PR #505, opened 2019-10-10
Extract validator into separate package

### LINKED ISSUE #466: Extract validator in separate package
Validator is quite useful on it's own so it should be separated in separate package.

Issues to be discussed:
- logger sharing with lodestar
- rpc api interfaces

mpetrunic:
- a lot of changes
- validator is now separate package

resolves #466


GregTheGreek (dismissed):

packages/lodestar-validator/README.md
```diff
@@ -0,0 +1,57 @@
+# Lodestar Validator
+[![Discord](https://img.shields.io/discord/593655374469660673.svg?label=Discord&logo=discord)](https://discord.gg/aMxzVcr)
+![ETH2.0_Spec_Version 0.8.3](https://img.shields.io/badge/ETH2.0_Spec_Version-0.8.3-2e86c1.svg)
+
+Typescript imlementation of the Eth2.0 beacon chain
```
wemeetagain: Readme should probably be stripped down for validator

packages/lodestar-validator/src/util/beaconChain.ts
```diff
@@ -0,0 +1,84 @@
+import {IBeaconConfig} from "@chainsafe/eth2.0-config";
```
wemeetagain, line 1: makes me sad :sob:

packages/lodestar/src/validator/nodejs.ts
```diff
@@ -0,0 +1,57 @@
+import defaultConfig, {IValidatorClientOptions} from "./options";
```
wemeetagain, line 1: Probably for a different PR:
I think this should live in the validator package.
Looks like we will probably need to split out our lodestar modules even more if we want to do that though.

packages/lodestar/test/unit/util/objects.test.ts
```diff
@@ -1,28 +1,12 @@
-// @ts-ignore-file
 import BN from "bn.js";
 import {expect} from "chai";
-
 import {uint64} from "@chainsafe/eth2.0-types";
 import {config} from "@chainsafe/eth2.0-config/lib/presets/mainnet";
-import {isPlainObject, mostFrequent} from "../../../src/util/objects";
-
+import {mostFrequent} from "../../../src/util/objects";
+import {describe, it} from "mocha";
```
wemeetagain, line 6: is this going in all of our tests now?

wemeetagain (changes requested):
roughly lgtm!
Makes me realize we need to split out the lodestar modules into packages to make things 'right'.

packages/lodestar-validator/README.md
mpetrunic: oh yeah, true

packages/lodestar-validator/src/util/beaconChain.ts
mpetrunic: I know :(

packages/lodestar/src/validator/nodejs.ts
mpetrunic: I think about it a lot. Probably it's much nicer UX to do `yarn add lodestar` and immediately have standalone validator. I stripped down validator package to be unbiased on logger and database implementation.

packages/lodestar/test/unit/util/objects.test.ts
mpetrunic: Dunno, cli works without it but I got Visual Code and Webstrom complaining when it's missing

packages/lodestar-validator/src/util/beaconChain.ts
ec2: :(

GregTheGreek: wait whats going on here...?

mpetrunic: haha. I had to copy few state transition methods in this file as we cannot easily rip state transition into separate module

packages/lodestar/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@chainsafe/lodestar",
```
ec2, line 2: Thoughts on renaming this to `lodestar-chain` or `lodestar-beacon` or something like that?

mpetrunic: reason?

GregTheGreek: tbh - i kind of back this idea to some degree... Lodestar has sort of evolved from this massive beast, to a small army of individual components.

packages/lodestar/src/cli/commands/validator.ts
```diff
@@ -2,13 +2,11 @@
  * @module cli/commands
```
ec2, line 2: Maybe for another PR, but we should totally have the CLI in a different package and have that require both the beacon chain and the validator.

packages/lodestar/package.json
ec2: ^ That. Lodestar will eventually include the light client or something too. Helps distinguish what is what

packages/lodestar/src/cli/commands/validator.ts
GregTheGreek: Yeah but here's a problem, debugging on the fly could be a pain. You'd have to keep lerna linking no?

packages/lodestar/test/unit/util/objects.test.ts
GregTheGreek: same for me

packages/lodestar/package.json
wemeetagain: I agree, "lodestar" is sort of a nebulous term now. I think `lodestar-beacon` makes sense.
Related, we could move cli into a separate package and reexport all relevant objects with a 'convenient' interface in the `lodestar` package.

packages/lodestar/src/validator/nodejs.ts
wemeetagain: if we rename lodestar to lodestar-beacon, and add a "convenience" lodestar package that re-exports everything, this code can/should go in the validator package.
Ofc needs the logger and db broken into packages.

packages/lodestar/src/cli/commands/validator.ts
wemeetagain: CLI in a separate package would be nice to separate concerns. Debugging is always a little trickier with more packages, but the monorepo makes debugging easier.
Lerna should handle the linking, just rebuild the needed packages.

ec2 (approved):
Other than the stuff we're discussing, which should be in another PR, LGTM.

wemeetagain (approved):

PR #509, opened 2019-10-18
Fast initial sync

### LINKED ISSUE #479: Improve initial sync
- if no peers, subscribe to peer:connected event and start from there
- pick target `head_slot` + `n` (make n option) slots
- split that range by number of peers and fetch each part from different peer
- once all blocks are obtained, verify only block headers
   - if headers are fine, run state transition with those blocks without any signature verification
   - if it doesn't match fetch blocks again
- repeat until all steps until finalized root encountered

mpetrunic:
resolves #479


packages/lodestar/src/sync/initial.ts
```diff
@@ -5,86 +5,106 @@
...
+    let chunks = chunkify(20, currentSlot, targetSlot);
+    const blocks: BeaconBlock[] = [];
+    //try to fetch chunks from different peers until all chunks are fetched
+    while(chunks.length > 0) {
+      //rotate peers
+      const peers = new RoundRobinArray(this.peers);
+      chunks = (await Promise.all(
+        chunks.map(async (chunk) => {
+          try {
+            const chunkBlocks = await this.syncChunk(peers.next(), chunk);
```
wemeetagain: does this request chunks in parallel? Or is it blocking on each request to finish?

packages/lodestar/src/sync/initial.ts
```diff
@@ -5,86 +5,106 @@
...
+      this.emit("synced", finalizedCheckPoint);
+    } else {
+      this.logger.debug(`Fast syncing to target ${targetEpoch}`);
+      const blocks = await this.syncToTarget(targetEpoch);
+      //TODO: verify block headers and run state transition without signature verification
+      blocks.forEach((block) => this.chain.receiveBlock(block, true));
+    }
+  };
+
+  private async syncToTarget(targetEpoch: Epoch): Promise<BeaconBlock[]> {
```
wemeetagain: maybe call this batchRequestBlocks, or something along those lines, we aren't really syncing here, just fetching blocks.

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -0,0 +1,39 @@
...
+      return true;
+    }
+  });
+  if(peersWithHigherFinalizedEpoch.length > 0) {
+    return currentCheckPoint.epoch + 1;
+  }
+  return currentCheckPoint.epoch;
+}
+
+export type Chunk = [Slot, Slot];
```
wemeetagain: Maybe call this BlockRange?

wemeetagain (commented):
I feel like we should be able to start processing blocks as we receive them, rather than saving them all up until the end.
Also I wonder what the tradeoffs with using the `step` param in beacon_block to interleave requests look like, rather than requesting contiguous ranges.

packages/lodestar/src/sync/initial.ts
mpetrunic: in parallel, map will return array of promises to resolve

packages/lodestar/src/sync/initial.ts
mpetrunic: yeah already renamed it

packages/lodestar/src/sync/utils/sync.ts
mpetrunic: > I feel like we should be able to start processing blocks as we receive them, rather than saving them all up until the end.

If you processing them all at once, you have to verify signatures and everything(as some peers might be malicious and sending wrong blocks). Here we collect them and verify if all block roots match (block.parentRoot) from different peers and execute fast state transition without signature and different root verifications (faster state transition). Target is still small (1 epoch) so we will process blocks in batches of 64 or something

> Also I wonder what the tradeoffs with using the `step` param in beacon_block to interleave requests look like, rather than requesting contiguous ranges.

Not sure, I would leave experimentation with this until we have some stable testnet against which we can benchmark syncing

mpetrunic:
@wemeetagain 
> I feel like we should be able to start processing blocks as we receive them, rather than saving them all up until the end.

If you processing them immediately after receiving, you have to verify signatures and everything(as some peers might be malicious and sending wrong blocks). Here we collect them and verify if all block roots match (block.parentRoot) from different peers and execute fast state transition without signature and different root verifications (faster state transition). Target is still small (1 epoch) so we will process blocks in batches of 64 or something

> Also I wonder what the tradeoffs with using the `step` param in beacon_block to interleave requests look like, rather than requesting contiguous ranges.

Not sure, I would leave experimentation with this until we have some stable testnet against which we can benchmark syncing

GregTheGreek:
Crazy idea what if we do a pre filter, where we do everything but verify the signature... And then process them later

mpetrunic:
> Crazy idea what if we do a pre filter, where we do everything but verify the signature... And then process them later

Could you clarify a bit better

wemeetagain:
> Target is still small (1 epoch) so we will process blocks in batches of 64 or something

I don't see where we're doing that?
It looks like we await all blocks before we start processing blocks. Trying to avoid an OOM error keeping all blocks in memory before processing.

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -0,0 +1,39 @@
+import {IReputation} from "../IReputation";
+import {Checkpoint, Epoch, Slot} from "@chainsafe/eth2.0-types";
+
+export function getTargetEpoch(peers: IReputation[], currentCheckPoint: Checkpoint): Epoch {
+  const peersWithHigherFinalizedEpoch = peers.filter(peer => {
+    if(!peer.latestHello) {
```
GregTheGreek, line 21: How recent are the hello's?

mpetrunic:
> > Target is still small (1 epoch) so we will process blocks in batches of 64 or something
> 
> I don't see where we're doing that?
> It looks like we await all blocks before we start processing blocks. Trying to avoid an OOM error keeping all blocks in memory before processing.

https://github.com/ChainSafe/lodestar/pull/509/files#diff-ae79fc115387d8b55109acf5b159e8e2R14

I will add some comments their, so in worst case scenario we are keeping 64 blocks in memory. 
We could optimize this better if we introduce event to signal which block **are we going to** process so we can download next epoch blocks while epoch transition is going on

packages/lodestar/src/sync/utils/sync.ts
mpetrunic: not recent, I don't think we are refreshing hellos, but I would do this in another PR as it's needed in regular sync as well. Possible solution, every SECONDS_PER_SLOT go trough all peers and refresh hellos.

GregTheGreek: Yep!

Intitially i was thinking they would be new because of startup, but the more i thought about it, we could lag at any point need to resync and our hello's our totally out of wack

Another thing to think about is when we actually start syncing. Do we have an ideal `PEER_THRESHOLD` where we hold off on syncing until we have an ideal amount of peers

mpetrunic: I put 1 for now, as we will test it against single node most of the time. But I will add that as cli option in this PR so user can configure (I think you can do this in current eth1 nodes).

I left peers as reference to peers in sync module so sync can refresh peer data (and it will reflect in regular and initial sync) without some complicated passing on.

mpetrunic:
It passes unit tests, but we need to improve "trusted" state transition to be able to catch up with other node so this potentially breaks sync. 

We could start dev branch until we merge all those big changes if we care about our current "stable" client?! cc @ChainSafe/eth2-0

packages/lodestar/src/chain/chain.ts
```diff
@@ -338,6 +339,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
     // Post-epoch processing
     const currentEpoch = computeEpochOfSlot(this.config, newState.slot);
     if (computeEpochOfSlot(this.config, preSlot) < currentEpoch) {
+      this.emit("processedEpoch", {epoch: currentEpoch, root: blockRoot});
```
wemeetagain: maybe `processedCheckpoint` to match the other events

packages/lodestar/src/sync/initial/initial.ts
```diff
@@ -0,0 +1,94 @@
...
+      return;
+    }
+    if(!this.chain.isInitialized()) {
+      this.logger.warn("Chain not initialized.");
+      this.emit("sync:completed", null);
+      return;
+    }
+
+    const chainCheckPoint = this.chain.latestState.currentJustifiedCheckpoint;
+    //listen on finalization events
```
wemeetagain: We also have a finalizedCheckpoint event. Is that the finalization event we want here?

packages/lodestar/src/chain/chain.ts
mpetrunic: sure. 

Side note: we have quite a lot of different event name styles. Do we wan't at some point to have processed:checkpoint or processedCheckpoint

packages/lodestar/src/sync/initial/initial.ts
mpetrunic: you would think so. But you will receive finalizedCheckpoint event only after you justify at least finalizedEpoch+1 which will never happen in initial sync

packages/lodestar/src/chain/chain.ts
wemeetagain: probably worth making an issue to straighten out the event names.
Maybe `chain:processedCheckpoint` would eventually be the best here.

packages/lodestar/src/sync/initial/initial.ts
ec2: That's confusing 😕 ...

packages/lodestar/src/sync/initial/initial.ts
```diff
@@ -0,0 +1,94 @@
...
+    //start syncing from current chain checkpoint
+    await this.sync(chainCheckPoint);
+  }
+
+  public async stop(): Promise<void> {
+    this.logger.info("initial sync stop");
+    this.chain.removeListener("processedEpoch", this.sync);
+  }
+
+  private sync = async (chainCheckPoint: Checkpoint) => {
```
ec2: sync = async 😂 Nothing wrong here, just think its funny haha

packages/lodestar/src/sync/initial/initial.ts
mpetrunic: think about finalized is really really justified epoch 😄

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -0,0 +1,83 @@
+import {IReputation, ReputationStore} from "../IReputation";
+import {BeaconBlock, BeaconBlockHeader, Checkpoint, Epoch, Slot} from "@chainsafe/eth2.0-types";
+import {IBeaconConfig} from "@chainsafe/eth2.0-config";
+import {hashTreeRoot} from "@chainsafe/ssz";
+import {blockToHeader} from "../../chain/stateTransition/util";
+import {IReqResp} from "../../network";
+
+export function isValidHeaderChain(config: IBeaconConfig, start: BeaconBlockHeader, blocks: BeaconBlock[]): boolean {
+  blocks = blocks.sort((b1, b2) => b1.slot - b2.slot);
+  let previousRoot = hashTreeRoot(start, config.types.BeaconBlockHeader);
```
wemeetagain: I think this should be signingRoot. Maybe call the variable parentRoot

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -0,0 +1,83 @@
...
+  for(const block of blocks) {
+    if(!previousRoot.equals(block.parentRoot)) {
+      return false;
+    }
+    previousRoot = hashTreeRoot(blockToHeader(config, block), config.types.BeaconBlockHeader);
```
wemeetagain: this should be signingRoot, also we don't need the extra `blockToHeader`, we can signingRoot the BeaconBlock directly.

packages/lodestar/src/sync/initial/initial.ts
wemeetagain: can we call this fastSync, or somehow differentiate it from a naive sync?
I think we probably will have support for several initial syncing strategies (eg: fast sync, naive sync, light sync, etc)

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -0,0 +1,83 @@
...
+  const peersWithHigherFinalizedEpoch = peers.filter(peer => {
+    if(!peer.latestHello) {
+      return false;
+    }
+    if(peer.latestHello.finalizedEpoch > currentCheckPoint.epoch) {
+      return true;
+    }
+  });
+  if(peersWithHigherFinalizedEpoch.length > 0) {
+    return currentCheckPoint.epoch + 1;
```
wemeetagain: I don't understand the + 1 logic.

I think we need to work with a Checkpoint, ie not throw away the root, because the final part of the fast sync strategy is to check the root against the last downloaded block to ensure we've downloaded the chain we think we've downloaded.

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -0,0 +1,83 @@
...
+  for(const block of blocks) {
+    if(!previousRoot.equals(block.parentRoot)) {
+      return false;
+    }
+    previousRoot = hashTreeRoot(blockToHeader(config, block), config.types.BeaconBlockHeader);
+  }
+  return true;
+}
+
+export function getTargetEpoch(peers: IReputation[], currentCheckPoint: Checkpoint): Epoch {
```
wemeetagain: I think we should change the name here, we use 'target' as a term in FFG votes, makes this kinda confusing.

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -0,0 +1,83 @@
+import {IReputation, ReputationStore} from "../IReputation";
+import {BeaconBlock, BeaconBlockHeader, Checkpoint, Epoch, Slot} from "@chainsafe/eth2.0-types";
+import {IBeaconConfig} from "@chainsafe/eth2.0-config";
+import {hashTreeRoot} from "@chainsafe/ssz";
+import {blockToHeader} from "../../chain/stateTransition/util";
+import {IReqResp} from "../../network";
+
+export function isValidHeaderChain(config: IBeaconConfig, start: BeaconBlockHeader, blocks: BeaconBlock[]): boolean {
```
wemeetagain: I think we should give this name a slightly more descriptive name.
Also, consider adding an end param. (Then we could check the downloaded blocks against a known checkpoint root)
eg:
```typescript
/**
 * Ensure a chain of blocks links properly from start to end
 */
isValidChainOfBlocks(config: IBeaconConfig, start: Hash, end: Hash, blocks: BeaconBlock[]): boolean;
```

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -0,0 +1,83 @@
...
+  blocks = blocks.sort((b1, b2) => b1.slot - b2.slot);
```
wemeetagain: feels sort of strange that this function mutates `blocks` (is relied upon in the calling code)

wemeetagain (changes requested):

packages/lodestar/src/sync/utils/sync.ts
mpetrunic: there is no method to fetch non latest finalized checkpoint from peers. I'll extract +1 to constant, basically if you put +2 here it will sync 2 epoch at a time. We can only sync next epoch if there is peer with higher finalized epoch

packages/lodestar/src/sync/utils/sync.ts
mpetrunic: again we don't have end hash at that point, only when we reach highest finalized epoch we can check if our checkpoint matches other peers checkpoint

mpetrunic:
@ChainSafe/eth2-0 this is also finished for quite some time

packages/lodestar/src/chain/interface.ts
```diff
@@ -9,6 +9,7 @@ import {ProgressiveMerkleTree} from "@chainsafe/eth2.0-utils";
 export interface IChainEvents {
   unknownBlockRoot: (root: Hash) => void;
   processedBlock: (block: BeaconBlock) => void;
+  "chain:processedCheckpoint": (checkPoint: Checkpoint) => void;
```
wemeetagain: can we stick with `processedCheckpoint` for consistency for now? Maybe tackle 'robust-ifying' our event names separately.

packages/lodestar/src/chain/chain.ts
```diff
@@ -132,7 +132,8 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
     });
   }
 
-  public async receiveBlock(block: BeaconBlock): Promise<void> {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public async receiveBlock(block: BeaconBlock, trusted: boolean = false): Promise<void> {
```
wemeetagain, line 141: You can string this through now if you merge master.

packages/lodestar/src/sync/initial/fast/index.ts
```diff
@@ -0,0 +1,98 @@
...
+import {ILogger} from "../../../logger";
+import {ISyncOptions} from "../../options";
+import {IInitialSyncModules, InitialSync, InitialSyncEventEmitter} from "../interface";
+import {EventEmitter} from "events";
+import {getSyncTargetEpoch, isValidChainOfBlocks, isValidFinalizedCheckPoint} from "../../utils/sync";
+import {BeaconState, Checkpoint} from "@chainsafe/eth2.0-types";
+import {computeStartSlotOfEpoch} from "@chainsafe/eth2.0-state-transition";
+import {getBlockRange} from "../../utils/blocks";
+
+export class FastSync
```
wemeetagain, line 17: :+1:

wemeetagain (changes requested):

wemeetagain (approved):

PR #513, opened 2019-10-21
Convert interop command to dev command

### LINKED ISSUE #511: Convert interop to "dev" cli command
- rename "interop" cli to "dev"
- make every param optional (timestamp = new Date(), validators =8)
- split quick params into number of validators and timestamp params

twoeths:
## Goal
+ Close #511 
## Test
+ I can see a justified epoch after running `./bin/lodestar dev`
image


packages/lodestar/src/cli/commands/interop.ts
```diff
@@ -30,9 +30,13 @@ import {ProgressiveMerkleTree} from "@chainsafe/eth2.0-utils";
 import {MerkleTreeSerialization} from "../../util/serialization";
 import {ApiClientOverInstance} from "@chainsafe/lodestar-validator/lib/api";
 import {ValidatorClient} from "../../validator/nodejs";
+import {BeaconState} from "@chainsafe/eth2.0-types";
+import {quickStartState} from "../../interop/state";
 
 interface IInteropCommandOptions {
```
mpetrunic: rename to IDevCommand options

packages/lodestar/src/cli/commands/interop.ts
```diff
@@ -120,19 +126,22 @@ export class InteropCommand implements ICliCommand {
       .then((peerId) => initializePeerInfo(peerId, conf.network.multiaddrs))
       .then((peerInfo) => new NodejsNode({peerInfo, bootnodes: conf.network.bootnodes}));
     const config = options.preset === "minimal" ? minimalConfig : mainnetConfig;
+    const tree = ProgressiveMerkleTree.empty(DEPOSIT_CONTRACT_TREE_DEPTH, new MerkleTreeSerialization(config));
+    let state: BeaconState;
     if (options.quickStart) {
```
mpetrunic: We should remove quickstart option and use genesisTime/validatorCount separately

mpetrunic (changes requested):
Could you print genesis time and number of validators so we can connect other nodes?

Also if you try to run it, you will see it won't work as it doesn't have multiaddrs param, you should make that param optional as well.

running `.bin/lodestar dev` should run and produce blocks

Update readme as well

packages/lodestar/src/cli/commands/interop.ts
twoeths: `quickstart` option also help initialize state from ssz/yml file (see `function quickStartOptionToState(`). If we remove it we'll not be able to do that, can you confirm?

mpetrunic: rename it than to genesisState flag and use just for ssz/yaml file loading

twoeths: in fact, this is the only place we call `quickStartOptionToState()` function. If we drop `quickstart` option then this function is maybe redundant.

packages/lodestar/src/cli/commands/interop.ts
```diff
@@ -50,9 +54,11 @@ export class InteropCommand implements ICliCommand {
...
-      .command("interop")
+      .command("dev")
       .description("Start lodestar beacon node and certain amount of validator nodes")
-      .option("-q, --quickStart [params]", "Start chain from known state")
+      .option("-t, --genesisTime [genesisTime]", "genesis time of Beacon state", Math.round(Date.now()/1000))
+      .option("-c, --validatorCount [validatorCount]", "Number of validator for Beacon state", 8)
+      .option("-s, --genesisState [params]", "Start chain from known state")
       // eslint-disable-next-line max-len
       .option("-v, --validators [range]", "Start validators, single number - validators 0-number, x,y - validators between x and y", 0)
       .option("-p, --preset [preset]", "Minimal/mainnet", "mainnet")
```
mpetrunic: since it's a dev command, maybe better for preset to be minimal

packages/lodestar/src/cli/commands/interop.ts
```diff
@@ -50,9 +54,11 @@ export class InteropCommand implements ICliCommand {
...
     const command = commander
-      .command("interop")
+      .command("dev")
       .description("Start lodestar beacon node and certain amount of validator nodes")
-      .option("-q, --quickStart [params]", "Start chain from known state")
+      .option("-t, --genesisTime [genesisTime]", "genesis time of Beacon state", Math.round(Date.now()/1000))
+      .option("-c, --validatorCount [validatorCount]", "Number of validator for Beacon state", 8)
+      .option("-s, --genesisState [params]", "Start chain from known state")
       // eslint-disable-next-line max-len
       .option("-v, --validators [range]", "Start validators, single number - validators 0-number, x,y - validators between x and y", 0)
```
mpetrunic: should be 8 by default

packages/lodestar/src/cli/commands/interop.ts
```diff
@@ -116,23 +122,32 @@ export class InteropCommand implements ICliCommand {
...
       peerId = createPeerId();
     }
+    const network: {multiaddrs?: string[]; bootnodes?: string[]} = conf.network || {};
+    const multiaddrs = network.multiaddrs || [];
+    const bootnodes = network.bootnodes || [];
     const libp2p = await Promise.resolve(peerId)
-      .then((peerId) => initializePeerInfo(peerId, conf.network.multiaddrs))
-      .then((peerInfo) => new NodejsNode({peerInfo, bootnodes: conf.network.bootnodes}));
+      .then((peerId) => initializePeerInfo(peerId, multiaddrs))
+      .then((peerInfo) => new NodejsNode({peerInfo, bootnodes: bootnodes}));
```
mpetrunic: @wemeetagain Why are we initializing libp2p here and not inside network module?

packages/lodestar/src/cli/commands/interop.ts
```diff
@@ -116,23 +122,32 @@ export class InteropCommand implements ICliCommand {
     } else {
       peerId = createPeerId();
     }
+    const network: {multiaddrs?: string[]; bootnodes?: string[]} = conf.network || {};
+    const multiaddrs = network.multiaddrs || [];
```
mpetrunic: default multiaddrs should be ["/ip4/127.0.0.1/tcp/30606"] but wait for caymans rsponse down below

mpetrunic (changes requested):

packages/lodestar/src/cli/commands/interop.ts
wemeetagain: I don't think we need the fallback `|| []` bc we have that default set in src/network/options.ts

packages/lodestar/src/cli/commands/interop.ts
wemeetagain: We have these async functions we need to wait for. There was trouble awaiting them in the network constructor. The `NodejsNode` is just a libp2p instance.
Another thought was that we might want to reuse this instance for other beacon nodes or shard nodes.

mpetrunic: can we arange this so its awaited in start method of network?

packages/lodestar/src/cli/commands/interop.ts
mpetrunic: those options are only available i side network module, thats why im asking to move libp2p stuff inside network module

packages/lodestar/src/cli/commands/interop.ts
wemeetagain: Before this solution, we created a promise in the BeaconNode constructor and passed in the promise to the network, which made sure it was finished before starting.

mpetrunic: Ill think about cleaner solution on monday

wemeetagain: Think about reusing the network instance between multiple nodes, too.

twoeths: I move it to a separate async factory function, please have a look.

packages/lodestar/src/network/nodejs/util.ts
```diff
@@ -22,3 +27,16 @@ export async function loadPeerId(path: string): Promise<PeerId> {
   const data = await promisify(fs.readFile)(path);
   return await promisify(PeerId.createFromJSON)(JSON.parse(data));
 }
+
+/**
+ * 
+ * @param peerId Create an instance of NodejsNode asynchronously
+ */
+export async function createLibP2p(peerIdOrPromise: PeerId | Promise<PeerId>,
```
wemeetagain: :+1:

wemeetagain (dismissed):

packages/lodestar/src/network/nodejs/util.ts
mpetrunic: I've modified network a bit in PR to this branch, there is a link in eth2 channel as I can't tag you on his repo

wemeetagain (dismissed):

packages/lodestar/src/cli/commands/interop.ts
```diff
@@ -5,10 +5,11 @@
 import {ICliCommand} from "./interface";
```
mpetrunic, line 5: rename file to dev

packages/lodestar/src/network/nodejs/util.ts
mpetrunic: I would rename it to createNodeJsLibp2p

mpetrunic (changes requested):

packages/lodestar/src/cli/commands/interop.ts
mpetrunic: the cli file, not this import

mpetrunic (approved):

wemeetagain (approved):

PR #519, opened 2019-10-24
Generic cache implementation

### LINKED ISSUE #485: Implement caching module
- choose some library or implement own cache
- reuse Bucket, Key and encodeKey from database module
- should support expiring

mpetrunic:
resolves #485

### DIFF
--- a/packages/lodestar/src/cache/abstract.ts
+++ b/packages/lodestar/src/cache/abstract.ts
@@ -50,12 +50,8 @@ export abstract class CacheItem<T> {
     this.cache.clear();
   }
 
-  public delete(id?: CacheId): void {
-    if(id) {
-      this.cache.delete(this.idToString(id));
-    } else {
-      this.clear();
-    }
+  public delete(id: CacheId): void {
+    this.cache.delete(this.idToString(id));
   }
 
   protected getOrFirst(id: CacheId): T | null {
--- a/packages/lodestar/src/cache/abstract.ts
+++ b/packages/lodestar/src/cache/abstract.ts
@@ -1,3 +1,7 @@
+/**
+ * @module cache
+ */
+
 import {AnySSZType} from "@chainsafe/ssz-type-schema";
 import {CacheId, CacheIdFunction} from "./types";
 import {clone, hashTreeRoot} from "@chainsafe/ssz";
@@ -10,6 +14,12 @@ export abstract class CacheItem<T> {
 
   protected cache: Map<CacheId, T> = new Map();
 
+  /**
+   *
+   * @param sszType
+   * @param idFunction function  to generate id for cache item. Depending on your use case
+   * it can return unique identifier or some constant ("latestState")
+   */
   protected constructor(sszType: AnySSZType, idFunction?: CacheIdFunction<T>) {
     this.sszType = sszType;
     this.idFunction = idFunction || this.hashRootIdFunction;
--- a/packages/lodestar/src/cache/abstract.ts
+++ b/packages/lodestar/src/cache/abstract.ts
@@ -1,0 +1,68 @@
+import {AnySSZType} from "@chainsafe/ssz-type-schema";
+import {CacheId, CacheIdFunction} from "./types";
+import {clone, hashTreeRoot} from "@chainsafe/ssz";
+
+export abstract class CacheItem<T> {
+
+  protected sszType: AnySSZType;
+
+  protected idFunction: CacheIdFunction<T>;
+
+  protected cache: Map<CacheId, T> = new Map();
+
+  protected constructor(sszType: AnySSZType, idFunction?: CacheIdFunction<T>) {
+    this.sszType = sszType;
+    this.idFunction = idFunction || this.hashRootIdFunction;
+  }
+
+  /**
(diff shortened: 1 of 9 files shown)


packages/lodestar/src/cache/abstract.ts
```diff
@@ -0,0 +1,68 @@
...
+import {CacheId, CacheIdFunction} from "./types";
+import {clone, hashTreeRoot} from "@chainsafe/ssz";
+
+export abstract class CacheItem<T> {
+
+  protected sszType: AnySSZType;
+
+  protected idFunction: CacheIdFunction<T>;
+
+  protected cache: Map<CacheId, T> = new Map();
```
GregTheGreek, line 15: could we abstract this further? What if we wanted to use Redis ?

packages/lodestar/src/cache/abstract.ts
```diff
@@ -0,0 +1,68 @@
+import {AnySSZType} from "@chainsafe/ssz-type-schema";
+import {CacheId, CacheIdFunction} from "./types";
+import {clone, hashTreeRoot} from "@chainsafe/ssz";
+
+export abstract class CacheItem<T> {
+
+  protected sszType: AnySSZType;
+
+  protected idFunction: CacheIdFunction<T>;
```
GregTheGreek, line 13: Comments explaining this?

packages/lodestar/src/cache/abstract.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  public clear(): void {
+    this.cache.clear();
+  }
+
+  public delete(id?: CacheId): void {
+    if(id) {
+      this.cache.delete(this.idToString(id));
+    } else {
```
GregTheGreek: I don't think we should have this - imo we should either mark `clear` private or don't allow for delete to call clear

packages/lodestar/src/cache/abstract.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  protected sszType: AnySSZType;
+
+  protected idFunction: CacheIdFunction<T>;
+
+  protected cache: Map<CacheId, T> = new Map();
+
+  protected constructor(sszType: AnySSZType, idFunction?: CacheIdFunction<T>) {
+    this.sszType = sszType;
+    this.idFunction = idFunction || this.hashRootIdFunction;
```
GregTheGreek, line 25: where does `this.hashRootIdFunction` come from? it will always be null?

GregTheGreek (dismissed):

packages/lodestar/src/cache/abstract.ts
mpetrunic: Can we really package redis with lodestar? I would leave further abstraction until we have some use case, should be fairly easy to abstract more

packages/lodestar/src/cache/abstract.ts
mpetrunic: it's defined in class. Although I could move it to separate file with some common id generators

packages/lodestar/src/cache/abstract.ts
GregTheGreek: no idea - just a thought

twoeths: Isn't it just installing redis npm package and then it's ready to use? It's in-memory so should be lightweight and doesn't require any special setup, it also supports expiring feature.

twoeths: if we support expiring feature then probably we'll end up with implement something like redis.

mpetrunic: I'm pretty sure thats just redis client. You need redis server for client to work. There are redis mocks, but they are a bit bloated and not maintained

mpetrunic:
@ChainSafe/eth2-0 Can we review this?

wemeetagain:
tbh i don't think we should go in this direction, rather explore using persistent immutable data structures.
This is because the object we're most likely to want to cache (state) doesn't change very much between blocks, and if we use a simple map of states, we will just be storing a lot of duplicate data.
This alternative path also fits nicely with the idea of persisting the merkle tree.

EDIT:
On second thought, we could still use something like this despite using different data structures.
I think my issue with this is I'd prefer to start with something simple and extensible, and grow something from there, and this feels like a full-blown caching framework.

mpetrunic:
> tbh i don't think we should go in this direction, rather explore using persistent immutable data structures.
> This is because the object we're most likely to want to cache (state) doesn't change very much between blocks, and if we use a simple map of states, we will just be storing a lot of duplicate data.
> This alternative path also fits nicely with the idea of persisting the merkle tree.
> 
> EDIT:
> On second thought, we could still use something like this despite using different data structures.
> I think my issue with this is I'd prefer to start with something simple and extensible, and grow something from there, and this feels like a full-blown caching framework.

I was thinking of using this for api for now, and expand later for storing states and stuff. Like caching ValidatorDuties and commitees so validators don't block us on start of every slot.

Imean it's simple to extend for storing parts of state for example, it's more about interfaces (there is not much implementation anyways)

PR #523, opened 2019-10-25
Blocks archiving

### LINKED ISSUE #480: Implement block by range fetching
- switch block id from root to slot
- add database mapping of root->slot
- change sync to fetch blocks by range

Reasoning:
- sync requires fetching blocks by slot range and that will request a lot of blocks while fetching of slots by root is only required for recent blocks(after finalization) for which we can hold root->slor or root->block mapping

@wemeetagain what do you think?

mpetrunic:
- introduces block archive where blocks are mapped slot -> block
- introduces slot -> root and root -> slot ref storing
- chores module
   - archive finalized blocks - move from block bucket to blockArchive bucket
- BlockByRange response is more optimized because it fetches all blocks in single query

resolves #480


packages/lodestar/src/chores/index.ts
```diff
@@ -0,0 +1,44 @@
+import {IService} from "../node";
+import {IBeaconConfig} from "@chainsafe/eth2.0-config";
+import {IBeaconDb} from "../db/api";
+import {IBeaconChain} from "../chain";
```
GregTheGreek: TS doc comments?

GregTheGreek (dismissed):
lgtm, but missing comments

twoeths (dismissed):

packages/lodestar/src/chores/tasks/arhiveBlocks.ts
```diff
@@ -0,0 +1,50 @@
+/**
```
wemeetagain: misspelled filename

packages/lodestar/src/chores/index.ts
```diff
@@ -0,0 +1,48 @@
+/**
```
wemeetagain: Can we rename `chores` to something more standard? Perhaps `taskManager` or even `tasks`.

packages/lodestar/src/chores/index.ts
```diff
@@ -0,0 +1,48 @@
...
+import {ArchiveBlocksTask} from "./tasks/arhiveBlocks";
+import {ILogger} from "../logger";
+
+export interface IChoreModules {
+  db: IBeaconDb;
+  logger: ILogger;
+  chain: IBeaconChain;
+}
+
+export class ChoreService implements IService {
```
wemeetagain: Can you add a comment here saying what this module is for?

packages/lodestar/src/chores/tasks/arhiveBlocks.ts
```diff
@@ -0,0 +1,50 @@
...
+    this.config = config;
+    this.finalizedCheckpoint = finalizedCheckpoint;
+  }
+
+  public async run(): Promise<void> {
+    const blocks = (await this.db.block.getAll()).filter(
+      (block) =>
+        computeEpochOfSlot(this.config, block.slot) <= this.finalizedCheckpoint.epoch
+    );
+    this.logger.info(`Started archiving ${blocks.length} finalized blocks...`);
```
wemeetagain: Can we reference the finalized checkpoint in some way in this log? either root or epoch or both.
Might make sense to also log the # of finalized epochs rather than blocks.

packages/lodestar/src/db/api/beacon/repositories/blockArhive.ts
```diff
@@ -0,0 +1,32 @@
+import {BeaconBlock} from "@chainsafe/eth2.0-types";
```
wemeetagain: filename misspelled

packages/lodestar/src/db/api/beacon/repositories/blockArhive.ts
```diff
@@ -0,0 +1,32 @@
...
+import {IBeaconConfig} from "@chainsafe/eth2.0-config";
+import {BulkRepository} from "../repository";
+import {IDatabaseController} from "../../../controller";
+import {Bucket, encodeKey} from "../../../schema";
+import {serialize} from "@chainsafe/ssz";
+
+/**
+ * Stores finalized blocks. Block slot is identifier.
```
wemeetagain: maybe say blocks are indexed by slot number, and why.

packages/lodestar/src/db/schema.ts
```diff
@@ -10,6 +10,9 @@ export enum Bucket {
   state, // hash -> BeaconState
   attestation, // hash -> Attestation
   block, // hash -> BeaconBlock
+  blockArchive, // hash -> BeaconBlock
+  blockSlotRefs,
```
wemeetagain, line 14: can you include the comments to the side? eg: `slot -> hash` or whatever

packages/lodestar/src/db/api/beacon/repositories/blockArhive.ts
```diff
@@ -0,0 +1,32 @@
...
+  public async addMultiple(blocks: BeaconBlock[]): Promise<void> {
+    await this.db.batchPut(
+      blocks.map((block) => ({
+        key: encodeKey(this.bucket, block.slot),
+        value: serialize(block, this.type)
+      }))
+    );
+  }
+
+  public async setUnderRoot(value: BeaconBlock): Promise<void> {
```
wemeetagain: I think we should rename `setUnderRoot` (everywhere) to something more generic, like `add` (also goes well with `addMultiple`)
Maybe we can also break it out into two functions:
``
```
createId(value: T): Id;

// and

add(value: T): Promise<void> {
  this.set(this.createId(value), value);
}
```
We probably would mostly just need to override `createId`.

We should also merge the naming disparities between `deleteMany` and `addMultiple`.

wemeetagain (changes requested):
The structure looks good, mostly comments re: naming

packages/lodestar/src/chores/index.ts
mpetrunic: i'm fine with tasks. Chores seems more appropriate at the time :smile:

packages/lodestar/src/chores/tasks/arhiveBlocks.ts
mpetrunic: yeah, I'll add finalized epoch number. Though it's not necessary since we have red message saying what epich is finalized :slightly_smiling_face:

packages/lodestar/src/db/api/beacon/repositories/blockArhive.ts
mpetrunic: `Block slot is identifier` says in comment. I'll add why

packages/lodestar/src/db/api/beacon/repositories/blockArhive.ts
mpetrunic: I think we should leave set under root as it's quite clear what's going on. We have method where you can set id whatever you wan't, but this is extremely common scenario where we store under hashTreeRoot or signingRoot. We can create another method with custom id function though. 

> We should also merge the naming disparities between deleteMany and addMultiple
agree

wemeetagain: Its actually not clear. (I was staring at this for a few minutes, and I'm sure other people will do the same) It assumes that all indices are roots.
And doesn't lend any additional clarity (eg: why should the chain care that the block was stored under a root? It just wants to add the block to the db)

I'm imagining a nice generic interface like:
```
createId(value: T): Id;

add(value: T): Promise<void>;
remove(value: T): Promise<void>;
bulkAdd(values: T[]): Promise<void>;
bulkRemove(values: T[]): Promise<void>;

set(id: Id, value: T): Promise<void>;
delete(id: Id): Promise<void>;
bulkSet(ids: Id[], values: T[]): Promise<void>;
bulkDelete(ids: Id[]): Promise<void>;
```

ec2: @wemeetagain is right. I thought it was indexed by roots 😂 until i read it closer

packages/lodestar/src/chores/index.ts
ec2: Is this supposed to be a generic task manager?

packages/lodestar/src/db/api/beacon/repositories/blockArhive.ts
mpetrunic: it is indexed by root 😂

mpetrunic: I like that interface as well.

ec2: is it not by slot there?

mpetrunic: oh yeah, thought we are talking about block bucket 😂

mpetrunic:
@wemeetagain @wemeetagain can we merge this, I've update this PR like 5 time already

wemeetagain (approved):

ec2 (approved):

PR #527, opened 2019-10-27
e2e to simulate node and validators until a justified epoch

### LINKED ISSUE #512: Create e2e justification simulation
- find quickest valid config(3 validators, 2-3 seconds per slot, 5-6 slots per epoch)
- create e2e simulation which will start beacon node and 3 validators and run until 1 justified epoch

twoeths:
## Goal
+ Close #512

### DIFF
--- a/packages/lodestar/package.json
+++ b/packages/lodestar/package.json
@@ -129,6 +129,7 @@
     "chai-as-promised": "^7.1.1",
     "mocha": "^6.2.0",
     "mockery": "^2.1.0",
+    "rimraf": "^3.0.0",
     "supertest": "^4.0.2",
     "webpack": "^4.39.1"
   }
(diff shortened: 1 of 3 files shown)


mpetrunic:
I don't think we should merge this with this hack, because solution is to fix error that was caused with extracting state transition, apparently state transition or some other stub isn't working properly, probably different import path or something

twoeths:
> I don't think we should merge this with this hack, because solution is to fix error that was caused with extracting state transition, apparently state transition or some other stub isn't working properly, probably different import path or something

I removed the hack, please re-review.

mpetrunic:
I was hoping you would fix bug that occured after state transition got moved to separate package.
After we moved state transition, state transition stub stopped working and was causing long unit tests.

Better way was to fix this, but I guess we will have to fix it in separate PR :disappointed:

mpetrunic (approved):

twoeths:
@mpetrunic I noted your comment in #537 , will look into that. This PR is merely about e2e test anyway.

wemeetagain (approved):

PR #536, opened 2019-10-31
add trusted flag to skip bls verify in state transition functions

### LINKED ISSUE #515: Add "trusted" flag in state transition
"trusted" param should be optional with default param false.
If param is true, state transition should not verify signatures when processing block

ec2:
closes #515

### DIFF
--- a/packages/eth2.0-state-transition/src/block/randao.ts
+++ b/packages/eth2.0-state-transition/src/block/randao.ts
@@ -24,7 +24,7 @@ export function processRandao(
   const currentEpoch = getCurrentEpoch(config, state);
   const proposer = state.validators[getBeaconProposerIndex(config, state)];
   // Verify RANDAO reveal
-  assert(trusted ? true : bls.verify(
+  assert(trusted || bls.verify(
     proposer.pubkey,
     hashTreeRoot(currentEpoch, config.types.Epoch),
     body.randaoReveal,
--- a/packages/eth2.0-state-transition/src/block/randao.ts
+++ b/packages/eth2.0-state-transition/src/block/randao.ts
@@ -18,12 +18,13 @@ import {getBeaconProposerIndex, getCurrentEpoch, getDomain, getRandaoMix,} from 
 export function processRandao(
   config: IBeaconConfig,
   state: BeaconState,
-  body: BeaconBlockBody
+  body: BeaconBlockBody,
+  trusted: boolean = false
 ): void {
   const currentEpoch = getCurrentEpoch(config, state);
   const proposer = state.validators[getBeaconProposerIndex(config, state)];
   // Verify RANDAO reveal
-  assert(bls.verify(
+  assert(trusted ? true : bls.verify(
     proposer.pubkey,
     hashTreeRoot(currentEpoch, config.types.Epoch),
     body.randaoReveal,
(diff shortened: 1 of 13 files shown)


packages/eth2.0-state-transition/src/slot.ts
```diff
@@ -17,7 +17,8 @@ import {processEpoch} from "./epoch";
 export function processSlots(
   config: IBeaconConfig,
   state: BeaconState,
-  slot: Slot
+  slot: Slot,
+  trusted: boolean = false
```
mpetrunic: we probably don't need this flag here

packages/eth2.0-state-transition/src/block/randao.ts
```diff
@@ -18,12 +18,13 @@ import {getBeaconProposerIndex, getCurrentEpoch, getDomain, getRandaoMix,} from
...
   state: BeaconState,
-  body: BeaconBlockBody
+  body: BeaconBlockBody,
+  trusted: boolean = false
 ): void {
   const currentEpoch = getCurrentEpoch(config, state);
   const proposer = state.validators[getBeaconProposerIndex(config, state)];
   // Verify RANDAO reveal
-  assert(bls.verify(
+  assert(trusted ? true : bls.verify(
```
mpetrunic: trusted || bls.verify?

mpetrunic: and in all other occurencies

mpetrunic (changes requested):

GregTheGreek:
Why not just wrap BLS, it would probably be easier that way

mpetrunic:
@GregTheGreek deposit verify must be checked anyways

packages/eth2.0-state-transition/src/index.ts
```diff
@@ -24,14 +24,15 @@ export function stateTransition(
   state: BeaconState,
```
wemeetagain, line 33: Can you add a block comment above saying what each option does/is for?
Will be helpful for future us and for autogenerating docs.

wemeetagain (changes requested):

packages/eth2.0-state-transition/src/block/randao.ts
ec2: Wow... No duh, eh?

mpetrunic: seems nicer haha "trusted or verify" :D

mpetrunic (dismissed):

packages/eth2.0-state-transition/src/index.ts
```diff
@@ -19,19 +19,29 @@ export * from "./slot";
...
+ * @param trusted Skip operations signature verification
+ */
 export function stateTransition(
   config: IBeaconConfig,
   state: BeaconState,
   block: BeaconBlock,
   validateStateRoot = false,
-  verifySignatures = true
+  verifySignatures = true,
+  trusted = false
```
wemeetagain: Can we rename these params?
`verifyProposer`
`verifySignatures`

elsewhere instead of `trusted` defaulted to false, we can use `verifySignatures`, defaulted to true

wemeetagain (changes requested):

wemeetagain (approved):

mpetrunic (approved):

Issue #522, opened 2019-10-25
Remove logs from e2e tests


mpetrunic:
network related tests are emitting logs. 

Solution: logger should be stubbed or silenced

aunyks:
@mpetrunic I can take this if it hasn't already been claimed elsewhere.

mpetrunic:
sure

PR #538, opened 2019-11-05
Silence logger during E2E tests

aunyks:
Fixing #522.

### DIFF
--- a/packages/lodestar/src/logger/winston.ts
+++ b/packages/lodestar/src/logger/winston.ts
@@ -47,7 +47,10 @@ export class WinstonLogger implements ILogger {
     });
     //@ts-ignore
     this._level = LogLevel[options.level];
-    this._silent = process.env.LODESTAR_SILENCE === 'true';
+    this._silent = false
+    if (typeof process !== 'undefined' && typeof process.env !== 'undefined') {
+      this._silent = process.env.LODESTAR_SILENCE === 'true';
+    }
   }
 
   public debug(message: string | object, context?: object): void {
--- a/packages/lodestar/src/logger/winston.ts
+++ b/packages/lodestar/src/logger/winston.ts
@@ -2,8 +2,8 @@
  * @module logger
  */
 
-import { createLogger, format, Logger, transports } from "winston";
-import { defaultLogLevel, LogLevel, ILogger, ILoggerOptions } from "./interface";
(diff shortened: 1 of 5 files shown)


aunyks:
Not very sure what Labeler is communicating.

mpetrunic (commented):
this works, although better solution might be to just go in e2e tests and silence logger there or create mocked logger. Reasoning is that we use IDE shortcuts to run tests (or subset) and this would require us to configure env variable in IDe or console.

packages/lodestar/test/e2e/network/network.test.ts
```diff
@@ -27,11 +27,12 @@ describe("[network] network", function () {
   this.timeout(5000)
   let netA: Libp2pNetwork, netB: Libp2pNetwork;
   const logger: ILogger = new WinstonLogger();
-  const metrics = new BeaconMetrics({enabled: true, timeout: 5000, pushGateway: false}, {logger});
+  logger.silent = true;
```
aunyks, line 30: @mpetrunic Line 30 should silence the logger during E2E without relying just on the envar.

mpetrunic (dismissed):

packages/lodestar/src/logger/winston.ts
```diff
@@ -47,7 +47,7 @@ export class WinstonLogger implements ILogger {
     });
     //@ts-ignore
     this._level = LogLevel[options.level];
-    this._silent = false;
+    this._silent = process.env.LODESTAR_SILENCE === 'true';
```
wemeetagain: Can you add a documentation note somewhere, maybe the README or elsewhere, about the `LODESTAR_SILENCE` env var?

Another thought, can you do a check for if `process` exists? I think in the browser `process` will be `undefined`

wemeetagain (changes requested):

packages/lodestar/src/logger/winston.ts
aunyks: Good eye on `process` possibly being undefined. I'll update this logic and make note of the envar in the README.

wemeetagain (dismissed):
LGTM :)

mpetrunic (dismissed):

mpetrunic:
@aunyks you have lint errors, check CI

mpetrunic (approved):

wemeetagain (approved):

PR #542, opened 2019-11-08
Update eth2.0-utils to be more browser-friendly

wemeetagain:
- Replace as-sha256 with bcrypto in eth2.0-utils
- Move nodejs-specific code to eth2.0-utils/src/nodejs (and do not export to root)

This doesn't update versions, will be done in a different PR


mpetrunic (approved):

ec2 (approved):
lgtm

PR #545, opened 2019-11-11
Update networking to 0.9.x

### LINKED ISSUE #532: Update networking to 0.9.0
- https://github.com/ethereum/eth2.0-specs/pull/1404
- https://github.com/ethereum/eth2.0-specs/pull/1440

mpetrunic:
- update networking types to 0.9.x
- update gossipsub to 0.9.x
- reworked gossipsub to be more testable and expandable

Note:
- this isn't working as it's missing 0.9.x state transation
- it's hard to extend event emitter because of StrictEventEmitter it has complex types and everything is not exported (solution might be to not enforce strinct event emitting inside Gossip class - just interface)

resolves #532


wemeetagain (approved):
:100: ez review

packages/eth2.0-types/src/ssz/generators/validator.ts
```diff
@@ -15,10 +15,29 @@ export const ValidatorDuty = (ssz: IBeaconSSZTypes): SimpleContainerType => ({
...
+
 export const SyncingStatus = (ssz: IBeaconSSZTypes): SimpleContainerType => ({
   fields: [
     ["startingBlock", ssz.uint64],
     ["currentBlock", ssz.uint64],
     ["highestBlock", ssz.uint64],
   ],
 });
+
+export const AggregateAndProof = (ssz: IBeaconSSZTypes): SimpleContainerType => ({
```
twoeths, line 37: We are missing ssz spec test for this type, we can add that spec test in this PR or after we merge to 0.9.0. Just a note.

packages/lodestar/src/network/gossip/gossip.ts
```diff
@@ -0,0 +1,199 @@
...
+    //backward compatible
+    await promisify(this.pubsub.publish.bind(this.pubsub))(
+      getGossipTopic(GossipEvent.ATTESTATION), serialize(attestation, this.config.types.Attestation)
+    );
+    this.logger.verbose(
+      `Publishing attestation ${toHex(hashTreeRoot(attestation, this.config.types.Attestation))} for subnet ${subnet}`
+    );
+  }
+
+  public async publishAggregatedAttestation(aggregateAndProof: AggregateAndProof): Promise<void> {
```
twoeths: What I read from the latest validator spec https://github.com/ethereum/eth2.0-specs/blob/v09x/specs/validator/0_beacon-chain-validator.md#aggregateandproof, validator gonna broadcast aggregated attestation. If that's the case then we may need to make this network gossip as a separate package?

Please correct me if I'm wrong.

packages/eth2.0-types/src/ssz/generators/validator.ts
mpetrunic: I'll add it

packages/lodestar/src/network/gossip/gossip.ts
mpetrunic: you got it wrong, validator publishes attestation using rest api, which means beacon node will publish it on the right topic. Validator is not connected to libp2p

mpetrunic:
@ChainSafe/eth2-0 This is ready now :)

packages/lodestar/src/network/gossip/handlers/aggregateAndProof.ts
```diff
@@ -0,0 +1,23 @@
+/**
```
wemeetagain, line 1: It feels strange the only the receive-side is broken into files like this? Rather than both the receive and send.
It might be nice to see:
```typescript
export function sendAggregateAndProof(...) {...}
export function receiveAggregateAndProof(...) {...}
```

packages/lodestar/src/network/gossip/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+    encoding,
+    new Map([["subnet", getAttestationSubnet(attestation)]])
+  );
+}
+
+export function getAttestationSubnet(attestation: Attestation): string {
+  return String(attestation.data.index % ATTESTATION_SUBNET_COUNT);
+}
+
+export function handleGossipMessage<T>(msg: IGossipMessage, type: AnySSZType): T {
```
wemeetagain: can we call this `deserializeGossipMessage`

wemeetagain (changes requested):
Looking good, good to reorganize this stuff

packages/lodestar/src/network/gossip/handlers/aggregateAndProof.ts
mpetrunic: Sure, kept sending as is since they are one liners but yeah I can move

wemeetagain (approved):

PR #546, opened 2019-11-11
update eth1 deposit to 0.9.1

mpetrunic:
- updated contract abi and bytecode
- updated methods and deposit params


twoeths:
Looks good to me. There is a failed spec test in my state transition PR regarding deposit. I guess it'll be passed if we put this PR in.

twoeths (approved):

PR #554, opened 2019-11-14
Fetch block state that matches fork choice rules

### LINKED ISSUE #520: Fetch block "prestate" when producing
When producing block, we should fetch state that matches forckChoice block.stateRoot instead of latest state which sometimes breaks block production if our validators are lagging behind beacon node

aunyks:
closes #520.


packages/lodestar/src/chain/factory/block/index.ts
```diff
@@ -22,12 +22,12 @@ export async function assembleBlock(
   eth1: IEth1Notifier,
   slot: Slot,
   randao: bytes96
-): Promise<BeaconBlock|null> {
+): Promise<BeaconBlock | null> {
   const [parentBlock, currentState] = await Promise.all([
     db.block.get(chain.forkChoice.head()),
-    db.state.getLatest(),
+    db.state.get(chain.forkChoice.head()),
```
twoeths: I think what @mpetrunic means is:
+ Get `parentBlock: BeaconBlock` like above
+ Get `stateRoot` from `parentBlock`
+ Get state like what you did: `db.state.get(stateRoot)`

So instead of doing Promise.all we get block first and then state.

mpetrunic:
@aunyks You need to fix stubs to reflect your change in related unit and e2e tests

aunyks:
@mpetrunic For some reason getting a state like this returns an `undefined` value and I'm not quite sure why. Because of that it's difficult to test at the moment.

mpetrunic:
> @mpetrunic For some reason getting a state like this returns an `undefined` value and I'm not quite sure why. Because of that it's difficult to test at the moment.

It suppose to fail, because you changed logic. You are suppose to modify unit tests to reflect your changes. Slot is missing because the stub is missing for new method you are calling

aunyks:
@mpetrunic I don't quite understand. Are you saying that on [line 27](https://github.com/ChainSafe/lodestar/blob/eee1b82e2eba0c7054aa5142469dccbb166104b6/packages/lodestar/src/chain/factory/block/index.ts#L27), this `db.state.get()` is *supposed* to return a Promise that resolves to an undefined value?  

I ask because it's not just `slot` that's undefined. That function returns an undefined value.

mpetrunic:
> @mpetrunic I don't quite understand. Are you saying that on [line 27](https://github.com/ChainSafe/lodestar/blob/eee1b82e2eba0c7054aa5142469dccbb166104b6/packages/lodestar/src/chain/factory/block/index.ts#L27), this `db.state.get()` is _supposed_ to return a Promise that resolves to an undefined value?
> 
> I ask because it's not `slot` that's just undefined. That function returns an undefined value.

You code is perfectly fine, but existing test code is wrong.
Look [here](https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/test/unit/chain/factory/block/blockAssembly.test.ts#L47), test thats failing, it has stubbed `beaconDB.state.getLatest` while `beaconDB.state.get` should be stubbed. Similar thing is probably in other tests that are failing

aunyks:
Ah, I think I understand. Thanks for the clarification @mpetrunic!

packages/lodestar-validator/coverage/lcov-report/base.css
```diff
@@ -0,0 +1,224 @@
+body, html {
+  margin:0; padding: 0;
+  height: 100%;
+}
+body {
```
mpetrunic: coverage directory should be in gitignore and shouldn't go on git

mpetrunic (changes requested):
Looks good, just remove coverage from git

aunyks:
@mpetrunic Just made the change. Should be good to go once Travis finishes.

mpetrunic (dismissed):

wemeetagain:
@aunyks one file conflict, then this PR should be good to go

wemeetagain (approved):

PR #559, opened 2019-11-18
Fork choice update for 0.9.1

### LINKED ISSUE #556: Update fork choice to 0.9.1 spec
- [ ] bounce attack resistance - ethereum/eth2.0-specs#1465
- [x] decoy flip-flop resistance - ethereum/eth2.0-specs#1466

twoeths:
## Goal
+ Close #556 
+ Only this spec PR is applicable: https://github.com/ethereum/eth2.0-specs/pull/1466

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -242,7 +242,8 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
   }
 
   private processAttestation = async (latestState: BeaconState, attestation: Attestation, attestationHash: Hash) => {
-    const currentEpoch = computeEpochAtSlot(this.config, latestState.slot);
+    const currentSlot = getCurrentSlot(this.config, latestState.genesisTime);
+    const currentEpoch = computeEpochAtSlot(this.config, currentSlot);
     const previousEpoch = currentEpoch > GENESIS_EPOCH ? currentEpoch - 1 : GENESIS_EPOCH;
     const epoch = attestation.data.target.epoch;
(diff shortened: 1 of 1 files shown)


packages/lodestar/src/chain/chain.ts
```diff
@@ -242,6 +242,11 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
   }
 
   private processAttestation = async (latestState: BeaconState, attestation: Attestation, attestationHash: Hash) => {
+    const currentEpoch = computeEpochAtSlot(this.config, latestState.slot);
```
wemeetagain: I think we need to calculate the epoch in a special way here, using `Date.now() - latestState.genesisTime / SLOTS_PER_EPOCH`, since our latestState.slot may be far behind the 'wall-clock' slot.

wemeetagain (changes requested):

wemeetagain (approved):

packages/lodestar/src/chain/chain.ts
```diff
@@ -242,6 +242,12 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
   }
 
   private processAttestation = async (latestState: BeaconState, attestation: Attestation, attestationHash: Hash) => {
+    const currentSlot = getCurrentSlot(this.config, latestState.genesisTime);
```
mpetrunic: latestState.slot?

this will always be slot 0

mpetrunic: @wemeetagain should our fork choice track it's own time? Not sure how often is onTick called in spec

mpetrunic (approved):

PR #561, opened 2019-11-20
Remove --incremental from build:types

wemeetagain:
Using the --incremental flag, typescript will add a buildinfo file in the published package.
We don't want to be publishing this file, since its useless and potentially a security concern (leaks personal information of the publisher).

### DIFF
--- a/packages/benchmark-utils/package.json
+++ b/packages/benchmark-utils/package.json
@@ -18,7 +18,7 @@
     "clean": "rm -rf lib && rm -f tsconfig.tsbuildinfo",
     "build": "yarn build:lib && yarn build:types",
     "build:release": "yarn clean && yarn build && yarn build:docs",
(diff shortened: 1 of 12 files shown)


mpetrunic (changes requested):
This reduces build time from 60 to 30s on my pc, it will reduce event more time in future.

I think "files" in package.json should have more strict definition than just whole lib directory
Maybe something like:
```
"files": [
"lib/**/*.js",
"lib/**/*.d.ts"
]
```

GregTheGreek (approved):

wemeetagain:
> This reduces build time from 60 to 30s on my pc, it will reduce event more time in future.
> 
> I think "files" in package.json should have more strict definition than just whole lib directory
> Maybe something like:
> 
> ```
> "files": [
> "lib/**/*.js",
> "lib/**/*.d.ts"
> ]
> ```

fair enough, closing this PR in favor of something like that

PR #567, opened 2019-11-23
Add optional generic typescript type to ssz types

wemeetagain:
- Adds an optional generic type to ssz types (eg: `FullSSZType<T>`).
Defaults to `any`, so filling in the generic is not required and nothing changes if someone's ssz types don't fill in the generic.

- `ssz.deserialize` now uses this type as its return type.
eg:
```typescript
interface Foo {
  bar: number;
}
const fooType: AnySSZType<Foo> = {
  fields: [["bar", "uint8"]],
}
const deserialized = deserialize(data, fooType);
// ^ typescript now knows that `deserialized` is a `Foo`
deserialized.bar // <- IDE will be able to tab-complete this
```

- eth2.0-types `IBeaconSSZTypes` have been updated to include the typescript types of each ssz type
eg:
```typescript
interface IBeaconSSZTypes {
  ...
  BeaconState: AnySSZType<BeaconState>;
  ...
}
```

- In lodestar, we no longer do `deserialize(...) as Type` since we're always deserializing one of the known `IBeaconSSZTypes`

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "typedoc-plugin-external-module-name": "^2.1.0",
     "typedoc-plugin-internal-external": "^2.0.2",
     "typedoc-plugin-markdown": "^2.0.6",
-    "typescript": "^3.5.1",
+    "typescript": "^3.7.2",
     "webpack": "^4.39.1"
   },
   "optionalDependencies": {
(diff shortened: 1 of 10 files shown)


mpetrunic (dismissed):
me gusta :smile:

mpetrunic (approved):

PR #569, opened 2019-11-26
Add defaultValue and createValue to ssz

wemeetagain:
depends on #568

Adds `defaultValue` and `createValue` utilities to ssz to create new ssz objects from an `AnySSZType`.

`defaultValue<T>(type: AnySSZType<T>): T` returns an object in its 'zeroed' form.
`createValue<T>(type: AnySSZType<T>, value?: RecursivelyPartial<T>): T`returns an object 'zeroed' out except for any partially filled data passed in with `value`


packages/ssz/src/core/createValue.ts
```diff
@@ -0,0 +1,61 @@
...
+      }
+      return Array.from({length: type.length}, (_, i) =>
+        _createValue(type.elementType, value[i]));
+    case Type.container:
+      if (value) {
+        assert(Object(value) === value);
+      } else {
+        value = {};
+      }
+      // eslint-disable-next-line no-case-declarations
```
mpetrunic, line 40: I think you can put type unknown or object or {[k: string]: unknown}

packages/ssz/src/core/defaultValue.ts
```diff
@@ -0,0 +1,47 @@
...
+      return Buffer.alloc(0);
+    case Type.byteVector:
+      return Buffer.alloc(type.length);
+    case Type.list:
+      return [];
+    case Type.vector:
+      return Array.from({length: type.length}, () =>
+        defaultValue(type.elementType));
+    case Type.container:
+      // eslint-disable-next-line no-case-declarations
```
mpetrunic, line 40: same as above

mpetrunic (commented):
Seems useful :smiley:

packages/ssz/src/core/createValue.ts
wemeetagain: This disable is bc eslint doesn't like declaring variables inside a switch/case.

mpetrunic: Oh, than my solution fixes, ts-ignore here, and using any in below case :stuck_out_tongue:

packages/ssz/src/core/createValue.ts
```diff
@@ -0,0 +1,60 @@
...
+        assertValidValue(value, type);
+        return value;
+      } else {
+        return defaultValue(type);
+      }
+    case Type.list:
+      if (value) {
+        assert(Array.isArray(value));
+      } else {
+        value = [];
```
ec2: What cases would this be hit? Not seeing it.

packages/ssz/src/core/createValue.ts
```diff
@@ -0,0 +1,60 @@
...
+      }
+      return value.map((v: any) =>
+        _createValue(type.elementType, v));
+    case Type.vector:
+      if (value) {
+        assert(Array.isArray(value));
+      } else {
+        value = [];
```
ec2: What cases would this be hit? Not seeing it.

packages/ssz/src/core/createValue.ts
wemeetagain: if `value === null`

ec2: Oh yeah, youre right. I misread :\

ec2: Should you use the default value then?

ec2: The function that you created

packages/ssz/src/core/createValue.ts
```diff
@@ -0,0 +1,60 @@
...
+import {assertValidValue} from "./assertValidValue";
+import {defaultValue} from "./defaultValue";
+
+/**
+ * Given an ssz type, create an empty value, optionally filled in with a default value
+ *
+ * The default value can be only partially specified.
+ */
+export function createValue<T>(type: AnySSZType<T>, value: any = null): T {
+  return _createValue(parseType(type), value) as T;
```
ec2, line 14: Correct me if im wrong, but it seems like we only fill in with default value when `value === null`, so we can remove a lot of code in `_createValue` if we check if null here and just return `defaultValue` here and remove the checks in `_createValue`

ec2:
LGTM, other than a little optimization as suggested above.

ec2 (dismissed):

packages/ssz/src/core/createValue.ts
wemeetagain: Oh yeah that would be better

packages/ssz/src/core/createValue.ts
wemeetagain: :+1:

wemeetagain: Unfortunately, we still need the null checks in the inner function. The intended behavior is that a user can enter a partially filled object, and the unspecified parts are filled in with default values. Since the inner function is called recursively, those calls to unspecified parts will still need the null check.

ec2 (dismissed):

mpetrunic (dismissed):

packages/ssz/src/core/clone.ts
```diff
@@ -62,10 +61,14 @@ function _clone(value: any, type: FullSSZType): any {
...
-        return value.clone();
+      switch (type.use) {
+        case UintImpl.bn:
+          return value.clone();
+        case UintImpl.bigint:
+        case UintImpl.number:
+          return value;
       }
-      return value;
+      break;
```
mpetrunic: I think defaulting to bigint or trowing exception would be better solution

packages/ssz/src/core/createValue.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+function _createValue(type: FullSSZType, value: any = null): any {
```
mpetrunic, line 17: why aren't you using generics here as well?

packages/ssz/test/util/testCases.ts
```diff
@@ -1 +1,57 @@
-export const TEST_CASE_LOCATION = "../../../../../../node_modules/@chainsafe/eth2-spec-tests";
\ No newline at end of file
+import {describeDirectorySpecTest, InputType, safeType} from "@chainsafe/eth2.0-spec-test-util";
```
mpetrunic, line 1: Isn't this file duplicate of /test/spec/util?

mpetrunic (changes requested):

packages/ssz/src/core/clone.ts
wemeetagain: ok, its meant to be unreachable code, will throw an error

packages/ssz/src/core/createValue.ts
wemeetagain: not really needed at this lower level, only the higher level functions are used regularly.
I remember attempting this at some point and running into trouble, and thinking it's not worth it.

packages/ssz/test/util/testCases.ts
wemeetagain: I don't think so, this is a templated function for running the "static" ssz spec tests

mpetrunic: I'm pretty sure they are identical:
packages/ssz/test/spec/util.ts

wemeetagain: doh, yeah, will remove duplicate

mpetrunic (approved):

PR #570, opened 2019-11-28
Switch apache milagro with herumi/bls-wasm

### LINKED ISSUE #566: BLS herumi wasm
Replace current apache milagro implementation with herumi bls wasm build

mpetrunic:
- api is still intact (thank you wrappers haha)
- there are some additional methods like (PrivateKey.from(int)) - should replace that custom pk generation
- still needs arguments checking and exception catching
- It should be possible to turn off order checking (big performance improvement) but I can't figure out how to call that wasm method) 

resolves #566


packages/bls/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@chainsafe/bls",
-  "version": "0.1.7",
```
ec2, line 3: didnt we say were gunna update version numbers in a different PR or something?

mpetrunic: Might be, tbh, I've updated version just to try it out in lodestar, I can remove

ec2 (dismissed):

ec2:
Other than the thing about versioning, lgtm

packages/bls/package.json
mpetrunic: That said, if we have to release something before we can use it in monorepo, whats the differences from having multiple repos? :smile: cc @wemeetagain

packages/bls/src/index.ts
```diff
@@ -1,15 +1,18 @@
...
-import {ElipticCurvePairing} from "./helpers/ec-pairing";
-import ctx from "./ctx";
 import {BLSPubkey, BLSSecretKey, BLSSignature, Domain, Hash} from "@chainsafe/eth2.0-types";
+import {init} from "./context";
+import {PUBLIC_KEY_LENGTH} from "./constants";
 
 export {Keypair, PrivateKey, PublicKey, Signature};
 
+
+export async function initLibrary(): Promise<void> {
```
wemeetagain: we could just `export {init};`

wemeetagain (changes requested):
Looks great! Much simpler.

Separate thought, if we simplify our code (eg: validator and keystore code), we probably won't even need to use the OO interface to bls, maybe just use a simple IKeypair interface :)
```typescript
interface IKeypair {
  private: Buffer;
  public: Buffer;
}
```
and the functional interface exported here in this file
```typescript
function getPublicKey(priv: Buffer): Buffer;
function sign(priv: Buffer, msg: Buffer, domain: Buffer): Buffer;
function aggregateSignatures(sigs: Buffer[]): Buffer;
function aggregatePublicKeys(pubs: Buffer[]): Buffer;
function verify(pub: Buffer, msg: Buffer, sig: Buffer, domain: Buffer): boolean;
function verifyMultiple(pubs: Buffer[], msgs: Buffer[], sig: Buffer, domain: Buffer): boolean;
```

packages/bls/src/index.ts
mpetrunic: I need to rename one of them or IDE is going wild with typehinting both of them haha

mpetrunic:
> Looks great! Much simpler.
> 
> Separate thought, if we simplify our code (eg: validator and keystore code), we probably won't even need to use the OO interface to bls, maybe just use a simple IKeypair interface :)
> 
> ```ts
> interface IKeypair {
>   private: Buffer;
>   public: Buffer;
> }
> ```
> 
> and the functional interface exported here in this file
> 
> ```ts
> function getPublicKey(priv: Buffer): Buffer;
> function sign(priv: Buffer, msg: Buffer, domain: Buffer): Buffer;
> function aggregateSignatures(sigs: Buffer[]): Buffer;
> function aggregatePublicKeys(pubs: Buffer[]): Buffer;
> function verify(pub: Buffer, msg: Buffer, sig: Buffer, domain: Buffer): boolean;
> function verifyMultiple(pubs: Buffer[], msgs: Buffer[], sig: Buffer, domain: Buffer): boolean;
> ```

Actually, I was thinking of removing those functional interfaces as they are somewhat slower and much less optimized (you have to deserialize buffer and allocate new array to create internal wasm implementation format). Example, keeping PrivateKey instance and using `sign` method on that instance removes unnecessary deserialization and allocation.

We could see potential speed up if we convert, our validator registry public keys into PublicKey instances once and just use instances. It would shave ~20-30% time on every aggregation and verification

packages/bls/src/index.ts
mpetrunic: What do you think of renaming to `initBLS`. Might be better name, because whomever is using this library would have to rename that export every time so he is aware what's going on. Random `await init()` in your code is confusing

wemeetagain: > What do you think of renaming to initBLS

sgtm :)

wemeetagain:
I see, good point. Public keys would be a good thing to store with an "optimized beacon state", along with the committees and other things.
We definitely shouldn't remove the functional interface though, as its convenient in the naive case. Buffer in, buffer out.

wemeetagain (approved):

PR #573, opened 2019-11-30
Fork choice update for 0.9.1 - Spec PR 1465

### LINKED ISSUE #556: Update fork choice to 0.9.1 spec

twoeths:
## Goal
+ Close #556 
+ Address PR 1465 of the spec - ethereum/eth2.0-specs#1465

### DIFF
--- a/packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
+++ b/packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
@@ -5,7 +5,7 @@
 import assert from "assert";
 import BN from "bn.js";
 
-import {Gwei, Hash, Slot, ValidatorIndex, number64,} from "@chainsafe/eth2.0-types";
+import {Gwei, Hash, Slot, ValidatorIndex, number64, Checkpoint, Epoch,} from "@chainsafe/eth2.0-types";
 
 import {ILMDGHOST} from "../interface";
 
@@ -166,12 +166,12 @@ export class StatefulDagLMDGHOST implements ILMDGHOST {
   /**
    * Last finalized block
    */
-  private finalized: Node|null;
+  private finalized: {node: Node; epoch: Epoch} | null;
 
   /**
    * Last justified block
    */
-  private justified: Node|null;
+  private justified: {node: Node; epoch: Epoch} | null;
   private synced: boolean;
 
   public constructor(config: IBeaconConfig) {
@@ -188,7 +188,8 @@ export class StatefulDagLMDGHOST implements ILMDGHOST {
     this.genesisTime = genesisTime;
   }
 
-  public addBlock(slot: Slot, blockRootBuf: Hash, parentRootBuf: Hash): void {
+  public addBlock(slot: Slot, blockRootBuf: Hash, parentRootBuf: Hash,
+    justifiedCheckpoint?: Checkpoint, finalizedCheckpoint?: Checkpoint): void {
     this.synced = false;
     const blockRoot = blockRootBuf.toString("hex");
     const parentRoot = parentRootBuf.toString("hex");
@@ -206,6 +207,12 @@ export class StatefulDagLMDGHOST implements ILMDGHOST {
     if (this.nodes[parentRoot]) {
       this.nodes[parentRoot].addChild(node);
     }
+    if (justifiedCheckpoint && (!this.justified || justifiedCheckpoint.epoch > this.justified.epoch)) {
+      this.setJustified(justifiedCheckpoint);
+    }
+    if (finalizedCheckpoint && (!this.finalized || finalizedCheckpoint.epoch > this.finalized.epoch)) {
+      this.setFinalized(finalizedCheckpoint);
+    }
   }
 
   public addAttestation(blockRootBuf: Hash, attester: ValidatorIndex, weight: Gwei): void {
@@ -215,21 +222,6 @@ export class StatefulDagLMDGHOST implements ILMDGHOST {
       attester,
       weight,
     });
-  }
-
-  public setFinalized(blockRoot: Hash): void {
-    this.synced = false;
-    const rootHex = blockRoot.toString("hex");
-    this.finalized = this.nodes[rootHex];
-    this.prune();
-    this.aggregator.prune();
-  }
-
-  public setJustified(blockRoot: Hash): void {
-    if (this.shouldUpdateJustifiedCheckpoint(blockRoot)) {
-      const rootHex = blockRoot.toString("hex");
-      this.justified = this.nodes[rootHex];
-    }
   }
 
   public syncChanges(): void {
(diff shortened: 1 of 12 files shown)


wemeetagain (approved):
LGTM
Great job :rocket:

packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
```diff
@@ -216,8 +226,10 @@ export class StatefulDagLMDGHOST implements ILMDGHOST {
   }
 
   public setJustified(blockRoot: Hash): void {
-    const rootHex = blockRoot.toString("hex");
-    this.justified = this.nodes[rootHex];
+    if (this.shouldUpdateJustifiedCheckpoint(blockRoot)) {
```
mpetrunic: Aren't we already doing this check in chain before calling this method?

packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
```diff
@@ -242,6 +254,45 @@ export class StatefulDagLMDGHOST implements ILMDGHOST {
...
+      return false;
+    }
+    if (this.getAncestor(blockRoot.toString("hex"), this.justified.slot) !== this.justified.blockRoot) {
+      return false;
+    }
+
+    return true;
+  }
+
+  private computeSlotsSinceEpochStart(slot: Slot): number {
```
mpetrunic: I think this method should go in state transition package if it's not already there. You could also have option param epcoh

packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
twoeths: No, this is new method and specific for fork choice.

packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
twoeths: yeah it's done

packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
mpetrunic: This could leave us in wierd state where we would update database and everything while we weren't actually justified.

As it seems we have two options:
1. change this method to return "true" if it updated fork choice - in chain module call this method first and update database and other stuff
2. Leave fork choice as is and call `shouldUpdateJustifiedCheckpoint` in chain module

I'm leaning on option 1 as I think chain module should follow fork choice decisions
 
cc @wemeetagain

twoeths: yeah I see the concern. From the spec of https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_fork-choice.md, I only see we use this new method in the "`store`".

mpetrunic: Actually on second look I might be wrong. Chain updates database depending on if state transition justified epoch (based on attestations). Still not sure what happens if state marks epoch as justified but fork choice doesn't set it.Might be irrelevant. It would be awesome if cayman can explain this better just to make it clear

wemeetagain: hmm. ok, so something is a little funky here, didn't catch it the first time around.
`forkChoice.setJustified` is currently only being called on an epoch boundary, once the chain sees a new justified epoch (src/chain/chain.ts#L356). This 'bounce' attack is meant to not update justified blocks in later slots within an epoch, but this is was never a case that was considered in our implementation, because we only update on epoch boundaries anyways.
I think we may want to take a step back at what we're currently doing, and maybe revamp the whole process.
It doesn't even look like we're applying the fork choice at all (src/chain/chain.ts#L338)

twoeths: The issue to me regarding understanding the spec is that there are spec for fork choice and spec for beacon chain but they do not link together, so it's not clear to me on when to set justified check point to `forkChoice/store` and when to call `get_head`. 

The other thing to consider is we have a different implementation compared to the spec. Should we update justified check point to fo forkChoice like this in the spec
```
if state.current_justified_checkpoint.epoch > store.justified_checkpoint.epoch:
        store.best_justified_checkpoint = state.current_justified_checkpoint
        if should_update_justified_checkpoint(store, state.current_justified_checkpoint):
            store.justified_checkpoint = state.current_justified_checkpoint
```
or we do that on epoch boundary like in the current implementation.

mpetrunic (approved):

wemeetagain (changes requested):
It looks good. The only difference I see with the spec is that they save the best justified root even if `shouldUpdateJustifiedCheckpoint` returns false. They then can set the justified root based on this, when a new epoch rolls around (based on clock time).
Imagine a scenario where the fork choice calculates that there is a better justified root, but it's already 3/4 thru the epoch, so it isn't immediately set. Then the next many blocks are left empty by validators that are offline/etc. The spec implementation will set the better justified root when the next epoch comes around, even if no blocks are around to trigger the setting of the justified root.
All in all, we should probably start a `setInterval` in the `start` method with the interval being triggered at the start of each new epoch, fulfilling the functionality of `on_tick` of the spec.

packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
```diff
@@ -161,24 +167,52 @@ export class StatefulDagLMDGHOST implements ILMDGHOST {
...
+   */
+  public async start(genesisTime: number): Promise<void> {
+    this.genesisTime = genesisTime;
+    const numSlot = computeSlotsSinceEpochStart(this.config, getCurrentSlot(this.config, this.genesisTime));
+    const timeToWaitTillNextEpoch = (this.config.params.SLOTS_PER_EPOCH - numSlot) * 
+      this.config.params.SECONDS_PER_SLOT * 1000;
+    // Make sure we call onTick at start of each epoch
+    await sleep(timeToWaitTillNextEpoch);
+    const epochInterval = this.config.params.SLOTS_PER_EPOCH * this.config.params.SECONDS_PER_SLOT * 1000;
+    setInterval(this.onTick.bind(this), epochInterval);
```
wemeetagain: We should have a `stop` method that does `clearInterval` that we call in `chain.stop`

wemeetagain (changes requested):
onTick behavior looks good, one minor thing before it looks ready to merge

wemeetagain (approved):
:tada:

PR #574, opened 2019-12-05
bump bls version in lodestar

mpetrunic:


wemeetagain (changes requested):
I think we need to call `initBLS` in lodestar somewhere

mpetrunic:
> I think we need to call `initBLS` in lodestar somewhere

Oh damm, it was called in node and validator start method but I seemed to have lost it when reverted in last PR :disappointed: 

good catch

wemeetagain (approved):

PR #575, opened 2019-12-06
Send gossip messages to OpPool

### LINKED ISSUE #551: Add operations from gossip to operations pool
- listen on VoluntaryExit, ProposerSlashing, AttesterSlashing and committee attestation from gossip and add to OpPool

twoeths:
## Goal
+ Close #551


packages/lodestar/src/sync/regular.ts
```diff
@@ -68,6 +84,51 @@ export class RegularSync {
...
+    }
+    const state = await this.db.state.getLatest();
+    const currentSlot = getCurrentSlot(this.config, state.genesisTime);
+    if (!(attestation.data.slot + ATTESTATION_PROPAGATION_SLOT_RANGE >= currentSlot &&
+      currentSlot >= attestation.data.slot)) {
+      return;
+    }
+    if(!isValidIndexedAttestation(this.config, state, getIndexedAttestation(this.config, state, attestation))) {
+      return;
+    }
```
mpetrunic: This is wrong place for validation since those validation needs to be hooked into gossipsub itself.
We shouldn't rebroadcast messages without validating them, and here is way too late to validate it.

I would suggest to just implement functions like:
- isValidAttestationGossipMessage 
- isValidAggregationAndProofGossipMessage 
etc

Might be even better to implement some class to contain those methods since you need database access for some

packages/lodestar/src/sync/regular.ts
```diff
@@ -68,6 +84,51 @@ export class RegularSync {
...
+    let count = 0;
+    for (let i = 0; i < aggregationBits.bitLength; i++) {
+      if (aggregationBits.getBit(i)) {
+        count++;
+      }
+    }
+    if (count !== 1) {
+      return;
+    }
+    if (await this.db.block.isBadBlock(attestation.data.beaconBlockRoot)) {
```
mpetrunic: Add check if we even have that block in database

packages/lodestar/src/sync/regular.ts
```diff
@@ -68,6 +84,51 @@ export class RegularSync {
...
+    const aggregationBits = attestation.aggregationBits;
+    let count = 0;
+    for (let i = 0; i < aggregationBits.bitLength; i++) {
+      if (aggregationBits.getBit(i)) {
+        count++;
+      }
+    }
+    if (count !== 1) {
+      return;
+    }
```
mpetrunic: extract this check to method

packages/lodestar/src/sync/regular.ts
```diff
@@ -68,6 +84,51 @@ export class RegularSync {
...
+    await Promise.all([
+      this.opPool.attestations.receive(attestation),
+      this.chain.receiveAttestation(attestation),
+    ]);
+  };
+
+  public receiveAggregateAndProof = async (aggregation: AggregateAndProof): Promise<void> => {
+    // a place holder for mainnet
+    // this is for aggregated attestattions
+    this.logger.debug(aggregation);
```
mpetrunic: This is not for mainnet, this is mandatory for 0.9.2, add Bucket in database to store those

packages/lodestar/src/sync/regular.ts
```diff
@@ -68,6 +84,51 @@ export class RegularSync {
...
+  };
+
   public receiveAttestation = async (attestation: Attestation): Promise<void> => {
```
mpetrunic, line 88: You need to ignore aggregated attestations here

mpetrunic (changes requested):

packages/eth2.0-utils/src/bytes.ts
```diff
@@ -17,6 +17,15 @@ export function intToBytes(value: BN | number, length: number): bytes {
...
 }
 
+export function bytesToInt(value: bytes): number {
+  const length = value.length;
+  let result = 0;
+  for (let i = 0; i < length; i++) {
+    result += value[i] * 2 ** (8 * i);
+  }
+  return result;
+}
```
mpetrunic, line 30: is this le or be? might wanna put that in comment or even in method name

mpetrunic (commented):
You didn't had to add validation to existing gossip handlers as they will need to be moved after we add gossipsub message validation hook

packages/eth2.0-utils/src/bytes.ts
wemeetagain: should be LE, matching `intToBytes` above

wemeetagain (approved):

wemeetagain:
@mpetrunic RE: gossip validation I was thinking of refactoring gossip / moving to new libp2p, after 0.9.x is merged back in master.

packages/eth2.0-utils/src/bytes.ts
twoeths: I've just added comment to method.

mpetrunic (approved):

PR #576, opened 2019-12-06
Chore: dependency updates

mpetrunic:
- updates eslint to latest version
   - removes warning about unsupported typescript version
   - "ban-ts-ignore" rule put to warning - we are using it pretty lightly as there are bunch of places where it shouldn't be used
- removed babel-runtime dependency and plugin
   - it it automatically added to support old browsers without Promises and esmodules but dropped those with bigint anyways (most bundlers will polyfill this stuff anyways)


mpetrunic:
I need #574 before we can merge because already released bls depends on babel-runtime

.babelrc
```diff
@@ -2,14 +2,21 @@
   "presets": [
     [
       "@babel/env",
-      {"targets": {"node":"current"}}
+      {
+        "targets": {
+          "node": "8",
```
wemeetagain: I think we'll be using features only in current LTS and beyond (eg: bigint)

wemeetagain (dismissed):

.babelrc
mpetrunic: oh yeah, >=10.4.0 supports bigint

Although, I would still keep node 8 support for libraries without bigint. What do you think?

wemeetagain: :+1: Yeah, as much support as possible

.babelrc
```diff
@@ -2,14 +2,21 @@
...
+          "browsers": ">1%, not ie 11"
```
ec2, line 8: :( no love for IE

mpetrunic: haha, just noticed that bigint also means we dropped support for safari haha

ec2: rip. thats actually quite a big chunk of people

ec2 (dismissed):

wemeetagain (approved):

PR #579, opened 2019-12-09
Allow Blocks to be skipped by step on RequestBlocksByRange

### LINKED ISSUE #521: RequestBlocksByRange block skipping
Currently we ignore step flag when returning blocks. We should use given step number and skip blocks depending on step amount.

aunyks:
Closes #521


aunyks:
Ah, this means I should add a set of tests for these changes, right?

packages/lodestar/src/db/api/beacon/repository.ts
```diff
@@ -64,12 +64,18 @@ export abstract class BulkRepository<T> extends Repository<T> {
     return (data || []).map((data) => deserialize(data, this.type));
   }
 
-  public async getAllBetween(lowerLimit: number|null, upperLimit: number|null): Promise<T[]> {
+  public async getAllBetween(lowerLimit: number | null, upperLimit: number | null, step?: number | null): Promise<T[]> {
```
mpetrunic:
```suggestion
  public async getAllBetween(lowerLimit: number | null, upperLimit: number | null, step: number | null = null): Promise<T[]> {
```
so you don't have to check for undefineed

packages/lodestar/src/db/api/beacon/repository.ts
```diff
@@ -64,12 +64,18 @@ export abstract class BulkRepository<T> extends Repository<T> {
...
       gt: encodeKey(this.bucket, lowerLimit || Buffer.alloc(0)),
       lt: encodeKey(this.bucket, upperLimit || Number.MAX_SAFE_INTEGER),
     });
-    return (data || []).map((data) => deserialize(data, this.type));
+    let processedData = data;
+    if (typeof step !== "undefined" && step !== null) {
+      processedData = data.filter((datum, index) => {
+        return index % step === 0;
+      });
+    }
```
mpetrunic: This is kinda awkward here and we are reading a lot of data that we dont need. Also this allows us to do same thing in other Buckets.

I would suggest to add this code here: https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/db/controller/impl/level.ts#L89

You can do this if you add optional param to ISearchOptions.

mpetrunic (changes requested):

packages/lodestar/src/db/controller/impl/level.ts
```diff
@@ -86,7 +86,7 @@ export class LevelDbController extends EventEmitter implements IDatabaseControll
         gt: opts.gt,
         lt: opts.lt,
       }).on("data", function (data) {
-        searchData.push(data);
+        searchData.push(data.filter((datum, index) => index % (opts.step || 1) === 0));
```
mpetrunic: ehm, pretty sure data is one item from database like, that function is called multiple times.
You should count how many times this method is called I guess

mpetrunic (changes requested):

packages/lodestar/src/db/controller/impl/level.ts
wemeetagain: Yeah, data is a single item. You'll want to wrap the `searchData.push(data)` in an `if` that does the proper check. You may need to use a [`createReadStream`](https://github.com/Level/levelup#dbcreatereadstreamoptions) so you have access to the keys that you want to filter by.

aunyks: Are the keys array indices or block heights, though? Like can I assume that some key is equal to the last key plus one?

wemeetagain: I don't think thats a safe assumption. There may be skip slots, where we only store eg: block 2, 3, 5, ... (skipping slot 4)

aunyks: Gotcha. And just to be sure, if a client calls this rpc method with a start slot 2, count 3, and step 2, do we return [2, 5] because 5 is two indices away from start or [2] because 4 is two slots away from start but we don't have it?

mpetrunic: I would say 2,3,5 as you he requested 3 blocks.

That said if 3 is "skipped slot" you would return just 2, 5

aunyks: Sounds good. Thanks for the clarification y'all.

aunyks:
Is Travis okay lol

wemeetagain:
:cry:  its not ok, working on it now

aunyks:
Can someone re-run Travis on this PR? I don't think I have permission.

wemeetagain:
closed and reopened to trigger travis

packages/lodestar/src/db/controller/impl/level.ts
```diff
@@ -82,11 +82,19 @@ export class LevelDbController extends EventEmitter implements IDatabaseControll
...
-      this.db.createValueStream({
+      this.db.createReadStream({
         gt: opts.gt,
         lt: opts.lt,
-      }).on("data", function (data) {
-        searchData.push(data);
+      }).on("data", function ({key, value}) {
+        // Typechecking keys as they
+        // can be defined as any type
+        if (typeof key === "number") {
```
wemeetagain: I think there's a tricky issue with this here.
Where this is called, in `getAllBetween`, you can see that in practice, we "encode" the keys, so here, the keys will not be numbers, and not the keys that are used within the application. They're prepended with the "bucket", an identifier of the category of the data, and are buffers.

wemeetagain (changes requested):

packages/lodestar/src/db/controller/impl/level.ts
twoeths: we can track by a new `count` variable internally so we don't have to base on type of key I think

aunyks:
This most recent change moves the step logic further up the call stack so we don’t make any assumptions about keys and stuff down at the DB Controller.

packages/lodestar/src/db/api/beacon/repository.ts
```diff
@@ -65,17 +65,22 @@ export abstract class BulkRepository<T> extends Repository<T> {
...
+      gt: encodeKey(this.bucket, safeLowerLimit),
+      lt: encodeKey(this.bucket, safeUpperLimit),
     });
-    let processedData = data;
-    if (step !== null) {
-      processedData = data.filter((datum, index) => {
-        return index % step === 0;
+    const processedData = (data || [])
+      .map((processedData) => deserialize(processedData, this.type))
+      .filter(block => {
```
mpetrunic: This method is not block specific, and this variabkle doesn't have slot field.

mpetrunic (changes requested):
Your first solution was on the right track, with tiny modification:
```javascript
public search(opts: ISearchOptions): Promise<any> {
    return new Promise<any[]>((resolve) => {
      let index = 0;
      const searchData: any[] = [];
      this.db.createValueStream({
        gt: opts.gt,
        lt: opts.lt,
      }).on("data", function (data) {
        if(index % opts.step === 0) {
           searchData.push(data);
        }
        index++;
      }).on("close", function () {
        resolve(searchData);
      }).on("end", function () {
        resolve(searchData);
      });
    });
  }
```

You could also add limit ini similar fashion (it should resolve and close stream if index > limit )

aunyks:
> Your first solution was on the right track, with tiny modification:
> 
> ```js
> public search(opts: ISearchOptions): Promise<any> {
>     return new Promise<any[]>((resolve) => {
>       let index = 0;
>       const searchData: any[] = [];
>       this.db.createValueStream({
>         gt: opts.gt,
>         lt: opts.lt,
>       }).on("data", function (data) {
>         if(index % opts.step === 0) {
>            searchData.push(data);
>         }
>         index++;
>       }).on("close", function () {
>         resolve(searchData);
>       }).on("end", function () {
>         resolve(searchData);
>       });
>     });
>   }
> ```
> 
> You could also add limit ini similar fashion (it should resolve and close stream if index > limit )

I moved on from that change, because it could be the case that we don't have every block between our upper and lower limits. For example, if the lower and upper limits are 1 and 5 and step is 2, we might not have block 3. That'd cause us to return 1 and 4, as that missing block will throw off the counting.  

Then I wanted to filter based on the keys in the database but I think @wemeetagain said that the keys aren't very easy to parse for filtering.

mpetrunic:
> I moved on from that change, because it could be the case that we don't have every block between our upper and lower limits. For example, if the lower and upper limits are 1 and 5 and step is 2, we might not have block 3. That'd cause us to return 1 and 4, as that missing block will throw off the counting.

I was under impression that we are suppose to return 1 and 4 in that case. Otherwise when syncing it could happen that we don't return any block if every second is skipped and thus defeat the purpose of interleaving blocks while syncing. I could be wrong. cc @wemeetagain

mpetrunic:
I'm probably wrong, this would cause some weird issues

wemeetagain:
copy pasta of relevant spec section for clarity:
Requests count beacon blocks from the peer starting from start_slot on the chain defined by head_block_root (= hash_tree_root(SignedBeaconBlock.message)). The response MUST contain no more than count blocks. step defines the slot increment between blocks. For example, requesting blocks starting at start_slot 2 with a step value of 2 would return the blocks at [2, 4, 6, …]. In cases where a slot is empty for a given slot number, no block is returned. For example, if slot 4 were empty in the previous example, the returned array would contain [2, 6, …]. A step value of 1 returns all blocks on the range [start_slot, start_slot + count).

wemeetagain:
@mpetrunic 
> Otherwise when syncing it could happen that we don't return any block if every second is skipped

I think this is the intended behavior, if step = 2 and every other slot is skipped, no blocks would be returned.

packages/lodestar/src/db/api/beacon/repository.ts
wemeetagain: I think we should override the `getAllBetween` method in src/db/api/beacon/repositories/blockArchive.ts w an implementation like this, where we know we're only handling blocks.

wemeetagain (dismissed):

packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
```diff
@@ -29,4 +29,27 @@ export class BlockArchiveRepository extends BulkRepository<BeaconBlock> {
...
+    step: number | null = null
+  ): Promise<BeaconBlock[]> {
+    const safeLowerLimit = lowerLimit || Buffer.alloc(0);
+    const safeUpperLimit = upperLimit || Number.MAX_SAFE_INTEGER;
+    const data = await this.db.search({
+      gt: encodeKey(this.bucket, safeLowerLimit),
+      lt: encodeKey(this.bucket, safeUpperLimit),
+    });
+    const processedData = (data || [])
+      .map((processedData) => deserialize(processedData, this.type))
```
mpetrunic: For the sake on typehinting, it is probably better to change it to deserialize<BeaconBlock>(...)

mpetrunic (dismissed):

aunyks:
Just noticed I flipped logic in that filter function. Pushing a fix right now.

packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
```diff
@@ -29,4 +29,27 @@ export class BlockArchiveRepository extends BulkRepository<BeaconBlock> {
...
+    const safeLowerLimit = lowerLimit || Buffer.alloc(0);
+    const safeUpperLimit = upperLimit || Number.MAX_SAFE_INTEGER;
+    const data = await this.db.search({
+      gt: encodeKey(this.bucket, safeLowerLimit),
+      lt: encodeKey(this.bucket, safeUpperLimit),
+    });
+    const processedData = (data || [])
+      .map((datum) => deserialize(this.type, datum))
+      .filter(block => {
+        if (step !== null && typeof safeLowerLimit === "number") {
```
mpetrunic, line 47: It would probably make a lot more sense to but step = 1 as default value instead of null :)

aunyks: Good point! @mpetrunic

mpetrunic (approved):

PR #583, opened 2019-12-11
Add ci test github action

wemeetagain:
Probably needs more work, but here's a start.
Meant to replace the current travis functionality.


.github/workflows/test.yml
```diff
@@ -0,0 +1,39 @@
+name: CI Tests
+
+on: [pull_request]
```
mpetrunic: push as well?

.github/workflows/test.yml
```diff
@@ -0,0 +1,39 @@
...
+      with:
+          cmd: run coverage
+    - name: E2e tests
+      uses: wemeetagain/actions-lerna@v0.1.0
+      with:
+          cmd: run test:e2e
+    - name: Spec tests
+      uses: wemeetagain/actions-lerna@v0.1.0
+      with:
+          cmd: run test:spec-min
```
mpetrunic: it should run full suite on master

mpetrunic (commented):
any idea how to integrate docs?

mpetrunic (commented):
Those parallel tests are a mess, I think running test one by one is probably cleaner. Our tests are still too short to make up for booting of virtual machine

Especially since you are running bootstrap in every single one of them haha

wemeetagain:
might make sense to run spec tests separately and everything else as one job (lint, check-types, unit, e2e)

mpetrunic:
> might make sense to run spec tests separately and everything else as one job (lint, check-types, unit, e2e)

maybe, but minimal spec tests are quite fast now though

.github/workflows/docs.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Spec tests
+      uses: wemeetagain/actions-lerna@v0.1.0
+      with:
+          cmd: run run build:release
+    - name: Build docs
+      run: node scripts/collect-docs.js
+    - name: Deploy
+      uses: peaceiris/actions-gh-pages@v2
+      env:
+        PERSONAL_TOKEN: ${{ secrets.GH_PAGES_TOKEN }}
```
priom, line 26: How's this getting triggered? I don't see any variable named GH_PAGES_TOKEN in the settings.

wemeetagain: you're right, I guess that was set on travis's secrets.

priom: yeah so, for github actions, you have to set the secrets in the individual repo

mpetrunic: can't you put secrets in repo -> settings -> secrets?

Isn't repo just for large secrets?

wemeetagain: added token to repo secrets

wemeetagain:
> https://github.com/ethereum/eth2.0-APIs/pull/14/files#diff-1d37e48f9ceff6d8030570cd36286a61R12 stuck_out_tongue
I couldn't get it to work, I think we should just merge so we can start using actions, we can incrementally improve what we have here

mpetrunic (approved):

twoeths (approved):

PR #584, opened 2019-12-13
Handle block forks

### LINKED ISSUE #578: Properly handle block forks
Right now, we maintain the `latestState`, which is the state after processing the latest received block. We're always using this to progress the chain.
This means we cannot process legitimate blocks that are not descendants of the latest block we've seen.

eg:
```
block A -----> block B
         \
          \
           \
            \
             > block C
```
If we receive block C after processing block B, we currently cannot process it, since we're not using the state at block A, only using the new `latestState` (post-block B).

We should instead fetch the (non-latest) state needed to process a block if we need to.

twoeths:
## Goal
+ Close #578

### DIFF
--- a/packages/lodestar/src/api/impl/validator/produceAttestation.ts
+++ b/packages/lodestar/src/api/impl/validator/produceAttestation.ts
@@ -3,7 +3,7 @@ import {Attestation, BLSPubkey, CommitteeIndex, Slot} from "@chainsafe/eth2.0-ty
 import {IBeaconDb} from "../../../db/api";
 import {IBeaconChain} from "../../../chain";
 import {IBeaconConfig} from "@chainsafe/eth2.0-config";
-import { processSlots } from "@chainsafe/eth2.0-state-transition";
+import {processSlots} from "@chainsafe/eth2.0-state-transition";
 
 export async function produceAttestation(
   {config, db, chain}: {config: IBeaconConfig; db: IBeaconDb; chain: IBeaconChain},
--- a/packages/lodestar/src/api/impl/validator/produceAttestation.ts
+++ b/packages/lodestar/src/api/impl/validator/produceAttestation.ts
@@ -1,9 +1,9 @@
 import {assembleAttestation} from "../../../chain/factory/attestation";
-import {Attestation, BeaconState, BLSPubkey, CommitteeIndex, Slot} from "@chainsafe/eth2.0-types";
+import {Attestation, BLSPubkey, CommitteeIndex, Slot} from "@chainsafe/eth2.0-types";
 import {IBeaconDb} from "../../../db/api";
 import {IBeaconChain} from "../../../chain";
 import {IBeaconConfig} from "@chainsafe/eth2.0-config";
-import {clone} from "@chainsafe/ssz";
+import { processSlots } from "@chainsafe/eth2.0-state-transition";
 
 export async function produceAttestation(
   {config, db, chain}: {config: IBeaconConfig; db: IBeaconDb; chain: IBeaconChain},
@@ -12,11 +12,12 @@ export async function produceAttestation(
   slot: Slot
 ): Promise<Attestation|null> {
   try {
-    const [headState, headBlock, validatorIndex] = await Promise.all([
-      clone(chain.latestState, config.types.BeaconState) as BeaconState,
+    const [headBlock, validatorIndex] = await Promise.all([
       db.block.get(chain.forkChoice.head()),
       db.getValidatorIndex(validatorPubKey)
     ]);
+    const headState = await db.state.get(headBlock.stateRoot);
+    await processSlots(config, headState, slot);
     return await assembleAttestation({config, db}, headState, headBlock, validatorIndex, index, slot);
   } catch (e) {
     throw e;
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -146,14 +146,20 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
       return;
     }
 
-    if(block.slot <= this.latestState.slot) {
+    if(await this.db.block.has(blockHash)) {
+      this.logger.warn(`Block ${blockHash} existed already, no need to process it.`)
+      return;
+    }
+
+    const headBlockRoot = this.forkChoice.head();
+    const headBlock = await this.db.block.get(headBlockRoot);
+    if(block.slot <= headBlock.slot) {
       this.logger.warn(
         `Block ${blockHash.toString("hex")} is in past. ` +
         "Probably fork choice/double propose/processed block. Ignored for now."
       );
       return;
     }
-
 
     await this.processBlock(block, blockHash);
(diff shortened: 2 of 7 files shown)


packages/lodestar/src/api/impl/validator/produceAttestation.ts
```diff
@@ -12,11 +12,12 @@ export async function produceAttestation(
...
 ): Promise<Attestation|null> {
   try {
-    const [headState, headBlock, validatorIndex] = await Promise.all([
-      clone(chain.latestState, config.types.BeaconState) as BeaconState,
+    const [headBlock, validatorIndex] = await Promise.all([
       db.block.get(chain.forkChoice.head()),
       db.getValidatorIndex(validatorPubKey)
     ]);
+    const headState = await db.state.get(headBlock.stateRoot);
+    await processSlots(config, headState, slot);
```
wemeetagain, line 20: I don't think we need this `processSlots` here?
It could go in `assembleAttestation` instead of the while loop at the top.

packages/lodestar/src/chain/chain.ts
```diff
@@ -118,18 +117,17 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
   public async receiveAttestation(attestation: Attestation): Promise<void> {
     const attestationHash = hashTreeRoot(attestation, this.config.types.Attestation);
     this.logger.info(`Received attestation ${attestationHash.toString("hex")}`);
-    const latestState = this.latestState;
     try {
       const attestationSlot: Slot = attestation.data.slot;
-      if(attestationSlot + this.config.params.SLOTS_PER_EPOCH < latestState.slot) {
+      if(attestationSlot + this.config.params.SLOTS_PER_EPOCH < this.latestState.slot) {
```
wemeetagain: Should this be the head state based on fork choice?

packages/lodestar/src/chain/chain.ts
```diff
@@ -338,7 +327,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
     await this.db.setChainHeadRoots(blockRoot, block.stateRoot);
     this.forkChoice.addBlock(block.slot, blockRoot, block.parentRoot, newState.currentJustifiedCheckpoint,
       newState.finalizedCheckpoint);
-    // await this.applyForkChoiceRule();
+    await this.applyForkChoiceRule();
     await this.updateDepositMerkleTree(newState);
```
wemeetagain, line 338: Do we need to handle different versions of the deposit tree in different forks?

wemeetagain (changes requested):

packages/lodestar/src/api/impl/validator/produceAttestation.ts
twoeths: Got that `processSlots` in https://github.com/ethereum/eth2.0-specs/blob/dev/specs/validator/0_beacon-chain-validator.md
```
First, the validator should construct attestation_data, an AttestationData object based upon the state at the assigned slot.

Let head_block be the result of running the fork choice during the assigned slot.
Let head_state be the state of head_block processed through any empty slots up to the assigned slot using process_slots(state, slot).
```

wemeetagain: Ok, we can remove the loop at the top of `assembleAttestation` then, since we're doing the same here.

packages/lodestar/src/chain/chain.ts
twoeths: What I see is fork choice spec does not mention about deposit tree.

Right now we get state of the parent block and process this block into `newState`. So I think storing deposit tree in this `newState` already has the meaning of "fork" here.

Not sure if that's your concern.

twoeths: Oh I found out that we use same merkleTree database in `updateDepositMerkleTree`  which may be an issue, will look into that..

twoeths: Probably we don't have any issues here.
All deposits are sorted, given 2 different states respective to 2 different forked blocks, as long as they have the same `eth1DepositIndex`, `merkleTree` would be the same because they process the same set of deposit.

mpetrunic (commented):
Shouldn't we change head state if "applyForkChoice" changes head?

wemeetagain:
@mpetrunic I think thats what `applyForkChoice` is doing

mpetrunic:
> @mpetrunic I think thats what `applyForkChoice` is doing

Yeah but we won't have state for that?

Shouldn't we run state transition from last common state?

packages/lodestar/src/chain/chain.ts
```diff
@@ -338,7 +329,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
     await this.db.setChainHeadRoots(blockRoot, block.stateRoot);
```
wemeetagain: I think this line should be removed, since we'll be doing it in applyForkChoiceRule if needed

wemeetagain:
`applyForkChoice` is getting called right after the state transition has been run, and the post state stored in the db.
I think it has everything it needs?

mpetrunic:
> `applyForkChoice` is getting called right after the state transition has been run, and the post state stored in the db.
> I think it has everything it needs?

Thsi works for:
```
block A -----> block B
         \
          \
           \
            \
             > block C
```
but what if we have:
```
block A-----> B -----> block C
         \
          \
            > B1-------->C1
```
I guess we don't have to support that scenario yet?

packages/lodestar/src/chain/chain.ts
```diff
@@ -142,6 +142,8 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
 
     if(!await this.db.block.has(block.parentRoot)) {
       this.emit("unknownBlockRoot", block.parentRoot);
+      this.blockProcessingQueue.add(block);
+      return;
     }
 
     if(block.slot <= this.latestState.slot) {
```
wemeetagain: this condition can be changed to discarding if the block.slot is <= the finalized block slot (according to fork choice)

We also probably want a check for if `this.db.block.has(blockHash)` already (duplicate block, already processed)

packages/lodestar/src/chain/chain.ts
```diff
@@ -142,35 +142,29 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
 
-    if(block.slot <= this.latestState.slot) {
+    if(await this.db.block.has(blockHash)) {
+      this.logger.warn(`Block ${blockHash} existed already, no need to process it.`)
+      return;
+    }
+
+    const headBlockRoot = this.forkChoice.head();
+    const headBlock = await this.db.block.get(headBlockRoot);
+    if(block.slot <= headBlock.slot) {
```
wemeetagain: This should be roughly `if(block.slot <= this.forkChoice.finalized.slot)`

and the log message should be something like "block ${b} is not after the finalized checkpoint ${c}"

wemeetagain (changes requested):

wemeetagain (approved):

PR #592, opened 2019-12-20
Fork choice update for 0.9.2

### LINKED ISSUE #588: [0.9.2] Fork choice update
Spec PR: https://github.com/ethereum/eth2.0-specs/pull/1477

twoeths:
## Goal
+ Close #588

### DIFF
--- a/packages/lodestar/src/chain/attestation.ts
+++ b/packages/lodestar/src/chain/attestation.ts
@@ -1,0 +1,100 @@
+import assert from "assert";
+import {IAttestationProcessor, ChainEventEmitter} from "./interface";
+import {ILMDGHOST} from ".";
+import {Attestation, BeaconBlock, Slot, Root} from "@chainsafe/eth2.0-types";
+import {IBeaconConfig} from "@chainsafe/eth2.0-config";
+import {IBeaconDb} from "../db";
+import {getCurrentSlot, computeEpochAtSlot, getAttestingIndices} from "@chainsafe/eth2.0-state-transition";
+import {GENESIS_EPOCH} from "../constants";
+import {ILogger} from "../logger";
+import {hashTreeRoot, signingRoot} from "@chainsafe/ssz";
+
+type RootHex = string;
+
+export class AttestationProcessor implements IAttestationProcessor {
+  private readonly config: IBeaconConfig;
+  private db: IBeaconDb;
+  private logger: ILogger;
+  private chain: ChainEventEmitter;
+  private forkChoice: ILMDGHOST;
+  private pendingAttestations: Record<RootHex, Record<RootHex, Attestation>>;
+  
+  public constructor(chain: ChainEventEmitter, forkChoice: ILMDGHOST, 
+    {config, db, logger}: {config: IBeaconConfig; db: IBeaconDb; logger: ILogger}) {
+    this.config = config;
+    this.db = db;
+    this.logger = logger;
+    this.chain = chain;
+    this.forkChoice = forkChoice;
+    this.pendingAttestations = {};
+  }
+  
+  public async receiveAttestation(attestation: Attestation): Promise<void> {
+    const attestationHash = hashTreeRoot(this.config.types.Attestation, attestation);
+    this.logger.info(`Received attestation ${attestationHash.toString("hex")}`);
+    try {
+      const attestationSlot: Slot = attestation.data.slot;
+      const headBlock = await this.db.block.get(this.forkChoice.head());
+      const state = await this.db.state.get(headBlock.stateRoot);
+      if(attestationSlot + this.config.params.SLOTS_PER_EPOCH < state.slot) {
+        this.logger.verbose(`Attestation ${attestationHash.toString("hex")} is too old. Ignored.`);
+        return;
+      }
+    } catch (e) {
+      return;
+    }
+    const targetRoot = attestation.data.target.root;
+    if (!await this.db.block.has(targetRoot)) {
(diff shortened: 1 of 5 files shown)


wemeetagain (approved):
lgtm

packages/lodestar/src/chain/chain.ts
```diff
@@ -84,6 +85,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
     this.attestationProcessingQueue = queue(async (task: Function) => {
       await task();
     }, 1);
+    this.attesationWaitingQueue = new FastPriorityQueue((a: Attestation, b: Attestation) => a.data.slot < b.data.slot);
```
mpetrunic: Why do you need them ordered by slot and how is that different from processing queue?

I think we just need to add check to prevent processing attestation if we don't yet have targeted block

packages/lodestar/src/chain/chain.ts
```diff
@@ -133,9 +135,34 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
+      this.emit("unknownBlockRoot", targetRoot);
+      this.attesationWaitingQueue.add(attestation);
+      return;
+    }
+    const beaconBlockRoot = attestation.data.beaconBlockRoot;
+    if (!await this.db.block.has(beaconBlockRoot)) {
+      this.emit("unknownBlockRoot", beaconBlockRoot);
+      this.attesationWaitingQueue.add(attestation);
+      return;
+    }
```
mpetrunic: You could squash those if statements into single one

twoeths: I can't because it's not doing the same thing in the body of these `if` statements.

packages/lodestar/src/chain/chain.ts
twoeths: At each slot validators will post attestation for that slot so it's reasonable to process attestation slot by slot.

The processing queue will process attestation without having to check for anything. The waiting queue sits before the processing queue and we stop in the middle if any attestation don't meet our checks below, just like the block queue.

We do need this waiting queue because the spec said "Attestations target be for a known block. If target block is unknown, delay consideration until the block is found"

mpetrunic: YOu are assuming we are dealing with local nodes and validators. In reality we might receive attestation but not a block or we receive attestation too late. There can also be multiple attestations at same slot for different block etc.

According to spec we have two scenarios:
1. We received attestation for known block -> we include it in fork choice
2. We received attestation for unknown block -> emit unknownBlockRoot event -> include block in forkchoice -> check in pending attestation if some are targeting that block -> include attestation

twoeths: So I suppose the approach is:
+ Don't need a new attestation queue
+ Create `Record<blockRoot, Attestation> pendingAttestations` inside chain.ts (or any similar ways)
+ Add check to prevent processing attestations, if it does not pass the check then put it in `pendingAttestations` above
+ When we receive block, get all pending attestations respective to that block and add to processing queue. Remove them from `pendingAttestations`

Is that correct?

mpetrunic: Yeah, few advices:
- try to extract those attestation processing outside of chain module (it will be easier to test and maintain)
- If you implement that, I don't think we need attestation processing queue because we have that queue just so we don't process attestation before block which should't happen now

packages/lodestar/src/chain/attestation.ts
```diff
@@ -0,0 +1,100 @@
...
+import {ILMDGHOST} from ".";
+import {Attestation, BeaconBlock, Hash, Slot} from "@chainsafe/eth2.0-types";
+import {IBeaconConfig} from "@chainsafe/eth2.0-config";
+import {IBeaconDb} from "../db";
+import {getCurrentSlot, computeEpochAtSlot, getAttestingIndices} from "@chainsafe/eth2.0-state-transition";
+import {GENESIS_EPOCH} from "../constants";
+import {ILogger} from "../logger";
+import {hashTreeRoot, signingRoot} from "@chainsafe/ssz";
+
+type RootHex = string;
```
mpetrunic: Can you move this to eth2-types or at least somewhere in utils

packages/lodestar/src/chain/attestation.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export class AttestationProcessor implements IAttestationProcessor {
+  private readonly config: IBeaconConfig;
+  private db: IBeaconDb;
+  private logger: ILogger;
+  private chain: ChainEventEmitter;
+  private forkChoice: ILMDGHOST;
+  private pendingAttestations: Record<RootHex, Record<RootHex, Attestation>>;
```
mpetrunic: It would make more sense like `Record<BlockRootHex, Record<AttestationRootHex, Attestation>>`

Why not Record<BlockRootHex, Array<Attestation>>?

mpetrunic: You could consider Map instead of plain object (Record) for the utility methods.

twoeths: > Why not Record<BlockRootHex, Array>?

It's because I don't want to have duplicate pending attestations for a block. Later on I don't want to process an attestation twice when receiving a block.

mpetrunic: That makes sense

wemeetagain (approved):
Great work! This looks good

PR #598, opened 2019-12-27
Handle wrong fork/finalization on hello

### LINKED ISSUE #478: Handle wrong fork/finalization on hello
Lodestar should disconnect on hello:
1. If `fork_version` doesn’t match the local fork version, since the client’s chain is on another fork. `fork_version` can also be used to segregate testnets.
2. If the (`finalized_root`, `finalized_epoch`) shared by the peer is not in the client's chain at the expected epoch. For example, if Peer 1 sends (root, epoch) of (A, 5) and Peer 2 sends (B, 3) but Peer 1 has root C at epoch 3, then Peer 1 would disconnect because it knows that their chains are irreparably disjoint.

twoeths:
## Goal
+ Close #478

### DIFF
--- a/packages/lodestar/src/sync/reqResp/reqResp.ts
+++ b/packages/lodestar/src/sync/reqResp/reqResp.ts
@@ -30,6 +30,12 @@ export interface ISyncReqRespModules {
   logger: ILogger;
 }
 
+enum GoodByeReasonCode {
+  CLIENT_SHUTDOWN = 1,
+  IRRELEVANT_NETWORK = 2,
+  ERROR = 3,
+}
+
 /**
  * The SyncReqResp module handles app-level requests / responses from other peers,
  * fetching state from the chain and database as needed.
@@ -66,7 +72,7 @@ export class SyncReqResp implements ISyncReqResp {
     this.network.reqResp.removeListener("request", this.onRequest);
     await Promise.all(
       this.network.getPeers().map((peerInfo) =>
-        this.network.reqResp.goodbye(peerInfo, 1n)));
+        this.network.reqResp.goodbye(peerInfo, BigInt(GoodByeReasonCode.CLIENT_SHUTDOWN))));
   }
 
   public onRequest = async (
@@ -100,7 +106,7 @@ export class SyncReqResp implements ISyncReqResp {
       this.network.reqResp.sendResponse(id, e, null);
     }
     if (await this.shouldDisconnectOnHello(request)) {
-      this.network.reqResp.goodbye(peerInfo, 2n);
+      this.network.reqResp.goodbye(peerInfo, BigInt(GoodByeReasonCode.IRRELEVANT_NETWORK));
     }
   }
 
@@ -111,7 +117,7 @@ export class SyncReqResp implements ISyncReqResp {
       return true;
     }
     const startSlot = computeStartSlotOfEpoch(this.config, request.finalizedEpoch);
-    const startBlock = await this.db.block.get(startSlot);
+    const startBlock = await this.db.blockArchive.get(startSlot);
     if (state.finalizedCheckpoint.epoch >= request.finalizedEpoch &&
        !request.finalizedRoot.equals(signingRoot(this.config.types.BeaconBlock, startBlock))) {
       return true;
--- a/packages/lodestar/src/sync/reqResp/reqResp.ts
+++ b/packages/lodestar/src/sync/reqResp/reqResp.ts
@@ -7,7 +7,7 @@ import {
   Epoch, Hash, Slot,
   RequestBody, Hello, Goodbye,
   BeaconBlocksByRangeRequest, BeaconBlocksByRangeResponse,
-  BeaconBlocksByRootRequest, BeaconBlocksByRootResponse,
+  BeaconBlocksByRootRequest, BeaconBlocksByRootResponse, Checkpoint,
 } from "@chainsafe/eth2.0-types";
 import {IBeaconConfig} from "@chainsafe/eth2.0-config";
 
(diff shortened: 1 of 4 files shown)


packages/lodestar/src/sync/reqResp/reqResp.ts
```diff
@@ -97,7 +99,31 @@ export class SyncReqResp implements ISyncReqResp {
...
-    // TODO handle incorrect forkVersion or disjoint finalizedCheckpoint
+    if (await this.shouldDisconnectOnHello(peerInfo, request)) {
+      this.network.reqResp.goodbye(peerInfo, 0n);
+    }
+  }
+
+  public async shouldDisconnectOnHello(peerInfo: PeerInfo, request: Hello): Promise<boolean> {
+    const headBlock = await this.db.block.getChainHead();
+    const state = await this.db.state.get(headBlock.stateRoot);
+    const fork = state.fork;
```
mpetrunic: I think chain module should expose getFork method

packages/lodestar/src/sync/reqResp/reqResp.ts
```diff
@@ -97,7 +99,31 @@ export class SyncReqResp implements ISyncReqResp {
...
+    const epoch = computeEpochOfSlot(this.config, request.headSlot);
+    const forkVersionHeadSlot = (epoch < fork.epoch)? fork.previousVersion : fork.currentVersion;
+    if (!forkVersionHeadSlot.equals(request.headForkVersion)) {
+      return true;
+    }
```
mpetrunic, line 118: I think we just need to check 
```javascript
fork.currentVersion.equals(request.headForkVersion) && request.finalizedRoot.equals(getBlockRoot(this.config, state, request.finalizedEpoch))
//maybe check if we have finalized that epoch
```

mpetrunic (changes requested):

packages/lodestar/src/sync/reqResp/reqResp.ts
twoeths: I also need state for subsequence calls, the new code looks better anyway.

packages/lodestar/src/sync/reqResp/reqResp.ts
```diff
@@ -97,7 +98,25 @@ export class SyncReqResp implements ISyncReqResp {
...
-    // TODO handle incorrect forkVersion or disjoint finalizedCheckpoint
+    if (await this.shouldDisconnectOnHello(peerInfo, request)) {
+      this.network.reqResp.goodbye(peerInfo, 0n);
+    }
+  }
+
+  public async shouldDisconnectOnHello(peerInfo: PeerInfo, request: Hello): Promise<boolean> {
+    const headBlock = await this.db.block.getChainHead();
+    if (!headBlock) {
+      return true;
```
mpetrunic: We should always have chain head at this point. Might be better to throw exception as we shouldn't proceed after this,.

packages/lodestar/src/sync/reqResp/reqResp.ts
```diff
@@ -97,7 +98,25 @@ export class SyncReqResp implements ISyncReqResp {
...
+    }
+    const state = await this.db.state.get(headBlock.stateRoot);
+    if (!state.fork.currentVersion.equals(request.headForkVersion)) {
+      return true;
+    }
+    if (state.finalizedCheckpoint.epoch >= request.finalizedEpoch &&
+       !request.finalizedRoot.equals(getBlockRoot(this.config, state, request.finalizedEpoch))) {
```
mpetrunic: if I'm not mistaken `getBlockRoot` only contains root for recent slots. I think it's safer to fetch block from database by request.finalizedEpoch start slot.

You should try to avoid fetching entire state here as that call isn't cheap, especially considering that this method will be called on each peer connect.

packages/lodestar/src/sync/reqResp/reqResp.ts
```diff
@@ -97,7 +98,25 @@ export class SyncReqResp implements ISyncReqResp {
     } catch (e) {
       this.network.reqResp.sendResponse(id, e, null);
     }
-    // TODO handle incorrect forkVersion or disjoint finalizedCheckpoint
+    if (await this.shouldDisconnectOnHello(peerInfo, request)) {
+      this.network.reqResp.goodbye(peerInfo, 0n);
```
mpetrunic: as per spec it should be 2n

mpetrunic (dismissed):

packages/lodestar/src/sync/reqResp/reqResp.ts
twoeths: I correct the one in `stop()` too

packages/lodestar/src/sync/reqResp/reqResp.ts
twoeths: Ah I had to do that earlier because of my unit test, now we don't have that anymore.
Let's remove it to make it consistent to other places.

packages/lodestar/src/sync/reqResp/reqResp.ts
twoeths: I replaced getBlockRoot by getting block root from database as you said.

Regarding the state, I still need it for other calculations.

packages/lodestar/src/sync/reqResp/reqResp.ts
```diff
@@ -65,7 +66,7 @@ export class SyncReqResp implements ISyncReqResp {
     this.network.reqResp.removeListener("request", this.onRequest);
     await Promise.all(
       this.network.getPeers().map((peerInfo) =>
-        this.network.reqResp.goodbye(peerInfo, 0n)));
```
mpetrunic, line 67: I would extract reason codes to enum

packages/lodestar/src/sync/reqResp/reqResp.ts
```diff
@@ -98,22 +99,21 @@ export class SyncReqResp implements ISyncReqResp {
...
     const headBlock = await this.db.block.getChainHead();
-    if (!headBlock) {
-      return true;
-    }
     const state = await this.db.state.get(headBlock.stateRoot);
     if (!state.fork.currentVersion.equals(request.headForkVersion)) {
       return true;
     }
+    const startSlot = computeStartSlotOfEpoch(this.config, request.finalizedEpoch);
+    const startBlock = await this.db.block.get(startSlot);
```
mpetrunic: If that block is finalized it will be in `this.db.blockArchive`

wemeetagain (approved):
:+1:

PR #599, opened 2019-12-28
Assert Genesis State Correctness

### LINKED ISSUE #557: Verify database genesis
When lodestar is starting, we aren't detecting if database has same genesis state as defined by cli flags
which sometimes causes weird errors.

Probably ideal would be for chain module to throw error on initialization if state is different from genesis state in database (check should be done before we store new genesis state).

error should be something like "Database contains different genesis state, please clean database"

aunyks:
Resolves #557.

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -3,18 +3,18 @@
  */
 
 import assert from "assert";
-import {EventEmitter} from "events";
-import {clone, hashTreeRoot, serialize, signingRoot} from "@chainsafe/ssz";
-import {Attestation, BeaconBlock, BeaconState, Hash, Slot, uint16, uint64, bool} from "@chainsafe/eth2.0-types";
-import {IBeaconConfig} from "@chainsafe/eth2.0-config";
+import { EventEmitter } from "events";
+import { clone, hashTreeRoot, serialize, signingRoot } from "@chainsafe/ssz";
+import { Attestation, BeaconBlock, BeaconState, Hash, Slot, uint16, uint64 } from "@chainsafe/eth2.0-types";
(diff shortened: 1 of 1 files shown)


wemeetagain (approved):
LGTM

packages/lodestar/src/chain/chain.ts
```diff
@@ -215,6 +217,13 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
     genesisBlock.stateRoot = stateRoot;
     const blockRoot = signingRoot(this.config.types.BeaconBlock, genesisBlock);
     this.latestState = genesisState;
+    // Determine whether a genesis state already in
+    // the database matches what we were provided
+    const storedGenesisBlock = await this.db.block.getBlockBySlot(GENESIS_SLOT);
```
mpetrunic, line 222: where is GENESIS_SLOT defined?

shouldn't this be this.config.params.GENESIS_SLOT

aunyks: It’s a [constant that was imported](https://github.com/aunyks/lodestar/blob/a4abbd9f9acad4b8f6e46f92d1ead186800c0b36/packages/lodestar/src/chain/chain.ts#L11).

mpetrunic: Could you try to remove that constant from that file and replace occurencies with this.config.params.GENESIS_SLOT. Not sure in what case genesis slot could be different from 0 but nevertheless

wemeetagain: I think that can be done separately, as needed. It may touch a bunch of other files and have some consequences we haven't fully thought through.

PR #600, opened 2019-12-28
Fix CI on 0.9.x

twoeths:
+ Fix e2e
+ Spec tests should be fixed in a separate PR


wemeetagain (approved):

packages/eth2.0-state-transition/src/util/duties.ts
```diff
@@ -25,8 +25,6 @@ export function getCommitteeAssignment(
   validatorIndex: ValidatorIndex
 ): CommitteeAssignment {
 
-  const nextEpoch = getCurrentEpoch(config, state) + 1;
-  assert(epoch <= nextEpoch);
```
mpetrunic, line 29: why did you remove this?

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -106,10 +106,21 @@ export class AttestationService {
...
+      if (!isFound && equals(this.config.types.Attestation, attestation, wireAttestation)) {
+        isFound = true;
+      }
+      if (equals(this.config.types.AttestationData, wireAttestation.data, attestation.data)) {
+        compatibleAttestations.push(wireAttestation);
+      }
+    }
+    if (!isFound) {
+      compatibleAttestations.push(attestation);
+    }
```
mpetrunic: I think this achieves same this but it's more readable:
```javascript
const compatibleAttestations = wireAttestations.filter((wireAttestation) => {
   return equals(this.config.types.AttestationData, wireAttestation.data, attestation.data)
                 // prevent including aggregator attestation twice
                 && ! equals(this.config.types.Attestation, attestation, wireAttestation)
});
 compatibleAttestations.push(attestation);
```

mpetrunic (changes requested):

packages/eth2.0-state-transition/src/util/duties.ts
twoeths: yeah I'm concerning about this too. When I run e2e `dev.test.ts` it failed at this line.
+ Given SLOTS_PER_EPOCH=5
+ At slot 5 => epoch = 1, the validator asks for duties of next epoch, which is 2. In the above context `epoch=2`
+ But `nextEpoch` is evaluated as 1 because it only see blocks 0,1,2,3,4 => `getCurrentEpoch` evaluated to 0

I don't see anything wrong here so I have to comment this out. Do you think this is because we changed something in validator?

mpetrunic: Hmm, I think it's okay to bump next epoch = currentEpoch +2 as this is just to protect us from doing way too many epoch transitions if there is malicious user

wemeetagain (approved):

PR #604, opened 2019-12-31
[DNM] Investigate 'too many open files' issue

twoeths:


twoeths:
Figured out the reason why the test failed

PR #606, opened 2020-01-07
Don't Return Unwanted Slot 0 on BeaconBlocksByRange Call

### LINKED ISSUE #602: beaconBlocksByRange doesn't respect startSlot
**Describe the bug**

When requesting blocks by range with start:1, it will return block with slot 0 and onward.
When requesting blocks by range with start: 2, it will return block with slot 2 and onward.

**Expected behavior**

If start:1 block with slot 0 should not be returned

aunyks:
Fixes #602. It appears that unwanted blocks with slot 0 are returned whenever BeaconBlocksByRange's `step` and `lowerLimit` arguments are equal, since `(0 - lowerLimit) % step === 0` is true whenever the two are equal.

Ensuring that some `slot` is greater than or equal to the `lowerLimit` should fix this.

### DIFF
--- a/packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
+++ b/packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
@@ -1,9 +1,9 @@
-import { BeaconBlock } from "@chainsafe/eth2.0-types";
-import { IBeaconConfig } from "@chainsafe/eth2.0-config";
-import { BulkRepository } from "../repository";
-import { IDatabaseController } from "../../../controller";
-import { Bucket, encodeKey } from "../../../schema";
-import { serialize, deserialize } from "@chainsafe/ssz";
+import {BeaconBlock} from "@chainsafe/eth2.0-types";
+import {IBeaconConfig} from "@chainsafe/eth2.0-config";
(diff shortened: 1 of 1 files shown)


packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
```diff
@@ -45,7 +45,7 @@ export class BlockArchiveRepository extends BulkRepository<BeaconBlock> {
       .map((datum) => deserialize(this.type, datum))
```
wemeetagain, line 45: Not exactly related to this PR, but it seems strange that `safeLowerLimit` is set to a number or Buffer.

aunyks: Yeah, I’ve been thinking the same thing. I assumed it has some significance, so I never decided to delete it in case it broke something somewhere else.

wemeetagain: It seems to work as expected, even with `Buffer.alloc(0)`. It should probably be just `0` though.
We can change in a later PR

wemeetagain (approved):

PR #608, opened 2020-01-11
Move cli to a separate package

### LINKED ISSUE #535: Create CLI as a seperate package
Currently, the CLI is within the lodestar package. The idea is to move that into it's own package, so that the executable lives there. This will allow us to have a unified CLI to spin up a beacon node or validator client all in one place.

This is important to maintain our design philosophy and architecture.
The idea is our modules should not care about the external resources that we use, as long as the interface is compliant. Configuration of external resources should happen outside.

twoeths:
## Goal
+ Close #535


twoeths:
@wemeetagain this is close to done. One question: ILogger and WinstonLogger are staying in lodestar/cli/lodestar-validator. I intend to move them to eth2.0-util, what do you think?

mpetrunic (commented):
In my opinion ConfigurationField related interfaces and logger should be imported from lodestar library instead of copy pasted to cli

packages/cli/package.json
```diff
@@ -0,0 +1,78 @@
+{
+  "name": "@chainsafe/cli",
```
wemeetagain: I would call this @chainsafe/lodestar-cli

packages/cli/package.json
```diff
@@ -0,0 +1,78 @@
...
+  "private": true,
+  "version": "0.1.0",
+  "description": "Command line interface for other packages",
```
wemeetagain: Command line interface for lodestar

wemeetagain (commented):
This is a great start!

I think ILogger and WinstonLogger should move to util. If we do something like below, we will want a lower level dependency to import these from. We might even consider breaking into a `lodestar-logger` package.

I think each module's ConfigurationModule can be pulled into the cli, so that all parsing and input validation is done on the cli level. We can also move toml processing to the cli.

This whole exercise, looking at this code, makes me think we should consider splitting out lodestar into packages of its own, though this is probably a separate discussion.
eg:
rename `eth2.0-util` -> `lodestar-util` (move `lodestar/src/util` there too)
`lodestar-db`
`lodestar-eth1`
`lodestar-network`
`lodestar-sync`
`lodestar-chain`
`lodestar-oppool`
`lodestar-sync`
`lodestar-api`
`lodestar-beacon`
`lodestar-validator`
`lodestar-cli`

The dependency tree would look something like in [this diagram](https://hackmd.io/L5vwdaXGTH-IqYtv6fqVRA#Beacon-Chain-Modules) with a few additions

ec2:
I like @wemeetagain 's suggestion in changing to `lodestar-*`. Should maybe be in another PR though

twoeths:
> I think each module's ConfigurationModule can be pulled into the cli, so that all parsing and input validation is done on the cli level

@wemeetagain  I also see IConfigurationModule being used in lodestar src/api/rpc and src/api/rest. Probably we have to keep it in lodestar as @mpetrunic commented above.

wemeetagain:
I think src/api/rpc and src/api/rest have IConfigurationModule just so the `api`-wide IConfigurationModule file (src/api/options.ts) isn't really big.

I think my issue is that IConfigurationModule currently does a lot of things, some of which are CLI responsibilities. I would prefer to split out the CLI responsibilities entirely to the CLI package. This includes option descriptions, CLI flags, toml file section and key names, and any user input parsing/hydrating steps.

The validation is something that should/could probably stay with each module. (I'd love to see us use something like https://validatejs.org/)

We will essentially be providing two interfaces for developers/users to chose from:
1. the API interface. This is the JS-level interface to our modules and chain, a way to extend lodestar, or use in a browser context. A JS developer would use any / all lodestar modules along with other JS code. We would export the module itself as well as a validation schema or validation function for config options. Each module will expect its configuration as a javascript object (eg: `IChainOptions`).
2. the CLI interface. This is a higher level interface for end users of lodestar. This would be a way to run the chain or a utility from the command line, configuring lodestar through a series of flags and a toml file. It would first parse all user input, derive a configuration object to pass to lodestar (by parsing command line and toml file input). Then it would run the appropriate module-level configuration validations + use the API interface to boot up the chain.

packages/lodestar-cli/src/commands/dev.ts
```diff
@@ -10,29 +10,32 @@ import PeerId from "peer-id";
...
+import {BeaconNode} from "@chainsafe/lodestar/lib/node";
+import {IBeaconNodeOptions} from "@chainsafe/lodestar/lib/node/options";
 import {generateCommanderOptions, optionsToConfig,} from "../util";
-import {getTomlConfig, rmDir} from "../../util/file";
+import {rmDir} from "@chainsafe/lodestar/lib/util/file";
 import {config as minimalConfig} from "@chainsafe/eth2.0-config/lib/presets/minimal";
-import {InteropEth1Notifier} from "../../eth1/impl/interop";
-import {quickStartOptionToState} from "../../interop/cli";
+import {InteropEth1Notifier} from "@chainsafe/lodestar/lib/eth1/impl/interop";
+import {quickStartOptionToState} from "../lodestar/interop/cli";
```
wemeetagain, line 20: We shouldn't have any relative path references across packages.
They should all be referenced through the package name.

wemeetagain (changes requested):

packages/lodestar-cli/src/commands/dev.ts
twoeths: `lodestar` is a folder name under `src` inside the cli package. The idea is to have package name level in folder tree so that we can move files from other package to cli if needed.

wemeetagain: ohhh, gotcha

wemeetagain (approved):

PR #609, opened 2020-01-14
Changes to utilize new gindex code

protolambda:
Changed merkle-tree api slightly, to use regular numbers where allowable (like depth). Also see PR in merkle-tree library repo.


wemeetagain (approved):
Thanks!

PR #610, opened 2020-01-15
Gossip message validation

### LINKED ISSUE #552: Add gossip message validation
- implement validation as per spec
- attach validation to messages resending (echoing) and publishing

twoeths:
## Goal
+ Close #552 
+ Refer to https://github.com/ChainSafe/gossipsub-js/issues/54

### DIFF
--- a/packages/lodestar/src/network/gossip/gossipsub.ts
+++ b/packages/lodestar/src/network/gossip/gossipsub.ts
@@ -4,7 +4,7 @@ import {getGossipTopic, isAttestationSubnetTopic, getSubnetFromAttestationSubnet
 import {GossipEvent} from "./constants";
 import {IBeaconConfig} from "@chainsafe/eth2.0-config";
 import {utils} from "libp2p-pubsub";
-import {ATTESTATION_SUBNET_COUNT, GOSSIP_MAX_SIZE} from "../../constants";
+import {GOSSIP_MAX_SIZE} from "../../constants";
 import {AnySSZType, deserialize} from "@chainsafe/ssz";
 import assert from "assert";
 import {ILogger} from "@chainsafe/eth2.0-utils/lib/logger";
@@ -66,10 +66,8 @@ export class LodestarGossipsub extends Gossipsub {
 
 
   private getTopicValidator(topic: string): GossipMessageValidatorFn {
-    for(let subnet = 0; subnet < ATTESTATION_SUBNET_COUNT; subnet++) {
-      if (topic === getGossipTopic(GossipEvent.ATTESTATION_SUBNET, "ssz", new Map([["subnet", String(subnet)]]))) {
-        return this.validator.isValidIncomingCommitteeAttestation as GossipMessageValidatorFn;
-      }
+    if (isAttestationSubnetTopic(topic)) {
+      return this.validator.isValidIncomingCommitteeAttestation as GossipMessageValidatorFn;
     }
 
     let result: Function;
--- a/packages/lodestar/src/network/gossip/gossipsub.ts
+++ b/packages/lodestar/src/network/gossip/gossipsub.ts
@@ -1,0 +1,137 @@
+import Gossipsub, {IGossipMessage, Registrar, Options} from "libp2p-gossipsub";
+import {IGossipMessageValidator, GossipObject, GossipMessageValidatorFn} from "./interface";
+import {getGossipTopic, isAttestationSubnetTopic, getSubnetFromAttestationSubnetTopic} from "./utils";
+import {GossipEvent} from "./constants";
+import {IBeaconConfig} from "@chainsafe/eth2.0-config";
+import {utils} from "libp2p-pubsub";
+import {ATTESTATION_SUBNET_COUNT, GOSSIP_MAX_SIZE} from "../../constants";
+import {AnySSZType, deserialize} from "@chainsafe/ssz";
+import assert from "assert";
+import {ILogger} from "@chainsafe/eth2.0-utils/lib/logger";
+
+/**
+ * This validates messages in Gossipsub and emit the transformed messages.
+ * We don't want to double deserialize messages for performance benefit.
+ */
+export class LodestarGossipsub extends Gossipsub {
+  private transformedObjects: Map<string, GossipObject>;
+  private config: IBeaconConfig;
+  private validator: IGossipMessageValidator;
+  private readonly  logger: ILogger;
+
+  constructor (config: IBeaconConfig, validator: IGossipMessageValidator, logger: ILogger, peerInfo: PeerInfo,
+    registrar: Registrar, options: Options = {}) {
+    super(peerInfo, registrar, options);
+    this.transformedObjects = new Map();
+    this.config = config;
+    this.validator = validator;
+    this.logger = logger;
+  }
+
+  public async validate(rawMessage: IGossipMessage): Promise<boolean> {
+    const message: IGossipMessage = utils.normalizeInRpcMessage(rawMessage);
+    assert(message.topicIDs && message.topicIDs.length === 1, `Invalid topicIDs: ${message.topicIDs}`);
+    assert(message.data.length <= GOSSIP_MAX_SIZE, `Message exceeds size limit of ${GOSSIP_MAX_SIZE} bytes`);
+    const topic = message.topicIDs[0];
+
+    let isValid;
+    let transformedObj: GossipObject;
+    try {
+      const validatorFn = this.getTopicValidator(topic);
+      const objSubnet = this.deserializeGossipMessage(topic, message);
+      transformedObj = objSubnet.object;
+      isValid = await validatorFn(transformedObj, objSubnet.subnet);
+    } catch (err) {
+      isValid = false;
+      this.logger.error(`Cannot validate message from ${message.from}, topic ${message.topicIDs}, error: ${err}`);
+    }
+    if (isValid && transformedObj) {
+      this.transformedObjects.set(this.getKey(message), transformedObj);
+    }
+    return isValid;
+  }
+
+  public _emitMessage(topics: string[], message: IGossipMessage): void {
+    const subscribedTopics = super.getTopics();
+    topics.forEach((topic) => {
+      if (subscribedTopics.includes(topic)) {
+        const transformedObj = this.transformedObjects.get(this.getKey(message));
+        if (transformedObj) {
+          super.emit(topic, transformedObj);
+          this.transformedObjects.delete(this.getKey(message));
+        }
+      }
+    });
+  }
+
+
+  private getTopicValidator(topic: string): GossipMessageValidatorFn {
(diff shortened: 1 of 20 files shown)


packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  constructor (config: IBeaconConfig, validator: IGossipMessageValidator, peerInfo: PeerInfo,
+    registrar: Registrar, options: Options = {}) {
+    super(peerInfo, registrar, options);
+    this.transformedMessages = new Map();
+    this.config = config;
+    this.validator = validator;
+  }
+
+  public async validate(rawMessage: IGossipMessage): Promise<boolean> {
```
mpetrunic, line 31: This method is not really maintainable.
I would suggest to extract each validation to separate function, create validation map which maps event => validation function. Same for deserialization

twoeths: I added `getTopicValidator` and `getGossipObject` to make `validate` easier to read and maintain

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -89,15 +49,67 @@ export class LodestarGossipsub extends Gossipsub {
...
+      case getGossipTopic(GossipEvent.ATTESTATION, "ssz"):
+        return this.validator.isValidIncomingUnaggregatedAttestation as GossipMessageValidatorFn;
+      case getGossipTopic(GossipEvent.AGGREGATE_AND_PROOF, "ssz"):
+        return this.validator.isValidIncomingAggregateAndProof as GossipMessageValidatorFn;
+      case getGossipTopic(GossipEvent.ATTESTER_SLASHING, "ssz"):
+        return this.validator.isValidIncomingAttesterSlashing as GossipMessageValidatorFn;
+      case getGossipTopic(GossipEvent.PROPOSER_SLASHING, "ssz"):
+        return this.validator.isValidIncomingProposerSlashing as GossipMessageValidatorFn;
+      case getGossipTopic(GossipEvent.VOLUNTARY_EXIT, "ssz"):
+        return this.validator.isValidIncomingVoluntaryExit as GossipMessageValidatorFn;
```
mpetrunic: I think codeclimate will complain about too many returns. You could assign that fn to variable and cast only once as well :)

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -89,15 +49,67 @@ export class LodestarGossipsub extends Gossipsub {
...
+      case getGossipTopic(GossipEvent.AGGREGATE_AND_PROOF, "ssz"):
+        return {object: deserializeGossipMessage<AggregateAndProof>(message, 
+          this.config.types.AggregateAndProof)};
+      case getGossipTopic(GossipEvent.ATTESTER_SLASHING, "ssz"):
+        return {object: deserializeGossipMessage<AttesterSlashing>(message, this.config.types.AttesterSlashing)};
+      case getGossipTopic(GossipEvent.PROPOSER_SLASHING, "ssz"):
+        return {object: deserializeGossipMessage<ProposerSlashing>(message, this.config.types.ProposerSlashing)};
+      case getGossipTopic(GossipEvent.VOLUNTARY_EXIT, "ssz"):
+        return {object: deserializeGossipMessage<SignedVoluntaryExit>(message,
+          this.config.types.SignedVoluntaryExit)};
```
mpetrunic: Same this as above, assign type config to variable and reduce code

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -89,15 +49,67 @@ export class LodestarGossipsub extends Gossipsub {
...
+      default:
+        throw new Error(`cannot get gossip object for topic ${topic}`);
```
mpetrunic: Maybe "Don't know how to deserialize object received under topic ${topic}" would be more appropriate

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -31,56 +31,16 @@ export class LodestarGossipsub extends Gossipsub {
...
-    if (isValid && message.topicIDs.includes(getGossipTopic(GossipEvent.BLOCK, "ssz"))) {
-      const signedBlock = deserializeGossipMessage<SignedBeaconBlock>(message, this.config.types.SignedBeaconBlock);
-      isValid = await this.validator.isValidIncomingBlock(signedBlock);
-      transformeObj = signedBlock;
+    let transformedObj: GossipObject;
+    for (const topic of message.topicIDs) {
+      const validatorFn = this.getTopicValidator(topic);
+      const objSubnet = this.getGossipObject(topic, message);
+      transformedObj = objSubnet.object;
+      isValid = isValid && await validatorFn(transformedObj, objSubnet.subnet);
```
mpetrunic: You should add try/catch in case deserialization or validation throws exception

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -89,15 +49,67 @@ export class LodestarGossipsub extends Gossipsub {
...
+      case getGossipTopic(GossipEvent.PROPOSER_SLASHING, "ssz"):
+        return this.validator.isValidIncomingProposerSlashing as GossipMessageValidatorFn;
+      case getGossipTopic(GossipEvent.VOLUNTARY_EXIT, "ssz"):
+        return this.validator.isValidIncomingVoluntaryExit as GossipMessageValidatorFn;
+      default:
+        throw new Error(`no validator for topic ${topic}`); 
+    }
+  }
+
+  private getGossipObject(topic: string, message: IGossipMessage): { object: GossipObject; subnet?: number} {
```
mpetrunic: I would rename this to `deserializeGossipMessage`

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  public async validate(rawMessage: IGossipMessage): Promise<boolean> {
+    const message: IGossipMessage = utils.normalizeInRpcMessage(rawMessage);
+    if (!message.topicIDs) {
+      return false;
+    }
+
+    let isValid = true;
+    let transformedObj: GossipObject;
+    for (const topic of message.topicIDs) {
```
mpetrunic: Should we discard messages with more than one topic id?

twoeths: yeah probably because when we only publish single topic id but let's double check.
@wemeetagain wdyt?

wemeetagain: asked this question on eth2 discord, here's proto's answer:
Either gracefully process for each topic, or discard (with an error log) to catch accidental multiple topic IDs

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -0,0 +1,140 @@
...
+      try {
+        const validatorFn = this.getTopicValidator(topic);
+        const objSubnet = this.deserializeGossipMessage(topic, message);
```
wemeetagain: Now that we have this validation method, I think the assertion in `deserializeGossipMessage`(from utils) can live here (and we can delete `deserializeGossipMessage`)

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -0,0 +1,140 @@
...
+        break;
+      default:
+        throw new Error(`No validator for topic ${topic}`); 
+    }
+    return result as GossipMessageValidatorFn;
+  }
+
+  private deserializeGossipMessage(topic: string, message: IGossipMessage): { object: GossipObject; subnet?: number} {
+    for(let subnet = 0; subnet < ATTESTATION_SUBNET_COUNT; subnet++) {
+      if (topic === getGossipTopic(GossipEvent.ATTESTATION_SUBNET, "ssz", new Map([["subnet", String(subnet)]]))) {
```
wemeetagain: Can we use regex here to avoid looping thru subnets?

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -0,0 +1,140 @@
...
+        return {object: deserializeGossipMessage<Attestation>(message, this.config.types.Attestation), subnet};
+      }
+    }
+    let objType: AnySSZType;
+    switch(topic) {
+      case getGossipTopic(GossipEvent.BLOCK, "ssz"):
```
wemeetagain, line 106: Right now "ssz" is hardcoded everywhere, which is fine for now but will likely need to be changed. We should anticipate having gossip managers for different encoding schemed (like ssz_snappy needed for mainnet)

wemeetagain (changes requested):

packages/lodestar/src/network/gossip/gossipsub.ts
mpetrunic: I'm working on generalizing that in snappy compression PR

packages/lodestar/src/network/gossip/gossipsub.ts
twoeths: I'll go with the 2nd option to make it simple.

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -0,0 +1,137 @@
...
+          super.emit(topic, transformedObj);
+          this.transformedObjects.delete(this.getKey(message));
+        }
+      }
+    });
+  }
+
+
+  private getTopicValidator(topic: string): GossipMessageValidatorFn {
+    for(let subnet = 0; subnet < ATTESTATION_SUBNET_COUNT; subnet++) {
```
wemeetagain: use regex here too

packages/lodestar/src/network/gossip/utils.ts
```diff
@@ -29,7 +28,14 @@ export function getAttestationSubnet(attestation: Attestation): string {
...
 }
 
-export function deserializeGossipMessage<T>(msg: IGossipMessage, type: AnySSZType): T {
-  assert(msg.data.length <= GOSSIP_MAX_SIZE, `Message exceeds size limit of ${GOSSIP_MAX_SIZE} bytes`);
-  return deserialize(type, msg.data);
+export function isAttestationSubnetTopic(topic: string): boolean {
+  return AttestationSubnetRegExp.test(topic);
+}
+
+export function getSubnetFromAttestationSubnetTopic(topic: string): number {
```
wemeetagain, line 34: You should be able to capture the subnet with regex: https://javascript.info/regexp-groups

wemeetagain (changes requested):

wemeetagain (dismissed):

wemeetagain:
file conflict

twoeths:
Resolved file conflict.

wemeetagain (approved):

PR #611, opened 2020-01-15
Use bls-keygen to generate random private keys

### LINKED ISSUE #571: Add bls-keygen to bls library
- integrate bls-keygen into PrivateKey class

aunyks:
Resolves #571.


wemeetagain (changes requested):
I think we only need bls-keygen as a dependency in the bls package.
Other than that, looks good!

aunyks:
Gotcha. Should be resolved now @wemeetagain.

wemeetagain (approved):
:tada:

PR #619, opened 2020-01-26
[Deposits] Read from json file

GregTheGreek:
This is a testing flag, it allows users to import their keys from a json file.


mpetrunic (approved):

wemeetagain (approved):
lgtm

PR #622, opened 2020-01-28
Implement chunkify response for network req/resp domain

### LINKED ISSUE #530: Implement chunkify response in sync protocol

twoeths:
## Goal
+ Close #530 
+ Waiting for the new libp2p integration

### DIFF
--- a/packages/lodestar/src/network/interface.ts
+++ b/packages/lodestar/src/network/interface.ts
@@ -17,12 +17,9 @@ import {Method, RequestId} from "../constants";
 import StrictEventEmitter from "strict-event-emitter-types";
 import {IGossip} from "./gossip/interface";
 
-// req/resp
-export type Response = {err?: Error; output?: ResponseBody};
+export type ResponseCallbackFn = (response: {err?: Error; output?: ResponseBody}) => void;
 
-export type ResponseCallbackFn = (response: Response) => void;
-
-export type ChunkResponse = {err?: Error; output: Status | Goodbye | SignedBeaconBlock};
+export type ResponseChunk = {err?: Error; output: Status | Goodbye | SignedBeaconBlock};
 
 interface IRespEvents {
   [responseEvent: string]: ResponseCallbackFn;
--- a/packages/lodestar/src/network/interface.ts
+++ b/packages/lodestar/src/network/interface.ts
@@ -10,7 +10,7 @@ import {
   BeaconBlocksByRootResponse,
   Goodbye,
   RequestBody,
-  ResponseBody, Status,
+  ResponseBody, Status, SignedBeaconBlock,
 } from "@chainsafe/eth2.0-types";
 
 import {Method, RequestId} from "../constants";
@@ -18,8 +18,11 @@ import StrictEventEmitter from "strict-event-emitter-types";
 import {IGossip} from "./gossip/interface";
 
 // req/resp
+export type Response = {err?: Error; output?: ResponseBody};
 
-export type ResponseCallbackFn = (err: Error|null, output: ResponseBody|null) => void;
+export type ResponseCallbackFn = (response: Response) => void;
+
+export type ChunkResponse = {err?: Error; output: Status | Goodbye | SignedBeaconBlock};
 
 interface IRespEvents {
   [responseEvent: string]: ResponseCallbackFn;
--- a/packages/lodestar/src/network/reqResp.ts
+++ b/packages/lodestar/src/network/reqResp.ts
@@ -27,7 +27,8 @@ import {
   RESP_TIMEOUT,
 } from "../constants";
 import {ILogger} from  "@chainsafe/eth2.0-utils/lib/logger";
-import {createResponseEvent, createRpcProtocol, randomRequestId,} from "./util";
+import {createResponseEvent, createRpcProtocol, randomRequestId, encodeChunkifyResponse, 
+  decodeChunkifyResponse} from "./util";
 
 import {IReqResp, ReqEventEmitter, RespEventEmitter, ResponseCallbackFn} from "./interface";
 import {INetworkOptions} from "./options";
@@ -169,26 +170,7 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements 
   }
 
   private encodeResponse(method: Method, body: ResponseBody): Buffer {
-    let output= Buffer.alloc(0);
-    switch (method) {
-      case Method.Status:
-        output = serialize(this.config.types.Status, body);
-        break;
-      case Method.Goodbye:
-        output = serialize(this.config.types.Goodbye, body);
-        break;
-      case Method.BeaconBlocksByRange:
-        output = serialize(this.config.types.BeaconBlocksByRangeResponse, body);
-        break;
-      case Method.BeaconBlocksByRoot:
-        output = serialize(this.config.types.BeaconBlocksByRootResponse, body);
-        break;
-    }
-    return Buffer.concat([
-      Buffer.alloc(1),
-      Buffer.from(varint.encode(output.length)),
-      output,
-    ]);
+    return encodeChunkifyResponse(this.config, method, body);
   }
   private encodeResponseError(err: Error): Buffer {
     const b = Buffer.from("c" + err.message);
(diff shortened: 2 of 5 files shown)


packages/lodestar/src/network/util.ts
```diff
@@ -52,3 +56,86 @@ export async function initializePeerInfo(peerId: PeerId, multiaddrs: string[]):
...
+      return decodeSingleChunk(config, method, chunks) as ResponseBody;
+    case Method.BeaconBlocksByRange:
+    case Method.BeaconBlocksByRoot:
+      return decodeBlockChunks(config, method, chunks);
+  }
+}
+
+function encodeBlockChunks(config: IBeaconConfig, blocks: BeaconBlock[]): Buffer {
+  const chunkArr = blocks.map(block => encodeSingleChunk(serialize(config.types.BeaconBlock, block)));
+  return Buffer.concat(chunkArr);
```
mpetrunic: Maybe I'm missing something but aren't we suppose to send each chunk as soon as it's ready instead of waiting to serialize all blocks?

twoeths: yeah this is a good point. I don't see a way to do that with the current version of libp2p, looking into v0.27 of libp2p now.

mpetrunic: Even with new version, I think code will need reorg. You will need access to connection I guess

twoeths: yeah. I'd keep this method unchanged and reorg the code in other parts. Whenever we have a bunch of blocks, we'll call this method to get encoded Buffer and send to client.

mpetrunic: Not sure you can, as you are suppose to do something like:

```javascript
blocks.forEach(block => {
  connection.write(
     encodeSingleChunk(serialize(config.types.BeaconBlock, block)))
  ); 
})
```

twoeths: do you think we should write to connection/stream per block or per n blocks? I thought I'd send per n block. This depends on the way we query blocks.

mpetrunic: per block

twoeths: I think I found the way it works in the new libp2p. We just need to give an array of buffer to the stream and libp2p will write to the stream one by one. At the client side, it'll need to read from the stream chunk by chunk and decode each until EOF.

mpetrunic: If I understood you, we will again have situation where we serialize all blocks into array of buffers before sending them?

twoeths: currently I concatenate all chunks into a single Buffer. I'll change to giving an array of buffers to libp2p stream. Internally, libp2p will send each buffer/chunk one by one.  I checked at the client side, it received chunks one by one like:
```
received from server: one
received from server: two
```

mpetrunic: But we will still wait for all blocks to serialize before sending them?

wemeetagain: depends on the how our db api returns blocks.
If we could yield block by block, we could serialize individually, then push each block one by one into the stream.
If the db just returns an array of blocks, then it probably just makes sense to map(serialize) and push the entire array at once

twoeths: if we don't want to serialize all blocks before sending to stream, we can use something like https://github.com/alanshaw/it-pipe , that's the lib I see in every example of the new libp2p.
Server will get one block => serialize => send to stream => client gets that serialized data => deserialize => server get next block => serialize => send to stream => ...
I think this clears everybody's concern.

wemeetagain: take a look at https://github.com/alanshaw/it-awesome too, plenty of tools to pick from

mpetrunic: We could expose stream method since database is returning blocks one-by-one as default behavior (we are just waiting for all of them. But I think priority is just sending blocks while serializing as I'm pretty sure it has more significant performance impact

twoeths: I got your point

packages/lodestar/src/network/interface.ts
```diff
@@ -10,16 +10,19 @@ import {
...
 import StrictEventEmitter from "strict-event-emitter-types";
 import {IGossip} from "./gossip/interface";
 
 // req/resp
+export type Response = {err?: Error; output?: ResponseBody};
 
-export type ResponseCallbackFn = (err: Error|null, output: ResponseBody|null) => void;
+export type ResponseCallbackFn = (response: Response) => void;
+
+export type ChunkResponse = {err?: Error; output: Status | Goodbye | SignedBeaconBlock};
```
wemeetagain: Can you call this `ResponseChunk`?
We shouldn't need a `Response` and a `ResponseChunk` type. (every response should be a `ResponseChunk`)
If we do have a higher level `Response`-ish interface it should expose an `Iterable<ResponseChunk>` rather than an array, eg: `SignedBeaconBlock[]`

See the "relaxed BNF grammer" code section in the spec:
https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/p2p-interface.md#reqresp-interaction

packages/lodestar/src/network/util.ts
```diff
@@ -50,3 +55,105 @@ export async function initializePeerInfo(peerId: PeerId, multiaddrs: string[]):
...
+function encodeResponseError(err: Error): Buffer {
+  const b = Buffer.from("c" + "error_message:" + err.message);
+  b[0] = err.message === ERR_INVALID_REQ ? 1 : 2;
+  return b;
+}
+
+function encodeSingleChunk(data: Buffer): Buffer {
+  return Buffer.concat([
+    Buffer.alloc(1),
+    Buffer.from(varint.encode(data.length)),
```
wemeetagain, line 124: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/p2p-interface.md#ssz-encoding-strategy-with-or-without-snappy

So one thing to keep in mind is that the "encoding-dependent" header is dependent on the header (lol).
Here we're hard-coding the varint, which should be ok for now, but should be somehow move the code in a direction where we can abide by different encodings.

packages/lodestar/src/network/util.ts
```diff
@@ -50,3 +55,105 @@ export async function initializePeerInfo(peerId: PeerId, multiaddrs: string[]):
...
+export function decodeChunkifyResponse(config: IBeaconConfig, method: Method, chunks: Buffer): ResponseBody {
+  switch (method) {
+    case Method.Status:
+    case Method.Goodbye:
+      return decodeSingleChunk(config, method, chunks) as ResponseBody;
+    case Method.BeaconBlocksByRange:
+    case Method.BeaconBlocksByRoot:
+      // lodestar response multple single block chunks
+      // but other clients might response multiple n block chunks
+      // this handle both cases
```
wemeetagain: We should not be supporting multiple-block chunks.
According to the spec, "Each successful response_chunk MUST contain a single SignedBeaconBlock payload."

packages/lodestar/src/network/util.ts
```diff
@@ -50,3 +55,105 @@ export async function initializePeerInfo(peerId: PeerId, multiaddrs: string[]):
   multiaddrs.forEach((ma) => peerInfo.multiaddrs.add(ma));
   return peerInfo;
 }
+
+export function encodeChunkifyResponse(config: IBeaconConfig, method: Method, {err, output}: ChunkResponse): Buffer {
```
wemeetagain: Can we call these functions `encodeResponseChunk` and `decodeResponseChunk` to match the `ResponseChunk` name?

wemeetagain (changes requested):
Looking good so far, just a few requests re: naming, spec compliance, simplifying interfaces

packages/lodestar/src/network/reqResp.ts
```diff
@@ -96,19 +96,19 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
                   const chunkResponses = blocks.map(block => ({output: block}));
                   result.push(...(chunkResponses));
                 } else {
-                  result.push(response as ChunkResponse);
+                  result.push(response as ResponseChunk);
                 }
               }  
               return result;
             },
```
mpetrunic, line 91: extract, handle and transform iterators into separate functions

```
pipe(
   stream.source,
   handleRpcRequestStream(this.handleRequest),
   handleResponseStream(),
   stream.sink
)
```

wemeetagain (approved):

PR #625, opened 2020-01-29
SSZ Rewrite

wemeetagain:
- SSZ rewritten using Type classes, SSZ operations attached as methods to classes.
eg: `serialize(Type, value)` is now `Type.serialize(value)`
- SSZ type classes include two (soon three) implementations
  - they can operate on "structural" data, ie: the data structures we've been using (`Object`s, `Array`s, etc)
  - they can operate on "tree" data, ie: a `Tree` from the `@chainsafe/persistent-merkle-tree` repo
```typescript
const structuralState: BeaconState = config.types.BeaconState.structural.defaultValue();
const treeState: BeaconState = config.types.BeaconState.tree.defaultValue();
config.types.BeaconState.equals(structuralState, treeState) === true;
```
  - the 'handlers' for each implementation are attached to the type under `Type.structural`, `Type.tree`, and `Type.byteArray`, Code available for these implementations available under `ssz/src/backings`
  - basic types have only a single implementation, code available under `ssz/src/types/basic`
- Non-structural data is wrapped in an ES6 `Proxy` to give a structural interface (getters/setters/iteration/ownKeys)
  - eg: 
```typescript
// treeState implements the BeaconState interface! (even though it is actually a Tree)
const treeState: BeaconState = config.types.BeaconState.tree.defaultValue();
treeState.slot = 5;
for (const validator of treeState.validators) {...}
```
- `ssz-type-schema` is gone, no longer used. Instead, the types are more tightly coupled to ssz operations. types can be found in `ssz/src/types`, and additionally in `ssz/src/backings/structural` and `ssz/src/backings/tree`.
- `eth2.0-types` have been updated to suit
  - Typescript types have changed, this significantly effects code downstream, eg: in the state transition, as the new types are not as expressive as the old types
    - `Array`s replaced with `Vector`, `List` (see `ssz/src/interface.ts` for definitions)
    - `BitVector`/`BitList` (from `@chainsafe/bit-utils`) replaced with `Vector<boolean>` and `List<boolean>` -- no more `bit-utils`, it is no longer used
    - `Bytes32`/`BytesX` replaced with `ByteVector`.
      - A `Uint8Array` can be returned by doing `valueOf()` on a `ByteVector`. Unfortunately, this must be explicitly cast to `Uint8Array` for now. eg: `bvector.valueOf() as Uint8Array`
    - `Uint8Array` is used directly (as opposed to `Buffer`) as input/output for hashing, etc.
    - lower case primitive names have been changed to uppercase for consistency and closer alignment with spec, eg: `bytes32` -> `Bytes32`

current benchmarks of various operations done on validator registry of 10k validators:
```
iterate - structural x 190,399 ops/sec ±0.51% (98 runs sampled)
iterate - tree x 74.05 ops/sec ±0.64% (67 runs sampled)

get validator - structural x 44,796,868 ops/sec ±1.63% (89 runs sampled)
get validator - tree x 469,510 ops/sec ±1.96% (84 runs sampled)

get validator property - structural x 39,335,530 ops/sec ±1.47% (89 runs sampled)
get validator property - tree x 241,673 ops/sec ±6.50% (74 runs sampled)

set validator - structural x 187,899 ops/sec ±5.51% (73 runs sampled)
set validator - tree x 9,461 ops/sec ±12.51% (74 runs sampled)

set validator property - structural x 20,798,474 ops/sec ±1.77% (88 runs sampled)
set validator property - tree x 108,128 ops/sec ±3.97% (82 runs sampled)

serialize - old x 18.07 ops/sec ±2.16% (48 runs sampled)
serialize - structural x 44.96 ops/sec ±0.61% (56 runs sampled)
serialize - tree x 14.96 ops/sec ±1.97% (38 runs sampled)

hashTreeRoot - old x 3.02 ops/sec ±5.00% (12 runs sampled)
hashTreeRoot - structural x 3.97 ops/sec ±3.06% (14 runs sampled)
hashTreeRoot - tree x 5,403 ops/sec ±1.36% (80 runs sampled)
```


packages/eth2.0-spec-test-util/src/single.ts
```diff
@@ -23,7 +23,7 @@ export interface ISpecTestOptions<TestCase, Result> {
    */
   inputTypes?: {[K in keyof NonNullable<TestCase>]?: InputType};
 
-  sszTypes?: {[K in keyof NonNullable<TestCase>]?: AnySSZType};
+  sszTypes?: {[K in keyof NonNullable<TestCase>]?: Type<any>};
```
mpetrunic, line 25: I would always opt-in for `unknown` instead of `any`

packages/eth2.0-spec-test-util/src/single.ts
```diff
@@ -69,7 +69,7 @@ export function describeDirectorySpecTest<TestCase, Result>(
   options: Partial<ISpecTestOptions<TestCase, Result>>
 ): void {
   // @ts-ignore
-  options = deepMerge(defaultOptions, options);
+  options = {...defaultOptions, ...options};
```
mpetrunic, line 71: ain't this gonna merge only first level?

P.S. We might not need ts-ignore (with proper deepMerge import)

packages/eth2.0-spec-test-util/src/transform.ts
```diff
@@ -1,29 +1,35 @@
...
+  if (type === byteType) {
+    return type;
+  } else if (type.isBasic()) {
+    if ((type as UintType<any>).byteLength) {
+      return new BigIntUintType({byteLength: (type as UintType<any>).byteLength});
+    } else {
       return type;
+    }
+  } else {
+    const props = Object.getOwnPropertyDescriptors(type) as any;
```
mpetrunic, line 19: as Type<unknown>

Just guessing, I could make better guess in  IDE. But those any types are dangerous

packages/eth2.0-state-transition/src/block/operations/attesterSlashing.ts
```diff
@@ -32,8 +32,8 @@ export function processAttesterSlashing(
   const attestation1 = attesterSlashing.attestation1;
   const attestation2 = attesterSlashing.attestation2;
   let slashedAny = false;
-  const attestingIndices1 = attestation1.attestingIndices;
-  const attestingIndices2 = attestation2.attestingIndices;
+  const attestingIndices1 = Array.from(attestation1.attestingIndices);
+  const attestingIndices2 = Array.from(attestation2.attestingIndices);
```
mpetrunic, line 36: Curious, why new array (considering it's performance sensitive  stuff :slightly_smiling_face: )?

packages/eth2.0-state-transition/src/block/operations/index.ts
```diff
@@ -67,7 +68,7 @@ export function processOperations(
     func,
     verifySignatures,
   }: {
-    operations: Operation[];
+    operations: List<Operation>;
```
mpetrunic, line 71: Is that necessary? Operation isn't ssz type

util functions?

packages/eth2.0-state-transition/src/block/randao.ts
```diff
@@ -26,11 +26,11 @@ export function processRandao(
...
     proposer.pubkey,
-    hashTreeRoot(config.types.Epoch, currentEpoch),
+    config.types.Epoch.hashTreeRoot(currentEpoch),
     body.randaoReveal,
     getDomain(config, state, DomainType.RANDAO),
   ));
   // Mix it in
   state.randaoMixes[currentEpoch % config.params.EPOCHS_PER_HISTORICAL_VECTOR] =
-    xor(getRandaoMix(config, state, currentEpoch), hash(body.randaoReveal));
+    xor(Buffer.from(getRandaoMix(config, state, currentEpoch)), Buffer.from(hash(body.randaoReveal)));
```
mpetrunic: `Buffer.from(getRandaoMix(config, state, currentEpoch).buffer)` will create just view instead off allocation new buffer.

Source: https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_arraybuffer_byteoffset_length

packages/eth2.0-state-transition/src/epoch/justification.ts
```diff
@@ -22,11 +22,15 @@ export function processJustificationAndFinalization(
...
+  const bits = state.justificationBits;
 
   // Process justifications
   state.previousJustifiedCheckpoint = state.currentJustifiedCheckpoint;
   // Rotate the justification bitfield up one epoch to make room for the current epoch
-  state.justificationBits.push(false);
+  for (let i = bits.length - 1; i >= 1; i--) {
+    bits[i] = bits[i-1];
+  }
+  bits[0] = false;
```
mpetrunic, line 33: Might be nice to add "unshift" method. We can open issue for that

packages/ssz/src/backings/backedValue.ts
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * A BackedValue is a value that is backed by a non-structural type
+ *
+ * It is implemented as an ES6 Proxy object that provides
+ * - convenient access to the structural properties corresponding to its type
+ * - additional methods for backing-specific implementations of ssz operations
+ */
+export type BackedValue<T extends object> = TreeBackedValue<T> | ByteArrayBackedValue<T>;
+
+export function isBackedValue<T extends object>(value: any): value is BackedValue<T> {
```
mpetrunic: value could be `unknown` and then you narrow it:

Article for reference:
https://mariusschulz.com/blog/the-unknown-type-in-typescript#narrowing-the-unknown-type

packages/ssz/src/backings/structural/array.ts
```diff
@@ -0,0 +1,180 @@
...
+export class CompositeArrayStructuralHandler<T extends ArrayLike<any>> extends StructuralHandler<T> {
+  _type: CompositeArrayType<T>;
+  getLength(value: T): number {
+    throw new Error("Not implemented");
+  }
+  size(value: T): number {
+    if (this._type.elementType.isVariableSize()) {
+      let s = 0;
+      for (let i = 0; i < this.getLength(value); i++) {
+        s += this._type.elementType.structural.size(value[i]) + 4;
```
mpetrunic, line 90: I see this "4" often, it could be some constant with meaningful name :smile:

mpetrunic (commented):
So much code haha

I'm liking the new syntax. I have few questions/comments:
- why are BitList/BitVector missing a lot of methods?
- How are you handling padding bit in BitList/BitVector since you are accessing byte array directly?
- You should use Buffer.from(arrayLike.buffer) to avoid allocating all those new buffers (https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_arraybuffer_byteoffset_length)
- We are also creating a lot of new arrays when you are wrapping?
- Could you try to ensure better type safety, like try to drop all "any" types?
- I'm a bit worried with array proxy, as from my experiment some methods and operations are avoiding proxy :worried:

packages/eth2.0-spec-test-util/src/single.ts
wemeetagain: deepMerge was causing problems with the new type instances, stripping methods.
I think we only need shallow copying here

packages/eth2.0-spec-test-util/src/transform.ts
wemeetagain: this is a terrible, terrible monkey-patch function.
property descriptors are described [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor#Description), unfortunately, its not `Type`, its a weird javascript thing, but Object.create takes property descriptors.
The intention with this function was to rewrite the new hierarchy of types, but with only BigInts.
Unfortunately, this still corrupts the existing types, so I think we can just run the ssz spec tests independently, and accept this little gross function.

packages/ssz/src/backings/structural/array.ts
wemeetagain: same with 32 and 256, :+1:

packages/eth2.0-state-transition/src/block/operations/index.ts
wemeetagain: Yeah, its necessary. If we accept that ssz objects can have different backings, we can only support the lowest-common-denominator interface between them.
This is what is currently defined in `ssz/src/interface.ts`.
The idea is we want to be able to seemlessly swap out a structural beacon state with a tree-backed beacon state.

packages/eth2.0-state-transition/src/block/operations/attesterSlashing.ts
wemeetagain: The state transition has been _minimally_ changed to support the ArrayLike/List/Vector interfaces.
In this case, we're converting a `List<ValidatorIndex>` to `ValidatorIndex[]`.
The plan is to go back through and optimize using a testbench once we're further along, using https://github.com/protolambda/eth2-docs and https://github.com/ethereum/eth2.0-specs/pull/1552 as starting points.

packages/eth2.0-spec-test-util/src/single.ts
wemeetagain: I'll try changing `any`s to `unknown`s

wemeetagain:
>     * why are BitList/BitVector missing a lot of methods?

So we're representing bitarrays differently than before, instead of using a custom bitarray class with nonstandard getters/setters (which mostly treated bitarrays as bytearrays), we're using `Vector<boolean>` and `List<boolean>`, which means that bitarrays can be treated like ArrayLikes of booleans. In the structural implementation, we're literally using `boolean[]`.

Unfortunately, that means that the custom methods that took advantage of the byte-representation (and convenience methods) are hard to implement _as methods_ across the board, since we're dealing with simple boolean arrays in one of the cases. These are fairly easy to recreate as utility functions (and will likely need to be added for readability).

>     * How are you handling padding bit in BitList/BitVector since you are accessing byte array directly?

Padding is now only around in the serialized form of BitLists. In the structural implementation, the BitList is a `boolean[]`. In the tree-backed implementation, the tree has leaves of 32 byte chunks (padded w/ zero chunks to the 'limit' of the list), with an additional level in the tree / mixin chunk keeping track of the length.

>     * You should use Buffer.from(arrayLike.buffer) to avoid allocating all those new buffers (https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_arraybuffer_byteoffset_length)

:+1: 

>     * We are also creating a lot of new arrays when you are wrapping?

Yeah, this is mostly so I could move forward with integration. It's in cases where the `Array` api is used heavily, and the quickest way forward is to just convert the `List`/`Vector` to an `Array`.

>     * Could you try to ensure better type safety, like try to drop all "any" types?

Will try out `unknown`s. From what I can tell, though, the `any`s are mostly internal to the ssz library, (mostly due to not knowing the subtypes of ArrayLikes and ObjectLikes) and no exposed typing uses `any`.

>     * I'm a bit worried with array proxy, as from my experiment some methods and operations are avoiding proxy worried

We don't have all methods, currently only the methods defined in `ssz/src/interface.ts`.
It's actually very tricky to get certain methods to work (from a typing perspective).
Eg: `map` has a different interface between `Array` and `TypedArray` (TypedArray map returns a TypedArray!). And since we're representing `ByteVector` as `Uint8Array` in the structural case, we're bumping against these inconsistencies.
The idea is to provide the biggest consistent API between the different backings that we can, so hopefully that means a few more can get added.

packages/eth2.0-spec-test-util/src/single.ts
mpetrunic: I think deepMerge has third param where you can correct his behavior on unknown types. Might be useful to have util function which we can just pass to deepmerge

packages/eth2.0-state-transition/src/block/operations/index.ts
mpetrunic: I get that, but this isn't part of state and Operation type is not ssz type

packages/eth2.0-types/src/ssz/interface.ts
```diff
@@ -1,87 +1,90 @@
...
-  "number64",
-  "uint64",
-  "uint256",
+  "Boolean",
+  "Bytes4",
+  "Bytes8",
+  "Bytes32",
+  "Bytes48",
+  "Bytes96",
+  "Uint16",
```
protolambda, line 87: What happened to the other uints?

packages/ssz/src/backings/tree/list.ts
```diff
@@ -0,0 +1,197 @@
...
+  getLength(target: Tree): number {
+    return number32Type.fromBytes(target.getRoot(BigInt(3)), 0);
+  }
+  setLength(target: Tree, length: number): void {
+    const chunk = new Uint8Array(32);
+    number32Type.toBytes(length, chunk, 0);
+    target.setRoot(BigInt(3), chunk);
+  }
+  fromBytes(data: Uint8Array, start: number, end: number): Tree {
+    const length = (end - start) / this._type.elementType.size();
```
protolambda, line 32: You may want to handle the case where `end - start` is not a multiple of element size.

packages/ssz/src/backings/tree/list.ts
```diff
@@ -0,0 +1,197 @@
...
+            data,
+            start + currentOffset,
+            start + nextOffset,
+          ),
+        );
+      }
+      this.setLength(target, offsets.length);
+    } else {
+      const elementSize = this._type.elementType.structural.size(null);
+      const length = (end - start) / elementSize;
```
protolambda, line 137: Ditto

packages/ssz/src/types/index.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * SSZ is a type system that defines:
```
protolambda, line 2: Nice docs :+1:

protolambda (dismissed):
Few minor comments. And remaining question; I see List<boolean> and Bitlist are being interchanged in some places, while not strictly the same (packing by byte vs packing bits). Same for vector. I might be missing some context in usage, just want to confirm what is happening.

Read through the complete PR diff, but couldn't find any big problems, other than the minor stuff mentioned above.

Impressed by the huge amount of work put into this, it looks great and I'm looking forward to data-sharing and hash-tree-root speed in the state, it will help performance a lot (together with memoization for transition functions) :clap:

packages/eth2.0-types/src/ssz/interface.ts
wemeetagain: We're currently only using those listed. Probably a good idea to declare them all just for completeness

wemeetagain:
:pray: thanks for the review

> Few minor comments. And remaining question; I see List and Bitlist are being interchanged in some places, while not strictly the same (packing by byte vs packing bits). Same for vector. I might be missing some context in usage, just want to confirm what is happening.

The `BitList` and `BitVector` types are declared as `List<boolean>` and `Vector<boolean>`, basically an array-ish of booleans. As opposed to being exposed as the packed bytearray, something that would be an array-ish of numbers. In the 'structural' implementation, we're using a concrete `boolean[]` directly. In that case, when we `serialize`/`hashTreeRoot` we create the tightly packed byte array (w padding bit in bitlist serialization). In the 'tree' implementation, the tightly packed chunks are stored in the tree, but we still expose this array-like of booleans, so getters return `boolean` and setters set a `boolean value.

But the special packing is the reason we need additional implementations of bitlist and bitvector vs just using the list and vector  impls.

wemeetagain:
@ChainSafe/eth2-0 ready for a final review!
:rocket:

packages/bls/src/publicKey.ts
```diff
@@ -49,16 +48,16 @@ export class PublicKey {
...
   }
 
-  public toBytesCompressed(): BLSPubkey {
+  public toBytesCompressed(): Buffer {
     return Buffer.from(this.value.serialize());
   }
 
   public toHexString(): string {
-    return `0x${this.toBytesCompressed().toString("hex")}`;
+    return `0x${Buffer.from(this.toBytesCompressed()).toString("hex")}`;
```
mpetrunic: Why another copy of buffer?

packages/bls/src/signature.ts
```diff
@@ -43,20 +42,20 @@ export class Signature {
...
     return publicKey.verifyMessage(this, message, domain);
   }
 
-  public verifyMultiple(publicKeys: PublicKey[], messages: bytes32[], domain: Domain): boolean {
+  public verifyMultiple(publicKeys: PublicKey[], messages: Uint8Array[], domain: Uint8Array): boolean {
     domain = padLeft(domain, 8);
     return this.value.verifyAggregatedHashWithDomain(
       publicKeys.map((key) => key.getValue()),
-      messages.map((message) => Buffer.concat([message, domain]))
+      messages.map((message) => Buffer.concat([Buffer.from(message as Uint8Array), domain as Uint8Array]))
```
mpetrunic: why those castings?
Can't you just put type on message argument?

packages/ssz/src/backings/structural/array.ts
mpetrunic: You forgot about this :disappointed:

mpetrunic (approved):

PR #626, opened 2020-01-30
Chain and validator improvements

mpetrunic:
This is first in series of PRs to fix current priority bugs preventing chain finalization.

Bugs solved:
- block processing queue doesn't process block if you submit it before previous one is processed
- validator won't propose or attest before epoch 1
- cli displays genesis time in user friendly format instead of format required by cli


packages/lodestar-cli/src/commands/dev.ts
```diff
@@ -136,8 +136,8 @@ export class DevCommand implements ICliCommand {
     if (options.genesisState) {
       state = quickStartOptionToState(config, tree, options.genesisState);
     } else if (options.genesisTime && options.validatorCount) {
-      logger.info(`Starting node with genesisTime ${new Date(parseInt(options.genesisTime)*1000)} and \
-       ${options.validatorCount} validators.`);
+      logger.info(`Starting node with genesisTime ${parseInt(options.genesisTime)} and `
```
wemeetagain: can we still print the int version of the genesis time?
like: ${new Date(...)} (${parseInt(...)})

packages/lodestar/src/chain/chain.ts
```diff
@@ -411,4 +373,28 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
     this.logger.info(`Initializing beacon chain with eth1 block ${eth1Block.hash}`);
     await this.initializeBeaconChain(genesisState, merkleTree);
   };
+
```
wemeetagain, line 401: Can you add a description/motivation of the algorithm/problem here?

wemeetagain (changes requested):
Looks much better :+1:

packages/lodestar-cli/src/commands/dev.ts
mpetrunic: You mean human readable version along with current int version?

packages/lodestar/src/chain/chain.ts
mpetrunic: sure

packages/lodestar-cli/src/commands/dev.ts
wemeetagain: yea

wemeetagain (approved):
LGTM

PR #628, opened 2020-01-30
Have Separate Instances of Lodestar CLI Use Separate DBs

### LINKED ISSUE #603: Database location mgmt
When running two `dev` cli commands in separate terminals, second `dev` command would partially delete log files of first command. If disabled, both will use same  database.

We need to generate unique directory for every `dev` command run (maybe use peerId as key) and delete database before exit.

aunyks:
Resolves #603.

### DIFF
--- a/packages/lodestar-cli/src/commands/dev.ts
+++ b/packages/lodestar-cli/src/commands/dev.ts
@@ -6,6 +6,7 @@ import {ICliCommand} from "./interface";
 import {CommanderStatic} from "commander";
 import deepmerge from "deepmerge";
 import fs, {mkdirSync} from "fs";
+import {join} from "path";
 import PeerId from "peer-id";
 import yaml from "js-yaml";
 import {config as mainnetConfig} from "@chainsafe/eth2.0-config/lib/presets/mainnet";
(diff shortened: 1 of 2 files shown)


packages/lodestar-cli/src/commands/dev.ts
```diff
@@ -114,20 +115,23 @@ export class DevCommand implements ICliCommand {
...
       peerId = PeerId.createFromHexString(options["peerId"]);
     } else if (options["peerIdFile"]) {
-      peerId = loadPeerId(options["peerId"]);
+      peerId = await loadPeerId(options["peerId"]);
     } else {
-      peerId = createPeerId();
+      peerId = await createPeerId();
+    }
+    if (!conf.db) {
+      conf.db = {name: join(".", "lodestar-db", peerId.toB58String())};
```
mpetrunic: This is great, but if everyone agrees, I would maybe propose convention to keep this stuff in `.tmp` directory which we would put in gitignore. This `.tmp` dir would contain all those databases(node, validators, keys etc) as my git status is getting clouded with all those stuff.

cc @wemeetagain @tuyennhv

twoeths: yeah I'm fine with that, let's put these stuff in `.tmp`

aunyks: Good idea!

mpetrunic (approved):

wemeetagain (approved):

PR #647, opened 2020-02-07
bls 0.10.1

mpetrunic:
- added new methods
- updated spec tests dep
- update bls spec tests to new test case format

TODO:
- ~~3 fast aggregate verify tests are failing~~


packages/bls/src/index.ts
```diff
@@ -78,18 +77,36 @@ export function aggregatePubkeys(publicKeys: BLSPubkey[]): BLSPubkey {
...
+export function verifyAggregate(publicKeys: BLSPubkey[], messageHash: bytes32, signature: BLSSignature): boolean {
+  assert(publicKeys, "publicKey is null or undefined");
+  assert(messageHash, "messageHash is null or undefined");
+  assert(signature, "signature is null or undefined");
+  try {
+    return Signature
+      .fromCompressedBytes(signature)
+      .verifyAggregate(publicKeys, messageHash);
   } catch (e) {
+    console.log(e)
```
mpetrunic: @wemeetagain Do you think it makes sense to introduce "debug" dependency for ssz and bls so we can debug when other people are using it?

mpetrunic:
@ChainSafe/eth2-0 we cannot pass CI until everything is updated to 0.10.x spec but we are passing bls stuff which should be enough to release it

packages/bls/src/index.ts
wemeetagain: possibly? Curious why this would throw in the first place?
Perhaps we can just bubble up errors

mpetrunic: Well sometimes invalid serialization, wrong order, there is bunch of things though. Something wrong with messages, length etc

wemeetagain: I feel like those things could trigger an error rather than a false.

mpetrunic: well invalid signature order should trigger false though

mpetrunic: Let's release like this. Invalid signatures in spec tests are breaking everything if we remove try/catch

wemeetagain (approved):

PR #648, opened 2020-02-11
Attestation collection

### LINKED ISSUE #533: Validator attestation aggregation 0.9.0
https://github.com/ethereum/eth2.0-specs/pull/1440

mpetrunic:
- add clock which emits new slot and epoch events (without slipping)
- add service which tracks at which slot needs to start collecting subnet attestation
- add api for validator to subscribe to committee subnet

resolves #533


packages/lodestar/src/sync/regular.ts
```diff
@@ -76,18 +82,9 @@ export class RegularSync {
...
-    const attestation = attestationSubnet.attestation;
-
-    // to see if we need special process for these unaggregated attestations
-    // not in the spec atm
-    // send attestation on to other modules
-    await Promise.all([
-      this.opPool.attestations.receive(attestation),
-      this.chain.receiveAttestation(attestation),
-    ]);
+  public receiveCommitteeAttestation = async (attestationSubnet: {attestation: Attestation; subnet: number}):
```
twoeths: Do we still need this as we have the AttestationCollector?

packages/lodestar/src/sync/subnet/attestation-collector.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  private checkDuties = (slot: Slot): void => {
+    const committees = this.aggregationDuties.get(slot) || new Set();
+    committees.forEach((committeeIndex) => {
+      this.network.gossip.subscribeToAttestationSubnet(
+        getCommitteeIndexSubnet(committeeIndex),
+        this.handleCommitteeAttestation
+      );
+      setTimeout(
```
twoeths, line 55: should we use 1 single timer after the for loop?

mpetrunic: We could if we pass array of subnets to unsubscribe from. But, most of the time, it will be 1 or max 2 to unsubscribe from so it's not really relevant, I guess

packages/lodestar/src/api/rest/routes/validator/subscribeToCommitteeSubnet.ts
```diff
@@ -0,0 +1,71 @@
...
+interface IBody extends DefaultBody {
+  committee_index: string;
+  slot: string;
+  slot_signature: string;
+  aggregator_pubkey: string;
+}
+
+export const registerSubscribeToCommitteeSubnet = (fastify: IFastifyServer, modules: IApiModules): void => {
+  fastify.post<DefaultQuery, DefaultParams, DefaultHeaders, IBody>(
+    "/beacon_committee_subscription",
```
twoeths, line 40: question: is this up to the client to implement? Because I cant find this in https://github.com/ethereum/eth2.0-APIs/blob/master/apis/validator/beacon-node-oapi.yaml

mpetrunic: You are looking at wrong one. This needs to be migrated here: https://ethereum.github.io/eth2.0-APIs/

Official 0.9.x validator spec is still not merged. Here is PR version: https://mpetrunic.github.io/eth2.0-APIs/

packages/lodestar/src/sync/regular.ts
mpetrunic: nope

packages/eth2.0-utils/src/math.ts
```diff
@@ -43,3 +43,18 @@ export function bigIntSqrt(n: bigint): bigint {
   }
   return x;
 }
+
+/**
+ * Renerates a random integer between min (included) and max (excluded).
```
wemeetagain: typo

packages/lodestar/src/chain/clock/local/LocalClock.ts
```diff
@@ -0,0 +1,80 @@
...
+    this.config = config;
+    this.genesisTime = genesisTime;
+    //this assumes clock time is trusted
+    this.currentSlot = getCurrentSlot(this.config, this.genesisTime);
+  }
+
+  public async start(): Promise<void> {
+    this.isRunning = true;
+    const diffTillNextSlot = this.getDiffTillNextSlot();
+    setTimeout(
```
wemeetagain: Can we keep track of the timerId here and below, so we can preemptively cancel on `stop`? small race condition where the timer will be called after stop.

packages/lodestar/src/chain/clock/local/LocalClock.ts
```diff
@@ -0,0 +1,80 @@
...
+    this.isRunning = false;
+  }
+
+  public getCurrentSlot(): number {
+    return this.currentSlot;
+  }
+
+  public onNewEpoch(cb: NewEpochCallback): void {
+    if(cb) {
+      this.newEpochCallbacks.push(cb);
```
wemeetagain: I don't think we should store callbacks here.
It makes it hard to unregister callbacks and it creates an additional self-made event queue (rather than just using the built-in event queue).
We should emit clock events and register event listeners inside the modules that need them

wemeetagain (changes requested):
Minor comments about the clock, the subscription logic all looks good :+1:

packages/lodestar/src/chain/clock/local/LocalClock.ts
```diff
@@ -0,0 +1,77 @@
...
+import {computeEpochAtSlot, getCurrentSlot} from "@chainsafe/eth2.0-state-transition";
+import {EventEmitter} from "events";
+
+export class LocalClock extends EventEmitter implements IBeaconClock {
+
+  private readonly config: IBeaconConfig;
+  private readonly genesisTime: number;
+  private currentSlot: number;
+  private isRunning: boolean;
+  private timeoutId: NodeJS.Timeout;
```
wemeetagain, line 12: this isn't used

wemeetagain (changes requested):

packages/lodestar/src/chain/clock/local/LocalClock.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.config = config;
+    this.genesisTime = genesisTime;
+    //this assumes clock time is trusted
+    this.currentSlot = getCurrentSlot(this.config, this.genesisTime);
+  }
+
+  public async start(): Promise<void> {
+    this.isRunning = true;
+    const diffTillNextSlot = this.getDiffTillNextSlot();
+    setTimeout(
```
wemeetagain: Can you set timeoutId here too xD

wemeetagain (changes requested):

packages/lodestar/src/chain/clock/local/LocalClock.ts
mpetrunic: :disappointed:

packages/lodestar/src/sync/subnet/attestation-collector.ts
```diff
@@ -0,0 +1,71 @@
...
+        this.unsubscribeSubnet,
```
twoeths, line 56: I'm thinking about a race condition when we may subscribe/unsubscribe the same subnet at the same time.

For example, at slot 2 we subscribe at subnet 10 then at the end of slot 2 we unsubscribe by `setTimeout`, but onNewSlot of slot 3 we want to subscribe at subnet 10 again. How can we make sure the earlier unsubscribe happens before this subscribe as it may come at the same time?

packages/lodestar/src/tasks/index.ts
```diff
@@ -23,30 +28,37 @@ export interface ITasksModules {
...
     this.db = modules.db;
     this.chain = modules.chain;
     this.logger = modules.logger;
+    this.sync = modules.sync;
+    this.network = modules.network;
   }
 
   public async start(): Promise<void> {
     this.chain.on("finalizedCheckpoint", this.handleFinalizedCheckpointChores);
+    this.sync.on("regularSyncStarted", this.handleRegularSyncStartedTasks);
```
twoeths, line 48: I can't find where is this event emitted from

packages/lodestar/src/tasks/index.ts
```diff
@@ -23,30 +28,37 @@ export interface ITasksModules {
...
     this.chain.removeListener("finalizedCheckpoint", this.handleFinalizedCheckpointChores);
+    this.sync.removeListener("regularSyncStarted", this.handleRegularSyncStartedTasks);
   }
 
   private handleFinalizedCheckpointChores = async (finalizedCheckpoint: Checkpoint): Promise<void> => {
     new ArchiveBlocksTask(this.config, {db: this.db, logger: this.logger}, finalizedCheckpoint).run();
   };
+  
+  private handleRegularSyncStartedTasks = async (): Promise<void> => {
+    new InteropSubnetsJoiningTask(this.config, {network: this.network});
```
twoeths: missing `.run()`?

packages/lodestar/src/sync/subnet/attestation-collector.ts
mpetrunic: Nope, this won't happen. I've changed logic of unsubscription. It will only unsubscribe from topic if there is 0 listeners.

packages/lodestar/src/tasks/tasks/interopSubnetsJoiningTask.ts
```diff
@@ -0,0 +1,58 @@
...
+  private readonly config: IBeaconConfig;
+  private readonly network: INetwork;
+
+  public constructor(config: IBeaconConfig, modules: IInteropSubnetsJoiningModules) {
+    this.config = config;
+    this.network = modules.network;
+  }
+
+  public async run(): Promise<void> {
+    for (let i = 0; i < this.config.params.RANDOM_SUBNETS_PER_VALIDATOR; i++) {
```
twoeths, line 22: can you show me `RANDOM_SUBNETS_PER_VALIDATOR` in the spec or any spec PR? I can't find it in https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/p2p-interface.md

mpetrunic: https://github.com/ethereum/eth2.0-specs/blob/v0.9.4/specs/validator/0_beacon-chain-validator.md#phase-0-attestation-subnet-stability

wemeetagain (dismissed):

wemeetagain:
conflict :/

twoeths (dismissed):

twoeths (approved):

PR #652, opened 2020-02-12
Enhance Best Justified Checkpoint Update Logic

### LINKED ISSUE #639: ensure best_justified_checkpoint updates to the best available
v0.9.3:  (ethereum/eth2.0-specs#1508)

aunyks:
Resolves #639


packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
```diff
@@ -197,7 +197,7 @@ export class StatefulDagLMDGHOST implements ILMDGHOST {
   public async start(genesisTime: number): Promise<void> {
     this.genesisTime = genesisTime;
     const numSlot = computeSlotsSinceEpochStart(this.config, getCurrentSlot(this.config, this.genesisTime));
-    const timeToWaitTillNextEpoch = (this.config.params.SLOTS_PER_EPOCH - numSlot) * 
+    const timeToWaitTillNextEpoch = (this.config.params.SLOTS_PER_EPOCH - numSlot) *
       this.config.params.SECONDS_PER_SLOT * 1000;
```
mpetrunic: This would probably benefit from clock I'm adding in https://github.com/ChainSafe/lodestar/pull/648/files#diff-c823e58d682201704d876febbedd0d52R4

aunyks: In that case, should I wait for that to be merged before merging this?

twoeths: yeah let's use the new clock in @mpetrunic PR, so please hold on a little bit.

packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
```diff
@@ -242,7 +242,14 @@ export class StatefulDagLMDGHOST implements ILMDGHOST {
       this.nodes[parentRoot].addChild(node);
     }
     if (justifiedCheckpoint && (!this.justified || justifiedCheckpoint.epoch > this.justified.epoch)) {
-      this.checkAndSetJustified(justifiedCheckpoint);
```
twoeths, line 245: the logic is right but probably we want to do everything inside `checkAndSetJustified`

aunyks: You're talking about the `if` statements surrounding that call to `checkAndSetJustified`, correct?

twoeths: yes. I mean we keep the `this.checkAndSetJustified(justifiedCheckpoint);` call and change the logic inside `checkAndSetJustified` because I think it's better to centralize the check and assignment of `bestJustifiedCheckpoint` in one place

twoeths: so sorry, I misunderstood you. What I mean is we keep both the `if (justifiedCheckpoint && (!this.justified || justifiedCheckpoint.epoch > this.justified.epoch)) {` and `this.checkAndSetJustified(justifiedCheckpoint);` and just do the requested change inside `checkAndSetJustified`

aunyks: I'm not sure I understand. `checkAndSetJustified` is already doing the change, no?

twoeths: yeah it does its job, just want to keep minimal change to the current code but probably this looks good enough too. I may need to change the flow a little bit in my other PR anyway.

twoeths (approved):

PR #654, opened 2020-02-14
Explicitly use GENESIS_SLOT

### LINKED ISSUE #638: Clean up handle of slots
v0.9.3: (ethereum/eth2.0-specs#1510)

aunyks:
Resolves #638.


mpetrunic (dismissed):

packages/lodestar/.__testdb/LOG
```diff
@@ -0,0 +1,3 @@
+2020/02/18-10:12:36.316468 70000e500000 Recovering log #31
+2020/02/18-10:12:36.318965 70000e500000 Delete type=3 #30
```
mpetrunic: remove testdb from git

mpetrunic (changes requested):

packages/lodestar/src/chain/attestation.ts
```diff
@@ -79,9 +84,14 @@ export class AttestationProcessor implements IAttestationProcessor {
     const target = attestation.data.target;
     assert([currentEpoch, previousEpoch].includes(target.epoch));
     assert(target.epoch === computeEpochAtSlot(this.config, attestation.data.slot));
+    assert(
+      getCurrentSlot(this.config, checkpointState.genesisTime) >= computeStartSlotAtEpoch(this.config, target.epoch)
+      , "egenesis time"
```
mpetrunic: typo

packages/lodestar/src/chain/attestation.ts
```diff
@@ -79,9 +84,14 @@ export class AttestationProcessor implements IAttestationProcessor {
...
+    );
     const block = await this.db.block.get(attestation.data.beaconBlockRoot);
-    assert(block.message.slot <= attestation.data.slot);
+    assert(block.message.slot <= attestation.data.slot, "blok slot");
```
mpetrunic: typo

packages/eth2.0-state-transition/src/util/slot.ts
```diff
@@ -3,12 +3,16 @@ import {number64, Slot, Epoch} from "@chainsafe/eth2.0-types";
 import {intDiv} from "@chainsafe/eth2.0-utils";
 import {computeStartSlotAtEpoch, computeEpochAtSlot} from ".";
 
-export function getCurrentSlot(config: IBeaconConfig, genesisTime: number64): Slot {
+export function getSlotsSinceGenesis(config: IBeaconConfig, genesisTime: number64): number64 {
```
mpetrunic: return type should be Slot

packages/lodestar/test/unit/chain/attestation.test.ts
```diff
@@ -89,28 +94,32 @@ describe("AttestationProcessor", function() {
...
+    const block = generateEmptySignedBlock();
+    dbStub.block.get.resolves(block);
+    const state = generateState();
+    dbStub.state.get.resolves(state);
+    forkChoiceStub.getJustified.returns({});
+    getAttestingIndicesStub.returns([0]);
+    state.balances = [];
+    state.validators = generateValidators(3, {})
+
+    await timeout(config.params.SECONDS_PER_SLOT)
```
mpetrunic: you already have method called sleep. That sait, why are we waiting here (we ain't suppose to do that in unit tests)?

you can mock time with sinon (checkout tests in attestation collection PR)

mpetrunic (changes requested):

packages/lodestar/test/unit/chain/attestation.test.ts
aunyks: I originally waited here because [this assertion](https://github.com/ChainSafe/lodestar/pull/654/files#diff-bcd99ae817459ea05e757f50adc3573aR94) would fail since the test's generated state's `genesisTime` wouldn't be `SECONDS_PER_SLOT` seconds earlier than the attestation's slot. To work around that, I just let that amount of time pass to satisfy the assertion.

aunyks: Now that you mention it, though, I can just subtract `SECONDS_PER_SLOT` seconds from the generated state's `genesisTime` and it should work.

aunyks:
Good catch on the typos lol I was using those for debugging. I'll replace them with more useful messages.

packages/lodestar-validator/src/util/beaconChain.ts
```diff
@@ -12,11 +12,15 @@ export enum DomainType {
   VOLUNTARY_EXIT = 4,
 }
 
-export function getCurrentSlot(config: IBeaconConfig, genesisTime: number64): Slot {
+export function getSlotsSinceGenesis(config: IBeaconConfig, genesisTime: number64): number64 {
```
mpetrunic: Return type should be slot here as well

packages/lodestar/src/chain/attestation.ts
```diff
@@ -79,9 +84,17 @@ export class AttestationProcessor implements IAttestationProcessor {
...
+      , "Current slot less than this target epoch's start slot"
+    );
     const block = await this.db.block.get(attestation.data.beaconBlockRoot);
-    assert(block.message.slot <= attestation.data.slot);
+    assert(block.message.slot <= attestation.data.slot, "Attestation is for past block");
 
+    assert(
+      getCurrentSlot(this.config, checkpointState.genesisTime) >= attestation.data.slot + 1,
+      "Attestation slot is not in the past"
+    );
```
mpetrunic: I don't think we should have this check as because of time diff some nodes might be ahead of us

mpetrunic: This check also created codeclimate issue :smile:

mpetrunic (changes requested):

packages/lodestar/src/chain/attestation.ts
aunyks: Didn't even think about that. Maybe we should bring that up in the spec. Pushing a fix right now.

mpetrunic (approved):

PR #656, opened 2020-02-17
newEpoch event

AyushKaul:
newEpoch event to clean up the blockPriorityQueue and keep a count of orphan blocks in the previous epoch.

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -333,6 +333,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
           this.db.chain.setFinalizedBlockRoot(finalizedBlockRoot),
         ]);
         this.emit("finalizedCheckpoint", newState.finalizedCheckpoint);
+        this.emit("newEpoch", computeEpochAtSlot(this.config, finalizedBlock.message.slot), this.blockProcessingQueue);
       }
       this.metrics.previousJustifiedEpoch.set(newState.previousJustifiedCheckpoint.epoch);
       this.metrics.currentJustifiedEpoch.set(newState.currentJustifiedCheckpoint.epoch);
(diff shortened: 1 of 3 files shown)


packages/lodestar/src/chain/chain.ts
```diff
@@ -333,6 +333,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
           this.db.chain.setFinalizedBlockRoot(finalizedBlockRoot),
         ]);
         this.emit("finalizedCheckpoint", newState.finalizedCheckpoint);
+        this.emit("newEpoch", computeEpochAtSlot(this.config, finalizedBlock.message.slot), this.blockProcessingQueue);
```
mpetrunic, line 336: This won't work if chain stops finalizing. I would suggest waiting for clock in https://github.com/ChainSafe/lodestar/pull/648/files#diff-7602fa529a155c161f8bf81ea5328d5bR6
to be merged as it will provide epoch listener

packages/lodestar/src/tasks/index.ts
```diff
@@ -39,14 +42,30 @@ export class TasksService implements IService {
...
 
   private handleFinalizedCheckpointChores = async (finalizedCheckpoint: Checkpoint): Promise<void> => {
     new ArchiveBlocksTask(this.config, {db: this.db, logger: this.logger}, finalizedCheckpoint).run();
   };
 
+  private blockProcessingQueueCleanUp = 
+  async (newEpoch: Epoch, blockProcessingQueue: FastPriorityQueue<IBlockProcessJob>): Promise<void> => {
+    return new Promise((resolve): void => {
+      while (!blockProcessingQueue.isEmpty()) {
+        const blockProcessJob = blockProcessingQueue.poll();
```
mpetrunic, line 61: poll method removes elements from queue which would also remove future blocks.
I suggest checking out `removeMany` method of https://www.npmjs.com/package/fastpriorityqueue

mpetrunic (changes requested):

wemeetagain:
Thanks for your contribution @AyushKaul 
Sorry for the very late reply / review.

Unfortunately, I don't think we'll need this block queue cleanup as is. @mpetrunic is currently replacing the current block queue (with polling) with a block pool that exposes an async iterator, has a different internal implementation.

I can tell you did a good bit of digging in our codebase to figure out this PR, we'd definitely welcome further contributions :) come join our discord chat https://discord.gg/yjyvFRP

wemeetagain:
Closing this PR, see above comment :)

PR #658, opened 2020-02-18
Fix libp2p unit test fail and suspend in CI

twoeths:
In CI environment, sometimes the libp2p unit test failed and unit test get suspended.
## Root Cause
+ It's too late to wait for `peer:connect`
## Fix
+ Use `Promise.all`

### DIFF
--- a/packages/lodestar/test/unit/network/nodejs/libp2p.test.ts
+++ b/packages/lodestar/test/unit/network/nodejs/libp2p.test.ts
@@ -13,54 +13,51 @@ describe("[network] nodejs libp2p", () => {
     assert.equal(node.isStarted(), false);
   });
   it.only("can connect/disconnect to a peer", async function ()  {
-    this.timeout(5000 * 100);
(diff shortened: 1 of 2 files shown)


packages/lodestar/test/unit/network/nodejs/libp2p.test.ts
```diff
@@ -24,15 +24,18 @@ describe("[network] nodejs libp2p", () => {
...
+    await Promise.all(
+      [
+        new Promise((resolve, reject) => {
+          const t = setTimeout(reject, 1000, 'connection timed out');
+          nodeB.once("peer:connect", () => {
+            clearTimeout(t);
+            resolve();
+          });
+        }),
+        nodeA.dial(nodeB.peerInfo),
```
mpetrunic, line 36: I wonder if just putting nodeA.dial after connect listener would solve issue as well

mpetrunic (approved):

PR #659, opened 2020-02-18
Fix mainnet spec tests

mpetrunic:


packages/eth2.0-state-transition/src/epoch/registryUpdates.ts
```diff
@@ -20,9 +20,9 @@ export function processRegistryUpdates(config: IBeaconConfig, state: BeaconState
   const currentEpoch = getCurrentEpoch(config, state);
   // Process activation eligibility and ejections
   const ejectionBalance = config.params.EJECTION_BALANCE;
-  state.validators.forEach((validator, index) => {
+  state.validators.forEach((validator, index, validators) => {
     if (isEligibleForActivationQueue(config, validator)) {
-      validator.activationEligibilityEpoch = currentEpoch + 1;
```
wemeetagain: why?

mpetrunic: stupid webstorm got confused and auto corrected I guess. I think it said that primitive value would be lost for some reason. I will revert tommorow

mpetrunic:
@ChainSafe/eth2-0 Can somebody else try this locally, all spec tests are passing locally for me

packages/eth2.0-state-transition/src/epoch/registryUpdates.ts
```diff
@@ -34,7 +34,7 @@ export function processRegistryUpdates(config: IBeaconConfig, state: BeaconState
   const activationQueue = state.validators.filter((validator) =>
     isEligibleForActivation(config, state, validator)
     // Order by the sequence of activation_eligibility_epoch setting and then index
-  ).sort((a, b) => (a.activationEligibilityEpoch - b.activationEligibilityEpoch) || 1);
+  ).sort((a, b) => (a.activationEligibilityEpoch - b.activationEligibilityEpoch) || -1);
```
wemeetagain: should this be changed?

mpetrunic: Yes according to docs, negative value makes a receive smaller index (moves to the front of array). This caused failed mainnet spec test on master.

mpetrunic: That said, I think we need validator index (or at least original index) because I doubt this is bubble sort where this would make sense

packages/eth2.0-params/src/presets/mainnet.ts
```diff
@@ -34,23 +34,23 @@ export const GENESIS_START_SHARD = 0;
...
+export const PERSISTENT_COMMITTEE_PERIOD = 2048; // epochs || 9 days
 
 export const MIN_EPOCHS_TO_INACTIVITY_PENALTY = 2 ** 2; // 25.6 minutes
 
 // State list lengths
 export const EPOCHS_PER_HISTORICAL_VECTOR = 2 ** 16;
 export const EPOCHS_PER_SLASHINGS_VECTOR = 2 ** 13;
 export const HISTORICAL_ROOTS_LIMIT = 2 ** 24;
-export const VALIDATOR_REGISTRY_LIMIT = 2 ** 40;
+export const VALIDATOR_REGISTRY_LIMIT = 1099511627776;
```
wemeetagain, line 53: Can we heep these params the way they were? I don't think this is much better (I think things like eg: `2 ** 40` are more readable).

wemeetagain (changes requested):

packages/eth2.0-params/src/presets/mainnet.ts
mpetrunic: It's hard to compare them with spec configs, it's probably best to keep same notation as spec file config. There is another reason for that, tersec (webpack minimizer) has a bug with those notation and bigint. For example  `2n**0n*BigInt(1e9)` will get minifed to `1*BigInt(1e9)` which will throw error in chrome. It's fixed in latest version but it's far from being included in webpack:
https://github.com/terser/terser/issues/546

wemeetagain: ok fair enough, maybe one day we can just ingest the upstream configuration files

wemeetagain (approved):

packages/eth2.0-params/src/presets/mainnet.ts
mpetrunic: Yeah we had issue, to automatically generate params from: https://github.com/ethereum/eth2.0-specs/blob/dev/configs/mainnet.yaml but it got closed by stalebot :(

PR #661, opened 2020-02-19
De-isolate Isolated Unit Test

aunyks:
An `it.only` block was accidentally merged into master. This changes it back to `it` so all Lodestar unit tests can be ran.


wemeetagain (approved):

aunyks:
I'm confused, because I can't see *how* the two branches are conflicting anywhere so I can't resolve the conflicts.

wemeetagain:
It looks like the change is already in master

PR #663, opened 2020-02-20
Update bls and ssz

wemeetagain:
Remove bls and ssz from the monorepo, use the latest versions


mpetrunic (changes requested):
Remove them from root readme?

mpetrunic (approved):

PR #665, opened 2020-02-24
Fix test compilation errors for remaining packages

### LINKED ISSUE #518: Fix lodestar unit test compilation errors
Currently our unit tests in lodestar are excluded from compilation (that's why IDEs are complaining) because it has too many compilation errors (after switching strict flag).

This issue includes following tasks:
- include test directory in tsconfig
- fix all compilation errors in tests
- create tsconfig.build.json
   - this should extend default tsconfig but exclude test directory (rootDir: src) so we don't compile test files into release build
- fix package.json to use this tsconfig.build.json for generating production builds
- apply this to other packages

twoeths:
## Goal
+ Close #518 
+ This is a continuation of #664


wemeetagain (dismissed):
LGTM

wemeetagain (approved):

PR #671, opened 2020-02-26
Snappy compression

### LINKED ISSUE #670: Add snappy compression

mpetrunic:
- added ssz and snappy encoders
- refactored Req/Resp to use new ReqRespEncoder
- fixed async iterable types

resolves #670


packages/lodestar/src/network/encoders/interface.ts
```diff
@@ -0,0 +1,6 @@
+import {BasicType, ContainerType} from "@chainsafe/ssz";
+
+export interface IReqRespEncoder {
+  encode(type: ContainerType<unknown>|BasicType<unknown>, data: unknown): unknown;
```
wemeetagain: Can you use `Type` here and elsewhere instead of ContainerType|BasicType? We may need to use `any` instead of `unknown`.

packages/lodestar/src/network/encoder.ts
```diff
@@ -0,0 +1,131 @@
...
+import {ERR_INVALID_REQ, MAX_CHUNK_SIZE, Method, ReqRespEncoding} from "../constants";
+import {RequestBody, ResponseBody} from "@chainsafe/lodestar-types";
+import {IReqRespEncoder} from "./encoders/interface";
+import {getRequestMethodSSZType, getResponseMethodSSZType} from "./util";
+import {SszEncoder} from "./encoders/ssz";
+import {SnappyEncoder} from "./encoders/snappy";
+import * as varint from "varint";
+import {Response, ResponseChunk} from "./interface";
+// eslint-disable-next-line import/no-extraneous-dependencies
+import BufferList from "bl";
```
wemeetagain, line 9: Why are we using BufferList?

packages/lodestar/src/network/util.ts
```diff
@@ -73,58 +65,30 @@ export function encodeResponseChunk(config: IBeaconConfig, method: Method, {err,
...
-  const bytes = varint.decode.bytes;
-  if (
-    length !== chunk.length - (bytes + 1) ||
-    length > MAX_CHUNK_SIZE
-  ) {
-    throw new Error(ERR_INVALID_REQ);
-  }
-  const data = chunk.slice(bytes + 1);
-  let output: Status | Goodbye | SignedBeaconBlock;
+export function getResponseMethodSSZType(
```
wemeetagain, line 77: Can we make getRequestMethodSSZType and getResponseMethodSSZType a single function?
eg: getMethodSSZType

packages/lodestar/src/network/encoder.ts
```diff
@@ -0,0 +1,131 @@
...
+    //decoding is done backwards
+    const encoders = this.getEncoders(method, this.encoding).reverse();
+    return encoders.reduce((result: unknown, encoder) => {
+      return encoder.decode(type, result);
+    }, data) as RequestBody;
+  }
+
+  public encodeResponse(
+    method: Method
+  ): ((source: AsyncIterable<ResponseChunk|Partial<Response>>) => AsyncGenerator<Buffer>) {
```
wemeetagain: Why are we iterating over `Partial<Response>`? I thought responses are always `Arrayish<ResponseChunk>`?

packages/lodestar/src/network/interface.ts
```diff
@@ -12,12 +12,13 @@ import {
...
 } from "@chainsafe/lodestar-types";
-
 import {Method, RequestId} from "../constants";
 import StrictEventEmitter from "strict-event-emitter-types";
 import {IGossip} from "./gossip/interface";
 
-export type ResponseCallbackFn = (response: {err?: Error; output?: ResponseBody}) => void;
+export type Response = {err?: Error; output: ResponseBody};
+
+export type ResponseCallbackFn = ((response: Partial<Response>) => void);
```
wemeetagain: Should't ResponseCallbackFn take a ResponseChunk?
It looks like responses are not yet given to the resp/req manager in a chunked format yet, but given as a full response body.

packages/lodestar/src/network/reqResp.ts
```diff
@@ -127,46 +113,37 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
...
-  ): AsyncIterator<Buffer> {
-    const config = this.config;
-    return (async function * () {
-      const sourceVal = await source;
-      for await (const val of sourceVal) {
-        // each yield result will be sent to stream.sink immediately
-        yield encodeResponseChunk(config, method, val);
-      }
-    })();
+  private handleRpcRequests(
```
wemeetagain, line 118: should be renamed `handleRpcRequest` since this is handling a single request.

packages/lodestar/src/network/reqResp.ts
```diff
@@ -175,70 +152,26 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
...
+        this.encoder.decodeResponse(method),
+        async (source: AsyncIterable<ResponseBody>) => {
           try {
-            const responses = [];
-            for await (const val of source) {
-              const data = Buffer.isBuffer(val) ? val : val.slice();
-              const response = decodeResponseChunk(this.config, method, data);
-              responses.push(response.output);
+            const  responses = [];
+            for await (const response of source) {
```
wemeetagain, line 172: there should only ever be one response, no?

wemeetagain (changes requested):

packages/lodestar/src/network/encoder.ts
mpetrunic: Because that is actual type returned from source. Check: https://github.com/jacobheun/it-handshake#handshakeduplex
We were using source.slice() which is method on BufferList to return Buffer.

mpetrunic: I've added @types/bl (not bl itself) as dev dependency

packages/lodestar/src/network/util.ts
mpetrunic: like add extra param "isResponse"?

packages/lodestar/src/network/encoder.ts
mpetrunic: When we handle request we are yielding each ResponseChunk, but sometimes, we yield error or empty Response.
If we hook up stream from database we could stream blocks directly to response

packages/lodestar/src/network/interface.ts
mpetrunic: I will change this in the morning, I don't think we need Response, BeaconBlocksByRangeResponse and BeaconBlocksByRootResponse

packages/lodestar/src/network/reqResp.ts
mpetrunic: It's handling multiple same method requests from same peer.Like if peer sends two BlocksByRange requests with different request data, this pipe will handle them both

packages/lodestar/src/network/reqResp.ts
mpetrunic: In case of BlocksByRange, source will collect all decoded blocks

packages/lodestar/src/network/util.ts
wemeetagain: I don't think we need to differentiate between request and response ssz types.
just smoosh the two `switch` statements together

mpetrunic: But we do need. BeaconBlocksByRangeRequest is reuqest type while SignedBeaconBlock is response ssz type

mpetrunic: they are completely different types

wemeetagain: Oh I see, yeah, ignore this comment xD

packages/lodestar/src/network/interface.ts
wemeetagain: and instead of `Partial<ResponseChunk>` we probably should make `output` in `ResponseChunk` optional

packages/lodestar/src/network/reqResp.ts
wemeetagain: I think this is not the intended behavior described in the spec.
> We use ONE stream PER request/response interaction. Streams are closed when the interaction finishes, whether in success or in error.

mpetrunic: I know, but that is a bit more complicated and I was planning on creatimg a separate issues for those stuff

packages/lodestar/src/network/reqResp.ts
wemeetagain: Ok, so we're still exposing a ResponseBody (unchunked response) :(

mpetrunic: naming is wrong, ResponseChunk={err, output:ResponseBody}

packages/lodestar/src/network/reqResp.ts
mpetrunic: #675 #676 , #677

twoeths: my understanding is we still use ONE stream per request/response interaction with this implementation. Every time client sends request, it dials and then a new stream returned (connection is the same). At the server side, stream is created and closed automatically per request as I don't see any example to close stream in libp2p api.

If we send 2 BlocksByRange request, it'll dial 2 times and 2 separated streams returned (connection is the same) and closed automatically once done. This same method are used across requests but each time its a separate stream.

am I missing something?

mpetrunic: https://github.com/libp2p/js-libp2p-mplex

wemeetagain: I think @tuyennhv is right. Even if we're using mplex or some other stream multiplexer, each dial should return a new stream.

mpetrunic: image

packages/lodestar/src/network/reqResp.ts
```diff
@@ -117,56 +105,44 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
...
+    peerId: PeerId, method: Method
+  ): ((source: AsyncIterable<Buffer|BufferList>) => AsyncGenerator<ResponseChunk>) {
+    const getResponse = this.getResponse;
+    return (source: AsyncIterable<Buffer|BufferList>) => {
+      return (async function * () {
+        for await (const val of source) {
+          const data = Buffer.isBuffer(val) ? val : val.slice();
+          const response = await getResponse(peerId, method, data);
+          if(response.length) {
+            yield* response;
```
twoeths, line 128: This `yield*` is nice :)

mpetrunic: you learn those tricks if you use react with saga :smile:

packages/lodestar/src/network/encoder.ts
```diff
@@ -0,0 +1,129 @@
...
+            status = chunk.err.message === ERR_INVALID_REQ ? 1 : 2;
+            //TODO: stop yielding
+          }
+          yield writeStatus(writeLP(encodedPayload), status);
+        }
+      })();
+    };
+  }
+
+  public decodeResponse(method: Method): ((source: AsyncIterable<Buffer|BufferList>) => AsyncGenerator<ResponseBody>) {
```
twoeths, line 68: this method may throw error at one of the below calls. Probably we want to handle try/catch at the caller side (`sendRequest`) to reject the promise?

mpetrunic: Not really. according to spec we need to reset stream on responder side (caller won't see exception). I was hopping that exception in async iterator will reset stream. Not sure what else we can so. If we catch exception as responder, we will process previous chunks.

mpetrunic:
@ChainSafe/eth2-0 Can we merge this and try to fix streams in separate PR?

twoeths (approved):

wemeetagain (approved):

PR #678, opened 2020-02-28
Fork choice: make sure head() have correct justified/finalized epoch like the store

### LINKED ISSUE #637: Filter out non-viable branches when call to get_head
v0.9.3: filter out non-viable branches when call to get_head (ethereum/eth2.0-specs#1495)

twoeths:
## Goal
+ Close #637 
## Approach
+ Maintain justified & finalized epochs in our "store"
+ Maintain justified & finalized epochs in our node
+ Change weighing system taking justified & finalized epochs of best target into account
+ When adding block, if it causes any best targets to be epoch conflict, we do "soft unlink" by setting `bestChild=null`, this causes those best targets unreachable from the justified root
+ When adding block, relink if those best target's epochs now correct (and previously it's not)

### DIFF
--- a/packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
+++ b/packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
@@ -14,6 +14,7 @@ import {ILMDGHOST} from "../interface";
 import {AttestationAggregator} from "./attestationAggregator";
 import {sleep} from "../../../util/sleep";
 import {RootHex, NodeInfo, HexCheckpoint} from "./interface";
+import {GENESIS_EPOCH} from "../../../constants";
 
 
 /**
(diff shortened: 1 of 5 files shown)


packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
```diff
@@ -44,15 +44,22 @@ class Node {
...
+  public finalizedEpoch: Epoch;
+
   /**
    * All direct children
    */
   public children: Record<RootHex, Node>;
 
-  public constructor({slot, blockRoot, parent}: {slot: Slot; blockRoot: RootHex; parent: Node}) {
+  public constructor({slot, blockRoot, parent, justifiedEpoch, finalizedEpoch}:
+  {slot: Slot; blockRoot: RootHex; parent: Node; justifiedEpoch: Epoch; finalizedEpoch: Epoch}) {
```
mpetrunic: It's getting too big, might  be nice to extract argument type to interface

packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
```diff
@@ -219,36 +243,64 @@ export class StatefulDagLMDGHOST implements ILMDGHOST {
...
+      finalizedEpoch,
     });
     // best target is the node itself
     node.bestTarget = node;
     this.nodes[blockRoot] = node;
 
     // if parent root exists, link to blockRoot
     if (this.nodes[parentRoot]) {
-      this.nodes[parentRoot].addChild(node);
+      this.nodes[parentRoot].addChild(node, justifiedEpoch, finalizedEpoch);
```
wemeetagain: I think we're supposed to be using the store's justified / finalized information, eg: `this.justified` and `this.finalized`.
Also, it would be good to check against the whole checkpoint (epoch + root), rather than just the epoch.

twoeths: we want to track the added block's checkpoints in the node, not the store's checkpoints. Later on we can compare block's checkpoints against the current store's checkpoints. This is similar to `head_state = store.block_states[block_root]` in the spec.

yeah I'll make the check against `epoch + root` instead of just `epoch`.

twoeths: ah you're right, should put in store's justified & finalzied checkpoint to compare ..

wemeetagain (approved):

PR #679, opened 2020-03-04
Handle network req/resp timeout

### LINKED ISSUE #676: Increase timeout after successful response chunk.
> The requester MUST wait a maximum of TTFB_TIMEOUT for the first response byte to arrive (time to first byte—or TTFB—timeout). On that happening, the requester allows a further RESP_TIMEOUT for each subsequent response_chunk received. For responses consisting of potentially many response_chunks (an SSZ-list) the requester SHOULD read from the stream until either; a) An error result is received in one of the chunks, b) The responder closes the stream, c) More than MAX_CHUNK_SIZE bytes have been read for a single response_chunk payload or d) More than the maximum number of requested chunks are read.

I would propose adding another transform function to pipe reading responses (after response decoding).

twoeths:
## Goal
+ Close #676

### DIFF
--- a/packages/lodestar/src/network/reqResp.ts
+++ b/packages/lodestar/src/network/reqResp.ts
@@ -162,19 +162,15 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements 
     return await new Promise((resolve, reject) => {
       this.logger.verbose(`send ${method} request to ${peerInfo.id.toB58String()}`);
       let responseTimer = setTimeout(() => reject(new Error(ERR_RESP_TIMEOUT)), TTFB_TIMEOUT);
+      const renewTimer = (): void => {
+        clearTimeout(responseTimer);
+        responseTimer = setTimeout(() => reject(new Error(ERR_RESP_TIMEOUT)), RESP_TIMEOUT);
+      };
       pipe(
         [this.encoder.encodeRequest(method, body)],
         stream,
         this.encoder.decodeResponse(method),
-        function handleTimeout(source: AsyncIterable<ResponseBody>): AsyncIterable<ResponseBody> {
-          return (async function * () {
-            for await (const resp of source) {
-              clearTimeout(responseTimer);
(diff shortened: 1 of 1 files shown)


packages/lodestar/src/network/reqResp.ts
```diff
@@ -161,11 +161,20 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
     const {stream} = await this.libp2p.dialProtocol(peerInfo, protocol) as {stream: Stream};
     return await new Promise((resolve, reject) => {
       this.logger.verbose(`send ${method} request to ${peerInfo.id.toB58String()}`);
-      const responseTimer = setTimeout(() => reject(new Error(ERR_RESP_TIMEOUT)), RESP_TIMEOUT);
+      let responseTimer = setTimeout(() => reject(new Error(ERR_RESP_TIMEOUT)), TTFB_TIMEOUT);
       pipe(
         [this.encoder.encodeRequest(method, body)],
         stream,
         this.encoder.decodeResponse(method),
+        function handleTimeout(source: AsyncIterable<ResponseBody>): AsyncIterable<ResponseBody> {
```
mpetrunic: can you extract this as separate method instead.

mpetrunic (approved):

mpetrunic (changes requested):
missclicked approval

packages/lodestar/src/network/reqResp.ts
wemeetagain: One thing to keep in mind (if you extract this function)
I think we will soon want to change the function signature of `sendRequest` to expose the async iterable (vs the arrayish promise).

twoeths: yeah we may want that when there's a need for `initial sync`, i.e. `BeaconBlockByRange` request. There are maybe 2 different `sendRequest` versions, 1 returns an Array, other returns an Async Iterable.

From what I see, the current `sendRequest` interface works fine with `GoodBye`, `Status` and `BeaconBlockByRoot` (since most of the time we only request a single root/block atm).

mpetrunic (approved):

wemeetagain (approved):

PR #680, opened 2020-03-05
Introduce RpcError

### LINKED ISSUE #677: Introduce Rpc specfic errors
We should create Error subset (`class RpcError extends Error`) which should have `status: number` field.
Req/Resp should send this status field instead of determining status from error message.

twoeths:
## Goal
+ Close #677

### DIFF
--- a/packages/lodestar/src/constants/network.ts
+++ b/packages/lodestar/src/constants/network.ts
@@ -19,10 +19,13 @@ export enum ReqRespEncoding {
   SSZ_SNAPPY = "ssz_snappy",
 }
 
-export const ERR_INVALID_REQ = "invalid request";
-export const ERR_RESP_TIMEOUT = "response timeout";
+export enum RpcErrorCode {
+  ERR_INVALID_REQ = 1,
+  ERR_RESP_TIMEOUT = 2,
+}
+
 export const GOSSIP_MAX_SIZE = 2**20;
 export const MAX_CHUNK_SIZE = 2**20;
-export const TTFB_TIMEOUT = 5 * 1000; // 10 sec
+export const TTFB_TIMEOUT = 5 * 1000; // 5 sec
 export const RESP_TIMEOUT = 10 * 1000; // 10 sec
 
--- a/packages/lodestar/src/network/encoder.ts
+++ b/packages/lodestar/src/network/encoder.ts
@@ -1,5 +1,5 @@
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
-import {ERR_INVALID_REQ, MAX_CHUNK_SIZE, Method, ReqRespEncoding} from "../constants";
+import {MAX_CHUNK_SIZE, Method, ReqRespEncoding, RpcErrorCode} from "../constants";
 import {RequestBody, ResponseBody} from "@chainsafe/lodestar-types";
 import {IReqRespEncoder, SnappyEncoder, SszEncoder} from "./encoders";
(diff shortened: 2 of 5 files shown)


packages/lodestar/src/network/encoder.ts
```diff
@@ -56,7 +57,7 @@ export class ReqRespEncoder {
           }
           let status = 0;
           if(chunk.err) {
-            status = chunk.err.message === ERR_INVALID_REQ ? 1 : 2;
+            status = chunk.err.status === ERR_INVALID_REQ ? 1 : 2;
```
mpetrunic: do we need this anymore?

Also add break statement so we stop yielding chunks

packages/lodestar/src/network/error.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Error of network req/resp
+ */
+
+export class RpcError extends Error {
+  public status: number;
+  constructor(message: string, status?: number) {
```
mpetrunic: status could be enum

packages/lodestar/src/network/encoder.ts
```diff
@@ -55,12 +55,12 @@ export class ReqRespEncoder {
               return encoder.encode(type, result);
             }, chunk.output) as Buffer;
           }
-          let status = 0;
-          if(chunk.err) {
-            status = chunk.err.status === ERR_INVALID_REQ ? 1 : 2;
-            //TODO: stop yielding
-          }
+          const status = chunk.err && chunk.err.status;
```
mpetrunic, line 58: ain't this gonna be true/false now?

twoeths: no, it's gonna be `undefined` or  `typeof chunk.err.status` which is number. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators

packages/lodestar/src/constants/network.ts
```diff
@@ -19,10 +19,13 @@ export enum ReqRespEncoding {
   SSZ_SNAPPY = "ssz_snappy",
 }
 
-export const ERR_INVALID_REQ = "invalid request";
-export const ERR_RESP_TIMEOUT = "response timeout";
+export enum RpcErrorEnum {
```
wemeetagain: Can we call this `RpcErrorType` or `RpcErrorCode`?

packages/lodestar/src/network/error.ts
```diff
@@ -0,0 +1,17 @@
+import {RpcErrorEnum} from "../constants";
+
+/**
+ * Error of network req/resp
+ */
+
+export class RpcError extends Error {
+  public status: RpcErrorEnum;
+  constructor(message: string, status?: RpcErrorEnum) {
```
wemeetagain: I think an `RpcError` will always require a status and only optionally include a message.
probably better to have the signature be `constructor(status: RpcErrorType, message?: string)`
(then we don't need the `withStatus` static method)

wemeetagain (changes requested):
Looks good, only minor comments

packages/lodestar/src/constants/network.ts
wemeetagain: bump

wemeetagain (changes requested):
LGTM other than minor renaming

twoeths:
Ready for review

wemeetagain (approved):
:100:

PR #683, opened 2020-03-10
Fix vulnerability in objectToCamelCase

wemeetagain:
Don't let objectToCamelCase overwrite an existing property

### DIFF
--- a/packages/lodestar-utils/src/misc.ts
+++ b/packages/lodestar-utils/src/misc.ts
@@ -5,6 +5,10 @@ export function objectToCamelCase(obj: object): object {
   if (Object(obj) === obj) {
     Object.getOwnPropertyNames(obj).forEach((name) => {
       const newName = camelcase(name);
+      // @ts-ignore
+      if (obj[newName]) {
+        throw new Error(`object already has a ${newName} property`);
+      }
       if (newName !== name) {
         // @ts-ignore
         obj[newName] = obj[name];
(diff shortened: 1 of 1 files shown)


packages/lodestar-utils/src/misc.ts
```diff
@@ -5,6 +5,10 @@ export function objectToCamelCase(obj: object): object {
   if (Object(obj) === obj) {
     Object.getOwnPropertyNames(obj).forEach((name) => {
       const newName = camelcase(name);
+      // @ts-ignore
+      if (obj[newName]) {
```
aunyks: I don’t believe this will correctly check if the property already exists. I think we should have `obj.hasOwnProperty(newName)` instead.

I say this because with this case, if `obj[newName]` is falsey but still already exists, this error won’t get thrown.

aunyks (changes requested):

packages/lodestar-utils/src/misc.ts
wemeetagain: good call

mpetrunic (dismissed):
Do we even need this method. Can't we just use camelcase lib inside ssz.toJson instead of doing it here?

wemeetagain:
> Do we even need this method. Can't we just use camelcase lib inside ssz.toJson instead of doing it here?

Good question, this was just a specific issue that the auditors found. I don't think the way we're using this function is currently a problem, but this is just to say that we've "addressed" the audit issues. 

Definitely would love to remove this function if we can.

aunyks (approved):

PR #684, opened 2020-03-12
Stream API for DB and Network req/resp

### LINKED ISSUE #682: Expose async iterable resources
Currently, we return `Promise<XXX[]>` in network and db apis.

We should instead return an `AsyncIterable<XXX[]>` so that consumers may begin processing items immediately, rather than awaiting all items to be fetched/returned.

twoeths:
## Goal
+ Close #682

### DIFF
--- a/packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
+++ b/packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
@@ -34,20 +34,37 @@ export class BlockArchiveRepository extends BulkRepository<SignedBeaconBlock> {
     upperLimit: number | null,
     step: number | null = 1
   ): Promise<SignedBeaconBlock[]> {
+    const result = [];
+    for await (const signedBlock of this.getAllBetweenStream(lowerLimit, upperLimit, step)) {
+      result.push(signedBlock);
+    }
+    return result;
+  }
+
+  public getAllBetweenStream(
+    lowerLimit: number | null,
+    upperLimit: number | null,
+    step: number | null = 1
+  ): AsyncIterable<SignedBeaconBlock> {
     const safeLowerLimit = lowerLimit || Buffer.alloc(0);
     const safeUpperLimit = upperLimit || Number.MAX_SAFE_INTEGER;
-    const data = await this.db.search({
+    const data = this.db.searchStream({
       gt: encodeKey(this.bucket, safeLowerLimit),
       lt: encodeKey(this.bucket, safeUpperLimit),
     });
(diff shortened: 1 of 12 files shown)


packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
```diff
@@ -34,20 +34,37 @@ export class BlockArchiveRepository extends BulkRepository<SignedBeaconBlock> {
...
-          return signedBlock.message.slot >= safeLowerLimit && (signedBlock.message.slot - safeLowerLimit) % step === 0;
-        } else {
-          return true;
+    const isEligible = (signedBlock: SignedBeaconBlock): boolean => {
+      if (step !== null && typeof safeLowerLimit === "number") {
+        return signedBlock.message.slot >= safeLowerLimit && (signedBlock.message.slot - safeLowerLimit) % step === 0;
+      } else {
+        return true;
+      }
+    };
```
mpetrunic: I would consider extracting this method to some other file so you can test it easily

packages/lodestar/src/db/controller/impl/level.ts
```diff
@@ -79,19 +80,26 @@ export class LevelDbController extends EventEmitter implements IDatabaseControll
...
-      });
+  public async search(opts: ISearchOptions): Promise<any> {
+    const searchData: any[] = [];
+    for await (const data of this.searchStream(opts)) {
+      searchData.push(data);
+    }
+    return searchData;
+  }
+
+  public searchStream(opts: ISearchOptions): AsyncIterable<any> {
```
mpetrunic, line 91: does it work if you put unknow as a type. Isn't result always gonna be buffer|string?

mpetrunic (commented):
Looks good

packages/lodestar/src/db/controller/impl/level.ts
twoeths: The whole DB controller uses `any` so it's hard to tell, I think this is something we need to improve as I forgot to deserialize and it compiled successfully.

Will create a separate issue to avoid `any` and work on that if you think so.

mpetrunic: Yeah any in controller is my bad, can't remember why I was doing that at the time

twoeths: Ok I created https://github.com/ChainSafe/lodestar/issues/685

wemeetagain (approved):
LGTM :)

PR #689, opened 2020-03-16
Regular sync

mpetrunic:
Status:
- a lot easier to start and connect two beacon nodes via shared state(check docs)
- nodes can sometimes catch up each other
- critical bugs fixed

It's extremely unstable, but it's something we can work from. I'll open issues from problems I've encountered so we can all work on fixing them.


packages/lodestar/README.md
```diff
@@ -32,13 +27,26 @@ Open resulting html file in favorite browser.
...
+
+#### Connect two beacon nodes
+
+`cd packages/lodestar-cli`
+
+Terminal 1:
+`bin/lodestar dev -r` - It will run beacon node and 8 validators with all the default settings. State will be written to .tmp/state.ssz
+
+Terminal 2:
+`bin/lodestar dev --bootnodes "/ip4/127.0.0.1/tcp30306" --multiaddrs "/ip4/127.0.0.1/tcp/30307" -v 0 -r` - Will connect to bootnode (node 1 default multiaddrs) and won't start validators.
```
wemeetagain: `tcp/30306`

packages/lodestar/src/api/rest/routes/validator/publishBlock.ts
```diff
@@ -31,11 +31,13 @@ export const registerBlockPublishEndpoint = (fastify: IFastifyServer, modules: I
...
-      await modules.chain.receiveBlock(
-        modules.config.types.SignedBeaconBlock.fromJson(
-          request.body.beacon_block
-        )
+      const block = modules.config.types.SignedBeaconBlock.fromJson(
+        request.body.beacon_block
       );
+      await Promise.all([
+        modules.chain.receiveBlock(block),
+        modules.network.gossip.publishBlock(block)
```
wemeetagain, line 39: I'm thinking we should wait for the chain to process the block before publishing via gossip? (And let the "sync" module relay the block to the network)

packages/lodestar/src/sync/utils/blocks.ts
```diff
@@ -24,11 +25,16 @@ export async function getBlockRange(
...
           return null;
         } catch (e) {
+          console.log(e);
           //if failed to obtain blocks, try in next round on another peer
           return chunk;
         }
       })
     )).filter((chunk) => chunk !== null);
+    retry++;
+    if(retry > 5) {
```
wemeetagain: can you put `retries` as optional param w/ default in this function?

wemeetagain (changes requested):

packages/lodestar/README.md
mpetrunic: That's default port, but you cannot start two nodes on same machine with same port

wemeetagain: sorry, pointing to typo in the first bootnode string

packages/lodestar/src/api/rest/routes/validator/publishBlock.ts
mpetrunic: But we will then publish blocks that we are syncing and potentially ruin our peer reputation

wemeetagain: I guess my comment is more about wondering if there is a way to hide network internals from the API, or at least this part of the code.
IMO the logic here should look something like:
"push new block into the system"
rather than
"add block here, add block there, etc"

mpetrunic: Probably, but api is only entrypoint where block needs to be processed as well as gossiped :/

wemeetagain: Ok, how about this goes to `src/api/impl`?

mpetrunic:
@tuyennhv what was the reason for e2e tests to stuck last time? It passes locally but never exits

packages/lodestar/src/api/rest/routes/validator/publishBlock.ts
mpetrunic: yes, but I'm gonna do it in separate PR where I will drop rpc?

twoeths:
> what was the reason for e2e tests to stuck last time? It passes locally but never exits

That's the incorrect order of libp2p dial then wait for connect event. See https://github.com/ChainSafe/lodestar/pull/658/files

packages/lodestar/README.md
```diff
@@ -32,13 +27,26 @@ Open resulting html file in favorite browser.
...
 You can always download spec test cases by running `yarn install --force`.
 
+### Development
+
+#### Connect two beacon nodes
+
+`cd packages/lodestar-cli`
+
+Terminal 1:
+`bin/lodestar dev -r` - It will run beacon node and 8 validators with all the default settings. State will be written to .tmp/state.ssz
```
twoeths: I have to add `-c 8` or it'll throw `cannot find state.ssz`

packages/lodestar/src/sync/regular/naive/naive.ts
```diff
@@ -0,0 +1,165 @@
...
+    const blocks = await getBlockRange(
+      this.network.reqResp,
+      this.reps,
+      this.peers,
+      {start: currentSlot + 1, end: this.targetSlot},
+      this.opts.blockPerChunk
+    );
+    const startBlockHeader = blockToHeader(
+      this.config,
+      (await this.db.block.get(blocks[0].message.parentRoot.valueOf() as Uint8Array)).message
```
twoeths: need to handle empty `blocks` array

wemeetagain (dismissed):

wemeetagain:
Whats the status of this PR?
I see its failing CI due to a typo?

mpetrunic:
> Whats the status of this PR?
> I see its failing CI due to a typo?

Should be fine now. I'm up for merging and then switching #709 #710 and #727 to merge into master

wemeetagain (approved):

PR #700, opened 2020-03-17
Update dev dependencies

wemeetagain:
We still have many "vulnerable" dependencies (run `yarn audit`), but not all are able to be fixed at this time.

Most stem from the following dependencies (which are only used in development):
ganache-core
web3


mpetrunic (dismissed):

twoeths (dismissed):

wemeetagain:
Looks like an eslint rule is now causing errors, will fix

mpetrunic (dismissed):

mpetrunic:
Maybe we should put mocha as peer dependency in spec-test-utils with some loose range of mocha dependencies?

mpetrunic (approved):

PR #709, opened 2020-03-18
Fix the validation of aggregate and proof

### LINKED ISSUE #691: AggregateAndProof gossip validation isn't working
To reproduce:
- uncomment validation
- run and connect two beacon nodes as per readme

My best guess is that we should fetch state/advance state matching attestation slot or something

twoeths:
## Goal
+ Close #691

### DIFF
--- a/packages/lodestar/src/network/gossip/validator.ts
+++ b/packages/lodestar/src/network/gossip/validator.ts
@@ -16,10 +16,15 @@ import {
   isValidIndexedAttestation,
   isValidProposerSlashing,
   isValidVoluntaryExit,
+  isAggregator,
+  getAttestingIndices,
+  getDomain,
+  computeEpochAtSlot,
 } from "@chainsafe/lodestar-beacon-state-transition";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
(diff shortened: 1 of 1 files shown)


packages/lodestar/src/network/gossip/validator.ts
```diff
@@ -85,48 +90,46 @@ export class GossipMessageValidator implements IGossipMessageValidator {
...
+    const attestorIndices = getAttestingIndices(this.config, state,
+      aggregationAndProof.aggregate.data, aggregationAndProof.aggregate.aggregationBits);
+    if (!attestorIndices.includes(aggregationAndProof.aggregatorIndex)) {
+      return false;
+    }
+    if (!isAggregator(
+      this.config,
+      state,
+      slot,
+      aggregationAndProof.aggregate.data.index,
```
twoeths, line 117: This is main fix, should be committee index instead of aggregator index

mpetrunic (changes requested):
Can you write test cases for this, I think it also shouldn't use latest state but not 100%

twoeths:
I think we'll end up with using the same state like the one to validate block in my other PR #710 . Need to apply for other validation too as long as we have a `slot`.

twoeths:
Replaced by #760

PR #710, opened 2020-03-18
Fix block signature validation

### LINKED ISSUE #692: Verify gossiped block signature bug
To reproduce:
- create following test cases:
   - received block from previous epoch
   - receive block from future epoch

We should fetch state matching block slot or advance latest state to block slot.
I guess we should not propagate very old blocks or blocks that are more than one epoch in future.

twoeths:
## Goal
+ Close #692

### DIFF
--- a/packages/lodestar/src/db/api/beacon/beacon.ts
+++ b/packages/lodestar/src/db/api/beacon/beacon.ts
@@ -2,7 +2,7 @@
  * @module db/api/beacon
  */
 
-import {BeaconState, BLSPubkey, ValidatorIndex, SignedBeaconBlock,} from "@chainsafe/lodestar-types";
+import {BeaconState, BLSPubkey, ValidatorIndex, SignedBeaconBlock, Slot,} from "@chainsafe/lodestar-types";
 import {DatabaseService, IDatabaseApiOptions} from "../abstract";
 import {IBeaconDb} from "./interface";
 import {
@@ -18,6 +18,7 @@ import {
   StateRepository,
   VoluntaryExitRepository
 } from "./repositories";
+import {computeEpochAtSlot, computeStartSlotAtEpoch, processSlots} from "@chainsafe/lodestar-beacon-state-transition";
 
 export class BeaconDb extends DatabaseService implements IBeaconDb {
 
@@ -100,6 +101,22 @@ export class BeaconDb extends DatabaseService implements IBeaconDb {
     const state = await this.state.getLatest();
     //TODO: cache this (hashmap)
     return state.validators.findIndex(value => this.config.types.BLSPubkey.equals(value.pubkey, publicKey));
+  }
+
+  public async getStateForSlot(slot: Slot): Promise<BeaconState> {
+    const epoch = computeEpochAtSlot(this.config, slot);
+    const startSlot = computeStartSlotAtEpoch(this.config, epoch);
+    const blockRoot = await this.chain.getBlockRoot(startSlot);
+    if (blockRoot) {
+      const block = await this.block.get(blockRoot);
+      return await this.state.get(block.message.stateRoot.valueOf() as Uint8Array);
+    }
+    const headBlock = await this.block.getChainHead();
+    const state = await this.state.get(headBlock.message.stateRoot.valueOf() as Uint8Array);
+    if(state.slot < startSlot) {
+      processSlots(this.config, state, startSlot);
+    }
+    return state;
   }
 
 }
--- a/packages/lodestar/src/network/gossip/validator.ts
+++ b/packages/lodestar/src/network/gossip/validator.ts
@@ -16,10 +16,15 @@ import {
   isValidIndexedAttestation,
   isValidProposerSlashing,
   isValidVoluntaryExit,
+  isAggregator,
+  getAttestingIndices,
+  getDomain,
+  computeEpochAtSlot,
 } from "@chainsafe/lodestar-beacon-state-transition";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
-import {ATTESTATION_PROPAGATION_SLOT_RANGE} from "../../constants";
+import {ATTESTATION_PROPAGATION_SLOT_RANGE, DomainType} from "../../constants";
 import {ILogger} from "@chainsafe/lodestar-utils/lib/logger";
+import {verify} from "@chainsafe/bls";
 
 export class GossipMessageValidator implements IGossipMessageValidator {
   private db: IBeaconDb;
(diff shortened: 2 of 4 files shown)


packages/lodestar/src/network/gossip/validator.ts
```diff
@@ -45,10 +53,14 @@ export class GossipMessageValidator implements IGossipMessageValidator {
     if (await this.db.block.has(root)) {
       return false;
     }
-    return true;
-    //TODO: fix this
-    // const state = await this.db.state.getLatest();
-    // return verifyBlockSignature(this.config, state, signedBlock);
```
twoeths, line 51: Main change in this PR.
Verified after running 2 beacon nodes in local env.

packages/lodestar/src/network/gossip/validator.ts
```diff
@@ -45,10 +53,14 @@ export class GossipMessageValidator implements IGossipMessageValidator {
...
+    const headBlock = await this.db.block.getChainHead();
+    const state = await this.db.state.get(headBlock.message.stateRoot.valueOf() as Uint8Array);
+    const epoch = computeEpochAtSlot(this.config, signedBlock.message.slot);
+    const startSlot = computeStartSlotAtEpoch(this.config, epoch);
+    if(state.slot < startSlot) {
+      processSlots(this.config, state, startSlot);
+    }
```
mpetrunic, line 54: We are doing this in couple of places. It probably make sense to expose method in state db `getStateForSlot` which will fetch existing stored state for slot <= head.slot or process slots in case of slot > head.slot

packages/lodestar/src/network/gossip/validator.ts
```diff
@@ -85,48 +97,46 @@ export class GossipMessageValidator implements IGossipMessageValidator {
...
-    // if (!verify(validatorPubKey.valueOf() as Uint8Array,
-    //   this.config.types.Slot.hashTreeRoot(slot),
-    //   aggregationAndProof.selectionProof.valueOf() as Uint8Array,
-    //   domain,
-    // )) {
-    //   return false;
-    // }
-    // const indexedAttestation = getIndexedAttestation(this.config, state, aggregationAndProof.aggregate);
-    // return isValidIndexedAttestation(this.config, state, indexedAttestation);
+    const root = this.config.types.AggregateAndProof.hashTreeRoot(aggregationAndProof);
```
mpetrunic: This doesn't work, so keep it commented until tested and fixed in #691

mpetrunic (dismissed):
Can you add test cases specified in issue

twoeths:
We had unit test here: https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/test/unit/network/gossip/validator.test.ts

Just need to modify/add test cases there.

mpetrunic:
> We had unit test here: https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/test/unit/network/gossip/validator.test.ts
> 
> Just need to modify/add test cases there.

Yeah, beacause it's missing important test cases :)

packages/lodestar/src/network/gossip/validator.ts
twoeths: I think again, probably we don't need `processSlots` call above. Wdyt?

mpetrunic: I think you have to do it because proposers change on every epoch.

packages/lodestar/test/unit/network/gossip/validator.test.ts
```diff
@@ -80,22 +81,47 @@ describe("GossipMessageValidator", () => {
     expect(await validator.isValidIncomingBlock(block)).to.be.equal(false);
   });
 
+  it("should return invalid incoming block - block is too old", async () => {
```
tuyennhv: not sure if this helps because we mock everything...

packages/lodestar/test/unit/network/gossip/validator.test.ts
```diff
@@ -80,22 +81,47 @@ describe("GossipMessageValidator", () => {
...
     dbStub.block.isBadBlock.resolves(false);
     dbStub.block.has.resolves(false);
     const state = generateState();
-    dbStub.state.getLatest.resolves(state);
+    dbStub.getStateForSlot.resolves(state);
+    verifyBlockSignatureStub.returns(true);
+    expect(await validator.isValidIncomingBlock(block)).to.be.equal(true);
+  });
+
+  it("should return valid incoming block - block in previous epoch", async () => {
```
twoeths, line 116: not sure if this helps because we mock everything... that's why I did not create another one for future epoch

packages/lodestar/src/db/api/beacon/beacon.ts
```diff
@@ -102,4 +103,20 @@ export class BeaconDb extends DatabaseService implements IBeaconDb {
...
+    const startSlot = computeStartSlotAtEpoch(this.config, epoch);
+    const blockRoot = await this.chain.getBlockRoot(startSlot);
+    if (blockRoot) {
+      const block = await this.block.get(blockRoot);
+      return await this.state.get(block.message.stateRoot.valueOf() as Uint8Array);
+    }
+    const headBlock = await this.block.getChainHead();
+    const state = await this.state.get(headBlock.message.stateRoot.valueOf() as Uint8Array);
+    if(state.slot < startSlot) {
+      processSlots(this.config, state, startSlot);
```
twoeths, line 117: is this good to put `processSlots` here in db? cannot find a better place ...

wemeetagain: I think using `processSlots` is fine, though we should be processing slots up to the requested slot.
We're trying to get the state at the slot in question to verify the block signature, and the proposer gets changed per-slot, so I believe we need the _exact_ slot.

One thing to keep in mind is this part of the database code will soon be changed. (Soon we'll be keeping recent states in memory).

An alternative to doing the `processSlots` here, you might return the first available state at or before the requested slot, and make the consumer (ie: the gossip validator) run `processSlots`.

wemeetagain (changes requested):

packages/lodestar/src/db/api/beacon/beacon.ts
twoeths: yeah I get your logical points, I can do that but still have concerns regarding having to do `processSlots` up until the requested slot (may be wrong but here's what I see from the specs). To get correct state to validate block proposer signature, we need 3 things: the seed, the slot(not a concern) and active validator indixes

+ the seed is calculated from `randao_mixes` of 2 epochs earlier. If our head is older than that, just return null and make the validation failed. Running `processSlots` does not give correct `randao_mixes` for the current epoch because it should be updated block by block (in `process_block`).
+ From the 1st look, running `process_slots` may help get correct active validator indexes because we run thru `process_epoch` then `processRegistryUpdates`. But a validator can only becomes active in `epoch + 1 + config.params.MAX_SEED_LOOKAHEAD` which is `epoch+5`. In most of the cases, the head state should have same active validator indexes. If not, running `process_slots` does not help because the head is 5 epochs older than the block and it failed because of the `seed` condition above.

My thinking is that it's not wrong to run `processSlots` here, it's just not necessary to run it at all because we have same result anyway.

mpetrunic: > a validator can only becomes active in epoch + 1 + config.params.MAX_SEED_LOOKAHEAD

If you have validator that deposited some time ago and it's suppose to become active in current epoch +1. Running process slots till current epoch + 1 will change active validator indices

twoeths: Maybe it's not like that, let's look into the detail of `processRegistryUpdates` (lmk if I look at wrong place):
```
activationQueue.slice(0, getValidatorChurnLimit(config, state)).forEach((validator) => {
    if (validator.activationEpoch === FAR_FUTURE_EPOCH) {
      validator.activationEpoch = computeActivationExitEpoch(config, currentEpoch);
    }
  });
```
and 
```
export function computeActivationExitEpoch(config: IBeaconConfig, epoch: Epoch): Epoch {
  return epoch + 1 + config.params.MAX_SEED_LOOKAHEAD;
}
```
Compared head state and slot, we always have same set of active validators unless head state is more than 4/5 epochs (MAX_SEED_LOOKAHEAD) earlier compared to slot. And in that case `randao_mixes` is not correct anyway if we run `processSlots` as analyzed above. Hence with or without running `processSlots`, we both have the result as `false`.

mpetrunic: You are looking at right place but interpreting code wrong.
Scenario:
- we are at epoch 10
- validator deposited, passed activation queue and now validator.activationEpoch = 11(+1 + MAX_SEED_LOOKAHEAD)
- we received block in epoch 11
- we process slots till epoch 11 and on epoch transition that validator is not active and changes active validator indices

twoeths: In my understanding, in your scenario we don't need to run `processSlots()`. If you look at my validation, I'll pick head state and use `{...state, slot}` to verify the block signature. It should get the correct active validators.

On the other hand in `processSlots => processRegistryUpdates`, it only update `activationEpoch` (normally `current epoch+5`) while the place to get active validators is decided in `getActiveValidatorIndices`. As long as we put in the correct slot to state, it should get correct active valiator index.

twoeths: If I look at your explanation and mine, I think the main confusion comes from using `processSlots()` as yours/Cayman's and using `{...headState,slot}` as mine (I studied this way in `getEpochProposers`). I think using `{..headState,slot}` should be enough.

mpetrunic: process_epoch also modifes randao_mixes in final updates. Anyways, if we don't do process_slots, we are exposing ourselves to bugs caused by change in spec which affects proposers and it's way too hard to determine what will or will not affect it.

twoeths: > process_epoch also modifes randao_mixes in final updates

this `randao_mixes` needs to be "mixed"/updated by `process_block` of every block of epoch (then `process_randao`). If we merely run `process_slots` then the resulting `randao_mixes` may not be correct in final update.

> Anyways, if we don't do process_slots, we are exposing ourselves to bugs caused by change in spec which affects proposers and it's way too hard to determine what will or will not affect it.

For this point, I agree with this. I wish it's part of the spec so that all client implementation can follow.

Let's go with `process_slots`, sorry took your time lol ..

wemeetagain:
file conflict

twoeths:
Ready for review

wemeetagain (dismissed):

twoeths:
Signed CLA ^^

wemeetagain (approved):

PR #713, opened 2020-03-19
Filter peers on init sync

### LINKED ISSUE #696: Fix initial sync bug
- If we are syncing from node/s that aren't yet finalized, we should skip initial sync

twoeths:
## Goal
+ Close #696

### DIFF
--- a/packages/lodestar/src/sync/utils/sync.ts
+++ b/packages/lodestar/src/sync/utils/sync.ts
@@ -3,7 +3,6 @@ import {BeaconBlockHeader, Checkpoint, Epoch, Slot, SignedBeaconBlock, Status,
   BeaconState} from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {IReqResp} from "../../network";
-import {GENESIS_EPOCH} from "../../constants";
 
 export function isValidChainOfBlocks(
   config: IBeaconConfig,
--- a/packages/lodestar/src/sync/utils/sync.ts
+++ b/packages/lodestar/src/sync/utils/sync.ts
@@ -72,9 +72,7 @@ export function isValidPeerForInitSync(config: IBeaconConfig, myState: BeaconSta
     return false;
   }
   // TODO: compare fork_digest in the latest spec?
-  if (!peerStatus.finalizedRoot || peerStatus.finalizedEpoch <= GENESIS_EPOCH) {
-    return false;
-  }
+
   if (peerStatus.finalizedEpoch < myState.finalizedCheckpoint.epoch) {
     return false;
   }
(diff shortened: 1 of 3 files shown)


packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -65,6 +67,22 @@ export function isValidFinalizedCheckPoint(peers: IReputation[], finalizedCheckp
   return peerCount >= (validPeers.length / 2);
 }
 
+export function isValidPeerForInitSync(config: IBeaconConfig, myState: BeaconState, peerStatus: Status): boolean {
+  if (!peerStatus) {
+    return false;
+  }
+  if (!config.types.Version.equals(myState.fork.currentVersion, peerStatus.headForkVersion)) {
```
mpetrunic: We should remove this check as our peer might be on different fork version after block x but it's still valid for block 0..x
I think this is specified in latest spec release

twoeths: yeah, just saw we have to check something called `fork_digest` in the latest spec. Should mark a TODO here instead.

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -65,6 +67,20 @@ export function isValidFinalizedCheckPoint(peers: IReputation[], finalizedCheckp
   return peerCount >= (validPeers.length / 2);
 }
 
+export function isValidPeerForInitSync(config: IBeaconConfig, myState: BeaconState, peerStatus: Status): boolean {
+  if (!peerStatus) {
+    return false;
+  }
+  // TODO: compare fork_digest in the latest spec?
+  if (!peerStatus.finalizedRoot || peerStatus.finalizedEpoch <= GENESIS_EPOCH) {
```
mpetrunic: I just realized this check is also not ok because if we are trying to sync two peers locally and peer 1 is at block 5(no finalized epochs), sync won't start

mpetrunic (changes requested):

packages/lodestar/src/sync/utils/sync.ts
twoeths: oh I thought that's what we expect as in the issue description: `If we are syncing from node/s that aren't yet finalized, we should skip initial sync`

mpetrunic: I think, I've pushed some fix to regular sync branch to handle this inside initialSync. Basically it would enter initial sync, figure out there is no peer with finalized epoch and mark initial sync compleated. We should probably close that issue with "regular sync" PR merge. This PR is still valid with just last check.

There is a lot of cases we should handle that, like you skipped initial sync but you discovered peer with higher finalized epoch, you should try to initial sync again

mpetrunic (approved):

wemeetagain (approved):
LGTM

PR #725, opened 2020-03-20
Add headStateRoot() to forkchoice

### LINKED ISSUE #723: Expose headStateRoot in fork choice
Currently, we only expose the head block root in our fork choice module.
It would be more convenient and efficient to also expose the state root.
(we have a pattern many places in the code where we operate on the head state)

twoeths:
## Goal
+ Close #723

### DIFF
--- a/packages/lodestar/src/chain/forkChoice/interface.ts
+++ b/packages/lodestar/src/chain/forkChoice/interface.ts
@@ -2,21 +2,17 @@
  * @module chain/forkChoice
  */
 
-import {Gwei, Slot, ValidatorIndex, Checkpoint} from "@chainsafe/lodestar-types";
+import {Gwei, ValidatorIndex, Checkpoint} from "@chainsafe/lodestar-types";
+import {BlockChainInfo} from "./statefulDag/interface";
 
 
 export interface ILMDGHOST {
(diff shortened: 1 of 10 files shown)


packages/lodestar/src/chain/forkChoice/interface.ts
```diff
@@ -2,21 +2,17 @@
  * @module chain/forkChoice
  */
 
-import {Gwei, Slot, ValidatorIndex, Checkpoint} from "@chainsafe/lodestar-types";
+import {Gwei, ValidatorIndex, Checkpoint} from "@chainsafe/lodestar-types";
+import {BlockChainInfo} from "./statefulDag/interface";
```
wemeetagain: `BlockChainInfo` can be defined in this file

wemeetagain (changes requested):
This looks good :)

packages/lodestar/src/chain/forkChoice/interface.ts
twoeths: I failed to do that because it's referenced in the outer `interface.ts`. The outer `interface.ts` should be a better place because it's part of the `ILMDGHOST` interface anyway.

wemeetagain (approved):

Issue #704, opened 2020-03-17
Use timestamp instead of block height in eth1 voting to allow for simpler caching


ColinSchwarz:
https://github.com/ethereum/eth2.0-specs/pull/1553

twoeths:
Need to make sure spec test passed after this one is done.

PR #756, opened 2020-03-25
Use GENESIS_SLOT, GENESIS_EPOCH as constants

### LINKED ISSUE #703: Move GENESIS_SLOT/GENESIS_EPOCH to constants
(they were not truly configurable in the first place)
https://github.com/ethereum/eth2.0-specs/pull/1556

twoeths:
## Goal
+ Close #703 
+ There are other changes regarding Eth1Data that should be addressed by #704


mpetrunic (commented):
I think you can calculate genesis epoch from genesis slot though

wemeetagain:
genesis state initialization spec tests failing

twoeths:
> genesis state initialization spec tests failing

It's because we failed to verify deposits' signatures in that test. I notice that there are signature differences between data of spec test 0.10.0 vs 0.10.1-fix, let's revisit this issue in the newly created unit test `deposit2.test.ts`

twoeths:
Do we want to completely remove `GENESIS_EPOCH` constant?

mpetrunic (approved):

wemeetagain (approved):

PR #757, opened 2020-03-26
Use forkchoice headStateRoot for Gossip Validator

twoeths:
+ PR #710 has this commit then I rebase to sign CLA and it lost this part

### DIFF
--- a/packages/lodestar/src/network/gossip/validator.ts
+++ b/packages/lodestar/src/network/gossip/validator.ts
@@ -18,17 +18,22 @@ import {
   isValidVoluntaryExit,
   verifyBlockSignature,
   computeStartSlotAtEpoch,
+  processSlots,
 } from "@chainsafe/lodestar-beacon-state-transition";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {ATTESTATION_PROPAGATION_SLOT_RANGE} from "../../constants";
 import {ILogger} from "@chainsafe/lodestar-utils/lib/logger";
+import {IBeaconChain} from "../../chain";
 
 export class GossipMessageValidator implements IGossipMessageValidator {
+  private chain: IBeaconChain;
   private db: IBeaconDb;
   private config: IBeaconConfig;
   private logger: ILogger;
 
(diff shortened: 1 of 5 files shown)


twoeths:
Tested the sync of 2 nodes in my local environment

packages/lodestar/src/network/gossip/validator.ts
```diff
@@ -78,7 +87,10 @@ export class GossipMessageValidator implements IGossipMessageValidator {
     if (!await this.db.block.has(blockRoot) || await this.db.block.isBadBlock(blockRoot)) {
       return false;
     }
-    const state = await this.db.state.getLatest();
+    const state = await this.db.state.get(this.chain.forkChoice.headStateRoot());
+    if (state.slot < attestation.data.slot) {
+      processSlots(this.config, state, attestation.data.slot);
+    }
```
mpetrunic, line 93: Any particular reason to move this from db?
I think you could ib db just pass slot and forkchoice head root which would enable us later to cache those states instead of recalculate them every time and tracking where we use this code?

twoeths: I can put it at a common place other than db. Just found it a little bit weird if I put in in DB because we mix state transition function `processSlot` in db code, do you think so?

mpetrunic: I'm ok with that, we use other state  transition helpers in db code. I agree it would be more appropriate to put it in some db wrapper but we don't have one atm. So I figured it would make more sense to combine all those duplicated code their so it's easier to refactor in some wrapper later on

mpetrunic: cc @wemeetagain

wemeetagain: :thinking: I currently like this better than putting it in the DB, even if there's a bit of duplication.
Reason being, its easier to majorly overhaul an implementation with an interface that's simple and doesn't imply much logic.
I'd like to revamp our state storage in the near future, and I can take a look at expanding the interface then.

wemeetagain (approved):

PR #758, opened 2020-03-27
Block process stream

### LINKED ISSUE #481: Create block pool
- should accept blocks and form chains (might contain multiple chains in case of forks)
- should request parent blocks if missing

### LINKED ISSUE #694: Use streams instead of queues
Convert current attestation and block queues into streams

mpetrunic:
I would like some feedback on this block processing stream proposal before I start writing tests and modifying rest of codebase
Block lifecycle is:
network/api -> chain -> blockProcessStream - > validate -> process -> postProcess
                                                                                          \
                                                                                             -> blockPool-> blockProcessStream
After we process block, we will notify block pool which will put blocks waiting for that parent back on process stream.

Pros:
- very much testable code so we can cover a lot of scenarios with unit tests
- clean modular code
- parallel validate/process/post processes

Cons:
- it seems like during sync every block except first will end up in block pool before being process
   - it because there is a great chance most block will go trough validate before block process starts
   - naive idea to mitigate it is to combine validate and process function into one but we will end up with huge function again and we wouldn't be able to discard some blocks while waiting for process operations

resolves #694 
resolves #481

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -82,7 +82,8 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
   }
 
   public isInitialized(): boolean {
-    throw new Error("Method not implemented.");
+    //TODO: implement
+    return true;
   }
 
   public async start(): Promise<void> {
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -41,9 +41,9 @@ export interface IBlockProcessJob {
 export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) implements IBeaconChain {
 
   public readonly chain: string;
-  public readonly forkChoice: ILMDGHOST;
-  public readonly chainId: Uint16;
-  public readonly networkId: Uint64;
+  public forkChoice: ILMDGHOST;
+  public chainId: Uint16;
+  public networkId: Uint64;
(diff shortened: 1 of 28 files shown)


twoeths:
This looks great. I suppose you're talking about the init sync? Regular sync should be fine I guess.

wemeetagain:
I like this structure so far, should be easy to tweak, eg: add stages, etc

mpetrunic:
We are now handling and processing block from different forks and starting from different state than head!!

And block process is unit tested

wemeetagain (approved):
LGTM

twoeths (approved):

packages/lodestar/src/chain/chain.ts
```diff
@@ -203,151 +156,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
-    const hexBlockHash = toHexString(blockHash);
-    this.logger.info(`${hexBlockHash} is valid, running state transition...`);
-
-    // process current slot
-    const post = await this.runStateTransition(job.signedBlock, pre);
-
-    this.logger.info(
-      `Slot ${job.signedBlock.message.slot} Block ${hexBlockHash} ` +
-      `State ${toHexString(this.config.types.BeaconState.hashTreeRoot(post))} passed state transition`
-    );
```
twoeths, line 242: Seems like we miss the below operations. Added back in https://github.com/ChainSafe/lodestar/pull/806

PR #759, opened 2020-03-28
Get eth1 data using eth1 block timestamp

### LINKED ISSUE #704: Use timestamp instead of block height in eth1 voting to allow for simpler caching

twoeths:
## Goal
+ Close #704
+ The main thing is to implement an eth1 block cache in order to get eth1 data

### DIFF
--- a/packages/lodestar/src/eth1/impl/eth1Vote.ts
+++ b/packages/lodestar/src/eth1/impl/eth1Vote.ts
@@ -1,36 +1,14 @@
 import {IEth1Notifier} from "../interface";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
-import {BeaconState, Eth1Data, Number64} from "@chainsafe/lodestar-types";
-import {intSqrt} from "@chainsafe/lodestar-utils";
+import {BeaconState, Eth1Data} from "@chainsafe/lodestar-types";
 import {arrayIntersection, mostFrequent, sszEqualPredicate} from "../../util/objects";
-import {Block} from "ethers/providers";
+import {votingPeriodStartTime} from "./utils";
 
 export async function getEth1Vote(
-  this: IEth1Notifier, config: IBeaconConfig, state: BeaconState, previousEth1Distance: Number64
-): Promise<Eth1Data> {
-  const eth1Head = await this.getHead();
-  const allEth1Data = await getEth1DataRange(
-    this.getEth1Data,
-    eth1Head,
-    config.params.ETH1_FOLLOW_DISTANCE,
-    previousEth1Distance
-  );
-
-  const periodTail =
-        state.slot % config.params.SLOTS_PER_ETH1_VOTING_PERIOD
-        >=
-        intSqrt(config.params.SLOTS_PER_ETH1_VOTING_PERIOD);
-  let votesToConsider: Eth1Data[];
-  if(periodTail) {
-    votesToConsider = allEth1Data;
-  } else {
-    votesToConsider = await getEth1DataRange(
-      this.getEth1Data,
-      eth1Head,
-      config.params.ETH1_FOLLOW_DISTANCE,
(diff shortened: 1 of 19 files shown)


packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -63,19 +70,53 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
     this.emit("block", block);
+    let requestedBlockNumber = this.blocksCache.requestNewBlock(block);
+    if (requestedBlockNumber) {
+      let requestedBlock: Block;
+      let retry = 0;
+      while (!requestedBlock && retry < GET_ETH1_BLOCK_RETRY) {
+        requestedBlock = await this.getBlock(requestedBlockNumber);
+        requestedBlockNumber++;
+        retry++;
+      }
```
mpetrunic, line 114: We are doing this in bunch of places now. Can you create some utility function like:
```javascript
export async function retryable<T>(func: (...args) => T|null , retry = 3): T|null;
```
so we would have:
```javascript
const requestedBlock =  retryable(this.getBlock.bind(this, requestedBlockNumber), GET_ETH1_BLOCK_RETRY)
```

packages/lodestar/src/eth1/impl/eth1Vote.ts
```diff
@@ -1,36 +1,14 @@
...
-    votesToConsider = await getEth1DataRange(
-      this.getEth1Data,
-      eth1Head,
-      config.params.ETH1_FOLLOW_DISTANCE,
-      2 * config.params.ETH1_FOLLOW_DISTANCE
-    );
-  }
+  this: IEth1Notifier, config: IBeaconConfig, state: BeaconState): Promise<Eth1Data> {
+  const periodStart = votingPeriodStartTime(config, state);
+  const promises: Promise<Eth1Data>[] = this.findBLocks(config, periodStart).map(block => this.getEth1Data(block));
```
mpetrunic:
```suggestion
  const promises: Promise<Eth1Data>[] = this.findBlocks(config, periodStart).map(block => this.getEth1Data(block));
```

mpetrunic: rename method implementation also

mpetrunic (commented):
Haven't checked against spec yet

packages/lodestar/src/eth1/impl/eth1Vote.ts
twoeths: not sure if I understand this correctly. Do you mean to rename the `findBlocks` method?

mpetrunic: notice lowercase L

twoeths: I see it now lol

packages/lodestar/src/eth1/impl/ethers.ts
twoeths: I have to increase `requestedBlockNumber` on every call so have to come up with something like this which is quite complex:
```
export async function retryable<T>(funcGen: Generator<(...args: any[]) => Promise<T|null>>, retry = 3): Promise<T|null> {
  let result: T;
  let count = 0;
  while(!result && count < retry) {
    const func = funcGen.next().value;
    result = await func();
    count++;
  }
  return result;
}
```
And the consumers have to implement a generator. Any simpler approach?

mpetrunic: I think you should support plain function as well and check inside `retryable` if argument is iterator or plain function

mpetrunic (approved):

wemeetagain (approved):

PR #763, opened 2020-03-31
Use MIN_GENESIS_DELAY to init state

### LINKED ISSUE #702: Make genesis delay configurable for ease of setting up testnets
https://github.com/ethereum/eth2.0-specs/pull/1557

twoeths:
## Goals
+ Close #702 
+ Temporarily added `deposit2.test.ts` to investigate the failed spec test. Should revisit the issue once we're in v0.10.1 because it should pass v0.10.0 spec test.

### DIFF
--- a/packages/lodestar-beacon-state-transition/test/unit/stateTransition/block/operations/deposit2.test.ts
+++ b/packages/lodestar-beacon-state-transition/test/unit/stateTransition/block/operations/deposit2.test.ts
@@ -1,0 +1,86 @@
+import { generatePublicKey } from "@chainsafe/bls";
+import { computeDomain } from "../../../../../src/util/domain";
+import {config} from "@chainsafe/lodestar-config/lib/presets/mainnet";
+import { computeSigningRoot } from "../../../../../src/util/signingRoot";
+import bls, {initBLS} from "@chainsafe/bls";
+import {assert} from "chai";
+import {DepositMessage} from "@chainsafe/lodestar-types";
+import { DomainType } from "../../../../../src/constants";
+
+
+it("should verify deposit 2222", async () => {
+  await initBLS();
+  // Same deposit used in genesis_initialization_minimal.test.ts
+  const specJson = `
+    { "proof":
+      [ "0x0000000000000000000000000000000000000000000000000000000000000000",
+        "0xf5a5fd42d16a20302798ef6ed309979b43003d2320d9f0e8ea9831a92759fb4b",
+        "0xdb56114e00fdd4c1f85c892bf35ac9a89289aaecb1ebd0a96cde606a748b5d71",
+        "0xc78009fdf07fc56a11f122370658a353aaa542ed63e44c4bc15ff4cd105ab33c",
+        "0x536d98837f2dd165a55d5eeae91485954472d56f246df256bf3cae19352a123c",
+        "0x9efde052aa15429fae05bad4d0b1d7c64da64d03d7a1854a588c2cb8430c0d30",
+        "0xd88ddfeed400a8755596b21942c1497e114c302e6118290f91e6772976041fa1",
+        "0x87eb0ddba57e35f6d286673802a4af5975e22506c7cf4c64bb6be5ee11527f2c",
+        "0x26846476fd5fc54a5d43385167c95144f2643f533cc85bb9d16b782f8d7db193",
+        "0x506d86582d252405b840018792cad2bf1259f1ef5aa5f887e13cb2f0094f51e1",
+        "0xffff0ad7e659772f9534c195c815efc4014ef1e1daed4404c06385d11192e92b",
+        "0x6cf04127db05441cd833107a52be852868890e4317e6a02ab47683aa75964220",
+        "0xb7d05f875f140027ef5118a2247bbb84ce8f2f0f1123623085daf7960c329f5f",
+        "0xdf6af5f5bbdb6be9ef8aa618e4bf8073960867171e29676f8b284dea6a08a85e",
+        "0xb58d900f5e182e3c50ef74969ea16c7726c549757cc23523c369587da7293784",
+        "0xd49a7502ffcfb0340b1d7885688500ca308161a7f96b62df9d083b71fcc8f2bb",
+        "0x8fe6b1689256c0d385f42f5bbe2027a22c1996e110ba97c171d3e5948de92beb",
+        "0x8d0d63c39ebade8509e0ae3c9c3876fb5fa112be18f905ecacfecb92057603ab",
+        "0x95eec8b2e541cad4e91de38385f2e046619f54496c2382cb6cacd5b98c26f5a4",
+        "0xf893e908917775b62bff23294dbbe3a1cd8e6cc1c35b4801887b646a6f81f17f",
+        "0xcddba7b592e3133393c16194fac7431abf2f5485ed711db282183c819e08ebaa",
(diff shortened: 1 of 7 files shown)


packages/lodestar-beacon-state-transition/test/unit/stateTransition/block/operations/deposit2.test.ts
```diff
@@ -0,0 +1,86 @@
...
+import { computeSigningRoot } from "../../../../../src/util/signingRoot";
+import bls, {initBLS} from "@chainsafe/bls";
+import {assert} from "chai";
+import {DepositMessage} from "@chainsafe/lodestar-types";
+import { DomainType } from "../../../../../src/constants";
+
+
+it("should verify deposit 2222", async () => {
+  await initBLS();
+  // Same deposit used in genesis_initialization_minimal.test.ts
```
twoeths, line 14: This is the data I extracted from the failed spec test.

packages/lodestar-beacon-state-transition/test/unit/stateTransition/block/operations/deposit2.test.ts
```diff
@@ -0,0 +1,86 @@
...
+  const deposit = config.types.Deposit.fromJson(JSON.parse(specJson));
+  const privKey: Uint8Array = new Uint8Array(32);
+  privKey[31] = 1;
+  // Make sure we use correct privKey
+  assert.deepEqual(deposit.data.pubkey, generatePublicKey(privKey));
+  const domain = computeDomain(DomainType.DEPOSIT, config.params.GENESIS_FORK_VERSION);
+  const signingRoot = computeSigningRoot(config, config.types.DepositMessage, deposit.data, domain);
+  const signature = bls.sign(privKey, signingRoot);
+  // This is different from the signature we saw above
+  // b04653bcd8bd711e653fbdfdf14325bcc5d0f8fe55f69388b0f547669e02746225bbac72e0b7dad758a404cf223b58040c287f112e4662a2cb9602a1bc0c99a0b03479c17f500ea3b00ec2efc7f0aca84ce71639e1da9270ee3bb75831ca1f63
```
twoeths, line 69: The spec tests failed because we can't verify deposit data's signature.
In this unit test, we use same deposit, same public key & private key to the spec test but the produced signature is different to the spec test. @mpetrunic do you have any ideas?

twoeths: just found that my resulting signature is the one in spec test v0.10.0 while the other one is v0.10.1.

So spec test can only be passed once we're up to v0.10.1

mpetrunic: Ignore spec tests 0.10.0, they are broken, we should target 0.10.1
Can you explain better what's the issue here in detail?

packages/lodestar-beacon-state-transition/test/unit/stateTransition/block/operations/deposit2.test.ts
```diff
@@ -0,0 +1,86 @@
...
+         "withdrawalCredentials":
+         "0x00cf478a431837728dcec3461f4f53b8749cdc4e03496dcaed459dea82b82eb8",
+        "amount": "32000000000",
+        "signature":
+         "0xa19e892d93edb98971268e68bdd7a9abbd02764706b1816aa2a1a94f205f1b2174849592307255acc60e1bbb8bd82aa5190764b41cc01262ff6a257981098a0431fd2d1f15c039b19fb92b013da786251926bdd21bd1754d823809e19129646b" } }
+  `;
+    
+  const deposit = config.types.Deposit.fromJson(JSON.parse(specJson));
+  const privKey: Uint8Array = new Uint8Array(32);
+  privKey[31] = 1;
```
mpetrunic: PrivateKey.fromInt(1)

packages/lodestar-beacon-state-transition/test/unit/stateTransition/block/operations/deposit2.test.ts
twoeths: The spec test failed when init state because the provided deposit in the spec test can't be validated in our updated state transition code. Then I extracted the deposit data here, and try to get correct private key/public key used in the test. Then I notice the generated signature here is totally different from the test. But actually it's good for 0.10.0 already.

twoeths: I'll remove this test in the next commit, it's for my own investigation only.

twoeths: I don't see any update in v0.10.1 that may cause the signature change, so the spec test will likely to be failed once we're in v0.10.1, probably will have to revisit this issue anyway.
Is it ok to temporarily keep this test?

packages/lodestar-beacon-state-transition/test/unit/stateTransition/block/operations/deposit2.test.ts
```diff
@@ -57,15 +58,16 @@ it("should verify deposit 2222", async () => {
...
-  privKey[31] = 1;
+  const privKey: Uint8Array = PrivateKey.fromInt(1).toBytes();
   // Make sure we use correct privKey
   assert.deepEqual(deposit.data.pubkey, generatePublicKey(privKey));
   const domain = computeDomain(DomainType.DEPOSIT, config.params.GENESIS_FORK_VERSION);
   const signingRoot = computeSigningRoot(config, config.types.DepositMessage, deposit.data, domain);
   const signature = bls.sign(privKey, signingRoot);
-  // This is different from the signature we saw above
+  // The signature above is from spec 0.10.1: https://media.githubusercontent.com/media/ethereum/eth2.0-spec-tests/v0.10.1/tests/minimal/phase0/genesis/initialization/pyspec_tests/initialize_beacon_state_from_eth1/deposits_0.yaml
+  // This is the same to spec 0.10.0: https://media.githubusercontent.com/media/ethereum/eth2.0-spec-tests/v0.10.0/tests/minimal/phase0/genesis/initialization/pyspec_tests/initialize_beacon_state_from_eth1/deposits_0.yaml
```
mpetrunic, line 68: Signature that we generate matches 0.10.0 spec?
Did you copy their signature from spec tests repo because you should copy one from tar.gz arhive since they didn't update test files in repo

We are passing 0.10.1 signing spec tests so it shouldn't be bls

twoeths: > Signature that we generate matches 0.10.0 spec?

Yes, it matches the value specified in the provided link of 0.10.0.
> Did you copy their signature from spec tests repo because you should copy one from tar.gz arhive since they didn't update test files in repo

No, I used `config.types.Deposit.toJson(deposit).toJson()` in spec test and `console.log` from there, then use the same value in this unit test by copying the string.

The spec test is to verify deposit data, it failed. This test is to use the same deposit data and create signature from there, it's totally different from the spec test (0.10.1) but the same to 0.10.0.

I have no ideas for now, don't see anything in v0.10.1 that may cause the signature change

twoeths:
Ready for review

wemeetagain (approved):

mpetrunic (approved):

PR #764, opened 2020-04-01
Eager attestation broadcasting

### LINKED ISSUE #705: Add eager attestation broadcasting to optimize attestation propagation in the normal case
https://github.com/ethereum/eth2.0-specs/pull/1555

twoeths:
## Goal
+ Close #705

### DIFF
--- a/packages/lodestar-validator/src/services/attestation.ts
+++ b/packages/lodestar-validator/src/services/attestation.ts
@@ -22,7 +22,6 @@ import {
   computeEpochAtSlot, DomainType, getDomain, isSlashableAttestationData, computeSigningRoot,
 } from "@chainsafe/lodestar-beacon-state-transition";
 
-import {sleep} from "../util";
 import {IAttesterDuty} from "../types";
 
 export class AttestationService {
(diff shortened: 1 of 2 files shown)


mpetrunic (changes requested):
This is not good solution.
Now if our validator connects to bn that  does not handle this it will break.
As per spec, our validator needs to listen on blocks(we don't have this api yet), and publish if it receives block to keep this change backward compatible. I think we should close this and first add optional block stream api. We should also handle situation where block stream api is not available with just waiting `this.config.params.SECONDS_PER_SLOT / 3 * 1000` before publishing

wemeetagain:
I agree with Marin in that this is something that should probably be tracked on the validator side.
I think ideally, validators subscribe to updates from the BN (eg: opening a websocket(s) or eventsource(s) with the BN) which gets updates pushed as they happen. New duties, new block notifications, reorgs, attestations related to the duties, etc can be pushed to the VC. And the VC can push back signed versions to be published when its rules say to do so.

In a case like that, the validator would start the timer at the start of the slot they have attestation duty, and if they receive notice of the block in question before it goes off, then they proceed.

Issue #707, opened 2020-03-17
Validate block not from future slot 


ColinSchwarz:
(and add a small clock disparity allowance)
https://github.com/ethereum/eth2.0-specs/pull/1563

twoeths:
It's better to wait for #692

PR #766, opened 2020-04-01
Merge master to 0.10.1

twoeths:
## Goal
+ I need this in order to work on #707


packages/lodestar-params/src/presets/minimal.ts
```diff
@@ -1,70 +1,12 @@
...
-export const MAX_VOLUNTARY_EXITS = 2 ** 4; // 16
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
+import minimalYaml from "./minimal.yaml";
+import {convertTypes, schema} from "../utils";
+import {typeMap} from "../types";
+import {load} from "js-yaml";
+
+
+// TODO: fix this
```
twoeths: `GENESIS_FORK_VERSION` is not loaded correctly, will create an issue for it

twoeths:
Unit tests are not passed, investigating ...

mpetrunic (approved):

packages/lodestar-params/src/presets/minimal.ts
mpetrunic: just remove it from [constants](https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar-params/src/constants.ts) as it was not in config before

twoeths: it works, thanks

PR #772, opened 2020-04-03
Handle skip slots in fork choice

### LINKED ISSUE #770: [0.10.1]: Properly handle skip slots in fork choice
Spec PR: https://github.com/ethereum/eth2.0-specs/pull/1579

twoeths:
## Goal
+ Close #770

### DIFF
--- a/packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
+++ b/packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
@@ -292,7 +292,7 @@ export class StatefulDagLMDGHOST implements ILMDGHOST {
       slot,
       blockRoot,
       stateRoot: stateRootBuf,
-      parent: this.nodes[parentRoot], 
+      parent: this.nodes[parentRoot],
       justifiedCheckpoint: {rootHex: toHexString(justifiedCheckpoint.root), epoch: justifiedCheckpoint.epoch},
       finalizedCheckpoint: {rootHex: toHexString(finalizedCheckpoint.root), epoch: finalizedCheckpoint.epoch},
     });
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
```diff
@@ -391,11 +404,8 @@ export class StatefulDagLMDGHOST implements ILMDGHOST {
       return true;
```
twoeths, line 404: Not related to this PR, I think this condition should be the other way as the other conditions:
```
if (computeSlotsSinceEpochStart(this.config, getCurrentSlot(this.config, this.genesisTime)) >=
      this.config.params.SAFE_SLOTS_TO_UPDATE_JUSTIFIED)
      return false;
```
The reason is we want to check other conditions as well to make sure we have good justified checkpoints.
What do you think @wemeetagain ? Although it's the same to the spec, I think it's more reasonable to switch the condition, or make it the last condition in this function.

wemeetagain: :+1: that makes sense to me

packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
```diff
@@ -290,13 +292,24 @@ export class StatefulDagLMDGHOST implements ILMDGHOST {
       slot,
       blockRoot,
       stateRoot: stateRootBuf,
-      parent: this.nodes[parentRoot],
+      parent: this.nodes[parentRoot],
```
wemeetagain: unnecessary space

packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
twoeths: I see the spec and Prysmatic implement the same to this. Should we make a PR to the spec as well?

wemeetagain: We should definitely clarify what the intended behavior is

wemeetagain (approved):

PR #774, opened 2020-04-05
discv5 integration

wemeetagain:
Draft status until `@chainsafe/discv5` is published
and a libp2p issue is resolved (https://github.com/libp2p/js-libp2p/pull/600)


packages/lodestar/src/network/nodejs/util.ts
```diff
@@ -31,11 +31,13 @@ export async function loadPeerIdFromJsonFile(path: string): Promise<PeerId> {
...
+  network: Partial<INetworkOptions> = {}
+): Promise<LibP2p> {
   const peerId = await Promise.resolve(peerIdOrPromise);
   const multiaddrs = network.multiaddrs || defaults.multiaddrs;
   const bootnodes = network.bootnodes || defaults.bootnodes;
   const peerInfo = await initializePeerInfo(peerId, multiaddrs);
-  return new NodejsNode({peerInfo, bootnodes: bootnodes});
-}
\ No newline at end of file
+  return new NodejsNode({peerInfo, bootnodes: bootnodes, discv5: network.discv5});
```
twoeths, line 42: Just a small note: we'll need to put in `ENRForkID` as spec 0.11 PR https://github.com/ethereum/eth2.0-specs/pull/1652/files

wemeetagain: :+1: yeah, the enr should be available through `libp2p._discovery.get("discv5").enr`.

I'm thinking of doing this in additional PRs.

packages/lodestar/src/network/discovery/discv5.ts
```diff
@@ -0,0 +1,75 @@
...
+      for (const enr of enrs) {
+        try {
+          const peerInfo = new PeerInfo(await enr.peerId());
+          const multiaddrTCP = enr.multiaddrTCP;
+          if (multiaddrTCP) {
+            peerInfo.multiaddrs.add(multiaddrTCP);
+          }
+          this.emit("peer", peerInfo);
+        } catch (e) {
+          continue;
```
mpetrunic: might be useful to log warning with error?

packages/lodestar/src/network/discovery/discv5.ts
```diff
@@ -0,0 +1,75 @@
...
+    options.bootEnrs.forEach((bootEnr) => this.discv5.addEnr(bootEnr));
+  }
+
+  async start(): Promise<void> {
+    if (this.started) {
+      return;
+    }
+    this.started = true;
+    await this.discv5.start();
+    setTimeout(() => this.findPeers, 1);
```
mpetrunic: Why wait 1ms before finding peers?

packages/lodestar/src/network/discovery/discv5.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+    this.started = false;
+    await this.discv5.stop();
+  }
+
+  async findPeers(): Promise<void> {
+    while (!this.started) {
+      // Search for random nodes
+      // emit discovered on all finds
+      const enrs = await this.discv5.findNode(createNodeId(randombytes(32)));
```
mpetrunic: Have you considered async iterable here so it yields everytime it find enr? :smile: 
#asynciterableseverywhere

packages/lodestar/src/network/discovery/discv5.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+export interface IDiscv5DiscoveryOptions extends IDiscv5DiscoveryInputOptions {
+  peerInfo: PeerInfo;
+}
+
+/**
+ * Discv5Discovery
+ */
+export class Discv5Discovery extends EventEmitter {
```
mpetrunic: I think we should release this class as libp2p-discv5 or inside @chainsafe/discv5 along with instructions how to add to libp2p stack?

packages/lodestar/src/network/discovery/discv5.ts
wemeetagain: So discv5 has a different event `discovered` that gets emitted for any enrs it finds, but what is returned here is a little different, we're only getting back the closest 16 peers that we were able to connect to as a result of the lookup.

Unfortunately, I don't think we can make it an async iterable.

packages/lodestar/src/network/discovery/discv5.ts
wemeetagain: :+1:

packages/lodestar/src/network/discovery/discv5.ts
wemeetagain: We don't need to, just was using the timeout to run asyncronously. 
Will replace this line with `this.findPeers();`

packages/lodestar/test/unit/network/util.ts
```diff
@@ -9,7 +8,7 @@ export async function createNode(multiaddr: string): Promise<NodejsNode> {
   const peerId = await createPeerId();
   const enr = ENR.createFromPeerId(peerId);
   const randomPort = Math.round(Math.random() * 40000) + 1000;
-  const bindAddr = Multiaddr(`/ip4/127.0.0.1/udp/${randomPort}`);
+  const bindAddr = `/ip4/127.0.0.1/udp/${randomPort}`;
```
mpetrunic, line 11: Not sure if works here, but usually if you put port "0" it will take random available port

mpetrunic (dismissed):

wemeetagain:
ready for re-review

mpetrunic (approved):

PR #777, opened 2020-04-06
Block stream api

mpetrunic:
Implements api for validator to listen on new blocks. I accidentally committed bug-fix in block process validate (prevented us from processing blocks in epoch 0) so this is included here.

Currently is not very pretty:
async iterable -> transform(mannually converting to sse message format)->convert to stream-> fastify

Proposed upgrade would be to create fastify plugin (which will decorate reply object) with additional `sse` method which would accept AsyncIterable<{id: string; type?: string; data: string}> and convert to stream along with appropriate headers. There is outdated plugin for that: 
https://github.com/lolo32/fastify-sse


mpetrunic:
SSE examplanation and example: https://alligator.io/nodejs/server-sent-events-build-realtime-app/

wemeetagain (dismissed):
LGTM
Agreed, it would be nice to use a fastify plugin eventually, this looks good for now.

wemeetagain (approved):

PR #783, opened 2020-04-08
Add fork digest

### LINKED ISSUE #751: Add fork digest to gossipsub topics for simple fork versioning of messages and to "Status" message
https://github.com/ethereum/eth2.0-specs/pull/1652

twoeths:
## Goal
+ Resolve #751

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/util/domain.ts
+++ b/packages/lodestar-beacon-state-transition/src/util/domain.ts
@@ -5,24 +5,37 @@ import {
   Epoch,
   Version,
   BeaconState,
+  Root,
 } from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 
-import {DomainType} from "../constants";
+import {DomainType, ZERO_HASH} from "../constants";
 import {intToBytes} from "@chainsafe/lodestar-utils";
 
 import {getCurrentEpoch} from "./epoch";
+import {computeForkDataRoot} from "./fork";
 
 /**
  * Return the domain for the [[domainType]] and [[forkVersion]].
  */
-export function computeDomain(config: IBeaconConfig, domainType: DomainType, forkVersion?: Version): Buffer {
+export function computeDomain(
+  config: IBeaconConfig,
+  domainType: DomainType,
+  forkVersion?: Version,
+  genesisValidatorRoot: Root = ZERO_HASH): Buffer {
   if (!forkVersion) {
     forkVersion = config.params.GENESIS_FORK_VERSION;
   }
+  const forkDataRoot = computeForkDataRoot(config, forkVersion, genesisValidatorRoot);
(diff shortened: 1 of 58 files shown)


packages/lodestar-beacon-state-transition/src/util/domain.ts
```diff
@@ -5,24 +5,37 @@ import {
...
+export function computeDomain(
+  config: IBeaconConfig,
+  domainType: DomainType,
+  forkVersion?: Version,
+  genesisValidatorRoot: Root = ZERO_HASH): Buffer {
   if (!forkVersion) {
     forkVersion = config.params.GENESIS_FORK_VERSION;
   }
+  const forkDataRoot = computeForkDataRoot(config, forkVersion, genesisValidatorRoot);
+  const forkDataRootForDomain = Buffer.alloc(28);
```
twoeths: The spec is just like this `fork_data_root[:28]`. Do we want to add `spice` or similar methods to our ssz `ArrayLike`?

wemeetagain: It would be nice to add, might be tricky to get the types right.
Until then we can do `(forkDataRoot.valueOf() as Uint8Array).slice(0,28)`

packages/lodestar/src/api/impl/beacon/beacon.ts
```diff
@@ -46,7 +46,8 @@ export class BeaconApi implements IBeaconApi {
     };
     return {
       fork,
-      chainId: networkId
+      chainId: networkId,
+      genesisValidatorsRoot: state.genesisValidatorsRoot,
```
twoeths, line 50: We need to enhance Etth2 API `/node/fork` in order to compute domain correctly in validator `block/attestation` service.
Does it look ok, any better way to get `genesisValidatorsRoot` to the validator?

wemeetagain: I think this is fine for now

packages/lodestar-beacon-state-transition/src/util/domain.ts
twoeths: thanks, changed it accordingly

wemeetagain (approved):

packages/lodestar-validator/src/services/block.ts
```diff
@@ -76,7 +78,7 @@ export default class BlockProposingService {
     const epoch = computeEpochAtSlot(this.config, slot);
     const randaoDomain = getDomain(
       this.config,
-      {fork} as BeaconState,
+      {fork, genesisValidatorsRoot} as BeaconState,
```
mpetrunic, line 81: Might be cool to finaly fix getDomain to require `Pick<BeaconState, "fork"|"genesisValidatorsRoot">` instead of whole BeaconState

mpetrunic (approved):

Issue #746, opened 2020-03-24
Add subnet validations for DoS resistance


ColinSchwarz:
https://github.com/ethereum/eth2.0-specs/pull/1615

PR #805, opened 2020-04-12
Validator api update

### LINKED ISSUE #778: Standardize validator api
https://github.com/ethereum/eth2.0-APIs/pull/15

mpetrunic:
resolves #778


packages/lodestar-cli/src/commands/dev/command.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * @module cli/commands
  */
-
+process.setMaxListeners(15);
```
wemeetagain, line 4: why?

packages/lodestar-types/src/ssz/generators/api.ts
```diff
@@ -15,4 +15,29 @@ export const ForkResponse = (ssz: IBeaconSSZTypes): ContainerType => new Contain
...
+    startingBlock: ssz.Uint64,
+    currentBlock: ssz.Uint64,
+    highestBlock: ssz.Uint64,
+  },
+});
+
+export const ValidatorResponse = (ssz: IBeaconSSZTypes): ContainerType => new ContainerType({
+  fields: {
+    ...ssz.Validator.fields,
+    index: ssz.ValidatorIndex
```
wemeetagain: `index` should be declared before `...Validator.fields`

packages/lodestar-validator/src/api/impl/rest/beacon/beacon.ts
```diff
@@ -1,14 +1,36 @@
...
-    this.client = new HttpClient({urlPrefix: `${restUrl}/node`}, {logger});
+  private readonly config: IBeaconConfig;
+
+  public constructor(config: IBeaconConfig, restUrl: string, logger: ILogger) {
+    this.client = new HttpClient({urlPrefix: `${restUrl}/validator`}, {logger});
+    this.config = config;
+  }
+
+
+  public async getValidator(pubkey: BLSPubkey): Promise<ValidatorResponse|null> {
```
wemeetagain, line 30: Why do we actually need the `ValidatorResponse` type?
The api seems to be designed to NOT involve `ValidatorIndex` (anywhere other than this addition), but instead rely on passing `BLSPubkey` to identify the validator.

wemeetagain: I see, this is a general beacon API. Can we use the same shape lighthouse is using? https://lighthouse-book.sigmaprime.io/http_beacon.html#beaconvalidators

packages/lodestar-validator/src/api/interface/beacon.ts
```diff
@@ -15,6 +15,11 @@ export interface IBeaconApi {
      */
   getFork(): Promise<{fork: Fork; chainId: Uint64}>;
 
+  /**
+     * Requests the BeaconNode to provide which fork version it is currently on.
```
wemeetagain: comment needs update

packages/lodestar-validator/src/api/interface/validators.ts
```diff
@@ -1,22 +1,22 @@
...
-  getProposerDuties(epoch: Epoch): Promise<Map<Slot, BLSPubkey>>;
+  getProposerDuties(epoch: Epoch, validatorPubKeys: BLSPubkey[]): Promise<Map<Slot, BLSPubkey>>;
 
-  getAttesterDuties(epoch: Epoch, validatorPubKey: BLSPubkey[]): Promise<ValidatorDuty[]>;
-
-  isAggregator(slot: Slot, committeeIndex: CommitteeIndex, slotSignature: BLSSignature): Promise<boolean>;
+  getAttesterDuties(epoch: Epoch, validatorPubKeys: BLSPubkey[]): Promise<AttesterDuty[]>;
 
   /**
    * Requests a BeaconNode to produce a valid block,
```
wemeetagain, line 22: You might rebase off 0.11.x branch, there's a change to `produceBlock` there, which currently adds a BLSPubkey to identify the proposer. Definitely suitable for validatorIndex, too.

packages/lodestar-validator/src/api/interface/validators.ts
```diff
@@ -45,11 +45,13 @@ export interface IValidatorApi {
   publishAttestation(attestation: Attestation): Promise<void>;
 
   publishAggregatedAttestation(
-    aggregated: Attestation, validatorPubKey: BLSPubkey, slotSignature: BLSSignature
```
wemeetagain, line 48: I'm uneasy about using ValidatorIndex here (instead of BLSPubkey) but not changing it everywhere else.

If the validator has access to its ValidatorIndex, the entire API can likely be reworked rely on it (instead of mapping from BLSPubkey).
VC  -> getValidator(pubkey) -> BN
VC  <- {validatorIndex} <- BN
VC -> subscribeToEvents(validatorIndex) -> BN
VC <- {events} <- BN
VC -> publishData(data[, validatorIndex]) -> BN

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -138,39 +139,16 @@ export class AttestationService {
...
-        duty.attestationSlot,
-        fork
-      )
+      {
+        aggregate: aggregatedAttestation,
+        selectionProof: this.getSlotSignature(
+          duty.attestationSlot,
+          fork
+        ),
+        aggregatorIndex: (await (this.provider.beacon.getValidator(this.publicKey))).index
```
wemeetagain: The index should likely be fetched and stored prior to this. It can be retrieved once, it will never change.

packages/lodestar-cli/src/commands/dev/command.ts
mpetrunic: to remove that warning about max listeners. It's because we have 11 libraries that are subscribed on "unhandledRejection" event in process

packages/lodestar-types/src/ssz/generators/api.ts
mpetrunic: any particular reason?

packages/lodestar-validator/src/api/impl/rest/beacon/beacon.ts
mpetrunic: > Why do we actually need the `ValidatorResponse` type?
> The api seems to be designed to NOT involve `ValidatorIndex` (anywhere other than this addition), but instead rely on passing `BLSPubkey` to identify the validator.

I was hoping we will stick with pubkey as identifier. But to construct AggregateAndProof, we need index which is a bit of a pickle.

> I see, this is a general beacon API. Can we use the same shape lighthouse is using? https://lighthouse-book.sigmaprime.io/http_beacon.html#beaconvalidators

I like that response format but I api is not really rest like:
- GET /validators?pubkeys=[]&stateRoot=
- GET /validators/{pubkey}
would be more aligned with rest conventions

packages/lodestar-validator/src/services/attestation.ts
mpetrunic: I agree but I would wait with such optimization until we have proper ways to deal with validator status(we might not be able to fetch index on start if validator deposit isn't yet detected)

packages/lodestar-validator/src/api/interface/validators.ts
mpetrunic: I see, this is a general beacon API. Can we use the same shape lighthouse is using? https://lighthouse-book.sigmaprime.io/http_beacon.html#beaconvalidators

packages/lodestar-validator/src/services/attestation.ts
twoeths: maybe another stream api? A validator UI would need it to display correct status.

packages/lodestar-validator/src/api/impl/rest/validator/validator.ts
```diff
@@ -37,25 +40,23 @@ export class RestValidatorApi implements IValidatorApi {
...
-    aggregatedAttestation: Attestation,
-    validatorPubkey: BLSPubkey,
-    slotSignature: BLSSignature
+    aggregate: AggregateAndProof,
   ): Promise<void> {
     return this.client.post(
-      `/aggregate?validator_pubkey=${toHexString(validatorPubkey)}&slot_signature=${toHexString(slotSignature)}`,
-      this.config.types.Attestation.toJson(aggregatedAttestation)
+      "/aggregate_and_proof",
+      [this.config.types.AggregateAndProof.toJson(aggregate)]
```
twoeths, line 51: I'm curious about the reason why the api wants an array here instead of a single AggregateAndProof

twoeths: probably because one may want to run multiple validators in a single client?

packages/lodestar-validator/src/api/interface/validators.ts
twoeths: yeah I need it for #746 as well, please rebase to 0.11.x if possible

packages/lodestar/src/api/rest/routes/beacon/validator.ts
```diff
@@ -0,0 +1,39 @@
...
+            properties: {
+                "pubkey": {
+                    type: "string"
+                }
+            }
+        },
+    }
+};
+
+export const registerGenesisTimeEndpoint: LodestarRestApiEndpoint = (server, {api, config}): void => {
```
twoeths: registerGetValidatorEndpoint?

packages/lodestar-validator/src/api/interface/validators.ts
mpetrunic: Sorry I don't get how this api is related to #746 or 0.11.x? in 0.11.x produceBlock method will just add proposer index to BeaconBlock model, it doesn't require validator input on this as it knows index because it knows proposer on given slot?

packages/lodestar-validator/src/services/attestation.ts
mpetrunic: Yes, but instead of introducing our own statuses and api, I would wait till we have some agreement on that

packages/lodestar-validator/src/api/impl/rest/validator/validator.ts
mpetrunic: yeah, afaik lighthouse is relying heavily on batch requests

packages/lodestar-validator/src/api/interface/validators.ts
twoeths: in #746, I need to create `AggregateAndProof` and `SignedAggregateAndProof` from the validator side. In order to do that I need to issue a REST call to get aggregator index which is available in this PR.

packages/lodestar-validator/src/api/impl/rest/beacon/beacon.ts
wemeetagain: :+1:  Yeah I was more just talking about the response format

packages/lodestar-types/src/ssz/generators/api.ts
wemeetagain: The order is important for ssz serialization

mpetrunic:
@wemeetagain @tuyennhv I would rather merge this here and open another pr to merge master into 0.11 branch as this PR doesn't work without existing changes on the master not present in 0.11 branch

twoeths:
@mpetrunic  that's totally fine to me

packages/lodestar-validator/src/api/impl/rest/beacon/beacon.ts
twoeths, line 30: As we create `AggregateAndProof` from the server side, my understanding is we don't need to call this from the validator side for now. It's just to conform to the Eth2.0 API?

mpetrunic: yes

packages/lodestar/src/api/rest/routes/validator/publishAggregateAndProof.ts
```diff
@@ -1,48 +1,32 @@
...
-          fromHexString(request.query.slot_signature)
-        );
-      } catch (e) {
-        reply.code(500).send();
-        return;
-      }
+      await Promise.all(
+        request.body.map((payload) => {
+          api.validator.publishAggregateAndProof(
+            config.types.AggregateAndProof.fromJson(payload)
```
twoeths, line 26: Looks like in 0.11, the validator publishes `SignedAggregateAndProof` instead of `AggregateAndProof`. The other endpoint to produce `AggregateAndProof` is the same.

packages/lodestar/src/opPool/opPool.ts
```diff
@@ -82,35 +78,7 @@ export class OpPool extends EventEmitter {
       this.depositData.removeOld(signedBlock.message.body.eth1Data.depositCount),
       this.proposerSlashings.remove(signedBlock.message.body.proposerSlashings),
       this.attesterSlashings.remove(signedBlock.message.body.attesterSlashings),
-      this.aggregateAndProofs.removeIncluded(signedBlock.message.body.attestations),
-      this.checkDuplicateProposer(signedBlock)
+      this.aggregateAndProofs.removeIncluded(signedBlock.message.body.attestations)
     ]);
   }
-
```
twoeths, line 89: why do we remove the duplicate check here, any replacement for that?

mpetrunic: It's invalid check that made validator not propose twice in same epoch(which is case when you have small amount of validators), proposerSlashing is only if validator signed two blocks for sam e slot. Also this check should be added to validate method of block processing so it checks blocks from network

packages/lodestar/src/api/rest/routes/validator/publishAggregateAndProof.ts
mpetrunic: yes

twoeths (approved):

wemeetagain (approved):
LGTM

PR #806, opened 2020-04-13
Add missing pieces after we run state transition

twoeths:
## Goal
+ Just found out we probably miss some operations after running state transition.

### DIFF
--- a/packages/lodestar/src/chain/attestation.ts
+++ b/packages/lodestar/src/chain/attestation.ts
@@ -69,6 +69,13 @@ export class AttestationProcessor implements IAttestationProcessor {
   }
 
   public async receiveBlock(signedBlock: SignedBeaconBlock): Promise<void> {
+    // process block's attestations
+    const promises: Promise<void>[] = [];
+    signedBlock.message.body.attestations.forEach((attestation: Attestation) => {
+      promises.push(this.receiveAttestation(attestation));
+    });
+    await Promise.all(promises);
+    // process pending attestations due to this block
     const blockRoot = this.config.types.BeaconBlock.hashTreeRoot(signedBlock.message);
     const blockPendingAttestations = this.pendingAttestations.get(toHexString(blockRoot)) ||
       new Map<AttestationRootHex, Attestation>();
--- a/packages/lodestar/src/chain/attestation.ts
+++ b/packages/lodestar/src/chain/attestation.ts
@@ -70,11 +70,8 @@ export class AttestationProcessor implements IAttestationProcessor {
 
   public async receiveBlock(signedBlock: SignedBeaconBlock): Promise<void> {
     // process block's attestations
-    const promises: Promise<void>[] = [];
-    signedBlock.message.body.attestations.forEach((attestation: Attestation) => {
-      promises.push(this.receiveAttestation(attestation));
-    });
-    await Promise.all(promises);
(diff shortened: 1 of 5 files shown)


packages/lodestar/src/chain/blocks/post.ts
```diff
@@ -1,18 +1,27 @@
...
 
 export function postProcess(
-  config: IBeaconConfig, db: IBeaconDb, logger: ILogger, metrics: IBeaconMetrics, eventBus: ChainEventEmitter
+  config: IBeaconConfig, db: IBeaconDb, logger: ILogger, metrics: IBeaconMetrics, eventBus: ChainEventEmitter,
+  opPool: OpPool, attestationProcessor: IAttestationProcessor
 ): (source: AsyncIterable<{preState: BeaconState; postState: BeaconState; block: SignedBeaconBlock}>) => Promise<void> {
   return async (source) => {
     return (async function() {
       for await(const item of source) {
+        await opPool.processBlockOperations(item.block);
```
mpetrunic, line 18: Oh yeah, not sure how I've forgot about this

packages/lodestar/src/chain/blocks/post.ts
```diff
@@ -1,18 +1,27 @@
...
+        const promises: Promise<void>[] = [];
+        item.block.message.body.attestations.forEach((attestation: Attestation) => {
+          promises.push(attestationProcessor.receiveAttestation(attestation));
+        });
```
mpetrunic: this should be handled inside `await attestationProcessor.receiveBlock(item.block);`

mpetrunic (changes requested):

packages/lodestar/src/chain/attestation.ts
```diff
@@ -69,6 +69,13 @@ export class AttestationProcessor implements IAttestationProcessor {
   }
 
   public async receiveBlock(signedBlock: SignedBeaconBlock): Promise<void> {
+    // process block's attestations
+    const promises: Promise<void>[] = [];
+    signedBlock.message.body.attestations.forEach((attestation: Attestation) => {
+      promises.push(this.receiveAttestation(attestation));
+    });
+    await Promise.all(promises);
```
wemeetagain: Can we clean this up to:
```typescript
await Promise.all(attestations.map((a) => this.receiveAttestation(a)));
```

wemeetagain (changes requested):

packages/lodestar/src/chain/attestation.ts
```diff
@@ -69,6 +69,10 @@ export class AttestationProcessor implements IAttestationProcessor {
   }
 
   public async receiveBlock(signedBlock: SignedBeaconBlock): Promise<void> {
+    // process block's attestations
+    const attestations = signedBlock.message.body.attestations.valueOf() as Attestation[];
+    await Promise.all(attestations.map((a) => this.receiveAttestation(a)));
```
twoeths, line 74: I have to do `.valueOf() as Attestation[]` because `.map` is not available in `List<Attestation>`. Is it ok, any better way?

wemeetagain: This is good

wemeetagain (approved):

mpetrunic (approved):

PR #807, opened 2020-04-14
Add metadata and ping

### LINKED ISSUE #795: [0.11.1] Add Ping, MetaData, and GetMetaData to the protocol
ethereum/eth2.0-specs#1684

wemeetagain:
Resolves #795

### DIFF
--- a/packages/lodestar/src/network/reqResp.ts
+++ b/packages/lodestar/src/network/reqResp.ts
@@ -12,6 +12,8 @@ import {
   ResponseBody,
   SignedBeaconBlock,
   Status,
+  Ping,
+  Metadata,
 } from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {Method, ReqRespEncoding, RequestId, RESP_TIMEOUT, RpcErrorCode, TTFB_TIMEOUT,} from "../constants";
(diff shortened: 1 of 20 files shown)


packages/lodestar/src/network/reqResp.ts
```diff
@@ -157,7 +168,7 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
   }
 
   private getResponse = (peerId: PeerId, method: Method, data: Buffer): AsyncIterable<ResponseChunk> => {
-    const request = this.encoder.decodeRequest(method, data);
+    const request = method === Method.Metadata ? undefined : this.encoder.decodeRequest(method, data);
```
mpetrunic, line 171: Maybe we can add check formissing data in `decodeRequest`?

packages/lodestar/src/network/reqResp.ts
```diff
@@ -148,6 +156,9 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
     const getResponse = this.getResponse;
     return (source: AsyncIterable<Buffer|BufferList>) => {
       return (async function * () {
+        if (method === Method.Metadata) {
+          yield* getResponse(peerId, method, undefined);
+        }
```
mpetrunic, line 161: Maybe we should introduce better way of handling methods without response

wemeetagain: Definitely agree we should revisit the request/response logic to better handle methods without request / response.
I'd propose we add an issue and tackle this in a separate PR.

mpetrunic: :+1:

mpetrunic (approved):

packages/lodestar/src/network/reqResp.ts
wemeetagain: https://github.com/ChainSafe/lodestar/issues/808

Issue #726, opened 2020-03-23
Clean up "beacon" cli


wemeetagain:
The beacon CLI (eg: `bin/lodestar beacon`) is the primary entryplace for running a single beacon node. Currently, it is not working.

It should, at minimum, be able to run and connect (via bootnode or local network discovery) to another lodestar node (using `bin/lodestar dev`).

This is in preparation for connecting to other client nodes.

twoeths:
Do we need to move validation part from `lodestar-cli` to `lodestar` using https://github.com/hapijs/joi in this issue?

wemeetagain:
no, I think there is even more low hanging fruit than that.
Eg: there is no libp2p instance created in the beacon command yet.

PR #809, opened 2020-04-16
Clean up Beacon CLI

twoeths:
## Goal
+ Part of #726 
+ First step is to make it work. Next steps are to figure out what we want to do to improve it
## Test
+ I can make 64 deposits to form genesis state
image

### DIFF
--- a/packages/lodestar-cli/src/commands/beacon.ts
+++ b/packages/lodestar-cli/src/commands/beacon.ts
@@ -5,7 +5,6 @@
 import {ICliCommand} from "./interface";
 import {CommanderStatic} from "commander";
 import deepmerge from "deepmerge";
-
 import {config as mainnetConfig} from "@chainsafe/lodestar-config/lib/presets/mainnet";
 import {config as minimalConfig} from "@chainsafe/lodestar-config/lib/presets/minimal";
 import {ILogger, LogLevel, WinstonLogger} from "@chainsafe/lodestar-utils/lib/logger";
@@ -14,11 +13,18 @@ import {IBeaconNodeOptions} from "@chainsafe/lodestar/lib/node/options";
 import {generateCommanderOptions, optionsToConfig} from "../util";
 import {BeaconNodeOptions} from "../lodestar/node/options";
 import {getTomlConfig} from "../lodestar/util/file";
+import {createNodeJsLibp2p} from "@chainsafe/lodestar/lib/network/nodejs";
+import {ENR} from "@chainsafe/discv5";
+import {getPeerId} from "./dev/utils";
+import {initBLS} from "@chainsafe/bls";
 
 interface IBeaconCommandOptions {
   [key: string]: string;
   configFile?: string;
+  preset?: string;
   loggingLevel?: string;
+  peerId?: string;
+  eth1BlockNum?: string;
 }
 
 export class BeaconNodeCommand implements ICliCommand {
@@ -33,6 +39,8 @@ export class BeaconNodeCommand implements ICliCommand {
       .command("beacon")
       .description("Start lodestar node")
       .option("-c, --configFile [config_file]", "Config file path")
+      .option("-p, --preset [preset]", "Minimal/mainnet", "minimal")
+      .option("--peer-id [peerId]", "peer id hex string or json file path")
       .action(async (options) => {
         // library is not awaiting this method so don't allow error propagation
         // (unhandled promise rejections)
@@ -45,27 +53,36 @@ export class BeaconNodeCommand implements ICliCommand {
     generateCommanderOptions(command, BeaconNodeOptions);
   }
 
-  public async action(options: IBeaconCommandOptions, logger: ILogger): Promise<void> {
-    let conf: Partial<IBeaconNodeOptions> = {};
+  public async action(cmdOptions: IBeaconCommandOptions, logger: ILogger): Promise<void> {
+    let nodeOptions: Partial<IBeaconNodeOptions> = {};
+    //find better place for this once this cli is refactored
+    await initBLS();
 
-    if (options.loggingLevel) {
+    if (cmdOptions.loggingLevel) {
       // eslint-disable-next-line no-undef
       // @ts-ignore
-      logger.setLogLevel(LogLevel[options.loggingLevel]);
+      logger.setLogLevel(LogLevel[cmdOptions.loggingLevel]);
     }
 
     //merge config file
-    if (options.configFile) {
-      const parsedConfig = getTomlConfig(options.configFile, BeaconNodeOptions);
+    if (cmdOptions.configFile) {
+      const parsedConfig = getTomlConfig(cmdOptions.configFile, BeaconNodeOptions);
       //cli will override toml config options
-      conf = deepmerge(conf, parsedConfig) as Partial<IBeaconNodeOptions>;
+      nodeOptions = deepmerge(nodeOptions, parsedConfig) as Partial<IBeaconNodeOptions>;
     }
     //override current config with cli config
-    conf = deepmerge(conf, optionsToConfig(options, BeaconNodeOptions));
-
-    const config = options.preset === "minimal" ? minimalConfig : mainnetConfig;
-
-    this.node = new BeaconNode(conf, {config, logger});
+    nodeOptions = deepmerge(nodeOptions, optionsToConfig(cmdOptions, BeaconNodeOptions));
+    const peerId = await getPeerId(cmdOptions.peerId);
+    const defaultDiscv5Opt = {
+      enr: ENR.createFromPeerId(peerId),
+      bindAddr: "/ip4/127.0.0.1/udp/0",
+      bootEnrs: [] as ENR[]};
+    const discv5 = nodeOptions.network? Object.assign(defaultDiscv5Opt, nodeOptions.network.discv5) : defaultDiscv5Opt;
+    const libp2pOpt = nodeOptions.network? Object.assign(nodeOptions.network, {discv5}) : {discv5};
+    const libp2p = await createNodeJsLibp2p(peerId, libp2pOpt);
+    const config = cmdOptions.preset === "minimal" ? minimalConfig : mainnetConfig;
+    // nodejs will create EthersEth1Notifier by default
+    this.node = new BeaconNode(nodeOptions, {config, logger, libp2p});
     await this.node.start();
   }
 }
--- a/packages/lodestar-cli/src/commands/beacon.ts
+++ b/packages/lodestar-cli/src/commands/beacon.ts
@@ -13,17 +13,16 @@ import {IBeaconNodeOptions} from "@chainsafe/lodestar/lib/node/options";
 import {generateCommanderOptions, optionsToConfig} from "../util";
 import {BeaconNodeOptions} from "../lodestar/node/options";
 import {getTomlConfig} from "../lodestar/util/file";
-import {createNodeJsLibp2p} from "@chainsafe/lodestar/lib/network/nodejs";
+import {createNodeJsLibp2p, loadPeerIdFromJsonFile} from "@chainsafe/lodestar/lib/network/nodejs";
 import {ENR} from "@chainsafe/discv5";
-import {getPeerId} from "./dev/utils";
 import {initBLS} from "@chainsafe/bls";
 
 interface IBeaconCommandOptions {
   [key: string]: string;
+  peerId: string;
   configFile?: string;
   preset?: string;
   loggingLevel?: string;
-  peerId?: string;
   eth1BlockNum?: string;
 }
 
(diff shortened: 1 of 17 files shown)


packages/lodestar-cli/package.json
```diff
@@ -23,7 +23,7 @@
     "lint:fix": "eslint --color --ext .ts src/ --fix",
     "pretest": "yarn run check-types",
     "test:unit": "TS_NODE_PROJECT=tsconfig.test.json nyc --cache-dir .nyc_output/.cache -e .ts mocha --file ./test/setup.ts --colors -r ts-node/register 'test/unit/**/*.test.ts'",
-    "test:e2e": "TS_NODE_PROJECT=tsconfig.test.json mocha --file ./test/setup.ts --colors --timeout 5000 -r ts-node/register 'test/e2e/**/*.test.ts'",
+    "test:e2e": "TS_NODE_PROJECT=tsconfig.test.json mocha --file ./test/setup.ts --colors --timeout 5000 --exit -r ts-node/register 'test/e2e/**/*.test.ts'",
```
twoeths: I failed to figure out why e2e does not stop, that's something related to libp2p because if I comment it out, it works.
Probably we leave it like this and resolve it after we migrate to next libp2p?

packages/lodestar-cli/test/e2e/commands/beacon.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    };
+    const cmd = new BeaconNodeCommand();
+    await cmd.action(cmdOptions, logger);
+    logger.verbose("cmd.action started node successfully");
+    await new Promise((resolve) => setTimeout(resolve, 10 * config.params.SECONDS_PER_SLOT * 1000));
+    await cmd.node.stop();
+    logger.verbose("cmd.stop stopped node successfully");
+  });
+
+  it("should work with a dev node thru '--bootnodes' until the blocks found thru init & regular sync", async function() {
```
twoeths: This test is to make sure `beacon` command works with a `dev` node

packages/lodestar-cli/src/commands/beacon.ts
```diff
@@ -45,27 +60,51 @@ export class BeaconNodeCommand implements ICliCommand {
...
     }
     //override current config with cli config
-    conf = deepmerge(conf, optionsToConfig(options, BeaconNodeOptions));
+    nodeOptions = deepmerge(nodeOptions, optionsToConfig(cmdOptions, BeaconNodeOptions));
+    const peerId = await getPeerId(cmdOptions.peerId);
+    const defaultDiscv5Opt = {
+      enr: ENR.createFromPeerId(peerId),
+      bindAddr: "/ip4/0.0.0.0/udp/5501",
+      bootEnrs: [] as ENR[]};
+    const discv5 = nodeOptions.network? Object.assign(defaultDiscv5Opt, nodeOptions.network.discv5) : defaultDiscv5Opt;
```
mpetrunic, line 79: Why isn't this config inline with other configurations?

packages/lodestar-cli/src/commands/beacon.ts
```diff
@@ -45,27 +60,51 @@ export class BeaconNodeCommand implements ICliCommand {
...
+    const libp2p = await createNodeJsLibp2p(peerId, libp2pOpt);
 
-    const config = options.preset === "minimal" ? minimalConfig : mainnetConfig;
+    const config = cmdOptions.preset === "minimal" ? minimalConfig : mainnetConfig;
+    // deposit takes time to be come active validator
+    config.params.MIN_GENESIS_ACTIVE_VALIDATOR_COUNT = 0;
 
-    this.node = new BeaconNode(conf, {config, logger});
+    // nodejs will create EthersEth1Notifier by default
+    const eth1 = cmdOptions.eth1 === "interop" ? new InteropEth1Notifier() : undefined;
```
mpetrunic: I don't think beacon command should use interop eth1 we have dev command for that

packages/lodestar-cli/src/commands/beacon.ts
```diff
@@ -45,27 +60,51 @@ export class BeaconNodeCommand implements ICliCommand {
...
+    this.node = new BeaconNode(nodeOptions, {config, logger, eth1, libp2p});
+    const depositDataRootList = config.types.DepositDataRootList.tree.defaultValue();
+    const genesisStateFilePath = cmdOptions.genesisState;
+    if (genesisStateFilePath) {
+      // genesis state from file
+      const genesisState = quickStartOptionToState(config, depositDataRootList, genesisStateFilePath);
+      await this.node.initFromKnownState(genesisState, depositDataRootList);
```
mpetrunic: I don't think we should support quick start params in beacon command (we now need genesisValidatorsRoot, genesis fork etc), it might be better if we just support starting from eth1 + genesis fork version and from ssz state.

packages/lodestar-cli/src/commands/dev/command.ts
```diff
@@ -88,12 +89,14 @@ export class DevCommand implements ICliCommand {
...
 
-    conf.network.discv5 = Object.assign(
-      {},
-      {enr: ENR.createFromPeerId(peerId), bindAddr: "/ip4/0.0.0.0/udp/5501"},
-      conf.network.discv5,
-    );
-    const libp2p = await createNodeJsLibp2p(peerId, conf.network);
+    const defaultDiscv5Opt = {
+      enr: ENR.createFromPeerId(peerId),
+      bindAddr: "/ip4/0.0.0.0/udp/5501",
```
mpetrunic, line 94:
```suggestion
      bindAddr: "/ip4/127.0.0.1/udp/0",
```
should bind on random port and so commands in readme still works.

mpetrunic (commented):
Tests are not exiting because some server/connection stayed open. I think we should investigate this here instead of masking with `--exit` flag

packages/lodestar-cli/package.json
wemeetagain: libp2p should already be updated to the version that awaits discovery starting on libp2p start.
Not sure why its not exiting

packages/lodestar-cli/src/commands/beacon.ts
```diff
@@ -33,6 +43,11 @@ export class BeaconNodeCommand implements ICliCommand {
       .command("beacon")
       .description("Start lodestar node")
       .option("-c, --configFile [config_file]", "Config file path")
+      .option("-p, --preset [preset]", "Minimal/mainnet", "minimal")
+      .option("-s, --genesisState [params]", "Start chain from known state")
```
wemeetagain: We will likely not want to be starting from a known genesis state but rather start with an eth1 provider and eth1 deposit contract address and use the eth1 chain and the deposit contract to determine when/what the genesis state is.

packages/lodestar-cli/src/commands/beacon.ts
```diff
@@ -33,6 +43,11 @@ export class BeaconNodeCommand implements ICliCommand {
...
+      .option("--peer-id [peerId]", "peer id hex string or json file path")
```
wemeetagain: we will likely only support the json file path option here (we need the private key as well as the public key)

packages/lodestar-cli/src/commands/beacon.ts
```diff
@@ -33,6 +43,11 @@ export class BeaconNodeCommand implements ICliCommand {
...
+      .option("-e, --eth1", "interop for using fake eth1", "chain")
```
wemeetagain: This 'beacon' command is the main entry point for running a production beacon chain, we likely will want to keep the interop-related commands on the 'dev' command.

packages/lodestar-cli/src/commands/beacon.ts
```diff
@@ -33,6 +43,11 @@ export class BeaconNodeCommand implements ICliCommand {
...
+      .option("-b, --eth1BlockNum", "eth1 block number to create genesis state")
```
wemeetagain: We should be watching the eth1 chain to determine the right eth1 block number

packages/lodestar-cli/src/commands/beacon.ts
wemeetagain: I think because we need the PeerId to create the local ENR if it isn't already specified.

packages/lodestar-cli/src/commands/beacon.ts
```diff
@@ -45,27 +60,51 @@ export class BeaconNodeCommand implements ICliCommand {
...
+      bindAddr: "/ip4/0.0.0.0/udp/5501",
+      bootEnrs: [] as ENR[]};
+    const discv5 = nodeOptions.network? Object.assign(defaultDiscv5Opt, nodeOptions.network.discv5) : defaultDiscv5Opt;
+    const libp2pOpt = nodeOptions.network? Object.assign(nodeOptions.network, {discv5}) : {discv5};
+    const libp2p = await createNodeJsLibp2p(peerId, libp2pOpt);
 
-    const config = options.preset === "minimal" ? minimalConfig : mainnetConfig;
+    const config = cmdOptions.preset === "minimal" ? minimalConfig : mainnetConfig;
+    // deposit takes time to be come active validator
+    config.params.MIN_GENESIS_ACTIVE_VALIDATOR_COUNT = 0;
```
wemeetagain: This doesn't seem right. We shouldn't be editing the params.

wemeetagain (changes requested):

packages/lodestar-cli/src/commands/beacon.ts
twoeths: Yeah I'll keep the watching code to look for the 1st eth1 block forming genesis state. Should we still keep this option too or not? 
After genesis time, if we want to launch a node we may need to specify the exact eth1 block number. Or in this case the genesis state of the new node is determined by `initial sync` to other peers?

packages/lodestar-cli/src/commands/beacon.ts
twoeths: sorry, I thought this is test file :D. Tried to find a way to decrease the min genesis active validator for my e2e test. Probably I'll add a param for it.

packages/lodestar-cli/src/commands/beacon.ts
wemeetagain: After genesis time, we should still be able to scan eth1 block by block in the same way to determine the genesis state. I'm in favor of removing this option.

packages/lodestar-cli/src/commands/beacon.ts
mpetrunic: Aren't we suppose to modify enr in runtime according to subnets, how do we do that?

twoeths:
Since we should not use `genesisState` for beacon node, I cannot see the issue anymore. I created #812 to track the issue.

packages/lodestar-cli/src/commands/beacon.ts
twoeths: so I just need to modify the option description or what else I have to do?

packages/lodestar-cli/src/commands/beacon.ts
wemeetagain: the ENR gets passed into the network in the constructor. As the network subscribes / unsubscribes to subnets and fork versions change, the ENR can get updated. We should ideally be creating the whole network identity (peer Id and ENR) up front, and storing it to disk, or reading it from disk.

packages/lodestar-cli/src/commands/beacon.ts
wemeetagain: instead of using `getPeerId`, L75, use `loadPeerIdFromJsonFile`

packages/lodestar/src/chain/chain.ts
```diff
@@ -159,20 +154,44 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
     const depositDataRootList = this.config.types.DepositDataRootList.tree.defaultValue();
-    depositDataRootList.push(...depositDatas.map(this.config.types.DepositData.hashTreeRoot));
     const tree = depositDataRootList.tree();
 
     const genesisState = initializeBeaconStateFromEth1(
       this.config,
       fromHexString(eth1Block.hash),
       eth1Block.timestamp,
       depositDatas.map((data, index) => {
+        depositDataRootList.push(this.config.types.DepositData.hashTreeRoot(data));
```
twoeths, line 205: I can't verify deposit without this change (although the old/current looks more reasonable to me). I copied the way to get proof from https://github.com/ChainSafe/lodestar/blob/0.11.x/packages/lodestar-cli/src/lodestar/interop/deposits.ts#L33

any ideas?

twoeths: found it. That's because of the way we process deposits in `initializeBeaconStateFromEth1` so the proof should be calculated accordingly.

wemeetagain:
It looks like this has a lot of changes from the 0.11.x branch.
I think we can merge 0.11.x into master now that it is passing all tests?

twoeths:
ah.. I think I should not rebase against 0.11.x as this should go to master

wemeetagain (dismissed):
:+1:

wemeetagain (approved):

PR #810, opened 2020-04-16
Fix duplicate peers

mpetrunic:
Noticed after we added discv5, libp2p  emits peer:connected for peer found as bootnodes as well as peer found using some discovery method. I've switched this to listen on peers added to peerStore which filters duplicates.

### DIFF
--- a/packages/lodestar/src/network/interface.ts
+++ b/packages/lodestar/src/network/interface.ts
@@ -52,7 +52,7 @@ export interface IReqResp extends ReqEventEmitter {
 // network
 
 export interface INetworkEvents {
-  ["peer:connect"]: (peerInfo: PeerInfo) => void;
+  ["peer:connect"]: (peerInfo: PeerInfo, direction: "inbound"|"outbound") => void;
   ["peer:disconnect"]: (peerInfo: PeerInfo) => void;
 }
 export type NetworkEventEmitter = StrictEventEmitter<EventEmitter, INetworkEvents>;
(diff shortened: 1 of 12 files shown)


packages/lodestar/src/sync/reqResp/reqResp.ts
```diff
@@ -64,12 +65,18 @@ export class SyncReqResp implements ISyncReqResp {
   }
 
   public async start(): Promise<void> {
+    // TODO: enable after we add peerbook persisting
```
wemeetagain: what does this mean?

mpetrunic: Well curently, when this code is run we don't have any peers since peerbook is wiped clean onstartup. I guess in future we can store peers in database or on disk and when starting we can try to reconnect them and fetch their latestStatus.

I should probably uncomment this as it's gonna work as soon as we have that! :+1:

wemeetagain (dismissed):

wemeetagain:
So whats strange to me is that the other libp2p discovery modules don't check for duplicates either.
It also looks like the libp2p core logic is supposed to handle duplicate connections from discovered peers (when autodial is on):
https://github.com/libp2p/js-libp2p/blob/master/src/index.js#L451

mpetrunic:
Yeah, I saw that also, it might be that discovery emits before we mannually dial

packages/lodestar/src/network/interface.ts
```diff
@@ -52,7 +52,7 @@ export interface IReqResp extends ReqEventEmitter {
 // network
 
 export interface INetworkEvents {
-  ["peer:connect"]: (peerInfo: PeerInfo) => void;
+  ["peer:connect"]: (peerInfo: PeerInfo, direction: "inbound"|"outbound") => void;
```
wemeetagain, line 55: :pray:

wemeetagain (approved):

PR #816, opened 2020-04-20
Refactor req/resp to better handle different method types

### LINKED ISSUE #808: Refactor req/resp to better handle different method types
The current network request/response code has several hack-ish checks on the Method to handle:
- Methods with no request data
- Methods with no response data
- Methods with only a single response chunk

twoeths:
## Goal
+ resolves #808

### DIFF
--- a/packages/lodestar/src/constants/network.ts
+++ b/packages/lodestar/src/constants/network.ts
@@ -16,6 +16,11 @@ export enum Method {
   BeaconBlocksByRoot = "beacon_blocks_by_root",
 }
 
+// Methods that returns 1 single response_chunk
+export const SINGLE_CHUNK_METHODS = [Method.Status, Method.Ping, Method.Metadata];
+// Request only method
+export const NO_CHUNK_METHODS = [Method.Goodbye];
+
 export enum ReqRespEncoding {
   SSZ = "ssz",
   SSZ_SNAPPY = "ssz_snappy",
--- a/packages/lodestar/src/constants/network.ts
+++ b/packages/lodestar/src/constants/network.ts
@@ -1,5 +1,8 @@
 // gossip
 
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
+import {Type} from "@chainsafe/ssz";
+
 export const ATTESTATION_SUBNET_COUNT = 64;
 export const ATTESTATION_PROPAGATION_SLOT_RANGE = 23;
 export const MAXIMUM_GOSSIP_CLOCK_DISPARITY = 500;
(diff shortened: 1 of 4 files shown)


packages/lodestar/src/network/reqResp.ts
```diff
@@ -156,19 +157,19 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
...
+          break;
         }
+        yield* getResponse(peerId, method, data);
       })();
     };
   }
 
   private getResponse = (peerId: PeerId, method: Method, data: Buffer): AsyncIterable<ResponseChunk> => {
-    const request = method === Method.Metadata ? undefined : this.encoder.decodeRequest(method, data);
+    const request = data ? this.encoder.decodeRequest(method, data) : undefined;
```
mpetrunic: Maybe it's better to handle this inside decodeRequest

packages/lodestar/src/network/reqResp.ts
```diff
@@ -156,19 +157,19 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
...
-        }
+        let data: Buffer;
+        // source is an array of 1 item or empty array
         for await (const val of source) {
-          const data = Buffer.isBuffer(val) ? val : val.slice();
-          yield* getResponse(peerId, method, data);
+          data = Buffer.isBuffer(val) ? val : val.slice();
+          break;
         }
+        yield* getResponse(peerId, method, data);
```
mpetrunic, line 171: You could put this before break

packages/lodestar/src/constants/network.ts
```diff
@@ -16,6 +16,11 @@ export enum Method {
   BeaconBlocksByRoot = "beacon_blocks_by_root",
 }
 
+// Methods that returns 1 single response_chunk
+export const SINGLE_CHUNK_METHODS = [Method.Status, Method.Ping, Method.Metadata];
+// Request only method
+export const NO_CHUNK_METHODS = [Method.Goodbye];
```
mpetrunic: I don't feel like those arrays are best way forward, as it's quite easy to forget to add it.
We could modify Method enum to contain more informations.
Like:
```javascript
export const Methods = {
   Method.Status: {
      responseType: MethodResponseType.SingleRespone
   },
   Method.Ping: {
      responseType: MethodResponseType.NoResponse
   },
   Method.BlocksByRange: {
      responseType: MethodResponseType.Stream
   }
}
```
Maybe ti makes more sense to add more stuff later like `maxRequestPerSecond` etc

mpetrunic: this could also hold ssz types for request and response so we can remove those stuff from encoders

mpetrunic (approved):

packages/lodestar/src/constants/network.ts
```diff
@@ -16,6 +19,46 @@ export enum Method {
   BeaconBlocksByRoot = "beacon_blocks_by_root",
 }
 
+export enum MethodResponseType {
+  SingleRespone = "SingleRespone",
```
wemeetagain, line 23: Typo

wemeetagain (dismissed):
One typo, otherwise lgtm

packages/lodestar/src/constants/network.ts
mpetrunic: I already fixed that in #841 which is based on this branch so we can merge this

PR #818, opened 2020-04-20
Add ENR shard bitfield

### LINKED ISSUE #642: add ENR attestation subnet (attnets) bitfield
v0.9.4: (ethereum/eth2.0-specs#1534)

twoeths:
## Goals
+ resolves #642 
+ This is our debt in v0.9.4
+ Fix gossip attestation subnet topic as 1 GossipEvent maps to multiple attestation subnet topics
+ Add gossip.test.ts unit test


packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -139,14 +143,24 @@ export class ValidatorApi implements IValidatorApi {
...
         epoch
       );
     });
+    const subnets: Set<string> = new Set();
+    attesterDuties.forEach(duty => subnets.add(getCommitteeIndexSubnet(duty.committeeIndex)));
+    subnets.forEach(subnet => {
+      const peerIds = this.reps.getPeerIdsBySubnet(subnet);
+      if (peerIds.length < 3) {
+        // TODO: If an insufficient number of current peers are subscribed to the topic,
+        // the validator must discover new peers on this topic
```
twoeths: @wemeetagain how can I use discv5 to discover new peer given a subnet?

wemeetagain: So discv5 doesn't have topic discovery yet, but we can fetch the ENRs for all the peers we've found, and check the attnet value.
```typescript
Promise.all(
discv5
  .kadValues()
  .filter((enr: ENR) => enr.get("attnet"))
  .filter((enr: ENR) => config.types.AttestationSubnets.deserialize(enr.get("attnet"))[subnet])
  .map((enr: ENR) => enr.peerId().then((peerId) => {
    const peerInfo = new PeerInfo(peerId);
    peerInfo.multiaddrs.add(enr.multiaddrTCP);
    return peerInfo;
  })
)
```

mpetrunic: If we already discovered them, we don't need to search it as they will send us stuff after we subscribe to topic?

wemeetagain: we may have discovered peers, but the libp2p machinery didn't dial them (because max number of peers reached)

mpetrunic: How do we decide what peers to disconnect to add new ones? :smile:

wemeetagain: I think libp2p would help figure that out.
They're planning on maintaining some sort of score for each peer in the peer store.
We would help feed the scores with data about success / fail network requests or good gossip.

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -139,14 +143,26 @@ export class ValidatorApi implements IValidatorApi {
...
         {publicKey: state.validators[validatorIndex].pubkey, index: validatorIndex},
         state,
         epoch
       );
     });
+    const subnets: Set<string> = new Set();
+    attesterDuties.forEach(duty => subnets.add(getCommitteeIndexSubnet(duty.committeeIndex)));
+    for (const subnet of subnets) {
+      const peerIds = this.reps.getPeerIdsBySubnet(subnet);
+      if (peerIds.length < 3) {
```
twoeths: The spec does not specify a constant, I think we should have a configuration variable here. Any ideas?

packages/lodestar/src/network/network.ts
```diff
@@ -99,6 +100,39 @@ export class Libp2pNetwork extends (EventEmitter as { new(): NetworkEventEmitter
...
+    const peerInfos = peerInfosForSubnet.filter(peerInfo => !peerIds.includes(peerInfo.id.toB58String()));
+    // make sure they still connect to same subnet
+    let count = 0;
+    for (const peerInfo of peerInfos) {
+      // we'll dial thru sendRequest so don't need to do connect(peerInfo) like in the spec
+      const metadata = await this.reqResp.metadata(peerInfo);
+      if (metadata.attnets[subnet]) {
+        count++;
+      }
+      if (count < 10) {
```
twoeths, line 144: Do we need to limit the max peer to connect per subnet like this? If yes, need to specify a number.

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -139,14 +143,26 @@ export class ValidatorApi implements IValidatorApi {
...
+        // If an insufficient number of current peers are subscribed to the topic,
+        // the validator must discover new peers on this topic
+        this.logger.info(`Found only ${peerIds.length} for subnett ${subnet}, finding new peers to connect`);
+        await this.network.connectToNewPeersBySubnet(parseInt(subnet), peerIds);
+      }
+    }
```
mpetrunic: Don't do that here, we should start looking only if validator is aggregator which he will notify us by calling "subscribeAttestationSubnet" method.

packages/lodestar/src/network/network.ts
```diff
@@ -99,6 +100,39 @@ export class Libp2pNetwork extends (EventEmitter as { new(): NetworkEventEmitter
     await this.libp2p.hangUp(peerInfo);
   }
 
+  // peerIds: peers already have this subnet
+  public async connectToNewPeersBySubnet(subnet: number, inPeerIds?: string[]): Promise<number> {
+    const peerIds = inPeerIds || [];
+    // @ts-ignore
+    const discovery: Discv5Discovery = this.libp2p._discovery.get("discv5");
```
mpetrunic: add to libp2p-types

packages/lodestar/src/network/network.ts
```diff
@@ -99,6 +100,39 @@ export class Libp2pNetwork extends (EventEmitter as { new(): NetworkEventEmitter
     await this.libp2p.hangUp(peerInfo);
   }
 
+  // peerIds: peers already have this subnet
```
mpetrunic: convert to proper jsdoc

packages/lodestar/src/network/network.ts
```diff
@@ -99,6 +100,39 @@ export class Libp2pNetwork extends (EventEmitter as { new(): NetworkEventEmitter
...
+  public async connectToNewPeersBySubnet(subnet: number, inPeerIds?: string[]): Promise<number> {
+    const peerIds = inPeerIds || [];
+    // @ts-ignore
+    const discovery: Discv5Discovery = this.libp2p._discovery.get("discv5");
+    const discv5: Discv5 = discovery.discv5;
+    const peerInfosForSubnet: PeerInfo[] = await Promise.all(
+      discv5.kadValues()
+        .filter((enr: ENR) => enr.get("attnets"))
+        .filter((enr: ENR) => this.config.types.AttestationSubnets.deserialize(enr.get("attnets"))[subnet])
```
mpetrunic: we cannot control what peers will put here, so we should catch malformed deserialization errors

packages/lodestar/src/network/network.ts
```diff
@@ -99,6 +100,39 @@ export class Libp2pNetwork extends (EventEmitter as { new(): NetworkEventEmitter
...
+          const peerInfo = new PeerInfo(peerId);
+          peerInfo.multiaddrs.add(enr.multiaddrTCP);
+          return peerInfo;
+        })));
+    const peerInfos = peerInfosForSubnet.filter(peerInfo => !peerIds.includes(peerInfo.id.toB58String()));
+    // make sure they still connect to same subnet
+    let count = 0;
+    for (const peerInfo of peerInfos) {
+      // we'll dial thru sendRequest so don't need to do connect(peerInfo) like in the spec
+      const metadata = await this.reqResp.metadata(peerInfo);
```
mpetrunic: I wonder if metadata request comes in before status will peer reject us? In which case, should we catch that error here.

packages/lodestar/src/sync/IReputation.ts
```diff
@@ -33,4 +34,21 @@ export class ReputationStore {
...
+    if (!new RegExp("^\\d+$").test(subnetStr)) {
+      throw new Error(`Invalid subnet ${subnetStr}`);
+    }
+    const subnet = parseInt(subnetStr);
+    if (subnet < 0 || subnet >= ATTESTATION_SUBNET_COUNT) {
+      throw new Error(`Invalid subnet ${subnetStr}`);
+    }
+    const peerIds = [];
+    for (const [peerId, rep] of this.reputations) {
+      if (rep.latestMetadata.attnets[subnet]) {
```
mpetrunic: check if latestMetadata exists

packages/lodestar/src/tasks/tasks/interopSubnetsJoiningTask.ts
```diff
@@ -25,9 +25,12 @@ export class InteropSubnetsJoiningTask implements ITask {
 
   public async run(): Promise<void> {
     const forkDigest = this.chain.currentForkDigest;
+    const attNets: AttestationSubnets = this.config.types.AttestationSubnets.defaultValue();
     for (let i = 0; i < this.config.params.RANDOM_SUBNETS_PER_VALIDATOR; i++) {
-      this.subscribeToRandomSubnet(forkDigest);
+      const subnet = this.subscribeToRandomSubnet(forkDigest);
+      attNets[subnet] = true;
     }
+    this.network.metadata.attnets = attNets;
```
mpetrunic: I wonder if it would be better to do this in single place like in:
- this.network.gossip.subscribeToAttestationSubnet
- this.network.gossip.unsubscribeFromAttestationSubnet

packages/lodestar/test/e2e/network/network.test.ts
```diff
@@ -55,8 +60,11 @@ describe("[network] network", function () {
       state,
       config
     });
-    netA = new Libp2pNetwork(opts, {config, libp2p: createNode(multiaddr) as unknown as Libp2p, logger, metrics, validator, chain});
-    netB = new Libp2pNetwork(opts, {config, libp2p: createNode(multiaddr) as unknown as Libp2p, logger, metrics, validator, chain});
+    peerIdB = await createPeerId();
+    libP2pA = await createNode(multiaddr) as unknown as Libp2p;
+    libP2pB = await createNode(multiaddr, peerIdB) as unknown as Libp2p;
```
mpetrunic: Promise.all both nodes

packages/lodestar/test/e2e/network/network.test.ts
```diff
@@ -106,18 +114,13 @@ describe("[network] network", function () {
...
     await new Promise((resolve) => netB.gossip.once("gossipsub:heartbeat", resolve));
     validator.isValidIncomingBlock.resolves(true);
     const block = generateEmptySignedBlock();
     block.message.slot = 2020;
-    for (let i = 0; i < 5; i++) {
-      await netB.gossip.publishBlock(block);
-    }
-    await received;
+    await Promise.all(new Array(5).fill(netB.gossip.publishBlock(block)));
+    await new Promise(resolve => setTimeout(resolve, 2000));
```
mpetrunic: I don't like solution to just wait 2s, I think I've fixed this in #810

packages/lodestar/test/e2e/network/util.ts
```diff
@@ -0,0 +1,15 @@
+import {ENR, Discv5Discovery} from "@chainsafe/discv5";
+import PeerId from "peer-id";
+
+import {NodejsNode} from "../../../src/network/nodejs";
+import {createPeerId, initializePeerInfo} from "../../../src/network";
+import defaults from "../../../src/network/options";
+
+export async function createNode(multiaddr: string, inPeerId?: PeerId): Promise<NodejsNode> {
```
mpetrunic: How is this new function, we already had this

mpetrunic (changes requested):

packages/lodestar/test/e2e/network/util.ts
twoeths: the other `createNode` is in unit test, this one is for e2e.
The differenc is I want an optional peer id and use only `Discv5Discovery` as discovery method.

mpetrunic: can you extract it to some common utils so we don't duplicate. Any reason why using only discv5?

twoeths: yeah will do. For this e2e test, I want to make sure nodeA to connect to nodeB thru discv5 after we find by subnet. If we use `MDNS`, it'll connect to each other automatically.

packages/lodestar/src/network/network.ts
twoeths: done

packages/lodestar/src/tasks/tasks/interopSubnetsJoiningTask.ts
twoeths: yeah I did that

packages/lodestar/src/network/gossip/gossip.ts
```diff
@@ -123,15 +128,59 @@ export class Gossip extends (EventEmitter as { new(): GossipEventEmitter }) impl
     subnet: number|string,
     callback?: (attestation: {attestation: Attestation; subnet: number}) => void
   ): void {
-    this.subscribe(forkDigest, GossipEvent.ATTESTATION_SUBNET, callback, new Map([["subnet", subnet.toString()]]));
```
twoeths, line 126: Can't use the common `subscribe()` method here because 1 GossipEvent.ATTESTATION_SUBNET maps to multiple gossip topics.
Use the same strategy with new `attNetListenerCount`

packages/lodestar-cli/src/commands/dev/command.ts
```diff
@@ -146,7 +147,8 @@ export class DevCommand implements ICliCommand {
       logger: new WinstonLogger({module: "API"}),
       chain: node.chain,
       network: node.network,
-      db: node.db
+      db: node.db,
+      reps: new ReputationStore()
```
mpetrunic: I'm pretty sure this is useless as you need reputation store from beacon node itself

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -141,14 +145,15 @@ export class ValidatorApi implements IValidatorApi {
...
-    return validatorIndexes.map((validatorIndex) => {
+    const attesterDuties = validatorIndexes.map((validatorIndex) => {
       return assembleAttesterDuty(
         this.config,
         {publicKey: state.validators[validatorIndex].pubkey, index: validatorIndex},
         state,
         epoch
       );
     });
+    return attesterDuties;
```
mpetrunic: Why not returning inline?

packages/lodestar/src/network/gossip/gossip.ts
mpetrunic: I think common subscribe method works fine here.
generic subscribe method create topic from GossipEvent.ATTESTATION_SUBNET, forkDigest and subnet so it should work as expected. we have unique topic for each subnet

packages/lodestar/src/network/gossip/gossip.ts
```diff
@@ -123,15 +128,59 @@ export class Gossip extends (EventEmitter as { new(): GossipEventEmitter }) impl
...
+    const count = this.attNetListenerCount.get(topic);
+    if (count === 0) {
+      this.pubsub.subscribe(topic);
+    }
+    this.attNetListenerCount.set(topic, count + 1);
+    this.on(GossipEvent.ATTESTATION_SUBNET, ({attestation, subnet}: {attestation: Attestation; subnet: number}) => {
+      if (subnet === subnetNum && callback) {
+        callback({attestation, subnet});
+      }
+    });
```
mpetrunic: Why did you copy code from "subscribe" here

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -221,6 +226,14 @@ export class ValidatorApi implements IValidatorApi {
...
       committeeIndex
     );
+    const subnet = getCommitteeIndexSubnet(committeeIndex);
+    const peerIds = this.reps.getPeerIdsBySubnet(subnet);
+    if (peerIds.length < 3) {
+      // If an insufficient number of current peers are subscribed to the topic,
+      // the validator must discover new peers on this topic
+      this.logger.info(`Found only ${peerIds.length} for subnett ${subnet}, finding new peers to connect`);
+      await this.network.connectToNewPeersBySubnet(parseInt(subnet), peerIds);
+    }
```
mpetrunic: I feel like this whole logic should be moved to network. Like we should just call 
```
this.network.searchSubnetPeers(subnet)
```
here

packages/lodestar/src/network/network.ts
```diff
@@ -103,6 +105,52 @@ export class Libp2pNetwork extends (EventEmitter as { new(): NetworkEventEmitter
...
+          } catch (err) {
+            return false;
+          }
+        })
+        .map((enr: ENR) => enr.peerId().then((peerId) => {
+          const peerInfo = new PeerInfo(peerId);
+          peerInfo.multiaddrs.add(enr.multiaddrTCP);
+          return peerInfo;
+        })));
+    const peerInfos = peerInfosForSubnet.filter(peerInfo => !peerIds.includes(peerInfo.id.toB58String()));
```
mpetrunic: Might be good idea to break this function into smaller chucnk for readability.
Like you could extract this chunk to searchDiscv5Peers or something

packages/lodestar/src/network/gossip/gossip.ts
twoeths: yeah the `subscribe` is unique but they share the same `GossipEvent.ATTESTATION_SUBNET` event. Given this code `this.on(event, listener as (...args: unknown[]) => void);` in subscribe method, if there is 1 event emitted to a topic it'll call the callback functions of all subscriptions. And we can't base on `this.listenerCount(event)` to subscribe/unsubscribe to pubsub because it's common to all subnets so I have to manage the listener count for each topic in `this.attNetListenerCount` map

mpetrunic (dismissed):

packages/lodestar/src/network/gossip/gossip.ts
mpetrunic: I see what you are saying now.
But still, you should change this code a bit to use existing EventEmmiteer, possible solution is 
`this.on(GossipEvent.ATTESTATION_SUBNET +"_" + subnet, listener)`

If you modify generic subscribe to accept custom string, you could pass `GossipEvent.ATTESTATION_SUBNET +"_" + subnet` instead of just `keyof IGossipEvents` so you can reuse it. Just make sure you change it gossip handler

packages/lodestar/src/network/gossip/gossip.ts
```diff
@@ -123,24 +126,39 @@ export class Gossip extends (EventEmitter as { new(): GossipEventEmitter }) impl
     subnet: number|string,
     callback?: (attestation: {attestation: Attestation; subnet: number}) => void
   ): void {
-    this.subscribe(forkDigest, GossipEvent.ATTESTATION_SUBNET, callback, new Map([["subnet", subnet.toString()]]));
+    const subnetNum: number = (typeof subnet === "string")? parseInt(subnet) : subnet as number;
+    this.subscribe(forkDigest, subnetNum, callback, new Map([["subnet", subnet.toString()]]));
```
mpetrunic:
```suggestion
    this.subscribe(forkDigest, GossipEvent.ATTESTATION_SUBNET + "_" +subnetNum, callback, new Map([["subnet", subnet.toString()]]));
```
because event name as just number is weird

packages/lodestar/src/network/gossip/handlers/attestation.ts
```diff
@@ -30,7 +30,7 @@ export function getCommitteeAttestationHandler(subnet: number): GossipHandlerFn
         `Received committee attestation for block ${toHexString(attestation.data.beaconBlockRoot)}`
           +`subnet: ${subnet}, (${attestation.data.source.epoch}, ${attestation.data.target.epoch})`
       );
-      this.emit(GossipEvent.ATTESTATION_SUBNET, {attestation, subnet});
+      this.emit(subnet, {attestation, subnet});
```
mpetrunic:
```suggestion
      this.emit(GossipEvent.ATTESTATION_SUBNET + "_" +subnet, {attestation, subnet});
```

packages/lodestar/src/network/gossip/gossip.ts
twoeths: I failed to do that. I don't want to declare IGossipEvents like:
```
"committee_index_beacon_attestation_0": (attestationSubnet: {attestation: Attestation; subnet: number}) => void;
"committee_index_beacon_attestation_1": (attestationSubnet: {attestation: Attestation; subnet: number}) => void;
...
62 more
```
If I simply do `[key: string]: ...` I got compilation error due to conflict to other events so I have to do `[key:number]` to avoid the conflict.
any ideas to declare attestation subnet events in `IGossipEvents` as a string pattern or keyof something?

twoeths: the only way I can do is to have a separate event emitter for attestation subnet, please have a look.

mpetrunic: That's a pickle, maybe the best way is to "ts-ignore" where it complains since dynamic event name is not supported in strict event emitter and event that is just number is wierd as heck

packages/lodestar/src/network/gossip/interface.ts
```diff
@@ -22,14 +22,21 @@ import {IBeaconChain} from "../../chain";
...
-  [GossipEvent.ATTESTATION_SUBNET]: (attestationSubnet: {attestation: Attestation; subnet: number}) => void;
+  // @ts-ignore
   [GossipEvent.ATTESTATION]: (attestation: Attestation) => void;
+  // @ts-ignore
   [GossipEvent.AGGREGATE_AND_PROOF]: (attestation: SignedAggregateAndProof) => void;
+  // @ts-ignore
   [GossipEvent.VOLUNTARY_EXIT]: (voluntaryExit: SignedVoluntaryExit) => void;
+  // @ts-ignore
   [GossipEvent.PROPOSER_SLASHING]: (proposerSlashing: ProposerSlashing) => void;
+  // @ts-ignore
```
mpetrunic: Oh sorry, I should been more clearer, don't declare dynamic event here, just add @ts-ignore on two places we are using it

mpetrunic (approved):

PR #828, opened 2020-04-21
Make lodestar and lodestar-cli not private

wemeetagain:


mpetrunic (commented):
Are we ready to release them yet? :open_mouth:

wemeetagain:
> Are we ready to release them yet? open_mouth

Not particularly, but since we plan on it anyways, and we've moved to the single-version across the monorepo, I think it makes sense to release all of what we have.
We're clearly at a pre-1.0 state right now, so things are going to be breaking version to version.

I'd also like to set up nightly releases at some point. Making these libraries non-private will be needed for that too.

mpetrunic (approved):

PR #841, opened 2020-04-24
Rpc snappy framing compression

### LINKED ISSUE #749: Use snappy frames to support streaming
https://github.com/ethereum/eth2.0-specs/pull/1606

mpetrunic:
resolves #749


packages/lodestar/src/network/encoders/response.ts
```diff
@@ -0,0 +1,107 @@
...
+        } catch (e) {
+          logger.warn(`Failed to uncompress data for method ${method}. Error: ${e.message}`);
+        }
+        if(uncompressed) {
+          uncompressedData.append(uncompressed);
+          if(uncompressedData.length > sszLength) {
+            logger.warn(`Received too much data for method ${method}`);
+            break;
+          }
+          if(uncompressedData.length === sszLength) {
```
tuyennhv, line 88: should this be `!==` instead?

packages/lodestar/src/network/encoders/response.ts
```diff
@@ -0,0 +1,107 @@
...
+        let uncompressed: Buffer;
+        try {
+          uncompressed = decompressor.uncompress(buffer.slice());
+          buffer.consume(buffer.length);
+        } catch (e) {
+          logger.warn(`Failed to uncompress data for method ${method}. Error: ${e.message}`);
+        }
+        if(uncompressed) {
+          uncompressedData.append(uncompressed);
+          if(uncompressedData.length > sszLength) {
```
twoeths, line 84: should this be `!==` instead

packages/lodestar/src/network/encoders/response.ts
```diff
@@ -0,0 +1,107 @@
...
+            logger.warn(`Received too much data for method ${method}`);
+            break;
+          }
+          if(uncompressedData.length === sszLength) {
+            try {
+              yield type.deserialize(uncompressedData.slice()) as ResponseBody;
+              buffer = new BufferList();
```
twoeths, line 91: what's the point to reuse `buffer` and every variable except for `decompressor ` instead of just declare them inside the `for` loop?

packages/lodestar/src/network/encoders/response.ts
```diff
@@ -0,0 +1,107 @@
...
+  config: IBeaconConfig, logger: ILogger, method: Method, encoding: ReqRespEncoding
+): (source: AsyncIterable<IResponseChunk>) => AsyncIterable<Buffer> {
+  return (source => {
+    return (async function*() {
+      const type = Methods[method].responseSSZType(config);
+      let compressor = getCompressor(encoding);
+      if(!type) {
+        return;
+      }
+      for await (const chunk of source) {
```
twoeths, line 20: question: why do we yield status, ssz length and compressed data separately instead of just concatenate them in a single buffer and yield it like the previous version?

packages/lodestar/src/network/encoders/response.ts
```diff
@@ -0,0 +1,107 @@
...
+          serializedData = Buffer.from(serialized.buffer, serialized.byteOffset, serialized.length);
+        } catch (e) {
+          logger.warn(`Failed to ssz serialize chunk of method ${method}. Error: ${e.message}`); 
+        }
+        //yield encoded ssz length
+        yield Buffer.from(encode(serializedData.length));
+        //yield compressed or uncompressed data chunks
+        yield* compressor(serializedData);
+        //reset compressor
+        compressor = getCompressor(encoding);
```
twoeths, line 40: do we have `reset` method like the decompressor in order to reuse the compressor?

packages/lodestar/src/network/encoders/response.ts
```diff
@@ -0,0 +1,107 @@
+import {IResponseChunk} from "./interface";
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
+import {ILogger} from "@chainsafe/lodestar-utils/lib/logger";
+import {Method, MethodResponseType, Methods, ReqRespEncoding, RpcResponseStatus} from "../../constants";
+import {decode, encode} from "varint";
+import {encodeResponseStatus, getCompressor, getDecompressor} from "./utils";
+import BufferList from "bl";
```
twoeths, line 7: what's the reason to use `BufferList` instead of `Buffer`?

packages/lodestar/src/network/encoders/snappy-frames/uncompress.ts
```diff
@@ -0,0 +1,101 @@
+import BufferList from "bl";
+import {uncompress} from "snappyjs";
+import {IDecompressor} from "../interface";
+
+const IDENTIFIER = new Buffer([
+  0x73, 0x4e, 0x61, 0x50, 0x70, 0x59
+]);
+
+export class SnappyFramesUncompress implements IDecompressor {
```
twoeths, line 9: Looks great. Should we enhance `snappyjs` with this?

packages/lodestar/src/network/encoders/response.ts
mpetrunic: Nope, then we would break if we don't have enough data, we are suppose to keep uncompressing until we reach our sszLength

packages/lodestar/src/network/encoders/response.ts
mpetrunic: I could call consume on buffer and uncompressData to remove all buffers but I feel like this is faster as it will leave current data for gc to clean while consume might be fiddling with existing arrays. Decompressor has reset method to this cleanup. Once we deserialize data it means we reached end of chunk, and next one will be exactly the same so we just reset everything. I  should add more code comments here

packages/lodestar/src/network/encoders/response.ts
mpetrunic: Because recipient needs status and length so it knows how much frames it needs to decompress. with compression we are sending compressed data frame by frame while previously we would just send one blob of data so we didn't really need length prefix anyways

packages/lodestar/src/network/encoders/response.ts
mpetrunic: not really since compressor is just function, we can convert it to sam class later on, but we can do that once we replace snappy-stream with our implementation using snappyjs

packages/lodestar/src/network/encoders/response.ts
mpetrunic: We are appending a lot of buffers, with Buffer we would have to allocate and copy new Buffer everytime, while BufferList keeps internal array of buffers

packages/lodestar/src/network/encoders/snappy-frames/uncompress.ts
mpetrunic: I don't think this implementation will be usefull to snappyjs as most common scenario should be to just yield data everytime you uncompress something and I don't know how to incorporate async generator like that in our case. But yeah, I would love to create separate library it-snappyjs that would use async iterators to compress and decompress using framing format

packages/lodestar/src/network/encoders/response.ts
twoeths: I see, if `<` then just accumulate more and try to uncompress again

twoeths (dismissed):

twoeths:
need to resolve conflicted files

twoeths (approved):

wemeetagain (approved):
lgtm

packages/lodestar/src/network/encoders/response.ts
dapplion: @mpetrunic yes please! Comments would be very helpful since there are many non-obvious design choices that are really good and should be preserved if the code is re-worked

packages/lodestar/src/network/encoders/response.ts
dapplion: @mpetrunic Should we also use BufferList in eth2RequestDecode?

PR #842, opened 2020-04-26
lodecli beacon init / run

wemeetagain:


jbrady42 (dismissed):
Looks good.
:+1:

packages/lodestar-cli/src/lodecli/cmds/beacon/cmds/run/run.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+export async function run(options: Arguments<IBeaconArgs & Partial<IBeaconNodeOptions>>): Promise<void> {
+  await initBLS();
+
+  options = deepmerge(defaultOptions, options) as Arguments<IBeaconArgs & Partial<IBeaconNodeOptions>>;
+
+  const peerId = await readPeerId(options.network.peerIdPath);
+  options.network.discv5.enr = await readEnr(options.network.enrPath);
+
+  const config = options.chain.name === "mainnet" ? mainnetConfig : mainnetConfig;
```
mpetrunic: chain.name? maybe chain.spec is better?

packages/lodestar-cli/src/lodecli/cmds/beacon/cmds/run/run.ts
```diff
@@ -0,0 +1,43 @@
...
+import {WinstonLogger} from "@chainsafe/lodestar-utils/lib/logger";
+import defaultOptions, {IBeaconNodeOptions} from "@chainsafe/lodestar/lib/node/options";
+
+import {readPeerId, readEnr, writeEnr} from "../../../../network";
+import {IBeaconArgs} from "../../options";
+
+/**
+ * Run a beacon node
+ */
+export async function run(options: Arguments<IBeaconArgs & Partial<IBeaconNodeOptions>>): Promise<void> {
```
mpetrunic, line 20: Maybe we could have something similar to lighthouse:
```terminal
lodecli run bn //to run beacon node
lodecli run vn //to run validator
etc
```

packages/lodestar-cli/src/lodecli/cmds/beacon/cmds/run/run.ts
```diff
@@ -0,0 +1,43 @@
...
+    logger,
+  });
+
+  async function cleanup(): Promise<void> {
+    await node.stop();
+    await writeEnr(options.network.enrPath, options.network.discv5.enr, peerId);
+  }
+
+  process.on("SIGTERM", cleanup);
+  process.on("SIGINT", cleanup);
```
mpetrunic, line 47: Maybe we should use https://github.com/jtlapp/node-cleanup/blob/master/node-cleanup.js as it handles all signals and exceptions

packages/lodestar-cli/src/lodecli/cmds/beacon/options/api.ts
```diff
@@ -0,0 +1,52 @@
+import {Options} from "yargs";
+
+export const apiEnabled: Options = {
+  alias: [
+    "api.enabled",
+    "api.rest.enabled",
```
mpetrunic: maybe also just "rest"

packages/lodestar-cli/src/lodecli/cmds/beacon/options/api.ts
```diff
@@ -0,0 +1,52 @@
...
+  ],
+  type: "boolean",
+  default: false,
```
mpetrunic: It would be nice if it could be just "--rest" which would mean enable rest api

packages/lodestar-cli/src/lodecli/cmds/beacon/options/eth1.ts
```diff
@@ -0,0 +1,41 @@
...
+  group: "eth1",
+};
+
+export const eth1DepositContractAddress: Options = {
+  alias: [
+    "eth1.depositContract.address",
+  ],
+  description: "Address of deposit contract",
+  type: "string",
+  default: "TBD",
```
mpetrunic: if default is null will flag be required?

packages/lodestar-cli/src/lodecli/cmds/beacon/cmds/run/run.ts
wemeetagain: I believe we can add command aliases, will play with some of that

packages/lodestar-cli/src/lodecli/cmds/beacon/options/eth1.ts
wemeetagain: yeah, flag or config would be required

packages/lodestar-cli/src/lodecli/cmds/beacon/options/api.ts
wemeetagain: It should be documented somewhere, but with these aliases, the first alias is the preferred name, (ideally consistently named across all options), the second alias is probably the option that `BeaconNode` expects as part of `IBeaconNodeOptions`.

Not opposed to adding more "convenience" aliases on top tho

packages/lodestar-cli/src/lodecli/cmds/beacon/cmds/run/run.ts
wemeetagain: This is from `IBeaconNodeOptions`. (Would love to change that tho, `name` is not a good.... name)

Currently named `chain.preset` on the CLI level.
(run `lodecli beacon init` and look at `.lodecli/beacon/beacon.config.json` to see the default config)

twoeths:
Looks great. Do we want to start moving some e2e tests to lodecli?

wemeetagain:
@tuyennhv Yes! I'll add the beacon e2e test. (and update lodecli to include the work in #843)

wemeetagain:
I added some sanity unit tests, but the beacon node is currently not stable enough to run in a meaningful e2e test. I'll make an issue for this

mpetrunic (approved):

PR #843, opened 2020-04-27
Add ENRForkID to ENR eth2

### LINKED ISSUE #750: Add "eth2" key/value to ENR
https://github.com/ethereum/eth2.0-specs/pull/1614

twoeths:
## Goal
+ resolves #750 
+ provide a way to configure planned hard forks
## TODOs work for discv 5:
+ Clients SHOULD connect to peers with fork_digest, next_fork_version, and next_fork_epoch that match local values.
+ Clients MAY connect to peers with the same fork_digest but a different next_fork_version/next_fork_epoch. Unless ENRForkID is manually updated to matching prior to the earlier next_fork_epoch of the two clients, these connecting clients will be unable to successfully interact starting at the earlier next_fork_epoch

### DIFF
--- a/packages/lodestar-params/src/interface.ts
+++ b/packages/lodestar-params/src/interface.ts
@@ -70,4 +70,17 @@ export interface IBeaconParams {
   MAX_ATTESTATIONS: number;
   MAX_DEPOSITS: number;
   MAX_VOLUNTARY_EXITS: number;
+
+  // Old and future forks
+  ALL_FORKS: IFork[];
 }
+
+interface IFork {
+  // 4 bytes
+  previousVersion: number;
+  // 4 bytes
+  currentVersion: number;
+  // Fork epoch number
+  epoch: number;
+}
+
--- a/packages/lodestar-beacon-state-transition/src/epoch/fork.ts
+++ b/packages/lodestar-beacon-state-transition/src/epoch/fork.ts
@@ -1,0 +1,19 @@
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
+import {BeaconState} from "@chainsafe/lodestar-types";
+import {getCurrentEpoch} from "..";
+import {intToBytes} from "@chainsafe/lodestar-utils";
+
+export function processForkChanged(config: IBeaconConfig, state: BeaconState): void {
+  const currentEpoch = getCurrentEpoch(config, state);
+  const nextEpoch = currentEpoch + 1;
+  const currentForkVersion = state.fork.currentVersion;
+  const nextFork = config.params.ALL_FORKS && config.params.ALL_FORKS.find(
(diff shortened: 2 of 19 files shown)


packages/lodestar-params/src/interface.ts
```diff
@@ -70,4 +70,17 @@ export interface IBeaconParams {
   MAX_ATTESTATIONS: number;
   MAX_DEPOSITS: number;
   MAX_VOLUNTARY_EXITS: number;
+
```
twoeths, line 73: I define ALL_FORKS here and load it from yaml file in lodestar-cli, is it ok? I want it to be loaded in 1 single space and be able to process it in multiple modules in `config.params.ALL_FORKS`. Maybe for historical hard forks we want to store it in database, just load the future ones.

wemeetagain: This is fine

wemeetagain (approved):
:+1: LGTM

mpetrunic (approved):
Curious, if we have scheduled hard forks that changes state transition, how do we handle multiple state transition dependencies. Can you even do that? :smile:

twoeths:
@mpetrunic what do you mean `state transition dependencies`, can you tell me more about that?

mpetrunic:
> @mpetrunic what do you mean `state transition dependencies`, can you tell me more about that?

If there is some breaking change in state transition and we have to schedule new hard fork. When you sync from genesis again you have to run with old state transition until hard fork then you need to switch to new beacon state transition I guess

twoeths:
it should work fine because the `ALL_FORKS` array contains all the hard fork history as well as future planned hard fork.
At epoch boundaries, it checks for any fork with `previousVersion = state.fork.currentVersion` and at the right `epoch` then it'll use the new fork for the new state.
So keeping `ALL_FORKS` up to date and contain all historical forks are very important in order to launch a new node.

mpetrunic:
> it should work fine because the `ALL_FORKS` array contains all the hard fork history as well as future planned hard fork.
> At epoch boundaries, it checks for any fork with `previousVersion = state.fork.currentVersion` and at the right `epoch` then it'll use the new fork for the new state.
> So keeping `ALL_FORKS` up to date and contain all historical forks are very important in order to launch a new node.

You didn't get my point, it's not related to this PR per se. So you are syncing chain that went trough hardfork.
```
if fork.currentVersion === FORK_VERSION_WHEN_STATE_TRANSITION_v1
   import processBlock from "@chainsafe/beacon-state-transition@v1"
   processBlock(block)
if fork.currentVersion === FORK_VERSION_WHEN_STATE_TRANSITION_v2
   import processBlock from "@chainsafe/beacon-state-transition@v2"
   processBlock(block)
```
I wonder if it's even possible to do something like that

twoeths:
ah... I can't come up with something like that atm ...

Issue #855, opened 2020-04-28
Fixup Eth1 Processing


wemeetagain:
Several issues with current approach:
- db - duplicatative deposit data lists stored
- pregenesis - duplicate work every eth1 block
- pregenesis - processing recent eth1 blocks before checking historical blocks

Ideally:
- db - store deposit roots indexed by depositIndex, maintain a single depositDataRootList tree
- pregenesis - maintain processed deposits, don't reprocess deposits every eth1 block
- pregenesis - process historical eth1 blocks before processing recent eth1 blocks

Issue #724, opened 2020-03-19
Consider data persistence / Caching strategy


wemeetagain:
We should more thoughtfully consider which pieces of data will be:
- in memory
- in a 'hot' database
- in a 'cold' database

PR #858, opened 2020-04-30
Fixup db

### LINKED ISSUE #685: Avoid using "any" in DB controller
Need to audit the usage of DB controller to have type for it, it can be `UInt8Array | Buffer | string` or something ..

wemeetagain:
- Resolves #685 
- Exposes more of the level interface
  - exposes `gt`, `gte`, `lt`, `lte`, `reverse`, `limit` when filtering (vs hard-coded gt and lt)
  - filter by `keys`, `values`, or `entries` (both), as streaming or as a promise
- split bad block into its own `Repository`


packages/lodestar/src/chain/blocks/process.ts
```diff
@@ -107,13 +107,13 @@ export async function updateDepositMerkleTree(
     newState.eth1Data.depositCount - newState.eth1DepositIndex
   );
   const [depositDatas, depositDataRootList] = await Promise.all([
-    db.depositData.getAllBetween(newState.eth1DepositIndex, upperIndex),
+    db.depositData.values({gt: newState.eth1DepositIndex, lt: upperIndex}),
```
mpetrunic, line 110: db.depositData.filter would be more common to use but I don't have strong preference

packages/lodestar/src/db/api/beacon/repositories/badBlock.ts
```diff
@@ -0,0 +1,30 @@
...
+  public constructor(
+    config: IBeaconConfig,
+    db: IDatabaseController<Buffer, Buffer>,
+  ) {
+    super(config, db, Bucket.invalidBlock, config.types.Boolean);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public getId(value: boolean): Uint8Array {
+    throw new Error("Cannot get block root from boolean");
```
mpetrunic, line 24: eh?

packages/lodestar/src/db/api/beacon/repositories/block.ts
```diff
@@ -1,72 +1,46 @@
...
     return this.config.types.BeaconBlock.hashTreeRoot(value.message);
   }
 
-  public async set(id: Uint8Array, value: SignedBeaconBlock): Promise<void> {
+  public async put(id: Uint8Array, value: SignedBeaconBlock): Promise<void> {
     await Promise.all([
-      this.db.put(encodeKey(Bucket.blockSlotRefs, value.message.slot), id),
-      this.db.put(encodeKey(Bucket.blockRootRefs, id), this.config.types.Slot.serialize(value.message.slot)),
-      super.set(id, value)
+      this.db.put(encodeKey(Bucket.blockSlotRefs, value.message.slot), id as Buffer),
```
mpetrunic, line 35: Actually, in non-finalized blocks, there can be more than one block for one slot. I think we should add root to array of roots belonging to that slot. Maybe even have separate repository to handle those.

packages/lodestar/src/chain/chain.ts
```diff
@@ -140,7 +140,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
     );
     // Determine whether a genesis state already in
     // the database matches what we were provided
-    const storedGenesisBlock = await this.db.block.getBlockBySlot(GENESIS_SLOT);
+    const storedGenesisBlock = await this.db.block.getBySlot(GENESIS_SLOT);
```
mpetrunic, line 143: This could be in blockArchive if we have finalized chain stored

packages/lodestar/src/db/api/beacon/repositories/block.ts
```diff
@@ -1,72 +1,46 @@
...
-  }
-
-  public async getJustifiedBlock(): Promise<SignedBeaconBlock | null> {
-    const root = await this.chain.getJustifiedBlockRoot();
-    if(!root) return null;
-    return await this.get(root);
-  }
-
-  public async getBlockBySlot(slot: Slot): Promise<SignedBeaconBlock | null> {
+  public async getBySlot(slot: Slot): Promise<SignedBeaconBlock | null> {
```
mpetrunic, line 39: This should return array of SignedBeaconBlock, only in archive we have 1-1 mapping between slot and root

packages/lodestar/test/unit/chain/factory/block/blockAssembly.test.ts
```diff
@@ -12,28 +12,30 @@ import { StatefulDagLMDGHOST } from "../../../../../../lodestar/src/chain/forkCh
...
+    chainStub = sandbox.createStubInstance(BeaconChain) as unknown as StubbedChain;
     chainStub.forkChoice = forkChoiceStub;
 
     opPool = sandbox.createStubInstance(OpPool);
     beaconDB = {
       block: sandbox.createStubInstance(BlockRepository),
       state: sandbox.createStubInstance(StateRepository),
       depositDataRootList: sandbox.createStubInstance(DepositDataRootListRepository)
-    };
+    } as StubbedBeaconDb;
```
mpetrunic, line 38: It will probably be useful if we had:
```typescript
export class MockDb extends IBeaconDb {
   block: SinnonStubbedInstance<IBlockRepository>;
   ...

   constructor(sinon) {
      block = sinon.createStubbedInstance(BlockRepository);
     ...
   }
}
```

mpetrunic (changes requested):

packages/lodestar/src/db/api/beacon/repositories/badBlock.ts
wemeetagain: this `getId` method infers the key from a value.  (its what helps transform `add(value)` and `remove(value)` methods, into `put(key, value)`, `delete(key, value)`)
In the case of bad block, the key is the block root, the value is a boolean. So its not possible to infer the key from the value there.

packages/lodestar/src/db/api/beacon/repositories/block.ts
wemeetagain: Yeah, I would have removed it entirely but its being used right now. Trying to limit the damage of a single PR XD

packages/lodestar/src/chain/chain.ts
wemeetagain: agreed, the genesis handling is not right, but trying to limit the scope of this PR to just fix up the db controller and not change too much outside that.

packages/lodestar/src/db/api/beacon/repositories/block.ts
wemeetagain: agreed that this is not right, will tackle as part of another PR, when thinking about https://github.com/ChainSafe/lodestar/issues/724

packages/lodestar/src/chain/blocks/process.ts
wemeetagain: reason for this convention is to expose the key-value nature of `Repository`
`keys` - return [] of keys
`keysStream` return stream of keys
`values` - return [] of values
`valuesStream` return stream of values
`entries` return [] of {key, value}
`entriesStream` return stream of {key,value}

(I'm planning on using `keys` soon in #855)

packages/lodestar/test/unit/chain/factory/block/blockAssembly.test.ts
wemeetagain: Agreed, would definitely be more robust. Going to make an issue for this.

wemeetagain: #859

wemeetagain:
not sure why e2e tests are failing, they pass locally :thinking:

mpetrunic:
> not sure why e2e tests are failing, they pass locally :thinking: 

Try to run clean and build locally, although I noticed similar thing in one of my branches

mpetrunic (dismissed):

mpetrunic (approved):

PR #861, opened 2020-05-01
E2e to sync 2 dev nodes

### LINKED ISSUE #812: Lipp2p: need to close resources after use
When I work on #726, I used a `beacon` node to connect to `dev` node and start/stop them, then the test cannot exit so I have to use `--exit` flag in package.json

Since we should not connect `beacon` node to `dev` node thru shared state, I don't have this issue anymore but it's worth to investigate.

To reproduce, write an e2e test:
+ Start a dev node with 8 validators
+ Start another dev node without validator
+ Wait for blocks to be received on the 2nd node thru initial/regular sync
+ Make sure we can start/stop/sync all nodes smoothly in the test

twoeths:
# Goal
+ resolve #812

### DIFF
--- a/packages/lodestar/src/opPool/modules/attestation.ts
+++ b/packages/lodestar/src/opPool/modules/attestation.ts
@@ -1,6 +1,7 @@
 import {Attestation, BeaconState, CommitteeIndex, Epoch,} from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
-import {computeEpochAtSlot, computeStartSlotAtEpoch,} from "@chainsafe/lodestar-beacon-state-transition";
+import {computeEpochAtSlot, computeStartSlotAtEpoch,isUnaggregatedAttestation}
+  from "@chainsafe/lodestar-beacon-state-transition";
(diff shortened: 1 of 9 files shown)


twoeths:
Due to #864 , the emitted events are all `inbound` instead of `inbound` & `outbound` hence no handshake happened and dev node 2 can't sync to dev node 1.

twoeths:
It works now as I connect 2 dev nodes manually

packages/lodestar/src/opPool/modules/attestation.ts
```diff
@@ -18,7 +19,7 @@ export class AttestationOperations extends OperationsModule<Attestation> {
       return attestation.data.index === committeeIndex
           && computeEpochAtSlot(this.config, attestation.data.slot) === epoch
           //filter out aggregated attestations
-          && attestation.aggregationBits.length === 1;
+          && isUnaggregatedAttestation(attestation);
```
twoeths, line 22: need confirmation

wemeetagain: :+1:

wemeetagain (approved):

PR #862, opened 2020-05-01
Merge opPool and db

wemeetagain:
Remove `OpPool`, in favor of consolidating "op pool functionality" in the `Repository`s associated with pending operations.
More clearly distinguish the purposes of various `IBeaconDb` `Repository`s.

The `OpPool` module has always been fraught with an identity crisis.
What is its purpose? What differentiates it from other data that's stored?

In practice, the `OpPool` module acted as a subset of the `IBeaconDb` with additional methods on several of the `Repository`s.

I think this distinction between `OpPool` `Module`s and `BeaconDb` `Repository`s is not actually helpful, and the additional level of indirection is just a complicating factor to understanding the codebase. We can remove a bunch of duplicate code and simplify the code by unifying these modules.


mpetrunic (dismissed):

twoeths:
need to resolve conflicting files

twoeths (approved):

PR #865, opened 2020-05-03
Bring back shouldDisconnectOnStatus

### LINKED ISSUE #844: Get shouldDisconnectOnStatus up to date
Uncomment the following lines, make sure it gets thru sync issues and up to spec 0.11.x:
```
public async shouldDisconnectOnStatus(request: Status): Promise<boolean> {
    const currentForkDigest = this.chain.currentForkDigest;
    return !this.config.types.ForkDigest.equals(currentForkDigest, request.forkDigest);

    //TODO: fix this, doesn't work if we are starting sync(archive is empty) or we don't have finalized epoch
    // const startSlot = computeStartSlotAtEpoch(this.config, request.finalizedEpoch);
    // const startBlock = await this.db.blockArchive.get(startSlot);
    // // we're on a further (or equal) finalized epoch
    // // but the peer's block root at that epoch doesn't match ours
    // if (
    //   state.finalizedCheckpoint.epoch >= request.finalizedEpoch &&
    //   !this.config.types.Root.equals(
    //     request.finalizedRoot,
    //     this.config.types.BeaconBlock.hashTreeRoot(startBlock.message)
    //   )
    // ) {
    //   return true;
    // }
    // return false;
  }
```

twoeths:
## Goal
+ resolve #844 
+ depends on #861

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -6,7 +6,8 @@ import {EventEmitter} from "events";
 import {fromHexString, List, toHexString, TreeBacked} from "@chainsafe/ssz";
 import {Attestation, BeaconState, Root, SignedBeaconBlock, Uint16, Uint64,
   ForkDigest,
-  ENRForkID} from "@chainsafe/lodestar-types";
+  ENRForkID,
+  Checkpoint} from "@chainsafe/lodestar-types";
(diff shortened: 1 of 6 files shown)


twoeths:
Ready for review.

wemeetagain (approved):

mpetrunic (approved):

PR #868, opened 2020-05-04
Fix up eth1 processing

### LINKED ISSUE #855: Fixup Eth1 Processing

wemeetagain:
Resolves https://github.com/ChainSafe/lodestar/issues/855

`IBeaconDb` has been modified to store deposits and eth1 data.
EthersEth1Notifier now processes eth1 blocks, starting at `depositContract.deployedAt`, maintaining the follow distance, parsing deposit events and eth1 data from each block and storing in `IBeaconDb`.

new `IBeaconDb` properties:
- `depositData` - `DepositData` indexed by deposit index. Data stored pending chain processing (as part of the 'op pool'.
- `depositDataRoot` - `Root<DepositData>` indexed by deposit index. Deposit data roots are stored permanently, and a `TreeBacked` list of deposit roots (the deposit tree) is maintained in memory for fast deposit processing when creating blocks. (this tree replaces our previous `db.depositDataRootList`)
- `eth1Data - `Eth1Data` indexed by timestamp. In `EthersEth1Notifier`, because historical eth1 state is not fetchable from providers, the deposit count and deposit root is calculated, not from the deposit contract, but from the `IBeaconDb`. (this replaces the eth1 block cache we had before)

new `Repository` methods:
- `firstKey`, `firstValue`, `firstEntry`
- `lastKey`, `lastValue`, `lastEntry`

new `IBeaconTypes` type:
- `DepositEvent` - `DepositData` with index (TODO: once ssz is updated, we can replace `EthersEth1Notifier#parseDepositEvent` with `DepositEvent.fromJson(_,{case: "snake"})`)

Eth1 voting was updated:
- We no longer fetch candidate eth1data from a `IEth1Notifier` eth1 block cache
- instead, we fetch `Eth1Data`, stored in `IBeaconDb`, filtered to the candidate voting range
- code moved to `chain/factory/block`


twoeths (dismissed):

packages/lodestar-cli/src/commands/beacon.ts
```diff
@@ -19,14 +20,16 @@ import {initBLS} from "@chainsafe/bls";
...
+export interface IBeaconCommandOptions {
   [key: string]: string;
   peerId: string;
   forkFile?: string;
   configFile?: string;
   preset?: string;
   loggingLevel?: string;
   eth1BlockNum?: string;
+  // @ts-ignore
+  config?: IBeaconConfig;
```
mpetrunic, line 32: I think this is suppose to be just cli overrides, I don't think we should have this here

packages/lodestar-cli/src/commands/dev/command.ts
```diff
@@ -110,7 +111,16 @@ export class DevCommand implements ICliCommand {
...
     this.node = new BeaconNode(conf, {config, logger, eth1: new InteropEth1Notifier(), libp2p});
-    await this.node.chain.initializeBeaconChain(state, depositDataRootList);
+    const deposits = interopDeposits(
+      config, config.types.DepositDataRootList.tree.defaultValue(), parseInt(options.validatorCount),
+    );
+    for (let i = 0; i < deposits.length; i++) {
+      await Promise.all([
+        this.node.db.depositData.put(i, deposits[i].data),
+        this.node.db.depositDataRoot.put(i, config.types.DepositData.hashTreeRoot(deposits[i].data)),
+      ]);
```
mpetrunic, line 121: We could utilize batch writing here but probably out of scope for now

packages/lodestar-cli/src/commands/beacon.ts
wemeetagain: We need to have some way of overriding the config to pass the e2e beacon test (to update the minimum active validator count so we can induce genesis). In practice, won't be used. Also hoping to allow for this more gracefully in `lodecli`.

mpetrunic (approved):

PR #884, opened 2020-05-06
allow for clean exit when starting the chain without a genesis state

### LINKED ISSUE #880: Ensure beacon node cleanly exits
**Describe the bug**
When starting the chain without a genesis state, not all modules are started, since `BeaconChain` blocks on `start` until a genesis state has been created.
If the node is stopped during this time, the node is not stopped properly, because other modules' `stop` method assumes that all modules have been started.

**Expected behavior**
Beacon node should exit cleanly, even if all modules have not started.

**Steps to Reproduce**
start lodecli, using eth1 to trigger genesis, eg: using https://hackmd.io/2wNvrO77RQ6BWc2ZA7M0Cg
Once the node has started, hit Ctrl+C
Instead of exiting, an unhandled promise rejection is thrown

lastmjs:
Fixes https://github.com/ChainSafe/lodestar/issues/880

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -113,7 +113,11 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
 
   public async stop(): Promise<void> {
     await this.forkChoice.stop();
-    await this.clock.stop();
+
+    if (this.clock) {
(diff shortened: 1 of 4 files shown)


wemeetagain (approved):
LGTM, very minimal change that fixes the immediate errors in #880

PR #887, opened 2020-05-07
Fix number key database serialization

mpetrunic:
Leveldb is sorting keys lexicographic so it seems best way to keep order with number keys is to serialize them as utf8 string.

### DIFF
--- a/packages/lodestar-cli/src/commands/dev/command.ts
+++ b/packages/lodestar-cli/src/commands/dev/command.ts
@@ -96,7 +96,7 @@ export class DevCommand implements ICliCommand {
       {discv5: {enr: ENR.createFromPeerId(peerId), bindAddr: "/ip4/127.0.0.1/udp/0"}},
       {isMergeableObject: isPlainObject}
     );
-    const libp2p = await createNodeJsLibp2p(peerId, conf.network, true);
+    const libp2p = await createNodeJsLibp2p(peerId, conf.network, false);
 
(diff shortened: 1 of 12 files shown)


wemeetagain:
> Leveldb is sorting keys lexicographic so it seems best way to keep order with number keys is to serialize them as utf8 string.

I don't see why this is necessary.
Serializing ints in big-endian format has the same properties, no?

mpetrunic:
> > Leveldb is sorting keys lexicographic so it seems best way to keep order with number keys is to serialize them as utf8 string.
> 
> I don't see why this is necessary.
> Serializing ints in big-endian format has the same properties, no?

It's necessary since it doesn't work :smile: 
You can try it yourself by running blockarchive e2e test.
As for reasoning, my best guess is since if it deserializes big int serialized number to string it gets some control characters instead of numbers (ascii 0 is 48!)

wemeetagain:
I'm not sure I get it. The tests are passing on master, no?
unit testing the block archive shows its returning keys/values in order. https://github.com/ChainSafe/lodestar/pull/901

Hardening the eth1 e2e tests, on the other hand :+1: :+1:

mpetrunic:
> I'm not sure I get it. The tests are passing on master, no?
> unit testing the block archive shows its returning keys/values in order. https://github.com/ChainSafe/lodestar/pull/901
> 

We arent testing searching for number value keys anywhere. See the new blockarchive e2e tests. Its not passing on master

mpetrunic:
Will opem another PR with just eth1 test hardening

PR #896, opened 2020-05-07
Update order of logger cli aliases

wemeetagain:
The first `alias` is meant to be the "canonical" alias, which is used in the json config.

We want `log.X` to be "canonical" rather than the more verbose (but currently required in lodestar) `logger.X`

Also removes `log.opPool`


mpetrunic (approved):
What happened to module name in logs?

wemeetagain:
> What happened to module name in logs?

Not sure, will investigate

PR #898, opened 2020-05-08
Forkchoice protoarray

### LINKED ISSUE #720: Switch to protoarray fork choice implementation
Low-ish priority right now, since what we have works.

Switching to protoarray fork choice may be slightly more efficient and it will be much easier to persist the fork choice state to disk if we decide to (eg as part of a shutdown process)

twoeths:
##  Goal
+ Resolves #720 
+ Need to implement restoring forkchoice in #911


packages/lodestar/src/chain/chain.ts
```diff
@@ -185,6 +186,30 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
+        slot: block.message.slot,
+        blockRootBuf: this.config.types.BeaconBlock.hashTreeRoot(block.message),
+        stateRootBuf: stateRoot,
+        parentRootBuf: block.message.parentRoot.valueOf() as Uint8Array,
+        justifiedCheckpoint: state.currentJustifiedCheckpoint,
+        finalizedCheckpoint: state.finalizedCheckpoint
+      });
+      await this.attestationProcessor.receiveBlock(block);
+    }
+    this.logger.info(`Restore forkchoice in ${Math.floor(Date.now() - start)/1000} seconds, numblock=${blocks.length}`);
```
twoeths: One possibility is to store the whole forkchoice to db on `stop()` and do simple check against latest db block to use it on `start()`. If the check returns false we will build forkchoice from scratch like this.

packages/lodestar/src/chain/forkChoice/statefulDag/lmdGhost.ts
```diff
@@ -11,7 +11,7 @@ import {computeSlotsSinceEpochStart,
   getCurrentSlot,
```
wemeetagain: If you want, you can move the protoarray implementation to a new directory. (Or change the name of the directory)
(This was originally for the "stateful DAG" implementation of the fork choice, now no longer relevant here.)

twoeths: do you want to keep both "stateful DAG" and "protoarray" or just a rename?

wemeetagain: pro of keeping: might be interesting to be able benchmark one against the other
con of keeping: maintenance of unused fork choice implementation might become a hassle in the future

Lets start with keeping the old implementation. If it turns out we don't need it at all, we can delete it later.

packages/lodestar/src/chain/chain.ts
twoeths: @wemeetagain I don't see Record/Map as basic ssz type so we can't store node list as ssz serialize(), can you confirm? The option for me is to use `JSON.stringnify()` to serialize node list and store to db (thanks to protoarray impl), when we start the chain we'll restore forkchoice and compare its head root against with `db.chain.getChainHeadRoot()`. If it's good we'll use it, otherwise we have to build forkchoice manually. Wdyt?

The other concern is how to restore AttestationAggregator and `bestJustifiedCheckpoint` inside the store. For this, I'd load all blocks and call `addBlock(), addAttestation()` manually to make sure it's good (as long as we don't load states, I think it's ok to load all blocks). Do you think it's necessary, is it heavy to load all blocks and process one by one?

twoeths: other option is probably to write `ForkChoiceNode` lodestar type with `childrenHexes: string[], childrenIndexes: number[]` to address the `children` Record/map inside the node and use our ssz serialize. Then we need to convert `ForkChoiceNode` and the `Node` type inside forkchoice store.

twoeths:
The protoarray forkchoice implementation works fine with the new sync. I can sync till slot 10000 in ~48min.

wemeetagain (dismissed):

wemeetagain (approved):

PR #900, opened 2020-05-08
Add schlesi preset config

### LINKED ISSUE #899: Schlesi config
Since Schlesi started already, I'd like to track this as another preset config: https://github.com/goerli/schlesi/blob/master/light/config.yaml

twoeths:
## Goal
+ resolves #899


wemeetagain:
I'm not sure we need a separate preset for schlesi, it appears to intentionally be using the mainnet preset.
https://github.com/goerli/schlesi#why-does-schlesi-use-the-mainnet-configuration

mpetrunic:
We could bake in genesis.ssz state and bootnodes so we can just add --network schesi and it would have all data already?

Or maybe on init it would download required files?

wemeetagain:
yeah, I think this is something we should provide thru `lodecli beacon init`. An additional flag which will download a beacon.config.json template, genesis.ssz, and anything else needed.

Not particularly worried about this right now, as our node is still not stable (so setting up a lodestar config will be the least of our users worries hahaha). Definitely will be useful in the future.

PR #901, opened 2020-05-08
Test block archive ordering

wemeetagain:
Test block archive key and value ordering.
It appears to be working as intended :)


mpetrunic (commented):
Try it with my e2e test though

wemeetagain:
@mpetrunic I updated the test here to include range filtering in different ways.
It looks like the problem on master is not having a `decodeKey` defined for the `BlockArchiveRepository`.

mpetrunic (approved):

PR #902, opened 2020-05-08
Optimize status req/resp

mpetrunic:
Since with mainnet config fetching and deserializing state is expensive and we were doing it 3 times for each status req/resp, this PR alleviate this and should make status a lot quicker.

Pls check this for sanity


packages/lodestar/src/chain/forkChoice/interface.ts
```diff
@@ -12,7 +12,7 @@ export interface ILMDGHOST {
   stop(): Promise<void>;
   addBlock(info: BlockChainInfo): void;
   addAttestation(blockRootBuf: Uint8Array, attester: ValidatorIndex, weight: Gwei): void;
-  head(): Uint8Array;
+  head(): {slot: Slot; root: Uint8Array};
```
wemeetagain: Can we break this out into `headBlockRoot` and `headSlot`?

When we convert to protoarray, we can take time to expose something with all head data like a `head(): BlockChainInfo`

packages/lodestar/src/chain/chain.ts
```diff
@@ -86,16 +86,15 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
   }
 
   public async getHeadState(): Promise<BeaconState|null> {
-    return this.db.state.get(this.forkChoice.headStateRoot());
+    return this.db.state.get(this.forkChoice.headStateRoot().valueOf() as Uint8Array);
```
wemeetagain: shouldn't need these `valueOf`s.

We only need `valueOf` to convert `TreeBacked<Vector<number>>` to `Uint8Array`

wemeetagain (changes requested):

packages/lodestar/src/chain/blocks/validate.ts
```diff
@@ -26,7 +26,7 @@ export function validateBlock(
           );
           continue;
         }
-        const currentSlot = (await db.block.get(forkChoice.head())).message.slot;
+        const currentSlot = (await db.block.get(forkChoice.head().root)).message.slot;
```
twoeths: shouldn't we just use `forkChoice.head().slot`?

packages/lodestar/src/chain/forkChoice/interface.ts
mpetrunic: What would we get other than huge bulk of changes and test fixes?
Feel like this is a good start to add more info when converting to protoarray. We already have those data in forkchoice but we aren't returning it

mpetrunic:
@wemeetagain @tuyennhv is forkChoice.getFinalized same as chain.headState().finalizedCheckpoint?

twoeths:
@mpetrunic  yeah should be the case.
+ `forkChoice.getFinalized` is the finalized checkpoint of chain
+ `chain.getHeadState()` calls `forkChoice.headStateRoot()` which returns `bestTarget` state root of justified checkpoint and a `bestTarget` should always have same finalized checkpoint like the store to be eligible.

packages/lodestar/src/chain/forkChoice/interface.ts
mpetrunic: or did you mean to rename slot, root into headBlockRoot and headBlockSlot?

wemeetagain: Sorry, I guess it doesn't have much to do with protoarray.

I was thinking we can/should return _all_ the data we have, the whole `BlockChainInfo` (or similar, with Uint8Array instead of hex string). All head-related methods then become "convenience" methods that call `head`.
```typescript
head(): BlockChainInfo {
  ...
  return this.justified.bestNode.target;
}
headStateRoot(): Uint8Array {
  return this.head().stateRoot;
}
headSlot(): Slot {
  return this.head().slot;
}
...
```

packages/lodestar/src/chain/forkChoice/interface.ts
```diff
@@ -21,11 +23,11 @@ export interface ILMDGHOST {
 /*
  * Info of Block and Chain for forkchoice
  */
-export interface BlockChainInfo {
+export interface BlockHeadInfo {
```
wemeetagain: Somewhat unrelated to the original PR, I'd like to expose the whole block-tree with these objects
eg: `getAtSlot(slot: Slot): BlockHeadInfo` (which would navigate from the head, backwards up the chain to return the info at the slot.)
In that case, perhaps we should have a slightly more neutral naming, eg: `ChainSummary` or `BlockSummary`, etc.
Also renaming `xxxRootBuf` to `xxxRoot` for more standard naming.

wemeetagain (approved):

PR #904, opened 2020-05-09
add new docker build file

### LINKED ISSUE #355: Build docker image in CI
We want our docker image built when the master branch gets updated.
Ideally it will not effect normal CI times.

mpetrunic:
After merge, I will add triggers on dockerhub, so our master is always present as image on DockerHub.

resolves #355


mpetrunic:
This is not really optimized image. I think we should have production image which will just install lodestar-cli package globally and be triggered on tag creation/npm publish. Currently we need whole monorepo inside docker because of symlinking

wemeetagain:
> ... I think we should have production image ...

Yeah, then we won't need the whole 4GB+ w spec tests, etc :joy:

wemeetagain (approved):

mpetrunic:
> > ... I think we should have production image ...
> 
> Yeah, then we won't need the whole 4GB+ w spec tests, etc :joy: 

we don't have that (ignore optional :) ) but it's still 700ish mb image because we have dev dependcies.
Production image will only do `yarn global add @chainsafe/lodestar-cli`

Issue #883, opened 2020-05-06
Optimize Eth1 Genesis


wemeetagain:
In our current implementation, creating a genesis state from eth1 gets slower and slower the more deposits are accumulated. This is because we naively recreate a new genesis deposit tree and state in every call to [`checkGenesis`](https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/chain/chain.ts#L218).

We ideally would maintain a genesis deposit tree and beacon state that we incrementally update with each new timestamp, eth1Data, and depositDatas.
This requires reworking the logic in [`initializeBeaconStateFromEth1`](https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/chain/genesis/genesis.ts#L32). We will still need `initializeBeaconStateFromEth1` to remain, to be used to test genesis in spec tests, but we won't be using the all-at-once `initializeBeaconStateFromEth1`, directly, in our beacon node.

PR #906, opened 2020-05-10
Process eth1 blocks in bulk

### LINKED ISSUE #889: Process eth1 deposit logs in bulk

We are currently fetching every eth1 block and deposit to check if it forms genesis.

Fetch deposit event logs in bulk (from deployedAt block till currentBlock-ETH1_FOLLOW_DISTANCE), while processing logs if we reach MIN_GENESIS_ACTIVE_VALIDATOR_COUNT deposits, start fetching eth1 blocks belonging to deposit and emit then to test if it forms valid genesis state.

We could also estimate which block number is most close to minimal eth1_timestamp to satisfy condition MIN_GENESIS_TIME condition and start processing eth1 blocks if there is enough deposits and block timestamp satisfies MIN_GENESIS_TIME

**Additional context**

twoeths:
## Goal
+ resolves #889 
+ eth1 pushes Eth1Data to chain only when needed so it mitigates the issue in #883 
## Approach
+ ethers is now stateless, no db getting involved there
+ Store lastProcessedEth1Block in db
+ New GenesisBuilder and WatchEth1ForProposingTask class to use ethers
+ ethers has 1 main interface `getDepositEventsFromBlock`, this is called by GenesisBuilder to build genesis state and by WatchEth1ForProposingTask for proposing data
+ At init sync time, only store deposit roots of events and last processed eth1 block because we don't need proposing data. After that time we store proposing data like Eth1Data, block number, timestamp
+ Genesis builder stores deposit roots and Eth1Data itself and save lastProcessedEth1Block so WatchEth1ForProposingTask just continue from there
+ Whenever there is a new finalized checkpoint, delete old deposit data and Eth1Data
## Test
+ I can detect schlesi genesis state in around 7s with 4 deposits, hope it does not break anything
image

## TODOs
+ [x] Rebase against master once it's ready, fix tests

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -9,7 +9,6 @@ import {
   BeaconState,
   Checkpoint,
   ENRForkID,
-  Eth1Data,
   ForkDigest,
   SignedBeaconBlock,
   Uint16,
@@ -25,7 +24,7 @@ import {EMPTY_SIGNATURE, GENESIS_SLOT} from "../constants";
 import {IBeaconDb} from "../db";
 import {IEth1Notifier} from "../eth1";
 import {IBeaconMetrics} from "../metrics";
-import {getEmptyBlock, initializeBeaconStateFromEth1, isValidGenesisState} from "./genesis/genesis";
+import {getEmptyBlock, GenesisBuilder} from "./genesis/genesis";
 import {ILMDGHOST, ArrayDagLMDGHOST} from "./forkChoice";
 
 import {ChainEventEmitter, IAttestationProcessor, IBeaconChain} from "./interface";
@@ -306,16 +305,10 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
     let state: TreeBacked<BeaconState> = await this.db.stateArchive.lastValue();
     if (!state) {
       this.logger.info("Chain not started, listening for genesis block");
-      state = await new Promise((resolve) => {
-        const genesisListener = async (timestamp: number, eth1Data: Eth1Data): Promise<void> => {
-          const state = await this.checkGenesis(timestamp, eth1Data);
-          if (state) {
-            this.eth1.removeListener("eth1Data", genesisListener);
-            resolve(state);
-          }
-        };
-        this.eth1.on("eth1Data", genesisListener);
-      });
+      const builder = new GenesisBuilder(this.config, {eth1: this.eth1, db: this.db, logger: this.logger});
+      state = await builder.genesis();
+      await this.initializeBeaconChain(state);
+      this.logger.info(`Genesis state is ready with ${state.validators.length} validators`);
     }
     // set metrics based on beacon state
     this.metrics.currentSlot.set(state.slot);
@@ -324,41 +317,4 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
     this.metrics.currentFinalizedEpoch.set(state.finalizedCheckpoint.epoch);
     return state;
   }
-
-  /**
-   * Create a candidate BeaconState from the deposits at a certain time and eth1 state
-   *
-   * Returns the BeaconState if it is valid else null
-   */
-  private checkGenesis = async (timestamp: number, eth1Data: Eth1Data): Promise<TreeBacked<BeaconState> | null> => {
-    const blockHashHex = toHexString(eth1Data.blockHash);
-    this.logger.info(`Checking if block ${blockHashHex} will form valid genesis state`);
-    const depositDatas = await this.db.depositData.values({lt: eth1Data.depositCount});
-    const depositDataRoots = await this.db.depositDataRoot.values({lt: eth1Data.depositCount});
-    this.logger.info(`Found ${depositDatas.length} deposits`);
-    const depositDataRootList = this.config.types.DepositDataRootList.tree.defaultValue();
-    const tree = depositDataRootList.tree();
-
-    const genesisState = initializeBeaconStateFromEth1(
-      this.config,
-      eth1Data.blockHash,
-      timestamp,
-      depositDatas.map((data, index) => {
-        depositDataRootList.push(depositDataRoots[index]);
-        return {
-          proof: tree.getSingleProof(depositDataRootList.gindexOfProperty(index)),
-          data,
-        };
-      })
-    );
-    if (!isValidGenesisState(this.config, genesisState)) {
-      this.logger.info(`Eth1 block ${blockHashHex} is NOT forming valid genesis state`);
-      return null;
-    }
-    this.logger.info(`Initializing beacon chain with eth1 block ${blockHashHex}`);
-    await this.initializeBeaconChain(genesisState as TreeBacked<BeaconState>);
-    this.logger.info(`Genesis state is ready with ${genesisState.validators.length} validators`);
-    return genesisState;
-  };
-
 }
--- a/packages/lodestar/src/chain/genesis/genesis.ts
+++ b/packages/lodestar/src/chain/genesis/genesis.ts
@@ -29,7 +29,7 @@ import {
   processDeposit,
 } from "@chainsafe/lodestar-beacon-state-transition";
 import {bigIntMin, ILogger} from "@chainsafe/lodestar-utils";
-import {TreeBacked, List, fromHexString} from "@chainsafe/ssz";
+import {TreeBacked, List, fromHexString, toHexString} from "@chainsafe/ssz";
 import {IBeaconDb} from "../../db";
 import {IEth1Notifier, IDepositEvent} from "../../eth1";
 import pipe from "it-pipe";
@@ -74,8 +74,8 @@ export class GenesisBuilder {
   public isFormingValidGenesisState = async (events: IDepositEvent[]): Promise<boolean> => {
     const block = await this.eth1.getBlock(events[0].blockNumber);
     applyTimestamp(this.config, this.state, block.timestamp);
-    applyEth1BlockHash(this.state, fromHexString(block.hash));
-    const isValid = isValidGenesisState(this.config, this.state);
+    applyEth1BlockHash(this.config, this.state, fromHexString(block.hash));
+    const isValid = isValidGenesisState(this.config, this.state, this.logger);
     this.logger.verbose(`genesis: Process block ${block.number}, valid genesis state=${isValid}`);
     if (isValid) {
       const eth1Data = {
@@ -208,8 +208,9 @@ export function initializeBeaconStateFromEth1(
   return state as TreeBacked<BeaconState>;
 }
 
-function applyEth1BlockHash(state: BeaconState, eth1BlockHash: Bytes32): void {
+function applyEth1BlockHash(config: IBeaconConfig, state: BeaconState, eth1BlockHash: Bytes32): void {
   state.eth1Data.blockHash = eth1BlockHash;
+  state.randaoMixes = Array<Bytes32>(config.params.EPOCHS_PER_HISTORICAL_VECTOR).fill(eth1BlockHash);
 }
 
 function applyTimestamp(config: IBeaconConfig, state: BeaconState, eth1Timestamp: number): void {
@@ -267,14 +268,23 @@ function applyDeposits(
 
 }
 
-export function isValidGenesisState(config: IBeaconConfig, state: BeaconState): boolean {
+export function isValidGenesisState(config: IBeaconConfig, state: BeaconState, logger: ILogger): boolean {
   if(state.genesisTime < config.params.MIN_GENESIS_TIME) {
+    logger.verbose(`genesis: not pass MIN_GENESIS_TIME of ${config.params.MIN_GENESIS_TIME}, ` +
+    `state.genesisTime=${state.genesisTime}`);
     return false;
   }
-  return getActiveValidatorIndices(state, computeEpochAtSlot(config, GENESIS_SLOT)).length
-      >=
-      config.params.MIN_GENESIS_ACTIVE_VALIDATOR_COUNT;
-
+  const numActiveValidator = getActiveValidatorIndices(state, computeEpochAtSlot(config, GENESIS_SLOT)).length;
+  const result = numActiveValidator >= config.params.MIN_GENESIS_ACTIVE_VALIDATOR_COUNT;
+  if (result) {
+    logger.info(`genesis: Found genesis state at eth1 block ${toHexString(state.eth1Data.blockHash)} ` +
+    `genesisTime=${state.genesisTime}`);
+  } else {
+    const numValidator = state.validators.length;
+    logger.verbose(`genesis: validator: ${numValidator}, active validator: ${numActiveValidator} , ` +
+    `require: ${config.params.MIN_GENESIS_ACTIVE_VALIDATOR_COUNT}`);
+  }
+  return result;
 }
 
 /**
--- a/packages/lodestar/src/chain/genesis/genesis.ts
+++ b/packages/lodestar/src/chain/genesis/genesis.ts
@@ -2,7 +2,6 @@
  * @module chain/genesis
  */
 
-import {TreeBacked} from "@chainsafe/ssz";
 import {
   BeaconBlock,
   BeaconBlockBody,
@@ -14,6 +13,7 @@ import {
   Bytes32,
   Fork,
   SignedBeaconBlock,
+  Root,
 } from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 
@@ -28,8 +28,160 @@ import {
   getTemporaryBlockHeader,
   processDeposit,
 } from "@chainsafe/lodestar-beacon-state-transition";
-import {bigIntMin} from "@chainsafe/lodestar-utils";
+import {bigIntMin, ILogger} from "@chainsafe/lodestar-utils";
+import {TreeBacked, List, fromHexString} from "@chainsafe/ssz";
+import {IBeaconDb} from "../../db";
+import {IEth1Notifier, IDepositEvent} from "../../eth1";
+import pipe from "it-pipe";
 
+export interface IGenesisBuilderModules {
+  db: IBeaconDb;
+  eth1: IEth1Notifier;
+  logger: ILogger;
+}
+
+export class GenesisBuilder {
+  private state: TreeBacked<BeaconState>;
+  private depositTree: TreeBacked<List<Root>>;
+  private readonly config: IBeaconConfig;
+  private readonly db: IBeaconDb;
+  private readonly eth1: IEth1Notifier;
+  private readonly logger: ILogger;
+
+  private deposits: Deposit[];
+
+  constructor(config: IBeaconConfig, {eth1, db, logger}: IGenesisBuilderModules) {
+    this.state = getGenesisBeaconState(
+      config,
+      config.types.Eth1Data.defaultValue(),
+      getTemporaryBlockHeader(
+        config,
+        getEmptyBlock()
+      )
+    );
+    this.depositTree = config.types.DepositDataRootList.tree.defaultValue();
+    this.config = config;
+    this.eth1 = eth1;
+    this.db = db;
+    this.logger = logger;
+    this.deposits = [];
+  }
+
+
+  /**
+   * This processes an eth1 block and check if it can form genesis state of eth1.
+   */
+  public isFormingValidGenesisState = async (events: IDepositEvent[]): Promise<boolean> => {
+    const block = await this.eth1.getBlock(events[0].blockNumber);
+    applyTimestamp(this.config, this.state, block.timestamp);
+    applyEth1BlockHash(this.state, fromHexString(block.hash));
+    const isValid = isValidGenesisState(this.config, this.state);
+    this.logger.verbose(`genesis: Process block ${block.number}, valid genesis state=${isValid}`);
+    if (isValid) {
+      const eth1Data = {
+        blockHash: fromHexString(block.hash),
+        depositRoot: this.depositTree.tree().root,
+        depositCount: events[events.length - 1].index + 1,
+      };
+      await this.db.eth1Data.put(block.timestamp, eth1Data);
+    }
+    return isValid;
+  };
+
+  public processDepositEvents(): ((source: AsyncIterable<IDepositEvent[]>) => AsyncIterable<IDepositEvent[]>) {
+    const {logger, doProcessDepositEvents, config} = this;
+    return ((source: AsyncIterable<IDepositEvent[]>): AsyncIterable<IDepositEvent[]> => {
+      return async function* () {
+        for await (const depositEvents of source) {
+          const lastBlockNumber = depositEvents[depositEvents.length - 1].blockNumber;
+          const firstBlockNumber = depositEvents[0].blockNumber;
+          for (let blockNumber = firstBlockNumber; blockNumber <= lastBlockNumber; blockNumber++) {
+            const blockDepositEvents = depositEvents.filter(event => event.blockNumber === blockNumber);
+            if (blockDepositEvents.length > 0) {
+              await doProcessDepositEvents(blockDepositEvents);
+              const depositCount = blockDepositEvents[blockDepositEvents.length - 1].index + 1;
+              logger.info(`genesis: Found ${blockDepositEvents.length} deposit events for block ${blockNumber}, ` +
+                `depositCount=${depositCount}`);
+              if (depositCount >= config.params.MIN_GENESIS_ACTIVE_VALIDATOR_COUNT) {
+                yield blockDepositEvents;
+              }
+            }
+          }
+        }
+      }();
+    });
+  }
+
+  public genesis = async (): Promise<TreeBacked<BeaconState>> => {
+    await this.initialize();
+    const eth1DataStream = await this.eth1.getDepositEventsByBlock(true, undefined);
+    const {isFormingValidGenesisState: processSingleBlock, state} = this;
+    const foundGenesis = this.eth1.foundGenesis.bind(this.eth1);
+    return await pipe(eth1DataStream,
+      this.processDepositEvents(),
+      async function(source: AsyncIterable<IDepositEvent[]>) {
+        for await (const depositEvents of source) {
+          const isValidGenesis = await processSingleBlock(depositEvents);
+          if (isValidGenesis) {
+            await foundGenesis();
+            return state;
+          }
+        }
+      });
+  };
+
+  /**
+   * This processes deposit events of a single eth1 block.
+   */
+  private doProcessDepositEvents = async (depositEvents: IDepositEvent[]): Promise<void> => {
+    const depositDataRoots = new Map<number, Root>();
+    depositEvents.forEach(
+      event => depositDataRoots.set(event.index, this.config.types.DepositData.hashTreeRoot(event)));
+    await Promise.all([
+      // op pool depositData
+      this.db.depositData.batchPut(depositEvents.map((depositEvent) => ({
+        key: depositEvent.index,
+        value: depositEvent,
+      }))),
+      // deposit data roots
+      this.db.depositDataRoot.batchPut(depositEvents.map((depositEvent) => ({
+        key: depositEvent.index,
+        value: depositDataRoots.get(depositEvent.index),
+      }))),
+      this.db.setLastProcessedEth1BlockNumber(depositEvents[0].blockNumber),
+    ]);
+    const deposits = depositEvents.map((depositEvent) => {
+      this.depositTree.push(depositDataRoots.get(depositEvent.index));
+      return {
+        proof: this.depositTree.tree().getSingleProof(this.depositTree.gindexOfProperty(depositEvent.index)),
+        data: depositEvent,
+      };
+    });
+    this.deposits.push(...deposits);
+    applyDeposits(this.config, this.state, this.deposits, this.depositTree);
+  };
+
+  private async initialize(): Promise<void> {
+    const depositDatas = await this.db.depositData.values();
+    const depositDataRoots = await this.db.depositDataRoot.values();
+    const deposits = depositDatas.map((event, index) => {
+      this.depositTree.push(depositDataRoots[index]);
+      return {
+        proof: this.depositTree.tree().getSingleProof(this.depositTree.gindexOfProperty(index)),
+        data: event,
+      };
+    });
+    this.deposits.push(...deposits);
+  }
+}
+
+/**
+ * Mainly used for spec test.
+ * @param config
+ * @param eth1BlockHash
+ * @param eth1Timestamp
+ * @param deposits
+ */
 export function initializeBeaconStateFromEth1(
   config: IBeaconConfig,
   eth1BlockHash: Bytes32,
@@ -37,7 +189,6 @@ export function initializeBeaconStateFromEth1(
   deposits: Deposit[]): TreeBacked<BeaconState> {
   const state = getGenesisBeaconState(
     config,
-    eth1Timestamp - eth1Timestamp % config.params.MIN_GENESIS_DELAY + 2 * config.params.MIN_GENESIS_DELAY,
     {
       depositCount: deposits.length,
       depositRoot: new Uint8Array(32),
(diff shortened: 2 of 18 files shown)


mpetrunic:
> Scan eth1block in bulk per n blocks starting from deployedAt. Check deposits per block, if it meets genesis state coondition then push eth1Data to chain to check

This is not scalable, it takes 1s to fetch and parse single block and logs and you have to do it for every block from deployedAt till genesis block, which is insane, it could take hours.

This is steps we should do:
- on start use filter [here](https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/eth1/impl/ethers.ts#L208) to fetch all deposit logs from deployedAt till latest (we can optimize later to fetch only deposits until MIN_GENESIS_TIME but it's fine for now) - this should fetch all deposits within minutes
- after we process deposit logs, start processing eth1 blocks from block that is closest to MIN_GENESIS_TIME and test if it forms valid genesis state (it can be firther optimized by taking genesis delay into account)

> Fetching deposits from deployedAt till currentBlock-ETH1_FOLLOW_DISTANCE is not an option for me because it's not scalable. For example next year if we launch a node for schlesi we'll fetch event logs of the whole year.

I don't think we need to worry about that now, it can be easily handler later on by bulk fetching deposits only up to MIN_GENESIS_TIME

twoeths:
> This is not scalable, it takes 1s to fetch and parse single block and logs and you have to do it for every block from deployedAt till genesis block, which is insane, it could take hours.

I don't parse every single block, only do it when it meets the genesis conditions. That's how I can detect genesis state in around 7s in my local environment. Whenever it meets genesis condition, we have to always scan every block from there. I think this is scalable.

> on start use filter here to fetch all deposit logs from deployedAt till latest (we can optimize later to fetch only deposits until MIN_GENESIS_TIME but it's fine for now) - this should fetch all deposits within minutes

MIN_GENESIS_TIME < deloyedAt for schlesi, that may happen for mainnet too. I also use filter, and I think fetching per n block until we find genesis state is better than fetching till latest as the latter way is not scalable, imagine `latest - deployedAt > 1 year, or 5 years...`

> I don't think we need to worry about that now, it can be easily handler later on by bulk fetching deposits only up to MIN_GENESIS_TIME

Again MIN_GENESIS_TIME < deployedAt timestamp for schlesi

mpetrunic:
> > This is not scalable, it takes 1s to fetch and parse single block and logs and you have to do it for every block from deployedAt till genesis block, which is insane, it could take hours.
> 
> I don't parse every single block, only do it when it meets the genesis conditions. That's how I can detect genesis state in around 7s in my local environment. Whenever it meets genesis condition, we have to always scan every block from there. I think this is scalable.

But then you have wrong depositRootList (invalid genesis), because you have to process all deposits until genesis state or you will create invalid genesis

twoeths:
I'm open to any approaches as long as it proves in e2e test and it's scalable. Let's see what @wemeetagain think about these, he may come up with his other solutions too :)

mpetrunic:
> I'm open to any approaches as long as it proves in e2e test and it's scalable. Let's see what @wemeetagain think about these, he may come up with his other solutions too :)

But we won't detect those in e2e we will detect those when we interop and fail because our state will be different. How do you get genesis validators if you don't process deposit logs from deployedAt till genesis?

twoeths:
> But then you have wrong depositRootList (invalid genesis), because you have to process all deposits until genesis state or you will create invalid genesis

I process per block, not per deposit, and I do it incrementally that's how I got valid genesis state in my test (see screenshot). I think this is basically just the same way to the current implementation of processing block by block so should be ok.

twoeths:
> But we won't detect those in e2e we will detect those when we interop and fail because our state will be different. How do you get genesis validators if you don't process deposit logs from deployedAt till genesis?

I think as long as we process block by block and we find a valid genesis state, it should be the correct one. And I do process logs from deployedAt till genesis. From what I see the genesis state of schlesi contains 4 deposits/validators, 2 from prysm and 2 from lighthouse. In my test I got 4 deposits/validators too.

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -91,14 +96,34 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
-      if(this.started) {
-        this.provider.on("block", this.onNewEth1Block.bind(this));
-      }
-    });
+    this.isScanEth1ForGenesis = true;
+    this.logger.info(
+      `Start scanning Eth1Data last block =${this.lastProcessedEth1BlockNumber} headBlockNumber=${headBlockNumber}`
+    );
+    while (this.lastProcessedEth1BlockNumber < headBlockNumber && this.isScanEth1ForGenesis) {
+      await this.onNewEth1Block(this.lastProcessedEth1BlockNumber + 100 + this.config.params.ETH1_FOLLOW_DISTANCE);
```
mpetrunic: what's 100 for?

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -91,14 +96,34 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
-    this.processBlocks(headBlockNumber).then(() => {
-      if(this.started) {
-        this.provider.on("block", this.onNewEth1Block.bind(this));
-      }
-    });
+    this.isScanEth1ForGenesis = true;
+    this.logger.info(
+      `Start scanning Eth1Data last block =${this.lastProcessedEth1BlockNumber} headBlockNumber=${headBlockNumber}`
+    );
+    while (this.lastProcessedEth1BlockNumber < headBlockNumber && this.isScanEth1ForGenesis) {
```
mpetrunic: This is only working because schlesi had minimal deposits within 4 mins after contract deployment so genesis is formed after couple of blocks 
Consider following scenario which will happen in mainnet and maybe multiclient testnet.
1. deposit contract is deployed at block n (I made that number up)
2. We will wait a month for minimal deposits and genesis time (31day*24hours * 4 blocks per hour ~= 3000 blocks)
3. We take usually around 1s to fetch and parse eth1 block (in which case this loop is gonna take 1 hour to execute this loop)

When mainnet launches it will probably be more than 1 month so we would take couple hours just to find genesis

twoeths: If you look into `processDepositEvents` you'll see I don't call `getBlock()` for all blocks. I only call it until we have enough `MIN_GENESIS_ACTIVE_VALIDATOR_COUNT` so I don't think we have issue with that scenario because we'll just keep filtering deposit events and parse them.

mpetrunic: Oh, you are going exactly what I'm saying:
onNewEth1Block ->processBlocksInBulk (deployedAt, eth1_head - follow distance)  -> getDepositEvents

processBlocksInBulk should be called processDepositEvents as it's not processing blocks just deposit events.

Also this needs a lot of renaming like onNewEth1Block is called to process bunch of blocks etc. Also, handle cases from https://github.com/ChainSafe/lodestar/issues/889#issuecomment-626302180

wemeetagain (commented):
So there are a lot of positives here:
- get deposits in batches, filter per block - This is definitely a good direction to go, as we can receive deposits for many (currently 100) blocks in the same time it takes to receive information about a single block.
- don't start fetching eth1 data until some threshold of deposits processed - This is also a big improvement. We know we won't need pre-genesis eth1 data, and processing `MIN_ACTIVE_VALIDATOR_COUNT`deposits before fetching eth1 data is an easy win.
- async iterables vs event emitter - much better to use an async iterable to push data to the chain module

I would love to see the above mechanisms driven by the chain (or a genesis builder chain submodule).
some pseudocode, ignoring some details:
```typescript
class GenesisBuilder {

  depositTree: TreeBacked<DepositDataRootList>;
  // (we likely don't need to add block
  depositStream: AsyncIterable<[number, DepositEvent[]]>;
  // may or may not need to track progress
  currentProcessedBlockNumber: number;

  constructor(eth1, db) {
    this.state = config.types.BeaconState.tree.defaultValue();
    this.depositTree = config.types.DepositDataRootList.tree.defaultValue();
    // This stream fetches blocks in batches of blocks, and transforms into block-by-block
    this.depositStream = eth1.getDepositEventsByBlock(deployedAt);
  }
  async genesis(): TreeBacked<BeaconState> {
    // process first MIN_ACTIVE_VALIDATOR_COUNT deposit events
    await this.processFirstNDepositEvents();
    // then we can process eth1 data
    await this.processSingleBlockEth1Data();
    // then we have to check block by block
    while(!this.isValidGenesis) {
      await this.processSingleBlockDepositEvents();
      await this.processSingleBlockEth1Data();
    }
    return this.state;
  }
  async processFirstNDepositEvents(): Promise<void> {
    for await (const [blockNumber, events] of this.depositStream) {
      // generate proofs, using this.depositTree
      const deposits = this.eventsToDeposits(events);
      applyDeposits(this.state, deposits);
      if (this.depositTree.length >= MIN_ACTIVE_VALIDATOR_COUNT) {
        // break out of this for await loop, when we have to start looking block by block
        // we can keep reusing this stream since it left off where we need
        break;
      }
    }
  }
}

// not actually used, but used in spec tests
// it uses the functions we use to incrementally update the state in GenesisBuilder
function initializeBeaconStateFromEth1(config, eth1BlockHash, eth1Timestamp, deposits: Deposit[]) {
  const state = config.types.BeaconState.tree.defaultValue();
  applyDeposits(state, deposits);
  applyEth1BlockHash(state, eth1BlockHash);
  applyEth1Timestamp(state, eth1Timestamp);
  return state;
}
```

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -27,16 +28,117 @@ import {
...
+    this.logger.verbose(`Process block ${block.number}, valid genesis state=${isValid}`);
+    return isValid;
+  };
+
+  public genesis = async (): Promise<BeaconState> => {
+    await this.initialize();
+    const eth1DataStream = await this.eth1.getDepositEventsByBlock();
+    const {depositTree, config, processEvents, processSingleBlock, state} = this;
+    const foundGenesis = this.eth1.foundGenesis.bind(this.eth1);
+    return new Promise(resolve => {
```
mpetrunic: You don't need this wrapper, pipe returns result of last function like this:
```typescript
return await pipe(
   eth1DataStream,
   async (source) => {
      for await {
        return state
     }
   }
)
```

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -27,16 +28,117 @@ import {
...
+      return {
+        proof: this.depositTree.tree().getSingleProof(this.depositTree.gindexOfProperty(event.index)),
+        data: event,
+      };
+    });
+    this.deposits.push(...deposits);
+    applyDeposits(this.config, this.state, this.deposits, this.depositTree);
+  };
+
+  public processSingleBlock = async (data: DepositEventsBlock): Promise<boolean> => {
```
mpetrunic: isFormingValidGenesisState is maybe better name

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -113,103 +139,79 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
+   * Process blocks in bulk.
+   * Must process blocks in order with no gaps.
+   * @param fromBlockNbr
+   * @param toBlockNbr
    */
-  public async processBlock(blockNumber: number): Promise<boolean> {
-    if (!this.started) {
+  public async getAndProcessDepositEvents(fromBlockNbr: number, toBlockNbr: number): Promise<void> {
+    if (!this.started && !this.isScanEth1ForGenesis) {
       this.logger.verbose("Eth1 notifier must be started to process a block");
```
mpetrunic: This should be error log

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -113,103 +139,79 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
     }
     this.processingBlock = true;
-    this.logger.verbose(`Processing eth1 block ${blockNumber}`);
-    if (blockNumber !== this.lastProcessedEth1BlockNumber + 1) {
-      this.logger.verbose(
-        `eth1 block out of order. expected: ${this.lastProcessedEth1BlockNumber + 1} actual: ${blockNumber}`
+    this.logger.verbose(`Processing eth1 block ${fromBlockNbr} to ${toBlockNbr}`);
+    try {
+      const depositEvents  = await this.getDepositEvents(fromBlockNbr, toBlockNbr);
+      this.logger.info(`Found ${depositEvents.length} deposit events`);
```
mpetrunic: maybe add block number in which deposits are found

mpetrunic (changes requested):
Close eth1Source in stop method.

Why do we have while loop with sleep in stop method of eth1?

packages/lodestar/src/eth1/impl/ethers.ts
twoeths: this is deposit events of a block range, logged the block range in above line. We can also log deposit events of a single block inside `processDepositEvents`

twoeths:
> we want to once stop() done we finish processing a block range(a block in previous version), for each block we have a couple of operations, we want it to be "all or nothing". We can't do `while` loop without `sleep()`, that makes the thread busy.
I can improve a bit to check for stop() method getting called per block instead of per block range

Why not use abortable pipe with event as source?
This way we can abort pipe and wait until it processed already fetched eth1 blocks.

twoeths:
> Why not use abortable pipe with event as source?
 This way we can abort pipe and wait until it processed already fetched eth1 blocks.

I thought about using pipe inside `processBlocksUntil`, it's more convenient to use normal while loop and break based on `started` variable

packages/lodestar/src/eth1/impl/ethers.ts
twoeths: on another look, this is not the error. It happens when we call stop(), I left it with verbose mode and change the message accordingly

packages/lodestar/src/eth1/impl/ethers.ts
twoeths: added that

wemeetagain:
Some thoughts on a tweak to this approach:
I think the eth1 notifier should be stateless, and it was probably a blunder to give it access to the database, it was just easier with the naive approach.
Instead, the eth1 notifier has mainly one interface `(blockNumber: number) => AsyncIterable<DepositEvent[]>`, which would return chunks of DepositEvent[] from `blockNumber` all the way to the follow distance, and keep pushing new deposit events at the follow distance, abortable, too. Ofc, for historical deposits, the underlying implementation can request deposits from ethers from a range of blocks, as is done in this PR.

Then we can use this interface to both build genesis and to collect deposit data roots, deposits, and eth1 data for proposing. We just need a few transforms. One transform would be `AsyncIterable<DepositEvent[]> => AsyncIterable<[blockNumber, DepositEvent[]>`. Another transform can take that further `AsyncInterable<[blockNumber, DepositEvent[]]> => AsyncIterable<[timestamp, Eth1Data, DepositEvent[]]>`.

For genesis, this PR is mostly there. If we don't have a genesis state, we get a stream starting at `depositContract.deployedAt`, We can process deposits until processing MIN_VALIDATORS deposits, then apply the above transforms to the stream and continue processing, just with additional processing for timestamp and Eth1Data. Also, we can modify the genesis builder to return the deposit roots, along with the genesis state, so we can store both in the db at the same time. This will help us not have to reprocess genesis deposits.

On the proposing side, sync can (eventually) start a stream that first just stores deposit roots, starting from the last stored deposit root, until we've finished initial syncing and have processed deposits to the deposit index in the state. Then we reusing the stream , add the transorms, and begins to store roots, deposits and eth1 data.

Perhaps in this case, the eth1 notifier doesn't need to be running all the time, and can just be instantiated in these two cases as needed
Also, if the eth1 notifier has the exact interface as above, we might need to store the latest processed block number, so we can efficiently restart from where we left off. Alternatively, we could start from a deposit index, and have the notifier determine how to serve that.

Thoughts?

twoeths:
> On the proposing side, sync can (eventually) start a stream that first just stores deposit roots, starting from the last stored deposit root, until we've finished initial syncing and have processed deposits to the deposit index in the state.

is eth1 be able to fetch deposit events by deposit index range? I only see we can do that by block number range, and we have to store all deposit roots so doing that after initial sync would be late. Correct me if I'm wrong but I'd store deposit roots right after chain starts, and only store eth1Data & deposit data after init sync is done for proposing. I'm in progress doing that.

> Also, if the eth1 notifier has the exact interface as above, we might need to store the latest processed block number

I prefer this way, doing that too

wemeetagain:
Yeah, seems easier to manage. In that case, the genesis builder can return the last processed block number along with the genesis state and deposit data roots, all three can be stored to db.

And the second transform could also yield the block number, eg `AsyncIterable<[block number, DepositEvent[]> => AsyncIterator<[timestamp, Eth1Data, block number, DepositEvent[]>`. Perhaps we don't even need that transform, just a public method `getEth1Data(blockNumber): Promise<Eth1Data>`

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -78,146 +76,95 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
-    );
-    const headBlockNumber = await this.provider.getBlockNumber();
-    // process historical unprocessed blocks up to curent head
-    // then start listening for incoming blocks
-    this.processBlocks(headBlockNumber).then(() => {
-      if(this.started) {
-        this.provider.on("block", this.onNewEth1Block.bind(this));
-      }
-    });
+    // if we try to form deposit, allow to scan deposit events even eth1 is disabled
```
twoeths: should be "if we try to form genesis"*

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -78,146 +76,95 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
+   * Should stop pushing deposit events to genesis builder.
+   */
+  public async foundGenesis(): Promise<void> {
+    this.eth1DataSource.end();
+    this.eth1DataSource = undefined;
+    this.targetBlockSource.end();
+    this.targetBlockSource = undefined;
+    // we'll start listening for new block after chain starts
+    this.provider.removeAllListeners("block");
+    while (this.processingBlock) {
```
twoeths, line 117: don't like this much but I have to because I want to make sure all the processing is done

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -226,26 +173,121 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
     }
     try {
       this.contract = new ethers.Contract(address, abi, this.provider);
     } catch (e) {
-      throw new Error("Eth1 deposit contract not found! Probably wrong eth1 rpc url");
+      throw new Error("eth1 deposit contract not found! Probably wrong eth1 rpc url");
     }
   }
 
+  private async doGetDepositEventsByBlock(): Promise<void> {
```
twoeths: doGetDepositEventsFromLastProcessedBlock

twoeths:
I can detect witty genesis state with this branch.
```
2020-05-26 15:14:47  []                 info: Initializing beacon chain with state root 0x773c694b47504d789dc768d2356f691866b47833d0d85e02511d7cd339925b17 and genesis block root 0x19aa2deaa02cac9774eb8948a8ead1ebe851ba9590878a10cd5767092e16ba12 
```

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -28,16 +28,167 @@ import {
...
+  }
+
+
+  /**
+   * This processes an eth1 block and check if it can form genesis state of eth1.
+   */
+  public isFormingValidGenesisState = async (events: IDepositEvent[]): Promise<boolean> => {
+    const block = await this.eth1.getBlock(events[0].blockNumber);
+    applyTimestamp(this.config, this.state, block.timestamp);
+    applyEth1BlockHash(this.config, this.state, fromHexString(block.hash));
```
mpetrunic, line 77: can you extract this to "updateGenesisState(state, eth1Block): state" method

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -28,16 +28,167 @@ import {
...
+    const isValid = isValidGenesisState(this.config, this.state);
+    this.logger.verbose(`genesis: Process block ${block.number}, valid genesis state=${isValid}`);
```
mpetrunic, line 79:
```suggestion
    this.logger.verbose(`genesis: Process eth1 block ${block.number}, valid genesis state=${isValid}`);
```

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -28,16 +28,167 @@ import {
...
+          key: depositEvent.index,
+          value: depositEvent,
+        }))),
+        this.db.eth1Data.put(block.timestamp, eth1Data)
+      ]);
+    }
+    return isValid;
+  };
+
+  public processDepositEvents(): ((source: AsyncIterable<IDepositEvent[]>) => AsyncIterable<IDepositEvent[]>) {
```
mpetrunic, line 99: I don't think this should be public method. Also can you extract it to separate file and write unit tests for it

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -28,16 +28,167 @@ import {
...
+import {IEth1Notifier, IDepositEvent} from "../../eth1";
+import pipe from "it-pipe";
 
+export interface IGenesisBuilderModules {
+  db: IBeaconDb;
+  eth1: IEth1Notifier;
+  logger: ILogger;
+}
+
+export class GenesisBuilder {
```
mpetrunic, line 43: Can you add interface for it so we can have proper stub types in tests

packages/lodestar/src/chain/genesis/genesis.ts
mpetrunic: return type is AsyncGenerator

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -28,16 +28,167 @@ import {
...
+        }))),
+        this.db.eth1Data.put(block.timestamp, eth1Data)
+      ]);
+    }
+    return isValid;
+  };
+
+  public processDepositEvents(): ((source: AsyncIterable<IDepositEvent[]>) => AsyncIterable<IDepositEvent[]>) {
+    const {logger, doProcessDepositEvents, config} = this;
+    return ((source: AsyncIterable<IDepositEvent[]>): AsyncIterable<IDepositEvent[]> => {
```
mpetrunic, line 101: types for this are inherited from function signature so they are not needed here

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -78,146 +76,95 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
-      }
-    });
+    // if we try to form genesis, allow to scan deposit events even eth1 is disabled
+    if (!this.opts.enabled) {
+      await this.initialize();
+    }
+    this.lastProcessedEth1BlockNumber = fromBlock || this.opts.depositContract.deployedAt - 1;
+    this.logger.info("eth1: Get deposit events by block from block " + this.lastProcessedEth1BlockNumber);
+    this.eth1DataSource = pushable<IDepositEvent[]>();
+    this.targetBlockSource = pushable<number>();
```
mpetrunic, line 100: Calling "getDepositEventsFromBlock" multiple times will override eth1DataSource and targetBlockSource.

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -28,16 +28,167 @@ import {
...
+                yield blockDepositEvents;
+              }
+            }
+          }
+        }
+      }();
+    });
+  }
+
+  public genesis = async (): Promise<TreeBacked<BeaconState>> => {
```
mpetrunic, line 123: probably better to rename "waitForGenesis"

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -78,146 +76,95 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
-      await this.initContract();
+    await this.initialize();
+  }
+
+  /**
+   * Main methods to be called either to build genesis state or to save data to propose blocks later.
+   * @param isScanEth1ForGenesis
+   * @param fromBlock
+   */
+  public async getDepositEventsFromBlock(
```
mpetrunic, line 87: this name implicates it will fetch deposit event from single block. Better name could be "getDepositEventsSource(abortSignal, fromBlock?: number)". I don't think you need "isScanEth1ForGenesis", we should have separate method "getEth1BlocksSource"

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -78,146 +76,95 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
+    // should not wait
+    this.doGetDepositEventsFromLastProcessedBlock();
+    return this.eth1DataSource;
+  }
+
+  /**
+   * Called by genesis builder to inform that we found genesis state.
+   * Should stop pushing deposit events to genesis builder.
+   */
+  public async foundGenesis(): Promise<void> {
```
mpetrunic, line 110: Instead of this method, it's better to solve it using abort signal as param, which will trigger source closing

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -226,26 +173,122 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
+  }
+
+
+  /**
+   * Scan deposit events per every n (100 currently) blocks until inBlockNumber
+   * If undefined inBlockNumber then this is a reusable pipe, process whenever we push to targetBlockSource.
+   * If inBlockNumber exists, we close pipe after use.
+   * @param inBlockNumber
+   */
+  private async process(inBlockNumber?: number): Promise<void> {
```
mpetrunic, line 206: I feel like this and methods it's using are way too complicated and have way too much responsibilities.

packages/lodestar/src/tasks/tasks/watchEth1ForProposingTask.ts
```diff
@@ -0,0 +1,166 @@
...
+      this.db.depositData.deleteOld(depositCount)
+    ]);
+    this.logger.info(`watchEth1: Deleted old eth1Data and deposit up to eth1 block ${eth1BlockHash} ` +
+    `timestamp=${eth1Timestamp}`);
+  }
+
+  /**
+   * Store last processed eth1 block number and deposit data roots.
+   */
+  private storeDepositDataRoots(): ((source: AsyncIterable<IDepositEvent[]>) => AsyncIterable<IDepositEvent[]>) {
```
mpetrunic, line 81: extract those to separate files so you can test them properly

packages/lodestar/src/tasks/tasks/watchEth1ForProposingTask.ts
```diff
@@ -0,0 +1,166 @@
...
+import {IBeaconSync} from "../../sync";
+import {Eth1Data} from "@chainsafe/lodestar-types";
+
+export interface IWatchEth1ForProposingModules {
+  logger: ILogger;
+  db: IBeaconDb;
+  eth1: IEth1Notifier;
+  sync: IBeaconSync;
+}
+export class WatchEth1ForProposingTask {
```
mpetrunic, line 18: Don't really like this as a task. It should go either in eth1 module or chain or separate module. Not really sure

packages/lodestar/src/chain/chain.ts
```diff
@@ -306,16 +305,10 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
     let state: TreeBacked<BeaconState> = await this.db.stateArchive.lastValue();
```
mpetrunic, line 305: we should pribably rename this from "waitForState" to "waitForGenesis"

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -28,16 +28,167 @@ import {
...
+    await this.initialize();
+    const eth1DataStream = await this.eth1.getDepositEventsFromBlock(true, undefined);
+    const {isFormingValidGenesisState: processSingleBlock, state} = this;
+    const foundGenesis = this.eth1.foundGenesis.bind(this.eth1);
+    return await pipe(eth1DataStream,
+      this.processDepositEvents(),
+      async function(source: AsyncIterable<IDepositEvent[]>) {
+        for await (const depositEvents of source) {
+          const isValidGenesis = await processSingleBlock(depositEvents);
```
mpetrunic, line 132: this line makes no sense processSingleBlock that receives deposit events returns isValidGenesis

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -28,16 +28,167 @@ import {
...
+    });
+  }
+
+  public genesis = async (): Promise<TreeBacked<BeaconState>> => {
+    await this.initialize();
+    const eth1DataStream = await this.eth1.getDepositEventsFromBlock(true, undefined);
+    const {isFormingValidGenesisState: processSingleBlock, state} = this;
+    const foundGenesis = this.eth1.foundGenesis.bind(this.eth1);
+    return await pipe(eth1DataStream,
+      this.processDepositEvents(),
```
mpetrunic, line 129: It makes more sense for this to have eth1Blocks as source than deposit events

mpetrunic (commented):
I really don't wan't to block this PR as we need it but it has so much hard to read and complex code and functions with illogical naming and params that I don't really wan't to aprove as it will will bring us a lot of bugs in the future. I won't request changes so @wemeetagain can approve it if he feels like we can fix this in the future.

I would suggest opening new branch from master and adding things from here there but more cleanly with better structure. Probably write some pseudocode of logic like:
waitForGenesis-> getDepositEvents(max = min_validators) -> listen on eth1 blocks -> add deposit events -> isValidGenesis or something

packages/lodestar/src/chain/genesis/genesis.ts
twoeths: this is to process deposit events of a single block, and we accumulate deposits to the cached state and check if that state is genesis.
I guess the confusion comes from the name. How do you want to name it?

packages/lodestar/src/eth1/impl/ethers.ts
twoeths: yeah but we don't have a case like that

packages/lodestar/src/eth1/impl/ethers.ts
twoeths: yeah really, I thought you want to use `pipe` in one of your comments. I can come back to the old way
```
while (this.lastProcessedEth1BlockNumber < followUp && (this.isStarted || this.isScanForGenesis)) {
  awai this.getAndProcessDepositEvents(this.lastProcessedEth1BlockNumber + 1, blockNumber);
}
```

packages/lodestar/src/chain/genesis/genesis.ts
twoeths: we want to process deposit events here by applying deposits to the cached state. Once we have enough deposits, we'll group deposit events by block, get eth1block and do `processSingleBlock`

packages/lodestar/src/eth1/impl/ethers.ts
mpetrunic: It's not pipe that's problematic, problem is that this method serves to fetch deposit events from single block (which should be separate method "getBlockDepositEvents("block")) and also does batch processing which feeds events into another "source" which would make more sense to be getBatchEth1EventSource where pipe would return AsyncGenerator \<DepositEvent\> or something

packages/lodestar/src/chain/genesis/genesis.ts
mpetrunic: why not have one pipe where you process just some batched eth1 deposit events source and when you have enough, end pipe and start another where source are eth1 blocks which you then process?

packages/lodestar/src/chain/chain.ts
wemeetagain: It isn't necessarily going to fetch the genesis state, will fetch latest finalized state if its in the DB

packages/lodestar/src/chain/genesis/genesis.ts
wemeetagain: I think this is generally better seen as a function that mutates a state rather than a pure function. Ofc we can return the mutated state, too.

packages/lodestar/src/db/api/beacon/beacon.ts
```diff
@@ -79,4 +81,19 @@ export class BeaconDb extends DatabaseService implements IBeaconDb {
     await super.stop();
     this.stateCache.clear();
   }
+
+  public async getLastProcessedEth1BlockNumber(): Promise<number> {
+    const data = await this.db.get(
+      encodeKey(Bucket.chainInfo, "lastProcessedBlockNumber")
```
wemeetagain, line 87: Can you extract this string into a constant? and reference "eth1" in the string?

packages/lodestar/src/tasks/tasks/watchEth1ForProposingTask.ts
wemeetagain: We might consider bringing all the chain factory code together into a separate module, eg: `src/chainFactory`. I think this is related to that

packages/lodestar/src/eth1/impl/ethers.ts
twoeths: no, we always get deposit events in batch here. Even in the case of new block, we get deposit events from last processed block up to the follow up block to make sure we don't miss any events of blocks.

So the only business here is to simply get events up to a block and feed it into another source.

packages/lodestar/src/chain/genesis/genesis.ts
twoeths: we want both block data (mainly hash + timestamp) and deposit events in order to save proposing data (eth1Data + deposit data). I'm thinking about moving the persistence of proposing away from here to make it single responsibility.

wemeetagain:
This PR is becoming a bit of a monster, partly my fault by introducing the genesis builder idea :joy:  I propose we break this work up into a few PRs:
1. deposit processing by range

2. Incremental genesis state creation w the genesis builder

3. Abortable eth1 streams, stream transforms, "the right" db handling.

This PR has some of all 3. If we go one by one, this might give us time to work on a design doc that can explain in more detail/formalize the strategy that we want.

twoeths:
yeah let's close this PR and replace by some following smaller PRs

PR #907, opened 2020-05-10
Fix compilation errors

twoeths:
## Goal
+ We have some compilation errors in test but CI did not fail
## Question
+ If we bring back https://github.com/ChainSafe/lodestar/pull/664/files#diff-4ab3bf1feca40a36f6c4967139b0fe10R27 , we'll see tests' compilation errors during `build` process. Do we want to do that?

### DIFF
--- a/packages/lodestar/src/network/network.ts
+++ b/packages/lodestar/src/network/network.ts
@@ -59,7 +59,8 @@ export class Libp2pNetwork extends (EventEmitter as { new(): NetworkEventEmitter
         this.peerInfo = libp2p.peerInfo;
         this.libp2p = libp2p;
         this.reqResp = new ReqResp(opts, {config, libp2p, logger});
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/network/network.ts
```diff
@@ -59,7 +59,8 @@ export class Libp2pNetwork extends (EventEmitter as { new(): NetworkEventEmitter
         this.peerInfo = libp2p.peerInfo;
         this.libp2p = libp2p;
         this.reqResp = new ReqResp(opts, {config, libp2p, logger});
-        const enr = (this.libp2p._discovery.get("discv5") as Discv5Discovery)?.discv5?.enr || undefined;
```
twoeths, line 62: curious about this `?.` syntax, is it for newer node version?

mpetrunic: optional chaining, not supported by nodejs until october 2020

mpetrunic (approved):

PR #910, opened 2020-05-13
Add state cache

wemeetagain:
Generally, this PR allows much of the chain processing to happen in memory, rather than fetching/storing to disk. Fork choice, recent states, unfinalized block(summary) tree are all in memory, while unfinalized blocks remain stored to disk.
Ideally this addition will be followed up with a backup/restore procedure to store/restore the unfinalized block summary and recent state cache on teardown/bootup.

- Add `forkChoice.getBlockSummaryAtSlot` - this lets us get rid of the (incorrect) Bucket.blockSlotRef slot -> blockRoot mapping. Instead, `forkChoice.getBlockSummaryAtSlot` exposes the whole BlockSummary, indexed by slot, as seen by the fork choice head.
- Add `chain.getBlockAtSlot` - use `forkChoice.getBlockSummaryAtSlot` to fetch a block, indexed by slot, from the database, works for both unfinalized or finalized blocks. This lets us get rid of `db.block.getBySlot` which relied on Bucket.blockSlotRef
- Remove `db.chain` and related - no longer store head information in db, instead, rely on `chain.forkChoice`
- Add `db.stateCache` and `db.stateArchive`, remove `db.state` - Instead of storing all states to disk, indexed by root, store finalized states to disk in `db.stateArchive`, indexed by epoch, store all unfinalized states in `db.stateCache`, indexed by root.
Note: state cache has a similar API to `Repository`, but is synchronous
Note: state cache and state archive return and rely on `TreeBacked<BeaconState>`, not just `BeaconState`.
- Fetch latest finalized state from disk on startup, store new finalized state to disk on finalized checkpoint event, clear old finalized states from cache on finalized checkpoint event. This can be improved by pruning the stateArchive so we don't store every newly finalized state.


packages/lodestar/src/chain/blocks/post.ts
```diff
@@ -5,77 +5,52 @@ import {IBeaconDb} from "../../db/api";
...
+            "processedCheckpoint",
+            {epoch: currentEpoch, root: forkChoice.getBlockSummaryAtSlot(preSlot).blockRoot},
+          );
+          // newly justified epoch
+          if (preJustifiedEpoch < postState.currentJustifiedCheckpoint.epoch) {
+            logger.important(`Epoch ${postState.currentJustifiedCheckpoint.epoch} is justified!`);
+            metrics.previousJustifiedEpoch.set(preJustifiedEpoch);
+            metrics.currentJustifiedEpoch.set(postState.currentJustifiedCheckpoint.epoch);
+            eventBus.emit("justifiedCheckpoint", postState.currentJustifiedCheckpoint);
+          }
```
mpetrunic, line 40: I liked it more in separate method as it makes everything easier to cover with tests

packages/lodestar/src/chain/blocks/post.ts
```diff
@@ -5,77 +5,52 @@ import {IBeaconDb} from "../../db/api";
...
-        await attestationProcessor.receiveBlock(item.block);
-        metrics.currentSlot.set(item.block.message.slot);
-        eventBus.emit("processedBlock", item.block);
-        const preSlot = item.preState.slot;
-        const preFinalizedEpoch = item.preState.finalizedCheckpoint.epoch;
-        const preJustifiedEpoch = item.preState.currentJustifiedCheckpoint.epoch;
-        const currentEpoch = computeEpochAtSlot(config, item.postState.slot);
+      for await(const {block, preState, postState} of source) {
+        await db.processBlockOperations(block);
+        await attestationProcessor.receiveBlock(block);
```
mpetrunic: await Promise.all()

- just realized

packages/lodestar/src/db/api/beacon/stateCache.ts
```diff
@@ -0,0 +1,33 @@
+import {toHexString, TreeBacked, ByteVector} from "@chainsafe/ssz";
+import {BeaconState} from "@chainsafe/lodestar-types";
+
+/**
+ * Simple BeaconState in-memory cache
+ *
+ * Similar API to Repository, but synchronous
+ */
+export class StateCache {
```
mpetrunic, line 9: make it implement Repository it will be async and we could stub it out or provide altertive implementation easily

mpetrunic (commented):
I like this very much!

I would maybe leave it async to allow more flexibility in future with different cache systems where some might require async api (in memory redis or separate process which would allow us to circumvent memory limits of nodejs process).

mpetrunic (dismissed):

mpetrunic (dismissed):

twoeths (approved):

PR #914, opened 2020-05-14
Set peer preferred encoding in reputation store

### LINKED ISSUE #871: Sending requests with wrong encoding to peers
**Describe the bug**

If peers connects to us and sends status request with ssz encoding, we will send future requests with ssz_snappy encoding which peer doesn't support.

**Expected behavior**

We should either store peer prefered encoding in metadata and use it when sending requests or send requests with both encoding at the same time.

I would prefer first solution as it reduces noise and number of requests.

**Steps to Reproduce**

1. start one node
2. start second node and connect to first node with ssz encoding
3. disconnect

twoeths:
## Goal
+ resolves #871

### DIFF
--- a/packages/lodestar/src/network/reqResp.ts
+++ b/packages/lodestar/src/network/reqResp.ts
@@ -27,6 +27,7 @@ import {RpcError} from "./error";
 import {eth2RequestDecode, eth2RequestEncode} from "./encoders/request";
 import {eth2ResponseDecode, eth2ResponseEncode} from "./encoders/response";
 import {IResponseChunk} from "./encoders/interface";
+import {IReputationStore} from "../sync/IReputation";
 
 interface IReqEventEmitterClass {
   new(): ReqEventEmitter;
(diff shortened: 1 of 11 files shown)


packages/lodestar/src/network/reqResp.ts
```diff
@@ -80,7 +84,7 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
             pipe(
               stream.source,
               eth2RequestDecode(this.config, this.logger, method, encoding),
-              this.handleRpcRequest(peerId, method),
+              this.handleRpcRequest(peerId, method, encoding),
```
mpetrunic: Feels like it would be a lot cleaner if you add additional function before request handling that would store encoding instead of passing that troughout stack.
like:
```typescript
pipe(
stream.source,
eth2RequestDecode,
storePeerEncodingPreference,
this.handleRpcRequest
)
```
This pipe will respond with proper encoding anyways, issue is just when we are sending request.

wemeetagain:
Just a note, once we migrate to libp2p v0.28, we plan on removing reputation store in favor of the libp2p peer store.

mpetrunic (approved):

wemeetagain (approved):

wemeetagain:
e2e test failure

mpetrunic:
e2e tests failing

twoeths:
I can't reproduce the failed e2e in local env. I restarted the check in CI and the issue is gone too.

PR #916, opened 2020-05-15
fix block target import on large chunks of skip slots

mpetrunic:


packages/lodestar/src/sync/initial/fast/index.ts
```diff
@@ -99,18 +99,21 @@ export class FastSync
     }
   }
 
+  private updateBlockImportTarget = (target: Slot): void => {
+    console.log("updating block target", target);
```
wemeetagain: replace with call to logger

wemeetagain (approved):
lgtm

PR #923, opened 2020-05-17
Fix log levels and modules

mpetrunic:
Fixes:
- missing module names
- log level restrictions
- setting log level per module

Currently global log level is set using --logger.node.level which is a bit awkward.
Do you think we should change that to --logger.level?


packages/lodestar-cli/src/lodecli/cmds/beacon/cmds/run/options/logger.ts
```diff
@@ -1,98 +1,98 @@
...
-  default: LogLevels[LogLevel.info],
+  default: LogLevel.info,
   group: "log",
 };
 
 export const logNode: Options = {
   alias: [
-    "log.node",
-    "logger.node",
+    "log.node.level",
```
wemeetagain, line 42: You can prepend another alias "log.level"

wemeetagain (approved):

packages/lodestar-cli/src/lodecli/cmds/beacon/cmds/run/options/logger.ts
mpetrunic: Oh yeah, thats true :D

mpetrunic: I'll do it in separate PR

mpetrunic: Actually, it won't work, because of default values for each module. "log.level" should change default values for all modules

PR #926, opened 2020-05-18
Restore chain state upon restart

### LINKED ISSUE #911: Backup/restore memory data when shutdown/startup
This is a continuation of #910  and #898 
+ On shutdown, persist state cache and forkchoice
+ On startup, restore state cache and forkchoice

Need to also handle the case where node is not shutdown properly (by a crash for example): start from all unfinalzied blocks and process block by block to generate state cache and forkchoice again.

twoeths:
## Goal
+ Close #911 
+ I can start/stop schlesy sync process multiple times
+ When we restore the state, we only access in-memory data by adding `reprocess` flag to `IBlockProcessJob`

## TODOs:
+ [x] Test the sync schlesi then stop then start again
image


packages/lodestar/src/chain/chain.ts
```diff
@@ -193,6 +194,70 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
+      const lastBlock = sortedBlocks[sortedBlocks.length - 1];
+      this.on("processedBlock", (signedBlock) => {
+        if (signedBlock.message.slot === lastBlock.message.slot &&
+          this.config.types.SignedBeaconBlock.equals(signedBlock, lastBlock)) {
+          resolve();
+        }
+      });
+    });
+    const duration = Math.floor(Date.now() - start) / 1000;
+    this.logger.important(`Restored chain state for ${sortedBlocks.length} blocks in ${duration}s`);
```
mpetrunic: https://github.com/winstonjs/winston#profiling should be used for that

packages/lodestar/src/chain/chain.ts
```diff
@@ -193,6 +194,70 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
     };
   }
 
+  /**
+   * Restore state cache and forkchoice from last finalized state.
+   */
+  private async restoreFromFinalized(): Promise<void> {
```
mpetrunic: restoreHeadState might be better name

packages/lodestar/src/chain/chain.ts
```diff
@@ -193,6 +194,70 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
+    const start = Date.now();
+    const lastFinalizedState: TreeBacked<BeaconState> = await this.db.stateArchive.lastValue();
```
mpetrunic: lastKnownState?

packages/lodestar/src/chain/chain.ts
```diff
@@ -193,6 +194,70 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
+    this.db.stateCache.add(lastFinalizedState);
+    // the block respective to finalized epoch is still in block db
+    const allBlocks = await this.db.block.values();
```
mpetrunic, line 227: Aren't you suppose to search blocks from lastFinalizedState.slot?

packages/lodestar/src/chain/chain.ts
```diff
@@ -193,6 +194,70 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
+    if (!allBlocks || allBlocks.length === 0) {
+      return;
+    }
+    this.logger.info(`Found ${allBlocks.length} blocks in database`);
+    if (allBlocks.length === 1) {
+      // start from scratch
+      const block = allBlocks[0];
+      const blockHash = this.config.types.BeaconBlock.hashTreeRoot(block.message);
+      if (this.config.types.Root.equals(blockHash, this.forkChoice.head().blockRoot)) {
```
mpetrunic: there should be this.forkChoice.headBlockRoot() func

packages/lodestar/src/chain/chain.ts
```diff
@@ -193,6 +194,70 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
+        this.logger.info("Chain is up to date, no need to restore");
+        return;
+      }
+    }
+    const finalizedStateRoot = this.config.types.BeaconState.hashTreeRoot(lastFinalizedState);
+    const sortedBlocks = allBlocks.sort((a, b) => a.message.slot - b.message.slot);
```
mpetrunic: we already haave "sortBlocks" method in codebase

packages/lodestar/src/chain/chain.ts
```diff
@@ -193,6 +194,70 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
+    }
+    await new Promise((resolve) => {
+      const lastBlock = sortedBlocks[sortedBlocks.length - 1];
+      this.on("processedBlock", (signedBlock) => {
+        if (signedBlock.message.slot === lastBlock.message.slot &&
+          this.config.types.SignedBeaconBlock.equals(signedBlock, lastBlock)) {
+          resolve();
+        }
+      });
+    });
```
mpetrunic: This feels like it would be a cool method to expose from chain as `public waitForBlockProcessed`

packages/lodestar/src/chain/chain.ts
twoeths: I think in the end we'll load all blocks in order to build forkchoice & stateCache so I load all from the beginning, did not compare the performance against loading block one by one from db on demand

packages/lodestar/src/chain/chain.ts
twoeths: that's nice

packages/lodestar/src/chain/chain.ts
mpetrunic: Makes sense, non finalized blocks are in archive anyways. We can leave this for now, we should probably store even some non finalized state in case of long non finalization like currently on schlesi.

packages/lodestar/src/chain/chain.ts
```diff
@@ -193,6 +195,74 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
+      finalizedCheckpoint: (isCheckpointNotGenesis)? lastKnownState.finalizedCheckpoint : blockCheckpoint
+    });
+
+    for (const block of sortedBlocks) {
+      await this.receiveBlock(block, true);
+    }
+    const lastBlock = sortedBlocks[sortedBlocks.length - 1];
+    await this.waitForBlockProcessed(this.config.types.BeaconBlock.hashTreeRoot(lastBlock.message));
+    this.logger.important(`Finish restoring chain head from ${allBlocks.length} blocks`);
+    this.logger.profile("restoreHeadState");
```
twoeths, line 288: It only prints `restoreHeadState` without duration. Any ideas @mpetrunic ?

mpetrunic: My best guess is because of our custom log formatting. It ignores metadata.
You should log `info` object here https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar-utils/src/logger/winston.ts#L32

I really wish we fix that so whatever metadata we sent there is logged like "<propertyName1>=<value1>, <propertyName2>=<value2>" at the end of log message

wemeetagain: :100: :100: i've been frustrated w our logger setup. I'd love to see more data exposed via a context object. (or more generally thinking of our logs as json-serializable objects) IMO we should at minimum fix up winston, possibly revisit other alternatives like pino and https://github.com/pinojs/pino-pretty . (bonus: it plays nicely w/ fastify)

mpetrunic: Ill fix logger in next pr so we can log context passed along with message

packages/lodestar/src/tasks/tasks/archiveStates.ts
```diff
@@ -38,10 +38,10 @@ export class ArchiveStatesTask implements ITask {
     this.logger.info(
       `Started archiving states (finalized epoch #${this.finalizedCheckpoint.epoch})...`
     );
-    // store the last available state in the finalized epoch
+    // store the state of finalized checkpoint
     const finalizedState = (await this.db.stateCache.values())
       .filter((state) => computeEpochAtSlot(this.config, state.slot) === this.finalizedCheckpoint.epoch)
-      .sort((a, b) => b.slot - a.slot)[0];
+      .sort((a, b) => a.slot - b.slot)[0];
```
twoeths, line 44: I want to store state of finalized check point so when I restore I'm confident with the finalized block based on `block.message.stateRoot = hashTreeRoot(lastKnownState)`.

mpetrunic: Don't you actually need first state where computeEpochAtSlot(this.config, state.slot) === this.finalizedCheckpoint.epoch) + 1?

twoeths: I need the 1st state to be the state of the finalized check point so that I initialize the state cache with that. When we reprocess block, we still need `preState` to exist, and that's the finalized state. So it's like:
+ Store state of finalized checkpoint here
+ When chain is started, find finalized block based on `block.message.stateRoot = hashTreeRoot(lastKnownState)`
+ Init forkchoice, init state cache with finalized block, finalized state
+ No need to process the finalized block, start processing blocks following the finalized block (normally finalized block.slot + 1 till end)

packages/lodestar/src/chain/blocks/validate.ts
```diff
@@ -11,6 +11,10 @@ export function validateBlock(
   return (source) => {
     return (async function*() {
       for await(const job of source) {
+        if (job.reprocess) {
+          yield job;
+          continue;
+        }
```
mpetrunic: why?

packages/lodestar/src/chain/chain.ts
```diff
@@ -121,6 +123,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
     await this.blockProcessor.start();
     this._currentForkDigest =  computeForkDigest(this.config, state.fork.currentVersion, state.genesisValidatorsRoot);
     this.on("forkDigestChanged", this.handleForkDigestChanged);
+    await this.restoreHeadState(state);
```
mpetrunic, line 126: why not call this inside waitForState?

packages/lodestar/src/chain/chain.ts
```diff
@@ -193,6 +202,88 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
+        }
+      });
+    });
+  }
+
+  /**
+   * Restore state cache and forkchoice from last finalized state.
+   */
+  private async restoreHeadState(lastKnownState: TreeBacked<BeaconState>): Promise<void> {
+    this.logger.info(`Found last known/finalized state at epoch #${lastKnownState.finalizedCheckpoint.epoch}`);
```
mpetrunic: print state root also

packages/lodestar/src/chain/chain.ts
```diff
@@ -193,6 +202,88 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
+    const allBlocks = await this.db.block.values();
+    if (!allBlocks || allBlocks.length === 0) {
+      return;
+    }
+    const sortedBlocks = sortBlocks(allBlocks);
+    const firstBlock = allBlocks[0];
+    const lastBlock = allBlocks[allBlocks.length - 1];
+    let firstSLot = firstBlock.message.slot;
+    let lastSlot = lastBlock.message.slot;
+    this.logger.info(`Found ${allBlocks.length} blocks in database, from slot ${firstSLot} to ${lastSlot}`);
```
mpetrunic: found x **non finalized blocks** in database

packages/lodestar/src/chain/chain.ts
```diff
@@ -193,6 +202,88 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
+    if (allBlocks.length === 1) {
```
mpetrunic, line 238: === 0?

packages/lodestar/src/chain/chain.ts
```diff
@@ -193,6 +202,88 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
+      (isCheckpointNotGenesis? this.config.types.Root.equals(block.message.stateRoot, stateRoot) : true);
+    });
+    if (!finalizedBlock) {
+      throw new Error(`Cannot find block for finalized state at slot ${lastKnownState.slot}`);
+    } else {
+      this.logger.info(`Found finalized block at slot ${finalizedBlock.message.slot},
+        root=${toHexString(this.config.types.BeaconBlock.hashTreeRoot(finalizedBlock.message))}`);
+    }
+    // init forkchoice
+    const blockCheckpoint = {
```
mpetrunic: missing Checkpoint type

packages/lodestar/src/chain/blocks/validate.ts
twoeths: otherwise this condition will be failed: `computeEpochAtSlot(config, job.signedBlock.message.slot) <= finalizedCheckpoint.epoch`.
If we do want to reprocess I think we can ignore the checks.

packages/lodestar/src/chain/chain.ts
twoeths: initially we stored the genesis block to database already so this is to catch that case

packages/lodestar/src/chain/blocks/validate.ts
mpetrunic: but let's say our last finalized epoch is 4, all non finalized blocks must have epoch >= 5 and won't trigger this check?

packages/lodestar/src/chain/chain.ts
mpetrunic: can you add comment about that as it's hard to figure out

packages/lodestar/src/chain/chain.ts
twoeths: we have to do this after forkchoice.start() and probably blockProcessor.start()

packages/lodestar/src/chain/blocks/validate.ts
twoeths: you're right, this just happened during my development. I removed it in the latest version.

packages/lodestar/src/chain/chain.ts
```diff
@@ -193,6 +202,92 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
+      this.logger.info("No need to process blocks");
+      return;
+    }
+    firstSLot = processedBlocks[0].message.slot;
+    lastSlot = processedBlocks[processedBlocks.length - 1].message.slot;
+    this.logger.info(`Start processing from slot ${firstSLot} to ${lastSlot} to rebuild state cache and forkchoice`);
+    for (const block of processedBlocks) {
+      await this.receiveBlock(block, true, true);
+    }
+    await this.waitForBlockProcessed(this.config.types.BeaconBlock.hashTreeRoot(lastBlock.message));
```
wemeetagain: might be a race condition here?
where the block processor processes the lastBlock before the processedBlock event is subscribed to?

packages/lodestar/src/chain/chain.ts
```diff
@@ -193,6 +202,92 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
+    this.db.stateCache.add(lastKnownState);
+    // the block respective to finalized epoch is still in block db
+    const allBlocks = await this.db.block.values();
+    if (!allBlocks || allBlocks.length === 0) {
+      return;
+    }
+    const sortedBlocks = sortBlocks(allBlocks);
+    const firstBlock = allBlocks[0];
+    const lastBlock = allBlocks[allBlocks.length - 1];
+    let firstSLot = firstBlock.message.slot;
```
wemeetagain:
```suggestion
    let firstSlot = firstBlock.message.slot;
```

packages/lodestar/src/tasks/tasks/archiveStates.ts
wemeetagain: I guess the confusing thing is that only the first block in a finalized epoch is finalized. The rest are are not finalized even tho the epoch is finalized. This is bc  the epoch processing happens at the beginning of the epoch, rather than at the end. So any further blocks processed happen after the finalization has already occurred.
this lgtm tho

wemeetagain (changes requested):
lgtm, just minor comments

wemeetagain (approved):

PR #927, opened 2020-05-18
Use forkChoice instead of db.block

twoeths:
## Goal
+ Use forkChoice whenever possible to avoid i/o operations

### DIFF
--- a/packages/lodestar/src/chain/attestation.ts
+++ b/packages/lodestar/src/chain/attestation.ts
@@ -43,7 +43,7 @@ export class AttestationProcessor implements IAttestationProcessor {
     this.pendingAttestations = new Map<BlockRootHex, Map<AttestationRootHex, Attestation>>();
   }
 
-  public async receiveAttestation(attestation: Attestation): Promise<void> {
+  public async  receiveAttestation(attestation: Attestation): Promise<void> {
     const attestationHash = this.config.types.Attestation.hashTreeRoot(attestation);
     this.logger.info(`Received attestation ${toHexString(attestationHash)}`);
     try {
@@ -57,13 +57,13 @@ export class AttestationProcessor implements IAttestationProcessor {
       return;
     }
     const targetRoot = attestation.data.target.root;
-    if (!await this.db.block.has(targetRoot.valueOf() as Uint8Array)) {
+    if (!this.forkChoice.hasBlock(targetRoot.valueOf() as Uint8Array)) {
       this.chain.emit("unknownBlockRoot", targetRoot);
       this.addPendingAttestation(targetRoot, attestation, attestationHash);
       return;
     }
     const beaconBlockRoot = attestation.data.beaconBlockRoot;
-    if (!await this.db.block.has(beaconBlockRoot.valueOf() as Uint8Array)) {
+    if (!this.forkChoice.hasBlock(beaconBlockRoot.valueOf() as Uint8Array)) {
       this.chain.emit("unknownBlockRoot", beaconBlockRoot);
       this.addPendingAttestation(beaconBlockRoot, attestation, attestationHash);
       return;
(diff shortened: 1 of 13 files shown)


packages/lodestar/src/chain/attestation.ts
```diff
@@ -57,13 +57,13 @@ export class AttestationProcessor implements IAttestationProcessor {
       return;
     }
     const targetRoot = attestation.data.target.root;
-    if (!await this.db.block.has(targetRoot.valueOf() as Uint8Array)) {
+    if (!this.forkChoice.hasBlock(targetRoot.valueOf() as Uint8Array)) {
```
mpetrunic, line 60: how much blocks does forkchoice hold? all non finalized?

packages/lodestar/src/chain/attestation.ts
```diff
@@ -99,8 +99,9 @@ export class AttestationProcessor implements IAttestationProcessor {
     const currentEpoch = computeEpochAtSlot(this.config, currentSlot);
     let checkpointState: BeaconState;
     if(justifiedCheckpoint.epoch > GENESIS_EPOCH) {
-      const justifiedBlock = await this.db.block.get(justifiedCheckpoint.root.valueOf() as Uint8Array);
-      checkpointState = await this.db.stateCache.get(justifiedBlock.message.stateRoot.valueOf() as Uint8Array);
+      const justifiedBlock =
+        this.forkChoice.getBlockSummaryByBlockRoot(justifiedCheckpoint.root.valueOf() as Uint8Array);
+      checkpointState = await this.db.stateCache.get(justifiedBlock.stateRoot);
```
mpetrunic: null check for justifiedBlock?

packages/lodestar/src/chain/attestation.ts
```diff
@@ -118,8 +119,8 @@ export class AttestationProcessor implements IAttestationProcessor {
       currentSlot >= computeStartSlotAtEpoch(this.config, target.epoch)
       , "Current slot less than this target epoch's start slot"
     );
-    const block = await this.db.block.get(attestation.data.beaconBlockRoot.valueOf() as Uint8Array);
-    assert(block.message.slot <= attestation.data.slot, "Attestation is for past block");
+    const block = this.forkChoice.getBlockSummaryByBlockRoot(attestation.data.beaconBlockRoot.valueOf() as Uint8Array);
+    assert(block.slot <= attestation.data.slot, "Attestation is for past block");
```
mpetrunic, line 129: null check for block?

packages/lodestar/src/chain/attestation.ts
twoeths: yeah basically it stores everything in db.block with (n-1) non finalized and 1 finalized node/block

packages/lodestar/src/chain/attestation.ts
```diff
@@ -120,6 +122,7 @@ export class AttestationProcessor implements IAttestationProcessor {
       , "Current slot less than this target epoch's start slot"
     );
     const block = this.forkChoice.getBlockSummaryByBlockRoot(attestation.data.beaconBlockRoot.valueOf() as Uint8Array);
+    assert(!!block, "The block of attestation data does not exist");
```
mpetrunic: add blockhash in message so we can investigate stuff like that

packages/lodestar/src/chain/attestation.ts
```diff
@@ -101,7 +101,9 @@ export class AttestationProcessor implements IAttestationProcessor {
     if(justifiedCheckpoint.epoch > GENESIS_EPOCH) {
       const justifiedBlock =
         this.forkChoice.getBlockSummaryByBlockRoot(justifiedCheckpoint.root.valueOf() as Uint8Array);
-      checkpointState = await this.db.stateCache.get(justifiedBlock.stateRoot);
+      if (justifiedBlock) {
+        checkpointState = await this.db.stateCache.get(justifiedBlock.stateRoot);
```
mpetrunic, line 105: I think we should throw error here or state will be null later on

twoeths: actually I think this will never happen, we should always get a not-null justified checkpoint from forkchoice

wemeetagain (approved):
looks great

PR #931, opened 2020-05-19
Fast state transition

### LINKED ISSUE #630: Fast state transition
see:
https://github.com/protolambda/eth2-docs
https://github.com/ethereum/eth2.0-specs/pull/1552

wemeetagain:
- resolves #630 (at least partially, there are other optimizations to be had)
- implements optimizations described in https://github.com/protolambda/eth2fastspec
- state transition code lives in `lodestar-beacon-state-transition/src/fast`
  - follows a similar 
  - reuses many state transition utils in `src/util`
  - cache machinery in `src/fast/util`
- spec tests passing: `yarn test:spec-fast` in spec-test-runner

- current status:
  - syncing is significantly faster, at ~15 blocks/sec (no bls, including networking), ~25 blocks/sec (no bls, only including the state transition)
  - ~syncing schlesi breaks at slot 4868, not sure why yet~


packages/lodestar-beacon-state-transition/src/fast/block/processAttestation.ts
```diff
@@ -0,0 +1,79 @@
...
+  const slot = state.slot;
+  const data = attestation.data;
+  if (!(data.index < epochCtx.getCommitteeCountAtSlot(data.slot))) {
+    throw new Error();
+  }
+  if (!(
+    data.target.epoch === epochCtx.previousShuffling.epoch ||
+    data.target.epoch === epochCtx.currentShuffling.epoch
+  )) {
+    throw new Error();
```
mpetrunic: I would love some meaningful messages instead of generic errors :smile:

packages/lodestar-beacon-state-transition/src/fast/block/processDeposit.ts
```diff
@@ -0,0 +1,70 @@
...
+  epochCtx: EpochContext,
+  state: BeaconState,
+  deposit: Deposit,
+): void {
+  const config = epochCtx.config;
+  const {EFFECTIVE_BALANCE_INCREMENT, MAX_EFFECTIVE_BALANCE} = config.params;
+  // verify the merkle branch
+  if (!verifyMerkleBranch(
+    config.types.DepositData.hashTreeRoot(deposit.data),
+    Array.from(deposit.proof).map((p) => p.valueOf() as Uint8Array),
```
mpetrunic: This could be optimized by not iterating twice(Array.from, map) just to cast type.
First step could be using mapFn as second argument to Array.from so it removes one extra looping

packages/lodestar-beacon-state-transition/src/fast/block/processAttestation.ts
```diff
@@ -15,25 +15,42 @@ export function processAttestation(
   const {MIN_ATTESTATION_INCLUSION_DELAY, SLOTS_PER_EPOCH} = config.params;
   const slot = state.slot;
   const data = attestation.data;
-  if (!(data.index < epochCtx.getCommitteeCountAtSlot(data.slot))) {
-    throw new Error();
+  const committeeCount = epochCtx.getCommitteeCountAtSlot(data.slot);
+  if (!(data.index < committeeCount)) {
+    throw new Error(
+      "Attestation committee index not within current committee count: " +
+      `committeeIndex=${data.index} committeeCount=${committeeCount}`
```
mpetrunic, line 22: You could've use features of new logger :smile:

mpetrunic (approved):

PR #936, opened 2020-05-22
Investigate the network e2e suspend issue

twoeths:
## Issue
+ See the issue in my local environment, not in any previous CI. Probably because yarn.lock is not up to date after we upgrade discv5. Let's try in CI.


packages/lodestar/test/e2e/network/network.test.ts
```diff
@@ -244,7 +244,7 @@ describe("[network] network", function () {
     netA.gossip.unsubscribeFromAttestationSubnet(forkDigest, "0", callback);
     expect(netA.gossip.listenerCount(getAttestationSubnetEvent(0))).to.be.equal(0);
   });
-  it("should connect to new peer by subnet", async function() {
+  it.only("should connect to new peer by subnet", async function() {
```
twoeths: this test is related to discv5 and maybe it's suspended because of this

wemeetagain (approved):

PR #969, opened 2020-05-27
Process deposit event logs in bulk

### LINKED ISSUE #889: Process eth1 deposit logs in bulk

twoeths:
## Goal
+ resolves #889  with minimal change
+ instead of doing `processBlock(blockNumber)` we get deposit events of block range and do `processBlock(blockNumber, blockDepositEvents)` and only process blocks that have deposit events

### DIFF
--- a/packages/lodestar/src/db/api/beacon/repositories/depositDataRoot.ts
+++ b/packages/lodestar/src/db/api/beacon/repositories/depositDataRoot.ts
@@ -51,12 +51,13 @@ export class DepositDataRootRepository extends Repository<number, Root> {
       await this.initTree();
     }
     const tree = this.depositRootTree.clone();
-    const maxIndex = tree.length - 1;
+    let maxIndex = tree.length - 1;
     if (depositIndex > maxIndex) {
       throw new Error(`Cannot get tree for unseen deposits: requested ${depositIndex}, last seen ${maxIndex}`);
     }
     while (maxIndex > depositIndex) {
       tree.pop();
+      maxIndex = tree.length - 1;
     }
     return tree;
   }
--- a/packages/lodestar/src/eth1/impl/ethers.ts
+++ b/packages/lodestar/src/eth1/impl/ethers.ts
@@ -129,6 +129,10 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
     }
   }
 
+  /**
+   * Process blocks from lastProcessedEth1BlockNumber + 1 until toNumber.
+   * @param toNumber
+   */
   public async processBlocks(toNumber: number): Promise<void> {
     let rangeBlockNumber = this.lastProcessedEth1BlockNumber;
     while (rangeBlockNumber < toNumber && this.started) {
@@ -141,15 +145,17 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
           + ` to ${blockNumber}`);
         continue;
       }
-      // get deposit events successfully, it's safe to update rangeBlockNumber
-      rangeBlockNumber = blockNumber;
-      if (!rangeDepositEvents || rangeDepositEvents.length === 0) {
-        continue;
-      }
+      let success = true;
       for (const [blockNumber, blockDepositEvents] of getDepositEventsByBlock(rangeDepositEvents)) {
         if (!await this.processBlock(blockNumber, blockDepositEvents)) {
+          success = false;
           break;
         }
+      }
+      // no error, it's safe to update rangeBlockNumber
+      if (success) {
+        rangeBlockNumber = blockNumber;
+        this.lastProcessedEth1BlockNumber = blockNumber;
       }
     }
   }
--- a/packages/lodestar/src/eth1/impl/ethers.ts
+++ b/packages/lodestar/src/eth1/impl/ethers.ts
@@ -10,12 +10,11 @@ import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {ILogger} from  "@chainsafe/lodestar-utils/lib/logger";
 
 import {isValidAddress} from "../../util/address";
-import {sleep} from "../../util/sleep";
 import {IBeaconDb} from "../../db";
 import {RetryProvider} from "./retryProvider";
 import {IEth1Options} from "../options";
 import {Eth1EventEmitter, IEth1Notifier, IDepositEvent} from "../interface";
-import {getDepositEventsByBlock} from "./util";
+import {groupDepositEventsByBlock} from "./util";
 
 export interface IEthersEth1Options extends IEth1Options {
   contract?: Contract;
@@ -47,7 +46,6 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
   private logger: ILogger;
 
   private started: boolean;
-  private processingBlock: boolean;
   private lastProcessedEth1BlockNumber: number;
 
   public constructor(opts: IEthersEth1Options, {config, db, logger}: IEthersEth1Modules) {
@@ -106,10 +104,6 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
     }
     this.provider.removeAllListeners("block");
     this.started = false;
-    this.logger.verbose("Eth1 notifier is stopping");
-    while (this.processingBlock) {
-      await sleep(5);
-    }
     this.logger.verbose("Eth1 notifier stopped");
   }
 
@@ -146,8 +140,8 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
         continue;
       }
       let success = true;
-      for (const [blockNumber, blockDepositEvents] of getDepositEventsByBlock(rangeDepositEvents)) {
-        if (!await this.processBlock(blockNumber, blockDepositEvents)) {
+      for (const [blockNumber, blockDepositEvents] of groupDepositEventsByBlock(rangeDepositEvents)) {
+        if (!await this.processDepositEvents(blockNumber, blockDepositEvents)) {
           success = false;
           break;
         }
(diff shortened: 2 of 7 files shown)


packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -129,10 +129,34 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
+      for (const [blockNumber, blockDepositEvents] of getDepositEventsByBlock(rangeDepositEvents)) {
+        if (!await this.processBlock(blockNumber, blockDepositEvents)) {
+          success = false;
+          break;
+        }
+      }
+      // no error, it's safe to update rangeBlockNumber
+      if (success) {
+        rangeBlockNumber = blockNumber;
+        this.lastProcessedEth1BlockNumber = blockNumber;
```
mpetrunic, line 152: Do we really need this? Can't we just fetch latest eth1Data in db and get block by hash?

I get it's easier to use this but we have to persist it anyway

packages/lodestar/src/eth1/impl/util.ts
```diff
@@ -0,0 +1,21 @@
+import {IDepositEvent} from "../interface";
+
+/**
+ * Return deposit events of blocks.
+ * @param depositEvents range deposit events
+ */
+export function getDepositEventsByBlock(rangeDepositEvents: IDepositEvent[]): [number, IDepositEvent[]][] {
```
mpetrunic: rename to "groupDepositEventsByBlock" so it's clear that it's not fetching block events

packages/lodestar/src/eth1/impl/util.ts
```diff
@@ -0,0 +1,21 @@
...
+  const firstBlockNumber = rangeDepositEvents[0].blockNumber;
+  const lastBlockNumber = rangeDepositEvents[rangeDepositEvents.length - 1].blockNumber;
+  const result: [number, IDepositEvent[]][] = [];
+  for (let blockNumber = firstBlockNumber; blockNumber <= lastBlockNumber; blockNumber ++) {
+    const blockDepositEvents = rangeDepositEvents.filter(event => event.blockNumber === blockNumber);
+    if (blockDepositEvents.length > 0) {
+      result.push([blockNumber, blockDepositEvents]);
+    }
+  }
+  return result;
```
mpetrunic: it would be nicer with reduce function but I'm fine with.
Like:
```
return depositEvents.reduce(<push depost event in map under block number key>, new Map()).entries()
```

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -143,79 +167,71 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
    *
    * Returns true if processing was successful
    */
-  public async processBlock(blockNumber: number): Promise<boolean> {
+  public async processBlock(blockNumber: number, blockDepositEvents: IDepositEvent[]): Promise<boolean> {
```
mpetrunic: Even though we fetched deposit events by range we are still fetching block by block even though we might not need it.

I would propose renaming this method into "processDepositEvents" and break eth1 block voting stuff into "processEth1Votes" (fetching block creating Eth1Data and storing into db). "processEth1Votes" would be only called if there are enough deposits (more than min validators) and spare us from useless block fetching.

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -108,9 +106,11 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
     }
     this.provider.removeAllListeners("block");
     this.started = false;
+    this.logger.verbose("Eth1 notifier is stopping");
     while (this.processingBlock) {
       await sleep(5);
     }
```
mpetrunic: I don't think we need this. Even if we stop eth1, processingBlock will still finish eventually because node won't kill process until something is still executing.

mpetrunic (changes requested):
That's much better now, I have few suggestions to make it more optimized and readable

packages/lodestar/src/eth1/impl/ethers.ts
twoeths: we need it because the latest eth1Data block hash may not be `blockNumber`, `blockNumber` may not contain any deposit events. We update `lastProcessedEth1BlockNumber` to `blockNumber` in order not to process old blocks again the next round.

packages/lodestar/src/eth1/impl/util.ts
twoeths: we need to sort by block number and deposit index (missing it now) so doing this is more convenient

mpetrunic: I think they are sorted when returned from rpc (not 100% on that though) but you can sort it before reduce

twoeths: done

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -143,79 +161,81 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
       }))),
       // deposit data roots
-      this.db.depositDataRoot.batchPut(depositEvents.map((depositEvent) => ({
+      this.db.depositDataRoot.batchPut(blockDepositEvents.map((depositEvent) => ({
         key: depositEvent.index,
         value: this.config.types.DepositData.hashTreeRoot(depositEvent),
       }))),
     ]);
-    const depositTree = await this.db.depositDataRoot.getTreeBacked(this.lastProcessedDepositIndex);
+    const depositCount = blockDepositEvents[blockDepositEvents.length - 1].index + 1;
```
mpetrunic, line 183: We should probably put notice that this is not 100% correct logic since deposit events can be small chunks or have invalid signatures (so they won't produce validator) but it won't matter much if we start processing blocks earlier

mpetrunic (dismissed):

packages/lodestar/src/db/api/beacon/repositories/depositDataRoot.ts
```diff
@@ -51,12 +51,13 @@ export class DepositDataRootRepository extends Repository<number, Root> {
...
     }
     const tree = this.depositRootTree.clone();
-    const maxIndex = tree.length - 1;
+    let maxIndex = tree.length - 1;
     if (depositIndex > maxIndex) {
       throw new Error(`Cannot get tree for unseen deposits: requested ${depositIndex}, last seen ${maxIndex}`);
     }
     while (maxIndex > depositIndex) {
       tree.pop();
+      maxIndex = tree.length - 1;
```
wemeetagain, line 60: XD :pray:

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -143,79 +161,81 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
+    return true;
+  }
+
+  /**
+   * Process proposing data of eth1 block
+   * @param blockNumber
+   * @param blockDepositEvents
+   * @returns true if success
+   */
+  public async processEth1Votes(blockNumber: number, blockDepositEvents: IDepositEvent[]): Promise<boolean> {
```
wemeetagain: recommend calling this `processEth1Data` since we aren't actually processing any votes here

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -143,79 +161,81 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
+      const blockNumber = log.blockNumber;
+      return this.parseDepositEvent(this.contract.interface.parseLog(log).values, blockNumber);
     });
   }
 
   public async getBlock(blockTag: string | number): Promise<Block> {
     try {
-      return this.provider.getBlock(blockTag, false);
+      // without await we can't catch error
+      const block = await this.provider.getBlock(blockTag, false);
```
wemeetagain: return await

packages/lodestar/src/eth1/impl/util.ts
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Return deposit events of blocks grouped/sorted by block number and deposit index
+ * @param depositEvents range deposit events
+ */
+export function groupDepositEventsByBlock(rangeDepositEvents: IDepositEvent[]): Map<number, IDepositEvent[]> {
+  if (!rangeDepositEvents || rangeDepositEvents.length === 0) {
+    return new Map();
+  }
+  rangeDepositEvents.sort((event1, event2) => event1.index - event2.index);
+  return rangeDepositEvents.reduce<Map<number, IDepositEvent[]>>((previousValue, currentValue) => {
```
wemeetagain: Might be nice to call the variables `groupedEvents` and `event`, or something like that.

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -143,79 +161,81 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
-      fromBlock: blockTag,
-      toBlock: blockTag,
+      fromBlock: fromBlockTag,
+      toBlock: toBLockTag || fromBlockTag,
       address: this.contract.address,
       topics: [this.contract.interface.events.DepositEvent.topic],
     })).map((log) => {
-      return this.parseDepositEvent(this.contract.interface.parseLog(log).values);
+      const blockNumber = log.blockNumber;
+      return this.parseDepositEvent(this.contract.interface.parseLog(log).values, blockNumber);
```
wemeetagain: You might consider putting the `this.contract.interface.parseLog` call inside `parseDepositEvent`, and only passing in the single arg `log`

wemeetagain (changes requested):
LGTM, just minor comments

mpetrunic (approved):

wemeetagain (approved):

PR #990, opened 2020-06-03
Add FFG-LMD vote consistency checks

### LINKED ISSUE #953: Add FFG-LMD vote consistency checks
https://github.com/ethereum/eth2.0-specs/pull/1742

twoeths:
## Goal
+ resolves #953

### DIFF
--- a/packages/lodestar/src/chain/forkChoice/interface.ts
+++ b/packages/lodestar/src/chain/forkChoice/interface.ts
@@ -18,6 +18,7 @@ export interface ILMDGHOST {
   headStateRoot(): Uint8Array;
   getJustified(): Checkpoint;
   getFinalized(): Checkpoint;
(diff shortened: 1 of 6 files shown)


packages/lodestar/src/chain/forkChoice/interface.ts
```diff
@@ -18,6 +18,7 @@ export interface ILMDGHOST {
   headStateRoot(): Uint8Array;
   getJustified(): Checkpoint;
   getFinalized(): Checkpoint;
+  getAncestor(root: RootHex, slot: Slot): RootHex | null;
```
wemeetagain: Can the input/output be Uint8Array instead of hex string? Perhaps an outer function, public, with Uint8Array input/output, and an inner function, private, as before, with hex string input/output

wemeetagain (changes requested):

wemeetagain (approved):

PR #991, opened 2020-06-04
Loosen restrictions for aggregate propagation to not require aggregator in bitfield

### LINKED ISSUE #958: Require "seen" aggregates to be valid to bound cache
https://github.com/ethereum/eth2.0-specs/pull/1749

### LINKED ISSUE #964: Loosen restrictions for aggregate propagation to not require aggregator in bitfield
https://github.com/ethereum/eth2.0-specs/pull/1801

twoeths:
## Goals
+ resolves #964 
+ resolves #958

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -59,7 +59,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
   public chainId: Uint16;
   public networkId: Uint64;
   public clock: IBeaconClock;
-  private epochCtx: EpochContext;
+  public epochCtx: EpochContext;
   private readonly config: IBeaconConfig;
   private readonly db: IBeaconDb;
   private readonly eth1: IEth1Notifier;
--- a/packages/lodestar/src/network/gossip/validator.ts
+++ b/packages/lodestar/src/network/gossip/validator.ts
@@ -24,10 +24,8 @@ import {
   getDomain,
   computeEpochAtSlot,
   computeSigningRoot,
-  getBeaconProposerIndex,
   isUnaggregatedAttestation,
-  computeSubnetForAttestation,
-  getBeaconCommittee,
+  computeSubnetForAttestation
 } from "@chainsafe/lodestar-beacon-state-transition";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {ATTESTATION_PROPAGATION_SLOT_RANGE, DomainType, MAXIMUM_GOSSIP_CLOCK_DISPARITY} from "../../constants";
@@ -92,7 +90,7 @@ export class GossipMessageValidator implements IGossipMessageValidator {
       return false;
     }
 
-    const supposedProposerIndex = getBeaconProposerIndex(this.config, state);
+    const supposedProposerIndex = this.chain.epochCtx.getBeaconProposer(signedBlock.message.slot);
     if (supposedProposerIndex !== signedBlock.message.proposerIndex) {
       return false;
     }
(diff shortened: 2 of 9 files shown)


twoeths:
`The valid aggregate attestation defined by `hash_tree_root(aggregate)` has _not_ already been seen (via aggregate gossip, within a verified block, or through the creation of an equivalent aggregate locally).` => not sure how do we check if a aggregated attestation stays in any verified blocks (currently we only check against AggregateAndProofDb - opPool)

packages/lodestar/src/network/gossip/validator.ts
```diff
@@ -175,8 +180,8 @@ export class GossipMessageValidator implements IGossipMessageValidator {
       return false;
     }
 
-    const attestorIndices = getAttestingIndices(this.config, state, attestationData, aggregate.aggregationBits);
-    if (!attestorIndices.includes(aggregatorIndex)) {
+    const committee = getBeaconCommittee(this.config, state, attestationData.slot, attestationData.index);
```
wemeetagain: We should have access to the committee in our `chain.epochCtx` cache.

I bet that we have many places in the gossip validation logic that could be refactored to use chain.epochCtx, might be worth tackling as as separate PR/issue.

wemeetagain (approved):

packages/lodestar/src/network/gossip/validator.ts
twoeths: I only see `getBeaconProposer` and `getBeaconCommittee` in EpochContext that can be applied so I applied that to this PR

packages/lodestar/src/sync/gossip/handler.ts
```diff
@@ -74,7 +74,7 @@ export class BeaconGossipHandler implements IGossipHandler {
   };
 
   private onAggregatedAttestation = async (aggregate: SignedAggregateAndProof): Promise<void> => {
-    await this.chain.receiveAttestation(aggregate.message.aggregate);
+    await this.db.aggregateAndProof.add(aggregate.message);
```
twoeths: need confirmation

wemeetagain: I think this should still be passed to the chain?
We need to process incoming attestations for the fork choice

twoeths: I'm not sure, currently we process attestations from received block so I thought this is to put it to the OpPool only to produce block and once the block is published we'll finally process attestations from block.

twoeths: if we pass to the chain here we may process again when the block containing the same attestation is published, isn't it a duplicate?

wemeetagain: the fork choice should handle duplicates by only counting the latest one received per attester.

I believe this is where the LMD (latest message driven) aspect of the fork choice shines through. We just feed it all valid messages, and it handles new/duplicate/updated attestations.

twoeths: I see, thanks for the clarification

packages/lodestar/src/chain/chain.ts
```diff
@@ -59,7 +59,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
   public chainId: Uint16;
   public networkId: Uint64;
   public clock: IBeaconClock;
-  private epochCtx: EpochContext;
```
mpetrunic, line 62: I wonder if it would be better to expose getter to copy of EpochContext as this posses danger that some code might change EpochContext accidentally. 

We could expose getBeaconProposers and getBeaconCommittee from chain which we also need in api.

wemeetagain: :+1: definitely anything returned by the API, we can't rely on being unmodified

wemeetagain: I think we can do this as a separate PR tho? what do yall think?

mpetrunic: I hoped @tuyennhv would do it in separate PR since couple of PR-s depends on this

wemeetagain (approved):

packages/lodestar/src/chain/chain.ts
twoeths: sure

PR #992, opened 2020-06-05
lodecli dev command

### LINKED ISSUE #983: Lodecli dev command
Should provide short command with a lot of predefined values so we can jumpstart chain and validators.

mpetrunic:
Some of the changes:
- depreacte old lodestar cli's
- ability to start dev chain with given count of genesis validators
- start from given ssz file
- fixed issue where default multiaddrs cannotbe overriden
- enabled reusing of ApiOverInstance amongst validators
- add interop key generation to validator package

resolves #983


mpetrunic:
Had to disable some e2e tests as they cannot work with new cli. Not sure why they are in cli instead of in lodestar (why they are using cli command instead of BeaconNode class directly)

packages/lodestar-cli/src/lodecli/cmds/dev/options/index.ts
```diff
@@ -0,0 +1,29 @@
...
+import * as logger from "./logger";
+import * as metrics from "./metrics";
+import * as network from "./network";
+import * as sync from "./sync";
+import {paramsOptions} from "./params";
+import {IGlobalArgs} from "../../../options";
+import {IBeaconNodeOptions} from "@chainsafe/lodestar/lib/node/options";
+import {IChainArgs} from "./chain";
+
+export const devRunOptions = canonicalOptions({
```
wemeetagain, line 16: It seems like the only options unique to the dev command are in `dev.ts`, the rest are copy/pasted from "beacon run" options.

I think, as a general rule, we'd want the dev options to be a superset of beacon run options.
Could you either make a note of that, eg: that the options are copy/pasted and/or re-factor the placement of the beacon run options to they can be reused here?

packages/lodestar-cli/src/lodecli/cmds/dev/run.ts
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * Run a beacon node
+ */
+export async function run(options: Arguments<IDevOptions>): Promise<void> {
+  await initBLS();
+
+  options = deepmerge(defaultOptions, options) as Arguments<IDevOptions>;
+  const peerId = await createPeerId();
+  options.network.discv5.enr = await createEnr(peerId);
```
wemeetagain, line 30: Can we save these to disk, something like the beacon init command.
EDIT: I see we delete everything on cleanup, ignore this comment

Side note:
In a future PR, perhaps we can roll the beacon init command into the beacon run command, like here.
So if the user wants to explicitly initialize the chain before running, they can (with beacon init).
But if they want the defaults and to just run the chain, they can just beacon run, and an on-disk conguration will be built for them on first bootup.

packages/lodestar-cli/src/lodecli/cmds/dev/run.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  const config = createIBeaconConfig({
+    ...(options.chain.name === "mainnet" ? mainnetParams : minimalParams),
+  });
+  const libp2p = await createNodeJsLibp2p(peerId, options.network);
+  const logger = new WinstonLogger();
+
+  const chainDir = join(options.rootDir, "dev", "beacon");
```
wemeetagain, line 39: I was flip-flopping about whether it makes sense to segregate the dev on-disk config from the beacon / validator command's config. (eg: whether we should save to `$ROOT_DIR/dev/beacon` or `$ROOT_DIR/beacon`)
I think it comes down to the intention behind the dev command.

(1) If the dev command is mostly about being a convenience during development, with a focus on spinning up an ephemeral beacon node with interop-style validators, then segregating to a `dev/` directory is likely the best approach.
(2) If the dev command is mostly about being a way to run lodecli with a beacon chain and all configured validators, we likely shouldn't segregate the configs into a `dev/` directory

I think we're clearly going for 1 here, not 2, so the current strategy seems appropriate.
We might clarify the intention behind the dev command in the yargs command description or set description to false (make `dev` a hidden command)

wemeetagain (changes requested):
LGTM, just minor comments
+1 for ez ephemeral testnet

packages/lodestar-cli/src/lodecli/cmds/dev/options/index.ts
mpetrunic: Well it's kinda weird, since I had to change default values for some option in every file. Not sure what would be best way to handle those?

packages/lodestar-cli/src/lodecli/cmds/dev/run.ts
mpetrunic: Yeah, I think that would be useful.

packages/lodestar-cli/src/lodecli/cmds/dev/run.ts
mpetrunic: Well, I did it mostly to avoid overwriting witti chain/config by accident when doing some tests. Makes it easier to keep it clean also

packages/lodestar-cli/src/lodecli/cmds/dev/options/index.ts
wemeetagain: ah, gotcha, didn't realize the defaults were changed. I think its good as-is, maybe just a comment to mention that its mostly copied but with different defaults. Maybe we can think of a better way in the future it it gets difficult to manage.

wemeetagain (approved):
:tada:

PR #999, opened 2020-06-09
SSZ size bound validation

### LINKED ISSUE #963: Req/Resp sizes constrained by SSZ bounds
https://github.com/ethereum/eth2.0-specs/pull/1800

twoeths:
## Goals
+ resolves #963 
+ We need to send `invalid request` status code to client if request does not pass validation
```
From requests: send back an error message, response code InvalidRequest. The request itself is ignored.
```

### DIFF
--- a/packages/lodestar/src/network/encoders/request.ts
+++ b/packages/lodestar/src/network/encoders/request.ts
@@ -1,10 +1,10 @@
-import {Method, Methods, ReqRespEncoding} from "../../constants";
+import {Method, Methods, ReqRespEncoding, RpcResponseStatus, ValidatedRequestBody} from "../../constants";
 import {ILogger} from "@chainsafe/lodestar-utils/lib/logger";
 import {RequestBody} from "@chainsafe/lodestar-types";
 import {decode, encode} from "varint";
 import BufferList from "bl";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
-import {getCompressor, getDecompressor} from "./utils";
+import {getCompressor, getDecompressor, maxEncodedLen} from "./utils";
 
 export function eth2RequestEncode(
   config: IBeaconConfig, logger: ILogger, method: Method, encoding: ReqRespEncoding
@@ -34,7 +34,7 @@ export function eth2RequestEncode(
 
 export function eth2RequestDecode(
   config: IBeaconConfig, logger: ILogger, method: Method, encoding: ReqRespEncoding
-): (source: AsyncIterable<Buffer|BufferList>) => AsyncGenerator<RequestBody> {
+): (source: AsyncIterable<Buffer|BufferList>) => AsyncGenerator<ValidatedRequestBody> {
   return (source) => {
     return (async function*() {
       let sszDataLength: number = null;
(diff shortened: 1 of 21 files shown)


packages/lodestar/src/network/encoders/response.ts
```diff
@@ -62,15 +62,26 @@ export function eth2ResponseDecode(
...
         if(sszLength === null) {
           sszLength = decode(buffer.slice());
+          if (decode.bytes > 10) {
+            throw new Error(`eth2ResponseDecode: Invalid number of bytes for protobuf varint ${decode.bytes}` +
+            `, method ${method}`);
+          }
           buffer.consume(decode.bytes);
         }
+        if (sszLength < type.minSize() || sszLength > type.maxSize()) {
+          throw new Error(`eth2ResponseDecode: Invalid szzLength of ${sszLength} for method ${method}`);
```
twoeths, line 79: According to the spec `From responses: ignore the response, the response MUST be considered bad server behavior.` so we have to throw Error here instead of log and break

packages/lodestar/src/network/encoders/request.ts
```diff
@@ -47,16 +47,34 @@ export function eth2RequestDecode(
         }
         if(sszDataLength === null) {
           sszDataLength = decode(chunk.slice());
+          if (decode.bytes > 10) {
+            logger.error(`eth2RequestDecode: Invalid number of bytes for protobuf varint ${decode.bytes}` +
+            `, method ${method}`);
+            yield RpcResponseStatus.ERR_INVALID_REQ;
```
mpetrunic: I don't think it's wise to yield that, body can be just number so it's confusing. Just print warn message and break so request doesn't get propagated

packages/lodestar/src/network/reqResp.ts
```diff
@@ -180,12 +180,16 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
...
   private handleRpcRequest(
     peerId: PeerId, method: Method
-  ): ((source: AsyncIterable<RequestBody>) => AsyncGenerator<IResponseChunk>) {
+  ): ((source: AsyncIterable<ValidatedRequestBody>) => AsyncGenerator<IResponseChunk>) {
     const getResponse = this.getResponse;
     return (source) => {
       return (async function * () {
         for await (const request of source) {
-          yield* getResponse(peerId, method, request);
+          if (request === RpcResponseStatus.ERR_INVALID_REQ) {
```
mpetrunic: This could be equal even for valid requests

mpetrunic (changes requested):

packages/lodestar/src/network/encoders/request.ts
twoeths: I see your point, but we have to send `ERR_INVALID_REQ` status to the client in this case, we can't just log and break ... definitely have to come up with a different way

mpetrunic (approved):

packages/lodestar/src/network/encoders/utils.ts
```diff
@@ -46,3 +46,10 @@ export function getDecompressor(encoding: ReqRespEncoding): IDecompressor {
 export function encodeResponseStatus(status: number): Buffer {
   return Buffer.from([status]);
 }
+
+export function maxEncodedLen(sszLength: number, encoding: ReqRespEncoding): number {
+  if (encoding === ReqRespEncoding.SSZ) {
+    return sszLength;
+  }
+  return 32 + sszLength + sszLength / 6;
```
wemeetagain, line 55: Is this padding from snappy compression? would you mind adding a brief comment on what/why this is?

twoeths: yeah added that. In the spec of p2p we have the `maxEncodedLen()` function too so this is just a 1-1 mapping. This is the worst-case compression for ssz and ssz-snappy.

wemeetagain (approved):

PR #1001, opened 2020-06-10
Add e2e tests for sync and finalization

mpetrunic:
- fixed bug when we encounter consensus split or possible target race issues
- finalization e2e tests without cli dependencies
- sync e2e tests without cli dependencies


mpetrunic:
Seems like 1s slot is too much for ci as it works locally always and on ci it sometimes fails

mpetrunic:
@wemeetagain Seems like it's stable now, think you forgot to increase event timeouts

wemeetagain (approved):

PR #1005, opened 2020-06-12
Return blocks on same chain in BeaconBlocksByRange

### LINKED ISSUE #962: Clarify blocks in BlocksByRange must be from a single chain
https://github.com/ethereum/eth2.0-specs/pull/1795

### LINKED ISSUE #966: Restrict and clarify BlocksByRange request
https://github.com/ethereum/eth2.0-specs/pull/1835

twoeths:
## Goals
+ resolves #966 
+ resolves #962

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -114,16 +114,17 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
     return this.epochCtx.copy();
   }
 
-  public async getUnfinalizedBlocksAtSlots(slots: Slot[]): Promise<(SignedBeaconBlock|null)[]> {
+  public async getUnfinalizedBlocksAtSlots(slots: Slot[]): Promise<SignedBeaconBlock[]|null> {
     if (!slots) {
-      return [];
+      return null;
     }
     const blockRoots = slots.map((slot) => {
       const summary = this.forkChoice.getBlockSummaryAtSlot(slot);
       return summary? summary.blockRoot : null;
-    });
+    }).filter((blockRoot) => !!blockRoot);
+    // these blocks are on the same chain to head
     return await Promise.all(blockRoots.map(
-      (blockRoot) => blockRoot? this.db.block.get(blockRoot) : Promise.resolve(null)));
+      (blockRoot) => this.db.block.get(blockRoot)));
   }
 
   public async getFinalizedCheckpoint(): Promise<Checkpoint> {
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -114,6 +114,18 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
     return this.epochCtx.copy();
   }
 
+  public async getUnfinalizedBlocksAtSlots(slots: Slot[]): Promise<(SignedBeaconBlock|null)[]> {
+    if (!slots) {
+      return [];
+    }
+    const blockRoots = slots.map((slot) => {
+      const summary = this.forkChoice.getBlockSummaryAtSlot(slot);
+      return summary? summary.blockRoot : null;
+    });
+    return await Promise.all(blockRoots.map(
+      (blockRoot) => blockRoot? this.db.block.get(blockRoot) : Promise.resolve(null)));
+  }
+
   public async getFinalizedCheckpoint(): Promise<Checkpoint> {
     return this.forkChoice.getFinalized();
   }
--- a/packages/lodestar/src/chain/interface.ts
+++ b/packages/lodestar/src/chain/interface.ts
@@ -66,6 +66,8 @@ export interface IBeaconChain extends ChainEventEmitter {
 
   getBlockAtSlot(slot: Slot): Promise<SignedBeaconBlock|null>;
 
+  getUnfinalizedBlocksAtSlots(slots: Slot[]): Promise<(SignedBeaconBlock|null)[]>;
+
   /**
    * Add attestation to the fork-choice rule
    */
--- a/packages/lodestar/src/chain/interface.ts
+++ b/packages/lodestar/src/chain/interface.ts
@@ -66,7 +66,7 @@ export interface IBeaconChain extends ChainEventEmitter {
 
   getBlockAtSlot(slot: Slot): Promise<SignedBeaconBlock|null>;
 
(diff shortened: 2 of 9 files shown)


packages/lodestar/src/sync/reqResp/reqResp.ts
```diff
@@ -175,16 +176,30 @@ export class BeaconReqRespHandler implements IReqRespHandler {
...
+    if (request.step < 1 || request.startSlot < GENESIS_SLOT || request.count < 1) {
+      this.logger.error(`Invalid request id ${id} start: ${request.startSlot} step: ${request.step}` +
+      ` count: ${request.count}`);
+      this.network.reqResp.sendResponse(
+        id,
+        new RpcError(RpcResponseStatus.ERR_INVALID_REQ, "Invalid request"),
+        null);
+      return;
+    }
+    if (request.count > 1000) {
```
twoeths, line 189: The spec allows us to have some limitation, is `1000` suitable?

packages/lodestar/src/tasks/tasks/archiveBlocks.ts
```diff
@@ -32,21 +33,40 @@ export class ArchiveBlocksTask implements ITask {
...
   }
 
+  /**
+   * Only archive blocks on the same chain to the finalized checkpoint.
+   */
   public async run(): Promise<void> {
-    const blocks = (await this.db.block.values()).filter(
+    const allBlocks = await this.db.block.values();
+    const finalizedBlock = allBlocks.find(block => {
+      const blockRoot = this.config.types.BeaconBlock.hashTreeRoot(block.message);
```
tuyennhv, line 42: does this return the cached root or it has to calculate again?

mpetrunic (changes requested):
You should probably also change in this PR chain.getBlockAtSlot to return Array of blocks and propagate it into database. Currently if we get and process block thats on fork it will override last block linked to that slot.

packages/lodestar/src/chain/interface.ts
```diff
@@ -65,6 +65,8 @@ export interface IBeaconChain extends ChainEventEmitter {
 
   getBlockAtSlot(slot: Slot): Promise<SignedBeaconBlock|null>;
 
+  getUnfinalizedBlocksAtSlots(slots: Slot[]): Promise<(SignedBeaconBlock|null)[]>;
```
mpetrunic:
```suggestion
  getUnfinalizedBlocksAtSlots(slots: Slot[]): Promise<SignedBeaconBlock[]|null>;
```

mpetrunic (changes requested):
I think there was misuunderstanding.
We should change how we store slot -> root mapping. Currently is 1:1 and it should be 1:n aka 1 slot can have many block roots attached. That also means that `db.blocks.getBlockAtSlot` should be  `db.blocks.getBlocksAtSlot`. slot -> root mapping is 1:1 only in finalized blocks (block archive)

twoeths:
@mpetrunic  I see what you mean now. Can you have a look at https://github.com/ethereum/eth2.0-specs/pull/1818/files

My understanding was we have 1-1 mapping for slot-block after that PR.

mpetrunic:
> @mpetrunic I see what you mean now. Can you have a look at https://github.com/ethereum/eth2.0-specs/pull/1818/files
> 
> My understanding was we have 1-1 mapping for slot-block after that PR.

I think this just prevents multiple blocks per slot in single head chain. In unfinalized chain we can have parallel chains (forks) where different block proposers are proposing different blocks for same slot we need to store multiple blocks per slot until we finalize this (decide on one fork to be correct).

twoeths:
> I think this just prevents multiple blocks per slot in single head chain. In unfinalized chain we can have parallel chains (forks) where different block proposers are proposing different blocks for same slot we need to store multiple blocks per slot until we finalize this (decide on one fork to be correct).

Thanks, clear to me now. With the latest implementation, I don't touch `getBlockAtSlot` so I'm not implementing `getBlocksAtSlot` in this PR, will open a separate issue for that.

Just one clarification, we don't have `db.blocks.getBlockAtSlot` (maybe we used to have that in the past, I'm not sure). In the meantime we get blockRoot from forkchoice and get block based on blockRoot.

packages/lodestar/src/chain/chain.ts
```diff
@@ -110,6 +110,19 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
 
+  public async getUnfinalizedBlocksAtSlots(slots: Slot[]): Promise<SignedBeaconBlock[]> {
+    if (!slots) {
+      return [];
+    }
+    const blockRoots = slots.map((slot) => {
+      const summary = this.forkChoice.getBlockSummaryAtSlot(slot);
+      return summary? summary.blockRoot : null;
+    }).filter((blockRoot) => !!blockRoot);
+    // these blocks are on the same chain to head
```
twoeths, line 125: Since we get block roots from the same head, we're sure all blocks are on the same chain to head

mpetrunic (approved):
lgtm, although solving multiple fork blocks for same slot was probably more important than avoiding it :disappointed:

wemeetagain:
merge conflict, looks ready to merge otherwise

packages/lodestar/test/utils/mocks/chain/chain.ts
```diff
@@ -51,6 +51,13 @@ export class MockBeaconChain extends EventEmitter implements IBeaconChain {
     return this.epochCtx.copy();
   }
 
+  public async getUnfinalizedBlocksAtSlots(slots: Slot[]): Promise<SignedBeaconBlock[]|null> {
+    if (!slots) {
+      return [];
+    }
+    return await Promise.all(slots.map(this.getBlockAtSlot));
```
mpetrunic, line 58: We could rename `getBlockAtSlot` to `getCannonicalBlockAtSlot` and keet this method `getBlocksAtSlot`?

twoeths: > We could rename getBlockAtSlot to `getCannonicalBlockAtSlot`

this is good to me

> and keet this method getBlocksAtSlot

as this method only gets unfinalized blocks, should we keep it `getUnfinalizedBlocksAtSlots` to avoid any confusion? Or any better name?

mpetrunic: > as this method only gets unfinalized blocks, should we keep it getUnfinalizedBlocksAtSlots to avoid any confusion? Or any better name?

oh scratch that, thought this method is on forkchoice.

wemeetagain (approved):

PR #1006, opened 2020-06-12
Use fastStateTransition when producing block

### LINKED ISSUE #957: Minor cleanups and validator sanity tests (non-substantive)
Spec PR: https://github.com/ethereum/eth2.0-specs/pull/1745

twoeths:
## Goal
+ resolves #957 
+ Don't need to modify `is_candidate_block` because we query from db instead of having that function
+ Write `compute_new_state_root` to easier maintain the spec, use fastStateTransition
+ Don't need to modify `get_block_signature` because we get signing root from block when calculating signature already in our validator


wemeetagain (commented):
lgtm, minus file conflict

twoeths:
I implemented `getEpochContext()` in this PR

mpetrunic (approved):

wemeetagain (approved):

Issue #970, opened 2020-05-27
Investigate using bigints for all uint64 types


wemeetagain:
Currently, we represent `Slot` and `Epoch` as a `number`, even though it is defined as a `Uint64`. In practice this has been ok,because we start from 0 and increment these values slowly.

One wrinkle to that assumption, `FAR_FUTURE_EPOCH`.We get around that by representing it as `infinity`.

We've also had pain running ssz spec tests and fuzz tests that exercise the higher bits of epoch and slot.

We'd like to explore changing these types from `number` to `bigint`.

dapplion:
@wemeetagain Should internal types in the beacon-state-transition function that represent Unix timestamps be also converted to BigInt? For example in this function when should the BigInt -> Number casting happen if any?
```js
export function computeTimeAtSlot(config: IBeaconConfig, slot: Slot, genesisTime: Number64): Number64 {
  return genesisTime + Number(slot) * config.params.SECONDS_PER_SLOT;
}
```

wemeetagain:
I think any internal types that don't touch ssz-defined datastructures can remain numbers, if they are naturally so.

I think the spirit of this issue is to have a more literal conformance to https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/beacon-chain.md#custom-types which defines all numbers in eth2 datastructures as uint64.

dapplion:
Last attempt on https://github.com/ChainSafe/lodestar/pull/1011 required massive changes everywhere on the code, so it was discarded. Maybe it would be best to discuss the strategy on how to deal with key issues such as how to use bigint for array indexing beforehand

PR #1007, opened 2020-06-12
0.12.0 non-substantive change

### LINKED ISSUE #948: Fix some ValidatorIndex type casting (non-substantive)
https://github.com/ethereum/eth2.0-specs/pull/1714

### LINKED ISSUE #954: Remove redundant check in validate_on_attestation (Non-substantive)
Spec PR: https://github.com/ethereum/eth2.0-specs/pull/1755

twoeths:
## Goals
+ resolves #948 
+ resolves #954

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/util/seed.ts
+++ b/packages/lodestar-beacon-state-transition/src/util/seed.ts
@@ -6,7 +6,6 @@ import assert from "assert";
 import {hash} from "@chainsafe/ssz";
 import {
   Epoch,
-  ValidatorIndex,
   BeaconState,
   Bytes32,
 } from "@chainsafe/lodestar-types";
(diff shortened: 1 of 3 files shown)


mpetrunic (commented):
I would rather keep our ValidatorIndex type as it makes it easier if we would like to switch it to BigInt because of #970

packages/lodestar-beacon-state-transition/src/util/seed.ts
```diff
@@ -26,7 +25,7 @@ import {DomainType} from "../constants";
  */
 export function computeShuffledIndex(
   config: IBeaconConfig,
-  index: ValidatorIndex,
```
twoeths, line 29: @mpetrunic the meaning of this is not validator index so we have to make it as `number` to avoid the confusion, that's what the spec PR asked for.

mpetrunic (approved):

wemeetagain (approved):

PR #1008, opened 2020-06-14
Add prettier as an eslint plugin

dapplion:
### Rationale

A code formating tool homogenizes style across contributors, which help new ones match the repo style and stay consistent.

The setup in this PR runs prettier as an eslint plugin. In CI, formatting errors will be shown in the `lint` stage. Most of the issues found by the prettier plugin can be fixed with the `lint:fix` script. Alternatively prettier can be run directly across the entire repo with
```
./node_modules/.bin/prettier --write 'packages/**/*.{ts,md}'
```

Locally, contributors may choose to run it manually, before committing with a git hook (i.e. husky), or incorporating it into their IDE and trigger it every save.

I've selected the settings in `.prettierrc.js` config that generated the smallest git diff, so they had the widest adoption in the codebase. Specific settings that were contentious and may be discussed are:
- [`trailingComma`](https://prettier.io/docs/en/options.html#trailing-commas): `"es5"` - Print trailing commas when multi-line and valid in ES5 (objects, arrays, etc.).
- [`arrowParens`](https://prettier.io/docs/en/options.html#arrow-function-parentheses): `"always"` - Always include parens. Example: `(x) => x`. 
>*from prettier docs*: 
>At first glance, avoiding parentheses may look like a better choice because of less visual noise. However, when Prettier removes parentheses, it becomes harder to add type annotations, extra arguments or default values as well as making other changes. Consistent use of parentheses provides a better developer experience.

### How to incorporate

To incorporate the formatter I would recommend waiting for a moment with no open PRs, since it will generate a big diff (as of opening this PR `599 files changed, 5729 insertions(+), 6242 deletions(-)`).


wemeetagain:
This is great!

Personally, I can get behind both of those contentious rules. Lets see if we can get consensus from @ChainSafe/eth2-0 

I'm curious to hear everyone's opinions about how we can go about integrating this, whether simply plugging into the linter is sufficient, or if we need to do more. I think hooking into the linter, perhaps adding a lint-fix script would be sufficient.

As far as timing, lets plan on doing this after we merge our 0.12.x branch into master (hopefully in a week or so).

mpetrunic:
I'm fine with that after 0.12 merge. I'm against running it on git commit as with many files, running commit from some GUI, makes it somewhat unresponsive. CI already checks lint so locally we just need to run lint:fix or let connect IDE with eslint

PR #1010, opened 2020-06-15
Parse yargs option types from the default config values

### LINKED ISSUE #977: CLI: hex string params overrides being parsed as number
Create a lodecli configuration
Edit beacon.config.json, add "chain.params.GENESIS_FORK_VERSION": "0x00000113"
Run beacon node, should error w/ something like "value.replace is not a function"

The reason is we're not setting the option type in `lodestar-cli/src/lodecli/cmds/beacon/cmd/run/options/params.ts`, and yargs parses hex strings into numbers if a "type" is not set.

Quick fix would be to add something like `type: Buffer.isBuffer(params[key]) ? "string" : undefined`
Ideally, we would have a more systematic way of handling params of different types, rather than inspecting runtime values of mainnet config.

dapplion:
Implements the quick fix outlined by https://github.com/ChainSafe/lodestar/issues/977

Fixes https://github.com/ChainSafe/lodestar/issues/977

### DIFF
--- a/packages/lodestar-cli/src/lodecli/cmds/beacon/cmds/run/options/params.ts
+++ b/packages/lodestar-cli/src/lodecli/cmds/beacon/cmds/run/options/params.ts
@@ -9,7 +9,17 @@ Object.keys(params).forEach((key) => {
       `chain.params.${key}`,
     ],
     hidden: true,
+    type: yargsOptionType(params[key as keyof typeof params])
   };
 });
+
+
+function yargsOptionType(value: unknown): Options["type"] {
+  if (typeof value === "string" || Buffer.isBuffer(value)) return "string";
+  if (typeof value === "boolean") return "boolean";
+  if (typeof value === "number") return "number";
+  if (Array.isArray(value)) return "array";
+  else return undefined;
+}
 
 export const paramsOptions = options;


packages/lodestar-cli/src/lodecli/cmds/beacon/cmds/run/options/params.ts
```diff
@@ -9,7 +9,18 @@ Object.keys(params).forEach((key) => {
       `chain.params.${key}`,
     ],
     hidden: true,
+    type: yargsOptionType(params[key as keyof typeof params])
   };
 });
 
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function yargsOptionType(value: any): Options["type"] {
```
mpetrunic: might be better to use `unknown` type

dapplion: Fixed

packages/lodestar-cli/src/lodecli/cmds/beacon/cmds/run/options/params.ts
```diff
@@ -9,7 +9,18 @@ Object.keys(params).forEach((key) => {
       `chain.params.${key}`,
     ],
     hidden: true,
+    type: yargsOptionType(params[key as keyof typeof params])
   };
 });
 
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
mpetrunic: you can now remove eslint disable rule :smile:

mpetrunic (approved):

wemeetagain (approved):

PR #1011, opened 2020-06-15
Use bigint for all epoch and slot related types

### LINKED ISSUE #970: Investigate using bigints for all uint64 types

dapplion:
Uses javascript native `bigint` type to represent Epoch and Slot types. Motivated to fix the issues outlined in https://github.com/ChainSafe/lodestar/issues/970
> One wrinkle to that assumption, FAR_FUTURE_EPOCH.We get around that by representing it as infinity.
We've also had pain running ssz spec tests and fuzz tests that exercise the higher bits of epoch and slot.

No big structural changes are required. The least appealing is having to cast to Number to index arrays
```js
this.proposers[Number(slot % this.config.params.SLOTS_PER_EPOCH)]
```
Native `Map`s support indexing with BigInt, but I don't know if it's appropriate for the requirements of this codebase. I would appreciate your feedback.

Fixes https://github.com/ChainSafe/lodestar/issues/970

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/epoch/finalUpdates.ts
+++ b/packages/lodestar-beacon-state-transition/src/epoch/finalUpdates.ts
@@ -8,14 +8,14 @@ import {
   getCurrentEpoch,
   getRandaoMix
 } from "../util";
-import {bigIntMin, intDiv} from "@chainsafe/lodestar-utils";
+import {bigIntMin} from "@chainsafe/lodestar-utils";
 
 
 export function processFinalUpdates(config: IBeaconConfig, state: BeaconState): void {
   const currentEpoch = getCurrentEpoch(config, state);
-  const nextEpoch = currentEpoch + 1;
+  const nextEpoch = currentEpoch + 1n;
   // Reset eth1 data votes
-  if (nextEpoch % config.params.EPOCHS_PER_ETH1_VOTING_PERIOD === 0) {
+  if (nextEpoch % config.params.EPOCHS_PER_ETH1_VOTING_PERIOD === 0n) {
     state.eth1DataVotes = [];
   }
   // Update effective balances with hysteresis
@@ -34,12 +34,12 @@ export function processFinalUpdates(config: IBeaconConfig, state: BeaconState): 
     }
   });
   // Reset slashings
-  state.slashings[nextEpoch % config.params.EPOCHS_PER_SLASHINGS_VECTOR] = 0n;
+  state.slashings[Number(nextEpoch % config.params.EPOCHS_PER_SLASHINGS_VECTOR)] = 0n;
   // Set randao mix
-  state.randaoMixes[nextEpoch % config.params.EPOCHS_PER_HISTORICAL_VECTOR] =
+  state.randaoMixes[Number(nextEpoch % config.params.EPOCHS_PER_HISTORICAL_VECTOR)] =
     getRandaoMix(config, state, currentEpoch);
   // Set historical root accumulator
-  if (nextEpoch % intDiv(config.params.SLOTS_PER_HISTORICAL_ROOT, config.params.SLOTS_PER_EPOCH) === 0) {
+  if (nextEpoch % BigInt(config.params.SLOTS_PER_HISTORICAL_ROOT) / config.params.SLOTS_PER_EPOCH === 0n) {
     const historicalBatch: HistoricalBatch = {
       blockRoots: state.blockRoots,
       stateRoots: state.stateRoots,
--- a/packages/lodestar-beacon-state-transition/src/fast/epoch/processFinalUpdates.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/epoch/processFinalUpdates.ts
@@ -1,5 +1,5 @@
 import {BeaconState} from "@chainsafe/lodestar-types";
-import {bigIntMin, intDiv} from "@chainsafe/lodestar-utils";
+import {bigIntMin} from "@chainsafe/lodestar-utils";
 
 import {getRandaoMix} from "../../util";
 import {EpochContext, IEpochProcess} from "../util";
@@ -11,7 +11,7 @@ export function processFinalUpdates(
 ): void {
   const config = epochCtx.config;
   const currentEpoch = process.currentEpoch;
-  const nextEpoch = currentEpoch + 1;
+  const nextEpoch = currentEpoch + 1n;
   const {
     EFFECTIVE_BALANCE_INCREMENT,
     EPOCHS_PER_ETH1_VOTING_PERIOD,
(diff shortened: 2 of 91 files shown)


packages/lodestar-beacon-state-transition/src/epoch/finalUpdates.ts
```diff
@@ -34,12 +34,12 @@ export function processFinalUpdates(config: IBeaconConfig, state: BeaconState):
...
   // Reset slashings
-  state.slashings[nextEpoch % config.params.EPOCHS_PER_SLASHINGS_VECTOR] = 0n;
+  state.slashings[Number(nextEpoch % config.params.EPOCHS_PER_SLASHINGS_VECTOR)] = 0n;
   // Set randao mix
-  state.randaoMixes[nextEpoch % config.params.EPOCHS_PER_HISTORICAL_VECTOR] =
+  state.randaoMixes[Number(nextEpoch % config.params.EPOCHS_PER_HISTORICAL_VECTOR)] =
     getRandaoMix(config, state, currentEpoch);
   // Set historical root accumulator
-  if (nextEpoch % intDiv(config.params.SLOTS_PER_HISTORICAL_ROOT, config.params.SLOTS_PER_EPOCH) === 0) {
+  if (nextEpoch % BigInt(config.params.SLOTS_PER_HISTORICAL_ROOT) / config.params.SLOTS_PER_EPOCH === 0n) {
```
mpetrunic, line 42:
```suggestion
  if (nextEpoch % (BigInt(config.params.SLOTS_PER_HISTORICAL_ROOT) / config.params.SLOTS_PER_EPOCH) === 0n) {
```

packages/lodestar-beacon-state-transition/src/fast/epoch/processFinalUpdates.ts
```diff
@@ -49,13 +49,13 @@ export function processFinalUpdates(
...
-  state.slashings[nextEpoch % EPOCHS_PER_SLASHINGS_VECTOR] = BigInt(0);
+  state.slashings[Number(nextEpoch % EPOCHS_PER_SLASHINGS_VECTOR)] = BigInt(0);
 
   // set randao mix
-  state.randaoMixes[nextEpoch  % EPOCHS_PER_HISTORICAL_VECTOR] = getRandaoMix(config, state, currentEpoch);
+  state.randaoMixes[Number(nextEpoch  % EPOCHS_PER_HISTORICAL_VECTOR)] = getRandaoMix(config, state, currentEpoch);
 
   // set historical root accumulator
-  if (nextEpoch % intDiv(SLOTS_PER_HISTORICAL_ROOT, SLOTS_PER_EPOCH) === 0) {
+  if (nextEpoch % SLOTS_PER_HISTORICAL_ROOT / SLOTS_PER_EPOCH === 0n) {
```
mpetrunic, line 58:
```suggestion
  if (nextEpoch % (SLOTS_PER_HISTORICAL_ROOT / SLOTS_PER_EPOCH) === 0n) {
```

packages/lodestar-types/src/ssz/interface.ts
```diff
@@ -20,8 +20,8 @@ export interface IBeaconSSZTypes {
   Uint64: BigIntUintType;
   Uint128: BigIntUintType;
   Uint256: BigIntUintType;
-  Slot: NumberUintType;
-  Epoch: NumberUintType;
+  Slot: BigIntUintType;
+  Epoch: BigIntUintType;
   CommitteeIndex: NumberUintType;
   ValidatorIndex: NumberUintType;
```
mpetrunic, line 26: Both CommitteeIndex and ValidatorIndex should be BitIntUintType.
One of the goals of this change was to prevent api response discrepancies ( we serialize bigints as string in json while numbers remain native)

mpetrunic (commented):
We should probably use `BigInt(1)` instead of `1n` to allow polyfilling in browsers that don't support it.
cc @wemeetagain ^^

You will have to fix type errors in other packages so it can pass ci (spec tests primarily) :slightly_frowning_face:

dapplion:
@mpetrunic I've converted to draft until I fix the types for all packages, sorry about that.

However, I believe we should evaluate performance before considering this PR. I've done a quick test to check the performance difference of BigInt math vs Number math and is significant

**Number addition**
```js
const count = 100000;
const label = String(count);
console.time(label);
let value = 1231254;
for (let i=0; i<count; i++) {
  value = value + i;
}
console.timeEnd(label);
```
```
100000: 3.381ms
100000: 4.139ms 
100000: 3.102ms
100000: 3.009ms
100000: 3.136ms
```
**BigInt addition**
```js
const count = 100000n;
const label = String(count);
console.time(label);
let value = 1231254n;
for (let i=0n; i<count; i++) {
  value = value + i;
}
console.timeEnd(label);
```
```
100000: 13.202ms
100000: 12.874ms 
100000: 14.014ms
100000: 14.134ms
100000: 13.600ms
100000: 13.969ms
100000: 13.879ms
```

Do you believe this can have a significant impact on the overall performance of Lodestar?

mpetrunic:
> I've converted to draft until I fix the types for all packages, sorry about that.

No worries.

> Do you believe this can have a significant impact on the overall performance of Lodestar?

Not sure, hopefully not since there isn't much bigint addition, rewards and penalties are using numbers. You could try running spec tests (mainnet) and see before and after.

wemeetagain:
> We should probably use `BigInt(1)` instead of `1n` to allow polyfilling in browsers that don't support 

Not sure this is a big issue here. In a few of our dependencies, eg: ssz, this was a bigger deal, as _it_ has optional dependencies that use BigInt. So avoiding BigInt literals in those subdependencies opened up a subset of functionality to non-BigInt-supporting browsers.
Here, there's no hope if your environment doesn't support BigInt. You _will_ hit a runtime error because BigInts are used in many places (even without this PR).

This whole issue should be resolved as browsers become compliant with ES2020. (The biggest current offender is Safari)

> > Do you believe this can have a significant impact on the overall performance of Lodestar?

> Not sure, hopefully not since there isn't much bigint addition, rewards and penalties are using numbers. 
> You could try running spec tests (mainnet) and see before and after.

I tend to think there will be a noticeable performance penalty, and we'll just have to make a decision whether its worth it or not when we see it. Agreed that benchmarking the spec tests would be a decent way of exercising the code.

> One of the goals of this change was to prevent api response discrepancies ( we serialize bigints as string in json while numbers remain native)

I'll make a separate issue (in the ssz repo) to just tackle just this issue, should be fairly trivial to standardize our handling of JSON<->uint conversion (https://github.com/ChainSafe/ssz/issues/34)

wemeetagain:
It looks like ssz uint64 number and bigint values are standardized to both be `toJson(value: uint64): string` already

wemeetagain:
Closing this PR until more thorough benchmarking is done

PR #1012, opened 2020-06-15
Add missing LICENSE files

### LINKED ISSUE #968: Add LICENSE files to all packages
Just noticed that not all of our packages in this monorepo have LICENSE files (even tho the license is stated in the READMEs and package.json). It would be good to include LICENSE files in each package to avoid ambiguity.

dapplion:
Fixes https://github.com/ChainSafe/lodestar/issues/968

I've respected the license as stated in each package, but I like to assert that the license of each package aligns with what the team wants:

| package | License |
| --- | --- |
| benchmark-utils | Apache-2.0
| lodestar | LGPL-3.0
| lodestar-beacon-state-transition | Apache-2.0
| lodestar-cli | LGPL-3.0
| lodestar-config | Apache-2.0
| lodestar-params | Apache-2.0
| lodestar-spec-test-util | Apache-2.0
| lodestar-types | Apache-2.0
| lodestar-utils | Apache-2.0
| lodestar-validator | LGPL-3.0
| spec-test-runner | LGPL-3.0

### DIFF
--- a/packages/lodestar-beacon-state-transition/LICENSE
+++ b/packages/lodestar-beacon-state-transition/LICENSE
@@ -1,0 +1,165 @@
+                   GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+
+  This version of the GNU Lesser General Public License incorporates
+the terms and conditions of version 3 of the GNU General Public
+License, supplemented by the additional permissions listed below.
+
+  0. Additional Definitions.
+
+  As used herein, "this License" refers to version 3 of the GNU Lesser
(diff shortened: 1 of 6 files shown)


mpetrunic:
@GregTheGreek I think you should take a look at this

packages/lodestar-beacon-state-transition/LICENSE
```diff
@@ -0,0 +1,165 @@
+                   GNU LESSER GENERAL PUBLIC LICENSE
```
wemeetagain, line 1: I think this one should be apache

wemeetagain (changes requested):
Thanks for tackling this

packages/lodestar-beacon-state-transition/LICENSE
GregTheGreek: I think we'll keep it how it is,,and explore dual licensing things.

wemeetagain (approved):

GregTheGreek (approved):

PR #1014, opened 2020-06-16
Replace NodeJS's assert with a minimal implementation

### LINKED ISSUE #929: Use minimal assert implementation
idea:
- write a minimalist `assert` implementation, eg: `function assert(condition: boolean, message?: string): void;`
- store in lodestar-utils
- use everywhere we use lodestar-utils instead of the current assert library

dapplion:
Note: In a few instances `assert.equal` were used instead of just `assert`. The main difference is that in case of error, the error would contain the expected and actual value. Should this behavior be respected in this PR?

Would it be in the scope of this PR to add assertion messages to all assert() calls?

Fixes https://github.com/ChainSafe/lodestar/issues/929

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/block/blockHeader.ts
+++ b/packages/lodestar-beacon-state-transition/src/block/blockHeader.ts
@@ -2,9 +2,9 @@
  * @module chain/stateTransition/block
  */
 
-import assert from "assert";
 import {BeaconBlock, BeaconState} from "@chainsafe/lodestar-types";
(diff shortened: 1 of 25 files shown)


mpetrunic (approved):
Adding messages to assert statement would be awesome and help with debugging errors but I think we can open separate issue for that. As it would include a lot of changes.

wemeetagain (approved):
This is 

Edit: whoops, submitted message too soon. Meant to say: This is perfect

PR #1022, opened 2020-06-16
Fix api array query

### LINKED ISSUE #1015: Fix api array query params
Instead of using common syntax `?param1=1&param1=2`, we are using `?param1[]=1,2`

mpetrunic:
- it fixes bug where single item in query would not be array even though schemas says it should be
- drops param[]=value syntax

resolves #1015


wemeetagain (approved):
lgtm

PR #1023, opened 2020-06-16
Abstract query stringify in the validator REST API

dapplion:
It should help with readability and maintainability. If the NodeJS `querystring` module is not appropriate for use I can substitute with a minimal implementation.

### DIFF
--- a/packages/lodestar-validator/src/util/httpClient.ts
+++ b/packages/lodestar-validator/src/util/httpClient.ts
@@ -7,6 +7,10 @@ export interface IHttpClientOptions {
   urlPrefix: string;
 }
 
+export interface IHttpQuery {
+  [key: string]: string | number | boolean | string[] | number[];
+}
+
 export class HttpClient {
   private client: AxiosInstance;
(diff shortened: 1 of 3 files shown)


wemeetagain (dismissed):
lgtm :+1:

packages/lodestar-validator/src/util/httpClient.ts
```diff
@@ -17,8 +18,9 @@ export class HttpClient {
     this.logger = logger;
   }
 
-  public async get<T>(url: string): Promise<T> {
+  public async get<T>(url: string, query?: querystring.ParsedUrlQueryInput): Promise<T> {
```
mpetrunic: What do you think about passing query object instead of ParsedUrlQueryInput?
 That way we could easily switch how we serialize query.

mpetrunic (commented):
It would probably make sense to add this to all httpClient methods along with tests

wemeetagain (approved):

twoeths (approved):

PR #1024, opened 2020-06-16
Add fastify hook to convert querystring keys to camelcase

### LINKED ISSUE #1021: Convert api query params to camel case

According to api spec, query params are snake-case which means in our api controllers we have to deal with snake case params which are against eslint rule. 

Fastify uses ajv schemas to validate and coerce data. Ideally, there would be some way to automatically convert query params to camel case.

We could write fastify plugin that would intercept query params and convert them to snakeCase

**Additional context**

dapplion:
Adds a step in the [fastify lifecycle](https://www.fastify.io/docs/latest/Lifecycle/) to mutate the query object before hitting the schema validation. The hook adds a cameCase version of each keys in the query string object.

Fixes https://github.com/ChainSafe/lodestar/issues/1021

Personally I will stick to snake case query string params if that is what the spec is using for clarity. Now there is a non-obvious step that might be problematic in the future.

### DIFF
--- a/packages/lodestar/src/api/rest/index.ts
+++ b/packages/lodestar/src/api/rest/index.ts
@@ -2,6 +2,7 @@ import * as fastify from "fastify";
 import {FastifyInstance} from "fastify";
 import fastifyCors from "fastify-cors";
 import {IService} from "../../node";
+import {camelcase} from "../../util/camelcase"
 import {IRestApiOptions} from "./options";
 import {ILogger} from "@chainsafe/lodestar-utils/lib/logger";
 import * as routes from "./routes";
@@ -70,6 +71,16 @@ export class RestApi implements IService {
       // applySyncingMiddleware(server, "/validator/*", modules);
       server.register(routes.validator, {prefix: "/validator", api, config: modules.config});
     }
+
+    // Make querystring keys also available in camelCase
+    server.addHook("preValidation", (request, reply, done) => {
+      if (typeof request.query === "object") {
+        for (const key in request.query) {
+          request.query[camelcase(key)] = request.query[key];
+        }
+      }
+      done();
+    });
 
     return server;
   }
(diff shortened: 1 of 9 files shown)


wemeetagain:
> Personally I will stick to snake case query string params if that is what the spec is using for clarity. Now there is a non-obvious step that might be problematic in the future.

I tend to agree. If eslint is giving errors, we can just disable that rule for those files. Better than adding a runtime step to transform the object keys. We can keep the underscore_case naming isolated to the api layer.

mpetrunic:
> Personally I will stick to snake case query string params if that is what the spec is using for clarity. Now there is a non-obvious step that might be problematic in the future.

@dapplion is there post validation hook? I would like to keep validation schema according to spec. snake to camel should be just our internal handling

> I tend to agree. If eslint is giving errors, we can just disable that rule for those files. Better than adding a runtime step to transform the object keys. We can keep the underscore_case naming isolated to the api layer.

We will quickly have bunch of eslint disable comments which is not really ideal and it's quite annoying while coding

wemeetagain:
> We will quickly have bunch of eslint disable comments which is not really ideal and it's quite annoying while coding

I think we can use file-level disables for this in all the rest api files, eg:
```typescript
/* eslint-disable camelcase */

...

const opts: fastify.RouteShorthandOptions = {
  non_camel_1: {},
  non_camel_2: {},
}
```
We would then only need one eslint-disable statement per file.
This seems like a fair tradeoff, the api is standardized to go against our internal conventions, so this boundary will just naturally have these issues.

mpetrunic:
> We would then only need one eslint-disable statement per file.
> This seems like a fair tradeoff, the api is standardized to go against our internal conventions, so this boundary will just naturally have these issues.

Whats your cons for not auto converting to camelCase internally? I mean we can have official validation and schemas as per spec, we will just get camelCase params in controller

dapplion:
> Whats your cons for not auto converting to camelCase internally? I mean we can have official validation and schemas as per spec, we will just get camelCase params in controller

Even if at the cost of verbosity, code that is painfully obvious to reason about is better months / years down the line than anything that may require non-obvious knowledge. It's a decision I've always regretted when coming back to an old codebase and having to spent more time than necessary.

wemeetagain:
Closing this in favor of fixing this with selective disabling of the eslint rule

PR #1025, opened 2020-06-17
Start MetadataController with an enr

### LINKED ISSUE #989: Syncing Witti: Don't see `eth2` being inserted in `enr.json`
## Actual:
+ After I sync Witti to slot ~1000, I stop and check `enr.json`, I don't see `eth2` key being inserted there.

## Expect:
+ See `eth2` key in `enr.json` after we find genesis state ourself, or load genesis state from a file
+ Fork digest should be `f6775d07`, fork version should be `00000113` for Witti

twoeths:
fixes #989 
## Cause
+ ENR is not available at the time of creating instance (constructor)
+ It's only available after libp2 started so make it `start(enr)`

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -51,7 +51,6 @@ export interface IBlockProcessJob {
   reprocess: boolean;
 }
 
-const MAX_VERSION = Buffer.from([255, 255, 255, 255]);
 export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) implements IBeaconChain {
 
   public readonly chain: string;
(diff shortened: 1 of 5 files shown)


packages/lodestar/src/chain/chain.ts
```diff
@@ -204,7 +203,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
       fork => this.config.types.Version.equals(currentVersion, intToBytes(fork.previousVersion, 4)));
     return {
       forkDigest: this.currentForkDigest,
-      nextForkVersion: nextVersion? intToBytes(nextVersion.currentVersion, 4) : MAX_VERSION,
+      nextForkVersion: nextVersion? intToBytes(nextVersion.currentVersion, 4) : currentVersion.valueOf() as Uint8Array,
```
twoeths, line 206: this is to address spec: `If no future fork is planned, set next_fork_version = current_fork_version to signal this fact`

twoeths:
Tested with Witti and `enr view --kv` in `rumor`
```
[DEFAULT_ACTOR][s1_19]INFO Decoded ENR key-value pairs                   enr_kv="map[eth2:f6775d0700000113ffffffffffff1f00 id:v4 ip:118.69.144.196 secp256k1:16483116747c1d100dcbc3607f6f278540e20d4d9d0676d17d0c18deb3a26ab49c4441236aea2e7df5d28f3bfe46ccacc85d4c27098ddf9f00e367ad951d3ae8 udp:9000]"
```

wemeetagain (approved):
Great!

mpetrunic (approved):
lgtm.

We don't have flow to start networking and discovery after we found genesis?

twoeths:
> We don't have flow to start networking and discovery after we found genesis?

We do start networking and discovery after we found genesis and this PR works thanks to that. Earlier we set `enr` in `MetadataController` constructor and it was not available at that time. This PR simply set `enr` when we start `MetadataController` instead of the constructor.

PR #1026, opened 2020-06-17
Fix epoch context copy

mpetrunic:
We broke our client, epochCty copy corrupts pubkey2index mapping because of that hidden implicit array => string conversion.


packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts
```diff
@@ -56,7 +57,11 @@ export class EpochContext {
   public copy(): EpochContext {
     const ctx = new EpochContext(this.config);
     // full copy of pubkeys, this can mutate
-    ctx.pubkey2index = new PubkeyIndexMap(this.pubkey2index.entries());
+    // ctx.pubkey2index = new PubkeyIndexMap(this.pubkey2index.entries());
```
wemeetagain: can remove this line

wemeetagain:
can we override PubkeyIndexMap.entries instead?

wemeetagain (approved):
https://discordapp.com/channels/593655374469660673/593655641445367808/722913119864750120
for posterity

PR #1028, opened 2020-06-18
Avoid deep copy pubkey2index map in EpochContext.copy()

### LINKED ISSUE #1027: Don't deep copy pubkey map
Copying the validator registry on every EpochContext data access outside the chain module is not sustainable.

Solution TBD

See https://discordapp.com/channels/593655374469660673/593655641445367808/722918727137493011
from the chat:
> You can return a read-only interface, and avoid that encapsulation problem
[4:04 PM]
And then just maintain a single dictionary that you never copy
[4:05 PM]
And then you have two options: either special-case when eth1 has a deep-reorg and we have much bigger things to worry about (lighthouse), or do a layered approach, where you wrap the old dictionary, and only use the common part that is still the same between deep reorgs (zrnt).
[4:06 PM]
It saves you a lot of time and allocations to not copy the full registry all the time (the registry is mostly pubkey data)

twoeths:
resolves #1027  with Proxy approach:
+ Share the same pubkey2Index map in EpochContext.copy()
+ pubkey2Index is now a getter method returning pubkey2index map proxy with `get` method only

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts
@@ -1,4 +1,4 @@
-import {ByteVector, toHexString, hash, fromHexString} from "@chainsafe/ssz";
+import {hash} from "@chainsafe/ssz";
 import {ValidatorIndex, Epoch, BeaconState, Slot, CommitteeIndex} from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {intToBytes} from "@chainsafe/lodestar-utils";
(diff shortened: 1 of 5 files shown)


mpetrunic:
I would rather go in direction of exposing method to read required data like:
```typescript
class EpochCtx {
   getValidatorIndex(pubkey): ValidatorIndex;

 // add methods for other data similary
}
```
wdyt @wemeetagain ?

wemeetagain:
@mpetrunic Yeah I think that would suffice. A get/set on EpochContext.
I don't think we should be engineering our node defensively at every level against internal data manipulation, or it will likely lead to an unmaintainable solution. I don't think we should be copying the EpochContext in every chain.getEpochContext() call, either. (We still should figure out how to copy the epoch context, w/o deep copying the validator indices tho)

As it stands, I could feed random data to the db controller, fork choice, gossip peers, etc and corrupt the node in a thousand different evil ways, given a reference to the beacon node. Not to say that this isn't an important issue, its just something that's better tackled at some boundary, at one layer, all at once, and optionally. (fake example solution: we have a SafeApi module that safely exposes the API w/o giving access to the underlying modules. We keep the reference to a beacon node in a closure, and only return the SafeApi.)

Here, I think the danger of overwriting / corrupting the pubkey map is in the case of deep eth1 reorgs. So whatever solution we figure, it should have that in mind. The solutions I've seen for that are either 1. ignore/explode on deep eth1 reorg. 2. special case on deep eth1 reorg (eg: recompute cache) 3. hierarchical approach as here: https://github.com/protolambda/zrnt/blob/master/eth2/beacon/epochs_context.go#L95
I'm leaning toward 1 or 2 atm since this is a highly unlikely scenario.

wemeetagain:
We might use something like https://github.com/immutable-js/immutable-js if we wanted to properly handle conflicts / deep copying

twoeths:
Thanks for clarifying that.

PR #1030, opened 2020-06-18
Error handling

mpetrunic:
- set fastify error handler to log unintentional errors
- add api logger with fastify adapter
- handler api rejections in lodestar-validator


packages/lodestar/src/api/rest/logger/fastify.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  public readonly serializers = {
+    req: (request: IncomingMessage&FastifyRequest) => {
+      return {
+        msg: `Request: ${request.ip} -> ${request.hostname}\t${request.method}:${request.url}\tRequestId: ${request.id}`
+      };
+    },
+  };
+
+  private winston: ILogger;
```
wemeetagain: I think should be called something like `log` or `logger` since this is generic `ILogger`

wemeetagain (changes requested):

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -169,10 +189,14 @@ export class AttestationService {
...
-    this.logger.info(
-      `Published signed aggregatte and proof for committee ${duty.committeeIndex} at slot ${duty.attestationSlot}`
-    );
+    try {
+      await this.provider.validator.publishAggregateAndProof(signedAggregateAndProof);
+      this.logger.info(
+        `Published signed aggregate and proof for committee ${duty.committeeIndex} at slot ${duty.attestationSlot}`
+      );
+    } catch (e) {
+      this.logger.error("Failed to publish signed aggregate and proof");
```
twoeths: Log error here

wemeetagain (approved):

twoeths (approved):

PR #1032, opened 2020-06-19
Readonly interface for EpochContext

twoeths:
Part of #1027 
+ Create readonly interfaces for EpochContext
+ chain.getEpochContext() should return that readonly interface instead of a copy

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/fast/util/epochShuffling.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/util/epochShuffling.ts
@@ -5,6 +5,14 @@ import {intDiv} from "@chainsafe/lodestar-utils";
 import {getSeed, unshuffleList} from "../../util";
 import {DomainType} from "../../constants";
 
+/**
+ * Readonly interface for IEpochShuffling.
+ */
+export interface IReadonlyEpochShuffling {
+  readonly epoch: Epoch;
+  readonly committees: Readonly<ValidatorIndex[][][]>;
+}
+
 export interface IEpochShuffling {
   /**
    * Epoch being shuffled
--- a/packages/lodestar-beacon-state-transition/src/fast/util/interface.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/util/interface.ts
@@ -1,0 +1,15 @@
+/* eslint-disable @typescript-eslint/interface-name-prefix */
+import {IReadonlyEpochShuffling} from ".";
+import {ValidatorIndex, Slot} from "@chainsafe/lodestar-types";
+import {ByteVector} from "@chainsafe/ssz";
+
+/**
+ * Readonly interface for EpochContext.
+ */
+export interface ReadonlyEpochContext {
+  readonly pubkey2index: ReadonlyMap<ByteVector, ValidatorIndex>;
+  readonly index2pubkey: Readonly<Uint8Array[]>;
+  readonly currentShuffling?: IReadonlyEpochShuffling;
+  readonly previousShuffling?: IReadonlyEpochShuffling;
+  getBeaconProposer: (slot: Slot) => ValidatorIndex;
+}
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -18,6 +18,7 @@ import {
 } from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {computeEpochAtSlot, computeForkDigest, EpochContext} from "@chainsafe/lodestar-beacon-state-transition";
+import {ReadonlyEpochContext} from "@chainsafe/lodestar-beacon-state-transition/lib/fast/util";
 import {ILogger} from "@chainsafe/lodestar-utils/lib/logger";
 import {intToBytes} from "@chainsafe/lodestar-utils";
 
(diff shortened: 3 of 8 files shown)


packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts
```diff
@@ -112,24 +113,40 @@ export class EpochContext {
...
 
+  public getValidatorIndex(pubkey: BLSPubkey): ValidatorIndex | undefined {
+    return this._pubkey2index.get(pubkey);
+  }
+
+  public getPublicKey(index: ValidatorIndex): Uint8Array {
+    return this._index2pubkey[index];
+  }
+
+  public get currentShuffling(): PartialEpochShuffling {
```
twoeths: This getter keeps references unchanged

packages/lodestar-beacon-state-transition/src/fast/util/epochShuffling.ts
```diff
@@ -5,21 +5,42 @@ import {intDiv} from "@chainsafe/lodestar-utils";
 import {getSeed, unshuffleList} from "../../util";
 import {DomainType} from "../../constants";
 
-export interface IEpochShuffling {
+export class PartialEpochShuffling {
```
twoeths: Readonly interface for EpochShuffling

packages/lodestar-beacon-state-transition/src/fast/util/interface.ts
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable @typescript-eslint/interface-name-prefix */
+import {PartialEpochShuffling} from ".";
+import {BLSPubkey, ValidatorIndex, Slot} from "@chainsafe/lodestar-types";
+
+export interface PartialEpochContext {
```
twoeths: Readonly interface for EpochContext

wemeetagain:
Thanks for putting this together.
After looking a little deeper, I think this might be still more than we need.
If we're just wanting to adjust the interface to be readonly, we could just do something like:
```typescript
class BeaconChain {
  ...
  public getEpochContext(): PartialEpochContext {
    return this.epochCtx as PartialEpochContext;
  }
}
```
and define the interface of PartialEpochContext to have `readonly` elements. Have the readonly characteristics completely enforced on the typescript interface level.

wemeetagain:
Do we actually need these separate getter functions? (sorry XD)

I think we can achieve the typescript-level safety with something like
```typescript
export interface ReadonlyEpochContext {
  readonly pubkey2Index: Readonly<Map<ByteVector, ValidatorIndex>>;
  readonly index2pubkey: Readonly<Uint8Array[]>;
  readonly previousShuffling ....
}
```
perhaps with inner Readonly-defined interfaces as needed
then
```typescript
// chain.ts
getEpochContext(): ReadonlyEpochContext {
  this.epochCtx as ReadonlyEpochContext;
}
```

packages/lodestar/src/chain/chain.ts
```diff
@@ -97,8 +98,8 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
     return this.db.block.get(this.forkChoice.headBlockRoot());
   }
 
-  public getEpochContext(): EpochContext {
-    return this.epochCtx.copy();
+  public getEpochContext(): ReadonlyEpochContext {
```
twoeths, line 101: I dont have to cast to ReadonlyEpochContext ;)

packages/lodestar-beacon-state-transition/src/fast/util/epochShuffling.ts
```diff
@@ -5,6 +5,14 @@ import {intDiv} from "@chainsafe/lodestar-utils";
 import {getSeed, unshuffleList} from "../../util";
 import {DomainType} from "../../constants";
 
+/**
+ * Readonly interface for IEpochShuffling.
+ */
+export interface IReadonlyEpochShuffling {
```
mpetrunic, line 11: you could add something like this:
```typescript
type Readonly<T> = {
    readonly [P in keyof T]: T[P];
}
export type ReadonlyEpochContext = ReadOnly<EpochContext>
```

packages/lodestar-beacon-state-transition/src/fast/util/interface.ts
```diff
@@ -0,0 +1,15 @@
+/* eslint-disable @typescript-eslint/interface-name-prefix */
+import {IReadonlyEpochShuffling} from ".";
+import {ValidatorIndex, Slot} from "@chainsafe/lodestar-types";
+import {ByteVector} from "@chainsafe/ssz";
+
+/**
+ * Readonly interface for EpochContext.
+ */
+export interface ReadonlyEpochContext {
```
mpetrunic, line 9: you could add something like this:
```typescript
type Readonly<T> = {
    readonly [P in keyof T]: T[P];
}
export type ReadonlyEpochContext = ReadOnly<EpochContext>;
```

twoeths: yeah but I want to enforce `currentShuffling` to be `IReadonlyEpochShuffling` not `IEpochShuffling` ...

mpetrunic: You could create:
```typescript
type DeepReadonly<T> = {
    readonly [P in keyof T]: DeepReadonly<T[P]>;
}
```

mpetrunic: this way, everything should be readonly without extra interfaces and casting

twoeths: I failed to do that with this error
```
@chainsafe/lodestar:   Type 'DeepReadonly<(key: Vector<number>) => number>' has no call signatures.
@chainsafe/lodestar: 49     const index = this.chain.getEpochContext().pubkey2index.get(pubkey);
```

also, I want to use ReadonlyMap<> instead of Readonly<Map<>> to avoid write operations to the map.

wemeetagain (approved):

PR #1034, opened 2020-06-19
Remove event listeners

### LINKED ISSUE #1029: Handle stream api closing
On our stream endpoints we have following logic:
- register event listener
- convert event listener into async generator
- convert generator into stream

That works until api consumer closes connection in which case we get:
`2020-06-18 12:10:48  [API]              warn: response terminated with an error with headers already sent `
and we are left we dangling event listener.

Solution:
From stream api implementation should return callback which can be invoked once `fastifyRequest.raw.on("close")` occurs which in term should stop EventIterator and unregister event listener.

dapplion:
- Unsubscribe from processedBlock on EventIterator remove. Fixes https://github.com/ChainSafe/lodestar/issues/1029
- Remove processedBlock listener on first event in waitForBlockProcessed

### DIFF
--- a/packages/lodestar/src/api/impl/beacon/beacon.ts
+++ b/packages/lodestar/src/api/impl/beacon/beacon.ts
@@ -92,7 +92,7 @@ export class BeaconApi implements IBeaconApi {
   }
 
   public getBlockStream(): AsyncIterable<SignedBeaconBlock> {
-    return new EventIterator<SignedBeaconBlock>((push) => {
+    return new EventIterator<SignedBeaconBlock>(({push}) => {
       this.chain.on("processedBlock", push);
       return () => this.chain.off("processedBlock", push);
     });
--- a/packages/lodestar/src/api/impl/beacon/beacon.ts
+++ b/packages/lodestar/src/api/impl/beacon/beacon.ts
@@ -93,9 +93,8 @@ export class BeaconApi implements IBeaconApi {
 
   public getBlockStream(): AsyncIterable<SignedBeaconBlock> {
     return new EventIterator<SignedBeaconBlock>((push) => {
-      this.chain.on("processedBlock", (block) => {
-        push(block);
-      });
+      this.chain.on("processedBlock", push);
+      return () => this.chain.off("processedBlock", push);
     });
   }
 }
--- a/packages/lodestar/package.json
+++ b/packages/lodestar/package.json
@@ -57,7 +57,7 @@
     "deepmerge": "^3.2.0",
     "es6-promisify": "6.0.2",
     "ethers": "^5.0.2",
-    "event-iterator": "^1.2.0",
+    "event-iterator": "^2.0.0",
     "fastify": "2.13.0",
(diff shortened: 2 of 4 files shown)


packages/lodestar/src/api/impl/beacon/beacon.ts
```diff
@@ -93,9 +93,8 @@ export class BeaconApi implements IBeaconApi {
 
   public getBlockStream(): AsyncIterable<SignedBeaconBlock> {
     return new EventIterator<SignedBeaconBlock>((push) => {
-      this.chain.on("processedBlock", (block) => {
-        push(block);
-      });
+      this.chain.on("processedBlock", push);
+      return () => this.chain.off("processedBlock", push);
```
mpetrunic, line 97: this won't be called unless you call stop (second argument in constructor which is awkward.

dapplion: Maybe we can update to event-iterator 2.0.0 which seems to have a simpler API https://www.npmjs.com/package/event-iterator#advanced-examples

mpetrunic: Yeah sure ( I think we have fork endpoint with this as well). I just briefly checked, but not sure if simpler api helps you.
It would be ideal if EventIntertor class would expose stop method. Maybe you can extend it with that method?
In endpoint implementation, you would get it and call `req.raw.on("close", () => lodestarEventIterator.stop())`

dapplion: So you mean that it's `fastify-sse-v2` who is not handling the close event properly? Note that the plugin is adding types and headers by itself https://github.com/NodeFactoryIo/fastify-sse-v2/blob/6a91fedb4a4aa82d1fd13f6cb14909cc028d8ed5/src/plugin.ts#L14 so this redundancy may be the cause?

mpetrunic: plugin doesn't have access to req so it cannot detect req closing, Once request is closed, fastify closes stream, but I cannot notify event-iterator to stop as I only get async iterable

mpetrunic: https://github.com/NodeFactoryIo/fastify-sse-v2/issues/2

wemeetagain (approved):

mpetrunic (changes requested):
This does not yet solve our issue. stop callback of EventIterator must be called on request close

dapplion:
> This does not yet solve our issue. stop callback of EventIterator must be called on request close

@mpetrunic  If you can please take over this PR as it fits more your realm of expertise.

mpetrunic:
> > This does not yet solve our issue. stop callback of EventIterator must be called on request close
> 
> @mpetrunic If you can please take over this PR as it fits more your realm of expertise.

sure :slightly_smiling_face:

mpetrunic:
closing in favor of https://github.com/ChainSafe/lodestar/pull/1050

PR #1035, opened 2020-06-19
SSZ types for p2p spec

### LINKED ISSUE #997: Fix invalid unbound SSZ list types
https://github.com/ethereum/eth2.0-specs/pull/1837

twoeths:
resolves #997 

+ If p2p response an error, we want to return error message along with status code
+ New P2pErrorMessage ssz type for p2p error
+ `MAX_REQUEST_BLOCKS` for beacon_block_by_range is enforced by validating the request
+ `MAX_REQUEST_BLOCKS` for beacon_block_by_root is enforced by ssz max size validation


packages/lodestar/src/network/encoders/response.ts
```diff
@@ -112,9 +122,25 @@ export function eth2ResponseDecode(
           }
         }
       }
+      if(status !== RpcResponseStatus.SUCCESS) {
```
twoeths, line 125: should we throw error here to mark an unsuccessful req/resp? that would make it the same treatment to others

wemeetagain (approved):
lgtm

PR #1042, opened 2020-06-23
Build genesis state incrementally

### LINKED ISSUE #883: Optimize Eth1 Genesis

twoeths:
resolves #883 

+ Build genesis state incrementally using newly created GenesisBuilder, it should be a good preparation for mainnet with `MIN_GENESIS_ACTIVE_VALIDATOR_COUNT > 16000`
+ Start processing eth1 blocks/events from either genesis builder or chain
+ Process eth1 blocks/events one by one using pipe instead of event

This doesn't address the following issues/concerns (should have a separate issue for it)
+ ethers.ts should be stateless, no db should be involved there. Create a separate service to manage that.
+ No need to store proposing data during initial sync

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -25,7 +25,7 @@ import {EMPTY_SIGNATURE, GENESIS_SLOT} from "../constants";
 import {IBeaconDb} from "../db";
 import {IEth1Notifier} from "../eth1";
 import {IBeaconMetrics} from "../metrics";
-import {getEmptyBlock, initializeBeaconStateFromEth1, isValidGenesisState} from "./genesis/genesis";
+import {getEmptyBlock, initializeBeaconStateFromEth1, isValidGenesisState, GenesisBuilder} from "./genesis/genesis";
 import {ILMDGHOST, ArrayDagLMDGHOST} from "./forkChoice";
 
 import {ChainEventEmitter, IAttestationProcessor, IBeaconChain} from "./interface";
@@ -130,6 +130,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
     this._currentForkDigest =  computeForkDigest(this.config, state.fork.currentVersion, state.genesisValidatorsRoot);
     this.on("forkDigestChanged", this.handleForkDigestChanged);
     await this.restoreHeadState(state);
+    await this.eth1.startProcessEth1Blocks();
   }
 
   public async stop(): Promise<void> {
@@ -309,16 +310,10 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
     let state: TreeBacked<BeaconState> = await this.db.stateArchive.lastValue();
     if (!state) {
       this.logger.info("Chain not started, listening for genesis block");
-      state = await new Promise((resolve) => {
-        const genesisListener = async (timestamp: number, eth1Data: Eth1Data): Promise<void> => {
-          const state = await this.checkGenesis(timestamp, eth1Data);
-          if (state) {
-            this.eth1.removeListener("eth1Data", genesisListener);
-            resolve(state);
-          }
-        };
-        this.eth1.on("eth1Data", genesisListener);
-      });
+      const builder = new GenesisBuilder(this.config, {eth1: this.eth1, db: this.db, logger: this.logger});
+      state = await builder.waitForGenesis();
+      await this.initializeBeaconChain(state);
+      this.logger.info(`Genesis state is ready with ${state.validators.length} validators`);
     }
     // set metrics based on beacon state
     this.metrics.currentSlot.set(state.slot);
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -9,7 +9,6 @@ import {
   BeaconState,
   Checkpoint,
   ENRForkID,
-  Eth1Data,
   ForkDigest,
   SignedBeaconBlock,
   Uint16,
@@ -25,7 +24,7 @@ import {EMPTY_SIGNATURE, GENESIS_SLOT} from "../constants";
 import {IBeaconDb} from "../db";
 import {IEth1Notifier} from "../eth1";
 import {IBeaconMetrics} from "../metrics";
-import {getEmptyBlock, initializeBeaconStateFromEth1, isValidGenesisState, GenesisBuilder} from "./genesis/genesis";
+import {GenesisBuilder} from "./genesis/genesis";
 import {ILMDGHOST, ArrayDagLMDGHOST} from "./forkChoice";
 
 import {ChainEventEmitter, IAttestationProcessor, IBeaconChain} from "./interface";
@@ -35,6 +34,7 @@ import {IBeaconClock} from "./clock/interface";
 import {LocalClock} from "./clock/local/LocalClock";
 import {BlockProcessor} from "./blocks";
 import {sortBlocks} from "../sync/utils";
+import {getEmptyBlock} from "./genesis/util";
 
 export interface IBeaconChainModules {
   config: IBeaconConfig;
@@ -130,7 +130,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
     this._currentForkDigest =  computeForkDigest(this.config, state.fork.currentVersion, state.genesisValidatorsRoot);
     this.on("forkDigestChanged", this.handleForkDigestChanged);
     await this.restoreHeadState(state);
-    await this.eth1.startProcessEth1Blocks();
+    await this.eth1.start();
   }
 
   public async stop(): Promise<void> {
@@ -168,7 +168,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
     genesisBlock.stateRoot = stateRoot;
     const blockRoot = this.config.types.BeaconBlock.hashTreeRoot(genesisBlock);
     this.logger.info(`Initializing beacon chain with state root ${toHexString(stateRoot)}`
-            + ` and genesis block root ${toHexString(blockRoot)}`
+            + ` and block root ${toHexString(blockRoot)}, number of validator: ${genesisState.validators.length}`
     );
     const justifiedFinalizedCheckpoint = {
       root: blockRoot,
@@ -313,7 +313,6 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
       const builder = new GenesisBuilder(this.config, {eth1: this.eth1, db: this.db, logger: this.logger});
       state = await builder.waitForGenesis();
       await this.initializeBeaconChain(state);
-      this.logger.info(`Genesis state is ready with ${state.validators.length} validators`);
     }
     // set metrics based on beacon state
     this.metrics.currentSlot.set(state.slot);
@@ -322,41 +321,5 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
     this.metrics.currentFinalizedEpoch.set(state.finalizedCheckpoint.epoch);
     return state;
   }
-
-  /**
-   * Create a candidate BeaconState from the deposits at a certain time and eth1 state
-   *
-   * Returns the BeaconState if it is valid else null
-   */
-  private checkGenesis = async (timestamp: number, eth1Data: Eth1Data): Promise<TreeBacked<BeaconState> | null> => {
-    const blockHashHex = toHexString(eth1Data.blockHash);
-    this.logger.info(`Checking if block ${blockHashHex} will form valid genesis state`);
-    const depositDatas = await this.db.depositData.values({lt: eth1Data.depositCount});
-    const depositDataRoots = await this.db.depositDataRoot.values({lt: eth1Data.depositCount});
-    this.logger.info(`Found ${depositDatas.length} deposits`);
-    const depositDataRootList = this.config.types.DepositDataRootList.tree.defaultValue();
-    const tree = depositDataRootList.tree();
-
-    const genesisState = initializeBeaconStateFromEth1(
-      this.config,
-      eth1Data.blockHash,
-      timestamp,
-      depositDatas.map((data, index) => {
-        depositDataRootList.push(depositDataRoots[index]);
-        return {
-          proof: tree.getSingleProof(depositDataRootList.gindexOfProperty(index)),
-          data,
-        };
-      })
-    );
-    if (!isValidGenesisState(this.config, genesisState)) {
-      this.logger.info(`Eth1 block ${blockHashHex} is NOT forming valid genesis state`);
-      return null;
-    }
-    this.logger.info(`Initializing beacon chain with eth1 block ${blockHashHex}`);
-    await this.initializeBeaconChain(genesisState as TreeBacked<BeaconState>);
-    this.logger.info(`Genesis state is ready with ${genesisState.validators.length} validators`);
-    return genesisState;
-  };
 
 }
--- a/packages/lodestar/src/chain/genesis/genesis.ts
+++ b/packages/lodestar/src/chain/genesis/genesis.ts
@@ -4,37 +4,29 @@
 
 import {TreeBacked, List, fromHexString} from "@chainsafe/ssz";
 import {
-  BeaconBlock,
-  BeaconBlockBody,
-  BeaconBlockHeader,
   BeaconState,
   Deposit,
-  Eth1Data,
   Number64,
   Bytes32,
-  Fork,
-  SignedBeaconBlock,
   Root,
 } from "@chainsafe/lodestar-types";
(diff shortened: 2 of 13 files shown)


packages/lodestar/src/chain/chain.ts
```diff
@@ -309,16 +310,10 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
-            this.eth1.removeListener("eth1Data", genesisListener);
-            resolve(state);
-          }
-        };
-        this.eth1.on("eth1Data", genesisListener);
-      });
+      const builder = new GenesisBuilder(this.config, {eth1: this.eth1, db: this.db, logger: this.logger});
+      state = await builder.waitForGenesis();
+      await this.initializeBeaconChain(state);
+      this.logger.info(`Genesis state is ready with ${state.validators.length} validators`);
```
mpetrunic: I don't think we need that message, you could put number of validators inside `initializeBeaconChain` log.
It might be useful to have log messages once we satisfy min number of validators and when delay timer has started

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -28,16 +29,104 @@ import {
...
+    this.eth1 = eth1;
+    this.db = db;
+    this.logger = logger;
+    this.deposits = [];
+  }
+
+  /**
+   * Get eth1 deposit events and blocks and apply to this.state until we found genesis.
+   */
+  public waitForGenesis = async (): Promise<TreeBacked<BeaconState>> => {
```
mpetrunic: any reason why this is property and not a function?

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -28,16 +29,104 @@ import {
...
+    await this.initialize();
+    const eth1DataStream = await this.eth1.startProcessEth1Blocks(true);
+    return await pipe(eth1DataStream,
+      async (source: AsyncIterable<[IDepositEvent[], ethers.providers.Block]>) => {
```
mpetrunic: can we move it to separate function `processDepositEvents` so we can test that independently and it would help with readability

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -28,16 +29,104 @@ import {
...
+        for await (const data of source) {
+          const depositEvents = data[0];
+          const deposits = depositEvents.map((depositEvent) => {
+            this.depositTree.push(this.config.types.DepositData.hashTreeRoot(depositEvent));
+            return {
+              proof: this.depositTree.tree().getSingleProof(this.depositTree.gindexOfProperty(depositEvent.index)),
+              data: depositEvent,
+            };
+          });
+          this.deposits.push(...deposits);
```
mpetrunic: if you replace map with foreach we can do this in single loop

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -28,16 +29,104 @@ import {
...
+          applyDeposits(this.config, this.state, this.deposits, this.depositTree);
```
mpetrunic: I would extract this and lines below into `assembleGenesis`

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -28,16 +29,104 @@ import {
...
+          const block = data[1];
+          if (!block) {
+            continue;
+          }
+          applyTimestamp(this.config, this.state, block.timestamp);
+          applyEth1BlockHash(this.config, this.state, fromHexString(block.hash));
+          const isValid = isValidGenesisState(this.config, this.state);
+          if (isValid) {
+            this.logger.info(`Found genesis state at eth1 block ${block.number}`);
+            this.eth1.unsubscribeEth1Blocks();
```
mpetrunic: I think we should do this after pipe so we reduce complexity of this pipe

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -28,16 +29,104 @@ import {
...
+import {IGenesisBuilder, IGenesisBuilderModules} from "./interface";
 
+export class GenesisBuilder implements IGenesisBuilder {
+  private readonly config: IBeaconConfig;
+  private readonly db: IBeaconDb;
+  private readonly eth1: IEth1Notifier;
+  private readonly logger: ILogger;
+  private state: TreeBacked<BeaconState>;
+  private depositTree: TreeBacked<List<Root>>;
+  private deposits: Deposit[];
```
mpetrunic: do we actually need this, isn't this in state after you applyDeposits?

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -28,16 +29,104 @@ import {
...
+    this.deposits = [];
+  }
+
+  /**
+   * Get eth1 deposit events and blocks and apply to this.state until we found genesis.
+   */
+  public waitForGenesis = async (): Promise<TreeBacked<BeaconState>> => {
+    await this.initialize();
+    const eth1DataStream = await this.eth1.startProcessEth1Blocks(true);
+    return await pipe(eth1DataStream,
```
mpetrunic: I'm think to break this down into following pseudocode:
```
const state = await pipe(
   eth1DataStream,
   processDepositEvents, //maps into DepositData and fills up deposit tree and returns deposits array and block
   assembleGenesisState // apply given deposits and block onto state, return state if valid
)
//eth1 unsubscribe
```

packages/lodestar/src/chain/genesis/genesis.ts
mpetrunic: also it's gonna return undefined, if eth1 notifier is disabled

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -49,14 +138,66 @@ export function initializeBeaconStateFromEth1(
...
+    eth1Timestamp - eth1Timestamp % config.params.MIN_GENESIS_DELAY + 2 * config.params.MIN_GENESIS_DELAY;
+}
+
+/**
+ * Apply deposits to state.
+ * @param config IBeaconConfig
+ * @param state BeaconState
+ * @param deposits full list of deposits
+ */
+function applyDeposits(
```
mpetrunic: could you move those util functions to separate file so we keep codeclimate happy

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -70,40 +70,63 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
       this.logger.verbose("Eth1 notifier is disabled" );
       return;
     }
-    if (this.started) {
-      this.logger.verbose("Eth1 notifier already started" );
+    await this.doStart();
```
mpetrunic: It's a bit hard to use eth1notifier now.
We could do following:
- remove starting from "node.ts" file as it makes no sense to be started there
- keep counter how much time it's started (every time some module calls start +1, and everytime module calls stop -1 and stop only we counter is gonna be zero after stop)

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -70,40 +70,63 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
-      `Started listening to eth1 provider ${this.opts.provider.url} on chain ${this.opts.provider.network}`
-    );
-    this.logger.verbose(
-      `Last processed block number: ${this.lastProcessedEth1BlockNumber}`
-    );
     const headBlockNumber = await this.provider.getBlockNumber();
     // process historical unprocessed blocks up to curent head
     // then start listening for incoming blocks
+    if (subscribe) {
+      this.eth1Source = pushable<[IDepositEvent[], ethers.providers.Block]>();
```
mpetrunic: It's gonna bi nicer to use if it's object {depositEvents: IDepositEvent[], block: Block}

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -70,40 +70,63 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
+    this.startedProcessEth1 = false;
+    this.logger.verbose("Eth1 notifier stopped");
+  }
+
+  /**
+   * This is triggered when building genesis or after chain gets started.
+   * @param subscribe true for genesis builder
+   * @returns a pushable for genesis builder, null otherwise.
+   */
+  public async startProcessEth1Blocks(subscribe = false): Promise<Pushable<[IDepositEvent[], ethers.providers.Block]>> {
```
mpetrunic: if we assume module will call start, we can drop "subscribe" variable and rename it to `getEth1BlockAndDepositEventsSource`

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -70,40 +70,63 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
+    if (this.startedProcessEth1) {
+      this.logger.info("Started processing eth1 blocks already");
```
mpetrunic: I think it should be fine if two modules get this source

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -70,40 +70,63 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
+      this.eth1Source = pushable<[IDepositEvent[], ethers.providers.Block]>();
+    }
     this.processBlocks(headBlockNumber - this.config.params.ETH1_FOLLOW_DISTANCE).then(() => {
-      if(this.started) {
+      if(this.startedProcessEth1) {
         this.provider.on("block", this.onNewEth1Block.bind(this));
       }
     });
+    this.startedProcessEth1 = true;
+    return this.eth1Source;
```
mpetrunic, line 96: Don't do that, two modules cannot process same iterator.
You could after you process eth1 block, emit `Eth1BlockAndDepositEvents` data and create new EventIterator everytime someone asks for this source

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -70,40 +70,63 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
-    );
     const headBlockNumber = await this.provider.getBlockNumber();
     // process historical unprocessed blocks up to curent head
     // then start listening for incoming blocks
+    if (subscribe) {
+      this.eth1Source = pushable<[IDepositEvent[], ethers.providers.Block]>();
+    }
     this.processBlocks(headBlockNumber - this.config.params.ETH1_FOLLOW_DISTANCE).then(() => {
-      if(this.started) {
+      if(this.startedProcessEth1) {
```
mpetrunic: you should probably run processBlocks on start

mpetrunic (commented):
I'm fine with flow, there just some nits to make it more readable and less complicated. Hopefully I managed to articulate my thoughts :smile:

packages/lodestar/src/eth1/impl/ethers.ts
wemeetagain: I tend to think we should keep it simple for now, only allow one consumer at a time.

packages/lodestar/src/eth1/impl/ethers.ts
wemeetagain: I think we can (eventually) decompose this `IEth1Notifier` into a few parts, that can be run on-demand, rather than started globally. If we want to reuse resources, we could keep the provider and keep that around (it could be started/stopped in node.ts). Decomposing the functionality will lead to increased maintainability (each thing can be in its own file) and flexibility in how we can use this:
```typescript
type BlockTag = string | number

/**
 * Thin wrapper around ethers provider, tailored for the deposit contract
 */
interface IEth1Provider {
  /**
   * initialize contract, error if it doesn't exist or is invalid
   */
  start(): Promise<void>;
  stop(): Promise<void>;

  getBlock(blockTag: BlockTag): Promise<Block>;
  getDepositEvents(fromBlockTag: BlockTag, toBlockTag?: BlockTag): Promise<IDepositEvent[]>;

  // event for new blocks
  on(event: 'block', block: Block): void;
  off(event: 'block', block: Block): void;

  // event for module stopping, allows for event consumers to gracefully shut down
  on(event: 'stop'): void;
  off(event: 'stop'): void;
}

interface IBatchDepositEvents {
  depositEvents: IDepositEvent[];
  blockNumber: number;
}

/**
 * Stream of batched deposit events
 */
function getHistoricalDepositEventsStream(provider: IEth1Provider, from: blockNumber, to: blockNumber): Promise<Pushable<IBatchDepositEvents>>;

/**
 * Abortable stream of batched deposit events
 * data starting at `from`, continues at ETH1_FOLLOW_DISTANCE until aborted or IEthProvider stops
 */
function getFollowDepositEventsStream(provider: IEth1Provider, from: blockNumber): Promise<Pushable<IBatchDepositEvents>>;
```

From there we've separated how the data is consumed, database storage concerns, whether a stream has a single or multiple consumers, etc. We should write up a design doc determining what we store, when we store, which modules manage storage, etc. Likely, we will have a strategy for genesis and a strategy for proposing.

Eth1Data processing requires a depositData tree, either fetched from the db module, or managed by the genesis builder.
```typescript
/**
 * Create an Eth1Data based on a block and a deposit data root tree
 */
function getEth1Data(block: Block, depositDataRootTree: TreeBacked<Root<DepositData>[]>): {timestamp: number, eth1Data: Eth1Data};
```

Not saying we should do all this in this PR, just brainstorming how to make this whole process simpler and more manageable. IMO building out good primitives like this will help everything else fall into place.

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -70,40 +70,63 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
+    if (this.startedProcessEth1) {
+      this.logger.info("Started processing eth1 blocks already");
       return;
     }
-    this.started = true;
-    if(!this.contract) {
-      await this.initContract();
+    if (!this.opts.enabled) {
+      if (subscribe) {
+        this.logger.info("Eth1 notifider is disabled but starting it to build genesis state");
```
wemeetagain: typo here and below

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -70,40 +70,63 @@ export class EthersEth1Notifier extends (EventEmitter as { new(): Eth1EventEmitt
...
+   */
+  public async startProcessEth1Blocks(subscribe = false): Promise<Pushable<[IDepositEvent[], ethers.providers.Block]>> {
+    if (this.startedProcessEth1) {
+      this.logger.info("Started processing eth1 blocks already");
       return;
     }
-    this.started = true;
-    if(!this.contract) {
-      await this.initContract();
+    if (!this.opts.enabled) {
```
wemeetagain, line 90: Right now it seems that this `eth1.enabled` is more like `proposing.enabled` rather than a hard on/off switch for eth1 processing as a whole. This is cool, just noting it. We might eventually want a toggle for both.

packages/lodestar/src/chain/genesis/genesis.ts
twoeths: I'll switched waitForGenesis to be a function. If eth1 is disabled, this function still try to find genesis state and stop eth1 after found it. Does that make sense?

packages/lodestar/src/eth1/impl/ethers.ts
twoeths: For now, we consume eth1 block + events either before genesis (for building genesis state), or after genesis (for storing proposing data), not both

packages/lodestar/src/eth1/impl/ethers.ts
twoeths: for now, we either call this from pre-genesis, or after genesis, not both.
And for pregenesis (GenesisBuilder) it'll consume the stream, for after genesis we don't need the stream. I can separate to 2 different functions to make it clear.

wemeetagain (approved):
LGTM

PR #1050, opened 2020-06-25
Remove event listener on sse stream closing

### LINKED ISSUE #1029: Handle stream api closing

mpetrunic:
- removes listener when sse request is closed

We will still see `response terminated with an error with headers already sent` since that is fastify warning if request closes before response but we cannot fix that.

resolves #1029


wemeetagain:
typescript error https://github.com/ChainSafe/lodestar/pull/1050/checks?check_run_id=806985906#step:8:313

wemeetagain:
file conflict, i think from the block api PR

wemeetagain (approved):

Issue #1045, opened 2020-06-25
Cannot read property 'fork' of null


wemeetagain:
Syncing witti, started a node, then stopped.
On re-start hit this error, which crashed and stopped the node:
```
...
2020-06-24 20:09:00  [CHAIN]            info: Epoch 29 is justified! 
2020-06-24 20:09:00  [CHAIN]            info: Epoch 28 is finalized! 
2020-06-24 20:09:00  [CHAIN]            info: Processed new chain head newChainHeadRoot=0x055ae68f7ff2da84ef8192fe2b587701fec57a3631878f6180df9fe55240d7a3, slot=962, epoch=30
2020-06-24 20:09:00  [CHAIN]            info: Processed new chain head newChainHeadRoot=0xcd1b6e12075a04b26ae1fa616642f5e61f23f079339f04d17ba894ac6eb2fdc6, slot=964, epoch=30
2020-06-24 20:09:01  [CHAIN]            info: Processed new chain head newChainHeadRoot=0x9aaa2fe92de92cc55dfca434e21e0fc35b0dfc9e8fbf80b0e833704ab0f6779a, slot=968, epoch=30
2020-06-24 20:09:01  [CHAIN]            info: Finish restoring chain head from 68 blocks 
2020-06-24 20:09:01  [CHAIN]            info: restoreHeadState - duration=7321ms
  discv5:service Starting discv5 service +0ms
TypeError: Cannot read property 'fork' of null
    at BeaconChain.getENRForkID (/home/cayman/Code/lodestar/packages/lodestar/lib/chain/chain.js:239:34)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async MetadataController.start (/home/cayman/Code/lodestar/packages/lodestar/lib/network/metadata/metadata.js:35:79)
    at async Libp2pNetwork.start (/home/cayman/Code/lodestar/packages/lodestar/lib/network/network.js:128:5)
    at async BeaconNode.start (/home/cayman/Code/lodestar/packages/lodestar/lib/node/nodejs.js:158:5)
    at async Object.run [as handler] (/home/cayman/Code/lodestar/packages/lodestar-cli/lib/lodecli/cmds/beacon/cmds/run/run.js:77:3)
```

twoeths:
I cannot reproduce this issue in my environment, need to try some more times...

wemeetagain:
[...]

PR #1054, opened 2020-06-26
Fix sorted blocks in restoreHeadState()

twoeths:
It's not a fix for #1045 because the printed values are the same, however it's worth to fix.


mpetrunic (approved):
I think blocks will already be sorted anyways since they are sorted in leveldb

twoeths:
curious how can we do it? I thought leveldb only sort based on keys but in this case its blockRoot ...

mpetrunic:
> curious how can we do it? I thought leveldb only sort based on keys but in this case its blockRoot ...

Oh yeah true, they aren't sorted then probably

wemeetagain (approved):

PR #1065, opened 2020-06-27
Add assertion messages

dapplion:
Add meaningful assert messages to all `assert()` statements.

Extends the assert util to include comparator methods (`equal`, `lte`, `lt`, `gte`, `gt`) so the error message contains the asserted values.

Assert errors currently printing 
```
AssertionError: Expect false == true
```

will print
```
AssertionError: Current slot less than this target epoch's start slot: 351 >= 354
```

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/block/operations/attesterSlashing.ts
+++ b/packages/lodestar-beacon-state-transition/src/block/operations/attesterSlashing.ts
@@ -45,5 +45,5 @@ export function processAttesterSlashing(
     }
   });
 
-  assert.true(slashedAny, "should slash any");
+  assert.true(slashedAny, "no slashable validators for attester slashing found");
 }
--- a/packages/lodestar-beacon-state-transition/src/block/operations/attesterSlashing.ts
+++ b/packages/lodestar-beacon-state-transition/src/block/operations/attesterSlashing.ts
@@ -27,7 +27,7 @@ export function processAttesterSlashing(
   verifySignatures = true,
 ): void {
   // Check that the attestations are conflicting
-  assert.true(isValidAttesterSlashing(config, state, attesterSlashing, verifySignatures), "invalid attester slashing");
+  assert.true(isValidAttesterSlashing(config, state, attesterSlashing, verifySignatures), "Invalid attester slashing");
 
   const attestation1 = attesterSlashing.attestation1;
   const attestation2 = attesterSlashing.attestation2;
@@ -45,5 +45,5 @@ export function processAttesterSlashing(
     }
   });
 
-  assert.true(slashedAny, "no slashable validators for attester slashing found");
+  assert.true(slashedAny, "No slashable validators for attester slashing found");
 }
--- a/packages/lodestar-beacon-state-transition/src/block/operations/attesterSlashing.ts
+++ b/packages/lodestar-beacon-state-transition/src/block/operations/attesterSlashing.ts
@@ -27,7 +27,7 @@ export function processAttesterSlashing(
   verifySignatures = true,
 ): void {
   // Check that the attestations are conflicting
-  assert(isValidAttesterSlashing(config, state, attesterSlashing, verifySignatures));
+  assert.true(isValidAttesterSlashing(config, state, attesterSlashing, verifySignatures), "invalid attester slashing");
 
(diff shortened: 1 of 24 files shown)


packages/lodestar-beacon-state-transition/src/block/operations/attesterSlashing.ts
```diff
@@ -45,5 +45,5 @@ export function processAttesterSlashing(
     }
   });
 
-  assert(slashedAny);
+  assert.true(slashedAny, "should slash any");
```
dapplion: What would be an appropriate assertion message here? I don't believe this one is clear to end users

packages/lodestar-beacon-state-transition/src/util/blockRoot.ts
```diff
@@ -25,8 +25,8 @@ import {computeStartSlotAtEpoch} from "./epoch";
  * Return the block root at a recent [[slot]].
  */
 export function getBlockRootAtSlot(config: IBeaconConfig, state: BeaconState, slot: Slot): Root {
-  assert(slot < state.slot);
-  assert(state.slot <= slot + config.params.SLOTS_PER_HISTORICAL_ROOT);
+  assert.lt(slot, state.slot, "cannot get future epochs");
+  assert.lte(state.slot, slot + config.params.SLOTS_PER_HISTORICAL_ROOT);
```
dapplion: Could you recommend an assert message here?

packages/lodestar-beacon-state-transition/src/block/operations/attesterSlashing.ts
wemeetagain: maybe "no slashable validators for attester slashing found"

packages/lodestar-beacon-state-transition/src/util/blockRoot.ts
wemeetagain: maybe "cannot get block root from slot more than SLOTS_PER_HISTORICAL_ROOT in the past"

packages/lodestar-beacon-state-transition/src/util/blockRoot.ts
```diff
@@ -25,8 +25,8 @@ import {computeStartSlotAtEpoch} from "./epoch";
  * Return the block root at a recent [[slot]].
  */
 export function getBlockRootAtSlot(config: IBeaconConfig, state: BeaconState, slot: Slot): Root {
-  assert(slot < state.slot);
-  assert(state.slot <= slot + config.params.SLOTS_PER_HISTORICAL_ROOT);
+  assert.lt(slot, state.slot, "cannot get future epochs");
```
wemeetagain: maybe "cannot get block root for slot in the future"

packages/lodestar-beacon-state-transition/src/util/duties.ts
```diff
@@ -26,7 +26,7 @@ export function getCommitteeAssignment(
 ): CommitteeAssignment {
 
   const next2Epoch = getCurrentEpoch(config, state) + 2;
-  assert(epoch <= next2Epoch);
+  assert.lte(epoch, next2Epoch, "too future epoch");
```
wemeetagain: maybe "cannot get committee assignment for epoch more than two ahead"

packages/lodestar/src/chain/forkChoice/arrayDag/lmdGhost.ts
```diff
@@ -341,7 +342,7 @@ export class ArrayDagLMDGHOST implements ILMDGHOST {
   }
 
   public headStateRoot(): Uint8Array {
-    assert(Boolean(this.justified));
+    assert.true(Boolean(this.justified), "Justified checkpoint does not exist");
```
wemeetagain, line 345: Maybe we should standardize between capitalized and non-capitalized errors. No preference for either, just consistency

wemeetagain (changes requested):
Looks great, the assert upgrade is really nice

wemeetagain:
linter error: https://github.com/ChainSafe/lodestar/pull/1065/checks?check_run_id=814292314#step:7:84
otherwise looks good

mpetrunic (approved):

wemeetagain (approved):

PR #1066, opened 2020-06-27
Remove @ts-ignore usage

dapplion:
Remove `@ts-ignore` usage that do not require big changes. I would love for @wemeetagain or experienced devs in this codebase to check if there's a reasoning for some of the @ts-ignore occurrences that I might be missing

### DIFF
--- a/packages/lodestar/src/metrics/server/http.ts
+++ b/packages/lodestar/src/metrics/server/http.ts
@@ -24,8 +24,13 @@ export class HttpMetricsServer implements IMetricsServer {
   public async start(): Promise<void> {
     if (this.opts.enabled) {
       this.logger.info(`Starting metrics HTTP server on port ${this.opts.serverPort}`);
-      // @ts-ignore
-      await promisify<void, number>(this.http.listen.bind(this.http))(this.opts.serverPort);
+      const listen = this.http.listen.bind(this.http);
+      const port = this.opts.serverPort;
+      return new Promise((resolve, reject) => {
+        listen(port)
+          .once("listening", resolve)
+          .once("error", reject);
+      });
     }
   }
   public async stop(): Promise<void> {
--- a/packages/benchmark-utils/src/index.ts
+++ b/packages/benchmark-utils/src/index.ts
@@ -1,7 +1,6 @@
 import fs, {writeFile} from "fs";
 import {BENCH_DIR} from "./constant";
 import Benchmark from "benchmark";
-//@ts-ignore
 import profiler from "v8-profiler-next";
 import {dirname} from "path";
 
--- a/packages/lodestar-spec-test-util/src/single.ts
+++ b/packages/lodestar-spec-test-util/src/single.ts
@@ -67,7 +67,6 @@ export function describeDirectorySpecTest<TestCase, Result>(
   testFunction: (testCase: TestCase, directoryName: string) => Result,
   options: Partial<ISpecTestOptions<TestCase, Result>>
 ): void {
-  // @ts-ignore
   options = {...defaultOptions, ...options};
   if(!isDirectory(testCaseDirectoryPath)) {
     throw new Error(`${testCaseDirectoryPath} is not directory`);
--- a/packages/lodestar-utils/src/logger/interface.ts
+++ b/packages/lodestar-utils/src/logger/interface.ts
@@ -14,8 +14,7 @@ export enum LogLevel {
   silly = "silly",
 }
 
-// @ts-ignore
-export const LogLevels = Object.keys(LogLevel).map(key => LogLevel[key]);
+export const LogLevels = Object.values(LogLevel);
 
 
 export const customColors = {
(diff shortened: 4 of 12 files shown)


packages/lodestar/src/metrics/server/http.ts
```diff
@@ -24,8 +24,13 @@ export class HttpMetricsServer implements IMetricsServer {
   public async start(): Promise<void> {
     if (this.opts.enabled) {
       this.logger.info(`Starting metrics HTTP server on port ${this.opts.serverPort}`);
-      // @ts-ignore
-      await promisify<void, number>(this.http.listen.bind(this.http))(this.opts.serverPort);
```
mpetrunic, line 28: Can you remember why ts-ignore was needed with promisify impl?

Also it would probably be nice to catch error here if port is taken^^

mpetrunic (commented):
This is awesome.

I think ts-ignores are mostly in tests or in places where we had bad types but were too lazy to fix them (especially libp2p parts)

packages/lodestar/src/metrics/server/http.ts
dapplion: Because Typescript is not able to figure out the complex overloads of the listen function and results in a bad type. So `listen(3000)` results in a type error.

mpetrunic: Oh yeah, that was so annoying, it was same thing in rpc server (when we had it).

dapplion: I'm still confused in what's the appropriate way of doing a `listen` handler. Does the callback return an error? According to the official types it doesn't.

mpetrunic: Not super familiar, but it could be that error is first param in callback

dapplion:
@mpetrunic Can you take a look at these type errors? I'm not familiar with pouch DB
- https://github.com/ChainSafe/lodestar/pull/1066/checks#step:5:381
- https://github.com/ChainSafe/lodestar/pull/1066/checks#step:5:384

mpetrunic:
> @mpetrunic Can you take a look at these type errors? I'm not familiar with pouch DB
> 
> * https://github.com/ChainSafe/lodestar/pull/1066/checks#step:5:381
> * https://github.com/ChainSafe/lodestar/pull/1066/checks#step:5:384

I can check on monday. Maybe pouchdb types aren't match for pouchdb version we are using

dapplion:
> I can check on monday. Maybe pouchdb types aren't match for pouchdb version we are using

We are using 
```
"pouchdb-adapter-memory": "^7.0.0",
"pouchdb-core": "^7.0.0",
```
And types
```
"@types/pouchdb-adapter-memory": "^6.1.3",
"@types/pouchdb-core": "^7.0.6",
```
`@types/pouchdb-adapter-memory` is at its latest version so maybe it has not been updated for the latest major version. However the conflicting types come from `@types/pouchdb-core`

packages/lodestar/src/metrics/server/http.ts
dapplion: Seems that the callback is subscribed to the `'listening'` event https://github.com/nodejs/node/blob/f63436d190b60e12131036aa9d1888d9023e9127/lib/net.js#L1378, which is emitted without any arguments.

wemeetagain:
> I think ts-ignores are mostly in tests or in places where we had bad types but were too lazy to fix them (especially libp2p parts)

Right, that's my understanding. Vast majority of ts-ignores should/can be removed with a bit of work.
I can only think of one reason for ts-ignore, which you've already encountered, dynamic events which can't be typed.

RE: pouchdb, seems like the declared types are just wrong. I'd be interested in revisiting if we even need the extra layer of DB abstraction, perhaps purging pouchdb and using level exclusively and concretely.

dapplion:
@wemeetagain  So for this PR would you recommend reverting the pouch DB commits and leave that part untouched?

wemeetagain:
Yeah, I think we can visit that separately. Will make an issue now

dapplion:
Besides for dynamic events, the other area where `@ts-ignore` is used and may be concerning is here
https://github.com/ChainSafe/lodestar/blob/7f69a144dcdc3b4c00f25d6c87ff8ae3c5aee24a/packages/lodestar-types/src/ssz/generators/index.ts#L30-L39

Do you think it's worth it to try to fix it?

wemeetagain:
Ah, not opposed to trying, that section of code is likely a little confusing to outside observers. The ts-ignore doesn't help.

Here, we're creating a big object of SSZ type objects, `IBeaconSSZTypes`, but there is a tree of heirarchy, where some type objects rely on other objects existing.  The reason we're ts-ignoring here is that we're using a list of object names, `typeNames: string[]`, which is sorted in this 'dependency order'. And `IBeaconSSZTypes` is not able to be dynamically indexed by strings. Perhaps if we say types[type as keyof IBeaconSSZTypes]` or declare `typeNames` as `keyof IBeaconSSZTypes`. Or there might be some other way you know of.

dapplion:
@wemeetagain Everything seems okay to merge, it's the labeler that failed because the source branch comes from a fork

mpetrunic (approved):

wemeetagain (approved):

Issue #1070, opened 2020-06-28
Get finalized block by root


mpetrunic:
**Is your feature request related to a problem? Please describe.**

When fetching block by block id ("head", slot, root, "genesis" etc), we should resolve block id to root.
When we do that, we cannot fetch finalized blocks since they are indexed by slot.

**Describe the solution you'd like**

Store root => slot references for finalized blocks.

**Describe alternatives you've considered**

Make function to resolve block id into root if not finalized and slot if finalized. I Think this alternative will make code quicke complex and we would propagate this complexity trough a lot of code.

**Additional context**

wemeetagain:
:+1:
With this and #1069 we will be able to retrieve finalized blocks by slot, root, or parentRoot.
We likely should expose helpers on our forkchoice/chain to retrieve non-finalized blocks by slot, root, or parentRoot, to mirror.

Issue #1069, opened 2020-06-28
Searching blocks by parentRoot property


mpetrunic:
**Is your feature request related to a problem? Please describe.**

In api, we need to be able to filter blocks by parentRoot.
Ref: https://ethereum.github.io/eth2.0-APIs/#/Beacon/getBlockHeaders

**Describe the solution you'd like**

Finalized blhttps://ethereum.github.io/eth2.0-APIs/#/Beacon/getBlockHeadersocks (block archive):
- store references parentRoot => slot (since finalized blocks are indexed by slot)

NonFinalized blocks:
- quick solution would be to just iterate all blocks and search mannually (they are already in memory)
   - could be problematic in long periods of non finalization

**Describe alternatives you've considered**

**Additional context**

wemeetagain:
Both of those solutions seem good.

Definitely wouldn't bother with anything fancier for non-finalized blocks until/unless we do something more advanced for the cases of non-finalization. Likely worth tackling as a separate issue once it arises.

PR #1071, opened 2020-06-28
Block endpoints in rest api

### LINKED ISSUE #1056: Add block rest endpoints
   - add beacon/block endpoints
   - add beacon/headers endpoints
- slight api refactor - separate routes from controllers (for easier navigation)

mpetrunic:
resolves #1056 

There are some things lacking like:
- fetching finalized blocks - depends on #1070 
- searching blocks by parent root - depends on #1069 

I wish we migrate all endpoints to look like this
- controller defines route (easier to change)
- controller files are named after operationId in spec

Side note: Aparently we will return response much faster if we define fastify response schema, it's not really important for now but later on, we could generate that schemas from ssz

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,4 +1,5 @@
-version: "2"         # required to adjust maintainability checks
+version: "2"  # required to adjust maintainability checks
+
 checks:
   argument-count:
     config:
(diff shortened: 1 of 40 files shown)


wemeetagain:
> I wish we migrate all endpoints to look like this
> - controller defines route (easier to change)
> - controller files are named after operationId in spec

:+1:

wemeetagain (approved):

PR #1074, opened 2020-06-29
0.12.x merge master

wemeetagain:
Mostly straightforward

One issue was with `ReadonlyEpochContext` which had to removed from chain, bc we rely on the full interface in some places.

Unit tests and e2e tests run, but there are a few failures


wemeetagain:
it seems the commit history got lost here -_- closing this

PR #1075, opened 2020-06-30
Fix metadata e2e test

twoeths:


packages/lodestar/src/network/reqResp.ts
```diff
@@ -206,7 +206,6 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
           }
           return;
         }
-        yield* getResponse(peerId, method, encoding);
```
twoeths, line 209: we never reach this statement because we always yield something in `eth2RequestDecode`

mpetrunic (approved):
lgtm, there are some wierd e2e tests. I'll try to reproduce locally

wemeetagain (approved):

PR #1077, opened 2020-06-30
Fix e2e tests

mpetrunic:
- fix validators skipping first slot in epoch
- fix archiving  task not storing finalized block so it's not returned in block by range
- add removal of past proposer duties
- fix block proposing modifying context

There are other bugs identified:
- something is wrong in block gossip validation, a lot of blocks are failing with block proposer index out of range (could be bad context or state)
- when checking if synced proper chain we would compare peer finalized root with processed checkpoint root but they are returning blocks for diffferent slot (processed checkpoint has root for last slot in epoch while finalized has first slot in epoch)

### DIFF
--- a/packages/lodestar/src/tasks/tasks/archiveBlocks.ts
+++ b/packages/lodestar/src/tasks/tasks/archiveBlocks.ts
@@ -5,9 +5,8 @@
 import {ITask} from "../interface";
 import {IBeaconDb} from "../../db/api";
 import {Checkpoint, SignedBeaconBlock} from "@chainsafe/lodestar-types";
-import {computeEpochAtSlot} from "@chainsafe/lodestar-beacon-state-transition";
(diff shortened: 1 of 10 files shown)


wemeetagain (approved):

packages/lodestar/src/tasks/tasks/archiveBlocks.ts
```diff
@@ -44,12 +43,12 @@ export class ArchiveBlocksTask implements ITask {
     });
     const blocks = allBlocks.filter(
       (block) =>
-        computeEpochAtSlot(this.config, block.message.slot) < this.finalizedCheckpoint.epoch
+        block.message.slot <= finalizedBlock.message.slot
```
twoeths, line 46: I created #1078 to follow up this change.

PR #1080, opened 2020-07-01
Libp2p 0.28.3 migration

### LINKED ISSUE #1036: Libp2p migration to 0.28
Refer to
+ https://github.com/libp2p/js-libp2p/issues/603
+ https://github.com/libp2p/js-libp2p/blob/master/doc/migrations/v0.27-v.28.md

twoeths:
resolves #1036 

+ Instead of peerInfo, we use peerId everywhere
+ Normally, nodes found by dscv5, node's multi addresses are added to `peerStore.addressBook` by libp2p so we don't have to care about `multiaddrs`
+ For tests, we have to do additional `network.connect(peerId, multiaddrs)` in order to let our node know the other node's `multiaddrs`
## TODO:
+ Need to ultilize peerStore metadata book instead of our reputation store. => should be done in #1081


packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -1,4 +1,7 @@
-import Gossipsub, {IGossipMessage, Options, Registrar} from "libp2p-gossipsub";
+import PeerId from "peer-id";
+import Gossipsub from "libp2p-gossipsub";
+import {Registrar} from "libp2p-gossipsub/src/peer";
+import {Message} from "libp2p-gossipsub/src/message";
```
twoeths, line 4: is it ok to import like this or gossipsub has to publish `Message` directly?

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -71,28 +72,11 @@ export function getCommonFinalizedCheckpoint(config: IBeaconConfig, peers: IRepu
   }
 }
 
-
-
-export function targetSlotToBlockChunks(
```
twoeths, line 76: Don't see this method being used so I removed all references to this

packages/lodestar/src/network/network.ts
```diff
@@ -56,7 +58,7 @@ export class Libp2pNetwork extends (EventEmitter as { new(): NetworkEventEmitter
     // `libp2p` can be a promise as well as a libp2p object
     this.inited = new Promise((resolve) => {
```
twoeths: I don't see in which case libp2p is a promise, probably this happens in old version of libp2p only?

packages/lodestar/package.json
```diff
@@ -67,18 +67,17 @@
...
-    "libp2p-bootstrap": "0.10.4",
-    "libp2p-gossipsub": "^0.3.0",
-    "libp2p-mdns": "^0.13.3",
+    "libp2p": "^0.28.3",
+    "libp2p-bootstrap": "^0.11.0",
+    "libp2p-gossipsub": "^0.4.5",
+    "libp2p-mdns": "^0.14.1",
     "libp2p-mplex": "^0.9.5",
-    "libp2p-noise": "^1.1.0",
+    "libp2p-noise": "^1.1.1",
```
mpetrunic: should be 1.1.2 to insure we have all bugfixes

packages/lodestar/src/network/network.ts
wemeetagain: Yeah, we used to maybe initialize a libp2p object in BeaconNode constructor. Since constructor is synchronous function, we couldn't await there, but would await here. Now, we properly create a libp2p object before creating a new BeaconNode, so we probably shouldn't support this anymore.

packages/lodestar/package.json
```diff
@@ -109,7 +108,7 @@
     "chai": "^4.2.0",
     "chai-as-promised": "^7.1.1",
     "eventsource": "^1.0.7",
-    "libp2p-ts": "https://github.com/ChainSafe/libp2p-ts.git",
+    "libp2p-ts": "https://github.com/ChainSafe/libp2p-ts.git#tuyen/libp2p-0.28-types",
```
wemeetagain: merged changes in branch to master

packages/lodestar/src/network/gossip/gossipsub.ts
wemeetagain: Should be ok to import like that. We can make sure to export `Message` to root for the 0.29 migration

packages/lodestar/src/network/nodejs/bundle.ts
```diff
@@ -8,12 +8,13 @@ import Mplex from "libp2p-mplex";
...
 import MDNS from "libp2p-mdns";
-import PeerInfo from "peer-info";
+import PeerId from "peer-id";
 import {ENRInput, Discv5Discovery} from "@chainsafe/discv5";
 
 
 export interface ILibp2pOptions {
-  peerInfo: PeerInfo;
+  peerId: PeerId;
+  listenAddrs: string[];
```
wemeetagain: Might be a good time to support all three, {listen, announce, noAnnounce}, even if its not yet supported downstream in the cli. Eg: for now we can do: `NodejsNode({addresses: {listen: multiaddrs}})`, later adding options to configure announce and noAnnounce.

wemeetagain (changes requested):

wemeetagain (approved):

PR #1082, opened 2020-07-01
Dont use epochContext inside gossip validator

twoeths:
Sometimes I got this error in e2e
```
error: Cannot validate message from 16Uiu2HAmRN1W6A8o1Yb7DdFA8UneV7XXy8pd5GiDjweEsvexY8Pt, topic /eth2/31681ad1/beacon_block/ssz_snappy, error: Error: beacon proposer index out of range
```
This is because `EpochContext.getBeaconProposer()` is only designed to get proposers for slots of current epoch.

Similarly, we should not use `EpochContext.getBeaconCommittee()` because it only caches committes for previous epoch, current epoch or next epoch. I sometimes see the error `crosslink committee retrieval: out of range epoch` in my e2e too

### DIFF
--- a/packages/lodestar/src/network/gossip/validator.ts
+++ b/packages/lodestar/src/network/gossip/validator.ts
@@ -24,7 +24,8 @@ import {
   getDomain,
   computeEpochAtSlot,
   computeSigningRoot,
-  computeSubnetForAttestation
+  computeSubnetForAttestation,
+  getBeaconProposerIndex
 } from "@chainsafe/lodestar-beacon-state-transition";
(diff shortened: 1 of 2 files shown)


mpetrunic (approved):

wemeetagain:
Lets wait until #1084 so we can select the epochCtx from the right epoch.

mpetrunic:
> Lets wait until #1084 so we can select the epochCtx from the right epoch.

I think #1084 already fixes this

wemeetagain:
closing in favor of #1084 
We can still use epochContext inside gossip validators, now that we can track the _correct_ epochContext to use.

PR #1083, opened 2020-07-01
Dont use EpochContext to calculate state root

twoeths:
In my environment, this error happens once per every 10 times when we produce a block: `Cannot get block root for slot in the future`

This is inside `packages/lodestar/src/chain/factory/block/index.ts` where we calculate state root, and `computeNewStateRoot` really depends on the correctness of EpochContext which changes very quickly, this is an uncertainty.

As we're not sure the correctness of EpochContext here, I propose we switch back to the legacy way of calculating state root. This doesnt happen very frequently at runtime so I think we don't have to worry about the performance here.

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -35,6 +35,7 @@ import {LocalClock} from "./clock/local/LocalClock";
 import {BlockProcessor} from "./blocks";
 import {sortBlocks} from "../sync/utils";
 import {getEmptyBlock} from "./genesis/util";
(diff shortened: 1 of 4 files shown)


mpetrunic (changes requested):
It's actually because of fork. See my comment in #lodestar channel

twoeths:
yeah I saw it. In my case, it happens when producing block/calculating root, not when running state transition. I'll try to catch the log and post here.

mpetrunic:
Its because of same thing, you try to rotate epoch context in block production as well, your solution is just gonna make things slower but its gonna have same issue because fork block on epoch boundary corrupts our epoch context

twoeths:
So I think I'd just close this PR as it's not tackling the root cause.

PR #1084, opened 2020-07-02
Add epoch context to state cache

mpetrunic:
- add epoch ctx alongside state in cache
- fix gossip attestation validators running processSlots before checking slot range
- fix gossip block validator validating against head state instead of parentRoot block state
- small fixes here and there


mpetrunic:
~~Seems like we don't finalize :cry: ~~

packages/lodestar/src/db/api/beacon/stateContextCache.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (!item) {
+      return null;
+    }
+    return {
+      state: item.state.clone(),
+      epochCtx: item.epochCtx?.copy()
+    };
+  }
+
+  public async add(state: TreeBacked<BeaconState>, epochContext?: EpochContext): Promise<void> {
```
wemeetagain: use `ITreeStateContext` param here

packages/lodestar/src/chain/interface.ts
```diff
@@ -56,14 +57,12 @@ export interface IBeaconChain extends ChainEventEmitter {
    */
   getENRForkID(): Promise<ENRForkID>;
 
-  getHeadState(): Promise<BeaconState|null>;
+  getHeadContext(): Promise<IStateContext|null>;
```
wemeetagain: Can we call this `getHeadStateContext`? I think better to be a little explicit here since there's also the epochCtx sharing the name.
It might be nice to also still expose the `getHeadState` and `getHeadEpochContext`.

packages/lodestar-beacon-state-transition/src/fast/util/index.ts
```diff
@@ -1,7 +1,15 @@
+import {BeaconState} from "@chainsafe/lodestar-types";
+import {EpochContext} from "../index";
```
wemeetagain: import from './epochContext'

packages/lodestar-beacon-state-transition/src/fast/index.ts
```diff
@@ -11,16 +11,15 @@ export {
 };
```
wemeetagain, line 12: lets export IStateContext here too so we can import from root in lodestar

packages/lodestar/src/chain/interface.ts
```diff
@@ -56,14 +57,12 @@ export interface IBeaconChain extends ChainEventEmitter {
    */
   getENRForkID(): Promise<ENRForkID>;
 
-  getHeadState(): Promise<BeaconState|null>;
```
wemeetagain, line 59: I think we should create a:
```typescript
export interface ITreeStateContext {
  epochCtx: EpochContext;
}
```
and use TreeBacked beacon state here and across the `lodestar` package, rather than currently at the boundary of `StateContextCache`.

eg:
```typescript
interface IBeaconChain {
  ...
  getHeadStateContext(): Promise<ITreeStateContext | null>;
  getHeadState(): Promise<TreeBacked<BeaconState> | null>;
}
```

We can also use `ITreeStateContext` as the `StateContextCache` cache item.

packages/lodestar/src/db/api/beacon/stateContextCache.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Similar API to Repository, but synchronous
+ */
+export class StateContextCache {
+  private cache: Record<string, IStateContextCacheItem>;
+  constructor() {
+    this.cache = {};
+  }
+
+  public async get(root: ByteVector): Promise<IStateContextCacheItem | null> {
```
wemeetagain: return `Promise<ITreeStateContext | null>` here.

packages/lodestar/src/db/api/beacon/stateContextCache.ts
```diff
@@ -0,0 +1,54 @@
...
+import {BeaconState} from "@chainsafe/lodestar-types";
+import {EpochContext} from "@chainsafe/lodestar-beacon-state-transition";
+
+export interface IStateContextCacheItem {
+  state: TreeBacked<BeaconState>;
+  epochCtx?: EpochContext;
+}
+
+/**
+ * Simple BeaconState in-memory cache
```
wemeetagain: Update comment

packages/lodestar/src/db/api/beacon/stateContextCache.ts
```diff
@@ -0,0 +1,54 @@
...
+  public async batchDelete(roots: ByteVector[]): Promise<void> {
+    await Promise.all(roots.map((root) => this.delete(root)));
+  }
+
+  public clear(): void {
+    this.cache = {};
+  }
+
+  public async values(): Promise<IStateContextCacheItem[]> {
+    return Object.values(this.cache);
```
wemeetagain: Looks like an existing bug, lurking, but the values should be `.copy/.clone`ed.

packages/lodestar/src/db/api/beacon/stateContextCache.ts
```diff
@@ -0,0 +1,54 @@
...
+  state: TreeBacked<BeaconState>;
+  epochCtx?: EpochContext;
+}
+
+/**
+ * Simple BeaconState in-memory cache
+ *
+ * Similar API to Repository, but synchronous
+ */
+export class StateContextCache {
```
wemeetagain, line 15: Considering we're always passing passing in both a state and epochCtx, and always needing to get both a state and epochCtx, I don't think its worth having the separate interface with optional epochCtx here, just keep it simple for now.

packages/lodestar/src/chain/factory/block/index.ts
```diff
@@ -22,32 +22,37 @@ export async function assembleBlock(
   randaoReveal: Bytes96,
   graffiti = ZERO_HASH,
 ): Promise<BeaconBlock | null> {
-  const [parentBlock, currentState] = await Promise.all([
+  const [parentBlock, stateContext] = await Promise.all([
     chain.getHeadBlock(),
-    chain.getHeadState()
+    chain.getHeadContext()
```
wemeetagain: If we keep the `getHeadState`, we can avoid a lot of these changes

packages/lodestar/src/chain/blocks/process.ts
```diff
@@ -8,36 +8,39 @@ import {ILogger} from "@chainsafe/lodestar-utils/lib/logger";
...
   logger: ILogger,
   db: IBeaconDb,
-  epochCtx: EpochContext,
   forkChoice: ILMDGHOST,
   pool: BlockPool,
   eventBus: ChainEventEmitter,
 ): (source: AsyncIterable<IBlockProcessJob>) =>
-  AsyncGenerator<{preState: BeaconState; block: SignedBeaconBlock; postState: BeaconState; finalized: boolean}> {
+  AsyncGenerator<{
+    preState: IStateContext; block: SignedBeaconBlock; postState: Required<IStateContext>; finalized: boolean;
```
wemeetagain: can we update this to `preStateContext` and `postStateContext`?

packages/lodestar/src/chain/blocks/process.ts
```diff
@@ -8,36 +8,39 @@ import {ILogger} from "@chainsafe/lodestar-utils/lib/logger";
...
+  }> {
   return (source) => {
     return (async function*() {
       for await(const job of source) {
         const blockRoot = config.types.BeaconBlock.hashTreeRoot(job.signedBlock.message);
-        const preState = await getPreState(config, db, forkChoice, pool, logger, job);
-        if(!preState) {
+        const stateContext = await getPreState(config, db, forkChoice, pool, logger, job);
```
wemeetagain: might be nice to call this preStateContext, and below postStateContext

wemeetagain (changes requested):

packages/lodestar/src/chain/blocks/process.ts
```diff
@@ -75,7 +78,12 @@ export async function getPreState(
     pool.addPendingBlock(job);
     return null;
   }
-  return await db.stateCache.get(parentBlock.stateRoot as Uint8Array);
+  const context = await db.stateCache.get(parentBlock.stateRoot as Uint8Array);
+  if(!context.epochCtx) {
```
twoeths: I think we always have epochCtx in stateCache except for genesis state.
Even when we restart node, we also go thru this process and we have epochCtx.

We can either handle specifically for genesis state, or if we want to have fallback like this then probably it's better to put in `db.stateCache`

packages/lodestar/src/chain/factory/block/index.ts
```diff
@@ -22,32 +22,37 @@ export async function assembleBlock(
...
     proposerIndex,
     parentRoot: config.types.BeaconBlockHeader.hashTreeRoot(parentHeader),
     stateRoot: undefined,
-    body: await assembleBody(config, db, {...currentState, slot}, randaoReveal, graffiti),
+    body: await assembleBody(config, db, {...stateContext.state, slot}, randaoReveal, graffiti),
   };
 
-
-  block.stateRoot = computeNewStateRoot(config, currentState, block, chain);
+  let epochCtx: EpochContext;
```
twoeths, line 38: let's handle this logic in single place

packages/lodestar/src/chain/interface.ts
mpetrunic: Why does rest of lodestar care about underlying backing? Isn't it only important inside state cache?

packages/lodestar/src/chain/factory/block/index.ts
mpetrunic: TBH, I initially removed so I'm forced to fix bugs where we use head state incorectly (that's how I've discovered bugs in gossip validation). I will restore getHeadState method

packages/lodestar/src/chain/interface.ts
wemeetagain: serving proofs :)
Also, because we rely on the efficiency of a tree-backed state, and practically can't run a production beacon node without it. If we accept that at the `lodestar` level, then there's not really a benefit to obscuring the backing.

This is somewhat different than the at the level of the state transition, in `lodestar-beacon-state-transition`. There, the actual runtime is not known. We don't know where/what the state transition is being used for. We're just exporting a few functions. There, we do get a benefit from obscuring the backing, since it opens options to downstream consumers. (to use whichever backing suits their application)

mpetrunic: Can we do that in separate PR?

wemeetagain: Yeah, lets do that in a separate PR (once we merge 0.12.x back to master)

packages/lodestar/src/api/impl/beacon/beacon.ts
```diff
@@ -61,7 +60,7 @@ export class BeaconApi implements IBeaconApi {
   }
 
   public async getFork(): Promise<ForkResponse> {
-    const state: BeaconState = await this.chain.getHeadState();
+    const {state} = await this.chain.getHeadStateContext();
```
wemeetagain: getHeadState

packages/lodestar/src/api/impl/beacon/beacon.ts
```diff
@@ -76,7 +75,7 @@ export class BeaconApi implements IBeaconApi {
   }
 
   public async getGenesisTime(): Promise<Number64> {
-    const state = await this.chain.getHeadState();
+    const {state} = await this.chain.getHeadStateContext();
```
wemeetagain: getHeadState

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -68,7 +68,7 @@ export class ValidatorApi implements IValidatorApi {
   }
 
   public async produceBlock(slot: Slot, validatorPubkey: BLSPubkey, randaoReveal: Bytes96): Promise<BeaconBlock> {
-    const validatorIndex = this.chain.getEpochContext().pubkey2index.get(validatorPubkey);
+    const validatorIndex = (await this.chain.getHeadStateContext()).epochCtx.pubkey2index.get(validatorPubkey);
```
wemeetagain: getHeadEpochContext

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -191,19 +191,21 @@ export class ValidatorApi implements IValidatorApi {
...
-      aggregatorIndex: this.chain.getEpochContext().pubkey2index.get(aggregator),
+      aggregatorIndex: epochCtx.pubkey2index.get(aggregator),
       selectionProof: EMPTY_SIGNATURE
     };
   }
 
   public async subscribeCommitteeSubnet(
     slot: Slot, slotSignature: BLSSignature, committeeIndex: CommitteeIndex, aggregatorPubkey: BLSPubkey
   ): Promise<void> {
+    const {state} = await this.chain.getHeadStateContext();
```
wemeetagain: getHeadState

packages/lodestar/src/db/api/beacon/stateContextCache.ts
```diff
@@ -0,0 +1,55 @@
+import {ByteVector, toHexString, TreeBacked} from "@chainsafe/ssz";
+import {BeaconState} from "@chainsafe/lodestar-types";
+import {EpochContext} from "@chainsafe/lodestar-beacon-state-transition";
+
+export interface IStateContextCacheItem {
```
wemeetagain: Can we call this `ITreeStateContext` to make future PR easier?

packages/lodestar/src/network/gossip/gossip.ts
```diff
@@ -153,7 +153,7 @@ export class Gossip extends (EventEmitter as { new(): GossipEventEmitter }) impl
   }
 
   public async getForkDigestByEpoch(epoch: Epoch): Promise<ForkDigest> {
-    const state = await this.chain.getHeadState();
+    const {state} = await this.chain.getHeadStateContext();
```
wemeetagain: getHeadState

packages/lodestar/src/network/gossip/handlers/attestation.ts
```diff
@@ -29,7 +29,7 @@ export function getCommitteeAttestationHandler(subnet: number): GossipHandlerFn
 
 export async function publishCommiteeAttestation(this: Gossip, attestation: Attestation): Promise<void> {
   const forkDigestValue = await this.getForkDigest(attestation.data.slot);
-  const headState = await this.chain.getHeadState();
+  const {state: headState} = await this.chain.getHeadStateContext();
```
wemeetagain: getHeadState

packages/lodestar/test/e2e/sync/sync.test.ts
```diff
@@ -22,7 +22,7 @@ describe("syncing", function () {
     const bn2 = await getDevBeaconNode(
       {SECONDS_PER_SLOT: 2, SLOTS_PER_EPOCH: 8},
       8,
-      (await bn.chain.getHeadState()).genesisTime
+      (await bn.chain.getHeadStateContext()).state.genesisTime
```
wemeetagain: getHeadState

wemeetagain (changes requested):
looks good, minor comments

wemeetagain (approved):

Issue #1085, opened 2020-07-02
Validator CLI


wemeetagain:
**Is your feature request related to a problem? Please describe.**
We need a validator CLI to run validators from the command line.

**Describe the solution you'd like**
A validator CLI with:
- ability to run N validators at once
- the big task here will be handling on-disk persistence, managing keys, user secrets, etc and passing them successfully to the validator nodes
- safely handle user secrets
- we likely only need to handle 'keystores', rather than 'walletstores'
- consistent code style, building off the current `beacon` command. Eg: Create a `validator` command, in the same way
- implementation following community standards and community discussion
  - ideally following closely with other client decisions, picking the best ideas and using them

**Describe alternatives you've considered**
No CLI? :sob: 

**Additional context**
Related to the validator CLI we may need a wallet CLI command. Depending on judgement, that may be completed in parallel or as a prerequisite.

[EIP-2335](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2335.md)
https://github.com/ethereum/eth2.0-pm/issues/161
See other clients validator clients:
https://github.com/sigp/lighthouse/tree/master/validator_client
https://github.com/PegaSysEng/teku
https://github.com/prysmaticlabs/prysm/
https://github.com/status-im/nimbus
https://github.com/ethereum/trinity/tree/master/eth2

See Eth2 R&D #key-management discord for important discussion on validator key management

dapplion:
Do you imagine it being close to `ethdo` https://github.com/wealdtech/ethdo/blob/master/docs/usage.md ?

wemeetagain:
I think that would be more related to an `account`/`wallet` command, which would manage the on-disk keystores, in a lodestar-opinionated way.

The `validator` cli command would read on-disk keystores, as configured via cli and on-disk configuration, and use those to init `Validator`s from `lodestar-validator`.

wemeetagain:
resolved with #1093

PR #1093, opened 2020-07-03
Add basic validator client CLI

dapplion:
Adds a validator client CLI following Lighthouse style and keystores directory structure. Fixes https://github.com/ChainSafe/lodestar/issues/1085

To run:
```
lodecli validator
```

Command options:
```js
  datadir: {
    default: "validator",
    alias: ["datadir", "d"],
    description: "Data directory for keys and databases",
    type: "string"
  },

  server: {
    default: "http://127.0.0.1:9596",
    alias: ["server"],
    description: "Address to connect to BeaconNode",
    type: "string"
  },

  spec: {
    default: "mainnet",
    alias: ["spec", "s"],
    description: "Specifies the default eth2 spec type",
    choices: ["mainnet", "minimal", "interop"],
    type: "string"
  },

  validatorsDir: {
    default: "~/.lodestar/validators",
    alias: ["validators-dir"],
    description: "The directory for storing validator keystores",
    type: "string"
  },

  secretsDir: {
    default: "~/.lodestar/secrets",
    alias: ["secrets-dir"],
    description: "The directory for storing validator keystore secrets",
    type: "string"
  }
```

The expected keystore structure is
```
$ tree ~/.lodestar/
/home/lion/.lodestar/
└── validator
    ├── secrets
    │   └── 0xb108ea25ebb658adb2aadc9a3d6ff62b4d015ce66f8024eb5fc94c31c9a54b6c84905022f4c8cb376dae7f0ac5181412
    └── validators
        └── 6aaa5de9-964e-479e-8682-12a1f0c044e3
```
Where `0xb108ea25ebb658adb2aadc9a3d6ff62b4d015ce66f8024eb5fc94c31c9a54b6c84905022f4c8cb376dae7f0ac5181412` is the public key of the validator with UUID `6aaa5de9-964e-479e-8682-12a1f0c044e3` (created with ethdo).

----

Necessary account commands can be added in a subsequent PR or in this one.


packages/lodestar-cli/src/cmds/validator/options.ts
```diff
@@ -0,0 +1,47 @@
...
+  server: {
+    default: "http://127.0.0.1:9596",
+    alias: ["server"],
+    description: "Address to connect to BeaconNode",
+    type: "string"
+  },
+
+  spec: {
+    default: "mainnet",
+    alias: ["spec", "s"],
```
wemeetagain: Can we call this `chain.preset` to align with the `beacon` command?

We will also want `chain.params` to allow for overriding `IBeaconParams`, as in the `beacon` command.

packages/lodestar-cli/src/cmds/validator/options.ts
```diff
@@ -0,0 +1,47 @@
...
+    description: "Specifies the default eth2 spec type",
+    choices: ["mainnet", "minimal", "interop"],
```
wemeetagain: We only support mainnet and minimal. Interop is no more :dagger:

packages/lodestar-cli/src/cmds/validator/options.ts
```diff
@@ -0,0 +1,47 @@
...
+  server: string;
+  spec: string;
+  validatorsDir: string;
+  secretsDir: string;
+}
+
+export const validatorOptions: CommandBuilder<{}, IValidatorCliOptions> = {
+  datadir: {
+    default: "validator",
+    alias: ["datadir", "d"],
```
wemeetagain: Can we call this `validatorDir` - the idea being each command has a root directory that by default, it will save all data to. Matches the `beaconDir` of the `beacon` command.

Or alternatively, we could rename `beaconDir` to `datadir` as here? just looking to consistency.

packages/lodestar-cli/src/cmds/validator/options.ts
```diff
@@ -0,0 +1,47 @@
...
+    default: "mainnet",
+    alias: ["spec", "s"],
+    description: "Specifies the default eth2 spec type",
+    choices: ["mainnet", "minimal", "interop"],
+    type: "string"
+  },
+
+  validatorsDir: {
+    default: "~/.lodestar/validators",
+    alias: ["validators-dir"],
```
wemeetagain: Can we call this `keystoresDir`? (`yargs` will auto hyphen-case any camelCased option name)

packages/lodestar-cli/src/cmds/validator/options.ts
```diff
@@ -0,0 +1,47 @@
...
+    description: "The directory for storing validator keystores",
+    type: "string"
+  },
+
+  secretsDir: {
+    default: "~/.lodestar/secrets",
+    alias: ["secrets-dir"],
+    description: "The directory for storing validator keystore secrets",
+    type: "string"
+  }
```
wemeetagain: I think `datadir`, `validatorsDir`, and `secretsDir` (gonna call them `validatorDir`, `keystoresDir`, `secretsDir` here) should be structured in a way similar to:
src/cmds/beacon/options/beaconDir.ts - for `validatorDir`
src/cmds/beacon/options/beaconFile.ts - for `keystoresDir` and `secretsDir`

See how those files are used: options are declared as functions, which let higher level defaults impact lower level defaults.

What we'd want to do is set defaults based on a tree of options:
- rootDir
  - validatorDir
    - keystoresDir
    - secretsDir

so that higher level options influence the default values of lower level options.
This way the user can override at any level in this hierarchy and have the effects downstream. 

eg: `lodestar validator --root-dir .altona` (like what we've been doing for our beacon testnet config)
Validator keys/passwords will be loaded from `./.altona/validator/keystores` and `./.altona/validator/secrets`
(we could also run the beacon node like `lodestar beacon run --root-dir .altona`, and all testnet data would be fully contained in the `./.altona` directory, very nice)

eg: `lodestar validator --validator-dir myvalidators` (more likely configuration choice once mainnet is live)
Validator keys/passwords will be loaded from `./myvalidators/keystores` and `./myvalidators/secrets`

eg: `lodestar validator --keystores-dir somekeystores --secrets-file secrets_on_temporarily_mounted_path`
Validator keys loaded from `./somekeystores`, passwords loaded from `./secrets_on_temporarily_mounted_path`

Doing this will likely remove the need for `resolveTildePath` in validator/run.ts below.

packages/lodestar-cli/src/cmds/validator/run.ts
```diff
@@ -0,0 +1,77 @@
...
+    throw new YargsError(`secretsDir ${secretsDir} does not exist`);
+
+  const validatorKeypairs = unlockDirKeypairs({keystoresDir: validatorsDir, secretsDir});
+  if (validatorKeypairs.length === 0)
+    throw new YargsError(`There are no validator keystores in ${validatorsDir}`);
+  logger.info(`Decrypted ${validatorKeypairs.length} validator keystores`);
+
+  const validators: Validator[] = validatorKeypairs.map((keypair): Validator => {
+    const id = keypair.publicKey.toHexString().slice(0, 10);
+    const dbPath = join(datadir, "validators", id);
```
wemeetagain: Probably should name the db path after the full public key, just for clarity.
Can we also include the name `db` in the path somewhere
eg: `join(datadir, "db", keypair.publicKey.toHexString())`

packages/lodestar-cli/src/index.ts
```diff
@@ -3,16 +3,32 @@ import * as yargs from "yargs";
...
   .help()
   .wrap(yargs.terminalWidth())
+  .fail((msg, err) => {
+    if (err) {
+      if (err instanceof YargsError) {
+        // eslint-disable-next-line no-console
+        console.error(` ✖ ${err.message}\n`);
+      } else {
+        // eslint-disable-next-line no-console
+        console.error(` ✖ ${err.stack}\n`);
```
wemeetagain, line 30: love the x

packages/lodestar-cli/src/cmds/account/utils/unlockKeypair.ts
```diff
@@ -0,0 +1,67 @@
...
+ * Name mapping of keystore to password file
+ * Note: to be defined in EIP-2334
+ * @param keystore Encrypted keystore object
+ * @param secretsDir Directory containing keystore passwords
+ */
+function readPasswordFile(keystore: Keystore, secretsDir: string): string {
+  const passwordPath = path.join(secretsDir, `0x${keystore.pubkey}`);
+  try {
+    // Data may end with '\n', trim it
+    return fs.readFileSync(passwordPath, "utf8").trim();
```
wemeetagain: I think we need to manually trim the string. `String#trim` seems too aggressive, trims any whitespace from both the front and back. (not sure who would start a password with whitespace, but :shrug: ).

I think we just want to trim a single `\n` from the end of the string? Not sure what the 'right' thing to do is here.

packages/lodestar-cli/src/cmds/account/utils/unlockKeypair.ts
```diff
@@ -0,0 +1,67 @@
...
+  keystorePath,
+  secretsDir,
+}: {
+  keystorePath: string;
+  secretsDir: string;
+}): Keypair {
+  try {
+    const keystore = Keystore.fromJSON(fs.readFileSync(keystorePath, "utf8"));
+    const password = readPasswordFile(keystore, secretsDir);
+    const privKey = keystore.decrypt(password);
```
wemeetagain, line 41: Not related to this PR, just for tracking somewhere:

I think the function signature may need to change here OR we make keystore password string->buffer conversion more robust. There is some ambiguity surrounding how to go convert from unicode characters to bytearray, requires investigation. From the Eth2 key management call:
> EIP 2335: Unicode and bytes as passwords for keystores
>
> Note UTF-8 is specifies how to encode but does not specify the order
> Pick one of the used form from unicode (like ENS, nfkd?)

packages/lodestar-cli/src/cmds/validator/run.ts
```diff
@@ -0,0 +1,77 @@
...
+  const secretsDir = resolveTildePath(options.secretsDir);
+  
+  await initBLS();
+
+  const config = getBeaconConfig(spec);
+
+  const logger = new WinstonLogger();
+
+  if (!fs.existsSync(validatorsDir))
+    throw new YargsError(`validatorsDir ${validatorsDir} does not exist`);
```
wemeetagain: :+1: `YargsError` would be great to add to the rest of the CLI

packages/lodestar-cli/src/cmds/validator/options.ts
```diff
@@ -0,0 +1,47 @@
+import {CommandBuilder} from "yargs";
+
+export interface IValidatorCliOptions {
```
wemeetagain: I see you followed the lighthouse cli https://lighthouse-book.sigmaprime.io/validator-create.html
solid foundation :+1:

wemeetagain (changes requested):
Solid foundation, looks great! much needed functionality!

I think we can tackle the validator lockfile issue separately.

packages/lodestar-cli/src/cmds/validator/options/validatorDir.ts
```diff
@@ -0,0 +1,25 @@
...
+import {Options} from "yargs";
+
+export interface IValidatorDirArgs extends IGlobalArgs {
+  dbDir: string;
+  validatorDir: string;
+}
+
+import {IGlobalArgs} from "../../../options";
+
+export const dbDir = (args: IGlobalArgs): Options => ({
```
wemeetagain: I think dbDir should go in `validatorFile.ts` so it can be placed in the validatorDir and its default can be `path.join(args.validatorDir, "validator-db")`

wemeetagain (changes requested):

packages/lodestar-cli/src/cmds/validator/run.ts
dapplion: Opened an issue for it https://github.com/ChainSafe/lodestar/issues/1099

packages/lodestar-cli/src/cmds/account/utils/unlockKeypair.ts
dapplion: Could you clarify the implications here?

packages/lodestar-cli/src/cmds/account/utils/unlockKeypair.ts
dapplion: Check the latest implementation which only trims a trailing newline. While it's not great to modify the contents of a password file, working with `ethdo` I've a had a ton of pain for trailing new-lines that got added automatically depending on the text editor

packages/lodestar-cli/src/cmds/validator/options/validatorDir.ts
dapplion: All options merged in the same file

packages/lodestar-cli/src/cmds/account/utils/unlockKeypair.ts
wemeetagain: I think we can just make an issue in bls-keystore, handle it there.
Basically, converting string->Buffer is not necessarily deterministic, which is problematic when doing crypto operations on 'string' passwords. The same UTF-8 string can create one of several different outcomes. There are normalization techniques to get a deterministic string->Buffer outcome, we need to apply the technique agreed on in the eth2 key-management standards conversation, and/or ensure that the default implementation uses the correct technique.

wemeetagain: edit: bls-keystore! not bls-keygen library

packages/lodestar-cli/src/cmds/validator/paths.ts
```diff
@@ -0,0 +1,21 @@
...
+  keystoresDir?: string;
+  secretsDir?: string;
+}
+
+/**
+ * Defines the dynamic path structure of the validator files
+ */
+export function processValidatorPaths(options: IValidatorPaths): Required<IValidatorPaths> {
+  const rootDir = options.rootDir;
+  const dbDir = path.join(rootDir, options.dbDir || "validator-db");
```
wemeetagain: Can we put this inside the `validatorDir` on the same level as the keystores and secrets?

wemeetagain (changes requested):

wemeetagain (approved):

twoeths:
Lgtm. Do we have any Readme on how to set up keystores and keystore secret somewhere?

dapplion:
> Lgtm. Do we have any Readme on how to set up keystores and keystore secret somewhere?

I'm working now on the account CLI who should do that. To play around with the validator CLI you should use `ethdo` or `lighthouse` account managers to generate the necessary keystores

twoeths (approved):

PR #1101, opened 2020-07-05
Detect Altona genesis state

twoeths:
+ Add `genesis.test.ts` to detect Altona genesis state, similar to what we did to Schlesi and compare found genesis state to https://github.com/goerli/altona/tree/master/altona
+ Need to save/process/push a block even it has no deposit event
+ Since Altona has a lot of invalid deposts, there are a lot of `getBlock` calls unnecessarily even it's way before genesis block. So I implement checkpoint strategy: if preGenesis only call `getBlock` if it's at checkpoint and set the next checkpoint based on `SECONDS_PER_ETH1_BLOCK` so the `getBlock` calls is likely reduced from `n` to `log2(n)`
+ SECONDS_PER_ETH1_BLOCK may not be the same to the reality but since we set checkpoint frequently at "`distance of current block to genesis/2`" , the "`delta`" is gradually smaller, it should be fine. Added unit tests to make sure.

## Tests
+ In CI we can find Altona genesis state in just 1 min but in my local environment, it can be up to 6 minutes depending on network situation
```
@chainsafe/lodestar: 2020-07-05 09:22:27  [ETH1]             info: Set checkpoint to 2947075 
@chainsafe/lodestar: 2020-07-05 09:22:27  [ETH1]             info: At block 2947075, probably 2 blocks to genesis time if there is enough validators 
@chainsafe/lodestar: 2020-07-05 09:22:27  [ETH1]             info: At block 2947076, probably 1 blocks to genesis time if there is enough validators 
@chainsafe/lodestar: 2020-07-05 09:22:27  [ETH1]             info: At block 2947077, probably 0 blocks to genesis time if there is enough validators 
@chainsafe/lodestar: 2020-07-05 09:22:33  [GENESIS]          info: Found genesis state at eth1 block 2947078 
@chainsafe/lodestar: 2020-07-05 09:22:33  [ETH1]             info: Unsubscribed eth1 blocks & depoosit events 
@chainsafe/lodestar: 2020-07-05 09:22:33  []                 info: detect altona genesis state - duration=59587ms
```

### DIFF
--- a/packages/lodestar/src/eth1/impl/ethers.ts
+++ b/packages/lodestar/src/eth1/impl/ethers.ts
@@ -14,6 +14,8 @@ import {IEth1Options} from "../options";
 import {IEth1Notifier, IDepositEvent, Eth1Block, Eth1EventsBlock} from "../interface";
 import {groupDepositEventsByBlock} from "./util";
 import pushable, {Pushable} from "it-pushable";
+import {calculateStateTime} from "../../chain/genesis/util";
+import {Eth1Data} from "@chainsafe/lodestar-types";
 
 export interface IEthersEth1Options extends IEth1Options {
   contract?: Contract;
@@ -46,6 +48,12 @@ export class EthersEth1Notifier implements IEth1Notifier {
 
   private startedProcessEth1: boolean;
   private lastProcessedEth1BlockNumber: number;
+  private lastDepositCount: number;
+  /**
+   * Pregenesis block number to check remaining time/block to form genesis.
+   * This helps avoid calling too many unnecessary getBlock() calls before genesis.
+   */
+  private preGenesisCheckpoint: number;
   private eth1Source: Pushable<Eth1EventsBlock>;
 
   public constructor(opts: IEthersEth1Options, {config, db, logger}: IEthersEth1Modules) {
@@ -63,6 +71,7 @@ export class EthersEth1Notifier implements IEth1Notifier {
       );
     }
     this.contract = opts.contract;
+    this.preGenesisCheckpoint = undefined;
   }
 
   /**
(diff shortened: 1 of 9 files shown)


packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -132,30 +140,38 @@ export class EthersEth1Notifier implements IEth1Notifier {
    * @param toNumber
    */
   public async processBlocks(toNumber: number): Promise<void> {
+    this.logger.info(`Processing eth1 blocks from ${this.lastProcessedEth1BlockNumber + 1} to ${toNumber}`);
     let rangeBlockNumber = this.lastProcessedEth1BlockNumber;
     while (rangeBlockNumber < toNumber && this.startedProcessEth1) {
-      const blockNumber = Math.min(this.lastProcessedEth1BlockNumber + 100, toNumber);
+      const endRangeBlockNumber = Math.min(this.lastProcessedEth1BlockNumber + 1000, toNumber);
```
twoeths, line 146: I has to tweak this a couple of times to find a best number. Do we want to make this as an eth1 option?

wemeetagain (commented):
Is this correct:
- The checkpointing feature stops us from fetching blocks after min_validators number of deposit events have been found, but still long before min_genesis_time?
- Reason being that checking deposit event count is not accurate to the number of validators, as there are invalid deposit events.
- So even though there are min_validators deposit events, there may be many fewer processed validators, and checking every block is expensive and slows down finding the genesis in cases where it has already happened.
- Once min_genesis_time, however, we resume checking every block for eth1 data.

twoeths:
@wemeetagain  yeah that's all correct. The checkpoint stuff is for preGenesis only, everything is normal after MIN_GENESIS_TIME as we resume checking block by block

wemeetagain (approved):
lgtm!
I'm hoping that in the future we can move to something like #1103 so that we won't need all this switching functionality in a monolithic Eth1Notifier. Its getting kinda hard to follow. Other than that, seems like a good strategy for now to cut down on calls to `getBlock`.

PR #1104, opened 2020-07-06
Use TreeBacked<BeaconState> everywhere in lodestar package

### LINKED ISSUE #1100: Use TreeBacked<BeaconState> in IBeaconChain
_Originally posted by @wemeetagain in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMjc5ODc3NTgxOnYy/pull_request_review_threads/discussion_

I think we should create a:
```typescript
export interface ITreeStateContext {
  epochCtx: EpochContext;
}
```
and use TreeBacked beacon state here and across the `lodestar` package, rather than currently at the boundary of `StateContextCache`.

eg:
```typescript
interface IBeaconChain {
  ...
  getHeadStateContext(): Promise<ITreeStateContext | null>;
  getHeadState(): Promise<TreeBacked<BeaconState> | null>;
}
```

Rationale:

We rely on the efficiency of a tree-backed state, and, practically, can't run a production beacon node without it. If we accept this, there's not a benefit to obscuring the backing throughout the `lodestar` package, as we have been doing.

This will also open the way for serving proofs :)

twoeths:
resolves #1100 
+ Fix unit test interopSubnetsJoiningTask.test.ts

### DIFF
--- a/packages/lodestar/test/utils/state.ts
+++ b/packages/lodestar/test/utils/state.ts
@@ -2,21 +2,26 @@ import {BeaconState} from "@chainsafe/lodestar-types";
 import {GENESIS_EPOCH, GENESIS_SLOT, ZERO_HASH} from "../../src/constants";
 import {generateEmptyBlock} from "./block";
 import {config as mainnetConfig} from "@chainsafe/lodestar-config/lib/presets/mainnet";
+import {TreeBacked} from "@chainsafe/ssz";
 
 /**
  * Copy of BeaconState, but all fields are marked optional to allow for swapping out variables as needed.
  */
 type TestBeaconState = Partial<BeaconState>;
 
+let treeBackedState: TreeBacked<BeaconState>;
+
 /**
  * Generate beaconState, by default it will use the initial state defined when the `ChainStart` log is emitted.
  * NOTE: All fields can be overridden through `opts`.
+*  should allow 1st test calling generateState more time since TreeBacked<BeaconState>.createValue api is expensive.
+ *
  * @param {TestBeaconState} opts
  * @param config
  * @returns {BeaconState}
  */
-export function generateState(opts?: TestBeaconState, config = mainnetConfig): BeaconState {
-  return {
+export function generateState(opts: TestBeaconState = {}, config = mainnetConfig): TreeBacked<BeaconState> {
+  const defaultState: BeaconState = {
     genesisTime: Math.floor(Date.now() / 1000),
     genesisValidatorsRoot: ZERO_HASH,
     slot: GENESIS_SLOT,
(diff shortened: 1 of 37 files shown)


packages/lodestar/test/utils/state.ts
```diff
@@ -61,6 +66,12 @@ export function generateState(opts?: TestBeaconState, config = mainnetConfig): B
       epoch: GENESIS_EPOCH,
       root: ZERO_HASH,
     },
-    ...opts,
   };
+  treeBackedState = treeBackedState || config.types.BeaconState.tree.createValue(defaultState);
```
twoeths, line 70: Generating tree backed BeaconState is expensive (~4s in my environment) so I made a default tree backed state here. For the 1st test calling this function, I have to increase timeout for it. The down side is if we run 1 single test in local environment using `it.only` we may see this timeout issue if that test calls `generateState()`

Not sure if there're any better ways to resolve this.

wemeetagain: Maybe we can add a mocha `before` (outside of any `describe`) to call generateState
eg: `test/unit/setupState.ts` and the same for e2e

twoeths: thanks, that's a good idea

packages/lodestar/test/utils/state.ts
```diff
@@ -2,21 +2,26 @@ import {BeaconState} from "@chainsafe/lodestar-types";
 import {GENESIS_EPOCH, GENESIS_SLOT, ZERO_HASH} from "../../src/constants";
 import {generateEmptyBlock} from "./block";
 import {config as mainnetConfig} from "@chainsafe/lodestar-config/lib/presets/mainnet";
```
mpetrunic:
```suggestion
import {config as minimal} from "@chainsafe/lodestar-config/lib/presets/minimal";
```

packages/lodestar/test/utils/state.ts
mpetrunic: Please use minimal preset for state, we cannot have mainnet state in unit tests

mpetrunic (changes requested):

packages/lodestar/test/utils/state.ts
twoeths: so we should use minimal preset for unit tests and mainnet preset for e2e tests? Fyi, this file is shared for both unit and e2e tests.

mpetrunic: In my opinion we can use minimal in e2e as well. Only place for mainnet would be benchmark

mpetrunic (approved):

wemeetagain (approved):

PR #1107, opened 2020-07-07
Expose getting finalized blocks by root

### LINKED ISSUE #1070: Get finalized block by root

mpetrunic:
resolves #1070

### DIFF
--- a/packages/lodestar/src/api/rest/controllers/beacon/blocks/getBlockHeaders.ts
+++ b/packages/lodestar/src/api/rest/controllers/beacon/blocks/getBlockHeaders.ts
@@ -1,13 +1,13 @@
 import {Root, Slot} from "@chainsafe/lodestar-types";
 import {ApiController} from "../../types";
 
-export const getBlockHeaders: ApiController<{slot?: string; parent_root?: string}> = {
+export const getBlockHeaders: ApiController<{slot?: string|number; parent_root?: string}> = {
 
   url: "/v1/beacon/headers",
 
   handler: async function (req, resp) {
     let slot: Slot|undefined;
-    if(req.query.slot || req.query.slot == "0") {
+    if(req.query.slot || req.query.slot === 0) {
       slot = this.config.types.Slot.fromJson(req.query.slot);
     }
     let parentRoot: Root|undefined;
--- a/packages/lodestar/src/api/rest/controllers/beacon/blocks/getBlockHeaders.ts
+++ b/packages/lodestar/src/api/rest/controllers/beacon/blocks/getBlockHeaders.ts
@@ -7,7 +7,7 @@ export const getBlockHeaders: ApiController<{slot?: string; parent_root?: string
 
   handler: async function (req, resp) {
     let slot: Slot|undefined;
-    if(req.query.slot) {
+    if(req.query.slot || req.query.slot == "0") {
       slot = this.config.types.Slot.fromJson(req.query.slot);
     }
     let parentRoot: Root|undefined;
--- a/packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
+++ b/packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
@@ -32,13 +32,13 @@ export class BlockArchiveRepository extends Repository<Slot, SignedBeaconBlock> 
 
   public async getByRoot(root: Root): Promise<SignedBeaconBlock|null> {
     const slot = await this.getSlotByRoot(root);
-    if(slot || slot === 0) {
+    if(Number.isInteger(slot)) {
       return this.get(slot);
     }
     return null;
   }
 
-  public async getSlotByRoot(root: Root): Promise<Slot> {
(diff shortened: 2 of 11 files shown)


packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
```diff
@@ -22,6 +23,42 @@ export class BlockArchiveRepository extends Repository<Slot, SignedBeaconBlock>
...
+  public async add(value: SignedBeaconBlock): Promise<void> {
+    await Promise.all([
+      super.add(value),
+      this.storeRootRef(value)
+    ]);
+  }
+
+  public async getByRoot(root: Root): Promise<SignedBeaconBlock|null> {
+    const slot = await this.getSlotByRoot(root);
+    if(slot || slot === 0) {
```
wemeetagain: `Number.isInteger`

packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
```diff
@@ -22,6 +23,42 @@ export class BlockArchiveRepository extends Repository<Slot, SignedBeaconBlock>
...
+      return this.get(slot);
+    }
+    return null;
+  }
+
+  public async getSlotByRoot(root: Root): Promise<Slot> {
```
wemeetagain: return Promise<Slot | null>

packages/lodestar/src/api/rest/controllers/beacon/blocks/getBlockHeaders.ts
```diff
@@ -7,7 +7,7 @@ export const getBlockHeaders: ApiController<{slot?: string; parent_root?: string
 
   handler: async function (req, resp) {
     let slot: Slot|undefined;
-    if(req.query.slot) {
+    if(req.query.slot || req.query.slot == "0") {
```
wemeetagain: Is this needed? `Boolean("0") === true`
Also seems like something that fastify schema validation should find?

If it is, should use ===

wemeetagain (changes requested):
Tiny things, Looks good!

packages/lodestar/src/api/rest/controllers/beacon/blocks/getBlockHeaders.ts
mpetrunic: > Also seems like something that fastify schema validation should find?

slot 0 is valid, and we are suppose to receive slots as string.
Used == as it would catch both number zero and string zero
req.query.slot === 0 is fine too I think

twoeths (dismissed):

twoeths:
Lgtm!

twoeths:
Not sure why fast spec test is failing in this PR.

packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
```diff
@@ -22,6 +23,42 @@ export class BlockArchiveRepository extends Repository<Slot, SignedBeaconBlock>
     super(config, db, Bucket.blockArchive, config.types.SignedBeaconBlock);
```
wemeetagain, line 23: Can you extend the whole interface for completeness?
Will be helpful for eg: a db sniffer utility. And there's nothing warning people from using methods that don't respect the secondary index.

mpetrunic: What methods does not respect secondary index?

Cannot really extend delete methods since they only receive slot but it doesn't really matter if it points to non existent slot, it will return null anyways.

wemeetagain: `put`, `delete`, `remove`, `batchPut`, `batchDelete`, `batchRemove`

mpetrunic: I should probably override put and batchPut instead of add and batchAdd. remove/delete/batchDelete/batchRemove are useless to overide

wemeetagain: remove / batchRemove is useful (and easy to implement) for db diagnostic tooling. We should have it for completeness, and because there is no warning to users that it won't respect the secondary index.

Hypothetical:
- synced a lot of blocks
- there's a bug in our state transition
- we want to "remove" a finalized block from the db

Other hypothetifcal:
- synced a lot of blocks
- we figure out a better on-disk format / db for storing _very old_ blocks but not recently old blocks
- we want to delete the oldest blocks and move them to another db

wemeetagain (approved):

Issue #1108, opened 2020-07-07
GENESIS_FORK_VERSION copied incorrectly from template-config-file


wemeetagain:
**Describe the bug**
When using a template config file eg: `lodestar beacon init --template-config-file foo.json`
If the template file contains a hex string, it will be copied incorrectly, converted to a number, and copied as a number.

**Expected behavior**
It copies hex string unmodified.

**Steps to Reproduce**
Follow instructions here: https://hackmd.io/2wNvrO77RQ6BWc2ZA7M0Cg
inspect generated beacon.config.json

Related to https://github.com/ChainSafe/lodestar/issues/977

3xtr4t3rr3str14l:
resolved in #1136

PR #1109, opened 2020-07-07
check for existing template-config-file

### LINKED ISSUE #1106: beacon init does not error when providing a non-existent template config file
**Describe the bug**
When I run `yarn run cli beacon init     --eth1.enabled=false     --chain.genesisStateFile=genesis.ssz     --template-config-file altona.json     --root-dir .altona/`, I get no errors and the output from the cli just shows what it would normally show when init has been ran successfully.

**Expected behavior**
I would expect to see some sort of error telling the user to provide a .json file

**Steps to Reproduce**
`lerna clean`
`lerna bootstrap`
`yarn run cli beacon init     --eth1.enabled=false     --chain.genesisStateFile=genesis.ssz     --template-config-file altona.json     --root-dir .altona/`

3xtr4t3rr3str14l:
-fixes #1106

### DIFF
--- a/packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
+++ b/packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
@@ -14,9 +14,14 @@ export function builder(yargs: Argv<IBeaconArgs>): Argv<IBeaconArgs> {
   yargs.options(beaconInitOptions);
   const args = parseArgs(yargs) as IBeaconInitArgs;
 
-  if (args.templateConfigFile) {
-    require("fs").readFileSync(args.templateConfigFile, "utf8");
-    yargs.config(readBeaconConfig(args.templateConfigFile));
+  if (args.chain && args.chain.genesisStateFile) {
+    require("fs").readFileSync(args.chain.genesisStateFile, "utf8");
+  }
+
+  const { templateConfigFile } = args;
+  if (templateConfigFile) {
+    require("fs").readFileSync(templateConfigFile, "utf8");
+    yargs.config(readBeaconConfig(templateConfigFile));
   }
   return yargs as unknown as Argv<IBeaconArgs>;
 }
--- a/packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
+++ b/packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
@@ -13,7 +13,6 @@ export const description = "Initialize lodestar beacon node";
 export function builder(yargs: Argv<IBeaconArgs>): Argv<IBeaconArgs> {
   yargs.options(beaconInitOptions);
   const args = parseArgs(yargs) as IBeaconInitArgs;
-
   if (args.templateConfigFile) {
     yargs.config(readBeaconConfig(args.templateConfigFile));
   }
--- a/packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
+++ b/packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
@@ -5,7 +5,6 @@ import {readBeaconConfig} from "../../config";
 import {IBeaconArgs} from "../../options";
 import {beaconInitOptions, IBeaconInitArgs} from "./options";
 import {init} from "./init";
-import fs from "fs";
 
 export const command = "init";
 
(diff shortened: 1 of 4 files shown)


3xtr4t3rr3str14l:
just realized a similar errors happens when providing a non-existent genesisStateFile, so i'm gonna push something for that in a second here too

wemeetagain (changes requested):
I think we just need to remove the silly try catch from readBeaconConfig. We don't want to coerce here.

Edit: alternatively, add a file-exists check in src/cmds/beacon/cmds/init/index.ts

packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
```diff
@@ -13,8 +14,15 @@ export const description = "Initialize lodestar beacon node";
 export function builder(yargs: Argv<IBeaconArgs>): Argv<IBeaconArgs> {
   yargs.options(beaconInitOptions);
   const args = parseArgs(yargs) as IBeaconInitArgs;
-  if (args.templateConfigFile) {
-    yargs.config(readBeaconConfig(args.templateConfigFile));
+
+  if (args.chain && args.chain.genesisStateFile) {
+    fs.readFileSync(args.chain.genesisStateFile, "utf8");
+  }
+
```
wemeetagain: This check shouldn't be done here.
Check beacon/cmds/run/run.ts. Add a try/catch around the genesis state file `fs.promises.readFile` and catch, throw a YargsError with a descriptive error msg.

packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
```diff
@@ -13,8 +14,15 @@ export const description = "Initialize lodestar beacon node";
...
+  const {templateConfigFile} = args;
+  if (templateConfigFile) {
+    fs.readFileSync(templateConfigFile, "utf8");
+    yargs.config(readBeaconConfig(templateConfigFile));
```
wemeetagain: Can you investigate why there's a try/catch around `readBeaconConfig`? Likely should be removed as a better solution than this.

wemeetagain (changes requested):

packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
3xtr4t3rr3str14l: yep.  just took out the try/catch

3xtr4t3rr3str14l: and took out my initial solution

packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
3xtr4t3rr3str14l: after this PR is done, i think i might try going after #1108 and maybe just include the fix for this in that PR instead

3xtr4t3rr3str14l: so for now, just gonna use this PR for template-config-file fix

mpetrunic (dismissed):

packages/lodestar-cli/src/cmds/beacon/config.ts
```diff
@@ -32,11 +32,7 @@ export async function writeBeaconConfig(filename: string, config: Partial<IBeaco
  * This needs to be a synchronous function because it will be run as part of the yargs 'build' step
  */
 export function readBeaconConfig(filename: string): Partial<IBeaconNodeOptions> {
-  try {
-    return readFileSync(filename) as Partial<IBeaconNodeOptions>;
-  } catch(e) {
-    return {};
-  }
+  return readFileSync(filename) as Partial<IBeaconNodeOptions>;
```
wemeetagain, line 35: I think this diff is the only one needed for this PR. The others should be removed

3xtr4t3rr3str14l: done.

3xtr4t3rr3str14l: done.

wemeetagain (approved):

PR #1110, opened 2020-07-07
Add account CLI

dapplion:
Adds an acount CLI feature-pair with `lighthouse`'s account CLI.

### How to test
Normal usage flow of the account CLI could be:
```
lodecli account wallet create --name wallet1 --passphraseFile wallet1.pass
```
```
lodecli account wallet list
```
```
lodecli account validator create --name wallet1 --passphraseFile wallet1.pass --count 1
```
```
lodecli account validator deposit --validator all --eth1Http http://your.unlock.node:8545
```

### Available commands
Summary of the command hierarchy, use --help at any level to get full details and options. 
```
account
  validator
    create    Creates new validators from an existing EIP-2386 wallet...
    deposit   Submits a deposit to an Eth1 validator...
  wallet
    create    Creates a new HD (hierarchical-deterministic) EIP-2386 wallet
    list      Lists the names of all wallets
```

### Directory structure
All major directories are one level deep of the global `rootDir` path. This account directory structure is very similar to `ethdo` and `ighthouse`'s, expect that this one mixes account data with db data.
```
.lodestar/
├── keystores
│   └── 0x871cb1c0b1e6b67473003ccfbdf9532d3b0d29bf51b07df6a3fae2ba25526d18e67a119babd44aa3b22b9bc4ef16f71f
│       ├── eth1-deposit-data.rlp
│       ├── eth1-deposit-gwei.txt
│       └── voting-keystore.json
├── secrets
│   └── 0x871cb1c0b1e6b67473003ccfbdf9532d3b0d29bf51b07df6a3fae2ba25526d18e67a119babd44aa3b22b9bc4ef16f71f
├── validator-db
│   └── 0x871cb1c0b1e6b67473003ccfbdf9532d3b0d29bf51b07df6a3fae2ba25526d18e67a119babd44aa3b22b9bc4ef16f71f
│       ├── 000003.log
│       ├── CURRENT
│       ├── LOCK
│       ├── LOG
│       └── MANIFEST-000002
└── wallets
    └── 05939fbd-923a-4559-b41f-2e43ad8fd80b
        └── 05939fbd-923a-4559-b41f-2e43ad8fd80b
```

### Other minor changes
- Clean CLI export style to prevent the use of 'as any' casting. Type-safety = :smile: 
- CLI now prints help when no command is provided and it's required.
- CLI now recommends possible command names if no command is matched
- Added welcome message in help of the root command
```
🌟 Lodestar: Ethereum 2.0 TypeScript Implementation of the Beacon Chain
```

### Next steps
- Provide more options to broadcast a deposit data, i.e. connect to Metamask, IPC, Ledger?
- Support non-HD wallets
- Add tests


packages/lodestar-cli/src/cmds/validator/paths.ts
```diff
@@ -1,21 +1,31 @@
...
-  const rootDir = options.rootDir;
-  const validatorDir = path.join(rootDir, options.validatorDir || "validator");
-  const dbDir = path.join(validatorDir, options.dbDir || "validator-db");
-  const keystoresDir = path.join(validatorDir, options.keystoresDir || "validators");
-  const secretsDir = path.join(validatorDir, options.keystoresDir || "secrets");
-  return {rootDir, dbDir, validatorDir, keystoresDir, secretsDir};
-}
\ No newline at end of file
+export function getValidatorPaths(options: Partial<IValidatorPaths> & Pick<IGlobalArgs, "rootDir">): IValidatorPaths {
+  const rootDir = options.rootDir || ".lodestart";
```
wemeetagain: lodestart intentional? We likely should pull this to a constant so it can be reused everywhere this pattern is used.

packages/lodestar-cli/src/cmds/account/cmds/validator/deposit.ts
```diff
@@ -0,0 +1,91 @@
...
+import {ValidatorDirManager} from "../../../../validatorDir";
+import {getAccountPaths} from "../../paths";
+import {getBeaconConfig} from "../../../../util";
+import {IAccountValidatorOptions} from "./options";
+
+const DEPOSIT_GAS_LIMIT = 400000;
+
+export const command = "deposit";
+
+export const description = `Submits a deposit to an Eth1 validator registration contract via an IPC endpoint
```
wemeetagain: remove IPC?

packages/lodestar-cli/src/cmds/account/cmds/validator/create.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export async function handler(options: IValidatorCreateOptions): Promise<void> {
+  const name = options.name;
+  const passphraseFile = options.passphraseFile;
+  const spec = options.chain.name;
+  const storeWithdrawalKeystore = options.storeWithdrawalKeystore;
+  const count = options.count;
+  const atMost = options.atMost;
+  const accountPaths = getAccountPaths(options);
+  const config = getBeaconConfig(spec);
```
wemeetagain: We will want a way to override params, like in `beacon run`.
Especially important here is `GENESIS_FORK_VERSION`.
So the chain.params will need to be an input here.

packages/lodestar-cli/src/cmds/account/cmds/validator/create.ts
```diff
@@ -0,0 +1,120 @@
...
+import {getAccountPaths} from "../../paths";
+import {WalletManager} from "../../../../wallet";
+import {ValidatorDirBuilder} from "../../../../validatorDir";
+import {stripOffNewlines, randomPassword, getBeaconConfig, YargsError} from "../../../../util";
+import {IAccountValidatorOptions} from "./options";
+
+export const command = "create";
+
+export const description = "Creates new validators from an existing EIP-2386 wallet using the EIP-2333 HD key \
+  derivation scheme.";
```
wemeetagain: Maybe mention what is created as a result of running this.

packages/lodestar-cli/src/cmds/account/cmds/validator/deposit.ts
```diff
@@ -0,0 +1,91 @@
...
+interface IAccountValidatorDepositOptions extends IAccountValidatorOptions {
+  keystoresDir: string;
+  validator: string;
+  eth1Http?: string;
+}
+
+export const builder: CommandBuilder<{}, IAccountValidatorDepositOptions> = {
+  validator: {
+    description: "The name of the validator directory in $keystoresDir for which to deposit. \
+    Set to 'all' to deposit all validators in the $keystoresDir.",
```
wemeetagain, line 31: Maybe nice to have a `account validator list` to list available names?

packages/lodestar-cli/src/cmds/account/cmds/validator/deposit.ts
```diff
@@ -0,0 +1,91 @@
...
+    normalize: true,
+    demandOption: true,
+    type: "string",
+  },
+
+  eth1Http: {
```
wemeetagain: eth1Url? eth1ProviderUrl?

packages/lodestar-cli/src/cmds/account/cmds/validator/deposit.ts
```diff
@@ -0,0 +1,91 @@
...
+  for (const validatorDir of validatorDirsToSubmit) {
+    const {rlp, depositData} = validatorDir.eth1DepositData(config);
+    const value = depositData.amount * BigInt(1e9);
+    const tx = await eth1Wallet.sendTransaction({
+      to: depositContractAddress,
+      gasLimit: DEPOSIT_GAS_LIMIT,
+      value: value.toString(),
+      data: rlp
+    });
+    const txHash = tx.hash || "";
```
wemeetagain: Why `|| ""`?
Then we `saveEth1DepositTxHash("")`
Should we warn in this case?

packages/lodestar-cli/src/cmds/account/cmds/validator/deposit.ts
```diff
@@ -0,0 +1,91 @@
...
+const DEPOSIT_GAS_LIMIT = 400000;
+
+export const command = "deposit";
+
+export const description = `Submits a deposit to an Eth1 validator registration contract via an IPC endpoint
+of an Eth1 client (e.g., Geth, OpenEthereum, etc.). The validators must already
+have been created and exist on the file-system. The process will exit immediately
+with an error if any error occurs. After each deposit is submitted to the Eth1
+node, a file will be saved in the validator directory with the transaction hash.
+The application does not wait for confirmations so there is not guarantee that
```
wemeetagain: It looks like it does wait for a confirmation w/ `tx.wait`

packages/lodestar-cli/src/cmds/account/cmds/validator/options.ts
```diff
@@ -0,0 +1,25 @@
...
+export interface IAccountValidatorOptions extends IGlobalArgs, IChainArgs {
+  keystoresDir?: string;
+  secretsDir?: string;
+}
+
+export const accountValidatorOptions = {
+  chainPreset,
+
+  keystoresDir: {
+    description: `The directory for storing validator keystores.\n[default: ${defaultAccountPaths.keystoresDir}]`,
```
wemeetagain: Should the default string thing be pulled into a utility function?

packages/lodestar-cli/src/depositContract/depositData.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  // deposit data with empty signature to sign
+  const depositData: DepositData = {
+    pubkey,
+    withdrawalCredentials,
+    amount,
+    signature: Buffer.alloc(96)
+  };
+
+  const domain = computeDomain(config, DomainType.DEPOSIT);
```
wemeetagain, line 64: domain is calculated with GENESIS_FORK_VERSION here

packages/lodestar-cli/src/wallet/Wallet.ts
```diff
@@ -0,0 +1,113 @@
...
+export interface IWalletKeystoreJson {
+  crypto: object;
+  uuid: string;
+  name: string;
+  nextaccount: number;
+  version: number;
+  type: string;
+}
+
+export class Wallet extends Keystore {
```
wemeetagain, line 26: Ideally we pull this out into a bls-walletstore library

wemeetagain (changes requested):
Looks really good, this is great work.
just minor comments and questions

packages/lodestar-cli/src/depositContract/depositData.ts
mpetrunic: I think this function uses GENESIS_FORK_VERSION by default if not submitted

packages/lodestar-cli/src/cmds/account/cmds/validator/create.ts
```diff
@@ -0,0 +1,120 @@
...
+  const config = getBeaconConfig(spec);
+  const maxEffectiveBalance = config.params.MAX_EFFECTIVE_BALANCE;
+  const depositGwei = BigInt(options.depositGwei || 0) || maxEffectiveBalance;
+
+  // To compute the publicKey of the validator keystores
+  await initBLS();
+
+  if (depositGwei > maxEffectiveBalance) {
+    throw new YargsError(`depositGwei ${depositGwei} is higher than MAX_EFFECTIVE_BALANCE ${maxEffectiveBalance}`);
+  }
```
mpetrunic: We could move all args validation to yargs.check

packages/lodestar-cli/src/cmds/account/cmds/validator/deposit.ts
```diff
@@ -0,0 +1,91 @@
...
+    description: "The name of the validator directory in $keystoresDir for which to deposit. \
+    Set to 'all' to deposit all validators in the $keystoresDir.",
+    normalize: true,
+    demandOption: true,
+    type: "string",
+  },
+
+  eth1Http: {
+    description: "URL to an Eth1 JSON-RPC endpoint with an unlock account to sign",
+    demandOption: true,
```
mpetrunic: we could make it optional and default to ethers default provider

packages/lodestar-cli/src/cmds/account/cmds/validator/deposit.ts
```diff
@@ -0,0 +1,91 @@
...
+  const validatorDirsToSubmit = validatorDirs
+    // txHash file is used as a flag of deposit submission
+    .filter(validatorDir => validatorDir.eth1DepositTxHashExists());
+  
+  if (validatorDirsToSubmit.length === 0)
+    throw Error("No validators to deposit");
+  // eslint-disable-next-line no-console
+  console.log(`Starting ${validatorDirsToSubmit.length} deposits`);
+
+  const eth1Wallet = new ethers.providers.JsonRpcProvider(eth1Http).getSigner();
```
mpetrunic: This assumes eth1 node/provider has enabled signer which is quite unsafe if you also exposes http api (lighthouse is using ipc for that reason). I think we should also have args to supply eth1 keystore and password

packages/lodestar-cli/src/cmds/account/cmds/validator/index.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const builder: CommandBuilder<{}, IAccountValidatorOptions> = (yargs) => {
+  return yargs
+    .options(accountValidatorOptions)
+    .command(create)
+    .command(deposit);
+};
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+export function handler(): void {}
```
mpetrunic: I think we should use "dev" command style export as it's much more typesafe in terms what are you suppose to export and with which type:
https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar-cli/src/cmds/dev/index.ts

cc @wemeetagain

packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  if (mnemonicOutputPath) {
+    writeFile600Perm(mnemonicOutputPath, mnemonic);
+  }
+
+  // eslint-disable-next-line no-console
+  console.log(`
+Your wallet's 12-word BIP-39 mnemonic is:
+
+\t${mnemonic}
```
mpetrunic, line 82: I wonder if we should print this since it would stay in bash history

mpetrunic (changes requested):

packages/lodestar-cli/src/cmds/account/cmds/validator/index.ts
dapplion: I'm not sure about that. Using this exports style the main weakness is misspelling the named exports. However, besides that, yargs does a great job at infering the option and handlers types and make sure they are consistent. Also using the `CommandModule` approach forced the author to cast the module entirely to `any` in `lodestar-cli/src/index.ts`:

https://github.com/ChainSafe/lodestar/blob/8524b8086320997bc91251111af45782b4da60cf/packages/lodestar-cli/src/index.ts#L14-L15

packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
dapplion: How could we provide the mnemonic to the user otherwise?

packages/lodestar-cli/src/cmds/account/cmds/validator/index.ts
dapplion: Agree! I've converted all commands to the style you suggest except for the third level child commands such as `account validator create`, which I get type errors on their parent builder that the options do not match.

packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
mpetrunic: It's inside wallet store, but we can leave this for now

packages/lodestar-cli/src/cmds/account/cmds/validator/deposit.ts
dapplion: I've added more options to connect to the eth1 node. With local keystore, IPC path, and RPC URL (locked and unlocked)

packages/lodestar-cli/src/depositContract/depositData.ts
dapplion: So should this be changed?

packages/lodestar-cli/src/cmds/account/cmds/validator/deposit.ts
dapplion: According to ethers types it should not happen but I'm throwing an error in case it does

packages/lodestar-cli/src/cmds/validator/paths.ts
dapplion: Removed since the default value to rootDir is assigned in the top-level options builder

packages/lodestar-cli/src/cmds/account/cmds/validator/create.ts
dapplion: Could we find a more friendly and intuitive way of achieving the same effect?

packages/lodestar-cli/src/wallet/Wallet.ts
dapplion: Agree, would mind taking care of creating the library's repo and NPM package?

packages/lodestar-cli/src/cmds/account/cmds/validator/create.ts
dapplion: Since `maxEffectiveBalance` is dynamically computed from `getBeaconConfig`, moving this to the `check` function will result in code duplication. Also by throwing a `YargsError` you achieve the same effect as `check` as it will show usage help. Do you think it's worth to move it to `check` anyway?

packages/lodestar-cli/src/cmds/account/cmds/validator/create.ts
wemeetagain: The direction I've been thinking lately is to do something like:
- create a IBeaconParams ssz type, `createIBeaconParams` to create IBeaconParams from a `Record<string, unknown>` / `Json` #1136 
- Use it to help create yargs `Options` if needed. All Options are `type: "string"`, will be parsed using `createIBeaconParams`
- Have a cli-wide params file. (should be copy-pastable from eg: https://github.com/eth2-clients/eth2-testnets/blob/master/shared/altona/config.yaml)
- Have all lodestar cli commands rely on this separate config file - Likely _every_ command needs the same IBeaconParams.
- First `lodestar-cli` bootup should require a config file, or have `lodestar init` to create it, or other similar behavior.

packages/lodestar-cli/src/depositContract/depositData.ts
wemeetagain: No, just pointing out where the GENESIS_FORK_VERSION is called since it isn't super obvious.

packages/lodestar-cli/src/wallet/Wallet.ts
wemeetagain: :+1: Should we wait until we review https://github.com/ChainSafe/bls-keystore/pull/6 ?

wemeetagain (approved):

packages/lodestar-cli/src/cmds/account/cmds/validator/create.ts
dapplion: Should this be addressed here on in a subsequent PR?

wemeetagain: Yea, likely will involve some edits to all cli commands

packages/lodestar-cli/src/wallet/Wallet.ts
dapplion: I've created an issue in https://github.com/ChainSafe/bls-keystore (for lack of a better place) to track this https://github.com/ChainSafe/bls-keystore/issues/8

mpetrunic (approved):

PR #1111, opened 2020-07-08
Use fast processSlots

wemeetagain:
Use fast `processSlots` across `lodestar`

See https://discordapp.com/channels/593655374469660673/593655641445367808/730077633064206397
for discussion

Things go :boom: when slow `processSlots` decides to `processEpoch`


wemeetagain:
Followup: Analyze gossip validation functions for other uses of slow state transition / utils.
Eg: `getAttestingIndices` should likely be split into `getAttestingIndices` and `getAttestingIndicesFromCommittee`, so we may use the fast `epochCtx.getBeaconCommittee` instead of naively recreating the committee as currently in `getAttestingIndices`.

mpetrunic (approved):

PR #1117, opened 2020-07-08
Start network gossip with sync gossip

wemeetagain:
Fixes the root cause which triggered the PR here: https://github.com/ChainSafe/lodestar/pull/1111

Gossip was started prematurely, DOSed our node by running processSlots across many epochs.
Now we start gossip when the sync gossip handler is started.
Required small tweaks to committee-joining task, to turn off and on with gossip.


packages/lodestar/test/e2e/node.test.ts
```diff
@@ -8,7 +8,7 @@ describe("beacon node", function () {
 
   it("should justify block", async function () {
     this.timeout(120000);
-    const bn = await getDevBeaconNode({SECONDS_PER_SLOT: 2, SLOTS_PER_EPOCH: 8});
+    const bn = await getDevBeaconNode({SECONDS_PER_SLOT: 2, SLOTS_PER_EPOCH: 8}, {sync: {minPeers: 0}});
```
mpetrunic, line 11: Should work without minPeers: 0 since you don't need to gossip if you don't have peers :slightly_smiling_face:

packages/lodestar/test/e2e/node.test.ts
```diff
@@ -24,7 +24,7 @@ describe("beacon node", function () {
 
   it("should finalize block", async function () {
     this.timeout(120000);
-    const bn = await getDevBeaconNode({SECONDS_PER_SLOT: 2, SLOTS_PER_EPOCH: 8});
+    const bn = await getDevBeaconNode({SECONDS_PER_SLOT: 2, SLOTS_PER_EPOCH: 8}, {sync: {minPeers: 0}});
```
mpetrunic, line 27: Should work without minPeers: 0 since you don't need to gossip if you don't have peers :slightly_smiling_face:

packages/lodestar/test/e2e/sync/sync.test.ts
```diff
@@ -9,7 +9,7 @@ describe("syncing", function () {
 
   it("should sync from other BN", async function () {
     this.timeout(0);
-    const bn = await getDevBeaconNode({SECONDS_PER_SLOT: 2, SLOTS_PER_EPOCH: 8}, {}, 8);
+    const bn = await getDevBeaconNode({SECONDS_PER_SLOT: 2, SLOTS_PER_EPOCH: 8}, {sync: {minPeers: 0}}, 8);
```
mpetrunic, line 12: I would keep it with condition to have 1 peer just to see if gossip will enable once it gets peer

packages/lodestar/test/e2e/node.test.ts
wemeetagain: removing this causes bugs, stemming from our sync attestation collector trying to interact with a stopped gossipsub.

Will fix in this PR, since it seems related.

wemeetagain: I think having gossip on (eg; being fully synced on a chain) is a key part of a validator running. If not, the validator's attestations will not be broadcast properly to the network.
In this case, I don't know where the boundary should be, but it is not good to have gossip off while the chain is being driven by validators.

Recommendation is to leave these changes as-is in this PR, make an issue about syncing status and validator input.
Eg: "validator processing should be dependent on syncing status." or something to that effect.
Perhaps there are two separable sub-issues.
One on the vc side: "validator client should initially check the bn sync status"
One on the bn side: "validator api should handle data received during unsynced chain"

mpetrunic: Actually, we should have middleware to return 403 or something from validator related api if node is syncing.

Can we make attestation collector to silently ignore errors if gossip sub is off. I mean there is no point in it if we don't have enough peers anyways?

wemeetagain: It likely shouldn't be a middleware, because we will want the same behavior in the JS api.

I think we can do "the log.error then return" in `sync.collectAttestations`. We should also do the sync check on the API level since that should throw an error, instead of just return.

packages/lodestar/src/sync/sync.ts
```diff
@@ -119,10 +119,8 @@ export class BeaconSync implements IBeaconSync {
     if(this.mode === SyncMode.STOPPED) return;
     this.mode = SyncMode.REGULAR_SYNCING;
     await this.initialSync.stop();
-    await Promise.all([
-      this.gossip.start(),
-      this.regularSync.start()
-    ]);
+    await this.regularSync.start();
```
twoeths: should we swap: make `await this.gossip.start();` first?

packages/lodestar/src/tasks/index.ts
```diff
@@ -50,14 +50,25 @@ export class TasksService implements IService {
...
-    await this.interopSubnetsTask.start();
+    this.network.gossip.on("gossip:start", this.handleGossipStart);
+    this.network.gossip.on("gossip:stop", this.handleGossipStop);
   }
 
   public async stop(): Promise<void> {
     this.chain.removeListener("finalizedCheckpoint", this.handleFinalizedCheckpointChores);
+    this.network.gossip.removeListener("gossip:start", this.handleGossipStart);
+    this.network.gossip.removeListener("gossip:stop", this.handleGossipStop);
     await this.interopSubnetsTask.stop();
```
twoeths, line 60: remove this as `this.interopSubnetsTask` is removed in `handleGossipStop`

wemeetagain: We want the interopSubnetsTask to shut down _regardless_, when task manager is shutdown.
In practice, the tasks manager is shutting down before the network, too.

packages/lodestar/test/e2e/node.test.ts
wemeetagain: Just changed this to throw an Error in `sync.collectAttestations` if the chain isn't in regular sync or synced mode. I think it will be better to throw an error here than silently return.
Sync is an internal module, so we know that anything that attempts to run `sync.collectAttestations` without the chain being in regular sync mode is an error.

packages/lodestar/src/sync/sync.ts
wemeetagain: perhaps we can revert to the previous Promise.all()

mpetrunic (approved):

PR #1119, opened 2020-07-09
once/on patch, eth1vote optimization, pubkey cache change, attestation optimizations

protolambda:
Collection of Altona optimizations/hotfixes

- once/on patch (not mine)
- eth1vote optimization
  - previously it hashed the eth1 data of the block over and over again, for each comparison. Super slow. Conver to tree backed, to cache hash-tree-root, and have much faster counting
- pubkey cache hack, index bound safety
  - the pubkeys don't change for eth1 data, unless a bad majority vote is made by an attacker and accepted by the local node.
  - added some safety checks, we don't want not-yet-existing pubkeys of a different beacon fork (but the same eth1 chain) to mess up the transition logic
     - optimized indices check in the attestation processing too

Edit, more optimization:
- convert the block to a tree-backed block. The amount of times the block is merkleized, its parts are merkleized, and how large parts end up in the state (attestation bitfields), hash-tee-root caching is too important.
- Optimize get-indexed-attestation. Don't unnecessarily copy. Don't use a Set when it's unique anyway. In-place sort the (small-ish) list of participants.
- Cut off eth1 vote logic when half of the period is not reached yet. No majority vote can exist before then, thus no state change anyway.

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/fast/index.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/index.ts
@@ -19,13 +19,20 @@ export function fastStateTransition(
   verifySignatures = true,
 ): IStateContext {
   const types = epochCtx.config.types;
-  const block = signedBlock.message;
+
+  // convert to tree-backed block: the root is computed multiple times,
+  // the contents are hash-tree-rooted multiple times, and some of the contents end up in the state as tree-form.
+  const {SignedBeaconBlock} = types;
+  const signedBlockCached = SignedBeaconBlock.tree.asTreeBacked(SignedBeaconBlock.tree.fromStructural(signedBlock));
+
+  const block = signedBlockCached.message;
   const postState = types.BeaconState.clone(state);
   // process slots (including those with no blocks) since block
   processSlots(epochCtx, postState, block.slot);
+
   // verify signature
   if (verifyProposer) {
-    if (!verifyBlockSignature(epochCtx.config, postState, signedBlock)) {
+    if (!verifyBlockSignature(epochCtx.config, postState, signedBlockCached)) {
       throw new Error("Invalid block signature");
     }
   }
--- a/packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts
@@ -17,7 +17,11 @@ class PubkeyIndexMap extends Map<ByteVector, ValidatorIndex> {
 }
 
 export class EpochContext {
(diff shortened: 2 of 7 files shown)


protolambda:
CI has some lint errors that are not mine (I think), tests pass otherwise :tada:

packages/lodestar-beacon-state-transition/src/fast/index.ts
```diff
@@ -19,13 +19,20 @@ export function fastStateTransition(
   verifySignatures = true,
 ): IStateContext {
   const types = epochCtx.config.types;
-  const block = signedBlock.message;
+
+  // convert to tree-backed block: the root is computed multiple times,
+  // the contents are hash-tree-rooted multiple times, and some of the contents end up in the state as tree-form.
+  const {SignedBeaconBlock} = types;
+  const signedBlockCached = SignedBeaconBlock.tree.asTreeBacked(SignedBeaconBlock.tree.fromStructural(signedBlock));
```
mpetrunic, line 26: We should probably do this in validate step of blockprocessing as we calculate root there few times already and state transition won't care about backing then

packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts
```diff
@@ -55,12 +59,9 @@ export class EpochContext {
 
   public copy(): EpochContext {
     const ctx = new EpochContext(this.config);
-    // full copy of pubkeys, this can mutate
-    ctx.pubkey2index = new PubkeyIndexMap();
-    for(const entry of this.pubkey2index.entries()) {
-      ctx.pubkey2index.set(fromHexString(entry[0] as unknown as string), entry[1]);
-    }
-    ctx.index2pubkey = [...this.index2pubkey];
+    // warning: pubkey cache is not copied, it is shared, as eth1 is not expected to reorder validators.
```
mpetrunic, line 62: https://github.com/ChainSafe/lodestar/pull/1120#discussion_r452050220

packages/lodestar-beacon-state-transition/src/fast/index.ts
protolambda: And I am doing it explicitly for the eth1 vote of the block, to make hash tree root cached, and do many repeated equality checks for the vote counting.
If the whole block ends up being tree backed anyway, the eth1 data change can be simplified.

mpetrunic:
> CI has some lint errors that are not mine (I think), tests pass otherwise

It's yours, you left some unused imports :slightly_smiling_face: 

image

protolambda:
Sorry, will try the lint tooling more next time :sweat_smile:

PR #1120, opened 2020-07-09
Proto/altona optimization

wemeetagain:
Proto's updates with with linter fix: See #1119

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/fast/block/processEth1Data.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/block/processEth1Data.ts
@@ -12,12 +12,19 @@ export function processEth1Data(
   const {EPOCHS_PER_ETH1_VOTING_PERIOD, SLOTS_PER_EPOCH} = epochCtx.config.params;
   const SLOTS_PER_ETH1_VOTING_PERIOD = EPOCHS_PER_ETH1_VOTING_PERIOD * SLOTS_PER_EPOCH;
   const newEth1Data = body.eth1Data;
-  state.eth1DataVotes.push(newEth1Data);
-  if (Eth1Data.equals(state.eth1Data, newEth1Data)) {
+  // Convert to tree-backed, this is what it looks like in the state,
+  // and hash-tree-root caching speeds things up a lot during the vote counting.
+  const newEth1DataCached = Eth1Data.tree.asTreeBacked(Eth1Data.tree.fromStructural(newEth1Data));
+  // Count it as vote
+  state.eth1DataVotes.push(newEth1DataCached);
+  // If there are not more than 50% votes, then we do not have to count to find a winner.
+  if (state.eth1DataVotes.length * 2 <= SLOTS_PER_ETH1_VOTING_PERIOD) {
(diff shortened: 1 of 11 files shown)


packages/lodestar-beacon-state-transition/src/fast/block/processEth1Data.ts
```diff
@@ -17,7 +17,8 @@ export function processEth1Data(
     return; // Nothing to do if the state already has this as eth1data (happens a lot after majority vote is in)
   }
   // TODO fast read-only iteration
-  const sameVotesCount = Array.from(state.eth1DataVotes).filter((e) => Eth1Data.equals(e, newEth1Data)).length;
+  const newEth1DataCached = Eth1Data.tree.createValue(newEth1Data);
+  const sameVotesCount = Array.from(state.eth1DataVotes).filter((e) => Eth1Data.equals(e, newEth1DataCached)).length;
```
twoeths: this is interesting! :)

twoeths (dismissed):

protolambda:
Updated my PR with more optimizations :sweat_smile:

packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts
```diff
@@ -55,12 +59,9 @@ export class EpochContext {
 
   public copy(): EpochContext {
     const ctx = new EpochContext(this.config);
-    // full copy of pubkeys, this can mutate
-    ctx.pubkey2index = new PubkeyIndexMap();
-    for(const entry of this.pubkey2index.entries()) {
-      ctx.pubkey2index.set(fromHexString(entry[0] as unknown as string), entry[1]);
-    }
-    ctx.index2pubkey = [...this.index2pubkey];
+    // warning: pubkey cache is not copied, it is shared, as eth1 is not expected to reorder validators.
```
mpetrunic, line 62: We copied it because state transition would modify pubkey cache for all previous epoch contexts.

Like you have context with 8 validators for epoch 1, and when transition to epoch 2 you copy this and for example add 3 and remove 2 validators, you will update epoch context for epoch 1 so we would have bad transition in case of reorg from epoch 1 block

protolambda: The pubkey cache won't reorg unless the deposit order reorgs, for now in phase 0 at least. So it's a case of only adding pubkeys, never overwriting/removing anything. The pubkeys that are not yet supposed to be in the cache w.r.t. the state can be ignored on the usage side by checking the validator index against the current registry size.
Now, the difficult part is avoiding bad reorgs, when e.g. an attacker manages to opportunistically fabricate a bad chain with censored votes.
If the cache is only used for finalized pubkeys it is much less of a problem, but more difficult.

Keeping the copy, but using an immutable map structure for cheap copies sounds like a clean better approach.

wemeetagain:
Ok, should have the latest optimizatoins in
Did the `tree.createValue` in a pre-validation step in our chain processing

mpetrunic (approved):

PR #1122, opened 2020-07-09
Fix chain restore

### LINKED ISSUE #1078: Finalized block when restarting chain
**Describe the bug**
When restarting change, chance is that we can't get finalized block in `block` database due to the change we have in #1077 

**Expected behavior**
Should find finalized block from either `block` or `archivedBlocks` database depending on if `ArchiveBlock` run already or not

twoeths:
resolves #1078 
+ Fix chain restore, add pre-finalized blocks to init forkchoice
+ Fix a small issue in DB
+ Order ArchiveBlock, ArchiveState
+ In eth1, don't need to await processing blocks

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -16,7 +16,11 @@ import {
   Uint64,
 } from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
-import {computeEpochAtSlot, computeForkDigest, EpochContext} from "@chainsafe/lodestar-beacon-state-transition";
+import {
+  computeEpochAtSlot,
+  computeForkDigest,
+  EpochContext,
+  ZERO_HASH} from "@chainsafe/lodestar-beacon-state-transition";
 import {ILogger} from "@chainsafe/lodestar-utils/lib/logger";
 import {intToBytes} from "@chainsafe/lodestar-utils";
 
@@ -211,8 +215,10 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
       !this.config.types.Root.equals(genesisBlock.stateRoot, storedGenesisBlock.message.stateRoot)) {
       throw new Error("A genesis state with different configuration was detected! Please clean the database.");
     }
+    const signedGenesisBlock = {message: genesisBlock, signature: EMPTY_SIGNATURE};
     await Promise.all([
-      this.db.block.add({message: genesisBlock, signature: EMPTY_SIGNATURE}),
+      this.db.block.add(signedGenesisBlock),
+      this.db.blockArchive.add(signedGenesisBlock),
       this.db.stateArchive.add(genesisState),
     ]);
     this.logger.info("Beacon chain initialized");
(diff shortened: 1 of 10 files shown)


packages/lodestar/src/chain/chain.ts
```diff
@@ -232,7 +234,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
 
   public async waitForBlockProcessed(blockRoot: Uint8Array): Promise<void> {
     await new Promise((resolve) => {
-      this.once("processedBlock", (signedBlock) => {
+      this.on("processedBlock", (signedBlock) => {
```
mpetrunic: unsubscribe on resolve

mpetrunic (changes requested):
Where do we set reprocess to true? Why do we need that?

Only thing I can think of is if we process block that isn't new head, we probably don't wan't to emit those events

twoeths:
> Where do we set reprocess to true? Why do we need that?

`reprocess` is true when we restart chain and reprocess blocks, at that time I got `forkChoice.getCanonicalBlockSummaryAtSlot(preSlot)` as undefined because initially we don't have "`justified`" node in forkchoice. I don't think we need to emit anything at that time.

mpetrunic:
> > Where do we set reprocess to true? Why do we need that?
> 
> `reprocess` is true when we restart chain and reprocess blocks, at that time I got `forkChoice.getCanonicalBlockSummaryAtSlot(preSlot)` as undefined because initially we don't have "`justified`" node in forkchoice. I don't think we need to emit anything at that time.

Wouldn't it make more sense to add justified and finalized node in forkchoice before processing?
I mean both justified and finalized node would be same (last block in block archive)

twoeths:
> Wouldn't it make more sense to add justified and finalized node in forkchoice before processing?
I mean both justified and finalized node would be same (last block in block archive)

I can add that. It's not the last block in block archive, it's the justified/finalized block of the last known state (the finalized/justified checkpoints of the state of finalized block)

twoeths:
In order to do that, I need to be able to get finalized block by root in #1107

packages/lodestar/src/db/api/beacon/repositories/abstract.ts
```diff
@@ -211,16 +211,16 @@ export abstract class Repository<I extends Id, T> {
       lt: _encodeKey(this.bucket + 1, Buffer.alloc(0)),
     };
     if (opts) {
-      if (opts.lt) {
+      if (opts.lt !== undefined) {
```
mpetrunic: This wouldn't catch opts.lt = null.

It's better to use opts.lt && Number.isNumber(opts.lt) if you are trying to catch 0 in opts. But since we don't use negative numbers 0 here is same as undefined, right?

packages/lodestar/test/unit/eth1/impl/ethers.test.ts
```diff
@@ -120,7 +121,8 @@ describe("Eth1Notifier", () => {
       logger,
     });
     await expect(
-      eth1.start()
+      // @ts-ignore
```
mpetrunic, line 124: why?

packages/lodestar/src/db/api/beacon/repositories/abstract.ts
twoeths: > It's better to use opts.lt && Number.isNumber(opts.lt) if you are trying to catch 0 in opts


yeah I got an issue in `gt: 0` case, will fix like that

packages/lodestar/test/unit/eth1/impl/ethers.test.ts
twoeths: `eth1.start()` does not await the `startProcessEth1Blocks` which is private so I have to do like this. The other way is to publish `startProcessEth1Blocks` just for test. Any ideas?

mpetrunic: ah, either leave like this or introduce some events like "eth1BlockProcessStarted". I'm fine either way :slightly_smiling_face:

packages/lodestar/src/db/api/beacon/repositories/abstract.ts
twoeths: Number.isInteger requires me to cast to number so I'd just compare to 0

mpetrunic (dismissed):

packages/lodestar/src/tasks/index.ts
```diff
@@ -69,8 +69,8 @@ export class TasksService implements IService {
   };
 
   private handleFinalizedCheckpointChores = async (finalized: BlockSummary, pruned: BlockSummary[]): Promise<void> => {
-    new ArchiveBlocksTask(this.config, {db: this.db, logger: this.logger}, finalized, pruned).run();
-    new ArchiveStatesTask(this.config, {db: this.db, logger: this.logger}, finalized, pruned).run();
+    await new ArchiveBlocksTask(this.config, {db: this.db, logger: this.logger}, finalized, pruned).run();
+    await new ArchiveStatesTask(this.config, {db: this.db, logger: this.logger}, finalized, pruned).run();
```
wemeetagain: can we do Promise.all here?

packages/lodestar/test/unit/eth1/impl/ethers.test.ts
```diff
@@ -61,6 +61,7 @@ describe("Eth1Notifier", () => {
     provider.getBlockNumber.resolves(currentBlockNumber);
 
     await eth1.start();
+    await Promise.resolve();
```
wemeetagain, line 64: a stray promise.resolve?

wemeetagain (changes requested):

packages/lodestar/src/tasks/index.ts
twoeths: I want to order it so that when we restart from last known state, we're sure finalized blocks stay in ArchiveBlock db already.
If we do Promise.all we're not sure about that if our node stops somewhere in the middle

packages/lodestar/test/unit/eth1/impl/ethers.test.ts
twoeths: remove it will call the test fail as `eth1` now does not await for `startProcessEth1Blocks`. This is like a hacky way but not sure how can we make it better.

packages/lodestar/src/tasks/index.ts
wemeetagain: If the node exits cleanly, I don't think it shouldn't matter. I think the node.stop would await any open promises.

wemeetagain (approved):

PR #1123, opened 2020-07-09
Avoid using db.stateCache.values()

### LINKED ISSUE #1118: Profiling: archiveStates is slow
While syncing Altona, CPU profiler shows significant time spent in `archiveStates`, specifically, because we call `stateCache.values`, which iterates thru all states/contexts and copies them. Even though this is relatively lightweight, it adds up.

Slow `stateCache.values` is a bottleneck.

avoid calling `stateCache.values` alltogether. Ideally, we should be able to use the information from the `forkChoice` (with some minor refactoring) to figure out which stateRoots to remove. Currently, it seems we immediately delete the old `forkChoice` data, and only delete the old `stateCache` afterwards as a "task".

Make copying/cloning cheaper (we should try to do this too)

twoeths:
resolves #1118 

+ We know that using `db.stateCache.values()` is expensive so just comment out that for now. The processing time is reduce from ~2500ms to ~200ms in my environment.

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -232,7 +232,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
 
   public async waitForBlockProcessed(blockRoot: Uint8Array): Promise<void> {
     await new Promise((resolve) => {
-      this.once("processedBlock", (signedBlock) => {
+      this.on("processedBlock", (signedBlock) => {
         const root = this.config.types.BeaconBlock.hashTreeRoot(signedBlock.message);
         if (this.config.types.Root.equals(root, blockRoot)) {
           resolve();
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -231,14 +231,17 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
   }
 
   public async waitForBlockProcessed(blockRoot: Uint8Array): Promise<void> {
+    let listener: (signedBlock: SignedBeaconBlock) => void;
     await new Promise((resolve) => {
-      this.on("processedBlock", (signedBlock) => {
+      listener = (signedBlock) => {
         const root = this.config.types.BeaconBlock.hashTreeRoot(signedBlock.message);
         if (this.config.types.Root.equals(root, blockRoot)) {
           resolve();
         }
-      });
+      };
+      this.on("processedBlock", listener);
     });
+    this.removeListener("processedBlock", listener);
   }
 
   /**
--- a/packages/lodestar/src/db/api/beacon/stateContextCache.ts
+++ b/packages/lodestar/src/db/api/beacon/stateContextCache.ts
@@ -1,6 +1,6 @@
 import {ByteVector, toHexString, TreeBacked} from "@chainsafe/ssz";
-import {BeaconState, Epoch, Slot} from "@chainsafe/lodestar-types";
-import {EpochContext, computeEpochAtSlot} from "@chainsafe/lodestar-beacon-state-transition";
+import {BeaconState} from "@chainsafe/lodestar-types";
+import {EpochContext} from "@chainsafe/lodestar-beacon-state-transition";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 
 export interface ITreeStateContext {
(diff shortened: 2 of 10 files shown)


wemeetagain:
I'd like to see about using the technique described in the issue.
Fork choice data mirrors the data in `db.stateCache` and `db.block`. The work of knowing "which to delete" is already done in `forkChoice.prune`. This would make it trivial to prune the cache:
```typescript
// archive state
forkChoice.on("prune", async (summaries: BlockSummary[]) => {
  await db.stateArchive.set(...);
  await db.stateCache.batchDelete(summaries.map(s => s.stateRoot));
});
// archive blocks
forkChoice.on("prune", async (summaries: BlockSummary[]) => {
  await db.blockArchive.set(...);
  await db.block.batchDelete(summaries.map(s => s.blockRoot));
});
```

mpetrunic:
I think fork choice prune isnt the same as cache prune aince fork choice keeps last finalized stuff.

twoeths:
just double check:
+ For fork choice, we keep finalized stuff, remove nodes < `finalized.slot`
+ For ArchiveState, we keep finalized state in `stateCache`, remove states < `finalized.slot`
+ For ArchiveBlock, we only use `finalized`, not `pruned` so it's just the same to the current implementation

so we're safe

packages/lodestar/src/chain/chain.ts
```diff
@@ -232,7 +232,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
 
   public async waitForBlockProcessed(blockRoot: Uint8Array): Promise<void> {
     await new Promise((resolve) => {
-      this.once("processedBlock", (signedBlock) => {
+      this.on("processedBlock", (signedBlock) => {
         const root = this.config.types.BeaconBlock.hashTreeRoot(signedBlock.message);
         if (this.config.types.Root.equals(root, blockRoot)) {
           resolve();
```
mpetrunic, line 239: you should probably unsubscribe processedBlock event at this point

packages/lodestar/src/db/api/beacon/stateContextCache.ts
```diff
@@ -42,9 +46,13 @@ export class StateContextCache {
...
-  public async values(): Promise<ITreeStateContext[]> {
-    return Object.values(this.cache).map(item => this.clone(item));
-  }
+  /**
+   * Should only use this with care as this is expensive.
+   * @param epoch
+   */
+  // public async values(): Promise<ITreeStateContext[]> {
+  //   return Object.values(this.cache).map(item => this.clone(item));
+  // }
```
mpetrunic: We could maybe call it `valuesUnsafe` or something though

twoeths:
I got this error after syncing up to block 4968
```
(node:28904) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'state' of null
    at BeaconChain.getHeadState (/Users/tuyennguyen/Projects/workshop/lodestar/packages/lodestar/lib/chain/chain.js:100:76)
    at async LocalClock.<anonymous> (/Users/tuyennguyen/Projects/workshop/lodestar/packages/lodestar/lib/sync/utils/attestation-collector.js:29:25)
```
need to investigate that ...

mpetrunic:
> I got this error after syncing up to block 4968
> 
> ```
> (node:28904) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'state' of null
>     at BeaconChain.getHeadState (/Users/tuyennguyen/Projects/workshop/lodestar/packages/lodestar/lib/chain/chain.js:100:76)
>     at async LocalClock.<anonymous> (/Users/tuyennguyen/Projects/workshop/lodestar/packages/lodestar/lib/sync/utils/attestation-collector.js:29:25)
> ```
> 
> need to investigate that ...

That's slot in which we have state transition split (we produce state with different root).
@wemeetagain already open PRs with fix for this in ssz.

packages/lodestar/src/tasks/tasks/archiveBlocks.ts
```diff
@@ -23,31 +25,31 @@ export class ArchiveBlocksTask implements ITask {
...
-    this.finalizedCheckpoint = finalizedCheckpoint;
+    this.finalized = finalized;
+    this.pruned = pruned;
   }
 
   /**
    * Only archive blocks on the same chain to the finalized checkpoint.
    */
   public async run(): Promise<void> {
+    this.logger.profile("Archieve Blocks");
```
wemeetagain: typo, the other logger messages here and in archiveState are typo'd too

wemeetagain (changes requested):
typos, this looks great tho!

wemeetagain (approved):

PR #1125, opened 2020-07-09
API: add /node/head endpoint, tweak /node/version endpoint

protolambda:
Working on Eth2stats support, and for this I need a few APIs.
- genesis time is already there
- version is there, but had an "undefined" in it. Defaulting the extra info to "dev" now
- head (or sometimes called chainhead) endpoint. Not standard yet, but this matches Lighthouse and Teku closely, so it's easy to adapt in eth2stats, and we'll very likely see something like this in the final API standard anyway.
- Looking at implementing a peers endpoint now. Think I'll add it to `/node` for now, and you can wait for the api standard to extend and move it to `/network`

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -10,13 +10,19 @@ import {
   Checkpoint,
   ENRForkID,
   ForkDigest,
+  HeadResponse,
   SignedBeaconBlock,
   Slot,
   Uint16,
   Uint64,
 } from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
-import {computeEpochAtSlot, computeForkDigest, EpochContext} from "@chainsafe/lodestar-beacon-state-transition";
+import {
+  computeEpochAtSlot,
+  computeForkDigest,
+  computeStartSlotAtEpoch,
+  EpochContext
+} from "@chainsafe/lodestar-beacon-state-transition";
 import {ILogger} from "@chainsafe/lodestar-utils/lib/logger";
 import {intToBytes} from "@chainsafe/lodestar-utils";
 
@@ -132,6 +138,18 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
 
   public async getFinalizedCheckpoint(): Promise<Checkpoint> {
     return this.forkChoice.getFinalized();
+  }
+
+  public async getHead(): Promise<HeadResponse> {
+    const head = this.forkChoice.head();
+    return {
+      headSlot: head.slot,
+      headBlockRoot: head.blockRoot,
+      finalizedSlot: computeStartSlotAtEpoch(this.config, head.finalizedCheckpoint.epoch),
+      finalizedBlockRoot: head.finalizedCheckpoint.root,
+      justifiedSlot: computeStartSlotAtEpoch(this.config, head.justifiedCheckpoint.epoch),
+      justifiedBlockRoot: head.justifiedCheckpoint.root,
+    };
   }
 
   public async start(): Promise<void> {
--- a/packages/lodestar-types/src/ssz/generators/api.ts
+++ b/packages/lodestar-types/src/ssz/generators/api.ts
@@ -59,3 +59,13 @@ export const ValidatorResponse = (ssz: IBeaconSSZTypes): ContainerType => new Co
   },
 });
 
+export const HeadResponse = (ssz: IBeaconSSZTypes): ContainerType => new ContainerType({
+  fields: {
+    headSlot: ssz.Slot,
+    headBlockRoot: ssz.Root,
+    finalizedSlot: ssz.Slot,
+    finalizedBlockRoot: ssz.Root,
+    justifiedSlot: ssz.Slot,
+    justifiedBlockRoot: ssz.Root
+  },
+});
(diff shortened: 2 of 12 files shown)


protolambda:
Eh, forgot to update the stubbed version of the API. Fixing that.

packages/lodestar/src/chain/chain.ts
```diff
@@ -134,6 +140,18 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
     return this.forkChoice.getFinalized();
   }
 
+  public async getHead(): Promise<HeadResponse> {
+    const head = this.forkChoice.head();
+    return {
+      headSlot: head.slot,
+      headBlockRoot: head.blockRoot,
+      finalizedSlot: computeStartSlotAtEpoch(this.config, head.finalizedCheckpoint.epoch),
```
mpetrunic, line 148: this would return incorrect value in case of skip slot on start of epoch

mpetrunic (commented):
I'm gonna deprecate current /node api in favor of standardized in next PR so not sure if we should merge this.

It probably makes more sense to add support for standardized api inside eth2stats instead of implementing non standardized api here.

> head (or sometimes called chainhead) endpoint. Not standard yet, but this matches Lighthouse and Teku closely, so it's easy to adapt in eth2stats, and we'll very likely see something like this in the final API standard anyway

I don't think it's gonna end up in standard as both LH and Teku are moving into standardized api where we have:
- https://ethereum.github.io/eth2.0-APIs/#/Beacon/getStateFinalityCheckpoints
- https://ethereum.github.io/eth2.0-APIs/#/Beacon/getBlockHeader

> Looking at implementing a peers endpoint now. Think I'll add it to /node for now, and you can wait for the api standard to extend and move it to /network

There is no /network in standardized api. I already have branch with /v1/node/peers and other endpoints implemented as per standard api:
https://ethereum.github.io/eth2.0-APIs/#/Node/getPeers

packages/lodestar/src/chain/chain.ts
protolambda: No I don't think it would, it is not specced to match the block root. Checkpoint epochs match the time of the checkpoint in addition to the block root, to point to skip slots, since those need to be checkpoint-ed too. Otherwise we would only be using the block root.

protolambda:
This is mostly just experimental to get eth2stats setup, I don't want lodestar to miss out on the Altona fun. If you want to wait for the standard API before implementing these endpoints (with the new api endpoint system, and those open API definitions), that's totally fine. But I'll be running this API for now, to get the data I need :).

protolambda:
Stub not updated again, whoops, I keep forgetting :sweat_smile:

mpetrunic:
> This is mostly just experimental to get eth2stats setup, I don't want lodestar to miss out on the Altona fun.

That's totally fine :smile: 

 > If you want to wait for the standard API before implementing these endpoints (with the new api endpoint system, and those open API definitions), that's totally fine. But I'll be running this API for now, to get the data I need :).

We already have standardized api or at least ones we need for eth2stats. It just makes more sense to use time to add standardized api support in eth2stats-client repo rather then adding deprecated endpoints here :slightly_smiling_face:

protolambda:
> It just makes more sense to use time to add standardized api support in eth2stats-client repo rather then adding deprecated endpoints here 

Agree to some extent. Not all clients are prioritizing these things the same way, so I imagine we will be stuck to the old APIs in eth2stats for at least another month, and rather not make big changes in the eth2stats client code, and just replicate the Teku/lighthouse api endpoints. Besides, the new open-api related typescript code looks more complicated, and I don't think I can get into that as fast.

Also, from the eth2stats perspective, in about two weeks the eth2stats server code should get open sourced (yay!). So then we can do a bigger proper overhaul, based on new standard API (throwing away the old things), and show more data on eth2stats with more ease.

mpetrunic:
> Agree to some extent. Not all clients are prioritizing these things the same way, so I imagine we will be stuck to the old APIs in eth2stats for at least another month, and rather not make big changes in the eth2stats client code, and just replicate the Teku/lighthouse api endpoints

It's easy to add new api support, I'll try to hack it over weekend to add standardized api support just like teku:
https://github.com/Alethio/eth2stats-client/blob/master/beacon/teku/teku.go

If I conform to existing expected data everything should work.

> Besides, the new open-api related typescript code looks more complicated, and I don't think I can get into that as fast.

It's same code, just registering differently:
https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/api/rest/routes/beacon/index.ts#L29

It provides better typesafety :blush: 
 
> Also, from the eth2stats perspective, in about two weeks the eth2stats server code should get open sourced (yay!). So then we can do a bigger proper overhaul, based on new standard API (throwing away the old things), and show more data on eth2stats with more ease.

We don't really need server, just to modify eth2stats client to know how to call standardized api :slightly_smiling_face:

protolambda:
Cool, if you have time to refactor to the standard API that is nice, and I'll help update the eth2stats client then. With these quick changes currently though, it's almost the same as lighthouse, so I'll have eth2stats support down in the next hour, I can't wait :sweat_smile: .

And regarding eth2stats: I see the standardized API + open source server more as a pivot point, where we can finally improve the dashboard (Alethio is great, but shuttering down direct support for eth2stats), while we are changing the data sources anyway.

mpetrunic:
I'm closing this since we have standard api for eth2stats now

PR #1129, opened 2020-07-10
genesis and skip slots case in finalized checkpoint check

protolambda:
Based on this testnet error:
```
2020-07-10 05:49:24  [SYNC]             warn: Received block parent root doesn't match our head head=0x7cb6a152924d493d6119407b732461d2addee6ec833b38fe169119d576db1dcd, headSlot=76069, blockParent=0x6f0f0920aa5d299df1245d925a2f2152c2550226c246ea24f169fcedc72ec700, blockSlot=76071
(node:201339) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'message' of null
    at BeaconReqRespHandler.shouldDisconnectOnStatus (/home/proto/projects/lodestar/packages/lodestar/lib/sync/reqResp/reqResp.js:214:130)
    at async BeaconReqRespHandler.onStatus (/home/proto/projects/lodestar/packages/lodestar/lib/sync/reqResp/reqResp.js:184:9)
    at async ReqResp.<anonymous> (/home/proto/projects/lodestar/packages/lodestar/lib/sync/reqResp/reqResp.js:74:18)
(Use `node --trace-warnings ...` to show where the warning was created)
```

As shared in discord:
> The problem is that the slot may be a skip slot
> And the block root may be from multiple epochs back even
> the epoch in the checkpoint is there to checkpoint the tail end of skip slots, even if there is no block

Changes:
- fix genesis case while we are at it
- cheap very common check against current finalized checkpoint first
- alternatively use head state for checking (mainnet config has a tree of 8000 or so roots to check against). This deals with skip-slots, it will return the most recent block for the given epoch.
- leave out the case for historical checking. There is no lookup for "latest finalized block root for epoch N", and looking up a block directly may return nil in case of a skip slot. To fix later, with DB changes to track finalized roots.

### DIFF
--- a/packages/lodestar/src/sync/reqResp/reqResp.ts
+++ b/packages/lodestar/src/sync/reqResp/reqResp.ts
@@ -21,7 +21,7 @@ import {INetwork} from "../../network";
 import {ILogger} from "@chainsafe/lodestar-utils/lib/logger";
 import {IReqRespHandler} from "./interface";
 import {IReputationStore} from "../IReputation";
-import {computeStartSlotAtEpoch, GENESIS_SLOT} from "@chainsafe/lodestar-beacon-state-transition";
+import {computeStartSlotAtEpoch, getBlockRoot, GENESIS_SLOT} from "@chainsafe/lodestar-beacon-state-transition";
 import {toHexString} from "@chainsafe/ssz";
 import {RpcError} from "../../network/error";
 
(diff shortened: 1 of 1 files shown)


wemeetagain (approved):
Looks good. I think it would be great to store finalized checkpoints to DB

PR #1130, opened 2020-07-10
Fix req/resp TTFB timeout

### LINKED ISSUE #1128: Altona: incorrect logic in shouldDisconnectOnStatus
**Describe the bug**
This issue is found by proto after he seems to finish initial sync
```
2020-07-10 05:49:23  [CHAIN]            info: Processed new chain head newChainHeadRoot=0x6f0f0920aa5d299df1245d925a2f2152c2550226c246ea24f169fcedc72ec700, slot=76069, epoch=2377
2020-07-10 05:49:24  [SYNC]             info: Imported blocks for slots: 76071,76072,76073,76074,76076,76077,76078,76079,76082,76083,76085,76086,76087,76088,76089,76091,76092,76093,76094,76095,76096,76097,76098,76099,76100,76102,76103,76104,76106,76107,76109,76110,76111,76112,76113,76115,76116,76117,76118,76119,76120,76122,76123,76124,76125,76127,76128,76129 
2020-07-10 05:49:24  [SYNC]             warn: Received block parent root doesn't match our head head=0x7cb6a152924d493d6119407b732461d2addee6ec833b38fe169119d576db1dcd, headSlot=76069, blockParent=0x6f0f0920aa5d299df1245d925a2f2152c2550226c246ea24f169fcedc72ec700, blockSlot=76071
(node:201339) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'message' of null
    at BeaconReqRespHandler.shouldDisconnectOnStatus (/home/proto/projects/lodestar/packages/lodestar/lib/sync/reqResp/reqResp.js:214:130)
    at async BeaconReqRespHandler.onStatus (/home/proto/projects/lodestar/packages/lodestar/lib/sync/reqResp/reqResp.js:184:9)
    at async ReqResp.<anonymous> (/home/proto/projects/lodestar/packages/lodestar/lib/sync/reqResp/reqResp.js:74:18)
```
More at https://gist.github.com/protolambda/c999779f766a0de1df1a6bb313081c47

**Expected behavior**
+ It seems the epoch misses the 1st slot
+ So we should find a way to get correct root of finalized checkpoint given an epoch

twoeths:
resolves #1128 

+ One timeouted response logged after 1h. We should handle TTFB_TIMEOUT in `sendRequestStream` instead of `sendRequest`
+ In `getBlockRange`, we keep sending to a disconnected peer, it's also the peer id of a timeout response earlier so I handled bad peer inside that

### DIFF
--- a/packages/lodestar/src/network/util.ts
+++ b/packages/lodestar/src/network/util.ts
@@ -61,23 +61,27 @@ export function isRequestSingleChunk(method: Method): boolean {
   return Methods[method].responseType === MethodResponseType.SingleResponse;
 }
 
-export function eth2ResponseTimer<T>(): (source: AsyncIterable<T>) => AsyncGenerator<T> {
-  return (source) => {
-    return (async function*() {
-      const controller = new AbortController();
-      let responseTimer = setTimeout(() => controller.abort(), TTFB_TIMEOUT);
-      const renewTimer = (): void => {
-        clearTimeout(responseTimer);
-        responseTimer = setTimeout(() => controller.abort(), RESP_TIMEOUT);
-      };
-      const cancelTimer = (): void => {
-        clearTimeout(responseTimer);
-      };
-      for await(const item of abortSource(source, controller.signal, {abortMessage: "response timeout"})) {
-        renewTimer();
-        yield item;
-      }
-      cancelTimer();
-    })();
+export function initResponseTimer<T>(): (() => ((source: AsyncIterable<T>) => AsyncGenerator<T>)) {
+  const controller = new AbortController();
+  let responseTimer = setTimeout(() => controller.abort(), TTFB_TIMEOUT);
+  const renewTimer = (): void => {
+    clearTimeout(responseTimer);
+    responseTimer = setTimeout(() => controller.abort(), RESP_TIMEOUT);
   };
+  const cancelTimer = (): void => {
+    clearTimeout(responseTimer);
+  };
+  const checkTimeout = (): ((source: AsyncIterable<T>) => AsyncGenerator<T>) => {
+    return (source) => {
+      return (async function*() {
+        const abortOpt = {abortMessage: "response timeout"};
+        for await(const item of abortSource(source, controller.signal, abortOpt)) {
+          renewTimer();
+          yield item;
+        }
+        cancelTimer();
+      })();
+    };
+  };
+  return checkTimeout;
 }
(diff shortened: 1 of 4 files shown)


packages/lodestar/src/sync/utils/blocks.ts
```diff
@@ -83,12 +84,15 @@ export async function getBlockRange(
...
           //if failed to obtain blocks, try in next round on another peer
-          return chunk;
+          return {chunk, peer};
         }
       })
-    )).filter((chunk) => chunk !== null);
+    )).filter((chunkPeer) => chunkPeer !== null);
+    chunks = chunkPeers.map((chunkPeer) => chunkPeer.chunk);
+    const badPeers = chunkPeers.map((chunkPeer) => chunkPeer.peer);
+    peers = peers.filter((peer) => !badPeers.includes(peer));
```
mpetrunic: I don't think we should handle this here, those thing should be sorted out once we get peer scoring.
And we probably shouldn't use peers with closed connection (that should be filtered when obtaining peer list)

packages/lodestar/src/network/util.ts
```diff
@@ -61,23 +61,27 @@ export function isRequestSingleChunk(method: Method): boolean {
...
-}
+  const cancelTimer = (): void => {
+    clearTimeout(responseTimer);
+  };
+  const checkTimeout = (): ((source: AsyncIterable<T>) => AsyncGenerator<T>) => {
+    return (source) => {
+      return (async function*() {
+        const abortOpt = {abortMessage: "response timeout"};
+        for await(const item of abortSource(source, controller.signal, abortOpt)) {
+          renewTimer();
```
wemeetagain, line 79: It seems that the RESP_TIMEOUT is 'The maximum time for complete response transfer.'
Shouldn't be reset, but rather tracked across the whole response

wemeetagain (changes requested):

packages/lodestar/src/network/util.ts
twoeths: I think RESP_TIMEOUT is for per response_chunk as in the p2p spec: `The requester MUST wait a maximum of TTFB_TIMEOUT for the first response byte to arrive (time to first byte—or TTFB—timeout). On that happening, the requester allows a further RESP_TIMEOUT for each subsequent response_chunk received.`

twoeths: I also see `The maximum time for complete response transfer.`, this is misleading

wemeetagain: Gotcha, you wanna make a PR to the spec to clarify?

wemeetagain (approved):

packages/lodestar/src/network/util.ts
twoeths: probably not at this time, it's too small and all clients understood it I think :)

PR #1131, opened 2020-07-10
Validate attestation

### LINKED ISSUE #1088: Validate attestations received in API
**Describe the bug**

Hard to tell, we sometimes produce invalid attestation, might be that we try to include it in block too soon.
IF we fail to process block with that attestation it gets included in every following block thus breaks chain with following errors:
```
@chainsafe/lodestar: 2020-07-03 17:42:42  []                error: Failed to produce block for slot 17 Error: Attestation is not valid
@chainsafe/lodestar: Error: Attestation is not valid
```

**Expected behavior**

We should validate attestation when submitted trough api.

**Steps to Reproduce**

**Screenshots**

 - OS: 
 - Version: 
 - Branch: 
 - Commit hash:

mpetrunic:
We should stop seeing random e2e tests (still happens sometimes because of gossipsub starting/not starting)

Features:
- validate attestations received on api

Bugfixes:
- prevent double attestation production and publishing
- correctly use fast state transaition in attestation and block production
- bring state to correct slot before block and attestation production/validation
- sortout bugs where validators would mess up timers and notify on new slots before it should actually happen (in future we will handle this with stream endpoint)

resolves #1088


packages/lodestar/src/util/validation/attestation.ts
```diff
@@ -0,0 +1,75 @@
...
+  );
+    // each attestation has only 1 validator index
+  const existingValidatorIndexes = existingAttestations.map(
+    item => getAttestingIndices(config, state, item.data, item.aggregationBits)[0]);
+    // attestation is unaggregated attestation as validated above
+  const validatorIndex = getAttestingIndices(config, state, attestation.data, attestation.aggregationBits)[0];
+  return !existingValidatorIndexes.includes(validatorIndex);
+}
+
+export async function validateAttestation(
```
wemeetagain, line 69: this still uses slow state transition

packages/lodestar/src/util/validation/attestation.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export function isUnaggregatedAttestation(
+  config: IBeaconConfig, state: BeaconState, epochCtx: EpochContext, attestation: Attestation
+): boolean {
+  if (state.slot < attestation.data.slot) {
+    processSlots(epochCtx, state, attestation.data.slot);
+  }
+  // Make sure this is unaggregated attestation
+  return getAttestingIndices(config, state, attestation.data, attestation.aggregationBits).length === 1;
```
wemeetagain: `getAttestingIndices` should be split in two functions, one looks like:
`getAttestingIndicesFromCommittee(committee: ValidatorIndex[], bits: BitList): ValidatorIndex`
So we can use `epochCtx.getBeaconCommittee` instead of slow committee calculation here.

packages/lodestar/src/util/validation/attestation.ts
```diff
@@ -0,0 +1,76 @@
...
+    // each attestation has only 1 validator index
+  const existingValidatorIndexes = existingAttestations.map(
+    item => getAttestingIndices(config, state, item.data, item.aggregationBits)[0]);
+    // attestation is unaggregated attestation as validated above
+  const validatorIndex = getAttestingIndices(config, state, attestation.data, attestation.aggregationBits)[0];
+  return !existingValidatorIndexes.includes(validatorIndex);
+}
+
+export async function validateAttestation(
+  config: IBeaconConfig, db: IBeaconDb, state: BeaconState, epochCtx: EpochContext, attestation: Attestation
```
wemeetagain: maybe epochCtx should come before state, just for consistency?

wemeetagain (changes requested):

packages/lodestar-validator/src/api/abstract.ts
```diff
@@ -118,4 +124,12 @@ export abstract class AbstractApiClient
     }
   };
 
+  /**
+   * Returns milis till next slot
+   */
+  private getDiffTillNextSlot(): number {
+    const diffInSeconds = Math.floor((Date.now() / 1000) - this.genesisTime);
```
twoeths, line 133: this function will likely return seconds to next slot * 1000. Should we improve it to calculate more precisely to return milli seconds to next slot?

packages/lodestar/src/network/gossip/validator.ts
```diff
@@ -100,39 +99,22 @@ export class GossipMessageValidator implements IGossipMessageValidator {
...
-    const validatorIndex = getAttestingIndices(this.config, state, attestationData, attestation.aggregationBits)[0];
-    if (existingValidatorIndexes.includes(validatorIndex)) {
-      return false;
-    }
-    const blockRoot = attestationData.beaconBlockRoot.valueOf() as Uint8Array;
-    if (!this.chain.forkChoice.hasBlock(blockRoot) || await this.db.badBlock.has(blockRoot)) {
+    try {
+      await validateAttestation(this.config, this.db, state, epochCtx, attestation);
+      return true;
+    } catch (e) {
```
twoeths, line 115: probably log the error here

packages/lodestar-validator/src/api/abstract.ts
mpetrunic: I find it hard to implement simple algo for this. That said,I don't think we care about milis there.

packages/lodestar/src/util/validation/attestation.ts
```diff
@@ -0,0 +1,89 @@
...
+): Promise<boolean> {
+  if (state.slot < attestation.data.slot) {
+    processSlots(epochCtx, state, attestation.data.slot);
+  }
+  const existingAttestations = await db.attestation.geAttestationsByTargetEpoch(
+    attestation.data.target.epoch
+  );
+    // each attestation has only 1 validator index
+  const existingValidatorIndexes = existingAttestations.map(
+    item => getAttestingIndicesFromCommittee(
```
wemeetagain, line 53: This might be more ergonomic as a method on epochCtx. Same with getCommitteeAssignment. (eg: `epochCtx.getAttestingIndices` and `epochCtx.getCommitteeAssignment`)

Maybe we just make an issue for it for now?

mpetrunic: Yes pls, since there is awfully lot of stubbing required

wemeetagain (approved):

PR #1136, opened 2020-07-10
beacon params as an ssz type

3xtr4t3rr3str14l:
-fix for #1108 by encapsulating the beacon params as an ssz type to be used as a yargs option in the beacon init process
-update mainnet and minimal yaml files in lodestar-params to latest spec (and update the beacon chain interface and beacon params files accordingly)


packages/lodestar-params/src/utils.ts
```diff
@@ -1,29 +1,90 @@
...
+    HYSTERESIS_DOWNWARD_MULTIPLIER: new NumberUintType({byteLength: 8}),
+    HYSTERESIS_UPWARD_MULTIPLIER: new NumberUintType({byteLength: 8}),
+    MIN_GENESIS_ACTIVE_VALIDATOR_COUNT: new NumberUintType({byteLength: 8}),
+    TARGET_AGGREGATORS_PER_COMMITTEE: new NumberUintType({byteLength: 8}),
+    RANDOM_SUBNETS_PER_VALIDATOR: new NumberUintType({byteLength: 8}),
+    EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION: new NumberUintType({byteLength: 8}),
+    SECONDS_PER_ETH1_BLOCK: new NumberUintType({byteLength: 8}),
+
+    // Deposit contract
+    DEPOSIT_CONTRACT_ADDRESS: new BitListType({limit: Number.MAX_SAFE_INTEGER - 1}),
```
wemeetagain: Should be ByteVectorType({length: 32})

packages/lodestar-params/src/utils.ts
```diff
@@ -1,29 +1,90 @@
...
+
+    // Gwei Values
+    MIN_DEPOSIT_AMOUNT: new BigIntUintType({byteLength: Number.MAX_SAFE_INTEGER - 1}),
+    MAX_EFFECTIVE_BALANCE: new BigIntUintType({byteLength: 32}),
+    EJECTION_BALANCE: new BigIntUintType({byteLength: 32}),
+    EFFECTIVE_BALANCE_INCREMENT: new BigIntUintType({byteLength: 32}),
```
wemeetagain: byteLength should be 8 for these

packages/lodestar-params/src/utils.ts
```diff
@@ -1,29 +1,90 @@
...
-import * as constants from "./constants";
+import {ContainerType, BigIntUintType, NumberUintType, BitVectorType, BitListType, Json} from "@chainsafe/ssz";
 
-export function createIBeaconParams(params: Record<string, unknown>): Partial<IBeaconParams> {
-  return convertTypes(params, typeMap);
-}
+const beaconParamsType = new ContainerType({
+  fields: {
+    // Misc
+    MAX_COMMITTEES_PER_SLOT: new NumberUintType({byteLength: 8}),
```
wemeetagain: You should factor each unique type out above the `beaconParamsType` declaration:
```typescript
const Number64 = new NumberUintType({byteLength: 8});
...

const beaconParamsType = new ContainerType({
  fields: {
    MAX_COMMITTEES_PER_SLOT: Number64,
    ...
```

packages/lodestar-params/src/utils.ts
```diff
@@ -1,29 +1,90 @@
...
+    // Gwei Values
+    MIN_DEPOSIT_AMOUNT: new BigIntUintType({byteLength: Number.MAX_SAFE_INTEGER - 1}),
+    MAX_EFFECTIVE_BALANCE: new BigIntUintType({byteLength: 32}),
+    EJECTION_BALANCE: new BigIntUintType({byteLength: 32}),
+    EFFECTIVE_BALANCE_INCREMENT: new BigIntUintType({byteLength: 32}),
+
+    // Initial values
+    SAFE_SLOTS_TO_UPDATE_JUSTIFIED: new NumberUintType({byteLength: 8}),
+    BLS_WITHDRAWAL_PREFIX: new BitVectorType({length: 8}),
+    GENESIS_FORK_VERSION: new BitVectorType({length: 32}),
```
wemeetagain: should be ByteVectorType({length: 4})

packages/lodestar-params/src/utils.ts
```diff
@@ -1,29 +1,90 @@
...
-      if(typeMap[k]) {
-        params[k] = typeMap[k](params[k] as string);
-      } else {
-        params[k] = Number(params[k]);
-      }
+export function createIBeaconParams(input: Record<string, unknown>): Partial<IBeaconParams> {
+  const params: Partial<IBeaconParams> = {};
+  Object.entries(beaconParamsType.fields).forEach(([fieldName, fieldType]) => {
+    if (input[fieldName]) {
+      (params as Record<string, unknown>)[fieldName] = fieldType.fromJson(input[fieldName] as Json) as Json;
```
wemeetagain:
```suggestion
      (params as Record<string, unknown>)[fieldName] = fieldType.fromJson(input[fieldName] as Json) as unknown;
```

packages/lodestar-params/src/utils.ts
```diff
@@ -1,29 +1,90 @@
 import {FAILSAFE_SCHEMA, Schema, Type} from "js-yaml";
-import {TypeMap, typeMap} from "./types";
 import {IBeaconParams} from "./interface";
-import * as constants from "./constants";
+import {ContainerType, BigIntUintType, NumberUintType, BitVectorType, BitListType, Json} from "@chainsafe/ssz";
 
-export function createIBeaconParams(params: Record<string, unknown>): Partial<IBeaconParams> {
-  return convertTypes(params, typeMap);
-}
+const beaconParamsType = new ContainerType({
```
wemeetagain:
```suggestion
export const BeaconParams = new ContainerType<IBeaconParams>({
```
Can you also export this type from src/index.ts?

wemeetagain: Can you move the ssz type to its own file? eg: src/sszType.ts

wemeetagain (changes requested):

packages/lodestar-params/src/beaconParams.ts
```diff
@@ -0,0 +1,86 @@
+import {IBeaconParams} from "./interface";
+import {
+  ContainerType, BigIntUintType, NumberUintType, BitVectorType, ByteVectorType,
+} from "@chainsafe/ssz";
+
+const Number64 = new NumberUintType({byteLength: 8});
+const BitVector8 = new BitVectorType({length: 8});
+const ByteVector32 = new ByteVectorType({length: 4});
```
wemeetagain: ByteVector32 should be length 32

This number is different than in number types.
We usually discuss numbers in bitlength (eg: uint64 is 64 bits, 8 bytes)
Bytevectors usually discussed in bytelength (eg: bytes32 is 32 bytes)

packages/lodestar-params/src/beaconParams.ts
```diff
@@ -0,0 +1,86 @@
...
+    HYSTERESIS_DOWNWARD_MULTIPLIER: Number64,
+    HYSTERESIS_UPWARD_MULTIPLIER: Number64,
+    MIN_GENESIS_ACTIVE_VALIDATOR_COUNT: Number64,
+    TARGET_AGGREGATORS_PER_COMMITTEE: Number64,
+    RANDOM_SUBNETS_PER_VALIDATOR: Number64,
+    EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION: Number64,
+    SECONDS_PER_ETH1_BLOCK: Number64,
+
+    // Deposit contract
+    DEPOSIT_CONTRACT_ADDRESS: ByteVector256,
```
wemeetagain: should be bytevector32

packages/lodestar-params/src/beaconParams.ts
```diff
@@ -0,0 +1,86 @@
...
+    // Gwei Values
+    MIN_DEPOSIT_AMOUNT: BigInt64,
+    MAX_EFFECTIVE_BALANCE: BigInt64,
+    EJECTION_BALANCE: BigInt64,
+    EFFECTIVE_BALANCE_INCREMENT: BigInt64,
+
+    // Initial values
+    SAFE_SLOTS_TO_UPDATE_JUSTIFIED: Number64,
+    BLS_WITHDRAWAL_PREFIX: BitVector8,
+    GENESIS_FORK_VERSION: ByteVector32,
```
wemeetagain: should be bytevector4

packages/lodestar-params/src/beaconParams.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    // Gwei Values
+    MIN_DEPOSIT_AMOUNT: BigInt64,
+    MAX_EFFECTIVE_BALANCE: BigInt64,
+    EJECTION_BALANCE: BigInt64,
+    EFFECTIVE_BALANCE_INCREMENT: BigInt64,
+
+    // Initial values
+    SAFE_SLOTS_TO_UPDATE_JUSTIFIED: Number64,
+    BLS_WITHDRAWAL_PREFIX: BitVector8,
```
wemeetagain: should be bytevector1

packages/lodestar-params/src/beaconParams.ts
```diff
@@ -0,0 +1,86 @@
...
+    EPOCHS_PER_SLASHINGS_VECTOR: Number64,
+    HISTORICAL_ROOTS_LIMIT: Number64,
+    VALIDATOR_REGISTRY_LIMIT: Number64,
+
+    // Reward and penalty quotients
+    BASE_REWARD_FACTOR: Number64,
+    BASE_REWARDS_PER_EPOCH: Number64,
+    WHISTLEBLOWER_REWARD_QUOTIENT: Number64,
+    PROPOSER_REWARD_QUOTIENT: Number64,
+    INACTIVITY_PENALTY_QUOTIENT: BigInt256,
```
wemeetagain: should be bigint64

packages/lodestar-params/src/presets/mainnet.yaml
```diff
@@ -52,7 +52,7 @@ SECONDS_PER_ETH1_BLOCK: 14
 # Deposit contract
 # ---------------------------------------------------------------
 # **TBD**
-DEPOSIT_CONTRACT_ADDRESS: 0x1234567890123456789012345678901234567890
+DEPOSIT_CONTRACT_ADDRESS: 0x1234567890123456789012345678901234567890123456789012345678901234
```
wemeetagain: this shouldn't be changed

packages/lodestar-params/src/beaconParams.ts
```diff
@@ -0,0 +1,86 @@
+import {IBeaconParams} from "./interface";
+import {
+  ContainerType, BigIntUintType, NumberUintType, BitVectorType, ByteVectorType,
+} from "@chainsafe/ssz";
+
+const Number64 = new NumberUintType({byteLength: 8});
+const BitVector8 = new BitVectorType({length: 8});
```
wemeetagain: we don't need this type

packages/lodestar-params/src/beaconParams.ts
```diff
@@ -0,0 +1,86 @@
...
+const ByteVector32 = new ByteVectorType({length: 4});
+const ByteVector256 = new ByteVectorType({length: 32});
```
wemeetagain: we don't need this type

packages/lodestar-params/src/beaconParams.ts
```diff
@@ -0,0 +1,86 @@
...
+const BigInt64 = new BigIntUintType({byteLength: 8});
+const BigInt256 = new BigIntUintType({byteLength: 32});
```
wemeetagain: we don't need this type

wemeetagain (changes requested):

packages/lodestar-params/package.json
```diff
@@ -49,6 +49,7 @@
     "mocha": "^6.2.2"
   },
   "dependencies": {
+    "@chainsafe/ssz": "^0.6.9",
```
mpetrunic, line 52: @wemeetagain Is this circular dep. Doesn't ssz depend on params?

mpetrunic: Maybe we should move this to lodestar-types?

wemeetagain: This shouldn't be a circular dependency, we will have:
- ssz -> params
- ssz -> types
  params -> types

The one tiny wrinkle is that we will have defined a few 'extra' ssz types, Number64, BigInt64, etc. As they are both needed here and in types. This is really really small, and likely not worth resolving fully. (or we could resolve it by creating these as 'well-known' types/singletons in the ssz package)

We can't move the `BeaconParams` ssz type to the types package because we want to parse the minimal/mainnet yaml here.

I think we can just treat this a a special type, we likely don't need to include it in our `IBeaconSSZTypes` interface, but just let it exist independently. It won't rely on the config to exist, and likely we can/should just let it exist as a singleton, here.

packages/lodestar-params/src/beaconParams.ts
3xtr4t3rr3str14l: changed to ByteVector20 as per latest spec config

packages/lodestar-params/src/interface.ts
```diff
@@ -3,6 +3,16 @@
  */
 
 export interface IBeaconParams {
+
```
wemeetagain, line 6: Can you order these fields in the same order as in the spec yaml files?
Also making sure its in the same order in the ssz type definition.

3xtr4t3rr3str14l: done!

wemeetagain (approved):

Issue #961, opened 2020-05-25
Bump gossipsub to v1.1 and note extended validator enums for gossip conditions 


ColinSchwarz:
https://github.com/ethereum/eth2.0-specs/pull/1796

https://github.com/ethereum/eth2.0-specs/pull/1828

dapplion:
@wemeetagain Is this resolved?

twoeths:
@dapplion we mostly implement the logic in the spec but for old gossip version.
Once we release gossip 1.1 and decide to use it in lodestar, there is a small modification in lodestar to make it work so let's keep this issue open.

twoeths:
should be done in #1206

PR #1140, opened 2020-07-11
Gossip Extended Validator

twoeths:
Part of #961 
+ Support Extended Validator in the current version of gossip. Once Gossip 1.1 is in, we just need to have a small change to remove `_processTopicValidatorResult`
+ Remove the check for parentRoot in `isValidIncomingBlock` as it's not part of the spec. If a block comes far away in the future, we have a check for it already.

### DIFF
--- a/packages/lodestar/src/network/gossip/gossipsub.ts
+++ b/packages/lodestar/src/network/gossip/gossipsub.ts
@@ -1,6 +1,6 @@
 import PeerId from "peer-id";
 import Gossipsub from "libp2p-gossipsub";
-import {Registrar} from "libp2p-gossipsub/src/peer";
+import {Registrar, Peer} from "libp2p-gossipsub/src/peer";
 import {Message} from "libp2p-gossipsub/src/message";
 import {Type} from "@chainsafe/ssz";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
@@ -16,7 +16,7 @@ import {
   normalizeInRpcMessage,
   topicToGossipEvent
 } from "./utils";
-import {GossipEvent} from "./constants";
+import {GossipEvent, ExtendedValidatorResult} from "./constants";
 import {GOSSIP_MAX_SIZE} from "../../constants";
 import {getTopicEncoding, GossipEncoding} from "./encoding";
 
@@ -65,7 +65,7 @@ export class LodestarGossipsub extends Gossipsub {
     }
   }
 
-  public async validate(rawMessage: Message): Promise<boolean> {
+  public async validate(rawMessage: Message, peer: Peer): Promise<boolean> {
     const message: ILodestarGossipMessage = normalizeInRpcMessage(rawMessage);
     assert.true(Boolean(message.topicIDs), "topicIds is not defined");
     assert.equal(message.topicIDs.length, 1, "topicIds array must contain one item");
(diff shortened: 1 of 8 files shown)


packages/lodestar/src/network/gossip/validator.ts
```diff
@@ -56,66 +57,68 @@ export class GossipMessageValidator implements IGossipMessageValidator {
...
     if (slot <= computeStartSlotAtEpoch(this.config, finalizedCheckpoint.epoch)) {
-      return false;
+      return ExtendedValidatorResult.ignore;
     }
 
     const parentBlock = await this.db.block.get(signedBlock.message.parentRoot.valueOf() as Uint8Array);
-    if(!parentBlock) {
-      return false;
-    }
-    const {state, epochCtx} = await this.db.stateCache.get(parentBlock.message.stateRoot);
```
twoeths, line 70: need confirmation

wemeetagain: I think we start doing validations against the head state, then, second-to-last rule (and last rule) is the check against the parent block.

twoeths: ok so I use the head state here.

I remove the check of `parentBlock` because:
+ it is not part of the spec
+ it causes the issue of not being able to sync to head after initial sync

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -76,23 +76,38 @@ export class LodestarGossipsub extends Gossipsub {
...
       const objSubnet = this.deserializeGossipMessage(topic, message);
       transformedObj = objSubnet.object;
-      isValid = await validatorFn(transformedObj, objSubnet.subnet);
+      validationResult = await validatorFn(transformedObj, objSubnet.subnet);
     } catch (err) {
-      isValid = false;
+      validationResult = ExtendedValidatorResult.reject;
       this.logger.error(`Cannot validate message from ${message.from}, topic ${message.topicIDs}, error: ${err}`);
     }
+    const isValid = this._processTopicValidatorResult(topic, peer, message, validationResult);
```
mpetrunic, line 90: If it's ignore, it will not be propagated to chain?

 I think you should store validation result in transformed object if it's accept|ignore and afterm,  prevent rebroadcast if it's status is "ignore".
That said, not sure how to achieve this without modifying:
https://github.com/ChainSafe/js-libp2p-gossipsub/blob/master/ts/index.ts#L223

mpetrunic: To summarize:
- accept - emitSelf and send to peers
- ignore - emitSelf only

mpetrunic (dismissed):

packages/lodestar/src/network/gossip/gossipsub.ts
twoeths: > If it's ignore, it will not be propagated to chain?
no it'll not propagate for reject & ignore, only propagate for accept.

My understanding is we treat ignore & reject the same in lodestar as invalid messages. The only difference is to trigger a penalty or not and it should be taken care by gossip-js in the next version of `_processTopicValidatorResult`.
The other thing to consider is to implement optional part like `a client MAY queue future blocks for processing at the appropriate slot` in p2p spec, we'll consider this later.
So I don't see we have to modify `emitSelf` and `transformed object`, please help clarify @wemeetagain @mpetrunic

mpetrunic: > The other thing to consider is to implement optional part like a client MAY queue future blocks for processing at the appropriate slot in p2p spec, we'll consider this later.

But we have both attestation and block queue implemented and we kinda need gossip to propagate future blocks and attestation

twoeths: I think if block is too far in the future then it should fail, otherwise it's covered by the spec by `MAXIMUM_GOSSIP_CLOCK_DISPARITY`. Probably what we want is the block in `current slot`, not future block.

mpetrunic: We wan't to queue at least 10ish blocks in future (gossiped blocks). So that once we reach head that we determined from current peer statuses, we can just use those queued blocks instead of sending and waiting for exchanging 20 status requests and setting new target head. 

Currently without queue of future blocks we can never reach head because once we exchange all status requests with our peers, head is already further away.

wemeetagain: > The other thing to consider is to implement optional part like a client MAY queue future blocks for processing at the appropriate slot in p2p spec, we'll consider this later.

Yeah, I think we can/should approach this in a separate PR, keep this one super straightforward.

There are a few MAYs that we will want to implement:
- a client MAY queue future blocks for processing at the appropriate slot
- If the proposer_index cannot immediately be verified against the expected shuffling, the block MAY be queued for later processing while proposers for the block's branch are calculated
- a client MAY queue blocks for processing once the parent block is retrieved

> My understanding is we treat ignore & reject the same in lodestar as invalid messages. The only difference is to trigger a penalty or not and it should be taken care by gossip-js in the next version of _processTopicValidatorResult.

This is indeed the default behavior of gossipsub. We'd want to extend this PR in some way to emit / queue blocks in those scenarios. I think it will be easier to tackle this in a separate PR, as there seems to be several possible solutions / directions to making this work.
My immediate thought is we can just queue these blocks directly from the validation function, but ideally we would use the same handling code as in the accept case. This may require additional thinking / refactoring of js-libp2p-gossipsub depending on the solution.

wemeetagain (dismissed):

wemeetagain:
gossip validator unit tests failing

wemeetagain (approved):

PR #1142, opened 2020-07-12
Add build:watch script

### LINKED ISSUE #1089: monorepo build:watch
It's frustrating when eg: Running the cli from `lodestar-cli`, after updating code in `lodestar`, we must run `yarn build` from the lodestar package, for `lodestar-cli` to pick up the changes.

The typescript source is not used by other downstream monorepo packages, only the transpiled javascript is used.

Maybe we can have something along the lines of `babel --watch && tsc --watch` running in each monorepo package, which will pick up any typescript file changes and automatically update the transpiled JS. This way, we won't need to be constantly be manually re-building each package.
If we can package this into something that we can run via `lerna`, we can get auto-building for the whole monorepo at once, very nice.

write a custom thing with https://github.com/paulmillr/chokidar with a nice glob that targets only what we need

**Additional context**
https://babeljs.io/docs/en/babel-cli#compile-files
https://www.typescriptlang.org/docs/handbook/configuring-watch.html

dapplion:
Fixes https://github.com/ChainSafe/lodestar/issues/1089.

- Adds build watch scripts to all packages. Devs can target the packages in development with lerna's scope.
- Otherwise, it also adds a concurrent script in the root package.json so everything is watched at once. Can be heavy on memory ~4GB on my laptop.


packages/lodestar-utils/src/assert.ts
```diff
@@ -20,6 +20,9 @@ export const assert = {
     }
   },
 
+  hello: 2,
+  bye: "a",
```
mpetrunic: eh? :smile:

mpetrunic (changes requested):
LGTM just some junk left

packages/lodestar-utils/src/assert.ts
dapplion: Thanks for catching that!

packages/lodestar-cli/src/index.ts
```diff
@@ -7,6 +7,10 @@ import {validatorCommandModule} from "./cmds/validator";
 import {globalOptions} from "./options";
 import {YargsError} from "./util/errors";
 
+import {assert} from "@chainsafe/lodestar-utils";
+
+assert.hello;
+assert.bye+1;
```
mpetrunic: ehem :rofl:

mpetrunic (changes requested):

packages/lodestar-cli/src/index.ts
dapplion: hahaha :shushing_face:

mpetrunic (approved):

wemeetagain (approved):

PR #1143, opened 2020-07-12
Remove unnecessary try / catch on mkdir recursive

dapplion:
You can call fs.mkdir with recursive = true on the same directory multiple times. Is there any specific version why the try / catch was added in the first place?


wemeetagain (approved):
>  Is there any specific version why the try / catch was added in the first place?

Not that I know of? XD

PR #1149, opened 2020-07-13
Fix undefined option in lodestar dev --help

### LINKED ISSUE #1148: CLI: genesisStateFile appears as undefined Option in CLI help instead of as a Command
**Describe the bug**

In these instructions https://chainsafe.github.io/lodestar/usage/local/, the command that's shown uses command  `--chain.genesisStateFile ./dev/genesis.ssz`, where the path of that file is relative to the root directory that you define with option `--rootDir`  (which is `./.lodestar` by default).

When you clone the latest 'master' branch and run `yarn run cli dev --help` it shows 
```
Command used to quickly bootstrap beacon node and validators

dev
  ...

Options:
  ...
  --undefined                   Genesis state in ssz-encoded format   [string]
```

Notice how it's missing the following, and just shows it as `--undefined` in the Options section, whereas I think it should instead be shown as a Command as shown below
```
chain
  --chain.genesisStateFile
```

**Expected behavior**

```
chain
  --chain.genesisStateFile    Genesis state file in ssz-encoded format. File location relative to Lodestar root directory.        [string] [ default: "" ]
```

**Steps to Reproduce**

1. Clone the latest Lodestart repo `git clone https://github.com/ChainSafe/lodestar`
2. Switch to a version of Node.js that's later than v12 (requirement of package.json) e.g. `nvm use 14.4.0`
3. Build from source `yarn install; yarn run build`
4. View CLI dev help `yarn run cli dev --help`
5. See error

 - OS: macOS Mojave 10.14.6
 - Version:
```
$ yarn run cli --version
0.9.0
```
 - Branch: master
 - Commit hash: 26a935ce7973c95d58570ef55c71ebc8a18bb61b

twoeths:
resolves #1148 

`./bin/lodestar dev --help` now prints:
```
chain
  --chain.genesisStateFile      Genesis state in ssz-encoded format                                                       [string]
  --chain.preset, --chain.name  Specifies the default eth2 spec type [string] [choices: "mainnet", "minimal"] [default: "minimal"]

```

### DIFF
--- a/packages/lodestar-cli/src/cmds/beacon/cmds/run/options/chain.ts
+++ b/packages/lodestar-cli/src/cmds/beacon/cmds/run/options/chain.ts
@@ -9,6 +9,7 @@ export const chainPreset: Options = {
   type: "string",
   choices: ["mainnet", "minimal"],
   default: "mainnet",
+  group: "chain"
 };
 
 export const chainGenesisStateFile: Options = {
@@ -18,4 +19,5 @@ export const chainGenesisStateFile: Options = {
   description: "Genesis state in ssz-encoded format",
   type: "string",
   normalize: true,
+  group: "chain"
 };
--- a/packages/lodestar-cli/src/cmds/dev/options/chain.ts
+++ b/packages/lodestar-cli/src/cmds/dev/options/chain.ts
@@ -1,6 +1,15 @@
 import {Options} from "yargs";
 
-export {chainGenesisStateFile} from "../../beacon/cmds/run/options/chain";
+// if we import from somewhere it'll print out "undefined" help
+export const chainGenesisStateFile: Options = {
+  alias: [
+    "chain.genesisStateFile",
+  ],
+  description: "Genesis state in ssz-encoded format",
+  type: "string",
+  normalize: true,
+  group: "chain"
+};
 
 export const chainPreset: Options = {
   alias: [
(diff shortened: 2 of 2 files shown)


mpetrunic:
I don't think that's ok, we will have bunch of shared option. Can you debug whats different inside yargs when using exported option?

wemeetagain:
I think I know whats going on, (haven't debugged it yet tho).
Its related to https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar-cli/src/util/canonicalOptions.ts

We're doing something questionable to get the naming of options that we want. By default, when using `yargs.options(options: Record<string, Options>)`, the first name shown for each option, eg: in `--help`, is the key in the options record.

So to enforce the key that we want, we've been using `canonicalOptions`, currently designed in a way to _abuse_ the existing `Options` interface.
We use `canonicalOptions` to pop the option.alias[0] if it exists, and use that as the key in the options object.
This "pop" behavior is not good, for the reasons we see here.
I think we've run `canonicalOptions` already (for the beacon run command), and the original exported optoins object has been modified!

Likely what we should do is extend the yargs `Options` interface like:
```typescript
interface ICLIOptions {
  canonicalAlias?: string;
}
```
Or something to that effect.
Then in `canonicalOptions`, we can use this additional metadata to set the alias without mutating the underlying object.

wemeetagain:
Closing in favor of #1210

PR #1153, opened 2020-07-14
Fetch blocks by parent root

### LINKED ISSUE #1069: Searching blocks by parentRoot property

mpetrunic:
resolves #1069

### DIFF
--- a/packages/lodestar/src/chain/forkChoice/arrayDag/lmdGhost.ts
+++ b/packages/lodestar/src/chain/forkChoice/arrayDag/lmdGhost.ts
@@ -379,6 +379,18 @@ export class ArrayDagLMDGHOST extends (EventEmitter as { new(): ForkChoiceEventE
     return (node)? this.toBlockSummary(node) : null;
   }
 
+  public getBlockSummaryByParentBlockRoot(blockRoot: Uint8Array): BlockSummary[] {
+    return Object.values(this.nodes)
+      .filter((node) => {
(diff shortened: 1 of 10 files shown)


packages/lodestar/src/chain/forkChoice/arrayDag/lmdGhost.ts
```diff
@@ -379,6 +379,18 @@ export class ArrayDagLMDGHOST extends (EventEmitter as { new(): ForkChoiceEventE
     return (node)? this.toBlockSummary(node) : null;
   }
 
+  public getBlockSummaryByParentBlockRoot(blockRoot: Uint8Array): BlockSummary[] {
```
twoeths, line 382: We indexed everything in forkchoice already so should be able to get parent by `nodeIndices` and `nodes` then we can get all child of that found node.

wemeetagain (approved):

PR #1157, opened 2020-07-14
Fix cli published files

### LINKED ISSUE #1145: cli bin in node_modules/.bin require is broken
**Describe the bug**

After installing the CLI and running it locally with `./node_modules/.bin/lodestar` you get a module node found error
```
lion@lion-gram:~/Code/eth2.0/test/test-cli-npm$ ./node_modules/.bin/lodestar 
internal/modules/cjs/loader.js:796
    throw err;
    ^

Error: Cannot find module './cmds/dev'
Require stack:
- /home/lion/Code/eth2.0/test/test-cli-npm/node_modules/@chainsafe/lodestar-cli/lib/index.js
- /home/lion/Code/eth2.0/test/test-cli-npm/node_modules/@chainsafe/lodestar-cli/bin/lodestar
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:793:17)
    at Function.Module._load (internal/modules/cjs/loader.js:686:27)
    at Module.require (internal/modules/cjs/loader.js:848:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/home/lion/Code/eth2.0/test/test-cli-npm/node_modules/@chainsafe/lodestar-cli/lib/index.js:5:12)
    at Module._compile (internal/modules/cjs/loader.js:955:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:991:10)
    at Module.load (internal/modules/cjs/loader.js:811:32)
    at Function.Module._load (internal/modules/cjs/loader.js:723:14)
    at Module.require (internal/modules/cjs/loader.js:848:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/home/lion/Code/eth2.0/test/test-cli-npm/node_modules/@chainsafe/lodestar-cli/lib/index.js',
    '/home/lion/Code/eth2.0/test/test-cli-npm/node_modules/@chainsafe/lodestar-cli/bin/lodestar'
  ]
}

```

**Expected behavior**
CLI should work

**Steps to Reproduce**
```
yarn add @chainsafe/lodestar-cli
./node_modules/.bin/lodestar
```

wemeetagain:
from @dapplion 

The lodestar-cli package published in NPM seems to be empty
```bash
lion@lion-gram:~/Code/eth2.0/test/test-cli-npm$ tree node_modules/@chainsafe/lodestar-cli/
node_modules/@chainsafe/lodestar-cli/
├── bin
│   └── lodestar
├── lib
│   └── index.js
├── LICENSE
├── node_modules
├── package.json
└── README.md

3 directories, 5 files
```

Fixes #1145
This is the resulting directory after installing the latest version from NPM

---

This should fix this issue. to test, clone down, `lerna bootstrap`, `cd lodestar-cli`, `npm publish --dry-run`

### DIFF
--- a/packages/lodestar-cli/package.json
+++ b/packages/lodestar-cli/package.json
@@ -9,7 +9,10 @@
   },
   "homepage": "https://github.com/ChainSafe/lodestar#readme",
   "main": "lib/index.js",
-  "files": [],
+  "files": [
+    "lib",
+    "bin"
+  ],
   "scripts": {
     "clean": "rm -rf lib && rm -f tsconfig.tsbuildinfo",
     "build": "yarn run build:lib && yarn run build:types",
--- a/packages/lodestar-cli/package.json
+++ b/packages/lodestar-cli/package.json
@@ -10,9 +10,13 @@
   "homepage": "https://github.com/ChainSafe/lodestar#readme",
   "main": "lib/index.js",
   "files": [
-    "lib",
-    "bin"
+    "lib/**/*.d.ts",
+    "lib/**/*.js",
+    "lib/**/*.js.map"
   ],
+  "bin": {
+    "lodestar": "lib/index.js"
+  },
   "scripts": {
     "clean": "rm -rf lib && rm -f tsconfig.tsbuildinfo",
     "build": "yarn run build:lib && yarn run build:types",
@@ -68,9 +72,6 @@
     "rimraf": "^3.0.0",
     "uuidv4": "^6.1.1",
     "yargs": "^15.3.1"
-  },
-  "bin": {
-    "lodestar": "./bin/lodestar"
   },
   "devDependencies": {
     "@types/chai": "4.2.0",


packages/lodestar-cli/package.json
```diff
@@ -9,7 +9,10 @@
   },
   "homepage": "https://github.com/ChainSafe/lodestar#readme",
   "main": "lib/index.js",
-  "files": [],
+  "files": [
+    "lib",
+    "bin"
+  ],
```
dapplion, line 16: To stay consistent with the rest of packages in the repo you can use
```
"files": [
    "lib/**/*.d.ts",
    "lib/**/*.js",
    "lib/**/*.js.map"
  ],
```
Also, after doing some tests, `npm` or `yarn` assigns the correct permissions to the symlinks created in `node_modules/.bin/`. So it should not be necessary to push the bin directory to NPM while changing `manifest.json#bin` to point to `lib/index.js`. The bin directory is still useful but for local scripting without using the `node` prefix

dapplion:
Fixes https://github.com/ChainSafe/lodestar/issues/1145

packages/lodestar-cli/package.json
wemeetagain: ah, right. we can avoid publishing the extra file. I'll update this PR to remove `bin` from `files`, add `lib/index.js` as the bin.

dapplion (approved):

packages/lodestar-cli/package.json
```diff
@@ -9,7 +9,14 @@
   },
   "homepage": "https://github.com/ChainSafe/lodestar#readme",
   "main": "lib/index.js",
-  "files": [],
+  "files": [
+    "lib/**/*.d.ts",
+    "lib/**/*.js",
+    "lib/**/*.js.map"
```
mpetrunic, line 15: include source files or this map file is useless :smile:

mpetrunic (dismissed):

packages/lodestar-cli/package.json
dapplion: In case the CLI throws an error, won't the source maps be useful to trace the stack to the TS source?

mpetrunic: yes, but it doesn't work if there aren't any source files, should add `src/**/*.ts` line here. 
That said, sourcemaps only work if you have node12.12 and add flag --enable-source-maps

dapplion: In other CLIs I've developed source files work fine without adding the source `*.ts` files. But I initially suggested this change for consistency since this `"files"` is the one currently used in all other Lodestar monorepo packages. If it's okay we can leave it like this for the sake of consistency and discuss in a subsequent issue if we should change the files strategy for the monorepo in all packages at once.

mpetrunic: > In other CLIs I've developed source files work fine without adding the source *.ts files

You are right, it will display correct path, we can leave it like this. Those stack traces are links in my terminal and opens files when clicked. I doubt many people are using it so it's totally fine not send it.

dapplion: Summoning @wemeetagain so we can have a working CLI in NPM :heart:

PR #1161, opened 2020-07-15
Ensure clearTimeout for all requests

### LINKED ISSUE #1127: Altona: Initial Sync suspends with timeout
**Describe the bug**
Sync Altona and the process suspends, there are a couple of `response timeout` errors

**Steps to Reproduce**
Sync Altona

**Logs**
This seems to be too slow to detect req/resp timeout as it took almost 1h (09:51:11 - 10:49:34):
```
2020-07-10 09:51:11  [NETWORK]       verbose: sending beacon_blocks_by_range request to 16Uiu2HAm6FBad3xnXj4WkkjpjaQQedn4rCmh8jCww2uw6MSSFNs2 requestId=6d3dc33918668bed, encoding=ssz_snappy
...
2020-07-10 10:49:34  [NETWORK]         error: failed to send request 6d3dc33918668bed to peer 16Uiu2HAm6FBad3xnXj4WkkjpjaQQedn4rCmh8jCww2uw6MSSFNs2 Error: response timeout
Error: response timeout
    at nextAbortHandler (/Users/tuyennguyen/Projects/workshop/lodestar/node_modules/abortable-iterator/index.js:37:20)
```

This seems to be too fast to detect req/resp timeout as it happens in same second (10:52:26):
```
2020-07-10 10:52:26  [NETWORK]       verbose: sending beacon_blocks_by_range request to 16Uiu2HAmV2BjFPcY4QYRuWE5DxpS2KpfLogiSmVKe9Lp9mTt3XBU requestId=2e576ea0a9ed7b15, encoding=ssz_snappy
...
2020-07-10 10:52:26  [NETWORK]         error: failed to send request 2e576ea0a9ed7b15 to peer 16Uiu2HAmV2BjFPcY4QYRuWE5DxpS2KpfLogiSmVKe9Lp9mTt3XBU Error: response timeout
Error: response timeout
    at nextAbortHandler (/Users/tuyennguyen/Projects/workshop/lodestar/node_modules/abortable-iterator/index.js:37:20
```

twoeths:
resolves #1127 
+ Close stream when timeout
+ Currently we don't cancel timers if we do `break` inside `eth2ResponseDecode`. It means we don't call `cancelTimer` for all single_response requests.
+ One solution is to put `eth2RequestEncode` after `eth2ResponseDecode`, but we can't handle TTFB in that case
+ So I have to split `eth2ResponseTimer` to `checkTimer` and `clearTimer`


packages/lodestar/src/network/util.ts
```diff
@@ -61,17 +61,23 @@ export function isRequestSingleChunk(method: Method): boolean {
...
-export function eth2ResponseTimer<T>(): (source: AsyncIterable<T>) => AsyncGenerator<T> {
+type PipeFn<T> = (source: AsyncIterable<T>) => AsyncGenerator<T>;
+export function eth2ResponseTimer<T>(stream: Stream): [PipeFn<T>, PipeFn<T>] {
   const controller = new AbortController();
-  let responseTimer = setTimeout(() => controller.abort(), TTFB_TIMEOUT);
+  const abort = (): void => {
+    controller.abort();
+    // @ts-ignore
+    stream.close();
+  };
```
mpetrunic: controller.signal.addEventListener('abort', () => stream.close())

mpetrunic (changes requested):
> * Currently we don't cancel timers if we do `break` inside `eth2ResponseDecode`. It means we don't call `cancelTimer` for all single_response requests.

That's shouldn't be true, line after for..await (clearTimer) will execute if you break loop somewhere.
Can you add testCase to demonstrate your bug?

 I don't think we should complicate it like this unless there is reason for that.

twoeths:
>That's shouldn't be true, line after for..await (clearTimer) will execute if you break loop somewhere.
Can you add testCase to demonstrate your bug?

@mpetrunic yeah it was a surprise to me too when I tried to print timeout when syncing altona, all of them come from either status or metadata request.
You can have a look at #1168

twoeths:
@mpetrunic so I tried the `controller.abort()` instead of `break` and it really does the job of clearing timeout, but there's a concern regarding `returnOnAbort` option. We want `controller.abort()` to raise error in case of timeout and not to raise error in case of a normal `break` of for loop.

So should we consider the change in this PR which is not really nice, or do you have any other ideas?

mpetrunic:
> @mpetrunic so I tried the `controller.abort()` instead of `break` and it really does the job of clearing timeout, but there's a concern regarding `returnOnAbort` option. We want `controller.abort()` to raise error in case of timeout and not to raise error in case of a normal `break` of for loop.
> 
> So should we consider the change in this PR which is not really nice, or do you have any other ideas?

Abort controller in eth2ResponseTimer should not have returnOnError true, while stream abort controller should.

twoeths:
> Abort controller in eth2ResponseTimer should not have returnOnError true, while stream abort controller should.

@mpetrunic so I tried with 2 different controllers and not able to get though unit tests, then I found out that `controller.abort()` is not equivalent to `break` because it'll run until the end of for loop (unlike `break`).
If you want to give it a try, it failed at this unit test
```
 it.only("should work - response stream with error - ssz", async function () {
    const chunks = generateBlockChunks(10);
    chunks[4].status = RpcResponseStatus.ERR_INVALID_REQ;
    chunks[4].body = encodeP2pErrorMessage(config, "Invalid request");
    const responses = await pipe(
      abortSource(chunks, controller.signal, {returnOnAbort: true}),
      eth2ResponseEncode(config, loggerStub, Method.BeaconBlocksByRange, ReqRespEncoding.SSZ),
      eth2ResponseDecode(config, loggerStub, Method.BeaconBlocksByRange, ReqRespEncoding.SSZ, "abc", controller),
      collect
    ) as ResponseBody[];
    expect(responses.length).to.be.equal(4);
  });
```
We can for sure adding a separate variable to make `controller.abort()` the same to `break` but afraid it'd be too far, we make things complex unnecessarily.

I hope you have a simpler/cleaner way.

mpetrunic:
Use continue after abort

twoeths:
that works, thanks ^^

packages/lodestar/src/network/reqResp.ts
```diff
@@ -277,14 +279,23 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
     const {libp2p, config, logger} = this;
     return (async function * () {
       const protocol = createRpcProtocol(method, encoding);
-      const {stream} = await libp2p.dialProtocol(peerId, protocol) as {stream: Stream};
+      let connectionTimer;
+      const {stream} = await Promise.race([
+        libp2p.dialProtocol(peerId, protocol),
```
mpetrunic: dial protocol accept third param options with abort signal :)

https://github.com/libp2p/js-libp2p/blob/v0.28.3/src/index.js#L316

packages/lodestar/src/network/reqResp.ts
```diff
@@ -277,14 +279,23 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
...
+      clearTimeout(connectionTimer);
+      const controller = new AbortController();
       logger.verbose(`sending ${method} request to ${peerId.toB58String()}`, {requestId, encoding});
       yield* pipe(
         (body !== null && body !== undefined) ? [body] : [null],
         eth2RequestEncode(config, logger, method, encoding),
-        stream,
-        eth2ResponseTimer(),
-        eth2ResponseDecode(config, logger, method, encoding, requestId)
+        // @ts-ignore
```
mpetrunic: can we fix this? Whats the error?

packages/lodestar/src/network/reqResp.ts
mpetrunic: I would also extract to function that accepts peerId, protocol and timeout

packages/lodestar/src/network/util.ts
```diff
@@ -61,8 +61,10 @@ export function isRequestSingleChunk(method: Method): boolean {
   return Methods[method].responseType === MethodResponseType.SingleResponse;
 }
 
-export function eth2ResponseTimer<T>(): (source: AsyncIterable<T>) => AsyncGenerator<T> {
+export function eth2ResponseTimer<T>(stream: Stream): (source: AsyncIterable<T>) => AsyncGenerator<T> {
   const controller = new AbortController();
+  // @ts-ignore
```
mpetrunic: why does it complain here?

packages/lodestar/src/network/reqResp.ts
```diff
@@ -277,14 +279,23 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
     const {libp2p, config, logger} = this;
     return (async function * () {
       const protocol = createRpcProtocol(method, encoding);
-      const {stream} = await libp2p.dialProtocol(peerId, protocol) as {stream: Stream};
+      let connectionTimer;
+      const {stream} = await Promise.race([
```
twoeths: this is to fix #1193

packages/lodestar/src/network/util.ts
twoeths: it can't get latest version of `libp2p-ts` although I do `yarn cache clean`

mpetrunic: does it work on ci? It would be better to investigate this error instead of doing ignore like this. Maybe you have it linked or something?

twoeths: I guess if I point to a different github branch then point back to master it should work, will give it a try

packages/lodestar-cli/refresh-peer.sh
```diff
@@ -0,0 +1,9 @@
+rm -rf .altona3
```
twoeths: will remove in my next commit when I switch to libp2p-ts master :D

packages/lodestar/src/network/util.ts
```diff
@@ -81,3 +82,15 @@ export function eth2ResponseTimer<T>(): (source: AsyncIterable<T>) => AsyncGener
     })();
   };
 }
+
+export async function dialProtocol(libp2p: LibP2p, peerId: PeerId, protocol: string, timeout: number): Promise<Stream> {
+  let connectionTimer;
+  const {stream} = await Promise.race([
+    libp2p.dialProtocol(peerId, protocol),
```
mpetrunic: you aren't using abort signal (I think abort signal will reject dial protocol anyways)

packages/lodestar/src/network/util.ts
```diff
@@ -61,8 +61,9 @@ export function isRequestSingleChunk(method: Method): boolean {
   return Methods[method].responseType === MethodResponseType.SingleResponse;
 }
 
-export function eth2ResponseTimer<T>(): (source: AsyncIterable<T>) => AsyncGenerator<T> {
+export function eth2ResponseTimer<T>(stream: Stream): (source: AsyncIterable<T>) => AsyncGenerator<T> {
```
mpetrunic: why not receive stream abort controller (call it streamCloseController)?

I don't think we need another abort source in this method now?

packages/lodestar/src/network/util.ts
twoeths: so create new separate AbortController here and put its signal to `dialProtocol` and do `controller.abort()` in `setTimeout`?

packages/lodestar/src/network/util.ts
twoeths: the other `abortDuplex` has `{returnOnAbort: true}`, this `abortSource` does not have that, we want pipe to throw Error in case of timeout

twoeths: > why not receive stream abort controller (call it streamCloseController)?

I'm not so sure we can use the same AbortController here. One is used to abort the original stream nicely, one is used to abort the stream with an Error.

packages/lodestar/src/network/util.ts
mpetrunic: yes

packages/lodestar/src/network/util.ts
mpetrunic: well if you use original one, it would close stream nicely (basically same thing as you did here on stream directly but removes dependency of this method to stream) and to a cleanup while one used here would throw for rest of methods in pipe from what I understand.

twoeths: If we use same controller, calling `controller.abort()` would both trigger abort the original iterator (which we don't want Error) and abort this internal iterator (which we want Error so we can catch in `sendRequest`) so the behavior is indeterministic.

so I'd use separate AbortControllers here unless there's performance issue later..

mpetrunic: Hmm it should be following flow in case of timeout:
1. streamAbortController.abort() - won't break anything, it will allow rest of transformations to end
2. timeoutController.abort() - will throw error from this function below to handle

twoeths: 👍 I'll give it a try

twoeths: so this unit test failed: `can send/receive status messages from connected peers` if I use the same controller, error thrown with `response timeout`.
It's because it sent a `status` requests which is single chunk response and it calls break (now replaced by `controller.abort();continue;` and caused timeout error.

mpetrunic (approved):

wemeetagain (approved):

PR #1166, opened 2020-07-15
Add --altona flag to beacon cmds

### LINKED ISSUE #1163: Add --altona flag to CLI to automate setup
As suggested in https://github.com/ChainSafe/lodestar/pull/1159#pullrequestreview-448701077
> I was hopping that this Pr includes --testnet altona flag that would bootstrap everything and run node against altona

Add a flag to `beacon run` that sets up everything and runs the chain with the necessary files and config in one command.

dapplion:
Fixes https://github.com/ChainSafe/lodestar/issues/1163.

To start and configure a Beacon node now you just do
```
lodestar beacon run --testnet altona
```

The flag will download the genesis file if not in disk, and apply the appropriate config values. This flag is at the `beacon` level, so it applies both to `run` and `init`.


packages/lodestar-cli/src/cmds/beacon/config.ts
```diff
@@ -30,9 +31,14 @@ export async function writeBeaconConfig(filename: string, config: Partial<IBeaco
...
  * This needs to be a synchronous function because it will be run as part of the yargs 'build' step
+ * If the config file is not found, the default values will apply.
  */
 export function readBeaconConfig(filename: string): Partial<IBeaconNodeOptions> {
-  return readFileSync(filename) as Partial<IBeaconNodeOptions>;
+  if (fs.existsSync(filename)) {
+    return readFileSync(filename) as Partial<IBeaconNodeOptions>;
+  } else {
+    return {};
+  }
```
dapplion, line 41: Letting the config file not exist is necessary to allow running `run` without running `init` first, which may not necessary anyway.

packages/lodestar-cli/package.json
```diff
@@ -65,6 +65,7 @@
     "deepmerge": "^3.2.0",
     "ethers": "^5.0.2",
     "expand-tilde": "^2.0.2",
+    "got": "^11.5.0",
```
dapplion, line 68: got is a lightweight and modern npm client already used in previous lodestar-cli dependencies

packages/lodestar-cli/src/cmds/beacon/testnets/index.ts
```diff
@@ -0,0 +1,47 @@
...
+export async function downloadGenesisFile(
+  testnet: "altona",
+  options: Partial<IBeaconArgs>
+): Promise<void> {
+  const genesisFileUrl = getGenesisFileUrl(testnet);
+  const genesisFilePath = options.chain.genesisStateFile;
+
+  if (!fs.existsSync(genesisFilePath)) {
+    fs.mkdirSync(path.parse(genesisFilePath).dir, {recursive: true});
+    await promisify(stream.pipeline)(
```
twoeths, line 52: @mpetrunic do you see if we can use `pipe` here?

twoeths (dismissed):
lgtm, from now on it's easier to setup to cloud environments like heroku ;)

packages/lodestar-cli/src/cmds/beacon/cmds/init/init.ts
```diff
@@ -1,14 +1,26 @@
...
 import {initPeerId, initEnr, readPeerId} from "../../../../network";
 import {initBeaconConfig} from "../../config";
+import {getTestnetConfig, downloadGenesisFile} from "../../testnets";
 
 /**
  * Initialize lodestar-cli with an on-disk configuration
  */
 export async function init(args: Arguments<IBeaconArgs>): Promise<void> {
+  // Auto-setup altona
+  if (args.altona) {
```
mpetrunic: Might be better to use --network <altona|mainnet> as other clients

packages/lodestar-cli/src/cmds/beacon/testnets/altona.json
```diff
@@ -0,0 +1,78 @@
...
+      "level": "verbose"
+    },
+    "network": {
+      "level": "verbose"
+    },
+    "eth1": {
+      "level": "verbose"
+    },
+    "sync": {
+      "level": "verbose"
```
mpetrunic: MAybe we should have info by default ?

packages/lodestar-cli/src/cmds/beacon/testnets/index.ts
```diff
@@ -0,0 +1,47 @@
...
+      return JSON.parse(fs.readFileSync(path.join(__dirname, "altona.json"), "utf8"));
+    default:
+      throw Error(`Testnet not supported: ${testnet}`);
+  }
+}
+
+function getGenesisFileUrl(testnet: Testnet): string {
+  switch (testnet) {
+    case "altona":
+      return "https://github.com/eth2-clients/eth2-testnets/raw/master/shared/altona/genesis.ssz";
```
mpetrunic: use permalink to git hash, so we don't break if they rearrange paths

packages/lodestar-cli/src/cmds/beacon/testnets/index.ts
```diff
@@ -0,0 +1,47 @@
...
+      throw Error(`Testnet not supported: ${testnet}`);
+  }
+}
+
+/**
+ * Downloads a genesis file per testnet if it does not exist
+ * @param options
+ */
+export async function downloadGenesisFile(
+  testnet: "altona",
```
mpetrunic:
```suggestion
  testnet: Testnet,
```

packages/lodestar-cli/src/cmds/beacon/testnets/index.ts
mpetrunic: They are streams so which means they are also iterators:
```
await pipe(
   got.stream(genesisFileUrl),
   fs.createWriteStream(genesisFilePath)
)
```
should work but this is same thing more or less

packages/lodestar-cli/src/cmds/beacon/testnets/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  options: Partial<IBeaconArgs>
+): Promise<void> {
+  const genesisFileUrl = getGenesisFileUrl(testnet);
+  const genesisFilePath = options.chain.genesisStateFile;
+
+  if (!fs.existsSync(genesisFilePath)) {
+    fs.mkdirSync(path.parse(genesisFilePath).dir, {recursive: true});
```
mpetrunic, line 51: path should be resolved relative to rootDir (either default one received as cli param)

packages/lodestar-cli/src/cmds/beacon/testnets/index.ts
mpetrunic: That said, genesis file is always the same but we should probably fetch bootnodes (merge with some hardcoded in config) from repo since they tend to change as peoples nodes go down.

mpetrunic (commented):
awesome

packages/lodestar-cli/src/cmds/beacon/testnets/index.ts
dapplion: I followed the suggested use within `got` https://www.npmjs.com/package/got#streams
Happy to change it if you deem it necessary.

packages/lodestar-cli/src/cmds/beacon/cmds/init/init.ts
dapplion: I like the suggestion but we are already using a `network` parameter here
https://github.com/ChainSafe/lodestar/blob/9c5fef9c83645e1c26e5fb43ec263237ca0c53fb/packages/lodestar-cli/src/cmds/beacon/options/beaconFile.ts#L8
Would you suggest to use a different name than `network` that's not used or go with altona?

packages/lodestar-cli/src/cmds/beacon/testnets/index.ts
dapplion: Switched to permalink. Where can we fetched the updated bootnodes from?

packages/lodestar-cli/src/cmds/beacon/cmds/init/init.ts
mpetrunic: argh, maybe chain.network then?

packages/lodestar-cli/src/cmds/beacon/testnets/index.ts
mpetrunic: https://github.com/eth2-clients/eth2-testnets/blob/master/shared/altona/bootstrap_nodes.txt

packages/lodestar-cli/src/cmds/beacon/cmds/init/init.ts
wemeetagain: `--testnet`? blank == mainnet

packages/lodestar-cli/src/cmds/beacon/testnets/index.ts
dapplion: added

packages/lodestar-cli/src/cmds/beacon/cmds/init/init.ts
```diff
@@ -1,14 +1,36 @@
...
 import {initBeaconConfig} from "../../config";
+import {getTestnetConfig, downloadGenesisFile, fetchBootnodes} from "../../testnets";
 
 /**
  * Initialize lodestar-cli with an on-disk configuration
  */
 export async function init(args: Arguments<IBeaconArgs>): Promise<void> {
+  // Auto-setup altona
+  if (args.testnet) {
+    const altonaConfig = getTestnetConfig(args.testnet);
```
wemeetagain: should these variable names reference altona now? we'll likely support "the multiclient testnet", and other testnets, as they come out

packages/lodestar-cli/src/cmds/beacon/cmds/run/run.ts
```diff
@@ -15,13 +15,19 @@ import {WinstonLogger} from "@chainsafe/lodestar-utils";
...
+import {init as initBeacon} from "../init/init";
 
 /**
  * Run a beacon node
  */
 export async function run(options: Arguments<IBeaconArgs & Partial<IBeaconNodeOptions>>): Promise<void> {
   await initBLS();
 
+  // Auto-setup altona
+  if (options.altona) {
```
wemeetagain: should this be options.testnet?

wemeetagain:
@dapplion why did you close this?

dapplion:
> @dapplion why did you close this?

Accidentally closed when migrating branches from my fork to the main repo

wemeetagain (approved):

PR #1168, opened 2020-07-15
[DNM] Demo clearTimer is not called for single_response message

twoeths:
This is just to demo the `cancelTimer` issue with single response requests.
Whenever we break inside for loop of `eth2ResponseDecode`, pipe decides to finish everything and `cancelTimer` is not called after `for await`

### DIFF
--- a/packages/lodestar/test/unit/network/reqResp.test.ts
+++ b/packages/lodestar/test/unit/network/reqResp.test.ts
@@ -101,20 +101,17 @@ describe("[network] rpc", () => {
         rpcA.sendResponse(id, null, body as Status);
       }, 100);
     });
-    try {
-      const statusExpected: Status = {
-        forkDigest: Buffer.alloc(4),
-        finalizedRoot: Buffer.alloc(32),
-        finalizedEpoch: 0,
-        headRoot: Buffer.alloc(32),
-        headSlot: 0,
-      };
+    const statusExpected: Status = {
+      forkDigest: Buffer.alloc(4),
+      finalizedRoot: Buffer.alloc(32),
+      finalizedEpoch: 0,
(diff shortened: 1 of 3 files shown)


packages/lodestar/test/unit/network/reqResp.test.ts
```diff
@@ -155,6 +152,7 @@ describe("[network] rpc", () => {
         }
         reqIndex ++;
       }
+      assert(loggerStub.info.calledWith("cancelTimer called"), "cancelTimer not called!!");
```
twoeths, line 155: `cancelTimer` is called for beacon_block_by_range_requests

packages/lodestar/test/unit/network/reqResp.test.ts
```diff
@@ -101,20 +101,17 @@ describe("[network] rpc", () => {
...
+    };
 
-      const statusActual = await rpcB.status(nodeA.peerId, statusExpected);
-      assert.deepEqual(statusActual, statusExpected);
-    } catch (e) {
-      assert.fail("status not received");
-    }
+    const statusActual = await rpcB.status(nodeA.peerId, statusExpected);
+    assert.deepEqual(statusActual, statusExpected);
+    assert(loggerStub.info.calledWith("cancelTimer called"), "cancelTimer not called!!");
```
twoeths, line 114: `cancelTimer` is not called for status requests

mpetrunic:
I've opened up issue for this: https://github.com/alanshaw/it-pipe/issues/3

It appears like bug but it might be like that by design not sure.

twoeths:
we should use `controller.abort()` instead of `break` in order for functions after `for await` gets called.

PR #1176, opened 2020-07-16
YargsError handling across whole lodestar-cli package

### LINKED ISSUE #1099: Use custom error for expected CLI exceptions
Use a custom error for CLI errors that are expected and should not trigger an exception nor show a stack trace. Then catch that error at the top level, log the error message and exit cleanly with `process.exit(1)`.

Check https://github.com/ChainSafe/lodestar/pull/1093 and https://github.com/ChainSafe/lodestar/pull/1093#discussion_r449721256 with its `YargsError` as a reference.

3xtr4t3rr3str14l:
fixes #1099 by using YargsOptions across the whole lodestar-cli package


packages/lodestar-cli/src/cmds/account/cmds/validator/deposit.ts
```diff
@@ -96,7 +96,7 @@ export async function handler(options: IAccountValidatorDepositOptions): Promise
       value: (depositData.amount * BigInt(1e9)).toString(),
       data: rlp
     });
-    if (!tx.hash) throw Error("No transaction hash");
+    if (!tx.hash) throw new YargsError("No transaction hash");
```
dapplion: This is not a quite "expected" error so I would throw a regular error

packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
```diff
@@ -14,7 +14,13 @@ export function builder(yargs: Argv<IBeaconArgs>): Argv<IBeaconArgs> {
...
   const args = parseArgs(yargs) as IBeaconInitArgs;
   if (args.templateConfigFile) {
-    yargs.config(readBeaconConfig(args.templateConfigFile));
+    try {
+      yargs.config(readBeaconConfig(args.templateConfigFile));
+    } catch(error) {
+      // eslint-disable-next-line no-console
+      console.log(` ✖ ${error.message}\n`);
+      process.exit(1);
+    }
```
dapplion: Errors thrown here are not catched by yargs.fail handler? I would avoid using process.exit() anywhere but a very predictable spot, like in the root index of the CLI.

packages/lodestar-cli/src/depositContract/depositData.ts
```diff
@@ -35,7 +36,7 @@ export function decodeEth1TxData(
   // Sanity check
   const depositDataRoot = config.types.DepositData.hashTreeRoot(depositData);
   if (depositDataRoot != root)
-    throw Error("deposit data root mismatch");
+    throw new YargsError("deposit data root mismatch");
```
dapplion: This is not an "expected" error, I would throw a regular error

packages/lodestar-cli/src/util/file.ts
```diff
@@ -31,7 +32,7 @@ export function stringify<T=Json>(obj: T, fileFormat: FileFormat): string {
       contents = JSON.stringify(obj, null, 2);
       break;
     default:
-      throw new Error("Invalid filetype");
+      throw new YargsError("Invalid filetype");
```
dapplion: This is not an "expected" error, I would throw a regular error

dapplion (changes requested):
I would exclusively use YargsError for expected errors that just by seeing the error message the user can take action and correct the error. Such as `"Must supply either keystorePath, rpcUrl, or ipcPath"`. Any other error that is not that obvious must have a stack trace so we can help users debug the problem.

packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
3xtr4t3rr3str14l: yeah i wasn't able to figure out how get the errors to be thrown to the top-level for `beacon`.  for now, i'll take out that process.exit, but after this PR, i might start to look into refactoring mergeBeaconOptions and see if maybe that's the culprit

packages/lodestar-cli/src/util/file.ts
3xtr4t3rr3str14l: done

packages/lodestar-cli/src/depositContract/depositData.ts
3xtr4t3rr3str14l: done

packages/lodestar-cli/src/cmds/account/cmds/validator/deposit.ts
3xtr4t3rr3str14l: done

packages/lodestar-cli/src/cmds/beacon/cmds/run/index.ts
```diff
@@ -14,9 +14,18 @@ export const description = "Run a lodestar beacon node";
...
+  try {
+    yargsConfig = yargs
+      .options(beaconRunOptions)
+      .config(readBeaconConfig(args.config)) as unknown as Argv<IBeaconArgs & Partial<IBeaconNodeOptions>>;
+  } catch(error) {
+    // eslint-disable-next-line no-console
+    console.log(` ✖ ${error.message}\n`);
+  }
+
+  return yargsConfig;
```
dapplion: Is this type-safe? In case of error it will be returning an undefined object that the Typescript compilers assumes is a defined object.

dapplion:
I would exclusively use YargsError in yargs handlers, so to files within `lodestar-cli/src/cmds`.

packages/lodestar-cli/src/cmds/beacon/cmds/run/index.ts
3xtr4t3rr3str14l: fixed

3xtr4t3rr3str14l:
> I would exclusively use YargsError in yargs handlers, so to files within `lodestar-cli/src/cmds`.

ah yeah i see what you mean.  going to push something for that in a sec

dapplion (dismissed):

packages/lodestar-cli/src/util/file.ts
```diff
@@ -21,7 +21,7 @@ export function parse<T=Json>(contents: string, fileFormat: FileFormat): T {
     case FileFormat.json:
       return JSON.parse(contents);
     default:
-      throw new Error("Invalid filetype");
+      throw Error("Invalid filetype");
```
wemeetagain: `new` got omitted off here, this should still be `throw new Error`, same with the error in `stringify` in this file.

3xtr4t3rr3str14l: fixed

wemeetagain (approved):

PR #1177, opened 2020-07-17
Validate passphraseFile before wallet create

### LINKED ISSUE #1174: cli account wallet creation allows for a file that is not a passphrase file to be passed to the --passphraseFile option
**Describe the bug**

when i run `yarn run cli account wallet create --name=asdf --passphraseFile=yarn.lock` (where `yarn.lock` could be any non-passphrase file & `asdf` can be whatever name you want), I am able to generate a wallet with no obvious errors thrown (the UUID is shown, the mnemonic is shown, everything appears to be fine).

**Expected behavior**

I'm not sure if we are supposed to be able pass arbitrary data to this cli parameter, but I would expect it to error if we don't give it an expected file (and i'm not sure which file types we should be allowing, for the record).
I did notice that when I tried to pass a non-existent file that wasn't a .pass file, I would get the error `passphraseFile must end with .pass, make sure to not provide the actual password`, so maybe we just need to stitch up that error handler?

**Steps to Reproduce**

run `yarn run cli account wallet create --name=AName --passphraseFile=yarn.lock` (where the yarn.lock file could be replaced with any existing file on the local filesystem that is ostensibly not a passphrase file)

dapplion:
Fixes https://github.com/ChainSafe/lodestar/issues/1174.

In CLI cmd `account wallet create` validates that the passphraseFile has `.pass` extension even if it exists. It also validates that its contents:
- Do not have multilines
- Do not have length > 512 chars

This prevents the scenario brought up by https://github.com/ChainSafe/lodestar/issues/1174 where the user might typo the passphraseFile path and create a wallet with the contents of some random file.

### DIFF
--- a/packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
+++ b/packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
@@ -66,14 +66,6 @@ export async function handler(options: IWalletCreateOptions): Promise<void> {
     writeFile600Perm(passphraseFile, randomPassword());
   }
   const password = readPassphraseFile(passphraseFile);
-  // Validate the passphraseFile contents to prevent the user to create a wallet with a password
-  // that is the contents a random unintended file
-  try {
-    if (password.includes("\n")) throw Error("contains multiple lines");
-    if (password.length > 512) throw Error("is really long");
-  } catch (e) {
-    throw new YargsError(`passphraseFile ${passphraseFile} ${e.message}. Is this a well-formated passphraseFile?`);
-  }
 
   const walletManager = new WalletManager(accountPaths);
   const wallet = walletManager.createWallet(name, type, mnemonic, password);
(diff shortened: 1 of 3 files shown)


wemeetagain (dismissed):
:+1:

packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
```diff
@@ -59,14 +59,21 @@ export async function handler(options: IWalletCreateOptions): Promise<void> {
...
-    }
     writeFile600Perm(passphraseFile, randomPassword());
   }
-  const password = fs.readFileSync(passphraseFile, "utf8");
+  const password = readPassphraseFile(passphraseFile);
+  // Validate the passphraseFile contents to prevent the user to create a wallet with a password
+  // that is the contents a random unintended file
+  try {
+    if (password.includes("\n")) throw Error("contains multiple lines");
+    if (password.length > 512) throw Error("is really long");
```
3xtr4t3rr3str14l: where do we get the 512 number from?  also, could we perhaps move it to a constants file?

dapplion: It's a completely arbitrary number, it just seemed like something long enough that would never be used as a password. Happy to change it to something else.

dapplion: @3xtr4t3rr3str14l I've added a comment explaining the arbitrary nature of 512.

mpetrunic (approved):

wemeetagain (approved):

PR #1178, opened 2020-07-17
Faster attestation processor

wemeetagain:
In attestation processor
- Remove slow committee computation
- Remove special-case genesis state fetching

### DIFF
--- a/packages/lodestar/src/chain/attestation.ts
+++ b/packages/lodestar/src/chain/attestation.ts
@@ -1,8 +1,7 @@
-import {fromHexString, toHexString, TreeBacked} from "@chainsafe/ssz";
+import {fromHexString, toHexString} from "@chainsafe/ssz";
 import {
   Attestation,
   AttestationRootHex,
-  BeaconState,
   BlockRootHex,
   Root,
   SignedBeaconBlock,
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/chain/attestation.ts
```diff
@@ -97,19 +97,16 @@ export class AttestationProcessor implements IAttestationProcessor {
...
-      checkpointState = await this.db.stateArchive.get(0);
+    let stateCtx: ITreeStateContext;
+    const justifiedBlock = this.forkChoice.getBlockSummaryByBlockRoot(
+      justifiedCheckpoint.root.valueOf() as Uint8Array
+    );
+    if (justifiedBlock) {
+      stateCtx = await this.db.stateCache.get(justifiedBlock.stateRoot);
+    }
+    if (!justifiedBlock || !stateCtx) {
+      // should not happen
```
3xtr4t3rr3str14l, line 108: is this comment meant to reinforce the idea that an error should be thrown under these conditions?  or is this also to maybe say "this shouldn't happen, but here's a sanity check just in case".  or both?  not a big deal, but just kinda caught my eye as being maybe a little ambiguous

wemeetagain: This is just a carry over from the existing impl. (the comment was already there)

I think this is a "this shouldn't happen, but here's a sanity check just in case" kind of situation tho.

twoeths (dismissed):

twoeths:
unit tests failed

mpetrunic (approved):

twoeths (approved):

Issue #1103, opened 2020-07-05
Decompose IEth1Notifier into parts


wemeetagain:
I think we can (eventually) decompose this `IEth1Notifier` into a few parts, that can be run on-demand, rather than started globally. If we want to reuse resources, we could keep the provider and keep that around (it could be started/stopped in node.ts). Decomposing the functionality will lead to increased maintainability (each thing can be in its own file) and flexibility in how we can use this:
```typescript
type BlockTag = string | number

/**
 * Thin wrapper around ethers provider, tailored for the deposit contract
 */
interface IEth1Provider {
  /**
   * initialize contract, error if it doesn't exist or is invalid
   */
  start(): Promise<void>;
  stop(): Promise<void>;

  getBlock(blockTag: BlockTag): Promise<Block>;
  getDepositEvents(fromBlockTag: BlockTag, toBlockTag?: BlockTag): Promise<IDepositEvent[]>;

  // event for new blocks
  on(event: 'block', block: Block): void;
  off(event: 'block', block: Block): void;

  // event for module stopping, allows for event consumers to gracefully shut down
  on(event: 'stop'): void;
  off(event: 'stop'): void;
}

interface IBatchDepositEvents {
  depositEvents: IDepositEvent[];
  blockNumber: number;
}

/**
 * Stream of deposit events, batched by block number.
 * Includes blocks with 0 deposit events
 */
function getHistoricalDepositEventsStream(provider: IEth1Provider, from: blockNumber, to: blockNumber): Promise<Pushable<IBatchDepositEvents>>;

/**
 * Abortable stream of batched deposit events
 * data starting at `from`, continues at ETH1_FOLLOW_DISTANCE until aborted or IEthProvider stops
 */
function getFollowDepositEventsStream(provider: IEth1Provider, from: blockNumber): Promise<Pushable<IBatchDepositEvents>>;
```

From there we've separated how the data is consumed, database storage concerns, whether a stream has a single or multiple consumers, etc. We should write up a design doc determining what we store, when we store, which modules manage storage, etc. Likely, we will have a strategy for genesis and a strategy for proposing.

Eth1Data processing requires a depositData tree, either fetched from the db module, or managed by the genesis builder.
```typescript
/**
 * Create an Eth1Data based on a block and a deposit data root tree
 */
function getEth1Data(block: Block, depositDataRootTree: TreeBacked<Root<DepositData>[]>): {timestamp: number, eth1Data: Eth1Data};
```

[...]

PR #1179, opened 2020-07-17
Enforce no-floating-promises to prevent unhandled rejections

### LINKED ISSUE #1063: await async functions to prevent unhandled rejections
As of now, some expected errors such as a wrongly configured deposit contract cause an unhandled rejection. It would be best to handle this as expected errors by catching them at the CLI level and then do a clean process exit.
```
2020-06-26 16:14:13  [NODE]             info: Starting eth2 beacon node - LODESTAR! 
2020-06-26 16:14:13  [DB]               info: Connected to LevelDB database at .lodestar/dev/beacon/db-16Uiu2HAmLXTUhNJZgYayrQc2EkRLbrHU49EqFfiE8pffrzJchVDq 
2020-06-26 16:14:13  [CHAIN]            info: Chain not started, listening for genesis block 
2020-06-26 16:14:13  [ETH1]             info: Eth1 notifier is disabled but starting it to build genesis state 
(node:14720) UnhandledPromiseRejectionWarning: Error: There is no deposit contract at given address: TBD
    at EthersEth1Notifier.initContract (/home/lion/Code/eth2.0/lodestar/packages/lodestar/src/eth1/impl/ethers.ts:247:13)
    at processTicksAndRejections (internal/process/task_queues.js:94:5)
    at EthersEth1Notifier.startProcessEth1Blocks (/home/lion/Code/eth2.0/lodestar/packages/lodestar/src/eth1/impl/ethers.ts:265:7)
(node:14720) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:14720) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
However, in this case doing a try / catch won't save the unhandled rejection because the causing function is not awaited
https://github.com/ChainSafe/lodestar/blob/e247acf95db46133d6766458c3414ea77dc393f7/packages/lodestar/src/eth1/impl/ethers.ts#L94-L95

There are other instances of this happening in the codebase, detected with the rule [@typescript-eslint/no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md)

```
@chainsafe/lodestar-validator: 
@chainsafe/lodestar-validator: /home/lion/Code/eth2.0/lodestar/packages/lodestar-validator/src/services/attestation.ts
@chainsafe/lodestar-validator:   66:5  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar-validator: 
@chainsafe/lodestar-validator: /home/lion/Code/eth2.0/lodestar/packages/lodestar-validator/src/services/block.ts
@chainsafe/lodestar-validator:   57:5  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar-validator: 
@chainsafe/lodestar-validator: /home/lion/Code/eth2.0/lodestar/packages/lodestar-validator/src/validator.ts
@chainsafe/lodestar-validator:   65:5  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar-validator:   66:5  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar-validator: 
@chainsafe/lodestar: 
@chainsafe/lodestar: /home/lion/Code/eth2.0/lodestar/packages/lodestar/src/api/rest/routes/validator/publishAggregateAndProof.ts
@chainsafe/lodestar:   25:11  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar: 
@chainsafe/lodestar: /home/lion/Code/eth2.0/lodestar/packages/lodestar/src/chain/blocks/processor.ts
@chainsafe/lodestar:   64:5  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar: 
@chainsafe/lodestar: /home/lion/Code/eth2.0/lodestar/packages/lodestar/src/chain/chain.ts
@chainsafe/lodestar:   129:5  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar:   233:5  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar:   281:5  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar: 
@chainsafe/lodestar: /home/lion/Code/eth2.0/lodestar/packages/lodestar/src/chain/genesis/genesis.ts
@chainsafe/lodestar:   66:5  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar: 
@chainsafe/lodestar: /home/lion/Code/eth2.0/lodestar/packages/lodestar/src/eth1/impl/ethers.ts
@chainsafe/lodestar:    95:5  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar:   278:5  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar: 
@chainsafe/lodestar: /home/lion/Code/eth2.0/lodestar/packages/lodestar/src/network/network.ts
@chainsafe/lodestar:   58:7  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar: 
@chainsafe/lodestar: /home/lion/Code/eth2.0/lodestar/packages/lodestar/src/node/nodejs.ts
@chainsafe/lodestar:   151:5  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar: 
@chainsafe/lodestar: /home/lion/Code/eth2.0/lodestar/packages/lodestar/src/tasks/index.ts
@chainsafe/lodestar:   62:5  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar:   63:5  error  Promises must be handled appropriately  @typescript-eslint/no-floating-promises
@chainsafe/lodestar: 
```

I would like to discuss what's the best route to deal with this issue
1. Make sure that all error throwing async functions are awaited so the error can be catched upstream
2. It's okay to have unhandled rejection warnings
3. Handle the [event](https://nodejs.org/api/process.html#process_event_unhandledrejection) at the process level (not recommended)

dapplion:
This PR adds the rule @typescript-eslint/no-floating-promises and fixes its errors.

This PR is opened as a draft because it requires to add await to places I'm not sure if it can extend the duration of certain functions beyond what is expected. This applies to most of the comments around `start` functions below. Please review my comments below.

Fixes https://github.com/ChainSafe/lodestar/issues/1063.


packages/lodestar-validator/src/services/block.ts
```diff
@@ -55,7 +55,8 @@ export default class BlockProposingService {
   public start = async (): Promise<void> => {
     //trigger getting duties for current epoch
     const slot = this.provider.getCurrentSlot();
-    this.onNewEpoch(computeEpochAtSlot(this.config, slot));
+    // ### Todo: Unknown promise duration
+    await this.onNewEpoch(computeEpochAtSlot(this.config, slot));
```
dapplion: What are the consequences of awaiting this function in the validator start sequence?

packages/lodestar-validator/src/validator.ts
```diff
@@ -56,15 +56,17 @@ export class Validator {
...
     this.apiClient.onNewEpoch(this.blockService.onNewEpoch);
     this.apiClient.onNewSlot(this.attestationService.onNewSlot);
     this.apiClient.onNewEpoch(this.attestationService.onNewEpoch);
-    this.blockService.start();
-    this.attestationService.start();
-  };
+    // ### Todo: Unknown promise duration
+    await this.blockService.start();
+    // ### Todo: Unknown promise duration
+    await this.attestationService.start();
```
dapplion: What are the consequences of awaiting this function in the validator start sequence?

packages/lodestar/src/chain/blocks/processor.ts
```diff
@@ -58,7 +58,8 @@ export class BlockProcessor implements IService {
 
   public async start(): Promise<void> {
     const abortSignal = this.controller.signal;
-    pipe(
+    // ### Todo: Unknown promise duration
+    await pipe(
```
dapplion: What are the consequences of awaiting this function in the node start sequence?

packages/lodestar/src/eth1/impl/ethers.ts
```diff
@@ -101,7 +101,8 @@ export class EthersEth1Notifier implements IEth1Notifier {
     }
     this.eth1Source = pushable<Eth1EventsBlock>();
     // no need await
-    this.startProcessEth1Blocks();
+    // ### Todo: Unknown promise duration
+    await this.startProcessEth1Blocks();
```
dapplion: What is the reason of the `no need await` comment? What happens if this is awaited?

packages/lodestar/src/node/nodejs.ts
```diff
@@ -148,7 +148,7 @@ export class BeaconNode {
     // eth1 is started in chain
     await this.chain.start();
     await this.network.start();
-    this.sync.start();
+    await this.sync.start();
```
dapplion: Not awaiting the sync was a typo or there was a reason?

packages/lodestar-validator/src/validator.ts
mpetrunic: I don't think they are any but you should do Promise.all to start them at once or we could miss first attestation

packages/lodestar/src/chain/blocks/processor.ts
mpetrunic: You would break chain start function(it's waiting for this method to end which will never happen). This is endless loop (it will only end if stop method is called)

packages/lodestar/src/node/nodejs.ts
mpetrunic: sync start method returns when we are in sync. We need some sync initialization for api and chores I think

packages/lodestar/src/eth1/impl/ethers.ts
mpetrunic: again endless loop, it would never resolved so this method wouldn't return eth1 block generator

packages/lodestar-validator/src/services/block.ts
mpetrunic: I don't think they are any

packages/lodestar-validator/src/validator.ts
```diff
@@ -56,15 +56,17 @@ export class Validator {
     this.apiClient.once("beaconChainStarted", this.run);
   }
 
-  public run = (): void => {
+  public async run(): Promise<void> {
```
mpetrunic: Careful, this could break some tests as we loose class context.
you could do  `public run = async (): Promise<void> => {` as I think this was intentional

packages/lodestar/src/eth1/impl/ethers.ts
wemeetagain: These cases could be refactored to something like:
```typescript
start(): Promise<void> {
  this.longRunningThingTimeout = setTimeout(this.longRunningThing.bind(this), N);
}
stop(): Promise<void> {
  // for if stop is called before longRunningThing starts
  cancelTimeout(this.longRunningThingTimeout);
  // for if stop is called after longRunningThing starts
  await this.abortLongRunningThing();
}
```

dapplion: @wemeetagain But this approach will create the same situation of unhandled rejected promises if `longRunningThing` throws inside of a setTimeout

packages/lodestar-validator/src/validator.ts
dapplion: What is the practical difference in javascript between `run = () => {}` and `run() {}`?

mpetrunic: When you call somewhere setTimeout(this.run, 0) you loose `this` (class) context when using run() {}.
run = () =.> {} is basically run.bind(this)()

packages/lodestar/src/eth1/impl/ethers.ts
wemeetagain: I guess longRunningThing should explicitly have a try/catch.

I'm just thinking of our needs this way:
- coherent/standard strategy for handling daemon-like subprocesses across modules
- standard behavior for start/stop across modules

packages/lodestar-validator/src/validator.ts
wemeetagain: We're using the `run = () => {}` pattern for event handlers.

eg:
```typescript
run = () => {
  this.foo();
}

start() {
  this.on('bar', this.run);
}
stop() {
  this.off('bar', this.run);
}
```

If you just use `.bind(this)` for the event handler, it causes a memory leak, since each call to `.bind` creates a new `function`.
eg
```typescript
run() {
  this.foo();
}

start() {
  this.on('bar', this.run.bind(this));
}
stop() {
  // this.run.bind(this) is a new function, not the same as the one above
  this.off('bar', this.run.bind(this));
}
```

packages/lodestar/src/eth1/impl/ethers.ts
dapplion: If this function never resolves (or takes a long time) it means that the `start` method never resolves either?

https://github.com/ChainSafe/lodestar/blob/bd8adf267acebd0a93efa693be8fcec538384431/packages/lodestar/src/eth1/impl/ethers.ts#L80-L85

 If yes, then `Chain.start()` will never resolve either?
https://github.com/ChainSafe/lodestar/blob/bd8adf267acebd0a93efa693be8fcec538384431/packages/lodestar/src/chain/chain.ts#L137
https://github.com/ChainSafe/lodestar/blob/bd8adf267acebd0a93efa693be8fcec538384431/packages/lodestar/src/chain/chain.ts#L152-L153

wemeetagain: right, this specific case might be resolved with #1103

dapplion: Got it. But for now, how doesn't this stall the node's start process?

packages/lodestar/src/node/nodejs.ts
dapplion: Related to the discussion above, seems that `sync.start()` won't resolve until Eth1 blocks finish processing. Is this correct?

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -65,7 +65,7 @@ export class AttestationService {
   public start = async (): Promise<void> => {
     const slot = this.provider.getCurrentSlot();
     //trigger getting duties for current epoch
-    this.onNewEpoch(computeEpochAtSlot(this.config, slot) - 1);
+    await this.onNewEpoch(computeEpochAtSlot(this.config, slot) - 1);
```
mpetrunic: We could use this
```suggestion
    void this.onNewEpoch(computeEpochAtSlot(this.config, slot) - 1);
```
to intentionally ignore result.

See comment:
https://github.com/Microsoft/TypeScript/issues/13376#issuecomment-273289748

packages/lodestar/src/chain/blocks/processor.ts
```diff
@@ -58,6 +58,8 @@ export class BlockProcessor implements IService {
 
   public async start(): Promise<void> {
     const abortSignal = this.controller.signal;
+    // An error in this pipe should panic the process, so it's not handled
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
     pipe(
```
mpetrunic:
```suggestion
   void pipe(
```

packages/lodestar/src/node/nodejs.ts
mpetrunic: hmm that shouldn't be connected but this await will prevent api and chores from starting which should be fatal

dapplion: Then what should we do in this case? IMO the root issue here is that `sync.start()` should resolve once the essential initialization is done

packages/lodestar-validator/src/services/attestation.ts
dapplion: What's the issue with using `await`? Don't we want to await `onNewEpoch`?

packages/lodestar/src/chain/blocks/processor.ts
dapplion: That's some crazy (and nice) javascript voodoo magic. For a new-comer I'm not sure what's more confusing `void` or `eslint-disable-next-line @typescript-eslint/no-floating-promises`

dapplion:
An important question with this PR is: 

> What do we want to do when a long-running process (daemon) throws?

As of now most NodeJS versions and browsers log unhandled rejections but don't stop the main process. There should be some way for this long-running process to tell the main process to panic. But, how can we do that in a browser & NodeJS friendly way?

packages/lodestar/src/eth1/impl/ethers.ts
wemeetagain: It does stall the nodes start process. IMO the takeaway of this PR should be:
1. Determine root causes of problems here (eg: needing an error-handling strategy for long-running processes)
2. Document problem areas in the code (there are a handful here)
3. Eslint rule in place to prevent further unhandled rejections

In the problem cases, just add an eslint-disable line for now and document the specific case. Don't try to add hacks that just mask the problem, it won't really get us towards our goal. A real solution is bigger than can be solved here, likely involves refactoring in some small or large way.

packages/lodestar/src/node/nodejs.ts
wemeetagain: We should refactor the sync module to respect the "start should resolve quickly" interface.
In this PR, just eslint-disable and document.

packages/lodestar/src/eth1/impl/ethers.ts
dapplion: Opened followup issue https://github.com/ChainSafe/lodestar/issues/1196

packages/lodestar/src/node/nodejs.ts
dapplion: Opened follow up issue https://github.com/ChainSafe/lodestar/issues/1196

packages/lodestar-validator/src/services/block.ts
mpetrunic: Well if attestation is on first slot, start method would wait quite a lot of time to resolve. Errors in those methods ain't suppose to propagate up. it should be handled and this service should just loop and wiat for next duty

packages/lodestar-validator/src/services/attestation.ts
mpetrunic: Well if attestation is on first slot, start method would wait quite a lot of time to resolve. Errors in those methods ain't suppose to propagate up. it should be handled and this service should just loop and wait for next duty

packages/lodestar/src/chain/blocks/processor.ts
```diff
@@ -58,7 +58,10 @@ export class BlockProcessor implements IService {
 
   public async start(): Promise<void> {
     const abortSignal = this.controller.signal;
-    pipe(
+    // An error in this pipe should panic the process, so it's not handled
+    // void indicates that a function is not awaited on purpose
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
mpetrunic, line 62: you don't need this eslint ignore. Also error in pipe should't panic process :slightly_smiling_face:

dapplion: I've tried without `eslint-disable-next-line @typescript-eslint/no-floating-promises` and the rule still kicks in, seems it doesn't accept `void` as a valid promise handler

mpetrunic: Try to explicitly set in eslint rule config:
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md#options

dapplion: @wemeetagain I thought it was said that pipe had to indeed let the error bubble up.

wemeetagain: I'm not sure that its accurate to say it panics the process? It'll just be an unhandled promise rejection, right?

In any case, if this pipe errors out, its not great. We stop processing any new blocks.

We might just add a TODO / issue to have more robust error handling here. Either with better error handling inside each segment of the pipeline, or surrounding the pipeline.
The current state of things is: the pipeline doesn't error in practice because there's input validation pre-pipeline. So this is more a hardening / stylistic issue than an urgent practical issue.

dapplion:
Related https://github.com/ChainSafe/lodestar/issues/1297

dapplion:
@wemeetagain PR appears ready to review and merge

wemeetagain:
I ran this branch (with master merged) and it didn't cause any noticeable regressions

wemeetagain (approved):

mpetrunic (approved):

PR #1186, opened 2020-07-20
Fix "Received block parent root doesn't match our head"

### LINKED ISSUE #1183: Altona sync: Received block parent root doesn't match our head
**Describe the bug**
Initial sync is suspended with:
```
2020-07-19 13:54:31  [SYNC]             warn: Received block parent root doesn't match our head head=0x9d388a55e7b865f8bf10cfbaf1312d94a91fd9a9d67b1bb6fa6505dfe91fd1a3, headSlot=142161, blockParent=0x729ab4115fd5c7949b90e24edfb4a92f9092b2714eba7570ca2e0ac03a6b269b, blockSlot=142163
```

Previous log:
```
2020-07-19 13:54:27  [SYNC]             info: Fetching blocks for 142131...142162 slot range
...
2020-07-19 13:54:29  [CHAIN]            info: Processed new chain head newChainHeadRoot=0x9d388a55e7b865f8bf10cfbaf1312d94a91fd9a9d67b1bb6fa6505dfe91fd1a3, slot=142144, epoch=4442
2020-07-19 13:54:29  [SYNC]             info: Sync progress - currentEpoch=4442, targetEpoch=4442, speed=10.1 slots/s, estimateTillComplete=0 hours
2020-07-19 13:54:29  [SYNC]             info: Fetching blocks for 142162...142225 slot range
2020-07-19 13:54:29  [SYNC]          verbose: updating block target 142225
...
2020-07-19 13:54:30  [CHAIN]            info: Processed new chain head newChainHeadRoot=0x729ab4115fd5c7949b90e24edfb4a92f9092b2714eba7570ca2e0ac03a6b269b, slot=142161, epoch=4442
...
2020-07-19 13:54:30  [CHAIN]            info: Processed new chain head newChainHeadRoot=0x729ab4115fd5c7949b90e24edfb4a92f9092b2714eba7570ca2e0ac03a6b269b, slot=142161, epoch=4442
2020-07-19 13:54:31  [SYNC]             info: Imported blocks for slots: 142163,142164,142165,142166,142168,142169,142170,142172,142173,142176,142177,142178,142179,142181,142182,142183,142184,142187,142188,142190,142192,142193,142194,142195,142196,142197,142199,142200,142201,142203,142204,142206,142207,142210,142211,142213,142214,142215,142216,142217,142218,142219,142220,142221,142224,142225
```

**Analyse**
+ Block 142163 has parent root at 0x729ab4115fd5c7949b90e24edfb4a92f9092b2714eba7570ca2e0ac03a6b269b which is block 142161 and it's a correct block
+ But our calculated head is somehow not correct and it points to an obsolete block  142144 which we processed already in previous round
+ It seems that we fetch 142162...142225 range too early right after block 142144 (should be after block 142161) when the previous block range process were not finished

twoeths:
## Goals
+ resolves #1183 
+ Get through initial sync with this PR

image

### DIFF
--- a/packages/lodestar/src/network/gossip/gossip.ts
+++ b/packages/lodestar/src/network/gossip/gossip.ts
@@ -70,7 +70,7 @@ export class Gossip extends (EventEmitter as { new(): GossipEventEmitter }) impl
     this.registerHandlers(this.chain.currentForkDigest);
     this.chain.on("forkDigest", this.handleForkDigest);
     this.emit("gossip:start");
-    this.logger.info("Gossip is started");
+    this.logger.verbose("Gossip is started");
   }
 
   public async stop(): Promise<void> {
@@ -78,6 +78,7 @@ export class Gossip extends (EventEmitter as { new(): GossipEventEmitter }) impl
     this.unregisterHandlers();
     this.chain.removeListener("forkDigest", this.handleForkDigest);
     await this.pubsub.stop();
+    this.logger.verbose("Gossip is stopped");
   }
 
   public publishBlock = publishBlock.bind(this);
--- a/packages/lodestar/src/sync/initial/fast/index.ts
+++ b/packages/lodestar/src/sync/initial/fast/index.ts
@@ -135,6 +135,7 @@ export class FastSync
         const setBlockImportTarget = this.setBlockImportTarget;
         const updateBlockImportTarget = this.updateBlockImportTarget;
         const getInitialSyncPeers = this.getInitialSyncPeers;
+        const forkChoice = this.chain.forkChoice;
         return (async function() {
           for await (const slotRange of source) {
             const lastSlot = await pipe(
@@ -142,13 +143,19 @@ export class FastSync
               fetchBlockChunks(
                 logger, chain, network.reqResp, getInitialSyncPeers, opts.blockPerChunk
               ),
-              processSyncBlocks(config, chain, logger, true)
+              processSyncBlocks(config, chain, logger, true, true)
             );
-            logger.verbose("last fetched slot=" + lastSlot);
+            logger.verbose("last processed slot=" + lastSlot);
             if(lastSlot) {
-              //set new target from last block we've received
-              // it will trigger new sync once last block is processed
-              updateBlockImportTarget(lastSlot);
+              if (lastSlot === forkChoice.headBlockSlot()) {
+                // failed at start of range
+                logger.warn(`Failed to process range, retry fetching range, lastSlot=${lastSlot}`);
+                setBlockImportTarget(lastSlot);
+              } else {
+                //set new target from last block we've processed
+                // it will trigger new sync once last block is processed
+                updateBlockImportTarget(lastSlot);
+              }
             } else {
               logger.warn("Didn't receive any valid block in given range");
               //we didn't receive any block, set target from last requested slot
--- a/packages/lodestar/src/sync/utils/sync.ts
+++ b/packages/lodestar/src/sync/utils/sync.ts
@@ -155,36 +155,41 @@ export function validateBlocks(
  * @param config
  * @param chain
  * @param logger
+ * @param isInitialSync
  * @param trusted
  */
 export function processSyncBlocks(
-  config: IBeaconConfig, chain: IBeaconChain, logger: ILogger, trusted = false
+  config: IBeaconConfig, chain: IBeaconChain, logger: ILogger, isInitialSync: boolean, trusted = false
 ): (source: AsyncIterable<SignedBeaconBlock[]>) => Promise<Slot|null> {
   return async (source) => {
     let blockBuffer: SignedBeaconBlock[] = [];
-    let headRoot = chain.forkChoice.headBlockRoot();
     let lastProcessedSlot: Slot|null = null;
     for await (const blocks of source) {
       logger.info("Imported blocks for slots: " + blocks.map((block) => block.message.slot).join(","));
       blockBuffer.push(...blocks);
     }
     blockBuffer = sortBlocks(blockBuffer);
+    let headRoot = chain.forkChoice.headBlockRoot();
+    let headSlot = chain.forkChoice.headBlockSlot();
     while(blockBuffer.length > 0) {
-      const block = blockBuffer.shift();
-      if(!trusted || (trusted && config.types.Root.equals(headRoot, block.message.parentRoot))) {
-        await chain.receiveBlock(block, trusted);
-        headRoot = config.types.BeaconBlockHeader.hashTreeRoot(blockToHeader(config, block.message));
-        if(block.message.slot > lastProcessedSlot) {
-          lastProcessedSlot = block.message.slot;
+      const signedBlock = blockBuffer.shift();
+      const block = signedBlock.message;
+      if(!isInitialSync ||
+        (isInitialSync && block.slot > headSlot && config.types.Root.equals(headRoot, block.parentRoot))) {
+        await chain.receiveBlock(signedBlock, trusted);
+        headRoot = config.types.BeaconBlockHeader.hashTreeRoot(blockToHeader(config, block));
+        headSlot = block.slot;
+        if(block.slot > lastProcessedSlot) {
+          lastProcessedSlot = block.slot;
         }
       } else {
         logger.warn(
           "Received block parent root doesn't match our head",
           {
             head: toHexString(headRoot),
-            headSlot: chain.forkChoice.headBlockSlot(),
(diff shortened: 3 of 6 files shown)


packages/lodestar/src/sync/initial/fast/index.ts
```diff
@@ -96,14 +96,13 @@ export class FastSync
     return computeStartSlotAtEpoch(this.config, this.targetCheckpoint.epoch);
   }
 
-  private getNewBlockImportTarget(fromSlot?: Slot): Slot {
-    const headSlot = fromSlot || this.chain.forkChoice.headBlockSlot();
+  private getNewBlockImportTarget(fromSlot: Slot): Slot {
```
twoeths: this is called from one single place and it had param there so this is just simply remove the optional thing

packages/lodestar/src/sync/initial/fast/index.ts
```diff
@@ -190,9 +189,10 @@ export class FastSync
       );
       if(newTarget.epoch > this.targetCheckpoint.epoch) {
         this.targetCheckpoint = newTarget;
-        this.setBlockImportTarget();
```
twoeths: This is to fix the `Received block parent root doesn't match our head` issue. This call would cause the next block range fetch while we're in the middle of current range.

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -171,7 +171,7 @@ export function processSyncBlocks(
     blockBuffer = sortBlocks(blockBuffer);
     while(blockBuffer.length > 0) {
       const block = blockBuffer.shift();
-      if(config.types.Root.equals(headRoot, block.message.parentRoot)) {
+      if(!trusted || (trusted && config.types.Root.equals(headRoot, block.message.parentRoot))) {
```
twoeths: for regular sync, we should not check for the `headRoot` since we're not sure block `slot` is parent of block `slot + 1`

packages/lodestar/src/network/gossip/gossip.ts
```diff
@@ -70,6 +70,7 @@ export class Gossip extends (EventEmitter as { new(): GossipEventEmitter }) impl
     this.registerHandlers(this.chain.currentForkDigest);
     this.chain.on("forkDigest", this.handleForkDigest);
     this.emit("gossip:start");
+    this.logger.info("Gossip is started");
```
wemeetagain: Can you make this verbose?
Also add "Gossip stopped"

packages/lodestar/src/sync/initial/fast/index.ts
mpetrunic: hmm but it won't trigger to sync forward if detects new finalized checkpoint (sync could stop now if new finalized epoch shows up while we are syncing current one)?
`Received block parent root doesn't match our head` - I don't think this happens because we are fetching next block range in middle of current one because thos functions are executed sequentially

packages/lodestar/src/sync/utils/sync.ts
mpetrunic: I think for initial sync we should also check if received block slot is greater than head block slot.

Also as a backup, if we get  `Received block parent root doesn't match our head` we should set lastBlockTarget to current head slot, so we try to refetch that range in case some peer sent response with some missing block

packages/lodestar/src/sync/initial/fast/index.ts
twoeths: > hmm but it won't trigger to sync forward if detects new finalized checkpoint (sync could stop now if new finalized epoch shows up while we are syncing current one)?

if it detects new finalized checkpoint, it just set new `targetCheckpoin`, `checkSyncProgress` will always trigger the next sync

> I don't think this happens because we are fetching next block range in middle of current one because thos functions are executed sequentially

this is how it happens: I have `maxSlotImport` as 64 = 2 * SLOTS_PER_EPOCH. 
+ It fetch and process 142131...142162 `Fetching blocks for 142131...142162 slot range`
+ At slot 142144 (the middle of the range), chain processor detects new epoch and it emits `"processedCheckpoint"`
+ Then `checkSyncCompleted` called and it called `setBlockImportTarget` and it triggered a new block range as in the log `Fetching blocks for 142162...142225 slot range`, although the range 142131...142162 not finished yet

I think this is quite clear as we can see in the issue description logs #1183

mpetrunic: I guess, that's odd since since it iterates sequentially it cannot skip stuff, it can start fetching blocks for next batch but they ain't suppose to be processed before first batch.

in #1183 it says in logs we processed 142161 but call to head returns different one, so it apears we have some race condition between block post process and this  processSyncBlocks. We could call method for getting headRoot a bit later in processSyncBlocks method though (after we receive all blocks in range)

twoeths: > in #1183 it says in logs we processed 142161 but call to head returns different one

that's because in `processSyncBlocks` we have `let headRoot = chain.forkChoice.headBlockRoot();` before `for await` so that `headRoot` became obsolete. The other way to fix the issue is to just have that call after `for await` to make sure we get up-to-date value, does that sound good to you?

mpetrunic: yeah, that was my suggestion too.

Still not sure about removing `this.setBlockImportTarget();` here. `checkSyncProgress` will most likely be called before `checkSyncCompleted` so it won't trigger new sync loop

twoeths: in this case, `checkSyncProgress` is not called before `checkSyncCompleted` because `checkSyncProgress` is called at the end of batch (which is 2 epochs) while `checkSyncProgress` is called in the middle (at end of 1st epoch)
+ we get batch 142131...142162
+ at 142144 `checkSyncCompleted` is called in the middle of the batch because it's new epoch
+ at 142162 `checkSyncProgress` is called (142131 + 2 * SLOTS_PER_EPOCH)

twoeths: if we configure `maxSlotImport = 1*SLOTS_PER_EPOCH` then yes, `checkSyncProgress` will most likely be called before `checkSyncCompleted`

mpetrunic: Yeah, but consider the case where we are one epoch from targeted finalized epoch. We would fetch missing blocks, call checkSyncProgress on last finalized block, then call checkSyncCompleted where new finalized target is detected from peers and sync would stall?

twoeths: > then call checkSyncCompleted where new finalized target is detected from peers 

if `newTarget epoch > targetCheckpoint.epoch` then we just set new targetCheckpoint, sync will continue for 1 more round
otherwise we stop the initial sync and initial sync is stopped and we transition to regular sync

packages/lodestar/src/sync/utils/sync.ts
wemeetagain: We can't use `trusted` to determine if this is used in initial vs regular sync. Bc we need to be running with trusted=false during initial sync too.

packages/lodestar/src/sync/initial/fast/index.ts
twoeths: > Still not sure about removing this.setBlockImportTarget(); here. checkSyncProgress will most likely be called before checkSyncCompleted so it won't trigger new sync loop

from what I see, `checkSyncProgress` will always trigger new sync loop regardless of `targetCheckpoint`

packages/lodestar/src/sync/initial/fast/index.ts
```diff
@@ -135,20 +135,27 @@ export class FastSync
...
-              processSyncBlocks(config, chain, logger, true)
+              processSyncBlocks(config, chain, logger, true, true)
             );
-            logger.verbose("last fetched slot=" + lastSlot);
+            logger.verbose("last processed slot=" + lastSlot);
             if(lastSlot) {
-              //set new target from last block we've received
-              // it will trigger new sync once last block is processed
-              updateBlockImportTarget(lastSlot);
+              if (lastSlot === forkChoice.headBlockSlot()) {
```
twoeths, line 150: handle the case when the processSyncBlocks fails at 1st range slot which cause the initial sync suspends

twoeths:
I used to be able to sync to head but with a new fresh sync, I can only get through initial sync and regular sync and receive a gossip block that > 10 blocks in the future, so there's a gap between regular sync's final target and gossip block.

Will test more and create a separate issue for it.

wemeetagain (approved):

mpetrunic (approved):

PR #1192, opened 2020-07-22
Remove multiaddr in ENR if its not local network interface

### LINKED ISSUE #1141: Altona: Cannot detect new peers upon restart
**Describe the bug**
After we initialize a beacon folder and sync Altona, we can always detect peers very quickly. But upon a restart, sometimes we can detect peers, sometimes we cannot and we end up with setting up a new beacon folder by `beacon run init`.

**Expected behavior**
We can detect new peers at any time

twoeths:
resolves #1141 

image

### DIFF
--- a/packages/lodestar/src/network/nodejs/util.ts
+++ b/packages/lodestar/src/network/nodejs/util.ts
@@ -9,6 +9,8 @@ import LibP2p from "libp2p";
 import {NodejsNode} from ".";
 import {INetworkOptions} from "../options";
 import defaults from "../defaults";
+import {isLocalMultiAddr, clearMultiaddrUDP} from "..";
+import {ENR} from "@chainsafe/discv5";
 
 /**
  * Save a peer id to disk
@@ -39,6 +41,13 @@ export async function createNodeJsLibp2p(
   const peerId = await Promise.resolve(peerIdOrPromise);
   const multiaddrs = network.multiaddrs || defaults.multiaddrs;
   const bootnodes = network.bootnodes || defaults.bootnodes;
+  const enr = network.discv5?.enr;
+  if (enr && typeof enr !== "string") {
+    const enrInstance = enr as ENR;
+    if (enrInstance.multiaddrUDP && !isLocalMultiAddr(enrInstance.multiaddrUDP)) {
+      clearMultiaddrUDP(enrInstance);
+    }
+  }
   return new NodejsNode({
     peerId,
     addresses: {listen: multiaddrs},
--- a/packages/lodestar/src/network/util.ts
+++ b/packages/lodestar/src/network/util.ts
@@ -5,11 +5,13 @@
 
 import PeerId from "peer-id";
 import {Type} from "@chainsafe/ssz";
+import AbortController from "abort-controller";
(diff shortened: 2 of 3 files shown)


packages/lodestar-cli/src/cmds/beacon/cmds/run/run.ts
```diff
@@ -24,28 +25,26 @@ export async function run(options: Arguments<IBeaconArgs & Partial<IBeaconNodeOp
...
+  const logger = new WinstonLogger();
   const peerId = await readPeerId(options.network.peerIdFile);
   // read local enr from disk
-  options.network.discv5.enr = await readEnr(options.network.enrFile);
+  const enr = await readEnr(options.network.enrFile);
+  if (enr.multiaddrUDP && !isLocalMultiAddr(enr.multiaddrUDP)) {
+    logger.info(`MultiAddress of ${enr.multiaddrUDP.toString()} is not part of local network interface so removed it`);
+    clearMultiaddrUDP(enr);
+  }
+  options.network.discv5.enr = enr;
```
mpetrunic: How come we don't handle this in network module?

packages/lodestar-cli/src/cmds/beacon/cmds/run/util.ts
```diff
@@ -0,0 +1,37 @@
+import Multiaddr from "multiaddr";
+import {networkInterfaces} from "os";
+import {ENR} from "@chainsafe/discv5";
+
+/**
+ * Check if multiaddr belongs to the local network interfaces.
+ */
+export function isLocalMultiAddr(multiaddr: Multiaddr | undefined): boolean {
```
mpetrunic: Feels like we should put this in lodestar/network/utils

packages/lodestar-cli/src/cmds/beacon/cmds/run/run.ts
twoeths: I think it's just more clear to put it here because we want to handle this logic at startup time.
what do you think @wemeetagain ?

dapplion: This fix changes should not apply to the validator when it's run by a different consumer than our CLI?

wemeetagain: I'm not sure we want the same behavior with a browser node, at least, not the _exact_ same behavior.
(eg: we might want to explicitly set a relay IP, or have a more intricate ENR initialization procedure in that case)

I think we can leave this here for now, or maybe move it to `createNodeJsLibp2p`?

wemeetagain: I guess a bigger issue is figuring out a strategy for splitting network agnostic code from nodejs / browser specific code. We have done some work towards that end, eg: with a nodejs-specific libp2p bundle, but it there likely will be a need for better organization.

mpetrunic (dismissed):

wemeetagain:
file conflict

wemeetagain (approved):

PR #1194, opened 2020-07-22
Eth2stats minimal api

### LINKED ISSUE #1154: Add minimal api for eth2 stats

mpetrunic:
- this implements minimal api needed for eth2stats client

resolves #1154


wemeetagain (approved):
lgtm

packages/lodestar-validator/src/api/impl/rest/beacon/beacon.ts
```diff
@@ -17,13 +17,15 @@ import {Json, toHexString} from "@chainsafe/ssz";
...
   private readonly client: HttpClient;
+  private readonly clientV2: HttpClient;
 
   private readonly logger: ILogger;
 
   private readonly config: IBeaconConfig;
 
   public constructor(config: IBeaconConfig, restUrl: string, logger: ILogger) {
     this.client = new HttpClient({urlPrefix: urlJoin(restUrl, "lodestar")}, {logger});
+    this.clientV2 = new HttpClient({urlPrefix: urlJoin(restUrl, "/v1/beacon")}, {logger});
```
dapplion, line 28: Should it be 
```
clientV1 = ... "/v1"
```
or 
```
clientV2 = ... "/v2"
```
?

packages/lodestar-validator/src/api/impl/rest/beacon/beacon.ts
```diff
@@ -39,12 +41,13 @@ export class RestBeaconApi implements IBeaconApi {
...
-  public async getGenesisTime(): Promise<Number64> {
+  public async getGenesis(): Promise<Genesis|null> {
     try {
-      return await this.client.get<Number64>("/genesis_time");
+      const genesisResponse = await this.clientV2.get<{data: Json}>("/genesis");
+      return this.config.types.Genesis.fromJson(genesisResponse.data, {case: "snake"});
     } catch (e) {
       this.logger.error("Failed to obtain genesis time", e);
-      return 0;
+      return null;
```
dapplion, line 50: Why return null instead of throwing the error? The spec allows returning 404 or 500 
https://github.com/ethereum/eth2.0-APIs/blob/1c6b9cb4ec8d8587f5739dd2d491a8742be89e48/apis/beacon/genesis.yaml#L25

dapplion (commented):
Great work, glad to be compatible with eth2stats :smile:

packages/lodestar-validator/src/api/impl/rest/beacon/beacon.ts
mpetrunic: We will drop first client. In future it will be just one haha

packages/lodestar-validator/src/api/impl/rest/beacon/beacon.ts
mpetrunic: I don't wanna throw rest errors here since we are using this directly in validator so we would have to catch and differentiate errors which leads to ugly code. 

We might also expose this over graphql or rpc where rest errors wouldn't make sense.

wemeetagain: Lion makes a good point. We will need to handle the error case appropriately, eventually. Right now, `pollBeaconNode` silently does _nothing_ if this errors.
We can tackle this in another PR, seems like that has been the behavior even before this PR :grimacing:

PR #1197, opened 2020-07-22
Add uncompressed pubkey caching

mpetrunic:


packages/lodestar-beacon-state-transition/src/fast/block/isValidIndexedAttestation.ts
```diff
@@ -40,9 +40,12 @@ export function isValidIndexedAttestation(
...
   const signingRoot = computeSigningRoot(config, config.types.AttestationData, indexedAttestation.data, domain);
-  return verifyAggregate(
-    pubkeys,
-    signingRoot,
-    indexedAttestation.signature.valueOf() as Uint8Array,
+  return pubkeys.reduce(
+    (aggregate, pubkey) => {
+      return aggregate.add(pubkey);
+    }, pubkeys.shift()
+  ).verifyMessage(
```
dapplion, line 47: What would happen here if pubkeys is an empty array?

packages/lodestar-beacon-state-transition/src/fast/block/processVoluntaryExit.ts
```diff
@@ -43,10 +43,10 @@ export function processVoluntaryExit(
   if (verifySignature) {
     const domain = getDomain(config, state, DomainType.VOLUNTARY_EXIT, voluntaryExit.epoch);
     const signingRoot = computeSigningRoot(config, config.types.VoluntaryExit, voluntaryExit, domain);
-    if (!verify(
-      validator.pubkey.valueOf() as Uint8Array,
+    const pubkey = epochCtx.index2pubkey[voluntaryExit.validatorIndex];
```
dapplion, line 46: Is it possible for `voluntaryExit.validatorIndex` to not exist in `epochCtx.index2pubkey`? If so, should we guard against it or not necessary?

dapplion (dismissed):

packages/lodestar-beacon-state-transition/src/fast/block/isValidIndexedAttestation.ts
mpetrunic: It would fail much earlier 😄 there is a check for that

packages/lodestar-beacon-state-transition/src/fast/block/processVoluntaryExit.ts
mpetrunic: Nope validator must exists for it to exit, state transition checks this before we check signature here

wemeetagain (dismissed):
Sweet!

wemeetagain:
Unit test failing

wemeetagain (approved):

PR #1200, opened 2020-07-22
lodestar-cli: chain.genesisStateFile alias fix

### LINKED ISSUE #1188: genesisStateFile not being copied over to beacon.config.json during beacon init
**Describe the bug**

when i run `yarn run cli beacon init --template-config-file altona.json --chain.genesisStateFile=genesis.ssz --rootDir .altona` (or if I omit the cli parameter and include chain.genesisStateFile as `"genesisStateFile": "genesis.ssz",` in my altona.json), I don't see the genesisStateFile parameter copied over in the resultant beacon.config.json

**Expected behavior**

the genesisStateFile parameter should be copied over to the resultant beacon.config.json both when passed as a cli option and when specified in the altona.json

3xtr4t3rr3str14l:
fixes #1188 

i noticed that the chain.genesisStateFile parameter for the yargs option only had one entry as it's `alias`.  yargs doesn't like it when you only provide one entry, apparently, in the case of the chain options.  so i added another alias to fix this (which is also what chain.preset has already done)


3xtr4t3rr3str14l:
maybe this issue is also related to the `canonicalOptions` issue described here?

https://github.com/ChainSafe/lodestar/pull/1149#issuecomment-657605782

mpetrunic (dismissed):
Yup it's related to that most likely

twoeths (dismissed):

dapplion (dismissed):

wemeetagain (changes requested):
:( I think we should just fix the underlying issue.

This fix will add a different name `chain.genesisFileName` as the "canonical option". In the case of chain.preset, we're adding two names because the second name, `chain.name` is what the BeaconNode requires. We're relying on yargs to create that entry to fulfill this req: https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/chain/chain.ts#L80

Here, the likely problem is that `canonicalOptions` is stripping away the first alias (and only) alias, which we need for another command. (Take a look at the linked comment We really should be looking at the fix described there)

If we do just want this cheap and ez solution, we should try pasting the same alias twice, eg:
```
 alias: [
    "chain.genesisStateFile",
    "chain.genesisStateFile",
  ],
```

dapplion:
This PR https://github.com/ChainSafe/lodestar/pull/1210 would address the root cause of this issue

wemeetagain:
closing in favor of fix in #1210

PR #1205, opened 2020-07-24
Regular sync until gossip parent block

### LINKED ISSUE #1201: Altona sync: Regular sync to head
**Describe the bug**
Until now we can transition from initial sync to regular sync, but cannot transition from regular sync to head where we can stop regular sync and receive gossip blocks.

```
2020-07-21 15:14:22  [SYNC]             info: Sync caught up to latest slot 157108
2020-07-21 15:15:06  [GOSSIP]        verbose: Incoming block at slot: 157125
```

twoeths:
resolves #1201 

+ Sync up to current slot and keep it up using `onNewSlot`
+ Once we see gossip block parent in our forkChoice, we can stop regular sync


twoeths:
@wemeetagain so I can consistently sync up to current slot but didn't see any gossip blocks in my test (although last week I did see). Do you have any ideas?
I see there are Aggregate_And_Proof gossip message so the `forkDigest` should be good. I didn't see any BLOCK messages even at LodestarGossipSub so it hasn't touched the gossip validator too.

wemeetagain:
I'm not sure what could have changed in the past week that could have stopped it :thinking: 
If it seems like an issue with gossipsub, you might run the cli prefixed with `DEBUG=libp2p:gossipsub ` to print gossipsub internal logs.

twoeths:
Today I can receive Gossip block, not sure why it did not work yesterday.
image

twoeths:
So I can sync to gossip blocks for a while then after a while gossip blocks cannot pass our validator due to incorrect proposer index
```
const supposedProposerIndex = epochCtx.getBeaconProposer(signedBlock.message.slot);
    if (supposedProposerIndex !== signedBlock.message.proposerIndex) {
      return ExtendedValidatorResult.reject;
    }
```
Will create a separate issue for it.

wemeetagain (approved):

PR #1208, opened 2020-07-24
Fix attestation  processing bug where we would use wrong state

mpetrunic:
Seems like I'm on a bug hunt today.

Fixed:
`warn: Failed to process attestation. Reason: crosslink committee retrieval: out of range epoch: 2`

This will speed up e2e tests as we would reach finalization sooner.

### DIFF
--- a/packages/lodestar/src/chain/attestation.ts
+++ b/packages/lodestar/src/chain/attestation.ts
@@ -1,12 +1,5 @@
 import {fromHexString, toHexString} from "@chainsafe/ssz";
-import {
-  Attestation,
-  AttestationRootHex,
-  BlockRootHex,
-  Root,
-  SignedBeaconBlock,
-  Slot
-} from "@chainsafe/lodestar-types";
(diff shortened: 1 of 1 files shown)


twoeths:
yeah that's why I just see it recently, quite annoying 💯

twoeths (approved):

mpetrunic:
> yeah that's why I just see it recently, quite annoying

I have no idea how we didn't noticed earlier that we were using epochctx of last justified state :sweat_smile:

wemeetagain (approved):

packages/lodestar/src/chain/attestation.ts
```diff
@@ -71,6 +63,7 @@ export class AttestationProcessor implements IAttestationProcessor {
     try {
       await this.processAttestation(attestation, attestationHash);
     } catch (e) {
+      console.log(e);
```
dapplion, line 66: @mpetrunic  Is this console.log meant to be here?

PR #1210, opened 2020-07-24
Refactor CLI option handling

dapplion:
This PR tackles various improvements to simplify the CLI option handling that wer not splittable between smaller PRs.

1. Remove `canonicalOptions` usage

For yargs to parse options whose name is a JSON path, `canonicalOptions` was used to assign their first alias as their name within the yargs options object. Now instead of declaring options individually, they are defined in an object where JSON paths are allowed as keys. This approach is more transparent and "common", but it does not guarantee that the JSON path of the variable matches its expected type.

```ts
export const metricsOptions = {
  "metrics.enabled": {
    type: "boolean",
    default: false,
    group: "metrics",
  } as Options
}
```

2. Compute paths in the handler to remove "dynamic" options

Beacon options were defined as functions so they could be computed dynamically and compute fs subpaths with the options provided by the user. Now beacon paths are computed in the handler with `getBeaconPaths` so the options can be static, following the same approach as the `account` commands

```
$rootDir
 └── $beaconDir
     ├── beacon.config.json
     ├── peer-id.json
     ├── enr.json
     └── chain-db
```

3. Reorg options

There where duplicated options between various commands. Now shared options can be found in `src/options` with the following groups

```
src/options
├── beaconNodeOptions/
├── globalOptions.ts
└── paramsOptions.ts
```

- `beaconNodeOptions/` includes options strictly part of the `BeaconNodeOptions` type.
- `paramsOptions` are options dynamically generated from the spec file
- `globalOptions`: Global options.

4. Simplify default parsing

Convert all "config" yargs Options from using `default` to `defaultDescription`. Yargs Options has a defaultDescription property which in practice makes that value show in `--help` as if it was a default, BUT it's not an actual default so we can know in the handler which options have been actually set by the user and which not. Then, at the handler we can easily merge default config with config files with CLI config.

```js
options = mergeConfigOptions(options);
```

Which does a deepmerge of `defaultOptions` > `templateConfigFile` > `configFile` > `userOptions`

5. Un-nest options outside of `BeaconNodeOptions`

Options not part of the actual type `BeaconNodeOptions` are moved to the root level of `options`:
- `options.chain.name` > `options.preset`
- `options.chain.params` > `options.params`
- `options.network.peerIdFile` > `options.peerIdFile`
- `options.network.enrFile` > `options.enrFile`


packages/lodestar-cli/src/cmds/beacon/cmds/init/init.ts
```diff
@@ -1,47 +1,49 @@
...
+import {initBeaconConfig, mergeConfigOptions} from "../../config";
 import {getTestnetConfig, downloadGenesisFile, fetchBootnodes} from "../../testnets";
+import {getBeaconPaths} from "../../paths";
 
 /**
  * Initialize lodestar-cli with an on-disk configuration
  */
-export async function init(args: Arguments<IBeaconArgs>): Promise<void> {
+export async function initHandler(options: IBeaconOptions): Promise<void> {
+  options = mergeConfigOptions(options);
```
wemeetagain: I don't think we want to merge from `configFile` here, since thats where we're writing out the configuration in `initBeaconConfig`.

wemeetagain (changes requested):
Looks so much better :pray: 
one minor issue

3xtr4t3rr3str14l:
nice!  you might want to double check, but i think this PR should resolve not only #1200 but also #1149 (and the issues referenced within them, of course)

packages/lodestar-cli/src/options/otherOptions.ts
```diff
@@ -0,0 +1,15 @@
+import {Options} from "yargs";
+
+export interface IGenesisStateFileOptions {
+  chain: {
+    genesisStateFile?: string;
+  };
+}
+
+export const genesisStateFileOptions = {
+  "chain.genesisStateFile": {
```
3xtr4t3rr3str14l: why not put this in `packages/lodestar-cli/src/options/beaconNodeOptions/chain.ts`?  what's the reasoning behind putting it in a separate file?

packages/lodestar-cli/src/cmds/beacon/index.ts
```diff
@@ -1,15 +1,19 @@
...
   command: "beacon <command>",
   describe: "Beacon node",
-  builder: (yargs) => mergeBeaconOptions(yargs)
-    .command(init)
-    .command(run),
+  builder: (yargs) => yargs
+    .options(beaconOptions)
+    // Deep type error with the command's options
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    .command(init as any)
```
3xtr4t3rr3str14l: maybe try replacing `any` with `CommandModule<IGlobalArgs, IBeaconOptions>`?

packages/lodestar-cli/src/cmds/beacon/index.ts
```diff
@@ -1,15 +1,19 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    .command(run as any),
```
3xtr4t3rr3str14l: same as above

packages/lodestar-cli/src/options/otherOptions.ts
dapplion: genesisStateFile is not actually part of the `BeaconNodeOptions` type. Mixing actual options of `BeaconNodeOptions` with random options scoped only to the CLI is confusing. The entire `options` object is passed to the BeaconNode constructor as if it was `BeaconNodeOptions`, so not merging this option into `beaconNodeOptions` attempts to make this nuance clearer

packages/lodestar-cli/src/cmds/beacon/index.ts
dapplion: Thanks works great!

packages/lodestar-cli/src/cmds/beacon/cmds/init/init.ts
dapplion: Right, actually `options = mergeConfigOptions(options)` could be not called at all in the init handler? All the options actually necessary are computed with defaults in `beaconPaths`. No option in `options` used needs to be assigned a default.

dapplion:
@wemeetagain the option `genesisStateFile` must it be a subproperty of `chain`? If it is not used outside of the CLI, why does it have to be mixed up inside part of the `BeaconNodeOptions`?

wemeetagain:
> @wemeetagain the option `genesisStateFile` must it be a subproperty of `chain`? If it is not used outside of the CLI, why does it have to be mixed up inside part of the `BeaconNodeOptions`?

:+1: for simplifying it. Doesn't need to be a subproperty of `chain`.

dapplion:
@wemeetagain Same for `network.peerIdFile`, `network.enrFile`. Can we evict them from being a sub-property of BeaconNodeOptions?

wemeetagain:
> @wemeetagain Same for `network.peerIdFile`, `network.enrFile`. Can we evict them from being a sub-property of BeaconNodeOptions?

Yeah sounds good to me. All of these CLI-only flags could be pulled out, so we aren't confusing them with `BeaconNodeOptions`

wemeetagain (dismissed):

dapplion:
Testing the command `lodestar beacon run --testnet altona` I got this error
```
lion@lion-gram:~/Code/eth2.0/lodestar/packages/lodestar-cli$ ts-node src beacon run --testnet altona
2020-07-26 19:53:55  []                error: uncaughtException: IO error: ./.tmp/lodestar-db/LOCK: No such file or directory
OpenError: IO error: ./.tmp/lodestar-db/LOCK: No such file or directory
    at /home/lion/Code/eth2.0/lodestar/node_modules/level-packager/node_modules/levelup/lib/levelup.js:119:23
    at /home/lion/Code/eth2.0/lodestar/node_modules/level-packager/node_modules/deferred-leveldown/node_modules/abstract-leveldown/abstract-leveldown.js:38:14
    at /home/lion/Code/eth2.0/lodestar/node_modules/level-packager/node_modules/deferred-leveldown/deferred-leveldown.js:31:21
    at /home/lion/Code/eth2.0/lodestar/node_modules/level-packager/node_modules/abstract-leveldown/abstract-leveldown.js:38:14
    at /home/lion/Code/eth2.0/lodestar/node_modules/leveldown/node_modules/abstract-leveldown/abstract-leveldown.js:38:14 
```
To fix it I had to manually create that directory `mkdir ./.tmp/lodestar-db`, then it worked fine.

@wemeetagain @mpetrunic What data is stored in this directory `lodestar-db`? Should it be under `./.tmp` or respect the path structure provided by the user? Maybe I'm not passing a parameter to set the DB path that was used before, but I can't figure out which.

**Edit**: Fixed with `Set options.db.name in beacon run cmd` https://github.com/ChainSafe/lodestar/pull/1210/commits/0e049e39cc7aa4a13f23bf0c4b27a19056343fb7

wemeetagain:
I think this is from the dev command? I can't find the file path, but found a reference to it in a local file:
`lodestar-cli/coverage/lcov-report/src/commands/dev/command.ts.html`

wemeetagain (approved):

mpetrunic (approved):

PR #1231, opened 2020-07-26
Review account docs comments

### LINKED ISSUE #1229: Documentation walkthrough bug reports
Ive been using the docs to run through setting up a lodestar node and validators and ill be documenting any gotchyas I find along the way.

## Install Issues
Environment: Digital Ocean - Ubuntu 18
Node: Installed via nvm v12
Bugs:
- Creating validator keys does not match the documentation, `$rootDir/validators` doesn't exist, and it seems there are no keys either.
```
> $yarn run cli account wallet create --name primary --passphrase-file ../password.pass 
Your wallet's 12-word BIP-39 mnemonic is:

[redacted]
[removed warning]

Done in 1.88s.

> $ tree .lodestar/
.lodestar/
└── wallets
    └── ec35c27f-329d-45ca-b58c-cc4bdb9ad03e
        └── ec35c27f-329d-45ca-b58c-cc4bdb9ad03e

> $ yarn run cli account validator create --name primary --passphrase-file ../password.pass 
Done in 1.95s.

> $ tree .lodestar/
.lodestar/
├── keystores
├── secrets
└── wallets
    └── ec35c27f-329d-45ca-b58c-cc4bdb9ad03e
        └── ec35c27f-329d-45ca-b58c-cc4bdb9ad03e

> $ yarn run cli account wallet list
primary
Done in 1.58s.

> $ yarn run cli account validator list

Done in 1.69s.
```

Gotchyas:
- `sudo apt install build-essential` was needed to remove random node-gyp errors

## Documentation
- The [create wallet](https://chainsafe.github.io/lodestar/usage/key-management/#create-a-wallet) section is a bit confusing, perhaps some clarity around the `--name` flag would be helpful to users. TBH i dont even understand what it is doing
- General confusion around what `$rootDir` means in the wallet section for `lodestar-cli wallet validator create`

dapplion:
Fixes https://github.com/ChainSafe/lodestar/issues/1229

- Fix account docs directory name holding validator keystores
- Don't use $rootDir in docs, but its default value
- Clarify the role of --name option in 'lodestar account wallet create'

### DIFF
--- a/docs/usage/key-management.md
+++ b/docs/usage/key-management.md
@@ -11,7 +11,7 @@ The 12-word string is randomly generated during wallet creation and printed out 
 To create a wallet, use the following command:
 
 ```
-lodestar account wallet create --name primary --passphrase-file primary.pass
+lodestar account wallet create --name primary --passphraseFile primary.pass --rootDir .lodestar
 ```
 
 This command will:
@@ -35,16 +35,16 @@ Validators are represented by a BLS keypair. It is recommended to generate valid
 To create a new validator use the following command:
 
 ```
-lodestar account validator create --name primary --passphrase-file primary.pass
+lodestar account validator create --name primary --passphraseFile primary.pass --rootDir .lodestar
 ```
 
 This command will:
 
 - Derive a new BLS keypair from the wallet `primary`.
-- Create a new directory in `./.lodestar/keystores` containing:
+- Create a new directory in `.lodestar/keystores` containing:
   - An encrypted Keystore with the validator voting keypair.
   - An eth1_deposit_data.rlp file with the precomputed Eth1 deposit transaction data ready to be submitted to the deposit contract.
-- Store the validator voting Keystore password in `./.lodestar/secrets`.
+- Store the validator voting Keystore password in `.lodestar/secrets`.
 - Print the validator public key to the terminal
 
 <!-- prettier-ignore-start -->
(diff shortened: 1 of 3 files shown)


docs/usage/key-management.md
```diff
@@ -38,10 +41,10 @@ lodestar account validator create --name primary --passphrase-file primary.pass
 This command will:
 
 - Derive a new BLS keypair from the wallet `primary`.
-- Create a new directory in `$rootDir/validators` containing:
+- Create a new directory in `./lodestar/keystores` containing:
```
wemeetagain: `./.lodestar` here and below.

wemeetagain (dismissed):

mpetrunic (dismissed):
Wonder if maybe it would be better to add rootDir flag explicitly in command?

mpetrunic (approved):

wemeetagain (approved):
When we resolve #1232 we can remove `--root-dir` from the docs and rely on the default

mpetrunic:
It would still be nice to use custom directory instead of default

wemeetagain:
> It would still be nice to use custom directory instead of default

Right, we should definitely explain the `--root-dir` and other options in the docs.
I think we could expand the docs to include something like "Advanced configuration", where we cover all of these options not directly related to running the lodestar instance.
This would be beyond a "Testnet configuration" page, which would likely also mention `--root-dir` as a way to partition data per-testnet.
In general tho, I think the docs should be geared to mainnet users so we should keep it simpler where we can. We might even mention that explicitly somewhere. "These docs assume mainnet usage. For testnet use, see 'Testnet configuration'" (or something to that effect).

PR #1237, opened 2020-07-27
Add customizable graffiti param to validator client

### LINKED ISSUE #1219: validator graffiti

dapplion:
Fixes https://github.com/ChainSafe/lodestar/issues/1219.

This option is available in the validator client CLI with `--graffiti`. It's also available programmatically in `IValidatorOptions` with an option of the same name. The graffiti value is handled as a UTF8 string until the `produceBlock` handler in the lodestar API impl, where it's converted to a Buffer, cropped to 32 bytes and right padded with zeros.

### Questions

- What should be the default value of graffiti if unspecified? Typical choices are either a zero hash, or the version of the client.
https://github.com/ChainSafe/lodestar/blob/de782e9ae74034e06cabf23e829187e70ee79ba7/packages/lodestar-cli/src/cmds/validator/options.ts#L42
- Does fastify URI encode query params so this string can contain unsafe characters? Should we do any sanitation of the graffiti?
https://github.com/ChainSafe/lodestar/blob/de782e9ae74034e06cabf23e829187e70ee79ba7/packages/lodestar/src/api/rest/routes/validator/produceBlock.ts#L46


packages/lodestar-cli/src/cmds/validator/options.ts
```diff
@@ -33,6 +34,13 @@ export const validatorOptions: CommandBuilder<{}, IValidatorCliOptions> = {
   force: {
     description: "Open validators even if there's a lockfile. Use with caution",
     type: "boolean"
-  } as Options
+  } as Options,
+
+  graffiti: {
+    description: "Specify your custom graffiti to be included in blocks (plain UTF8 text, 32 characters max)",
```
mpetrunic, line 40: But utf8 characters can be up to 4 bytes, right? 

Maybe expect ascii text? :smile:

packages/lodestar/src/api/rest/routes/validator/produceBlock.ts
```diff
@@ -41,7 +42,8 @@ export const registerBlockProductionEndpoint: LodestarRestApiEndpoint = (fastify
       const block = await api.validator.produceBlock(
         request.query.slot,
         fromHex(request.query.proposer_pubkey),
-        fromHex(request.query.randao_reveal)
+        fromHex(request.query.randao_reveal),
+        request.query.graffiti
```
mpetrunic, line 49: add to opts properties graffitti

packages/lodestar-cli/src/cmds/validator/options.ts
dapplion: Truee

wemeetagain: Maybe we can be more flexible, convert to Buffer, then either pad or slice to 32 bytes.

wemeetagain (dismissed):

packages/lodestar/src/util/graffiti.ts
```diff
@@ -0,0 +1,12 @@
+import {GRAFFITI_SIZE} from "../constants";
+
+/**
+ * Parses a graffiti ASCII string and returns a 32 bytes buffer right padded with zeros
+ * @param graffiti
+ */
+export function toGraffitiBuffer(graffiti: string): Buffer {
+  return Buffer.concat([
+    Buffer.from(graffiti, "ascii"),
```
wemeetagain: I think we should allow for utf-8. seems that block explorers are parsing the graffiti as utf-8 and hex.
As long as we slice to 32 bytes, we're fine.

wemeetagain (dismissed):

wemeetagain (approved):

packages/lodestar-cli/src/util/graffiti.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Guesses the package version finding up the closest package.json
+ */
+function guessVersion(): string {
+  const res = readPkgUp.sync();
+  if (!res) throw Error("No package.json found");
+  const version = res.packageJson.version;
+  if (!version) throw Error(`No version in ${res.path}`);
+  return version;
```
mpetrunic, line 14: We should probably use that in api getVersion endpoint

packages/lodestar-cli/src/util/graffiti.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+/**
+ * Computes a default graffiti fetching dynamically the package info
+ */
+export function getDefaultGraffiti(): string {
+  try {
+    return `${lodestarPackageName}-${guessVersion()}`;
+  } catch (e) {
+    // eslint-disable-next-line no-console
```
mpetrunic, line 24: how about we disable this rule for all of cli package?

mpetrunic (approved):
We can do those in separate PRs I guess

PR #1240, opened 2020-07-27
Fix reqResp logging, simplify logger context type

wemeetagain:
Using JSON.stringify on ssz types occasionally blows up when the input is/has BigInt. This was happening to coerce a request body into a log entry.

Adds a reverse mapping for Method => Req SSZ type, uses that to toJson request body.
Simplifies logger `Context` type to `Json`.

### DIFF
--- a/packages/lodestar-utils/test/unit/logger/format.test.ts
+++ b/packages/lodestar-utils/test/unit/logger/format.test.ts
@@ -1,21 +1,25 @@
 import {serializeContext} from "../../../src/logger";
 import {expect} from "chai";
-import {config} from "@chainsafe/lodestar-config/lib/presets/minimal";
 
 describe("log format", function () {
 
   describe("serialize context", function () {
     it("should serialize context", function () {
       const serialized = serializeContext({
-        array: config.types.Uint64.hashTreeRoot(10n),
-        buffer: Buffer.alloc(2, 0),
+        object: {
+          foo: "bar",
+          baz: 1,
+        },
+        array: ["a", "small", "array"],
         string: "test",
-        bigint: 10n
+        number: 3,
+        bool: true,
       });
       expect(serialized)
         .to.equal(
-          "array=0x0a00000000000000000000000000000000000000000000000000000000000000,"
-          +" buffer=0x0000, string=test, bigint=10"
+          "object={\"foo\":\"bar\",\"baz\":1},"
+          +" array=[\"a\",\"small\",\"array\"],"
+          +" string=test, number=3, bool=true"
         );
     });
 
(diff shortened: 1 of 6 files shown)


twoeths (approved):

packages/lodestar-utils/test/unit/logger/format.test.ts
```diff
@@ -1,21 +1,25 @@
...
+        },
+        array: ["a", "small", "array"],
         string: "test",
-        bigint: 10n
+        number: 3,
+        bool: true,
       });
       expect(serialized)
         .to.equal(
-          "array=0x0a00000000000000000000000000000000000000000000000000000000000000,"
```
mpetrunic, line 17: I really liked auto serializing arrays to hex since we have a lot of those throughout the code

packages/lodestar-utils/test/unit/logger/format.test.ts
```diff
@@ -1,21 +1,25 @@
...
 
   describe("serialize context", function () {
     it("should serialize context", function () {
       const serialized = serializeContext({
-        array: config.types.Uint64.hashTreeRoot(10n),
-        buffer: Buffer.alloc(2, 0),
+        object: {
+          foo: "bar",
+          baz: 1,
+        },
```
mpetrunic, line 12: It was designed to handle single level context to force more readable and cleaner logs. If you wish to log object you should serialize it and submit to context. This change makes logger smarter but it could easily result in log clutter.

If you are worried about serializing logs unnecessary for debug logs that aren't displayed we could accept function as context to resolve only if it's going to be displayed

wemeetagain: I know it was designed that way, I do think we should minimize nested structures when we can. 
I'm just trying to make our logs more machine readable, and handling arbitrary `Json` IMO is part of the process, especially for debug or verbose-level logs.

IMO, in an ideal setting, we would output logs to JSON-LD format in production, or with a certain "productionized logs" flag.
We would still have a pretty-printer, like in this file, for daily use, but the underlying representation is simply `Json`. This will allow us to easily ingest logs into log aggregators and do more advanced analytics on logs. This is how logging will be used in production settings.
Logs that look like `[CHAIN] "Attestation ${X} processed at epoch ${Y}"` should be refactored into `[CHAIN] "processed attestation", {attestationRoot, epoch}`. Which has a machine representation of `{timestamp: 9999999, module: "chain", message: "processed attestation", attestationRoot, epoch}`

packages/lodestar-utils/test/unit/logger/format.test.ts
wemeetagain: There was actually only one instance of using this feature, all else already are either `toJson`/`toHexString`ed.

Issue #1238, opened 2020-07-27
CLI flag documentation


wemeetagain:
**Is your feature request related to a problem? Please describe.**
The lodestar docs have CLI usage pages, but nothing detailing the available CLI flags.

**Describe the solution you'd like**
We should have generated docs showing all available CLI options per command.
Teku does a good job of this here: https://docs.teku.pegasys.tech/en/latest/Reference/CLI/CLI-Syntax/

**Describe alternatives you've considered**
Direct users to run `--help` for option discovery.

3xtr4t3rr3str14l:
For any options that have missing descriptions, I will also add those to the corresponding cli options object.

3xtr4t3rr3str14l:
closed via #1241

PR #1241, opened 2020-07-27
Dynamic generation of CLI docs

3xtr4t3rr3str14l:
solves #1238 

- Added a script to generate CLI docs, which can be ran from the monorepo package.json
- the generated cli.md is included in this PR as well
- added descriptions for CLI flags that do not have descriptions


mpetrunic (commented):
LGTM

docs/usage/cli.md
```diff
@@ -0,0 +1,98 @@
...
+| network.bootnodes | array | undefined | 
+| network.multiaddrs | array | undefined | 
+| templateConfigFile | string | Template configuration used to initialize beacon node | 
+| genesisStateFile | string | Genesis state in ssz-encoded format | 
+| testnet | string | Use a testnet configuration and genesis file | 
+| beaconDir | string | Beacon root dir | 
+| dbDir | string | Beacon DB dir | 
+| configFile | string | Beacon node configuration file | 
+| peerIdFile | string | Peer ID file | 
+| enrFile | string | ENR file |
```
dapplion: Actually it would be nice to have description for all of these options. You may browse the code in BeaconNodeOptions and it's nested objects for descriptions in the Typescript interfaces. Spec params options can be excluded from this document entirely.

3xtr4t3rr3str14l: yep, spec params are already exlcuded via https://github.com/ChainSafe/lodestar/pull/1241/commits/398e6867a6d6c22cf8870ae4add01ad5ffebc021
other descriptions coming next

wemeetagain (commented):
Can you add the generation script to the docs github action?

In that case, should we check in the `docs/usage/cli.md`? @dapplion thoughts?

packages/lodestar-cli/tmp.json
```diff
@@ -0,0 +1,53 @@
+{
```
mpetrunic: I don't think this file belongs here?

mpetrunic (dismissed):

docs/usage/cli.md
```diff
@@ -0,0 +1,136 @@
+
```
dapplion: If this file is dynamically generated it should not be committed and `.gitignore`ed instead. Otherwise, it might create unnecessary diffs on other PRs.

dapplion (dismissed):

dapplion:
See https://github.com/ChainSafe/lodestar/pull/1357 as proposed requirement to this PR

wemeetagain:
#1357 now merged

3xtr4t3rr3str14l:
pushed integrated changes from #1357 by adapting @dapplion 's `docGeneration.ts`

docs/usage/cli.md
dapplion: @3xtr4t3rr3str14l bump

scripts/generateCLIDocs.ts
```diff
@@ -0,0 +1,86 @@
+import { writeFile } from "fs";
```
dapplion: I think it would make more sense to have this file is the lodestar-cli package. You could require the src instead of requiring `@chainsafe/lodestar-cli`

docs/usage/cli.md
3xtr4t3rr3str14l: done!

scripts/generateCLIDocs.ts
3xtr4t3rr3str14l: done!

wemeetagain (dismissed):
:tada: :1st_place_medal: 
Awesome endeavor

dapplion:
@3xtr4t3rr3str14l You should configure a git hook in your local environment to run `yarn lint` or `lerna run lint` before pushing
https://git-scm.com/docs/githooks
This way you will prevent the need for `linting` commits that often

3xtr4t3rr3str14l:
> @3xtr4t3rr3str14l You should configure a git hook in your local environment to run `yarn lint` or `lerna run lint` before pushing
> https://git-scm.com/docs/githooks
> This way you will prevent the need for `linting` commits that often

Right on.  Maybe this is something we should consider implementing across the repo?  Or would that be too annoying?

dapplion:
> > @3xtr4t3rr3str14l You should configure a git hook in your local environment to run `yarn lint` or `lerna run lint` before pushing
> > https://git-scm.com/docs/githooks
> > This way you will prevent the need for `linting` commits that often
> 
> Right on. Maybe this is something we should consider implementing across the repo? Or would that be too annoying?

That's something each developer can configure on it's own. It doesn't affect Github, only your local git repo

wemeetagain (approved):

PR #1246, opened 2020-07-28
Regular sync: retry requests to out-of-sync peer with best peer

### LINKED ISSUE #1234: Regular Sync: some peers are out-of-sync
**Describe the bug**
In my environment, I can sync to gossip parent blocks 5-6 times. However there are 1-2 times I cannot do that because we sent `beacon_block_by_range` requests to out-of-date peers. 
```
2020-07-26 21:56:28  [SYNC]             info: Requesting blocks from slot 194977 to slot 195131
2020-07-26 21:56:28  [NETWORK]       verbose: sending beacon_blocks_by_range request to 16Uiu2HAmUghihDaHutopxWPoxjUedJZBLVuuP1NBDq2LfpQZTnad requestId=3fb6e9d3489b0875, encoding=ssz_snappy
2020-07-26 21:56:28  [NETWORK]       verbose: sending beacon_blocks_by_range request to 16Uiu2HAm2BfJQmNSNdAtGfJLH1CyQKKRVr425uJgrSq9ps8TLJe2 requestId=c618c049d122bcee, encoding=ssz_snappy
...
2020-07-26 21:56:29  [NETWORK]       verbose: receive beacon_blocks_by_range response with 0 chunks from 16Uiu2HAmUghihDaHutopxWPoxjUedJZBLVuuP1NBDq2LfpQZTnad requestId=3fb6e9d3489b0875, encoding=ssz_snappy, body={"startSlot":194977,"step":1,"count":78}
2020-07-26 21:56:31  [NETWORK]       verbose: receive beacon_blocks_by_range response with 52 chunks from 16Uiu2HAm2BfJQmNSNdAtGfJLH1CyQKKRVr425uJgrSq9ps8TLJe2 requestId=c618c049d122bcee, encoding=ssz_snappy, body={"startSlot":195055,"step":1,"count":77}
```

Since the 1st peer above returned 0 chunk because it's not up to date, we cannot process any blocks. So we have to filter to up-to-date peers only.

There are 2 ways I can think of:
+ Filter based on headSlot of Status message, send status messages at init time of regular sync. This maybe does not work if peers have `headSlot` not equal to `currentSlot`
+ Find a way to keep peerIds of gossip blocks and just get peers based on that - but a peer's chain may not be up to date even it published gossip blocks

twoeths:
resolves #1234 

```
Given 3 peers, we can query by interleaving way, assuming we want to get block 10 to 20 for regular sync:
Round 0

Peer 0: start_slot=10, step=3 => return 10,13,16,19
Peer 1: start_slot=11, step=3 => return 11,14 - this peer is out-of-sync because expected latest slot=20
Peer 2: start_slot=12, step=3 => return null - error querying this peer - timeout for example
=> bestPeer=peer has max(lastSlot)=peer 0
Round 1 - fallback

Query bestPeer with start_slot=17, step=3
Query bestPeer with start_slot=12, step=3
```

Tested successfully a couple of times. Cannot test the latest version after I rebase to master because I got a some recent issues with "initial sync".

### DIFF
--- a/packages/lodestar/src/sync/regular/naive/naive.ts
+++ b/packages/lodestar/src/sync/regular/naive/naive.ts
@@ -131,6 +131,7 @@ export class NaiveRegularSync implements IRegularSync {
     const reqResp = this.network.reqResp;
     const getSyncPeers = this.getSyncPeers;
     const setTarget = this.setTarget;
+    const doneFirstSync = this.doneFirstSync;
     await pipe(
       this.targetSlotRangeSource,
       (source) => {
@@ -145,7 +146,9 @@ export class NaiveRegularSync implements IRegularSync {
               await setTarget(lastFetchedSlot, false);
             } else {
               // some peers maybe not up to date, retry the range again next time
-              await setTarget(range.start - 1, false);
+              // first sync: force it runs again
+              // after first sync: onNextSlot will trigger
+              await setTarget(range.start - 1, !doneFirstSync);
             }
           }
         })();
--- a/packages/lodestar/src/sync/regular/naive/naive.ts
+++ b/packages/lodestar/src/sync/regular/naive/naive.ts
@@ -12,7 +12,7 @@ import {IReputationStore} from "../../IReputation";
 import {SignedBeaconBlock, Slot, Root} from "@chainsafe/lodestar-types";
 import pushable, {Pushable} from "it-pushable";
 import pipe from "it-pipe";
-import {fetchBlockChunks, processSyncBlocks} from "../../utils";
+import {processSyncBlocks, fetchBlockInterleave} from "../../utils";
 import {ISlotRange} from "../../interface";
 import {getCurrentSlot} from "@chainsafe/lodestar-beacon-state-transition";
 import {GossipEvent} from "../../../network/gossip/constants";
@@ -138,7 +138,7 @@ export class NaiveRegularSync implements IRegularSync {
           for await (const range of abortSource(source, controller.signal, {returnOnAbort: true})) {
             const lastFetchedSlot = await pipe(
               [range],
-              fetchBlockChunks(logger, chain, reqResp, getSyncPeers),
+              fetchBlockInterleave(logger, reqResp, getSyncPeers),
               processSyncBlocks(config, chain, logger, false)
             );
             if(lastFetchedSlot) {
--- a/packages/lodestar/src/sync/utils/blocks.ts
+++ b/packages/lodestar/src/sync/utils/blocks.ts
@@ -39,14 +39,15 @@ export function chunkify(blocksPerChunk: number, currentSlot: Slot, targetSlot: 
 export async function getBlockRangeFromPeer(
   rpc: IReqResp,
   peer: PeerId,
-  chunk: ISlotRange
+  chunk: ISlotRange,
+  step = 1
 ): Promise<SignedBeaconBlock[]|null> {
   return await rpc.beaconBlocksByRange(
     peer,
     {
       startSlot: chunk.start,
-      step: 1,
-      count: chunk.end - chunk.start + 1
+      step,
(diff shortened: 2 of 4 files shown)


mpetrunic:
I decided to not use interleave since you need to fetch all blocks in range  before you begin to process them which is not suitable for js singlethreaded way. We need to utilize time waiting for block by processing ones we already receive. I think our sync is a bit slower because somewhere along the way we ended up waiting for all blocks before processing

twoeths:
@mpetrunic `getBlockRange` and `getBlockRangeInterleave` are the same in term of waiting for all requests to finish.
+ getBlockRange: wait for chunk1 (1-5) and chunk 2 (6-10)
+ getBlockRangeInterleave: wait for request 1 (1,3,5,7,9) and request 2 (2,4,6,8,10)

mpetrunic:
> @mpetrunic `getBlockRange` and `getBlockRangeInterleave` are the same in term of waiting for all requests to finish.
> 
> * getBlockRange: wait for chunk1 (1-5) and chunk 2 (6-10)
> * getBlockRangeInterleave: wait for request 1 (1,3,5,7,9) and request 2 (2,4,6,8,10)

But in getBlockRange you can process all 5 blocks while waiting fot chunk2 while in inteleave you cannot process block 3 until you receive response for request 2. Difference is even bigger if you increase chunk size to 20 blocks. Assuming you send request based on peer score. You can request last chunks from slower peers and when you finish processing earlier chunks you will receive slower ones without sync downtime

twoeths:
> But in getBlockRange you can process all 5 blocks while waiting fot chunk2 while in inteleave you cannot process block 3 until you receive response for request 2

in sync util we work with range not chunk so even you get chunk1 done, you still wait for chunk2 to be done and pass to processSyncBlocks, so I think it's the same in both case https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/sync/utils/sync.ts#L99

also, it's only applied for Regular Sync (not Initial Sync) and normally it takes less than 1 min to be up to head and we're done

mpetrunic:
> > But in getBlockRange you can process all 5 blocks while waiting fot chunk2 while in inteleave you cannot process block 3 until you receive response for request 2
> 
> in sync util we work with range not chunk so even you get chunk1 done, you still wait for chunk2 to be done and pass to processSyncBlocks, so I think it's the same in both case https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/sync/utils/sync.ts#L99
> 
> also, it's only applied for Regular Sync (not Initial Sync) and normally it takes less than 1 min to be up to head and we're done

As I said, somehwere along the way we lost general idea to process chunk while waiting for another :smile: 

Regular sync should actually sync all heads from all peers but it's not immediate concern

twoeths:
> As I said, somehwere along the way we lost general idea to process chunk while waiting for another

I'm thinking about this... not sure if we can do that because we have a check that a block is parent of next block after we get all chunks and sort by slots (chunk2 maybe done before chunk1)... But if we remove the check, we'll get stuck if any peers return us incorrect response...
```
if(!isInitialSync ||
        (isInitialSync && block.slot > headSlot && config.types.Root.equals(headRoot, block.parentRoot))) {
```

mpetrunic:
> > As I said, somehwere along the way we lost general idea to process chunk while waiting for another
> 
> I'm thinking about this... not sure if we can do that because we have a check that a block is parent of next block after we get all chunks and sort by slots (chunk2 maybe done before chunk1)... But if we remove the check, we'll get stuck if any peers return us incorrect response...
> 
> ```
> if(!isInitialSync ||
>         (isInitialSync && block.slot > headSlot && config.types.Root.equals(headRoot, block.parentRoot))) {
> ```

Yes but when we receive chunk, we can check if some block is descendant of head, if not we store it for after we try some other chunk

twoeths:
closing this as the bigger concern is we query peers with different forkchoice branches and we may not be up to head() due to that, see the issue.

PR #1260, opened 2020-07-30
Regular sync: find best peer at startup and use it

### LINKED ISSUE #1234: Regular Sync: some peers are out-of-sync

twoeths:
+ resolves #1234 
+ remove `onNewSlot` as it may cause missing blocks and we can process chain up to head()

At regular sync, if we query blocks from multiple peers we may get blocks from different forkchoice branch so we'll *not* be able to sync to head in that case.
This PR helps find best peer at Regular Sync start time and use it as the only peer.

image

### DIFF
--- a/packages/lodestar/src/sync/regular/naive/naive.ts
+++ b/packages/lodestar/src/sync/regular/naive/naive.ts
@@ -12,7 +12,7 @@ import {IReputationStore} from "../../IReputation";
 import {SignedBeaconBlock, Slot, Root} from "@chainsafe/lodestar-types";
 import pushable, {Pushable} from "it-pushable";
 import pipe from "it-pipe";
-import {fetchBlockChunks, processSyncBlocks, createStatus} from "../../utils";
+import {fetchBlockChunks, processSyncBlocks, createStatus, syncPeersStatus} from "../../utils";
 import {ISlotRange} from "../../interface";
 import {getCurrentSlot} from "@chainsafe/lodestar-beacon-state-transition";
 import {GossipEvent} from "../../../network/gossip/constants";
@@ -59,6 +59,10 @@ export class NaiveRegularSync implements IRegularSync {
     this.currentTarget = this.chain.forkChoice.headBlockSlot();
     const state = await this.chain.getHeadState();
     this.slotAtStart = getCurrentSlot(this.config, state.genesisTime);
+    if (this.currentTarget >= this.slotAtStart) {
+      this.logger.info(`Regular Sync: node is up to date, headSlot=${this.currentTarget}`);
+      return;
+    }
     this.logger.verbose(`Regular Sync: Current slot at start: ${this.slotAtStart}`);
     this.targetSlotRangeSource = pushable<ISlotRange>();
     await this.waitForBestPeer();
@@ -104,7 +108,8 @@ export class NaiveRegularSync implements IRegularSync {
       if(await this.checkSyncComplete()) {
         return;
       }
-      this.logger.info(`Regular Sync: Synced up to slot ${lastProcessedBlock.message.slot}`);
+      this.logger.info(`Regular Sync: Synced up to slot ${lastProcessedBlock.message.slot} ` +
+        `gossipParentBlockRoot=${this.gossipParentBlockRoot && toHexString(this.gossipParentBlockRoot)}`);
       if (lastProcessedBlock.message.slot >= this.slotAtStart && !this.isGossipStarted) {
         this.isGossipStarted = true;
         this.network.gossip.subscribeToBlock(this.chain.currentForkDigest, this.onGossipBlock);
@@ -175,23 +180,25 @@ export class NaiveRegularSync implements IRegularSync {
   };
 
   private waitForBestPeer = async (): Promise<void> => {
-    // check 2 times per slot
-    const waitingTime = this.config.params.SECONDS_PER_SLOT * 1000 / 2;
+    // check every slot
+    const waitingTime = this.config.params.SECONDS_PER_SLOT * 1000;
     const state = await this.chain.getHeadState();
     const status = createStatus(this.chain);
-    while (!this.bestPeer) {
+    this.controller = new AbortController();
+    let isAborted = false;
+    this.controller.signal.addEventListener("abort", () => {
+      this.logger.verbose("RegularSync: Abort waitForBestPeer");
+      isAborted = true;
+    });
+    while (!this.bestPeer && !isAborted) {
       const peers = this.network.getPeers();
       const previousSlot = getCurrentSlot(this.config, state.genesisTime) - 1;
-      await Promise.all(peers.map(peerId => async () => {
-        try {
-          this.reps.get(peerId.toB58String()).latestStatus = await this.network.reqResp.status(peerId, status);
-        } catch (e) {
-          this.logger.error(`Regular Sync: Failed to get peer ${peerId.toB58String()} status. Error: ` + e.message);
-        }
-      }));
+      await syncPeersStatus(this.reps, this.network, status);
+      const maxHeadSlot = Math.max(...peers.map(
+        (peerId) => this.reps.get(peerId.toB58String()).latestStatus?.headSlot || 0));
       this.bestPeer = peers.find(peerId => {
         const headSlot = this.reps.get(peerId.toB58String()).latestStatus?.headSlot;
-        return headSlot >= previousSlot;
+        return headSlot >= previousSlot && headSlot === maxHeadSlot;
       });
       await sleep(waitingTime);
       if (this.bestPeer) {
--- a/packages/lodestar/src/sync/regular/naive/naive.ts
+++ b/packages/lodestar/src/sync/regular/naive/naive.ts
@@ -12,11 +12,12 @@ import {IReputationStore} from "../../IReputation";
 import {SignedBeaconBlock, Slot, Root} from "@chainsafe/lodestar-types";
 import pushable, {Pushable} from "it-pushable";
 import pipe from "it-pipe";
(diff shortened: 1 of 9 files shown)


packages/lodestar/src/sync/regular/naive/naive.ts
```diff
@@ -154,12 +166,39 @@ export class NaiveRegularSync implements IRegularSync {
...
+      // bestPeer disconnected
+      this.bestPeer = undefined;
+      await this.waitForBestPeer();
+    }
+    return [this.bestPeer];
+  };
+
+  private waitForBestPeer = async (): Promise<void> => {
+    // check 2 times per slot
+    const waitingTime = this.config.params.SECONDS_PER_SLOT * 1000 / 2;
```
mpetrunic: Isn't this too often? How often does other clients refresh sync status?

Maybe we should run status update as daemon somewhere else?

packages/lodestar/src/sync/regular/naive/naive.ts
```diff
@@ -154,12 +171,41 @@ export class NaiveRegularSync implements IRegularSync {
...
+    if (!this.network.getPeers().includes(this.bestPeer)) {
+      // bestPeer disconnected
+      this.bestPeer = undefined;
+      await this.waitForBestPeer();
+    }
+    return [this.bestPeer];
+  };
+
+  private waitForBestPeer = async (): Promise<void> => {
+    // check every slot
```
wemeetagain: Can we only send a status if our head slot is < getCurrentSlot to avoid spamming peers?

packages/lodestar/src/sync/regular/naive/naive.ts
```diff
@@ -154,12 +171,41 @@ export class NaiveRegularSync implements IRegularSync {
...
+      const peers = this.network.getPeers();
+      const previousSlot = getCurrentSlot(this.config, state.genesisTime) - 1;
+      await syncPeersStatus(this.reps, this.network, status);
+      const maxHeadSlot = Math.max(...peers.map(
+        (peerId) => this.reps.get(peerId.toB58String()).latestStatus?.headSlot || 0));
+      this.bestPeer = peers.find(peerId => {
+        const headSlot = this.reps.get(peerId.toB58String()).latestStatus?.headSlot;
+        return headSlot >= previousSlot && headSlot === maxHeadSlot;
+      });
+      await sleep(waitingTime);
```
wemeetagain, line 189: should we be waiting _before_ we send statuses, so we get a more current status?

packages/lodestar/src/sync/regular/naive/naive.ts
twoeths: I checked that in start() so if we detect genesis at the same time with other peer we don't call `waitForBestPeer` and finish RegularSync immediately

packages/lodestar/src/sync/regular/naive/naive.ts
twoeths: I changed `waitingTime` to per slot as it's more reasonable.

I'm not sure the need to create a daemon for now unless we want to prepare something for the future. In that case we'll decide the waitingTime depending on different use cases for it.

For this use case, we just need this for a short period, created `syncPeersStatus()` util for it.

packages/lodestar/src/sync/regular/naive/naive.ts
```diff
@@ -154,12 +169,43 @@ export class NaiveRegularSync implements IRegularSync {
...
-        validPeers.push(peer);
+    if (!this.network.getPeers().includes(this.bestPeer)) {
+      // bestPeer disconnected
+      this.bestPeer = undefined;
+      await this.waitForBestPeer();
+    }
+    return [this.bestPeer];
+  };
+
+  private waitForBestPeer = async (): Promise<void> => {
```
mpetrunic: Can we make this cleaner like have two functions:
getBestHead(currentSlot, peers, reps): {root, slot} // highest slot
getBestPeer(root, reps): PeerId // takes all peers with this headRoot and get one with best score (once we land new gossipsub, take random for now)



Those methods should be standalone so we can test them properly

packages/lodestar/src/sync/regular/naive/naive.ts
```diff
@@ -154,12 +169,43 @@ export class NaiveRegularSync implements IRegularSync {
...
+    this.controller.signal.addEventListener("abort", () => {
+      this.logger.verbose("RegularSync: Abort waitForBestPeer");
+      isAborted = true;
+    });
+    while (!this.bestPeer && !isAborted) {
+      // wait first to make sure we have latest status
+      await sleep(waitingTime);
+      const status = createStatus(this.chain);
+      const previousSlot = getCurrentSlot(this.config, state.genesisTime) - 1;
+      await syncPeersStatus(this.reps, this.network, status);
```
mpetrunic: I would separate process of obtaining status so it's running during both initial and regular sync on some time interval

packages/lodestar/src/sync/regular/naive/naive.ts
```diff
@@ -154,12 +169,43 @@ export class NaiveRegularSync implements IRegularSync {
...
+      await this.waitForBestPeer();
+    }
+    return [this.bestPeer];
+  };
+
+  private waitForBestPeer = async (): Promise<void> => {
+    // check every slot
+    const waitingTime = this.config.params.SECONDS_PER_SLOT * 1000;
+    const state = await this.chain.getHeadState();
+    this.controller = new AbortController();
```
mpetrunic: this method should just take signal

packages/lodestar/src/sync/regular/naive/naive.ts
```diff
@@ -52,9 +57,17 @@ export class NaiveRegularSync implements IRegularSync {
     this.logger.info("Started regular syncing");
     this.chain.on("processedBlock", this.onProcessedBlock);
     this.currentTarget = this.chain.forkChoice.headBlockSlot();
+    const state = await this.chain.getHeadState();
+    this.slotAtStart = getCurrentSlot(this.config, state.genesisTime);
```
mpetrunic: Does this work if there are a lot of skip slots?

Why not start gossip as soon as regular sync starts?

twoeths: Gossip is started as soon as regular sync starts, this is all about listen for its event.
I'll change to listen in start bc I removed `onNewSlot` already, this would make it simpler

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -213,4 +213,18 @@ export async function syncPeersStatus(reps: IReputationStore, network: INetwork,
...
+}
+
+export function getBestHead(peers: PeerId[], reps: IReputationStore): number {
+  return Math.max(...peers.map(
+    (peerId) => reps.get(peerId.toB58String()).latestStatus?.headSlot || 0));
+}
+
+// should add peer score later
+export function getBestPeer(peers: PeerId[], reps: IReputationStore): PeerId {
+  const bestHead = getBestHead(peers, reps);
```
mpetrunic: receive targetHead as param and compare root, not slot

packages/lodestar/src/sync/regular/naive/naive.ts
```diff
@@ -58,16 +58,18 @@ export class NaiveRegularSync implements IRegularSync {
     this.chain.on("processedBlock", this.onProcessedBlock);
     this.currentTarget = this.chain.forkChoice.headBlockSlot();
```
mpetrunic: can we store it in const headSlot, so code below is more clearer.

I was so confused that target larger than current slot means we are synced :smile:

wemeetagain (dismissed):

wemeetagain:
file conflict

wemeetagain (approved):

PR #1263, opened 2020-07-30
Fix cmd account validator create - no validator bug

### LINKED ISSUE #1254: lodestar-cli: "account validator create" not creating validator
**Describe the bug**

The following process is based on the instructions given here: https://chainsafe.github.io/lodestar/usage/key-management/

1.  run `yarn run cli account wallet create --name primary --passphraseFile primary.pass --rootDir .altona`
2.  run `yarn run cli account validator create --name primary --passphraseFile primary.pass --rootDir .altona`

When I run these commands, as per the expectations set in the document linked above, I would expect to see entries in the `.altona/keystore/` and `.altona/secrets/ directories, but see nothing.  Also, the validator public key is not printed to the terminal.

**Expected behavior**

I would expect to see the following happen as described in the document linked above:
```
    Derive a new BLS keypair from the wallet primary.
    Create a new directory in .lodestar/keystores containing:
    An encrypted Keystore with the validator voting keypair.
    An eth1_deposit_data.rlp file with the precomputed Eth1 deposit transaction data ready to be submitted to the deposit contract.
    Store the validator voting Keystore password in .lodestar/secrets.
    Print the validator public key to the terminal
```

dapplion:
Fixes https://github.com/ChainSafe/lodestar/issues/1254.

FIxes a bug where if no --count or --atMost option was specified, validatorCount became `NaN` and no validator was created while not failing

### DIFF
--- a/packages/lodestar-cli/src/cmds/account/cmds/validator/create.ts
+++ b/packages/lodestar-cli/src/cmds/account/cmds/validator/create.ts
@@ -82,6 +82,7 @@ export async function handler(options: IValidatorCreateOptions): Promise<void> {
   const n = typeof count === "number" ? count
     : typeof atMost === "number" ? atMost - wallet.nextaccount
       : 1;
+  if (isNaN(n)) throw new Error(`Error computing validatorCount: ${n}`);
   if (n <= 0) throw new YargsError("No validators to create");
 
   const walletPassword = readPassphraseFile(passphraseFile);
--- a/packages/lodestar-cli/src/cmds/account/cmds/validator/create.ts
+++ b/packages/lodestar-cli/src/cmds/account/cmds/validator/create.ts
@@ -18,7 +18,6 @@ interface IValidatorCreateOptions extends IAccountValidatorOptions {
   depositGwei?: string;
   storeWithdrawalKeystore?: boolean;
   count?: number;
-  atMost?: number;
 }
 
 export const builder: CommandBuilder<{}, IValidatorCreateOptions> = {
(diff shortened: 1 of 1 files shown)


packages/lodestar-cli/src/cmds/account/cmds/validator/create.ts
```diff
@@ -79,7 +79,10 @@ export async function handler(options: IValidatorCreateOptions): Promise<void> {
   const validatorDirBuilder = new ValidatorDirBuilder(accountPaths);
   const walletManager = new WalletManager(accountPaths);
   const wallet = walletManager.openByName(name);
-  const n = count || atMost - wallet.nextaccount;
+  const n = typeof count === "number" ? count
+    : typeof atMost === "number" ? atMost - wallet.nextaccount
+      : 1;
+  if (isNaN(n)) throw new Error(`Error computing validatorCount: ${n}`);
```
wemeetagain: Can we throw a YargsError if neither count or atMost are specified?

3xtr4t3rr3str14l: @wemeetagain  what's the rationale for having those fields be required?

3xtr4t3rr3str14l: i know that's what caused this bug, but just wondering why it wasn't required in the first place or how this worked before without it being required (if it worked before)

wemeetagain: I misread this. It appears its defaulting to count=1 if neither is set.
We don't need them to be required, maybe we can be more clear in comments or cli helptext what the defaults are.
We should check if both are set tho. Error or handle that case.

It worked before if you assumed that either count or atMost were required, and set one of them.

dapplion: @wemeetagain there's already a yargs option configuration `conflicts` that prevents both options from being set at the same time.

I've inherited `atMost` from Lighthouse, tho it's a weird option. Would it make sense to remove `atMost` and just keep `count` defaulting it to 1?

wemeetagain: :+1:

wemeetagain (approved):

Issue #1226, opened 2020-07-25
CLI variables rename


dapplion:
Change some variable names to a more descriptive name:
- `network.multiaddrs` to `network.localMultiaddrs`
- `network.bootnodes` to `network.bootMultiaddrs`

Make sure that its use in all components is changed consistently

3xtr4t3rr3str14l:
Can we also change `api` to not nest it's options within the `rest` subcategory?
Currently, we are required to configure `api` options like this:
```
"api": {
    "rest": {
      "enabled": true,
      "namespaces": ["beacon", "validator"],
      "host": "127.0.0.1",
      "port": 9596,
      "cors": {
        "origin": "*"
      }
    }
  },
```

but i think we should structure them like this:
```
  "api": {
    "enabled": true,
    "namespaces": ["beacon", "validator"],
    "host": "127.0.0.1",
    "port": 9596,
    "cors": {
      "origin": "*"
    }
  },
```

mpetrunic:
It will be awkward if we introduce rpc or graphql at some point^^

3xtr4t3rr3str14l:
@mpetrunic yeah I see what you mean.  If that's something we think we will have in the future (or some other API type besides what we currently have), then might be worth leaving as is.

dapplion:
All code in the `lodestar` subpackage expects that option type so it would be best to not change it.

PR #1267, opened 2020-07-30
lodestar-cli: fix init fields copy bug

### LINKED ISSUE #1266: --testnet command copies over extra fields
**Describe the bug**

When running `yarn run cli beacon init --testnet altona`, the resultant beacon.config.json also contains these parameters:
```
  "templateConfigFile": null,
  "testnet": "altona"
```

**Expected behavior**

I would expect to see the config created without those 2 parameters mentioned above.

3xtr4t3rr3str14l:
Fixes #1266 to copy over the necessary fields to the beacon.config.json when running `yarn run cli beacon init --testnet altona`

### DIFF
--- a/packages/lodestar-cli/src/options/beaconNodeOptions/api.ts
+++ b/packages/lodestar-cli/src/options/beaconNodeOptions/api.ts
@@ -2,37 +2,37 @@ import {Options} from "yargs";
 import defaultOptions from "@chainsafe/lodestar/lib/node/options";
 
 export const apiOptions = {
-  "api.namespaces": {
-    alias: ["api.namespaces", "api.rest.api"],
+  "api.rest.api": {
+    alias: ["api.namespaces"],
     type: "array",
     choices: ["beacon", "validator"],
     defaultDescription: JSON.stringify(defaultOptions.api.rest.api),
     group: "api",
   } as Options,
 
-  "api.cors.origin": {
-    alias: ["api.cors.origin", "api.rest.cors"],
+  "api.rest.cors": {
+    alias: ["api.cors.origin"],
     type: "string",
     defaultDescription: defaultOptions.api.rest.cors,
     group: "api",
   } as Options,
 
-  "api.enabled": {
-    alias: ["api.enabled", "api.rest.enabled"],
+  "api.rest.enabled": {
+    alias: ["api.enabled"],
     type: "boolean",
     defaultDescription: String(defaultOptions.api.rest.enabled),
     group: "api",
   } as Options,
 
-  "api.host": {
-    alias: ["api.host", "api.rest.host"],
+  "api.rest.host": {
+    alias: ["api.host"],
     type: "string",
     defaultDescription: defaultOptions.api.rest.host,
     group: "api",
   } as Options,
 
-  "api.port": {
-    alias: ["api.port", "api.rest.port"],
+  "api.rest.port": {
+    alias: ["api.port"],
     type: "number",
     defaultDescription: String(defaultOptions.api.rest.port),
     group: "api",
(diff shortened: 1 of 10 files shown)


packages/lodestar-cli/src/options/beaconNodeOptions/api.ts
```diff
@@ -2,37 +2,37 @@ import {Options} from "yargs";
...
+    alias: ["api.host", "api.rest.host"],
     type: "string",
     defaultDescription: defaultOptions.api.rest.host,
     group: "api",
   } as Options,
 
-  "api.rest.port": {
-    alias: ["api.port"],
+  "api.port": {
+    alias: ["api.port", "api.rest.port"],
```
dapplion: The property names here cannot be changed, they must match the actual BeaconNodeOption types.

packages/lodestar-cli/src/options/beaconNodeOptions/network.ts
```diff
@@ -27,13 +27,13 @@ export const networkOptions = {
...
 
-  "network.bootnodes": {
+  "network.bootMultiaddrs": {
     type: "array",
     defaultDescription: JSON.stringify(defaultOptions.network.bootnodes),
     group: "network",
   } as Options,
 
-  "network.multiaddrs": {
+  "network.localMultiaddrs": {
```
dapplion: The property names here cannot be changed, they must match the actual BeaconNodeOption types.

dapplion (changes requested):

packages/lodestar-cli/src/options/beaconNodeOptions/api.ts
3xtr4t3rr3str14l: so does that mean we need to restructure our altona configs going forward?  for example, right now in `testnets/altona.ts`, the structure for the `api` options are as follows:

```
  "api": {
    "enabled": true,
    "namespaces": ["beacon", "validator"],
    "host": "127.0.0.1",
    "port": 9596,
    "cors": {
      "origin": "*"
    }
  },
```

Do you mean to say that now it should look something like this instead?
```
 "api": {
    "rest": {
      "enabled": true,
      "namespaces": ["beacon", "validator"],
      "host": "127.0.0.1",
      "port": 9596,
      "cors": {
        "origin": "*"
      }
    }
  },
```

I'm not sure why we need the `.rest` part.  I thought that was just an alias for when we were still using the old `mergeBeaconOptions` method?

3xtr4t3rr3str14l: To clarify, there aren't any other subcategories of `api` options, so I don't see a reason to subcategorize them with `rest`

wemeetagain: @3xtr4t3rr3str14l 
We've simplified the configuration to _require_ one-for-one matching of yargs `Options` to `BeaconNodeOptions`. (if there's a configuration file that doesn't match the structure of `BeaconNodeOptions` (with a few additional fields, eg: `genesisStateFile`), you can think of that as a bug)
I think what you're hoping for is a matter of renaming the options in `BeaconNodeOptions`. (in the same vein as https://github.com/ChainSafe/lodestar/issues/1226) and pushing those changes to the yargs level as well. Might add a comment there, as there are quite a few fields in `BeaconNodeOptions` that could/should be renamed.

packages/lodestar-cli/src/options/beaconNodeOptions/network.ts
3xtr4t3rr3str14l: changes for this file reverted

packages/lodestar-cli/src/options/beaconNodeOptions/api.ts
3xtr4t3rr3str14l: changes for this file reverted

3xtr4t3rr3str14l: also, made my comment in #1226 accordingly to request to remove the `api` nest

wemeetagain:
Should be resolved with #1301

PR #1272, opened 2020-07-31
Add basic beacon option to log to file

mpetrunic:
Later on we should probably introduce log rotatio but for current status it will suffice to just log everything to file so we can attach log file to issue and debug later on.

### DIFF
--- a/packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
+++ b/packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
@@ -1,8 +1,9 @@
 import {CommandModule} from "yargs";
 import {IBeaconOptions} from "../../options";
 import {initHandler} from "./init";
+import {IGlobalArgs} from "../../../../options";
 
-export const init: CommandModule<IBeaconOptions, IBeaconOptions> = {
+export const init: CommandModule<IGlobalArgs, IBeaconOptions> = {
   command: "init",
   describe: "Initialize lodestar beacon node",
   handler: initHandler
--- a/packages/lodestar-cli/src/cmds/beacon/cmds/run/run.ts
+++ b/packages/lodestar-cli/src/cmds/beacon/cmds/run/run.ts
@@ -3,7 +3,7 @@ import process from "process";
 import {initBLS} from "@chainsafe/bls";
 import {BeaconNode} from "@chainsafe/lodestar/lib/node";
 import {createNodeJsLibp2p} from "@chainsafe/lodestar/lib/network/nodejs";
-import {WinstonLogger} from "@chainsafe/lodestar-utils";
+import {fileTransport, WinstonLogger} from "@chainsafe/lodestar-utils";
 import {IBeaconOptions} from "../../options";
 import {readPeerId, readEnr, writeEnr} from "../../../../network";
 import {ENR} from "@chainsafe/discv5";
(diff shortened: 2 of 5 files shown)


packages/lodestar-cli/src/cmds/beacon/options.ts
```diff
@@ -14,9 +13,10 @@ export type IBeaconOptions =
     templateConfigFile?: string;
     genesisStateFile?: string;
     testnet?: TestnetName;
+    logFile?: string;
   };
 
-export const beaconOptions = {
+export const beaconOptions: {[k: string]: Options} = {
```
dapplion, line 19: By doing this you lose access to each option in a type safe way in other parts of the code such as `beaconOptions.genesisStateFile`

packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
```diff
@@ -1,8 +1,9 @@
 import {CommandModule} from "yargs";
 import {IBeaconOptions} from "../../options";
 import {initHandler} from "./init";
+import {IGlobalArgs} from "../../../../options";
 
-export const init: CommandModule<IBeaconOptions, IBeaconOptions> = {
+export const init: CommandModule<IGlobalArgs, IBeaconOptions> = {
```
dapplion, line 6: Why is this type change necessary?

packages/lodestar-cli/src/cmds/beacon/cmds/run/run.ts
```diff
@@ -34,7 +36,13 @@ export async function runHandler(options: IBeaconOptions): Promise<void> {
 
   const config = getBeaconConfig(options.preset, options.params);
   const libp2p = await createNodeJsLibp2p(peerId, options.network);
-  const logger = new WinstonLogger();
+  const loggerTransports = [
+    consoleTransport
+  ];
+  if(options.logFile) {
+    loggerTransports.push(fileTransport(path.join(options.rootDir, options.logFile)));
```
dapplion, line 43: What if the user specifies `options.logFile` as an absolute path?

dapplion (changes requested):

packages/lodestar-cli/src/cmds/beacon/cmds/run/run.ts
mpetrunic: it won't work, but same goes for all path we use :(

packages/lodestar-cli/src/cmds/beacon/options.ts
mpetrunic: how so?
We are passing this variable to yargs.options so all commands are using IBeaconOptions interface anyways.
This change just adds more typesafety by forcing every option to have type Options. Without this you can submit (as I did with typo) invalid option that fails at runtime

packages/lodestar-cli/src/cmds/beacon/cmds/init/index.ts
mpetrunic: Not sure, I'm trying to fix some type error but it would seem we are using these types completely wrong.
First one is passed in buidler commands and second in handler

packages/lodestar-cli/src/cmds/beacon/cmds/run/run.ts
dapplion: True

dapplion (approved):

wemeetagain (approved):

Issue #1255, opened 2020-07-29
Create more complete unit tests for lodestar-cli


3xtr4t3rr3str14l:
Create unit tests for all CLI commands in lodestar-cli

PR #1276, opened 2020-07-31
lodestar-cli e2e tests

3xtr4t3rr3str14l:
Fixing #1255 
Decided to consolidate everything into the account tests since a lot of account partially depends on `init` anyway


packages/lodestar-cli/test/unit/lodestar/account.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      name: "primary",
+      passphraseFile: "primary.pass",
+      // @ts-ignore
+    }).command(account).help().parse(["account", "wallet", "create"], resolve));
+    await new Promise(resolve => setTimeout(resolve, 500));
+    expect(fs.existsSync(`${tmpDir}/wallets/`)).to.be.true;
+  });
+
+  // lodestar account wallet list --rootDir .tmp
+  it("should list existing wallets", async function() {
```
dapplion: This test should assert that the output of the command contains the name of the new wallet

packages/lodestar-cli/test/unit/lodestar/account.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  it("should create new validator", async function() {
+    await new Promise(resolve => yargs().default({
+      name: "primary",
+      passphraseFile: "primary.pass",
+      rootDir: tmpDir,
+      preset: "mainnet", // @TODO: do we really need this?
+      // @ts-ignore
+    }).command(account).help().parse(["account", "validator", "create"], resolve));
+    await new Promise(resolve => setTimeout(resolve, 500));
+    assert(spy.called);
```
dapplion: This test should assert that the resulting files and file structure are correct.

packages/lodestar-cli/test/unit/lodestar/account.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  // lodestar account validator list --rootDir .tmp
+  it("should list validators", async function() {
+    await new Promise(resolve => yargs().default({
+      rootDir: tmpDir,
+      // @ts-ignore
+    }).command(account).help().parse(["account", "validator", "list"], resolve));
+    await new Promise(resolve => setTimeout(resolve, 500));
+    assert(spy.called);
```
dapplion: This test should assert that the output contains the public key of the new validator

packages/lodestar-cli/test/unit/lodestar/account.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  // @TODO: lodestar account validator deposit --validator <validatorID> --rootDir .tmpDir
+  it("should make a deposit to validator registration contract", async function() {
+    await new Promise(resolve => yargs().default({
+      rootDir: tmpDir,
+      // @ts-ignore
+    }).command(account).help().parse(["account", "validator", "deposit"], resolve));
+    await new Promise(resolve => setTimeout(resolve, 500));
```
dapplion: Where is it making the deposit to?

packages/lodestar-cli/test/unit/lodestar/dev.test.ts
```diff
@@ -0,0 +1,16 @@
+import * as fs from "fs";
+import yargs from "yargs/yargs";
+import {tmpDir} from "../../constants";
+import { dev } from "../../../src/cmds/dev";
+import { expect } from "chai";
+
+describe.only("dev cli", function() {
```
dapplion: Remove .only

packages/lodestar-cli/test/unit/lodestar/validator.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    await new Promise(resolve => setTimeout(resolve, 500));
+
+
+    // run validator command
+    await new Promise(resolve => yargs().default({
+      rootDir: tmpDir,
+      preset: "mainnet", // @TODO: do we really need this?
+      // @ts-ignore
+    }).command(validator).help().parse(["validator"], resolve));
+    await new Promise(resolve => setTimeout(resolve, 500));
```
dapplion: This would leave an uncontrolled failing validator process running in the background.

dapplion (changes requested):
I would focus in doing tests exclusively for the `account` commands and maybe to assert that the configuration is properly loaded. I would skip `run` or long-running commands for now.

packages/lodestar-cli/package.json
```diff
@@ -65,6 +65,7 @@
     "deepmerge": "^3.2.0",
     "ethers": "^5.0.2",
     "expand-tilde": "^2.0.2",
+    "ganache-core": "^2.10.2",
```
mpetrunic: dev dependency? 
Maybe we should use something more typesafe and with less problematic dependenvies like builderevm?

dapplion: +1 for buidler

packages/lodestar-cli/test/unit/lodestar/account.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    const accountPaths = getAccountPaths({rootDir});
+    const validatorDirManager = new ValidatorDirManager(accountPaths);
+    const validatorPubKeys = validatorDirManager.iterDir();
+    expect(spy.calledWith(validatorPubKeys.join("\n"))).to.be.true;
+  });
+
+  it("should make a deposit to validator registration contract", async function() {
+    const validatorId = readdirSync(`${rootDir}/keystores`)[0];
+    expect(validatorId).to.not.be.undefined;
+    unlinkSync(`${rootDir}/keystores/${validatorId}/.lock`)
```
dapplion: .lock files should never be manually deleted. If you have to do this it means something is wrong and the root cause should be addressed first

packages/lodestar-cli/test/unit/lodestar/account.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    ...initDefaults,
+    name: walletName,
+    passphraseFile: "primary.pass",
+  };
+
+  it("should init beacon configuration with --testnet option", async function() {
+    await new Promise(resolve => yargs()
+      .default(initDefaults)
+      .command(init).help().parse(["init"], resolve));
+    await new Promise(resolve => setTimeout(resolve, 6000));
```
dapplion: Why does it need a timeout? doesn't the command resolve when done?

packages/lodestar-cli/test/unit/lodestar/account.test.ts
```diff
@@ -0,0 +1,114 @@
...
+import { account } from "../../../src/cmds/account";
+import { ValidatorDirManager } from "../../../src/validatorDir";
+import { getAccountPaths } from "../../../src/cmds/account/paths";
+import { init } from "../../../src/cmds/init";
+import { TestnetName } from "../../../src/testnets";
+import { medallaConfig } from "../../../src/testnets/medalla";
+import { existsSync, readFileSync, readdirSync, unlinkSync } from "fs";
+
+describe.only("account cli", function() {
+  this.timeout("10 min");
```
dapplion: A unit test should be very quick. If you do something that takes more than seconds move it to e2e tests

packages/lodestar-cli/test/unit/lodestar/account.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    expect(spy.calledWith(walletName)).to.be.true;
+  });
+
+  it("should create new validator", async function() {
+    const wallets = readdirSync(walletsDir);
+    expect(wallets.length > 0).to.be.true;
+    await new Promise(resolve => yargs().default(accountDefaults)
+      .command(account).help().parse(["account", "validator", "create"], resolve));
+    await new Promise(resolve => setTimeout(resolve, 500));
+    expect(spy.calledWith(match.string)).to.be.true;
```
dapplion: It should assert that the validator was created by checking the expected files are there or by some other means. Matching a console.log output seems too brittle.

dapplion (changes requested):

packages/lodestar-cli/test/unit/lodestar/account.test.ts
3xtr4t3rr3str14l: When `parse` resolves, it only resolves the parsing of `init`.  Meanwhile, the `init` handler itself runs as a separate process, and I can't find a way within yargs to be able to hook into the handler to see when it's done running.  So, the timeout is a way to get around this by estimating how long it will take for the underlying handler to finish resolving.  Admittedly, not the best option, and I'm open to other suggestions if you have any.

packages/lodestar-cli/test/unit/lodestar/account.test.ts
3xtr4t3rr3str14l: The reason for this timeout is connected to the reason for the timeout for init below.  By default, an individual unit test function will run for 2000ms, and the beacon init takes 6000ms (as estimated in the Promise resolve timeout below).  So this makes the timeout longer to get past the 2000ms limit.

packages/lodestar-cli/package.json
3xtr4t3rr3str14l: Sorry, meant to clean up some of that ganache-core code that i was using in the setup.  Just cleaned that up.  Does it look better to you now?

3xtr4t3rr3str14l: My interpretation of your comment @mpetrunic was that you were commenting on the way I used the dependency (i.e. not as a dev dep) and used `any` previously for the server listen/close methods.  If that's an incorrect assumption on my part, then what is problematic and/or not typesafe about ganache-core?

packages/lodestar-cli/test/unit/lodestar/account.test.ts
3xtr4t3rr3str14l: fixed

packages/lodestar-cli/test/unit/lodestar/account.test.ts
3xtr4t3rr3str14l: resolved via 7a04dea

packages/lodestar-cli/package.json
mpetrunic: builderevm is alternative to ganache-core so I'm suggesting to swap them since ganache-core doesn't have types or they don't work and has some native dependencies that are hard compile on some OS

3xtr4t3rr3str14l: ok cool.  just swapped ganache-core for buidler in dc0d18843c9d411d976550254ed152fa65960774

mpetrunic:
lodestar unit tests have error but for some reason check has passed. Seems like those lodestar-cli tests are taking quite long (maybe make them e2e)?

dapplion:
> lodestar unit tests have error but for some reason check has passed. Seems like those lodestar-cli tests are taking quite long (maybe make them e2e)?

given that they run the yargs instance directly I would make them e2e too

mpetrunic (dismissed):

3xtr4t3rr3str14l:
@mpetrunic sorry, had to merge from master.  everything up-to-date now

packages/lodestar-cli/test/e2e/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const rootDir = ".tmp";
```
mpetrunic, line 1: Maybe it would be better to use tmp directory dependency just to make sure we don't collide with some directory someone was using and delete him that

mpetrunic (approved):

dapplion (approved):

packages/lodestar-cli/test/e2e/constants.ts
3xtr4t3rr3str14l: sorry, just saw this after i merged.  do you mean that we should rename it to something like `.tmpLodestarE2E` or something like that?

wemeetagain: There's a library called `tmp` which can be used, it's used in `lodestar/test`

3xtr4t3rr3str14l: ahh ok.  i'll start whipping up a PR for that here in a sec

Issue #1150, opened 2020-07-13
EpochContext convenience methods


wemeetagain:
**Is your feature request related to a problem? Please describe.**
We should provide a full-featured interface to EpochContext.
It's nice to have a consistent interface for interacting with EpochContext.

We could add some useful methods to EpochContext, which may lead to more readable code, better devex.

**Describe the solution you'd like**
Add convenience methods `getAttestingIndices` and `getCommitteeAssignment` to `EpochContext`.
Consider adding state transition methods to `EpochContext`. Eg: `processSlots`, `stateTransition`.
Use in place of lower level functions, where applicable.

**Describe alternatives you've considered**
Leave as is.

**Additional context**
See https://github.com/ChainSafe/lodestar/pull/1131#discussion_r453713607

wemeetagain:
Resolved in #1235

PR #1278, opened 2020-08-01
Gossip validation logging

mpetrunic:
- refactored gossip validation so it's not 400lines file (needs to be improved still)
- added logging when some gossip validation fails
- faster subnet calculations and block verification
- faster gossip validation

Since you already applied fix to multinode test branch, I can prepare this to merge separately.

TODO: fix lint and unit tests

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/fast/util/attestation.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/util/attestation.ts
@@ -1,0 +1,23 @@
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
+import {Attestation, ATTESTATION_SUBNET_COUNT} from "@chainsafe/lodestar-types";
+import {computeSlotsSinceEpochStart} from "../../util";
+import {EpochContext} from "../index";
+
+/**
+ * Compute the correct subnet for an attestation for Phase 0.
+ */
+export function computeSubnetForAttestation
+(config: IBeaconConfig, epochCtx: EpochContext, attestation: Attestation): number {
+  const {slot, index} = attestation.data;
+  return computeSubnetForSlot(config, epochCtx, slot, index);
+}
+
+/**
+ * Compute the correct subnet for a slot/committee index for Phase 0.
+ */
+export function computeSubnetForSlot
+(config: IBeaconConfig, epochCtx: EpochContext, slot: number, committeeIndex: number): number {
+  const slotsSinceEpochStart = computeSlotsSinceEpochStart(config, slot);
+  const committeesSinceEpochStart = epochCtx.getCommitteeCountAtSlot(slot) * slotsSinceEpochStart;
+  return (committeesSinceEpochStart + committeeIndex) % ATTESTATION_SUBNET_COUNT;
+}
--- a/packages/lodestar-beacon-state-transition/src/util/attestation.ts
+++ b/packages/lodestar-beacon-state-transition/src/util/attestation.ts
@@ -138,22 +138,3 @@ export function isUnaggregatedAttestation(attestation: Attestation): boolean {
   }
   return count === 1;
 }
-
-/**
- * Compute the correct subnet for an attestation for Phase 0.
- */
-export function computeSubnetForAttestation
-(config: IBeaconConfig, state: BeaconState, attestation: Attestation): number {
-  const {slot, index} = attestation.data;
-  return computeSubnetForSlot(config, state, slot, index);
-}
-
-/**
- * Compute the correct subnet for a slot/committee index for Phase 0.
- */
-export function computeSubnetForSlot
-(config: IBeaconConfig, state: BeaconState, slot: number, committeeIndex: number): number {
-  const slotsSinceEpochStart = computeSlotsSinceEpochStart(config, slot);
-  const committeesSinceEpochStart = getCommitteeCountAtSlot(config, state, slot) * slotsSinceEpochStart;
-  return (committeesSinceEpochStart + committeeIndex) % ATTESTATION_SUBNET_COUNT;
-}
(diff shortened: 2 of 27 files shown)


packages/lodestar-beacon-state-transition/src/fast/util/attestation.ts
```diff
@@ -0,0 +1,23 @@
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
+import {Attestation, ATTESTATION_SUBNET_COUNT} from "@chainsafe/lodestar-types";
+import {computeSlotsSinceEpochStart} from "../../util";
+import {EpochContext} from "../index";
+
+/**
+ * Compute the correct subnet for an attestation for Phase 0.
+ */
+export function computeSubnetForAttestation
```
wemeetagain, line 9: Can you put these as methods on `EpochContext`? (in the same vein as https://github.com/ChainSafe/lodestar/issues/1150)

`config` won't be needed as an arg as its included in `epochCtx.config`

packages/lodestar-beacon-state-transition/src/util/attestation.ts
```diff
@@ -138,22 +138,3 @@ export function isUnaggregatedAttestation(attestation: Attestation): boolean {
   }
   return count === 1;
 }
-
-/**
- * Compute the correct subnet for an attestation for Phase 0.
- */
-export function computeSubnetForAttestation
```
wemeetagain, line 145: We should keep the slow versions for now.

Maybe we eventually rm -r all of the slow implementation?

packages/lodestar-beacon-state-transition/src/util/block.ts
```diff
@@ -19,3 +20,18 @@ export function verifyBlockSignature(
     signedBlock.signature.valueOf() as Uint8Array,
   );
 }
+
+export function verifyBlockSignatureFast(
```
wemeetagain, line 24: Can you make this an epochCtx convenience method? Can you move this to `src/fast` subdir?

If you're worried about epochContext.ts getting big, you could still define the method in a separate file.

packages/lodestar-cli/src/cmds/beacon/cmds/run/run.ts
```diff
@@ -34,7 +36,13 @@ export async function runHandler(options: IBeaconOptions): Promise<void> {
 
   const config = getBeaconConfig(options.preset, options.params);
   const libp2p = await createNodeJsLibp2p(peerId, options.network);
-  const logger = new WinstonLogger();
+  const loggerTransports = [
```
wemeetagain, line 39: We should merge master into dapplion/e2e+ so we don't get these diffs here

packages/lodestar/src/chain/utils.ts
```diff
@@ -0,0 +1,9 @@
+import {BeaconBlock} from "@chainsafe/lodestar-types";
+
+export function getBlockAttestationCount(block: BeaconBlock): number {
+  return Array.from(
```
wemeetagain, line 4: If this is to be used, and not just commented/used in debug settings, we should use readOnlyMap here instead of Array.from.

wemeetagain (changes requested):

packages/lodestar-beacon-state-transition/src/util/attestation.ts
mpetrunic: Well they aren't used in state transition so maybe we should pull it out allthogether?

packages/lodestar-beacon-state-transition/src/util/block.ts
mpetrunic: How would that be epoch convinience method?

packages/lodestar/src/chain/utils.ts
mpetrunic: That just for debugging. Too many iterations to keep it in log method

packages/lodestar-beacon-state-transition/src/util/attestation.ts
wemeetagain: My current thought is eventually we could split state transition into two packages, normal and fast (like eth2spec and eth2fastspec python libraries). Slow implementations usually look like `fn(config, state, a,b,c)`, while fast looks like `epochCtx.fn(a, b,c)` or `epochCtx.fn(state, a, b, c)` depending if state is needed (in most cases it is not). We can re-export some low-level utils from slow in the fast package. And lodestar could import solely from fast.

These attestation helpers are a weird case, since they're not used directly by state transition, but still require committee info. Imo we leave them in state transition as state transition "helpers/misc utilities".

packages/lodestar-beacon-state-transition/src/util/block.ts
wemeetagain: `epochCtx.verifyBlockSignature(state, signedBlock)`
Or we might want to keep fork information in the epochCtx? `epochCtx.verifyBlockSignature(signedBlock)`

packages/lodestar-cli/src/cmds/beacon/cmds/run/run.ts
dapplion: What about merging `dapplion/e2e+` to `master` as skipped and un-skip them in the PR that is able to fix it's issues?

packages/lodestar-beacon-state-transition/src/util/block.ts
mpetrunic: I guess that's fine, but it seems like clutter of epoch ctx for code that executed exactly in one place :smile:

wemeetagain:
Is this PR dead? seems part of this was resolved in #1352

mpetrunic:
> Is this PR dead? seems part of this was resolved in #1352

Yes, splitting into smaller PRs, will close once all gossip topics are fixed, keeping for quick reference :)

wemeetagain:
@mpetrunic can we close?

Issue #885, opened 2020-05-06
Enable strict null checking


lastmjs:
As I was fixing this issue: https://github.com/ChainSafe/lodestar/issues/880, I realized that the main `tsconfig.json` has `strictNullChecks` set to false. I believe that if this had been set to true, then the issue from https://github.com/ChainSafe/lodestar/issues/880 would already have been apparent. There are probably many more hidden problems of a similar nature lurking in the codebase.

We need to remove `strictNullCheck`: false from `tsconfig.json`. This will probably cause a massive amount of TypeScripte errors throughout the entire codebase. The longer we wait, the harder it will be to make the switch.

Also it might be good to have a null/undefined strategy if there isn't one already. Such as always using undefined, always using null, or deciding when to use each. There's also the option of never using undefined or null if possible, such as creating types that are more descriptive, i.e. `NOT_SET`.

Example:

```typescript
  public clock: IBeaconClock | "NOT_SET";
```
You could also say "NOT_INITIALIZED", "NOT_STARTED", or something. I have found this to be a pretty nice way to do things, not without its own problems of course. `null` and `undefined` are just such problematic and broad types.

wemeetagain:
unstale

dapplion:
I will start to tackle this issue progressively up to the bigger packages to prevent creating too many conflicts at once. I'll follow the topological order of this repo's packages

```
lion@lion-gram:~/Code/eth2.0/lodestar$ ./node_modules/.bin/lerna list --toposort
lerna notice cli v3.20.2
@chainsafe/lodestar-params
@chainsafe/lodestar-utils
@chainsafe/lodestar-spec-test-util
@chainsafe/lodestar-types
@chainsafe/lodestar-config
@chainsafe/lodestar-beacon-state-transition
@chainsafe/lodestar-validator
@chainsafe/lodestar
@chainsafe/lodestar-cli
```

PR #1284, opened 2020-08-01
Set strictNullChecks = true in lodestar-beacon-state-transition

dapplion:
Parallelizable effort to tackle https://github.com/ChainSafe/lodestar/issues/885. Sets `strictNullChecks = true` in a single package so eventually it can be set in the root tsconfig.

### Specific issues discussion

- **Cast empty arrays to ssz type List<T>**: We may want to have the ssz library export empty lists of functions that create type-safe empty lists to prevent unsafe casting.
- **Cast arrays to ssz type List<T>**: Is typecasting good enough here? Maybe it would make sense to have a casting function to do that more safely.
- **Assert Number is not undefined on top of Number.isSafeInteger**: Satisfy the tsc compiler
- **Add null to getCommitteeAssignment return signature**: Consumers of this method will be unsafe until they have `strictNullChecks = true`.
- **Ensure pubkeys array is not empty before reducing it**: Even if it never happens, let's make the tsc compiler happy. 
- **Ensure shuffling properties are not undefined before accessing them**: This required quite a lot of changes. Do these properties really have to be defined as maybe `undefined`? Why may they be undefined?


packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts
```diff
@@ -186,15 +189,15 @@ export class EpochContext {
...
+            validators: committee as List<number>,
             committeeIndex: i,
             slot,
           };
         }
       }
     }
 
-    return null;
+    throw Error("no committeeAssignment");
```
mpetrunic: shouldn't be error it's perfectly valid not to have assignment

packages/lodestar-beacon-state-transition/src/util/duties.ts
```diff
@@ -35,15 +36,15 @@ export function getCommitteeAssignment(
...
+          validators: committee as List<number>,
           committeeIndex: i,
           slot,
         };
       }
     }
   }
 
-  return null;
+  throw Error("No committeeAssignment");
```
mpetrunic: same as above

packages/lodestar-beacon-state-transition/src/epoch/finalUpdates.ts
```diff
@@ -16,7 +17,7 @@ export function processFinalUpdates(config: IBeaconConfig, state: BeaconState):
   const nextEpoch = currentEpoch + 1;
   // Reset eth1 data votes
   if (nextEpoch % config.params.EPOCHS_PER_ETH1_VOTING_PERIOD === 0) {
-    state.eth1DataVotes = [];
+    state.eth1DataVotes = [] as Eth1Data[] as List<Eth1Data>;
```
wemeetagain, line 20: I think these can be casted directly to `List<X>`

dapplion: Typescript errors with:
```
Conversion of type 'never[]' to type 'List<Eth1Data>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  The types returned by 'pop()' are incompatible between these types.
    Type 'undefined' is not comparable to type 'Eth1Data'.ts(2352)
```
Because it understands `[]` as `never[]`

wemeetagain (dismissed):

wemeetagain:
tsc errors compiling tests

dapplion:
I've added two helper functions to cast to SSZ `List` type. @wemeetagain is `lodestar-types` a good place to host them?

```ts
export function emptyList<T>(): List<T> {
  return [] as T[] as List<T>;
}
```

```ts
export function toList<T>(arr: T[]): List<T> {
  return arr as List<T>;
}
```

If turning an array to an SSZ List requires some logic in the future it will be centralized in these functions.

dapplion:
@wemeetagain @mpetrunic Is there a cleaner solution for the type errors of `previousShuffling` and similar? Having to call
```ts
if (epochCtx.previousShuffling === undefined) throw Error("previousShuffling not defined");
```
on every use doesn't seem like the best solution.

wemeetagain:
> I've added two helper functions to cast to SSZ List type. 
> If turning an array to an SSZ List requires some logic in the future it will be centralized in these functions.

Generally, I don't think we should be solving these type annoyances using runtime tools. The types are here to assist the correctness of the runtime, but especially in trivially-correct cases, it doesn't make sense to adjust the runtime to appease the types.
In this case, there won't be any case where turning an array into an ssz list requires logic. `List` and `Vector` interfaces are designed to be subsets of `Array` (and `TypedArray`).
I think using a function here is not necessary and less clear than just typecasting.

>  Is there a cleaner solution for the type errors of previousShuffling and similar?

You can do `epochCtx.previousShuffling!` to assert that previousShuffling is not undefined.

packages/lodestar-beacon-state-transition/test/unit/stateTransition/epoch/fork.test.ts
```diff
@@ -18,11 +18,11 @@ describe("processForkChanged", () => {
   });
 
   afterEach(() => {
-    config.params.ALL_FORKS = undefined;
+    config.params.ALL_FORKS = [];
   });
 
   it("should not update fork if no matched next fork", () => {
-    config.params.ALL_FORKS = undefined;
+    config.params.ALL_FORKS = [];
```
dapplion, line 25: Is this equivalent? Tests seem to pass but I would love an OK on this change.

dapplion:
> You can do `epochCtx.previousShuffling!` to assert that previousShuffling is not undefined.

After investigating, the Non-Null Assertion Operator only exist in the Typescript domain. `previousShuffling` will have a possible value of undefined until the code is changed, so asserting that it is not undefined provides no value. If for some reason it becomes undefined, it will still crash at runtime

dapplion:
@wemeetagain Following your suggestion I've added `state` in the `EpochContext` constructor. All `EpochContext` instantiation in non-test code were immediately followed by a method call adding state. So I've refactored the instantiations to

**from**
```ts
const epochCtx = new EpochContext(config);
epochCtx.loadState(state);
```
**to**
```ts
const epochCtx = new EpochContext(config, state);
```

With this change, all `*Shuffling` properties don't need to be optional, so no checks for undefined are necessary.

Some test code required extra modifications to inject empty state.

packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts
```diff
@@ -43,19 +43,19 @@ export class EpochContext {
...
+  public previousShuffling: IEpochShuffling;
+  public currentShuffling: IEpochShuffling;
+  public nextShuffling: IEpochShuffling;
   public config: IBeaconConfig;
+  public state: BeaconState;
 
-  constructor(config: IBeaconConfig) {
+  constructor(config: IBeaconConfig, state: BeaconState) {
     this.config = config;
+    this.state = state;
```
dapplion: Is there any problem with keeping the state as an internal property? Otherwise, is there a better way to do the `copy` method below without keeping an internal reference to the state?

wemeetagain: I don't think we want to keep the state as an internal property. We could make the constructor look like:
`constructor({config, state, epochCtx}: {config: IBeaconConfig; state?: BeaconState; epochCtx?: EpochContext})`, and switch behavior to either copy or loadState, depending on if `state` or `epochCtx` is provided.

Tbh this feels strange, though I guess this is buying us that we won't be able to misuse the EpochContext.

What if we did something like

```typescript
interface EpochContext {
  shuffling: EpochShuffling;
  ...
}

interface UninitializedEpochContext {
  shuffling?: EpochShuffling;
  // no copy method
}

class EpochContext {
  // don't use directly
  constructor(config) {}
  static create(config): UninitializedEpochContext {}
  loadState(state): EpochContext {}
  copy(): EpochContext {}
}
```

packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts
```diff
@@ -43,68 +36,74 @@ export class EpochContext {
...
+  }) {
+    this.pubkey2index = ctx.pubkey2index;
+    this.index2pubkey = ctx.index2pubkey;
+    this.proposers = ctx.proposers;
+    this.previousShuffling = ctx.previousShuffling;
+    this.currentShuffling = ctx.currentShuffling;
+    this.nextShuffling = ctx.nextShuffling;
+    this.config = ctx.config;
+  }
+
```
dapplion: This approach allows to do a copy without having to edit values manually, nor use `Object.create`. At the same time by using the static method `fromState` the API is very simple to consume and makes Typescript happy as the shuffling are defined when constructing the instance.

packages/lodestar-beacon-state-transition/src/fast/util/pubkeys.ts
```diff
@@ -0,0 +1,26 @@
...
+    throw new Error("Pubkey indices have fallen out of sync");
+  }
+  const newCount = state.validators.length;
+  for (let i = currentCount; i < newCount; i++) {
+    const pubkey = state.validators[i].pubkey.valueOf() as Uint8Array;
+    pubkey2index.set(pubkey, i);
+    index2pubkey.push(PublicKey.fromBytes(pubkey));
+  }
+}
+
```
dapplion: Since some code was shared between `EpochContext` instance methods and static methods, I pulled the logic into an external utility for easier reusing. Let me know if it's best to have this utility as a static method of `EpochContext`. No-one is using this utility outside of `EpochContext` so it seemed cleaner to pull it out.

packages/lodestar-beacon-state-transition/src/fast/util/resetProposers.ts
```diff
@@ -0,0 +1,25 @@
...
+      computeProposerIndex(
+        config,
+        state,
+        currentShuffling.activeIndices,
+        hash(Buffer.concat([epochSeed, intToBytes(slot, 8)])),
+      )
+    );
+  }
+  return proposers;
+}
```
dapplion: Same as [comment above ^](https://github.com/ChainSafe/lodestar/pull/1284/files#r467475846)

wemeetagain (dismissed):

wemeetagain:
Unit test errors

dapplion:
Reverted changes regarding the `EpochContext` and used @wemeetagain suggestion to just mark the shufflings with the non-null assertion operator. All other paths are too complicated, require too many modifications, so are not worth the reward.

wemeetagain (approved):

PR #1285, opened 2020-08-01
Set strictNullChecks = true in lodestar-validator

dapplion:
Parallelizable effort to tackle https://github.com/ChainSafe/lodestar/issues/885. Sets `strictNullChecks = true` in a single package so eventually it can be set in the root tsconfig.

### Specific issues discussion

- **Throw Error if failed to createAttestation**: Is this the desired behaviour?
- **tsc friendly way of accessing Map values**: tsc does not like using `.has()`, better to access directly with `.get()` and then check for undefined
- **Set non initialized variables in ClassX as optional**: Is this the best **easy** solution to this problem for now? Eventually the setup sequence could be improved to prevent those.


mpetrunic (dismissed):

packages/lodestar-validator/test/utils/block.ts
```diff
@@ -1,26 +0,0 @@
-import { BeaconBlock } from "@chainsafe/lodestar-types";
```
dapplion, line 1: Removed duplicate function `generateEmptyBlock` that is already used from `packages/lodestar`

mpetrunic (dismissed):

dapplion:
I don't understand how but seems that `CI Tests / Quick tests (pull_request) ` is running the tests off a different commit. The code that causes the error does not exist on this branch's latest commit. Any ideas what may be happening?

mpetrunic:
> I don't understand how but seems that `CI Tests / Quick tests (pull_request) ` is running the tests off a different commit. The code that causes the error does not exist on this branch's latest commit. Any ideas what may be happening?

PR CI merges master first and runs on that commit (aka what will happen once you merge it). Try to merge master and see what happens?

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -170,7 +171,9 @@ export class AttestationService {
 
     if (duty.isAggregator) {
       setTimeout(() => {
-        this.aggregateAttestations(duty.attesterIndex, duty, attestation, fork, genesisValidatorsRoot);
+        if (attestation) {
```
mpetrunic, line 174: we Already check that 6 lines above

mpetrunic (approved):

packages/lodestar-validator/src/services/attestation.ts
dapplion: I know but since the variable is declared with `let` Typescript thinks that the variable can change to undefined again between the check and the execution of the timeout. So within the timeout block, attestation = T | undefined

wemeetagain (approved):

PR #1286, opened 2020-08-01
Fix deposit data decoding

wemeetagain:
See https://discordapp.com/channels/593655374469660673/593655641445367808/738871830567452684

### DIFF
--- a/packages/lodestar-cli/src/depositContract/depositData.ts
+++ b/packages/lodestar-cli/src/depositContract/depositData.ts
@@ -21,10 +21,12 @@ export function decodeEth1TxData(
   const {
     deposit_data_root: root,
   } = inputs;
-  // attach `amount` to decoded deposit inputs so it can be parsed to a DepositData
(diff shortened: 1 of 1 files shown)


twoeths (approved):

GregTheGreek (approved):
This does solve that error, but theres still an error when using infura endpoint. I've opened a related issue on ethers i think there is a discrepancy.

wemeetagain:
Can you sign a transaction using infura? We're submitting a raw tx to the rpc/ipc endpoint to be signed and broadcast, so if your provider can't sign a tx and provide the amount requested (eg: 32 eth), there will be an error.

For me running this command, I ran into very obscure-looking ethers issues until I realized my geth instance had its account locked. Once I unlocked the account, it ran w/o error.

GregTheGreek:
waiiit, it looks the provider on the wallet is actually nulll.... I tihnk we need to add a check to the eth1 signer that the provider actually exists. debugging.

PR #1290, opened 2020-08-02
Set strictNullChecks true in lodestar/api/impl

dapplion:

### DIFF
--- a/packages/lodestar/src/api/impl/beacon/blocks/index.ts
+++ b/packages/lodestar/src/api/impl/beacon/blocks/index.ts
@@ -32,17 +32,17 @@ export class BeaconBlockApi implements IBeaconBlocksApi {
       }
       const nonFinalizedBlockSummaries
           = this.chain.forkChoice.getBlockSummaryByParentBlockRoot(filters.parentRoot.valueOf() as Uint8Array);
-      result.push(... await Promise.all(
-        nonFinalizedBlockSummaries.map(async (summary) => {
-          const block = await this.db.block.get(summary.blockRoot);
+      for (const summary of nonFinalizedBlockSummaries) {
+        const block = await this.db.block.get(summary.blockRoot);
+        if (block) {
           const cannonical = this.chain.forkChoice.getCanonicalBlockSummaryAtSlot(block.message.slot);
-          return toBeaconHeaderResponse(
+          result.push(toBeaconHeaderResponse(
             this.config,
             block,
             this.config.types.Root.equals(cannonical.blockRoot, summary.blockRoot)
-          );
-        })
-      ));
+          ));
+        }
+      }
       return result.filter(
         (item) =>
         //skip if no slot filter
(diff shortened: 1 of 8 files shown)


packages/lodestar/src/api/impl/node/node.ts
```diff
@@ -43,6 +46,7 @@ export class NodeApi implements INodeApi {
   public async getPeers(): Promise<NodePeer[]> {
     return this.network.getPeers().map((peer) => {
       const conn = this.network.getPeerConnection(peer);
+      if (!conn) throw Error(`No peer connection for ${peer.id.toString()}`);
```
dapplion: Should it throw an error here? Or maybe ignore this peer

packages/lodestar/src/api/impl/beacon/blocks/index.ts
```diff
@@ -32,49 +32,55 @@ export class BeaconBlockApi implements IBeaconBlocksApi {
...
+      const canonicalBlock = await this.chain.getBlockAtSlot(filters.slot);
+      //skip slot
+      if(!canonicalBlock) {
+        return [];
+      }
+      const canonicalRoot = this.config.types.BeaconBlock.hashTreeRoot(canonicalBlock.message);
+      result.push(toBeaconHeaderResponse(this.config, canonicalBlock, true));
+
+      //fork blocks
+      for (const summary of this.chain.forkChoice.getBlockSummariesAtSlot(filters.slot)) {
```
dapplion: Is it necessary to parallelize calls to `this.db.block.get` with Promise.all?

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -208,6 +211,11 @@ export class ValidatorApi implements IValidatorApi {
       throw Error("No matching attestations found for attestationData");
     }
 
+    const aggregatorIndex = epochCtx.pubkey2index.get(aggregator);
+    if (aggregatorIndex === undefined) {
+      throw Error("Aggregator pubkey not in epochCtx");
```
dapplion: Should this error contain the pubkey that is not found?

packages/lodestar/src/api/impl/beacon/blocks/index.ts
```diff
@@ -32,49 +32,55 @@ export class BeaconBlockApi implements IBeaconBlocksApi {
       }
       const nonFinalizedBlockSummaries
           = this.chain.forkChoice.getBlockSummaryByParentBlockRoot(filters.parentRoot.valueOf() as Uint8Array);
-      result.push(... await Promise.all(
-        nonFinalizedBlockSummaries.map(async (summary) => {
-          const block = await this.db.block.get(summary.blockRoot);
+      for (const summary of nonFinalizedBlockSummaries) {
+        const block = await this.db.block.get(summary.blockRoot);
```
mpetrunic, line 36: Original implementation ran this "concurrently" but it's api so it's ok for now

mpetrunic: cannonical can also be null

packages/lodestar/src/api/impl/beacon/blocks/index.ts
mpetrunic: well it's suppose be faster but it's api so it's not really important

packages/lodestar/src/api/impl/node/node.ts
mpetrunic: We should ignore.  Not sure but I think network.getPeers doesn't return peers without connection anyways

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -75,6 +75,7 @@ export class ValidatorApi implements IValidatorApi {
     graffiti = ""
   ): Promise<BeaconBlock> {
     const validatorIndex = (await this.chain.getHeadEpochContext()).pubkey2index.get(validatorPubkey);
+    if (validatorIndex === undefined) throw Error("Validator pubKey not in epochCtx");
```
mpetrunic: Not in this PR but we should make this error more sensible so we can trow 400 in case of this error.

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -159,15 +162,15 @@ export class ValidatorApi implements IValidatorApi {
     }
     const validatorIndexes = validatorPubKeys.map((key) => {
       const validatorIndex = epochCtx.pubkey2index.get(key);
-      if (!Number.isInteger(validatorIndex)) {
-        throw Error(`Validator pubkey ${Buffer.from(key).toString("hex")} not in epochCtx`);
+      if (validatorIndex === undefined || !Number.isInteger(validatorIndex)) {
+        throw Error(`Validator pubKey ${Buffer.from(key).toString("hex")} not in epochCtx`);
```
mpetrunic:
```suggestion
        throw Error(`Validator pubKey ${toHexString(key)} not in epochCtx`);
```

packages/lodestar/src/api/impl/validator/validator.ts
mpetrunic: sure. We will need to change this anyways to return meaningful error messages in api

mpetrunic (approved):

packages/lodestar/src/api/impl/validator/validator.ts
dapplion: Opened https://github.com/ChainSafe/lodestar/issues/1293

mpetrunic (approved):

PR #1291, opened 2020-08-02
fix ethers wallet provider

GregTheGreek:
`.connect()` returns a new instance and does not modify the original wallet.

### DIFF
--- a/packages/lodestar-cli/src/util/ethers.ts
+++ b/packages/lodestar-cli/src/util/ethers.ts
@@ -23,8 +23,7 @@ export async function getEthersSigner({
     const eth1Provider = rpcUrl
       ? new ethers.providers.JsonRpcProvider(rpcUrl)
       : ethers.getDefaultProvider();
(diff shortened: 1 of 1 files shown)


packages/lodestar-cli/src/util/ethers.ts
```diff
@@ -23,8 +23,8 @@ export async function getEthersSigner({
     const eth1Provider = rpcUrl
       ? new ethers.providers.JsonRpcProvider(rpcUrl)
       : ethers.getDefaultProvider();
-    wallet.connect(eth1Provider);
-    return wallet;
+    const providerWallet = wallet.connect(eth1Provider);
+    return providerWallet;
```
wemeetagain: Maybe just:
```suggestion
    return wallet.connect(eth1Provider);
```

wemeetagain (dismissed):

mpetrunic (approved):

dapplion (approved):

PR #1296, opened 2020-08-02
Set strictNullChecks true in lodestar/chain

dapplion:

### DIFF
--- a/packages/lodestar/src/chain/attestation.ts
+++ b/packages/lodestar/src/chain/attestation.ts
@@ -113,7 +113,7 @@ export class AttestationProcessor implements IAttestationProcessor {
     const targetSlot = computeStartSlotAtEpoch(this.config, target.epoch);
     const ancestor = this.forkChoice.getAncestor(attestation.data.beaconBlockRoot as Uint8Array, targetSlot);
     assert.true(
-      ancestor && this.config.types.Root.equals(target.root, ancestor),
+      Boolean(ancestor && this.config.types.Root.equals(target.root, ancestor)),
       "FFG and LMD vote must be consistent with each other");
     const stateCtx = await this.db.stateCache.get(block.stateRoot);
     assert.true(
--- a/packages/lodestar/src/chain/attestation.ts
+++ b/packages/lodestar/src/chain/attestation.ts
@@ -116,10 +116,9 @@ export class AttestationProcessor implements IAttestationProcessor {
       Boolean(ancestor && this.config.types.Root.equals(target.root, ancestor)),
       "FFG and LMD vote must be consistent with each other");
     const stateCtx = await this.db.stateCache.get(block.stateRoot);
-    assert.true(
-      !!stateCtx,
-      `Missing state context for attestation block with stateRoot ${toHexString(block.stateRoot)}`
-    );
+    if (stateCtx === null) {
+      throw Error(`Missing state context for attestation block with stateRoot ${toHexString(block.stateRoot)}`)
+    }
     const validators = stateCtx.epochCtx.getAttestingIndices(
       attestation.data,
       attestation.aggregationBits
--- a/packages/lodestar/src/chain/blocks/pool.ts
+++ b/packages/lodestar/src/chain/blocks/pool.ts
@@ -27,9 +27,9 @@ export class BlockPool {
 
   public addPendingBlock(job: IBlockProcessJob): void {
     const key = this.getKey(job.signedBlock);
-    const jobPool = this.pool.get(key);
-    if(jobPool) {
-      jobPool.push(job);
+    const pendingBlockPool = this.pool.get(key);
+    if(pendingBlockPool) {
+      pendingBlockPool.push(job);
     } else {
       this.pool.set(key, [job]);
       //this prevents backward syncing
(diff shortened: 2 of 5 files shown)


packages/lodestar/src/chain/attestation.ts
```diff
@@ -113,13 +113,12 @@ export class AttestationProcessor implements IAttestationProcessor {
...
+      Boolean(ancestor && this.config.types.Root.equals(target.root, ancestor)),
       "FFG and LMD vote must be consistent with each other");
     const stateCtx = await this.db.stateCache.get(block.stateRoot);
-    assert.true(
-      !!stateCtx,
-      `Missing state context for attestation block with stateRoot ${toHexString(block.stateRoot)}`
-    );
+    if (stateCtx === null) {
+      throw Error(`Missing state context for attestation block with stateRoot ${toHexString(block.stateRoot)}`)
+    }
```
dapplion, line 121: Is it necessary to throw the error with `assert`? If not better, because tsc can see that in this branch `stateCtx` is defined

packages/lodestar/src/chain/blocks/pool.ts
```diff
@@ -27,8 +27,9 @@ export class BlockPool {
 
   public addPendingBlock(job: IBlockProcessJob): void {
     const key = this.getKey(job.signedBlock);
-    if(this.pool.has(key)) {
-      this.pool.get(key).push(job);
+    const jobPool = this.pool.get(key);
+    if(jobPool) {
+      jobPool.push(job);
```
dapplion: Appropriate name?

packages/lodestar/src/chain/chain.ts
```diff
@@ -94,15 +95,17 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
   }
   public async getHeadState(): Promise<TreeBacked<BeaconState>> {
     //head state should always have epoch ctx
-    return (await this.db.stateCache.get(this.forkChoice.headStateRoot())).state;
+    return (await this.getHeadStateContext()).state;
   }
   public async getHeadEpochContext(): Promise<EpochContext> {
     //head should always have epoch ctx
-    return (await this.db.stateCache.get(this.forkChoice.headStateRoot())).epochCtx;
+    return (await this.getHeadStateContext()).epochCtx;
```
dapplion, line 108: Is it okay here to use `this.getHeadStateContext` and deduplicate code?

packages/lodestar/src/chain/chain.ts
```diff
@@ -237,17 +242,16 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
   }
 
   public async waitForBlockProcessed(blockRoot: Uint8Array): Promise<void> {
-    let listener: (signedBlock: SignedBeaconBlock) => void;
     await new Promise((resolve) => {
-      listener = (signedBlock) => {
+      const listener = (signedBlock: SignedBeaconBlock): void => {
         const root = this.config.types.BeaconBlock.hashTreeRoot(signedBlock.message);
         if (this.config.types.Root.equals(root, blockRoot)) {
+          this.removeListener("processedBlock", listener);
```
dapplion, line 249: The logic here should be identical. Is there any reason to call `this.removeListener` outside of the promise?

packages/lodestar/src/chain/chain.ts
```diff
@@ -94,15 +95,17 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
 
   public async getHeadStateContext(): Promise<ITreeStateContext> {
     //head state should always exist
-    return (await this.db.stateCache.get(this.forkChoice.headStateRoot()));
+    const headStateRoot = await this.db.stateCache.get(this.forkChoice.headStateRoot());
+    if (!headStateRoot) throw Error("headStateRoot does not exist");
```
dapplion, line 99: This check is potentially going to have the greatest impact. Since "*head state should always exist*", is throwing an error the best solution?

packages/lodestar/src/chain/attestation.ts
mpetrunic: Well it's not but I don't think there is a way to indicate tsc that this method will throw if param is false

packages/lodestar/src/chain/blocks/pool.ts
mpetrunic: pendingBlockPool would be better

packages/lodestar/src/chain/chain.ts
mpetrunic: yup

packages/lodestar/src/chain/chain.ts
mpetrunic: yup

packages/lodestar/src/chain/chain.ts
mpetrunic: probably not

mpetrunic (commented):
You should probably fix https://github.com/ChainSafe/lodestar/issues/1297
if you add this handling in fork choice

packages/lodestar/src/chain/attestation.ts
dapplion: What do you mean with "if param is false"? Should I do any further changes to this file?

mpetrunic: well it would seem tsc doesn't check method implementation to see if it will throw so nope, we cannot do anything about it, I think

mpetrunic (approved):

wemeetagain (approved):

PR #1301, opened 2020-08-02
lodestar-cli init

### LINKED ISSUE #1155: Refactor beacon config param parsing in CLI
We will want a way to override params, like in beacon run.
Especially important here is GENESIS_FORK_VERSION.
So the chain.params will need to be an input here.

The direction I've been thinking lately is to do something like:
- [x] create a IBeaconParams ssz type, `createIBeaconParams` to create IBeaconParams from a `Record<string, unknown>` / `Json` #1136 
- [x] Use it to help create yargs `Options` if needed. All Options are `type: "string"`, will be parsed using `createIBeaconParams`
- [x] Have a cli-wide params file. (should be copy-pastable from eg: https://github.com/eth2-clients/eth2-testnets/blob/master/shared/altona/config.yaml)
- [x] Have all lodestar cli commands rely on this separate config file - Likely _every_ command needs the same IBeaconParams.
- [x] First `lodestar-cli` bootup should require a config file, or have `lodestar init` to create it, or other similar behavior.

_Originally posted by @wemeetagain in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMjgxMzExNjg4OnYy/pull_request_review_threads/discussion_

### LINKED ISSUE #1165: CLI init
we might have a `lodestar init` to burn in the chain params to disk for use across `beacon`, `validator`, etc. The `init` wouldn't be very useful to run by itself, but its there conceptually as initialization required across lodestar commands. In the same way that `beacon run` should run `beacon init` if it needs to be run.

In the CLI, some initialization will likely be required for use across different commands. This initialization looks like writing some files to disk, for future consumption by cli commands.

Add `init` command to write beacon params to disk. Have `init` called before other commands if chain params doesn't exist on disk.

**Additional context**
Prereq: https://github.com/ChainSafe/lodestar/pull/1136 and separating chain params from `beacon.config.json`

### LINKED ISSUE #1271: Re-org beacon commands
The inital idea behind the different options and commands was as follows (things may have changed, this is the original conception):
- `init` will write a beacon configuration (+ other configuration files) to disk
  - the configuration it writes to disk is, by default, the default values of the Options
  - the `templateConfigFile` overrides any default values
  - cli flags (and env vars) can further override values
  - the `configFile`, in this `init` context, is the _output_ of the command. It isn't used as input, because the purpose of the `init` command is to create the config when it does not exist. Overriding this value sets the location of where to write the generated configuation.

- `run` will run the beacon chain with a configuration (+ other configuration files)
  - the default values are, by default, the default values of the Options
  - the `configFile` overrides any default values
  - cli flags (and env vars) can further override values
  - the `templateConfigFile`, in the `run` context, has no meaning. It isn't used because its purpose is simply as a _template_ for the config file written in the `init` step

My proposal is to refactor the process:
- replace `beacon init` command with `init` with the following responsibilities
  - initialize all configuration to disk (params (point 3 of https://github.com/ChainSafe/lodestar/issues/1155) + beacon configuration + other configuration files)
  - don't overwrite existing configuration, if it exists, check first and bail if it exists
  - support verbose logs
- rename `beacon run` command to `beacon`
- in every command other than `init`, if `init` has not been called yet for that `rootDir`, call it first (possibly suppress this w an explicit cli flag)
- `init` is left as a command that may be run explicitly, for the same reasons `beacon init` was separated from `beacon run` initially: to provide a step where sysadmins can verify/edit on-disk configuration files before running lodestar commands
- `templateConfigFile` is either removed entirely, or only exists as an _advanced option_ in `init` using the original semantics. If it remains, it is intentionally _only_ available as a flag in the `init` command, having no effect in any other command.

_Originally posted by @wemeetagain in https://github.com/ChainSafe/lodestar/issues/1251#issuecomment-666527008_

wemeetagain:
Resolves #1271 
Resolves #1165 
Resolves #1155 

- Add `paramsFile` to GlobalOptions - use this to load params for `IBeaconParams`, used across lodestar cli via `getMergedIBeaconConfig` function
- Add `testnet` to GlobalOptions - testnet will effect `rootDir` if it is not explicitly set. Eg: `--testnet medalla` will point the rootDir to `./.medalla`. This is passed thru the app by using `getGlobalPaths`. `getBeaconPaths`, `getValidatorPaths`, etc all will implicitly call `getGlobalPaths` so paths are always set correctly
- `--testnet` only downloads files if an IBeaconParams config is not already written to disk
- `init` won't overwrite existing configs/files
- Move `lodestar beacon init` to `lodestar init`
- Move `beacon run` to `beacon`
- testnet file configuration now includes downloading a `paramsFile`, a yaml file for IBeaconParams
- Add yaml file processing - for IBeaconParams yaml parsing / writing
- set `beaconDir` to `rootDir`, one less level of nesting

At a high level:
- It is now easier to run a beacon node `lodestar beacon --testnet foo`
- It is now easy run validators for testnets: eg: `lodestar validator --testnet foo`
- It is now easy to make validators and submit deposits for testnets eg: `lodestar account validator create --testnet foo` & `lodestar account  validator deposit --testnet foo`


twoeths:
need to update the docs also

packages/lodestar-cli/src/testnets/index.ts
```diff
@@ -20,6 +20,17 @@ export function getTestnetConfig(testnet: TestnetName): IBeaconNodeOptionsPartia
   }
 }
 
+export function getTestnetParamsUrl(testnet: TestnetName): string | null {
+  switch (testnet) {
+    case "altona":
+      return "https://raw.githubusercontent.com/eth2-clients/eth2-testnets/master/shared/altona/config.yaml";
+    case "medalla":
+      return "https://raw.githubusercontent.com/eth2-clients/eth2-testnets/master/shared/medalla/config.yaml";
```
mpetrunic, line 28: maybe pin links to commit in case dir structure changes?

mpetrunic (dismissed):

packages/lodestar-cli/src/testnets/index.ts
```diff
@@ -20,6 +20,17 @@ export function getTestnetConfig(testnet: TestnetName): IBeaconNodeOptionsPartia
   }
 }
 
+export function getTestnetParamsUrl(testnet: TestnetName): string | null {
+  switch (testnet) {
+    case "altona":
+      return "https://raw.githubusercontent.com/eth2-clients/eth2-testnets/master/shared/altona/config.yaml";
```
twoeths, line 26: I don't see `DEPOSIT_CHAIN_ID` in this link, probably it prevents me from doing a deposit
```
Starting 1 deposits
 ✖ Error: chainId address mismatch (argument="transaction", value={"to":"0x1234567890123456789012345678901234567890","gasLimit":400000,"value":{"_hex":"0x01bc16d674ec800000","_isBigNumber":true},"data":"0x22895118000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000001203a11b4afbcb000d77cb24e6e14d87c962965fa9632f59dc2502ddecc05a4bcd4000000000000000000000000000000000000000000000000000000000000003091aadbb5fcc07a2215e7056847f07aaa1fff03ead480d5b5da45c3d58c03c72f863ff50447cf670d0ea93eabad77a33d00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000cfa79e4e42951cfbb687df5916899a8d15fdf05d33b0f01d84dd7ac38d1d5200000000000000000000000000000000000000000000000000000000000000608288fa3b2a74f3ad18ba6adfc451a2a34c70aa5f7b47075a51cbf8f48a4f693408b22b8cc7a6728871e45433866d2d26187f4ba7975a59c9053ba28a34d374c12844dfe1444b7c6ec0fe5e09c36683e0d389cfd2a52e8ba599a8a753947be9f5","chainId":1}, code=INVALID_ARGUMENT, version=abstract-signer/5.0.1)
```

twoeths: failed to get medalla deposit events too, maybe it's the same error
```
(node:36549) UnhandledPromiseRejectionWarning: Error: could not detect network (event="noNetwork", code=NETWORK_ERROR, version=providers/5.0.2)
```
I found that chainId in eth1 provider is 1

packages/lodestar-cli/src/util/file.ts
```diff
@@ -16,10 +29,14 @@ export enum FileFormat {
...
 }
 
+
+
 export function parse<T=Json>(contents: string, fileFormat: FileFormat): T {
   switch (fileFormat) {
     case FileFormat.json:
       return JSON.parse(contents);
+    case FileFormat.yaml:
+      return load(contents, {schema: yamlSchema});
```
dapplion, line 39: JSON is a subset of YAML, so actually you can load both with the same yaml.load function

dapplion (commented):
Much needed rework! Awesome job. Tested account commands, init, and beacon and all work fine with `--testnet altona` but for `--testnet medalla` ethers throws with 
```
(node:17620) UnhandledPromiseRejectionWarning: Error: could not detect network (event="noNetwork", code=NETWORK_ERROR, version=providers/5.0.2)
    at Logger.makeError (/home/lion/Code/eth2.0/lodestar/node_modules/@ethersproject/logger/lib/index.js:179:21)
    at Logger.throwError (/home/lion/Code/eth2.0/lodestar/node_modules/@ethersproject/logger/lib/index.js:188:20)
    at RetryProvider.<anonymous> (/home/lion/Code/eth2.0/lodestar/node_modules/@ethersproject/providers/lib/json-rpc-provider.js:335:54)
    at step (/home/lion/Code/eth2.0/lodestar/node_modules/@ethersproject/providers/lib/json-rpc-provider.js:46:23)
    at Object.throw (/home/lion/Code/eth2.0/lodestar/node_modules/@ethersproject/providers/lib/json-rpc-provider.js:27:53)
    at rejected (/home/lion/Code/eth2.0/lodestar/node_modules/@ethersproject/providers/lib/json-rpc-provider.js:19:65)
    at processTicksAndRejections (internal/process/task_queues.js:94:5)
```

Is the Prylabs endpoint still working for Goerli? Or maybe it's a network ID issue?

twoeths:
> Is the Prylabs endpoint still working for Goerli? Or maybe it's a network ID issue?

I modified `https://goerli.prylabs.net` (https instead of http) in `.medalla/beacon.config.json` and 
```# Ethereum PoW Mainnet
DEPOSIT_CHAIN_ID: 5
DEPOSIT_NETWORK_ID: 5```
in `.medalla/config.yaml`

it seems to get through the issue

mpetrunic (approved):

PR #1305, opened 2020-08-03
Fix array option merging

mpetrunic:
- as said in comment using cli to override array option will result in merger
- with this cli flag will replace default


wemeetagain (dismissed):

wemeetagain:
file conflicts

wemeetagain (approved):

PR #1308, opened 2020-08-03
Add waitForGenesisTime

wemeetagain:


packages/lodestar/src/chain/chain.ts
```diff
@@ -372,4 +374,18 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
...
 
+  /**
+   * Wait until now >= genesisTime
+   */
+  private async waitForGenesisTime(genesisTime: number): Promise<void> {
+    let now = Date.now();
+    while (genesisTime - now > 0) {
+      // wait 15s or time to genesis, whichever is shorter
+      const diff = Math.min(genesisTime - now, 15000);
+      this.logger.info("Waiting for genesis", {waitTime: Math.floor(diff / 1000)});
```
mpetrunic: - maybe print it in minutes?
- add time unit to log

dapplion (dismissed):

packages/lodestar/src/chain/chain.ts
mpetrunic: oh noes wait, but it would just print 15 s all the time? I think log should display genesisTime-now not diff :smile: @wemeetagain

3xtr4t3rr3str14l (dismissed):

wemeetagain:
We might consider not merging this, or instead, let `chain.start` finish without waiting for the `GENESIS_DELAY`. (Add a  "waiting  for genesis" log in a different part of the code)
This way the network and other modules can be started at the appropriate time.

packages/lodestar/src/chain/chain.ts
wemeetagain: XD yea, fixed

wemeetagain:
Closing this after discussion on the above comment

PR #1309, opened 2020-08-03
Fix testnet params generation in init

wemeetagain:
`appendTestnetParamsConfig` did not work, because `readParamsConfig` and `writeParamsConfig` are not inverses of each other. Reading doesn't return `IBeaconParams`, just a `Record<string, unknown>`. This lets us merge params from different sources before a single conversion to `IBeaconParams`. Current behavior generates config.yaml with invalid hex strings.

Instead of patching the testnet config.yaml files we download, add params to the testnet beacon.config.json.


mpetrunic (approved):

dapplion:
Much better solution :+1:

PR #1317, opened 2020-08-03
Regular sync v2

mpetrunic:
Still WIP since I didn't yet managed to reach altona head, but looks promising and stable


dapplion:
@mpetrunic @tuyennhv What's the status of this? Should it still be pursued?

mpetrunic:
> @mpetrunic @tuyennhv What's the status of this? Should it still be pursued?

This was open just to provide some ideas, I think it's better to start from scratch at this point

PR #1328, opened 2020-08-04
Regular Sync: Emit syncCompleted if head is up to date

twoeths:
Currently we're before medalla genesis.
Regular sync is started and it notices it's up to date but it didn't emit `syncCompleted` event and peer status timer just continue to run

## Fix:
+ Emit `syncCompleted` in that case
+ Make timer to run per 3 slots instead of per slot

### DIFF
--- a/packages/lodestar/src/sync/regular/naive/naive.ts
+++ b/packages/lodestar/src/sync/regular/naive/naive.ts
@@ -61,6 +61,7 @@ export class NaiveRegularSync extends (EventEmitter as { new(): RegularSyncEvent
     const currentSlot = getCurrentSlot(this.config, state.genesisTime);
     if (headSlot >= currentSlot) {
       this.logger.info(`Regular Sync: node is up to date, headSlot=${headSlot}`);
+      this.emit("syncCompleted");
       return;
     }
     this.currentTarget = headSlot;
--- a/packages/lodestar/src/sync/regular/naive/naive.ts
+++ b/packages/lodestar/src/sync/regular/naive/naive.ts
@@ -62,6 +62,7 @@ export class NaiveRegularSync extends (EventEmitter as { new(): RegularSyncEvent
     if (headSlot >= currentSlot) {
       this.logger.info(`Regular Sync: node is up to date, headSlot=${headSlot}`);
       this.emit("syncCompleted");
+      await this.stop();
       return;
     }
     this.currentTarget = headSlot;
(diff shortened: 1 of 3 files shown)


packages/lodestar/src/sync/regular/naive/naive.ts
```diff
@@ -61,6 +61,7 @@ export class NaiveRegularSync extends (EventEmitter as { new(): RegularSyncEvent
     const currentSlot = getCurrentSlot(this.config, state.genesisTime);
     if (headSlot >= currentSlot) {
       this.logger.info(`Regular Sync: node is up to date, headSlot=${headSlot}`);
+      this.emit("syncCompleted");
```
mpetrunic, line 64: should we also call stop here so it can remove chain listener and close source or sync calls that?

twoeths: `targetSlotRangeSource` and `controller` are not even initialized at that time

mpetrunic: but we registered chain listener which will get called nonetheless

mpetrunic: and targetSlotRangeSource is initialized in constructor

twoeths: oh didn't notice `targetSlotRangeSource` is initialized at both constructor and start()

so let's do everything in stop()

twoeths: have to emit `syncCompleted` separatedly because sync also call `stop()`

packages/lodestar/src/sync/regular/naive/naive.ts
```diff
@@ -62,11 +62,11 @@ export class NaiveRegularSync extends (EventEmitter as { new(): RegularSyncEvent
     if (headSlot >= currentSlot) {
       this.logger.info(`Regular Sync: node is up to date, headSlot=${headSlot}`);
       this.emit("syncCompleted");
+      await this.stop();
       return;
     }
     this.currentTarget = headSlot;
     this.logger.verbose(`Regular Sync: Current slot at start: ${currentSlot}`);
-    this.targetSlotRangeSource = pushable<ISlotRange>();
```
mpetrunic: I would leave this actually, since we should also add restarting regular sync if we encounter peer with higher slot, in which we would all start again and it would fail because of closed source.

mpetrunic (dismissed):

mpetrunic (approved):

PR #1345, opened 2020-08-04
Revert strict yargs

wemeetagain:
Yargs `strict` currently creates an issue for `--options.likeThis`. Reverting until we find a better solution

Reproduce:
```bash
$ lodestar beacon --testnet medalla --metrics.enabled
 ✖ Unknown argument: metrics


twoeths (approved):

GregTheGreek (approved):
Rip. I didn't test this.

PR #1353, opened 2020-08-05
remove IInitOptions

### LINKED ISSUE #1251: Config options not copied over when --template-config-file used
**Describe the bug**

When I run `yarn run cli beacon init --template-config-file altona.json` (where altona.json is the one specified here: https://chainsafe.github.io/lodestar/usage/altona/), the resultant beacon.config.json just contains an empty object, `{}`

**Expected behavior**

I would expect the resultant beacon.config.json to contain all of the options contained in the altona.json

3xtr4t3rr3str14l:
removing IInitOptions since we don't need `templateConfigFile` anymore
will close #1251 in the process


mpetrunic (commented):
Can we sort out types so that we don't have this unsafe castings?

Maybe create initCmdOptions and every command that uses init must have it's options extend initCmdOptions?

wemeetagain:
closing in favor of #1357

PR #1357, opened 2020-08-06
CliCommand builder

dapplion:
#### Problem
Reviewing https://github.com/ChainSafe/lodestar/pull/1241 I found un-scalable that the CLI structure had to be hardcoded in the CLI docs generation script. The ideal scenario would be for the script to consume the CLI itself without having to redefine it's structure again, which would make maintainability problematic.

Yargs does not offer any way to access it's internal cmd structure or registry in a way that would allow automatic docs to be generated. Nor there seem to be any tool out there that do so.

#### Proposed solution

We have a good standard way of defining commands. If we define then in a consumable format instead of using the yargs builder, the resulting command tree object can be both consumed by the documentation buidler and yargs itself. I propose to define commands with:

```ts
export interface ICliCommand<OwnOptions = {}, ParentOptions = {}> {
  command: string;
  describe: string;
  options?: Required<{[key in keyof OwnOptions]: Options}>;
  // 1st arg: any = free own sub command options
  // 2nd arg: subcommand parent options is = to this command options + parent options
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  subcommands?: ICliCommand<any, OwnOptions & ParentOptions>[];
  handler: (args: OwnOptions & ParentOptions) => void;
}
```

The key is that subcommands are recursively available, instead of being fed into the yargs blackbox with `yargs.command` in the command parent builder. It also achieves better type safety, as it will enforce that a command's `OwnOptions` match its options object.

:warning: This PR is a demo with this structure applied to only two commands. **Code owners** reply with yes / no / comments, and if the idea is liked I will apply it to the rest of the commands.

#### To docs

After implementing this, the script of https://github.com/ChainSafe/lodestar/pull/1241 can fed the command data structure and generate the docs without having to manually import and declare everything.


dapplion:
Leaving Teku CLI docs are a goal reference https://docs.teku.pegasys.tech/en/latest/Reference/CLI/CLI-Syntax/ but having them autogenerated

wemeetagain:
So the tradeoff is: If we can accept this non-yargs-standard interface, the helper function, and tweaked usage, we get programmatic access to the command tree and increased type safety.

I think this is really great. :+1:  from me

One thing that might be good to highlight, the difference between Options and Args. (Or whatever you want to call it). Options are objects that describe/define a cli option. Args are the actual values provided.
When does the code expect Options? When does the code expect Args?
I think we should be clear in the naming of these things so we aren't confusing the two related, but different concepts.

dapplion:
@wemeetagain That's a good point, I'm myself confused sometimes with options / args naming. With this distinction the signature `ICliCommand` should be
```ts
interface ICliCommand<OwnArgs = {}, ParentArgs = {}>
```
?

3xtr4t3rr3str14l:
I dig it!  I had thought about how it would be nice to have something like this when working on #1241 so I think this will help a lot.

dapplion:
Changes summary:
 - Converted all commands to use `ICliCommandOptions`. There is a small pitfall with `IBeaconOptions`, since the Options do not match the Args I've to cast the options in `ICliCommandOptions` for commands that use `IBeaconOptions`.
- `ICliCommandOptions` prefers that a command Options and Args match. That way, we can ensure that all Args are actual options and the other way around. For that:
  - Un-nested the `dev` cmd options so Options and Args match.
  - Move the `dev` option `sync.minPeers` to `IBeaconOptions`
  - Split other commands Options in a group which Options and Args match and a group that doesn't.
- I've added a version of the initHandler `initCmd` so it can be easily consumed by other commands.

@3xtr4t3rr3str14l I've added an example script to dynamically generate the docs as a reference here `packages/lodestar-cli/src/docGeneration.ts`, hope it's useful!

packages/lodestar-cli/src/cmds/init/index.ts
```diff
@@ -1,11 +1,11 @@
-import {CommandModule} from "yargs";
+import {ICliCommand, ICliCommandOptions} from "../../util";
 import {IBeaconOptions, beaconOptions} from "../beacon/options";
-import {initHandler} from "./init";
+import {initHandler} from "./handler";
 import {IGlobalArgs} from "../../options";
 
-export const init: CommandModule<IGlobalArgs, IBeaconOptions> = {
+export const init: ICliCommand<IBeaconOptions, IGlobalArgs> = {
```
wemeetagain: Since you renamed `IValidatorCLIOptions` to `IValidatorCLIArgs`, should also rename `IBeaconOptions` to `IBeaconArgs` and any other

wemeetagain (changes requested):
Looks great :)

One minor comment

packages/lodestar-cli/src/cmds/init/index.ts
dapplion: I was unsure about that because in `packages/lodestar` it's called `IBeaconOptions`, won't it create confusion?

wemeetagain: I thought that was `IBeaconNodeOptions` in packages/lodestar

dapplion: Oh, right! No problem then for `IBeaconArgs`

wemeetagain:
file conflicts

packages/lodestar-cli/src/cmds/dev/options.ts
```diff
@@ -1,90 +1,82 @@
...
+};
+
+/**
+ * Add custom defaults different than the ones in `beaconOptions`:
+ * - In dev command we don't wanna connect to other peers, 
+ * - but we do wanna get out of syncing (min peers) 
+ * - and have api enabled by default (as it's used by validator)
+ * Note: use beaconNodeOptions and globalOptions to make sure option key is correct
+ */
+const externalOptionsOverrides: {[k: string]: Options} = {
```
dapplion, line 47: Modified latest @mpetrunic PR edit of the dev options to deduplicate the option definition. It also makes sure the option key is correct, if we change the key over to beacon node options, tsc will error here.

wemeetagain (approved):

PR #1360, opened 2020-08-06
Add enr cli options

wemeetagain:
Depends on https://github.com/ChainSafe/discv5/pull/84 and https://github.com/ChainSafe/discv5/pull/85 and subsequent release

- Set enr UDP/TCP fields based on configured libp2p multiaddr / discv5 multiaddr.
- Override enr values based on enr cli options
- If enr ip or ip6 options are set, ensure that discv5 enrUpdate=false


packages/lodestar-cli/src/db.ts
```diff
@@ -0,0 +1,34 @@
...
+import {getBeaconPaths} from "./cmds/beacon/paths";
+import {mergeConfigOptions} from "./config/beacon";
+import {getBeaconConfig} from "./util";
+
+const rootDir = ".altona";
+const bucket = "badBlock";
+const action = "delete";
+const value = "0xe37b83f1e25ce3ebc2e10fa308413160eb36e4de6d295079104222d67229136c";
+
+(async function() {
```
dapplion: Why does this function need to run in the body of the script?

wemeetagain: doh, this file isn't meant to be added, its just some `beacon db` cli command i've been toying with

packages/lodestar-cli/src/util/enr.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function updateENR(enr: ENR, args: IENRArgs & IBeaconNodeOptions): void {
+  if (args.network.multiaddrs.length) {
+    try {
+      const tcpOpts = Multiaddr(args.network.multiaddrs[0]).toOptions();
+      if (tcpOpts.transport === "tcp") {
+        enr.tcp = tcpOpts.port;
+      }
+    } catch (e) {
+      throw new Error("Invalid tcp multiaddr");
```
dapplion: For sanity check append the error message (at least) after the nice error message. So many things can happen in `Multiaddr`. I would do the same for the error below too

dapplion (approved):

mpetrunic (approved):

PR #1363, opened 2020-08-07
Set strictNullChecks true in lodestar/util

dapplion:


packages/lodestar/src/util/function.ts
```diff
@@ -1,15 +0,0 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
```
dapplion, line 1: This function had a bunch of Typescript (strictNullChecks) errors, but since it's not used anywhere I delete it.

wemeetagain (approved):

PR #1366, opened 2020-08-07
Set strictNullChecks true in lodestar/db

dapplion:


packages/lodestar/src/db/api/beacon/repositories/depositDataRoot.ts
```diff
@@ -29,28 +29,22 @@ export class DepositDataRootRepository extends Repository<number, Root> {
...
+      depositRootTree[key] = value as TreeBacked<Root>;
     }
   }
 
   public async getTreeBacked(depositIndex: number): Promise<TreeBacked<List<Root>>> {
-    if (!this.depositRootTree) {
-      await this.initTree();
-    }
-    const tree = this.depositRootTree.clone();
+    const depositRootTree = await this.getDepositRootTree();
```
dapplion, line 46: Same logic but by calling this function tsc can guarantee `depositRootTree` is defined

wemeetagain (approved):

Issue #1370, opened 2020-08-08
Node v12- v14 support


wemeetagain:
**Describe the bug**
Lodestar does not work, either hangs, or hits errors when using newer versions of node.

**Expected behavior**
Lodestar runs flawlessly on newer versions of node

**Steps to Reproduce**
Install node v12.13.x, v13, v14
run the beacon chain

**Context**
See https://discordapp.com/channels/593655374469660673/593655641445367808/741637064810037258
https://github.com/ChainSafe/lodestar/issues/1320

dapplion:
Could we recap in this issue the specific errors or reasons why Lodestar breaks in higher versions?

- Issue fixed downgrading to Node v12 https://github.com/ChainSafe/lodestar/issues/1320: `TypeError: Cannot perform %TypedArray%.prototype.set on a detached ArrayBuffer`

wemeetagain:
v12, later minor versions: Lodestar hangs on boot
v14, detached Arraybuffer error

wemeetagain:
I just tried running off latest master with v12.18.3, didn't have any problems. Can anyone else confirm?

dapplion:
We should revisit this. It would be a valuable experiment to try to the current release in Node v14 and see what happens

dapplion:
I've pushed a branch running the sim tests with Node 10,11,12,13,14,15 https://github.com/ChainSafe/lodestar/runs/2138730025 and the results are
- 10: :heavy_multiplication_x: js-libp2p declares Node >= v12 in the package.json
- 11: :heavy_multiplication_x: js-libp2p declares Node >= v12 in the package.json
- 12: :heavy_check_mark: 
- 13: :heavy_check_mark: 
- 14: :heavy_check_mark: 
- 15: :heavy_check_mark: 

I've also run the Docker-ized setup with the `node:14-alpine` image and works fine.

If you can please test Lodestar in more environments with v14 and we can consider lifting the restriction

wemeetagain:
Woohoo! I guess the issues we were seeing have just vanished from dependencies being updated over time?

Definitely time to move to v14 once we can: https://nodejs.org/en/about/releases/
v12 no longer 'active', just in maintenance. v14 is the new active LTS.

PR #1371, opened 2020-08-08
Narrow node engine version

wemeetagain:
Temporary fix until #1370 is resolved

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "root",
   "private": true,
   "engines": {
-    "node": ">=12.0.0"
+    "node": ">=12.0.0 <12.13.0"
   },
   "workspaces": [
     "packages/*"
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "root",
   "private": true,
   "engines": {
-    "node": ">=12.0.0 <12.13.0"
+    "node": ">=12.0.0 <=12.14.1"
   },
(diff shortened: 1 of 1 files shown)


GregTheGreek (dismissed):

mpetrunic:
I'm running it with 12.14.1 for last 3-4 months so I don't think 12.13 is issue. in channel someone built it with 12.13 which you've excluded

wemeetagain:
@mpetrunic Oh, right. should we use <12.18.0?

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "root",
   "private": true,
   "engines": {
-    "node": ">=12.0.0"
+    "node": ">=12.0.0 <12.13.0"
```
wemeetagain:
```suggestion
    "node": ">=12.0.0 <12.18.0"
```

dapplion (dismissed):

mpetrunic:
> @mpetrunic Oh, right. should we use <12.18.0?

I would go <=12.14.1 since it's highest confirmed working version. You could try 12.17 if you wanna pinpoint version with issue :smile:

package.json
wemeetagain:
```suggestion
    "node": ">=12.0.0 <=12.14.1"
```

mpetrunic (approved):

PR #1378, opened 2020-08-09
Set strictNullChecks true in lodestar - part 2

dapplion:
Parallelizable effort to tackle #885. Sets strictNullChecks = true in a single package so eventually it can be set in the root tsconfig.

Note: There are 80 errors! (down from 271 on part 1) when compiling the current lodestar package with strictNullChecks = true. Extends the progress made on https://github.com/ChainSafe/lodestar/pull/1289 to prevent too big changes at once potentially creating conflicts with other PRs. The PR that manages to bring the error count to 0 will actually set `strictNullChecks = true`.

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -328,7 +328,6 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
     // add justified block to forkchoice so "this.justified" in forkchoice really map to a block
     if (isStateNotGenesis) {
       const preJustifiedBlock = await this.db.blockArchive.getByRoot(lastKnownState.currentJustifiedCheckpoint.root);
-      if (!preJustifiedBlock) throw Error("lastJustifiedBlock not in db");
       let preFinalizedBlocks = await this.db.blockArchive.values({
         gt: preJustifiedBlock.message.slot, lt: finalizedSlot});
       preFinalizedBlocks = sortBlocks([preJustifiedBlock, ...preFinalizedBlocks]);
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -63,7 +63,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
   public forkChoice: ILMDGHOST;
   public chainId: Uint16;
   public networkId: Uint64;
-  public clock?: IBeaconClock;
+  public clock: IBeaconClock;
   private readonly config: IBeaconConfig;
   private readonly db: IBeaconDb;
   private readonly eth1: IEth1Notifier;
@@ -113,8 +113,10 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
     return (await this.getHeadStateContext()).epochCtx;
   }
 
-  public async getHeadBlock(): Promise<SignedBeaconBlock|null> {
-    return this.db.block.get(this.forkChoice.headBlockRoot());
+  public async getHeadBlock(): Promise<SignedBeaconBlock> {
+    const headBlock = await this.db.block.get(this.forkChoice.headBlockRoot());
+    if (headBlock === null) throw Error("headBlock not in db");
+    return headBlock;
   }
 
   public async getBlockAtSlot(slot: Slot): Promise<SignedBeaconBlock|null> {
@@ -326,6 +328,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
     // add justified block to forkchoice so "this.justified" in forkchoice really map to a block
     if (isStateNotGenesis) {
       const preJustifiedBlock = await this.db.blockArchive.getByRoot(lastKnownState.currentJustifiedCheckpoint.root);
+      if (!preJustifiedBlock) throw Error("lastKnownState not in db");
       let preFinalizedBlocks = await this.db.blockArchive.values({
         gt: preJustifiedBlock.message.slot, lt: finalizedSlot});
       preFinalizedBlocks = sortBlocks([preJustifiedBlock, ...preFinalizedBlocks]);
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -328,7 +328,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
     // add justified block to forkchoice so "this.justified" in forkchoice really map to a block
     if (isStateNotGenesis) {
       const preJustifiedBlock = await this.db.blockArchive.getByRoot(lastKnownState.currentJustifiedCheckpoint.root);
-      if (!preJustifiedBlock) throw Error("lastKnownState not in db");
+      if (!preJustifiedBlock) throw Error("lastJustifiedBlock not in db");
       let preFinalizedBlocks = await this.db.blockArchive.values({
         gt: preJustifiedBlock.message.slot, lt: finalizedSlot});
       preFinalizedBlocks = sortBlocks([preJustifiedBlock, ...preFinalizedBlocks]);
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -85,7 +85,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter }) 
     this.eth1 = eth1;
     this.logger = logger;
     this.metrics = metrics;
-    this.forkChoice = forkChoice || new ArrayDagLMDGHOST(config);
+    this.forkChoice = forkChoice || new ArrayDagLMDGHOST(config) as ILMDGHOST;
     this.chainId = 0; // TODO make this real
     this.networkId = BigInt(0); // TODO make this real
     this.attestationProcessor = new AttestationProcessor(this, this.forkChoice, {config, db, logger});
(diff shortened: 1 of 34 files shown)


packages/lodestar/src/chain/chain.ts
```diff
@@ -326,6 +328,7 @@ export class BeaconChain extends (EventEmitter as { new(): ChainEventEmitter })
     // add justified block to forkchoice so "this.justified" in forkchoice really map to a block
     if (isStateNotGenesis) {
       const preJustifiedBlock = await this.db.blockArchive.getByRoot(lastKnownState.currentJustifiedCheckpoint.root);
+      if (!preJustifiedBlock) throw Error("lastKnownState not in db");
```
dapplion: Many error messages are raw, feel free to propose better more descriptive messages where necessary.

dapplion:
**CC Reviewers**: Only relevant changes are around nullish types where it will (a) throw an Error or (b) ignore subsequent logic. All the rest of logic is meant to be identical, so I'll appreciate help on making sure it didn't ignore any implementation details.

packages/lodestar/src/chain/chain.ts
mpetrunic: if you got this far in code, you already have lastKnwon state so message should be lastJustifiedBlock not in database

packages/lodestar/src/chain/factory/duties/index.ts
```diff
@@ -31,8 +31,8 @@ export function generateEmptyAttesterDuty(publicKey: BLSPubkey, duty?: Partial<A
   return {
     validatorPubkey: publicKey,
     aggregatorModulo: 1,
-    attestationSlot: null,
-    committeeIndex: null,
+    attestationSlot: -1,
+    committeeIndex: -1,
```
mpetrunic: this could be breaking change, not sure how ssz handles negative numbers and our VC don't handle -1 at all, rather put those fields as optional in type

packages/lodestar/src/network/gossip/utils.ts
```diff
@@ -56,9 +57,9 @@ export function isAttestationSubnetTopic(topic: string): boolean {
 }
 
 export function getSubnetFromAttestationSubnetTopic(topic: string): number {
-  assert.true(isAttestationSubnetTopic(topic), "Should be an attestation topic");
   const groups = topic.match(AttestationSubnetRegExp);
-  const subnetStr = groups[4];
+  const subnetStr = groups && groups[4];
+  if (subnetStr === null || subnetStr === undefined) throw Error(`Bad attestation topic format: ${topic}`);
```
mpetrunic:
```suggestion
  if (!subnetStr) throw Error(`Bad attestation topic format: ${topic}`);
```

packages/lodestar/src/network/gossip/utils.ts
```diff
@@ -38,7 +38,8 @@ export function mapGossipEvent(event: keyof IGossipEvents | string): GossipEvent
 
 export function topicToGossipEvent(topic: string): GossipEvent {
   const groups = topic.match(GossipTopicRegExp);
-  const topicName = groups[3] as keyof typeof GossipEvent;
+  const topicName = groups && groups[3];
+  if (topicName === null || topicName === undefined) throw Error(`Bad gossip topic format: ${topic}`);
```
mpetrunic:
```suggestion
  if (!topicName) throw Error(`Bad gossip topic format: ${topic}`);
```

packages/lodestar/src/sync/sync.ts
```diff
@@ -166,14 +167,16 @@ export class BeaconSync implements IBeaconSync {
...
-    let peer = peerBalancer.next();
-    let block: SignedBeaconBlock;
-    while (!block && peer) {
-      block = (await this.network.reqResp.beaconBlocksByRoot(peer, [root]))[0];
-      peer = peerBalancer.next();
-    }
-    if(block) {
-      await this.chain.receiveBlock(block);
+    const maxAttempts = 1000;
+    for (let attempt = 0; attempt < maxAttempts; attempt++) {
```
mpetrunic: why switch to for loop?

dapplion: @wemeetagain Commented he got stuck in an infinite loop sometimes in this function, so it might make sense to add some abort mechanism, i.e. a limited loop. No clue on what would be a reasonable maxAttempts value here, nor if it's the best approach to prevent the infinite loop possibility without addressing the underlying cause.

Maybe it would be best to not change the logic here, and tackle the infinite loop issue in another PR

mpetrunic: oh yeah, because someone changed peerBalalncer to loop around. I changed that in regular sync v2.
So we have two options, either reset RoundRobinArray to return null if all peers are discarded or reduce maxAttempts here to this.getPeers().length x 2 (so we try each peer twice).

if you have 10 peers and block doesn't exists this will surely make you banned from those 10 peers

dapplion:
@mpetrunic in `lodestar/packages/lodestar/src/chain/factory/duties/index.ts` `assembleAttesterDuty` was returning invalid `AttesterDuty` that contained null types if was `committeeAssignment` false.

https://github.com/ChainSafe/lodestar/blob/1d916a6e8de045575c6eda0786b58eed0e6d32ea/packages/lodestar/src/chain/factory/duties/index.ts#L13-L16

Is this a good approach? Asking because now a test is failing. Otherwise, should the test be fixed or removed?

```
  1) assemble validator duty
       should produce empty duty:
     Error: No committeeAssignment for validator 2 at epoch 3
      at Object.assembleAttesterDuty (src/chain/factory/duties/index.ts:1:1818)
      at Context.<anonymous> (test/unit/chain/factory/duties/assembleValidatorDuty.test.ts:43:33)
      at processImmediate (internal/timers.js:439:21)
      at process.topLevelDomainCallback (domain.js:130:23)
```

---

The tested action `should produce empty duty`, is something `assembleAttesterDuty` should be capable of doing? Is that needed somewhere?

dapplion:
Porting from issue https://github.com/ChainSafe/lodestar/issues/1402, @wemeetagain found an error while running the branch of this PR

---- 

Started node, stopped, restarted to this message:
```
2020-08-13 11:58:29  [CHAIN]            info: Found finalized state at slot 192, starting chain from there 
2020-08-13 11:58:29  [NODE]             info: Starting eth2 beacon node - LODESTAR! 
2020-08-13 11:58:29  [METRICS]          info: Starting metrics HTTP server on port 8008 
2020-08-13 11:58:29  [DB]               info: Connected to LevelDB database at .medalla/chain-db 
2020-08-13 11:58:37  [CHAIN]            info: Chain started, waiting blocks and attestations 
2020-08-13 11:58:37  [CHAIN]            info: Found last known finalized state at epoch #0 root 0x0000000000000000000000000000000000000000000000000000000000000000 
2020-08-13 11:58:38  [CHAIN]            info: Found 85 nonfinalized blocks in database from slot 192 to 316 
2020-08-13 11:58:38  [CHAIN]            info: Found finalized block at slot 192,
        root=0x5ce25fca09469f8473c31213a213b8ebbd76eebc41b47c19f984c5dabee6fe3a 
 ✖ Error: lastJustifiedBlock not in db
    at BeaconChain.initForkChoice (/home/cayman/Code/lodestar/packages/lodestar/lib/chain/chain.js:348:37)
    at async BeaconChain.restoreHeadState (/home/cayman/Code/lodestar/packages/lodestar/lib/chain/chain.js:321:5)
    at async BeaconChain.start (/home/cayman/Code/lodestar/packages/lodestar/lib/chain/chain.js:176:5)
    at async BeaconNode.start (/home/cayman/Code/lodestar/packages/lodestar/lib/node/nodejs.js:157:5)
    at async Object.run [as handler] (/home/cayman/Code/lodestar/packages/lodestar-cli/lib/cmds/beacon/run.js:96:3)
```

-----

@wemeetagain Do you think this error was triggered when it shouldn't? Or the error is correctly thrown and the underlying error should be addressed later?

wemeetagain:
I think it was triggered when it shouldn't. When I switched to the master branch, using the same db, the startup swept past that point w/o error.

dapplion:
@wemeetagain I've reviewed the PR for throw conditions

##### `chain/chain.ts`
Remove added throw on lastJustifiedBlock not in db. Now it only adds this throw condition 
https://github.com/ChainSafe/lodestar/blob/e50ce09f4a9348895d3f933d486a6092e18e8d76/packages/lodestar/src/chain/chain.ts#L116-L120
All code consuming `getHeadBlock()` immediately calls `headBlock.message` so without that throw it will blow up downstream for sure.

##### `forkChoice/arrayDag/lmdGhost.ts`
 I've reverted the strict NodeIndeces Map, there are now two main groups of changes:
- Normalize the assertion of node existence with the null type instead of using the getters `hasParent()`, `hasBestChild()`, etc.
- throw conditions in the class getters `headNode()`, `getJustified()`, `getFinalized()`. There is code downstream that will blow up if those return null, some that does not.

Should these throw conditions also be reverted?

dapplion:
:rofl: 

```
@chainsafe/lodestar:   1 failing
@chainsafe/lodestar:   1) ArrayDagLMDGHOST
@chainsafe/lodestar:        update justified checkpoint after finalized checkpoint is set
@chainsafe/lodestar:          should update justified checkpoint - new justified is conflict to the previous justified:
@chainsafe/lodestar:      TypeError: Cannot read property 'bestChild' of undefined
@chainsafe/lodestar:       at ArrayDagLMDGHOST.isBestChildOfParent (src/chain/forkChoice/arrayDag/lmdGhost.ts:63:498)
@chainsafe/lodestar:       at ArrayDagLMDGHOST.onRemoveWeight (src/chain/forkChoice/arrayDag/lmdGhost.ts:53:33)
@chainsafe/lodestar:       at ArrayDagLMDGHOST.propagateWeightChange (src/chain/forkChoice/arrayDag/lmdGhost.ts:46:599)
```

https://github.com/ChainSafe/lodestar/runs/986369593#step:9:1486

wemeetagain:
re: fork choice, I think you should work only on the type-level, try not to change any runtime behavior

wemeetagain:
closing in favor of #1420

Issue #1086, opened 2020-07-03
Change tsconfig target to es2019


mpetrunic:
Currently we are targeting esnext in every tsoconfig file in lodestar.
Esnext is not fully supported in nodejs 12 which means we could have runtime errors if we use such unsupported feature.

Solution:
- change target to es2018 (fully supported by node12)
- replace all 12n bigint literals with BigInt(12) - following regex does that automatically `\s+(\d+)n` => `BigInt($1)`

This will allow us to use all node12 goodies like optional chaining and it will warn us if we try to use some unsupported esnext stuff (we enabled only esnext.bigint lib as it's supported by node12)

mpetrunic:
I already did it in lodestar package to fix optional chaining runtime error but we should probably do it for all packages

dapplion:
I had issues in `lodestar-cli` since @wemeetagain added optional chaining there too. Should we change it at the root tsconfig?

mpetrunic:
YES :D but it will involve some changes in terms of 1n -> BigInt(1) since es2018 doesn't support 1n syntax

dapplion:
Shouldn't Babel took care of the transpilation of those features?

mpetrunic:
> Shouldn't Babel took care of the transpilation of those features?

Context: https://stackoverflow.com/questions/59787574/typescript-tsconfig-settings-for-node-js-12

 tldr nodejs supports some bigint features available in estarget beyond es2019 but laks some other, so if we set target to esnext, 
transpiled code could contain some features not supported by node12 but it's gonna fail only at runtime.

Babel is fine since it transpiles to code compatible with nodejs version while tsonfig (which we use for typechecking and some other stuff) targets es versions.

wemeetagain:
standardize on node v12.9+, 
```json
{
  "lib": "es2020",
  "target": "es2019"
}
```

PR #1387, opened 2020-08-11
target es2019 in lodestar-cli development tsconfig

dapplion:
Since @wemeetagain added optional chaining in the CLI the code can't be run with ts-node in development. Related to https://github.com/ChainSafe/lodestar/issues/1086


mpetrunic:
Node12 fully supports es2019 so we should target that instead

mpetrunic (approved):

wemeetagain (approved):

PR #1393, opened 2020-08-12
Allow all node v12 versions

wemeetagain:


twoeths (approved):

mpetrunic (approved):

dapplion:
What has changed to be confident that it works with all Node v12 versions?

wemeetagain:
We ran the code and didn't experience the described hanging in the beacon node

PR #1395, opened 2020-08-12
Set strictNullChecks true in lodestar-cli

dapplion:
Parallelizable effort to tackle #885. Sets strictNullChecks = true in a single package so eventually it can be set in the root tsconfig.

Since `lodestar-cli` depends on `lodestar` until all errors are resolved (see https://github.com/ChainSafe/lodestar/pull/1378) there the check can't be turned on here. This PR fixes the existing errors to ease future work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "@types/chai": "4.2.0",
     "@types/chai-as-promised": "^7.1.2",
     "@types/mocha": "^5.2.7",
-    "@types/node": "^12.12.21",
+    "@types/node": "^14.0.27",
     "@types/sinon": "^9.0.0",
     "@typescript-eslint/eslint-plugin": "2.23.0",
     "@typescript-eslint/parser": "2.23.0",
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "@types/chai": "4.2.0",
     "@types/chai-as-promised": "^7.1.2",
     "@types/mocha": "^5.2.7",
-    "@types/node": "^14.0.27",
+    "@types/node": "^12.12.21",
     "@types/sinon": "^9.0.0",
     "@typescript-eslint/eslint-plugin": "2.23.0",
     "@typescript-eslint/parser": "2.23.0",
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
     "@types/chai": "4.2.0",
     "@types/chai-as-promised": "^7.1.2",
     "@types/mocha": "^5.2.7",
-    "@types/node": "^12.12.21",
+    "@types/node": "^14.0.27",
```
mpetrunic: why? we aren't using node14

dapplion: I bumped because a dependency uses incompatible types that where not causing issues because strictNullChecks was false. A github issue about recommended to bump `@types/node`. I didn't knew the `@types/node` semver was related to NodeJS version.

dapplion: Since I'm not turning `strictNullChecks` on yet, I'll revert the version bump and we can deal with the dependency issue latter.

yarn.lock
```diff
@@ -2540,11 +2540,16 @@
   resolved "https://registry.yarnpkg.com/@types/node/-/node-10.17.20.tgz#e6d8b3631af1e59bbb4fda04926b078acdd3c2ef"
   integrity sha512-XgDgo6W10SeGEAM0k7FosJpvLCynOTYns4Xk3J5HGrA+UI/bKZ30PGMzOP5Lh2zs4259I71FSYLAtjnx3qhObw==
 
-"@types/node@^12.12.21", "@types/node@^12.6.1", "@types/node@^12.7.2":
+"@types/node@^12.6.1", "@types/node@^12.7.2":
   version "12.12.36"
   resolved "https://registry.yarnpkg.com/@types/node/-/node-12.12.36.tgz#162c8c2a2e659da480049df0e19ae128ad3a1527"
   integrity sha512-hmmypvyO/uTLFYCYu6Hlb3ydeJ11vXRxg8/WJ0E3wvwmPO0y47VqnfmXFVuWlysO0Zyj+je1Y33rQeuYkZ51GQ==
 
+"@types/node@^14.0.27":
```
wemeetagain: I think yarn lock should be fixed?

mpetrunic (approved):

wemeetagain (approved):

PR #1397, opened 2020-08-12
Connect to the correct chainId with ethers default provider

### LINKED ISSUE #1340: ChainId mismatch when rpcUrl not specified
**Describe the bug**

When making a deposit, if the rpcurl isn't specified an error is thrown stating that there is a chainId address mismatch.

**Expected behavior**

It should error that there are missing params.

**Steps to Reproduce**

 - OS:  ubuntu 18
 - Version:  latest master
 - Branch:  master
 - Commit hash: db3ad729923dc4021479c56624b02762e7b9fe7e

dapplion:
On CLI cmd `accounts validator deposit`, fetch deposit chain ID from params to initialize an ethers default provider

Fixes https://github.com/ChainSafe/lodestar/issues/1340.

### DIFF
--- a/packages/lodestar-cli/src/cmds/account/cmds/validator/deposit.ts
+++ b/packages/lodestar-cli/src/cmds/account/cmds/validator/deposit.ts
@@ -91,7 +91,10 @@ The deposit contract address will be determined by the spec config flag.",
     // eslint-disable-next-line no-console
     console.log(`Starting ${validatorDirsToSubmit.length} deposits`);
 
-    const eth1Signer = await getEthersSigner(options);
+    const eth1Signer = await getEthersSigner({
+      ...options,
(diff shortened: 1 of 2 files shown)


mpetrunic (approved):

mpetrunic:
Does that even work, I vaguely remember that it quickly hits infura rate limit

dapplion:
> Does that even work, I vaguely remember that it quickly hits infura rate limit

You use ethers own project ID https://github.com/ethers-io/ethers.js/blob/9640e864a68b4a9e84e820f0ceaf1eb56c66715f/packages/providers/src.ts/infura-provider.ts#L16
Since this command only does a single request to the node it shouldn't cause a rate limit. If for some reason rate limit happened, users can use the `rpcUrl` flag.

wemeetagain (approved):

Issue #1394, opened 2020-08-12
Bump bls-keygen and use eth2ValidatorPaths


dapplion:
After merging https://github.com/ChainSafe/bls-keygen/pull/21/files, bump the library in lodestar and use `eth2ValidatorPaths` in the CLI instead of hardcoding the derivation paths.

3xtr4t3rr3str14l:
closed via #1400

PR #1400, opened 2020-08-12
upgrade bls-keystore and bls-keygen

### LINKED ISSUE #1389: Upgrade bls-keystore
We're still using an old version of bls-keystore

Use the latest version. Fix the minor breaking changes (renaming of a few methods)

3xtr4t3rr3str14l:
fixes #1389 and  #1394 by upgrading bls-keygen and bls-keystore to 0.2.1 and 0.2.0 respectively and making the appropriate changes.
-change function names where needed as well as converting them to their new async definitions
-removed prefixing of public key (and consequently, the validator keystore directory name)
-corrected the generation of the wallet keystore master key to be 32 bytes instead of 64 bytes


packages/lodestar-cli/src/validatorDir/ValidatorDirManager.ts
```diff
@@ -67,8 +67,8 @@ export class ValidatorDirManager {
   /**
    * Opens all the validator directories in `this` and decrypts the validator keypairs.
    */
-  decryptAllValidators(options?: IValidatorDirOptions): Keypair[] {
+  async decryptAllValidators(options?: IValidatorDirOptions): Promise<Keypair[]> {
     const validators = this.openAllValidators(options);
-    return validators.map(validator => validator.votingKeypair(this.secretsDir));
+    return await Promise.all(validators.map(async validator => await validator.votingKeypair(this.secretsDir)));
```
wemeetagain: You can avoid using the innermost await, the Promise.all will take care of awaiting for you.

packages/lodestar-cli/src/wallet/Wallet.ts
```diff
@@ -97,10 +99,10 @@ export class Wallet extends Keystore {
       withdrawal: `m/12381/3600/${validatorIndex}/0`,
     };
 
-    const keystores = mapValues(privKeys, (privKey, key) => {
+    const keystores = mapValues(privKeys, async (privKey, key) => {
       const type = key as keyof typeof privKeys;
       const privateKey = PrivateKey.fromBytes(privKey);
-      const keystore = Keystore.encrypt(privKey, emptyPubkey, passwords[type], paths[type]);
+      const keystore = await Keystore.create(passwords[type], privKey, emptyPubkey, paths[type]);
```
wemeetagain: I think you can replace `emptyPubkey` with `privateKey.toPublicKey().toBytesCompressed()` here, then remove the below comment and `keystore.pubkey` line

packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
```diff
@@ -18,6 +19,15 @@ interface IWalletCreateArgs {
   mnemonicOutputPath?: string;
 }
 
+function validatePassphraseFile(passphraseFile: string): void {
+  if (path.parse(passphraseFile).ext !== ".pass") {
+    throw new YargsError("passphraseFile must end with .pass, make sure to not provide the actual password");
+  }
+  if (!fs.existsSync(passphraseFile)) {
```
mpetrunic: whats the logic for generating random passphrase file if there isn't any. Ain't this gonna fail with wrong password next time you use it?

packages/lodestar-cli/src/validatorDir/paths.ts
```diff
@@ -14,13 +13,13 @@ export const LOCK_FILE = ".lock";
 // Dynamic paths computed from the validator pubkey
 
 export function getValidatorDirPath(
-  {keystoresDir, pubkey, prefixed}: {keystoresDir: string; pubkey: string; prefixed?: boolean}
```
mpetrunic: Wasn't this causing errors because we don't have 0x prefix when using some command?

mpetrunic (commented):
I'm reluctant to approve this since we have 0 unit tests for cli util functions.

packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
```diff
@@ -18,6 +19,15 @@ interface IWalletCreateArgs {
   mnemonicOutputPath?: string;
 }
 
+function validatePassphraseFile(passphraseFile: string): void {
```
dapplion, line 22: Why is this utility move into the command handler?

packages/lodestar-cli/src/validatorDir/paths.ts
dapplion: @3xtr4t3rr3str14l  The rest of the code expects this path prefixed with `0x`. Is the current Keystore returning the pubkey prefixed?

packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
3xtr4t3rr3str14l: I didn't write this.  I just moved it because codecov was complaining bout the function where that logic used to be having too many lines.  @dapplion ?

packages/lodestar-cli/src/validatorDir/paths.ts
3xtr4t3rr3str14l: The new bls-keystore library returns the pubkey without a prefix, which is why I made this change (bear in mind that this was upgraded from 1.0.0-beta8 to 2.0.0 so a number of things have changed since we last upgraded that pkg here)

packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
wemeetagain: You should pull out the second `if` here into a `maybeCreatePassphraseFile`. Also call this function `validatePassphraseFilename`.

So then below, the flow will look like:
```
validatePassphraseFilename(passphraseFile);
maybeCreatePassphraseFile(passphraseFile);
```

packages/lodestar-cli/src/validatorDir/paths.ts
wemeetagain: It seems all the hex strings in the new keystore library are non-prefixed. And that is actually a good thing for cross-client compatability, other clients don't use 0x prefix in keystores, including eth2.0-deposit-cli.

In that case, we should keep this logic in this function so we can add prefixes asap in this part of the codebase.

wemeetagain:
This is a good change, but FYI it will break our existing wallets and keystores in several ways that we should be prepared for:

1. Our wallet master private keys are currently just wrong. We don't create a master private key for our wallet, but rather use the seed/entropy from mnemonic directly! The newer bls-keygen caught this error with now checking that the input is 32 bytes. (the seed is 64 bytes)

2. Our keystores previously used '0x'-prefixed pubkeys. Now we have strict json schema validation, which will fail for the previous implementation. Attempting to create a new keystore with '0x'-prefixed pubkeys will also fail.

packages/lodestar-cli/src/validatorDir/paths.ts
3xtr4t3rr3str14l: ok, just reverted paths.ts and format.ts

packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
wemeetagain: bump

3xtr4t3rr3str14l: gotchu @wemeetagain

packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
```diff
@@ -58,22 +71,21 @@ export const create: ICliCommand<IWalletCreateArgs, IAccountWalletArgs & IGlobal
...
     const mnemonic = bip39.generateMnemonic();
 
-    if (path.parse(passphraseFile).ext !== ".pass") {
-      throw new YargsError("passphraseFile must end with .pass, make sure to not provide the actual password");
-    }
-    if (!fs.existsSync(passphraseFile)) {
-      writeFile600Perm(passphraseFile, randomPassword());
-    }
+    validatePassphraseFilename(passphraseFile);
+    maybeCreatePassphraseFile(passphraseFile);
```
mpetrunic: @dapplion Why do we create random password here?

packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
```diff
@@ -18,6 +19,18 @@ interface IWalletCreateArgs {
   mnemonicOutputPath?: string;
 }
 
+function validatePassphraseFilename(passphraseFile: string): void {
+  if (path.parse(passphraseFile).ext !== ".pass") {
```
dapplion, line 23: @3xtr4t3rr3str14l If you are creating this functions just to please CodeClimate I don't think it's a good idea. In my personal opinion having to jump to another function for something this simple that is not re-used, it decreases code readability. But I would love to know others opinions.

packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
dapplion: I think it's a better approach than requesting user input. This keystore is likely to only exist locally for the validator client so what the actual password is, is irrelevant as long as it's strong. What would be another alternative?

packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
dapplion: @wemeetagain Do you think it's a good idea to create these very small utility functions that are not re-used just to please CodeClimate errors? In my personal opinion, it reduces code readability.

wemeetagain: Not really. I usually just CodeClimate to help identify _potential_ issues, not enforce the rules rigorously. Would be in favor of reverting this specific change.

wemeetagain: @3xtr4t3rr3str14l Could you revert these two small functions? Feel free to ignore CodeClimate in cases like these

wemeetagain (changes requested):

packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
3xtr4t3rr3str14l: ok cool, will do.  this is the first time i've ran into code climate issues on this repo, so i wasn't sure if it would prevent a merge or not.  will push that reversion here in a sec

3xtr4t3rr3str14l: pushed

mpetrunic (approved):

wemeetagain (approved):

PR #1401, opened 2020-08-13
Faster attestation validation and processing

mpetrunic:
Since this is quite extensive change, I would like some sanity review before I start refactoring tests for this new logic.

Some noticable improvements:
- introduce checkpoint state cache (a lot of attestations has same target checkpoint so we can reuse checkpoint states instead processing blocks all over again)
- introduce seen attestation lru cache (for "*first seen attestation for validator*" - prevent database lookups)
- reduce duplicate validation across board
- attestation processor has both pending block and slot pool where attestation wait until they are ready


packages/lodestar/src/network/gossip/validation/attestation.ts
```diff
@@ -0,0 +1,131 @@
...
+    );
+    return ExtendedValidatorResult.reject;
+  }
+  await db.seenAttestationCache.addCommitteeAttestation(attestation);
+  logger.profile("gossipAttestationValidation");
+  logger.info("Received valid committee attestation", attestationLogContext);
+  return ExtendedValidatorResult.accept;
+}
+
+export async function isAttestingToInValidBlock(db: IBeaconDb, attestation: Attestation): Promise<boolean> {
```
wemeetagain, line 111: Should we be checking for block in the block db? Eg: checking for included good block, rather than bad block.

wemeetagain (dismissed):
Generally, looks good :100:

packages/lodestar/src/network/gossip/validation/attestation.ts
mpetrunic: This is gossip validation so if we don't have valid block we would put it in pool and ignore so it makes sense to first check if it's invalid as we need to reject immediately

mpetrunic:
@ChainSafe/eth2-0  Should be ready now!

There is no more attestation validation errors in e2e tests so we reach finalization faster.

Unfortunately, I had to rollback changes to AggregateAndProof as they've caused e2e errors. I will tackle that in separate PR along with unit tests for it.

wemeetagain (approved):

PR #1404, opened 2020-08-14
Protect Validator API

### LINKED ISSUE #1380: Ensure VC knows of BN syncing status
**Describe the bug**
See https://discordapp.com/channels/593655374469660673/593655641445367808/742254419156009072

The validator never realizes that the beacon node is still syncing. Leads to confusing logs and beacon node errors.

**Expected behavior**
Validator API responses are tailored / conditional on syncing status
Validator is aware of syncing status

### LINKED ISSUE #1391: Validator can trigger errors in Beacon Node
**Describe the bug**
See https://discordapp.com/channels/593655374469660673/593655641445367808/742868789452079194

> So everytime I get one of these in the validator

> It freezes up the beacon chain for 5-10 mins with this:

> so i have shut down the validator just so the beacon can catch back up

**Expected behavior**
Validator cannot cause errors in the beacon node

twoeths:
resolves #1391 
resolves #1380 

For validator api:
+ Do a check before `processSlots()` to make sure our node is not suspended if epoch is too far in the future
+ throw Error with detailed sync status if our node is syncing

image

### DIFF
--- a/packages/lodestar/src/api/impl/validator/validator.ts
+++ b/packages/lodestar/src/api/impl/validator/validator.ts
@@ -43,6 +43,7 @@ import {assembleAttesterDuty} from "../../../chain/factory/duties";
 import {assembleAttestation} from "../../../chain/factory/attestation";
 import {IBeaconSync} from "../../../sync";
 import {toGraffitiBuffer} from "../../../util/graffiti";
+import {ApiError} from "../errors/api";
 import {processSlots} from "@chainsafe/lodestar-beacon-state-transition/lib/fast/slot";
 
 export class ValidatorApi implements IValidatorApi {
@@ -75,6 +76,7 @@ export class ValidatorApi implements IValidatorApi {
     randaoReveal: Bytes96,
     graffiti = ""
   ): Promise<BeaconBlock> {
+    await this.checkSyncStatus();
     const validatorIndex = (await this.chain.getHeadEpochContext()).pubkey2index.get(validatorPubkey);
     if (validatorIndex === undefined) {
       throw Error(`Validator pubKey ${toHexString(validatorPubkey)} not in epochCtx`);
@@ -96,6 +98,7 @@ export class ValidatorApi implements IValidatorApi {
     slot: Slot,
   ): Promise<Attestation> {
     try {
+      await this.checkSyncStatus();
       const [headBlockRoot, {state: headState, epochCtx}] = await Promise.all([
         this.chain.forkChoice.headBlockRoot(),
         this.chain.getHeadStateContext(),
@@ -123,6 +126,7 @@ export class ValidatorApi implements IValidatorApi {
   }
 
   public async publishBlock(signedBlock: SignedBeaconBlock): Promise<void> {
+    await this.checkSyncStatus();
     await Promise.all([
       this.chain.receiveBlock(signedBlock),
       this.network.gossip.publishBlock(signedBlock)
@@ -130,6 +134,7 @@ export class ValidatorApi implements IValidatorApi {
   }
 
   public async publishAttestation(attestation: Attestation): Promise<void> {
+    await this.checkSyncStatus();
     //it could discard attestations that would can be valid a bit later
     // await validateGossipAttestation(
     //   this.config, this.db, headStateContext.epochCtx, headStateContext.state, attestation
@@ -141,8 +146,9 @@ export class ValidatorApi implements IValidatorApi {
   }
 
   public async getProposerDuties(epoch: Epoch): Promise<ProposerDuty[]> {
-    const {state, epochCtx} = await this.chain.getHeadStateContext();
+    await this.checkSyncStatus();
     assert.gte(epoch, 0, "Epoch must be positive");
+    const {state, epochCtx} = await this.chain.getHeadStateContext();
     assert.lte(
       epoch,
       computeEpochAtSlot(this.config, state.slot) + 2,
@@ -162,6 +168,8 @@ export class ValidatorApi implements IValidatorApi {
   }
 
   public async getAttesterDuties(epoch: number, validatorPubKeys: BLSPubkey[]): Promise<AttesterDuty[]> {
+    await this.checkSyncStatus();
+    if (!validatorPubKeys || validatorPubKeys.length === 0) throw new Error("No validator to get attester duties");
     const {epochCtx, state} = await this.chain.getHeadStateContext();
     const currentSlot = getCurrentSlot(this.config, state.genesisTime);
     if(state.slot < currentSlot) {
@@ -191,6 +199,7 @@ export class ValidatorApi implements IValidatorApi {
   public async publishAggregateAndProof(
     signedAggregateAndProof: SignedAggregateAndProof,
   ): Promise<void> {
+    await this.checkSyncStatus();
     await Promise.all([
       this.db.aggregateAndProof.add(signedAggregateAndProof.message),
       this.network.gossip.publishAggregatedAttestation(signedAggregateAndProof)
@@ -204,6 +213,7 @@ export class ValidatorApi implements IValidatorApi {
   public async produceAggregateAndProof(
     attestationData: AttestationData, aggregator: BLSPubkey
   ): Promise<AggregateAndProof> {
+    await this.checkSyncStatus();
     const attestations = (await this.getWireAttestations(
       computeEpochAtSlot(this.config, attestationData.slot),
       attestationData.index
@@ -252,6 +262,7 @@ export class ValidatorApi implements IValidatorApi {
   public async subscribeCommitteeSubnet(
     slot: Slot, slotSignature: BLSSignature, committeeIndex: CommitteeIndex, aggregatorPubkey: BLSPubkey
   ): Promise<void> {
+    await this.checkSyncStatus();
     const state = await this.chain.getHeadState();
     const domain = getDomain(
       this.config,
@@ -273,6 +284,13 @@ export class ValidatorApi implements IValidatorApi {
     );
     const subnet = computeSubnetForSlot(this.config, state, slot, committeeIndex);
     await this.network.searchSubnetPeers(String(subnet));
+  }
+
+  private async checkSyncStatus(): Promise<void> {
+    if (!this.sync.isSynced()) {
+      const syncStatus = this.config.types.SyncingStatus.toJson(await this.sync.getSyncStatus());
+      throw new ApiError(503, `Node is syncing, status: ${JSON.stringify(syncStatus)}`);
+    }
   }
 
 }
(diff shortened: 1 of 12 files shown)


wemeetagain (dismissed):
Awesome. :+1:

Do you happen to know which http response code fastify sends back in the errored case?

twoeths:
> Do you happen to know which http response code fastify sends back in the errored case?

it's 500 as in `{"statusCode":500,"error":"Internal Server Error","message":"Node is syncing, status: {\"headSlot\":\"69568\",\"syncDistance\":\"26318\"}"}`

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -75,6 +75,10 @@ export class ValidatorApi implements IValidatorApi {
     randaoReveal: Bytes96,
     graffiti = ""
   ): Promise<BeaconBlock> {
+    if (!this.sync.isSynced()) {
+      const syncStatus = this.config.types.SyncingStatus.toJson(await this.sync.getSyncStatus());
+      throw new Error(`Node is syncing, status: ${JSON.stringify(syncStatus)}`);
```
mpetrunic: Can you throw ApiError that contains prop statusCode and throw it here with status code set to 503.
Rest api should handle this error and return proper status code

mpetrunic (changes requested):
You should add test case to rest api unit tests

packages/lodestar/src/api/impl/validator/validator.ts
twoeths: I switched to using our middlware, it used 503 there.
Since we use it in single place, I don't think we need ApiError?

twoeths:
> You should add test case to rest api unit tests

I added `should return 503 when node is syncing`

packages/lodestar/src/api/impl/validator/validator.ts
mpetrunic: I don't think we should use middleware as we might consume this api directly (not trough rest)

mpetrunic (changes requested):
We should deal with sync status bug before merging. We currently don't handle situation before head  block, when we are in waitingForPeers and when target < current slot

wemeetagain:
> We should deal with sync status bug before merging. We currently don't handle situation before head block, when we are in waitingForPeers and when target < current slot

This seems like it should be tackled as a separate PR.
We also may want the API to still work in that case, or at least when our head is at the current epoch.

mpetrunic:
> > We should deal with sync status bug before merging. We currently don't handle situation before head block, when we are in waitingForPeers and when target < current slot
> 
> This seems like it should be tackled as a separate PR.
> We also may want the API to still work in that case, or at least when our head is at the current epoch.

True but we cannot merge this before fixing this silly bug 😅

packages/lodestar/src/sync/gossip/handler.ts
```diff
@@ -49,7 +50,9 @@ export class BeaconGossipHandler implements IGossipHandler {
     this.subscribe(forkDigest);
   };
 
-  private subscribe = (forkDigest: ForkDigest): void => {
+  private subscribe = async (forkDigest: ForkDigest): Promise<void> => {
+    // TODO: remove this once we fix in gossipsub-js
```
twoeths: @wemeetagain I was not able to subscribe to `/beacon_block/` immediately after we start gossip (`/beacon_block/` was not even shown in `gossip subscriptions` log) so I have to delay 100ms like this in order to pass e2e.

packages/lodestar/src/api/impl/validator/util.ts
```diff
@@ -0,0 +1,22 @@
...
+import {assert} from "@chainsafe/lodestar-utils";
+import {computeEpochAtSlot, computeStartSlotAtEpoch} from "@chainsafe/lodestar-beacon-state-transition";
+import {processSlots} from "@chainsafe/lodestar-beacon-state-transition/lib/fast/slot";
+
+export async function safeProcessSlots
+(config: IBeaconConfig, epoch: Epoch, chain: IBeaconChain): Promise<ITreeStateContext> {
+  const {epochCtx, state} = await chain.getHeadStateContext();
+  assert.lte(
+    epoch,
+    computeEpochAtSlot(config, state.slot) + 2,
```
mpetrunic: I don't think we need this anymore because if we are in sync, this would stall us

packages/lodestar/test/e2e/noEth1Sim.test.ts
```diff
@@ -68,21 +72,29 @@ describe("no eth1 sim (multi-node test)", function () {
         }
       }
 
+      // wait for all nodes to be synced first or Validator API would failed
```
mpetrunic: Isn't it the goal for api to fail until we are synced?

mpetrunic: Maybe we can just put warning in VC instead of error in case of syncing

packages/lodestar/test/unit/sync/gossip/handler.test.ts
```diff
@@ -42,6 +43,7 @@ describe("gossip handler", function () {
     });
     const handler = new BeaconGossipHandler(chainStub, networkStub, dbStub, logger);
     await handler.start();
+    await sleep(100);
```
mpetrunic: We need to investigate this, putting random sleep in unit tests cannot be solution. That worked before?!

twoeths: it seems there is an issue in gossip or the way we handle gossip so I have to sleep(100) for unit tests in order to confirm the e2e tests, I'm investigating it.

packages/lodestar/src/network/gossip/gossip.ts
```diff
@@ -147,6 +147,7 @@ export class Gossip extends (EventEmitter as { new(): GossipEventEmitter }) impl
     event: keyof IGossipEvents | string,
     listener?: unknown,
     params: Map<string, string> = new Map()): void {
+    if (!this.listeners(event as keyof IGossipEvents).includes(listener as (...args: unknown[]) => void)) return;
```
twoeths, line 150: @wemeetagain so gossip is fine, this caused the incorrect unsubscription of gossip topic

packages/lodestar/test/e2e/noEth1Sim.test.ts
twoeths: I removed this, validator is handling pregenesis nicely with logs like `Waiting for genesis time waitTime=26s`

packages/lodestar/test/unit/sync/gossip/handler.test.ts
twoeths: resolved the issue

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -189,19 +198,22 @@ export class ValidatorApi implements IValidatorApi {
...
   ): Promise<void> {
+    await this.checkSyncStatus();
     await Promise.all([
       this.db.aggregateAndProof.add(signedAggregateAndProof.message),
       this.network.gossip.publishAggregatedAttestation(signedAggregateAndProof)
     ]);
   }
 
   public async getWireAttestations(epoch: number, committeeIndex: number): Promise<Attestation[]> {
+    await this.checkSyncStatus();
```
mpetrunic: We don't need it here?

wemeetagain:
file conflicts

mpetrunic (approved):
How does logs look like on validator side?

twoeths:
@mpetrunic 

> How does logs look like on validator side?

there is no error, it's like
```
2020-08-18 15:13:24  []                 info: Checking genesis time and beacon node connection
2020-08-18 15:13:24  []                 info: Waiting for genesis time waitTime=2s
2020-08-18 15:13:24  []                 info: Checking genesis time and beacon node connection
2020-08-18 15:13:24  []                 info: Waiting for genesis time waitTime=2s
2020-08-18 15:13:27  []                 info: Checking genesis time and beacon node connection
2020-08-18 15:13:27  [VALIDATOR 0-8]    info: Chain start has occured!
2020-08-18 15:13:27  [VALIDATOR 0-8]    info: on new block epoch epoch=0, validator=0xa99a76ed7796f7be22d5b7e85deeb7c5677e88e511e0b337618f8c4eb61349b4bf2d153f649f7b53359fe8b94a38e44c
```

mpetrunic:
> @mpetrunic
> 
> > How does logs look like on validator side?
> 
> there is no error, it's like
> 
> ```
> 2020-08-18 15:13:24  []                 info: Checking genesis time and beacon node connection
> 2020-08-18 15:13:24  []                 info: Waiting for genesis time waitTime=2s
> 2020-08-18 15:13:24  []                 info: Checking genesis time and beacon node connection
> 2020-08-18 15:13:24  []                 info: Waiting for genesis time waitTime=2s
> 2020-08-18 15:13:27  []                 info: Checking genesis time and beacon node connection
> 2020-08-18 15:13:27  [VALIDATOR 0-8]    info: Chain start has occured!
> 2020-08-18 15:13:27  [VALIDATOR 0-8]    info: on new block epoch epoch=0, validator=0xa99a76ed7796f7be22d5b7e85deeb7c5677e88e511e0b337618f8c4eb61349b4bf2d153f649f7b53359fe8b94a38e44c
> ```

Maybe we should have logs so user knows something is wrong?

twoeths:
> > @mpetrunic
> > > How does logs look like on validator side?
> > 
> > 
> > there is no error, it's like
> > ```
> > 2020-08-18 15:13:24  []                 info: Checking genesis time and beacon node connection
> > 2020-08-18 15:13:24  []                 info: Waiting for genesis time waitTime=2s
> > 2020-08-18 15:13:24  []                 info: Checking genesis time and beacon node connection
> > 2020-08-18 15:13:24  []                 info: Waiting for genesis time waitTime=2s
> > 2020-08-18 15:13:27  []                 info: Checking genesis time and beacon node connection
> > 2020-08-18 15:13:27  [VALIDATOR 0-8]    info: Chain start has occured!
> > 2020-08-18 15:13:27  [VALIDATOR 0-8]    info: on new block epoch epoch=0, validator=0xa99a76ed7796f7be22d5b7e85deeb7c5677e88e511e0b337618f8c4eb61349b4bf2d153f649f7b53359fe8b94a38e44c
> > ```
> 
> Maybe we should have logs so user knows something is wrong?

Last time I did see a lot of errors thrown at validator side because nodes run regular sync. With the latest code it does not run regular sync and nodes are up to date from the beginning (since I added genesisDelay in multinode test) so we don't see any errors anymore.

wemeetagain (approved):
Lets merge this, tackle additional logging in a separate PR

PR #1405, opened 2020-08-14
Update discv5 + Fix maxPeers

wemeetagain:


mpetrunic (changes requested):
I think you should fix maxPeer option to be passed to conn manager and set default value to 50 otherwise we will probably cause some issues with too many peers (from lighthouse discord, even they are having issues with validating 200 peers) :smile:

dapplion (dismissed):

mpetrunic (approved):

PR #1409, opened 2020-08-15
Optimize post block processing

mpetrunic:
Changelog:
- fix "message of undefined" errors in getSupportedPeerProtocol method
- hide scary errors and stacktraces when probing peers - removed stacktrace, message level changed to warn
- rather huge performance boost in method that removes old data from db
   - tldr we were trying to delete 20k deposits from db every time even though they don't exists. This removed 20k key encoding calls as well as level db searches for those keys (@wemeetagain  seems like that was gray blocks in profiler since I don't see them anymore)


wemeetagain (approved):
such a small change for a great effect :joy:

PR #1411, opened 2020-08-17
Fix sync status: handle waitForPeers and target < our head

twoeths:
Precondition for #1404 

https://github.com/ChainSafe/lodestar/pull/1404#pullrequestreview-467609880

### DIFF
--- a/packages/lodestar/src/sync/sync.ts
+++ b/packages/lodestar/src/sync/sync.ts
@@ -90,22 +90,32 @@ export class BeaconSync implements IBeaconSync {
   }
 
   public async getSyncStatus(): Promise<SyncingStatus> {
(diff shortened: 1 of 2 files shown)


mpetrunic (approved):

wemeetagain (approved):
LGTM.

The sync status endpoint is not what I'd expected, and the eth2.0-apis is scant with details.
Some questions I had:
- should the `headSlot` just be the clock slot? It means something slightly different, but I'm not sure the intention of the endpoint.
- `headSlot` is a confusing name. In all cases (except for this, apparently), the references to "head" are local, always the head _relative to our own state_. Or in the case of req/resp, our peers' view of the head. (From the perspective of the sender, its still "relative to their own state"). Here it seems `headSlot` means something like "the furthest `headSlot` of any of our peers".

PR #1417, opened 2020-08-18
Regular Sync: Simpler getBestPeer and handle empty range

twoeths:
Current:
+ get genesisTime from headState
+ getBestPeer with headSlot >= current_slot() - 1
+ keep retrying empty range

Change:
+ get genesisTime from chain
+ getBestPeer  headSlot > our headSlot
+ based on the best peer's head we know when to get next range, when to retry


packages/lodestar/src/chain/attestation/processor.ts
```diff
@@ -64,7 +64,7 @@ export async function processAttestation(
         balances[i]
       );
     }
-    logger.verbose(`Attestation ${toHexString(attestationHash)} passed to fork choice`);
+    // logger.verbose(`Attestation ${toHexString(attestationHash)} passed to fork choice`);
```
twoeths: this log is quite annoying bc it appears too much, not sure if we want to keep it

twoeths:
With this change we aren't stuck at this range anymore `Regular Sync: Requesting blocks from slot 73488 to slot 73507`

packages/lodestar/src/chain/attestation/processor.ts
mpetrunic: Maybe put it to debug? Generally it's somewhat usefull to see if attestations are managing to get into fork choice

packages/lodestar/src/sync/regular/naive/naive.ts
```diff
@@ -180,24 +191,28 @@ export class NaiveRegularSync extends (EventEmitter as { new(): RegularSyncEvent
...
     while (!this.bestPeer && !isAborted) {
       // wait first to make sure we have latest status
       await sleep(waitingTime);
-      const previousSlot = getCurrentSlot(this.config, state.genesisTime) - 1;
       const peers = this.network.getPeers();
-      const {slot: maxHeadSlot} = getBestHead(peers, this.reps);
-      if (maxHeadSlot >= previousSlot) {
-        this.bestPeer = getBestPeer(this.config, peers, this.reps);
-        this.logger.verbose(`Regular Sync: Found best peer ${this.bestPeer.toB58String()}, headSlot=${maxHeadSlot}`);
+      this.bestPeer = getBestPeer(this.config, peers, this.reps);
```
mpetrunic, line 203: We cannot sync 20k slots from one peer, we need to change logic to find best head and get all peers on same head.

mpetrunic (dismissed):

packages/lodestar/src/sync/regular/naive/naive.ts
twoeths: this is naive regular sync anyway, I prefer to fix to let it work first. Does v2 implement this?

even through we can get multiple peers with same head, it's not a guarantee we can get blocks from same chain since they can have different heads later, that's the concern we have in #1234 and we decided to get blocks from single best peer for this "naive" approach.

mpetrunic: @wemeetagain any ideas?

wemeetagain: I don't think syncing from a single peer would be so bad, if we could recover from that peer disconnecting us / timing out.
It seems we currently only do peer selection once, upfront.
I think the bigger thing we should add is the notion of regular sync failure, which should then recover from.
Strawman proposal: Regular sync knows when sync peers are no longer viable (eg: using heuristic N timeouts or disconnection), recalculates target + sync peers.

twoeths: with the latest update, whenever we get next range we call `getSyncPeers()` which then call `checkBestPeer()`. We'll find new best peer if
+ peer is not connected to node anymore
+ or peer head is not better than us

wemeetagain (approved):

PR #1418, opened 2020-08-18
Sync uses forkchoice latest() instead of head()

twoeths:
this is a copy of #1288

+ the current situation with Medalla regular sync is we started with head block, there are more blocks after that in our db. Then we fetch blocks from the best peer and chain processor found we had those blocks already and it does not emit "processedBlock" and regular sync gets stuck.
+ so we should start with `forkchoice.latest()` instead of `forkchoice.head()`


twoeths:
we may not need this PR if we decide to still emit "processedBlock" when processing an existing block.

mpetrunic:
> we may not need this PR if we decide to still emit "processedBlock" when processing an existing block.

we should emit processedBlock every time we process block new or existing. We will need to add "head" event that will just be emitted on new head for new event stream api

twoeths:
closing this, "processedBLock" is handled in #1425

PR #1420, opened 2020-08-19
Set strictNullChecks = true in root

### LINKED ISSUE #885: Enable strict null checking

dapplion:
Sets `strictNullChecks = true` in root tsconfig.json.

For all sub-packages except `lodestar` work was done beforehand with preceding PRs. This one required to fix a large number of errors in `lodestar` and took the approach of not modifying any logic when possible while doing extensive use of the non-null-assertion operator. This is a bad practice that should not be followed in new code, but it was the only way to enable this rule fast. 

Supersedes https://github.com/ChainSafe/lodestar/pull/1378
Fixes https://github.com/ChainSafe/lodestar/issues/885


dapplion:
@wemeetagain Only changes in logic are
https://github.com/ChainSafe/lodestar/blob/124ca24f0968688746a29c25c5ac3ff92658e844/packages/lodestar/src/network/gossip/gossipsub.ts#L77-L78
https://github.com/ChainSafe/lodestar/blob/124ca24f0968688746a29c25c5ac3ff92658e844/packages/lodestar/src/network/gossip/utils.ts#L43-L45
All other issues including fork choice and sync are fixed with the non-null-assertion operator

wemeetagain:
There's a test error
https://github.com/ChainSafe/lodestar/pull/1420/checks?check_run_id=1003809261#step:8:13

dapplion:
> There's a test error
> https://github.com/ChainSafe/lodestar/pull/1420/checks?check_run_id=1003809261#step:8:13

Fixed with https://github.com/ChainSafe/lodestar/pull/1420/commits/124ca24f0968688746a29c25c5ac3ff92658e844

wemeetagain:
file confiict

mpetrunic (dismissed):

wemeetagain:
damn, one last file conflict

wemeetagain (approved):

PR #1425, opened 2020-08-20
Initial sync to start from the finalized block

### LINKED ISSUE #1423: Initial Sync keeps retrying a range on restart
**Describe the bug**
I keep getting this in my node, can't move forward.
```
2020-08-20 05:58:51  [SYNC]             info: Imported blocks for slots: 74438,74439,74441,74442,74443,74444,74445,74446,74447,74449,74451,74454,74455,74456,74458,74459,74461,74462,74464,74466,74471,74473,74474,74475,74476,74478,74479,74481,74485,74490,74491,74493,74495,74496,74498,74499,74501
2020-08-20 05:58:51  [SYNC]             warn: Received block parent root doesn't match our head head=0xe1aeb5e8b60a65c5013716c19c4c26ad93b42a00c7cebb4bef903f11208df955, headSlot=74437, blockParent=0xcd03f1aafa136f6284052e37739f92adafd0b5883c1745642e31df446fc6fc45, blockSlot=74438
2020-08-20 05:58:51  [SYNC]          verbose: last processed slot=74437
2020-08-20 05:58:51  [SYNC]             warn: Failed to process range, retry fetching range, lastSlot=74437
```
This is because I have block 74437 in db which is not part of finalized chain, it came from regular sync 2 days ago and now it becomes my head while the real parent of block 74438 is 74435.
Can't recover from that atm.
**Expected behavior**
Initial Sync currently only expects finalized blocks in db which is good for most of the time.
But for the 1st range it fetches, it should expect some unfinalized blocks in db since it may come from an old regular sync some days ago.

**Steps to Reproduce**

1. Sync Medalla until Regular Sync so we have some unfinalized block in db and it becomes head
2. Stop
3. Wait for some time so Medalla can have new finalized epoch (in my case it's 2 days)
4. Start our node again to sync, it may be stuck if our head is not part of finalized chain.

twoeths:
resolves #1423 

In case our head does not stay on the finalized chain, initial sync can't proceed.
This PR helps recover from that DB by starting from the finalized block, can proceed to slot > 100k then node is crashed due to out of memory issue.

### DIFF
--- a/packages/lodestar/src/sync/initial/fast/index.ts
+++ b/packages/lodestar/src/sync/initial/fast/index.ts
@@ -18,6 +18,7 @@ import {ISlotRange} from "../../interface";
 import {fetchBlockChunks, getCommonFinalizedCheckpoint, processSyncBlocks} from "../../utils";
 import {GENESIS_EPOCH, Method} from "../../../constants";
 import {ISyncStats, SyncStats} from "../../stats";
+import {IBeaconDb} from "../../../db";
 
 export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}) implements InitialSync {
   private readonly opts: ISyncOptions;
@@ -27,6 +28,7 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
   private readonly network: INetwork;
   private readonly logger: ILogger;
   private readonly stats: ISyncStats;
+  private readonly db: IBeaconDb;
 
   /**
    * Targeted finalized checkpoint. Initial sync should only sync up to that point.
@@ -41,10 +43,14 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
    * Trigger for block import
    */
   private syncTriggerSource: Pushable<ISlotRange>;
+  /**
+   * The last processed block
+   */
+  private lastProcessedBlock: SignedBeaconBlock;
 
   public constructor(
     opts: ISyncOptions,
-    {config, chain, network, reputationStore, logger, stats}: IInitialSyncModules
+    {config, chain, network, reputationStore, logger, db, stats}: IInitialSyncModules
   ) {
     super();
     this.config = config;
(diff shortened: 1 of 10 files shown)


packages/lodestar/src/chain/interface.ts
```diff
@@ -24,7 +24,7 @@ import {IService} from "../node";
 
 export interface IChainEvents {
   unknownBlockRoot: (root: Root) => void;
-  processedBlock: (signedBlock: SignedBeaconBlock) => void;
+  processedBlock: (signedBlock: SignedBeaconBlock, isNew: boolean) => void;
```
mpetrunic: what does isNew means?

packages/lodestar/src/sync/initial/fast/index.ts
```diff
@@ -62,20 +69,25 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
     this.chain.on("processedCheckpoint", this.checkSyncCompleted);
     this.chain.on("processedBlock", this.checkSyncProgress);
     this.syncTriggerSource = pushable<ISlotRange>();
-    this.blockImportTarget = this.chain.forkChoice.headBlockSlot();
+    const headSlot = this.chain.forkChoice.headBlockSlot();
```
mpetrunic: why not fetch finalized slot block from db on start? It would ignore all unfinalized blocks

packages/lodestar/src/chain/interface.ts
mpetrunic: I don't think we need it

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -184,14 +181,45 @@ export function processSyncBlocks(
...
   };
 }
 
+/**
+ * Order blocks and pass to chain.
+ * @param chain BeaconChain
+ * @param logger the logger
+ * @param trusted trusted
+ */
+export function processRegularSyncBlocks(
```
mpetrunic: this and inital sync is nearly identical, can we deduplicate?

mpetrunic (approved):

wemeetagain (approved):

PR #1428, opened 2020-08-21
Diversify peers by subnet

### LINKED ISSUE #1383: Find diverse peers

We should aim to have at least one peer from each subnet.

Use discv5 to periodically search for peers with attestation subnets we are missing

**Additional context**

twoeths:
resolves #1383

New DiversifyPeersBySubnetTask task to run periodically to search for new peers if we miss any subnets.

## Test
Able to get all subnets after 1-2 rounds.
Screen Shot 2020-08-21 at 16 52 39

### DIFF
--- a/packages/lodestar/src/network/network.ts
+++ b/packages/lodestar/src/network/network.ts
@@ -18,6 +18,8 @@ import {IBeaconChain} from "../chain";
 import {MetadataController} from "./metadata";
 import {Discv5, Discv5Discovery, ENR} from "@chainsafe/discv5";
 import {IReputationStore} from "../sync/IReputation";
+import {DiversifyPeersBySubnetTask} from "./tasks/diversifyPeersBySubnetTask";
+import {CheckPeerAliveTask} from "./tasks/checkPeerAliveTask";
 
 interface ILibp2pModules {
   config: IBeaconConfig;
@@ -41,6 +43,8 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
   private logger: ILogger;
   private metrics: IBeaconMetrics;
   private peerReputations: IReputationStore;
+  private diversifyPeersTask: DiversifyPeersBySubnetTask;
+  private checkPeerAliveTask: CheckPeerAliveTask;
 
   public constructor(
     opts: INetworkOptions,
@@ -58,6 +62,16 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
     this.reqResp = new ReqResp(opts, {config, libp2p, peerReputations: this.peerReputations, logger});
     this.metadata = new MetadataController({}, {config, chain, logger});
     this.gossip = (new Gossip(opts, {config, libp2p, logger, validator, chain}) as unknown) as IGossip;
+    this.diversifyPeersTask = new DiversifyPeersBySubnetTask(this.config, {
+      network: this,
+      reps: this.peerReputations,
+      logger: this.logger,
+    });
+    this.checkPeerAliveTask = new CheckPeerAliveTask(this.config, {
+      network: this,
+      reps: this.peerReputations,
+      logger: this.logger,
+    });
   }
 
   public async start(): Promise<void> {
@@ -79,6 +93,11 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
     await this.gossip.stop();
     await this.reqResp.stop();
     await this.libp2p.stop();
+    await Promise.all([this.diversifyPeersTask.stop(), this.checkPeerAliveTask.stop()]);
+  }
+
+  public async handleSyncCompleted(): Promise<void> {
+    await Promise.all([this.diversifyPeersTask.start(), this.checkPeerAliveTask.start()]);
   }
 
   public getEnr(): ENR | undefined {
--- a/packages/lodestar/src/network/network.ts
+++ b/packages/lodestar/src/network/network.ts
@@ -107,6 +107,10 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
     return peers || [];
   }
 
+  public getMaxPeer(): number {
+    return this.opts.maxPeers;
+  }
+
   public hasPeer(peerId: PeerId, connected = false): boolean {
     const peer = this.libp2p.peerStore.get(peerId);
     if (!peer) {
@@ -142,42 +146,39 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
 
   public async searchSubnetPeers(subnet: string): Promise<void> {
     const peerIds = this.peerReputations.getPeerIdsBySubnet(subnet);
-    if (peerIds.length < 3) {
-      // If an insufficient number of current peers are subscribed to the topic,
+    if (peerIds.length === 0) {
       // the validator must discover new peers on this topic
-      this.logger.verbose(`Found only ${peerIds.length} for subnett ${subnet}, finding new peers to connect`);
-      const count = await this.connectToNewPeersBySubnet(parseInt(subnet), peerIds);
-      this.logger.verbose(`Connected to ${count} new peers for subnet ${subnet}`);
+      this.logger.verbose(`Finding new peers for subnet ${subnet}`);
+      const found = await this.connectToNewPeersBySubnet(parseInt(subnet));
+      if (found) {
+        this.logger.verbose(`Found new peer for subnet ${subnet}`);
+      } else {
+        this.logger.verbose(`Not found any peers for subnet ${subnet}`);
+      }
     }
   }
 
   /**
-   * Connect to new peers given a subnet.
+   * Connect to 1 new peer given a subnet.
    * @param subnet the subnet calculated from committee index
-   * @param inPeerIds peers already have this subnet
    */
-  private async connectToNewPeersBySubnet(subnet: number, inPeerIds: string[] = []): Promise<number> {
+  private async connectToNewPeersBySubnet(subnet: number): Promise<boolean> {
     const discv5Peers = (await this.searchDiscv5Peers(subnet)) || [];
-    const peerIds = discv5Peers.filter((peerId) => !inPeerIds.includes(peerId.toB58String()));
-    // make sure they still connect to same subnet
-    let count = 0;
+    const knownPeers = Array.from(await this.libp2p.peerStore.peers.values()).map((peer) => peer.id.toB58String());
+    const peerIds = discv5Peers.filter((peerId) => !knownPeers.includes(peerId.toB58String()));
+    let found = false;
     for (const peerId of peerIds) {
-      // we'll dial thru sendRequest so don't need to do connect() like in the spec
+      // will automatically get metadata once we connect
       try {
-        const metadata = await this.reqResp.metadata(peerId);
-        if (metadata === null) throw Error("No metadata");
-        if (metadata.attnets[subnet]) {
-          count++;
-        }
-      } catch (err) {
-        this.logger.warn(`Cannot get metadata from ${peerId.toB58String()}: ${err.message}`);
-      }
-      if (count < 10) {
-        // TODO: decide max peers per subnet to connect?
+        await this.connect(peerId);
+        found = true;
         break;
+      } catch (e) {
+        // this runs too frequently so make it verbose
+        this.logger.verbose(`Cannot connect to peer ${peerId.toB58String()} for subnet ${subnet}`, e.message);
       }
     }
-    return count;
+    return found;
   }
 
   private searchDiscv5Peers = async (subnet: number): Promise<PeerId[]> => {
--- a/packages/lodestar/src/network/network.ts
+++ b/packages/lodestar/src/network/network.ts
@@ -165,23 +165,23 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
   private async connectToNewPeersBySubnet(subnet: number): Promise<boolean> {
     const discv5Peers = (await this.searchDiscv5Peers(subnet)) || [];
     const knownPeers = Array.from(await this.libp2p.peerStore.peers.values()).map((peer) => peer.id.toB58String());
-    const peerIds = discv5Peers.filter((peerId) => !knownPeers.includes(peerId.toB58String()));
+    const candidatePeers = discv5Peers.filter((peer) => !knownPeers.includes(peer.peerId.toB58String()));
     let found = false;
-    for (const peerId of peerIds) {
+    for (const peer of candidatePeers) {
       // will automatically get metadata once we connect
       try {
-        await this.connect(peerId);
+        await this.connect(peer.peerId, [peer.multiaddr]);
         found = true;
         break;
       } catch (e) {
         // this runs too frequently so make it verbose
-        this.logger.verbose(`Cannot connect to peer ${peerId.toB58String()} for subnet ${subnet}`, e.message);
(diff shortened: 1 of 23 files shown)


packages/lodestar/src/network/util.ts
```diff
@@ -135,7 +135,7 @@ export async function dialProtocol(
     // eslint-disable-next-line no-useless-catch
   } catch (e) {
     const err = new Error(e.code || e.message);
-    err.stack = e.stack;
+    // err.stack = e.stack;
```
twoeths, line 151: It's too noisy if I enable this

packages/lodestar/src/network/network.ts
```diff
@@ -118,10 +119,10 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
 
   public async searchSubnetPeers(subnet: string): Promise<void> {
     const peerIds = this.peerReputations.getPeerIdsBySubnet(subnet);
-    if (peerIds.length < 3) {
+    if (peerIds.length < MIN_NUM_PEERS_PER_SUBNET) {
       // If an insufficient number of current peers are subscribed to the topic,
       // the validator must discover new peers on this topic
-      this.logger.verbose(`Found only ${peerIds.length} for subnett ${subnet}, finding new peers to connect`);
+      this.logger.verbose(`Found only ${peerIds.length} for subnet ${subnet}, finding new peers to connect`);
```
mpetrunic:
```suggestion
      this.logger.verbose(`Found only ${peerIds.length} peers for subnet ${subnet}, finding new peers to connect`);
```

packages/lodestar/src/network/network.ts
```diff
@@ -145,11 +146,11 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
         if (metadata.attnets[subnet]) {
           count++;
         }
+        this.peerReputations.getFromPeerId(peerId).latestMetadata = metadata;
```
mpetrunic: not related to this line, but we could also search not connected peers in peerstore (we might have dc-ed them before but now we need them again)

mpetrunic: That said, can we even ask for metadata before exchanging status (maybe we will get bad score because of doing this)?

mpetrunic:
@tuyennhv I was thinking, this might be too invasive (too much cpu usage for little benefit).
Since this is run periodically, how about on each loop:
- group peers per subnet
   - we need some ranking system since you cannot just drop one random peer from subnet group with a lot of peers
     since that peer could also be in other subnet groups where he is only one
- if we are close to max peers (less than 10% free?!), prune connected peers (disconnect ones that doesn't increase peer diversity or have bad score)
- search discv5 for missing peer subnet and initiate connection (don't wait for connection and metadata - it will be pruned in next loop if useless)

Do we have some loop that updates metadata?

packages/lodestar/src/network/network.ts
twoeths: I think status should be the 1st one because some peers may run different forks/networks

twoeths:
@mpetrunic it sounds good to choose to select peers to disconnect if it's close to maxConnection.
> I was thinking, this might be too invasive (too much cpu usage for little benefit).

we run per `EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION` which is per > 2 hours, any recommendation to increase it like `2*EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION`?
tbn, it's not a cpu intensive task for me because it's i/o operations in the end and all of the operations are in-memory.

> Do we have some loop that updates metadata?
no, we just implement `metadata` api and noone calls that

mpetrunic:
> we run per `EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION` which is per > 2 hours, any recommendation to increase it like `2*EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION`?
> tbn, it's not a cpu intensive task for me because it's i/o operations in the end and all of the operations are in-memory.

I meant  we are waiting for all those peer responses which might take long and have response timeouts.

I think we should run it way more often like every epoch or every n slots (we can exit if we already have diverse peers) but it should be quick (without waiting for connection response). My idea was that it should just try to connect fresh peers and on next run we would clean bad ones and try again.
 
> > Do we have some loop that updates metadata?
> > no, we just implement `metadata` api and noone calls that

We should probably have separate service that updates that on some interval. We should also have service that sends ping request for checking peer liveness.

packages/lodestar/src/sync/utils/reputation.ts
```diff
@@ -0,0 +1,122 @@
...
+  if (peers.length < maxPeer * retainRatio) return [];
+  const numDisconnect = peers.length + newPeer - maxPeer;
+  if (numDisconnect <= 0) return [];
+  // should not disconnect important peers
+  const importantPeers = getImportantPeers(peers, reps);
+  const candidatePeers = peers.filter((peer) => !importantPeers.has(peer));
+  // worse peer on top
+  const sortedPeers = candidatePeers.sort((peer1, peer2) => {
+    const peer1Rep = reps.getFromPeerId(peer1);
+    const peer2Rep = reps.getFromPeerId(peer2);
```
twoeths, line 88: Don't want to lose peers with better supported protocols

twoeths:
Update:
+ New CheckPeerAliveTask to ping peers and call network `metadata` to update reputation store if needed
+ DiversifyPeersBySubnetTask only find peers with subnet and connect to them to save time.
+ Automatically call `metadata` during handshake
+ DiversifyPeersBySubnetTask will disconnect selected peers if we have more than 90% of maxPeers
+ Handle `onPing`: Update our ReputationStore and issue `metadata` if needed

packages/lodestar/src/network/network.ts
```diff
@@ -132,25 +137,24 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
    * @param subnet the subnet calculated from committee index
    * @param inPeerIds peers already have this subnet
    */
-  private async connectToNewPeersBySubnet(subnet: number, inPeerIds: string[] = []): Promise<number> {
+  public async connectToNewPeersBySubnet(subnet: number, inPeerIds: string[] = []): Promise<number> {
     const discv5Peers = (await this.searchDiscv5Peers(subnet)) || [];
     const peerIds = discv5Peers.filter((peerId) => !inPeerIds.includes(peerId.toB58String()));
```
mpetrunic: I wonder if you should also skip peers in peerstore?

packages/lodestar/src/network/network.ts
```diff
@@ -132,25 +137,24 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
...
+          break;
         }
-      } catch (err) {
-        this.logger.warn(`Cannot get metadata from ${peerId.toB58String()}: ${err.message}`);
-      }
-      if (count < 10) {
-        // TODO: decide max peers per subnet to connect?
-        break;
+      } catch (e) {
+        this.logger.warn(
```
mpetrunic: put this in verbose, as there are a lot of dead and useless peers which just clutters logs

packages/lodestar/src/network/network.ts
```diff
@@ -132,25 +137,24 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
...
+          `Cannot connect to peer ${peerId.toB58String()} for subnet ${subnet}`,
+          e.message.substring(0, 30)
```
mpetrunic: why substring?

mpetrunic: in case of some new error we could loose valuable info

packages/lodestar/src/network/network.ts
```diff
@@ -132,25 +137,24 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
...
+    // make sure they still connect to same subnet
     for (const peerId of peerIds) {
-      // we'll dial thru sendRequest so don't need to do connect() like in the spec
+      // will automatically get metadata once we connect
       try {
-        const metadata = await this.reqResp.metadata(peerId);
-        if (metadata === null) throw Error("No metadata");
-        if (metadata.attnets[subnet]) {
-          count++;
+        await this.connect(peerId);
```
mpetrunic: This could be infinite loop, we have connection manager set to maxPeers connections, so this could loop until we prune peers. Maybe just attempt to connect one peer?

packages/lodestar/src/sync/utils/reputation.ts
```diff
@@ -0,0 +1,122 @@
...
+      logger.warn("Cannot get metadata for peer " + peer.toB58String());
+    }
+  }
+}
+
+/**
+ * Update new metadata to reputation store.
+ */
+export function updateMetadata(reps: IReputationStore, peerId: string, newMetadata: Metadata | null): void {
+  if (!reps) return;
```
mpetrunic: somewhat useless check as it's gonna fail in `handlePeerMetadataSequence`
Probably should be checked in network?

packages/lodestar/src/sync/utils/reputation.ts
mpetrunic: hmm this could be wierd as we might prefer peers that supports bunch of useless libp2p protocols. Maybe during sync prefer peers supporting blocks_by_range and after sync prefer blocks_by_root and gossip peers (not sure how to check those)?

packages/lodestar/src/sync/utils/reputation.ts
```diff
@@ -0,0 +1,122 @@
...
+ * before connecting to new peers with missing subnets.
+ * @param peers current connected peers
+ * @param newPeer number of peers with new subnets to connect to
+ * @param maxPeer max peer provided as option
+ * @param reps our reputation store
+ * @param retainRatio: 0 to 1, the ratio to consider disconnecting peers, 90% by default.
+ */
+export function selectPeersToDisconnect(
+  peers: PeerId[],
+  newPeer: number,
```
mpetrunic, line 70: Even if we don't have new peers, I think we should keep max peers at retain ration to allow new peers to connect and prune useless peers

packages/lodestar/src/tasks/index.ts
```diff
@@ -32,8 +36,11 @@ export class TasksService implements IService {
   private readonly sync: IBeaconSync;
   private readonly network: INetwork;
   private readonly logger: ILogger;
+  private readonly reps: IReputationStore;
 
   private interopSubnetsTask: InteropSubnetsJoiningTask;
+  private diversifyPeersTask: DiversifyPeersBySubnetTask;
+  private checkPeerAliveTask: CheckPeerAliveTask;
```
mpetrunic: Maybe we shouldn't just pile tasks here? I imagine this as a place where we execute tasks that affect multiple beacon chain components. But those two are strictly network tasks and should live there.

packages/lodestar/src/tasks/tasks/checkPeerAliveTask.ts
```diff
@@ -0,0 +1,66 @@
...
+    this.config = config;
+    this.network = modules.network;
+    this.logger = modules.logger;
+    this.reps = modules.reps;
+  }
+
+  public async start(): Promise<void> {
+    this.interval = setInterval(
+      this.run,
+      1 * this.config.params.SLOTS_PER_EPOCH * this.config.params.SECONDS_PER_SLOT * 1000
```
mpetrunic: I don't like sending batch of ping requests, maybe store last sent ping timestamp and send another if sufficient time has passed. That said, we can leave for now, I'm migrating metadata to peerstore anyways

packages/lodestar/src/sync/utils/reputation.ts
twoeths: yeah we can do that using sync, for now only blocks_by_range and blocks_by_root is meaningful to us so we can ignore others.

I'm not sure if we have protocols for gossip peers but let's double check with @wemeetagain

packages/lodestar/src/network/network.ts
twoeths: yeah that's a good idea to me because  there are a lot of errors when we connect to peers from discv5, doing that will filter some bad peers that we disconnected in the past, will try that

packages/lodestar/src/network/network.ts
twoeths: there are a lot of error, one of them is `onAbort` and it appears in multilines, also we run this task frequently for all missing subnets
can we go like this and monitor it for a while, will change to display more if we think it's valuable to do so?

packages/lodestar/src/network/network.ts
twoeths: there is `MAX_NUM_PEERS_PER_SUBNET`, we can tweak based on that

packages/lodestar/src/tasks/index.ts
twoeths: how about having them in subfolder of `tasks` like `tasks/network`? when I want to audit all routines we have I can just look at this

would like to hear others ideas too @wemeetagain @dapplion @3xtr4t3rr3str14l , to see if we want `tasks/network` or `network/tasks`

packages/lodestar/src/sync/utils/reputation.ts
twoeths: how to know useless peers? probably peers that don't connect to any subnets and has `supportedProtocols=[]`?

packages/lodestar/src/network/network.ts
mpetrunic: hmm, I think on Abort can be overriden?
I'm afraid if we receive some unique error it would be too late to change, rather leave it multilined until we figure out fix for it

packages/lodestar/src/network/network.ts
mpetrunic: Well yeah, but if we reach max connection, I'm afraid every connect will fail and we would never increase counter

packages/lodestar/src/sync/utils/reputation.ts
mpetrunic: peer that timeouts, that doesn't support required protocols (we still keep those as they might be useful to gossip) or that have bad gossip score later on

packages/lodestar/src/network/network.ts
twoeths: I just loop through the filtered peers that have the specified subnet, did not loop until we reach `MAX_NUM_PEERS_PER_SUBNET` :D

but yeah I'll switch to connect to the 1st possible peer as 2 peers per subnets are 128 which is too much

packages/lodestar/src/sync/utils/reputation.ts
twoeths: @mpetrunic seems like `DiversifyPeersBySubnetTask` should only be started after sync? Before node is synced, diversifying peers by subnet is not a concerned.

twoeths:
Update: handle `syncCompleted` event in `TaskService` and start `diversifyPeersTask, checkPeerAliveTask` once we see it

packages/lodestar/src/tasks/index.ts
wemeetagain: I agree with @mpetrunic's thoughts on these tasks being things that require multiple components.

I think these could be candidates here tho, both rely on sync completion to start. Imo this is fine for now.
I think one possibility would be to introduce a module for managing validator resources, handling any additional resources required for attesting/proposing. It would track connected validators, manage eth1data fetching/retrieving, manage subnet subscriptions, etc. All things that a simple read-only beacon chain doesn't require.

mpetrunic: Peer liveness and metadata tracking is just related to networking and should live there

twoeths:
Update: move network specific tasks to under `/network/tasks`

mpetrunic (approved):

wemeetagain (approved):

PR #1431, opened 2020-08-22
Fix ghost-failing unit tests

dapplion:
Current master will report unit tests as successfully run when no test is run. It uses an old version of mocha with doesn't handle Typescript parsing well and exits with code 0 when ts-node fails. To fix:

**Necessary changes**
- Bumped test dependencies to latest version. Mocha has breaking changes.
- Migrate deprecated mocha.opts to .mocharc.yaml
- Move mocha options from command arguments to the config file (i.e. colors)
- Fixed type errors that caused the tests to fail initially

With this changes now mocha exits with an error code if the test typescript compilation fails

**Other changes**
- Run tests across all packages with lerna and --no-bail, so all tests are run even if the first topological package fails. The entire suite fails with the first exit code.
- Remove useless imports of mocha functions
- Added a mock logger to a validator test that spammed the stdout
- Skipped a strange test that triggered an unhandled rejection after the test had completed


package.json
```diff
@@ -15,6 +15,8 @@
     "build:lib:watch": "lerna run build:lib:watch --parallel",
     "build:types:watch": "lerna run build:types:watch --parallel",
     "build:watch": "run-p build:lib:watch build:types:watch",
+    "test:unit": "node_modules/.bin/lerna run test:unit --no-bail",
+    "test:e2e": "node_modules/.bin/lerna run test:e2e --no-bail",
```
mpetrunic:
```suggestion
    "test:e2e": "lerna run test:e2e --no-bail",
```
works inside package.json^^

package.json
```diff
@@ -15,6 +15,8 @@
     "build:lib:watch": "lerna run build:lib:watch --parallel",
     "build:types:watch": "lerna run build:types:watch --parallel",
     "build:watch": "run-p build:lib:watch build:types:watch",
+    "test:unit": "node_modules/.bin/lerna run test:unit --no-bail",
```
mpetrunic:
```suggestion
    "test:unit": "lerna run test:unit --no-bail",
```

packages/lodestar-beacon-state-transition/package.json
```diff
@@ -44,9 +44,9 @@
   },
   "devDependencies": {
     "@types/chai": "4.2.0",
-    "@types/mocha": "^5.2.7",
+    "@types/mocha": "^8.0.3",
     "chai": "^4.2.0",
-    "mocha": "^6.2.2"
+    "mocha": "^8.1.1"
```
mpetrunic: since we have those in root package.json, does it work if you remove it here?

mpetrunic: if it works in other packages, we should probably remove it everywhere

.github/workflows/test.yml
```diff
@@ -35,11 +35,11 @@ jobs:
       - name: Lint
         run: node_modules/.bin/lerna run lint
```
mpetrunic: could you update this as well

mpetrunic (commented):
Awesome, much needed update.

cc @wemeetagain @tuyennhv Are we intentionally not running check types in ci anymore?

packages/lodestar-beacon-state-transition/package.json
dapplion: I was wondering about that and didn't find any conclusive answer. If a package is declared in the root does it have to be declared in the subpackages?

mpetrunic: Only prod dependencies should be kept in each package I think, because of npm publish

wemeetagain (approved):

twoeths:
@mpetrunic I prefer bringing it back, it was removed when we optimize CI last time as I remember, need to check with @wemeetagain

PR #1432, opened 2020-08-23
Fix loop edge case in chunkify function

strawdar:
While investigating some other stuff I found a likely bug in the chunkify function in `blocks.ts`. It seems to not produce the last chunk if it's supposed to be of length 1 (where `start` and `end` both equal `targetSlot`).

An alternate explanation is that the loop exits early if `i + blocksPerChunk` ever ends up equaling `targetSlot - 1`.

Along with a fix, I also included a new test that covers this case.


wemeetagain:
Thanks @strawberrydarling 
Looks good! Can you sign our CLA https://github.com/ChainSafe/lodestar/pull/1432#issuecomment-678747672 and we can merge this

mpetrunic (approved):

wemeetagain (approved):

PR #1435, opened 2020-08-24
Sync: handle failed to get block range

### LINKED ISSUE #1433: Initial Sync: handle waitForPeers
**Describe the bug**
Initial Sync misunderstand that "no peer" means no block in range
image

**Expected behavior**
Initial Sync should retry same range when waiting for peers

twoeths:
resolves #1433 

In case of failed to fetch block range, we need to yield null instead of yield nothing so that we can handle in `processSyncBlocks`. Otherwise it's not possible to distinguish empty range vs failed range in `processSyncBlocks` since it always pass `for..await`.

### DIFF
--- a/packages/lodestar/src/sync/initial/fast/index.ts
+++ b/packages/lodestar/src/sync/initial/fast/index.ts
@@ -20,7 +20,7 @@ import {GENESIS_EPOCH, Method} from "../../../constants";
 import {ISyncStats, SyncStats} from "../../stats";
 import {IBeaconDb} from "../../../db";
 
-export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}) implements InitialSync {
+export class FastSync extends (EventEmitter as {new(): InitialSyncEventEmitter}) implements InitialSync {
   private readonly opts: ISyncOptions;
   private readonly config: IBeaconConfig;
   private readonly chain: IBeaconChain;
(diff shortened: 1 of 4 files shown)


packages/lodestar/src/sync/initial/fast/index.ts
```diff
@@ -121,7 +121,7 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
   };
 
   private setBlockImportTarget = (fromSlot?: Slot): void => {
-    const lastTarget = fromSlot || this.blockImportTarget;
+    const lastTarget = typeof fromSlot === "number" ? fromSlot : this.blockImportTarget;
```
mpetrunic: I think you are looking for https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing :smile:

packages/lodestar/src/sync/regular/naive/naive.ts
```diff
@@ -95,7 +95,7 @@ export class NaiveRegularSync extends (EventEmitter as {new (): RegularSyncEvent
   }
 
   private setTarget = (newTarget?: Slot, triggerSync = true): void => {
-    newTarget = newTarget || this.getNewTarget();
+    newTarget = typeof newTarget === "number" ? newTarget : this.getNewTarget();
```
mpetrunic: same as above

mpetrunic (commented):
Seems good, just small syntax sugar notice

mpetrunic (dismissed):

mpetrunic:
@tuyennhv seems like there are lint errors

mpetrunic (approved):

wemeetagain (approved):

PR #1438, opened 2020-08-24
Optimize gossip AggregateAndProof validation

mpetrunic:
There is still some outstanding issues like:
- we don't check if gossiped aggregate was something we already received and checked in block or trough api (not a major issue)
- ideally, we would have some pool that would reverify once we have prestate and attestation block (it's fine since we will eventually receive it trough some other block)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -66,9 +66,6 @@ module.exports = {
       "peerDependencies": false
     }],
     "func-call-spacing": "off",
-    "max-len": ["error", {
-      "code": 120
-    }],
     //if --fix is run it messes imports like /lib/presets/minimal & /lib/presets/mainnet
     "import/no-duplicates": "off",
     "node/no-deprecated-api": "error",
--- a/packages/lodestar/src/network/gossip/validation/aggregateAndProof.ts
+++ b/packages/lodestar/src/network/gossip/validation/aggregateAndProof.ts
@@ -1,0 +1,204 @@
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
+import {IBeaconChain} from "../../../chain";
+import {IBeaconDb} from "../../../db/api";
+import {Context, ILogger} from "@chainsafe/lodestar-utils";
+import {
+  AggregateAndProof,
+  Attestation,
+  BeaconState,
+  Epoch,
+  SignedAggregateAndProof,
+  Slot,
+} from "@chainsafe/lodestar-types";
+import {ExtendedValidatorResult} from "../constants";
+import {toHexString} from "@chainsafe/ssz";
+import {ATTESTATION_PROPAGATION_SLOT_RANGE, DomainType, MAXIMUM_GOSSIP_CLOCK_DISPARITY} from "../../../constants";
+import {
+  computeEpochAtSlot,
+  computeSigningRoot,
+  computeStartSlotAtEpoch,
+  getCurrentSlot,
+  getDomain,
+  isAggregatorFromCommitteeLength,
+} from "@chainsafe/lodestar-beacon-state-transition";
+import {isAttestingToInValidBlock} from "./attestation";
+import {getAttestationPreState} from "../utils";
+import {processSlots} from "@chainsafe/lodestar-beacon-state-transition/lib/fast/slot";
+import {PublicKey, Signature} from "@chainsafe/bls";
+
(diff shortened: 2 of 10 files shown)


packages/lodestar/src/network/gossip/validation/aggregateAndProof.ts
```diff
@@ -0,0 +1,170 @@
...
+  const attestationPreState = await getAttestationPreState(config, chain, db, attestation.data.target);
+  if (!attestationPreState) {
+    logger.warn("Ignored gossip aggregate and proof", {reason: "missing attestation prestate", ...logContext});
+    return ExtendedValidatorResult.ignore;
+  }
+  const {state, epochCtx} = attestationPreState;
+  //committee changes on epoch, so advance only if different epoch
+  const attEpoch = computeEpochAtSlot(config, attestation.data.slot);
+  if (attEpoch > computeEpochAtSlot(config, state.slot)) {
+    processSlots(epochCtx, state, computeStartSlotAtEpoch(config, attEpoch));
```
twoeths, line 109: since we have out-of-memory issue, we have to care about the use of `processSlots` more.

what's the plan for DOS attack? i.e. some peers can send us an attestation with `attestation.data.slot` so far away from current slot.

similar concern for `processSlots` inside `getAttestationPreState`

mpetrunic: Attestation pre state:
Well attestation.data.slot can only be at most current slot or is rejected.
So issue is if it's way ahead of our head slot.
In that case we check if we have checkpoint state in cache.
If we don't have it we search for target.root amongst unfinalized blocks and process him up to start slot of target.epoch. In normal conditions there won't be slot processing. Issue could occur if during non-finalization, some validators are deliberately producing attestation with head target epoch and very old non finalized block. I would wait till we see how other clients solved this issue. Simply limiting this to few slots could introduce bugs and maybe get us on wrong chain.

Referenced line:
This happens if slots in next epoch are skip slots so target epoch is still previous epoch.

.eslintrc.js
```diff
@@ -66,9 +66,6 @@ module.exports = {
       "peerDependencies": false
     }],
     "func-call-spacing": "off",
-    "max-len": ["error", {
```
wemeetagain, line 69: is this meant to be here?

mpetrunic: Rule conflicts with prettier on import statement

packages/lodestar/src/network/gossip/validation/aggregateAndProof.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  if (await isAttestingToInValidBlock(db, aggregate)) {
+    logger.warn("Rejected gossip aggregate and proof", {
+      reason: "attesting to invalid block",
+      ...logContext,
+    });
+    return ExtendedValidatorResult.reject;
+  }
+
+  //TODO: check pool of aggregates if already seen (not a dos vector check)
```
3xtr4t3rr3str14l, line 69: is this something that will be done in this PR?  or does a separate issue need to be made?

packages/lodestar/src/network/gossip/validation/aggregateAndProof.ts
```diff
@@ -0,0 +1,170 @@
...
+      computeEpochAtSlot(config, attestation.data.slot),
+      aggregator,
+      aggregateAndProof
+    )
+  ) {
+    logger.warn("Rejected gossip aggregate and proof", {reason: "invalid signature", ...logContext});
+    return ExtendedValidatorResult.reject;
+  }
+
+  //TODO: once we have pool, check if aggregate block is seen and has target as ancestor
```
3xtr4t3rr3str14l, line 147: same question as the previous TODO

packages/lodestar/src/network/gossip/validation/aggregateAndProof.ts
mpetrunic: Separate issue, not really a priority i guess

packages/lodestar/src/network/gossip/validation/aggregateAndProof.ts
wemeetagain: separate PR

wemeetagain (approved):

PR #1440, opened 2020-08-24
clean validator exit

### LINKED ISSUE #1437: Ensure clean validator exit
Whens validator stop is called:
- we should invoke stop on both block and attestation service
- in both service stop should clear interval and timeouts

3xtr4t3rr3str14l:
fixes #1437


mpetrunic:
I think you need to clear interval in attestation service. And you should unsubscribe from new slot/epoch rather then calling disconnect i think

3xtr4t3rr3str14l:
> I think you need to clear interval in attestation service. And you should unsubscribe from new slot/epoch rather then calling disconnect i think

@mpetrunic , I'm not seeing any intervals in the attestation service class itself, but `disconnect` clears the `beaconNodeInterval` of the api provider for each service, which is the interval I assumed you meant.  Is that not the correct interval?  Which interval(s) do you think need to be cleared?

mpetrunic:
Line 168 and 196 in attestation service. Aggregator timeout is simple. Block wait should probably be solved by abort signal

3xtr4t3rr3str14l:
@mpetrunic just pushed changes for the attestation.stop that i think might work, but i can't test it yet because i haven't been able to fully sync the beacon yet.  if you have suggestions for testing this, i'm open to hearing about it.  otherwise, i may try later if i'm fully synced (currently waiting for sync now in my local dev environment)

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -47,6 +48,8 @@ export class AttestationService {
   private readonly logger: ILogger;
 
   private nextAttesterDuties: Map<Slot, IAttesterDuty[]> = new Map();
+  private aggregatorTimeout?: NodeJS.Timeout;
```
mpetrunic: I'm afraid we can have multiple aggreagtor timeouts at the same time.
Might be better to wrap setting timeout into function with abort signal.
You can look at examples here:
https://ckeditor.com/blog/Aborting-a-signal-how-to-cancel-an-asynchronous-task-in-JavaScript/

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -47,6 +48,8 @@ export class AttestationService {
...
+  private controller: AbortController = new AbortController();
```
mpetrunic: might make more sense to initialize in on start (soo we can start-stop multiple times)

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -191,11 +198,16 @@ export class AttestationService {
...
-    this.logger.debug("Waiting for slot block", {slot});
+  private async waitForAttestationBlock(slot: Slot, signal: AbortSignal): Promise<void> {
+    this.logger.info("Waiting for slot block", {slot});
     const eventSource = new EventSource(`${this.provider.url}/lodestar/blocks/stream`, {
       https: {rejectUnauthorized: false},
     });
+
+    if (signal.aborted) {
+      this.logger.verbose("AttestationService: Abort waitForAttestationBlock");
+    }
```
mpetrunic: signal.on("abort", () => {clean(); resolve();}) should be inside promise, since with this, you cannot abort once it goes past this line

mpetrunic (changes requested):

packages/lodestar-validator/src/services/attestation.ts
3xtr4t3rr3str14l: what would clean() be in this case?  the clearTimeout?

mpetrunic: Yes and close event stream

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -66,16 +68,24 @@ export class AttestationService {
...
 
     if (computeStartSlotAtEpoch(this.config, computeEpochAtSlot(this.config, slot)) !== slot) {
-      //trigger next epoch duties
+      // trigger next epoch duties
       await this.onNewEpoch(computeEpochAtSlot(this.config, slot));
     }
   };
 
+  public stop = async (): Promise<void> => {
+    await this.provider.disconnect();
```
mpetrunic: I don't think we need that in services (unsubscribe would be okeish but we don't need it). It's fine just to call disconnect in validator class

mpetrunic:
It's much better now

3xtr4t3rr3str14l:
using bigint literals was messing with the tsconfig target (and wasn't being used outside of the unit tests) when i started using the optional chaining operator in src/, so i changed the literals to `BigInt(number)`s and changed the tsconfig target to es2019 to allow the optional chaining operator

mpetrunic (dismissed):
You should've rebased, now you leaked password forever :smile: 
https://github.com/ChainSafe/lodestar/pull/1440/commits/c60581139a67ac88274fdfa98e0912fe4c658e9e

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -165,7 +175,13 @@ export class AttestationService {
     }
 
     if (duty.isAggregator) {
-      setTimeout(async () => {
+      const timeout = setTimeout(async (signal = abortSignal) => {
+        this.logger.debug("AttestationService: Start waitForAggregate");
+        signal?.addEventListener("abort", () => {
```
dapplion: You could just do 
```ts
if (signal) signal.addEventListener
```
And won't need the optional chaining operator. We can figure out in a latter PR a common policy for ES versioning

3xtr4t3rr3str14l: Right, but I figured we don't need the bigint literals in lodestar-validator anyway, so no reason to target ES2020 regardless.

3xtr4t3rr3str14l:
> You should've rebased, now you leaked password forever smile
> [c605811](https://github.com/ChainSafe/lodestar/commit/c60581139a67ac88274fdfa98e0912fe4c658e9e)

ya glad it's just a test pass file from cli testing hah

packages/lodestar-validator/src/services/attestation.ts
3xtr4t3rr3str14l: Also, worth mentioning that the other lodestar packages don't have es2020 target in their tsconfig either.

mpetrunic: I think we need to remove abort event listener as well :(

You can also pass third argument {once: true} to automatically clear if event is triggered

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -187,16 +203,25 @@ export class AttestationService {
...
       https: {rejectUnauthorized: false},
     });
+
     await new Promise((resolve) => {
       const timeout = setTimeout(() => {
         this.logger.debug("Timed out slot block waiting");
         resolve();
       }, (this.config.params.SECONDS_PER_SLOT / 3) * 1000);
+
+      signal?.addEventListener("abort", () => {
```
mpetrunic: same here

3xtr4t3rr3str14l: just added in both spots

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -177,10 +177,14 @@ export class AttestationService {
...
-          this.logger.debug("AttestationService: Abort waitForAggregate");
-        });
+        signal?.addEventListener(
+          "abort",
+          () => {
+            clearTimeout(timeout);
+            this.logger.debug("AttestationService: Abort waitForAggregate");
+          },
+          {once: true}
+        );
```
dapplion: The abortable setTimeout should be extracted into a utility function and unit tested

dapplion (dismissed):

packages/lodestar-validator/src/services/attestation.ts
3xtr4t3rr3str14l: done

3xtr4t3rr3str14l: er, well, actually, forgot to add the unit test.  will do that here in a min

mpetrunic (dismissed):

packages/lodestar-validator/src/util/misc.ts
```diff
@@ -1,3 +1,11 @@
 export function sleep(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
+
+export function abortableTimeout(
+  signal: AbortSignal | undefined,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  callback: (this: AbortSignal, ev: Event) => any
```
mpetrunic:
```suggestion
  callback: (this: AbortSignal, ev: Event) => void
```

packages/lodestar-validator/src/util/misc.ts
```diff
@@ -1,3 +1,11 @@
 export function sleep(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
+
+export function abortableTimeout(
+  signal: AbortSignal | undefined,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
mpetrunic, line 7:
```suggestion
```

packages/lodestar-validator/src/validator.ts
```diff
@@ -72,6 +72,7 @@ export class Validator {
   public async stop(): Promise<void> {
     this.isRunning = false;
     await this.apiClient.disconnect();
+    if (this.attestationService) await this.attestationService.stop();
```
mpetrunic, line 75: stop block service as well so if something gets added to block service stop method,  gets stopped as well

mpetrunic (changes requested):

packages/lodestar-validator/src/util/misc.ts
```diff
@@ -1,3 +1,11 @@
 export function sleep(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
+
+export function abortableTimeout(
```
dapplion, line 5: An abortableTimeout should start the timeout for you and cancel it on abort
```ts
export function abortableTimeout(callback, timeoutMs, signal): Timeout {
  const timeout = setTimeout(callback, timeoutMs)
  signal?.addEventListener("abort", () => {
    clearTimeout(timeout)
  }, {once: true});

  return timeout;
}
```
Or something like this I think would make more sense

dapplion: You could even use this abortable sleep function
https://github.com/ChainSafe/lodestar/blob/4295ce0f4c3f2e8e8a39e6ce01f813219b668abc/packages/lodestar/src/util/sleep.ts#L8-L30

And use the sleep as if it was a setTimeout do 
```ts
sleep(1000, signal).then(
  () => {
    // timeout completed, do stuff
  },
  () => {
    // timeout aborted, do stuff
  }
)
```
I would like to know what others think about this approach

3xtr4t3rr3str14l: Not sure how everyone else feels, but I like the first option (the modification of `abortableTimeout`) you brought up in your first comment more because it's simpler.  If we were to go that route though, we would also need an abortCallback or similar to handle logging, closing the EventSource, etc.

Also, `sleep` would force the whole timeout to be a Promise, which I'm not sure would be applicable in the same way here:
https://github.com/ChainSafe/lodestar/pull/1440/files#diff-cd304f421d116e19fd26363114c73e94R179

3xtr4t3rr3str14l: Or would it be applicable there?  I might not be seeing it correctly.

packages/lodestar-validator/src/validator.ts
3xtr4t3rr3str14l: done

mpetrunic (approved):

PR #1443, opened 2020-08-24
Decompose IEth1Notifier into functional components

### LINKED ISSUE #1103: Decompose IEth1Notifier into parts

dapplion:
This PR executes the idea of https://github.com/ChainSafe/lodestar/issues/1103 to separate concerns in the previously monolithic `IEth1Notifier`. It has two components
- `Eth1Provider`: Simple wrapper around the ethers provider with convenience methods to consume just the necessary data.
- `getDepositsStream()` and `getDepositsAndBlockStreamForGenesis()`: Functions that take the `Eth1Provider` as argument and return an AsyncGenerator of deposit events. The second function is optimized with @tuyennhv strategy to fetch blocks efficiently when close to the min genesis time.

Interacting with the database is out of the scope of these components and should be done in its consumers. This PR will be followed by a couple more to
- Query deposit and eth1 block data for block production
- Read and write an intermediary state of the GenesisBuilder

Fixes https://github.com/ChainSafe/lodestar/issues/1103.


packages/lodestar/src/eth1/stream.ts
```diff
@@ -0,0 +1,66 @@
...
+import {IDepositEvent, IBatchDepositEvents, IEth1Block, IEth1Provider, IEth1StreamParams} from "./interface";
+import {groupDepositEventsByBlock} from "./utils/groupDepositEventsByBlock";
+import {optimizeNextBlockDiffForGenesis} from "./utils/optimizeNextBlockDiffForGenesis";
+import {sleep} from "../util/sleep";
+
+/**
+ * Phase 1 of genesis building.
+ * Not enough validators, only stream deposits
+ */
+export async function* getDepositsStream(
```
wemeetagain, line 16: We may need to pass an abort signal to these functions (and modify the `sleep` function).
Otherwise we will have to wait for the sleep on exit

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -19,90 +25,112 @@ import {
...
+          proof: this.depositTree.tree().getSingleProof(this.depositTree.gindexOfProperty(depositEvent.index)),
+          data: depositEvent,
+        };
+      });
+
+    applyDeposits(this.config, this.state, newDeposits, this.depositTree);
+
+    this.logger.verbose(`Found ${this.state.validators.length} validators to genesis`);
+
+    // TODO: If necessary persist deposits here to this.db.depositData, this.db.depositDataRoot
```
wemeetagain, line 133: IMO genesis builder can/should be stateless, not have access to `db`, just return data to be saved to db by its consumer.

We might not even need/want genesis builder to be a class:
idea:
```typescript
// in chain/chain.ts
try {
  const { state, depositTree, blockNumber } = await buildGenesis(
    this.eth1, this.logger, this.db.getGenesisWork()
  );
  ...
} catch (e) {
  if (e instanceof BuildGenesisError) {
    await this.db.setGenesisWork(e.stuff);
  }
}

File: packages/lodestar/src/chain/genesis/genesis.ts [hunk 3]
```diff
@@ -19,90 +25,112 @@ import {
...
+    this.eth1Params = {
+      ...config.params,
+      MAX_BLOCKS_PER_POLL: 10000,
+    };
   }
 
   /**
    * Get eth1 deposit events and blocks and apply to this.state until we found genesis.
    */
   public async waitForGenesis(): Promise<TreeBacked<BeaconState>> {
```
wemeetagain: Should probably return state, depositTree, and blockNumber

dapplion:
To share some insights I found while developing this:
- get_logs calls to eth1 nodes can handle a very large amount of logs per request, > 10000 blocks with a reply of many MB. Going for the highest possible amount of blocks per request significantly speeds up genesis state generation. Compare to querying 100 blocks at a time for example.
- If you are querying large amounts of blocks (> 10000) the bottleneck becomes BLS signature verification. To test it, disabling BLS signature verification the node was able to process deposits x2 ~ x3 faster.
- Merkle tree verification of each deposit leaf does not affect genesis state generation speed in a noticeable way.
- Generating the genesis state of a large network will take some time no matter what. It has to download a lot of data and do +16k signature checks. If the Eth1 node is not in the local network the roundtrip of each HTTP request also adds up fast.

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -5,104 +5,132 @@
...
+  public async waitForGenesis(): Promise<IGenesisResult> {
+    await this.eth1Provider.validateContract();
+
+    // TODO: Load data from data from this.db.depositData, this.db.depositDataRoot
+    // And start from a more recent fromBlock
+    const blockNumberLastestInDb = this.eth1Provider.deployBlock;
+
+    const blockNumberValidatorGenesis = await this.waitForGenesisValidators(blockNumberLastestInDb);
+
+    const controller = new AbortController();
```
mpetrunic: I think you should receive signal as method argument so high level module calling this can abort this upon "stop" method invocation

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -5,104 +5,132 @@
...
-        }
-      })();
-    };
+  /**
+   * First phase of waiting for genesis.
+   * Stream deposits events in batches as big as possible without querying block data
+   * @returns Block number at which there are enough active validators is state for genesis
+   */
+  private async waitForGenesisValidators(fromBlock: number): Promise<number> {
+    const controller = new AbortController();
```
mpetrunic: same as above

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -5,104 +5,132 @@
...
-    return state;
-  }
+  public async waitForGenesis(): Promise<IGenesisResult> {
+    await this.eth1Provider.validateContract();
+
+    // TODO: Load data from data from this.db.depositData, this.db.depositDataRoot
+    // And start from a more recent fromBlock
+    const blockNumberLastestInDb = this.eth1Provider.deployBlock;
+
+    const blockNumberValidatorGenesis = await this.waitForGenesisValidators(blockNumberLastestInDb);
```
mpetrunic, line 68: handle if this ends up undefined because it's stopped

mpetrunic (changes requested):
Noice!

packages/lodestar/test/e2e/eth1/ethers.test.ts
```diff
@@ -0,0 +1,47 @@
...
+import {Eth1Provider, IEth1Block} from "../../../src/eth1";
+import {IEth1Options} from "../../../src/eth1/options";
+import {getMedallaConfig, medalla} from "./util";
+
+describe("eth1 / Eth1Provider", function () {
+  this.timeout("2 min");
+
+  const eth1Options: IEth1Options = {
+    enabled: true,
+    providerUrl: medalla.providerUrl,
```
mpetrunic, line 12: I don't think it's okay to run our e2e tests against prylabs public goerli node but
alternative is to setup private chain with deposit contract and deposits :disappointed:

mpetrunic (commented):
Maybe some unit tests to ensure aborting of streams work?

packages/lodestar/test/e2e/eth1/ethers.test.ts
wemeetagain: I think we were doing this before :sweat: 
Maybe we make an issue for this.

dapplion: Let's get some infura keys meanwhile?

packages/lodestar/src/eth1/stream.ts
```diff
@@ -0,0 +1,75 @@
...
+    const toBlock = Math.min(remoteFollowBlock, fromBlock + params.MAX_BLOCKS_PER_POLL);
+    const logs = await provider.getDepositEvents(fromBlock, toBlock);
+    for (const batchedDeposits of groupDepositEventsByBlock(logs)) {
+      yield batchedDeposits;
+    }
+
+    fromBlock = toBlock;
+
+    if (signal && signal.aborted) throw new ErrorAborted();
+    if (toBlock >= remoteFollowBlock) await sleep(params.SECONDS_PER_ETH1_BLOCK * 1000, signal);
```
dapplion: `sleep()` will throw ErrorAbort if the abort signal is already aborted. So the check above it might be unnecessary. Should I delete it or leave as a sanity check?

packages/lodestar/test/unit/chain/genesis/genesis.test.ts
```diff
@@ -20,77 +21,101 @@ describe("genesis builder", function () {
...
     );
   });
 
-  function generateDeposit(index: ValidatorIndex): DepositData {
-    const domain = computeDomain(config, DomainType.DEPOSIT);
-    const depositMessage = {
-      pubkey: keypairs[index].publicKey.toBytesCompressed(),
-      withdrawalCredentials: Buffer.alloc(32, index),
-      amount: BigInt(32 * 1000000000000000000),
+  it("should abort building genesis state", async () => {
```
dapplion, line 83: @mpetrunic I'm testing stream aborting through this test. Should I test them directly anyway? stream abort tests are very similar to the one below anyway.

packages/lodestar/src/util/sleep.ts
```diff
@@ -1,3 +1,30 @@
-export function sleep(ms: number): Promise<void> {
-  return new Promise((resolve) => setTimeout(resolve, ms));
+import {AbortSignal} from "abort-controller";
+import {ErrorAborted} from "./errors";
+
+/**
+ * Abortable sleep function. Cleans everything on all cases preventing leaks
+ * On abort throws ErrorAborted
+ */
+export async function sleep(ms: number, signal?: AbortSignal): Promise<void> {
```
dapplion, line 8: Changed the behaviour of sleep on abort to throw an Error. It makes flow control much easier. In the genesis builder it would be as simple to catch in `Chain` for `ErrorAborted` and log or ignore

packages/lodestar/src/eth1/depositContract.ts
```diff
@@ -0,0 +1,15 @@
...
...05080610b0052610b009050602060c0825160208401600060025af1610ede57600080fd5b60c0519050602082610c800101526020810190506000610300600880602084610c0001018260208501600060046012f15050805182019150506000601860208206610b800160208284011115610f3357600080fd5b602080610ba082610700600060046015f1505081815280905090509050601880602084610c0001018260208501600060046014f150508051820191505061084051602082610c0001015260208101905080610c0052610c009050602060c0825160208401600060025af1610fa657600080fd5b60c0519050602082610c8001015260208101905080610c8052610c809050602060c0825160208401600060025af1610fdd57600080fd5b60c0519050610ae052606435610ae05114610ff757600080fd5b600180546001825401101561100b57600080fd5b6001815401815550600154610d0052610d2060006020818352015b60016001610d005116141561105b57610ae051610d20516020811061104a57600080fd5b600060c052602060c02001556110ef565b6000610d20516020811061106e57600080fd5b600060c052602060c0200154602082610d40010152602081019050610ae051602082610d4001015260208101905080610d4052610d409050602060c0825160208401600060025af16110bf57600080fd5b60c0519050610ae052610d0060026110d657600080fd5b60028151048152505b8151600101808352811415611026575b5050005b5b60006000fd5b61017f6112790361017f60003961017f611279036000f3"
```
dapplion: Is it necessary to keep the bytecode of the contract in the lodestar source? Doesn't seem to be used by anyone and if it was it should be stored in the tests not the source.

packages/lodestar/src/eth1/stream.ts
wemeetagain: I don't think we should have it. 
IMO we're fine with just the assumption that the signal has not already been thrown if you want to use it to abort the stream.

packages/lodestar/src/chain/genesis/genesis.ts
dapplion: Removed `db` from genesis builder modules

packages/lodestar/src/eth1/depositContract.ts
wemeetagain: Yea i agree. If at some point we decide to make a deposit contract utility library it could go there, for now just in tests.

packages/lodestar/src/chain/genesis/genesis.ts
```diff
@@ -5,104 +5,128 @@
...
+          state: this.state,
+          depositTree: this.depositTree,
+          block,
+        };
       }
-      throw Error("Did not found genesis state and block source stopped");
-    };
+    }
+
+    throw Error("depositsStream stopped without a valid genesis state");
```
mpetrunic, line 93: I'm pretty sure this error would end up as unhandled rejection, as we don't handle error all the way till chain.start

packages/lodestar/src/eth1/stream.ts
```diff
@@ -0,0 +1,75 @@
...
+      provider.getBlock(toBlock),
+    ]);
+    yield [logs, block];
+
+    const remoteFollowBlock = await getRemoteFollowBlock(provider, params);
+    const nextBlockDiff = optimizeNextBlockDiffForGenesis(block, params);
+    fromBlock = toBlock;
+    toBlock = Math.min(remoteFollowBlock, fromBlock + Math.min(nextBlockDiff, params.MAX_BLOCKS_PER_POLL));
+
+    if (signal && signal.aborted) throw new ErrorAborted();
```
mpetrunic: you don't need this as well now

packages/lodestar/src/chain/genesis/genesis.ts
dapplion: What would you do here instead?

mpetrunic: I think that's fine, we should probably catch error in chain.start

wemeetagain (dismissed):

wemeetagain:
file conflicts

mpetrunic (dismissed):

wemeetagain (dismissed):

wemeetagain:
file conflict

mpetrunic:
there are still conflicts :smile:

packages/lodestar/test/unit/chain/genesis/genesis.test.ts
mpetrunic: think this is fine

mpetrunic (approved):

wemeetagain (approved):

PR #1444, opened 2020-08-24
Allow exiting process by force

dapplion:
The current way of handling process signals causes almost always to create a zombie unexited process.
According to the docs https://nodejs.org/api/process.html#process_signal_events

> 'SIGTERM' and 'SIGINT' have default handlers on non-Windows platforms that reset the terminal mode before exiting with code 128 + signal number. If one of these signals has a listener installed, its default behavior will be removed (Node.js will no longer exit).

So by attaching a cleanup listener to SIGINT and never calling process.exit() afterwards the only option for the process to exit is if the event loop becomes empty. This would be ideal but it's a very difficult task to accomplish having so many components and dependencies. A single unresolved promise or listener would leave the process zombie.

This PR adds a second listener to SIGINT so that if the user hits CTRL+C twice the process will exit by force.

### DIFF
--- a/packages/lodestar-cli/src/util/process.ts
+++ b/packages/lodestar-cli/src/util/process.ts
@@ -6,17 +6,22 @@ const exitSignals = ["SIGTERM", "SIGINT"] as NodeJS.Signals[];
  * user forcibly kills the process by doing CTRL+C again
  * @param cleanUpFunction
  */
-export function onProcessSIGINT(cleanUpFunction: () => Promise<void>): void {
+export function onProcessSIGINT(
+  cleanUpFunction: () => Promise<void>,
+  // eslint-disable-next-line no-console
+  logFn: (msg: string) => void = console.log
+): void {
   for (const signal of exitSignals) {
     process.once(signal, async () => {
-      // eslint-disable-next-line no-console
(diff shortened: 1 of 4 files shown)


packages/lodestar-cli/src/util/process.ts
```diff
@@ -0,0 +1,22 @@
+const exitSignals = ["SIGTERM", "SIGINT"] as NodeJS.Signals[];
+
+/**
+ * All CLI handlers should register this callback to exit properly and not leave
+ * a process hanging forever. Pass a clean function that will be run until the
+ * user forcibly kills the process by doing CTRL+C again
+ * @param cleanUpFunction
+ */
+export function onProcessSIGINT(cleanUpFunction: () => Promise<void>): void {
```
dapplion: Not super happy about this function's name. Any suggestion that's better self-describing is super appreciated!

mpetrunic (approved):

packages/lodestar-cli/src/util/process.ts
mpetrunic: just onProcessStop ?

twoeths (approved):

wemeetagain (approved):

PR #1447, opened 2020-08-25
Better supported protocols handling

### LINKED ISSUE #1325: Log the peer count
See https://discordapp.com/channels/593655374469660673/593655641445367808/740074151775043594

### LINKED ISSUE #1430: Delete peer from peer store if unsupported protocol
**What is your question?**

When we find peer, it get's stored in peer store and persisted on disk. When we restart we loose a lot of time probing those same peers that doesn't support status rpc protocol (maybe different chain or something).

Should we delete peer from peer store on unsupported protocol? Would be even better if we could keep blacklist in peerstore so we don't attempt to connect again

mpetrunic:
requires: https://github.com/ChainSafe/libp2p-ts/pull/16

Changelog:
- update libp2p latest minor version
- use builtin peer identified protocols
- exterminate unhandled rejection in goodbye request
- log peer count periodically resolves #1325 
- reduced warn log clutter

resolves #1430


packages/lodestar/src/sync/initial/fast/index.ts
```diff
@@ -209,6 +204,15 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
     }
   };
 
+  private getPeerReputations(): IReputation[] {
```
wemeetagain, line 207: Should this live in network module? It would be nice to store peer reputations/scores in metadata store.

Maybe separate PR?

mpetrunic: Well it's kinda specific to sync, as it gets only reputations from peers supporting sync protocols

packages/lodestar/src/sync/reqResp/reqResp.ts
```diff
@@ -104,33 +106,23 @@ export class BeaconReqRespHandler implements IReqRespHandler {
...
       this.network.reqResp.sendResponse(id, null, status);
-      isSuccess = true;
     } catch (e) {
       this.logger.error("Failed to create response status", e.message);
       this.network.reqResp.sendResponse(id, e);
     }
-    if (isSuccess) {
-      // response Status request first
-      try {
-        this.reps.get(peerId.toB58String()).supportedProtocols = await getPeerSupportedProtocols(
```
twoeths, line 124: if we don't use `getPeerSupportedProtocols` we can remove it.

packages/lodestar/src/sync/initial/fast/index.ts
```diff
@@ -220,17 +224,18 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
...
-        rep &&
-        rep.supportedProtocols.includes(Method.BeaconBlocksByRange) &&
-        rep.latestStatus &&
-        rep.latestStatus.finalizedEpoch >= this.targetCheckpoint!.epoch
-      ) {
-        validPeers.push(peer);
-      }
-      return validPeers;
-    }, [] as PeerId[]);
+    return this.network
```
twoeths, line 227: initially how can we have `peer.protocols`? curious how libp2p populate it without test requests

mpetrunic: It's using identify protocol:
https://github.com/libp2p/specs/tree/master/identify

If we start encountering that peer lies about it, we can add piece of code to remove it from protobook

twoeths:
Lgtm, it's nice not to have to deal with supported protocols. I synced to slot 8k without seeing the UNSUPPORTED_PROTOCOL error.

twoeths (approved):

wemeetagain (approved):

PR #1450, opened 2020-08-27
Track validators connected to node to subscribe random subnets

### LINKED ISSUE #1382: Lack of random subnet subscription
**Describe the bug**

We don't track how many active validators are attached to node so we are always to subscribed to one random subnet.

**Expected behavior**

Use validator duties or attestation publish to track how many active validators we have and subscribe to appropriate number of random subnets.

**Steps to Reproduce**

**Screenshots**

 - OS: 
 - Version: 
 - Branch: 
 - Commit hash:

twoeths:
resolves #1382 

+ Track validators connected to node in newly created `activeValidatorCache`
+ Update it in `beacon_committee_subscriptions` api
+ Remove inactive validators using `pruneFinalized`
+ Use `activeValidatorCache` in `interopJoiningSubnetTask` per new epoch

### DIFF
--- a/packages/lodestar/src/api/impl/validator/validator.ts
+++ b/packages/lodestar/src/api/impl/validator/validator.ts
@@ -3,6 +3,7 @@
  */
 
 import bls, {Signature} from "@chainsafe/bls";
+import {computeEpochAtSlot} from "@chainsafe/lodestar-beacon-state-transition";
 import {computeStartSlotAtEpoch, computeSubnetForCommitteesAtSlot} from "@chainsafe/lodestar-beacon-state-transition";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {
@@ -119,6 +120,7 @@ export class ValidatorApi implements IValidatorApi {
     if (validatorIndices.length === 0) throw new ApiError(400, "No validator to get attester duties");
     if (epoch > this.chain.clock.currentEpoch + 1)
       throw new ApiError(400, "Cannot get duties for epoch more than one ahead");
+    if (epoch < this.chain.clock.currentEpoch) throw new ApiError(400, "Not support getting duties for old epochs");
     const {epochCtx, state} = await this.chain.getHeadStateContextAtCurrentEpoch();
     return validatorIndices
       .map((validatorIndex) => {
@@ -202,6 +204,20 @@ export class ValidatorApi implements IValidatorApi {
     isAggregator: boolean
   ): Promise<void> {
     await checkSyncStatus(this.config, this.sync);
+    const currentSlot = this.chain.clock.currentSlot;
+    const epoch = computeEpochAtSlot(this.config, slot);
+    const currentEpoch = this.chain.clock.currentEpoch;
+    if (slot <= currentSlot) {
+      throw new ApiError(400, `Cannot prepare subnet for invalid slot ${slot}, current slot: ${currentSlot}`);
+    }
+    if (epoch > currentEpoch + 1) {
+      throw new ApiError(400, `Cannot prepare subnet for future epoch ${epoch}, current epoch: ${currentEpoch}`);
+    }
+    const headEpochCtx = await this.chain.getHeadEpochContext();
+    if (headEpochCtx.getCommitteeCountAtSlot(slot) !== committeesAtSlot) {
+      throw new ApiError(400, `Incorrect committee count ${committeesAtSlot} at slot ${slot}`);
+    }
+    await this.db.activeValidatorCache.add(validatorIndex, epoch);
     if (isAggregator) {
       await this.sync.collectAttestations(slot, committeeIndex);
     }
--- a/packages/lodestar/test/unit/tasks/interopSubnetsJoiningTask.test.ts
+++ b/packages/lodestar/test/unit/tasks/interopSubnetsJoiningTask.test.ts
@@ -15,12 +15,14 @@ import {MetadataController} from "../../../src/network/metadata";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {computeForkDigest} from "@chainsafe/lodestar-beacon-state-transition";
 import {TreeBacked} from "@chainsafe/ssz";
+import {StubbedBeaconDb} from "../../utils/stub";
 
 describe("interopSubnetsJoiningTask", () => {
   const sandbox = sinon.createSandbox();
   let clock: SinonFakeTimers;
   let networkStub: SinonStubbedInstance<INetwork>;
   let gossipStub: SinonStubbedInstance<IGossip>;
+  let dbStub: StubbedBeaconDb;
 
   let chain: IBeaconChain;
   const logger = new WinstonLogger();
@@ -53,9 +55,13 @@ describe("interopSubnetsJoiningTask", () => {
       config,
     });
     networkStub.metadata = new MetadataController({}, {config, chain, logger});
+    dbStub = new StubbedBeaconDb(sinon);
+    // 2 active validators
+    dbStub.activeValidatorCache.values.resolves([0, 5]);
     task = new InteropSubnetsJoiningTask(config, {
       network: networkStub,
       chain,
+      db: dbStub,
       logger,
     });
     await task.start();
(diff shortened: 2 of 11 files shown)


packages/lodestar/src/tasks/tasks/cleanUpTask.ts
```diff
@@ -0,0 +1,49 @@
...
+  private run = async (epoch: Epoch): Promise<void> => {
+    this.logger.info("Run CleanUpTask at epoch ", epoch);
+    this.logger.profile("CleanUpTask");
+    const indexes = await this.db.activeValidatorCache.values();
+    const headState = await this.chain.getHeadState();
+    await Promise.all(
+      indexes
+        .filter((index) => headState.validators[index].exitEpoch <= epoch)
+        .map((index) => this.db.activeValidatorCache.delete(index))
+    );
```
mpetrunic: Aren't you suppose to delete all active validators, else validator might disconnect and we still might track him?

twoeths: so I'll have to track the last active epoch of each validator, not just validator index
+ at each epoch, check if lastActiveEpoch < epoch - 1 then delete it

packages/lodestar/src/tasks/tasks/cleanUpTask.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  public async stop(): Promise<void> {
+    this.chain.clock.unsubscribeFromNewEpoch(this.run);
+  }
+
+  private run = async (epoch: Epoch): Promise<void> => {
+    this.logger.info("Run CleanUpTask at epoch ", epoch);
+    this.logger.profile("CleanUpTask");
+    const indexes = await this.db.activeValidatorCache.values();
```
wemeetagain: If this CleanupTask can be used for other things, we should separate each thing into separate functions.

packages/lodestar/src/tasks/tasks/cleanUpTask.ts
mpetrunic: We just care about active indexes in current epoch I guess? Not sure what's the best way to do it

packages/lodestar/src/tasks/tasks/cleanUpTask.ts
```diff
@@ -0,0 +1,53 @@
...
+export interface ICleanUpModules {
+  chain: IBeaconChain;
+  db: IBeaconDb;
+  logger: ILogger;
+}
+
+/**
+ * Cleanup task to run per epoch or per slot.
+ */
+export class CleanUpTask {
```
mpetrunic: Maybe call it epoch cleanup task?

packages/lodestar/src/tasks/tasks/cleanUpTask.ts
twoeths: I can't think of a better way. `EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION` = 256 so a subscription last from 256 to 512 so it's not a big difference to know active indexes in previous epoch vs current epoch.

packages/lodestar/src/tasks/tasks/cleanUpTask.ts
twoeths: it depends on if we want this to cleanup per epoch only or we may want to cleanup per slot later for other businesses ...

If we leave like this we'll likely have `onNewEpoch, onNewSlot` etc..

If we separate to multiple cleanup tasks then we may have EpochCleanUpTask, SlotCleanUpTask

twoeths: I changed the method name to `onNewEpoch` so that we can clean up per slot or per events later...

wemeetagain:
wait for https://github.com/ethereum/eth2.0-APIs/pull/110

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -195,6 +197,20 @@ export class ValidatorApi implements IValidatorApi {
...
+    const epoch = computeEpochAtSlot(this.config, slot);
+    const currentEpoch = this.chain.clock.currentEpoch;
+    if (slot <= currentSlot) {
+      throw new ApiError(400, `Cannot prepare subnet for invalid slot ${slot}, current slot: ${currentSlot}`);
+    }
+    if (epoch > currentEpoch + 1) {
+      throw new ApiError(400, `Cannot prepare subnet for future epoch ${epoch}, current epoch: ${currentEpoch}`);
+    }
+    const headEpochCtx = await this.chain.getHeadEpochContext();
+    if (headEpochCtx.getCommitteeCountAtSlot(slot) !== committeesAtSlot) {
```
twoeths, line 217: for other clients, the point to have `committeesAtSlot` param is to avoid the node calculating it again (using `processSlots` and maybe `processEpoch`) but for us it's a simple get from EpochContext so I want an extra check here to potentially avoid DOS attack

twoeths:
go with the latest api spec and updated the description accordingly, ready for review

mpetrunic (dismissed):

mpetrunic (dismissed):

dapplion:
Bump

packages/lodestar/test/unit/tasks/interopSubnetsJoiningTask.test.ts
```diff
@@ -7,6 +7,7 @@ import {ChainEvent, IBeaconChain} from "../../../src/chain";
 import {Gossip} from "../../../src/network/gossip/gossip";
 import {InteropSubnetsJoiningTask} from "../../../src/tasks/tasks/interopSubnetsJoiningTask";
 import {WinstonLogger, bytesToInt, intToBytes} from "@chainsafe/lodestar-utils";
+import * as mathUtils from "@chainsafe/lodestar-utils/lib/math";
```
dapplion, line 10: This type of stubs is preventing us from upgrading to Typescript v4. Could you please find a different way to test this without them?

dapplion:
Would it be possible to move all the logic around counting the num of connected validators, the cache, and persisting to db from the `api` dir to the InteropSubnetsJoiningModules instance? It would keep the logic consistent in one single place

packages/lodestar/test/unit/tasks/interopSubnetsJoiningTask.test.ts
twoeths: @dapplion could you send me the breaking change of Typescript v4 regarding this? I notice this patterns happen at a lot of places in our code.

twoeths:
> Would it be possible to move all the logic around counting the num of connected validators, the cache, and persisting to db from the `api` dir to the InteropSubnetsJoiningModules instance? It would keep the logic consistent in one single place

I'm afraid no, the requirement is to track connected validators via api and use it in InteropSubnetsJoiningModules

packages/lodestar/test/unit/tasks/interopSubnetsJoiningTask.test.ts
twoeths: is it ok to do `export * as mathUtils from "./math";` in lodestar-utils and `import {mathUtils} from "@chainsafe/lodestar-utils"` in Typescript v4?

wemeetagain:
bump @dapplion what are your thoughts here wrt how this interacts with the new peer store / range sync PR?

dapplion:
> bump @dapplion what are your thoughts here wrt how this interacts with the new peer store / range sync PR?

It should be very easy to implement now since all peer logic is grouped in the PeerManager. However the approach should be completely re-done.

twoeths:
there is a conflict between AttestationCollector (dealing with short-lived subnets) and InteropSubnetsJoiningTask (long lived random subnets) regarding subscribe/unsubscribe attestation subnet topics. The AttestationCollector itself has a conflict on the subscribe/unsubscribe short-lived subnets too.

as suggested by @dapplion , I'm looking into unifying them into a service like https://github.com/sigp/lighthouse/blob/3a24ca5f14c6e6d6612fd43eca82aa0c1e6aba16/beacon_node/network/src/attestation_service/mod.rs#L114 and handle connected validators there

PR #1454, opened 2020-08-27
fix genesisStateFile flag

### LINKED ISSUE #1188: genesisStateFile not being copied over to beacon.config.json during beacon init

3xtr4t3rr3str14l:
fixes #1188

### DIFF
--- a/packages/lodestar-cli/src/cmds/init/handler.ts
+++ b/packages/lodestar-cli/src/cmds/init/handler.ts
@@ -43,7 +43,7 @@ export async function initHandler(options: IBeaconArgs & IGlobalArgs): Promise<v
     Object.assign(options, deepmerge(options, testnetConfig));
     const genesisFileUrl = getGenesisFileUrl(options.testnet);
     if (genesisFileUrl) {
-      const genesisStateFile = path.join(options.beaconDir, "genesis.ssz");
+      const genesisStateFile = path.join(options.beaconDir, options.genesisStateFile || "genesis.ssz");
       options.genesisStateFile = genesisStateFile;
       await downloadFile(options.genesisStateFile, genesisFileUrl);
       options.eth1.enabled = false;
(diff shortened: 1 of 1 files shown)


packages/lodestar-cli/src/cmds/init/handler.ts
```diff
@@ -43,7 +43,7 @@ export async function initHandler(options: IBeaconArgs & IGlobalArgs): Promise<v
     Object.assign(options, deepmerge(options, testnetConfig));
     const genesisFileUrl = getGenesisFileUrl(options.testnet);
     if (genesisFileUrl) {
-      const genesisStateFile = path.join(options.beaconDir, "genesis.ssz");
+      const genesisStateFile = path.join(options.beaconDir, options.genesisStateFile || "genesis.ssz");
```
wemeetagain: Can you use `joinIfRelative` instead of `path.join`?

wemeetagain (changes requested):

packages/lodestar-cli/src/cmds/init/handler.ts
3xtr4t3rr3str14l: done

dapplion (approved):

mpetrunic (approved):

wemeetagain (approved):

Issue #1451, opened 2020-08-27
Start from Weak Subjectivity State


wemeetagain:
**Describe the solution you'd like**
Lodestar should be able to boot from a trusted BeaconState that is not genesis. See Weak Subjectivity doc for motivation.

**Additional context**
https://notes.ethereum.org/@adiasg/weak-subjectvity-eth2

wemeetagain:
This should be ready for mainnet launch.

twoeths:
So I suppose this helps save a lot of time for Initial Sync.

Does this include calculating Weak Subjectivity Checkpoint States and export it somewhere through Rest api for example?

wemeetagain:
no, i think storage/retrieval should be tackled separately

wemeetagain:
#1455 updates chain initialization, tweaks `chain.initForkChoice` and others, will be helpful for this

PR #1455, opened 2020-08-27
Add chain event emitter

wemeetagain:
Steps towards https://hackmd.io/jS7QMXR2Q4Os-JS-ZFHBAQ

- ~Remove `StatefulDagLMDGhost` - we aren't planning on using this, maintenance burden~ #1470 
- Add `ChainEventEmitter` and use everywhere. All chain events now are attached to `chain.emitter`.
- Refactor `LocalClock` and `ArrayDagLMDGhost` to remove start/stop
  - `LocalClock` uses `AbortSignal` to close
  - `ArrayDagLMDGhost` no longer calls its own `onTick` method, instead the chain calls it
- slight chain module rework:
  - `forkChoice`, `attestationProcessor`, `blockProcessor` now constructed in `chain#start`
  - `initializeBeaconChain` now doesn't initialize the fork choice
  - `restoreHeadState` now initializes the fork choice (no matter what)
- use `chain.clock.currentSlot` where immediately possible instead of using `getCurrentSlot`

Need help diagnosing issues with e2e tests (edit: and hanging unit tests :( )


packages/lodestar/src/api/impl/beacon/beacon.ts
```diff
@@ -92,9 +92,9 @@ export class BeaconApi implements IBeaconApi {
 
   public getBlockStream(): LodestarEventIterator<SignedBeaconBlock> {
     return new LodestarEventIterator<SignedBeaconBlock>(({push}) => {
-      this.chain.on("processedBlock", push);
+      this.chain.emitter.on("processedBlock", push);
```
mpetrunic: since you are already refactoring chain emitting. Maybe rename event to just "block" and have another "head" which should help us implement standardized event stream api

packages/lodestar/src/chain/attestation/pool.ts
```diff
@@ -34,10 +34,10 @@ export class AttestationProcessor implements IAttestationProcessor {
 
   public async start(): Promise<void> {
     //checks if attestation is waiting on specific slot
-    this.chain.clock.onNewSlot(this.onNewSlot);
+    this.chain.emitter.on("clock:slot", this.onNewSlot);
   }
   public async stop(): Promise<void> {
-    this.chain.clock.unsubscribeFromNewSlot(this.onNewSlot);
+    this.chain.emitter.on("clock:slot", this.onNewSlot);
```
mpetrunic:
```suggestion
    this.chain.emitter.off("clock:slot", this.onNewSlot);
```

packages/lodestar/src/chain/attestation/processor.ts
```diff
@@ -64,7 +64,7 @@ export async function processAttestation(
       );
     }
     logger.debug(`Attestation ${toHexString(attestationHash)} passed to fork choice`);
-    chain.emit("processedAttestation", attestation);
+    chain.emitter.emit("processedAttestation", attestation);
```
mpetrunic: maybe processedAttestation -> attestation

packages/lodestar/src/chain/chain.ts
```diff
@@ -176,27 +168,46 @@ export class BeaconChain extends (EventEmitter as {new (): ChainEventEmitter}) i
...
+    });
+    this.blockProcessor = new BlockProcessor(
+      this.config,
+      this.logger,
+      this.db,
+      this.forkChoice,
+      this.metrics,
+      this.emitter,
+      this.attestationProcessor
+    );
```
mpetrunic, line 196: Why constructing in start method?

packages/lodestar/src/chain/chain.ts
```diff
@@ -176,27 +168,46 @@ export class BeaconChain extends (EventEmitter as {new (): ChainEventEmitter}) i
...
     await this.db.stateCache.add({state, epochCtx});
     this.logger.info("Chain started, waiting blocks and attestations");
-    this.clock = new LocalClock(this.config, state.genesisTime);
-    await this.clock.start();
-    await this.forkChoice.start(state.genesisTime, this.clock);
+    this.clock = new LocalClock({
+      config: this.config,
+      emitter: this.emitter,
+      genesisTime: state.genesisTime,
+      signal: this.abortController.signal,
```
mpetrunic, line 176: why not create separate abort controller inside clock?

packages/lodestar/src/chain/chain.ts
```diff
@@ -226,24 +237,9 @@ export class BeaconChain extends (EventEmitter as {new (): ChainEventEmitter}) i
...
-      slot: genesisBlock.slot,
-      blockRoot: blockRoot,
-      stateRoot: stateRoot,
-      parentRoot: Buffer.alloc(32),
-      justifiedCheckpoint: justifiedFinalizedCheckpoint,
-      finalizedCheckpoint: justifiedFinalizedCheckpoint,
-    });
-    const epochCtx = new EpochContext(this.config);
-    epochCtx.loadState(genesisState);
-    await this.db.stateCache.add({state: genesisState, epochCtx});
```
mpetrunic, line 241: where did this part of code go?

packages/lodestar/src/chain/chain.ts
```diff
@@ -281,11 +277,11 @@ export class BeaconChain extends (EventEmitter as {new (): ChainEventEmitter}) i
...
         if (this.config.types.Root.equals(root, blockRoot)) {
-          this.removeListener("processedBlock", listener);
+          this.emitter.removeListener("processedBlock", listener);
           resolve();
         }
       };
-      this.on("processedBlock", listener);
+      this.emitter.on("processedBlock", listener);
     });
   }
```
mpetrunic, line 284: I think this method should also receive abort signal or timeout?
And on chain stop should cancel

packages/lodestar/src/chain/clock/LocalClock.ts
```diff
@@ -0,0 +1,69 @@
...
+
+import {IBeaconClock} from "./interface";
+
+/**
+ * A local clock, the clock time is assumed to be trusted
+ */
+export class LocalClock implements IBeaconClock {
+  private readonly config: IBeaconConfig;
+  private readonly genesisTime: number;
+  private timeoutId: NodeJS.Timeout;
```
mpetrunic, line 15:
```suggestion
  private timeoutId?: NodeJS.Timeout;
```

packages/lodestar/src/chain/clock/LocalClock.ts
```diff
@@ -0,0 +1,69 @@
...
+  public get currentSlot(): Slot {
+    return this._currentSlot;
+  }
+
+  public get currentEpoch(): Epoch {
+    return computeEpochAtSlot(this.config, this._currentSlot);
+  }
+
+  private abort = (): void => {
+    clearTimeout(this.timeoutId);
```
mpetrunic, line 51: if(this.timeoutId) {
}

packages/lodestar/src/chain/clock/LocalClock.ts
```diff
@@ -0,0 +1,69 @@
...
+    emitter: ChainEventEmitter;
+    signal: AbortSignal;
+  }) {
+    this.config = config;
+    this.genesisTime = genesisTime;
+    this.timeoutId = setTimeout(this.onNextSlot, this.msUntilNextSlot());
+    this.signal = signal;
+    this.emitter = emitter;
+    this._currentSlot = getCurrentSlot(this.config, this.genesisTime);
+    signal.addEventListener("abort", this.abort);
```
mpetrunic: https://github.com/mysticatea/event-target-shim/blob/master/src/event-target.mjs#L202

There is this param option, where you can set `once: true` to avoid removing event listener.
cc @dapplion since he asked me this ^^

packages/lodestar/src/chain/emitter.ts
```diff
@@ -0,0 +1,31 @@
...
+import {BlockSummary} from "./forkChoice";
+
+export interface IChainEvents {
+  // old, to be deprecated
+  unknownBlockRoot: (root: Root) => void;
+  processedBlock: (signedBlock: SignedBeaconBlock) => void;
+  processedCheckpoint: (checkpoint: Checkpoint) => void;
+  processedAttestation: (attestation: Attestation) => void;
+  justifiedCheckpoint: (checkpoint: Checkpoint) => void;
+  finalizedCheckpoint: (checkpoint: Checkpoint) => void;
```
mpetrunic: block, head, checkpoint, attestation, justified, finalized for events?

packages/lodestar/test/unit/chain/attestation/pool.test.ts
```diff
@@ -24,6 +24,9 @@ describe("attestation pool", function () {
 
   beforeEach(function () {
     chain = sinon.createStubInstance(BeaconChain);
+    chain.clock = sinon.createStubInstance(LocalClock);
+    (chain.clock as any).config = config;
```
mpetrunic, line 28: why do you need to set config?

packages/lodestar/test/unit/chain/attestation/pool.test.ts
```diff
@@ -35,17 +38,19 @@ describe("attestation pool", function () {
 
   it("should not receive old attestation", async function () {
     const attestation = generateAttestation({data: {target: {epoch: 0}}});
-    chain.getGenesisTime.returns(
+    (chain.clock as any)._currentSlot = getCurrentSlot(
```
mpetrunic: :vomiting_face:  :rofl: 

stub getter function.
I think it's something like `sinon.stub(chain.clock, "currentSlot").get(() => 15)`

mpetrunic (changes requested):

packages/lodestar/src/chain/clock/LocalClock.ts
dapplion: Nice find! Thanks

mpetrunic: Keep in mind that if signal don't get triggered, we still need to remove manually

packages/lodestar/src/chain/chain.ts
wemeetagain: fork choice, clock, and processors now require an anchor state to be constructed, allows us to make them all simpler.

packages/lodestar/src/chain/chain.ts
wemeetagain: We don't need this since we're doing the same in chain.start

packages/lodestar/test/unit/chain/attestation/pool.test.ts
wemeetagain: :+1:  will try that

packages/lodestar/src/chain/clock/LocalClock.ts
wemeetagain: We don't need that anymore with the clock refactoring, `this.timeout` is always set.

packages/lodestar/src/chain/chain.ts
wemeetagain: :shrug: It is nice to be able to control from an outside source (that may not even be in scope here).

If we want to make the `IBeaconClock` interface more flexible, we can include a public `abort` method, make `signal` an optional input to `LocalClock`.

packages/lodestar/src/api/impl/beacon/beacon.ts
wemeetagain: :+1: I was trying not to do too much in this PR, but I'll go ahead and rename these

packages/lodestar/src/chain/chain.ts
mpetrunic: I liked that we had start/stop standard everywhere but ok

packages/lodestar/src/chain/clock/LocalClock.ts
mpetrunic: what about calling this method twice in a row?

packages/lodestar/test/unit/chain/attestation/pool.test.ts
wemeetagain: for some reason this doesn't work. It sets the stub getter appropriately, but isn't set when called thru another module :thinking:

wemeetagain: the unit tests also seem to be hanging bc in `test/unit/sync/initial/fast.test.ts`, the stubbed `networkStub.getPeers` isn't being used, and its causing an unhandled promise rejection.

dapplion:
> Need help diagnosing issues with e2e tests (edit: and hanging unit tests :( )

You can add the flag `--exit` to mocha to kill its process when tests are done

https://mochajs.org/#-exit

mpetrunic:
Fixed hanging in unit tests and e2e tests (mockchain wasn't stopped). There is a bug somewhere where current slot is behind even current head slot:
image

wemeetagain:
> There is a bug somewhere where current slot is behind even current head slot

Fixed the bug. ~e2e test still not justifying/finalizing, only error I see is `error: Failed to obtain attester duty for epoch 7 Error: Cannot get committee assignment for epoch more than 1 ahead: 7 <= 6`~

~Edit: e2e test still not justifying/finalizing, no visible errors~
Edit: tests passing locally yay

packages/lodestar/src/chain/chain.ts
```diff
@@ -345,54 +345,37 @@ export class BeaconChain extends (EventEmitter as {new (): ChainEventEmitter}) i
...
+   * This state is set as the finalized state.
+   */
+  private async initForkChoice(anchorState: TreeBacked<BeaconState>): Promise<void> {
+    let blockRoot;
+    if (anchorState.latestBlockHeader.slot === GENESIS_SLOT) {
+      const block = getEmptyBlock();
+      block.stateRoot = this.config.types.BeaconState.hashTreeRoot(anchorState);
+      blockRoot = this.config.types.BeaconBlock.hashTreeRoot(block);
+    } else {
+      blockRoot = this.config.types.BeaconBlockHeader.hashTreeRoot(anchorState.latestBlockHeader);
```
twoeths: what I read is each time we process block we only assign temporary block header with state root like `state.latestBlockHeader = getTemporaryBlockHeader(config, block);`
so only when we process the next block, we know how to calculate previous block root and store it to `blockRoots` in state (using `processSlot()`)

so I suppose here we have to set stateRoot to latestBlockHeader in this case before calculating blockRoot, can you double check?

wemeetagain: Ah, you're right, good catch!
Yeah, the work here is attempting to remove the need to pass _both_ a state and a block to initialize the fork choice. Ideally, we only need _just_ a state to boot the system. This should help in #1451 so we won't need both the weak subjectivity state and latest block.

packages/lodestar/src/chain/chain.ts
```diff
@@ -345,54 +346,39 @@ export class BeaconChain extends (EventEmitter as {new (): ChainEventEmitter}) i
...
+   */
+  private async initForkChoice(anchorState: TreeBacked<BeaconState>): Promise<void> {
+    let blockRoot;
+    if (anchorState.latestBlockHeader.slot === GENESIS_SLOT) {
+      const block = getEmptyBlock();
+      block.stateRoot = this.config.types.BeaconState.hashTreeRoot(anchorState);
+      blockRoot = this.config.types.BeaconBlock.hashTreeRoot(block);
+    } else {
+      const blockHeader = anchorState.latestBlockHeader as BeaconBlockHeader;
+      blockHeader.stateRoot = this.config.types.BeaconState.hashTreeRoot(anchorState);
```
twoeths, line 350: maybe `anchorState.hashTreeRoot()`?

If I use `this.config.types.BeaconState.hashTreeRoot(anchorState)`, I got a wrong state root, do you know why?

wemeetagain: the two should be equivalent. If you can get a reproducible testcase theres a bug!

twoeths: thanks for figuring that out, that's my environment issue.
probably we should clone `latestBlockHeader` before assigning new `stateRoot` so that the original `anchorState` is unchanged? or just clone the original anchorState.

packages/lodestar/src/sync/sync.ts
```diff
@@ -128,7 +128,7 @@ export class BeaconSync implements IBeaconSync {
   }
 
   public isSynced(): boolean {
-    return this.mode === SyncMode.SYNCED;
+    return this.mode === SyncMode.SYNCED || this.mode === SyncMode.REGULAR_SYNCING;
```
mpetrunic: Oh noes, this is wrong!! Regular sync could be way of head

mpetrunic (dismissed):

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -261,11 +261,16 @@ export class ValidatorApi implements IValidatorApi {
     if (!this.sync.isSynced()) {
       let syncStatus;
       try {
-        syncStatus = this.config.types.SyncingStatus.toJson(await this.sync.getSyncStatus());
+        syncStatus = await this.sync.getSyncStatus();
```
dapplion, line 264: Just an opinionated comment, isn't
```ts
const syncStatus = await this.sync.getSyncStatus().catch(e => {
  throw new ApiError(503, "Node is stopped");
})
```
cleaner? No need to make `syncStatus` mutable

wemeetagain: yeah, this logic is still likely wrong. Lets fix it in another PR.

wemeetagain: alternatively
```typescript
try {
  const syncStatus . await this.sync.getSyncStatus();
  ...
} catch (e) {
  throw new ApiError(503, "Node is stopped");
}
```

dapplion: Also, you should log the error so we are not helpless with the non-specific "Node is stopped" error

wemeetagain: https://github.com/ChainSafe/lodestar/issues/1492

twoeths (approved):

PR #1474, opened 2020-08-29
Rename chain.getBlockAtSlot to chain.getCanonicalBlockAtSlot

### LINKED ISSUE #1018: Rename chain getBlockAtSlot to getCanonicalBlockAtSlot
Currently, we have getBlockAtSlot() in chain but it only return 1 block on the same chain to head or null.
In reality, for single slot we may have multiple blocks staying on multiple chains of fork choice. We may need to implement getBlocksAtSlot() to return all blocks of that slot and check if we want/should call getBlocksAtSlot() instead of getBlockAtSlot().

mpetrunic:
resolves #1018


packages/lodestar/src/chain/chain.ts
```diff
@@ -124,7 +124,7 @@ export class BeaconChain extends (EventEmitter as {new (): ChainEventEmitter}) i
     return this.db.block.get(this.forkChoice.headBlockRoot());
   }
 
-  public async getBlockAtSlot(slot: Slot): Promise<SignedBeaconBlock | null> {
+  public async getCanonicalBlockAtSlot(slot: Slot): Promise<SignedBeaconBlock | null> {
```
dapplion, line 127: Could you add a JSDoc description to the function to clarify to future maintainers the meaning of CanonicalBlockAtSlot?

mpetrunic: hmm should be self explanatory, canonical is term for block in current head chain. What would you like me to write there?

dapplion: Explain that there can be more than 1 block per slot due to fork choice forks, and this function returns the block from the fork that we consider head. Basically Tuyen explained it great in his issue to be fixed with this PR.

dapplion (approved):

wemeetagain (approved):

PR #1477, opened 2020-08-30
Use named arguments for beacon state transition options

### LINKED ISSUE #1421: Use a more descriptive way to communicate different opt modes of the state transition fn
It will provide better readability than the current approach
```ts
fn(true, false, true)
```

Inspired by Lighthouse's 
```rust
/// The strategy to be used when validating the block's signatures.
pub enum BlockSignatureStrategy {
    /// Do not validate any signature. Use with caution.
    NoVerification,
    /// Validate each signature individually, as its object is being processed.
    VerifyIndividual,
    /// Verify all signatures in bulk at the beginning of block processing.
    VerifyBulk,
}
```
https://github.com/sigp/lighthouse/blob/b73c497be2bf70715327654eb3bf5ff4df9aaab8/consensus/state_processing/src/per_block_processing.rs#L41

dapplion:
Fixes https://github.com/ChainSafe/lodestar/issues/1421

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/fast/index.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/index.ts
@@ -10,10 +10,9 @@ export {IStateContext, EpochContext};
 export function fastStateTransition(
   {state, epochCtx}: IStateContext,
   signedBlock: SignedBeaconBlock,
-  verifyStateRoot = true,
-  verifyProposer = true,
-  verifySignatures = true
+  options?: {verifyStateRoot: boolean; verifyProposer: boolean; verifySignatures: boolean}
 ): IStateContext {
+  const {verifyStateRoot = false, verifyProposer = true, verifySignatures = true} = options || {};
   const types = epochCtx.config.types;
 
   const block = signedBlock.message;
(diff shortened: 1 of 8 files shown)


packages/lodestar-beacon-state-transition/src/fast/index.ts
```diff
@@ -10,10 +10,9 @@ export {IStateContext, EpochContext};
 export function fastStateTransition(
   {state, epochCtx}: IStateContext,
   signedBlock: SignedBeaconBlock,
-  verifyStateRoot = true,
-  verifyProposer = true,
-  verifySignatures = true
+  options?: {verifyStateRoot?: boolean; verifyProposer?: boolean; verifySignatures?: boolean}
 ): IStateContext {
+  const {verifyStateRoot = false, verifyProposer = true, verifySignatures = true} = options || {};
```
mpetrunic: I think we should have everything to true by default

dapplion: I respected the original values and logic. Should I change it anyway?

mpetrunic: DO we every use default values anywhere?

dapplion: In one test in packages/lodestar

mpetrunic: I think it's fine to change defaults then and just update there

wemeetagain (approved):
:tada: sometimes its the little things

mpetrunic (approved):

PR #1478, opened 2020-08-30
Remove unused chain options

### LINKED ISSUE #1228: Change `chain.name` option to `spec`
Use a more intuitive name to choose between the default eth2 spec type. Also, align the API to other major Eth2.0 clients

- [x] Remove `chain.name` option from `BeaconChain` since it's not used: https://github.com/ChainSafe/lodestar/blob/9fceb034282609f11efb0ce8acb74a0cb2107e25/packages/lodestar/src/chain/chain.ts#L80
- [ ] Rename the library API and the CLI API to be on par and both use a `spec` name option
- [ ] Consider deprecating `IChainOptions` entirely since none of its options seem to be used. https://github.com/ChainSafe/lodestar/blob/9fceb034282609f11efb0ce8acb74a0cb2107e25/packages/lodestar/src/chain/options.ts#L1-L4

dapplion:
Fixes https://github.com/ChainSafe/lodestar/issues/1228
I've left the IChainOpts there empty. It's more consistent with other modules and would require less changes if we need to add options in the future. Should I remove it?

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -61,7 +61,6 @@ export interface IBlockProcessJob {
 }
 
 export class BeaconChain extends (EventEmitter as {new (): ChainEventEmitter}) implements IBeaconChain {
-  public readonly chain: string;
   public forkChoice: ILMDGHOST;
   public chainId: Uint16;
   public networkId: Uint64;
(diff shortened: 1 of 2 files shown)


mpetrunic (approved):

wemeetagain (approved):
I think this looks good. Like you said, will be easier to add options later if the need arises

PR #1479, opened 2020-08-30
Remove unsafe uses of setTimeout

### LINKED ISSUE #1184: Not type-safe use of setTimeout
Using `setTimeout` in the 3+ arguments form:
```js
setTimeout(handler: TimerHandler, timeout?: number, ...arguments: any[])
```
is not type-safe. If we ever change the function signature of the handler Typescript will not alert that these setTimeout calls are broken.

I've found some uses of this pattern throughout the codebase.

https://github.com/ChainSafe/lodestar/blob/e71f29432f0a67b5ad467c88c4c6dcaab96fb2d5/packages/lodestar-validator/src/services/attestation.ts#L135-L139

https://github.com/ChainSafe/lodestar/blob/e71f29432f0a67b5ad467c88c4c6dcaab96fb2d5/packages/lodestar/src/sync/utils/attestation-collector.ts#L64-L68

https://github.com/ChainSafe/lodestar/blob/e71f29432f0a67b5ad467c88c4c6dcaab96fb2d5/packages/lodestar/src/tasks/tasks/interopSubnetsJoiningTask.ts#L78

https://github.com/ChainSafe/lodestar/blob/e71f29432f0a67b5ad467c88c4c6dcaab96fb2d5/packages/lodestar/src/tasks/tasks/interopSubnetsJoiningTask.ts#L136-L144

Find an alternative approach that ensures argument type-safety. Aren't class methods bind automatically?

dapplion:
Fixes https://github.com/ChainSafe/lodestar/issues/1184

No floating promises rule complains about the handlers inside setTimeout. I've added an eslint ignore since that's not the focus of this PR. But if you have a definitive answer about how those should be handled please leave a comment.

### DIFF
--- a/packages/lodestar/src/sync/utils/attestation-collector.ts
+++ b/packages/lodestar/src/sync/utils/attestation-collector.ts
@@ -65,11 +65,9 @@ export class AttestationCollector implements IService {
       const subnet = computeSubnetForSlot(this.config, headState, slot, committeeIndex);
       this.network.gossip.subscribeToAttestationSubnet(forkDigest, subnet, this.handleCommitteeAttestation);
       this.timers.push(
-        (setTimeout(
-          this.unsubscribeSubnet,
-          this.config.params.SECONDS_PER_SLOT * 1000,
-          subnet
-        ) as unknown) as NodeJS.Timeout
+        setTimeout(() => {
+          this.unsubscribeSubnet(subnet);
+        }, this.config.params.SECONDS_PER_SLOT * 1000)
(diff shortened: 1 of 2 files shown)


mpetrunic:
Cant you just pass async function to setTimeout?

dapplion:
> Cant you just pass async function to setTimeout?

They will just be called and caused an unhandled rejection on error. This is bad but this was the original behaviour

wemeetagain (approved):

PR #1484, opened 2020-08-30
Migration of reputation store to metabook

### LINKED ISSUE #1081: Libp2p MetadataBook
Currently we implement ReputationStore to store data of a node. From libp2p 0.28 it introduced MetadataBook with same functionalities so we should switch to using it instead.

mpetrunic:
We don't actually need status field persisted, although method is synchronous so I don't think we are waiting for it to be persisted.
If it becomes performance issue, we can transparently switch to pure map for status storage.

This is base for BlockProvider score tracking. 

Let's wait for #1428 

resolves #1081


packages/lodestar/src/network/peers/metastore.ts
```diff
@@ -0,0 +1,73 @@
...
+  public setScore(peer: PeerId, score: number): void {
+    return this.set(peer, MetadataKey.SCORE, this.config.types.Number64, score);
+  }
+
+  public setStatus(peer: PeerId, status: Status): void {
+    return this.set(peer, MetadataKey.STATUS, this.config.types.Status, status);
+  }
+
+  private set<T>(peer: PeerId, key: MetadataKey, type: BasicType<T> | ContainerType<T>, value: T): void {
+    this.metabook.set(peer, key, Buffer.from(type.serialize(value)));
```
twoeths: does `MetadataBook` in libp2p 0.28 support Uint8Array?

libp2p 0.29 switches from Buffer to UInt8Array to support browser later so probably we want to use Uint8Array here if possible

mpetrunic: Buffer is uint8array and deserialize works with uint8array so all of this will work as expected

wemeetagain (approved):

PR #1487, opened 2020-08-31
Faster attestation verification

mpetrunic:
we now utilize decompressed public keys, and aggregation is done without extra wasm calls.


wemeetagain (approved):

dapplion:
Would love to see a benchmark of this!

PR #1490, opened 2020-09-01
Explanation of generating keys

darcys22:
Steps followed to generate the validator keystores

### DIFF
--- a/docs/usage/medalla.md
+++ b/docs/usage/medalla.md
@@ -76,4 +76,40 @@ You should see confirmation that modules have started.
 
 ## Submit your deposit to Goerli
 
-_TBD_
+Navigate to https://github.com/ethereum/eth2.0-deposit-cli/releases
+
+Pick the most recent release to download
+
+```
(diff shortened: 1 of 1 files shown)


dapplion:
Thanks for your contribution! Above all, I think we should forward users to the eth2-deposit-cli guide so we don't have to actively maintain this doc page. @wemeetagain what do you think?

wemeetagain:
Thanks for getting the ball rolling @darcys22 

I think we would benefit from some kind of documentation here. Definitely with focus on the lodestar side, rather than the deposit cli specifics.

One thing to note, other clients have help documentation baked into the eth2.0-deposit launchpad. We don't (although I don't feel comfortable adding lodestar help here until lodestar is more stable): https://github.com/ethereum/eth2.0-deposit/tree/dev/src/pages/ValidatorClients

Perhaps whatever documentation we have here could be used in preparation for adding to the eth2.0-deposit repo.

dapplion:
Keys should always be generated with the eth2.0-deposit-cli tool https://github.com/ethereum/eth2.0-deposit-cli/
We may keep the wallet and key generation functionality but since users should not use it we may not include docs for it.

PR #1494, opened 2020-09-02
Simple state cache pruning

wemeetagain:
Tested against medalla, memory stays <800MB

### DIFF
--- a/packages/lodestar/src/db/api/beacon/stateContextCache.ts
+++ b/packages/lodestar/src/db/api/beacon/stateContextCache.ts
@@ -42,6 +42,25 @@ export class StateContextCache {
     this.cache = {};
   }
 
+  public get size(): number {
+    return Object.keys(this.cache).length;
+  }
+
+  /**
+   * TODO make this more robust.
+   * Without more thought, this currently breaks our assumptions about recent state availablity
+   */
+  public prune(): void {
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/db/api/beacon/stateContextCache.ts
```diff
@@ -42,6 +42,25 @@ export class StateContextCache {
...
+  /**
+   * TODO make this more robust.
+   * Without more thought, this currently breaks our assumptions about recent state availablity
+   */
+  public prune(): void {
+    const MAX_STATES = 128;
+    const keys = Object.keys(this.cache);
+    if (keys.length > MAX_STATES) {
+      // object keys are stored in insertion order, delete keys starting from the front (but keeping the first)
+      keys.slice(1, MAX_STATES - keys.length).forEach((key) => {
```
twoeths, line 58: do we keep the 1st item in order to restore other deleted states/epoch contexts from there?

wemeetagain: Yeah, its not really necessary or used right now but thats the idea. The first one is likely the oldest, so we could _probably_ recreate all future states from that one.

twoeths (approved):

PR #1495, opened 2020-09-02
Remove recursion from fork choice

wemeetagain:
Simple rewrite of `propagateWeightChange` method.
Not intended to change any behavior, just rewrite to not use recursion.


twoeths (approved):
lgtm, this is simple change with great effect 👍

PR #1496, opened 2020-09-02
Start from a weak subjectivity state

### LINKED ISSUE #1451: Start from Weak Subjectivity State

twoeths:
resolves #1451 

+ To try this branch, unzip the attached and issue something like `./bin/lodestar  beacon --testnet medalla --weak-subjectivity-state-file "/Users/tuyennguyen/Downloads/state181216" --root-dir .medalla.weak`
[state181216.zip](https://github.com/ChainSafe/lodestar/files/5243029/state181216.zip)

### DIFF
--- a/packages/lodestar-cli/src/testnets/index.ts
+++ b/packages/lodestar-cli/src/testnets/index.ts
@@ -58,6 +58,18 @@ function getBootnodesFileUrl(testnet: TestnetName): string {
 }
 
 /**
+ * Fet a remote file from either a local file part or url
+ */
+export async function getRemoteFile(filepath: string, urlOrPath: string): Promise<void> {
+  if (urlOrPath.startsWith("http")) {
+    await downloadFile(filepath, urlOrPath);
+  } else {
+    fs.mkdirSync(path.parse(filepath).dir, {recursive: true});
+    await fs.promises.copyFile(urlOrPath, filepath);
+  }
+}
+
+/**
  * Downloads a genesis file per testnet if it does not exist
  */
 export async function downloadFile(filepath: string, url: string): Promise<void> {
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -255,6 +255,21 @@ export class BeaconChain implements IBeaconChain {
     this.logger.info("Beacon chain initialized");
   }
 
+  /**
+   * Initialize our chain from a weak subjectivity state.
+   * @param state a weak subjectivity state
+   */
+  public async initializeWeakSubjectivityState(weakSubjectivityState: TreeBacked<BeaconState>): Promise<void> {
+    // don't want to initialize if already run beacon node
+    const lastKnownState = await this.db.stateArchive.lastValue();
+    if (lastKnownState) {
+      this.logger.info(`Found finalized state at slot ${lastKnownState.slot}, starting chain from there`);
+      return;
+    }
+    await this.db.stateArchive.add(weakSubjectivityState);
+    this.logger.info("Beacon chain initialized with weak subjectivity state at slot", weakSubjectivityState.slot);
+  }
+
   public async getENRForkID(): Promise<ENRForkID> {
     const state = await this.getHeadState();
     const currentVersion = state.fork.currentVersion;
--- a/packages/lodestar/src/sync/initial/fast/index.ts
+++ b/packages/lodestar/src/sync/initial/fast/index.ts
@@ -13,7 +13,7 @@ import {Checkpoint, SignedBeaconBlock, Slot, Status} from "@chainsafe/lodestar-t
 import pushable, {Pushable} from "it-pushable";
 import {computeEpochAtSlot, computeStartSlotAtEpoch} from "@chainsafe/lodestar-beacon-state-transition";
 import pipe from "it-pipe";
-import {ISlotRange} from "../../interface";
+import {ISlotRange, ISyncCheckpoint} from "../../interface";
 import {fetchBlockChunks, getCommonFinalizedCheckpoint, processSyncBlocks} from "../../utils";
 import {GENESIS_EPOCH} from "../../../constants";
 import {ISyncStats, SyncStats} from "../../stats";
@@ -45,7 +45,7 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
   /**
    * The last processed block
    */
-  private lastProcessedBlock!: SignedBeaconBlock;
+  private lastProcessedBlock!: ISyncCheckpoint;
 
   public constructor(opts: ISyncOptions, {config, chain, network, logger, db, stats}: IInitialSyncModules) {
     super();
@@ -67,8 +67,8 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
     this.targetCheckpoint = getCommonFinalizedCheckpoint(this.config, this.getPeerStatuses());
     // head may not be on finalized chain so we start from finalized block
     // there are unfinalized blocks in db so we reprocess all of them
-    this.lastProcessedBlock = (await this.db.blockArchive.lastValue())!;
-    this.blockImportTarget = this.lastProcessedBlock.message.slot;
+    this.lastProcessedBlock = this.chain.forkChoice.getFinalizedBlock();
+    this.blockImportTarget = this.lastProcessedBlock.slot;
     if (
       !this.targetCheckpoint ||
       this.targetCheckpoint.epoch == GENESIS_EPOCH ||
@@ -140,7 +140,7 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
           );
           logger.verbose("last processed slot=" + lastSlot + ` range=${JSON.stringify(slotRange)}`);
           if (typeof lastSlot === "number") {
-            if (lastSlot === getLastProcessedBlock().message.slot) {
+            if (lastSlot === getLastProcessedBlock().slot) {
               // failed at start of range
               logger.warn(`Failed to process range, retrying, lastSlot=${lastSlot} range=${JSON.stringify(slotRange)}`);
               setBlockImportTarget(lastSlot);
@@ -163,7 +163,10 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
 
   private checkSyncProgress = async (processedBlock: SignedBeaconBlock): Promise<void> => {
     if (processedBlock.message.slot === this.blockImportTarget) {
-      this.lastProcessedBlock = processedBlock;
+      this.lastProcessedBlock = {
+        slot: processedBlock.message.slot,
+        blockRoot: this.config.types.BeaconBlock.hashTreeRoot(processedBlock.message),
+      };
       this.setBlockImportTarget();
     }
   };
@@ -212,7 +215,7 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
   /**
    * Make sure we get up-to-date lastProcessedBlock from sync().
    */
-  private getLastProcessedBlock = (): SignedBeaconBlock => {
+  private getLastProcessedBlock = (): ISyncCheckpoint => {
     return this.lastProcessedBlock;
   };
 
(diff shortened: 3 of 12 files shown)


packages/lodestar/src/chain/chain.ts
```diff
@@ -362,6 +395,30 @@ export class BeaconChain implements IBeaconChain {
...
+   * This state is set as the finalized state.
+   */
+  private async initForkChoiceFromWeakSubjectivityState(anchorState: TreeBacked<BeaconState>): Promise<void> {
+    const blockHeader = anchorState.latestBlockHeader.clone() as BeaconBlockHeader;
+    blockHeader.stateRoot = anchorState.hashTreeRoot();
+    const blockRoot = this.config.types.BeaconBlockHeader.hashTreeRoot(blockHeader);
+    const epoch = computeEpochAtSlot(this.config, anchorState.slot);
+    const blockCheckpoint = {
+      root: blockRoot,
+      // TODO: is it possible to be `epoch + n` where n >= 2?
```
twoeths: The epoch of checkpoint is not always `computeEpochAtSlot(this.config, anchorState.slot)` due to this
https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar-beacon-state-transition/src/epoch/justification.ts#L54

In my case, I started with state of slot 671 and if I use `computeEpochAtSlot(this.config, anchorState.slot)` I'll receive a lot of goodbye requests with reason = 2.

This change helps me got through that but not sure how can we calculate the epoch of checkpoint given a state/BeaconBlockHeader @wemeetagain @mpetrunic

packages/lodestar/src/chain/blocks/validate.ts
```diff
@@ -25,7 +25,7 @@ export function validateBlock(
         if (
           finalizedCheckpoint &&
           finalizedCheckpoint.epoch > 0 &&
-          computeEpochAtSlot(config, job.signedBlock.message.slot) <= finalizedCheckpoint.epoch
+          computeEpochAtSlot(config, job.signedBlock.message.slot) < finalizedCheckpoint.epoch
```
twoeths: this is the same concern I have below. When I start from finalized state of slot 671 (which is epoch 20), the checkpoint is like:
```
{root: hashTreeRoot(block 671), epoch: 21}
```
next slots are 675+ which is epoch 21

packages/lodestar-cli/src/testnets/index.ts
```diff
@@ -63,7 +63,11 @@ function getBootnodesFileUrl(testnet: TestnetName): string {
 export async function downloadFile(filepath: string, url: string): Promise<void> {
   if (!fs.existsSync(filepath)) {
     fs.mkdirSync(path.parse(filepath).dir, {recursive: true});
-    await promisify(stream.pipeline)(got.stream(url), fs.createWriteStream(filepath));
+    if (url.startsWith("http")) {
+      await promisify(stream.pipeline)(got.stream(url), fs.createWriteStream(filepath));
+    } else {
+      await fs.promises.copyFile(url, filepath);
```
mpetrunic: this doesn't belong here, downloadfile should only download from internet, not handling local file copy

packages/lodestar/src/chain/blocks/validate.ts
mpetrunic: I don't think you should change this line, as it opens us to processing finalized blocks

packages/lodestar/src/chain/chain.ts
mpetrunic: What happens if you set checkpoints to:
```
{
   root: getBlockRoot(config, anchorState, epoch),
   epoch
}
```

Maybe check how others temas implemented this?

packages/lodestar/src/sync/initial/fast/index.ts
```diff
@@ -72,8 +73,10 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
     this.targetCheckpoint = getCommonFinalizedCheckpoint(this.config, this.getPeerReputations());
     // head may not be on finalized chain so we start from finalized block
     // there are unfinalized blocks in db so we reprocess all of them
-    this.lastProcessedBlock = (await this.db.blockArchive.lastValue())!;
-    this.blockImportTarget = this.lastProcessedBlock.message.slot;
+    const lastKnownState = (await this.db.stateArchive.lastValue())?.clone() as TreeBacked<BeaconState>;
```
mpetrunic: I don't think you need clone here db returns new object

packages/lodestar/src/sync/initial/fast/index.ts
```diff
@@ -72,8 +73,10 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
...
+    this.lastProcessedBlock = lastKnownState.latestBlockHeader.clone();
```
mpetrunic: or here

packages/lodestar/src/sync/initial/fast/index.ts
```diff
@@ -72,8 +73,10 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
...
+    this.lastProcessedBlock.stateRoot = lastKnownState.hashTreeRoot();
```
mpetrunic: why do you need state root in sync anyways?

packages/lodestar/src/chain/chain.ts
twoeths: other nodes will disconnect us with reason code 2 due to conditions in checking status message (our `shouldDisconnectOnStatus()` is more relaxed and not 100% follow the spec I think)

packages/lodestar/src/chain/blocks/validate.ts
twoeths: initially I don't have this change :).

I can't sync because initial sync doesn't received "processedBlock" event if I don't change that. The reason is like I explained above:
+ we start with slot 671, *but* finalized epoch is 21 (due to the concern I have), not 20 (which is the real epoch of slot 671)
+ initial sync fetches blocks 675+ which is epoch 21

packages/lodestar-cli/src/testnets/index.ts
twoeths: yeah true but I found comfortable as a developer option, it helps saved us a lot of time without having to sync from slot 0 ...

packages/lodestar/src/sync/initial/fast/index.ts
twoeths: without doing this we'll not get correct block root calculated from `this.lastProcessedBlock`
every time we process a block, we don't have a final `latestBlockHeader` as in https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar-beacon-state-transition/src/block/blockHeader.ts#L24 bc we don't have stateRoot at that time

packages/lodestar/src/chain/chain.ts
mpetrunic: I feel like we should check how other teams are doing that, we probably need to set checkpoints from state and refetch blocks between finalized and head slot via rpc

packages/lodestar-cli/src/testnets/index.ts
mpetrunic: create another function called getRemoteFile(localPath, urlOrPath) or something that calls downloadFile if url

packages/lodestar/src/sync/initial/fast/index.ts
mpetrunic: maybe just get finalized block root from forkChoice?

packages/lodestar/src/chain/chain.ts
```diff
@@ -253,6 +255,27 @@ export class BeaconChain implements IBeaconChain {
     this.logger.info("Beacon chain initialized");
   }
 
+  /**
+   * Initialize our chain from a weak subjectivity state.
+   * @param state a weak subjectivity state
+   */
+  public async initializeWeakSubjectivityState(weakSubjectivityState: TreeBacked<BeaconState>): Promise<void> {
```
wemeetagain, line 262: Can we just reuse `initializeChain` and `initForkChoice`? These methods should be doing what we need already (should be tweaked if need be).

packages/lodestar/src/chain/chain.ts
```diff
@@ -362,6 +395,30 @@ export class BeaconChain implements IBeaconChain {
...
+      root: blockRoot,
+      // TODO: is it possible to be `epoch + n` where n >= 2?
+      epoch: anchorState.slot === computeStartSlotAtEpoch(this.config, epoch) ? epoch : epoch + 1,
+    };
+    this.forkChoice.addBlock({
+      slot: anchorState.slot,
+      blockRoot,
+      stateRoot: this.config.types.BeaconState.hashTreeRoot(anchorState),
+      parentRoot: anchorState.latestBlockHeader.parentRoot.valueOf() as Uint8Array,
+      justifiedCheckpoint: blockCheckpoint,
```
wemeetagain: I believe these checkpoints should be state.currentJustifiedCheckpoint and state.finalizedCheckpoint.

twoeths: in that case we have to at least fetch blocks of `state.currentJustifiedCheckpoint` and `state.finalizedCheckpoint`, do we need to fetch all blocks from those checkpoints to the current head?

at this time, network is not initialized. I guess I have to do that in `sync` right before initial sync and call `chain` again from there, will figure that out.

packages/lodestar/src/chain/chain.ts
twoeths: > we probably need to set checkpoints from state 

yeah will do that, that's the below concern too ... In that case we don't have to create our self initiated checkpoint (`blockCheckpoint` in the above code snippet)

packages/lodestar/src/chain/chain.ts
wemeetagain: Oh, I see... I guess there's two interpretations of 'what to do', both with their own issues:
- set this state.latestBlockHeader as the finalized/justified checkpoint, in which case we don't actually have a valid block to add to the fork choice (we don't have a block where this state is the one thats finalized)
or
- set this state.latestBlockHeader as the block in the fork choice, in which case we don't have the blocks from the finalized/justified checkpoint up to the lastestBlock

These do different things, and I think we should figure out which is the 'correct' one.
My current thought is that the first interpretation is correct. In which case, what you have here for the checkpoints is correct. I think we will need an anchor block as well as a state in order to restore the chain. (eg: `--weakSubjectivityBlockFile` as well)

This aligns with the needs during genesis as well. Its just that for genesis, we can compute the genesis block, and don't need to provide it separately.

twoeths: if we go with "`state.latestBlockHeader as the finalized/justified checkpoint`" then I think we don't need anchor block since we have `stateRoot` as well, so that's enough to add to forkchoice

the concern for this approach is which epoch for the finalized/justified checkpoint is correct? in a lot of cases its `computeEpochAtSlot()` but it's not always be like that.

I wonder if one of the criteria to select a weak subjectivity state is it should not have a skipped slot at start of that epoch ...

mpetrunic: ain't wrong finalized checkpoint mess with peer status exchange for peer that check if we have proper finalized checkpoint?

twoeths: yeah that's why I got goodbye requests a lot using `computeEpochAtSlot()`. We need to make sure we have correct finalized checkpoint.

packages/lodestar/src/sync/initial/fast/index.ts
twoeths: we want both blockRoot and slot so `lastProcessedBlock` is being model as `BeaconBlockHeader`

packages/lodestar/src/chain/chain.ts
twoeths: > I feel like we should check how other teams are doing that

I don't see this in prysmatic and lighthouse

mpetrunic: https://github.com/sigp/lighthouse/pull/1468

prysm probably has Pr as well

wemeetagain (approved):

Issue #1499, opened 2020-09-02
Sync not abortable


mpetrunic:
Both regular and initial sync doesn't contain code required to abort. Especially problematic if there are missing peers since it contains loop and sleep.

Solution:
When sync stop is called make sure to abort all long running tasks via abort controller, abort iterator and abortable sleep

dapplion:
Is this fixed already?

mpetrunic:
No idea. I don't think we abort block fetching from peers.

dapplion:
With the review from https://github.com/ChainSafe/lodestar/pull/2164 I would consider this issue closed. When stopping the node libp2p.stop() closes the streams of all open connections which should abort all active ReqResp requests including fetching blocks from peers

PR #1505, opened 2020-09-03
Make waitForBestPeer sleep abortable

dapplion:
Partial https://github.com/ChainSafe/lodestar/issues/1499 fix. @mpetrunic seems that other streams and pushable are aborted on stop. Let me know if I'm missing something


mpetrunic:
This is here is issue:
https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/sync/utils/sync.ts#L96

Also, we should be able to abort rpc requests.

packages/lodestar/src/util/sleep.ts
```diff
@@ -1,3 +1,30 @@
-export function sleep(ms: number): Promise<void> {
-  return new Promise((resolve) => setTimeout(resolve, ms));
+import {AbortSignal} from "abort-controller";
+import {ErrorAborted} from "./errors";
+
+/**
+ * Abortable sleep function. Cleans everything on all cases preventing leaks
+ * On abort throws ErrorAborted
+ */
+export async function sleep(ms: number, signal?: AbortSignal): Promise<void> {
```
wemeetagain: Maybe this function should go in lodestar-utils ? we can use it in the validator and other places

wemeetagain:
bump

wemeetagain:
> ##[error]@chainsafe/lodestar: test/e2e/chain/genesis/genesis.test.ts(64,11): error TS2304: Cannot find name 'sleep'.

mpetrunic (dismissed):

dapplion:
Rebased to master to fix conflicts

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -85,7 +86,8 @@ export function fetchBlockChunks(
   reqResp: IReqResp,
   getPeers: () => Promise<PeerId[]>,
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  maxBlocksPerChunk?: number
+  maxBlocksPerChunk?: number,
```
mpetrunic, line 89: I think you can remove this, we don't even use this in method

dapplion: Better to do it in another PR to not mixup changes?

mpetrunic (approved):

wemeetagain (approved):

PR #1516, opened 2020-09-09
Save checkpoint states in block processor

wemeetagain:
This PR centers around splitting up the state transition inside our block processor to expose true checkpoint states (eg: states where `state.slot % SLOTS_PER_EPOCH === 0`).
We do this by stopping after each epoch transition, and emitting the state directly after the epoch transition.
(This can be extended to `await` and allow other tasks to complete after each epoch :dancers: )

This PR doesn't change any behavior, just adds an additional source of data for the checkpoint state cache. In the future, we can persist some of these states or use the checkpoint state cache to regenerate pruned states.

### DIFF
--- a/packages/lodestar/src/chain/blocks/process.ts
+++ b/packages/lodestar/src/chain/blocks/process.ts
@@ -169,8 +169,7 @@ export async function runStateTransition(
       verifyProposer: !job.trusted,
       verifySignatures: !job.trusted,
     }) as ITreeStateContext;
-    const blockSlot = job.signedBlock.message.slot;
-    if (blockSlot % SLOTS_PER_EPOCH === 0) {
+    if (postSlot % SLOTS_PER_EPOCH === 0) {
       emitCheckpointEvent(emitter, postStateContext);
     }
     return postStateContext;
--- a/packages/lodestar/src/chain/blocks/process.ts
+++ b/packages/lodestar/src/chain/blocks/process.ts
@@ -3,6 +3,7 @@ import {
   computeEpochAtSlot,
   computeStartSlotAtEpoch,
   fastStateTransition,
+  ZERO_HASH,
 } from "@chainsafe/lodestar-beacon-state-transition";
 import {IStateContext} from "@chainsafe/lodestar-beacon-state-transition/lib/fast/util";
 import {processSlots} from "@chainsafe/lodestar-beacon-state-transition/lib/fast/slot";
@@ -134,10 +135,14 @@ export function emitCheckpointEvent(emitter: ChainEventEmitter, checkpointStateC
   const config = checkpointStateContext.epochCtx.config;
   const slot = checkpointStateContext.state.slot;
   assert.true(slot % config.params.SLOTS_PER_EPOCH === 0, "Checkpoint state slot must be first in an epoch");
+  const blockHeader = config.types.BeaconBlockHeader.clone(checkpointStateContext.state.latestBlockHeader);
+  if (config.types.Root.equals(blockHeader.stateRoot, ZERO_HASH)) {
+    blockHeader.stateRoot = config.types.BeaconState.hashTreeRoot(checkpointStateContext.state);
+  }
   emitter.emit(
     "checkpoint",
     {
-      root: config.types.BeaconBlockHeader.hashTreeRoot(checkpointStateContext.state.latestBlockHeader),
+      root: config.types.BeaconBlockHeader.hashTreeRoot(blockHeader),
       epoch: computeEpochAtSlot(config, slot),
     },
     checkpointStateContext
--- a/packages/lodestar/src/chain/blocks/process.ts
+++ b/packages/lodestar/src/chain/blocks/process.ts
@@ -136,6 +136,8 @@ export async function runStateTransition(
   job: IBlockProcessJob
 ): Promise<IStateContext | null> {
   const config = stateContext.epochCtx.config;
+  const {SLOTS_PER_EPOCH} = config.params;
+  const {BeaconBlockHeader} = config.types;
   try {
     const preSlot = stateContext.state.slot;
     const postSlot = job.signedBlock.message.slot;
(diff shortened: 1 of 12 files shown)


packages/lodestar/src/chain/blocks/process.ts
```diff
@@ -126,27 +131,43 @@ export function updateForkChoice(
...
+    const postStateContext = fastStateTransition(stateContext, job.signedBlock, {
       verifyStateRoot: true,
       verifyProposer: !job.trusted,
       verifySignatures: !job.trusted,
-    });
+    }) as ITreeStateContext;
+    const blockSlot = job.signedBlock.message.slot;
+    if (blockSlot % config.params.SLOTS_PER_EPOCH === 0) {
+      const checkpoint: Checkpoint = {
+        root: stateContext.state.blockRoots[Math.max(0, blockSlot - 1) % config.params.SLOTS_PER_HISTORICAL_ROOT],
```
twoeths: I don't get why it is `blockSlot - 1` instead of `blockSlot`. Should we use `getBlockRootAtSlot` instead?

packages/lodestar/src/chain/blocks/process.ts
```diff
@@ -126,27 +131,43 @@ export function updateForkChoice(
...
+  const config = stateContext.epochCtx.config;
   try {
+    const preSlot = stateContext.state.slot;
+    const postSlot = job.signedBlock.message.slot;
+    const preEpoch = computeEpochAtSlot(config, preSlot);
+    let nextEpochSlot = computeStartSlotAtEpoch(config, preEpoch + 1);
+    while (nextEpochSlot < postSlot) {
+      processSlots(stateContext.epochCtx, stateContext.state, nextEpochSlot);
+      const checkpoint: Checkpoint = {
+        root: stateContext.state.blockRoots[(nextEpochSlot - 1) % config.params.SLOTS_PER_HISTORICAL_ROOT],
```
twoeths: this results in `getBlockRootAtSlot(nextEpochSlot - 1)`. Did you mean `getBlockRootAtSlot(nextEpochSlot )`?

wemeetagain: I think we need something like:
```typescript
const blockHeader = stateContext.state.latestBlockHeader.clone();
blockHeader.stateRoot = stateContext.state.hashTreeRoot();
const root = blockHeader.hashTreeRoot();
```

Below, we can just hashTreeRoot the job.block.message

(we can't just use `getBlockRootAtSlot(nextEpochSlot)`, because that only gets set on `nextEpochSlot+1`)

packages/lodestar/src/chain/blocks/process.ts
```diff
@@ -126,27 +131,47 @@ export function updateForkChoice(
...
+  const {BeaconBlockHeader} = config.types;
   try {
+    const preSlot = stateContext.state.slot;
+    const postSlot = job.signedBlock.message.slot;
+    const preEpoch = computeEpochAtSlot(config, preSlot);
+    let nextEpochSlot = computeStartSlotAtEpoch(config, preEpoch + 1);
+    while (nextEpochSlot < postSlot) {
+      processSlots(stateContext.epochCtx, stateContext.state, nextEpochSlot);
+      const blockHeader = BeaconBlockHeader.clone(stateContext.state.latestBlockHeader);
+      blockHeader.stateRoot = stateContext.state.hashTreeRoot();
```
twoeths: it seems that we should only set `stateRoot` if it's ZERO_HASH https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar-beacon-state-transition/src/slot.ts#L32

I think checkpoint `root` is the same for this whole while loop because these are skipped slots, only checkpoint epoch is different. It means checkpoint `root` can even be calculated before this `while` loop.

mpetrunic (commented):
Not 100% but this might break initial sync or at least it's making "wrong checkpoint events". Idea of initial sync is to sync up to agreed finalized checkpoint. So if we listen to checkpoint events peer finalized checkpoint won't match emitted checkpoint. We cannot listen on finalized since we need to go at least 2 epochs above finalized to get that event.

I would propose to keep existing checkpoint event as is and emit epochBoundaryCheckpoint or something for creating/storing those states. wdyt?

wemeetagain:
> Not 100% but this might break initial sync or at least it's making "wrong checkpoint events".

This shouldn't be changing the checkpoints that are emitted. If it doesn't match, we're doing something wrong here. 

This does two things (which are mostly backwards compatible):
- adds the state at the start-slot of the checkpoint to the event. (This is important because we want to later determine the checkpoint from the state itself. In that case we need the state.slot to be dialed to the correct epoch. We can't just rely on the state with the stateRoot of the checkpoint root.)
- makes sure to emit after _every_ epoch (rather than skip checkpoints that are fully skip-slotted)

3xtr4t3rr3str14l (dismissed):
LGTM

packages/lodestar/src/chain/blocks/process.ts
```diff
@@ -125,28 +130,52 @@ export function updateForkChoice(
...
     // if block is trusted don't verify proposer or op signature
-    return fastStateTransition(stateContext, job.signedBlock, {
+    const postStateContext = fastStateTransition(stateContext, job.signedBlock, {
       verifyStateRoot: true,
       verifyProposer: !job.trusted,
       verifySignatures: !job.trusted,
-    });
+    }) as ITreeStateContext;
+    if (postSlot % SLOTS_PER_EPOCH === 0) {
+      emitCheckpointEvent(emitter, postStateContext);
```
twoeths, line 178: I suppose `postStateContext.state.latestBlockHeader.stateRoot` is ZERO_HASH so we can't use `emitCheckpointEvent` here.

The old emitter code of getting root from `config.types.BeaconBlock.hashTreeRoot(job.signedBlock.message)` should work fine.

wemeetagain: we never run state transition against the genesis block so we should be good here :)

twoeths: I refer to this https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar-beacon-state-transition/src/block/blockHeader.ts#L24

my understanding is `postStateContext.state.latestBlockHeader.stateRoot` is always ZERO_HASH after we run a state transition since we don't know stateRoot at that time. Next `stateTransition` or `processSlot` should set previous `blockRoot` to state

twoeths (approved):

Issue #1381, opened 2020-08-10
[waiting for db upload]beacon chain halt on incorrect fork


GregTheGreek:
**Describe the bug**

Note: This is reproducible, after many restarts it seems that I cannot proceed from this state. [link to db](https://lodestar-db-dumps.s3.us-east-2.amazonaws.com/lodestar-db.tar)

It seems when we encounter a fork the beacon node blows up and halts.
```
2020-08-10 18:25:49  [CHAIN]            info: Processed new chain head newChainHeadRoot=0x3ca7be2b487ac5203e40c6f7a4fa7bf5f5e884265e7e2e43b1872af7bbabc060, slot=44480, epoch=1390
2020-08-10 18:25:49  [CHAIN]            info: Epoch 1389 is justified at root     0xe3fd646d11c20f2610bb71331a483e591e6aa1a81f2d214bc7aac191ffacc570! 
2020-08-10 18:25:49  [CHAIN]            info: Epoch 1388 is finalized at root     0x1126b9aa2f6e1c50f739bdc8e54e1d47ab16b382b8d312aec3cdf1dfe495031b! 
(node:24368) UnhandledPromiseRejectionWarning: Error: Fork choice: Block slot 44481 is not on the same chain, finalized slot=44416: null === 0x1126b9aa2f6e1c50f739bdc8e54e1d47ab16b382b8d312aec3cdf1dfe495031b
```

**Expected behavior**

Not blow up :+1: 

**Steps to Reproduce**
See attached db, run from the db.

**Desktop (please complete the following information):**
 - OS:  ubuntu 
 - Version:  18
 - Branch:  master
 - Commit hash:  #c803315

PR #1517, opened 2020-09-09
Separate fork choice

wemeetagain:
Current Status:
- able to run the chain from genesis and restore from checkpoints
- tests untouched and likely failing

Big items:

- `StateContextCheckpointCache` has been internally refactored, and its public interface extended
  - add `pruneFinalized(cp: Checkpoint)` - will prune checkpoints from epochs before the finalized epoch
  - add `prune(finalized: Checkpoint, justified: Checkpoint)` - similar to `stateCache.prune` - if there are more epochs present than `MAX_EPOCHS`, will prune starting from the lowest epoch to the highest epoch, skipping the finalized and justified epochs.
  - add `getLatest(cp: Checkpoint): ITreeStateContext` - starting from `cp.epoch` and going backwards, searches the cache for hits with `cp.root`. "Finds the latest state context at the checkpoint where the checkpoint's root is cp.root" and the checkpoint's epoch is lte cp.epoch."
- `StateRegenerator` has been added - `src/chain/regen.ts`
  - it can regenerate states from blocks _that have already been processed_ , given `forkChoice`, `stateCache`, `checkpointStateCache`, and `db.blocks`
  - it intelligently attempts to perform the smallest amount of regeneration necessary to regen the requested state. To that end, it exposes three different methods.
  - it reuses some block processing machinery (`emitCheckpointEvent`, `runStateTransition`) so as to enable re-seeding the checkpoint cache
  - it throws a `RegenError` with well-defined error codes and types, this allows us to tighten the interface to disallow nulls.
  - it has one major TODO, it must queue requests instead of allowing multiple concurrent requests
  - `getPreState(block: BeaconBlock): Promise<ITreeStateContext>` - the same as `getPreState` that we were using in our block processor. It can request a more recent checkpoint state if the block occurs after many skip-slots, skipping epoch transitions :) 
  - `getCheckpointState(cp: Checkpoint): Promise<ITreeStateContext>` - returns the checkpoint state. It minimizes regeneration by first searching the checkpoint cache for checkpoint states with cp.root, which may then just be `processSlots`ed. Otherwise regens the 'ol-fashioned way', by state root.
  - `getState(stateRoot: Root): Promise<ITreeStateContext>` - returns the state whose root is stateRoot. This is where the actual block replay happens. It iterates backwards thru the block tree from the requested state, to the first state that we have cached, (searching the checkpointStateCache as well!), then replaying the blocks that were iterated thru to the requested state, once a state has been found.
- Adds an optional `EpochProcess` object to `EpochContext`.
  - This object is used in epoch transitions to simplify validator balance counting. It holds flattened validators, validator statuses (whether they can attest, whether they've attested for important recent states, etc), total attested balances, etc.
  - This ends up being very useful for updating the fork choice balances.
  - We will find more uses for this after this PR, likely will be very nice to scrape for metrics re network participation.
- Block processing refactored slightly
  - `getPreState` is now handled by the `StateRegenerator`
  - `runStateTransition` and other utilities moved to `src/chain/blocks/stateTransition.ts`
- Block/state archival tasks refactored
  - they are now both triggered by the `forkChoice:finalized` event, the `forkChoice:pruned` event is no more.
  - we also prune state caches, triggered by the `checkpoint` event, but not in a task as its a one-liner
  - block archival no longer loads _all_ blocks in `block` repository into memory at once, instead streams
  - block archival strategy, run before fork choice pruning
    - stream all blocks in `block` repository, one by one
    - if the block is more recent than the finalized block, ignore
    - if the finalized block is a descendant of the block, store in the blockArchive
    - delete all un-ignored blocks
  - state archival now archives the finalized state from `checkpointStateCache` (these will be candidates for weak subjectivity :) )
- Add `@chainsafe/lodestar-fork-choice` package
  - split into two 'levels', protoArray, and forkChoice
  - protoArray - the block dag, not much else
  - forkChoice
    - provides an interface very close to the spec
    - includes an `IForkChoiceStore`, very similar to `Store` in the spec -- this is implemented in the lodestar package with knowledge of the `ChainEventEmitter`
    - stores votes, balances, queued attestations
  - Attestation weighting is completely different than before
    - Instead of feeding attestations in with weights attached per attestation, we feed in something like a `LatestMessage` from the spec.
    - attestations are reweighted all-at-once, using `forkChoice.updateBalances`. Using `epochContext.epochProcess`, this is fast, lots of the work was already done as part of epoch processing.
    - attestation reweighting is triggered by the `forkChoice:justified` event.


packages/lodestar/src/chain/regen.ts
```diff
@@ -0,0 +1,205 @@
...
+    for (const b of blocksToReplay.reverse()) {
+      const block = await this.db.block.get(b.blockRoot);
+      if (!block) {
+        throw new RegenError({
+          code: RegenErrorCode.ERR_BLOCK_NOT_IN_DB,
+          blockRoot: b.blockRoot,
+        });
+      }
+      try {
+        stateCtx = await runStateTransition(this.emitter, this.forkChoice, stateCtx, {
```
twoeths, line 191: in case of regen, we may not want to emit anything to trigger unnecessary "onCheckpoint" at a couple of places

packages/lodestar/src/tasks/tasks/archiveBlocks.ts
```diff
@@ -4,81 +4,60 @@
...
-        archivedBlocks.push(lastBlock);
+    const finalizedSlot = computeStartSlotAtEpoch(this.config, this.finalized.epoch);
+    const keysToDelete = [];
+    let totalArchived = 0;
+    for await (const {key, value} of await this.db.block.entriesStream()) {
+      if (value.message.slot > finalizedSlot) {
+        continue;
       }
+      if (this.forkChoice.isDescendant(key, this.finalized.root)) {
+        await this.db.blockArchive.add(value);
```
twoeths, line 51: sometimes it comes up to 3.5s to finish the task: `2020-09-12 12:31:37  [CHORES]           info: Archive Blocks - duration=3509ms`. Maybe just collect all and do `batchAdd` to save time? If it's too much then probably collect and save per n blocks.

packages/lodestar/src/chain/chain.ts
```diff
@@ -192,19 +203,22 @@ export class BeaconChain implements IBeaconChain {
     await this.attestationProcessor.start();
     this._currentForkDigest = computeForkDigest(this.config, state.fork.currentVersion, state.genesisValidatorsRoot);
     this.emitter.on("forkVersion", this.handleForkVersionChanged);
-    this.emitter.on("clock:epoch", this.onClockEpoch);
+    this.emitter.on("clock:slot", this.onClockSlot);
     this.emitter.on("checkpoint", this.onCheckpoint);
+    this.emitter.on("forkChoice:justified", this.onForkChoiceJustified);
+    this.emitter.on("forkChoice:finalized", this.onForkChoiceFinalized);
     this.emitter.on("error:block", this.onErrorBlock);
-    await this.restoreHeadState(state, epochCtx);
```
twoeths, line 198: we don't call this intentionally? if we don't use it, we should remove.

packages/lodestar/src/tasks/index.ts
```diff
@@ -70,8 +74,27 @@ export class TasksService implements IService {
     await this.interopSubnetsTask.stop();
   };
 
-  private handleFinalizedCheckpointChores = async (finalized: BlockSummary, pruned: BlockSummary[]): Promise<void> => {
-    await new ArchiveBlocksTask(this.config, {db: this.db, logger: this.logger}, finalized, pruned).run();
-    await new ArchiveStatesTask(this.config, {db: this.db, logger: this.logger}, finalized, pruned).run();
+  private onFinalizedCheckpoint = async (finalized: Checkpoint): Promise<void> => {
+    await Promise.all([
+      new ArchiveStatesTask(this.config, {db: this.db, logger: this.logger}, finalized).run(),
+      new ArchiveBlocksTask(
```
twoeths, line 80: I think we want to make sure ArchiveBlock finish before starting ArchiveTask due to https://github.com/ChainSafe/lodestar/issues/1381

packages/lodestar/src/chain/chain.ts
wemeetagain: Yeah, the same thing happened to me that happened to you :stuck_out_tongue:. Replayed half the blocks thru the medalla chasm (eg: to 85k) then had to start all the way back over.
Just removing this line was enough to stop restoration. :+1:  for removing restoration for now. Eventually I'd like to bring it back in certain circumstances, but with a different implementation. Fork choice persistence + restoration.

packages/lodestar/src/tasks/tasks/archiveBlocks.ts
wemeetagain: I refactored this task bc it caused issues archiving the medalla chasm. Pulling all blocks into memory could cause OOM. The slowdown here is the tradeoff of not pulling all blocks in memory at once.
Unfortunately, collecting all and doing a batchAdd will have the same issue as pulling all into memory at once.

Collecting N blocks at a time seems like a good middle ground. I'll make an issue for this as this PR is getting too big.

packages/lodestar/src/chain/regen.ts
wemeetagain: I think we should emit these events any time blocks or checkpoints are processed (or reprocessed). I'd like to also trigger the `block` event here (`runStateTransition`) too.

Re: emitting events in the state regenerator, it was intentional. The thought is would help re-seed our checkpointStateCache, since that's the trigger for updating that cache. (I'd like to do the same with the `block` event and the stateCache)

If it turns out emitting events all the time isn't what we want, we have some clear paths forward. We can always hook into the job metadata to inform whether or not to emit. Eg: don't emit events when job.reprocess === true

wemeetagain:
closing in favor of smaller PRs

PR #1518, opened 2020-09-09
Refactor eth1 data for block production

dapplion:
Refactors (enables) eth1 data for block production. The previous iteration had flaws and PR https://github.com/ChainSafe/lodestar/pull/1443 deactivated block production. This PR will to re-enable it.

#### Architecture overview

The architecture of this PR is heavily inspired by Lighthouse. There is a specific class for block production with a minimal interface meant to be called by the block production API for a given state. The logic to prepare both the eth1Data vote and the deposits with proofs is encapsulated in the class (moved from its previous location of `chain/factory/block`
```ts
export interface IEth1ForBlockProduction {
  getEth1DataAndDeposits(state: TreeBacked<BeaconState>): Promise<{eth1Data: Eth1Data; deposits: Deposit[]}>;
}
```

The class `IEth1ForBlockProduction` upon instantiation will fetch both deposits and blocks from an eth1 node at intervals. The class consumes two other classes who wrap the `db` module and handle read / write opts
- `Eth1BlockHeaderCache`: Eth1 block headers indexed by blockNumber
- `Eth1DepositsCache`: Eth1 deposit logs indexed by validator index + partial eth1 data (depositCount, depositRoot) indexed by blockNumber. The latter is derived from deposit logs exclusively.

#### Goals

All logic whenever possible is stored in pure utility functions which are unit tested without stubs. The classes should contain as little logic as possible and just connect the logic heavy function with internal state and data sources.

Database and data fetching concerns are split into different classes, brought together only in the main class `IEth1ForBlockProduction`.

`IEth1ForBlockProduction` starts its long-running process in the constructor, not requiring a `start` method. To stop and destroy the class, it accepts an `AbortSignal` as a constructor parameter. The signal will stop the main setInterval handle and also interrupt HTTP requests with the new httpEth1Client functions.


dapplion:
I've decided to try to split this PR into smaller ones for the health of reviewers

packages/lodestar-beacon-state-transition/src/block/eth1Data.ts
```diff
@@ -2,19 +2,33 @@
...
+export function getNewEth1Data(config: IBeaconConfig, state: BeaconState, eth1Data: Eth1Data): Eth1Data | null {
+  const numVotes = Array.from(state.eth1DataVotes).filter((vote) => config.types.Eth1Data.equals(vote, eth1Data))
+    .length;
+
+  // The +1 is to account for the `eth1Data` supplied to the function.
+  if ((numVotes + 1) * 2 > config.params.EPOCHS_PER_ETH1_VOTING_PERIOD * config.params.SLOTS_PER_EPOCH) {
+    return eth1Data;
+  } else {
+    return null;
+  }
```
wemeetagain: Likely not worth it to change `processEth1Data` here.
We're not actually even using this implementation right now, but an optimized implementation in `src/fast/block/processEth1Data.ts`

Edit: nvm, you're using this in `Eth1ForBlockProduction`

packages/lodestar/src/eth1/types.ts
```diff
@@ -0,0 +1,50 @@
+import {Bytes32, Number64, Root, DepositData, IBeaconSSZTypes} from "@chainsafe/lodestar-types";
+import {ContainerType} from "@chainsafe/ssz";
+
+export interface IDepositEvent {
+  depositData: DepositData;
+  /// The block number of the log that included this `DepositData`.
+  blockNumber: Number64;
+  /// The index included with the deposit log.
+  index: Number64;
+}
```
wemeetagain: You might just update the `DepositEvent` type in lodestar-types. This type is basically that + blockNumber.

packages/lodestar/src/eth1/utils/eth1Vote.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export function pickEth1Vote(
+  config: IBeaconConfig,
+  state: TreeBacked<BeaconState>,
+  votesToConsider: Eth1Data[]
+): Eth1Data {
+  const votesToConsiderHashMap = new Set<string>();
+  for (const eth1Data of votesToConsider) votesToConsiderHashMap.add(serializeEth1Data(eth1Data));
+
+  const validVotes = Array.from(state.eth1DataVotes).filter((eth1Data) =>
```
wemeetagain: Places where we're doing this (`Array.from(state.XXX)`) can be replaced with `readOnlyMap(state.XXX, (x) => x)` for a speedup.

packages/lodestar/src/eth1/eth1DepositsCache.ts
```diff
@@ -0,0 +1,122 @@
...
+    const depositRootTree = await this.db.depositDataRoot.getDepositRootTree();
+    return getDepositsWithProofs(depositEvents, depositRootTree, depositCount);
+  }
+
+  /**
+   * Add log to cache
+   * This function enforces that `logs` are imported one-by-one with no gaps between
+   * `log.index`, starting at `log.index == 0`.
+   */
+  async insertLogs(depositEvents: IDepositEvent[]): Promise<void> {
```
wemeetagain: It might be nice to standardize to use either "log" or "event", not both.

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -52,18 +53,20 @@ export class ValidatorApi implements IValidatorApi {
...
   public constructor(
     opts: Partial<IApiOptions>,
-    modules: Pick<IApiModules, "config" | "chain" | "db" | "sync" | "network" | "logger">
+    modules: Pick<IApiModules, "config" | "chain" | "db" | "eth1" | "sync" | "network" | "logger">
   ) {
     this.namespace = ApiNamespace.VALIDATOR;
     this.config = modules.config;
     this.chain = modules.chain;
     this.db = modules.db;
+    this.eth1 = modules.eth1;
```
wemeetagain, line 69: Why does the validator API need IEth1ForBlockProduction? Seems it is unused?

packages/lodestar/src/db/api/beacon/repositories/depositEvent.ts
```diff
@@ -0,0 +1,45 @@
...
+import {IDepositEvent, DepositEventGenerator} from "../../../../eth1";
+
+import {IDatabaseController} from "../../../controller";
+import {Bucket} from "../../schema";
+import {Repository} from "./abstract";
+
+/**
+ * DepositData indexed by deposit index
+ *
+ * ### TODO: Are deposits meant to be included via gossip? They are not
```
wemeetagain: I believe since the eth1 chain exists as an independent data source (that can be subscribed to separately), we don't need to gossip deposits.

We should likely be pruning deposit logs included before a finalized checkpoint, just to free up disc space.

packages/lodestar/src/eth1/types.ts
```diff
@@ -0,0 +1,50 @@
...
+export const Eth1BlockGenerator = (ssz: IBeaconSSZTypes): ContainerType<IEth1Block> =>
+  new ContainerType({
+    fields: {
+      blockHash: ssz.Bytes32,
+      blockNumber: ssz.Number64,
+      timestamp: ssz.Number64,
+    },
+  });
+
+export interface IEth1DataDeposit {
```
wemeetagain: This is an Eth1Data, but instead of a blockHash, has a blockNumber. Maybe consider renaming this? I don't think it really has anything more to do with a Deposit than an Eth1Data. Maybe something like `IOrderedEth1Data` or `INumberedEth1Data`.

wemeetagain (commented):
Great job. I like the provided interface in `IEth1ForBlockProduction`.

I'm a little confused why storing `DepositEvent`s (indexed by deposit index), `Eth1Data`s (indexed by timestamp) and lastStoredBlockNumber directly isn't sufficient.

Instead we're storing `IEth1Block`s (indexed by block number) and `IEth1DataDeposit`s (indexed by block number). Then, to create valid `Eth1Data`s, we have to stream backwards to get the candidate block range, then joining with `Eth1DataDeposit`s on block number.

I guess I don't see why this joining process is done after storing eth1 blocks and deposits, rather than before storing. IMO doing this work up front is better bc it can be done once, up-front, (rather than on every block production) and it makes candidate `Eth1Data` trivial.

packages/lodestar-beacon-state-transition/src/block/eth1Data.ts
dapplion: Oh right, I didn't realize there's the fast version. I should use that instead of the regular one.

packages/lodestar/src/eth1/types.ts
dapplion: True! Is it used by anything in current master? Seems not.

packages/lodestar/src/eth1/eth1DepositsCache.ts
dapplion: Right, on the smaller PRs I'm settling to depositEvents

packages/lodestar/src/api/impl/validator/validator.ts
dapplion: It is used for block production
https://github.com/ChainSafe/lodestar/blob/3492e8e7973baeb4a75e4575d7017cfa2310cd67/packages/lodestar/src/api/impl/validator/validator.ts#L86-L95

but the class is instantiated in a parent class
https://github.com/ChainSafe/lodestar/blob/3492e8e7973baeb4a75e4575d7017cfa2310cd67/packages/lodestar/src/node/nodejs.ts#L137-L145

packages/lodestar/src/eth1/types.ts
dapplion: Agree, those are much better names

dapplion:
> Instead we're storing `IEth1Block`s (indexed by block number) and `IEth1DataDeposit`s (indexed by block number). Then, to create valid `Eth1Data`s, we have to stream backwards to get the candidate block range, then joining with `Eth1DataDeposit`s on block number.
> 
> I guess I don't see why this joining process is done after storing eth1 blocks and deposits, rather than before storing. IMO doing this work up front is better bc it can be done once, up-front, (rather than on every block production) and it makes candidate `Eth1Data` trivial.

I considered that option but the tradeoff is that you couple block fetching + storing with log fetching + storing more. If you fetch a block range but the deposits for that range are not received yet, you have to wait or throw and re-fetched the same blocks again latter. Lighthouse uses a similar approach to this PR. There's `getDepositCountForBlockNumber` and `getDepositRootForBlockNumber` which stream deposits from the DB to find the closest `<= deposit.blockNumber` that matches the query. A third option would be to not store this partial eth1 data (`IOrderedEth1Data`) at all. Just keep it in memory and preprocess once at start.

wemeetagain:
> I considered that option but the tradeoff is that you couple block fetching + storing with log fetching + storing more. If you fetch a block range but the deposits for that range are not received yet, you have to wait or throw and re-fetched the same blocks again latter.

Blocks and logs are coupled either way though, right? just a matter of when the coupling happens, between fetching and db storing, or after db retrieval.

either (pseudocode):
```typescript
// storage is made simpler, most transformation work is deferred to retrieval
async function store(blocks: IEth1Block[], logs: DepositEvent[]): Promise<void> {
  const orderedEth1Datas = transformLogsToOrderedEth1Data(logs);
  await Promise.all([
    db.batchAdd(blocks),
    db.batchAdd(logs),
    db.batchAdd(orderedEth1Datas),
  ]);
}

async function retrieve(from: number, to: number): Promise<[Eth1Data[], Deposit[]]> {
  const blocks = await db.getBlocksByTimestamp(from, to);
  const logs = await db.getLogsByBlocks(blocks);
  const orderedEth1Datas = await db.getOrderedEth1DataByBlocks(blocks);
  // join done on retrieval
  const eth1Datas = joinBlocksAndOrderedEth1Data(blocks, orderedEth1Datas);
  const deposits = transformLogsToDeposits(logs);
  return [eth1Datas, deposits];
}
```

or

```typescript
// storage incudes most transformation work, retrieval is simpler
async function store(blocks: IEth1Block[], logs: DepositEvent[]): Promise<void> {
  const orderedEth1Datas = transformLogsToOrderedEth1Data(logs);
  // join done on storage
  const eth1Datas = joinBlocksAndOrderedEth1Data(blocks, orderedEth1Datas);
  const eth1DatasIndices = blocks.map(b => b.timestamp);
  await Promise.all([
    db.batchAdd(logs),
    db.batchAdd(etn1DatasIndices, eth1Datas),
  ]);
}

async function retrieve(from: number, to: number): Promise<[Eth1Data[], Deposit[]]> {
  const eth1Datas = await db.getEth1DatasByTimestamp(from, to);
  const logs = await db.getLogsByEth1Datas(eth1Datas);
  const deposits = transformLogsToDeposits(logs);
  return [eth1Datas, deposits];
}
```
Both snippets perform roughly the same operations, just at different times.

I think the tradeoffs lean towards doing the work up front. In the current approach, we simplify fetching and storing at the expense of storing double data (OrderedEth1Data and IEth1Blocks) as well as reindexing IEth1Blocks by timestamp and recomputing Eth1Data, on every block proposed.
This might make sense if there was a clear need to avoid doing the work up front. Like if fetching and storing was a bottleneck and we risked not having the data in time. But it seems that the only time that large swaths of eth1 are queried are on initial sync. In that case, the work of syncing blocks, running the state transition, will strongly outweigh any transformation work done here.
That being the case, we should take advantage of the strategy of doing the work once, rather than 'every time', as 'every time' gets worse, approaching 'every block' as the number of attached validators increases towards state.validators.length.
In Lighthouse's case, they are fortunate to have fast code and the ability to easily push work out to workers. In our case, we have to be greedy with every opportunity to do less work when we can.

dapplion:
Isn't indexing eth1Data by timestamp a bad idea? That is not a unique identifier for blocks. Although not likely N numbers of blocks could have the same timestamp. Is it more expensive to stream values and start / stop when the timestamp conditions are met?

wemeetagain:
> Isn't indexing eth1Data by timestamp a bad idea? That is not a unique identifier for blocks. Although not likely N numbers of blocks could have the same timestamp.

Has this ever happened in Ethereum's history? :joy: I would argue that its a practical impossibility. Blocks must be broadcast over a network and processed sequentially and new blocks require work to calculate. This practically ensures that every block happens measurably _after_ its parent.

If it does happen though, we should just ignore / overwrite the earlier-blocknumber-but-same-timestamp blocks. In the worst case, we end up voting for a more recent eth1data b/c the most-voted eth1data is an earlier-blocknumber-but-same-timestamp block. This won't effect our block proposal, and only delays beacon consensus on eth1data, so not the end of the world, especially given the 1 in 10M+ odds.

Is it more expensive to stream values and start / stop when the timestamp conditions are met?

Stream from the database? It would be significantly simplified and much cheaper since we'd be requesting data exactly as its indexed.
```typescript
db.eth1Data.values({
  gte: periodStart - SECONDS_PER_ETH1_BLOCK * ETH1_FOLLOW_DISTANCE * 2,
  lte: periodStart - SECONDS_PER_ETH1_BLOCK * ETH1_FOLLOW_DISTANCE,
});
```
Stream from an eth1 provider? It doesn't have to be more complicated. We can still stream using ETH1_FOLLOW_DISTANCE, not timestamp.

dapplion:
> Has this ever happened in Ethereum's history?  I would argue that its a practical impossibility. Blocks must be broadcast over a network and processed sequentially and new blocks require work to calculate. This practically ensures that every block happens measurably _after_ its parent.

Alright I agree that's a practical impossibility so let's index by timestamp. I'll edit the cache PRs but this two don't have to be modified
- https://github.com/ChainSafe/lodestar/pull/1519
- https://github.com/ChainSafe/lodestar/pull/1524

dapplion:
I've simplified the storage and transformation of blocks + deposits following @wemeetagain recommendations.

All transformation work is done upfront before storing two DB. Now it uses only the existing 3 db repositories, but slightly modified.

- `eth1Data`: `Eth1DataOrdered` type. Stores blockNumber to reliably know the latest fetched blockNumber to fetch the next range. It could be sufficient to store `latestFetchedBlockNumber` but adding a u64 field adds negligible disk overhead and simplifies things.
```ts
interface Eth1DataOrdered extends Eth1Data {
    blockNumber: Number64;
}
```
- `depositEvent`: `DepositEvent` type. Required depositData plus metadata to compute a `depositCountByBlockNumber` afterwards.
```ts
interface DepositEvent {
    depositData: DepositData;
    blockNumber: Number64;
    index: Number64;
}
```
- `depositDataRoot`: same

---

Eth1 blocks and deposits and fetched in parallel. 
- `fetch deposits` => `store depositEvent, store depositDataRoots`
- `fetch blocks` => `compute eth1Data for block range` => `store eth1Data`

To compute eth1Data for a block range, it computes two mappings from the already fetched deposits:
- depositCountByBlockNumber by streaming deposits from the db.depositEvent repository
- depositRootByDepositCount by popping values from a clone of the db.depositDataRoot repository's cached tree
Those mappings are computed just for the necessary range of block numbers. So each deposit should only be read once after a successful fetch of eth1 block headers.

dapplion:
Added extensive unit tests to all added utilities + e2e tests on parent classes querying Medalla data.

The main integration test is `packages/lodestar/test/e2e/eth1/eth1ForBlockProduction.test.ts` which should assert the entire work of this PR. It starts fetching deposits and blocks from a Goerli node and stops when receives both events and blocks. Then it queries `getEth1DataAndDeposits()` asserting the returning data as good for block production.

Coverage report (unit + e2e)

```
-------------------------------------|---------|----------|---------|---------|---------------------------------------
File                                 | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s                     
-------------------------------------|---------|----------|---------|---------|---------------------------------------
 eth1                                |   91.84 |    59.62 |   90.91 |   92.59 |                                       
  depositContract.ts                 |     100 |      100 |     100 |     100 |                                       
  eth1DataCache.ts                   |     100 |       50 |     100 |     100 | 24                                    
  eth1DepositsCache.ts               |   85.19 |       50 |     100 |   84.62 | 40-44                                 
  eth1ForBlockProduction.ts          |   93.75 |    68.75 |   83.33 |   93.75 | 60,112,116-117                        
  eth1ForBlockProductionDisabled.ts  |      50 |      100 |       0 |      50 | 20                                    
  ethers.ts                          |    93.1 |       70 |     100 |    96.3 | 40                                    
  httpEth1Client.ts                  |     100 |      100 |     100 |     100 |                                       
  httpRpcClient.ts                   |      75 |       50 |   83.33 |   77.78 | 25-35,71,86                           
  index.ts                           |     100 |      100 |     100 |     100 |                                       
  options.ts                         |     100 |      100 |     100 |     100 |                                       
  stream.ts                          |     100 |       50 |     100 |     100 | 36-68                                 
 eth1/utils                          |   91.22 |     82.5 |   84.21 |   92.42 |                                       
  deposits.ts                        |   86.67 |       75 |   71.43 |   88.46 | 46,69,75                              
  eth1Data.ts                        |   94.83 |    94.44 |   91.67 |   96.15 | 112-113                               
  eth1DepositEvent.ts                |     100 |      100 |     100 |     100 |                                       
  eth1Vote.ts                        |   88.89 |    66.67 |   77.78 |   91.67 | 49-52                                 
  groupDepositEventsByBlock.ts       |     100 |      100 |     100 |     100 |                                       
  optimizeNextBlockDiffForGenesis.ts |     100 |      100 |     100 |     100 |                                       
  retryProvider.ts                   |      75 |        0 |      80 |      75 | 24-27                                 
 util                                |   79.01 |    68.97 |   47.37 |   79.49 |                                                                        
  binarySearch.ts                    |      75 |       75 |   33.33 |      75 | 3,13,34,42-44  
```

wemeetagain (approved):

PR #1519, opened 2020-09-09
Add eth1 JSON RPC client capable of batched requests

dapplion:
Performing a batch request allows to query block headers orders of magnitude faster. This is necessary for block processing

This PR is a step for https://github.com/ChainSafe/lodestar/pull/1518


mpetrunic (dismissed):

twoeths (dismissed):

packages/lodestar/src/eth1/httpEth1Client.ts
```diff
@@ -0,0 +1,61 @@
...
+  for (let i = fromBlockNumber; i <= toBlockNumber; i++) {
+    blockNumbers.push(i);
+  }
+  return fetchBlocksByNumber(url, blockNumbers, signal);
+}
+
+/**
+ * Fetches an array of block numbers.
+ * Roundtrip times with a good connection and remote provider
+ * - 100 blocks: 1481.319ms
```
twoeths: is this statistic for eth1 mainnet or goerli?

dapplion: Goerli

twoeths: this is not very important, I think we need statistic for mainnet instead since it's way slower (I experienced rinkeby vs mainnet in the past, did not try mainnet vs goerli)

dapplion: Thanks, good to know I'll delete the comment

wemeetagain (dismissed):
It seems somewhat strange that we would need our own json-rpc client, and couldn't piggyback off of ethers, but this is also very simple so :shrug:

wemeetagain:
file conflict

dapplion:
> It seems somewhat strange that we would need our own json-rpc client, and couldn't piggyback off of ethers, but this is also very simple so

ethers does not support batch requests

dapplion:
Rebased to master to fix file conflicts

wemeetagain (approved):

PR #1520, opened 2020-09-09
Add Eth1Blocks cache for eth1 data for block production

dapplion:
Adds a new repository to the DB and it's wrapper `Eth1BlockCache`. Part of https://github.com/ChainSafe/lodestar/pull/1518. See https://github.com/ChainSafe/lodestar/pull/1518 for an overview of the target architecture that this component will be part of.

The current repository `eth1Data` will be deprecated. Eth1 data is best fetched and stored separately: (1) `blockHash` fetched from block headers directly, and (2) `depositCount` and `depositRoot` fetched from the deposit logs directly.


dapplion:
No longer necessary after rethinking https://github.com/ChainSafe/lodestar/pull/1518

PR #1521, opened 2020-09-09
Add Eth1DepositsCache cache for eth1 data for block production

dapplion:
Adds two new repository to the DB and their wrapper Eth1DepositsCache. Part of #1518. See #1518 for an overview of the target architecture that this component will be part of.

The current repositories eth1Data and depositData will be deprecated. Eth1 data is best fetched and stored separately: (1) blockHash fetched from block headers directly, and (2) depositCount and depositRoot fetched from the deposit logs directly. To do so it's necessary to store each deposit log blockNumber and index, so just keeping depositData is not sufficient.

Most of the logic added in this PR is to convert non-consecutive deposit logs into consecutive consumable eth1Data.

### DIFF
--- a/packages/lodestar/src/db/api/beacon/beacon.ts
+++ b/packages/lodestar/src/db/api/beacon/beacon.ts
@@ -16,6 +16,7 @@ import {
   DepositDataRootRepository,
   Eth1DataRepository,
   DepositEventRepository,
+  Eth1DataDepositRepository,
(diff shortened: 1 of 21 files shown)


mpetrunic (approved):
seems good to me

dapplion:
Superseded by https://github.com/ChainSafe/lodestar/pull/1518

PR #1524, opened 2020-09-09
Decouple eth1Vote logic from DB

dapplion:
Abstracts eth1 vote filtering and picking from a specific db implementation and adds tests to its logic.

It's a necessary step in preparation to refactor eth1 data for block production https://github.com/ChainSafe/lodestar/pull/1518


packages/lodestar/test/unit/eth1/utils/eth1Vote.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    ...eth1DataBlock,
+  };
+}
+
+/**
+ * Util: Get a mock timestamp that passes isCandidateBlock validation
+ * @param config
+ * @param state
+ */
+function getTimestampInRage(config: IBeaconConfig, state: TreeBacked<BeaconState>): number {
```
mpetrunic:
```suggestion
function getTimestampInRange(config: IBeaconConfig, state: TreeBacked<BeaconState>): number {
```

wemeetagain (approved):

Issue #1466, opened 2020-08-28
Track open handles on close


wemeetagain:
Currently, the beacon node hangs on close because several handles are still open. This is happening because some of our modules to not exit cleanly. (Either event handlers, open promises, etc)

We should use `wtf-node` or similar to track open handles, document all open handles, and ensure we're closing everything properly.

dapplion:
Are all open handlers closed for good on multiple modes of operation of the CLI?

mpetrunic:
probably not, it got autoclosed when I've dragged pr into closed column

3xtr4t3rr3str14l:
it's been a long time since i've ran `wtf-node` on the codebase, but i guess this isn't an issue anymore since we aren't using `yarn` to run the beacon node anymore?

wemeetagain:
We rearranged how the node is started/stopped. No longer a problem, we have a clean stop.

PR #1525, opened 2020-09-09
clear timeouts in network

3xtr4t3rr3str14l:
part of #1466 
found via wtfnode that there were a couple of timeouts that weren't being cleared in network.  this takes care of that.

### DIFF
--- a/packages/lodestar/src/network/reqResp.ts
+++ b/packages/lodestar/src/network/reqResp.ts
@@ -221,7 +221,6 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements 
   ): (source: AsyncIterable<IValidatedRequestBody>) => AsyncGenerator<IResponseChunk> {
     const getResponse = this.getResponse;
     const config = this.config;
-    const signal = this.controller?.signal;
     return (source) => {
       return (async function* () {
         for await (const request of source) {
@@ -232,7 +231,7 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements 
               body: encodeP2pErrorMessage(config, "Invalid Request"),
             };
           } else {
-            yield* getResponse(peerId, method, encoding, request.body, signal);
+            yield* getResponse(peerId, method, encoding, request.body);
           }
           return;
         }
@@ -244,9 +243,9 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements 
     peerId: PeerId,
     method: Method,
     encoding: ReqRespEncoding,
-    request?: RequestBody,
-    signal?: AbortSignal
+    request?: RequestBody
   ): AsyncIterable<IResponseChunk> => {
+    const signal = this.controller?.signal;
     const requestId = randomRequestId();
     this.logger.verbose(`receive ${method} request from ${peerId.toB58String()}`, {requestId, encoding});
     // eslint-disable-next-line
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/network/util.ts
```diff
@@ -139,8 +139,10 @@ export async function dialProtocol(
   const timer = setTimeout(() => {
     abortController.abort();
   }, timeout);
+  const signal = abortController.signal;
+  signal?.addEventListener("abort", () => clearTimeout(timer), {once: true});
```
twoeths: controller is aborted at timeout so there's no point to do `clearTimeout() ` after a timeout, lets node handle it

packages/lodestar/src/network/reqResp.ts
```diff
@@ -240,7 +244,8 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
     peerId: PeerId,
     method: Method,
     encoding: ReqRespEncoding,
-    request?: RequestBody
+    request?: RequestBody,
+    signal?: AbortSignal
```
twoeths: can we access `signal` directly from `this.controller`? this helps avoid adding a new param here.

packages/lodestar/src/network/util.ts
twoeths: we should use the abort controller in reqResp, not this local `abortController`

packages/lodestar/src/network/reqResp.ts
3xtr4t3rr3str14l: done

packages/lodestar/src/network/util.ts
3xtr4t3rr3str14l: > we should use the abort controller in reqResp, not this local abortController

done

packages/lodestar/src/network/util.ts
```diff
@@ -133,14 +133,16 @@ export async function dialProtocol(
...
+  signal: AbortSignal | undefined
 ): ReturnType<LibP2p["dialProtocol"]> {
   const abortController = new AbortController();
   const timer = setTimeout(() => {
     abortController.abort();
   }, timeout);
+  signal?.addEventListener("abort", () => clearTimeout(timer), {once: true});
   try {
-    const conn = await libp2p.dialProtocol(peerId, protocol, {signal: abortController.signal} as object);
+    const conn = await libp2p.dialProtocol(peerId, protocol, {signal} as object);
```
twoeths: I think we want `abortController.signal` as original

3xtr4t3rr3str14l: ah yeah, good catch

3xtr4t3rr3str14l: reverted

twoeths (dismissed):

packages/lodestar/src/network/util.ts
```diff
@@ -133,12 +133,14 @@ export async function dialProtocol(
   libp2p: LibP2p,
   peerId: PeerId,
   protocol: string,
-  timeout: number
+  timeout: number,
+  signal: AbortSignal | undefined
```
mpetrunic:
```suggestion
  signal?: AbortSignal
```

mpetrunic (dismissed):
Other  than minor type comment, LGTM

packages/lodestar/src/network/util.ts
```diff
@@ -133,12 +133,14 @@ export async function dialProtocol(
...
   protocol: string,
-  timeout: number
+  timeout: number,
+  signal?: AbortSignal
 ): ReturnType<LibP2p["dialProtocol"]> {
   const abortController = new AbortController();
   const timer = setTimeout(() => {
     abortController.abort();
   }, timeout);
+  signal?.addEventListener("abort", () => clearTimeout(timer), {once: true});
```
wemeetagain: I don't think this is right.
If the reqResp controller aborts, we want to also abort any in-progress dials.
Currently, this would STOP the aborting of any in-progress dials in that case (by clearing the timeout that would otherwise abort them).

So this listener should be something like:
```suggestion
  signal?.addEventListener("abort", () => abortController.abort(), {once: true});
```

wemeetagain (approved):

PR #1527, opened 2020-09-10
Disregard unfinalized blocks upon restart

### LINKED ISSUE #1526: Restore Chain: need to make it minimal
After syncing medalla and restart, we have to load 8000 unfinalized blocks and reprocess which takes time and memory, it's redundant because:
+ After chain is started, initial sync redownload same blocks again
+ Some unfinalized blocks are even not in the finalized chain so it's redundant to restore

This leads to some weird issues like head slot is way behind latest slot when I restart.

When we restart, we only need to feed forkchoice with last known state:
+ Justified and finalized nodes are ones from the last known (finalized) state 
+ Get respective blocks and feed to forkchoice: 1 for head, 1 for justified, 1 for finalized
+ Continue with initial sync and regular sync

need confirmation from @wemeetagain @mpetrunic

twoeths:
resolves #1526 
## Current flow
+ Upon restart, load and restore from block 73249 to block 110000 through block processor in order to populate state cache and fork choice
+ Initial sync to start querying from block 73249 again and continue until blocks 110000 which is a duplicate and potentially cause OOM issue
## New flow
+ Upon restart, disregard unfinalized blocks
+ Populate forkchoice from previous finalized slot until current finalized slot to make it head, this is quick
+ Only populate state cache with last finalized state
+ Don't even touch block processor in most cases
+ Initial sync start querying from block 73249 

## Testing
+ Restore: With my database of 110000 blocks, I can go to initial sync quickly without having to load and reprocess from block 73249 till 110000
+ Start from scratch: Starting from a new dir works fine as normal

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -18,9 +18,15 @@ import {
   Uint64,
 } from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
-import {computeEpochAtSlot, computeForkDigest, EpochContext} from "@chainsafe/lodestar-beacon-state-transition";
+import {
+  computeEpochAtSlot,
+  computeForkDigest,
+  EpochContext,
+  ZERO_HASH,
+  GENESIS_EPOCH,
+} from "@chainsafe/lodestar-beacon-state-transition";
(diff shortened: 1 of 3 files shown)


twoeths:
#1517  will resolve this issue

PR #1530, opened 2020-09-11
tidying network

3xtr4t3rr3str14l:
-clean up null checks in reqResp
-create handleResponses function for use in sendRequest of reqResp
-check for type safety of slow before processing slots


packages/lodestar/src/network/reqResp.ts
```diff
@@ -131,7 +132,7 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
       id,
       err,
       (async function* () {
-        if (response !== null && response !== undefined) {
+        if (notNullish(response)) {
```
dapplion: I created `notNullish` to filter items in an array that contain nullish items. IMO here it's best to just use the original form because it's more obvious and transparent. Future code readers now have to worry about what is `notNullish` actually doing.

3xtr4t3rr3str14l: gotcha.  i replaced the places where i used notNullish with !isNil from lodash instead

packages/lodestar/package.json
```diff
@@ -82,6 +83,7 @@
     "libp2p-noise": "^1.1.2",
     "libp2p-pubsub": "^0.4.7",
     "libp2p-tcp": "^0.14.6",
+    "lodash": "^4.17.20",
```
wemeetagain: can you use this package: https://www.npmjs.com/package/lodash.isnil

3xtr4t3rr3str14l: done

packages/lodestar/package.json
```diff
@@ -82,6 +83,7 @@
     "libp2p-noise": "^1.1.2",
     "libp2p-pubsub": "^0.4.7",
     "libp2p-tcp": "^0.14.6",
+    "lodash.isnil": "^4.0.0",
```
dapplion: This package is literally one line of code:
```ts
function isNil(value: any): value is null | undefined {
      return value == null;
}
```
https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L11973
Do we really have to install a new package for this? Seems like a big stretch

packages/lodestar/src/network/reqResp.ts
```diff
@@ -132,7 +133,7 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
       id,
       err,
       (async function* () {
-        if (response !== null && response !== undefined) {
+        if (!isNil(response)) {
```
dapplion: See comment above https://github.com/ChainSafe/lodestar/pull/1530/files#r488414419
This could be replaced with 
```ts
if (response != null)
```
and it would be equivalent

packages/lodestar/package.json
3xtr4t3rr3str14l: my original thinking was that maybe we would need lodash in this package in the future (and we are already using it in lodestar-cli), but then again, that's just a hypothetical.  i'll change it to this for now

packages/lodestar/src/network/reqResp.ts
3xtr4t3rr3str14l: done

packages/lodestar/package.json
3xtr4t3rr3str14l: done

packages/lodestar/package.json
```diff
@@ -53,6 +53,7 @@
     "@chainsafe/ssz": "^0.6.13",
     "@types/async": "^3.0.1",
     "@types/datastore-level": "^1.1.1",
+    "@types/lodash.isnil": "^4.0.6",
```
wemeetagain: I don't think we need this anymore

packages/lodestar/src/network/gossip/utils.ts
```diff
@@ -92,7 +92,7 @@ export async function getBlockStateContext(
   if (!stateEpochCtx) return null;
   //only advance state transition if asked for future block
   slot = slot ?? parentSummary.slot;
-  if (stateEpochCtx.state.slot < slot) {
+  if (slot && stateEpochCtx.state.slot < slot) {
```
wemeetagain: I don't think this is necessary

wemeetagain (commented):
a few extraneous things remaining, other than that looks good

packages/lodestar/src/network/gossip/utils.ts
3xtr4t3rr3str14l: done

packages/lodestar/package.json
3xtr4t3rr3str14l: done

mpetrunic (approved):
lgtm

wemeetagain (approved):

PR #1536, opened 2020-09-13
Use it-all util

dapplion:
Since we are already using functions from the it-* family, it makes sense to use it-all instead of writing its verbose version in multiple places


wemeetagain (approved):
Cool :+1:

PR #1539, opened 2020-09-14
Start gossip on boot

### LINKED ISSUE #1531: Start gossip protocol when network starts
**Describe the bug**
I believe we're being dropped by lighthouse because we don't 'support' gossipsub until we get past initial sync.

**Expected behavior**
We should start gossipsub **but not subscribe to any topics** on network start.
Then when we finish initial sync, we can subscribe to topics.
This will allow us to stay connected to clients which require peers to be gossipsub peers while we sync.

wemeetagain:
Resolves #1531

### DIFF
--- a/packages/lodestar/src/network/gossip/gossip.ts
+++ b/packages/lodestar/src/network/gossip/gossip.ts
@@ -37,9 +37,9 @@ import {toHexString} from "@chainsafe/ssz";
 export type GossipHandlerFn = (this: Gossip, obj: GossipObject) => void;
 
 export class Gossip extends (EventEmitter as {new (): GossipEventEmitter}) implements IGossip {
+  public readonly pubsub: IGossipSub;
   protected readonly opts: INetworkOptions;
   protected readonly config: IBeaconConfig;
-  protected readonly pubsub: IGossipSub;
   protected readonly chain: IBeaconChain;
   protected readonly logger: ILogger;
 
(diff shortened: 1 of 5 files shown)


packages/lodestar/src/network/network.ts
```diff
@@ -78,13 +78,15 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
     await this.reqResp.start();
     const enr = this.getEnr();
     await this.metadata.start(enr!);
+    await this.gossip.pubsub.start();
```
twoeths: How about having `pubsub` as part of network instead of gossip? So `pubsub` could be created/started/stopped inside network, gossip just use it. We have `pubsub` as part of `IGossipModules` anyway.

So this could become: `this.pubsub.start()`

wemeetagain: :+1: sounds good

wemeetagain: In fact, libp2p already comes with a pubsub attached! We might look into just using that one in a later PR

twoeths (dismissed):

packages/lodestar/src/network/gossip/gossip.ts
```diff
@@ -63,7 +63,6 @@ export class Gossip extends (EventEmitter as {new (): GossipEventEmitter}) imple
   }
 
   public async start(): Promise<void> {
-    await this.pubsub.start();
```
twoeths: still need to keep this?

packages/lodestar/src/network/gossip/gossip.ts
```diff
@@ -75,7 +74,6 @@ export class Gossip extends (EventEmitter as {new (): GossipEventEmitter}) imple
     this.emit("gossip:stop");
     this.unregisterHandlers();
     this.chain.emitter.removeListener("forkDigest", this.handleForkDigest);
-    await this.pubsub.stop();
```
twoeths: same

twoeths (dismissed):

twoeths (approved):

PR #1540, opened 2020-09-14
Separate fork choice

wemeetagain:
part of #1517 

state regeneration saved for a separate PR
other updates already made in in #1537 and #1538

- Block processing refactored slightly
  - `runStateTransition` and other utilities moved to `src/chain/blocks/stateTransition.ts`
- Block/state archival tasks refactored
  - they are now both triggered by the `forkChoice:finalized` event, the `forkChoice:pruned` event is no more.
  - we also prune state caches, triggered by the `checkpoint` event, but not in a task as its a one-liner
  - block archival no longer loads _all_ blocks in `block` repository into memory at once, instead streams
  - block archival strategy, run before fork choice pruning
    - stream all blocks in `block` repository, one by one
    - if the block is more recent than the finalized block, ignore
    - if the finalized block is a descendant of the block, store in the blockArchive
    - delete all un-ignored blocks
  - state archival now archives the finalized state from `checkpointStateCache` (these will be candidates for weak subjectivity :) )
- Add `@chainsafe/lodestar-fork-choice` package
  - split into two 'levels', protoArray, and forkChoice
  - protoArray - the block dag, not much else
  - forkChoice
    - provides an interface very close to the spec
    - includes an `IForkChoiceStore`, very similar to `Store` in the spec -- this is implemented in the lodestar package with knowledge of the `ChainEventEmitter`
    - stores votes, balances, queued attestations
  - Attestation weighting is completely different than before
    - Instead of feeding attestations in with weights attached per attestation, we feed in something like a `LatestMessage` from the spec.
    - attestations are reweighted all-at-once, using `forkChoice.updateBalances`. Using `epochContext.epochProcess`, this is fast, lots of the work was already done as part of epoch processing.
    - attestation reweighting is triggered by the `forkChoice:justified` event.


packages/lodestar/src/chain/chain.ts
```diff
@@ -328,35 +331,51 @@ export class BeaconChain implements IBeaconChain {
...
+      if (this.config.types.Root.equals(blockHeader.stateRoot, ZERO_HASH)) {
+        blockHeader.stateRoot = anchorState.hashTreeRoot();
+      }
+      const finalizedCheckpoint = {
+        root: this.config.types.BeaconBlockHeader.hashTreeRoot(blockHeader),
+        epoch: computeEpochAtSlot(this.config, anchorState.slot),
+      };
+      const justifiedCheckpoint = {
+        root: this.config.types.BeaconBlockHeader.hashTreeRoot(blockHeader),
+        epoch: computeEpochAtSlot(this.config, anchorState.slot) + 1,
```
twoeths, line 370: what's the reason to add 1 to justified checkpoint epoch?

This is most likely not the correct justified checkpoint for a specific state but a reasonable one to start with and forkchoice will update to the correct one later.

I think if we just leave this justified checkpoint `epoch` the same to finalized checkpoint `epoch`, we'll update to correct justified checkpoint sooner due to the condition `if state.current_justified_checkpoint.epoch > store.justified_checkpoint.epoch:` in https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/fork-choice.md#on_block

packages/lodestar/src/chain/chain.ts
```diff
@@ -328,35 +331,51 @@ export class BeaconChain implements IBeaconChain {
...
-      blockHeader.stateRoot = this.config.types.BeaconState.hashTreeRoot(anchorState);
-      blockRoot = this.config.types.BeaconBlockHeader.hashTreeRoot(blockHeader);
-      justifiedCheckpoint = anchorState.currentJustifiedCheckpoint;
-      finalizedCheckpoint = anchorState.finalizedCheckpoint;
+      if (this.config.types.Root.equals(blockHeader.stateRoot, ZERO_HASH)) {
+        blockHeader.stateRoot = anchorState.hashTreeRoot();
+      }
+      const finalizedCheckpoint = {
+        root: this.config.types.BeaconBlockHeader.hashTreeRoot(blockHeader),
+        epoch: computeEpochAtSlot(this.config, anchorState.slot),
```
twoeths, line 362: I'm afraid this is `computeEpochAtSlot(this.config, anchorState.slot) + n` where `n` is unknown if 1st slot of finalized epoch is a skipped slot, it'll use "last known blockRoot" in that case.

I used to face this issue in my weak subjectivity state PR, most peers say goodbye to us due to incorrect finalized checkpoint.

twoeths: ignore the above comment, I think it should work bc the new ArchiveState uses CheckpointStateCache instead of the regular state cache so we're fine but correct me if I'm wrong.

packages/lodestar/src/tasks/index.ts
```diff
@@ -70,8 +74,27 @@ export class TasksService implements IService {
     await this.interopSubnetsTask.stop();
   };
 
-  private handleFinalizedCheckpointChores = async (finalized: BlockSummary, pruned: BlockSummary[]): Promise<void> => {
-    await new ArchiveBlocksTask(this.config, {db: this.db, logger: this.logger}, finalized, pruned).run();
-    await new ArchiveStatesTask(this.config, {db: this.db, logger: this.logger}, finalized, pruned).run();
+  private onFinalizedCheckpoint = async (finalized: Checkpoint): Promise<void> => {
+    await new ArchiveStatesTask(this.config, {db: this.db, logger: this.logger}, finalized).run();
```
twoeths, line 83: we want to run ArchiveStateTask after ArchiveBlockTask so that when we restart, we start from last finalized state and we want to make sure finalized blocks stored in db already

packages/lodestar/src/chain/chain.ts
wemeetagain: The state archival task has been updated so that we only _store_ valid checkpoint states (state where state.slot % SLOTS_PER_EPOCH === 0) to db. Now the epoch will always be correct, if we use those checkpoint states.

packages/lodestar/src/chain/chain.ts
wemeetagain: There's an edge-case that we may hit if we set the finalized and justified epochs the same.
(I ran into it syncing medalla)
There will be a period upon restore where we will be feeding in blocks, but those blocks will not be the 'head' of the chain. This will be the case until we process the block that actually finalized the epoch of the finalized epoch we restore with.
The issue we can run into is that while we are getting to this epoch, we may create a head that triggers ERR_INVALID_BEST_NODE, which can happen if the best node (which falls back to the justified node) is not viable for head (doesn't have the same finalized and justified and finalized epochs as is stored in the fork choice). This happens when we set the justified checkpoint, but don't set the finalized checkpoint, and then the justified node no longer matches the fork choice justified and finalized epochs.

mpetrunic: can you add this as comment explaining +1 :)

packages/lodestar-fork-choice/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "@chainsafe/lodestar-config": "^0.11.0",
+    "@chainsafe/lodestar-params": "^0.11.0",
+    "@chainsafe/lodestar-types": "^0.11.0",
+    "@chainsafe/lodestar-utils": "^0.11.0",
+    "@chainsafe/ssz": "^0.6.13"
+  },
+  "devDependencies": {
+    "chai": "^4.2.0",
+    "chai-as-promised": "^7.1.1",
+    "mocha": "^6.2.2"
```
mpetrunic: do we need those since they are already in root package.json.
If we keep them all in one place, it's easier to keep same version in every package

packages/lodestar-fork-choice/src/protoArray/interface.ts
```diff
@@ -0,0 +1,59 @@
+/* eslint-disable @typescript-eslint/interface-name-prefix */
```
mpetrunic: Don't think you need this since all interfaces have "I" prefix.

packages/lodestar-fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -0,0 +1,771 @@
...
+ * - Management of validators latest messages and balances
+ * - Management of the justified/finalized checkpoints as seen by fork choice
+ * - Queuing of attestations from the current slot
+ *
+ * This class MUST be used with the following considerations:
+ *
+ * - Time is not updated automatically, updateTime MUST be called every slot
+ * - Justified balances are not updated automatically, updateBalances MUST be called when Store justifiedCheckpoint is updated
+ */
+export class ForkChoice {
```
mpetrunic: Why not create IForkChoice interface and rename this class to ProtoArrayForkChoice?

I see multiple pros for that:
- when we initialize forkchoice in code we know exactly what implementation we are using
- we can add different implementations (faster, slower etc) without breaking api
- it's easier to stub it in tests

 If at some point we wan't to experiment with different implementations

packages/lodestar/test/unit/api/impl/beacon/blocks/utils.test.ts
```diff
@@ -14,18 +14,18 @@ use(chaiAsPromised);
...
 describe("block api utils", function () {
   describe("resolveBlockId", function () {
-    let forkChoiceStub: SinonStubbedInstance<ILMDGHOST>;
+    let forkChoiceStub: SinonStubbedInstance<ForkChoice> & ForkChoice;
 
     let dbStub: StubbedBeaconDb;
 
     beforeEach(function () {
-      forkChoiceStub = sinon.createStubInstance(ArrayDagLMDGHOST);
+      forkChoiceStub = sinon.createStubInstance(ForkChoice) as SinonStubbedInstance<ForkChoice> & ForkChoice;
```
mpetrunic: with interface, you wouldn't have to cast like this everywhere

wemeetagain:
Passing tests :100:

twoeths (approved):
💯 🥇

PR #1541, opened 2020-09-15
Migrate to libp2p 0.29

### LINKED ISSUE #1206: Update to libp2p 0.29.x, gossipsubv1.1
It's that time again, time to start looking into upgrading from libp2p 0.28.x to libp2p 0.29.x. At the same time, we will need to upgrade gossipsub (0.5.x).
Even tho neither libp2p 0.29.x nor gossipsub 0.5.x have been published, we may still begin the process of upgrading using 
libp2p/js-libp2p branch 0.29.x and ChainSafe/js-libp2p-pubsub branch gsv1.1

Seemingly all major changes to gossipsub have been merged in, with only additional testing and non-breaking changes left, see:
https://github.com/ChainSafe/js-libp2p-gossipsub/issues/73

Upgrade libp2p stack, be ready for the official 0.29.x and 0.5.x release.

**Additional context**
I believe the main difference, for us, between versions is in gossipsub, not much in libp2p core:
- peer scoring - we can pass in [`scoreParams`](https://github.com/ChainSafe/js-libp2p-gossipsub/blob/gsv1.1/ts/index.ts#L33) and `scoreThresholds` when they get defined in the spec
- the `Message` object is now only used minimally in pubsub components. Instead [`InMessage`](https://github.com/ChainSafe/js-libp2p-gossipsub/blob/gsv1.1/ts/message/index.ts#L69) is used throughout. Everywhere that operates on a "message" is operating on an `InMessage`. Gossipsub emits `InMessage`s on subscribed topics. Note it now has a `receivedFrom` field, the string peer id of the direct sender.
- topic validators - instead of returning an `ExtendedValidatorResult`, topic validator functions now `throw` an Error. Valid errors are defined [here](https://github.com/ChainSafe/js-libp2p-gossipsub/blob/gsv1.1/ts/constants.ts#L216).  Example [here](https://github.com/ChainSafe/js-libp2p-gossipsub/blob/gsv1.1/test/pubsub.spec.js#L148).  The function signature is also changed, now `(topic: string, message: InMessage): void` - peerId is no longer a parameter, instead `message.receivedFrom` should be used.
- Because of the validation refactor, we no longer use the `_processTopicValidatorResult` function, its been deleted.

twoeths:
resolves #1206


packages/lodestar/package.json
```diff
@@ -74,17 +74,16 @@
...
-    "libp2p-gossipsub": "^0.4.6",
-    "libp2p-mdns": "^0.14.3",
-    "libp2p-mplex": "^0.9.5",
-    "libp2p-noise": "^1.1.2",
-    "libp2p-pubsub": "^0.4.7",
-    "libp2p-tcp": "^0.14.6",
-    "multiaddr": "^7.4.3",
+    "libp2p": "^0.29.0",
+    "libp2p-bootstrap": "^0.12.0",
+    "libp2p-gossipsub": "/Users/tuyennguyen/Projects/workshop/gossipsub-js",
```
twoeths: waiting for `gossipsub-js` update with `libp2p-interfaces` types

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -70,16 +69,22 @@ export class LodestarGossipsub extends Gossipsub {
     }
   }
 
-  public async validate(rawMessage: Message, peer: Peer): Promise<boolean> {
-    const message: ILodestarGossipMessage = normalizeInRpcMessage(rawMessage);
+  /**
+   * Override https://github.com/ChainSafe/js-libp2p-gossipsub/blob/v0.6.0/ts/index.ts#L442
+   * Should throw ERR_TOPIC_VALIDATOR_IGNORE or ERR_TOPIC_VALIDATOR_REJECT
+   * If we don't throw it means accept
```
wemeetagain, line 102: We'll need to update the topic validators in `src/network/gossip/validation/*` to do this, rather than rely on transformation after the fact. Results are no longer collected: https://github.com/libp2p/js-libp2p-interfaces/blob/master/src/pubsub/index.js#L515

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -70,16 +69,22 @@ export class LodestarGossipsub extends Gossipsub {
...
     assert.true(Boolean(message.topicIDs), "topicIds is not defined");
     assert.equal(message.topicIDs.length, 1, "topicIds array must contain one item");
     if (!message.data) throw Error("Message has not data");
     assert.lte(message.data.length, GOSSIP_MAX_SIZE, "Message exceeds byte size limit");
     const topic = message.topicIDs[0];
     // avoid duplicate
     if (this.transformedObjects.get(message.messageId)) {
-      return false;
+      throw new GossipValidationError(ERR_TOPIC_VALIDATOR_IGNORE);
     }
```
wemeetagain: gossipsub should already be handling duplicates. I think we can `assert.false(this.transformedObjects.get(...))` here

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -1,24 +1,24 @@
-import PeerId from "peer-id";
 import Gossipsub from "libp2p-gossipsub";
```
wemeetagain, line 1: fyi https://github.com/ethereum/eth2.0-specs/pull/2044
it isn't in a release yet, but we'll need to support it soon enough.
This might require an update to gossipsub, not sure. Currently, topics are `string`, and I think there may be issues with this one way or another.

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -70,16 +69,22 @@ export class LodestarGossipsub extends Gossipsub {
...
 
-  public async validate(rawMessage: Message, peer: Peer): Promise<boolean> {
-    const message: ILodestarGossipMessage = normalizeInRpcMessage(rawMessage);
+  /**
+   * Override https://github.com/ChainSafe/js-libp2p-gossipsub/blob/v0.6.0/ts/index.ts#L442
+   * Should throw ERR_TOPIC_VALIDATOR_IGNORE or ERR_TOPIC_VALIDATOR_REJECT
+   * If we don't throw it means accept
+   * @param rawMessage
+   */
+  public async validate(rawMessage: InMessage): Promise<void> {
```
wemeetagain: We need to rely on super.validate, and call it bc we track rejected / ignored messages in super.validate.
One direction might be to push all validation into the topic validators.

```typescript

async function genericIsValid (message: InMessage): Promise<boolean> {
  return (
    message.topicIDs && message.topicIDs.length === 1 &&
    message.data && message.data.length > GOSSIP_MAX_SIZE
  );
}

async function topicValidator (topic: string, objValidatorFn: (obj: GossipObject) => Promise<void>): (msg: InMessage) => Promise<void> {
  return async (topic: string, msg: InMessage): Promise<void> => {
    if (!genericIsValid(msg)) {
      throw new GossipValidationError(ERR_TOPIC_VALIDATOR_REJECT);
    }
    try {
      const {object} = deserializeGossipMessage(topic, message);
      await objValidatorFn(object);
    } catch (e) {
      throw new GossipValidationError(ERR_TOPIC_VALIDATOR_REJECT);
    }
  }
}

// in gossip.ts / gossipsub.ts, set the topic validators
// they fail fast here: https://github.com/libp2p/js-libp2p-interfaces/blob/master/src/pubsub/index.js#L515
this.topicValidators.set(
  getGossipTopic(BLOCK),
  topicValidator(getGossipTopic(BLOCK), validateGossipBlock)
);
```
Then we wouldn't override `validate` at all, just rely on the underlying implementation.
This could be cleaned up, just showing the idea.

packages/lodestar/src/network/gossip/gossipsub.ts
twoeths: I only see the way to calculate message-id there which is not a big change, are there other PRs about topics? Anyway it's good to notice that, we should focus on libp2p migration to make sure gossip works fine with medalla first.
In the mean time, topics are still string in dev branch.

packages/lodestar/src/network/gossip/gossip.ts
```diff
@@ -56,14 +57,15 @@ export class Gossip extends (EventEmitter as {new (): GossipEventEmitter}) imple
     // need to improve Gossipsub type to implement EventEmitter to avoid this cast
     this.pubsub =
       pubsub ||
-      ((new LodestarGossipsub(config, validator, this.logger, libp2p.peerId, libp2p.registrar, {
+      ((new LodestarGossipsub(config, validator, this.logger, (libp2p as unknown) as Libp2p, {
```
mpetrunic: I don't think we should do this. We need to fix this

packages/lodestar/src/network/gossip/gossip.ts
```diff
@@ -33,6 +33,7 @@ import {IBeaconChain} from "../../chain";
 import {computeEpochAtSlot, computeForkDigest} from "@chainsafe/lodestar-beacon-state-transition";
 import {GossipEncoding} from "./encoding";
 import {toHexString} from "@chainsafe/ssz";
+import {Libp2p} from "libp2p-gossipsub/src/interfaces";
```
mpetrunic: why do we have separate interface in libp2p-gossipsub (should be reused from libp2p-ts?)

packages/lodestar/src/network/gossip/interface.ts
```diff
@@ -105,3 +106,11 @@ export type GossipMessageValidatorFn = (message: GossipObject, subnet?: number)
 export interface ILodestarGossipMessage extends Message {
   messageId: string;
 }
+
+export class GossipValidationError extends Error {
```
mpetrunic: maybe separate file?

packages/lodestar/src/network/gossip/interface.ts
```diff
@@ -105,3 +106,11 @@ export type GossipMessageValidatorFn = (message: GossipObject, subnet?: number)
...
+  public code: string;
+  constructor(code: string, message?: string) {
```
mpetrunic: ain't this code suppose to be enum?

twoeths: I got error "Only numeric enums can have computed members" if I import error constants from gossipsub to make an enum in lodestar. I think gossipsub can make an enum instead if we want to.

packages/lodestar/src/network/gossip/gossip.ts
twoeths: gossipsub has its own libp2p type, libp2p-interface uses `any`. Also gossipsub uses peer-id 0.14.0 and lodestar uses 0.14.1 (drived by other dependencies) and it causes type conflict with this minor version change.
I heard that libp2p has a plan to generate its type from jsdoc like libp2p-interface, can we wait for it? Temporarily made it as ts-ignore to avoid the confusion and version conflict issue.

mpetrunic: > I heard that libp2p has a plan to generate its type from jsdoc like libp2p-interface, can we wait for it?

Nope, it's last one that's gonna receive this update and it's probably gonna take a long time. Why aren't we using libp2p type from ts-libp2p? Since libp2p-interfaces has any type it's gonna work

twoeths: gossipsub-js uses its own libp2p type. @wemeetagain can we use libp2p-ts libp2p type in gossipsub-js?

packages/lodestar/src/network/gossip/gossip.ts
```diff
@@ -56,14 +56,16 @@ export class Gossip extends (EventEmitter as {new (): GossipEventEmitter}) imple
     // need to improve Gossipsub type to implement EventEmitter to avoid this cast
     this.pubsub =
       pubsub ||
-      ((new LodestarGossipsub(config, validator, this.logger, libp2p.peerId, libp2p.registrar, {
+      // @ts-ignore
```
3xtr4t3rr3str14l, line 59: Could you find a way to avoid using ts-ignore here?

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -70,71 +75,97 @@ export class LodestarGossipsub extends Gossipsub {
...
     });
   }
 
-  publish(topic: string, data: Buffer): void {
-    const encoding = getTopicEncoding(topic);
+  /**
+   * Override https://github.com/ChainSafe/js-libp2p-gossipsub/blob/v0.6.3/ts/index.ts#L1034
+   */
+  public _publish(message: InMessage): Promise<void> {
+    assert.true(message.topicIDs && message.topicIDs.length === 1, "lodestar only support 1 topic per message");
```
3xtr4t3rr3str14l: minor grammar issue, but should be "supports" instead of "support"

packages/lodestar/test/unit/network/gossip/gossipsub.test.ts
```diff
@@ -1,66 +1,73 @@
...
-    const registrar = {
-      handle: (): null => null,
-      register: (): string => "",
-      unregister: (): boolean => false,
-    };
-    const peerId = await createPeerId();
-    gossipSub = new LodestarGossipsub(config, validator, new WinstonLogger(), peerId, registrar, {});
+    const multiaddr = "/ip4/127.0.0.1/tcp/0";
+    const libp2p = await createNode(multiaddr);
+    gossipSub = new LodestarGossipsub(config, validator, new WinstonLogger(), (libp2p as unknown) as Libp2p, {});
```
3xtr4t3rr3str14l, line 38: Can you try to typecast without the `unknown`?

packages/lodestar/src/network/gossip/gossip.ts
3xtr4t3rr3str14l: oh, just saw this comment that explains why it's here
https://github.com/ChainSafe/lodestar/pull/1541#discussion_r499922451

packages/lodestar/test/unit/network/gossip/gossipsub.test.ts
3xtr4t3rr3str14l: Nvm, related to above comment

https://github.com/ChainSafe/lodestar/pull/1541#discussion_r499922451

wemeetagain (approved):
LGTM! :tada: 
We can further tweak things (eg: gossip validators, message ID) in future PRs

PR #1545, opened 2020-09-17
Block providers score tracking

### LINKED ISSUE #1460: Track peer block provider score
- peer scores should be stored in peerstore metabook
- we can have something simple to start:
   - every peer starts with score 1
   - every block_by_root or block_by_range successfully requests adds 0.1
   - every bad response or response timeout reduces score by 0.2
- in sync sort peers by score and take best N

Later on we can add more complex stuff like score decay over time so we give peers another chance after some time etc.

mpetrunic:
TODO: add more unit tests after reviews

﻿- add score provider tracker
- keep track of block provider scores
- use peer block score tracking in initial sync
- post merge fixes
- fix tests

resolves #1460


packages/lodestar/src/sync/initial/fast/index.ts
```diff
@@ -220,18 +221,10 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
...
-      .map((peer) => peer.id)
-      .reduce((validPeers: PeerId[], peer: PeerId) => {
-        const status = this.network.peerMetadata.getStatus(peer);
-        if (status && status.finalizedEpoch >= this.targetCheckpoint!.epoch) {
-          validPeers.push(peer);
-        }
-        return validPeers;
-      }, [] as PeerId[]);
+    const targetCheckpoint = this.targetCheckpoint;
+    return getSyncPeers(this.network, (peer) => {
```
twoeths, line 228: do we want to apply this to regular sync `naive.ts` strategy too? probably in `checkBestPeer()`

mpetrunic: yeah, makes sense

wemeetagain (commented):
This is really awesome, simple and easy.

The only thing I would change is the references to "blockProvider". Likely, we will want to add more scoring events and it doesn't seem worth it to segregate the scoring modules into different domain-specific "scorers". Rather, we should just have a IScoreTracker, which can be fed any variety of ScoreEvents about a peer. A sophisticated score tracker may be aggregating scores from several underlying params (eg: like how gossipsub v1.1 tracks time-in-mesh separately from message-deliveries), but the interface doesn't need to worry about that, and we can just append to the list of ScoreEvents.

eg:
`BlockProviderScoreEvent` => `ScoreEvent`
`IBlockProviderScoreTracker` => `IScoreTracker`
`SimpleBlockProviderScoreTracker` => `SimpleScoreTracker` (or `ScoreTracker`)
`blockProviderScores` => `scores`

mpetrunic:
> This is really awesome, simple and easy.
> 
> The only thing I would change is the references to "blockProvider". Likely, we will want to add more scoring events and it doesn't seem worth it to segregate the scoring modules into different domain-specific "scorers". 

I originally thought the same, but opted for specific scorers in case in future we have optional state rpc or something. But agree, I will make it more generic, we can easily change it later if need be.

packages/lodestar/src/network/error.ts
```diff
@@ -11,3 +13,19 @@ export class RpcError extends Error {
...
   }
 }
+
+export const RESPONSE_TIMEOUT_ERR = "response timeout";
+
+export function isResponseTimeout(err: Error): boolean {
+  return err.message.includes(RESPONSE_TIMEOUT_ERR);
+}
+
+export function updateBlockProviderErrorScore(scoreTracker: IRpcScoreTracker, peer: PeerId, e: Error): void {
```
wemeetagain: one last reference to block provider, `updateRpcScore`?

mpetrunic: dammit

wemeetagain (approved):

PR #1547, opened 2020-09-18
Add state regenerator

wemeetagain:
- `IStateRegenerator` interface has been added, along with implementations `StateRegenerator` and `QueuedStateRegenerator`, - `src/chain/regen/`
  - it can regenerate states from blocks _that have already been processed_ , given `forkChoice`, `stateCache`, `checkpointStateCache`, and `db.blocks`
  - it intelligently attempts to perform the smallest amount of regeneration necessary to regen the requested state. To that end, it exposes 4 different methods.
  - it reuses some block processing machinery (`emitCheckpointEvent`, `runStateTransition`) so as to enable re-seeding the checkpoint cache (using the emitted events)
  - it throws a `RegenError` with well-defined error codes and types, this allows us to tighten the interface to disallow nulls.
  - it queues requests, one regeneration allowed at a time
  - `getPreState(block: BeaconBlock): Promise<ITreeStateContext>` - similar to `getPreState` that we were using in our block processor. It can request a more recent checkpoint state if the block occurs after many skip-slots, skipping epoch transitions :) 
  - `getCheckpointState(cp: Checkpoint): Promise<ITreeStateContext>` - returns the checkpoint state. It minimizes regeneration by first searching the checkpoint cache for checkpoint states with cp.root, which may then just be `processSlots`ed. Otherwise regens the 'ol-fashioned way', by state root.
  - `getBlockSlotState(blockRoot: Root, slot: Slot): Promise<ITreeStateContext>` - returns the block state at the requested slot. This is a lot like `getCheckpointState`, but more flexible, especially useful for dealing with the head state, eg: `regen.getBlockSlotState(forkChoice.getHeadRoot(), clock.currentSlot)`. This is nice, because in periods of skip slots, we will only process the checkpoint with the blockRoot once.
  - `getState(stateRoot: Root): Promise<ITreeStateContext>` - returns the state whose root is stateRoot. This is where the actual block replay happens. It iterates backwards thru the block tree from the requested state, to the first state that we have cached, (searching the checkpointStateCache as well!), then replaying the blocks that were iterated thru to the requested state, once a state has been found.

### DIFF
--- a/packages/lodestar/src/api/impl/validator/validator.ts
+++ b/packages/lodestar/src/api/impl/validator/validator.ts
@@ -29,7 +29,6 @@ import {
   getDomain,
 } from "@chainsafe/lodestar-beacon-state-transition";
 import {assert, ILogger} from "@chainsafe/lodestar-utils";
-import {processSlots} from "@chainsafe/lodestar-beacon-state-transition/lib/fast/slot";
 
 import {IBeaconDb} from "../../../db";
 import {IBeaconChain} from "../../../chain";
@@ -46,6 +45,7 @@ import {ApiError} from "../errors/api";
 import {notNullish} from "../../../util/notNullish";
 import {ApiNamespace, IApiModules} from "../interface";
 import {IValidatorApi} from "./interface";
+import {waitForBlockSlot} from "../../../network/gossip/validation";
 
 export class ValidatorApi implements IValidatorApi {
   public namespace: ApiNamespace;
@@ -98,18 +98,16 @@ export class ValidatorApi implements IValidatorApi {
   public async produceAttestation(validatorPubKey: BLSPubkey, index: CommitteeIndex, slot: Slot): Promise<Attestation> {
     try {
       await this.checkSyncStatus();
-      const [headBlockRoot, {state: headState, epochCtx}] = await Promise.all([
-        this.chain.forkChoice.getHeadRoot(),
-        this.chain.getHeadStateContext(),
-      ]);
+      const headBlock = this.chain.forkChoice.getHead();
+      const {state, epochCtx} = await this.chain.regen.getBlockSlotState(
+        headBlock.blockRoot,
+        this.chain.clock.currentSlot
+      );
       const validatorIndex = epochCtx.pubkey2index.get(validatorPubKey);
       if (validatorIndex === undefined) {
         throw Error(`Validator pubKey ${toHexString(validatorPubKey)} not in epochCtx`);
       }
-      if (headState.slot < slot) {
-        processSlots(epochCtx, headState, slot);
-      }
-      return await assembleAttestation(epochCtx, headState, headBlockRoot, validatorIndex, index, slot);
+      return await assembleAttestation(epochCtx, state, headBlock.blockRoot, validatorIndex, index, slot);
     } catch (e) {
       this.logger.warn(`Failed to produce attestation because: ${e.message}`);
       throw e;
@@ -118,6 +116,8 @@ export class ValidatorApi implements IValidatorApi {
 
   public async publishBlock(signedBlock: SignedBeaconBlock): Promise<void> {
     await this.checkSyncStatus();
+    await waitForBlockSlot(this.config, this.chain.getGenesisTime(), signedBlock.message.slot);
+    this.chain.clock.currentSlot;
     await Promise.all([this.chain.receiveBlock(signedBlock), this.network.gossip.publishBlock(signedBlock)]);
   }
 
@@ -136,16 +136,12 @@ export class ValidatorApi implements IValidatorApi {
   public async getProposerDuties(epoch: Epoch): Promise<ProposerDuty[]> {
     await this.checkSyncStatus();
     assert.gte(epoch, 0, "Epoch must be positive");
-    const {state, epochCtx} = await this.chain.getHeadStateContext();
-    assert.lte(
-      epoch,
-      computeEpochAtSlot(this.config, state.slot) + 2,
-      "Cannot get duties for epoch more than two ahead"
+    assert.lte(epoch, this.chain.clock.currentEpoch + 2, "Cannot get duties for epoch more than two ahead");
+    const {state, epochCtx} = await this.chain.regen.getBlockSlotState(
+      this.chain.forkChoice.getHeadRoot(),
+      this.chain.clock.currentSlot
     );
     const startSlot = computeStartSlotAtEpoch(this.config, epoch);
-    if (state.slot < startSlot) {
-      processSlots(epochCtx, state, startSlot);
-    }
     const duties: ProposerDuty[] = [];
 
     for (let slot = startSlot; slot < startSlot + this.config.params.SLOTS_PER_EPOCH; slot++) {
@@ -158,11 +154,11 @@ export class ValidatorApi implements IValidatorApi {
   public async getAttesterDuties(epoch: number, validatorPubKeys: BLSPubkey[]): Promise<AttesterDuty[]> {
     await this.checkSyncStatus();
     if (!validatorPubKeys || validatorPubKeys.length === 0) throw new Error("No validator to get attester duties");
-    const {epochCtx, state} = await this.chain.getHeadStateContext();
-    const currentSlot = this.chain.clock.currentSlot;
-    if (state.slot < currentSlot) {
-      processSlots(epochCtx, state, currentSlot);
-    }
+    assert.lte(epoch, this.chain.clock.currentEpoch + 2, "Cannot get duties for epoch more than two ahead");
+    const {epochCtx, state} = await this.chain.regen.getBlockSlotState(
+      this.chain.forkChoice.getHeadRoot(),
+      this.chain.clock.currentSlot
+    );
     const validatorIndexes = validatorPubKeys.map((key) => {
       const validatorIndex = epochCtx.pubkey2index.get(key);
       if (validatorIndex === undefined || !Number.isInteger(validatorIndex)) {
--- a/packages/lodestar/src/api/impl/validator/validator.ts
+++ b/packages/lodestar/src/api/impl/validator/validator.ts
@@ -98,10 +98,7 @@ export class ValidatorApi implements IValidatorApi {
     try {
       await this.checkSyncStatus();
       const headBlock = this.chain.forkChoice.getHead();
-      const {state, epochCtx} = await this.chain.regen.getBlockSlotState(
-        headBlock.blockRoot,
-        this.chain.clock.currentSlot
-      );
+      const {state, epochCtx} = await this.chain.getHeadStateContextAtCurrentSlot();
       const validatorIndex = epochCtx.pubkey2index.get(validatorPubKey);
       if (validatorIndex === undefined) {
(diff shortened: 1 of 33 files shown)


packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -119,6 +116,8 @@ export class ValidatorApi implements IValidatorApi {
 
   public async publishBlock(signedBlock: SignedBeaconBlock): Promise<void> {
     await this.checkSyncStatus();
+    await waitForBlockSlot(this.config, this.chain.getGenesisTime(), signedBlock.message.slot);
+    this.chain.clock.currentSlot;
```
mpetrunic: this line seems useless?

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -119,6 +116,8 @@ export class ValidatorApi implements IValidatorApi {
 
   public async publishBlock(signedBlock: SignedBeaconBlock): Promise<void> {
     await this.checkSyncStatus();
+    await waitForBlockSlot(this.config, this.chain.getGenesisTime(), signedBlock.message.slot);
```
mpetrunic: maybe it would be better to have waitForSlot method on chain clock?

It would be nice if we would use chain clock everywhere to get time as it would make it easier to implement some time sync strategy in future

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -137,16 +136,12 @@ export class ValidatorApi implements IValidatorApi {
...
     assert.gte(epoch, 0, "Epoch must be positive");
-    const {state, epochCtx} = await this.chain.getHeadStateContext();
-    assert.lte(
-      epoch,
-      computeEpochAtSlot(this.config, state.slot) + 2,
-      "Cannot get duties for epoch more than two ahead"
+    assert.lte(epoch, this.chain.clock.currentEpoch + 2, "Cannot get duties for epoch more than two ahead");
+    const {state, epochCtx} = await this.chain.regen.getBlockSlotState(
+      this.chain.forkChoice.getHeadRoot(),
+      this.chain.clock.currentSlot
```
mpetrunic: actually, we can optimize stuff a bit here:
- if epoch > currentEpoch - process slots up to start slot of epoch
- else - use headState without slot processing

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -159,11 +154,11 @@ export class ValidatorApi implements IValidatorApi {
...
     if (!validatorPubKeys || validatorPubKeys.length === 0) throw new Error("No validator to get attester duties");
-    const {epochCtx, state} = await this.chain.getHeadStateContext();
-    const currentSlot = this.chain.clock.currentSlot;
-    if (state.slot < currentSlot) {
-      processSlots(epochCtx, state, currentSlot);
-    }
+    assert.lte(epoch, this.chain.clock.currentEpoch + 2, "Cannot get duties for epoch more than two ahead");
+    const {epochCtx, state} = await this.chain.regen.getBlockSlotState(
+      this.chain.forkChoice.getHeadRoot(),
+      this.chain.clock.currentSlot
```
mpetrunic: same as above

packages/lodestar/src/chain/regen/regen.ts
```diff
@@ -0,0 +1,167 @@
...
+  async getState(stateRoot: Root): Promise<ITreeStateContext> {
+    // Trivial case, stateCtx at stateRoot is already cached
+    const cachedStateCtx = await this.db.stateCache.get(stateRoot);
+    if (cachedStateCtx) {
+      return cachedStateCtx;
+    }
+    // Otherwise we have to use the fork choice to traverse backwards, block by block,
+    // searching the state caches
+    // then replay blocks forward to the desired stateRoot
+    const _Root = this.config.types.Root;
```
mpetrunic:
```suggestion
    const rootType = this.config.types.Root;
```
to keep naming consistent

packages/lodestar/src/chain/regen/regen.ts
```diff
@@ -0,0 +1,167 @@
...
+    if (latestCheckpointStateCtx) {
+      // If a checkpoint state exists with the given checkpoint root, it either is in requested epoch
+      // or needs to have empty slots processed until the requested epoch
+      return await processSlotsByCheckpoint(this.emitter, latestCheckpointStateCtx, slot);
+    } else {
+      // Otherwise, use the fork choice to get the stateRoot from block at the checkpoint root
+      // regenerate that state,
+      // then process empty slots until the requested epoch
+      const blockStateCtx = await this.getState(block.stateRoot);
+      return await processSlotsByCheckpoint(this.emitter, blockStateCtx, slot);
```
mpetrunic, line 98: So checkpoint event isn't just cannonical checkpoints?
Since when processing attestation, this method is called which might emit checkpoint events on non cannonical chain.

packages/lodestar/src/util/queue.ts
```diff
@@ -0,0 +1,78 @@
...
+  constructor(code: QueueErrorCode, msg?: string) {
+    super(msg || code);
+    this.code = code;
+  }
+}
+
+/**
+ * EventEmitter-based job queue
+ */
+export class JobQueue extends EventEmitter {
```
mpetrunic: This queue implementation seems weird and overly comlicated. Maybe use async generators like here:
https://github.com/ChainSafe/lodestar/pull/1551/files#diff-009319ca2922f4fbfff417a57221eb69R14

mpetrunic (commented):
Much better than what we had before. Actually found some optimization hints while looking trough it :)

packages/lodestar/src/chain/regen/regen.ts
wemeetagain: right, checkpoint events get triggered for _all_ processed checkpoints. We won't know which are canonical and which aren't until finalization.
This will help us seed our checkpoint cache with any useful checkpoint, whether it ends up being canonical or not.

packages/lodestar/src/api/impl/validator/validator.ts
wemeetagain: Yeah, I can add add it to this PR.

packages/lodestar/src/api/impl/validator/validator.ts
wemeetagain: I added this to sync the clock state with the current time. (so that forkchoice ERR_FUTURE_SLOT couldn't be triggered)
Calling `clock.currentSlot` is a getter, which can potentially update the current slot if the time has changed.
If we have a `waitForSlot` method on clock, or something functionally similar, then we won't have to worry about synchronization with the clock.

packages/lodestar/src/api/impl/validator/validator.ts
wemeetagain: I think we need to progress to at least the currentEpoch no matter what?
The optimization would be to not advance slots if the head block slot is already in currentEpoch.

packages/lodestar/src/chain/clock/LocalClock.ts
```diff
@@ -47,12 +46,29 @@ export class LocalClock implements IBeaconClock {
     return computeEpochAtSlot(this.config, this.currentSlot);
   }
 
-  public abort = (): void => {
-    clearTimeout(this.timeoutId);
-    if (this.signal) {
-      this.signal.removeEventListener("abort", this.abort);
+  public async waitForSlot(slot: Slot): Promise<void> {
```
mpetrunic, line 49: might make sense to add signal as param as well but whoever needs it can do it :)

mpetrunic (dismissed):

wemeetagain:
e2e still failing :( seems attestations can't be produced in some cases 
```
2020-09-24 23:34:54  [VALIDATOR 0-8]   error: Failed to produce attestation slot=95, committee=0, error=Failed to obtain attestation at slot 95 and committee 0: Validator not in given committee
```

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -98,18 +97,13 @@ export class ValidatorApi implements IValidatorApi {
   public async produceAttestation(validatorPubKey: BLSPubkey, index: CommitteeIndex, slot: Slot): Promise<Attestation> {
     try {
       await this.checkSyncStatus();
-      const [headBlockRoot, {state: headState, epochCtx}] = await Promise.all([
-        this.chain.forkChoice.getHeadRoot(),
-        this.chain.getHeadStateContext(),
-      ]);
+      const headRoot = this.chain.forkChoice.getHeadRoot();
+      const {state, epochCtx} = await this.chain.regen.getBlockSlotState(headRoot, slot);
```
mpetrunic, line 101: I'm pretty sure we only need state at first slot of attestation slot epoch. As we will have same committee if we are on first slot of epoch or in attestation slot but in case of skip slots we would have to roll state unnecessary

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -118,6 +112,7 @@ export class ValidatorApi implements IValidatorApi {
 
   public async publishBlock(signedBlock: SignedBeaconBlock): Promise<void> {
     await this.checkSyncStatus();
+    await this.chain.clock.waitForSlot(signedBlock.message.slot);
```
mpetrunic: do we even need this now that we have pending block pool?

packages/lodestar/src/chain/factory/block/index.ts
```diff
@@ -23,26 +23,26 @@ export async function assembleBlock(
   randaoReveal: Bytes96,
   graffiti = ZERO_HASH
 ): Promise<BeaconBlock> {
-  const [parentBlock, stateContext] = await Promise.all([chain.getHeadBlock(), chain.getHeadStateContext()]);
-  const parentHeader: BeaconBlockHeader = blockToHeader(config, parentBlock!.message);
-  const headState = stateContext!.state.clone();
+  const head = chain.forkChoice.getHead();
+  const stateContext = await chain.regen.getBlockSlotState(head.blockRoot, slot - 1);
```
mpetrunic: I don't think -1 should be here, it says in spec that state should be rolled to assigned slot not assigned slot -1:

> Note: In this section, state is the state of the slot for the block proposal without the block yet applied. That is, state is the previous_state processed through any empty slots up to the assigned slot using process_slots(previous_state, slot).

mpetrunic (commented):
Just few more comments and I think we are there! :smile:

packages/lodestar/src/chain/factory/block/index.ts
wemeetagain: lmao, wow, we were never doing this right...

packages/lodestar/src/api/impl/validator/validator.ts
wemeetagain: Spec says:
> Let head_state be the state of head_block processed through any empty slots up to the assigned slot using process_slots(state, slot).

mpetrunic (dismissed):

mpetrunic:
oh man, linting error :(

mpetrunic (dismissed):

mpetrunic:
failing unit test :rofl:

twoeths (approved):

PR #1551, opened 2020-09-18
State transition worker thread

### LINKED ISSUE #1464: Block state transition in worker thread
Currently the state transition is a synchronous function, this leads to problems when many epochs are processed at once, since we cannot perform any other tasks (eg: respond to network requests).

Create a worker thread that will process the state transition if block.message.slot - state.slot > MAX_SYNC_SLOTS.

mpetrunic:
This is some POC and I would like your opinion on direction I've taken.
I've used singleton because:
- block processing and attestation validation are somewhat pure functions so passing another object would be major change in a lot of places
- we have just one thread for that anyways so it isn't really important to create multiple instances of that

Done:
- add threads.js library which provides abstraction so everything works in browser
- add fifo queue with async source and generator so if we have duplicated jobs we can return pregenerated state
- serialize state transition and mark as Transferable (avoids buffer copy)
- add EpochContext regeneration since we cannot serialize it yet
   - I wonder how much time this consumes since it's only done twice but we can definitely improve that
- I'm sending IBeaconParams only and regenerating IBeaconConfig in thread
   - I wish I could do this only once while spawning thread but I don't see any option, I could hack it by exposing additional method in worker thread which I would call to initialize config

TODO:
- add e2e to ensure everything works
- return state from cache if duplicate jobs
- integrate in block process and gossip validation

resolves #1464

### DIFF
--- a/packages/lodestar/package.json
+++ b/packages/lodestar/package.json
@@ -85,6 +85,7 @@
     "snappyjs": "^0.6.0",
     "stream-to-it": "^0.2.0",
     "strict-event-emitter-types": "^2.0.0",
+    "threads": "^1.6.3",
     "varint": "^5.0.0"
   },
   "devDependencies": {
--- a/packages/lodestar/src/chain/blocks/stateTransition/index.ts
+++ b/packages/lodestar/src/chain/blocks/stateTransition/index.ts
@@ -1,0 +1,1 @@
+export * from "./stateTransition";
--- a/packages/lodestar/src/chain/blocks/stateTransition/stateTransition.ts
+++ b/packages/lodestar/src/chain/blocks/stateTransition/stateTransition.ts
@@ -10,11 +10,11 @@ import {
 import {processSlots} from "@chainsafe/lodestar-beacon-state-transition/lib/fast/slot";
 import {IForkChoice} from "@chainsafe/lodestar-fork-choice";
 
-import {ITreeStateContext} from "../../db/api/beacon/stateContextCache";
-import {ChainEventEmitter} from "../emitter";
-import {IBlockProcessJob} from "../interface";
+import {ITreeStateContext} from "../../../db/api/beacon/stateContextCache";
+import {ChainEventEmitter} from "../../emitter";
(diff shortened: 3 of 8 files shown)


mpetrunic:
unstale

dapplion:
@mpetrunic For this to work do we have to serialize the entire state to the worker?

mpetrunic:
> @mpetrunic For this to work do we have to serialize the entire state to the worker?

yeah, but state and params are easy, main issue is epochCtx serialization, otherwise we would have to run slow state transition,

dapplion:
> > @mpetrunic For this to work do we have to serialize the entire state to the worker?
> 
> yeah, but state and params are easy, main issue is epochCtx serialization, otherwise we would have to run slow state transition,

Right, that's a serious issue. I question then if running the fast state transition in a worker thread is possible

mpetrunic:
> > > @mpetrunic For this to work do we have to serialize the entire state to the worker?
> > 
> > 
> > yeah, but state and params are easy, main issue is epochCtx serialization, otherwise we would have to run slow state transition,
> 
> Right, that's a serious issue. I question then if running the fast state transition in a worker thread is possible

My idea was, before I stopped, to recalculate epochCtx in worker thread. Since we are only doing that for quite long epoch transition it might be worth doing double epochCtx recalculation (one in thread and one when we get new state back from thread)

dapplion:
> My idea was, before I stopped, to recalculate epochCtx in worker thread. Since we are only doing that for quite long epoch transition it might be worth doing double epochCtx recalculation (one in thread and one when we get new state back from thread)

We should first optimize the state transition in a single thread as much as possible and then look into multi-threading

dapplion:
Thanks to @tuyennhv work we've been able to optimize the state transition significantly without recurring to threads. It seems that it may be enough for our needs, so I would be in favor of not pursuing this PR. The concepts should be applied to multi-threading BLS verification which may be much easier

wemeetagain:
closing for now, we can revisit later if multithreaded bls and state transition optimizations are insufficient.

PR #1559, opened 2020-09-21
Avoid signing conflicting attestation

### LINKED ISSUE #1549: E2e test error - avoided signing conflicting attestation
**Describe the bug**

Validator client stores attestations under "attester index" key, which our internal index in array
of validators in validator client. When we have 2 validator clients (example multi threaded tests),
two different validators will have same "attester index" and thus second validator thinks he is about to produce 
slashable attestation.

**Expected behavior**

We should either store validator published attestation under pubkey or obtain real validator index and use that.

**Steps to Reproduce**

Run e2e tests.

**Screenshots**

image

 - OS: Ubuntu
 - Version: 20
 - Branch: master

mpetrunic:
resolves #1549

### DIFF
--- a/packages/lodestar-validator/src/services/attestation.ts
+++ b/packages/lodestar-validator/src/services/attestation.ts
@@ -354,19 +354,33 @@ export class AttestationService {
 
   private async isConflictingAttestation(attesterIndex: number, other: AttestationData): Promise<boolean> {
     const potentialAttestationConflicts = await this.db.getAttestations(this.publicKeys[attesterIndex], {
-      gt: other.target.epoch - 1,
+      gte: other.target.epoch,
     });
     return potentialAttestationConflicts.some((attestation) => {
-      return isSlashableAttestationData(this.config, attestation.data, other);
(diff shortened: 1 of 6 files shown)


packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -354,19 +354,33 @@ export class AttestationService {
...
+          targetEpoch: other.target.epoch,
+          conflictx: JSON.stringify(potentialAttestationConflicts),
+        });
+      }
+      return result;
     });
   }
 
   private async storeAttestation(attesterIndex: number, attestation: Attestation): Promise<void> {
+    // this.logger.error("storing attestation", {
```
wemeetagain: maybe logger.verbose or logger.debug?

wemeetagain (dismissed):
:+1:

packages/lodestar-validator/src/services/attestation.ts
mpetrunic: Ill remove, i started planting logs to debug behavior 😅

wemeetagain (approved):

PR #1560, opened 2020-09-21
Remove signal handler after finish

wemeetagain:
heap_sept_21

Before and after


3xtr4t3rr3str14l (approved):

dapplion:
Wow, pretty critical! @wemeetagain will this implementation leak or be safe? https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/util/sleep.ts

wemeetagain:
@dapplion seems like it should be safe.
We just need to be sure there's a `signal.removeEventListener` that gets called when an abortable thing is completed and not aborted.

PR #1561, opened 2020-09-21
Gossip Validation Spec Compliance

### LINKED ISSUE #1544: Gossip Block Validation: the current finalized_checkpoint is an ancestor of block
From the network specs (https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/p2p-interface.md#beacon_block)
>   -[REJECT] The current finalized_checkpoint is an ancestor of block -- i.e. get_ancestor(store, block.parent_root, compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)) == store.finalized_checkpoint.root

It doesn't look like we are checking for this in the block validation explicitly (i.e. REJECTing a given gossip block during validation if the above conditions are not met), and we should be doing this to be up-to-spec.

### LINKED ISSUE #1572: Add self-consistency checks to attestation gossip validation
https://github.com/ethereum/eth2.0-specs/pull/2001

### LINKED ISSUE #1573: Add two more simple gossip validations
https://github.com/ethereum/eth2.0-specs/pull/2030

3xtr4t3rr3str14l:
I've been going through the network spec (https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/p2p-interface.md#global-topics) trying to verify that we are up-to-spec, and have found some holes in this that are fixed here.

```[IGNORE] The valid aggregate attestation defined by hash_tree_root(aggregate) has not already been seen (via aggregate gossip, within a verified block, or through the creation of an equivalent aggregate locally).```

- fixing this by calling addAggregateAndProof inside produceAggregateAndProof (so that it can be verified in the seenAttestationCache check during AggregateAndProof validation)

```[REJECT] The aggregator's validator index is within the committee -- i.e. aggregate_and_proof.aggregator_index in get_beacon_committee(state, aggregate.data.slot, aggregate.data.index).```

- we currently throw an error in getBeaconCommittee if `index >= slotCommittees.length`, which would bubble up to the gossip validation try/catch which returns an IGNORE, but we need to throw a REJECT.  I added try/catch to make sure we throw a REJECT when appropriate within the aggregationAndProof gossip validation.

```[IGNORE] attestation.data.slot is within the last ATTESTATION_PROPAGATION_SLOT_RANGE slots (within a MAXIMUM_GOSSIP_CLOCK_DISPARITY allowance) -- i.e. attestation.data.slot + ATTESTATION_PROPAGATION_SLOT_RANGE >= current_slot >= attestation.data.slot (a client MAY queue future attestations for processing at the appropriate slot).```

- we aren't currently using MAXIMUM_GOSSIP_CLOCK_DISPARITY, so I fixed this and consolidated the check into a common function that can be used for validating slots for both attestation and aggregateAndProof gossip validation.

```[REJECT] The current finalized_checkpoint is an ancestor of the block defined by attestation.data.beacon_block_root -- i.e. get_ancestor(store, attestation.data.beacon_block_root, compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)) == store.finalized_checkpoint.root```
- we weren't checking for this, so i added a check for this

fixes #1544 
fixes #1573 
fixes #1572 

More to come, possibly.
I'm also going to check our unit/e2e tests and make sure we are running tests that check explicitly for the rules outlined for gossip validation in the network spec.

I would like to change some of the tests to where we aren't as many stubs and instead use some real-ish data (except maybe in places where we're trying to force values for conditions that we've already tested for with real-ish data), but if I go that route, I will do it in another PR.

### CODE REVIEW

File: packages/lodestar/src/api/impl/validator/validator.ts [hunk 1]
```diff
@@ -235,11 +235,14 @@ export class ValidatorApi implements IValidatorApi {
...
     });
 
-    return {
+    const aggregateAndProof = {
       aggregate,
       aggregatorIndex,
       selectionProof: EMPTY_SIGNATURE,
     };
+
+    await this.db.seenAttestationCache.addAggregateAndProof(aggregateAndProof);
```
mpetrunic: I think this should go in publishAggregateAndProof

packages/lodestar/src/network/gossip/validation/aggregateAndProof.ts
```diff
@@ -102,59 +92,67 @@ export async function validateAggregateAttestation(
...
 
-  if (!isValidIndexedAttestation(epochCtx, state, epochCtx.getIndexedAttestation(attestation))) {
-    logger.warn("Rejected gossip aggregate and proof", {reason: "invalid indexed attestation", ...logContext});
+    if (!isValidIndexedAttestation(epochCtx, state, epochCtx.getIndexedAttestation(attestation))) {
+      logger.warn("Rejected gossip aggregate and proof", {reason: "invalid indexed attestation", ...logContext});
+      return ExtendedValidatorResult.reject;
+    }
+  } catch (e) {
+    // if any errors are thrown in the process of checking for REJECTs, return a REJECT
+    logger.warn("Rejected gossip aggregate and proof", {reason: e.message, ...logContext});
```
mpetrunic, line 139: can you just change that in GossipMessageValidator.isValidIncomingAggregateAndProof to be reject as we already catch errors there

3xtr4t3rr3str14l: yeah i was wondering if there was any reason why the ignore was thrown in that particular try/catch.  do you happen to know?  or is it mostly just an arbitrary default value?  if we change that try/catch to throw a reject, then there could be a possibility that while we're computing some values for a case that wants to throw an ignore that we would run into the same problem by throwing a reject by default.

3xtr4t3rr3str14l: i will probably also be restructuring the way the errors are thrown once we get the error type that @wemeetagain  said he was going to be making a PR for, so this might not apply anymore when that happens.

packages/lodestar/src/api/impl/validator/validator.ts
3xtr4t3rr3str14l: done

packages/lodestar/src/network/gossip/validation/aggregateAndProof.ts
mpetrunic: > yeah i was wondering if there was any reason why the ignore was thrown in that particular try/catch. do you happen to know?

No particular reason, I've added those top level try catch in case we have some bug or encounter some unexpected value. My reasoning for ignore instead of reject was that if we have bug it's better not to censor messages. But you are probably right, it's better to reject in those situations.

3xtr4t3rr3str14l: Well, I think it depends on where the bug would be happening.  If the bug happens while trying to check for a condition that should throw an ignore if the condition is not met, then we should throw ignore, and if the bug happens when we are checking for a condition that should throw a reject if the condition is not met, then we should throw a reject.  I think maybe to keep this PR minimal and specific to spec-compliance, I will make a separate PR for restructuring the errors and incorporating the work done in #1579

3xtr4t3rr3str14l: Though I'm not 100% if my assumption is correct that we should throw a reject or ignore depending on whether the bug happens in a reject or ignore condition check.  Open to questioning that assumption, if there's a good case for it.
On one hand, we could have a condition where we intentionally throw an ignore or reject because some data is invalid/missing from the gossip, but on the other hand, we may also have a bug that is something wrong with Lodestar, so maybe in that case we want a generic ignore or reject for catching these bugs.  I think a generic ignore for catching bugs could be good since it doesn't punish the user like reject would, but I'm not sure.

For example, in the case of this error here (https://github.com/ChainSafe/lodestar/blob/7249507d26601f8c14864ace476f6808601e4da9/packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts#L133) that could get thrown while retrieving the beacon committee to see if the aggregator is in the committee (the case brought up in the description of this PR), would this error only happen if there was a bug in the code?  Or could this also happen because of some bad data in the incoming gossip?

3xtr4t3rr3str14l: To me, it seems like a reject should be returned in that specific case, but I'm not sure about all cases.

3xtr4t3rr3str14l:
I would like to change some of the tests to where we aren't as many stubs and instead use some real-ish data (except maybe in places where we're trying to force values for conditions that we've already tested for with real-ish data), but if I go that route, I will do it in another PR.

packages/lodestar/src/network/gossip/validation/attestation.ts
```diff
@@ -120,20 +158,47 @@ export async function validateGossipAttestation(
...
-/**
- * is ready to be included in block
- */
-export function hasValidAttestationSlot(config: IBeaconConfig, genesisTime: number, attestation: Attestation): boolean {
-  const currentSlot = getCurrentSlot(config, genesisTime);
+export function isUnaggregatedAttestation(attestation: Attestation): boolean {
+  return Array.from(attestation.aggregationBits).filter((bit) => !!bit).length === 1;
+}
+
+export function isCommitteeIndexWithinRange(attestationData: AttestationData, epochCtx: EpochContext): boolean {
```
mpetrunic: can you make epochCtx first param to keep consistency

packages/lodestar/src/network/gossip/validation/attestation.ts
```diff
@@ -120,20 +158,47 @@ export async function validateGossipAttestation(
...
+export function isAncestorOfBlock(
+  config: IBeaconConfig,
+  chain: IBeaconChain,
+  attestationData: AttestationData,
+  checkpoint: Checkpoint
+): boolean {
+  if (!checkpoint) return false;
+
+  let ancestor;
+  // TODO: added try/catch to account for undefined block which throws ForkChoiceError but isn't caught up here
```
mpetrunic: you already have try catch?

packages/lodestar/src/network/gossip/validation/attestation.ts
```diff
@@ -120,20 +158,47 @@ export async function validateGossipAttestation(
...
+  try {
+    const startSlot = computeStartSlotAtEpoch(config, checkpoint.epoch);
+    ancestor = chain.forkChoice.getAncestor(attestationData.beaconBlockRoot, startSlot);
+  } catch (e) {
+    return false;
+  }
+  return toHexString(ancestor) === toHexString(checkpoint.root);
```
mpetrunic: it's faster to use `config.types.Root.equals(ancestor,checkpoint.root)`

packages/lodestar/src/network/gossip/validation/attestation.ts
```diff
@@ -120,20 +158,47 @@ export async function validateGossipAttestation(
...
 
-export function isUnaggregatedAttestation(attestation: Attestation): boolean {
-  return Array.from(attestation.aggregationBits).filter((bit) => !!bit).length === 1;
+export function doesEpochSlotMatchTarget(config: IBeaconConfig, attestationData: AttestationData): boolean {
+  return attestationData.target.epoch === computeEpochAtSlot(config, attestationData.slot);
+}
+
+export function isAncestorOfBlock(
+  config: IBeaconConfig,
+  chain: IBeaconChain,
```
mpetrunic: you can just pass forkChoice to make testing easier

packages/lodestar/src/network/gossip/validation/attestation.ts
```diff
@@ -120,20 +158,47 @@ export async function validateGossipAttestation(
...
   );
 }
 
-export function isUnaggregatedAttestation(attestation: Attestation): boolean {
-  return Array.from(attestation.aggregationBits).filter((bit) => !!bit).length === 1;
+export function doesEpochSlotMatchTarget(config: IBeaconConfig, attestationData: AttestationData): boolean {
+  return attestationData.target.epoch === computeEpochAtSlot(config, attestationData.slot);
+}
+
+export function isAncestorOfBlock(
```
mpetrunic: don't we already have similar method in forkchoice?

packages/lodestar/src/network/gossip/validation/block.ts
```diff
@@ -90,6 +94,16 @@ export async function validateGossipBlock(
     });
     return ExtendedValidatorResult.reject;
   }
+  if (
+    toHexString(await chain.forkChoice.getAncestor(blockRoot, finalizedSlot)) !== toHexString(finalizedCheckpoint.root)
```
mpetrunic: config.types.Root.equals()

mpetrunic: also you aren't accounting for when getAncestor returns undefined. YOu could make generic method which works here and in attestation validation

packages/lodestar/src/network/gossip/validation/attestation.ts
3xtr4t3rr3str14l: ah yeah meant to take this out after i fixed some unit tests.  will remove in next commit

packages/lodestar/src/network/gossip/validation/attestation.ts
3xtr4t3rr3str14l: yep, looks like isDescendant and isDescendantOfFinalized will do the trick.  will put this in next commit

packages/lodestar/src/network/gossip/validation/block.ts
3xtr4t3rr3str14l: going to use isDescendantOfFinalized for both

packages/lodestar/src/network/gossip/validation/utils/hasValidAttestationSlot.ts
```diff
@@ -0,0 +1,13 @@
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
+import {getCurrentSlot} from "@chainsafe/lodestar-beacon-state-transition";
+import {Slot} from "@chainsafe/lodestar-types";
+import {ATTESTATION_PROPAGATION_SLOT_RANGE, MAXIMUM_GOSSIP_CLOCK_DISPARITY} from "../../../../constants";
+
+export function hasValidAttestationSlot(config: IBeaconConfig, genesisTime: number, slot: Slot): boolean {
```
wemeetagain: I think this function have the following signature:
`(config: IBeaconConfig, currentSlot: Slot, slot: Slot): boolean`
and be called like `hasValidAttestationSlot(config, chain.clock.currentSlot, aggregate.data.slot)`

wemeetagain (changes requested):
Looking good

one minor comment

packages/lodestar/src/network/gossip/validation/attestation.ts
```diff
@@ -120,20 +157,28 @@ export async function validateGossipAttestation(
...
- * is ready to be included in block
- */
-export function hasValidAttestationSlot(config: IBeaconConfig, genesisTime: number, attestation: Attestation): boolean {
-  const currentSlot = getCurrentSlot(config, genesisTime);
+export function isUnaggregatedAttestation(attestation: Attestation): boolean {
+  return Array.from(attestation.aggregationBits).filter((bit) => !!bit).length === 1;
+}
+
+export function isCommitteeIndexWithinRange(epochCtx: EpochContext, attestationData: AttestationData): boolean {
+  return attestationData.index < epochCtx.getCommitteeCountAtSlot(attestationData.target.epoch);
```
mpetrunic: method expects slot not epoch!!
Please create unit test for this:
```
@chainsafe/lodestar: 2020-10-06 21:34:14 [NETWORK]         error: Error while validating gossip attestation message=crosslink committee retrieval: out of range epoch: 1 
@chainsafe/lodestar: Error: crosslink committee retrieval: out of range epoch: 1
@chainsafe/lodestar:     at EpochContext._getSlotCommittees (/home/runner/work/lodestar/lodestar/packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts:236:13)
@chainsafe/lodestar:     at EpochContext.getCommitteeCountAtSlot (/home/runner/work/lodestar/lodestar/packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts:139:17)
@chainsafe/lodestar:     at isCommitteeIndexWithinRange (/home/runner/work/lodestar/lodestar/packages/lodestar/src/network/gossip/validation/attestation.ts:169:43)
@chainsafe/lodestar:     at Object.validateGossipAttestation (/home/runner/work/lodestar/lodestar/packages/lodestar/src/network/gossip/validation/attestation.ts:123:8)
@chainsafe/lodestar:     at GossipMessageValidator.isValidIncomingCommitteeAttestation (/home/runner/work/lodestar/lodestar/packages/lodestar/src/network/gossip/validator.ts:59:14)
@chainsafe/lodestar:     at LodestarGossipsub.validate (/home/runner/work/lodestar/lodestar/packages/lodestar/src/network/gossip/gossipsub.ts:91:26)
```

mpetrunic (changes requested):

packages/lodestar/src/network/gossip/validation/attestation.ts
```diff
@@ -120,20 +157,28 @@ export async function validateGossipAttestation(
...
-    attestation.data.slot + ATTESTATION_PROPAGATION_SLOT_RANGE >= currentSlot && currentSlot >= attestation.data.slot
+    attestation.aggregationBits.length ===
+    attestationPreStateContext.epochCtx.getBeaconCommittee(attestation.data.slot, attestation.data.index).length
   );
 }
 
-export function isUnaggregatedAttestation(attestation: Attestation): boolean {
-  return Array.from(attestation.aggregationBits).filter((bit) => !!bit).length === 1;
+export function doesEpochSlotMatchTarget(config: IBeaconConfig, attestationData: AttestationData): boolean {
+  return attestationData.target.epoch === computeEpochAtSlot(config, attestationData.slot);
```
dapplion: @wemeetagain Should we use the SSZ equality helper here in case we switch types?

packages/lodestar/test/unit/network/gossip/validation/attestation.test.ts
```diff
@@ -175,23 +181,207 @@ describe("gossip attestation validation", function () {
...
+    forkChoice.hasBlock.returns(true);
+    const attestationPreState = {
+      state: generateState(),
+      epochCtx: new EpochContext(config),
+    };
+    attestationPreState.epochCtx.previousShuffling = {
+      activeIndices: [],
+      epoch: 0,
+      committees: [[[]]],
+    } as any;
```
dapplion: Is this `any` necessary? Try to do a better type casting if possible

packages/lodestar/test/unit/network/gossip/validation/attestation.test.ts
```diff
@@ -175,23 +181,207 @@ describe("gossip attestation validation", function () {
...
+    attestationPreState.epochCtx.getIndexedAttestation = () => {
+      return (attestation as unknown) as IndexedAttestation;
```
dapplion: Is it possible to do a better type casting?

dapplion (commented):
comments

packages/lodestar/src/network/gossip/validation/attestation.ts
3xtr4t3rr3str14l: ah yeah, good catch.  thanks for that.  will put this in next commit.

mpetrunic (approved):

mpetrunic (dismissed):

mpetrunic:
I'm trying to add comment unsuccessfully, does  e2e tests pass locally'?

3xtr4t3rr3str14l:
> I'm trying to add comment unsuccessfully, does e2e tests pass locally'?

@mpetrunic it does happen locally.  still not sure why this is going on.

wemeetagain (dismissed):

3xtr4t3rr3str14l:
> > I'm trying to add comment unsuccessfully, does e2e tests pass locally'?
> 
> @mpetrunic it does happen locally. still not sure why this is going on.

resolved via 26b11fa

wemeetagain (approved):

Issue #1346, opened 2020-08-05
Backward sync controlled by sync


wemeetagain:
**Describe the bug**
Currently we abuse selectively emiting "unknownBlockRoot" in the block pool to suppress backwards sync. This is confusing, since the block pool should not be concerned with syncing strategy. There might also be eventually other consumers of "unknownBlockRoot" which need every unknown block root to fire an event.

**Expected behavior**
Sync controls backwards sync.

**Screenshots**
https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/chain/blocks/pool.ts#L35

PR #1563, opened 2020-09-22
Avoid write EPIPE error at regular sync

### LINKED ISSUE #1498: UnhandledPromiseRejectionWarning: Error: write EPIPE
**Describe the bug**
Error recevied while syncing beacon
```
(node:107746) UnhandledPromiseRejectionWarning: Error: write EPIPE
    at afterWriteDispatched (internal/stream_base_commons.js:150:25)
    at writeGeneric (internal/stream_base_commons.js:141:3)
    at Socket._writeGeneric (net.js:770:11)
    at Socket._write (net.js:782:8)
    at doWrite (_stream_writable.js:431:12)
    at writeOrBuffer (_stream_writable.js:415:5)
    at Socket.Writable.write (_stream_writable.js:305:11)
    at /home/seamonkey/Downloads/lodestar/packages/lodestar/node_modules/libp2p-tcp/node_modules/stream-to-it/sink.js:71:20
    at async Object.sink (/home/seamonkey/Downloads/lodestar/packages/lodestar/node_modules/libp2p-tcp/src/socket-to-conn.js:31:9)
```

**Expected behavior**
no error

**Steps to Reproduce**
Sync beacon

 - OS: Linux Mint
 - Version: 20
 - Branch: Master
 - Commit hash: 78b3764349b12651ddfc0fb15ff537aea9231217

twoeths:
resolves #1498

### DIFF
--- a/packages/lodestar/src/sync/sync.ts
+++ b/packages/lodestar/src/sync/sync.ts
@@ -197,6 +197,11 @@ export class BeaconSync implements IBeaconSync {
   }
 
   private onUnknownBlockRoot = async (root: Root): Promise<void> => {
+    // at regular sync time, we fetch follow-up range so don't need to handle this
+    // otherwise, got "write EPIPE" error bc we call beacon_block_by_root too much
+    // at the same time to same peer and it disconnects us
+    // this happens with "missing parent" gossip blocks
+    if (!this.isSynced()) return;
     const peerBalancer = new RoundRobinArray(this.getPeers());
     // eslint-disable-next-line no-constant-condition
     while (true) {
--- a/packages/lodestar/src/sync/sync.ts
+++ b/packages/lodestar/src/sync/sync.ts
@@ -60,7 +60,6 @@ export class BeaconSync implements IBeaconSync {
     this.mode = SyncMode.WAITING_PEERS as SyncMode;
     await this.reqResp.start();
     await this.attestationCollector.start();
-    this.chain.emitter.on("unknownBlockRoot", this.onUnknownBlockRoot);
     // so we don't wait indefinitely
     await this.waitForPeers();
     if (this.mode === SyncMode.STOPPED) {
@@ -155,6 +154,7 @@ export class BeaconSync implements IBeaconSync {
 
   private syncCompleted = async (): Promise<void> => {
     this.stopSyncTimer();
+    this.chain.emitter.on("unknownBlockRoot", this.onUnknownBlockRoot);
     await this.network.handleSyncCompleted();
   };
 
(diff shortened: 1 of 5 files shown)


packages/lodestar/src/sync/sync.ts
```diff
@@ -197,6 +197,11 @@ export class BeaconSync implements IBeaconSync {
   }
 
   private onUnknownBlockRoot = async (root: Root): Promise<void> => {
+    // at regular sync time, we fetch follow-up range so don't need to handle this
+    // otherwise, got "write EPIPE" error bc we call beacon_block_by_root too much
+    // at the same time to same peer and it disconnects us
+    // this happens with "missing parent" gossip blocks
+    if (!this.isSynced()) return;
```
mpetrunic, line 204: but if we are on `current slot - 2` and receive gossip block at `current slot - 2` we would need that to fetch missing parent? Or are you hopping we would fetch range that would include missing block?

Side question after regular sync is completed, what service handles unknown block root events?

twoeths: > but if we are on current slot - 2 and receive gossip block at current slot - 2 we would need that to fetch missing parent? Or are you hopping we would fetch range that would include missing block?

we should have gossip parent block at that time bc we fetch ranges from behind until missing parent gossip blocks, given the fact that we use single peer for regular sync so it should return the same chain to head

> Side question after regular sync is completed, what service handles unknown block root events?

this `sync.ts` still handles "unknownBlockRoot" event after the node is synced. I think it not only manages initial sync/regular sync but also after that time. Wdyt?
One possible fix for this issue could be only register `unknownBlockRoot` event after regular sync completed.

mpetrunic: > we should have gossip parent block at that time bc we fetch ranges from behind until missing parent gossip blocks, given the fact that we use single peer for regular sync so it should return the same chain to head

I'm worried because we are syncing from one peer which might not have latest head at the time we need it. I agree there is low chance for that but it could still stall us I guess.

> One possible fix for this issue could be only register unknownBlockRoot event after regular sync completed.

But we kinda need handling that before and during sync as peers could return range with some block missing (it happens quite often actually).

twoeths: > I'm worried because we are syncing from one peer which might not have latest head at the time we need it. I agree there is low chance for that but it could still stall us I guess.

I found that "missing parent" gossip blocks come regularly so as long as the best peer does not way below head, we're good since "follow up" range and "missing parent gossip blocks" range will gradually overlap and we get through regular sync

> But we kinda need handling that before and during sync as peers could return range with some block missing (it happens quite often actually).

Agree. So this main `sync.ts` service takes care "after sync", and the regular sync should handle "unknowBlockRoot" event itself (will do it in an upcoming PR for regular sync).

twoeths:
this is related to #1346 too

packages/lodestar/src/sync/sync.ts
```diff
@@ -155,6 +154,7 @@ export class BeaconSync implements IBeaconSync {
 
   private syncCompleted = async (): Promise<void> => {
     this.stopSyncTimer();
+    this.chain.emitter.on("unknownBlockRoot", this.onUnknownBlockRoot);
```
twoeths, line 157: so this becomes an event for afterSync. Do we want to rename it to "afterSync:unknownBlockRoot" or "synced:unknownBlockRoot"?

twoeths:
replaced by #1594

Issue #1249, opened 2020-07-28
Structured logs


wemeetagain:
I know it was designed that way, I do think we should minimize nested structures when we can. 
I'm just trying to make our logs more machine readable, and handling arbitrary `Json` IMO is part of the process, especially for debug or verbose-level logs.

IMO, in an ideal setting, we would output logs to JSON-LD format in production, or with a certain "productionized logs" flag.
We would still have a pretty-printer, like in this file, for daily use, but the underlying representation is simply `Json`. This will allow us to easily ingest logs into log aggregators and do more advanced analytics on logs. This is how logging will be used in production settings.
Logs that look like `[CHAIN] "Attestation ${X} processed at epoch ${Y}"` should be refactored into `[CHAIN] "processed attestation", {attestationRoot, epoch}`. Which has a machine representation of `{timestamp: 9999999, module: "chain", message: "processed attestation", attestationRoot, epoch}`

_Originally posted by @wemeetagain in https://github.com/ChainSafe/lodestar/pull/1240#discussion_r461631285_

dapplion:
Could all mantainers of Lodestar agree or disagree to 
- consistently follow this proposal
- not follow this proposal

Otherwise, we would continue our current state of different log styles all over the place

PR #1566, opened 2020-09-22
Add pending slot block pool

### LINKED ISSUE #1550: E2e test error - invalid block
**Describe the bug**

Occasionally in e2e tests we are producing bad blocks. Unfortunately, as seen in screenshot, error is undefined.
To investigate further we would need exact error. Since that worked before, my best guess it's related to either recent block processing changes or eth1 data. 

**Expected behavior**

Should not happen so e2e tests can pass consistently.

**Steps to Reproduce**

**Screenshots**

image

 - OS: Ubuntu
 - Version: 20
 - Branch: master
 - Commit hash:

### LINKED ISSUE #1565: Add pending slot blocks support to block pool

Currently if block comes too early we mark it as bad block which is somewhat bad since we cannot ever process it again and thus cannot form valid chain. Currently we experience similar stuff in e2e tests (#1550 ) but since we have majority we form chain without that block.

Similar to attestation pool, block pool should index blocks by their slot and listen on slot ticker to send block back to processing queue.

Let it be!

**Additional context**

mpetrunic:
tldr: slots are short and blocks come in in millseconds on edge of slot, I've increased clock precision but eventually
I had to add pending block slot pool to mitigate.

We don't have bad blocks in e2e tests anymore :)

resolves #1565 
resolves #1550


packages/lodestar/src/chain/blocks/process.ts
```diff
@@ -29,6 +29,10 @@ export function processBlock(
   return (source) => {
     return (async function* () {
       for await (const job of source) {
+        if (job.signedBlock.message.slot > forkChoice.getTime()) {
```
wemeetagain: I think we should just pass in our `IBeaconClock` and use `clock.currentSlot` here.
I have a branch where I did the same.

packages/lodestar/src/chain/blocks/pool.ts
```diff
@@ -27,25 +28,42 @@ export class BlockPool {
     this.forkChoice = forkChoice;
   }
 
+  public async start(): Promise<void> {
+    this.eventBus.on("clock:slot", this.onNewSlot);
```
wemeetagain: Can you add this handler in src/chain/chain.ts instead? This will require the block pool to be exposed to the chain.

I'd like to capture all event subscriptions there and prevent these submodules from growing start/stop methods. In a future PR, I'd even like to remove this `eventBus` and `forkChoice` dependencies, as the "unknown block root" scenario can/will be handled by a certain `BlockError` outcome.

packages/lodestar-fork-choice/src/forkChoice/errors.ts
```diff
@@ -174,7 +174,7 @@ export class ForkChoiceError extends Error {
   type: ForkChoiceErrorType;
 
   constructor(type: ForkChoiceErrorType) {
-    super();
+    super(type.code + " " + JSON.stringify(type));
```
wemeetagain, line 177: I tend to think this JSONification should either happen downstream, eg: in logging, or pre-logging, or we should have a custom `toString` method here. Perhaps this can be tackled in a separate PR.

wemeetagain (changes requested):

packages/lodestar-fork-choice/src/forkChoice/errors.ts
mpetrunic: I feel like this is specific to error and to respect error message property, it should be handled here like mentioned in https://github.com/ChainSafe/lodestar/pull/1579#discussion_r493242136

wemeetagain: The problem with mashing the stringified contents here is that you end up with a bunch of duplicate information (and the duplicated info is nasty, stringified and inflexible to different formatting.) This goes against the goal of https://github.com/ChainSafe/lodestar/issues/1249:
```
{
  message: "{\"code\":\"ERR_FUTURE_SLOT\",\"blockSlot\":5,\"currentSlot\":4}",
  stack: "...",
  type: {
    code: "ERR_FUTURE_SLOT",
    blockSlot: 5,
    currentSlot: 4,
  }
}
```

My current thought is that we should provide a custom method on our errors, eg: `toObject`, or `toJson`, and our logger should attempt to convert errors to objects, and do the stringification.  Would have something like the following logic:
```typescript
function defaultErrorObj(e: Error): object {
  const obj = {};
  e.message && obj.message = e.message;
  e.code && obj.code = e.code;
  e.stack && obj.stack = e.stack;
  return obj;
}

function convertError(e: Error): object {
  if (e.toObject) {
    return e.toObject(); // custom errors can supply metadata here
  } else {
    return defaultErrorObj(e); // regular errors handled here
  }
}

// alternatively, substitute `object` for `Json`
```
This allows each specific error to provide its own relevant object-ification, but lets the actual string-ification be deferred to the level of logging, which knows better how strings should be formatted.

mpetrunic: I honestly don't think we should have metadata on Error and we should respect existing Error interface of message/stacktrace.
We can just as easily user logger serialization to serialize metadata into error message field which should work nicely with our logger. I don't think we are using error metadata anywhere in runtime?!

wemeetagain: > I honestly don't think we should have metadata on Error and we should respect existing Error interface of message/stacktrace.

The two aren't mutually exclusive. We can have metadata on an `Error` subclass, and it will respect the superclass interface (message/stacktrace). If we really want to respect the interface, semantically, we should just always pass in a string description of the error (the `message`).

Error metadata gives error handlers _actual data_ to handle error scenarios with more sophistication.

Eg: Do you want an error handler to handle `ERR_FUTURE_SLOT`, but only when the `blockSlot` is within 5 of the `currentSlot`?

If you rely on `err.message` only, thats just too bad, your errors are simply not expressive enough. Or perhaps the handler must now perform some error prone string parsing or regex. If you have error metadata, this becomes possible, and trivial.

> We can just as easily user logger serialization to serialize metadata into error message field which should work nicely with our logger.

Its actually not just as easy. The reason is that string representation requires _serialization_ of structured data. If you preserialize Error metadata, you're literally recreating the situation we had before the logger context object was available!

The structured logger initiative is about _retaining_ structured data, and having the logger perform the serialization that is necessary. By jamming structured data into the error.message you remove the ability of the logger to do its own serialization. 

> I don't think we are using error metadata anywhere in runtime?!

We likely will very soon :) we've only just integrated these (and `BlockError`).

dapplion: Error subclass (with message, stack) + metadata provides the best of both worlds. Having some somewhat expressive message is useful too in case the error leaks to unexpected places, so it's more debuggable.

dapplion: @wemeetagain @mpetrunic would it be best to open a discussion issue and agree on the Lodestar Error strategy?

mpetrunic: > @wemeetagain @mpetrunic would it be best to open a discussion issue and agree on the Lodestar Error strategy?

yup

packages/lodestar-beacon-state-transition/src/util/slot.ts
```diff
@@ -1,12 +1,11 @@
...
-import {computeStartSlotAtEpoch, computeEpochAtSlot} from ".";
+import {Epoch, Number64, Slot} from "@chainsafe/lodestar-types";
+import {computeEpochAtSlot, computeStartSlotAtEpoch} from ".";
 import {GENESIS_SLOT} from "../constants";
 
 export function getSlotsSinceGenesis(config: IBeaconConfig, genesisTime: Number64): Slot {
-  const diffInSeconds = Date.now() / 1000 - genesisTime;
-  return intDiv(diffInSeconds, config.params.SECONDS_PER_SLOT);
+  const diffInMiliSeconds = Date.now() - genesisTime * 1000;
+  return Math.round(diffInMiliSeconds / (config.params.SECONDS_PER_SLOT * 1000));
```
dapplion, line 8: Math.round is not equivalent to intDiv, but Math.floor is

mpetrunic: I'm a bit worried about this change, but we have issue wher 1ms diff is causing us to skip slot counting because of 
https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar-validator/src/api/abstract.ts#L114
where 1ms causes us to skip slot counting (condition is true and by the time we start calculating diff till next slot, it makes us wait 2 second)

wemeetagain: I think we should just make the validator logic smarter. Its probably more important that this function always return the correct result.

wemeetagain: maybe we can add a function here that returns two values, current slot and milliseconds til next slot (both using the same call to `Date.now()`)

mpetrunic: I think I can maybe revert back and just remove that clunky logic in validator as it was needed because we didn't have pending slot block pool

wemeetagain (approved):

PR #1579, opened 2020-09-22
Add block error

wemeetagain:
Code has been sitting around for a while. This just adds `BlockError` and `AttestationError`, does nothing with them.

### DIFF
--- a/packages/lodestar/src/chain/errors/attestationError.ts
+++ b/packages/lodestar/src/chain/errors/attestationError.ts
@@ -1,0 +1,199 @@
+import {CommitteeIndex, Epoch, Slot, ValidatorIndex} from "@chainsafe/lodestar-types";
+
+export enum AttestationErrorCode {
+  /**
+   * The attestation is from a slot that is later than the current slot (with respect to the gossip clock disparity).
+   */
+  ERR_FUTURE_SLOT = "ERR_FUTURE_SLOT",
+  /**
+   * The attestation is from a slot that is prior to the earliest permissible slot
+   * (with respect to the gossip clock disparity).
+   */
+  ERR_PAST_SLOT = "ERR_PAST_SLOT",
+  /**
+   * The attestations aggregation bits were empty when they shouldn't be.
+   */
+  ERR_EMPTY_AGGREGATION_BITFIELD = "ERR_EMPTY_AGGREGATION_BITFIELD",
+  /**
+   * The `selection_proof` on the aggregate attestation does not elect it as an aggregator.
+   */
+  ERR_INVALID_SELECTION_PROOF = "ERR_INVALID_SELECTION_PROOF",
+  /**
+   * The `selection_proof` on the aggregate attestation selects it as a validator,
+   * however the aggregator index is not in the committee for that attestation.
+   */
+  ERR_AGGREGATOR_NOT_IN_COMMITTEE = "ERR_AGGREGATOR_NOT_IN_COMMITTEE",
+  /**
+   * The aggregator index refers to a validator index that we have not seen.
+   */
+  ERR_AGGREGATOR_PUBKEY_UNKNOWN = "ERR_AGGREGATOR_PUBKEY_UNKNOWN",
+  /**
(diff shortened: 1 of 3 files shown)


packages/lodestar/src/chain/errors/blockError.ts
```diff
@@ -0,0 +1,160 @@
...
+export class BlockError extends Error {
+  public type: BlockErrorType;
+  public job: IBlockProcessJob;
+
+  constructor(type: BlockErrorType & JobObject) {
+    super(type.code);
+    this.job = type.job;
+    this.type = type;
+    // delete this.type.job for easier serialization of this.type
+    delete (this.type as BlockErrorType & Partial<JobObject>).job;
```
dapplion: A more typesafe implementation should be easy, by setting
```
public type: Omit<BlockErrorType, "job">
```
And then
```ts
constructor({job, ...type}: BlockErrorType & JobObject) {
  ...
}
```

packages/lodestar/src/chain/errors/attestationError.ts
```diff
@@ -0,0 +1,199 @@
...
+  | {
+      code: AttestationErrorCode.ERR_BEACON_CHAIN_ERROR;
+      error: Error;
+    };
+
+export class AttestationError extends Error {
+  public type: AttestationErrorType;
+
+  constructor(type: AttestationErrorType) {
+    super(type.code);
```
mpetrunic, line 196: We can do in another PR, but I feel like we should handle type serialization to message in constructor of error. Something like:
```typescript
super(this.getMessage(type));


private getMessage(type: AttestationErrorType) {
   let message = type.code;
   switch(type.code) {
      case something:
            message += " details: " + JSON.stringify(type.block)
   }
   return message;
}
```

mpetrunic (dismissed):

packages/lodestar/src/chain/errors/attestationError.ts
wemeetagain: Yeah, we definitely need a way to get this metadata into a printable state for logging purposes.

packages/lodestar/src/chain/errors/attestationError.ts
```diff
@@ -0,0 +1,199 @@
...
+export enum AttestationErrorCode {
+  /**
+   * The attestation is from a slot that is later than the current slot (with respect to the gossip clock disparity).
+   */
+  ERR_FUTURE_SLOT = "ERR_FUTURE_SLOT",
+  /**
+   * The attestation is from a slot that is prior to the earliest permissible slot
+   * (with respect to the gossip clock disparity).
+   */
+  ERR_PAST_SLOT = "ERR_PAST_SLOT",
```
3xtr4t3rr3str14l, line 12: do we actually need to throw different errors for past and future slots?  the network spec rule (for attestation gossip validation, at least) is outlined to where a single IGNORE is thrown when the slot is not within the allowed range.  maybe we combine ERR_FUTURE_SLOT and ERR_PAST_SLOT into one error and call it ERR_SLOT_OUT_OF_RANGE or something like that?

Here's the rule from the spec, btw:
```    [IGNORE] attestation.data.slot is within the last ATTESTATION_PROPAGATION_SLOT_RANGE slots (within a MAXIMUM_GOSSIP_CLOCK_DISPARITY allowance) -- i.e. attestation.data.slot + ATTESTATION_PROPAGATION_SLOT_RANGE >= current_slot >= attestation.data.slot (a client MAY queue future attestations for processing at the appropriate slot).```

wemeetagain: These errors aren't necessarily supposed to be 1 for 1 with gossip validation rules.
They're meant to enumerate meaningfully distinct errors that can happen in the course of processing a block.
Likely, when we're incorporating them into Lodestar further, some cases will be deleted, some will be added.
This list is taken mainly from Lighthouse's block errors.

IMO we should save this discussion for a future PR, when we incorporate these errors into validation functions. At that point it will be more clear whether past vs future slot a meaningful distinction to make.

3xtr4t3rr3str14l (approved):

PR #1593, opened 2020-09-25
spadina testnet lodestar-cli setup

### LINKED ISSUE #1584: Prepare lodestar-cli for spadina testnet
- [ ] Add a new `--testnet` argument, `spadina`
- [ ] Create a new default config for spadina
- [ ] Create everything else that is required to have `init` setup an environment for spadina and for `beacon` to run successfully on spadina (if anyone wants to drill this down further in this issue, feel free, but we can probably use whatever is setup for `--testnet medalla` as an example to create `--testnet spadina`)

3xtr4t3rr3str14l:
resolves #1584

### DIFF
--- a/packages/lodestar-cli/src/testnets/index.ts
+++ b/packages/lodestar-cli/src/testnets/index.ts
@@ -6,8 +6,9 @@ import got from "got";
 import {IBeaconNodeOptionsPartial} from "../options";
 import {altonaConfig} from "./altona";
 import {medallaConfig} from "./medalla";
+import {spadinaConfig} from "./spadina";
 
-export type TestnetName = "altona" | "medalla";
+export type TestnetName = "altona" | "medalla" | "spadina";
 
 export function getTestnetConfig(testnet: TestnetName): IBeaconNodeOptionsPartial {
   switch (testnet) {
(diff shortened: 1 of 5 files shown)


3xtr4t3rr3str14l:
putting this here to announce my desire to skip the codeclimate check this time.  what it's suggesting doesn't seem necessary because it would be good to be able to look at each testnet's config in whole (without having to reference some external common objects, like they suggest with the logger params).  open to objections or other opinions, but just wanna make that known.

dapplion:
> putting this here to announce my desire to skip the codeclimate check this time. what it's suggesting doesn't seem necessary because it would be good to be able to look at each testnet's config in whole (without having to reference some external common objects, like they suggest with the logger params). open to objections or other opinions, but just wanna make that known.

Yeah I'm okay too with the testnet config containing duplicate code.

packages/lodestar-cli/src/testnets/index.ts
```diff
@@ -41,6 +46,8 @@ export function getGenesisFileUrl(testnet: TestnetName): string | null {
       return "https://github.com/eth2-clients/eth2-testnets/raw/b84d27cc8f161cc6289c91acce6dae9c35096845/shared/altona/genesis.ssz";
     case "medalla":
       return "https://github.com/eth2-clients/eth2-testnets/blob/master/shared/medalla/genesis.ssz?raw=true";
+    case "spadina":
+      return ""; // TODO: add genesis.ssz file here
```
dapplion: Better to return null than an empty string

dapplion (dismissed):

wemeetagain (approved):

PR #1596, opened 2020-09-27
Add forkChoice head and reorg events

wemeetagain:
A few updates here:
- update fork choice balances synchronously, as part of `onBlock`. This is a nice change, because it ensures that our fork choice is never out of sync. (Previously, once the fork choice justified a new checkpoint, the `updateBalances` method _must_ be called immediately, or else any `getHead` call will result in an error!!)
- update fork choice initialization in BeaconChain. Should be followed up in a future PR with refactored initializer method(s).
- add `forkChoice:head` and `forkChoice:reorg` events, in a similar style that other events are called in the block processor.


mpetrunic (approved):
Awesome

PR #1600, opened 2020-09-29
Fix dialProtocol after peer disconnected

### LINKED ISSUE #1597: libp2p 0.29: Handle dialProtocol timeout
**Describe the bug**
Sync stalled due to a request timeout does not throw Error.
This happens to me a couple of times with libp2p 0.29: during initial sync/regular sync, peers get disconnected after we called dialProtocol but it did not trigger an error when timeout
```
2020-09-28 19:37:35  [SYNC]             warn: Failed to obtain chunk {"start":395488,"end":395551} from peer 16Uiu2HAmEPBjHgMH4YZeNMFsMC6wrGWKcJR6uFL5j8oahyG2NKzz
2020-09-28 19:37:35  [NETWORK]       verbose: sending beacon_blocks_by_range request to 16Uiu2HAmEPBjHgMH4YZeNMFsMC6wrGWKcJR6uFL5j8oahyG2NKzz requestId=5c42cd4fd33136f3, encoding=ssz_snappy
2020-09-28 19:37:35  [NETWORK]       verbose: peer disconnected 16Uiu2HAmEPBjHgMH4YZeNMFsMC6wrGWKcJR6uFL5j8oahyG2NKzz
```

**Expected behavior**
Network req/resp should throw error for sync to handle

One candidate to fix is #1277

twoeths:
resolves #1597 
+ When timeout we still abort controller to inform libp2p
+ But we throw error ourself to make sure network request finishes

So basically it's the same to the current behavior, we just base on ourself to throw Error instead of libp2p (which is not reliable as reported in the issue)

### DIFF
--- a/packages/lodestar/src/network/util.ts
+++ b/packages/lodestar/src/network/util.ts
@@ -138,17 +138,23 @@ export async function dialProtocol(
   signal?: AbortSignal
 ): ReturnType<LibP2p["dialProtocol"]> {
   const abortController = new AbortController();
-  const timer = setTimeout(() => {
-    abortController.abort();
-  }, timeout);
   const abortHandler = (): void => abortController.abort();
   signal?.addEventListener("abort", abortHandler, {once: true});
+  let timer: NodeJS.Timeout | null = null;
   try {
-    const conn = await libp2p.dialProtocol(peerId, protocol, {signal: abortController.signal} as object);
+    const conn = await Promise.race([
+      libp2p.dialProtocol(peerId, protocol, {signal: abortController.signal} as object),
+      new Promise((_, reject) => {
+        timer = setTimeout(() => {
+          abortController.abort();
+          reject(new Error(RESPONSE_TIMEOUT_ERR));
+        }, timeout);
+      }),
+    ]);
     if (!conn) {
-      throw new Error("timeout");
+      throw new Error(RESPONSE_TIMEOUT_ERR);
     }
-    return conn;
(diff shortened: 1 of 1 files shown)


mpetrunic:
Seems odd, I would ask libp2p guys first rather than doing those weird patches until we figure out reason.

packages/lodestar/src/network/util.ts
```diff
@@ -138,24 +138,32 @@ export async function dialProtocol(
...
   const abortController = new AbortController();
-  const timer = setTimeout(() => {
-    abortController.abort();
-  }, timeout);
   const abortHandler = (): void => abortController.abort();
   signal?.addEventListener("abort", abortHandler, {once: true});
+  let timer: NodeJS.Timeout | null = null;
   try {
-    const conn = await libp2p.dialProtocol(peerId, protocol, {signal: abortController.signal} as object);
+    const conn = await Promise.race([
```
dapplion, line 145: Is it worth to abstract this construction so we can properly test it and make sure it doesn't leak?
```ts
abortableTimeoutCallback(
  fn: () => Promise<void>,
  timeoutMs: number,
  signal: AbortSignal
)
```

wemeetagain: Yeah, I think that would be really useful to pull into its own function. Basically like `setTimeout` but with a signal. (and not supporting the 'evil' extra arguments that `setTimeout` supports)

wemeetagain:
updated libp2p has a fix for this

PR #1603, opened 2020-09-29
Spec v0.12.3 - use raw SHA256 as message-id and reduce size to 8 bytes #2044

### LINKED ISSUE #1574: Use raw SHA256 as message-id and reduce size to 8 bytes
https://github.com/ethereum/eth2.0-specs/pull/2044

dapplion:
Fixes https://github.com/ChainSafe/lodestar/issues/1574

Update to spec v0.12.3

 - Remove the base64 encoding as described in the previous spec.
 - Use only the first 8 bytes as the message id instead of the whole
32 bytes.

### DIFF
--- a/packages/lodestar/src/network/gossip/utils.ts
+++ b/packages/lodestar/src/network/gossip/utils.ts
@@ -75,8 +75,17 @@ export function normalizeInRpcMessage(rawMessage: Message): ILodestarGossipMessa
   };
 }
 
+/**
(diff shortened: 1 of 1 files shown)


mpetrunic (approved):

twoeths (approved):

wemeetagain:
Closing until we have https://github.com/ChainSafe/js-libp2p-gossipsub/pull/143

Issue #1595, opened 2020-09-26
Discussion: Error strategy


dapplion:
Let's discuss and agree on an Error strategy. We should define at least:
- How to define and throw Errors. How to add and handle metadata
- Typing
- Logging

------

The problem with mashing the stringified contents here is that you end up with a bunch of duplicate information (and the duplicated info is nasty, stringified and inflexible to different formatting.) This goes against the goal of https://github.com/ChainSafe/lodestar/issues/1249:
```
{
  message: "{\"code\":\"ERR_FUTURE_SLOT\",\"blockSlot\":5,\"currentSlot\":4}",
  stack: "...",
  type: {
    code: "ERR_FUTURE_SLOT",
    blockSlot: 5,
    currentSlot: 4,
  }
}
```

My current thought is that we should provide a custom method on our errors, eg: `toObject`, or `toJson`, and our logger should attempt to convert errors to objects, and do the stringification.  Would have something like the following logic:
```typescript
function defaultErrorObj(e: Error): object {
  const obj = {};
  e.message && obj.message = e.message;
  e.code && obj.code = e.code;
  e.stack && obj.stack = e.stack;
  return obj;
}

function convertError(e: Error): object {
  if (e.toObject) {
    return e.toObject(); // custom errors can supply metadata here
  } else {
    return defaultErrorObj(e); // regular errors handled here
  }
}

// alternatively, substitute `object` for `Json`
```
This allows each specific error to provide its own relevant object-ification, but lets the actual string-ification be deferred to the level of logging, which knows better how strings should be formatted.

_Originally posted by @wemeetagain in https://github.com/ChainSafe/lodestar/pull/1566#discussion_r494435876_

PR #1609, opened 2020-09-29
Error logger formatting

wemeetagain:
- add `toJson` utility function
  - simple best-effort conversion of `unknown` to `Json`
  - includes converting `Error`s, will first attempt to call `err.toJson()` if it exists
- add `BlockError#toJson` - returns the error in a json-serializable way
- update `serializeContext` to convert the error to json, attempt to print serialized error metadata

Other errors can add a `toJson` method to have pretty-printed error metadata


packages/lodestar-utils/src/logger/format.ts
```diff
@@ -23,7 +24,12 @@ export const defaultLogFormat = format.combine(
 export function serializeContext(context?: Context | Error): string {
   if (!context) return "";
   if (context instanceof Error) {
-    return "Error: " + context.message + "\n" + context.stack;
+    const errObj = toJson(context);
+    if (errObj != null && typeof errObj === "object" && !Array.isArray(errObj)) {
+      delete errObj["message"];
```
mpetrunic: Why aren't we printing error message?

We still have and will have bunch or errors originating from libraries and stuff that won't have meta so we would end-up being clueless about error.

mpetrunic (commented):
I'm a bit confused as this doesn't help us much.

packages/lodestar-utils/src/logger/format.ts
dapplion: Agree this should work with Errors that we don't control

packages/lodestar-utils/src/json.ts
```diff
@@ -0,0 +1,32 @@
...
+  } else if (obj instanceof Error) {
+    return (obj as ICustomError).toJson?.() ?? defaultErrorToJson(obj);
+  } else if (Array.isArray(obj)) {
+    return obj.map(toJson);
+  } else if (typeof obj === "object") {
+    const jsonObj: Json = {};
+    Object.entries(obj as object).forEach(([k, v]) => (jsonObj[k] = toJson(v)));
+    return jsonObj;
+  }
+  throw new Error("Unable to convert unknown to json");
```
dapplion: This throw can potentially prevent the actual error from being logged, `toJson` should never throw and return something

packages/lodestar-utils/src/logger/format.ts
wemeetagain: This should work well for errors we don't control.
The `error.message` is part of the `error.stack` :o)

Eg:
```typescript
logger = new WinstonLogger();
logger.info("A terrific error", new Error("this is an error"));
```
will output
```
2020-09-30 13:38:59  []                 info: A terrific error
Error: this is an error
    at repl:1:33
    at Script.runInThisContext (vm.js:120:18)
    at REPLServer.defaultEval (repl.js:433:29)
    at bound (domain.js:427:14)
    at REPLServer.runBound [as eval] (domain.js:440:12)
    at REPLServer.onLine (repl.js:760:10)
    at REPLServer.emit (events.js:327:22)
    at REPLServer.EventEmitter.emit (domain.js:483:12)
    at REPLServer.Interface._onLine (readline.js:329:10)
    at REPLServer.Interface._line (readline.js:658:8)
```

Edit:
A `BlockError` looks like:
```typescript
logger.info("Uh oh!", new be.BlockError({
  code: be.BlockErrorCode.ERR_FUTURE_SLOT,
  currentSlot: 5,
  blockSlot: 6
}));
```
```
2020-09-30 14:37:25  []                 info: Uh oh! code=ERR_FUTURE_SLOT, currentSlot=5, blockSlot=6
Error: ERR_FUTURE_SLOT
    at repl:1:23
    at Script.runInThisContext (vm.js:120:18)
    at REPLServer.defaultEval (repl.js:433:29)
    at bound (domain.js:427:14)
    at REPLServer.runBound [as eval] (domain.js:440:12)
    at REPLServer.onLine (repl.js:760:10)
    at REPLServer.emit (events.js:327:22)
    at REPLServer.EventEmitter.emit (domain.js:483:12)
    at REPLServer.Interface._onLine (readline.js:329:10)
    at REPLServer.Interface._line (readline.js:658:8)
```

packages/lodestar-utils/src/json.ts
wemeetagain: This should be unreachable, so it doesn't practically matter. But philosophically, I think it makes sense to error and not "error silently" for a general purpose `toJson` function. See our past struggle with this in the lodestar library. We had many "nullish" functions that simply hide error cases.

wemeetagain:
> I'm a bit confused as this doesn't help us much.

This is basically the idea in #1595

This adds:
- generic `toJson` can be used in many places, and especially surrounding the logger:
eg: instead of `logger.info("message", config.types.Foo.toJson(data))` => `logger.info("message", toJson(data))`
- a "pattern" for exposing error data as `Json` - custom errors implement `toJson` method, a default fallback implementation returns `{message, stack}`
- exposes error metadata in the logger - if a custom error is passed to the logger, its metadata is printed

This PR is a minimal change to focus on these things, can be rolled out across the repo to greater effect.
Eg:
one PR to use `toJson` helper where appropriate in calls to logger (simplifies many logger call sites)
one PR to add `toJson` method to other custom errors
one PR to log errors where appropriate - this is where most of the runtime benefit will be, when we will see error metadata printed via the logger.

packages/lodestar-utils/src/logger/format.ts
dapplion: True, tho it is possible for some insane dev to throw Errors that do not have the stack property. Maybe we can log the message only if there is no stack.

packages/lodestar-utils/src/json.ts
dapplion: Then add a stringified json arg to the error message in the edge case it happens we know what data caused the error

packages/lodestar-utils/src/json.ts
```diff
@@ -0,0 +1,38 @@
...
+  } else if (Array.isArray(obj)) {
+    return obj.map(toJson);
+  } else if ((obj as Partial<IToJson>).toJson) {
+    return (obj as IToJson).toJson();
+  } else if (typeof obj === "object") {
+    const jsonObj: Json = {};
+    Object.entries(obj as object).forEach(([k, v]) => (jsonObj[k] = toJson(v)));
+    return jsonObj;
+  }
+  throw new Error("Unable to convert unknown to json: " + obj);
```
dapplion: You are forgetting "boolean" and "functions" types, which will throw and error and are frequent types. I strongly argue against ever throwing an error here.
```ts
function toJson(arg: unknown): Json {
  switch (typeof arg) {
    case "bigint":
    case "symbol":
    case "function":
      return arg.toString();

    case "object":
      if (arg === null) return "null";
      if (Array.isArray(arg)) return arg.map(toJson);
      if (arg instanceof Error) return errorToJson(arg);
      return lodash.mapValues(arg, (value) => toJson(value));

    // Already valid JSON
    case "number":
    case "string":
    case "undefined":
    case "boolean":
    default:
      return arg;
  }
}
```

wemeetagain:
Ready for re-review

dapplion (approved):
I'm a partial author so don't count on this approve too much

PR #1610, opened 2020-09-30
Deprecate block stream

### LINKED ISSUE #1590: Deprecate old block event stream in favour of standard event stream
- change validator to use standard event stream (probably abstract inside validator api)
- remove old block event stream endpoint

mpetrunic:
resolves #1590


wemeetagain (dismissed):
:+1: :+1:

wemeetagain:
linter errors

wemeetagain (approved):

PR #1612, opened 2020-09-30
Emit ancestor root for unknownBlockRoot event

### LINKED ISSUE #1611: Improve unknownBlockRoot event emit in block pool
Currently, if we have a branch of blocks like block 1000 -> 1001 -> 1002 in block pools, when block 1003 comes, which is child of 1002, we emit parentRoot of 1003 which is 1002 as "unknownBlockRoot" event, but we have block 1002 in pool already.

We should emit the ancestor root of that branch instead, which is parentRoot of 1000.

In regular sync, I see we have a lot of blocks in pool but we keep emitting the latest block parent which we can't find in any sync peers because it's quite new, something like `410504,410505,410506,410507,410508,410510,410511,410489,410489,410491,410494,410513,410514,410488,410487,410485`
We should emit parent root of 410504, which is probably 410503 and gradually we go overlap to the followup beacon_block_by_range requests of regular sync faster.

twoeths:
resolves #1611

### DIFF
--- a/packages/lodestar/src/chain/blocks/pool.ts
+++ b/packages/lodestar/src/chain/blocks/pool.ts
@@ -6,6 +6,7 @@ import {toHexString} from "@chainsafe/ssz";
 import {IBlockProcessJob} from "../interface";
 import {ChainEventEmitter} from "../emitter";
 import {IForkChoice} from "@chainsafe/lodestar-fork-choice";
+import {findUnknownAncestor} from "./util";
 
 export class BlockPool {
   private unknownParentBlockPool = new Map<string, IBlockProcessJob[]>();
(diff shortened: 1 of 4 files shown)


mpetrunic (commented):
Wouldn't it make more sense just to check if NewBlock.parentRoot is in Object.keys(unknownParentBlockPool) or in pendingSlotBlockPool and if it is do nothing.

My reasoning is that if we already have parent block in pool. Parent block already emitted event for it's missing parent

twoeths:
> Wouldn't it make more sense just to check if NewBlock.parentRoot is in Object.keys(unknownParentBlockPool) or in pendingSlotBlockPool and if it is do nothing.
> 
> My reasoning is that if we already have parent block in pool. Parent block already emitted event for it's missing parent

In a lot of cases we don't have gossip block parent bc it's too new to all sync peers at that time, and the whole unknown block pools keep growing due to 1 single missing ancestor block root.
+ At the time of regular sync, it'd help our node to be in synced faster
+ At the time of synced, if we don't do that our node may become out of synced again due to 1 missing ancestor root. So reemitting unknown ancestor root would help avoid that bc at that time, the ancestor is not so new to all sync peers anymore.

wemeetagain (approved):

PR #1616, opened 2020-10-01
Error logger formatting - proposal 2

dapplion:
Extends https://github.com/ChainSafe/lodestar/pull/1609 with some opinionated decisions.

- Adds a placeholder implementation of a human-readable and a JSON logger. Can be configured with the logger option `format`
- Adds two "serialization" functions: `toJson(arg: unknown): Json` and `toString(json: Json): string`. These can then are used in the winston format functions. They never throw and I've added many test cases for most possible types.
- Adds a `LodestarError` which all custom errors should extend and which contains a `toObj` method. `toObj` returns the data that should be printed by the logger associated to that error. Extenders of `LodestarError` can customize that function if necessary.

#### Goals

- If a `LodestarError` is catched by external code it will have a message property with that error's code
- When a `LodestarError` is logged by our logger there will not be duplicate information
- Serializing a `LodestarError` in our logger is type-safe without castings
- Our logger will log external errors with their message property as well as all their metadata


wemeetagain (approved):
I really like this direction.
`LodestarError` will make it easier to standardize errors across all of the libraries.
Also, better implementation of `toJson` and nice cleanup of the logger formatting code.

wemeetagain (approved):

PR #1618, opened 2020-10-01
Regular sync: Handle empty range and avoid listening to gossip blocks

### LINKED ISSUE #1614: Regular Sync is stalled: do not trust empty range result
**Describe the bug**
Currently in regular sync, if we receive an empty range and later on we check best peer's head is above range end, we assume that it's really an empty range and we just move forward.
In reality, the peer may return beacon_block_by_range incorrectly, or because peer head is updated after we call beacon_block_by_range. So we have to be safe: should start from range.start and expand range.end.

```
2020-10-01 10:09:44  [SYNC]             info: Regular Sync: Setting target newTargetSlot=414580
2020-10-01 10:09:44  [SYNC]             info: Regular Sync: Requesting blocks from slot 414561 to slot 414580
2020-10-01 10:09:50  [SYNC]             info: Imported blocks for slots: 414561,414562,414564,414565,414567,414568,414569,414570,414571,414572,414573,414574,414575,414576,414577,414578,414579,414580
2020-10-01 10:10:52  [SYNC]             info: Regular Sync: Synced up to slot 414580  currentSlot=414653, gossipParentBlockRoot=undefined
2020-10-01 10:10:52  [SYNC]             info: Regular Sync: Requesting blocks from slot 414581 to slot 414600
2020-10-01 10:10:55  [SYNC]             info: Imported blocks for slots:
2020-10-01 10:10:55  [SYNC]             info: Regular Sync: Requesting blocks from slot 414601 to slot 414620
2020-10-01 10:10:55  [SYNC]             info: Regular Sync: wait for best peer
2020-10-01 10:11:36  [SYNC]             info: Imported blocks for slots: 414601,414602,414603,414604,414606,414607,414608,414611,414612,414614,414615,414616,414617,414618,414619,414620
```
In the above log, fetching `414581 -> 414600` returns empty range (without error) and we trust that and move forward, but since we really have some blocks in that range, Regular Sync stalled (bc no "processedBlock" event emitted due to missing blocks in the middle) We should not trust that and fetch `414581 -> 414620` to be safe if possible.
**Expected behavior**
Regular sync is not stalled even best peer does not return correctly or whatever reason in the case of empty blocks returned.

twoeths:
resolves #1614 
This is all about regular sync:
+ When best peer returns an empty range, do not trust that. Just use same start and expand end.
+ Do not listen for gossip blocks, otherwise it'll trigger unexpected fetch
+ Tweak best peer min score, set bestPeer to undefined in order to find a new best peer

### DIFF
--- a/packages/lodestar/src/network/peers/score/simpleRpcScore.ts
+++ b/packages/lodestar/src/network/peers/score/simpleRpcScore.ts
@@ -11,7 +11,7 @@ const scoreConstants: Record<RpcScoreEvent, number> = {
   [RpcScoreEvent.UNKNOWN_ERROR]: -10,
 };
 
-const DEFAULT_SCORE = 100;
+export const DEFAULT_SCORE = 100;
 const MAX_SCORE = 200;
 const MIN_SCORE = 0;
 
--- a/packages/lodestar/src/network/peers/score/simpleRpcScore.ts
+++ b/packages/lodestar/src/network/peers/score/simpleRpcScore.ts
@@ -11,7 +11,7 @@ const scoreConstants: Record<RpcScoreEvent, number> = {
   [RpcScoreEvent.UNKNOWN_ERROR]: -10,
 };
 
-export const DEFAULT_SCORE = 100;
+export const DEFAULT_RPC_SCORE = 100;
 const MAX_SCORE = 200;
 const MIN_SCORE = 0;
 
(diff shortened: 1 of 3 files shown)


packages/lodestar/src/network/peers/score/simpleRpcScore.ts
```diff
@@ -11,7 +11,7 @@ const scoreConstants: Record<RpcScoreEvent, number> = {
   [RpcScoreEvent.UNKNOWN_ERROR]: -10,
 };
 
-const DEFAULT_SCORE = 100;
+export const DEFAULT_SCORE = 100;
```
mpetrunic: if you are exporting, pls rename to DEFAULT_RPC_SCORE

mpetrunic (commented):
I don't think it's wise not to listen gossip blocks during regular sync since we need to be aware of other head branches and sync them as well. If you are getting to many fetches, best way would be to reduce unknown parent depth filter:
https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/chain/blocks/pool.ts#L39

twoeths:
> I don't think it's wise not to listen gossip blocks during regular sync since we need to be aware of other head branches and sync them as well. If you are getting to many fetches, best way would be to reduce unknown parent depth filter:

the issue is not about having too many fetches but the control of `currentTarget` variable in naive strategy. If we have multiple fetches, we can't control that variable. I think the original design of naive stragegy is to have 1 single fetch, if we want multiple fetches then probably we should do that in another strategy

wemeetagain (approved):

PR #1620, opened 2020-10-02
Update validator event handling

wemeetagain:
Update validator event handling to mimic the style in the beacon node.
Currently there is a lot of code duplication. In the future, we should reorganize our packages to allow for more code reuse.

- Add a `BeaconEventEmitter` to the validator (similar to `ChainEventEmitter`, but for consuming clock + events API data)
- Add a `LocalClock` to the validator (with minor tweaks)
- Update block and attestation services to:
  - subscribe to the `BeaconEventEmitter` for clock events
  - re-request duties when a new head+epoch_transition event is emitted (eg: a new head checkpoint is found)

Results: no more errors in the multithreaded e2e test!~


packages/lodestar-validator/src/api/impl/rest/events/events.ts
```diff
@@ -30,4 +31,21 @@ export class RestEventsApi implements IEventsApi {
       };
     });
   }
+
+  public getEventEmitter(topics: BeaconEventType[], signal: AbortSignal): EventSourceBeaconEventEmitter {
```
mpetrunic: Why not on start use getEventStream and emit events from there instead of duplicate implementation with different interface?

packages/lodestar-validator/src/api/interface/events.ts
```diff
@@ -1,6 +1,58 @@
...
+import {BlockEventPayload, ChainHead, ChainReorg} from "@chainsafe/lodestar-types";
 import {IStoppableEventIterable} from "@chainsafe/lodestar-utils";
-import {BeaconEvent, BeaconEventType} from "../impl/rest/events/types";
+
+export enum BeaconEventType {
+  BLOCK = "block",
+  CHAIN_REORG = "chain_reorg",
+  HEAD = "head",
+  CLOCK_SLOT = "clock_slot",
+  CLOCK_EPOCH = "clock_epoch",
```
mpetrunic: I think we should try to upstream those first, before adding it here. I also have personal interested in this since we would like lodestar-validator to work with lighthouse in Chainguardian.

wemeetagain: These events are not being sent via the API. A local clock is reusing the event emitter to emit clock events.
Definitely would be interested to get the clock exposed in the standard API tho

packages/lodestar-validator/src/api/impl/rest/events/events.ts
wemeetagain: I guess because this implementation is a lot simpler this way? Not having to deal with async functions or generators inside the EventEmitter implementation.

mpetrunic: Hmm but you can implement this without using EventSource directly. 
Basically:
```typescript
getEventEmmitter(topics) {

   for await (const event in this.getEventStream(topics)) {
      something.emit(event.topic, event.message);
   }

}
```
And if you put it as utility function somewhere, you don't have to reimplement on lodestar side

mpetrunic (commented):
Seems like your types wen't a bit mayhem, would you mind me fixing those?

mpetrunic (approved):
Asked in typescript discord for help so we have proper typing for pipeToEmitter! :)

PR #1628, opened 2020-10-04
Add retry on large eth1 requests

dapplion:
Current eth1 module implementation can request very large amounts of data, which some Eth1 nodes seem to return with a success 200 HTTP code but the body of the request is truncated and contains invalid JSON. 

This PR adds a retry wrapper on the batch blocks request and the logs request. On error, if the error is caused to bad JSON, it will retry the same range of blocks but splitting the range requested into smaller chunks (original range /1, /2, /4).

### DIFF
--- a/packages/lodestar/src/eth1/eth1JsonRpcClient.ts
+++ b/packages/lodestar/src/eth1/eth1JsonRpcClient.ts
@@ -1,7 +1,10 @@
 import {fromHexString} from "@chainsafe/ssz";
 import {Eth1Block} from "@chainsafe/lodestar-types";
 import {AbortSignal} from "abort-controller";
-import {JsonRpcHttpClient} from "./jsonRpcHttpClient";
+import {chunkifyInclusiveRange} from "../util/chunkify";
+import {linspace} from "../util/numpy";
+import {retry} from "../util/retry";
+import {ErrorParseJson, JsonRpcHttpClient} from "./jsonRpcHttpClient";
 
 /**
  * Binds return types to Ethereum JSON RPC methods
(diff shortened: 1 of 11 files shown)


packages/lodestar/src/eth1/eth1JsonRpcClient.ts
```diff
@@ -37,13 +40,31 @@ export class Eth1JsonRpcClient {
...
+        // Large batch requests can return with code 200 but truncated, with broken JSON
+        // This retry will split a given block range into smaller ranges exponentially
+        // The underlying http client should handle network errors and retry
+        const chunkCount = 2 ** (attempt - 1);
+        const blockRanges = chunkifyInclusiveRange(fromBlock, toBlock, chunkCount);
+        return Promise.all(
+          blockRanges.map(([from, to]) =>
+            this.rpc.fetchBatch<IEthJsonRpcTypes[typeof method]>(
+              linspace(from, to).map((blockNumber) => ({method, params: [toHex(blockNumber), false]})),
+              signal
```
mpetrunic, line 56: if signal is activated, won't rpc client throw and this will retry?

packages/lodestar/src/util/chunkify.ts
```diff
@@ -0,0 +1,27 @@
...
+ * ```
+ * [[a,b], [c,d] ... Sn] = chunkifyInclusiveRange([a,z], n)
+ * // where
+ * [a,z] = [a,b] U [c,d] U ... U Sn
+ * ```
+ * @param from range start inclusive
+ * @param to range end inclusive
+ * @param chunks Maximum number of chunks, if range is big enough
+ */
+export function chunkifyInclusiveRange(from: number, to: number, chunkCount: number): number[][] {
```
mpetrunic, line 12: isn't sync chunkify doing same thing?

packages/lodestar/src/util/retry.ts
```diff
@@ -0,0 +1,36 @@
...
+  const shouldRetry = opts?.shouldRetry;
+
+  let lastError: Error = Error("RetryError");
+  for (let i = 1; i <= maxRetries; i++) {
+    try {
+      return await fn(i);
+    } catch (e) {
+      lastError = e;
+      if (shouldRetry && !shouldRetry(lastError)) {
+        break;
```
mpetrunic, line 31: you could just throw current error here?

packages/lodestar/src/eth1/eth1JsonRpcClient.ts
dapplion: It will only retry if the error is an instance of `ErrorParseJson`

packages/lodestar/src/util/chunkify.ts
dapplion: Right, didn't know that function existed. What do you suggest to do, to move sync chunkify to utils and make it more generic?

packages/lodestar/src/util/retry.ts
dapplion: Yes, but I felt it was repeating code

packages/lodestar/src/util/chunkify.ts
mpetrunic: Hmm yeah, let's do that

dapplion: Can we do that in another PR not minimize the diff?

mpetrunic: hmm shouldn't be too much diff, it had identical interface, right?

dapplion: No, the interface is different. The `chunkifyInclusiveRange` of this PR passes as parameter `chunkCount` while `chunkify` of sync passes `itemsPerChunk`

mpetrunic (approved):

wemeetagain (approved):

PR #1629, opened 2020-10-05
Backward sync controlled by sync

### LINKED ISSUE #1346: Backward sync controlled by sync

twoeths:
resolves #1346 
+ block pool always emit `unknownBlockRoot`, which is a missing ancestor root. Consumers decide how to handle it.
+ For sync, it'll handle `unknownBlockRoot` from the time of regular sync and keep doing that until the node is stopped.

This should help the current regular sync work better as gossip/pending blocks will grow in 2 directions and will soon overlap with the follow up `beacon_blocks_by_range` fetch.

### DIFF
--- a/packages/lodestar/src/chain/blocks/pool.ts
+++ b/packages/lodestar/src/chain/blocks/pool.ts
@@ -5,7 +5,6 @@ import {toHexString} from "@chainsafe/ssz";
 
 import {IBlockProcessJob} from "../interface";
 import {ChainEventEmitter} from "../emitter";
-import {IForkChoice} from "@chainsafe/lodestar-fork-choice";
 import {findUnknownAncestor} from "./util";
 
 export class BlockPool {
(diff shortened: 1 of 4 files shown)


mpetrunic (commented):
Hmm but we still need that logic in pending block pool in case of missing out on gossip block. 

I would suggest that sync searches for missing blocks before submitting block range instead of emitting event, wdyt?

packages/lodestar/src/chain/blocks/pool.ts
```diff
@@ -36,15 +28,12 @@ export class BlockPool {
...
-      //this prevents backward syncing, tolerance is 20 blocks
-      if (job.signedBlock.message.slot <= this.forkChoice.getHead().slot + 20) {
-        const pendingBlocks = Array.from(this.unknownParentBlockPool.values()).flat();
-        this.eventBus.emit(
-          "unknownBlockRoot",
-          findUnknownAncestor(this.config, pendingBlocks, job.signedBlock.message.parentRoot)
-        );
-      }
     }
+    const pendingBlocks = Array.from(this.unknownParentBlockPool.values()).flat();
```
twoeths, line 32: @mpetrunic the main change is to emit `unknownBlockRoot` without caring about forkchoice head, logic is still the same. In the case of missing gossip blocks, sync would search for them.

mpetrunic: Wait but this then allows backwards sync? If we receive block far away from our head we would start syncing backwards as each time it finds missing block it would emit events again?

twoeths: yeah we allow that from regular sync time, for initial sync we don't allow backward sync.

I think from regular sync, syncing backward is a good thing bc it helps the regular sync to be finished faster: regular sync starts from our head and do `beacon_blocks_by_range` requests. At the same time gossip blocks come and they emit `unknownBlockRoot` for missing ancestor root and soon it'd overlap with following up `beacon_blocks_by_range` requests (since they'll be in 2 directions) and then the node is fully synced.

Also with the latest change, we emit the missing ancestor root instead of the parent block root.

twoeths:
need to work with the latest block processor/pool

PR #1630, opened 2020-10-05
Unify validator event emitter and slight type correction

mpetrunic:


packages/lodestar-validator/src/api/impl/rest/events/util.ts
```diff
@@ -22,3 +24,16 @@ export function deserializeBeaconEventMessage(config: IBeaconConfig, msg: Messag
...
 }
+
+export async function pipeToEmitter(
+  stream: IStoppableEventIterable<BeaconEvent>,
+  emitter: ApiClientEventEmitter
+): Promise<void> {
+  for await (const evt of stream) {
+    //I have no idea how to type this.
+    // I've tried conditional typing but it didn't work
+    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
```
wemeetagain, line 35: Only way I saw to do this is to type it as a specific type of event/payload. See the old version of this fn.

The reason it doesn't seem to work with eg: `BlockEventPayload[keyof BlockEventPayload]` is that these approaches create a union type that doesn't match the type needed for the emitter.

mpetrunic: Yeah, not sure if it's possible to extract type based on type property. 
I've tried conditional types:
```typescript
type GetEventPayload<T: BeaconEventType> = BeaconEvent extends {type: T} ? BeaconEvent : never
```
But resulting type is always union. If I make BeaconEvent generic, I have to manually type possible events.

I just wanted to remove as unknown and double event emitter, than I noticed this haha

wemeetagain (approved):

PR #1633, opened 2020-10-05
Update block/attestation processor/pool

wemeetagain:
- block processor
  - use `JobQueue` for queueing
  - use `IBlockJob` for input (prereq for skipping gossip/etc validations)
  - `validateBlock`/`processBlock` simplified, now just normal async functions
  - remove all side-effects from the processor, all now handled in event handlers in `src/chain/chain.ts`
- block pool
  - now exposed to `src/chain/chain.ts`
  - exists as a cache
- attestation processor
  - split into files similarly to block processor
- attestation pool
  - similar to block pool
- `src/chain/chain.ts`
  - add event handlers for `block` event, used to update block/attestation pool
  - update `checkpoint` event handler, use to emit `justified`/`finalized` checkpoint event
  - update `error:block` event handler, selectively push blocks into the block pool or add to db.badBlock
  - add `error:attestation` event handler, selectively push blocks into the block pool or remove from db
  - add informational handlers for `forkChoice:head` and `forkChoice:reorg`


packages/lodestar/src/chain/attestation/pool.ts
```diff
@@ -1,161 +1,134 @@
...
-    for (const blockRoot of [target.root, attestation.data.beaconBlockRoot]) {
-      if (!this.forkChoice.getBlock(blockRoot)) {
-        this.logger.verbose("Delaying attestation", {
-          reason: "missing either target or attestation block",
-          blockRoot: toHexString(blockRoot),
-          ...attestationLogContext,
-        });
-        this.addPendingBlockAttestation(blockRoot, attestation, attestationHash);
-        return;
+    const bbrKey = toHexString(job.attestation.data.beaconBlockRoot);
```
mpetrunic: can we use more descriptive names instead.
Like `attestationBlockRootKey`

packages/lodestar/src/chain/attestation/pool.ts
```diff
@@ -1,161 +1,134 @@
...
+    const attestationsAtBbr = this.attestationsByBlock.get(bbrKey);
```
mpetrunic: pendingBlockAttestations ?

packages/lodestar/src/chain/blocks/pool.ts
```diff
@@ -1,81 +1,128 @@
...
-      }
+  public addByParent(job: IBlockJob): void {
+    const {signedBlock} = job;
+    // put block in two indices:
+    // blocks
+    const blockKey = this.getBlockKey(signedBlock);
+    this.blocks.set(blockKey, job);
+    // blocks by parent
+    const parentKey = this.getParentKey(signedBlock);
+    let blocksWithParent = this.blocksByParent.get(parentKey);
```
mpetrunic, line 45: blocksWithoutParent?

packages/lodestar/src/chain/blocks/pool.ts
```diff
@@ -1,81 +1,128 @@
...
-    const key = toHexString(this.config.types.BeaconBlock.hashTreeRoot(block.message));
-    const jobs = this.unknownParentBlockPool.get(key);
-    if (jobs) {
-      this.unknownParentBlockPool.delete(key);
-      jobs
-        .sort((a, b) => a.signedBlock.message.slot - b.signedBlock.message.slot)
-        .forEach((job) => {
-          this.blockProcessorSource.push(job);
-        });
+  public remove(job: IBlockJob): void {
```
mpetrunic, line 69: why is this public?

packages/lodestar/src/chain/blocks/pool.ts
```diff
@@ -1,81 +1,128 @@
...
-    forkChoice: IForkChoice
-  ) {
+  constructor({config, processor}: {config: IBeaconConfig; processor: BlockProcessor}) {
     this.config = config;
-    this.blockProcessorSource = blockProcessorSource;
-    this.eventBus = eventBus;
-    this.forkChoice = forkChoice;
+    this.processor = processor;
+
+    this.blocks = new Map();
```
mpetrunic, line 32: why do we need this?

packages/lodestar/src/chain/blocks/pool.ts
```diff
@@ -1,81 +1,128 @@
...
+   */
+  public blocks: Map<string, IBlockJob>;
+  /**
+   * Blocks indexed by parentRoot, then blockRoot
+   */
+  public blocksByParent: Map<string, Map<string, IBlockJob>>;
+  /**
+   * Blocks indexed by slot, then blockRoot
+   */
+  public blocksBySlot: Map<Slot, Map<string, IBlockJob>>;
```
mpetrunic: I think we should expose getters like "getFutureBlocks" and "getBlocksWithoutParent" that returns IBlockJob[]

packages/lodestar/src/chain/chain.ts
```diff
@@ -485,14 +534,114 @@ export class BeaconChain implements IBeaconChain {
...
+      case BlockErrorCode.ERR_PER_BLOCK_PROCESSING_ERROR:
+      case BlockErrorCode.ERR_BLOCK_IS_NOT_LATER_THAN_PARENT:
+      case BlockErrorCode.ERR_UNKNOWN_PROPOSER:
+        await this.db.badBlock.put(blockRoot);
+        this.logger.warn("Found bad block", {
+          blockRoot: toHexString(blockRoot),
+          error: toJson(err),
+        });
+        break;
+    }
```
mpetrunic, line 675: I think we can separate handlers into separate files/functions but I can do it after we merge this since this diff is already huge

packages/lodestar/src/chain/blocks/pool.ts
wemeetagain: It's not used currently, its just something that seems like we'll eventually need.
Eg: for quickly checking if we've 'seen' a block.

packages/lodestar/src/chain/blocks/pool.ts
wemeetagain: We're not using this now, but removing blocks from the pool is a natural compliment to adding blocks to the pool. Gives us more flexibility for pruning strategies.

packages/lodestar/src/chain/attestation/pool.ts
wemeetagain: agree bbr is a little too dense. will go with `beaconBlockRoot` instead

packages/lodestar/src/chain/blocks/pool.ts
mpetrunic: @wemeetagain seems like you missed this one

mpetrunic:
Looks good to me, there is just that tiny comment and type errors in unit tests

packages/lodestar/src/chain/blocks/pool.ts
wemeetagain: I think this should still be called `blocksWithParent`, or perhaps `blocksAtParent`.
These blocks all _do_ have a parent, its the key that they're all indexed by.

mpetrunic: Yeah but they are indexed by it because it's missing. I find current name quite confusing though

wemeetagain: Think of this module as a _just_ a cache with three indices. blocks by root, parentRoot, slot.
when we get blocks by slot, these aren't blocks _without_ a slot. They're blocks _at_ a slot.
Similarly, when we get a block by root, its a block _with/of_ a root.
Same thinking goes for blocks by parent root.

mpetrunic: But they are blocks without slot, their slot has not yet come or it has already passed. I don't feel like arguing about that anymore but if I find it confusing and I do know what this is about it's highly likely other contributors would as well. At least put comment explaining it since it's not really self descriptive variable name

dapplion (approved):

PR #1635, opened 2020-10-05
Run lodestar with node directly not yarn

dapplion:
I've found that `yarn` does a really bad job at handling child script signals. On CTRL+C it only forwards the first SIGINT and ignores the rest, breaking our graceful shutdown pattern. `npm` does propagate signals correctly but I would recommend users to run lodestar directly with node to prevent further issues.

Also Windows, MacOS, Linux handle CTRL+C differently so the simpler the better

#### Test

Try running
```
node --trace-deprecation ./packages/lodestar-cli/bin/lodestar beacon --testnet spadina
```
vs
```
yarn run cli beacon --testnet spadina
```
And look for this log https://github.com/ChainSafe/lodestar/blob/d9fcdad8a5ea6e43515bdd6e4ed1c4012eb7332b/packages/lodestar-cli/src/util/process.ts#L19 after hitting CTRL+C twice fast 

#### Solution

I propose to run the CLI bin file directly. In this PR I've added a softlink to it, but I'm open to better suggestions if any.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -34,7 +34,7 @@ yarn run build
 Lodestar should now be ready for use:
 
 ```bash
-yarn run cli --help
+./lodestar --help
 ```
 
 ## Install from NPM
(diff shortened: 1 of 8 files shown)


mpetrunic (approved):
I'm fine with this

wemeetagain (approved):

PR #1638, opened 2020-10-06
Standardize fork endpoint

mpetrunic:
- update fork endpoint to be inline with standardized api
- add wrapper for beacon api implementation to conform with validator api interface for ApiOverInstance implementation (easier than replicating api structure in validator client)


packages/lodestar-validator/src/api/types/beacon.ts
```diff
@@ -0,0 +1,11 @@
+import {BLSPubkey, Fork, Genesis, ValidatorResponse} from "@chainsafe/lodestar-types";
+
+export interface IBeaconApiClient {
```
wemeetagain, line 3: I don't think the name should be changed from `IBeaconApi`. We want to remain standardized with the interface in the `lodestar` package. Eventually we'll want to break out lodestar into packages and use the interface directly.

wemeetagain (changes requested):

packages/lodestar-validator/src/api/types/beacon.ts
mpetrunic: We can discuss this more, but  lodestar api is growing in complexity and has namespaced methods which is hard to replicate in lodestar-validator (and it complicates validator for no reason). I also found it a lot confusing because you never know where your type errors are or if you are using/importing proper interface.

I'm in favor of defining api client interface in lodestar-validator with only methods lodestar-validator needs and in cli where we use ApiClientOverInstance, we use some lodestar api wrapper that implements validator api client interface and calls lodestar api implementation. It would make everything a lot more verbose and easier to read and understand

Example: https://github.com/ChainSafe/lodestar/pull/1638/files#diff-40ecd6cb1b00979ab71468039e53fb61R5

wemeetagain: I don't really understand why we need _any_ divergence between how the API works on the `lodestar` side vs the `lodestar-validator` side.
imo `ApiClientOverInstance` should not even be necessary, we should just be using a `beaconNode.api` (of type `Api`) directly. If there are interface differences, they should be resolved, bc ultimately `Api` and `ApiClient` should behave _exactly the same_ (eg: have the same interface, `IApi`). The only difference is that one calls into a beacon node directly, and the other is remote. 
We should literally copy-paste `IApi` interfaces from `lodestar` and `lodestar-validator` if we need to, until we can unify the the imports. The ideal that we are moving towards is: we have that single interface that is imported in both `lodestar` and `lodestar-validator`.

If we aren't using the entire API in the validator, then a good way of thinking of this is:
We should be building a reusable `ApiClient` that supports the full standard API, and 'importing' the client for use in `lodestar-validator`.

mpetrunic: > imo ApiClientOverInstance should not even be necessary, we should just be using a beaconNode.api (of type Api) directly.

That would mean we depend on lodestar package and it would make this much simpler but since we aren't, best way it to make code simple on lodestar-validator and just wrap it. Wrapper is just used in dev anyways.

Trying to keep lodestar-validator and lodestar api interface in sync is quite difficult and confusing. Given the fact that lodestar api has multiple levels it makes it even weirder when implementing and using on lodestar-validator side since you need to create a lot of dumb interfaces and they have to match those on lodestar side. When implemnting that api over rest you need to support all params required by interface which are useless for lodestar-validator.

Third thing is that we are over-complicating api interface on lodestar-validator side for anyone who wishes to use it with some other than lodestar client

wemeetagain: > That would mean we depend on lodestar package and it would make this much simpler but since we aren't, best way it to make code simple on lodestar-validator and just wrap it. 

It only needs to be wrapped if the interfaces are not compatible. We should be moving in that direction.
The original point made in this thread is that we need to minimize divergences where we can. Not add a superfluous adapter layer, but make the minimal changes to let the `IBeaconApi` on the lodestar-validator side be a subset of the `IBeaconApi` on the lodestar side.

> Trying to keep lodestar-validator and lodestar api interface in sync is quite difficult and confusing.

We don't need the full standard API right now, but the pieces we _are_ using should be in sync. I'm willing to help if you want.

> Third thing is that we are over-complicating api interface on lodestar-validator side for anyone who wishes to use it with some other than lodestar client

I disagree, we're setting the prerequisites for pulling out the api client into its own package. The end result is that the validator won't deal with api client logic at all.
Its much more difficult to do this if we're building in incompatibilities.

mpetrunic: It's not that it's much work but it's completely wrong work, we are making our validator client api interface dependent on lodestar internal api structure so if we ever decide to change lodestar api structure we have to make breaking changes on lodestar-validator which we did 3 times since we started using lodestar-validator in Chainguardian.
If we decide to move into direction of breaking lodestar-validator interface yet again and make it dependent on lodestar api internal structure, we will write our own implementation of eth2 validator on Chainguardian side. 

I understand we are in heavy development and we have to break api but this in unacceptable.

PR #1651, opened 2020-10-09
Extract chain event handlers to separate files

### LINKED ISSUE #1644: Separate chain event handlers into separate files
I think we can separate handlers into separate files/functions but I can do it after we merge this since this diff is already huge

_Originally posted by @mpetrunic in https://github.com/ChainSafe/lodestar/pull/1633#r500049602_

mpetrunic:
resolves #1644


mpetrunic:
Why did you do that awful last commit?

They were already typesafe and we didn't had huge file.

wemeetagain:
IMO it's a lot easier to manage all of the side effects in a single file for now

You're right, it was typesafe tho :joy:

mpetrunic:
What kind of screen you have where you can navigate 300 lines of function code? 

I think we should aim to keep files <50 lines of code , maybe even enforce it in eslint as looking for effect in this huge file is nightmare 😅

Can you rollback to the cleaner typesafe code at least(without all those register/deregister)?

wemeetagain (approved):

PR #1655, opened 2020-10-12
Sync: Handle unknown block root event from regular sync

### LINKED ISSUE #1346: Backward sync controlled by sync

twoeths:
resolves #1346 

+ This is based on the new block pool work
+ Update sync status to` SYNCED` on `syncCompleted` event
+ Sync: Separate `getUnknownRootPeers` and `getSyncPeer`


packages/lodestar/src/chain/emitter.ts
```diff
@@ -40,6 +39,8 @@ export interface IChainEvents {
   "forkChoice:reorg": (head: IBlockSummary, oldHead: IBlockSummary, depth: number) => void;
   "forkChoice:justified": (checkpoint: Checkpoint) => void;
   "forkChoice:finalized": (checkpoint: Checkpoint) => void;
+
+  "block:unknownRoot": (root: Root) => void;
```
mpetrunic: isn't that present in error:block?

twoeths: when doing block `validate` we throw `error:block` with `ERR_PARENT_UNKNOWN` code and chain handle that, then chain puts to pending block pool and emits `block:unknownRoot` with missing ancestor root then sync will handle that.
So I'd separate `error:block` and `block:unknownRoot` event.

mpetrunic: hmm but do we need to fire block:unknownRoot from pool again, I mean everyone can listen to error:block and check if code is missing parent, right?

twoeths: I thought about it, my understanding is `BlockPool` merely a data storage, chain is the controller so I did not fire `block:unknownRoot` inside block pool and let chain does the job.
cc @wemeetagain to see if this is the correct way

twoeths:
Update: in regular sync, we want to handle ERR_BLOCK_IS_ALREADY_KNOWN because it maybe auto-processed by pending block pool already (previously we still emit "block" event in `validate.ts` but now we switch to ERR_BLOCK_IS_ALREADY_KNOWN block error code)

wemeetagain (approved):

PR #1659, opened 2020-10-13
Add attestation logic for slashing protection

dapplion:
Adds attestation logic for a refactored slashing protection module compatible with the "Slashing Protection Database Interchange Format"

### DIFF
--- a/packages/lodestar-db/src/schema.ts
+++ b/packages/lodestar-db/src/schema.ts
@@ -35,6 +35,8 @@ export enum Bucket {
   validator = 16,
   lastProposedBlock = 17,
   proposedAttestations = 18,
+  // validator slashing protection
+  slashingProtectionAttestation = 21,
 }
 
 export enum Key {
--- a/packages/lodestar-validator/src/slashingProtection/attestation/checkAndInsertAttestation.ts
+++ b/packages/lodestar-validator/src/slashingProtection/attestation/checkAndInsertAttestation.ts
@@ -24,9 +24,8 @@ export async function checkAndInsertAttestation(
   // Merging those ranges
   //   0 <= s2 < t1 - 1
   //   s1 + 1 <= t2
-  const potentialAttestationConflicts = await signedAttestationDb.getByPubkeyAndTargetEpoch(pubKey, {
-    gte: att.sourceEpoch + 1,
-  });
+  // EDIT: fetch the entire range or tests do not pass
+  const potentialAttestationConflicts = await signedAttestationDb.getByPubkeyAndTargetEpoch(pubKey, {});
 
   // Although it's not required to avoid slashing, we disallow attestations
   // which are obviously invalid by virtue of their source epoch exceeding their target.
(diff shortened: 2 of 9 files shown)


packages/lodestar-db/src/schema.ts
```diff
@@ -35,6 +35,8 @@ export enum Bucket {
   validator = 16,
   lastProposedBlock = 17,
   proposedAttestations = 18,
+  // validator slashing protection
```
mpetrunic, line 38: bucket above is also used for same thing, so you can probably drop it?

packages/lodestar-validator/src/slashingProtection/attestation/checkAndInsertAttestation.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  const minTargetEpoch = minEpoch(potentialAttestationConflicts.map((att) => att.targetEpoch));
+  if (minTargetEpoch && att.targetEpoch <= minTargetEpoch) {
+    throw new InvalidAttestationError({
+      code: InvalidAttestationErrorCode.TARGET_LESS_THAN_OR_EQ_LOWER_BOUND,
+      targetEpoch: att.targetEpoch,
+      minTargetEpoch,
+    });
+  }
```
mpetrunic, line 76: I counted 5 iterations over unbounded array. Can't we do all of that in single loop (find min source and target epoch as well as other checks)?

packages/lodestar-db/src/schema.ts
dapplion: By dropping you mean using the `18` bucket? We should discuss how to handle changes in the DB. If we change the schema and users update it will break in a very nasty form and the db would have to be wiped out, all without any helpful message.

packages/lodestar-validator/src/slashingProtection/attestation/checkAndInsertAttestation.ts
dapplion: This code will be optimized with https://github.com/ChainSafe/lodestar/tree/dapplion/slashing-protection/min-max-surround which is able to pick both surround votes with O(1). Then we can query the DB directly for the double vote and both minimum bounds, so 5 O(1) lookups.

packages/lodestar-db/src/schema.ts
mpetrunic: I thought you are going to integrate slashing protection into Vc in this PR in which case you need to replace `proposedAttestations = 18` and it's usage with your own

dapplion: But the schema for the binary data stored in bucket 18 will be different. Once users run the client they will get a nasty SSZ decode error.

mpetrunic: True, we can just delete it and keep a skip bucket index

dapplion:
Superseded by https://github.com/ChainSafe/lodestar/pull/1674

packages/lodestar-db/src/schema.ts
dapplion: Suggestion applied in https://github.com/ChainSafe/lodestar/pull/1674/commits/e879f745b0f2624bd0b34b4bac0edc0154627acc

PR #1660, opened 2020-10-13
Add block proposal logic for slashing protection

dapplion:
Adds block logic for a refactored slashing protection module compatible with the "Slashing Protection Database Interchange Format"

### DIFF
--- a/packages/lodestar-validator/src/slashingProtection/utils.ts
+++ b/packages/lodestar-validator/src/slashingProtection/utils.ts
@@ -1,0 +1,15 @@
+import {Epoch, Root} from "@chainsafe/lodestar-types";
+import {toHexString} from "@chainsafe/ssz";
+import {ZERO_HASH} from "@chainsafe/lodestar-beacon-state-transition";
+
+export function isEqualRoot(root1: Root, root2: Root): boolean {
+  return toHexString(root1) === toHexString(root2);
+}
+
+export function isZeroRoot(root: Root): boolean {
+  return isEqualRoot(root, ZERO_HASH);
+}
+
+export function minEpoch(epochs: Epoch[]): Epoch | null {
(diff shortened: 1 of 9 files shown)


packages/lodestar-validator/src/slashingProtection/utils.ts
```diff
@@ -0,0 +1,15 @@
+import {Epoch, Root} from "@chainsafe/lodestar-types";
+import {toHexString} from "@chainsafe/ssz";
+import {ZERO_HASH} from "@chainsafe/lodestar-beacon-state-transition";
+
+export function isEqualRoot(root1: Root, root2: Root): boolean {
+  return toHexString(root1) === toHexString(root2);
```
mpetrunic, line 6: Think it's better/faster to use config.types.Root.equals as it doesn't convert to string first and then compare strings.

mpetrunic (changes requested):

packages/lodestar-validator/src/slashingProtection/utils.ts
dapplion: Good point! Could you link the implementation of `equals` for Root type?

mpetrunic: not 100% but https://github.com/ChainSafe/ssz/blob/master/src/backings/byteArray/abstract.ts#L20

dapplion: Suggestion applied in https://github.com/ChainSafe/lodestar/pull/1674/commits/1c10190d60378a812ea4a3be81873e054c46493c

dapplion:
Superseded by https://github.com/ChainSafe/lodestar/pull/1674

PR #1661, opened 2020-10-13
Add interchange format parser and serializer

dapplion:
Adds interchange format parser and serializer according to the "Slashing Protection Database Interchange Format"

From https://hackmd.io/@sproul/Bk0Y0qdGD implements:
- Complete v4 format

### DIFF
--- a/packages/lodestar-validator/src/slashingProtection/interchange/errors.ts
+++ b/packages/lodestar-validator/src/slashingProtection/interchange/errors.ts
@@ -1,0 +1,19 @@
+import {Root} from "@chainsafe/lodestar-types";
+import {LodestarError} from "@chainsafe/lodestar-utils";
+
+export enum InterchangeErrorErrorCode {
+  UNSUPPORTED_FORMAT = "ERR_INTERCHANGE_UNSUPPORTED_FORMAT",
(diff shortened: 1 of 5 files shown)


mpetrunic (commented):
I think you would reduce a lot of code if you write ssz definition for interchange format and use sszDefinition.fromJson/toJson as it provides all conversions out of box?

dapplion:
> I think you would reduce a lot of code if you write ssz definition for interchange format and use sszDefinition.fromJson/toJson as it provides all conversions out of box?

That's a good idea! In that case how would you handle the conversion from camel case to snake case and vice-versa if any?

mpetrunic:
> That's a good idea! In that case how would you handle the conversion from camel case to snake case and vice-versa if any?
```
sszType.toJson({case: "snake"})
sszType.fromJson({case: "snake"})
```
is already supported

dapplion:
> > That's a good idea! In that case how would you handle the conversion from camel case to snake case and vice-versa if any?
> 
> ```
> sszType.toJson({case: "snake"})
> sszType.fromJson({case: "snake"})
> ```
> 
> is already supported

SSZ de-serialization from JSON does not support optional fields, which the interchange format has. Therefore I would stick to custom parsing which is not that lengthy in this case

dapplion:
Superseded by https://github.com/ChainSafe/lodestar/pull/1677

PR #1664, opened 2020-10-14
Regular sync to start from last processed block instead of forkchoice head

### LINKED ISSUE #1657: Forkchoice head does not seem right after initial sync
**Describe the bug**
After initial sync, we processed up to slot 495550. Right after that we start regular sync and head slot is at 495327.

```
2020-10-12 16:13:33 [SYNC]             ^[[32minfo^[[39m: Imported blocks for slots: 495520,495521,495524,495525,      495526,495527,495528,495529,495530,495531,495533,495534,495535,495539,495540,495543,495544,495547,495548,495549,      495550
  368 2020-10-12 16:13:35 [CHAIN]         ^[[36mverbose^[[39m: Checkpoint processed epoch=15485, root=0x6ab0c7a7159257      cd51e87b483af63996f5fd64c69a528e0875181566ea2d7e17
  369 2020-10-12 16:13:35 [SYNC]             ^[[32minfo^[[39m: Sync progress currentEpoch=15485, targetEpoch=15485, sp      eed=2.0 slots/s, estimatedTillComplete=0 hours
  370 2020-10-12 16:13:35 [SYNC]             ^[[32minfo^[[39m: Reach common finalized checkpoint at epoch 15485
  371 2020-10-12 16:13:35 [SYNC]             ^[[32minfo^[[39m: initial sync stop
  372 2020-10-12 16:13:35 [CHAIN]           ^[[34mdebug^[[39m: Block processed slot=495520, root=0x6ab0c7a7159257cd51e      87b483af63996f5fd64c69a528e0875181566ea2d7e17
  373 2020-10-12 16:13:36 [SYNC]             ^[[32minfo^[[39m: initial sync stop
  374 2020-10-12 16:13:36 [CHAIN]           ^[[34mdebug^[[39m: Block processed slot=495521, root=0x6e7a8d9527110d78952      fe60de5cb20b691556fb05d0dfdd94c24e6d9c8823bd0
  375 2020-10-12 16:13:36 [SYNC]             ^[[32minfo^[[39m: Started regular syncing currentSlot=495667, headSlot=49      5327
```

The result of that is Regular Sync tried to fetch an old range `495328-495347` which we had blocks already.

```
2020-10-12 16:13:48 [SYNC]             ^[[32minfo^[[39m: Regular Sync: Found best peer 16Uiu2HAmFYMVzKz9wC2JfiMP      ZodA322v5iV9HAwnETnwHM5m53Gy peerHeadSlot=495662, currentSlot=495668
  800 2020-10-12 16:13:48 [SYNC]             ^[[32minfo^[[39m: Regular Sync: Setting target newTargetSlot=495347
  801 2020-10-12 16:13:48 [SYNC]             ^[[32minfo^[[39m: Regular Sync: Requesting blocks from slot 495328 to slo      t 495347
```

Although I can still catch up with head slot using `unknownBlockRoot` event, this is not nice.

**Expected behavior**
Head slot should normally around 495550 at the time of starting regular sync.

twoeths:
resolves #1657
+ At the time of regular sync, forkchoice head() is not updated until the next finalized epoch so we should not use it, use the `lastProcessedBlock` in initial sync instead
+ Change `blockPerChunk` default value from 20 to 64 to be more efficient, similar to initial sync.
+ Fix initial sync bypass condition: finalized epoch maybe different from epoch of finalized block. This happens in case start slot of finalized epoch is a skipped slot.

### DIFF
--- a/packages/lodestar/src/sync/initial/fast/index.ts
+++ b/packages/lodestar/src/sync/initial/fast/index.ts
@@ -11,7 +11,7 @@ import {IInitialSyncModules, InitialSync, InitialSyncEventEmitter} from "../inte
 import {EventEmitter} from "events";
 import {Checkpoint, SignedBeaconBlock, Slot, Status} from "@chainsafe/lodestar-types";
 import pushable, {Pushable} from "it-pushable";
-import {computeEpochAtSlot, computeStartSlotAtEpoch} from "@chainsafe/lodestar-beacon-state-transition";
+import {computeStartSlotAtEpoch} from "@chainsafe/lodestar-beacon-state-transition";
 import pipe from "it-pipe";
 import {ISlotRange, ISyncCheckpoint} from "../../interface";
 import {fetchBlockChunks, getCommonFinalizedCheckpoint, processSyncBlocks} from "../../utils";
@@ -73,7 +73,7 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
     if (
       !this.targetCheckpoint ||
       this.targetCheckpoint.epoch == GENESIS_EPOCH ||
-      this.targetCheckpoint.epoch <= computeEpochAtSlot(this.config, this.blockImportTarget)
+      this.targetCheckpoint.epoch <= this.chain.forkChoice.getFinalizedCheckpoint().epoch
     ) {
       this.logger.info("No peers with higher finalized epoch");
       await this.stop();
--- a/packages/lodestar/src/sync/initial/fast/index.ts
+++ b/packages/lodestar/src/sync/initial/fast/index.ts
@@ -93,6 +93,13 @@ export class FastSync extends (EventEmitter as {new (): InitialSyncEventEmitter}
     this.chain.emitter.removeListener("checkpoint", this.checkSyncCompleted);
   }
 
+  /**
+   * Make sure we get up-to-date lastProcessedBlock from sync().
+   */
+  public getLastProcessedBlock(): ISyncCheckpoint {
+    return this.lastProcessedBlock;
+  }
+
   public getHighestBlock(): Slot {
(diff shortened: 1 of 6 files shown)


mpetrunic:
> At the time of regular sync, forkchoice head() is not updated until the next finalized epoch so we should not use it, use the lastProcessedBlock in initial sync instead

How is this possible? Ain't head suppose to update depending on attestations and new blocks?

twoeths:
> we start the node with last known state as the finalized checkpoint in forkchoice

Is this because of weak subjectivity start? Why aren't we setting finalizedcheckpoint based on last known state finalized checkpoint?
Seems odd to have forkchoice broken during regular sync though because we might not get new finalized checkpoint. Does that mean we won't have head after regular sync until we reach finalization?

twoeths:
> > we start the node with last known state as the finalized checkpoint in forkchoice
> 
> Is this because of weak subjectivity start? Why aren't we setting finalizedcheckpoint based on last known state finalized checkpoint?
> Seems odd to have forkchoice broken during regular sync though because we might not get new finalized checkpoint. Does that mean we won't have head after regular sync until we reach finalization?

you're all right. We use last known state as finalized checkpoint because we want to support to start from weak subjectivity state with same API `initForkchoice`. I think we have 2 options to get through the issue:
1. [Current approach of this PR] Still use last known state as finalized checkpoint. The down side of this is we don't have new forkchoice head for a while as explained
2. Or create new API `initForkchoiceWeakSubjectivity`.
+ `initForkChoice` would use last known state's justified/finalized checkpoints to initialize checkpoints. We may have to initialize forkchoice with corresponding blocks from `archivedBlocks` repository.
+ `initForkchoiceWeakSubjectivity` would be the same to the current `initForkchoice`, and it's for initializing the node with a weak subjectivity state without having to query any blocks.

cc @wemeetagain to see which options we want to go.

PR #1665, opened 2020-10-14
Run lodestar e2e with validators over rest api

mpetrunic:


mpetrunic:
- changed max validators per duty request to 64
- changed external error json serialization to only include error messages since some errors have huge metadata and even circular references.

twoeths (dismissed):

packages/lodestar-utils/src/json.ts
```diff
@@ -7,27 +7,31 @@ export function errorToObject(obj: Error): Json {
...
-      if (arg instanceof Uint8Array) return toHexString(arg);
-      if (arg instanceof LodestarError) return toJson(arg.toObject());
-      if (arg instanceof Error) return toJson(errorToObject(arg));
-      return mapValues(arg, (value) => toJson(value)) as Json;
+      case "object":
+        if (arg === null) return "null";
+        if (Array.isArray(arg)) return arg.map(toJson);
+        if (arg instanceof Uint8Array) return toHexString(arg);
+        if (arg instanceof LodestarError) return toJson(arg.toObject());
+        if (arg instanceof Error) return arg.message;
```
dapplion: We should include both the message and stack. Fixes for this issue are included in this PR https://github.com/ChainSafe/lodestar/pull/1669 too

wemeetagain:
the unit test for `toJson` util function is failing

packages/lodestar-utils/src/json.ts
```diff
@@ -2,26 +2,29 @@ import {Json, toHexString} from "@chainsafe/ssz";
...
-    case "function":
-      return arg.toString();
-
-    case "object":
-      if (arg === null) return "null";
-      if (arg instanceof Uint8Array) return toHexString(arg);
-      if (arg instanceof LodestarError) return toJson(arg.toObject());
-      if (arg instanceof Error) return toJson(errorToObject(arg));
-      return arg as Json;
+  try {
```
dapplion: The `toJson` helper was fixed in https://github.com/ChainSafe/lodestar/pull/1669. You can rebase to master and use its solution if you believe its sufficient

3xtr4t3rr3str14l (approved):
LGTM

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -109,7 +109,7 @@ export class AttestationService {
     try {
       attesterDuties = await this.provider.validator.getAttesterDuties(epoch, this.publicKeys);
     } catch (e) {
-      this.logger.error(`Failed to obtain attester duty for epoch ${epoch}`, e);
+      this.logger.error(`Failed to obtain attester duty for epoch ${epoch}`, e.message);
```
dapplion, line 112: Now that the logger issues are fixed, should we send the entire error object to the logger to print the stack trace?

mpetrunic: TBH, those stacktraces can be too much on validator side. Maybe if they are only enabled on debug level. But just useless clutter here.

PR #1666, opened 2020-10-15
Avoid keeping our node busy processing blocks

### LINKED ISSUE #1619: Regular Sync: not enough connected peers
**Describe the bug**
This happens with my libp2p branch. I notice during initial sync, I can see ~20 connected peers but at regular sync, it reduced gradually (mostly with 129 goodbye code) and in worse case, we have no sync peers.

**Expected behavior**
Should maintain ~20 connected peers like in initial sync

twoeths:
resolves #1619

### DIFF
--- a/packages/lodestar/src/chain/blocks/stateTransition.ts
+++ b/packages/lodestar/src/chain/blocks/stateTransition.ts
@@ -13,6 +13,7 @@ import {IBlockSummary, IForkChoice} from "@chainsafe/lodestar-fork-choice";
 import {ITreeStateContext} from "../../db/api/beacon/stateContextCache";
 import {ChainEventEmitter} from "../emitter";
 import {IBlockJob} from "../interface";
+import {sleep} from "@chainsafe/lodestar-utils";
 
 /**
  * Emits a properly formed "checkpoint" event, given a checkpoint state context
@@ -70,6 +71,8 @@ export async function processSlotsToNearestCheckpoint(
     processSlots(postCtx.epochCtx, postCtx.state, nextEpochSlot);
     emitCheckpointEvent(emitter, postCtx);
     postCtx = cloneStateCtx(postCtx);
+    // this avoids keeping our node busy processing blocks
+    await sleep(0);
   }
   return postCtx;
 }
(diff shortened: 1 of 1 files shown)


packages/lodestar/src/util/queue.ts
```diff
@@ -37,6 +38,9 @@ export class JobQueue {
     this.processor = pipe(this.queue, async (source) => {
       for await (const job of source) {
         await this.processJob(job);
+        // in case we have adjacent events in the event loop of the same job like block processing
+        // we want other event to have a chance to execute too
+        await sleep(0);
```
twoeths: right now we use JobQueue for block processing and for regen module. I think both of them need this trick.

dapplion: We should benchmark the performance implications of this. Maybe it's not necessary to do this on every block but every N blocks

twoeths: I don't think it has a big performance impact as this basically means we put the execution of this event to the end of the event loop. Jobs putting to this JobQueue is meant to be heavy already, others should be lightweight or io operations and it'll be executed soon.

But I also have some statistics in my local environment for the current performance of block processing: processing a normal block in regular sync takes 1.5s-2s in average, processing an epoch takes 4s-5s.

twoeths:
with this I saw we can process other network req/resp in the middle of block range processing, did not see we lose peers like previously although we need to improve quality of peers (see #1668):

```
2020-10-15 15:03:15 [SYNC]             info: Peer status activePeers=21, syncPeers=2, unknownRootPeers=2
2020-10-15 15:03:51 [SYNC]             info: Peer status activePeers=24, syncPeers=2, unknownRootPeers=2
```

mpetrunic (dismissed):

wemeetagain (changes requested):
I think this is a good idea, though I think we should put this inside the functions in `lodestar/src/chain/blocks/stateTransition.ts` which are used in both block processing and regeneration.
This way we can avoid the issue in both cases, not just in new block processing.

Eg: add the sleep to `processSlotsToNearestCheckpoint` (so we sleep after every epoch processed)
and add the sleep to `runStateTransition` (so we sleep after every block processed)

mpetrunic (approved):

wemeetagain (approved):

PR #1669, opened 2020-10-15
Deal with circular references

dapplion:
In Zinken launch, a circular reference prevented the logger from showing an important error. 

```
(node:19660) UnhandledPromiseRejectionWarning: RangeError: Maximum call stack size exceeded
    at Object (<anonymous>)
    at baseGetTag (/home/ubuntu/lodestar/packages/lodestar-utils/node_modules/lodash/lodash.js:3067:51)
    at isFunction (/home/ubuntu/lodestar/packages/lodestar-utils/node_modules/lodash/lodash.js:11679:17)
    at isArrayLike (/home/ubuntu/lodestar/packages/lodestar-utils/node_modules/lodash/lodash.js:11359:58)
    at keys (/home/ubuntu/lodestar/packages/lodestar-utils/node_modules/lodash/lodash.js:13333:14)
    at /home/ubuntu/lodestar/packages/lodestar-utils/node_modules/lodash/lodash.js:4920:21
    at baseForOwn (/home/ubuntu/lodestar/packages/lodestar-utils/node_modules/lodash/lodash.js:2990:24)
    at mapValues (/home/ubuntu/lodestar/packages/lodestar-utils/node_modules/lodash/lodash.js:13426:7)
    at toJson (/home/ubuntu/lodestar/packages/lodestar-utils/src/json.ts:22:14)
    at /home/ubuntu/lodestar/packages/lodestar-utils/src/json.ts:22:40
```

This PR allows the toJson and toString methods to deal with circular references and not throw errors. 

Then, in case of circular references the logger will still print part of the original intended message. However, if a circular reference occurs, it typically will come from un-intended metdata so it's okay to omit it.

### DIFF
--- a/packages/lodestar-utils/src/json.ts
+++ b/packages/lodestar-utils/src/json.ts
@@ -1,5 +1,5 @@
 import {Json, toHexString} from "@chainsafe/ssz";
-import {mapValues, pick} from "lodash";
+import {pick} from "lodash";
 import {LodestarError} from "./errors";
 
 export function errorToObject(obj: Error): Json {
@@ -15,11 +15,10 @@ export function toJson(arg: unknown): Json {
 
     case "object":
       if (arg === null) return "null";
-      if (Array.isArray(arg)) return arg.map(toJson);
       if (arg instanceof Uint8Array) return toHexString(arg);
       if (arg instanceof LodestarError) return toJson(arg.toObject());
       if (arg instanceof Error) return toJson(errorToObject(arg));
-      return mapValues(arg, (value) => toJson(value)) as Json;
+      return arg as Json;
 
     // Already valid JSON
     case "number":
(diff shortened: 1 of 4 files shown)


packages/lodestar-utils/src/json.ts
```diff
@@ -15,11 +15,10 @@ export function toJson(arg: unknown): Json {
 
     case "object":
       if (arg === null) return "null";
-      if (Array.isArray(arg)) return arg.map(toJson);
       if (arg instanceof Uint8Array) return toHexString(arg);
       if (arg instanceof LodestarError) return toJson(arg.toObject());
       if (arg instanceof Error) return toJson(errorToObject(arg));
```
mpetrunic, line 15: I think only circular references are here and this is totally wrong since external errors have huge metadata. 600+lines so in #1665 I've changed this to just print error.message.

dapplion: Oh damn, then I'll pick only message and stack `errorToObject`. Would that resolve your point?

mpetrunic: Probably, yeah

mpetrunic (approved):

wemeetagain (approved):

Issue #1667, opened 2020-10-15
Network req/resp timeout without error


twoeths:
**Describe the bug**
This is not the `dialProtocol` timeout issue in #1597 since we got the stream (I applied my hot fix #1600 for my local branch). After that we receive no response from the peer and regular sync stalled.

```
2020-10-15 14:13:49 [SYNC]             info: Regular Sync: Requesting blocks from slot 501677 to slot 501740
2020-10-15 14:13:49 [SYNC]             info: Regular Sync: Synced up to slot 501676  currentSlot=516668, gossipParentBlockRoot=undefined
2020-10-15 14:13:50 [NETWORK]       ^[[36mverbose^[[39m: sending beacon_blocks_by_range request to 16Uiu2HAmBHHixkAU7ikai94zvxWgbJS4J9BBNr2ZsHkAt4ghWCBd requestId=f69141a4794db63f, encoding=ssz_snappy
2020-10-15 14:13:59 [NETWORK]       ^[[36mverbose^[[39m: got stream to 16Uiu2HAmBHHixkAU7ikai94zvxWgbJS4J9BBNr2ZsHkAt4ghWCBd requestId=f69141a4794db63f, encoding=ssz_snappy
```

**Expected behavior**
We should either receive/trigger an error or the log `receive beacon_blocks_by_range request from` with same requestId.

I still saw the peer stay connected to us after that:
```
2020-10-15 14:14:17 [NETWORK]       ^[[36mverbose^[[39m: receive ping request from 16Uiu2HAmBHHixkAU7ikai94zvxWgbJS4J9BBNr2ZsHkAt4ghWCBd requestId=d7c30682f8f20c01, encoding=ssz_snapp
```
and after that we did do beacon_blocks_by_root request to the same peer successfully too:
```
2020-10-15 14:14:24 [NETWORK]       ^[[36mverbose^[[39m: sending beacon_blocks_by_root request to 16Uiu2HAmBHHixkAU7ikai94zvxWgbJS4J9BBNr2ZsHkAt4ghWCBd requestId=8a24e9262bff231f, encoding=ssz_snappy
2020-10-15 14:14:25 [NETWORK]       ^[[36mverbose^[[39m: got stream to 16Uiu2HAmBHHixkAU7ikai94zvxWgbJS4J9BBNr2ZsHkAt4ghWCBd requestId=8a24e9262bff231f, encoding=ssz_snappy
[...]

PR #1670, opened 2020-10-15
Store pending blocks to db

### LINKED ISSUE #1663: Optimize block pool memory usage
We got OOM issue again in non-finality time in Medalla, did not do a memory inspection but I think in-memory block pool is not sustainable and we need to improve that first.
In my test, I got ~1000 gossip/pending blocks and it need to be ~8000 to catchup with head block/slot.

I think "BlockPool" should maintain maps of block roots/slots, the real block data is stored in a new "pendingBlock" db (this is similar to forkchoice which is an index store vs "block" db which is real repo).

twoeths:
resolves #1663 

in this infinality time of medalla, it's not sustainable to store pending blocks in memory. So I created `pendingBlock` repository for this purpose and `BlockPool` acts as an index for it, similar to what forkchoice does (vs `block` repository)

### DIFF
--- a/packages/lodestar/src/chain/blocks/pool.ts
+++ b/packages/lodestar/src/chain/blocks/pool.ts
@@ -99,6 +99,10 @@ export class BlockPool {
     return this.blocks.size;
   }
 
+  public has(blockRoot: Root): boolean {
+    return Boolean(this.blocks.get(this.getBlockKeyByRoot(blockRoot)));
+  }
+
   public getByParent(parentRoot: Root): Uint8Array[] {
     const hexArr = Array.from(this.blocksByParent.get(toHexString(parentRoot))?.values() ?? []);
     return hexArr.map((hex) => fromHexString(hex));
@@ -119,5 +123,9 @@ export class BlockPool {
 
   private getBlockKey(block: SignedBeaconBlock): string {
     return toHexString(this.config.types.BeaconBlock.hashTreeRoot(block.message));
+  }
+
+  private getBlockKeyByRoot(root: Root): string {
+    return toHexString(root);
   }
 }
--- a/packages/lodestar/src/chain/blocks/pool.ts
+++ b/packages/lodestar/src/chain/blocks/pool.ts
@@ -95,6 +95,10 @@ export class BlockPool {
     return fromHexString(root);
   }
 
+  public getTotalPendingBlocks(): number {
+    return this.blocks.size;
+  }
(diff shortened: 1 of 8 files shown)


mpetrunic:
If it works ok, I guess it's fine. But I think it would make more sense to prune block pool into db rather than storing everything.
Not sure what would be pruning and restoration rules though

twoeths:
I tend to think of it as an index store, similar to what we have with forkchoice. We don't store any data other than index in forkchoice.
The first sync with it works fine until block 501676, then I got issue #1667, need to test it again.

twoeths:
Ready for review.

packages/lodestar/src/chain/blocks/pool.ts
```diff
@@ -91,20 +87,26 @@ export class BlockPool {
...
-    return this.blocks.get(toHexString(blockRoot));
+  public getMissingAncestor(blockRoot: Root): Root {
+    let root = toHexString(blockRoot);
+    while (this.blocks.has(root)) {
+      root = this.blocks.get(root)!;
+    }
+    return fromHexString(root);
   }
 
-  public has(blockRoot: Root): boolean {
```
wemeetagain, line 98: Can we keep this method :)

wemeetagain (commented):
LGTM

Two thoughts (not really actionable):
- we're storing blockRoot => parentRoot as the primary index in `BlockPool`. For now, likely enough, I wonder if we will want/need to store something closer to a `BeaconBlockHeader` in the future. (Eg: to aid in checking if we've already gotten a block at a slot from a proposer) For now, this seems good enough tho.
- `BlockPool` now operates on `SignedBeaconBlock` instead of `IBlockJob`.  The reason we were storing "job" objects here is so we can track which validations have already been done, so jobs can avoid duplicating validations when pushed to the block processor. But now, looking at the flow of validations, it seems that we never actually needed the job metadata here!
Blocks can only go into this block pool when pending bc of "having a future slot" or "having an unknown parent". And these validations (in gossip and in block processing) are performed before any others. So there is no chance for us to have validated the block signature and then put the block in the block pool. These blocks always are pre-validated.

packages/lodestar/src/chain/blocks/pool.ts
twoeths: yeah I got it back

wemeetagain (approved):

PR #1671, opened 2020-10-15
Standard rest pool endpoints

### LINKED ISSUE #1058: Add pool rest endpoints
   - add beacon/pool endpoints

mpetrunic:
- some endpoints that are used by validator are duplicated as I plan to migrate validator to standard endpoint in separate PR to reduce diff

resolves #1058


wemeetagain (approved):
Solid!

PR #1673, opened 2020-10-16
Response sent feedback

### LINKED ISSUE #1465: Refactor RPC sending to await completion of response
We want to track when a response gets fully sent.
Currently sendResponse is an artificially synchronous function.
It returns quickly, but contains a generator that emits bytes.
Needs to be wrap generator in a Promise.

mpetrunic:
Instead of emitting response events and waiting for them, we are passing destination stream for response.

TODO:
- add libp2p fix so e2e tests can pass
- use latest any-signal with types

resolves #1465


wemeetagain (commented):
holy cow, this is looking good :+1: :+1:

mpetrunic:
Seems like despite libp2p fixes stuck dialProtocol problem still exists. I've added dial protocol to our request timeout to unblock this PR.

wemeetagain (approved):

PR #1677, opened 2020-10-16
Add slashing protection and interchange format

dapplion:
Refactors current slashing protection in the validator client to accommodate the interchange format as defined in https://hackmd.io/@sproul/Bk0Y0qdGD

The slashing protection refactor improves the current implementation by covering all possible surround vote cases. It does so with the min-max-surround algorithm https://github.com/protolambda/eth2-surround#min-max-surround in O(1) time. In total, this slashing protection module uses 5 repositories:
- slashingProtectionBlockBySlot: To check block double vote in O(1), and fetch the first block (stream values with `limit=1`) to assert interchange format condition `block.slot > minBlock.slot`.
- slashingProtectionAttestationByTarget: To check attestation double vote in O(1)
- slashingProtectionMinSpanDistance: To check surrounding votes in O(1)
- slashingProtectionMaxSpanDistance: To check surrounded votes in O(1)
- slashingProtectionAttestationLowerBound: To assert the interchange format conditions `att.sourceEpoch >= minSourceEpoch` and `att.targetEpoch > minTargetEpoch`. Computed once on interchange import.

The core logic for min-max-surround is tested with the same data used in Prysm's implementation. The interchange format is tested with a common set of spec-ed tests from https://github.com/eth2-clients/slashing-protection-interchange-tests 

#### Next steps
- Develop CLI commands to call import / export interchange functions


packages/lodestar-db/src/schema.ts
```diff
@@ -32,9 +32,15 @@ export enum Bucket {
...
-  proposedAttestations = 18,
+  // validator = 16, // DEPRECATED on v0.11.0
+  // lastProposedBlock = 17, // DEPRECATED on v0.11.0
+  // proposedAttestations = 18, // DEPRECATED on v0.11.0
+  // validator slashing protection
+  slashingProtectionBlockBySlot = 20,
+  slashingProtectionAttestationByTarget = 21,
+  slashingProtectionAttestationLowerBound = 22,
+  slashingProtectionMinSpanDistance = 23,
+  slashingProtectionMaxSpanDistance = 24,
```
mpetrunic, line 43: nitpick and probably for separate PR: we should probably split beacon node db from validator db. It made sense to be together earlier because of how dependencies were arranged

packages/lodestar-validator/src/slashingProtection/attestation/dbRepository.ts
```diff
@@ -0,0 +1,71 @@
...
+      }))
+    );
+  }
+
+  private encodeKey(pubkey: BLSPubkey, targetEpoch: Epoch): Buffer {
+    return encodeKey(this.bucket, Buffer.concat([Buffer.from(pubkey), intToBytes(BigInt(targetEpoch), 8, "be")]));
+  }
+}
+
+export class AttestationLowerBoundRepository {
```
mpetrunic, line 49: can we keep each class in separate file?

I don't think it's just me since I noticed it on other colleagues, when you open a file if first class name you see isn't what you looking for you will go to next file

packages/lodestar-validator/src/slashingProtection/attestation/service.ts
```diff
@@ -0,0 +1,145 @@
...
+  async insertAttestation(pubKey: BLSPubkey, att: SlashingProtectionAttestation): Promise<void> {
+    await this.attestationByTarget.set(pubKey, [att]);
+    await this.minMaxSurround.insertAttestation(pubKey, {source: att.sourceEpoch, target: att.targetEpoch});
+  }
+
+  /**
+   * Interchange import / export functionality
+   */
+  async importAttestations(pubkey: BLSPubkey, atts: SlashingProtectionAttestation[]): Promise<void> {
+    await this.attestationByTarget.set(pubkey, atts);
```
mpetrunic, line 124: ain't this gonna be set in insertAttestation?

wemeetagain (approved):
Very strong work :1st_place_medal: 
Thanks for adding extensive documentation and tests.
IMO this is very readable and will be very easy to maintain.

mpetrunic (changes requested):
Can you add exports for slashing protection stuff from lodestar-validator. Like everything I need as user to have slashing protection where I only need to provide database storage layer.

mpetrunic:
:sob:

packages/lodestar-validator/src/slashingProtection/attestation/dbRepository.ts
dapplion: Yes I agree, I'll split it in a new PR

packages/lodestar-db/src/schema.ts
dapplion: It's indeed confusing. The advantage of using the same schema is that if you re-use db instances there won't be conflicts.

packages/lodestar-validator/src/slashingProtection/attestation/service.ts
dapplion: Yes, but this `importAttestations` a specialized method to import a (large) number of attestations from an interchange file. Could you develop your question more?

dapplion:
> Can you add exports for slashing protection stuff from lodestar-validator. Like everything I need as user to have slashing protection where I only need to provide database storage layer.

Are you referring to a CLI functionality? I don't fully understand your question

packages/lodestar-validator/src/slashingProtection/attestation/service.ts
mpetrunic: Seems like you don't need this line as you are gonna overwrite those values with individual insertAttestation in code below.

mpetrunic:
> > Can you add exports for slashing protection stuff from lodestar-validator. Like everything I need as user to have slashing protection where I only need to provide database storage layer.
> 
> Are you referring to a CLI functionality? I don't fully understand your question

Lodestar-validator is used as a library outside of lodestar context with different database implemention. Previously, we just sent database implementing required interface. So it would be convenient for this library to export slashing protection stuff and interfaces from root index.ts

packages/lodestar-validator/src/slashingProtection/attestation/dbRepository.ts
dapplion: New PR https://github.com/ChainSafe/lodestar/pull/1679

dapplion:
> > > Can you add exports for slashing protection stuff from lodestar-validator. Like everything I need as user to have slashing protection where I only need to provide database storage layer.
> > 
> > 
> > Are you referring to a CLI functionality? I don't fully understand your question
> 
> Lodestar-validator is used as a library outside of lodestar context with different database implemention. Previously, we just sent database implementing required interface. So it would be convenient for this library to export slashing protection stuff and interfaces from root index.ts

Should this be good enough?
https://github.com/ChainSafe/lodestar/blob/00012384f2fefe57805e368d815c6677214baf05/packages/lodestar-validator/src/index.ts#L9

What items from slashing protection are not exported that should be?

packages/lodestar-validator/src/slashingProtection/attestation/service.ts
dapplion: The line below calls `this.minMaxSurround.insertAttestation` which inserts to a different repository than `attestationByTarget`. Let me know if the naming is not clear

dapplion:
> Previously, we just sent database implementing required interface

I thought it was simpler to get as a parameter a generic slashing protection instance that abstracts further than the db. @mpetrunic let me know if you think that we should revert back to passing a more complex db instance object as parameter

mpetrunic:
> > Previously, we just sent database implementing required interface
> 
> I thought it was simpler to get as a parameter a generic slashing protection instance that abstracts further than the db. @mpetrunic let me know if you think that we should revert back to passing a more complex db instance object as parameter

It's fine in current context, might be weird in future if validator might have to store more stuff than just slashing protection. But I guess we can keep this for now.

PR #1678, opened 2020-10-16
Add maxEpoch timeout in e2e justification tests

dapplion:
I think it might be a good idea to limit the how many epochs the justification tests can run, on top of limiting by time.

I've assigned `maxEpoch = 10` just as a placeholder random value, but we should discuss what is it appropriate value for justification and finalization event tests.

### DIFF
--- a/packages/lodestar/test/e2e/multiNodeMultiThread.test.ts
+++ b/packages/lodestar/test/e2e/multiNodeMultiThread.test.ts
@@ -3,10 +3,10 @@ import os from "os";
 import {Worker} from "worker_threads";
 import {IBeaconParams} from "@chainsafe/lodestar-params";
 import {Checkpoint} from "@chainsafe/lodestar-types";
-import {waitForEvent} from "../utils/events/resolver";
 
 describe("Run multi node multi thread interop validators (no eth1) until checkpoint", function () {
   const checkpointEvent = "justified";
+  const maxEpoch = 10;
   const validatorsPerNode = 8;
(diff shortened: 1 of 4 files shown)


mpetrunic (commented):
Feels a bit useless since we already have max time. Why do you think it's necessary?

dapplion:
> Feels a bit useless since we already have max time. Why do you think it's necessary?

The total timeout in mocha would have to consider the start-up time and other factors. This way is more precise to kill the test at a specified epoch. However, you are right we could use the timeout for this. If so, we should compute a more thoughtful value than just "10 min". So the question still stands
- how fast do we expect these tests to reach (a) justification, (b) finality?

mpetrunic:
But aren't we already have timeout in waitForEvent method:
https://github.com/ChainSafe/lodestar/pull/1678/files#diff-59dcf7b40e1c8fd56f18dccccb69ddfec995e934d15333e7bd6a92b4ecc13ca3R90

We just aren't using it in multi node/thread test

dapplion:
> But aren't we already have timeout in waitForEvent method:

@mpetrunic So do you believe a timeout of 10 min is appropriate? Feels kinda random. I would be okay with a timebased timeout if comes from a rationalized number: `maxEpoch x secondsPerEpoch`

mpetrunic:
> > But aren't we already have timeout in waitForEvent method:
> 
> @mpetrunic So do you believe a timeout of 10 min is appropriate? Feels kinda random. I would be okay with a timebased timeout if comes from a rationalized number: `maxEpoch x secondsPerEpoch`

Don't really care, we can put some other number if you think it's more appropriate :)

PR #1682, opened 2020-10-17
Add internal event emitter

wemeetagain:
This is a solution to the following problem:
External (non-chain) handlers of `ChainEventEmitter` events can't be guaranteed that the proper side effects (internal to the chain) have been performed, at the time that the external handler is triggered.

This problem has been hinted at in #1655, where a new event is introduced, rather than relying on the existing events.
It's done for the following reasons:
1. the BlockPool is currently not part of the public IBeaconChain interface, so the sync can't retrieve the 'missing ancestor'
2. the sync can't really rely on the BlockPool being properly updated when the "error:block" event is fired/handled

Number 1. can be rectified fairly easily, either with a public method or public member on IBeaconChain
its 2. that this PR deals with.

The problem stems from the fact that multiple event handlers can be run concurrently (in the case that handlers are async / have async steps).
This means that external chain event handlers can be interwoven with internal chain event handlers in a way that exposes external handlers to an inconsistent chain state.

There are several ways to handle the problem, this is an attempt to systematically resolve it.

This PR adds an "internal event emitter" that is used to wire together chain submodules.
The chain's event handlers subscribe to this internal emitter.
These handlers perform various important side effects (eg: updating caches/db) that we want completed before external handers can run.
Once the chain's event handler finishes in its entirety (async function completes), the same event, with the same args is emitted from the public/external event emitter, where downstream consumers can attach event handlers.

### DIFF
--- a/packages/lodestar/src/chain/eventHandlers.ts
+++ b/packages/lodestar/src/chain/eventHandlers.ts
@@ -1,12 +1,12 @@
 import {readOnlyMap, toHexString} from "@chainsafe/ssz";
 import {Attestation, Checkpoint, SignedBeaconBlock, Slot} from "@chainsafe/lodestar-types";
-import {toJson} from "@chainsafe/lodestar-utils";
+import {ILogger, toJson} from "@chainsafe/lodestar-utils";
 import {IBlockSummary} from "@chainsafe/lodestar-fork-choice";
 
 import {ITreeStateContext} from "../db/api/beacon/stateContextCache";
 import {AttestationError, AttestationErrorCode, BlockError, BlockErrorCode} from "./errors";
 import {IBlockJob} from "./interface";
-import {IChainEvents} from "./emitter";
+import {ChainEventEmitter, IChainEvents} from "./emitter";
 import {BeaconChain} from "./chain";
 
 interface IEventMap<Events, Key extends keyof Events = keyof Events, Value extends Events[Key] = Events[Key]>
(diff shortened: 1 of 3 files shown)


twoeths:
so the internal emitter and emitter throws same event with same data.

in the case of #1655 , `validate.ts` throws `error:block` with missing parent root while sync doesn't want that missing parent root, it wants missing ancestor root.

if we go with this, we have to publish `BlockPool` to chain.ts and sync should consult `BlockPool` itself to get missing ancestor root.

packages/lodestar/src/chain/eventHandlers.ts
```diff
@@ -1,48 +1,86 @@
...
+ * then emits the event on the event emitter using the same args
+ */
+function wrapHandler<
+  Key extends keyof IChainEvents = keyof IChainEvents,
+  Value extends IChainEvents[Key] = IChainEvents[Key]
+>(event: Key, emitter: ChainEventEmitter, logger: ILogger, handler: (...args: Parameters<Value>) => Promise<void>) {
+  return async (...args: Parameters<Value>): Promise<void> => {
+    try {
+      await handler(...args);
+      emitter.emit(event, ...((args as unknown) as ListenerType<Value>));
```
mpetrunic: It's odd that you have to cast it as unknown, Parameters<Value> should be sufficient

packages/lodestar/src/chain/eventHandlers.ts
```diff
@@ -1,48 +1,86 @@
...
+type ListenerType<T> = [T] extends [(...args: infer U) => any] ? U : [T] extends [void] ? [] : [T];
+
+/**
+ * Returns a function that runs an async handler function with input args,
+ * If the handler function successfully completes,
+ * then emits the event on the event emitter using the same args
+ */
+function wrapHandler<
+  Key extends keyof IChainEvents = keyof IChainEvents,
+  Value extends IChainEvents[Key] = IChainEvents[Key]
```
mpetrunic: can we change generics name to Event and Callback for the sake of readability :slightly_smiling_face:

packages/lodestar/src/chain/eventHandlers.ts
```diff
@@ -1,48 +1,86 @@
...
+import {ChainEventEmitter, IChainEvents} from "./emitter";
 import {BeaconChain} from "./chain";
 
 interface IEventMap<Events, Key extends keyof Events = keyof Events, Value extends Events[Key] = Events[Key]>
   extends Map<Key, Value> {
   set<Key extends keyof Events>(key: Key, value: Events[Key]): this;
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type ListenerType<T> = [T] extends [(...args: infer U) => any] ? U : [T] extends [void] ? [] : [T];
```
mpetrunic, line 18:
```suggestion
type ListenerType<T> = [T] extends [(...args: infer U) => void] ? U : [T] extends [void] ? [] : [T];
```
?

mpetrunic (commented):
I wen't trough this bunch of times.

I would propose slightly simpler but more readable option.
Introduce "internal" event prefix. So for example we would have:
- clock:slot -> when chain clock slot handler is executed
- internal:clock:slot -> one thats emitted by BeaconClock

this would allow chain module users to decide whether they want to wait for internal state or they don't care. It would also be more explicit on what we are trying to acomplish,

packages/lodestar/src/chain/eventHandlers.ts
wemeetagain: For some reason (unknown to me), `Parameters<Value>` wasn't working.
Thats why `ListenerType` is copied here (direct copy-paste from the events type declarations).

wemeetagain:
> Introduce "internal" event prefix.

That was my original thought as well. I ended up not doing this for a few reasons.


IMO these internal events shouldn't be available at all to outside consumers. Its not like the internal events are any different semantically, in fact, they're the exact same. The only difference to consumers is that one fires before the chain is in a coherent state, and one fires after the chain is in a coherent state.
Making a distinction between the internal and external events means we need duplicate event types. Again, the events are exactly the same, one-for-one copies of each other.
Using internal events mean that chain modules now need to be aware of the internal/external distinction (by emitting internal events).
While it is more explicit to have internal events, IMO it will be more straightforward and DRY to have a single event interface, just used internally and externally. The fact that the second (internal) event emitter exists, which also fulfills the same interface, should be an implementation detail irrelevant to chain consumers.

Edit: I'd like to pull out the ChainEventEmitter event names into an enum (with the extensive comments from the events hackmd doc). I'm hoping that will also go a long way towards making the events interface clearer to consumers.

3xtr4t3rr3str14l (approved):

twoeths (approved):

PR #1683, opened 2020-10-19
Disconnect useless peers at sync time

twoeths:
resolves #1668 

+ if node is not synced, disconnect half of non-sync peers order by score asc. Sometimes I have 25 peers and non of them are sync peers, I'd think disconnecting all of them at once may be a shock so I just disconnect half of them per time. In the end quality of peers should be good gradually.
+ if node is synced, we care more about missing subnets. Nothing changes.

PR #1685, opened 2020-10-19
OneRangeAhead Regular Sync 

### LINKED ISSUE #1676: Improve Regular Sync
This is at infinality time of Medalla and we have to start from block ~500000 to block ~525000 (at least) and it's not efficient to query a  block range and wait for it to be processed before querying the next one.

We should be able to query 1 range ahead and process the current buffered block range at the same time. This is likely the approach we'll need to apply to initial sync if it's proved to be efficient.

Some code refactoring is expected due to this change.

twoeths:
resolves #1676 

At this time, we need to catch up with ~40000 slots in order to sync to head.

I intended to improve our naive strategy but it was too much change so decided to come up with a new one. Naive was good to start with anyway and we leave it there for backward reference.
+ In term of strategy, the difference is we fetch and process block range at the same time. We do this by fetch one range ahead and buffer it. It's like `Promise.all([fetcher.next(), processor.processUntilComplete())`
+ I separate the fetch to a fetcher - `BlockRangeFetcher` and processor to `BlockRangeProcessor`. The main sync class `ORARegularSync` acts as a controller.
+ It's easier to do unit test for each of them and maintain, I reproduced cases I got in the past in unit tests
+ Performance should be better since we either `fetch` or `process` in naive, now we do both at the same time

I tried syncing a couple of times and felt good about it. Not able to sync up to head (25000 max) bc of different issues:
+ Forkchoice getCanonicalBlockAtSlot()
+ Not enough sync peers
+ Sometimes network req/res still timeout

### DIFF
--- a/packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
+++ b/packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
@@ -49,13 +49,15 @@ export class BlockRangeFetcher implements IBlockRangeFetcher {
       let slotRange: ISlotRange | null = null;
       try {
         const peers = await this.getPeers();
-        // node is stopped
-        if (!peers || !peers.length) return [];
         if (result && !result!.length) await this.handleEmptyRange(peers);
         slotRange = {start: this.rangeStart, end: this.rangeEnd};
         result = await getBlockRange(this.logger, this.network.reqResp, peers, slotRange);
       } catch (e) {
-        this.logger.debug("Failed to get block range " + JSON.stringify(slotRange || {}) + ". Error: " + e.message);
+        this.logger.debug(
+          "Regular Sync: Failed to get block range " + JSON.stringify(slotRange || {}) + ". Error: " + e.message
+        );
+        // node is stopped for whatever reasons
+        if (e.message.trim() === "Aborted") return [];
         result = null;
       }
     }
--- a/packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
+++ b/packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
@@ -8,6 +8,7 @@ import {defaultOptions, IRegularSyncOptions} from "../options";
 import {IBeaconChain} from "../../../chain";
 import {INetwork} from "../../../network";
 import {getBlockRange} from "../../utils";
+import {ISlotRange} from "../../interface";
 
 // TODO: reusable, unit test
 export class BlockRangeFetcher {
@@ -20,22 +21,14 @@ export class BlockRangeFetcher {
   private rangeStart: Slot = 0;
   // exclusive
   private rangeEnd: Slot = 0;
-  private bestPeer: PeerId | undefined;
   private getPeers: () => Promise<PeerId[]>;
 
-  constructor(
-    options: Partial<IRegularSyncOptions>,
-    modules: IRegularSyncModules,
-    rangeStart: Slot,
-    getPeers: () => Promise<PeerId[]>
-  ) {
+  constructor(options: Partial<IRegularSyncOptions>, modules: IRegularSyncModules, getPeers: () => Promise<PeerId[]>) {
     this.config = modules.config;
     this.network = modules.network;
     this.chain = modules.chain;
     this.logger = modules.logger;
     this.opts = deepmerge(defaultOptions, options);
-    this.rangeStart = rangeStart;
-    this.rangeEnd = this.rangeStart + 1;
     this.getPeers = getPeers;
   }
 
@@ -47,13 +40,16 @@ export class BlockRangeFetcher {
     this.rangeEnd = this.getNewTarget();
     let result: SignedBeaconBlock[] | null = null;
     while (!result || !result!.length) {
-      if (result && !result!.length) await this.handleEmptyRange();
-      const slotRange = {start: this.rangeStart, end: this.rangeEnd};
+      let slotRange: ISlotRange | null = null;
       try {
         const peers = await this.getPeers();
+        // node is stopped
+        if (!peers || !peers.length) return [];
+        if (result && !result!.length) await this.handleEmptyRange(peers);
+        slotRange = {start: this.rangeStart, end: this.rangeEnd};
         result = await getBlockRange(this.logger, this.network.reqResp, peers, slotRange);
       } catch (e) {
-        this.logger.debug("Failed to get block range " + JSON.stringify(slotRange) + ". Error: " + e.message);
+        this.logger.debug("Failed to get block range " + JSON.stringify(slotRange || {}) + ". Error: " + e.message);
         result = null;
       }
     }
@@ -61,12 +57,12 @@ export class BlockRangeFetcher {
     return result!;
   }
 
-  private async handleEmptyRange(): Promise<void> {
-    if (!this.bestPeer) {
+  private async handleEmptyRange(peers: PeerId[] = []): Promise<void> {
+    if (!this.getPeers.length) {
       return;
     }
     const range = {start: this.rangeStart, end: this.rangeEnd};
-    const peerHeadSlot = this.network.peerMetadata.getStatus(this.bestPeer)?.headSlot ?? 0;
+    const peerHeadSlot = Math.max(...peers.map((peer) => this.network.peerMetadata.getStatus(peer)?.headSlot ?? 0));
     this.logger.verbose(`Regular Sync: Not found any blocks for range ${JSON.stringify(range)}`);
     if (range.end <= peerHeadSlot) {
       // range contains skipped slots, query for next range
--- a/packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
+++ b/packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
@@ -1,6 +1,6 @@
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {SignedBeaconBlock, Slot} from "@chainsafe/lodestar-types";
-import {ILogger, sleep} from "@chainsafe/lodestar-utils";
+import {ErrorAborted, ILogger, sleep} from "@chainsafe/lodestar-utils";
 import deepmerge from "deepmerge";
 import PeerId from "peer-id";
 import {IRegularSyncModules} from "..";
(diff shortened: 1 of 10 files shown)


packages/lodestar/src/sync/regular/oneRangeAhead/oneRangeAhead.ts
```diff
@@ -0,0 +1,147 @@
...
+  private onGossipBlock = async (block: SignedBeaconBlock): Promise<void> => {
+    const gossipParentBlockRoot = block.message.parentRoot;
+    if (this.chain.forkChoice.hasBlock(gossipParentBlockRoot as Uint8Array)) {
+      this.logger.important("Regular Sync: caught up to gossip block parent " + toHexString(gossipParentBlockRoot));
+      this.emit("syncCompleted");
+      await this.stop();
+    }
+  };
+
+  private async sync(): Promise<void> {
```
twoeths, line 95: the main method to start from

packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
```diff
@@ -0,0 +1,108 @@
...
+import {IBlockRangeFetcher} from "./interface";
+
+export class BlockRangeFetcher implements IBlockRangeFetcher {
+  protected readonly config: IBeaconConfig;
+  private readonly network: INetwork;
+  private readonly chain: IBeaconChain;
+  private readonly logger: ILogger;
+  private readonly opts: IRegularSyncOptions;
+  // for control range across next() calls
+  private lastFetchBlock: ISyncCheckpoint;
```
mpetrunic: lstFetchCheckpoint?

packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
```diff
@@ -0,0 +1,108 @@
...
+      }
+    }
+    // success
+    const lastBlock = result[result.length - 1].message;
+    this.lastFetchBlock = {blockRoot: this.config.types.BeaconBlock.hashTreeRoot(lastBlock), slot: lastBlock.slot};
+    return result!;
+  }
+
+  // always set range based on last fetch block bc sometimes the previous fetch may not return all blocks
+  private getNextRange(): void {
```
mpetrunic:
```suggestion
  private updateNextRange(): void {
```

packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
```diff
@@ -0,0 +1,108 @@
...
+   * Get next block range.
+   */
+  public async next(): Promise<SignedBeaconBlock[]> {
+    this.getNextRange();
+    let result: SignedBeaconBlock[] | null = null;
+    while (!result || !result!.length) {
+      let slotRange: ISlotRange | null = null;
+      try {
+        const peers = await this.getPeers();
+        if (result && !result!.length) await this.handleEmptyRange(peers);
```
mpetrunic:
```suggestion
        if (result && !result.length) await this.handleEmptyRange(peers);
```

packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
```diff
@@ -0,0 +1,108 @@
...
+        slotRange = {start: this.rangeStart, end: this.rangeEnd};
+        result = await getBlockRange(this.logger, this.network.reqResp, peers, slotRange);
+      } catch (e) {
+        this.logger.debug(
+          "Regular Sync: Failed to get block range " + JSON.stringify(slotRange || {}) + ". Error: " + e.message
```
mpetrunic:
```suggestion
          "Regular Sync: Failed to get block range ", {...(slotRange ?? {}), error: e.message}
```

packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
```diff
@@ -0,0 +1,108 @@
...
+        );
+        // node is stopped for whatever reasons
+        if (e.message.trim() === "Aborted") return [];
```
mpetrunic: I feel like this is place where we should use codes. This code would break if we change error message at some point

packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
```diff
@@ -0,0 +1,108 @@
...
+    this.rangeEnd = this.getNewTarget();
+  }
+
+  private async handleEmptyRange(peers: PeerId[] = []): Promise<void> {
+    if (!this.getPeers.length) {
+      return;
+    }
+    const range = {start: this.rangeStart, end: this.rangeEnd};
+    const peerHeadSlot = Math.max(...peers.map((peer) => this.network.peerMetadata.getStatus(peer)?.headSlot ?? 0));
+    this.logger.verbose(`Regular Sync: Not found any blocks for range ${JSON.stringify(range)}`);
```
mpetrunic:
```suggestion
    this.logger.verbose("Regular Sync: Not found any blocks for range", {range});
```

packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
```diff
@@ -0,0 +1,108 @@
...
+    if (range.end <= peerHeadSlot) {
+      // range contains skipped slots, query for next range
+      this.logger.verbose("Regular Sync: queried range is behind peer head, fetch next range", {
+        range: JSON.stringify(range),
```
mpetrunic:
```suggestion
        ...range,
```

packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
twoeths: I should use `if (e instanceof ErrorAborted)`

packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  public setLastProcessedBlock(lastProcessedBlock: ISyncCheckpoint): void {
+    this.lastFetchCheckpoint = lastProcessedBlock;
+  }
+
+  /**
+   * Get next block range.
+   */
+  public async next(): Promise<SignedBeaconBlock[]> {
```
dapplion: Maybe rename to the comment above `getNextBlockRange`

packages/lodestar/src/sync/regular/oneRangeAhead/oneRangeAhead.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  private bestPeer: PeerId | undefined;
+
+  private fetcher: IBlockRangeFetcher;
+
+  private processor: IBlockRangeProcessor;
+
+  private controller!: AbortController;
+
+  private blockBuffer: SignedBeaconBlock[];
```
dapplion, line 32: For consistency with other Lodestar classes let's condense the property declarations without newlines

packages/lodestar/src/sync/regular/oneRangeAhead/processor.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  /**
+   * Main method.
+   */
+  public async processUntilComplete(blocks: SignedBeaconBlock[], signal: AbortSignal): Promise<void> {
+    if (!blocks || !blocks.length) return;
+    await new Promise((resolve) => {
+      const sortedBlocks = sortBlocks(blocks);
+      this.logger.info("Imported blocks for slots: " + blocks.map((block) => block.message.slot).join(","));
```
dapplion: The logger should handle the formatting you, whenever possible give it a message + JSON data

packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
```diff
@@ -0,0 +1,106 @@
...
+    let result: SignedBeaconBlock[] | null = null;
+    while (!result || !result!.length) {
+      let slotRange: ISlotRange | null = null;
+      try {
+        const peers = await this.getPeers();
+        if (result && !result.length) await this.handleEmptyRange(peers);
+        slotRange = {start: this.rangeStart, end: this.rangeEnd};
+        result = await getBlockRange(this.logger, this.network.reqResp, peers, slotRange);
+      } catch (e) {
+        this.logger.debug("Regular Sync: Failed to get block range ", {...(slotRange ?? {}), error: e.message});
```
dapplion, line 67: I think Javascript allows spreading nullish values
```
{...null, ...undefined}
```
Also, give the entire error to the logger, it will handle the formating.
```
this.logger.debug("Regular Sync: Failed to get block range", slotRange, error)
```

packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
```diff
@@ -0,0 +1,106 @@
...
+      this.logger.verbose("Regular Sync: queried range is behind peer head, fetch next range", {
+        ...range,
+        peerHead: peerHeadSlot,
+      });
+      // don't trust empty range as it's rarely happen, peer may return it incorrectly or not up to date
+      // same range start, expand range end
+      this.rangeEnd = this.getNewTarget();
+    } else {
+      this.logger.verbose("Regular Sync: Queried range passed peer head, sleep then try again", {
+        range: JSON.stringify(range),
```
dapplion: `range` is stringified here but not above. Unless there's a strong reason to stringify it, pass the entire JSON to the logger

dapplion (commented):
Great work! Much needed strategy improvement. Added comments mainly around the logger; since we want to eventually allow JSON logging we should not redact logs whenever possible
do
```js
logger.info("Event happen", {id, location, time})
```
instead of
```js
logger.info(`Event ${id} happen in ${location} at ${time}`)
```

twoeths:
So I was able to sync up to head (~120000 slots) and stay synced there for 5.5h with this PR along with:
+ Pass balances of justified checkpoint to forkchoice #1702 
+ Don't remove the head state from cache #1703 
+ Disconnect useless peers #1683 
+ No signature verification
+ MAX_EPOCH = 10
+ No gossip subscription except for block topic

RAM stays consistently at 1.7 GB - 2 GB

wemeetagain (approved):

PR #1687, opened 2020-10-20
Use ChainEvent enum

wemeetagain:
- add thorough comments on `ChainEvent` and members
- use `ChainEvent` throughout
- remove `voluntaryExit` event, derivative of `block` event


twoeths (approved):
lgtm

Issue #1053, opened 2020-06-25
Save ENR to disk on ENR update


wemeetagain:
See https://discordapp.com/channels/593655374469660673/593655641445367808/725848665327140917

We should save the ENR to disk whenever it updates.
It may be difficult to have ENR extend EventEmitter (as it currently already extends Map), but we may find another way to keep the ENR updated on disk.
A strawman proposal:
Extend ENR with a subclass which accepts a path in its constructor, overrides `set` to save to disk after updating.

dapplion:
@wemeetagain Is this solved with the recent ENR changes?

3xtr4t3rr3str14l:
resolved in #1694

PR #1694, opened 2020-10-22
FileENR

3xtr4t3rr3str14l:
implements suggested change in #1053


packages/lodestar-cli/src/cmds/beacon/fileEnr.ts
```diff
@@ -0,0 +1,40 @@
+import {createKeypairFromPeerId, ENR, ENRKey, ENRValue} from "@chainsafe/discv5";
```
wemeetagain: can you move this file to `src/network`

packages/lodestar-cli/src/cmds/beacon/fileEnr.ts
```diff
@@ -0,0 +1,40 @@
...
+import {writeFileSync} from "fs";
+import PeerId from "peer-id";
+import {readFileSync} from "../../util";
+
```
wemeetagain: Can you add a jsdoc comment here?
```
/**
 * `FileENR` is an `ENR` that saves the ENR contents to a file on every modification
 */
```

packages/lodestar-cli/src/cmds/beacon/fileEnr.ts
```diff
@@ -0,0 +1,40 @@
...
+  private filename: string;
+  private localPeerId: PeerId;
+
+  constructor(filename: string, peerId: PeerId) {
+    super();
+    this.filename = filename;
+    this.localPeerId = peerId;
+  }
+
+  static initFromFile(filename: string, peerId: PeerId): FileENR {
```
wemeetagain: We will likely want/need another static method `initFromENR(filename: string, peerId: PeerId, enr: ENR): FileENR`
to be used :point_down: in `src/network/enr.ts`

packages/lodestar-cli/src/network/enr.ts
```diff
@@ -13,10 +12,6 @@ export async function writeEnr(filename: string, enr: ENR, peerId: PeerId): Prom
   await writeFile(filename, enr.encodeTxt(keypair.privateKey));
 }
 
-export async function readEnr(filename: string): Promise<ENR> {
-  return ENR.decodeTxt(await readFile(filename));
-}
-
 export async function initEnr(filename: string, peerId: PeerId): Promise<void> {
```
wemeetagain, line 10: This `initEnr` will write the ENR to the file differently than how FileENR does.
The difference between:
`"enr:....."` - a valid json file
and
`enr:.....` - not a valid json file, plain text file

We should standardize how the enr is written across the board to _always_ either write quoted enr or _always_ write non-quoted enr.

Imo we don't need valid json here, we can just write the enr as plain text:

so `writeEnr` should be deleted, and `initEnr` should create and save an enr using `FileENR.initFromENR(...).saveToFile();`
Also, we would change the `beaconPaths.enrFile` defaults to be `$path/enr` rather than `$path/enr.json` since we won't be writing valid json anymore.

packages/lodestar-cli/src/network/test.js
```diff
@@ -0,0 +1,10 @@
+const {FileENR} = require('./lib/cmds/beacon/fileEnr.js');
```
wemeetagain: Test file should be converted to a unit test.

Unit test idea:
- create tmp file/path (we have a dependency which helps)
- create a new ENR, then FileENR (either using `new FileENR(...)` or `FileENR.initFromENR(...)`)
- update the enr
- check file for update (read the file into a new ENR, check the contents for the update)
- pass if the file has been written/updated

packages/lodestar-cli/src/cmds/beacon/handler.ts
```diff
@@ -28,8 +29,8 @@ export async function beaconHandler(options: IBeaconArgs & IGlobalArgs): Promise
 
   options = mergeConfigOptions(options);
   const peerId = await readPeerId(beaconPaths.peerIdFile);
-  // read local enr from disk
-  const enr = await readEnr(beaconPaths.enrFile);
+  // read local enr from disk; returns a FileENR which persists enr updates to a file
+  const enr = FileENR.initFromFile(beaconPaths.enrFile, peerId);
```
wemeetagain, line 33: If this PR is putting `FileENR` to use, it should also remove the "save ENR on shutdown" logic.

wemeetagain (changes requested):

packages/lodestar-cli/test/unit/network/fileEnr.test.ts
```diff
@@ -0,0 +1,23 @@
...
+import {expect} from "chai";
+import {existsSync} from "fs";
+import {initEnr, initPeerId, readPeerId} from "../../../src/network";
+import {FileENR} from "../../../src/network/fileEnr";
+import {rootDir} from "../../constants";
+
+describe("fileENR", function () {
+  it("create ENR from file", async function () {
+    const enrFilePath = `${rootDir}/enr`;
+    const peerIdFile = `${rootDir}/peer-id.json`;
```
dapplion: You can use the `getBeaconPaths()` utility to compute these paths

packages/lodestar-cli/test/unit/network/fileEnr.test.ts
```diff
@@ -0,0 +1,23 @@
...
+    await initPeerId(peerIdFile);
+    const peerId = await readPeerId(peerIdFile);
+    await initEnr(enrFilePath, peerId);
+    const enr = FileENR.initFromFile(enrFilePath, peerId);
+    const newValue = new Uint8Array(55);
+    enr.set("tcp", newValue);
+    expect(existsSync(enrFilePath)).to.be.true;
```
dapplion: When using multiple expect in a file use a clear error message such as
```ts
expect(existsSync(enrFilePath), `ENR does not exist at ${enrFilePath}`).to.be.true;
```

packages/lodestar-cli/test/unit/setup.ts
```diff
@@ -0,0 +1,11 @@
+import {rootDir} from "../constants";
```
dapplion: Not a fan of a global setup file that can affect other tests. If you need these steps in the ENR test, put this before and after hooks in the ENR test, and don't use any setup file for the unit tests

packages/lodestar-cli/src/network/fileEnr.ts
```diff
@@ -0,0 +1,46 @@
...
+    this.filename = filename;
+    this.localPeerId = peerId;
+  }
+
+  static initFromFile(filename: string, peerId: PeerId): FileENR {
+    const enr = FileENR.decodeTxt(readFileSync(filename, "utf8")) as FileENR;
+    return this.initFromENR(filename, peerId, enr);
+  }
+
+  static initFromENR(filename: string, peerId: PeerId, enr: FileENR): FileENR {
```
dapplion: Add a unit test for `initFromENR` to make sure the prototype manipulation is ok

dapplion (dismissed):
Looks really good overall! Nice idea, just some comments

packages/lodestar-cli/src/network/fileEnr.ts
3xtr4t3rr3str14l: `initFromFile` calls `initFromENR`, so the logic is tested there.

wemeetagain: `enr` should be type `ENR` here

wemeetagain (approved):
lgtm :+1:

PR #1700, opened 2020-10-23
Reorg lodestar options & config handling

### LINKED ISSUE #1684: Discussion: remove deposit functionality
> From what I understand the general consensus these days is that only audited deposit tools should be used, and as such some clients are removing above mentioned functionality (e.g Prysm)

See https://github.com/sigp/lighthouse/issues/1727

dapplion:
### Goals

- Typesafe all options in the CLI. Ensure each arg has an option and that it matches an actual option in Lodestar
- Enable yargs strict options. If the user types `--api.rest.enabled2` the CLI will throw an "Unknown option" error
- Remove variable mutation. Mutations are bad. Mutations make the code really harder to maintain. Mutations must go.
- Deduplicate code
- Remove default exports in Lodestar's options. Properly name each option object and export as named

### Changes

Changes necessary to achieve the goals above

- Disable yargs parser dot-notation. It's a good idea but you get no static Typesafety and it has bugs with the strict options code that make both unsuitable.
- Manually map a doted CLI arg `"api.rest.enabled"` to an IBeaconNodeOptions object, to replace dot-notation. It's not that bad code-wise, and you get static typesafety.
- Refactor the init handler and how other commands use it to initialize things to remove mutation. Use a regular function approach that returns initialized variables for other commands to use.
- Re-do the CLI e2e test to not rely and a setTimeout to guess completed time + stubes. Added a `getLodestarCliTestRunner` utility with which you can run commands directly and await them
```ts
const lodestar = getLodestarCliTestRunner()
const walletNames = await lodestar(`account wallet list --rootDir ${rootDir}`
```
- Re-org testnet settings to reduce code duplication. Since a testnet's config will never change, keep its diff in the src, instead of fetching the config remotely.
- Use a `getBeaconConfig` function to get the config instead of relying on mutations in the init command
- Isolate config and options handling in separate utils. Unit test those.
- Unit test path parsers (getBeaconPaths, getAccountPaths)
- Remove mutation from the path parsers

Others / minor
- Remove the `account validator deposit` command. Fixes https://github.com/ChainSafe/lodestar/issues/1684
- Rename all CLI args to `args`. Previously they were mixed `options` and `args`
- Add `minConnections` option from updated libp2p dependency
- Improve the RecursivePartial type. The previous version caused issues with deeply nested types like network's discv5 options.
- Use synchronous fs methods to interact with small files = simpler code
- Fix init command description


mpetrunic (commented):
Generally looks good to me. It's a huge diff so it's hard to spot any major bugs

dapplion:
> Generally looks good to me. It's a huge diff so it's hard to spot any major bugs

Running the branch locally with your common CLI configurations would be really helpful if you have a chance.

wemeetagain (dismissed):

3xtr4t3rr3str14l (commented):
are the beacon.test.ts and validator.test.ts in lodestar-cli e2e tests going to have anything in them for this PR?

packages/lodestar-cli/src/cmds/dev/handler.ts
```diff
@@ -1,97 +1,89 @@
...
-  const chainDir = join(options.rootDir, "dev", "beacon");
-  const validatorsDir = join(options.rootDir, "dev", "validators");
-  mkdirSync(chainDir, {recursive: true});
-  mkdirSync(validatorsDir, {recursive: true});
+  mkdir(chainDir);
+  mkdir(validatorsDir);
 
-  options.db.name = join(chainDir, "db-" + peerId.toB58String());
-  options.eth1.enabled = false;
+  // TODO: Rename db.name to db.path or db.location
```
3xtr4t3rr3str14l, line 46: is this something that will be handled in this PR?  or do we need to make a separate issue for it?

packages/lodestar-cli/test/e2e/cmds/account.test.ts
```diff
@@ -0,0 +1,142 @@
...
+      // ⏎
+      "account validator list",
+      `--rootDir ${rootDir}`,
+    ]);
+
+    // Write order is not guarranteed, sort before comparing
+    expect(validatorPubKeys.sort()).to.deep.equal(createdPubkeys.sort(), "Wrong validator pubkeys");
+  });
+
+  // it("should make a deposit to validator registration contract", async function () {
```
3xtr4t3rr3str14l: can you remove this commented code since we are removing the deposit command?

packages/lodestar-cli/src/options/beaconNodeOptions/index.ts
```diff
@@ -1,28 +1,38 @@
...
+  sync.ISyncArgs;
 
-// Re-export for convenience
-export type IBeaconNodeOptions = _IBeaconNodeOptions;
-export type IBeaconNodeOptionsPartial = RecursivePartial<_IBeaconNodeOptions>;
+export function parseBeaconNodeArgs(args: IBeaconNodeArgs): RecursivePartial<IBeaconNodeOptions> {
+  // Remove undefined values to allow deepmerge to inject default values downstream
+  return removeUndefinedRecursive({
+    api: api.parseArgs(args),
+    // chain: {},
```
3xtr4t3rr3str14l, line 21: are these two comments here for a particular reason?  if not, can you remove them?

3xtr4t3rr3str14l (commented):
awesome!  this significantly cleans and tightens the cli.  a few minor, mostly cosmetic comments

packages/lodestar-cli/src/cmds/dev/handler.ts
dapplion: A new PR / issue

packages/lodestar-cli/src/options/beaconNodeOptions/index.ts
dapplion: They state that these properties exist in IBeaconNodeOptions but none of them are purposefully exposed as CLI options directly

dapplion:
e2e tests are passing but the simulation tests are failing. I don't see how these CLI changes may affect that test. Let me know if you think the errors should be addressed here

mpetrunic:
> e2e tests are passing but the simulation tests are failing. I don't see how these CLI changes may affect that test. Let me know if you think the errors should be addressed here

seems like minPeers/maxPeers issue, did you change anything on that front?

dapplion:
> > e2e tests are passing but the simulation tests are failing. I don't see how these CLI changes may affect that test. Let me know if you think the errors should be addressed here
> 
> seems like minPeers/maxPeers issue, did you change anything on that front?

Yes I did change something there. Could you paste here the error that points in that direction? I don't see it in the logs

wemeetagain (approved):

Issue #1701, opened 2020-10-24
State regen causes OOM


twoeths:
**Describe the bug**
Sync to slot 578307 and regen state from 502561 which caused a huge block range to reprocess and OOM happened
```
2020-10-24 11:59:20 [CHAIN]           ^[[34mdebug^[[39m: Block processed slot=578307, root=0xf147a2017ddb16d89b       4e82e0f9097adde4ec6d8e63064177da59351f45819eef
2020-10-24 11:59:24 [CHAIN]           ^[[34mdebug^[[39m: Block processed slot=502561, root=0xfe34181492bd5ae731       b03b39ade8e40ccb388d32cd4a9222a8b532848f22b10a
2020-10-24 11:59:24 [CHAIN]           ^[[34mdebug^[[39m: Block processed slot=502562, root=0x9e0abf1d2cd25f241e       3415768ac3bbef11be0da78f5bd6f6bb2d85b33f8f4222
...
2020-10-24 12:00:08 [CHAIN]           ^[[34mdebug^[[39m: Block processed slot=502684, root=0xc6832f690a0b129728       e7ef322d999147e50d297ff4aa0e3aeb856b512bfecee9
173099 2020-10-24 12:00:08 [CHAIN]           ^[[34mdebug^[[39m: Block processed slot=502687, root=0x5bce3c5f9d31986171       80ad012b8cd5b994b202df52d2dcd321c0c97170de21e1
173100
173101 <--- Last few GCs --->
173102
173103 [98471:0x10284d000] 16936548 ms: Mark-sweep 2004.3 (2067.5) -> 1988.3 (2067.5) MB, 1242.8 / 0.1 ms  (average mu        = 0.116, current mu = 0.029) allocation failure scavenge might not succeed
173104 [98471:0x10284d000] 16937903 ms: Mark-sweep 2004.6 (2067.8) -> 1988.6 (2067.8) MB, 1318.9 / 0.1 ms  (average mu        = 0.072, current mu = 0.027) allocation failure scavenge might not succeed
```

Note that block 502558 formed the justified checkpoint earlier, that's the last cached state where regen started from.
```
2020-10-24 07:36:39 [CHAIN]         ^[[36mverbose^[[39m: Fork choice justified epoch=15705, root=0x66ba71dfb29b       ada27c3f99e9823dac4272ff1a057814d0672353358571cb0142
```
**Expected behavior**
State regen should not cause OOM.

twoeths:
[...]

PR #1703, opened 2020-10-27
Avoid removing stateRoot of forkchoice head from state context cache

twoeths:
remedy #1701 
+ throw regen error if we have to replay too many blocks to regen state

### DIFF
--- a/packages/lodestar/src/chain/regen/regen.ts
+++ b/packages/lodestar/src/chain/regen/regen.ts
@@ -141,7 +141,8 @@ export class StateRegenerator implements IStateRegenerator {
         code: RegenErrorCode.ERR_NO_SEED_STATE,
       });
     }
-    if (blocksToReplay.length > 5 * this.config.params.SLOTS_PER_EPOCH) {
+    const MAX_EPOCH_TO_PROCESS = 5;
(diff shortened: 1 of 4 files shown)


packages/lodestar/src/chain/regen/regen.ts
```diff
@@ -141,6 +141,12 @@ export class StateRegenerator implements IStateRegenerator {
         code: RegenErrorCode.ERR_NO_SEED_STATE,
       });
     }
+    if (blocksToReplay.length > 5 * this.config.params.SLOTS_PER_EPOCH) {
```
mpetrunic: extract 5 to some constant like MAX_EPOCH_TO_PROCESS

3xtr4t3rr3str14l (approved):
LGTM

packages/lodestar/src/chain/regen/regen.ts
dapplion: Consider enforcing this comment programatically with this rule 
**Disallow Magic Numbers (no-magic-numbers)**
https://eslint.org/docs/rules/no-magic-numbers#:~:text=The%20no%2Dmagic%2Dnumbers%20rule,to%20make%20their%20meaning%20explicit.

wemeetagain (approved):

PR #1706, opened 2020-10-27
Split long e2e tests

dapplion:
Run the node heavy integration tests in a separate CI/CD step to clearly know that is those that are failing


package.json
```diff
@@ -18,6 +18,7 @@
     "lint": "lerna run lint --no-bail",
     "test:unit": "lerna run test:unit --no-bail",
     "test:e2e": "lerna run test:e2e --no-bail",
+    "test:e2e:long": "lerna run test:e2e:long --no-bail",
```
mpetrunic: maybe call it "sim" (simulation) instead of "long" since long is relative in e2e :smile:

dapplion: A much better name yes! Thanks

wemeetagain (approved):

PR #1707, opened 2020-10-28
Refactor node initialization

wemeetagain:
This PR refactors the `BeaconChain`,  `BeaconNode`, and `ForkChoice` initialization.

Major changes:
- initialize an anchor state (either from file, db, or eth1)
  - the three different state initializations are refactored into functions, previously living in various chain methods
- `BeaconNode#init` is used to initialize a node from an anchor state
  - it creates all non-external submodules, wired together and started
  - the `BeaconNode` constructor simply attaches preconfigured submodules to the node as properties
- `BeaconChain` now is fully initialized in the constructor, no need for an asynchronous initialization, much simplified
  - chain requires a genesis/anchor state passed in the constructor, used to seed all chain submodules
  - chain no longer includes any Eth1Provider / GenesisBuilder code, its handled ^ when the anchor state is initialized
- `ForkChoice` is now subclassed - moves the fork choice initialization previously living as a chain method into the new constructor
  - this subclass can/will be extended to emit `forkChoice:head` and `forkChoice:reorg` events in a future PR
- as a result of this PR, the node now seems to stop cleanly, no longer hanging


packages/lodestar-cli/src/cmds/beacon/handler.ts
```diff
@@ -38,29 +42,33 @@ export async function beaconHandler(options: IBeaconArgs & IGlobalArgs): Promise
...
+  const anchorState = await initBeaconState(options, config, db, logger, abortController.signal);
+  const node = await BeaconNode.init({
+    opts: options,
+    config,
+    db,
+    logger,
+    libp2p: await createNodeJsLibp2p(peerId, options.network, options.peerStoreDir),
+    anchorState,
+  });
+  abortController.signal.addEventListener("abort", () => node.close(), {once: true});
```
mpetrunic, line 72: But if database closes at the same time, some node close actions might get corrupted.

packages/lodestar/src/api/impl/beacon/beacon.ts
```diff
@@ -66,7 +66,7 @@ export class BeaconApi implements IBeaconApi {
 
   public async getFork(): Promise<ForkResponse> {
     const state = await this.chain.getHeadState();
-    const networkId: Uint64 = this.chain.networkId;
```
mpetrunic: what's networkId, do we even need it since we use forkVersion?

packages/lodestar/src/chain/forkChoice/forkChoice.ts
```diff
@@ -0,0 +1,55 @@
...
+import {toHexString} from "@chainsafe/ssz";
+import {BeaconState, Slot} from "@chainsafe/lodestar-types";
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
+import {ForkChoice, ProtoArray} from "@chainsafe/lodestar-fork-choice";
+
+import {computeAnchorCheckpoint} from "../initState";
+import {ChainEventEmitter} from "../emitter";
+import {ForkChoiceStore} from "./store";
+
+export class EventedForkChoice extends ForkChoice {
```
mpetrunic: maybe call it LodestarForkChoice? I get that you are trying to hint that we extend fork choice with event emitting but it's tied to lodestar event emitting and forkchoice store. Just a suiggestion, not a strong opinion.

packages/lodestar/src/node/nodejs.ts
```diff
@@ -2,171 +2,234 @@
...
+    config,
+    db,
+    logger,
+    libp2p,
+    anchorState,
+  }: IBeaconNodeInitModules): Promise<T> {
+    const controller = new AbortController();
+    await initBLS();
+    // start db if not already started
+    await (db as BeaconDb).start();
```
mpetrunic: why casting, doesn't IBeaconDb have start method?

packages/lodestar/src/api/impl/beacon/beacon.ts
wemeetagain: nope haha, i think after your standard fork API PR, we don't even use it at all

mpetrunic (approved):

packages/lodestar/src/chain/forkChoice/forkChoice.ts
```diff
@@ -0,0 +1,58 @@
...
+    config: IBeaconConfig;
+    emitter: ChainEventEmitter;
+    currentSlot: Slot;
+    anchorState: BeaconState;
+  }) {
+    const {blockHeader, checkpoint} = computeAnchorCheckpoint(config, anchorState);
+    const justifiedCheckpoint = {...checkpoint};
+    const finalizedCheckpoint = {
+      ...checkpoint,
+      // If not genesis epoch, justified checkpoint epoch must be set to finalized checkpoint epoch + 1
```
twoeths, line 33: the code does not go with the comment, maybe switch `finalizedCheckpoint` and `justifiedCheckpoint` assignment?

twoeths: created https://github.com/ChainSafe/lodestar/pull/1735/files

PR #1712, opened 2020-10-30
Only subscribe to validator topics when node is synced

### LINKED ISSUE #1697: Node has to process slots through giant range
**Describe the bug**
At some point I saw my node kept processing through checkpoints. The debug shown that it has to do `processSlotsToNearestCheckpoint` from slot 503119 to 572736, almost 70000 slots!

Screen Shot 2020-10-23 at 09 57 38

Note this log: `Checkpoint processed epoch=15725, root=0x3b931133a8897f0785da2754e57c1e535a77e3388aaccaac26496cc882c56e8e` happened a lot. Root is the same, epoch is increased gradually

The result is the node is very busy processing that request.

**Expected behavior**
+ Figure out what caused the issue and prevent it
+ This is the log file
[processSlotsToNearestCheckpoint.log.zip](https://github.com/ChainSafe/lodestar/files/5427059/processSlotsToNearestCheckpoint.log.zip)

twoeths:
resolves #1697 

This is part of the formation that helped me sync up to head:
+ Only subscribe to gossip validator topics when node is synced
+ Reduce `MAX_EPOCHS` in checkpoint state cache to 10 (otherwise we'll get OOM error)

### DIFF
--- a/packages/lodestar/src/db/api/beacon/stateContextCheckpointsCache.ts
+++ b/packages/lodestar/src/db/api/beacon/stateContextCheckpointsCache.ts
@@ -75,7 +75,7 @@ export class CheckpointStateCache {
     const epochs = Object.keys(this.epochIndex)
       .map(Number)
       .filter((epoch) => epoch !== finalizedEpoch && epoch !== justifiedEpoch);
-    const MAX_EPOCHS = 20;
+    const MAX_EPOCHS = 10;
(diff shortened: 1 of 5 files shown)


mpetrunic (commented):
I guess this is fine, would be cleaner if we do this via events

wemeetagain (approved):

PR #1719, opened 2020-11-01
Split finality tests

dapplion:
`mocha -parallel` is used to reduce the total time to run spec fast tests. However, it distributes between workers on a file basis. The finality tests are way longer than the rest so that single file is the limiting factor. By splitting the finality tests between two files mocha --parallel can kick in and do its magic, reducing the step time by 2-3 min

I've changed the spec test runner to allow me to pass an array of test directories instead of their parent


dapplion:
After testing a few rounds, there seems to be no time improvement

PR #1721, opened 2020-11-01
Split state transition naive / fast packages

dapplion:
- Naive state transition -> `@chainsafe/lodestar-beacon-state-transition`
- Fast state transition -> `@chainsafe/lodestar-beacon-state-transition-fast`

Some tests are falling in the lodestar package due to very weird stubs of the beacon-state-transition-fast utilities. Are they really necessary? Can the same tests by done tweaking data instead of stubbing imports?


mpetrunic:
> Some tests are falling in the lodestar package due to very weird stubs of the beacon-state-transition-fast utilities. Are they really necessary? Can the same tests by done tweaking data instead of stubbing imports?

That would defeat point of unit testing since change in state transition package would invalidate unit tests in lodestar. Stubs are you friends (if written and used properly), while data tweaking is enemy :stuck_out_tongue: 

There are some build errors as well.

Issue #1715, opened 2020-11-01
OOM after node is synced to head


twoeths:
**Describe the bug**
Node is synced to head:
```
2020-11-01 01:36:48 [SYNC]             info: Regular Sync: processed up to current slot 635283
2020-11-01 01:36:48 [SYNC]             info: Regular Sync: fetcher returns empty array, finish sync now
```
saw some valid gossip blocks:
```
2020-11-01 01:37:27 [GOSSIP]        ^[[36mverbose^[[39m: Incoming gossip block at slot: 635286
...
 2020-11-01 01:52:34 [CHAIN]           ^[[34mdebug^[[39m: Block processed slot=635360, root=0xfe797cb4160009147d       901b9b49db4871f52124a50bb3aa422a0a527128fbbc94
```
2 minutes after that node is crashed due to OOM

```
<--- JS stacktrace --->
739036
739037 ==== JS stack trace =========================================
739038
739039     0: ExitFrame [pc: 0x1009288f9]
739040 Security context: 0x21ede0380919 <JSObject>
739041     1: rebindLeft [0x21ed62d0ecb1] [/Users/tuyennguyen/Projects/workshop/lodestar/node_modules/@chainsafe/persi       stent-merkle-tree/lib/node.js:~57] [pc=0x376646bdff53](this=0x21ed386566f9 <Node map = 0x21edad3809f9>,0x21ed38       668f59 <Node map = 0x21edad3809f9>)
739042     2: /* anonymous */(aka /* anonymous */) [0x21ed38656e61] [/Users/tuyennguyen/Projects/workshop/lodest...
```
**Expected behavior**
Node can stay synced

I notice that previously I disabled subscribing to all gossip topics (except for `block`), now I apply  #1712 so that should cause the issue.

twoeths:
Right before OOM happened, there is `ERR_QUEUE_THROTTLED` which means node had to regen a lot.

```
For help, see: https://nodejs.org/en/docs/inspector
(node:21583) UnhandledPromiseRejectionWarning: Error: ERR_QUEUE_THROTTLED
    at JobQueue.enqueueJob (/Users/tuyennguyen/Projects/workshop/lodestar/packages/lodestar/src/util/queue.ts:63:13)
    at QueuedStateRegenerator.getState (/Users/tuyennguyen/Projects/workshop/lodestar/packages/lodestar/src/chain/regen/queued.ts:53:32)
[...]

PR #1723, opened 2020-11-02
Do not need to get head state through regen

twoeths:
part of #1715 
head state is always available in the state cache so we don't need to go through regen module to get it. Otherwise:
+ the consume api has to wait for existing jobs in the queue unnecessarily
+ It'll throw `ERR_QUEUE_THROTTLED` because we get head state very frequently


twoeths:
not able to start node due to "headStateRoot does not exist" but still leave here to see if this is correct approach

twoeths:
open #1872 to track this

PR #1731, opened 2020-11-03
get genesisStateFileUrl if testnet arg is passed to beacon

### LINKED ISSUE #1730: lodestar-cli not downloading genesis.ssz for medalla
when i run `lodestar init --testnet medalla`, it doesn't download the genesis.ssz

3xtr4t3rr3str14l:
fixes #1730  (a regression issue caused by the functionality for downloading of genesis.ssz being removed) by downloading the genesis.ssz when running `beacon` if the `--testnet` param is passed

### DIFF
--- a/packages/lodestar-cli/src/cmds/beacon/handler.ts
+++ b/packages/lodestar-cli/src/cmds/beacon/handler.ts
@@ -15,6 +15,7 @@ import {getBeaconPaths} from "./paths";
 import {onGracefulShutdown} from "../../util/process";
 import {FileENR, overwriteEnrWithCliArgs, readPeerId} from "../../config";
 import {initBeaconState} from "./initBeaconState";
+import {getGenesisFileUrl} from "../../testnets";
 
 /**
  * Run a beacon node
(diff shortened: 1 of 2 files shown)


packages/lodestar-cli/src/cmds/beacon/handler.ts
```diff
@@ -58,6 +59,10 @@ export async function beaconHandler(args: IBeaconArgs & IGlobalArgs): Promise<vo
   abortController.signal.addEventListener("abort", dbClose, {once: true});
   await db.start();
 
+  if (args.testnet && !args.genesisStateFile) {
```
dapplion: Would it make more sense to move this logic into `src/cmds/beacon/initBeaconState.ts`?

3xtr4t3rr3str14l: yeah i think so.  just pushed an update for this.

dapplion (approved):

wemeetagain (approved):

PR #1736, opened 2020-11-05
Regular Sync: check linear chain segment to avoid orphaned block

### LINKED ISSUE #1727: Forkchoice head getting stuck at an orphaned block
**Describe the bug**
This is the latest head I have
```
2020-11-03 01:18:25 [CHAIN]         verbose: New chain head headSlot=622522, headRoot=0x8778d4254b41f48534757e7a1fd1f9b82471494d3f1765b5f7b3f003de43405a
```
I notice that's an orphaned block
**Expected behavior**
Head should move forward during the sync.

**Steps to Reproduce**
Sync Medalla to head

twoeths:
resolves #1727 
and fix some found issues during the test of regular sync (hard coded to get through the big unfinality period)

+ avoid orphaned block to be imported by using a linear chain
+ a linear chain segment means every block being parent of next block and child of previous block
+ always switch peer in case of error, `getBestPeer()` now support an `excludedList` of bad peers to make sure we don't retry the old bad peers
+ slowly expand range end in the case of `handleEmptyRange`
+ have to go with sandbox fake timers instead of sinon to get through unit tests

### DIFF
--- a/packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
+++ b/packages/lodestar/src/sync/regular/oneRangeAhead/fetcher.ts
@@ -67,11 +67,9 @@ export class BlockRangeFetcher implements IBlockRangeFetcher {
         if (peer) {
           // peers may return incorrect empty range, or 1 block, or 2 blocks or unlinear chain segment
           // if we try the same peer it'll just return same result so switching peer here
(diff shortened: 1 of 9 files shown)


twoeths:
have tested this with Medalla for a while, ready for review

wemeetagain (approved):

PR #1737, opened 2020-11-06
Rest API state endpoints

### LINKED ISSUE #1057: Add state endpoints
   - add beacon/state endpoints
   - add beacon/genesis endpoint

mpetrunic:
Sorry for huge diff, a lot of stuff broke along the way.

resolves #1057


packages/lodestar/src/api/impl/beacon/state/interface.ts
```diff
@@ -1,8 +1,45 @@
...
 
 export type StateId = string | "head" | "genesis" | "finalized" | "justified";
+
+export type ValidatorStatus = "active";
+
+export interface IValidatorFilters {
+  indices?: (BLSPubkey | ValidatorIndex)[];
+  statuses?: ValidatorStatus[];
+}
+export interface ICommittesFilters {
```
wemeetagain: typo

packages/lodestar/src/api/impl/beacon/state/state.ts
```diff
@@ -1,26 +1,155 @@
...
+        .filter(notNullish);
+    }
+    return readOnlyMap(state.state.validators, (v, index) => {
+      return {
+        index,
+        balance: state.state.balances[index],
+      };
+    });
+  }
+  public async getStateCommittes(stateId: StateId, filters?: ICommittesFilters): Promise<BeaconCommitteeResponse[]> {
```
twoeths: typo

packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -37,8 +50,68 @@ export async function resolveStateId(
...
+
+export function validatorPubkeyToIndex(
+  config: IBeaconConfig,
+  stateCtx: ApiStateContext,
+  pubkey: BLSPubkey
+): ValidatorIndex | null {
+  if (stateCtx.epochCtx) {
+    return stateCtx.epochCtx.pubkey2index.get(pubkey) ?? null;
+  }
+  for (const [index, validator] of Array.from(stateCtx.state.validators).entries()) {
```
twoeths: this place can potentially receive DOS attach. As head state/context is always available in cache, can we just use it?

packages/lodestar/src/api/impl/errors/api.ts
```diff
@@ -5,3 +5,14 @@ export class ApiError extends Error {
     this.statusCode = statusCode;
   }
 }
+
+export class StateNotFound extends ApiError {
+  constructor() {
+    super(404, "State not found");
```
twoeths, line 11: can we put state id here so that we have this example error:
```
{
  "code": 400,
  "message": "Invalid state ID: current"
}
```

packages/lodestar/src/api/impl/beacon/state/errors.ts
```diff
@@ -0,0 +1,7 @@
+import {ApiError} from "../../errors/api";
+
+export class MissingState extends ApiError {
```
twoeths: how is this different from `StateNotFound` error?

packages/lodestar/src/api/impl/errors/api.ts
mpetrunic: That's not the same error, this is when there is correct state id but we don't have it in database.

packages/lodestar/src/api/impl/beacon/state/utils.ts
mpetrunic: Oh true, validators are never deleted, so we could always use head epochCtx

packages/lodestar/src/api/impl/beacon/state/errors.ts
mpetrunic: It isn't, gonna remove this one

twoeths (dismissed):
not related to this PR, I think we missed this interface https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar-beacon-state-transition/src/fast/util/interface.ts#L9

maybe we should create an issue to bring it back since we really only need `ReadOnly` interface in lodestar?

packages/lodestar/src/api/impl/beacon/state/state.ts
```diff
@@ -1,26 +1,159 @@
...
+        .filter(notNullish);
+    }
+    return readOnlyMap(state.state.validators, (v, index) => {
+      return {
+        index,
+        balance: state.state.balances[index],
+      };
+    });
+  }
+  public async getStateCommittees(stateId: StateId, filters?: ICommitteesFilters): Promise<BeaconCommitteeResponse[]> {
```
dapplion, line 129: Could you add more space (newlines) between code blocks? It helps readability a lot

mpetrunic: I hoped prettier would take care of that :disappointed:

packages/lodestar-types/src/ssz/utils/string.ts
```diff
@@ -0,0 +1,19 @@
+import {BasicType} from "@chainsafe/ssz";
+
+export class StringType<T extends string = string> extends BasicType<T> {
```
3xtr4t3rr3str14l, line 3: i noticed we also have a `StringType` in `packages/lodestar` (https://github.com/ChainSafe/lodestar/blob/f0dcca441ccab7364ce529f5736b8d7b9880ac4c/packages/lodestar/src/network/peers/sszTypes.ts#L3).  is this something we should consider adding to the ssz package?

mpetrunic: No. This is not standard or supported by ssz. This is just hack to allow json serilization of string vslues using ssz schema

wemeetagain (approved):

PR #1740, opened 2020-11-09
Update attester duties endpoint

mpetrunic:
- updates attester duties endpoint to standardized
- updates are on both lodestar and lodestar-validator side


packages/lodestar-validator/src/api/impl/rest/validator/validator.ts
```diff
@@ -12,23 +13,25 @@ import {
...
 
 export class RestValidatorApi implements IValidatorApi {
   private readonly client: HttpClient;
+  private readonly clientV2: HttpClient;
 
   private readonly config: IBeaconConfig;
 
   public constructor(config: IBeaconConfig, restUrl: string, logger: ILogger) {
     this.client = new HttpClient({urlPrefix: urlJoin(restUrl, "validator")}, {logger});
+    this.clientV2 = new HttpClient({urlPrefix: urlJoin(restUrl, "/eth/v1/validator")}, {logger});
```
dapplion, line 34: clientV2 = /eth/v1 ?

Would it be more clear to do clientV1 = /eth/v1

mpetrunic: When I'm done it will be just "client", this is just so I don't have to update everything at once

wemeetagain (approved):
:100: lgtm

PR #1742, opened 2020-11-09
reqResp cleanups

3xtr4t3rr3str14l:
just a few things i found when making sure we are up to spec for reqResp:

-  remove unnecessary check in onBeaconBlocksByRange
-  remove unnecessary function params
-  clarifying error message in reqUtils->sendRequest

### DIFF
--- a/packages/lodestar/src/network/reqresp/reqUtils.ts
+++ b/packages/lodestar/src/network/reqresp/reqUtils.ts
@@ -29,7 +29,7 @@ export async function sendRequest<T extends ResponseBody | ResponseBody[]>(
   signal?: AbortSignal
 ): Promise<T | null> {
   if (signal?.aborted) {
-    throw new Error("aborted");
+    throw new Error("sendRequest aborted");
   }
   const requestOnly = isRequestOnly(method);
   const requestSingleChunk = isRequestSingleChunk(method);
--- a/packages/lodestar/src/network/reqresp/reqUtils.ts
+++ b/packages/lodestar/src/network/reqresp/reqUtils.ts
@@ -29,7 +29,7 @@ export async function sendRequest<T extends ResponseBody | ResponseBody[]>(
   signal?: AbortSignal
 ): Promise<T | null> {
   if (signal?.aborted) {
-    throw new ErrorAborted("sendRequest aborted");
+    throw new ErrorAborted("sendRequest");
   }
   const requestOnly = isRequestOnly(method);
   const requestSingleChunk = isRequestSingleChunk(method);
--- a/packages/lodestar/src/network/reqresp/reqUtils.ts
+++ b/packages/lodestar/src/network/reqresp/reqUtils.ts
@@ -14,7 +14,7 @@ import {
   randomRequestId,
 } from "..";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
-import {ILogger} from "@chainsafe/lodestar-utils";
+import {ErrorAborted, ILogger} from "@chainsafe/lodestar-utils";
 import {Method, MethodRequestType, ReqRespEncoding, TTFB_TIMEOUT, REQUEST_TIMEOUT} from "../../constants";
 import {eth2RequestEncode} from "../encoders/request";
 import {eth2ResponseDecode} from "../encoders/response";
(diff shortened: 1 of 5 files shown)


packages/lodestar/src/sync/reqResp/reqResp.ts
```diff
@@ -97,11 +98,11 @@ export class BeaconReqRespHandler implements IReqRespHandler {
       case Method.Ping:
         return await this.onPing(request as ReqRespRequest<Ping>, peerId, sink);
       case Method.Metadata:
-        return await this.onMetadata(request as ReqRespRequest<null>, peerId, sink);
+        return await this.onMetadata(request as ReqRespRequest<null>, sink);
       case Method.BeaconBlocksByRange:
-        return await this.onBeaconBlocksByRange(request as ReqRespRequest<BeaconBlocksByRangeRequest>, peerId, sink);
+        return await this.onBeaconBlocksByRange(request as ReqRespRequest<BeaconBlocksByRangeRequest>, sink);
       case Method.BeaconBlocksByRoot:
-        return await this.onBeaconBlocksByRoot(request as ReqRespRequest<BeaconBlocksByRootRequest>, peerId, sink);
```
mpetrunic: I would keep peerId on those methods. We could/should track those requests by peerId so we can blacklist peer if requests too much blocks or something

packages/lodestar/src/sync/reqResp/reqResp.ts
```diff
@@ -280,10 +280,6 @@ export class BeaconReqRespHandler implements IReqRespHandler {
       );
       return;
     }
-    if (request.body.count > 1000) {
-      this.logger.warn(`Request id ${request.id} asked for ${request.body.count} blocks, just return 1000 maximum`);
-      request.body.count = 1000;
-    }
```
mpetrunic, line 286: why did you remove this?
You could change this so that "1000" (maxRequestedBlocks) be an ReqResp option.

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -226,10 +226,7 @@ export async function createStatus(chain: IBeaconChain): Promise<Status> {
 export async function syncPeersStatus(network: INetwork, status: Status): Promise<void> {
   await Promise.all(
     network.getPeers({connected: true, supportsProtocols: getStatusProtocols()}).map(async (peer) => {
-      try {
-        network.peerMetadata.setStatus(peer.id, await network.reqResp.status(peer.id, status));
-        // eslint-disable-next-line no-empty
-      } catch {}
+      network.peerMetadata.setStatus(peer.id, await network.reqResp.status(peer.id, status));
```
mpetrunic: we need try catch here otherwise it's unhandled rejection when `await network.reqResp.status(peer.id, status)` throws

mpetrunic (changes requested):

packages/lodestar/src/network/reqresp/reqUtils.ts
```diff
@@ -29,7 +29,7 @@ export async function sendRequest<T extends ResponseBody | ResponseBody[]>(
   signal?: AbortSignal
 ): Promise<T | null> {
   if (signal?.aborted) {
-    throw new Error("aborted");
+    throw new Error("sendRequest aborted");
```
dapplion: You can use `ErrorAborted` from lodestar-utils

packages/lodestar/src/sync/reqResp/reqResp.ts
3xtr4t3rr3str14l: We have an identical check just below it with the correct number specified in the spec:
https://github.com/ChainSafe/lodestar/pull/1742/files#diff-7c5831b4dde5a221fa8a043078154dcd1d94033976ba1c25cc0e493ee3c7d156R284

mpetrunic (dismissed):

packages/lodestar/src/network/reqresp/reqUtils.ts
```diff
@@ -29,7 +29,7 @@ export async function sendRequest<T extends ResponseBody | ResponseBody[]>(
   signal?: AbortSignal
 ): Promise<T | null> {
   if (signal?.aborted) {
-    throw new Error("aborted");
+    throw new ErrorAborted("sendRequest aborted");
```
dapplion: This would print `Aborted sendRequest aborted`. Either modify the `ErrorAborted` constructor or this message

dapplion (dismissed):

wemeetagain (approved):

PR #1743, opened 2020-11-09
fix slashingProtection reference

3xtr4t3rr3str14l:
I was getting an error `Error: Cannot find module '@chainsafe/lodestar-validator/src/slashingProtection/utils'` when trying to run `init`, so i changed the reference to be local rather than referencing the global package name to get past the error.


dapplion:
Sorry, I have to watch out for these bad import statements added by VSCode automatically

dapplion:
Scanned the entire codebase and there are no more instances of self imports

dapplion (approved):

twoeths (approved):

twoeths (approved):

PR #1746, opened 2020-11-10
Fix vscode self reference

twoeths:
I found one more place in lodestar-validator that refer to itself


mpetrunic (approved):

twoeths:
duplicate to #1743

PR #1752, opened 2020-11-11
allow bigints in logger and apply to onGoodbye log message

### LINKED ISSUE #1750: some goodbye reasons misprinting reason contents
**Describe the bug**

I have seen some messages like this when running the beacon node:
```2020-11-10 17:20:20 [SYNC]             info: Received goodbye request from 16Uiu2HAmMUNbsTEnkSGWAmfw9YEr8BeuBgTJ5mfyZ35WgpRyo9n6, reason=[object Object]```

**Expected behavior**

1) the reason should be printed in full instead of showing as `[object Object]`

### PR DESCRIPTION
**Pull request by 3xtr4t3rr3str14l**
resolves #1750 

Made an update to allow bigints in the logger by changing the Context object and upgrading winston. If everyone likes this, I'll go to the other instances where we are needlessly stringifying for the logger and just pass in the raw object instead (where applicable, like i've done for the goodbye log here).

### DIFF
--- a/packages/lodestar/src/sync/reqResp/reqResp.ts
+++ b/packages/lodestar/src/sync/reqResp/reqResp.ts
@@ -14,7 +14,7 @@ import {
   SignedBeaconBlock,
   Status,
 } from "@chainsafe/lodestar-types";
-import {ILogger, toJson} from "@chainsafe/lodestar-utils";
+import {ILogger} from "@chainsafe/lodestar-utils";
 import {toHexString} from "@chainsafe/ssz";
 import PeerId from "peer-id";
 import {IBeaconChain} from "../../chain";
@@ -219,12 +219,7 @@ export class BeaconReqRespHandler implements IReqRespHandler {
     sink: Sink<unknown, unknown>
   ): Promise<void> {
     this.logger.info(
-      `Received goodbye request from ${peerId.toB58String()}, reason=${JSON.stringify({
-        id: toJson(request.id),
-        method: toJson(request.method),
-        body: this.config.types.Uint64.toJson(request.body),
-        encoding: toJson(request.encoding),
-      })}`
+      `Received goodbye request from ${peerId.toB58String()}, reason=${this.config.types.Uint64.toJson(request.body)}`
     );
     await sendResponse(
       {config: this.config, logger: this.logger},
--- a/packages/lodestar/src/sync/reqResp/reqResp.ts
+++ b/packages/lodestar/src/sync/reqResp/reqResp.ts
@@ -14,7 +14,7 @@ import {
   SignedBeaconBlock,
   Status,
 } from "@chainsafe/lodestar-types";
-import {ILogger} from "@chainsafe/lodestar-utils";
+import {ILogger, toJson} from "@chainsafe/lodestar-utils";
 import {toHexString} from "@chainsafe/ssz";
 import PeerId from "peer-id";
 import {IBeaconChain} from "../../chain";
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by mpetrunic — DISMISSED** [review]

File: packages/lodestar/src/sync/reqResp/reqResp.ts [hunk 1]
```diff
@@ -218,7 +218,9 @@ export class BeaconReqRespHandler implements IReqRespHandler {
     peerId: PeerId,
     sink: Sink<unknown, unknown>
   ): Promise<void> {
-    this.logger.info(`Received goodbye request from ${peerId.toB58String()}, reason=${request}`);
+    this.logger.info(
+      `Received goodbye request from ${peerId.toB58String()}, reason=${this.config.types.Uint64.toJson(request.body)}`
+    );
```
dapplion: Isn't the logger able to stringify it by itself? This should work or we have to make it work
```ts
this.logger.info("Received goodbye request", {peer: peerId.toB58String(), reason: request.body});
```

3xtr4t3rr3str14l: you're right.  just pushed a change for this.

packages/lodestar/src/sync/reqResp/reqResp.ts
```diff
@@ -218,7 +218,7 @@ export class BeaconReqRespHandler implements IReqRespHandler {
     peerId: PeerId,
     sink: Sink<unknown, unknown>
   ): Promise<void> {
-    this.logger.info(`Received goodbye request from ${peerId.toB58String()}, reason=${request}`);
+    this.logger.info(`Received goodbye request from ${peerId.toB58String()}, reason=${request.body}`);
```
dapplion: Note that here you are still stringifying the data instead of passing JSON to the logger. I would suggest to do this:
```ts
this.logger.info("Received goodbye request", {peer: peerId.toB58String(), reason: request.body});
```
If you follow this suggestion please test at least once that the output looks as expected

3xtr4t3rr3str14l: ah yeah, the logger doesn't like bigints apparently.  when i try your line i get 
```
Types of property 'reason' are incompatible.
    Type 'bigint' is not assignable to type 'string | number | boolean | { [property: string]: Json; } | Json[] | null | undefined'.
```

in my IDE

dapplion: Then we should modify the logger to accept bigints. Do a similar approach like describe here https://github.com/ChainSafe/lodestar/pull/1755#issuecomment-726914113
Add test cases for the data that needs to be logged and modify the logger functions accordingly

3xtr4t3rr3str14l:
Made an update to allow bigints in the logger by changing the `Context` object and upgrading winston.  If everyone likes this, I'll go to the other instances where we are needlessly stringifying for the logger and just pass in the raw object instead (where applicable, like i've done for the goodbye log here).

@dapplion

mpetrunic (approved):

dapplion (approved):
Looks great!

3xtr4t3rr3str14l:
Actually, looks like there's a lot of spots where we are stringifying data to the logger, but it looks like it's because we want it formatted for the sake of the readability of the log (for example: https://github.com/ChainSafe/lodestar/blob/379fa22aa833d13058822baad67b7e21065b96cc/packages/lodestar/src/network/network.ts#L184).  In the interest of keeping this PR simple and since making a judgement call about the wording/format of every log in the lodestar package is probably out of the scope of this PR, I'm going to call this done for now.

wemeetagain (approved):

dapplion:
> Actually, looks like there's a lot of spots where we are stringifying data to the logger, but it looks like it's because we want it formatted for the sake of the readability of the log (for example:
> 
> https://github.com/ChainSafe/lodestar/blob/379fa22aa833d13058822baad67b7e21065b96cc/packages/lodestar/src/network/network.ts#L184
> 
> ). In the interest of keeping this PR simple and since making a judgement call about the wording/format of every log in the lodestar package is probably out of the scope of this PR, I'm going to call this done for now.

Sounds good, it's best to open a new issue tracker + PR to make all other log statements JSON logger friendly

PR #1753, opened 2020-11-12
set source mapping in tsconfig

### LINKED ISSUE #1708: VSCode to reference package ts src instead of lib
VSCode has a great functionality that with SHIFT+F12 takes you directly to the source code of whatever function you want. However, for the monorepo packages, it takes you to the built `lib/*.d.ts` files which is not helpful at all. There must exist some configuration to be done in the tsconfig.json so the actual source code `src/*.ts` is referenced instead

3xtr4t3rr3str14l:
resolves #1708

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,7 +7,7 @@
     "lib": ["es2020", "esnext.bigint", "es2020.string", "es2020.symbol.wellknown"],
     // Export declaration files '.d.ts' to /lib
     "declaration": true,
-    // "declarationMap": true,                /* Generates a sourcemap for each corresponding '.d.ts' file. */
+    "declarationMap": true,                /* Generates a sourcemap for each corresponding '.d.ts' file. */
     // "sourceMap": true,                     /* Generates corresponding '.map' file. */
 
     /* Strict Type-Checking Options */


3xtr4t3rr3str14l:
oh wait, this works when you go to source of an input, but not when you try to go to source when it's referenced outside of the import.  currently looking for the option to catch all cases, will push that once i find it

dapplion (approved):
:+1: Perfect, tho this does not fix https://github.com/ChainSafe/lodestar/issues/1708
EDIT: it does fix https://github.com/ChainSafe/lodestar/issues/1708

3xtr4t3rr3str14l:
> +1 Perfect, tho this does not fix #1708

It doesn't for you?  For me, when I build from this branch and hit Ctrl+click on an imported function from a monorepo package (which I'm assuming is the equivalent for the desired behavior from Shift+F12), it brings me to `src/` instead of `lib/` of that function in that monorepo package.  Is there something I'm missing in the requirements of this issue?

dapplion:
> > +1 Perfect, tho this does not fix #1708
> 
> It doesn't for you? For me, when I build from this branch and hit Ctrl+click on an imported function from a monorepo package (which I'm assuming is the equivalent for the desired behavior from Shift+F12), it brings me to `src/` instead of `lib/` of that function in that monorepo package. Is there something I'm missing in the requirements of this issue?

Tried again and worked perfectly! Sorry for not checking twice. Reverted the PR description change

wemeetagain (approved):

PR #1755, opened 2020-11-13
expand default object case for lodestar-utils toJson

### LINKED ISSUE #1754: ERR_UNKNOWN_PARENT not logging human-readable format
**Describe the bug**

i saw this log in both sync and gossip validation:
```2020-11-10 20:46:20 [NETWORK]          warn: Ignoring gossip block code=ERR_PARENT_UNKNOWN, parentRoot={"0":22,"1":178,"2":226,"3":54,"4":42,"5":160,"6":222,"7":100,"8":237,"9":34,"10":246,"11":200,"12":125,"13":159,"14":131,"15":135,"16":97,"17":156,"18":224,"19":189,"20":212,"21":248,"22":26,"23":96,"24":225,"25":13,"26":41,"27":65,"28":23,"29":149,"30":219,"31":138}, blockRoot=0xe2c7466a69829041405cf220f58125307535790dd4d7cca6f34c58720a6034d6, blockSlot=709731```

**Expected behavior**

print `parentRoot` as a hex string

### PR DESCRIPTION
**Pull request by 3xtr4t3rr3str14l**
resolves #1754 by expanding the default object case for the lodestar-utlis toJson

### DIFF
--- a/packages/lodestar-utils/src/json.ts
+++ b/packages/lodestar-utils/src/json.ts
@@ -1,16 +1,15 @@
 import {Json, toHexString} from "@chainsafe/ssz";
 import {LodestarError} from "./errors";
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-function convertArrayBufferViews(copyArg: any): void {
-  Object.entries(copyArg).forEach(([k, v]) => {
-    if (ArrayBuffer.isView(copyArg[k])) copyArg[k] = toJson(v);
+function convertArrayBufferViews(arg: {[key: string]: Json}): Json {
+  const argClone = {...arg};
+  Object.entries(argClone).forEach(([k, v]) => {
+    if (ArrayBuffer.isView(argClone[k])) argClone[k] = toJson(v);
   });
+  return argClone;
 }
 
 export function toJson(arg: unknown): Json {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  let copyArg: any;
   switch (typeof arg) {
     case "bigint":
     case "symbol":
@@ -21,15 +20,11 @@ export function toJson(arg: unknown): Json {
       if (arg === null) return "null";
       if (arg instanceof Uint8Array) return toHexString(arg);
       if (arg instanceof LodestarError) {
-        const copyArg = arg.toObject();
-        convertArrayBufferViews(copyArg);
-        return toJson(copyArg);
+        return toJson(convertArrayBufferViews(arg.toObject()));
       }
       if (arg instanceof Error) return toJson(errorToObject(arg));
       if (arg instanceof Array) return arg as Json;
-      copyArg = Object.assign({}, arg);
-      convertArrayBufferViews(copyArg);
-      return copyArg as Json;
+      return convertArrayBufferViews(Object.assign({}, arg));
 
     // Already valid JSON
     case "number":
--- a/packages/lodestar-utils/src/json.ts
+++ b/packages/lodestar-utils/src/json.ts
@@ -1,15 +1,10 @@
 import {Json, toHexString} from "@chainsafe/ssz";
 import {LodestarError} from "./errors";
+import {mapValues} from "./objects";
 
-function convertArrayBufferViews(arg: {[key: string]: Json}): Json {
-  const argClone = {...arg};
-  Object.entries(argClone).forEach(([k, v]) => {
-    if (ArrayBuffer.isView(argClone[k])) argClone[k] = toJson(v);
-  });
-  return argClone;
-}
+export const CIRCULAR_REFERENCE_TAG = "CIRCULAR_REFERENCE";
 
-export function toJson(arg: unknown): Json {
+export function toJson(arg: unknown, refs = new WeakMap()): Json {
   switch (typeof arg) {
     case "bigint":
     case "symbol":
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packages/lodestar/src/network/gossip/validator.ts [hunk 1]
```diff
@@ -69,17 +69,17 @@ export class GossipMessageValidator implements IGossipMessageValidator {
         case BlockErrorCode.ERR_PROPOSAL_SIGNATURE_INVALID:
         case BlockErrorCode.ERR_INCORRECT_PROPOSER:
         case BlockErrorCode.ERR_KNOWN_BAD_BLOCK:
-          this.logger.warn("Rejecting gossip block", {...e.getMetadata(), ...logContext});
+          this.logger.warn("Rejecting gossip block", toJson({...e.getMetadata(), ...logContext}));
```
dapplion: You shouldn't have to call `toJson` on the logger context. It is already called here
https://github.com/ChainSafe/lodestar/blob/10e261546c4c054811e272ebcb12bf736ef627dd/packages/lodestar-utils/src/logger/format.ts#L84
Why is it necessary? Doing 
```ts
this.logger.warn("Rejecting gossip block", {...e.getMetadata(), ...logContext});
```
results in bad logs?

packages/lodestar-utils/src/json.ts
```diff
@@ -13,7 +16,8 @@ export function toJson(arg: unknown): Json {
       if (arg instanceof Uint8Array) return toHexString(arg);
       if (arg instanceof LodestarError) return toJson(arg.toObject());
       if (arg instanceof Error) return toJson(errorToObject(arg));
-      return arg as Json;
+      Object.entries(arg).forEach(([k, v]) => (copyArg[k] = toJson(v)));
+      return copyArg as Json;
```
dapplion: Warning! You are introducing recursion here, which has caused issues in the past. Object can have self-referencing properties which would create an infinite loop

dapplion (changes requested):

packages/lodestar-utils/src/json.ts
dapplion: Unit tests caught the recursion here https://github.com/ChainSafe/lodestar/pull/1755/checks?check_run_id=1393717117#step:8:149

dapplion:
To address the issue https://github.com/ChainSafe/lodestar/issues/1754 I suggest you add test cases to the formating unit tests toJson and toString
https://github.com/ChainSafe/lodestar/blob/10e261546c4c054811e272ebcb12bf736ef627dd/packages/lodestar-utils/test/unit/json.test.ts#L10
With data like the gossip module sends to the logger
```ts
{code: "ERR_PARENT_UNKNOWN", parentRoot: fromHexString("0x111111111111111111111111111111111...")}
```

packages/lodestar/src/network/gossip/validator.ts
3xtr4t3rr3str14l: Yeah, in the case of an error having some metadata that is not immediately printable (such as a Uint8Array), it doesn't print it correctly and must be converted to hex string first.

3xtr4t3rr3str14l: However, I was trying to call it here before I found out that toJson had a bug, so maybe I don't need to call it anymore.  Will double check.

3xtr4t3rr3str14l: Yup, don't need to call it.  Just pushed up a change for it.

packages/lodestar-utils/src/json.ts
```diff
@@ -11,9 +20,16 @@ export function toJson(arg: unknown): Json {
...
-      if (arg instanceof LodestarError) return toJson(arg.toObject());
+      if (arg instanceof LodestarError) {
+        const copyArg = arg.toObject();
+        convertArrayBufferViews(copyArg);
+        return toJson(copyArg);
+      }
       if (arg instanceof Error) return toJson(errorToObject(arg));
-      return arg as Json;
+      if (arg instanceof Array) return arg as Json;
+      copyArg = Object.assign({}, arg);
```
dapplion: It's hard for me to follow what's going on here. Could you describe the purpose of these statements and the global untyped `copyArg` variable?

3xtr4t3rr3str14l: I didn't want to have to modify the argument of the function (for the sake of maintaining a "functional" programming style, as it has been recommended to me to use for this codebase in the past), so I made a copy of it.

packages/lodestar-utils/test/unit/json.test.ts
```diff
@@ -135,6 +165,14 @@ describe("Json helper", () => {
       {id: "object of basic types", json: {a: 1, b: 2}, output: "a=1, b=2"},
       // eslint-disable-next-line quotes
       {id: "object of objects", json: {a: {b: 1}}, output: `a={"b":1}`},
+      {
+        id: "error metadata",
+        json: toJson({
```
dapplion: This is part of the test data, you should not use here the function to be tested

packages/lodestar-utils/src/json.ts
dapplion: Can you abstract this functionality in a way that you don't need to mutate arguments out of the scope of the if statement?

packages/lodestar-utils/src/json.ts
```diff
@@ -1,7 +1,16 @@
 import {Json, toHexString} from "@chainsafe/ssz";
 import {LodestarError} from "./errors";
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function convertArrayBufferViews(copyArg: any): void {
```
dapplion: Could you make this function not mutate `copyArg` and also type it?

dapplion (dismissed):

packages/lodestar-utils/src/json.ts
3xtr4t3rr3str14l: Yes, my latest commit (0cc57106367b0a0158f4a944a25d195fce00c3d9) should address this and your other comments.  Let me know what you think.

wemeetagain (approved):

PR #1756, opened 2020-11-13
Update block production api

mpetrunic:
- added new block production endpoint
- validator changed to use new block production and block publish endopoints


packages/lodestar-validator/src/api/impl/rest/beacon/beacon.ts
```diff
@@ -3,11 +3,13 @@ import {BeaconBlock, BeaconState, BLSPubkey, Genesis, ValidatorResponse} from "@
...
-import {IBeaconApi, IBeaconStateApi} from "../../../interface/beacon";
+import {IBeaconApi, IBeaconBlocksApi, IBeaconStateApi} from "../../../interface/beacon";
+import {RestBeaconBlocksApi} from "./blocks";
 import {RestBeaconStateApi} from "./state";
 
 export class RestBeaconApi implements IBeaconApi {
   public readonly state: IBeaconStateApi;
+  public readonly blocks: IBeaconBlocksApi;
 
   private readonly client: HttpClient;
```
twoeths: seems like we don't use this atm. Do we want to reserve it for any future use?

packages/lodestar-validator/src/api/impl/rest/validator/validator.ts
```diff
@@ -66,21 +65,14 @@ export class RestValidatorApi implements IValidatorApi {
...
-    proposerPubkey: BLSPubkey,
-    randaoReveal: Bytes96,
-    graffiti: string
-  ): Promise<BeaconBlock> {
-    const url = "/block";
+  public async produceBlock(slot: Slot, randaoReveal: Bytes96, graffiti: string): Promise<BeaconBlock> {
+    const url = `/blocks/${slot}`;
     const query = {
-      slot: slot,
-      proposer_pubkey: toHexString(proposerPubkey),
```
twoeths, line 78: so since we don't need to transfer pubkey to produce a block anymore, what's the plan to avoid DOS attack? I can write a script asking our beacon node to produce blocks multiple times to keep our node busy.

packages/lodestar-validator/src/api/impl/rest/beacon/beacon.ts
mpetrunic: nah, I'm gonna remove it

packages/lodestar-validator/src/api/impl/rest/validator/validator.ts
mpetrunic: how was pubkey preventing dos?

You can open issue to add rate limiting to certain endpoints if you wish

twoeths: hah so an attacker can easily find a pubkey and still can do the DOS attack.
digging into this, I think maybe we should validate randao_reveal when we produce a block?

it's not related to this PR anyway, just general concern.

mpetrunic: I mean producing block isn't performance intensive, limiting it to 1 req per second or something could suffice. While randao_revel verification is probably more cpu intenssive :smile:

twoeths: oh yeah that's why I don't see other clients are validating that

I'm fine with leaving it as is for now bc if we limit 1 req per second, an attacker can disturb our node and the real validators never has a chance to produce blocks. (if later on if we want to do it then probably we can base on client ip to ban attackers)

mpetrunic: sure, fyi rate limits are per ip not overall

twoeths (dismissed):

wemeetagain:
file conflict, other than that LGTM

wemeetagain (dismissed):

wemeetagain (approved):

PR #1757, opened 2020-11-13
Use bls.verify across codebase for consistency

dapplion:


twoeths:
what's the rationale of this @dapplion ?

mpetrunic:
> what's the rationale of this @dapplion ?

I'm also curious, because named imports should always be preferred over default ones.

dapplion:
As discussed in the call, global config changes to BLS can be achieved in the init call, so this is not necessary

Issue #1738, opened 2020-11-08
OOM due to ArchiveBlocks having to process too many blocks


twoeths:
**Describe the bug**
I increased heap memory to 3GB and was able to reach the finalized state after long infinality period, however it crashed due to ArchiveBlocks task:
```
2020-11-08 04:21:52 [CHORES]           info: @@@ archiveBlocks number of canonical blocks 77614
2020-11-08 04:21:52 [CHORES]           info: @@@ archiveBlocks getting canonical blocks from db
2020-11-08 04:21:52 [CHAIN]            info: Checkpoint justified epoch=20559, root=0x135a6b7134b11ff838848dfe37a2daafccaeda333de83cd50981d4635dc59b98
2020-11-08 04:21:52 [CHAIN]            info: Checkpoint finalized epoch=20557, root=0x3fabb30bf04bc51c87a232ba81919821547c5f822f3305665df127577a234f87
```
We should be able to separate the job into smaller batches/chunks instead of loading 77614 blocks from db and process all of them at the same time.
**Expected behavior**
We run ArchiveBlocks task successfully

**Steps to Reproduce**
```
node --max-old-space-size=3072 --inspect  lib/index.js beacon --testnet medalla --root-dir .medalla.weak
```

twoeths:
even after I process blocks in chunks and put finalized checkpoint process in queue, I still see we have a lot of canonical blocks in forkchoice
```
2020-11-11 03:21:19 [CHORES]           ^[[32minfo^[[39m: @@@ archiveBlocks number of canonical blocks 77614
 2020-11-11 05:14:39 [CHORES]           ^[[32minfo^[[39m: Archive Blocks  - duration=6802173ms
185496 2020-11-11 05:14:39 [CHORES]           ^[[32minfo^[[39m: Archiving of finalized blocks complete. totalArchived=77614, finalizedEpoch=20557
2020-11-11 05:14:47 [CHORES]           ^[[32minfo^[[39m: @@@ archiveBlocks number of canonical blocks 79261
 2020-11-11 05:57:54 [CHORES]           ^[[32minfo^[[39m: Archive Blocks  - duration=2595480ms
186981 2020-11-11 05:57:54 [CHORES]           ^[[32minfo^[[39m: Archiving of finalized blocks complete. totalArchived=79261, finalizedEpoch=20638
[...]

PR #1759, opened 2020-11-14
DB API: Implement getBinary(), putBinary(), batchPutBinary() to improve performance of blockArchive task

twoeths:
part of #1738 
+ During the big unfinality time of medalla, loading `block` db and persisting to `blockArchive` takes more or less 1h so we need to improve it
+ I see that when loading from `block` db, we don't have to convert to `SignedBeaconBlock`, just use the raw Buffer data and store it to `blockArchive` db. This makes it faster and takes less memory (so we can tweak chunk size bigger too)
+ Implement `getBinary()`, `putBinary()` and `batchPutBinary()` accordingly
+ It's 10 times faster as tested in e2e

### DIFF
--- a/packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
+++ b/packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
@@ -4,9 +4,14 @@ import {bytesToInt, intToBytes} from "@chainsafe/lodestar-utils";
 import {IDatabaseController, Bucket, Repository, IFilterOptions, IKeyValue, encodeKey} from "@chainsafe/lodestar-db";
 import {ArrayLike} from "@chainsafe/ssz";
 import all from "it-all";
+import {IBlockSummary} from "@chainsafe/lodestar-fork-choice";
 
 export interface IBlockFilterOptions extends IFilterOptions<Slot> {
   step?: number;
+}
+
+export interface IKeyValueSummary<K, V, S> extends IKeyValue<K, V> {
+  summary: S;
 }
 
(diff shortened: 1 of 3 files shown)


mpetrunic (commented):
Hmm I couldn't find code where you are using getBinary/putBinary? I guess you didn't change archive blocks?

twoeths:
> Hmm I couldn't find code where you are using getBinary/putBinary? I guess you didn't change archive blocks?

@mpetrunic I'll do it in my main PR of #1738 , want to make smaller diff to make it easier to review. The usage is the same to the e2e test I make in this PR with block summaries available in forkchoice.

packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts
```diff
@@ -18,14 +23,36 @@ export class BlockArchiveRepository extends Repository<Slot, SignedBeaconBlock>
...
+      }),
+      Array.from(items).map((item) => {
+        const slot = item.value.message.slot;
+        const parentRoot = item.value.message.parentRoot;
+        return this.storeParentRootIndex(slot, parentRoot);
+      }),
+    ]);
+  }
+
+  public async batchPutBinary(items: ArrayLike<IKeyValueSummary<Slot, Buffer, IBlockSummary>>): Promise<void> {
```
twoeths, line 51: this is the one to be used in our `archiveBlocks` task later (block summaries are available in forkchoice)

wemeetagain (approved):

PR #1763, opened 2020-11-17
add metrics.listenAddr option

### LINKED ISSUE #1294: Configure Metrics listening address via CLI
There's no way to configure the metrics listening address, it simply defaults (in `http.listen` to 0.0.0.0.

Expose listening address via CLI flag + BeaconNodeOptions

**Additional context**
See https://discordapp.com/channels/593655374469660673/593655641445367808/739555083166941214

3xtr4t3rr3str14l:
resolves #1294

### DIFF
--- a/packages/lodestar-cli/test/unit/options/beaconNodeOptions.test.ts
+++ b/packages/lodestar-cli/test/unit/options/beaconNodeOptions.test.ts
@@ -26,6 +26,7 @@ describe("options / beaconNodeOptions", () => {
       "metrics.pushGateway": true,
       "metrics.serverPort": 8765,
       "metrics.timeout": 5000,
+      "metrics.listenAddr": "0.0.0.0",
 
       "network.discv5.enabled": true,
       "network.discv5.bindAddr": "addr",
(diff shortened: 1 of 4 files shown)


packages/lodestar/src/metrics/server/http.ts
```diff
@@ -28,7 +28,7 @@ export class HttpMetricsServer implements IMetricsServer {
       const listen = this.http.listen.bind(this.http);
       const port = this.opts.serverPort;
       return new Promise((resolve, reject) => {
-        listen(port).once("listening", resolve).once("error", reject);
+        listen(port, this.opts.listenAddr).once("listening", resolve).once("error", reject);
```
dapplion: For consistency assign both port and listenAddr to a const variable before hand or none

packages/lodestar-cli/test/unit/options/beaconNodeOptions.test.ts
```diff
@@ -64,6 +65,7 @@ describe("options / beaconNodeOptions", () => {
         pushGateway: true,
         serverPort: 8765,
         timeout: 5000,
+        listenAddr: "0.0.0.0",
```
mpetrunic, line 68: I feel like 127.0.0.1 would be more sensible default value

3xtr4t3rr3str14l: I just used 0.0.0.0 because that's what the node http server uses by default.

wemeetagain (approved):

PR #1764, opened 2020-11-18
Implement initiateMultipleValidatorExits

### LINKED ISSUE #1761: Improve validator exit process
**Describe the bug**
It takes so much time to process blocks with validator exits, we need to improve it

Screen Shot 2020-11-16 at 19 35 02

it showed that processing this method may take up to 6s

twoeths:
resolves #1761 

+ Keep `initiateValidatorExit` for backward reference, it's the same to the spec
+ Implement `initiateMultipleValidatorExits` which loop through validator array inside state only 1 time
+ Spec test goes through `initiateMultipleValidatorExits`

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/fast/block/initiateValidatorExit.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/block/initiateValidatorExit.ts
@@ -42,7 +42,7 @@ export function initiateMultipleValidatorExits(
   state: BeaconState,
   indexes: ValidatorIndex[] = []
 ): void {
-  if (!indexes.length) return;
+  if (!indexes || indexes.length === 0) return;
   const config = epochCtx.config;
   // compute exit queue epoch
   const validatorExitEpochs = readOnlyMap(state.validators, (v) => v.exitEpoch);
@@ -51,10 +51,10 @@ export function initiateMultipleValidatorExits(
   const currentEpoch = epochCtx.currentShuffling.epoch;
   const activationExitEpoch = computeActivationExitEpoch(config, currentEpoch);
   for (const index of indexes) {
-    // return if validator already initiated exit
+    // continue if validator already initiated exit
     const validator = state.validators[index];
     if (validator.exitEpoch !== FAR_FUTURE_EPOCH) {
-      return;
+      continue;
     }
     let exitQueueEpoch = Math.max(...exitEpochs, activationExitEpoch);
     const exitQueueChurn = validatorExitEpochs.filter((exitEpoch) => exitEpoch === exitQueueEpoch).length;
--- a/packages/lodestar-beacon-state-transition/src/fast/block/initiateValidatorExit.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/block/initiateValidatorExit.ts
@@ -5,6 +5,9 @@ import {FAR_FUTURE_EPOCH} from "../../constants";
 import {computeActivationExitEpoch, getChurnLimit} from "../../util";
 import {EpochContext} from "../util";
 
+/**
+ * Old implementation following the spec, mainly for the spec test.
+ */
 export function initiateValidatorExit(epochCtx: EpochContext, state: BeaconState, index: ValidatorIndex): void {
   const config = epochCtx.config;
   // return if validator already initiated exit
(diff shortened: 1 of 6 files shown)


packages/lodestar-beacon-state-transition/src/fast/block/processOperations.ts
```diff
@@ -41,10 +42,12 @@ export function processOperations(
...
     [body.attestations, processAttestation],
     [body.deposits, processDeposit],
-    [body.voluntaryExits, processVoluntaryExit],
+    [body.voluntaryExits, verifyVoluntaryExit],
   ] as [List<Operation>, OperationFunction][]).forEach(([operations, processOp]) => {
     readOnlyForEach(operations, (op) => {
       processOp(epochCtx, state, op, verifySignatures);
     });
   });
+  const indexes = readOnlyMap(body.voluntaryExits, (voluntaryExit) => voluntaryExit.message.validatorIndex);
```
dapplion: What about using
```ts
processVoluntaryExits(body.voluntaryExits, epochCtx, state)
```
to not spread the logic in multiple files

packages/lodestar-beacon-state-transition/src/fast/block/processOperations.ts
```diff
@@ -41,10 +42,12 @@ export function processOperations(
     [body.attesterSlashings, processAttesterSlashing],
     [body.attestations, processAttestation],
     [body.deposits, processDeposit],
-    [body.voluntaryExits, processVoluntaryExit],
+    [body.voluntaryExits, verifyVoluntaryExit],
```
dapplion: I wonder if we can modify this entire statement to
```js
processAttesterSlashings(...)
processAttestations(...)
processDeposits(...)
```
And look for similar optimizations in the other opts

packages/lodestar-beacon-state-transition/src/fast/block/initiateValidatorExit.ts
```diff
@@ -29,3 +32,40 @@ export function initiateValidatorExit(epochCtx: EpochContext, state: BeaconState
...
+export function initiateMultipleValidatorExits(
+  epochCtx: EpochContext,
+  state: BeaconState,
+  indexes: ValidatorIndex[] = []
+): void {
+  if (!indexes || indexes.length === 0) return;
+  const config = epochCtx.config;
+  // compute exit queue epoch
+  const validatorExitEpochs = readOnlyMap(state.validators, (v) => v.exitEpoch);
+  const exitEpochs = validatorExitEpochs.filter((exitEpoch) => exitEpoch !== FAR_FUTURE_EPOCH);
```
mpetrunic: would it be better to filter first then map?

dapplion: Definitely! Also, why not do both in one loop?

mpetrunic: Can't in a single loop unless we add filerAndmapReadOnly iterator to ssz. I hope we have readOnlyFilter if not, this mihgt be useless comment

dapplion: Can you do this?
```ts
const exitEpochs = []
for (const v of state.validators) {
  if (v.exitEpoch !== FAR_FUTURE_EPOCH) {
    exitEpochs.push(v.exitEpoch)
  }
}
```

mpetrunic: Keep in mind that state is actually TreeBacked so if I'm not mistaken, your code would actually deserialize and allocate new validator objects. But your idea could be done with readOnlyForEach. Ref: https://github.com/ChainSafe/ssz/blob/master/src/backings/tree/array.ts#L141

twoeths: `readOnlyForEach` would be good, will do it

packages/lodestar-beacon-state-transition/src/fast/block/processOperations.ts
twoeths: the `initiateValidatorExit()` was called in `slashValidator.ts` also but we couldn't optimize the same way to this because it's required to process one by one there in order to use validator `withdrawableEpoch` correctly

other than that I don't see any similar improvement can be made so I keep other functions as is.

twoeths:
closing this in favor of #1816   as I think the bigger issue is we loop through `validators` array in other flows too

PR #1769, opened 2020-11-19
readable goodbye reason codes

3xtr4t3rr3str14l:
I got a list of known goodbye reason codes from someone at Prysm (or at least the ones that Prysm knows about: https://github.com/prysmaticlabs/prysm/blob/master/beacon-chain/p2p/types/rpc_goodbye_codes.go#L8) and added them to our codebase to where we can print a human-readable form of the codes when we receive them.

### DIFF
--- a/packages/lodestar/src/sync/reqResp/reqResp.ts
+++ b/packages/lodestar/src/sync/reqResp/reqResp.ts
@@ -43,6 +43,21 @@ enum GoodByeReasonCode {
   ERROR = 3,
 }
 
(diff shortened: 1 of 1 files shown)


mpetrunic (approved):

mpetrunic:
same unit test failing in both runs so I doubt it's just CI error

twoeths (approved):
it passed when I run them again. We should create an issue if we see it next time.

wemeetagain (approved):

PR #1783, opened 2020-11-21
Inline config files | Bump spec tests to v1.0.0

### LINKED ISSUE #1787: Use tar archive instead of npm for spec tests
Problem:
We used npm to distribute spec tests, but spec tests reached size limit of npm registry.

Solution:
Extend lodestar-spec-test-util to download spec tests archive from https://github.com/ethereum/eth2.0-spec-tests/releases/download/v1.0.0/general.tar.gz and resolve file path relative to extracted spec tests.

Method that should be extended:
https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar-spec-test-util/src/single.ts#L61

My suggestion is to add spec tests version as required param and extraction location as optional (defaulting to node_modules/.cache/@chainsafe/lodestar-spec-test-utils).
Upon executing describeDirectorySpecTest, we should check whether directory <spec_tests_location>/<version> directory exists.
If it does resolve spec test path against it.
If it doesn't, download tar archives for given version extract them all into <spec_tests_location>/<version> location

dapplion:
- Decouple the dependency with babel to inline config .yaml files
- Ensure with a test that our config yaml is synced with the remote config from `ethereum/eth2.0-specs`
- Bump spec tests to v1.0.0
- Download spec tests from GitHub instead of NPM

Fixes https://github.com/ChainSafe/lodestar/issues/1787


packages/lodestar-params/package.json
```diff
@@ -46,6 +46,8 @@
     "js-yaml": "^3.13.1"
   },
   "devDependencies": {
-    "babel-plugin-inline-import": "^3.0.0"
+    "@types/node-fetch": "^2.5.7",
+    "babel-plugin-inline-import": "^3.0.0",
```
mpetrunic: remove

packages/lodestar-params/package.json
```diff
@@ -46,6 +46,8 @@
...
+    "node-fetch": "^2.6.1"
```
mpetrunic: maybe better to use axios since we are already using it in lodestar-validator?

packages/lodestar-params/src/utils.ts
```diff
@@ -13,7 +13,13 @@ export function createIBeaconParams(input: Record<string, unknown>): Partial<IBe
...
 }
 
-export const schema = new Schema({
+export function loadConfigYaml(configYaml: string): Record<string, unknown> {
+  return load(configYaml, {
+    schema: schemaParseAllNumsAsStrings,
+  });
+}
+
+export const schemaParseAllNumsAsStrings = new Schema({
```
mpetrunic: I would keep schema name as it was because we might gonna need to add more rules for different types

packages/lodestar-params/test/ensure-config-is-synced.test.ts
```diff
@@ -0,0 +1,27 @@
+import {expect} from "chai";
+import fetch from "node-fetch";
+import {loadConfigYaml} from "../src/utils";
+import {yaml as mainnetYaml} from "../src/presets/mainnet";
+import {yaml as minimalYaml} from "../src/presets/minimal";
+
+const commit = "1005b5baf8c043982b9a72282996311f9c194569";
```
mpetrunic: we should probably put 1.0.0. tag here

mpetrunic (changes requested):

packages/lodestar-params/package.json
dapplion: For blst-ts I was forced to add an HTTP client as dependency to fetch the pre-built bindings. I researched what was the least invasive HTTP client and came across this comparison https://github.com/request/request/issues/3143.
- node-fetch has 0 dependencies and is tiny
- axios has 1 dependency and is also quite small

That was the reasoning for picking node-fetch but I would love to know your opinion on which to use given this criteria

mpetrunic: in blst I would go with pure nodejs and just pipe to file:
```js
const http = require('http');
const fs = require('fs');

const file = fs.createWriteStream("binding.node");
const request = http.get("http://binding.node", function(response) {
  response.pipe(file);
});
```

I would go with axios just because we already have it so we can avoid installing multiple dependencies that are doing same thing. Also it's quite configurable and works on web

dapplion: I did that initially but it broke because this code does not follow redirect and Github assets are not in the original URL

dapplion: I've replaced node-fetch with axios

mpetrunic: It's easy to fix, works with:
```typescript
const https = require('https');
const fs = require('fs');

const file = fs.createWriteStream("./binding.node");
const request = https.get("<release url>", (response) => {
  if(response.statusCode == 302) {
     https.get(response.headers.location, (res) => {
        res.pipe(file);
     })
  } else {
     response.pipe(file);
  }
});


```

mpetrunic (dismissed):

mpetrunic:
we need 1.0.0. tests, gonna publish that on monday

dapplion:
@wemeetagain This test is failing, could you advise what to do?
```
@chainsafe/lodestar-types:   1) should calculate correct minSize and maxSize:
@chainsafe/lodestar-types:       AssertionError: expected 141837543039377 to equal 141837542965649
@chainsafe/lodestar-types:       + expected - actual
@chainsafe/lodestar-types:       -141837543039377
@chainsafe/lodestar-types:       +141837542965649
```
https://github.com/ChainSafe/lodestar/pull/1783/checks?check_run_id=1435942310#step:8:141

wemeetagain:
Did any of the params that control the max-lengths of lists change? That would change the number in the tests

dapplion:
> Did any of the params that control the max-lengths of lists change? That would change the number in the tests

You can check the diff here https://github.com/ChainSafe/lodestar/pull/1783/commits/a2e132c8b69a69c047df05506d3cd0ee0be77459#diff-b932715bd65532d39074e134c4331abcac00cadd47785b015d388740dede3125
`EPOCHS_PER_ETH1_VOTING_PERIOD` was doubled

twoeths:
the BeaconState minSize maxSize comes from https://gist.github.com/protolambda/db75c7faa1e94f2464787a480e5d613e and it goes with the old params and that link is still part of the spec.

If we just tweak the minSize/maxSize to make it passed, it does not have a lot of meaning.
If we want to keep it to avoid any regressions, we can do something like:
```
import {params} from "@chainsafe/lodestar-params/lib/presets/mainnet";
import {expect} from "chai";
import {createIBeaconSSZTypes} from "../../src/ssz/generators";

it.only("should calculate correct minSize and maxSize", () => {
  // TODO: try to get param that goes with the test result
  const testParams = {...params, EPOCHS_PER_ETH1_VOTING_PERIOD: 32};
  const types = createIBeaconSSZTypes(testParams);
  const minSize = types.BeaconState.minSize();
  const maxSize = types.BeaconState.maxSize();
  // https://gist.github.com/protolambda/db75c7faa1e94f2464787a480e5d613e
  expect(minSize).to.be.equal(2687377);
  expect(maxSize).to.be.equal(141837542965649);
});
```

wemeetagain:
Seems to be breaking bc we need 1.0 spec tests
@mpetrunic  pls commit on top of this branch when ready

mpetrunic:
> Seems to be breaking bc we need 1.0 spec tests
> @mpetrunic  pls commit on top of this branch when ready

I already written in chat, tests grew too large, we cannot upload them anymore 😓

Ive made issue.
https://github.com/ChainSafe/lodestar/issues/1787

wemeetagain (approved):

PR #1786, opened 2020-11-22
Initial Sync: do not import orphaned blocks

### LINKED ISSUE #1785: Initial sync getting stuck at orphaned block
**Describe the bug**
Node keeps trying the same range:

```
2020-11-22 08:33:20 [SYNC]             info: Fetching blocks for 606921...606984 slot range
2020-11-22 08:33:22 [SYNC]             info: Imported blocks for slots: 606921,606923,606924,606925,606926,606927,606928,606929,606931,606932,606933,606934,606936,606938,606939,606941,606942,606947,606949,606950,606951,606954,606956,606958,606959,606960,606962,606963,606964,606965,606972,606973,606977,606979,606980,606981,606982
2020-11-22 08:33:22 [SYNC]             warn: Received block parent root doesn't match our head head=0x1664a3191404d5c5aafd7f24c22a417bc0b8628d009bed48ae0c9847b0198980, headSlot=606920, blockParent=0x6313aacdc01f649c765918506d90e4a8551077175c2dc7124860e1c6bca13e21, blockSlot=606921
2020-11-22 08:33:22 [SYNC]             warn: Failed to process range, retrying, lastSlot=606920 range={"start":606921,"end":606984}
2020-11-22 08:33:22 [SYNC]             info: Fetching blocks for 606921...606984 slot range
```

Note that last slot is 606920 and it's an orphaned block.

**Expected behavior**
Chain should ignore orphaned block

**Steps to Reproduce**
Sync Medalla

Need to have the same check like in https://github.com/ChainSafe/lodestar/pull/1736

twoeths:
resolves #1785 
+ avoid orphaned block
+ avoid req/resp timeout which cause the sync stall
+ set `minPeers=3`

### DIFF
--- a/packages/lodestar/src/sync/options.ts
+++ b/packages/lodestar/src/sync/options.ts
@@ -9,7 +9,7 @@ export interface ISyncOptions {
 }
 
 export const defaultSyncOptions: Required<ISyncOptions> = {
-  minPeers: 4,
+  minPeers: 3,
   //2 epochs
   maxSlotImport: 64,
   blockPerChunk: 64,
--- a/packages/lodestar/src/sync/options.ts
+++ b/packages/lodestar/src/sync/options.ts
@@ -9,7 +9,7 @@ export interface ISyncOptions {
 }
 
 export const defaultSyncOptions: Required<ISyncOptions> = {
-  minPeers: 2,
+  minPeers: 4,
   //2 epochs
   maxSlotImport: 64,
   blockPerChunk: 64,
--- a/packages/lodestar/src/sync/utils/sync.ts
+++ b/packages/lodestar/src/sync/utils/sync.ts
@@ -105,7 +105,17 @@ export function fetchBlockChunks(
           return;
         }
         try {
-          yield await getBlockRange(logger, reqResp, peers, slotRange);
+          // a work around of timeout issue that cause our sync stall
+          let timer: NodeJS.Timeout | null = null;
+          yield (await Promise.race([
+            getBlockRange(logger, reqResp, peers, slotRange),
+            new Promise((_, reject) => {
+              timer = setTimeout(() => {
+                reject(new Error("beacon_blocks_by_range timeout"));
+              }, 3 * 60 * 1000); // 3 minutes
+            }),
+          ])) as SignedBeaconBlock[] | null;
+          if (timer) clearTimeout(timer);
         } catch (e) {
           logger.debug("Failed to get block range " + JSON.stringify(slotRange) + ". Error: " + e.message);
           yield null;
--- a/packages/lodestar/src/sync/utils/sync.ts
+++ b/packages/lodestar/src/sync/utils/sync.ts
@@ -16,6 +16,9 @@ import {IPeerMetadataStore} from "../../network/peers/interface";
 import {getSyncPeers} from "./peers";
 import {DEFAULT_RPC_SCORE} from "../../network/peers";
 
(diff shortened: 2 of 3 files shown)


packages/lodestar/src/sync/options.ts
```diff
@@ -9,7 +9,7 @@ export interface ISyncOptions {
 }
 
 export const defaultSyncOptions: Required<ISyncOptions> = {
-  minPeers: 2,
+  minPeers: 4,
```
3xtr4t3rr3str14l: just looking for a little bit of background here....where did we get our initial number of 2 for minPeers and how do we know that increasing it to 4 won't have any adverse side-effects?

packages/lodestar/src/sync/utils/sync.ts
```diff
@@ -105,7 +105,17 @@ export function fetchBlockChunks(
...
         try {
-          yield await getBlockRange(logger, reqResp, peers, slotRange);
+          // a work around of timeout issue that cause our sync stall
+          let timer: NodeJS.Timeout | null = null;
+          yield (await Promise.race([
+            getBlockRange(logger, reqResp, peers, slotRange),
+            new Promise((_, reject) => {
+              timer = setTimeout(() => {
+                reject(new Error("beacon_blocks_by_range timeout"));
+              }, 3 * 60 * 1000); // 3 minutes
```
3xtr4t3rr3str14l: where does the number of 3 minutes come from?

twoeths: in initial sync, we divide 64 slot range to a couple of small range and query it to eligible peers i parallel. Normally, it's pretty fast (only 2s - 10s) as I can see but if a chunk is failed, it'll try the same chunk with another peer.

In the case of timeout and an error is not thrown, it'll never finish the call and initial sync get stalled, the 3 minutes param is the maximum time we can wait and that's my estimate in the case. This rarely happens but quite annoying.

packages/lodestar/src/sync/options.ts
twoeths: this is the minimal number of peers sync wait to start its process. If we start from scratch we can quickly get ~20 peers. As I remember, initially we had a hard time getting peers, that's quite it's quite low.

With the 2 peers minimal, the common finalized checkpoint may be too small if the found peers are syncing peer too (this may happen if we're unlucky). The consequence is we use Regular Sync instead of Initial Sync even there are a lot of finalized epochs ahead bc it bypass initial sync unintentionally.

https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/sync/initial/fast/index.ts#L78

Increasing to 4 peers help me fix the issue locally, we can even increase more. ~It has no side-effect.~

Edit: the side effect is we have to wait for a couple of seconds more to get more peers before we sync

packages/lodestar/src/sync/utils/sync.ts
mpetrunic: make it a constant

packages/lodestar/src/sync/options.ts
mpetrunic: what value does other clients have?

twoeths: it's 3 for Prysm, not sure for Lighthouse

3xtr4t3rr3str14l (approved):
LGTM

wemeetagain (approved):

PR #1788, opened 2020-11-23
Phase1 types and params

### LINKED ISSUE #1458: Phase1 Params + Types
If we want to prototype any phase1-related work, we will need the new phase1 params and types.
Ph1 params only add additional params.
Ph1 types, on the other hand, both extend existing types and add additional types.

Whatever solution we choose, we should be mindful of backwards compatibility.

One solution would be to namespace our types by phase:
```typescript
import {phase0, phase1} from "@chainsafe/lodestar-types";
const v: phase0.Validator = {...};
const vv: phase1.Validator = {...};
```
Every extended type would simply be a new type definition, and we would be able to retain all existing types. All untouched types can 

The config would be similarly namespaced:
```typescript
import {config} from "@chainsafe/lodestar-config";
config.types.phase1.Validator.toJson({...});
```

- Separate repos for different phases
- Rename types to include phase in the name eg: Phase1Validator or ValidatorPhase1

**Additional context**
See https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase1/beacon-chain.md

mpetrunic:
Does this look ok?

Should I try to adjust tests for phase 1 stuff? Do we have some other ideas how to test those type generators?

resolves #1458


wemeetagain:
LGTM :)

mpetrunic:
> LGTM :)

Didnt read pr description, eh? : 😁

mpetrunic:
@wemeetagain I manage to make type generation full type safe so it don't actually need those wierd unit tests since typescript checks whether we have ssz definition for every interface and if they match.

 Do you wan't me to change phase0 types in separate PR?

wemeetagain (approved):

PR #1794, opened 2020-11-25
Run type and code build tasks in parallel

mpetrunic:
Instead of running sequentially, we now run both babel for code generation and tsc for type generation in parallel.

I wish there was some faster type compiler :disappointed:


dapplion:
What's the real time difference to build with this PR vs current master?
How do tsc compile errors look like through the concurrently renderer? Does it look the same?

mpetrunic:
> What's the real time difference to build with this PR vs current master?

Small 10s I think, basically time that babel takes to compile everything. On every task we are waiting for type generation.

> How do tsc compile errors look like through the concurrently renderer? Does it look the same?

Yeah same, if any of commands fail, concurrently fails also

twoeths (approved):

wemeetagain (approved):

PR #1799, opened 2020-11-29
Performance tests for state transition functions

twoeths:
## Goal
+ This should give us an idea of the current performance of state transition functions
+ Data is generated from Medalla state/block 756416 and the performance is close to what I see in Medalla sync


packages/lodestar-beacon-state-transition/test/perf/util.ts
```diff
@@ -0,0 +1,143 @@
...
+      "0x8a5d2673c48f22f6ed19462efec35645db490df29eed2f56321dbe4a89b2463b0c902095a7ab74a2dc5b7f67edb1a19507ea3d4361d5af9cb0a524945c91638dfd6568841486813a2c45142659d6d9403f5081febb123a7931edbc248b9d0025"
+    );
+    // eth1Data, graffiti, attestations
+    signedBlock = config.types.SignedBeaconBlock.tree.createValue(block);
+    logger.info("Loaded block at slot", signedBlock.message.slot);
+  }
+  return signedBlock.clone();
+}
+
+const CURVE_ORDER = BigInt("52435875175126190479447740508185965837690552500527637822603658699938581184513");
```
twoeths: this part was copied from lodestar-validator, should we move it to lodestar-util?

dapplion:
Much needed effort to start tracking performance! Tho I'm not sure if asserting an upper bound is problematic: 
```
timeElapsed < 500
``` 
- Will the tests fail often if for some reason performance is locally slower? 
- Can we keep a historical progression of performance to evaluate and detect trends?

twoeths:
@dapplion this is just the 1st effort to improve state transition functions (if any), I'm not sure if we want to put it in CI or not.

The statistic comes from my own environment, and some test durations are different from time to time, maybe even up to 1s!

I just want to have a way to calculate performance of state transition functions without having to sync Pyrmont or Medalla.

dapplion:
> @dapplion this is just the 1st effort to improve state transition functions (if any), I'm not sure if we want to put it in CI or not.
> 
> The statistic comes from my own environment, and some test durations are different from time to time, maybe even up to 1s!
> 
> I just want to have a way to calculate performance of state transition functions without having to sync Pyrmont or Medalla.

That's very reasonable! Thanks for the clarification :heart:

packages/lodestar-beacon-state-transition/test/perf/util.ts
wemeetagain: Yeah :+1:

dapplion (dismissed):

wemeetagain (approved):

PR #1800, opened 2020-11-29
Update bls

wemeetagain:
todo: release bls and update package.json / yarn.lock

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/block/operations/deposit.ts
+++ b/packages/lodestar-beacon-state-transition/src/block/operations/deposit.ts
@@ -2,7 +2,7 @@
  * @module chain/stateTransition/block
  */
 
-import {verify} from "@chainsafe/bls";
+import bls from "@chainsafe/bls";
 import {BeaconState, Deposit, Validator} from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {DEPOSIT_CONTRACT_TREE_DEPTH, DomainType, FAR_FUTURE_EPOCH} from "../../constants";
@@ -38,7 +38,7 @@ export function processDeposit(config: IBeaconConfig, state: BeaconState, deposi
     // Verify the deposit signature (proof of possession)
     // Note: The deposit contract does not check signatures.
     // Note: Deposits are valid across forks, thus the deposit domain is retrieved directly from `computeDomain`.
-    if (!verify(pubkey.valueOf() as Uint8Array, signingRoot, deposit.data.signature.valueOf() as Uint8Array)) {
+    if (!bls.verify(pubkey.valueOf() as Uint8Array, signingRoot, deposit.data.signature.valueOf() as Uint8Array)) {
       return;
     }
     // Add validator and balance entries
--- a/packages/lodestar-beacon-state-transition/src/fast/util/block.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/util/block.ts
@@ -1,6 +1,5 @@
 import bls from "@chainsafe/bls";
 import {BeaconState, SignedBeaconBlock} from "@chainsafe/lodestar-types";
-
 import {computeSigningRoot, getDomain} from "../../util";
 import {DomainType} from "../../constants";
 import {EpochContext} from "../index";
--- a/packages/lodestar-beacon-state-transition/src/fast/util/block.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/util/block.ts
@@ -1,7 +1,8 @@
+import bls from "@chainsafe/bls";
 import {BeaconState, SignedBeaconBlock} from "@chainsafe/lodestar-types";
+
 import {computeSigningRoot, getDomain} from "../../util";
 import {DomainType} from "../../constants";
-import {Signature} from "@chainsafe/bls";
 import {EpochContext} from "../index";
 
 export function verifyBlockSignature(
@@ -16,8 +17,8 @@ export function verifyBlockSignature(
     signedBlock.message,
     domain
   );
-  return epochCtx.index2pubkey[signedBlock.message.proposerIndex].verifyMessage(
-    Signature.fromCompressedBytes(signedBlock.signature.valueOf() as Uint8Array),
+  return bls.Signature.fromBytes(signedBlock.signature.valueOf() as Uint8Array).verify(
+    epochCtx.index2pubkey[signedBlock.message.proposerIndex],
     signingRoot
   );
 }
--- a/packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts
@@ -1,5 +1,5 @@
 import {ByteVector, hash, toHexString, readOnlyMap, BitList, List} from "@chainsafe/ssz";
-import {PublicKey} from "@chainsafe/bls";
+import bls, {IPublicKey} from "@chainsafe/bls";
 import {
   Attestation,
   AttestationData,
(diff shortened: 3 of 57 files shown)


packages/lodestar-beacon-state-transition/src/fast/util/block.ts
```diff
@@ -1,7 +1,8 @@
+import bls from "@chainsafe/bls";
 import {BeaconState, SignedBeaconBlock} from "@chainsafe/lodestar-types";
+
```
dapplion: unnecessary newline

packages/lodestar-cli/src/cmds/account/cmds/validator/create.ts
```diff
@@ -72,7 +72,7 @@ and pre-computed deposit RPL data",
 
   handler: async (args) => {
     // Necessary to compute validator pubkey from privKey
-    await initBLS();
+    await init("blst-native");
```
dapplion: Use fallback to WASM

packages/lodestar-cli/src/cmds/beacon/handler.ts
```diff
@@ -20,7 +20,7 @@ import {initBeaconState} from "./initBeaconState";
  * Run a beacon node
  */
 export async function beaconHandler(args: IBeaconArgs & IGlobalArgs): Promise<void> {
-  await initBLS();
+  await init("blst-native");
```
dapplion: Here use fallback to WASM but if it fails add a fat message of performance warning

packages/lodestar-cli/src/wallet/Wallet.ts
```diff
@@ -43,7 +43,7 @@ export class Wallet extends Keystore {
    */
   static async fromMnemonic(mnemonic: string, password: string, name: string): Promise<Wallet> {
     const seed = deriveKeyFromMnemonic(mnemonic);
-    const publicKey = generatePublicKey(seed);
+    const publicKey = bls.PrivateKey.fromBytes(seed).toPublicKey().toBytes();
```
dapplion: All this code should be removed, we should not handle validator creation in the CLI

packages/lodestar-validator/test/unit/services/block.test.ts
```diff
@@ -35,51 +40,41 @@ describe("block proposing service", function () {
...
+    const service = new BlockProposingService(config, [keypair], rpcClientStub, slashingProtectionStub, logger);
     const result = await service.createAndPublishBlock(0, slot, generateFork(), ZERO_HASH);
     expect(result).to.not.be.null;
     expect(rpcClientStub.beacon.blocks.publishBlock.calledOnce).to.be.true;
   });
 
   it("should produce correct block - last signed in previous epoch", async function () {
+    const privateKey = bls.PrivateKey.fromBytes(toBufferBE(BigInt(98), 32));
+    const publicKey = privateKey.toPublicKey();
+    const keypair = {privateKey, publicKey};
```
dapplion: These three lines are repeated a lot. Probably it's not necessary to pass around both the privateKey and the publicKey, only the privateKey. If it's out the scope of this PR add a note on issue as todo

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -129,9 +129,10 @@ export class ValidatorApi implements IValidatorApi {
 
     const aggregate = attestations.reduce((current, attestation) => {
       try {
-        current.signature = Signature.fromCompressedBytes(current.signature.valueOf() as Uint8Array)
-          .add(Signature.fromCompressedBytes(attestation.signature.valueOf() as Uint8Array))
-          .toBytesCompressed();
+        current.signature = bls.aggregateSignatures([
+          current.signature.valueOf() as Uint8Array,
+          attestation.signature.valueOf() as Uint8Array,
+        ]);
```
dapplion: This is very inefficient serializing and de-serializing every time.

dapplion (changes requested):

packages/lodestar-validator/test/unit/services/block.test.ts
wemeetagain: seems like a bigger refactor of the validator

packages/lodestar-cli/src/wallet/Wallet.ts
wemeetagain: This is just to fix build errors. We can tackle this as a separate PR

packages/lodestar-beacon-state-transition/src/block/operations/deposit.ts
```diff
@@ -2,7 +2,7 @@
  * @module chain/stateTransition/block
  */
 
-import {verify} from "@chainsafe/bls";
+import bls from "@chainsafe/bls";
```
mpetrunic, line 5: why removing named import? We should use named import everywhere

dapplion (approved):

packages/lodestar-beacon-state-transition/src/block/operations/deposit.ts
dapplion: To have a globally available bls object that gets set conditionally named exports didn't work. Instead it exports a bls variable that gets mutated on init()
https://github.com/ChainSafe/bls/blob/c8ddfcaa3b12bdc81557686958e36b07535f42c3/src/index.ts#L9

dapplion: Using a specific implementation directly allows the use of named exports. Because you are not going through that switch statement
```ts
import {verify} from "@chainsafe/bls/browser"
```

mpetrunic: oh man, than we made bls worse. And it sucks in  web usage unless you import from /browser.
I guess that was big arg for my way of abstracting it. :sob:

dapplion: In your abstraction you never got to add blst and herumi in the same PR. Only abstracted herumi without BLST. I'm down to change the abstraction if it works well on all settings!

mpetrunic: we just needed to implement given interface with blst and it would work...Why haven't you raised concern that you failed to implement named export?

dapplion: > we just needed to implement given interface with blst and it would work 

I'm not so sure about that. Again, if you know a way to make it work I would love to merge it

> Why haven't you raised concern that you failed to implement named export?

The current export strategy was there from the first iteration of the PR

dapplion: Also you don't have to use default exports, you can do
```ts
import {bls} from "@chainsafe/bls";
```
And it would work as well

dapplion (changes requested):
We may want to use this PR to simplify the change and keep using named import 
https://github.com/ChainSafe/bls/pull/44

packages/lodestar-beacon-state-transition/src/block/operations/deposit.ts
wemeetagain: resolved w chainsafe/bls#44, will update here

packages/lodestar-beacon-state-transition/src/fast/util/epochContext.ts
```diff
@@ -1,5 +1,5 @@
 import {ByteVector, hash, toHexString, readOnlyMap, BitList, List} from "@chainsafe/ssz";
-import {PublicKey} from "@chainsafe/bls";
+import bls, {IPublicKey} from "@chainsafe/bls";
```
dapplion, line 2: Use PublicKey class type

dapplion (changes requested):

dapplion (dismissed):

dapplion (approved):

PR #1802, opened 2020-11-30
Generalize spec download tool

dapplion:
- Move download script to spec utils repo to be accessible by chainsafe/bls
- Expose a bin CLI


wemeetagain:
Good idea, tho it seems something about this update is causing the script to now exit early before everything is downloaded.

dapplion:
> Good idea, tho it seems something about this update is causing the script to now exit early before everything is downloaded.

Right, using pipeline now to await competition + removed the process.exit(0) on success

packages/lodestar-spec-test-util/src/downloadTests.ts
```diff
@@ -0,0 +1,39 @@
...
+import tar from "tar";
+import stream from "stream";
+import {promisify} from "util";
+
+const BASE_URL = "https://github.com/ethereum/eth2.0-spec-tests/releases/download";
+
+// eslint-disable-next-line prettier/prettier
+const TESTS = ["general", "mainnet", "minimal"];
+
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
```
mpetrunic: where is floating promise?

packages/lodestar-spec-test-util/src/downloadTests.ts
```diff
@@ -0,0 +1,39 @@
...
+): Promise<void> {
+  if (log) log(`outputDir = ${outputDir}`);
+
+  await Promise.all(
+    TESTS.map(async (test) => {
+      const URL = `${BASE_URL}/${specVersion}/${test}.tar.gz`;
+      const OUTPUT = `${outputDir}`;
+
+      // download tar
+      const {data, headers} = await axios({
```
mpetrunic, line 47: can we check if tests are already downloaded and extracted?

packages/lodestar-spec-test-util/src/downloadTests.ts
```diff
@@ -0,0 +1,39 @@
...
+export async function downloadTests(
+  {specVersion, outputDir}: {specVersion: string; outputDir: string},
+  log?: (msg: string) => void
+): Promise<void> {
+  if (log) log(`outputDir = ${outputDir}`);
+
+  await Promise.all(
+    TESTS.map(async (test) => {
+      const URL = `${BASE_URL}/${specVersion}/${test}.tar.gz`;
+      const OUTPUT = `${outputDir}`;
```
mpetrunic: I think we should separate them by version:
```suggestion
      const OUTPUT = `${outputDir}/${specVersion}`;
```

dapplion: We should review how to deal with version changes. The tests are really heavy so maybe we should keep only a single version locally. What do you think?

packages/lodestar-spec-test-util/src/downloadTests.ts
dapplion: What condition should it use for that? That there are some files in `outputDir`? That there are files with certain names in `outputDir`?

mpetrunic: We can check if we have {specVersion} directory and not download then

packages/lodestar-spec-test-util/src/downloadTests.ts
mpetrunic: Yeah we can remove all non {specVersion} directories. Maybe we should have flag for cleanup? I would much rather keep spec tests locally instead of downloading everytime if I'm switching between versions.

dapplion: Sounds good. Should I bring `yargs` at this point, or it's overkill?

mpetrunic: probably overkill

packages/lodestar-spec-test-util/src/downloadTests.ts
wemeetagain: Would prefer to have this logic outside this function/file. "if tests don't exist, run this function"
We may have this condition checked in a github action, for example. "if cache hit, restore tests, else, download tests"

mpetrunic: I would keep it in script if not in this method.

wemeetagain (approved):

PR #1805, opened 2020-11-30
Cache balances when processing epoch

twoeths:
we can save one `balances` loop in `TreeBacked<BeaconState>` by

+ New `balances` array inside EpochProcess
+ Populate it in `processRewardsAndPenalties`
+ Use it in `processFinalUpdates`
+ Save 165ms with over 110000 validators (same validator number in Pyrmont and Medalla)

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/fast/epoch/processRewardsAndPenalties.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/epoch/processRewardsAndPenalties.ts
@@ -24,6 +24,7 @@ export function processRewardsAndPenalties(epochCtx: EpochContext, process: IEpo
       newBalances[i] -= penalty;
     }
   });
+  process.balances = newBalances;
   // important: do not change state one balance at a time
   // set them all at once, constructing the tree in one go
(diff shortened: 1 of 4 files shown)


twoeths:
waiting for #1799

twoeths:
ready for review

packages/lodestar-beacon-state-transition/src/fast/epoch/processRewardsAndPenalties.ts
```diff
@@ -24,6 +24,7 @@ export function processRewardsAndPenalties(epochCtx: EpochContext, process: IEpo
       newBalances[i] -= penalty;
     }
   });
+  process.balances = newBalances;
```
dapplion, line 27: For a second I thought we are adding stuff the global NodeJS `process` object xD I another PR may we want to change the name so it doesn't match any existent global variable?

dapplion (dismissed):

dapplion (approved):

3xtr4t3rr3str14l (approved):

wemeetagain (approved):

PR #1813, opened 2020-12-02
Update eslint and typescript-eslint

mpetrunic:
- this allows us to use modern typescript features like:
   - import type only
   -  `export * as ns from "..."` which is quite important in phase1 stuff since types gets namespaced

There were a lot of breaking changes but all rules now work better. We can probably try to enable some rules that didn't work previously.


packages/lodestar-cli/src/cmds/account/cmds/validator/deposit.ts
```diff
@@ -7,7 +7,7 @@ const deprecatedDescription =
 - eth2.0-deposit-cli: https://github.com/ethereum/eth2.0-deposit-cli \
 - Ethereum Foundation launchpad: https://medalla.launchpad.ethereum.org";
 
-export const deposit: ICliCommand<{}, IAccountValidatorArgs & IGlobalArgs> = {
+export const deposit: ICliCommand<Record<string, unknown>, IAccountValidatorArgs & IGlobalArgs> = {
```
dapplion: Is the empty object type `{}` forbidden in this version? I'm afraid `Record<string, unknown>` might allow arbitrary keys.

mpetrunic: {} apparently accepts any non null type. As claimed by eslint rule

dapplion: I see it. For empty object you should use `Record<string, never>`

dapplion: That rule is lying, behaviour is not equivalent. I want that access of non declared keys to be wrong:
Screenshot from 2020-12-02 19-17-32

dapplion (changes requested):
See comment above https://github.com/ChainSafe/lodestar/pull/1813#discussion_r534420172
Accessing arbitrary options and args on the CLI should no be permited

packages/lodestar-cli/src/cmds/account/cmds/validator/deposit.ts
mpetrunic: > That rule is lying, behaviour is not equivalent. I want that access of non declared keys to be wrong:
> Screenshot from 2020-12-02 19-17-32
yes, but you can call function fn1(1) and it would be ok.

image

Eslint suggestion is to use Record<never, never>
image

That said, that is out of scope for this PR as types were wrong in the first place

dapplion: `Record<never, never>` works great! Thanks for updating the PR

dapplion (approved):

dapplion (approved):

packages/lodestar-cli/test/unit/config/beaconParams.test.ts
```diff
@@ -17,6 +17,7 @@ describe("config / beaconParams", () => {
   const testCases: {
     id: string;
     kwargs: Parameters<typeof getBeaconParams>[0];
+    // eslint-disable-next-line @typescript-eslint/naming-convention
```
3xtr4t3rr3str14l, line 20: <s>it feels to me like we are disabling this particular rule in a lot of places in this PR.  should we maybe just disable it at the eslint config level instead?</s>
nevermind, i see why it's being done

3xtr4t3rr3str14l (approved):

packages/lodestar-cli/test/unit/config/beaconParams.test.ts
mpetrunic: I had briliant solution to allow snake case if property is quoted but prettier doesnt let me

wemeetagain (approved):

PR #1816, opened 2020-12-02
Implement CachedValidatorsBeaconState

### LINKED ISSUE #1789: POC: Create a StateTransitionBeaconState
Right now an epoch processing take at least 2 validators loop, note that each validator loop is very expensive with TreeBacked data structure.

I used to see some epoch state transition take more than 5s, some are up to 10s because we loop through validators multiple times (to prepare epoch, to rotate epoch, to slash validators, to process validator exits etc), this does not count the BLS signature verification. The goal is to loop through validators array only 1 time.

+ Create a wrapped BeaconState inside lodestar-beacon-state-transition. The interface should be extended from BeaconState to ensure we don't break the current code.
+ Cache a validator array inside that when preparing epoch.
+ For reading loop, use the cached validators array. The performance should be way better than looping through TreeBacked data structure.
+ When writing a validator, write to both the cache and the real TreeBacked<Validator>
+ The cached validators array should be short-lived during an epoch process in order not to consume memory

twoeths:
resolves #1789 

## Goal
+ In case of a slow state transition, looping through `validators` inside `TreeBacked` structure takes a significant time
+ Create CachedValidatorsBeaconState that cache a regular `validators` array after the 1st loop, use it in lodestar-state-transition only. Note that when processing epoch, we always have to do the loop to prepare EpochProcess.
+ When update/add a validator, update to both the cache and `TreeBacked<BeaconState>`
+ When loop/read, use the cached `validators` if we have it

## Trade off
+ We have to use `setValidator`, `addValidator`, `getValidator` api in state transition instead of accessing `validators` directly. I guess we can enforce it at compile time using `Omit<BeaconState, "validators">` or something if we want to.

## Performance gain with 100k validators as in Pyrmont/Medalla
+ Process 1st epoch: save 500ms - 600ms
+ From the 2nd epoch, it only takes around 1100ms per epoch (instead of >=2800ms)
+ When we have to process 16 validator exits as in the performance test, the time to process block is reduced from 6000ms to 1400ms
+ See more details in the performance tests of lodestar-beacon-state-transition

## Test
+ Passed spec tests
+ Synced Pyrmont successfully

## Memory
+ When we process a regular block it even does not create a cached validators
+ It short-live in state transition only then released

Although the performance is improved, I'd like to have everyone review to see if we have any potential issues with this approach. Thanks!

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/fast/util/interface.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/util/interface.ts
@@ -1,6 +1,8 @@
 import {IReadonlyEpochShuffling} from ".";
-import {ValidatorIndex, Slot} from "@chainsafe/lodestar-types";
-import {ByteVector} from "@chainsafe/ssz";
+import {ValidatorIndex, Slot, BeaconState, Validator} from "@chainsafe/lodestar-types";
+import {ByteVector, readOnlyForEach} from "@chainsafe/ssz";
+import {createIFlatValidator, IFlatValidator} from "./flatValidator";
+import {config} from "@chainsafe/lodestar-config/lib/presets/mainnet";
 
 /**
  * Readonly interface for EpochContext.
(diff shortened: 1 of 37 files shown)


packages/lodestar-beacon-state-transition/src/fast/util/interface.ts
```diff
@@ -13,3 +15,139 @@ export interface ReadonlyEpochContext {
   readonly previousShuffling?: IReadonlyEpochShuffling;
   getBeaconProposer: (slot: Slot) => ValidatorIndex;
 }
+
```
twoeths, line 17: main file to review

twoeths:
replaced by #1925

Issue #1775, opened 2020-11-20
Code Explanation


wemeetagain:
- Explaining non-obvious design decisions in top-level comments
- Guiding thru the overall structure/architecture
- Adding typedocs to every public interface/constant/class/etc

Current Progress (checked indicates that documentation/comment polishing for that package has been merged into master):

- [x] benchmark-utils
- [x] lodestar
- [x] lodestar-beacon-state-transition
- [x] lodestar-cli
- [x] lodestar-config
- [x] lodestar-db
- [x] lodestar-fork-choice
- [x] lodestar-params
- [x] lodestar-spec-test-util
- [x] lodestar-types
- [x] lodestar-utils
- [x] lodestar-validator
- [x] spec-test-runner

3xtr4t3rr3str14l:
I'm starting this off by making sure we have our comments covered in `lodestar-cli`.  I'm thinking it's probably a good idea to break this up into chunks so we don't have a massive diff, if we end up needing a lot of more comments, so I will probably make a PR just for the lodestar-cli comment updates.

3xtr4t3rr3str14l:
I'm moving on to continue this process for the `lodestar-validator` package

3xtr4t3rr3str14l:
I just added the list of checkmarks above in @wemeetagain 's original comment.  I figured it might less less unwieldy than having an issue for every package.  If enough people disagree, I'll make issues for each package, but for now, if anyone starts working on this effort for a particular package, just say so.  That way we don't accidentally step on each others toes :)

3xtr4t3rr3str14l:
i'm starting on `lodestar-beacon-state-transition` next

3xtr4t3rr3str14l:
moving on to `lodestar-fork-choice` next

EDIT: `lodestar-fork-choice` is already pretty well-documented/commented, but I'll probably add some high-level things to the README

3xtr4t3rr3str14l:
calling this done

PR #1817, opened 2020-12-02
lodestar-validator comment updates

3xtr4t3rr3str14l:
part of the effort tracked in #1775 

adds some clarifying comments to the lodestar-validator package

also
- removed isRunning from validator.ts (it wasn't being used)
- turned off validator events in blockService.stop()
- corrected the naming of  param in the public interface for `getAttesterDuties`

### DIFF
--- a/packages/lodestar-validator/src/services/attestation.ts
+++ b/packages/lodestar-validator/src/services/attestation.ts
@@ -37,6 +37,9 @@ import {isValidatorAggregator} from "../util/aggregator";
 import {abortableTimeout} from "../util/misc";
 import {getAggregationBits, getAggregatorModulo} from "./utils";
 
+/**
+ * Service that sets up and handles validator attester duties.
+ */
 export class AttestationService {
   private readonly config: IBeaconConfig;
   private readonly provider: IApiClient;
@@ -70,6 +73,9 @@ export class AttestationService {
     this.logger = logger;
   }
 
+  /**
+   * Starts the AttestationService by updating the validator attester duties and turning on the relevant listeners for clock events.
+   */
   public start = async (): Promise<void> => {
     this.controller = new AbortController();
     const currentEpoch = this.provider.clock.currentEpoch;
@@ -82,6 +88,9 @@ export class AttestationService {
     this.provider.on(BeaconEventType.HEAD, this.onHead);
   };
 
+  /**
+   * Stops the AttestationService by turning off the relevant listeners for clock events.
+   */
   public stop = async (): Promise<void> => {
     if (this.controller) {
       this.controller.abort();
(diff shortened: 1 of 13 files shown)


dapplion:
I appreciate the effort to document the codebase further but adding comments to already explicit code may be counterproductive. For example, if we have
```ts
/**
 * Gets validator from index
 */
getValidatorFromIndex(index: number) {
   ...
}
```
If in the future we have to change the function to fetch from pubkey instead of index we now have to change two things instead of one. Otherwise we can end up with
```ts
/**
 * Gets validator from index
 */
getValidatorFromPubkey(pubkey: PublickKey) {
   ...
}
```

----

The original goal of the issue and good practice with comments is to only use them when it's extremely necessary to explain something that can be obvious with clearer code. An example would be
https://github.com/ChainSafe/lodestar/blob/5162c9136854c8c0b3816bb7622468ee0f3cc61b/packages/lodestar/test/sim/threaded/noEth1SimWorker.ts#L12-L15
There is no other way to explain the decision of using herumi's implementation there than with a comment.

---

I would love to know other's opinion on the topic since this is very opinionated @wemeetagain @mpetrunic @tuyennhv

3xtr4t3rr3str14l:
> 
> The original goal of the issue and good practice with comments is to only use them when it's extremely necessary to explain something that can be obvious with clearer code. 

Perhaps there are some comments that some may see as redundant here, but I more or less commented what wasn't immediately clear to me based on the function name and params alone.  Also, some comments were added for the sake of having it popup in the IDE tool tips.  I personally feel that commenting only in cases of "extreme" necessity is maybe a little drastic.  My general rule of thumb is, if I can't tell exactly what the function does based on the function name and params (or maybe if the function body is only a few lines that also more closely resemble the English language upon immediate inspection), then a comment on top of that function can be very useful.  That's just my take though.  I too am open to hear others' opinions.

3xtr4t3rr3str14l:
I will say that maybe the comment for `getStateValidator` is redundant (perhaps a better solution is renaming that function to `getValidatorFromState` and removing the comment), but I'm not sure how other people see the other comments.  Are there specific comments that I added that you have issue with @dapplion ?  Or do you feel that my overall strategy is off?

dapplion:
Right, IDE IntelliSense documentation is a big plus. I would say that unless anyone else says anything continue with your effort

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -165,6 +184,11 @@ export class AttestationService {
     }
   }
 
+  /**
+   * Perform attestation/aggregation duties.
+   * IFF a validator is an attester, create and submit an attestation.
```
twoeths, line 189: typo

twoeths: oh so this `IFF` is intentional? :)

3xtr4t3rr3str14l: correct :) :+1:

wemeetagain (approved):

PR #1820, opened 2020-12-03
chore: update libp2p to 0.30

vasco-santos:
This PR updates libp2p to its next release with type definitions created on release.

The initial goal here was to just remove `libp2p-ts` but the transition is not so smoothly as plugging in the new libp2p versions.  This will need follow up work to slowly transition into libp2p generated types as there are a lot of imported types across the codebase from `libp2p-ts` that need to be updated to the paths/names of the types exported by libp2p.


vasco-santos:
@wemeetagain this should be ready to be merged 🎉 
Feel free to ping me here while you are moving from `libp2p-ts` to libp2p's own types for reviews/discussion

wemeetagain:
closing in favor of #1999

PR #1821, opened 2020-12-04
Add signature sets getters

dapplion:
Building block for BLS batch verify functionality. Recycles existing code in the fast state transition function.

Lodestar as a consumer can do 
```ts
const signatureSets = getAllBlockSignatureSets(epochCtx, state, signedBlock)
verifySignatureSetsBatch(signatureSets)
```


dapplion:
Partially related to this, how would a multi-block signature set collection look like? Can you collect signatures from N consecutive blocks without running the state transition? The modified state influences which pubkeys gets collected. So, it should run the state transition first with verify off and then collect the signatures, then verify the whole thing and revert if invalid?

EDIT: Lighthouse uses `cheap_state_advance_to_obtain_committees` https://github.com/sigp/lighthouse/blob/65dcdc361b0c91dabe53c886e32aa7952251be68/beacon_node/beacon_chain/src/block_verification.rs#L320

packages/lodestar-beacon-state-transition/src/fast/block/isValidIndexedAttestation.ts
```diff
@@ -35,15 +34,34 @@ export function isValidIndexedAttestation(
...
+export function getIndexedAttestationSignatureSet(
+  epochCtx: EpochContext,
+  state: BeaconState,
+  indexedAttestation: IndexedAttestation,
+  indices?: number[]
+): ISignatureSet {
+  const config = epochCtx.config;
+  const domain = getDomain(config, state, DomainType.BEACON_ATTESTER, indexedAttestation.data.target.epoch);
+
+  // TODO: Should the indexes be sorted for signature validation?
```
mpetrunic: I don't think it's relavant in context of signature verification

packages/lodestar-beacon-state-transition/src/fast/signatureSets/proposerSlashings.ts
```diff
@@ -0,0 +1,18 @@
...
+import {EpochContext} from "../index";
+import {getProposerSlashingSignatureSets} from "../block/processProposerSlashing";
+import {flatten} from "./util";
+
+export function getProposerSlashingsSignatureSets(
+  epochCtx: EpochContext,
+  state: BeaconState,
+  signedBlock: SignedBeaconBlock
+): ISignatureSet[] {
+  return flatten(
```
mpetrunic: regular array.flat should work here

packages/lodestar-beacon-state-transition/src/fast/signatureSets/index.ts
```diff
@@ -0,0 +1,31 @@
...
+import {getVoluntaryExitsSignatureSets} from "./voluntaryExits";
+
+export * from "./types";
+export * from "./verify";
+
+/**
+ * Includes all signatures on the block (except the deposit signatures) for verification.
+ * Deposits are not included because they can legally have invalid signatures.
+ */
+export function getAllBlockSignatureSets(
```
mpetrunic, line 18: cannot find where do we use this. I assume in processBlock and if it passes all verifySignatures should be set to false?

packages/lodestar-beacon-state-transition/src/fast/signatureSets/types.ts
```diff
@@ -0,0 +1,18 @@
...
+
+interface ISignatureSinglePubkeySet {
+  type: "single-pubkey";
+  pubkey: PublicKey;
+  signingRoot: Root;
+  signature: BLSSignature;
+}
+
+interface ISignatureMultiplePubkeySet {
+  type: "multiple-pubkeys";
```
mpetrunic: maybe make this enum. Also I think single/aggregate/multiple should be better names

packages/lodestar-beacon-state-transition/src/fast/signatureSets/index.ts
dapplion: This is not used anywhere yet but should be used as a pre-step to running the state-transition. My idea was to add the signature sets getters first in this PR and consume them in a latter PR

packages/lodestar-beacon-state-transition/src/fast/signatureSets/types.ts
dapplion: Damn right, much more cleaner

packages/lodestar-beacon-state-transition/src/fast/block/processRandao.ts
```diff
@@ -16,23 +15,39 @@ export function processRandao(
...
+ * Extract signatures to allow validating all block signatures at once
+ */
+export function getRandaoRevealSignatureSet(
+  epochCtx: EpochContext,
+  state: BeaconState,
+  body: BeaconBlockBody
+): ISignatureSet {
+  const config = epochCtx.config;
+  const epoch = epochCtx.currentShuffling.epoch;
+  const proposerIndex = epochCtx.getBeaconProposer(state.slot);
```
wemeetagain: should pass in the whole beacon block, so we can do `epochCtx.getBeaconProposer(block.slot);` and avoid needing the state dialed per-slot.

See https://github.com/sigp/lighthouse/blob/c6baa0eed131c5e8ecc5860778ffc7d4a4c18d2d/consensus/state_processing/src/per_block_processing/signature_sets.rs#L121

wemeetagain (changes requested):
> Partially related to this, how would a multi-block signature set collection look like?
> Can you collect signatures from N consecutive blocks without running the state transition? 
> The modified state influences which pubkeys gets collected. 
> So, it should run the state transition first with verify off and then collect the signatures, then verify the whole thing and revert if invalid?

I'm hoping we can avoid real state transitions, but just roll forward to the highest slot to get the committees and any additional validator pubkeys required, similar to what lighthouse does.

We want to be sure that these signature sets don't rely on the state being dialed to particular slots or other things that would require finer grained tweaking of the state in between processing signatures of consecutive blocks.

wemeetagain (approved):

PR #1831, opened 2020-12-06
Improve lodestar-config exports

dapplion:
Prevent having to import from /lib and deliver what the README promises.

Adds two files in the root of lodestar-config that re-export the preset files


packages/lodestar-config/README.md
```diff
@@ -14,8 +14,8 @@ Configuration variables for an Eth2 network -- consists of params and ssz types
 
 ```typescript
 // mainet and minimal presets are available under non-default exports
-import {config as mainnetConfig} from "@chainsafe/lodestar-config/lib/presets/mainnet";
-import {config as minimalConfig} from "@chainsafe/lodestar-config/lib/presets/mainnet";
+import {mainnetConfig} from "@chainsafe/lodestar-config/mainnet";
+import {minimalConfig} from "@chainsafe/lodestar-config/mainnet";
```
twoeths: `import {config} from "@chainsafe/lodestar-config/mainnet";`?

packages/lodestar-config/mainnet.d.ts
```diff
@@ -0,0 +1 @@
+export * from "./lib/presets/mainnet";
```
twoeths, line 1: can we have `lodestar-config/mainnet.ts` instead?

packages/lodestar-config/README.md
dapplion: Do you suggest to only use a single import statement in the README example?

packages/lodestar-config/mainnet.d.ts
dapplion: That would require compiling code outside of src, I haven't found a clean way to do it

packages/lodestar-config/README.md
twoeths: ah I just mean the change you did everywhere is different from this README :D

wemeetagain (dismissed):

wemeetagain:
file conflict

wemeetagain (approved):

PR #1835, opened 2020-12-06
Remove unnecessary try / catch expect.fail() pattern

dapplion:
The pattern
```ts
    try {
      await jobs;
    } catch (e) {
      expect.fail();
    }
```
provides no extra functionality than just
```ts
      await jobs;
```
expect that it's muting the error so it's less useful to debug


wemeetagain (dismissed):

wemeetagain:
file conflict

wemeetagain (approved):

PR #1839, opened 2020-12-07
Assemble block: process epoch at most 1 time

### LINKED ISSUE #1838: Block Delay at epoch transition slot
**Describe the bug**
Following up the conversation in https://discord.com/channels/595666850260713488/598292067260825641/785242071857627147

when producing a block for `slot % 32 === 0`, we process epoch 2 times

**Expected behavior**
when producing a block, we want to process epoch at most 1 time

twoeths:
resolves #1838

### DIFF
--- a/packages/lodestar/src/chain/factory/block/index.ts
+++ b/packages/lodestar/src/chain/factory/block/index.ts
@@ -2,11 +2,12 @@
  * @module chain/blockAssembly
  */
 
-import {fastStateTransition, IStateContext} from "@chainsafe/lodestar-beacon-state-transition";
+import {processBlock} from "@chainsafe/lodestar-beacon-state-transition/lib/fast/block";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {BeaconBlock, Bytes96, Root, Slot} from "@chainsafe/lodestar-types";
-import {EMPTY_SIGNATURE, ZERO_HASH} from "../../../constants";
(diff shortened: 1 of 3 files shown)


mpetrunic (dismissed):

mpetrunic (approved):

packages/lodestar/src/chain/factory/block/index.ts
```diff
@@ -30,25 +31,22 @@ export async function assembleBlock(
     stateRoot: ZERO_HASH,
     body: await assembleBody(config, db, eth1, stateContext.state, randaoReveal, graffiti),
   };
-
-  const preStateContext = await chain.regen.getBlockSlotState(head.blockRoot, slot - 1);
```
wemeetagain, line 34: Don't we still need this preStateContext?

twoeths: no, we don't need to bc we only run `processBlock` instead of a `fastStateTransition` to leverage the 1st/initial `stateContext`
Current:
+ process head state up to slot
+ process head state up to (slot - 1) (which is redundant)
+ run `fastStateTransition` which is processing slot (epoch possibly) + process block

This PR changes to:
+ process head state up to slot
+ ~process head state up to (slot - 1)~
+ ~run `fastStateTransition` which is processing slot (epoch possibly) + process block~ => process block only

wemeetagain (approved):

PR #1843, opened 2020-12-07
Remove self invoking functions where not strictly necessary

dapplion:
Self invoking functions add 1 indentation level and are a little harder to reason about. Let me know if some instances are necessary


dapplion:
Definitely still necessary somewhere :laughing:

wemeetagain (approved):

PR #1844, opened 2020-12-07
Add whitespace to lodestar/network

dapplion:
+ better readability

### DIFF
--- a/packages/lodestar/src/network/encoders/utils.ts
+++ b/packages/lodestar/src/network/encoders/utils.ts
@@ -6,14 +6,14 @@ import {SnappyFramesUncompress} from "./snappy-frames/uncompress";
 
 export function getCompressor(encoding: ReqRespEncoding): (data: Buffer) => AsyncGenerator<Buffer> {
   switch (encoding) {
-    case ReqRespEncoding.SSZ_SNAPPY: {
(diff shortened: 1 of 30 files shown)


twoeths (approved):

packages/lodestar/src/network/encoders/utils.ts
```diff
@@ -6,14 +6,14 @@ import {SnappyFramesUncompress} from "./snappy-frames/uncompress";
 
 export function getCompressor(encoding: ReqRespEncoding): (data: Buffer) => AsyncGenerator<Buffer> {
   switch (encoding) {
-    case ReqRespEncoding.SSZ_SNAPPY: {
```
mpetrunic, line 9: We don't have rule to enforce brackets on case statements. :disappointed: 
Brackets really helps out to notice case boundaries.

mpetrunic (approved):

wemeetagain (approved):

packages/lodestar/src/network/encoders/utils.ts
dapplion: For switch statements with a lot of short cases not using brackets is pretty clean. If the case code grows brackets are nice so I'm not sure about making them mandatory

PR #1845, opened 2020-12-07
Use for of instead of forEach where possible

dapplion:
for of instead of forEach:
- It works on any iterable
- It supports all kinds of control flow in the loop body, like continue, break, return, yield and await.
- on errors there's one less stack line
- forEach can potentially hide problems such as unhandled promises
- personally, I find it more readable (tho this is subjective)


mpetrunic (commented):
TBH I like forEach more:
- it's easier to cast type of item being iterated
- easier access to item index and update of array being iterated
- I frequently confuse for-in and for-of (having hard time figuring what iterates over what)

dapplion:
> TBH I like forEach more:
> 
> * it's easier to cast type of item being iterated
> * easier access to item index and update of array being iterated
> * I frequently confuse for-in and for-of (having hard time figuring what iterates over what)

.forEach is great for the scenarios that you described. This PR only changes the cases where none of these conditions are met. Also `for ... in` should never be used, but use `for ... of Object.keys()`

dapplion:
By using for of I found to instances of not handleded promises. Tagging the previous author according to git blame

1. @tuyennhv could you take a look at this one
https://github.com/ChainSafe/lodestar/blob/8059f347e91dde3a75f7cd5db7a9dfd43d2d06a9/packages/lodestar/src/sync/regular/oneRangeAhead/processor.ts#L51
Before it was
```ts
sortedBlocks.forEach((block) => this.chain.receiveBlock(block, false));
```
Should it be awaited or not?

2. @mpetrunic could you take a look at this one
https://github.com/ChainSafe/lodestar/blob/8059f347e91dde3a75f7cd5db7a9dfd43d2d06a9/packages/lodestar/test/e2e/sync/sync.test.ts#L31
```ts
validators.forEach((v) => v.start());
```

mpetrunic:
> 1. @mpetrunic could you take a look at this one
>    https://github.com/ChainSafe/lodestar/blob/8059f347e91dde3a75f7cd5db7a9dfd43d2d06a9/packages/lodestar/test/e2e/sync/sync.test.ts#L31
> 
> ```ts
> validators.forEach((v) => v.start());
> ```

We don't really care about awaiting there. If we are gonna await, Promise.all should be better fit.

That said, it has nothing to do with for-of vs forEach. It's just that in forEach callback we didn't put async keyword (`validators.forEach(async (v) => {v.start()});`). And once you moved it into async scope it shows error. It wouldn't display error if for-of is executed in non-async context/method.

dapplion:
> It wouldn't display error if for-of is executed in non-async context/method.

The error `@typescript-eslint/no-floating-promises` still shows in non-async context, you can try it

mpetrunic:
> > It wouldn't display error if for-of is executed in non-async context/method.
> 
> The error `@typescript-eslint/no-floating-promises` still shows in non-async context, you can try it

add brackets in forEach :)

dapplion:
> add brackets in forEach :)

Sure, but here we should compare the one-line syntax to be fair.
```ts
validators.forEach(v => v.start())
```
```ts
for (const v of validators) v.start()
```
The error will show up in the second and not in the first. In multiline, it will show up in both styles

mpetrunic:
> > add brackets in forEach :)
> 
> Sure, but here we should compare the one-line syntax to be fair.
> 
> ```
> validators.forEach(v => v.start())
> ```
> 
> ```
> for (const v of validators) v.start()
> ```
> 
> The error will show up in the second and not in the first. In multiline, it will show up in both styles

That's not equivalent.
Without brackets there is implicit return so in for-of that would be:
```typescript
for (const v of validators) {
   (function() { return return v.start() })()
}
```

dapplion:
> > > add brackets in forEach :)
> > 
> > 
> > Sure, but here we should compare the one-line syntax to be fair.
> > ```
> > validators.forEach(v => v.start())
> > ```
> > 
> > 
> > ```
> > for (const v of validators) v.start()
> > ```
> > 
> > 
> > The error will show up in the second and not in the first. In multiline, it will show up in both styles
> 
> That's not equivalent.
> Without brackets there is implicit return so in for-of that would be:
> 
> ```ts
> for (const v of validators) {
>    (function() { return return v.start() })()
> }
> ```

Why would you want that in a for loop?

mpetrunic:
> Why would you want that in a for loop?

You don't. You don't wan't that in forEach which is what we/I did there since void-ing promise wasn't working in previous eslint version when you didn't want to await promise.

twoeths:
> By using for of I found to instances of not handleded promises. Tagging the previous author according to git blame
> 
> 1. @tuyennhv could you take a look at this one
>    https://github.com/ChainSafe/lodestar/blob/8059f347e91dde3a75f7cd5db7a9dfd43d2d06a9/packages/lodestar/src/sync/regular/oneRangeAhead/processor.ts#L51
>    
>    Before it was
> 
> ```ts
> sortedBlocks.forEach((block) => this.chain.receiveBlock(block, false));
> ```
> 
> Should it be awaited or not?


no we don't have to but it should also be ok to use await here too

dapplion:
> no we don't have to but it should also be ok to use await here too

Would you mind taking a look yourself @tuyennhv? That function uses non straightforward Promise pattern that I don't understand

packages/lodestar/src/sync/regular/oneRangeAhead/processor.ts
```diff
@@ -47,7 +47,10 @@ export class BlockRangeProcessor implements IBlockRangeProcessor {
         resolve,
         signal,
       };
-      sortedBlocks.forEach((block) => this.chain.receiveBlock(block));
+      for (const block of sortedBlocks) {
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        this.chain.receiveBlock(block);
```
twoeths: we can use `await` here and remove the `eslint-disable-next-line` comment

bajohn: Hi, new to this repo but confirming you don't want to do something like this? I think what @mpetrunic mentioned:
```
Promise.all(sortedBlocks.map(block => this.chain.receiveBlock(block)));
```

twoeths: we really want to send blocks to processor queue in order, that's why I intentionally want to do that.
I think initially `Promise.all` was my first try but it's not a guarantee to execute them sequentially, at least in the spec.

bajohn: Makes sense!

twoeths: we can do something like this, if you feel comfortable with it:
```
public async processUntilComplete(blocks: SignedBeaconBlock[], signal: AbortSignal): Promise<void> {
    if (!blocks || !blocks.length) return;
    const sortedBlocks = sortBlocks(blocks);
    await Promise.all([
      new Promise((resolve) => {
        this.logger.info("Imported blocks for slots", {blocks: blocks.map((block) => block.message.slot)});
        const lastRoot = this.config.types.BeaconBlock.hashTreeRoot(sortedBlocks[sortedBlocks.length - 1].message);
        this.target = {
          root: lastRoot,
          resolve,
          signal,
        };
      }),
      (async function (chain) {
        for (const block of sortedBlocks) {
          await chain.receiveBlock(block);
        }
      })(this.chain)
    ]
    ) // end Promise.all
  }
```

wemeetagain (commented):
Can you see if there is a noticeable performance cost to using for-of instead of forEach?

dapplion:
> Can you see if there is a noticeable performance cost to using for-of instead of forEach?

This is a very popular question and different benchmarks show different results, where sometimes forEach wins others for ... of. I guess it depends on the platform or engine then? If performance is important somewhere (like SSZ bytes comparison), we should use the regular for loop which is faster and noticeable once you do very large numbers of iterations (+1e6)

bajohn:
> > Can you see if there is a noticeable performance cost to using for-of instead of forEach?
> 
> This is a very popular question and different benchmarks show different results, where sometimes forEach wins others for ... of. I guess it depends on the platform or engine then? If performance is important somewhere (like SSZ bytes comparison), we should use the regular for loop which is faster and noticeable once you do very large numbers of iterations (+1e6)

I think the difference between the two methods would be insignificant compared to the async operation happening inside the loop, probably not the most important place to optimize!

dapplion:
> I think the difference between the two methods would be insignificant compared to the async operation happening inside the loop, probably not the most important place to optimize!

Agree, the reasons for choosing between for of and forEach should be code readability, features and subjective preference

wemeetagain (approved):

Issue #1196, opened 2020-07-22
refactor (some) modules to respect the "start should resolve quickly" interface


dapplion:
Follow up of https://github.com/ChainSafe/lodestar/pull/1179. At least the `sync` module should be refactored so it's start function resolves after the critical initial steps are completed.

But then: **How are the long-running processes handled**? How do they do error handling and communication with the top instance consumer or main process?

As of now most NodeJS versions and browsers log unhandled rejections but don't stop the main process. There should be some way for this long-running process to tell the main process to panic. But, how can we do that in a browser & NodeJS friendly way?

----

Relevant code snippets
https://github.com/ChainSafe/lodestar/blob/5e8af2502c2716daa4efe90341d2a3923ded9d13/packages/lodestar/src/node/nodejs.ts#L151-L154

https://github.com/ChainSafe/lodestar/blob/5e8af2502c2716daa4efe90341d2a3923ded9d13/packages/lodestar/src/eth1/impl/ethers.ts#L85-L86

https://github.com/ChainSafe/lodestar/blob/5e8af2502c2716daa4efe90341d2a3923ded9d13/packages/lodestar/src/eth1/impl/ethers.ts#L104-L109

https://github.com/ChainSafe/lodestar/blob/5e8af2502c2716daa4efe90341d2a3923ded9d13/packages/lodestar/src/chain/blocks/processor.ts#L61-L64

----

IMO the takeaway of PR https://github.com/ChainSafe/lodestar/pull/1179 should be:
1. Determine root causes of problems here (eg: needing an error-handling strategy for long-running processes)
2. Document problem areas in the code (there are a handful here)
3. Eslint rule in place to prevent further unhandled rejections

In the problem cases, just add an eslint-disable line for now and document the specific case. Don't try to add hacks that just mask the problem, it won't really get us towards our goal. A real solution is bigger than can be solved here, likely involves refactoring in some small or large way.

_Originally posted by @wemeetagain in https://github.com/ChainSafe/lodestar/pull/1179#discussion_r458841614_

PR #1852, opened 2020-12-09
Prevent using recursion in LocalClock

dapplion:
Now that we use AbortSignals it's not necessary to use recursion in the LocalClock. Functionality should be the same but without relying on setTimeout. While errors on `onNextSlot` are probably impossible, now the stack should come from LocalClock's instantiation instead of a timeout.

### DIFF
--- a/packages/lodestar-validator/src/api/LocalClock.ts
+++ b/packages/lodestar-validator/src/api/LocalClock.ts
@@ -1,5 +1,7 @@
+import {AbortSignal} from "abort-controller";
 import {Epoch, Slot} from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
+import {sleep, ErrorAborted} from "@chainsafe/lodestar-utils";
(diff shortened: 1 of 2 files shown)


packages/lodestar-validator/src/api/LocalClock.ts
```diff
@@ -45,12 +50,13 @@ export class LocalClock implements IBeaconClock {
     return computeEpochAtSlot(this.config, this.currentSlot);
   }
 
-  private abort = (): void => {
-    clearTimeout(this.timeoutId);
-    if (this.signal) {
-      this.signal.removeEventListener("abort", this.abort);
+  private async start(): Promise<void> {
```
wemeetagain, line 53: https://github.com/ChainSafe/lodestar/issues/1196 :(

dapplion: This function may not even exist as you can run this code directly on the constructor. I can rename to `runInternalProcess` or something to differentiate the pattern

dapplion:
Consider reviewing after https://github.com/ChainSafe/lodestar/pull/2022 and move the clock to core

PR #1853, opened 2020-12-09
Import AbortSignal wherever used

dapplion:
Would it be possible to remove AbortSignal from Typescript's global scope? Probably it's baked into DOM types, but it results in different types which cause types problems, and not sure if it can cause problems in runtime


wemeetagain (approved):

wemeetagain:
I don't think its possible to remove baked-in types :(

PR #1854, opened 2020-12-09
Add babel plugins to optimise code

mpetrunic:
If you compare spec tests execution time in ci on master, it seems like this is shaving ~1s on mainnet sanity tests.


dapplion:
> If you compare spec tests execution time in ci on master, it seems like this is shaving ~1s on mainnet sanity tests.

The 1s second savings is against how much as base execution time?

mpetrunic:
> > If you compare spec tests execution time in ci on master, it seems like this is shaving ~1s on mainnet sanity tests.
> 
> The 1s second savings is against how much as base execution time?

13s -> 12s

It is probably more significant on prymont because it mostly optimises loops and callbacks which depends on number of validators

dapplion:
> 13s -> 12s
> 
> It is probably more significant on prymont because it mostly optimises loops and callbacks which depends on number of validators

The 1s difference is with a resolution of 1s o milliseconds? Is it possible to benchmark this PR for a few blocks on Pyrmont?

mpetrunic:
> > 13s -> 12s
> > It is probably more significant on prymont because it mostly optimises loops and callbacks which depends on number of validators
> 
> The 1s difference is with a resolution of 1s o milliseconds? Is it possible to benchmark this PR for a few blocks on Pyrmont?

milliseconds. Between 1100 - 900 ms.
It is but we need exported state and block in ssz or some format.

dapplion:
Why is `babel-plugin-object-to-json-parse` relevant in Lodestar's context? I thought this is to optimize page load times in the web

mpetrunic:
> Why is `babel-plugin-object-to-json-parse` relevant in Lodestar's context? I thought this is to optimize page load times in the web

https://v8.dev/blog/cost-of-javascript-2019#json

That said. I'm gonna try tommorow to remove it, I don't think we initialize a lot of objects anyways. And don't have exact numbers on difference in NodeJs (should be same because they both use v8 engine I guess). Keep in mind that this code should be ran on browser as well.

dapplion:
> https://v8.dev/blog/cost-of-javascript-2019#json

A blockchain node is meant to run for at least a while so I'm not sure that would make a difference since it only affects parsing time once. That sounds like an optimization that a browser consumer could do if it's found appropriate

mpetrunic:
> > https://v8.dev/blog/cost-of-javascript-2019#json
> 
> A blockchain node is meant to run for at least a while so I'm not sure that would make a difference since it only affects parsing time once. That sounds like an optimization that a browser consumer could do if it's found appropriate

As I said, both chrome and nodejs are run by v8 so it makes no difference. I was thinking of code where we have
`const something = committee.map((c) => {index, c, validators: [1, 5, 7, 8]})` where it would get transpiled to `const something = committee.map((c) => JSON.parse("{index, c, validators: [1, 5, 7, 8]}"))` which is alot faster to initialize especially if you are doing that in a loop a lot of times.

dapplion:
> As I said, both chrome and nodejs are run by v8 so it makes no difference. I was thinking of code where we have
> `const something = committee.map((c) => {index, c, validators: [1, 5, 7, 8]})` where it would get transpiled to `const something = committee.map((c) => JSON.parse("{index, c, validators: [1, 5, 7, 8]}"))` which is alot faster to initialize especially if you are doing that in a loop a lot of times.

I know the purpose of the optimization but I'm not sure it will have a noticeable effect on Lodestar. People don't benchmark if Prysm started running in 500ms instead of 750ms, it's more like it synced in 4 hours instead of 8 hours.

mpetrunic:
> I know the purpose of the optimization but I'm not sure it will have a noticeable effect on Lodestar. People don't benchmark if Prysm started running in 500ms instead of 750ms, it's more like it synced in 4 hours instead of 8 hours.

We actually initialize quite a lot of objects during block processing and epoch transitions. Like for every attestation we initialize new pendingAttestation object, signature sets etc.

dapplion:
> We actually initialize quite a lot of objects during block processing and epoch transitions. Like for every attestation we initialize new pendingAttestation object, signature sets etc.

If get this right, here you will have to parse the json data every call of the function
```js
const fn = (c) => JSON.parse("{index, c, validators: [1, 5, 7, 8]}")
```
While here that js code will be parsed only once by v8 on start
```js
const fn = (c) => {index, c, validators: [1, 5, 7, 8]}
```
is that right?

Also, in the blog post here https://v8.dev/blog/cost-of-javascript-2019#json
- Recommends using with objects of > 10kB
- Measure the actual impact before making any changes

----

Anyway I think the best would be to benchmark the impact of each of the 3 pluggings you are adding to make an informed decision of which speed things up and which don't

PR #1857, opened 2020-12-09
Lightclient params and types

mpetrunic:
I've based branch and PR against already dones types and params of phase1 as it includes a lot more generic way to handle namespacing and I'm starting to work on state transition updates based on this branch.

### DIFF
--- a/packages/lodestar-params/src/beaconParams.ts
+++ b/packages/lodestar-params/src/beaconParams.ts
@@ -2,6 +2,7 @@
 import {IBeaconParams} from "./interface";
 import {ContainerType, BigIntUintType, NumberUintType, ByteVectorType} from "@chainsafe/ssz";
 import {Phase1Params} from "./phase1";
+import {LightClientParamsSSZ} from "./presets/lightclient";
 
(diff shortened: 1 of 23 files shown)


packages/lodestar-params/src/beaconParams.ts
```diff
@@ -13,6 +14,7 @@ const ByteVector20 = new ByteVectorType({length: 20});
 export const BeaconParams = new ContainerType<IBeaconParams>({
   fields: {
     phase1: Phase1Params,
+    lightclient: LightClientParamsSSZ,
```
wemeetagain: should be `lightClient` for consistency

wemeetagain: Actually, reading thru the eth2 spec, it seems that Lightclient is one word (not two).

So should be `LightClient` => `Lightclient`, `lightClient` => `lightclient` everywhere for consistency.

mpetrunic: I dont think they can agree but yeah lets stick to the second one

wemeetagain (approved):

PR #1858, opened 2020-12-09
lodestar-fork-choice README elaboration

3xtr4t3rr3str14l:
part of #1775 

adds some bare minimal detail to the lodestar-fork-choice README to demonstrate the ways in which a fork choice object can be instantiated.

### DIFF
--- a/packages/lodestar-fork-choice/README.md
+++ b/packages/lodestar-fork-choice/README.md
@@ -16,7 +16,7 @@ import {toHexString} from "@chainsafe/ssz";
 const config = createIBeaconConfig({params}); // see @chainsafe/lodestar-config for more info
 let someExistingBlockData = {
   // block data from somewhere to initialize the fork choice
-  // ArrayBuffer(0) is just a placeholder
+  // Buffer(32) is just a placeholder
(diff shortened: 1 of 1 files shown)


dapplion:
I bet this README will get out-dated really fast and we won't remember to update it as it evolves. Would it be better to link to where it's used in the Lodestar code as an example of usage?

3xtr4t3rr3str14l:
> I bet this README will get out-dated really fast and we won't remember to update it as it evolves. Would it be better to link to where it's used in the Lodestar code as an example of usage?

yeah i like that idea.  i'll do that instead

packages/lodestar-fork-choice/README.md
```diff
@@ -4,11 +4,8 @@
 
 ## Usage
 
-```
-const lodestarForkChoice = require('@chainsafe/lodestar-fork-choice');
-
-// TODO: DEMONSTRATE API
-```
+See here for the lodestar implementation of the interface provided by this package:
```
wemeetagain:
```suggestion
 See usage in the `lodestar` package here: 
```

wemeetagain (approved):

PR #1861, opened 2020-12-10
Use TreeBacked<SignedBeaconBlock> to improve sync time

### LINKED ISSUE #1860: Use TreeBacked<SignedBeaconBlock> to reduce sync time
Currently we use `TreeBacked<SignedBeaconBlock>` before we process a block.
But sync services also calculate roots before that time to check linear chain.
We should be able to build `TreeBacked<SignedBeaconBlock>` at network layer to reduce sync time, need to do that and benchmark.

twoeths:
resolves #1860 
+ convert to `TreeBacked<SignedBeaconBlock>` from beacon_blocks_by_range, beacon_blocks_by_root and gossip
+ the below is statistic from syncing Pyrmont till slot 5000

|#|Branch|Time (in min)|
|-|-|-|
|1|master|22|
|2|master|21.5|
|1|this PR|17.5|
|2|this PR|18|

### DIFF
--- a/packages/lodestar/src/network/encoders/response.ts
+++ b/packages/lodestar/src/network/encoders/response.ts
@@ -1,6 +1,7 @@
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {P2pErrorMessage, ResponseBody} from "@chainsafe/lodestar-types";
 import {ILogger} from "@chainsafe/lodestar-utils";
+import {CompositeType} from "@chainsafe/ssz";
 import {AbortController} from "abort-controller";
 import BufferList from "bl";
 import {decode, encode} from "varint";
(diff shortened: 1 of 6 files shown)


mpetrunic (dismissed):

wemeetagain:
It might be worth trying to deserialize into `TreeBacked<SignedBeaconBlock>` directly "off the wire". Eg: in `src/network/encoders/...` and elsewhere.
Instead of deserializing to ObjBacked first, then converting to TreeBacked.

packages/lodestar/src/constants/network.ts
```diff
@@ -37,34 +39,56 @@ export enum MethodResponseType {
...
+      isTreeBacked: false,
+    }),
     responseType: MethodResponseType.SingleResponse,
   },
   [Method.Ping]: {
     requestSSZType: (config: IBeaconConfig): typeof config.types.Ping => config.types.Ping,
-    responseSSZType: (config: IBeaconConfig): typeof config.types.Ping => config.types.Ping,
+    responseSSZType: (config: IBeaconConfig): {type: BigIntUintType; isTreeBacked: boolean} => ({
+      type: config.types.Ping,
+      isTreeBacked: false,
```
mpetrunic: Can't we determine backing from value directly instead of passing separate boolean flags?

twoeths: @wemeetagain could you help with this :)

packages/lodestar/src/network/encoders/response.ts
```diff
@@ -130,7 +131,14 @@ export function eth2ResponseDecode(
 
         if (uncompressedData.length === sszLength) {
           try {
-            yield type.deserialize(uncompressedData.slice()) as ResponseBody;
+            if (typeInfo.isTreeBacked) {
```
wemeetagain: `if (method === this || method === that) {`

wemeetagain (approved):

Issue #1772, opened 2020-11-20
Add block range API to chain


wemeetagain:
Allow the chain to accept and process a range of blocks at once

Eventually we'll integrate #1473

PR #1863, opened 2020-12-10
Use correct epoch context to get signature set

### LINKED ISSUE #1859: Cannot get signature set at epoch transition
**Describe the bug**
If I enable signature verification for initial sync, the sync gets stuck at slot 32, seems like the checkpoint epoch context is not correct to get proposer index, and we should rotate epoch.

But an additional epoch rotate is not a cheap at all as it loops through state validators (500ms to 600ms as far as I know for Pyrmont), we need to benchmark verifying signature signature set in batch vs. old flow (verify one by one) if we do that.

**Expected behavior**
+ Initial sync should get through epoch transition with signature verification
+ make sure it logs out error, in this case it takes time to debug `beacon proposer index out of range` error
Screen Shot 2020-12-10 at 09 01 58

**Steps to Reproduce**
Sync Peermont from scratch with signature verification

twoeths:
resolves #1859 

+ at epoch transition, we have to rotate epoch context in order to get signature set, i.e. after `processSlots`
+ verifying signatures inside `lodestar-beacon-state-transition` also makes sense in term of spec tests


twoeths:
verified through a Pyrmont sync with enabled bls for initial sync, it took me 39.5 minutes to reach slot 5000 (22 minutes without it as shown in #1861)

wemeetagain:
One thing to note, when we add multi-block processing (#1772), we will need to be verifying the signatures outside of `lodestar-beacon-state-transition`. I believe this is why @dapplion originally put this logic in `lodestar`.

twoeths:
> One thing to note, when we add multi-block processing (#1772), we will need to be verifying the signatures outside of `lodestar-beacon-state-transition`. I believe this is why @dapplion originally put this logic in `lodestar`.

I see so this is the wrong direction to go. I'll open another PR to fix this issue

PR #1865, opened 2020-12-10
Follow XDG Base Directory Specification

### LINKED ISSUE #1232: Change CLI default root-dir
The current default root-dir is `./.lodestar` in the current working directory when running `lodestar`.
Using a local root-dir has been convenient for development/debug purposes, but its not standard behavior for systemwide-installed programs.

Brainstorm / implement a better default root-dir path. Some possible options:
- $HOME/lodestar
- Something following the XDG standard: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#basics
- ??

dapplion:
Fixes https://github.com/ChainSafe/lodestar/issues/1232

Note: This PR offers no migration plan so it's a breaking change. To keep using an existing DB users will have to do a mv o use a custom --rootDir

### DIFF
--- a/packages/lodestar-cli/src/cmds/dev/handler.ts
+++ b/packages/lodestar-cli/src/cmds/dev/handler.ts
@@ -19,8 +19,6 @@ import {IGlobalArgs} from "../../options";
 import {IDevArgs} from "./options";
 import {initializeOptionsAndConfig} from "../init/handler";
 import {mkdir, initBLS} from "../../util";
-import {getGlobalPaths} from "../../paths/global";
-import {TestnetName} from "../../testnets";
 
 /**
  * Run a beacon node with validator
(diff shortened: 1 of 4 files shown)


packages/lodestar-cli/src/cmds/dev/handler.ts
```diff
@@ -35,9 +36,9 @@ export async function devHandler(args: IDevArgs & IGlobalArgs): Promise<void> {
   beaconNodeOptions.set({network: {discv5: {enr}}});
 
   // Custom paths different than regular beacon, validator paths
-  const rootDir = path.join(args.rootDir || defaultRootDir, "dev");
-  const chainDir = path.join(rootDir, "beacon");
-  const validatorsDir = path.join(rootDir, "validators");
+  const globalPaths = getGlobalPaths({...args, testnet: "dev" as TestnetName});
```
mpetrunic: I don't think dev chain should be stored in home directory. Let's keep it in working directory

mpetrunic (approved):

wemeetagain (approved):

PR #1873, opened 2020-12-11
Light client state transition

mpetrunic:
This will probably receive a lot more update as there are ongoing effort to make lightclient stuff executable. Will update once they merge it.


packages/lodestar-beacon-state-transition/src/lightclient/block.ts
```diff
@@ -0,0 +1,13 @@
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
+import {Lightclient} from "@chainsafe/lodestar-types";
+import * as phase0Block from "../block";
```
wemeetagain, line 3: might be nice to segregate phase0-specific code to a `phase0/` directory

wemeetagain (approved):

packages/lodestar-beacon-state-transition/src/lightclient/block.ts
mpetrunic: > might be nice to segregate phase0-specific code to a `phase0/` directory

I would love that. But it's enormus diff and change because we didn't use exports in root index 😭

wemeetagain: the exports from `src/fast/` ?

mpetrunic: We use imports from fast and normal and they are all over the place

PR #1877, opened 2020-12-14
POC: immutablejs for CachedValidatorsBeaconState

twoeths:
just want to have some statistics regarding ImmutableJS as @wemeetagain 's ideas in last meeting
+ this is based on #1816. The difference is we use ImmutableJs for the validators cache and apply CachedValidatorsBeaconState globally across lodestar instead of just state transition.
+ the statistic

|branch|average epoch transition(ms)|time to sync to slot 10000(mins)|
|-|-|-|
|master|4114|49|
|this branch|1395|27|

+ I really don't see any differences in terms of heap memory in my local environment. Perhaps someone can try in your environments too!


dapplion:
Oh mama, those results are great!

twoeths:
replaced by #1925

PR #1884, opened 2020-12-14
Simpify eth2RequestEncode

dapplion:
eth2RequestEncode is not used anywhere for more than a request at once which adds complexity to the implementation.

The goal of this PR is to simplify things assuming there's a single request to be streamed
- Split utility into three functions
  - `serializeRequestBody`: json => Uint8Array
  - `getRequestBodyEncodedStream`: Instantiate compressor and return stream
  - `streamRequestBodyTo`: Calls above fns and pipes to connection sink
- If the body fails to serialize a proper error will be thrown since it happens outside of pipe
- Tests are updated with a `simulateRequest` utility which creates a streamSink to compare the streamed body. It also reduces code duplication with a for loop over encodings

### DIFF
--- a/packages/lodestar/src/network/encoders/request.ts
+++ b/packages/lodestar/src/network/encoders/request.ts
@@ -1,4 +1,5 @@
 import {Method, Methods, ReqRespEncoding} from "../../constants";
+import {toBuffer} from "../../util/buffer";
 import {ILogger} from "@chainsafe/lodestar-utils";
 import {RequestBody} from "@chainsafe/lodestar-types";
 import {decode, encode} from "varint";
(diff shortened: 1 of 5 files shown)


packages/lodestar/src/util/buffer.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Cast Uint8Array to Buffer efficiently
+ */
+export function toBuffer(bytes: Uint8Array): Buffer {
```
mpetrunic:
```suggestion
export function toBufferUnsafe(bytes: Uint8Array): Buffer {
```
This cast also means that change on original array would be reflected on Buffer which is not what we always wan't

packages/lodestar/src/network/encoders/request.ts
```diff
@@ -1,42 +1,59 @@
...
+  config: IBeaconConfig,
+  method: Method,
+  requestBody: RequestBody
+): Uint8Array | null {
+  const type = Methods[method].requestSSZType(config);
+  if (type == null || requestBody == null) {
+    return null;
+  } else {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    return type.serialize(requestBody as any);
```
mpetrunic: I feel like catching error here and logging it makes more sense. Currently it would end up as unhandled rejection and we won't close streams we opened.

packages/lodestar/test/unit/network/encoders/request.test.ts
```diff
@@ -1,124 +1,70 @@
...
 
-  it("should work - multiple request - ssz_snappy", async function () {
-    const requests = await pipe(
-      [BigInt(1), BigInt(2)],
-      eth2RequestEncode(config, logger, Method.Ping, ReqRespEncoding.SSZ_SNAPPY),
-      eth2RequestDecode(config, logger, Method.Ping, ReqRespEncoding.SSZ_SNAPPY),
-      all
-    );
+        await expect(simulateRequest(Method.Status, encoding, requestBody)).to.be.rejectedWith(
+          "Cannot convert undefined or null to object"
```
mpetrunic: That's not really error you would expect isn't it?

packages/lodestar/test/unit/network/encoders/request.test.ts
```diff
@@ -1,124 +1,70 @@
...
+        );
+      });
```
mpetrunic, line 46: Can you add test for request without body?

mpetrunic (changes requested):

packages/lodestar/src/network/encoders/request.ts
dapplion: Added a commit to close stream on error

packages/lodestar/test/unit/network/encoders/request.test.ts
dapplion: SSZ errors could be improved upstream TBH, meanwhile I've wrapped this with `ReqRespSerializeError`

mpetrunic: I mean't it would be more appropriate to be `Failed to serialize request <request body> as type Status` or something

packages/lodestar/test/unit/network/encoders/request.test.ts
dapplion: streamRequestBodyTo is not called for a request without a body, there's an if in sendRequestStream

mpetrunic (approved):

packages/lodestar/test/unit/network/encoders/request.test.ts
dapplion: Oh I like the text, tho showing the request body might be too much since it could potentially large? Or not for requests

mpetrunic: Not sure, I doubt this error will appear often. I think it can only apear as a result of developer error anyways

wemeetagain:
close in favor of #1901

dapplion:
> close in favor of #1901

They are different efforts but I'll this from scratch anyway

PR #1886, opened 2020-12-14
Split state transition package (2nd attempt)

dapplion:
@mpetrunic please can you fix the tests yourself please :pray: I've done the bulk of the work but can't get the tests working after many attempts. If you agree on splitting the package help me get the PR green asap before it gets unsalvagable again due to conflicts.


mpetrunic:
I need to think about it more.
I wouldn't separate fast state transition into separate package. I think that would be really bad in terms of future planned hardforks. We would new package for every planned hardfork then

I think best way forward would be following directory structure in current state transition:

- phase0
   - fast 
      - index.ts - export * as fast from
   - original
      - index.ts - export * from
   - utils (shared things between fast and original)
   - index.ts - exports utils, fast and original
- lightclient
   - index.ts - export * as lightclient
- index.ts - export phase0 and lightclient

That way everything will be namespaced and we can map epoch => hardfork namespace.

In code it would be much clearer since we would write:
`phase0.fast.processBlock` or `phase0.processBlock`

cc @wemeetagain I will put this in that issue you created as well

wemeetagain:
It seems @dapplion is suggesting something like this (I only see 2 packages):
- naive/
  - index.ts
  - phase0/
  - lightclient/
  - utils/
- fast/
  - index.ts
  - phase0/
  - utils/

wemeetagain:
Closing in favor of reorganizing beacon state transition package (as mentioned in planning meeting / standup)

PR #1890, opened 2020-12-15
blockService and attestationService validator data consolidations

3xtr4t3rr3str14l:
consolidating some of the validator data in `blockService` and `attestationService` in `lodestar-validator`
for example, in attestationService, we have 3 separate objects for keeping track of validator data, but we need all of those objects to be in the same order, so it seems best to consolidate them into one object.


packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -35,17 +35,16 @@ import {ISlashingProtection} from "../slashingProtection";
...
 export class AttestationService {
   private readonly config: IBeaconConfig;
   private readonly provider: IApiClient;
-  // order is important
-  private readonly secretKeys: SecretKey[] = [];
-  // order is important
-  private readonly publicKeys: BLSPubkey[] = [];
-  // order is important
-  private readonly validators: (ValidatorResponse | null)[] = [];
+  private readonly validatorClients: {
```
mpetrunic:
```suggestion
  private readonly validators: {
```
1 validator client process can host multiple validators

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -314,7 +309,7 @@ export class AttestationService {
       computeEpochAtSlot(this.config, aggregate.data.slot)
     );
     const signingRoot = computeSigningRoot(this.config, this.config.types.AggregateAndProof, aggregateAndProof, domain);
-    return this.secretKeys[aggregatorIndex].sign(signingRoot).toBytes();
+    return this.validatorClients[aggregatorIndex].secretKey.sign(signingRoot).toBytes();
```
mpetrunic: I wonder if we should add utility method to fetch validator by pubkey from our array of validators. This is quite confusing because in eth2 context aggregatorIndex is index of validator in validator registry not in array of our validators

mpetrunic: or better yet instead of array have hexpubkey => {validator, secretKey} map so it's faster in case we are having multiple validators

3xtr4t3rr3str14l: I'm not sure I follow.  Why/how would we want to fetch by pubkey if we are given a `number` index?  We could also maybe just call the array `validatorStore` instead of `validators` if that is the confusing part.

mpetrunic: Index in our validators array which we call throughout block and attestation service is not the same index as in eth2 spec is confusing part.
Example:
We are running 6 validators in our validator client. There is total of 10k validators in beacon chain.
If we wanna take some action with validator in position 2 in our client. We will store "2" as value in variable called validatorIndex or aggregatorIndex but in beacon context that's compleatly different validator. This is especially confusing where duty.validatorIndex is not the index of validator in our array.

So we should have map pubkey => {secretKey, validator} so our code is more clear and we can find validator faster since validator is identified with pubkey in most of api responses

3xtr4t3rr3str14l: just pushed a change for this

3xtr4t3rr3str14l:
would also like to get some feedback on this approach in particular:

https://github.com/ChainSafe/lodestar/pull/1890/files#diff-ff30b1253b91df678666c2f076e07037711e7bac916b98fc0257c4ee7046339eR187

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -43,16 +42,11 @@ import {getAggregationBits, getAggregatorModulo} from "./utils";
...
-  // order is important
-  private readonly publicKeys: BLSPubkey[] = [];
-  // order is important
-  private readonly validators: (ValidatorResponse | null)[] = [];
+  private readonly validators: Map<string, ValidatorAndSecret> = new Map();
   private readonly slashingProtection: ISlashingProtection;
   private readonly logger: ILogger;
 
-  private nextAttesterDuties: Map<Slot, Map<number, IAttesterDuty>> = new Map();
+  private nextAttesterDuties: Map<Slot, Map<BLSPubkey, IAttesterDuty>> = new Map();
```
twoeths: is there a reason to use `BLSPubkey` as a key for Map? Normally we use number or string as key for Map to easily compare by value, it should help with `Map.get()`, `Map.has()` etc.

3xtr4t3rr3str14l: nope, no good reason. i meant to change this in accordance with the new `validators` object (which uses the string instead of BLSPubKey).  will change this

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -399,18 +401,15 @@ export class AttestationService {
    * Update the local list of validators based on the current head state.
    */
   private async updateValidators(): Promise<void> {
-    let index = 0;
-    for (const pubkey of this.publicKeys) {
-      // fetch validator details if missing
-      if (!this.validators[index]) {
+    this.validators.forEach(async (v, pk) => {
```
dapplion: Using an async callback inside a forEach callback is not super clean. Would recommend using `for..of` loop like before which handles rejections properly or await Promise.all() if this needs to be paralelized

dapplion (dismissed):

packages/lodestar-validator/src/types.ts
```diff
@@ -1,14 +1,18 @@
...
 };
 
 export interface IAttesterDuty extends AttesterDuty {
-  // internal index
-  attesterIndex: number;
   isAggregator: boolean;
 }
+
+export type ValidatorAndSecret = {
+  validator: ValidatorResponse | null;
```
wemeetagain, line 17: why is this `| null`?

3xtr4t3rr3str14l: no good reason.  i think this is a vestige of a previous step of this PR.  will remove the `| null`

3xtr4t3rr3str14l: oh wait, actually, it's because this is what was there before:
https://github.com/ChainSafe/lodestar/pull/1890/files#diff-ff30b1253b91df678666c2f076e07037711e7bac916b98fc0257c4ee7046339eL51

might actually need to put this back as the tests are failing after i changed it:

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -43,16 +41,11 @@ import {getAggregationBits, getAggregatorModulo} from "./utils";
 export class AttestationService {
   private readonly config: IBeaconConfig;
   private readonly provider: IApiClient;
-  // order is important
-  private readonly secretKeys: SecretKey[] = [];
-  // order is important
-  private readonly publicKeys: BLSPubkey[] = [];
-  // order is important
-  private readonly validators: (ValidatorResponse | null)[] = [];
+  private readonly validators: Map<string, ValidatorAndSecret> = new Map();
```
mpetrunic: for the sake of readability might be good to add comment what this string key really is or create string alias
`type PublicKeyHex = string; Map<PublicKeyHex, ValidatorAndSecret>`

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -43,16 +41,11 @@ import {getAggregationBits, getAggregatorModulo} from "./utils";
...
   private readonly slashingProtection: ISlashingProtection;
   private readonly logger: ILogger;
 
-  private nextAttesterDuties: Map<Slot, Map<number, IAttesterDuty>> = new Map();
+  private nextAttesterDuties: Map<Slot, Map<string, IAttesterDuty>> = new Map();
```
mpetrunic: same here

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -65,9 +58,7 @@ export class AttestationService {
     this.config = config;
     this.provider = rpcClient;
     for (const secretKey of secretKeys) {
-      this.secretKeys.push(secretKey);
-      this.publicKeys.push(secretKey.toPublicKey().toBytes());
-      this.validators.push(null);
+      this.validators.set(toHexString(secretKey.toPublicKey().toBytes()), {validator: null, secretKey});
```
mpetrunic: Doesn't `secretKey.toPublicKey()` have toHex or something method?

3xtr4t3rr3str14l: looks like it does!  will fix this in next commit

packages/lodestar-validator/src/services/attestation.ts
```diff
@@ -43,16 +41,11 @@ import {getAggregationBits, getAggregatorModulo} from "./utils";
 export class AttestationService {
   private readonly config: IBeaconConfig;
   private readonly provider: IApiClient;
-  // order is important
-  private readonly secretKeys: SecretKey[] = [];
-  // order is important
-  private readonly publicKeys: BLSPubkey[] = [];
-  // order is important
-  private readonly validators: (ValidatorResponse | null)[] = [];
+  private readonly validators: Map<PublicKeyHex, ValidatorAndSecret> = new Map();
```
wemeetagain: Is there a reason that the AttestationService uses a `Map<PublicKeyHex, XXX>` but the BlockService uses a `XXX[]` to store validator pubkeys/secretkeys/etc?

Somehow I'd think we'd use one way to organize them.

3xtr4t3rr3str14l: yeah that sounds like a good idea.  i'll look into it

3xtr4t3rr3str14l: just pushed a fix for this

wemeetagain (approved):

Issue #1881, opened 2020-12-14
Refresh Lodestar User Guide


wemeetagain:
**Is your feature request related to a problem? Please describe.**
Documentation page is out of date

Refresh user guide for content and style
Add FAQ / usecases
Add dev guide / ecosystem packages
Add motivation / architecture descriptions / diagrams

**Additional context**
https://docs.ethermint.zone/

3xtr4t3rr3str14l:
i'm gonna start heading in this general direction

3xtr4t3rr3str14l:
open to suggestions if anyone has anything more specific or different than what is listed above

dapplion:
User guides were updates 2 months ago and are fairly updated

PR #1891, opened 2020-12-15
refresh user guide docs

3xtr4t3rr3str14l:
part of #1881 .  specifically the part about:
> Refresh user guide for content

just updates the existing info for now.  new content to come later

- generalizes instructions for the user to run on mainnet/testnet rather than having an entry for each testnet
- updates guides for existing commands to have latest info


docs/installation.md
```diff
@@ -31,6 +31,12 @@ Build across all packages
 yarn run build
 ```
 
+Or if you are using [Lerna](https://lerna.js.org/):
+
+```bash
+lerna bootstrap
+```
+
```
mpetrunic, line 39: not sure we need this section. Since we are using workspaces, yarn install should always suffice. We should remove `yarn run build` section since install will build code as well

docs/usage/index.md
```diff
@@ -1,12 +1,12 @@
 # Usage
 
 - [Local tesnet](local): To quickly test and run Lodestar we recommend to start a local testnet.
-- [Connect to Medalla](medalla)
-- [Connect to Altona](altona)
-- [Connect to Witti](witti) *(deprecated)*
+- [Connect to mainnet or public testnte](testnet)
```
mpetrunic:
```suggestion
- [Connect to mainnet or public testnet](testnet)
```

mpetrunic: I think we should add pyrmont

docs/usage/key-management.md
```diff
@@ -66,38 +66,21 @@ This command will:
 
 To import a keystore that was created via the ETH2.0 Deposit Launch Pad:
 
-```
-./lodestar account validator import --testnet medalla --directory <path to your launchpad keys>
+```bash
+./lodestar account validator import --testnet .$TESTNET_NAME --directory <path to your launchpad keys>
```
dapplion: Why a dot before the variable name TESTNET_NAME?

3xtr4t3rr3str14l: woops!  just changed it

wemeetagain (approved):

PR #1892, opened 2020-12-16
Debug api to download state

### LINKED ISSUE #1856: Implement debug api to download state
https://ethereum.github.io/eth2.0-APIs/#/Debug/getState

this could be used to
+ debug any issues we have
+ or to download weak subjectivity checkpoint state

twoeths:
resolves #1856 

Some implementation notes:
+ Store finalized state by slot instead of epoch
+ Support storing finalized state by state root
+ Download full beacon state by using `accept` header `application/octet-stream`, otherwise it'll return json instead

## Question: 

+ For old slots, this is only able to download states of finalized checkpoints. We could also download any finalized states by slot by regenerating it from a persisted finalized checkpoint state and run state transitions through related finalized blocks, using a WorkerThread is suitable in this use case. Do we want to do it?

### DIFF
--- a/packages/lodestar/src/api/impl/beacon/state/utils.ts
+++ b/packages/lodestar/src/api/impl/beacon/state/utils.ts
@@ -1,8 +1,17 @@
 // this will need async once we wan't to resolve archive slot
-import {computeEpochShuffling} from "@chainsafe/lodestar-beacon-state-transition/lib/fast/util";
+import {GENESIS_SLOT} from "@chainsafe/lodestar-beacon-state-transition";
+import {computeEpochShuffling, EpochContext} from "@chainsafe/lodestar-beacon-state-transition/lib/fast/util";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {IForkChoice} from "@chainsafe/lodestar-fork-choice";
-import {Epoch, Validator, ValidatorIndex, ValidatorStatus, ValidatorResponse, Gwei} from "@chainsafe/lodestar-types";
+import {
+  Epoch,
+  Validator,
+  ValidatorIndex,
+  ValidatorStatus,
+  ValidatorResponse,
+  Gwei,
+  BeaconState,
+} from "@chainsafe/lodestar-types";
 import {fromHexString, readOnlyMap} from "@chainsafe/ssz";
 import {IBeaconChain} from "../../../../chain";
 import {IBeaconDb} from "../../../../db/api";
@@ -19,7 +28,7 @@ export async function resolveStateId(
     return (await db.stateCache.get(forkChoice.getHead().stateRoot)) ?? null;
   }
   if (stateId === "genesis") {
-    const state = await db.stateArchive.get(0);
+    const state = await db.stateArchive.get(GENESIS_SLOT);
     if (!state) return null;
     return {
       state,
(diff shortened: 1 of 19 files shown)


packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -32,21 +41,24 @@ export async function resolveStateId(
...
-    throw new Error("Invalid block id");
+    throw new Error("Invalid state id");
   }
-  //todo: resolve archive slot -> state
   const blockSummary = forkChoice.getCanonicalBlockSummaryAtSlot(slot);
   if (blockSummary) {
     return (await db.stateCache.get(blockSummary.stateRoot)) ?? null;
+  } else {
+    const finalizedState = await db.stateArchive.get(slot);
+    return stateToApiStateContext(config, finalizedState);
```
mpetrunic: we should probably split this method into bunch of utilities functions. I didn't expect it to grow that much tbh haha

packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -92,3 +104,13 @@ export function getEpochBeaconCommittees(
   }
   return committees;
 }
+
+function stateToApiStateContext(config: IBeaconConfig, state: BeaconState | null): ApiStateContext | null {
```
mpetrunic: I don't think we ever need epochCtx for archive state. That why ApiStateContext allows epochCtx to be null. So I don't think we need that method and unfinelized state comes with it anyways

packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -92,3 +104,13 @@ export function getEpochBeaconCommittees(
...
+  if (!state) return null;
+  const epochCtx = new EpochContext(config);
+  epochCtx.loadState(state);
```
mpetrunic: I don't think we ever need epochCtx for archive state. That why ApiStateContext allows epochCtx to be null. So I don't think we need that method and unfinelized state comes with it anyways

packages/lodestar/src/api/rest/controllers/debug/beacon/getStates.ts
```diff
@@ -0,0 +1,51 @@
...
+export const getState: ApiController<DefaultQuery, {stateId: string}> = {
+  url: "/beacon/states/:stateId",
+
+  handler: async function (req, resp) {
+    try {
+      const state = await this.api.debug.beacon.getState(req.params.stateId);
+      if (!state) {
+        return resp.status(404).send();
+      }
+      if (req.headers["accept"] === "application/octet-stream") {
```
mpetrunic: extract to constant or enum.

packages/lodestar/src/api/rest/controllers/debug/beacon/getStates.ts
```diff
@@ -0,0 +1,51 @@
...
+        throw {
+          statusCode: 400,
+          validation: [
+            {
+              dataPath: "state_id",
+              message: e.message,
+            },
+          ],
+        } as FastifyError;
+      }
```
mpetrunic, line 27: Maybe it would be better to change original error to extend from fastifyError and set this details? That said we should stick to official spec error format.

twoeths: the Fastify error handler already set status code to 400 in case of validation error so we don't have to.
I think this is a good place to put in failed validation field and corresponding message, implemented `toRestValidationError()` and applied to other places.

packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -15,38 +25,18 @@ export async function resolveStateId(
...
-  //todo: resolve archive slot -> state
-  const blockSummary = forkChoice.getCanonicalBlockSummaryAtSlot(slot);
-  if (blockSummary) {
-    return (await db.stateCache.get(blockSummary.stateRoot)) ?? null;
+  if (stateId === "head" || stateId === "genesis" || stateId === "finalized" || stateId === "justified") {
+    return await stateByName(db, forkChoice, stateId);
+  } else if (stateId.startsWith("0x")) {
+    return await stateByRoot(db, stateId);
+  } else {
+    //block id must be slot
```
mpetrunic:
```suggestion
    //state id must be slot
```

mpetrunic (commented):
LGTM, just some typo

mpetrunic (approved):

wemeetagain (approved):

PR #1893, opened 2020-12-17
Lightclient upgrade

mpetrunic:
This just provide method to upgrade state. More concerning issue is that we depend on epochCtx from fast transition in a rest of lodestar code and we should use regular one now.

Only solution I see is to add some logic in block processing to generate new epochCtx after state transition. It's slow, but we don't really care when prototyping lightclient with minimal spec. Is that ok?

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/lightclient/index.ts
+++ b/packages/lodestar-beacon-state-transition/src/lightclient/index.ts
@@ -1,3 +1,4 @@
 export * from "./sync_committee";
 export * from "./block";
+export * from "./upgrade";
 export * from "./epoch";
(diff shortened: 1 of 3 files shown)


wemeetagain:
We can always just use the "naive" lightclient/upgrade pieces of the state transition in the fast impl.
Since the naive impl just relies on the state, it should work no problem.

mpetrunic:
> We can always just use the "naive" lightclient/upgrade pieces of the state transition in the fast impl.
> Since the naive impl just relies on the state, it should work no problem.

Lightclient stuff adds additional state balance updates which needs to reflect in epoch ctx

PR #1899, opened 2020-12-18
Lightclient sync

mpetrunic:


packages/lodestar-beacon-state-transition/src/lightclient/sync/update.ts
```diff
@@ -0,0 +1,71 @@
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
```
wemeetagain: We should maybe make a new package for this.
Isn't really part of the "beacon-state-transition"

mpetrunic: Yeah, totally makes sense.

mpetrunic:
I've moved sync stuff to separate package. I reckon we can start adding some browser db etc there and create additional demo package with some react app that would utilize lightclient

PR #1901, opened 2020-12-18
Write request decode validation as async fn

dapplion:
- Move validation logic out of the async generator into a standalone function
- Use LodestarError pattern for validation errors
- Use RequestDecodeErrorCode in tests instead of asserting raw strings

### DIFF
--- a/packages/lodestar/src/network/encoders/request.ts
+++ b/packages/lodestar/src/network/encoders/request.ts
@@ -1,5 +1,5 @@
 import {Method, Methods, ReqRespEncoding} from "../../constants";
-import {ILogger} from "@chainsafe/lodestar-utils";
+import {ILogger, LodestarError} from "@chainsafe/lodestar-utils";
 import {RequestBody} from "@chainsafe/lodestar-types";
 import {decode, encode} from "varint";
 import BufferList from "bl";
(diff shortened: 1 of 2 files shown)


wemeetagain (approved):

packages/lodestar/src/network/encoders/request.ts
```diff
@@ -53,76 +53,123 @@ export function eth2RequestDecode(
...
+
+  for await (let chunk of source) {
+    if (!chunk || chunk.length === 0) {
+      continue;
+    }
 
-      if (sszDataLength === null) {
-        sszDataLength = decode(chunk.slice());
-        if (decode.bytes > 10) {
-          logger.error("eth2RequestDecode: Invalid number of bytes for protobuf varint", {
```
mpetrunic, line 68: IMO having method in log could be useful

packages/lodestar/src/network/encoders/request.ts
```diff
@@ -53,76 +53,123 @@ export function eth2RequestDecode(
...
+    if (buffer.length > sszDataLength) {
+      throw new RequestDecodeError({code: RequestDecodeErrorCode.TOO_MANY_BYTES, sszDataLength});
     }
-  };
+
+    // buffer.length === sszDataLength
+
+    // only one request body accepted, so return
+    try {
+      return type.deserialize(buffer.slice(0, sszDataLength));
```
mpetrunic, line 136: I wonder if you don't break when reading from source, do you even close stream. It's been a while,
but I remember that it could hang open if you don't read everything or break. libp2p handles open and closes in post loop cleanup. I know we had some issues because if you don't finish reading that post iteration cleanup did not occur.

PR #1902, opened 2020-12-21
Get connected peers from connection manager

### LINKED ISSUE #1897: Get all connected peers from libp2p connection manager
Currently we get connected peers by looping through all peers in the peer store and check if we have any connections for that peer and we have a performance issue with it.
In my medalla libp2p peer store, I got more than 800 peers.

We should be able to get that directly from the connection manager using its `connections` map https://github.com/libp2p/js-libp2p/blob/v0.30.0/src/connection-manager/index.js#L92

twoeths:
resolves #1897 

+ get connected peers directly from libp2p connection manager should be faster than looping through every peers in the libp2p store. Note that libp2p peer store may contain peers from testnets as well.
+ `getPeers()` default to get connected peers, added `getAllPeers()` which loop through peer store just for the api
+ correct `getPeer` Rest api: add `state` and `direction` query params
+ improve `searchSubnetPeers` by not loading all known peers from the peer store

Screen Shot 2020-12-24 at 16 21 13

### DIFF
--- a/packages/lodestar/src/api/types/node.ts
+++ b/packages/lodestar/src/api/types/node.ts
@@ -13,5 +13,6 @@ export type NodePeer = {
   enr: string;
   lastSeenP2pAddress: string;
   state: "disconnected" | "connecting" | "connected" | "disconnecting";
-  direction: "inbound" | "outbound" | "NA";
+  // the spec does not specify direction for a disconnected peer, lodestar uses null in that case
+  direction: "inbound" | "outbound" | null;
 };
--- a/packages/lodestar/src/api/types/node.ts
+++ b/packages/lodestar/src/api/types/node.ts
@@ -11,7 +11,7 @@ export type NodeIdentity = {
 export type NodePeer = {
   peerId: string;
   enr: string;
-  address: string;
+  lastSeenP2pAddress: string;
   state: "disconnected" | "connecting" | "connected" | "disconnecting";
-  direction: "inbound" | "outbound";
+  direction: "inbound" | "outbound" | "NA";
 };
--- a/packages/lodestar/src/network/network.ts
+++ b/packages/lodestar/src/network/network.ts
@@ -4,7 +4,7 @@
 
 import {EventEmitter} from "events";
 import LibP2p from "libp2p";
-import PeerId from "peer-id";
+import PeerId, {createFromCID} from "peer-id";
 import Multiaddr from "multiaddr";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {ILogger} from "@chainsafe/lodestar-utils";
@@ -23,6 +23,7 @@ import {IPeerMetadataStore} from "./peers/interface";
 import {Libp2pPeerMetadataStore} from "./peers/metastore";
 import {getPeerCountBySubnet} from "./peers/utils";
 import {IRpcScoreTracker, SimpleRpcScoreTracker} from "./peers/score";
+import {notNullish} from "../util/notNullish";
 
 interface ILibp2pModules {
   config: IBeaconConfig;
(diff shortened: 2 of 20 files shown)


packages/lodestar/package.json
```diff
@@ -101,7 +101,7 @@
     "@types/varint": "^5.0.0",
     "benchmark": "^2.1.4",
     "eventsource": "^1.0.7",
-    "libp2p-ts": "https://github.com/ChainSafe/libp2p-ts.git",
```
twoeths: waiting for https://github.com/ChainSafe/libp2p-ts/pull/19

packages/lodestar/src/network/network.ts
```diff
@@ -113,27 +114,32 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
...
-      });
+    let peers: LibP2p.Peer[];
+    if (opts.connected) {
+      const peerIdStrs = Array.from(this.libp2p.connectionManager.connections.keys());
+      const peerIds = peerIdStrs
+        .map((peerIdStr) => createFromCID(peerIdStr))
+        .filter((peerId) => this.getPeerConnection(peerId));
+      peers = peerIds.map((peerId) => this.libp2p.peerStore.get(peerId)).filter(notNullish);
+    } else {
+      peers = Array.from(this.libp2p.peerStore.peers.values());
```
twoeths: We have a performance issue with this api, I'd like to remove this completely and default to connected peers only if that's ok @mpetrunic

twoeths: fyi I checked that we always use `{connected: true}` when we use this api and in my pyrmont folder I got more than 6600 peers and it takes around 300ms for this call

mpetrunic: hmm don't we use `connected: false` in rest api where we need to display disconnected peers as well?
I'm fine with current code where if we wan't connected only, we use connection manager.

mpetrunic: We could split it in two methods though.

twoeths: I don't see we use `connected: false` in our api https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/api/impl/node/node.ts#L52 , also I don't see the spec states to display connected or disconnected peers, can you double check?

mpetrunic: That might be bug  😄 

Check filter params regarding state:
https://ethereum.github.io/eth2.0-APIs/#/Node/getPeers

twoeths: it's good to know, I'll fix it

packages/lodestar/test/unit/api/impl/node/node.test.ts
```diff
@@ -94,24 +94,51 @@ describe("node api implementation", function () {
...
+      expect(peers[0].lastSeenP2pAddress).not.empty;
+      expect(peers[0].peerId).not.empty;
+      // expect(peers[0].enr).not.empty;
+      expect(peers[0].direction).to.equal("outbound");
+      expect(peers[0].state).to.equal("disconnected");
+      expect(peers[1].lastSeenP2pAddress).to.be.empty;
+      expect(peers[1].peerId).not.empty;
+      // expect(peers[1].enr).not.empty;
+      expect(peers[1].direction).to.equal("NA");
+      expect(peers[1].state).to.equal("disconnected");
```
dapplion: Probably it's not in the scope of this PR but it would be much more helpful to do
```ts
expect(toPeerSummary(peers[0]).to.deep.equal({
  direction: "outbound",
  state: "disconnected",
  hasPeerId: true,
  hasP2pAddress: true
})
```
Otherwise, on failure you have a message like
```
expected null to not be empty
```
that provides no help at all on knowing the actual issue

dapplion: Or do
```ts
function assertPeer(data) {
  expect(peers[0].lastSeenP2pAddress, `peer ${id} has not lastSeenP2pAddress`).not.empty;
  expect(peers[0].peerId, `peer ${id} has no peerId`).not.empty;
  ....
}
```

packages/lodestar/src/network/network.ts
```diff
@@ -112,31 +113,44 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
...
-      });
-
-      if (opts?.supportsProtocols) {
-        for (const protocol of opts.supportsProtocols) {
+    let peers: LibP2p.Peer[];
+    const peerIdStrs = Array.from(this.libp2p.connectionManager.connections.keys());
+    const peerIds = peerIdStrs
+      .map((peerIdStr) => createFromCID(peerIdStr))
+      .filter((peerId) => this.getPeerConnection(peerId));
+    peers = peerIds.map((peerId) => this.libp2p.peerStore.get(peerId)).filter(notNullish);
```
dapplion: You can do `let peers` here instead of declaring above

packages/lodestar/src/network/network.ts
```diff
@@ -112,31 +113,44 @@ export class Libp2pNetwork extends (EventEmitter as {new (): NetworkEventEmitter
...
+
+    if (opts?.supportsProtocols) {
+      const supportsProtocols = opts.supportsProtocols;
+      peers = peers.filter((peer) => {
```
dapplion: Depending on how big the peers array is it might be worth it to join this .filter with the one above

packages/lodestar/src/api/types/node.ts
```diff
@@ -11,7 +11,7 @@ export type NodeIdentity = {
 export type NodePeer = {
   peerId: string;
   enr: string;
-  address: string;
+  lastSeenP2pAddress: string;
   state: "disconnected" | "connecting" | "connected" | "disconnecting";
-  direction: "inbound" | "outbound";
+  direction: "inbound" | "outbound" | "NA";
```
dapplion: What does it mean `NA`? Unknown direction? Do you know under which conditions does it happen? Might be worth to add a comment

mpetrunic: if you are disconnected.

dapplion: Have we decided to use the string `NA` or does it come from libp2p or the specs?

mpetrunic: neither. It isn't covered in api spec so every client did it's own thing :(

dapplion: Then can we use the string `disconnected`?

mpetrunic: I would rather use null.

dapplion: Sounds good, with a comment explaining this nuance

wemeetagain (approved):
lgtm :+1:

PR #1910, opened 2020-12-22
Fix it-pipe types

### LINKED ISSUE #4: Potential ts-ignore effect in current type definition
Current type definitions compile fine with invalid arguments such as
```ts
await pipe(1, "e", 0, null)
```

This can be an issue for strongly typed projects since Typescript won't notify of type issues within it-pipe arguments. Seems that the issue comes from this overload

https://github.com/alanshaw/it-pipe/blob/825195720c72bf40c08335223bfa40ae2624d150/index.d.ts#L82

if the pipe argument types are invalid and don't satisfy any of the above well-crafted types line 82 acts as a sort of `@ts-ignore`.

While this final overload allows for relaxed usage it's a strong cause of concern for Typescript consumers

CC: @mpetrunic @wemeetagain

dapplion:
Overwrite types until this issue is fixed https://github.com/alanshaw/it-pipe/issues/4
Original types contain
```
export function pipe (first: any, ...rest: any[]): any
```
Which act as a @ts-ignore compromising type safety

### DIFF
--- a/packages/lodestar/src/network/reqresp/reqResp.ts
+++ b/packages/lodestar/src/network/reqresp/reqResp.ts
@@ -59,8 +59,9 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements 
       for (const encoding of Object.values(ReqRespEncoding)) {
         this.libp2p.handle(createRpcProtocol(method, encoding), async ({connection, stream}) => {
           const peerId = connection.remotePeer;
+          // eslint-disable-next-line @typescript-eslint/no-floating-promises
           pipe(
-            stream.source,
+            stream.source as AsyncIterable<Buffer>,
             eth2RequestDecode(this.config, this.logger, method, encoding),
             this.storePeerEncodingPreference(peerId, method, encoding),
             this.handleRpcRequest(peerId, method, encoding, stream.sink)
--- a/packages/lodestar/src/network/reqresp/reqResp.ts
+++ b/packages/lodestar/src/network/reqresp/reqResp.ts
@@ -59,8 +59,7 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements 
       for (const encoding of Object.values(ReqRespEncoding)) {
         this.libp2p.handle(createRpcProtocol(method, encoding), async ({connection, stream}) => {
           const peerId = connection.remotePeer;
-          // eslint-disable-next-line @typescript-eslint/no-floating-promises
-          pipe(
+          void pipe(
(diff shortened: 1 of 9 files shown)


packages/lodestar/src/network/reqresp/reqResp.ts
```diff
@@ -59,8 +59,9 @@ export class ReqResp extends (EventEmitter as IReqEventEmitterClass) implements
       for (const encoding of Object.values(ReqRespEncoding)) {
         this.libp2p.handle(createRpcProtocol(method, encoding), async ({connection, stream}) => {
           const peerId = connection.remotePeer;
+          // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
mpetrunic: With new eslint, you can just add "void" in front of promise you don't wan't to await

packages/lodestar/src/network/reqresp/reqUtils.ts
```diff
@@ -87,10 +87,10 @@ export async function* sendRequestStream<T extends ResponseBody>(
   logger.verbose("sent request", {peer: peerId.toB58String(), method});
 
   yield* pipe(
-    abortDuplex(conn!.stream, controller.signal, {returnOnAbort: true}),
+    (abortDuplex(conn!.stream, controller.signal, {returnOnAbort: true}) as unknown) as AsyncIterable<Buffer>,
```
mpetrunic: maybe we can add better types for abort duplex?

dapplion: Agree, but would you mind opening an issue for it? I think it's out the scope of this PR

ams-201062 (dismissed):

packages/lodestar/src/network/reqresp/reqUtils.ts
mpetrunic: But if you don't fix it, isn't it basically same thing as before(if something changes it will happily cast it into type that fits)? 😄

dapplion: I'm not adding type unsafety, this change just shows what was already there. The goal of the PR is to allow future work to be type safe. Same thing when turning strictNullChecks on, the goal of the work was not to fix existing issues but to allow new code to work with the rule

mpetrunic:
As I sadi in chat. I think it would better to upstream proper types to it-pipe instead of hacking code and types like this just to make it work. IMO there is no rush to change this as this code works (without full typesafety) and we won't be required to change anything anytime soon so we can wait to get better it-pipe types upstream.

PR #1911, opened 2020-12-22
Enable Typescript strict opts

dapplion:
Enable remaining strict options in Typescript with the exception of `noUnusedParameters`, which should be taken care of by eslint
- If we agree to enable them I'll start fixing the issues

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/phase0/fast/util/cachedBeaconState.ts
+++ b/packages/lodestar-beacon-state-transition/src/phase0/fast/util/cachedBeaconState.ts
@@ -92,6 +92,8 @@ export const CachedBeaconStateProxyHandler: ProxyHandler<CachedBeaconState<allFo
     } else if (key in target.type.tree) {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       return (target.type.tree as any)[key].bind(target.type.tree, target.tree);
+    } else {
+      return undefined;
     }
   },
   set(target: CachedBeaconState<allForks.BeaconState>, key: string, value: unknown): boolean {
--- a/packages/lodestar-spec-test-util/src/single.ts
+++ b/packages/lodestar-spec-test-util/src/single.ts
@@ -107,7 +107,8 @@ function generateTestCase<TestCase, Result>(
   it(name, function () {
     const testCase = loadInputFiles(testCaseDirectoryPath, options);
     if (options.shouldSkip && options.shouldSkip(testCase, name, index)) {
-      return this.skip();
+      this.skip();
+      return;
     }
     if (options.shouldError && options.shouldError(testCase)) {
       try {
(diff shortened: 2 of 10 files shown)


tsconfig.json
```diff
@@ -7,24 +7,22 @@
...
-    /* Additional Checks */
-    // "noUnusedLocals": true,                /* Report errors on unused locals. */
+    "strictFunctionTypes": true /* Enable strict checking of function types. */,
+    "strictBindCallApply": true /* Enable strict 'bind', 'call', and 'apply' methods on functions. */,
+    "strictPropertyInitialization": true /* Enable strict checking of property initialization in classes. */,
+    "noImplicitAny": true /* Raise error on expressions and declarations with an implied 'any' type. */,
+    "noImplicitThis": true /* Raise error on 'this' expressions with an implied 'any' type. */,
+    "noImplicitReturns": true /* Report error when not all code paths in function return a value. */,
+    "noFallthroughCasesInSwitch": true /* Report errors for fallthrough cases in switch statement. */,
+    "noUnusedLocals": true /* Report errors on unused locals. */,
```
mpetrunic: that should be checked by eslint

tsconfig.json
```diff
@@ -7,24 +7,22 @@
...
-
-    /* Additional Checks */
-    // "noUnusedLocals": true,                /* Report errors on unused locals. */
+    "strictFunctionTypes": true /* Enable strict checking of function types. */,
+    "strictBindCallApply": true /* Enable strict 'bind', 'call', and 'apply' methods on functions. */,
+    "strictPropertyInitialization": true /* Enable strict checking of property initialization in classes. */,
+    "noImplicitAny": true /* Raise error on expressions and declarations with an implied 'any' type. */,
+    "noImplicitThis": true /* Raise error on 'this' expressions with an implied 'any' type. */,
+    "noImplicitReturns": true /* Report error when not all code paths in function return a value. */,
+    "noFallthroughCasesInSwitch": true /* Report errors for fallthrough cases in switch statement. */,
```
mpetrunic: this one is a bit annoying

dapplion: Since we use a lot of enums + switch I think it can bring more value than pain. You can always use a default if necessary

tsconfig.json
dapplion: True, I guess adding it here adds unnecessary compile time

tsconfig.json
mpetrunic: it prevents grouping like:
```
Enum.ONE:
Enum.TWO: doSomething(); break;
Enum.THREE:
Enum.FOUR: doSomethingElse(); break;
default: throw new Error();
```
I don't really see what value it brings, never seen project with that flag enabled tbh haha If you enable it using if-else makes more sense than switch-case

dapplion:
So is there a consensus on turning some of these rules on?

dapplion:
Review after merging https://github.com/ChainSafe/lodestar/pull/1924 should fix the type issues breaking CI

mpetrunic (dismissed):

wemeetagain:
dang, one more file conflict

wemeetagain (approved):

PR #1915, opened 2020-12-23
Add network test vectors

dapplion:
The purpose of this PR is to ensure that our ReqResp code behaves exactly the same from version to version. Current unit tests only assert that the current encoder works with the current decoder, but doesn't guarantee correctness. This PR leverages the fact that our current code has been widely tested and interop-ed so it's treated as correct. I'm adding this now to safely incorporate my upcoming ReqResp code re-org.

Tests are structured resembling the eth2 spec tests, but are generated with our own code. Include three directories
- `network/spec-test/format`: write test vectors to FS, parse FS to test vectors logic
- `network/spec-test/generate`: create test vectors and store to FS. Should be run rarely run, i.e. when the spec changes and we are confident our code interops
- `network/spec-test/run`: read test vectors from FS and run them
- `network/spec-test/vectors`: test vectors

Currently this PR only includes tests for decoding a response given a stream of bytes. I will extend to response encoding, request encoding / decoding and errors if this structure is good with other repo mantainers


wemeetagain (approved):
LGTM
I think this is a good idea

mpetrunic (commented):
I don't see any error test case?

dapplion:
Superseded by https://github.com/ChainSafe/lodestar/pull/1930 [`packages/lodestar/test/unit/network/reqresp/encodingStrategies/sszSnappy/testData.ts`](https://github.com/ChainSafe/lodestar/pull/1930/files#diff-06c33d8dfa5e808f8568447465c16b6765f91b864a29d23cdc22cfc501d8fcc6)

Above PR includes serialized payloads for Ping, Status and SignedBeaconBlock for ssz_snappy, captured in current master.

PR #1919, opened 2020-12-25
Fix out-of-sync node due to timed out beacon_blocks_by_root

### LINKED ISSUE #1916: Node is out of sync
Node stays out of sync, gossip blocks received frequently, number of peers is 25. I see a lot of logs like this:
```
Ignoring gossip block blockRoot=0x6ca67ba20f940f25e5d6        64af590b0f50f7537c9e204c25253ab6204cec8806d1, blockSlot=148684 code=BLOCK_ERROR_PARENT_UNKNOWN, parentRoot={}
   5334 Error: BLOCK_ERROR_PARENT_UNKNOWN
```

Ideally the sync service should issue "beacon_blocks_by_root" request and node should be synced again, need to turn debug mode on to get more information.

twoeths:
not the root cause of #1916 but this could prevent same issue

+ if a beacon_blocks_by_root request is timed out, the same missing ancestor root stays in sync cache forever and node will be out of sync due to that
+ this change handle the `processingRoots` inside the sync as a Map (with key as root hex, value as timestamp) instead of Set. If it's in the missing root for more than 1 minute, we allow to retry finding the same root.

### DIFF
--- a/packages/lodestar/src/sync/sync.ts
+++ b/packages/lodestar/src/sync/sync.ts
@@ -15,6 +15,7 @@ import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {List, toHexString} from "@chainsafe/ssz";
 import {BlockError, BlockErrorCode} from "../chain/errors";
 import {ORARegularSync} from "./regular/oneRangeAhead/oneRangeAhead";
+import {notNullish} from "../util/notNullish";
 
 export enum SyncMode {
   WAITING_PEERS,
@@ -41,7 +42,8 @@ export class BeaconSync implements IBeaconSync {
   private statusSyncTimer?: NodeJS.Timeout;
   private peerCountTimer?: NodeJS.Timeout;
   // avoid finding same root at the same time
-  private processingRoots: Set<string>;
+  // key as root hex and value as timestamp
+  private processingRoots: Map<string, number>;
 
   constructor(opts: ISyncOptions, modules: ISyncModules) {
     this.opts = opts;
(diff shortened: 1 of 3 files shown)


packages/lodestar/src/sync/sync.ts
```diff
@@ -212,10 +214,13 @@ export class BeaconSync implements IBeaconSync {
     const blockRoot = this.config.types.BeaconBlock.hashTreeRoot(err.job.signedBlock.message);
     const unknownAncestorRoot = this.chain.pendingBlocks.getMissingAncestor(blockRoot);
     const missingRootHex = toHexString(unknownAncestorRoot);
-    if (this.processingRoots.has(missingRootHex)) {
+    const MAX_TIME_IN_CACHE = 60 * 1000;
+    const lastTimeInCache = this.processingRoots.get(missingRootHex);
+    // in case beaconBlocksByRoot timeout, we can retry after 1 minute
+    if (notNullish(lastTimeInCache) && Date.now() - lastTimeInCache < MAX_TIME_IN_CACHE) {
```
mpetrunic: maybe better to use some lru cache implementation or write our own generic one that can be used here

twoeths: gossipsub-js uses `time-cache`, we should use it here too

packages/lodestar/src/sync/sync.ts
```diff
@@ -246,7 +247,22 @@ export class BeaconSync implements IBeaconSync {
...
+   * Check if we should process a missing root.
+   * Time-cache does not do the sweep() automatically,
+   * it only does that per put() call.
+   */
+  private shouldProcessRoot(missingRootHex: string): boolean {
+    if (this.processingRoots.has(missingRootHex)) {
+      // time-cache only do sweep() in this case
+      this.processingRoots.put(missingRootHex);
+      // check again to see if the item is there after the sweep()
+      return !this.processingRoots.has(missingRootHex);
```
twoeths, line 263: this is the sweep() implementation: https://github.com/daviddias/time-cache/blob/master/src/index.js#L27

3xtr4t3rr3str14l (approved):

twoeths:
this doesn't seem like we have timedout req/resp and we're refactoring the sync so closing for now

PR #1920, opened 2020-12-26
Fix out-of-sync node due to FUTURE_SLOT error

### LINKED ISSUE #1916: Node is out of sync

twoeths:
resolves #1916 

+ as I see `BLOCK_ERROR_FUTURE_SLOT` quite frequently, I think we need to accept some deltas when determining that error due to time differences. This is part of the spec and was somehow removed after some code refactoring
+ when handling that error, we should add to the cache first then add to `pendingBlock` db
+ at `onClockSlot` we should get all pending blocks of past slots too and process all of them
+ minor point: improve `Block error` debug log to include block slot as well

### DIFF
--- a/packages/lodestar/src/chain/clock/LocalClock.ts
+++ b/packages/lodestar/src/chain/clock/LocalClock.ts
@@ -2,11 +2,12 @@ import {AbortSignal} from "abort-controller";
 import {Epoch, Slot} from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {ErrorAborted} from "@chainsafe/lodestar-utils";
-import {computeEpochAtSlot, getCurrentSlot} from "@chainsafe/lodestar-beacon-state-transition";
+import {computeEpochAtSlot, computeTimeAtSlot, getCurrentSlot} from "@chainsafe/lodestar-beacon-state-transition";
 
 import {ChainEvent, ChainEventEmitter} from "../emitter";
 
 import {IBeaconClock} from "./interface";
+import {MAXIMUM_GOSSIP_CLOCK_DISPARITY} from "../../constants";
 
 /**
  * A local clock, the clock time is assumed to be trusted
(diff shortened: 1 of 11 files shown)


packages/lodestar/src/chain/clock/LocalClock.ts
```diff
@@ -43,6 +44,17 @@ export class LocalClock implements IBeaconClock {
     return getCurrentSlot(this.config, this.genesisTime);
   }
 
+  /**
+   * Maximum currentSlot of peers.
+   * If it's too close to next slot given MAXIMUM_GOSSIP_CLOCK_DISPARITY, return currentSlot + 1.
+   * Otherwise return currentSlot
+   */
+  public get maxPeerCurrentSlot(): Slot {
```
mpetrunic: I don't get how is this max slot of peers. I guess you meant "wall clock"?

twoeths: yeah it's related to that. When we're too close to next slot, other peers may switched to next slot already and it submits/spread the new block to the network and our validation failed if we use our own `currentSlot`. Since the spec allow a `MAXIMUM_GOSSIP_CLOCK_DISPARITY`, I implement this method accordingly.

The method name may be confusing, any recommendation?

mpetrunic: maybe networkWallClock or assumedWallClock? Not sure

wemeetagain: Imo `currentSlotWithGossipDisparity`

3xtr4t3rr3str14l:
@tuyennhv  Just curious, have you seen incoming gossip be errored with FUTURE_SLOT erroneously?  In the example you have here in the original issue (https://github.com/ChainSafe/lodestar/issues/1916#issuecomment-751333824), it looks like the time difference is about 5 seconds, which is beyond the MAXIMUM_GOSSIP_CLOCK_DISPARITY and should be marked as a FUTURE_SLOT.

twoeths:
> @tuyennhv Just curious, have you seen incoming gossip be errored with FUTURE_SLOT erroneously? In the example you have here in the original issue ([#1916 (comment)](https://github.com/ChainSafe/lodestar/issues/1916#issuecomment-751333824)), it looks like the time difference is about 5 seconds, which is beyond the MAXIMUM_GOSSIP_CLOCK_DISPARITY and should be marked as a FUTURE_SLOT.

@3xtr4t3rr3str14l the error `BLOCK_ERROR_FUTURE_SLOT` happened a lot in the log

wemeetagain (approved):

PR #1924, opened 2021-01-01
Bump Typescript to v4

### LINKED ISSUE #1784: Bump Typescript to 4.1
Should we?

----

There are some type issues to fix in some packages after quickly checking

dapplion:
Fixes type breaking changes to be compatible with Typecript v4, so we can benefit from the performance improvements and new features.

See https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/
Fixes https://github.com/ChainSafe/lodestar/issues/1784

### DIFF
--- a/packages/lodestar-utils/src/bytes.ts
+++ b/packages/lodestar-utils/src/bytes.ts
@@ -1,5 +1,4 @@
 import {toBufferLE, toBigIntLE, toBufferBE, toBigIntBE} from "bigint-buffer";
-import {ArrayLike} from "@chainsafe/ssz";
 
 type Endianness = "le" | "be";
 
@@ -35,7 +34,7 @@ export function bytesToBigInt(value: Uint8Array, endianness: Endianness = "le"):
   throw new Error("endianness must be either 'le' or 'be'");
 }
 
-export function toHex(buffer: ArrayLike<number>): string {
+export function toHex(buffer: Parameters<typeof Buffer.from>[0]): string {
   if (Buffer.isBuffer(buffer)) {
     return "0x" + buffer.toString("hex");
   } else if (buffer instanceof Uint8Array) {
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "@types/chai": "4.2.0",
     "@types/chai-as-promised": "^7.1.2",
     "@types/mocha": "^8.0.3",
-    "@types/node": "^12.12.21",
+    "@types/node": "^14.14.17",
     "@types/sinon": "^9.0.0",
     "@typescript-eslint/eslint-plugin": "4.9.0",
     "@typescript-eslint/parser": "4.9.0",
@@ -62,12 +62,12 @@
     "prettier": "^2.0.5",
     "sinon": "^9.0.2",
     "supertest": "^4.0.2",
-    "ts-node": "^9.0.0",
+    "ts-node": "^9.1.1",
     "typedoc": "^0.19.2",
     "typedoc-plugin-external-module-name": "^4.0.3",
     "typedoc-plugin-internal-external": "^2.2.0",
     "typedoc-plugin-markdown": "^2.4.0",
-    "typescript": "^3.8.3",
+    "typescript": "^4.1.3",
     "webpack": "^4.42.0"
   },
   "dependencies": {}
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "typedoc-plugin-external-module-name": "^4.0.3",
     "typedoc-plugin-internal-external": "^2.2.0",
     "typedoc-plugin-markdown": "^2.4.0",
-    "typescript": "^4.1.3",
+    "typescript": "^4.2.0",
     "webpack": "^4.42.0"
   },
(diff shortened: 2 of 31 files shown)


dapplion:
A lot of unit tests are failing, anyone has a clue what might have changed to cause so many failures?

packages/lodestar-utils/src/bytes.ts
```diff
@@ -35,14 +34,6 @@ export function bytesToBigInt(value: Uint8Array, endianness: Endianness = "le"):
   throw new Error("endianness must be either 'le' or 'be'");
 }
 
-export function toHex(buffer: ArrayLike<number>): string {
```
mpetrunic, line 38: Can you put this method back. We should use this method instead of Buffer.from().toHex().

This change doesn't belong in this PR anyways

mpetrunic (commented):
Only breaking changes that could cause tests to fail are:
- https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#getterssetters-are-no-longer-enumerable
- https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#parsing-differences-in-optional-chaining-and-non-null-assertions

You should probably debug one tests and detect line thats causing trouble

wemeetagain:
Seems the stubs are no longer working

Eg:
in lodestar-beacon-state-transition/test/unit/stateTransition/epoch/justification.test.ts
set breakpoint on L43
getCurrentEpochStub.called === false

Seems related to the getters no longer being enumerable?
https://github.com/microsoft/TypeScript/issues/38568
@mpetrunic any thoughts?

mpetrunic:
> Seems the stubs are no longer working
> 
> Eg:
> in lodestar-beacon-state-transition/test/unit/stateTransition/epoch/justification.test.ts
> set breakpoint on L43
> getCurrentEpochStub.called === false
> 
> Seems related to the getters no longer being enumerable?
> [microsoft/TypeScript#38568](https://github.com/microsoft/TypeScript/issues/38568)
> @mpetrunic any thoughts?

Typical response from Ryan...
There is nice tread on sinon repo https://github.com/sinonjs/sinon/issues/2258

I think it worked before because typescript didn't generated esm compliant code (aka module imports were mutable). If my memory serves me correctly, babel was already doing that so we switched from babel-register to ts-node/register.

Few ideas:
- try this for running tests https://github.com/jasmine/jasmine/issues/1817#issuecomment-630110087
- switch to jest - it hacks into module loading system and allows stubbing but I think it also caches loaded modules
- switch back to babel-register with plugin like this that would transpile modules to be mutable for test only https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs

dapplion:
We could use an import mocking library such as https://www.npmjs.com/package/rewiremock

I've used it with success and offers full type support. Example:

```ts
  const { default: getPortsToOpen } = await rewiremock.around(
      () => import("../../../src/watchers/natRenewal/getPortsToOpen"),
      mock => {
        mock(() => import("../../../src/modules/docker/listContainers"))
          .with({ listContainers })
          .toBeUsed();
      }
    );
```

From https://github.com/dappnode/DNP_DAPPMANAGER/blob/cbd7cbcf3e155f0dd7e9e969ee37b431fb916819/packages/dappmanager/test/watchers/natRenewal/getPortsToOpen.test.ts#L74

mpetrunic:
> We could use an import mocking library such as https://www.npmjs.com/package/rewiremock
> 
> I've used it with success and offers full type support. Example:
> 
> ```ts
>   const { default: getPortsToOpen } = await rewiremock.around(
>       () => import("../../../src/watchers/natRenewal/getPortsToOpen"),
>       mock => {
>         mock(() => import("../../../src/modules/docker/listContainers"))
>           .with({ listContainers })
>           .toBeUsed();
>       }
>     );
> ```
> 
> From https://github.com/dappnode/DNP_DAPPMANAGER/blob/cbd7cbcf3e155f0dd7e9e969ee37b431fb916819/packages/dappmanager/test/watchers/natRenewal/getPortsToOpen.test.ts#L74
> 

Im pretty sure that only works because you are using typescript <3.9. All those libs were just modifying exports. Feel free to try it though

dapplion:
> Im pretty sure that only works because you are using typescript <3.9. All those libs were just modifying exports. Feel free to try it though

RIght, I have not tried it with Typescript >= 3.9

dapplion:
rewiremock works fine with Typescript v4.
- I've skipped all tests in the beacon state transition that were mocking imports
- I've refactored the tests in lodestar that were mocking imports with rewiremock. It's not pretty, and slow but allows us to move forward.

All our code does dependency injection everywhere, where we pass entire instances of the `chain`, `network`, etc. However, we don't do dependency injection for BLS and state transition utils. To fix this type of issues with mocking imports that result in very ugly verbose test code we could also do dependency injection for these remaining items

wemeetagain (approved):
This is good for now to move forward.
I think ideally, we can build out better testing infrastructure that can help us test on valid-ish states so we can avoid relying on mocking and just have better e2e tests.

PR #1925, opened 2021-01-02
Speed up epoch transition using persistent vector

### LINKED ISSUE #1761: Improve validator exit process

### LINKED ISSUE #1789: POC: Create a StateTransitionBeaconState

### LINKED ISSUE #1878: Explore Epoch Transition Speedups
Profile epoch transition, explore what can be done to speed it up.
Eg: Create flattened validators with fast iteration through all validator fields at once.

twoeths:
resolves #1789, resolves #1878, and resolves #1761

+ Create a CachedValidatorsBeaconState that cache validators using persistent vector (see https://github.com/tuyennhv/persistent-ts/pull/1) and use it everywhere
+ Persistent vector is an immutable data structure, similar to the persistent-merkle-tree we used: it shares a lot of data across beacon state since validators are rarely changed and it has cheap `clone()`
+ Persistent vector loop speed is comparable to regular javascript array loop while it's slow if we loop through `TreeBacked<BeaconState>.validators`
+ When we want to mutate validators inside state, use `setValidator()` and `addValidator()` instead of accessing `validators` directly
+ Average epoch transition time in my local environment for Pyrmont sync was reduced from `3.5s-4.2s` to `1.2s-1.6s`. This should speed up the validator exit process too.
## Testing
+ Pyrmont: synced from slot 0 to head and stay synced for more than 4 hours (tested 2 times)
+ Don't see any memory differences during Pyrmont sync
+ Performance tests

|test|master|this branch|
|----|------|-----------|
|prepareEpochProcessState|1500|100|
|process block of 16 validator exits|6000|200|
|transition - 32 slots|2800|1100|
|transition - 64 slots|5300|2200|
|transition - 128 slots|11000|4300|

## TODO
+ Persistent vector was started from https://github.com/cronokirby/persistent-ts, we should consider having a separate repository for it or just have a separate package like in this PR

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/fast/util/interface.ts
+++ b/packages/lodestar-beacon-state-transition/src/fast/util/interface.ts
@@ -1,6 +1,8 @@
 import {IReadonlyEpochShuffling} from ".";
-import {ValidatorIndex, Slot} from "@chainsafe/lodestar-types";
-import {ByteVector} from "@chainsafe/ssz";
+import {ValidatorIndex, Slot, BeaconState, Validator} from "@chainsafe/lodestar-types";
+import {ByteVector, readOnlyForEach} from "@chainsafe/ssz";
+import {createIFlatValidator, IFlatValidator} from "./flatValidator";
+import {config} from "@chainsafe/lodestar-config/lib/presets/mainnet";
 
 /**
  * Readonly interface for EpochContext.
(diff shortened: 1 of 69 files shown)


packages/lodestar-beacon-state-transition/src/fast/util/interface.ts
```diff
@@ -12,3 +15,128 @@ export type ReadonlyEpochContext = {
...
+ * Instead of accessing `validators` array directly inside BeaconState, use:
+ * + flatValidators() for the loop
+ * + setValidator() for an update
+ * + addValidator() for a creation
+ * that'd update both the cached validators and the one in the original state.
+ */
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export interface CachedValidatorsBeaconState extends BeaconState {
+  flatValidators(): Vector<IFlatValidator>;
+  setValidator(i: ValidatorIndex, value: Partial<IFlatValidator>): void;
```
mpetrunic: updateValidator?

packages/lodestar-beacon-state-transition/src/fast/util/interface.ts
```diff
@@ -12,3 +15,128 @@ export type ReadonlyEpochContext = {
...
+}
+
+/**
+ * Looping through validators inside TreeBacked<BeaconState> is so expensive.
+ * Cache validators from TreeBacked<BeaconState> using a persistent vector for efficiency.
+ * When write, write to both the cache and TreeBacked.
+ * When read, just return the cache.
+ */
+export class CachedValidatorsBeaconState {
+  // this is for the proxy handler, not for the external use
```
mpetrunic: why not make it private then?

packages/lodestar-beacon-state-transition/src/fast/util/interface.ts
```diff
@@ -12,3 +15,128 @@ export type ReadonlyEpochContext = {
...
+      this._cachedValidators = this._cachedValidators.set(i, {...validator!, ...value});
+    }
+    const validator = this._state.validators[i];
+    if (value.activationEligibilityEpoch !== undefined)
+      validator.activationEligibilityEpoch = value.activationEligibilityEpoch;
+    if (value.activationEpoch !== undefined) validator.activationEpoch = value.activationEpoch;
+    if (value.effectiveBalance !== undefined) validator.effectiveBalance = value.effectiveBalance;
+    if (value.exitEpoch !== undefined) validator.exitEpoch = value.exitEpoch;
+    if (value.slashed !== undefined) validator.slashed = value.slashed;
+    if (value.withdrawableEpoch !== undefined) validator.withdrawableEpoch = value.withdrawableEpoch;
```
mpetrunic: Can't we replace all those if-elses with `Object.assign(validator, value)`. This won't create new instance and it will automatically update it in array? 

It also reduces chance that we forget to update this in case some new prop comes along

packages/lodestar-beacon-state-transition/src/fast/util/persistentVector.ts
```diff
@@ -0,0 +1,332 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+const BIT_WIDTH = 5;
```
mpetrunic: maybe move this to separate package in monorepo so it's easier to move to separate repository?

packages/lodestar-beacon-state-transition/src/fast/util/persistentVector.ts
```diff
@@ -0,0 +1,332 @@
...
+
+function isFullBranch(length: number): boolean {
+  return (
+    // initially we initialize Vector with an empty branch (DEFAULT_LEVEL_SHIFT)
+    // length === 1 << 5 ||
+    length === 1 << 10 || length === 1 << 15 || length === 1 << 20 || length === 1 << 25 || length === 1 << 30
+  );
+}
+
+interface VLeaf<T> {
```
mpetrunic: What's up with V prefix. Aren't we using I everywhere?

packages/lodestar-beacon-state-transition/src/fast/util/interface.ts
twoeths: thanks that's a good ideas, it works (I thought it doesn't initially so I have to check each property), unit tests passed.

packages/lodestar-beacon-state-transition/src/fast/util/persistentVector.ts
twoeths: V may stands for `Vector`, that's the upstream implementation, will fix it

mpetrunic (approved):

dapplion (approved):
Awesome work! Looks great. I would vote for `@chainsafe/persistent-ts` to be eventually spin-out into a different repo

wemeetagain (approved):
This is great, we should merge for now.

Going forward, we should take some time to possibly re-organize the beacon state, validator cache, epoch context, epoch process, etc. as needed.

This is mainly an issue of how we organize the various datastructures (`BeaconState`, `CachedValidatorsBeaconState`, `EpochContext`, `EpochProcess`, `IStateContext`). I'm not sure what the right organization is (possibly no re-organization needed) but what we have seems to be getting more cluttered as we've added more and more optimizations. We have several nested datastructures that are all related to the beacon state / epoch state.

My hunch is that we may be able to expand the `CachedValidatorBeaconState` to be more of a general `CachedBeaconState` at the 'top-level', like `IStateContext`, which exposes the `BeaconState` + extra methods, with various caches 'under the hood' (each cache using 'persistent-ts' or similar for cheap cloning).

PR #1930, opened 2021-01-06
Network ReqResp modularization

dapplion:
Refactors network reqResp code with the general goals of modularization and better maintainability. The overall logic is not changed with a few exceptions outlined below.

- Do not rely on events to handle reqResp responses. Instead, the Sync registers a `ReqRespHandler` to network's ReqResp instance
  - Allows keeping only the app-level logic in Sync: given a request, yield a response.
  - Error handling is done exclusively in network's ReqResp, any error on any method will return error to the requester and close the stream
```ts
type ReqRespHandler = (
  method: Method, 
  requestBody: RequestBody,
  peerId: PeerId
) => AsyncIterable<ResponseBody>
```
- De-deduplicate ssz-snappy codec logic. Allows to be recycled in requestEncode and responseEncode
  - Current organization allows to easily add another encoding strategy with minimal logic duplication
  - Use `LodestarError` style with rich documented errors
- Abstract response timeout handling in separate function `responseTimeoutsHandler` 
  - RESP_TIMEOUT waits for an entire response_chunk as indicated by the spec, instead of just a raw bytes chunk from libp2p
- Only log in the main response or request handler, other functions either return or throw
  - Reduces the modules that need to be passed around and simplifies flow control
- Document non-obvious code decisions and general function purpose with content from the p2p spec
- For ssz list requests, limits the number of responses collected to the requested amount. Prevents a malicious node from sending us infinite blocks and DOS us.

This branch is currently syncing Pyrmont in `lodestar-cloud-02` with a stable peer count. I has also synced Pyrmont on my local machine without issues.

Happy to jump on a call if more context is needed by a reviewer.


wemeetagain (approved):
Great PR! I think this is long-overdue refactor of the req/resp code.

I checked this out and saw a lot of positive things:
- significantly more comprehensive comments
- encoding/decoding logic pulled into pure functions (rather than embedded in fns that return async generators)
- more correct handling of req/resp timeout and response length limits
- fully defined request/resp error codes/types
- simplified/linearized req/resp flows

Things like the types can be pushed upstream (eg: it-pipe and libp2p) in the future.

PR #1934, opened 2021-01-11
Process chain segment

### LINKED ISSUE #1772: Add block range API to chain

twoeths:
resolves #1772 

+ new api `chain.processChainSegment()` to process a linear chain until completed
+ Regular Sync to use `chain.processChainSegment()`, removed its own processor
+ Tested with Pyrmont: synced to head 2 times
```
2021-01-11 16:09:24 [SYNC]             info: Regular Sync: processed up to current slot slot=387946
2021-01-11 16:09:33 [SYNC]             info: Regular Sync: wait for best peer
2021-01-11 16:09:33 [SYNC]             info: Regular Sync: fetcher returns empty array, finish sync now
```


packages/lodestar/src/chain/blocks/process.ts
```diff
@@ -3,42 +3,46 @@ import {IForkChoice} from "@chainsafe/lodestar-fork-choice";
...
   emitter: ChainEventEmitter;
   db: IBeaconDb;
-  job: IBlockJob;
+  jobs: IBlockJob[];
 }): Promise<void> {
+  let preStateContext: ITreeStateContext;
   try {
-    const preStateContext = await regen.getPreState(job.signedBlock.message);
-    await runStateTransition(emitter, forkChoice, db, preStateContext, job);
+    preStateContext = await regen.getPreState(jobs[0].signedBlock.message);
```
dapplion, line 26: An alternative more succinct syntax would be
```ts
const x = await getX().catch(e => {
  throw WrapperError(e)
})
```

packages/lodestar/src/chain/blocks/process.ts
```diff
@@ -3,42 +3,46 @@ import {IForkChoice} from "@chainsafe/lodestar-fork-choice";
...
   } catch (e) {
-    if (e instanceof RegenError) {
+    throw new BlockError({
+      code: BlockErrorCode.PRESTATE_MISSING,
+      job: jobs[0],
+    });
+  }
+  for (const job of jobs) {
```
dapplion, line 34: Suggestion to use whitespace before this for block

dapplion (dismissed):

packages/lodestar/src/chain/interface.ts
```diff
@@ -105,4 +112,8 @@ export interface IBeaconChain {
    * Pre-process and run the per slot state transition function
    */
   receiveBlock(signedBlock: SignedBeaconBlock, trusted?: boolean): Promise<void>;
```
dapplion, line 114: Can we change this to return `void` so it's not misleading? the behaviour is very different to `processChainSegment`

packages/lodestar/src/chain/blocks/processor.ts
```diff
@@ -58,14 +61,28 @@ export class BlockProcessor {
...
+  try {
+    const blockJobs: IBlockJob[] = chainSegmentJob.signedBlocks.map((signedBlock) => ({
+      signedBlock,
+      ...chainSegmentJob,
+    }));
+    await validateBlocks({...modules, jobs: blockJobs});
+    await processBlocks({...modules, jobs: blockJobs});
   } catch (e) {
     // above functions only throw BlockError
     modules.emitter.emit(ChainEvent.errorBlock, e);
```
dapplion, line 88: I think this should throw an error here, otherwise consumers of `processChainSegment` won't know if it was successful or not

packages/lodestar/src/chain/interface.ts
twoeths: I'd say `yes`, would like @wemeetagain to confirm through. I can to create a separate PR if we want to do that.

at least what I see is the name `receive*` vs. `process*` is the main difference between these 2 apis.

dapplion (dismissed):

dapplion (approved):

PR #1937, opened 2021-01-12
Store state to db only every n epochs

### LINKED ISSUE #1931: Store state only every N epochs
Currently lodestar takes a lot of disk space, because it stores every new finalized state to disk.

An alternative would be to store states to disk every N epochs, or at most every N epochs. (I think some clients store as few as 1 state per 8196 epochs)

There's some discussion to be had around how specifically this should/could be accomplished. Here are two ideas.

Proposal 1:
Track epoch of latest stored state in `stateArchive`
On new finalized state: if newEpoch - latestStoredEpoch > N, store new state in `stateArchive`

On boot (eg: inside `BeaconNode#init`) feed any additional archived blocks into the chain.

Proposal 2:
Create a new db bucket or special entry `latestFinalizedState`
On new finalized state, always overwrite `latestFinalizedState`
Track epoch of latest stored state in `stateArchive`
On new finalized state: if newEpoch - latestStoredEpoch > N, store new state in `stateArchive`

On boot, use `latestFinalizedState` as anchor state to init the node.

--
Would like to get feedback / consensus on a direction before implementation.

wemeetagain:
Resolves #1931 

Stores state to `stateArchive` at most every `MIN_EPOCHS_PER_DB_STATE` epochs, currently set to 1024.
On shutdown, store the latest finalized state to `stateArchive`.

Reusing the `stateArchive` bucket for the latest finalized, rather than creating a new bucket, avoids any potential issues with deserializing states before/after forks.

With N set to 1024, syncing mainnet, disk usage is ~1GB.

Possible enhancements:
- expose `MIN_EPOCHS_PER_DB_STATE` as CLI-configurable
- occasionally prune unnecessarily stored states from db (if a node is frequently restarted, current approach will lead to more "unnecessary" states stored to disk)

### DIFF
--- a/packages/lodestar-cli/src/util/process.ts
+++ b/packages/lodestar-cli/src/util/process.ts
@@ -21,8 +21,6 @@ export function onGracefulShutdown(
       });
 
       await cleanUpFunction();
-      logFn("Cleanup completed");
-      process.exit(0);
     });
   }
 }
--- a/packages/lodestar/src/tasks/index.ts
+++ b/packages/lodestar/src/tasks/index.ts
@@ -15,6 +15,11 @@ import {IBeaconSync} from "../sync";
 import {InteropSubnetsJoiningTask} from "./tasks/interopSubnetsJoiningTask";
 import {INetwork} from "../network";
 
+/**
+ * Minimum number of epochs between archived states
+ */
+export const MIN_EPOCHS_PER_DB_STATE = 1024;
+
 export interface ITasksModules {
   db: IBeaconDb;
   logger: ILogger;
@@ -64,6 +69,12 @@ export class TasksService implements IService {
     this.network.gossip.removeListener("gossip:start", this.handleGossipStart);
     this.network.gossip.removeListener("gossip:stop", this.handleGossipStop);
     await this.interopSubnetsTask.stop();
+    // Archive latest finalized state
+    await new ArchiveStatesTask(
+      this.config,
+      {db: this.db, logger: this.logger},
+      await this.chain.getFinalizedCheckpoint()
+    ).run();
   }
 
   private handleGossipStart = async (): Promise<void> => {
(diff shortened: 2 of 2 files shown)


packages/lodestar/src/tasks/index.ts
```diff
@@ -82,7 +93,10 @@ export class TasksService implements IService {
         finalized
       ).run();
       // should be after ArchiveBlocksTask to handle restart cleanly
-      await new ArchiveStatesTask(this.config, {db: this.db, logger: this.logger}, finalized).run();
+      const lastStoredEpoch = (await this.db.stateArchive.lastKey()) as number;
```
twoeths, line 96: we changed from storing by epoch to by slot https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/db/api/beacon/repositories/stateArchive.ts#L17

mpetrunic (approved):

twoeths:
I wonder how can we generate a weak subjectivity checkpoint if we do this? Perhaps we should implement a worker thread to start from a finalized state and reprocess finalized blocks up until a slot/epoch to generate/download a weak subjectivity checkpoint state. This could also be used for the debug API where we can download whatever state by slot (currently we can only download states of finalized slot/epoch and after this PR, we could only download less states - i.e. only states available in `archiveState` repository).

packages/lodestar-cli/src/util/process.ts
```diff
@@ -21,8 +21,6 @@ export function onGracefulShutdown(
       });
 
       await cleanUpFunction();
```
dapplion, line 23: Add a note for historical reasons saying,
```
Ensure no open handlers so the javascript process can exit properly without having to explicitly call process.exit(0)
```

packages/lodestar/src/tasks/index.ts
```diff
@@ -15,6 +15,11 @@ import {IBeaconSync} from "../sync";
 import {InteropSubnetsJoiningTask} from "./tasks/interopSubnetsJoiningTask";
 import {INetwork} from "../network";
 
+/**
+ * Minimum number of epochs between archived states
+ */
+export const MIN_EPOCHS_PER_DB_STATE = 1024;
```
dapplion, line 21: Should this constant be in a more visible place like the `constants` file?

dapplion (approved):

PR #1939, opened 2021-01-12
Remove passing gossip attestations to forkchoice

### LINKED ISSUE #1938: Do not pass gossip attestation to forkchoice
**Describe the bug**
A lot of gossip attestations are not included in any blocks so when receiving them through gossip, we should not include them in forkchoice. If we add them, potentially we'll have different forkchoice head compared to other clients.

**Expected behavior**
We should only pass attestations included in blocks to forkchoice.

twoeths:
resolves #1938

A lot of gossip attestations are not included in any blocks so when receiving them through gossip, we should not include them in forkchoice. If we add them, potentially we'll have different forkchoice head compared to other clients.

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -11,7 +11,6 @@ import {
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {IForkChoice} from "@chainsafe/lodestar-fork-choice";
 import {
-  Attestation,
   BeaconState,
   Checkpoint,
   ENRForkID,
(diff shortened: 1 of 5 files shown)


dapplion (approved):

wemeetagain:
I think we _want_ to apply as many attestations to the fork choice as possible, including those received via gossip.
Our view of the head will naturally be slightly different than other clients depending on the attestations we've received.

See https://github.com/sigp/lighthouse/blob/9ed65a64f8fc518c98b381e628556df9bf906acb/beacon_node/network/src/beacon_processor/worker/gossip_methods.rs#L128

twoeths:
> I think we _want_ to apply as many attestations to the fork choice as possible, including those received via gossip.
> Our view of the head will naturally be slightly different than other clients depending on the attestations we've received.
> 
> See https://github.com/sigp/lighthouse/blob/9ed65a64f8fc518c98b381e628556df9bf906acb/beacon_node/network/src/beacon_processor/worker/gossip_methods.rs#L128

so I'm closing this PR, but notice that the current implementation does not pass all possible attestations to our forkchoice, will open an issue for it

PR #1946, opened 2021-01-13
Avoid going thru regen module whenever possible

### LINKED ISSUE #1872: Avoid going through regen module whenever possible
This comes from #1723 , for some flows we can get states immediately instead of having to go through regen module and stay in queue to wait.
+ first candidate is to modify `getHeadStateContext()` to get head state right from our state cache
+ look for similar scenarios and make the improvement

twoeths:
resolves #1872 

+ I don't see other flows that could bypass the regen module except for the head state

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -139,13 +139,9 @@ export class BeaconChain implements IBeaconChain {
   public async getHeadStateContext(): Promise<ITreeStateContext> {
     // head state should always exist
     const head = this.forkChoice.getHead();
-    const headStateRoot =
-      (await this.db.checkpointStateCache.getLatest({
-        root: head.blockRoot,
-        epoch: Infinity,
-      })) || (await this.regen.getState(head.stateRoot));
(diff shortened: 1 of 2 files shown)


wemeetagain (approved):

packages/lodestar/src/chain/chain.ts
```diff
@@ -139,13 +139,9 @@ export class BeaconChain implements IBeaconChain {
...
     // head state should always exist
     const head = this.forkChoice.getHead();
-    const headStateRoot =
-      (await this.db.checkpointStateCache.getLatest({
-        root: head.blockRoot,
-        epoch: Infinity,
-      })) || (await this.regen.getState(head.stateRoot));
-    if (!headStateRoot) throw Error("headStateRoot does not exist");
-    return headStateRoot;
+    const headState = await this.db.stateCache.get(head.stateRoot);
```
twoeths, line 142: this should keep the same logic to get from `checkpointStateCache` then `stateCache`, otherwise we can't start from an anchor state at orphaned block

PR #1951, opened 2021-01-14
Validate and broadcast voluntary exit, propsoser/attester slashing

### LINKED ISSUE #1944: Broadcast voluntary exit
Implement `/v1/beacon/pool/voluntary_exits` POST api: Submits SignedVoluntaryExit object to node's pool and if passes validation node MUST broadcast it to network.

+ Right now we miss the validation and broadcasting of the voluntary exit to the network through gossip, need to do that at https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/api/impl/beacon/pool/pool.ts#L90
+ Not sure if we want to do the same thing for AttesterSlashing and ProposerSlashing

twoeths:
resolves #1944

### DIFF
--- a/packages/lodestar/src/api/impl/beacon/pool/pool.ts
+++ b/packages/lodestar/src/api/impl/beacon/pool/pool.ts
@@ -5,6 +5,9 @@ import {SignedVoluntaryExit} from "../../../../../../lodestar-types/lib/types/op
 import {IAttestationJob, IBeaconChain} from "../../../../chain";
 import {AttestationError, AttestationErrorCode} from "../../../../chain/errors";
 import {validateGossipAttestation} from "../../../../chain/validation";
+import {validateGossipAttesterSlashing} from "../../../../chain/validation/attesterSlashing";
+import {validateGossipProposerSlashing} from "../../../../chain/validation/proposerSlashing";
+import {validateGossipVoluntaryExit} from "../../../../chain/validation/voluntaryExit";
 import {IBeaconDb} from "../../../../db/api";
 import {INetwork} from "../../../../network";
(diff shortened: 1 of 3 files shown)


packages/lodestar/src/api/impl/beacon/pool/pool.ts
```diff
@@ -71,22 +74,25 @@ export class BeaconPoolApi implements IBeaconPoolApi {
...
   }
 
   public async getProposerSlashings(): Promise<ProposerSlashing[]> {
     return this.db.proposerSlashing.values();
   }
 
   public async submitProposerSlashing(slashing: ProposerSlashing): Promise<void> {
-    await this.db.proposerSlashing.add(slashing);
+    await validateGossipProposerSlashing(this.config, this.chain, this.db, slashing);
+    await Promise.all([this.network.gossip.publishProposerSlashing(slashing), this.db.proposerSlashing.add(slashing)]);
```
wemeetagain, line 87: I think we may want to be 'greedy' with the slashings we receive and not publish them in gossip.

twoeths: what's the reason to do that?  From the spec: `Submits ProposerSlashing object to node's pool and if passes validation node MUST broadcast it to network.`

3xtr4t3rr3str14l: so i guess this issue is tracking this now?
https://github.com/ChainSafe/lodestar/issues/1956

3xtr4t3rr3str14l (approved):

wemeetagain (approved):

PR #1952, opened 2021-01-14
Hard fork friendly db

### LINKED ISSUE #1927: Implemented hard fork compatible db
Since we store everything in ssz format to db, we could end up being unable to read pre fork data.
As suggested in chat, we could add `getTypesByEpoch` method to BeaconConfig. This assumes only time anchored types would change (contains slot or epoch) otherwise we would need another repository.

mpetrunic:
resolves #1927


mpetrunic:
@ChainSafe/eth2-0 A bit of a crossroad:
1. Continue like this creating repository levels (seems like a lot of code). Should I just implement methods we need?
2. Change base Repository class to consider forkVersion on every method
    - this probably means that we would have to change a lot of code interacting with database
    - every repository is already ready for changes in hard forks

packages/lodestar-db/src/const.ts
```diff
@@ -0,0 +1,2 @@
+export const BUCKET_LENGTH = 1;
+export const FORK_VERSION_LENGTH = 8;
```
dapplion: Would that mean that all keys in the DB will consume 8 extra bytes? How many keys do we have in the DB usually?

mpetrunic: Yes, no idea. But we cannot create new bucket on every fork since we have finite amount of buckets (256). 

But even if we have 100k keys, that's 0.8mb db size increase (probably even less since it will be compressed)

wemeetagain: Alternative:
Increase bucket length to allow for more buckets (eg: increase to 2 or 4)
Create a bucket(+"repository") for every new datatype (eg: Bucket.blockArchive_phase0, Bucket.blockArchive_lightClient)
In this case, the database doesn't need to enforce or understand fork versions. Rather, the enforcement will come from consumer logic which stores to different buckets according to fork-specific implementations.
The database is only responsible for storing different types of data. The node / higher-level logic will be responsible for storing/fetching in a fork-compatible way.
Of course, any wrapper "repositories" can still be done as in this PR.

packages/lodestar-params/src/presets/mainnet.ts
```diff
@@ -10,8 +10,18 @@ export const commit = "v1.0.0";
...
+const phase0Params = createParams<IBeaconParams>(phase0Yaml, BeaconParams);
+
 export const params: IBeaconParams = {
-  ...createParams<IBeaconParams>(phase0Yaml, BeaconParams),
+  ...phase0Params,
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  ALL_FORKS: [
+    {
+      epoch: 0,
+      currentVersion: 16777216,
```
dapplion, line 23: Could you add a note on where to get this number from?

packages/lodestar/src/db/api/beacon/repositories/blockArchive/abstract.ts
```diff
@@ -0,0 +1,116 @@
...
+      : null;
+    if (firstSlot === null) throw Error("specify opts.gt or opts.gte");
+    const valuesStream = super.valuesStream(opts);
+    const step = (opts && opts.step) || 1;
+
+    for await (const value of valuesStream) {
+      if ((value.message.slot - firstSlot) % step === 0) {
+        yield value;
+      }
+    }
```
dapplion, line 114: Not completely related to this PR but if someone sends a blocksByRangeRequest to us with
```
startSlot: 0,
count: 2,
step: 360000
```
we would read the entire block DB?

packages/lodestar-params/src/presets/mainnet.ts
mpetrunic: for conversation sake, I've converted array to number mannually to avoid bringing lodestar-utils here just for that.
I'll add todo to remove this since this should go in yaml config but didn't hit dev branch yet

packages/lodestar/src/db/api/beacon/repositories/blockArchive/abstract.ts
mpetrunic: probably. We can add some validation to blocksByRange to limit step size.

dapplion: Or if step size is > stepSizeLimit don't stream and fetch one by one

packages/lodestar-db/src/const.ts
dapplion: > Increase bucket length to allow for more buckets (eg: increase to 2 or 4)

Sound that with 65536 it's plenty of buckets

packages/lodestar/src/db/api/beacon/repositories/block/ligthclient.ts
```diff
@@ -0,0 +1,21 @@
...
+import {Lightclient} from "@chainsafe/lodestar-types";
+
+/**
+ * Blocks by root
+ *
+ * Used to store unfinalized blocks
+ */
+export class LightClientBlockRepository extends Repository<Uint8Array, Lightclient.SignedBeaconBlock> {
+  public constructor(config: IBeaconConfig, db: IDatabaseController<Buffer, Buffer>) {
+    super(config, db, Bucket.lightclientBlock, config.types.lightclient.SignedBeaconBlock);
```
twoeths, line 12: should we pass in fork version here?

twoeths: ah seems like we're not using the `forkVersion` inside the abstract Repository

mpetrunic:
FYI, this is ready to go. Tests are expected to fail until we start producing post lc fork blocks since they are not stored currently :)

dapplion (approved):

packages/lodestar-db/src/abstractRepository.ts
```diff
@@ -23,11 +24,20 @@ export abstract class Repository<I extends Id, T> {
...
 
-  protected constructor(config: IBeaconConfig, db: IDatabaseController<Buffer, Buffer>, bucket: Bucket, type: Type<T>) {
+  protected forkVersion: Version;
+
+  protected constructor(
+    config: IBeaconConfig,
+    db: IDatabaseController<Buffer, Buffer>,
+    bucket: Bucket,
+    type: Type<T>,
+    forkVersion?: Version
```
wemeetagain: Repository doesn't need forkVersion anymore?

wemeetagain (changes requested):
LGTM

one minor comment

dapplion (approved):

wemeetagain (approved):

PR #1959, opened 2021-01-15
--testnet -> --network

### LINKED ISSUE #1808: Separate mainnet cli flag from testnet cli flag?
Currently, (as per #1806 ), we do `--testnet mainnet` to run lodestar on mainnet.  I did this for convenience so i didn't have to restructure the current --testnet logic.  Said current current logic (i.e. much of the architecture of `lodestar-cli`) is centered around the assumption that we have a `--testnet` flag passed in order to bootstrap us to a valid eth2 environment.  

This issue proposes to allow the user to bootstrap to mainnet without the `--testnet` flag (i.e. like `node lodestar beacon` instead of `node lodestar beacon --testnet mainnet`.

Or maybe some feel it's not a big deal to leave as is (i.e. running as `--testnet mainnet`)?  Would be interested to hear everyone's thoughts.

3xtr4t3rr3str14l:
Proposed solution for #1808 as inspired by lighthouse.

Fixes https://github.com/ChainSafe/lodestar/issues/1808


dapplion (dismissed):

wemeetagain (commented):
Can you do one last check to change all instances of `testnet` to `network`? I think your last design doc PR, just merged, may have included more.

3xtr4t3rr3str14l:
> Can you do one last check to change all instances of `testnet` to `network`? I think your last design doc PR, just merged, may have included more.

yeah forgot to change the docs.  they are included in these changes now

packages/lodestar-cli/src/cmds/account/cmds/validator/import.ts
```diff
@@ -35,7 +35,7 @@ Ethereum Foundation utility.",
 
   examples: [
     {
-      command: "account validator import --testnet medalla --directory $HOME/eth2.0-deposit-cli/validator_keys",
+      command: "account validator import --network medalla --directory $HOME/eth2.0-deposit-cli/validator_keys",
```
dapplion, line 38: We should replace all mentions of medalla (and any other dead testnets) with pyrmont (in a future issue / PR)

packages/lodestar-cli/src/options/globalOptions.ts
```diff
@@ -17,10 +17,10 @@ const globalSingleOptions: ICliCommandOptions<IGlobalSingleArgs> = {
     type: "string",
   },
 
-  testnet: {
-    description: "Use a testnet configuration",
+  network: {
+    description: "Use a network configuration",
```
dapplion: Rephrase for `Name of the Eth2 chain network to join`

packages/lodestar-cli/src/cmds/account/cmds/validator/import.ts
3xtr4t3rr3str14l: made an issue for it #1972

wemeetagain (approved):

dapplion (approved):

PR #1961, opened 2021-01-17
Initial sync as state machine

dapplion:
Refactors initial sync with a heavily Lightouse inspired sync.

> TLDR: higher sync speeds - works as a state machine: easier testing, debugging and predictability

Record of successfully completing initial sync on cloud instances

- `lodestar-cloud-01`: Pyrmont on container `lodestar_lodestar_1`: 2 times
- `lodestar-cloud-02`: Mainnet on container `lodestar_lodestar_1`: 3 times

---

The Sync has two integral components:
### `Batch`
Represent a fix range of slots to fetch and process. A batch is a state machine that transitions through multiple status with public methods:
- `AwaitingDownload`: Initial state, or the batch has failed either downloading or processing, but can be requested again.
- `Downloading`: The batch is being downloaded
- `AwaitingProcessing`: The batch has been completely downloaded and is ready for processing
- `Processing`: The batch is being processed.
- `AwaitingValidation`: The batch was successfully processed and is waiting to be validated. It is not sufficient to process a batch successfully to consider it correct. This is because batches could be erroneously empty, or incomplete. Therefore, a batch is considered valid, only if the next sequential batch imports at least a block.

Tracking the status in this way simplifies a lot the logic to concurrently handle and download multiple batches at once. The `AwaitingValidation` also allow to handle infinitely long periods of completely skipped slots without issues.

Batches always start a slot 1 to ease committee forecasting when batch signature verification is ON.
```
 Epoch boundary |                                   |
  ... | 30 | 31 | 32 | 33 | 34 | ... | 61 | 62 | 63 | 64 | 65 |
       Batch 1       |              Batch 2              |  Batch 3
 ```
 
Batches also track the failed attemps which would allow sofisticate peer scoring in upcomming PRs. So far it's used to reduce the chance of re-requesting a batch to a peer with a failed attempt.
 
### `SyncChain`
 
Main driver that handles a map of `Batch`es. It uses an AsyncIterable `batchProcessor` to processes batches to be processed in order and only one at once. All the state necessary to chose the next to be downloaded batch and when syncing is done is derived from the batches. Plus `startEpoch` which tracks the latest validated epoch.
 
It interfaces with the sync manager through 3 methods only
 - `processChainSegment`
 - `downloadBeaconBlocksByRange`
 - `getPeersAndTargetEpoch`: Abstracts the peer handling logic upstream. Allows to sync manager to enforce minPeers even throughout the sync

`SyncChain` instead of `start()` `stop()` exposes a single `sync()` method that resolves when sync. It uses an abort signal for stopping.

`SyncChain` is able to detect that the sync has stalled with a recurring `maybeStuckTimeout`. It will print detailed data to help debugging and try to re-fetch new batches and reprocess pending. It also detects specific invalid internal state combinations and will nuke itself if those happen.

### CODE REVIEW

**Comment by twoeths** [comment]
this is great, the sync speed is really impressive to me! `/eth/v1/node/syncing` does not seem to be correct anymore during initial sync.

File: packages/lodestar/src/tasks/tasks/archiveBlocks.ts [hunk 1]
```diff
@@ -57,8 +56,7 @@ export class ArchiveBlocksTask implements ITask {
       i = upperBound;
     }
     await this.deleteNonCanonicalBlocks();
-    this.logger.profile("Archive Blocks epoch #" + this.finalized.epoch);
```
twoeths, line 60: why do you remove this? it's used to track how much time we use to archive blocks

packages/lodestar/src/sync/range/chain.ts
```diff
@@ -0,0 +1,350 @@
...
+      }
+
+      // Potentially process next AwaitingProcessing batch
+      this.triggerBatchProcessor();
+    } catch (e) {
+      batch.processingError(e);
+
+      // At least one block was successfully verified and imported, so we can be sure all
+      // previous batches are valid and we only need to download the current failed batch.
+      if (e instanceof BlockProcessorError && e.importedBlocks.length > 0) {
```
twoeths, line 279: I don't see this BlockProcessorError being thrown anywhere

packages/lodestar/src/sync/range/chain.ts
```diff
@@ -0,0 +1,350 @@
...
+
+      // Pull the current targetEpoch
+      const {targetEpoch} = this.getPeersAndTargetEpoch() || {};
+      if (targetEpoch === undefined) {
+        this.logger.verbose("No targetEpoch available");
+        continue;
+      }
+
+      // TEMP
+      validateBatchesStatus(toArr(this.batches));
```
twoeths: I guess this is needed for development, do we still need it for production?

packages/lodestar/src/sync/range/batches.ts
```diff
@@ -0,0 +1,76 @@
+import {Epoch} from "@chainsafe/lodestar-types";
```
twoeths, line 1: these util functions are nice, could we have unit tests for them?

packages/lodestar/src/tasks/tasks/archiveBlocks.ts
dapplion: As it's currently it's very noisy to log that to info. I agree it's valuable info but it should not be logged to info always. We should profile both the archive blocks task, the state transition and everything else but on demand only.

For now I've remove it since there also this statement
```
this.logger.verbose("Archiving of finalized blocks complete"
```

packages/lodestar/src/sync/range/chain.ts
dapplion: Right, I'm waiting for Cayman to tweak the chain segment processor and add this error. The error is not necessary to work, it just adds a performance boost in some situations

packages/lodestar/src/sync/range/chain.ts
dapplion: I'm not sure, I think we should keep testing and see if invalid situations happen or not. If `validateBatchesStatus` finds invalid batches something very wrong has happened potentially un-recoverable. In that case, we may end the sync or drop all pending progress and start from the previous validated checkpoint

dapplion:
> `/eth/v1/node/syncing` does not seem to be correct anymore during initial sync.

What behavior specifically do you see that does not match expectations? I've changed the target to just be `target = clock.currentSlot` (as Lighthouse does) as an interim solution for a future PR to handle the case of skipped slots on target.

dapplion:
I've completed successfully initial sync with this branch at commit db8c69f on Mainnet and Pyrmont and cloud instances

packages/lodestar/src/tasks/tasks/archiveBlocks.ts
dapplion: As discussed in the meeting we will implement profiling through metrics in an upcoming PR

twoeths:
> > `/eth/v1/node/syncing` does not seem to be correct anymore during initial sync.
> 
> What behavior specifically do you see that does not match expectations? I've changed the target to just be `target = clock.currentSlot` (as Lighthouse does) as an interim solution for a future PR to handle the case of skipped slots on target.

it returns `{"data":{"head_slot":"438615","sync_distance":"422901"}}` (head_slot is wrong) although I just started the sync in 30 minutes

dapplion:
> it returns {"data":{"head_slot":"438615","sync_distance":"422901"}} (head_slot is wrong) although I just started the sync in 30 minutes

Thanks for testing it out! Fixed with https://github.com/ChainSafe/lodestar/pull/1961/commits/88feee95d36723206b8676616029ba483afa9092. After a couple of minutes of testing it returns

```
/usr/app # curl http://localhost:9596/eth/v1/node/syncing
{"data":{"head_slot":"386","sync_distance":"467256"}}
```

packages/lodestar/src/sync/range/chain.ts
twoeths: looks like we'll use `ChainSegmentError`? https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/chain/errors/blockError.ts#L125

packages/lodestar/src/sync/sync.ts
```diff
@@ -132,18 +146,45 @@ export class BeaconSync implements IBeaconSync {
...
   }
 
-  private async startInitialSync(): Promise<void> {
-    if (this.mode === SyncMode.STOPPED) return;
-    this.mode = SyncMode.INITIAL_SYNCING;
-    this.startSyncTimer(this.config.params.SLOTS_PER_EPOCH * this.config.params.SECONDS_PER_SLOT * 1000);
-    await this.regularSync.stop();
-    await this.initialSync.start();
-  }
+  private processChainSegment: ProcessChainSegment = async (blocks) => {
```
wemeetagain, line 149: Remove in a future PR

packages/lodestar/src/sync/range/chain.ts
wemeetagain: will change in a separate PR

wemeetagain (approved):

PR #1966, opened 2021-01-18
Update block/chain segment processing

wemeetagain:
- Add `ChainSegmentError` - mirrors `BlockError` but includes # of `importedBlocks` before error
- Add `processChainSegment` - mirrors `processBlock` but with epoch-by-epoch signature verification
- Move batch signature verification into `processBlock` / `processChainSegment` rather than lower-level `runStateTransition`
- Move check for parent parent from `validateBlock` into `processBlock` / `processChainSegment`
- Update `processChainSegmentJob` to do filtering of irrelevant blocks & use `processChainSegment`

### DIFF
--- a/packages/lodestar/src/chain/blocks/process.ts
+++ b/packages/lodestar/src/chain/blocks/process.ts
@@ -14,6 +14,7 @@ import {IStateRegenerator, RegenError} from "../regen";
 import {BlockError, BlockErrorCode, ChainSegmentError} from "../errors";
 import {IBeaconDb} from "../../db";
 import {verifySignatureSetsBatch} from "../bls";
+import {findLastIndex} from "../../util/array";
 
 export async function processBlock({
   forkChoice,
@@ -109,15 +110,6 @@ export async function processChainSegment({
     }
     const startEpoch = computeEpochAtSlot(config, firstBlock.message.slot);
 
-    const findLastIndex = <T>(array: T[], predicate: (value: T) => boolean): number => {
-      let i = array.length;
-      while (i--) {
-        if (predicate(array[i])) {
-          return i;
-        }
-      }
-      return -1;
-    };
     // The `lastIndex` indicates the position of the last block that is in the current
     // epoch of `startEpoch`.
     const lastIndex = findLastIndex(blocks, (block) => computeEpochAtSlot(config, block.message.slot) === startEpoch);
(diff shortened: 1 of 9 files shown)


packages/lodestar/src/chain/blocks/processor.ts
```diff
@@ -72,20 +74,85 @@ export async function processBlockJob(modules: BlockProcessorModules, job: IBloc
...
-    // above functions only throw BlockError
-    modules.emitter.emit(ChainEvent.errorBlock, e);
-    throw e;
-  }
+export async function processChainSegmentJob(modules: BlockProcessorModules, job: IChainSegmentJob): Promise<void> {
+  const blocks = job.signedBlocks;
+
+  // Validate and filter out irrelevant blocks
+  const filteredChainSegment: SignedBeaconBlock[] = [];
+  blocks.forEach(async (block, i) => {
```
dapplion: Would recommend to use
```ts
for (const [i, block] of blocks.entries()) {
  //
}
```
instead

packages/lodestar/src/chain/blocks/validate.ts
```diff
@@ -6,86 +6,63 @@ import {IBlockJob} from "../interface";
 import {IBeaconClock} from "../clock";
 import {BlockError, BlockErrorCode} from "../errors";
 
-export async function validateBlocks({
+export async function validateBlock({
```
dapplion: No need for this function to be async anymore, there are no awaits inside

packages/lodestar/src/chain/blocks/processor.ts
dapplion: Also `validateBlock` should not be async, thus this function neither. See https://github.com/ChainSafe/lodestar/pull/1966/files#r559793686

packages/lodestar/src/chain/blocks/process.ts
```diff
@@ -1,48 +1,189 @@
...
+      throw new ChainSegmentError({
+        code: BlockErrorCode.PARENT_UNKNOWN,
+        parentRoot: firstBlock.message.parentRoot.valueOf() as Uint8Array,
+        job,
+        importedBlocks,
+      });
+    }
+    const startEpoch = computeEpochAtSlot(config, firstBlock.message.slot);
+
+    const findLastIndex = <T>(array: T[], predicate: (value: T) => boolean): number => {
```
dapplion: This should go to src/util/array and be unit tested

dapplion (approved):

mpetrunic (approved):

PR #1978, opened 2021-01-20
Faster CI by running scripts in parallel

### LINKED ISSUE #1977: Run unit tests in parallel
Currently lerna runs in ci unit tests from every package sequentially. To improve CI times, we should add `--parallel` flag to `test:unit` and `lint` script in root package.json

mpetrunic:


mpetrunic:
Don't see any noticable ci increase (likely lack of cores). There is noticable difference on my pc: https://github.com/chainsafe/lodestar/issues/1977#issuecomment-763548394

PR #1982, opened 2021-01-20
 Replace all mentions of medalla (and any other dead testnets) with pyrmont

### LINKED ISSUE #1972: Replace all mentions of medalla (and any other dead testnets) with pyrmont
We should replace all mentions of medalla (and any other dead testnets) with pyrmont (in a future issue / PR)

_Originally posted by @dapplion in https://github.com/ChainSafe/lodestar/pull/1959#discussion_r559796021_

3xtr4t3rr3str14l:
resolves #1972


dapplion (dismissed):

packages/lodestar/test/utils/pyrmont.ts
```diff
@@ -3,28 +3,29 @@ import {createIBeaconConfig, IBeaconConfig} from "@chainsafe/lodestar-config";
...
-  config.params.GENESIS_FORK_VERSION = Buffer.from("00000001", "hex");
+  config.params.DEPOSIT_CONTRACT_ADDRESS = Buffer.from(pyrmontTestnet.beaconParams.DEPOSIT_CONTRACT_ADDRESS as string);
+  config.params.MIN_GENESIS_TIME = pyrmontTestnet.beaconParams.MIN_GENESIS_TIME as number;
+  config.params.GENESIS_DELAY = pyrmontTestnet.beaconParams.GENESIS_DELAY as number;
+  config.params.GENESIS_FORK_VERSION = Buffer.from(pyrmontTestnet.beaconParams.GENESIS_FORK_VERSION as string, "hex");
   return config;
 }
 
-export const goerliMedallaLogs = [
+export const goerliPyrmontLogs = [
```
dapplion: The data is from Medalla not Pyrmont. To prevent confusion you should change it to goerliTestnetDepositLog and add a JSDoc above clarifying that this is from the Medalla deposit contract `0x07b39F4fDE4A38bACe212b546dAc87C58DfE3fDC`. Same for goerliPyrmontDepositEvents below

3xtr4t3rr3str14l: hmm, yeah now i'm wondering if maybe i should revert this file to its original medalla.ts and just use that?  otherwise, we might have to update this every time the current testnet dies.  i think it's probably okay to use data from an old testnet since it's only testing based on the data structure and not necessarily trying to do live testnet testing.  how do you feel about that?

dapplion: Just call it something generic like sample testnet data and add a comment explaining that the data is taken from Medalla

twoeths (dismissed):

dapplion (approved):

Issue #1970, opened 2021-01-19
Add voluntary exit functionality to the CLI


dapplion:
Nice to have to clean-up after testing Pyrmont.

3xtr4t3rr3str14l:
do you mean that you'd want a flag or something like that to add as an option to perform a voluntary exit once the node process is quit?  or maybe just something to perform a voluntary exit in general regardless of when it would be ran?  just trying to clarify the intent here

dapplion:
> do you mean that you'd want a flag or something like that to add as an option to perform a voluntary exit once the node process is quit? or maybe just something to perform a voluntary exit in general regardless of when it would be ran? just trying to clarify the intent here

You can take a look at the Pyrsm CLI which has this functionality. It should be done upon express intention of the users with a command similar to
```
lodestar accounts voluntary-exit
```

PR #1986, opened 2021-01-22
add voluntary exit to lodestar-cli

3xtr4t3rr3str14l:
solution for #1970 
currently testing with an active pyrmont validator to see if/when it proposes a block.  after the validator has done this and maybe after it has ran for a good while, i'll perform voluntary exit as a test for this PR

i also added a check for the publish voluntary exit gossip call to check and make sure we have gossip peers in the voluntary exit topic before sending a voluntary exit so it will error if we have no relevant peers.  depending on how the team likes the design of this particular case, i may try to apply it to other parts of gossip and try to add some code to track peers by topic


3xtr4t3rr3str14l:
was able to successfully perform a voluntary exit for this validator (https://pyrmont.beaconcha.in/validator/b6e9b6ba69fbd330993b8167c2e1daf4dd04eaa6670a21e776ddc4dec3cc2cedd50bd5c3d7a134d91447ad905f1c1cce#attestations) so this PR is ready for review

packages/lodestar-cli/src/cmds/account/cmds/validator/voluntaryExit.ts
```diff
@@ -0,0 +1,122 @@
...
+    state.fork = (await api.beacon.state.getFork("head")) || state.fork;
+    state.genesisValidatorsRoot = (await api.beacon.getGenesis())?.genesisValidatorsRoot || state.genesisValidatorsRoot;
+
+    const domain = getDomain(config, state, DomainType.VOLUNTARY_EXIT, epoch);
+    const signingRoot = computeSigningRoot(config, config.types.VoluntaryExit, voluntaryExit, domain);
+
+    const signedVoluntaryExit: SignedVoluntaryExit = {
+      message: voluntaryExit,
+      signature: secretKey.sign(signingRoot).toBytes(),
+    };
```
mpetrunic: Can you move voluntary exit creation process to lodestar-validator as it would allow users of lodestar-validator to use that as well?

Something like:
```
const validator = new LodestarValidator(...)

await validator.voluntaryExit()
```

packages/lodestar-validator/src/validator.ts
```diff
@@ -73,6 +75,54 @@ export class Validator {
...
+    };
+
+    try {
+      this.logger.info(`Waiting for validator ${publicKey} to be exited...`);
+      await this.apiClient.beacon.pool.submitVoluntaryExit(signedVoluntaryExit);
+      this.logger.info(`Successfully exited validator ${publicKey}`);
+      await this.apiClient.disconnect();
+    } catch (error) {
+      this.logger.error(error);
+      await this.apiClient.disconnect();
```
dapplion, line 119: You can move `apiClient.disconnect();` to a finally clause to run it for both error and success. I would let the error bubble up here, why is being catched and logged?

packages/lodestar-validator/src/validator.ts
```diff
@@ -73,6 +75,54 @@ export class Validator {
...
+   * Perform a voluntary exit for the given validator by its public key.
+   */
+  public async voluntaryExit(publicKey: string): Promise<void> {
+    await this.apiClient.connect();
+
+    const stateValidator = await this.apiClient.beacon.state.getStateValidator(
+      "head",
+      this.config.types.BLSPubkey.fromJson(publicKey)
+    );
+    if (!stateValidator) throw new Error("No validator found in validator store.");
```
mpetrunic:
```suggestion
    if (!stateValidator) throw new Error("Validator not found in beacon chain.");
```
?

packages/lodestar-validator/src/validator.ts
```diff
@@ -73,6 +75,54 @@ export class Validator {
...
+
+    const epoch = this.apiClient.clock.currentEpoch;
```
mpetrunic: I think this should be passed as argument to function (and cli) and can default to current epoch. I think it's idea that you can schedule exit in the future

packages/lodestar-validator/src/validator.ts
```diff
@@ -73,6 +75,54 @@ export class Validator {
...
+    const voluntaryExit = {
+      epoch,
+      validatorIndex: stateValidator.index,
+    };
+
+    const state: BeaconState = this.config.types.BeaconState.tree.defaultValue();
+    state.fork = (await this.apiClient.beacon.state.getFork("head")) || state.fork;
+    state.genesisValidatorsRoot =
+      (await this.apiClient.beacon.getGenesis())?.genesisValidatorsRoot || state.genesisValidatorsRoot;
+    const domain = getDomain(this.config, state, DomainType.VOLUNTARY_EXIT, epoch);
```
mpetrunic: there is computeDomain so you don't have to recreate state

packages/lodestar-validator/src/validator.ts
```diff
@@ -73,6 +75,54 @@ export class Validator {
...
+
+    const epoch = this.apiClient.clock.currentEpoch;
+
+    const voluntaryExit = {
+      epoch,
+      validatorIndex: stateValidator.index,
+    };
+
+    const state: BeaconState = this.config.types.BeaconState.tree.defaultValue();
+    state.fork = (await this.apiClient.beacon.state.getFork("head")) || state.fork;
```
mpetrunic: you should fetch https://ethereum.github.io/eth2.0-APIs/#/Config/getForkSchedule and pick fork version based on voluntary exit epoch (if you wanna schedule exit in the future you need to use future fork version). I think it's fine to use (fallback) to (this.apiClient.beacon.state.getFork("head")) if nothing is scheduled

packages/lodestar-validator/src/validator.ts
```diff
@@ -73,6 +75,54 @@ export class Validator {
...
+    state.genesisValidatorsRoot =
+      (await this.apiClient.beacon.getGenesis())?.genesisValidatorsRoot || state.genesisValidatorsRoot;
+    const domain = getDomain(this.config, state, DomainType.VOLUNTARY_EXIT, epoch);
+    const signingRoot = computeSigningRoot(this.config, this.config.types.VoluntaryExit, voluntaryExit, domain);
+
+    let secretKey;
+    for (const sk of this.opts.secretKeys) {
+      if (sk.toPublicKey().toHex() === publicKey) secretKey = sk;
```
mpetrunic: probably better to use config.types.PublicKey.equals() since 0x prefix or capitalization could screw with you here

packages/lodestar-validator/src/validator.ts
```diff
@@ -73,6 +75,54 @@ export class Validator {
...
+
+    const signedVoluntaryExit: SignedVoluntaryExit = {
+      message: voluntaryExit,
+      signature: secretKey.sign(signingRoot).toBytes(),
+    };
+
+    try {
+      this.logger.info(`Waiting for validator ${publicKey} to be exited...`);
+      await this.apiClient.beacon.pool.submitVoluntaryExit(signedVoluntaryExit);
+      this.logger.info(`Successfully exited validator ${publicKey}`);
```
mpetrunic: this is not exactly true, validator will first wait for exit to be included in block than wait in queue.

Maybe better is to write submitted voluntary exit to the network.

packages/lodestar-validator/src/validator.ts
3xtr4t3rr3str14l: no good reason.  probably just a remnant of a previous phase of development.  i'll change it to bubble up

packages/lodestar-validator/src/validator.ts
3xtr4t3rr3str14l: working on this api implementation here: #2036

mpetrunic (approved):

PR #1997, opened 2021-01-26
Implement CachedBeaconState

### LINKED ISSUE #1950: Implement CachedBeaconState
The idea came from https://github.com/ChainSafe/lodestar/pull/1925#pullrequestreview-564581634

> Going forward, we should take some time to possibly re-organize the beacon state, validator cache, epoch context, epoch process, etc. as needed.

> This is mainly an issue of how we organize the various datastructures (BeaconState, CachedValidatorsBeaconState, EpochContext, EpochProcess, IStateContext). I'm not sure what the right organization is (possibly no re-organization needed) but what we have seems to be getting more cluttered as we've added more and more optimizations. We have several nested datastructures that are all related to the beacon state / epoch state.

> My hunch is that we may be able to expand the CachedValidatorBeaconState to be more of a general CachedBeaconState at the 'top-level', like IStateContext, which exposes the BeaconState + extra methods, with various caches 'under the hood' (each cache using 'persistent-ts' or similar for cheap cloning).

Need to come to an agreement regarding how it looks like before implementation.

twoeths:
resolves #1950 

+ Implement CachedBeaconState
+ Cheap `clone()`
+ Passed spec tests
+ Pyrmont sync is in progress


twoeths:
not able to test this due to #2005

wemeetagain:
closing in favor of #2172 :pray:

PR #2009, opened 2021-01-29
Fix validator not found and wrong events url

mpetrunic:
- fix error in sim tests where validator with index 0 cannot be found
- fixed error where validator is using wrong events url

### DIFF
--- a/packages/lodestar-validator/src/api/impl/rest/events/events.ts
+++ b/packages/lodestar-validator/src/api/impl/rest/events/events.ts
@@ -16,7 +16,7 @@ export class RestEventsApi implements IEventsApi {
 
   public getEventStream(topics: BeaconEventType[]): IStoppableEventIterable<BeaconEvent> {
     const eventSource = new EventSource(
-      urlJoin(this.baseUrl, "/eth/v1/events?" + topics.map((topic) => `topics=${topic}`).join("&"))
+      urlJoin(this.baseUrl, "/v1/events?" + topics.map((topic) => `topics=${topic}`).join("&"))
     );
     return new LodestarEventIterator(({push}) => {
       eventSource.onmessage = (event) => {
--- a/packages/lodestar-validator/src/api/impl/rest/events/events.ts
+++ b/packages/lodestar-validator/src/api/impl/rest/events/events.ts
@@ -16,7 +16,7 @@ export class RestEventsApi implements IEventsApi {
 
   public getEventStream(topics: BeaconEventType[]): IStoppableEventIterable<BeaconEvent> {
     const eventSource = new EventSource(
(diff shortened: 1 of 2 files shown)


mpetrunic:
@ChainSafe/eth2-0 how is it possible for e2e tests to fail without single warning or error. I also didn't make any significant changes 😞

packages/lodestar-validator/src/api/impl/rest/events/events.ts
```diff
@@ -16,7 +16,7 @@ export class RestEventsApi implements IEventsApi {
 
   public getEventStream(topics: BeaconEventType[]): IStoppableEventIterable<BeaconEvent> {
     const eventSource = new EventSource(
-      urlJoin(this.baseUrl, "/v1/events?" + topics.map((topic) => `topics=${topic}`).join("&"))
+      urlJoin(this.baseUrl, "/eth/v1/events?" + topics.map((topic) => `topics=${topic}`).join("&"))
```
dapplion, line 19: You didn't update the validators RestEventsApi.getEventStream() url

dapplion:
> @ChainSafe/eth2-0 how is it possible for e2e tests to fail without single warning or error. I also didn't make any significant changes

I would love to find a way for sim tests to provide valuable feedback, but is it possible?

mpetrunic:
cc @wemeetagain Seems like when I fix events sim tests aren't working. My best guess is that something is wrong with using events for duties so it works if they are disabled (not working)

packages/lodestar/src/api/impl/beacon/state/state.ts
```diff
@@ -85,7 +86,7 @@ export class BeaconStateApi implements IBeaconStateApi {
         validatorIndex = undefined;
       }
     }
-    if (!validatorIndex) {
+    if (!validatorIndex && validatorIndex !== 0) {
```
wemeetagain: `validatorIndex == null`? :joy:

mpetrunic: What about undefined 😅

dapplion: > What about undefined

`variable == null` with only two `=` will be true for `null` and `undefined`

mpetrunic: Lol wth thats so confusing. 😓

dapplion: Using Javascript extremely confusing type coercion to our advantage =D https://www.freecodecamp.org/news/js-type-coercion-explained-27ba3d9a2839/

I think 
```ts
if (arg != null) {}
if (arg == null {}
```
Are legit uses so you don't have to check for undefined and null, it's kind of a nullish coalescing operator on a if statement

dapplion (approved):

wemeetagain (approved):

PR #2011, opened 2021-01-29
Resolve OOM issue

### LINKED ISSUE #2005: JS heap blows up when running beacon node
**Describe the bug**

When running a synced lodestar beacon node (cmd: `/lodestar beacon --network pyrmont --rootDir .pyrmont/ --eth1.providerUrl https://goerli.infura.io/v3/3ee38f4a9c4447dfbd050e5afbb4b9c6`) and the validator (cmd: `./lodestar validator --network pyrmont --rootDir .pyrmont/ --graffiti "lodestar wuz here"`) with an activated pyrmont validator, on pyrmont, everything runs smoothly for about 16-18 hours (beacon node is stable, validator is attesting, etc.), and then I got an OOM issue.  I'm not sure if it's connected to the fact that both the validator and the beacon node were ran together, but the fact that other people have ran the beacon node without an OOM issue for quite some time (and that no one has really been running an active testnet validator alongside the beacon nod in lodestar, to my knowledge) gives me a hunch that it's connected to the validator.

EDIT: this issue happens when running just the beacon node by itself and does not require the validator to be ran to replicate the issue.

This is what my heap space looked like in grafana (the big empty space is between when the crash happened and when I noticed the crash)
image

here is my crash report:
https://pastebin.com/WJCEe6R6

twoeths:
resolves #2005 

+ The current `sleep(0)` inside `runStateTransition()` causes BeaconState/EpochContext stays inside the outer Promise scope until the inner Promise (`sleep(0)`) resolved, gradually it causes a lot of objects in memory like this

Screen Shot 2021-01-29 at 21 26 26

+ Simply move `sleep(0)` to the consumer help fix the issue. **__Note__**: we cannot simply remove these `sleep(0)` calls, otherwise we'll lose peers
+ Work in my local environment: memory stays from 1.1GB to 1.8GB (currently it's 4GB after 1h if I start from epoch 4xxx), testing in contavo 01 now

### DIFF
--- a/packages/lodestar/src/chain/blocks/process.ts
+++ b/packages/lodestar/src/chain/blocks/process.ts
@@ -14,6 +14,7 @@ import {BlockError, BlockErrorCode, ChainSegmentError} from "../errors";
 import {IBeaconDb} from "../../db";
 import {verifySignatureSetsBatch} from "../bls";
 import {groupBlocksByEpoch} from "./util";
+import {sleep} from "@chainsafe/lodestar-utils";
 
 export async function processBlock({
   forkChoice,
@@ -153,6 +154,8 @@ export async function processChainSegment({
           validSignatures: true,
         });
         importedBlocks++;
+        // we'll likely lose peers without this
+        await sleep(0);
       }
     } catch (e) {
       if (e instanceof RegenError) {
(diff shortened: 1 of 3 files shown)


3xtr4t3rr3str14l:
The beacon node has not yet crashed running off a build of the branch for this PR, but it's pretty much following the same pattern of heap usage as before (middle graph is the last time it crashed when running on master, and right side graph is what's running now).  @tuyennhv are seeing different behavior than this?

Screenshot from 2021-01-30 01-41-00

twoeths:
when I sync from scratch, I still can see these "zombie" CachedValidatorsBeaconState objects so this is not ready. It only works in my scenario when I used `--inspect` and restarted from epoch 9xxx and left there for 1 hour: master left a lot of zombie state objects and heap memory increased to 4GB, this branch did not and maintained a heap memory of < 2GB, maybe that's not enough time to test.

Plus there are concerns/uncertainties with this PR so I'm closing this, at least I think we found the root cause of "zombie" CachedValidatorsBeaconState objects

PR #2022, opened 2021-02-02
Lightclient block production

### LINKED ISSUE #1985: Produce lightclient blocks
We need to implement switch to api where depending on current epoch/fork we need to produce different blocks.

mpetrunic:
resolves #1985 
blocked by #1988 

Added:
- lodestar-core package to share common logic between end products like lodestar-validator, lodestar. lodestar-light-client
- producing and fetching both phase0 and lightclient blocks as well as their signing in lodestar-validator


dapplion:
> lodestar-core package to share common logic

Could you define more precisely what common logic should go into lodestar-core, and what should go into lodestar-utils?

mpetrunic:
> > lodestar-core package to share common logic
> 
> Could you define more precisely what common logic should go into lodestar-core, and what should go into lodestar-utils?

lodestar utils is basically no-dep utils because you cannot depend on lodestar-types or config or db because of circular dependcies while lodestar-core could be package that depends on most/all core lodestar packages (except end products like cli, validator, lightclient or lodestar itself) 
 and contain for example various serialization or hashing logic (depending on slot return ssz type for appropriate fork) , http/api client or shared interfaces.

dapplion:
> > > lodestar-core package to share common logic
> > 
> > 
> > Could you define more precisely what common logic should go into lodestar-core, and what should go into lodestar-utils?
> 
> lodestar utils is basically no-dep utils because you cannot depend on lodestar-types or config or db because of circular dependcies while lodestar-core could be package that depends on most/all core lodestar packages (except end products like cli, validator, lightclient or lodestar itself)
> and contain for example various serialization or hashing logic (depending on slot return ssz type for appropriate fork) , http/api client or shared interfaces.

Could you add this clarification to the lodestar-core and lodestar-utils README's?

packages/lodestar-core/test/setup.ts
```diff
@@ -0,0 +1,10 @@
+import {init} from "@chainsafe/bls";
+import {before} from "mocha";
+
+before(async function () {
+  try {
+    await init("blst-native");
+  } catch (e) {
+    console.log(e);
+  }
+});
```
dapplion: Can you use mocha hooks instead of setup file? It's currently the recommended way
Example: https://github.com/ChainSafe/lodestar/blob/master/packages/spec-test-runner/test/mocha-hooks.js
Ref: https://simplecheatsheet.com/mocha-js-root-hook-plugins-cheat-sheet/

packages/lodestar/test/unit/chain/forkChoice/forkChoice.test.ts
```diff
@@ -37,6 +37,7 @@ describe("LodestarForkChoice", function () {
       const finalizedRoot = config.types.BeaconBlockHeader.hashTreeRoot(blockHeader);
       const targetBlock = generateSignedBlock({message: {slot: 32}});
       targetBlock.message.parentRoot = finalizedRoot;
+      //
```
dapplion: Typo?

dapplion (approved):

packages/lodestar-core/test/setup.ts
mpetrunic: Cool. I've written hooks it in typescript though.

dapplion (dismissed):

package.json
```diff
@@ -63,7 +63,7 @@
     "prettier": "^2.0.5",
     "sinon": "^9.0.2",
     "supertest": "^4.0.2",
-    "ts-node": "^9.0.0",
+    "ts-node": "^9.1.1",
```
wemeetagain, line 66: Maybe best to keep this PR to specific changes

packages/lodestar-beacon-state-transition/src/slot.ts
```diff
@@ -2,15 +2,19 @@
...
 import {processEpoch} from "./epoch";
-
-export function processSlots(config: IBeaconConfig, state: BeaconState, slot: Slot): void {
+import {lightclient} from ".";
+
+export function processSlots<TState extends BeaconState = BeaconState>(
+  config: IBeaconConfig,
+  state: TState,
+  slot: Slot,
+  ignoreFork = false
```
wemeetagain: There shouldn't be an option to ignore the fork

packages/lodestar-beacon-state-transition/src/index.ts
```diff
@@ -36,9 +38,9 @@ export function stateTransition(
   const {verifyStateRoot = true, verifyProposer = true, verifySignatures = true} = options || {};
 
   // Clone state because process slots and block are not pure
-  const postState = config.types.BeaconState.clone(state);
+  let postState = config.types.BeaconState.clone(state);
   // Process slots (including those with no blocks) since block
-  processSlots(config, postState, signedBlock.message.slot);
+  postState = processSlots(config, postState, signedBlock.message.slot);
```
wemeetagain: `lightclientUpgrade` should happen here
> After process_slots of Phase 0 finishes, if state.slot == LIGHTCLIENT_PATCH_FORK_SLOT, an irregular state change is made to upgrade to light-client patch.

packages/lodestar-beacon-state-transition/src/util/fork.ts
```diff
@@ -22,13 +21,10 @@ export function computeForkDigest(
   return (root.valueOf() as Uint8Array).slice(0, 4);
 }
 
-export function epochToCurrentForkVersion(config: IBeaconConfig, epoch: Epoch): Version | null {
-  const forks = config.params.ALL_FORKS;
-  for (const fork of forks) {
-    if (epoch < fork.epoch) {
-      return intToBytes(fork.previousVersion, 4, "le");
-    }
+export function epochToCurrentForkVersion(config: IBeaconConfig, slot: Slot): Version | null {
```
wemeetagain: return type can just be `Version`

packages/lodestar-core/src/types/block.ts
```diff
@@ -0,0 +1,30 @@
+import {Lightclient, SignedBeaconBlock, BeaconBlock, Slot} from "@chainsafe/lodestar-types";
+import {ContainerType} from "@chainsafe/ssz";
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
+import {EMPTY_SIGNATURE} from "@chainsafe/lodestar-beacon-state-transition";
+
+export type SignedBeaconBlockType = SignedBeaconBlock | Lightclient.SignedBeaconBlock;
+export type BeaconBlockType = BeaconBlock | Lightclient.BeaconBlock;
+
+export function getSignedBeaconBlockSSZTypeBySlot(
```
wemeetagain, line 9: I thought we were going to go with:
`config.getTypes(epoch/slot).SignedBeaconBlock`

In that case we shouldn't need this `lodestar-core` library, at least not right now. Preference for not having it to make merging this branch easier when we need to.

wemeetagain (changes requested):
I think this will be easier to follow once we get #2034 merged here

mpetrunic:
Gonna start this in new PR/branch

PR #2028, opened 2021-02-08
Initialize persistent vector with array

### LINKED ISSUE #2027: RangeError: Maximum call stack size exceeded
Got this when restaring from a finalized state:

```
RangeError: Maximum call stack size exceeded
    at createCachedValidatorsBeaconState (/Users/tuyennguyen/Projects/workshop/lodestar/packages/lodestar-beacon-state-transition/src/fast/util/interface.ts:109:56)
    at restoreStateCaches (/Users/tuyennguyen/Projects/workshop/lodestar/packages/lodestar/src/chain/initState.ts:142:28)
    at Function.init (/Users/tuyennguyen/Projects/workshop/lodestar/packages/lodestar/src/node/nodejs.ts:127:11)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at Object.beaconHandler [as handler] (/Users/tuyennguyen/Projects/workshop/lodestar/packages/lodestar-cli/src/cmds/beacon/handler.ts:63:16)
```

lodestar passes so many arguments to persistent-ts, we may need it to accept an array instead

twoeths:
resolves #2027

### DIFF
--- a/packages/persistent-ts/src/Vector.ts
+++ b/packages/persistent-ts/src/Vector.ts
@@ -65,7 +65,7 @@ export class Vector<T> implements Iterable<T> {
    *
    * @param values the values that this vector will contain
    */
-  public static of<T>(...values: T[]): Vector<T> {
+  public static of<T>(values: T[]): Vector<T> {
     let acc = Vector.empty<T>();
     for (const v of values) acc = acc.append(v);
     return acc;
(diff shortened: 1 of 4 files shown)


packages/persistent-ts/src/Vector.ts
```diff
@@ -65,7 +65,7 @@ export class Vector<T> implements Iterable<T> {
    *
    * @param values the values that this vector will contain
    */
-  public static of<T>(...values: T[]): Vector<T> {
```
wemeetagain, line 68: Maybe best to call this `from` since it matches Array.from more closely

wemeetagain: Then the param could be `values: Iterable<T>`, even closer to Array.from

wemeetagain (approved):

PR #2030, opened 2021-02-08
Synch chain cache

wemeetagain:
- Move `StateContextCache` and `CheckpointStateCache` to `src/chain/stateCache/*`
- Move state caches to be managed by IBeaconChain
- Make all state cache methods synchronous, make downstream head accessors in chain synchronous
- Add eslint rule (set to warn) for showing unnecessary awaits


dapplion (approved):

dapplion:
Add https://eslint.org/docs/rules/require-await in a subsequent PR

PR #2032, opened 2021-02-08
implement getValidatorStatus and use it in getStateValidator(s)

3xtr4t3rr3str14l:
while working on #1986 , i noticed that we weren't getting the validator(s)'s actual state and just returning `"active"` unconditionally.  this PR fixes that

calculations of the various validator statuses based on https://github.com/ethereum/eth2.0-APIs/blob/e36b026ef1265dba9b9c6bc984236088d3759f87/types/api.yaml#L31 which is based on https://hackmd.io/ofFJ5gOmQpu1jjHilHbdQQ
i also updated the ValidatorStatus list in `lodestar-types` to match the eth2 api spec

### DIFF
--- a/packages/lodestar/src/api/impl/beacon/state/utils.ts
+++ b/packages/lodestar/src/api/impl/beacon/state/utils.ts
@@ -43,7 +43,7 @@ export async function resolveStateId(
  * Get the status of the validator
  * based on conditions outlined in https://hackmd.io/ofFJ5gOmQpu1jjHilHbdQQ
  */
-function getValidatorStatus(validator: Validator, currentEpoch: Epoch): ValidatorStatus {
+export function getValidatorStatus(validator: Validator, currentEpoch: Epoch): ValidatorStatus {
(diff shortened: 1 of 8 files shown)


packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -39,10 +39,49 @@ export async function resolveStateId(
   }
 }
 
-export function toValidatorResponse(index: ValidatorIndex, validator: Validator, balance: Gwei): ValidatorResponse {
+/**
+ * Get the status of the validator
+ * based on conditions outlined in https://hackmd.io/ofFJ5gOmQpu1jjHilHbdQQ
+ */
+function getValidatorStatus(validator: Validator, currentEpoch: Epoch): ValidatorStatus {
```
mpetrunic: could you add unit tests to cover all code paths of this function?

mpetrunic (approved):

PR #2053, opened 2021-02-17
Apply pruneFinalized() for StateContextCache

### LINKED ISSUE #2051: Proposal to store less State Context inside StateContextCache
As noted in #2005, we store too many State Contexts inside StateContextCache during initial sync unnecessarily

+ Create one more index inside StateContextCache to know the epoch of each State Contexts
+ do `pruneFinalized` if an epoch is finalized
+ keep doing the existing `prune` per check point and make sure `MAX_STATES` there

twoeths:
resolves #2051

### DIFF
--- a/packages/lodestar/src/chain/stateCache/stateContextCache.ts
+++ b/packages/lodestar/src/chain/stateCache/stateContextCache.ts
@@ -47,7 +47,11 @@ export class StateContextCache {
   }
 
   public delete(root: ByteVector): void {
-    delete this.cache[toHexString(root)];
+    const key = toHexString(root);
+    const item = this.cache[key];
+    if (!item) return;
+    this.epochIndex[item.epochCtx.currentShuffling.epoch].delete(key);
+    delete this.cache[key];
   }
 
   public batchDelete(roots: ByteVector[]): void {
@@ -73,6 +77,8 @@ export class StateContextCache {
       // object keys are stored in insertion order, delete keys starting from the front
       for (const key of keys.slice(0, keys.length - this.maxState)) {
         if (key !== headStateRootHex) {
+          const item = this.cache[key];
+          this.epochIndex[item.epochCtx.currentShuffling.epoch].delete(key);
           delete this.cache[key];
         }
       }
@@ -82,7 +88,7 @@ export class StateContextCache {
   /**
    * Prune per finalized epoch.
    */
-  public async pruneFinalized(finalizedEpoch: Epoch): Promise<void> {
+  public async deleteAllBeforeEpoch(finalizedEpoch: Epoch): Promise<void> {
     for (const epoch of Object.keys(this.epochIndex).map(Number)) {
       if (epoch < finalizedEpoch) {
         this.deleteAllEpochItems(epoch);
--- a/packages/lodestar/src/chain/stateCache/stateContextCache.ts
+++ b/packages/lodestar/src/chain/stateCache/stateContextCache.ts
@@ -1,6 +1,8 @@
 import {ByteVector, toHexString, TreeBacked} from "@chainsafe/ssz";
-import {phase0} from "@chainsafe/lodestar-types";
+import {phase0, Epoch} from "@chainsafe/lodestar-types";
 import {ITreeStateContext} from "../interface";
+
+const MAX_STATES = 96;
 
 /**
  * In memory cache of BeaconState and connected EpochContext
@@ -9,8 +11,17 @@ import {ITreeStateContext} from "../interface";
  */
 export class StateContextCache {
   private cache: Record<string, ITreeStateContext>;
-  constructor() {
+  /**
+   * Epoch -> Set<blockRoot>
+   */
+  private epochIndex: Record<Epoch, Set<string>>;
+
+  private maxState: number;
+
+  constructor(maxState = MAX_STATES) {
     this.cache = {};
+    this.epochIndex = {};
+    this.maxState = maxState;
   }
 
   public get(root: ByteVector): ITreeStateContext | null {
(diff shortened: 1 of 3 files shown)


packages/lodestar/src/chain/stateCache/stateContextCache.ts
```diff
@@ -1,16 +1,27 @@
...
+  /**
+   * Epoch -> Set<blockRoot>
+   */
+  private epochIndex: Record<Epoch, Set<string>>;
+
+  private maxState: number;
+
+  constructor(maxState = MAX_STATES) {
     this.cache = {};
+    this.epochIndex = {};
```
dapplion, line 23: Should we use a plain object or a Map?

packages/lodestar/src/chain/stateCache/stateContextCache.ts
```diff
@@ -22,9 +33,17 @@ export class StateContextCache {
...
+    this.cache[key] = this.clone(item);
+    const epoch = item.epochCtx.currentShuffling.epoch;
+    if (this.epochIndex[epoch]) {
+      this.epochIndex[epoch].add(key);
+    } else {
+      this.epochIndex[epoch] = new Set([key]);
+    }
   }
 
   public delete(root: ByteVector): void {
```
wemeetagain, line 49: need to update `delete` to remove the item from the epochIndex set if it exists

packages/lodestar/src/chain/stateCache/stateContextCache.ts
```diff
@@ -48,19 +67,29 @@ export class StateContextCache {
...
-    const MAX_STATES = 96;
     const keys = Object.keys(this.cache);
-    if (keys.length > MAX_STATES) {
+    if (keys.length > this.maxState) {
       const headStateRootHex = toHexString(headStateRoot);
       // object keys are stored in insertion order, delete keys starting from the front
-      for (const key of keys.slice(0, keys.length - MAX_STATES)) {
+      for (const key of keys.slice(0, keys.length - this.maxState)) {
         if (key !== headStateRootHex) {
           delete this.cache[key];
```
wemeetagain, line 82: need to update to remove the item from epochIndex set if it exists

packages/lodestar/src/chain/stateCache/stateContextCache.ts
```diff
@@ -48,19 +67,29 @@ export class StateContextCache {
...
         }
       }
     }
   }
 
+  /**
+   * Prune per finalized epoch.
+   */
+  public async pruneFinalized(finalizedEpoch: Epoch): Promise<void> {
```
wemeetagain: Might be better to rename as `deleteAllBeforeEpoch`

packages/lodestar/src/chain/stateCache/stateContextCache.ts
twoeths: I don't have a preference for it, just do consistently with other caches but this is a good point.

  For `cache`, maybe using `Map` is better than `Record` since looping through keys respect the insertion order in `Map` while `Object.keys(Record)` is not

```
const cache: Record<string, string> = {};
    cache["1"] = "1";
    cache["3"] = "3";
    cache["5"] = "5";
    cache["2"] = "2";
    cache["4"] = "4";
    // this proves Record does not respect insertion order
    expect(Object.keys(cache)).to.be.deep.equal(["1", "2", "3", "4", "5"]);
```

could you confirm @wemeetagain

wemeetagain: Map and Record both respect insertion order. From that perspective, it doesn't really matter.
I think the only place it matters is the type interface. I think Map has a 'safer' interface in that it can return undefined. IMO it doesn't really matter here as long as the implementation is correct tho :joy:

twoeths: but the test above shows that `Object.keys(Record)` does not respect insertion order

wemeetagain: the order inserted is 1 3 5 2 4, not 1 2 3 4 5

twoeths: the order inserted is 1 3 5 2 4 and Object.keys(Record) shows 1 2 3 4 5, we want it to be 1 3 5 2 4 too

wemeetagain: Oh I see... learn something new every day :thinking: 
I'm not too sure it matters for this purpose? Are we using the order somewhere?
I thought either way we'd have some code that looks like `deleteAllBeforeEpoch`

twoeths: it matters in `prune` flow https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/chain/stateCache/stateContextCache.ts#L56

I think let's switch to using `Map` instead

wemeetagain: looks like object uses the following rules:
    First, the keys that are integer indices, in ascending numeric order.
    Then, all other string keys, in the order in which they were added to the object.
    Lastly, all symbol keys, in the order in which they were added to the object.

wemeetagain: According to those rules, we likely don't need to change the cache implementation here. If we want to refactor both of the state caches in a future PR, go for it.

twoeths: agree

wemeetagain (approved):

PR #2056, opened 2021-02-18
Expose p2p port in docker-compose setup

dapplion:


docker-compose.yml
```diff
@@ -5,7 +5,9 @@ services:
     restart: always
     volumes:
       - lodestar:/root/.local/share/lodestar
-    command: beacon --network ${NETWORK:-pyrmont}
+    command: beacon --network ${NETWORK:-mainnet}
```
3xtr4t3rr3str14l: should this maybe be changed back to pyrmont?

docker-compose.yml
```diff
@@ -6,6 +6,8 @@ services:
     volumes:
       - lodestar:/root/.local/share/lodestar
     command: beacon --network ${NETWORK:-pyrmont}
+    ports:
+      - "9000:9000"
```
twoeths, line 10: this is disc5 port, and I still see we can detect new peers without this so I'm curious how could this help?

dapplion: If you don't expose this port you will get only outbound peers. Opening the p2p port so other nodes can connect to you is very recommended and necessary to test a network participant with inbound and outbound connections

dapplion: https://docs.prylabs.network/docs/prysm-usage/p2p-host-ip/#port-forwarding
https://lighthouse-book.sigmaprime.io/docker.html?highlight=port#ports

twoeths (approved):

Issue #1941, opened 2021-01-12
Fix Beacon x Validator docker compose setup


dapplion:
User @baysa from discord reported issues trying to connect the validator with the beacon chain on a dockerized environment.

Review and fix the issues + add a one liner to spin-up a working setup with both components. Current setup only includes a beacon node.

-----

Discord message https://discord.com/channels/593655374469660673/593655641445367808/798215178625351740

I want to run lodestar using docker compose. Beacon node (beacon-container) configurations:

beacon 
--testnet pyrmont 
--eth1.providerUrl http://geth:8545 
--rootDir /opt/app/beacon 
--api.rest.host 0.0.0.0 
--metrics.enabled true 
--metrics.serverPort 8008 
--api.rest.enabled true
--api.rest.port 9596

Validator (validator-container) configurations:

validator 
--testnet pyrmont 
--server http://beacon:9596 
--keystoresDir /opt/app/keystores 
--secretsDir /opt/app/secrets 
--validatorsDbDir /opt/app/validator-db

but it doesn't work for me, got error (validator-logs shown below). Anyone know how to fix this? thank you

![unknown](https://user-images.githubusercontent.com/35266934/104335857-15d58400-54f4-11eb-949e-108a847b3e59.png)

dapplion:
Fixed with https://github.com/ChainSafe/lodestar/pull/2057

PR #2057, opened 2021-02-18
Add validator service to docker-compose

3xtr4t3rr3str14l:
sets up a validator service on the existing docker-container


twoeths:
@3xtr4t3rr3str14l would this resolve https://github.com/ChainSafe/lodestar/issues/1941?

docker-compose.yml
```diff
@@ -5,7 +5,23 @@ services:
...
+    network_mode: host
+
+  validator:
+    build: .
+    restart: always
+    volumes:
+      - lodestar:/root/.local/share/lodestar
+      # NOTE: mapping of pyrmont volume assumes that a validator has been manually imported or created here
+      - /root/.local/share/lodestar/pyrmont:/root/.local/share/lodestar/pyrmont
+    command: validator --network ${NETWORK:-pyrmont} --graffiti "lodestar wuz here"
```
mpetrunic: you should set beacon node url to be `http://lodestar:9596` then you won't need `network_mode: host` which would be recommended to avoid.

docker-compose.yml
```diff
@@ -5,7 +5,23 @@ services:
     restart: always
     volumes:
       - lodestar:/root/.local/share/lodestar
+      - /root/.local/share/lodestar/pyrmont:/root/.local/share/lodestar/pyrmont
     command: beacon --network ${NETWORK:-pyrmont}
```
mpetrunic: you should set rest api host to be `0.0.0.0`

mpetrunic (dismissed):

docker-compose.yml
dapplion: :+1: For a configuration with network_mode: host now there's https://github.com/ChainSafe/lodestar/blob/master/docker/docker-compose-local.yml

The main docker-compose should use docker's DNS and only expose the necessary services

3xtr4t3rr3str14l: @mpetrunic before i found out about `network_mode: host` i had already tried what you suggested, and i wasn't able to get the dockerized beacon node to accept requests from the dockerized validator

3xtr4t3rr3str14l: ~~@dapplion so assuming i keep the `network_mode: host`, then do you think i should just add a beacon and validator service to that docker-compose-local.yml?~~

EDIT: going to use https://github.com/ChainSafe/lodestar/pull/2057#discussion_r579042163 instead

mpetrunic: You were probably missing this:
https://github.com/ChainSafe/lodestar/pull/2057#discussion_r579042163

3xtr4t3rr3str14l:
> @3xtr4t3rr3str14l would this resolve #1941?

@tuyennhv once i get this PR stable, i will run on contabo-2 to verify, but it looks like it might?  i did notice in the screenshot of that issue that it looks like the validator is connected to the api, but just not getting duties, so maybe that person's beacon node wasn't synced?

docker-compose.yml
```diff
@@ -5,17 +5,34 @@ services:
     restart: always
     volumes:
       - lodestar:/root/.local/share/lodestar
-    command: beacon --network ${NETWORK:-pyrmont}
+      - /root/.local/share/lodestar/pyrmont:/root/.local/share/lodestar/pyrmont
```
dapplion, line 8: Why are you mapping the volume twice, and only for the pyrmont network?

dapplion: Why are you mapping the volume twice, and only for the pyrmont network?

docker-compose.yml
```diff
@@ -5,17 +5,34 @@ services:
...
+    command: beacon --network ${NETWORK:-pyrmont} --api.rest.host 0.0.0.0
+    expose:
+      - "9596"
     ports:
       - "9000:9000"
 
+  validator:
+    build: .
```
dapplion: You should give the image a name, and the same in the lodestar service so it's only build once.

docker-compose.yml
```diff
@@ -5,17 +5,34 @@ services:
...
       - lodestar:/root/.local/share/lodestar
-    command: beacon --network ${NETWORK:-pyrmont}
+      - /root/.local/share/lodestar/pyrmont:/root/.local/share/lodestar/pyrmont
+    command: beacon --network ${NETWORK:-pyrmont} --api.rest.host 0.0.0.0
+    expose:
+      - "9596"
     ports:
       - "9000:9000"
 
+  validator:
```
dapplion: Rename lodestar service to beacon_node

docker-compose.yml
```diff
@@ -5,17 +5,34 @@ services:
     restart: always
     volumes:
       - lodestar:/root/.local/share/lodestar
-    command: beacon --network ${NETWORK:-pyrmont}
+      - /root/.local/share/lodestar/pyrmont:/root/.local/share/lodestar/pyrmont
+    command: beacon --network ${NETWORK:-pyrmont} --api.rest.host 0.0.0.0
+    expose:
+      - "9596"
```
dapplion: What do you want to achieve here? 9565 should be available to the validator service without this

> `expose`: Expose ports without publishing them to the host machine - they’ll only be accessible to linked services. Only the internal port can be specified.

https://docs.docker.com/compose/compose-file/compose-file-v3/#expose

docker-compose.yml
```diff
@@ -5,17 +5,34 @@ services:
...
   prometheus:
     build: docker/prometheus
     restart: always
     volumes:
       - "prometheus:/prometheus"
     depends_on:
       - lodestar
+    expose:
+      - "9090"
+      - "8008"
```
dapplion: What's the purpose of exposing this ports?

docker-compose.yml
```diff
@@ -5,17 +5,34 @@ services:
...
+    expose:
+      - "9596"
     ports:
       - "9000:9000"
 
+  validator:
+    build: .
+    restart: always
+    volumes:
+      - lodestar:/root/.local/share/lodestar
```
dapplion: The validator should use a different volume

dapplion (dismissed):

docker-compose.yml
3xtr4t3rr3str14l: I want to map the manually imported validators for pyrmont.  Do you have a suggestion for doing it a different way, or is this okay?

docker-compose.yml
3xtr4t3rr3str14l: i made the build happen in the beacon_node service and then used that image in validator.  if this is off in any way, i'm open to other ways

docker-compose.yml
dapplion: You should use the CLI flags to use a constant directory for the keystores independent of the network. This setup must work for mainnet and pyrmont

docker-compose.yml
```diff
@@ -1,21 +1,34 @@
...
-    build: .
+  beacon_node:
+    build: ../lodestar
+    image: lodestar/lodestar
     restart: always
     volumes:
-      - lodestar:/root/.local/share/lodestar
-    command: beacon
+      - beacon_node:/root/.local/share/lodestar
+      - /root/.local/share/lodestar/${NETWORK:-pyrmont}:/root/.local/share/lodestar/${NETWORK:-pyrmont}
```
dapplion: Why is it necessary to map the `beacon_node` lodestar dir to the host's?

docker-compose.yml
```diff
@@ -1,21 +1,34 @@
...
     ports:
       - "9000:9000"
 
+  validator:
+    image: lodestar/lodestar
+    restart: always
+    volumes:
+      - validator:/root/.local/share/validator
+      # NOTE: mapping of ${NETWORK} volume assumes that a validator has been manually imported or created here
+      - /root/.local/share/lodestar/${NETWORK:-pyrmont}:/root/.local/share/lodestar/${NETWORK:-pyrmont}
```
dapplion: Assuming `/root/.local/share/lodestar` exists in all hosts that may run this command will cause issues. Especially when running it in non-VPS setups. I would suggest to use a relative path here with some validator path flags

docker-compose.yml
```diff
@@ -1,21 +1,34 @@
...
+  beacon_node:
+    build: ../lodestar
+    image: lodestar/lodestar
     restart: always
     volumes:
-      - lodestar:/root/.local/share/lodestar
-    command: beacon
+      - beacon_node:/root/.local/share/lodestar
+      - /root/.local/share/lodestar/${NETWORK:-pyrmont}:/root/.local/share/lodestar/${NETWORK:-pyrmont}
+    command: beacon --network ${NETWORK:-pyrmont} --api.rest.host 0.0.0.0
```
dapplion: We should default to mainnet, to be consistent with the CLI default behaviour

docker-compose.yml
3xtr4t3rr3str14l: i'm not sure.  this was already there before.  i just changed the name from `lodestar` to `beacon_node`

3xtr4t3rr3str14l: but yeah, didn't seem necessary, so i removed it

dapplion:
@3xtr4t3rr3str14l I've made the setup compatible with different configs by using multiple commands documented in CONTRIBUTING.md. Could you give it a try and ensure it works as you intend?

3xtr4t3rr3str14l:
@dapplion LGTM.  just added the `/keystores` and `/secrets` to their respective flags.

docker-compose.yml
```diff
@@ -1,21 +1,24 @@
...
-  lodestar:
+  beacon_node:
     build: .
+    image: lodestar
     restart: always
     volumes:
-      - lodestar:/root/.local/share/lodestar
-    command: beacon
+      - beacon_node:/data
+    env_file: .env
```
wemeetagain, line 9: is the `cp default.env .env` step optional?

dapplion: No, it's mandatory. Otherwise docker-compose errors

docker-compose.yml
```diff
@@ -1,21 +1,24 @@
...
     ports:
-      - "9000:9000"
+      - "9000:9000" # P2P port
+      - "9596:9596" # REST API port
+    command: beacon --rootDir /data --api.rest.host 0.0.0.0
```
twoeths: it doesn't seem `NETWORK` env variable is carried all the way to the container, should we do `--network ${NETWORK}` here?

also metrics is disabled by default, I think we need to enable it here

dapplion: Oh right! The ENV has to be prefixed by `LODESTAR_` to be picked up by yargs

docker-compose.validator.yml
```diff
@@ -11,7 +11,7 @@ services:
       - ./keystores:/keystores
       - ./secrets:/secrets
     env_file: .env
-    command: validator --rootDir /data --keystoresDir /keystores --secretsDir /secrets --server http://beacon_node:9596
+    command: validator --rootDir /data --keystoresDir /keystores --secretsDir /secrets --server http://beacon_node:9596 --network ${NETWORK}
```
dapplion: Does it work with `LODESTAR_NETWORK` on the .env file?

3xtr4t3rr3str14l: No, I didn't have to use the `LODESTAR_` prefix at all.  I just passed in NETWORK=pyrmont in the .env file.

dapplion (dismissed):

wemeetagain:
Weird that the tests are all pending

wemeetagain (approved):

PR #2058, opened 2021-02-18
ReqResp does not return null in any case

dapplion:
reqResp request never returns null. For protocols with multiple responses, returns an array possibly empty. For single response protocols, returns 1 response or throw an error. Current types added unnecessary complexity for an out-of-spec situation

### DIFF
--- a/packages/lodestar/src/network/reqresp/request/collectResponses.ts
+++ b/packages/lodestar/src/network/reqresp/request/collectResponses.ts
@@ -1,6 +1,7 @@
 import {ResponseBody} from "@chainsafe/lodestar-types";
 import {Method} from "../../../constants";
 import {isRequestSingleChunk} from "../../util";
+import {RequestErrorCode, RequestInternalError} from "./errors";
 
 /**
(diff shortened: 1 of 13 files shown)


wemeetagain (approved):

packages/lodestar/src/network/reqresp/request/collectResponses.ts
```diff
@@ -12,13 +13,13 @@ import {isRequestSingleChunk} from "../../util";
...
   maxResponses?: number
-): (source: AsyncIterable<ResponseBody>) => Promise<T | null> {
+): (source: AsyncIterable<ResponseBody>) => Promise<T> {
   return async (source) => {
     if (isRequestSingleChunk(method)) {
       for await (const response of source) {
         return response as T;
       }
-      return null;
+      throw new RequestInternalError({code: RequestErrorCode.EMPTY_RESPONSE});
```
dapplion, line 22: methods that `isRequestSingleChunk` must return exactly 1 response chunk

PR #2061, opened 2021-02-18
Add bulldozer config

dapplion:
Automate merging PRs on an opt-in basis. Prevents having to come back to merge if reviews are ready before CI passing

### DIFF
--- a/.bulldozer.yml
+++ b/.bulldozer.yml
@@ -1,0 +1,33 @@
+# "version" is the configuration version, currently "1".
+version: 1
+
+# "merge" defines how and when pull requests are merged. If the section is
+# missing, bulldozer will consider all pull requests and use default settings.
+merge:
+  # "trigger" defines the set of pull requests considered by bulldozer. If
+  # the section is missing, bulldozer considers all pull requests not excluded
+  # by the ignore conditions.
+  trigger:
+    # Pull requests with any of these labels (case-insensitive) are added to
+    # the trigger.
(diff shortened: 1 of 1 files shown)


mpetrunic:
There is github option for that, bulldozer is no longer used:
image

dapplion:
> There is github option for that, bulldozer is no longer used:
> image

Yeah, but that would make it mandatory for all PRs. I would like to be an opt-in feature only. Is it possible with the native Github feature?

mpetrunic:
> > There is github option for that, bulldozer is no longer used:
> > image
> 
> Yeah, but that would make it mandatory for all PRs. I would like to be an opt-in feature only. Is it possible with the native Github feature?

probably not. but like, just don't ad second approval 😄

dapplion:
> probably not. but like, just don't ad second approval

Sometimes I do want to add my approval to a PR but I want to wait for Cayman's opinion for example before merging. This setup would prevent me from doing this.

wemeetagain:
> > probably not. but like, just don't ad second approval
> 
> Sometimes I do want to add my approval to a PR but I want to wait for Cayman's opinion for example before merging. This setup would prevent me from doing this.

Seems useful in that case. Probably won't be used every day, but occasionally, you know you want a PR merged, but don't want to wait around for the tests to finish.

wemeetagain (approved):

dapplion:
@wemeetagain Would be awesome to have this working. Can you take a look at the repo settings if it's all good, and the app is configured fine?

PR #2063, opened 2021-02-18
Abstract assertPeerRelevance function

dapplion:
Isolate logic to decide if a peer is relevant or not in a standalone function that return if relevant, throws if not. The assert format allows the consumer to log the reason.

Also tweaks the logic to align it with Lighthouse interpretation of the spec.

Attaches `createStatus` method to the chain instance. Reduces code duplication and allows for easier testing mocking a single method instead of having to stub imports.

### DIFF
--- a/packages/lodestar/src/sync/utils/assertPeerRelevance.ts
+++ b/packages/lodestar/src/sync/utils/assertPeerRelevance.ts
@@ -1,0 +1,110 @@
+import {computeStartSlotAtEpoch, getBlockRootAtSlot} from "@chainsafe/lodestar-beacon-state-transition";
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
+import {Epoch, ForkDigest, Root, Status} from "@chainsafe/lodestar-types";
+import {LodestarError} from "@chainsafe/lodestar-utils";
+import {toHexString} from "@chainsafe/ssz";
+import {IBeaconChain} from "../../chain";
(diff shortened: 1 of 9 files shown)


packages/lodestar/src/sync/utils/assertPeerRelevance.ts
```diff
@@ -0,0 +1,110 @@
...
+  | {code: IrrelevantPeerErrorCode.GENESIS_NONZERO; root: string}
+  | {code: IrrelevantPeerErrorCode.DIFFERENT_FINALIZED; expectedRoot: string; remoteRoot: string};
+
+export class IrrelevantPeerError extends LodestarError<IrrelevantPeerErrorType> {}
+
+/**
+ * Process a `Status` message to determine if a peer is relevant to us. If the peer is
+ * irrelevant the reason is returned.
+ */
+export async function assertPeerRelevance(remote: Status, chain: IBeaconChain, config: IBeaconConfig): Promise<void> {
```
wemeetagain, line 31: just an aside, i think `IBeaconChain` should publicly expose a `IBeaconChain#config`

wemeetagain (dismissed):
Looks good :+1:

packages/lodestar/src/sync/utils/assertPeerRelevance.ts
dapplion: That's a good idea!

mpetrunic (approved):

Issue #2046, opened 2021-02-15
Investigate long epoch transition during Pyrmont sync


twoeths:
I was able to catch a 4.84s epoch transition in my local environment using Chrome profiler
<img width="1124" alt="Screen Shot 2021-02-15 at 18 42 17" src="https://user-images.githubusercontent.com/10568965/107942360-b6b8d280-6fbd-11eb-976d-08a8bad6f2b1.png">

An overview:
+ prepareEpochProcessState takes 530ms ++, and there are some small executions happened after that
+ processRewardsAndPenalties takes 2.4s, getAttestationDeltas takes 1.17s, epochBalances setter takes 764ms
+ rotateEpoch takes more than 1s, computeEpochShuffling takes 687ms
+ hashTreeRoot takes 425ms

event loop lag somehow was at 1.5s only, I'm attaching the snapshot as well.
[CPU-20210215T150254-4s-epoch-transition.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/5981369/CPU-20210215T150254-4s-epoch-transition.cpuprofile.zip)

wemeetagain:
IMO we have likely milked as much performance out of things as reasonable and should just attempt to mitigate the slowness. 
We should just break up epoch processing with `await sleep(0);` to just prevent any single piece of the epoch transition from taking 

For instance, there's not much that can be done regarding re-hashing the balances. It's just computationally intensive work that has to be done.
Similar issue with just setting the state balances. The balances list needs to be converted to a tree, and that just takes time.

If we yield back to the event loop in between our longer-running subtasks, it likely won't even matter that the epoch transition takes time.

dapplion:
@wemeetagain I had a chat with Paul from Lighthouse and their epoch transitions take 20-30 ms, 100 times faster than ours. That's an insane difference that should be at least investigated.

wemeetagain:
It looks like we might not be iterating thru our cached validators/balances where we can?

twoeths:
`getAttesationDeltas()`/BigInt performance is really unpredictable (ranging from 98ms to 1375ms) as shown in PR #2064 , one possibility is to implement the function in assemblyscript. From a quick look, we can just  pass balances, config params (5 of them), EpochProcess (may pass 20 different params), finalized epoch and receive back new balances (if assemblyscript does not support passing objects as argument).

not sure if we want to do a POC for it at this time.

dapplion:
> not sure if we want to do a POC for it at this time.

[...]

PR #2064, opened 2021-02-19
Add getAttestationDeltas performance test

twoeths:
part of #2046 

+ Fix and restructure the current performance tests
+ Add `getAttestationDeltas.test.ts` showing that it may finish in more than 1s as in the profiler:

|#|minTime (ms)|maxTime (ms)|average (ms)|
|-|---------|---------|--------|
|1st|98|1375|130|
|2nd|97|699|131|

### DIFF
--- a/packages/lodestar-beacon-state-transition/test/perf/phase0/fast/block/block.test.ts
+++ b/packages/lodestar-beacon-state-transition/test/perf/phase0/fast/block/block.test.ts
@@ -3,8 +3,8 @@ import {SignedBeaconBlock, SignedVoluntaryExit} from "@chainsafe/lodestar-types"
 import {WinstonLogger} from "@chainsafe/lodestar-utils";
 import {List} from "@chainsafe/ssz";
 import {expect} from "chai";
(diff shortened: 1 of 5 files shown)


twoeths:
I'm not familiar with assembly-script. Since `getAttestationDeltas` mainly do`BigIntUintType({byteLength: 8})` operations, can we use `u64` type of assembly-script and implement the function in wasm @ChainSafe/eth2-0 ? at least we want it to be predictable.

mpetrunic:
> I'm not familiar with assembly-script. Since `getAttestationDeltas` mainly do`BigIntUintType({byteLength: 8})` operations, can we use `u64` type of assembly-script and implement the function in wasm @ChainSafe/eth2-0 ? at least we want it to be predictable.

We could but not sure if worth it since I think that method is removed in HF1 and replaced with https://github.com/ethereum/eth2.0-specs/blob/dev/specs/lightclient/beacon-chain.md#get_flag_deltas

It might be better to write new method in assemblyscript later on

wemeetagain (approved):

PR #2072, opened 2021-02-19
Remove unnecessary async functions and awaits

dapplion:
Removes all unnecessary async function expect to meet a required interface and omitting the api module entirely.

### DIFF
--- a/packages/lodestar/src/db/api/beacon/seenAttestationCache.ts
+++ b/packages/lodestar/src/db/api/beacon/seenAttestationCache.ts
@@ -18,22 +18,22 @@ export class SeenAttestationCache {
     this.cache = new Map();
   }
 
-  public async addCommitteeAttestation(attestation: phase0.Attestation): Promise<void> {
+  public addCommitteeAttestation(attestation: phase0.Attestation): void {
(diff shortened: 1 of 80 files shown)


packages/lodestar/src/db/api/beacon/seenAttestationCache.ts
```diff
@@ -18,22 +18,22 @@ export class SeenAttestationCache {
     this.cache = new Map();
   }
 
-  public async addCommitteeAttestation(attestation: Attestation): Promise<void> {
+  public addCommitteeAttestation(attestation: Attestation): void {
```
mpetrunic: I guess this all was async to allow easier change in implementation

dapplion: The chain API was made sync where possible too. ***When*** we change the implementation this can become async.

wemeetagain (approved):
I don't think this needs to be applied militantly, but generally ok with sucking the waste out.

Especially in places where we have interfaces, we often create APIs designed to avoid change. In those cases, it could make sense to expose an async method, anticipating a known need.

PR #2073, opened 2021-02-19
Namespace phase0 types

wemeetagain:
This namespaces ssz types under `config.types.phase0` or `import {phase0} from "@chainsafe/lodestar-types";`
This allows us to always refer unambiguously to a specific version of a type.
It also allows us to add types for future HFs without breaking the 'backward compatibility' of the existing types.

This PR creates `IPhase0Params` and `IPhase0SSZTypes` and integrates them in the params and types libraries in a standard way.

Future HF params are expected to define new params and never update the value of a pre-existing param. `IBeaconParams` will be the union of all HF params.
(eg: `type IBeaconParams = IPhase0Params & IHFParams ....;`)

Future HF ssz types are expected to define new types and possibly update existing types (but only by adding additional fields to those types).
`IBeaconSSZTypes` will have a separate namespace for each HF that allows for unambiguously selecting a type at a specific version.
(eg: `type IBeaconSSZTypes = { phase0: IPhase0SSZTypes; hf: IHFSSZTypes };`)
These same namespaces are exported from the root of the `@chainsafe/lodestar-types` to access the typescript types.
(eg: `import {phase0, hf} from "@chainsafe/lodestar-types";`)

For improved ergonomics, a few tweaks have been made.
Primitive types are also available directly on `IBeaconSSZTypes` and are also exported from the roof of `@chainsafe/lodestar-types`. This list of un-namespaced types may only ever grow to include types that we _know_ will never change, to protect backwards compatibility.

Additionally, phase0 typescript types have been re-exported under phase0 exports in `@chainsafe/lodestar-beacon-state-transition`. This allows importing a single `phase0` object from `@chainsafe/lodestar-beacon-state-transition` to both access phase0 state transition functions AND phase0 typescript types.


mpetrunic (approved):
SSo many changed files, even octotree extension has frozen haha

PR #2077, opened 2021-02-21
Enable single thread multi node test

dapplion:
Enables the multi-node single thread test. It provides valuable information if an issue is caused due to a networking issue or something related to multi-threading. It also reduces the node count from 4 to 2 to reduce the load of the test


wemeetagain:
I think we should keep testing with 4 nodes and also keep the multithreaded test.
Testing with 4 (rather than 2) nodes allows each node to track multiple peers to determine how to sync.
Testing in multiple threads allows us to simulate a 'full testnet' in different processes. Each node can process tasks without being synchronized to a single event queue.

dapplion:
> I think we should keep testing with 4 nodes and also keep the multithreaded test.
> Testing with 4 (rather than 2) nodes allows each node to track multiple peers to determine how to sync.
> Testing in multiple threads allows us to simulate a 'full testnet' in different processes. Each node can process tasks without being synchronized to a single event queue.

Maybe we should even go further and put the validators in separate processes too

dapplion:
> I think we should keep testing with 4 nodes and also keep the multithreaded test.
> Testing with 4 (rather than 2) nodes allows each node to track multiple peers to determine how to sync.
> Testing in multiple threads allows us to simulate a 'full testnet' in different processes. Each node can process tasks without being synchronized to a single event queue.

@wemeetagain Reverted to test 4 nodes

wemeetagain (approved):

PR #2089, opened 2021-02-22
Skip initial sync if already synced

dapplion:
Should help increase the stability of sim tests.

However does not fix the spammed error

```
 error: Failed to subscribe to committee subnet message=Cannot collect attestations before regular sync
Error: Cannot collect attestations before regular sync
```


wemeetagain:
closing in favor of #2090

PR #2090, opened 2021-02-23
Fix asssertPeerRelevance during pre-genesis

wemeetagain:
Peers were not connecting before genesis (due to clock skew between headSlot=0 and currentSlot=-N)

Fixes issues seen in sim tests

### DIFF
--- a/packages/lodestar/test/unit/sync/utils/assertPeerRelevance.test.ts
+++ b/packages/lodestar/test/unit/sync/utils/assertPeerRelevance.test.ts
@@ -99,4 +99,20 @@ describe("sync / utils / assertPeerRelevance", () => {
       }
     });
   }
+
+  it("Accept during pre-genesis clock", () => {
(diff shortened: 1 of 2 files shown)


packages/lodestar/test/unit/sync/utils/assertPeerRelevance.test.ts
```diff
@@ -99,4 +99,20 @@ describe("sync / utils / assertPeerRelevance", () => {
       }
     });
   }
+
+  it("Accept during pre-genesis clock", () => {
+    // clock slot pre-genesis (< 0) by a good margin
+    (chain as any).clock.currentSlot = -50;
```
mpetrunic, line 105: Would be nicer to not cast as any (in case we rename or move clock we won't detect error here) but unlikely to happen so I guess it's fine

mpetrunic (approved):

twoeths (approved):

Issue #2049, opened 2021-02-16
Epoch transition performance improvement - low hanging fruit


dapplion:
Research potential ways to improve epoch transition total CPU time

- We might already have the balances in an array, would not need to compute this again
https://github.com/ChainSafe/lodestar/blob/a5e8116af83df67faa0da2bff47d90f731fb523c/packages/lodestar-beacon-state-transition/src/fast/epoch/processRewardsAndPenalties.ts#L14
- Improve creating the tree from balances. We might be able to use a BigUint64Array, then just chop up the underlying buffer into Uint8Arrays and manually create the leaf nodes. This solution might be faster than the default SSZ implementation
- Investigate using a number instead of a BigInt when possible. For example for single validator deltas, math is much faster.
- Remove forEach, reduce for a regular for loop if iterating over very large arrays. Merge multiple iterations if possible.

Based on the profile of https://github.com/ChainSafe/lodestar/issues/2046

dapplion:
All useful ideas are already implemented

PR #2092, opened 2021-02-23
Improve getAttestationDeltas using Number

twoeths:
part of #2046 #2049

+ use Number instead of BigInt whenever possible to improve `getAttestationDeltas`
+ Average time to run `getAttestationDeltas` was reduced from 130 to 79

### DIFF
--- a/packages/lodestar-beacon-state-transition/src/phase0/fast/epoch/getAttestationDeltas.ts
+++ b/packages/lodestar-beacon-state-transition/src/phase0/fast/epoch/getAttestationDeltas.ts
@@ -1,4 +1,4 @@
-import {phase0, Gwei} from "@chainsafe/lodestar-types";
+import {phase0} from "@chainsafe/lodestar-types";
 import {bigIntSqrt, bigIntMax, intDiv} from "@chainsafe/lodestar-utils";
 import {BASE_REWARDS_PER_EPOCH as BASE_REWARDS_PER_EPOCH_CONST} from "../../../constants";
 
--- a/packages/lodestar-beacon-state-transition/src/phase0/fast/epoch/getAttestationDeltas.ts
+++ b/packages/lodestar-beacon-state-transition/src/phase0/fast/epoch/getAttestationDeltas.ts
@@ -20,11 +20,11 @@ export function getAttestationDeltas(
   epochCtx: EpochContext,
   process: IEpochProcess,
   state: phase0.BeaconState
-): [Gwei[], Gwei[]] {
+): [number[], number[]] {
   const params = epochCtx.config.params;
   const validatorCount = process.statuses.length;
-  const rewards = Array.from({length: validatorCount}, () => BigInt(0));
-  const penalties = Array.from({length: validatorCount}, () => BigInt(0));
+  const rewards = Array.from({length: validatorCount}, () => 0);
+  const penalties = Array.from({length: validatorCount}, () => 0);
 
   const increment = params.EFFECTIVE_BALANCE_INCREMENT;
   let totalBalance = bigIntMax(process.totalActiveStake, increment);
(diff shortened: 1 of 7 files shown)


packages/lodestar-beacon-state-transition/src/phase0/fast/epoch/getAttestationDeltas.ts
```diff
@@ -49,29 +49,30 @@ export function getAttestationDeltas(
 
   process.statuses.forEach((status, i) => {
     const effBalance = status.validator.effectiveBalance;
-    const baseReward = (effBalance * BASE_REWARD_FACTOR) / balanceSqRoot / BASE_REWARDS_PER_EPOCH;
-    const proposerReward = baseReward / PROPOSER_REWARD_QUOTIENT;
+    const baseReward = Number((effBalance * BASE_REWARD_FACTOR) / balanceSqRoot / BASE_REWARDS_PER_EPOCH);
```
wemeetagain, line 51: Number division should be replaced with `intDiv`

twoeths:
not able to pass spec tests using Number. Also `getAttestationDeltas` spec tests was so strict, currently we do `sum+= c1 * a / c2; sum += c1 * b / c;`, if I do `sum += c1 * (a+b) / c2;` it does not pass spec tests, I guess its due to BigInt calculation

dapplion:
> not able to pass spec tests using Number. Also `getAttestationDeltas` spec tests was so strict, currently we do `sum+= c1 * a / c2; sum += c1 * b / c;`, if I do `sum += c1 * (a+b) / c2;` it does not pass spec tests, I guess its due to BigInt calculation

What's the specific test breaking and the error message?

twoeths:
> > not able to pass spec tests using Number. Also `getAttestationDeltas` spec tests was so strict, currently we do `sum+= c1 * a / c2; sum += c1 * b / c;`, if I do `sum += c1 * (a+b) / c2;` it does not pass spec tests, I guess its due to BigInt calculation
> 
> What's the specific test breaking and the error message?

nvm, it works now. Thanks @wemeetagain for the great suggestion of using `intDiv`

packages/lodestar-beacon-state-transition/src/phase0/fast/epoch/processRewardsAndPenalties.ts
```diff
@@ -18,14 +18,14 @@ export function processRewardsAndPenalties(
   const newBalances = readOnlyMap(state.balances, (balance) => balance);
```
wemeetagain: As an additional optimization, you might try to apply the rewards and penalties at the same time.
eg:
```ts
const newBalances = readOnlyMap(state.balances, (balance, i) => {
  const newBalance = balance + BigInt(rewards[i] - penalties[i]);
  return newBalance < 0 ? BigInt(0) : newBalance;
});
```

twoeths: @wemeetagain this helps reduce the time in `processRewardsAndPenalties.test.ts` ~5%

packages/lodestar-beacon-state-transition/test/perf/phase0/fast/slot/slots.test.ts
```diff
@@ -27,7 +27,7 @@ describe("Process Slots Performance Test", function () {
     const start = Date.now();
     phase0.fast.processSlots(epochCtx, state, state.slot + numSlot);
     logger.profile(`Process ${numSlot} slots`);
-    expect(Date.now() - start).lt(1100);
+    expect(Date.now() - start).lt(570);
```
twoeths, line 30: with new ssz + this PR, the new epoch transition takes 500ms - 600ms in average instead of ~1100ms as previously

dapplion (approved):

wemeetagain (approved):

PR #2098, opened 2021-02-24
Add block processor utilization chart

dapplion:
Adds a metric to compute the total async time spent running async jobs in the block processor queue. With it, we can compute the utilization ratio with prometheus function `rate`.

This utilization ratio is very important for sync. It should be always 1 or the sync won't be as fast as it could be. If the sync manager doesn't handle peers correctly and is stuck often waiting for timeouts, the utilization ratio will drop indicating this problem.

### DIFF
--- a/packages/lodestar/src/util/queue.ts
+++ b/packages/lodestar/src/util/queue.ts
@@ -1,5 +1,5 @@
 import {AbortSignal} from "abort-controller";
-import pushable, {Pushable} from "it-pushable";
+import pushable from "it-pushable";
 import pipe from "it-pipe";
 import {LodestarError} from "@chainsafe/lodestar-utils";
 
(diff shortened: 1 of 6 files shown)


packages/lodestar/src/util/queue.ts
```diff
@@ -25,43 +25,52 @@ type JobQueueItem<T extends Job = Job> = {
...
       reject(new QueueError({code: QueueErrorCode.QUEUE_ABORTED}));
     } else {
+      const start = Date.now();
       try {
         const result = await job();
         resolve(result);
       } catch (e) {
         reject(e);
+      } finally {
+        this.currentSize--;
```
wemeetagain: previously, `this.currentSize` was decremented in both the if and else.

dapplion: Fixed

wemeetagain (approved):

PR #2105, opened 2021-02-26
Simplify API impl for node peers

dapplion:
- Take advantage of the Map returned by the libp2p connection manager to minimize data manipulation
- Don't calculate peerId from the key, since it's available in the remotePeer property
- SImplify filtering function in a single filter run
- Isolate parsing logic in formatNodePeer

This PR helps minimize the diff of the range-sync branch for a non-critical change. The method `network.getPeers` will be removed in that branch.

### DIFF
--- a/packages/lodestar/src/api/impl/node/node.ts
+++ b/packages/lodestar/src/api/impl/node/node.ts
@@ -2,12 +2,12 @@ import {phase0} from "@chainsafe/lodestar-types";
 import {createKeypairFromPeerId} from "@chainsafe/discv5";
 
 import {NodeIdentity, NodePeer} from "../../types";
-import {INetwork} from "../../../network";
+import {INetwork, PeerDirection, PeerState} from "../../../network";
 import {IBeaconSync} from "../../../sync";
 
 import {IApiOptions} from "../../options";
 import {ApiNamespace, IApiModules} from "../interface";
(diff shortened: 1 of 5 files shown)


packages/lodestar/src/api/impl/node/node.ts
```diff
@@ -43,30 +43,20 @@ export class NodeApi implements INodeApi {
...
-        enr: "",
-        lastSeenP2pAddress: conn ? conn.remoteAddr.toString() : "",
-        direction: conn ? conn.stat.direction : null,
-        state: conn ? getPeerState(conn.stat.status) : "disconnected",
-      });
-    }
-    return nodePeers;
+  public async getPeers(state?: PeerState[], direction?: PeerDirection[]): Promise<NodePeer[]> {
+    return Array.from(this.network.getConnectionsByPeer().entries())
+      .map(([peerIdStr, connections]) => formatNodePeer(peerIdStr, connections))
```
mpetrunic, line 54: It probably makes more sense to filter before map?

dapplion: The map run formats the objects from libp2p into our own API terms `state`, `direction`. So it's necessary to do that conversion first on all items to then be able to filter.

wemeetagain (approved):

PR #2111, opened 2021-02-26
Range sync refactor

### LINKED ISSUE #1921: Implement one-range-ahead strategy for initial sync
One range ahead for regular sync (ORARegularSync) turns out to be good in terms of performance and more maintainable code.

We should implement the same strategy for Initial Sync and maybe refactor ORARegularSync code a little bit to have some common code for both.

dapplion:
## TLDR

Implements Lighthouse's range sync which features:

- Able to go from synced to syncing dynamically depending on current connected peer status
- Resilient against malicious peers feeding a bad chain. Able to detect this case, keep track of the peers, report them and sync from another chain.
- Merges the initial sync + regular sync code.
- Able to sync multiple head chains at once.

#### `RangeSync`

Sync is triggered by adding and removing peers, so it's convenient to follow the logic from `addPeer` https://github.com/ChainSafe/lodestar/blob/03f356b35be94c51985dc1be87c1489da3d54790/packages/lodestar/src/sync/range/range.ts#L105

Key sensitive logic:
- [`updateChains`](https://github.com/ChainSafe/lodestar/blob/dapplion%2Frange-sync-rebased/packages/lodestar/src/sync/range/utils/updateChains.ts): How to decide with SyncChain to sync from

----

Fixes https://github.com/ChainSafe/lodestar/issues/1921


packages/lodestar/src/network/peers/discover.ts
```diff
@@ -0,0 +1,159 @@
...
+};
+
+export type PeerDiscoveryModules = {
+  libp2p: LibP2p;
+  peerRpcScores: IPeerRpcScoreStore;
+  logger: ILogger;
+  config: IBeaconConfig;
+};
+
+export class PeerDiscovery {
```
wemeetagain: Should `PeerDiscovery` take an abort signal?

Edit: It looks like its relying on libp2p and discv5 to properly abort.

wemeetagain: Would be good to have a class-level documentation here. Eg:
```ts
/**
 * PeerDiscovery discovers unconnected peers and issues connection requests
 */
```

packages/lodestar/src/network/peers/utils/peerMapDelay.ts
```diff
@@ -0,0 +1,36 @@
...
+  /** timeToReq = now() -> Request after `INTERVAL` */
+  requestAfter(peer: PeerId, ms = this.interval): void {
+    this.lastMsMap.set(peer, Date.now() - this.interval + ms);
+  }
+
+  /** Return array of peers with expired interval + call requestAfter on them */
+  pollNext(): PeerId[] {
+    const peers: PeerId[] = [];
+    for (const [peer, lastMs] of this.lastMsMap.entries()) {
+      if (Date.now() - lastMs > this.interval) {
```
wemeetagain: Maybe `Date.now()` should be pulled out of the for loop so we're fairly treating every peer with the same time.
```ts
const now = Date.now();
...
  if (now - lastMs > this.interval) {
```

Edit: Will it be a problem that each peer in this loop that triggers `this.requestAfter` will use a slightly different "now"? Should `requestAfter` look like `requestAfter(peer: PeerId, ms = this.interval, now = Date.now()): void`?

Broader question with this comment: The order that peers are added to the map effects the order they're looped, and ultimately, the values in the map (based on use of `Date.now`). Do we need to pre-calc the `Date.now()` when we're handling these peers in this loop?

packages/lodestar/src/network/peers/utils/peerMapDelay.ts
```diff
@@ -0,0 +1,36 @@
+import PeerId from "peer-id";
+import {PeerMap} from "../../../util/peerMap";
+
+export class PeerMapDelay {
```
wemeetagain: Can you add a class level description?

packages/lodestar/src/network/peers/discover.ts
```diff
@@ -0,0 +1,159 @@
+import PeerId from "peer-id";
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
+import {ILogger} from "@chainsafe/lodestar-utils";
+import {Discv5, Discv5Discovery} from "@chainsafe/discv5";
+import {shuffle} from "../../util/shuffle";
+import {getConnectedPeerIds} from "./utils";
+import {IPeerRpcScoreStore, ScoreState} from "./score";
+
+export type Discv5Query = {subnetId: number; maxPeersToDiscover: number};
```
wemeetagain: Might be better to call this `SubnetQuery`, `AttSubnetQuery`, or similar.

Also, just thinking about the fact that new committees will be introduced (eg: sync committee) and there may be ways to determine if peers are in those commitees (eg: additions to ENR or metadata endpoint). Definitely not required to solve this now, just FYI.

packages/lodestar/src/network/peers/utils/prioritizePeers.ts
```diff
@@ -0,0 +1,84 @@
...
+export function prioritizePeers(
+  connectedPeers: {id: PeerId; attnets: phase0.AttestationSubnets; score: number}[],
+  activeSubnetIds: number[],
+  {targetPeers, maxPeers}: {targetPeers: number; maxPeers: number}
+): {peersToDisconnect: PeerId[]; peersToConnect: number; discv5Queries: Discv5Query[]} {
+  const peersToDisconnect: PeerId[] = [];
+  let peersToConnect = 0;
+  const discv5Queries: Discv5Query[] = [];
+
+  // Dynamically compute 1 >= TARGET_PEERS_PER_SUBNET <= MAX_TARGET_SUBNET_PEERS
```
wemeetagain: assuming this should be 
```suggestion
  // Dynamically compute 1 <= TARGET_PEERS_PER_SUBNET <= MAX_TARGET_SUBNET_PEERS
```

packages/lodestar/src/network/peers/utils/prioritizePeers.ts
```diff
@@ -0,0 +1,84 @@
...
+  const targetPeersPerSubnet = Math.min(
+    MAX_TARGET_SUBNET_PEERS,
+    Math.max(1, Math.floor(maxPeers / activeSubnetIds.length))
+  );
+
+  // To filter out peers that are part of 1+ attnets of interest
```
wemeetagain:
```suggestion
  // To filter out peers that are part of 1+ attnets of interest from possible disconnection
```

packages/lodestar/src/network/peers/utils/subnetMap.ts
```diff
@@ -0,0 +1,41 @@
+import {Slot} from "@chainsafe/lodestar-types";
+
+export type RequestedSubnet = {
+  subnetId: number;
+  /**
+   * Slot after which the network will stop to mantain a min number of peers
```
wemeetagain:
```suggestion
   * Slot after which the network will stop maintaining a min number of peers
```

packages/lodestar/src/network/peers/utils/subnetMap.ts
```diff
@@ -0,0 +1,41 @@
...
+   *  connected to `subnetId`
+   */
+  toSlot: Slot;
+};
+
+/**
+ * Track request subnets by `toSlot`
```
wemeetagain:
```suggestion
 * Track requested subnets by `toSlot`
```

packages/lodestar/src/network/reqresp/reqResp.ts
```diff
@@ -62,21 +59,16 @@ export class ReqResp implements IReqResp {
         this.libp2p.handle(createRpcProtocol(method, encoding), async ({connection, stream}) => {
           const peerId = connection.remotePeer;
 
-          // Store peer encoding preference for this.sendRequest
+          // TODO: Do we really need this now that there is only one encoding?
```
wemeetagain: :shrug:

packages/lodestar/src/network/peers/peerManager.ts
```diff
@@ -0,0 +1,443 @@
...
+import {
+  getConnectedPeerIds,
+  PeerMapDelay,
+  SubnetMap,
+  RequestedSubnet,
+  assertPeerRelevance,
+  prioritizePeers,
+} from "./utils";
+
+export interface IPeerManager {
```
wemeetagain: Just an aside:
I'm not sure how useful the `INetwork`, `IPeerManager`, etc. interfaces are.

packages/lodestar/src/network/peers/peerManager.ts
```diff
@@ -0,0 +1,443 @@
...
+/**
+ * Tasks:
+ * - Ping peers every `PING_INTERVAL_MS`
+ * - Status peers every `STATUS_INTERVAL_MS`
+ * - Execute discovery query if under target peers
+ * - Execute discovery query if need peers on some subnet: TODO
+ * - Disconnect peers if over target peers
+ */
+export class PeerManager implements IPeerManager {
+  // TODO: Reorg - TEMP
```
wemeetagain: ?

packages/lodestar/src/network/peers/peerManager.ts
```diff
@@ -0,0 +1,443 @@
...
+
+  start(): void {
+    this.libp2p.connectionManager.on(Libp2pEvent.peerConnect, this.onLibp2pPeerConnect);
+    this.libp2p.connectionManager.on(Libp2pEvent.peerDisconnect, this.onLibp2pPeerDisconnect);
+    this.networkEventBus.on(NetworkEvent.reqRespRequest, this.onRequest);
+
+    // On start-up will connected to existing peers in libp2p.peerStore, same as autoDial behaviour
+    this.heartbeat();
+    this.intervals = [
+      setInterval(() => this.pingAndStatusTimeouts(), 2 * 1000),
```
wemeetagain: Maybe pull `2 * 1000` into a constant?

packages/lodestar/src/network/peers/peerManager.ts
```diff
@@ -0,0 +1,443 @@
...
+  /** Map of PeerId -> Time of last PING'd request in ms */
+  private peersToPingOutbound = new PeerMapDelay(PING_INTERVAL_INBOUND_MS);
+  private peersToPingInbound = new PeerMapDelay(PING_INTERVAL_OUTBOUND_MS);
+  /** Map of PeerId -> Time of last STATUS'd request in ms */
+  private peersToStatus = new PeerMapDelay(STATUS_INTERVAL_MS);
+  private opts: PeerManagerOpts;
+  private intervals: NodeJS.Timeout[] = [];
+
+  /** Map of subnets and the slot until they are needed */
+  private subnets = new SubnetMap();
```
wemeetagain: So we no longer track peer duty duration? Just tracking subnet duty duration?

packages/lodestar/src/sync/sync.ts
```diff
@@ -1,113 +1,85 @@
...
       modules.gossipHandler || new BeaconGossipHandler(modules.chain, modules.network, modules.db, this.logger);
     this.attestationCollector = modules.attestationCollector || new AttestationCollector(modules.config, modules);
-    this.mode = SyncMode.STOPPED;
-    this.processingRoots = new Set();
+    this.slotImportTolerance = modules.config.params.SLOTS_PER_EPOCH;
+
+    this.rangeSync.on(RangeSyncEvent.completedChain, this.updateSyncState);
   }
 
   public async start(): Promise<void> {
```
wemeetagain: We can merge `start` into the `constructor` now

packages/lodestar/src/sync/interface.ts
```diff
@@ -1,30 +1,33 @@
...
-  STOPPED,
-}
+import {SyncChainDebugState} from "./range/chain";
+export {SyncChainDebugState};
 
 export interface IBeaconSync {
-  state: SyncMode;
-  start(): Promise<void>;
-  stop(): Promise<void>;
+  state: SyncState;
```
wemeetagain, line 14: Add `close` method?

packages/lodestar/src/node/nodejs.ts
```diff
@@ -227,7 +228,7 @@ export class BeaconNode {
     if (this.status === BeaconNodeStatus.started) {
       this.status = BeaconNodeStatus.closing;
       await this.chores.stop();
-      await (this.sync as BeaconSync).stop();
+      (this.sync as BeaconSync).close();
```
wemeetagain: If `BeaconSync#start` is removed and `close` added to `IBeaconSync` this file can be cleaned up

packages/lodestar/src/sync/range/chain.ts
```diff
@@ -16,119 +22,225 @@ import {
...
-/**
- * Blocks are downloaded in batches from peers. This constant specifies how many epochs worth of
- * blocks per batch are requested _at most_. A batch may request less blocks to account for
- * already requested slots. There is a timeout for each batch request. If this value is too high,
- * we will negatively report peers with poor bandwidth. This can be set arbitrarily high, in which
- * case the responder will fill the response up to the max request size, assuming they have the
- * bandwidth to do so.
- */
-const EPOCHS_PER_BATCH = 2;
+export class SyncChain {
```
wemeetagain, line 81: class-level comment

packages/lodestar/src/sync/range/range.ts
```diff
@@ -0,0 +1,256 @@
...
+  /** Full debug state for lodestar API */
+  getSyncChainsDebugState(): SyncChainDebugState[] {
+    return Array.from(this.chains.values())
+      .map((syncChain) => syncChain.getDebugState())
+      .reverse(); // Newest additions first
+  }
+
+  /** Convenience method for `SyncChain` */
+  private processChainSegment: SyncChainFns["processChainSegment"] = async (blocks) => {
+    const trusted = true; // TODO: Verify signatures
```
wemeetagain, line 184: On my local computer:
w/o signature verification -> 14 slots/s
w/ signature verification -> 4.5 slots/s (also seems to introduce lots of choppiness in grafana reporting)

wemeetagain (changes requested):
:100: :rocket: This looks really good, mostly small comments.

I'm running this locally on pyrmont, will update with how it goes.

packages/lodestar/src/network/peers/discover.ts
dapplion: Right, `PeerDiscovery` does not execute queries now. Relies on discv5 automatic periodic queries and feeds from cached kadValues.

packages/lodestar/src/network/peers/utils/peerMapDelay.ts
dapplion: I see your point but I'm not sure it would have any impact. This data structure is somewhat equivalent to discv5 timeout map, but with a single setInterval in the parent class. If there are small differences in the timing it can cause some peers to be poll'ed in the next interval run. If a peer is ping'ed or status'ed -+ 2 seconds shouldn't matter much.

If calling `Date.now()` has some cost then we can optimize it as you propose.

packages/lodestar/src/network/reqresp/reqResp.ts
dapplion: :woman_shrugging:

packages/lodestar/src/network/peers/peerManager.ts
dapplion: Indeed the `IPeerManager` is useless. It can be a private property on Network. However `Network` is used everywhere, and referencing the interface instead of the actual class maintains allows for some flexibility and is consistent with other code. We can discuss in an issue our pattern of interfaces.

packages/lodestar/src/network/peers/peerManager.ts
dapplion: Yes, I think it's easier since you can compute `activeSubnetIds: number[]` from the `SubnetMap` and then contrast it with the current peer's metadata. It requires no intermediate cached state, which is what the peer duty duration was.

https://github.com/ChainSafe/lodestar/blob/03f356b35be94c51985dc1be87c1489da3d54790/packages/lodestar/src/network/peers/utils/prioritizePeers.ts#L40-L42

packages/lodestar/src/network/peers/discover.ts
```diff
@@ -0,0 +1,163 @@
...
+    this.logger = modules.logger;
+    this.config = modules.config;
+    this.maxPeers = opts.maxPeers;
+  }
+
+  /**
+   * Request to find peers. First, looked at cached peers in peerStore
+   */
+  discoverPeers(maxPeersToDiscover: number): void {
+    const notConnectedPeers = Array.from(this.libp2p.peerStore.peers.values()).filter(
```
wemeetagain: probably a small thing, but `peerStore.peers` has been an expensive call in the past.

dapplion: @wemeetagain What alternatives does libp2p offer to get a full list of available peers? If this call is expensive we can consider using an alternative peerStore

wemeetagain: libp2p doesn't provide anything other than a few Map objects, all keyed by peerIdStr
If you look at the implementation, you should see how you can get the keys w/o the extra bloat.

Ideally, eventually libp2p moves entirely to peerIdStr identifiers instead of PeerId objects.

wemeetagain:
I ran this locally and the sync stalled, also visible is OOM


[sync-chains-debug-state](https://cdn.discordapp.com/attachments/726486475943837786/815003557752733696/message.txt)

twoeths:
I also got OOM after syncing Pyrmont for only 2 hours

3xtr4t3rr3str14l:
i got OOM too after a few hours on node v12.18.3

wemeetagain:
I'm not sure that this branch, specifically is causing the OOM. I'm seeing the same thing on master.

packages/lodestar/src/network/peers/discover.ts
dapplion: You warned https://github.com/ChainSafe/lodestar/pull/2238

wemeetagain:
Whats the status of this? I'd love to get it merge-ready if it is stable.

dapplion:
> Whats the status of this? I'd love to get it merge-ready if it is stable.

This PR has a fundamental flaw caused by our high peer churn (which Lighthouse doesn't have). Tackling this issue is not trivial, and I feel we have more pressing priorities right now than this. I would love to get it merged soon tho, but maybe in a few weeks or after the hackaton

dapplion:
Closed for https://github.com/ChainSafe/lodestar/pull/2464

PR #2123, opened 2021-03-02
Fix params creation

### LINKED ISSUE #2120: Config parser error on every node restart
The lodestar-cli stores an invalid config YAML to disk with
```yaml
# ~/.local/share/lodestar/mainnet/config.yaml
'PHASE_1_FORK_VERSION': '0x01000000'
'PHASE_1_FORK_SLOT': 'undefined'
```
The string "undefined" is not a number so it always throws when starting the node with
```
 ✖ SyntaxError: Cannot convert undefined to a BigInt
    at BigInt (<anonymous>)
    at NumberUintType.fromJson (/root/lodestar/node_modules/@chainsafe/ssz/src/types/basic/uint.ts:90:18)
    at /root/lodestar/packages/lodestar-params/lib/utils.js:11:43
    at Array.forEach (<anonymous>)
    at Object.createIBeaconParams (/root/lodestar/packages/lodestar-params/lib/utils.js:8:56)
    at getBeaconParams (/root/lodestar/packages/lodestar-cli/lib/config/beaconParams.js:56:58)
    at getBeaconParamsFromArgs (/root/lodestar/packages/lodestar-cli/lib/config/beaconParams.js:24:12)
    at Object.getBeaconConfigFromArgs (/root/lodestar/packages/lodestar-cli/lib/config/beaconParams.js:16:50)
    at Object.initializeOptionsAndConfig (/root/lodestar/packages/lodestar-cli/lib/cmds/init/handler.js:41:29)
    at Object.beaconHandler [as handler] (/root/lodestar/packages/lodestar-cli/lib/cmds/beacon/handler.js:21:59)
```

---

Seems that it's SSZ the one doing the serialization. Any idea how to fix this?
https://github.com/ChainSafe/lodestar/blob/423124c3bdfbb86a852a67518d6ad60798c699c6/packages/lodestar-cli/src/config/beaconParams.ts#L85

wemeetagain:
Fixes #2120 

Sets lightclient and phase1 fork slot to `Infinity`


packages/lodestar-params/src/presets/mainnet/phase1.ts
```diff
@@ -8,7 +8,7 @@ export const phase1Json = {
   // ---------------------------------------------------------------
   PHASE_1_FORK_VERSION: "0x01000000",
   // [STUB]
-  PHASE_1_FORK_SLOT: 0,
+  PHASE_1_FORK_SLOT: "99999999999",
```
3xtr4t3rr3str14l: why make this a string if the type is a number?  `INITIAL_ACTIVE_SHARDS` (just below this one) is a number and has a number value as well

wemeetagain: I just copied the value in lightclient.ts, `LIGHTCLIENT_PATCH_FORK_SLOT`
Shouldn't really matter either way, but these objects mirror the yaml config in the specs (https://github.com/ethereum/eth2.0-specs/tree/dev/configs).
`createIBeaconParams` treats each value as a `Json`, and coerces it to the right type. (this is more important for byte arrays, eg: `0x01000000` -> `Uint8Array[1, 0, 0, 0]`)

3xtr4t3rr3str14l (approved):

Issue #2127, opened 2021-03-04
blocks by range responds with orphaned blocks


dapplion:
AgeManning reported that a Lighthouse node in Pyrmont was served orphaned blocks and thus banned us. Investigate.

> We've also noticed that when syncing from lodestar you guys dont seem to send us blocks from the canonical chain from blocks by range requests
You seem to send orphaned or blocks that don't exist in the chain back and we are kicking you guys
Think this was Pyrmont. When sync errors we keep track of the client and the cause

twoeths:
from https://github.com/sigp/lighthouse/blob/v1.2.0/beacon_node/network/src/sync/range_sync/chain.rs#L682

regarding `beacon_block_by_range`, what I see is we enforce canonical blocks pretty well for archived blocks and unarchived blocks but we need to do more to make sure the concatenation of these 2 arrays/streams (when handle request) form a complete linear chain segment that other clients can process for `step=1`

I'm thinking about the edge case when we handle a `beacon_block_by_range` request and run `archiveBlocks` task for that same range at the same time, this could result in a chain segment with missing blocks.

dapplion:
We could do a last loop over all blocks before sending to ensure that all blocks are part of the chain or drop them

twoeths:
> We could do a last loop over all blocks before sending to ensure that all blocks are part of the chain or drop them

it works or `step=1` but not for `step>1`, I think we want to return correct results for all steps. Maybe we want to coordinate with `ArchiveBlocksTask` to return error if requested range overlaps with `ArchiveBlocksTask` range?

dapplion:
> it works or step=1 but not for step>1

What's the issue with step > 1 ?

twoeths:
> > it works or step=1 but not for step>1
> 
> What's the issue with step > 1 ?

I mean we can't check parent-child relationship before we return if `step>1`

dapplion:
> I mean we can't check parent-child relationship before we return if step>1

Can we ask the forkchoice if blocks belong to our current head chain?

twoeths:
> > I mean we can't check parent-child relationship before we return if step>1
> 
> Can we ask the forkchoice if blocks belong to our current head chain?

forkchoice is pruned frequently per finalized checkpoint so we can't count on it

dapplion:
Current version returns only canonical blocks

PR #2130, opened 2021-03-05
Set state cache size to 1 during initial sync

wemeetagain:


mpetrunic (approved):

twoeths:
in the past I used to see some peers sent us orphaned blocks during initial sync so should we increase it a little bit? not sure if it happens again through

dapplion:
> in the past I used to see some peers sent us orphaned blocks during initial sync so should we increase it a little bit? not sure if it happens again through

Is this related? https://github.com/ChainSafe/lodestar/issues/2127

twoeths:
> > in the past I used to see some peers sent us orphaned blocks during initial sync so should we increase it a little bit? not sure if it happens again through
> 
> Is this related? #2127

yeah it's same issue but I'm not sure whose nodes I received data from

wemeetagain:
> in the past I used to see some peers sent us orphaned blocks during initial sync so should we increase it a little bit? not sure if it happens again through

We should be able to handle orphaned blocks using the regen. In that case, we'd just replay a few blocks starting from the last checkpoint state.

PR #2131, opened 2021-03-05
Lightclient patch spec tests

mpetrunic:
- missing ssz static tests - not sure how I'm suppose modify them to accomodate hf1
- removed process_final_updates fast spec tests - method was refactored in phase0 that we didn't reflect in our code


packages/lodestar-beacon-state-transition/src/util/blockRoot.ts
```diff
@@ -25,20 +27,27 @@ export function getBlockRootAtSlot(config: IBeaconConfig, state: phase0.BeaconSt
...
+    type = config.types.phase0.BeaconBlockBody;
+  }
   return {
     slot: block.slot,
     proposerIndex: block.proposerIndex,
     parentRoot: block.parentRoot,
     // `state_root` is zeroed and overwritten in the next `process_slot` call
     stateRoot: ZERO_HASH,
-    bodyRoot: config.types.phase0.BeaconBlockBody.hashTreeRoot(block.body),
+    bodyRoot: type.hashTreeRoot(block.body as lightclient.BeaconBlockBody),
```
wemeetagain: Should be able to do this without passing in a type with: `config.getTypes(block.slot).BeaconBlockBody.hashTreeRoot(block.body)`

packages/lodestar-beacon-state-transition/src/lightclient/naive/sync_committee.ts
```diff
@@ -73,7 +76,7 @@ export function processSyncCommittee(
  */
 export function getSyncCommitteeIndices(
   config: IBeaconConfig,
-  state: lightclient.BeaconState,
+  state: lightclient.BeaconState | phase0.BeaconState,
```
wemeetagain: Can you use `import {allForks} from "@chainsafe/lodestar-types";` `allForks.BeaconState` (and `allForks.BeaconBlock`) in all these places? Trying to use that everywhere for "multi-fork" types.

Feel free to remove `phase1.BeaconState` from `allForks.BeaconState`, for now, if it causes trouble.

wemeetagain (changes requested):

packages/lodestar-beacon-state-transition/src/util/blockRoot.ts
mpetrunic: You would think that but spec test blocks don't have slot after hf1 fork slot

wemeetagain: we just need to create new `IBeaconConfig`s for the spec tests.
```ts
import {params as minimalParams} from "@chainsafe/lodestar-params/minimal";
import {params as mainnetParams} from "@chainsafe/lodestar-params/mainnet";

export const lightclientSpecTestMinimal = createIBeaconConfig({...minimalParams, LIGHTCLIENT_PATCH_FORK_SLOT: 0});
export const lightclientSpecTestMainnet = createIBeaconConfig({...mainnetParams, LIGHTCLIENT_PATCH_FORK_SLOT: 0});
```

wemeetagain (approved):

PR #2132, opened 2021-03-06
Refactor gossip code

### LINKED ISSUE #2083: Simplify Gossipsub validator fn Map
In Gossipsub we register the same function for all topics. Then this function has to parse the incoming topic and decide which dedicated validator fn to use.

Would it be simpler / possible to directly register validator fn A for topic A and remove all the parsing?

Ref: https://github.com/ChainSafe/lodestar/blob/c013fbf6c3c7b623a6f476cbd5a7544026e009c3/packages/lodestar/src/network/gossip/gossipsub.ts#L83-L91

wemeetagain:
Refactors gossip code and adds easier support for supporting future HFs

- A few general fork helpers, the `IForkInfo` interface, `config.getForkInfoRecord()`, and `computeForkNameFromForkDigest`
- Remove outer `Gossip` wrapper, all logic is on `Eth2Gossipsub`, a class that extends `Gossipsub`.
- Rework gossip topic typing, add type `GossipTopic` that supports referencing a topic in a friendly way
- Generally, validation functions and gossip event handlers are added on initialization, removed on destruction. (rather than added/removed during subscription/unsubscription)
- Subscription/unsubscription is toggled by starting/stopping the sync gossip handler, attestation collector, and interopSubnetTask
- Generally, `ForkDigest` event handlers can/should be replaced with `IForkName` event handlers.
  - in many cases, the way to handle events is known ahead of time in a per-fork manner
  - fork digest is not immediately useful, rather, its derivative of the fork version/name
  - since `genesisValidatorsRoot` is a constant, we can code in terms of `IForkName` and derive a `ForkDigest` if/when/where its needed
- Adds a few metrics:
Screenshot from 2021-03-05 18-52-47

Fixes https://github.com/ChainSafe/lodestar/issues/2083


packages/lodestar/src/network/gossip/encoding.ts
```diff
@@ -1,65 +1,92 @@
...
-    case GossipEncoding.SSZ_SNAPPY:
+    case GossipEncoding.ssz_snappy:
       return uncompress(data);
 
-    default:
+    case GossipEncoding.ssz:
       return data;
+
+    default:
+      throw new Error(`Unable to decode message with encoding ${encoding}`);
```
dapplion: Message should be `Unsupported encoding ${encoding}`

packages/lodestar/src/network/gossip/validator.ts
```diff
@@ -1,279 +1,366 @@
...
+    {
+      type: GossipType.proposer_slashing,
+      objectValidatorFn: validateProposerSlashing as ObjectValidatorFn,
+    },
+    {
+      type: GossipType.attester_slashing,
+      objectValidatorFn: validateAttesterSlashing as ObjectValidatorFn,
+    },
+  ];
+  staticTopics.forEach(({type, objectValidatorFn}) => {
```
dapplion: `for .. of`

dapplion (dismissed):

dapplion (approved):

PR #2136, opened 2021-03-07
Simplify tsc setup

dapplion:
- Simplifies all tsconfig files throughout the repo with the minimum necessary settings
- Uniforms them to the same style and scripts
- Hoists common options to the root tsconfig

Also:
- Uses a faster clean script with file globs instead of using lerna (5s -> 0.5s)
- Required to fix various type errors, that were not visible due to the prev tsc setup


wemeetagain (dismissed):

package.json
```diff
@@ -8,9 +8,10 @@
     "packages/*"
   ],
   "scripts": {
-    "postinstall": "lerna run clean && lerna run build",
+    "postinstall": "yarn clean && lerna run build",
+    "clean": "rm -rf ./packages/*/lib ./packages/*/*.tsbuildinfo",
```
mpetrunic, line 12: should we use rimraf instead everywhere? It would be nice to support developing lodestar on windows?

dapplion: Can we tackle that in another PR? I'm sure there are more places with unix syntax

wemeetagain (approved):

PR #2142, opened 2021-03-07
Refactor peer managment

### LINKED ISSUE #823: Disconnect after goodbye is sent
Currently we don't disconnect once we receive goodbye rpc message as we would close connection before response is sent.

Solution:
"sendResponse" method should resolve promise once response is sent

### LINKED ISSUE #1989: Consider doing sync peer management ourselves
Sync peer management in Eth2.0 requires particular features such as 
- Ban peers and don't let them connect to us
- Don't connect to peers that have ban us
- Reverse both actions after some time (TTL/decay for bans/scoring)
- Limit connections per peer
- Handle autoDial with minPeer and maxPeers, not connections

Consider the dev cost of doing peer management ourselves instead of relying in libp2p connection manager

CC: @wemeetagain

dapplion:
Group all peer management functionality into a single class PeerManager. Current master segregates the functionality between tasks, network, and sync. PeerManager does:
- Perform handshake on peer connect, notify Sync only when peer is relevant. Ping and status peers every interval
- Mantain a set of peer by discovering + disconnecting according to conditions: targetPeers, minPeers, requested subnets. Groups logic to do so in testable pure functions

All ReqResp logic moved to network/reqResp so it can easily notify the PeerManager on incoming requests.

----

This PR is a spin-off from https://github.com/ChainSafe/lodestar/pull/2111 which is still unclear if it causes more memory and lag issues than master. I will A/B test this PR too to ensure it performs at least as well as master

---

#### `PeerManager`

PeerManager runs an internal `heartbeat` function which is a convenient start to follow its logic https://github.com/ChainSafe/lodestar/blob/03f356b35be94c51985dc1be87c1489da3d54790/packages/lodestar/src/network/peers/peerManager.ts#L119

Key sensitive logic:
- [`prioritizePeers`](https://github.com/ChainSafe/lodestar/blob/dapplion%2Frange-sync-rebased/packages/lodestar/src/network/peers/utils/prioritizePeers.ts): How to decide which peers to connect to and which to disconnect given an existing peer set and target peer count + target subnets

----

Fixes https://github.com/ChainSafe/lodestar/issues/823 fixes https://github.com/ChainSafe/lodestar/issues/1989


dapplion:
Rebased into a merge-able status. Ready for review.

packages/lodestar-cli/src/options/beaconNodeOptions/network.ts
```diff
@@ -6,7 +6,7 @@ export interface INetworkArgs {
   "network.discv5.bindAddr": string;
   "network.discv5.bootEnrs": string[];
   "network.maxPeers": number;
-  "network.minPeers": number;
+  "network.targetPeers": number;
```
mpetrunic, line 9: whats the difference between that and maxPeers? 
Idea behind minPeers was to not allow sync if peer count is below that

packages/lodestar/src/constants/network.ts
```diff
@@ -159,3 +162,9 @@ export const goodbyeReasonCodeDescriptions: Record<string, string> = {
   250: "Peer score too low",
   251: "Peer banned this node",
 };
+
+/** Until js-libp2p types its events */
```
mpetrunic, line 166: Aren't we maintaining types for that still in libp2p-ts?

packages/lodestar/src/network/events.ts
```diff
@@ -0,0 +1,25 @@
+import {EventEmitter} from "events";
+import PeerId from "peer-id";
+import StrictEventEmitter from "strict-event-emitter-types";
+import {phase0} from "@chainsafe/lodestar-types";
+import {Method} from "../constants";
+
+export enum NetworkEvent {
+  /** A relevant peer has connected or has been re-STATUS'd */
+  peerConnected = "PeerManager-peerConnected",
+  peerDisconnected = "PeerManager-peerDisconnected",
```
mpetrunic: can we make this enum values have consistent naming?

packages/lodestar-cli/src/options/beaconNodeOptions/network.ts
dapplion: You refer to `sync.minPeers`, which is different option with the purpose you describe. In current master `network.minPeers` is used by libp2p's autoDial logic. In the Range Sync PR https://github.com/ChainSafe/lodestar/pull/2111 `sync.minPeers` is eliminated, so we will then only have `network.minPeers` and `network.targetPeers`

mpetrunic: ok, and whats the difference between network.targetPeers and network.maxPeers? 😄

packages/lodestar/src/constants/network.ts
dapplion: libp2p-ts types these events but as a string uninon, not an enum. 
```ts
export type ConnectionEvents = 'peer:connect' | 'peer:disconnect';
```
https://github.com/ChainSafe/libp2p-ts/blob/86c1d9eca212194c8a5ce1ea6572a3bd694103e5/types/libp2p/index.d.ts#L95

mpetrunic: Why not change that there instead of here? 😄

packages/lodestar-cli/src/options/beaconNodeOptions/network.ts
dapplion: - `targetPeers`: Below that number actively discover and connect to peers. Above that number actively disconnect peers
- `maxPeers`: Hard cap to not allow any peer connections above that number.

You want `maxPeers` to be a little larger than `targetPeers` to allow new inbound peers that just joined the network connect to you (for the sake of the network's health).

packages/lodestar/src/constants/network.ts
dapplion: Just opened an issue https://github.com/ChainSafe/libp2p-ts/issues/21

wemeetagain (approved):

PR #2160, opened 2021-03-10
API Test Data Consolidation

### LINKED ISSUE #2080: Review API unit tests
Beacon node API unit tests
- feature a lot of code repetition, consider defining common utils.
- They are also quite slow, consider options to speed them up or move to e2e

3xtr4t3rr3str14l:
partially fixes #2080 by puling some of the common data between unit api tests into common objects.  The 2 big ones are in `unit/api/impl/index.test.ts' and `unit/api/rest/index.test.ts`, but there are also some similar actions taken within specific tests as well.


packages/lodestar/test/unit/api/rest/index.test.ts
```diff
@@ -0,0 +1,54 @@
...
+import {testLogger} from "../../../utils/logger";
+import {DebugBeaconApi} from "../../../../src/api/impl/debug/beacon";
+import {BeaconStateApi} from "../../../../src/api/impl/beacon/state";
+
+export const BEACON_PREFIX = "/eth/v1/beacon";
+export const CONFIG_PREFIX = "/eth/v1/config";
+export const NODE_PREFIX = "/eth/v1/node";
+export const VALIDATOR_PREFIX = "/eth/v1/validator";
+
+beforeEach(async function () {
```
dapplion: How does this architecture control to only run this beforeEach block on the tests that you want? By importing it selectively?

packages/lodestar/test/unit/api/impl/beacon/blocks/getBlock.test.ts
```diff
@@ -1,47 +1,20 @@
...
-        network: sinon.createStubInstance(Network),
-        sync: sinon.createStubInstance(BeaconSync),
-      }
-    );
-  });
-
-  afterEach(function () {
-    sandbox.restore();
+  beforeEach(() => {
+    resolveBlockIdStub = this.ctx.sandbox.stub(blockUtils, "resolveBlockId");
```
dapplion: Is `this.ctx` fully typed? If so how does that work?

packages/lodestar/test/unit/api/rest/index.test.ts
3xtr4t3rr3str14l: As it stands right now, it just attaches the objects to `this` in these outer tests which are then accessed inside of each individual test via `this.ctx` or `this.test.ctx`, depending on where you're accessing it.  But also, yeah I just realized that if you declare more than one of these index.test.ts with the beforeEach and afterEach in the global scope, then it calls both even if they are in different directories.  So, maybe it would be better to find a way to make sure that each index.test.ts is only ran for the tests in their corresponding directories (i.e. `rest/` and `impl/`)

packages/lodestar/test/unit/api/impl/beacon/blocks/getBlock.test.ts
3xtr4t3rr3str14l: Not really sure, honestly.  I just found this way to access the global test data, but didn't look too much under the hood of it yet.

3xtr4t3rr3str14l: I'll look into though.

dapplion:
Why don't you define a `setupApiTest()` function as a test util, and run that before each test?
```ts
// test/util/api.ts
export function setupApiTest() {
  return server
}
```
```ts
// test/api/a.test.ts
import {setupApiTest} from "test/util/api"
describe("", () => {
  it("", () => {
    const server = await setupApiTest()
    server.request(testPayload)
  })
})
```

dapplion:
What do you think about returning all modules as a plain object instead of a class? The `ApiImplTest` class holds no state and gets disassembled immediately, so it might make more sense
```ts
export type ApiImplTestModules = {
  sandbox: SinonSandbox;
  forkChoiceStub: SinonStubbedInstance<ForkChoice>;
  chainStub: SinonStubbedInstance<BeaconChain>;
  syncStub: SinonStubbedInstance<BeaconSync>;
  dbStub: StubbedBeaconDb;
  networkStub: SinonStubbedInstance<Network>;
  blockApi: BeaconBlockApi;
  config: IBeaconConfig;
};

export function setupApiImplTestServer(): ApiImplTestModules {
  const sandbox = sinon.createSandbox();
  const forkChoiceStub = sinon.createStubInstance(ForkChoice);
  const chainStub = sinon.createStubInstance(BeaconChain);
  const syncStub = sinon.createStubInstance(BeaconSync);
  const dbStub = new StubbedBeaconDb(sinon, config);
  const networkStub = sinon.createStubInstance(Network);
  const blockApi = new BeaconBlockApi(
    {},
    {
      chain: chainStub,
      config,
      db: dbStub,
      network: networkStub,
      sync: syncStub,
    }
  );
  chainStub.forkChoice = forkChoiceStub;
  return {
    sandbox,
    forkChoiceStub,
    chainStub,
    syncStub,
    dbStub,
    networkStub,
    blockApi,
    config,
  };
}
```

wemeetagain (dismissed):

wemeetagain (approved):

PR #2021, opened 2021-02-02
Process finalized checkpoints in order

twoeths:
resolves #2015 
+ use `JobQueue` to order the finalized checkpoint execution: one finalized checkpoint needs to be processed completely before the next one

PR #2161, opened 2021-03-11
beaconBlockByRange handler to work with BlockArchiver

### LINKED ISSUE #2127: blocks by range responds with orphaned blocks

twoeths:
resolves #2127 

Most of the time a `beaconBlocksByRange` handler does not need to deal with BlockArchiver, just in edge case when both run at the same time with an overlapped range 

+ Add new methods to BlockArchive task to know its status: `getArchivingStatus()` which returns a last finalized slot and a finalizing slot (`null` if not running); and `waitUntilComplete()` to wait for a `run()` to be done
+ if `beaconBlocksByRange` handler found that `BlockArchiver` is running with an overlapped range, it'd wait for BlockArchiver to finish first before querying database (a `BlockArchiver.run` takes less than 300ms so I hope it's ok)
+ if `step=1` do a linear chain segment check to make sure we return a good result to clients in order to avoid the downvote

### DIFF
--- a/packages/lodestar/src/network/reqresp/handlers/beaconBlocksByRange.ts
+++ b/packages/lodestar/src/network/reqresp/handlers/beaconBlocksByRange.ts
@@ -1,18 +1,25 @@
 import {GENESIS_SLOT} from "@chainsafe/lodestar-beacon-state-transition";
-import {MAX_REQUEST_BLOCKS, phase0} from "@chainsafe/lodestar-types";
+import {MAX_REQUEST_BLOCKS, phase0, Slot} from "@chainsafe/lodestar-types";
 import {IBlockFilterOptions} from "../../../db/api/beacon/repositories";
 import {IBeaconChain} from "../../../chain";
 import {IBeaconDb} from "../../../db";
 import {RpcResponseStatus} from "../../../constants";
 import {ResponseError} from "../response";
+import {IArchivingStatus, ITaskService} from "../../../tasks/interface";
+import {checkLinearChainSegment} from "../../../../src/util/chain";
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
 
-// TODO: Unit test
-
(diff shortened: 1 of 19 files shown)


dapplion:
Do you think it would make sense to add another job queue to schedule block ops? Then we could guarantee that archive, read, never happen at once.

With that approach `onBeaconBlocksByRange` won't have to know that `chores?: ITaskService` exists, modularizing the components a bit more.

Archive and read blocks should not be too long so if they are waiting for each other the overall time to satisfy a `BeaconBlocksByRange` shouldn't be affected in a significant manner.

twoeths:
> Do you think it would make sense to add another job queue to schedule block ops? Then we could guarantee that archive, read, never happen at once.
> 
> With that approach `onBeaconBlocksByRange` won't have to know that `chores?: ITaskService` exists, modularizing the components a bit more.
> 
> Archive and read blocks should not be too long so if they are waiting for each other the overall time to satisfy a `BeaconBlocksByRange` shouldn't be affected in a significant manner.

yeah I agree that's the good point, but we need to consider these as well:
+ the range of beacon_blocks_by_range request may not overlap to ArchiveBlocks and it still has to wait
+ multiple beacon_blocks_by_range come at the same time, do we want to put them to queue?
+ we also have an issue or ordering the finalized checkpoint process in #2021, which requires us to have a job queue there

wemeetagain:
Good job highlighting this corner-case issue. I'd like to hold off on this PR until we can find a more general way of handling the issue.

PR #2164, opened 2021-03-11
Review ReqResp stream handling

dapplion:
Follow up to https://github.com/ChainSafe/lodestar/pull/2163

After inspecting closely our ReqResp code, `js-libp2p`, `js-libp2p-mplex` and, `it-pushable` with the goals:

- Use abort interators and signals only if necessary
- Ensure to close the stream in all cases

this PR fixes some extra errors caused by a miss-understanding of the implications in upstream communication of the `for .. of` iterator protocol. It adds more clarifying comments to justify the changes and aid maintainability.

### DIFF
--- a/packages/lodestar/src/network/reqresp/request/index.ts
+++ b/packages/lodestar/src/network/reqresp/request/index.ts
@@ -1,5 +1,4 @@
 import {AbortSignal} from "abort-controller";
-import {source as abortSource} from "abortable-iterator";
 import pipe from "it-pipe";
 import PeerId from "peer-id";
 import {phase0} from "@chainsafe/lodestar-types";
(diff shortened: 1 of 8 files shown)


packages/lodestar/src/network/reqresp/request/index.ts
```diff
@@ -87,20 +86,17 @@ export async function sendRequest<T extends phase0.ResponseBody | phase0.Respons
...
-          stream.sink
-        ),
+      async () => await pipe(requestEncode(config, method, encoding, requestBody), stream.sink),
       REQUEST_TIMEOUT,
       signal
     ).catch((e) => {
-      // Must close the stream read side (stream.source) manually
-      stream.close();
+      // Must close the stream read side (stream.source) manually AND the write side
+      stream.reset();
```
wemeetagain: I think we want to `abort` here?

dapplion: I end up using reset because abort is not typed in your libp2p types and has the same exact effect with the difference that abort logs the error to the libp2p internal debugger.

wemeetagain: I think they're not exactly the same, see https://github.com/libp2p/js-libp2p-mplex/blob/master/src/stream.js#L101-L106
I think we should still use `abort` here.
For now, we can update our types to match (or ts-ignore :shrug: ). Slightly longer term, do this: #2167

packages/lodestar/src/network/reqresp/request/index.ts
```diff
@@ -111,21 +107,26 @@ export async function sendRequest<T extends phase0.ResponseBody | phase0.Respons
...
+      // Note: libp2p.stop() will close all connections, so not necessary to abort this pipe on parent stop
+      const responses = await pipe(
+        stream.source,
+        responseTimeoutsHandler(responseDecode(config, method, encoding), options),
+        collectResponses(method, maxResponses)
+      );
+
+      // NOTE: Only log once per request to verbose, intermediate steps to debug
+      // NOTE: Do not log the response, logs get extremely cluttered
+      // NOTE: add double space after "Req  " to match "Resp "
```
3xtr4t3rr3str14l: it took me a second to figure out what this meant.  this might be pedantic, but maybe a little more clearer would be something like
```javascript
// NOTE: add double space after "Req  " to align log with the "Resp " log
```

dapplion: Thanks! That's great feedback, comments should be easy to understand

wemeetagain (approved):

PR #2169, opened 2021-03-12
fix getHeaders/finalized call

### LINKED ISSUE #2157: Can't fetch finalized block via API
### Describe the bug

Fetching the finalized block using the API fails with 404

### Steps to Reproduce

`curl localhost:9596/eth/v1/beacon/headers/finalized`

3xtr4t3rr3str14l:
resolves #2157

### DIFF
--- a/packages/lodestar/src/api/impl/beacon/blocks/utils.ts
+++ b/packages/lodestar/src/api/impl/beacon/blocks/utils.ts
@@ -36,7 +36,7 @@ export async function resolveBlockId(
     return db.blockArchive.get(GENESIS_SLOT);
   }
   if (blockId === "finalized") {
-    return db.block.get(forkChoice.getFinalizedCheckpoint().root.valueOf() as Uint8Array);
+    return await db.blockArchive.get(forkChoice.getFinalizedCheckpoint().epoch);
   }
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/api/impl/beacon/blocks/utils.ts
```diff
@@ -36,7 +36,7 @@ export async function resolveBlockId(
     return db.blockArchive.get(GENESIS_SLOT);
   }
   if (blockId === "finalized") {
-    return db.block.get(forkChoice.getFinalizedCheckpoint().root.valueOf() as Uint8Array);
+    return await db.blockArchive.get(forkChoice.getFinalizedCheckpoint().epoch);
```
mpetrunic, line 39: don't we have last finalized block in memory?

wemeetagain: I don't think so, I don't think we have any blocks in memory for much time. Either pushed into the `block`, `pendingBlock` or `blockArchive` dbs.

wemeetagain:
looks like a unit test is failing

wemeetagain (approved):

PR #2171, opened 2021-03-13
Rename WinstonLogger to Logger

dapplion:
The fact that we use winston as a logger library is an implementation detail that other components should not be aware of. We only have 1 logger, with a generic interface ILogger, to it makes that the class is named Logger


mpetrunic (commented):
I'm conflicted about this. People using lodestar libraries might implement their own logger and they may also ask us to provide different logger implementation so I don't imagine we will always have single logger. 
If we consider switching to pino, I don't think we should just remove winston since there are always some integrations available in one logger but not other.

dapplion:
@mpetrunic I totally agree, closing PR

PR #2172, opened 2021-03-14
Add CachedBeaconState

wemeetagain:
Adds a `CachedBeaconState` abstraction and uses it throughout. (ie a refactored #1997)

The basic idea is that `IStateContext` is has become unwieldy as caches have been added and added.
Instead of dealing with `IStateContext`, with underlying proxy-backed `CachedValidatorsBeaconState` and `EpochContext`,
the (fast) state transition operates on a single object, the `CachedBeaconState`, which maintains all of the caches underneath.

In practice, this roughly looks like modifying `CachedValidatorsBeaconState` to include the `EpochContext`, plus a few other items.

Chunks of work worth calling out:

- Maintain the beacon state merkle tree directly, rather than as a proxy-wrapped `TreeBacked<BeaconState>`. This has twofold benefit.
  - 1. This removes a layer of proxy, which should minorly help performance.
  - 2. This forces us to maintain a separate reference to the beacon state ssz type. Exposing the type directly should make it easier to support updated beacon state ssz types, eg: the altair beacon state.
- Add a `MutableVector`, which is a reference to a `Vector` from `persistent-ts`. This is similar to how `persistent-merkle-tree` exposes a mutable `Tree`, which references an immutable root `Node`. This gives a more ergonomic setter interface, which doesn't return a _new vector_, but updates the reference in place.
- Add a `CachedValidatorsList`, which syncronizes the contents of a validator list merkle tree and a `MutableVector` of validators. (Again, the tree is operated on directly, and the ssz type is also explicitly maintained.)
  - This `CachedValidatorsList` is wrapped in a Proxy to provide the full `List` interface (from the ssz lib).
  - The only differences between a `CachedValidatorsList` and a `TreeBacked<List<Validator>>` is that the getter and iterator operate on the `MutableVector`. Because `get`ing a validator returns a flat validator, a normal object, validator entries must use the `update` method to synchronize updates to the tree (and `push` to add a validator).
- Simplify `EpochContext` initialization and methods
  - No more `EpochContext#loadState`, `EpochContext` must be initialized with real values, and not empty.
  - No more setters. `EpochContext` is treated as a read-only cache in practice. Mutator methods are pulled out into mutator functions.
- The `CachedBeaconState` is proxy-wrapped, with getter/setters which give it the interface `BeaconStateContext & EpochContext & ITreeBacked & BeaconState`
  - The proxy getter selectively returns a `CachedValidatorList` bound to the validator subtree (if the property is `"validator"`), an `EpochContext` property, a `BeaconStateContext` property, or a `BeaconState` property (as a `TreeBacked` value).
  - The proxy setter updates the state tree values
- Since `CachedBeaconState` supersedes `IStateContext`, many function interfaces in the state transition and lodestar, are simplified.
  - Several patterns emerge: `const {config, ...} = state;` to pull out the config and other properties. Many methods that reference "StateContext" can be entirely removed.
- `TreeBacked` has a recursive `TreeBackedify` component that causes incompatibilities between subtypes of TreeBacked and non-TreeBacked interfaces. It should likely be modified in the future (the ssz refactor already has this change). `RelaxedTreeBacked` is created and used in lodestar to temporarily get around this issue. (Ofc nothing comes for free, the tradeoff using a non-recursively TreeBacked type is that subtypes will need to be explicitly cast to `ITreeBacked` in order to access those methods)
- One difference from #1997 is that `CachedBeaconState` does not `extend EpochContext`. This turns out to be very convenient for stubbing the epochCtx in tests

TODO: rename `StateContextCache` to `StateCache` (separate PR?)

I've run this locally, syncing pyrmont and not seen any regressions

Some example code:
```ts

// before
function stateTransitionFoo(epochContext: EpochContext, state: CachedValidatorsBeaconState, ...args) {
  const {config} = epochContext;
  epochContext.getBeaconProposer(...);
  state.updateValidator(validatorIndex, {...});
  state.flatValidators().readOnlyForEach(...);
}

// after
function stateTransitionFoo(state: CachedBeaconState<phase0.BeaconState>, ...args) {
  const {config} = state;
  state.getBeaconProposer(...);
  state.validators.update(validatorIndex, {...});
  state.validators.forEach(...);
}

// before
function lodestarFoo(chain: IBeaconChain) {
  const stateCtx = chain.getHeadStateContext();
  stateCtx.epochCtx.getBeaconProposer(...);
  stateCtx.state.hashTreeRoot();
}

// after
function lodestarFoo(chain: IBeaconChain) {
  const state = chain.getHeadState();
  state.getBeaconProposer(...);
  state.hashTreeRoot();
}
```


dapplion (approved):
Awesome work! Really like the new ergonomics of CachedBeaconState

packages/lodestar-beacon-state-transition/src/phase0/fast/util/cachedBeaconState.ts
```diff
@@ -0,0 +1,105 @@
...
+    this.type = state.type() as ContainerType<T>;
+    this.tree = state.tree();
+    this.epochCtx = epochCtx;
+    this._validatorCache = validatorCache;
+  }
+
+  get validators(): CachedValidatorList<T["validators"][number]> & T["validators"] {
+    return (new Proxy(
+      new CachedValidatorList(
+        (this.type.tree.getProperty(this.tree, "validators") as unknown) as TreeBacked<List<T["validators"][number]>>,
```
mpetrunic, line 64: not for this Pr but we could maybe add optional generic to getProperty to make it easier to set expected type:
```
this.type.tree.getProperty<TreeBacked<List<T["validators"][number]>>>(this.tree, "validators")
```

mpetrunic (approved):

PR #2173, opened 2021-03-14
removed try / catch from abstractRepository

### LINKED ISSUE #2139: Dangerous try / catch all in lodestar-db
Lodestar-db has a try / catch that swallows all errors, without doing any triage first. Is this expected? Can we define what types of errors do we want to ignore and catch only those?

https://github.com/ChainSafe/lodestar/blob/a5973ea9551e3feae25a9c53d04d73616fe1d770/packages/lodestar-db/src/abstractRepository.ts#L54-L56

leviathanbeak:
Fixes #2139

### DIFF
--- a/packages/lodestar-db/src/abstractRepository.ts
+++ b/packages/lodestar-db/src/abstractRepository.ts
@@ -47,13 +47,9 @@ export abstract class Repository<I extends Id, T> {
   }
 
   async get(id: I): Promise<T | null> {
-    try {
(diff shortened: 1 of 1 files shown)


dapplion (approved):

mpetrunic (approved):

wemeetagain:
Thanks for the contribution @leviathan88 :)

wemeetagain (approved):

Issue #1733, opened 2020-11-04
Protection best practices


dapplion:
From https://github.com/ethereum/eth2.0-specs/pull/2082/files#diff-e3f50f31ab8c39b7d634a722cdbdf3575c0ae70c787da2007b4ab5ae24a6b3eaR613

> 3) Recovered validator -- Recovering a validator from a private key will result in an empty local slashing db. Best practice is to import (from a trusted source) that validator's attestation history. See [EIP 3076](https://github.com/ethereum/EIPs/pull/3076/files) for a standard slashing interchange format.
> 4) Far future signing requests -- A validator client can be requested to sign a far into the future attestation, resulting in a valid non-slashable request. If the validator client signs this message, it will result in it blocking itself from attesting any other attestation until the beacon-chain reaches that far into the future epoch. This will result in an inactivity leak and potential ejection due to low balance. 
A validator client should prevent itself from signing such requests by: a) keeping a local time clock if possible and following best practices to stop time server attacks and b) refusing to sign, by default, any message that has a large (>6h) gap from the current slashing protection database indicated a time "jump" or a long offline event. The administrator can manually override this protection to restart the validator after a genuine long offline event.

dapplion:
Is point (4) something we want to implement eventually?

dapplion:
@mpetrunic  What's your opinion on the value of implementing this protections in lodestar-validator?

mpetrunic:
nice to have, I would prefer if there is option to override this stuff from code as well.

dapplion:
Our current implementation follows this recommendations

PR #2178, opened 2021-03-16
Add CLI command: account wallet recovery

3xtr4t3rr3str14l:
adds new cli functionality to recover a wallet from mnemonic.  since there are similar processes used for wallet creation, i pulled some of the common functionality into separate, common functions as well.

based heavily on the lighthouse implementation:
https://github.com/sigp/lighthouse/blob/stable/account_manager/src/wallet/recover.rs

if everyone likes this, i'll use this as a basis to add validator recovery to the cli as well

### DIFF
--- a/packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
+++ b/packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
@@ -1,8 +1,8 @@
 import * as bip39 from "bip39";
-import {ICliCommand, ICliCommandOptions} from "../../../../util";
+import {ICliCommand, ICliCommandOptions, initBLS} from "../../../../util";
 import {IGlobalArgs} from "../../../../options";
 import {accountWalletsOptions, IAccountWalletArgs} from "./options";
-import {createWalletFromArgsAndMnemonic, printUuidData} from "./utils";
+import {createWalletFromArgsAndMnemonic} from "./utils";
 
 export const command = "create";
 
@@ -72,6 +72,8 @@ export const create: ICliCommand<IWalletCreateArgs, IAccountWalletArgs & IGlobal
   options: walletCreateOptions,
 
   handler: async (args) => {
+    await initBLS();
+
     // Create a new random mnemonic.
     const mnemonic = bip39.generateMnemonic();
 
@@ -95,8 +97,11 @@ export const create: ICliCommand<IWalletCreateArgs, IAccountWalletArgs & IGlobal
   recover your private keys in the case of data loss. Writing it on 
   a piece of paper and storing it in a safe place would be prudent.
 
-  `,
-      printUuidData(uuid)
+  Your wallet's UUID is:
+
+  \t${uuid}
+
+  You do not need to backup your UUID or keep it secret.`
     );
 
     // Return values for testing
--- a/packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
+++ b/packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
@@ -1,22 +1,55 @@
-import fs from "fs";
-import path from "path";
 import * as bip39 from "bip39";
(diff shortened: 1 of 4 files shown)


dapplion:
Related, see point 3 https://github.com/ChainSafe/lodestar/issues/1733#issuecomment-801882988

wemeetagain (dismissed):
LGTM

packages/lodestar-cli/src/cmds/account/cmds/wallet/create.ts
```diff
@@ -116,12 +95,9 @@ wallets are supported presently.",
...
   a piece of paper and storing it in a safe place would be prudent.
 
-  Your wallet's UUID is:
-
-  \t${uuid}
-
-  You do not need to backup your UUID or keep it secret.
-  `);
+  `,
+      printUuidData(uuid)
```
dapplion: This util is not necessary and may mess up the print style if tabulations are not computed properly. No problem in duplicating the output in two commands

packages/lodestar-cli/src/cmds/account/cmds/wallet/recover.ts
```diff
@@ -0,0 +1,59 @@
...
+  handler: async (args) => {
+    const {mnemonicInputPath} = args;
+    let mnemonic;
+
+    console.log("\nWARNING: KEY RECOVERY CAN LEAD TO DUPLICATING VALIDATORS KEYS, WHICH CAN LEAD TO SLASHING.\n");
+
+    if (mnemonicInputPath) {
+      mnemonic = await fs.promises.readFile(mnemonicInputPath, "utf8");
+    } else {
+      const input = await inquirer.prompt([
```
dapplion, line 45: Do other clients have an option to input the mnemonic manually? If it leaves a trace in bash history this is security risk and we should only offer reading from a file.

packages/lodestar-cli/src/cmds/account/cmds/wallet/utils.ts
```diff
@@ -0,0 +1,46 @@
+import fs from "fs";
+import path from "path";
+import {IGlobalArgs} from "../../../../options";
+import {YargsError, writeFile600Perm, randomPassword, readPassphraseFile, initBLS} from "../../../../util";
+import {WalletManager} from "../../../../wallet";
+import {getAccountPaths} from "../../paths";
+import {IWalletRecoverArgs} from "./recover";
+
+export async function createWalletFromArgsAndMnemonic(
```
dapplion, line 9: If `packages/lodestar-cli/src/cmds/account/cmds/wallet/recover.ts` is the only one using this logic it's not quite useful to split the logic in this utility. I would just dump this logic in recover, and if in the future it is used in multiple places, then we can abstract it away.

packages/lodestar-cli/src/cmds/account/cmds/wallet/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+  args: Pick<IWalletRecoverArgs & IGlobalArgs, "name" | "type" | "passphraseFile" | "mnemonicOutputPath" | "rootDir">,
+  mnemonic: string
+): Promise<{uuid: string; password: string}> {
+  await initBLS();
```
dapplion: To be consistent with other commands, could you place this only at the top of each command's handler function?

packages/lodestar-cli/src/cmds/account/cmds/wallet/recover.ts
3xtr4t3rr3str14l: lighthouse does it :shrug:

3xtr4t3rr3str14l: but yeah i can see the potential for security risk there.  we could potentially change the inquirer prompt to be "password" type instead of "input" like how we do for importing a validator

dapplion: If Lighthouse does it then I'm cool with it :+1:

dapplion (approved):

PR #2184, opened 2021-03-17
Run postStateTransition function in a separate context

### LINKED ISSUE #2179: BeaconStateContext instances are retained in Promise
**Describe the bug**

This is from a heap snapshot I used to sync Pyrmont

retained_BeaconStateContext

The promise being retained is the one in processChainSegment https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/chain/blocks/process.ts#L141

**Expected behavior**

The promise and BeaconStateContext instance are both not retained in memory

**Steps to Reproduce**
+ Sync Pyrmont
+ The more `await sleep(0);`  calls we add to `stateTransition` function, the more Promises are retained. 

- Commit hash: a855a08

twoeths:
resolves #2179 

+ Running the `ChainEvent.block` (and others) async handlers in the same Promise of `stateTransition` sometimes cause a lot of states being retained in memory
+ The fix is to use `setImmediate` to run `postStateTransition` in a separate context. `setImmediate` is called before the next event loop so we have the event processed before we finish `runStateTransition` (thanks to `await sleep(0)` at the end of `runStatetransition`). Reference: https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/#setimmediate-vs-settimeout

## Test with Pyrmont
+ master retains 615 MB of BeaconStateContext instances in memory (note: the dash `-` in `Distance` column means that the instance stays in memory forever)
BeaconStateContext_master

+ this branch retains 77MB of BeaconStateContext instances in memory - almost no dash `-` in `Distance` column
BeaconStateContext_fix_retained_state

### DIFF
--- a/packages/lodestar/src/chain/blocks/stateTransition.ts
+++ b/packages/lodestar/src/chain/blocks/stateTransition.ts
@@ -66,7 +66,7 @@ export async function processSlotsToNearestCheckpoint(
     nextEpochSlot += SLOTS_PER_EPOCH
   ) {
     phase0.fast.processSlots(postState, nextEpochSlot);
-    emitCheckpointEvent(emitter, postState.clone());
+    setImmediate(() => emitCheckpointEvent(emitter, postState.clone()));
     // this avoids keeping our node busy processing blocks
     await sleep(0);
   }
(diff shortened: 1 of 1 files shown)


dapplion:
This is an awesome step towards fixing our memory issues! :fire: However, I'm not sure if this is the best final solution but it somehow masks the underlying problem. Could we investigate a bit more to understand what is actually retaining memory?

twoeths:
> This is an awesome step towards fixing our memory issues! 🔥 However, I'm not sure if this is the best final solution but it somehow masks the underlying problem. Could we investigate a bit more to understand what is actually retaining memory?

Sorry I forgot to mention this https://github.com/ChainSafe/lodestar/issues/2005#issuecomment-770611519

it's this promise that keeps our state in memory https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/chain/blocks/process.ts#L141
retained_BeaconStateContext

wemeetagain:
Good progress!
I'm curious more about the root cause of:
> Running the ChainEvent.block (and others) async handlers in the same Promise of stateTransition sometimes cause a lot of states being retained in memory

Can you try to see if there are any similarities between the retained states? Maybe they are all checkpoint states, etc. It would be great to try to figure out _why_ the states are retained. Maybe there are some underlying errors, etc, that might be uncovered by digging deeper.
If they are all checkpoint states, we can take a deeper look at the checkpoint event handlers. If not, we can take a deeper look at block event handlers. Hopefully we will be able to narrow down some reason why certain promises are kept and others not.

wemeetagain:
Closing this as the issue has been resolved

PR #2185, opened 2021-03-17
Type error in catch clause

dapplion:
The `e` arg in `} catch (e) {` is typed as `any` effectively ts-ignoring all downstream code that consumes it. Typescript 4.0 allows to type the `e` as unknown https://github.com/microsoft/TypeScript/pull/39015 which is practically the correct type.

**Approach 1: no types at all**
```ts
} catch (e) {
  handleError(e)
}
```

**Approach 2: cast e to Error**
If `handleError` signature changes you will be alerted
```ts
} catch (e: unknown) {
  handleError(e as Error)
}
```

**Approach 3: check for Error**
Probably overkill since the likelyhood of an external library throwing a non-Error instance is extremely low
```ts
} catch (e: unknown) {
  if (e instanceof Error) {
    handleError(e)
  }
}
```

eslint rule [no-implicit-any-catch](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-implicit-any-catch.md) will enforce this pattern

This PR is meant to gauge interest from Lodestar contributors to adopt this type strategy with errors, before actually doing all the changes necessary. I've just bumped prettier to 2.1 to support [type annotations on catch clauses](https://prettier.io/blog/2020/08/24/2.1.0.html#type-annotations-on-catch-clauses-8805httpsgithubcomprettierprettierpull8805-by-fiskerhttpsgithubcomfisker) and run a find and replaces

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -86,6 +86,7 @@ module.exports = {
     "@typescript-eslint/type-annotation-spacing": "error",
     "@typescript-eslint/no-floating-promises": "error",
     "@typescript-eslint/explicit-member-accessibility": ["error", {accessibility: "no-public"}],
+    "@typescript-eslint/no-implicit-any-catch": "error",
     "import/no-extraneous-dependencies": [
       "error",
       {
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "mocha": "^8.3.0",
     "npm-run-all": "^4.1.5",
     "nyc": "^15.1.0",
-    "prettier": "^2.0.5",
+    "prettier": "^2.2.0",
     "sinon": "^9.0.2",
     "supertest": "^4.0.2",
     "ts-node": "^9.1.1",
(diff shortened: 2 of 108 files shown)


wemeetagain:
+1 for approach 2

wemeetagain:
I think we should move forward w approach 2

3xtr4t3rr3str14l:
i feel like it'd almost be easier to make a new PR and change all the errors rather than try to tackle those merge conflicts?  maybe if no one wants to resolve all those conflicts, we put this in an issue and close this?

dapplion:
> i feel like it'd almost be easier to make a new PR and change all the errors rather than try to tackle those merge conflicts? maybe if no one wants to resolve all those conflicts, we put this in an issue and close this?

@3xtr4t3rr3str14l  I just did a search and replace of `} catch (e) {` for `} catch (e: unknown) {`. You could reset hard this branch to current master and re-do the changes on top, then force push. This way we keep the PR and it's conversation

dapplion:
Closing for https://github.com/ChainSafe/lodestar/pull/3072

PR #2189, opened 2021-03-17
Customize log level for file transport

dapplion:
- Abstracts the `transport` argument with a generic non-Winston class description of the transports
- Adds `--logLevelFile <LogLevel>` CLI param to be able to customize the log level for file logging
- Uploads debug level logs in sim tests as test artifacts. This is extremely useful when the test break and you need more info, but won't pollute the regular output if all is good. Check the Sim Test GA run for a demo of the artifacts


.github/workflows/test-sim.yml
```diff
@@ -36,3 +37,10 @@ jobs:
       - name: Simulation multi thread multi node test
         run: yarn test:sim:multiThread
         working-directory: packages/lodestar
+
+      - name: Upload debug log test files
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: debug-test-logs
```
mpetrunic, line 45: I would include commit hash had some weird issues where other runs would overwrite files with same name

dapplion: That's so weird, in which repo did that happen? Is this a known public issue in Github Actions runners?

mpetrunic: private one, just checked it, it appears that issue was only when downloading artifact in another job.

mpetrunic (approved):

wemeetagain (approved):

PR #2192, opened 2021-03-18
logger.profile default level = verbose

### LINKED ISSUE #1436: Change default level for profiling logs
In winston logger, `logger.profile`, if level is not explicitly set, set to verbose.

**Edit:**
Log profiling logs to verbose, not info.

dapplion:
Fixes https://github.com/ChainSafe/lodestar/issues/1436

@tuyennhv Would this mess up your performance tests?

### DIFF
--- a/packages/lodestar-utils/src/logger/winston.ts
+++ b/packages/lodestar-utils/src/logger/winston.ts
@@ -2,7 +2,7 @@
  * @module logger
  */
 
-import {createLogger, Logger, transports as winstonTransports} from "winston";
+import {createLogger, LogEntry, Logger, transports as winstonTransports} from "winston";
 import {Context, defaultLogLevel, ILogger, ILoggerOptions, LogLevel} from "./interface";
 import chalk from "chalk";
 import {getFormat} from "./format";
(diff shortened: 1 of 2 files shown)


packages/lodestar-utils/src/logger/winston.ts
```diff
@@ -80,7 +80,7 @@ export class WinstonLogger implements ILogger {
   }
 
   profile(message: string, option?: {level: string; message: string}): void {
-    this.winston.profile(message, option);
+    this.winston.profile(message, {level: LogLevel.verbose, ...option} as LogEntry);
```
wemeetagain, line 83: I seem to remember that this winston didn't respect the log level here. Haven't tested this out, so not sure if thats still the case.

twoeths:
> Fixes #1436
> 
> @tuyennhv Would this mess up your performance tests?

we can either set default log level to `verbose` for the performance tests (`test:perf` in package.json) or update every `logger.profile` to use info level.

wemeetagain:
bump

dapplion:
Not necessary since we don't use profile in anywhere else than tests

PR #2193, opened 2021-03-18
Ban imports from lib

### LINKED ISSUE #1462: Consider not requiring from `modules/lib/path/to/file`
I believe requiring specific module paths is an anti-pattern since we are coupling the file structure of a module with its consumer. The internal structure of a module should not matter, everything that has to be used should be exported in the root.

Export everything that's used in the module main file.

dapplion:
Fixes https://github.com/ChainSafe/lodestar/issues/1462 for good. 

This rule will prevent adding more un-wanted imports from lib, which happens regularly due to VSCode's bad auto-import logic (see https://github.com/ChainSafe/lodestar/pull/2160/files#diff-be5db15737cd57cc0783725ea5bad27ab02ae94329d3b07372d698c3587087ebR11).

Current offenders of the rule are tests that need to stub certain imports from other modules, and common test utilities that have not been extracted to a common module.

This PR is meant to gauge interest by Lodestar maintainers. If you agree with enforcing this please throw in ideas of how to solve the issues above or we can just eslint ignore the rule in those files.


.eslintrc.js
```diff
@@ -111,6 +111,7 @@ module.exports = {
     "prefer-const": "error",
     quotes: ["error", "double"],
     semi: "off",
+    "no-restricted-imports": ["error", {patterns: ["../lib/*", "@chainsafe/*/lib/*"]}],
```
mpetrunic, line 119: I'm pretty sure we have a lot of places where we need "@chainsafe/*/lib/*" because not everything is exported at index.

dapplion: You can see the list in this PR's lint job, it's not too long now. With the root imports in the types and config packages most of the necessary usages are covered. Then it's mostly tests utils and stubs. We can add an
```ts
/* eslint-disable no-restricted-imports */
```
line there. The goal of this is to prevent unnecessary / un-intended usage of lib imports. Happy to hear other proposals

wemeetagain:
LGTM, just need to go all the way with it

wemeetagain (approved):

PR #2198, opened 2021-03-18
Log hygiene - info level

### LINKED ISSUE #2078: Improve lodestar-validator logging
Current lodestar-validator logging at the info level is very spammy. It should log some status message every slot at most like https://github.com/ChainSafe/lodestar/pull/2059 All other status logging should be moved to verbose or removed

dapplion:
- Beacon node: Move gossip subscription log to verbose so the only periodic info log is the periodic notifier https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/node/notifier.ts
- Validator: Log only once per successful published block, aggregate and attestation. Move other logs to verbose
- Genesis: Log once every 30s instead of spam

Fixes https://github.com/ChainSafe/lodestar/issues/2078

### DIFF
--- a/packages/lodestar-validator/src/services/attestation.ts
+++ b/packages/lodestar-validator/src/services/attestation.ts
@@ -167,7 +167,7 @@ export class AttestationService {
     // TODO: is this how we should handle a non-matching validator?
     if (!validator) return;
 
-    this.logger.info("Handling attestation duty", {
+    this.logger.debug("Handling attestation duty", {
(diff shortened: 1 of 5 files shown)


wemeetagain:
:tada: i like the idea: I've had this general idea for the log levels:
- info - high-level summary view of the node, fatal errors requiring user intervention
- verbose - internal view of the node without the details
- debug - implementation details

wemeetagain (approved):

PR #2201, opened 2021-03-18
Remove version restrictions

### LINKED ISSUE #1370: Node v12- v14 support

dapplion:
Lodestar seems compatible with newer version of NodeJS. Versions lower to v12 are currently blocked by libp2p requirements. Should we keep the restriction for lower versions anyway?

Fixes https://github.com/ChainSafe/lodestar/issues/1370

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12.13-alpine as build
+FROM node:14-alpine as build
 WORKDIR /usr/app
 RUN apk update && apk add --no-cache git g++ make python && rm -rf /var/cache/apk/*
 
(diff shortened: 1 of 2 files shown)


wemeetagain:
I think its probably a good idea keeping the lower version restriction, not a big deal though.

wemeetagain (approved):

PR #2204, opened 2021-03-18
Allow to regen states in API debug states

### LINKED ISSUE #2177: Serve Weak Subjectivity State

We would like to upload certain historical states, weak subjectivity states, to the cloud, for use in bootstrapping sync.

We need to expand our debug/states API endpoint to be able to serve historical checkpoint states at weak subjectivity epochs.
Currently, the endpoint just checks our state cache, 404s if the requested state is not found.

**Additional context**

See https://notes.ethereum.org/@adiasg/weak-subjectvity-eth2#Weak-Subjectivity-Checkpoint

dapplion:
**Motivation**

We need to serve old states for Weak Subjectivity see https://github.com/ChainSafe/lodestar/issues/2177

**Description**

Only for the debug states endpoint, uses the chain.regen module to fetch states.
Closes https://github.com/ChainSafe/lodestar/issues/2177

I have not tested any code path

### DIFF
--- a/packages/lodestar/src/api/impl/beacon/state/utils.ts
+++ b/packages/lodestar/src/api/impl/beacon/state/utils.ts
@@ -2,31 +2,35 @@
 import {GENESIS_SLOT, FAR_FUTURE_EPOCH, CachedBeaconState} from "@chainsafe/lodestar-beacon-state-transition";
 import {phase0} from "@chainsafe/lodestar-beacon-state-transition";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
-import {IForkChoice} from "@chainsafe/lodestar-fork-choice";
-import {Epoch, ValidatorIndex, Gwei, Slot} from "@chainsafe/lodestar-types";
+import {Epoch, ValidatorIndex, Gwei, Slot, Root} from "@chainsafe/lodestar-types";
 import {fromHexString, readOnlyMap} from "@chainsafe/ssz";
 import {IBeaconChain} from "../../../../chain";
-import {StateContextCache} from "../../../../chain/stateCache";
 import {IBeaconDb} from "../../../../db/api";
 import {StateId} from "./interface";
+
+export enum RegenType {
+  CacheOnly,
+  AllowRegen,
+}
 
 export async function resolveStateId(
   chain: IBeaconChain,
   db: IBeaconDb,
-  stateId: StateId
+  stateId: StateId,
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -2,31 +2,35 @@
...
-import {IForkChoice} from "@chainsafe/lodestar-fork-choice";
-import {Epoch, ValidatorIndex, Gwei, Slot} from "@chainsafe/lodestar-types";
+import {Epoch, ValidatorIndex, Gwei, Slot, Root} from "@chainsafe/lodestar-types";
 import {fromHexString, readOnlyMap} from "@chainsafe/ssz";
 import {IBeaconChain} from "../../../../chain";
-import {StateContextCache} from "../../../../chain/stateCache";
 import {IBeaconDb} from "../../../../db/api";
 import {StateId} from "./interface";
 
+export enum RegenType {
```
dapplion, line 11: Suggestion for a better name? :pray:

packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -2,31 +2,35 @@
...
+  CacheOnly,
+  AllowRegen,
+}
+
 export async function resolveStateId(
   chain: IBeaconChain,
   db: IBeaconDb,
-  stateId: StateId
+  stateId: StateId,
+  type: RegenType = RegenType.CacheOnly
```
dapplion, line 20: Only the debug state endpoint is able to trigger regen. This is a serious DOS vector so I think it makes that is only allowed for the debug API which is not enabled in normal usage.

twoeths:
I think `chain.regen` is supposed to regenerate state from a last finalized state in the checkpoint state cache while what we want is to regenerate an old finalized state that was not stored in db. The work may involve (if we're not able to regenerate state from `chain.regen`)
+ start from a finalized state in db
+ get all blocks from that state to a slot we want
+ do the regen by keep running `phase0.fast.fastStateTransition()` without forkchoice/cache being involved

wemeetagain:
Yeah, this PR by itself shouldn't close #2177 for the reasons @tuyennhv mentioned. Imo this is still a useful step in the right direction.

dapplion:
Replaced by https://github.com/ChainSafe/lodestar/pull/2369

PR #2207, opened 2021-03-19
docker refac plus primary dev setup

g11tech:
**Motivation**

To make the code base docker dev and deploy friendly

**Description**
Dockerfile is refactored to allow for development image, respective docker-compose.yml's are changed to map to the new Dockerfile, development.yml is added for someone to spin up a development container, with code base mappings from outside container to inside. Modified the docker section in the contributions.md.


Dockerfile
```diff
@@ -1,7 +1,12 @@
-FROM node:14-alpine as build
+FROM node:14-alpine as base
 WORKDIR /usr/app
 RUN apk update && apk add --no-cache git g++ make python && rm -rf /var/cache/apk/*
 
+FROM base as development
+RUN apk add --no-cache bash
+RUN yarn global add ts-node typescript
```
dapplion, line 7: Tried locally and this code ran always even when targeting production. Could you find a different arrangement so the production build doesn't have to do extra work exclusive to development?

development.yml
```diff
@@ -0,0 +1,10 @@
+version: "3.4"
+services:
+  beacon_node:
+    build:
+      target: development
+    image: lodestar:dev
+    volumes:
+      - ./:/usr/app
+      - ../dev/lodestar/node_modules:/usr/app/node_modules
```
dapplion: This `dev` folder is specific to your local environment directory structure? Could you make it more generic?

development.yml
```diff
@@ -0,0 +1,10 @@
...
+    command: tail -F anything
```
dapplion: Why this command?

dapplion:
Hi @g11in ! Thank you so much for contributing :heart: Happy to explore different ways to develop, tho could you move non-production items to either `/scripts` or `/docker`? To not pollute the root dir more than necessary

development.yml
g11tech: this is to keep the container spinning, so as to provide shells for the developer to run installs, beacon and/or val or dev/test individual segments, this overrides the main docker-compose command so the image doesn't start executing default code.

development.yml
g11in: umm, yes but it doesn't has any other significance other than separating out installs from the code base inside the container so as to not dirty the codebase in the IDE as well as save reinstall process if development image is rebuild. do suggest how to make it generic. with environment variable (from .env) may be?

Dockerfile
g11tech: i will check it out and see how to optimize.

development.yml
g11in: i will also be abstracting out lib files in the docker dev image (same reason as to not corrupt the IDE) plus docker cache use on the rebuilds (as the build artifacts will not corrupt main codebase)

g11tech:
> 
> 
> Hi @g11in ! Thank you so much for contributing ❤️ Happy to explore different ways to develop, tho could you move non-production items to either `/scripts` or `/docker`? To not pollute the root dir more than necessary

sure, will do the  rearrangement :)

development.yml
g11tech: i will hookup docker persistent volume here as the only relevance is to prevent build artifacts in codebase. will also link up lib dirs for the same reason.

Dockerfile
g11tech: hey @dapplion , i did a rebuild with no cache for just production, it actually didn't build the development target. image attached.
image

Dockerfile
```diff
@@ -1,7 +1,12 @@
-FROM node:14-alpine as build
+FROM node:14-alpine as base
 WORKDIR /usr/app
 RUN apk update && apk add --no-cache git g++ make python && rm -rf /var/cache/apk/*
 
+FROM base as development
+RUN apk add --no-cache bash
```
mpetrunic, line 6: why not add this to L3?

g11in: whats L3?

mpetrunic: Line 3 of this file :)

g11tech: makes sense, could shift there as final "production" gets the copies from "build" target, but only catch is a little bit extra time it might take to add bash package for production build. so in my opinion this is just fine unless you strongly feel otherwise.

mpetrunic: From what I see, production image is fresh nodejs-alpine image so it won't have bash installed. I'm asking because this new apk add probably adds new docker layer which probably makes image size larger. Do you have docker image sizes before and after?

g11tech: yes, final production image gets copy from build, i think it shouldn't affect as in my production target build the development target was not even executed, so i think it shouldnt be adding any extra layers. however will confim on image sizes.
image

g11tech: @mpetrunic yep negligible diff ~1MB, no development target layers included

Dockerfile
g11tech: @dapplion if you are using docker build ., then it would be building all targets, if you are using docker build . --target production then it won't. just trying to guess why you saw all the targets building :)

development.yml
g11tech: generalized using docker volumes, also moved development.yml to docker folder.

docker/development.yml
```diff
@@ -0,0 +1,109 @@
...
+      - devinstall_ls_utils_lib:/usr/app/packages/lodestar-utils/lib
+
+      - devinstall_ls_val_node_modules:/usr/app/packages/lodestar-validator/node_modules
+      - devinstall_ls_val_lib:/usr/app/packages/lodestar-validator/lib
+
+      - devinstall_persist_node_modules:/usr/app/packages/persistent-ts/node_modules
+      - devinstall_persist_lib:/usr/app/packages/persistent-ts/lib
+
+      - devinstall_str_node_modules:/usr/app/packages/spec-test-runner/node_modules
+      - devinstall_str_lib:/usr/app/packages/spec-test-runner/lib
```
dapplion, line 104: What's the purpose of all this volumes?

g11tech: so as to not let build artifacts mix in with the codebase outside the container, each volume to mount for node_modules/lib for reach of packages at their respective places

dapplion: That's a very verbose strategy that requires editing when packages change. Is there a more succinct and decoupled approach?

g11tech: ideally the project should build files in one place (which could then be mounted to single volume), but unfortunately its not and in my opinion should go restructuring which I believe some it you guys are doing. 
Currently, one that that can be done, is a script can build sim-links at all these places to only one particular location mounted by a single volume and that script can be made to run in development build target of dockerfile. if you like that approach, can implement that one.

g11tech:
@dapplion there is another thing that i feel should be changed, depends_on field could be removed, for the reason that if for e.g. beacon_node restarts or shutdowns, it will force other containers depending on it to shutdown/restart, which shouldn't be the case,  validator/grafana/prom.. can/should handle the timeout/not reachable response.

wemeetagain:
closing, since we discussed this in our meeting today

PR #2211, opened 2021-03-20
Cleaner validator info logs

dapplion:
Instead of logging full hashes of what's being signed, log only the slot and who is signing it with prettyBytes. Lighthouse does not log hashes on the validator client.

Should we log more info about what's being signed at the debug or verbose level?

### DIFF
--- a/packages/lodestar-utils/src/format.ts
+++ b/packages/lodestar-utils/src/format.ts
@@ -1,0 +1,9 @@
+import {ByteVector, toHexString} from "@chainsafe/ssz";
+
+/**
+ * Format bytes as `0x1234…1234`
+ */
+export function prettyBytes(root: Uint8Array | ByteVector): string {
+  const str = toHexString(root);
+  return `${str.slice(0, 6)}…${str.slice(-4)}`;
+}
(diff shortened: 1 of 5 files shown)


wemeetagain (approved):
👍 to moving prettyBytes to lodestar-utils

I think its still useful to have a more verbose log of what the validator is signing. Ofc even more useful would be metrics.

dapplion:
>  to moving prettyBytes to lodestar-utils
> 
> I think its still useful to have a more verbose log of what the validator is signing. Ofc even more useful would be metrics.

What would you log about block, aggregate and attestation? i.e. just the root of the entire object, certain metadata, etc. Check Lighthouse's client as example https://github.com/sigp/lighthouse/blob/b34a79dc0b02e04441ba01fd0f304d1e203d877d/validator_client/src/attestation_service.rs#L411

mpetrunic:
> > to moving prettyBytes to lodestar-utils
> > I think its still useful to have a more verbose log of what the validator is signing. Ofc even more useful would be metrics.
> 
> What would you log about block, aggregate and attestation? i.e. just the root of the entire object, certain metadata, etc. Check Lighthouse's client as example https://github.com/sigp/lighthouse/blob/b34a79dc0b02e04441ba01fd0f304d1e203d877d/validator_client/src/attestation_service.rs#L411

I would say as much as possible, root of objects definetly and some metadata like slot, validator. I like info unaggregated/aggregated attestation.

I wonder why they excluded hash root. Seems pretty useful to find you block in explorer.

PR #2212, opened 2021-03-20
Increase sim test stability

dapplion:
**Motivation**

Multi-thread sim tests are very unstable due to wrong assumptions in our sync module. It is not able to handle nicely the genesis situation that occurs in precisely that sim test, causing nodes to fork and not reach justification together.

**Description**

This PR introduces two main changes:
- Adds a somewhat "dirty" hack to mark the sync as completed and start gossip if epoch = 0, and node has peers. This should be properly fixed eventually with https://github.com/ChainSafe/lodestar/pull/2111. Do you see any dangers with this condition? What other condition or event could we listen for to find genesis more reliably?
- Disable MDNS for all the sim tests and manually connect the nodes with each other. In the multi-thread test, p2p ports and peerIds are pre-computed and sent as worker data to all threads. Then nodes wait a bit and connect to each other. If the timing is correct nodes should be ready to receive inbound connections by the time they try to connect.

Minor issues:
- Added process.exit(1) if worker start sequence fails. Currently it would just throw and unhandled rejection and do nothing.
- Added a condition in network to skip trying to connect to subnet peers if discv5 is not available
https://github.com/ChainSafe/lodestar/blob/76da10ac8a485b3cc8f1f444a6ed0f873dd90b88/packages/lodestar/src/network/network.ts#L244-L245

This PR is a pre-requisite to merge https://github.com/ChainSafe/lodestar/pull/2142


wemeetagain (approved):
LGTM

The "hack" looks good

PR #2213, opened 2021-03-21
num to bytes and bytes to num for safe ints, defaults to bigint for u…

### LINKED ISSUE #199: Improve intToBytes, bytesToInt for number
Right now these utils convert to BigInt unnecessarily and we know that we have performance issue with using BigInt

Reimplement these functions without going through `bigint-buffer`

### LINKED ISSUE #2121: Improve intToBytes, bytesToInt for number
Right now these utils convert to BigInt unnecessarily and we know that we have performance issue with using BigInt

Reimplement these functions without going through `bigint-buffer`

g11tech:
…nsafe

**Motivation**

to provide native int to bytes and reverse implementation wherever possible instead of using bigint

**Description**
used javascript dataview for a number to extract bytes whenever number <= MAX_SAFE_INTEGER

Closes #2121

### DIFF
--- a/packages/lodestar-utils/src/buf64copy.ts
+++ b/packages/lodestar-utils/src/buf64copy.ts
@@ -1,0 +1,178 @@
+export function copyFromBuf64LE(sbuffer: Buffer, tbuffer: Buffer, length: number): void {
+  //unrolled fast copy
+  switch (length) {
+    case 8:
+      tbuffer[7] = sbuffer[7];
+    /* falls through */
+    case 7:
+      tbuffer[6] = sbuffer[6];
+    /* falls through */
+    case 6:
+      tbuffer[5] = sbuffer[5];
+    /* falls through */
+    case 5:
+      tbuffer[4] = sbuffer[4];
+    /* falls through */
+    case 4:
+      tbuffer[3] = sbuffer[3];
+    /* falls through */
+    case 3:
+      tbuffer[2] = sbuffer[2];
+    /* falls through */
+    case 2:
+      tbuffer[1] = sbuffer[1];
+    /* falls through */
+    case 1:
+      tbuffer[0] = sbuffer[0];
+      break;
+    default:
+  }
+}
+
+export function copyFromBuf64BE(sbuffer: Buffer, tbuffer: Buffer, length: number): void {
+  switch (
+    length //we need to unroll it for fast computation, buffer copy/looping takes too much time
+  ) {
+    case 8:
+      tbuffer[0] = sbuffer[0]; //length is 8 here
+    /* falls through */
+    case 7:
+      tbuffer[length - 7] = sbuffer[1];
+    /* falls through */
+    case 6:
+      tbuffer[length - 6] = sbuffer[2];
+    /* falls through */
+    case 5:
+      tbuffer[length - 5] = sbuffer[3];
+    /* falls through */
+    case 4:
+      tbuffer[length - 4] = sbuffer[4];
+    /* falls through */
+    case 3:
+      tbuffer[length - 3] = sbuffer[5];
+    /* falls through */
+    case 2:
+      tbuffer[length - 2] = sbuffer[6];
+    /* falls through */
+    case 1:
+      tbuffer[length - 1] = sbuffer[7];
+      break;
+    default:
+  }
+}
+
+export function copyToBuf64LE(sbuffer: Uint8Array, tbuffer: Buffer, length: number): void {
+  switch (
+    length //we need to unroll it for fast computation, buffer copy/looping takes too much time
+  ) {
+    case 8:
+      tbuffer[7] = sbuffer[7];
+    /* falls through */
+    case 7:
+      tbuffer[6] = sbuffer[6];
+    /* falls through */
+    case 6:
+      tbuffer[5] = sbuffer[5];
+    /* falls through */
+    case 5:
+      tbuffer[4] = sbuffer[4];
+    /* falls through */
+    case 4:
+      tbuffer[3] = sbuffer[3];
+    /* falls through */
+    case 3:
+      tbuffer[2] = sbuffer[2];
+    /* falls through */
+    case 2:
+      tbuffer[1] = sbuffer[1];
+    /* falls through */
+    case 1:
+      tbuffer[0] = sbuffer[0];
+      break;
+    default:
+  }
+  switch (
+    length //zero the rest
+  ) {
+    case 1:
+      tbuffer[1] = 0;
+    /* falls through */
+    case 2:
+      tbuffer[2] = 0;
+    /* falls through */
+    case 3:
+      tbuffer[3] = 0;
+    /* falls through */
+    case 4:
+      tbuffer[4] = 0;
+    /* falls through */
+    case 5:
+      tbuffer[5] = 0;
+    /* falls through */
+    case 6:
+      tbuffer[6] = 0;
+    /* falls through */
+    case 7:
+      tbuffer[7] = 0;
+      break;
+    default:
+  }
+}
+
+export function copyToBuf64BE(sbuffer: Uint8Array, tbuffer: Buffer, length: number): void {
+  switch (
+    length //we need to unroll it for fast computation, buffer copy/looping takes too much time
+  ) {
+    case 8:
+      tbuffer[0] = sbuffer[0];
+    /* falls through */
+    case 7:
+      tbuffer[1] = sbuffer[length - 7];
+    /* falls through */
+    case 6:
+      tbuffer[2] = sbuffer[length - 6];
+    /* falls through */
+    case 5:
+      tbuffer[3] = sbuffer[length - 5];
+    /* falls through */
+    case 4:
+      tbuffer[4] = sbuffer[length - 4];
+    /* falls through */
+    case 3:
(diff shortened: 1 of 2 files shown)


dapplion:
Hi @g11in ! Thanks again for contributing. Make sure CI passes, and follow the repo's prettier style.

Also, could you add some detailed benchmarks for the before and after of the functions you have optimized?

g11tech:
> 
> 
> Hi @g11in ! Thanks again for contributing. Make sure CI passes, and follow the repo's prettier style.
> 
> Also, could you add some detailed benchmarks for the before and after of the functions you have optimized?
with little bit of more refac (will regen the PR) i am seeing average gains of ~30% (for 64bits).
image
script:
image

g11tech:
@dapplion  updated stats:

image

dapplion:
> @dapplion updated stats:
> 
> image

Thanks! Could you detail what data and code did you use to benchmark? I would also like to see the ops/sec before and after with data inputs of 8, 32, 64 bytes, which is what we use in practice.

Overall while there is a speed up, if believe the complexity introduced is way to high. Probably not worth the trade-off. One of the goals of Lodestar is to be easy to understand, easy to hack, and educational. We have to mantain a delicate balance between performance and clarity.

g11tech:
> Thanks! Could you detail what data and code did you use to benchmark? I would also like to see the ops/sec before and after with data inputs of 8, 32, 64 bytes, which is what we use in practice.
> 
> Overall while there is a speed up, if believe the complexity introduced is way to high. Probably not worth the trade-off. One of the goals of Lodestar is to be easy to understand, easy to hack, and educational. We have to mantain a delicate balance between performance and clarity.

a javascript number would never be bigger than 8 bytes. This implementation delegates to bigint if a number is bigger than Number.MAX_SAFE_INTEGER as number manipulation becomes non precise beyond this point.

benchmarking code snip: https://github.com/g11in/utilities/blob/main/fastnumbytesconv.js . it generates random numbers between 1 to MAX_SAFE_INTEGER (above which the code anyway delegates to bigint), does a byte xtraction of length between 1-8 bytes in round robin fashion and convert it back. checks any mismatches between this and bigint, and then time them.

it is a little bit low level implementation as it requires one to understand computer architecture and how JavaScript treat numbers. however someone snooping around to play with the implementation is/would be expected to understand few concepts. thankfully it will all be hidden under inttobytes or bytestoint in the end.

if you think its worth it, then i can generate the benchmarks for 8 bytes, **(for 8 bytes, the speedup is 30-40%)**

packages/lodestar-utils/src/bytes.ts
```diff
@@ -1,19 +1,96 @@
...
+  if (length > 8 || value > Number.MAX_SAFE_INTEGER || value < -Number.MAX_SAFE_INTEGER)
+    return intToBytes(BigInt(value), length, endianness);
+  //can't handle if above conds are true
+  //length less than required will lead to truncation on the msb side, same as bigint inttobytes
+
+  if (value < 0) value = -value; //bigint inttobytes always return the absolute value
+
+  let mvalue;
+  const view = viewG;
+  const result = Buffer.allocUnsafe(length);
```
wemeetagain, line 38: We need a solution that is not nodejs specific

packages/lodestar-utils/src/bytes.ts
```diff
@@ -1,19 +1,96 @@
...
  * Return a byte array from a number or BigInt
  */
 export function intToBytes(value: bigint | number, length: number, endianness: Endianness = "le"): Buffer {
-  return bigIntToBytes(BigInt(value), length, endianness);
+  switch (typeof value) {
+    case "number":
+      return numberToBytes(value, length, endianness);
+    case "bigint":
+      return bigIntToBytes(value, length, endianness);
+      break;
```
wemeetagain, line 15: unnecessary break

packages/lodestar-utils/src/bytes.ts
```diff
@@ -1,19 +1,96 @@
...
+    default:
+      throw new Error(`unsupported number type ${typeof value}`);
+  }
 }
 
 /**
  * Convert byte array in LE to integer.
  */
```
wemeetagain, line 23: jsdoc needs to remain right above the function its documenting

packages/lodestar-utils/src/bytes.ts
```diff
@@ -1,19 +1,96 @@
...
+const bufferG = new ArrayBuffer(8);
+const viewG = new DataView(bufferG);
+
+export function numberToBytes(value: number, length: number, endianness: Endianness = "le"): Buffer {
+  if (length > 8 || value > Number.MAX_SAFE_INTEGER || value < -Number.MAX_SAFE_INTEGER)
+    return intToBytes(BigInt(value), length, endianness);
+  //can't handle if above conds are true
+  //length less than required will lead to truncation on the msb side, same as bigint inttobytes
+
+  if (value < 0) value = -value; //bigint inttobytes always return the absolute value
```
wemeetagain, line 34: this should be rethought. perhaps our top-level functions should operate on "uint"s instead of ints, this would match the spec closer, too.

packages/lodestar-utils/src/bytes.ts
```diff
@@ -1,19 +1,96 @@
...
 /**
  * Convert byte array in LE to integer.
  */
+
+const bufferG = new ArrayBuffer(8);
+const viewG = new DataView(bufferG);
+
+export function numberToBytes(value: number, length: number, endianness: Endianness = "le"): Buffer {
+  if (length > 8 || value > Number.MAX_SAFE_INTEGER || value < -Number.MAX_SAFE_INTEGER)
+    return intToBytes(BigInt(value), length, endianness);
```
wemeetagain, line 30: should call `bigIntToBytes`

packages/lodestar-utils/src/buf64copy.ts
```diff
@@ -0,0 +1,178 @@
+export function copyFromBuf64LE(sbuffer: Buffer, tbuffer: Buffer, length: number): void {
```
wemeetagain, line 1: might be better named `source` and `target`. We know they're a buffer from their type :slightly_smiling_face:

wemeetagain (commented):
I appreciate the inventiveness here, but I'd echo @dapplion 's sentiment.

There is always a tradeoff between maintainability/readability and highly tuned implementations. We can navigate this tradeoff by profiling and analysis of algorithms used throughout the project/product. Blind optimization of _everything_ results in unmaintainable code for little payoff. Generally speaking, the low hanging fruit of optimization, for us, still exist at the algorithmic level, and we should approach micro-optimizations carefully, save this for known hotspots, and thoroughly document them as they tend to be non-trivial, or at least not readily accessible.

In this case, we should look at how/when/under what conditions `intToBytes` and `bytesToInt` are used. Just for background, these functions were originally meant to model [`uint_to_bytes`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/beacon-chain.md#uint_to_bytes) and [`bytes_to_uint64`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/beacon-chain.md#bytes_to_uint64) from the spec.

Taking a look at when/where we use these functions in our codebase, there are two distinct uses that stand out:
- in the spec, to seed a hash
- serializing/deserializing db keys

While I haven't done extensive benchmarking, my intuition is that in both cases, the routines are actually dominated by other operations. (In the first case, by hashing, in the second, by reading/writing from disk) In fact, in the profiling I've done, I've never seen either function appear.

With this tempered perspective, this code addition takes on a different character. We view our code not just as a means so solve a problem, but also as a medium to embed meaning to developers (ourselves and others, now and in the future).
The immediate questions come down to style:
- It's not immediately clear what the algorithm performed here is. I spent over 10 minutes combing through the code trying to figure out what is actually happening, between the global buffer, multiple copying functions, etc. We would need to document what is happening here and why we need it
- It appears that there's a lot of duplication here, is it all needed? can it be eliminated?
- Why do we need an unrolled loop to copy a buffer? Does `TypedArray#set` not suffice?
- There are several inconsistencies which make it harder to figure out
  - `numberToBytes` is broken into its own function, while `bytesToNumber` isn't
  - In one endian switch/case, there's a default, in another, there isn't
  - some variables are `camelCase`, some are `notcamelcase`
  - some variable names are prepended with a letter, some are appended with a letter

g11tech:
> 
> 
> I appreciate the inventiveness here, but I'd echo @dapplion 's sentiment.


I wouldn't disagree (with you @wemeetagain and @dapplion ) that in scheme of things, there would be other operations that would actually be dominating the computation. 
I could work on the other issues/documentation for this, if you this it would be worth it to merge this, complexity won't come down (some duplication might but duplication is because of unrolling, i used buffer copy methods before unrolling, and they were too costly to offset any gain made) because in the end you got to look at the javascript implementation and hack it there if any more juice needs to be extracted. else bigint has sufficient performance in bigger scheme of things. :)

dapplion:
During prater genesis this are the total time on each of the two target functions of this PR

Screenshot from 2021-03-25 08-39-43
Screenshot from 2021-03-25 08-39-07

g11tech:
> 
> 
> During prater genesis this are the total time on each of the two target functions of this PR

yea  not worth it.

wemeetagain:
Closing this PR. Thanks for working thru it with us.

PR #2186, opened 2021-03-17
Enable no-unsafe eslint rules

dapplion:
Eslint v7 has 4 new rules very useful to guard against using `any` types inadvertently.

- Disallows assigning any to variables and properties ([no-unsafe-assignment](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unsafe-assignment.md))
- Disallows calling an any type value ([no-unsafe-call](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unsafe-call.md))
- Disallows member access on any typed variables ([no-unsafe-member-access](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unsafe-member-access.md))
- Disallows returning any from a function ([no-unsafe-return](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unsafe-return.md))

The purpose of this PR is gauge interest by Lodestar maintainers before actually working on the issues they raise. I've counted 800 lint errors of no-unsafe-* rules, mostly in tests accessing un-typed stubs

3xtr4t3rr3str14l EDIT: added in the changes required by these rules being applied.

PR #2221, opened 2021-03-22
use createStubInstance helper function everywhere in lodestar tests

3xtr4t3rr3str14l:
uses @dapplion 's createStubInstance function (as mentioned in https://github.com/ChainSafe/lodestar/pull/2186#issuecomment-803149369) in every spot where applicable in `packages/lodestar/test/`

### DIFF
--- a/packages/lodestar/test/unit/api/impl/events/events.test.ts
+++ b/packages/lodestar/test/unit/api/impl/events/events.test.ts
@@ -9,12 +9,13 @@ import {
   VoluntaryExitEvent,
 } from "../../../../../src/api/impl/events";
 import {config} from "@chainsafe/lodestar-config/minimal";
-import sinon, {SinonStubbedInstance} from "sinon";
+import {SinonStubbedInstance} from "sinon";
 import {BeaconChain, ChainEvent, ChainEventEmitter, IBeaconChain} from "../../../../../src/chain";
 import {generateBlockSummary, generateEmptySignedBlock, generateSignedBlock} from "../../../../utils/block";
 import {expect} from "chai";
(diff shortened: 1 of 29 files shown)


dapplion:
Looks great! Tho we should only use that utility where it simplifies the types. It should not be used everywhere where using `sinon.createStubInstance` is good enough for typing to compile

3xtr4t3rr3str14l:
> Looks great! Tho we should only use that utility where it simplifies the types. It should not be used everywhere where using `sinon.createStubInstance` is good enough for typing to compile

ok, i'll revert the ones that were originally like `sinon.createStubInstance(Something)` (as opposed to the ones that were outlined in the original issue like `sinon.createStubInstance(Something) & Something`

packages/lodestar/test/unit/api/impl/events/events.test.ts
```diff
@@ -9,12 +9,13 @@ import {
...
 } from "../../../../../src/api/impl/events";
 import {config} from "@chainsafe/lodestar-config/minimal";
-import sinon, {SinonStubbedInstance} from "sinon";
+import {SinonStubbedInstance} from "sinon";
 import {BeaconChain, ChainEvent, ChainEventEmitter, IBeaconChain} from "../../../../../src/chain";
 import {generateBlockSummary, generateEmptySignedBlock, generateSignedBlock} from "../../../../utils/block";
 import {expect} from "chai";
 import {generateAttestation, generateEmptySignedVoluntaryExit} from "../../../../utils/attestation";
 import {generateState} from "../../../../utils/state";
+import sinon from "sinon";
```
dapplion: Merge the two "sinon" imports as 
```ts
import sinon, {SinonStubbedInstance} from "sinon";
```

dapplion (approved):

PR #2234, opened 2021-03-23
implement getStateValidators in lodestar-validator

### LINKED ISSUE #2232: Smarter validator client with multiple validators
**Describe the bug**

When running our validator client with multiple validators, our validator client spams the beacon node with repeated requests that could be batched

```
2021-03-23 15:00:00 [API]             debug: Request: 127.0.0.1 -> 127.0.0.1:9596	GET:/eth/v1/beacon/states/head/validators/0x810c608e7eda878ac8bed378f6e29504698422345895b74eb40717e5501a5b554d2e19b2e468fcd9ec05131ad4db557e	RequestId: 18275
2021-03-23 15:00:00 [API]             debug: Request: 127.0.0.1 -> 127.0.0.1:9596	GET:/eth/v1/beacon/states/head/validators/0x8136274d1fe87eb017d1e9a7ad36b35014d71601241100688f06c6b64d1ada508b97b2f8b54b79076360a9ecf735c5a0	RequestId: 18276
2021-03-23 15:00:00 [API]             debug: Request: 127.0.0.1 -> 127.0.0.1:9596	GET:/eth/v1/beacon/states/head/validators/0x81d4b7ca812383c670abb67ce04dddaa6b378dda7144869f4fa69d7fde47a12199e2eca3231d2aa1884b78c9721b45ef	RequestId: 18277
2021-03-23 15:00:00 [API]             debug: Request: 127.0.0.1 -> 127.0.0.1:9596	GET:/eth/v1/beacon/states/head/validators/0x82744af45bbb63fd5b74e0e6bec230b02ffc39972180232ae830b9ef535faaf534eaf6d1bf8f273b5c56d4d62b316c93	RequestId: 18278
```

Also for
```
POST:/eth/v1/validator/beacon_committee_subscriptions
```

And probably more

**Expected behavior**

validator client should be smarter and batch the requests

### LINKED ISSUE #2235: implement rest api server-side filters for getStateValidators
as noted here:
https://github.com/ChainSafe/lodestar/blob/4f19c7910259e3b1d7cd794053273e3bc5e793d4/packages/lodestar/src/api/impl/beacon/state/state.ts#L53

### LINKED ISSUE #2293: implement status filters for getStateValidators
see 
https://github.com/ChainSafe/lodestar/blob/27a6282187d349a5dfc830e822b568e6286c6919/packages/lodestar/src/api/impl/beacon/state/state.ts#L70

3xtr4t3rr3str14l:
resolve #2232, #2235, and #2293   by implementing `getStateValidators` on the lodestar-validator client API side and then use that inside `updateValidators()`


wemeetagain:
We want to update `getStateValidators` to accept an array of validator indices/pubkeys (https://github.com/ChainSafe/lodestar/blob/4f19c7910259e3b1d7cd794053273e3bc5e793d4/packages/lodestar-validator/src/api/impl/rest/beacon/state.ts#L10)
(using https://ethereum.github.io/eth2.0-APIs/#/Beacon/getStateValidators), and then only call that once in `updateValidators`

3xtr4t3rr3str14l:
> We want to update `getStateValidators` to accept an array of validator indices/pubkeys (
> 
> https://github.com/ChainSafe/lodestar/blob/4f19c7910259e3b1d7cd794053273e3bc5e793d4/packages/lodestar-validator/src/api/impl/rest/beacon/state.ts#L10
> )
> (using https://ethereum.github.io/eth2.0-APIs/#/Beacon/getStateValidators), and then only call that once in `updateValidators`

for the request you linked (from the eth2 api docs), that would return all the validators for the whole state.  the other similar call is this one, which just gets one validator per id:
https://ethereum.github.io/eth2.0-APIs/#/Beacon/getStateValidator

so in order to stay within the spec, wouldn't we just call getStateValidator in batch like i'm doing here?  (since i'm assuming we don't want to fetch the whole state validators every time we run `updateValidators`).  Or do you think it's better to deviate from the spec in this case?

wemeetagain:
The `getStateValidators` request has an `id` param that takes an array of validator indices/pubkeys, used to filter the response to only the validators provided.

3xtr4t3rr3str14l:
> The `getStateValidators` request has an `id` param that takes an array of validator indices/pubkeys, used to filter the response to only the validators provided.

ah gotcha, i didn't see that.  i'll restructure this PR accordingly

packages/lodestar/src/api/impl/beacon/state/state.ts
```diff
@@ -44,13 +45,31 @@ export class BeaconStateApi implements IBeaconStateApi {
...
+            state.validators[validatorIndex],
+            state.balances[validatorIndex],
+            getCurrentEpoch(this.config, state)
+          );
+          validators.push(validatorResponse);
+        }
+      }
+      return validators;
+    } else if (filters?.statuses) {
+      // TODO: implement status filters when needed
```
wemeetagain: Probably should make an issue to track this

packages/lodestar/src/api/rest/controllers/beacon/state/getValidators.ts
```diff
@@ -1,18 +1,33 @@
...
+export const getStateValidators: ApiController<ValidatorsQuery, Params> = {
   url: "/states/:stateId/validators",
 
   handler: async function (req, resp) {
-    const validators = await this.api.beacon.state.getStateValidators(req.params.stateId);
+    const filters: IValidatorFilters = {};
+    if (req.query.indices) {
+      filters.indices = mapValidatorIndices(
+        this.config,
+        typeof req.query.indices === "string" ? [req.query.indices] : req.query.indices
```
wemeetagain: I think you should be able to tell fastify (in the fastify schema) that req.query.indices should be an array

3xtr4t3rr3str14l: When i send the indices via the getStateValidators api call from the validator client as an array to the beacon node, if the array has only one item, it looks like fastify is supposed to coerce the query param to an array (as per our server setup here: https://github.com/ChainSafe/lodestar/blob/ea43a754d42831e5ad7bef2b4c3741822467926e/packages/lodestar/src/api/rest/index.ts#L54), and yet, if i send only one item in the indices array, it comes across on the beacon side as a string, even if i change the type in the schema for the getStateValidators controller.  

It looks like we also have other places where we are doing this manual array coercion (such as here: https://github.com/ChainSafe/lodestar/blob/ea43a754d42831e5ad7bef2b4c3741822467926e/packages/lodestar/src/api/rest/controllers/node/getPeers.ts#L37) so maybe this is something we should handle in a separate issue?  Or if I'm misunderstanding and/or you have another suggestion, then I'm open to that too.

packages/lodestar/src/api/rest/controllers/beacon/state/getValidators.ts
```diff
@@ -27,6 +42,19 @@ export const getStateValidators: ApiController<DefaultQuery, Params> = {
           },
         },
       },
+      querystring: {
+        type: "object",
+        required: [],
+        properties: {
+          indices: {
+            types: "string",
```
mpetrunic:
```suggestion
            type: "array",
```
should allow you to always get array even with single item

packages/lodestar/src/api/rest/controllers/beacon/state/getValidators.ts
mpetrunic: Check my other comment.

packages/lodestar/src/api/rest/controllers/beacon/state/getValidators.ts
3xtr4t3rr3str14l: ahh yeah ok i messed up `types` vs `type` when i had tried this originally.  just pushed a change for this.

packages/lodestar/src/api/impl/beacon/state/state.ts
```diff
@@ -44,17 +45,45 @@ export class BeaconStateApi implements IBeaconStateApi {
...
+            state.validators[validatorIndex],
+            state.balances[validatorIndex],
+            currentEpoch
+          );
+          validators.push(validatorResponse);
+        }
+      }
+      return validators;
+    }
+    if (filters?.statuses) {
```
mpetrunic: what if both indices and status are passed?

mpetrunic (changes requested):

packages/lodestar/src/api/impl/beacon/state/state.ts
3xtr4t3rr3str14l: ah yeah i meant to implement that, but forgot.  i will take care of that

packages/lodestar/src/api/impl/beacon/state/state.ts
```diff
@@ -44,17 +51,42 @@ export class BeaconStateApi implements IBeaconStateApi {
...
+            validatorIndex,
+            validator,
+            state.balances[validatorIndex],
+            currentEpoch
+          );
+          validators.push(validatorResponse);
+        }
+      }
+      return validators;
+    } else if (filters?.statuses) {
```
mpetrunic, line 81: If there is indices filter, I'm pretty sure we won't go trough this code path and filter by status as well.

You should add test case

3xtr4t3rr3str14l: I assumed your comment here (https://github.com/ChainSafe/lodestar/pull/2234#discussion_r605451789) was implying that we needed to account for when both filter types are passed in.  I've tested it locally, but I'll add a formal test case as well.

3xtr4t3rr3str14l: Either way, the current structure does allow for both filter types to be passed in.

3xtr4t3rr3str14l: Also, this particular line you pointed out could also account for if only the statuses are passed in.  Still, I'm adding more tests now.  Also just realized there were two tests that had a `skip` on them for `getStateValidators`, so I will enable those as well and modify as needed.

mpetrunic: Maybe I'm mistaken but if you pass both indices and status filter. Isn't this code gonna return all validators for given indices but it won't filter them by status?

mpetrunic: ah, nvm, missed some code

mpetrunic (approved):

wemeetagain (approved):

PR #2236, opened 2021-03-24
[DNM] Demo retained BeaconStateContext instance 

twoeths:
This is to make it simpler to debug the retained BeaconStateContext in #2184
+ Write a simple `runStateTransition3()` which run state transition and store to db only
+ Query blocks of 2 epochs in your existing dir and reprocess them using the above function
+ Start peer manager only, (and ignore sync + chores + notifier)
+ Heap size reach to 2GB at around loop 100-150
+ Check the retained memory  with `BeaconStateContext` row (500MB in the screenshot)
Screen Shot 2021-03-24 at 11 12 17

How to run:
+ `node --inspect lib/index.js beacon --network ${NETWORK} --root-dir ${YOUR_EXISING_DB}`

### DIFF
--- a/packages/lodestar/src/node/nodejs.ts
+++ b/packages/lodestar/src/node/nodejs.ts
@@ -224,7 +224,7 @@ export class BeaconNode {
             logger.info("Error dialing discovered peer", {peer: peer.toB58String()}, e);
           });
         }
-        await sleep(30000);
+        await sleep(15000);
       }
     }, 1000);
 
--- a/packages/lodestar/src/node/nodejs.ts
+++ b/packages/lodestar/src/node/nodejs.ts
@@ -203,51 +203,16 @@ export class BeaconNode {
 
     // void runNodeNotifier({network, chain, sync, config, logger, signal});
 
+    // we'll not see issue if we comment out the next 7 lines
     await libp2p.start();
     setTimeout(async function () {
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/node/nodejs.ts
```diff
@@ -193,11 +196,57 @@ export class BeaconNode {
...
+      for (let i = 0; ; i++) {
+        const toDialPeers: PeerId[] = [];
+        for (const peer of libp2p.peerStore.peers.values()) {
+          if (!libp2p.connectionManager.get(peer.id)) {
+            toDialPeers.push(peer.id);
+          }
+          if (toDialPeers.length >= 25) break;
+        }
+        logger.info("Run libp2p.dial", {i, toDialPeers: toDialPeers.length});
+        for (const peer of toDialPeers) {
```
twoeths: this is same code I found in https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/network/peers/discover.ts#L153

twoeths:
Update: the issue can be reproduce if we do `libp2p.dial` and reprocess blocks concurrently

wemeetagain:
Is this still an issue?

twoeths:
> Is this still an issue?

not with nodejs 14, closing for now

Issue #2145, opened 2021-03-08
Validator does invalid prepareBeaconCommitteeSubnet request


dapplion:
The multinode multithread test for the `dapplion/peer-manager` branch showed this repeatedly in the tests https://github.com/ChainSafe/lodestar/pull/2142/checks?check_run_id=2054916882

```
2021-03-08T08:51:26.1865174Z 2021-03-08 08:51:26 [VALI-A]          [31merror[39m: Failed to subscribe to committee subnet message=Request failed with response status 400
2021-03-08T08:51:26.1885124Z 2021-03-08 08:51:26 [VALI-A]          [31merror[39m: Failed to subscribe to committee subnet message=Request failed with response status 400
2021-03-08T08:51:26.1907499Z 2021-03-08 08:51:26 [NODE-A]           [33mwarn[39m: Request 451 failed validation. Reason: body[0] should have required property 'validator_index'
2021-03-08T08:51:26.1912117Z 2021-03-08 08:51:26 [NODE-A]           [33mwarn[39m: Request 452 failed validation. Reason: body[0] should have required property 'validator_index'
2021-03-08T08:51:26.1917276Z 2021-03-08 08:51:26 [VALI-A]          [31merror[39m: Failed to subscribe to committee subnet message=Request failed with response status 400
2021-03-08T08:51:26.1936306Z 2021-03-08 08:51:26 [VALI-A]          [31merror[39m: Failed to subscribe to committee subnet message=Request failed with response status 400
2021-03-08T08:51:26.1957755Z 2021-03-08 08:51:26 [NODE-A]           [33mwarn[39m: Request 453 failed validation. Reason: body[0] should have required property 'validator_index'
2021-03-08T08:51:26.1964305Z 2021-03-08 08:51:26 [NODE-A]           [33mwarn[39m: Request 454 failed validation. Reason: body[0] should have required property 'validator_index'
```

How could the validator send this invalid request?

See full test logs attached:

[logs.txt](https://github.com/ChainSafe/lodestar/files/6104276/logs.txt)

dapplion:
There's also this error

```
2021-03-08T08:47:16.8189999Z 2021-03-08 08:47:16 [NODE-A]           [33mwarn[39m: Route GET:/v1/events?topics=block&topics=head&topics=chain_reorg not found
[...]

PR #2260, opened 2021-03-25
Add BeaconCommitteeSubscription ssz type

twoeths:
part of #2145

**Motivation**

There is no ssz type for BeaconCommitteeSubscription so validator does not use snack case to conform to the standard api

**Description**

Add ssz type for it and validator should use `toJson({case: "snake"}) to communicate with node.

Run lodestar e2e test.

### DIFF
--- a/packages/validator/src/api/interface/validators.ts
+++ b/packages/validator/src/api/interface/validators.ts
@@ -1,13 +1,5 @@
 import {BLSPubkey, CommitteeIndex, Epoch, Root, phase0, Slot, ValidatorIndex} from "@chainsafe/lodestar-types";
 
-export type BeaconCommitteeSubscription = {
-  validatorIndex: number;
-  committeeIndex: number;
-  committeesAtSlot: number;
-  slot: number;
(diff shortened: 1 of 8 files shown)


packages/validator/src/api/interface/validators.ts
```diff
@@ -1,13 +1,5 @@
 import {BLSPubkey, CommitteeIndex, Epoch, Root, phase0, Slot, ValidatorIndex} from "@chainsafe/lodestar-types";
 
-export type BeaconCommitteeSubscription = {
-  validatorIndex: number;
-  committeeIndex: number;
-  committeesAtSlot: number;
-  slot: number;
-  isAggregator: boolean;
-};
-
```
dapplion, line 10: Type is also declared and used here https://github.com/ChainSafe/lodestar/blob/683af73def6c5150e86c75477be98cedefd11a15/packages/lodestar/src/api/impl/validator/interface.ts#L6

dapplion (approved):

mpetrunic (approved):

PR #2261, opened 2021-03-25
Use checkpoint state in validateAggregateAttestation

### LINKED ISSUE #2257: Use Checkpoint State in Attestation Gossip Validation
We may be able to use a checkpoint state directly in processing `AggregateAndProof`s, rather than dialing the state forward to the `attestation.data.slot`.
Dialing the state forward is a small bit of work (hashing as a result of `processSlot`) that may be avoided.
Investigate using the checkpoint state directly.
See https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/chain/validation/aggregateAndProof.ts#L84

Would save 21% of CPU time

Screenshot from 2021-03-24 20-51-47

dapplion:
**Motivation**

See https://github.com/ChainSafe/lodestar/issues/2257

**Description**

Closes https://github.com/ChainSafe/lodestar/issues/2257

### DIFF
--- a/packages/lodestar/src/chain/validation/aggregateAndProof.ts
+++ b/packages/lodestar/src/chain/validation/aggregateAndProof.ts
@@ -81,7 +81,7 @@ export async function validateAggregateAttestation(
   let attestationPreState;
   try {
     // the target state, advanced to the attestation slot
-    attestationPreState = await chain.regen.getBlockSlotState(attestation.data.target.root, attestation.data.slot);
+    attestationPreState = await chain.regen.getCheckpointState(attestation.data.target);
   } catch (e) {
     throw new AttestationError({
       code: AttestationErrorCode.MISSING_ATTESTATION_PRESTATE,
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/chain/validation/aggregateAndProof.ts
```diff
@@ -81,7 +81,7 @@ export async function validateAggregateAttestation(
   let attestationPreState;
   try {
     // the target state, advanced to the attestation slot
-    attestationPreState = await chain.regen.getBlockSlotState(attestation.data.target.root, attestation.data.slot);
+    attestationPreState = await chain.regen.getCheckpointState(attestation.data.target);
```
wemeetagain, line 84: maybe should be called `targetState` or `attestationTargetState`

dapplion: It's used in many places I can rename in a new PR

wemeetagain (approved):

PR #2266, opened 2021-03-25
Ensure all ChainEvents are re-emitted to external emitter

dapplion:
**Motivation**

Working on the Range Sync PR I notice that we forgot to re-emit the clockEpoch chain event.

**Description**

 I've refactored how eventHandlers are subscribed to Typescript ensures that this does not happen again, that all events are re-emitted from internal emitter to external emitter


wemeetagain (approved):
LGTM

Issue #2265, opened 2021-03-25
Add queue in gossip validation


dapplion:
Prater genesis has shown that we can get seriously overloaded with gossip validation work. Where our node spends 15 seconds doing nothing else than validating aggregates

**Goals**
For our node to survive:
- We can't do too much of 1 thing, each module has to yield fast to the macro queue and let other things happen.

**Proposal**
Add one or multiple validation queue(s) per gossip topic validation.
- Ensure that only one aggregate validation is performed at once, instead of 400 as Prater genesis showed.
- Add some mechanism for the validation function to yield to the macro queue often. To break long promise chains, we can run a `sleep(0)` every N jobs.
- Limit the queue size and drop jobs after the limit to contain the total load.
- Consider doing the queue a LIFO (Last In, First Out). The more recent aggregates are the most valuable so we should prioritize those.

This queue should be implemented in Lodestar side, not gossipsub's. However consider extending this method in our gossipsub implementation so each message is processed serially 
https://github.com/libp2p/js-libp2p-interfaces/blob/master/src/pubsub/index.js#L392-L399

----

Similarly, consider adding a queue to AttestationProcessor. Then the regen queue may not be needed anymore. Otherwise consider splitting the regen queue into different based on regen method.

PR #2274, opened 2021-03-26
Re-org gossip validation fns

dapplion:
**Motivation**

Preparing to develop https://github.com/ChainSafe/lodestar/issues/2265 it is much more convenient to modularize these validation functions in multiple files. To reduce the diff of an upcoming PR with sensitive logic I'm doing this bigger change in a previous single purpose PR

**Description**

Move `packages/lodestar/src/network/gossip/validator.ts` functions to smaller files

### DIFF
--- a/packages/lodestar/src/network/gossip/validatorFns/index.ts
+++ b/packages/lodestar/src/network/gossip/validatorFns/index.ts
@@ -1,0 +1,16 @@
+import {GossipType} from "../interface";
+import {validateAggregatedAttestation} from "./aggregatedAttestation";
+import {validateCommitteeAttestation} from "./attestation";
+import {validateAttesterSlashing} from "./attesterSlashing";
+import {validateBeaconBlock} from "./block";
+import {validateProposerSlashing} from "./proposerSlashing";
+import {validateVoluntaryExit} from "./voluntaryExit";
+
+export const objectValidatorFns = {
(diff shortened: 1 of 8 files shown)


mpetrunic (approved):

packages/lodestar/src/network/gossip/validatorFns/index.ts
```diff
@@ -0,0 +1,16 @@
+import {GossipType} from "../interface";
+import {validateAggregatedAttestation} from "./aggregatedAttestation";
+import {validateCommitteeAttestation} from "./attestation";
+import {validateAttesterSlashing} from "./attesterSlashing";
+import {validateBeaconBlock} from "./block";
+import {validateProposerSlashing} from "./proposerSlashing";
+import {validateVoluntaryExit} from "./voluntaryExit";
```
mpetrunic, line 7: we could rexport them since I believe we are reusing them in api?

dapplion: chain `packages/lodestar/src/chain/validation/*` exports the actual validator functions. Then the API and this `network/gossip/validatorFns/*` consume those. The functions being moved here only do error handling of the validation but not the validation itself.

wemeetagain (approved):

PR #2276, opened 2021-03-26
Move runStateTransition sleep

### LINKED ISSUE #2179: BeaconStateContext instances are retained in Promise

wemeetagain:
**Motivation**

With #2275, iterating through jobs in the queue yield to the event loop (with a `sleep(0)`).
This PR explores moving the `runStateTransition` `sleep(0)` call into the async iterators that call it.
This may be preferable for a few reasons:
- More abstractly, yielding to the event loop is something that should be done when async iterating over CPU-heavy operations, between iterations. It's more clear to put the `sleep(0)` in the async iteration, since it can be argued that the iteration over many values is the reason for needing the `sleep(0)` in the first place.
- Possibly resolves #2179. As seen in #2184, the interaction in `runStateTransition` between the logic and `sleep` cause states to occasionally be retained. By removing the `sleep`, this PR might resolve the issue. Further testing required.

### DIFF
--- a/packages/lodestar/src/chain/blocks/process.ts
+++ b/packages/lodestar/src/chain/blocks/process.ts
@@ -1,5 +1,6 @@
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {IForkChoice} from "@chainsafe/lodestar-fork-choice";
+import {sleep} from "@chainsafe/lodestar-utils";
 import {ChainEventEmitter} from "../emitter";
 import {IBlockJob, IChainSegmentJob} from "../interface";
 import {runStateTransition} from "./stateTransition";
@@ -146,6 +147,8 @@ export async function processChainSegment({
           validSignatures: true,
         });
         importedBlocks++;
+        // this avoids keeping our node busy processing blocks
+        await sleep(0);
       }
     } catch (e) {
       if (e instanceof RegenError) {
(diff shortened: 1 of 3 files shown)


dapplion:
Could detail here the reasoning for the change?

wemeetagain:
@dapplion Updated the description

twoeths:
so I tested this PR in my local Pyrmont db (which could be used to reproduce retained state issue)
+ The peer count is 7 maximum
+ the retained state issue is still there

also as reproduced in #2236, the retained state issue is not caused by `sleep(0)`

wemeetagain:
@ChainSafe/eth2-0 Can this get a review? I think its still worthwhile

3xtr4t3rr3str14l (approved):
i feel like i don't have enough knowledge about the chain code to be able to provide a counter-opinion or think about the implications beyond what you've mostly already said, but based on what you're saying, LGTM

PR #2277, opened 2021-03-26
Add JobQueue for each gossip topic

### LINKED ISSUE #2241: Throttle libp2p validation
During Prater genesis our node did nothing more than validate and process AggregateAndProof from gossip. Download and inspect the profile below.

- Our node should have some queue to prevent being DDOS by gossip messages.
- How should this queue be designed?
- Should we prioritize some messages (blocks) over others (attestations)?

[lodestar-64vali-pratergenesis-CPU-20210323T092838.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/6196079/lodestar-64vali-pratergenesis-CPU-20210323T092838.cpuprofile.zip)

----

Between `2021-03-23 15:14:19` and `2021-03-23 15:15:41` our Beacon node on Prater validated 487 AggregateAndProof and did nothing else.

### LINKED ISSUE #2265: Add queue in gossip validation

dapplion:
**Motivation**

See https://github.com/ChainSafe/lodestar/issues/2265

**Description**

Add a single queue for each gossip topic. Numbers are copied from here to start

https://github.com/sigp/lighthouse/blob/b34a79dc0b02e04441ba01fd0f304d1e203d877d/beacon_node/network/src/beacon_processor/mod.rs#L772-L825

but should be adjusted after experimentation in production settings

Closes https://github.com/ChainSafe/lodestar/issues/2265 closes https://github.com/ChainSafe/lodestar/issues/2241


packages/lodestar/src/network/gossip/interface.ts
```diff
@@ -36,49 +36,39 @@ export interface IGossipTopic {
...
-  | IAttesterSlashingTopic;
-
-export type GossipFn =
-  | ((signedBlock: phase0.SignedBeaconBlock) => Promise<void> | void)
-  | ((attestation: phase0.SignedAggregateAndProof) => Promise<void> | void)
-  | ((attestation: phase0.Attestation) => Promise<void> | void)
-  | ((voluntaryExit: phase0.SignedVoluntaryExit) => Promise<void> | void)
-  | ((proposerSlashing: phase0.ProposerSlashing) => Promise<void> | void)
-  | ((attesterSlashing: phase0.AttesterSlashing) => Promise<void> | void);
+export type GossipTopic = GossipTopicMap[keyof GossipTopicMap];
```
wemeetagain, line 51: :+1:

wemeetagain (approved):
LGTM

PR #2284, opened 2021-03-28
Use BLS batch verify in AggregateAndProof

### LINKED ISSUE #2225: validateAggregateAttestation could use batch signature verification

When running a beacon node with many validators (512) **28%** (13000 ms / 45000 ms) of total CPU time is spent in validateAggregateAttestation. This node runs all validators in the network, so this is not a realistic metric.

However, `validateAggregateAttestation()` verifies three sets of signatures separately, which can be optimized.

Screenshot from 2021-03-23 10-48-20

If it use batch signature verification it could reduce the pairings needed from 3 * 2 to 3 + 1, -33% max time savings. Is it really necessary to know which of the signatures is wrong? Could we throw an generic error `AttestationErrorCode.INVALID_SIGNATURES` 

**Alternatives**

We could also consider to implementing some verification queue which could de-duplicate verification if that's shows to be an issue

dapplion:
**Motivation**

See https://github.com/ChainSafe/lodestar/issues/2225

**Description**

- Moves the BLS signature sets types to the utils folder since it is not tied to phase0 or fast.
- Gets 3 signature sets from AggregateAndProof, and verifies them in batch 

I have not confirmed the performance improvement in production

Closes https://github.com/ChainSafe/lodestar/issues/2225


wemeetagain (approved):
:+1: We really don't need to know which signature is incorrect/why the aggAndProof is invalid. Will be interesting to see the improvement.

packages/beacon-state-transition/src/util/index.ts
```diff
@@ -2,21 +2,22 @@
...
-export * from "./proposer";
-export * from "./balance";
-export * from "./validatorStatus";
-export * from "./duties";
 export * from "./slot";
-export * from "./activation";
-export * from "./shuffle";
-export * from "./genesis";
+export * from "./validator";
+export * from "./validatorStatus";
```
dapplion, line 23: Sorted alphabetically to be able to track what we are exporting

mpetrunic (approved):

3xtr4t3rr3str14l (approved):

PR #2285, opened 2021-03-28
Add state cache basic metrics

### LINKED ISSUE #2153: Implement regen metrics
- Cache hits vs cache miss
- Cache frequency usage, last usage per item

dapplion:
**Motivation**

See https://github.com/ChainSafe/lodestar/issues/2153

**Description**

Adds basic metrics for the state cache. Total gets + hit, allows to compute frequency, total usage and hit ratio.

Closes https://github.com/ChainSafe/lodestar/issues/2153

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -87,8 +87,8 @@ export class BeaconChain implements IBeaconChain {
       genesisTime: this.genesisTime,
       signal: this.abortController.signal,
     });
-    this.stateCache = new StateContextCache();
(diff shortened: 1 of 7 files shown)


twoeths (approved):

wemeetagain:
:+1: 
It would be (more) useful to have metrics in the regen module. We would calculate the same sorts of things (hit/miss) but be able to measure the cost/why of misses.

3xtr4t3rr3str14l (approved):

wemeetagain:
@dapplion I think we should close this in favor of suggestion above:

add metrics to regen module

PR #2287, opened 2021-03-29
Prevent copying .git to the final layer

dapplion:
Unnecessary bloats the size of the final layer

```
  517.8 MiB [##########] /node_modules
  389.2 MiB [#######   ] /.git
   29.2 MiB [          ] /packages
  568.0 KiB [          ]  yarn.lock
  168.0 KiB [          ] /audits
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,8 @@ COPY . .
 RUN yarn install --non-interactive --frozen-lockfile
 
 RUN node ./scripts/getGitData /usr/app/.git-data.json
+# Prevent copying .git to the final layer
+RUN rm -r .git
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -10,6 +10,8 @@ COPY . .
 RUN yarn install --non-interactive --frozen-lockfile
 
 RUN node ./scripts/getGitData /usr/app/.git-data.json
```
mpetrunic:
```suggestion
RUN node ./scripts/getGitData /usr/app/.git-data.json && rm -r .git
```
to avoid another docker layer

mpetrunic (approved):

PR #2291, opened 2021-03-29
Fix wasteful number serialization in DB

### LINKED ISSUE #2255: Fix wasteful number serialization in DB
**Describe the bug**

Several slot indices store 64 bytes per entry where 8 bytes would suffice

See https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/db/api/beacon/repositories/blockArchive.ts#L134 for an example

**Expected behavior**

Should only store 8 bytes per slot.

3xtr4t3rr3str14l:
fixes #2255


wemeetagain:
Looks like the unit tests need to be updated, but the implementation looks good :+1:

twoeths (approved):

mpetrunic (approved):

wemeetagain (approved):

PR #2295, opened 2021-03-30
Index attestation repository by attestation data root

### LINKED ISSUE #2222: getAttestationsByDataRoot excessive hashing
**Describe the bug**

When running a beacon node with many validators (512) **10%** (4030 ms / 45000 ms) of total CPU time is spent in `getAttestationsByDataRoot` since it is hashing attestations over and over and dropping the result.

https://github.com/ChainSafe/lodestar/blob/0bddc7ac75a0fab0356d29ebce6ea14f94a96458/packages/lodestar/src/db/api/beacon/repositories/attestation.ts#L29-L38

**Expected behavior**

It should be something easy to optimize for example doing what the TODO says
```
//TODO: add secondary index slot => root => AttestationData
```

**Steps to Reproduce**

Run the multiNodeSingleThread test with 128 validators per node and attach a profiler

**Screenshots**

This is a very special condition since 100% of the validators of the network are connected to the same node. So the node is doing the same getAttestationsByDataRoot checks over and over, which won't happen as much in a real network, when you run a fraction of total validators. However the pool of attestations is likely to be bigger and have a greater impact

Screenshot from 2021-03-23 10-24-37

Screenshot from 2021-03-23 10-23-37

Screenshot from 2021-03-23 10-24-09

twoeths:
**Motivation**

We hash have to hash AttestationData too much in order to get attestation by data root

**Description**

+ Create an index inside attestation repository
+ This index should be technically deleted along with an attestation but I chose to delete per finalized epoch

Closes #2222

### DIFF
--- a/packages/lodestar/src/db/api/beacon/repositories/attestation.ts
+++ b/packages/lodestar/src/db/api/beacon/repositories/attestation.ts
@@ -25,14 +25,16 @@ export class AttestationRepository extends Repository<Uint8Array, phase0.Attesta
     const slot = value.data.slot;
     const dataRootHex = toHexString(attestationDataRoot);
     const rootHex = toHexString(key);
-    if (!this.dataRootIndex.get(slot)) {
-      this.dataRootIndex.set(slot, new Map());
+    let slotIndex = this.dataRootIndex.get(slot);
+    if (!slotIndex) {
+      slotIndex = new Map();
+      this.dataRootIndex.set(slot, slotIndex);
     }
-    if (!this.dataRootIndex.get(slot)?.get(dataRootHex)) {
-      this.dataRootIndex.get(slot)?.set(dataRootHex, [rootHex]);
+    if (!slotIndex.get(dataRootHex)) {
+      slotIndex.set(dataRootHex, [rootHex]);
     } else {
-      const rootHexes = new Set([rootHex, ...(this.dataRootIndex.get(slot)?.get(dataRootHex) || [])]);
-      this.dataRootIndex.get(slot)?.set(dataRootHex, Array.from(rootHexes));
+      const rootHexes = new Set([rootHex, ...(slotIndex.get(dataRootHex) || [])]);
+      slotIndex.set(dataRootHex, Array.from(rootHexes));
     }
     await super.put(key, value);
   }
@@ -54,12 +56,7 @@ export class AttestationRepository extends Repository<Uint8Array, phase0.Attesta
     if (!dataRootsAtSlot) return [];
     const rootHexes = dataRootsAtSlot.get(toHexString(attestationDataRoot));
     if (!rootHexes) return [];
-    const attestations = await Promise.all(
-      rootHexes.map((rootHex) => {
-        const root = fromHexString(rootHex);
-        return this.get(root);
-      })
-    );
+    const attestations = await Promise.all(rootHexes.map((rootHex) => this.get(fromHexString(rootHex))));
     return attestations.filter(notNullish);
   }
 
@@ -76,8 +73,7 @@ export class AttestationRepository extends Repository<Uint8Array, phase0.Attesta
         return a.data.slot < finalizedEpochStartSlot;
       })
     );
-    const indexedSlots = Array.from(this.dataRootIndex.keys());
-    for (const slot of indexedSlots) {
+    for (const slot of this.dataRootIndex.keys()) {
       if (slot < finalizedEpochStartSlot) this.dataRootIndex.delete(slot);
     }
   }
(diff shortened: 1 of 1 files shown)


packages/lodestar/src/db/api/beacon/repositories/attestation.ts
```diff
@@ -50,5 +76,9 @@ export class AttestationRepository extends Repository<Uint8Array, phase0.Attesta
         return a.data.slot < finalizedEpochStartSlot;
       })
     );
+    const indexedSlots = Array.from(this.dataRootIndex.keys());
+    for (const slot of indexedSlots) {
```
dapplion: You can iterate `this.dataRootIndex.keys()` directly

packages/lodestar/src/db/api/beacon/repositories/attestation.ts
```diff
@@ -27,14 +50,17 @@ export class AttestationRepository extends Repository<Uint8Array, phase0.Attesta
...
-    });
+    const dataRootsAtSlot = this.dataRootIndex.get(slot);
+    if (!dataRootsAtSlot) return [];
+    const rootHexes = dataRootsAtSlot.get(toHexString(attestationDataRoot));
+    if (!rootHexes) return [];
+    const attestations = await Promise.all(
+      rootHexes.map((rootHex) => {
+        const root = fromHexString(rootHex);
+        return this.get(root);
+      })
```
dapplion: Consider for succinctness
```ts
rootHexes.map((rootHex) => this.get(fromHexString(rootHex)))
```

packages/lodestar/src/db/api/beacon/repositories/attestation.ts
```diff
@@ -10,10 +12,31 @@ import {IDatabaseController, Bucket, Repository} from "@chainsafe/lodestar-db";
  * Removed when included on chain or old
  */
 export class AttestationRepository extends Repository<Uint8Array, phase0.Attestation> {
+  // For each slot, index attestation roots by attestation data root hex
+  // do batch remove per finalized checkpoint
+  private dataRootIndex = new Map<Slot, Map<string, string[]>>();
```
dapplion, line 17: If this map is stored in memory but the attestations to disk, if the node restarts will it loose access to the stored attestations? Should we persist this index to disk too?

twoeths: this index is used for aggregator and its per slot. If we restart we need to index the current slot going forward, not the past slot. So I think we don't need to persist it.

packages/lodestar/src/db/api/beacon/repositories/attestation.ts
```diff
@@ -10,10 +12,31 @@ import {IDatabaseController, Bucket, Repository} from "@chainsafe/lodestar-db";
...
+    const slot = value.data.slot;
+    const dataRootHex = toHexString(attestationDataRoot);
+    const rootHex = toHexString(key);
+    if (!this.dataRootIndex.get(slot)) {
+      this.dataRootIndex.set(slot, new Map());
+    }
+    if (!this.dataRootIndex.get(slot)?.get(dataRootHex)) {
+      this.dataRootIndex.get(slot)?.set(dataRootHex, [rootHex]);
+    } else {
+      const rootHexes = new Set([rootHex, ...(this.dataRootIndex.get(slot)?.get(dataRootHex) || [])]);
```
wemeetagain: why is `rootHex` included here?

Edit: nvm, the index goes: slot -> dataRoot -> root[]

packages/lodestar/src/db/api/beacon/repositories/attestation.ts
```diff
@@ -10,10 +12,31 @@ import {IDatabaseController, Bucket, Repository} from "@chainsafe/lodestar-db";
...
   constructor(config: IBeaconConfig, db: IDatabaseController<Buffer, Buffer>) {
     super(config, db, Bucket.phase0_attestation, config.types.phase0.Attestation);
   }
 
+  async put(key: Uint8Array, value: phase0.Attestation): Promise<void> {
+    const attestationDataRoot = this.config.types.phase0.AttestationData.hashTreeRoot(value.data);
+    const slot = value.data.slot;
+    const dataRootHex = toHexString(attestationDataRoot);
+    const rootHex = toHexString(key);
+    if (!this.dataRootIndex.get(slot)) {
```
wemeetagain: You might structure this part like:
```ts
let slotIndex = this.dataRootIndex.get(slot);
if (!slotIndex) {
  slotIndex = new Map();
  this.dataRootIndex.set(slot, slotIndex);
}
slotIndex.get(...)....
```
to avoid repeated `this.dataRootIndex.get(slot)?` in the rest.

wemeetagain (approved):

PR #2303, opened 2021-03-30
Prepare fast state transition for forks

wemeetagain:
**Motivation**

We want our "fast" state transition to be multi-fork aware. To that end, there should be fork-aware utilities and a fork-aware state transition function. These files should also be organized in a fork-agnostic location.

**Description**

- Move `phase0.fast` top-level state transition function (`fastStateTransition`), fast utils (cached beacon state, etc), and  signature set utils to a top-level `fast/` directory.
- Add stubbed switch/case(s) in `fastStateTransition` for other-fork logic (eg: switching out `processSlots`, `processBlocks` and upgrading the `BeaconState`)
- Re-export top-level `fast/` exports as `fast` namespace.
- Minimal renaming/importing in `lodestar` and `spec-test-runner` packages

A followup PR can update the chain/regen/etc to use `allForks.BeaconState` rather than `phase0.BeaconState` (which will remove a lot of the explicit castings to `CachedBeaconState<allForks.BeaconState>`


packages/beacon-state-transition/src/fast/stateTransition.ts
```diff
@@ -0,0 +1,61 @@
...
+      throw new Error(`Slot processing not implemented for fork ${preFork}`);
+  }
+
+  // perform state upgrades
+  // TODO
+
+  // verify signature
+  if (verifyProposer) {
+    if (!verifyProposerSignature(postState, signedBlock)) {
+      throw new Error("Invalid block signature");
```
twoeths, line 41: maybe add block slot/root here

packages/beacon-state-transition/src/fast/stateTransition.ts
```diff
@@ -0,0 +1,61 @@
...
+      processBlock(postState as CachedBeaconState<phase0.BeaconState>, block as phase0.BeaconBlock, verifySignatures);
+      break;
+    default:
+      throw new Error(`Block processing not implemented for fork ${blockFork}`);
+  }
+
+  // verify state root
+  if (verifyStateRoot) {
+    if (!config.types.Root.equals(block.stateRoot, postState.tree.root)) {
+      throw new Error("Invalid state root");
```
twoeths, line 57: maybe add block stateRoot, and the expected stateRoot here

packages/beacon-state-transition/src/fast/util/epochProcess.ts
```diff
@@ -8,8 +8,8 @@ import {
   computeStartSlotAtEpoch,
   getChurnLimit,
   isActiveValidator,
-} from "../../../util";
-import {FAR_FUTURE_EPOCH} from "../../../constants";
+} from "../../util";
+import {FAR_FUTURE_EPOCH} from "../../constants";
 import {
```
twoeths, line 13: The `statusProcessEpoch` function below seems to be better to put into `phase0` folder, it's fine for now through

packages/beacon-state-transition/src/fast/stateTransition.ts
dapplion: Should we use detailed typed `LodestarError`s in the state transition function too? To be tackled in new PR

dapplion (approved):

packages/beacon-state-transition/src/fast/stateTransition.ts
wemeetagain: These errors (this and below) are untouched from before. I'll make an issue to revisit state transition errors for a future PR!

PR #2312, opened 2021-03-31
Refactor metrics

dapplion:
**Motivation**

- Our metrics setup is unnecessary verbose, having to declare each metric 3 times in different places
- `prom-client` native classes do not have sufficient features to allow the gossip validation queue metrics to be separated by labels instead of different metric names. Having a single `gossip_validation_queue` and then each topic use a different label simplifies a lot the charts in Grafana.

**Description**

- Create all metrics in functions without return type, to prevent declaring metrics multiple times.
- Create `GaugeExtra` and `HistogramExtra` which allow to:
  - Create child classes with pre-defined labels
  - Attach multiple `collect()` functions after instantiation.
- Rename `IBeaconMetrics` to `IMetrics`
- Remove un-used PushGateway code
- Update metric name in Grafana dashboard


wemeetagain (approved):
Sad to see the PushGateway code go, but we can always bring it back if/when we need.

Other than that, looks great.

PR #2316, opened 2021-03-31
Add getLatestWeakSubjectivityCheckpointEpoch function

### LINKED ISSUE #2304: Add `getLatestWeakSubjectivityCheckpointEpoch` function

We would like to implement various features surrounding the weak subjectivity state.
In order to do so, we need to implement a few core utility functions, possibly just 1 to begin.

Implement `get_latest_weak_subjectivity_checkpoint_epoch` as described here:
https://notes.ethereum.org/@adiasg/weak-subjectvity-eth2#Distributing-Weak-Subjectivity-Checkpoint-States

Ideally we would implement twice, once for `allForks.BeaconState`, and another for `CachedBeaconState<allForks.BeaconState>`.

The code can likely live in `packages/beacon-state-transition/src/util/weakSubjectivity.ts` and
 `packages/beacon-state-transition/src/fast/util/weakSubjectivity.ts`

3xtr4t3rr3str14l:
resolves #2304  for both allForks.BeaconState, and CachedBeaconState<allForks.BeaconState>

### DIFF
--- a/packages/beacon-state-transition/src/util/weakSubjectivity.ts
+++ b/packages/beacon-state-transition/src/util/weakSubjectivity.ts
@@ -12,9 +12,17 @@ export function getLatestWeakSubjectivityCheckpointEpoch(
   state: allForks.BeaconState,
   safetyDecay = 0.1
 ): Epoch {
-  let weakSubjectivityMod = config.params.MIN_VALIDATOR_WITHDRAWABILITY_DELAY;
   const valCount = getActiveValidatorIndices(state, getCurrentEpoch(config, state)).length;
+  return getEpochFromValCount(config, state, safetyDecay, valCount);
+}
 
+export function getEpochFromValCount(
+  config: IBeaconConfig,
+  state: allForks.BeaconState,
+  safetyDecay = 0.1,
+  valCount: number
+): Epoch {
+  let weakSubjectivityMod = config.params.MIN_VALIDATOR_WITHDRAWABILITY_DELAY;
   if (valCount >= config.params.MIN_PER_EPOCH_CHURN_LIMIT * config.params.CHURN_LIMIT_QUOTIENT) {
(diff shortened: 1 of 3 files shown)


packages/beacon-state-transition/src/util/weakSubjectivity.ts
```diff
@@ -0,0 +1,33 @@
...
+export function getLatestWeakSubjectivityCheckpointEpoch(
+  config: IBeaconConfig,
+  state: allForks.BeaconState,
+  safetyDecay = 0.1
+): Epoch {
+  const valCount = getActiveValidatorIndices(state, getCurrentEpoch(config, state)).length;
+  return getEpochFromValCount(config, state, safetyDecay, valCount);
+}
+
+export function getEpochFromValCount(
```
wemeetagain: Can you make this function look like this:
```ts
function getWeakSubjectivityCheckpointEpoch(
  config: IBeaconConfig,
  finalizedEpoch: Epoch,
  activeValidatorCount: number,
  safetyDecay = 0.1
): Epoch;
```

wemeetagain (approved):
:100:

dapplion (approved):
Looks good! @wemeetagain What's the next step now? Where should this function be used? Consider opening a follow up issue

wemeetagain:
> Looks good! @wemeetagain What's the next step now? Where should this function be used?

Depending on the route we go next, we may not need to use this, its just nice to have.

The end goal is to have infrastructure in place to provide syncing by weak subjectivity checkpoint+state "by default" for our beacon node, similar to the experience we currently give for starting from genesis.

Imo we can break the problem down into a few parts: creating the ws state, distributing the ws state, loading the ws state.

As far as creating the state, we use something like https://github.com/adiasg/eth2-ws-provider (which just uses the standard API) or add a custom API endpoint to retrieve the weak subjectivity state.

As far as serving the state, it would be nice to use IPFS or similar to leverage existing ChainSafe Files tooling.

And whatever the distribution, we will need to extend our CLI to do this fetch of the ws state on initial boot.

PR #2317, opened 2021-04-01
pretty print peer info

### LINKED ISSUE #2250: Prater logs - pretty print peer info
Instead of printing 
```
2021-03-23 15:24:05 [NETWORK]         debug: action peer=16Uiu2HAkznk3ZL5qgBMR1UDunzSkdxTvuECqehWzRLRdT2DEnamK, requestId=62
```
We could print the peerId shorter and add the client metadata
```
2021-03-23 15:24:05 [NETWORK]         debug: action client=prysm/v1.1.0 peer=16..DEnamK, requestId=62
```

3xtr4t3rr3str14l:
resolves #2250

### DIFF
--- a/packages/lodestar/src/network/util.ts
+++ b/packages/lodestar/src/network/util.ts
@@ -8,7 +8,7 @@ import {Method, MethodResponseType, Methods, ReqRespEncoding, RequestId} from ".
 import Multiaddr from "multiaddr";
 import {networkInterfaces} from "os";
 import {ENR} from "@chainsafe/discv5";
-import PeerStore from "libp2p/src/peer-store";
+import MetadataBook from "libp2p/src/peer-store/metadata-book";
 
 // req/resp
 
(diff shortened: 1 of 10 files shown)


packages/lodestar/src/network/util.ts
```diff
@@ -100,3 +101,12 @@ export function clearMultiaddrUDP(enr: ENR): void {
...
   enr.delete("udp6");
 }
+
+export function prettyPrintPeerId(peerId: PeerId): string {
+  const id = peerId.toB58String();
+  return `${id.substr(0, 2)}...${id.substr(id.length - 6, id.length)}`;
+}
+
+export function getAgentVersionFromPeerStore(peerId: PeerId, peerStore: PeerStore): string {
+  return peerStore.get(peerId)?.metadata?.get("AgentVersion")?.toString() || "";
```
dapplion: Interacting with the peer store can be very expensive. Please benchmark this in isolation to check its cost

3xtr4t3rr3str14l: as discussed on the planning meeting last Friday, I changed this function to use the metadatabook instead of accessing via `peerstore.get()`.  when i benchmark the function in it's latest form (via the test added in this PR) all 2000 of the added peers take less than 1ms to access the `AgentVersion` except for roughly about 9 of said peers on average.

cc @wemeetagain

wemeetagain (approved):

Issue #2306, opened 2021-03-31
Benchmark + optimize aggregate and proof validation


dapplion:
Metrics from Prater in one of our Contabo VPS S size nodes, show that when the node is synced 80% of CPU time is spent validating aggregate and proof gossip messages. To keep up we also drop 35% of all received messages.

![Screenshot from 2021-03-31 09-21-52](https://user-images.githubusercontent.com/35266934/113114663-dcb3d280-920b-11eb-8656-8b4b77b66be0.png)

The average job duration is 20-30ms. In Prater stable conditions all attestation target states should be in the cache and cost 0 to get. Then the only big cost is signature validation, which has 3 (selection proof + aggregator sig + att sig aggregate). A BLS sig costs between 1-2 ms, and since the 3 sigs are verified in batch it should have a discount of ~50%. So the total job time should be between 1.5-3ms.

We should investigate the performance of that validation since there is significant room for improvement.

tuyennhv:
found duplicate SignedAggregateAndProof
```
Apr-06 02:20:50.858 [NETWORK]          ^[[32minfo^[[39m: @@@ validateAggregatedAttestation slot=96703, aggregator=7760, signature=0xa7950b19d1805589aa59427701d9c9770842addc071f4b262af21468bc4371f9b48487d90b089b9f5d6f61ab4ddc30d419489aab6104dd2800aae5ba54bdd07cc065820a370ebb2372c973dc9c8b6a8fd51bc24374a926adfb84e7768d2de6f7, hash=0x03e7c0720870e581af028a305ef87ad53a2e3eb8dd885905630928e9c0205473
Apr-06 02:20:51.454 [NETWORK]          ^[[32minfo^[[39m: @@@ validateAggregatedAttestation gossip - AggregateAndProof - accept slot=96703, aggregator=7760, signature=0xa7950b19d1805589aa59427701d9c9770842addc071f4b262af21468bc4371f9b48487d90b089b9f5d6f61ab4ddc30d419489aab6104dd2800aae5ba54bdd07cc065820a370ebb2372c973dc9c8b6a8fd51bc24374a926adfb84e7768d2de6f7, hash=0x03e7c0720870e581af028a305ef87ad53a2e3eb8dd885905630928e9c0205473
```

twoeths:
[...]

PR #2329, opened 2021-04-06
Improve AggregateAndProof gossip validation

twoeths:
**Motivation**

+ AggregateAndProof as TreeBacked to avoid multiple `struct_hashTreeRoot`
+ Increase JobQueue for AggregateAndProof 
+ Filter out known aggregated attestation

**Description**

part of #2306

### DIFF
--- a/packages/lodestar/src/chain/validation/aggregateAndProof.ts
+++ b/packages/lodestar/src/chain/validation/aggregateAndProof.ts
@@ -68,8 +68,6 @@ export async function validateGossipAggregateAndProof(
     });
   }
 
-  // TODO: check pool of aggregates if already seen (not a dos vector check)
-
   await validateAggregateAttestation(config, chain, signedAggregateAndProof, attestationJob);
 }
 
--- a/packages/lodestar/src/db/beacon.ts
+++ b/packages/lodestar/src/db/beacon.ts
@@ -48,7 +48,7 @@ export class BeaconDb extends DatabaseService implements IBeaconDb {
     this.badBlock = new BadBlockRepository(this.config, this.db);
     this.block = new BlockRepository(this.config, this.db);
     this.pendingBlock = new PendingBlockRepository(this.config, this.db);
-    this.seenAttestationCache = new SeenAttestationCache(5000);
+    this.seenAttestationCache = new SeenAttestationCache(this.config, 5000);
     this.blockArchive = new BlockArchiveRepository(this.config, this.db);
     this.stateArchive = new StateArchiveRepository(this.config, this.db);
     this.attestation = new AttestationRepository(this.config, this.db);
--- a/packages/lodestar/src/db/beacon.ts
+++ b/packages/lodestar/src/db/beacon.ts
@@ -48,7 +48,7 @@ export class BeaconDb extends DatabaseService implements IBeaconDb {
     this.badBlock = new BadBlockRepository(this.config, this.db);
     this.block = new BlockRepository(this.config, this.db);
     this.pendingBlock = new PendingBlockRepository(this.config, this.db);
-    this.seenAttestationCache = new SeenAttestationCache(this.config, 5000);
+    this.seenAttestationCache = new SeenAttestationCache(this.config, 2048);
     this.blockArchive = new BlockArchiveRepository(this.config, this.db);
     this.stateArchive = new StateArchiveRepository(this.config, this.db);
     this.attestation = new AttestationRepository(this.config, this.db);
--- a/packages/lodestar/src/db/seenAttestationCache.ts
+++ b/packages/lodestar/src/db/seenAttestationCache.ts
@@ -4,7 +4,8 @@
  */
 import {phase0} from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
-import {toHexString} from "@chainsafe/ssz";
+import {readonlyValues, toHexString} from "@chainsafe/ssz";
+import {assert} from "@chainsafe/lodestar-utils";
 
 /**
  * USed to verify gossip attestation. When there are multiple
@@ -12,6 +13,8 @@ import {toHexString} from "@chainsafe/ssz";
  */
 export class SeenAttestationCache {
   private cache: Map<string, boolean>;
+  // key as AttestationDelta root hex, value as aggregationBits
+  private attDataCache: Map<string, boolean[]>;
   private readonly config: IBeaconConfig;
   private readonly maxSize: number;
 
@@ -19,6 +22,7 @@ export class SeenAttestationCache {
     this.config = config;
     this.maxSize = maxSize;
     this.cache = new Map<string, boolean>();
+    this.attDataCache = new Map<string, boolean[]>();
   }
 
   addCommitteeAttestation(attestation: phase0.Attestation): void {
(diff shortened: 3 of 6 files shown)


packages/lodestar/src/chain/validation/aggregateAndProof.ts
```diff
@@ -69,7 +69,12 @@ export async function validateGossipAggregateAndProof(
     });
   }
 
-  // TODO: check pool of aggregates if already seen (not a dos vector check)
+  if (await db.aggregateAndProof.hasAttestation(aggregate)) {
```
dapplion: Is this cache of seen aggregates in memory or in disk?

twoeths: it's in disc with key as aggregated attestation root

dapplion: If we get 600 aggregates per slot would that be "a lot" of disk reads? Would it be worth it to keep a the roots in memory in a set? Keep them only for 2 epochs and them drop them

twoeths: seems like SeenAttestationCache was not hit, will leverage it instead of disk read.

twoeths:
some statistic (found cached AggregateAndProof / total)
```
slot 104922: 32/596
slot 104956: 99/555
slot 104954: 38/437
slot 104948: 50/525
slot 105057: 0/14
slot 105056: 2/49
```

twoeths:
based on prater block explorer, I think there are at most 64 attestation data based on 64 different committee index of the same slot. There are ~600 different AggregateAndProof messages of the same slot because each aggregator see different `aggregationBits`. So we can improve the cache strategy by:
+ cache by AttestationData root + number of attesters (based on `aggregationBits`)
+ if same AttestationData root and not greater number of attesters, we can ignore that AggregationAndProof

what do you think @dapplion ?

dapplion:
> based on prater block explorer, I think there are at most 64 attestation data based on 64 different committee index of the same slot. There are ~600 different AggregateAndProof messages of the same slot because each aggregator see different `aggregationBits`. So we can improve the cache strategy by:
> 
> * cache by AttestationData root + number of attesters (based on `aggregationBits`)
> * if same AttestationData root and not greater number of attesters, we can ignore that AggregationAndProof
> 
> what do you think @dapplion ?

@wemeetagain Can confirm but we should take any aggregate that includes new bits so the fork choice is aware of as many attestations as possible

twoeths:
in that case we should cache AttestationData root + XOR of (`aggregationBits`):

+ {rootA, [0, 1, 1, 0]} comes => cache {rootA => [0, 1, 1, 0]}
+ {rootA, [1, 0, 1, 0]} comes => cache {rootA => [1, 1, 1, 0]}
+ {rootA, [1, 1, 0, 0]} comes => IGNORE since its `aggregationBits` are subset of the cached one
+ {rootA, [1, 1, 1, 0]} comes => IGNORE since its `aggregationBits` are same to the cached one
+ {rootA, [0, 0, 0, 1} comes => cache {rootA => [1, 1, 1, 1]}
+ from now on whenever rootA comes => IGNORE

twoeths:
with the cache by AttestationData root hex + XOR (aggregationBits) we have (found seen gossip AggregationAndProof/total):

```
slot 113716: 397/546
slot 113720: 384/553
slot 113736: 394/596
```

so we don't have to validate every SignedAggregateAndProof like before, just around 30% of that

packages/lodestar/src/db/seenAttestationCache.ts
```diff
@@ -3,19 +3,26 @@
...
+import {assert} from "@chainsafe/lodestar-utils";
 
 /**
  * USed to verify gossip attestation. When there are multiple
  * attestation from same validator
  */
 export class SeenAttestationCache {
   private cache: Map<string, boolean>;
-
+  // key as AttestationDelta root hex, value as aggregationBits
```
wemeetagain:
```suggestion
  // key as AttestationData root hex, value as aggregationBits
```

packages/lodestar/src/db/seenAttestationCache.ts
```diff
@@ -59,8 +93,10 @@ export class SeenAttestationCache {
...
   }
 
-  // serialize aggregate key as concatenation of interested properties
-  private aggregateAndProofKey(aggreate: phase0.AggregateAndProof): string {
-    return "" + aggreate.aggregatorIndex + aggreate.aggregate.data.target.epoch;
+  /**
+   * We're only interested in the AttestationData + aggregationBits inside AggregateAndProof.
+   */
+  private aggregateAndProofKey(value: phase0.AggregateAndProof): string {
+    return toHexString(this.config.types.phase0.AttestationData.hashTreeRoot(value.aggregate.data));
```
wemeetagain:
```suggestion
    return toHexString(this.config.getTypes(value.aggregate.slot).AttestationData.hashTreeRoot(value.aggregate.data));
```

packages/lodestar/src/db/seenAttestationCache.ts
```diff
@@ -25,7 +32,25 @@ export class SeenAttestationCache {
...
+    } else {
+      assert.equal(
+        aggBit.length,
+        cachedAggBit.length,
+        "Length of AggregateAndProof aggregationBits is not the same to cache"
+      );
+      this.attDataCache.set(
+        key,
+        aggBit.map((_, i) => aggBit[i] || cachedAggBit[i])
+      );
```
wemeetagain: Might consider something like:
```ts
for (const [i, bit] of aggBit.entries()) {
  if (bit) {
    cachedAggBit[i] = true;
  }
}
```
tbh i'm not sure which is better.

packages/lodestar/src/db/beacon.ts
```diff
@@ -48,7 +48,7 @@ export class BeaconDb extends DatabaseService implements IBeaconDb {
     this.badBlock = new BadBlockRepository(this.config, this.db);
     this.block = new BlockRepository(this.config, this.db);
     this.pendingBlock = new PendingBlockRepository(this.config, this.db);
-    this.seenAttestationCache = new SeenAttestationCache(5000);
+    this.seenAttestationCache = new SeenAttestationCache(this.config, 2048);
```
wemeetagain, line 51: curious why this is decreased?

wemeetagain (changes requested):

packages/lodestar/src/db/beacon.ts
twoeths: currently it's used for 2 caches: (committee) Attestation and AttestationAndProof but this PR changes to cache them separately so mathematically it should be `5000 / 2 = 2500`, but I changed it to `2048`

packages/lodestar/src/network/gossip/validator.ts
```diff
@@ -20,7 +20,8 @@ import {
 // Numbers from https://github.com/sigp/lighthouse/blob/b34a79dc0b02e04441ba01fd0f304d1e203d877d/beacon_node/network/src/beacon_processor/mod.rs#L69
 const gossipQueueOpts: {[K in GossipType]: {maxLength: number; type: QueueType}} = {
   [GossipType.beacon_block]: {maxLength: 1024, type: QueueType.FIFO},
-  [GossipType.beacon_aggregate_and_proof]: {maxLength: 1024, type: QueueType.LIFO},
+  // this is different from lighthouse's, there are more gossip aggregate_and_proof than gossip block
+  [GossipType.beacon_aggregate_and_proof]: {maxLength: 4096, type: QueueType.LIFO},
```
3xtr4t3rr3str14l, line 24: what's the rationale for changing `maxLength` here?

dapplion (approved):

wemeetagain (approved):

packages/lodestar/src/network/gossip/validator.ts
twoeths: the issue itself is the queue length is not enough to handle SignedAggregateAndProof gossip topic, it's ~600 gossip messages per slot

Issue #2325, opened 2021-04-04
Investigate createFlat() function


twoeths:
**Describe the bug**

The createFlat was implemented in a generic way
```ts
export function createFlat<T = ObjectLike>(value: T): T {
  const flat = {} as Record<string, T[keyof T]>;
  for (const [k, v] of readonlyEntries(value)) {
    // eslint-disable-next-line @typescript-eslint/ban-types
    flat[k as string] = ((v as unknown) as object).valueOf() as T[keyof T];
  }
  return (flat as unknown) as T;
}
```

but I found it's much better (in term of performance) if we use this

```ts
export function createValidatorFlat(v: Validator): Validator {
  return {
    pubkey: v.pubkey.valueOf() as BLSPubkey,
    withdrawalCredentials: v.withdrawalCredentials.valueOf() as Bytes32,
    effectiveBalance: v.effectiveBalance,
    slashed: v.slashed,
    activationEligibilityEpoch: v.activationEligibilityEpoch,
    activationEpoch: v.activationEpoch,
    exitEpoch: v.exitEpoch,
    withdrawableEpoch: v.withdrawableEpoch,
  };
}
```

here's the performance test

```ts
it.only("Array[createFlat] vs Array[createValidatorFlat", () => {
    const numValidators = 114038;
    const numKeyPairs = 100;
    const secretKeys = interopSecretKeys(numKeyPairs);
    const validators = Array.from({length: numValidators}, (_, i) => {
      return config.types.phase0.Validator.createTreeBackedFromStruct ({
        pubkey: secretKeys[i % numKeyPairs].toPublicKey().toBytes(),
        withdrawalCredentials: Buffer.alloc(32, i),
        effectiveBalance: BigInt(31000000000),
        slashed: false,
        activationEligibilityEpoch: 0,
        activationEpoch: 0,
        exitEpoch: Infinity,
        withdrawableEpoch: Infinity,
      });
    });
    const flatValidators = validators.map((v) => createFlat(v));
    const flatValidators2 = validators.map((v) => createValidatorFlat(v));
    expect(flatValidators).to.be.deep.equal(flatValidators2);
    let start = Date.now();
    for (let i = 0; i < 16; i++) {
      flatValidators.map((v) => v.exitEpoch);
    }
    console.log("@@@ map using createFlat in ", Date.now() - start);
    start = Date.now();
    for (let i = 0; i < 16; i++) {
      flatValidators2.map((v) => v.exitEpoch);
    }
[...]

PR #2331, opened 2021-04-06
[DNM] Demo createFlat issue

twoeths:
**Description**
+ `createFlat` is generalized version of `createValidatorFlat`, the result is the same as in the test
+ but looping through array from `createValidatorFlat` is much faster than the one from `createFlat`

Reproduce the issue in #2325

+ `cd packages/beacon-state-transition`
+ `yarn test:perf`

@ChainSafe/eth2-0 please try :pray: !


3xtr4t3rr3str14l:
these are the results i got when running the steps outlined:

```  createFlat
@@@ map using createFlat in  138
@@@ map using createValidatorFlat in  47
    ✓ Array[createFlat] vs Array[createValidatorFlat (20485ms)
```

PR #2334, opened 2021-04-06
Prepare fast state transition functions

wemeetagain:
**Motivation**

In the process of adding a fast implementation of the altair state transition, it is useful to make as many existing functions shared across forks as possible.

**Description**

- Move `processBlockHeader`, `processEth1Data`, `processRandao` from `fast/phase0/block` to `fast/block` (changing function signatures to `allForks` types
- Split `processFinalUpdates` into spec-defined functions, move all shared functions to `fast/epoch`
- Update downstream to use new import location


3xtr4t3rr3str14l (approved):
LGTM

PR #2341, opened 2021-04-08
Precompute OR flags

dapplion:
**Motivation**

Reduce unnecessary opts in hot paths

**Description**

Pre-computes the OR of flags to not do that on every loop of getAttestationDeltas().
- With my local estimates this PR reduces the total time of `getAttestationDeltas()` by 16%
- In Prater that means about 150ms per epoch transition


dapplion:
Screenshot from 2021-04-08 21-41-08

twoeths:
this could be coordinate with #2348 . Could you give it a try @dapplion ?

dapplion:
> this could be coordinate with #2348 . Could you give it a try @dapplion ?

Let's merge both

twoeths (dismissed):

wemeetagain:
Can we hold off on this and #2348 until after we add the participation cache?

wemeetagain:
Actually, lets just merge #2348 first, then this (with this PR rebased to take advantage of #2348). And I'll base the participation cache off that.

wemeetagain (approved):

PR #2345, opened 2021-04-08
Don't use intDiv

dapplion:
**Motivation**

Benchmarking getAttestationDeltas()

**Description**

Calling `intDiv()` added 25% extra CPU time in `getAttestationDeltas()`. `Math.floor(a / b)` in almost as readable as `intDiv()`, and in hot paths it makes sense to save up on extra function invocation. I've replaced intDiv() in the places that could be hot paths only.

### DIFF
--- a/packages/beacon-state-transition/src/fast/util/epochShuffling.ts
+++ b/packages/beacon-state-transition/src/fast/util/epochShuffling.ts
@@ -77,8 +77,8 @@ export function computeEpochShuffling(
 
   const sliceCommittee = (slot: number, committeeIndex: number): ValidatorIndex[] => {
     const index = slot * committeesPerSlot + committeeIndex;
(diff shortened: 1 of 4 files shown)


twoeths:
lgtm, could you also drop some comments for it? I also found if I define constants directly in this file, it saves quite some more time, will make a PR for that.

dapplion:
> lgtm, could you also drop some comments for it? I also found if I define constants directly in this file, it saves quite some more time, will make a PR for that.

Where would you place the comment?

twoeths:
> > lgtm, could you also drop some comments for it? I also found if I define constants directly in this file, it saves quite some more time, will make a PR for that.
> 
> Where would you place the comment?

nvm, thought the change in 1/2 files

twoeths (approved):

wemeetagain (approved):

PR #2347, opened 2021-04-08
Add CachedBalanceList

wemeetagain:
**Motivation**

On the path towards fast altair state transition, it would be nice to align our phase0 fast impl to be more suitable to work with altair.
This is the first part of several to align the phase0 fast impl.

**Description**

Add a balances cache (backed by persistent-ts `MutableVector`) to the `CachedBeaconState`. This can be quickly iterated and cheaply copied, so it avoids iterating thru the tree-backed `state.balances` in `processRewardsAndPenalties`.

- Add `CachedBalanceList` (which matches `CachedValidatorList` in almost every way)
  - One difference: Updating the entire balances array can be done with `CachedBalanceList.updateAll`
- Remove `IEpochProcess.balances` in favor of `CachedBeaconState.balances` cache
- Add `unsafeUint8ArrayToTree` function to create a merkle tree from a `Uint8Array`.
  - In epoch processing, a `BigUint64Array` of new balances is created and populated with the updated balances. This typed array is converted to `Uint8Array` and is able to be converted to merkle tree without any additional serialization/copying.

This speeds up epoch transition times, mainly thru speeding up `processRewardsAndPenalties`, screenshots below:
Before: avg 1.75s
before

After: avg 1.5s
after


twoeths (dismissed):
lgtm 👍

dapplion (dismissed):

packages/beacon-state-transition/src/fast/util/unsafeUint8ArrayToTree.ts
```diff
@@ -0,0 +1,26 @@
+import {LeafNode, Node, subtreeFillToContents} from "@chainsafe/persistent-merkle-tree";
+
+/**
+ * Convert a Uint8Array to a merkle tree, using the underlying array's underlying ArrayBuffer
+ *
+ * The array MUST NOT be modified after this, or risk the merkle nodes being modified.
```
wemeetagain:
```suggestion
 * `data` MUST NOT be modified after this, or risk the merkle nodes being modified.
```

3xtr4t3rr3str14l (approved):
LGTM

Issue #2352, opened 2021-04-09
unshuffleList performance


dapplion:
unshuffleList() can represent a significant part of an epoch transition. This issue recaps some work to check it performance.

1. Comparison against Lighthouse
 - Lighthouse [swap or not bench](https://github.com/sigp/lighthouse/blob/stable/consensus/swap_or_not_shuffle/benches/benches.rs)
 - Lodestar [shuffle bench in a branch](https://github.com/ChainSafe/lodestar/blob/dapplion%2Fbenchmark-shuffle/packages/beacon-state-transition/test/perf/shuffle/shuffle.perf.ts)

| list size |         Lightouse | Lodestar
| --- | --- | --- |
| 512     | 254.04 us | 1.6034 ms (x6)
| 16384  |  6.2046 ms | 18.272 ms (x3)
| 4000000 | 1.5617 s  | 4.9690 s  (x3)

2. Cost of hashing in shuffle

Changing the hash function with an echo function, estimate the total time spent hashing which is around 80%. Our hashing function could be optimized with native bindings, with a ~1.5-2x improve. With that unshuffleList() cpu time could be reduced by 40%.

3. Profile for the record

![Screenshot from 2021-04-09 11-37-37](https://user-images.githubusercontent.com/35266934/114164063-2a76bc00-992b-11eb-8f67-57f0ed8bb756.png)
![Screenshot from 2021-04-09 11-37-53](https://user-images.githubusercontent.com/35266934/114164080-2ea2d980-992b-11eb-80b1-4b474204b57a.png)

dapplion:
With the improvements in https://github.com/ChainSafe/lodestar/pull/2353 that's as much performance as we can squeeze. There are many other areas with higher priority to focus now

dapplion:
unshuffleList has never shown up in benchmarks recently. Also thanks to the eager state transition ahead of the epoch start, shaving off some milliseconds won't make a difference here

PR #2353, opened 2021-04-09
Benchmark shuffle

dapplion:
**Motivation**

Research performance improvements for the shuffle routine. See https://github.com/ChainSafe/lodestar/issues/2352

**Description**

- Adds multiple benchmarks to understand the cost of each component
- Inline closure. It's the only optimization that consistently showed significant improvements. It also matches the implementations in Go and Rust. It reduces the total cpu time (expect spent on hashing) by 50%

**Next steps**

- Hashing takes 80% of total CPU time so increasing the performance of our hashing library would have a big impact.
- Then we could consider paralleling this function. Given the overhead cost of message passing in Javascript this is not a route we should take until the validator set is really big.

---

> Proto: Each round can be computed independently (or in groups, e.g. rounds 0-10, 10-20, etc.), and then you can merge them afterwards. If you use some webworker approach, that may be may improve things.

> Proto: The parallelizing thing works like:
> - create a list of 0...N for each worker
> - assign a round start and end to each worker. E.g. worker 2 does rounds 20...30 of the total 90 rounds.
> - every worker runs the list shuffle for their range of rounds
> - a final merge step: apply the "mapping" of each worker to the input, in the order of the rounds

> Memory-wise: Each worker basically requires as much memory as the original algo: N uint32 values.
> Plus some small constant, to cover memory of the hash function and temporary variables.

### DIFF
--- a/packages/beacon-state-transition/src/util/shuffle.ts
+++ b/packages/beacon-state-transition/src/util/shuffle.ts
@@ -64,6 +64,14 @@ Go: https://github.com/protolambda/eth2-shuffle
 All three implemented by @protolambda, but meant for public use, like the original spec version.
 */
 
+function setPositionUint32(value: number, buf: Buffer): void {
+  // Little endian, optimized version
+  buf[_SHUFFLE_H_PIVOT_VIEW_SIZE] = (value >> 0) & 0xff;
+  buf[_SHUFFLE_H_PIVOT_VIEW_SIZE + 1] = (value >> 8) & 0xff;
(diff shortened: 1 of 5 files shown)


dapplion:
> Proto: The parallelizing thing works like:
> - create a list of 0...N for each worker
> - assign a round start and end to each worker. E.g.  worker 2 does rounds 20...30 of the total 90 rounds.
> - every worker runs the list shuffle for their range of rounds
> - a final merge step: apply the "mapping" of each worker to the input, in the order of the rounds

wemeetagain (approved):

PR #2360, opened 2021-04-12
Use createValidatorFlat instead of createFlat

### LINKED ISSUE #2325: Investigate createFlat() function

twoeths:
**Motivation**

Due to #2331 , all flow looping through `state.validators` and access validator properties takes so much time (>= 3x) like `processValidatorExit`, `getEffectiveBalances` ...

**Description**

Drop the generic version `createFlat` in favor of `createValidatorFlat`.

Closes #2325

### DIFF
--- a/packages/beacon-state-transition/src/fast/util/cachedValidatorList.ts
+++ b/packages/beacon-state-transition/src/fast/util/cachedValidatorList.ts
@@ -1,8 +1,7 @@
-import {CompositeListType, List, TreeBacked} from "@chainsafe/ssz";
+import {CompositeListType, isTreeBacked, List, TreeBacked} from "@chainsafe/ssz";
 import {phase0} from "@chainsafe/lodestar-types";
 import {Tree} from "@chainsafe/persistent-merkle-tree";
 import {MutableVector} from "@chainsafe/persistent-ts";
-import {createFlat} from "./flat";
 
 /**
  * Validator registry that synchronizes changes between two underlying implementations:
@@ -13,11 +12,18 @@ export class CachedValidatorList<T extends phase0.Validator> implements List<T> 
   tree: Tree;
   type: CompositeListType<List<T>>;
   persistent: MutableVector<T>;
+  createFlat: (t: TreeBacked<T>) => T;
 
-  constructor(type: CompositeListType<List<T>>, tree: Tree, persistent: MutableVector<T>) {
+  constructor(
+    type: CompositeListType<List<T>>,
+    tree: Tree,
+    persistent: MutableVector<T>,
+    createFlat: (t: TreeBacked<T>) => T
+  ) {
     this.type = type;
     this.tree = tree;
     this.persistent = persistent;
+    this.createFlat = createFlat;
   }
 
   get length(): number {
@@ -29,7 +35,7 @@ export class CachedValidatorList<T extends phase0.Validator> implements List<T> 
   }
 
   set(index: number, value: T): void {
-    this.persistent.set(index, createFlat(value as TreeBacked<T>));
+    this.persistent.set(index, this.createFlat(value as TreeBacked<T>));
     this.type.tree_setProperty(this.tree, index, this.type.elementType.struct_convertToTree(value));
   }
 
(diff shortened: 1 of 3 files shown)


dapplion (dismissed):

packages/beacon-state-transition/src/fast/util/flat.ts
```diff
@@ -1,13 +1,16 @@
-import {ObjectLike, readonlyEntries} from "@chainsafe/ssz";
+import {BLSPubkey, Bytes32, Validator} from "@chainsafe/lodestar-types/phase0";
```
wemeetagain: Can you import `phase0` and use it?
eg: `import {..., phase0} from "@chainsafe/lodestar-types";`

packages/beacon-state-transition/src/fast/util/cachedValidatorList.ts
```diff
@@ -13,11 +12,18 @@ export class CachedValidatorList<T extends phase0.Validator> implements List<T>
...
   type: CompositeListType<List<T>>;
   persistent: MutableVector<T>;
+  createFlat: (t: TreeBacked<T>) => T;
 
-  constructor(type: CompositeListType<List<T>>, tree: Tree, persistent: MutableVector<T>) {
+  constructor(
+    type: CompositeListType<List<T>>,
+    tree: Tree,
+    persistent: MutableVector<T>,
+    createFlat: (t: TreeBacked<T>) => T
```
wemeetagain: I don't think we need this to be passed in, just use `createValidatorFlat` directly.

wemeetagain (changes requested):

wemeetagain (approved):

PR #2363, opened 2021-04-12
Simplify API REST routes registration

dapplion:
**Motivation**

Reduce boilerplate and disperse logic

**Description**

Groups the route method declaration with route url declaration.


mpetrunic (approved):
LGTM

In my experience, it seems better to assign route names in single file. It makes it a bit easier to find controller and to check whether you have that endpoint or not.

PR #2369, opened 2021-04-12
extend beacon state API to get unarchived finalized states

3xtr4t3rr3str14l:
As part of the weak subjectivity state creation effort, we need to be able to fetch finalized states (which may not be in the state archive) from the beacon state API given a weak subjectivity epoch.  This PR allows that.


packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -182,3 +185,46 @@ export function filterStateValidatorsByStatuses(
...
+  return state;
+}
+
+async function getFinalizedState(
+  db: IBeaconDb,
+  stateCache: StateContextCache,
+  forkChoice: IForkChoice,
+  slot: Slot
+): Promise<CachedBeaconState<allForks.BeaconState>> {
+  // @TODO: we should probably use SLOTS_PER_EPOCH instead of hard-coding `32`, but i'm not sure how without passing the config all the way from all the upstream calls
```
wemeetagain: Yeah, you can just add `config: IBeaconConfig` as the first param in this function.

Perhaps we can make a public `IBeaconChain.config` property, so it can be retrieved from `chain` in `resolveStateId`

packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -12,6 +12,9 @@ import {StateContextCache} from "../../../../chain/stateCache";
 import {IBeaconDb} from "../../../../db";
 import {getStateValidatorIndex} from "../../utils";
 import {StateId} from "./interface";
+import {sleep, assert} from "@chainsafe/lodestar-utils";
+import {processSlots} from "@chainsafe/lodestar-beacon-state-transition/lib/phase0/fast";
```
wemeetagain: Can you do `import {phase0} from "@chainsafe/lodestar-beacon-state-transition";`? And below: `phase0.fast.processSlots(...)`
We're trying to keep all of our phase-specific code explicitly namespaced where its used.

packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -198,3 +204,46 @@ export function filterStateValidatorsByStatuses(
...
+): Promise<CachedBeaconState<allForks.BeaconState>> {
+  assert.lte(slot, forkChoice.getFinalizedCheckpoint().epoch * config.params.SLOTS_PER_EPOCH);
+  let state = getNearestArchivedState(config, stateCache, forkChoice, slot);
+
+  // process blocks up to the requested slot
+  for await (const block of db.blockArchive.valuesStream({gt: state.slot, lte: slot})) {
+    state = fast.fastStateTransition(state, block);
+    // yield to the event loop
+    await sleep(0);
+  }
```
dapplion, line 258: @wemeetagain Should we run this through the regen to queue things?

dapplion: Also this can destroy the node by overloading it. This functionality should not be exposed by default.

wemeetagain: We don't want to use regen bc regen (and `runStateTransition`) emits events meant for the handling of unfinalized blocks. These blocks and state are completely useless to the function of the chain since its already been finalized, so we want to run this more ephemerally.
This definitely can wreck our node and would only be available behind the `debug` api.

3xtr4t3rr3str14l: i changed this PR to where this functionality is only called when using debug api

packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -28,9 +37,11 @@ export async function resolveStateId(
 }
 
 async function resolveStateIdOrNull(
+  config: IBeaconConfig,
   chain: IBeaconChain,
   db: IBeaconDb,
-  stateId: StateId
+  stateId: StateId,
+  fallbackToArchive: boolean
```
dapplion: Could you use an opts arg here `opts: {fallbackToArchive: boolean}`

dapplion (commented):
Minor comment otherwise looks good

packages/lodestar/src/api/impl/beacon/state/utils.ts
3xtr4t3rr3str14l: would it be better to do that further upstream (i.e. here: https://github.com/ChainSafe/lodestar/pull/2369/files/c15a18a7d1e3a019a5636fd3171a5818b191984d#diff-ce989f0c32e2a4da5e07b6b7d259b3b7d9eb8fa5f07329f4b86761c9aba29c50R29)?  and then just pass `fallbackToArchive` to the downstream functions?

dapplion: Both places for consistency

packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -21,14 +21,18 @@ import {ApiError, ValidationError} from "../../errors";
 import {StateId} from "./interface";
 import {sleep, assert} from "@chainsafe/lodestar-utils";
 
+type ResolveStateIdOpts = {
+  fallbackToArchive: boolean;
```
dapplion:
```ts
fallbackToArchive?: boolean;
```

packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -21,14 +21,18 @@ import {ApiError, ValidationError} from "../../errors";
...
+};
+
 export async function resolveStateId(
   config: IBeaconConfig,
   chain: IBeaconChain,
   db: IBeaconDb,
   stateId: StateId,
-  fallbackToArchive = false
+  opts: ResolveStateIdOpts = {fallbackToArchive: false}
```
dapplion: No need to give a full default value, just do
```ts
if (opts?.fallbackToArchive)
```
below

packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -198,3 +218,38 @@ export function filterStateValidatorsByStatuses(
...
+  db: IBeaconDb,
+  forkChoice: IForkChoice,
+  slot: Slot
+): Promise<CachedBeaconState<allForks.BeaconState>> {
+  assert.lte(slot, forkChoice.getFinalizedCheckpoint().epoch * config.params.SLOTS_PER_EPOCH);
+  let state = await getNearestArchivedState(config, db, slot);
+
+  // process blocks up to the requested slot
+  for await (const block of db.blockArchive.valuesStream({gt: state.slot, lte: slot})) {
+    state = fast.fastStateTransition(state, block);
```
twoeths: by default, all verification flags are `true`. I suppose we only need to regen state so probably make them all `false`?

packages/lodestar/src/api/impl/beacon/state/utils.ts
```diff
@@ -1,25 +1,38 @@
...
 import {IBeaconChain} from "../../../../chain";
 import {StateContextCache} from "../../../../chain/stateCache";
 import {IBeaconDb} from "../../../../db";
 import {getStateValidatorIndex} from "../../utils";
 import {ApiError, ValidationError} from "../../errors";
 import {StateId} from "./interface";
+import {sleep, assert} from "@chainsafe/lodestar-utils";
+
+type ResolveStateIdOpts = {
+  fallbackToArchive?: boolean;
```
wemeetagain: Can you add a comment here for what this opt is for/does?

wemeetagain (changes requested):
Minor comment, otherwise LGTM :+1:

wemeetagain (approved):

PR #2383, opened 2021-04-15
Reprocess blocks as TreeBacked

### LINKED ISSUE #2359: Improve processEth1Data
**Describe the bug**
Process eth1 data takes 300ms - 500ms per block which is too much.
Screen Shot 2021-04-12 at 08 38 23

the reason is probably we use `readOnlyValues` which returns structs and we compare that to TreeBacked Eth1Data inside block. 

**Expected behavior**

We should be able to get TreeBacked Eth1Data from `state.eth1DataVotes` and compare TreeBacked values.
[0412_sync_prater_local.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/6293288/0412_sync_prater_local.cpuprofile.zip)

twoeths:
**Motivation**

+ Reprocess blocks take time because it's a struct, we should use TreeBacked blocks to make it faster.
+ At `onBlock` event handler, skip some operations if `reprocess=true` 

**Description**

Closes #2359

### DIFF
--- a/packages/lodestar/src/chain/eventHandlers.ts
+++ b/packages/lodestar/src/chain/eventHandlers.ts
@@ -240,6 +240,7 @@ export async function onBlock(
       }),
     ]);
   }
+  // if reprocess job, don't have to reprocess block operations or pending blocks
   if (!job.reprocess) {
     await this.db.processBlockOperations(block);
     await Promise.all(
--- a/packages/lodestar/src/chain/eventHandlers.ts
+++ b/packages/lodestar/src/chain/eventHandlers.ts
@@ -240,25 +240,25 @@ export async function onBlock(
       }),
     ]);
   }
-
-  await this.db.processBlockOperations(block);
-
-  await Promise.all(
-    this.pendingBlocks.getByParent(blockRoot).map(async ({root, slot}) => {
-      const pendingBlock = await this.db.pendingBlock.get(root, slot);
-      if (pendingBlock) {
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/chain/eventHandlers.ts
```diff
@@ -240,25 +240,25 @@ export async function onBlock(
...
-          signedBlock: pendingBlock,
-          reprocess: false,
-          prefinalized: false,
-          validSignatures: false,
-          validProposerSignature: false,
-        });
-      }
-    })
-  );
+  if (!job.reprocess) {
```
dapplion, line 244: What the reason of this change? Would you mind adding a comment?

packages/lodestar/src/chain/regen/regen.ts
```diff
@@ -174,7 +174,8 @@ export class StateRegenerator implements IStateRegenerator {
     }
 
     for (const b of blocksToReplay.reverse()) {
-      const block = await this.db.block.get(b.blockRoot, b.slot);
+      const structBlock = (await this.db.block.get(b.blockRoot, b.slot))!;
+      const block = this.config.getTypes(b.slot).SignedBeaconBlock.createTreeBackedFromStruct(structBlock);
```
dapplion, line 178: :tada:

dapplion (approved):

wemeetagain (approved):

Issue #2321, opened 2021-04-02
Sync maybe stuck


twoeths:
**Describe the bug**

It shows this all the time and sync could not move forward

```
Apr-02 03:30:45.627 [SYNC]          verbose: SyncChain maybe stuck
startEpoch: 418
batches: 7
        epoch   status
        -----   ------
        418     AwaitingValidation
        420     Downloading
        422     AwaitingProcessing
        424     AwaitingProcessing
        426     AwaitingProcessing
        428     AwaitingProcessing
        430     AwaitingProcessing

Apr-02 03:32:42.001 []                 info: Syncing - ? left - 0.00 slots/s - finalized: 418 0x6aca…1bdd - head: 13440 0x8ea5…e178 - currentSlot: 68263 - peers: 26
```
It shows 418 as the finalized epoch but also "AwaitingValidation" in the batch status.

**Expected behavior**

**Steps to Reproduce**
+ Sync Prater
+ This does not always happen

twoeths:
one more log:
```
startEpoch: 162
batches: 7
        epoch   status
        -----   ------
        162     AwaitingValidation
        164     Downloading
        166     AwaitingProcessing
        168     Downloading
        170     AwaitingProcessing
        172     AwaitingProcessing
        174     AwaitingProcessing
```

twoeths:
caught some logs for it

```
	epoch 	status
	----- 	------
	174 	AwaitingValidation
	176 	Downloading
	178 	AwaitingProcessing
	180 	AwaitingProcessing
	182 	AwaitingProcessing
	184 	AwaitingProcessing
	186 	AwaitingProcessing
```
[syncMayBeStuck.txt](https://github.com/ChainSafe/lodestar/files/6323263/syncMayBeStuck.txt)

PR #2384, opened 2021-04-15
Add participation caches

wemeetagain:
**Motivation**

Towards the goal of building a fast altair state transition, we'd like to align the phase0 state transition.
Specifically here, we can align attestation participation tracking (and speed up epoch processing in the process).

**Description**

Refactors how attestations are processed in the fast state transition to be faster and more aligned with altair.

Before, we followed the spec quite directly. Processing attestations saw them added to `state.previousEpochAttestations` and `state.currentEpochAttestations` as `PendingAttestations`. Then, on epoch transitions, these in-state attestations are analyzed to determine per-validator rewards based on participation. This participation analysis was performed in `prepareEpochProcessState`.

Now, more attestation processing happens up front (similarly/exactly as in altair). Processing attestations will still include them in the state (as changing this would change the state root). But -- per-validator participation is _also_ updated at this point (rather than being deferred to epoch processing time). And on epoch transition, the on-chain attestations are not needed, since the participation has already been tallied.

To that end, several new caches are added:
- `CachedEpochParticipation` - `state.currentEpochParticipation`, `state.previousEpochParticipation` - these participation caches are designed to align with the in-state participation fields in altair. In altair, these caches will synchronize `PersistentVector` and `Tree` backings, like is done in `CachedValidatorList`. In phase0, they are just caches.
- `epochCtx.currentInclusionData`, `epochCtx.previousInclusionData` - these caches are included to track the inclusion bonus paid in epoch processing. In altair, they won't be used, since the inclusion bonus is paid during attestation processing.

Other things of note:
- `@chainsafe/persistent-ts` is updated. Now `MutableVector` has `asTransient` and `asPersistent`, which can toggle the backing vector to use immutability or not. All `MutableVector` caches are toggled to transient at the start of state transition, toggled back to persistent at the end of state transition.

TODO:
- Run more tests/add screenshots measuring the relative speedup


twoeths:
This looks great. I was able to sync for a while until I got "Sync maybe stuck" issue #2321

packages/beacon-state-transition/src/fast/util/epochContext.ts
```diff
@@ -147,6 +162,20 @@ export function rotateEpochs(
   ]);
   epochCtx.nextShuffling = computeEpochShuffling(epochCtx.config, state, indicesBounded, nextEpoch);
   epochCtx.proposers = computeProposers(epochCtx.config, state, epochCtx.currentShuffling);
+
+  const forkName = epochCtx.config.getForkName(state.slot);
+  if (forkName === "phase0") {
+    epochCtx.previousInclusionData = epochCtx.currentInclusionData;
+    epochCtx.currentInclusionData = MutableVector.from(
```
twoeths: this seems to use PersistentVector, can we use TransientVector here instead?

packages/beacon-state-transition/src/fast/util/cachedBalanceList.ts
```diff
@@ -34,7 +34,7 @@ export class CachedBalanceList implements List<Gwei> {
   }
 
   updateAll(balances: BigUint64Array): void {
-    this.persistent.vector = Vector.from(balances);
+    this.persistent.vector = PersistentVector.from(balances);
```
twoeths, line 37: I suppose we call this from epoch transition only. Can we use the new `TransientVector.from()` instead?

packages/beacon-state-transition/src/phase0/fast/epoch/processParticipationRecordUpdates.ts
```diff
@@ -1,9 +1,21 @@
...
 import {CachedBeaconState} from "../../../fast";
 
 export function processParticipationRecordUpdates(state: CachedBeaconState<phase0.BeaconState>): void {
   // rotate current/previous epoch attestations
   state.previousEpochAttestations = state.currentEpochAttestations;
   state.currentEpochAttestations = ([] as phase0.PendingAttestation[]) as List<phase0.PendingAttestation>;
+  // rotate participation caches
+  state.previousEpochParticipation.updateAllStatus(state.currentEpochParticipation.persistent.vector);
+  state.currentEpochParticipation.updateAllStatus(
+    PersistentVector.from(
```
twoeths: maybe `TransientVector.from`?

packages/beacon-state-transition/src/fast/util/cachedBalanceList.ts
wemeetagain: `PersistentVector.from` uses `TransientVector` behind the scenes :)
https://github.com/ChainSafe/persistent-ts/blob/master/src/Vector.ts#L72

dapplion (commented):
Overall looks good, but I'm not able to vouch for correctness, would need to look way more into it. @wemeetagain What are the memory implications of the extra caches?

twoeths:
I got OOM with this branch, but used to get it with master too (even with NodeJS 14). Just something to notice.

dapplion (approved):

wemeetagain:
Documenting the learnings of this PR:

This PR started by updating participation _in every block_, _in every attestation_ (like is done in Altair). This ended up being slower than our current approach despite decreasing the epoch transition time by roughly half. The reason is the increased block processing time offset the epoch processing time (and then some). So while epoch transition times were decreased from 1.5s to 600ms, block processing time increased from 5ms to 45ms.

In the interest of making progress (and continuing on Altair) the PR has been amended to simply include the participation caches for Altair, but not use them in phase 0.

PR #2385, opened 2021-04-16
Save 1 epoch transition when the node is synced

### LINKED ISSUE #2382: Investigate double epoch transition
**Describe the bug**

The prater node is synced and there are 2 adjacent epoch transition calls (1 for `getBlockSlotState` and 1 for `runEpochTransition`), I suspect it's for the same epoch:
Screen Shot 2021-04-15 at 14 52 17

And each epoch transition in the profiler is around 6s!

What I guess:
+ we receive a gossip block at start of an epoch and we have to do epoch transition to check block proposer https://github.com/ChainSafe/lodestar/blob/5ed277e2005484b0c8e77a2e2857d501d5b938c5/packages/lodestar/src/chain/validation/block.ts#L59
+ then we do regular `runStateTransition()` which process slots (and epoch) and block inside `beacon-state-transition` module https://github.com/ChainSafe/lodestar/blob/5ed277e2005484b0c8e77a2e2857d501d5b938c5/packages/lodestar/src/chain/blocks/stateTransition.ts#L124

It happens 2 times in this profiler result 
[0415_prater_server_synced.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/6316154/0415_prater_server_synced.cpuprofile.zip)

**Expected behavior**

We should be able to do exactly 1 epoch transition leveraging our checkpoint cache there.

**Steps to Reproduce**

Sync prater until the node is synced and take profiler.

twoeths:
**Motivation**

Reduce 1 epoch transition when the node is synced.
+ we receive a gossip block at start of an epoch and we have to do epoch transition to check block proposer https://github.com/ChainSafe/lodestar/blob/5ed277e2005484b0c8e77a2e2857d501d5b938c5/packages/lodestar/src/chain/validation/block.ts#L59
+ then we do regular `runStateTransition()` which process slots (and epoch) and block inside `beacon-state-transition` module https://github.com/ChainSafe/lodestar/blob/5ed277e2005484b0c8e77a2e2857d501d5b938c5/packages/lodestar/src/chain/blocks/stateTransition.ts#L124

**Description**

Closes #2382

**Tested with prater synced node**

After processing epoch to validate gossip block, we only need to do `processSlot` inside `runStateTransition`
Screen Shot 2021-04-16 at 08 31 06
[01416_server_prater_synced.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/6321922/01416_server_prater_synced.cpuprofile.zip)

### DIFF
--- a/packages/beacon-state-transition/src/fast/stateTransition.ts
+++ b/packages/beacon-state-transition/src/fast/stateTransition.ts
@@ -26,10 +26,7 @@ export function fastStateTransition(
   // process slots (including those with no blocks) since block
   switch (preFork) {
     case "phase0":
-      if (postState.slot < block.slot) {
-        // a cached checkpoint state doesn't need to process slots
-        processSlots(postState as CachedBeaconState<phase0.BeaconState>, block.slot);
-      }
+      processSlots(postState as CachedBeaconState<phase0.BeaconState>, block.slot);
       break;
     default:
       throw new Error(`Slot processing not implemented for fork ${preFork}`);
--- a/packages/beacon-state-transition/src/fast/stateTransition.ts
+++ b/packages/beacon-state-transition/src/fast/stateTransition.ts
@@ -26,7 +26,10 @@ export function fastStateTransition(
   // process slots (including those with no blocks) since block
   switch (preFork) {
     case "phase0":
-      processSlots(postState as CachedBeaconState<phase0.BeaconState>, block.slot);
+      if (postState.slot < block.slot) {
+        // a cached checkpoint state doesn't need to process slots
+        processSlots(postState as CachedBeaconState<phase0.BeaconState>, block.slot);
+      }
       break;
     default:
       throw new Error(`Slot processing not implemented for fork ${preFork}`);
--- a/packages/lodestar/src/chain/regen/regen.ts
+++ b/packages/lodestar/src/chain/regen/regen.ts
@@ -61,11 +61,16 @@ export class StateRegenerator implements IStateRegenerator {
     const parentEpoch = computeEpochAtSlot(this.config, parentBlock.slot);
     const blockEpoch = computeEpochAtSlot(this.config, block.slot);
 
+    // This may save us at least one epoch transition.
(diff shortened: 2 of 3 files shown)


packages/lodestar/src/chain/regen/regen.ts
```diff
@@ -61,11 +61,22 @@ export class StateRegenerator implements IStateRegenerator {
...
+    // if (parentEpoch < blockEpoch && !isCheckpointBlock) {
+    if (parentEpoch < blockEpoch) {
+      const checkpointState = await this.getCheckpointState({root: block.parentRoot, epoch: blockEpoch});
+      if (!isCheckpointBlock) {
+        // if block is at a checkpoint, we're sure we have the parent state inside the checkpoint state cache
+        return checkpointState;
+      } else if (isCheckpointBlock && checkpointState) {
+        // if block is not at a checkpoint, maybe we have the parent state iside the checkpoint state cache
+        // through gossip validation. Otherwise just fallback.
+        return checkpointState;
```
dapplion, line 72: If this both branches execute the same code could you re-org as
```ts
if (
  // Explain condition 1
  condition1 ||
  // Explain condition 2
  condition2
) {
  // Common code
}
```

twoeths:
with this improvement, our validator was able to publish some attestations (currently, all requests are timed out)

dapplion (dismissed):

packages/beacon-state-transition/src/fast/stateTransition.ts
```diff
@@ -26,7 +26,10 @@ export function fastStateTransition(
   // process slots (including those with no blocks) since block
   switch (preFork) {
     case "phase0":
-      processSlots(postState as CachedBeaconState<phase0.BeaconState>, block.slot);
+      if (postState.slot < block.slot) {
```
wemeetagain: This condition can be added to `processSlots` by tweaking the assertion there. eg: `assert.lte(state.slot, block.slot)`

wemeetagain (changes requested):

wemeetagain (approved):

PR #2389, opened 2021-04-18
Add api REST response time metric

### LINKED ISSUE #2307: Add metrics for API response times
It is critical for the validator client to reply "fast" (less than 1/3 of the slot) to its requests for duties, blocks, attestations and aggregates.

Add metrics for API response times to those (at least) and watch them to be less than 1/3 of the slot.

dapplion:
**Motivation**

Our beacon node - validator client interaction is problematic causing frequent timeouts on the validator client requests.

**Description**

This PR adds metrics to know which routes are slow and by how much. It adds a histogram metric labeled by operationId. If many routes are hit, the metrics can get very verbose. If so, the buckets can be reduced to one or use multiple gauges.

It also:
- Normalize naming of rest files and functions to the operationId in the eth2 api specs
- Add missing route `/states/:stateId/root`

Fixes https://github.com/ChainSafe/lodestar/issues/2307


twoeths:
this is nice to have. Due to the nature of async/await, I'm afraid the metric is not accurate if epoch transition or any heavy operations happen randomly at any api calls.

dapplion:
> this is nice to have. Due to the nature of async/await, I'm afraid the metric is not accurate if epoch transition or any heavy operations happen randomly at any api calls.

Right, but what we want to track and minimize is the total async time to fullfill a duties request for example. The validator doesn't care about the internal queue system of the beacon node, it just needs to get the responses under x seconds

wemeetagain (approved):

PR #2393, opened 2021-04-19
Improve sim tests a bit

dapplion:
- Use blst-native in multi-thread tests
- Use `getDevValidators()` everywhere to get the same validator format, which logs the index range it handles
- Log participation every slot to ease debugging
- Reduce clean-up duplication with finally {} blocks
- Improve logging and shutdown with more sleeps

### DIFF
--- a/packages/lodestar/test/e2e/sync/sync.test.ts
+++ b/packages/lodestar/test/e2e/sync/sync.test.ts
@@ -24,7 +24,6 @@ describe("sync", function () {
 
     const loggerNodeA = testLogger("Node-A", LogLevel.info);
     const loggerNodeB = testLogger("Node-B", LogLevel.info);
-    const loggerValiA = testLogger("Vali-A", LogLevel.info);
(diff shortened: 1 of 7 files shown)


wemeetagain (dismissed):

packages/lodestar/test/e2e/sync/sync.test.ts
```diff
@@ -56,11 +56,15 @@ describe("sync", function () {
       genesisTime: bn.chain.getHeadState().genesisTime,
       logger: loggerNodeB,
     });
-    const head = await bn.chain.getHeadBlock()!;
+
+    const head = await bn.chain.getHeadBlock();
+    if (!head) throw Error("First beacon node has not head block");
```
3xtr4t3rr3str14l:
```suggestion
    if (!head) throw Error("First beacon node has no head block");
```

wemeetagain (approved):

PR #2398, opened 2021-04-19
Add dependentRoot to duties API

### LINKED ISSUE #2378: Implement dependent_root on getProposerDuties
See for context https://github.com/ethereum/eth2.0-APIs/blob/ceb555f9b40ff9c2094d038e9f70a19419e5b652/apis/validator/duties/proposer.yaml#L13

> Duties should only need to be checked once per epoch,
    however a chain reorganization could occur that results in a change of duties. For full safety,
    you should monitor head events and confirm the dependent root in this response matches:

Implement both on the server and client

dapplion:
**Motivation**

Update duties API endpoints to return dependentRoot as detailed here https://github.com/ethereum/eth2.0-APIs/blob/46d2b82127cb1ffce51bbc748a7df2677fc0215a/apis/validator/duties/attester.yaml#L50

It allows to know when a re-org has changed the duties.

**Description**

This PR extends the existing API endpoints to return dependentRoot but the validator client does nothing with them yet. This will be tackled in a future PR.

Closes https://github.com/ChainSafe/lodestar/issues/2378

### DIFF
--- a/packages/beacon-state-transition/src/util/shufflingDecisionRoot.ts
+++ b/packages/beacon-state-transition/src/util/shufflingDecisionRoot.ts
@@ -8,10 +8,8 @@ import {computeEpochAtSlot, computeStartSlotAtEpoch, getCurrentEpoch, getPreviou
  * Returns the block root which decided the proposer shuffling for the current epoch. This root
  * can be used to key this proposer shuffling.
  *
- * ## Notes
- *
  * Returns `null` on the one-off scenario where the genesis block decides its own shuffling.
- * It should be set to the latest block applied to `self` or the genesis block root.
+ * It should be set to the latest block applied to this `state` or the genesis block root.
  */
 export function proposerShufflingDecisionRoot(config: IBeaconConfig, state: BeaconState): Root | null {
   const decisionSlot = proposerShufflingDecisionSlot(config, state);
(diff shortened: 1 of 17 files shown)


packages/beacon-state-transition/src/util/shufflingDecisionRoot.ts
```diff
@@ -0,0 +1,85 @@
...
+ * can be used to key the proposer shuffling for the current epoch.
+ */
+function attesterShufflingDecisionSlot(config: IBeaconConfig, state: BeaconState, requestedEpoch: Epoch): Slot {
+  const epoch = attesterShufflingDecisionEpoch(config, state, requestedEpoch);
+  const slot = computeStartSlotAtEpoch(config, epoch);
+  return Math.max(slot - 1, 0);
+}
+
+/**
+ * Returns the epoch at which the proposer shuffling was decided.
```
twoeths: should be attester instead of proposer

packages/beacon-state-transition/src/util/shufflingDecisionRoot.ts
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * Throws an error when:
+ * - `EpochTooLow` when `requestedEpoch` is more than 1 prior to `currentEpoch`.
+ * - `EpochTooHigh` when `requestedEpoch` is more than 1 after `currentEpoch`.
+ */
+function attesterShufflingDecisionEpoch(config: IBeaconConfig, state: BeaconState, requestedEpoch: Epoch): Epoch {
```
twoeths, line 72: could you link the spec reference for the below logic? also maybe we want the fast stuff in this util, i.e. `state.currentShuffling.epoch`

dapplion: This logic assumes that `state.currentShuffling.epoch` is equal to that state's epoch, which should always be the case. Using `state.currentShuffling.epoch` here would involve a bit of refactoring that I would prefer to tackle in another PR.

wemeetagain (approved):

PR #2414, opened 2021-04-23
add state root to finalized checkpoint event

### LINKED ISSUE #2413: add state root to finalized checkpoint event
as mentioned here:
https://github.com/ChainSafe/lodestar/blob/f2bd8ac442045fd582af113022e1fb4296884460/packages/lodestar/src/api/impl/events/handlers.ts#L78

3xtr4t3rr3str14l:
resolves #2413

### DIFF
--- a/packages/lodestar/src/api/impl/events/handlers.ts
+++ b/packages/lodestar/src/api/impl/events/handlers.ts
@@ -69,14 +69,13 @@ export function handleFinalizedCheckpointEvent(
   config: IBeaconConfig,
   callback: (value: BeaconEvent) => void
 ): ChainEventListener<ChainEvent.finalized> {
-  return (payload) => {
+  return (payload, state) => {
     callback({
       type: BeaconEventType.FINALIZED_CHECKPOINT,
       message: {
         block: payload.root,
         epoch: payload.epoch,
-        //TODO: add state root to finalized checkpoint event
-        state: Buffer.alloc(32, 0),
+        state: config.getTypes(state.slot).BeaconState.hashTreeRoot(state),
       },
     });
   };
--- a/packages/lodestar/src/api/impl/events/handlers.ts
+++ b/packages/lodestar/src/api/impl/events/handlers.ts
@@ -75,7 +75,7 @@ export function handleFinalizedCheckpointEvent(
       message: {
         block: payload.root,
         epoch: payload.epoch,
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/api/impl/events/handlers.ts
```diff
@@ -69,14 +69,13 @@ export function handleFinalizedCheckpointEvent(
...
-  return (payload) => {
+  return (payload, state) => {
     callback({
       type: BeaconEventType.FINALIZED_CHECKPOINT,
       message: {
         block: payload.root,
         epoch: payload.epoch,
-        //TODO: add state root to finalized checkpoint event
-        state: Buffer.alloc(32, 0),
+        state: config.getTypes(state.slot).BeaconState.hashTreeRoot(state),
```
dapplion: Are we 100% sure that this state has been hashed before? Hashing a full BeaconState is extremely expensive (in the order of 500-1000ms) so doing this just because would be a big loss in performance

3xtr4t3rr3str14l: I personally haven't seen a spot where the state has been hashed upstream (such as where the chain emits the event), but maybe there's somewhere else upstream that I could be getting the state root from that I'm not aware of.  

However, we do ultimately need the state root to be in this event handler in order to be compliant with the API spec (in the link, see the Examples dropdown and find entry for `finalized_checkpoint`):
https://ethereum.github.io/eth2.0-APIs/#/Events/eventstream

twoeths: I think a finalized state should be pre-hashed, so this is fine.
Btw, It maybe better to rename `payload` to `checkpoint`.

dapplion: > Btw, It maybe better to rename payload to checkpoint

Better to open a new issue since everything in this file uses payload

dapplion (dismissed):

packages/lodestar/src/api/impl/events/handlers.ts
wemeetagain: If `state` is a `CachedBeaconState` here, we can use `state.hashTreeRoot()`

wemeetagain (dismissed):

twoeths (approved):

wemeetagain (approved):

PR #2416, opened 2021-04-24
Refactor validator client

### LINKED ISSUE #2380: Validators miss proposing blocks at epoch transition
**Describe the bug**

What I see from the log
```
2021-03-26 11:23:54 [VALI-A]           info: Published block slot=3, validator=0xab8d…6ff9
2021-03-26 11:23:56 [VALI-A]           info: Published block slot=4, validator=0x9698…dd5c
2021-03-26 11:23:58 [VALI-A]           info: Published block slot=5, validator=0xa8d4…b8ac
2021-03-26 11:24:00 [VALI-A]           info: Published block slot=6, validator=0x9297…95d2
2021-03-26 11:24:02 [VALI-A]           info: Published block slot=7, validator=0x8128…4a5e
2021-03-26 11:24:06 [VALI-A]           info: Published block slot=9, validator=0x8f46…9be7
2021-03-26 11:24:08 [VALI-A]           info: Published block slot=10, validator=0x8105…6fe2
2021-03-26 11:24:10 [VALI-A]           info: Published block slot=11, validator=0x876d…f268
2021-03-26 11:24:12 [VALI-A]           info: Published block slot=12, validator=0xb5bb…c478
2021-03-26 11:24:14 [VALI-A]           info: Published block slot=13, validator=0x925b…4bbc
2021-03-26 11:24:16 [VALI-A]           info: Published block slot=14, validator=0xab8d…6ff9
2021-03-26 11:24:18 [VALI-A]           info: Published block slot=15, validator=0x9314…1327
2021-03-26 11:24:22 [VALI-A]           info: Published block slot=17, validator=0xb397…c006
2021-03-26 11:24:24 [VALI-A]           info: Published block slot=18, validator=0xa48c…dd6b
2021-03-26 11:24:26 [VALI-A]           info: Published block slot=19, validator=0x92a9…a249
2021-03-26 11:24:28 [VALI-A]           info: Published block slot=20, validator=0xaecc…3573
2021-03-26 11:24:30 [VALI-A]           info: Published block slot=21, validator=0x8d5d…b62e
2021-03-26 11:24:32 [VALI-A]           info: Published block slot=22, validator=0x954e…593c
2021-03-26 11:24:34 [VALI-A]           info: Published block slot=23, validator=0xaef9…72d5
```

we miss blocks 8, 16, 24

at validator block service, I found that `onClockSlot` happens before `updateDuties` (which happens at onClockEpoch)

also, is it possible to get duties 1 epoch in advance? right now we get duties at current epoch.

**Expected behavior**

Validators should propose blocks at epoch transition

**Steps to Reproduce**

`yarn test:sim:singleThread`

dapplion:
**Motivation**

Our validator client worked but was sub-optimal. This PR attempts to solve existing issues from their root, and pave the way for future optimizations

**Description**

Known fixed issues:
- Handles pre-genesis properly. Will fetch both proposer and attester duties before genesis. Is able to attest block 0 and propose block 1.
- Fix bug where the validator will consistently not propose on the first block of the epoch. Closes https://github.com/ChainSafe/lodestar/issues/2380

Improvements:
- Batch all operations possible. Create a single attestation and aggregate per committee, then sign by everyone and send once.
- Re-fetch block duties on slot to handle re-orgs safely.
- Fetch fork only once per epoch and cache the result for future epochs in case of failure.
- Fetch validator indices in parallel to attester duties. Will refetch duties if it finds new indices.
- Throttle operations so if an AttesterService slot cycle takes too long, skips the next. This prevents overloading the node continuously.


packages/lodestar/src/api/impl/beacon/blocks/index.ts
```diff
@@ -109,7 +108,6 @@ export class BeaconBlockApi implements IBeaconBlocksApi {
   }
 
   async publishBlock(signedBlock: phase0.SignedBeaconBlock): Promise<void> {
-    await checkSyncStatus(this.config, this.sync);
```
3xtr4t3rr3str14l, line 112: should we still call `notWhileSyncing` (or similar) here?  even if the assumption is that someone has already called `produceBlock` which checks sync status, someone could still theoretically call `publishBlock` and having the sync check there would still be helpful, no?

packages/lodestar/src/api/impl/beacon/pool/pool.ts
```diff
@@ -43,7 +42,6 @@ export class BeaconPoolApi implements IBeaconPoolApi {
   }
 
   async submitAttestations(attestations: phase0.Attestation[]): Promise<void> {
-    await checkSyncStatus(this.config, this.sync);
```
3xtr4t3rr3str14l, line 46: same as in `publishBlock`

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -25,14 +25,28 @@ import {ZERO_HASH} from "../../../constants";
...
 import {ApiError} from "../errors";
 import {ApiNamespace, IApiModules} from "../interface";
-import {checkSyncStatus} from "../utils";
 import {IValidatorApi} from "./interface";
 
+/**
+ * Validator clock may be advanced from beacon's clock. If the validator requests a resource in a
+ * future slot, wait some time instead of rejecting the request because it's in the future
+ */
+const MAX_API_CLOCK_DISPARITY_MS = 1000;
```
3xtr4t3rr3str14l, line 39: where does the value for this and `SYNC_TOLERANCE_EPOCHS` below come from?

packages/lodestar/test/sim/singleNodeSingleThread.test.ts
```diff
@@ -13,14 +13,15 @@ import {sleep, TimestampFormatCode} from "@chainsafe/lodestar-utils";
...
 describe("Run single node single thread interop validators (no eth1) until checkpoint", function () {
-  const timeout = 120 * 1000;
   const testParams: Pick<IBeaconParams, "SECONDS_PER_SLOT" | "SLOTS_PER_EPOCH"> = {
     SECONDS_PER_SLOT: 2,
     SLOTS_PER_EPOCH: 8,
   };
   const manyValidatorParams: Partial<IBeaconParams> = {
     ...testParams,
-    TARGET_AGGREGATORS_PER_COMMITTEE: 1,
+    // Reduce from 16 of minimal but ensure there's almost always an aggregator per committee
```
3xtr4t3rr3str14l, line 22: what does "Reduce from 16 of minimal" mean?

packages/validator/src/util/clock.ts
```diff
@@ -0,0 +1,102 @@
...
+    if (timeItem === TimeItem.Slot) {
+      return miliSecondsPerSlot - Math.abs(msFromGenesis % miliSecondsPerSlot);
+    } else {
+      const miliSecondsPerEpoch = this.config.params.SLOTS_PER_EPOCH * miliSecondsPerSlot;
+      return miliSecondsPerEpoch - Math.abs(msFromGenesis % miliSecondsPerEpoch);
+    }
+  }
+}
+
+// function useEventStream() {
```
3xtr4t3rr3str14l, line 93: why is this commented function here?  can it be removed?

3xtr4t3rr3str14l (commented):
this is a big'un!  i only reviewed the `packages/lodestar/*` half of it (except for the commented function in `packages/validator`, but will try to look at the validator half later.

packages/lodestar/src/api/impl/beacon/blocks/index.ts
dapplion: Lighthouse doesn't have this check here, only in duties. It good to not prevent actions unless there's a strong reason to do so, specially publishing a block which is a sensitive operation. I don't any immediate issue with publishing a block even if the chain is syncing.

packages/lodestar/src/api/impl/validator/validator.ts
dapplion: SYNC_TOLERANCE_EPOCHS comes from Lighthouse and this 1000 come from my own observations of our metrics, it's a bit arbitrary. It's double the gossip clock disparity to account for our overloaded node.

packages/lodestar/test/sim/singleNodeSingleThread.test.ts
dapplion: The minimal config has 16 as value

packages/validator/src/util/clock.ts
dapplion: It's sort of TODO. We should use the event stream here but in a future PR

3xtr4t3rr3str14l: can you maybe add a TODO comment and/or throw this into a new issue?

wemeetagain (approved):

packages/validator/src/util/clock.ts
wemeetagain: #2430

PR #2424, opened 2021-04-26
Raise default memory limit

dapplion:
**Motivation**

NodeJS applications have a default memory limit of 2.5GB. This limit is bit tight for a Prater node, it is recommended to raise the limit since memory may spike during certain network conditions.

We are now focusing on Altair + the Lightclient hackaton. This PR would give us some time to avoid out of memory errors during this time. In about a month we should focus again on reducing memory spikes. Our node seems to not have any memory leaks, so I believe raising the limit is safe.


twoeths (approved):

wemeetagain (approved):
:rocket:

PR #2427, opened 2021-04-26
Attestation Service

### LINKED ISSUE #1382: Lack of random subnet subscription

### LINKED ISSUE #2456: prepareBeaconCommitteeSubnet performance bottleneck
**Describe the bug**

`prepareBeaconCommitteeSubnet` takes too much potentially breaking the validator's workflow. In average it takes about 12 seconds but this snapshot it takes 4s. The cost comes from `computeSubnetForSlot()`

Screenshot from 2021-05-01 16-03-28
Screenshot from 2021-05-01 16-05-46

**Expected behavior**

`prepareBeaconCommitteeSubnet` should almost be free since that information should already be cached in the epoch context.

twoeths:
**Motivation**

+ We need to coordinate short-lived subnets and long-lived subnets subscriptions/unsubscriptions
+ We need to track known validators

**Description**

+ Remove most of logic in AttestationCollector, only handle committee attestation when we really need to (see `shouldProcessAttestation`)
+ Remove InteropSubnetsJoiningTask
+ Create new AttestationService to handle subnets and track connected/known validators there
+ Known validators are expired after 150 slots as in lighthouse implementation

Closes #1382 

Closes https://github.com/ChainSafe/lodestar/issues/2456 by removing the calls to the extremely slow `computeSubnetForSlot()`


packages/lodestar/src/network/attestation-service/service.ts
```diff
@@ -0,0 +1,384 @@
...
+   * Duplicate randomSubnets to nextForkRandomSubnets to make sure same ENR advertisement
+   * Subscribe to respective gossip topics with next fork.
+   * No need to update ENR
+   * @param nextFork
+   */
+  private subscribeToNextForkRandomSubnets(nextFork: ForkName): void {
+    const currentSlot = this.chain.clock.currentSlot;
+    const activeSubnets = this.randomSubnets.getActive(currentSlot);
+    for (const subnet of activeSubnets) {
+      this.nextForkRandomSubnets!.request({
```
dapplion: Could you find a way to not have to use the `!` operator? Maybe ensure it's defined first of change the data structure

packages/lodestar/src/network/attestation-service/service.ts
```diff
@@ -0,0 +1,384 @@
...
+  }
+
+  /**
+   * Subscribe to a committee subnet.
+   */
+  private handleSubscription(subnetId: number, slot: Slot): void {
+    // Check if the subnet currently exists as a long-lasting random subnet
+    const randomSubnetToSlot = this.randomSubnets.getToSlot(subnetId);
+    const fork = this.chain.getForkName();
+    if (randomSubnetToSlot && randomSubnetToSlot >= 0) {
```
dapplion: Check for `randomSubnetToSlot !== undefined`

packages/lodestar/src/network/attestation-service/service.ts
```diff
@@ -0,0 +1,384 @@
...
+
+  /**
+   * A known validator has not sent a subscription in a while. They are considered offline and the
+   * beacon node no longer needs to be subscribed to the allocated random subnets.
+   *
+   * We don't keep track of a specific validator to random subnet, rather the ratio of active
+   * validators to random subnets. So when a validator goes offline, we can simply remove the
+   * allocated amount of random subnets.
+   */
+  private handleKnownValidatorExpiry(currentSlot: Slot): void {
```
dapplion: You could re-org this function by doing
1. Iterate `this.knownValidators` and delete expired directly without storing them on a set
2. Then compute `targetRandomSubnetCount = this.knownValidators.size * this.config.params.RANDOM_SUBNETS_PER_VALIDATOR`
3. Then if `activeRandomSubnets < targetRandomSubnetCount` add more, if `activeRandomSubnets > targetRandomSubnetCount` remove some

dapplion: Points 2 + 3 could be a new function `rebalanceRandomSubnets` and be called too after adding a new known validator, deduplicating logic

packages/lodestar/src/network/attestation-service/service.ts
```diff
@@ -0,0 +1,384 @@
...
+  };
+
+  private transitionToNextFork(): void {
+    const fork = this.chain.getForkName();
+    assert.true(this.nextForkRandomSubnets !== undefined, "No next fork random subnets subscriptions");
+    assert.equal(
+      fork,
+      this.nextForkRandomSubnets?.getForkName(),
+      "Chain fork name is not same to next fork random subnet"
+    );
```
dapplion: If this conditions throw they will interrupt the flow of onSlot. We may not want that

twoeths: this is just like a precondition before running this function to make it easy to understand, we had a check in `onSlot` already so it should be safe.
It's ok to remove the `assert()` but I found it's easier to understand that way.

packages/lodestar/src/network/attestationService.ts
```diff
@@ -343,30 +337,38 @@ export class AttestationService implements IAttestationService {
...
+  private rebalanceRandomSubnets(): void {
+    const {RANDOM_SUBNETS_PER_VALIDATOR} = this.config.params;
+    const numValidators = this.knownValidators.size;
+    // random subnet is renewed automatically
+    let numRandomSubnets = this.randomSubnets.getAll().length;
+    // subscribe to more random subnets
+    const currentFork = this.chain.getForkName();
+    const targetRandomSubnetCount = numValidators * RANDOM_SUBNETS_PER_VALIDATOR;
+    while (targetRandomSubnetCount > numRandomSubnets && numRandomSubnets < ATTESTATION_SUBNET_COUNT) {
+      this.subscribeToRandomSubnet(currentFork);
```
dapplion: You could pass the number of new subnets to subscribe to here and won't need to iterate with a while. The logic inside `subscribeToRandomSubnet` can be re-used if subscribing to multiple subnets at once. Consider moving the `ATTESTATION_SUBNET_COUNT` check there too to compact the logic

packages/lodestar/src/network/attestationService.ts
```diff
@@ -0,0 +1,400 @@
...
+
+  // Map of random subnets and the slot until they are needed of current fork
+  private randomSubnets: SubnetMap;
+  // Map of random subnets and the slot until they are needed of next fork
+  // TODO: handle multiple hard forks?
+  private nextForkRandomSubnets: SubnetMap | undefined;
+  // Map of committee subnets and the slot until they are needed
+  private committeeSubnets: SubnetMap;
+  // subset of committeeSubnets with exact slot for aggregator
+  private subscribedCommitteeSubnets: SubnetMap;
```
wemeetagain: Might be helpful for these comments to be `/** */` comments so they show up as help text in IDE.

packages/lodestar/src/network/attestationService.ts
wemeetagain: Yeah, might be a small gain, but agree, `subscribeToRandomSubnet` could take a second `count` param. Then we only filter/select subnets / update the ENR once.

packages/lodestar/src/network/attestationService.ts
```diff
@@ -0,0 +1,400 @@
...
+    if (this.nextForkRandomSubnets) {
+      return this.nextForkRandomSubnets.getActive(this.chain.clock.currentSlot);
+    }
+    return [];
+  }
+
+  /**
+   * Called from /eth/v1/validator/beacon_committee_subscriptions api when validator is an attester.
+   */
+  validatorSubscriptions(subscriptions: phase0.BeaconCommitteeSubscription[] = []): void {
```
wemeetagain: might call this function `addValidatorSubscriptions`, `addBeaconCommitteeSubscriptions` or similar.

packages/lodestar/src/network/peers/peerManager.ts
```diff
@@ -141,11 +136,9 @@ export class PeerManager {
   }
 
   /**
-   * Request to find peers on a given subnet.
+   * Run after validator subscriptions request.
    */
-  requestAttSubnets(requestedSubnets: RequestedSubnet[]): void {
-    this.subnets.request(requestedSubnets);
-
+  onValidatorSubscriptions(): void {
```
wemeetagain: Not a huge fan of the term "validator subscriptions" as we know we will be adding different types of subscriptions (beacon committee vs sync committee) and "validator subscriptions" is/will be vague.
Suggestion to use "beacon committee subscriptions" or similar.

wemeetagain (dismissed):
Minor feedback, overall this looks really good :100:

packages/lodestar/src/network/util.ts
```diff
@@ -110,3 +112,19 @@ export function prettyPrintPeerId(peerId: PeerId): string {
 export function getAgentVersionFromPeerStore(peerId: PeerId, metadataBook: MetadataBook): string {
   return new TextDecoder().decode(metadataBook.getValue(peerId, "AgentVersion")) || "N/A";
 }
+
```
twoeths, line 115: @dapplion I moved the `getCurrentAndNextFork` over here so that it can be used for `AttestationServices`, made some small changes and unit tests for it

dapplion: Looks good! On demand from cayman config.forks are sorted in ascending order. We can use that fact to prevent having to sort the forks everytime. So passing `config` to forks instead of an array of ForkInfo

dapplion:
I've review the design to simplify how we handle fork transitions. Instead of having multiple data structures per fork this approach just does:
- 2 epochs before the fork epoch: re-subscribe all existing long lived subnet subs to the next fork (keeping the old)
- between 2 epochs of the fork epoch: any new subscription gets subscribed to both forks.
- 2 epochs after the fork epoch: un-subscribe all 64 subnets.

With this approach we satisfy the spec with much less complexity. The SubnetMaps don't have to be aware of forks and we don't need to keep timeouts around.

dapplion (approved):

PR #2433, opened 2021-04-27
Remove try catch ignore in db impl

dapplion:
Try catch and ignore the error is very dangerous pattern. Is there a reason why we do this? We are already guarding against the NotFound error. I have not tested this change in production


wemeetagain:
file conflict

wemeetagain (approved):

PR #2441, opened 2021-04-28
Cache fork digest

### LINKED ISSUE #2431: Handle Altair fork in chain

1. Form here https://github.com/ChainSafe/lodestar/blob/84a03e020d99d8a3cee681e4d5d90ff5a9407160/packages/lodestar/src/chain/chain.ts#L277

2. Chain should be able return `getForkDigest` and `getForkName` immediately using a cache.
3. Gossip should use `chain.getForkDigest()` instead of computing it itself
4. Get fork based on epoch instead of slot (spec updated `ALTAIR_FORK_EPOCH` instead of `ALTAIR_FORK_SLOT`)

twoeths:
**Motivation**

+ Currently we compute fork digest at many places, this PR caches it in `chain`
+ Gossip to use `chain` to get available fork digest there
+ Spec updated to `ALTAIR_FORK_EPOCH` instead of `ALTAIR_FORK_SLOT`

**Description**
+ Remove `computeForkNameFromForkDigest`
+ Pass `chain` to gossip instead of `genesisValidatorsRoot`

Closes #2431


packages/params/src/presets/minimal/altair.ts
```diff
@@ -14,7 +14,7 @@ export const altairJson = {
   DOMAIN_CONTRIBUTION_AND_PROOF: "0x09000000",
 
   ALTAIR_FORK_VERSION: "0x01000001",
-  ALTAIR_FORK_SLOT: "0xffffffffffffffff",
```
dapplion, line 17: Why do we change from slot to epoch? Is this how it's defined in the spec?

packages/lodestar/src/chain/chain.ts
```diff
@@ -257,38 +258,83 @@ export class BeaconChain implements IBeaconChain {
...
       finalizedEpoch: finalizedCheckpoint.epoch,
       headRoot: head.blockRoot,
       headSlot: head.slot,
     };
   }
+
+  /**
+   * Precompute current fork and next fork if it's available
+   */
+  private initForkDigestCache(): void {
```
dapplion: I think we should prevent adding methods to a class when they are not strictly necessary. You can have a standalone pure function like:
```ts
class {
  constructor() {
    this.forkDigestByForkName = computeForkDigestByForkName(config, genesisValidatorsRoot)
  }
}

function computeForkDigestByForkName() {}
```
This way it's very obvious that we will never call the equivalent of `initForkDigestCache` again. Also super easy to unit test

packages/lodestar/src/chain/chain.ts
```diff
@@ -257,38 +258,83 @@ export class BeaconChain implements IBeaconChain {
...
+    );
+    if (nextFork) {
+      this.forkDigestCache.set(
+        nextFork.name,
+        computeForkDigest(this.config, nextFork.version, this.genesisValidatorsRoot)
+      );
+    }
+  }
+
+  private getForks(): {currentFork: IForkInfo; nextFork?: IForkInfo} {
```
dapplion: getForks could be a method from config, so we don't have to manually do Object.values here

packages/params/src/presets/minimal/altair.ts
twoeths: yeah it's changed 8 days ago https://github.com/ethereum/eth2.0-specs/blob/dev/configs/mainnet/altair.yaml

dapplion:
I've tuned the solution a bit with a `ForkDigestContext` object. Using that prevents having to pass the entire chain to Gossip and ReqResp handlers. I've also simplified and strengthen the types in IBeaconConfig and IBeaconChain.

packages/lodestar/src/chain/interface.ts
```diff
@@ -54,69 +55,56 @@ export interface IAttestationJob {
 export interface IBeaconChain {
   readonly genesisTime: Number64;
   readonly genesisValidatorsRoot: Root;
+
   bls: IBlsVerifier;
-  emitter: ChainEventEmitter;
-  clock: IBeaconClock;
   forkChoice: IForkChoice;
+  clock: IBeaconClock;
+  emitter: ChainEventEmitter;
```
dapplion, line 62: Sorted methods here to be in the same order as implementation

packages/lodestar/src/network/gossip/topic.ts
```diff
@@ -52,19 +49,28 @@ export function getSubnetFromAttestationSubnetTopic(topic: string): number {
...
+export function getGossipTopic(forkDigestContext: IForkDigestContext, topic: string): GossipTopic {
   const groups = topic.match(GossipTopicRegExp);
-  if (!groups || !groups[4]) {
+  if (!groups) {
+    throw Error(`Bad gossip topic string: ${topic}`);
+  }
+
+  const forkDigestHex = groups[2] as string | undefined;
+  const type = groups[3] as GossipType | undefined;
+  const encoding = groups[4] as GossipEncoding | undefined;
```
dapplion, line 60: De-structure first at once for context. Cast to T | undefined which is the real type

packages/lodestar/src/network/metadata/utils.ts
```diff
@@ -0,0 +1,36 @@
+import {IBeaconConfig, IForkInfo} from "@chainsafe/lodestar-config";
+import {Epoch, phase0} from "@chainsafe/lodestar-types";
+import {FAR_FUTURE_EPOCH} from "../../constants";
+import {IForkDigestContext} from "../../util/forkDigestContext";
+
+export function getENRForkID(
```
dapplion, line 6: This function is specific to the Metadata class and only used here. It doesn't make sense to be attached to the BeaconChain class

packages/lodestar/src/util/forkDigestContext.ts
```diff
@@ -0,0 +1,41 @@
...
+import {ForkName, IBeaconConfig} from "@chainsafe/lodestar-config";
+import {ForkDigest, Root} from "@chainsafe/lodestar-types";
+import {mapValues} from "@chainsafe/lodestar-utils";
+import {toHexString} from "@chainsafe/ssz";
+
+type ForkDigestHex = string;
+
+export type IForkDigestContext = {
+  forkDigest2ForkName(forkDigest: ForkDigest | ForkDigestHex): ForkName | undefined;
+  forkName2ForkDigest(forkName: ForkName): ForkDigest;
```
dapplion, line 11: We can be certain that there will always be a ForkDigest available for known forks. Types reflect this.

packages/lodestar/src/network/metadata/utils.ts
twoeths: I think we put that in BeaconChain because it links to the head state while the implementation below always link to current clock slot / epoch.

The difference happens when we're at altair (current slot / epoch is at altair) and we're syncing from phase0, in that case ENRForkID inside the ENR should be respective to phase0, not altair

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -59,26 +60,26 @@ export class Eth2Gossipsub extends Gossipsub {
    */
   private statusInterval?: NodeJS.Timeout;
 
-  constructor({config, genesisValidatorsRoot, libp2p, validatorFns, logger, metrics}: IGossipsubModules) {
+  constructor(modules: IGossipsubModules) {
```
dapplion, line 63: Follow same style of other classes with a `modules` arg

packages/lodestar/test/unit/network/gossip/topic.test.ts
```diff
@@ -1,48 +1,50 @@
...
+  const topics: GossipTopic[] = [
+    {type: GossipType.beacon_block, fork: ForkName.phase0, encoding: DEFAULT_ENCODING},
+    {type: GossipType.beacon_aggregate_and_proof, fork: ForkName.phase0, encoding: DEFAULT_ENCODING},
+    {type: GossipType.beacon_attestation, fork: ForkName.phase0, subnet: 5, encoding: DEFAULT_ENCODING},
+    {type: GossipType.voluntary_exit, fork: ForkName.phase0, encoding: DEFAULT_ENCODING},
+    {type: GossipType.proposer_slashing, fork: ForkName.phase0, encoding: DEFAULT_ENCODING},
+    {type: GossipType.attester_slashing, fork: ForkName.phase0, encoding: DEFAULT_ENCODING},
+  ];
+  for (const topic of topics) {
+    it(`should round trip encode/decode gossip topic ${topic.type} ${topic.fork} ${topic.encoding}`, async () => {
```
dapplion, line 26: One unit test per topic, easier debugging

packages/lodestar/src/network/metadata/utils.ts
dapplion: RIght, good pick! Fixed

packages/lodestar/test/utils/mocks/chain/chain.ts
```diff
@@ -36,7 +35,7 @@ export class MockBeaconChain implements IBeaconChain {
   readonly genesisTime: Number64;
   readonly genesisValidatorsRoot: Root;
   readonly bls: IBlsVerifier;
-  forkChoice!: IForkChoice;
+  forkChoice: IForkChoice;
```
dapplion, line 38: Remove evil null assertion here. ForkChoice has to be defined in a mock BeaconChain

packages/config/src/index.ts
```diff
@@ -1,42 +1,35 @@
...
-          version: params.GENESIS_FORK_VERSION,
-        },
-        altair: {
-          name: ForkName.altair,
-          slot: params.ALTAIR_FORK_SLOT,
-          version: params.ALTAIR_FORK_VERSION,
-        },
-      };
+
+    forks: {altair, phase0},
```
wemeetagain: We need / I'd like a place that has the `IForkInfo` objects _in order of occurance_. (I've been using `getForkInfoRecord` for that purpose in a local branch)

Edit: Its worth calling out in a comment that the order here matters

wemeetagain (changes requested):

wemeetagain (approved):

PR #2442, opened 2021-04-29
chore: update libp2p modules

vasco-santos:
**Motivation**

Update libp2p to its newest release with better types support, removed Buffer, smaller bundle and other improvements

**Description**

N/A

N/A

Needs:

- [x] https://github.com/libp2p/js-libp2p-interfaces/pull/95
- [x] https://github.com/libp2p/js-libp2p/pull/934
- [ ] https://github.com/ChainSafe/discv5/pull/115
- [x] https://github.com/pgte/moving-average/pull/11 (using a published version on my npm namespace for now)


packages/lodestar/types/it-pipe/index.d.ts
```diff
@@ -1,120 +1,120 @@
...
-    f1: Duplex<A, R>,
-    f2: (source: AsyncIterable<A>) => B,
-    f3: (source: B) => C,
-    f4: (source: C) => D,
-    f5: (source: D) => E
-  ): E;
-
-  export default pipe;
-}
+// // Overwrite types until this issue is fixed https://github.com/alanshaw/it-pipe/issues/4
```
vasco-santos, line 1: I know this was not fixed yet, but this was the only way I could get the types to work. I think that all Buffer types need to be updated to Uint8Array , in order to have this back in. Can I get help on this?

packages/lodestar/src/network/reqresp/response/index.ts
```diff
@@ -61,9 +61,9 @@ export async function handleRequest(
           } else {
             throw new ResponseError(RpcResponseStatus.INVALID_REQUEST, (e as Error).message);
           }
-        });
+        })) as phase0.BeaconBlocksByRootRequest;
 
-        logger.debug("Resp received request", {...logCtx, requestBody} as Context);
+        logger.debug("Resp received request", {...logCtx, requestBody} as unknown as Context);
```
vasco-santos: Consequence of types commented for `it-pipe` I guess?

packages/lodestar/types/it-pipe/index.d.ts
dapplion: @wemeetagain

dapplion:
@vasco-santos Would you mind running prettier on the modified files for linting to pass?

vasco-santos:
> @vasco-santos Would you mind running prettier on the modified files for linting to pass?

just did it, also added a script for lint fix. Anyway, there are a few problems with the linter. Some of them are related to my changes and the `it-pipe` I believe: `Unsafe return of an any typed value`

dapplion:
Thanks for the PR @vasco-santos I think Cayman recycled parts of it in his bump deps PR, closing since we are fully up to date now :+1:

Issue #2388, opened 2021-04-16
Light Client Hackathon TODO


ColinSchwarz:
Project public URL https://hack.ethglobal.co/showcase/lodestar-light-client-recElTbWnTEw2ehQL

- [x] 1. Light client server
  - [x] a. Light Client Update creation https://github.com/ChainSafe/lodestar/pull/2494
  - [x] b. Sync API - (depends on 1a) (Lion) https://github.com/ChainSafe/lodestar/pull/2494
  - [x] c. Proof API - Request/response proofs API and proof serialization (Cayman, PR open that will unblock a light client) 
- [x] 2. Light client client (able to hit consumer endpoints, sync state machine)
  - [x] a. Consume update and sync (Lion, simple/naive approach https://github.com/ChainSafe/lodestar/pull/2400) "fast" with mock backed https://github.com/ChainSafe/lodestar/pull/2494
  - [x] b. Consume proofs and display partial data (Cayman) https://github.com/ChainSafe/lodestar/pull/2493
  - [x] c. test sync and proof interaction https://github.com/ChainSafe/lodestar/pull/2494
  - [x] d. Combine into a website (no style, eg https://yieldfarming.info/) https://github.com/ChainSafe/lodestar/pull/2494
- [x] 3. Implement Altair fork on beacon node
  - [x] a. State transition
  - [x] b. State transition fast (optimize)(Cayman) https://github.com/ChainSafe/lodestar/pull/2508
  - [x] c. Update validator flow (Tuyen, refresh of block + attestation creation)
    - [x] 1. Beacon node server https://github.com/ChainSafe/lodestar/pull/2514 https://github.com/ChainSafe/lodestar/pull/2521 https://github.com/ChainSafe/lodestar/pull/2522
    - [x] 2. Validator client (includes adding sync committee services and extra signatures) https://github.com/ChainSafe/lodestar/pull/2520
  - [x] d. Networking
    - [x] 1. Gossip https://github.com/ChainSafe/lodestar/pull/2479 https://github.com/ChainSafe/lodestar/pull/2519
    - [x] 2. Request/response (Lion) https://github.com/ChainSafe/lodestar/pull/2444
    - [x] 3. Discv5 https://github.com/ChainSafe/lodestar/pull/2484

- [ ] 4. Bonus: Smuch it all together
  - [x] Add sim tests for Altair with ALTAIR_FORK_EPOCH = 0
[...]

PR #2444, opened 2021-04-29
Implement Altair ReqResp

dapplion:
**Motivation**

Part of the Altair fork effort and our Light Client Hackathon https://github.com/ChainSafe/lodestar/issues/2388

**Description**

Required updates:
- Support V2 protocol for block methods.
- Update response parsing with a new chunk `<context-bytes>`.
- Update our dialProtocol call to negotiate V1 or V2 protocol variants.

Extra changes:
- Move all ReqResp only types and constants from `src/constants` to `src/network/reqResp/types`
- Update the responseType and requestType object maps to functions to support switching with two variables (Method + ForkName)


packages/lodestar/src/constants/network.ts
```diff
@@ -21,88 +17,9 @@ export const ATTESTATION_SUBNET_COUNT = 64;
  */
 export const ATTESTATION_PROPAGATION_SLOT_RANGE = 23;
 
-/**
```
dapplion, line 24: Moved all the ReqResp only constants and data to `network/reqResp`

packages/lodestar/src/network/events.ts
```diff
@@ -17,7 +17,7 @@ export enum NetworkEvent {
 export type NetworkEvents = {
   [NetworkEvent.peerConnected]: (peer: PeerId, status: phase0.Status) => void;
   [NetworkEvent.peerDisconnected]: (peer: PeerId) => void;
-  [NetworkEvent.reqRespRequest]: (method: Method, requestBody: phase0.RequestBody, peer: PeerId) => void;
+  [NetworkEvent.reqRespRequest]: (request: RequestTypedContainer, peer: PeerId) => void;
```
dapplion, line 20: Using this typed container prevents having to cast types on the event listener

packages/lodestar/src/network/reqresp/encoders/responseDecode.ts
```diff
@@ -44,12 +51,15 @@ export function responseDecode(
...
       // For multiple chunks, only the last chunk is allowed to have a non-zero error
       // code (i.e. The chunk stream is terminated once an error occurs
-      if (status !== RpcResponseStatus.SUCCESS) {
+      if (status !== RespStatus.SUCCESS) {
         const errorMessage = await readErrorMessage(bufferedSource);
         throw new ResponseError(status, errorMessage);
       }
 
-      yield await readEncodedPayload<phase0.ResponseBody>(bufferedSource, encoding, type, {isSszTree});
+      const forkName = await readForkName(forkDigestContext, bufferedSource, contextBytesType);
```
dapplion, line 59: While context-bytes has a single type / purpose readForkName will either read context bytes as forkDigest or return phase0

packages/lodestar/src/network/reqresp/encodingStrategies/sszSnappy/decode.ts
```diff
@@ -2,14 +2,14 @@ import BufferList from "bl";
...
+import {RequestOrResponseType, RequestOrResponseBody} from "../../types";
 import {SnappyFramesUncompress} from "./snappyFrames/uncompress";
-import {BufferedSource} from "../../utils/bufferedSource";
-import {RequestOrResponseType, RequestOrResponseBody} from "../../interface";
 import {maxEncodedLen} from "./utils";
 import {SszSnappyError, SszSnappyErrorCode} from "./errors";
 
 export interface ISszSnappyOptions {
-  isSszTree?: boolean;
+  deserializeToTree?: boolean;
```
dapplion, line 12: Better self-describing option name

packages/lodestar/src/network/reqresp/handlers/beaconBlocksByRoot.ts
```diff
@@ -1,18 +1,16 @@
...
 import {IBeaconChain} from "../../../chain";
 import {IBeaconDb} from "../../../db";
 
 export async function* onBeaconBlocksByRoot(
   requestBody: phase0.BeaconBlocksByRootRequest,
   chain: IBeaconChain,
   db: IBeaconDb
-): AsyncIterable<phase0.SignedBeaconBlock> {
-  const getBlock = db.block.get.bind(db.block);
-  const getFinalizedBlock = db.blockArchive.getByRoot.bind(db.blockArchive);
```
dapplion, line 11: Is there a reason why we binded the functions here instead of calling them directly?

packages/lodestar/src/network/reqresp/reqResp.ts
```diff
@@ -55,73 +67,45 @@ export class ReqResp implements IReqResp {
...
-              // TODO: Do error peer scoring here
-              // Must not throw since this is an event handler
-            }
-          }
-        );
-      }
+    for (const [method, version, encoding] of protocolsSupported) {
+      this.libp2p.handle(
+        formatProtocolId(method, version, encoding),
+        this.getRequestHandler({method, version, encoding})
```
dapplion, line 73: Move logic to a separate function `getRequestHandler()` to keep code more modular

packages/lodestar/test/unit/network/reqresp/encoders/request.test.ts
```diff
@@ -0,0 +1,73 @@
+import chai, {expect} from "chai";
```
dapplion, line 1: Merged requestEncode and requestDecode tests in a single file with common test cases

packages/lodestar/test/unit/network/reqresp/encoders/response.test.ts
```diff
@@ -0,0 +1,316 @@
+import chai from "chai";
```
dapplion, line 1: Merged responseEncode and responseDecode tests in a single file with common test cases

packages/lodestar/src/network/reqresp/reqResp.ts
```diff
@@ -55,73 +67,45 @@ export class ReqResp implements IReqResp {
...
-              );
-              // TODO: Do success peer scoring here
-            } catch {
-              // TODO: Do error peer scoring here
-              // Must not throw since this is an event handler
-            }
-          }
-        );
-      }
+    for (const [method, version, encoding] of protocolsSupported) {
```
twoeths, line 70: it seems we handle v1 and v2 all the times, do we manage to address this spec somewhere? "`The v1 method MAY be unregistered at the fork boundary. In the event of a request on v1 for an Altair specific payload, the responder MUST return the InvalidRequest response code.`"

twoeths: when we start the node after `ALTAIR_FORK_EPOCH`, we don't need to handle v1

dapplion: > it seems we handle v1 and v2 all the times, do we manage to address this spec somewhere? "The v1 method MAY be unregistered at the fork boundary. In the event of a request on v1 for an Altair specific payload, the responder MUST return the InvalidRequest response code."

SSZ decoding will fail because phase0 is assumed, so it will reject with an InvalidRequest error

dapplion: > when we start the node after ALTAIR_FORK_EPOCH, we don't need to handle v1

Can I handle that on an upcoming PR? We have to review our fork transition code throughly and in the case you note we also want the node to stop listening to v1 when it crosses the fork transition

wemeetagain (approved):

PR #2445, opened 2021-04-29
Add state transition function metrics

### LINKED ISSUE #2426: Add metric to time epoch transition
Add a histogram metric to time epoch transitions

dapplion:
**Motivation**

We need to know how long a epoch transition and block processing takes. This is arguably one of the most important metrics for us that we are currently trying to minimize, so we need to measure it.

**Description**

Closes https://github.com/ChainSafe/lodestar/issues/2426

The consumer code of the beacon state transition function is a bit messy so I've re-arranged it a bit so it's clearer what functions are local and which are exported


wemeetagain (dismissed):
:100:

twoeths (approved):
👍

PR #2451, opened 2021-04-30
fetch ws state from IPFS via cli

3xtr4t3rr3str14l:
Changes our cli to have it load from a weak subjectivity state by default (if a state is not already in the db)
Also adds a state root file of the corresponding ws state to verify against
As it stands right now with this scheme, we will have to update the IPFS link and the state root file manually for every release.

The current ws state is stored here:
https://ipfs.io/ipfs/QmVsGxwZs4acTDfhQrNYVqccEVxyhALqJuYqgJrbbNAd1Z

This is all for the Prater testnet.  I can generate one for mainnet as well once this PR and the parallel PR for beacon-state-upload (https://github.com/ChainSafe/beacon-state-upload/pull/3) are merged to their respective master branches.

**NOTE: DO NOT MERGE:** added the DO NOT MERGE label because we need both this PR and https://github.com/ChainSafe/beacon-state-upload/pull/3 to be approved before merging, but also because the downloading of the state from IPFS isn't always stable during my testing, so I want to investigate this further.


packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -44,6 +44,15 @@ export async function initBeaconState(
     return await initFromFile(args.weakSubjectivityStateFile);
   } else if (dbHasSomeState) {
     return await initStateFromDb(config, db, logger);
+  } else if (args.fetchWeakSubjectivityStateFromIPFS) {
+    const state = await initFromFile("https://ipfs.io/ipfs/QmVsGxwZs4acTDfhQrNYVqccEVxyhALqJuYqgJrbbNAd1Z/state.ssz");
+    const storedStateRoot = await downloadOrLoadFile("./ws-state-root.ssz");
```
wemeetagain: Can we pull out the ipfs path and state root into a separate file, eg: weakSubjectivityState.json.
Something that ends up looking like:
```json
{
  "ipfsPath": "/ipfs/QmVsGxwZs4acTDfhQrNYVqccEVxyhALqJuYqgJrbbNAd1Z/state.ssz",
  "stateRoot": "0xabcd..."
}
```

I think this will be easier to manage.

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -44,6 +44,15 @@ export async function initBeaconState(
...
+    state;
+    // verify downloaded state against locally stored state root
+    if (!config.types.Root.equals(state.hashTreeRoot(), storedStateRoot)) {
+      throw new Error("Unable to verify state root downloaded from ");
```
wemeetagain: maybe missing "IPFS" in this error message?

wemeetagain (changes requested):

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -44,6 +44,14 @@ export async function initBeaconState(
     return await initFromFile(args.weakSubjectivityStateFile);
   } else if (dbHasSomeState) {
     return await initStateFromDb(config, db, logger);
+  } else if (args.fetchWeakSubjectivityStateFromIPFS) {
+    const stateData = readFile("./weakSubjectivityState.json") as {ipfsPath: string; stateRoot: string};
```
dapplion: Why would this file `./weakSubjectivityState.json` exist?

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -44,6 +44,14 @@ export async function initBeaconState(
...
+    const state = await initFromFile("https://ipfs.io" + stateData.ipfsPath);
```
dapplion: The IPFS gateway url must be a customizable parameter

packages/cli/src/cmds/beacon/initBeaconState.ts
3xtr4t3rr3str14l: as it stands right now, it would be because we copied it from the output of beacon-state-upload (from here: https://github.com/ChainSafe/beacon-state-upload/pull/3/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R39)
perhaps we can have a more automated process, but that's the way it is right now (or at least inasmuch as this PR and that beacon-state-upload PR are "right now")

3xtr4t3rr3str14l:
added the DO NOT MERGE label because we need both this PR and https://github.com/ChainSafe/beacon-state-upload/pull/3 to be approved before merging, but also because the downloading of the state from IPFS isn't always stable during my testing, so I want to investigate this further.

packages/cli/src/cmds/beacon/initBeaconState.ts
dapplion: I still don't follow why the file should exist in any environment that you may run the cli.

3xtr4t3rr3str14l: Two reasons
1) so we can have the state root of the weak subjectivity state to check against when downloading from IPFS so as to not rely purely on the security of the SSL certs (as discussed on one of the eth2 implementer's calls about Teku's state loading process from Infura)
2) so we can have the IPFS link from which to download the weak subjectivity state

Maybe it's better to just embed directly into the code though rather than rely on a file.  We'll just have to remember to update the code for every release that we have a new weak subjectivity state, but that's just as much of a process as remembering to copy the file anyway.

dapplion: But again if we add the file at the project root `/weakSubjectivityState.json`, and you run the command from the root it will work. If you run the command from `/packages/cli` it will break since `./weakSubjectivityState.json` does not exist

3xtr4t3rr3str14l: ahh ok i gotcha.  yeah i'll move it into `packages/lodestar/src/cmds/...`

3xtr4t3rr3str14l:
> added the DO NOT MERGE label because we need both this PR and [ChainSafe/beacon-state-upload#3](https://github.com/ChainSafe/beacon-state-upload/pull/3) to be approved before merging, but also because the downloading of the state from IPFS isn't always stable during my testing, so I want to investigate this further.

To follow up to this comment, I have been running the beacon state uploader for the past few days with no issues on mainnet.  Currently setting one up on contabo-3 to run alongside our prater node as well, but I expect similar results.  Thus, I feel that this PR is working well enough to where I don't need to have the DO NOT MERGE label anymore.

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -16,6 +16,32 @@ import {downloadOrLoadFile} from "../../util";
...
+  logger: ILogger
+): Promise<TreeBacked<allForks.BeaconState>> {
+  logger.info("Fetching weak subjectivity state from IPFS at " + args.ipfsGatewayUrl + wsStateData.ipfsPath);
+  const state = await initFunc(args.ipfsGatewayUrl + wsStateData.ipfsPath);
+  if (!state) {
+    throw new Error("Weak subjectivity state not found for network " + args.network);
+  }
+  // verify downloaded state against locally stored state root
+  if (toHexString(state.hashTreeRoot()) !== wsStateData.stateRoot) {
+    throw new Error("Unable to verify state root downloaded from IPFS");
```
dapplion, line 41: You could verify it but it is invalid. You should print the expect root and the actual root

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -44,6 +70,16 @@ export async function initBeaconState(
...
+  } else if (args.fetchWeakSubjectivityStateFromIPFS) {
+    if (args.network === "mainnet") {
+      return await initAndVerifyWeakSujectivityState(args, mainnetWeakSubjectivityState, initFromFile, logger);
+    } else if (args.network === "prater") {
+      return await initAndVerifyWeakSujectivityState(args, praterWeakSubjectivityState, initFromFile, logger);
+    } else if (args.network === "pyrmont") {
+      return await initAndVerifyWeakSujectivityState(args, pyrmontWeakSubjectivityState, initFromFile, logger);
+    } else {
+      throw new Error("No matching network with weak subjectivity state.");
+    }
```
dapplion, line 82: Can be simplified by getting the `weakSubjectivityState` and then calling 
```ts
return await initAndVerifyWeakSujectivityState(args, weakSubjectivityState, initFromFile, logger);
```
once

packages/cli/src/cmds/weakSubjectivityState.ts
```diff
@@ -0,0 +1,16 @@
...
+export const praterWeakSubjectivityState = {
+  stateRoot: "0x9a4e6155e7341303e595529cd27529b4a61b7be6a38b7f437c68dc2dd9baf3da",
+  ipfsPath: "/ipfs/QmXM4yuVAZxd8jm2st6nfSy8gijnun4owePPDp3kqtHbyd/state.ssz",
+};
+
+// manually copied from the weakSubjectivityState.json ouput in beacon-state-upload
+export const pyrmontWeakSubjectivityState = {
+  stateRoot: "0xdab52683fc625aa156ca1861ddcabe642e56ffc3c0d6f04a11a9b721b92eb2d9",
+  ipfsPath: "/ipfs/QmPinARkXH7CnCq4A3fFwL6Gf46rubzbXZn1ggHdHULqRr/state.ssz",
+};
```
dapplion, line 16: How would we update this values every X time?

dapplion: Who is pinning those IPFS paths and with what replication factor?

3xtr4t3rr3str14l:
@dapplion just pulled a draft PR for a new version of the weak subjectivity CLI here:
https://github.com/ChainSafe/lodestar/pull/2571

Keeping this one (#2451) open for now so we can still have both solutions available to review, if needed.

packages/cli/src/cmds/weakSubjectivityState.ts
dapplion: @3xtr4t3rr3str14l bump. This is the most critical question to answer around this PR

3xtr4t3rr3str14l: Under this design, the pinning would be done by https://github.com/ChainSafe/beacon-state-upload/pull/3 every time a new weak subjectivity period is fetched via our API.  `beacon-state-upload` would be subscribed to the finalized checkpoints events stream via the API, and for every new finalized checkpoint, we would check to see if the WS API (via `lodestar/ws_epoch` custom API endpoint) gives us a new WS checkpoint.  If we have a new WS checkpoint, then we get the WS state from the API, and upload the WS checkpoint data and the WS state to IPFS & IPNS (which pins the data automatically via the IPFS HTTP API).  But under this setup, we would only use the IPFS link in this embedded file, for stability.

However, none of this may be relevant if we go with the newer solution in #2571 .  I'm also open to suggestions on a fusion of the two strategies though, if that is desired.

wemeetagain:
closing in favor of #2571 which is nearing completion

Issue #2473, opened 2021-05-03
Add validator monitor metrics


dapplion:
Reviewed this file and select interesting metrics to add
https://github.com/sigp/lighthouse/blob/stable/beacon_node/beacon_chain/src/metrics.rs

**Validator monitoring (own validators)**

High priority to add these. All metrics are indexed by validator (probably index)

Validator Monitor Metrics (per-epoch summaries)

- [ ] `validator_monitor_prev_epoch_on_chain_attester_hit`: Incremented if the validator is flagged as a previous epoch attester during per epoch processing
- [ ] `validator_monitor_prev_epoch_on_chain_attester_miss`: Incremented if the validator is not flagged as a previous epoch attester during per epoch processing
- [ ] `validator_monitor_prev_epoch_on_chain_head_attester_hit`: Incremented if the validator is flagged as a previous epoch head attester during per epoch processing
- [ ] `validator_monitor_prev_epoch_on_chain_head_attester_miss`: Incremented if the validator is not flagged as a previous epoch head attester during per epoch processing
- [ ] `validator_monitor_prev_epoch_on_chain_target_attester_hit`: Incremented if the validator is flagged as a previous epoch target attester during per epoch processing
- [ ] `validator_monitor_prev_epoch_on_chain_target_attester_miss`: Incremented if the validator is not flagged as a previous epoch target attester during per epoch processing
- [ ] `validator_monitor_prev_epoch_on_chain_inclusion_distance`: The attestation inclusion distance calculated during per epoch processing
- [ ] `validator_monitor_prev_epoch_attestations_total`: The number of unagg. attestations seen in the previous epoch.
- [ ] `validator_monitor_prev_epoch_attestations_min_delay_seconds`: The min delay between when the validator should send the attestation and when it was received.
- [ ] `validator_monitor_prev_epoch_attestation_aggregate_inclusions`: The count of times an attestation was seen inside an aggregate.
- [ ] `validator_monitor_prev_epoch_attestation_block_inclusions`: The count of times an attestation was seen inside a block.
- [ ] `validator_monitor_prev_epoch_attestation_block_min_inclusion_distance`: The minimum inclusion distance observed for the inclusion of an attestation in a block.
- [ ] `validator_monitor_prev_epoch_beacon_blocks_total`: The number of beacon_blocks seen in the previous epoch.
- [ ] `validator_monitor_prev_epoch_beacon_blocks_min_delay_seconds`: The min delay between when the validator should send the block and when it was received.
[...]

PR #2474, opened 2021-05-03
Add local validator monitor

dapplion:
**Motivation**

Automatically track the validators using a beacon node. This is critical to debug and understand how to improve the performance of the Lodestar validator.

**Description**

Implements the metrics described in https://github.com/ChainSafe/lodestar/issues/2473 which are compatible with Lighthouse's validator monitoring. Consider adding their Grafana dashboard when this PR is merged.

@wemeetagain I need help to figure out how to connect this validator monitoring to the head so it has sure access to the precomputed cache-s


wemeetagain:
> I need help to figure out how to connect this validator monitoring to the head so it has sure access to the precomputed cache-s

Are you talking about how to feed data on epoch transitions to `processValidatorStatuses`?

dapplion:
> > I need help to figure out how to connect this validator monitoring to the head so it has sure access to the precomputed cache-s
> 
> Are you talking about how to feed data on epoch transitions to `processValidatorStatuses`?

Yes

dapplion:
Brainstormed offline with @wemeetagain and settled for:

- Add a property to the metrics object sent to the state transition functions
```ts
interface IMetrics {
  registerParticipation(epochProcess): void
}
```
- Call this after processEpoch and make that function return something that you can efficiently extract participation data from.
https://github.com/ChainSafe/lodestar/blob/13bb12031c31a2fe2d1af9b5ff9c7436e914e5e7/packages/beacon-state-transition/src/allForks/stateTransition.ts#L145

- For altair the relevant data is in the CachedBeaconState right before rotating epochs here
https://github.com/ChainSafe/lodestar/blob/13bb12031c31a2fe2d1af9b5ff9c7436e914e5e7/packages/beacon-state-transition/src/altair/epoch/processParticipationFlagUpdates.ts#L7
- ParticipationFlags are available in object form as here:
https://github.com/ChainSafe/lodestar/blob/13bb12031c31a2fe2d1af9b5ff9c7436e914e5e7/packages/beacon-state-transition/src/allForks/util/cachedEpochParticipation.ts#L41

TODO:
- How to deal with phase0 + altair in the validator monitor? The rewards calculation changes significantly and we want our code to work both ways and Prometheus + Grafana don't understand forking.

dapplion:
There's an issue with Altair because now inclusion is not tracked.

@mpetrunic Suggests to just loop over the attestations of a block afterwards and compute the attestation inclusion only for the validators of interest afterwards.

packages/lodestar/src/chain/eventHandlers.ts
```diff
@@ -259,7 +261,17 @@ export async function onBlock(
         return this.attestationProcessor.processAttestationJob(job);
       }),
     ]);
+
+    if (this.metrics) {
+      // Register attestations in metrics
+      for (const attestation of indexedAttestations) {
+        if (attestation) {
+          this.metrics.registerAttestationInBlock(attestation, block.message);
```
wemeetagain, line 269: I think we will likely want to register attestations while processing attestations in block processing (since this will save us one iteration thru tree-backed attestations).

This should be fine for now tho. We can benchmark and move from there.

wemeetagain (approved):

Issue #2412, opened 2021-04-23
Altair: Sync Committee API


twoeths:
**Is your feature request related to a problem? Please describe.**

Part of 3c in #2388

**Describe the solution you'd like**

Start with a proposal as mentioned at https://discord.com/channels/595666850260713488/710831784991916072/832015975879540736

HackMD link: https://hackmd.io/@QYHAVYiHRg65pdI_CEm7Eg/syncommitteeapi

twoeths:
duplicate to #2512

Issue #2415, opened 2021-04-24
Pre-aggregate attestations in Aggregation pool


dapplion:
In this PR https://github.com/ChainSafe/lodestar/pull/2295/files we introduced an index to efficiently check which attestations have the same data root.

However checking Lighthouse Aggregation Pool https://github.com/sigp/lighthouse/blob/stable/beacon_node/beacon_chain/src/naive_aggregation_pool.rs we could just pre-aggregate attestations and keep a single attestation per data root.

This would substitute the need for the index in the first place and reduce the amount of data we have to keep around.

twoeths:
this looks great 👍 . My understanding is we don't even need the `attestation` repo with the new index by attestation data root?

We may want to do the same thing for altair `SyncCommitteeSignature`.

PR #2479, opened 2021-05-04
Implement Altair networking gossip spec

twoeths:
**Motivation**

Part of #2388

**Description**

Implement Altair networking gossip spec at https://github.com/ethereum/eth2.0-specs/blob/v1.1.0-alpha.3/specs/altair/p2p-interface.md:
+ New `sync_committee_contribution_and_proof` and `sync_committee_{subnet_id}` topic
+ Add respective modification for phase0 gossip and unit tests

This PR does not cover:
+ Handle pending ContributionAndProof due to FUTURE_SLOT, UNKOWN_BEACON_BLOCK_ROOT error, created #2480
+ Aggregate SyncCommitteeSignature into ContributionAndProof, should do that in #2412


packages/lodestar/src/chain/validation/contributionAndProof.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  const headState = chain.getHeadState();
+  const aggregatorPubkey = headState.validators[contributionAndProof.aggregatorIndex].pubkey;
+  if (
+    !getSyncSubCommitteePubkeys(
+      config,
+      headState as CachedBeaconState<altair.BeaconState>,
+      contribution.subCommitteeIndex
+    ).includes(aggregatorPubkey)
```
dapplion: Is this approach efficient enough?
- What's the cost of the getSyncSubCommitteePubkeys() utility? We may have to run it 1000 times per slot if we have enough validators
- Note that doing an .includes requires checking the entire array and is O(n)

packages/lodestar/src/chain/validation/syncCommittee.ts
```diff
@@ -0,0 +1,94 @@
...
+  if (!chain.forkChoice.hasBlock(syncCommittee.beaconBlockRoot)) {
+    throw new SyncCommitteeError({
+      code: SyncCommitteeErrorCode.UNKNOWN_BEACON_BLOCK_ROOT,
+      beaconBlockRoot: syncCommittee.beaconBlockRoot as Uint8Array,
+      job: syncCommitteeJob,
+    });
+  }
+
+  // There has been no other valid sync committee signature with same slot and validatorIndex
+  if (db.seenSyncCommiteeCache.hasSyncCommitteeSignature(syncCommittee)) {
```
dapplion: If we already have a validated sync committee object that includes this one, then this one is also valid right? should we then return early and accept it instead of throwing?

packages/lodestar/src/chain/validation/syncCommittee.ts
```diff
@@ -0,0 +1,94 @@
...
+  // ) {
+  //   throw new SyncCommitteeError({
+  //     code: SyncCommitteeErrorCode.VALIDATOR_NOT_IN_SYNC_COMMITTEE,
+  //     validatorIndex: syncCommittee.validatorIndex,
+  //     job: syncCommitteeJob,
+  //   });
+  // }
+
+  // validate subnet
+  const expectedSubnets = computeSubnetsForSyncCommittee(
```
dapplion: Similar to above, what the cost of this utility `computeSubnetsForSyncCommittee`? Ensure it's efficient by maybe adding some performance tests to ensure we don't have the same issues as with prepareBeaconCommitteeSubnet

dapplion:
I think we can optimize what we store:

Merge the db.contributionAndProof and sync committee seen cache into a single only-memory cache. contributionAndProof are useless after 1 block, so there's no need to persist them in the db.

Once you have verified the ContributionAndProof you can throw away the proof. You want to turn ContributionAndProof into SyncCommitteeContribution on the fly and only persist SyncCommitteeContribution.
As you verify them you pre-aggregate all SyncCommitteeContribution that sign the same data and keep merging the bits
Only when you have to create a block you will go into the SyncCommitteeContribution memory cache and turn it into a SyncAggregate. I would limit the max slots SyncCommitteeContribution are kept around to a strictly low number like Lighthouse: for example 3 slots.

To verify this condition efficiently you can add an extra index in that memory cache.
`[IGNORE] There has been no other valid sync committee signature for the declared slot for the validator referenced by sync_committee_signature.validator_index`

packages/lodestar/src/chain/validation/contributionAndProof.ts
twoeths: + `getSyncSubCommitteePubkeys` utility just extract pubkeys from a bigger `pubkeys` array inside `state.currentSyncCommittee` so it's not expensive.
But we should cache it as it's only changed per ~27h, I believe we'll create a cache for it in `CachedBeaconState`. I'd mark a TODO for this and wait for @wemeetagain fast state transition work for altair
+ once we have the cache, we just do `subCommitteeIndices.includes(aggregatorIndex)` so `.includes` is not a concern here

packages/lodestar/src/chain/validation/syncCommittee.ts
twoeths: we throw here and the gossip validation would cache it then it'll throw `IGNORE` error, the spec said to throw `IGNORE` error in this case instead of an accept

packages/lodestar/src/chain/validation/syncCommittee.ts
twoeths: I'll add: `// TODO: cache syncCommitteeIndices in CachedBeaconState`

packages/beacon-state-transition/src/fast/signatureSets/syncCommittee.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+  const syncCommittee = (state as CachedBeaconState<altair.BeaconState>).currentSyncCommittee;
+  const blsPubkeys = indicesInSyncCommittee.map((i) => syncCommittee.pubkeys[i]);
+  const currentEpoch = computeEpochAtSlot(config, state.slot);
+  const domain = getDomain(config, state, config.params.DOMAIN_SYNC_COMMITTEE, currentEpoch);
+  return {
+    type: SignatureSetType.aggregate,
+    // TODO: should index this somewhere?
+    pubkeys: blsPubkeys.map((blsPubkey) =>
+      bls.PublicKey.fromBytes(blsPubkey.valueOf() as Uint8Array, CoordType.jacobian)
```
dapplion: > TODO: should index this somewhere?

Yeah! Deserializing a pubkey from crompressed form is relatively expensive. It costs 30524.00 ns/op so per 1024 pubkeys it would be 31 ms

packages/lodestar/src/db/seenSyncCommitteeCache.ts
```diff
@@ -0,0 +1,164 @@
...
+        bls.Signature.fromBytes(preContribution.signature.valueOf() as Uint8Array),
+        bls.Signature.fromBytes(syncCommitteeSignature.signature.valueOf() as Uint8Array),
+      ];
+      // accumulate the aggregation, same slot and beaconBlockRoot here
+      const newContribution = {
+        slot,
+        beaconBlockRoot,
+        subCommitteeIndex,
+        aggregationBits,
+        signature: bls.Signature.aggregate(signatures).toBytes(),
```
dapplion: Serializing and de-serializing has some cost. This won't be done as frequently as the pubkeys issue noted above but we should keep the signatures deserialized in this cache and then serialize when extracting the contribution from the cache

packages/lodestar/src/chain/validation/contributionAndProof.ts
```diff
@@ -0,0 +1,107 @@
...
+  if (chain.clock.currentSlot > contribution.slot) {
+    throw new SyncCommitteeError({
+      code: SyncCommitteeErrorCode.PAST_SLOT,
+      currentSlot: chain.clock.currentSlot,
+      syncCommitteeSlot: contribution.slot,
+      job: contributionAndProofJob,
+    });
+  }
+
+  // The block being signed over has been seen
```
dapplion: has **not** been

packages/lodestar/src/chain/validation/syncCommittee.ts
```diff
@@ -0,0 +1,107 @@
...
+  if (chain.clock.currentSlot > syncCommittee.slot) {
+    throw new SyncCommitteeError({
+      code: SyncCommitteeErrorCode.PAST_SLOT,
+      currentSlot: chain.clock.currentSlot,
+      syncCommitteeSlot: syncCommittee.slot,
+      job: syncCommitteeJob,
+    });
+  }
+
+  // The block being signed over has been seen
```
dapplion: has **not** been

packages/lodestar/src/chain/validation/syncCommittee.ts
```diff
@@ -0,0 +1,107 @@
...
+      beaconBlockRoot: syncCommittee.beaconBlockRoot as Uint8Array,
+      job: syncCommitteeJob,
+    });
+  }
+
+  // There has been no other valid sync committee signature with same slot and validatorIndex
+  if (db.seenSyncCommiteeCache.hasSyncCommitteeSignature(subnet, syncCommittee)) {
+    throw new SyncCommitteeError({
+      code: SyncCommitteeErrorCode.SYNC_COMMITTEE_ALREADY_KNOWN,
+      root: config.types.altair.SyncCommitteeSignature.hashTreeRoot(syncCommittee),
```
dapplion: Is this root really necessary here? If no-one uses that let's not spend time doing unnecessary hashing.

packages/lodestar/src/chain/validation/syncCommittee.ts
```diff
@@ -0,0 +1,107 @@
...
+      job: syncCommitteeJob,
+    });
+  }
+
+  const headState = chain.getHeadState();
```
dapplion: Note the conversation about forkChoice efficiency. Calling getHead() here potentially means 1000 times per slot

packages/lodestar/src/db/seenSyncCommitteeCache.ts
```diff
@@ -0,0 +1,164 @@
...
+      };
+      slotRootCache.set(key, newContribution);
+    } else {
+      const aggregationBits = Array.from(
+        {length: intDiv(this.config.params.SYNC_COMMITTEE_SIZE, altair.SYNC_COMMITTEE_SUBNET_COUNT)},
+        (_, i) => indicesInSubSyncCommittee.includes(i)
```
dapplion: Note: This has quadratic time

packages/lodestar/src/chain/validation/syncCommittee.ts
```diff
@@ -0,0 +1,107 @@
...
+    const signatureSet = fast.getSyncCommitteeSignatureSet(headState, syncCommittee);
+    if (!(await chain.bls.verifySignatureSets([signatureSet]))) {
+      throw new SyncCommitteeError({
+        code: SyncCommitteeErrorCode.INVALID_SIGNATURE,
+        job: syncCommitteeJob,
+      });
+    }
+  }
+
+  const indicesInSubSyncCommittee = getIndicesInSubSyncCommittee(
```
dapplion: A bit worried about performance here, we should really work on the SyncCommittee cache as soon as we can

packages/lodestar/src/db/seenSyncCommitteeCache.ts
```diff
@@ -0,0 +1,164 @@
...
+   * Each array item is respective to a subCommitteeIndex.
+   * Preaggregate into SyncCommitteeContribution.
+   * */
+  private contributionCaches: Map<phase0.Slot, Map<SlotRootKey, altair.SyncCommitteeContribution>>[];
+
+  /**
+   * Each array item is respective to a subCommitteeIndex.
+   * A seen SyncCommitteeSignature is decided by slot + validator index.
+   */
+  private seenCaches: Map<phase0.Slot, Set<phase0.ValidatorIndex>>[];
```
dapplion: If I understand, here we need to know if we have seen a signature for a combination of slot + validatorIndex + subnet. Would it make things simpler just to have a single 
```ts
type SubnetValidatorIndex = string; // `${subnet}-${validatorIndex}`
Map<phase0.Slot, Set<SubnetValidatorIndex>>
```

packages/lodestar/src/db/seenSyncCommitteeContributionCache.ts
```diff
@@ -0,0 +1,90 @@
...
+ * Cache SyncCommitteeContribution and seen ContributionAndProof.
+ * This is used for SignedContributionAndProof validation and block factory.
+ * This stays in-memory and should be pruned per slot.
+ */
+export class SeenSyncCommitteeContributionCache {
+  private contributionsCache: Map<phase0.Slot, ContriButionsBySlotRoot> = new Map<
+    phase0.Slot,
+    ContriButionsBySlotRoot
+  >();
+  private seenCache: Map<phase0.Slot, Map<SeenCacheKey, boolean>> = new Map<phase0.Slot, Map<SeenCacheKey, boolean>>();
```
dapplion: No need to type if you already give a typed initial value, this should work:
```ts
private seenCache = new Map<phase0.Slot, Map<SeenCacheKey, boolean>>();
```

packages/lodestar/src/db/seenSyncCommitteeContributionCache.ts
```diff
@@ -0,0 +1,90 @@
...
+
+type ContriButionsBySlotRoot = Map<SlotRootKey, altair.SyncCommitteeContribution[]>;
+type SeenCacheKey = string;
+
+/**
+ * Cache SyncCommitteeContribution and seen ContributionAndProof.
+ * This is used for SignedContributionAndProof validation and block factory.
+ * This stays in-memory and should be pruned per slot.
+ */
+export class SeenSyncCommitteeContributionCache {
```
dapplion: Would it make sense to merge the `SeenSyncCommitteeContributionCache` and `SeenSyncCommitteeCache` into a single class? You would only need to namespace the `seenCache` maps

packages/lodestar/src/db/seenSyncCommitteeContributionCache.ts
dapplion: Also you can use a Set instead of a Map of booleans
```ts
private seenCache = new Map<phase0.Slot, Set<SeenCacheKey>>();
```

packages/lodestar/src/db/seenSyncCommitteeContributionCache.ts
```diff
@@ -0,0 +1,90 @@
+import {phase0, altair} from "@chainsafe/lodestar-types";
+import {NUM_SLOTS_IN_CACHE, slotRootKey, SlotRootKey} from "./repositories/utils/syncCommittee";
+
+type ContriButionsBySlotRoot = Map<SlotRootKey, altair.SyncCommitteeContribution[]>;
+type SeenCacheKey = string;
```
dapplion: Add a JSDoc comment with the meaning of this string:
```
`slot + "_" + aggregatorIndex + "_" + subCommitteeIndex`
```

packages/lodestar/src/db/seenSyncCommitteeCache.ts
dapplion: Doing it this way would be consistent with the seenCache of `SeenSyncCommitteeContributionCache`

packages/lodestar/src/db/seenSyncCommitteeContributionCache.ts
```diff
@@ -0,0 +1,90 @@
...
+   * This is for the validation.
+   */
+  hasContributionAndProof(contributionAndProof: altair.ContributionAndProof): boolean {
+    const slot = contributionAndProof.contribution.slot;
+    const seenContributions = this.seenCache.get(slot);
+    if (seenContributions) {
+      const key = seenCacheKey(contributionAndProof);
+      return seenContributions.has(key);
+    }
+    return false;
```
dapplion: Could be simplified to
```ts
return this.seenCache.get(slot)?.has(seenCacheKey(contributionAndProof)) === true
```

packages/lodestar/src/db/seenSyncCommitteeContributionCache.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  /**
+   * This is for the block factory
+   */
+  getSyncCommitteeContributions(slot: phase0.Slot, beaconBlockRoot: phase0.Root): altair.SyncCommitteeContribution[] {
```
dapplion: Shouldn't this function return a single already pre-aggregated contribution?

packages/lodestar/src/chain/validation/syncCommittee.ts
twoeths: in this scenario we don't really need head state since what we want to get validator pubkey and sync committees. I suppose we should create `chain.getHeadStateUnsafe()` and audit all use cases of `chain.getHeadState()` to use that instead?

packages/lodestar/src/db/seenSyncCommitteeContributionCache.ts
twoeths: I'll add `getSyncAggregate()`

packages/lodestar/src/db/seenSyncCommitteeContributionCache.ts
twoeths: I'd like to keep it separated because it cache is for different purposes respective to 2 separate gossip validations.
+ SeenSyncCommitteeCache has `contributionCaches` as an array by subnet and each one is a map, it's for `sync_committee_{subnet_id}` topic
+ SeenSyncCommitteeContributionCache has `contributionsCache` has a map by slot and `Map<SlotRootKey, altair.SyncCommitteeContribution[]>`, it's for sync_committee_contribution_and_proof

merging those to a single cache would make it confused to me when I want to look back

packages/beacon-state-transition/src/fast/signatureSets/syncCommittee.ts
twoeths: once we have `validatorIndices` in sync committee, we can use `epochCtx` to get `pubkey` directly, this is the same `TODO` to other places

wemeetagain (commented):
I think we can fix the performance issues (mainly from not having a sync committee cache) in a later PR

dapplion:
Commit https://github.com/ChainSafe/lodestar/pull/2479/commits/ee8b0e8df50967447a52d86ef71d03dd5195396f fixes the performance issues:
- Added sync committee indexes cache to efficiently validate and process items
- Pre-aggregate sync signatures and sync contributions with the same strategy here https://github.com/ChainSafe/lodestar/issues/2415 simplifies the caches and reduces the memory footprint

Note: altair code is not super stable and should be reviewed once we have the whole stack

dapplion (approved):

PR #2484, opened 2021-05-04
Altair Metadata/Discv5

3xtr4t3rr3str14l:
part of discv5 spec of altair (https://github.com/ethereum/eth2.0-specs/blob/dev/specs/altair/p2p-interface.md#the-discovery-domain-discv5)

### DIFF
--- a/packages/lodestar/src/network/metadata/metadata.ts
+++ b/packages/lodestar/src/network/metadata/metadata.ts
@@ -1,6 +1,6 @@
 import {BitVector, toHexString} from "@chainsafe/ssz";
 import {ENR} from "@chainsafe/discv5";
-import {phase0} from "@chainsafe/lodestar-types";
+import {altair, phase0} from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {ChainEvent, IBeaconChain} from "../../chain";
 import {ILogger} from "@chainsafe/lodestar-utils";
@@ -8,7 +8,10 @@ import {getENRForkID} from "./utils";
 import {computeEpochAtSlot} from "@chainsafe/lodestar-beacon-state-transition";
 
 export interface IMetadataOpts {
-  metadata?: phase0.Metadata;
+  metadata?: {
+    phase0?: phase0.Metadata;
+    altair?: altair.Metadata;
+  };
 }
 
 export interface IMetadataModules {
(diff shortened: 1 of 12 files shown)


packages/types/src/phase0/types/wire.ts
```diff
@@ -33,6 +33,7 @@ export type Ping = Uint64;
 export interface Metadata {
```
wemeetagain: You'll need to make a `Metadata` type for altair, this one is for phase0 and can't change.

Also, the `MetadataController` will need an additional method / tweaked methods eg: `getPhase0(): phase0.Metadata` and `getAltair(): altair.Metadata` so that we can support both protocols `/eth2/beacon_chain/req/metadata/1/` and `/eth2/beacon_chain/req/metadata/2/`

packages/lodestar/src/network/metadata/metadata.ts
```diff
@@ -26,22 +29,25 @@ export class MetadataController {
   private enr?: ENR;
   private config: IBeaconConfig;
   private chain: IBeaconChain;
-  private _metadata: phase0.Metadata;
+  private _metadata: {phase0: phase0.Metadata; altair: altair.Metadata};
```
wemeetagain: You only need to keep a single `altair.Metadata` around since `altair.Metadata` extends `phase0.Metadata`, so it satisfies the `phase0.Metadata` interface.

eg:
```ts
  private _metadata: altair.Metadata;
  ...
  get allPhase0(): phase0.Metadata {
    return this._metadata as phase0.Metadata;
  }
  ...
```

wemeetagain (changes requested):
Ssz/type additions looks good, one comment on the metadata controller

packages/lodestar/src/network/metadata/metadata.ts
3xtr4t3rr3str14l: If `_metadata.syncnets` gets set though, then when we call `allPhase0` it would still have syncnets in the object even though it was casted to `phase0.Metadata`.  Maybe we just decompose the `_metadata` object and send only the phase0-relevant parts?  like:

```ts
return {
   attnets: this._metadata.attnets,
   seqNumber: this._metadata.seqNumber
} as phase0.Metadata.
```

3xtr4t3rr3str14l: pushed this^

dapplion (approved):

wemeetagain (approved):
LGTM

PR #2486, opened 2021-05-04
Fast path for regen

wemeetagain:
**Motivation**

Looking at our prater node, many regen jobs are dropped due to the queue always being full.
In practice, many regen jobs may be resolved with already-cached states and do not need to be queued.

**Description**

Add several opportunities for regen methods to first look in the caches before putting the regen job in the queue.


twoeths (approved):

dapplion:
Updated to master with this branch half an hour ago and you can see the difference.

Screenshot from 2021-05-05 10-19-45

PR #2493, opened 2021-05-05
Add light client getState

wemeetagain:
**Motivation**

Part of the light client hackathon.
We want the light client to be able to fetch a state proof, reassemble it into an ssz partial. This PR introduces minimal code for a POC of this.

**Description**

- Add light client `getState` function. It uses the lodestar `/proof/:stateId` REST endpoint to request a proof of a state and return a partial `TreeBacked<BeaconState>`.
The proof is specified by the stateId, eg: `"head"` and the paths, eg: `[["slot"], ["validators", 0, "exitEpoch"]]`.
This is a POC because the proof is not verified against a known state root -- but it is actually reassembling a state from a multiproof.

### DIFF
--- a/packages/light-client/src/getState.ts
+++ b/packages/light-client/src/getState.ts
@@ -1,0 +1,24 @@
+import axios from "axios";
+
+import {TreeBacked} from "@chainsafe/ssz";
+import {phase0} from "@chainsafe/lodestar-types";
+import {deserializeProof} from "@chainsafe/persistent-merkle-tree";
+import {config} from "@chainsafe/lodestar-config/mainnet";
+
+const BEACON_URL = "http://localhost:9596";
+const STATE_PROOF_URL = `${BEACON_URL}/eth/v1/lodestar/proof`;
+
+const client = axios.create();
+
+type Path = (string | number)[];
+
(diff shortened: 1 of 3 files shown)


packages/light-client/src/getState.ts
```diff
@@ -0,0 +1,24 @@
+import axios from "axios";
+
+import {TreeBacked} from "@chainsafe/ssz";
+import {phase0} from "@chainsafe/lodestar-types";
+import {deserializeProof} from "@chainsafe/persistent-merkle-tree";
+import {config} from "@chainsafe/lodestar-config/mainnet";
+
+const BEACON_URL = "http://localhost:9596";
```
twoeths, line 8: probably make this a configuration variable?

dapplion: This should be a class, `LightclientApiClient` that accepts the URL as constructor param

dapplion:
Should add an e2e test to start a BeaconNode instance with a listening REST API endpoint, give it a generated state and then call the LightclientApiClient to get a proof and verify it.

dapplion (dismissed):

dapplion (approved):

PR #2494, opened 2021-05-06
Lightclient sync update producer + API

dapplion:
**Motivation**

Part of Hackaton Lightclient effort https://github.com/ChainSafe/lodestar/issues/2388

**Description**

**API spec**

`GET /eth/v1/lightclient/best_update/:periods`

Returns 0 or more updates per period, targeting 1 update per period, finalized and with the most bits and latest.

`GET GET /eth/v1/lightclient/latest_update_finalized/`

Returns the best update of the latest finalized checkpoint as known by the node. Must return exactly one update. May return an update for the previous finalized checkpoint if no update exist for the latest

*TODO*

`WS /best-updates`

Unidirectional node to client. Node will push updates whenever a call to any of the above methods would result is a different value

**LightClientUpdater***

Class to efficiently create and store updates to satisfy the above API subscribing to onHead and onFinalized events


dapplion:
Ready to merge once credentials for `Publish LC demo` action are added. The build of the ui is done in a different command `yarn build-ui` so it doesn't interfere with the default build run of all other packages.

wemeetagain (approved):

PR #2508, opened 2021-05-11
Altair fast state transition

wemeetagain:
**Motivation**
On the quest towards an altair testnet, we need to update our "fast" state transition to work with altair

**Description**
Adds a "fast" altair state transition (although its not fast without a sync committee cache)


dapplion (approved):
Read it all, looks good :fire: . Can't guarantee correctness but spec tests are passing. Looking forward to reduce code duplication in future PRs

Issue #2502, opened 2021-05-11
Array.from(readonlyValues(aggregationBits)) performance issue


twoeths:
This is from the new performance test in #2501 - it takes almost 1.5s for `Array.from(readonlyValues(aggregationBits))`
<img width="1460" alt="statusProcessEpoch2" src="https://user-images.githubusercontent.com/10568965/117744292-9a31dc80-b232-11eb-860c-0497e90c199e.png">
(notice that the statistic is 1 line shift to the downside somehow)

Currently for  `readonlyValues` of BitList, we return an iterator by yielding each separate bit and then the consumer (lodestar) aggregate them again as `boolean[]`.

We should be able to write an util to access the respective chunks and the consumer can just use that.

dapplion:
The `attestation` is a tree-backed object?

twoeths:
yeah everything inside CachedBeaconState is TreeBacked objects

PR #2510, opened 2021-05-12
Faster aggregation bits read

twoeths:
**Motivation**

+ `readonlyValues(aggregationBits)` is not bad, it's just not worth to create an array of boolean from there. Use precomputed arrays of boolean from all possible bytes instead.
+ Avoid using closure whenever possible

**Description**
+ Precompute arrays of booleans from all possible bytes (0 - 255)
+ Write `getAggregationBytes` which extract byte data from chunks
+ Write `getAggregationBit` to get boolean value of a specific index. Note: don't use this to access aggregation bits of all indices in committee, see the `attestation.test.ts` performance test.

Also I move the closure `statusProcessEpoch` inside `epochProcess.ts` to a separate file and it gains ~0.7s in the performance test.

part of #2502

**Performance tests**

|#|master|this branch|
|-|-------|-----------|
|epoch transition|3.5s|1.5s|
|access aggregation bits|14.5s|3.1s|

**Question/concern**
+ Although this epoch transition improvement is not important as we're close to the end of phase0, accessing `aggregationBits` is still necessary in a couple of places and for altair sync committee as well
+ Consider applying this to other places?
+ Do we want/is it possible to apply to ssz somehow?


dapplion (approved):
Really nice idea to speed this up! @wemeetagain we should consider moving this to SSZ to speed up the entire library.

I've re-organized the code a bit to reduce complexity so it's easier to understand and mantain. Now the trade-off between readability and performance is a bit better =D

Issue #2354, opened 2021-04-09
Remove naive state transition


wemeetagain:
**Is your feature request related to a problem? Please describe.**

Navigating the state transition package can be frustrating, and it can be confusing determining if/when an implementation is used in production or part of the naive suite. The naive/fast distinction is confusing. The naive code is not used or useful to us.

**Describe the solution you'd like**

Remove naive code entirely.
IMO the naive code is a net detriment and not worth maintaining.
Time spent making naive code more readable can be used making the fast code more readable (eg: with more significant code comments).
Any important exported naive interfaces can be retained by wrapping the fast implementation, if exporting those interfaces is important.

**Describe alternatives you've considered**

Status quo

dapplion:
Agree with above

PR #2513, opened 2021-05-12
Move naive state transition

wemeetagain:
**Motivation**
See #2354 

**Description**

This does not _remove_ the naive state transition but rather reorganizes the state transition code to put the naive code in a subdirectory. This reorganization can give us most of the benefits of #2354 without entirely throwing the naive code away. A future PR can remove the naive code entirely if desired.

Important implementation items:
- all naive code moved to `beacon-state-transition/src/naive/{altair, phase0}/*`
- `beacon-state-transition/src/{altair, phase0}/fast/*` moved to  ``beacon-state-transition/src/{altair, phase0}/*`
- `beacon-state-transition/src/fast/` renamed to `beacon-state-transition/src/allForks`
  - `fastStateTransition` renamed to `stateTransition`
  - allForks re-exports the allForks types, similarly to how phase0 and altair types are re-exported
- In lodestar, `allForks.stateTransition` and `allForks.processSlots` used across the board
  - `allForks.processSlots` function signature differs from `phase0.processSlots`, returns a post state instead of just mutating the state.
- spec test reorganization
  - naive spec tests moved to `spec-test-runner/test/spec/naive/*`
  - phase0 tests grouped under `spec-test-runner/test/spec/phase0/*`

Important future items:
- Our altair state transition relies on naive implementations in several places, listed below. This will likely need to be resolved / sped up.
  - reward calculation in block / sync committee processing
  - eligibility calculation in balance state accessors
- As it stands, adding/editing spec tests is a very verbose and time consuming task. As an enhancement, it would be great to refactor the spec tests to make them more clearly correct and comprehensive.
  - the test:spec-min and test:spec-main were testing the naive implementations
  - we don't currently implement both mainnet and minimal spec tests for every test for our fast impl. In fact, we usually only implement one or the other, inconsistently


twoeths (approved):

.github/workflows/test-spec.yml
```diff
@@ -32,13 +32,10 @@ jobs:
...
-      - name: Spec min
-        run: yarn test:spec-min
+      - name: Spec tests
+        run: yarn test:spec
         working-directory: packages/spec-test-runner
-      - name: Spec fast
-        run: yarn test:spec-fast
-        working-directory: packages/spec-test-runner
-      - name: Spec main
-        run: yarn test:spec-main
```
dapplion, line 42: Why un-group this different runs?

dapplion (approved):

dapplion:
This is an awesome step forward. However I'm in favor of removing the naive code entirely. Here the code will still show-up in searches and *be there* and having tests means it may have to be updated on changes.

If we rely on naive code for altair that's fine. The those naive parts can just be the "fast" for now, until we optimize it.

If needed we can park the naive code in a separate repo for historical value.

.github/workflows/test-spec.yml
wemeetagain: Good question
- the test:spec-min and test:spec-main were testing the naive implementations
- we don't currently implement both mainnet and minimal spec tests for every test for our fast impl. In fact, we usually only implement one or the other, inconsistently

As it stands, adding/editing spec tests is a very verbose and time consuming task. As an enhancement, it would be great to refactor the spec tests to make it more clearly correct and comprehensive.

wemeetagain:
> This is an awesome step forward. However I'm in favor of removing the naive code entirely. Here the code will still show-up in searches and _be there_ and having tests means it may have to be updated on changes.
> 
> If we rely on naive code for altair that's fine. The those naive parts can just be the "fast" for now, until we optimize it.
> 
> If needed we can park the naive code in a separate repo for historical value.

I agree, I'd like to see the naive code removed entirely.
I started this PR just removing the naive code, but it turned out to be more manageable to reorganize things first.
A followup PR to remove the naive code entirely will now be very straightforward, just remove a few directories named `naive`.

PR #2514, opened 2021-05-13
Implement Altair API

### LINKED ISSUE #2512: Altair API
https://github.com/ethereum/eth2.0-APIs/blob/master/beacon-node-oapi.yaml 

- [ ] `/eth/v1/beacon/states/{state_id}/sync_committees`
- [ ] `/eth/v2/beacon/blocks/{block_id}`
- [ ] `/eth/v2/debug/beacon/states/{state_id}`
- [ ] `/eth/v2/validator/blocks/{slot}`
- [ ] `/eth/v1/validator/sync_committee_subscriptions`

Update by Tuyen on May 13:
- [ ] POST `/eth/v1/validator/duties/sync/{epoch}`
- [ ] POST `/eth/v1/beacon/pool/sync_committees`
- [ ] GET `/eth/v1/validator/sync_committee_contribution`
- [ ] POST `/eth/v1/validator/contribution_and_proofs`

dapplion:
**Motivation**

Implement altair API. Closes https://github.com/ChainSafe/lodestar/issues/2512. Part of https://github.com/ChainSafe/lodestar/issues/2388

**Description**

- Implements all new routes for syncCommittees
- Updates existing routes to support v2 and type multiplexing
- Extends network and PeerManager to deal with syncnets, generalizing the existing AttnetsService
- Adds `currSyncComitteeValidatorIndexMap` to epochCtx to efficiently validate and compute subnets


packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -182,6 +206,47 @@ export class ValidatorApi implements IValidatorApi {
...
+   */
+  async getSyncCommitteeDuties(epoch: number, validatorIndices: ValidatorIndex[]): Promise<altair.SyncDutiesApi> {
+    this.notWhileSyncing();
+
+    if (validatorIndices.length === 0) {
+      throw new ApiError(400, "No validator to get attester duties");
+    }
+
+    // May request for an epoch that's in the future
+    await this.waitForNextClosestEpoch();
```
twoeths, line 230: I think we don't need to as we have current sync committee and next sync committee, and each sync period is ~27h

packages/lodestar/src/network/network.ts
```diff
@@ -152,9 +155,14 @@ export class Network implements INetwork {
...
-  prepareBeaconCommitteeSubnet(subscriptions: phase0.BeaconCommitteeSubscription[]): void {
-    this.attService.addBeaconCommitteeSubscriptions(subscriptions);
-    this.peerManager.onBeaconCommitteeSubscriptions();
+  prepareBeaconCommitteeSubnet(subscriptions: CommitteeSubscription[]): void {
+    this.attnetsService.addCommitteeSubscriptions(subscriptions);
+    if (subscriptions.length > 0) this.peerManager.onCommitteeSubscriptions();
+  }
+
+  prepareSyncCommitteeSubnets(subscriptions: CommitteeSubscription[]): void {
+    this.syncnetsService.addCommitteeSubscriptions(subscriptions);
```
twoeths, line 164: I suppose this is not the same to `prepareBeaconCommitteeSubnet`? the subnet of SyncCommittee is totally different from attestation so it's not the same business logic there.

Instead, we only need to use gossip to subscribe to sync committee subnet here?!

packages/lodestar/src/api/rest/validator/duties/getSyncCommitteeDuties.ts
```diff
@@ -0,0 +1,34 @@
+import {ValidatorIndex} from "@chainsafe/lodestar-types";
+import {ApiController} from "../../types";
+
+export const getAttesterDuties: ApiController<null, {epoch: number}, ValidatorIndex[]> = {
```
twoeths: should be `getSyncCommitteeDuties`

packages/lodestar/src/api/impl/validator/validator.ts
```diff
@@ -226,41 +291,100 @@ export class ValidatorApi implements IValidatorApi {
...
+      const committeeIndices = state.currSyncComitteeValidatorIndexMap.get(sub.validatorIndex);
+      if (committeeIndices) {
+        for (const committeeIndex of committeeIndices) {
+          const subnet = Math.floor(committeeIndex / SYNC_COMMITTEE_SUBNET_SIZE);
+          subs.push({
+            validatorIndex: sub.validatorIndex,
+            subnet: subnet,
+            // Subscribe until the end of `untilEpoch`: https://github.com/ethereum/eth2.0-APIs/pull/136#issuecomment-840315097
+            slot: computeStartSlotAtEpoch(this.config, sub.untilEpoch + 1),
+            isAggregator: true,
```
twoeths, line 379: similar to the other concern, in my understanding this is different from `prepareBeaconCommitteeSubnet` so we don't have to refactor to have `CommitteeSubscription` type

packages/lodestar/src/sync/gossip/handler.ts
```diff
@@ -117,10 +117,11 @@ export class BeaconGossipHandler {
     signature: altair.SyncCommitteeSignature
   ): Promise<void> => {
     // TODO: Review if we need to check shouldProcessAttestation() like with onAttestation
-
-    // TODO: Do this much better to be able to access this property in the handler
-    const indexInSubCommittee = (signature as SyncCommitteeSignatureIndexed).indexInSubCommittee;
-    this.db.syncCommitee.add(subnet, signature, indexInSubCommittee);
+    if (this.network.syncnetsService.shouldProcess(subnet, signature.slot)) {
```
twoeths: I think we don't need this check, we should always process

packages/lodestar/src/network/subnetsService.ts
```diff
@@ -34,10 +42,18 @@ export interface IAttestationServiceModules {
...
+}
+
+export function getSyncnetsService(modules: IAttestationServiceModules): SubnetsService {
+  return new SubnetsService(modules, GossipType.sync_committee, "syncnets");
+}
+
 /**
  * Manage random (long lived) subnets and committee (short lived) subnets.
  */
-export class AttestationService implements IAttestationService {
```
twoeths, line 40: maybe we don't need to change this class as it's specific for attestation only

packages/lodestar/src/api/impl/validator/validator.ts
dapplion: It's important because the validator client will request the next duties exactly at the border between periods. If the clock of the beacon node lags behind the beacon node will reject the request because the validator client is requesting the next period too early. Since our node is way more overloaded than the validator client this happens regularly

packages/lodestar/src/network/network.ts
dapplion: No, it's not the same. Note that there are two different classes
- `attnetsService`: Manage attestation subnets
- `syncnetsService`: Manage sync committee subnets

Both instantiate the same class `SubnetsService` since the logic to handle attnets and syncnets is *almost* identical. I've created this generic `CommitteeSubscription` type that gives just the minimal info needed for `SubnetsService`

packages/lodestar/src/network/subnetsService.ts
dapplion: I've generalized it for attnets and syncnets

packages/lodestar/src/api/rest/validator/duties/getSyncCommitteeDuties.ts
dapplion: fixed

packages/lodestar/src/api/impl/validator/validator.ts
dapplion: See other replies

packages/lodestar/src/sync/gossip/handler.ts
dapplion: Sounds good, removing check

wemeetagain:
There are some file conflicts, otherwise LGTM

wemeetagain (approved):

PR #2515, opened 2021-05-13
Cache domains in cachedBeaconState.ts

### LINKED ISSUE #2505: Cache forkDataRoot
**Feature request**

Currently we compute forkDataRoot every time we need to verify a signature. We could compute all `forkDataRoot`s for the known forks after genesis is known, and cache them.

We could even cache the domains as we compute them.

**Proposal**

Expose somewhere an interface like

```ts
export interface IDomainCache {
  getDomainAtSlot(domainType: DomainType, slot: Slot): Buffer;
}
```

----

**EDIT**: Note: I have not data to prove that this is an actual performance issue, it was just an idea that would be nice to de-duplicate work but maybe it has a negligible impact.

One may browse the issues with the performance label which are proven to have a serious impact on the node performance https://github.com/ChainSafe/lodestar/labels/Performance

twoeths:
**Motivation**

We don't want to compute domain all the time for the same values, should cache it somewhere.

**Description**

Cache domains in `cacheBeaconState.ts`

Closes #2505

### DIFF
--- a/packages/beacon-state-transition/src/allForks/util/cachedBeaconState.ts
+++ b/packages/beacon-state-transition/src/allForks/util/cachedBeaconState.ts
@@ -10,7 +10,7 @@ import {
   readonlyValues,
   TreeBacked,
 } from "@chainsafe/ssz";
-import {allForks, altair, ParticipationFlags} from "@chainsafe/lodestar-types";
+import {allForks, altair, DomainType, ParticipationFlags, phase0} from "@chainsafe/lodestar-types";
 import {ForkName, IBeaconConfig} from "@chainsafe/lodestar-config";
 import {Tree} from "@chainsafe/persistent-merkle-tree";
 import {MutableVector} from "@chainsafe/persistent-ts";
@@ -24,6 +24,7 @@ import {
   fromParticipationFlags,
   IParticipationStatus,
 } from "./cachedEpochParticipation";
+import {computeDomain} from "../../util";
 
 /**
  * `BeaconState` with various caches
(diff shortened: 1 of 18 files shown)


packages/beacon-state-transition/src/allForks/util/cachedBeaconState.ts
```diff
@@ -253,3 +276,13 @@ export const CachedBeaconStateProxyHandler: ProxyHandler<CachedBeaconState<allFo
...
+
+/**
+ * Caution: not for production, only used for sim tests.
+ */
+export function clearDomainCache(): void {
+  domainCache.clear();
+}
+
+/** Domain cache, share across beacon states */
+const domainCache = new Map<ForkName, Map<DomainType, Buffer>>();
```
dapplion, line 288: Having non-universal data attached to the global scope is a bad idea. If someone runs two nodes with different genesis it's can break everything. Also it makes it harder to track memory and not good etiquette anyway. We should only do it for very specific occasions like caching hex to uint conversions and uint to boolean[].

dapplion, line 288: Having non-universal data attached to the global scope is a bad idea. If someone runs two nodes with different genesis it's can break everything. Also it makes it harder to track memory and not good etiquette anyway. We should only do it for very specific occasions like caching hex to uint conversions and uint to boolean[].

dapplion (changes requested):

packages/beacon-state-transition/src/allForks/util/cachedBeaconState.ts
```diff
@@ -192,6 +193,28 @@ export class BeaconStateContext<T extends allForks.BeaconState> {
...
+      domainCache.set(forkName, domainByType);
+    }
+    let domain = domainByType.get(domainType);
+    if (!domain) {
+      const forkVersion = this.config.getForkVersion(slot);
+      domain = computeDomain(this.config, domainType, forkVersion, state.genesisValidatorsRoot);
+      domainByType.set(domainType, domain);
+    }
+    return domain;
+  }
```
wemeetagain, line 217: Maybe move this method to EpochContext? This class (intentionally) doesn't have any application logic but rather defers to the various underlying caches.

In that case, `slot` can be a required param.

This shouldn't effect downstream consumers of the method since CachedBeaconState implements EpochContext. eg: usage will still be `state.getDomain(...)`

Also, the EpochContext can cache the genesisValidatorsRoot. (we could begin to consilidate the domain, fork digest, etc caching that requires config + genesisValidatorsRoot)

packages/beacon-state-transition/src/allForks/util/cachedBeaconState.ts
wemeetagain: Right. The cache can be attached on the cached state / epochCtx and copied by reference as the cache is cloned.

wemeetagain (changes requested):

wemeetagain:
After talking to @dapplion further about this, it might be nice to attach this cache/method to the IBeaconConfig so the domain logic can be shared across the beacon node and validator (the validator doesn't have a CachedBeaconState available).

The IBeaconConfig could load in a genesisValidatorsRoot, then be able to handle fork digests, domains, etc (anything combining the fork schedule and genesisValidatorsRoot information).

dapplion:
Also I have not data to prove that this is an actual performance issue, it was just an idea that would be nice to de-duplicate work but maybe it has a negligible impact.

You can browse the issues with the `performance` label which are proven to have a serious impact on the node performance https://github.com/ChainSafe/lodestar/labels/Performance

dapplion:
We should revisit this after merging https://github.com/ChainSafe/lodestar/pull/2585 seeing it has a noticeable performance improvement

twoeths:
waiting for #2585

PR #2519, opened 2021-05-13
Handle fork transitions in gossip

dapplion:
**Motivation**

Handle fork transition properly in Gossip handler

**Description**

Follow same strategy than in SubnetService https://github.com/ChainSafe/lodestar/blob/765067145921fa1d2fe1209b6699d437cb65b0f8/packages/lodestar/src/network/subnetsService.ts#L167

- On start-up: subscribe to 
- ONLY ONCE: Two epoch before the fork, re-subscribe all existing random subscriptions to the new fork
- ONLY ONCE: Two epochs after the fork, un-subscribe all subnets from the old fork

Also re-organizes the gossip handler code a bit :)

### DIFF
--- a/packages/lodestar/src/network/interface.ts
+++ b/packages/lodestar/src/network/interface.ts
@@ -38,6 +38,12 @@ export interface INetwork {
   /** Subscribe, search peers, join long-lived syncnets */
   prepareSyncCommitteeSubnets(subscriptions: CommitteeSubscription[]): void;
   reStatusPeers(peers: PeerId[]): void;
+
+  // Gossip handler
+  subscribeGossipCoreTopics(): void;
+  unsubscribeGossipCoreTopics(): void;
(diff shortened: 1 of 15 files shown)


packages/lodestar/src/network/interface.ts
```diff
@@ -38,6 +38,12 @@ export interface INetwork {
   /** Subscribe, search peers, join long-lived syncnets */
   prepareSyncCommitteeSubnets(subscriptions: CommitteeSubscription[]): void;
   reStatusPeers(peers: PeerId[]): void;
+
+  // Gossip handler
+  subscribeGossipCoreTopics(): void;
+  unsubscribeGossipCoreTopics(): void;
+  isSubscribedToGossipCoreTopics(): boolean;
```
wemeetagain, line 45: :+1:

wemeetagain (approved):

PR #2520, opened 2021-05-13
Add Altair VC client

dapplion:
**Motivation**

Completes the work started with https://github.com/ChainSafe/lodestar/pull/2514. Part of #2388

**Description**

- Adds a new service pair: `SyncCommitteeService` + `SyncCommitteeDutiesService` which work very similar to the attester services.
- Abstract getting indices in a different service `IndicesService` used by sync and att services

### DIFF
--- a/packages/validator/src/services/syncCommittee.ts
+++ b/packages/validator/src/services/syncCommittee.ts
@@ -24,11 +24,6 @@ export class SyncCommitteeService {
     private readonly validatorStore: ValidatorStore,
     indicesService: IndicesService
   ) {
-    this.config = config;
-    this.logger = logger;
-    this.apiClient = apiClient;
-    this.clock = clock;
-    this.validatorStore = validatorStore;
     this.dutiesService = new SyncCommitteeDutiesService(
       config,
       logger,
--- a/packages/validator/src/services/syncCommittee.ts
+++ b/packages/validator/src/services/syncCommittee.ts
@@ -152,7 +152,7 @@ export class SyncCommitteeService {
     }
 
     this.logger.verbose("Producing SyncCommitteeContribution", logCtx);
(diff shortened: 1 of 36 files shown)


packages/validator/src/services/syncCommittee.ts
```diff
@@ -0,0 +1,182 @@
...
+   */
+  private async produceAndPublishSyncCommittees(slot: Slot, duties: SyncDutyAndProof[]): Promise<Root> {
+    const logCtx = {slot};
+
+    // /eth/v1/beacon/blocks/:blockId/root -> at slot -1
+
+    // Produce one attestation data per slot and subcommitteeIndex
+    // Spec: the validator should prepare a SyncCommitteeSignature for the previous slot (slot - 1)
+    // as soon as they have determined the head block of slot - 1
+    const beaconBlockRoot = await this.apiClient.beacon.blocks.getBlockRoot(slot - 1).catch((e) => {
```
twoeths, line 96: my understanding is we want to get block root at the current slot since we wait for 1/3 slot already, so `getBlockRoot(slot)`

correct me if I'm wrong

twoeths (approved):

packages/validator/src/services/syncCommittee.ts
twoeths: will address in future PRs

dapplion: Yeah that's right, we want to sign the head at that slot since the signatures will be included in the next. Even we can / should do `getBlockRoot("head")`

dapplion: See https://github.com/ChainSafe/lodestar/pull/2522 as conclusion

PR #2521, opened 2021-05-14
Altair block production

dapplion:
**Motivation**

Update block production to be able to do Altair blocks

**Description**

Add extra field SyncAggregate conditionally based on the block root parent

### DIFF
--- a/packages/lodestar/src/chain/factory/block/body.ts
+++ b/packages/lodestar/src/chain/factory/block/body.ts
@@ -3,9 +3,9 @@
  */
 
 import {List} from "@chainsafe/ssz";
-import {Bytes96, Bytes32, phase0, allForks} from "@chainsafe/lodestar-types";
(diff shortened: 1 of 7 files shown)


packages/lodestar/src/chain/factory/block/body.ts
```diff
@@ -38,6 +40,16 @@ export async function assembleBody(
     deposits: deposits as List<phase0.Deposit>,
     voluntaryExits: voluntaryExits as List<phase0.SignedVoluntaryExit>,
   };
+
+  if (computeEpochAtSlot(config, slot) >= config.params.ALTAIR_FORK_EPOCH) {
+    const block: altair.BeaconBlockBody = {
+      ...blockCommon,
+      syncAggregate: db.syncCommitteeContribution.getSyncAggregate(slot, parentBlockRoot),
```
twoeths: should this be the parent slot? `getSyncAggregate(parentSlot, parentBlockRoot)`

twoeths (approved):

PR #2524, opened 2021-05-14
add LODESTAR_NETWORK env var to rootDir path in docker-compose.yml

3xtr4t3rr3str14l:
**Motivation**

This makes it more convenient to switch to a different network when running lodestar as a dockerized process.  For example, when running as `docker-compose -p myProject up` you can change the network that the process runs on by changing the .env file without having to worry about using the same directory that was used for the previous network under `myProject`.  So in this PR, if you ran the same project on two separate occassions on two different networks (say, mainnet and prater), then you would have two directories (one for each network) under the mapped `myProject` volume.


dapplion:
Not a big fan of this approach. You can achieve the same results with the ENV `XDG_DATA_HOME`, like this:
```bash
# .env
XDG_DATA_HOME=/data
```
```yaml
# docker-compose.yml
command: beacon --api.rest.host 0.0.0.0 --metrics.enabled --logFile /logs/beacon.log --logLevelFile debug
```

dapplion:
Still I'm not in favor of merging this. Only us developers will run the exact same environment in multiple networks. You should apply the configuration you describe here to the environment you run the node in. I maintain a modified docker-compose in the nodes I run in the cloud and you can do the same with this approach if it's more convenient to you.

PR #2531, opened 2021-05-18
Use same workflow prepare steps block for all actions

dapplion:
- Recycle same steps on all actions.
- Force to run yarn clean on all builds


dapplion:
CI passing

wemeetagain (approved):
:100:

PR #2536, opened 2021-05-19
Split attnets and syncnets services

dapplion:
**Motivation**

Currently the class managing attnets and syncnets was the same. However that implementation was not following the spec since:
- syncnets do not need long lived random subscriptions
- syncnets persist the subscriptions requested by the validators to the ENR instead of the long lived random subscriptions

**Description**

Split into two classes `SyncnetsService` and `AttnetsService`. The design is similar but the functionality is customized to each requirements.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,10 +27,10 @@ jobs:
         run: yarn build
         if: steps.cache-deps.outputs.cache-hit == 'true'
       # </common-build>
-      - name: Lint
-        run: yarn lint
       - name: Check Types
         run: yarn run check-types
+      - name: Lint
+        run: yarn lint
       - name: Unit tests
         run: yarn test:unit
       - name: Upload coverage data
--- a/packages/lodestar/src/network/subnets/interface.ts
+++ b/packages/lodestar/src/network/subnets/interface.ts
@@ -1,0 +1,20 @@
+import {Slot, ValidatorIndex} from "@chainsafe/lodestar-types";
+
+/** Generic CommitteeSubscription for both beacon attnets subs and syncnets subs */
+export type CommitteeSubscription = {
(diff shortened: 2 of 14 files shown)


.github/workflows/test.yml
```diff
@@ -27,10 +27,10 @@ jobs:
         run: yarn build
         if: steps.cache-deps.outputs.cache-hit == 'true'
       # </common-build>
-      - name: Lint
-        run: yarn lint
       - name: Check Types
         run: yarn run check-types
+      - name: Lint
+        run: yarn lint
```
dapplion, line 33: Run lint **after** check types. It's annoying to catch type issues with a linter instead of the compiler

packages/lodestar/src/network/subnets/interface.ts
```diff
@@ -0,0 +1,20 @@
...
+};
+
+export interface ISubnetsService {
+  start(): void;
+  stop(): void;
+  addCommitteeSubscriptions(subscriptions: CommitteeSubscription[]): void;
+  getActiveSubnets(): number[];
+}
+
+export interface IAttnetsService extends ISubnetsService {
```
wemeetagain, line 18: Either remove this interface or add a `ISyncnetsService` interface.
My preference is removing this interface (so that click thru navigates to the implementation and not just the interface, which only has a single implementation anyways)

wemeetagain (changes requested):
LGTM, one minor comment

wemeetagain (approved):

packages/lodestar/src/network/subnets/interface.ts
dapplion: Agree with the concerns but let's discuss that in an upcoming issue

PR #2538, opened 2021-05-20
Add ReqResp Metadata V2

dapplion:
Altair ReqResp PR https://github.com/ChainSafe/lodestar/pull/2444 did not implemented `metadata/v2` which is required by the spec. See https://github.com/ethereum/eth2.0-specs/blob/dev/specs/altair/p2p-interface.md#getmetadata-v2

It also fixes some issues introduced in https://github.com/ChainSafe/lodestar/pull/2484 around handling the metadata class and types for altair


packages/lodestar/src/network/peers/metastore.ts
```diff
@@ -37,13 +41,23 @@ export class Libp2pPeerMetadataStore implements IPeerMetadataStore {
...
+    const metadataV2Serdes = typeToSerdes(this.config.types.altair.Metadata);
+    const stringSerdes: BucketSerdes<ReqRespEncoding> = {
+      serialize: (v) => Buffer.from(v, "utf8"),
+      deserialize: (b) => Buffer.from(b).toString("utf8") as ReqRespEncoding,
+    };
+
+    this.encoding = this.typedStore("encoding", stringSerdes);
+    // Discard existing `metadata` stored values. Store both phase0 and altair Metadata objects as altair
+    // Serializing altair.Metadata instead of phase0.Metadata has a cost of just `SYNC_COMMITTEE_SUBNET_COUNT // 8` bytes
+    this.metadata = this.typedStore("metadata-altair", metadataV2Serdes);
```
dapplion, line 55: The strategy to deal with the different Metadata types is to coerce `phase0.Metadata` into `altair.Metadata` with a zeroed syncnets value. Consumers that want the phase0 can just ignore the syncnets property. SSZ serialization works just fine since it ignores extra properties. Since metadata is ephemeral information the old `metadata` key is abandoned and not migrated.

packages/lodestar/src/network/peers/metastore.ts
```diff
@@ -37,13 +41,23 @@ export class Libp2pPeerMetadataStore implements IPeerMetadataStore {
...
-    this.encoding = this.typedStore("encoding", new StringType()) as PeerStoreBucket<ReqRespEncoding>;
-    this.metadata = this.typedStore("metadata", this.config.types.phase0.Metadata);
-    this.rpcScore = this.typedStore("score", this.config.types.Number64);
-    this.rpcScoreLastUpdate = this.typedStore("score-last-update", this.config.types.Number64);
+
+    const number64Serdes = typeToSerdes(this.config.types.Number64);
+    const metadataV2Serdes = typeToSerdes(this.config.types.altair.Metadata);
+    const stringSerdes: BucketSerdes<ReqRespEncoding> = {
+      serialize: (v) => Buffer.from(v, "utf8"),
+      deserialize: (b) => Buffer.from(b).toString("utf8") as ReqRespEncoding,
```
dapplion, line 49: Same logic as previous `StringType` class but in two lines

packages/lodestar/src/network/metadata.ts
```diff
@@ -103,3 +97,20 @@ export class MetadataController {
...
+
+  return {
+    // Current fork digest
+    forkDigest: forkDigestContext.forkName2ForkDigest(currentFork.name),
+    // next planned fork versin
+    nextForkVersion: nextFork ? nextFork.version : currentFork.version,
+    // next fork epoch
+    nextForkEpoch: nextFork ? nextFork.epoch : FAR_FUTURE_EPOCH,
+  };
+}
```
dapplion, line 121: This util is only used by network/metadata so I've un-nested the metadata directory to the network root for simplicity.

wemeetagain (approved):
LGTM :+1:

twoeths (approved):

PR #2541, opened 2021-05-20
Re-org types organization

dapplion:
**Motivation**

Currently out types are accessible from multiple paths, for example:
```
config.types.Root
config.types.phase0.Root
config.types.altair.Root
config.types.allForks.Root
```

This leads to inconsistencies and makes it harder to reason about if this type belong to a fork or is just propagated from the previous fork.

For example, if you see this `config.types.altair.AttestationData` an observer has no way of knowing if that's actually the phase AttestationData type or a new AttestationData introduced in altair.

**Description**

Re-org types such that:
- Primitive types are only accessible from `config.types.*`
- Phase0 types are only accessible from `config.types.phase0.*`
- `config.types.altair.*` contains only types that were changed or introduced in altair
- `config.types.allForks` contains only *some* types that are present in both forks and must be accessed through the `getTypes()` runtime switcher.


packages/types/src/index.ts
```diff
@@ -23,19 +23,17 @@ export {AllForksSSZTypes} from "./allForks";
...
   const primitive = getPrimitiveTypes();
   const phase0 = getPhase0Types(params, primitive);
-  const altair = getAltairTypes(params, {...primitive, ...phase0});
+  const altair = getAltairTypes(params, primitive, phase0);
   return {
     ...primitive,
-    phase0: {...primitive, ...phase0},
-    altair: {...primitive, ...phase0, ...altair},
+    phase0,
+    altair,
```
dapplion, line 37: Now phase0 = phase0, and altair = altair

wemeetagain:
I appreciate the desire to make things simpler, but I don't think this works well when we need to programmatically access a type that changes between forks. Especially once we include multiple forks. My preference is keeping what we have or resolving this issue before continuing.

eg: The sharding fork will update AttestationData. During altair, but before the sharding fork, how do we programmatically refer to the correct AttestationData? We can't do `config.getTypes(data.slot).AttestationData` anymore, because the altair namespace doesn't actually include the AttestationData type anymore.

> For example, if you see this config.types.altair.AttestationData an observer has no way of knowing if that's actually the phase AttestationData type or a new AttestationData introduced in altair.

The observer actually does know if the type is the same: `config.types.phase0.AttestationData === config.types.altair.AttestationData`.
But I'd say that its not actually important to be able to know the difference. Whats important (imo) is that given a fork (or a slot that resolves to a fork), all available types are immediately known.

dapplion:
> The sharding fork will update AttestationData. During altair, but before the sharding fork, how do we programmatically refer to the correct AttestationData? We can't do config.getTypes(data.slot).AttestationData anymore, because the altair namespace doesn't actually include the AttestationData type anymore.

In that case we can add `AttestationData` to the altair namespace. My point is to restrict getTypes() to the minimum number of types that have to be multiplexed at runtime. getTypes() could even be a dedicated switch like

```ts
function getTypes(fork: ForkName) {
  switch(fork) {
    case ForkName.phase0:
      return {
        SignedBeaconBlock: phase0.SignedBeaconBlock,
        AttestationData: phase0.AttestationData
      }
    case ForkName.altair:
      return {
        SignedBeaconBlock: altair.SignedBeaconBlock,
        AttestationData: phase0.AttestationData
      }
    case ForkName.sharding:
      return {
        SignedBeaconBlock: sharding.SignedBeaconBlock,
        AttestationData: sharding.AttestationData
      }
  }
}
```

wemeetagain:
>  My point is to restrict getTypes() to the minimum number of types that have to be multiplexed at runtime.

Ok I see! So once we decide a type changes between forks, is part of `AllForksSSZTypes`, it just needs to be properly handled in `getTypes`.

wemeetagain (approved):

PR #2548, opened 2021-05-22
Allow to run validator with interop key range

dapplion:
**Motivation**

It is useful for local testnets to run a validator with interop keys quickly

**Description**

Add CLI arg to use a contiguous inclusive range of interop keys
```
--interopIndexes 17..32
```


wemeetagain:
file conflict

wemeetagain (approved):

PR #2565, opened 2021-05-26
enr extension update

### LINKED ISSUE #2561: Error if enr exists
**Describe the bug**

If you already have enr, starting with beacon command fails with:
```
✖ Error: Invalid filetype
    at parse (/Users/nodefactory/Projects/lodestar/packages/cli/src/util/file.ts:47:13)
    at readFile (/Users/nodefactory/Projects/lodestar/packages/cli/src/util/file.ts:89:10)
    at readEnr (/Users/nodefactory/Projects/lodestar/packages/cli/src/config/enr.ts:28:24)
    at persistOptionsAndConfig (/Users/nodefactory/Projects/lodestar/packages/cli/src/cmds/init/handler.ts:85:17)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at Object.beaconHandler [as handler] (/Users/nodefactory/Projects/lodestar/packages/cli/src/cmds/beacon/handler.ts:24:3)
```
because enr doesn't have extension.

**Expected behavior**
should work

**Steps to Reproduce**

**Screenshots**

- OS: 
- Version: 
- Branch: 
- Commit hash:

g11tech:
**Motivation**

**Description**

Closes #2561

### DIFF
--- a/packages/cli/src/cmds/beacon/paths.ts
+++ b/packages/cli/src/cmds/beacon/paths.ts
@@ -37,7 +37,7 @@ export function getBeaconPaths(
   const peerStoreDir = args.peerStoreDir || path.join(beaconDir, "peerstore");
   const configFile = args.configFile || path.join(beaconDir, "beacon.config.json");
   const peerIdFile = args.peerIdFile || path.join(beaconDir, "peer-id.json");
-  const enrFile = args.enrFile || path.join(beaconDir, "enr");
+  const enrFile = args.enrFile || path.join(beaconDir, "enr.yml");
   const logFile = args.logFile;
 
   return {


dapplion:
Why change to `.yml` extension? Could you add rationale in the PR description? This also looks like it will break current nodes since existing ENRs won't be read at their previous path `enr`

g11tech:
> 
> 
> Why change to `.yml` extension? Could you add rationale in the PR description? This also looks like it will break current nodes since existing ENRs won't be read at their previous path `enr`

this is to close issue #2561, as this code is already breaking while reading at the previous path `enr`, as the readFile expects a fileformat/extension of type: `yml`, `yaml` or `json`.

image

image

image

dapplion:
But then the issue is that our file reader is bad. Changing the enr extension is not the solution

g11tech:
> 
> 
> But then the issue is that our file reader is bad. Changing the enr extension is not the solution

even when you not specify the extension (first time around), its writing the `enr` file in `yml` format only ( `enr:..` ).  readfile seems to be returning a parsed file, and unless you want to make an assumption that if `fileFormat` is empty string, its `yml`, which in my opinion is the worse idea. 
or `readFile` interface having an optional `default` format parameter, which can still make sense.

let me know which one sounds good.

dapplion:
> > But then the issue is that our file reader is bad. Changing the enr extension is not the solution
> 
> even when you not specify the extension (first time around), its writing the `enr` file in `yml` format only ( `enr:..` ). readfile seems to be returning a parsed file, and unless you want to make an assumption that if `fileFormat` is empty string, its `yml`, which in my opinion is the worse idea.
> or `readFile` interface having an optional `default` format parameter, which can still make sense.
> 
> let me know which one sounds good.

I would not give it a .yml format. It can be either `enr` or `enr.dat` (Lighthouse).

Here instead of throwing an error it should just return the file in plain text format
https://github.com/ChainSafe/lodestar/blob/f6778740075fe2b75edf94d1db0b5691039cb500/packages/cli/src/util/file.ts#L46-L47

```ts
 default:
   return contents
   ```

Same for stringify
https://github.com/ChainSafe/lodestar/blob/f6778740075fe2b75edf94d1db0b5691039cb500/packages/cli/src/util/file.ts#L64-L65

g11tech:
ok assign the #2561  to me, i will send in the proper fix.

dapplion:
@g11in thanks! If you create new branches from current master, CI will pass. We fixed it in Friday

PR #2568, opened 2021-05-27
Separate lodestar-api package

dapplion:
**Motivation**

It was overdue to move the REST api code into its own package so it can be consumed by the validator, lightclient and dependees on NPM.

I've taken the opportunity to review how we declare and consume the REST routes to be more typesafe and succint. We can use Typescript to our advantage to ensure the client and server are compatible at build time and all the logic matches the spec definition.

I've tried to come up with a setup that derives this code (or part of it) from the openapi definition but I was unsuccessful. The code that you get is good but not good enough and requires significant tweaking defeating its purpose. We also have our own SSZ types objects which the openapi is not aware of complicating the logic in connecting both worlds.

See inline comments for rationale

https://github.com/ChainSafe/lodestar/blob/0b53679f1eaadb9d612ec7e38ddbfb7950f5a469/packages/api/src/routes/index.ts#L10-L47

**Description**

Compact the client and server REST api logic into a separate package.
- The client can be consumed directly, using `cross-fetch`. Consumers can provide their own http client if they want.
- For the server, exposes a `registerRoutes` function which does that given the api implementation `Api` and a fastify server.
- REST parsing tests are moved to the api package and test the client and server directly.
- Re-write Api impl code in function form to automatically type all methods.

Other:
- Bump Fastify to v3

# BREAKING CHANGES

- Ensures our `Api` implementation type matchs the JSON return type of all routes. This means adding a `{data: T}` wrapper to most routes

before
```ts
getGenesis(): Promise<phase0.Genesis>;
```
after
```ts
getGenesis(): Promise<{data: phase0.Genesis}>;
```

- Flatten the Api object structure so beacon routes are not nested. This is necessary for the type tricks used through the routes definitions.

before
```ts
api.beacon.state.getFork()
api.beacon.pool.submitAttestations()
```
after
```ts
api.beacon.getStateFork()
api.beacon.submitPoolAttestations()
```

- All types that exclusive to the API have been removed from lodestar-types and exist now only in that route namespace definition file.

before
```ts
import {SyncDuty} from "@chainsafe/lodestar-types";
const duty: SyncDuty = ...
```
after
```ts
import {routes} from "@chainsafe/lodestar-api";
const duty: routes.validator.SyncDuty = ...
```


packages/beacon-state-transition/src/allForks/util/epochContext.ts
```diff
@@ -24,12 +24,25 @@ export type EpochContextOpts = {
   skipSyncPubkeys?: boolean;
 };
 
-export class PubkeyIndexMap extends Map<ByteVector, ValidatorIndex> {
-  get(key: ByteVector): ValidatorIndex | undefined {
-    return super.get((toHexString(key) as unknown) as ByteVector);
+type PubkeyHex = string;
+
+function toHexStringMaybe(hex: ByteVector | string): string {
```
dapplion, line 29: Allow to pass string as hex string to prevent unnecessary parsing in the API

packages/lodestar/src/sync/sync.ts
```diff
@@ -66,21 +66,21 @@ export class BeaconSync implements IBeaconSync {
...
-          syncDistance: BigInt(currentSlot - headSlot),
+          headSlot: headSlot,
+          syncDistance: currentSlot - headSlot,
         };
       case SyncState.Synced:
         return {
-          headSlot: BigInt(headSlot),
-          syncDistance: BigInt(0),
+          headSlot: headSlot,
+          syncDistance: 0,
```
dapplion, line 83: No need to return a slot value as BigInt, return as `Slot`

wemeetagain (approved):
LGTM, we discussed the details offline.
Long overdue update :pray:

Issue #2556, opened 2021-05-24
BlocksByRange under WS


twoeths:
**Is your feature request related to a problem? Please describe.**

+ Implement this https://github.com/ethereum/eth2.0-specs/pull/2131/files
+ Part of #2530
+ Since it's quite isolated I make it a separate issue to avoid pulling us from being compliant to the whole spec

**Detail**
1. Since we only need to serve blocks within `MIN_EPOCHS_FOR_BLOCK_REQUESTS`  epochs, we can prune our db accordingly in order to avoid having our db grow forever
2. Validate `beacon_blocks_by_range` request accordingly
3. We have to backfill blocks if starting from a ws checkpoint
+ Don't need to get through state transition, just validate block proposer and parent-child relationship until recent block roots inside the ws state
+ Consider hooking into / add new sync type  https://github.com/ChainSafe/lodestar/blob/074649e35fe6a20a5aeab8898b67b73dca2e2900/packages/lodestar/src/sync/range/range.ts#L112

dapplion:
This new sync type can be actually fully independent from the other sync types. A node can be syncing forward at the same time it's backfilling.

I would create a new class BackfillSync that re-uses the `SyncChain` class which contains all the logic to download in parallel, load balance peers, etc. But with custom instantiation parameters

```ts
processChainSegment: (blocks) => {
  verifyParentChildSeg(blocks)
  db.blocks.put(blocks)
}
```

Also override the computeTarget function too to make it static and equal to the ws checkpoint

g11tech:
some conversation  context:

![image](https://user-images.githubusercontent.com/76567250/135662917-021c55db-ce9a-4ca0-a8fb-4b333a63f09a.png)
![image](https://user-images.githubusercontent.com/76567250/135663009-0bb1b48b-e9c5-4e6c-9d20-d96c3389d9c4.png)
![image](https://user-images.githubusercontent.com/76567250/135663074-823dcdd2-d687-4973-9500-82fecc1072f8.png)

PR #2571, opened 2021-05-27
Weak subjectivity chain initialization

3xtr4t3rr3str14l:
Alternative to the IPFS setup for weak subjectivity CLI where we query Infura instead of IPFS.

Implements `isWithinWeakSubjectivityPeriod` and as a consequence, `computeWeakSubjectivityPeriod` (as spec'd here: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/weak-subjectivity.md#is_within_weak_subjectivity_period) to verify the fetched weak subjectivity state (which also handles the state root verification).

Corrects a flaw in the original ws spec of `is_within_weak_subjectivity_period` (via our implementation) by checking block roots.

TODO (after merge):
Make a pull request on https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/weak-subjectivity.md to tell them about the flaw in `is_within_weak_subjectivity_period`.


docker-compose.yml
```diff
@@ -11,7 +11,7 @@ services:
     ports:
       - "9000:9000" # P2P port
       - "9596:9596" # REST API port
-    command: beacon --rootDir /data --api.rest.host 0.0.0.0 --metrics.enabled --logFile /logs/beacon.log --logLevelFile debug
+    command: beacon --rootDir /data --api.rest.host 0.0.0.0 --metrics.enabled --logFile /logs/beacon.log --logger.api.level debug --api.rest.api debug beacon config node validator events lodestar
```
dapplion: We must not enable the debug API by default

twoeths:
> Still looking into it, but every time I try to access state.latestBlockHeader.stateRoot where state is a WS state, it tells me the stateRoot is 0x0000...00 every time. This may be due to needing #2556 implemented (or the block backfilling part, rather), but I'm still not sure yet.

it's expected bc we only know state root after we finish the state transition https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/beacon-chain.md#block-header


due to that this code snippet is quite common https://github.com/ChainSafe/lodestar/blob/ee75995812cbe60b4056b6800e34eea03c2c052b/packages/lodestar/src/chain/blocks/stateTransition.ts#L138

packages/beacon-state-transition/src/allForks/util/weakSubjectivity.ts
```diff
@@ -15,3 +30,62 @@ export function getLatestWeakSubjectivityCheckpointEpoch(
...
+    wsPeriod +=
+      epochsForValidatorSetChurn > epochsForBalanceTopUps ? epochsForValidatorSetChurn : epochsForBalanceTopUps;
+  } else {
+    wsPeriod += Math.floor((3 * N * D * t) / (200 * Delta * (T - t)));
+  }
+  return BigInt(wsPeriod);
+}
+
+function getLatestBlockRoot(config: IBeaconConfig, state: allForks.BeaconState): Root {
+  const header = config.types.phase0.BeaconBlockHeader.clone(state.latestBlockHeader);
```
mpetrunic: why clone header?

packages/beacon-state-transition/src/allForks/util/weakSubjectivity.ts
```diff
@@ -15,3 +30,62 @@ export function getLatestWeakSubjectivityCheckpointEpoch(
...
+  wsState: allForks.BeaconState,
+  wsCheckpoint: Checkpoint
+): boolean {
+  const blockRoot = getLatestBlockRoot(config, wsState);
+  if (!config.types.Root.equals(blockRoot, wsCheckpoint.root)) {
+    throw new Error(
+      `Roots do not match.  expected=${toHexString(wsCheckpoint.root)}, actual=${toHexString(blockRoot)}`
+    );
+  }
+  assert.equal(computeEpochAtSlot(config, wsState.slot), wsCheckpoint.epoch);
```
mpetrunic: you should probably add third param message

packages/beacon-state-transition/src/allForks/util/weakSubjectivity.ts
```diff
@@ -15,3 +30,62 @@ export function getLatestWeakSubjectivityCheckpointEpoch(
...
+    const epochsForValidatorSetChurn = Math.floor(
+      (N * (t * (200 + 12 * D) - T * (200 + 3 * D))) / (600 * delta * (2 * t + T))
+    );
+    const epochsForBalanceTopUps = Math.floor((N * (200 + 3 * D)) / (600 * Delta));
+    wsPeriod +=
+      epochsForValidatorSetChurn > epochsForBalanceTopUps ? epochsForValidatorSetChurn : epochsForBalanceTopUps;
+  } else {
+    wsPeriod += Math.floor((3 * N * D * t) / (200 * Delta * (T - t)));
+  }
+  return BigInt(wsPeriod);
```
mpetrunic: why is wsPeriod bigint?

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -13,10 +16,59 @@ import {
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ws_state: any;
+};
+
+type WeakSubjectivityData = {
+  state: TreeBacked<allForks.BeaconState>;
+  checkpoint: Checkpoint;
+};
+
+async function getWeakSubjectivityData(
```
mpetrunic: I would move getting weak subjectivity data to separate file

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -13,10 +16,59 @@ import {
...
 import {getGenesisFileUrl} from "../../networks";
+import {WeakSubjectivityServers} from "../weakSubjectivityState";
+
+type WSResponse = {
+  current_epoch: number;
+  ws_checkpoint: string;
+  ws_period: number;
+  is_safe: boolean;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ws_state: any;
```
mpetrunic:
```suggestion
  ws_state: Json;
```
?

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -13,10 +16,59 @@ import {
...
+async function getWeakSubjectivityData(
+  config: IBeaconConfig,
+  args: IBeaconArgs & IGlobalArgs,
+  server: string,
+  logger: ILogger
+): Promise<WeakSubjectivityData> {
+  logger.info("Fetching weak subjectivity state from ChainSafe at " + server);
+  const response = await got(server, {searchParams: {checkpoint: args.weakSubjectivityCheckpoint}});
+  const responseBody = JSON.parse(response.body) as WSResponse;
+  const data = responseBody.ws_state.data;
```
mpetrunic: can ws_state be null? If yess you should add null checkhere

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -13,10 +16,59 @@ import {
...
+  const state = config.getForkTypes(data.slot).BeaconState.createTreeBackedFromJson(data, {case: "snake"});
+  if (!state) {
+    throw new Error("Weak subjectivity state not found for network " + args.network);
+  }
+  const checkpointData = (args.weakSubjectivityCheckpoint || responseBody.ws_checkpoint).split(":");
```
mpetrunic:
```suggestion
  const [checkpointRoot, checkpointEpoch] = (args.weakSubjectivityCheckpoint || responseBody.ws_checkpoint).split(":");
```

packages/cli/src/cmds/beacon/options.ts
```diff
@@ -8,6 +8,8 @@ interface IBeaconExtraArgs {
   forceGenesis?: boolean;
   genesisStateFile?: string;
   weakSubjectivityStateFile?: string;
+  fetchChainSafeWeakSubjecitivtyState: boolean;
+  weakSubjectivityCheckpoint?: string;
```
mpetrunic, line 11: you should allow passing custom server url and also check if it returns expected response

packages/cli/src/cmds/weakSubjectivityState.ts
```diff
@@ -4,5 +4,5 @@ export enum WeakSubjectivityServers {
   prater = "http://localhost:8081",
   pyrmont = "http://localhost:8081",
   dev = "http://localhost:8081",
-  coorparoo = "http://localhost:8081",
```
dapplion: What are these servers / services?

packages/beacon-state-transition/src/allForks/util/weakSubjectivity.ts
dapplion: Because it may be mutated below at line 68

packages/beacon-state-transition/src/allForks/util/weakSubjectivity.ts
dapplion: Yes please :pray: or just throw an error. It inconsistent to use assert here when you are throwing the error yourself above

packages/beacon-state-transition/src/allForks/util/weakSubjectivity.ts
mpetrunic: completely missed that

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -13,10 +16,59 @@ import {
...
+  const checkpointRoot = checkpointData[0];
+  const checkpointEpoch = checkpointData[1];
+  const checkpoint = {root: fromHex(checkpointRoot), epoch: parseInt(checkpointEpoch)};
+  return {state, checkpoint};
+}
+
+async function initAndVerifyWeakSujectivityState(
```
dapplion: This is very weak verification, you can do much more

packages/cli/src/cmds/beacon/options.ts
```diff
@@ -25,6 +27,19 @@ export const beaconExtraOptions: ICliCommandOptions<IBeaconExtraArgs> = {
...
+  fetchChainSafeWeakSubjecitivtyState: {
+    description:
+      "Tell the beacon node to fetch the latest weak subjectivity state for the given network from the ChainSafe weak subjectivity state server.",
+    type: "boolean",
+    default: false,
+  },
+
+  weakSubjectivityCheckpoint: {
+    description:
+      "Tell the beacon node to fetch a weak subjectivity state at the specified checkpoint.  The string arg must be in the form <blockRoot>:<epoch>.  For example, 0x1234:100 would ask for the weak subjectivity state at checkpoint of epoch 100 with block root 0x1234.",
```
dapplion: Remove double spaces

packages/cli/src/cmds/weakSubjectivityState.ts
```diff
@@ -0,0 +1,8 @@
+// TODO put the real server names/IPs once we have them
+export enum WeakSubjectivityServers {
+  mainnet = "http://localhost:8081",
+  prater = "http://localhost:8081",
+  pyrmont = "http://localhost:8081",
+  dev = "http://localhost:8081",
+  oonoonba = "http://localhost:8081",
```
dapplion: What are this servers / services? What software are they expected to run? Until this URLs are changed for something reachable publicly the feature won't work.

packages/beacon-state-transition/src/allForks/util/weakSubjectivity.ts
3xtr4t3rr3str14l: Because that's what the spec calls for:
https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/weak-subjectivity.md#compute_weak_subjectivity_period

packages/cli/src/cmds/beacon/initBeaconState.ts
3xtr4t3rr3str14l: This ultimately comes from calling the `debug/states/<stateId>` endpoint, so if the stateId is not resolved, the API will throw an error when we try to fetch it.

packages/cli/src/cmds/weakSubjectivityState.ts
3xtr4t3rr3str14l: These would be addresses for servers that would be running this code (assuming we decide to go with this scheme):
https://github.com/ChainSafe/eth2-ws-provider/pull/1

packages/cli/src/cmds/beacon/initBeaconState.ts
3xtr4t3rr3str14l: What else would you like to see here?  The `isWithinWeakSubjectivityPeriod` function also does some root and epoch verification (i.e. it does more than just check that the state is within the ws period).  That's just what they call the function in the spec.

packages/cli/src/cmds/weakSubjectivityState.ts
3xtr4t3rr3str14l: These would be addresses for servers that would be running this code (assuming we decide to go with this scheme):
ChainSafe/eth2-ws-provider#1

> Until this URLs are changed for something reachable publicly the feature won't work.

That's why I put the `TODO` at the top of this list.  This item depends on consensus from the server code (e.g. at `eth2-ws-provider`)

3xtr4t3rr3str14l: I'll add a note in the description to say that we shouldn't merge until we arrive at consensus for the server code (for any would-be mergers of this PR).

packages/beacon-state-transition/src/allForks/util/weakSubjectivity.ts
mpetrunic: true, but they have everything as uint64, I think it safe to say, that this period will never be more than native javascript number

packages/cli/src/cmds/weakSubjectivityState.ts
3xtr4t3rr3str14l: These are now Infura endpoints instead, as per our discussion.

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -46,6 +62,23 @@ export async function initBeaconState(
     return await initFromFile(args.weakSubjectivityStateFile);
   } else if (dbHasSomeState) {
     return await initStateFromDb(config, db, logger);
+  } else if (args.fetchChainSafeWeakSubjecitivtyState && args.network !== "dev" && args.network !== "oonoonba") {
```
wemeetagain: probably should have a more robust check against a supported network, maybe throw with a helpful message if not.
eg: `Object.keys(WeakSubjectivityServers).includes(args.network) || args.weakSubjectivityServerUrl`

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -17,6 +17,22 @@ import {downloadOrLoadFile} from "../../util";
...
+import {SLOTS_PER_EPOCH} from "@chainsafe/lodestar-params";
+
+async function initAndVerifyWeakSujectivityState(
+  config: IBeaconConfig,
+  db: IBeaconDb,
+  wsState: TreeBacked<allForks.BeaconState>,
+  logger: ILogger,
+  wsCheckpoint?: Checkpoint
+): Promise<TreeBacked<allForks.BeaconState>> {
+  if (!isWithinWeakSubjectivityPeriod(config, wsState.genesisTime, wsState, wsCheckpoint)) {
```
wemeetagain: It looks like this function can throw, error needs to be handled.

packages/beacon-state-transition/src/allForks/util/weakSubjectivity.ts
```diff
@@ -15,3 +31,69 @@ export function getLatestWeakSubjectivityCheckpointEpoch(
...
+    header.stateRoot = config.getForkTypes(state.slot).BeaconState.hashTreeRoot(state);
+  }
+  return ssz.phase0.BeaconBlockHeader.hashTreeRoot(header);
+}
+
+export function isWithinWeakSubjectivityPeriod(
+  config: IBeaconConfig,
+  genesisTime: number,
+  wsState: allForks.BeaconState,
+  wsCheckpoint?: Checkpoint
```
wemeetagain: How come this is optional? Are we using that optionality somewhere? Just noticing its not optional in the spec.

packages/beacon-state-transition/src/allForks/util/weakSubjectivity.ts
```diff
@@ -15,3 +31,69 @@ export function getLatestWeakSubjectivityCheckpointEpoch(
...
+    const epochsForBalanceTopUps = Math.floor((N * (200 + 3 * D)) / (600 * Delta));
+    wsPeriod +=
+      epochsForValidatorSetChurn > epochsForBalanceTopUps ? epochsForValidatorSetChurn : epochsForBalanceTopUps;
+  } else {
+    wsPeriod += Math.floor((3 * N * D * t) / (200 * Delta * (T - t)));
+  }
+  return wsPeriod;
+}
+
+function getLatestBlockRoot(config: IBeaconConfig, state: allForks.BeaconState): Root {
```
wemeetagain: Can you export this fn? It may be useful in other places

packages/cli/src/cmds/weakSubjectivityState.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export async function getWeakSubjectivityState(
+  config: IBeaconConfig,
+  args: IBeaconArgs & IGlobalArgs,
+  stateId = "finalized",
+  server: string,
+  logger: ILogger
+): Promise<TreeBacked<allForks.BeaconState>> {
+  logger.info("Fetching weak subjectivity state from ChainSafe at " + server + "/" + stateId);
```
wemeetagain: remove "ChainSafe at "

packages/cli/src/cmds/beacon/initBeaconState.ts
wemeetagain: I think we need things from point 2 here: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/weak-subjectivity.md#weak-subjectivity-sync-procedure

In the spec where it says `store`, it means the latest state stored in our state archive (which may or may not exist).

For the first bullet point:
> IF epoch_number > store.finalized_checkpoint.epoch, then ASSERT during block sync that block with root block_root is in the sync path at epoch epoch_number. Emit descriptive critical error if this assert fails, then exit client process.

We will probably need a new option to pass to the beacon node on init so that during backfill, this check can happen.

The second bullet point:
> IF epoch_number <= store.finalized_checkpoint.epoch, then ASSERT that the block in the canonical chain at epoch epoch_number has root block_root. Emit descriptive critical error if this assert fails, then exit client process.

That can be checked here.

packages/cli/src/cmds/beacon/initBeaconState.ts
wemeetagain: This check should use the latest stored state's genesisTime, if it exists.

wemeetagain (changes requested):
Looking pretty good, left a few comments on minor things

packages/cli/src/cmds/beacon/initBeaconState.ts
3xtr4t3rr3str14l: I don't think there would be a case where we would have a stored state since we are only getting the weak subjectivity state in the first place if we don't have a state already stored.

packages/beacon-state-transition/src/allForks/util/weakSubjectivity.ts
3xtr4t3rr3str14l: It's optional because in the case that the user doesn't provide a wsCheckpoint, we are implicitly trusting the state that we get from Infura.  This may not be ideal, but if the user does not provide us a block root + epoch, we need to fetch the data for the latest block root + epoch from somewhere (not sure where, at the moment...maybe beaconscan?  but then we are offloading trust to beaconscan) because it seems like it would be pointless to verify the wsState against data that is inside the wsState (i.e. the checkpoint inside the wsState) at least inasmuch as we are trying to reinforce some security on the fetched Infura state.

packages/cli/src/cmds/beacon/initBeaconState.ts
3xtr4t3rr3str14l: Again, `isWithinWeakSubjectivityPeriod` does those checks here:

https://github.com/ChainSafe/lodestar/pull/2571/files/3e3a81a2818e81929d9cbf1a471ef32713ccdce4#diff-51a07be7e62b602a2219ab9614f4e88b9fb63ab0d590886478c6504120349edeR73

However, as discussed in my reply to your other comment, it does not use the `store` because, as it currently stands, we implicitly don't have an existing store when fetching from wsState because if we have something stored in the db, then we init from that and not from a wsState.

3xtr4t3rr3str14l: @wemeetagain  Maybe we can just fetch from the db and then if we also have the `fetchWS` flag enabled, then just use the db from state as the `store` for `isWithinWeakSubjectivityPeriod`?

wemeetagain: Oh, I see.
I think we'll want to allow ws sync even when a state exists in the db. The thing is, the long-range attack can happen whether or not the user already has a stored state or not. So if the user explicitly enables ws sync, it should be able to fetch/use the latest ws state.

As far as the cli goes, we can have:
- `wsStateFile`: string, default to undefined - if set, use this state for ws sync
- `wsCheckpoint`: string, default to undefined - if set, use this checkpoint for safety checks and (possibly) for fetching the ws state
- `wsServerUrl`: string, default to undefined - if undefined and ws sync is enabled and a state file is not provided (whether the user provides a ws checkpoint or not), use our provided infura url
- `wsSyncLatest`: boolean, default to false - if true, and `wsStateFile` and `wsCheckpoint` are not defined, fetch the latest ws checkpoint from infura, use that for safety check and fetching the ws state

so if `wsStateFile`, `wsCheckpoint`, or `wsSyncLatest` are set, the chain will be initialized with a ws state as the anchor state. Depending on how the db is populated, additional safety checks would be performed.

if `wsCheckpoint` or `wsSyncLatest` are set, `wsServerUrl` is used to fetch a ws state.

if `wsStateFile` and `wsCheckpoint` are set, `wsCheckpoint` is used for safety checks even tho no ws state is fetched via url.

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -17,6 +17,26 @@ import {downloadOrLoadFile} from "../../util";
...
+import {SLOTS_PER_EPOCH} from "@chainsafe/lodestar-params";
+
+async function initAndVerifyWeakSujectivityState(
+  config: IBeaconConfig,
+  db: IBeaconDb,
+  wsState: TreeBacked<allForks.BeaconState>,
+  logger: ILogger,
+  wsCheckpoint?: Checkpoint
+): Promise<TreeBacked<allForks.BeaconState>> {
+  try {
```
wemeetagain: I think I messed up here, the error was already being "handled" by just being thrown. My bad :sweat_smile:

packages/beacon-state-transition/src/allForks/util/weakSubjectivity.ts
wemeetagain: I think we should keep the param as non-optional just to simplify the code. If no wsCheckpoint is provided via cli, we can just create a checkpoint from the ws state to pass here and to perform the safety checks.

packages/cli/src/cmds/beacon/initBeaconState.ts
3xtr4t3rr3str14l: @wemeetagain pushed up this ^

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -17,6 +20,29 @@ import {downloadOrLoadFile} from "../../util";
...
+import {Checkpoint} from "@chainsafe/lodestar-types/phase0";
+import {SLOTS_PER_EPOCH} from "@chainsafe/lodestar-params";
+import {computeEpochAtSlot} from "../../../../beacon-state-transition/lib";
+
+function getCheckpointFromArg(checkpointStr: string): Checkpoint {
+  const wsCheckpointData = checkpointStr.split(":");
+  return {root: fromHex(wsCheckpointData[0]), epoch: parseInt(wsCheckpointData[1])};
+}
+
+async function initAndVerifyWeakSujectivityState(
```
wemeetagain: Typo in function name

wemeetagain (dismissed):

dapplion (dismissed):
Remove file `packages/beacon-state-transition/state1.ssz`

dapplion (approved):

wemeetagain (approved):

PR #2574, opened 2021-05-27
Fetch fork in validatorStore

### LINKED ISSUE #2573: Check Validator + BeaconNode Config

In a discussion w @dapplion:
> Oh, the node is running minimal, and the validator is running mainnet

We'd like to prevent that scenario from happening.

Add a check in the validator to ensure that the validator is connecting to a beacon node w the same configuration.
Eg: fetch from the `config.getSpec` endpoint and compare against local values. Throw if the beacon node isn't configured the same as the validator.

dapplion:
**Motivation**

Reverts a critical bug introduced in https://github.com/ChainSafe/lodestar/pull/2529/files#diff-f315d8cdde0f9f1d3f00aab8ea4d888c125bd20ad3813633535e9f971764f38bL198 where if the validator no longer fetches the fork version from the beacon node. If the beacon node is running in a non-standard fork and the validator is not properly configured all signatures will be rejected.

**Description**

Fetch the fork in the validatorStore

Closes https://github.com/ChainSafe/lodestar/issues/2573

### DIFF
--- a/packages/validator/src/services/validatorStore.ts
+++ b/packages/validator/src/services/validatorStore.ts
@@ -1,10 +1,5 @@
 import {SecretKey} from "@chainsafe/bls";
-import {
-  computeDomain,
-  computeEpochAtSlot,
-  computeSigningRoot,
-  computeStartSlotAtEpoch,
-} from "@chainsafe/lodestar-beacon-state-transition";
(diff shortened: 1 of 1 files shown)


wemeetagain:
Related to #2573

wemeetagain (approved):

Issue #2542, opened 2021-05-20
Multi-thread CI seg-fault


dapplion:
Our multi-thread tests are failing very often. This issue is a tracker to aggregate the errors and hopefully figure out what's going on. I have no idea

May 20 2021 *Simulation multi thread multi node test altair epoch genesis* https://github.com/ChainSafe/lodestar/pull/2541/checks?check_run_id=2629966394   
```
node: ../binding.cc:545: Iterator::~Iterator(): Assertion `ended_' failed.
Aborted (core dumped)
error Command failed with exit code 134.
```

---

May 20 2021 *Simulation multi thread multi node test altair epoch genesis* https://github.com/ChainSafe/lodestar/pull/2540/checks?check_run_id=2629455362
```
Success: Terminating workers
FATAL ERROR: v8::FromJust Maybe value is Nothing.
 1: 0xa04200 node::Abort() [/opt/hostedtoolcache/node/14.16.0/x64/bin/node]
 2: 0x94e4e9 node::FatalError(char const*, char const*) [/opt/hostedtoolcache/node/14.16.0/x64/bin/node]
 3: 0xb794aa v8::Utils::ReportApiFailure(char const*, char const*) [/opt/hostedtoolcache/node/14.16.0/x64/bin/node]
 4: 0xab5e65 node::worker::Worker::JoinThread() [/opt/hostedtoolcache/node/14.16.0/x64/bin/node]
 5: 0x9a4298 node::Environment::RunAndClearNativeImmediates(bool) [/opt/hostedtoolcache/node/14.16.0/x64/bin/node]
 6: 0x137bb06  [/opt/hostedtoolcache/node/14.16.0/x64/bin/node]
 7: 0x138e5e5  [/opt/hostedtoolcache/node/14.16.0/x64/bin/node]
 8: 0x137c438 uv_run [/opt/hostedtoolcache/node/14.16.0/x64/bin/node]
 9: 0xabc4a3 node::worker::Worker::Run() [/opt/hostedtoolcache/node/14.16.0/x64/bin/node]
10: 0xabcf28  [/opt/hostedtoolcache/node/14.16.0/x64/bin/node]
11: 0x7f8b1276c609  [/lib/x86_64-linux-gnu/libpthread.so.0]
12: 0x7f8b12693293 clone [/lib/x86_64-linux-gnu/libc.so.6]
Aborted (core dumped)
error Command failed with exit code 134.
```
```
[...]

PR #2575, opened 2021-05-27
Run multi-thread tests in spawned processes

dapplion:
**Motivation**

Our CI sim tests fail at a >90% rate due to a seg-fault see https://github.com/ChainSafe/lodestar/issues/2542. Running the multiple nodes in separate processes instead of separate threads may solve the issue.

**Description**

Run each node and validator in a separate spawned process


dapplion:
- Tracked in https://github.com/ChainSafe/lodestar/issues/4275

**Do not delete branch**

PR #2590, opened 2021-05-30
List all API namespaces in CLI args

dapplion:
**Motivation**

CLI `api.rest.api` option was not in sync with new namespaces.

**Description**

This approach enforces with Typescript that all existing yargs will always be aware of all existing API namespaces.

### DIFF
--- a/packages/api/src/interface.ts
+++ b/packages/api/src/interface.ts
@@ -17,3 +17,17 @@ export type Api = {
   node: NodeApi;
   validator: ValidatorApi;
 };
+
+// Declare namespaces for CLI options
+export type ApiNamespace = keyof Api;
+const allNamespacesObj: {[K in keyof Api]: true} = {
+  beacon: true,
+  config: true,
+  debug: true,
+  events: true,
+  lightclient: true,
+  lodestar: true,
+  node: true,
+  validator: true,
+};
+export const allNamespaces = Object.keys(allNamespacesObj) as ApiNamespace[];
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -1,5 +1,5 @@
 export * as routes from "./routes";
-export {Api} from "./interface";
+export * from "./interface";
 export {getClient} from "./client";
 
 // Node: Don't export server here so it's not bundled to all consumers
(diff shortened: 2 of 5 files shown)


mpetrunic (commented):
Maybe it would be better to introduce ApiNamespace enum, which can be single source of truth and also enforce proper types here https://github.com/ChainSafe/lodestar/blob/master/packages/api/src/interface.ts#L10

dapplion:
> Maybe it would be better to introduce ApiNamespace enum, which can be single source of truth and also enforce proper types here https://github.com/ChainSafe/lodestar/blob/master/packages/api/src/interface.ts#L10

You can't derive an enum from object keys. The goal of this approach is that the ApiNamespaces derive from the shape of the Api interface. Then Typescript enforces that the namespace matches exactly the key with it associated routes group.

Using an enum is the previous approach. And we forgot all the time to declare new namespaces. Here the code won't compile if you forget.

mpetrunic:
> > Maybe it would be better to introduce ApiNamespace enum, which can be single source of truth and also enforce proper types here https://github.com/ChainSafe/lodestar/blob/master/packages/api/src/interface.ts#L10
> 
> You can't derive an enum from object keys. The goal of this approach is that the ApiNamespaces derive from the shape of the Api interface. Then Typescript enforces that the namespace matches exactly the key with it associated routes group.
> 
> Using an enum is the previous approach. And we forgot all the time to declare new namespaces. Here the code won't compile if you forget.

Hmm but if you declare Api as
```
export type Api : Record<ApiNamespaceEnum, something> = {
  ...
}

and

const allNamespacesObj: {[K in keyof Api]: true} = Object.values(ApiNamespaceEnum).reduce((o,n) => {o[n]=true}, {})
```
everything should be typesafe and enforced while we also have single source of truth(enum), right?

dapplion:
> everything should be typesafe and enforced while we also have single source of truth(enum), right?

Right but with that approach you can't associate different types to each property. All properties of `Api` would have the same routes which is not what we want.

packages/api/src/interface.ts
```diff
@@ -17,3 +17,17 @@ export type Api = {
   node: NodeApi;
   validator: ValidatorApi;
 };
+
+// Declare namespaces for CLI options
+export type ApiNamespace = keyof Api;
+const allNamespacesObj: {[K in keyof Api]: true} = {
```
wemeetagain, line 23: Annoying that we had to do this, but I understand why. `allNamespacesObj`, and its `K in keyof...` type is what enforces the connection between the namespace list as a type (`Api`), and the namespace list as a javascript object/array.

wemeetagain (approved):

PR #2599, opened 2021-05-31
Review gossip validation

dapplion:
**Motivation**

Currently our node is able to process only 0-20% of the the attestations received through gossip. Reviewing our gossip validation I found many inneficiencies:
- Checked badBlock twice
- badBlock check involves a disk read, should be cached in memory
- Committee calculations seem like could be optimized further

**Description**

- Added the exact spec directive as a comment associated with each validation. This is extremely useful to audit the code quickly against the spec. Will be very useful if the gossip checks evolve.
- Re-ordered the gossip validation checks to match the spec, see above.

### DIFF
--- a/packages/lodestar/src/chain/validation/aggregateAndProof.ts
+++ b/packages/lodestar/src/chain/validation/aggregateAndProof.ts
@@ -5,13 +5,11 @@ import {IBeaconDb} from "../../db";
 import {
   phase0,
   allForks,
-  CachedBeaconState,
   computeEpochAtSlot,
   isAggregatorFromCommitteeLength,
 } from "@chainsafe/lodestar-beacon-state-transition";
-import {isAttestingToInValidBlock} from "./attestation";
 import {getSelectionProofSignatureSet, getAggregateAndProofSignatureSet} from "./signatureSets";
-import {AttestationError, AttestationErrorCode} from "../errors";
+import {AttestationGossipError, AttestationErrorCode, GossipAction} from "../errors";
 import {ATTESTATION_PROPAGATION_SLOT_RANGE} from "../../constants";
 
 export async function validateGossipAggregateAndProof(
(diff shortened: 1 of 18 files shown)


packages/lodestar/src/chain/validation/attestation.ts
```diff
@@ -14,176 +14,157 @@ export async function validateGossipAttestation(
...
   }
 
-  if (!config.types.Epoch.equals(attestation.data.target.epoch, computeEpochAtSlot(config, attestationSlot))) {
-    throw new AttestationError({
-      code: AttestationErrorCode.BAD_TARGET_EPOCH,
+  // [IGNORE] There has been no other valid attestation seen on an attestation subnet that has an
+  // identical attestation.data.target.epoch and participating validator index.
+  //
+  // no other validator attestation for same target epoch has been seen
+  if (db.seenAttestationCache.hasCommitteeAttestation(attestation)) {
```
twoeths, line 114: there are a lot of duplicate attestations, should this one be checked first?

packages/lodestar/src/chain/validation/attestation.ts
```diff
@@ -14,176 +14,157 @@ export async function validateGossipAttestation(
   subnet: number
 ): Promise<void> {
   const attestation = attestationJob.attestation;
-  const numBits = getAttestationAttesterCount(attestation);
```
twoeths, line 17: it's good to remove this manual `aggregationBits` read 👍

packages/lodestar/src/chain/validation/aggregateAndProof.ts
```diff
@@ -24,126 +22,124 @@ export async function validateGossipAggregateAndProof(
...
-  let attestationTargetState: CachedBeaconState<allForks.BeaconState>;
+  let committee: ValidatorIndex[];
   try {
-    // the target state, advanced to the attestation slot
-    attestationTargetState = await chain.regen.getCheckpointState(attestation.data.target);
-  } catch (e) {
-    throw new AttestationError({
-      code: AttestationErrorCode.MISSING_ATTESTATION_TARGET_STATE,
-      error: e as Error,
+    committee = attestationTargetState.getBeaconCommittee(aggregate.data.slot, aggregate.data.index);
```
twoeths, line 85: not important but the `aggregateAndProof` is a TreeBacked here so we want to get each property (for example `aggregate.data`) only one time

packages/lodestar/src/chain/validation/aggregateAndProof.ts
```diff
@@ -24,126 +22,124 @@ export async function validateGossipAggregateAndProof(
...
+  // [REJECT] The current finalized_checkpoint is an ancestor of the block defined by aggregate.data.beacon_block_root
+  // -- i.e. get_ancestor(store, aggregate.data.beacon_block_root, compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)) == store.finalized_checkpoint.root
 
+  // [REJECT] The aggregate_and_proof.selection_proof is a valid signature of the aggregate.data.slot
+  // by the validator with index aggregate_and_proof.aggregator_index.
+  // [REJECT] The aggregator signature, signed_aggregate_and_proof.signature, is valid.
+  // [REJECT] The signature of aggregate is valid.
+  const slot = aggregate.data.slot;
+  const epoch = computeEpochAtSlot(config, slot);
+  const aggregator = attestationTargetState.index2pubkey[aggregateAndProof.aggregatorIndex];
```
twoeths, line 131: similarly maybe assign a separate `aggregatorIndex` variable to avoid duplicate `aggregateAndProof.aggregatorIndex`

not sure how significant it is but for TreeBacked, we store everything in byte array and a simple `aggregateAndProof.aggregatorIndex` requires us to extract byte array to number, similarly `state.slot` or `state.currentFinalizedCheckpoint.epoch` etc.

dapplion:
I also think we have an issue with early attestations and blocks. We throw MISSING_PRESTATE error, add the attestation to the pending pool but then we never run the full gossip verification chain.

dapplion:
Closing for https://github.com/ChainSafe/lodestar/pull/2760

PR #2611, opened 2021-06-02
Fix bootfile parsing

wemeetagain:
**Description**
Fix bootfile parsing. Enrs must begin with "enr:"


twoeths (approved):

dapplion:
I was wondering why it won't connect to the bootnodes :sweat_smile:

PR #2617, opened 2021-06-02
Validate SyncCommitteeSignatureSet

### LINKED ISSUE #2610: Add SyncCommittee signature set in all block sigs
**Describe the bug**

We do not include the SyncCommittee signature in this all blocks sig bundle

https://github.com/ChainSafe/lodestar/blob/13bb12031c31a2fe2d1af9b5ff9c7436e914e5e7/packages/beacon-state-transition/src/allForks/signatureSets/index.ts#L33-L43

**Expected behavior**

We must verify all signatures in a block

dapplion:
**Motivation**

See https://github.com/ChainSafe/lodestar/issues/2610

**Description**

Closes https://github.com/ChainSafe/lodestar/issues/2610

### DIFF
--- a/packages/beacon-state-transition/src/altair/block/processSyncCommittee.ts
+++ b/packages/beacon-state-transition/src/altair/block/processSyncCommittee.ts
@@ -1,13 +1,13 @@
-import {verifyAggregate} from "@chainsafe/bls";
 import {altair} from "@chainsafe/lodestar-types";
-import {assert} from "@chainsafe/lodestar-utils";
-
 import {
   computeEpochAtSlot,
   computeSigningRoot,
   getBlockRootAtSlot,
   getDomain,
   increaseBalance,
+  ISignatureSet,
+  SignatureSetType,
+  verifySignatureSet,
   zipIndexesInBitList,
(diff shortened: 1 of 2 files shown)


twoeths (approved):

packages/beacon-state-transition/src/altair/block/processSyncCommittee.ts
```diff
@@ -1,59 +1,86 @@
...
+
+export function getSyncCommitteeSignatureSet(
+  state: CachedBeaconState<altair.BeaconState>,
+  syncAggregate: altair.SyncAggregate,
+  /** Optional parameter to prevent computing it twice */
+  participantIndices?: number[]
+): ISignatureSet | null {
+  const {config, epochCtx} = state;
+
+  const previousSlot = Math.max(state.slot, 1) - 1;
```
twoeths, line 48: @dapplion previousSlot should be from `Math.max(block.slot, 1) - 1`

dapplion: Good pick!

PR #2621, opened 2021-06-03
Benchmark workflow

twoeths:
**Motivation**

We want to track lodestar performance in Github page

**Description**

+ Rewrite performance tests with new runner
+ Write new benchmark workflow using https://github.com/rhysd/github-action-benchmark
+ In the end we want statistics to be uploaded to `gh-pages/dev/bench`

Also I move the flag 'OR` operation outside of the for loop in EpochProcess.


dapplion (approved):
Awesome job :fire: !

We need to be aware that the current approach will break if we do benchmarks in different packages of the monorepo since each will override the `benchmark-output.txt` file.

PR #2624, opened 2021-06-03
Fix signature verification at fork boundary

twoeths:
**Motivation**

Fix signature verification at fork boundary
**Description**

Closes #2623

**Steps to test or reproduce**

`./bin/lodestar beacon --network oonoonba`

PR #2625, opened 2021-06-03
Verify SyncCommittee signatures in batch

twoeths:
**Motivation**

A continuation of #2617

**Description**

+ when getting sync committee signature set, previous slot should be from the block instead of state to make sure it's correct
+ signed roots of sync committee signatures are only known after we run state transition for all blocks in segment
+ revert #2624 as it's not necessary anymore

+ `./bin/lodestar beacon --network oonoonba`

### DIFF
--- a/packages/beacon-state-transition/src/altair/block/processSyncCommittee.ts
+++ b/packages/beacon-state-transition/src/altair/block/processSyncCommittee.ts
@@ -23,7 +23,9 @@ export function processSyncCommittee(
 
   // different from the spec but not sure how to get through signature verification for default/empty SyncAggregate in the spec test
   if (verifySignatures) {
-    const signatureSet = getSyncCommitteeSignatureSet(state, syncAggregate, participantIndices);
+    // This is to conform to the spec - we want the signature to be verified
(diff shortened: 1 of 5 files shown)


twoeths:
synced fine now
Screen Shot 2021-06-03 at 16 52 01

packages/beacon-state-transition/src/altair/block/processSyncCommittee.ts
```diff
@@ -7,29 +7,18 @@ import {
...
   const participantIndices = getParticipantIndices(state, syncAggregate);
 
-  // different from the spec but not sure how to get through signature verification for default/empty SyncAggregate in the spec test
-  if (verifySignatures) {
-    const signatureSet = getSyncCommitteeSignatureSet(state, syncAggregate, participantIndices);
-    // When there's no participation we consider the signature valid and just ignore i
-    if (signatureSet !== null && !verifySignatureSet(signatureSet)) {
-      throw Error("Sync committee signature invalid");
-    }
-  }
```
dapplion: Please re-add signature verification

dapplion (approved):

PR #2633, opened 2021-06-04
Fix debug state api limitation

3xtr4t3rr3str14l:
There were cases when I was testing the weak subjectivity state fetching where I couldn't get a very recently finalized state by its slot.  The `debug/state` would not fetch the state if we couldn't find a block summary in the fork choice and it wouldn't trigger the state API's state regen functionality.  This PR extends the existing solution to regen a state regardless of whether a block summary is found.

### DIFF
--- a/packages/lodestar/src/api/impl/beacon/state/utils.ts
+++ b/packages/lodestar/src/api/impl/beacon/state/utils.ts
@@ -212,13 +212,13 @@ async function stateBySlot(
 ): Promise<allForks.BeaconState | null> {
   const blockSummary = forkChoice.getCanonicalBlockSummaryAtSlot(slot);
   if (blockSummary) {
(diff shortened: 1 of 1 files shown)


mpetrunic (commented):
I don't see any logic change here? Seems like just different syntax. If i'm not mistaken in old code is same logic, if block summary isn't found, it will regen state

dapplion:
> I don't see any logic change here? Seems like just different syntax. If i'm not mistaken in old code is same logic, if block summary isn't found, it will regen state

This change allows to reach the getFinalizedState() code path even if there is a blockSummary available

dapplion (approved):

mpetrunic (approved):

wemeetagain (approved):

PR #2637, opened 2021-06-04
Backfill sync

### LINKED ISSUE #2556: BlocksByRange under WS

mpetrunic:
resolves #2556


packages/lodestar/src/chain/chain.ts
```diff
@@ -70,6 +70,7 @@ export class BeaconChain implements IBeaconChain {
   private abortController = new AbortController();
 
   constructor({opts, config, db, logger, metrics, anchorState}: IBeaconChainModules) {
+    config.getForkTypes(0)
```
dapplion: Typo?

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,189 @@
...
+
+export enum BackfillSyncEvent {
+  completedChain = "BackfillSync-completedChain",
+}
+
+type BackfillSyncEvents = {
+  [BackfillSyncEvent.completedChain]: () => void;
+};
+
+type BackfillSyncEmitter = StrictEventEmitter<EventEmitter, BackfillSyncEvents>;
```
dapplion: If there's no consumer for this event I would remove the event emmiter code.

packages/lodestar/src/sync/backfill/verify.ts
```diff
@@ -0,0 +1,24 @@
...
+    if (
+      lastBlock &&
+      !config.types.Root.equals(
+        config.getForkTypes(lastBlock.message.slot).BeaconBlock.hashTreeRoot(lastBlock.message),
+        block.message.parentRoot
+      )
+    ) {
+      throw new Error("BackfillSync - Non linear blocks");
+    }
+    if (!verifyBlockSignature(config, state, block)) {
```
dapplion: We should use BLS batch verification to verify all proposer signatures in one go

dapplion: How can you retrieve the pubkey of the block proposer for a block that's pretty old? Do we keep the shuffling somewhere?

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,189 @@
...
+      this.close();
+    } else {
+      this.syncChain?.startSyncing(oldestRequiredEpoch);
+    }
+  }
+
+  private async addPeer(peerId: PeerId, peerStatus: phase0.Status): Promise<void> {
+    //Could be caveat if you start from genesis, than stop and continue from WS state
+    // Currently we are relying on block archive prune to delete old blocks after which this is going to work
+    const oldestStoredBlock = await this.db.blockArchive.firstValue();
```
dapplion: We don't really need to check this for every new peer, but once at startup and that's it.

dapplion:
I think it would be simpler to download them a verify them in reverse.

1. Check if we have ws checkpoint block root in db
2. if not, do a blocks_by_root request with ws checkpoint block root -> block0
3. do a blocks_by_range request for [block.slot - N, block.slot]
4. Verify linear block segment such that block-1 == block0.parentRoot and so on. + verify proposer signature
5. Repeat (3) until we have a required block in the db or backfill the required slots

I've been playing around with this for a bit and maybe we want to not re-use the existing range sync code since the requirements are very different. Unlike with current sync we don't need backfill sync to be fast, it just has to complete eventually. The feature of downloading batches in parallel is not useful here since speed is not a concern and we can potentially be doing more work than necessary.

Probably just doing the above 5 steps in an infinite retry and picking a random peer would suffice. We don't need to track status or emit any events, just log that we have to backfill, log errors and log on done.

EDIT: modified point 4 to add "+ verify proposer signature"

packages/lodestar/src/chain/chain.ts
mpetrunic: lol yeah

packages/lodestar/src/sync/backfill/verify.ts
mpetrunic: I don't think we can that's why I'm using this slow method with current state we have

mpetrunic:
> So to backfill sync we can't really do any verification on the blocks but ensure that all parents of our trusted weak subjectivity checkpoint. Therefore we must download them a verify them in reverse.

We need to verify proposer signature as it's not part of hash chain

> Check if we have ws checkpoint block root in db
Do we need it? Can't we check with state.lastestBlockHeader.parentRoot?

mpetrunic:
@dapplion does this flow looks better? No idea if I'm using processor as expected :)

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,182 @@
...
+        verifyBlockSignature(this.config, this.chain.getHeadState(), block);
+      } catch (e) {
+        throw new BackfillSyncError({code: BackfillSyncErrorCode.INVALID_SIGNATURE});
+      }
+
+      await this.db.blockArchive.put(block.message.slot, block);
+      this.anchorBlock = block;
+      this.processor.trigger();
+    } catch (e) {
+      this.peers.delete(peer);
```
dapplion: You should report the peer and expect the network to disconnect the peer eventually once it is banned. This approach may be too punishing and overlaps with the scoring system.

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,182 @@
...
+      setTimeout(this.processor.trigger, 2000);
```
dapplion: Why wait 2 seconds here?

packages/lodestar/src/sync/backfill/verify.ts
```diff
@@ -0,0 +1,37 @@
...
+        config.getForkTypes(block.message.slot).BeaconBlock.hashTreeRoot(block.message),
+        nextRoot
+      )
+    ) {
+      if (config.types.Root.equals(nextRoot, anchorRoot)) {
+        throw new BackfillSyncError({code: BackfillSyncErrorCode.NOT_ANCHORED});
+      }
+      throw new BackfillSyncError({code: BackfillSyncErrorCode.NOT_LINEAR});
+    }
+    signatures.push(allForks.getProposerSignatureSet(state, block));
```
dapplion: Should we gather the signatures after this for loop in a map? No need to do `ReturnType<typeof allForks["getProposerSignatureSet"]>[]` then

packages/lodestar/src/tasks/tasks/pruneBlockArchive.ts
```diff
@@ -0,0 +1,53 @@
...
+  db: IBeaconDb;
+  clock: IBeaconClock;
+  logger: ILogger;
+}
+
+/**
+ * Since we only need to serve MIN_EPOCHS_FOR_BLOCK_REQUESTS of history blocks.
+ * We can delete it periodically.
+ */
+export class PruneBlockArchiveTask implements ITask {
```
dapplion: For testnets we may want to keep all blocks for debugging purposes. Should we add a flag to prevent nuking the block db for everyone?

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,182 @@
...
+              this.logger.info("Backfill sync completed");
+              break;
+            }
+            await this.syncRange(fromSlot, toSlot, this.anchorBlock.message.parentRoot);
+          } catch (e) {
+            this.logger.debug("Error while backfiling by range", e);
+          }
+        } else {
+          try {
+            await this.syncBlock(this.chain.getHeadState().latestBlockHeader.parentRoot);
```
dapplion: We must backfill from the weak subjectivity point which becomes our finalized checkpoint. Shouldn't we get the finalized checkpoint instead of the head here?

packages/lodestar/src/sync/backfill/backfill.ts
mpetrunic: I don't think we really need this anymore. If we ran out of peers, adding new peer will trigger again

packages/lodestar/src/sync/backfill/verify.ts
mpetrunic: but bls.verifySignatureSets expects array. How does map prevents requiring type?
Better solution would be to export SignatureSet type but I assumed there was a reason why you didn't do that

packages/lodestar/src/tasks/tasks/pruneBlockArchive.ts
mpetrunic: yeah I think we should add flag, not just for testnet but for other consumers which might need to store all blocks

packages/lodestar/src/sync/backfill/backfill.ts
mpetrunic: Yeah, I think that makes more sense.

packages/lodestar/src/sync/backfill/verify.ts
dapplion:
```ts
const signatures = blocks.map(block => allForks.getProposerSignatureSet(state, block))
if (!(await bls.verifySignatureSets(signatures))) {
    throw new BackfillSyncError({code: BackfillSyncErrorCode.INVALID_SIGNATURE});
  }
```

packages/lodestar/src/constants/network.ts
```diff
@@ -34,9 +37,13 @@ export type RpcResponseStatusError = Exclude<RespStatus, RespStatus.SUCCESS>;
 
 /** The maximum allowed size of uncompressed gossip messages. */
 export const GOSSIP_MAX_SIZE = 2 ** 20;
+
+export function getMinEpochForBlockRequests(config: IBeaconConfig): number {
```
dapplion: Can you add a link with source for this?

dapplion:
Is the any blocker on this PR? Happy to help get it to a review-able state

mpetrunic:
> Is the any blocker on this PR? Happy to help get it to a review-able state

Got busy with db metrics, getting back to this. Hopefully I can get this working and figure some test for it

mpetrunic:
@dapplion this should be better now.

- no idea how to listen on backfill sync completed in test now that we removed events
- there is bunch of different errors when running test that I would appreciate help debugging

dapplion:
Merge conflicts

packages/lodestar/src/sync/backfill/errors.ts
```diff
@@ -0,0 +1,21 @@
...
+import {Root} from "@chainsafe/lodestar-types";
+import {LodestarError} from "../../../../utils/lib";
+
+export enum BackfillSyncErrorCode {
+  //fetched block doesn't connect to anchor block
+  NOT_ANCHORED = "not_anchored",
+  // fetched blocks are not linear
+  NOT_LINEAR = "not_linear",
+  // peer doesn't have required blocks
+  MISSING_BLOCKS = "mssing_blocks",
```
wemeetagain: typo

packages/lodestar/src/sync/backfill/errors.ts
```diff
@@ -0,0 +1,21 @@
+import PeerId from "peer-id";
+import {Root} from "@chainsafe/lodestar-types";
+import {LodestarError} from "../../../../utils/lib";
```
wemeetagain: relative import

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,208 @@
...
+
+export enum BackfillSyncStatus {
+  /** Required history blocks are syncing */
+  Syncing,
+
+  /** There are no suitable peers or we synced all required history blocks */
+  Idle,
+}
+
+export type RangeSyncModules = {
```
wemeetagain: Maybe should be called `BackfillSyncModules`

packages/lodestar/src/sync/backfill/errors.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export enum BackfillSyncErrorCode {
+  //fetched block doesn't connect to anchor block
```
wemeetagain: Maybe should be `/** */` style comments so they show up in IDE

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,208 @@
...
+              this.logger.info("Backfill sync completed");
+              break;
+            }
+            this.logger.debug("BackfillSync - range sync", {
+              fromSlot,
+              toSlot,
+              anchorSlot: this.anchorBlock.message.slot,
+              lastFetched: this.lastFetchedSlot ?? -1,
+            });
+            await this.syncRange(fromSlot, toSlot, this.anchorBlock.message.parentRoot);
```
wemeetagain: Should we be downloading multiple batches at a time as we do in range sync?

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,208 @@
...
+          } catch (e) {
+            this.logger.debug("Error while backfiling by range", e);
```
wemeetagain: typo

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,208 @@
...
+  /**
+   * Remove this peer from all sync chains
+   */
+  private removePeer(peerId: PeerId): void {
+    this.network.peerRpcScores.applyAction(peerId, PeerAction.MidToleranceError);
+    this.peers.delete(peerId);
+  }
+
+  private async syncBlock(root: Root): Promise<void> {
+    const peer = getRandomPeer(this.peers.keys());
```
wemeetagain: I think we have ways of ranking peers by score?

wemeetagain (changes requested):

packages/lodestar/src/sync/backfill/backfill.ts
mpetrunic: Had that discussion with dapplion, came to conclusion that speed is not that relevant to backfill sync

packages/lodestar/src/sync/backfill/backfill.ts
mpetrunic: don't think we do that even in range sync. We use score to determine when to disconnect peer which will remove it from this set as well.

wemeetagain (dismissed):

packages/lodestar/src/constants/network.ts
```diff
@@ -38,9 +41,14 @@ export type RpcResponseStatusError = Exclude<RespStatus, RespStatus.SUCCESS>;
 
 /** The maximum allowed size of uncompressed gossip messages. */
 export const GOSSIP_MAX_SIZE = 2 ** 20;
+
+// source: https://github.com/ethereum/eth2.0-specs/blob/f52f067b8ea3f8adbebc936207b06459d1956e72/specs/phase0/p2p-interface.md#why-are-blocksbyrange-requests-only-required-to-be-served-for-the-latest-min_epochs_for_block_requests-epochs
+export function getMinEpochForBlockRequests(config: IChainForkConfig): number {
+  return config.MIN_VALIDATOR_WITHDRAWABILITY_DELAY + intDiv(config.CHURN_LIMIT_QUOTIENT, 2);
```
dapplion: Is this formula right? The link above states
```
MIN_VALIDATOR_WITHDRAWABILITY_DELAY
    + MAX_SAFETY_DECAY * CHURN_LIMIT_QUOTIENT // (2 * 100)
```

mpetrunic: Seems like same thing, since MAX_SAFETY_DECAY is 100. Probably added later to ease in changing MAX_SAFETY_DECAY. Pushing fix

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,220 @@
...
+    this.chain = modules.chain;
+    this.network = modules.network;
+    this.db = modules.db;
+    this.config = modules.config;
+    this.logger = modules.logger;
+    this.opts = opts ?? {batchSize: BATCH_SIZE};
+    this.network.events.on(NetworkEvent.peerConnected, this.addPeer);
+    this.network.events.on(NetworkEvent.peerDisconnected, this.removePeer);
+
+    this.sync()
```
dapplion, line 86: Merged sync() and init() functions to it's easy to detect a completed event in all cases, even when early exit conditions are found

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,220 @@
...
+      this.anchorBlock?.message.slot ?? computeStartSlotAtEpoch(this.chain.getFinalizedCheckpoint().epoch);
+    this.logger.debug("BackfillSync add peer", {peerhead: peerStatus.headSlot, requiredSlot});
+    if (peerStatus.headSlot >= requiredSlot) {
+      this.peers.add(peerId);
+      this.processor.trigger();
+    }
+  };
+
+  private removePeer(peerId: PeerId): void {
+    this.peers.delete(peerId);
```
dapplion, line 207: We cannot report peers here, otherwise we would downscore peers just for disconnecting.

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,220 @@
...
+        await this.syncRange(peer, fromSlot, toSlot, this.anchorBlock.message.parentRoot);
+      } catch (e) {
+        this.logger.debug("BackfillSync sync error", e);
+
+        if (e instanceof BackfillSyncError) {
+          switch (e.type.code) {
+            case BackfillSyncErrorCode.INVALID_SIGNATURE:
+            case BackfillSyncErrorCode.NOT_LINEAR:
+            case BackfillSyncErrorCode.NOT_ANCHORED:
+              this.network.reportPeer(peer, PeerAction.LowToleranceError, "BadSyncBlocks");
```
dapplion, line 185: Report peers only when they do something bad according to internal rules of the BackfilSync. Other reqresp errors like bad ssz formating, timeouts etc are handled and reported in the ReqResp class

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,220 @@
...
+      return;
+    }
+
+    verifyBlockSequence(this.config, blocks, anchorRoot);
+    await verifyBlockProposerSignature(this.chain.bls, this.chain.getHeadState(), blocks);
+
+    await this.db.blockArchive.batchAdd(blocks);
+    // first block is oldest
+    if (blocks[0]) {
+      this.anchorBlock = blocks[0];
```
dapplion, line 233: Check range is not empty otherwise we would unset the previous anchorBlock

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  private async syncRange(peer: PeerId, from: Slot, to: Slot, anchorRoot: Root): Promise<void> {
+    const blocks = await this.network.reqResp.beaconBlocksByRange(peer, {startSlot: from, count: to - from, step: 1});
+    if (blocks.length === 0) {
+      return;
+    }
+
+    verifyBlockSequence(this.config, blocks, anchorRoot);
+    await verifyBlockProposerSignature(this.chain.bls, this.chain.getHeadState(), blocks);
```
dapplion, line 227: Splited the verifyBlock function into two so it can be reused in `syncBlockByRoot`. @mpetrunic Could you please add a unit test for verifyBlockSequence? Should be not complex to test with actual data, this logic is very important.

packages/lodestar/src/sync/backfill/verify.ts
```diff
@@ -0,0 +1,39 @@
...
+import {BackfillSyncError, BackfillSyncErrorCode} from "./errors";
+
+export function verifyBlockSequence(
+  config: IBeaconConfig,
+  blocks: allForkTypes.SignedBeaconBlock[],
+  anchorRoot: Root
+): void {
+  let nextRoot: Root = anchorRoot;
+  for (const block of blocks.slice(0).reverse()) {
+    const blockRoot = config.getForkTypes(block.message.slot).BeaconBlock.hashTreeRoot(block.message);
```
dapplion, line 15: How long does it take to hash a block? Now that the batch size is 64 for how long will this hang the event loop?

packages/lodestar/src/sync/backfill/verify.ts
```diff
@@ -0,0 +1,39 @@
+import {allForks, CachedBeaconState} from "@chainsafe/lodestar-beacon-state-transition";
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
+import {Root, allForks as allForkTypes, ssz} from "@chainsafe/lodestar-types";
+import {GENESIS_SLOT} from "../../../../params/lib";
```
dapplion, line 4: Relative import of a monorepo library!

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    for await (const _ of this.processor) {
+      const peer = shuffleOne(this.peers.values());
+      if (!peer) {
+        this.logger.debug("BackfillSync no peers yet");
+        continue;
+      }
+
+      try {
+        // First get anchor block
```
dapplion, line 155: Re-organized to first try to get anchorBlock, then sync the range of blocks. Allows to organize the completedness check is a clearer way

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,220 @@
...
+
+export class BackfillSync extends (EventEmitter as {new (): BackfillSyncEmitter}) {
+  private readonly chain: IBeaconChain;
+  private readonly network: INetwork;
+  private readonly db: IBeaconDb;
+  private readonly config: IBeaconConfig;
+  private readonly logger: ILogger;
+  private opts: BackfillSyncOpts;
+  /** last trusted block */
+  private anchorBlock: allForks.SignedBeaconBlock | null = null;
```
dapplion, line 69: @mpetrunic Removed the `lastFetchedSlot` since you can get that data from the anchorBlock. Let me know if there was a use for `lastFetchedSlot` that I did not notice.

mpetrunic: I was using that to handle skip slots

dapplion:
@mpetrunic I've been reviewing this branch this morning and applied some changes directly that I though helped improve the design. Please review my commit in case I worsen any area.

There is still a big problem in the current design which is that it will always backfill the entire range over and over on every restart. Even if the node stops for 60 seconds it will try to backfill the entire period again, despite those blocks being already in the db. Since there can be skipped slots or only part of the blocks may be backfilled how can the node reliably tell if it has to backfill or not?
- I've added a very naive check in startup to find at least one block in the target range to backfill. If there's at least 1 block, skip backfill sync. This is not great but better than backfilling always.

Please let me know if you can think of any solution to this problem.

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,220 @@
...
+        this.logger.info("BackfillSync completed", {oldestSlotSynced});
+        // Sync completed, unsubscribe listeners and don't run the processor again
+        // Backfill is never necessary again until the node shuts down
+        this.close();
+      })
+      .catch((e) => {
+        if (e instanceof ErrorAborted) {
+          return; // Ignore
+        }
+        this.logger.error("BackfillSync processor error", e);
```
mpetrunic, line 99: should we call this.close here as well?

packages/lodestar/src/sync/backfill/verify.ts
mpetrunic: Aren't we doing that in regular range sync as well?

packages/lodestar/src/sync/backfill/verify.ts
```diff
@@ -0,0 +1,39 @@
...
+  bls: IBlsVerifier,
+  state: CachedBeaconState<allForks.BeaconState>,
+  blocks: allForkTypes.SignedBeaconBlock[]
+): Promise<void> {
+  const signatures = blocks
+    // genesis block doesn't have valid signature
+    .filter((block) => block.message.slot !== GENESIS_SLOT)
+    .map((block) => allForks.getProposerSignatureSet(state, block));
+
+  if (!(await bls.verifySignatureSets(signatures))) {
```
mpetrunic, line 36: in case of skip slots this will fail because signatures is empty array

mpetrunic:
> @mpetrunic I've been reviewing this branch this morning and applied some changes directly that I though helped improve the design. Please review my commit in case I worsen any area.
> 
> There is still a big problem in the current design which is that it will always backfill the entire range over and over on every restart. Even if the node stops for 60 seconds it will try to backfill the entire period again, despite those blocks being already in the db. Since there can be skipped slots or only part of the blocks may be backfilled how can the node reliably tell if it has to backfill or not?
> 
> * I've added a very naive check in startup to find at least one block in the target range to backfill. If there's at least 1 block, skip backfill sync. This is not great but better than backfilling always.
> 
> Please let me know if you can think of any solution to this problem.

I vaguely remember having check which fetches first block in archive to check that.

packages/lodestar/src/sync/backfill/backfill.ts
dapplion: Yes indeed

packages/lodestar/src/sync/backfill/backfill.ts
dapplion: hhmm, good point. Tho how can we then protect ourselves against nodes that send skip slots when they shouldn't? In range sync there's a two step "verification" process to guard against that. Can we just use a very big batch size and hope it's larger than a range of skipped slots?

packages/lodestar/src/sync/backfill/verify.ts
dapplion: Right! Then I guess it's okay.

packages/lodestar/src/sync/backfill/verify.ts
dapplion: Before calling, we do
```ts
if (blocks.length === 0) {
      return;
    }
```

packages/lodestar/src/sync/backfill/backfill.ts
mpetrunic: Well hash chain will fail after which we will go back and try from last anchor block again?

wemeetagain:
Linter error

packages/lodestar/test/unit/sync/backfill/verify.test.ts
```diff
@@ -0,0 +1,51 @@
...
+
+describe("backfill sync - verify block sequence", function () {
+  //mainnet validators root
+  const beaconConfig = createIBeaconConfig(
+    config,
+    ssz.Root.fromJson("0x4b363db94e286120d76eb905340fdd4e54bfe9f06bf33ff6cf5ad27f511bfe95")
+  );
+
+  it("should verify valid chain of blocks", function () {
+    const blocks = getBlocks();
```
dapplion, line 19: Can we read the file and parse it only once for the describe block? You can either cache the return or call getBlocks in a before() block

mpetrunic: It takes like 10ms to do that so I think we are fine, skipped caching since tests and code is modifying original array so we would have to copy array of blocks between tests

dapplion:
@mpetrunic This PR still has an unresolved issue where it may backfill sync more than necessary. 

To be aware of the problem can you add metrics to backfill sync? I think we need at least:
- total count of backfilled blocks. Increment by one for each processed block
- total count of errors. Increment by one on each download or processor error
- current backfill status. Add a collect function to get the status, similar to range sync

Is it possible for backfill sync to get stuck processing the same range over and over forever? Should it stop trying after N attempts? However if we do that, the node on the next startup may think it's fully backfilled and not do anything.

mpetrunic:
> @mpetrunic This PR still has an unresolved issue where it may backfill sync more than necessary.
> 
> To be aware of the problem can you add metrics to backfill sync? I think we need at least:
> 
> * total count of backfilled blocks. Increment by one for each processed block
> * total count of errors. Increment by one on each download or processor error
> * current backfill status. Add a collect function to get the status, similar to range sync
> 
> Is it possible for backfill sync to get stuck processing the same range over and over forever? Should it stop trying after N attempts? However if we do that, the node on the next startup may think it's fully backfilled and not do anything.

I can do that. What have we figure out with skip slots, will you revert back to supporting it or figure out something new?

dapplion:
> What have we figure out with skip slots, will you revert back to supporting it or figure out something new?

I trust your criteria, please take over the PR if you can

dapplion:
@mpetrunic Teku and Lighthouse backfill to genesis. From the discussion at the eth2 call it's was clearly stated that we must backfill to genesis until there's a strong platform for long term block storage.

I would remove the pruning code and always backfill to genesis.

To know if we have to backfill or not Paul from Lighthouse suggested to persist in the db some notion of ranges that we would read know until when we have to backfill. If this ranges don't exist, then the node may read the block archive db and figure them out.

mpetrunic:
> @mpetrunic Teku and Lighthouse backfill to genesis. From the discussion at the eth2 call it's was clearly stated that we must backfill to genesis until there's a strong platform for long term block storage.
> 
> I would remove the pruning code and always backfill to genesis.
> 
> To know if we have to backfill or not Paul from Lighthouse suggested to persist in the db some notion of ranges that we would read know until when we have to backfill. If this ranges don't exist, then the node may read the block archive db and figure them out.

Doubt there will ever be strong platform for long term storage as long as there are clients storing all blocks but thats just my 2c

Gonna remove pruning and blocks_by_range condition

dapplion:
From Michael Sproul (Lighthouse)

> > If the client stops during backfill sync, on startup how do you decide if you are backfilled or not?
>
> We keep an oldest_block_slot: Slot in the database, which decrements as new blocks are written. We also store the oldest_block_parent: Hash256 to know which block to fetch next.

> > If you have ran the client a month ago and now start from a recent weak subjectivity cp, should the node stop backfilling when it reaches the latest persisted block from a month ago? If so, how can it differentiate between a failed backfilled sync and the chain from the previous run?
>
> At the moment we'll just keep syncing from the old checkpoint. If we add a test to make sure that checkpoint isn't "too old", I think we'll just recommend dropping the DB so we don't have to deal with multiple backfills

mpetrunic:
> From Michael Sproul (Lighthouse)
> 
> > > If the client stops during backfill sync, on startup how do you decide if you are backfilled or not?
> > 
> > 
> > We keep an oldest_block_slot: Slot in the database, which decrements as new blocks are written. We also store the oldest_block_parent: Hash256 to know which block to fetch next.
> 
> > > If you have ran the client a month ago and now start from a recent weak subjectivity cp, should the node stop backfilling when it reaches the latest persisted block from a month ago? If so, how can it differentiate between a failed backfilled sync and the chain from the previous run?
> > 
> > 
> > At the moment we'll just keep syncing from the old checkpoint. If we add a test to make sure that checkpoint isn't "too old", I think we'll just recommend dropping the DB so we don't have to deal with multiple backfills

We do that now as well except I'm not sure why di we need separate counter in db since we can just check first key/value stored in block archive, maybe they don't use slots as keys in block archive.

dapplion:
> We do that now as well except I'm not sure why di we need separate counter in db since we can just check first key/value stored in block archive, maybe they don't use slots as keys in block archive.

If I understand they store the oldest_block_slot from backfill sync. The problem here is how you differentiate between backfill sync and regular block achiving if you just check the first key?

mpetrunic:
> > We do that now as well except I'm not sure why di we need separate counter in db since we can just check first key/value stored in block archive, maybe they don't use slots as keys in block archive.
> 
> If I understand they store the oldest_block_slot from backfill sync. The problem here is how you differentiate between backfill sync and regular block achiving if you just check the first key?

Ah you mean, when people start and stop nodes that we won't backfillsync gap? Not sure how storing separate last fetch helps there though. I think we should also recommend to just drop db haha

dapplion:
Closing for https://github.com/ChainSafe/lodestar/pull/3384

PR #2639, opened 2021-06-05
Set SYNC_COMMITTEE_SIZE to 512 for mainnet

twoeths:
**Motivation**

Got "Invalid state root" at fork transition slot 320 of oonoonba sync

**Description**

Change SYNC_COMMITTEE_SIZE to 512 as in https://github.com/ethereum/eth2.0-specs/blob/v1.1.0-alpha.6/presets/mainnet/altair.yaml


wemeetagain (approved):
Probably worth doing a deeper review on all changes from alpha.5 to alpha.6

PR #2657, opened 2021-06-08
Fix getParticipantIndices from SyncAggregate struct

twoeths:
**Motivation**
`getParticipantIndices` is correct for tree backed object but not struct

**Description**

+ same way to get attesting indexes from a struct https://github.com/ChainSafe/lodestar/blob/0c4f8315fa4596e5bb0ff8ada8e8d8c38499b6da/packages/beacon-state-transition/src/allForks/util/epochContext.ts#L414

### DIFF
--- a/packages/beacon-state-transition/src/altair/block/processSyncCommittee.ts
+++ b/packages/beacon-state-transition/src/altair/block/processSyncCommittee.ts
@@ -4,16 +4,15 @@ import {DOMAIN_SYNC_COMMITTEE} from "@chainsafe/lodestar-params";
 import {
   computeEpochAtSlot,
   computeSigningRoot,
+  extractParticipantIndices,
   getBlockRootAtSlot,
   getDomain,
   increaseBalance,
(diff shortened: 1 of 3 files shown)


packages/beacon-state-transition/src/altair/block/processSyncCommittee.ts
```diff
@@ -91,11 +91,19 @@ function getParticipantIndices(
...
+      ssz.altair.SyncCommitteeBits
+    );
+  } else {
+    const participantIndices = [];
+    for (const [i, index] of committeeIndices.entries()) {
+      if (syncAggregate.syncCommitteeBits[i]) {
+        participantIndices.push(index);
+      }
+    }
+    return participantIndices;
```
dapplion: Should we move this to a util function and unit test it?

dapplion (approved):
Great work!

wemeetagain (approved):

PR #2667, opened 2021-06-09
Publish nightly version to Docker Hub too

dapplion:
**Motivation**

With https://github.com/ChainSafe/lodestar/pull/2647 we are releasing nightly versions to NPM. We should do the same for Docker Hub.

**Description**

Grab the tag from lerna's logic and publish to Docker Hub.

### DIFF
--- a/.github/workflows/release-nightly.yml
+++ b/.github/workflows/release-nightly.yml
@@ -58,7 +58,7 @@ jobs:
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
-      - run: docker build --tag chainsafe/lodestar:next .
+      - run: docker build --tag chainsafe/lodestar:next --build-arg VERSION=${{ needs.publish.outputs.version }} .
       - run: docker tag chainsafe/lodestar:next chainsafe/lodestar:${{ needs.publish.outputs.version }}
         # Push all tags
       - run: docker push chainsafe/lodestar
--- a/.github/workflows/release-nightly.yml
+++ b/.github/workflows/release-nightly.yml
@@ -1,7 +1,7 @@
 name: Release nightly
 
 # only one per github sha can be run
-concurrency: 
+concurrency:
   group: ${{ github.sha }}
   cancel-in-progress: true
 
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: "14.16.0"
-          registry-url: 'https://registry.npmjs.org'
+          registry-url: "https://registry.npmjs.org"
       - name: Restore dependencies
         uses: actions/cache@master
         id: cache-deps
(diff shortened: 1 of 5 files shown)


.github/workflows/release-nightly.yml
```diff
@@ -40,3 +40,24 @@ jobs:
         run: yarn run publish:nightly
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Get latest tag
+        uses: actions-ecosystem/action-get-latest-tag@v1
```
mpetrunic: ain't this loading latest git tag?

Nightly releases don't have tag.

You could do something like:
```
runs: |
   packageJson=`cat ./package.json`
   version=`echo $(jq -r '.version' <<< "$packageJson")`
   echo "::set-output name=version::$version"
```
not sure if this works needs testing 😅

.github/workflows/release-nightly.yml
```diff
@@ -40,3 +40,24 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: publish
+    steps:
+      - uses: actions/checkout@v2
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      # TODO: Prevent Docker Hub tagging this image as "latest", tag as "next" instead
```
mpetrunic: just add 
`docker tag chainsafe/lodestar:${{ needs.publish.outputs.tag }} chainsafe/lodestar:next`
and also push chainsafe/lodestar:next

mpetrunic (commented):
I think it would be wise to also change dockerfile to only install npm package otherwise we will spend huge amount of ci time just building that docker images

.github/workflows/release-nightly.yml
dapplion: Thanks! What about preventing dockerhub from tagging it as latest?

dapplion:
> I think it would be wise to also change dockerfile to only install npm package otherwise we will spend huge amount of ci time just building that docker images

Yeah I'm on it https://github.com/ChainSafe/lodestar/pull/2672

.github/workflows/release-nightly.yml
mpetrunic: I don't think dockerhub does that by default. You have to push latest tag in order to update it or push image without tag

mpetrunic: or not hmm
We should probably start pushing stable tag with our main release

dapplion:
Warning! There can be a race condition if NPM does not serve the recently pushed version immediately. The build should target the exact version of this nightly build.

We could use this API https://registry.npmjs.org/@chainsafe/lodestar-cli to ensure that our target nightly version is already available.

mpetrunic (dismissed):

mpetrunic (approved):

PR #2674, opened 2021-06-10
get head metric

g11tech:
**Motivation**

this PR is to add fork-choice metrics, with getHead metric : `beacon_fork_choice_find_head_seconds` as a pilot/reference to add other fork choice metrics post this.

image

**Description**

Closes #issue_number


packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -157,31 +166,46 @@ export class ForkChoice implements IForkChoice {
...
+          root: fromHexString(headRoot),
+        });
+      }
+
+      const blockSummary = toBlockSummary(headNode);
+      if (timer) timer();
+      return blockSummary;
+    } catch (e) {
+      if (timer) timer();
+      throw e;
```
dapplion: Oh nice idea to stop the time on error! you could do that in a finally block and won't have to re-throw

packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -202,9 +202,8 @@ export class ForkChoice implements IForkChoice {
       const blockSummary = toBlockSummary(headNode);
       if (timer) timer();
```
dapplion: Now with the finally block there's no need to also stop the timer in the try block

g11tech: yea i figured, wait for the new commit, you review too fast! :)

g11tech:
@dapplion  check now pls. :)

packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -157,31 +166,44 @@ export class ForkChoice implements IForkChoice {
...
+      }
+      const headNode = this.protoArray.nodes[headIndex];
+      if (headNode === undefined) {
+        throw new ForkChoiceError({
+          code: ForkChoiceErrorCode.MISSING_PROTO_ARRAY_BLOCK,
+          root: fromHexString(headRoot),
+        });
+      }
+
+      blockSummary = toBlockSummary(headNode);
```
dapplion: You should return blockSummary here without setting it to a higher scope variable. Does it break otherwise?

g11tech: i will need to test if the finally segment will still run.

g11tech: hmmm its running, do you want me to make the change to bring it inside?

g11tech:
@dapplion , should be good now.

wemeetagain (dismissed):
lgtm

dapplion (changes requested):
I've been thinking about it and there are two different routines that should be tracked separaterly
1. computeDeltas + applyScoreChanges
2. findHead

findHead should be very cheap so it's not that important to track. Why really want to track computeDeltas + applyScoreChanges, but they are called in more places.

I think we should move all computeDeltas + applyScoreChanges calls to a dedicated method and time that. They we will be sure to capture all calls to the expensive fork choice routine.

g11tech:
> 
> 
> I've been thinking about it and there are two different routines that should be tracked separaterly
> 
>     1. computeDeltas + applyScoreChanges
> 
>     2. findHead
> 
> 
> findHead should be very cheap so it's not that important to track. Why really want to track computeDeltas + applyScoreChanges, but they are called in more places.
> 
> I think we should move all computeDeltas + applyScoreChanges calls to a dedicated method and time that. They we will be sure to capture all calls to the expensive fork choice routine.

ok, let me look into the same 👍

wemeetagain:
@g11in check out https://github.com/ChainSafe/lodestar/pull/2696
Just moved all the heavy computation to the `updateHead` function, which resolves @dapplion's comment about consolidating the calls to computeDeltas and applyScoreChanges.
Fixing the small file conflict, you should be able to wrap that method in the same way you wrapped `getHead` and we'll be ready to merge this :)

g11tech:
> @g11in check out #2696
> Just moved all the heavy computation to the `updateHead` function, which resolves @dapplion's comment about consolidating the calls to computeDeltas and applyScoreChanges.
> Fixing the small file conflict, you should be able to wrap that method in the same way you wrapped `getHead` and we'll be ready to merge this :)

ok so I take it after `updateJustified` (via onblock), `updateHead` is being called in the refactored code? because that was the only other place from where computeDeltas and applyScoreChanges were being called apart from getHead?
yea this would be easy to resolve.

g11tech:
sending a new PR, some git merging issue

PR #2679, opened 2021-06-10
Add AttestationPool to aggregate attestations on the go

### LINKED ISSUE #2415: Pre-aggregate attestations in Aggregation pool

dapplion:
**Motivation**

See https://github.com/ChainSafe/lodestar/issues/2415

**Description**

Substitutes the current disk-persisted attestation repo with a memory-only attestation pool. It pre-aggregates attestations saving memory and reducing disk reads and writes.

Closes https://github.com/ChainSafe/lodestar/issues/2415


packages/lodestar/src/chain/eventHandlers.ts
```diff
@@ -336,8 +338,9 @@ export async function onErrorAttestation(this: BeaconChain, err: AttestationErro
       this.pendingAttestations.putByBlock(err.type.beaconBlockRoot, err.job);
       break;
 
-    default:
-      await this.db.attestation.remove(err.job.attestation);
+    // TODO: Why is necessary to remove the attestation from the DB on error?
```
mpetrunic, line 341: I think this was failsafe since we didn't have api validation before so on some occasion, validator would put some invalid attestation which would prevent all blocks from being processed

packages/lodestar/src/chain/factory/block/body.ts
```diff
@@ -25,6 +25,18 @@ export async function assembleBody(
   blockSlot: Slot,
   syncAggregateData: {parentSlot: Slot; parentBlockRoot: Root}
 ): Promise<allForks.BeaconBlockBody> {
+  // TODO:
```
mpetrunic, line 28: I wonder if we should, if number of agg attestations is < MAX_ATTESTATIONS, take some attestations from attestations memory pool. Assuming we delete attestation that are aggregated?

packages/lodestar/src/db/attestationPool.ts
```diff
@@ -0,0 +1,230 @@
...
+  add(attestation: phase0.Attestation): void {
+    const slot = attestation.data.slot;
+    const lowestPermissibleSlot = this.lowestPermissibleSlot;
+
+    // Reject any attestations that are too old.
+    if (slot < lowestPermissibleSlot) {
+      throw new AttestationPoolError({code: AttestationPoolErrorCode.SLOT_TOO_LOW, slot, lowestPermissibleSlot});
+    }
+
+    const dataRoot = ssz.phase0.AttestationData.hashTreeRoot(attestation.data);
```
mpetrunic, line 101: wouldn't more future safe way be getForkTypes(slot).AttestationData?

dapplion: Yes but we should change it everywhere at once in another PR. This logic in unchanged in this PR just moved around.

packages/lodestar/src/chain/eventHandlers.ts
dapplion: How could an invalid attestation prevent all blocks from being processed?

packages/lodestar/src/chain/factory/block/body.ts
dapplion: Right, the logic to do that successfully is tricky. Since we were not doing that previously it should be done in another PR if we decide to.

packages/lodestar/src/chain/eventHandlers.ts
mpetrunic: process block failed when processing attestations and attestation was put in every block we were producing

dapplion: But how could an invalid attestation got included in the attestation pool?

twoeths: this error handler for attestation was for forkchoice, not when we process block.
Since it was for forkchoice, it's not really necessary to remove that attestation from db. The next time we see that same error from that same attestation, we could just throw error and handle the same way but this is rare.

twoeths (approved):

PR #2681, opened 2021-06-10
fix: Api should not return TreeBacked objects

### LINKED ISSUE #2572: Api should not return TreeBacked objects
**Describe the bug**

Currently the Api impl returns tree backed objects. The return types are typed as regular Javascript objects and unsuspecting users may encounter issues. This happens on the beacon.state routes (at least).

**Expected behavior**

Since everything coming out of the Api impl isn't ever going to be hashed we should drop the trees. If someone **really** wants to consume tree backed objects they should connect directly the BeaconChain instance.

3xtr4t3rr3str14l:
fixes #2572

### DIFF
--- a/packages/lodestar/src/api/impl/beacon/state/utils.ts
+++ b/packages/lodestar/src/api/impl/beacon/state/utils.ts
@@ -260,7 +260,7 @@ async function getFinalizedState(
   db: IBeaconDb,
   forkChoice: IForkChoice,
   slot: Slot
-): Promise<CachedBeaconState<allForks.BeaconState>> {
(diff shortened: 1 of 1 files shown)


dapplion (changes requested):
This will break the lightclient API which requires a tree backed state.

Also, note that converting a tree backed object to struct is very slow, see benchmarks here https://github.com/ChainSafe/ssz/runs/2722203471

I didn't know that before opening #2572 but it takes x1.5 times a deserialize operation which can be 20-40 seconds for a full beacon state. So this can be a serious performance bottleneck. We should study this change more and apply to very specific places instead of deep in a utility.

Issue #467, opened 2019-09-25
Implement missing ssz generic spec tests


mpetrunic:
Currently our ssz spec tests covers only static tests, our spec-tes-utils package provides spec test directory as param so it's possible to implement generic tests as well.

mpetrunic:
unstale

dapplion:
@mpetrunic Could you detail what tests specifically are lacking?

mpetrunic:
> @mpetrunic Could you detail what tests specifically are lacking?

https://github.com/ethereum/eth2.0-specs/tree/dev/tests/formats/ssz_generic 

I had them implemented but contrainer tests werent passing and meanwhile we removed ssz from lodestar so that branch got deleted :(

dapplion:
@mpetrunic :sweat_smile: So should this issue be moved to the SSZ repo?

mpetrunic:
cc @wemeetagain since he does not want spec tests dep in ssz repo :)

dapplion:
@wemeetagain bump

dapplion:
I've been working on this issue today and there are a lot of errors and very serious on the SSZ side. We must incorporate these tests in the ssz repo and fix the issues asap.
- Numbers being deserialized to `Infinity`
- Strange rounding errors, `13097433917148758497` being deserialized to `13097433917148758000`
- Wrong hash tree roots

Branch WIP https://github.com/ChainSafe/lodestar/compare/dapplion/spec-test-ssz-generic

dapplion:
@mpetrunic I would prefer for some else to take this issue since I'm not that familiar with SSZ internals

wemeetagain:
assigning to myself

dapplion:
Implemented

PR #2684, opened 2021-06-10
Adds generic ssz spec tests

wemeetagain:
**Motivation**

See #467 

**Description**

Adds generic ssz spec tests - minus one container type which relies on https://github.com/ChainSafe/ssz/issues/128


dapplion (dismissed):

mpetrunic (commented):
I guess we should pull latest ssz changes before merging

twoeths (approved):

PR #2694, opened 2021-06-13
De-duplicate code in altair state transition fn

### LINKED ISSUE #2691: De-duplicate processAttestation()
**Describe the bug**

processAttestation() in phase0 and altair contains a lot of duplicated code.

https://github.com/ChainSafe/lodestar/blob/16b856f45cf6dcf0c63f28b2ec13d1939bcfc8d9/packages/beacon-state-transition/src/phase0/block/processAttestation.ts#L8

https://github.com/ChainSafe/lodestar/blob/16b856f45cf6dcf0c63f28b2ec13d1939bcfc8d9/packages/beacon-state-transition/src/altair/block/processAttestation.ts#L22

**Expected behavior**

De-duplicate code as much as possible

dapplion:
**Motivation**

De-duplicate code from altair's state transition fn methods.

**Description**

Multiplex fork differences inside each function when the differences are small enough. This approach involves one extra function call than necessary but it should be okay in terms of performance.

Closes https://github.com/ChainSafe/lodestar/issues/2691


packages/beacon-state-transition/src/phase0/block/processProposerSlashing.ts
```diff
@@ -2,17 +2,32 @@ import {allForks, phase0, ssz} from "@chainsafe/lodestar-types";
...
+): void {
+  processProposerSlashingAllForks(
+    ForkName.phase0,
+    state as CachedBeaconState<allForks.BeaconState>,
+    proposerSlashing,
+    verifySignatures
+  );
+}
+
+export function processProposerSlashingAllForks(
```
twoeths: should this be in `allForks` instead?

packages/beacon-state-transition/src/phase0/block/processAttesterSlashing.ts
```diff
@@ -1,14 +1,29 @@
...
+): void {
+  processAttesterSlashingAllForks(
+    ForkName.phase0,
+    state as CachedBeaconState<allForks.BeaconState>,
+    attesterSlashing,
+    verifySignatures
+  );
+}
+
+export function processAttesterSlashingAllForks(
```
twoeths: similar to the other place, I think we want to put this in `allForks`, so for common function:
+ `phase0` => `allForks`
+ `altair` => `allForks`

I hope it'd resolve the module loading issue in the spec test

dapplion:
Ready for review

twoeths (approved):

PR #2695, opened 2021-06-14
Improve validator exit flow

### LINKED ISSUE #2342: Potential performance issue in initiateValidatorExit()
**Describe the bug**

Every time `initiateValidatorExit()` is called it iterates over the entire validators array. initiateValidatorExit may be called multiple times per block with the right conditions.

https://github.com/ChainSafe/lodestar/blob/e481a8434c0b658890fcf700cb074ddd5d3cf7cd/packages/beacon-state-transition/src/util/validatorStatus.ts#L37

**Expected behavior**

We should do something smarter and cache this computation

twoeths:
**Motivation**

+ Processing a block with 16 validator exits is 100x times slower than regular block
+ Slashing validator triggers validator exit also

**Description**

+ Most of the time was for looping through validators, create a cache (`ValidatorExitProcess`) per block to make sure we loop through it at most 1 time
+ Minimize for loop inside `initiateValidatorExit`
+ The result in `block.perf.ts` is reduced from 100x to 6x, notice that processing 1 validator exit takes almost the same time to processing 16 validator exits
```
// Process block
// ================================================================
// Process block with 0 validator exit                                    233.6434 ops/s      4.280027 ms/op   3491 runs    15.01 s
// Process block with 1 validator exit                                    41.33581 ops/s      24.19210 ms/op    619 runs    15.00 s
// Process block with 16 validator exits                                  42.34492 ops/s      23.61558 ms/op    635 runs    15.02 s
```

Closes #2342


dapplion (commented):
Awesome performance boost! Tho I think it would be simpler to refactor the ValidatorExitProcess into a `BlockProcess` and make it mandatory param in the function signature.

dapplion (approved):
Looks great! Nice perf boost :rocket: We should add a benchmark for attester slashing and proposer slashing too

wemeetagain (approved):

PR #2696, opened 2021-06-14
Cache head in fork choice

wemeetagain:
**Motivation**

We currently recompute the head 'too much'. We would like more control over when the fork choice is doing heavy work and when it's doing light work.

**Description**

Minimal update to add a cache of the head inside the fork choice module.

`forkChoice.getHead` and `getHeadRoot` now use a cached value.
To trigger an update of the head, possibly recomputing balance and vote changes, call `forkChoice.updateHead`

Now we can decide when/where to call `updateHead`.
This PR adds `updateHead` during block processing.

### DIFF
--- a/packages/fork-choice/src/forkChoice/forkChoice.ts
+++ b/packages/fork-choice/src/forkChoice/forkChoice.ts
@@ -76,6 +76,11 @@ export class ForkChoice implements IForkChoice {
   synced: boolean;
 
   /**
+   * Cached head
+   */
+  head: IBlockSummary;
+
+  /**
    * Instantiates a Fork Choice from some existing components
    *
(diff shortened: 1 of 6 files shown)


packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -538,14 +548,7 @@ export class ForkChoice implements IForkChoice {
   }
 
   private updateJustified(justifiedCheckpoint: phase0.Checkpoint, justifiedBalances: Gwei[]): void {
```
wemeetagain, line 556: need to set this.synced to false here?

dapplion (approved):

PR #2698, opened 2021-06-15
fix getCommitteeAssignment performance issue

### LINKED ISSUE #2690: getCommitteeAssignment performance issue
**Describe the bug**

`getCommitteeAssignment()` will loop through the entire shuffled index list (worst case) for every validator connected to the beacon node.

https://github.com/ChainSafe/lodestar/blob/16b856f45cf6dcf0c63f28b2ec13d1939bcfc8d9/packages/beacon-state-transition/src/allForks/util/epochContext.ts#L440-L449

**Expected behavior**

Find a more efficient approach

3xtr4t3rr3str14l:
fixes #2690 

on my laptop, the performance increased from ~3-4ms to < 1 ms (when testing with a state generated by [`generatePerformanceState()`](https://github.com/ChainSafe/lodestar/blob/5c3331bebfbe233f053c4dcb19abc6e6e1c34394/packages/beacon-state-transition/test/perf/util.ts#L82))

### DIFF
--- a/packages/beacon-state-transition/src/allForks/util/epochContext.ts
+++ b/packages/beacon-state-transition/src/allForks/util/epochContext.ts
@@ -439,19 +439,16 @@ export class EpochContext {
 
     const epochStartSlot = computeStartSlotAtEpoch(epoch);
     for (let slot = epochStartSlot; slot < epochStartSlot + SLOTS_PER_EPOCH; slot++) {
-      const committeeCount = this.getCommitteeCountAtSlot(slot);
-      for (let i = 0; i < committeeCount; i++) {
-        const committee = this.getBeaconCommittee(slot, i);
-        if (committee.includes(validatorIndex)) {
-          return {
-            validators: committee as List<number>,
-            committeeIndex: i,
-            slot,
-          };
-        }
-      }
+      const slotCommittee = this._getSlotCommittees(slot);
+      const committeeIndex = slotCommittee.findIndex((v) => v.includes(validatorIndex));
+      const committee = slotCommittee[committeeIndex];
+      if (!committee) continue;
+      return {
+        validators: committee as List<number>,
+        committeeIndex,
+        slot,
+      };
     }
-
     return null;
   }
 
--- a/packages/beacon-state-transition/src/allForks/util/epochContext.ts
+++ b/packages/beacon-state-transition/src/allForks/util/epochContext.ts
@@ -442,12 +442,13 @@ export class EpochContext {
       const slotCommittee = this._getSlotCommittees(slot);
       const committeeIndex = slotCommittee.findIndex((v) => v.includes(validatorIndex));
       const committee = slotCommittee[committeeIndex];
-      if (!committee) continue;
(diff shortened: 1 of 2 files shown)


dapplion:
> performance increased from ~3-4ms to < 1 ms

Can you share the full details of the benchmark including run count and name of the runs?

packages/beacon-state-transition/src/allForks/util/epochContext.ts
```diff
@@ -439,19 +439,17 @@ export class EpochContext {
 
     const epochStartSlot = computeStartSlotAtEpoch(epoch);
     for (let slot = epochStartSlot; slot < epochStartSlot + SLOTS_PER_EPOCH; slot++) {
```
twoeths, line 441: at api side we call this function in a for loop with a list of `validatorIndices`. I wonder if we can do that in batch, i.e supporting a list of `ValidatorIndex` in this function too.

3xtr4t3rr3str14l:
> > performance increased from ~3-4ms to < 1 ms
> 
> Can you share the full details of the benchmark including run count and name of the runs?

@dapplion I ran the following test locally and got the following results (a little faster today than when i initially ran on master yesterday):

master
   - avg perf:  0.591

this PR
   - avg perf:  0.099

```typescript
import {config} from "@chainsafe/lodestar-config/default";
import { BeaconState, Epoch } from "../../phase0";
import {CachedBeaconState, createCachedBeaconState} from "../../src/allForks";
import {generatePerformanceState} from "../perf/util";

describe.only("getCommitteeAssignment perf test group", function () {
  this.timeout(0);

  const numRuns = 1000;
  let perfCumulative = 0;
  let epoch: Epoch, state, cstate: CachedBeaconState<BeaconState>;
    
  before(() => {
    epoch = 23638;
    state = generatePerformanceState();
    cstate = createCachedBeaconState(config, state);
  });

  after(() => {
    console.log('avg perf: ', perfCumulative / numRuns);
  });

  for (let i = 0; i < numRuns; i++) {
    it("getCommitteeAssignment indivdual test", () => {
      const start = Date.now();
      const ca = cstate.getCommitteeAssignment(epoch, 0);
      const perf = Date.now() - start;
      console.log("perf: ", Date.now() - start);
      perfCumulative += perf;
    });
  }
});

```

packages/beacon-state-transition/src/allForks/util/epochContext.ts
wemeetagain: Good idea. Lets merge this first and attempt that separately.

wemeetagain (changes requested):
@3xtr4t3rr3str14l
Can you commit the perf test? Other than that looks good.

As a followup, I'd take a look at @tuyennhv 's suggestion, adding a
`getCommitteeAssignments(epoch: Epoch, validatorIndices: ValidatorIndex[]): (phase0.CommitteeAssignment | null)[]`
and proving out w a benchmark.

wemeetagain (approved):

PR #2700, opened 2021-06-15
IndexedSyncCommittee

### LINKED ISSUE #2518: Create IndexedSyncCommittee

Right now we put indexed data of a sync committee inside EpochContext:
```ts
  currSyncCommitteeIndexes: ValidatorIndex[];
  nextSyncCommitteeIndexes: ValidatorIndex[];
  currSyncComitteeValidatorIndexMap: SyncComitteeValidatorIndexMap;
```

My proposal is to put index data inside an `IndexedSyncCommittee`, similar to `IndexedAttestation`
```ts
export interface IndexedSyncCommittee extends SyncCommittee {
  attestingIndices: List<ValidatorIndex>;
  validatorIndexMap: SyncComitteeValidatorIndexMap;
  // add more if needed
}

type SyncComitteeValidatorIndexMap = Map<ValidatorIndex, number[]>;
```

**Additional context**
+ The `IndexedSyncCommittee` is cached inside `CachedBeaconState` (similar to `balances`, `validators`) with the same `currentSyncCommittee`, `nextSyncCommittee` property key
+ Ensure the tree backed data structure is the same
+ When cloning `CachedBeaconState` we can just share the same `IndexedSyncCommittee` instead of doing a real clone

What do you think @wemeetagain @dapplion ?

twoeths:
**Motivation**

+ Currently we have sync committee in state and its indexed data in epoch context
+ We want to manage and update them in a single place

**Description**

+ Create IndexedSyncComittee with indexed data inside, store it in `CachedBeaconState`
+ Use the same instance per normal epoch transition
+ Do `rotateSyncCommittee` per sync committee period transition
+ Sync speed is the same to master when I sync oonoonba

Closes #2518


packages/beacon-state-transition/src/allForks/util/cachedBeaconState.ts
```diff
@@ -57,7 +60,11 @@ export function createCachedBeaconState<T extends allForks.BeaconState>(
   const cachedBalances = MutableVector.from(readonlyValues(state.balances));
   let cachedPreviousParticipation, cachedCurrentParticipation;
   const forkName = config.getForkName(state.slot);
-  if (forkName === ForkName.phase0) {
+  const isOnPhase0 = forkName === ForkName.phase0;
+  let currIndexedSyncCommittee: IndexedSyncCommittee | undefined;
+  let nextIndexedSyncCommittee: IndexedSyncCommittee | undefined;
+  const epochCtx = createEpochContext(config, state, cachedValidators, opts);
+  if (isOnPhase0) {
```
dapplion: Can we do here `forkName === ForkName.phase0` directly for consistency with other code?

packages/beacon-state-transition/src/allForks/util/cachedBeaconState.ts
```diff
@@ -120,6 +129,8 @@ export class BeaconStateContext<T extends allForks.BeaconState> {
     balanceCache: MutableVector<T["balances"][number]>,
     previousEpochParticipationCache: MutableVector<IParticipationStatus>,
     currentEpochParticipationCache: MutableVector<IParticipationStatus>,
+    currentSyncCommittee: IndexedSyncCommittee | undefined,
+    nextSyncCommittee: IndexedSyncCommittee | undefined,
```
dapplion: Should we not make this properties optional and initialize them to empty on phase0? Would that be very wasteful?

twoeths: it's not wasteful but why do we want to do that? If `undefined` is not clear enough, we can make it `null`

dapplion: Not having optional types is great for consumer code. We did a big refactor a few months ago around the strictNullChecks to not have optional state when avoidable.

twoeths: 👍 will do that, we have the same thing for `cachedPreviousParticipation`

dapplion (approved):
:fire: :+1:

wemeetagain (approved):
:+1: lgtm

Issue #2703, opened 2021-06-15
Can't install @next version


dapplion:
**Describe the bug**

```
lion@lion-gram:~/Code/eth2.0/eth2-light-client-demo$ yarn add @chainsafe/lodestar-light-client@next
yarn add v1.22.5
[1/4] Resolving packages...
[2/4] Fetching packages...
error An unexpected error occurred: "ENAMETOOLONG: name too long, unlink '/home/lion/.cache/yarn/v6/npm-@chainsafe-lodestar-beacon-state-transition-0.24.3-next.f0f46706b561dc5ed13af1f667f296731473abf4.0-13505f390d25231072c6a25191c97a3af0b004b4-integrity/node_modules/@chainsafe/lodestar-beacon-state-transition'".
```

**Expected behavior**

package should be installable

**Steps to Reproduce**

```
yarn add @chainsafe/lodestar-light-client@next
```

PR #2705, opened 2021-06-15
Fix nightly release

dapplion:
**Motivation**

Nightly release is broken.

Also the version is so long that NPM breaks, see https://github.com/ChainSafe/lodestar/issues/2703

**Description**

- Use `--no-git-reset` flag to access version artifacts in next step
- Use shorter commit sha

### DIFF
--- a/.github/workflows/release-nightly.yml
+++ b/.github/workflows/release-nightly.yml
@@ -9,7 +9,6 @@ on:
   push:
     branches:
       - master
-      - nightly-release
 
 jobs:
   publish:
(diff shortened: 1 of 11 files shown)


wemeetagain (dismissed):

dapplion:
@wemeetagain @mpetrunic is the `prepublishOnly` script necessary? It causes an unnecessary double build only in some packages.

mpetrunic:
> @wemeetagain @mpetrunic is the `prepublishOnly` script necessary? It causes an unnecessary double build only in some packages.

Probably not but you never know what some package needs.

wemeetagain (approved):

PR #2707, opened 2021-06-15
Simplify required params in light client init

wemeetagain:
**Motivation**

Currently, initializing the light client against a testnet is more difficult than necessary, due to requiring the genesis time and genesis validators root at the point of light client intialization.

The configuration can be simplified by simply querying the information required as part of the initial proof payload.

**Description**

Sets the genesis time and genesis validators root to be _optional_ in `initializeFromTrustedStateRoot`.
This allows the light client to be initialized with only: a config, a beacon node url, a state root checkpoint.

### DIFF
--- a/packages/light-client/src/client/index.ts
+++ b/packages/light-client/src/client/index.ts
@@ -77,17 +77,17 @@ export class Lightclient {
         nextSyncCommittee: deserializeSyncCommittee(state.nextSyncCommittee),
       },
     };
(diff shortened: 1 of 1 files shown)


3xtr4t3rr3str14l (dismissed):

dapplion (approved):
Yeah much simpler

3xtr4t3rr3str14l (approved):

PR #2710, opened 2021-06-16
Add benchmark for ssz.Root.equals

### LINKED ISSUE #2708: ssz.Root.equals performance
**Describe the bug**

In the profiler result below, Root comparison and state.setStateCachesAsPersistent seems to be too high compared to others:

Screen Shot 2021-06-16 at 08 22 04

We need to analyse why and if we can improve it

## Update
My other profiler shows that `postState.setStateCachesAsPersistent();` is only 0.1ms (so we can iggnore it) while Root comparison is still that high (2.6ms)

twoeths:
**Motivation**

+ `byteArrayEquals` is better than `ssz.Root.equals` in terms of performance

**Description**

Added benchmark result:
```
// Compare state root
// ================================================================
// ssz.Root.equals                                                        17262.51 ops/s      57.92900 us/op 256558 runs    15.01 s
// ssz.Root.equals with valueOf()                                         61656.08 ops/s      16.21900 us/op 903088 runs    15.03 s
// byteArrayEquals with valueOf()                                         848896.4 ops/s      1.178000 us/op 9904067 runs    15.38 s
```

Closes #2708

**Update**
+ Only add performance test, `ssz.Root.equals` is improved in ssz

### DIFF
--- a/packages/beacon-state-transition/test/perf/phase0/block/block.perf.ts
+++ b/packages/beacon-state-transition/test/perf/phase0/block/block.perf.ts
@@ -1,8 +1,9 @@
 import {init} from "@chainsafe/bls";
 import {MAX_VOLUNTARY_EXITS} from "@chainsafe/lodestar-params";
 import {phase0} from "@chainsafe/lodestar-types";
+import {ssz} from "@chainsafe/lodestar-types";
 import {BenchmarkRunner} from "@chainsafe/lodestar-utils/test_utils/benchmark";
-import {List} from "@chainsafe/ssz";
+import {byteArrayEquals, List, fromHexString} from "@chainsafe/ssz";
 import {allForks} from "../../../../src";
 import {generatePerformanceBlock, generatePerfTestCachedBeaconState} from "../../util";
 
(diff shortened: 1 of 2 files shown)


mpetrunic (commented):
Can we make ssz.Root use byteArrayEquals instead? 😄

wemeetagain:
> Can we make ssz.Root use byteArrayEquals instead? smile

We can! I ran the benchmark with https://github.com/ChainSafe/ssz/pull/130 and https://github.com/ChainSafe/lodestar/pull/2713 applied, got these results:

```
Compare state root                                                                                                                      
ssz.Root.equals                                                         1101322 ops/s      908.0000 ns/op 12559206 runs    15.51 s      
ssz.Root.equals with valueOf()                                         731528.9 ops/s      1.367000 us/op 8654589 runs    15.37 s       
byteArrayEquals with valueOf()                                         785546.0 ops/s      1.273000 us/op 9585978 runs    15.39 s
```

dapplion (dismissed):
Looks good, but why is the ssz lib version so slow?? Can we implement an optimization there too?

3xtr4t3rr3str14l:
since #2713 is merged do we not need this PR anymore?

wemeetagain:
> since #2713 is merged do we not need this PR anymore?

We should still commit the perf tests here

wemeetagain (approved):

PR #2717, opened 2021-06-16
getCommitteeAssignments

### LINKED ISSUE #2716: implement batch getCommitteeAssignment
at api side we call this function in a for loop with a list of `validatorIndices`. I wonder if we can do that in batch, i.e supporting a list of `ValidatorIndex` in this function too.

_Originally posted by @tuyennhv in https://github.com/ChainSafe/lodestar/pull/2698#discussion_r651674791_

3xtr4t3rr3str14l:
Resolves #2716 by implementing `getCommitteeAssignments`, a replacement of `assembleAttesterDuty`.

current benchmark results:
```
get attester duties
new way: getCommitteeAssignments                                      0.6062110 ops/s      1.649591  s/op      2 runs    3.299 s
new way (plus index2pubkey): getCommitteeAssignments                  0.7766266 ops/s      1.287620  s/op      2 runs    2.576 s
old way: assembleAttesterDuty batch                                   0.2985871 ops/s      3.349106  s/op      2 runs    6.698 s
    ✓ performance comparison (12678ms)
```

The benchmark test I used to get those numbers is in this PR. It also checks to make sure that `getCommitteeAssignments` gets the same values that batch `assembleAttesterDuty` previously did.

Removed `assembleAttesterDuty` as well.


3xtr4t3rr3str14l:
NOTE: noticed something may be off in my benchmark, so I'm going to double check it and put the results in the description once I'm done checking & benchmarking.

3xtr4t3rr3str14l:
Had to fix a bug (which the fix for is now in this PR), but the latest performance metrics of this PR are in the description.  I could maybe go further with some of the optimizations (not sure just yet), but as I'm heading out for the day, I'm leaving what I got here for now until I can take a second look.

dapplion:
Benchmarks should be the same because the code is fundamentally the same. You've replaced a for loop for a map and another for loop for a find index. The time complexity of the algorithm is the same, but maybe less obvious.

It should even be slower because callback-based iterators are less optimized than for of loops (but I'm not sure). However a basic for loop will always be much faster.

To make this faster you have to remove one loop and iterate the committee once instead of one time per validator.

dapplion:
Current algorytm loops through the entire shuffled indices array for each validator. Not only that but it returns the committee array so it is looped **again** afterwards.

```ts
for (const validatorIndex of validatorIndices) {
    const committeeAssignment = (function () {
      for (let slot = epochStartSlot; slot < epochStartSlot + SLOTS_PER_EPOCH; slot++) {
        const committeeCount = this.getCommitteeCountAtSlot(slot);
        for (let i = 0; i < committeeCount; i++) {
          const committee = this.getBeaconCommittee(slot, i);
          if (committee.includes(validatorIndex)) {
            return {
              validators: committee as List<number>,
              committeeIndex: i,
              slot,
            };
          }
        }
      }
    })();

    if (committeeAssignment) {
      let validatorCommitteeIndex = -1;
      let index = 0;
      for (const i of readonlyValues(committeeAssignment.validators)) {
        if (ssz.ValidatorIndex.equals(i, validator.index)) {
          validatorCommitteeIndex = index;
          break;
        }
        index++;
      }
      return {
        pubkey: validator.pubkey,
        validatorIndex: validator.index,
        committeeLength: committeeAssignment.validators.length,
        committeesAtSlot: epochCtx.getCommitteeCountAtSlot(committeeAssignment.slot),
        validatorCommitteeIndex,
        committeeIndex: committeeAssignment.committeeIndex,
        slot: committeeAssignment.slot,
      };
    }
  }
```

----

A more reasonable approach would be to iterate the shuffled array once for all validatorIndices and not loop twice through the committee array

3xtr4t3rr3str14l:
@dapplion I followed your suggestion and got significant performance increases:

```
number of runs:  50
assembleAttesterDuty batch performance avg:  3332.42
getCommitteeAssignments performance avg:  4.22
```

Now the solution just uses `getCommitteeAssignments` (new epochCtx batch function) directly and does the work that running `assembleAttesterDuty` in batch previously did.

The benchmark test I used to get those numbers is in this PR.  It also checks to make sure that `getCommitteeAssignments` gets the same values that batch `assembleAttesterDuty` previously did.

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -227,20 +226,16 @@ export function getValidatorApi({
...
-          config,
-          {pubkey: validator.pubkey, index: validatorIndex},
-          state.epochCtx,
-          epoch
-        );
-        if (duty) duties.push(duty);
-      }
+      const duties = state.epochCtx.getCommitteeAssignments(
+        epoch,
+        state.validators.map((validator, k) => {
```
twoeths: I suppose we need to map `validatorIndices` instead of `state.validators`

3xtr4t3rr3str14l: I use `state.validators` because we need the `pubkey` as well.

dapplion: Running `state.validators.map` must be a very slow operation, does that show up in the benchmarks? If you need the pubkey use the index2pubkey cache

packages/lodestar/test/unit/api/impl/validator/duties/attester.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    const cachedState = createCachedBeaconState(config, state);
+    chainStub.getHeadStateAtCurrentEpoch.resolves(cachedState);
+
+    indices = Array.from({length: numValidators}, (_, i) => i);
+  });
+
+  after(() => {
+    console.log("number of runs: ", numRuns);
+    console.log("assembleAttesterDuty batch performance avg: ", totalPerfAAD / numRuns);
+    console.log("getCommitteeAssignments performance avg: ", totalPerfGCA / numRuns);
```
dapplion: Please use a proper benchmark runner to measure performance. Follow this example https://github.com/ChainSafe/lodestar/blob/f52728f6b3c7b82333840613db73930bdc9111d9/packages/beacon-state-transition/test/perf/phase0/block/block.perf.ts#L48

dapplion (dismissed):

packages/lodestar/src/api/impl/validator/index.ts
3xtr4t3rr3str14l: @dapplion when I try using index2pubkey, it seems it's actually slower than using state.validators (maybe because I have to convert using `toBytes()`.  here's the benchmark (I updated the benchmark test to have all 3 of these run calls in there too):

```
get attester duties
new way: getCommitteeAssignments                                       56.21777 ops/s      17.78797 ms/op      6 runs   0.1070 s
new way (plus index2pubkey): getCommitteeAssignments                   14.70149 ops/s      68.02031 ms/op      2 runs   0.1360 s
old way: assembleAttesterDuty batch                                   0.1696719 ops/s      5.893728  s/op      1 runs    11.79 s
```

wemeetagain: I think @tuyennhv  means:
```ts
const validators = state.validators.persistent;
const duties = state.epochCtx.getCommitteeAssignments(
  epoch,
  validatorIndices.map((index) => {
    const validator = validators.get(index);
    if (!validator) {
      throw new ApiError(400, `Validator index ${index} not in state`);
    }
    return {pubkey: validator.pubkey, index};
  })
);
```

packages/beacon-state-transition/src/allForks/util/epochContext.ts
```diff
@@ -325,6 +327,36 @@ export class EpochContext {
     return validatorIndices;
   }
 
+  getCommitteeAssignments(
+    epoch: Epoch,
+    validatorData: {pubkey: BLSPubkey; index: ValidatorIndex}[]
+  ): routes.validator.AttesterDuty[] {
```
wemeetagain: I don't think this is the return type we want here?

Or maybe we do. There's also the `CommitteeAssignment` type. Maybe a chance to just stop using this `CommitteeAssignment` type which is only used during intermediate steps.

Thoughts @dapplion ?

packages/lodestar/src/api/impl/validator/index.ts
wemeetagain: As it stands, this is just passing _all_ validators' pubkey + a different index to `getCommitteeAssignments`? Doesn't seem right.

packages/beacon-state-transition/src/allForks/util/epochContext.ts
3xtr4t3rr3str14l: yeah we can change the name of this function to `getAttesterDuties` or something like that, if that's better.

packages/lodestar/src/api/impl/validator/index.ts
3xtr4t3rr3str14l: ok, made a few changes and now the benchmarks are as follows (which make a lot more sense than the original ones i posted):

```
get attester duties
new way: getCommitteeAssignments                                      0.6062110 ops/s      1.649591  s/op      2 runs    3.299 s
new way (plus index2pubkey): getCommitteeAssignments                  0.7766266 ops/s      1.287620  s/op      2 runs    2.576 s
old way: assembleAttesterDuty batch                                   0.2985871 ops/s      3.349106  s/op      2 runs    6.698 s
    ✓ performance comparison (12678ms)
```

so i'm using `index2pubkey` now

packages/lodestar/test/perf/api/impl/validator/attester.test.ts
```diff
@@ -0,0 +1,98 @@
...
+      const validator = state.validators[validatorIndex];
+      const duty = assembleAttesterDuty(config, {pubkey: validator.pubkey, index: validatorIndex}, state.epochCtx, 0);
+      if (duty) oldDuties.push(duty);
+    }
+  });
+
+  // verify that both the old and new ways get the same data
+  newDuties.sort((x, y) => x.validatorIndex - y.validatorIndex);
+  oldDuties.sort((x, y) => x.validatorIndex - y.validatorIndex);
+  expect(newDuties).to.deep.equal(oldDuties);
```
dapplion: Doing an expect() in a describe() block will kinda break mocha. expect() must only be called inside an it() block.

Tho, separate testing and benchmark in different runs. Here exclusively test performance and then assert correctnes in `test/unit/**` with a different file

packages/lodestar/test/perf/api/impl/validator/attester.test.ts
```diff
@@ -0,0 +1,98 @@
...
+      const validator = validators.get(index);
+      if (!validator) {
+        throw new Error(`Validator at index ${index} not in state`);
+      }
+      validatorData[index] = validator.pubkey;
+    });
+    newDuties = state.epochCtx.getCommitteeAssignments(0, validatorData);
+  });
+
+  itBench("new way (plus index2pubkey): getCommitteeAssignments", () => {
```
dapplion: Use simpler names with the same prefix for easier lookup. This name will be used as index and may end up in plot titles. Something like:
```
getCommitteeAssignments - index2pubkey - more data
```

packages/lodestar/test/perf/api/impl/validator/attester.test.ts
```diff
@@ -0,0 +1,98 @@
...
+);
+const state = createCachedBeaconState(config, _state);
+chainStub.getHeadStateAtCurrentEpoch.resolves(state);
+
+const indices = Array.from({length: numAttachedValidators}, (_, i) => i * 5);
+
+describe("getCommitteeAssignments vs assembleAttesterDuties performance test", async () => {
+  setBenchOpts({
+    maxMs: 10 * 1000,
+    runs: 10,
```
dapplion: 10 runs is extremely low, it will add too much variability in the results. For fast benchmarks it should do at least 1000. You should use minMs to ensure a sufficient number is run like:
```
maxMs: 10 * 1000,
minMs: 2 * 1000,
runs: 1024,
```

packages/lodestar/test/perf/api/impl/validator/attester.test.ts
```diff
@@ -0,0 +1,98 @@
...
+const _state = generateState(
+  {
+    slot: 0,
+    validators,
+    balances: generateInitialMaxBalances(config, numValidators),
+  },
+  config
+);
+const state = createCachedBeaconState(config, _state);
+chainStub.getHeadStateAtCurrentEpoch.resolves(state);
```
dapplion: Move all this logic inside the describe() block, so it's only run when calling mocha, not with just an import

packages/lodestar/test/perf/api/impl/validator/attester.test.ts
```diff
@@ -0,0 +1,98 @@
...
+        throw new Error(`Validator pubkey at validator index ${index} not found in state.`);
+      }
+      validatorData[index] = pubkey.toBytes();
+    });
+    newDuties = state.epochCtx.getCommitteeAssignments(0, validatorData);
+  });
+
+  // the old way of getting the attester duties
+  let oldDuties: AttesterDuty[] = [];
+  itBench("old way: assembleAttesterDuty batch", () => {
```
dapplion: It's not useful to benchmark old code we are not gonna use anymore. Run the benchmark, share the results on the PR to appreciate the performance improvement, then delete it.

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -227,20 +226,16 @@ export function getValidatorApi({
       // the first `MAXIMUM_GOSSIP_CLOCK_DISPARITY` duration of the epoch `tolerantCurrentEpoch`
       // will equal `currentEpoch + 1`
 
-      const duties: routes.validator.AttesterDuty[] = [];
-      for (const validatorIndex of validatorIndices) {
-        const validator = state.validators[validatorIndex];
+      const validators = state.validators.persistent;
+      const validatorData: BLSPubkey[] = [];
```
dapplion: Should this be a Map? This is going to be a hugely sparse array. Not sure if it's dangerous memory-wise for Node.

dapplion:
Would it make sense to keep a reverse shuffled list to speed this even more?

If the shuffled array looks like `[3,0,2,1]` the reverse array would be `[1,3,2,0]`. Given a validator index you could get its postion in the shuffled array in O(1) time. From there with modulos you can figure out the slot, committee and index.

Would that be worth it the extra memory? At 250_000 validators x 8 bytes = 2MB

packages/lodestar/test/perf/api/impl/validator/attester.test.ts
3xtr4t3rr3str14l: I think once we decide that the new strategy is sound, we'll just remove the oldDuties anyways, so there will be no need to compare anymore.  So we can just remove that section once the main part of this PR is complete.

packages/lodestar/test/perf/api/impl/validator/attester.test.ts
3xtr4t3rr3str14l: Right, that's the plan (https://github.com/ChainSafe/lodestar/pull/2717#discussion_r660068417)

packages/lodestar/src/api/impl/validator/index.ts
3xtr4t3rr3str14l: I'm not sure either.  @wemeetagain thoughts?

wemeetagain: I don't think its dangerous, but there's no harm in using a Map to make this clearer

wemeetagain:
> Would it make sense to keep a reverse shuffled list to speed this even more?
> 
> If the shuffled array looks like `[3,0,2,1]` the reverse array would be `[1,3,2,0]`. Given a validator index you could get its postion in the shuffled array in O(1) time. From there with modulos you can figure out the slot, committee and index.
> 
> Would that be worth it the extra memory? At 250_000 validators x 8 bytes = 2MB

Worth discussing in a separate issue. W can also take a look at what other clients do.

This seems to speed up getCommitteeAssignments 2x for small # of validators, 30x for large # of validators, so it may be "good enough"?

wemeetagain:
This lgtm

wemeetagain (dismissed):

packages/beacon-state-transition/package.json
```diff
@@ -38,6 +38,7 @@
   "types": "lib/index.d.ts",
   "dependencies": {
     "@chainsafe/bls": "6.0.1",
+    "@chainsafe/lodestar-api": "^0.25.1",
```
dapplion: beacon-state-transition must not depend on the API. Define the `AttesterDuty` type again in that repo without importing it. You may skip the pubkey since it adds no info

packages/beacon-state-transition/src/allForks/util/epochContext.ts
```diff
@@ -329,6 +331,41 @@ export class EpochContext {
     return validatorIndices;
   }
 
+  /**
+   * @param validatorData pubkeys indexed by validator index
+   */
+  getCommitteeAssignments(epoch: Epoch, validatorData: Map<number, BLSPubkey>): routes.validator.AttesterDuty[] {
```
dapplion: Change to 
```ts
getCommitteeAssignments(epoch: Epoch, requestedValidatorIndices: Set<number>): AttesterDuty[]
```
and use the set to validate if index belongs in set instead of map. The consumer of this function may add the pubkey if necessary.

packages/lodestar/package.json
```diff
@@ -36,6 +36,7 @@
     "pretest": "yarn run check-types",
     "test": "yarn test:unit && yarn test:e2e",
     "test:unit": "LODESTAR_PRESET=minimal nyc --cache-dir .nyc_output/.cache -e .ts mocha 'test/unit/**/*.test.ts'",
+    "test:perf": "LODESTAR_PRESET=minimal nyc --cache-dir .nyc_output/.cache -e .ts mocha 'test/perf/**/*.test.ts'",
```
dapplion: Benchmarks must be run with the `benchmark` runner. The `nyc` binary tracks coverage data which makes no sense in a benchmark.

Follow this script https://github.com/ChainSafe/lodestar/blob/86f9cf24b2546c362b874265539b90e3a6db50ec/packages/beacon-state-transition/package.json#L36
You may not need the extra memory if the test doesn't deal with huge states

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -227,20 +226,16 @@ export function getValidatorApi({
...
-      const duties: routes.validator.AttesterDuty[] = [];
-      for (const validatorIndex of validatorIndices) {
-        const validator = state.validators[validatorIndex];
+      const validators = state.validators.persistent;
+      const validatorData: Map<number, BLSPubkey> = new Map<number, BLSPubkey>();
+      for (const index of validatorIndices) {
+        const validator = validators.get(index);
         if (!validator) {
-          throw new ApiError(400, `Validator index ${validatorIndex} not in state`);
+          throw new ApiError(400, `Validator pubkey at index ${index} not in state`);
```
dapplion: I think the original message is more correct
```
`Validator index ${validatorIndex} not in state`
```

packages/beacon-state-transition/package.json
3xtr4t3rr3str14l: Ok, I'll copy in that AttesterDuty, but I think it might be more convenient and efficient to still have the pubkey there because otherwise we would have to re-map the pubkeys to the duties once we get the duties.

dapplion: It's def more efficient but since it's an operation you don't do 200_000 times it's doesn't really matter.

packages/lodestar/package.json
3xtr4t3rr3str14l: When I try to run that command (both when I run for `beacon-state-transition` and when i copy it and run it for `lodestar`) I get this error:

```
✖ Error: Must specify a history option
    at Object.getHistoryProvider (/home/seraph/work/ChainSafe/lodestar/node_modules/@dapplion/benchmark/lib/history/index.js:22:11)
    at Object.run (/home/seraph/work/ChainSafe/lodestar/node_modules/@dapplion/benchmark/lib/run.js:36:39)
    at Object.handler (/home/seraph/work/ChainSafe/lodestar/node_modules/@dapplion/benchmark/lib/cli.js:24:30)
    at /home/seraph/work/ChainSafe/lodestar/node_modules/@dapplion/benchmark/node_modules/yargs/build/index.cjs:1:8973
    at j (/home/seraph/work/ChainSafe/lodestar/node_modules/@dapplion/benchmark/node_modules/yargs/build/index.cjs:1:4852)
    at _.applyMiddlewareAndGetResult (/home/seraph/work/ChainSafe/lodestar/node_modules/@dapplion/benchmark/node_modules/yargs/build/index.cjs:1:8942)
    at _.runCommand (/home/seraph/work/ChainSafe/lodestar/node_modules/@dapplion/benchmark/node_modules/yargs/build/index.cjs:1:7127)
    at Bt.[runYargsParserAndExecuteCommands] (/home/seraph/work/ChainSafe/lodestar/node_modules/@dapplion/benchmark/node_modules/yargs/build/index.cjs:1:52753)
    at Bt.parse (/home/seraph/work/ChainSafe/lodestar/node_modules/@dapplion/benchmark/node_modules/yargs/build/index.cjs:1:36100)
    at Object.<anonymous> (/home/seraph/work/ChainSafe/lodestar/node_modules/@dapplion/benchmark/lib/cli.js:59:6)
```

Do you know what's up with that?

wemeetagain (dismissed):

packages/lodestar/package.json
dapplion: Yes, you must provide a history option to tell the tool where to get benchmark data from. Locally you should use `--local` to persist it in the local filesystem.

In CI it will detect `S3` ENV is set and use aws to persist benchmark data

packages/lodestar/test/perf/api/impl/validator/attester.test.ts
```diff
@@ -0,0 +1,79 @@
...
+      const validator = validators.get(index);
+      if (!validator) {
+        throw new Error(`Validator pubkey at index ${index} not in state`);
+      }
+      validatorData.set(index, validator.pubkey);
+    }
+    state.epochCtx.getCommitteeAssignments(0, validatorData);
+  });
+
+  itBench("getCommitteeAssignments - index2pubkey - more data", () => {
```
dapplion: The `more data` part was not literal but an indication of "put more data here if you need". Try to understand what the difference between this two implementations and use a good description name for each

packages/lodestar/package.json
3xtr4t3rr3str14l: I keep getting this error locally as well as in CI when I try to run `yarn run benchmark`

```
Connected to historyProvider: LocalHistoryProvider, dirpath ./benchmark_data
No previous bencharmk found for latestCommitInBranch 'master'
 ✖ TypeError: Cannot read property 'fromBytes' of undefined
    at syncPubkeys (/home/seraph/work/ChainSafe/lodestar/packages/beacon-state-transition/src/allForks/util/epochContext.ts:172:37)
    at createEpochContext (/home/seraph/work/ChainSafe/lodestar/packages/beacon-state-transition/src/allForks/util/epochContext.ts:100:5)
    at Object.createCachedBeaconState (/home/seraph/work/ChainSafe/lodestar/packages/beacon-state-transition/src/allForks/util/cachedBeaconState.ts:70:20)
    at Object.generateCachedState (/home/seraph/work/ChainSafe/lodestar/packages/lodestar/test/utils/state.ts:149:10)
    at Suite.<anonymous> (/home/seraph/work/ChainSafe/lodestar/packages/lodestar/test/perf/chain/stateCache/stateContextCheckpointsCache.test.ts:15:17)
    at Object.create (/home/seraph/work/ChainSafe/lodestar/node_modules/mocha/lib/interfaces/common.js:148:19)
    at context.describe.context.context (/home/seraph/work/ChainSafe/lodestar/node_modules/mocha/lib/interfaces/bdd.js:42:27)
    at Object.<anonymous> (/home/seraph/work/ChainSafe/lodestar/packages/lodestar/test/perf/chain/stateCache/stateContextCheckpointsCache.test.ts:7:1)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Module.m._compile (/home/seraph/work/ChainSafe/lodestar/node_modules/ts-node/src/index.ts:1056:23)
 ```
 
 Not sure why yet.

Comment (3xtr4t3rr3str14l) [reply packages/lodestar/package.json]: nvm, think i got it.  will push up once resolved

File: packages/lodestar/src/api/impl/validator/index.ts [hunk 15]
```diff
@@ -227,25 +226,19 @@ export function getValidatorApi({
...
-          {pubkey: validator.pubkey, index: validatorIndex},
-          state.epochCtx,
-          epoch
-        );
-        if (duty) duties.push(duty);
+      const validators = state.validators.persistent;
+      const duties = state.epochCtx.getCommitteeAssignments(epoch, new Set(validatorIndices));
+      const data: routes.validator.AttesterDuty[] = [];
+      for (const duty of duties) {
+        const validator = validators.get(duty.validatorIndex);
```
dapplion: Why not use the index2pubkey cache?

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -227,25 +226,19 @@ export function getValidatorApi({
...
-        const duty = assembleAttesterDuty(
-          config,
-          {pubkey: validator.pubkey, index: validatorIndex},
-          state.epochCtx,
-          epoch
-        );
-        if (duty) duties.push(duty);
+      const validators = state.validators.persistent;
+      const duties = state.epochCtx.getCommitteeAssignments(epoch, new Set(validatorIndices));
+      const data: routes.validator.AttesterDuty[] = [];
```
dapplion: Rename to `duties`

packages/lodestar/test/perf/chain/stateCache/stateContextCheckpointsCache.test.ts
```diff
@@ -1,23 +1,33 @@
 import {itBench, setBenchOpts} from "@dapplion/benchmark";
 import {config} from "@chainsafe/lodestar-config/default";
-import {ssz} from "@chainsafe/lodestar-types";
+import {allForks, ssz} from "@chainsafe/lodestar-types";
 import {generateCachedState} from "../../../utils/state";
 import {CheckpointStateCache} from "../../../../src/chain/stateCache";
+import {Checkpoint} from "../../../../../types/phase0";
+import {CachedBeaconState} from "@chainsafe/lodestar-beacon-state-transition";
 
 describe("CheckpointStateCache perf tests", function () {
```
dapplion: Mixup? This changes do not belong to this PR

packages/lodestar/src/api/impl/validator/index.ts
dapplion: Turns out using the persistent validators array is much faster. Added a benchmark to prove it.

dapplion:
> This seems to speed up getCommitteeAssignments 2x for small # of validators, 30x for large # of validators, so it may be "good enough"?

These are the current numbers in my local machine
```
    ✓ getCommitteeAssignments - req 1 vs - 200000 vc                      225.6465 ops/s    4.431711 ms/op        -       1024 runs   4.54 s
    ✓ getCommitteeAssignments - req 100 vs - 200000 vc                    141.3410 ops/s    7.075087 ms/op        -       1024 runs   7.25 s
    ✓ getCommitteeAssignments - req 1000 vs - 200000 vc                   124.7096 ops/s    8.018632 ms/op        -       1024 runs   8.25 s
```

Since getCommitteeAssignments() is something we should call twice per epoch I think it's good for now. The node will spend ~10ms every 6min doing executing this logic (0.003% rate).

Note that prior to this PR, for 1000 validators the node will do 4.4ms x 1000 = 4 sec total. AKA not acceptable

packages/lodestar/test/perf/chain/stateCache/stateContextCheckpointsCache.test.ts
3xtr4t3rr3str14l: I was getting some errors after changing the `yarn run test:perf` command to `yarn run benchmark`, so I made changes here accordingly.

twoeths (approved):
Looks good to me, just a small note that if we use `state.validators.persistent` frequently, it's maybe better to write a method in `CachedValidatorList` so that we get through the proxy a single time

dapplion:
> Looks good to me, just a small note that if we use `state.validators.persistent` frequently, it's maybe better to write a method in `CachedValidatorList` so that we get through the proxy a single time

Good idea! Can you open an issue for it with more details?

PR #2721, opened 2021-06-18
Improve processInactivityUpdates

### LINKED ISSUE #2709: Improve processInactivityUpdates
**Describe the bug**

This is from oonoonba sync, it shows that epoch transition takes significant time for `processInactivityUpdates`

Screen Shot 2021-06-16 at 13 26 00

Profiler
Screen Shot 2021-06-16 at 13 26 48

[0616_processInactivityUpdates_local.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/6660691/0616_processInactivityUpdates_local.cpuprofile.zip)

We also want to track the benchmark for altair epoch transition

twoeths:
**Motivation**

+ Altair epoch transition `processInactivityUpdates` is so slow, we need to improve it
+ This is because we access the TreeBacked too frequently there

**Description**
+ Create `CachedInactivityScoreList` similar to `CachedBalanceList`: restrict the access to the tree by using persistent-ts
+ Inside `processInactivityUpdates`, only update the inactivity score if necessary
+ As tested in the benchmark, it's a 3x improvement

Also, I refactored the performance test util to support altair

Closes #2709


packages/beacon-state-transition/src/altair/epoch/processInactivityUpdates.ts
```diff
@@ -16,16 +16,21 @@ export function processInactivityUpdates(state: CachedBeaconState<altair.BeaconS
...
+        inactivityscore -= Math.min(1, inactivityscore);
       } else {
-        state.inactivityScores[i] += Number(INACTIVITY_SCORE_BIAS);
+        inactivityscore += Number(INACTIVITY_SCORE_BIAS);
       }
       if (!inActivityLeak) {
-        state.inactivityScores[i] -= Math.min(Number(INACTIVITY_SCORE_RECOVERY_RATE), state.inactivityScores[i]);
+        inactivityscore -= Math.min(Number(INACTIVITY_SCORE_RECOVERY_RATE), inactivityscore);
+      }
+      if (inactivityscore !== state.inactivityScores[i]) {
```
dapplion: Instead of doing a second state read you could keep the previous value scoped here like `inactivityscorePrev` then
```ts
if (inactivityscore !== inactivityscorePrev) ...
```

dapplion:
Overall looks great! @wemeetagain can you take a look?

wemeetagain (commented):
Approach looks good to me.
Just needs to be updated for merge conflicts and new benchmarking framework

dapplion:
@wemeetagain We've debugged the perf issue with Tuyen and the root cause is that the test state changed. It had more active balance thus required more processing. Feel free to merge even if the performance tests are failing.

wemeetagain (approved):

PR #2725, opened 2021-06-18
Update abort-controller import

wemeetagain:
**Motivation**

See https://github.com/mysticatea/abort-controller/issues/32
Due to a missing export statement in the `abort-controller` package, using the named export `AbortController` will break when run in the browser.

**Description**

Update all named import `AbortController` with the default import, named the same.
This goes against one of our linter rules `import/no-named-as-default`, so every import is accompanied with an eslint-disable statement.

We should discuss whether or not this is the best course of action, considering the underlying issue may come back when we use `AbortController` in the future (and use the named export).

### DIFF
--- a/packages/api/src/client/utils/httpClient.ts
+++ b/packages/api/src/client/utils/httpClient.ts
@@ -1,5 +1,5 @@
 import {fetch} from "cross-fetch";
-import {AbortSignal, AbortController} from "abort-controller";
+import AbortController, {AbortSignal} from "abort-controller";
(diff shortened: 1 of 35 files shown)


dapplion (approved):

mpetrunic:
should we rather fork and use our fork until they accept and release our fix?

wemeetagain:
I just released `@chainsafe/abort-controller` v3.0.1 with the fix

Closing this in favor of using that for now

PR #2730, opened 2021-06-22
Poll node in batch to get validator indices

### LINKED ISSUE #2727: Validator: attestation and sync committee services stop working
**Describe the bug**

I synced oonoonba with 160 validators, got this right after I start validator:
```
Jun-20 10:56:46.448 []                 info: Verified node and validator have same config
Jun-20 10:56:46.546 []                error: Error on poll indices and attesters epoch=4118 message=Bad Request: Client Error
Error: Bad Request: Client Error
    at HttpClient.request (/usr/app/packages/api/src/client/utils/httpClient.ts:85:15)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at HttpClient.json (/usr/app/packages/api/src/client/utils/httpClient.ts:57:12)
    at Object.request [as getStateValidators] (/usr/app/packages/api/src/client/utils/client.ts:71:19)
    at IndicesService.pollValidatorIndicesInternal (/usr/app/packages/validator/src/services/indices.ts:62:29)
Jun-20 10:56:46.574 []                error: Error on poll indices and SyncDuties epoch=4118 message=Bad Request: Client Error
Error: Bad Request: Client Error
    at HttpClient.request (/usr/app/packages/api/src/client/utils/httpClient.ts:85:15)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at HttpClient.json (/usr/app/packages/api/src/client/utils/httpClient.ts:57:12)
    at Object.request [as getStateValidators] (/usr/app/packages/api/src/client/utils/client.ts:71:19)
    at IndicesService.pollValidatorIndicesInternal (/usr/app/packages/validator/src/services/indices.ts:62:29)
(node:1) UnhandledPromiseRejectionWarning: Error: Bad Request: Client Error
    at HttpClient.request (/usr/app/packages/api/src/client/utils/httpClient.ts:85:15)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at HttpClient.json (/usr/app/packages/api/src/client/utils/httpClient.ts:57:12)
    at Object.request [as getStateValidators] (/usr/app/packages/api/src/client/utils/client.ts:71:19)
    at IndicesService.pollValidatorIndicesInternal (/usr/app/packages/validator/src/services/indices.ts:62:29)
(Use `node --trace-warnings ...` to show where the warning was created)
```

I don't see attestation + sync committee services' logs, only block proposal service continues after that
```
Jun-20 10:59:20.358 []                 info: Published block slot=131792, validator=0x91a4…f7e2, graffiti=chainsafe/lodestar-0.24.3
```

also, no error at node side

**Expected behavior**

No error

twoeths:
**Motivation**

+ Imported 160 oononba validators into contabo-1 and it failed to request validator indices due to the query max length

**Description**

+ Query in batches, max batch=5
+ Each batch contains at most 5 http requests to the node
+ Each http request could work on 40 validators at most
+ Work on at most 1000 validator pubkeys per poll

Let me know if we want to tweak the above numbers.

Closes #2727

**Tested in contabo-1**
```
Jun-22 09:24:26.394 []                 info: Decrypted 160 validator keystores
Jun-22 09:24:26.459 []                 info: Genesis available
Jun-22 09:24:26.477 []                 info: Verified node and validator have same config
Jun-22 09:24:26.727 []                 info: Discovered new validators count=160
Jun-22 09:24:36.614 []                 info: Published SyncCommitteeSignature slot=145718, count=160
Jun-22 09:24:36.832 []                 info: Published attestations slot=145718, index=0, count=5
Jun-22 09:24:40.068 []                 info: Published SyncCommitteeContribution slot=145718, index=1, count=1
```


dapplion:
Looks good! What numbers do Teku use? We could base ours on them

twoeths:
> Looks good! What numbers do Teku use? We could base ours on them

@dapplion  it's 10 per batch https://github.com/ConsenSys/teku/blob/79e49c48e0a6a78eac5b7760860d680a046e78af/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/RemoteValidatorApiHandler.java#L75

I think it's quite conservative there, I used to be able to query 80 validator indices successfully, the issue only happened with 160 validators per batch. So
+ 10 as Teku
+ or 40 as in this PR
+ or 20 which is in the middle

wemeetagain:
Teku may be going off of this: https://stackoverflow.com/questions/3091485/what-is-the-limit-on-querystring-get-url-parameters#answer-56527673 I've seen this answer copy/pasted around, tho I can't find the original

Also related, I just posted this, maybe this can be more definitely resolved at the standards level: https://github.com/ethereum/eth2.0-APIs/issues/153

packages/validator/test/unit/services/indices.test.ts
```diff
@@ -0,0 +1,50 @@
...
+          ["1", "2"],
+          ["3", "4"],
+          ["5", "6"],
+        ],
+        [["7"]],
+      ],
+    },
+  ];
+  const maxPubkeysPerRequest = 2;
+  const maxRequesPerBatch = 3;
```
3xtr4t3rr3str14l, line 44:
```suggestion
  const maxRequestPerBatch = 3;
```

3xtr4t3rr3str14l: typo, should be `maxRequestPerBatch`

packages/validator/src/services/indices.ts
```diff
@@ -71,7 +85,42 @@ export class IndicesService {
...
+ * each request can work on at most 40 pubkeys.
+ * @param pubkeysHex
+ */
+export function pubkeysToBatches(
+  pubkeysHex: string[],
+  maxPubkeysPerRequest: number = PUBKEYS_PER_REQUEST,
+  maxRequesPerBatch: number = REQUESTS_PER_BATCH
+): Batch[] {
+  if (!pubkeysHex || pubkeysHex.length === 0) {
+    return [[[]]];
```
dapplion: Why nesting three levels deep? We run a Promise.all() inside a Promise.all() so it should be equivalent to having a `[[]]` data structure + a single Promise.all()

twoeths: I intended to make a `for loop` instead but missed that. So the intention is:
+ Given 10 validators per http request,  with 2000 validators we'll have to run 200 http requests
+ So I divided into `x` batches, each batch contains `y` http requests and run each batch one by one. We can tweak number of requests per batch to a suitable number

I was afraid running all http requests at the same time is not a scalable solution if we increase the validators, do you think so?

dapplion: > I was afraid running all http requests at the same time is not a scalable solution if we increase the validators, do you think so?

Yeah I agree but aren't we just doing all the HTTP calls at the same time?

twoeths: I should change the 1st `Promise.all` as a `for loop` instead

packages/validator/src/services/indices.ts
```diff
@@ -71,7 +84,42 @@ export class IndicesService {
...
+
+/**
+ * Divide pubkeys into batches, each batch contains at most 5 http requests,
+ * each request can work on at most 40 pubkeys.
+ * @param pubkeysHex
+ */
+export function pubkeysToBatches(
+  pubkeysHex: string[],
+  maxPubkeysPerRequest: number = PUBKEYS_PER_REQUEST,
+  maxRequesPerBatch: number = REQUESTS_PER_BATCH
```
dapplion: typo `maxRequestsPerBatch`

packages/validator/src/services/indices.ts
```diff
@@ -71,7 +84,42 @@ export class IndicesService {
...
+): Batch[] {
+  if (!pubkeysHex || pubkeysHex.length === 0) {
+    return [[[]]];
+  }
+  const batches: Batch[] = [];
+
+  const pubkeysPerBatch = maxPubkeysPerRequest * maxRequesPerBatch;
+  let batch: Batch = [];
+  let pubkeysPerRequest: string[];
```
dapplion: Can find a similar approach without mutating scoped variables? You could have two for loops and scope the arrays inside them only with const

wemeetagain (approved):

Issue #1554, opened 2020-09-20
Massive repo size


dapplion:
`ChainSafe/lodestar` currently weighs 131.75MB on clone. This is way too much for a src, and we should fix it.

```
lion@lion:~/Code/eth2.0$ git clone git@github.com:ChainSafe/lodestar.git
Cloning into 'lodestar'...
remote: Enumerating objects: 3768, done.
remote: Counting objects: 100% (3768/3768), done.
remote: Compressing objects: 100% (681/681), done.
remote: Total 303063 (delta 2007), reused 3339 (delta 1669), pack-reused 299295
Receiving objects: 100% (303063/303063), 131.75 MiB | 892.00 KiB/s, done.
Resolving deltas: 100% (185861/185861), done.
```

There seems to be a binary of prometheus in the scripts directory https://github.com/ChainSafe/lodestar/blob/master/scripts/prometheus is that an error?

mpetrunic:
i'm in favor of removing that. We could create some docker-compose to set those things up

dapplion:
Repo size is still huge since this file will leave forever in the git history https://github.com/ChainSafe/lodestar/commit/79ac7f76

The only way to get rid of it is by re-writing this repo's git history from that commit (see the process below). It's quite dramatic because all existing tags and commits hash will change, but it's doable.

```     
ubuntu@lodestar-cloud-01:~$ java -jar bfg-1.13.2.jar  --strip-blobs-bigger-than 50M lodestar/

Using repo : /home/ubuntu/lodestar/.git

Scanning packfile for large blobs: 614687
Scanning packfile for large blobs completed in 9,751 ms.
Found 1 blob ids for large blobs - biggest=84346472 smallest=84346472
Total size (unpacked)=84346472
Found 1180 objects to protect
Found 3 tag-pointing refs : refs/tags/v0.13.0, refs/tags/v0.14.0, refs/tags/v0.6.0
Found 65 commit-pointing refs : HEAD, refs/heads/master, refs/remotes/origin/HEAD, ...

Protected commits
-----------------


These are your protected commits, and so their contents will NOT be altered:

 * commit 83f69eee (protected by 'HEAD')

Cleaning
--------

Found 7221 commits
Cleaning commits:       100% (7221/7221)
Cleaning commits completed in 15,058 ms.

Updating 65 Refs
----------------

	Ref                                                                      Before     After   
	--------------------------------------------------------------------------------------------
[...]

PR #2732, opened 2021-06-22
Add brand assets

sweetpea22:
**Motivation**

Adding new brand assets ✨

**Description**

You'll need to delete the old ones after, didn't want to remove them yet at risk of breaking live imgs. 

suggestions: 
- use the square icon for touch icon in index.html
- when using full mark for web apps, set maxHeight to 40px 
- stacked mark works well when we're sitting next to other brands :D


wemeetagain (dismissed):
:rocket: :rocket:

wemeetagain:
@dapplion we might save & delete a few of the larger files to keep the repo size down. See https://github.com/ChainSafe/lodestar/issues/1554

dapplion (approved):

dapplion (dismissed):
Agree on triming the size down a bit!

wemeetagain:
I'm going to commit over this, removing the MB+ images. I've got them saved locally so we can still use them where desired.

wemeetagain (approved):

PR #2741, opened 2021-06-23
Db read/write metrics

### LINKED ISSUE #2603: Add metrics for total db read / writes

Add metrics for total db read / writes. Quantify by Bucket how much we use the db.

Add a total counter labeled by Bucket name

mpetrunic:
**Motivation**

**Description**

Closes #2603 

### DIFF
--- a/packages/cli/src/cmds/beacon/handler.ts
+++ b/packages/cli/src/cmds/beacon/handler.ts
@@ -49,11 +49,16 @@ export async function beaconHandler(args: IBeaconArgs & IGlobalArgs): Promise<vo
 
   logger.info("Lodestar", {version: lodestarGitData.version, network: args.network});
 
-  const dbMetrics = createDbMetrics();
+  let dbMetrics: null | ReturnType<typeof createDbMetrics> = null;
+  const metricRegistries = [];
+  if (options.metrics.enabled) {
+    dbMetrics = createDbMetrics();
+    metricRegistries.push(dbMetrics.registry);
+  }
   const db = new BeaconDb({
     config,
     controller: new LevelDbController(options.db, {logger: logger.child(options.logger.db)}),
-    metrics: dbMetrics.metrics,
+    metrics: dbMetrics?.metrics,
   });
 
   await db.start();
@@ -68,7 +73,7 @@ export async function beaconHandler(args: IBeaconArgs & IGlobalArgs): Promise<vo
       logger,
       libp2p: await createNodeJsLibp2p(peerId, options.network, {peerStoreDir: beaconPaths.peerStoreDir}),
       anchorState,
-      metricRegistries: [dbMetrics.registry],
+      metricRegistries,
     });
 
     abortController.signal.addEventListener("abort", () => node.close(), {once: true});
--- a/packages/cli/src/cmds/beacon/handler.ts
+++ b/packages/cli/src/cmds/beacon/handler.ts
@@ -13,6 +13,7 @@ import {initBLS, onGracefulShutdown, getCliLogger} from "../../util";
 import {readLodestarGitData} from "../../util/gitData";
 import {FileENR, overwriteEnrWithCliArgs, readPeerId} from "../../config";
 import {initBeaconState} from "./initBeaconState";
+import {createDbMetrics} from "@chainsafe/lodestar/lib/metrics";
 
 /**
  * Run a beacon node
@@ -48,9 +49,11 @@ export async function beaconHandler(args: IBeaconArgs & IGlobalArgs): Promise<vo
 
   logger.info("Lodestar", {version: lodestarGitData.version, network: args.network});
 
+  const dbMetrics = createDbMetrics();
   const db = new BeaconDb({
     config,
     controller: new LevelDbController(options.db, {logger: logger.child(options.logger.db)}),
+    metrics: dbMetrics.metrics,
   });
 
   await db.start();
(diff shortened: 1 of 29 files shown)


packages/db/src/abstractRepository.ts
```diff
@@ -63,14 +78,18 @@ export abstract class Repository<I extends Id, T> {
...
   }
 
   async putBinary(id: I, value: Buffer): Promise<void> {
+    this.dbWriteMetrics?.inc();
     await this.db.put(this.encodeKey(id), value);
   }
 
   async delete(id: I): Promise<void> {
+    // do we wanna intorduce db deletes or treat delete as write?
+    // this.dbWriteMetrics?.inc();
```
dapplion: I think threating delete as write is a great idea, let's do it

packages/db/src/abstractRepository.ts
```diff
@@ -124,35 +145,41 @@ export abstract class Repository<I extends Id, T> {
...
       yield decodeKey(key);
     }
   }
   async values(opts?: IFilterOptions<I>): Promise<T[]> {
+    this.dbReadsMetrics?.inc();
     const data = await this.db.values(this.dbFilterOptions(opts));
     return (data || []).map((data) => this.decodeValue(data));
   }
   async *valuesStream(opts?: IFilterOptions<I>): AsyncIterable<T> {
+    this.dbReadsMetrics?.inc();
```
dapplion, line 166: Here we are counting 1 for potentially reading the entire DB. However for writes we count 2 if writing two items. Should we just count 1 for batch writes? Level DB optimizes batch write opts so maybe our metrics should help us optimize DB usage in the direction, rewarding using batch writes vs one-off writes

mpetrunic: How would you count it since you don't know how many results we got though. I feel like this should be separate metrics like number of read streames opened.

packages/db/src/abstractRepository.ts
```diff
@@ -23,11 +25,22 @@ export abstract class Repository<I extends Id, T> {
...
+    db: IDatabaseController<Buffer, Buffer>,
+    bucket: Bucket,
+    type: Type<T>,
+    metrics?: IDbMetrics
+  ) {
     this.config = config;
     this.db = db;
     this.bucket = bucket;
     this.type = type;
+    this.dbReadsMetrics = metrics?.dbReads.labels({bucket: getBucketNameByValue(bucket) ?? ""});
```
dapplion: Should you compute getBucketNameByValue() only once and re-use?

packages/db/src/abstractRepository.ts
mpetrunic: Does it make more sense to add delete metric instead?

packages/db/src/util.ts
```diff
@@ -0,0 +1,12 @@
...
+export function getBucketNameByValue<T extends Bucket>(enumValue: T): keyof typeof Bucket | null {
+  const keys = Object.keys(Bucket).filter((x) => {
+    if (isNaN(parseInt(x))) {
+      return Bucket[x as keyof typeof Bucket] == enumValue;
+    } else {
+      return false;
+    }
+  }) as (keyof typeof Bucket)[];
+  return keys.length > 0 ? keys[0] : null;
+}
```
dapplion, line 15: Should we throw here instead of returning null? If for some reason this function returns null the metrics will break and it won't be immediately obvious why. Adding a unit test should help ensure that it doesn't break on production first.

packages/db/src/abstractRepository.ts
mpetrunic: Yes, since every repository is separate instance of abastract repo, it's safe to compute bucket name only once in constructor if that was your question.

packages/db/src/util.ts
mpetrunic: I should test it but my understanding is that it will only go in metrics with no bucket label

packages/lodestar/src/db/single/preGenesisState.ts
```diff
@@ -2,26 +2,30 @@ import {TreeBacked, ContainerType} from "@chainsafe/ssz";
...
     this.config = config;
     this.db = db;
-    this.bucket = Bucket.phase0_preGenesisState as Bucket;
+    this.bucket = Bucket.phase0_preGenesisState;
     this.key = Buffer.from(new Uint8Array([this.bucket]));
+    this.metrics = metrics;
   }
 
   async put(value: TreeBacked<allForks.BeaconState>): Promise<void> {
+    this.metrics?.dbWrites.labels({bucket: "phase0_preGenesisState"}).inc();
```
dapplion, line 23: Why is it necessary to manually call the metrics here? Shouldn't `this.db` call it on the Abstract class?

packages/cli/src/cmds/beacon/handler.ts
```diff
@@ -48,9 +49,11 @@ export async function beaconHandler(args: IBeaconArgs & IGlobalArgs): Promise<vo
 
   logger.info("Lodestar", {version: lodestarGitData.version, network: args.network});
 
+  const dbMetrics = createDbMetrics();
```
dapplion: Metrics are disabled by default, here seems that they will always exists even if the user doesn't use --metrics.enable

packages/lodestar/src/db/single/preGenesisState.ts
mpetrunic: this repo is not extending AbstractRepository. this.db is controller, not repository

packages/cli/src/cmds/beacon/handler.ts
mpetrunic: you are right, will fix

packages/db/src/abstractRepository.ts
dapplion: I don't think so, I would just do `this.dbWriteMetrics?.inc();`

packages/db/src/abstractRepository.ts
dapplion: What you did for reads is fine. I would change for writes to always inc(1) without counting the num of items in the batch

dapplion (dismissed):

wemeetagain (approved):

Issue #2729, opened 2021-06-21
Remove travis CI badges in READMEs


wemeetagain:
**Describe the bug**
There are a lot of stale badges in this repo and in our other repos

**Expected behavior**

Links/badges should only exist to relevant items.
Old/stale items should be removed

dapplion:
Thank you @3xtr4t3rr3str14l !

PR #2743, opened 2021-06-23
remove stale badges

3xtr4t3rr3str14l:
part of #2729

### DIFF
--- a/packages/lodestar/README.md
+++ b/packages/lodestar/README.md
@@ -1,6 +1,5 @@
 # Lodestar
 
-[![](https://img.shields.io/travis/com/ChainSafe/lodestar/master.svg?label=master&logo=travis "Master Branch (Travis)")](https://travis-ci.com/ChainSafe/lodestar)
 [![Discord](https://img.shields.io/discord/593655374469660673.svg?label=Discord&logo=discord)](https://discord.gg/aMxzVcr)
 ![ETH2.0_Spec_Version 0.12.1](https://img.shields.io/badge/ETH2.0_Spec_Version-0.12.1-2e86c1.svg)
 ![ES Version](https://img.shields.io/badge/ES-2020-yellow)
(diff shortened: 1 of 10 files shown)


wemeetagain:
As a followup, we should add a badge for the github action

wemeetagain (approved):

packages/lodestar/README.md
```diff
@@ -1,6 +1,5 @@
 # Lodestar
 
-[![](https://img.shields.io/travis/com/ChainSafe/lodestar/master.svg?label=master&logo=travis "Master Branch (Travis)")](https://travis-ci.com/ChainSafe/lodestar)
 [![Discord](https://img.shields.io/discord/593655374469660673.svg?label=Discord&logo=discord)](https://discord.gg/aMxzVcr)
 ![ETH2.0_Spec_Version 0.12.1](https://img.shields.io/badge/ETH2.0_Spec_Version-0.12.1-2e86c1.svg)
```
wemeetagain: Looks like there's still some stale badges

wemeetagain (changes requested):

packages/lodestar/README.md
3xtr4t3rr3str14l: do you think we should remove that eth2 spec version badge?  on one hand, we'd have to update it every so often, but on the other hand, it might be a good thing to keep us in check and make sure we actually are up to spec.  if we wanna keep it, i can go through the past version tags in https://github.com/ethereum/eth2.0-specs/tags since 0.12.1 and make sure we are up to spec (unless you already know if we are or not).

wemeetagain: I think we could just bump them all to 1.0.0, and once the altair release is locked down, then bump to 1.1.0

wemeetagain (approved):

PR #2745, opened 2021-06-23
fix: Don't hash checkpoint as cache id

### LINKED ISSUE #2685: Don't hash checkpoint as cache id
**Describe the bug**

In checkpoint state cache we hash the Checkpoint SSZ object to get a unique ID

https://github.com/ChainSafe/lodestar/blob/1a68e18dd69983e6fcbb2a407e0d650a4e1e8f69/packages/lodestar/src/chain/stateCache/stateContextCheckpointsCache.ts#L25

**Expected behavior**

We can just do concat the root and epoch to get a much cheaper also unique id

3xtr4t3rr3str14l:
resolves #2685 

benchmark results (test included in this PR):

```
this PR:
  CheckpointStateCache: add, get, delete     20073.07 ops/s      49.81800 us/op   2013 runs   0.1020 s

master:
  CheckpointStateCache: add, get, delete     6836.251 ops/s      146.2790 us/op    683 runs   0.1010 s
```

Also adds the `yarn run test:perf` command and puts the benchmark for this PR in the test/perf folder


packages/lodestar/src/chain/stateCache/stateContextCheckpointsCache.ts
```diff
@@ -21,13 +21,17 @@ export class CheckpointStateCache {
     this.config = config;
   }
 
+  checkpointToKey(cp: phase0.Checkpoint): string {
```
twoeths: this could be either private or a standalone function staying out of this class

packages/lodestar/test/unit/chain/stateCache/stateContextCheckpointsCache.test.ts
```diff
@@ -0,0 +1,31 @@
...
+
+describe("CheckpointStateCache perf tests", function () {
+  this.timeout(0);
+
+  it("should run test", async () => {
+    const checkpointStateCache = new CheckpointStateCache(config);
+    const state = generateCachedState();
+    const checkpoint = ssz.phase0.Checkpoint.defaultValue();
+    const runner = new BenchmarkRunner("CheckpointStateCache perf test", {
+      maxMs: 999999 * 1000,
```
dapplion: Why 999999 sec? You are telling the benchmark to run for 11 days if necessary in order to complete the 64 runs. You should set minMs to 1000 ms, runs to 1024, otherwise there's to much variability

packages/lodestar/test/unit/chain/stateCache/stateContextCheckpointsCache.test.ts
```diff
@@ -0,0 +1,31 @@
+import {config} from "@chainsafe/lodestar-config/default";
+import {BenchmarkRunner} from "@chainsafe/lodestar-utils/test_utils/benchmark";
+import {ssz} from "@chainsafe/lodestar-types";
+import {generateCachedState} from "../../../utils/state";
+import {CheckpointStateCache} from "../../../../src/chain/stateCache";
+
+describe("CheckpointStateCache perf tests", function () {
+  this.timeout(0);
```
dapplion: Why timeout(0) here?

3xtr4t3rr3str14l: old, will remove

packages/lodestar/test/unit/chain/stateCache/stateContextCheckpointsCache.test.ts
3xtr4t3rr3str14l: when i init the BenchmarkRunner with these options:
```
const runner = new BenchmarkRunner("CheckpointStateCache perf test", {
  minMs: 1000,
  runs: 1024,
});
```

I get these results:
```
this PR:
  CheckpointStateCache: add, get, delete     21211.61 ops/s      47.14400 us/op  20930 runs    1.002 s
master:
  CheckpointStateCache: add, get, delete     13629.18 ops/s      73.37200 us/op  13501 runs    1.002 s
```

It doesn't do the number of runs that I ask it, but the total run time (1.002 s) is consistent between both.  Is this expected?

3xtr4t3rr3str14l: (my original options for BenchmarkRunner had the same thing happen (just with different numbers), but still just wondering if that's expected)

packages/lodestar/test/unit/chain/stateCache/stateContextCheckpointsCache.test.ts
```diff
@@ -0,0 +1,28 @@
+import {config} from "@chainsafe/lodestar-config/default";
+import {BenchmarkRunner} from "@chainsafe/lodestar-utils/test_utils/benchmark";
+import {ssz} from "@chainsafe/lodestar-types";
+import {generateCachedState} from "../../../utils/state";
+import {CheckpointStateCache} from "../../../../src/chain/stateCache";
+
+describe("CheckpointStateCache perf tests", function () {
```
dapplion: You are running a performance test in the unit tests run. You must ensure that everything run in the `yarn test:unit` completes very quickly ideally in < 100 ms.

This perf test must be run through a different command if applicable.

wemeetagain (approved):

Issue #2726, opened 2021-06-18
Missed blocks -  can't produce blocks on time


dapplion:
**Describe the bug**

Watching our Prater validators, all blocks produced with Lodestar are orphaned:
- https://prater.beaconcha.in/validator/200983#blocks (node with 1 validator)
- https://prater.beaconcha.in/validator/200984#blocks (node with 16 validators)

I tracked a missed block proposal in the REST API metrics and got:
- produceBlock: 4.01s
- publishBlock: 64.1ms

**Expected behavior**

Blocks must be produced way faster. Otherwise we won't get attestations on it and the network will ignore it. Other clients produce blocks in ~ 100ms tops if I recall.

- [ ] Add benchmark to produceBlock
- [ ] Debug and improve performance

twoeths:
the produceBlock itself may not be an issue as when I call that api in my server without any validator, it returned immediately.

a lot of apis takes ~4s as in http://167.86.102.209:3000/d/lodestar/lodestar?orgId=1&refresh=30s, for example `getAttesterDuties` or `produceAttestationData`

we need to improve out node in general, it's most likely it's the attestation gossip tasks as shown below from a node with full attestation subscriptions
<img width="1301" alt="Screen Shot 2021-06-24 at 15 05 39" src="https://user-images.githubusercontent.com/10568965/123226284-b7c6b700-d4fd-11eb-8d11-11b1c2eac4c8.png">
[0624-prater-full-attestation-subscriptions.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/6707467/0624-prater-full-attestation-subscriptions.cpuprofile.zip)

dapplion:
> we need to improve out node in general, it's most likely it's the attestation gossip tasks as shown below from a node with full attestation subscriptions

I worry if it's that too, however some API calls take very little time. Maybe we should have a script next to the node that calls a free API method like version every second and plot response times through a full epoch.

3xtr4t3rr3str14l:
Regardless of what the cause may be, I have updated the #2746 to include the (mostly, a few stubs here and there) full test with benchmark results for `produceBlock` so we can at least easily know how it's performing.

dapplion:
Recent metrics across all our infra show that we produce all blocks in time

PR #2746, opened 2021-06-24
produceBlock performance test

3xtr4t3rr3str14l:
part of #2726 

Adds a performance test to benchmark `produceBlock()`

current results:

```
produce block
benchmark produceBlock            2209.037 ops/s      452.6860 us/op   2201 runs    1.001 s
```


dapplion:
This benchmark shows that producing a block costs 0.000452 seconds. While valuable, note that we are still missing 99.9999% of the real time in production shown here https://github.com/ChainSafe/lodestar/issues/2726

PR #2753, opened 2021-06-27
Fix bug in aggregator verification

dapplion:
**Motivation**

Note: **must** use bytesToBigInt() otherwise a JS number is not able to represent the latest digits of the remainder, resulting in `14333559117764833000` for example, where the last three digits are always zero. Using bytesToInt() may cause isSelectionProofValid() to always return false.

**Description**

- Use `bytesToBigInt()`
- Use correct constant


dapplion:
This bug slipped through because it's not covered by spec tests. This logic is purely p2p, so:
- it's not covered by any spec tests, nor unit tests
- it won't cause a consensus failure in our nodes

To prevent this from happening again we should move all logic not used in the state transition from the beacon-state-transition package. Put it somewhere else (like in the lodestar package) and heavily test it.

twoeths:
seems like we have the other` isSyncCommitteeAggregator` in `packages/validator/src/util/aggregator.ts` which uses BigInt there, we should deduplicate it

wemeetagain (approved):
Good catch!

PR #2756, opened 2021-06-28
Simplify BlsVerifier instantiation

dapplion:
**Motivation**

Remove intermediate wrapper class BlsVerifier to use the `implements` pattern directly. Allows for easier testing and prototyping being able to use the single thread class directly.

**Description**

- Remove intermediate class BlsVerifier
- Move switching between single-multi thread variants to chain class level


wemeetagain (approved):
Awesome, much better

PR #2757, opened 2021-06-28
Remove unused naive code

dapplion:
**Motivation**

We still have naive code in the beacon-state-transition package. This PR deletes code that's not used by anyone. There is still some work to do to not use any naive code at all.

**Description**

- Remove unused naive code


wemeetagain:
conflict, otherwise lgtm

wemeetagain (approved):

Issue #1523, opened 2020-09-09
Discussion: unit test stubbing


dapplion:
#### Background

When doing the eth1 data for block processing refactor I had to update a lot of stubbed unit tests just because I changed the implementation slightly. A great example of a stubbed unit test is this one:

https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/test/unit/chain/factory/block/blockAssembly.test.ts

What is this test trying to verify or prove? Apparently, it only asserts a bunch of functions are called in the implementation. It's almost an identical implementation of the original implementation but with more code.

#### Proposal

In my personal opinion, the test linked above is more of a liability than an asset to this codebase. I would like to know other contributors' thoughts about testing to try to improve it going forward.

1. Whenever possible isolate all logic into pure functions that can be tested just by feeding them values and asserting the result.
2. "Delegator functions" like `blockAssembly` should not be tested at all. Prevented that, (1) their implementation is so simple that a test is not necessary and (2) exist an end-to-end test (maybe many abstractions above) that uses the function and would fail if delegation is not done properly.

mpetrunic:
blockAssembly method is not simple and it currently just checks if in happy path all block fields are set. Value correctness are not tested as it depends on other methods which we expect to work correctly. What this test do lack is most important part of unit testing, error paths like state context is missing, head block is missing etc.

dapplion:
Following your point I would suggest do abstract blockAssembly to fit this layout
```ts
async function doSomething(db: Db) {
  const values = await db.something.values();
  return doSomethingLogic(values);
}
```
Where `doSomething` exclusively collects values and calls pure functions encapsulating the logic. Then you test error paths and everything else in `doSomethingLogic` with mock values only, instead of stubbed classes.

**EDIT**
This issue is inspired by the feeling that the current usage of stubs in the lodestar codebase is not optimal. In many tests stubs are defined in the parent scope with `any` type which is evil
https://github.com/ChainSafe/lodestar/blob/3492e8e7973baeb4a75e4575d7017cfa2310cd67/packages/lodestar-beacon-state-transition/test/unit/stateTransition/block/operations/voluntaryExit.test.ts#L16
As a new-comer, I have a hard time reading the tests and understanding what they are trying to prove. Most of the code ends up being boilerplate that doesn't add meaningful information. The goal of this issue is to find better ways to test the codebase in more elegant and correct ways.

mpetrunic:
[...]

Issue #2748, opened 2021-06-24
Improve processAttestation


twoeths:
**Describe the bug**

One of the most heavy operation from this profile is `processAttestation` - signature verification, it takes more than 23% of the time
<img width="1335" alt="Screen Shot 2021-06-24 at 15 45 55" src="https://user-images.githubusercontent.com/10568965/123232374-4ab62000-d503-11eb-9e2c-546707476712.png">

1. In some cases when we pass errorred attestations to chain, we make `verifySignature` as true by default just in the end put these errorred attestations in `eventHandlers.onErrorAttestation`
2. In `eventHandler` when we reprocess attestations from pending blocks or new slots we process attestations separately

**Expected behavior**
1. We should make `verifySignature` as false in scenario 1 above. When we reprocess these errored attestations, we'll verify signatures again anyway
2. In scenario 2, we want to do signature verification in batch.

[0624-prater-full-attestation-subscriptions.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/6707770/0624-prater-full-attestation-subscriptions.cpuprofile.zip)

**Steps to Reproduce**

Sync prater with full attestation subscriptions

twoeths:
not an issue anymore with latest gossip attestation update #2760

PR #2760, opened 2021-06-28
Gossip attestation validation review

### LINKED ISSUE #2771: Don't re-index attestations
**Describe the bug**

At least in event handlers we re-index attestations that have already been indexed for signature verification.

**Expected behavior**

Don't re-index attestations if not necessary

dapplion:
**Motivation**

Our synced nodes struggle to process the attestations and aggregates received through gossip. Metrics show that the node may reject between 30-99% of those objects when connected to +16 subnets.

To start addressing the problem I did an in-depth review of the gossip validation conditions focusing on performance.

**Description**

 I've refactored the gossip validation logic completely basing it heavily on Lighthouse approach. I've ensured that we satisfy all spec conditions by adding a comment for each condition.

**Important changes**

- Deleted the future / unknown target attestation cache. I strongly believe it's very unsafe, since it's completely unbounded. Our implementation is dangerous and we will be better off without anything. In the short term we should look at implementing a DOS resistant cache. There is this great write-up by Paul (Lighthouse) about the complexities of a safe attestation cache https://hackmd.io/@CzkIH-W1R42R-S4vdcDJeQ/rJmmumLPu
- Since there are no pending / cached attestations the entire processor architecture + job wrappers can be deleted for simplicity

- I've changed the validation unit tests from using stubs to using real data. The tests are much shorter now and easier to follow. They run pretty fast also, taking 40-70ms for all 10 validation tests. I strongly argue in favor of phasing out stubbing where possible, circling back to this discussion https://github.com/ChainSafe/lodestar/issues/1523

Closes https://github.com/ChainSafe/lodestar/issues/2771


packages/lodestar/src/chain/validation/aggregateAndProof.ts
```diff
@@ -1,143 +1,122 @@
...
-  // TODO: once we have pool, check if aggregate block is seen and has target as ancestor
-
-  // verifySignature = false, verified in batch above
-  if (!phase0.isValidIndexedAttestation(attestationTargetState, indexedAttestation, false)) {
-    throw new AttestationError({
-      code: AttestationErrorCode.INVALID_INDEXED_ATTESTATION,
-      job: attestationJob,
-    });
+  if (chain.seenAggregators.isKnown(targetEpoch, aggregatorIndex)) {
+    throw new AttestationError({code: AttestationErrorCode.AGGREGATOR_ALREADY_KNOWN, targetEpoch, aggregatorIndex});
```
twoeths, line 109: why is aggregator unique per target epoch?

packages/lodestar/src/chain/validation/attestation.ts
```diff
@@ -1,177 +1,286 @@
...
-    throw new AttestationError({
-      code: AttestationErrorCode.BAD_TARGET_EPOCH,
-      job: attestationJob,
-    });
+  // Now that the attestation has been fully verified, store that we have received a valid attestation from this validator.
+  //
+  // It's important to double check that the attestation still hasn't been observed, since
+  // there can be a race-condition if we receive two attestations at the same time and
+  // process them in different threads.
+  if (chain.seenAttesters.isKnown(targetEpoch, validatorIndex)) {
```
twoeths, line 144: this is a duplicate, checked above already

packages/lodestar/src/chain/validation/attestation.ts
```diff
@@ -1,177 +1,286 @@
...
-    const signatureSet = allForks.getIndexedAttestationSignatureSet(attestationTargetState, indexedAttestation);
-    if (!(await chain.bls.verifySignatureSets([signatureSet]))) {
-      throw new AttestationError({
-        code: AttestationErrorCode.INVALID_SIGNATURE,
-        job: attestationJob,
-      });
-    }
+  // [IGNORE] There has been no other valid attestation seen on an attestation subnet that has an
+  // identical attestation.data.target.epoch and participating validator index.
+  if (chain.seenAttesters.isKnown(targetEpoch, validatorIndex)) {
```
twoeths, line 124: seems like this cache strategy is a great improvement compared to our current one 👍 
confirm that each validator is a unique attester for an epoch as in `computeEpochShuffling`

packages/lodestar/src/chain/validation/aggregateAndProof.ts
twoeths: an aggregator index is just a validator index so it's unique per epoch as confirmed by `computeEpochShuffling`

packages/lodestar/src/chain/validation/aggregateAndProof.ts
```diff
@@ -1,143 +1,122 @@
...
 
-  // TODO: once we have pool, check if aggregate block is seen and has target as ancestor
-
-  // verifySignature = false, verified in batch above
-  if (!phase0.isValidIndexedAttestation(attestationTargetState, indexedAttestation, false)) {
-    throw new AttestationError({
-      code: AttestationErrorCode.INVALID_INDEXED_ATTESTATION,
-      job: attestationJob,
-    });
+  if (chain.seenAggregators.isKnown(targetEpoch, aggregatorIndex)) {
```
twoeths, line 108: this is a lightweight check, should we move it to the early stage?

packages/lodestar/src/chain/validation/attestation.ts
```diff
@@ -1,177 +1,286 @@
...
-}
+/**
+ * Verifies that the `attestation.data.target.root` is indeed the target root of the block at
+ * `attestation.data.beacon_block_root`.
+ */
+function verifyAttestationTargetRoot(headBlock: IBlockSummary, targetRoot: Root, attestationEpoch: Epoch): void {
+  // Check the attestation target root.
+  const headBlockEpoch = computeEpochAtSlot(headBlock.slot);
+
+  if (headBlockEpoch > attestationEpoch) {
```
twoeths, line 226: this does not go with the comment, please double check

twoeths: ok so headBlock here is not the head of our forkchoice, the name is misleading

packages/lodestar/src/chain/validation/aggregateAndProof.ts
twoeths: this is a duplicate, checked above

packages/beacon-state-transition/src/util/aggregationBits.ts
```diff
@@ -134,3 +135,55 @@ export function bitsToUint8Array<BitArr extends BitList | BitVector>(
   // the last chunk has 32 bytes but we don't use all of them
   return Buffer.concat(chunks).subarray(0, Math.ceil(bits.length / BITS_PER_BYTE));
 }
+
+/**
+ * Variant to extract a single bit (for un-aggregated attestations)
+ */
+export function getSingleBitIndex(bits: BitList | TreeBacked<BitList>): number {
```
twoeths, line 142: missing unit test for this

packages/lodestar/src/chain/clock/LocalClock.ts
```diff
@@ -9,6 +9,10 @@ import {ChainEvent, ChainEventEmitter} from "../emitter";
 import {IBeaconClock} from "./interface";
 import {MAXIMUM_GOSSIP_CLOCK_DISPARITY} from "../../constants";
 
+// let latest_permissible_slot = chain
```
twoeths: remove this?

packages/lodestar/src/chain/eventHandlers.ts
```diff
@@ -244,32 +236,19 @@ export async function onBlock(
...
-
-    if (this.metrics) {
-      // Register attestations in metrics
-      for (const attestation of indexedAttestations) {
-        if (attestation) {
-          this.metrics.registerAttestationInBlock(attestation, block.message);
-        }
+    for (const attestation of attestations) {
+      try {
+        const indexedAttestation = postState.epochCtx.getIndexedAttestation(attestation);
```
twoeths, line 245: we did get indexed attestation to verify signatures. I wonder if we could get it from some caches to avoid creating it again?

packages/lodestar/src/chain/validation/attestation.ts
```diff
@@ -1,177 +1,286 @@
...
-      job: attestationJob,
+      code: AttestationErrorCode.BAD_TARGET_EPOCH,
     });
   }
 
-  // No need to check if `attestation.data.beaconBlockRoot` is a badBlock
-  // If it is, the ForkChoice won't know about it so the `chain.forkChoice.hasBlock` is sufficient
+  // Ensure attestation is within the last ATTESTATION_PROPAGATION_SLOT_RANGE slots (within a MAXIMUM_GOSSIP_CLOCK_DISPARITY allowance).
+  //
+  // TODO (LH): Do not queue future attestations for later processing. Investigate why LH doesn't do it
```
twoeths: I think this is because they have `toleranceSec` check per slot range validation, that should be enough (far future slot should be invalid)

twoeths: so we can remove this 2 `TODO` concern

packages/lodestar/src/chain/validation/attestation.ts
```diff
@@ -1,177 +1,286 @@
...
-    });
+/**
+ * Verify that the `attestation` is within the acceptable gossip propagation range, with reference
+ * to the current slot of the `chain`.
+ *
+ * Accounts for `MAXIMUM_GOSSIP_CLOCK_DISPARITY`.
+ */
+export function verifyPropagationSlotRange(chain: IBeaconChain, attestationSlot: Slot): void {
+  // TODO (LH): Do not queue future attestations for later processing. Investigate why LH doesn't do it
+  // TODO: use MAXIMUM_GOSSIP_CLOCK_DISPARITY to compute the clock.currentSlot
```
twoeths: remove these 2 TODOs as we count on `MAXIMUM_GOSSIP_CLOCK_DISPARITY_SEC` already

packages/lodestar/src/chain/validation/attestation.ts
dapplion: > this is a duplicate, checked above already

The duplication is intentional, since there as async call before to the bls verification there's a race condition.

```
// It's important to double check that the attestation still hasn't been observed, since
  // there can be a race-condition if we receive two attestations at the same time and
  // process them in different threads.
```

packages/lodestar/src/chain/validation/aggregateAndProof.ts
dapplion: The duplication is intentional, since there as async call before to the bls verification there's a race condition.

I should add a comment there tho

packages/lodestar/src/chain/validation/attestation.ts
dapplion: What name would you use here?

packages/lodestar/src/chain/eventHandlers.ts
dapplion: Right! I though about this optimization too, it should be doable but I would tackle in another PR for sure.

twoeths: agree 👍

packages/lodestar/src/chain/validation/attestation.ts
twoeths: just read it again, `headBlock` means the head block the attester voted for so it's still valid, but in lodestar we mean the current head of our forkchoice
maybe just adding a comment in the method header would be enough

packages/lodestar/src/chain/eventHandlers.ts
dapplion: https://github.com/ChainSafe/lodestar/issues/2771

twoeths (dismissed):
looks good to me. Something I notice from this PR
+ we'll not have the issue in #2748
+ the simple cache of validator index/aggregator index is really an improvement to us

for gossip `AggregatedAndProof`, we need to be prepared to process ~600 messages per slot (1 slot = 64 committee indices, each committee has 8-9 aggregators), we'll see if we can handle that well with this update

dapplion:
Huge difference in our throughput of gossip objects. Ran this branch from 15:00 in the chart below (note it's logarithmic)

Screenshot from 2021-06-30 17-03-18

dapplion:
Been running overnight, still showing a massive improvement in throughput. Ran this branch after 06/30 15:00

Screenshot from 2021-07-01 12-11-11

wemeetagain (approved):

PR #2761, opened 2021-06-28
Remove relative import

wemeetagain:
Bug report from discord:
```
yorick@ethlinux:~/eth-docker-devel$ docker-compose run --rm consensus --help
internal/modules/cjs/loader.js:905
  throw err;
  ^

Error: Cannot find module '../../../../beacon-state-transition/lib'
```


dapplion (approved):
We should finish and merge https://github.com/ChainSafe/lodestar/pull/2193

PR #2769, opened 2021-06-29
Reuse variables in forkchoice

### LINKED ISSUE #2755: Improve forkchoice addLatestMessage
**Describe the bug**

This is the profile from a full attestation subnet subscriptions
Screen Shot 2021-06-28 at 14 02 09

It takes almost 14% of the time for the following

+ We access the same tree data in the loop and pass to `addLatestMessage`
+ Then `addLatestMessage` keeps calling `toHexString` of the same root

**Expected behavior**

+ Avoid accessing the same tree property in a for loop
+ May want string instead of Root for `addLatestMessage`

twoeths:
**Motivation**
This is from profiling a full attestation subnet subscriptions node: redundant call of tree backed `get` and `toHexString` causes 14% of total time

+ attestation data is a tree backed and we keep accessing same data in for loops
+ `addLatestMessage` keeps doing `toHexString` for the same root
+ some functions keep accessing same properties of tree backed objects

**Description**
+ add more variables to functions, reuse them as much as possible
+ `addLatestMessage` now accepts a RootHex to avoid redundant `toHexString` call

Closes #2755

### DIFF
--- a/packages/fork-choice/src/forkChoice/forkChoice.ts
+++ b/packages/fork-choice/src/forkChoice/forkChoice.ts
@@ -321,15 +321,16 @@ export class ForkChoice implements IForkChoice {
 
     let shouldUpdateJustified = false;
     const {currentJustifiedCheckpoint, finalizedCheckpoint} = state;
+    const stateJustifiedEpoch = currentJustifiedCheckpoint.epoch;
     // Update justified checkpoint.
-    if (currentJustifiedCheckpoint.epoch > this.fcStore.justifiedCheckpoint.epoch) {
+    if (stateJustifiedEpoch > this.fcStore.justifiedCheckpoint.epoch) {
       if (!justifiedBalances) {
         throw new ForkChoiceError({
           code: ForkChoiceErrorCode.UNABLE_TO_SET_JUSTIFIED_CHECKPOINT,
           error: new Error("No validator balances supplied"),
         });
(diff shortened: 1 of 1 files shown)


packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -747,30 +753,28 @@ export class ForkChoice implements IForkChoice {
     // is from a prior epoch to the attestation, then the target root must be equal to the root
     // of the block that is being attested to.
     const expectedTarget =
-      target.epoch > computeEpochAtSlot(block.slot)
-        ? indexedAttestation.data.beaconBlockRoot
-        : fromHexString(block.targetRoot);
+      target.epoch > computeEpochAtSlot(block.slot) ? beaconBlockRoot : fromHexString(block.targetRoot);
```
dapplion: Instead of converting `block.targetRoot` to Root, can we convert `beaconBlockRoot` to string and compare them as strings below?

dapplion (dismissed):
Looks great!

packages/fork-choice/src/forkChoice/forkChoice.ts
twoeths: will do that

wemeetagain (approved):
:+1:

PR #2770, opened 2021-06-29
Fix batch signature verification

twoeths:
**Motivation**

Currently we both validate signature separately and in batch

**Description**

Should only validate signatures in batch

### DIFF
--- a/packages/lodestar/src/chain/blocks/process.ts
+++ b/packages/lodestar/src/chain/blocks/process.ts
@@ -111,9 +111,9 @@ async function processBlocksInEpoch(
         reprocess: job.reprocess,
         prefinalized: job.prefinalized,
         signedBlock: block,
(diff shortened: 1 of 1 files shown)


dapplion (changes requested):
Are you sure? We want to be able to switch from using the batch verify sig verification

- `disableBlsBatchVerify = true && job.validSignatures = true`: don't check sigs
- `disableBlsBatchVerify = true && job.validSignatures = false`: check sigs with `runStateTransition()`
- `disableBlsBatchVerify = false && job.validSignatures = true`: don't check sigs
- `disableBlsBatchVerify = false && job.validSignatures = false`: check sigs with `bls.verifySignatureSets()`

twoeths:
seems that not an issue, not sure why I see it verified 2 times in my profile (which I did not save), closing for now

PR #2776, opened 2021-06-30
New IBeaconConfig with the cache apis

twoeths:
## Motivation
+ This starts from #2515 
+ Even through CI failed, I need your reviews before I proceed since the modification is a lot

## Description
+ The old `IBeaconConfig` becomes` IChainForkConfig`
+ New interface `ICachedGenesis` supporting a `getDomain()` api which cache domains inside, need a `genesisValidatorRoot` for this
+ The new `IBeaconConfig` is `IChainForkConfig & ICachedGenesis`

With this new `IBeaconConfig`, we can get cached domain from everywhere including lodestar & validator. A lot of flows don't need that (and we don't have `genesisValidatorRoot`) so they'd use `IChainForkConfig` instead

Once I get a green light on this, I'll close this PR and replace with an official PR.

Part of #2505

### DIFF
--- a/packages/config/src/beaconConfig.ts
+++ b/packages/config/src/beaconConfig.ts
@@ -1,18 +1,31 @@
+import {Root} from "../../types";
 import {createIChainConfig, IChainConfig} from "./chainConfig";
 import {createIForkConfig, IForkConfig} from "./forkConfig";
+import {createICachedGenesis} from "./genesisConfig";
+import {ICachedGenesis} from "./genesisConfig/types";
 
 /**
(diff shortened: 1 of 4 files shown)


dapplion:
> Once I get a green light on this, I'll close this PR and replace with an official PR. 

:green_circle: Looks good to me!

wemeetagain:
:green_heart: LGTM! Definitely good to keep exposing what _was_ the IBeaconConfig for consumers that don't have the genesisValidatorsRoot available.

twoeths:
close this in favor of #2787

PR #2782, opened 2021-07-01
Add light client init proof builder

wemeetagain:
**Motivation**

Initializing a light client requires fetching a few pieces of state, given a trusted checkpoint. Unfortunately, we usually don't have that required state lying around to be fetched.

Instead of regenerating the state to create the proofs, the proofs can be made ahead of time, as the chain is progressing. These proofs can be stored thru the ws period, and fetched via API.

**Description**

Add light client init proof builder, which creates/stores init proof to db
Add new API endpoint to fetch init proofs given a state root or epoch
Deduplicates sync committee proofs - An init proof is 69k, most of which is the committee pubkeys
Prunes proofs outside ws period

TODO:
- more tests


packages/light-client/src/client/index.ts
```diff
@@ -70,16 +69,7 @@ export class Lightclient {
...
-    // fetch a proof with everything needed to bootstrap the client
-    const paths = [
-      // initial sync committee list
-      ...getSyncCommitteesProofPaths(),
-      // required to initialize a slot clock
-      ["genesisTime"],
-      // required to verify signatures
-      ["genesisValidatorsRoot"],
-    ];
-    const proof = await api.lightclient.getStateProof(toHexString(stateRoot), paths);
```
dapplion, line 82: Should we assert that state.genesisTime and state.genesisValidatorsRoot are not zero explicitly as a sanity check?

wemeetagain: I don't think they'll ever be zero. If the proof doesn't contain the data, an error will be thrown rather than zeros being returned.

packages/lodestar/src/chain/lightClient/LightClientIniter.ts
```diff
@@ -1,12 +1,12 @@
...
+import {Proof, ProofType, TreeOffsetProof} from "@chainsafe/persistent-merkle-tree";
+import {SLOTS_PER_EPOCH} from "@chainsafe/lodestar-params";
+import {Epoch, phase0} from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
-import {CachedBeaconState} from "@chainsafe/lodestar-beacon-state-transition";
+import {computeSyncPeriodAtEpoch} from "@chainsafe/lodestar-beacon-state-transition";
 import {ForkChoice} from "@chainsafe/lodestar-fork-choice";
 import {IBeaconDb} from "../../db";
 import {StateContextCache} from "../stateCache";
+import { computeWeakSubjectivityPeriod } from "../../../../beacon-state-transition/lib/allForks/util/weakSubjectivity";
```
dapplion: Dangerous relative import!

packages/api/src/routes/lightclient.ts
```diff
@@ -39,15 +38,15 @@ export const routesData: RoutesData<Api> = {
   getBestUpdates: {url: "/eth/v1/lightclient/best_updates/", method: "GET"},
   getLatestUpdateFinalized: {url: "/eth/v1/lightclient/latest_update_finalized/", method: "GET"},
   getLatestUpdateNonFinalized: {url: "/eth/v1/lightclient/latest_update_nonfinalized/", method: "GET"},
-  getInitProof: {url: "/eth/v1/lightclient/init_proof/:stateId", method: "GET"},
+  getInitProof: {url: "/eth/v1/lightclient/init_proof/:epoch", method: "GET"},
```
dapplion, line 41: If we query by epoch, the consumer hopes that the state is the same as it got in the checkpoint?

wemeetagain: Yeah
1. start with trusted `checkpoint`
2. use `checkpoint.root` to fetch block `header`
3. verify `hashTreeRoot(header)` against `checkpoint.root`
4. use `checkpoint.epoch` to fetch init `proof`
5. verify init `proof` against `header.stateRoot`

packages/lodestar/src/chain/lightClient/LightClientIniter.ts
wemeetagain: fixed

dapplion (approved):

PR #2784, opened 2021-07-01
Updatable preset

wemeetagain:
**Motivation**

In some cases, consumers of our libraries may not be able to set environment variables, and so not be able to select alternative presets (eg: minimal, for testing purposes).
For example, `create-react-app` doesn't allow for injecting arbitrary environment variables, only environment variables that are prefixed with `REACT_APP_`.

**Description**

Refactor the active preset selection and add a `setActivePreset` function that can switch the active preset.

Note: manually setting the active preset is potentially dangerous, as downstream lodestar libraries expect preset values to be _constant_.
Some care was given to document the situation in the `setActivePreset` docstring.

### DIFF
--- a/packages/params/src/activePreset.ts
+++ b/packages/params/src/activePreset.ts
@@ -1,27 +1,32 @@
 // load in an "active" preset
 
-import {IBeaconPreset, PresetName} from "./preset";
+import {PresetName} from "./preset";
 
 import {preset as mainnet} from "./presets/mainnet";
 import {preset as minimal} from "./presets/minimal";
 
-let ACTIVE_PRESET: PresetName;
-let preset: IBeaconPreset;
+const presets = {
+  [PresetName.mainnet]: mainnet,
+  [PresetName.minimal]: minimal,
+};
 
-switch (process.env.LODESTAR_PRESET) {
-  case PresetName.minimal:
-    ACTIVE_PRESET = PresetName.minimal;
-    preset = minimal;
-    break;
-  case PresetName.mainnet:
-  default:
-    ACTIVE_PRESET = PresetName.mainnet;
-    preset = mainnet;
-    break;
+function parsePresetName(name: string): PresetName {
+  return PresetName[name as PresetName] ?? PresetName.mainnet;
 }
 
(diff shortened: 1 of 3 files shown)


packages/params/src/activePreset.ts
```diff
@@ -60,4 +65,57 @@ export const {
...
+ *
+ * Only call this function if you _really_ know what you are doing.
+ */
+export function setActivePreset(presetName: PresetName): void {
+  ACTIVE_PRESET = parsePresetName(presetName);
+  const preset = presets[ACTIVE_PRESET];
+
+  // Unfortunately, there is way to dynamically update all of these variables
+  // so they must be updated individually and explicitly
+  MAX_COMMITTEES_PER_SLOT = preset.MAX_COMMITTEES_PER_SLOT;
```
3xtr4t3rr3str14l, line 83: Would doing something like 
```typescript
Object.entries(preset).forEach((k, v) => globalThis[k] = v
```

or some variant of that be dangerous terrible idea? :laughing:

packages/params/src/activePreset.ts
```diff
@@ -60,4 +65,57 @@ export const {
...
+ *
+ * WARNING: Lodestar libraries rely on preset values being _constant_, so the active preset must be set _before_ loading any other lodestar libraries.
+ *
+ * Only call this function if you _really_ know what you are doing.
+ */
+export function setActivePreset(presetName: PresetName): void {
+  ACTIVE_PRESET = parsePresetName(presetName);
+  const preset = presets[ACTIVE_PRESET];
+
+  // Unfortunately, there is way to dynamically update all of these variables
```
3xtr4t3rr3str14l: i'm assuming you mean this?

```suggestion
  // Unfortunately, there is no way to dynamically update all of these variables
```

packages/params/src/activePreset.ts
wemeetagain: There's no way to access the module scope (as far as I know). `globalThis` isn't the same thing.

3xtr4t3rr3str14l (approved):

PR #2787, opened 2021-07-02
Cache domains in config

### LINKED ISSUE #2505: Cache forkDataRoot

twoeths:
**Motivation**

We keep computing domain from same values, on a node with full attestation subnets subscription, it takes 2% - 3% for this redundancy.

**Description**

+ Make the new `IBeaconConfig` being able to cache domains, we need `genesisValidatorsRoot` for this
+ The old `IBeaconConfig` becomes `IChainForkConfig`, use this for contexts without `genesisValidatorsRoot`
+ Remove `getDomain` function
+ Clean up unused config everywhere

The main change was reviewed in #2776, this PR applies `IChainForkConfig` or `IBeaconConfig` to different places
Closes #2505


dapplion (approved):
Read everything, looks great! Looking forward to this improvement :fire: 


I've pushed two commits:
- Fixed conflicts
- Convert to calls to get fork info to a single one, here https://github.com/ChainSafe/lodestar/pull/2787/files#diff-f44ca031217bdd7aa85c293fb5fecb35cded012172694ebdeb1262a4d824ce3bR10

wemeetagain (approved):

PR #2798, opened 2021-07-05
Fix SyncCommitteeContribution and SyncCommitteeMessage

### LINKED ISSUE #2797: Yeerongpilly: Handle SYNC_COMMITTEE_CONTRIBUTION_ERROR_ALREADY_KNOWN
**Describe the bug**

This error happens a lot in our log:
```
(node:1) UnhandledPromiseRejectionWarning: Error: SYNC_COMMITTEE_CONTRIBUTION_ERROR_ALREADY_KNOWN
    at aggregateContributionInto (/usr/app/packages/lodestar/src/db/syncCommitteeContribution.ts:146:15)
    at SyncCommitteeContributionCache.add (/usr/app/packages/lodestar/src/db/syncCommitteeContribution.ts:63:7)
    at Eth2Gossipsub.<anonymous> (/usr/app/packages/lodestar/src/network/gossip/handler.ts:116:39)
```

**Expected behavior**

That error happens in gossip handler, meaning it passed gossip validation function.
Investigate why gossip validation function does not detect these messages as `ALREADY_KNOWN`

twoeths:
**Motivation**
This is all about gossip for altair, detected with yeerongpilly

1. This error happens all the time
```
(node:3783) UnhandledPromiseRejectionWarning: Error: SYNC_COMMITTEE_CONTRIBUTION_ERROR_ALREADY_KNOWN
    at aggregateContributionInto (/root/workspace/node/lodestar/packages/lodestar/src/db/syncCommitteeContribution.ts:146:15)
```
it's not really an error, a sub committee has multiple aggregators, hence multiple contributions.

2. We miss adding altair validator functions to topic string, hence SyncCommittee messages never go through our validator function and we miss its indexed data

```
(node:1) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 34732)
(node:1) UnhandledPromiseRejectionWarning: Error: Already aggregated SyncCommitteeMessage - subCommitteeIndex=3 indexInSubCommittee=undefined
    at aggregateSignatureInto (/usr/app/packages/lodestar/src/db/syncCommittee.ts:144:11)
    at SyncCommitteeCache.add (/usr/app/packages/lodestar/src/db/syncCommittee.ts:79:7)
    at GossipHandler.onSyncCommitteeSignature (/usr/app/packages/lodestar/src/network/gossip/handler.ts:131:27)

```

**Description**

1. Instead of throwing error, just ignore the latter contribution.
2. Add missing altair validator functions to topic string map

Closes #2797

Sync `yeerongpilly`

### DIFF
--- a/packages/lodestar/src/db/syncCommitteeContribution.ts
+++ b/packages/lodestar/src/db/syncCommitteeContribution.ts
@@ -143,11 +143,9 @@ function aggregateContributionInto(
     if (participated) {
       const syncCommitteeIndex = indexOffset + index;
       if (aggregate.syncCommitteeBits[syncCommitteeIndex] === true) {
-        throw new SyncContributionError({
-          code: SyncContributionErrorCode.ALREADY_KNOWN,
-          syncCommitteeIndex,
-          slot: contribution.slot,
-        });
+        // there are multiple SyncCommitteeContribution messaages of the same sub committee
+        // this means we aggregated a contribution of same sub committee already
+        return;
       }
 
       aggregate.syncCommitteeBits[syncCommitteeIndex] = true;
--- a/packages/lodestar/src/network/gossip/validator.ts
+++ b/packages/lodestar/src/network/gossip/validator.ts
@@ -1,5 +1,5 @@
 import {AbortSignal} from "@chainsafe/abort-controller";
-import {ATTESTATION_SUBNET_COUNT, ForkName} from "@chainsafe/lodestar-params";
+import {ATTESTATION_SUBNET_COUNT, ForkName, SYNC_COMMITTEE_SUBNET_COUNT} from "@chainsafe/lodestar-params";
 import {mapValues} from "@chainsafe/lodestar-utils";
 import {IMetrics} from "../../metrics";
 import {JobQueue, JobQueueOpts, QueueType} from "../../util/queue";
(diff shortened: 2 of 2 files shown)


packages/lodestar/src/db/syncCommitteeContribution.ts
```diff
@@ -143,11 +143,9 @@ function aggregateContributionInto(
...
       const syncCommitteeIndex = indexOffset + index;
       if (aggregate.syncCommitteeBits[syncCommitteeIndex] === true) {
-        throw new SyncContributionError({
-          code: SyncContributionErrorCode.ALREADY_KNOWN,
-          syncCommitteeIndex,
-          slot: contribution.slot,
-        });
+        // there are multiple SyncCommitteeContribution messaages of the same sub committee
+        // this means we aggregated a contribution of same sub committee already
+        return;
```
dapplion, line 148: I don't think we should return here. The consumer should be informed that we are not including this aggregate. The problem is that gossip handlers are not handled and they just reject unhandled.

I would add a try / catch here and return for that typed error like we do in the API
https://github.com/ChainSafe/lodestar/blob/3a63d7f8948234424335b8f4fae67e974259121d/packages/lodestar/src/network/gossip/handler.ts#L115-L117

twoeths: if there are 10 contributions for the same subnet we'd ignore 9 of them and this is expected, why do you think we have to throw error here? I see that you refactor a lot in #2803 anyway so this is not a big deal.

actually we're not strictly following the spec `Of these contributions, proposers should select the best contribution seen across all aggregators for each subnet/subcommittee. A contribution with more valid signatures is better than a contribution with fewer signatures.`. Is this possible to do that in #2803 ?

twoeths:
close in favor of #2816

PR #2801, opened 2021-07-05
Add generic batch BLS verification

### LINKED ISSUE #2686: Batch BLS verification for attestations

Very promising optimization, it has been very successful in Nimbus and Lighthouse. See results here: https://github.com/sigp/lighthouse/pull/2399. We struggle to process all attestations and aggregates in Prater, this will help a lot.

https://github.com/sigp/lighthouse/pull/2399

dapplion:
**Motivation**

Very promising optimization, it has been very successful in Nimbus and Lighthouse. See https://github.com/ChainSafe/lodestar/issues/2686

**Description**

- Buffer submitted work for 100ms
- Then construct packets of work that will be verified in batch, if some fail re-verify all individually. This logic happens in the worker side to reduce communication.
- Only allow this behaviour on certain work packets. The BLS queue does not apply this strategy to work with a lot a signature sets, only to those with fewer than 3-5 (aggregates + attesations) that manually requested.

Closes https://github.com/ChainSafe/lodestar/issues/2686

**Testing**

The parameters need tweaking and more debugging. I've tested this PR over the weekend and didn't got good results.
- **EDIT: fixed** - 3% of all signatures are invalid according to metrics which makes the batching strategy less efficient
- **EDIT: fixed** - As so, batches are retried very often
- There are some memory issues that cause far OOM. Still need more debugging

### DIFF
--- a/docker/grafana/provisioning/dashboards/lodestar.json
+++ b/docker/grafana/provisioning/dashboards/lodestar.json
@@ -3456,7 +3456,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "delta(lodestar_bls_thread_pool_batch_retries[$__rate_interval])",
+              "expr": "rate(lodestar_bls_thread_pool_batch_retries[$__rate_interval])",
               "interval": "",
               "legendFormat": "pool",
               "refId": "A"
--- a/docker/grafana/provisioning/dashboards/lodestar.json
+++ b/docker/grafana/provisioning/dashboards/lodestar.json
@@ -2867,11 +2867,50 @@
       "id": 92,
       "panels": [
         {
+          "datasource": null,
+          "fieldConfig": {
+            "defaults": {
+              "color": {},
+              "custom": {},
+              "thresholds": {
+                "mode": "absolute",
+                "steps": []
+              }
+            },
+            "overrides": []
+          },
+          "gridPos": {
+            "h": 3,
+            "w": 24,
+            "x": 0,
+            "y": 21
+          },
+          "id": 154,
+          "options": {
+            "content": "Verifies signature sets in a thread pool of workers. Must ensure that signatures are verified fast and efficiently.",
+            "mode": "markdown"
+          },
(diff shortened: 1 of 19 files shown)


dapplion:
After debugging `prom-client` it will increase by 1 if you increase by 0.

https://github.com/siimon/prom-client/blob/c7b9a9ddd048f6a1921fa1b33c56a5296713aa2b/lib/gauge.js#L156

@tuyennhv @g11tech @dapplion  Make sure to not call `inc(value)` where value may be exactly 0. Instead do
```ts
if (value > 0) inc(value)
```

dapplion:
This branch still has occasional fast OOM issues. Debugging,

Screenshot from 2021-07-05 22-14-25

dapplion:
I've been running this branch over the weekend, subscribed to all subnets without any memory issues

Screenshot from 2021-07-12 21-41-27

dapplion:
The unstable memory pattern is also happening in another machine running master on Prater with 16 validators attached

Screenshot from 2021-07-13 21-33-07

dapplion:
While I'm still not able to measure a significant drop in CPU, clearly it's verifying signatures at a cheaper cost. This chart shows that the average verification time per sig went from 3.5 ms / sig (with master) to 2.0 ms / sig (with this branch).

Screenshot from 2021-07-23 02-11-09

dapplion:
This branch has been running for 3.5 days now and the results are good. Memory is stable, signatures are processed faster and the worker threads are less bussy. The throughput did not increase dramatically indicating that the bottleneck is still the main thread, that busy doing other stuff.

IMO this is ready to be merged

*On all charts below this branch was ran in the middle*

Screenshot from 2021-07-26 22-03-24
Screenshot from 2021-07-26 22-03-34
Screenshot from 2021-07-26 22-04-14
Screenshot from 2021-07-26 22-03-55

Screenshot from 2021-07-26 22-08-52
Screenshot from 2021-07-26 22-04-28

wemeetagain (approved):
:rocket:

PR #2802, opened 2021-07-05
Add DB metrics charts

dapplion:
**Motivation**

Add charts for https://github.com/ChainSafe/lodestar/pull/2741

**Description**

Last row in dashboard. Charts below are from Prater

Screenshot from 2021-07-05 18-29-59
Screenshot from 2021-07-05 18-30-08

### DIFF
--- a/docker/grafana/provisioning/dashboards/lodestar.json
+++ b/docker/grafana/provisioning/dashboards/lodestar.json
@@ -15,7 +15,7 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
-  "id": 1,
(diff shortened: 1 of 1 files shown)


wemeetagain (approved):

dapplion:
Zoom in chart, clearly we are doing a lot of reads at epoch boundaries
Screenshot from 2021-07-05 22-27-50

Below with delta() function it shows the absolute value difference between scrapes. We read 2000-4000 blocks. Are the metrics right? Is our archiving function inefficient?
Screenshot from 2021-07-05 22-28-27

PR #2803, opened 2021-07-05
Review gossipsub handlers

dapplion:
**Motivation**

Our gossipsub validator + handlers use an inconvenient architecture that prevents us from re-using state within them. It also forces us to mutate objects to attach data to them to prevent repeating work.

**Description**

Now, gossip handlers perform validation + handling in a single function.
 - This gossip handlers MUST only be registered as validator functions. No handler is registered for any topic.
 - All `chain/validation/*` functions MUST throw typed GossipActionError instances so they gossip action is captured
   by `getGossipValidatorFn()` try catch block.
 - This gossip handlers should not let any handling errors propagate to the caller. Only validation errors must be thrown.

Note: `libp2p/js-libp2p-interfaces` would normally indicate to register separate validator functions and handler functions.
This approach is not suitable for us because:
 - We do expensive processing on the object in the validator function that we need to re-use in the handler function.
 - The validator function produces extra data that is needed for the handler function. Making this data available in the handler function scope is hard to achieve without very hacky strategies
 - Eth2.0 gossipsub protocol strictly defined a single topic for message

----

Adds three caches to attach data to objects without mutating them:
- gossipTopicCache: Map, store gossip for each known topicStr
- uncompressCache: WeakMap, caches uncompressedData for each gossip message.data
- msgIdCache: WeakMap, caches msgID for each gossip object


wemeetagain (dismissed):
LGTM

packages/lodestar/src/network/gossip/validation/validatorFnsByTopic.ts
```diff
@@ -0,0 +1,70 @@
...
+export function createValidatorFnsByTopic(
+  config: IBeaconConfig,
+  forkDigestContext: IForkDigestContext,
+  validatorFnsByType: ValidatorFnsByType
+): Map<string, GossipValidatorFn> {
+  const validatorFnsByTopic = new Map<string, GossipValidatorFn>();
+
+  const encoding = DEFAULT_ENCODING;
+  const allForkNames = Object.values(config.forks).map((fork) => fork.name);
+  const allForksAfterAltair = allForkNames.filter((fork) => fork !== ForkName.phase0);
```
wemeetagain: allForksAfterPhase0 ?

twoeths (approved):

twoeths:
Tested with a `yeerongpilly` node with 400 validators, everything is fine except for the `SYNC_COMMITTEE_CONTRIBUTION_ERROR_ALREADY_KNOWN` error as being discussed in #2798

PR #2808, opened 2021-07-06
Fix relative import

twoeths:
**Motivation**

The relative imports made the docker build broken
```
Error: Cannot find module '../../../types'
```


dapplion (approved):

dapplion:
I've been researching eslint rules to help us there but haven't found anything promising. We may have to roll our own.

PR #2810, opened 2021-07-06
Write heapdump to disk on request through API

dapplion:
**Motivation**

Since NodeJS v12 you can take a heapdump without requiring the external library `heapdump`. It's convenient to manually trigger heapdumps remotely when watching the metrics.

**Description**

When querying a route, write a heapdump to the current path or `dirpath`. All modules are imported lazily.

```
curl http://localhost:9596/eth/v1/lodestar/writeheapdump/?dirpath=/headdumps
```


dapplion:
I have issues with this strategy, NodeJS doesn't actually write the file to disk. Needs more testing

twoeths (approved):
lgtm, note that this does work with Node 16 but not with Node 14

For Node 14, node crashes with `segmentation fault` error and it happens with `heapdump` or Chrome Dev Tools as well.

wemeetagain (approved):
I tested locally w node v14, it worked, it created the heapsnapshot and the request resolved, but later when exiting the node, instead of exiting nicely, it crashed with OOM

PR #2816, opened 2021-07-06
Move sync pools to chain

### LINKED ISSUE #2797: Yeerongpilly: Handle SYNC_COMMITTEE_CONTRIBUTION_ERROR_ALREADY_KNOWN

dapplion:
**Motivation**

Extends https://github.com/ChainSafe/lodestar/pull/2798 consolidating the op pools clases together in the chain submodule.

**Description**

Previously we had some "repositories" in the db that contained strictly in-memory data, never persisted to disk. We should keep in-memory data in closes closer to where it makes sense and only db read - write logic in the db submodule.

Also, the SyncCommittee and SyncContribution classes where being two things at once:
- Op pool
- Gossip 'seen' cache

This PR splits them and re-orgs the logic into
- `chain/opPools` for classes handling op pool objects. We may want to persist them latter to db, on shutdown for example.
- `chain/seenCache` for classes handling the cache to support gossip validation

Closes #2797


wemeetagain (dismissed):

wemeetagain (approved):
LGTM :+1:

PR #2831, opened 2021-07-11
Improve rotateEpochs and exitQueueEndChurn

### LINKED ISSUE #2830: Save one validator loop in rotateEpoch

In Prater, a validator loop in `rotateEpoch` may take up to 700 ms:

Screen Shot 2021-07-11 at 14 25 00

`rotateEpoch` always run after we process an epoch, which we can extract some validator data from epoch process

twoeths:
**Motivation**

+ The validator for loop inside `rotateEpochs` may take up to 700ms in Prater

**Description**
+ Save 1 validator loop in `rotateEpochs` by preparing necessary validator data in `prepareEpochProcessState`
+ Calculate exitQueueEndChurn in the main for loop of `prepareEpochProcessState`

Closes #2830

Sync prater

### DIFF
--- a/packages/beacon-state-transition/src/allForks/util/epochProcess.ts
+++ b/packages/beacon-state-transition/src/allForks/util/epochProcess.ts
@@ -121,8 +121,13 @@ export function prepareEpochProcessState<T extends allForks.BeaconState>(state: 
       activeCount += 1;
     }
 
-    if (v.exitEpoch !== FAR_FUTURE_EPOCH && v.exitEpoch > exitQueueEnd) {
-      exitQueueEnd = v.exitEpoch;
+    if (v.exitEpoch !== FAR_FUTURE_EPOCH) {
+      if (v.exitEpoch > exitQueueEnd) {
+        exitQueueEnd = v.exitEpoch;
+        exitQueueEndChurn = 1;
+      } else if (v.exitEpoch === exitQueueEnd) {
+        exitQueueEndChurn += 1;
+      }
     }
 
     if (v.activationEligibilityEpoch === FAR_FUTURE_EPOCH && v.effectiveBalance === MAX_EFFECTIVE_BALANCE) {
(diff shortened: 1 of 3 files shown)


packages/beacon-state-transition/src/allForks/util/epochProcess.ts
```diff
@@ -118,8 +121,13 @@ export function prepareEpochProcessState<T extends allForks.BeaconState>(state:
       activeCount += 1;
     }
 
-    if (v.exitEpoch !== FAR_FUTURE_EPOCH && v.exitEpoch > exitQueueEnd) {
-      exitQueueEnd = v.exitEpoch;
+    if (v.exitEpoch !== FAR_FUTURE_EPOCH) {
+      if (v.exitEpoch > exitQueueEnd) {
+        exitQueueEnd = v.exitEpoch;
+        exitQueueEndChurn = 1;
```
wemeetagain, line 128: Is this right?
exitQueueEndChurn will be reset to 1 in this case?

twoeths: yeah it should not be zero. `exitQueueEndChurn` means number of `max(exitEpoch)` so if we find a new higher `exitEpoch`, we found the 1st validator with that highest `exitEpoch`

tbh I made it start from 0 and then spec tests failed, so I fixed it this way :D

wemeetagain: Oh, I see

Before:
- Loop thru validators, find the max exit epoch that isn't `FAR_FUTURE_EPOCH`, set that to `exitQueueEnd`.
- Loop thru validators again, count all validators with exit epoch equal to `exitQueueEnd`, set that to `exitQueueEndChurn`

Now:
- Loop once thru validators, start counting validators with the largest exit epoch (that isn't `FAR_FUTURE_EPOCH`), reset the count and largest exit epoch if a larger exit epoch is found

wemeetagain (approved):

PR #2833, opened 2021-07-12
Avoid recalculating rewards penalties for same effective balance

### LINKED ISSUE #2832: Avoid recalculating rewards and penalties

This shows that effectiveBalance is always multiples of 1Eth and should not be more than 32Eth: https://github.com/ChainSafe/lodestar/blob/b49ac4c084c21be52cd9b98e2151c18851b5af73/packages/beacon-state-transition/src/allForks/epoch/processEffectiveBalanceUpdates.ts#L26

a confirmation of effective balance in mainnet: the value set is like `@@@ effectiveBalances of all validators Set(3) { 32000000000, 31000000000, 30000000000 }`

We should be able to avoid recalculating rewards and penalties by using a cache to help:
+ improve state transition speed
+ save memory due to BigInt calculation

current situation in Prater shows that it may take more than 1s to process rewards and penalties

Screen Shot 2021-07-12 at 08 38 55

twoeths:
**Motivation**

+ Effective balance is always a multiple of EFFECTIVE_BALANCE_INCREMENT (1ETH) and not greater than MAX_EFFECTIVE_BALANCE (32ETH) so we don't want to calculate rewards and penalties deltas from the same data
+ For altair, we calculate source/target/head/penalty rewards/penalties deltas separately, it requires multiple loops over `process.statuses` and multiple BigInt arrays

**Description**

+ Use a cache (for both phase0 and altair)
+ Altair: loop though `process.statuses` only 1 time, this combines with the cache gives an (>10x) improvement as in the benchmark diff

Closes #2832

### DIFF
--- a/packages/beacon-state-transition/src/altair/epoch/balance.ts
+++ b/packages/beacon-state-transition/src/altair/epoch/balance.ts
@@ -20,7 +20,102 @@ import {
 } from "../../allForks/util";
 import {isInInactivityLeak, newZeroedBigIntArray} from "../../util";
 
+interface IRewardPenaltyItem {
+  baseReward: bigint;
+  timelySourceReward: bigint;
+  timelySourcePenalty: bigint;
+  timelyTargetReward: bigint;
+  timelyTargetPenalty: bigint;
+  timelyHeadReward: bigint;
+}
+
 /**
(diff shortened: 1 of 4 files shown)


dapplion:
Damn! Nice improvement on processRewardsAndPenalties!

wemeetagain (approved):

PR #2834, opened 2021-07-13
Add subscribeAllSubnets flag

dapplion:
**Motivation**

Useful for testing and to capture the entire network

**Description**

Subscribes to all subnets for attestations and sync committees

### DIFF
--- a/packages/lodestar/src/network/subnets/attnetsService.ts
+++ b/packages/lodestar/src/network/subnets/attnetsService.ts
@@ -15,7 +15,7 @@ import {Eth2Gossipsub, GossipType} from "../gossip";
 import {MetadataController} from "../metadata";
 import {SubnetMap, RequestedSubnet} from "../peers/utils";
 import {getActiveForks} from "../forks";
-import {IAttnetsService, CommitteeSubscription} from "./interface";
+import {IAttnetsService, CommitteeSubscription, SubnetsServiceOpts} from "./interface";
 
 /**
  * The time (in slots) before a last seen validator is considered absent and we unsubscribe from the random
@@ -52,7 +52,8 @@ export class AttnetsService implements IAttnetsService {
     private readonly chain: IBeaconChain,
     private readonly gossip: Eth2Gossipsub,
     private readonly metadata: MetadataController,
-    private readonly logger: ILogger
+    private readonly logger: ILogger,
+    private readonly opts?: SubnetsServiceOpts
   ) {}
 
   start(): void {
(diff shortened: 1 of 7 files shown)


packages/lodestar/src/network/subnets/attnetsService.ts
```diff
@@ -271,14 +274,18 @@ export class AttnetsService implements IAttnetsService {
...
     for (const subnet of subnets) {
       if (
         !this.subscriptionsCommittee.isActiveAtSlot(subnet, slot) &&
         !this.subscriptionsRandom.isActiveAtSlot(subnet, slot)
       ) {
         for (const fork of forks) {
-          this.gossip.unsubscribeTopic({type: gossipType, fork, subnet});
+          if (!this.opts?.subscribeAllSubnets) {
+            this.gossip.unsubscribeTopic({type: gossipType, fork, subnet});
+          }
```
twoeths: I suppose we wanna change `shouldProcess` to return true if we have this flag

dapplion: Lighthouse has another flag for it, `--import-all-subnets` since subscribing is a different action that importing the attestation.

However, I noticed that I removed the shouldProcess() call when refactoring the gossip handlers. So I will fix both in another PR

wemeetagain: Is this check needed if L277 is there?

packages/lodestar/src/network/subnets/syncnetsService.ts
```diff
@@ -82,7 +83,9 @@ export class SyncnetsService implements ISubnetsService {
     if (prevFork === ForkName.phase0) return;
     this.logger.info("Unsuscribing to random attnets from prev fork", {prevFork});
     for (let subnet = 0; subnet < SYNC_COMMITTEE_SUBNET_COUNT; subnet++) {
-      this.gossip.unsubscribeTopic({type: gossipType, fork: prevFork, subnet});
+      if (!this.opts?.subscribeAllSubnets) {
```
wemeetagain, line 86: Maybe we should have a short circuit at the top of this and `unsubscribeSubnets`
`if (this.opts?.subscribeAllSubnets) return;`

wemeetagain (changes requested):

packages/lodestar/src/network/subnets/syncnetsService.ts
dapplion: I think this syntax is more obvious, since it's immediately closer to the unsubscribeTopic() topic call.

packages/lodestar/src/network/subnets/attnetsService.ts
dapplion: Removed line 277, I think it's more obvious this way

wemeetagain (approved):

PR #2835, opened 2021-07-13
Add fromMnemonic option

dapplion:
**Motivation**

Running validators in testnets from mnemonics is very time consuming. Keystores have to be generated, encrypted and then decrypted.

**Description**

Add a hidden flag to run keys straight from a mnemonic. This option is labeled as "UNSAFE", hidden and disabled for mainnet.


3xtr4t3rr3str14l (approved):
lgtm

twoeths:
@dapplion should we put it in README/doc?

dapplion:
> @dapplion should we put it in README/doc?

Which README? This is a very dangerous option that should only be used by developers and testers

PR #2836, opened 2021-07-14
Cache known block root and accumulate proposer reward

### LINKED ISSUE #2528: Experiment caching block root in CachedBeaconState

A simple `const blockRootsAtSlot = new Map<phase0.Slot, phase0.Root>();` at https://github.com/ChainSafe/lodestar/blob/e04410b5704f04738c03a75673efff6b829f1700/packages/beacon-state-transition/src/phase0/epoch/processPendingAttestations.ts#L29

helps save ~0.5s per epoch transition in the performance test (~2.1s in total).

Cache blockRoots in CachedBeaconState similar to `validators, balances` so that we don't have to go to persistent-merkle-tree to get data, this should help speed up the whole state transition instead of just `processPendingAttestations` in phase0.

twoeths:
**Motivation**

+ For both altair and phase0, we keep getting block root for known slot since most attestations of same block point to same value
+ For altair, there are proposal rewards at different places. Note that updating a balance to `persistent-merkle-tree` is not cheap. 

**Description**

+ Cache known block roots for phase0 `processPendingAttesatations` and altair `BlockProcess`
+ Altair: Accumulate proposal reward and only update it once per block
+ Add missing `asTransient()` and `asPersistent()` calls

Closes #2528

### DIFF
--- a/packages/beacon-state-transition/src/altair/block/processAttestation.ts
+++ b/packages/beacon-state-transition/src/altair/block/processAttestation.ts
@@ -86,7 +86,19 @@ export function processAttestation(
   const totalIncrements = totalBalancesWithWeight / EFFECTIVE_BALANCE_INCREMENT;
   const proposerRewardNumerator = totalIncrements * state.baseRewardPerIncrement;
   const proposerReward = proposerRewardNumerator / PROPOSER_REWARD_DOMINATOR;
-  increaseBalance(state, epochCtx.getBeaconProposer(state.slot), proposerReward);
+  const proposerIndex = epochCtx.getBeaconProposer(state.slot);
+  // we increase proposer reward multiple times per block process, it's better to do that in batch
+  if (blockProcess.increaseBalanceCache) {
+    let increaseBalanceValue = blockProcess.increaseBalanceCache.get(proposerIndex);
+    if (increaseBalanceValue === undefined) {
+      increaseBalanceValue = BigInt(0);
+    }
+    increaseBalanceValue += proposerReward;
+    blockProcess.increaseBalanceCache.set(proposerIndex, increaseBalanceValue);
+  } else {
+    // for spec test only
+    increaseBalance(state, proposerIndex, proposerReward);
+  }
 }
 
 /**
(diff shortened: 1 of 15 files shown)


packages/beacon-state-transition/src/altair/block/processAttestation.ts
```diff
@@ -86,7 +86,19 @@ export function processAttestation(
...
-  increaseBalance(state, epochCtx.getBeaconProposer(state.slot), proposerReward);
+  const proposerIndex = epochCtx.getBeaconProposer(state.slot);
+  // we increase proposer reward multiple times per block process, it's better to do that in batch
+  if (blockProcess.increaseBalanceCache) {
+    let increaseBalanceValue = blockProcess.increaseBalanceCache.get(proposerIndex);
+    if (increaseBalanceValue === undefined) {
+      increaseBalanceValue = BigInt(0);
+    }
+    increaseBalanceValue += proposerReward;
+    blockProcess.increaseBalanceCache.set(proposerIndex, increaseBalanceValue);
```
dapplion, line 97: Maybe it's worth it to create a util BalanceMap with a single method
```ts
add(validatorIndex, balanceDelta): void
```
So the code here can be simplified

packages/beacon-state-transition/src/altair/block/processAttestation.ts
```diff
@@ -86,7 +86,19 @@ export function processAttestation(
...
+  } else {
+    // for spec test only
+    increaseBalance(state, proposerIndex, proposerReward);
```
dapplion, line 100: It doesn't work in spec tests? We should only have one path and test the same with spec tests and production execution

packages/beacon-state-transition/src/altair/block/processAttestation.ts
```diff
@@ -95,7 +107,8 @@ export function processAttestation(
 export function getAttestationParticipationStatus(
   state: CachedBeaconState<altair.BeaconState>,
   data: phase0.AttestationData,
-  inclusionDelay: number
+  inclusionDelay: number,
+  blockProcess: BlockProcess = getEmptyBlockProcess()
```
dapplion, line 111: Why is this argument optional?

dapplion:
This optimization is already implemented :+1:

PR #2838, opened 2021-07-14
validator config file for picking account paths and graffiti

### LINKED ISSUE #2814: Add --configFile to validator
From discord (@stefa2k):

Q: Hi, I'm getting keystoresDir /opt/app/.local/share/lodestar/mainnet/keystores does not exist but I set keystoresDir: /opt/app/keystores in the config yaml via --paramsFile /opt/app/validator.yaml
Am I missing something?

A: it looks the paramsFile only supports the testnet/mainnet params (the ones from the eth2 spec)
our beacon node lets you override all of the cli options via a file --configFile but it doesn't look like the validator supports that
We should add the same thing to the validator, but for now, you'd have to add --keystoresDir via cli

Allow for validator cli options to be specified in a file, like is possible for the beacon node

g11tech:
**Motivation**
This draft PR refactors and enables ValidatorOptions to be read and written to a config file. (will rebase it before final PR for merge compatibility). the config can be provided in the json format (default validator.config.json)
{
grafitti : ... ,
account: {secretsDir:... ,keystoresDir: ...}
}

@dapplion  @wemeetagain  Apart from general feedback on PR, feedback  required on:

1. in the dev handler, both beacon options and validator options are being made from the cli args, so should the  config file param name (--configFile) needs to be different for them?
2. since getAccountPaths is now modified to have an optional opts param build from config file, does that needs to be percolated to the account handlers which are being used for generating/managing account keys? 

**Description**

resultant behaviour:
image
1. when no option of keystoreDir is provided, it looks in default dir.
2. now the keystoreDir is provided, so it overrides the opts being build from the file.
3.  in 3.1 the option is again omitted and it should now be picked from the config file which was written in the last run. 3.2 confirms the path that is now being searched was provided (and written to config) in last run, and 3.3 confirms the same.

Closes #2814


dapplion:
@g11tech I think we can simplify the entire design by just using `yargs.config` parser. That's what mocha uses so they don't have to define another custom file options parser.

https://yargs.js.org/docs/#api-reference-configobject

https://github.com/mochajs/mocha/blob/master/lib/cli/config.js

g11tech:
Yes it does sound simpler just to read/dump all/any cli args to a config. will investigate mocha config code. 👍

wemeetagain:
@dapplion do you remember the reason why we created a `BeaconNodeOptions` class?
I seem to remember some sort of limitation / annoyance around using `yargs.config`, which is why its not used in the `beacon` command.

dapplion:
> @dapplion do you remember the reason why we created a `BeaconNodeOptions` class?
> I seem to remember some sort of limitation / annoyance around using `yargs.config`, which is why its not used in the `beacon` command.

Because we have a strange flow where options go through the init() function, mutated persisted without defaults and then consumed with defaults. The class is ugly but the previous approach was chaos.

In this case there is no init function, so please @g11tech can you find an approach that is simpler, with just regular objects that does not require a class nor mutation?

packages/cli/src/cmds/validator/handler.ts
```diff
@@ -4,22 +4,30 @@ import {Validator, SlashingProtection} from "@chainsafe/lodestar-validator";
...
+    validatorOptionsCli: parseValidatorArgs(args),
+  });
+  validatorOptions.writeTo(validatorPaths.configFile);
 
-  const graffiti = args.graffiti || getDefaultGraffiti();
+  const opts = validatorOptions.getWithDefaults();
+  const graffiti = opts.graffiti;
 
-  const validatorPaths = getValidatorPaths(args);
+  //Object.assign(args,{configFile:undefined}); //configFile if provided was for validator
```
dapplion, line 30: Remove commented code if not used

packages/utils/package.json
```diff
@@ -45,7 +45,8 @@
     "js-yaml": "^3.13.1",
     "winston": "^3.3.3",
     "winston-daily-rotate-file": "^4.5.5",
-    "winston-transport": "^4.3.0"
+    "winston-transport": "^4.3.0",
+    "find-up": "^5.0.0"
```
dapplion, line 49: This module relies on being run on NodeJS but lodestar-utils is used in the browser. Why should this utils be here and not in the CLI only?

packages/validator/src/options.ts
```diff
@@ -0,0 +1,16 @@
...
+  secretsDir?: string;
+}
+
+export interface IValidatorOptions {
+  graffiti: string;
+  account: IAccountPathOptions;
+}
+
+export const defaultOptions: IValidatorOptions = {
+  graffiti: getDefaultGraffiti(),
```
dapplion, line 14: lodestar-validator runs on the browser too. It must not do anything that interacts with the filesystem. Move this logic back to the CLI

g11tech:
thanks @dapplion for the feedback. will try to figure out a simpler approach.

g11tech:
@dapplion @wemeetagain  i studied the mocha config, its config loading code is massaging node based args, merging package config as well as normal config as well as setting default values itself in the config object. 

in light of above discussions, I propose that 
1) we can have a fixed cli args config file .lodestarrc.json/yml/js config irrespective of which command is being run to supplement cli args. it will act like env file and right .lodestarrc file would need to be copied/created. we need to have a common file (like global options) across all cmds as the yargs configuring time, we don't know what file will be invoked.
OR theother way could be 
2) to provide a nested structure for supplying the options for the beacon, validator, account etc in the same file, but  will require a little more handling to parse and supply the right config object at load time, sort of basically 2 rounds parsing process which mocha employs.

in both of these solutions beacon options and params file will keep working in the same way as it is but we will be able to provide an easy way to the user to dump his runtime args in a file in whatever way he/she wants unconstrained by the deeper parsing and handling of other config files.

let me know your thought process on this.

dapplion:
> @dapplion @wemeetagain i studied the mocha config, its config loading code is massaging node based args, merging package config as well as normal config as well as setting default values itself in the config object.
> 
> in light of above discussions, I propose that
> 
> 1. we can have a fixed cli args config file .lodestarrc.json/yml/js config irrespective of which command is being run to supplement cli args. it will act like env file and right .lodestarrc file would need to be copied/created. we need to have a common file (like global options) across all cmds as the yargs configuring time, we don't know what file will be invoked.
>    OR theother way could be
> 2. to provide a nested structure for supplying the options for the beacon, validator, account etc in the same file, but  will require a little more handling to parse and supply the right config object at load time, sort of basically 2 rounds parsing process which mocha employs.
> 
> in both of these solutions beacon options and params file will keep working in the same way as it is but we will be able to provide an easy way to the user to dump his runtime args in a file in whatever way he/she wants unconstrained by the deeper parsing and handling of other config files.
> 
> let me know your thought process on this.

We should read `.lodestarrc` if exists + have a flag to pass a config file.

Using a nested structure per command sounds complicated and it's not what JS users are used to in other JS cli apps.

One issue I faced while using an .env file was that if you use validator only options, mocha throws due to unknown options when running the beacon node. We may do our own option validation where we check options are valid on any command, and only error if no command declares that option

g11tech:
> We should read `.lodestarrc` if exists + have a flag to pass a config file.

+1 on `.lodestarrc`, additional config flag will require a 2 stage args parsing (first one for config flag) . if we are doing 2 stage args parsing, then we can can also have a different .rc files for different commands and on runtime load/look for the corresponding file to the build.
also .lodestarrc is to be json or yml file? or try looking to json,yml.js as the mocha does
> 
> Using a nested structure per command sounds complicated and it's not what JS users are used to in other JS cli apps.
> 
hmmmm 2 stage args parsing looks good.

> One issue I faced while using an .env file was that if you use validator only options, mocha throws due to unknown options when running the beacon node. We may do our own option validation where we check options are valid on any command, and only error if no command declares that option

yes 2 stage parsing  with different .rcs for different commands sounds much better as no such issues should arise parsing the corresponding file

g11tech:
closing the PR in favor of much simple solution: https://github.com/ChainSafe/lodestar/pull/2961

PR #2841, opened 2021-07-14
README checks

### LINKED ISSUE #2828: Investigate and implement tooling to ensure README files are accurate across all packages
Do this for all lodestar-relevant codebases (i.e. all inside monorepo as well as externals that are related such as ssz, bls-*, etc).

> Would be great to have tooling to ensure the readme is up to date. I think rust has something like that to ensure code snippets compile

from:
https://discord.com/channels/593655374469660673/605818244036821012/862719541838348318

3xtr4t3rr3str14l:
Resolves #2828 by adding a command in package.json's called `check-readme` and makes a CI action for running this check on all packages in the monorepo that have a README file.
Uses https://github.com/bbc/typescript-docs-verifier
Also fixed all errors in READMEs that the docs verifier found.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,3 +37,5 @@ jobs:
         run: yarn coverage
       - name: E2e tests
         run: yarn test:e2e
+      - name: README check
+        run: yarn run check-readme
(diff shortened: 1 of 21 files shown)


wemeetagain (approved):
awaesome, ezpz solution :+1:

PR #2842, opened 2021-07-15
Fix gossip block error handler

### LINKED ISSUE #2840: Node fails to track correct head
**Describe the bug**

In our nodes @tuyennhv and I have independently noticed that the node considers as head a block that's > 20 slots old (in Prater with healthy conditions).

Maybe due to https://github.com/ChainSafe/lodestar/pull/2696 we are not running updateHead() when we should? Also https://github.com/ChainSafe/lodestar/issues/2839 prevents to know if fork choice is actually being run.

This issue causes a significant drop in proposer rewards since the head is always wrong.

**Expected behavior**

Node should track the correct head.

**Steps to Reproduce**

Run node for a while, then query `/eth/v1/beacon/headers/head` and compare with a block explorer

twoeths:
**Motivation**

+ Node fails to track latest head due to incorrect error handler of Gossip Block handler when a future block or unknown parent root gossip block comes
+ `onBeaconBlockByRoot` handler is not correct for finalized block so we may get lower score as found by @mpetrunic and hence we may not receive gossip block frequently

**Description**

+ Use the new `BlockGossipError` instead of `BlockError`
+ Fix `onBeaconBlockByRoot` for finalized block
+ Add more logs

Other than that, move forkchocie `updateHead` after we emit/process block in order to process updated attestations in the block

Closes #2840

Sync prater

```
Jul-15 04:43:42.049 []                 info: Synced - finalized: 25561 0xc276…85a0 - head: 818018 0x94c9…d6f7 - clockSlot: 818018 - peers: 26
Jul-15 04:43:54.022 []                 info: Synced - finalized: 25561 0xc276…85a0 - head: 818019 0x0261…af05 - clockSlot: 818019 - peers: 26
Jul-15 04:44:06.002 []                 info: Synced - finalized: 25561 0xc276…85a0 - head: 818020 0xf595…464f - clockSlot: 818020 - peers: 27
```

### DIFF
--- a/packages/lodestar/src/network/gossip/handlers/index.ts
+++ b/packages/lodestar/src/network/gossip/handlers/index.ts
@@ -1,10 +1,10 @@
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
-import {ILogger} from "@chainsafe/lodestar-utils";
+import {ILogger, prettyBytes} from "@chainsafe/lodestar-utils";
 import {IMetrics} from "../../../metrics";
 import {OpSource} from "../../../metrics/validatorMonitor";
 import {IBeaconDb} from "../../../db";
 import {IBeaconChain} from "../../../chain";
-import {BlockError, BlockErrorCode} from "../../../chain/errors";
+import {BlockErrorCode, BlockGossipError} from "../../../chain/errors";
 import {GossipTopicMap, GossipType, GossipTypeMap} from "../interface";
 import {
   validateGossipAggregateAndProof,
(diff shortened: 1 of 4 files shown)


packages/lodestar/src/network/gossip/handlers/index.ts
```diff
@@ -71,9 +75,10 @@ export function getGossipHandlers(modules: ValidatorFnsModules): GossipHandlers
         }
       } catch (e) {
         if (
-          e instanceof BlockError &&
+          e instanceof BlockGossipError &&
```
twoeths, line 78: the main fix

twoeths:
the validator is profitable again https://prater.beaconcha.in/validator/0x811af134a14e68b8b14102bd8c5339864faa592b977eab427b245cc4e602c91f245f51ffb1ca04d0b36eeeaceac3ca6a#attestations

Screen Shot 2021-07-15 at 11 41 02

dapplion (approved):
Nice pick!

PR #2844, opened 2021-07-15
Add head drift chart

dapplion:
**Motivation**

Add metric track the issue from https://github.com/ChainSafe/lodestar/issues/2840

**Description**

Screenshot from 2021-07-15 12-09-44

This chart confirms that the issue had been happening for a while. Note that this value will be non zero if at least:
- The node is syncing
- There are many skipped slots
- Another issue like https://github.com/ChainSafe/lodestar/issues/2840


3xtr4t3rr3str14l (approved):
lgtm!

wemeetagain (approved):

PR #2849, opened 2021-07-17
Empty sync committee obj with G2_POINT_AT_INFINITY

dapplion:
**Motivation**

Current master does not follow the spec. It was detected on altair-devnet-1, where Lodestar nodes produce invalid blocks.

**Description**

- Must reject blocks with empty sync committee with signature not equal to infinity
- When producing blocks, must set sync committee signature to infinity if empty


dapplion:
Merging since this issue causes consensus failures :pray:

PR #2851, opened 2021-07-19
Implement aggregated attestation pool

### LINKED ISSUE #2845: Proposed block include old aggregated attestations
**Describe the bug**

On altair-dev-1, all other clients have 2 attestations per block like https://altair-devnet-1-explorer.ethdevops.io/block/95 , but block proposed by lodestar have more than 2 https://altair-devnet-1-explorer.ethdevops.io/block/96

Screenshot from 2021-07-15 15-14-46

Some more context from the discussion with other teams:
```
Danny Ryan, [Jul 15, 2021 at 20:02:51]:
Looks like you arent clearing your pool or something

dappLion, [Jul 15, 2021 at 20:02:55]:
so likely there are duplicates

Danny Ryan, [Jul 15, 2021 at 20:03:10]:
Because you have attestations from prior slots thag are full and already have been included
```

**Expected behavior**

We should prune old aggregated attestations so that they'll not be included in our proposed block

**Steps to Reproduce**

Run altair-devnet-1 with validators

twoeths:
**Motivation**

+ We implemented aggregated attestation pool as a repository and it includes so many redundant attestations in lodestar proposed block
+ This issue was found in altair-devnet-1

**Description**

Implement aggregated attestation pool referring to other client's implementation:
+ Partition aggregated attestations by slot and attestation data
+ Preaggregate attestation if there is no intersection regarding attestingIndices (or aggregationBits) to maximize profit
+ Use `attestingIndices` instead of `aggregationBits` for better performance
+ Remove attestations from seen attestations that are included in block chain
+ Validate all attestations before we include in block
+ Sort attestations by inclusion distance, then by number of attesters
+ Prune per slot
+ api should use it instead of the unaggregated attestation pool
+ Remove `aggregationAndProof` repository

Closes #2845

**Test**
Tested with prater by the api


packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -0,0 +1,216 @@
...
+ * or api. The returned attestations are aggregated to maximise the number of validators that can be
+ * included.
+ * Note that we want to remove attestations with attesters that were included in the chain.
+ */
+export class AggregatedAttestationPool {
+  private readonly attestationGroupByDataHashBySlot = new MapDef<
+    phase0.Slot,
+    Map<DataRootHex, MatchingDataAttestationGroup>
+  >(() => new Map<DataRootHex, MatchingDataAttestationGroup>());
+  // no need to maintain lowestPermissibleSlot since it's checked via gossip validation fn
```
dapplion: We must have this check here too. When our node gets extremely busy promises can be pending for many seconds even minutes. This protection is useful against ourselves

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -0,0 +1,216 @@
...
+      const attestationGroupByDataHash = this.attestationGroupByDataHashBySlot.get(slot);
+      // should not happen
+      if (!attestationGroupByDataHash) {
+        throw Error(`No aggregated attestation pool for slot=${slot}`);
+      }
+      const attestationGroups = Array.from(attestationGroupByDataHash.values());
+      attestations.push(...attestationGroups.map((group) => group.getAttestations()).flat());
+    }
+    // consumer should limit MAX_ATTESTATIONS items
+    return attestations.filter((attestation) => safeValidateAttestation(state, attestation));
```
dapplion: Is this safe performance wise? Why do we need to re-validate the attestation here?

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -0,0 +1,216 @@
...
+  add(attestation: phase0.Attestation, attestingIndices: ValidatorIndex[], committee: ValidatorIndex[]): InsertOutcome {
+    const slot = attestation.data.slot;
+    const attestationGroupByDataHash = this.attestationGroupByDataHashBySlot.getOrDefault(slot);
+    const dataRoot = ssz.phase0.AttestationData.hashTreeRoot(attestation.data);
+    const dataRootHex = toHexString(dataRoot);
+    let attestationGroup = attestationGroupByDataHash.get(dataRootHex);
+    if (attestationGroup) {
+      return attestationGroup.add({attestation, attestingIndices});
+    } else {
+      attestationGroup = new MatchingDataAttestationGroup({attestation, attestingIndices}, committee);
```
dapplion: Try out this Map variant to reduce code complexity here https://github.com/ChainSafe/lodestar/blob/70949d5d74956c692b243578d9845b8ab0cc0eea/packages/lodestar/src/util/map.ts#L1

Look for examples in the code on how to use it. May be helpful here

dapplion:
An aggregateAndProof is valuable during ~2 epochs. If our node restarts are we okay losing that data? We may want to consider persisting the opPool to disk on shutdown via a `close()` hook or with an abort signal, then restoring on start-up. Maybe we can defer doing that in another PR but keep it in mind during the design of the data structures

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
twoeths: this is part of the spec `The attestations added must satisfy the verification conditions found in attestation processing`, Teku and lighthouse both do that.

The `validateAttestation` is part of our state transition and contains some simple checks there https://github.com/ChainSafe/lodestar/blob/70949d5d74956c692b243578d9845b8ab0cc0eea/packages/beacon-state-transition/src/phase0/block/processAttestation.ts#L60

dapplion: I would ask to profile very well all this code in a testnet node or with performance tests to ensure we are good performance-wise

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
dapplion: Ou already used :+1:

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
dapplion: It still feels like lots of repeated work. The attestation was validated before to get here, what conditions can be invalidated from then and now?

twoeths:
> An aggregateAndProof is valuable during ~2 epochs. If our node restarts are we okay losing that data? We may want to consider persisting the opPool to disk on shutdown via a `close()` hook or with an abort signal, then restoring on start-up. Maybe we can defer doing that in another PR but keep it in mind during the design of the data structures

with the current good network condition, a block only contains attestations voting for previous block like in https://www.beaconcha.in/block/1660479#attestations because attestations in pool are all seen by chain

when we restart, it takes > 1 minute to search for peers, then `syncing...` then `Synced`, all of attestations in pool would be old and not valuable anymore even if we restart immediately. In addition, the possibility to publish a block right after we restart is so low.

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
twoeths: one condition is `attestation.aggregationBits.length !== committee.length`, another is `!(data.target.epoch === epochCtx.previousShuffling.epoch || data.target.epoch === epochCtx.currentShuffling.epoch)`

these were not checked in `validateGossipAggregateAndProof`

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
twoeths: added that!

twoeths:
I just changed the strategy to only filter attestations in pool when we produce a block. TODOs:
- [x] Implement getAttestationsForBlock for phase0, this is low priority since we're close to altair. It's mainly for the e2e.
- [x] Test with `altair-devnet-1`
- [x] Performance test shows that it takes 60ms - 80ms to run `getAttestationsForBlock()`

## Update:
+ TODOs is done
+ Attestations in pool are sorted by inclusion distance, then number of fresh (not-seen) attesters
+ The performance test shows that it takes 60ms-80ms for `getAttestationsForBlock()`
+ The pool api `curl http://localhost:9596/eth/v1/beacon/pool/attestations` should return same result to producing a block. Tested with `altair-devnet-1` and it returns 2-4 attestations which is similar to other clients.

packages/lodestar/src/chain/factory/block/body.ts
```diff
@@ -46,7 +46,7 @@ export async function assembleBody(
   const [proposerSlashings, attesterSlashings, attestations, voluntaryExits, {eth1Data, deposits}] = await Promise.all([
     db.proposerSlashing.values({limit: MAX_PROPOSER_SLASHINGS}),
     db.attesterSlashing.values({limit: MAX_ATTESTER_SLASHINGS}),
-    db.aggregateAndProof.getBlockAttestations(currentState).then((value) => value.slice(0, MAX_ATTESTATIONS)),
+    chain.aggregatedAttestationPool.getAttestationsForBlock(currentState).slice(0, MAX_ATTESTATIONS),
```
dapplion, line 49: Instead of slicing here the pool should be aware of MAX_ATTESTATIONS and stop doing work when it gets to MAX_ATTESTATIONS

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -0,0 +1,310 @@
...
+  phase0,
+  zipIndexesCommitteeBits,
+} from "@chainsafe/lodestar-beacon-state-transition";
+import {List, readonlyValues, toHexString} from "@chainsafe/ssz";
+import {MapDef} from "../../util/map";
+import {pruneBySlot} from "./utils";
+import {InsertOutcome, OpPoolError, OpPoolErrorCode} from "./types";
+
+type DataRootHex = string;
+/**
```
dapplion, line 30: Add a new line after the type for whitespace

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -0,0 +1,310 @@
...
+  >(() => new Map<DataRootHex, MatchingDataAttestationGroup>());
+  private lowestPermissibleSlot = 0;
+
+  add(attestation: phase0.Attestation, attestingIndices: ValidatorIndex[], committee: ValidatorIndex[]): InsertOutcome {
+    const slot = attestation.data.slot;
+    const lowestPermissibleSlot = this.lowestPermissibleSlot;
+
+    // Reject any attestations that are too old.
+    if (slot < lowestPermissibleSlot) {
+      throw new OpPoolError({code: OpPoolErrorCode.SLOT_TOO_LOW, slot, lowestPermissibleSlot});
```
dapplion: Seeng that this error happens frequently on the attestation pool, should it just be an InsertOutcome?

twoeths: I don't see any of it for this pool but think it's good to do that since the unaggregated pool throws a lot of errors like that. So we should return `AlreadyKnown`?

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -0,0 +1,310 @@
...
+    state: CachedBeaconState<allForks.BeaconState>,
+    isValidAttestation = false
+  ): phase0.Attestation[] {
+    if (state.config.getForkName(state.slot) !== ForkName.phase0) {
+      throw new Error(`Not support proposing block at slot ${state.slot}`);
+    }
+    const phase0State = state as CachedBeaconState<phase0.BeaconState>;
+    const {epochCtx} = phase0State;
+    const currentEpoch = computeEpochAtSlot(state.slot);
+    const extractParticipation = (attestations: List<phase0.PendingAttestation>): Set<ValidatorIndex> => {
```
dapplion: Could we prevent having closures if possible?

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -0,0 +1,310 @@
...
+   * Get attestations to be included in a phase0 block.
+   * As we are close to altair, this is not really important, it's mainly for e2e.
+   * The performance is not great due to the different BeaconState data structure to altair.
+   */
+  private getAttestationsForPhase0Block(
+    state: CachedBeaconState<allForks.BeaconState>,
+    isValidAttestation = false
+  ): phase0.Attestation[] {
+    if (state.config.getForkName(state.slot) !== ForkName.phase0) {
+      throw new Error(`Not support proposing block at slot ${state.slot}`);
```
dapplion: Is it necessary to do this check again here after https://github.com/ChainSafe/lodestar/pull/2851/files#diff-d9b2c635e7fcd4e4b3d0bddf21deb28e7d1b5f16a0ee899de077347bcbba77adR66 ?

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -0,0 +1,310 @@
...
+    const attestations: phase0.Attestation[] = [];
+    for (const slot of slots) {
+      const attestationGroupByDataHash = this.attestationGroupByDataHashBySlot.get(slot);
+      // should not happen
+      if (!attestationGroupByDataHash) {
+        throw Error(`No aggregated attestation pool for slot=${slot}`);
+      }
+      const attestationGroups = Array.from(attestationGroupByDataHash.values());
+      const epoch = computeEpochAtSlot(slot);
+      const participants =
```
dapplion: It necessary to compute this `participants` variable on every slot of the loop?

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -0,0 +1,310 @@
...
+    for (const existingAttestation of this.attestations) {
+      const existingAttestingIndices = existingAttestation.attestingIndices.valueOf() as ValidatorIndex[];
+      let numIntersection = 0;
+      for (const index of attestingIndices) {
+        if (existingAttestingIndices.includes(index)) numIntersection++;
+      }
+      // no intersection
+      if (numIntersection === 0) {
+        aggregateInto(existingAttestation, attestation, this.committee);
+        insertResult = InsertOutcome.Aggregated;
```
dapplion, line 247: Can we just return here
```ts
return InsertOutcome.Aggregated
```
?

twoeths, line 252: we want to aggregate all possible existing attestations so should not return there

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
twoeths: each slot may belong to either previous epoch or current epoch so we have to compute it here, it's just a `if/else` clause here since `currentEpochParticipants` and `previoiusEpochParticipants` are precomputed once before the loop.

we should focus more on the same function for altair btw

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -0,0 +1,310 @@
...
+      this.attestations.push(attestation);
+    }
+    return insertResult;
+  }
+
+  /** Remove all attestations with seen validators. */
+  removeBySeenValidators(seenAttestingIndices: ValidatorIndex[]): number {
+    let index = this.attestations.length - 1;
+    let numRemoved = 0;
+    while (index >= 0) {
```
dapplion: Why not just do a regular for loop of attestations?
```ts
for (let i=0; i < this.attestations.length; i++) {}
```
You can keep a `indexesToRemove` and then splice the array at once.

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -0,0 +1,310 @@
...
+      const existingAttestation = this.attestations[index];
+      existingAttestation.freshAttesterCount = existingAttestation.attestingIndices.filter(
+        (attIndex) => !seenAttestingIndices.includes(attIndex)
```
dapplion: seenAttestingIndices must be a Set or this is quadratic time

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -0,0 +1,310 @@
...
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+interface AttestationWithIndex {
+  attestation: phase0.Attestation;
+  attestingIndices: ValidatorIndex[];
+  // this is <= attestingIndices.count since some attesters may be seen by the chain
+  // this is only updated and used in removeBySeenValidators function
+  freshAttesterCount?: number;
```
dapplion: Should use `notSeenAttesterCount` for consistency with other function names

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -0,0 +1,310 @@
...
+   */
+  add(attestation: AttestationWithIndex): InsertOutcome {
+    const attestingIndices = attestation.attestingIndices.valueOf() as ValidatorIndex[];
+    // preaggregate
+    let insertResult = InsertOutcome.NewData;
+    for (const existingAttestation of this.attestations) {
+      const existingAttestingIndices = existingAttestation.attestingIndices.valueOf() as ValidatorIndex[];
+      let numIntersection = 0;
+      for (const index of attestingIndices) {
+        if (existingAttestingIndices.includes(index)) numIntersection++;
```
dapplion: Quadratic time, this should be a set

twoeths: not sure why using Set takes 5-10ms more than List ?! still go with a Set for now

dapplion: Oh damn! That's counter intuitive

dapplion: If not using a Set proves to be faster, then just use arrays

dapplion:
@tuyennhv I've applied some revisions on top of your branch. Didn't had time to update the tests, please if you accept my changes update those.

- Remove the need to get head state on API pool attestations
- Apply validateAttestation conditions once per slot instead of once per attestation
- Limit the max number of attestations per group
- Filter out attestations based on participation without persisting the result in the pool. We may propose blocks in differents forks so this is not something we want to persist.

I'm very curious to know what's the pattern of attestations in the pool, how does participations looks like? i.e. do we have attestations with wide participation that barely overlap each other,

Screenshot from 2021-07-22 22-34-52


or smaller attestations like the diagram below? 

Screenshot from 2021-07-22 22-35-09

We should run some tests after merging this branch and tune the constants for the behaviour we see fit.

twoeths:
Thanks for the commits, I think it's all good.

> I'm very curious to know what's the pattern of attestations in the pool, how does participations looks like? i.e. do we have attestations with wide participation that barely overlap each other,

having a look at https://www.beaconcha.in/block/1682189#attestations , I don't see attestations that overlap much each other, however there are some attestations with only 1-2 participations

however, an inspection from prater shows that it's common to have attestations with a lot of overlapped `aggregationBits` like:
```
0xdb5c476cfe355b7fffffbedbcadeddaf07
0xdb5c476cfe355b7fffffbedbcadfdd2f07
0xdb5c476cfe355b7fffffbedbcadfddaf07
0xdb5c476cfe357b7fffffbedbcbdfddaf07
```

Given `MAX_ATTESTATION=128` and 64 committees, we should include only 2 of them at most, or perhaps only the best one (similar to what I see on mainnet)

dapplion (approved):

PR #2852, opened 2021-07-19
regen metrics reference  impl

g11tech:
**Motivation**
This draft PR is to illustrate the regen metrics implementation talked here: 
https://github.com/ChainSafe/lodestar/issues/2153#issuecomment-877608829

right now it shows how state cache total/hits can be collected, and all other metrics mentioned (https://github.com/ChainSafe/lodestar/issues/2153#issuecomment-879171006)  could be collected in a similar fashion.

@wemeetagain  @dapplion  feedback on approach. thanks!

**Description**
just a sample image of how metrics would come out,
image

Closes #issue_number
#2153


packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -161,7 +161,19 @@ export function getValidatorApi({
...
-      const state = await chain.regen.getBlockSlotState(headRoot, slot);
+
+      const mChildLabels = {entrypointFn: "getBlockSlotState", callingModule: "produceAttestationData"};
+
+      const state = await chain.regen.getBlockSlotState(
+        headRoot,
+        slot,
+        metrics
+          ? {
+              stateLookupsTotal: metrics.regenStateCacheLookupTotal.child(mChildLabels),
```
dapplion: You are instantiating a child metrics instance on each call. Instantiation should only occur once. The .child pattern is probably not the right solution here

packages/lodestar/src/chain/chain.ts
```diff
@@ -169,11 +169,31 @@ export class BeaconChain implements IBeaconChain {
...
+            stateLookupHits: this.metrics.regenStateCacheLoopkupHits.child(mChildLabels),
+          }
+        : undefined
+    );
   }
 
   async getHeadStateAtCurrentSlot(): Promise<CachedBeaconState<allForks.BeaconState>> {
-    return await this.regen.getBlockSlotState(this.forkChoice.getHeadRoot(), this.clock.currentSlot);
+    const mChildLabels = {entrypointFn: "getBlockSlotState", callingModule: "getProposerDuties"};
+    return await this.regen.getBlockSlotState(
```
dapplion, line 178: Here you are telling the "getBlockSlotState" function that it's own name is "getBlockSlotState". That information should already be known, no need to pass it as an argument.

dapplion: The function only needs to know who is it calling it. There's not need to pass an entire metrics instance, you can just pass the name of the callingModule as an enum
```ts
function getBlockSlotState (blockRoot: Root, slot: Slot, caller?: GetBlockSlotStateCaller) {}

enum GetBlockSlotStateCaller {
  getProposerDuties = "getProposerDuties"
}
```

g11tech:
thanks @dapplion . let me rework and update!

g11tech:
thanks @dapplion . let me rework and update!

wemeetagain:
@g11tech do you think this is ready for another review?

g11tech:
> @g11tech do you think this is ready for another review?

yes please, just the dashboarding left, most probably couple will be in by tomorrow.

packages/lodestar/src/chain/regen/interface.ts
```diff
@@ -1,5 +1,38 @@
...
+  produceAttestationData = "produceAttestationData",
+  getStateByBlockRoot = "getStateByBlockRoot",
+  processBlocksInEpoch = "processBlocksInEpoch",
+  validateGossipAggregateAndProof = "validateGossipAggregateAndProof",
+  validateGossipAttestation = "validateGossipAttestation",
+  validateGossipVoluntaryExit = "validateGossipVoluntaryExit",
+  onForkChoiceFinalized = "onForkChoiceFinalized",
+}
+
+export enum RegenEntrypoint {
```
dapplion: Rename to `RegenFnName`

dapplion:
@g11tech Please checkout how https://github.com/ChainSafe/lodestar/pull/2876 changed the regen queue. You can simplify the design a lot by doing metrics in the jobQueueProcessor function

g11tech:
> @g11tech Please checkout how #2876 changed the regen queue. You can simplify the design a lot by doing metrics in the jobQueueProcessor function

so i take it that all calls to regen go through the jobQueueProcessor? then it would be simpler wrap around for counts, timing and errors.

g11tech:
regen cache dash
image

image

image

image

g11tech:
regen fn stats:

image

image

image

image

dapplion:
@g11tech Please don't stack the charts, it makes it much harder to reason about IMO. Unless the total amount means something (like total peers) I won't stack them

g11tech:
> @g11tech Please don't stack the charts, it makes it much harder to reason about IMO. Unless the total amount means something (like total peers) I won't stack them

Ummm the reason they are stacked is to show the spread clearly as well as the total volume (which is all the spreads together).
In all of the above stackings it's a drill down and hence total volume does mean something. for e.g. total calls in sucesses from vs total errors, total errors spead in the errors.

 I find them meaningful to give that info especially in debug times. 
However if you think it's not meaningful, I can unstack them. Let me know. 👍

dapplion:
> > @g11tech Please don't stack the charts, it makes it much harder to reason about IMO. Unless the total amount means something (like total peers) I won't stack them
> 
> Ummm the reason they are stacked is to show the spread clearly as well as the total volume (which is all the spreads together).
> In all of the above stackings it's a drill down and hence total volume does mean something. for e.g. total calls in sucesses from vs total errors, total errors spead in the errors.
> 
> I find them meaningful to give that info especially in debug times.
> However if you think it's not meaningful, I can unstack them. Let me know.

That makes sense, let's keep them stacked

packages/lodestar/src/chain/regen/queued.ts
```diff
@@ -52,72 +53,125 @@ export class QueuedStateRegenerator implements IStateRegenerator {
...
-      case "getPreState":
-        return this.regen.getPreState(regenRequest.args[0]);
-      case "getCheckpointState":
-        return this.regen.getCheckpointState(regenRequest.args[0]);
-      case "getBlockSlotState":
-        return this.regen.getBlockSlotState(...regenRequest.args);
-      case "getState":
-        return this.regen.getState(regenRequest.args[0]);
-    }
+    const wrappedWithMetrics = async (): Promise<CachedBeaconState<allForks.BeaconState>> => {
```
dapplion: Here, why do you declare a function to then call it?

packages/lodestar/src/chain/regen/queued.ts
```diff
@@ -52,72 +53,125 @@ export class QueuedStateRegenerator implements IStateRegenerator {
...
+    try {
+      // First attempt to fetch the state from cache before queueing
+      const state = this.stateCache.get(stateRoot, rmetrics);
+      if (state) {
+        return state;
+      }
+      // The state is not immediately available in the cache, enqueue the job
+      return this.jobQueue.push({key: "getState", args: [stateRoot, rCaller]});
+    } catch (e) {
+      this.metrics?.regenFnTotalErrors.inc({caller, entrypoint: RegenFnName.getPreState});
```
dapplion: Isn't this metric duplicated with the code in jobQueueProcessor? Same for all the methods

packages/lodestar/src/chain/regen/interface.ts
```diff
@@ -9,21 +43,28 @@ export interface IStateRegenerator {
...
+  getPreState: (block: allForks.BeaconBlock, rCaller: IRegenCaller) => Promise<CachedBeaconState<allForks.BeaconState>>;
 
   /**
    * Return a valid checkpoint state
    * This will always return a state with `state.slot % SLOTS_PER_EPOCH === 0`
    */
-  getCheckpointState: (cp: phase0.Checkpoint) => Promise<CachedBeaconState<allForks.BeaconState>>;
+  getCheckpointState: (
+    cp: phase0.Checkpoint,
+    rCaller: IRegenCaller
```
dapplion: You don't need to know the function name in `IRegenCaller`, this function signature can just be

```ts
getCheckpointState(cp: phase0.Checkpoint, caller: RegenCaller)
```

packages/lodestar/src/chain/regen/queued.ts
g11tech: because it was returning non awaited call in the original async jobqueueprocessor, so without looking into the queue implementation, i had to wrap it and return non awaited, and await the actual regen function inside to measure the metric.

packages/lodestar/src/chain/regen/queued.ts
```diff
@@ -52,72 +53,125 @@ export class QueuedStateRegenerator implements IStateRegenerator {
...
   }
 
-  async getState(stateRoot: Root): Promise<CachedBeaconState<allForks.BeaconState>> {
-    // First attempt to fetch the state from cache before queueing
-    const state = this.stateCache.get(stateRoot);
-    if (state) {
-      return state;
+  async getState(stateRoot: Root, rCaller: IRegenCaller): Promise<CachedBeaconState<allForks.BeaconState>> {
+    const {caller} = rCaller || {};
+    this.metrics?.regenFnCallTotal.inc({caller, entrypoint: RegenFnName.getState});
```
dapplion: I would differentiate between total calls that get here and calls that get queued. At this level I would just metric `regenFnCallTotal`, not errors, do that on jobQueueProcessor. Then at the top of jobQueueProcessor count `regenFnCallQueueTotal`. Then we can compute the ratio between those to see the cache hit vs cache miss.

packages/lodestar/src/chain/regen/queued.ts
g11tech: no, this counts any error thrown before it can be queued, once queued, this will be a successful return and any errors there will be caught mutually exclusively.

packages/lodestar/src/chain/stateCache/stateContextCache.ts
```diff
@@ -23,11 +27,13 @@ export class StateContextCache {
...
   }
 
-  get(root: ByteVector): CachedBeaconState<allForks.BeaconState> | null {
+  get(root: ByteVector, metrics?: IStateCacheMetrics): CachedBeaconState<allForks.BeaconState> | null {
+    metrics?.stateLookupsTotal.inc();
     const item = this.cache.get(toHexString(root));
     if (!item) {
       return null;
     }
+    metrics?.stateLookupHits.inc();
```
dapplion: I'm not sure how will we interpret this metrics. While the regen callers make sense, there are so many paths this get function can be called, register a miss but not actually be a problem. If you see https://github.com/ChainSafe/lodestar/pull/2852/files#diff-c09440f0657ffeff9d7ba94c89ff56b27b2f44c4eaf61b641c7ab28b8813be2aR183 there's a loop that may cause many cache miss while eventually resolving. We could start by only computing cache hit miss here but without labels.

packages/lodestar/src/chain/regen/regen.ts
```diff
@@ -152,7 +180,7 @@ export class StateRegenerator implements IStateRegenerator {
     const blocksToReplay = [block];
     let state: CachedBeaconState<allForks.BeaconState> | null = null;
     for (const b of this.forkChoice.iterateBlockSummaries(block.parentRoot)) {
-      state = this.stateCache.get(b.stateRoot);
+      state = this.stateCache.get(b.stateRoot, rmetrics);
```
dapplion: Instead of passing a full metrics instance as a function argument, why not pass a `IRegenCaller`? You can pass the metrics object through the state cache constructor

packages/lodestar/src/chain/regen/regen.ts
```diff
@@ -29,6 +29,7 @@ export class StateRegenerator implements IStateRegenerator {
   private checkpointStateCache: CheckpointStateCache;
   private db: IBeaconDb;
   private metrics: IMetrics | null;
+  private rmetricsMap: Map<string, IRegenFnMetrics>;
```
dapplion: This Map should not be necessary. You can find alternative patterns, like passing the metrics to the state cache constructor

packages/lodestar/src/chain/stateCache/stateContextCheckpointsCache.ts
```diff
@@ -15,8 +20,10 @@ export class CheckpointStateCache {
   /** Epoch -> Set<blockRoot> */
   private epochIndex = new Map<Epoch, Set<string>>();
 
-  get(cp: phase0.Checkpoint): CachedBeaconState<allForks.BeaconState> | null {
+  get(cp: phase0.Checkpoint, metrics?: IStateCheckpointCacheMetrics): CachedBeaconState<allForks.BeaconState> | null {
+    metrics?.stateCpLookupsTotal.inc();
     const item = this.cache.get(toCheckpointKey(cp));
+    if (item) metrics?.stateCpLookupsHits.inc();
```
dapplion: Same as https://github.com/ChainSafe/lodestar/pull/2852/files#r686155282

packages/lodestar/src/chain/validation/aggregateAndProof.ts
```diff
@@ -60,12 +61,14 @@ export async function validateGossipAggregateAndProof(
   // -- i.e. get_ancestor(store, aggregate.data.beacon_block_root, compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)) == store.finalized_checkpoint.root
   // > Altready check in `chain.forkChoice.hasBlock(attestation.data.beaconBlockRoot)`
 
-  const targetState = await chain.regen.getCheckpointState(attTarget).catch((e) => {
-    throw new AttestationError(GossipAction.REJECT, {
-      code: AttestationErrorCode.MISSING_ATTESTATION_TARGET_STATE,
-      error: e as Error,
+  const targetState = await chain.regen
+    .getCheckpointState(attTarget, {caller: RegenCaller.validateGossipAggregateAndProof})
```
dapplion: `getCheckpointState(attTarget, RegenCaller.validateGossipAggregateAndProof)`

packages/lodestar/src/chain/regen/interface.ts
```diff
@@ -1,5 +1,39 @@
...
+  produceBlock = "produceBlock",
+  validateGossipBlock = "validateGossipBlock",
+  produceAttestationData = "produceAttestationData",
+  getStateByBlockRoot = "getStateByBlockRoot",
+  processBlocksInEpoch = "processBlocksInEpoch",
+  validateGossipAggregateAndProof = "validateGossipAggregateAndProof",
+  validateGossipAttestation = "validateGossipAttestation",
+  validateGossipVoluntaryExit = "validateGossipVoluntaryExit",
+  onForkChoiceFinalized = "onForkChoiceFinalized",
+  dummyCallerfromTests = "dummyCallerfromTests",
```
dapplion: Remove `dummyCallerfromTests` and just use a random value from this enum

packages/lodestar/src/chain/regen/regen.ts
g11tech: actually, i abstracted out the cache metrics to be agnostic of the caller and pass it the object in which labels are pre bundeled, even when it could be called outside the regen modules otherwise it gets deeply tied to rcaller.

packages/lodestar/src/chain/regen/queued.ts
dapplion: I would remove anyway, the chance of getting an error here is very low

packages/lodestar/src/chain/regen/queued.ts
dapplion: It doesn't matter, behavior should be identical. You can remove the wrappedWithMetrics fn

packages/lodestar/src/chain/regen/regen.ts
dapplion: I would remove all labels from the state cache metrics, see https://github.com/ChainSafe/lodestar/pull/2852#discussion_r686155282

dapplion:
Good job overall! I would add a few more metrics if possible:

For both (1) state cache and (2) checkpoint state cache:
- cache size, use a collect function
- counter on the set / add function

For checkpoint state cache:
- count of epochs in epochIndex, use a collect function

packages/lodestar/src/chain/regen/queued.ts
g11tech: understood, do we need 'regenFnCallQueueTotal'? `regenFnCallDuration` is a histogram and gives a count along with it to do this ratio.

g11tech:
@dapplion  refactored as per feedback, also added the state/checkpoint state cache/add metrics plus checkpoint state epoch map size metrics.
grafana dashboard update pending, will come in the next commit.

g11tech:
new (updated) dashboards:

stateCache and stateCheckpointCache Stats:
image

Regen Fn Stats:
image
image

g11tech:
@dapplion  let me know if this looks good, will then rebase with current master and push after checking grafana dashboard json alignment with other changes that might have been pushed.

dapplion (approved):
Looks really good! Thank you for your this PR :rocket:

PR #2871, opened 2021-07-22
EpochProcess: Prepare activeValidatorIndices for the next epoch

### LINKED ISSUE #2866: Improve altair rotateEpoch

The `getTotalActiveBalance` function is expensive as shown below from a profile of altair sync and it takes 3.5% of the total time

Screen Shot 2021-07-21 at 14 16 57

We should be access that data from the epoch process

twoeths:
**Motivation**

+ On an altair syncing node, it takes 3.5% of the time to get active validator indices on `rotateEpoch`

**Description**

+ Calculate it in `IEpochProcess` at the end of the previous epoch to save 1 validators loop

Closes #2866

+ Sync `altair-devnet-1`
+ was able to sync to head with this branch
```
Jul-22 11:44:30.003 []                 info: Synced - finalized: 1551 0x5b68…4f6f - head: 49721 0xf9bd…6132 - clockSlot: 49722 - peers: 17
```

### DIFF
--- a/packages/beacon-state-transition/src/allForks/util/epochProcess.ts
+++ b/packages/beacon-state-transition/src/allForks/util/epochProcess.ts
@@ -55,7 +55,7 @@ export interface IEpochProcess {
   validators: phase0.Validator[];
   balances?: BigUint64Array;
   // to be used for rotateEpochs()
-  indicesBounded: [ValidatorIndex, Epoch, Epoch][];
+  nextEpochActiveValidatorIndices: ValidatorIndex[];
 }
 
 export function createIEpochProcess(): IEpochProcess {
@@ -79,7 +79,7 @@ export function createIEpochProcess(): IEpochProcess {
     churnLimit: 0,
     statuses: [],
     validators: [],
-    indicesBounded: [],
+    nextEpochActiveValidatorIndices: [],
   };
 }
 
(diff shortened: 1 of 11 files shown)


packages/beacon-state-transition/src/allForks/util/epochProcess.ts
```diff
@@ -145,6 +148,11 @@ export function prepareEpochProcessState<T extends allForks.BeaconState>(state:
 
     out.statuses.push(status);
     out.indicesBounded.push([i, v.activationEpoch, v.exitEpoch]);
+    if (nextEpoch >= config.ALTAIR_FORK_EPOCH) {
```
wemeetagain: This can be precalculated above this loop

twoeths:
Since we have `activeIndices` cached in `epochShuffling`, this PR is changed to making sure there is exactly 1 validators loop during epoch transition

```
Aug-04 12:14:18.012 []                 info: Synced - finalized: 1284 0x314c…667d - head: 41171 0x0da7…ae2c - clockSlot: 41171 - peers: 15
Aug-04 12:14:30.002 []                 info: Synced - finalized: 1284 0x314c…667d - head: 41172 0x62e7…77b5 - clockSlot: 41172 - peers: 15
```

dapplion:
:+1: Nice approach! This PR relies on `MAX_SEED_LOOKAHEAD >= 1`, otherwise it can't guarantee that the return value of `isActiveValidator(nextEpoch)` will be the same for the next epoch. You should add a check in somewhere so the application exists in a panic if that happens. `MAX_SEED_LOOKAHEAD` is a customizable variable, someone could run lodestar with `--params.MAX_SEED_LOOKAHEAD 0`

@tuyennhv please if possible tackle in a new PR

dapplion (approved):

PR #2875, opened 2021-07-22
Fix finalized block retrieval in API

wemeetagain:
**Motivation**
Stumbled upon these bugs.

**Description**
Several bugs in `resolveBlockIdOrNull` function.
- if `"finalized"` is provided, the finalized epoch was being used to fetch a block instead of the finalized block slot
  - The finalized block slot should be used to retrieve the finalized block
- if block hash or slot was provided, if the hash or slot corresponded to the finalized block, the block repository was searched, instead of the block archive repository
  - the block archive should be used to retrieve the finalized block

### DIFF
--- a/packages/lodestar/src/api/impl/beacon/blocks/utils.ts
+++ b/packages/lodestar/src/api/impl/beacon/blocks/utils.ts
@@ -5,7 +5,7 @@ import {IChainForkConfig} from "@chainsafe/lodestar-config";
 import {IForkChoice} from "@chainsafe/lodestar-fork-choice";
 import {IBeaconDb} from "../../../../db";
 import {GENESIS_SLOT} from "../../../../constants";
-import {fromHexString} from "@chainsafe/ssz";
+import {byteArrayEquals, fromHexString} from "@chainsafe/ssz";
 import {ApiError, ValidationError} from "../../errors";
 
 export function toBeaconHeaderResponse(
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/api/impl/beacon/blocks/utils.ts
```diff
@@ -52,14 +52,21 @@ async function resolveBlockIdOrNull(
...
 
   if (blockId.startsWith("0x")) {
     const root = fromHexString(blockId);
     const summary = forkChoice.getBlock(root);
     if (summary) {
-      return await db.block.get(summary.blockRoot);
+      // All unfinalized blocks and the finalized block are tracked by the fork choice.
+      // Unfinalized blocks are stored in the block repository, but the finalized block is in the block archive
+      const finalized = forkChoice.getFinalizedBlock();
+      if (byteArrayEquals(summary.blockRoot, finalized.blockRoot)) {
```
dapplion: Since it's finalized, can you compare by slot which should be cheaper?

packages/lodestar/src/api/impl/beacon/blocks/utils.ts
```diff
@@ -73,7 +80,14 @@ async function resolveBlockIdOrNull(
...
   if (blockSummary) {
-    return db.block.get(blockSummary.blockRoot);
+    // All unfinalized blocks and the finalized block are tracked by the fork choice.
+    // Unfinalized blocks are stored in the block repository, but the finalized block is in the block archive
+    const finalized = forkChoice.getFinalizedBlock();
+    if (byteArrayEquals(blockSummary.blockRoot, finalized.blockRoot)) {
+      return await db.blockArchive.get(blockSummary.slot);
+    } else {
+      return await db.block.get(blockSummary.blockRoot);
+    }
```
dapplion, line 83: Move to a function to deduplicate with above

dapplion (dismissed):

twoeths (dismissed):

packages/lodestar/src/api/impl/beacon/blocks/utils.ts
```diff
@@ -52,29 +52,37 @@ async function resolveBlockIdOrNull(
...
+
   if (blockId.startsWith("0x")) {
-    const root = fromHexString(blockId);
-    const summary = forkChoice.getBlock(root);
-    if (summary) {
-      return await db.block.get(summary.blockRoot);
-    } else {
-      return await db.blockArchive.getByRoot(root);
+    const blockHash = fromHexString(blockId);
+    blockSummary = forkChoice.getBlock(blockHash);
```
mpetrunic, line 63: isn't this gonna return null for finalized blocks?

wemeetagain: Yeaj, it will be null for finalized blocks (except _the_ finalized block).
See the comments lower in the function that describe what should happen.

dapplion (approved):

PR #2881, opened 2021-07-23
Make processEth1Data faster

dapplion:
**Motivation**

`getNewEth1Data` has shown up in CPU profiles in the past. The issue is comparing many eth1Data objects using ssz utility when it's not strictly necessary.

**Description**

Use serializeEth1Data to compare eth1Data objects

### DIFF
--- a/packages/beacon-state-transition/src/allForks/block/processEth1Data.ts
+++ b/packages/beacon-state-transition/src/allForks/block/processEth1Data.ts
@@ -1,6 +1,6 @@
 import {EPOCHS_PER_ETH1_VOTING_PERIOD, SLOTS_PER_EPOCH} from "@chainsafe/lodestar-params";
-import {allForks, phase0, ssz} from "@chainsafe/lodestar-types";
-import {readonlyValues} from "@chainsafe/ssz";
+import {allForks, phase0} from "@chainsafe/lodestar-types";
+import {readonlyValues, toHexString} from "@chainsafe/ssz";
 
 import {CachedBeaconState} from "../util";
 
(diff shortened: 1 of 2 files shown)


packages/beacon-state-transition/src/allForks/block/processEth1Data.ts
```diff
@@ -22,22 +22,39 @@ export function getNewEth1Data(
...
+    }
   }
+
+  // Not enough votes, won't change the result
+  return null;
+}
+
+/**
+ * Serialize eth1Data types to a unique string ID. It is only used for comparison.
+ * This serializer is x15 times faster than ssz.phase0.Eth1Data.equals()
```
mpetrunic, line 56: what about comparing each field individually with ssz.Root.Equals?

dapplion:
@mpetrunic Indeed it's faster even in the worst case!

```
  eth1Data isEqual
    ✓ each field with ssz.Root.equals                                      1364256 ops/s    733.0000 ns/op        -    4619587 runs   5.28 s
    ✓ compare with serializeEth1Data                                      257267.8 ops/s    3.887000 us/op        -    1183689 runs   5.08 s
    ✓ ssz.phase0.Eth1Data.equals                                          16932.79 ops/s    59.05700 us/op        -      84084 runs   5.01 s
```

@wemeetagain Why is ssz.phase0.Eth1Data.equals so much slower than manually comparing each field?

packages/beacon-state-transition/test/perf/allForks/util/eth1Data.test.ts
```diff
@@ -0,0 +1,54 @@
...
+// ✓ ssz.phase0.Eth1Data.equals                                          16932.79 ops/s    59.05700 us/op        -      84084 runs   5.01 s
+
+describe("eth1Data isEqual", () => {
+  setBenchOpts({
+    maxMs: 10 * 1000,
+    minMs: 5 * 1000,
+    runs: 1024,
+  });
+
+  const eth1Data1: phase0.Eth1Data = {
```
wemeetagain, line 19: This benchmark should be comparing two tree-backed eth1 datas.

dapplion:
Closing since when using tree backed this is not an issue

PR #2883, opened 2021-07-24
Aggregated attestation pool: handle 1 attestation includes another

### LINKED ISSUE #2882: [altair-devnet-1] Improve aggregated attestation pool
**Describe the bug**

There are still some redundant attestations in our aggregated attestation pool, for example block 62091
+ When we add an attestation to a group, we only check if existing attestation contains new one. We should check which one has greater attesting indices first before doing that.
+ Consider only getting 1 or 2 best attestations from a group, see https://github.com/ChainSafe/lodestar/pull/2851#issuecomment-885392204

**Expected behavior**

+ On both `mainnet` and `altair-devnet-1`, normally there are only 1 attestation per group (attestation data)

twoeths:
**Motivation**

+ We include up to 4 attestations of same data, 1 includes another. This is because we only check if an existing attestation contains another one, if a new attestation contains a superset of `aggregationBits` we don't check that
+ We include up to 4 attestations of same data in block which is too many. In mainnet, a slot has 64 committees and a block has ~~64~~ 128 attestations at most so we should reduce to 2.

**Description**

+ When an attestation comes, check which one has more `attestingIndices` first before checking for an inclusion. If new attestation's `aggregationBits` is a superset of an existing one, remove the existing one and add new.
+ Introduce `MAX_RETAINED_ATTESTATIONS_PER_GROUP` vs `MAX_ATTESTATIONS_PER_GROUP`

Closes #2882

**Test result with altair-devnet-1**

|Slot|This PR|The explorer (other clients)|
|----|-------|-------------|
|63819|2 * 63818 attestations|2 * 63818 attestations|
|63829|2 * 63828 attestations + 2 * 63822 attestations|2 * 63828 attestations|
|63844|2 * 63843 attestations + 2 * 63822 attestations|2 * 63843 attestations|
|63877|2 * 63876 attestations + 1 * 63856 attestation | 2 * 63876 attestations|

Not sure why we still have 1-2 extra old attestations included in block? if any attesters inside any attestation haven't seen by chain, we always include those, not sure why other clients don't!

At least, this PR confirms that we always have 2 attestations voted for previous block/slot which is the main point of the PR.

### DIFF
--- a/packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
+++ b/packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
@@ -25,7 +25,13 @@ type GetParticipationFn = (epoch: Epoch, committee: number[]) => Set<number> | n
  * After merging AggregatedAttestationPool, gather numbers from a real network and investigate
  * how does participation looks like in attestations.
  */
-const MAX_ATTESTATIONS_PER_GROUP = 4;
+const MAX_RETAINED_ATTESTATIONS_PER_GROUP = 4;
+
+/**
+ * On mainnet, each slot has 64 committees, and each block has 128 attestations max so we don't
+ * want to store more than 2 per group.
+ */
+const MAX_ATTESTATIONS_PER_GROUP = 2;
 
 /**
  * Maintain a pool of aggregated attestations. Attestations can be retrieved for inclusion in a block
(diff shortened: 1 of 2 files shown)


wemeetagain (approved):

dapplion (approved):
Nice find!

PR #2890, opened 2021-07-27
Enforce no relative package imports

dapplion:
**Motivation**

Importing from relative packages in the src structure breaks packages when distributed. This rule enforces that we don't do that.

https://github.com/import-js/eslint-plugin-import/blob/master/docs/rules/no-relative-packages.md

**Description**

Enforce no relative package imports with linter


mpetrunic (approved):

twoeths (approved):
👍 it's so useful to have this rule

PR #2891, opened 2021-07-27
Harden docker setup

dapplion:
**Motivation**

Our setup was too focused on our needs of rapid prototyping. This PR improves it a bit to be more secure.

**Description**

- Disable anonymous login in Grafana.
  - Users that just `docker-compose up` the setup, Grafana will prompt to change the password on the first login. Default credentials are u: `admin`, p: `admin`.
  - For us that we constantly re-build Grafana to update the charts we can set `GF_SECURITY_ADMIN_PASSWORD` ENV to a well known password by the team so we can always login with it across builds
 - Don't expose API port by default
 - Move NODE_OPTIONS to ENV to split beacon_node and validator limits. Only extend the memory ceiling for the beacon_node process, not also the validator.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "test:spec-main": "lerna run test:spec-main --no-bail",
     "benchmark": "node --max-old-space-size=4096 -r ts-node/register -r ./packages/lodestar/test/setup.ts ./node_modules/.bin/benchmark 'packages/*/test/perf/**/*.test.ts'",
     "benchmark:local": "yarn benchmark --local",
+    "cli": "node --trace-deprecation --max-old-space-size=8192 ./packages/cli/bin/lodestar",
     "publish:release": "lerna publish from-package --yes --no-verify-access",
     "release": "lerna version --no-push --sign-git-commit",
     "postrelease": "git tag -d $(git describe --abbrev=0)",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -26,7 +26,6 @@
     "test:spec-main": "lerna run test:spec-main --no-bail",
     "benchmark": "node --max-old-space-size=4096 -r ts-node/register -r ./packages/lodestar/test/setup.ts ./node_modules/.bin/benchmark 'packages/*/test/perf/**/*.test.ts'",
     "benchmark:local": "yarn benchmark --local",
-    "cli": "node --trace-deprecation --max-old-space-size=8192 ./packages/cli/bin/lodestar",
```
mpetrunic, line 29: whats the reason for removing this?

mpetrunic:
>  Remove cli link in package.json. Running the node through yarn is very problematic because signals are not properly propagated potentially causing zombie processes.

I don't think I had issues with that when not running in docker and this command is quite useful. Maybe we can replace it with some sh script

dapplion:
> > Remove cli link in package.json. Running the node through yarn is very problematic because signals are not properly propagated potentially causing zombie processes.
> 
> I don't think I had issues with that when not running in docker and this command is quite useful. Maybe we can replace it with some sh script

If we can force people to exclusive use npm not yarn to run this command then I'm okay with keeping it. Having a sh script is a good idea. Should it replace the soft link we have in the root?

dapplion:
@mpetrunic I've moved the CLI script change to a different PR to separate concerns, check https://github.com/ChainSafe/lodestar/pull/2893

wemeetagain (approved):

PR #2893, opened 2021-07-27
Move CLI script to .sh file

dapplion:
**Motivation**

Remove cli link in package.json. Running the node through yarn is very problematic because signals are not properly propagated potentially causing zombie processes.

**Description**

This script forces people to not use yarn to run this command, so signal handling is fine. It replaces the existing soft link so it can also extend the memory limit of node.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,8 @@ jobs:
         run: yarn build
         if: steps.cache-deps.outputs.cache-hit == 'true'
       # </common-build>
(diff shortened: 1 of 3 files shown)


mpetrunic (approved):

twoeths (approved):

wemeetagain (approved):
@dapplion in chat:
From running testnets for +1 month we consistently stay in the 3.5-4.5GB and only leave it when something really bad happens
In that case the node should die and not drag along

PR #2906, opened 2021-07-29
Process sync objects with concurrency

dapplion:
**Motivation**

In altair-devnet-2 we are too slow processing sync objects.

**Description**

Increase concurrency to be able to process more objects in one slot


dapplion:
Trying this fix live on altair-devnet-2 node B2

dapplion:
We can process all objects now

PR #2921, opened 2021-08-05
Set caches to transient/persistent

wemeetagain:
**Motivation**
Digging into slow `processInactivityUpdates` in perf tests.

It looks like `inactivityScores` cache wasn't being toggled to transient mode.
Also perf tests weren't properly toggling to transient mode either.


twoeths (approved):

dapplion:
I run a profile and it doesn't really matter if it's persistent or transient since CachedInactivityScoreList.set() will write to the tree anyway.
```ts
export class CachedInactivityScoreList implements List<Number64> {
  set(index: number, value: Number64): void {
    this.persistent.set(index, value);
    this.type.tree_setProperty(this.tree, index, value);
  }
}
```

The profile shows that most of the time is spent navigating the tree and setting individual values.

PR #2926, opened 2021-08-08
Benchmark array-ish structures

dapplion:
**Motivation**

To properly optimize our beacon state transition performance and memory usage we need to understand the tradeoffs or our different approaches.

**Description**

Add informational tests (not run in CI) with hardcoded results.

Notable results:
- Iterating an array is x10 faster than iterating a MutableVector
- Iterating a MutableVector is x100 times faster than iterating a Tree
- Regular JS arrays of numbers take 8 bytes per element
- MutableVector of numbers take 15 bytes per element
- Cloning a MutableVector has a fixed cost of ~1000 bytes. Cloning a MutableVector to mutate a few elements is very memory efficient even with the initial 1000 bytes cost

### DIFF
--- a/packages/beacon-state-transition/test/perf/dataStructures/arrayish.memory.ts
+++ b/packages/beacon-state-transition/test/perf/dataStructures/arrayish.memory.ts
@@ -15,10 +15,12 @@ enum TestType {
   MutableVector,
   MutableVectorClone,
   MutableVectorCloneAndMutate,
+  Set,
+  Map,
 }
 
-const size = 1000;
-const testType = TestType.MutableVector;
+const size = 100;
+const testType = TestType.Set;
 
 let arrayNumGlobal: number[] | null = null;
 let mutableVectorGlobal: MutableVector<number> | null = null;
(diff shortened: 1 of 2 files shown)


twoeths (approved):
Thanks for having this statistic 👍 , as we'll have more and more validators especially after the Merge, I suppose the loop speed is more and more important and we want to take a scalable approach.

To deduplicate validator data, I suggest keeping only validator roots in the tree (i.e. `validatorRoots: new ListType({elementType: Root, limit: VALIDATOR_REGISTRY_LIMIT})` and still keep `CachedValidatorList` to get the best of both world: the `hash`, the `loop` and access validator properties. I'm not sure how `serialize()` works for `CachedBeaconState` through.

What do you think @wemeetagain @dapplion ?

wemeetagain:
> To deduplicate validator data, I suggest keeping only validator roots in the tree

Definitely agree. I think the only question is how we should go about that.

You mentioned the tradeoff of storing the deserialized validators separately. Done naively, it breaks ssz serialization/deserialization (and proof generation).

Another approach may be to work within the ssz library to support hybrid tree-backed / struct-backed values. This could make it easier to maintain compatibility with the full range of ssz operations. The tradeoff being that it may be harder to customize / get the exact performance characteristics we're wanting in lodestar.

PR #2929, opened 2021-08-09
Compute totalActiveBalance in existing validators loop

dapplion:
**Motivation**

Noticed that we run an extra unnecessary loop over validators in epochCtx creation

**Description**

Compute total active balance on the existing loop over validators

### DIFF
--- a/packages/beacon-state-transition/src/allForks/util/epochContext.ts
+++ b/packages/beacon-state-transition/src/allForks/util/epochContext.ts
@@ -101,6 +101,8 @@ export function createEpochContext(
   const previousEpoch = currentEpoch === GENESIS_EPOCH ? GENESIS_EPOCH : currentEpoch - 1;
   const nextEpoch = currentEpoch + 1;
 
+  let totalActiveBalance = BigInt(0);
+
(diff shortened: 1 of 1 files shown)


wemeetagain (dismissed):

packages/beacon-state-transition/src/allForks/util/epochContext.ts
```diff
@@ -132,7 +135,6 @@ export function createEpochContext(
   // Only after altair, compute the indices of the current sync committee
   const onAltairFork = currentEpoch >= config.ALTAIR_FORK_EPOCH;
```
wemeetagain, line 142: After the loop, needs to check for 0, eg:
```ts
if (!totalActiveBalance) totalActiveBalance = BigInt(1);
```

wemeetagain (changes requested):

packages/beacon-state-transition/src/allForks/util/epochContext.ts
dapplion: Done

wemeetagain (approved):

PR #2933, opened 2021-08-10
Remove babel

dapplion:
**Motivation**

Through multiple tsc version it has been shown that babel does not add any performance increase in code compilation. Lodestar is not distributed directly to the browser, but only to NodeJS. We should only ensure compatibility with NodeJS v12.9.0. Browser compatibility should be ensured by Lodestar consumers, such as the Lightclient demo UI, which has babel and transpiles the code there.

See "StackOverflow Why do I need Babel JS?" https://stackoverflow.com/questions/39401836/why-do-i-need-babel-js

**Description**

Compile code with Typescript aiming at doing the least transformations possible to ensure high performance of direct NodeJS consumers.

----

PS: Sometimes the generated code by transpilers to support CommonJS has shown up in profilers, indicating it has a performance cost. We should:
- Investigate which features are left out when transpiling to es2019, and their performance cost
- Investigate the cost of CommonJS exports vs ES6 exports. In this branch I'm testing how feasible it is to switch https://github.com/ChainSafe/lodestar/tree/dapplion/esnext


tsconfig.build.json
```diff
@@ -19,7 +19,6 @@
     // Babel builds source, Typescript is used only for .d.ts files
```
wemeetagain: remove this line

wemeetagain (approved):

PR #2938, opened 2021-08-10
Process altair attestations in batch

dapplion:
**Motivation**

Small optimizations on altair attestation processing

**Description**

- Only update the balance of the proposer once for all attestations
- Prepare signature set more efficiently
- Cache block roots to reach each one once

### DIFF
--- a/packages/beacon-state-transition/src/altair/block/processAttestation.ts
+++ b/packages/beacon-state-transition/src/altair/block/processAttestation.ts
@@ -1,8 +1,8 @@
-import {allForks, altair, phase0, ssz} from "@chainsafe/lodestar-types";
+import {allForks, altair, Epoch, phase0, Root, Slot, ssz} from "@chainsafe/lodestar-types";
 import {intSqrt} from "@chainsafe/lodestar-utils";
 
 import {getBlockRoot, getBlockRootAtSlot, increaseBalance, verifySignatureSet} from "../../util";
-import {CachedBeaconState} from "../../allForks/util";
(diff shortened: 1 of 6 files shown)


packages/beacon-state-transition/src/altair/block/processAttestation.ts
```diff
@@ -17,69 +16,80 @@ import {
...
-      BigInt(!status.timelyTarget && timelyTarget) * TIMELY_TARGET_WEIGHT +
-      BigInt(!status.timelyHead && timelyHead) * TIMELY_HEAD_WEIGHT;
-    if (totalWeight > 0) {
-      totalBalancesWithWeight += state.validators[index].effectiveBalance * totalWeight;
+    const epochParticipation =
+      data.target.epoch === epochCtx.currentShuffling.epoch
+        ? state.currentEpochParticipation
+        : state.previousEpochParticipation;
+
+    const {timelySource, timelyTarget, timelyHead} = getAttestationParticipationStatus(
```
twoeths, line 63: attestations in same block tend to have same source/target/head, could you find a way to cache block roots queried from the state? it's not efficient to keep calling `state.blockRoots[i]` multiple times for same value

dapplion: Agree, but we should tackle that in another PR, not here

dapplion: Done! :+1:

wemeetagain (approved):

PR #2961, opened 2021-08-14
rcconfig option for any cmd

### LINKED ISSUE #2814: Add --configFile to validator

g11tech:
**Motivation**

This PR enables supplementing any cli arguments with additional options by providing a corresponding `--rcConfig` file to the lodestar cli in jso, tnl or yaml format.
 

**Description**

here is how it looks in the cli help:
image

config example for beacon (yml):
image

config example for validator (yml): 
image

Closes #2814

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -1,7 +1,7 @@
 // Must not use `* as yargs`, see https://github.com/yargs/yargs/issues/1131
 import yargs from "yargs";
 import {cmds} from "./cmds";
-import {globalOptions} from "./options";
+import {globalOptions, rcConfigOption} from "./options";
 import {registerCommandToYargs} from "./util";
 
 const topBanner = "🌟 Lodestar: Ethereum 2.0 TypeScript Implementation of the Beacon Chain";
@@ -38,6 +38,7 @@ export function getLodestarCli(): yargs.Argv {
 
   // throw an error if we see an unrecognized cmd
   lodestar.recommendCommands().strict();
+  lodestar.config(...rcConfigOption);
 
   return lodestar;
 }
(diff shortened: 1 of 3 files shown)


g11tech:
@wemeetagain  @dapplion turned out that 2 phase parsing isn't even required if we just allow the user to link the correct corresponding rcconfig with the command. :smiley: . I think this simple solution should suffice for now.

wemeetagain:
Would this replace the need for beacon.config.json?

g11tech:
> Would this replace the need for beacon.config.json?

umm in this avatar i don;t think so, as beacon config gets saved with final beacon (non default) options and gets reloaded on the next run even if last run cli options are skipped (though I am not sure of the scenario in which this will be useful), however i think its useful to inspect what final non default options object the beacon got instantiated with.

rcconfig however does no jazzy things and is a simple way to supplement command line args and therein lies its flexibility as well, as it can supplement any/all options, while beacon.config.json for e.g. options doesn't cover (for e.g.) network or rootdir param.

g11tech:
> umm in this avatar i don;t think so, as beacon config gets saved with final beacon (non default) options and gets reloaded on the next run even if last run cli options are skipped (though I am not sure of the scenario in which this will be useful)

if we don't need this, then beacon.config.json file reading can be done away with, though we can still dump the final options that goes into beacon init for inspection use, so its use will become more for debugging.

dapplion (approved):
Looks great! Thanks

PR #2963, opened 2021-08-16
Call state.slot once per function

twoeths:
**Motivation**

+ A `state.slot` call means navigating a persistent-merkle-tree node, extract number from bytes
+ A `state.slot ++` means we navigate the Node 2 times

**Description**

+ Avoid duplicate calls of `state.slot` per function by using a separate variable for it


dapplion:
Why closing?

twoeths:
@dapplion we have to maintain both `postStateSlot` and `postState.slot` and keep them in synced due to the `while` loop, I don't feel convenient with having to do that just for a small improvement so I closed it in the end, let me know if you think we should do that.

PR #2969, opened 2021-08-18
cli/cmds: fix dev startValidator index to 0:7

q9f:
**Motivation**

`lodestar dev` fails with:

```
 ✖ Error: Invalid startValidators arg '0:8' - state has 7 validators
    at Object.devHandler [as handler] (/home/user/.opt/chainsafe/lodestar/packages/cli/src/cmds/dev/handler.ts:116:13)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

**Description**

Default `startValidators=0:8` seems to be off by 1 and should be `0:7` for 8 validators.

I also clarified the variable names.

```sh
git checkout master
lodestar dev
```

### DIFF
--- a/packages/cli/src/cmds/dev/handler.ts
+++ b/packages/cli/src/cmds/dev/handler.ts
@@ -102,7 +102,7 @@ export async function devHandler(args: IDevArgs & IGlobalArgs): Promise<void> {
   if (args.startValidators) {
     const secretKeys: SecretKey[] = [];
     const [fromIndex, valCount] = args.startValidators.split(":").map((s) => parseInt(s));
-    const toIndex = valCount - 1;
+    const toIndex = valCount - fromIndex - 1;
     const maxIndex = anchorState.validators.length - 1;
 
     if (fromIndex > toIndex) {
--- a/packages/cli/src/cmds/dev/handler.ts
+++ b/packages/cli/src/cmds/dev/handler.ts
@@ -102,7 +102,7 @@ export async function devHandler(args: IDevArgs & IGlobalArgs): Promise<void> {
   if (args.startValidators) {
     const secretKeys: SecretKey[] = [];
     const [fromIndex, toIndex] = args.startValidators.split(":").map((s) => parseInt(s));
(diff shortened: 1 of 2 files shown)


packages/cli/src/cmds/dev/handler.ts
```diff
@@ -101,15 +101,16 @@ export async function devHandler(args: IDevArgs & IGlobalArgs): Promise<void> {
...
+    const toIndex = valCount - 1;
     const maxIndex = anchorState.validators.length - 1;
 
     if (fromIndex > toIndex) {
       throw Error(`Invalid startValidators arg '${args.startValidators}' - fromIndex > toIndex`);
     }
 
     if (toIndex > maxIndex) {
-      throw Error(`Invalid startValidators arg '${args.startValidators}' - state has ${maxIndex} validators`);
+      throw Error(`Invalid startValidators arg '${args.startValidators}' - state has ${valCount} validators`);
```
wemeetagain, line 113: should be `${valCount - fromIndex}`?
Maybe better to retain the old `toIndex` variable name

wemeetagain (changes requested):

packages/cli/src/cmds/dev/handler.ts
q9f: `const toIndex = valCount - fromIndex - 1;` like this? I was looking for confirmation that the old `toIndex` is actually the validator count. Maybe @dapplion can clarify.

wemeetagain: I think the `--startValidators` option would be `fromIndex:toIndex`
[fromIndex, toIndex)
including fromIndex and up to, but not including toIndex.

q9f:
Ok, thanks for the reviews. I identified the issue was in the default value of `startValidators = "0:8"`.

wemeetagain (approved):

dapplion (approved):

dapplion:
Thanks @q9f !

PR #2971, opened 2021-08-18
Fix issue in parseJsonRpcErrorCode

0xFredrik:
Checks were turned around and always returned false.

**Motivation**

Fix a minor bug

**Description**

Checks were turned around in "if (code >= -32000 && code <= -32099) return "Server error";" and so always returned false.


wemeetagain (approved):
Thanks for the contribution :)

dapplion (approved):
Nice pick! Thank you so much for contributing the fix :+1:

PR #2980, opened 2021-08-19
Add more accurate epoch perf tests

dapplion:
**Motivation**

Our current performance tests are good but don't cover all functions in all possible conditions. Therefore when doing tradeoffs we can underestimate the cost of performance in unfavorable network conditions.

**Description**

Ensure that our performance tests reflect real network conditions. Add tests for hypotetical worst network conditions so we aren't caught by surprise and our client crashes.

- Analyze network conditions https://github.com/ChainSafe/lodestar/pull/2979
- Add normal case + worst case for each expensive epoch transition step
- Re-organize tests.
  - `/$fork/epoch.test.ts` runs the entire epoch transition + hashing against real states
  - Individual function tests run hypothetical optimal case + hypothetical worst case.


dapplion:
About length (push / pr):
- Spec tests: 7m 59s / 8m 31s
- Sim test: 12m 16s / 12m 9s
- Tests: 10m 31s / 10m 49s

Build and setup is about 1m 40s, so benchmark tests in total should take ~9m

packages/beacon-state-transition/src/altair/epoch/processInactivityUpdates.ts
```diff
@@ -23,12 +30,13 @@ export function processInactivityUpdates(
...
+  const {config, inactivityScores} = state;
   const {INACTIVITY_SCORE_BIAS, INACTIVITY_SCORE_RECOVERY_RATE} = config;
+  const {statuses} = epochProcess;
   const inActivityLeak = isInInactivityLeak((state as unknown) as phase0.BeaconState);
-  const {inactivityScores} = state;
-  for (let i = 0; i < epochProcess.statuses.length; i++) {
-    const status = epochProcess.statuses[i];
+
+  for (let i = 0; i < statuses.length; i++) {
+    const status = statuses[i];
```
dapplion, line 39: Only change in `/src`, just cosmetic to get variables once before the loop

packages/beacon-state-transition/src/allForks/util/attesterStatus.ts
```diff
@@ -60,3 +60,16 @@ export function parseAttesterFlags(flags: number): AttesterFlags {
     eligibleAttester: hasMarkers(flags, FLAG_ELIGIBLE_ATTESTER),
   };
 }
+
+export function toAttestarFlags(flagsObj: AttesterFlags): number {
```
wemeetagain: typo: Attestar -> Attester

wemeetagain (approved):
Awesome! This adds so many great benchmarks

PR #2988, opened 2021-08-22
Run tests across all packages directly

dapplion:
**Motivation**

Running tests with lerna is very annoying because you have to scroll a lot to figure out which package actually had a failing test.

*before*
```
// Package A
✓ A1 Ok
x A2 Error

// Package B
✓ B1 Ok
x B2 Error
```
*after this PR*
```
✓ A1 Ok
✓ B1 Ok
x A2 Error
x B2 Error
```

**Description**

Run the unit tests across all packages with a single mocha run. If tests fail in multiple packages they will be nicely summarized at the end :+1: 

**TODO**
- [ ] Make all tests work with the same preset, types and tsconfig options


dapplion:
- Closing for issue https://github.com/ChainSafe/lodestar/issues/4291 as won't be fixed now

PR #2989, opened 2021-08-22
Move network configs to lodestar-config

dapplion:
**Motivation**

Working on improving the performance tests I needed the config of Pyrmont. However it was buried deep in the CLI code. I believe it's useful to expose well-known testnet configs in higher up package like lodestar-config. The CLI can still declare less useful networks like devnets if it really wants.

**Description**

- Move mainnet, pyrmont and prater configs to lodestar-config
- Expose network configs and presets configs in root package on `lodestar-config/networks` and `lodestar-config/presets`. Don't use specific names so we don't have to update them once new testnets appear.

Example usage:

```ts
import {NetworkName, networksChainConfig} from "@chainsafe/lodestar-config/networks";

export function getNetworkConfig(network: NetworkName): IChainConfig {
  return networksChainConfig[network];
}
```

### DIFF
--- a/packages/cli/src/config/beaconParams.ts
+++ b/packages/cli/src/config/beaconParams.ts
@@ -62,15 +62,18 @@ export function getBeaconConfig(args: IBeaconParamsArgs): IChainForkConfig {
  * - CLI flags
  */
 export function getBeaconParams({network, paramsFile, additionalParamsCli}: IBeaconParamsArgs): IChainConfig {
-  const additionalParams = mergeBeaconParams(
-    // Default network params
-    network ? getNetworkBeaconParams(network) : {},
-    // Existing user custom params from file
-    readBeaconParamsIfExists(paramsFile),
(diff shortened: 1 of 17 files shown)


twoeths (approved):

q9f:
Is there any way to also use the module's minimal preset like:

```
import {minimalChainConfig} from "@chainsafe/lodestar-config/presets";
```

For some reason, I can only access the mainnet preset. (I want to stub a `dev` network.)

```
src/networks/dev.ts:3:9 - error TS2305: Module '"@chainsafe/lodestar-config/presets"' has no exported member 'minimalChainConfig'.
  import {minimalChainConfig} from "@chainsafe/lodestar-config/presets";
            ~~~~~~~~~~~~~~~~~~
```

Or does this require a release of the package first because it uses an old version of the lodestar-config package?

Edit: fixed in #2995

PR #2998, opened 2021-08-23
Update ssz and persistent-merkle-tree

wemeetagain:


twoeths (approved):

twoeths:
it looks like `processRewardsAndPenalties` already improved a lot with new ssz, I guess it's due to less memory used?

wemeetagain:
We also made iterating thru merkle nodes faster (use arrays and simple for loops rather than iterable), may also be part of it

PR #3001, opened 2021-08-24
config file is now used to dump finalized options

g11tech:
**Motivation**
Beacon node now will not use configFile to read beacon options, rather the use of--configFile is now to dump finalized beaconNode options, 
even though we discussed logging the options instead of writing them back to file (for inspection/debugging), found  writing to file much cleaner instead of logging the options on the logger. let know if you feel differently. @dapplion @wemeetagain 

**Description**
image

Closes #issue_number

### DIFF
--- a/packages/cli/src/cmds/beacon/handler.ts
+++ b/packages/cli/src/cmds/beacon/handler.ts
@@ -24,11 +24,20 @@ export async function beaconHandler(args: IBeaconArgs & IGlobalArgs): Promise<vo
   await initBLS();
 
   const {beaconNodeOptions, config} = await initializeOptionsAndConfig(args);
-  await persistOptionsAndConfig(args, beaconNodeOptions, config);
+  await persistOptionsAndConfig(args, config);
 
   const version = getVersion();
   const gitData = getVersionGitData();
   const beaconPaths = getBeaconPaths(args);
+
+  const logger = getCliLogger(args, beaconPaths, config);
+  const abortController = new AbortController();
+  onGracefulShutdown(async () => {
+    abortController.abort();
+  }, logger.info.bind(logger));
+
+  logger.info("Lodestar", {version: version, network: args.network});
+
   // TODO: Rename db.name to db.path or db.location
   beaconNodeOptions.set({db: {name: beaconPaths.dbDir}});
   // Add metrics metadata to show versioning + network info in Prometheus + Grafana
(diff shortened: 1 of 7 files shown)


packages/cli/src/config/beaconNodeOptions.ts
```diff
@@ -16,16 +16,13 @@ export class BeaconNodeOptions {
...
-    configFile,
     beaconNodeOptionsCli,
   }: {
     network?: NetworkName;
-    configFile?: string;
     beaconNodeOptionsCli: RecursivePartial<IBeaconNodeOptions>;
   }) {
     this.beaconNodeOptions = mergeBeaconNodeOptions(
       network ? getNetworkBeaconNodeOptions(network) : {},
-      configFile ? readBeaconNodeOptionsIfExists(configFile) : {},
```
dapplion, line 28: Why not read from the config file? We want to *not* persist the options, but we do want to *read* them right? @wemeetagain

packages/cli/src/cmds/beacon/handler.ts
```diff
@@ -24,11 +24,20 @@ export async function beaconHandler(args: IBeaconArgs & IGlobalArgs): Promise<vo
   await initBLS();
 
   const {beaconNodeOptions, config} = await initializeOptionsAndConfig(args);
-  await persistOptionsAndConfig(args, beaconNodeOptions, config);
+  await persistOptionsAndConfig(args, config);
```
dapplion, line 27: Note that the bug in Pyrmont was caused by reading the config object, not the beaconNodeOptions

dapplion:
I think the behavior you propose is very un-intuitive. If I use an option called `--config` I expect the CLI to read from it, not write to it.

packages/cli/src/config/beaconNodeOptions.ts
g11tech: ummm i thought after merging rcconfig we don't need to read options from any other config, i can change the name of the flag to something like `dumpConfig`. Let me know if I got the discussion wrong.

g11tech:
> I think the behavior you propose is very un-intuitive. If I use an option called `--config` I expect the CLI to read from it, not write to it.

my understanding was to remove the reading from the beacon config, just dump the final object for inspection/debugging purposes, or if the beacon config is important, then can add the dumpConfig as separate options and remove just persisting the  options  on beacon config.

dapplion:
> > I think the behavior you propose is very un-intuitive. If I use an option called `--config` I expect the CLI to read from it, not write to it.
> 
> my understanding was to remove the reading from the beacon config, just dump the final object for inspection/debugging purposes, or if the beacon config is important, then can add the dumpConfig as separate options and remove just persisting the options on beacon config.

The source problem is:
1. Persisting default values
2. Reading those default values

What we have to change is:
- Do not persist default values. Taking it further, do not persist values unless requested explicitly by the user
- Don't read a file by default. Only read a config file if the user requested it explicitly via a flag

If you thing a different flow provides better UX let me know. I think this is what all nodes usually do.

g11tech:
Ok, my doubt is a bit fundamental, basically I want to understand which scenario will the beacon options config be useful since now we have rcconfig. 

May be understanding that scenario will help me figure out the right way to handle the beacon options config.

dapplion:
> Ok, my doubt is a bit fundamental, basically I want to understand which scenario will the beacon options config be useful since now we have rcconfig.
> 
> May be understanding that scenario will help me figure out the right way to handle the beacon options config.

Now you have multiple ways of providing options, which can cause confusion. Keep this PR with the goals of fixing the "reading defaults" issue which is serious urgent and should be fixed ASAP: If you want we can discuss latter how to rethink options and configs overall. However I think we have much more urgent needs elsewhere in the project.

dapplion:
This issue almost kills our nodes in the Prater fork :cry: @g11tech can you fix it asap? Otherwise I'll take over the PR

g11tech:
> This issue almost kills our nodes in the Prater fork cry @g11tech can you fix it asap? Otherwise I'll take over the PR

ok, i'll work on it :+1:

g11tech:
addressing this issue with another approach(es)

PR #3003, opened 2021-08-24
cli/networks: expose a dev network based on minimal preset

### LINKED ISSUE #2594: Wrong config in CLI dev cmd
**Describe the bug**

Running the `dev` command in the CLI with `--minimal` I got 12 seconds per slot. Something is wrong the the config generation.

**Expected behavior**

Minimal should have 6 seconds per slot.

q9f:
_WIP_

**Motivation**

Goal is to fix #2594

**Description**

Dev is _supposed_ to use the minimal preset but something something inserted mainnet defaults into all networks. _WIP_


packages/config/presets.d.ts
```diff
@@ -1 +1 @@
-export * from "./lib/networks";
+export * from "./lib/presets";
```
q9f: It took me a day to figure out why `mainnet` works but `minimal` didn't ;)

dapplion: Sorry :pray:

dapplion:
@q9f Can you rebase to master so CI passes?

q9f:
Currently, there is no easy way to use the minimal preset in the lodestar CLI because the ACTIVE_PRESET gets locked with the first import of params, and this usually pulls in the Mainnet defaults. For now, we can just keep dev as per #2594 constraints.

In the future, if necessary, we could either remove or split out dev into its own application that specifically pulls in params with the minimal active preset.

PR #2672, opened 2021-06-09
Dockerfile build w/ NPM

dapplion:
**Motivation**

Simplify and speed up builds

**Description**

Use NPM instead of building from source.

For nightly builds do

```
docker-compose build --build-arg VERSION=next
```
or
```
docker-compose build --build-arg VERSION=0.24.3-next.48c325709d978a574b215beb34c8a18a7d45cffe.0
```

PR #3006, opened 2021-08-24
github: allow release workflow to wait up to 15 minutes for npm package to publish

### LINKED ISSUE #2999: Releases not uploaded to Dockerhub
**Describe the bug**

At least version 0.28.0 is not uploaded to Dockerhub. The publish to Dockerhub action failed. https://github.com/ChainSafe/lodestar/runs/3255490028

```
Step 5/12 : ENV VERSION=$VERSION
 ---> Running in ce678a409691
Removing intermediate container ce678a409691
 ---> 47fbf0c68360
Step 6/12 : RUN npm install @chainsafe/lodestar-cli@$VERSION
 ---> Running in eade1e96c7d9
npm ERR! code ETARGET
npm ERR! notarget No matching version found for @chainsafe/lodestar-cli@0.28.0.
npm ERR! notarget In most cases you or one of your dependencies are requesting
npm ERR! notarget a package version that doesn't exist.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2021-08-05T18_50_04_526Z-debug.log
The command '/bin/sh -c npm install @chainsafe/lodestar-cli@$VERSION' returned a non-zero code: 1
```

I think the problem is a race condition with NPM. We have a prior action that uploads that version to NPM, then run the Dockerfile that installs from NPM. Since the timing is tight, it may happen that NPM doesn't acknowledge the existence of that version.

**Expected behavior**

All releases should be uploaded to Dockerhub

q9f:
**Motivation**

Trying to fix #2999 where releases are no longer pushed to docker hub.

**Description**

The issue was introduced in #2672 unknowingly that dist releases might be delayed in the publication on the npm package registry side.

This adds a script that only runs on release build to verify that the npm version exists before attempting and gracefully takes a nap in case it cannot find the required version.

For now I limited it to ~~60~~ 900 seconds but we can play around with this number. ~~Honestly, 60 should be fine, we can test that though.~~

Remaining question: Why are nightly builds not failing? I was staring at the workflows and nightly releases should face the same issue but they don't. Maybe NPM handles them differently.

Fix #2999


.github/workflows/release.yml
```diff
@@ -116,6 +116,7 @@ jobs:
     if: needs.tag.outputs.tag != ''
     steps:
       - uses: actions/checkout@v2
+      - run: scripts/await-release.sh ${{ needs.tag.outputs.tag }} 60
```
dapplion: I would do much more than 60 seconds, like 10 or 20 minutes haha

dapplion (commented):
Nice job! :+1:

q9f:
Bumped it to 15 minutes. We can test this with #3012 WDYT?

dapplion:
> Bumped it to 15 minutes. We can test this with #3012 WDYT?

Yeah let's do it

dapplion (approved):

PR #3009, opened 2021-08-24
Drop flat caches

dapplion:
**Motivation**

CachedBeaconState duplicates much of the data in the state. Lodestar struggles with too much memory, so it's imperative data is represented only once.

**Description**

WIP


dapplion:
Replaced by https://github.com/ChainSafe/lodestar/pull/3760

PR #3013, opened 2021-08-25
Add config for Prater fork

dapplion:
**Motivation**

See https://github.com/eth2-clients/eth2-networks/pull/58

**Description**

Update Prater config


q9f:
What about `INACTIVITY_SCORE_BIAS` and `INACTIVITY_SCORE_RECOVERY_RATE`? Are they unrelated to Altair fork?

wemeetagain:
> What about `INACTIVITY_SCORE_BIAS` and `INACTIVITY_SCORE_RECOVERY_RATE`? Are they unrelated to Altair fork?

It looks like those are already covered by the mainnet config and don't change from the default.

wemeetagain (approved):

PR #3023, opened 2021-08-26
Altair: precalculate nextEpoch total active balance

twoeths:
**Motivation**

We calculate total active balance for altair in `afterProcessEpoch` while we do some loops during processing epoch

**Description**

Inside `processEffectiveBalanceUpdates`, precalculate next epoch's total active balance if it's altair.


dapplion:
Please add a nice multi-line comment explaining why there's non-spec-ed code in this function, the tradeoff and the approx performance benefit. As inspiration please take a look at https://github.com/ChainSafe/lodestar/pull/3009/files#diff-972c30309af0b5e9c82e95c1729f11612914cd5ce2b843e265e397f2773b98a8R172 this is extremely detailed but something in that direction is necessary. Also check Lighthouse source code comments on complex tradeoffs as an example too

dapplion (approved):
Thanks for adding the comments :pray:

PR #3028, opened 2021-08-27
Improve perf test for altair block

twoeths:
**Motivation**

For perf test for altair block, we want to produce a block as closed to reality as possible

**Description**

+ Attestation's committee index should not always be 0
+ Correct the head/target root of attestation
+ Reset epoch participation flag of state in order to mark the attesting indices have not been seen by the network, hence we'll calculate "totalWeight" for the proposer reward https://github.com/ChainSafe/lodestar/blob/cd6e6a4f3b51c9fbb48dfc73390da933e6fc5b8f/packages/beacon-state-transition/src/altair/block/processAttestation.ts#L91


twoeths:
these benchmark tests take more time as expected



|Benchmark suite | Current: b097982 | Previous: cd6e6a4 | Ratio|
| ---------------------------------------------------- | -------------- | -------------- | ---- |
| altair processBlock - 250000 vs - 7PWei normalcase | 136.09 ms/op | 92.186 ms/op | 1.48 |
| altair processBlock - 250000 vs - 7PWei worstcase | 338.97 ms/op | 234.27 ms/op | 1.45 |

dapplion (approved):

PR #3031, opened 2021-08-27
Fix typo in processSlashings perf test

dapplion:
**Motivation**

processSlashings perf test isn't actually testing processSlashings fn

**Description**

Test processSlashings fn in processSlashings perf test


dapplion:
> ##  **Performance Alert** 
> Possible performance regression was detected for some benchmarks.
> Benchmark result of this commit is worse than the previous benchmark result exceeding threshold.
> 
> Benchmark suite	Current: [6777148](https://github.com/ChainSafe/lodestar/commit/6777148da0e0c2a00bbfbf97e4368876da180754)	Previous: [f360283](https://github.com/ChainSafe/lodestar/commit/f360283da69c4d560eabac65467bf1ac212faa6a)	Ratio
> phase0 processSlashings - 250000 worstcase	87.690 ms/op	23.748 us/op	3692.52
> Full benchmark results
> > by benchmarkbot/action

Known regression the test was not running the correct function

PR #3043, opened 2021-08-30
Fix hashing perf test

dapplion:
**Motivation**

hashing perf tests results show the same values for all cases. The issue is that the state is not being cloned so the hashing cache is shared between tests.

**Description**

Fix hashing perf test

### DIFF
--- a/packages/beacon-state-transition/test/perf/allForks/hashing.test.ts
+++ b/packages/beacon-state-transition/test/perf/allForks/hashing.test.ts
@@ -14,7 +14,7 @@ describe("BeaconState hashTreeRoot", () => {
   before(function () {
     this.timeout(300_000);
     const {pubkeys} = getPubkeys();
-    stateOg = ssz.phase0.BeaconState.createTreeBacked(generatePerformanceStatePhase0(pubkeys).tree);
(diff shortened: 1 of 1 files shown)


packages/beacon-state-transition/test/perf/allForks/hashing.test.ts
```diff
@@ -6,15 +6,15 @@ import {unshuffleList} from "../../../src";
...
+describe("BeaconState hashTreeRoot", () => {
   const vc = numValidators;
   const indicesShuffled: number[] = [];
   let stateOg: TreeBacked<phase0.BeaconState>;
 
   before(function () {
     this.timeout(300_000);
     const {pubkeys} = getPubkeys();
-    stateOg = generatePerformanceStatePhase0(pubkeys);
+    stateOg = ssz.phase0.BeaconState.createTreeBacked(generatePerformanceStatePhase0(pubkeys).tree);
```
wemeetagain: no need?

wemeetagain:
```suggestion
    stateOg = generatePerformanceStatePhase0(pubkeys);
```

PR #2973, opened 2021-08-19
Track balances as Number64UintType

twoeths:
**Motivation**

Track `balances` as the new `Number64UintType` of ssz for faster processing, same for slot, epoch

**Description**

This depends on https://github.com/ChainSafe/ssz/pull/159

+ For all Number64 type, use the new `Number64UintType` of ssz
+ Track balances as a number too `new ListType({elementType: Number64, limit: VALIDATOR_REGISTRY_LIMIT})`
+ Update CachedBalancesList: remove `persistent-ts` array there to save memory, move it to `balanceList.ts`
+ Increase balance/decrease balance: use the new api `tree_applyDeltaAtIndex()` or `tree_applyDeltaInBatch()` to avoid the pattern `balances[i] = balances[i] + delta` (which have to do `deserialize` then `serialize`)
+ `processRewardsAndPenalties`: use the new api `tree_newTreeFromDeltas()` to build balances tree and avoid BigInt calculation and Uint8Array allocation, it return new number[] array to be used for next step of epoch transition

Closes #2972 #2049

More related improvement could be done after this PR:
+ Track effectiveBalance as number too, may simplify calculation at different places a bit (see #2981)

PR #3044, opened 2021-08-31
Track effective balance as number

### LINKED ISSUE #2981: Track effectiveBalance as small number

We should use number whenever possible as BigInt calculation is about x250 times slower as analysed by @dapplion 

1. effectiveBalance is always a multiple of 1e9 and it's <= 32*1e9 so should be able to track 30, 31, 32 instead of BigInt(30 * 1e9), BigInt(31 * 1e9), BigInt(32 * 1e9). This would resolve issue #2747
2. In `epochProcess.ts`, totalActiveStake, prevEpochUnslashedStake and currEpochUnslashedTargetStake could also be tracked as number since they are sum of effectiveBalances
    + May need to tweak a little bit for some functions like `computeBaseRewardPerIncrement(), processEffectiveBalanceUpdates()`
    + Need to do an inventory for the usage of all of these variables but as far as I see it's all multiply/divide by `EFFECTIVE_BALANCE_INCREMENT` and this constant (1e9) is not likely to be changed

This should speed up `beforeProcessEpoch()` and some other calculations a lot since we don't have to deal with BigInt() there

twoeths:
**Motivation**

+ BigInt is not necessary for effective balance
+ This should avoid a lot of params being configured as BigInt
+ forkchoice `computeDeltas()` wants to use number for its calculation instead of BigInts

waiting for https://github.com/ChainSafe/lodestar/pull/2973

**Description**

+ Change ssz type of Validator to use Number64 for `effectiveBalance`
+ In `beforeEpochProcess()`, we track total balance as multiple of `EFFECTIVE_BALANCE_INCREMENT` then when processing rewards and penalties we divide them to `EFFECTIVE_BALANCE_INCREMENT`, so we should track them "by increment" initially, in eth. It means we can use number there too.

Closes #2981


packages/lodestar/test/utils/state.ts
```diff
@@ -83,7 +83,8 @@ export function generateState(
     eth1DataVotes: ([] as phase0.Eth1Data[]) as List<phase0.Eth1Data>,
     eth1DepositIndex: 0,
     validators: validators as List<phase0.Validator>,
-    balances: Array.from({length: 4}, () => MAX_EFFECTIVE_BALANCE) as List<Gwei>,
+    // TODO: fix once effective balance becomes number
+    balances: Array.from({length: 4}, () => Number(MAX_EFFECTIVE_BALANCE)) as List<number>,
```
dapplion: TODO

packages/types/src/primitive/sszTypes.ts
```diff
@@ -17,7 +18,7 @@ export const Bytes96 = new ByteVectorType({length: 96});
 export const Uint8 = byteType;
 export const Uint16 = new NumberUintType({byteLength: 2});
 export const Uint32 = number32Type;
-export const Number64 = new NumberUintType({byteLength: 8});
+export const Number64 = new Number64UintType();
```
dapplion: Should we rename to `Number64Uint` to prevent issues in the future if we need a true `Number64`?

twoeths: yeah, maybe Uint64 since we have Uint8, Uint16, Uint32...

dapplion (approved):

dapplion:
Awesome work on the PR thanks for the push on this effort! Looking forward to see all the improvements compounded

PR #3046, opened 2021-08-31
Drop flat validators cache

dapplion:
**Motivation**

A Lodestar beacon node uses too much memory compared to what is strictly necessary. See this issue for the analysis https://github.com/ChainSafe/lodestar/issues/2885

We must:
- Not duplicate expensive data = drop flat caches where possible
- Represent the validators registry in a memory efficient manner

**Description**

- Drop flat validators array from CacheBeaconState. In normal network conditions it will not affect performance. Only when the validator registry drastically changes it will slow the node a bit. See performance tests for actual numbers
- Use new SSZ type ContainerLeafNodeStructType to represent the leafs of the state.validators tree as regular JS values. Much less memory, fast reads but slow hashing. The tradeoff is worth it it normal network conditions.

**Performance cost**

Surprisingly it has very little!
- Hashing many validator objects is a bit more expensive ~ x2, but it a rare operation. For example, it increases epoch processing time by 8ms if all block are fulled with deposits.
- getPubkeys (for duties API) is more expensive, extra 0.5ms when requesting duties for 1000 validators
- beforeProcessingEpoch takes extra 30ms to traverse the tree and grab all the root nodes.

**Memory gains**

current master
```
getPubkeys()         rss +106.91 MB       heapTotal +41.41 MB        heapUsed +39.16 MB        external +11.47 MB        arrayBuffers +11.47 MB       
.defaultValue()      rss +14.16 MB        heapTotal +14 MB           heapUsed +15.1 MB         external -259.28 KB       arrayBuffers -259.28 KB      
build raw state      rss +97.44 MB        heapTotal +97.32 MB        heapUsed +103.35 MB       external  0 B             arrayBuffers  0 B            
addPendingAtt        rss +138.94 MB       heapTotal +144 MB          heapUsed +127.61 MB       external  0 B             arrayBuffers  0 B            
toTreeBacked         rss +457.91 MB       heapTotal +450.48 MB       heapUsed +461.4 MB        external -13.3 KB         arrayBuffers -12.69 KB       
CachedBeaconState    rss +814.43 MB       heapTotal +749.26 MB       heapUsed +746.37 MB       external +36.39 KB        arrayBuffers +36.39 KB  
```

this branch
```
getPubkeys()         rss +96.79 MB        heapTotal +33.91 MB        heapUsed +40.16 MB        external +11.44 MB        arrayBuffers +11.44 MB       
[...]

PR #3049, opened 2021-08-31
Add effectiveBalances cache

dapplion:
**Motivation**

To help speed up the PR https://github.com/ChainSafe/lodestar/pull/3046 adds and effectiveBalances cache that allows to not rely on the validators registry for some expensive ops

**Description**

Attach effectiveBalances cache as a `MutableVector<number>` to the epochCtx and use it in all places that's available.


packages/beacon-state-transition/src/util/balance.ts
```diff
@@ -25,17 +24,6 @@ export function getTotalBalance(state: allForks.BeaconState, indices: ValidatorI
...
 }
 
-/**
- * Call this function with care since it has to loop through validators which is expensive.
- * Return the combined effective balance of the active validators.
- * Note: `getTotalBalance` returns `EFFECTIVE_BALANCE_INCREMENT` Gwei minimum to avoid divisions by zero.
- *
- * SLOW CODE - 🐢
- */
-export function getTotalActiveBalance(state: allForks.BeaconState): Gwei {
```
twoeths, line 35: 👍

dapplion:
:warning: This very optimistic performance test uncovered a bug vector that we don't have tests for

> phase0 processEffectiveBalanceUpdates - 250000 worstcase 0.5 | 9.4982 ms/op | 2.6382  s/op | 0.00
-- | -- | -- | --

epochCtx.effectiveBalances was not being .clone() so the performance test run once doing changes, and did none on subsequent runs. No spec tests nor the simulated tests picked up this bug. @tuyennhv @wemeetagain how could we cover this scenario in a test?

packages/beacon-state-transition/src/allForks/util/epochContext.ts
```diff
@@ -137,7 +138,11 @@ export function createEpochContext(
         exitQueueChurn += 1;
       }
     }
+
+    // TODO: Should have 0 for not active validators to be re-usable in ForkChoice
```
twoeths, line 142: I think we can just pass the whole effectiveBalances to forkchoice, please take a look at https://github.com/ChainSafe/lodestar/blob/7afe0aab0eb09d2ff4c6a5b98b4c9e33693b8e1b/packages/fork-choice/src/protoArray/computeDeltas.ts#L42

for inactive validator, currentRoot should be nextRoot and old balance should be the same to new balance so delta should be 0 for those inactive validator indices anyway.

@wemeetagain can you confirm again?

twoeths:
> ⚠️ This very optimistic performance test uncovered a bug vector that we don't have tests for
> 
> > phase0 processEffectiveBalanceUpdates - 250000 worstcase 0.5 | 9.4982 ms/op | 2.6382  s/op | 0.00
> > -- | -- | -- | --
> 
> epochCtx.effectiveBalances was not being .clone() so the performance test run once doing changes, and did none on subsequent runs. No spec tests nor the simulated tests picked up this bug. @tuyennhv @wemeetagain how could we cover this scenario in a test?

I have no idea, I suppose spec test or our sim test would fail without the clone. And probably we should not trust any performance gain that's too good to be true without a good reason for it? :D

twoeths (approved):
looks good to me, I saw some performance tests takes +30% time to run but don't see any code issue may cause this, I run again in my local environment and don't see this issue.

packages/beacon-state-transition/src/allForks/util/epochContext.ts
dapplion: @tuyennhv I think it's unsafe to pass to the forkchoice since in case of heavy forking there may still be attestations around for attackers. We should ping the researchers before going that way

twoeths: @dapplion we could loop through the cached effectiveBalances persistent-ts 1 time anyway, will create a PR for that

dapplion:
> looks good to me, I saw some performance tests takes +30% time to run but don't see any code issue may cause this, I run again in my local environment and don't see this issue.

Perf runs in CI have high variability, +30% -30% is normal

Issue #2747, opened 2021-06-24
Improve forkchoice updateHead


twoeths:
**Describe the bug**

This is from a prater node with no validator
<img width="1275" alt="Screen Shot 2021-06-24 at 09 00 35" src="https://user-images.githubusercontent.com/10568965/123191016-c47de780-d4ca-11eb-9205-86772410adf3.png">
[0624_prater_forkchoice_updateHead.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/6705773/0624_prater_forkchoice_updateHead.cpuprofile.zip)

a normal state transition takes 227.6m, 212.2 ms is for `updateHead` 

We need to find a way to improve `computeDeltas` in forkchoice implementation.

**Expected behavior**

Not sure what the expectation is but I think forkchoice is designed to be lightweight, taking 90% of time per normal state transition for forkchoice is too much.

twoeths:
I think one of the cost is the slowness of BigInt calculation in `computeDeltas`. One idea is to track `effectiveBalance` as number instead of BigInt since `MAX_EFFECTIVE_BALANCE=32000000000` which is still in the range of number.

We need to write benchmark test for any solutions we want to go.

dapplion:
Related https://github.com/ChainSafe/lodestar/issues/2455

g11tech:
@dapplion assign pls

dapplion:
![Screenshot from 2021-07-06 18-54-40](https://user-images.githubusercontent.com/35266934/124638810-b1382800-de8b-11eb-9563-f6217a01d19b.png)
![Screenshot from 2021-07-06 18-54-50](https://user-images.githubusercontent.com/35266934/124638814-b301eb80-de8b-11eb-9e89-6ed34bcbc939.png)

After implementing head caching fork-choice runs take 0.5% of CPU time in a not overloaded node (no subnets). Worth to look into it but not an high priority.

Also note that metrics reads match what Tuyen got in the profile.

twoeths:
As noted in #2332, effectiveBalance is always multiple of  `EFFECTIVE_BALANCE_INCREMENT` so ideally we want to track the weight as `32, 31, 30` instead of `32000000000, 31000000000, 30000000000`

dapplion:
Current metrics the fork-choice cost is very low :+1:

dapplion:
Fork-choice updateHead cost is about ~1% of total CPU time currently. I think it can still be improved, but it's low priority

[...]

PR #3065, opened 2021-09-03
Improve getEffectiveBalances()

twoeths:
**Motivation**

Looping through persistent-ts is faster than persistent-merkle-tree

**Description**

+ Also we can use `epochCtx.currentShuffling.activeIndices` to know if a validator is active for currentShuffling epoch or not
+ The benchmark showed that it's a 5x improvement

### DIFF
--- a/packages/beacon-state-transition/src/util/balance.ts
+++ b/packages/beacon-state-transition/src/util/balance.ts
@@ -51,21 +51,26 @@ export function decreaseBalance(
 
 /**
  * This method is used to get justified balances from a justified state.
+ * This is consumed by forkchoice which based on delta so we return "by increment" (in ether) value,
+ * ie [30, 31, 32] instead of [30e9, 31e9, 32e9]
  */
 export function getEffectiveBalances(justifiedState: CachedBeaconState<allForks.BeaconState>): number[] {
   const {activeIndices} = justifiedState.currentShuffling;
   // 5x faster than using readonlyValuesListOfLeafNodeStruct
-  const effectiveBalances = justifiedState.effectiveBalances.toArray();
+  const count = justifiedState.validators.length;
+  const {effectiveBalances} = justifiedState;
+  const effectiveBalancesArr = new Array<number>(count);
   let j = 0;
-  for (let i = 0; i < effectiveBalances.length; i++) {
-    // same logic to checking activeIndices.includes(i) since activeIndices is sorted
-    if (i !== activeIndices[j]) {
-      // inactive validator
-      effectiveBalances[i] = 0;
-    } else {
-      // active validator, keep effective balance as is
+  for (let i = 0; i < count; i++) {
+    if (i === activeIndices[j]) {
+      // active validator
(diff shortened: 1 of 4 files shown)


packages/beacon-state-transition/src/util/balance.ts
```diff
@@ -52,18 +50,22 @@ export function decreaseBalance(
...
-  const effectiveBalances: number[] = [];
-  const validators = readonlyValuesListOfLeafNodeStruct(justifiedState.validators);
-  for (let i = 0, len = validators.length; i < len; i++) {
-    const validator = validators[i];
-    effectiveBalances.push(isActiveValidator(validator, justifiedEpoch) ? validator.effectiveBalance : 0);
+  const {activeIndices} = justifiedState.currentShuffling;
+  // 5x faster than using readonlyValuesListOfLeafNodeStruct
+  const effectiveBalances = justifiedState.effectiveBalances.toArray();
+  let j = 0;
+  for (let i = 0; i < effectiveBalances.length; i++) {
```
dapplion: If you look at the benchmarks

```
MutableVector 250000 iterate all - toArray() + loop | 8.0552 ms/op | 8.9640 ms/op | 0.90
MutableVector 250000 iterate all - get(i) | 3.4577 ms/op | 3.8777 ms/op | 0.89
```
it's actually cheaper to iterate once with individual get(i) ops. Try creating an empty array and push after each get(i)

twoeths: but we need the number[] in the end, it's convenient to create this way and just update to 0 if it's not an active validator

twoeths: also this needs to address https://github.com/ChainSafe/lodestar/issues/2747#issuecomment-879568043

twoeths: @dapplion in the end I switched to use `persistentTS.get(i)`

packages/beacon-state-transition/test/perf/util/balance.test.ts
```diff
@@ -0,0 +1,18 @@
+import {itBench} from "@dapplion/benchmark";
+import {generatePerfTestCachedStatePhase0, perfStateId} from "../util";
+import {State} from "../types";
+import {getEffectiveBalances} from "../../../src/util";
+
+describe("getEffectiveBalances", () => {
+  itBench<State, State>({
+    id: `getEffectiveBalances - ${perfStateId}`,
```
dapplion, line 8: Use `runsFactor` option to scale down the 100 for loop iterations
https://github.com/dapplion/benchmark/blob/8f2fffda28262e08a446d8a8284d1d2235dc479f/src/types.ts#L40

twoeths: @dapplion I use `runsFactor: 100` 8 lines below, is that correct?

dapplion: Oh I'm blind, didn't see it hahaha. Yes it's correct :+1:

dapplion (approved):

Issue #3066, opened 2021-09-03
Invalid state root


twoeths:
**Describe the bug**

Got this on contabo-0 running Prater
```
Sep-03 03:01:58.707 [CHAIN]           error: Block error slot=1177408 code=BLOCK_ERROR_BEACON_CHAIN_ERROR, error={"message":"Invalid state root at slot 1177408, expected=0xc7db41ea510dae0fbf32caae531c9e29e95b0318b4c4c6ff192654146f295a66, actual=0x17eb9ac943f2a3aa5b2a53640cd7c48088ef086eae7f825cda97dd9d6bd73ffb"}
Error: Invalid state root at slot 1177408, expected=0xc7db41ea510dae0fbf32caae531c9e29e95b0318b4c4c6ff192654146f295a66, actual=0x17eb9ac943f2a3aa5b2a53640cd7c48088ef086eae7f825cda97dd9d6bd73ffb
    at Object.stateTransition (/usr/app/packages/beacon-state-transition/src/allForks/stateTransition.ts:67:13)
    at Object.runStateTransition (/usr/app/packages/lodestar/src/chain/blocks/stateTransition.ts:83:30)
    at processBlocksInEpoch (/usr/app/packages/lodestar/src/chain/blocks/process.ts:108:24)
    at runMicrotasks (<anonymous>)
    at runNextTicks (internal/process/task_queues.js:60:5)
    at processTimers (internal/timers.js:497:9)
    at Object.processBlock (/usr/app/packages/lodestar/src/chain/blocks/process.ts:42:3)
    at processBlockJob (/usr/app/packages/lodestar/src/chain/blocks/processor.ts:66:5)
    at Timeout.JobItemQueue.runJob [as _onTimeout] (/usr/app/packages/lodestar/src/util/queue/itemQueue.ts:85:22)
Error: BLOCK_ERROR_BEACON_CHAIN_ERROR
    at processBlocksInEpoch (/usr/app/packages/lodestar/src/chain/blocks/process.ts:149:11)
    at runMicrotasks (<anonymous>)
    at runNextTicks (internal/process/task_queues.js:60:5)
    at processTimers (internal/timers.js:497:9)
    at Object.processBlock (/usr/app/packages/lodestar/src/chain/blocks/process.ts:42:3)
    at processBlockJob (/usr/app/packages/lodestar/src/chain/blocks/processor.ts:66:5)
[...]

PR #3067, opened 2021-09-03
Persist invalid ssz objects

twoeths:
**Motivation**

+ When "Invalid state root" issue happen inside our state transition, there is no way to investigate. We should be able to persist the preState/block/postState and investigate
+ Teku always persist error items, it's really convenient for us to do the same so that does not take time to reproduce/investigate the issue

**Description**

+ New "archiveItem" function in our util, call it when error in our api/gossip/state transition. For invalid gossip message, we only persist if its a "REJECT" action
+ Since `archiveItem` stays in util so that we can call it either from lodestar or beacon-state-transition, I make LODESTAR_ARCHIVE_DIR environment variable for the parent dir, set it with new "archiveDir" option
+ Although it takes time to serialize items, we don't do that frequently so hopefully it does not have performance issue
+ Persist by date, something like `/Users/tuyennguyen/.local/share/lodestar/prater/archive/2021-09-03/state_invalid_weak_subjectivity_1630662789679.ssz`

part of #3066

### DIFF
--- a/packages/cli/src/cmds/beacon/handler.ts
+++ b/packages/cli/src/cmds/beacon/handler.ts
@@ -31,6 +31,7 @@ export async function beaconHandler(args: IBeaconArgs & IGlobalArgs): Promise<vo
   const beaconPaths = getBeaconPaths(args);
   // TODO: Rename db.name to db.path or db.location
   beaconNodeOptions.set({db: {name: beaconPaths.dbDir}});
+  beaconNodeOptions.set({chain: {persistInvalidSszObjectsDir: beaconPaths.persistInvalidSszObjectsDir}});
   // Add metrics metadata to show versioning + network info in Prometheus + Grafana
   beaconNodeOptions.set({metrics: {metadata: {...gitData, version, network: args.network}}});
 
--- a/packages/cli/src/cmds/beacon/options.ts
+++ b/packages/cli/src/cmds/beacon/options.ts
@@ -102,6 +102,13 @@ export const beaconPathsOptions: ICliCommandOptions<IBeaconPaths> = {
     type: "string",
   },
 
+  persistInvalidSszObjectsDir: {
+    description: "Directory to persist invalid ssz objects",
+    defaultDescription: defaultBeaconPaths.persistInvalidSszObjectsDir,
+    hidden: true,
+    type: "string",
+  },
+
   configFile: {
     description: "Beacon node configuration file path",
     defaultDescription: defaultBeaconPaths.configFile,
--- a/packages/lodestar/src/api/impl/beacon/pool/index.ts
+++ b/packages/lodestar/src/api/impl/beacon/pool/index.ts
@@ -1,6 +1,6 @@
 // eslint-disable-next-line no-restricted-imports
 import {Api as IBeaconPoolApi} from "@chainsafe/lodestar-api/lib/routes/beacon/pool";
-import {Epoch} from "@chainsafe/lodestar-types";
+import {Epoch, ssz} from "@chainsafe/lodestar-types";
 import {allForks} from "@chainsafe/lodestar-beacon-state-transition";
 import {SYNC_COMMITTEE_SIZE, SYNC_COMMITTEE_SUBNET_COUNT} from "@chainsafe/lodestar-params";
 import {validateGossipAttestation} from "../../../../chain/validation";
@@ -10,6 +10,8 @@ import {validateGossipVoluntaryExit} from "../../../../chain/validation/voluntar
 import {validateSyncCommitteeSigOnly} from "../../../../chain/validation/syncCommittee";
 import {ApiModules} from "../../types";
 import {OpSource} from "../../../../metrics/validatorMonitor";
+import {toHexString} from "@chainsafe/ssz";
+import {AttestationError, GossipAction, SyncCommitteeError} from "../../../../chain/errors";
 
 export function getBeaconPoolApi({
   chain,
@@ -64,6 +66,14 @@ export function getBeaconPoolApi({
               {slot: attestation.data.slot, index: attestation.data.index},
               e
             );
+            if (e instanceof AttestationError && e.action === GossipAction.REJECT) {
+              const archivedPath = chain.persistInvalidSszObject(
+                "attestation",
+                ssz.phase0.Attestation.serialize(attestation),
+                toHexString(ssz.phase0.Attestation.hashTreeRoot(attestation))
+              );
+              logger.debug("Submitted invalid attestation was written to", archivedPath);
+            }
           }
         })
       );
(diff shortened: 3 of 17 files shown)


packages/lodestar/src/api/impl/beacon/pool/index.ts
```diff
@@ -139,9 +151,18 @@ export function getBeaconPoolApi({
...
+                "syncCommittee",
+                ssz.altair.SyncCommitteeMessage.serialize(signature),
+                toHexString(ssz.altair.SyncCommitteeMessage.hashTreeRoot(signature))
+              );
+            }
+
             logger.error(
               `Error on submitPoolSyncCommitteeSignatures [${i}]`,
-              {slot: signature.slot, validatorIndex: signature.validatorIndex},
+              {slot: signature.slot, validatorIndex: signature.validatorIndex, archivedPath},
```
dapplion: Don't add the archivedPath to the error log itself, just log it afterwards

```ts
logger.error(
              `Error on submitPoolSyncCommitteeSignatures [${i}]`,
              {slot: signature.slot, validatorIndex: signature.validatorIndex},
              e
            );

// then
if (e instanceof SyncCommitteeError && e.action === GossipAction.REJECT) {
              const archivedPath = archiveItem(
                "syncCommittee",
                ssz.altair.SyncCommitteeMessage.serialize(signature),
                toHexString(ssz.altair.SyncCommitteeMessage.hashTreeRoot(signature))
              );
              // Log "Invalid SSZ object written to ${archivedPath}"
            }
```
Then you can abstract the archive logic way better, and don't need `let` vars

packages/cli/src/networks/index.ts
```diff
@@ -112,11 +113,23 @@ export async function fetchWeakSubjectivityState(
   config: IChainForkConfig,
   url: string
 ): Promise<TreeBacked<allForks.BeaconState>> {
+  let stateBytes: Buffer | undefined;
   try {
     const response = await got(url, {headers: {accept: "application/octet-stream"}});
-    const stateBytes = response.rawBody;
+    stateBytes = response.rawBody;
```
dapplion: Refactor to avoid let variables

```ts
const response = await got(url, {headers: {accept: "application/octet-stream"}}).catch(e => {
      throw new Error("Unable to fetch weak subjectivity state: " + (e as Error).message);
})
try {
      return getStateTypeFromBytes(config, stateBytes).createTreeBackedFromBytes(stateBytes);
} catch (e) {
  // archive test and rethrow error
}
```

packages/cli/src/cmds/beacon/options.ts
```diff
@@ -102,6 +102,13 @@ export const beaconPathsOptions: ICliCommandOptions<IBeaconPaths> = {
     type: "string",
   },
 
+  archiveDir: {
```
dapplion: This name is way to vague, only debuggers should use this. The name should imply it's for debugging
```bash
persistInvalidSszObjectsDir
# or
persistInvalidSszObjectsDebugDir
```

packages/utils/src/file.ts
```diff
@@ -0,0 +1,29 @@
...
+  | "contributionAndProof";
+
+/**
+ * Persist bad items to archive dir, for example invalid state, attestations etc.
+ */
+export function archiveItem(type: ArchiveType, bytes: Uint8Array, suffix = ""): string {
+  const now = new Date();
+  // yyyy-MM-dd
+  const date = now.toISOString().split("T")[0];
+  const parentDir = process.env.LODESTAR_ARCHIVE_DIR;
```
dapplion: This is problematic, utils must be able to run in the browser and this is NodeJS specific code. I would prefer to have this code in packages/lodestar only unless there's a strong reason to put it here

packages/lodestar/src/network/gossip/handlers/index.ts
```diff
@@ -75,7 +82,7 @@ export function getGossipHandlers(modules: ValidatorFnsModules): GossipHandlers
         try {
           chain.receiveBlock(signedBlock);
         } catch (e) {
-          logger.error("Error receiving block", {}, e);
+          logger.error("Error receiving block ", {}, e);
```
dapplion: Typo

packages/lodestar/src/network/gossip/handlers/index.ts
```diff
@@ -18,6 +24,7 @@ import {
   validateGossipVoluntaryExit,
 } from "../../../chain/validation";
 import {INetwork} from "../../interface";
+import {toHexString} from "@chainsafe/ssz";
```
dapplion: Sort dependencies by first external deps, then `@chainsafe` deps, then local path imports

packages/cli/src/cmds/beacon/handler.ts
```diff
@@ -31,6 +31,8 @@ export async function beaconHandler(args: IBeaconArgs & IGlobalArgs): Promise<vo
   const beaconPaths = getBeaconPaths(args);
   // TODO: Rename db.name to db.path or db.location
   beaconNodeOptions.set({db: {name: beaconPaths.dbDir}});
+  // we need to store to env variable so that lodestar-utils can read this
+  process.env.LODESTAR_ARCHIVE_DIR = beaconPaths.archiveDir;
```
dapplion: This is very ugly, please just pass a regular beaconNode options even if it requires a lot of plumbing to get the options to the network class

twoeths:
**Update on Sep 07**
+ New chain options `persistInvalidSszObjects` (default as true for easier debug, we may turn it to false in the future if our node is very stable) and `persistInvalidSszObjectsDir` (default as `${root_dir}/invalidSszObjects`)
+ New method `chain.persistInvalidSszObject`, store by date
+ Do not verify state root in beacon-state-transition so that it's convenient to call `chain.persistInvalidSszObject` in lodestar only, verify state root at lodestar side, new `BlockErrorCode.INVALID_STATE_ROOT` and handle it in errorHandler

the folder structure is like

```
invalidSszObjects
 |-2021-09-07
 | |-state_2012959_invalid_state_root_preState_0x8bad721001f95044215c769ea57860ad90d739e258e526b71e37c3b1bde07925_1630979595245.ssz
 | |-signedBlock_2012960_invalid_state_root_0x8bad721001f95044215c769ea57860ad90d739e258e526b71e37c3b1bde07925_1630979593992.ssz
 | |-state_2012959_invalid_state_root_preState_0x8bad721001f95044215c769ea57860ad90d739e258e526b71e37c3b1bde07925_1630979664056.ssz
 | |-state_2012960_invalid_state_root_postState_0x8bad721001f95044215c769ea57860ad90d739e258e526b71e37c3b1bde07925_1630979665637.ssz
 | |-signedBlock_2012960_invalid_state_root_0x8bad721001f95044215c769ea57860ad90d739e258e526b71e37c3b1bde07925_1630979663326.ssz
 | |-state_2012960_invalid_state_root_postState_0x8bad721001f95044215c769ea57860ad90d739e258e526b71e37c3b1bde07925_1630979597718.ssz
```

dapplion (approved):
Looks great thanks!

PR #3068, opened 2021-09-03
Remove dev databases if chain starts from scratch

dapplion:
**Motivation**

When running dev command again, the db must be whiped out or bad things happens:
- slashing protection db invalidates everything

**Description**

Remove dev databases if no genesisTime option is passed


q9f:
:+1: I was also looking into that. Also, the `--reset` flag for `--dev` is broken too.

wemeetagain (approved):

PR #3073, opened 2021-09-03
no default read of the configFile and paramsFile, persistence of these files removed

### LINKED ISSUE #2282: Review if `node-prometheus-gc-stats` is needed
See https://github.com/siimon/prom-client/pull/361#discussion_r400430717

From https://github.com/SimenB/node-prometheus-gc-stats/issues/72

### LINKED ISSUE #2983: Persisted config overwrites new Pyrmont fork config
**Describe the bug**

Our Pyrmont node did not apply the new config merged in https://github.com/ChainSafe/lodestar/pull/2975

The cause is that the cli beacon cmd had persisted a `config.yaml` in the rootDir with these values

```
'PRESET_BASE': 'mainnet'
'MIN_GENESIS_ACTIVE_VALIDATOR_COUNT': '16384'
'MIN_GENESIS_TIME': '1605700800'
'GENESIS_FORK_VERSION': '0x00002009'
'GENESIS_DELAY': '432000'
'ALTAIR_FORK_VERSION': '0x01000000'
'ALTAIR_FORK_EPOCH': '18446744073709551615'
'MERGE_FORK_VERSION': '0x02000000'
'MERGE_FORK_EPOCH': '18446744073709551615'
'SHARDING_FORK_VERSION': '0x03000000'
'SHARDING_FORK_EPOCH': '18446744073709551615'
'TRANSITION_TOTAL_DIFFICULTY': '4294967296'
'SECONDS_PER_SLOT': '12'
'SECONDS_PER_ETH1_BLOCK': '14'
'MIN_VALIDATOR_WITHDRAWABILITY_DELAY': '256'
'SHARD_COMMITTEE_PERIOD': '256'
'ETH1_FOLLOW_DISTANCE': '2048'
'INACTIVITY_SCORE_BIAS': '4'
'INACTIVITY_SCORE_RECOVERY_RATE': '16'
'EJECTION_BALANCE': '16000000000'
'MIN_PER_EPOCH_CHURN_LIMIT': '4'
'CHURN_LIMIT_QUOTIENT': '65536'
'DEPOSIT_CHAIN_ID': '5'
'DEPOSIT_NETWORK_ID': '5'
'DEPOSIT_CONTRACT_ADDRESS': '0x8c5fecdc472e27bc447696f431e425d02dd46a8c'
```

When the command starts again, reads this config and overwrites the new default values with those. Note that it includes `ALTAIR_FORK_EPOCH`.

**Expected behavior**

- If the user hasn't explicitly customized a config we **should not** persist the defaults. It will prevent our nodes from forking to altair in existing networks.
- I'm not even sure that we should persist the config file at all. We have to expose an option for users to provide a custom config file, but what's the point storing it if the user doesn't need to? Note that this routine comes back from when we used a two step process with init then run.
- Add more logging in the beacon node about current fork:
  - On start, the cli should log the current fork
  - When a fork transition happens, the chain should log that
  - Metrics must include a text field with the current fork name

g11tech:
**Motivation**
PR removes default reading of the configFile and paramsFile files without explicit flags. also when flags provided but files don't exists logs error onto the terminal to bring it to the user attentions as follows:

image

**Description**

partially closes #2983


dapplion (commented):
Thanks for the PR! If the user provided a flag for a file and the file is not found, the CLI must error and stop, not show an error and continue.

Also please check CI, type errors

g11tech:
> Thanks for the PR! If the user provided a flag for a file and the file is not found, the CLI must error and stop, not show an error and continue.
> 
> Also please check CI, type errors

@dapplion  understood for cli to throw error.
e2e test cases that are failing are where we init (`init.test.ts`), and then test the beacon config file options that are written. these test cases won't make sense in the new scheme. should i just remove them for now and later figure out a way to test the executed configuration?

g11tech:
or we dump the options like before (in a separate beaon.config.dump.json, network.params.dump.yml , just to inspect/run test cases against them)

dapplion (dismissed):
Looks great thanks!

PR #3086, opened 2021-09-06
config api interoperability with lighthouse

g11tech:
**Motivation**
The goal is to make validator interoperable with lighthouse
The commit is trying to match case of data received from lighthouse beacon node, issue is our beacon servers snake case for config data using jsonopts while lighthouse serves constant case for the spec config data. This PR demos one way to do that.

DEPENDENCY: ssz PR to extend expected case: https://github.com/ChainSafe/ssz/pull/184 to be merged in ssz with ssz released and package version updated.

even after matching case, got this missing data field error in lighthouse data:

image

which basically implies that lighthouse latest stable branch doesn't has the  field. this raises some questions of tracking the version of api.

Also the expectedCase filling of the container data doesn't seem to be the correct approach, because of container nesting like {data:configObj} where data is snake, and configObj is constant case. Ideally if there has to be any case specification of container fields it has to be with the container structure itself (along with fields) and then we don't need to pass any expected case opts.
This needs more discussion and debate.

**Description**

Closes #issue_number

### DIFF
--- a/packages/api/src/client/utils/client.ts
+++ b/packages/api/src/client/utils/client.ts
@@ -72,7 +72,7 @@ export function generateGenericJsonClient<
       const res = await fetchFn.json<Json>(fetchOptsSerializer(...args));
       if (returnType) {
         // eslint-disable-next-line @typescript-eslint/no-unsafe-return
-        return returnType.fromJson(res, jsonOpts) as ReturnType<Api[keyof Api]>;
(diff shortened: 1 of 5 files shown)


g11tech:
casing issue has been resolved, being tracked in the master issue : https://github.com/ChainSafe/lodestar/issues/3368

PR #3092, opened 2021-09-07
Move archiver to BeaconChain

dapplion:
**Motivation**

While working on a memory leak issue I caused another leak because the pruning caller for state caches was outside of the BeaconChain class.

Moving the archive blocks and states into the BeaconChain makes more sense and is more intuitive as now 100% of that functionality is handled by the same class without external dependencies.

**Description**

- Move archive blocks and states logic into the BeaconChain class, under an Archiver class.
- Convert archive blocks into a function since it does not require state and was instantiated every time.


twoeths (approved):
looks good to me!

PR #3096, opened 2021-09-09
Avoid getTotalBalance()

twoeths:
**Motivation**
+ `getTotalBalance` is called per finalized checkpoint, basically it means per epoch
+ `getTotalBalance` is very slow, it's >3s in the below profiler in my local environment (and usually the server is 2x slower than my environment)
Screen Shot 2021-09-09 at 15 52 20

**Description**

+ `totalActiveBalanceByIncrement` is available for epochProcess and altair, we should calculate, use it and cache it in  EpochContext for both phase0 and altair


dapplion:
Oh nice pick, but why is the node computing the weak subjectivity period every epoch? Why does it need that info?

EDIT: Oh, for the lightclient initer

dapplion (approved):

wemeetagain (approved):

PR #3102, opened 2021-09-09
Improve attestation production

dapplion:
**Motivation**

Ensure correctness in more corner cases, where the vc and beacon node may not agree on time.

**Description**

- Clean up and improve attestation production logic. Based on https://github.com/sigp/lighthouse/blob/b0ac3464ca5fb1e9d75060b56c83bfaf990a3d25/beacon_node/beacon_chain/src/beacon_chain.rs#L1264


packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -165,9 +166,46 @@ export function getValidatorApi({
...
+      const targetRoot =
+        targetSlot >= headSlot
+          ? // If the state is earlier than the target slot then the target *must* be the head block root.
+            headBlockRoot
+          : getBlockRootAtSlot(headState, targetSlot);
+
+      // To get the correct source we must get a state in the same epoch as the attestation's epoch.
+      // An epoch transition may change state.currentJustifiedCheckpoint
+      const attEpochState =
+        attEpoch === headEpoch
```
twoeths: I think it's better to use ">=" here?

twoeths: ah no, "<="

dapplion: Right good point

twoeths (approved):

PR #3104, opened 2021-09-09
Query discv5 peers on demand

wemeetagain:
**Motivation**

See #3041 

**Description**

Instead of just continuously performing discv5 lookups, also perform a lookup when the peer discovery module determines that peers need to be discovered.


wemeetagain:
from @dapplion 
 we relied on discv5 always searching to ensure that we can always find subnet peers

we need to also ensure there's a way to find subnet peers when we need

dapplion (approved):

PR #3106, opened 2021-09-09
Cache head state in BeaconChain instance

dapplion:
**Motivation**

As a previous step to having a state cache with WeakRefs, the head state must be strongly held somewhere. This PR implements and approach inspired by Lighthouse, where the canonical head state is kept in the BeaconChain class

**Description**

- Keep head state in BeaconChain class
- Required changing the API of all processors and regen to get be able to call `chain.updateHead()` inside the state transition fn runner


dapplion:
Will do again from scratch after latest merges

PR #3107, opened 2021-09-10
Review gossip block validation

dapplion:
**Motivation**

Apply the same revision and style as https://github.com/ChainSafe/lodestar/pull/2760

**Description**

- Inline spec comments to ensure we cover all validation cases.
- Use a seen cache in memory instead of doing db reads
- Port useful optimizations from Lighthouse gossip validation
- Update test with helpful comments and updated sorted conditions

### DIFF
--- a/packages/lodestar/src/chain/seenCache/seenBlockProposers.ts
+++ b/packages/lodestar/src/chain/seenCache/seenBlockProposers.ts
@@ -2,7 +2,7 @@ import {Epoch, Slot, ValidatorIndex} from "@chainsafe/lodestar-types";
 import {MapDef} from "../../util/map";
 
 /**
- * Keeps a cache to filter unaggregated attestations from the same validator in the same epoch.
+ * Keeps a cache to filter block proposals from the same validator in the same slot.
  *
  * This cache is not bounded and for extremely long periods of non-finality it can grow a lot. However it's practically
  * limited by the possible shufflings in those epochs, and the stored data is very cheap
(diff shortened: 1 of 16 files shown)


packages/lodestar/src/chain/seenCache/seenBlockProposers.ts
```diff
@@ -0,0 +1,34 @@
+import {Epoch, Slot, ValidatorIndex} from "@chainsafe/lodestar-types";
+import {MapDef} from "../../util/map";
+
+/**
+ * Keeps a cache to filter unaggregated attestations from the same validator in the same epoch.
```
twoeths: need to fix this description?

dapplion: Fixed

packages/lodestar/src/chain/validation/block.ts
```diff
@@ -26,56 +40,85 @@ export async function validateGossipBlock(
...
+  if (parentBlock === null) {
+    // If fork choice does *not* consider the parent to be a descendant of the finalized block,
+    // then there are two more cases:
+    //
+    // 1. We have the parent stored in our database. Because fork-choice has confirmed the
+    //    parent is *not* in our post-finalization DAG, all other blocks must be either
+    //    pre-finalization or conflicting with finalization.
+    // 2. The parent is unknown to us, we probably want to download it since it might actually
+    //    descend from the finalized root.
+    if ((await db.block.get(parentRoot as Uint8Array)) !== null) {
```
twoeths: this never happen with the current implementation. When `archiveBlocks` runs, if it found any blocks conflicting to the finalized block it'll delete it from db

Just afraid we query db for nothing here.

dapplion: That's a good point, and this check won't save us from blocks from unknown forks if they are two children deep from our latest known forked block. I'll remove

twoeths (approved):

PR #3119, opened 2021-09-12
rearrangement of regen fn stats panel

g11tech:
**Motivation**

**Description**

Rearrangement of the panels for regenFn stats from [][][] format to [][] 

panels:
image

image

and so on... for all 4 regen Fns

Closes #issue_number


dapplion (approved):
Thanks!

Issue #3084, opened 2021-09-06
Incorrect attestations submitted by lodestar validator in prater


twoeths:
**Describe the bug**

Validator consistently submit incorrect source/target checkpoints and head block in prater after altair launched

<img width="1566" alt="Screen Shot 2021-09-06 at 08 15 46" src="https://user-images.githubusercontent.com/10568965/132147628-16ea32a8-5f41-4209-9bb0-5a2729b0fee9.png">

**Expected behavior**

Validator should submit correct attestations most of the time

twoeths:
this is due to the performance of gossip block, we tend to receive gossip block from previous epoch

```
Sep-08 03:06:48.298 [CHAIN]           debug: Block pools:  pendingBlocks=68, currentSlot=1213534
Sep-08 03:06:51.349 [NETWORK]       verbose: Received gossip block slot=1213534, root=0x69ee…50a6
Sep-08 03:06:52.091 [NETWORK]       verbose: Received gossip block slot=1213502, root=0xf2b8…74ad
Sep-08 03:07:01.220 [CHAIN]           debug: Block pools:  pendingBlocks=68, currentSlot=1213535
Sep-08 03:07:03.186 [NETWORK]       verbose: Received gossip block slot=1213503, root=0x6f01…e57e
Sep-08 03:07:03.237 [NETWORK]       verbose: Received gossip block slot=1213535, root=0xdc10…51a8
Sep-08 03:07:12.480 [CHAIN]           debug: Block pools:  pendingBlocks=68, currentSlot=1213536
Sep-08 03:07:15.868 [NETWORK]       verbose: Received gossip block slot=1213504, root=0x80a8…6d7e
Sep-08 03:07:15.954 [NETWORK]       verbose: Received gossip block slot=1213536, root=0xcd73…240e
...
```

+ at slot 1213434, we do receive gossip block 1213434
+ at slot 1213435, we receive `old/wrong` gossip block 1213403
+ at slot 1213436, we do receive gossip block 1213436
+ at slot 1213437, we receive `old/wrong` gossip block 1213417
+ at slot 1213438, we receive `old/wrong` gossip block 1213406
+ at slot 1213539, we do receive both gossip block 1213539 and `old` gossip block 1213507

also pendingBlocks size of 68 means that we have some missing parent blocks

twoeths:
with #3186 we can receive gossip blocks more regularly but some blocks come quite late (>4s) after the slot starts
```
[...]

PR #3126, opened 2021-09-13
Implement gossip topic scoring params

twoeths:
**Motivation**

+ gossip v1.1 has an implementation for scoring params but we only use default params

**Description**

+ go with lighthouse's implementation at https://github.com/sigp/lighthouse/blob/b0ac3464ca5fb1e9d75060b56c83bfaf990a3d25/beacon_node/eth2_libp2p/src/behaviour/gossipsub_scoring_parameters.rs#L83

part of #3084


packages/lodestar/src/network/gossip/scoringParameters.ts
```diff
@@ -0,0 +1,276 @@
...
+  (MAX_IN_MESH_SCORE + MAX_FIRST_MESSAGE_DELIVERIES_SCORE) *
+  (BEACON_BLOCK_WEIGHT +
+    +BEACON_AGGREGATE_PROOF_WEIGHT +
+    beaconAttestationSubnetWeight * ATTESTATION_SUBNET_COUNT +
+    VOLUNTARY_EXIT_WEIGHT +
+    PROPOSER_SLASHING_WEIGHT +
+    ATTESTER_SLASHING_WEIGHT);
+
+/**
+ * The following params is implemented by Lighthouse.
```
dapplion: Can you link to exact code where this is taken from?

packages/lodestar/test/unit/network/gossip/scoringParameters.ts
```diff
@@ -0,0 +1,254 @@
...
+import {expect} from "chai";
+import {stringifyGossipTopic} from "../../../../src/network/gossip/topic";
+import {GossipType} from "../../../../src/network";
+import {TopicScoreParams} from "libp2p-gossipsub/src/score";
+
+/**
+ * Refer to Teku tests at
+ * https://github.com/ConsenSys/teku/blob/e18ab9903442410aa04b590c4cc46734e13d3ffd/networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/config/GossipScoringConfiguratorTest.java#L38
+ */
+export function runGossipScoringParametersTests(): void {
```
twoeths: this test needs to run with mainnet profile so I have to call it through `mainnetTest.ts`, any better ways to do it?

dapplion: Yeah, can you place the tests under `test/unit-mainnet` for consistency with other tests? Then add to package json
```
"test:unit": "LODESTAR_PRESET=mainnet mocha 'test/unit-mainnet/**/*.test.ts'"
```

The minimal preset is injected in this script for reference https://github.com/ChainSafe/lodestar/blob/7beccf0128c94d9e1720d4e844a53035ff4975ea/packages/lodestar/.mocharc.yaml#L5

twoeths:
issue #3132 found during testing this branch, however it does not look like that comes from this branch itself. I can't reproduce it in my local environment or contabo server.

packages/lodestar/src/chain/chain.ts
```diff
@@ -38,6 +38,7 @@ import {ForkDigestContext, IForkDigestContext} from "../util/forkDigestContext";
 import {LightClientIniter} from "./lightClient";
 import {AggregatedAttestationPool} from "./opPools/aggregatedAttestationPool";
 import {Archiver} from "./archiver";
+import {Eth2Context} from ".";
```
dapplion: Import explicitly from module

packages/lodestar/src/network/gossip/scoringParameters.ts
```diff
@@ -0,0 +1,275 @@
...
+};
+
+type MeshMessageInfo = {
+  decaySlots: number;
+  capFactor: number;
+  activationWindow: number;
+  currentSlot: number;
+};
+
+export class GossipPeerScoreParamsBuilder {
```
dapplion: This should not be a class, it holds not state and is only called one. Please convert to function that returns the params

packages/lodestar/src/chain/chain.ts
```diff
@@ -136,6 +138,11 @@ export class BeaconChain implements IBeaconChain {
       opts,
     });
 
+    this.eth2Context = {
```
dapplion: Don't attach this data to the state, it's unnecessary. Just pass the Eth2Context down to gossipsub

dapplion (approved):
Looks good! Overall structure and approach is sound but I have not verified all numbers and constants.

wemeetagain (approved):
Lgtm

PR #3134, opened 2021-09-14
Add op pool for slashings and exits

dapplion:
**Motivation**

Our current operation pools have a critical flaw:
- They don't verify that the existing items are valid in the state when proposing blocks. This will very likely invalidate all blocks Lodestar tries to propose in the epochs close to slashings and exists being propagated through the network.

Also the gossip validation of slashings and exits required db reads, specially for attester slashings where all values had to be read.

**Description**

- Use in-memory cache for attester slashing, proposer slashing and voluntary exists
- Use this fast memory cache for gossip validation, with a fast index for attester slashing intersected indexes
- Read from DB once on start-up, and persist to DB once on shutdown


wemeetagain (approved):
LGTM :tada:

Issue #3173, opened 2021-09-15
Improve processInactivityUpdates


twoeths:
**Is your feature request related to a problem? Please describe.**

1. Right now `processInactivityUpdates` takes the same amount of time to `processRewardsAndPenalties`, and all of the time is for the loop not for the update itself
```
Sep-15 08:25:18.745 []                 info: Syncing - 16 minutes left - 2.55 slots/s - head: 1260909 0x6d2a…7479 - finalized: 1260832 0x9238…3285 (39401) - clockSlot: 1263426 - peers: 3
@@@ processInactivityUpdates for epoch 39403 in 108 num change 0
@@@ processRewardsAndPenalties for epoch 39403 in 120
@@@ processInactivityUpdates for epoch 39404 in 106 num change 0
@@@ processRewardsAndPenalties for epoch 39404 in 111
```
this is due to `FLAG_ELIGIBLE_ATTESTER` being imported

![Screen Shot 2021-09-15 at 08 25 37](https://user-images.githubusercontent.com/10568965/133355931-3f93edb2-2599-4302-9d6a-872647a02eda.png)

2. When we have to process empty epoch, it's likely that we have to update `inactivityScore` for all validators and it makes the function run like crazy
![Screen Shot 2021-09-15 at 08 48 19](https://user-images.githubusercontent.com/10568965/133357133-d8b975da-b244-4d39-b85a-3cfe22ae75fe.png)

the above profile was taken when we have OOM issue (node is almost synced to head), the whole epoch transition takes > 11s, most of that is for `processInactivityUpdates` (`processRewardsAndPenalties` takes only 700ms in this case) 

**Describe the solution you'd like**

1. this is due to the way `tsc` compiles our code to always import inline
```
if ((0, util_1.hasMarkers)(status.flags, util_1.FLAG_ELIGIBLE_ATTESTER)) {
```
we should find a way to import that only 1 time, or redefine the constant

2. If we have to update all `inactivityScores` we should rebuild the whole tree (similar to `balances.UpdateAll`) instead of doing individual updates. The worse performance of this function should be close to `processRewardsAndPenalties` where we rebuild the whole tree.

PR #3178, opened 2021-09-15
Improve processInactivityUpdates

twoeths:
**Motivation**

+ `processInactivityUpdates` is slow and almost equal to `processRewardsAndPenalties` even we don't update anything
+ the profile is shown in #3173

**Description**

1. Loop through the persistent-ts array
2. New method `setMultiple` so that we go through the proxy only 1 time. We'll enhance `setMultiple` in another PR to build the whole tree from scratch if we have to update so many values.
3. Avoid importing functions/constants from other modules inside for loop

This gives 5x-7x improvement when I run altair and 3x in the performance test. Note that the improvement #3 gives only 10% speed increasement but we can apply same strategy for other functions 

resolves issue 1 of #3173

### DIFF
--- a/packages/beacon-state-transition/test/perf/altair/epoch/processInactivityUpdates.test.ts
+++ b/packages/beacon-state-transition/test/perf/altair/epoch/processInactivityUpdates.test.ts
@@ -1,4 +1,4 @@
-import {itBench, setBenchOpts} from "@dapplion/benchmark";
+import {itBench} from "@dapplion/benchmark";
 import {altair} from "../../../../src";
 import {FlagFactors, generateBalanceDeltasEpochProcess} from "../../phase0/epoch/util";
 import {StateAltairEpoch} from "../../types";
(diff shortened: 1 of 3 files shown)


twoeths:
```
pyrmont_e62330 - altair processInactivityUpdates | 8.9175 ms/op | 72.662 ms/op | 0.12
```

@dapplion looks like we miss the nice performance improvement alert in the old version of benchmark tool

dapplion:
> ```
> pyrmont_e62330 - altair processInactivityUpdates | 8.9175 ms/op | 72.662 ms/op | 0.12
> ```
> 
> @dapplion looks like we miss the nice performance improvement alert in the old version of benchmark tool

It will only show up if the improvement is better than the current threshold. In CI that's 3, so the factor in CI has to be lower than 0.33.

Also the benchmark `pyrmont_e62330 - altair processInactivityUpdates` has a threshold of Infinity, so it won't ever show here as an improvement

packages/beacon-state-transition/test/perf/altair/epoch/processInactivityUpdates.test.ts
```diff
@@ -21,6 +21,10 @@ import {mutateInactivityScores} from "./util";
 
 describe("altair processInactivityUpdates", () => {
   const vc = numValidators;
+  setBenchOpts({
+    maxMs: 60 * 1000,
+    minMs: 50 * 1000,
```
dapplion: Why are this parameters necessary?

twoeths: I read the benchmark result wrong, thought the test was stable.
Reverted the change.

dapplion (approved):

Issue #3076, opened 2021-09-03
Implement merge spec + spec tests


dapplion:
**What is your question?**

See https://github.com/ethereum/consensus-specs/blob/dev/specs/merge/beacon-chain.md

- [x] #3093
- [x] #3185
- [x] #3194
- [x] #3196
- [x] #3198
- [x] #3254
- [x] #3259
- [x] #3263
- [x] #3252
- [x] Test merge interop with Geth Catalyst

--------

Pending updates:
- [x] https://github.com/ethereum/consensus-specs/pull/2605
- [x] https://github.com/ethereum/consensus-specs/pull/2606
- [x] https://github.com/ethereum/consensus-specs/pull/2609
- [x] https://github.com/ethereum/consensus-specs/pull/2613
- [ ] https://github.com/ethereum/consensus-specs/pull/2617
- [ ] https://github.com/ethereum/consensus-specs/pull/2621
- [ ] https://github.com/ethereum/consensus-specs/pull/2682
- [ ] https://github.com/ethereum/consensus-specs/pull/2683
- [ ] https://github.com/ethereum/consensus-specs/pull/2684
- [ ] https://github.com/ethereum/consensus-specs/pull/2686
- [ ] https://github.com/ethereum/consensus-specs/pull/2687
- [ ] https://github.com/ethereum/consensus-specs/pull/2699

Non-PR todo:
- [ ] Polish optimistic sync, only attest and build blocks on fully verified execution payloads

--------

Additional material
- The Merge Mainnet Readiness Checklist: [ethereum/pm/blob/master/Merge/mainnet-readiness](https://github.com/ethereum/pm/blob/master/Merge/mainnet-readiness.md)
- Rayonism spec https://github.com/ethereum/rayonism/blob/master/specs/merge.md#Consensus-JSON-RPC
- Merge tracking issue Prysm: https://github.com/prysmaticlabs/prysm/issues/9278
- Merge tracking issue Teku: https://github.com/ConsenSys/teku/issues/4226
- Teku rayonism work > current master https://github.com/ConsenSys/teku/compare/master...txrx-research:post-rayonism-merge?expand=1

dapplion:
Added https://github.com/ethereum/consensus-specs/pull/2687 and more merged PRs

dapplion:
Closing for https://github.com/ChainSafe/lodestar/issues/3410

PR #3184, opened 2021-09-16
Add merge state transition

### LINKED ISSUE #3185: Implement merge state transition

dapplion:
**Motivation**

Part of https://github.com/ChainSafe/lodestar/issues/3076

**Description**

Closes https://github.com/ChainSafe/lodestar/issues/3185

### DIFF
--- a/packages/beacon-state-transition/src/merge/block/processExecutionPayload.ts
+++ b/packages/beacon-state-transition/src/merge/block/processExecutionPayload.ts
@@ -1,0 +1,108 @@
+import {GAS_LIMIT_DENOMINATOR, MIN_GAS_LIMIT} from "@chainsafe/lodestar-params";
+import {merge, ssz} from "@chainsafe/lodestar-types";
+import {byteArrayEquals, List} from "@chainsafe/ssz";
+import {CachedBeaconState} from "../../allForks";
+import {getCurrentEpoch, getRandaoMix} from "../../util";
+import {ExecutionEngine} from "../executionEngine";
+
+export function isExecutionEnabled(state: merge.BeaconState, body: merge.BeaconBlockBody): boolean {
+  return isMergeBlock(state, body) || isMergeComplete(state);
+}
+
+function isMergeBlock(state: merge.BeaconState, body: merge.BeaconBlockBody): boolean {
+  return (
+    !isMergeComplete(state) &&
+    !ssz.merge.ExecutionPayload.equals(body.executionPayload, ssz.merge.ExecutionPayload.defaultValue())
+  );
+}
+
+function isMergeComplete(state: merge.BeaconState): boolean {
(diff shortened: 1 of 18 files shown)


packages/beacon-state-transition/src/merge/block/processExecutionPayload.ts
```diff
@@ -0,0 +1,108 @@
...
+import {getCurrentEpoch, getRandaoMix} from "../../util";
+import {ExecutionEngine} from "../executionEngine";
+
+export function isExecutionEnabled(state: merge.BeaconState, body: merge.BeaconBlockBody): boolean {
+  return isMergeBlock(state, body) || isMergeComplete(state);
+}
+
+function isMergeBlock(state: merge.BeaconState, body: merge.BeaconBlockBody): boolean {
+  return (
+    !isMergeComplete(state) &&
```
wemeetagain, line 14: Something seems wrong here

should this be `||`?

packages/beacon-state-transition/src/merge/block/processExecutionPayload.ts
```diff
@@ -0,0 +1,107 @@
...
+    !ssz.merge.ExecutionPayload.equals(body.executionPayload, ssz.merge.ExecutionPayload.defaultValue())
+  );
+}
+
+function isMergeComplete(state: merge.BeaconState): boolean {
```
wemeetagain, line 19: This is probably just naive implementation, but if we pass in a `TreeBacked<merge.BeaconState>`, we can get a (probably) faster implementation:
```ts
return byteArrayEquals(
  state.latestExecutionPayloadHeader.hashTreeRoot(),
  ssz.merge.ExecutionPayloadHeader.defaultTree.root
);
```

Maybe better saved for future PRs after more benchmarking

packages/beacon-state-transition/src/merge/block/processExecutionPayload.ts
dapplion: Python code is
```py
def is_merge_block(state: BeaconState, body: BeaconBlockBody) -> bool:
    return not is_merge_complete(state) and body.execution_payload != ExecutionPayload()
```   

https://github.com/ethereum/consensus-specs/blob/0eb0a934a3/specs/merge/beacon-chain.md#is_merge_block

packages/beacon-state-transition/src/merge/block/processExecutionPayload.ts
dapplion: I'm not sure because the very first property in latestExecutionPayloadHeader is parentHash which will likely be non-zero for any value but the empty one. So in most cases a single byte will be compared. However your solution could trigger much more work if used incorrectly.

wemeetagain (approved):

Issue #3117, opened 2021-09-12
script to prohibit exemplar use in grafana dashboard


g11tech:
As in some grafana instances, exemplar query is giving 404 not found, add test script to prohibit exemplar use in grafana lodestar.json, as by default in new panel exemplar is on by default.

 https://github.com/ChainSafe/lodestar/blob/6fae28646d2b783acd0fe1b2d95f012efcc68769/.github/workflows/test.yml#L30 

**Additional context**
context: 
![image](https://user-images.githubusercontent.com/76567250/132984096-fe60686d-b748-463d-9b93-cc14ad82013f.png)

PR #3192, opened 2021-09-16
inspecting lodestar.json with regexp to invalidate use of exemplar

g11tech:
**Motivation**
since some of the grafana panels were showing 404 not found on using the exemplar, it was decided to avoid exemplar query use as we don't use the feature at all. however exemplar is turned on by default while creating new query.
This PR adds the check for the same.

**Description**

Closes #issue_number
#3117

example of github checks failing (mock test):

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,9 @@ jobs:
         if: steps.cache-deps.outputs.cache-hit == 'true'
       # </common-build>
 
+
+      - name: Grafana Dashboard
+        run: yarn validate-gdash
       - name: Test root binary exists
         run: ./lodestar --version
       - name: Check Types
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,7 +32,7 @@ jobs:
       # </common-build>
 
 
-      - name: Grafana Dashboard
+      - name: Lint Grafana Dashboard
         run: yarn validate-gdash
       - name: Test root binary exists
         run: ./lodestar --version
--- a/package.json
+++ b/package.json
@@ -28,7 +28,8 @@
     "publish:release": "lerna publish from-package --yes --no-verify-access",
     "release": "lerna version --no-push --sign-git-commit",
     "postrelease": "git tag -d $(git describe --abbrev=0)",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -47,6 +48,7 @@
     "eslint-plugin-no-only-tests": "^2.4.0",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^3.1.4",
+    "jsonpath": "^1.1.1",
```
dapplion: No need to add a dependency for this, we want to have as little dependencies as possible in lodestar. Just do a regex yourself :+1:

dapplion (changes requested):

package.json
g11tech: hmmmm ok :+1:

scripts/validate-grafana-dashboard.js
```diff
@@ -0,0 +1,6 @@
+let lodestarDash = require("../docker/grafana/provisioning/dashboards/lodestar.json");
+lodestarDash = JSON.stringify(lodestarDash); //get everything in the same line
```
dapplion, line 2: Here you are parsing a JSON file to immediately serialize it again. You can just read it with fs.readFileSync which is what require() is doing under the hood.

.github/workflows/test.yml
```diff
@@ -31,6 +31,9 @@ jobs:
         if: steps.cache-deps.outputs.cache-hit == 'true'
       # </common-build>
 
+
+      - name: Grafana Dashboard
```
dapplion: name: Lint Grafana dashboard

scripts/validate-grafana-dashboard.js
g11tech: yes, doing it intentionally to remove new lines and  bring it back to same line to avoid this kind of scenario:
exemplar :
    true

also side benefit: it is sort of check that JSON is actually parse-able .

dapplion: Oh! Good point, then sounds good to me

dapplion (approved):

dapplion (approved):

PR #3200, opened 2021-09-17
Remove balanceList proxy

twoeths:
**Motivation**

+ Most of the time we use direct method of `BalanceList` so we don't need proxy
+ Performance of proxy is a concern, see #3109

**Description**
+ Remove BalanceList proxy
+ So BalanceList does not conform to List interface anymore, I have to make a separate property `balanceList` vs `balances`. Using `balances` would go through the CachedBeaconState proxy but let's only use `balanceList` instead of `balances` from now on.

it's not really convenient since BeaconState has `balances` in its interface but we always use `balanceList` in CachedBeaconState from this PR, please let me know your ideas @wemeetagain @dapplion


dapplion (approved):
Looks great!

PR #3202, opened 2021-09-18
Add extra_data field to the execution payload

dapplion:
**Motivation**

See https://github.com/ethereum/consensus-specs/pull/2606/files

**Description**

Add extra_data field to the execution payload


dapplion:
Fixed with https://github.com/ChainSafe/lodestar/pull/3260

PR #3209, opened 2021-09-20
Fix MESSAGE_DOMAIN_VALID_SNAPPY

twoeths:
**Motivation**

+ Everytime we receive IHAVE rpc control message and we issue IWANT rpc control message in response, we end up penaltilizing that same peer because that same message id never comes and timeout occurred through gossip tracer
+ Then I found that our messageId function is incorrect due to incorrect MESSAGE_DOMAIN_VALID_SNAPPY defined, it means when subsequent gossip messages come, we can't create a message id that match the one in IHAVE

**Description**

+ Fix MESSAGE_DOMAIN_VALID_SNAPPY
+ Refer to https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md

part of #3084

### DIFF
--- a/packages/lodestar/src/network/gossip/constants.ts
+++ b/packages/lodestar/src/network/gossip/constants.ts
@@ -7,13 +7,13 @@ import {GossipEncoding} from "./interface";
 export const GOSSIP_MSGID_LENGTH = 20;
 
 /**
- * 4-byte domain for gossip message-id isolation of *invalid* snappy messages
- */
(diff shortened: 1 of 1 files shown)


dapplion (approved):
Very nice find! Thanks for deeply researching the issue :pray: Shout-out to Lighthouse too for the invaluable help

twoeths:
@dapplion  this is just part of the issue of missing HEAD, merging this for a continued investigation

PR #3210, opened 2021-09-20
Don't store future blocks

dapplion:
**Motivation**

We already check gossip disparity when asserting correct slot. It's not useful for us to keep future blocks.

**Description**

Don't store future blocks

### DIFF
--- a/packages/lodestar/src/chain/eventHandlers.ts
+++ b/packages/lodestar/src/chain/eventHandlers.ts
@@ -322,14 +322,9 @@ export async function onErrorBlock(this: BeaconChain, err: BlockError): Promise<
   // err type data may contain CachedBeaconState which is too much to log
   this.logger.error("Block error", {slot: err.signedBlock.message.slot, errCode: err.type.code});
 
-  if (err.type.code === BlockErrorCode.FUTURE_SLOT) {
-    this.pendingBlocks.addBySlot(err.signedBlock);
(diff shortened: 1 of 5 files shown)


packages/lodestar/src/chain/eventHandlers.ts
```diff
@@ -322,14 +322,9 @@ export async function onErrorBlock(this: BeaconChain, err: BlockError): Promise<
   // err type data may contain CachedBeaconState which is too much to log
   this.logger.error("Block error", {slot: err.signedBlock.message.slot, errCode: err.type.code});
 
-  if (err.type.code === BlockErrorCode.FUTURE_SLOT) {
```
twoeths, line 325: we need to modify this https://github.com/ChainSafe/lodestar/blob/6cf7c2c3f3df6012bd4ce773ab3b5310c347e884/packages/lodestar/src/chain/blocks/validate.ts#L38

to allow some tolerance as well, otherwise node will be out of synced for some slots

dapplion: I've added the functionality of a simple queue if the block in between 0 and 500ms early with a promise in the validation function itself. Do you think this is good enough?

wemeetagain (approved):
At first I thought it might not be a good idea waiting inside the gossip validation. But it's fine considering we only should be getting 1 block per slot or so. Simple approach checks out here.

PR #3212, opened 2021-09-20
Refactor unknown block sync

### LINKED ISSUE #2409: Add sim test for onUnknownBlockRoot
**Describe the bug**

In this issue https://github.com/ChainSafe/lodestar/issues/2197 we found that if a node advances on it's own to epoch 3 without having gossip enabled and then enables receiving blocks from other 3 nodes of the correct chain, it is not able to correctly sync and switch to the correct chain.

**Expected behavior**

Add a sim test to purposely create that situation and ensure the node can recover

dapplion:
**Motivation**

Our unknown block sync works, but lacks key protections against various attacks

**Description**

- Add strong peer scoring against malicious peers feeding us bad chains
  - Note: penalizes all peers that referenced any descendant of an invalid block.
- Verify the received block against the hash request
- Limit the amount of blocks cached in unknown parent block chains
- Add unit tests, and full sync e2e tests for unknown parent block syncs

Also it changes the flow of handling an unknown block. Now the BeaconChain does not participate in it.
1. A block is received through gossip and throws UNKNWON_PARENT in validation
2. Gossip handler picks it up and emits the block to the sync
3. Sync fetches parent and checks against the fork choice directly if the parent is know
4. (a) if know, feed all descendant blocks to chain, (b) if not know keep searching.

----

Fixes https://github.com/ChainSafe/lodestar/issues/2409


packages/lodestar/src/api/impl/beacon/blocks/index.ts
```diff
@@ -147,7 +149,18 @@ export function getBeaconBlockApi({
...
       metrics?.registerBeaconBlock(OpSource.api, seenTimestampSec, signedBlock.message);
 
-      await Promise.all([chain.receiveBlock(signedBlock), network.gossip.publishBeaconBlock(signedBlock)]);
+      await Promise.all([
+        // Send the block, regardless of whether or not it is valid. The API
+        // specification is very clear that this is the desired behaviour.
+        network.gossip.publishBeaconBlock(signedBlock),
+
+        chain.processBlock(signedBlock).catch((e) => {
+          if (e instanceof BlockError && e.type.code === BlockErrorCode.PARENT_UNKNOWN) {
```
twoeths, line 158: this does not happen if a validator publish a block produced by us. Is this to protect against an attacker trying to publish random blocks? How about using CSRF token for all of our apis?

twoeths: nvm, CSRF token should not be used for lodestar only as it should be part of the spec

packages/lodestar/src/sync/unknownBlock.ts
```diff
@@ -0,0 +1,295 @@
...
+  private triggerUnknownBlockSearch = (): void => {
+    const blocks = this.pendingBlocks.size > 0 && getLowestPendingUnknownParents(this.pendingBlocks);
+
+    // Cheap early stop to prevent calling the network.getConnectedPeers()
+    if (!blocks) {
+      return;
+    }
+
+    // If the node loses all peers with pending unknown blocks, the sync will stall
+    const connectedPeers = this.network.getConnectedPeers();
```
twoeths, line 111: is it better to only get peers with `status.head >= ours`? it's no use to ask for blocks from syncing peers

twoeths (dismissed):
Looks great, just add one minor comment

packages/lodestar/src/sync/unknownBlock.ts
dapplion: Oh good point! Will do in another PR :pray:

twoeths (approved):

PR #3214, opened 2021-09-21
Add Gossip Block metrics

twoeths:
**Motivation**

In order to make sure our validators vote for correct HEAD

+ We want to know delay time when a gossip block is received
+ We want to know delay time when a gossip block is processed

**Description**

+ New metric `gossip_block_elappsed_time_till_received` and `gossip_block_elappsed_time_till_processed`

part of #3084

The new panels are grouped into "Gossip Block" row as below:
Screen Shot 2021-09-21 at 11 42 35

### DIFF
--- a/packages/lodestar/src/chain/blocks/process.ts
+++ b/packages/lodestar/src/chain/blocks/process.ts
@@ -11,8 +11,6 @@ import {allForks, ISignatureSet, CachedBeaconState} from "@chainsafe/lodestar-be
 import {CheckpointStateCache} from "../stateCache";
 import {IMetrics} from "../../metrics";
 import {toHexString} from "@chainsafe/ssz";
-import {IChainForkConfig} from "@chainsafe/lodestar-config";
-import {Number64} from "@chainsafe/lodestar-types";
 
 export type BlockProcessOpts = {
   /**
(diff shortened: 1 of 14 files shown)


packages/lodestar/src/chain/blocks/process.ts
```diff
@@ -40,7 +43,13 @@ export async function processBlock(modules: BlockProcessModules, job: IBlockJob)
     });
   }
 
-  await processBlocksInEpoch(modules, job, [job.signedBlock]);
+  const {signedBlock} = job;
+  await processBlocksInEpoch(modules, job, [signedBlock]);
+  if (metrics) {
+    // Returns the delay between the start of `block.slot` and `current time`
+    const delaySec = Date.now() / 1000 - (genesisTime + signedBlock.message.slot * config.SECONDS_PER_SLOT);
+    metrics.gossipBlock.elappsedTimeTillProcessed.observe(delaySec);
```
dapplion: Note sync and API submited blocks also come through this method so they will distort the metrics

dapplion: Oh, now that I just merge the unknown sync PR you can just use the processBlock promise in the gossip handler to time when it's done processing

dapplion:
@tuyennhv Can you please add a screen capture of how the metrics charts that you added look like?

dapplion (approved):

PR #3221, opened 2021-09-22
Refactor chain block verify > import flow

dapplion:
**Motivation**

Our current architecture for verifying and importing blocks in sub-optimal.

- The block processor has a function `runStateTransition` which triggers many permanent side-effects, i.e. updating fork-choice, db, etc.
- After implementing altair, `runStateTransition` is ran before verifying the signatures of the block, due to a requirement of the syncAggregate signatures.
- Critical import logic is split between `chain/blocks` and `chain/eventHandlers` reducing clarity and maintenance. That's was probably caused the above problem to go unnoticed.
- This architecture requires to have two event emitters, and internal one that runs the after import steps and an external one that is guaranteed to run after the import is fully completed.

**Description**

Same logic, but simplified architecture:

- processBlock calls:
  - (1) verifyBlock - no side effects, stricly after
  - (2) importBlock - contains ALL side effects and required steps to fully import the block. Then events are emitted


packages/lodestar/src/network/gossip/handlers/index.ts
```diff
@@ -93,29 +86,35 @@ export function getGossipHandlers(modules: ValidatorFnsModules): GossipHandlers
...
-                PeerId.createFromB58String(peerIdStr),
-                PeerAction.LowToleranceError,
-                "BadGossipBlock"
-              );
+      // Handler - MUST NOT `await`, to allow validation result to be propagated
+
+      metrics?.registerBeaconBlock(OpSource.gossip, seenTimestampSec, signedBlock.message);
+
+      // `validProposerSignature = true`, in gossip validation the proposer signature is checked
+      chain
```
dapplion, line 94: Gossip handlers must not be awaited. This nuance could be enforced latter with a better architecture

twoeths (approved):

PR #3232, opened 2021-09-22
Use closest state if justified state is not available

dapplion:
**Motivation**

If the justified state is not found, our node stops.

**Description**

Use closest state if justified state is not available


wemeetagain (approved):
using the least wrong balances > stalling

Issue #3211, opened 2021-09-20
ERR_INVALID_ATTESTATION_DOUBLE_VOTE


dapplion:
**Describe the bug**

A user in discord reported  https://discord.com/channels/593655374469660673/593655641445367808/889555125448892427

```
error: Error signing attestation slot=1304012, index=17, head=0xdca5…3d80, validator=0xa87e…afb1 code=ERR_INVALID_ATTESTATION_DOUBLE_VOTE, attestation={"sourceEpoch":40749,"targetEpoch":40750,"signingRoot":"0x667cbb5b329650705425bddc216491a059be6e4e4dcab96727b5f1989a7bd847"}, prev={"sourceEpoch":40749,"targetEpoch":40750,"signingRoot":"0x4818486efdecad5d700e2b048c7ca3c4f85132cfbc929fc5f932b85199a952c3"}
```

Lodestar version=v0.30.0 (npm), network=prater
Neither my vc or bn node were restarting. I am running both in cli. I couldn't find any grafana dashboards for metrics. Can you please link it?
I plan on running  lodestar using systemd but will probably do that later on today. Unfortunately, I don't have any saved logs.

Perhaps it's my network settings? I have a lighthouse node running on port 9000 so I switched lodestar to use 9001. These are the settings I'm using. 

```
lodestar beacon --network prater --network.localMultiaddrs /ip4/0.0.0.0/tcp/9001 --network.discv5.bindAddr /ip4/0.0.0.0/udp/9001 --eth1.providerUrls https://goerli.infura.io/v3/XXX --network.targetPeers 50 --network.maxPeers 55
```

**Expected behavior**

Ǹot happen

twoeths:
this happens in our contabo-7 too: 
`grep -e "ERR_INVALID_ATTESTATION_DOUBLE_VOTE" -nr validator*.log`

```
Sep-22 22:12:53.569 []                ^[[34mdebug^[[39m: Signed attestation slot=1320064, index=30, head=0x8ab7…5d87, valida      tor=0xa207…f892
```
that's maybe the wrong assignment for that validator since node can't verify it

```
 Sep-22 22:12:55.862 []                ^[[31merror^[[39m: Error publishing attestations slot=1320064, index=30 message=Intern      al Server Error: ATTESTATION_ERROR_INVALID_SIGNATURE
62005 Error: Internal Server Error: ATTESTATION_ERROR_INVALID_SIGNATURE
[...]

PR #3235, opened 2021-09-23
Insert SlashingProtectionAttestation after publish successfully

twoeths:
**Motivation**

+ As shown in https://github.com/ChainSafe/lodestar/issues/3211#issuecomment-925447606 when our node is not stable, attester duties may not be correct
+ But the validator insert SlashingProtectionAttestation right after it signs the attestation
+ Then the validation publish attestation unsucessfully (maybe with an invalid signature error)
+ When the next/correct duties come in same epoch, we got double vote error

**Description**

+ Only insert SlashingProtectionAttestation after we successfully publish attestation

part of #3211 

**TODOs**
- [ ] Determine if a specific attestation failed the validation at api side
- [ ] Consider doing that for block? We may not need it.

### DIFF
--- a/packages/validator/src/services/attestation.ts
+++ b/packages/validator/src/services/attestation.ts
@@ -8,6 +8,7 @@ import {ValidatorStore} from "./validatorStore";
 import {AttestationDutiesService, AttDutyAndProof} from "./attestationDuties";
 import {groupAttDutiesByCommitteeIndex} from "./utils";
 import {IndicesService} from "./indices";
+import {AttestationSigningResult} from "../types";
 
 /**
  * Service that sets up and handles validator attester duties.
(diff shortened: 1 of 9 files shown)


dapplion:
@tuyennhv That may be super dangerous. The spec and common security advice strictly says that the validator client MUST add the attestation to the slashing protection db BEFORE signing, and only sign after confirming it's permanently written in the db.

EDIT: We must not merge this strategy, this will get people slashed.

twoeths:
this is the spec for later reference

```
Specifically, when signing an Attestation, a validator should perform the following steps in the following order:

Save a record to hard disk that an attestation has been signed for source (i.e. attestation_data.source.epoch) and target (i.e. attestation_data.target.epoch).
Generate and broadcast attestation.
```

PR #3241, opened 2021-09-24
Validator: log full validator root hex for debug context

twoeths:
**Motivation**

To debug an issue in validator, it's hard to get validator index from `prettyBytes()`

**Description**

+ For debug context, log full validator root hex
+ For info context, keep `prettyBytes()`/short root hex

from #3211

### DIFF
--- a/packages/validator/src/services/attestation.ts
+++ b/packages/validator/src/services/attestation.ts
@@ -1,13 +1,14 @@
 import {AbortSignal} from "@chainsafe/abort-controller";
 import {phase0, Slot, CommitteeIndex, ssz} from "@chainsafe/lodestar-types";
 import {computeEpochAtSlot} from "@chainsafe/lodestar-beacon-state-transition";
-import {prettyBytes, sleep} from "@chainsafe/lodestar-utils";
+import {sleep} from "@chainsafe/lodestar-utils";
 import {Api} from "@chainsafe/lodestar-api";
 import {extendError, IClock, ILoggerVc} from "../util";
 import {ValidatorStore} from "./validatorStore";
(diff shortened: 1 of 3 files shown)


dapplion:
@tuyennhv Good point, why not logging the validator index instead? It should be available in those debug logs

twoeths:
> @tuyennhv Good point, why not logging the validator index instead? It should be available in those debug logs

@dapplion in the info logContext, we show `prettyBytes` hex root of validator for the user so I want to match that in our debug logContext to ease the investigation in case we have too many validators in the client. Do you think we want to log validator index in the debug logContext too? To me it's as simple as going to to the explorer to get it like https://prater.beaconcha.in/validator/0x811af134a14e68b8b14102bd8c5339864faa592b977eab427b245cc4e602c91f245f51ffb1ca04d0b36eeeaceac3ca6a

dapplion:
> > @tuyennhv Good point, why not logging the validator index instead? It should be available in those debug logs
> 
> @dapplion in the info logContext, we show `prettyBytes` hex root of validator for the user so I want to match that in our debug logContext to ease the investigation in case we have too many validators in the client. Do you think we want to log validator index in the debug logContext too? To me it's as simple as going to to the explorer to get it like https://prater.beaconcha.in/validator/0x811af134a14e68b8b14102bd8c5339864faa592b977eab427b245cc4e602c91f245f51ffb1ca04d0b36eeeaceac3ca6a

Right, it's just that the pubkey is so long and clutters the logs a lot. If we can just use the index it would be so much shorter and the logs cleaner

dapplion (approved):

PR #3242, opened 2021-09-24
Handle attester duties reorg

### LINKED ISSUE #3211: ERR_INVALID_ATTESTATION_DOUBLE_VOTE

twoeths:
**Motivation**

+ When getting attester duties for current epoch, we have `notWhileSyncing()` check and it's good but not enough when validator ask for next epoch
+ For example in https://github.com/ChainSafe/lodestar/issues/3211#issuecomment-926320834, head epoch is at 41250 while validator asked for attester duties at epoch 41252, node should throw error instead of returning unreliable duties

**Description**

+ Check requested epoch vs headEpoch (instead of currentEpoch), it should be <=1
+ Closes #3211

### DIFF
--- a/packages/lodestar/src/api/impl/events/index.ts
+++ b/packages/lodestar/src/api/impl/events/index.ts
@@ -37,7 +37,7 @@ export function getEventsApi({chain, config}: Pick<ApiModules, "chain" | "config
       }
       const currentEpoch = state.currentShuffling.epoch;
       const [previousDutyDependentRoot, currentDutyDependentRoot] = [currentEpoch - 1, currentEpoch].map((epoch) =>
-        toHexString(getBlockRootAtSlot(state, computeStartSlotAtEpoch(epoch) - 1))
+        toHexString(getBlockRootAtSlot(state, Math.max(computeStartSlotAtEpoch(epoch) - 1, 0)))
       );
       return [
         {
--- a/packages/lodestar/src/api/impl/events/index.ts
+++ b/packages/lodestar/src/api/impl/events/index.ts
@@ -1,11 +1,11 @@
-import {ApiModules} from "../types";
-import {ChainEvent, IChainEvents} from "../../../chain";
-import {routes} from "@chainsafe/lodestar-api";
 import {
   computeEpochAtSlot,
   computeStartSlotAtEpoch,
   getBlockRootAtSlot,
 } from "@chainsafe/lodestar-beacon-state-transition";
+import {ApiModules} from "../types";
+import {ChainEvent, IChainEvents} from "../../../chain";
+import {routes} from "@chainsafe/lodestar-api";
 import {ApiError} from "../errors";
 import {toHexString} from "@chainsafe/ssz";
 
--- a/packages/lodestar/src/api/impl/events/index.ts
+++ b/packages/lodestar/src/api/impl/events/index.ts
@@ -1,8 +1,11 @@
 import {ApiModules} from "../types";
 import {ChainEvent, IChainEvents} from "../../../chain";
 import {routes} from "@chainsafe/lodestar-api";
-import {computeEpochAtSlot, computeStartSlotAtEpoch} from "@chainsafe/lodestar-beacon-state-transition";
-import {ZERO_HASH} from "../../../constants";
+import {
+  computeEpochAtSlot,
+  computeStartSlotAtEpoch,
+  getBlockRootAtSlot,
+} from "@chainsafe/lodestar-beacon-state-transition";
 import {ApiError} from "../errors";
 import {toHexString} from "@chainsafe/ssz";
 
@@ -19,8 +22,6 @@ const chainEventMap = {
 };
 
 export function getEventsApi({chain, config}: Pick<ApiModules, "chain" | "config">): routes.events.Api {
-  const ZERO_HASH_HEX = toHexString(ZERO_HASH);
-
   /**
    * Mapping to convert internal `ChainEvents` payload to the API spec events data
    */
@@ -29,17 +30,26 @@ export function getEventsApi({chain, config}: Pick<ApiModules, "chain" | "config
       ...args: Parameters<IChainEvents[typeof chainEventMap[K]]>
     ) => routes.events.EventData[K][];
   } = {
-    [routes.events.EventType.head]: (head) => [
-      {
-        block: head.blockRoot,
-        epochTransition: computeStartSlotAtEpoch(computeEpochAtSlot(head.slot)) === head.slot,
-        slot: head.slot,
-        state: head.stateRoot,
-        // Todo implement
-        previousDutyDependentRoot: ZERO_HASH_HEX,
-        currentDutyDependentRoot: ZERO_HASH_HEX,
-      },
-    ],
+    [routes.events.EventType.head]: (head) => {
+      const state = chain.stateCache.get(head.stateRoot);
+      if (!state) {
+        throw Error("cannot get state for head " + head.stateRoot);
+      }
+      const currentEpoch = state.currentShuffling.epoch;
+      const [previousDutyDependentRoot, currentDutyDependentRoot] = [currentEpoch - 1, currentEpoch].map((epoch) =>
+        toHexString(getBlockRootAtSlot(state, computeStartSlotAtEpoch(epoch) - 1))
+      );
+      return [
+        {
+          block: head.blockRoot,
+          epochTransition: computeStartSlotAtEpoch(computeEpochAtSlot(head.slot)) === head.slot,
+          slot: head.slot,
+          state: head.stateRoot,
+          previousDutyDependentRoot,
+          currentDutyDependentRoot,
+        },
+      ];
+    },
     [routes.events.EventType.block]: (block) => [
       {
         block: toHexString(config.getForkTypes(block.message.slot).BeaconBlock.hashTreeRoot(block.message)),
--- a/packages/lodestar/src/api/impl/events/index.ts
+++ b/packages/lodestar/src/api/impl/events/index.ts
@@ -35,10 +35,12 @@ export function getEventsApi({chain, config}: Pick<ApiModules, "chain" | "config
       if (!state) {
         throw Error("cannot get state for head " + head.stateRoot);
       }
+
       const currentEpoch = state.currentShuffling.epoch;
       const [previousDutyDependentRoot, currentDutyDependentRoot] = [currentEpoch - 1, currentEpoch].map((epoch) =>
         toHexString(getBlockRootAtSlot(state, Math.max(computeStartSlotAtEpoch(epoch) - 1, 0)))
       );
+
       return [
         {
           block: head.blockRoot,
--- a/packages/lodestar/src/api/impl/validator/index.ts
+++ b/packages/lodestar/src/api/impl/validator/index.ts
@@ -275,6 +275,7 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
       // May request for an epoch that's in the future
       await waitForNextClosestEpoch();
 
+      // should not compare to headEpoch in order to handle skipped slots
       // Check if the epoch is in the future after waiting for requested slot
       if (epoch > chain.clock.currentEpoch + 1) {
         throw new ApiError(400, "Cannot get duties for epoch more than one ahead");
--- a/packages/validator/src/services/attestationDuties.ts
+++ b/packages/validator/src/services/attestationDuties.ts
@@ -19,10 +19,12 @@ export type AttDutyAndProof = {
   selectionProof: BLSSignature | null;
 };
 
+// To assist with readability
+type AttDutiesAtEpoch = {dependentRoot: RootHex; dutiesByIndex: Map<ValidatorIndex, AttDutyAndProof>};
+
 export class AttestationDutiesService {
   /** Maps a validator public key to their duties for each epoch */
-  private readonly dutiesByEpochByIndex = new Map<ValidatorIndex, Map<Epoch, AttDutyAndProof>>();
-  private readonly dependentRootByEpoch = new Map<Epoch, RootHex>();
+  private readonly dutiesByIndexByEpoch = new Map<Epoch, AttDutiesAtEpoch>();
   /**
    * We may receive new dependentRoot of an epoch but it's not the last slot of epoch
    * so we have to wait for getting close to the next epoch to redownload new attesterDuties.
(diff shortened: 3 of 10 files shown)


packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -280,7 +280,13 @@ export function getValidatorApi({
       await waitForNextClosestEpoch();
 
       // Check if the epoch is in the future after waiting for requested slot
-      if (epoch > chain.clock.currentEpoch + 1) {
+      const headEpoch = computeEpochAtSlot(chain.forkChoice.getHead().slot);
+      if (epoch > headEpoch + 1) {
```
g11tech: ~~ideally you should || i.e. epoch > headEpoch + 1 || epoch > chain.clock.currentEpoch + 1
in both scenarios reject api call~~
notWhileSyncing wouldn't take care of this?

g11tech:
just to mention @wemeetagain 's comment on the lodestar-testnet-alerts channel which might be relevant:

>You may wonder, why even allow getting duties if we're not fully synced? Well it's not always clear whether you're fully synced or not.
>In normal network conditions, yes, its easy to determine that your node is synced. Your clock ticks to the next slot, you receive and process that slot's block via gossip, your node determines that the new block is the new head of the chain.
>However when the network is degraded, for example during a long chain of skip slots, being synced looks just like being unsynced. So a node always needs to be able to handle this unsure area and ensure that their validators will be able to progress the chain even in these conditions.

in case of skipped slots, this will cause the attestation to freeze and not get involved in voting for zero blocks (atleast)
duties reorg with throwing of the double vote error behavior is a valid behavior and is harmless.

dapplion (changes requested):
This can be very dangerous is there are many skipped slots, like +64. It won't allow validators to participate anymore right? Because the head will be too far back.

dapplion:
I think that if the validator tracks the head event with dependant root this issue will be fixed much cleanly since the validator would realize in time when the re-org happens

g11tech:
> I think that if the validator tracks the head event with dependant root this issue will be fixed much cleanly since the validator would realize in time when the re-org happens

true, as long as validators dont double vote, duties reorg is much preferred behavior

twoeths:
**Update**
1. Only change validator, keep the node as is. Right now validator only detect duties reorg passively
2. We want to track duties reorg actively through ChainHeadTracker
  + If new head comes at the same to dependentRoot's epoch of any dutyEpoch, and the new root is different to what's the validator is tracking, and it's the last slot of an epoch => poll duties again
  + If it's not the last slot of epoch, put it to pending state. When it's 2/3 slot to the next epoch, poll duties again. (in this case, dependentRoot may not be the last slot of an epoch)
3. Refactor AttestationDutiesService to store dependent root by epoch in 2 new maps, and prune them when needed

TODOs
- [x] Test on prater (not easy to detect attesterDuties reorg but at least want to make sure it does not affect normal flow)

In my test node, I was able to catch an error like that when the node is syncing
```
Sep-28 06:56:17.478 []                 info: Found attesterDuties reorg through new head slot=1358079, dutyEpoch=42441, root=0x3ad269556b768a43571b5dfc212ed09fb056a13e3573054d008145ef94ebd8d1
Sep-28 06:56:19.665 []                 warn: Attester duties re-org. This may happen from time to time priorDependentRoot=0x816f62c55d1d95b99dcba47909fdc67e30da83c03010b58f1226dec3f165a27b, dependentRoot=0x3ad269556b768a43571b5dfc212ed09fb056a13e3573054d008145ef94ebd8d1
```

Validator works well after that

g11tech (dismissed):
just an idea or rather curiosity, why shouldn't we connect through websocket for just piping the events from bn to the connected validator processes rather than polling to bn, then validator would be more naturally processing events.

twoeths:
need to `implement/consume previousDutyDependentRoot and currentDutyDependentRoot` in https://ethereum.github.io/beacon-APIs/#/Events/eventstream

twoeths:
Update: 
+ handle attester duties reorg due to the node is syncing
+ (new) handle attester duties reorg due to the head is reorg

packages/lodestar/src/api/impl/events/index.ts
```diff
@@ -29,17 +30,26 @@ export function getEventsApi({chain, config}: Pick<ApiModules, "chain" | "config
...
-        // Todo implement
-        previousDutyDependentRoot: ZERO_HASH_HEX,
-        currentDutyDependentRoot: ZERO_HASH_HEX,
-      },
-    ],
+    [routes.events.EventType.head]: (head) => {
+      const state = chain.stateCache.get(head.stateRoot);
+      if (!state) {
+        throw Error("cannot get state for head " + head.stateRoot);
+      }
```
dapplion, line 37: This is dangerous. The state is really not needed. You can figure out the dependant root only with the forkchoice. The dependant root is the parent of the target block. The `head` event argument here should already include all the data necessary

packages/validator/src/services/chainHeaderTracker.ts
```diff
@@ -28,12 +38,30 @@ export class ChainHeaderTracker {
     return null;
   }
 
+  runOnNewHead(fn: RunEveryFn): void {
+    this.fns.push(fn);
+  }
+
```
dapplion, line 44: What about the reorg event? Should we subscribe to it too? Are other clients using the head + reorg events or only head?

packages/validator/src/services/chainHeaderTracker.ts
```diff
@@ -28,12 +38,30 @@ export class ChainHeaderTracker {
...
     if (event.type === EventType.head) {
       const {message} = event;
-      this.headBlockSlot = message.slot;
-      this.headBlockRoot = fromHexString(message.block);
-      this.logger.verbose("Found new chain head", {slot: message.slot, blockRoot: message.block});
+      const {slot, block, previousDutyDependentRoot, currentDutyDependentRoot} = message;
+      this.headBlockSlot = slot;
+      this.headBlockRoot = fromHexString(block);
+      for (const fn of this.fns) {
+        void fn({
```
dapplion: Better to .catch() here than void just in case

packages/lodestar/src/api/impl/events/index.ts
dapplion: Here these is some example code on how to compute dependant roots using the fork-choice https://github.com/ChainSafe/lodestar/blob/93a179e6fd9b64febbcf4239dccab366b215ec3f/packages/lodestar/src/chain/regen/interface.ts?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L271

It should be improved, removing the `!` operator

packages/validator/src/services/attestationDuties.ts
```diff
@@ -16,23 +20,29 @@ export type AttDutyAndProof = {
...
 };
 
-// To assist with readability
-type AttDutyAtEpoch = {dependentRoot: Root; dutyAndProof: AttDutyAndProof};
-
 export class AttestationDutiesService {
   /** Maps a validator public key to their duties for each epoch */
-  private readonly dutiesByEpochByIndex = new Map<ValidatorIndex, Map<Epoch, AttDutyAtEpoch>>();
+  private readonly dutiesByEpochByIndex = new Map<ValidatorIndex, Map<Epoch, AttDutyAndProof>>();
+  private readonly dependentRootByEpoch = new Map<Epoch, RootHex>();
```
dapplion: I think you can merge this two data structures together using
```ts
dutiesByEpochByIndex = new Map<
  Epoch, 
  {dependantRoot: RootHex, dutiesByIndex: Map<ValidatorIndex, AttDutyAndProof>}
>();
```
Indexing by epoch -> validator should be as fast as now, and it allows faster pruning too

packages/validator/src/services/attestationDuties.ts
```diff
@@ -167,24 +198,94 @@ export class AttestationDutiesService {
...
+    head,
+    previousDutyDependentRoot,
+    currentDutyDependentRoot,
+  }: HeadEventData): Promise<void> => {
+    const currentEpoch = computeEpochAtSlot(slot);
+    const nextEpoch = currentEpoch + 1;
+    const nextTwoEpoch = currentEpoch + 2;
+    const nextTwoEpochDependentRoot = this.dependentRootByEpoch.get(currentEpoch + 2);
+    // node is syncing, it may cause the attester duties reorg
+    if (nextTwoEpochDependentRoot && head !== nextTwoEpochDependentRoot) {
```
dapplion, line 237: Please add a nice comment here explaining why this 3rd case is necessary, linking to the original issue that caused it. Also please add white space (1 new line) between each of the three cases

packages/validator/src/services/chainHeaderTracker.ts
twoeths: Teku handles both https://github.com/ConsenSys/teku/blob/669c608d5a5a3eae1345e8a751d3c6c1754a6e3d/validator/client/src/main/java/tech/pegasys/teku/validator/client/AbstractDutyScheduler.java#L74 (see `onChainHead()` and `onChainReorg()`)

packages/lodestar/src/api/impl/events/index.ts
twoeths: @dapplion forkchoice may not store the block of `compute_start_slot_at_epoch(currentEpoch - 1) - 1` because it's pruned (given the current status of https://prater.beaconcha.in/) so I have to go with getting the head state here. I think getting headState is safe enough as it should be the same data to forkchoice, let me know if you still think it's dangerous.

just check that Teku also fetch dependent root from state https://github.com/ConsenSys/teku/blob/669c608d5a5a3eae1345e8a751d3c6c1754a6e3d/storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java#L322

dapplion: Got it, but they don't assume that the head state will be in the cache. I understand that `retrieveStateAndBlockSummary` is sort of a regen call. I'm okay merging this now and cleaning up our regen strategy latter.

packages/validator/src/services/chainHeaderTracker.ts
twoeths: Lighthouse instead keep polling duties per slot so it does not need to handle reorg or head event.

With the mature `on_head` event api at this time, I think handling `on_head` is enough for us. As long as dependent roots are not changed, it's not necessary to poll duties again even through it's a reorg.

twoeths:
tested with a node of 4 validators, I think the performance is the same. We'll need to produce attestation data as soon as validator see new head in a separate PR.

Screen Shot 2021-10-19 at 21 21 01

packages/validator/src/services/attestationDuties.ts
```diff
@@ -17,39 +20,72 @@ export type AttDutyAndProof = {
...
       }
     }
 
     return duties;
   }
 
+  /**
+   * If a reorg dependent root comes at a slot other than last slot of epoch
+   * just update this.pendingDependentRootByEpoch() and process here
+   * @returns
```
wemeetagain: stray `@returns` line

packages/validator/src/services/attestationDuties.ts
```diff
@@ -17,39 +20,71 @@ export type AttDutyAndProof = {
...
 
-    for (const dutiesByEpoch of this.dutiesByEpochByIndex.values()) {
-      const dutyAtEpoch = dutiesByEpoch.get(epoch);
-      if (dutyAtEpoch && dutyAtEpoch.dutyAndProof.duty.slot === slot) {
-        duties.push(dutyAtEpoch.dutyAndProof);
+    const epochDuties = this.dutiesByIndexByEpoch.get(epoch);
+    if (epochDuties === undefined) {
+      return duties;
+    }
+    for (const validatorDuty of epochDuties.dutiesByIndex.values()) {
```
dapplion, line 58: Add white space before for loop

packages/validator/src/services/attestationDuties.ts
```diff
@@ -139,50 +173,122 @@ export class AttestationDutiesService {
...
     });
 
-    let alreadyWarnedReorg = false;
-    for (const duty of relevantDuties) {
-      let dutiesByEpoch = this.dutiesByEpochByIndex.get(duty.validatorIndex);
-      if (!dutiesByEpoch) {
-        dutiesByEpoch = new Map<Epoch, AttDutyAtEpoch>();
-        this.dutiesByEpochByIndex.set(duty.validatorIndex, dutiesByEpoch);
+    const prior = this.dutiesByIndexByEpoch.get(epoch);
+    const priorDependentRoot = prior?.dependentRoot;
```
dapplion: You can merge this lines into 
```ts
const priorDependentRoot = this.dutiesByIndexByEpoch.get(epoch)?.dependentRoot;
```
And use `priorDependentRoot` in the if statement below

dapplion (approved):

dapplion:
Awesome! :fire:

PR #3245, opened 2021-09-24
Range sync: add ignoreIfFinalized flag

### LINKED ISSUE #3240: RangeSync should handle PARENT_UNKNOWN errors
**Describe the bug**

From this comment in the block processor, if the sync ignores WOULD_REVERT_FINALIZED_SLOT the sync must also be able to recover. 

https://github.com/ChainSafe/lodestar/blob/4e8e969fa4223749ce760a02563913f3dbb44a94/packages/lodestar/src/chain/blocks/index.ts#L104-L117

twoeths:
**Motivation**

+ In Range Sync, some preFinalized blocks come and it throws error BLOCK_ERROR_PARENT_UNKNOWN
+ The old range sync works by filtering those blocks out before processing remaining blocks

**Description**

+ For range sync, add `isPreFinalized` flags that ignore the error
+ This is not applied for `processBlock`
+ Switch conditions in `verifyBlockSanityChecks`, more specific errors should be checked first

Closes #3240

### DIFF
--- a/packages/lodestar/src/chain/blocks/index.ts
+++ b/packages/lodestar/src/chain/blocks/index.ts
@@ -114,6 +114,9 @@ export async function processChainSegment(
       ) {
         continue;
       }
+      if (partiallyVerifiedBlock.ignoreIfFinalized && err.type.code == BlockErrorCode.WOULD_REVERT_FINALIZED_SLOT) {
+        continue;
+      }
 
       modules.emitter.emit(ChainEvent.errorBlock, err);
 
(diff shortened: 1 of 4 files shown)


packages/lodestar/src/chain/blocks/index.ts
```diff
@@ -114,6 +114,9 @@ export async function processChainSegment(
       ) {
         continue;
       }
+      if (partiallyVerifiedBlock.ignoreIfFinalized && err.type.code == BlockErrorCode.WOULD_REVERT_FINALIZED_SLOT) {
```
g11tech, line 117: should all skips (continues) be logged with logger.debug("skipping block ....",{errCode:BlockErrorCode...})
would be use in debug.

very minor" also could move in the same if block (especially if the logger.debug is to be added)

dapplion: Good idea! Could you do a new PR?

dapplion (approved):

Issue #3243, opened 2021-09-24
slot less than finalized block


twoeths:
**Describe the bug**

Got this error on contabo-0 with latest master (version=v0.30.0/master/+42/72628fb0 (git))

```
442630 Sep-24 04:18:44.275 [SYNC]          ^[[36mverbose^[[39m: Batch process error id=Finalized, startEpoch=41400, status=Processing code=BLOCK_ERROR_BEACON_CHAIN_ERR       OR, error={"message":"slot less than finalized block"}
442631 Error: slot less than finalized block
442632     at ProtoArray.getAncestorAtSlot (/usr/app/node_modules/@chainsafe/lodestar-fork-choice/src/protoArray/protoArray.ts:655:11)
442633     at ProtoArray.getCommonAncestor (/usr/app/node_modules/@chainsafe/lodestar-fork-choice/src/protoArray/protoArray.ts:630:21)
442634     at ForkChoice.getCommonAncestorDistance (/usr/app/node_modules/@chainsafe/lodestar-fork-choice/src/forkChoice/forkChoice.ts:635:44)
442635     at importBlock (/usr/app/node_modules/@chainsafe/lodestar/src/chain/blocks/importBlock.ts:129:39)
442636     at processChainSegment (/usr/app/node_modules/@chainsafe/lodestar/src/chain/blocks/index.ts:102:24)
442637     at Timeout.JobItemQueue.runJob [as _onTimeout] (/usr/app/node_modules/@chainsafe/lodestar/src/util/queue/itemQueue.ts:85:22)
442638 Error: slot less than finalized block
442639     at ProtoArray.getAncestorAtSlot (/usr/app/node_modules/@chainsafe/lodestar-fork-choice/src/protoArray/protoArray.ts:655:11)
442640     at ProtoArray.getCommonAncestor (/usr/app/node_modules/@chainsafe/lodestar-fork-choice/src/protoArray/protoArray.ts:630:21)
442641     at ForkChoice.getCommonAncestorDistance (/usr/app/node_modules/@chainsafe/lodestar-fork-choice/src/forkChoice/forkChoice.ts:635:44)
442642     at importBlock (/usr/app/node_modules/@chainsafe/lodestar/src/chain/blocks/importBlock.ts:129:39)
442643     at processChainSegment (/usr/app/node_modules/@chainsafe/lodestar/src/chain/blocks/index.ts:102:24)
[...]

PR #3250, opened 2021-09-27
Modify error message for getAncestorAtSlot

twoeths:
**Motivation**

The error "slot less than finalized block" is kind of misleading and missing specific information

**Description**

Change to `cannot get ancestor for block ${node.blockRoot} at slot ${slot}`

part of #3243


twoeths:
this method was removed with latest master

PR #3258, opened 2021-09-27
propogating skipped block logic to processblock as well and debug logging on block skip

g11tech:
**Motivation**
This PR makes the block skip logic same on processBlock and processChainSegment, plus adds debug logs on block skip

context: https://github.com/ChainSafe/lodestar/pull/3245#discussion_r716029538

**Description**

Closes #issue_number

### DIFF
--- a/packages/lodestar/src/chain/blocks/index.ts
+++ b/packages/lodestar/src/chain/blocks/index.ts
@@ -3,6 +3,7 @@ import {AbortSignal} from "@chainsafe/abort-controller";
 import {allForks} from "@chainsafe/lodestar-types";
 import {sleep} from "@chainsafe/lodestar-utils";
 import {ChainEvent} from "../emitter";
+import {PendingEvents} from "./utils/pendingEvents";
 import {JobItemQueue} from "../../util/queue";
 import {BlockError, BlockErrorCode, ChainSegmentError} from "../errors";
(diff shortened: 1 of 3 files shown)


dapplion:
Good idea to de-duplicate! However we should
- emit error events immediately as they happen (don't add to pending)
- the skip event is not useful for anyone, so we should not emit it at all since it has some performance cost
- getBlockErrorEvent could be renamed to getBlockAction, and should return either action skip or action error.
- However may never want to allow single blocks to be skipped with the ignoreIfFinalized option. That only really makes sense for chain segments

g11tech:
> Good idea to de-duplicate! However we should
> 
>     * emit error events immediately as they happen (don't add to pending)
> 
error event is immediate emit (not being added to pending) as there can only be one error event  immediately throwing with the segmenterror (same as previous code)
>     * the skip event is not useful for anyone, so we should not emit it at all since it has some performance cost
> 
didnt have the logger handle here, will bring it down here.

>     * getBlockErrorEvent could be renamed to getBlockAction, and should return either action skip or action error.
> 
sounds good.
>     * However may never want to allow single blocks to be skipped with the ignoreIfFinalized option. That only really makes sense for chain segments
ok, will try to go deeper into ignoreIfFinalized option

g11tech:
putting to ice as per discussion

Issue #3120, opened 2021-09-13
investigate: getState related all regenFns's metric show empty data plus getPreState regenFn job metrics


g11tech:
![image](https://user-images.githubusercontent.com/76567250/133033750-06a62ded-5f9a-4c1a-b451-83eeee603fcd.png)

![image](https://user-images.githubusercontent.com/76567250/133034063-9e316f1e-dd05-4298-b25f-a7fdd0ee734d.png)

g11tech:
on a cursory glance, seems like queued `getState` seems to be never called (which is our entrypoint for starting any regen metric).
however will there is a getState in api/impl/beacon/state/index.ts that further calls resolveStateId:

resolveStateId either picks from stateCache, or stateArchive or constructs from stateArchive.
seems like getState is not being called from outside.

g11tech:
seems like `getPreState` computation is never getting queued for computation and is getting resolved in cache only.
it does get called from `processBlocksInEpoch` but gets resolved using cache's only in regen/queued.ts and never further ahead.

here is an effort to look up any observation across last 30 days in a prater running node"
![image](https://user-images.githubusercontent.com/76567250/133906907-203d114f-c45b-4209-a04b-d18389142efe.png)

g11tech:
@dapplion  should these metrics panels be removed from regenFn Stats of the metrics dashboard? or should I aggregate all these metrics in a single widget (multiple targets), someday if something shows up we can dig deeper
atleast we can remove the crowding from regenFn Stats

dapplion:
> @dapplion should these metrics panels be removed from regenFn Stats of the metrics dashboard? or should I aggregate all these metrics in a single widget (multiple targets), someday if something shows up we can dig deeper
> atleast we can remove the crowding from regenFn Stats

Sounds good do a PR with screen captures and let's see how it looks. If you can make all panels by 2 per row format would be great, there are still 2 full row width panels.

g11tech:
get preState graphs are now showing up when the node is experiencing the sync failure, which means these graphs can be retained,

![image](https://user-images.githubusercontent.com/76567250/134779532-30e8dbdc-8532-4cbd-9c8c-03cc1a2d044e.png)

will collapse only getState graphs (apart from also resizing the full width graphs into half width)

dapplion:
@g11tech Why is the cache hit ratio negative sometimes?

g11tech:
[...]

PR #3261, opened 2021-09-27
Fixing the negative cache hit artifacts in regenFn stats, collapsing the getSate metric panels

g11tech:
**Motivation**
Fixing the negative cache hit artifacts in regenFn stats, collapsing the getSate metric panels into 1 single one as per our analysis is doesn't seemed to be called from anywhere outside regen. 
plus shortning the full length cache panels to half length.

context: https://github.com/ChainSafe/lodestar/issues/3120

how the regen fn panels now look (e.g)

image

how getState is collapsed into just 1:

image

**Description**

Closes #issue_number

### DIFF
--- a/docker/grafana/provisioning/dashboards/lodestar.json
+++ b/docker/grafana/provisioning/dashboards/lodestar.json
@@ -9575,10 +9575,11 @@
         "x": 0,
         "y": 19
       },
-      "id": 193,
+      "id": 270,
       "panels": [
         {
           "datasource": null,
+          "description": "",
           "fieldConfig": {
             "defaults": {
               "color": {
@@ -9589,10 +9590,9 @@
                 "axisPlacement": "auto",
                 "barAlignment": 0,
                 "drawStyle": "line",
-                "fillOpacity": 20,
+                "fillOpacity": 0,
                 "gradientMode": "none",
                 "hideFrom": {
-                  "graph": false,
                   "legend": false,
                   "tooltip": false,
                   "viz": false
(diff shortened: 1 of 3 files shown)


dapplion:
@g11tech The screen capture shows that one row only has one panel. Please add an empty panel such that all rows have exactly 2 panels with the same width. Otherwise Grafana may bug

dapplion:
Thanks! Looks great overall. Why not add the total count metric call above this line? Then you will only need one

https://github.com/ChainSafe/lodestar/blob/027a62088870727132c2b97353a269149fae6f33/packages/lodestar/src/chain/regen/queued.ts#L141

g11tech:
> @g11tech The screen capture shows that one row only has one panel. Please add an empty panel such that all rows have exactly 2 panels with the same width. Otherwise Grafana may bug

its the last panel (the getState empty one having all its collapsed  metric targets), so wouldnt bug. but will add another empty one just the same.

g11tech:
> Thanks! Looks great overall. Why not add the total count metric call above this line? Then you will only need one
> 
> https://github.com/ChainSafe/lodestar/blob/027a62088870727132c2b97353a269149fae6f33/packages/lodestar/src/chain/regen/queued.ts#L141

thats what is causing negative artifacts, as there is a slight delay between queuing and jobqueue  executing  as well as gets spread over as jobqueue picks and executes. so cant move it in jobqueue's processing scope. this would be same as regen_fn_call_duration_count which was being used earlier.

g11tech:
discord conversation context for condensing the charts further:

image

g11tech:
sample screenshots:
image
image

g11tech:
@dapplion

dapplion (dismissed):

dapplion:
Used the histogram counter `regen_fn_call_duration_count` instead of the new `regen_fn_queued_total` since they should be identical with no added code

g11tech:
> Used the histogram counter `regen_fn_call_duration_count` instead of the new `regen_fn_queued_total` since they should be identical with no added code

@dapplion as discussed over the discord, `regen_fn_queued_total` is required as duration_count is timeshifted (as its is being processed inside job processorr). As agreed, reverted the code to original. formatting feedback included!

dapplion (approved):

Issue #3275, opened 2021-09-30
Amphora Tracking


wemeetagain:
Execution clients look here -> [Amphora README](https://github.com/ChainSafe/lodestar/tree/master/amphora)

## M1
> Spec Implemented

- [x] pass all released consensus test vectors https://github.com/ChainSafe/lodestar/issues/3076
- [x] run through engine [test vectors](https://notes.ethereum.org/@9AeMAlpyQYaAAyuj47BzRw/rkwW3ceVY) calling against geth - @dapplion 
- [x] run against [mergemock](https://github.com/protolambda/mergemock) - @tuyennhv 
- [x] Create Intro to Lodestar issue for other teams: e.g. Run these commands to start devnet with X config - @wemeetagain 
- [x] Run geth test vectors - @g11tech 

~~Add geth client into CI  @g11tech~~ (added by @dapplion )

## M2/M3
> 1-to-1 Interop – One EL and one CL team pair off for initial interop testing

- Team 1 - 
  - [x] run 1:1 devnet without transition (TTD == 0)  
  - [x] run 1:1 devnet with transition (TTD > 10)

- Team 2 - 
  - [ ] run 1:1 devnet without transition (TTD == 0)
  - [ ] run 1:1 devnet with transition (TTD > 10)

**Interoperability Status:**
*with execution*
- [x] Lodestar <> Geth  :hourglass_flowing_sand: 
- [x] Lodestar <> Besu  :hourglass_flowing_sand: 
- [ ] Lodestar <> Erigon  :hourglass_flowing_sand: 
- [x] Lodestar <> Nethermind  :hourglass_flowing_sand: 
- [x] Lodestar <> EthereumJS  :hourglass_flowing_sand: 

*with consensus*
- [ ] Lodestar <> Nimbus  :hourglass_flowing_sand: 
- [x] Lodestar <> Lighthouse  :hourglass_flowing_sand: 
- [ ] Lodestar <> Prysmatic  :hourglass_flowing_sand: 
- [x] Lodestar <> Teku  :hourglass_flowing_sand: 

## M4
> Many-to-Many Interop – 2 or more EL and CL teams perform interop testing

- [x] Run many:many devnet with transition (TTD > 50)
- [ ] Record teams and devnet config/results in separate issue

## M5
> All-to-All persistent devnet – All ready EL and CL teams create a long-lived devnet

- [x] Run many:many devnet with transition (TTD > 100)
- [x] Keep the devnet up for continuous testing and fun
[...]

PR #3281, opened 2021-10-02
Test UnionType

dapplion:
**Motivation**

Part of #3275, UnionType breaks when transactions are non-empty

**Description**

PR to show errors with UnionType

### DIFF
--- a/packages/types/test/unit/union.test.ts
+++ b/packages/types/test/unit/union.test.ts
@@ -1,0 +1,15 @@
+import {ssz} from "../../src";
+
+describe("union", () => {
+  it("struct hashTreeRoot", () => {
+    const defaultValue = ssz.merge.ExecutionPayload.defaultValue();
+    defaultValue.transactions.push(Buffer.alloc(512, 0));
+    ssz.merge.ExecutionPayload.hashTreeRoot(defaultValue);
+  });
+
+  it("tree hashTreeRoot", () => {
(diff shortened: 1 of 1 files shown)


packages/types/test/unit/union.test.ts
```diff
@@ -0,0 +1,15 @@
+import {ssz} from "../../src";
+
+describe("union", () => {
+  it("struct hashTreeRoot", () => {
+    const defaultValue = ssz.merge.ExecutionPayload.defaultValue();
+    defaultValue.transactions.push(Buffer.alloc(512, 0));
```
twoeths, line 6: for Union, value is actually a wrapped value, it needs a selector (uint) specifying the type used for that Union. For example this is from the spec test (which was passed)

```
 transactions:
 - {selector: 0, value: '0xfa61e831ac3cab767c448be074e4f555537ea08bf0341d8aba3682d6b78313d9442e2ef03905e61f65433fac    13d591b5a84af52249d7dc78dfe14276820a2ed6bd164e8f6e49abb32a05ce809d772a6233977db550da180fc4f74d54d7847895568963db44    1a6745ffa3197daef756976993185af5bafc7518f34799cac1d2a1ca97fc7fd2effd36841ac418ef8826b8a5494961dc0b068742f2d04e873c    f512bcb02c748eafc1e6'}
```

this works
```
// @ts-ignore
    defaultValue.transactions.push({selector: 0, value: Buffer.alloc(512, 0)});
```

twoeths:
created #3285 to correct the merge Transaction type

PR #3288, opened 2021-10-02
ssz package update with casing map supplied for slashing containers

g11tech:
**Motivation**
This PR updates ssz package to `0.8.18` and with that allows the ability to provide declaration time casing map for resolving casing mismatch especially in number involving field transformations (e.g. `signed_header_1` <=> `signedHeader1`) as well as super fast lookup whenever the `casingMap` is provided.

The slashing related `AttesterSlashing` and `ProposerSlashing` container have been updated as first and foremost as their data consumption would throw an error without this fix.
The tests for serialization destabilization of above containers have been added for `toJson`/`fromJson` fields search.

**Description**
Post this PR, containers involved in rest api module will be targeted for providing `casingMap`.

Closes #issue_number


packages/types/test/unit/ssz.test.ts
```diff
@@ -0,0 +1,54 @@
...
+import {ssz} from "../../src";
+
+describe("size", function () {
+  it("should calculate correct minSize and maxSize", () => {
+    const minSize = ssz.phase0.BeaconState.minSize();
+    const maxSize = ssz.phase0.BeaconState.maxSize();
+    // https://gist.github.com/protolambda/db75c7faa1e94f2464787a480e5d613e
+    expect(minSize).to.be.equal(2687377);
+    expect(maxSize).to.be.equal(141837543039377);
+  });
```
dapplion, line 11: This test is not necessary here

packages/types/test/unit/ssz.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const result = ssz.phase0.ProposerSlashing.fromJson(json, {
+      case: "snake",
+    });
+    expect(ssz.phase0.ProposerSlashing.equals(test, result)).to.be.true;
+    const back = ssz.phase0.ProposerSlashing.toJson(result, {
+      case: "snake",
+    });
+    expect(back).to.be.deep.equal(json);
+  });
```
dapplion, line 53: Could you add also tests and casing for `eth1_data` and all `eth1_*` fields? Check https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md

packages/types/test/unit/ssz.test.ts
g11tech: that was an old test (renamed size.test.ts to this file) or i can just restore size.test.ts and remove this from here.

packages/types/test/unit/ssz.test.ts
g11tech: in this PR or new one?
eth1_ fields should already work fine, though I intend to add casingMap (and relevant tests) to all the containers which serialize/de-serialize from data, just for speed of to/from json even if the casing works without casingMap

packages/types/test/unit/ssz.test.ts
dapplion: Yes please remove, this test may go stale

dapplion (approved):

PR #3294, opened 2021-10-03
Notify execution client on fork choice head update

wemeetagain:
**Motivation**

Merge interop

**Description**

We should notify the execution client when our head updates


dapplion (approved):
Good find!

PR #3295, opened 2021-10-03
a test script for checking multiple execution client merge api spec

g11tech:
**Motivation**
This PR reimagines the `merge-interop.test.ts` as `merge-interop-multi.test.ts` which can be interfaced with a number of available execution clients for running  tests and may be going forward running sims against.
depending on the enviornment variable the appropriate client can be interfaced, especially outside CI where docker enviornment is available allowing any execution client to be easily interfaced.
 

**Description**
Amphora #3275

with dockerized geth (& skipped binary geth)
image

Closes #issue_number


dapplion:
If I understand the PR, it currently only interfaces with Geth right? It would be nice to extend it to actually interface with other clients, i.e. keep current master for Geth and add other clients like Nethermind, Besu. De-duplicating code is not really important yet, but nice if possible without sacrificing test clarity.

g11tech:
> If I understand the PR, it currently only interfaces with Geth right? It would be nice to extend it to actually interface with other clients, i.e. keep current master for Geth and add other clients like Nethermind, Besu. De-duplicating code is not really important yet, but nice if possible without sacrificing test clarity.

thats what I was going for (though didnt get the chance to clean this up ), linkup with all dockerized execution clients (this one links up with dockerized geth)
How about I  keep this multi dockerized client test file separate (like as in the current draft), and move the test cases to utils as calling them up is a clear abstraction point? 
if you like the abstraction will extend it to other the normal merge file.

g11tech:
addressed in https://github.com/ChainSafe/lodestar/pull/3436

PR #3301, opened 2021-10-04
More sim tests for merge

twoeths:
**Motivation**

Part of #3275 

**Description**

+ Refactor eth1 option to "rpcClient" or "mock" (new) or "disabled"
+ Mock the Eth1Provider with 4 options
```typescript
export type Eth1ProviderMockOpts = {
  startDifficulty: number;
  difficultyIncrement: number;
  mergeBlockDifficulty: number;
  mergeBlockHash: string;
};
```
+ mergeBlockHash is the same to the `genesisBlockHash` option of `ExecutionEngineMock` in order for the mock execution engine to create payload from a known parent hash
+ More tests for merge: MERGE_EPOCH = 1, TERMINAL_TOTAL_DIFFICULTY = 0 and MERGE_EPOCH = 2, TERMINAL_TOTAL_DIFFICULTY = 10 
+ Assertion for the delta/offset between the consensus block slot and the eth1 block number in the payload
+ Assertion for the found merge block

### DIFF
--- a/packages/cli/src/options/beaconNodeOptions/eth1.ts
+++ b/packages/cli/src/options/beaconNodeOptions/eth1.ts
@@ -2,7 +2,7 @@ import {defaultOptions, IBeaconNodeOptions} from "@chainsafe/lodestar";
 import {ICliCommandOptions} from "../../util";
 
 export interface IEth1Args {
-  "eth1.enabled": boolean;
+  "eth1.mode": "rpcClient" | "disabled";
   "eth1.providerUrl": string;
   "eth1.providerUrls": string[];
   "eth1.depositContractDeployBlock": number;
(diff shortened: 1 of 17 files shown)


packages/cli/src/options/beaconNodeOptions/eth1.ts
```diff
@@ -2,10 +2,14 @@ import {defaultOptions, IBeaconNodeOptions} from "@chainsafe/lodestar";
 import {ICliCommandOptions} from "../../util";
 
 export interface IEth1Args {
-  "eth1.enabled": boolean;
+  "eth1.mode": "rpcClient" | "mock" | "disabled";
   "eth1.providerUrl": string;
   "eth1.providerUrls": string[];
   "eth1.depositContractDeployBlock": number;
+  "eth1.startDifficulty": number;
+  "eth1.difficultyIncrement": number;
```
dapplion: This params are only used for merge mock testing, should not be part of the regular options. Can you find a different strategy to customize this without affecting the production path?

dapplion: A potential strategy would be to serve the data in a real JSON RPC server, that way to code would be more real but still super fast

twoeths: I see your point 👍 
in this case, the cli should not support "mock" option of eth1 and then the the prod options are clean

PR #3308, opened 2021-10-05
Fix notifyConsensusValidated

twoeths:
**Motivation**

+ part of #3275

**Description**

+ when we failed to get through state transition, we want to `notifyConsensusValidated` as `false` to execution engine
+ also in executionEngine mock, delete from `preparingPayloads` in `getPayload()` as we don't need that anymore after this method


wemeetagain (approved):
:eyes: Good catch!

dapplion (approved):
Good pick! Thanks

Issue #3276, opened 2021-10-01
Multiple epoch transitions per epoch


twoeths:
**Describe the bug**

Right now for each epoch, we have 3 or 4 or 5 epoch transitions

<img width="807" alt="Screen Shot 2021-10-01 at 17 22 08" src="https://user-images.githubusercontent.com/10568965/135605003-4792e687-01c7-40df-b1f0-81e566255d21.png">

The checkpoint state cache shows that we have no more than 3 states per epoch

<img width="791" alt="Screen Shot 2021-10-01 at 17 36 47" src="https://user-images.githubusercontent.com/10568965/135606603-3075b861-5b39-493a-b0f5-5949a050191a.png">

**Expected behavior**

1. There are 1 or 2 redundant epoch transitions per epoch as we can simply get it from the cache
2. Investigate why we have up to 3 epoch transitions per epoch? Ideally most of the time we should have exactly 1 epoch transition per epoch.

At least when getting attester duties, we may not need to rotate to the current epoch (most of the time) thanks to the `nextEpochShuffling` cache https://github.com/ChainSafe/lodestar/blob/96ca4d319b386f1c448defe236eba7b7370b2cbe/packages/lodestar/src/api/impl/validator/index.ts#L283

dapplion:
@tuyennhv The first chart screenshot can be very miss-leading, three dots don't mean 3 epoch transitions. You should look into the actual count of epoch transitions and look at the step delta.

dapplion:
![Screenshot from 2021-10-02 05-48-06](https://user-images.githubusercontent.com/35266934/135702698-7dd8f541-c311-470a-8e88-f6aaec5eeb2e.png)

This is Contabo-5 and here it's irrefutable that is most of the epochs 2 epoch transitions are run

twoeths:
@dapplion do you think 2 epoch transitions per epoch is good enough? In my understanding, it should be 1 most of the time

dapplion:
> @dapplion do you think 2 epoch transitions per epoch is good enough? In my understanding, it should be 1 most of the time

No, 2 is very bad! hahaha It would be very strange that there a re-org always at the epoch boundary, we must be doing something wrong

twoeths:
I was able to catch the additional non-regular epoch transition from https://github.com/ChainSafe/lodestar/blob/98144ce2ad5df5208eeb941fc81d6ae029cc25d4/packages/lodestar/src/chain/validation/aggregateAndProof.ts#L64

[...]

PR #3310, opened 2021-10-05
Epoch transition count metric

twoeths:
**Motivation**

+ We want to track epoch transition count while the current panel is misleading

**Description**
+ It clearly shows that we have 2 epoch transitions per epoch

Screen Shot 2021-10-05 at 11 20 55

part of #3276

### DIFF
--- a/docker/grafana/provisioning/dashboards/lodestar.json
+++ b/docker/grafana/provisioning/dashboards/lodestar.json
@@ -3520,7 +3520,7 @@
           "pluginVersion": "7.4.5",
           "targets": [
             {
-              "exemplar": true,
+              "exemplar": false,
               "expr": "lodestar_stfn_epoch_transition_seconds_count",
               "interval": "",
               "legendFormat": "number of epoch transition",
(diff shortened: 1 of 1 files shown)


docker/grafana/provisioning/dashboards/lodestar.json
```diff
@@ -12232,15 +12233,15 @@
...
+  "refresh": false,
   "schemaVersion": 30,
   "style": "dark",
   "tags": [],
   "templating": {
     "list": []
   },
   "time": {
-    "from": "now-12h",
+    "from": "now-1h",
```
dapplion: Don't modify the time window

docker/grafana/provisioning/dashboards/lodestar.json
```diff
@@ -12232,15 +12233,15 @@
       "type": "row"
     }
   ],
-  "refresh": "30s",
+  "refresh": false,
```
dapplion: Leave as 30s refresh

dapplion (approved):

PR #3320, opened 2021-10-06
Add connect and disconect peer routes

dapplion:
**Motivation**

Very useful for testnet debugging when discv5 only returns useless peers

**Description**

Add connect and disconect peer routes


packages/api/src/routes/lodestar.ts
```diff
@@ -94,6 +99,8 @@ export const routesData: RoutesData<Api> = {
   getCheckpointStateCacheItems: {url: "/eth/v1/lodestar/checkpoint-state-cache-items", method: "GET"},
   runGC: {url: "/eth/v1/lodestar/gc", method: "POST"},
   dropStateCache: {url: "/eth/v1/lodestar/drop-state-cache", method: "POST"},
+  connectPeer: {url: "/eth/v1/node/connect_peer", method: "GET"},
```
twoeths: should it be `POST` instead?

dapplion: Sounds good

twoeths (approved):

PR #3322, opened 2021-10-06
Add port option

dapplion:
**Motivation**

Changing the port is a UX nightmare

**Description**

Add port option

### DIFF
--- a/packages/cli/src/cmds/beacon/options.ts
+++ b/packages/cli/src/cmds/beacon/options.ts
@@ -16,7 +16,7 @@ interface IBeaconExtraArgs {
 
 export const beaconExtraOptions: ICliCommandOptions<IBeaconExtraArgs> = {
   port: {
-    description: "P2P port for discv5 and lipbp",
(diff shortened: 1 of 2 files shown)


packages/cli/src/cmds/beacon/options.ts
```diff
@@ -14,6 +15,13 @@ interface IBeaconExtraArgs {
 }
 
 export const beaconExtraOptions: ICliCommandOptions<IBeaconExtraArgs> = {
+  port: {
+    description: "P2P port for discv5 and lipbp",
```
wemeetagain: pbp

wemeetagain:
would like to see these cli flags:
```
--discovery-port <PORT>
            The UDP port that discovery will listen on. Defaults to `port`

--port <PORT>
            The TCP/UDP port to listen on. The UDP port can be modified by the --discovery-port flag. [default:
            9000]
```

wemeetagain (approved):
:100: :heart:

PR #3328, opened 2021-10-06
If minimal preset log on start

dapplion:
**Motivation**

Have a log line to ensure preset is correctly set

**Description**

If minimal preset log on start


wemeetagain (approved):

packages/cli/src/cmds/dev/handler.ts
```diff
@@ -66,6 +67,7 @@ export async function devHandler(args: IDevArgs & IGlobalArgs): Promise<void> {
   const libp2p = await createNodeJsLibp2p(peerId, options.network);
   const logger = getCliLogger(args, beaconPaths, config);
   logger.info("Lodestar", {version: getVersion(), network: args.network});
+  if (ACTIVE_PRESET === PresetName.minimal) logger.info("ACTIVE_PRESET == minimal preset");
```
wemeetagain, line 70: maybe if ACTIVE_PRESET !== mainnet, print

PR #3330, opened 2021-10-06
Implement optimistic sync post-merge

dapplion:
**Motivation**

It's okay to ignore SYNCING status because:
- We MUST verify execution payloads of blocks we attest
- We are NOT REQUIRED to check the execution payload of blocks we don't attest. 

When EL syncs from genesis to a chain post-merge, it doesn't know what the head, CL knows. However, we must verify (complete this fn) and import a block to sync. Since we are syncing we only need to verify consensus and trust that whatever the chain agrees is valid, is valid; no need to verify. When we verify consensus up to the head we notify forkchoice update head and then EL can sync to our head. At that point regular EL sync kicks in and it does verify the execution payload (EL blocks). If after syncing EL gets to an invalid payload or we can prepare payloads on what we consider the head that's a critical error

**Description**

Accept SYNCING status on execution payload verification only when we are syncing

### DIFF
--- a/packages/lodestar/src/executionEngine/http.ts
+++ b/packages/lodestar/src/executionEngine/http.ts
@@ -67,14 +67,13 @@ export class ExecutionEngineHttp implements IExecutionEngine {
       params: [serializeExecutionPayload(executionPayload)],
     });
 
-    switch (status) {
-      case ExecutePayloadStatus.VALID:
-        return ExecutePayloadStatus.VALID;
-      case ExecutePayloadStatus.INVALID:
-        return ExecutePayloadStatus.INVALID;
-      case ExecutePayloadStatus.SYNCING:
-        return ExecutePayloadStatus.SYNCING;
+    // Validate status is known
+    const statusEnum = ExecutePayloadStatus[status];
+    if (statusEnum === undefined) {
+      throw Error(`Unknown status ${status}`);
     }
+
+    return statusEnum;
   }
 
   /**
--- a/packages/lodestar/src/executionEngine/http.ts
+++ b/packages/lodestar/src/executionEngine/http.ts
@@ -12,7 +12,7 @@ import {
   QUANTITY,
 } from "../eth1/provider/utils";
 import {IJsonRpcHttpClient} from "../eth1/provider/jsonRpcHttpClient";
-import {IExecutionEngine, PayloadId} from "./interface";
+import {ExecutePayloadStatus, IExecutionEngine, PayloadId} from "./interface";
 import {BYTES_PER_LOGS_BLOOM} from "@chainsafe/lodestar-params";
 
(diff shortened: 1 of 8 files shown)


g11tech (dismissed):

packages/lodestar/src/executionEngine/http.ts
```diff
@@ -67,8 +67,14 @@ export class ExecutionEngineHttp implements IExecutionEngine {
       params: [serializeExecutionPayload(executionPayload)],
     });
 
-    // TODO: Handle invalid status
-    return status === ExecutePayloadStatus.VALID;
+    switch (status) {
```
twoeths: I wonder why we have to use `switch` here

twoeths (dismissed):
lgtm, just drop a minor comment

wemeetagain:
Problems for other PRs:
This is fine for now, should be extended in various ways:
- we should track last validated block or otherwise limit attesting/proposing/serving peers from unvalidated blocks. Note, this validation can happen one of two ways, as described in this PR:
- 1. validated by consensus (so all finalized slots are valid)
- 2. fully validated by our node (so all blocks, and their ancestors, validated by our state transition, _including execution_ are valid)

This PR will over-optimistically validate blocks during the initial sync because we have no notion/marker/pointer/etc. of fully validated vs partially validated blocks. So the problem presents itself during finalized sync when we would serve unfinalized and unvalidated blocks and during unfinalized sync to head, when we would serve/attest/propose unfinalized and unvalidated blocks.

packages/lodestar/src/executionEngine/http.ts
wemeetagain: Looks like it should be just `return status`

wemeetagain (changes requested):

dapplion:
> Problems for other PRs: This is fine for now, should be extended in various ways:
> 
>     * we should track last validated block or otherwise limit attesting/proposing/serving peers from unvalidated blocks. Note, this validation can happen one of two ways, as described in this PR:
> 
>     * 1. validated by consensus (so all finalized slots are valid)
> 
>     * 1. fully validated by our node (so all blocks, and their ancestors, validated by our state transition, _including execution_ are valid)
> 
> 
> This PR will over-optimistically validate blocks during the initial sync because we have no notion/marker/pointer/etc. of fully validated vs partially validated blocks. So the problem presents itself during finalized sync when we would serve unfinalized and unvalidated blocks and during unfinalized sync to head, when we would serve/attest/propose unfinalized and unvalidated blocks.

It's even more complicated than that. Developing a resilient sync strategy will take significant though and I think its implementation is out of scope now. We **need** this PR to participate in the merge interop so we must accept progressive increased security against edge cases

twoeths (approved):

wemeetagain (approved):

PR #3331, opened 2021-10-07
Check genesisTime when starting validator

### LINKED ISSUE #3326: Track genesisTime in validator's metadata db bucket
**Describe the bug**

Track genesisTime in validator's metadata bucket

https://github.com/ChainSafe/lodestar/blob/9221c29efed52dd136dfbecfbdcc67926f33b647/packages/validator/src/validator.ts#L98

In merge interop devnets the genesisValidatorsRoot was constant but we just rolled the genesisTime forward.

twoeths:
**Motivation**

We want to check both genesis validator root and genesis time when starting validator to make sure they are the same to node's values.

**Description**

Closes #3326


dapplion (approved):
Looks great!

wemeetagain (approved):

Issue #3336, opened 2021-10-08
refactor: add casingMap in all the ssz types as per the spec


g11tech:
Lodestar creates container objects in the `camel` casing with to and fro with 'snake` on serialization/deserialization. This transformation happens via ssz `toJson`/`fromJson` calls.
Aim is to provide `casingMap`  to all the ssz containers so that computation is saved as well as there is no room any discrepancy with the field's in the container spec: https://github.com/ethereum/consensus-specs/

PR #3351, opened 2021-10-11
casingMap update for all ssz containers

g11tech:
**Motivation**
This PR adds casingMap to all the ssz containers both for the reasons of
1. Being verbose in stating the matching of container fields to the field names in specs expected in data apis and not leaving to doubt the library based conversion
2. casingMap speeds up the case conversion in toJson/fromJson by skipping the actual case conversion call
3. corrected case conversion of `subCommitteeIndex` to the `subcommittee_index` as specified in spec. earlier  the case conversion library  was  **erroneously** converting it to `sub_committee_index` potentially leading to errors and bad performance.
4. The `PowBlock` ssz type has been corrected to `merge.PowBlock`. earlier it was `merge.BeaconState`. 

**Description**
This PR labels if the container casing is from beacon apis or custom made (there are couple of the custom fields). Rest were all found and matched from the consensus specs.

Closes #issue_number
#3336

### DIFF
--- a/packages/api/src/routes/beacon/state.ts
+++ b/packages/api/src/routes/beacon/state.ts
@@ -232,6 +232,12 @@ export function getReturnTypes(): ReturnTypes<Api> {
       currentJustified: ssz.phase0.Checkpoint,
       finalized: ssz.phase0.Checkpoint,
     },
+    // From beacon apis
+    casingMap: {
+      previousJustified: "previous_justified",
+      currentJustified: "current_justified",
+      finalized: "finalized",
+    },
   });
 
   const ValidatorResponse = new ContainerType<ValidatorResponse>({
@@ -241,6 +247,8 @@ export function getReturnTypes(): ReturnTypes<Api> {
       status: new StringType<ValidatorStatus>(),
       validator: ssz.phase0.Validator,
     },
+    // From beacon apis
+    expectedCase: "notransform",
   });
 
   const ValidatorBalance = new ContainerType<ValidatorBalance>({
@@ -248,6 +256,8 @@ export function getReturnTypes(): ReturnTypes<Api> {
       index: ssz.ValidatorIndex,
       balance: ssz.Number64,
     },
+    // From beacon apis
+    expectedCase: "notransform",
   });
 
   const EpochCommitteeResponse = new ContainerType<EpochCommitteeResponse>({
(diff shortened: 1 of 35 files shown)


packages/params/src/preset/altair/ssz.ts
```diff
@@ -15,3 +15,9 @@ export const AltairPreset = new ContainerType<IAltairPreset>({
     PROPORTIONAL_SLASHING_MULTIPLIER_ALTAIR: Number64,
   },
 });
+
+//expected and container fields are the same here
+AltairPreset.casingMap = Object.keys(AltairPreset.fields).reduce((obj: Record<string, string>, key) => {
```
dapplion: Can you add a space before the first character of the comment for better readability?

packages/types/src/altair/sszTypes.ts
```diff
@@ -45,17 +44,24 @@ export const SyncSubnets = new BitVectorType({
 
 export const Metadata = new ContainerType<altair.Metadata>({
   fields: {
-    seqNumber: Uint64,
-    attnets: phase0Ssz.AttestationSubnets,
+    ...phase0Ssz.Metadata.fields,
```
dapplion, line 48: Since order is critical in SSZ containers I would not spread them. It's just a few extra lines and ensures the reader the implementation is correct

EDIT: We do that in the block type, if the new key is the last then it's okay. Maybe add a comment?
```
// New keys are strictly appended, phase0 key order is preserved
```

packages/api/src/routes/beacon/state.ts
```diff
@@ -241,13 +247,25 @@ export function getReturnTypes(): ReturnTypes<Api> {
       status: new StringType<ValidatorStatus>(),
       validator: ssz.phase0.Validator,
     },
+    //from beacon apis
+    casingMap: {
+      index: "index",
+      balance: "balance",
+      status: "status",
+      validator: "validator",
+    },
```
dapplion: With the goal of doing the minimal work possible, could you find a way to skip doing casingMap transformations in the casing is exactly the same?

g11tech: a reduce code segment like in Phase0Preset.casingMap?

dapplion:
@g11tech Feel free to trigger a release on SSZ, just merged the PR there

dapplion (dismissed):
Looks great! Thank you so much :fire: Could you now try to connect Lighthouse validator to a Lodestar beacon node and a Lodestar validator to a Lighthouse beacon node to ensure we are compatible?

g11tech:
> Looks great! Thank you so much fire Could you now try to connect Lighthouse validator to a Lodestar beacon node and a Lodestar validator to a Lighthouse beacon node to ensure we are compatible?

sounds good :+1:

dapplion:
@g11tech conflicts

g11tech:
@dapplion rebased :+1:

dapplion (approved):

PR #3356, opened 2021-10-12
Fix bootEnrs cli issue

twoeths:
**Motivation**

+ When no network is specified, "mainnet" is used
+ In amphora environment, even we specify `network.discv5.bootEnrs` it still gets the bootEnrs from mainnet and discv5 sends lodestar wrong peers all the times

**Description**

+ Only download bootEnrs from the network if cli does not specify its own `network.discv5.bootEnrs` arguments

**Note**
+ There is a TODO found in this PR, will open a separate issue for it.


dapplion (approved):
Looks great! Thanks

PR #3364, opened 2021-10-14
Strict boolean expression

### LINKED ISSUE #3324: Disallow if coercion
**Describe the bug**

Coercing values in if statements can be problematic if the variable can be a 0 which should be truthy:

```ts
let num: number | null;
if (num) {}
```

Require

```ts
let num: number | null;
if (num !== null && num > 0) {}
```

Check: https://eslint.org/docs/rules/no-implicit-coercion

twoeths:
**Motivation**

+ For if clause, we want to treat 0 and `null | undefined` differently case by case

**Description**

+ Apply `strict-boolean-expression` fule for nullable number (will consider applying for nullable string and nullable boolean later if we find it as an issue)
+ There is no configured rule for object so instead of `if (obj)` we have to do something like `if (obj !== undefined)`

Closes #3324

### DIFF
--- a/packages/api/src/utils/urlFormat.ts
+++ b/packages/api/src/utils/urlFormat.ts
@@ -23,7 +23,7 @@ export function compileRouteUrlFormater(path: string): (arg: Args) => string {
     const currentToken: Token | undefined = tokens[tokens.length - 1];
     switch (path[i]) {
       case ":": {
-        if (currentToken && currentToken.type === TokenType.Variable) {
+        if (currentToken !== undefined && currentToken.type === TokenType.Variable) {
           throw Error(`Invalid path token ':' not closed: ${path}`);
         }
         tokens.push({type: TokenType.Variable, start: i});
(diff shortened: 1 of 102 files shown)


dapplion (changes requested):
Looks great! Thanks for auditing all existing expressions

dapplion (approved):
Looks great! Thanks for auditing all existing expressions

packages/api/src/utils/urlFormat.ts
```diff
@@ -23,22 +23,22 @@ export function compileRouteUrlFormater(path: string): (arg: Args) => string {
     const currentToken: Token | undefined = tokens[tokens.length - 1];
     switch (path[i]) {
       case ":": {
-        if (currentToken && currentToken.type === TokenType.Variable) {
+        if (currentToken !== undefined && currentToken.type === TokenType.Variable) {
```
filipesmedeiros, line 26: wait why not `currentToken?.type`?

dapplion: Good suggestion! Please do a PR if you want :+1:

Issue #3370, opened 2021-10-15
validator interop: investigate and fix the produceBlock api not working against lighthouse beacon


g11tech:
master issue: #3368

the `produceBlock` api is not working against the lighthouse beacon:
![image](https://user-images.githubusercontent.com/76567250/137482456-03ce3080-180e-4558-a035-ae19fd26ee22.png)
investigate/fix

PR #3376, opened 2021-10-18
publish block fix for validator interoperability with lighthouse beacon

g11tech:
**Motivation**
This is to make the lodestar validator interop with lighthouse beacon

**Description**
The lodestar validator was throwing errors on the `produceBlock` api response as documented in isssue: #3370 . 

~~Reason was that the lodestar validator assumes the version  information (`phase0`, `altair`) availability in response which was not being provided by lighthouse beacon as its not the part of the spec. This PR adds the ability to extract the version from the slot to get the version.~~

making `WithVersion` back compatible with pre altair `v1/produceBlock` call

The `publishBlock` api is now being properly consumed and the validator is able to produce blocks.
image

Closes #issue_number
#3370

### DIFF
--- a/packages/api/src/utils/types.ts
+++ b/packages/api/src/utils/types.ts
@@ -128,10 +128,15 @@ export function WithVersion<T>(getType: (fork: ForkName) => TypeJson<T>): TypeJs
       data: getType(version || ForkName.phase0).toJson(data, opts),
       version,
     }),
-    fromJson: ({data, version}: {data: Json; version: string}, opts) => ({
-      data: getType((version as ForkName) || ForkName.phase0).fromJson(data, opts),
-      version: version as ForkName,
-    }),
+    fromJson: ({data, version}: {data: Json; version: string}, opts) => {
+      // Un-safe external data, validate version is known ForkName value
+      if (!ForkName[version as ForkName]) throw Error(`Invalid version ${version}`);
+
+      return {
+        data: getType(version as ForkName).fromJson(data, opts),
+        version: version as ForkName,
+      };
+    },
   };
 }
 
--- a/packages/api/src/utils/types.ts
+++ b/packages/api/src/utils/types.ts
@@ -125,11 +125,11 @@ export function ContainerData<T>(dataType: Type<T>): ContainerType<{data: T}> {
 export function WithVersion<T>(getType: (fork: ForkName) => TypeJson<T>): TypeJson<{data: T; version: ForkName}> {
   return {
     toJson: ({data, version}, opts) => ({
-      data: getType(version).toJson(data, opts),
+      data: getType(version || ForkName.phase0).toJson(data, opts),
       version,
     }),
     fromJson: ({data, version}: {data: Json; version: string}, opts) => ({
-      data: getType(version as ForkName).fromJson(data, opts),
+      data: getType((version as ForkName) || ForkName.phase0).fromJson(data, opts),
       version: version as ForkName,
     }),
   };
--- a/packages/api/src/routes/validator.ts
+++ b/packages/api/src/routes/validator.ts
@@ -123,11 +123,7 @@ export type Api = {
    * @returns any Success response
    * @throws ApiError
    */
-  produceBlock(
-    slot: Slot,
-    randaoReveal: BLSSignature,
-    graffiti: string
-  ): Promise<{data: allForks.BeaconBlock; version: ForkName}>;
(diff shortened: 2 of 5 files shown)


dapplion:
@g11tech It is part of the REST API spec! Can you check with Lighthouse devs if they are not adding this info? If not, you should raise an issue on their side.

dapplion:
@g11tech According to their source code Lighthouse is adding the fork version to the response:

```rs
                            let fork_name = block.fork_name(&chain.spec).ok();
                            fork_versioned_response(endpoint_version, fork_name, block)
                                .map(|res| warp::reply::json(&res).into_response())
```

https://github.com/sigp/lighthouse/blob/fff01b24ddedcd54486e374460855ca20d3dd232/beacon_node/http_api/src/lib.rs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L1019

g11tech:
> @g11tech According to their source code Lighthouse is adding the fork version to the response:
> 
> ```rust
>                             let fork_name = block.fork_name(&chain.spec).ok();
>                             fork_versioned_response(endpoint_version, fork_name, block)
>                                 .map(|res| warp::reply::json(&res).into_response())
> ```
> 
> https://github.com/sigp/lighthouse/blob/fff01b24ddedcd54486e374460855ca20d3dd232/beacon_node/http_api/src/lib.rs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L1019

@dapplion  actually on the testnet, when slots start from the genesis, till altair fork epoch, v1/produceBlock is being called instead of v2/productBlock by our validator:
https://github.com/ChainSafe/lodestar/blob/master/packages/validator/src/services/block.ts#L76:L80
and v1 doesn't have the version type.
a very simple approach would be that in `WithVersion`, `getType(version)` can be replaced by `getType(version||'phase0')`

packages/api/src/utils/types.ts
```diff
@@ -125,11 +125,11 @@ export function ContainerData<T>(dataType: Type<T>): ContainerType<{data: T}> {
 export function WithVersion<T>(getType: (fork: ForkName) => TypeJson<T>): TypeJson<{data: T; version: ForkName}> {
   return {
     toJson: ({data, version}, opts) => ({
-      data: getType(version).toJson(data, opts),
+      data: getType(version || ForkName.phase0).toJson(data, opts),
```
dapplion, line 128: Why default to phase0 specifically? Some devnets may start genesis at a different fork the phase0

g11tech: true, but validator's `produceBlock` (https://github.com/ChainSafe/lodestar/blob/master/packages/validator/src/services/block.ts#L76:L80) will then call v2/produceBlock api end point, and that would have the version

Or I can revert/force push to the previous commit where an extractVersion fn option was passed.

dapplion:
Oh the root of the problem is that our API is not following the spec. The route produceBlock does not include version. https://github.com/ethereum/beacon-APIs/blob/295ce41a45e297aa21bea5d1695b3060066320e0/apis/validator/block.yaml?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L39

I proposed this solution, please take a look and adopt it in your PR if you agree https://github.com/ChainSafe/lodestar/compare/master...linterop

g11tech:
> Oh the root of the problem is that our API is not following the spec. The route produceBlock does not include version. https://github.com/ethereum/beacon-APIs/blob/295ce41a45e297aa21bea5d1695b3060066320e0/apis/validator/block.yaml?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L39
> 
> I proposed this solution, please take a look and adopt it in your PR if you agree [master...linterop](https://github.com/ChainSafe/lodestar/compare/master...linterop)

yes, separate serialization for `produceBlock` makes sense. will incorporate the suggestion and bring it to a working stage! :+1:

g11tech:
@dapplion  now the validator is nicely publishing blocks on the lighthouse beacon!

image

dapplion (approved):
🚀🚀🙏🙏🎉🎉

Issue #3369, opened 2021-10-15
validator interop: a strategy to autofill non operational fields in spec/config api


g11tech:
master issue: #3368

Since our development pace/branch strategy can be different from the others, we need a way to auto fix the spec mismatch for those fields

for e.g. `TERMINAL_TOTAL_DIFFICULTY` here is a merge milestone field not yet operational

![image](https://user-images.githubusercontent.com/76567250/137481833-fec33c74-0934-4e64-8919-57ccd8a8b828.png)

PR #3379, opened 2021-10-19
fill the fields locally that might not be available in the interop spec config api

g11tech:
**Motivation**
When we interop with other clients like lighthouse, their spec api might not have fields that we expect as they might not have been operational, for e.g. those belonging to the merge spec.
By populating those fields in the `--paramsFile` file with appropriate values , this PR allows them to be filled in the spec, and hence let the validator run in an interop fashion.

**Description**
Example the current branch of lighthouse doesnt provides: `TERMINAL_TOTAL_DIFFICULTY`, `MERGE_FORK_VERSION`, `SHARDING_FORK_VERSION` ...
defining them in a params config file and using them via `--paramsFile`
image
allows us to supply those fields to the validator and move forward.

Closes #issue_number
https://github.com/ChainSafe/lodestar/issues/3369

### DIFF
--- a/packages/api/src/utils/types.ts
+++ b/packages/api/src/utils/types.ts
@@ -150,12 +150,3 @@ export function sameType<T>(): TypeJson<T> {
     fromJson: (json) => (json as unknown) as T,
   };
 }
-
-/** Helper to fill missing fields that might not been activated, and thus can be missing
- * especially in interop with other client beacons */
-export function withJsonFilled<T>(dataType: Type<T>, fillWith: Json): TypeJson<{data: T}> {
-  return {
-    toJson: ({data}, opts) => ({data: dataType.toJson(data, opts)}),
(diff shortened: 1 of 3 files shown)


packages/api/src/utils/types.ts
```diff
@@ -150,3 +150,12 @@ export function sameType<T>(): TypeJson<T> {
     fromJson: (json) => (json as unknown) as T,
   };
 }
+
+/** Helper to fill missing fields that might not been activated, and thus can be missing
+ * especially in interop with other client beacons */
+export function withJsonFilled<T>(dataType: Type<T>, fillWith: Json): TypeJson<{data: T}> {
```
wemeetagain: Might be more appropriate to inline the logic here into the `getSpec` value, as this logic is not quite reusable (specialized to the getSpec endpoint case, a shallow 'fill').

wemeetagain (approved):

dapplion (approved):

PR #3383, opened 2021-10-21
Precompute epoch transition

### LINKED ISSUE #3382: Early epoch transition when node is synced

When node is synced, we want to do precompute epoch transition so that:
+ attesters can vote for head block on time
+ proposer can submit for block on time

twoeths:
**Motivation**

+ We want to have early epoch transition when the node is synced

**Description**

+ 1/3 slot before epoch boundary, run epoch transition
Closes #3382

**Test**
+ "Gossip Block" metrics in contabo-1 shows good result
Screen Shot 2021-10-21 at 18 08 02

### DIFF
--- a/packages/lodestar/src/chain/precomputeEpoch.ts
+++ b/packages/lodestar/src/chain/precomputeEpoch.ts
@@ -53,7 +53,8 @@ export class PrecomputeEpochScheduler {
     // we want to make sure headSlot === clockSlot to do early epoch transition
     const nextSlot = clockSlot + 1;
     this.logger.verbose("Precompute epoch transition", {nextEpoch, headSlot, nextSlot});
-    await this.chain.regen.getBlockSlotState(blockRoot, nextSlot, RegenCaller.preComputeEpoch).catch((e) => {
+    // this takes ~2s as of Oct 2021, no need to wait for this or the clock drift
+    this.chain.regen.getBlockSlotState(blockRoot, nextSlot, RegenCaller.preComputeEpoch).catch((e) => {
       this.logger.error("Failed to precompute epoch transition", nextEpoch, e);
     });
   };
--- a/packages/lodestar/src/chain/precomputeEpoch.ts
+++ b/packages/lodestar/src/chain/precomputeEpoch.ts
@@ -1,0 +1,54 @@
+import {AbortSignal} from "@chainsafe/abort-controller";
+import {computeStartSlotAtEpoch} from "@chainsafe/lodestar-beacon-state-transition";
+import {IBeaconConfig} from "@chainsafe/lodestar-config";
+import {SLOTS_PER_EPOCH} from "@chainsafe/lodestar-params";
+import {Epoch} from "@chainsafe/lodestar-types";
+import {ILogger, sleep} from "@chainsafe/lodestar-utils";
+import {ChainEvent} from "./emitter";
+import {IBeaconChain} from "./interface";
+import {RegenCaller} from "./regen";
+
+/**
+ * When node is synced and 1/3 slot before an epoch, we want to prepare for the next epoch
+ * transition from our head so that:
+ * + validators vote for block head on time through attestation
+ * + validators propose blocks on time
+ */
+export class PrecomputeEpochScheduler {
+  constructor(
+    private readonly chain: IBeaconChain,
+    private readonly config: IBeaconConfig,
+    private readonly logger: ILogger,
+    private readonly signal: AbortSignal
+  ) {
+    this.chain.emitter.on(ChainEvent.clockEpoch, this.onEpoch);
+
+    this.signal.addEventListener(
+      "abort",
+      () => {
+        this.chain.emitter.off(ChainEvent.clockEpoch, this.onEpoch);
+      },
+      {once: true}
+    );
+  }
+
+  private onEpoch = async (epoch: Epoch): Promise<void> => {
+    // Precalculate epoch transition 2/3 of the way through the last slot of the epoch
+    const msSecondsPerSlot = this.config.SECONDS_PER_SLOT * 1000;
+    const msToPrecalculateTime = msSecondsPerSlot * (SLOTS_PER_EPOCH - 1 / 3);
+    await sleep(msToPrecalculateTime, this.signal);
+    const clockSlot = this.chain.clock.currentSlot;
+    const {slot: headSlot, blockRoot} = this.chain.forkChoice.getHead();
+    // node may be syncing or out of synced
+    if (headSlot < clockSlot) {
+      this.logger.verbose("No need to precompute epoch transition", {headSlot, clockSlot});
+      return;
+    }
+    // we want to make sure headSlot === clockSlot to do early epoch transition
+    const nextSlot = computeStartSlotAtEpoch(epoch + 1);
+    this.logger.verbose("Precompute epoch transition", {epoch, headSlot, nextSlot});
+    await this.chain.regen.getBlockSlotState(blockRoot, nextSlot, RegenCaller.preComputeEpoch).catch((e) => {
+      this.logger.error("Failed to precompute epoch transition", epoch, e);
+    });
+  };
+}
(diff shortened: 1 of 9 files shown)


g11tech (dismissed):

packages/lodestar/src/chain/precomputeEpoch.ts
```diff
@@ -0,0 +1,61 @@
...
+      return;
+    }
+    // Precalculate epoch transition 2/3 of the way through the last slot of the epoch
+    const msToPrecalculateTime = (this.config.SECONDS_PER_SLOT * 1000 * 2) / 3;
+    await sleep(msToPrecalculateTime, this.signal);
+    const {slot: headSlot, blockRoot} = this.chain.forkChoice.getHead();
+    const nextEpoch = computeEpochAtSlot(clockSlot) + 1;
+    // node may be syncing or out of synced
+    if (headSlot < clockSlot) {
+      this.logger.verbose("No need to precompute epoch transition", {nextEpoch, headSlot, slot: clockSlot});
```
wemeetagain: Maybe these loggers should be debug level

wemeetagain (dismissed):
lgtm

twoeths:
Tested this for a while:
+ Majority of blocks could be processed <=3s after slot starts, the main reason block is processed after 1/3 slot is due to the gossip block received late
Screen Shot 2021-10-28 at 10 39 08

packages/lodestar/src/chain/precomputeEpoch.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    // Precalculate epoch transition 2/3 of the way through the last slot of the epoch
+    const msToPrecalculateTime = (this.config.SECONDS_PER_SLOT * 1000 * 2) / 3;
+    await sleep(msToPrecalculateTime, this.signal);
+
+    const {slot: headSlot, blockRoot} = this.chain.forkChoice.getHead();
+    const nextEpoch = computeEpochAtSlot(clockSlot) + 1;
+    // node may be syncing or out of synced
+    if (headSlot < clockSlot) {
+      this.logger.debug("No need to precompute epoch transition", {nextEpoch, headSlot, slot: clockSlot});
```
dapplion: When logging think that this logs are burried in a sea of information thus need content. For example
```
Skipping PrecomputeEpochScheduler node is syncing
```

packages/lodestar/src/chain/precomputeEpoch.ts
```diff
@@ -0,0 +1,64 @@
...
+      return;
+    }
+
+    // we want to make sure headSlot === clockSlot to do early epoch transition
+    const nextSlot = clockSlot + 1;
+    this.logger.verbose("Precompute epoch transition", {nextEpoch, headSlot, nextSlot});
```
dapplion: Log `Running PrecomputeEpochScheduler`

packages/lodestar/src/chain/precomputeEpoch.ts
```diff
@@ -0,0 +1,64 @@
...
+    // this takes ~2s as of Oct 2021, no need to wait for this or the clock drift
+    this.chain.regen.getBlockSlotState(blockRoot, nextSlot, RegenCaller.preComputeEpoch).catch((e) => {
```
dapplion: Log both .then() `Completed PrecomputeEpochScheduler` and catch `Error on PrecomputeEpochScheduler`

dapplion: Add basic metrics:
- counter metric for total PrecomputeEpochScheduler runs
- counter metric for total PrecomputeEpochScheduler errors

dapplion: Another super important metric but harder to implement would to answer the question
- When does the result of this `getBlockSlotState()` epoch transition goes to waste? For that we could track re-orgs and check if the epoch transition got re-org-ed out. Also, if the last block of the epoch came very late it won't technically trigger a re-org but it would invalidate this epoch transition run.

dapplion: You should also add a nice long comment on top `getBlockSlotState` call explaining what assumptions this `PrecomputeEpochScheduler` class does on the regen such that the resulting checkpoint state is cached. Also might be worth to add a test, to prevent regressions

twoeths: >  Also, if the last block of the epoch came very late it won't technically trigger a re-org but it would invalidate this epoch transition run.

@dapplion in that case we skip the PreComputeEpoch task, no waste epoch transition here

> When does the result of this getBlockSlotState() epoch transition goes to waste

when noone uses the cached checkpoint, added `preComputedCheckpoint` to `CheckpointStateCache` in order to have that metric

twoeths:
New precompute epoch metrics:
+ Success vs Error vs Skip (no data for Error/Skip legend so it doesn not show)
+ Checkpoint state cache hits vs waste
Screen Shot 2021-10-29 at 17 33 44

packages/lodestar/src/metrics/metrics/lodestar.ts
```diff
@@ -538,5 +538,22 @@ export function createLodestarMetrics(
       name: "unhandeled_promise_rejections",
       help: "UnhandeledPromiseRejection count",
     }),
+
+    // Precompute epoch
+    preComputeEpoch: {
```
dapplion: The name `preComputeEpoch` is bit too generic could we do `precomputeNextEpochTransition`? If sounds good please rename files and classes on all the PR

packages/lodestar/src/metrics/metrics/lodestar.ts
```diff
@@ -538,5 +538,22 @@ export function createLodestarMetrics(
...
+      count: register.counter<"result">({
+        name: "precompute_epoch_result",
```
dapplion: See https://github.com/ChainSafe/lodestar/issues/3399

packages/lodestar/src/metrics/metrics/lodestar.ts
```diff
@@ -538,5 +538,22 @@ export function createLodestarMetrics(
...
+        labelNames: ["result"],
+        help: "Number of precompute epoch skip/success/error",
```
dapplion: Total number of precomputeNextEpochTransition runs by result

packages/lodestar/src/chain/precomputeEpoch.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    // this takes 2s - 4s as of Oct 2021, no need to wait for this or the clock drift
+    // assuming there is no reorg, it caches the checkpoint state & helps avoid doing a full state transition in the next slot
+    //  + when gossip block comes, we need to validate and run state transition
+    //  + if next slot is a skipped slot, it'd help getting target checkpoint state faster to validate attestations
+    this.chain.regen
+      .getBlockSlotState(blockRoot, nextSlot, RegenCaller.preComputeEpoch)
+      .then(() => {
+        this.metrics?.preComputeEpoch.count.inc({result: "success"}, 1);
+        const previousHits = this.chain.checkpointStateCache.updatePreComputedCheckpoint(blockRoot, nextEpoch);
```
dapplion: Can you find a a way to track this without adding custom code in the checkpointStateCache? Ideally this class should have sufficient info to know if the previous run was re-orged or not.

twoeths: @dapplion even when our node is reorged, it does not mean the pre computed epoch transition is wasteful as some nodes can still send attestations with our old/obsolete target checkpoint and we still need that cached target checkpoint. So for this `waste` metric, it does not depend on the node being re-orged or not, it depends on the produced checkpoint state is used or not.

dapplion: Yeah, that's true. Let's leave it like this then

wemeetagain (changes requested):
Can you make the naming consistent?
Either precompute or preCompute, not both

Other than that lgtm

wemeetagain (approved):

dapplion (approved):

PR #3384, opened 2021-10-21
backfill sync from an anchor checkpoint state

### LINKED ISSUE #2556: BlocksByRange under WS

g11tech:
**Motivation**
This PR is to backfill in a scenario when wss checkpoint or any other anchor state is used to start the beacon node.

**Description**
Right now, this PR rebases https://github.com/ChainSafe/lodestar/pull/2637 on current master, making it workable base to build on top of the mentioned previous PR

Fixes:
- [x]  Don't reject the entire syncrange sequence if a block in mid is missing making it a non linear
- [x] Fixe the scenario where the syncrange keeps on resulting in a non anchored block at the end, i.e. end block is orphaned, and the previous head could be multi blocks behind, likely scenario in reorgs. Resolution, load the previous head in a solo request to set it as a new anchor
- [x] before rangeSync or sync from outside, check the DB as DB could have had blocks saved from previous run, no need to verify their signatures and blocks in db should already be verified.
- [x] Do a checkpoint check if a wss checkpoint has been provided which was in the past from the initialized beacon state (from DB because of a previous run). In case not provided compare against genesis when at genesis and panic if not same.
*Created a backfilledSequences repository module, which tracks backfilled sequences like 384:key=> 211:value i.e. backfilled  from slot 384 to 211 both including, in normal operation will typically store two sequences, one used by the backfill process and one used by the forward moving chain (which will upon finalizing will write a new sequence finalizedSlot=>chainStarted/anchorSlot). Any gaps in the db because of restarting with checkpoint sync will result into a new sequence entry, but the previous entry will be eliminated when the backfill syncs up the gap*
- [x] Strategy to skip already filled sequence (needs a strategy to track gaps or filled sequences)
- [x]  `InvalidWSCheckpoint` error (https://github.com/ChainSafe/lodestar/issues/3385)
- [x] Schedule the processing/verification when the chain is supposedly free *throttling chain by `BATCH_SIZE` in a single trigger cycle (in db reads or syncRange fetches)*
- [x] grafana dashboard

~~skipBackFillSync option?
backfillTill:  options i) till  wsCheckpoint  to be validated, ii)till GENESIS_SLOT (default) iii)till provided Slot~~

Tested Scenarios:
- [x] sync test case: sync/wss.test.ts
image
- [x] prater backfill sync to genesis (despite multiple restarts interrupting the sync, recovering using backfilled Sequences)
image
- [x] stopping and starting multiple times creating gaps in the back filled sequences, backfiled sync filled the gaps and combined them with backfilled sequences to avoid redoing previous backfill sync. 
*example of backfillSync  recovering on a restart using previous backfilled sequence (jumped back from `1633696` to `1437564`, and then continuing backfilling from there on*:
image
- [x] recovering from a persistent non anchored sequence in range sync (because of the orphaned block being served at the edge of range sync)
image
- [x] recovering the **backfilled to genesis** state on a fully backfilled node restart
image
- [x] stopped a fully backfilled node and restarted after a gap with new recent checkpoint - backfilled from the new recent checkpoint to previous finalized checkpoint when the node was stopped, and then correlated with already backfilled sequence till the GENESIS to switch to backfilled complete state
image

Closes #2556

### DIFF
--- a/packages/cli/src/cmds/beacon/initBeaconState.ts
+++ b/packages/cli/src/cmds/beacon/initBeaconState.ts
@@ -38,7 +38,7 @@ async function initAndVerifyWeakSubjectivityState(
   store: TreeBacked<allForks.BeaconState>,
   wsState: TreeBacked<allForks.BeaconState>,
   wsCheckpoint: Checkpoint
-): Promise<TreeBacked<allForks.BeaconState>> {
+): Promise<{anchorState: TreeBacked<allForks.BeaconState>; wsCheckpoint: Checkpoint}> {
   // Check if the store's state and wsState are compatible
   if (
     store.genesisTime !== wsState.genesisTime ||
--- a/packages/cli/src/cmds/beacon/initBeaconState.ts
+++ b/packages/cli/src/cmds/beacon/initBeaconState.ts
@@ -64,7 +64,8 @@ async function initAndVerifyWeakSubjectivityState(
     throw new Error("Fetched weak subjectivity checkpoint not within weak subjectivity period.");
   }
 
-  return await initStateFromAnchorState(config, db, logger, anchorState);
+  await initStateFromAnchorState(config, db, logger, anchorState);
+  return {anchorState: wsState, wsCheckpoint};
 }
 
 /**
@@ -83,7 +84,7 @@ export async function initBeaconState(
   db: IBeaconDb,
   logger: ILogger,
   signal: AbortSignal
-): Promise<TreeBacked<allForks.BeaconState>> {
+): Promise<{anchorState: TreeBacked<allForks.BeaconState>; wsCheckpoint?: Checkpoint}> {
   // fetch the latest state stored in the db
   // this will be used in all cases, if it exists, either used during verification of a weak subjectivity state, or used directly as the anchor state
   const lastDbState = await db.stateArchive.lastValue();
@@ -132,16 +133,19 @@ export async function initBeaconState(
   } else if (lastDbState) {
     // start the chain from the latest stored state in the db
     const config = createIBeaconConfig(chainForkConfig, lastDbState.genesisValidatorsRoot);
-    return await initStateFromAnchorState(config, db, logger, lastDbState);
+    const anchorState = await initStateFromAnchorState(config, db, logger, lastDbState);
+    return {anchorState};
   } else {
     const genesisStateFile = args.genesisStateFile || getGenesisFileUrl(args.network || defaultNetwork);
     if (genesisStateFile && !args.forceGenesis) {
       const stateBytes = await downloadOrLoadFile(genesisStateFile);
       const anchorState = getStateTypeFromBytes(chainForkConfig, stateBytes).createTreeBackedFromBytes(stateBytes);
       const config = createIBeaconConfig(chainForkConfig, anchorState.genesisValidatorsRoot);
-      return await initStateFromAnchorState(config, db, logger, anchorState);
+      await initStateFromAnchorState(config, db, logger, anchorState);
+      return {anchorState};
     } else {
-      return await initStateFromEth1({config: chainForkConfig, db, logger, opts: options.eth1, signal});
+      const anchorState = await initStateFromEth1({config: chainForkConfig, db, logger, opts: options.eth1, signal});
+      return {anchorState};
     }
   }
 }
--- a/packages/cli/src/cmds/beacon/initBeaconState.ts
+++ b/packages/cli/src/cmds/beacon/initBeaconState.ts
@@ -64,8 +64,10 @@ async function initAndVerifyWeakSubjectivityState(
     throw new Error("Fetched weak subjectivity checkpoint not within weak subjectivity period.");
   }
 
-  await initStateFromAnchorState(config, db, logger, anchorState);
-  return {anchorState: wsState, wsCheckpoint};
+  anchorState = await initStateFromAnchorState(config, db, logger, anchorState);
+
+  // Return the latest anchorState but still return original wsCheckpoint to validate in backfill
+  return {anchorState, wsCheckpoint};
 }
 
 /**
@@ -139,9 +141,9 @@ export async function initBeaconState(
     const genesisStateFile = args.genesisStateFile || getGenesisFileUrl(args.network || defaultNetwork);
     if (genesisStateFile && !args.forceGenesis) {
       const stateBytes = await downloadOrLoadFile(genesisStateFile);
-      const anchorState = getStateTypeFromBytes(chainForkConfig, stateBytes).createTreeBackedFromBytes(stateBytes);
+      let anchorState = getStateTypeFromBytes(chainForkConfig, stateBytes).createTreeBackedFromBytes(stateBytes);
       const config = createIBeaconConfig(chainForkConfig, anchorState.genesisValidatorsRoot);
-      await initStateFromAnchorState(config, db, logger, anchorState);
+      anchorState = await initStateFromAnchorState(config, db, logger, anchorState);
       return {anchorState};
     } else {
       const anchorState = await initStateFromEth1({config: chainForkConfig, db, logger, opts: options.eth1, signal});
--- a/packages/lodestar/src/chain/interface.ts
+++ b/packages/lodestar/src/chain/interface.ts
@@ -39,6 +39,7 @@ export interface IBeaconChain {
   readonly executionEngine: IExecutionEngine;
   // Expose config for convenience in modularized functions
   readonly config: IBeaconConfig;
+  readonly anchorSlot: Slot;
 
   bls: IBlsVerifier;
   forkChoice: IForkChoice;
--- a/packages/lodestar/src/chain/interface.ts
+++ b/packages/lodestar/src/chain/interface.ts
@@ -39,6 +39,8 @@ export interface IBeaconChain {
   readonly executionEngine: IExecutionEngine;
   // Expose config for convenience in modularized functions
   readonly config: IBeaconConfig;
+
+  /** The initial slot that the chain is started with */
   readonly anchorSlot: Slot;
 
   bls: IBlsVerifier;
--- a/packages/lodestar/src/db/beacon.ts
+++ b/packages/lodestar/src/db/beacon.ts
@@ -18,7 +18,7 @@ import {
   CheckpointHeaderRepository,
   SyncCommitteeRepository,
   SyncCommitteeWitnessRepository,
-  BackfilledSequences,
+  BackfilledRanges,
 } from "./repositories";
 import {PreGenesisState, PreGenesisStateLastProcessedBlock} from "./single";
 
@@ -45,7 +45,7 @@ export class BeaconDb extends DatabaseService implements IBeaconDb {
   syncCommittee: SyncCommitteeRepository;
   syncCommitteeWitness: SyncCommitteeWitnessRepository;
 
-  backfilledSequences: BackfilledSequences;
+  backfilledRanges: BackfilledRanges;
 
   constructor(opts: IDatabaseApiOptions) {
     super(opts);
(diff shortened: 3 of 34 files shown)


g11tech:
~~example of backfillSync  recovering on a restart using previous backfilled sequence (jumped back from `1633696` to `1437564`, and then continuing backfilling from there on~~ (updated the scenario in the Tested Scenario section)

packages/lodestar/src/metrics/metrics/lodestar.ts
```diff
@@ -411,6 +411,21 @@ export function createLodestarMetrics(
       }),
     },
 
+    backfillSync: {
+      totalBlocks: register.gauge({
+        name: "lodestar__backfill_sync_total_blocks",
```
dapplion: Remove double underscore from `lodestar__backfill`

packages/lodestar/src/node/notifier.ts
```diff
@@ -59,6 +59,11 @@ export async function runNodeNotifier({
       const skippedSlots = clockSlot - headInfo.slot;
       const clockSlotRow = `slot: ${clockSlot}` + (skippedSlots > 0 ? ` (skipped ${skippedSlots})` : "");
 
+      // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
```
dapplion: Don't disable this rule, use strict comparisions even if it's more verbose

packages/lodestar/src/node/notifier.ts
```diff
@@ -75,20 +80,21 @@ export async function runNodeNotifier({
             clockSlotRow,
             headRow,
             finalizedCheckpointRow,
+            ...backfilledStat,
```
dapplion: I won't add any info to the notifier. Users don't really care about backfill status. We must ensure it works and they trust us that it does. We should use metrics and API debug endpoints to learn about the status of backfill sync

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,336 @@
...
+  private readonly chain: IBeaconChain;
+  private readonly network: INetwork;
+  private readonly db: IBeaconDb;
+  private readonly config: IBeaconConfig;
+  private readonly logger: ILogger;
+  private readonly metrics: IMetrics | null;
+  private wsCheckpoint?: phase0.Checkpoint | null;
+  private opts: BackfillSyncOpts;
+
+  // At any given point, we should have either the anchorBlock, or anchorBlockRoot, the reversed head of the backfill sync, from where we need to backfill
```
dapplion: Limit line width to strictly less than 120 characters wide for readability

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,336 @@
...
+        this.backfillStartFromSlot = anchorCp.epoch * SLOTS_PER_EPOCH;
+        this.logger.info("BackfillSync - initializing from Checkpoint", {
+          root: toHexString(anchorCp.root),
+          epoch: anchorCp.epoch,
+        });
+      }
+
+      // Update from the backfilled sequence
+
+      // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
```
dapplion: Please remove all disablings of the rule `@typescript-eslint/strict-boolean-expressions` from this PR

packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -0,0 +1,336 @@
...
+
+  /** Throw / return all AsyncGenerators */
+  close(): void {
+    this.network.events.off(NetworkEvent.peerConnected, this.addPeer);
+    this.network.events.off(NetworkEvent.peerDisconnected, this.removePeer);
+    this.processor.end(new ErrorAborted("BackfillSync"));
+  }
+
+  /** Returns oldestSlotSynced */
+  private async sync(): Promise<Slot> {
```
dapplion, line 280: This function is 134 lines long and hard to follow, please
- Add code comments throughout the code blocks explaining what is going on and how this is intended to work
- If possible try to modularize the code with good abstractions that help the above point self-documenting the code
- As bonus (if applicable) you can add a diagram or summary of the entire flow at the top of the function

packages/lodestar/src/sync/backfill/errors.ts
```diff
@@ -0,0 +1,21 @@
+import {LodestarError} from "@chainsafe/lodestar-utils";
+import PeerId from "peer-id";
+import {Root} from "@chainsafe/lodestar-types";
+
+export enum BackfillSyncErrorCode {
+  /** fetched block doesn't connect to anchor block */
+  NOT_ANCHORED = "not_anchored",
+  // fetched blocks are not linear
```
dapplion: Make all these comments JSDocs
```ts
/** fetched blocks are not linear */
```

packages/lodestar/src/sync/backfill/verify.ts
```diff
@@ -0,0 +1,49 @@
...
+): {
+  nextAnchor: allForkTypes.SignedBeaconBlock | null;
+  verifiedBlocks: allForkTypes.SignedBeaconBlock[];
+  error?: BackfillSyncErrorCode.NOT_LINEAR;
+} {
+  let nextRoot: Root = anchorRoot;
+  let nextAnchor: allForkTypes.SignedBeaconBlock | null = null;
+
+  const verifiedBlocks: allForkTypes.SignedBeaconBlock[] = [];
+  for (const block of blocks.slice(0).reverse()) {
```
dapplion: Just loop backwards (with a regular for loop) without allocating a new array twice

packages/lodestar/src/sync/backfill/verify.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export async function verifyBlockProposerSignature(
+  bls: IBlsVerifier,
+  state: CachedBeaconState<allForks.BeaconState>,
+  blocks: allForkTypes.SignedBeaconBlock[]
+): Promise<void> {
+  const signatures = blocks
+    // genesis block doesn't have valid signature
+    .filter((block) => block.message.slot !== GENESIS_SLOT)
+    .map((block) => allForks.getProposerSignatureSet(state, block));
```
dapplion: .filter and .map can be merged into a single loop to prevent looping twice

packages/lodestar/src/sync/sync.ts
```diff
@@ -21,6 +24,9 @@ export class BeaconSync implements IBeaconSync {
   private readonly rangeSync: RangeSync;
   private readonly unknownBlockSync: UnknownBlockSync;
 
+  // avoid finding same root at the same time
+  private readonly processingRoots = new Set<string>();
```
dapplion: Is this used?

g11tech: nope, will remove this :+1:

dapplion:
@g11tech Remember to mark as "Ready for review" when appropiate! Draft PRs should be a transient state for unfinished work

g11tech:
Graphana dashboard:
image
analysis: 
1. on starting beacon node with weakSubjectivitySync latest, the backfill sync started (1st graph), 
2. backfillTillSlot (2nd graph) started decreasing via syncing blocks from the network using rangesync (blue line in 3rd graph), 
3. till around 17:50 time mark, where backfill sync encountered a previous backfilled saved sequence (from the previous run few days ago) all the way to genesis (yellow spike in 3rd graph).
4. then backfill sync status turns to completed, backfillTillSlot goes to 0 and backfill sync exits.

4rth graph is the error rate.

dapplion:
This looks super good! Thank you so much for the detailed charts and progression :heart: Will do another review in depth asap and should be good to go.

g11tech:
> This looks super good! Thank you so much for the detailed charts and progression heart Will do another review in depth asap and should be good to go.

thank you! :+1:

packages/lodestar/src/chain/interface.ts
```diff
@@ -40,6 +40,7 @@ export interface IBeaconChain {
   readonly executionEngine: IExecutionEngine;
   // Expose config for convenience in modularized functions
   readonly config: IBeaconConfig;
+  readonly anchorSlot: Slot;
```
wemeetagain, line 44: would be nice to have a js doc:

Something like:
The initial slot that the chain is started with.

packages/lodestar/src/db/beacon.ts
```diff
@@ -53,6 +54,8 @@ export class BeaconDb extends DatabaseService implements IBeaconDb {
   lightClientInitProofIndex: LightClientInitProofIndexRepository;
   lightClientSyncCommitteeProof: LightClientSyncCommitteeProofRepository;
 
+  backfilledSequences: BackfilledSequences;
```
wemeetagain: I think using the term "ranges" would be more descriptive than "sequences".

wemeetagain (changes requested):

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -134,16 +135,19 @@ export async function initBeaconState(
...
+    const anchorState = await initStateFromAnchorState(config, db, logger, lastDbState);
+    return {anchorState};
   } else {
     const genesisStateFile = args.genesisStateFile || getGenesisFileUrl(args.network || defaultNetwork);
     if (genesisStateFile && !args.forceGenesis) {
       const stateBytes = await downloadOrLoadFile(genesisStateFile);
       const anchorState = getStateTypeFromBytes(chainForkConfig, stateBytes).createTreeBackedFromBytes(stateBytes);
       const config = createIBeaconConfig(chainForkConfig, anchorState.genesisValidatorsRoot);
-      return await initStateFromAnchorState(config, db, logger, anchorState);
+      await initStateFromAnchorState(config, db, logger, anchorState);
```
wemeetagain: I think we should be returning the return value of initStateFromAnchorState here

wemeetagain (dismissed):

packages/cli/src/cmds/beacon/initBeaconState.ts
g11tech: `initStateFromAnchorState` returns anchorState itself, but yes it would look cleaner

g11tech:
synopsis of a test run with the latest PR code:

image
1. Started backfill sync around 16:30
2. Restarted coouple of times between ~17:00-18:30, sync recovered the previous progress using backfillRanges
3. Synced to genesis at 20:00 but sync aborted with failiure as code was checking signature of genesis block as well
4. Fixed it around 22:00, and the sync recovered previous progress to genesis which it complete successfully

@dapplion : Kindly review and let know if anything further needs to be fine tuned.

g11tech:
@dapplion updated the PR with the discussed changes (and rebased with current master). test run where backfill sync recovered previous progress and completed sync to genesis. :slightly_smiling_face: 

image

dapplion (approved):
Excellent! Looking forward to many more like this in 2022 :tada:

g11tech:
> Excellent! Looking forward to many more like this in 2022 tada

Thank you! lodestar rocks! :rocket: 
latest run on prater
image

Issue #2483, opened 2021-05-04
Track UnhandledPromiseRejectionWarning count with metric 


dapplion:
**Feature request**

UnhandledPromiseRejectionWarning signal that something is potentially really wrong

Listen for them add track with metric

---

Also consider tracking the count of error logs by module

dapplion:
@g11tech Do you want to take this issue?

g11tech:
Sure pls assign

PR #3386, opened 2021-10-23
tracking UnhandledPromiseRejection(s)

g11tech:
**Motivation**
Some modules throw error in promises which have not been caught and handled, those errors need to be tracked and monitored (and investigated if required).

This PR tracks the UnhandledPfomiseRejection count in the metrics, and add a panel to the `VM Stats` section in metrics dashboard.
**Description**

image

Closes #issue_number
#2483


twoeths (approved):

wemeetagain (approved):
Good idea, good first step to tracking them down

PR #3391, opened 2021-10-25
weak subjecitivity checks

### LINKED ISSUE #3385: investigate: weak subjectivity checkpoint improper use
The weak subjectivity seems to be not being properly used. As per my understanding of the code, following are the issues:

1. when `weakSubjectivityStateFile` is being used, `isWithinWeakSubjectivityPeriod` check will almost always pass doesn't matter hold old the state if starting from blank block db (first time use)
2. Beacon should init from the db's last (biggest height) checkpoint instead from the provided checkpoint if db's last checkpoint height is more (forkChoice rule: build on top of the last seen checkpoint). In this case however point 4 check should be made
~~. `isWithinWeakSubjectivityPeriod` check should be done when the node gets into `Synced` state for the first time since being started. This check should be done irrespective of weak subjectivity flags provisioning, and should be checked against last anchor checkpoint state from where the sync started (even if the node was previously synced but went out of sync for any reason)~~
3. If a checkpoint is provided, and db's last save checkpoint > than the provided checkpoint, there needs to be a strict check made that provided checkpoint is a  ancestor of db's checkpoint from where the state is being fetched (apart from early check of being with the wss period).
4. documentation on `weakSubjectivitySyncLatest` always fetching the weak subjectivity checkpoint on starts/restarts.

Resolution: fix if any issue is deemed correct upon discussion.

TRACKER:
~~-  When the node syncs to chain head for the first time after starting or falling out of sync, the weak subjectivity checks should be done. This might not be an issue  in normal operation but an attacker could devise scenarios to server just borderline checkpoints which on sync might be out of sync. or in extreme scenarios where the validator set shrinks so much that the wss period is small.~~
~~Check the anchor state's checkpoint is the parent of the finalized checkpoint when the chain syncs for first time to head else panic~~

- [x] In case finalized state from DB is ahead of the provide wss state, make the check that checkpoint is parent of db's finalized and then move forward else again start from the wss checkpoint 
- [x] rectify check on wss's subjectivity period of the anchor state w.r.t. the clockSlot while initializing the chain
- [x] if the wss state is  before db's archieved state, then validate if the wss checkpoint is the ancestor of the db's state (https://github.com/ChainSafe/lodestar/pull/3384)
- [ ] Documentation

g11tech:
**Motivation**

The aim of the PR is to rectify weak subjectivity checks:
- [x] while intializing if DB's finalized is ahead of the provided anchor states finalized (though wssStateFile), use the db's finalized as anchor state instead of again starting from the wss checkpoint  
- [x] add a simpler check on wss's subjectivity period of the anchor state w.r.t. the clockSlot while initializing the chain
**Description**

UPDATE:
previous resolution of making the wss check on the head when chain syncs to synced state for the first time was deemed overkill, so reverted that and replaced with the initialization time `clockEpoch` check.

example error of providing an checkpoint wss subjectivity period behind the `clockEpoch`
image

Closes #issue_number
partially closes #3385

### DIFF
--- a/packages/cli/src/cmds/beacon/initBeaconState.ts
+++ b/packages/cli/src/cmds/beacon/initBeaconState.ts
@@ -49,12 +49,19 @@ async function initAndVerifyWeakSubjectivityState(
     );
   }
 
-  if (!allForks.isWithinWeakSubjectivityPeriod(config, wsState, wsCheckpoint)) {
+  // Pick the state which is ahead as an anchor to initialize the beacon chain
+  let anchorState = wsState;
+  let anchorCheckpoint = wsCheckpoint;
+  if (store.slot > wsState.slot) {
+    anchorState = store;
+    anchorCheckpoint = getCheckpointFromState(config, store);
+    logger.info("Db state is ahead of the provided checkpoint state, using the same to initialize the beacon chain");
+  }
+
+  if (!allForks.isWithinWeakSubjectivityPeriod(config, anchorState, anchorCheckpoint)) {
     throw new Error("Fetched weak subjectivity checkpoint not within weak subjectivity period.");
   }
 
-  // Pick the state which is ahead as an anchor to initialize the beacon chain
-  const anchorState = wsState.slot > store.slot ? wsState : store;
   return await initStateFromAnchorState(config, db, logger, anchorState);
 }
 
--- a/packages/cli/src/cmds/beacon/initBeaconState.ts
+++ b/packages/cli/src/cmds/beacon/initBeaconState.ts
@@ -55,7 +55,7 @@ async function initAndVerifyWeakSubjectivityState(
   if (store.slot > wsState.slot) {
     anchorState = store;
     anchorCheckpoint = getCheckpointFromState(config, store);
-    logger.info("Db state is ahead of the provided checkpoint state, using the same to initialize the beacon chain");
+    logger.verbose("Db state is ahead of the provided checkpoint state, using the db state to initialize the beacon chain");
   }
 
   if (!allForks.isWithinWeakSubjectivityPeriod(config, anchorState, anchorCheckpoint)) {
--- a/packages/cli/src/cmds/beacon/initBeaconState.ts
+++ b/packages/cli/src/cmds/beacon/initBeaconState.ts
@@ -48,7 +48,7 @@ async function initAndVerifyWeakSubjectivityState(
       "Db state and checkpoint state are not compatible, either clear the db or verify your checkpoint source"
     );
   }
-  
+
   if (!allForks.isWithinWeakSubjectivityPeriod(config, wsState, wsCheckpoint)) {
     throw new Error("Fetched weak subjectivity checkpoint not within weak subjectivity period.");
   }
--- a/packages/cli/src/cmds/beacon/initBeaconState.ts
+++ b/packages/cli/src/cmds/beacon/initBeaconState.ts
@@ -1,5 +1,5 @@
 import {AbortSignal} from "@chainsafe/abort-controller";
-
+import {ssz} from "@chainsafe/lodestar-types";
 import {TreeBacked} from "@chainsafe/ssz";
 import {createIBeaconConfig, IBeaconConfig, IChainForkConfig} from "@chainsafe/lodestar-config";
 import {fromHex, ILogger} from "@chainsafe/lodestar-utils";
@@ -39,10 +39,23 @@ async function initAndVerifyWeakSubjectivityState(
   wsState: TreeBacked<allForks.BeaconState>,
   wsCheckpoint: Checkpoint
 ): Promise<TreeBacked<allForks.BeaconState>> {
-  if (!allForks.isWithinWeakSubjectivityPeriod(config, store, wsState, wsCheckpoint)) {
+  // Check if the store's state and wsState are compatible
+  if (
+    store.genesisTime !== wsState.genesisTime ||
+    !ssz.Root.equals(store.genesisValidatorsRoot, wsState.genesisValidatorsRoot)
+  ) {
+    throw new Error(
+      "Db state and checkpoint state are not compatible, either clear the db or verify your checkpoint source"
+    );
+  }
+  
+  if (!allForks.isWithinWeakSubjectivityPeriod(config, wsState, wsCheckpoint)) {
     throw new Error("Fetched weak subjectivity checkpoint not within weak subjectivity period.");
   }
-  return await initStateFromAnchorState(config, db, logger, wsState);
+
+  // Pick the state which is ahead as an anchor to initialize the beacon chain
+  const anchorState = wsState.slot > store.slot ? wsState : store;
+  return await initStateFromAnchorState(config, db, logger, anchorState);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/chain/eventHandlers.ts
```diff
@@ -282,3 +284,26 @@ export async function onErrorBlock(this: BeaconChain, err: BlockError): Promise<
...
+  const anchorCheckpoint = {
+    epoch: computeEpochAtSlot(this.anchorState.latestBlockHeader.slot),
+    root: allForksBST.getLatestBlockRoot(this.config, this.anchorState),
+  };
+  if (!allForksBST.isWithinWeakSubjectivityPeriod(this.config, store, this.anchorState, anchorCheckpoint)) {
+    this.logger.error(
+      "Anchor state used for starting the chain not within the weak subjectivity period from the synced head."
+    );
+    this.close();
+    process.exit(1);
```
dapplion: That's a no-go, this is very bad, we must _never_ call process.exit() anywhere but the CLI. Lodestar code must be runnable as a library as should never do this kind a decisions. It must communicate this error condition to the CLI via some event or callback and then the CLI decides to exit properly, stopping the components gracefully _then_ exiting

packages/lodestar/test/utils/mocks/chain/chain.ts
```diff
@@ -63,6 +63,7 @@ export class MockBeaconChain implements IBeaconChain {
   emitter: ChainEventEmitter;
   lightclientUpdater: LightClientUpdater;
   lightClientIniter: LightClientIniter;
+  anchorState: TreeBacked<allForks.BeaconState>;
```
dapplion: A TreeBacked state takes hundreds of MB in memory, this will significantly increase the memory footprint of Lodestar. Why do you need to keep a strong reference to this state for the entire duration of the application?

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -42,7 +42,7 @@ async function initAndVerifyWeakSubjectivityState(
   if (!allForks.isWithinWeakSubjectivityPeriod(config, store, wsState, wsCheckpoint)) {
     throw new Error("Fetched weak subjectivity checkpoint not within weak subjectivity period.");
   }
-  return await initStateFromAnchorState(config, db, logger, wsState);
+  return await initStateFromAnchorState(config, db, logger, wsState.slot > store.slot ? wsState : store);
```
dapplion: Please add comments explaining critical logic like this

packages/lodestar/src/chain/emitter.ts
```diff
@@ -117,6 +123,8 @@ export interface IChainEvents {
   [ChainEvent.forkChoiceReorg]: (head: IProtoBlock, oldHead: IProtoBlock, depth: number) => void;
   [ChainEvent.forkChoiceJustified]: (checkpoint: CheckpointWithHex) => void;
   [ChainEvent.forkChoiceFinalized]: (checkpoint: CheckpointWithHex) => void;
+
+  [ChainEvent.chainSynced]: () => void;
```
dapplion: Use consistent white-spacing with adjacent code. In this case don't add a newline since that's what the rest of the interface does, but in `onChainSynced()` add whitespace to separate related blocks of logic, for reference https://github.com/ChainSafe/lodestar/pull/3392/commits/1c1508c18621ce1ea0985eff703b47c20b84d42a

packages/lodestar/src/chain/eventHandlers.ts
```diff
@@ -7,6 +7,7 @@ import {
   CachedBeaconState,
   computeEpochAtSlot,
   computeStartSlotAtEpoch,
+  allForks as allForksBST,
```
dapplion: You should not need to rename this, the types are part of the `@chainsafe/lodestar-beacon-state-transition` allforks export

packages/lodestar/src/chain/eventHandlers.ts
```diff
@@ -282,3 +284,26 @@ export async function onErrorBlock(this: BeaconChain, err: BlockError): Promise<
...
+
+function onChainSynced(this: BeaconChain): void {
+  const store = this.getHeadState();
+  const anchorCheckpoint = {
+    epoch: computeEpochAtSlot(this.anchorState.latestBlockHeader.slot),
+    root: allForksBST.getLatestBlockRoot(this.config, this.anchorState),
+  };
+  if (!allForksBST.isWithinWeakSubjectivityPeriod(this.config, store, this.anchorState, anchorCheckpoint)) {
+    this.logger.error(
+      "Anchor state used for starting the chain not within the weak subjectivity period from the synced head."
```
dapplion: Why does one need to wait to be synced to check call isWithinWeakSubjectivityPeriod()?

dapplion (dismissed):

packages/lodestar/src/chain/eventHandlers.ts
g11tech: I am trying to be extra cautious here, considering a hacker who could distribute a checkpoint which is barely in range but by the point they are in synced (lets say some sort of a long running attack, or peer/internet issues) the first time the person is synced, his head's finalized is not in the range)

packages/lodestar/test/utils/mocks/chain/chain.ts
g11tech: umm, yes i can do optimization here, store the checkpoint and its weak subjectivyt range to check when synced. thats the only thing this is being currently used for!

packages/lodestar/src/chain/eventHandlers.ts
dapplion: If the researchers don't instruct us to do this check we should not do it. WeakSubjectity is not binary; the closer you are to the threshold the harder it is for the attacker to execute the attack

g11tech: ok, what we should atleast do is (last point in the PR description tracker): add a simpler check on wss's subjectivity period of the anchor state w.r.t. the clockSlot while initializing the chain, which is currently missing

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -39,10 +39,22 @@ async function initAndVerifyWeakSubjectivityState(
...
+  // Check if the store's state and wsState are compatible
+  if (
+    store.genesisTime !== wsState.genesisTime ||
+    !ssz.Root.equals(store.genesisValidatorsRoot, wsState.genesisValidatorsRoot)
+  ) {
+    throw new Error(
+      "Db state and checkpoint state are not compatible, either clear the db or verify your checkpoint source"
+    );
+  }
+  if (!allForks.isWithinWeakSubjectivityPeriod(config, wsState, wsCheckpoint)) {
```
dapplion: Please add whitespace between big if blocks for readability

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -39,10 +39,30 @@ async function initAndVerifyWeakSubjectivityState(
...
+    );
+  }
+
+  // Pick the state which is ahead as an anchor to initialize the beacon chain
+  let anchorState = wsState;
+  let anchorCheckpoint = wsCheckpoint;
+  if (store.slot > wsState.slot) {
+    anchorState = store;
+    anchorCheckpoint = getCheckpointFromState(config, store);
+    logger.info("Db state is ahead of the provided checkpoint state, using the same to initialize the beacon chain");
```
wemeetagain:
```suggestion
    logger.verbose("Db state is ahead of the provided checkpoint state, using the db state to initialize the beacon chain");
```

wemeetagain (dismissed):

twoeths (dismissed):

wemeetagain (approved):

dapplion (approved):

Issue #3390, opened 2021-10-25
Investigate performance issue in pithos testnet


twoeths:
**Describe the bug**

The total state transition takes so much time in lodestar-geth instance of pithos network
```
Oct-25 09:26:49.206 [CHAIN]         verbose: Clock slot slot=78409
Oct-25 09:26:49.208 [NETWORK]       verbose: Received gossip block slot=78409, root=0x961a…0d3a, curentSlot=78409
Oct-25 09:26:51.338 [CHAIN]         verbose: Block processed slot=78409, root=0x961acd847841288c4584419e3f0530cee625779160c729821f3fa273f60c0d3a
Oct-25 09:27:00.623 [CHAIN]         verbose: Clock slot slot=78410
Oct-25 09:27:02.152 [NETWORK]       verbose: Received gossip block slot=78410, root=0xd888…5d09, curentSlot=78410
Oct-25 09:27:12.025 [CHAIN]         verbose: Clock slot slot=78411
Oct-25 09:27:13.964 [NETWORK]       verbose: Received gossip block slot=78411, root=0x5ac8…8ad4, curentSlot=78411
Oct-25 09:27:15.908 [CHAIN]         verbose: Block processed slot=78410, root=0xd888b52b3ac5089ba6ad7630592a6e956b8c29c1f0254c29be6b3c09dd645d09
Oct-25 09:27:24.076 [CHAIN]         verbose: Clock slot slot=78412
Oct-25 09:27:24.185 [NETWORK]       verbose: Received gossip block slot=78412, root=0x398b…5f83, curentSlot=78412
Oct-25 09:27:25.557 [CHAIN]         verbose: Block processed slot=78411, root=0x5ac82b0192fa834a322e10f7d3542dcdc16e65f65fa23ad9122f436e61898ad4
Oct-25 09:27:26.267 [CHAIN]         verbose: Block processed slot=78412, root=0x398b5d8b077289c2828ad9ee22cae0ae344394d0f32f77482234d8072d775f83
Oct-25 09:27:36.024 [CHAIN]         verbose: Clock slot slot=78413
Oct-25 09:27:37.965 [CHAIN]         verbose: Block processed slot=78413, root=0xe30594ab7157bf1d360bb094c28e1246fe112550ef31f93f92d35c80aea2221e
```

The contabo-3 instance shows the same
[...]

Issue #3034, opened 2021-08-29
Improve gossip attestation processing capacity


dapplion:
A busy Lodestar beacon node sometimes struggles to process all aggregates and attestations from gossip. I'm opening this issue to explore different queue and processing strategies to improve capacity.

### Sync batch validation

Currently all incoming gossip objects are placed in a queue and processed sequentially (LIFO) with a concurrency of 16.

A different approach would be modify our `validateGossipAggregateAndProof` to take a batch of aggregates with the same target so process the synchronous steps at once. To test I've patched this code in the API of a busy node:

```ts
async function timeQueueStyles(
  chain: BeaconChain,
  signedAggregateAndProof: phase0.SignedAggregateAndProof,
  count: number
): Promise<{seriesMs: number; parallelMs: number; batchMs: number}> {
  const indexes = linspace(0, count - 1);

  // time seriesMs
  for (let i = 0; i < count; i++) {
    await validateGossipAggregateAndProof(chain, signedAggregateAndProof);
  }

  // time parallelMs
  await Promise.all(indexes.map(() => validateGossipAggregateAndProof(chain, signedAggregateAndProof)));

  // time batchMs
  await validateGossipAggregateAndProofBatch(chain, indexes.map(() => signedAggregateAndProof));
}
```

`validateGossipAggregateAndProofBatch` does the syncronous checks in a for loop, while getting the target state once and directly verifying the signatures in batch. The code is committed at https://github.com/ChainSafe/lodestar/compare/master...dapplion/queue-batch

```ts
export async function validateGossipAggregateAndProofBatch(
  chain: IBeaconChain,
  signedAggregateAndProofs: phase0.SignedAggregateAndProof[]
): Promise<void> {
  for (const signedAggregateAndProof of signedAggregateAndProofs) {
    // Pre target state checks
  }

  const targetState = await chain.regen.getCheckpointState(attTarget, RegenCaller.validateGossipAggregateAndProof);

  for (const signedAggregateAndProof of signedAggregateAndProofs) {
    // Post target state checks
  }

  // Get signatureSets for all aggregates
  if (!(await chain.bls.verifySignatureSets(signatureSets, {batchable: true}))) {
    throw new AttestationError(GossipAction.REJECT, {code: AttestationErrorCode.INVALID_SIGNATURE});
  }

  for (const signedAggregateAndProof of signedAggregateAndProofs) {
    // Post signature checks
  }
}
```

[...]

PR #3392, opened 2021-10-26
Submit API endpoints: verify in batch

twoeths:
**Motivation**

+ Submitting attestations requires verifying signatures separately and regen same target state

**Description**

+ Regen same target state first
+ Validate all attestations
+ Get all signatures and verify once
+ If one attestation failed, handle separately like what we're doing

part of #3390

### DIFF
--- a/packages/lodestar/src/api/impl/beacon/pool/index.ts
+++ b/packages/lodestar/src/api/impl/beacon/pool/index.ts
@@ -1,7 +1,7 @@
 // eslint-disable-next-line no-restricted-imports
 import {Api as IBeaconPoolApi} from "@chainsafe/lodestar-api/lib/routes/beacon/pool";
-import {Epoch, ssz} from "@chainsafe/lodestar-types";
-import {allForks, phase0} from "@chainsafe/lodestar-beacon-state-transition";
+import {altair, Epoch, ssz} from "@chainsafe/lodestar-types";
+import {allForks, CachedBeaconState, phase0} from "@chainsafe/lodestar-beacon-state-transition";
 import {SYNC_COMMITTEE_SIZE, SYNC_COMMITTEE_SUBNET_COUNT} from "@chainsafe/lodestar-params";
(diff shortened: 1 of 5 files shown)


packages/lodestar/src/api/impl/beacon/pool/index.ts
```diff
@@ -43,39 +53,105 @@ export function getBeaconPoolApi({
...
-            errors.push(e as Error);
-            logger.error(
-              `Error on submitPoolAttestations [${i}]`,
-              {slot: attestation.data.slot, index: attestation.data.index},
-              e as Error
+      // we want all attestations to have same target so that we can validate in batch
+      const attTarget = attestations[0].data.target;
+      for (let i = 1; i < attestations.length; i++) {
+        if (!ssz.phase0.Checkpoint.equals(attestations[i].data.target, attTarget)) {
+          throw Error("Target checkpoints are not unique");
```
dapplion, line 75: This is a dangerous assumption not spec-ed in the beacon APIs repo. Our beacon node could break validators that submit multiple attestations with different targets. What this can be refactored instead is into grouping attestations by target, and then running them on a for loop, or Promise.all(). If all attestations have the same target then the result would be the same as your logic

dapplion:
@tuyennhv I did some research about improving our gossip throughput here https://github.com/ChainSafe/lodestar/issues/3034 and found no improvement by batching attestation processing. I was very surprised by the results, but I guess JS promises are very cheap. Do you have any proof that your PR improves processing times?

twoeths:
> @tuyennhv I did some research about improving our gossip throughput here #3034 and found no improvement by batching attestation processing. I was very surprised by the results, but I guess JS promises are very cheap. Do you have any proof that your PR improves processing times?

it was just my guess, closing as it's not the root cause of the issue https://github.com/ChainSafe/lodestar/issues/3390#issuecomment-956042391 and this was not proved to be a good improvement

Issue #3348, opened 2021-10-08
Not enough mesh peers in attestation subnets


dapplion:
**Describe the bug**

Contabo-6 with 16 validators should have peers in at least 32 subnets at all times. However the behavior is very erratic. Sometimes it has peers in a single subnet, then it has peers in all subnets. I added a new chart with equation
```
count(lodestar_gossip_mesh_peers_by_beacon_attestation_subnet > 0)
```
which shows the total count of subnets with 1 peer or more

![Screenshot from 2021-10-08 18-25-34](https://user-images.githubusercontent.com/35266934/136583738-b5cf3419-ec74-45e9-abc5-ec08d212b788.png)

![Screenshot from 2021-10-08 18-22-34](https://user-images.githubusercontent.com/35266934/136583748-7d809d37-cc80-4605-a477-8bf86bcbf399.png)

**Expected behavior**

Stay connected to a stable number of subnets

twoeths:
should be fixed by #3403

dapplion:
> should be fixed by #3403

I'm not sure, it will fix it. We need more data

dapplion:
Duplicate of https://github.com/ChainSafe/lodestar/issues/3555

PR #3397, opened 2021-10-29
Validator to submit attestations when seeing new head

twoeths:
**Motivation**

+ We want to submit attestation as long as validator see new head or 1/3 slot, whichever comes first

**Description**

+ this does not resolve the missing head metric of validator monitor completely, just an improvement
+ need to work on having subnet peers in order to resolve that, see #3348

### DIFF
--- a/packages/lodestar/src/chain/validation/attestation.ts
+++ b/packages/lodestar/src/chain/validation/attestation.ts
@@ -172,8 +172,6 @@ export async function validateGossipAttestation(
  * Note: We do not queue future attestations for later processing
  */
 export function verifyPropagationSlotRange(chain: IBeaconChain, attestationSlot: Slot): void {
-  // don't want to have the check for dev network with small SECONDS_PER_SLOT
-  const maximumGossipClockDisparity = chain.config.SECONDS_PER_SLOT <= 2 ? 0 : MAXIMUM_GOSSIP_CLOCK_DISPARITY_SEC;
   // slot with future tolerance of MAXIMUM_GOSSIP_CLOCK_DISPARITY_SEC
   const latestPermissibleSlot = chain.clock.slotWithFutureTolerance(MAXIMUM_GOSSIP_CLOCK_DISPARITY_SEC);
   const earliestPermissibleSlot = Math.max(
--- a/packages/lodestar/src/chain/validation/attestation.ts
+++ b/packages/lodestar/src/chain/validation/attestation.ts
@@ -172,6 +172,8 @@ export async function validateGossipAttestation(
  * Note: We do not queue future attestations for later processing
  */
 export function verifyPropagationSlotRange(chain: IBeaconChain, attestationSlot: Slot): void {
+  // don't want to have the check for dev network with small SECONDS_PER_SLOT
+  const maximumGossipClockDisparity = chain.config.SECONDS_PER_SLOT <= 2 ? 0 : MAXIMUM_GOSSIP_CLOCK_DISPARITY_SEC;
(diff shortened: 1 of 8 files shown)


packages/lodestar/src/chain/validation/attestation.ts
```diff
@@ -172,12 +172,14 @@ export async function validateGossipAttestation(
  * Note: We do not queue future attestations for later processing
  */
 export function verifyPropagationSlotRange(chain: IBeaconChain, attestationSlot: Slot): void {
+  // don't want to have the check for dev network with small SECONDS_PER_SLOT
+  const maximumGossipClockDisparity = chain.config.SECONDS_PER_SLOT <= 3 ? 0 : MAXIMUM_GOSSIP_CLOCK_DISPARITY_SEC;
```
dapplion: Why is this necessary? Does this break the sim tests?

packages/validator/src/services/attestation.ts
```diff
@@ -42,8 +44,10 @@ export class AttestationService {
     // Fetch info first so a potential delay is absorved by the sleep() below
     const dutiesByCommitteeIndex = groupAttDutiesByCommitteeIndex(this.dutiesService.getDutiesAtSlot(slot));
 
-    // Lighthouse recommends to always wait to 1/3 of the slot, even if the block comes early
-    await sleep(this.clock.msToSlotFraction(slot, 1 / 3), signal);
+    // A validator should create and broadcast the attestation to the associated attestation subnet when either
+    // (a) the validator has received a valid block from the expected block proposer for the assigned slot or
+    // (b) one-third of the slot has transpired (SECONDS_PER_SLOT / 3 seconds after the start of slot) -- whichever comes first.
+    await Promise.race([sleep(this.clock.msToSlotFraction(slot, 1 / 3), signal), this.waitForBlockSlot(slot)]);
```
dapplion, line 50: What if the block has already been seen before calling waitForBlockSlot()? You could check this edge case with the head tracker

packages/lodestar/src/chain/validation/attestation.ts
twoeths: yeah otherwise it breaks e2e and sim tests since validators submit attestations too early

packages/validator/src/services/attestation.ts
twoeths: `waitForBlockSlot` does use the head tracker, this tracks the block from the slot starts so should be safe enough

packages/lodestar/src/chain/validation/attestation.ts
dapplion: I understand but it feels too arbitrary to have this here this way. With too early, do you mean the beacon node is busy and it's internal clock is too slow?

twoeths: no, the clock is right but due to the MAXIMUM_GOSSIP_CLOCK_DISPARITY_SEC (0.5s), it can only accept attestations of current slot after `clockSlot + 0.5s` while this PR make validator submit attestations within `clockSlot` and `clockSlot + 0.5s` which is too early

twoeths: @dapplion thanks for pointing that out, it's a bug. I fixed it by implementing `slotWithFutureTolerance` and `slotWithPastTolerance` to make it clear

twoeths: merged via #3409

dapplion (approved):

PR #3409, opened 2021-11-03
Clock: Implement slotWithFutureTolerance and slotWithPastTolerance

### LINKED ISSUE #3408: Verify attestation slot range
**Describe the bug**

We want to accept more attestations given clock disparity, however currently we narrow down the chance to get attestation at https://github.com/ChainSafe/lodestar/blob/13b61a32d2aa68177cc752918af57c24290f899f/packages/lodestar/src/chain/validation/attestation.ts#L176

this is due to this implementation
```ts
slotWithTolerance(toleranceSec: number): Slot {
    return getCurrentSlot(this.config, this.genesisTime + toleranceSec);
  }
```

which may lead to the confusion that it's the tolerance for the future but actually it's for the past

**Expected behavior**
+ Implement `slotWithFutureTolerance` and `slotWithPastTolerance` api in clock to avoid the confusion

twoeths:
**Motivation**

+ Remove `slotWithTolerance` and replace with `slotWithFutureTolerance and slotWithPastTolerance`

**Description**

Closes #3408

### DIFF
--- a/packages/lodestar/src/chain/clock/LocalClock.ts
+++ b/packages/lodestar/src/chain/clock/LocalClock.ts
@@ -64,7 +64,14 @@ export class LocalClock implements IBeaconClock {
   }
 
   /** Returns the slot if the internal clock were advanced by `toleranceSec`. */
-  slotWithTolerance(toleranceSec: number): Slot {
+  slotWithFutureTolerance(toleranceSec: number): Slot {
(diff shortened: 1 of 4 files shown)


wemeetagain (approved):

dapplion (commented):
Much clearer thanks!

Issue #3171, opened 2021-09-14
Fast OOM when syncing close to head


dapplion:
**Describe the bug**

After stopping the node for a few minutes when it tries to sync to the head its memory spikes fast until it causes an OOM expection. The process is repeated after in loop.

![Screenshot from 2021-09-14 20-51-44](https://user-images.githubusercontent.com/35266934/133316975-86c82a13-5890-4b30-865b-4ee6f129c738.png)

![Screenshot from 2021-09-14 20-52-02](https://user-images.githubusercontent.com/35266934/133316999-d16226cf-3378-43b9-9360-c489bdb552a9.png)

The memory happens consistenly when the node is ~230 slots away from the head (7-8 epochs)

- The issue does **not** happen with using the flag `--weakSubjectivitySyncLatest`
- The issue does happen without validators attached

**Expected behavior**

Not OOM

**Steps to Reproduce**

Run a node close to head and let it sync with validators attached

Issue #3094, opened 2021-09-07
Reference cached beacon states with WeakRefs


dapplion:
Teku does it with good results. Since the cache memory size is technically unbounded this strategy is great for preventing OOMs.

- [ ] The head state should be kept separately with a strong reference.
- [ ] Test how JS chooses to garbage collect multiple WeakRef. By age? By least used?
- [ ] Test performance of WeakRef.unref() if it's called very often (attestation processing)
- [ ] Consider adding a shufflings cache to mitigate the risk of a key state getting dropped
- [ ] Don't regen for attestation verification if the head state already has the correct shuffling

Note: this is a huge en-devour with uncertain payoff. Teku has had issues in the past with this approach as all states disapears and the node death spirals

- Also see https://github.com/ChainSafe/lodestar/pull/3420

philknows:
Refer to #6008 . May not be required anymore.

wemeetagain:
@twoeths I don't think this is necessary anymore after n-historical-states?

twoeths:
yes, block state cache is bounded by max size and checkpoint state cache is bounded by max epoch count

PR #3420, opened 2021-11-08
Keep states in cache as WeakRef

dapplion:
**Motivation**

Lodestar has suffered serious memory issues when the state cache grows unbounded due to some bug (for example https://github.com/ChainSafe/lodestar/issues/3171)

The problem is that the state caches are unbounded memory wise. Current state cache design only limits by state count. In normal conditions that's good enough thanks to high structural sharing. However if the states are too different it will cause an out of memory (OOM) exception.

**Description**

Keep all states but the head as WeakRefs. NodeJS will garbage collect (GC) the oldest WeakRefs when the app's memory is considered too high.

Script `packages/lodestar/test/perf/weakrefOrder.ts` demonstrates the behavior of WeakRef (tested in NodeJS v14 and v16).
- WeakRef are GC'ed strictly by creation order
- For a WeakRef to become GC-able the macro queue must tick at least once
- After calling `WeakRef.deref()`, for that WeakRef to become GC-able the macro queue must tick at least once

### DIFF
--- a/packages/lodestar/src/chain/blocks/importBlock.ts
+++ b/packages/lodestar/src/chain/blocks/importBlock.ts
@@ -14,18 +14,20 @@ import {IMetrics} from "../../metrics";
 import {IEth1ForBlockProduction} from "../../eth1";
 import {IExecutionEngine} from "../../executionEngine";
 import {IBeaconDb} from "../../db";
+import {ZERO_HASH_HEX} from "../../constants";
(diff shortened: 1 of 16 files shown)


dapplion:
- Too far from a reviewable state, concept is already tracked in https://github.com/ChainSafe/lodestar/issues/3094

**Do not delete branch plz**

Issue #3423, opened 2021-11-09
Bootnode and P2P discovery doesn't work sometimes


skylenet:
**Describe the bug**

I have a private cluster and I'm running multiple lodestar nodes. I've noticed that sometimes the bootnode connectiviy seems to work fine and lodestar gets peers. Other times it just doesn't and nothing else happens. See the following example of 2 nodes, where 1 is working fine and the other isn't discoverying any peers, even though they have the same bootnode configured. 

**Expected behavior**

Bootnode should work, or on failure I should see any logs why it didn't work. 

**Steps to Reproduce**

### Config

I'm running many lodestar clients with this config. Note that each client has it's own `/data` dir and it's not shared between them.

```
node /usr/app/node_modules/.bin/lodestar beacon
--rootDir=/data
--network.discv5.bindAddr=/ip4/0.0.0.0/udp/9000
--network.localMultiaddrs=/ip4/0.0.0.0/tcp/9000
--enr.ip=10.244.6.73
--enr.tcp=9000
--enr.udp=9000
--api.rest.enabled=true
--api.rest.host=0.0.0.0
--api.rest.port=9596
--metrics.enabled=true
--metrics.listenAddr=0.0.0.0
--metrics.serverPort=8008
--genesisStateFile=/data/testnet_spec/genesis.ssz
--paramsFile=/data/testnet_spec/config.yaml
--network.discv5.bootEnrs=enr:-Ly4QPfI8vDLJRr5z14BBEVq1bPq07BhNbCcyftRUhmFf1eWR2Wn_VVnsEelOzYecvJBROTKUsxotWYO3CKPHcmwQIQBh2F0dG5ldHOIAAAAAAAAAACEZXRoMpAy95SQAQAQIDSPAAAAAAAAgmlkgnY0gmlwhAr0BsCJc2VjcDI1NmsxoQM-99Lt2OAst8ZF343Up2AGD_dOY1BzO85p2bGjm__AgohzeW5jbmV0cwCDdGNwgiMog3VkcIIjKA
--network.connectToDiscv5Bootnodes=true
--eth1.enabled
--eth1.providerUrls=http://geth.ethereum-private.svc.cluster.local:8545
--logLevel=silly
--eth1.depositContractDeployBlock=0
```

### Working node that was able to discover peers

[lodestar_works.log](https://github.com/ChainSafe/lodestar/files/7505862/lodestar_works.log)

### Broken node that failed discovering peers

[...]

PR #3429, opened 2021-11-12
Discv5: Do not issue FINDNODES right after the start

twoeths:
**Motivation**

Sometimes we're not able to send/receive discv5 packets successfully right after it starts. Ideally we should investigate discv5 for that, but I found we all `await` for the discv5 start process correctly.

```
2021-11-11T11:44:03.605Z discv5:service Starting discv5 service with node id a94979c6e3449280c849ddfcde174ddadd3cd4db280193d28743aeeef291a3fb
2021-11-11T11:44:03.605Z discv5:sessionService Starting session service with node id a94979c6e3449280c849ddfcde174ddadd3cd4db280193d28743aeeef291a3fb
2021-11-11T11:44:03.611Z discv5:service Starting a new lookup. Id: 1
```

**Description**

+ Skip doing `FINDNODES` in the 1st heart beat of peer manager

part of #3423

### DIFF
--- a/packages/lodestar/src/network/peers/discover.ts
+++ b/packages/lodestar/src/network/peers/discover.ts
@@ -19,6 +19,7 @@ const MAX_CACHED_ENR_AGE_MS = 5 * 60 * 1000;
 
 export type PeerDiscoveryOpts = {
   maxPeers: number;
+  discv5FirstQueryDelayMs: number;
   discv5: Omit<IDiscv5DiscoveryInputOptions, "metrics" | "searchInterval" | "enabled">;
 };
 
@@ -86,6 +87,8 @@ export class PeerDiscovery {
 
   /** The maximum number of peers we allow (exceptions for subnet peers) */
   private maxPeers: number;
+  private discv5StartMs: number;
+  private discv5FirstQueryDelayMs: number;
 
   constructor(modules: PeerDiscoveryModules, opts: PeerDiscoveryOpts) {
     const {libp2p, peerRpcScores, metrics, logger, config} = modules;
@@ -95,6 +98,8 @@ export class PeerDiscovery {
     this.logger = logger;
     this.config = config;
     this.maxPeers = opts.maxPeers;
+    this.discv5StartMs = 0;
+    this.discv5FirstQueryDelayMs = opts.discv5FirstQueryDelayMs;
 
     this.discv5 = Discv5.create({
       enr: opts.discv5.enr,
@@ -119,6 +124,7 @@ export class PeerDiscovery {
 
   async start(): Promise<void> {
     await this.discv5.start();
+    this.discv5StartMs = Date.now();
     this.discv5.on("discovered", this.onDiscovered);
   }
 
@@ -196,6 +202,10 @@ export class PeerDiscovery {
    * Request to find peers. First, looked at cached peers in peerStore
    */
   private async runFindRandomNodeQuery(): Promise<void> {
+    // Delay the 1st query after starting discv5
+    if (Date.now() - this.discv5StartMs <= this.discv5FirstQueryDelayMs) {
+      return;
+    }
     // Run a general discv5 query if one is not already in progress
     if (this.randomNodeQuery.code === QueryStatusCode.Active) {
       this.metrics?.discovery.findNodeQueryRequests.inc({action: "ignore"});
--- a/packages/lodestar/src/network/peers/discover.ts
+++ b/packages/lodestar/src/network/peers/discover.ts
@@ -203,9 +203,11 @@ export class PeerDiscovery {
    */
   private async runFindRandomNodeQuery(): Promise<void> {
     // Delay the 1st query after starting discv5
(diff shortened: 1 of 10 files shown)


packages/lodestar/src/network/peers/peerManager.ts
```diff
@@ -113,6 +113,8 @@ export class PeerManager {
 
   private seenPeers = new Set<string>();
 
+  private firstHearbeat = true;
```
dapplion: We could generalize this into N seconds after starting discv5, and fine tune accordingly

packages/lodestar/test/e2e/network/peers/peerManager.test.ts
```diff
@@ -85,7 +85,12 @@ describe("network / peers / PeerManager", function () {
         attnetsService: mockSubnetsService,
         syncnetsService: mockSubnetsService,
       },
-      {targetPeers: 30, maxPeers: 50, discv5: null}
+      {
+        targetPeers: 30,
+        maxPeers: 50,
+        discv5: null,
+        firstHeartBeatDelayMs: 1,
```
dapplion, line 92: Miliseconds, should be 1000 not 1

packages/lodestar/test/e2e/network/network.test.ts
```diff
@@ -55,6 +55,7 @@ describe("network", function () {
       targetPeers: 1,
       bootMultiaddrs: [],
       localMultiaddrs: [],
+      firstHeartBeatDelayMs: 1,
```
dapplion: Make it 0 for clarity, on all the tests

packages/lodestar/src/network/peers/peerManager.ts
```diff
@@ -45,6 +46,11 @@ export type PeerManagerOpts = {
   targetPeers: number;
   /** The maximum number of peers we allow (exceptions for subnet peers) */
   maxPeers: number;
+  /**
+   * Delay the 1st heartbeat after starting discv5
+   * See https://github.com/ChainSafe/lodestar/issues/3423
+   */
+  firstHeartBeatDelayMs: number;
```
dapplion: Do we want to delay the whole heartbeat or only discv5 queries? The hearbeat is doing other important actions like disconnecting bad peers that is unrelated to this issue.

dapplion: Another approach is to record the start time of discv5 and in this block:

https://github.com/ChainSafe/lodestar/blob/c5cef290fd3eab076fc9872bce05d10f913e7dc0/packages/lodestar/src/network/peers/peerManager.ts#L403-L405

Check

```ts
if (Date.now() - this.discv5StartMs > this.discv5FirstQueryDelayMs) {
  // do query
}
```

dapplion: You can even move this code into the `PeerDiscovery` class to be closer to discv5 logic

twoeths: discv5 queries is part of the `heartbeat` so I delayed the whole heartbeat only at the 1st time to make it simple. The default delay in ms is 1000 (1s)

twoeths:
@dapplion even through we define `discv5FirstQueryDelayMs`, it'll always start issueing `FINDNODES` request from the 2nd heartbeat
```
021-11-15T10:10:51.530Z discv5:service Starting discv5 service with node id 5ba50119c3f282470441414915f897f587c1fb48ac2bdaef82039a9dc8053038
2021-11-15T10:10:51.531Z discv5:sessionService Starting session service with node id 5ba50119c3f282470441414915f897f587c1fb48ac2bdaef82039a9dc8053038


2021-11-15T10:11:21.539Z discv5:service Starting a new lookup. Id: 1
```

packages/lodestar/src/network/peers/discover.ts
```diff
@@ -196,6 +202,10 @@ export class PeerDiscovery {
    * Request to find peers. First, looked at cached peers in peerStore
    */
   private async runFindRandomNodeQuery(): Promise<void> {
+    // Delay the 1st query after starting discv5
+    if (Date.now() - this.discv5StartMs <= this.discv5FirstQueryDelayMs) {
+      return;
+    }
```
dapplion, line 209: Whitespace after this if please :pray:

packages/lodestar/src/network/peers/discover.ts
```diff
@@ -196,6 +202,10 @@ export class PeerDiscovery {
    * Request to find peers. First, looked at cached peers in peerStore
    */
   private async runFindRandomNodeQuery(): Promise<void> {
+    // Delay the 1st query after starting discv5
```
dapplion, line 205: Link to the Github issue that explain why this is necessary

dapplion:
> even through we define `discv5FirstQueryDelayMs`, it'll always start issueing `FINDNODES` request from the 2nd heartbeat

Makes sense, not a problem :+1: but now we can add longer delays if we need that would go beyond the second heartbeat

dapplion (approved):

PR #3430, opened 2021-11-12
Monitor validator balance

twoeths:
**Motivation**

+ We want to know income of validators per epoch/day

**Description**

+ Add `validator_monitor_prev_epoch_on_chain_balance_total` that's updated per epoch
+ Add "Correct Head Percentage" panel to Grafana

Screen Shot 2021-11-12 at 14 56 13


dapplion (approved):
Looks great!

PR #3434, opened 2021-11-16
optimistic sync p1 - attest only validated head

### LINKED ISSUE #3505: Optimistic sync: Track Execution status and don't publish/produce on a syncing head
Track the payload execution status and don't perform following duties on syncing head
- attestation
- sync committee contributions
- block

g11tech:
**Motivation**
There is a high possibility that EL is syncing while CL asks for execution payloads validation (or preparation in case of publish block).

Optimistic sync is for CL to accept the blocks even when the status is not known (tracked via `payloadStatusUnknown`).
However when the calls to EL for executing payload stops responding as syncing,  the payloads status/forkchoice needs to be updated

Checklist:
- [x] Track valid/invalid/syncing and flow the information to the forkChoice/protoarray
- [x] DONOT participate in the if the head's payload status unkown or EL failure
  - [x] block production - if merge not complete still propose a pre-merge block and keep chain going otherwise fail hard
  - [x] attestation, 
  - [x] sync committee (sort of done w.r.t what other implementations have been doing i.e. publishing signature but failing on contribution)
    - [x] publish contribution

Considered and Rejected
~~-[ ] Fail hard on invalid executePayload call after EL itself prepared payload via getPayload ~~ publishBlock will fail for now, could be escalated later
To be moved to other Task/PR:
- publish signature ?? in flux about the best way forward, LH impl context : https://discord.com/channels/595666850260713488/892088344438255616/917362756242522143
- Mark Valid , run merge block validation on being valid
- Prune invalid and reorg new head
- Serve the validated head in p2p, req/res, api
 
**Description**
UPDATE: working smoothly (producing blocks etc)  on a devnet3 node

Closes #3505


packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -375,6 +374,10 @@ export class ForkChoice implements IForkChoice {
       justifiedRoot: toHexString(state.currentJustifiedCheckpoint.root),
       finalizedEpoch: finalizedCheckpoint.epoch,
       finalizedRoot: toHexString(state.finalizedCheckpoint.root),
+
+      // Optimistic sync based status to verify/prune later
+      payloadStatusUnknown: preCachedData?.payloadStatusUnknown,
+      isMergeBlock: preCachedData?.isMergeBlock,
```
dapplion: Is it really necessary to track `isMergeBlock`? ProtoBlocks already track executionPayloadBlockHash. Would that be enough?

dapplion: Checked Lighthouse's implementation and they only use
```rs
execution_status: ExecutionStatus

pub enum ExecutionStatus {
    /// An EL has determined that the payload is valid.
    Valid(Hash256),
    /// An EL has determined that the payload is invalid.
    Invalid(Hash256),
    /// An EL has not yet verified the execution payload.
    Unknown(Hash256),
    /// The block is either prior to the merge fork, or after the merge fork but before the terminal
    /// PoW block has been found.
    ///
    /// # Note:
    ///
    /// This `bool` only exists to satisfy our SSZ implementation which requires all variants
    /// to have a value. It can be set to anything.
    Irrelevant(bool), // TODO(merge): fix bool.
}
```

https://github.com/sigp/lighthouse/pull/2686/files#diff-446884cdc88e364a89436e4af6ef3b563c32556d6900679c9117bf50e93b3b72R40

g11tech: if we wish to enumerate the `payloadStatus` field, these two fields can be combined in a structure: 
enum ('VALID','INVALID','UNKNOWN','PREMERGE','MERGEVALIDPOW','MERGEUNKNOWNPOW')

merge block need special tracking to run the `assertValidTerminalPowBlock` / TTD difficulty validation if it was skipped because of EL syncing.
we can use `executionPayloadBlockHash` for merge block tracking as its parent will have that as null (so we can skip `MERGEVALIDPOW`, `MERGEUNKNOWNPOW` but then will always need to check the parent of the first validation for merge ttd checks)

another Q: 
1. do we need to track 'INVALID' status as they would anyway be pruned? may be it will help on rethrowing critical error on restarts if the finalized/justified checkpoint block becomes invalid 
2. also what sounds more intuitive: a) PREMERGE or IRRELEVANT b) UNKNOWN or SYNCING?

dapplion: Right let's do
```ts
enum ExecutionStatus {
  Valid,
  Syncing,
  PreMerge
}

type IProtoBlock = 
  { executionStatus: ExecutionStatus.Valid | ExecutionStatus.Syncing, executionPayloadBlockHash: RootHex }
| { executionStatus: ExecutionStatus.PreMerge, executionPayloadBlockHash: null }
```

packages/fork-choice/test/unit/protoArray/protoArray.test.ts
```diff
@@ -47,11 +51,13 @@ describe("ProtoArray", () => {
...
       stateRoot,
       targetRoot: finalizedRoot,
-      executionPayloadBlockHash: null,
+
       justifiedEpoch: genesisEpoch,
       justifiedRoot: stateRoot,
       finalizedEpoch: genesisEpoch,
       finalizedRoot: stateRoot,
+
+      ...{executionPayloadBlockHash: null, executionStatus: ExecutionStatus.PreMerge},
```
dapplion, line 60: Why spread a new object instead of assigning directly?

g11tech: umm... typescript was giving error as assigning directly might not be helping it identify/validate the union conditions

dapplion: Oh makes sense :+1:

g11tech: unless the entire object is explicity typecased as `{...} as IProtoBlock`, let know what is preferable: this or the explicity typecasing

dapplion: It's a test so it doesn't really matter, just leave as is

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -133,6 +135,15 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
     }
   }
 
+  function notOnOptimisticBlockRoot(beaconBlockRoot: Root): void {
+    const protoBeaconBlock = chain.forkChoice.getBlock(beaconBlockRoot);
+    if (!protoBeaconBlock) throw Error("BLOCK_NOT_IN_FORKCHOICE");
```
dapplion: This should be a 400 bad param error

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -133,6 +135,15 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
...
+    if (protoBeaconBlock.executionStatus === ExecutionStatus.Syncing)
+      throw new Error(
+        `InvalidExecutionStatus: beaconBlockRoot ${protoBeaconBlock.blockRoot}'s payload ${protoBeaconBlock.executionPayloadBlockHash} on slot ${protoBeaconBlock.slot} is not yet validated`
```
dapplion: There's a specific code for syncing errors define in the api spec

EDIT: https://github.com/ethereum/beacon-APIs/blob/d3e4a379f4235b5912978bba5e60a9de6d872761/apis/validator/duties/attester.yaml?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L69 it's a good placeholder while we spec the specific case for the merge

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -133,6 +135,15 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
     }
   }
 
+  function notOnOptimisticBlockRoot(beaconBlockRoot: Root): void {
+    const protoBeaconBlock = chain.forkChoice.getBlock(beaconBlockRoot);
+    if (!protoBeaconBlock) throw new ApiError(400, "Block not in forkChoice");
```
dapplion: Please use multiline if syntax for better readability
```ts
if (condition) {
  statement;
}
```

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -133,6 +135,15 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
     }
   }
 
+  function notOnOptimisticBlockRoot(beaconBlockRoot: Root): void {
```
dapplion, line 158: Currently this is applied to:
- `produceBlock`
- `produceSyncCommitteeContribution`

Should it be applied to attestations too? Could you add a nice long comment at the top of `notOnOptimisticBlockRoot` explaining why we don't allow this actions?

g11tech: you mean we are doing it on `produceAttestationData` and `produceSyncCommitteeContribution` only and not on `produceBlock`? 
`produceBlock` when calls assembleBody, which in turn call EL's fetch payload sequence (notify/getpayload), the EL will revert with Syncing there if not yet synced (and get aborted). The reason it has not been done here is for a optimization that EL could get synced between the last time it responded syncing and this opportunity of produceBlock.

dapplion: My bad, miss-understood the code. Current approach looks good!

packages/lodestar/src/executionEngine/http.ts
```diff
@@ -63,20 +65,40 @@ export class ExecutionEngineHttp implements IExecutionEngine {
...
     // Validate status is known
     const statusEnum = ExecutePayloadStatus[status];
-    if (statusEnum === undefined) {
-      throw Error(`Unknown status ${status}`);
+    if (statusEnum === undefined || (statusEnum !== ExecutePayloadStatus.SYNCING && latestValidHash === undefined)) {
+      throw Error(`Invalid respose status ${status} latestValidHash: ${latestValidHash}`);
+    }
+
+    // Validate blocks in forkchoice till latestValidHash, Invalidate decendants  if response status is invalid.
+    if (forkChoice) {
```
dapplion: Is there a situation where `executePayload()` is called without `forkChoice`?

dapplion (dismissed):

dapplion (approved):

PR #3435, opened 2021-11-17
p2p: stream bytes from db

### LINKED ISSUE #2490: In ReqResp block method stream bytes from db
**Describe the bug**

Right now when serving ReqResp block methods we read from DB, deserialize from bytes to struct and then re-serialize to send over the wire.

**Expected behavior**

We could just serve the bytes stream directly from db, prevent any serdes and require no memory.

twoeths:
**Motivation**

+ When serving p2p `beacon_blocks_by_range` or `beacon_blocks_by_roots`, it's not necessary to deserialize bytes from db to SignedBeaconBlock struct and serialize it back to bytes before doing ssz snappy and response to the the other node
+ In some cases, it takes 20% of our cpu time and our validators missed chain head root due to that

**Description**

+ No need to convert to temporary `SignedBeaconBlock` objects  for `beacon_blocks_by_range` and `beacon_blocks_by_roots` requests
+ Model `P2pBlockResponse` type which contains slot and bytes data of `SignedBeaconBlock`. We need `slot` because we use it when fetching from db and to determine correct forkdigest for p2p
+ New `LodestarResponseBodyByMethod` type which use `P2pBlockResponse`  for `beacon_blocks_by_range` and `beacon_blocks_by_roots`

Closes #2490

### DIFF
--- a/packages/lodestar/src/db/repositories/blockArchive.ts
+++ b/packages/lodestar/src/db/repositories/blockArchive.ts
@@ -2,7 +2,7 @@ import all from "it-all";
 import {ArrayLike} from "@chainsafe/ssz";
 import {IChainForkConfig} from "@chainsafe/lodestar-config";
 import {IDatabaseController, Repository, IKeyValue, IFilterOptions, Bucket, IDbMetrics} from "@chainsafe/lodestar-db";
-import {Slot, Root, allForks, ssz, ReqRespBlockResponse} from "@chainsafe/lodestar-types";
+import {Slot, Root, allForks, ssz} from "@chainsafe/lodestar-types";
 import {bytesToInt} from "@chainsafe/lodestar-utils";
 import {getSignedBlockTypeFromBytes} from "../../util/multifork";
 import {getRootIndexKey, getParentRootIndexKey} from "./blockArchiveIndex";
@@ -16,6 +16,13 @@ export type BlockArchiveBatchPutBinaryItem = IKeyValue<Slot, Buffer> & {
   slot: Slot;
   blockRoot: Root;
   parentRoot: Root;
+};
+
+// TODO: remove once we remove reqRespBlockStream
+export type BlockEntry = {
+  /** Deserialized data of allForks.SignedBeaconBlock */
+  bytes: Buffer;
+  slot: Slot;
 };
 
 /**
@@ -101,7 +108,8 @@ export class BlockArchiveRepository extends Repository<Slot, allForks.SignedBeac
     ]);
   }
 
-  async *reqRespBlockStream(opts?: IBlockFilterOptions): AsyncIterable<ReqRespBlockResponse> {
+  // TODO: only support step 1
+  async *reqRespBlockStream(opts?: IBlockFilterOptions): AsyncIterable<BlockEntry> {
     const firstSlot = this.getFirstSlot(opts);
     const binaryEntriesStream = super.binaryEntriesStream(opts);
     const step = (opts && opts.step) || 1;
--- a/packages/lodestar/src/db/repositories/blockArchive.ts
+++ b/packages/lodestar/src/db/repositories/blockArchive.ts
@@ -2,9 +2,9 @@ import all from "it-all";
 import {ArrayLike} from "@chainsafe/ssz";
 import {IChainForkConfig} from "@chainsafe/lodestar-config";
 import {IDatabaseController, Repository, IKeyValue, IFilterOptions, Bucket, IDbMetrics} from "@chainsafe/lodestar-db";
-import {Slot, Root, allForks, ssz} from "@chainsafe/lodestar-types";
+import {Slot, Root, allForks, ssz, P2pBlockResponse} from "@chainsafe/lodestar-types";
 import {bytesToInt} from "@chainsafe/lodestar-utils";
-import {getSignedBlockTypeFromBytes} from "../../util/multifork";
+import {getSignedBlockTypeFromBytes, getSlotFromBytes} from "../../util/multifork";
 import {getRootIndexKey, getParentRootIndexKey} from "./blockArchiveIndex";
 import {deleteParentRootIndex, deleteRootIndex, storeParentRootIndex, storeRootIndex} from "./blockArchiveIndex";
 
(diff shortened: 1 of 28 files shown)


packages/lodestar/src/db/repositories/blockArchive.ts
```diff
@@ -101,14 +101,21 @@ export class BlockArchiveRepository extends Repository<Slot, allForks.SignedBeac
     ]);
   }
 
+  async *p2pBlockStream(opts?: IBlockFilterOptions): AsyncIterable<P2pBlockResponse> {
+    const firstSlot = this.getFirstSlot(opts);
+    const binaryValuesStream = super.binaryValuesStream(opts);
```
dapplion: You can stream entries as bytes and get the slot from the key and well as the value in bytes. This should prevent the need for `getSlotFromBytes` on the archive repository

packages/types/src/types.ts
```diff
@@ -7,3 +9,10 @@ export {ts as allForks} from "./allForks";
 
 /** Common non-spec type to represent roots as strings */
 export type RootHex = string;
+
+/** This type helps response to beacon_block_by_range and beacon_block_by_root more efficiently */
+export type P2pBlockResponse = {
```
dapplion: Since this is restricted to ReqResp only and not gossip you can rename to `ReqRespBlockResponse`

packages/lodestar/src/network/reqresp/types.ts
```diff
@@ -141,20 +141,33 @@ export function getResponseSzzTypeByMethod(protocol: Protocol, forkName: ForkNam
...
+type CommonResponseBodyByMethod = {
   [Method.Status]: phase0.Status;
   [Method.Goodbye]: phase0.Goodbye;
   [Method.Ping]: phase0.Ping;
   [Method.Metadata]: phase0.Metadata;
+};
+
+// Used internally by lodestar to response to beacon_blocks_by_range and beacon_blocks_by_root
+// without having to deserialize and serialize from/to bytes
+export type LodestarResponseBodyByMethod = CommonResponseBodyByMethod & {
```
dapplion: Instead of Lodestar you can denote this with `Incoming` vs `Outgoing` responses.
- On an `Incoming` response we deserialize
- On an `Outgoing` response we do not deserialize

packages/lodestar/src/network/reqresp/encodingStrategies/sszSnappy/encode.ts
```diff
@@ -1,21 +1,23 @@
...
  * ```
  */
-export async function* writeSszSnappyPayload<T extends RequestOrResponseBody>(
+export async function* writeSszSnappyPayload<T extends RequestOrLodestarResponseBody>(
   body: T,
-  type: RequestOrResponseType
+  type: RequestOrResponseType | null
 ): AsyncGenerator<Buffer> {
-  const serializedBody = serializeSszBody(body, type);
+  const serializedBody = type === null ? (body as P2pBlockResponse).bytes : serializeSszBody(body, type);
```
dapplion: Instead of nulling the type you can define the type as an object that
```ts
const reqRespOutgoingResponseBlockType = {
  serialize(serializedBlockWrapper: SerializedBlockWrapper) {
    return serializedBlockWrapper.bytes
  }
}
```

packages/lodestar/src/network/reqresp/encoders/responseEncode.ts
```diff
@@ -26,20 +27,23 @@ import {
...
+      // <context-bytes> - from altair
       const forkName = getForkNameFromResponseBody(config, protocol, chunk);
       yield* writeContextBytes(config, contextBytesType, forkName);
 
       // <encoding-dependent-header> | <encoded-payload>
-      const type = getResponseSzzTypeByMethod(protocol, forkName);
+      const type =
+        protocol.method === Method.BeaconBlocksByRange || protocol.method === Method.BeaconBlocksByRoot
+          ? null
+          : getResponseSzzTypeByMethod(protocol, forkName);
```
dapplion: Adding this information here breaks the abstracting layer of the encoders decoders. I would propose to move this info inside `getResponseSzzTypeByMethod`, and here do
```ts
getResponseSzzTypeByMethod(protocol, forkName, Direction.Outgoing)
```

Then edit
```ts
export function getResponseSzzTypeByMethod(protocol: Protocol, forkName: ForkName, direction: Direction) {
  switch (protocol.method) {
    ....
      // SignedBeaconBlock type is changed in altair
      switch (Direction.Outgoing) {
        case Direction.Incoming:
          return ssz[forkName].SignedBeaconBlock;
        case Direction.Outgoing:
          return null;
          // or
          return bytesToBytesType
  }
}
```
This way the definition is packed together in `packages/lodestar/src/network/reqresp/types.ts`. If you think this approach is more convoluted let me know.

packages/lodestar/src/db/repositories/blockArchive.ts
```diff
@@ -101,14 +101,21 @@ export class BlockArchiveRepository extends Repository<Slot, allForks.SignedBeac
...
+    const step = (opts && opts.step) || 1;
+
+    for await (const bytes of binaryValuesStream) {
+      const slot = getSlotFromBytes(bytes);
+      if ((slot - firstSlot) % step === 0) {
+        yield {bytes, slot};
```
dapplion: The reasoning for attaching the slot here is to get the fork latter when constructing the response? Or is there other usages

twoeths: @dapplion  it's just to get the fork later

packages/lodestar/src/network/reqresp/encoders/responseEncode.ts
twoeths: Modifying `getResponseSzzTypeByMethod()` would escalate the change to `responseDecode` and other type definition that get return type from this method so I implement the change to a new method `getOutgoingSerializerByMethod()`

packages/types/src/types.ts
```diff
@@ -7,3 +9,10 @@ export {ts as allForks} from "./allForks";
 
 /** Common non-spec type to represent roots as strings */
 export type RootHex = string;
+
+/** This type helps response to beacon_block_by_range and beacon_block_by_root more efficiently */
+export type ReqRespBlockResponse = {
```
dapplion: The `types` package should only contain types defined in the spec. `RootHex` is the exception since it's used in lodestar, fork-choice and validator. Since `ReqRespBlockResponse` is a type that only exist and makes sense in reqResp, I would define it in `network/reqResp`.

dapplion (commented):
Looks great! Thank you!

Just a minor comment

dapplion (approved):
Excellent!

PR #3440, opened 2021-11-18
Update nodejs to v16 - gallium

wemeetagain:
**Motivation**

The active LTS is now v16. It could be useful to upgrade our tooling to use v16.
We've seen some internal reports that heap snapshots are more reliable on v16 than v14.

**Description**

Update node v14 -> v16. (fermium -> gallium)


twoeths (dismissed):
yay it's a good idea to go with NodeJS 16, I've used it for a while

twoeths:
@wemeetagain some sim tests failed

dapplion (approved):

Issue #3416, opened 2021-11-05
Performance issue with subscribeAllSubnets flag


twoeths:
**Describe the bug**

If we start a node with `--network.subscribeAllSubnets` flag, it may take up to 40s to process a gossip block

<img width="1614" alt="Screen Shot 2021-11-05 at 17 20 57" src="https://user-images.githubusercontent.com/10568965/140495988-0d371776-3fd5-4128-b545-818f66547caf.png">

**Expected behavior**
No performance issue with  `--network.subscribeAllSubnets` flag

twoeths:
[1105_contabo_1_subscribeAllSubnets.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/7482539/1105_contabo_1_subscribeAllSubnets.cpuprofile.zip)

twoeths:
Looks like it's a memory issue that cause the node pause more frequently due to gc

<img width="1587" alt="Screen Shot 2021-11-08 at 15 06 31" src="https://user-images.githubusercontent.com/10568965/140705791-234e56d9-30f2-4fe8-a651-69df703ce0c9.png">

gc pause time is almost 15% while it's 8.5% on normal node

twoeths:
one other issue was with throwing GossipActionError as noted in #3402, it took > 4% just to throw that

<img width="1481" alt="Screen Shot 2021-11-08 at 16 21 20" src="https://user-images.githubusercontent.com/10568965/140716138-f0452e38-43b4-42de-9699-7402a4ed2703.png">

dapplion:
@tuyennhv In the profile that you posted there's an insane amount of CPU time spent with async hook
- `registerDestroyHook`
- `promiseInitHookWithDestroyTrackinginternal`

![Screenshot from 2021-11-08 14-56-15](https://user-images.githubusercontent.com/35266934/140754439-079f034c-513f-449d-9561-d7df55b81780.png)


Could you investigate what's going on there?

twoeths:
> @tuyennhv In the profile that you posted there's an insane amount of CPU time spent with async hook
> registerDestroyHook
> promiseInitHookWithDestroyTrackinginternal

[...]

Issue #3402, opened 2021-10-30
Refactor gossip validation to return instead of throwing


dapplion:
**Describe the bug**

throw + catch is x1000 times more expensive than returning objects. See https://github.com/ChainSafe/lodestar/pull/3401

dapplion:
Closing as this is has no noticeable impact on recent benchmarks. The return not throw flow change is useful for doing batch verification of attestations, so implement there if needed

PR #3448, opened 2021-11-20
[DNM] Demo throwing AttestationError performance

twoeths:
**Motivation**

This profile with `--subscribeAllSubnets` flag shows that creating AttestationError takes 4% of the cpu time
Screen Shot 2021-11-20 at 16 38 10

**Description**

Throwing a regular object could improve the speed 500%:
```
test throw AttestationError vs object
    ✓ AttestationError                                                    115.3952 ops/s    8.665875 ms/op        -        197 runs   2.21 s
    ✓ Object                                                              578.6963 ops/s    1.728022 ms/op        -        410 runs   1.21 s

```
+ found from #3416
+ this is another option to improve performance compared to #3402

### DIFF
--- a/packages/lodestar/test/perf/chain/validation/attestation.test.ts
+++ b/packages/lodestar/test/perf/chain/validation/attestation.test.ts
@@ -4,6 +4,7 @@ import {validateGossipAttestation} from "../../../../src/chain/validation";
 import {generateTestCachedBeaconStateOnlyValidators} from "@chainsafe/lodestar-beacon-state-transition/test/perf/util";
 import {getAttestationValidData} from "../../../utils/validationData/attestation";
 import {AttestationError, AttestationErrorCode, GossipAction} from "../../../../src/chain/errors";
+import {ZERO_HASH} from "../../../../src/constants";
 
 describe("validate gossip attestation", () => {
   const vc = 64;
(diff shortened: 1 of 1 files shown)


dapplion:
@tuyennhv I did some more analysis on this and we should benchmark async operations not sync. The difference here is not throw X vs throw Y vs return, but reject X vs reject Y vs resolve.

twoeths:
> more analysis on this and we should benchmark async operations not sync. The difference here is not throw X vs throw Y vs r

@dapplion I just want to show that it takes time to construct LodestarError (4% of the time), if we throw/reject a regular object it's much better

dapplion:
> > more analysis on this and we should benchmark async operations not sync. The difference here is not throw X vs throw Y vs r
> 
> @dapplion I just want to show that it takes time to construct LodestarError (4% of the time), if we throw/reject a regular object it's much better

Tho promise handling is relatively expensive too, it would be very useful to have async comparisions of rejecting error vs rejecting object

twoeths:
@dapplion the async benchmark test also shows that it's a 5x improvement

```
test throw AttestationError vs object
    ✓ AttestationError                                                    63.24967 ops/s    15.81036 ms/op        -         57 runs   1.41 s
    ✓ Object                                                              315.5363 ops/s    3.169207 ms/op        -       1112 runs   4.04 s
```

packages/lodestar/test/perf/chain/validation/attestation.test.ts
```diff
@@ -26,3 +27,46 @@ describe("validate gossip attestation", () => {
     });
   }
 });
+
+describe.only("test throw AttestationError vs object", () => {
+  itBench({
+    id: "AttestationError",
+    fn: async () => {
```
dapplion, line 36: If you want to loop for 1000 times, use `runsFactor: 1000`. You should compare the results of looping 1000 times vs not looping. It's important to know the actual time of each operation.

packages/lodestar/test/perf/chain/validation/attestation.test.ts
```diff
@@ -26,3 +27,46 @@ describe("validate gossip attestation", () => {
     });
   }
 });
+
+describe.only("test throw AttestationError vs object", () => {
+  itBench({
+    id: "AttestationError",
```
dapplion: The `id` of the benchmark must be globally unique in the entire Lodestar monorepo. This ID will be displayed in PRs to it needs more context. For example:
- `Reject Error x1000`
- `Reject Error x1`
- `Reject object x1000`
- `Reject object x1`

packages/lodestar/test/perf/chain/validation/attestation.test.ts
```diff
@@ -26,3 +27,46 @@ describe("validate gossip attestation", () => {
...
+  if (b) {
+    throw new AttestationError(GossipAction.IGNORE, {
+      code: AttestationErrorCode.UNKNOWN_BEACON_BLOCK_ROOT,
+      root: Buffer.alloc(32),
+    });
+  }
+}
+
+async function throwObject(b: boolean): Promise<void> {
+  if (b) {
```
dapplion: b is always true, remove

packages/lodestar/test/perf/chain/validation/attestation.test.ts
```diff
@@ -26,3 +27,46 @@ describe("validate gossip attestation", () => {
...
+      }
+    }
+  });
+});
+
+async function throwAttestationError(b: boolean): Promise<void> {
+  if (b) {
+    throw new AttestationError(GossipAction.IGNORE, {
+      code: AttestationErrorCode.UNKNOWN_BEACON_BLOCK_ROOT,
+      root: Buffer.alloc(32),
```
dapplion: Don't allocate bytes here, this will distort the test's results. Declare this root above once for all tests or use ZERO_HASH constant

twoeths:
@dapplion  latest statistic in my local environment:

```
test throw AttestationError vs object
    ✓ AttestationError x1000                                              88479.32 ops/s    11.30208 us/op        -        123 runs   1.89 s
    ✓ Object x1000                                                        510569.6 ops/s    1.958597 us/op
```

twoeths:
This came at the time when our node had a lot of performance issues, so when we validate gossip messages there are a lot of AttestationErrors. This PR suggested a more performant way to throw errors.

As of now, the performance is so much better so there are so few AttestationErrors, closing for now.

dapplion:
This PR still shows returning objects vs throwing errors is more efficient. We should eventually adopt that, is there any issue for it?

twoeths:
> This PR still shows returning objects vs throwing errors is more efficient. We should eventually adopt that, is there any issue for it?

I created https://github.com/ChainSafe/lodestar/issues/3890, maybe a low priority as we don't have a lot of gossip errors like in the past

Issue #3344, opened 2021-10-08
Implement self rate-limiting


dapplion:
**Describe the bug**

In merge-interop Teku constantly bans us due to rate-limiting. Teku self rate limits and we should do so with some sane params taken from other clients. Causes https://github.com/ChainSafe/lodestar/issues/2622

**Expected behavior**

Prevent rate limiting when connected to few peers

twoeths:
Got this in Prater at the non-finality time:

```
beacon-2022-01-06.log:623:Jan-06 00:04:35.885 [NETWORK]       verbose: Req  error method=beacon_blocks_by_root, encoding=ssz_snappy, client=Prysm, peer=16...3rfnWg, requestId=1766 code=RESPONSE_STATUS_ERROR, status=1, errorMessage=sNaPpYuV~2rate limited
```

twoeths:
Refer to https://github.com/ethereum/consensus-specs/pull/2690

philknows:
If we spam other nodes, it'll eventually work itself out. Low priority.

PR #3454, opened 2021-11-23
Implement rate limit

### LINKED ISSUE #3451: Implement rate limit

We don't want to serve `beacon_blocks_by_range` and `beacon_blocks_by_root` too much

+ Per peer: Allow up to x requests per period
+ Per peer: Allow up to x blocks per period
+ Across peers: Allow up to x requests per period
+ Across peers: Allow up to x blocks per period

Reference params from Teku:
+ Per peer: up to 50 requests per 1 minute
+ Per peer: up to 500 blocks per 1 minute

twoeths:
**Motivation**

+ At some points, there are a lot of peers connect to us and it takes 20% of our CPU time to serve block requests. #3435 should mitigate the issue and this is the 2nd part
+ ~~We want to apply rate limit when we receive `beacon_blocks_by_range` and `beacon_blocks_by_root` requests~~
+ We want to apply **block count** rate limit when we receive beacon_blocks_by_range and beacon_blocks_by_root requests and **request count** rate limit for all request types

**Description**

+ Implement a `RateTracker` class that check the rate limit of request count and block count
+ Added 4 params to network: per peer `requestCountPeerLimit, blockCountPeerLimit` and total `requestCountTotalLimit, blockCountTotalLimit`
+ Ban peer if it violates per peer limit
+ By default, we limit 500 blocks and 50 requests per peer per minute or the peer is banned, make the `total` params 4x the `peer` params

```typescript
  requestCountTotalLimit: 200,
  requestCountPeerLimit: 50,
  blockCountTotalLimit: 2000,
  blockCountPeerLimit: 500,
  rateTrackerTimeoutMs: 60 * 1000,
```

**Note**
We will need to apply `RateTracker` when we issue rpc block requests too or we'll get banned by other peers, see #3344

Closes #3451


shanghan345:
98982021年11月23日 14:14，tuyennhv ***@***.***>寫道：Motivation
At some points, there are a lot of peers connect to us and it takes 20% of our CPU time to serve block requests. #3435 should mitigate the issue and this is the 2nd partWe want to apply rate limit when we receive beacon_blocks_by_range and beacon_blocks_by_root requests
Description
Implement a RateTracker class that check the rate limit of request count and block countAdded 4 params to network: requestCountTotalLimit, requestCountPeerLimit, blockCountTotalLimit, blockCountPeerLimitBy default, we limit 500 blocks and 5 requests per peer per minute or the peer is banned, make the total params 4x the peer params
Note
We will need to apply RateTracker when we issue rpc block requests too or we'll get banned by other peers, see #3344
Closes #3451



You can view, comment on, or merge this pull request online at:
  https://github.com/ChainSafe/lodestar/pull/3454

Commit Summary
fe361d3  Implement Req Resp Rate Limit8f221b0  Make total params 4x peer params

File Changes  (13 files)

    M
    packages/cli/src/options/beaconNodeOptions/network.ts
    (45)
  
    M
    packages/cli/test/unit/options/beaconNodeOptions.test.ts
    (11)
  
    M
    packages/lodestar/src/network/network.ts
    (3)
  
    M
    packages/lodestar/src/network/options.ts
    (11)
  
    M
    packages/lodestar/src/network/reqresp/interface.ts
    (2)
  
    M
    packages/lodestar/src/network/reqresp/reqResp.ts
    (15)
  
    A
    packages/lodestar/src/network/reqresp/response/rateTracker.ts
    (120)
  
    M
    packages/lodestar/test/e2e/network/gossipsub.test.ts
    (3)
  
    M
    packages/lodestar/test/e2e/network/network.test.ts
    (3)
  
    M
    packages/lodestar/test/e2e/network/reqresp.test.ts
    (3)
  
    M
    packages/lodestar/test/e2e/sync/finalizedSync.test.ts
    (8)
  
    M
    packages/lodestar/test/e2e/sync/unknownBlockSync.test.ts
    (8)
  
    A
    packages/lodestar/test/unit/network/reqresp/response/rateTracker.test.ts
    (119)
  

Patch Links:
https://github.com/ChainSafe/lodestar/pull/3454.patchhttps://github.com/ChainSafe/lodestar/pull/3454.diff

packages/lodestar/src/network/reqresp/rateTracker.ts
```diff
@@ -0,0 +1,47 @@
+import {MapDef} from "../../util/map";
+
+/**
+ * The generic rate tracker allows up to `limit` objects in a period of time.
+ * This could apply to both request count or block count, for both requests and responses.
+ */
+export class RateTracker {
+  private requestsWithinWindow = 0;
+  private requests = new MapDef<number, number>(() => 0);
```
dapplion: What do each number in `<number, number>` represent? Add JSDocs to `requests` explaining

packages/lodestar/src/network/reqresp/reqResp.ts
```diff
@@ -218,31 +217,43 @@ export class ReqResp implements IReqResp {
 
     switch (requestTyped.method) {
       case Method.Ping:
+        if (!this.responseRateLimiter.allowToProcess(peerId)) {
+          throw new ResponseError(RespStatus.SERVER_ERROR, "rate limit");
+        }
```
dapplion: Why not call once at the top of this function with 
```ts
       if (!this.responseRateLimiter.allowToProcess(protocol.method, peerId)) {
          throw new ResponseError(RespStatus.SERVER_ERROR, "rate limit");
        }
```

dapplion: Also, please review or ask Teku and Lighthouse if they use a different error code for rate limiting. I've seen before an unoficial list they mantain with different frequent error codes. To potentially replace the generic `SERVER_ERROR`

twoeths: @dapplion Teku uses `INVALID_REQUEST`, we'll follow Lighthouse to use new `RATE_LIMITED = 139` error code

twoeths: regarding `allowToProcess()` call, `BeaconBlocksByRange` and `BeaconBlocksByRoot` put in different param (block count) compared to others

packages/lodestar/src/network/reqresp/response/rateLimiter.ts
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Allow to process requests from peer or not based on the rate limit params configured.
+   * @param numBlock only apply for beacon_blocks_by_range and beacon_blocks_by_root
+   */
+  allowToProcess(peerId: PeerId, numBlock?: number): boolean {
+    const peerIdStr = peerId.toB58String();
+    const requestCountPeerTracker = this.requestCountTrackersByPeer.getOrDefault(peerIdStr);
+    const blockCountPeerTracker = this.blockCountTrackersByPeer.getOrDefault(peerIdStr);
+
+    const rateTrackers = [
```
dapplion: This implementation requires to create a new array on every request, adding GC cost. This arrays should be pre-computed. Why not pass `requestTyped` to `allowToProcess()` and do a switch here?

packages/lodestar/src/network/reqresp/response/rateLimiter.ts
```diff
@@ -0,0 +1,90 @@
...
+            name: "blockCountTotalTracker",
+          },
+          {tracker: blockCountPeerTracker, count: numBlock, applyPenalty: true, name: "blockCountPeerTracker"},
+        ]
+      );
+    }
+
+    for (const {tracker, count, applyPenalty, name} of rateTrackers) {
+      if (tracker.requestObjects(count) === 0) {
+        this.logger.warn("Do not serve request due to rate limit", {
```
dapplion: Should this be a warn? A peer could spam us but the user should not see a warning since there's nothing the user can do about it, and it's "expected behavior"?

packages/lodestar/src/network/reqresp/rateTracker.ts
```diff
@@ -0,0 +1,48 @@
...
+  requestObjects(objectCount: number): number {
+    if (objectCount <= 0) throw Error("Invalid objectCount " + objectCount);
+    this.prune();
+    if (this.requestsWithinWindow >= this.limit) {
+      return 0;
+    }
+
+    this.requestsWithinWindow += objectCount;
+    const now = Date.now();
+    const curObjectCount = this.requests.getOrDefault(now);
```
dapplion: `curObjectCount` equals to the count of objects in that exact ms? So we are limiting how many objects can happen in a per milisecond basis?

twoeths: @dapplion 
> curObjectCount equals to the count of objects in that exact ms

yes, we could switch to `second` instead to reduce the memory through

> So we are limiting how many objects can happen in a per milisecond basis?

the quota depends on the `timeoutMs` in the constructor of this class, not the key of the `this.request` map. Again, I think I should switch to tracking by `second` instead of `ms` to make it simpler and take less memory

packages/lodestar/src/network/reqresp/response/rateLimiter.ts
twoeths: I'll switch to `verbose` instead 👍

dapplion:
I'm concerned about the memory and performance cost of this implementation.
- Could you do a realistic analysis of memory + performance cost? Ideally both theoretical with ~25 peers or measuring in benchmarks
- What also the expected memory cost of all data structures as a function of peers?
- Do all data structures tied to peers get clean-up properly?

packages/lodestar/src/network/reqresp/rateTracker.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  private prune(): void {
+    const now = Date.now();
+
+    for (const [timeInSec, count] of this.requests.entries()) {
+      // reclaim the quota for old requests
+      if (now - timeInSec * 1000 >= this.timeoutMs) {
+        this.requestsWithinWindow -= count;
+        this.requests.delete(timeInSec);
+      }
```
wemeetagain, line 58: I think we can break out of the loop early?

``` ts
// Break after the first entry within the timeout window.
// Since the entries are added in order, all the rest will be within the window.
else break;
```

packages/lodestar/src/network/reqresp/response/rateLimiter.ts
```diff
@@ -0,0 +1,87 @@
...
+    );
+    this.logger = modules.logger;
+    this.peerRpcScores = modules.peerRpcScores;
+  }
+
+  /**
+   * Allow to process requests from peer or not based on the rate limit params configured.
+   * @param numBlock only apply for beacon_blocks_by_range and beacon_blocks_by_root
+   */
+  allowToProcess(peerId: PeerId, numBlock?: number): boolean {
```
wemeetagain: maybe call this `allowRequest`?

Also tweak to the jsdoc
```
/**
 * Tracks a request from a peer and returns whether to allow the request based on the configured rate limit params.
 * @param numBlock only applies to beacon_blocks_by_range and beacon_blocks_by_root
 */
```

packages/lodestar/src/network/reqresp/response/rateLimiter.ts
```diff
@@ -0,0 +1,87 @@
...
+  requestCountPeerLimit: number;
+  blockCountTotalLimit: number;
+  blockCountPeerLimit: number;
+  rateTrackerTimeoutMs: number;
+};
+
+/**
+ * The rate tracker for all peers.
+ */
+export class ResponseRateLimiter implements IReqRespRateLimiter {
```
wemeetagain: Maybe call this `InboundRateLimiter`

packages/lodestar/src/network/reqresp/interface.ts
```diff
@@ -69,3 +70,13 @@ export type Libp2pStream = {
    */
   abort: (err: Error) => void;
 };
+
+/**
+ * Rate limiter interface for request and response.
```
wemeetagain: Rate limiter interface for inbound and outbound requests

packages/lodestar/src/network/reqresp/interface.ts
```diff
@@ -69,3 +70,13 @@ export type Libp2pStream = {
...
+ */
+export interface IReqRespRateLimiter {
+  /**
+   * Allow to request or response based on rate limit params configured.
+   */
+  allowToProcess(peerId: PeerId, objectCount?: number): boolean;
```
wemeetagain: see the comment below

wemeetagain (changes requested):

packages/lodestar/src/network/reqresp/rateTracker.ts
twoeths: yeah should break early there 👍

twoeths:
Update with some analysis and benchmark tests:

**Memory**
+ Rate limiter can keep up to 15MB (this includes `25 + 25` rate trackers for peer and 2 `total` rate tracker, each is full of requests (60)) as shown in the test
+ We prune rate limiter when a peer is disconnected so memory will not be higher than that
+ I leave the machine for 3h, take a snapshot and it takes only 125kb
Screen Shot 2021-12-07 at 14 18 47

+ We don't allocate resource per request so no need to cleanup


**Performance**
+ A `rateTracker.requestObjects()` call takes from `80ns/op` to `160ns/op` if it does not have to prune. If it has to prune all old requests, it takes around `3.7 us/op` but this rarely happen since the next time it does not have to prune that much
+ A `rateLimiter.allowRequest()` call include `4x` or `2x` `rateTracker.requestObjects()` call


@dapplion please let me know if you have any concern, or any other tests that I can add

packages/lodestar/src/network/reqresp/reqResp.ts
```diff
@@ -215,26 +222,45 @@ export class ReqResp implements IReqResp {
...
+        if (!this.inboundRateLimiter.allowRequest(peerId)) {
+          throw new ResponseError(RespStatus.RATE_LIMITED, "rate limit");
+        }
         // V1 -> phase0, V2 -> altair. But the type serialization of phase0.Metadata will just ignore the extra .syncnets property
         // It's safe to return altair.Metadata here for all versions
         yield this.metadataController.json;
         break;
       case Method.Goodbye:
+        // no need to call this.inboundRateLimiter.allowRequest
+        this.inboundRateLimiter.prune(peerId);
```
dapplion: Rely exclusively on the libp2p disconnect event for pruning. A goodbye is not guaranteed to happen

packages/lodestar/src/network/reqresp/rateTracker.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  requestObjects(objectCount: number): number {
+    if (objectCount <= 0) throw Error("Invalid objectCount " + objectCount);
+    this.prune();
+    if (this.requestsWithinWindow >= this.limit) {
+      return 0;
+    }
+
+    this.requestsWithinWindow += objectCount;
+    const key = Math.floor(Date.now() / 1000);
```
dapplion: Make this 1000 a const BUCKET_SIZE_MS, and justify the value

packages/lodestar/src/network/reqresp/response/rateLimiter.ts
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * The rate tracker for all peers.
+ */
+export class InboundRateLimiter implements IRateLimiter {
+  private readonly logger: ILogger;
+  private readonly peerRpcScores: IPeerRpcScoreStore;
+  private requestCountTotalTracker: RateTracker;
+  private requestCountTrackersByPeer: MapDef<string, RateTracker>;
+  private blockCountTotalTracker: RateTracker;
```
dapplion, line 63: Extensively justify why this limit is necessary for us, it's not usually found in other impl

packages/lodestar/src/network/peers/peerManager.ts
```diff
@@ -492,6 +492,7 @@ export class PeerManager {
 
     this.logger.verbose("peer disconnected", {peer: prettyPrintPeerId(peer), direction, status});
     this.networkEventBus.emit(NetworkEvent.peerDisconnected, peer);
+    this.reqResp.prune(peer);
```
dapplion: If prune() here only prunes rate limit data make the name longer to clarify that it does exactly that and nothing else

packages/lodestar/src/network/options.ts
```diff
@@ -22,5 +23,13 @@ export const defaultNetworkOptions: INetworkOptions = {
...
   localMultiaddrs: ["/ip4/0.0.0.0/tcp/9000"],
   bootMultiaddrs: [],
+  // rate tracker options per 1 minute
+  // per peer per minute, allow to serve up to 5 requests and 500 blocks
+  // total: make 4x peer params
+  requestCountTotalLimit: 200,
+  requestCountPeerLimit: 50,
+  blockCountTotalLimit: 2000,
+  blockCountPeerLimit: 500,
+  rateTrackerTimeoutMs: 60 * 1000,
```
dapplion: If dependency loops allow consider moving to rate limit file so that all the logic and reasoning is in one place

packages/lodestar/src/network/reqresp/response/rateLimiter.ts
```diff
@@ -0,0 +1,99 @@
...
+  rateTrackerTimeoutMs: number;
+};
+
+/**
+ * The rate tracker for all peers.
+ */
+export class InboundRateLimiter implements IRateLimiter {
+  private readonly logger: ILogger;
+  private readonly peerRpcScores: IPeerRpcScoreStore;
+  private requestCountTotalTracker: RateTracker;
```
dapplion: I don't believe this one is necessary since serving non-blocks request _should_ be cheap

packages/lodestar/src/network/reqresp/response/rateLimiter.ts
```diff
@@ -0,0 +1,99 @@
...
+  allowRequest(peerId: PeerId, numBlock?: number): boolean {
+    const peerIdStr = peerId.toB58String();
+
+    // rate limit check for request
+    const requestCountPeerTracker = this.requestCountTrackersByPeer.getOrDefault(peerIdStr);
+    if (requestCountPeerTracker.requestObjects(1) === 0) {
+      this.logger.verbose("Do not serve request due to request count rate limit", {
+        peerId: peerIdStr,
+        requestsWithinWindow: requestCountPeerTracker.getRequestedObjectsWithinWindow(),
+      });
```
dapplion, line 110: Add with metrics to track total rate limits

packages/lodestar/src/network/network.ts
```diff
@@ -68,6 +69,7 @@ export class Network implements INetwork {
     const metadata = new MetadataController({}, {config, chain, logger});
     const peerMetadata = new Libp2pPeerMetadataStore(libp2p.peerStore.metadataBook);
     const peerRpcScores = new PeerRpcScoreStore(peerMetadata);
+    const inboundRateLimiter = new InboundRateLimiter(opts, {...modules, peerRpcScores});
```
dapplion: Instantiate inside ReqResp, since no-one else needs access to it

packages/lodestar/test/unit/network/reqresp/rateTracker.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    rateTracker = new RateTracker({limit: 500, timeoutMs: 60 * 1000});
+    sandbox.useFakeTimers();
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it("should request objects up to limit", () => {
+    expect(rateTracker.getRequestedObjectsWithinWindow()).to.be.equal(0);
```
dapplion, line 28: Add a comment explaining the idea of the test and what is it simulating, for example: "Simulates a peer doing a fast request of 4 objects in X time, and it should get banned after N etc etc."

packages/lodestar/test/unit/network/reqresp/response/rateLimiter.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+    expect(inboundRateLimiter.allowRequest(await PeerId.create())).to.be.false;
+    expect(peerRpcScoresStub.applyAction.calledOnce).to.be.false;
+
+    sandbox.clock.tick(60 * 1000);
+
+    expect(inboundRateLimiter.allowRequest(await PeerId.create())).to.be.true;
+  });
+
+  it("requestCountPeerLimit", async () => {
```
dapplion, line 38: Peer2 gets banned due to total request

packages/lodestar/test/unit/network/reqresp/response/rateLimiter.test.ts
```diff
@@ -0,0 +1,117 @@
...
+      expect(inboundRateLimiter.allowRequest(peerId)).to.be.true;
+    }
+    const peerId2 = await PeerId.create();
+    // it's ok to request blocks for another peer
+    expect(inboundRateLimiter.allowRequest(peerId2)).to.be.true;
+    expect(peerRpcScoresStub.applyAction.calledOnce).to.be.false;
+    // not ok for the same peer id as it reached the limit
+    expect(inboundRateLimiter.allowRequest(peerId)).to.be.false;
+    // this peer id abuses us
+    expect(peerRpcScoresStub.applyAction.calledOnce).to.be.true;
```
dapplion: Add assert message `"peer must be reported"`

packages/cli/src/options/beaconNodeOptions/network.ts
```diff
@@ -104,4 +114,39 @@ export const options: ICliCommandOptions<INetworkArgs> = {
...
+
+  "network.blockCountPeerLimit": {
+    type: "number",
+    description: "Max block count requested per peer per rateTrackerTimeoutMs",
+    defaultDescription: String(defaultOptions.network.blockCountPeerLimit),
+    group: "network",
+  },
+
+  "network.rateTrackerTimeoutMs": {
+    type: "number",
```
dapplion, line 141: `hide: true` to all these new network options

packages/lodestar/src/network/reqresp/rateTracker.ts
wemeetagain: i think this is just turning ms into s ?

twoeths: yeah it's just to convert ms into s and it's not likely we'll change it, we use this 1000 constant in `prune()` method as well so it's good to make a separate constant for it

twoeths:
@dapplion it turns out that some requests come to the rate limiter after libp2p disconnect event so I have to delay (5s) the prune by using a `pruneQueue` in order to cleanly remove unused rate tracker, see https://github.com/ChainSafe/lodestar/pull/3454/commits/5962dc691ae7fbf082e782a30cb1515f98190887

dapplion:
> @dapplion it turns out that some requests come to the rate limiter after libp2p disconnect event so I have to delay (5s) the prune by using a `pruneQueue` in order to cleanly remove unused rate tracker, see [5962dc6](https://github.com/ChainSafe/lodestar/commit/5962dc691ae7fbf082e782a30cb1515f98190887)

This solution looks dangerous assuming 5 seconds isn't safe. What if the node is very overloaded and pormises get stuck for a long time in the event loop?

What do you think about implementing the solution I proposed offline:
- Track latest seen message timestamp per tracker
- Once every some time, i.e. 1 min, 5 min loop over all trackers and clear trackers where now - lastRcvMsg > timeWindow
?

twoeths:
@dapplion so every 10 minutes, I remove tracker if there is no request for a peer in the last 5 minutes, does the number make sense to you https://github.com/ChainSafe/lodestar/pull/3454/commits/2d60bf1f214e8b53c60151869b3035add20a323e ? That's for minority of peers, most of the trackers of peers will be removed by libp2p disconnect event

wemeetagain (approved):
This LGTM but I've got a question before merging.

If we only have an inbound rate limiter, don't we run the risk of disconnecting from ourselves on the network?
Should we also have the outbound rate limiter implemented before merging this?

twoeths:
> If we only have an inbound rate limiter, don't we run the risk of disconnecting from ourselves on the network?
Should we also have the outbound rate limiter implemented before merging this?

@wemeetagain given the threshold of 500 blocks per minute, I think the issue may only happen on a very small devnet, it's safe for prater/pyrmont/mainnet, even if we have the issue we can always tweak through cli params, also we have #3344. What's your thought @dapplion ?

dapplion (approved):

dapplion:
> @wemeetagain given the threshold of 500 blocks per minute, I think the issue may only happen on a very small devnet, it's safe for prater/pyrmont/mainnet, even if we have the issue we can always tweak through cli params, also we have #3344. What's your thought @dapplion ?

I think it's good, we can fix latter if we see issues

PR #3461, opened 2021-11-25
LightClient server no state cache + tracks head

### LINKED ISSUE #3207: LightClientIniter should store checkpoints not finalized checkpoint
**Describe the bug**

LightClientIniter assumes the finalized state is in the cache. That's a big strech and will break in long non-finality

https://github.com/ChainSafe/lodestar/blob/6cf7c2c3f3df6012bd4ce773ab3b5310c347e884/packages/lodestar/src/chain/lightClient/LightClientIniter.ts#L97

Intead, store all checkpoints, or all checkpoints after N slots (i.e. 32) to guard against short lived forks. Then store by checkpoint (epoch | block-root) in reverse for fast pruning and update the init proof route to have the full checkpoint as parameter

dapplion:
**Motivation**

Refactor our current Lightclient library + Lightclient server that supports it to:
1. Track head, instead of just latest finalized checkpoint. Implements https://github.com/ethereum/consensus-specs/pull/2746
2. Lightclient server capable of generating valid updates without access to state caches, only the postState after importing a block

Closes https://github.com/ChainSafe/lodestar/issues/3207

## Routes description

There are three services that a lightclient server can provide. This PR implements all three but each node or backed may choose to support some of them

### Committee sync provider

**getCommitteeUpdates**

Returns an array of best updates in the requested periods within the inclusive range `from` - `to`. Best is defined by (in order of priority): (1) Is finalized update, (2) Has most bits, (3) Oldest update

```
GET eth/v1/lightclient/committee_updates?from&to

returns altair.LightClientUpdate[]
```

**getSnapshot**

Fetch a snapshot with a proof to a trusted block root. The trusted block root should be fetched with similar means to a weak subjectivity checkpoint. Only block roots for checkpoints are guaranteed to be available.

```
GET /eth/v1/lightclient/snapshot/:blockRoot

returns {
  header: phase0.BeaconBlockHeader;
  currentSyncCommittee: altair.SyncCommittee;
  // TODO: Not really necessary since it can be fetched with an update
  nextSyncCommittee: altair.SyncCommittee;
  /** Single branch proof from state root to currentSyncCommittee parent */
  syncCommitteesBranch: Uint8Array[];
}
```

### Head update provider

New SSE event type `lightclient_update`. Emitted at least once per unique header seen signed by a valid sync committee. Provider may choose to send multiple events for the same header if a syncAggregate with more participation is found. Providers should throttle events and only send updates when a syncAggregate is sufficiently better, i.e. at least 16 new participants.

```
emits {
  type: `lightclient_update`,
  message: {
    syncAggregate: altair.SyncAggregate;
    header: phase0.BeaconBlockHeader;
  }
}
```

### State proofs provider

Returns a multiproof of `paths` at the requested `stateId`. The requested `stateId` may not be available. Regular nodes only keep recent states in memory.

```
GET /eth/v1/lightclient/proof/:stateId?paths

returns ArrayBuffer, serialized tree offset multiproof
```


packages/api/src/routes/events.ts
```diff
@@ -1,9 +1,16 @@
...
 import {ContainerType, Json, Type} from "@chainsafe/ssz";
 import {jsonOpts, RouteDef, TypeJson} from "../utils";
 
 // See /packages/api/src/routes/index.ts for reasoning and instructions to add new routes
 
+export type LightclientHeaderUpdate = {
+  syncAggregate: altair.SyncAggregate;
+  header: phase0.BeaconBlockHeader;
+  /** Precomputed root to prevent re-hashing */
+  blockRoot: RootHex;
```
etan-status: Won't the client have to re-hash the header anyway for verification?

packages/api/src/routes/lightclient.ts
```diff
@@ -1,52 +1,54 @@
...
 // See /packages/api/src/routes/index.ts for reasoning and instructions to add new routes
 
+export type LightclientSnapshotWithProof = {
+  header: phase0.BeaconBlockHeader;
+  currentSyncCommittee: altair.SyncCommittee;
+  // TODO: Not really necessary since it can be fetched with an update
+  nextSyncCommittee: altair.SyncCommittee;
+  /** Single branch proof from state root to currentSyncCommittee parent */
+  syncCommitteesBranch: Uint8Array[];
+};
```
etan-status: This no longer includes `genesis_time` and `genesis_validators_root` from the previous `init_proof` endpoint, making it difficult to have the client sync its beacon clock.

packages/api/src/routes/lightclient.ts
```diff
@@ -1,52 +1,54 @@
...
-  getLatestUpdateFinalized(): Promise<{data: altair.LightClientUpdate}>;
-  /** TODO: description */
-  getLatestUpdateNonFinalized(): Promise<{data: altair.LightClientUpdate}>;
   /**
-   * Fetch a proof needed for light client initialization
+   * Returns an array of best updates in the requested periods within the inclusive range `from` - `to`.
+   * Best is defined by (in order of priority):
+   * - Is finalized update
+   * - Has most bits
+   * - Oldest update
```
etan-status, line 27: Shouldn't newer update (i.e., later slot) be preferred over older one, to delay landing in the fallback branch of `process_light_client_update` after initial sync? https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/sync-protocol.md

packages/api/src/routes/events.ts
```diff
@@ -142,6 +152,19 @@ export function getTypeByEvent(): {[K in EventType]: Type<EventData[K]>} {
         epoch: "epoch",
       },
     }),
+
+    [EventType.lightclientUpdate]: new ContainerType<EventData[EventType.lightclientUpdate]>({
```
etan-status: How to obtain the initial latest update in case there is none for a while (missed blocks)? 
Or, is an initial update sent on subscription? And, are finalized / non-finalized updates pushed over the same stream?

packages/api/src/routes/lightclient.ts
```diff
@@ -57,16 +59,13 @@ export function getReqSerializers(): ReqSerializers<Api, ReqTypes> {
...
       writeReq: (from, to) => ({query: {from, to}}),
       parseReq: ({query}) => [query.from, query.to],
       schema: {query: {from: Schema.UintRequired, to: Schema.UintRequired}},
     },
 
-    getLatestUpdateFinalized: reqEmpty,
-    getLatestUpdateNonFinalized: reqEmpty,
-
-    getInitProof: {
+    getSnapshot: {
```
etan-status, line 67: Cool, this new format is way more compact than previously. Furthermore, a client who just wants to sync and is not interested in the state-proof functionality does not require multiproof functionality anymore.

packages/api/src/routes/events.ts
dapplion: Yeah it's a left-over from some testing. I agree on removing it

packages/api/src/routes/lightclient.ts
dapplion: genesis_time and genesis_validators_root should be known ahead of time. For all known networks it should be hardcoded in the lightclient source. For devnets this information can be gathered once by the user through some other means, in a similar way like we distribute config files to join networks.

etan-status: Yeah, but how to validate that the checkpoint is indeed rooted in that particular chain? Or is this assumed implicitly?

dapplion: The user needs to provide a trusted checkpoint, if the block is from another chain the lightclient server won't have any snapshot associated with that block root. Also, updates from a different change won't pass signature validation

packages/api/src/routes/lightclient.ts
dapplion: This implementation of the server diverges from the spec's proposal in that it has two types of updates:
- committee_update: You want to oldest to minimize the chance or re-org
- header_update: You want to newer / latest

packages/api/src/routes/events.ts
dapplion: > How to obtain the initial latest update in case there is none for a while (missed blocks)?

The user can provide any checkpoint. if there's a long range of skipped slots just pick the latest checkpoint available. There's no need for this checkpoint to be in a particular time window, it could be 1 hour old or 1 day old. As long as the blockchain includes 1 single block there will exist at least one snapshot to initialize with.

> Or, is an initial update sent on subscription?

The current flow of the consumer is
1. Get init snapshot, get a sync committee
2. Get sync_committee updates from the period of the initial sync committee to current clock period
3. Subscribe to SSE lightclient_update feed to get header updates

> And, are finalized / non-finalized updates pushed over the same stream?

SSE only pushes header_update, which never includes next_sync_committee. The API REST route for sync committee updates may return a finalized or a non-finalized update depending on what's considered "best" by the server.

twoeths (dismissed):

packages/lodestar/src/chain/lightClient/index.ts
```diff
@@ -1 +1,579 @@
...
+        await this.storeSyncCommittee(postState.nextSyncCommittee, syncCommitteeWitness.nextSyncCommitteeRoot);
+        this.logger.debug("Stored nextSyncCommittee", {period, slot: blockSlot, dependantRoot});
+      }
+    }
+
+    // Ensure referenced syncCommittee are persisted before persiting this one
+    await this.db.syncCommitteeWitness.put(blockRoot, syncCommitteeWitness);
+
+    // Store header in case it is referenced latter by a future finalized checkpoint
+    await this.db.finalizedHeader.put(blockRoot, header);
```
twoeths: @dapplion could you double check this? I suppose we want to store the header of the finalized block instead?

packages/lodestar/src/chain/lightClient/index.ts
```diff
@@ -1 +1,579 @@
...
+  }
+
+  private async persistPostBlockImportData(
+    block: altair.BeaconBlock,
+    postState: CachedBeaconState<allForks.BeaconState>,
+    parentBlock: {blockRoot: RootHex; slot: Slot}
+  ): Promise<void> {
+    const blockSlot = block.slot;
+
+    const header: phase0.BeaconBlockHeader = {
```
twoeths, line 321: we could use `blockToHeader()` util

packages/lodestar/src/chain/lightClient/index.ts
```diff
@@ -1 +1,579 @@
...
+
+    // Store genesis witness in case this block becomes a checkpoint
+    const genesisWitness = getGenesisWitness(postState);
+    await this.db.genesisWitness.put(blockRoot, genesisWitness);
+
+    const syncCommitteeWitness = getSyncCommitteesWitness(postState);
+
+    // Only store current sync committee once per run
+    if (!this.storedCurrentSyncCommittee) {
+      await this.storeSyncCommittee(postState.currentSyncCommittee, syncCommitteeWitness.currentSyncCommitteeRoot);
```
twoeths: is it better to use `Promise.all()` here?

packages/lodestar/src/chain/lightClient/index.ts
dapplion: I've removed beacon-state-transition as dependency to make this lib lighter

packages/lodestar/src/chain/lightClient/index.ts
dapplion: The naming here is the problem, this header is not finalized now but it may in the future. I'll rename to checkpointHeader

packages/api/src/routes/events.ts
```diff
@@ -22,6 +27,8 @@ export enum EventType {
   finalizedCheckpoint = "finalized_checkpoint",
   /** The node has reorganized its chain */
   chainReorg = "chain_reorg",
+  /** New or better header update available */
+  lightclientUpdate = "lightclient_update",
```
wemeetagain: might be good to include the word header in there? Not super sure on this
Eg: "lightclient_header_update" or somehting like that

packages/light-client/src/index.ts
```diff
@@ -0,0 +1,473 @@
...
+      blockRoot: toHexString(ssz.phase0.BeaconBlockHeader.hashTreeRoot(snapshot.header)),
+    };
+  }
+
+  static async initializeFromCheckpoint({
+    config,
+    logger,
+    beaconApiUrl,
+    genesisData,
+    checkpoint,
```
wemeetagain: If we're not using the `checkpoint.epoch` at all (and are really relying on the `checkpoint.block.slot` for timing) we should consider only passing the block root.

packages/light-client/src/utils/logger.ts
```diff
@@ -0,0 +1,24 @@
...
+};
+
+/* eslint-disable no-console */
+
+/**
+ * With `console` module and ignoring debug logs
+ */
+export function getLcLoggerConsole(opts?: {logDebug?: boolean}): ILcLogger {
+  return {
+    error: console.log,
```
wemeetagain: we could use `console.error` and `console.warn` (and `console.info` and `console.debug`)

packages/lodestar/src/node/notifier.ts
```diff
@@ -93,6 +97,15 @@ export async function runNodeNotifier({
       }
       logger.info(nodeState.join(" - "));
 
+      // Log important chain time-based events
+      // Log sync committee change
```
wemeetagain, line 102: :heart:

packages/lodestar/src/chain/lightClient/index.ts
```diff
@@ -1 +1,579 @@
...
+
+interface ILightClientIniterModules {
+  config: IChainForkConfig;
+  db: IBeaconDb;
+  emitter: ChainEventEmitter;
+  logger: ILogger;
+}
+
+/* paths needed to bootstrap the light client */
+export const stateProofPaths = [
```
wemeetagain: remove this dead const and maybe other code related to genesis proofs

wemeetagain (changes requested):
Looks really good!
Just minor comments

packages/light-client/src/index.ts
```diff
@@ -0,0 +1,473 @@
...
+    });
+
+    this.head = {
+      participation: 0,
+      header: snapshot.header,
+      blockRoot: toHexString(ssz.phase0.BeaconBlockHeader.hashTreeRoot(snapshot.header)),
+    };
+  }
+
+  static async initializeFromCheckpoint({
```
wemeetagain: Should the name of this also be changed? `initializeFromBlockRoot`? `initializeFromCheckpointRoot`?

packages/lodestar/src/chain/lightClient/types.ts
```diff
@@ -0,0 +1,78 @@
...
+  isFinalized: false;
+  header: phase0.BeaconBlockHeader;
+  /** Precomputed root to prevent re-hashing */
+  blockRoot: Uint8Array;
+  // Finalized data
+} & altair.SyncAggregate;
+
+export type PartialLightClientUpdate = PartialLightClientUpdateFinalized | PartialLightClientUpdateNonFinalized;
+
+export type InitGenesisProof = {
```
wemeetagain: I think this is a dead type

wemeetagain (changes requested):

wemeetagain (approved):
:tada:

PR #3468, opened 2021-11-30
Use mainnet altair state

dapplion:
**Motivation**

Infura discontinued support for Pyrmont

**Description**

Download states from bucket

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -44,6 +44,13 @@ jobs:
         if: steps.cache-deps.outputs.cache-hit == 'true'
       # </common-build>
 
+      # Restore performance downloaded states
+      - name: Restore performance state cache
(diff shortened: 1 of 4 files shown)


.github/workflows/benchmark.yml
```diff
@@ -44,6 +44,13 @@ jobs:
         if: steps.cache-deps.outputs.cache-hit == 'true'
       # </common-build>
 
+      # Restore performance downloaded states
```
twoeths, line 47: what do we cache this for?

twoeths (approved):

.github/workflows/benchmark.yml
dapplion: The states downloaded from Infura for performance benchmarks of epoch processing, so we don't have to download them every time

PR #3471, opened 2021-11-30
Use Uint8Array in db interface

dapplion:
**Motivation**

In preparation for https://github.com/ChainSafe/lodestar/pull/3461 the db interface is excesively restrictive requiring Buffer when the underlying db lib needs Uint8Arrays

**Description**

- Use Uint8Array in db interface


wemeetagain (approved):
LGTM

PR #3472, opened 2021-12-02
Categorize peer by agent version

twoeths:
**Motivation**

+ Categorize peer by agent version

### DIFF
--- a/packages/lodestar/src/network/peers/client.ts
+++ b/packages/lodestar/src/network/peers/client.ts
@@ -7,55 +7,21 @@ export enum ClientKind {
   Unknown = "Unknown",
 }
 
-export type Client = {
-  kind: ClientKind;
-  version: string;
-  osVersion: string;
-};
-
-export function clientFromAgentVersion(agentVersion: string): Client {
-  const agentSplit = agentVersion.split("/");
-  switch (agentSplit[0]) {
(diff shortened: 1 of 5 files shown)


dapplion (changes requested):
We should only log the known ClientKind enum
```
Dec-02 16:00:53.413 [NETWORK]         debug: Resp sending chunk method=metadata, client=Teku, peer=16...yhBjgf, requestId=146
```

wemeetagain (dismissed):

packages/lodestar/src/network/peers/client.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export type Client = {
+  kind: ClientKind;
+  version: string;
+  osVersion: string;
+};
+
+export function clientFromAgentVersion(agentVersion: string): Client {
+  const agentSplit = agentVersion.split("/");
```
dapplion: split has to manipulate the string more than necessary. If we only care about the ClientKind split the string once with
```ts
const slashIndex = agentVersion.indexOf("/")
const agentKind = slashIndex >= 0 ? agentVersion.slice(0, slashIndex) : null
```
Then return `ClientKind`

dapplion (approved):

PR #3477, opened 2021-12-03
Offset sync period by ALTAIR_FORK_EPOCH

dapplion:
**Motivation**

The first sync period for networks with ALTAIR_FORK_EPOCH >> 0 should be sync period 0.

**Description**

- When the sync period is used as a value for indexing, offset by ALTAIR_FORK_EPOCH
- When the sync period is used purely for comparison, don't offset by ALTAIR_FORK_EPOCH


twoeths (approved):

etan-status:
This PR seems incorrect. ALTAIR does not necessarily start on a sync committee period boundary, making the initial sync committee period shorter than the remaining ones.

Issue #3036, opened 2021-08-30
Road to Stability


q9f:
We currently to nightly "unstable" releases and biweekly "alpha" releases, however, at this early stage of the project, we do not have any strategies towards stabilizing the code and the releases.

I use this meta ticket to outline some strategies to revisit in future in case we want to move into a "beta" or "stable" phase.

Useful resouce: 
- https://hackmd.io/HOaKGHWGT-aLzF6kh1KQfA
- I had this call with the Teku, Lighthouse, and Prysm teams last year
- Everything in this document also applies to Lodestar obviously

The key element is the release checklist (we can already tick two boxes):
> Release components checklist for teams:
> - [x] dedicate a team member to commit to releases (release manager, release coordinator) --> @q9f?
> - [x] commit to a strict versioning scheme (semver, calver, etc.) --> **semver** (lerna)
> - [x] enable concurrent development that reflect different expectations of release quality (stable, beta, nightly, etc.)
> - [x] conduct pre-release testing (release candidates, CI, QA)
> - [x] distribute packages (binaries, signatures, checksums) --> **NPM/Docker**

The hackmd above also outlines two different release strategies (Flow vs. Maturity). I always give the example that Geth uses the _"Flow"_ strategy and Parity ~~uses~~ used _"Maturity."_ I'm personally a big fan of the latter.

Current State of Lodestar https://github.com/ChainSafe/lodestar/releases
- [x] has development branches
    - [x] `master` (pull requests default to `master`)
- [x] has releases (23)
    - [x] `v0.29.1` on :warning: `master`
    - [x] `v0.29.0` on :warning: `master`
    - [x] ...
- [x] :warning: has release branches
    - [x] :warning: no release branch for current track (0.29 ?)

philknows:
- [ ] Create naming conventions for minor/major releases

philknows:
Updated above comment @dapplion for reviewing our process

dapplion:
Spoke with Afri and got a good plan that should be relatively easy to follow tho it requires non-trivial changes to our CI.

This diagram explains the process for the first release going into this flow (v0.33.0 release), and a regular release already being into this flow (v0.34.0 release)

![Screenshot from 2021-11-23 19-03-34](https://user-images.githubusercontent.com/35266934/143079662-038459a5-ac92-4154-800b-44a76411a258.png)

This git flow allows us to:
- Branch a potential release into a "candidate" stage while we test for 3 or more days
[...]

PR #3504, opened 2021-12-10
Trigger release with tag push

dapplion:
**Motivation**

Modifies current release github action to follow https://github.com/ChainSafe/lodestar/issues/3036#issuecomment-976956044

**Description**

- Don't trigger release on package.json bump
- Trigger release on tag push if doesn't include 'beta'
- Get previous tag with new script


wemeetagain (approved):
LGTM, we should open an issue to extend this to also publish beta releases to npm and docker.

PR #3520, opened 2021-12-15
Fix peer score serialization

### LINKED ISSUE #3517: Ban peers with bad scores
**Describe the bug**

this `W5ot96` is a bad peer and we can't dial it for all protocols

the first time I see the "request selection failed" of peer `W5ot96`
```
Dec-12 00:00:01.015 [NETWORK]       ^[[36mverbose^[[39m: Req  error method=ping, encoding=ssz_snappy, client=Unknown, peer=16...W5ot96, requestId=727018 code=REQUEST_ERROR_DIAL_ERROR, error={"message":"protocol selection failed"}
Error: protocol selection failed
```
the last time of the day
```
Dec-12 23:56:43.364 [NETWORK]       ^[[36mverbose^[[39m: Req  error method=status, encoding=ssz_snappy, client=Unknown, peer=16...W5ot96, requestId=1134921 code=REQUEST_ERROR_DIAL_ERROR, error={"message":"protocol selection failed"}
Error: protocol selection failed
```

**Expected behavior**

as we dial this peer quite frequently, we should ban it right after seconds/minutes because `DIAL_ERROR` has lower tolerance so this peer should be banned after 5 errors like this

twoeths:
**Motivation**

+ When there are a lot of `DIAL_ERROR`, I don't see we say goodbye to peers with "Peer score too low" or "Peer banned"
+ We can't dial some peers with `"protocol selection failed"` and it's still in our connected peer without being banned

**Description**

+ Right now we use `ssz.Number64` type to serialize/deserialize peer score which is float, a peer score of `-20` is serialized and deserialized back as a very big uint64

+ Use a new `floatSerdes` which is similar to `stringSerdes` but we parse float as an extra step

+ If method is `ping` and we get `"protocol selection failed"`, ban the peer immediately

Closes #3517

**Tested in contabo-19**
Screen Shot 2021-12-15 at 15 36 01

### DIFF
--- a/packages/lodestar/src/network/reqresp/score.ts
+++ b/packages/lodestar/src/network/reqresp/score.ts
@@ -12,6 +12,14 @@ const libp2pErrorCodes = {
   ERR_UNSUPPORTED_PROTOCOL: "ERR_UNSUPPORTED_PROTOCOL",
 };
 
+/**
+ * Multi stream select error code
+ * https://github.com/multiformats/js-multistream-select/blame/cf4e297b362a43bde2ea117085ceba78cbce1c12/src/select.js#L50
+ */
+const multiStreamSelectErrorCodes = {
+  protocolSelectionFailed: "protocol selection failed",
+};
+
 export function onOutgoingReqRespError(e: Error, method: Method): PeerAction | null {
   if (e instanceof RequestError) {
     switch (e.type.code) {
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/network/reqresp/score.ts
```diff
@@ -12,6 +12,14 @@ const libp2pErrorCodes = {
   ERR_UNSUPPORTED_PROTOCOL: "ERR_UNSUPPORTED_PROTOCOL",
 };
 
+/**
+ * Multi stream select error code
+ * https://github.com/multiformats/js-multistream-select/blame/cf4e297b362a43bde2ea117085ceba78cbce1c12/src/select.js#L50
+ */
+const multiStreamSelectErrorCodes = {
+  protocolSelectionFailed: "protocol selection failed",
```
twoeths, line 20: once this error happens, it'll happen all the time and the peer is useless to us. It's similar to the `ERR_UNSUPPORTED_PROTOCOL` error below

dapplion (approved):
Nice find! :pray:

wemeetagain (approved):

PR #3522, opened 2021-12-15
Implement standard keymanager API

### LINKED ISSUE #3523: Add standard keymanager API

Add the standard key manager API from https://ethereum.github.io/keymanager-APIs/, formerly ethereum/beacon-APIs#151

dapplion:
**Motivation**

Implements the standard key manager API from https://ethereum.github.io/keymanager-APIs/, formerly ethereum/beacon-APIs#151

Closes https://github.com/ChainSafe/lodestar/issues/3523

**Description**

- [x] Add all of the new endpoints from the standard API: GET, POST and DELETE.
- [x] Ensure atomic disable + export.
- [x] Add tests for all the common sequential accesses of the API
- [x] When a key is removed from validator store, delete the key from index service and all the duty services (att, block, sync)
- [x] Add tests for interactions with remote signer validators
- [x] Add end-to-end tests for migration of validators from one VC to another
- [x] Implement the authentication scheme from the standard (token bearer auth)


packages/utils/package.json
```diff
@@ -45,7 +45,9 @@
     "js-yaml": "^3.13.1",
     "winston": "^3.3.3",
     "winston-daily-rotate-file": "^4.5.5",
-    "winston-transport": "^4.3.0"
+    "winston-transport": "^4.3.0",
+    "@types/lockfile": "^1.0.1",
+    "lockfile": "^1.0.4"
```
dapplion: lodestar-utils is a package meant for platform agnostic code, and a lock files requires a filesystem. Could you look into placing this code in a package that is meant only for NodeJS?

dadepo: gotcha! Actually moved it here because I wanted some re-usability :/ But I'll remove it from `lodestar-utils` in a bit. For my own understanding, the packages that are supposed to be node agnostic are:

api, beacon-state-transition, config, fork-choice, params, types, utils and validator.

Will that be accurate?

dapplion: Yes!

packages/keymanager-server/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "build:lib:watch": "yarn run build:lib --watch",
+    "build:typedocs": "typedoc --exclude src/index.ts --out typedocs src",
+    "build:types:watch": "yarn run build:types --watch",
+    "check-types": "tsc",
+    "lint": "eslint --color --ext .ts src/ test/",
+    "lint:fix": "yarn run lint --fix",
+    "pretest": "yarn run check-types",
+    "test:unit": "nyc --cache-dir .nyc_output/.cache -e .ts mocha 'test/unit/**/*.test.ts'",
+    "test": "yarn test:unit",
+    "download-spec-tests": "node -r ts-node/register test/spec/downloadTests.ts",
```
twoeths: delete this?

dadepo: Deleted now.

packages/cli/package.json
```diff
@@ -51,20 +51,22 @@
...
     "@chainsafe/lodestar-config": "^0.35.0",
     "@chainsafe/lodestar-db": "^0.35.0",
+    "@chainsafe/lodestar-keymanager-server": "^0.35.0",
     "@chainsafe/lodestar-params": "^0.35.0",
     "@chainsafe/lodestar-types": "^0.35.0",
     "@chainsafe/lodestar-utils": "^0.35.0",
     "@chainsafe/lodestar-validator": "^0.35.0",
     "@chainsafe/ssz": "^0.8.20",
     "@types/lockfile": "^1.0.1",
+    "lockfile": "^1.0.4",
```
dapplion: Dependencies must be sorted alphabetically, otherwise a new `yarn add` run will cause a diff here

packages/cli/src/cli.ts
```diff
@@ -26,6 +26,7 @@ export function getLodestarCli(): yargs.Argv {
       // As of yargs v16.1.0 dot-notation breaks strictOptions()
       // Manually processing options is typesafe tho more verbose
       "dot-notation": false,
+      "parse-numbers": false,
```
dapplion: What motivated this change? Are you sure this won't cause problems will other arguments that may assume numbers?

packages/cli/src/cmds/dev/handler.ts
```diff
@@ -146,17 +147,43 @@ export async function devHandler(args: IDevArgs & IGlobalArgs): Promise<void> {
...
       logger: logger.child({module: "vali"}),
-      signers: secretKeys.map((secretKey) => ({
-        type: SignerType.Local,
-        secretKey,
-      })),
+      signers,
     });
 
+    // Start keymanager API backend
+    if (args.keymanagerEnabled) {
```
dapplion: This code must not be duplicated with the production run code path

packages/cli/src/cmds/dev/options.ts
```diff
@@ -28,7 +36,6 @@ const devOwnOptions: ICliCommandOptions<IDevOwnArgs> = {
 
   startValidators: {
     description: "Start interop validators in given range",
-    default: "0:7",
```
dapplion: Why removing this default?

packages/cli/src/cmds/validator/handler.ts
```diff
@@ -94,12 +96,37 @@ export async function validatorHandler(args: IValidatorCliArgs & IGlobalArgs): P
...
+        host: args.keymanagerHost,
+        port: args.keymanagerPort,
+        cors: args.keymanagerCors,
+        auth: args.keymanagerAuthEnabled,
+        tokenDir: dbPath,
+      },
+      {config, logger, api: keymanagerApi}
+    );
+    await keymanagerServer.listen();
+    onGracefulShutdownCbs.push(() => keymanagerServer.close());
```
dapplion, line 123: Register close before listen

packages/cli/src/cmds/validator/handler.ts
```diff
@@ -94,12 +96,37 @@ export async function validatorHandler(args: IValidatorCliArgs & IGlobalArgs): P
...
+  }
+
+  onGracefulShutdownCbs.push(() => validator.stop());
```
dapplion: Register stop() before starting the keymanager. onGracefulShutdown() could be called between initializeFromBeaconNode() and starting the keymanager server

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,226 @@
...
+} from "@chainsafe/lodestar-api/keymanager";
+import {fromHexString} from "@chainsafe/ssz";
+import {Interchange, ISlashingProtection, SecretKeyInfo, SignerType, Validator} from "@chainsafe/lodestar-validator";
+import {PubkeyHex} from "@chainsafe/lodestar-validator/src/types";
+import {Root} from "@chainsafe/lodestar-types";
+import {unlink, writeFile} from "fs/promises";
+import lockfile from "lockfile";
+import {ILogger} from "@chainsafe/lodestar-utils";
+
+export const LOCK_FILE_EXT = ".lock";
```
dapplion: Declaring the path of the lock file in two places is extremely dangerous. If we update one and not the other it could cause keystores to appear unlocked when they are not.

packages/cli/src/options/beaconNodeOptions/keymanager.ts
```diff
@@ -0,0 +1,43 @@
+import {ICliCommandOptions} from "../../util";
+import {restApiOptionsDefault} from "@chainsafe/lodestar-keymanager-server";
+
+export type KeymanagerArgs = {
```
dapplion: These are not beaconNodeOptions. BeaconNodeOptions are only the options passed to the lodestar's BeaconNode class. Move this options somewhere else, like the validator command. Then dev can import from there

packages/cli/src/util/lockfile.ts
```diff
@@ -4,14 +4,16 @@ type Lockfile = {
...
 
 let lockFile: Lockfile | null = null;
-
+export const LOCK_FILE_EXT = ".lock";
 /**
  * When lockfile it's required it registers listeners to process
  * Since it's only used by the validator client, require lazily to not pollute
  * beacon_node client context
  */
 export function getLockFile(): Lockfile {
```
dapplion, line 13: See warning on duplicated lock logic

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,226 @@
...
+
+export class KeymanagerApi implements Api {
+  constructor(
+    private readonly logger: ILogger,
+    private readonly validator: Validator,
+    private readonly slashingProtection: ISlashingProtection,
+    private readonly genesisValidatorRoot: Uint8Array | Root,
+    private readonly importKeystoresPath?: string[],
```
dapplion: Why is this an array of paths? Also why is it optional?

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,226 @@
...
+        if (this.validator.validatorStore.hasVotingPubkey(keystore.pubkey)) {
+          statuses[i] = {status: ImportStatus.duplicate};
+          continue;
+        }
+
+        const secretKey = SecretKey.fromBytes(await keystore.decrypt(password));
+
+        this.validator.validatorStore.addSigner({type: SignerType.Local, secretKey});
+
+        const importKeystoresPath = this.importKeystoresPath?.[0];
```
dapplion: This may be undefined creating a broken path below
```
undefined/key_imported_a.json
```

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,226 @@
...
+        const keystorePath = `${importKeystoresPath}/key_imported_${Date.now()}.json`;
```
dapplion: Use NodeJS module path for path manipulation. Otherwise this breaks on windows

dapplion: Why use Date.now() for file name? Why not using the pubkey as hex which is deterministic and allow to know if the same keystore is already there?

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,226 @@
...
+        const lockFilePath = `${keystorePath}${LOCK_FILE_EXT}`;
```
dapplion: Use NodeJS module path for path manipulation

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,226 @@
...
+
+        // Persist keys for latter restarts
+        if (importKeystoresPath) {
```
dapplion: Why is this an optional step?

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,226 @@
...
+          await writeFile(keystorePath, keystoreStr, {encoding: "utf8"});
+          lockfile.lockSync(lockFilePath);
+        }
+
+        this.secretKeysInfo?.push({
```
dapplion: Why is `this.secretKeysInfo` optional?

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,226 @@
...
+    slashingProtection: SlashingProtectionData;
+  }> {
+    const deletedKey: boolean[] = [];
+
+    for (let i = 0; i < pubkeysHex.length; i++) {
+      const pubkeyHex = pubkeysHex[i];
+
+      // Skip unknown keys or remote signers
+      if (
+        !this.validator.validatorStore.getSigner(pubkeyHex) ||
```
dapplion: You can call `this.validator.validatorStore.getSigner(pubkeyHex)` once and recycle the result in the three calls here

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,226 @@
...
+      // Remove key from blockduties
+      // Remove from attestation duties
+      // Remove from Sync committee duties
+      // Remove from indices
+      this.validator.removeSignerFromDutiesAndIndices(pubkeyHex);
+
+      // Remove key from persistent storage
+      if (this.secretKeysInfo) {
+        for (const [key, secretKeyInfo] of this.secretKeysInfo.entries()) {
+          if (secretKeyInfo.secretKey.toPublicKey().toHex() === pubkeyHex) {
```
dapplion: `.toPublicKey().toHex()` this is relatively expensive operation, why not make secretKeysInfo a Map by pubkeyHex? Easier retrieval. Also if you make the keystore path deterministic by pubkey you don't need to keep this data around.

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,226 @@
...
+        const lockFilePath = `${keystorePath}${LOCK_FILE_EXT}`;
+
+        // Persist keys for latter restarts
+        if (importKeystoresPath) {
+          await writeFile(keystorePath, keystoreStr, {encoding: "utf8"});
+          lockfile.lockSync(lockFilePath);
+        }
+
+        this.secretKeysInfo?.push({
+          secretKey,
```
dapplion: Why do you need to keep the secretKey here? Keeping secret keys in multiple places is a security risk, please remove unless it's extremely necessary

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,226 @@
...
+      // Remove from attestation duties
+      // Remove from Sync committee duties
+      // Remove from indices
+      this.validator.removeSignerFromDutiesAndIndices(pubkeyHex);
+
+      // Remove key from persistent storage
+      if (this.secretKeysInfo) {
+        for (const [key, secretKeyInfo] of this.secretKeysInfo.entries()) {
+          if (secretKeyInfo.secretKey.toPublicKey().toHex() === pubkeyHex) {
+            this.secretKeysInfo.splice(key, 1);
```
dapplion: Map.delete()

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,226 @@
...
+            secretKeyInfo?.unlockSecretKeys?.();
+            if (secretKeyInfo?.keystorePath) {
+              try {
+                await unlink(secretKeyInfo?.keystorePath);
+              } catch (e) {
+                this.logger.error("Error deleting keystorePath", {}, e as Error);
```
dapplion: This error should be propagated as a DeletionStatus, why not do a big try catch under `for (let i = 0; i < pubkeysHex.length; i++) {` ?

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,226 @@
...
+
+    const pubkeysBytes = pubkeysHex.map((pubkeyHex) => fromHexString(pubkeyHex));
+
+    const interchangeV5 = await this.slashingProtection.exportInterchange(this.genesisValidatorRoot, pubkeysBytes, {
+      version: "5",
+    });
+
+    // After exporting slashing protection data in bulk, render the status
+    const pubkeysWithSlashingProtectionData = new Set(interchangeV5.data.map((data) => data.pubkey));
+    const statuses: {status: DeletionStatus; message?: string}[] = [];
```
dapplion:
```ts
    const statuses = new Array<{status: DeletionStatus; message?: string}>(pubkeysHex.length)
```

packages/keymanager-server/src/server.ts
```diff
@@ -0,0 +1,163 @@
...
+  host: string;
+  cors: string;
+  port: number;
+  auth: boolean;
+  tokenDir?: string;
+};
+
+export const restApiOptionsDefault: RestApiOptions = {
+  host: "127.0.0.1",
+  port: 9597,
```
dapplion: This port is too close to our current beacon node API port. What are other clients using as the keymanager port? We could adopt their numbering for equivalence

packages/keymanager-server/src/server.ts
```diff
@@ -0,0 +1,163 @@
...
+  private readonly bearerToken: string;
+
+  constructor(optsArg: Partial<RestApiOptions>, modules: IRestApiModules) {
+    // Apply opts defaults
+    const opts = {
+      ...restApiOptionsDefault,
+      ...Object.fromEntries(Object.entries(optsArg).filter(([_, v]) => v != null)),
+    };
+    if (opts.auth) {
+      this.apiTokenPath = `${optsArg.tokenDir}/${apiTokenFileName}`;
```
dapplion: Use NodeJS path module

packages/keymanager-server/src/server.ts
```diff
@@ -0,0 +1,163 @@
...
+import bearerAuthPlugin from "fastify-bearer-auth";
+import querystring from "querystring";
+import {IncomingMessage} from "node:http";
+import {Api} from "@chainsafe/lodestar-api/keymanager";
+import {getRoutes} from "@chainsafe/lodestar-api/keymanager_server";
+import {registerRoutesGroup, RouteConfig} from "@chainsafe/lodestar-api/server";
+import {ErrorAborted, ILogger} from "@chainsafe/lodestar-utils";
+import {IChainForkConfig} from "@chainsafe/lodestar-config";
+import crypto from "node:crypto";
+import {unlink, writeFile} from "fs/promises";
```
dapplion: Prefix `node:fs`. Also please sort imports by
1. Non-chainsafe modules
2. @ chainsafe modules
3. Local paths

We should add this to the contributing guide

packages/keymanager-server/src/server.ts
```diff
@@ -0,0 +1,163 @@
...
+  private readonly logger: ILogger;
+  private readonly activeRequests = new Set<IncomingMessage>();
+  private readonly apiTokenPath: string;
+  private readonly bearerToken: string;
+
+  constructor(optsArg: Partial<RestApiOptions>, modules: IRestApiModules) {
+    // Apply opts defaults
+    const opts = {
+      ...restApiOptionsDefault,
+      ...Object.fromEntries(Object.entries(optsArg).filter(([_, v]) => v != null)),
```
dapplion, line 53: Why is this manipulation necessary? What option can be null and must be discarded?

packages/keymanager-server/src/server.ts
```diff
@@ -0,0 +1,163 @@
...
+    };
+    if (opts.auth) {
+      this.apiTokenPath = `${optsArg.tokenDir}/${apiTokenFileName}`;
+      this.bearerToken = `api-token-${toHexString(crypto.randomBytes(32))}`;
+
+      const initToken = async (): Promise<void> => {
+        await writeFile(this.apiTokenPath, this.bearerToken, {encoding: "utf8"});
+      };
+
+      void initToken();
```
dapplion: If you are going to run a promise unsafely without handling errors, please use the sync fs.writeFileSync api.

packages/keymanager-server/src/server.ts
```diff
@@ -0,0 +1,163 @@
...
+    // Apply opts defaults
+    const opts = {
+      ...restApiOptionsDefault,
+      ...Object.fromEntries(Object.entries(optsArg).filter(([_, v]) => v != null)),
+    };
+    if (opts.auth) {
+      this.apiTokenPath = `${optsArg.tokenDir}/${apiTokenFileName}`;
+      this.bearerToken = `api-token-${toHexString(crypto.randomBytes(32))}`;
+
+      const initToken = async (): Promise<void> => {
```
dapplion: Why pre-declare a function before immediately calling it?

packages/keymanager-server/src/server.ts
```diff
@@ -0,0 +1,163 @@
...
+    });
+
+    if (opts.cors) {
+      void server.register(fastifyCors, {origin: opts.cors});
+    }
+
+    if (opts.auth && this.bearerToken !== "") {
+      void server.register(bearerAuthPlugin, {keys: new Set([this.bearerToken])});
+    }
+    // Log all incoming request to debug (before parsing). TODO: Should we hook latter in the lifecycle? https://www.fastify.io/docs/latest/Lifecycle/
```
dapplion, line 96: Add white space between if and the next comment

packages/keymanager-server/src/server.ts
```diff
@@ -0,0 +1,163 @@
...
+  private readonly apiTokenPath: string;
+  private readonly bearerToken: string;
+
+  constructor(optsArg: Partial<RestApiOptions>, modules: IRestApiModules) {
+    // Apply opts defaults
+    const opts = {
+      ...restApiOptionsDefault,
+      ...Object.fromEntries(Object.entries(optsArg).filter(([_, v]) => v != null)),
+    };
+    if (opts.auth) {
```
dapplion: Add whitespace, between each conceptually different if statement. This should also be part of the contributing guide :pray:

packages/keymanager-server/src/server.ts
```diff
@@ -0,0 +1,163 @@
...
+      } else {
+        void res.status(500).send(err);
+      }
+    });
+
+    if (opts.cors) {
+      void server.register(fastifyCors, {origin: opts.cors});
+    }
+
+    if (opts.auth && this.bearerToken !== "") {
```
dapplion: What happens if `this.bearerToken` is empty? Also the log below would be lying, because

```ts
    if (!opts.auth) {
      this.logger.warn("Keymanager server started without authentication");
    }
```

This won't be logged while there is no authentication. Why no putting this log here in this if statement and its else?

packages/keymanager-server/src/server.ts
```diff
@@ -0,0 +1,163 @@
...
+      this.bearerToken = `api-token-${toHexString(crypto.randomBytes(32))}`;
+
+      const initToken = async (): Promise<void> => {
+        await writeFile(this.apiTokenPath, this.bearerToken, {encoding: "utf8"});
+      };
+
+      void initToken();
+    } else {
+      this.apiTokenPath = "";
+      this.bearerToken = "";
```
dapplion: Don't really like setting this variables to empty strings, it's kinda confusing. Can you find a cleaner approach?

packages/keymanager-server/src/server.ts
```diff
@@ -0,0 +1,163 @@
...
+    // In NodeJS land calling close() only causes new connections to be rejected.
+    // Existing connections can prevent .close() from resolving for potentially forever.
+    // In Lodestar case when the BeaconNode wants to close we will just abruptly terminate
+    // all existing connections for a fast shutdown.
+    // Inspired by https://github.com/gajus/http-terminator/
+    for (const req of this.activeRequests) {
+      req.destroy(Error("Closing"));
+    }
+    if (this.opts.auth && this.apiTokenPath !== "") {
+      await unlink(this.apiTokenPath);
```
dapplion: Why has the api token path be deleted? We need the password to stay the same otherwise users would have to log in again every time the validator client restarts

packages/keymanager-server/src/server.ts
```diff
@@ -0,0 +1,163 @@
...
+    }
+    await this.server.close();
```
dapplion, line 157: White-space between semantically different ifs :pray:

packages/keymanager-server/test/unit/impl.test.ts
```diff
@@ -0,0 +1,138 @@
...
+    slashingProtectionStub = sinon.createStubInstance(SlashingProtection);
+    genesisValidatorRootStub = sinon.createStubInstance(Uint8Array);
+    keystore = sinon.createStubInstance(Keystore);
+  });
+
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  describe("Keymanager / importKeystores", () => {
```
dapplion: Is this test necessary to be fully stubbed? With my experience this type of fully stubbed tests end-up offering very little coverage and are not harder to write that a test against the real code without stubs. Could you try changing to test against real data without any stubs?

packages/lodestar/test/e2e/keymanager/keymanager.test.ts
```diff
@@ -0,0 +1,402 @@
...
+    // 3. IMPORT PUBKEY K1 to SECOND VALIDATOR CLIENT
+
+    // 3.a Confirmation before import, that signing with k1 on vc2 throws
+    expect(async () => {
+      await vc2Info.validator.validatorStore.signAttestation(
+        createAttesterDuty(fromHexString(key1), 0, 0, 1),
+        ssz.phase0.AttestationData.defaultValue(),
+        1
+      );
+    }).to.throw;
```
dapplion: Warning! throw API does not support promises, you have to use chai-as-promised

packages/lodestar/test/e2e/keymanager/keymanager.test.ts
```diff
@@ -0,0 +1,402 @@
...
+
+    // Import k1 to vc2
+    const importResult = await clientKM2.importKeystores(
+      [vc1Info.keystoreContent],
+      ["test123!"],
+      JSON.parse(km1DeleteKeyResult.slashingProtection)
+    );
+
+    // 3.b. COMFIRM IMPORT RESPONSE
+    expect(importResult.data[0].status).to.be.equal("imported");
```
dapplion: Add your comments as asserting messages

```ts
expect(importResult.data[0].status).to.be.equal("imported", "3.b. confirm import response")
```

packages/lodestar/test/e2e/keymanager/keymanager.test.ts
```diff
@@ -0,0 +1,402 @@
...
+    // 4 CONFIRM PRESENCE OF IMPORTED KEY IN SECOND VALIDATOR CLIENT.
+    // 4.a Confirm via api
+    km2ListKeyResult = await clientKM2.listKeys();
+    expect(km2ListKeyResult.data.map((d) => d.validatingPubkey)).to.deep.equal(
+      [key2, key1],
+      "confirm imported keys in vc 2"
+    );
+
+    // 4.b Confirm imported and previous key still exist via file system
+    expect(dirContainFileWithContent(vc2Info.tempDirs.keystoreDir.name, [key1, key2])).to.be.true;
```
dapplion: Assertion messages please :pray: specially when checking it true / false. Just use the comments you already have!
```ts
expect(dirContainFileWithContent(vc2Info.tempDirs.keystoreDir.name, [key1, key2])).to.equal(true, "4.b Confirm imported and previous key still exist via file system")
```

dapplion: You actually have to use this utility because the keystore paths are not determistic! Just use the pubkey in the name and this is much more simple

```ts
expect(fs.readdirSync(dir)).includes(
  keys.map(getKeystorePath),
  "4.b Confirm imported and previous key still exist via file system"
)
```

packages/lodestar/test/e2e/keymanager/keymanager.test.ts
```diff
@@ -0,0 +1,402 @@
...
+
+    // 4.c Confirm vc1 cannot sign with k1
+    expect(async () => {
+      await vc1Info.validator.validatorStore.signAttestation(
+        createAttesterDuty(fromHexString(key1), 0, 0, 1),
+        ssz.phase0.AttestationData.defaultValue(),
+        1
+      );
+    }).to.throw;
```
dapplion: use chai-as-promised

packages/lodestar/test/e2e/keymanager/keymanager.test.ts
```diff
@@ -0,0 +1,402 @@
...
+
+    // 4.d Confirm vc2 can now sign with k1
+    expect(async () => {
+      await vc2Info.validator.validatorStore.signAttestation(
+        createAttesterDuty(fromHexString(key1), 0, 0, 1),
+        ssz.phase0.AttestationData.defaultValue(),
+        1
+      );
+    }).to.not.throw;
```
dapplion: Assertion messagesssss

packages/lodestar/test/e2e/keymanager/keymanager.test.ts
```diff
@@ -0,0 +1,402 @@
...
+
+    // 5. CLEAN UP
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-member-access
+    afterEachCallbacks.push(async () => {
```
dapplion: This clean up is done too late. You must register each step before its start call

packages/lodestar/test/e2e/keymanager/keymanager.test.ts
```diff
@@ -0,0 +1,402 @@
...
+        // prettier-ignore
+        expect((e as Error).message).to.equal("Unauthorized: {\"error\":\"missing authorization header\"}");
+      }
+
+      // Deleting keys is denied
+      try {
+        await client.deleteKeystores([key1]);
+      } catch (e) {
+        // prettier-ignore
+        expect((e as Error).message).to.equal("Unauthorized: {\"error\":\"missing authorization header\"}");
```
dapplion: Use chai-as-promised

packages/lodestar/test/e2e/keymanager/keymanager.test.ts
```diff
@@ -0,0 +1,402 @@
...
+      // importing keys is denied
+      try {
+        await client.importKeystores(["some keystore string"], ["some password"], "some slashing protecting)");
+      } catch (e) {
+        // prettier-ignore
+        expect((e as Error).message).to.equal("Unauthorized: {\"error\":\"missing authorization header\"}");
+      }
+
+      // clean up
+      afterEachCallbacks.push(async () => {
```
dapplion: clean-up too late

packages/lodestar/test/e2e/keymanager/keymanager.test.ts
```diff
@@ -0,0 +1,402 @@
...
+
+      const client = getKeymanagerClient(config, new HttpClient({baseUrl: `http://127.0.0.1:${kmPort}`}));
+
+      expect((await client.listKeys()).data).to.be.deep.equal([
+        {
+          validatingPubkey: `${secretKeys[0].toPublicKey().toHex()}`,
+          derivationPath: "",
+          readonly: true,
+        },
+      ]);
```
dapplion: Assertion messages, clean up too late

packages/lodestar/test/e2e/keymanager/keymanager.test.ts
```diff
@@ -0,0 +1,402 @@
...
+    logger,
+    vc,
+    slashingProtection,
+    vc.genesis.genesisValidatorsRoot,
+    importKeystoresPath,
+    secretKeysInfo
+  );
+
+  return new KeymanagerServer(
+    {host: "127.0.0.1", port, cors: "*", auth: false, tokenDir: "."},
```
dapplion: Use a tokenDir that is a test directory and part of the .gitignore file, otherwise you pollute the repository

packages/lodestar/test/utils/node/validator.ts
```diff
@@ -1,10 +1,234 @@
...
+  SignerLocal,
+} from "@chainsafe/lodestar-validator";
 import {BeaconNode} from "../../../src/node";
 import {testLogger, TestLoggerOpts} from "../logger";
 import {SecretKey} from "@chainsafe/bls";
+import {getLocalSecretKeys} from "@chainsafe/lodestar-cli/src/cmds/validator/keys";
+import {IValidatorCliArgs} from "@chainsafe/lodestar-cli/src/cmds/validator/options";
+import {IGlobalArgs} from "@chainsafe/lodestar-cli/src/options";
+
+function getFirstKeystoreContent(): string {
```
dapplion: Move to these long functions `packages/lodestar/test/utils/node/keymanager.ts`

packages/lodestar/test/utils/node/validator.ts
```diff
@@ -1,10 +1,234 @@
...
+    },
+    description: "",
+    pubkey: "a74e11fd129b9bafc2d6afad4944cd289c238139130a7abafe7b28dde1923a0e4833ad776f9e0d7aaaecd9f0acbfedd3",
+    path: "m/12381/3600/0/0/0",
+    uuid: "5c0169d3-c132-4581-8e7c-afcbf45000cf",
+    version: 4,
+  }).trim();
+}
+
+export async function getAndInitValidatorsWithKeystoreOne({
```
dapplion: Is this function really necessary? You inline the contents where it's consumed

packages/lodestar/test/utils/node/validator.ts
```diff
@@ -1,10 +1,234 @@
...
+  keystoreContent: string;
+  signers: SignerLocal[];
+  slashingProtection: ISlashingProtection;
+  tempDirs: {
+    keystoreDir: DirResult;
+    passwordFile: FileResult;
+  };
+}> {
+  const keystoreDir = tmp.dirSync({unsafeCleanup: true});
+  const keystoreFile = `${keystoreDir.name}/keys_test_${Date.now().toString()}.json`;
```
dapplion: Put in .gitginored directory, use NodeJS path module, use deterministic keystore path

packages/validator/package.json
```diff
@@ -47,6 +47,7 @@
   "dependencies": {
     "@chainsafe/abort-controller": "^3.0.1",
     "@chainsafe/bls": "6.0.3",
+    "@chainsafe/bls-keystore": "^2.0.0",
```
dapplion: Why is this dependency necessary?

packages/validator/src/services/attestationDuties.ts
```diff
@@ -46,6 +47,19 @@ export class AttestationDutiesService {
     chainHeadTracker.runOnNewHead(this.onNewHead);
   }
 
+  remove(signer: PubkeyHex): void {
+    mapValues(Object.fromEntries(this.dutiesByIndexByEpoch), (attDutiesAtEpoch, epoch) => {
```
dapplion: Why do you have to use mapValues here? Just iterate the Map as normal, then you don't need to do parseInt on the keys

packages/validator/src/services/attestationDuties.ts
```diff
@@ -46,6 +47,19 @@ export class AttestationDutiesService {
     chainHeadTracker.runOnNewHead(this.onNewHead);
   }
 
+  remove(signer: PubkeyHex): void {
```
dapplion:
```ts
remove(pubkeyHex: PubkeyHex): void {
```

packages/validator/src/services/blockDuties.ts
```diff
@@ -58,6 +60,14 @@ export class BlockDutiesService {
     return Array.from(publicKeys.values());
   }
 
+  remove(signer: PubkeyHex): void {
+    mapValues(Object.fromEntries(this.proposers), (blockDutyAtEpoch, _epoch) => {
```
dapplion: See above, just iterate. Repeat for all below

packages/validator/src/services/validatorStore.ts
```diff
@@ -72,13 +72,26 @@ export class ValidatorStore {
...
 
     this.slashingProtection = slashingProtection;
     this.genesisValidatorsRoot = genesis.genesisValidatorsRoot;
   }
 
+  addSigner(signer: Signer): void {
+    this.validators.set(getSignerPubkeyHex(signer), signer);
+  }
+
+  getSigner(signer: string | PubkeyHex): Signer | undefined {
```
dapplion: The argument name here is confusing, what is a signer? Are you using a signer to get a signer?

packages/validator/src/validator.ts
```diff
@@ -69,13 +86,39 @@ export class Validator {
...
-    new SyncCommitteeService(config, loggerVc, api, clock, validatorStore, this.chainHeaderTracker, indicesService);
+
+    this.blockDutiesService = new BlockProposingService(
+      config,
+      loggerVc,
+      api,
+      clock,
+      validatorStore,
+      graffiti
+    ).dutiesService;
```
dapplion: I think it would be cleaner for each service to forward the calls, that way we don't need to expose the duties service unnecessarily. i.e. add to BlockProposingService and others
```ts
  removeDutiesForKey(pubkey: PubkeyHex): boolean {
    this.dutiesService.removeDutiesForKey(pubkey)
  }
```

packages/validator/src/services/validatorStore.ts
```diff
@@ -72,13 +72,26 @@ export class ValidatorStore {
...
+    return this.validators.get(signer);
+  }
+
+  removeSigner(signer: string | PubkeyHex): boolean {
```
dapplion: `PubkeyHex` is of type string, why `string | PubkeyHex`?

packages/validator/src/services/validatorStore.ts
```diff
@@ -72,13 +72,26 @@ export class ValidatorStore {
...
+    const pubkeyHex = typeof signer === "string" ? signer : toHexString(signer);
```
dapplion: This condition will always be true `typeof signer === "string"`

packages/validator/src/validator.ts
```diff
@@ -19,12 +19,25 @@ import {toHexString} from "@chainsafe/ssz";
...
+import {SecretKey} from "@chainsafe/bls";
+import {BlockDutiesService} from "./services/blockDuties";
+import {AttestationDutiesService} from "./services/attestationDuties";
+import {SyncCommitteeDutiesService} from "./services/syncCommitteeDuties";
+import {PubkeyHex} from "./types";
+
+export type SecretKeyInfo = {
+  secretKey: SecretKey;
+  keystorePath?: string;
+  unlockSecretKeys?: () => void;
```
dapplion: Does really the validator have to be aware that keystores require an unlock call? The keymanager should handle this on its end. Feels like leaked separation of concerns

packages/validator/src/validator.ts
```diff
@@ -19,12 +19,25 @@ import {toHexString} from "@chainsafe/ssz";
...
+};
 
+// Combined the two new keys
```
dapplion: What does this comment mean?

packages/validator/src/validator.ts
```diff
@@ -107,6 +150,13 @@ export class Validator {
     return new Validator(opts, genesis);
   }
 
+  removeSignerFromDutiesAndIndices(signer: PubkeyHex): void {
```
dapplion:
```ts
removeDutiesForKey(pubkey: PubkeyHex)
```

packages/validator/test/unit/services/attestationDuties.test.ts
```diff
@@ -126,4 +126,67 @@ describe("AttestationDutiesService", function () {
...
+      {
+        4: {duty: duty, selectionProof: null},
+      },
+      "Wrong dutiesService.attesters Map at current epoch"
+    );
+    // then remove
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+    dutiesService.remove(toHexString(pubkeys[0]));
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+    expect(Object.fromEntries(dutiesService["dutiesByIndexByEpoch"]) || new Map()).to.deep.equal(
```
dapplion: Is this `|| new Map()` in the wrong parenthesis place?

packages/validator/test/unit/services/indicesService.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    const pubkey2 = toHexString(pubkeys[secondValidatorIndex]);
+
+    indicesService.index2pubkey.set(firstValidatorIndex, pubkey1);
+    indicesService.index2pubkey.set(secondValidatorIndex, pubkey2);
+
+    indicesService.pubkey2index.set(pubkey1, firstValidatorIndex);
+    indicesService.pubkey2index.set(pubkey2, secondValidatorIndex);
+
+    // remove pubkey2
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
dapplion: Why do you need to suppress `@typescript-eslint/no-unsafe-call`? Isn't this typed? I've seen this lint rule suppression used in many places in this PR, please remove from everywhere it must not be necessary-

packages/cli/src/cmds/validator/handler.ts
dadepo: Done

packages/cli/src/cmds/validator/handler.ts
dadepo: Done

packages/cli/src/options/beaconNodeOptions/keymanager.ts
dadepo: Done. Moved keymanager options out of the beaconNodeOptions

packages/cli/src/cmds/dev/options.ts
dadepo: It was removed while debugging a code path. Added it back now.

packages/cli/src/cli.ts
dadepo: It was added while working with the `--externalSignerPublicKeys` flag and the fact that if the pubkey is provided with "0x" it is parsed as number and hence breaks. I'll create a separate issue to properly look into how to handle that without affecting other arguments.

packages/keymanager-server/src/server.ts
dadepo: Based on reference here https://chainsafe.github.io/lodestar/reference/cli/ the default ports we currently have are: (9000, 9596) beacon related, (8008, and 5000) Metrics related. I can pick a port number in 7000 range here instead if that is fine by you.

wemeetagain: looks like lighthouse uses 5062

https://github.com/sigp/lighthouse/blob/stable/validator_client/src/cli.rs#L184

packages/keymanager-server/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "@chainsafe/bls-keystore": "^2.0.0",
+    "@chainsafe/lodestar-api": "^0.35.0",
+    "@chainsafe/lodestar-config": "^0.35.0",
+    "@chainsafe/lodestar-db": "^0.35.0",
+    "@chainsafe/lodestar-params": "^0.35.0",
+    "@chainsafe/lodestar-types": "^0.35.0",
+    "@chainsafe/lodestar-utils": "^0.35.0",
+    "@chainsafe/lodestar-validator": "^0.35.0",
+    "@chainsafe/ssz": "^0.8.20",
+    "@types/lockfile": "^1.0.1",
```
wemeetagain: I think this can be a dev dependency

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,226 @@
...
+      derivationPath?: string;
+      /** The key associated with this pubkey cannot be deleted from the API */
+      readonly?: boolean;
+    }[];
+  }> {
+    const pubkeys = this.validator.validatorStore.votingPubkeys();
+    return {
+      data: pubkeys.map((pubkey) => ({
+        validatingPubkey: pubkey,
+        derivationPath: "",
```
wemeetagain, line 54: I think this can be undefined since we aren't checking against any stored keystore.

(should this be populated? should we be checking against stored keystores?)

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,226 @@
...
+import {ILogger} from "@chainsafe/lodestar-utils";
+
+export const LOCK_FILE_EXT = ".lock";
+
+export class KeymanagerApi implements Api {
+  constructor(
+    private readonly logger: ILogger,
+    private readonly validator: Validator,
+    private readonly slashingProtection: ISlashingProtection,
+    private readonly genesisValidatorRoot: Uint8Array | Root,
```
wemeetagain: I think these two should be removed here and the functionality exported from `Validator`.
```ts
class Validator {
...
  async importInterchange(interchange: Interchange): Promise<void>
  async exportInterchange(pubkeys: PubkeyHex[], version: InterchangeFormatVersion = {version: "5"}): Promise<Interchange>
```

packages/validator/src/signer/local.ts
```diff
@@ -0,0 +1,57 @@
...
+  addKey(secretKey: SecretKey): void;
+  removeKey(pubkey: BLSPubkey | PubkeyHex): boolean;
+  hasKey(pubkeyHex: PubkeyHex): boolean;
+  hasSomeKeys(): boolean;
+  getKeys(): PubkeyHex[];
+}
+
+/**
+ * Signs messages while hidding secret keys from consumers.
+ * Does not use ES6 classes to implement really private properties, enforced at runtime.
```
wemeetagain: Maybe say: "Uses a closure to implement ..."

packages/validator/src/validator.ts
```diff
@@ -19,12 +19,24 @@ import {toHexString} from "@chainsafe/ssz";
...
+  keystorePath?: string;
+  unlockSecretKeys?: () => void;
+};
 
 export type ValidatorOptions = {
   slashingProtection: ISlashingProtection;
   dbOps: IDatabaseApiOptions;
   api: Api | string;
   signers: Signer[];
+  secretKeysInfo?: SecretKeyInfo[];
```
wemeetagain: I don't think this is used here? remove if possible, so the key manager becomes the sole 'owner' of this.

wemeetagain (changes requested):

packages/keymanager-server/src/server.ts
dapplion: Let's use that one

packages/cli/src/cmds/dev/handler.ts
dadepo: Keymanager support removed from the dev command

packages/keymanager-server/src/impl.ts
dadepo: was initially using the `importKeystoresPath` as provided by the cli argument where it is optional. But now updated to only start the Keymanager if this argument is present hence removed the need to be optional in the Keymanager.

packages/keymanager-server/src/impl.ts
dadepo: Should be resolved now.

packages/keymanager-server/src/impl.ts
dadepo: Switched to using pubkey as part of the filename

packages/keymanager-server/src/impl.ts
dadepo: Switched to using the path module for path construction. This line though is appending a file extension to a string. I believe that should be fine.

packages/keymanager-server/src/impl.ts
dadepo: Updated and this is no longer the case

packages/keymanager-server/src/impl.ts
dadepo: Updated and this is no longer the case

packages/keymanager-server/src/impl.ts
dadepo: Updated and this is no longer the case

packages/keymanager-server/src/impl.ts
dadepo: Updated

packages/keymanager-server/src/impl.ts
dadepo: Switched to using pubkey as part of the filename as suggested

packages/keymanager-server/src/impl.ts
dadepo: Updated as suggested

packages/keymanager-server/src/impl.ts
dadepo: updated

packages/keymanager-server/test/unit/impl.test.ts
dadepo: Switching to real data would probably be still cumbersome. Since I later added code to easily perform e2e testing, I'll remove this test completely to have the main flow covered by the e2e test

packages/lodestar/test/utils/node/validator.ts
dadepo: Done

packages/validator/package.json
dadepo: No longer necessary. Was added while moving some functionalities around. Removed now

packages/validator/test/unit/services/indicesService.test.ts
dadepo: Screenshot 2022-02-22 at 11 24 01

I do get the red lines even in places where the variable is typed, and should not show up. Maybe an issue with IDE integration or somewhere else. Anyways I'll remove the suppress comments

packages/keymanager-server/src/server.ts
dadepo: Imports sorted as suggested

packages/validator/src/validator.ts
dadepo: Already removed. This diff is outdated.

packages/lodestar/test/e2e/keymanager/keymanager.test.ts
dadepo: When used with async/await it works as well, but updated to use chai-as-promised as that seems to be the convention.

packages/lodestar/test/utils/node/validator.ts
dadepo: Not anymore. Removed now

packages/validator/src/services/blockDuties.ts
dadepo: Done

packages/validator/src/services/attestationDuties.ts
dadepo: mapValues removed

packages/validator/src/services/attestationDuties.ts
dadepo: renamed as suggested

packages/validator/src/validator.ts
dadepo: Renamed as suggested

packages/validator/src/validator.ts
dadepo: Updated as suggested

packages/validator/src/signer/local.ts
dadepo: Updated as suggested

packages/keymanager-server/src/impl.ts
dadepo: Implemented suggestion

docs/usage/local.md
```diff
@@ -7,7 +7,7 @@ To quickly test and run Lodestar we recommend starting a local testnet. We recom
 Run a beacon node, with 8 validators with the following command.
 
 ```bash
-./lodestar dev --genesisValidators 8 --genesisTime 1578787200 --enr.ip 127.0.0.1 --rootDir </path/to/node1> --reset
+./lodestar dev --genesisValidators 8 --genesisTime 1578787200 --startValidators 0:1 --enr.ip 127.0.0.1 --rootDir </path/to/node1> --reset
```
dapplion: It would be better to run with 8 validators in the main example. A network with only 1 validator is weird

packages/api/src/index.ts
```diff
@@ -1,5 +1,6 @@
 export * as routes from "./routes";
 export * from "./interface";
 export {getClient, HttpClient, HttpClientOptions, HttpError} from "./client";
+export {getClient as getKeymanagerClient} from "./keymanager/client";
```
dapplion: Don't import in the root index. The above `keymanager.js` file in the root are meant to allow to only import this subset of the API. Note that this initial WIP I did is a bit awkward, I won't vouch for this as the best long term solution

packages/cli/src/cmds/validator/handler.ts
```diff
@@ -11,6 +10,8 @@ import {getValidatorPaths} from "./paths";
 import {IValidatorCliArgs} from "./options";
 import {getLocalSecretKeys, getExternalSigners, groupExternalSignersByUrl} from "./keys";
 import {getVersion} from "../../util/version";
+import {SignerType, Signer, SlashingProtection, Validator} from "@chainsafe/lodestar-validator";
+import {KeymanagerServer, KeymanagerApi} from "@chainsafe/lodestar-keymanager-server";
```
dapplion: Sort imports from `@chainsafe` before local file imports

packages/api/test/unit/keymanager.test.ts
```diff
@@ -0,0 +1,30 @@
...
+import {runGenericServerTest} from "../utils/genericServerTest";
+
+describe("keymanager", () => {
+  runGenericServerTest<Api, ReqTypes>(config, getClient, getRoutes, {
+    listKeys: {
+      args: [],
+      res: {
+        data: [
+          {
+            validatingPubkey: "0x",
```
dapplion: Use proper example data here, a full pubkey of length 48 bytes and a real derivation path

packages/cli/package.json
```diff
@@ -31,7 +31,6 @@
     "lint": "eslint --color --ext .ts src/ test/",
     "lint:fix": "yarn run lint --fix",
     "pretest": "yarn run check-types",
-    "test:unit": "nyc --cache-dir .nyc_output/.cache -e .ts mocha 'test/unit/**/*.test.ts'",
```
dapplion: Why remove? This command should work even with no files

packages/keymanager-server/package.json
dapplion: This must be a devDependency

packages/keymanager-server/src/impl.ts
dapplion: Empty string is not a valid value. If derivationPath is not available, don't include in the JSON response. However the derivation path SHOULD be available since this is dealing with in disk keystores

packages/keymanager-server/src/impl.ts
dapplion: To just concatenate the extension it's okay :+1: For the line above, in my subjective opinion use `path.join`, a function named `join` is too vague and the path namespace helps provide context quicker

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,207 @@
...
+        // Remove from indices
+        this.validator.removeDutiesForKey(pubkeyHex);
+
+        // Remove key from persistent storage
+        for (const keystoreFile of await fs.promises.readdir(this.importKeystoresPath)) {
+          if (keystoreFile === `${KEY_IMPORTED_PREFIX}_${pubkeyHex}.json`) {
+            const keystorePath = join(this.importKeystoresPath, `key_imported_${pubkeyHex}.json`);
+            const lockFilePath = `${keystorePath}${LOCK_FILE_EXT}`;
+            await fs.promises.unlink(lockFilePath);
+            await fs.promises.unlink(keystorePath);
```
dapplion: Remove the lock last

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,207 @@
...
+        // Remove key from blockduties
+        // Remove from attestation duties
+        // Remove from Sync committee duties
+        // Remove from indices
+        this.validator.removeDutiesForKey(pubkeyHex);
+
+        // Remove key from persistent storage
+        for (const keystoreFile of await fs.promises.readdir(this.importKeystoresPath)) {
+          if (keystoreFile === `${KEY_IMPORTED_PREFIX}_${pubkeyHex}.json`) {
+            const keystorePath = join(this.importKeystoresPath, `key_imported_${pubkeyHex}.json`);
```
dapplion: Where are this keystore names defined? You should define helper functions to ensure paths are the same in import / delete. Is this expecting 
```
`${KEY_IMPORTED_PREFIX}_${pubkeyHex}.json`
```
to be equal to 
```
join(this.importKeystoresPath, `key_imported_${pubkeyHex}.json`)
```
?

This logic feels very awkward

packages/keymanager-server/src/impl.ts
```diff
@@ -0,0 +1,207 @@
...
+            const lockFilePath = `${keystorePath}${LOCK_FILE_EXT}`;
```
dapplion: Don't duplicate path manipulation logic

packages/validator/src/services/attestation.ts
```diff
@@ -11,12 +11,13 @@ import {IndicesService} from "./indices";
...
 import {ChainHeaderTracker, HeadEventData} from "./chainHeaderTracker";
 import {ValidatorEvent, ValidatorEventEmitter} from "./emitter";
+import {PubkeyHex} from "../types";
 
 /**
  * Service that sets up and handles validator attester duties.
  */
 export class AttestationService {
-  private readonly dutiesService: AttestationDutiesService;
+  readonly dutiesService: AttestationDutiesService;
```
dapplion: Now this can be private right? (same for all duties services parent class)

packages/validator/src/services/attestationDuties.ts
```diff
@@ -46,6 +47,19 @@ export class AttestationDutiesService {
...
+    this.dutiesByIndexByEpoch.forEach((attDutiesAtEpoch, epoch) => {
+      attDutiesAtEpoch.dutiesByIndex.forEach((attDutyAndProof, vIndex) => {
+        if (toHexString(attDutyAndProof.duty.pubkey) === signer) {
+          attDutiesAtEpoch.dutiesByIndex.delete(vIndex);
+          if (attDutiesAtEpoch.dutiesByIndex.size === 0) {
+            this.dutiesByIndexByEpoch.delete(epoch);
+          }
+        }
+      });
+    });
```
dapplion: While forEach is totally valid syntax, please use `for of` to be consistent with the rest of the validator package style. (same for all duties services)

packages/validator/src/services/attestationDuties.ts
dapplion: Still see `signer` as argument name. Also, might look simpler to name all this functions `removeDutiesForKey` to stay consistent

docs/usage/local.md
dadepo: changes reverted

packages/keymanager-server/src/impl.ts
dadepo: updated as suggested

packages/keymanager-server/src/server.ts
dadepo: Done

packages/validator/src/services/attestationDuties.ts
dadepo: updated

packages/validator/src/services/validatorStore.ts
dadepo: renamed

packages/validator/src/services/attestationDuties.ts
dadepo: Replaced with `for of`

packages/validator/src/services/attestation.ts
dadepo: Now private

packages/keymanager-server/src/impl.ts
dadepo: de-deuplicated

packages/keymanager-server/src/impl.ts
dadepo: Updated. Should be clearer now

packages/keymanager-server/src/impl.ts
dadepo: Done

packages/keymanager-server/src/server.ts
dadepo: since optsArg is a `Partial` type, any of its properties can be undefined. For example the port may not be set, and it would be undefined, which would then override the default port value in `restApiOptionsDefault` to be undefined.

packages/keymanager-server/src/impl.ts
dadepo: Ok so this is the situation: Both the cli package and the keymanager package need the lock file extension. It was originally defined in the cli package (in keys.ts) but I moved it to `packages/cli/src/util/lockfile.ts`, with the hope of reusing it across `cli` package and the `keymanager` when I moved the lockfile.ts into the `lodestar-util` - but I have since learnt that `lodestar-util` should be node agnostic.

But for the lock file extension to not be duplicated, it needs to be defined somewhere it can be used from both `cli` and `keymanager` package. Where do you suggest?

packages/cli/src/util/lockfile.ts
dadepo: See response [here](https://github.com/ChainSafe/lodestar/pull/3522/files#r816413902)

packages/cli/package.json
dadepo: undone

packages/api/test/unit/keymanager.test.ts
dadepo: Done

packages/cli/src/cmds/validator/handler.ts
dadepo: done

jjlitt1 (approved):
Thank u

packages/cli/src/cmds/account/cmds/validator/voluntaryExit.ts
```diff
@@ -1,5 +1,4 @@
-import {SlashingProtection, Validator} from "@chainsafe/lodestar-validator";
-import {SignerType} from "@chainsafe/lodestar-validator";
+import {SignerType, SlashingProtection, Validator} from "@chainsafe/lodestar-validator";
```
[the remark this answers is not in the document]
jjlitt1: Thanks

dapplion:
@dadepo Failing tests and conflicts

packages/keymanager-server/src/impl.ts
dadepo: I read [EIP-2334](https://eips.ethereum.org/EIPS/eip-2334) over the weekend and I realised the changes I made based on this comment where I now set derivationPath to `m/12381/3600/0/0/0` is wrong, because the derivation path should be `m/12381/3600/i/0/0` where i indicates the ith set of validator keys. I do not see a quick way to get the i vallue with the current code, so as to make it possible to merge this PR, I suggest to have the deriviationPath set back to "", and then create a follow up where the proper `derivationPath` can be computed. Acceptable?

dapplion:
@dadepo Failing tests and conflicts

wemeetagain (dismissed):

dapplion:
@dadepo failed build

wemeetagain (approved):
LGTM. Looking for another +1 before merging

g11tech (approved):
LGTM! should we in future save this in a leveldb repo?

dadepo:
>... should we in future save this in a leveldb repo?

Thats an interesting one. So far, what I have observed is: it seem the general practice is to have the keystore files stored on the filesystem, but we can definitely consider/discuss pros/cons of moving it to a leveldb repo

dapplion:
> LGTM! should we in future save this in a leveldb repo?

I don't think so, I doesn't feel right to store a keystore in a level db. Regular JSON files feel more "auditable" and transparent

PR #3532, opened 2021-12-17
Use attestation head state to validate attestations

### LINKED ISSUE #3276: Multiple epoch transitions per epoch

twoeths:
**Motivation**

+ We have multiple epoch transitions per epoch due to the call of `regen.getCheckpointState()` when validating attestation and AggregateAndProof

**Description**

+ The additional epoch transitions come from an out-of-sync node where the headBlockRoot belongs to old epochs

+ The more attestations like that, the more epoch transitions we have. I used to see we have 6-7 epoch transitions in just 2 minutes

+ In all scenarios, using the state at block root yields the same result to the current `getCheckpointState()`, and we don't need any additional epoch transition for it

+ Teku also uses the state at block root for that validation https://github.com/ConsenSys/teku/blob/85af19f4ac8398d62061dbfc77f27281000df6c1/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/AggregateAttestationValidator.java#L103

Closes #3276

**Test Result**

+ Epoch transition shows that there is exactly 1 epoch transition per epoch
Screen Shot 2021-12-17 at 08 53 52

+ This shows "process restart" time
Screen Shot 2021-12-17 at 08 57 23

+ Memory is managed better in a `"subscribeAllSubnets"` node
Screen Shot 2021-12-17 at 08 55 11

+ Number of processed attestation/AggregateAndProof gossip messages are the same
Screen Shot 2021-12-17 at 08 58 52

### DIFF
--- a/packages/lodestar/src/chain/errors/attestationError.ts
+++ b/packages/lodestar/src/chain/errors/attestationError.ts
@@ -118,9 +118,9 @@ export enum AttestationErrorCode {
    */
   COMMITTEE_INDEX_OUT_OF_RANGE = "ATTESTATION_ERROR_COMMITTEE_INDEX_OUT_OF_RANGE",
   /**
-   * Missing attestation pre-state.
+   * Missing attestation head state
    */
-  MISSING_ATTESTATION_TARGET_STATE = "ATTESTATION_ERROR_MISSING_ATTESTATION_TARGET_STATE",
+  MISSING_ATTESTATION_HEAD_STATE = "ATTESTATION_ERROR_MISSING_ATTESTATION_HEAD_STATE",
   /**
    * Invalid aggregator.
    */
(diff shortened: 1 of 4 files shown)


g11tech (approved):

dapplion (approved):
Love the detailed analysis with metrics in the PR body :heart_eyes: thanks for investigating this. @wemeetagain can you review?

wemeetagain (approved):
looks good! thanks for the detailed analysis

Issue #3527, opened 2021-12-16
[PROD] Validator miss attestation


twoeths:
**Describe the bug**

Validator 270069 missed an attestation at slot 2733664. For more information visit: https://beaconcha.in/validator/270069

**Expected behavior**

Validator should not miss any attestations

twoeths:
Validator log shows that the attestation is published on time with correct head

```
Dec-16 04:13:14.649 []                ^[[34mdebug^[[39m: Signed attestation slot=2733664, index=27, head=0xeb5d6fccd8d86989f576e83b22bac72de531abb858a2314e7283a3bce9b3b17f, validatorIndex=270069
```

committeesPerSlot = 131, epoch slot = 0, committee index = 27 => attestation subnet = 27

Node log shows that attestation is published by gossipsub

```
Dec-16 04:13:11.010 [CHAIN]         ^[[36mverbose^[[39m: Clock slot slot=2733664
Dec-16 04:13:14.651 [API]             ^[[34mdebug^[[39m: Req req-6gh 127.0.0.1 POST:/eth/v1/beacon/pool/attestations
Dec-16 04:13:14.770 [NETWORK]       ^[[36mverbose^[[39m: Publish to topic topic=/eth2/afcaaba0/beacon_attestation_27/ssz_snappy
```

maybe it's an issue of having attestation subnet 27 peers at that time to spread the attestation to the network

<img width="995" alt="Screen Shot 2021-12-16 at 11 50 36" src="https://user-images.githubusercontent.com/10568965/146310238-0f08820d-76b3-44fd-8898-fad6a2e05edc.png">

dapplion:
@tuyennhv We should have some internal log of an attestation's journey.

1. Submitted by validator
2. Received in some aggregate
3. Received in block

twoeths:
one reason could be we submit attestation too early https://github.com/sigp/lighthouse/issues/635#issuecomment-784387073 , and the other peer (could be another lodestar node) ignore the attestation since they haven't seen the voted block

dapplion:
> one reason could be we submit attestation too early [sigp/lighthouse#635 (comment)](https://github.com/sigp/lighthouse/issues/635#issuecomment-784387073) , and the other peer (could be another lodestar node) ignore the attestation since they haven't seen the voted block

That seems extremely unlikely, there are multiple aggregators per subnet and Lodestar represents a tiny fraction of the network

twoeths:
[...]

PR #3534, opened 2021-12-17
Log attestation's journey published by local validators

twoeths:
**Motivation**

We want to debug #3527 - missing attestations in mainnet

**Description**

+ Log attestation when it's published to topic and included in AggregateAndProof

+ Don't log when we receive gossip block because I think it's not worth to do so considering the performance to scan through all indices of aggregate and proofs, @dapplion let me know if there's a cheap way to do it. Once an attestation is included in AggregateAndProof, it's likely it's included in upcoming blocks.

part of #3527

### DIFF
--- a/packages/lodestar/src/api/impl/validator/index.ts
+++ b/packages/lodestar/src/api/impl/validator/index.ts
@@ -25,7 +25,6 @@ import {toGraffitiBuffer} from "../../../util/graffiti";
 import {ApiError, NodeIsSyncing} from "../errors";
 import {validateSyncCommitteeGossipContributionAndProof} from "../../../chain/validation/syncCommitteeContributionAndProof";
 import {CommitteeSubscription} from "../../../network/subnets";
-import {OpSource} from "../../../metrics/validatorMonitor";
 import {computeSubnetForCommitteesAtSlot, getPubkeysForIndices} from "./utils";
 import {ApiModules} from "../types";
 import {RegenCaller} from "../../../chain/regen";
@@ -424,7 +423,6 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
     async publishAggregateAndProofs(signedAggregateAndProofs) {
       notWhileSyncing();
 
-      const seenTimestampSec = Date.now() / 1000;
       const errors: Error[] = [];
 
       await Promise.all(
@@ -436,21 +434,13 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
               signedAggregateAndProof
             );
 
-            metrics?.registerAggregatedAttestation(
-              OpSource.api,
-              seenTimestampSec,
-              signedAggregateAndProof,
-              indexedAttestation
+            chain.aggregatedAttestationPool.add(
+              signedAggregateAndProof.message.aggregate,
+              indexedAttestation.attestingIndices,
+              committeeIndices
             );
-
-            await Promise.all([
-              chain.aggregatedAttestationPool.add(
-                signedAggregateAndProof.message.aggregate,
-                indexedAttestation.attestingIndices,
-                committeeIndices
-              ),
-              network.gossip.publishBeaconAggregateAndProof(signedAggregateAndProof),
-            ]);
+            const sentPeers = await network.gossip.publishBeaconAggregateAndProof(signedAggregateAndProof);
+            metrics?.submitAggregatedAttestation(indexedAttestation, sentPeers);
           } catch (e) {
             if (e instanceof AttestationError && e.type.code === AttestationErrorCode.AGGREGATOR_ALREADY_KNOWN) {
               logger.debug("Ignoring known signedAggregateAndProof");
--- a/packages/lodestar/src/api/impl/validator/index.ts
+++ b/packages/lodestar/src/api/impl/validator/index.ts
@@ -423,6 +423,7 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
     async publishAggregateAndProofs(signedAggregateAndProofs) {
       notWhileSyncing();
 
+      const seenTimestampSec = Date.now() / 1000;
       const errors: Error[] = [];
 
       await Promise.all(
@@ -440,7 +441,7 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
               committeeIndices
             );
             const sentPeers = await network.gossip.publishBeaconAggregateAndProof(signedAggregateAndProof);
-            metrics?.submitAggregatedAttestation(indexedAttestation, sentPeers);
+            metrics?.submitAggregatedAttestation(seenTimestampSec, indexedAttestation, sentPeers);
           } catch (e) {
             if (e instanceof AttestationError && e.type.code === AttestationErrorCode.AGGREGATOR_ALREADY_KNOWN) {
               logger.debug("Ignoring known signedAggregateAndProof");
--- a/packages/lodestar/src/metrics/metrics.ts
+++ b/packages/lodestar/src/metrics/metrics.ts
@@ -1,6 +1,7 @@
 /**
  * @module metrics
  */
+import {ILogger} from "@chainsafe/lodestar-utils";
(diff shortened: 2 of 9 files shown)


g11tech (dismissed):

dapplion:
I think this log statements should be part of the validator monitor. And be activated behind a flag since this could be extremely verbose?.

Also, consider using a sub-logger with different module name so it's easier to filter logs with grep.
```ts
logger.child({module: "VMON"})
```

> Don't log when we receive gossip block because I think it's not worth to do so considering the performance to scan through all indices of aggregate and proofs

We already do that in the validator monitor

https://github.com/ChainSafe/lodestar/blob/3d05dd6c2f03bb9ff88d2f37ce6b15d4d77c4a8d/packages/lodestar/src/metrics/validatorMonitor.ts#L305

packages/lodestar/src/metrics/metrics/lodestar.ts
```diff
@@ -530,6 +530,11 @@ export function createLodestarMetrics(
         help: "The delay between when the validator should send the attestation and when it was received",
         labelNames: ["index", "src"],
       }),
+      unaggregatedAttestationMissed: register.gauge<"index" | "subnet">({
```
dapplion: This is not correct since it doesn't mean it was missed. A better name would be
```
unaggregatedAttestationSubmittedNoSubnetPeers
```
but you should clarify if you are counting
- no mesh peers
- no peers in the same topic

packages/lodestar/src/network/peers/peerManager.ts
```diff
@@ -215,6 +215,19 @@ export class PeerManager {
     this.pingAndStatusTimeouts();
   }
 
+  hasBeaconCommitteeSubnetPeer(subnet: number): boolean {
```
dapplion, line 218: This function should look directly into gossipsub data, not the peer's metadata. Peers can lie, so what's the ultimate truth is the gossipsub state. @wemeetagain what do you recommend to look at?

dapplion: This function should look directly into gossipsub data, not the peer's metadata. Peers can lie, so what's the ultimate truth is the gossipsub state. @wemeetagain what do you recommend to look at?

Also whatever code is in hasBeaconCommitteeSubnetPeer() has to be extremely fast to not add more performance load to the node. Current implementation I think is way to heavy

dapplion (changes requested):

packages/lodestar/src/network/peers/peerManager.ts
wemeetagain: You can look at `gossipsub.topics`, which is a `Map<topicString, Set<peerIdString>>`.
This is a map of all our subscribed topics to a set of all connected peers that are also subscribed to the topic.

There's also `gossipsub.mesh`, which is also a  `Map<topicString, Set<peerIdString>>`.
This is a map of all our subscribed topics to a set of all connected peers that are also subscribed to the topic and also in the mesh.

The gossipsub publishing behavior is:  We will try to publish to mesh peers in the topic first. If we have no mesh peers, we get some random peers subscribed in the topic (that have a good enough score), and publish to them.

twoeths: since we have `floodPublish` option as `true` and we're the publisher, we'll publish to all topic peers that meet `publishThreshold` score https://github.com/ChainSafe/js-libp2p-gossipsub/blob/349d5a6303b9f1cd6e73819a495c1d42609bcb3f/ts/index.ts#L1103

the issue is we don't know which topic peers meet that condition unless we duplicate the condition in lodestar. The best way is to return number of `toSend` peers in `libp2p-interface` & `libp2p-gossipsub` in the `publish()/_publish()` method so that we can use it in lodestar metric, is that ok @wemeetagain ?

dapplion: > The best way is to return number of toSend peers in libp2p-interface & libp2p-gossipsub in the publish()/_publish() method so that we can use it in lodestar metric, is that ok @wemeetagain?

Yeah that's a great idea! Feel free to do a PR, will review

twoeths: @dapplion I created https://github.com/libp2p/js-libp2p-interfaces/pull/119

dapplion:
@tuyennhv What's the status of this? Waiting for some changes upstream?

twoeths:
> @tuyennhv What's the status of this? Waiting for some changes upstream?

yeah, we also need to upgrade gossipsub-js, gossipsub-js also needs to work with latest libp2p-interface

packages/lodestar/src/network/peers/peerManager.ts
wemeetagain: We're blocked on updating gossipsub to the latest libp2p-interfaces / libp2p-pubsub by https://github.com/libp2p/js-libp2p-interfaces/pull/127

As an interim solution we could:
- update gossipsub `_publish` to return sent peers, but cast as void
- add a lodestar gossipsub `publish` that duplicates `PubsubBase.publish` but returns the result of `_publish`

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -424,7 +423,6 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
     async publishAggregateAndProofs(signedAggregateAndProofs) {
       notWhileSyncing();
 
-      const seenTimestampSec = Date.now() / 1000;
```
dapplion: By removing the seenTimestamp from here, now the metric is slightly different:
- before: Tracks time attestation is received at the start of handling the REST call
- now: Track the time the attestation is published to gossip. I agree this is the one that matters to us

We should track both, or track the validation time individually. But can be done in a new PR

packages/lodestar/src/metrics/metrics.ts
```diff
@@ -18,14 +19,15 @@ export function createMetrics(
   opts: IMetricsOptions,
   config: IChainForkConfig,
   anchorState: BeaconStateAllForks,
+  logger: ILogger,
```
wemeetagain, line 22: Do we need to be logging the validator monitor / metrics?

twoeths: yeah at the first look it's a little bit strange to add logger to metric, but we need it to log attestation journey since `ValidatorMonitor` is the only place that manage our connected validators

packages/lodestar/src/metrics/validatorMonitor.ts
```diff
@@ -236,12 +239,27 @@ export function createValidatorMonitor(
...
+          metrics.validatorMonitor.prevEpochOnChainAttesterMiss.inc({index});
         }
 
         const balance = balances && balances[index];
         if (balance !== undefined) {
           metrics.validatorMonitor.prevEpochOnChainBalance.set({index}, balance);
         }
+
+        if (failedAttestation) {
+          logger.verbose("Failed attestation in previous epoch", {
```
twoeths: change this log level to debug

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -182,17 +183,17 @@ export class Eth2Gossipsub extends Gossipsub {
     await this.publishObject<GossipType.beacon_block>({type: GossipType.beacon_block, fork}, signedBlock);
   }
 
-  async publishBeaconAggregateAndProof(aggregateAndProof: phase0.SignedAggregateAndProof): Promise<void> {
+  async publishBeaconAggregateAndProof(aggregateAndProof: phase0.SignedAggregateAndProof): Promise<number> {
     const fork = this.config.getForkName(aggregateAndProof.message.aggregate.data.slot);
-    await this.publishObject<GossipType.beacon_aggregate_and_proof>(
+    return await this.publishObject<GossipType.beacon_aggregate_and_proof>(
```
dapplion, line 188: Change in all `publish*` methods for consistency

dapplion (dismissed):
LGMT!

dapplion (approved):

Issue #3438, opened 2021-11-18
Fork choice consensus updates


dapplion:
This issue is a remainder that the Fork choice spec has and will be patched against some attacks. Implementation is not immediate but must be completed prior to the merge.

- [x] https://github.com/ethereum/consensus-specs/pull/2727
- [x] https://github.com/ethereum/consensus-specs/pull/2730

This changes should be optional behind a flag. See https://github.com/sigp/lighthouse/issues/2838 for a great overview of the required changes

dapplion:
Āina-kāri release https://github.com/ethereum/consensus-specs/releases/tag/v1.1.6 is out. We should implement this changes by mid January 2022.

> _From https://github.com/ethereum/consensus-specs/pull/2733_
> Q: CL clients are expected to implement #2730 at any time between now and the merge? Is there any time target or just whenever possible?
> A: I'll put some notes in the release, but ideally, these changes are on merge testnets by roughly end of January and released on the normal release cycle at or before the Merge itself

PR #3540, opened 2021-12-19
spec 1.1.6 - Applying proposer boost to fork-choice

### LINKED ISSUE #3483: Merge spec 1.1.6
main tracker: https://github.com/ChainSafe/lodestar/issues/3410
  - [ ] Cleanup/Devnet tasks
    - [x] #3492
    - [x] is_merge -> is_merge_transition
    - [ ] Genesis execution block handling
  - [x] #3506
    - [x] Remove difficulty from PowBlock (UPDATE: specs 1.1.5 specs on spec-runner fail on this and 1.1.6 spec can be updated once proposer boosting is implemented)
    - [x] Update Forkchoice : 1. proposer boosting 2. atomic update to justified
  - ~~[ ] #3541~~ Independently tracked

### LINKED ISSUE #3506: CL spec 1.1.6: proposer boost
- [x] correct proposer boosting as per 1.1.7 spec

g11tech:
**Motivation**
In order to prevent some attack scenarios, timely receipt of the block needs to be rewarded with proper boost as per CL spec 1.1.6 (the calculation of which was fixed in spec 1.1.7)

This PR implements following CL spec 1.1.6 changes along with spec 1.1.7 calc fixes
- [x] proposer boost
- [x] enables the proposer boost checks in fork-choice spec tests version 1.1.8
- [x] putting the proposer boost behind the flag `--chain.proposerBoostEnabled`

Closes #3506, #3438, 
Partially closes #3483

### DIFF
--- a/packages/cli/src/options/beaconNodeOptions/chain.ts
+++ b/packages/cli/src/options/beaconNodeOptions/chain.ts
@@ -48,7 +48,6 @@ Will double processing times. Use only for debugging purposes.",
   },
 
   "chain.proposerBoostEnabled": {
-    hidden: true,
     type: "boolean",
     description: "Enable proposer boost to reward a timely block",
     defaultDescription: String(defaultOptions.chain.proposerBoostEnabled),
--- a/packages/cli/src/options/beaconNodeOptions/chain.ts
+++ b/packages/cli/src/options/beaconNodeOptions/chain.ts
@@ -5,6 +5,7 @@ export interface IChainArgs {
   "chain.useSingleThreadVerifier": boolean;
   "chain.disableBlsBatchVerify": boolean;
   "chain.persistInvalidSszObjects": boolean;
+  "chain.proposerBoostEnabled": boolean;
   // this is defined as part of IBeaconPaths
   // "chain.persistInvalidSszObjectsDir": string;
 }
@@ -16,6 +17,7 @@ export function parseArgs(args: IChainArgs): IBeaconNodeOptions["chain"] {
     persistInvalidSszObjects: args["chain.persistInvalidSszObjects"],
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any
     persistInvalidSszObjectsDir: undefined as any,
+    proposerBoostEnabled: args["chain.proposerBoostEnabled"],
   };
 }
 
@@ -42,6 +44,14 @@ Will double processing times. Use only for debugging purposes.",
     hidden: true,
     type: "boolean",
     description: "Persist invalid ssz objects or not for debugging purpose",
+    group: "chain",
+  },
+
+  "chain.proposerBoostEnabled": {
+    hidden: true,
+    type: "boolean",
+    description: "Enable proposer boost to reward a timely block",
+    defaultDescription: String(defaultOptions.chain.proposerBoostEnabled),
     group: "chain",
   },
 };
--- a/packages/fork-choice/src/forkChoice/forkChoice.ts
+++ b/packages/fork-choice/src/forkChoice/forkChoice.ts
@@ -1,7 +1,8 @@
 import {isTreeBacked, readonlyValues, toHexString, TreeBacked} from "@chainsafe/ssz";
-import {SAFE_SLOTS_TO_UPDATE_JUSTIFIED, SLOTS_PER_HISTORICAL_ROOT} from "@chainsafe/lodestar-params";
+import {SAFE_SLOTS_TO_UPDATE_JUSTIFIED, SLOTS_PER_HISTORICAL_ROOT, SLOTS_PER_EPOCH} from "@chainsafe/lodestar-params";
 import {Slot, ValidatorIndex, phase0, allForks, ssz, RootHex, Epoch, Root} from "@chainsafe/lodestar-types";
 import {
+  getCurrentInterval,
   computeSlotsSinceEpochStart,
   computeStartSlotAtEpoch,
   computeEpochAtSlot,
@@ -68,6 +69,8 @@ export class ForkChoice implements IForkChoice {
    * Only cache attestation data root hex if it's tree backed since it's available.
    **/
   private validatedAttestationDatas = new Set<string>();
+  /** Boost the entire branch with this proposer root as the leaf */
+  private proposerBoost?: {root: RootHex; score: number} | null;
   /**
    * Instantiates a Fork Choice from some existing components
    *
@@ -146,6 +149,13 @@ export class ForkChoice implements IForkChoice {
   }
 
   /**
+   * Get the proposer boost root
+   */
+  getProposerBoostRoot(): RootHex {
+    return this.proposerBoost?.root ?? HEX_ZERO_HASH;
+  }
+
+  /**
    * Run the fork choice rule to determine the head.
    * Update the head cache.
    *
@@ -162,18 +172,27 @@ export class ForkChoice implements IForkChoice {
 
     let timer;
     this.metrics?.forkChoiceRequests.inc();
-
     try {
+      let deltas: number[];
+
+      // Check if scores need to be calculated/updated
       if (!this.synced) {
         timer = this.metrics?.forkChoiceFindHead.startTimer();
-        const deltas = computeDeltas(
-          this.protoArray.indices,
-          this.votes,
-          this.justifiedBalances,
-          this.justifiedBalances
-        );
+        deltas = computeDeltas(this.protoArray.indices, this.votes, this.justifiedBalances, this.justifiedBalances);
+        /**
+         * The structure in line with deltas to propogate boost up the branch
+         * starting from the proposerIndex
+         */
+        const boosts: number[] = [];
+        if (this.proposerBoost) {
+          const proposerIndex = this.protoArray.indices.get(this.proposerBoost.root);
+          if (proposerIndex == undefined) throw Error("InvalidProposerIndex");
+          boosts[proposerIndex] = this.proposerBoost.score;
+        }
+
         this.protoArray.applyScoreChanges({
           deltas,
+          boosts,
           justifiedEpoch: this.fcStore.justifiedCheckpoint.epoch,
           justifiedRoot: this.fcStore.justifiedCheckpoint.rootHex,
           finalizedEpoch: this.fcStore.finalizedCheckpoint.epoch,
@@ -181,6 +200,7 @@ export class ForkChoice implements IForkChoice {
         });
         this.synced = true;
       }
+
       const headRoot = this.protoArray.findHead(this.fcStore.justifiedCheckpoint.rootHex);
       const headIndex = this.protoArray.indices.get(headRoot);
       if (headIndex === undefined) {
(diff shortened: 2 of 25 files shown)


packages/lodestar/src/chain/blocks/importBlock.ts
```diff
@@ -86,9 +86,15 @@ export async function importBlock(chain: ImportBlockModules, fullyVerifiedBlock:
   // current justified checkpoint should be prev epoch or current epoch if it's just updated
   // it should always have epochBalances there bc it's a checkpoint state, ie got through processEpoch
   const justifiedCheckpoint = postState.currentJustifiedCheckpoint;
-  const onBlockPrecachedData: OnBlockPrecachedData = {executionStatus};
+  const state = getStateForJustifiedBalances(chain, postState, block);
+
+  const onBlockPrecachedData: OnBlockPrecachedData = {
+    executionStatus,
+    justifiedActiveValidators: state.totalActiveValidators,
+    justifiedTotalActiveBalanceByIncrement: state.totalActiveBalanceByIncrement,
```
g11tech: This is an optimization point, could move justifiedActiveValidators, justifiedTotalActiveBalanceByIncrement as private members of forkchoice and update same like justifiedBalances and move back getStateForJustifiedBalances to previous state.

packages/lodestar/src/chain/blocks/importBlock.ts
```diff
@@ -86,9 +86,15 @@ export async function importBlock(chain: ImportBlockModules, fullyVerifiedBlock:
...
+    blockReceptionTime: Math.floor(Date.now() / 1000),
```
twoeths: should this be `Math.floor()` or `Math.ceil()`?

packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -349,15 +370,48 @@ export class ForkChoice implements IForkChoice {
       this.updateJustified(currentJustifiedCheckpoint, justifiedBalances);
     }
 
-    const targetSlot = computeStartSlotAtEpoch(computeEpochAtSlot(slot));
     const blockRoot = this.config.getForkTypes(slot).BeaconBlock.hashTreeRoot(block);
+    const blockRootHex = toHexString(blockRoot);
+
+    if (this.proposerBoostEnabled && slot === this.fcStore.currentSlot) {
+      let secondsIntoSlot;
+      if (preCachedData?.blockReceptionTime !== undefined) {
```
twoeths: why do we have this `blockReceptionTime` as optional? we should always have it

packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -162,25 +173,35 @@ export class ForkChoice implements IForkChoice {
...
-          this.justifiedBalances,
-          this.justifiedBalances
-        );
+        deltas = computeDeltas(this.protoArray.indices, this.votes, this.justifiedBalances, this.justifiedBalances);
+        /**
+         * The structure in line with deltas to propogate boost up the branch
+         * starting from the proposerIndex
+         */
+        const boosts: number[] = [];
+        if (this.proposerBoost) {
```
twoeths: if we call multiple `updateHead()` (with some `onAttestation()`) in the middle, we should not apply boost multiple times, please add this case in a unit test if you can

packages/lodestar/test/spec/allForks/forkChoice.ts
```diff
@@ -208,7 +215,11 @@ function runStateTranstion(
     justifiedBalances = getEffectiveBalances(justifiedState);
   }
   try {
-    forkchoice.onBlock(signedBlock.message, postState, {justifiedBalances});
+    forkchoice.onBlock(signedBlock.message, postState, {
+      justifiedBalances,
+      justifiedActiveValidators,
+      justifiedTotalActiveBalanceByIncrement,
+    });
```
twoeths, line 225: I wonder how does the spec test specify `blockReceptionTime`?

packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -349,15 +370,48 @@ export class ForkChoice implements IForkChoice {
...
+      let secondsIntoSlot;
+      if (preCachedData?.blockReceptionTime !== undefined) {
+        secondsIntoSlot = (preCachedData.blockReceptionTime - state.genesisTime) % this.config.SECONDS_PER_SLOT;
+        // This will only update fork-choice time into slot
+        this.updateTime(this.fcStore.currentSlot, secondsIntoSlot);
+      } else {
+        secondsIntoSlot = this.fcStore.secondsIntoSlot;
+      }
+
+      // Add proposer score boost if the block is timely
```
twoeths: please consider making a separate function if possible as this `onBlock()` function is already long, for example `computeProposerBoostScore()`

packages/fork-choice/src/forkChoice/store.ts
```diff
@@ -22,6 +22,7 @@ export type CheckpointWithHex = phase0.Checkpoint & {rootHex: RootHex};
  */
 export interface IForkChoiceStore {
   currentSlot: Slot;
+  secondsIntoSlot: number;
```
twoeths: I'm not sure if we need this in store, I don't see this in the spec as well. Maybe if we always have `blockReceptionTime`, we don't need this anymore? this does not look like other variables which hold the state of the forkchoice.

packages/lodestar/src/chain/blocks/importBlock.ts
twoeths: nvm, I see `proposerInterval < 1` condition below so 👍

packages/fork-choice/src/forkChoice/forkChoice.ts
g11tech: computeDeltas always use `balanceWeight` of the node (which is the original weight without proposer boost) and  adds boost to calculate the final `weight`, so this will not add `boost` again.

packages/fork-choice/src/forkChoice/store.ts
g11tech: This is actually needed to make the spec tests work as there is a step in it to compare `time`: over here https://github.com/ChainSafe/lodestar/pull/3540/files#diff-49a2a7fb7007eb44eff9a66634a30ff5475e4529132d81b6960675a86b98e1b3R108-R109

packages/lodestar/test/spec/allForks/forkChoice.ts
g11tech: there is a `step` for update time here: https://github.com/ChainSafe/lodestar/pull/3540/files#diff-49a2a7fb7007eb44eff9a66634a30ff5475e4529132d81b6960675a86b98e1b3R57-R58

packages/fork-choice/src/forkChoice/forkChoice.ts
g11tech: reason for this was the way spec tests steps are , update time, block, ... , so if blockReceptionTime is not provided it uses time from the forkchoice secondsIntoSlot, however it can be made mandatory and then i can pull time in the spec test file using getSecondsIntoSlot and pass it as blockReceptionTime in spec

dapplion: Lighthouse implementation looks so simple! Is the complexity here necessary? https://github.com/sigp/lighthouse/blob/a836e180f9ad51f31767c5ffb33bebdeff1a9f3f/consensus/fork_choice/src/fork_choice.rs#L549

packages/fork-choice/src/forkChoice/store.ts
dapplion: Lighthouse doesn't include this data point in the store and can pass spec tests https://github.com/sigp/lighthouse/blob/unstable/consensus/fork_choice/src/fork_choice_store.rs

packages/lodestar/src/chain/blocks/importBlock.ts
dapplion: What is this optimization for specifically? To speed up which function and by how much? Please document and justify as much as possible any changes that add complexity

packages/fork-choice/src/forkChoice/forkChoice.ts
dapplion: Should not be really necessary, the implementation of proposer boost in onBlock is just

```python
    if get_current_slot(store) == block.slot and is_before_attesting_interval:
        store.proposer_boost_root = hash_tree_root(block)
```

https://github.com/ethereum/consensus-specs/pull/2730/files#diff-d4c285a1dd3b0e1fa4b0039ca81142fec96a7e76342b0dc4487883bee593062aR178

packages/fork-choice/src/forkChoice/store.ts
g11tech: because they seem to be matching only slot on the check_time,  which was fine before proposer boost as there were time ticks in tests only on slot boundaries, but now we get time ticks even within the slot, but they are still matching just the slots
https://github.com/sigp/lighthouse/blob/unstable/testing/ef_tests/src/cases/fork_choice.rs#L436:L442

packages/fork-choice/src/forkChoice/forkChoice.ts
g11tech: I have calculated and set score here only, while they calculate the score later: https://github.com/sigp/lighthouse/blob/unstable/consensus/proto_array/src/proto_array.rs#L150 in their `apply_score_changes`.
I can also move score calculation up there.

packages/lodestar/src/chain/blocks/importBlock.ts
g11tech: This is done for computing the score where total active balance, and total total active validators, however could just iterate and count the `justifiedBalances` like how lighthouse does in their compute score fn

g11tech: reverting this optimization and directly calculating these values inside computeProposerBoostScoreFromBalances as the cost is insignificant:
```
✓ computeProposerBoostScoreFromBalances                               6907.508 ops/s    144.7700 us/op        -     206061 runs   30.0 s
```

g11tech:
since lighthouse's proposer score computation is still pretty cheap, reverted  to that approach in order to keep the forkChoice closer to lighthouse's implementation.
```
✓ computeProposerBoostScoreFromBalances                               6907.508 ops/s    144.7700 us/op        -     206061 runs   30.0 s
```

packages/fork-choice/src/forkChoice/forkChoice.ts
g11tech: test added here: https://github.com/ChainSafe/lodestar/pull/3540/files#diff-7153f5b72303272f1458508e23986e3ed13dc46ad1347903db75dca5f8821d7fR86 :+1:

packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -162,25 +176,42 @@ export class ForkChoice implements IForkChoice {
...
+         * The structure in line with deltas to propogate boost up the branch
+         * starting from the proposerIndex
+         */
+        const boosts: number[] = [];
+        if (this.proposerBoostEnabled && this.proposerBoost) {
+          const proposerIndex = this.protoArray.indices.get(this.proposerBoost.root);
+          if (proposerIndex == undefined) throw Error("InvalidProposerIndex");
+          const proposerBoostScore =
+            this.proposerBoost.score ??
+            computeProposerBoostScoreFromBalances(this.justifiedBalances, {
```
twoeths, line 193: this function should be called once per justified balances at most, we can even precompute it in `getEffectiveBalances()` to avoid an extra `for` loop over justified balances, and send both `proposerBoostScore` and `justifiedBalances` to forkchoice over `onBlock()`

packages/fork-choice/src/forkChoice/interface.ts
```diff
@@ -148,6 +148,8 @@ export type PowBlockHex = {
 export type OnBlockPrecachedData = {
   /** `justifiedBalances` balances of justified state which is updated synchronously. */
   justifiedBalances?: number[];
+  /** Time in seconds when the block was received */
+  blockDelay: number;
```
twoeths: name it `blockDelaySec` to make it the same to other places

packages/fork-choice/src/protoArray/protoArray.ts
```diff
@@ -143,6 +147,7 @@ export class ProtoArray {
 
         // back-propagate the nodes delta to its parent
         deltas[parentIndex] += nodeDelta;
+        if (nodeBoost) boosts[parentIndex] = nodeBoost;
```
twoeths: this is also good option to handle boost, however I still prefer lighthouse's implementation https://github.com/sigp/lighthouse/blob/9ed92d6e7830893670e2176a4e3bfd571f344cbf/consensus/proto_array/src/proto_array.rs#L192
+ memory: it does not need another array
+ performance: it does not need another get/set from/to the array

the difference maybe not a lot through. I'd like to know your ideas @dapplion @wemeetagain

packages/fork-choice/src/forkChoice/forkChoice.ts
g11tech: yes, good idea :+1:

packages/fork-choice/src/protoArray/protoArray.ts
dapplion: Agree to go for maximum simplicity here

packages/fork-choice/src/forkChoice/forkChoice.ts
g11tech: actually it could require a bit more bookeeping as there is one place where justified is updated with the bestJustifiedBalances, so will also have to track bestProposerBoostScore

packages/fork-choice/src/protoArray/protoArray.ts
g11tech: :+1:

packages/fork-choice/src/forkChoice/forkChoice.ts
g11tech: @tuyennhv now score is calculated atmost once per justified balances

packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -945,6 +990,13 @@ export class ForkChoice implements IForkChoice {
 
     // Update store time
     this.fcStore.currentSlot = time;
+    if (this.proposerBoostRoot) {
+      // Since previous weight was boosted, we need would now need to recalculate the
+      // scores but without the boost
+      this.proposerBoostRoot = null;
+      this.synced = false;
```
twoeths, line 997: maybe we don't need this with latest approach, adjust the comment accordingly

twoeths: ignore this comment, we have to keep it

twoeths (dismissed):

packages/cli/src/options/beaconNodeOptions/chain.ts
```diff
@@ -44,4 +46,12 @@ Will double processing times. Use only for debugging purposes.",
     description: "Persist invalid ssz objects or not for debugging purpose",
     group: "chain",
   },
+
+  "chain.proposerBoostEnabled": {
+    hidden: true,
```
dapplion: I think we can show this one, it's an important feature that will be switched on by default "soon"

dapplion (approved):
Excellent! Looks great :fire: :fire: I've committed on top to make the feature flag visible

twoeths (approved):

PR #3542, opened 2021-12-20
Time scrape metrics

dapplion:
**Motivation**

Prometheus metrics show that the scrape time of an overloaded node is ~10 sec. However, I did a CPU profile of that exact same node on such a scrape and internally it only takes ~35ms.

Screenshot from 2021-12-20 09-43-16

**Description**

Measure the scrape time of metrics to understand the cost of running metrics without the huge extra time that's introduced somewhere else.

**Next steps**

- [ ] Understand why a 35ms scrape, causes the `curl :/metrics` to take 10 sec.


g11tech (approved):

twoeths (approved):

dapplion:
Tested locally and works as expected :+1:

PR #3545, opened 2021-12-21
Handle errors  thrown by EL's executePayload

### LINKED ISSUE #3537: CL<>EL interop: Peers are getting downscored when EL errors on payload execution
While importing blocks, EL errors with various kind of errors (bad merkle state, internal error etc) as well as possible timeout/connection refused error.

The treatment of all these currently is that that block verification errors and the peer is downscored, which leads to downscoring/banning of peer and soon making lodestar loosing all peers and going out of sync! The lodestar doesn't find peers till the peerstore directory is removed and then the same cycle happens again (based on EL again erroring)

Expected:
The EL executePayload call should be try catched for these errors, and a separate status responded from executionEngine, which should log the error, but lodesstar shouldn;t penalize these peers and also accept this blocks with syncing.

However there is debate in the devnet debug group is that CLs should try discriminate these error (like treat merkle state error as invalid). Waiting till that discussion resolves.

Observed with: lodestar <> nethermind

g11tech:
**Motivation**
Currently in kintsugi devnets it has been observed that EL throws error like `internal error`, `invalid merkles` and sometimes `connection refused`. There could be host of other scenarios where EL is not able to respond. This causes block to be treated as invalid in the current handling, leading to downscoring of the peers. 
In a very limited peer scenario, lodestar eventually kicks all peers and goes out of sync unless peerstore is removed.

This PR 
- tracks the EL operation errors separately on executionEngine level and passes the information upstream. 
- Where these errors are treated in optimistic scenario allowing CL  to move its consensus forward. 
- The user is also logged a `warn` for him to check and take appropriate action/investigation.

and unnecessary/invalid down-scoring of the peer is also prevented.

Closes #3537

### DIFF
--- a/packages/lodestar/src/chain/blocks/verifyBlock.ts
+++ b/packages/lodestar/src/chain/blocks/verifyBlock.ts
@@ -169,7 +169,7 @@ export async function verifyBlockStateTransition(
     }
   }
 
-  let executionStatus = ExecutionStatus.PreMerge;
+  let executionStatus: ExecutionStatus;
   if (executionPayloadEnabled) {
     // TODO: Handle better executePayload() returning error is syncing
     const execResult = await chain.executionEngine.executePayload(
@@ -185,6 +185,7 @@ export async function verifyBlockStateTransition(
         executionStatus = ExecutionStatus.Valid;
         chain.forkChoice.validateLatestHash(execResult.latestValidHash, null);
         break; // OK
+
       case ExecutePayloadStatus.INVALID: {
         // If the parentRoot is not same as latestValidHash, then the branch from latestValidHash
         // to parentRoot needs to be invalidated
@@ -198,6 +199,7 @@ export async function verifyBlockStateTransition(
           errorMessage: execResult.validationError ?? "",
         });
       }
+
       case ExecutePayloadStatus.SYNCING: {
         // It's okay to ignore SYNCING status as EL could switch into syncing
         // 1. On intial startup/restart
@@ -264,6 +266,9 @@ export async function verifyBlockStateTransition(
           errorMessage: execResult.validationError,
         });
     }
+  } else {
+    // isExecutionEnabled() -> false
+    executionStatus = ExecutionStatus.PreMerge;
   }
 
   // Check state root matches
--- a/packages/lodestar/src/chain/blocks/verifyBlock.ts
+++ b/packages/lodestar/src/chain/blocks/verifyBlock.ts
@@ -229,7 +229,7 @@ export async function verifyBlockStateTransition(
           block.message.slot + SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY > clockSlot
         ) {
           throw new BlockError(block, {
-            code: BlockErrorCode.EXECUTION_ENGINE_SYNCING,
+            code: BlockErrorCode.EXECUTION_ENGINE_ERROR,
             errorMessage: `not safe to import SYNCING payload within ${SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY} of currentSlot`,
           });
         }
@@ -255,14 +255,9 @@ export async function verifyBlockStateTransition(
       // back. But for now, lets assume other mechanisms like unknown parent block of a future
       // child block will cause it to replay
       case ExecutePayloadStatus.ELERROR:
-        throw new BlockError(block, {
-          code: BlockErrorCode.EXECUTION_ENGINE_ERRORED,
-          errorMessage: execResult.validationError,
-        });
-
       case ExecutePayloadStatus.UNAVAILABLE:
         throw new BlockError(block, {
-          code: BlockErrorCode.EXECUTION_ENGINE_UNAVAILABLE,
+          code: BlockErrorCode.EXECUTION_ENGINE_ERROR,
           errorMessage: execResult.validationError,
         });
     }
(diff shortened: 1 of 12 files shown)


packages/lodestar/src/executionEngine/http.ts
```diff
@@ -68,24 +68,62 @@ export class ExecutionEngineHttp implements IExecutionEngine {
...
+    const {status, latestValidHash, validationError} = await this.rpc
+      .fetch<EngineApiRpcReturnTypes[typeof method], EngineApiRpcParamTypes[typeof method]>({
+        method,
+        params: [serializedExecutionPayload],
+      })
+      /**
+       * If there are errors by EL like connection refused, internal error, they need to be
+       * treated seperate from being INVALID. For now, just pass the error upstream.
+       */
+      .catch((e: Error) => ({status: ExecutePayloadStatus.ELERROR, latestValidHash: null, validationError: e.message}));
```
dapplion: This catch is too generic, it should be able to differentiate between network errors and legit internal EL errors

g11tech: Do you think they will need to be processed differently in the upstream `verifyBlockStateTransition` function? or we want the differentiation for being detailed/thorough?

packages/lodestar/src/chain/blocks/verifyBlock.ts
```diff
@@ -204,6 +204,21 @@ export async function verifyBlockStateTransition(
           chain.forkChoice.validateLatestHash(execResult.latestValidHash, null);
         }
         break;
+      case ExecutePayloadStatus.ELERROR:
+        // There can be many reasons for which EL failed some of the observed ones are
+        // 1. Connection refused / can't connect to EL port
```
dapplion, line 209: As discussed in EthR&D network errors must not be treated the same as EL being unable to process the block

dapplion: As discussed in EthR&D network errors must not be treated the same as EL being unable to process the block

dapplion (changes requested):

packages/lodestar/src/chain/blocks/verifyBlock.ts
```diff
@@ -199,11 +209,59 @@ export async function verifyBlockStateTransition(
...
+        // conditions are met:
+        //
+        // 1. The justified checkpoint has execution enabled
+        // 2. The current slot (as per the system clock) is at least
+        //    SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY ahead of the slot of the block being
+        //    imported.
+        const justifiedBlock = chain.forkChoice.getJustifiedBlock();
+        const clockSlot = getCurrentSlot(chain.config, postState.genesisTime);
+        if (
+          justifiedBlock.executionStatus === ExecutionStatus.PreMerge ||
```
twoeths: maybe `&&`? what I read is if one of these conditions is not met, it's ok for an optimistic import

packages/lodestar/src/executionEngine/http.ts
```diff
@@ -68,24 +69,77 @@ export class ExecutionEngineHttp implements IExecutionEngine {
...
+      .catch((e: Error) => {
+        let errResponse: EngineApiRpcReturnTypes[typeof method];
+        if (e instanceof TimeoutError) {
+          errResponse = {status: ExecutePayloadStatus.UNAVAILABLE, latestValidHash: null, validationError: e.message};
+        } else if (e instanceof ErrorAborted) {
+          errResponse = {status: ExecutePayloadStatus.UNAVAILABLE, latestValidHash: null, validationError: e.message};
+
+          // What would be better way of handling this
+        } else if (e.message === "Only absolute URLs are supported") {
+          errResponse = {status: ExecutePayloadStatus.UNAVAILABLE, latestValidHash: null, validationError: e.message};
```
twoeths: can we merge these e conditions into 1? `if (condition1 || condition2 || condition3)`

packages/lodestar/src/chain/blocks/verifyBlock.ts
g11tech: ~~condition for safe import is justified block is not execution enabled and block slot is SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY behind clock slot, so this condition is !(Not) of that,
i.e. throw error if justified block is pre merge or blockslot is within SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY from clockSlot, then one can't import the block with SYNCING status (throw error)~~

g11tech: ok I got what you mean, it won't let a syncing node which is PreMerge to import blocks even though merge has happened and is way ahead :+1:

g11tech:
@dapplion reverted the retry mechanism(s). Kindly review.

Would want to tackle the retry mechanism of executionEngine(https://github.com/ChainSafe/lodestar/issues/3567) in separate PR (along with through review of as how jsonRPC client behaves in various unavailability scenarios) as the aim of this PR is to prevent down-scoring peers because of EL errors. 
Local testing (with EL being shutdown) shows that peer count stays good (instead of previous version where it would go to zero)

g11tech:
Been running locally on kintsugi for an hour+ now with EL shutdown, peer count still strong 25.

packages/lodestar/src/sync/range/chain.ts
```diff
@@ -201,6 +201,10 @@ export class SyncChain {
     return toArr(this.batches).map((batch) => batch.getMetadata());
   }
 
+  get startEpochValue(): Epoch {
```
dapplion: Is this function used somewhere?

g11tech: nope :slightly_smiling_face: , thank you for pointing out!, removed.

dapplion (approved):
Looks great! Excellent work

PR #3551, opened 2021-12-23
Add async lag metric

dapplion:
**Motivation**

The event loop lag doesn't tell the whole picture of how busy the event loop is. Some tests show that the event loop lag doesn't really correlate with how slow task processing really is.

The metric added in this PR seems to track better bussiness. However, the results should be similar, since `prom-client` uses `setImmediate()`

https://github.com/siimon/prom-client/blob/d00dbd55736595887ae77f1431e4326296e3c5ec/lib/metrics/eventLoopLag.js#L54

event-loop-final-phases-1400x735

**Description**

Add async lag metric


dapplion:
Not merge-able now, rebase and re-open latter

Issue #3555, opened 2021-12-28
Investigate why there are so few mesh peers per attestation subnet


twoeths:
**Describe the bug**

This happens in both prater and mainnet: we have a lot of topic peers

<img width="1250" alt="Screen Shot 2021-12-28 at 10 48 00" src="https://user-images.githubusercontent.com/10568965/147525988-bd3e0ccc-d271-4c43-9c6b-6717ede07624.png">

but so few mesh peers
<img width="1290" alt="Screen Shot 2021-12-28 at 10 49 48" src="https://user-images.githubusercontent.com/10568965/147526074-a87c7d01-c45b-4d01-8724-1324e06bd975.png">

**Expected behavior**
+ Given `options.D = 8`, we should have up to 8 mesh peers per subnet
+ Track both mesh peers (already exist) and topic peers in the grafana dashboard

twoeths:
when joining a topic, mesh is created with peers that have score >=0, so we should add gossip peer score metric.

twoeths:
It shows that at some points we have so few peers at mesh score threshold (score >=0)

<img width="1537" alt="Screen Shot 2022-01-28 at 14 51 38" src="https://user-images.githubusercontent.com/10568965/151508299-c83f5a7c-2f41-4f89-9fd4-8a9b8271ce99.png">

@wemeetagain what's the consequence of having so few mesh peers?

twoeths:
Adrian at lighthouse mentioned that `Most are positive and hit the max of 100. Rarely see negative scores` at lighthouse, we need to investigate why peer score is so low in lodestar

twoeths:
this could mitigate the issue https://github.com/ChainSafe/lodestar/issues/3705, should assess the situation after that's done

twoeths:
https://github.com/ChainSafe/lodestar/pull/3781 helps us gain good gossip peer score, hence more mesh peers. This could help us mitigate the issue.

Also, we need to care about Job Wait Time. If it's >3s, peers potentially get penalized due to broken IWANT/IHAVE promises, this is from a node of 30 validators

<img width="826" alt="Screen Shot 2022-02-24 at 09 28 57" src="https://user-images.githubusercontent.com/10568965/155446095-99afc722-76bb-4fc8-a199-629e808191f7.png">

twoeths:
There are good number of mesh peers in mainnet nodes - v0.36.0

[...]

PR #3556, opened 2021-12-28
Add gossip peer score metric

twoeths:
**Motivation**

We want to know gossip peer score in order to investigate different gossip issues

**Description**

Part of #3555

### DIFF
--- a/packages/lodestar/src/metrics/metrics/lodestar.ts
+++ b/packages/lodestar/src/metrics/metrics/lodestar.ts
@@ -160,6 +160,17 @@ export function createLodestarMetrics(
       }),
     },
 
+    gossipPeer: {
+      scoreByThreshold: register.gauge<"threshold">({
+        name: "lodestar_gossip_peer_score_by_threshold_count",
+        help: "Gossip peer score by threashold",
+        labelNames: ["threshold"],
+      }),
+      score: register.avgMinMax({
+        name: "lodestar_gossip_score_avg_min_max",
+        help: "Avg min max of all gossip peer scores",
+      }),
+    },
     gossipMesh: {
       peersByType: register.gauge<"type" | "fork">({
         name: "lodestar_gossip_mesh_peers_by_type_count",
--- a/packages/lodestar/src/network/gossip/gossipsub.ts
+++ b/packages/lodestar/src/network/gossip/gossipsub.ts
@@ -456,6 +456,31 @@ export class Eth2Gossipsub extends Gossipsub {
         }
       }
     }
+
+    // track gossip peer score
+    let peerCountScoreGraylist = 0;
+    let peerCountScorePublish = 0;
+    let peerCountScoreGossip = 0;
+    let peerCountScoreMesh = 0;
+    const {graylistThreshold, publishThreshold, gossipThreshold} = gossipScoreThresholds;
+    const {scoreByThreshold, score: scoreMetric} = metrics.gossipPeer;
+    const gossipScores = [];
+
+    for (const peerIdStr of this.peers.keys()) {
+      const score = this.score.score(peerIdStr);
+      if (score >= graylistThreshold) peerCountScoreGraylist++;
+      if (score >= publishThreshold) peerCountScorePublish++;
+      if (score >= gossipThreshold) peerCountScoreGossip++;
+      if (score >= 0) peerCountScoreMesh++;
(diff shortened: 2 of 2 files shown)


packages/lodestar/src/metrics/metrics/lodestar.ts
```diff
@@ -160,6 +160,13 @@ export function createLodestarMetrics(
       }),
     },
 
+    gossipPeer: {
+      score: register.gauge<"peer">({
+        name: "lodestar_gossip_peer_score_total",
+        help: "Gossip peer score",
+        labelNames: ["peer"],
+      }),
+    },
```
dapplion, line 169: The label peer has too much cardinality (too many different values) and will break our metrics. Refer to https://www.robustperception.io/cardinality-is-key as example on rationale.

What we want to know is an idea of the peer's score overall, not per peer specifically. You should use the `AvgMinMax` metric and dump an array of scores on each scrape

dapplion (changes requested):
The label peer has too much cardinality (too many different values) and will break our metrics. Refer to https://www.robustperception.io/cardinality-is-key as example on rationale.

What we want to know is an idea of the peer's score overall, not per peer specifically. You should use the `AvgMinMax` metric and dump an array of scores on each scrape

twoeths:
> The label peer has too much cardinality (too many different values) and will break our metrics. Refer to https://www.robustperception.io/cardinality-is-key as example on rationale.
> 
> What we want to know is an idea of the peer's score overall, not per peer specifically. You should use the `AvgMinMax` metric and dump an array of scores on each scrape

thanks for the guide, in the gossip case I'd like to know number of peers in each threshold:

```ts
export const gossipScoreThresholds: PeerScoreThresholds = {
  meshThreshold: 0,
  gossipThreshold: -4000,
  publishThreshold: -8000,
  graylistThreshold: -16000,
};
```
so I'd create a new Bucket metric for this purpose so that we can reuse later.

@dapplion please let me know if there's a better way to do it

twoeths:
the generated metric looks like this
Screen Shot 2021-12-29 at 11 20 50

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -431,6 +431,10 @@ export class Eth2Gossipsub extends Gossipsub {
         }
       }
     }
+
+    // track gossip peer score
+    const gossipScore = Array.from(this.peers.keys()).map((peer) => this.score.score(peer));
+    metrics.gossipPeer.score.set(gossipScore);
```
dapplion: I think you can achieve the same results with regular Gauges, without a new metric type

```ts
    let peerCountScoreGraylist = 0;
    let peerCountScorePublish = 0;
    let peerCountScoreGossip = 0;
    let peerCountScoreMesh = 0;
    for (const peerIdStr of this.peers.keys()) {
      const score = this.score.score(peerIdStr);
      if (score > gossipScoreThresholds.graylistThreshold) peerCountScoreGraylist++;
      if (score > gossipScoreThresholds.publishThreshold) peerCountScorePublish++;
      if (score > gossipScoreThresholds.gossipThreshold) peerCountScoreGossip++;
      if (score > 0) peerCountScoreMesh++;
    }
    metrics.gossipPeer.scoreByThreshold.set({threshold: "graylist"}, peerCountScoreGraylist);
    metrics.gossipPeer.scoreByThreshold.set({threshold: "publish"}, peerCountScorePublish);
    metrics.gossipPeer.scoreByThreshold.set({threshold: "gossip"}, peerCountScoreGossip);
    metrics.gossipPeer.scoreByThreshold.set({threshold: "mesh"}, peerCountScoreMesh);
``` 

I think your current approach breaks the separation of concerns, the metrics have to know "too much" about gossipsub's internals getting the specific values. 

With my proposal the metrics only need to know that there's a `threshold` label. The Grafana charts would look nicer showing the thresholds by name instead of by numbers.

If you want each threshold gauge to include peers in that range only use `else if` in the code snippet above.

dapplion: We should also track the AvgMinMax of all scores, to answer to different questions:
- `scoreByThreshold` label by threshold: How many peers are in each gossip score threshold?
- `score` AvgMinMax: How well or bad scored are the peers numerically?

dapplion (approved):

PR #3557, opened 2021-12-29
Updated documentation regarding starting a beacon chain

dadepo:
**Motivation**

Some portion of the documentation has since diverged from the current behaviour of the cli. 
This PR updates this outdated part of the documentation.

**Description**

Updated the command listed in the documentation to reflect the fact that `init` is no longer a sub command of `beacon`. Also provided more instructions regarding the location of an eth1 client that the beacon node expects.


dapplion (approved):
Thanks @dadepo and welcome to the repo! :heart:

Issue #3446, opened 2021-11-19
(concatenated string) takes 23% of beacon node's heap


dapplion:
**Describe the bug**

In a synced healthy node in mainnet, according to a recent heap snapshot (concatenated string) take
- 23% of the heap
- 366MB
- 11M instances

![Screenshot from 2021-11-19 22-29-21](https://user-images.githubusercontent.com/35266934/142694423-8659cc4e-5bb9-41b0-b075-8556cbe329e9.png)

**Expected behavior**

??? Strings should take so much memory?

dapplion:
I've manually opened `2021-11-21T10:09:09.713Z.heapsnapshot` with Sublime Text (took 18GB RAM lol) and inspected the "strings" array, which lists all individual strings. Reference to the heapsnapshot format https://github.com/v8/v8/blob/17a99fec258bcc07ea9fc5e4fabcce259751db03/include/v8-profiler.h#L586

Search for `"strings":["<dummy>",`, in the above referenced heapsnapshot it starts at the line number 216_203_711

Notable strange things found:

Base64 text (extreme amounts)
```
m6Iw+9ey1UiRGGI0lCv5DHdbf0YE
mnnGLYdBbZ5B/DfFotXGtfJQEiPQ
mvyxSol88xUg8HPjigsyXFZQPU+s
```
> Seem to be part of `PeerScore.deliveryRecords` and `MessageDeliveries.records`

Short hex values (extreme amounts)
```
0x9983659130
0x8876951e38
0x8ecb05a455
```
> From https://github.com/ChainSafe/lodestar/issues/3446#issuecomment-974910152

Code snippets
```
[...]

PR #3561, opened 2022-01-03
Use memory efficient toHex in pubkey2index map

dapplion:
**Motivation**

From the benchmarks in https://github.com/ChainSafe/lodestar/issues/3446 it's shown that concatenated strings take significant more memory than a string produced with Buffer.toString(). NodeJS's Buffer uses C++ bindings to produce the hex strings thus the different result.

For a network with 250_000 the concatenated strings for the pubkeys alone take 1400 * 250_000 = 350 MB.

**Description**

- Add memory and performance benchmarks for concatenated strings and ES6 Maps with BLS pubkeys
- Use Buffer.toString("hex") to create the keys of pubkey2index Map

This PR should be followed by a general replacement of toHexString by this memory efficient approach. However, this PR does this minimal change first since pubkey2index takes the bulk of memory usage for concatenated strings.


packages/lodestar/test/scripts/blsPubkeyBytesFrequency.ts
```diff
@@ -0,0 +1,128 @@
...
+// bytes 1, collision rate 1
+// bytes 2, collision rate 0.92230224609375
+// bytes 3, collision rate 0.00013267993927001953
+// bytes 4, collision rate 2.0954757928848267e-9
+//
+// (hashed)
+// bytes 1, collision rate 1
+// bytes 2, collision rate 0.92401123046875
+// bytes 3, collision rate 0.00013625621795654297
+// bytes 4, collision rate 3.026798367500305e-9
```
g11tech, line 33: ~~I guess first bytes are extracted and then hashed? (or otherway around?)~~ ok found the answer on L85, its first hashed. But why would byte 1 collision rate be still 1

dapplion: Because there are 250_000 entries so if you only index by the first byte (256 possible values), 100% of those 256 values have collisions.

g11tech (approved):

packages/lodestar/test/memory/testRunnerMemory.ts
```diff
@@ -0,0 +1,215 @@
+export type TestRunnerMemoryOpts<T> = {
```
wemeetagain, line 1: When you're ready, maybe this should be pulled into its own package. I see you're using it in ssz too [here](https://github.com/ChainSafe/ssz/commit/4245cb6c79602e773a17ee7c1c4bad2fec66f35a#diff-9ef6530321876b3689cc72de2d1fa1b506461dc5718761816f3c8ed99ad1cd0dR1)

twoeths (approved):

packages/lodestar/test/memory/testRunnerMemory.ts
dapplion: Agree it should eventually, tho it still needs more testing.

PR #3562, opened 2022-01-03
Add CLI flag to read bootnode ENRs from a file

### LINKED ISSUE #3558: Add cli option to read bootnode ENRs from file
The functionality is there but only triggered for known networks. Expose option to read and parse bootnode ENRs file.

**Spec**

Add flag `--bootnodesFile` in the beacon command, specifically here https://github.com/ChainSafe/lodestar/blob/240085facb5e542defd8d190d5d2a0abbc574ecf/packages/cli/src/cmds/beacon/options.ts#L156

This flag's path must be processed with the same path manipulation logic from `getBeaconPaths()` https://github.com/ChainSafe/lodestar/blob/240085facb5e542defd8d190d5d2a0abbc574ecf/packages/cli/src/cmds/beacon/paths.ts#L29

Parsing logic may be based on https://github.com/ChainSafe/lodestar/blob/240085facb5e542defd8d190d5d2a0abbc574ecf/packages/cli/src/networks/index.ts#L67

and be able to parse any of this file formats
```
### Ethereum Node Records
- enr:-KG4QOWkRj
```
```
### Ethereum Node Records
- "enr:-KG4QOWkRj"
```
```
enr:-KG4QOWkRj
```
```
"enr:-KG4QOWkRj"
```

Parsed bootnode ENRs must be "injected" into the beacon option `network.discv5.bootEnrs` like done here https://github.com/ChainSafe/lodestar/blob/240085facb5e542defd8d190d5d2a0abbc574ecf/packages/cli/src/networks/index.ts#L50

If `bootnodesFile` is specified it MUST overwrite the bootnodes injected by https://github.com/ChainSafe/lodestar/blob/240085facb5e542defd8d190d5d2a0abbc574ecf/packages/cli/src/networks/index.ts#L41

**Tests**

- Test parsing of sample files with at least the four formats outlined above
- Ensure param overwrite rules work as expected and the resulting `beaconNodeOptions` object contains the target bootnode ENRs

Evalir:
**Motivation**

As commented by @dapplion on #3558, the functionality is there but it's only triggered for known networks, and it's a handy feature to test different bootnodes as well.

**Description**
Adds a CLI flag, `--bootnodesFile`, that if added, parses the bootnodes ENRs from a file. Note that if the more specific CLI flag `--network.discv5.bootEnrs` is provided, the bootnodes will **not** be used as the latter flag is considered of higher priority (subject to discussion).

As defined in the spec (#3558):
- Adds `bootnodesFile` flag to `beacon/options`
- Parses the flag with the logic in `beacon/paths`
- On the `BeaconNodeOptions` constructor, parses the file (if possible) and merges the injected bootnodes the rest of the config and _before_ any other CLI arguments, so that if the `--network.discv5.bootEnrs` flag is provided, it overrides the bootnode ENR file.
- Adds a parsing function `parseBootnodesFile` which can parse the formats defined in the spec, plus `.env` style files and JSON files, which might have enrs separated by commas. The parsing function takes advantage of the fact that it can remove any symbols not contained in the [URL and filename safe base64 alphabet](https://datatracker.ietf.org/doc/html/rfc4648#section-5).
- Adds corresponding tests as per the spec.

Closes #3558.

- Running the tests will test injecting the bootnodes, overriding the injected bootnodes with the `--network.discv5.bootEnrs` argument and the parsing function.
- To use the flag, the simplest way is just to run `lodestar beacon --bootnodesFile path/to/bootnodes`.

### DIFF
--- a/packages/cli/src/cmds/beacon/options.ts
+++ b/packages/cli/src/cmds/beacon/options.ts
@@ -157,7 +157,7 @@ export const beaconPathsOptions: ICliCommandOptions<IBeaconPaths> = {
 
   bootnodesFile: {
     hidden: true,
-    description: "Bootnode ENR file path",
+    description: "Bootnodes file path",
     defaultDescription: defaultBeaconPaths.bootnodesFile,
     type: "string",
   },
--- a/packages/cli/src/cmds/beacon/options.ts
+++ b/packages/cli/src/cmds/beacon/options.ts
@@ -158,7 +158,6 @@ export const beaconPathsOptions: ICliCommandOptions<IBeaconPaths> = {
   bootnodesFile: {
     hidden: true,
     description: "Bootnodes file path",
-    defaultDescription: defaultBeaconPaths.bootnodesFile,
     type: "string",
   },
 };
--- a/packages/cli/src/cmds/beacon/options.ts
+++ b/packages/cli/src/cmds/beacon/options.ts
@@ -154,6 +154,13 @@ export const beaconPathsOptions: ICliCommandOptions<IBeaconPaths> = {
     description: "Path to output all logs to a persistent log file",
     type: "string",
   },
+
+  bootnodesFile: {
+    hidden: true,
+    description: "Bootnode ENR file path",
+    defaultDescription: defaultBeaconPaths.bootnodesFile,
+    type: "string",
+  },
 };
 
 export type IBeaconArgs = IBeaconExtraArgs & ILogArgs & IBeaconPaths & IBeaconNodeArgs & IENRArgs;
--- a/packages/cli/src/cmds/init/handler.ts
+++ b/packages/cli/src/cmds/init/handler.ts
@@ -1,16 +1,8 @@
 import fs from "fs";
-import {
-  BeaconNodeOptions,
-  getBeaconConfigFromArgs,
-  initPeerId,
-  initEnr,
-  readPeerId,
-  readEnr,
-  mergeBeaconNodeOptions,
-} from "../../config";
+import {BeaconNodeOptions, getBeaconConfigFromArgs, initPeerId, initEnr, readPeerId, readEnr} from "../../config";
 import {IGlobalArgs, parseBeaconNodeArgs} from "../../options";
 import {mkdir} from "../../util";
-import {fetchBootnodes, getInjectableBootEnrs} from "../../networks";
+import {fetchBootnodes} from "../../networks";
 import {getBeaconPaths} from "../beacon/paths";
 import {IBeaconArgs} from "../beacon/options";
 import {IChainForkConfig} from "@chainsafe/lodestar-config";
@@ -32,16 +24,11 @@ export async function initHandler(args: IBeaconArgs & IGlobalArgs): Promise<Retu
 export async function initializeOptionsAndConfig(args: IBeaconArgs & IGlobalArgs): Promise<ReturnType> {
   const beaconPaths = getBeaconPaths(args);
 
-  const injectableBootEnrs = await getInjectableBootEnrs(beaconPaths.bootnodesFile);
-  // Merge CLI-provided beacon options in order:
-  // - Injected bootnode ENRs if its specified
-  // - Any other options obtained through CLI
-  const beaconNodeOptionsCli = mergeBeaconNodeOptions(injectableBootEnrs, parseBeaconNodeArgs(args));
-
   const beaconNodeOptions = new BeaconNodeOptions({
     network: args.network || "mainnet",
     configFile: beaconPaths.configFile,
-    beaconNodeOptionsCli,
+    bootnodesFile: beaconPaths.bootnodesFile,
+    beaconNodeOptionsCli: parseBeaconNodeArgs(args),
   });
 
   // Auto-setup network
--- a/packages/cli/src/cmds/init/handler.ts
+++ b/packages/cli/src/cmds/init/handler.ts
@@ -44,9 +44,6 @@ export async function initializeOptionsAndConfig(args: IBeaconArgs & IGlobalArgs
     beaconNodeOptionsCli,
   });
 
-  // eslint-disable-next-line no-console
-  console.log(beaconNodeOptions.get()?.network?.discv5?.bootEnrs);
-
   // Auto-setup network
   // Only download files if network.discv5.bootEnrs arg is not specified
   const bOpts = beaconNodeOptions.get();
--- a/packages/cli/src/cmds/init/handler.ts
+++ b/packages/cli/src/cmds/init/handler.ts
@@ -1,8 +1,16 @@
 import fs from "fs";
-import {BeaconNodeOptions, getBeaconConfigFromArgs, initPeerId, initEnr, readPeerId, readEnr} from "../../config";
+import {
+  BeaconNodeOptions,
+  getBeaconConfigFromArgs,
+  initPeerId,
+  initEnr,
+  readPeerId,
+  readEnr,
+  mergeBeaconNodeOptions,
+} from "../../config";
 import {IGlobalArgs, parseBeaconNodeArgs} from "../../options";
 import {mkdir} from "../../util";
-import {fetchBootnodes} from "../../networks";
+import {fetchBootnodes, getInjectableBootEnrs} from "../../networks";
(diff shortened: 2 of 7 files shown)


packages/cli/test/unit/config/beaconNodeOptions.test.ts
```diff
@@ -40,6 +88,110 @@ describe("config / beaconNodeOptions", () => {
...
+
+  for (const {name, input, expected} of testCases) {
+    it(name, () => {
+      expect(parseBootnodesFile(input)).to.be.deep.equal(expected);
+    });
+  }
+});
+
 describe("mergeBeaconNodeOptions", () => {
   const enrsToNetworkConfig = (enrs: string[]): RecursivePartial<IBeaconNodeOptions> => {
```
Evalir: Wanted to point out that I lifted up the `enrsToNetworkConfig` function from this test—but didn't replace it here if we decide to change anything. Lmk if we should just remove it and use the one from `networks/index.ts`!

packages/cli/src/cmds/beacon/options.ts
```diff
@@ -154,6 +154,13 @@ export const beaconPathsOptions: ICliCommandOptions<IBeaconPaths> = {
     description: "Path to output all logs to a persistent log file",
     type: "string",
   },
+
+  bootnodesFile: {
+    hidden: true,
+    description: "Bootnodes file path",
+    defaultDescription: defaultBeaconPaths.bootnodesFile,
```
g11tech: I guess we don't need default path for this as this is optional param

packages/cli/src/networks/index.ts
```diff
@@ -77,6 +78,63 @@ export async function fetchBootnodes(network: NetworkName): Promise<string[]> {
   return enrs;
 }
 
+/**
+ * Reads and parses a list of bootnodes for a network from a file.
+ */
+export async function readBootnodes(bootnodesFilePath: string): Promise<string[]> {
+  if (!fs.existsSync(bootnodesFilePath)) {
```
g11tech: since this shouldn't be default, may be this should throw error if user provided a file that we can't read (in that case readFileSync should directly throw error)

Evalir: Ah yeah, I was wondering about this and if we should just bail out instead of just telling the user "sorry, couldn't do anything". Sure then, we can just let `readFileSync` throw!

Should we also throw if no bootnodes are found in a readable file?

g11tech: >Should we also throw if no bootnodes are found in a readable file?


yes i think that would be a more predictable behavior

packages/cli/src/cmds/beacon/options.ts
Evalir: Done on https://github.com/ChainSafe/lodestar/pull/3562/commits/b8b796e8218b08e6f82e73110ece98ab8e93e20e

packages/cli/src/networks/index.ts
Evalir: Done in https://github.com/ChainSafe/lodestar/pull/3562/commits/71c3af90156f9402ad466aba6ff4693dd72aac99  @g11tech. Happy to iron out this error behavior further so we're 100% happy with it!

packages/cli/src/networks/index.ts
```diff
@@ -77,6 +78,64 @@ export async function fetchBootnodes(network: NetworkName): Promise<string[]> {
   return enrs;
 }
 
+/**
+ * Reads and parses a list of bootnodes for a network from a file.
+ */
+export async function readBootnodes(bootnodesFilePath: string): Promise<string[]> {
```
g11tech: does this function needs to be async as you are using readFileSync?

packages/cli/src/cmds/init/handler.ts
```diff
@@ -23,10 +31,17 @@ export async function initHandler(args: IBeaconArgs & IGlobalArgs): Promise<Retu
 
 export async function initializeOptionsAndConfig(args: IBeaconArgs & IGlobalArgs): Promise<ReturnType> {
   const beaconPaths = getBeaconPaths(args);
+
+  const injectableBootEnrs = await getInjectableBootEnrs(beaconPaths.bootnodesFile);
```
g11tech: since this function doesn;t need to be async, may be the file can be read and merged in the similar fashion as configFile, sort of to give a unified interface

packages/cli/test/unit/config/beaconNodeOptions.test.ts
g11tech: I guess it would be nice :+1:

packages/cli/src/networks/index.ts
Evalir: Whoops, forgot to remove the async flag :).

Done in https://github.com/ChainSafe/lodestar/pull/3562/commits/cd7d0fae3d546c98d0e8c9930f0e7406d73961c1

packages/cli/src/cmds/init/handler.ts
Evalir: Yup, that's true—We could merge it just like `configFile` on the `BeaconNodeOptions` constructor. This would also make the `getInjectableBootEnrs` function quite nicer.

EDIT: Done in https://github.com/ChainSafe/lodestar/pull/3562/commits/cc8a8ae9f83271b5c99e304d5d662e5ddfa3a956. I think it looks nice there :).

packages/cli/test/unit/config/beaconNodeOptions.test.ts
Evalir: Cool, did it at https://github.com/ChainSafe/lodestar/pull/3562/commits/cfdb07209f9eb6ec1eb298533a6c81af3d25c920

g11tech (dismissed):

packages/cli/src/networks/index.ts
```diff
@@ -77,6 +78,62 @@ export async function fetchBootnodes(network: NetworkName): Promise<string[]> {
...
+/**
+ * Parses a file to get a list of bootnodes for a network.
+ * Bootnodes file is expected to contain bootnode ENR's concatenated by newlines, or commas for
+ * parsing plaintext, YAML, JSON and/or env files.
+ */
+export function parseBootnodesFile(bootnodesFile: string): string[] {
+  const enrs = [];
+  for (const line of bootnodesFile.trim().split(/\r?\n/)) {
+    for (const entry of line.split(",")) {
+      if (entry.includes("enr:")) {
```
twoeths: maybe `startsWith()`?

Evalir: ~Ah yes, that'd work well :).~ No sure if that might work well. We could have a totally valid but inline JSON input e.g:

`{ "enrs": ["enr:-deadbeef"], ["enr:-cabbage"] }`

And we wouldn't be able to parse it. We could, of course, detect if something is JSON, YAML, and loop through the properties, but ENR files are pretty specific with their format so it might be better to just look for a `enr:-` entry than to be too clever. Happy to discuss this however :).

packages/cli/src/networks/index.ts
```diff
@@ -77,6 +78,64 @@ export async function fetchBootnodes(network: NetworkName): Promise<string[]> {
...
+ * Reads and parses a list of bootnodes for a network from a file.
+ */
+export function readBootnodes(bootnodesFilePath: string): string[] {
+  const bootnodesFile = fs.readFileSync(bootnodesFilePath, "utf8");
+
+  const bootnodes = parseBootnodesFile(bootnodesFile);
+
+  if (bootnodes.length === 0) {
+    throw new Error(`No bootnodes found on file ${bootnodesFilePath}`);
+  }
```
wemeetagain, line 91: It looks like there's also a check for 0 bootnodes in `enrsToNetworkConfig`

Evalir: Yup, it's not required anymore with the previous changes and this check. Removed in https://github.com/ChainSafe/lodestar/pull/3562/commits/f20a640b0cf7189909d96feff8bab456856fa8d5

wemeetagain (approved):
LGTM :+1:  Thanks @Evalir

dapplion (approved):
Looks great! Thank you so much for the contribution :heart:

g11tech (approved):

PR #3564, opened 2022-01-04
Reprocess unknown block root attestations

### LINKED ISSUE #3560: Reprocess attestations

Right now when an (unaggregated) Attestation or AggregateAndProof are ignored until the voted block reach the node and gets processed

```
Jan-03 02:14:14.116 [NETWORK]         debug: gossip - beacon_attestation - ignore slot=2055371, subnet=17, index=17, code=ATTESTATION_ERROR_UNKNOWN_BEACON_BLOCK_ROOT, root=0x7c8054d7137308f036d977d796a9bff70cea774efb275f7eb3c4b0d7c5600cdb
Jan-03 02:14:14.116 [NETWORK]         debug: gossip - beacon_attestation - ignore slot=2055371, subnet=12, index=12, code=ATTESTATION_ERROR_UNKNOWN_BEACON_BLOCK_ROOT, root=0x7c8054d7137308f036d977d796a9bff70cea774efb275f7eb3c4b0d7c5600cdb
Jan-03 02:14:14.117 [NETWORK]         debug: gossip - beacon_attestation - ignore slot=2055371, subnet=36, index=36, code=ATTESTATION_ERROR_UNKNOWN_BEACON_BLOCK_ROOT, root=0x7c8054d7137308f036d977d796a9bff70cea774efb275f7eb3c4b0d7c5600cdb
Jan-03 02:14:14.663 [CHAIN]         verbose: Block processed slot=2055371, root=0x7c8054d7137308f036d977d796a9bff70cea774efb275f7eb3c4b0d7c5600cdb
```

+ We should be able to cache and reprocess them.
+ Make sure we don't have performance issue with this
+ Make sure the cache is properly bounded

The best way to test is to confirm this error code `ATTESTATION_ERROR_UNKNOWN_BEACON_BLOCK_ROOT` rarely happens

twoeths:
**Motivation**

Right now when attestations come before the block, we ignore attestations. We should follow other clients to queue and reprocess those attestations when the block becomes known. This:
+ May help us with the missing attestation issue #3527, right now we swallow unknown block root attestations
+ Potentially increase gossip peer score after we propagating those attestations

**Description**
Since we used to have performance issue with this `reprocess` work, the approach is quite conservative

+ New `ReprocessController` to wait for the block, add awaiting promises to `awaitingPromisesByRootBySlot`
+ Prune `awaitingPromisesByRootBySlot` per clock slot since we don't want to reprocess old slot attestations
+ If block comes right close to next slot, do not reprocess as the node may be struggling to resync, see `REPROCESS_MIN_TIME_TO_NEXT_SLOT_SEC` constant
+ New related metrics for this reprocess work

Closes #3560

**Some considerations**
+ Ideally we should have 1 separate queue for reprocess work but since we have to hook to the life cycle of `js-libp2p-gossipsub`, I do this reprocess in the gossip handler
+ I increase attestation & AggregateAndProof gossip queue since we use single queue for this reprocess and Lighthouse has 2 separate queues, does this make sense? The test does not show any differences if I increase this

**Test**

~~I don't see any big differences with the metrics, except for the peer count with `subscribeAllSubnets` flag~~ => peer count is the same after running 2 instances for 9 days. Although total attestations reprocessed per slot is only around 20 - 60/slot, the total attestations accepted is a big difference after I leave 2 instances (contabo 19 vs cotabo 20) for 9 days
Screen Shot 2022-01-13 at 10 15 24

+ Master: Gossip aggregate and proofs accepted per slot
Screen Shot 2022-01-13 at 09 52 48

+ This branch:  Gossip aggregate and proofs accepted per slot is higher
Screen Shot 2022-01-13 at 09 53 31

+ Master: Gossip beacon attestations accepted per slot
Screen Shot 2022-01-13 at 09 58 50

+ This branch: Gossip beacon attestations accepted per slot is higher
Screen Shot 2022-01-13 at 09 59 12

### DIFF
--- a/packages/lodestar/src/chain/reprocess.ts
+++ b/packages/lodestar/src/chain/reprocess.ts
@@ -1,0 +1,159 @@
+import {Slot, RootHex} from "@chainsafe/lodestar-types";
+import {IMetrics} from "../metrics";
+import {MapDef} from "../util/map";
+
+/**
+ * To prevent our node from having to reprocess while struggling to sync,
+ * we only want to reprocess attestations if block reaches our node before this time.
+ */
+export const REPROCESS_MIN_TIME_TO_NEXT_SLOT_SEC = 2;
+
+/**
+ * Reprocess status for metrics
+ */
+enum ReprocessStatus {
+  /**
+   * There are too many attestations that have unknown block root.
+   */
+  reached_limit = "reached_limit",
+  /**
+   * The awaiting attestation is pruned per clock slot.
+   */
+  expired = "expired",
+}
+
+type AwaitingAttestationPromise = {
+  resolve: (foundBlock: boolean) => void;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  promise: Promise<boolean>;
+  // there are multiple subnet/aggregated attestations waiting for same promise
+  awaitingAttestationsCount: number;
+  // we only resolve to true or false to make the code simpler so no need to keep reject here
+  addedTimeMs: number;
+};
+
+// How many attestations (aggregate + unaggregate) we keep before new ones get dropped.
+const MAXIMUM_QUEUED_ATTESTATIONS = 16_384;
+
+type SlotRoot = {slot: Slot; root: RootHex};
+
+/**
+ * Some attestations may reach our node before the voted block, so we manage a cache to reprocess them
+ * when the block come.
+ * (n)                                               (n + 1)
+ *  |----------------|----------------|----------|------|
+ *                   |                |          |
+ *                  att           agg att        |
+ *                                              block
+ * Since the gossip handler has to return validation result to js-libp2p-gossipsub, this class should not
+ * reprocess attestations, it should control when the attestations are ready to reprocess instead.
+ */
+export class ReprocessController {
+  private readonly awaitingPromisesByRootBySlot: MapDef<Slot, Map<RootHex, AwaitingAttestationPromise>>;
+  private awaitingPromisesCount = 0;
+
+  constructor(private readonly metrics: IMetrics | null) {
+    this.awaitingPromisesByRootBySlot = new MapDef(() => new Map<RootHex, AwaitingAttestationPromise>());
+  }
+
+  /**
+   * Returns Promise that resolves either on block found or once 1 slot passes.
+   * Used to handle unknown block root for both unaggregated and aggregated attestations.
+   * @returns true if blockFound
+   */
+  waitForBlockOfAttestation(slot: Slot, root: RootHex): Promise<boolean> {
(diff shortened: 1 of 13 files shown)


twoeths:
thanks to the more gossip attestations we accepted, average peer score is an improvement too

+ master
Screen Shot 2022-01-13 at 10 38 42

+ this branch
Screen Shot 2022-01-13 at 10 39 02

dapplion:
@tuyennhv I've changed the design a bit as in my opinion make the change more manageable:
- I've moved the re-process waiting logic into the validation function itself. With that the logic doesn't have to be repeated and affects attestation, aggregate gossip validation and API validation.
- I've changed how the reprocess handles each case, making it always resolve instead of reject. That makes it much simpler to handle all cases:
  - if block found: resolve true
  - if queue full: resolve false
  - if timeout: resolve false
  
Please review my changes and keep them only if you agree. Please make sure I haven't broken any assumptions from our previous design

packages/lodestar/src/chain/reprocess.ts
```diff
@@ -0,0 +1,138 @@
...
+ * (n)                                               (n + 1)
+ *  |----------------|----------------|----------|------|
+ *                   |                |          |
+ *                  att           agg att        |
+ *                                              block
+ * Since the gossip handler has to return validation result to js-libp2p-gossipsub, this class should not
+ * reprocess attestations, it should control when the attestations are ready to reprocess instead.
+ */
+export class ReprocessController {
+  private readonly awaitingPromisesByRootBySlot: MapDef<Slot, MapDef<RootHex, AwaitingAttestationPromise[]>>;
```
wemeetagain: I wonder if we can make do with `MapDef<Slot, MapDef<RootHex, AwaitingAttestationPromise>>`
Reusing the same promise for attestations relying on the same block.
We would just be tracking the maximum wait time before resolve/reject per block.

dapplion: Good idea! Implemented :+1:

dapplion:
@tuyennhv I've moved back the code to handlers to implemented https://github.com/ChainSafe/lodestar/issues/3613 in the future without nasty stuff. I think it looks decently clean now and allows for extension here:

https://github.com/ChainSafe/lodestar/blob/dfc5af60cfc55e8754ea0cf15858801e5cd2ff2d/packages/lodestar/src/network/gossip/handlers/index.ts#L294

twoeths:
thanks @dapplion , it looks great now. The latest update works well on contabo-20 👍

wemeetagain (dismissed):

dapplion (dismissed):

dapplion:
@tuyennhv File conflicts from pass attestations to the forkchoice

dapplion (approved):
Looks great!

PR #3565, opened 2022-01-04
Remove unbounded set seenPeers

dapplion:
**Motivation**

PeerManager.seenPeers is an unbounded data structure that tracks all seen peerIds and never gets pruned. From https://github.com/ChainSafe/lodestar/issues/3446 concatenated strings take significant memory so for a long running node seenPeers can take significant memory

**Description**

- Remove unbounded set seenPeers

### DIFF
--- a/packages/lodestar/src/metrics/metrics/lodestar.ts
+++ b/packages/lodestar/src/metrics/metrics/lodestar.ts
@@ -68,10 +68,6 @@ export function createLodestarMetrics(
       help: "Total number of goodbye sent, labeled by reason",
       labelNames: ["reason"],
     }),
-    peersTotalUniqueConnected: register.gauge({
-      name: "lodestar_peers_total_unique_connected",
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/metrics/metrics/lodestar.ts
```diff
@@ -68,10 +68,6 @@ export function createLodestarMetrics(
       help: "Total number of goodbye sent, labeled by reason",
       labelNames: ["reason"],
     }),
-    peersTotalUniqueConnected: register.gauge({
-      name: "lodestar_peers_total_unique_connected",
```
twoeths, line 72: we need to remove this from `lodestar.json` too

dapplion: We can do that latter, many charts need adjustments in that row

wemeetagain (approved):

PR #3579, opened 2022-01-06
Updating SPEC_TEST_REPO_URL

0xFredrik:
Was renamed from eth2-clients to eth-clients


wemeetagain (approved):
Thanks @fredriksvantes

PR #3585, opened 2022-01-07
Apply gossip attestations to forkchoice

### LINKED ISSUE #1949: Apply gossip attestations to forkchoice
As noted in https://github.com/ChainSafe/lodestar/pull/1939#issuecomment-759604120

Currently we don't pass gossip attestations to forkchoice, we need to do it.

twoeths:
**Motivation**

Apply gossip attestations to forkchoice

**Description**

Closes #1949

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "test:spec-min": "lerna run test:spec-min --no-bail",
     "test:spec-fast": "lerna run test:spec-fast --no-bail",
     "test:spec-main": "lerna run test:spec-main --no-bail",
-    "benchmark": "yarn benchmark:files 'packages/*/test/perf/**/*.test.ts'",
+    "benchmark": "yarn benchmark:files packages/*/test/perf/**/*.test.ts",
     "benchmark:files": "LODESTAR_PRESET=mainnet NODE_OPTIONS=--max-old-space-size=4096 benchmark --config .benchrc.yaml",
     "publish:release": "lerna publish from-package --yes --no-verify-access",
     "release": "lerna version --no-push --sign-git-commit",
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "test:spec-min": "lerna run test:spec-min --no-bail",
     "test:spec-fast": "lerna run test:spec-fast --no-bail",
     "test:spec-main": "lerna run test:spec-main --no-bail",
-    "benchmark": "yarn benchmark:files packages/*/test/perf/**/*.test.ts",
+    "benchmark": "yarn benchmark:files 'packages/*/test/perf/**/*.test.ts'",
     "benchmark:files": "LODESTAR_PRESET=mainnet NODE_OPTIONS=--max-old-space-size=4096 benchmark --config .benchrc.yaml",
     "publish:release": "lerna publish from-package --yes --no-verify-access",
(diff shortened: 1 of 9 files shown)


dapplion:
Have you benchmarked the cost of doing this? I would love to see the CPU charts of a node subscribed to all subnets with this enabled and this disabled. Consider adding a flag with `hide: true` to be able to turn it on and off.

twoeths:
**Update**
+ only pass unaggregated gossip attestations to forkchoice if `shouldProcess` (i.e. there's an aggregator connecting to this node)
+ pass all aggregated gossip attestations to forkchoice
+ add `network.passGossipAttestationsToForkchoice` hidden cli flag, default as `true`
+ add benchmark: 700 aggregated attestations (as seen in prater/master) + 400 unaggregated attestations (assuming there are 3 aggregators connected to node), the result shows cpu time as ~30ms/slot

twoeths:
**Update**
Given the fact that for each slot, there are only limited set of AttestationData, we don't have to validate all attestation data of all attestations. In a good net work condition, there should be only 64 attestation data max (same slot, different committee index)

+ For AggregateAndProof messages, since they are tree backed, we already cache the attestation data root there. After a successful validation, add them to a `Set()` that's renew per slot. Bypass validation if we already validated that same AttestationData root.
+ For unaggregated attestations, they are just struct so using `hashTreeRoot()` there is not a good idea

with this improvement, total time to process gossip attestations in forkchoice is reduced from 30ms to more than 16ms (almost double the speed)

packages/cli/src/options/beaconNodeOptions/network.ts
```diff
@@ -16,6 +16,7 @@ export interface INetworkArgs {
   "network.blockCountTotalLimit": number;
   "network.blockCountPeerLimit": number;
   "network.rateTrackerTimeoutMs": number;
+  "network.passGossipAttestationsToForkchoice": boolean;
```
dapplion: I think we should make the flag in reverse such that the default is false. Change to
```
dontSendGossipAttestationsToForkchoice
```

package.json
```diff
@@ -24,7 +24,7 @@
     "test:spec-min": "lerna run test:spec-min --no-bail",
     "test:spec-fast": "lerna run test:spec-fast --no-bail",
     "test:spec-main": "lerna run test:spec-main --no-bail",
-    "benchmark": "yarn benchmark:files 'packages/*/test/perf/**/*.test.ts'",
+    "benchmark": "yarn benchmark:files packages/*/test/perf/**/*.test.ts",
```
dapplion: Could you review what error happened without quotes? The quotes are there to prevent bash expansion and let mocha do it.

dapplion: Warning! Many benchmarks are missing now https://github.com/ChainSafe/lodestar/runs/4772657466?check_suite_focus=true

twoeths: so I added the single quotes back but I have to change the root tsconfig.json
```
"compilerOptions": {
  "typeRoots": ["node_modules/@types", "./packages/lodestar/types"],
}
```

@dapplion do you see any issues with it, is it possible/necessary to provide a separate tsconfig.json just for `yarn:benchmark`?

dapplion: Let's roll with this and if anyone has issues find another strategy

dapplion (approved):
Looks great!

PR #3586, opened 2022-01-07
Ensure there are 10 epochs max in StateContextCheckpoint

### LINKED ISSUE #3583: Ensure 10 epochs maximum in CheckpointStateCache
**Describe the bug**

CheckpointStateCache is designed to have 10 epoch maximum but it's used to be up to 14-16 epochs during non-finality time in prater as shown in the below metrics

Screen Shot 2022-01-07 at 11 22 45

right now we prune per checkpoint, but if we have multiple skipped checkpoint slots then we'll not prune CheckpointStateCache for a while https://github.com/ChainSafe/lodestar/blob/d78d0c5790eb7010a8368698ff5d7f2cfb24331b/packages/lodestar/src/chain/blocks/importBlock.ts#L138

we should prune per epoch instead

**Expected behavior**

Max 10 epochs in CheckpointStateCache

twoeths:
**Motivation**

During non-finality period of prater, sometimes the StateContextCheckpoint contains up to 16 epochs, we want to ensure it's 10 max

**Description**

+ We control that by Checkpoint event but we miss emitting it when checkpoint is at a skipped slot
+ Add it back in regen module

Closes #3583

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -133,6 +133,7 @@ export class BeaconChain implements IBeaconChain {
       checkpointStateCache,
       db,
       metrics,
+      emitter,
       signal,
     });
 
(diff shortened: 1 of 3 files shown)


dapplion (approved):
This fix should do for now but it's not a long term solution. The node will still OOM if there are enough diverging states in the cache

g11tech (approved):

PR #3589, opened 2022-01-08
forward sync with wsCheckpoint

### LINKED ISSUE #3385: investigate: weak subjectivity checkpoint improper use

g11tech:
**Motivation**
Currently lodestar only provides starting sync from a `wsCheckpoint` (if wsCheckpoint is ahead of db state)  by syncing state from `weakSubjectivityServerUrl`.

This PR enables forward sync anchored in future by `wsCheckpoint` (while starting from DB/genesis), which basically validates finding `wsCheckpoint` as a finalized checkpoint as the chain syncs forward. 

However if a `weakSubjectivityServerUrl` is also provided, then lodestar will just use that url to fetch state and again sync from  wsCheckpoint and validate wsCheckpoint connecting to genesis in a backfill sync.

**Description**
example usage:
`./lodestar beacon --network prater --rootDir /data  --weakSubjectivityCheckpoint 0xbff6ef213be7c53b0181b2daf7ed68fe1ca29eebed04bf7ef46166c51fdcea55:65429`

This will result in the following successful log output (after it syncs past the wsCheckpoint epoch  65429 as a finalized checkpoint):
`Jan-08 09:58:22.913 [CHAIN]            info: wsCheckpoint found and validated as finalized root=0xbff6ef213be7c53b0181b2daf7ed68fe1ca29eebed04bf7ef46166c51fdcea55, epoch=65429`

Closes #3385

### DIFF
--- a/packages/cli/src/cmds/beacon/initBeaconState.ts
+++ b/packages/cli/src/cmds/beacon/initBeaconState.ts
@@ -90,6 +90,9 @@ export async function initBeaconState(
   // fetch the latest state stored in the db
   // this will be used in all cases, if it exists, either used during verification of a weak subjectivity state, or used directly as the anchor state
   const lastDbState = await db.stateArchive.lastValue();
+  const wsCheckpoint: Checkpoint | undefined = args.weakSubjectivityCheckpoint
+    ? getCheckpointFromArg(args.weakSubjectivityCheckpoint)
+    : undefined;
 
   if (args.weakSubjectivityStateFile) {
     // weak subjectivity sync from a provided state file:
@@ -99,24 +102,23 @@ export async function initBeaconState(
     const wsState = getStateTypeFromBytes(chainForkConfig, stateBytes).createTreeBackedFromBytes(stateBytes);
     const config = createIBeaconConfig(chainForkConfig, wsState.genesisValidatorsRoot);
     const store = lastDbState ?? wsState;
-    const checkpoint = args.weakSubjectivityCheckpoint
-      ? getCheckpointFromArg(args.weakSubjectivityCheckpoint)
-      : getCheckpointFromState(config, wsState);
+    const checkpoint = wsCheckpoint ?? getCheckpointFromState(config, wsState);
     return initAndVerifyWeakSubjectivityState(config, db, logger, store, wsState, checkpoint);
-  } else if (args.weakSubjectivitySyncLatest) {
+  } else if (args.weakSubjectivitySyncLatest || (wsCheckpoint && args.weakSubjectivityServerUrl)) {
     // weak subjectivity sync from a state that needs to be fetched:
-    // if a weak subjectivity checkpoint has been provided, it is used to inform which state to download and used for additional verification
+    // if a weak subjectivity checkpoint has been provided, it is used to inform which
+    // state to download and used for additional verification
     // otherwise, the 'finalized' state is downloaded and the state itself is used for verification (all trust delegated to the remote beacon node)
     const remoteBeaconUrl = args.weakSubjectivityServerUrl;
     if (!remoteBeaconUrl) {
       throw Error(`Must set arg --weakSubjectivityServerUrl for network ${args.network}`);
     }
 
-    let stateId = "finalized";
-    let checkpoint: Checkpoint | undefined;
-    if (args.weakSubjectivityCheckpoint) {
-      checkpoint = getCheckpointFromArg(args.weakSubjectivityCheckpoint);
-      stateId = (checkpoint.epoch * SLOTS_PER_EPOCH).toString();
+    let stateId: string;
+    if (wsCheckpoint != null) {
+      stateId = (wsCheckpoint.epoch * SLOTS_PER_EPOCH).toString();
+    } else {
+      stateId = "finalized";
     }
     const url = `${remoteBeaconUrl}/eth/v1/debug/beacon/states/${stateId}`;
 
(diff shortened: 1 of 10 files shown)


dapplion:
So this covers the case where I know what's the correct state in the future but I don't want to download it. Some issue I see:

- Isn't this a super niche use-case? While this has conceptual value, will users want to use this? Do other clients support this mode?
- I think the UX of this feature is confusing. For a user it will not be obvious that when providing `--weakSubjectivityCheckpoint` the node will **not** do a weak subjectivity sync. Instead it will do a full sync but with verification at the end.

For users weak subjectivity sync = jump straight to that state.

What about exposing this feature with a separate flag so that's obvious? Like `--forwardWeakSubjectivityCheckpoint` or something that clearly indicates that a weak subjectivity sync will not happen.

g11tech:
> So this covers the case where I know what's the correct state in the future but I don't want to download it. Some issue I see:
> 
Exactly! :slightly_smiling_face: 

>  Isn't this a super niche use-case? While this has conceptual value, will users want to use this? Do other clients support this mode?


Lighthouse has a cli option: 
```bash
 --wss-checkpoint <WSS_CHECKPOINT>
            Specify a weak subjectivity checkpoint in `block_root:epoch` format to verify the node's sync against. The
            block root should be 0x-prefixed. Note that this flag is for verification only, to perform a checkpoint sync
            from a recent state use --checkpoint-sync-url.
```
similarly teku: https://consensys.net/blog/blockchain-explained/how-to-safely-migrate-your-ethereum-2-0-validator-client/, https://docs.teku.consensys.net/en/latest/Reference/CLI/CLI-Syntax/#ws-checkpoint

I agree that this might appeal to only finicky people, who want to first verify and then participate in the network (weak subjectivity sync would verify retro actively using the backfill). Or to those who don't want to depend on loading a full state from a centralized provider (this %age of people could still be more than the over cautious ones).

I have used it couple of times while resetting lighthouse, but it was before when the weak subjectivity sync became a thing.
But it could still be handy in attack scenarios.

> 
>  I think the UX of this feature is confusing. For a user it will not be obvious that when providing `--weakSubjectivityCheckpoint` the node will **not** do a weak subjectivity sync. Instead it will do a full sync but with verification at the end.
> 
It would only do forward sync if `weakSubjectivityServerUrl` is not provided. but if one provides `weakSubjectivityServerUrl` and `weakSubjectivityCheckpoint` it would fetch the state from the server and start from there.

> 
> For users weak subjectivity sync = jump straight to that state.
> 
> What about exposing this feature with a separate flag so that's obvious? Like `--forwardWeakSubjectivityCheckpoint` or something that clearly indicates that a weak subjectivity sync will not happen.

Hmmm, what you are saying makes sense though I think if we clarify in the help that  if weakSubjectivityServerUrl is not provided, it will forward sync and verify, else it will sync state and back verify. If you think that would still not be clear enough for user, we can go a separate flag way.

dapplion:
> Hmmm, what you are saying makes sense though I think if we clarify in the help that if weakSubjectivityServerUrl is not provided, it will forward sync and verify, else it will sync state and back verify. If you think that would still not be clear enough for user, we can go a separate flag way.

I think this is not clear at all. I can see myself failing for this error constantly, where I don't set `weakSubjectivityServerUrl` and I don't understand why the node hasn't jumped to the checkpoint I provided. How could I user tell that this is not a bug? I can see lots of support cases generated because of this. Now, if you forget `weakSubjectivityServerUrl` the node immediately throws, which is much better.

g11tech:
> > Hmmm, what you are saying makes sense though I think if we clarify in the help that if weakSubjectivityServerUrl is not provided, it will forward sync and verify, else it will sync state and back verify. If you think that would still not be clear enough for user, we can go a separate flag way.
> 
> I think this is not clear at all. I can see myself failing for this error constantly, where I don't set `weakSubjectivityServerUrl` and I don't understand why the node hasn't jumped to the checkpoint I provided. How could I user tell that this is not a bug? I can see lots of support cases generated because of this. Now, if you forget `weakSubjectivityServerUrl` the node immediately throws, which is much better.

then the separate flag is the better approach, will update :+1:

dapplion:
Turning draft, since it has been sitting for a while. Can this PR only include the future check, if wsCp > current DB; but not if wsCp is in the past (backFill suicide mode)

g11tech:
> Turning draft, since it has been sitting for a while. Can this PR only include the future check, if wsCp > current DB; but not if wsCp is in the past (backFill suicide mode)

sure, i can do that and revive this :+1:

dapplion:
> > Turning draft, since it has been sitting for a while. Can this PR only include the future check, if wsCp > current DB; but not if wsCp is in the past (backFill suicide mode)
> 
> sure, i can do that and revive this +1

<3 note, not a priority, we were just reviewing old PRs with @wemeetagain

g11tech:
reworking here: https://github.com/ChainSafe/lodestar/pull/4604

PR #3593, opened 2022-01-08
Bugfix: don't throw error if finalized state not found for backfill ranges while archiving state

g11tech:
**Motivation**
In the backfill sync, in the archival process, updating backfill ranges required finalized slot. On not finding, it was made to error which messed up with the hot db to cold db movement of the blocks. This was observed in kintsugi network:
```
Jan-08 18:18:30.000 []                 info: Syncing - 8.5 hours left - 5.44 slots/s - slot: 167167 (skipped 165688) - head: 1479 0x81d0…f181 - execution: valid(0x32d6…a73e) - finalized: 0x5738…e489:43 - peers: 14
Jan-08 18:18:33.008 [CHAIN]           error: State for currentJustifiedCheckpoint not available, using closest state checkpointEpoch=46, checkpointRoot=0xdaf0940d9a9bc2bdf88ffd11b237104587290ccd5cb06441c6a2e96b447aec1f, stateSlot=1475, stateRoot=0x3ad08611a1a61a4431c3258f6a86e990ed3d0bbe746fb3f666d8ff8be6e3bd15
Jan-08 18:18:35.142 [CHAIN]           error: Error processing finalized checkpoint epoch=46 message=No state in cache for finalized checkpoint state epoch #46
Error: No state in cache for finalized checkpoint state epoch #46
    at StatesArchiver.maybeArchiveState (/usr/app/packages/lodestar/s
```

This PR fixes the issue by not throwing, as updating backfill ranges on a missed finalized state is not a critical as it will get updated next time around the archival happens. PR changes `throw` to an positive `if` scenario for updating backfill ranges.

### DIFF
--- a/packages/lodestar/src/chain/archiver/index.ts
+++ b/packages/lodestar/src/chain/archiver/index.ts
@@ -98,17 +98,22 @@ export class Archiver {
       // Mark the sequence in backfill db from finalized block's slot till anchor slot as
       // filled.
       const finalizedBlockFC = this.chain.forkChoice.getBlockHex(finalized.rootHex);
-      if (finalizedBlockFC && finalizedBlockFC.slot > this.chain.anchorSlot) {
-        await this.db.backfilledRanges.put(finalizedBlockFC.slot, this.chain.anchorSlot);
+      if (finalizedBlockFC && finalizedBlockFC.slot > this.chain.anchorStateLatestBlockSlot) {
+        await this.db.backfilledRanges.put(finalizedBlockFC.slot, this.chain.anchorStateLatestBlockSlot);
 
-        // Clear previously marked sequence till anchorSlot, without touching backfill sync
-        // process sequence which are at <=anchorSlot i.e. clear >anchorSlot and < currentSlot
+        // Clear previously marked sequence till anchorStateLatestBlockSlot, without
+        // touching backfill sync process sequence which are at
+        // <=anchorStateLatestBlockSlot i.e. clear >anchorStateLatestBlockSlot
+        // and < currentSlot
         const filteredSeqs = await this.db.backfilledRanges.keys({
-          gt: this.chain.anchorSlot,
+          gt: this.chain.anchorStateLatestBlockSlot,
           lt: finalizedBlockFC.slot,
         });
         await this.db.backfilledRanges.batchDelete(filteredSeqs);
-        this.logger.debug("Updated backfilledRanges", {key: finalizedBlockFC.slot, value: this.chain.anchorSlot});
+        this.logger.debug("Updated backfilledRanges", {
+          key: finalizedBlockFC.slot,
+          value: this.chain.anchorStateLatestBlockSlot,
+        });
       }
     } catch (e) {
       this.logger.error("Error updating backfilledRanges on finalization", {epoch: finalized.epoch}, e as Error);
--- a/packages/lodestar/src/chain/archiver/index.ts
+++ b/packages/lodestar/src/chain/archiver/index.ts
@@ -63,24 +63,52 @@ export class Archiver {
   };
 
   private processFinalizedCheckpoint = async (finalized: CheckpointWithHex): Promise<void> => {
+    const logError = (e: Error): void => {
+      this.logger.error("Error processing finalized checkpoint", {epoch: finalized.epoch}, e);
+    };
     try {
       const finalizedEpoch = finalized.epoch;
       this.logger.verbose("Start processing finalized checkpoint", {epoch: finalizedEpoch});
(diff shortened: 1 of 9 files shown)


dapplion:
As chatted offline:

- Since this code path covers the "forward sync" of blocks part of regular beacon node operations, the ranges should be updated update the archiveBlocks function. They are not related with states. https://github.com/ChainSafe/lodestar/blob/758a958aec9b92d9c8e60a6ea4cb113151059ff3/packages/lodestar/src/chain/archiver/index.ts#L70
- Add a nice comment explaining why the ranges are updated on every finalized event. Consider adding the code in a separate method with a meaningful name to clearly show that's a foreign step to the archival process.

packages/lodestar/src/chain/stateCache/stateContextCheckpointsCache.ts
```diff
@@ -83,7 +83,7 @@ export class CheckpointStateCache {
     return previousHits;
   }
 
-  pruneFinalized(finalizedEpoch: Epoch): void {
+  async pruneFinalized(finalizedEpoch: Epoch): Promise<void> {
```
g11tech: made this async so that processFinalizedCheckpoint doesn't need to wait for this to complete

dapplion: Please don't add unnecessary async functions. You can do error handling with try / catch for sync functions

packages/lodestar/src/chain/archiver/index.ts
```diff
@@ -65,24 +65,52 @@ export class Archiver {
...
-
       // tasks rely on extended fork choice
       this.chain.forkChoice.prune(finalized.rootHex);
+
+      // The following tasks need not be awaited and can run independently
+      // should be after ArchiveBlocksTask to handle restart cleanly
+      this.statesArchiver.maybeArchiveState(finalized).catch(logError);
+      this.updateBackfillRange(finalized).catch(logError);
+      this.chain.stateCache.deleteAllBeforeEpoch(finalizedEpoch).catch(logError);
+      this.chain.checkpointStateCache.pruneFinalized(finalizedEpoch).catch(logError);
```
dapplion: Restore previous sequence flow, maybeArchiveState is fully completed before pruning states

dapplion: Do you have any reason to believe that this prune functions will error very frequently? Why do they have dedicated error handling?

dapplion: Given that `this.updateBackfillRange(finalized).catch(logError);` is the least critical step it should be done last

dapplion: > Do you have any reason to believe that this prune functions will error very frequently? Why do they have dedicated error handling?

If no, just use the original single try / catch without any additional .catch(). If latter when running the node data shows that some of this statements error often we can add additional try / catch blocks

packages/lodestar/src/chain/chain.ts
```diff
@@ -110,7 +110,7 @@ export class BeaconChain implements IBeaconChain {
     this.logger = logger;
     this.metrics = metrics;
     this.genesisTime = anchorState.genesisTime;
-    this.anchorSlot = anchorState.slot;
+    this.anchorSlot = anchorState.latestBlockHeader.slot;
```
dapplion: `anchorSlot` is too vague and can be confusing since this is on the public facing API of the BeaconChain object. Since this is used just by an internal module use a long but precise name like:
`anchorStateLatestBlockSlot`

packages/lodestar/src/chain/archiver/index.ts
g11tech: done! :+1: thank you for pointing out the improper flow :heart:

packages/lodestar/src/chain/archiver/index.ts
```diff
@@ -69,18 +69,54 @@ export class Archiver {
...
-        this.chain.checkpointStateCache.pruneFinalized(finalizedEpoch),
-        this.chain.stateCache.deleteAllBeforeEpoch(finalizedEpoch),
-      ]);
+      await this.statesArchiver.maybeArchiveState(finalized);
 
+      this.chain.checkpointStateCache.pruneFinalized(finalizedEpoch);
+      this.chain.stateCache.deleteAllBeforeEpoch(finalizedEpoch);
       // tasks rely on extended fork choice
       this.chain.forkChoice.prune(finalized.rootHex);
+      void this.updateBackfillRange(finalized);
```
dapplion: While the errors are handled above, why not await for completeness? updateBackfillRange() should take too long

dapplion (approved):

PR #3600, opened 2022-01-10
Should only add to bootMultiaddrs if given bootEnrs supports TCP

### LINKED ISSUE #3599: Incorrect multiaddress gets added to networkOpts.bootMultiaddrs
**Describe the bug**
Incorrect multiaddress gets added to networkOpts.bootMultiaddrs when given networkOpts.discv5.bootEnrs does not support tcp.

If the `enr` value passed in via the `--network.discv5.bootEnrs` argument flag does not contain `tcp` entry, a garbled value is added to `network.bootMultiaddrs`

**Expected behavior**

Skip populating the `network.bootMultiaddrs` if passed in enr does not contain tcp

**Steps to Reproduce**

To reproduce start the node and pass an enr that does not contain tcp. After calling `createNodeJsLibp2p` the `network.bootMultiaddrs` array will contain an entry where the multiaddr is undefined. eg `undefined/p2p/16Uiu2HAmNj9zgHUrNj2LYymY5A2EC2Y4fdcAM7QFEDZS7stZL4Xq`

dadepo:
**Motivation**

Correctly parses enr passed in via `--network.discv5.bootEnrs` to avoid populating `network.bootMultiaddrs` with wrong values.

Closes #3599

### DIFF
--- a/packages/lodestar/src/network/nodejs/util.ts
+++ b/packages/lodestar/src/network/nodejs/util.ts
@@ -55,10 +55,11 @@ export async function createNodeJsLibp2p(
 
     for (const enrOrStr of networkOpts.discv5.bootEnrs) {
       const enr = typeof enrOrStr === "string" ? ENR.decodeTxt(enrOrStr) : enrOrStr;
-      const multiaddrTCP = enr.getLocationMultiaddr("tcp");
-      const peerId = await enr.peerId();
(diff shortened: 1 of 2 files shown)


wemeetagain (approved):
LGTM

twoeths (approved):

PR #3610, opened 2022-01-12
Update brand assets

sweetpea22:
**Motivation**

updating the assets to make experience of lodestar more awesome. kept file names the same so hopefully no breaking changes.


dapplion (changes requested):
Can you please ensure that the files are in similar pixel sizes and KB sizes? Github PRs are now finally very rich in image diff and show huge reduction in pixel sizes but increases in the KB size. Could you compress the PNGs with some compressor like https://compresspng.com/ ?

sweetpea22:
they're the same pixel size and i'll compress again

dapplion:
> they're the same pixel size and i'll compress again

According to the diff
- `assets/lodestar_icon_text_black.png` was  W: 3505px | H: 930px now is  W: 645px | H: 210px 
- `assets/lodestar_icon_text_black_stroke.png` was  W: 3505px | H: 930px  now is  W: 646px | H: 210px 
- `assets/lodestar_icon_text_white.png` was  W: 3580px | H: 970px  and now is  W: 645px | H: 210px

sweetpea22:
oooops, read heights as width 😭 i don't think we need the 3000px width since svgs will be readily available and it would save on bundle size, would that be ok with you?

dapplion:
> oooops, read heights as width sob i don't think we need the 3000px width since svgs will be readily available and it would save on bundle size, would that be ok with you?

Can we keep the original sizes with similar compression since the rendered versions are super convenient?

dapplion (approved):
Thank you so much for re-uploading the assets!

dapplion (dismissed):
Sorry the `assets/lodestar_icon_text_black_stroke.png` asset is missing a white stroke or it won't be legible in Github dark mode. Please if you can add a simple white stroke in similar fashion to the original asset

wemeetagain (approved):

PR #3614, opened 2022-01-13
Revert Offset sync period by ALTAIR_FORK_EPOCH #3477

### LINKED ISSUE #3494: Revert sync period offset

See https://github.com/ChainSafe/lodestar/pull/3477

In #3477 we offset the sync period calculation by `ALTAIR_FORK_EPOCH` so that the first sync period starts at 0.
This is nice in some ways, as a light client could start syncing from period 0 thru N, but it also comes with a small bit of complexity.

Now calculating the sync period requires a config, rather than just requiring the genesis time.

Revert this offset. This would simplify calculating the sync period.
The difference is that the first available sync period would not be 0. And this is not an issue, since nothing relies on this assumption. The sync period is always calculated from a header slot or epoch.

dapplion:
**Motivation**

ALTAIR_FORK_EPOCH does not necessarily start on a sync committee period boundary.

Closes https://github.com/ChainSafe/lodestar/issues/3494

**Description**

- Revert Offset sync period by ALTAIR_FORK_EPOCH #3477


wemeetagain:
https://github.com/ChainSafe/lodestar/runs/4800061070?check_suite_focus=true#step:7:22

wemeetagain (approved):

PR #3625, opened 2022-01-14
Add renderIrrelevantPeerType to flatten log

dapplion:
**Motivation**

Current master sends a nested object to the logger. This won't render properly since now we restrict to 1 level deep max

**Description**

Add renderIrrelevantPeerType to flatten log


wemeetagain (approved):
Lgtm

PR #3629, opened 2022-01-16
Fix issue 3611

### LINKED ISSUE #3611: Validator handler is expecting objects only in /keystores
**Describe the bug**

When other files, including the deposit_data.json are in the lodestar/keystores folder, the validator client returns the following error on startup:

```
Jan-13 04:43:12.373 []                 info: Lodestar version=v0.33.0 (npm), network=prater
 ✖ Error: : should be object
    at Object.validateKeystore (/usr/app/node_modules/@chainsafe/bls-keystore/lib/functional.js:111:15)
    at Function.fromObject (/usr/app/node_modules/@chainsafe/bls-keystore/lib/class.js:46:22)
    at Function.parse (/usr/app/node_modules/@chainsafe/bls-keystore/lib/class.js:53:25)
    at /usr/app/node_modules/@chainsafe/lodestar-cli/src/cmds/validator/keys.ts:65:44
    at Array.map (<anonymous>)
    at getSecretKeys (/usr/app/node_modules/@chainsafe/lodestar-cli/src/cmds/validator/keys.ts:64:21)
    at Object.validatorHandler [as handler] (/usr/app/node_modules/@chainsafe/lodestar-cli/src/cmds/validator/handler.ts:32:79)
```

**Expected behavior**

Lodestar validator should ignore files other than the keystore-m.json files, rather than crash with an error if other files are detected.

**Steps to Reproduce**
1. Put other files, specifically the `deposit_data.json` files in the /keystores folder. 
2. Run the validator client
3. Receive object expected error.

- OS: Ubuntu Server 20.04.3
- Version: v0.33.0
- Branch: v0.33.0 release branch
- Commit hash: 56de734

dadepo:
**Motivation**

Prevents lodestar from crashing when there are files that cannot be parsed as a keystore on startup.

**Description**

Have the parsing oppression in a try, so errors can be caught without terminating process.

Closes #3611

- Put other files, specifically the deposit_data.json files in the /keystores folder.
- Run the validator client
- Receive object expected error.

### DIFF
--- a/packages/cli/src/cmds/validator/keys.ts
+++ b/packages/cli/src/cmds/validator/keys.ts
@@ -5,7 +5,7 @@ import {SecretKey} from "@chainsafe/bls";
 import {deriveEth2ValidatorKeys, deriveKeyFromMnemonic} from "@chainsafe/bls-keygen";
 import {interopSecretKey} from "@chainsafe/lodestar-beacon-state-transition";
 import {defaultNetwork, IGlobalArgs} from "../../options";
-import {parseRange, stripOffNewlines, YargsError} from "../../util";
+import {parseRange, stripOffNewlines, warnLogger, YargsError} from "../../util";
 import {getLockFile} from "../../util/lockfile";
 import {ValidatorDirManager} from "../../validatorDir";
 import {getAccountPaths} from "../account/paths";
@@ -60,12 +60,24 @@ export async function getSecretKeys(
       lockFile.lockSync(lockFilePath);
     }
 
-    const secretKeys = await Promise.all(
-      keystorePaths.map(async (keystorePath) =>
-        SecretKey.fromBytes(await Keystore.parse(fs.readFileSync(keystorePath, "utf8")).decrypt(passphrase))
-      )
-    );
+    const secretKeysPromise = keystorePaths.map(async (keystorePath) => {
+      try {
+        const keystore = Keystore.parse(fs.readFileSync(keystorePath, "utf8"));
+        return SecretKey.fromBytes(await keystore.decrypt(passphrase));
+      } catch (e) {
+        return Promise.reject(`Failed to parse keystore file: ${keystorePath} with error ${(e as Error).toString()}`);
+      }
+    });
 
+    const results = await Promise.allSettled(secretKeysPromise);
+    const secretKeys: SecretKey[] = [];
+    for (const result of results) {
+      if (result.status !== "rejected") {
+        secretKeys.push(result.value);
+      } else {
+        warnLogger().warn(result.reason);
+      }
+    }
     return {
       secretKeys,
       unlockSecretKeys: () => {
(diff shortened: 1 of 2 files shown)


dapplion:
@dadepo Please name PRs with meaningful titles. Just puting the issue number doesn't give context. Specifically:

- Short and descriptive summary
- Should be capitalized and written in imperative present tense
- Not end with period

Example: "Add Edit on Github button to all the pages"

Also please name your branches starting with your github handle: `dadepo/fix-issue-3611`

----

Sorry that this info is not persisted in https://github.com/ChainSafe/lodestar/blob/master/CONTRIBUTING.md

packages/cli/src/cmds/validator/keys.ts
```diff
@@ -60,12 +60,24 @@ export async function getSecretKeys(
...
+      }
+    });
 
+    const results = await Promise.allSettled(secretKeysPromise);
+    const secretKeys: SecretKey[] = [];
+    for (const result of results) {
+      if (result.status !== "rejected") {
+        secretKeys.push(result.value);
+      } else {
+        warnLogger().warn(result.reason);
```
dapplion: Creating a new logger instance for every single log is not optimal

packages/cli/src/cmds/validator/keys.ts
```diff
@@ -60,12 +60,24 @@ export async function getSecretKeys(
       lockFile.lockSync(lockFilePath);
     }
 
-    const secretKeys = await Promise.all(
-      keystorePaths.map(async (keystorePath) =>
-        SecretKey.fromBytes(await Keystore.parse(fs.readFileSync(keystorePath, "utf8")).decrypt(passphrase))
-      )
-    );
+    const secretKeysPromise = keystorePaths.map(async (keystorePath) => {
```
dapplion, line 64: A better solution would be to filter the `keystorePaths` by known name formats. Specifically excluding files that look like the deposit data json: `deposit_data*.json`

dapplion: It also must apply to any place that keystores are read from disk, specially here https://github.com/ChainSafe/lodestar/blob/c581c18a0fd2f9db867c4f9852651bea64c51e4a/packages/cli/src/cmds/validator/keys.ts#L52

dadepo: > A better solution would be to filter the `keystorePaths` by known name formats. Specifically excluding files that look like the deposit data json: `deposit_data*.json`

I thought about this, but was not sure if the pattern of the filename should be the determinant or the content of the file itself? Like would the filename pattern always remain this way? If that is a yes then I can update to filter using the file name

dadepo: Also there could be situations where the file has the correct file name but fails to parse for whatever reason. The application should still not choke on startup but should emit a warning. So I’m still leaning on using the content as being a more robust approach... (except if failing to parse the file should be considered a fatal error?) Let me know what you think.

dapplion: > The application should still not choke on startup but should emit a warning

The application must error! This is a validator client, if it can not decrypt the keys it can not do anything. It MUST be a fatal error

dapplion: We can use the pattern `deposit_data*.json` safely because the only key generator tool recommended by all clients uses this very specific convention

https://github.com/ethereum/eth2.0-deposit-cli/blob/256ea2148cd088d6a8672bd0a3f5b12617159960/eth2deposit/credentials.py#L199

dadepo: > We can use the pattern `deposit_data*.json` safely because the only key generator tool recommended by all clients uses this very specific convention

Gotcha!

PR #3648, opened 2022-01-21
Trigger block search when receiving unknown block root attestations

### LINKED ISSUE #3217: Use synced peers only for unknown block sync
is it better to only get peers with `status.head >= ours`? it's no use to ask for blocks from syncing peers

_Originally posted by @tuyennhv in https://github.com/ChainSafe/lodestar/pull/3212#discussion_r712649826_

### LINKED ISSUE #3613: Trigger block search when receiving unknown block root gossip attestations

In our mainnet node, some gossip blocks are ignored
Screen Shot 2022-01-13 at 14 18 01

The log shows that it's all `BLOCK_ERROR_PARENT_UNKNOWN` unknown error

Given this example

```
Jan-13 00:37:47.577 [NETWORK]         ^[[34mdebug^[[39m: gossip - beacon_attestation - ignore slot=2934187, subnet=23, index=23, code=ATTESTATION_ERROR_UNKNOWN_BEACON_BLOCK_ROOT, root=0x144b0458d9fecfe537b2df6f07abda4e1ba5f708dcadb1454d0f11cad2789469
```

right now we only search for block until the next block comes
```
Jan-13 00:38:00.259 [NETWORK]         ^[[34mdebug^[[39m: Gossip block has error slot=2934188, root=0xa11e…3e12, code=BLOCK_ERROR_PARENT_UNKNOWN
Jan-13 00:38:00.527 [NETWORK]         ^[[34mdebug^[[39m: Req  sending request method=beacon_blocks_by_root, encoding=ssz_snappy, client=Prysm, peer=16...zQhddr, requestId=661829, requestBody=["0x144b0458d9fecfe537b2df6f07abda4e1ba5f708dcadb1454d0f11cad2789469"]
Jan-13 00:38:00.527 [NETWORK]         ^[[34mdebug^[[39m: Req  request sent method=beacon_blocks_by_root, encoding=ssz_snappy, client=Prysm, peer=16...zQhddr, requestId=661829
```

+ we should be able to search for the block right after we receive unknown block root gossip attestations. 
+ consider downscoring peer if we can't fetch that block.
+ this is related to #3560

twoeths:
**Motivation**

+ a missed block is only searched when we receive `PARENT_UNKNOWN` block error while usually we receive unknown block root attestations at previous slot

**Description**

+ Trigger block search when receiving unknown block attestations
+ Refactor Unknown block sync 
  + support both `unknown_block` and `unknown_parent` error
  +  only query synced peers, a synced peer is peer with `head >= current_slot - SLOTS_PER_EPOCH` at the time of exchanging status
  + don't query a peer for unknown block root if we already do that in less than 0.5s
  + only increase block attempt if we have at least 1 attempt to find block root
+ at each slot, there's a race between gossip block and UnknownBlock sync triggered by an unknown block root attestation, whichever comes first is fine to us
+ add related metrics

Closes #3613
Closes #3217

**Test**
This is from a node with `subscribeAllSubnets` flag

Screen Shot 2022-01-21 at 14 07 22

### DIFF
--- a/packages/lodestar/src/sync/interface.ts
+++ b/packages/lodestar/src/sync/interface.ts
@@ -80,6 +80,8 @@ export type PendingBlock = {
   | {type: PendingBlockType.UNKNOWN_BLOCK}
 );
 
+export type UnknownParentPendingBlock = PendingBlock & {type: PendingBlockType.UNKNOWN_PARENT};
+
 export enum PendingBlockStatus {
   pending = "pending",
   fetching = "fetching",
--- a/packages/lodestar/src/sync/interface.ts
+++ b/packages/lodestar/src/sync/interface.ts
@@ -2,6 +2,7 @@ import {ILogger} from "@chainsafe/lodestar-utils";
 import {allForks, RootHex, Slot, phase0} from "@chainsafe/lodestar-types";
 import {IBeaconConfig} from "@chainsafe/lodestar-config";
 import {routes} from "@chainsafe/lodestar-api";
+import PeerId from "peer-id";
 import {INetwork} from "../network";
 import {IBeaconChain} from "../chain";
 import {IMetrics} from "../metrics";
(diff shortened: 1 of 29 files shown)


twoeths:
It shows that Unknown Block Sync now triggers very frequently due to unknown block root gossip attestation, but most of the time gossip blocks still come before Unknown Block Sync, however it could help sometime

Screen Shot 2022-05-13 at 10 20 11

dapplion:
How does this interact with the cache for attestations that reference blocks still not seen? This triggers of unknown block sync, are due to attestations arriving before the block or actual blocks that where not received due to some other reason?

packages/lodestar/src/sync/interface.ts
```diff
@@ -58,14 +59,29 @@ export interface ISyncModules {
   wsCheckpoint?: phase0.Checkpoint;
 }
 
+export enum PendingBlockType {
+  UNKNOWN_BLOCK = "UNKNOWN_BLOCK",
+  UNKNOWN_PARENT = "UNKNOWN_PARENT",
```
dapplion: If used in metrics, don't use all caps for the value as it's not the best for reading. I would use
```
= "unknown_block"
```
or
```
 = "unknownBlock"
```

twoeths:
> How does this interact with the cache for attestations that reference blocks still not seen? 

This does not interact with that cache, it just help finding the unknown block faster if the gossip block comes late. 

> This triggers of unknown block sync, are due to attestations arriving before the block or actual blocks that where not received due to some other reason?

it's due to attestations arriving before the block

dapplion:
@tuyennhv Just pushed a commit to simplify the data structures and make the code agnostic to if the request comes unknown parent or unknown block.

Also, regarding metrics:
- Only add labels to where it provides real value. Adding labels everywhere makes creating dashboard much harder and adds cost to metrics processing. Adding labels to the histograms won't add much value for now.
- For the "type" label, just track the origin of the block: api | gossip | range sync | unknown block, and use simple counters for submitted, processed.

dapplion:
- Closing for https://github.com/ChainSafe/lodestar/pull/5485

PR #3651, opened 2022-01-21
Remove SSZ from params

dapplion:
**Motivation**

Currently config and params use SSZ to define the type of its values. This approach is sub-optimal because:
- Config API is too strict, some node may return not all values and SSZ will throw an Error, breaking the route
- CLI doesn't do enough verification, we may want to validate that any additional params are not ignored
- SSZ is overkill, since only the toJson fromJson methods are used, not merkleization bytes serialization etc

**Description**

- Remove SSZ from params


packages/config/src/chainConfig/types.ts
```diff
@@ -1,10 +1,11 @@
-/* eslint-disable @typescript-eslint/naming-convention */
 import {PresetName} from "@chainsafe/lodestar-params";
 
+/* eslint-disable @typescript-eslint/naming-convention */
+
 /**
  * Run-time chain configuration
  */
-export interface IChainConfig {
+export type IChainConfig = {
```
wemeetagain, line 8: why change to type?

packages/config/src/chainConfig/types.ts
```diff
@@ -47,4 +48,70 @@ export interface IChainConfig {
   DEPOSIT_CHAIN_ID: number;
   DEPOSIT_NETWORK_ID: number;
   DEPOSIT_CONTRACT_ADDRESS: Uint8Array;
-}
+};
+
+export const chainConfigTypes: SpecTypes<IChainConfig> = {
+  PRESET_BASE: "string",
```
wemeetagain, line 57: should these values be enum? or overkill?

packages/params/src/index.ts
```diff
@@ -23,7 +25,7 @@ presetStatus.frozen = true;
  *
  * The active preset can be manually overridden with `setActivePreset`
  */
-export const ACTIVE_PRESET: PresetName =
+export const ACTIVE_PRESET =
```
wemeetagain, line 28: is this type now inferred to be `PresetName`?

packages/params/src/interface/phase0.ts
```diff
@@ -16,7 +15,7 @@ export interface IPhase0Preset {
   SAFE_SLOTS_TO_UPDATE_JUSTIFIED: number;
 
   // Gwei Values
-  MIN_DEPOSIT_AMOUNT: bigint;
+  MIN_DEPOSIT_AMOUNT: number;
```
wemeetagain, line 18: This isn't used anywhere?! I guess its hard to argue against changing it to a number

packages/params/src/presets/mainnet/bellatrix.ts
```diff
@@ -1,7 +1,7 @@
-/* eslint-disable @typescript-eslint/naming-convention */
-import {IBellatrixPreset} from "../../preset";
+import {BellatrixPreset} from "../../interface/bellatrix";
 
-export const bellatrix: IBellatrixPreset = {
+/* eslint-disable @typescript-eslint/naming-convention */
```
wemeetagain, line 3: ? is this needed here?

packages/config/src/chainConfig/types.ts
```diff
@@ -1,10 +1,11 @@
-/* eslint-disable @typescript-eslint/naming-convention */
 import {PresetName} from "@chainsafe/lodestar-params";
 
+/* eslint-disable @typescript-eslint/naming-convention */
```
wemeetagain, line 3: Is this needed here?

wemeetagain (dismissed):
Structure lgtm

packages/params/src/index.ts
dapplion: Apparently it's not needed, I can't think of a reason why it would be necessary. Not sure if it's a legacy forgotten though or an old uncommented reason

packages/config/src/chainConfig/types.ts
dapplion: Otherwise Typescript doesnt allow this to be used as a Record<string,unknown> argument in functions

packages/config/src/chainConfig/types.ts
dapplion: Overkill? This whole approach will be review soon anyway to make parsing im the validator client and cli more resilient

packages/params/src/interface/phase0.ts
dapplion: Yeah, not used anywhere 🤷

packages/params/src/presets/mainnet/bellatrix.ts
dapplion: Yes to allow having all caps

packages/config/src/chainConfig/types.ts
dapplion: Yes to allow having all caps

g11tech (dismissed):
LGTM :heart: , as discussed on discord, can/will do a followup PR to fill config for beacon <> validator interop and to match preset more deeply :+1:

g11tech (approved):

PR #3660, opened 2022-01-22
Use the correct api version for wss sync

g11tech:
**Motivation**
To do the wss sync, we only use v1 to fetch the state, but post phase0 this end point may only be available in v2.

This PR determines which end point (v1/v2) to use based upon the epoch of the checkpoint at which to fetch the state.

### DIFF
--- a/packages/cli/src/cmds/beacon/initBeaconState.ts
+++ b/packages/cli/src/cmds/beacon/initBeaconState.ts
@@ -2,7 +2,7 @@ import {AbortSignal} from "@chainsafe/abort-controller";
 import {ssz} from "@chainsafe/lodestar-types";
 import {TreeBacked} from "@chainsafe/ssz";
 import {createIBeaconConfig, IBeaconConfig, IChainForkConfig} from "@chainsafe/lodestar-config";
-import {fromHex, ILogger} from "@chainsafe/lodestar-utils";
+import {ILogger} from "@chainsafe/lodestar-utils";
 import {computeEpochAtSlot, allForks} from "@chainsafe/lodestar-beacon-state-transition";
 import {IBeaconDb, IBeaconNodeOptions, initStateFromAnchorState, initStateFromEth1} from "@chainsafe/lodestar";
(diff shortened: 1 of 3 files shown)


packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -112,26 +112,23 @@ export async function initBeaconState(
...
     const store = lastDbState ?? wsState;
-    return initAndVerifyWeakSubjectivityState(
-      config,
-      db,
-      logger,
-      store,
-      wsState,
-      checkpoint || getCheckpointFromState(config, wsState)
-    );
+    return initAndVerifyWeakSubjectivityState(config, db, logger, store, wsState, checkpoint);
```
dapplion: All this logic would be better placed if in `fetchWeakSubjectivityState` I think, including the fetch of the checkpoint. By printing the checkpoint epoch + the lodestar version it's enough info to know version was used

packages/cli/src/networks/index.ts
```diff
@@ -136,12 +143,25 @@ export function enrsToNetworkConfig(enrs: string[]): RecursivePartial<IBeaconNod
...
+    let wsCheckpoint;
+    if (weakSubjectivityCheckpoint) {
+      wsCheckpoint = getCheckpointFromArg(weakSubjectivityCheckpoint);
+    } else {
+      wsCheckpoint = await fetchFinalizedCheckpoint(
+        `${weakSubjectivityServerUrl}/eth/v1/beacon/states/finalized/finality_checkpoints`
+      );
+    }
+    const stateSlot = wsCheckpoint.epoch * SLOTS_PER_EPOCH;
+    const apiVersion = config.getForkName(stateSlot) === ForkName.phase0 ? "v1" : "v2";
```
dapplion: Should you use the lodestar-api client for this? Since we already have it, and exposes both v1 and v2 returning bytes

dapplion (approved):

PR #3661, opened 2022-01-24
Migrate to libp2p > 0.35.8 and gossipsub 0.12.2

### LINKED ISSUE #3548: Migrate to libp2p-gossipsub and libp2p

In libp2p-gossipsub 0.12.x, we have some improvement regarding caching peer score, accepting messages without computing peer score, fast message id cache and we want to use it in lodestar soon

As a preparation for that:

+ Migrate libp2p from 0.32.4 to 0.35.x
+ Migrate libp2p-gossipsub from 0.11.1 to 0.12.x
+ Migrate libp2p-interface from 1.1.0 to 2.0.x

twoeths:
**Motivation**

We want to migrate to latest versions of libp2p and gossipsub which contains some improvements to be applied to lodestar

**Description**

Closes #3548

TODO:
- [x] wait for libp2p to be released, latest master fixed the typing issue
- [x] use datastore-core 7.0.1
- [ ] Test on a node


packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -151,31 +149,32 @@ export class Eth2Gossipsub extends Gossipsub {
...
-          this._processRpc(idB58Str, peerStreams, rpcMsg).catch((e) => {
-            this.log("_processRpc error", (e as Error).stack);
-          });
-        }
-      });
-    } catch (err) {
-      this._onPeerDisconnected(peerStreams.id, err as Error);
-    }
-  }
+  // async _processMessages(
```
twoeths: need some confirmation before removing this function completely

wemeetagain:
js-libp2p v0.36.0 just released

packages/lodestar/src/network/gossip/gossipsub.ts
dapplion: This is the override that suppress the queue in gossipsub right?

wemeetagain: No this override was originally used to stop this behavior:
```ts
for await (const bytes of source) {
  await this.processRpc(decodeBytes(bytes));
}
```

But the recent upstream now essentially does whats commented out below:
```ts
for await (const bytes of source) {
  this.processRpc(decodeBytes(bytes)).catch(...);
}
```

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -1,5 +1,8 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import Gossipsub from "libp2p-gossipsub";
+import {messageIdToString} from "libp2p-gossipsub/src/utils/messageIdToString";
```
dapplion, line 3: Is it safe to import from `/src`? Can this be exported at root?

packages/lodestar/src/network/peers/datastore.ts
```diff
@@ -0,0 +1,187 @@
...
+ * This datastore includes a memory datastore and fallback to db datastore
+ * Use an in-memory datastore with last accessed time and _maxMemoryItems, on start it's empty (lazy load)
+ * - get: Search in-memory datastore first, if not found search from db.
+ *     - If found from db, add back to the in-memory datastore
+ *     - Update lastAccessedMs
+ * - put: move oldest items from memory to db if there are more than _maxMemoryItems items in memory
+ *     -  update memory datastore, only update db datastore if there are at least _threshold dirty items
+ *     -  Update lastAccessedMs
+ */
+export class Eth2PeerDataStore extends BaseDatastore {
```
dapplion, line 23: Cool design! :fire: :raised_hands:

packages/lodestar/src/network/peers/metastore.ts
```diff
@@ -1,4 +1,4 @@
-import MetadataBook from "libp2p/src/peer-store/metadata-book";
+import {MetadataBook} from "libp2p/src/peer-store/types";
```
dapplion: Safe to import from `/src`?

packages/lodestar/test/e2e/network/peers/peerManager.test.ts
```diff
@@ -122,6 +122,9 @@ describe("network / peers / PeerManager", function () {
     // We get a ping by peer1, don't have it's metadata so it gets requested
     networkEventBus.emit(NetworkEvent.reqRespRequest, {method: ReqRespMethod.Ping, body: seqNumber}, peerId1);
 
+    // Allow to set metadata in onPing
+    await sleep(0);
```
dapplion: Why is this necessary? Could you clarify "Allow to set metadata in onPing"?

dapplion (dismissed):
Looks good overall! Some minor comments

packages/lodestar/src/network/peers/metastore.ts
twoeths: libp2p bundles `src` in its npm package and we've used it for a while so I think it's safe, @wemeetagain could you confirm

packages/lodestar/test/e2e/network/peers/peerManager.test.ts
twoeths: method `onPing` of PeerManager becomes async so we need this line to make sure it complete and we can confirm `reqResp.metadata.callCount` below

twoeths:
warning: we'll get the below error when starting the node if we use the old peerstore dir:

```
Feb-08 11:46:06.834 [API]              info: Started REST api server address=http://0.0.0.0:9596, namespaces=["beacon","config","debug","events","lightclient","lodestar","node","validator"]
 ✖ Error: Unable to decode multibase string "addrs", base32 decoder only supports inputs prefixed with b
    at Decoder.decode (/root/lodestar/node_modules/libp2p/node_modules/multiformats/cjs/src/bases/base.js:35:17)
    at Codec.decode (/root/lodestar/node_modules/libp2p/node_modules/multiformats/cjs/src/bases/base.js:84:25)
    at PersistentStore.all (/root/lodestar/node_modules/libp2p/src/peer-store/store.js:256:26)
    at DefaultPeerStore.getPeers (/root/lodestar/node_modules/libp2p/src/peer-store/index.js:52:24)
    at NodejsNode._onDidStart (/root/lodestar/node_modules/libp2p/src/index.js:698:22)
    at NodejsNode.start (/root/lodestar/node_modules/libp2p/src/index.js:389:7)
    at Network.start (/root/lodestar/packages/lodestar/src/network/network.ts:134:5)
    at Function.init (/root/lodestar/packages/lodestar/src/node/nodejs.ts:210:5)
    at Object.beaconHandler [as handler] (/root/lodestar/packages/cli/src/cmds/beacon/handler.ts:84:18)
```

removing it or pointing to a new peerstore dir should get us through the error

dapplion:
> removing it or pointing to a new peerstore dir should get us through the error

What the keys that changed encoding? Should there be a migration routine?

dapplion:
I suggest to just change the data key and abandon the old data. No need to prune that

twoeths:
> > removing it or pointing to a new peerstore dir should get us through the error
> 
> What the keys that changed encoding? Should there be a migration routine?

0.36.x stores all peer data in a single common name space https://github.com/libp2p/js-libp2p/blob/v0.36.0/src/peer-store/store.js#L50
0.32.x stores different peer data in different name spaces https://github.com/libp2p/js-libp2p/blob/v0.32.0/src/peer-store/persistent/index.js#L14

so I don't see we can migrate easily, is peer store data is so valuable to do a migration?
@wemeetagain do you have any ideas regarding the data migration?

wemeetagain:
I think we should do something like:
- On startup, check for old peer store. If found, destroy the db.

dapplion:
> I think we should do something like:
> 
>     * On startup, check for old peer store. If found, destroy the db.

How do you differentiate between the old peerstore and the new peerstore? With your idea now Lodestar has to mantain this legacy code migration forever which is annoying

twoeths:
> > I think we should do something like:
> > ```
> > * On startup, check for old peer store. If found, destroy the db.
> > ```
> 
> How do you differentiate between the old peerstore and the new peerstore? With your idea now Lodestar has to mantain this legacy code migration forever which is annoying

the other thing is we don't want to maintain old libp2p logic of store/load peer data in lodestar code

Can we do this migration step manually? For each release, if there are migration guide we should note it somewhere and include it in the release.

I don't see if's critical to keep the existing peer store, I'd go with removing the peerstore folder manually before we deploy the new version.

dapplion:
What's the problem with abandoning the previous data? Can't we do that?

wemeetagain:
> the other thing is we don't want to maintain old libp2p logic of store/load peer data in lodestar code

Right, I was thinking we would use some heuristic to determine if its the old version. Eg: Before startup, fetch a few keys and attempt a pattern match.

> Can we do this migration step manually? For each release, if there are migration guide we should note it somewhere and include it in the release.
> I don't see if's critical to keep the existing peer store, I'd go with removing the peerstore folder manually before we deploy the new version.

Definitely not critical to keep the existing peer store. Its just a matter of what we expect the UX of running and upgrading lodestar to be.
I had thought that we want folks to be able to just npm install / docker pull the next version and have it "just continue to work" -- to the best of our abilities.
If we go with manual peerstore removal, we will be breaking this workflow. Maybe thats not a problem if we just write a migration guide and publish it alongside the release. But it will probably waste a few people's time.

> What's the problem with abandoning the previous data? Can't we do that?

Doing nothing results in the fatal error @tuyennhv posted above. Some intervention is needed to get lodestar to run.

packages/lodestar/src/network/peers/metastore.ts
wemeetagain: Yeah its safe

philknows:
From a UX perspective, breaking this workflow will just require some good communication through the release notes/blog/etc. and it's better to do this now while most of our current users are generally still technical people. I'm ok with either method of a migration guide or @wemeetagain 's script to determine and delete the old peerstore. We should just get moving on this because it's a blocker on some of the other problems we are seeing.

packages/cli/src/migrations/deleteOldPeerstore.ts
```diff
@@ -0,0 +1,31 @@
...
+import {LevelDbController} from "@chainsafe/lodestar-db";
+import {ILogger} from "@chainsafe/lodestar-utils";
+
+/**
+ * As of libp2p v0.36.0 (https://github.com/libp2p/js-libp2p/commit/978eb3676fad5d5d50ddb28d1a7868f448cbb20b)
+ * the peerstore format has changed in a breaking way.
+ *
+ * Because of that, we need to wipe the old peerstore if it exists.
+ */
+export async function deleteOldPeerstore(peerStoreDir: string, logger: ILogger): Promise<void> {
```
dapplion: Name with exact version, "old" is too relative. `deletePeerstoreV036`

packages/cli/src/migrations/deleteOldPeerstore.ts
```diff
@@ -0,0 +1,31 @@
...
+  const db = new LevelDbController({name: peerStoreDir}, {logger});
+  await db.start();
+
+  const keys = await db.keys();
```
dapplion: The issue with this approach is that we need to read the entire db always. What if:
- We add a new key value pair with metadata information like the version of the libp2p consumer of this peerstore
- If this key is not found, delete db + write key after wipe
- If key is found use for future migrations

wemeetagain: This DB isn't managed by lodestar. If you do do that, there will be a fatal error, see @tuyennhv 's error above
The peerstore doesn't currently allow any stray keys in the database, nor should we (lodestar) be using this DB for that purpose.

dapplion: Oh see, how many keys are there in a typical peer store? We should do a rough benchmark of the performance cost of reading all of them at start-up, at least to be aware

**EDIT**: Not an issue now after checking only one key

packages/cli/src/migrations/deleteOldPeerstore.ts
```diff
@@ -12,11 +12,18 @@ export async function deleteOldPeerstore(peerStoreDir: string, logger: ILogger):
   const db = new LevelDbController({name: peerStoreDir}, {logger});
   await db.start();
 
-  const keys = await db.keys();
+  // only read a single key
+  const keys = await db.keys({limit: 1});
```
dapplion, line 16: Oh cool! That's much better :+1:

dapplion (dismissed):
Migration solution looks good to me! I have not tested this to make sure it doesn't break existing nodes

packages/cli/src/migrations/deleteOldPeerstore.ts
twoeths: should we also rename the file? for example "migrations/migrate-to-v0.35.0.ts". Also, maybe lodestar should have its own `metadata` repo to store current version and use it the next time we start the node

twoeths:
have tested this branch in contabo-19, it shows good result regarding "Gossip Block Processed Delay" metric, no Unknown Block sync happened
Screen Shot 2022-02-16 at 17 04 56

while other nodes (contabo-5, same number of validators connected) need helps a lot from UnknownBlock sync
Screen Shot 2022-02-16 at 17 07 14

Memory is stable (ranges from 2GB - 2.7GB)

wemeetagain (dismissed):
Approving but not merging

dapplion (changes requested):
I'm really disliking having to convert so many functions to async due to the peerStore. We keep a lot of stuff in memory: all operations pool, state caches, etc. And now we are introducing eager writes and reads for so little data like the peerstore. It doesn't make sense when gossip keeps all its scores in memory.

- score: It's just a number. You can keep thousands and not cause a memory issue. Prune after X time, like 1 hour after disconnection.
- addressBook: Only relevant while the peer is connected, prune after disconnection so bounded by peer count
- metadata: Same as above, only relevant for connected peers

packages/lodestar/src/network/peers/datastore.ts
```diff
@@ -0,0 +1,187 @@
...
+import {Key, KeyQuery, Query, Options, Pair} from "interface-datastore";
+
+type MemoryItem = {
+  lastAccessedMs: number;
+  data: Uint8Array;
+};
+
+/**
+ * Before libp2p 0.35, peerstore stays in memory and periodically write to db after n dirty items
+ * This has a memory issue because all peer data stays in memory and loaded at startup time
```
dapplion, line 12: Have we ever documented a memory issue specifically due to the peerstore? The information for the peerstore doesn't have to be kept around forever

twoeths: no we haven't. I just found the memory issue from the description of this PR https://github.com/ChainSafe/js-libp2p-gossipsub/pull/182

packages/lodestar/src/network/peers/peerManager.ts
```diff
@@ -560,7 +560,7 @@ export class PeerManager {
       if (openCnx) {
         const direction = openCnx.stat.direction;
         peersByDirection.set(direction, 1 + (peersByDirection.get(direction) ?? 0));
-        const client = getClientFromPeerStore(openCnx.remotePeer, this.libp2p.peerStore.metadataBook);
+        const client = await getClientFromPeerStore(openCnx.remotePeer, this.libp2p.peerStore.metadataBook);
```
dapplion, line 563: Shit forgot about this one too. @tuyennhv can you keep the client also in memory in the same way that this PR? https://github.com/ChainSafe/lodestar/pull/3838 there's not need to write the client implementation to disk. Do it in another PR I can review quickly once ready and unblock this one

dapplion (changes requested):

packages/lodestar/src/network/peers/peerManager.ts
dapplion: @tuyennhv Just opened https://github.com/ChainSafe/lodestar/pull/3841 please take a look and feel free to commit on top if necessary

twoeths: @dapplion "AgentVersion" is written by libp2p to metadataBook and we can't control that

we can duplicate this data to our PeerManager PeerData, the issue is we also need this data in req/resp log too. Right now we always query from libp2p peerstore metadataBook

twoeths: addressed in #3841

dapplion (approved):
I'm down to merge as is. @tuyennhv can you complete the fixes for in-memory db on a future PR?

wemeetagain (approved):

PR #3667, opened 2022-01-25
Use node:* import protocol for safer module resolution

### LINKED ISSUE #3664: Use `node:*` import protocol for safer module resolution
See https://nodejs.org/api/esm.html#esm_node_imports

Forces a specific import to be loaded exclusively as a built-in NodeJS library. Current imports `"fs"` can be overwritten by a malicious party if it manages to add a node_modules package with the same name.

:+1: security

ghost:
This PR uses `node:*` imports for `fs` and `path` modules.

Note that using `node:*` imports requires us to whitelist `node:fs` and `node:path` in the `import/core-modules` `eslint` setting. If we do not whitelist it then our linter fails because it can not identify `node:fs` as a node module.

Closes [3664](https://github.com/ChainSafe/lodestar/issues/3664)


wemeetagain (dismissed):
LGTM

Thanks @chainoperator

ghost:
@wemeetagain is there an issue with the Contributor License Agreement bot? I keep signing but it's not showing as signed in this PR.

CleanShot 2022-01-27 at 20 49 11@2x

wemeetagain:
@chainoperator 
you made the commit as @johnomarkid  so that account must sign the CLA too

ghost:
@wemeetagain good catch! That's my work account :) I amended the commit to use the correct account.

Looks like the workflows need approval again since I switched accounts. Sorry about that.

wemeetagain (approved):

dapplion (approved):
Thank you so much @chainoperator !

ghost:
My pleasure!

Just curious, what's the process for getting PRs merged after they have been approved?

philknows:
We usually take care of it from there, thanks for the contribution @chainoperator!

PR #3669, opened 2022-01-26
Optimize backfill sync to efficiently use reqresp fetched block data 

### LINKED ISSUE #3657: Skip serializing blocks when persisting to db

A profile from contabo-17 which has low peer count shows that it takes 8% to serialize blocks due to Backfill sync:
Screen Shot 2022-01-22 at 17 09 02

When fetching blocks from p2p, we already have binary data, we should be able to use that to persist to db without having to serialize() again

g11tech:
**Motivation**
As @tuyennhv pointed out after doing profiler runs, backfill sync could be optimized to use treebacked data from the sync range/block by root.

This PR changes the following:
- Use treebacked block's hashtree root of the synced block to get its blockroot to verify parent/child relationship in verify block sequence
- Use blockArchive's batch put binary using treebacked block's data
- add a hidden cli option `--sync.backfillBatchSize` to speciify the batch size for backfill sync
**Description**

Closes #3657

### DIFF
--- a/packages/lodestar/src/network/reqresp/encodingStrategies/sszSnappy/decode.ts
+++ b/packages/lodestar/src/network/reqresp/encodingStrategies/sszSnappy/decode.ts
@@ -3,13 +3,19 @@ import varint from "varint";
 import {CompositeType} from "@chainsafe/ssz";
 import {MAX_VARINT_BYTES} from "../../../../constants";
 import {BufferedSource} from "../../utils";
-import {RequestOrResponseType, RequestOrIncomingResponseBody} from "../../types";
+import {
+  RequestOrResponseType,
+  RequestOrIncomingResponseBody,
+  cachedTreeBackedProxyHandler,
+  cachedProxyHandler,
+} from "../../types";
 import {SnappyFramesUncompress} from "./snappyFrames/uncompress";
 import {maxEncodedLen} from "./utils";
 import {SszSnappyError, SszSnappyErrorCode} from "./errors";
 
 export interface ISszSnappyOptions {
   deserializeToTree?: boolean;
+  cacheBytes?: boolean;
 }
 
 /**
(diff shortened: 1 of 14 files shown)


packages/lodestar/src/sync/backfill/backfill.ts
```diff
@@ -821,6 +825,21 @@ export class BackfillSync extends (EventEmitter as {new (): BackfillSyncEmitter}
     }
     if (error) throw new BackfillSyncError({code: error});
   }
+
+  private async archiveBlocksTreeBacked(blocks: TreeBacked<allForks.SignedBeaconBlock>[]): Promise<void> {
+    const blockEntries: BlockArchiveBatchPutBinaryItem[] = blocks.map((signedBlock) => {
+      const block = signedBlock.message as TreeBacked<allForks.BeaconBlock>;
+      return {
+        key: block.slot,
+        value: signedBlock.serialize(),
```
twoeths: what I mean in the issue is that we return both the binary and the signed block in `beaconBlocksByRange` and `beaconBlocksByRoot`. So instead of returning `Promise<allForks.SignedBeaconBlock[]>` in network req/resp api, return something like

```typescript
type CachedBytes<T> = {bytes: Uint8Array} & T;

async beaconBlocksByRange(
    peerId: PeerId,
    request: phase0.BeaconBlocksByRangeRequest
  ): Promise<CachedBytes<allForks.SignedBeaconBlock>[]>
```

and consume that cached `bytes` without having to call `serialize()` again in BackFill sync and similar places

@dapplion do we have this type in ssz v2?

g11tech: yes, I was wondering about the same :slightly_smiling_face:
This format should currently save on hashTreeRoot calcs which would amount for 9% of hashTreeRoot in verifiBlockSequence (https://github.com/ChainSafe/lodestar/issues/3657#issuecomment-1019160634), and may be some 2-3% on the serialize. 

CachedBytes<allForks.SignedBeaconBlock> would be awesome :heart: awaiting @dapplion 's input regarding ssz v2.

wemeetagain: Right now, ssz v2 only (optionally) caches the hashTreeRoot (of non-tree-backed values); doesn't cache the serialized bytes.

wemeetagain: https://github.com/ChainSafe/ssz/pull/223/files#diff-bcfddac4baebdcc3adc704a754ecd03d1b5f8c78ba7234f3ac088e4042337e06R32-R38

packages/lodestar/src/network/reqresp/encodingStrategies/sszSnappy/decode.ts
```diff
@@ -138,9 +144,18 @@ function deserializeSszBody<T extends RequestOrIncomingResponseBody>(
   try {
     if (options?.deserializeToTree) {
       const typeTree = (type as unknown) as CompositeType<Record<string, unknown>>;
-      return (typeTree.createTreeBackedFromBytes(bytes) as unknown) as T;
+      const treeBacked = typeTree.createTreeBackedFromBytes(bytes) as unknown;
+      if (options?.cacheBytes) {
+        return (new Proxy({treeBacked, bytes}, cachedTreeBackedProxyHandler) as unknown) as T;
```
g11tech, line 144: Needed another proxy as tree's proxy wasn't letting assign `bytes` without going through the tree setters. May be post ssz v2 merge, it could be removed.

packages/lodestar/src/sync/backfill/backfill.ts
g11tech: added caching capability as that was discussed to be out of scope for ssz v2 :+1:

dapplion:
Closing for now since the network code has diverged. This optimization is good and should definitely be included on a future review of backfill sync

PR #3672, opened 2022-01-27
Connect to network discv5 boot enrs on start up

### LINKED ISSUE #3568: Documentation on how to run Lodestar on local testnet is no longer up to date

The documentation on how to run Lodestar on local testnet [here](https://chainsafe.github.io/lodestar/usage/local/) is no longer accurate and hence following the instructions does not work.

On a quick look I found the following issues:

- It states that _State will be written to .tmp/state.ssz_ It seems state is now stored in $XDG_DATA_HOME/lodestar/$NETWORK/chain-db instead
- The instructions included the `sync.minPeers` flag. This flag no longer exist
- Instructions states to pass `--genesisStateFile ./dev/genesis.ssz` to the cli, but no mention on how to get the `genesis.ssz` file
- The second command to run lodestar should probably include `--api.rest.port` to set the port number to something other than the default. Without this, a connection refused error prevents it from running as it classes with the first instance.

Having fixed the above issue locally, and was able to run the lodestar instances it looks like everything is still not quite right As I only keep getting the follow log message:

> Eph 0/11 7.864 [VALI]             info: Node is syncing - Error on getProposerDuties - Service Unavailable: Node is syncing - waiting for peers
Eph 0/12 0.001 []                 info: Searching peers - peers: 0 - slot: 2065010 (skipped 2065010) - head: 0 0x8c0e…ee87 - finalized: 0x0000…0000:0

Looks like no peer is being found.

I'll like to find out the correct steps needed to currently have lodestar run in local testnet. After that, update the documentation as needed.

dadepo:
**Motivation**

While using the dev command it was realised that the documentation on how to start two nodes that connects to each other is outdated. This changes updates the documentation and also fixes the code to ensures nodes are connected as peers based on supplied discv5.boot.enrs values

**Description**

Updated the documentation for using the dev command.
Updated the code to ensure nodes are connected as peers based on supplied discv5.boot.enrs values.

Closes #3568

### DIFF
--- a/docs/usage/local.md
+++ b/docs/usage/local.md
@@ -30,6 +30,7 @@ Start the second node without starting any validators and connect to the first n
   --rootDir /path/to/node2 \
   --port 9001 \
   --api.rest.port 9597 \
+  --network.connectToDiscv5Bootnodes true \
   --network.discv5.bootEnrs <enr value>
   --reset
 ```
@@ -40,7 +41,10 @@ the `--startValidators` option. Passing a value of `0:0` means no validators sho
 Also, take note that the values of `--genesisValidators` and `--genesisTime` must be the same as the ones passed to the first node in other for the two nodes
 to have the same beacon chain. 
 
-Finally `port` and `api.rest.port` are supplied since the default values will already be in use by the first node.
+Finally `--port` and `--api.rest.port` are supplied since the default values will already be in use by the first node.
+
+The `--network.connectToDiscv5Bootnodes` flags needs to be set to true as this is needed to allow connection to boot enrs on local devnet. 
+The exact enr of node to connect to is then supplied via the `--network.discv5.bootEnrs` flag.
 
 Once the second node starts, you should see an output similar to the following in either of the terminals:
 
--- a/docs/usage/local.md
+++ b/docs/usage/local.md
@@ -1,58 +1,76 @@
 # Local testnet
 
-To quickly test and run Lodestar we recommend to start a local testnet. We recommend a simple configuration of two beacon nodes with multiple validators
+To quickly test and run Lodestar we recommend starting a local testnet. We recommend a simple configuration of two beacon nodes with multiple validators
 
 **Terminal 1**
 
-Run a beacon node with 8 validators and default settings. State will be written to .tmp/state.ssz
+Run a beacon node, with 8 validators with the following command. 
 
 ```bash
-./lodestar dev --genesisValidators 8 --reset
+./lodestar dev --genesisValidators 8 --genesisTime 1578787200 --network prater --rootDir </path/to/node1> --reset
 ```
+
+`--genesisValidators` and `--genesisTime` define the genesis state of the beacon chain. `--rootDir` defines a path where lodestar should store the beacon state
+while the `--reset` flag ensures the state is cleared on each restart - which is useful when testing locally.
+
+Once the node has started, make a request to `curl http://localhost:9596/eth/v1/node/identity` and copy the `enr` value. 
+
+This would be used to connect from the second node.
+
+> enr stands for ethereum node records, which is a format for conveying p2p connectivity information for ethereum nodes.
+> For more info see [eip-778](https://eips.ethereum.org/EIPS/eip-778)
 
 **Terminal 2**
 
-Connect to bootnode (node 1 default multiaddrs) but without starting validators.
+Start the second node without starting any validators and connect to the first node by supplying the copied `enr` value:
 
 ```bash
 ./lodestar dev --startValidators 0:0 \
-  --genesisStateFile ./dev/genesis.ssz \
-  --network.localMultiaddrs /ip4/127.0.0.1/tcp/30607 \
-  --sync.minPeers 1
+  --genesisValidators 8 --genesisTime 1578787200 \
+  --network prater \
+  --rootDir /path/to/node2 \
+  --port 9001 \
+  --api.rest.port 9597 \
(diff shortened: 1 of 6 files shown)


docs/usage/local.md
```diff
@@ -1,58 +1,76 @@
...
+To quickly test and run Lodestar we recommend starting a local testnet. We recommend a simple configuration of two beacon nodes with multiple validators
 
 **Terminal 1**
 
-Run a beacon node with 8 validators and default settings. State will be written to .tmp/state.ssz
+Run a beacon node, with 8 validators with the following command. 
 
 ```bash
-./lodestar dev --genesisValidators 8 --reset
+./lodestar dev --genesisValidators 8 --genesisTime 1578787200 --network prater --rootDir </path/to/node1> --reset
```
wemeetagain: maybe remove `--network prater` here and below

I think this is a little confusing as an example in the docs, folks may think they are actually connecting to the prater testnet

packages/lodestar/src/network/peers/discover.ts
```diff
@@ -127,6 +127,8 @@ export class PeerDiscovery {
     await this.discv5.start();
     this.discv5StartMs = Date.now();
     this.discv5.on("discovered", this.onDiscovered);
+    // on start, dial the discv5.bootEnrs that has been added to the routing table
```
wemeetagain: I think we have two choices here:
1. only do this if network.connectToDiscv5Bootnodes is true
2. remove the network.connectToDiscv5Bootnodes option entirely

wemeetagain (changes requested):

docs/usage/local.md
g11tech: may be `network` option can be removed from the dev cli options, so that it will throw error if someone even tries to specify

packages/lodestar/src/network/peers/discover.ts
dapplion: We are not supposed to connect to the Mainnet bootnodes, this has to be behind a cli flag and used only for local devnets

packages/validator/src/validator.ts
```diff
@@ -86,6 +86,9 @@ export class Validator {
   /** Waits for genesis and genesis time */
   static async initializeFromBeaconNode(opts: ValidatorOptions, signal?: AbortSignal): Promise<Validator> {
     const {config} = opts.dbOps;
+    // TODO [DA] is there any objection to renaming api to client?
+    // it then becomes obvious this is a client that makes a request to an api
+    // instead of this being an api that is being made available
```
dapplion: Sounds good to me! `api` name is an old legacy option so feel free to change

docs/usage/local.md
dadepo: > may be `network` option can be removed from the dev cli options, so that it will throw error if someone even tries to specify

Good tip. I'll see to that

packages/lodestar/src/network/peers/discover.ts
dadepo: I have added it behind the `network.connectToDiscv5Bootnodes` cli flag

packages/validator/src/validator.ts
dadepo: Good to know! I'll remove the comment now but do the renaming in a separate PR (with other renaming propositions) as I think it is different from the main changes in this PR. I left that comment as just questions I was having about the code not necessary intended to be part of this change.

docs/usage/local.md
```diff
@@ -1,58 +1,75 @@
...
 ./lodestar dev --startValidators 0:0 \
-  --genesisStateFile ./dev/genesis.ssz \
-  --network.localMultiaddrs /ip4/127.0.0.1/tcp/30607 \
-  --sync.minPeers 1
+  --genesisValidators 8 --genesisTime 1578787200 \
+  --rootDir /path/to/node2 \
+  --port 9001 \
+  --api.rest.port 9597 \
+  --network.discv5.bootEnrs <enr value>
+  --reset
```
g11tech, line 35: may be this will now need connectToDiscv5Bootnodes arg as well?

dadepo: Indeed. Updated.

docs/usage/local.md
dadepo: @g11tech so I looked into your suggestion. The `network` option is a global option and hence the changes that will be required to make it not applicable when the dev cli option is used feels a bit out of scope to this PR. So I created a separate issue: #3695 for the suggestion to be implemented separately.

g11tech (dismissed):

packages/lodestar/src/network/peers/discover.ts
```diff
@@ -127,6 +130,13 @@ export class PeerDiscovery {
     await this.discv5.start();
     this.discv5StartMs = Date.now();
     this.discv5.on("discovered", this.onDiscovered);
+    if (this.connectToDiscv5BootnodesOnStart) {
+      // We are not supposed to connect to the Mainnet bootnodes
+      // so only connect to them on local devnet if the  cli flag
+      // connectToDiscv5Bootnodes has been set to true. If so then
+      // on start, dial the discv5.bootEnrs that has been added to the routing table
```
wemeetagain:
```suggestion
      // In devnet scenarios, especially, we want more control over which peers we connect to.
      // Only dial the discv5.bootEnrs if the option
      // network.connectToDiscv5Bootnodes has been set to true.
```

wemeetagain (approved):

g11tech (approved):

PR #3684, opened 2022-01-31
Increase default timeout of api instances except for validator duties

g11tech:
**Motivation**
A user reported over discord that lodestar was timing out doing a wss state fetch from infura. On a bit of digging through, figured out that lodestar uses timeout of 12 seconds, which is fine for validator calling beacon api, but could be relaxed else where, as discussed with @dapplion [here on discord](https://discord.com/channels/593655374469660673/593655641445367808/937360994710204456)

**Description**
This PR raises the default  api timeout from 12 seconds to 40 seconds, which should be a comfortable range for downloading the mainnet sync state and adds couple of comments to enure/highlight validator using a strict 1 slot timeout.


dapplion (approved):
Looks good!

wemeetagain (approved):

PR #3697, opened 2022-02-05
Use correct fork type to create state proof

wemeetagain:
**Motivation**

Offline discussion with @zsfelfoldi:
"Looks like "latestExecutionPayloadHeader" field id not supported, even if lodestar is running kintsugi..."

**Description**

Properly switch the type (based on fork) when creating state proofs.


g11tech (approved):

zsfelfoldi:
Works fine now, thanks!

Issue #3692, opened 2022-02-04
Docker containers on Docker Hub for x64, arm64, and with a single multi-arch manifest that contains both


philknows:
**Is your feature request related to a problem? Please describe.**

Required for Rocketpool implementation.

**Describe the solution you'd like**

Docker containers on Docker Hub for x64, arm64, and with a single multi-arch manifest that contains both.

philknows:
We need to upstream changes from blst-ts before closing.

dapplion:
> We need to upstream changes from blst-ts before closing.

Linking current blocker https://github.com/ChainSafe/blst-ts/pull/59

philknows:
Was merged. Ready to close @dapplion ?

dapplion:
Changes still not live in lodestar. Requires a release in blst-ts then a version bump here in the mono-repo. @dadepo could you do the PRs if avail?

dadepo:
Sure. I'll get to it today

PR #3698, opened 2022-02-05
Add multiarch docker build to CI

wemeetagain:
**Motivation**

See #3692

**Description**

Followed instructions here: https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

### DIFF
--- a/.github/workflows/release-nightly.yml
+++ b/.github/workflows/release-nightly.yml
@@ -91,14 +91,16 @@ jobs:
     needs: npm
     steps:
       - uses: actions/checkout@v2
-      - name: install buildx
-        id: buildx
-        uses: crazy-max/ghaction-docker-buildx@v1
-        with:
-          version: latest
+      # https://github.com/docker/setup-qemu-action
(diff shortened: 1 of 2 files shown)


dapplion:
Looks good! Tho we should do some checks on the image to ensure:
- It's in the right arch: `docker run chainsafe/lodestar uname -a`
- Lodestar actually runs: `docker run chainsafe/lodestar --help`

wemeetagain:
I don't think we need the architecture check (this is what the buildx tool does).

The --help check should be there to ensure the image was built properly (on our side).

dapplion (approved):
Looks good!

g11tech (approved):

dapplion:
Oh @wemeetagain I've experience with qemu and we can try this tricks to speed up the build
- The lodestar source (lib .js files) is platform agnostic, so we can build that in the host platform that should be much faster

wemeetagain:
The docker file just installs via npm so I don't think there's much that can be done to speed it up.

PR #3712, opened 2022-02-07
Dedupe transition spec runner and refac fork runner

g11tech:
This PR dedupes `transition` spec runner and does a small refac on the `fork` runner.


dapplion (approved):
Looks great! thanks :heart:

PR #3716, opened 2022-02-07
Check, validate and skip if we got any deposit events already present in DB

### LINKED ISSUE #3674: Handle recurring ETH1_ERROR_DUPLICATE_DISTINCT_LOG
On doing test runs of kintsugi with nethermind with deposit tracker enabled, following deposit cache error log dump was observed:

```
{
  depositEvents: [
    { blockNumber: 7521, index: 9, depositData: [Object] },
    { blockNumber: 8167, index: 10, depositData: [Object] }
  ]
}
Jan-27 20:10:02.474 [ETH1]            error: Error updating eth1 chain cache code=ETH1_ERROR_DUPLICATE_DISTINCT_LOG, newIndex=9, prevIndex=11
Error: ETH1_ERROR_DUPLICATE_DISTINCT_LOG
    at Eth1DepositsCache.add (/root/lodestar/packages/lodestar/src/eth1/eth1DepositsCache.ts:45:15)
    at Eth1DepositDataTracker.updateDepositCache (/root/lodestar/packages/lodestar/src/eth1/eth1DepositDataTracker.ts:164:5)
    at Eth1DepositDataTracker.update (/root/lodestar/packages/lodestar/src/eth1/eth1DepositDataTracker.ts:144:33)
    at Eth1DepositDataTracker.runAutoUpdate (/root/lodestar/packages/lodestar/src/eth1/eth1DepositDataTracker.ts:118:29)
parseDepositLog {
  log: {
```

This error shows that the tracker supposedly got a duplicate entry.  However once this happens, this error repeats ad infinitum

Expected: Investigate the error , if genuine the error should be handled and tracker should move ahead, else find/fix the bug causing the error.

g11tech:
**Motivation**
Previously, we were checking and throwing error if we got any deposit events which were already in DB. however a "reset" in remote EL/eth1 ' head can cause our tracker to import events which are already in DB causing our tracker to throw error and preventing the node from proposing.

This PR fixes that behavior by checking and validating if we get events which are already in DB. If the event doesn't match up it throws the DUPLICATE_DISTINCT_LOG error as it means either the DB is corrupted or the EL/eth1 did a reorg, both catastrophic.

Closes #3674

### DIFF
--- a/packages/cli/src/options/beaconNodeOptions/eth1.ts
+++ b/packages/cli/src/options/beaconNodeOptions/eth1.ts
@@ -7,6 +7,7 @@ export interface IEth1Args {
   "eth1.providerUrls": string[];
   "eth1.depositContractDeployBlock": number;
   "eth1.disableEth1DepositDataTracker": boolean;
+  "eth1.unsafeAllowDepositDataOverwrite": boolean;
 }
 
 export function parseArgs(args: IEth1Args): IBeaconNodeOptions["eth1"] {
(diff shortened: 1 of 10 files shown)


dapplion (approved):
Really tidy now! Thanks for the logic re-write and the comments :heart:

wemeetagain (approved):

PR #3718, opened 2022-02-07
Cache effectiveBalanceIncrements as Uint8Array

dapplion:
**Motivation**

I propose to change effectiveBalance array we keep in the epochCtx from MutableVector to a Uint8Array

**Description**

- Change it from effectiveBalance cache to effectiveBalanceIncrements cache
- Since EFFECTIVE_BALANCE_INCREMENT is a hard constant in the preset the increment is always less that 1 byte. The size of the data will be ~300KB (1 byte per validator)
- On every epoch transition we just clone the Uint8Array, cloning bytes is really cheap
- We will have extremely fast writes and extremely fast reads

Since effectiveBalances don't change much at all using MutableVector makes sense. However a Uint8Array will be extremely cheap on memory and since we are not expected to have 10s or 100s of states per epoch.

From @tuyennhv 

> the reason we use MutableVector there is cheap clone and shared memory. Divide it by EFFECTIVE_BALANCE_INCREMENT so that we can use Uint8Array sounds good potentially. The other down side is we have to multiply by EFFECTIVE_BALANCE_INCREMENT in all calculation, hope it does not cause cpu time.

Actually I think it reduces the amount of multiplications because we end up dividing by EFFECTIVE_BALANCE_INCREMENT almost everywhere

**Future proof**

May main worry is:
- What's the chance of MAX_EFFECTIVE_BALANCE_INCREMENTS > 255

Right now it's not practically possible to increase MAX_EFFECTIVE_BALANCE_INCREMENTS in Lodestar since all parameters that affect it can't be changed by consumers. However if we wanted to change the source to support higher MAX_EFFECTIVE_BALANCE_INCREMENTS I've made it easy to be able to change from Uint8Array to another TypedArray or a regular array.

**Memory**

Memory wise, it will be a net reduction since we are not supposed to keep more than 14 effectiveBalanceIncrements in memory at once. While MutableVector is extremely cheap for clones with small modifications the initial cost is still significant compared to Uint8Array

```
EffectiveBalanceIncrements Uint8Array 300000              - 299873.5 bytes / instance
EffectiveBalanceIncrements array 300000                   - 2400093.1 bytes / instance
EffectiveBalanceIncrements MutableVector 300000           - 4380557.0 bytes / instance
EffectiveBalanceIncrements MutableVector 300000 cloned 10 - 4399575.0 bytes / instance
```

**Performance**

Uint8Arrays are very cheap to copy. MutableVectors are even cheaper but at this scales doesn't matter (0.054 ms vs 0.00028 ms).

The game changer is with read and write speeds, where Uint8Array is 500 times faster.

```
  effectiveBalanceIncrements
    ✓ effectiveBalanceIncrements clone Uint8Array 300000                  18214.60 ops/s    54.90100 us/op        -      61779 runs   3.64 s
    ✓ effectiveBalanceIncrements clone MutableVector 300000                3496503 ops/s    286.0000 ns/op        -    5928161 runs   3.33 s
    ✓ effectiveBalanceIncrements rw all Uint8Array 300000                 3470.523 ops/s    288.1410 us/op        -       1750 runs  0.805 s
    ✓ effectiveBalanceIncrements rw all MutableVector 300000              8.313258 ops/s    120.2898 ms/op        -         21 runs   3.11 s
```


wemeetagain:
> However if we wanted to change the source to support higher MAX_EFFECTIVE_BALANCE_INCREMENTS I've made it easy to be able to change from Uint8Array to another TypedArray or a regular array.

:+1: likely Uint16Array would be plenty, but we can cross that bridge if we need to

packages/beacon-state-transition/src/allForks/util/epochContext.ts
```diff
@@ -163,9 +163,8 @@ export function createEpochContext(
     }
 
     // TODO: Should have 0 for not active validators to be re-usable in ForkChoice
-    effectiveBalancesArr.push(validator.effectiveBalance);
+    effectiveBalanceIncrements[i] = validator.effectiveBalance / EFFECTIVE_BALANCE_INCREMENT;
```
twoeths: need `Math.floor()` here

packages/beacon-state-transition/src/allForks/util/epochContext.ts
```diff
@@ -129,14 +129,14 @@ export function createEpochContext(
   let exitQueueEpoch = computeActivationExitEpoch(currentEpoch);
   let exitQueueChurn = 0;
 
-  const effectiveBalancesArr: number[] = [];
+  const validators = readonlyValuesListOfLeafNodeStruct(state.validators);
+  const validatorCount = validators.length;
+
+  const effectiveBalanceIncrements = new Uint8Array(getEffectiveBalanceIncrementsByteLen(validatorCount));
```
twoeths, line 93: in what case `effectiveBalanceIncrements` length is different from `validatorCount`? could it be `new Uint8Array(validatorCount)`

packages/beacon-state-transition/src/allForks/block/processDeposit.ts
```diff
@@ -75,7 +75,12 @@ export function processDeposit(fork: ForkName, state: CachedBeaconStateAllForks,
       slashed: false,
     });
     state.balanceList.push(Number(amount));
-    epochCtx.effectiveBalances.push(effectiveBalance);
+
+    // Updating here is better than updating at once on epoch transition
+    // - Simplify genesis fn applyDeposits(): effectiveBalanceIncrements is populated immediately
+    // - Keep related code together to reduce risk of breaking this cache
+    // - Should have equal performance since it sets a value in a flat array
+    epochCtx.effectiveBalanceIncrementsSet(validators.length, effectiveBalance);
```
twoeths, line 83: in average how many deposits we process per block? this create new `Uint8Array`, could we do this in batch?

dapplion: The effectiveBalanceIncrements Uint8Array is recreated in intervals of 1024 items. Once the length exceeds 1024 the length will be bumped to 2048.

packages/beacon-state-transition/src/allForks/util/epochContext.ts
dapplion: effectiveBalanceIncrements will be equally or greater than validatorCount. The `getEffectiveBalanceIncrementsByteLen()` modulo is used to reduce the frequency at which its Uint8Array is recreated. if effectiveBalanceIncrements has length greater than validatorCount it's not a problem since those values would never be accessed.

packages/beacon-state-transition/src/allForks/util/epochContext.ts
dapplion: Fixed

packages/beacon-state-transition/src/allForks/util/epochContext.ts
twoeths: that's great, please drop a comment in that function

twoeths (approved):
Looks good to me!

PR #3726, opened 2022-02-07
Cache syncCommittee contribution participant count

dapplion:
**Motivation**

Our current syncCommittee contribution opPool has the simple rules of:
- Replace existing contribution if the new one has more participants

This logic is okay, but the implementation does unnecessary work:
- De-serialize Signature for aggregation, before it's really necessary
- Recompute participation count when it's already computed in validation

**Description**

- Cache syncCommittee contribution participant count. Return validation artifact similar to what's done now with aggregates


twoeths (approved):
lgtm

PR #3729, opened 2022-02-08
Switch to Proxy-less SyncCommitteeCached

dapplion:
Blocked by https://github.com/ChainSafe/lodestar/pull/3728

**Motivation**

With SSZ refactor we have the option to drop the CachedBeaconState Proxies and added flat vectors.
- Performance benchmarks have shown that going through Proxies is significantly slower that regular objects
- Memory benchmarks have shown that data duplication is significant.

This PR moves SyncCommittee caches from Proxied CachedBeaconState to epochCtx. Overall the logic is very similar, but this re-organization allows to drop the root Proxy eventually

**Description**

- Switch to Proxy-less SyncCommitteeCached


packages/beacon-state-transition/src/util/genesis.ts
```diff
@@ -89,7 +89,7 @@ export function getGenesisBeaconState(
   state.eth1Data = genesisEth1Data;
   state.randaoMixes = randaoMixes;
 
-  return createCachedBeaconState(config, state);
+  return createCachedBeaconState(config, state, {skipSyncCommitteeCache: true});
```
wemeetagain, line 100: Why is this added here? Don't we want the cache populated if the chain is initialized with ALTAIR_FORK_EPOCH=0 ?

packages/lodestar/src/chain/lightClient/proofs.ts
```diff
@@ -1,9 +1,9 @@
-import {allForks} from "@chainsafe/lodestar-types";
+import {altair} from "@chainsafe/lodestar-types";
 import {TreeBacked} from "@chainsafe/ssz";
 import {FINALIZED_ROOT_GINDEX} from "@chainsafe/lodestar-params";
 import {SyncCommitteeWitness} from "./types";
 
-export function getSyncCommitteesWitness(state: TreeBacked<allForks.BeaconState>): SyncCommitteeWitness {
+export function getSyncCommitteesWitness(state: TreeBacked<altair.BeaconState>): SyncCommitteeWitness {
```
wemeetagain, line 6: shouldn't this stay `allForks` to allow for bellatrix and future forks?

wemeetagain (dismissed):

packages/beacon-state-transition/src/allForks/util/epochContext.ts
```diff
@@ -570,6 +631,28 @@ export class EpochContext {
 
     this.effectiveBalanceIncrements[index] = Math.floor(effectiveBalance / EFFECTIVE_BALANCE_INCREMENT);
   }
+
+  /**
+   * Note: The range of slots a validator has to perform duties is off by one.
+   * The previous slot wording means that if your validator is in a sync committee for a period that runs from slot
+   * 100 to 200,then you would actually produce signatures in slot 99 - 199.
+   */
+  getIndexedSyncCommittee(slot: Slot): SyncCommitteeCache {
```
twoeths, line 646: Some consumers have to convert from epoch to slot, and this converts from slot to syncPeriod.
Should we separate this function to `getIndexedSyncCommitteeAtSlot()` and `getIndexedSyncCommitteeAtEpoch()`

packages/beacon-state-transition/src/allForks/util/epochContext.ts
```diff
@@ -362,7 +379,36 @@ export class EpochContext {
    */
   exitQueueChurn: number;
 
-  constructor(data: IEpochContextData) {
+  /** TODO: Indexed SyncCommitteeCache */
```
twoeths: this SyncCommitteeCache is same to the old IndexedSyncCommittee, so drop this TODO and below?

packages/beacon-state-transition/src/allForks/util/epochContext.ts
dapplion: Added getIndexedSyncCommitteeAtEpoch

packages/beacon-state-transition/src/allForks/util/epochContext.ts
dapplion: Fixed

packages/lodestar/src/chain/lightClient/proofs.ts
dapplion: allForks doesn't have the fields for sync committees and this doesn't compile. The correct type would be allForks except phase0, but with current allforks types that's not possible. With SSZ refactor I can do that since allForks is purely a union

packages/beacon-state-transition/src/util/genesis.ts
dapplion: We need a CachedBeaconState to run processDeposit() which uses various caches. However at this point the state's syncCommittees are not known. This function can be called by:
  - 1. genesis spec tests: Don't care about the committee cache
  - 2. genesis builder: Only supports starting from genesis at phase0 fork
  - 3. interop state: Only supports starting from genesis at phase0 fork

So it's okay to skip syncing the sync committee cache here and expect it to be populated latter when the altair fork happens for cases 2, 3.

dapplion: Note that in current master the sync committee caches are populated with wrong data, because the logic wasn't checking that the pubkeys in the state's syncCommittee actually existed

twoeths: maybe we just need to drop a comment why we skip computing sync committees here, master works the same way (and it misses the comment that make me search for a while)

dapplion: Already added the comment https://github.com/ChainSafe/lodestar/pull/3729/commits/7218fb011524b1cef43930f2adc6e8efe7305b7a

twoeths (approved):

dapplion:
@wemeetagain I think all concerns are covered, and I'll merge to continue ssz refactor related PRs. If you have more concerns please comment and I can open fix PRs

PR #3733, opened 2022-02-08
Mask out the infura style wss credentials for the log

### LINKED ISSUE #3703: Weak Subjectivity sync outputs credentials to log
Weak subjectivity sync from a URL that uses credentials will log those credentials. E.g.
`Fetching weak subjectivity state from https://PROJECTID:PROJECTSECRET@eth2-beacon-mainnet.infura.io/eth/v1/debug/beacon/states/finalized`

It'd be better if it displayed only the FQDN, such as
`Fetching weak subjectivity state from https://eth2-beacon-mainnet.infura.io`. That way, credentials as auth or as tokens in the path will not be displayed.

g11tech:
**Motivation**
A user requested to mask out the infura style credentials of the wss url that we log while starting the beacon node.

This PR  Mask out the wss credentials
New SUCCESS Log:
```
Feb-08 11:14:42.630 []                 info: Fetching weak subjectivity state from weakSubjectivityServerUrl=https://eth2-beacon-prater.infura.io

```

NEW ERROR Log, it still prints the full url string so that the user can validate and see what lodestar cli is picking as the url
```
Feb-08 11:14:15.437 []                error: Invalid weakSubjectivityServerUrl=(real url here..) message=Invalid URL
```

Closes #3703

### DIFF
--- a/packages/cli/src/cmds/beacon/initBeaconState.ts
+++ b/packages/cli/src/cmds/beacon/initBeaconState.ts
@@ -108,7 +108,7 @@ export async function initBeaconState(
       // Validate the weakSubjectivityServerUrl and only log the origin to mask the
       // username password credentials
       const wssUrl = new URL(args.weakSubjectivityServerUrl);
-      logger.info("Fetching weak subjectivity state from", {
+      logger.info("Fetching weak subjectivity state", {
         weakSubjectivityServerUrl: wssUrl.origin,
       });
     } catch (e) {
--- a/packages/cli/src/cmds/beacon/initBeaconState.ts
+++ b/packages/cli/src/cmds/beacon/initBeaconState.ts
@@ -104,7 +104,17 @@ export async function initBeaconState(
     if (!args.weakSubjectivityServerUrl) {
       throw Error(`Must set arg --weakSubjectivityServerUrl for network ${args.network}`);
     }
-    logger.info("Fetching weak subjectivity state", {weakSubjectivityServerUrl: args.weakSubjectivityServerUrl});
+    logger.info("Fetching weak subjectivity state,", {
+      // Mask out the wss credentials of the format:
+      //  1. "(http|https)://<alphanumeric>:<alphanumeric>@any.domain.baseurl" =>
+      //     "(http|https)://*****@any.domain.baseurl"
+      //  2. "<alphanumeric>:<alphanumeric>@any.domain.baseurl" =>
+      //     "*****@any.domain.baseurl"
+      weakSubjectivityServerUrl: args.weakSubjectivityServerUrl.replace(
+        /((?<=^)|(?<=\/\/))((\w)+:(\w)+)(?=@([\w\-.])+)/,
+        "*****"
+      ),
+    });
 
     const {wsState, wsCheckpoint} = await fetchWeakSubjectivityState(
       chainForkConfig,
--- a/packages/cli/src/cmds/beacon/initBeaconState.ts
+++ b/packages/cli/src/cmds/beacon/initBeaconState.ts
@@ -104,17 +104,17 @@ export async function initBeaconState(
     if (!args.weakSubjectivityServerUrl) {
       throw Error(`Must set arg --weakSubjectivityServerUrl for network ${args.network}`);
     }
-    logger.info("Fetching weak subjectivity state,", {
-      // Mask out the wss credentials of the format:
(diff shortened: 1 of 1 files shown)


packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -104,7 +104,17 @@ export async function initBeaconState(
...
+    logger.info("Fetching weak subjectivity state,", {
+      // Mask out the wss credentials of the format:
+      //  1. "(http|https)://<alphanumeric>:<alphanumeric>@any.domain.baseurl" =>
+      //     "(http|https)://*****@any.domain.baseurl"
+      //  2. "<alphanumeric>:<alphanumeric>@any.domain.baseurl" =>
+      //     "*****@any.domain.baseurl"
+      weakSubjectivityServerUrl: args.weakSubjectivityServerUrl.replace(
+        /((?<=^)|(?<=\/\/))((\w)+:(\w)+)(?=@([\w\-.])+)/,
+        "*****"
+      ),
```
dapplion: You can just use NodeJS's builtin URL module
```ts
import {URL} from "url"
console.log(new URL("https://user:pass@google.es/"))
```
Prints
```
URL {
  href: 'https://user:pass@google.es/',
  origin: 'https://google.es',
  protocol: 'https:',
  username: 'user',
  password: 'pass',
  host: 'google.es',
  hostname: 'google.es',
  port: '',
  pathname: '/',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```
and print origin

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -104,7 +104,17 @@ export async function initBeaconState(
...
-    logger.info("Fetching weak subjectivity state", {weakSubjectivityServerUrl: args.weakSubjectivityServerUrl});
+    try {
+      // Validate the weakSubjectivityServerUrl and only log the origin to mask the
+      // username password credentials
+      const wssUrl = new URL(args.weakSubjectivityServerUrl);
+      logger.info("Fetching weak subjectivity state from", {
+        weakSubjectivityServerUrl: wssUrl.origin,
+      });
+    } catch (e) {
+      logger.error("Invalid", {weakSubjectivityServerUrl: args.weakSubjectivityServerUrl}, e as Error);
```
g11tech, line 115: still prints out the full url incase an invalid url is passed, helps user in reconciling what he/she is passing and what lodestar is picking up.

dapplion:
Thinking about this, do we _really_ have to print the URL? I mean it's a CLI parameter the user can just see what has typed, what's the value on printing that?

What the problem with logging
```
info: Fetching weak subjectivity state
```
Also, our fetcher could easily print the URL to logs anyway, we should test that

g11tech:
> Thinking about this, do we _really_ have to print the URL? I mean it's a CLI parameter the user can just see what has typed, what's the value on printing that?
> 

I find a bit of value as it looks more prudent to log the url, else just logging `info: Fetching weak subjectivity state` works too, as we do need to prompt user because it could easily take 20-30 seconds for the wss state to be fetched. so user shouldn't feel that lodestar is stuck.

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -104,7 +104,17 @@ export async function initBeaconState(
     if (!args.weakSubjectivityServerUrl) {
       throw Error(`Must set arg --weakSubjectivityServerUrl for network ${args.network}`);
     }
-    logger.info("Fetching weak subjectivity state", {weakSubjectivityServerUrl: args.weakSubjectivityServerUrl});
+    try {
+      // Validate the weakSubjectivityServerUrl and only log the origin to mask the
+      // username password credentials
+      const wssUrl = new URL(args.weakSubjectivityServerUrl);
+      logger.info("Fetching weak subjectivity state from", {
```
wemeetagain:
```suggestion
      logger.info("Fetching weak subjectivity state", {
```

g11tech: @wemeetagain I added *from* looking at the final generated log: `info: Fetching weak subjectivity state from weakSubjectivityServerUrl=https://eth2-beacon-` which reads good with from. Let me know what you think. :slightly_smiling_face:

dapplion: You shouldn't assume how the log data renders, we could change that in the future

dapplion (approved):

PR #3734, opened 2022-02-08
Disable arm64 docker build

### LINKED ISSUE #51: Publish arm64 bindings

See https://github.com/ChainSafe/lodestar/issues/3692

Build and publish arm64 bindings (for linux at least)

wemeetagain:
**Motivation**

See https://github.com/ChainSafe/lodestar/runs/5113003430?check_suite_focus=true
and
https://github.com/ChainSafe/blst-ts/issues/51

After #3698 , the docker build is breaking because blst isn't prebuilt for arm64, and the final docker layer doesn't have build tools to build blst itself.

We should first resolve chainsafe/blst-ts#51, then revert this when we're ready.

**Description**

Quick way to resolve the currently breaking docker build, just don't build the arm64 image.

### DIFF
--- a/.github/workflows/release-nightly.yml
+++ b/.github/workflows/release-nightly.yml
@@ -102,5 +102,5 @@ jobs:
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


dapplion:
@wemeetagain For next attempts you can modify the github actions workflow to also trigger on a specific branch temporarily to prevent polluting master

dapplion (approved):

Issue #3732, opened 2022-02-08
Gossip Block Processed delay increased in v0.34.0


twoeths:
**Describe the bug**

Right now with v0.34.0 in hetzner-test1
<img width="807" alt="Screen Shot 2022-02-08 at 14 46 17" src="https://user-images.githubusercontent.com/10568965/152941163-0b7b97ed-7d5d-44a3-bb8e-e4817b4bf465.png">

**Expected behavior**

This is from v0.33.0
<img width="811" alt="Screen Shot 2022-02-08 at 14 48 11" src="https://user-images.githubusercontent.com/10568965/152941447-e996ad73-478d-46e7-a9b7-b4c6a953e423.png">

twoeths:
one issue is BackfillSync query BlockArchive db a lot which takes 18% of cpu time
<img width="1127" alt="Screen Shot 2022-02-08 at 15 05 15" src="https://user-images.githubusercontent.com/10568965/152944111-727066cb-0133-4f5c-b83d-f3e5721927bb.png">

Backfill Sync also uses 3% of cpu time for `blockToHeader()` util
<img width="855" alt="Screen Shot 2022-02-08 at 15 07 31" src="https://user-images.githubusercontent.com/10568965/152944413-2184d111-760e-407a-a256-e08eef7733d5.png">

[0207_hetzner_test1_backfill_sync.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/8021875/0207_hetzner_test1_backfill_sync.cpuprofile.zip)

twoeths:
the other issue is `array.shift()` in `JobItem.push()`, which takes 17% cpu time in another profile

<img width="1042" alt="Screen Shot 2022-02-09 at 13 51 38" src="https://user-images.githubusercontent.com/10568965/153141700-24bef54e-b2a2-4fdf-8082-2beccaf51b01.png">

<img width="1137" alt="Screen Shot 2022-02-09 at 14 26 15" src="https://user-images.githubusercontent.com/10568965/153141936-a7947548-2557-4eb8-a804-0602eb10585c.png">

[0209_hetzner_test1_processRpc.cpuprofile.zip]
(https://github.com/ChainSafe/lodestar/files/8030365/0209_hetzner_test1_processRpc.cpuprofile.zip)

@dapplion could we use a linked list in JobItemQueue?

twoeths:
part of the issue is there are a lot of late gossip blocks recently in prater, even when I try with v0.33.0

[...]

PR #3738, opened 2022-02-10
Use LinkedList for JobItemQueue

twoeths:
**Motivation**

As shown in #3732, it takes up to 17% of cpu time just to do the `array.shift()`

**Description**

+ Since we don't need random access in the `jobs` array, write a LinkedList to improve the performance
+ The benchmark shows that it's way faster to use LinkedList to do shift / push /pop
```
LinkedList vs Regular Array
    ✓ array of 16000 items push then shift                                276390.5 ops/s    3.618070 us/op   x1.140       1017 runs   5.29 s
    ✓ LinkedList of 16000 items push then shift                        7.052684e+7 ops/s    14.17900 ns/op   x0.955       8744 runs   1.73 s
    ✓ array of 16000 items push then pop                                   6606896 ops/s    151.3570 ns/op   x0.947       1720 runs   3.04 s
    ✓ LinkedList of 16000 items push then pop                          7.412898e+7 ops/s    13.49000 ns/op   x1.006       8163 runs   1.82 s
    ✓ array of 24000 items push then shift                                167710.8 ops/s    5.962646 us/op   x0.994        389 runs   3.62 s
    ✓ LinkedList of 24000 items push then shift                        7.102777e+7 ops/s    14.07900 ns/op   x0.914       2824 runs   1.16 s
    ✓ array of 24000 items push then pop                                   4541347 ops/s    220.1990 ns/op   x1.180        525 runs   3.18 s
    ✓ LinkedList of 24000 items push then pop                          7.258474e+7 ops/s    13.77700 ns/op   x1.023       4409 runs   1.65 s
```
part of #3732

### DIFF
--- a/packages/lodestar/src/util/array.ts
+++ b/packages/lodestar/src/util/array.ts
@@ -112,4 +112,17 @@ export class LinkedList<T> {
 
     return arr;
   }
+
+  map<U>(fn: (t: T) => U): U[] {
+    let node = this.head;
+    if (!node) return [];
+
+    const arr: U[] = [];
+    while (node) {
+      arr.push(fn(node.data));
+      node = node.next;
+    }
+
+    return arr;
+  }
 }
--- a/packages/lodestar/src/util/array.ts
+++ b/packages/lodestar/src/util/array.ts
@@ -10,3 +10,106 @@ export function findLastIndex<T>(array: T[], predicate: (value: T) => boolean): 
   }
   return -1;
 }
+
+/**
+ * The node for LinkedList below
+ */
+class Node<T> {
+  data: T;
+  next: Node<T> | null = null;
+  prev: Node<T> | null = null;
+
+  constructor(data: T) {
+    this.data = data;
+  }
+}
+
+/**
+ * We want to use this if we only need push/pop/shift method
+ * without random access.
+ * The shift() method should be cheaper than regular array.
+ */
+export class LinkedList<T> {
+  private _length: number;
+  private head: Node<T> | null;
+  private tail: Node<T> | null;
+
+  constructor() {
+    this._length = 0;
+    this.head = null;
+    this.tail = null;
+  }
+
+  get length(): number {
+    return this._length;
+  }
+
+  push(data: T): void {
+    if (this._length === 0) {
+      this.tail = this.head = new Node(data);
+      this._length++;
+      return;
+    }
+
+    if (!this.head || !this.tail) {
+      // should not happen
+      throw Error("No head or tail");
+    }
+
+    const newTail = new Node(data);
+    this.tail.next = newTail;
+    newTail.prev = this.tail;
+    this.tail = newTail;
+    this._length++;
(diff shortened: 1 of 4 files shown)


dapplion:
Looks overall! @tuyennhv have you been able to test this in a beacon node?

twoeths:
> Looks overall! @tuyennhv have you been able to test this in a beacon node?

I've run it 7h in hetzner-test1 http://contabo-metrics:3000/d/lodestar/lodestar?orgId=1&var-ByJob=instance%7C%3D%7Chetzner-test1&from=now-24h&to=now&refresh=30s

at least we can maintain good number of peers there, profiles show that we don't take time for that `this.jobs.shift()` code anymore

some gossip blocks still come late, I suppose there are some validators in prater network publish late blocks (tried with v0.33.0 and it's the same)

dapplion:
@tuyennhv I would like to understand better first how often does master calls .shift(). Looking at your benchmarks the most expensive .shift() takes ~6us, so to take 17% of total CPU time

```
1e6 us/s * 0.17 / 5.9 us/op = 28813 ops/s
```

This number is absurdly high, maybe there's some other issue on top of the .shift inefficiency?

twoeths:
> @tuyennhv I would like to understand better first how often does master calls .shift(). Looking at your benchmarks the most expensive .shift() takes ~6us, so to take 17% of total CPU time
> 
> ```
> 1e6 us/s * 0.17 / 5.9 us/op = 28813 ops/s
> ```
> 
> This number is absurdly high, maybe there's some other issue on top of the .shift inefficiency?

@dapplion  I'm not sure how the profiler calculate the cpu percentage, maybe it's calculate against the total time for nodejs (not the total time to run profiler). From the profile, total time is 275100ms, time for `jobs.shift` is 7066ms so the ratio per total time is 2.57% (not 17%)

```
1e6 us/s * 0.0257 / 5.9 us/ops = 4353 ops/s
```

given 275_000 active validators in prater, attestation messages per second would be: `275_000/32/12 = 716 msg/s`

`4353 / 716 ~ 6` => due to gossip message redundancy, for each attestation there is approximately 6 duplicate instances. (We did not calculate gossip blocks, aggregated attestations, sync committee messages so the duplicate messages should be less than that)

packages/lodestar/src/util/array.ts
```diff
@@ -10,3 +10,119 @@ export function findLastIndex<T>(array: T[], predicate: (value: T) => boolean):
...
+    } else {
+      this.head = oldHead.next;
+      if (this.head) this.head.prev = null;
+      oldHead.prev = oldHead.next = null;
+    }
+
+    return oldHead.data;
+  }
+
+  prune(): void {
```
wemeetagain: Can we name this `clear`?

wemeetagain (changes requested):
LGTM, minor comment

wemeetagain (approved):

PR #3739, opened 2022-02-10
Show github branch in grafana dashboard if using ansible with lodestar_github_tag

twoeths:
**Motivation**

+ Using ansible with "lodestar_github_tag", it always show the current local version, which is "v0.35.0 (npm)"
+ The code assumes if no "semver", there is no git but actually there is git and we can get github branch in this case

**Description**
+ Show github branch

Screen Shot 2022-02-10 at 14 53 07

+ Note that with feature branch, we can't get "numCommit" but it should be fine with master

### DIFF
--- a/packages/cli/src/util/gitData/index.ts
+++ b/packages/cli/src/util/gitData/index.ts
@@ -68,7 +68,7 @@ export function readLodestarGitData(): GitData {
     }
 
     return {
-      semver: gitData?.semver || "N/A",
+      semver: gitData?.semver,
       branch: gitData?.branch || "N/A",
       commit: gitData?.commit || "N/A",
       numCommits: gitData?.numCommits || "",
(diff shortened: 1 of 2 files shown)


g11tech (approved):

dapplion (approved):

dapplion:
I have not tested it, but does this work fine with a stable release?

twoeths:
> I have not tested it, but does this work fine with a stable release?

I don't know how to test it until a release but this command gives exactly "0" in my local with a tag branch:
```
git rev-list v0.34.0-beta.0..HEAD --count
0
```

so it should return "v0.34.0-beta.0" in https://github.com/ChainSafe/lodestar/pull/3739/files#diff-7d989bc091ca0cdf47b26b676c0b4aa43669bcfd8cecc4a10c77524ef28d08f3L42

wemeetagain (approved):

PR #3740, opened 2022-02-10
Cleanup discontinued pyrmont network

g11tech:
Cleanup the pyrmont as it is now discontinued.


twoeths (approved):

packages/light-client/src/networks.ts
```diff
@@ -3,7 +3,7 @@ import {RootHex} from "@chainsafe/lodestar-types";
 enum NetworkName {
   mainnet = "mainnet",
   prater = "prater",
-  pyrmont = "pyrmont",
+  kintsugi = "kintsugi",
```
wemeetagain, line 6: did you mean to add this here?

g11tech: yes, but i defer it to you if it makes sense here as noticed kintsugi was missing (soon to be replaced by kiln but saw one PR from you regarding a lightclient fix for kintsugi) :slightly_smiling_face:, 

but I guess since Kiln execution engine  api changes PR is merge, kintsugi target is no more compatible with current master, but would like to retain so that when kiln config gets updated, this one will get updated too.

wemeetagain (approved):

PR #3742, opened 2022-02-10
Prevent circular dependencies with eslint

dapplion:
**Motivation**

'import/no-cycle' is expensive computationally, consider restricting its `maxdepth`

See https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/no-cycle.md#maxdepth

A lint run in master currently takes `Done in 121.30s.` in Github Actions (from https://github.com/ChainSafe/lodestar/runs/5137987474?check_suite_focus=true#step:13:38)

In this PR for a failed run it took `in 137.5s` so the extra time is okay :+1: 

**Description**

Prevent circular dependencies with eslint


dapplion:
Closing for https://github.com/ChainSafe/lodestar/issues/3929

PR #3743, opened 2022-02-10
Re-org state transition caches

dapplion:
**Motivation**

This PR lays down some ground work in anticipation to bring new SSZ version changes.

We have the opportunity to drop some of the caches, so I'm re-orging the existing in a way that makes more sense and is abstracted from the code that actually processes the state.

**Description**

Move to cache 2 main entities:

- EpochContext is the parent object of: Any data-structures not part of the spec'ed BeaconState, necessary to only compute data once, must be kept at all times through an epoch.
- EpochProcess is the parent object of: Any data-structures not part of the spec'ed BeaconState, necessary to only compute data once, only necessary for epoch processing, can be disposed immediatelly.

Other caches to be deprecated
- CachedBeaconState: balanceList, EpochParticipation, inactivityScores

It also clean-up unnecessary utils and move the used ones to `src/utils`


wemeetagain (approved):
LGTM

twoeths (approved):

Issue #3680, opened 2022-01-29
Add state cloneCount metric


dapplion:
**Is your feature request related to a problem? Please describe.**

New SSZ version uses a strategy of state cache transfer which provides great performance improvements. However, after cloning a state, subsequent clones will get an empty cache that must be re-populated.

Then it's important to know how many times a single state is cloned now, to estimate the effectiveness of this cache transfer strategy.

From https://github.com/ChainSafe/lodestar/pull/3760

**Describe the solution you'd like**

- Add a metric to get the number of clone() calls at a single state object. Every time when calling clone() add 1 to a counter of `clonedCount`, and register a histogram metric of cloneCount at cloned(). If states are used exactly once then the average will be 0. If states are clone multiple times the average will be greater than zero

PR #3760, opened 2022-02-15
Bump SSZ to v0.9.0

### LINKED ISSUE #2408: Make sure to not get unwanted subTrees
Even though currently SubTree instances use a WeakRef it would be nice to review our code and ensure that we don't get a SubTree for unwanted items that are likely never gonna change, such as withdrawal credentials or roots.

### LINKED ISSUE #2419: Better .valueOf()
**Describe the issue**

Using `.valueOf() as SomeType` is not type-safe at all and can become dangerous an unmaintainable. 

We are using it for a very specific purpose which is to drop the tree and get the actual internal value without wrappers. This meaning is not easy to get at first glance.

**Proposed solution**

Create a helper function that clearly shows what it does that AND is type safe.

```ts
function dropTree<T>(treeBackedValue: T): T extends ByteVector ? Uint8Array : T {
  return treeBackedValue.valueOf()
}
```

With this code would read like

```ts
const randaoReveal = dropTree(block.body.randaoReveal)
```

instead of 

```ts
const randaoReveal = block.body.randaoReveal.valueOf() as Uint8Array;
```

----

Related to https://github.com/ChainSafe/lodestar/issues/2408

### LINKED ISSUE #2859: Don't deserialize AggregateAndProof to tree
**Describe the bug**

Taking a CPU profile in Prater on a node with 1 validator 10.36% of CPU time is spent on tree_deserializeFromBytes.

I think we don't really need to deserialize to tree backed. We can manually cache the roots that we need.

The lifetime of an AggregateAndProof is relatively simple:

1. Gossip
2. Validation queue
3. Gossip validation
  3A. Verify AttestationData signature. Compute the AttestationData root and keep it.
  3B. Verify AggregateAndProof signature. Recycle AttestationData root to compute the signing root.
4. Op pool: Reuse the already computed AttestationData root
  4A. Block production
  4B. Pruned

A struct or even binary backed takes less memory, it's cheaper to deserialize and it's cheaper to access it's properties.

### LINKED ISSUE #2966: Improve updating balances to tree

In altair when processing a block, we update balances multiple time and every time we update BeaconState tree root due to the hook, this is inefficient.

Update balances on a cloned balances tree and once it's done, we escalate the change to BeaconState tree root

In `CachedBalanceList`

```ts
constructor(type: BasicListType<List<number>>, tree: Tree) {
    this.type = type;
    // this.persistentTree always link to the parent tree, this.tree may belong to the parent tree or a cloned tree
    this.tree = tree;
    this.persistentTree = tree;
  }
  
  // all methods should work on this.tree

  asTransient(): void {
    // from this time all increaseBalances should work on a cloned tree, this will not update the BeaconState parent tree
    this.tree = this.persistentTree.clone();
    // the persistent-ts list, may remove later
    this.persistent.asTransient();
  }

  asPersistent: void {
    // updating the root of persistentTree would escalate the change to BeaconState tree root due to the hook
    this.persistentTree.root = this.tree.root;
    // the persistent-ts list, may remove later
    this.tree = this.persistentTree;
  }

  // may need to update Proxy
```

### LINKED ISSUE #3080: altair processAttestations performance issue
**Describe the bug**

Looking at the metrics for our altair nodes processing a block takes way longer (300ms) than what the performance tests say (100ms).

Taking a profile shows that all the time is spent on processing attestations

Screenshot from 2021-09-03 18-40-18

**Expected behavior**

Should not take so long

------

Note: this CPU profile was taken in an overloaded contabo node.

Running the altair block performance benchmarks in my laptop do

```
  altair processBlock
    ✓ altair processBlock - 250000 vs - 7PWei normalcase                  10.01820 ops/s    99.81831 ms/op        -         33 runs   17.2 s
    ✓ altair processBlock - 250000 vs - 7PWei worstcase                   4.261845 ops/s    234.6401 ms/op        -         26 runs   7.50 s
```

While running the same in that contabo's host

```
  altair processBlock
    ✓ altair processBlock - 250000 vs - 7PWei normalcase                  7.083636 ops/s    141.1704 ms/op   x0.601         14 runs   20.8 s
    ✓ altair processBlock - 250000 vs - 7PWei worstcase                   2.356299 ops/s    424.3943 ms/op   x0.629         24 runs   12.1 s
```

And Github CI does

```
altair processBlock - 250000 vs - 7PWei normalcase | 103.91 ms/op | 97.787 ms/op | 1.06
altair processBlock - 250000 vs - 7PWei worstcase  | 261.36 ms/op | 265.06 ms/op | 0.99
```

### LINKED ISSUE #3082: processEth1Data performance issue
**Describe the bug**

processEth1Data takes more time than it should (5-6ms, 25% of phase0 block processing times). Looks like iteration takes signficant time + it has to convert the root of each entry from hashObject to Uint8Array and then compare. A bit stupid when it could just use the hashObject numbers

Screenshot from 2021-09-03 19-29-31

Screenshot from 2021-09-03 19-29-52

----

> From duplicate issue https://github.com/ChainSafe/lodestar/issues/2453

In Prater our node spends 1400 ms processing Eth1Data. This happens on every block when there's no majority vote.

Screenshot from 2021-05-01 15-20-47

Same cost every block

Screenshot from 2021-05-01 15-22-05

detail of the stack

Screenshot from 2021-05-01 15-23-11

- [ ] Add a dedicated performance test
- [ ] Optimize

dapplion:
**Draft note**

This PR installs a placeholder version of SSZ with this awesome tool https://gitpkg.vercel.app/. This PR is open now to proof that Lodestar can run with this new SSZ version before releasing it.

https://github.com/ChainSafe/lodestar/blob/d7197294d5f312bc64625e95e19c9a7cce88016f/packages/api/package.json#L43-L44

**Motivation**

Lodestar has implemented many optimizations around our data-structures. This PR together with https://github.com/ChainSafe/ssz/pull/223 upstreams them to ssz.

See https://github.com/ChainSafe/ssz/pull/223 for full rationale

**Description**

- Bump SSZ to upcoming major version
- Drop flat caches

**TODO** After merging

- [x] Make sure to remove installation from custom repositories off ssz and persistent merkle tree
- [x] Stop using `ssz.Root.equals`, use local bytesEqual utilities https://github.com/ChainSafe/lodestar/issues/3996
- [x] Optimize hashing of struct values, 30-40% potential improvements to be gained. Matters when hashing the state for the first time. https://github.com/ChainSafe/lodestar/issues/3997 https://github.com/ChainSafe/ssz/issues/245
- [ ] Add test for mutability of effectiveBalanceIncrements
- [x] Add state clone metrics, track the transfer of caches vs instantiating fresh state caches. Add a metric to get the number of clone() calls at a single state object. Every time when calling clone() add 1 to a counter of clonedCount, and register a histogram metric of cloneCount at cloned(). If states are used exactly once then the average will be 0. If states are clone multiple times the average will be greater than zero https://github.com/ChainSafe/lodestar/issues/3680
- [x] Review if the RootCache for attestation processing is necessary https://github.com/ChainSafe/lodestar/issues/3998
- [x] Review if all the data in epochCtx shufflings is necessary https://github.com/ChainSafe/lodestar/issues/3999
- [x] Add coverage tracker in SSZ https://github.com/ChainSafe/ssz/issues/252

Closes https://github.com/ChainSafe/lodestar/issues/2859
Closes https://github.com/ChainSafe/lodestar/issues/2419
Closes https://github.com/ChainSafe/lodestar/issues/2408
Closes https://github.com/ChainSafe/lodestar/issues/3082
Closes https://github.com/ChainSafe/lodestar/issues/2966
Closes https://github.com/ChainSafe/lodestar/issues/3080


packages/beacon-state-transition/src/cache/stateCache.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  function clone(this: T & BeaconStateCache): T & BeaconStateCache {
+    const viewDUCloned = viewDUClone();
+    return getCachedBeaconState(viewDUCloned, {
+      config: this.config,
+      epochCtx: this.epochCtx.clone(),
+    }) as T & BeaconStateCache;
+  }
+
+  cachedState.clone = clone as typeof viewDUClone;
```
dapplion, line 148: This is a specially critical an awkward part of the state cache. It re-writes the clone method set by the TreeViewDU. Would be better to use [Object.defineProperty()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty)?

packages/beacon-state-transition/src/allForks/block/processEth1Data.ts
```diff
@@ -63,8 +63,11 @@ export function becomesNewEth1Data(
   }
 }
 
-function isEqualEth1DataView(eth1DataA: TreeBacked<phase0.Eth1Data>, eth1DataB: TreeBacked<phase0.Eth1Data>): boolean {
-  return isEqualNode(eth1DataA.tree.rootNode, eth1DataB.tree.rootNode);
+function isEqualEth1DataView(
+  eth1DataA: CompositeViewDU<typeof ssz.phase0.Eth1Data>,
+  eth1DataB: CompositeViewDU<typeof ssz.phase0.Eth1Data>
+): boolean {
+  return isEqualNode(eth1DataA.node, eth1DataB.node);
```
twoeths, line 70: should we escalate this function to ssz v2 so that it can be reused?

dapplion:
I was able to sync with this branch from genesis to head in mainnet @tuyennhv 

```
Feb 27 05:22:53 hetzner-test1 beacon_run.sh[1663142]: Feb-27 05:22:53.001[]                 info: Synced - slot: 3259312 - head: 3259312 0x3a46…e64e - finalized: 0x8931…ed7b:101851 - peers: 19
Feb 27 05:23:05 hetzner-test1 beacon_run.sh[1663142]: Feb-27 05:23:05.001[]                 info: Synced - slot: 3259313 - head: 3259313 0x015a…8c09 - finalized: 0x8931…ed7b:101851 - peers: 19
Feb 27 05:23:17 hetzner-test1 beacon_run.sh[1663142]: Feb-27 05:23:17.000[]                 info: Synced - slot: 3259314 - head: 3259314 0xc427…ce1a - finalized: 0x8931…ed7b:101851 - peers: 18
Feb 27 05:23:28 hetzner-test1 beacon_run.sh[1663142]: Feb-27 05:23:28.999[]                 info: Synced - slot: 3259315 - head: 3259315 0x0530…8e1b - finalized: 0x8931…ed7b:101851 - peers: 19
```

dapplion (commented):
:heavy_check_mark: Meta-approve! Since I originally opened the PR, can't approve

----

TODOs for upcoming PRs

- https://github.com/ChainSafe/lodestar/pull/3760/files#r810760677

wemeetagain (dismissed):

yarn.lock
```diff
@@ -505,6 +525,15 @@
     buffer-from "^1.1.1"
     snappy "^6.3.5"
 
+"@chainsafe/ssz@^0.8.20":
```
dapplion: Warning, this should not be in the yarn.lock. Who is depending in the old versions?

wemeetagain (approved):

PR #3768, opened 2022-02-17
Deduplicate invalidSszObject files

twoeths:
**Motivation**

Right now same invalid state root objects are stored in different files with different date suffixes:
```
...
state_2368896_invalid_state_root_preState_0x4dc200dc344c587fcdb0b2bd6bc8c1dd67249c0f4563345ecab67b6b0830d333_1644935390189.ssz
state_2368896_invalid_state_root_preState_0x4dc200dc344c587fcdb0b2bd6bc8c1dd67249c0f4563345ecab67b6b0830d333_1644935499837.ssz
state_2368896_invalid_state_root_preState_0x4dc200dc344c587fcdb0b2bd6bc8c1dd67249c0f4563345ecab67b6b0830d333_1644935608920.ssz
...
```

**Description**

+ Remove the date suffixes as part of file name
+ For invalidStateRoot error, use block slot in file name to group all files together so that it's easier to investigate later
+ Fix the log typo

### DIFF
--- a/packages/lodestar/src/chain/chain.ts
+++ b/packages/lodestar/src/chain/chain.ts
@@ -283,8 +283,12 @@ export class BeaconChain implements IBeaconChain {
     if (!fs.existsSync(byDate)) {
       fs.mkdirSync(byDate, {recursive: true});
     }
-    const fileName = `${byDate}/${type}_${suffix}_${Date.now()}.ssz`;
-    fs.writeFileSync(fileName, bytes);
+    const fileName = `${byDate}/${type}_${suffix}.ssz`;
+    // as of Feb 17 2022 there are a lot of duplicate files stored with different date suffixes
+    // remove date suffixes in file name, and check duplicate to avoid redundant persistence
+    if (!fs.existsSync(fileName)) {
+      fs.writeFileSync(fileName, bytes);
+    }
     return fileName;
   }
 }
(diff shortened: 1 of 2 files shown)


dapplion (approved):

packages/lodestar/src/chain/chain.ts
```diff
@@ -283,8 +283,12 @@ export class BeaconChain implements IBeaconChain {
     if (!fs.existsSync(byDate)) {
       fs.mkdirSync(byDate, {recursive: true});
     }
-    const fileName = `${byDate}/${type}_${suffix}_${Date.now()}.ssz`;
-    fs.writeFileSync(fileName, bytes);
+    const fileName = `${byDate}/${type}_${suffix}.ssz`;
```
dapplion, line 286: Do they need to be organized in directories by date? You can use the linux time to see what file is at what date. With this files will still be duplicated 1 per day

twoeths: grouping by date would help investigate issues better imo, we can for sure use linux time to group/count files but it's just more convenient to go to sub-folder and see how many inValidSSZObject files there

yeah there's a chance files are stored end of day `n` and beginning of day `n + 1`, we'll see if it's a lot and cause any issues

PR #3770, opened 2022-02-18
Add SignedAggregateAndProof root to gossip validation accept metadata

twoeths:
**Motivation**

Different nodes show different subsets of invalid signature SignedAggregateAndProof messages

**Description**

+ Add SignedAggregateAndProof pretty bytes root to the below log:
```
Dec-24 14:11:13.820 [NETWORK]         ^[[34mdebug^[[39m: gossip - beacon_aggregate_and_proof - accept slot=198        5152, index=39
```

+ I'd like to do same investigation for `Attestation` and `sync_committee_contribution_and_proof` but worry that additional `hashTreeRoot()` calls would make the performance worse (ssz v2 could help?). Printing root for SignedAggregateAndProof is not an issue because it's a TreeBacked.

part of #3555


twoeths:
discussed with @dapplion , we don't want this on master as sszv2 only caches root for the SignedAggregateAndProof.message

PR #3771, opened 2022-02-18
Don't error on LightClientServer finalizedHeader not available

### LINKED ISSUE #3495: Error: finalityHeader not available
**Describe the bug**

During initial sync, the light-client server errors with

```
Error: finalityHeader not available 0xbfd6e68ea9ab8be2665a8662743e82ee0a72f846182fdf6232f9ff1a4bac9854
    at LightClientServer.getFinalizedHeader (/usr/app/node_modules/@chainsafe/lodestar/src/chain/lightClient/index.ts:511:13)
    at LightClientServer.maybeStoreNewBestPartialUpdate (/usr/app/node_modules/@chainsafe/lodestar/src/chain/lightClient/index.ts:474:28)
    at LightClientServer.onSyncAggregate (/usr/app/node_modules/@chainsafe/lodestar/src/chain/lightClient/index.ts:453:5)
```

**Expected behavior**

- The light-client server should not prepare updates during sync
- finalityHeader may not be available and it's not an issue, it's just that the node hasn't seen that finality header

dapplion:
**Motivation**

During sync finalizedHeader will not be available since the LightClientServer has never seen that block.

**Description**

Don't error on LightClientServer finalizedHeader not available.

If update if finalized retrieve the previously stored header from DB. Only checkpoint candidates are stored, and not all headers are guaranteed to be available
- If finalizedHeader is available (should be most times) create a finalized update
- If finalizedHeader is not available (happens on startup) create a non-finalized update

Closes https://github.com/ChainSafe/lodestar/issues/3495

### DIFF
--- a/packages/lodestar/src/chain/lightClient/index.ts
+++ b/packages/lodestar/src/chain/lightClient/index.ts
@@ -9,6 +9,7 @@ import {ILogger} from "@chainsafe/lodestar-utils";
 import {routes} from "@chainsafe/lodestar-api";
 import {BitVector, toHexString} from "@chainsafe/ssz";
 import {IBeaconDb} from "../../db";
+import {IMetrics} from "../../metrics";
 import {MapDef, pruneSetToMax} from "../../util/map";
 import {ChainEvent, ChainEventEmitter} from "../emitter";
 import {
@@ -40,17 +41,13 @@ type SyncAttestedData = {
     }
 );
 
-type GenesisData = {
-  genesisTime: number;
-  genesisValidatorsRoot: Uint8Array;
-};
-
-interface ILightClientIniterModules {
+type LightClientServerModules = {
   config: IChainForkConfig;
   db: IBeaconDb;
+  metrics: IMetrics | null;
   emitter: ChainEventEmitter;
   logger: ILogger;
-}
+};
 
 const MAX_CACHED_FINALIZED_HEADERS = 3;
 const MAX_PREV_HEAD_DATA = 32;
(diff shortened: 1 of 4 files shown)


dapplion:
Tested locally against Prater and I see no more errors, logs still appear but are expected and in debug level

g11tech (dismissed):

packages/lodestar/src/chain/lightClient/index.ts
```diff
@@ -468,23 +468,36 @@ export class LightClientServer {
...
+      // If update if finalized retrieve the previously stored header from DB.
+      // Only checkpoint candidates are stored, and not all headers are guaranteed to be available
+      const finalizedCheckpointRoot = attestedData.finalizedCheckpoint.root as Uint8Array;
+      const finalizedHeader = await this.getFinalizedHeader(finalizedCheckpointRoot);
+      if (finalizedHeader) {
+        // If finalizedHeader is available (should be most times) create a finalized update
+        newPartialUpdate = {...attestedData, finalizedHeader, syncAggregate};
+      } else {
+        // If finalizedHeader is not available (happens on startup) create a non-finalized update
+        newPartialUpdate = {...attestedData, isFinalized: false, syncAggregate};
```
twoeths, line 483: if it's not at startup time, is it an issue of missing finalizedHeader? Should we add new metric to track it `missingFinalizedHeader`?

dapplion: Just added a metric to track that in https://github.com/ChainSafe/lodestar/pull/3771/commits/9ef5e36f6dfa476a68b2b07db320b44ee9548f77

Yes it can be a problem if it's not available latter, but it's a situation that can happen depending on network conditions. We should monitor this new metric from now on

twoeths (approved):

PR #3774, opened 2022-02-18
BackfillSync boundary condition fixes

### LINKED ISSUE #3765: Investigate BackfillSync hetzner-test0 issue of "Re-doing" the previous backfill from DB instead from backfilledRanges
@dapplion pointed out an issue that hetzner nodes were facing, on restart the backfill sync started backfilling all previous backfilled blocks from db instead of it using a backfilled Range to skip back. 
This caused a huge disk resource consumption paired with hashTreeRoot severly degrading the performance of the node. A temp fix was send to disable backfill on default.

This issue is to investigate why the backfill range was not found by analyzing the logs from the affected machines.

g11tech:
**Motivation**
As described here: https://github.com/ChainSafe/lodestar/issues/3765#issuecomment-1041500809, the backfill had an issue where it was re-doing previous backfill, because it erroneously deleted a previous backfillRange. 

tl-dr: backfill saved completed backfilled range using starting `anchorState.slot` i.e `2342432,0` (i.e. range from 2342432 to 0), but  this was a skipped slot, so the actual  `anchorSlot = anchorState.latestBlockHeader.slot` lay at 2342431. 

This (2342431) is the same value in `anchorStateLatestBlockSlot` that  the forward sync correctly uses to update `finalizedBlockFC.slot, this.chain.anchorStateLatestBlockSlot`  as a backfill entry and deletes all previous "front" written entries `> anchorStateLatestBlockSlot`.  So in this case wrote `2351040,2342431` as backfill entry and deleted any entry `> 2342431`, which lead to deletion of `2342432,0`:
https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/chain/archiver/index.ts#L100:L113

This PR fixes the assignment of `backfillStartFromSlot` which is used to save the backfill's progress. This correction will prevent its deletion in the "forward" sync's update of the backfilledRange.
It also fixes the point where a wsCheckpoint slot should be found as epoch for a wsCheckpoint could be floor of the checkpoint block's slot /SLOTS_PER_EPOCH

**Description**

Closes #3765

### DIFF
--- a/packages/lodestar/src/chain/archiver/index.ts
+++ b/packages/lodestar/src/chain/archiver/index.ts
@@ -105,15 +105,21 @@ export class Archiver {
         // touching backfill sync process sequence which are at
         // <=anchorStateLatestBlockSlot i.e. clear >anchorStateLatestBlockSlot
         // and < currentSlot
-        const filteredSeqs = await this.db.backfilledRanges.keys({
+        const filteredSeqs = await this.db.backfilledRanges.entries({
           gt: this.chain.anchorStateLatestBlockSlot,
           lt: finalizedBlockFC.slot,
         });
-        await this.db.backfilledRanges.batchDelete(filteredSeqs);
-        this.logger.debug("Updated backfilledRanges", {
+        this.logger.debug("updated backfilledRanges", {
           key: finalizedBlockFC.slot,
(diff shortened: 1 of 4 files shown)


dapplion (approved):
LGTM!

PR #3781, opened 2022-02-22
Apply peer action when there are gossip validation errors

### LINKED ISSUE #3772: Apply peer action when there are gossip validation errors
**Describe the bug**

Hetzner-s2 shows a lot of rejected ContributionAndProof gossip messages but other nodes show nothing
Screen Shot 2022-02-18 at 14 28 24

The `lodestar_gossip_validation_accept_total` metric is the same for all nodes.

```
ls 2022-02-16 | grep "contribution" |  wc -l
174
 ls 2022-02-15 | grep "contribution" |  wc -l
125
```

**Expected behavior**

+ It should be consistent across nodes

twoeths:
**Motivation**

+ if a node keeps receiving invalid messages from another node, right now it only decreases gossip score of peer, we also should decrease score at lodestar side, and in the end disconnect the peer

**Description**

+ Decay score in PeerManager hearbeat (we missed that)
+ GossipError now has PeerAction, based on that we'll penalizing peer at lodestar side when errors happen, refer to lighthouse implementation https://github.com/sigp/lighthouse/blob/c3a793fd73a3b11b130b82032904d39c952869e4/beacon_node/network/src/beacon_processor/worker/gossip_methods.rs#L1339

Closes #3772

### DIFF
--- a/packages/lodestar/src/network/gossip/validation/index.ts
+++ b/packages/lodestar/src/network/gossip/validation/index.ts
@@ -18,7 +18,7 @@ import {decodeMessageData, UncompressCache} from "../encoding";
 import {createValidationQueues} from "./queue";
 import {DEFAULT_ENCODING} from "../constants";
 import {getGossipAcceptMetadataByType, GetGossipAcceptMetadataFn} from "./onAccept";
-import {IPeerRpcScoreStore} from "../../peers/score";
+import {IPeerRpcScoreStore, PeerAction} from "../../peers/score";
 import PeerId from "peer-id";
 
 type ValidatorFnModules = {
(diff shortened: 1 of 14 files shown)


twoeths:
I leave the test node for 24h, Gossip peer score is increased, number of mesh peers are increased as well

Screen Shot 2022-02-23 at 10 08 54

packages/lodestar/src/network/gossip/validation/index.ts
```diff
@@ -95,7 +99,7 @@ function getGossipValidatorFn<K extends GossipType>(
         throw new GossipActionError(GossipAction.REJECT, {code: (e as Error).message});
```
twoeths: need to apply action here

dapplion (dismissed):
LGTM!

dapplion (approved):

PR #3782, opened 2022-02-22
Implement support for validator next-epoch proposer duties

### LINKED ISSUE #3746: Support validator next-epoch proposer duties

Referencing Rocketpool's implementation of this feature on an open issue in Lighthouse: https://github.com/sigp/lighthouse/issues/2880

Prysm Issue for Feature: https://github.com/prysmaticlabs/prysm/issues/10035

Already implemented in Nimbus and Teku.

Support Rocketpool's integration of this feature alongside the four different clients.

dadepo:
**Motivation**

**Description**

Provides the ability to request block proposers one epoch in the future. See the corresponding issue #3746 for more background information

Closes #3746

If current Epoch is N,

Make a request to get the proposer duties for epoch N+1:

```
curl -X GET "http://beacon-node/eth/v1/validator/duties/proposer/N+1" -H "accept: application/json"
```

Take note of the returned validator indices.

Using a chain explorer like https://prater.beaconcha.in/, wait till the N+1 epoch becomes the current epoch. Then confirm that the validator indices now listed as part of N+1 are the same returned from the previous curl request to the beacon endpoint.


packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -166,6 +171,20 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
       );
   }
 
+  function computeProposersForEpoch(state: CachedBeaconStateAllForks, epoch: Epoch): ValidatorIndex[] {
+    const effectiveBalanceIncrementsWithLen: EffectiveBalanceIncrements = getEffectiveBalanceIncrementsWithLen(
+      state.validators.length
+    );
+
+    for (let i = 0; i < state.validators.length; i++) {
```
twoeths: This should be available in EpochContext, we don't have to recalculate again

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -166,6 +171,20 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
       );
   }
```
twoeths: Maybe we need some comments here regarding why we calculate it here instead of the state transition

packages/beacon-state-transition/src/cache/epochContext.ts
```diff
@@ -469,7 +469,8 @@ export class EpochContext {
 
   getBeaconProposer(slot: Slot): ValidatorIndex {
     const epoch = computeEpochAtSlot(slot);
-    if (epoch !== this.currentShuffling.epoch) {
+    // disallow retrieving proposer for past epochs and epochs more than 1 epoch in the future
+    if (epoch < this.currentShuffling.epoch || epoch > this.currentShuffling.epoch + 1) {
```
twoeths: why do we need this change? state-transition only caches proposers for current epoch, not next epoch

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -277,16 +296,43 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
       const startSlot = computeStartSlotAtEpoch(epoch);
       await waitForSlot(startSlot); // Must never request for a future slot > currentSlot
```
twoeths: If we request for current epoch, we need this call. If we request for next epoch, we don't need it. And also the comment was misleading, please fix it

@dadepo how does your test in prater work even with this check?

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -277,16 +296,43 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
...
+        // Gather indexes to get pubkeys in batch (performance optimization)
+        for (let i = 0; i < SLOTS_PER_EPOCH; i++) {
+          // getBeaconProposer ensures the requested epoch is correct
+          const validatorIndex = state.getBeaconProposer(startSlot + i);
+          indexes.push(validatorIndex);
+        }
+
+        for (const cachedEpoch of nextEpochProposerDutyCache.keys()) {
+          // Do not keep past cached future proposal duties.
+          if (cachedEpoch < epoch) {
```
twoeths: in this case, epoch is current epoch, maybe we even don't need the cache if `cachedEpoch === currentEpoch`, so `<=` instead?

packages/beacon-state-transition/src/cache/epochContext.ts
dadepo: Without this extra check, if proposers for more than 1 epoch is requested, the proposer for current epoch is returned. That is because of `return this.proposers[slot % SLOTS_PER_EPOCH]` on line 478. 

This checks ensures an error message is returned instead when proposers for more than 1 epoch ahead is requested.

packages/lodestar/src/api/impl/validator/index.ts
dadepo: Updated

packages/lodestar/src/api/impl/validator/index.ts
dadepo: I was a bit confused by this method initially but looking at the implementation, it only waits if the gap between `now` and the next epoch is less than `MAX_API_CLOCK_DISPARITY_MS`, which is set to 1000ms.

Testing on prater seems to work as expected also.

packages/lodestar/src/api/impl/validator/index.ts
dadepo: I think using `<=` would mean that if next epoch is computed, and then if the current epoch is requested and after that, the next epoch is requested again, it would have to be computed again (because requesting the current epoch would have deleted it)

packages/beacon-state-transition/src/cache/epochContext.ts
twoeths: I mean the old check is more strict and we should keep it. Do we have any consumers calling this function with `epoch !== this.currentShuffling.epoch`?

packages/lodestar/src/api/impl/validator/index.ts
twoeths: >  if next epoch is computed, and then if the current epoch is requested and after that, the next epoch is requested again, it would have to be computed again (because requesting the current epoch would have deleted it)

I think it's not deleted because `cachedEpoch` is the next epoch so `cachedEpoch <= epoch` is `false` (because `epoch` is `current epoch`)

dadepo: oh I mean in the case of a malicious actor, who sees that during the current epoch N, the first call to epoch N+1 is expensive before caching, and hence during the same epoch N, continues to make request for epoch N+1, and then epoch N (to delete the previous cache) and then epoch N + 1 again, in other to overwhelm the system

packages/beacon-state-transition/src/cache/epochContext.ts
dadepo: You correct. The change was not needed. Reverted.

packages/lodestar/src/api/impl/validator/index.ts
dadepo: I am not sure I follow, are you referring to adding a comment to `getGenesisBlockRoot` function?

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -58,6 +59,7 @@ const SYNC_TOLERANCE_EPOCHS = 1;
  */
 export function getValidatorApi({chain, config, logger, metrics, network, sync}: ApiModules): routes.validator.Api {
   let genesisBlockRoot: Root | null = null;
+  const nextEpochProposerDutyCache = new Map<Epoch, ValidatorIndex[]>();
```
dapplion: Keeping caches here is not a good separation of concerns. We must keep all caches attached to the BeaconChain class, this functions should be state-less

packages/lodestar/src/api/impl/validator/index.ts
twoeths: nvm, this function just calls `computeProposers` without adding more logics

packages/lodestar/src/api/impl/validator/index.ts
dadepo: You are right. It should be fine to delete the proposers in the cache at the current epoch (the cache value that was previous computed as at a future epoch) when the current epoch is finally reached. I'll update the code now to reflect this.

packages/lodestar/src/chain/chain.ts
```diff
@@ -78,6 +82,9 @@ export class BeaconChain implements IBeaconChain {
   readonly seenSyncCommitteeMessages = new SeenSyncCommitteeMessages();
   readonly seenContributionAndProof = new SeenContributionAndProof();
 
+  // caches computed proposers for epoch N+1 in the future
+  readonly nextEpochProposerDutyCache = new Map<Epoch, ValidatorIndex[]>();
```
twoeths: not sure if we need a separate class to manage this like the above caches and add a `prune(epoch)` method for it, leave it for @dapplion to consider

twoeths (dismissed):

packages/lodestar/src/chain/chain.ts
dadepo: I considered this also but wanted to keep it simple. But happy to create a separate class for it if @dapplion also agree thats preferred

dapplion (changes requested):
I don't think this is the best architecture. The result `futureProposers` is dependant on that specific state. If you just cache by epoch you are committing to a "head" that may change latter. Instead this result of future proposers should be cached in the epoch context, and computed lazily on demand. @tuyennhv What do you think?

wemeetagain:
>  The result futureProposers is dependant on that specific state. If you just cache by epoch you are committing to a "head" that may change latter. Instead this result of future proposers should be cached in the epoch context, and computed lazily on demand. 

I agree. We shouldn't attach the cache to the chain directly, because a reorg can make the proposer cache invalid, and that's currently not handled here. Caching at the level of the beacon state (in our epoch-long cache object, "epoch contect") would be the way to go :+1:

twoeths:
Since the result of next-epoch proposer duties computation is not final at all and it's only consumed by Rocket Pool, initially I didn't want to cache in beacon-state-transition because we may make it confused to cache both current and next epoch's proposers.

@dadepo if the direction is to cache in beacon-state-transition, I guess we just need to name it carefully and make a comment that it's not the final result of next epoch's proposers, at each epoch transition we still need to compute proposers for current epoch again

wemeetagain (dismissed):

packages/beacon-state-transition/src/cache/epochContext.ts
```diff
@@ -400,6 +417,7 @@ export class EpochContext {
     const nextEpoch = currEpoch + 1;
     this.nextShuffling = computeEpochShuffling(state, epochProcess.nextEpochShufflingActiveValidatorIndices, nextEpoch);
     this.proposers = computeProposers(state, this.currentShuffling, this.effectiveBalanceIncrements);
+    this.nextEpochProposers = computeProposers(state, this.nextShuffling, this.effectiveBalanceIncrements);
```
twoeths: I think we should only calculate this on demand (when the api call) and cache it, instead of on every epoch transition

dadepo: I actually thought about that also, but decided to go with this implementation due to its simplicity. (No need for code to only generate this after the first call, and then code to purge previous values when a new epoch is reached) It also does not introduce any performance regression, that is why I went with it.

dapplion: > I actually thought about that also, but decided to go with this implementation due to its simplicity

Should not be complicated at all to implement. Just type as T | null and compute if requested and null.

> It also does not introduce any performance regression, that is why I went with it.

Our performance CI only rejects x3 fn run time changes. computeProposers() has a cost of ~100ms over the state transition which takes 1000ms, so a ballpark 10% increase aprox.

dapplion:
@g11tech Clone nethermind interop branch Sim merge tests step failed, can you take a look?

g11tech:
> @g11tech Clone nethermind interop branch Sim merge tests step failed, can you take a look?

@dadepo  can you rebase your branch and push again?

g11tech (dismissed):

packages/beacon-state-transition/src/cache/epochContext.ts
dadepo: > Should not be complicated at all to implement. Just type as T | null and compute if requested and null.

The complexity is not about the typing though. It's more about passing `CachedBeaconStateAllForks` on the http requests to `computeProposers` - something which is already handled in the transition phase. Also adding a prune call to remove past caches - something which is not needed with this approach as current and next proposers are always refreshed on state transition.

> Our performance CI only rejects x3 fn run time changes. computeProposers() has a cost of ~100ms over the state transition which takes 1000ms, so a ballpark 10% increase aprox.

Okay. Avoiding a ~10% performance hit not a bad thing. I'll update to have it computed on demand.

dadepo: Together with @tuyennhv we looked at how to get the state needed by `computeProposers` which makes implementing the suggestion to compute on demand possible as can be seen in this commit [here](https://github.com/ChainSafe/lodestar/pull/3782/commits/7d08adaf1f98ec5a9151122477b9e816956cc908). 

A couple of observation where I think this approach makes the code less straightforward:

- `getNextEpochBeaconProposer` has to be moved to `BeaconStateContext` in other to have access to the `allForks.BeaconState` needed for `computeProposers` (while the other similar method `getBeaconProposer` remains on `EpochContext`)
- A `getState` has to be added to `BeaconStateContext` to get value of `allForks.BeaconState`
- Even though `getNextEpochBeaconProposer` is on `BeaconStateContext` it still has to reach to `EpochContext` to update the cache
-`getBeaconProposer` method now need to have knowledge of the previous proposer cache since there is now the need to clear the previous `nextEpochProposers` in the cache on request for current proposer.

I guess at the end of the day it boils down to whether this is okay and better than incurring the ~10% decrease in performace - which I am fine with not incurring.

Also open to other suggestions that could simply the compute at demand approach.

dapplion:
@dadepo lint errors

packages/lodestar/test/unit/api/impl/validator/duties/proposer.test.ts
```diff
@@ -40,6 +40,8 @@ describe("get proposers api impl", function () {
     chainStub.clock = server.sandbox.createStubInstance(LocalClock);
     chainStub.forkChoice = server.sandbox.createStubInstance(ForkChoice);
     chainStub.getCanonicalBlockAtSlot.resolves(ssz.phase0.SignedBeaconBlock.defaultValue());
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    (chainStub as any).nextEpochProposerDutyCache = new Map();
```
dapplion: This can be removed right? Also please never cast to `any` unless extremely necessary

twoeths (dismissed):

g11tech (dismissed):

packages/beacon-state-transition/src/cache/cachedBeaconState.ts
```diff
@@ -273,6 +274,22 @@ export class BeaconStateContext<T extends allForks.BeaconState> {
     ) as CachedBeaconState<T>;
   }
 
+  getState(): allForks.BeaconState {
+    return this.type.createTreeBacked(this.tree);
```
dapplion: Why is this necessary? `this` should already be treeBacked

packages/beacon-state-transition/src/cache/epochContext.ts
```diff
@@ -103,6 +103,17 @@ export class EpochContext {
...
+  /**
+   * Map of Indexes of the block proposers keyed by the next epoch.
+   *
+   * We allow requesting proposal duties only one epoch in the future
+   * Note: There is a small probability that returned validators differs
+   * than what is returned when the epoch is reached.
+   *
+   * 32 x Number
+   */
+  nextEpochProposers: Map<Epoch, ValidatorIndex[]> = new Map<Epoch, ValidatorIndex[]>();
```
dapplion: Cache exactly 1 or 0 next proposers for the next epoch of this epochCtx. We should research the safety of +2, +3 epoch duties before generalizing.

packages/beacon-state-transition/src/cache/epochContext.ts
```diff
@@ -474,6 +485,13 @@ export class EpochContext {
         `Requesting beacon proposer for different epoch current shuffling: ${epoch} != ${this.currentShuffling.epoch}`
       );
     }
+
+    for (const cachedEpoch of this.nextEpochProposers.keys()) {
+      // Do not keep past cached future proposal duties.
+      if (cachedEpoch <= epoch) {
+        this.nextEpochProposers.delete(cachedEpoch);
```
dapplion: Why is this necessary? Once you have computed nextEpochProposers entry it can be left there

packages/beacon-state-transition/src/cache/cachedBeaconState.ts
```diff
@@ -273,6 +274,22 @@ export class BeaconStateContext<T extends allForks.BeaconState> {
...
+  }
+
+  getNextEpochBeaconProposer(): ValidatorIndex[] {
```
dapplion: This is dangerous because it uses a state, which mutates every slot to mutate data in the epoch context, which is shared between multiple states. `getNextEpochBeaconProposer` should be in the epoch context and use only data available there immutable during an epoch

dadepo:
Converting this PR to draft. Based on discussing with Lion based on his last observation [here](https://github.com/ChainSafe/lodestar/pull/3782#discussion_r829657687), it was decided to freeze this until the work on ssz is done and merged in, as that would affect the implementation of this feature.

packages/beacon-state-transition/src/cache/epochContext.ts
dadepo: > Cache exactly 1 or 0 next proposers for the next epoch of this epochCtx

Yeah. That is what is being done. Are you hinting at something else?

packages/beacon-state-transition/src/cache/epochContext.ts
```diff
@@ -389,7 +418,10 @@ export class EpochContext {
     const nextEpoch = currEpoch + 1;
 
     this.nextShuffling = computeEpochShuffling(state, epochProcess.nextEpochShufflingActiveValidatorIndices, nextEpoch);
-    this.proposers = computeProposers(state, this.currentShuffling, this.effectiveBalanceIncrements);
+    this.currentProposerSeed = getSeed(state, this.currentShuffling.epoch, DOMAIN_BEACON_PROPOSER);
+    this.nextProposerSeed = getSeed(state, this.nextShuffling.epoch, DOMAIN_BEACON_PROPOSER);
```
twoeths: we only need to call `getSeed()` once for `this.nextProposerSeed`, `this.currrentProposerSeed = this.nextPrroposerSeed`

dapplion:
`packages/beacon-state-transition/src/cache/cachedBeaconState.ts` is an empty file

packages/beacon-state-transition/src/cache/epochContext.ts
dapplion: If there's exactly 1 possible value for that why do a map?

```ts
nextEpochProposers: ValidatorIndex[] | null;
```
is more intuitive

packages/beacon-state-transition/src/cache/epochContext.ts
```diff
@@ -150,6 +163,9 @@ export class EpochContext {
   epoch: Epoch;
   syncPeriod: SyncPeriod;
 
+  currentProposerSeed: Uint8Array;
```
dapplion: Why does needs to be persisted? What's the advantage of keeping this around?

packages/beacon-state-transition/src/cache/epochContext.ts
```diff
@@ -150,6 +163,9 @@ export class EpochContext {
...
+  nextProposerSeed: Uint8Array;
```
dapplion: Same here, what's the advantage of keeping this around? Why not compute on demand and discard after caching `nextEpochProposers`?

packages/beacon-state-transition/src/cache/epochContext.ts
```diff
@@ -467,6 +499,18 @@ export class EpochContext {
...
   }
 
+  getNextEpochBeaconProposer(): ValidatorIndex[] {
+    const nextShuffling = this.nextShuffling;
+    let nextProposers = this.nextEpochProposers.get(nextShuffling.epoch);
+    if (!nextProposers) {
+      nextProposers = computeProposers(this.nextProposerSeed, nextShuffling, this.effectiveBalanceIncrements);
+      // Only keep proposers for one epoch in the future, so clear before setting new value
+      this.nextEpochProposers.clear();
+      this.nextEpochProposers.set(nextShuffling.epoch, nextProposers);
```
dapplion: In a specific `epochCtx` does `nextShuffling.epoch` ever have different values? This .clear, then .set pattern is weird and miss-leading. Maps are meant to hold values of arbitrary keys, but here nextEpochProposers seems to ever hold a single value

packages/beacon-state-transition/test/perf/allForks/util/shufflings.test.ts
```diff
@@ -1,13 +1,15 @@
...
 import {Epoch} from "@chainsafe/lodestar-types";
+import {DOMAIN_BEACON_PROPOSER} from "@chainsafe/lodestar-params";
 import {
   computeEpochAtSlot,
   CachedBeaconStateAllForks,
   computeEpochShuffling,
   getNextSyncCommittee,
 } from "../../../../src";
+import {computeProposers} from "../../../../src";
+import {getSeed} from "../../../../lib";
```
dapplion: Warning bad import

packages/beacon-state-transition/src/cache/epochContext.ts
dadepo: The issue in that, is related to what you observed in the comment [here](https://github.com/ChainSafe/lodestar/pull/3782#discussion_r829657687)

> getNextEpochBeaconProposer should be in the epoch context and use only data available there immutable during an epoch

Computing the seed via `getSeed` method, requires the state which makes it impossible to compute using data available in the context.

dapplion: Then please add comments :pray: all non-obvious decisions must be rationalized in committed comments

packages/beacon-state-transition/src/cache/epochContext.ts
dapplion: The JSDoc for `nextProposerSeed` must be placed in the variable declaration ([here](https://github.com/ChainSafe/lodestar/pull/3782/files#diff-3e6d2b0cf933b99848484935101c7016da5a077e85ca7e16fa132d86a202ab91R156)), here it has no effect. You can duplicate the comment here too if appropiate but use regular `//` comment syntax. Also try to fit the comment to 120 wide characters for consistent formatting.

dapplion: Also here you can comment (with `//`) that the cost of getSeed is negligible provided that this path runs once per epoch.

packages/beacon-state-transition/test/perf/allForks/util/shufflings.test.ts
dapplion: Still bad import

packages/beacon-state-transition/src/cache/epochContext.ts
```diff
@@ -477,6 +511,60 @@ export class EpochContext {
...
+   *
+   * **2**: The active validator set can be predicted `MAX_SEED_LOOKAHEAD` in advance due to how activations are
+   * processed. We already compute the active validator set for the next epoch to optimize epoch processing, so it's
+   * reused here.
+   *
+   * **3**: Effective balance is not known ahead of time, but it rarely changes. Even if it changes, only a few
+   * balances are sampled to adjust the probability of the next selection (32 per epoch on average). So to invalidate
+   * the prediction the effective of one of those 32 samples should change and change the random_byte inequality.
+   */
+  getBeaconProposersNextEpoch(): ValidatorIndex[] {
```
dapplion, line 551: Added a rationale on when and why the proposer prediction stands or is invalidated

packages/beacon-state-transition/src/cache/epochContext.ts
```diff
@@ -477,6 +511,60 @@ export class EpochContext {
...
+    if (!this.proposersNextEpoch.computed) {
+      const indexes = computeProposers(
+        this.proposersNextEpoch.seed,
+        this.nextShuffling,
+        this.effectiveBalanceIncrements
+      );
+      this.proposersNextEpoch = {computed: true, indexes};
+    }
```
dapplion, line 559: Now computation of proposers is deferred AND cached when computed

twoeths (approved):

dapplion (approved):

packages/beacon-state-transition/src/cache/epochContext.ts
dadepo: @dapplion this won't cache though, and on every request to get next proposers, the `this.proposersNextEpoch.computed` check [here](https://github.com/ChainSafe/lodestar/pull/3782/files/ab35439d7109101dc162cb997c42c7f7be454843#diff-3e6d2b0cf933b99848484935101c7016da5a077e85ca7e16fa132d86a202ab91R556) will always return false, and the `if` branch check always run.

You can confirm this by running the application, and trigger the end point multiple times, the if branch will always run.

I noticed this while working on the implementation and I think the reason why this is the case is how the `chain.getHeadStateAtCurrentEpoch` [here](https://github.com/ChainSafe/lodestar/pull/3782/files/ab35439d7109101dc162cb997c42c7f7be454843#diff-b909bf67569ad6ffedfbd4ebe6c611c60e828d24b1e5d04fbafb058abc4eba50R279) works, but I did not go ahead with caching when I realised that the request to the endpoint itself was actually pretty fast and no need to even cache.

dapplion: > I noticed this while working on the implementation and I think the reason why this is the case is how the chain.getHeadStateAtCurrentEpoch [here](https://github.com/ChainSafe/lodestar/pull/3782/files/ab35439d7109101dc162cb997c42c7f7be454843#diff-b909bf67569ad6ffedfbd4ebe6c611c60e828d24b1e5d04fbafb058abc4eba50R279) works

If this is true, then it's a big issue! Can you open a dedicated issue for it? It should be investigated latter

> I did not go ahead with caching when I realised that the request to the endpoint itself was actually pretty fast and no need to even cache.

This code runs on the main thread and according to benchmarks it takes ~100ms. That's still a significant amount of time to block the main thread and repeated computations must be avoided if possible

philknows: And please link back to this context after opening a new dedicated issue. :)

wemeetagain: #4039

Issue #3731, opened 2022-02-08
Kiln 🔥🧱 (the merge) devnets 2022 tracker


g11tech:
## Kiln 🔥🧱 

Kiln 🔥🧱  is the followup sprint of [Kintsugi :tea: ](https://github.com/ChainSafe/lodestar/issues/3452) with aim to move client implementations into the production readiness stage. This is the supposed to be last devnet before pre-production testsnets.

Kiln v1 spec: https://hackmd.io/@n0ble/kiln-spec
Kiln v2 spec: https://hackmd.io/@n0ble/kiln-spec#v2-change-set
Milestones tracker: https://notes.ethereum.org/@timbeiko/kiln-milestones
Engine API test vectors: https://notes.ethereum.org/rmVErCfCRPKGqGkUe89-Kg?view
### CL Spec Tracker 
- [CL spec 1.1.9](https://github.com/ChainSafe/lodestar/issues/3677)
#### Kiln v1:
- [x] Interface lodestar with the ELs on Kiln spec 1.0.0.alpha.6 https://github.com/ChainSafe/lodestar/pull/3736
- [x] basic interop with
  - [x] nethermind
  - [x] geth
- [x] #3751
- [x] devnet4 participation https://notes.ethereum.org/D5xK4XrmTb6MKGudf_hcrg
  - [x] setup of devnet by EL team
  - [x] Merge transition
  - [x] Issues/Observations 
    - This devnet was important for the EL's `beacon sync` apart from the changed api semantics. 
    - At the merge points, there were atleast 2 pow chains, and hence when the terminal block came in from the clients that were ahead, the lodestar node (whose EL  - nethermind was on another fork), got accepted and didn't move ahead  
    - Because of some issue (already being investigated by nethermind team), it didn't sync to right chain, with ACCEPTED status being rejected till the head of the chain (of synced ELs) move 128 slots ahead, after that it started syncing, and synced up to chain head with an execution status of syncing. Lodestar's behavior in this is as expected and as per the optimistic specs.
#### Kiln v2
- [x] https://github.com/ChainSafe/lodestar/pull/3777
- [x] Rename random to prevRandao https://github.com/ChainSafe/lodestar/pull/3787
~~- [ ] Add engine_exchangeTransitionConfigurationV1 https://github.com/ChainSafe/lodestar/pull/3785~~  moved to https://github.com/ChainSafe/lodestar/issues/3945

#### Devnet -5
[...]

PR #3785, opened 2022-02-24
exchangeTransitionConfiguration heartbeat implementation

g11tech:
**Motivation**
Execution engines will now hosts an api to specify and match transition configuration https://github.com/ethereum/execution-apis/pull/172, which is supposed to act as a heatbeat between EL and CL.

This PR implements and matches the same and aims to implements the heart beat functionality
**Description**
PS: this work is currently WIP and some details/clarifications are awaited. Also this is not the part of Kiln v2 spec.
Part of [Kiln tracker](https://github.com/ChainSafe/lodestar/issues/3731)

### DIFF
--- a/packages/lodestar/src/executionEngine/http.ts
+++ b/packages/lodestar/src/executionEngine/http.ts
@@ -21,6 +21,9 @@ import {
   PayloadId,
   PayloadAttributes,
   ApiPayloadAttributes,
+  TransitionConfigOpts,
+  ApiTransitionConfig,
+  TransitionConfig,
 } from "./interface";
 
 export type ExecutionEngineHttpOpts = {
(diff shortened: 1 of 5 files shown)


packages/lodestar/src/executionEngine/http.ts
```diff
@@ -44,14 +47,65 @@ export const defaultExecutionEngineHttpOpts: ExecutionEngineHttpOpts = {
...
+  constructor(opts: ExecutionEngineHttpOpts & TransitionConfigOpts, signal: AbortSignal, rpc?: IJsonRpcHttpClient) {
+    this.transitionConfig = serializeTransitionConfig(opts.transitionConfig);
+    this.heartBeatInSec = opts.heartBeatInSec;
     this.rpc =
       rpc ??
       new JsonRpcHttpClient(opts.urls, {
         signal,
         timeout: opts.timeout,
       });
+    void this.exchangeTransitionConfiguration();
```
dapplion, line 63: void Promise that may reject is a bad idea.

packages/lodestar/src/executionEngine/http.ts
```diff
@@ -44,14 +47,65 @@ export const defaultExecutionEngineHttpOpts: ExecutionEngineHttpOpts = {
...
+    // TODO: determine the throw condition, also identify how to update terminal blocktracker hash
+    // and the terminal number depending upon
+    // 1. eth1 tracker 2. the terminal block received
+    // as per spec, this must not be dynamic
+    if (
+      transitionConfig.terminalTotalDifficulty !== this.transitionConfig.terminalTotalDifficulty ||
+      transitionConfig.terminalBlockHash !== this.transitionConfig.terminalBlockHash ||
+      transitionConfig.terminalBlockNumber !== this.transitionConfig.terminalBlockNumber
+    ) {
+      throw Error(
```
dapplion, line 99: This http engine should not be responsible to throw this Error. There should be something at a higher level that does this recurring calls and logs loudly if there's a mismatch. Keep in mind that there may be multiple transports in the future besides http. The return of this function could be a status

```ts
type ReturnType = {match: true} | {match: false; local: TransitionConfig; remote: TransitionConfig}
```

packages/lodestar/src/executionEngine/http.ts
```diff
@@ -44,14 +47,65 @@ export const defaultExecutionEngineHttpOpts: ExecutionEngineHttpOpts = {
...
+        `Transition config mismatch, actual=${JSON.stringify(transitionConfig)}, expected=${JSON.stringify(
+          this.transitionConfig
+        )}`
+      );
+    }
+    // Use this lastHeartBeatAt in the actual engine api calls to identify if this needs to
+    // be called again
+    this.lastHeartBeatAt = Math.floor(new Date().getTime() / 1000);
```
dapplion, line 107: This class should not be responsible to track the frequency of calls, this class should be state-less

g11tech:
closing this to begin afresh as the code has diverged quite a bit

PR #3791, opened 2022-02-25
Make it possible to run devnet setup script on macOS

dadepo:
**Motivation**

There are slight differences between the docker api on linux and macOS. 
The current devnet setup script assumes linux and hence the start up script does not work on macOs.
This PR updates the startup script making it possible to also use the startup script on a macOS

### DIFF
--- a/kiln/devnets/setup.sh
+++ b/kiln/devnets/setup.sh
@@ -16,6 +16,8 @@ if [ ! -n "$dataDir" ] || [ ! -n "$devnetVars" ] || ([ "$elClient" != "geth" ] &
 then
   echo "usage: ./setup.sh --dataDir <data dir> --elClient <geth | nethermind> --devetVars <devnet vars file> [--dockerWithSudo --withTerminal \"gnome-terminal --disable-factory --\"]"
   echo "example: ./setup.sh --dataDir devnet4-data --elClient nethermind --devnetVars ./devnet4.vars --dockerWithSudo --withTerminal \"gnome-terminal --disable-factory --\""
+  echo "Note: if running on macOS where gnome-terminal is not available, remove the gnome-terminal related flags."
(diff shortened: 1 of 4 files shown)


kiln/devnets/setup.sh
```diff
@@ -127,7 +151,13 @@ then
 	echo "you can watch observe the client logs at the respective terminals"
 	echo "use ctl + c on any of these three (including this) terminals to stop the process"
 	echo "waiting ..."
-	wait -n $elPid $clPid
+	if [ $platform == 'Darwin' ]
+	then # macOs ships with an old version of bash with wait that does not have the -n flag
+	  wait $elPid
+	  wait $clPid
```
g11tech, line 157: Is there a way to wait for both the processes together in Mac ? Because the intend here is to check if any process exits then do cleanup

dadepo: > Is there a way to wait for both the processes together in Mac ?

This is exactly what happens. The non-mac version which uses `-n` flag, modifies the `wait` command to only wait for a single job from the provided process ID or job specs to finish before returning its exit status [source](https://linuxhint.com/bash-wait-command-linux/#:~:text=Using%20%E2%80%93n%20option%3A,if%20no%20arguments%20are%20provided). 

I tested this out also and confirmed this behaviour of `-n` flag on a linux server.

So the mac version won't short-circuit, while the -n flag will. There would probably be a _hackish_ way to have the same circuiting behaviour for mac, but I consider losing that behaviour to be acceptable as it does not radically change the utility of the script.

wemeetagain (dismissed):

g11tech (dismissed):

g11tech:
@dadepo I have updated the lodestar image with devnet4 compatible build since master has now kiln v2 changes not compatible with devnet4. :+1:

g11tech (approved):

wemeetagain (approved):

PR #3793, opened 2022-02-26
Verify signatures on both main thread and worker threads

### LINKED ISSUE #3792: Improve BLS thread pool - job wait time

The BLS Thread pool - job wait time is designed to be consistently less than 100ms, however a lot of time it's > 100ms, maybe even up to 500ms.

Note that this metric is important for the gossip Job Wait Time, we want it to be consistently <3s to avoid the gosisp peer penalty due to broken IWANT promises.

+ The reason is when non-batchable jobs are sent to bls, it forces the worker thread to run and we don't utilize the worker threads at that time. If we only send batchable jobs to worker threads, it works as designed

Screen Shot 2022-02-26 at 14 00 37

+ In average, non-batchable signature sets happen only when we validate gossip blocks or run state transition, and it's <2ms in contabo or <1ms in hetzner so it's ok to run them in main-thread
+ Verify batchable signature sets in worker threads

twoeths:
**Motivation**

We want a stable BLS - Job Wait Time metric, ideally it should be < 100ms consistently

**Description**

+ Write a `BlsMixedVerifier` to send batchable bls jobs to worker threads and non-batchable ones to main thread
+ It's not likely we want to try using singleThreadVerifier, so change the flag to `useMultiThreadVerifier`, default to false
+ More explanation on #3792

Closes #3792


dapplion:
I would rather add another flag when sending jobs to the BLS verifier where you opt-out of multi-thread

https://github.com/ChainSafe/lodestar/blob/060918ecf2c8b2c1dbe02ea2ca2100f2efeb2b18/packages/lodestar/src/chain/bls/interface.ts#L11

Opting-out and using more resources on the main thread should be done very carefully and which good backing data that's it's a good trade-off

dapplion (changes requested):
See above

twoeths:
Test result in contabo-18 (250 validators) for ~16h (previously it's `tuyen/gossip_penalize_peer` branch)

+ BLS job wait time is improved 
Screen Shot 2022-02-28 at 08 20 21

+ Thanks for that gossip validation time is improved
Screen Shot 2022-02-28 at 08 22 19

+ Gossip Block Processed delay has been great since we deploy this branch, even through we verify gossip blocks' signatures in the main thread
Screen Shot 2022-02-28 at 08 24 15

Looks like it's a good trade-off. Also @dapplion I notice the "gossip validation gap" issue in previous branch (`tuyen/gossip_penalize_peer`), which hasn't happened in this branch so far. At least we know that issue is not specific to `ssz-v2` branch.

dapplion:
Some feedback items

1. Remove BlsMixed class, just add and if inside the BlsMultiThread class

```ts
  async verifySignatureSets(sets: ISignatureSet[], opts: VerifySignatureOpts = {}): Promise<boolean> {
    if (opts.useMainThread) {
      return verifySignatureSetsMaybeBatch(
        sets.map((set) => ({
          publicKey: getAggregatedPubkey(set),
          message: set.signingRoot.valueOf() as Uint8Array,
          signature: set.signature,
        }))
      );
    }
```

2. Verifying blocks on mainthread during sync will be a performance issue. Add a flag in the block processor `{blsVerifyMainThread: true}`, and set to true only on gossip handler and unknown block sync

3. To clean-up the flags I think we should have
- `--blsVerifyAllMainThread` default false. In BeaconChain constructor uses the BlsSingleThread class
- `--blsVerifyAllMultiThread` default false. Ignores the useMainThread flag in the BlsMultiThreadWorkerPool

4. Add a metric that counts total CPU time spent verifying signatures on the main thread. Just an incremental gauge of seconds using `process.hrtime.bigint()`. Now there will be two places potentially verifying signatures:
- `verifySignatureSetsMaybeBatch` inside `BlsMultiThreadWorkerPool.verifySignatureSets`: histogram `lodestar_bls_thread_pool_main_thread_time_seconds` with tight bands `[0.1, 1]`
- `verifySignatureSetsMaybeBatch` inside `BlsSingleThreadVerifier.verifySignatureSets`: histogram `lodestar_bls_single_thread_time_seconds` with tight bands `[0.1, 1]`

packages/lodestar/src/chain/blocks/types.ts
```diff
@@ -38,6 +38,10 @@ export type PartiallyVerifiedBlockFlags = FullyVerifiedBlockFlags & {
    * From RangeSync module, we won't attest to this block so it's okay to ignore a SYNCING message from execution layer
    */
   fromRangeSync?: boolean;
+  /**
+   * Verify signatures on main thread or not.
+   */
+  blsVerifyMainThread?: boolean;
```
dapplion: Maybe better name `blsVerifyOnMainThread`?

packages/lodestar/src/chain/bls/interface.ts
```diff
@@ -9,6 +9,12 @@ export type VerifySignatureOpts = {
    * Only non-time critical objects should be marked as batchable, since the pool may hold them for 100ms.
    */
   batchable?: boolean;
+
+  /**
+   * Use main thread to verify signatures, use this with care.
+   * Ignore the batchable option if this is true.
+   */
+  useMainThread?: boolean;
```
dapplion: Should we name `verifyOnMainThread` for consistency with `blsVerifyOnMainThread`?

packages/lodestar/src/chain/bls/multithread/index.ts
```diff
@@ -136,6 +143,22 @@ export class BlsMultiThreadWorkerPool implements IBlsVerifier {
...
+      const isValid = verifySignatureSetsMaybeBatch(
+        sets.map((set) => ({
+          publicKey: getAggregatedPubkey(set),
+          message: set.signingRoot.valueOf() as Uint8Array,
+          signature: set.signature,
+        }))
+      );
+
+      const endNs = process.hrtime.bigint();
+      this.metrics?.blsTime.mainThreadDurationInThreadPool.observe(Number(endNs - startNs) / 1e9);
```
dapplion: You can use the timer function which does calls the process time for you. Also wrap in try {} finally {} in case verifySignatureSetsMaybeBatch rejects

```ts
const timer = metrics.startTimer()
try {
  return fn()
} finally {
  timer()
}
```

packages/lodestar/src/chain/chain.ts
```diff
@@ -119,9 +120,13 @@ export class BeaconChain implements IBeaconChain {
...
     const signal = this.abortController.signal;
     const emitter = new ChainEventEmitter();
-    const bls = opts.useSingleThreadVerifier
-      ? new BlsSingleThreadVerifier()
-      : new BlsMultiThreadWorkerPool({logger, metrics, signal: this.abortController.signal});
+    const blsModules = {logger, metrics, signal: this.abortController.signal};
+    const bls = opts.blsVerifyAllMainThread
+      ? new BlsSingleThreadVerifier({metrics})
+      : opts.blsVerifyAllMultiThread
+      ? new BlsMultiThreadWorkerPool({blsVerifyAllMultiThread: true}, blsModules)
```
dapplion: Just do
```ts
new BlsMultiThreadWorkerPool(opts, blsModules)
```

packages/lodestar/src/chain/bls/singleThread.ts
```diff
@@ -1,16 +1,29 @@
 import {ISignatureSet} from "@chainsafe/lodestar-beacon-state-transition";
+import {IMetrics} from "../../metrics";
 import {IBlsVerifier} from "./interface";
 import {verifySignatureSetsMaybeBatch} from "./maybeBatch";
 import {getAggregatedPubkey} from "./utils";
 
 export class BlsSingleThreadVerifier implements IBlsVerifier {
+  private readonly metrics: IMetrics | null;
+
+  constructor({metrics = null}: {metrics: IMetrics | null}) {
```
dapplion, line 10: Equivalent but constructor arguments like this look much cleaner
```ts
constructor(private readonly metrics: IMetrics | null) {}
```

packages/lodestar/src/chain/bls/singleThread.ts
```diff
@@ -1,16 +1,29 @@
...
+    const isValid = verifySignatureSetsMaybeBatch(
       sets.map((set) => ({
         publicKey: getAggregatedPubkey(set),
         message: set.signingRoot.valueOf() as Uint8Array,
         signature: set.signature,
       }))
     );
+
+    const endNs = process.hrtime.bigint();
+    this.metrics?.blsTime.singleThreadDuration.observe(Number(endNs - startNs) / 1e9);
```
dapplion: Use metrics.startTimer() and try {} finally {}

packages/lodestar/src/metrics/metrics/lodestar.ts
```diff
@@ -367,6 +367,20 @@ export function createLodestarMetrics(
...
     },
 
+    // BLS time
+    blsTime: {
+      singleThreadDuration: register.histogram({
+        name: "lodestar_bls_single_thread_time_seconds",
+        help: "Time to verify signatures with single thread mode",
+        buckets: [0.1, 1],
+      }),
+      mainThreadDurationInThreadPool: register.histogram({
```
dapplion: There's already a group `blsThreadPool`, add `mainThreadDurationInThreadPool` to there

packages/lodestar/src/metrics/metrics/lodestar.ts
```diff
@@ -367,6 +367,20 @@ export function createLodestarMetrics(
       }),
     },
 
+    // BLS time
+    blsTime: {
```
dapplion: Put right after blsThreadPool, with group `blsSingleThread` or no group at all

packages/lodestar/src/network/gossip/handlers/index.ts
```diff
@@ -111,7 +111,7 @@ export function getGossipHandlers(modules: ValidatorFnsModules, options: GossipH
 
       // `validProposerSignature = true`, in gossip validation the proposer signature is checked
       chain
-        .processBlock(signedBlock, {validProposerSignature: true})
+        .processBlock(signedBlock, {validProposerSignature: true, blsVerifyMainThread: true})
```
dapplion: Add explanation above the `blsVerifyMainThread` option linking to the issue and investigation that motivated this change

packages/lodestar/src/sync/unknownBlock.ts
```diff
@@ -164,7 +164,9 @@ export class UnknownBlockSync {
     }
 
     pendingBlock.status = PendingBlockStatus.processing;
-    const res = await wrapError(this.chain.processBlock(pendingBlock.signedBlock, {ignoreIfKnown: true}));
+    const res = await wrapError(
+      this.chain.processBlock(pendingBlock.signedBlock, {ignoreIfKnown: true, blsVerifyMainThread: true})
```
dapplion: Add explanation above the blsVerifyMainThread option linking to the issue and investigation that motivated this change. Copy paste from `packages/lodestar/src/network/gossip/handlers/index.ts`.

Then please add a note why unknown block sync should use true and range sync false

packages/lodestar/src/sync/range/range.ts
```diff
@@ -194,6 +194,8 @@ export class RangeSync extends (EventEmitter as {new (): RangeSyncEmitter}) {
       ignoreIfFinalized: true,
       // We won't attest to this block so it's okay to ignore a SYNCING message from execution layer
       fromRangeSync: true,
+      // When syncing, we should verify signatures in worker threads
+      blsVerifyMainThread: false,
```
dapplion: add a note why unknown block sync should use true and range sync false

twoeths:
with #3812, gossip p7 penalty (broken IWANT promises issue) is avoided most of the time so this PR is less important

also in hetzner-test0 (with >900 validators), BLS Job wait time is consistently low
Screen Shot 2022-03-01 at 17 34 21

dapplion (approved):

dapplion:
> with #3812, gossip p7 penalty (broken IWANT promises issue) is avoided most of the time so this PR is less important

I want to revert #3812 (see https://github.com/ChainSafe/lodestar/pull/3815), so I think we should merge this PR.

PR #3797, opened 2022-02-27
Persist invalid SSZ objects only if enabled

### LINKED ISSUE #4069: Stop using file system sync api
**Describe the bug**

As clearly stated in https://nodejs.org/api/fs.html#synchronous-example, `The synchronous APIs block the Node.js event loop and further JavaScript execution until the operation is complete` so we should only use it in testing context

**Expected behavior**

Switch to equivalent async api

dapplion:
**Motivation**

Currently the the beacon node persists to disk invalid SSZ objects. This functionality is absolutely critical to debug consensus bugs and issues with gossip validation. However, this feature should not be enabled by default in production servers. Only us or debuggers should use this feature.

**Description**

- Persist invalid SSZ objects only if enabled
- Modularize a bit more to abstract naming and logging

Closes #4069 
part of #4002

### DIFF
--- a/packages/cli/src/cmds/beacon/options.ts
+++ b/packages/cli/src/cmds/beacon/options.ts
@@ -98,7 +98,7 @@ export const beaconPathsOptions: ICliCommandOptions<IBeaconPaths> = {
   },
 
   persistInvalidSszObjectsDir: {
-    description: "Directory to persist invalid ssz objects",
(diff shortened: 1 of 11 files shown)


dapplion:
Note: without this PR, the persist invalid SSZ objects is a DOS vector as it will fill the disk with junk until running out of space

g11tech (approved):

wemeetagain (approved):

PR #3800, opened 2022-02-28
JobItemQueue: only run job on push if there're free slots

twoeths:
**Motivation**

If there are `maxConcurrency` running jobs and we push more jobs to the queue, we don't have to trigger `runJob` on push. Jobs will be triggered to run from the current running jobs.

**Description**

```ts
if (this.runningJobs < this.opts.maxConcurrency) {
        setTimeout(this.runJob, 0);
      }
```


dapplion (approved):
LGTM! When jobs finish, they should trigger runJob anyway

PR #3801, opened 2022-02-28
Kurtosis multi client run in CI

g11tech:
**Motivation**
Currently, we only do lodestar sims in the CI. Kurtosis enables us to do multi client CI runs

This PR rips off from @parithosh's CI scripts, adapting them for lodestar! thanks @parithosh!
Over time, a lodestar beacon/validator interop could be setup as well with other client's beacon and validators.
**Description**
PS: WIP


dapplion:
This could be huge! Thanks. Let's see how the Github actions handle this load

dapplion:
This task is being taken over by devops

PR #3808, opened 2022-02-28
Add detailed score metrics

dapplion:
**Motivation**

We need more tools to debug why gossip peer scores are so low

**Description**

Add very detailed metrics of gossip scores

- Avg / min / max across all peers labeled by p parameter and topic type

Note: Code duplication from js-libp2p-gossipub, some unit test should ensure the logic is consistent. For the short term it's not an issue, and this debugging tool is very useful now.

Screenshot from 2022-02-28 23-23-42

Charts above with 

_left_
```
sum (lodestar_gossip_score_weights_avg_min_max_avg) by (p)
```

_right_
```
sum (lodestar_gossip_score_weights_avg_min_max_avg{p=~"p1|p2|p3|p4"} ) by (topic)
```

### DIFF
--- a/packages/lodestar/src/network/gossip/scoreMetrics.ts
+++ b/packages/lodestar/src/network/gossip/scoreMetrics.ts
@@ -1,0 +1,196 @@
+import {PeerScoreParams} from "libp2p-gossipsub/src/score";
+import {PeerStats} from "libp2p-gossipsub/src/score/peer-stats";
+import {MapDef} from "../../util/map";
+import {GossipType} from "./interface";
+import {GossipTopicCache} from "./topic";
+
+export type TopicScoreWeights<T> = {p1w: T; p2w: T; p3w: T; p3bw: T; p4w: T};
+export type ScoreWeights<T> = {
+  byTopic: MapDef<GossipType, TopicScoreWeights<T>>;
+  p5w: T;
+  p6w: T;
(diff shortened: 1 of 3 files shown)


dapplion:
Also running in contabo-5, shows similar results where p7 is the biggest contributor to low scores by a lot

Screenshot from 2022-02-28 23-41-19

packages/lodestar/src/network/gossip/scoreMetrics.ts
```diff
@@ -0,0 +1,196 @@
...
+  byTopic: MapDef<GossipType, TopicScoreWeights<T>>;
+  p5w: T;
+  p6w: T;
+  p7w: T;
+  score: T;
+};
+
+const gossipTypes = Object.values(GossipType);
+
+export function computeScoreWeights(
```
twoeths, line 18: `computeScores()` used to be a performance issue in gossipsub and we have to implement cache strategy there, I wonder if this affects lodestar's performance

dapplion: The only deployment I did didn't show to increase metrics scrape time. This function here is run once per peer and scrape peer. While in gossipsub is run for every message, which we receive hundreds per second.

wemeetagain (approved):

PR #3777, opened 2022-02-19
Add jwt based token auth to the engine api calls

g11tech:
**Motivation**
[Kiln spec v2](https://github.com/ethereum/execution-apis/pull/167) mandates jwt based auth for engine api calls to EL, using a shared jwt-secret fovided via a hex encoded file. 

This PR implements providing the file with hex encoded 256 bit secret via cli param `--jwt-secret /path/to/jwtsecret/file`. 
**Description**
Part of [Kiln tracker](https://github.com/ChainSafe/lodestar/issues/3731)
TODO:
- [x] instead of passing secret read from file and authenticate its length as per spec
- [x] test it against geth
- [x] Update CI sim merge for geth

PR #3809, opened 2022-02-28
Don't track unstable LinkedList benchmark

dapplion:
**Motivation**

We got the results of the data-structure performance, which is already implemented. Disable since it's failing CI builds

**Description**

Don't track unstable LinkedList benchmark

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2363,7 +2363,7 @@
   resolved "https://registry.yarnpkg.com/@types/json5/-/json5-0.0.29.tgz#ee28707ae94e11d2b827bcbe5270bcea7f3e71ee"
   integrity sha1-7ihweulOEdK4J7y+UnC86n8+ce4=
 
-"@types/jwt-simple@^0.5.33":
+"@types/jwt-simple@0.5.33":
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -2363,7 +2363,7 @@
   resolved "https://registry.yarnpkg.com/@types/json5/-/json5-0.0.29.tgz#ee28707ae94e11d2b827bcbe5270bcea7f3e71ee"
   integrity sha1-7ihweulOEdK4J7y+UnC86n8+ce4=
 
-"@types/jwt-simple@^0.5.33":
+"@types/jwt-simple@0.5.33":
```
twoeths, line 2366: not sure why this is ~~changed~~ modified

twoeths (approved):

wemeetagain (approved):

yarn.lock
dapplion: This PR https://github.com/ChainSafe/lodestar/pull/3777 did not update the yarn.lock after requiring a different version. Running yarn causes this diff

PR #3812, opened 2022-02-28
Call deliverMessage before gossip queue

### LINKED ISSUE #3811: Gossip P7 penalty causes peers to have very low scores
**Describe the bug**

Metrics from https://github.com/ChainSafe/lodestar/pull/3808 show that peers have high P7 penalties dragging the score down a lot. Without evindence yet, it seems that it's caused to broken promises. The peer actually delivers the message but _pubish is not called in time since wait times in queues can be very long.

Screenshot from 2022-02-28 23-41-19

**Expected behavior**

Don't penalize peers for long queue times

dapplion:
**Motivation**

See https://github.com/ChainSafe/lodestar/issues/3811

**Description**

- Call deliverMessage before gossip queue

**Todo**

Is there any security risk with this approach? Unless it's high, I think we must adopt this change while queue times are long. In the future this change can be reverted

Closes https://github.com/ChainSafe/lodestar/issues/3811

### DIFF
--- a/packages/lodestar/src/network/gossip/gossipsub.ts
+++ b/packages/lodestar/src/network/gossip/gossipsub.ts
@@ -230,6 +230,11 @@ export class Eth2Gossipsub extends Gossipsub {
    */
   async validate(message: InMessage): Promise<void> {
     try {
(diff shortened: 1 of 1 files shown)


wemeetagain (approved):
LGTM

PR #3816, opened 2022-03-01
Add node exporter metrics section

dapplion:
**Motivation**

It's very common to run node-exporter on any deployment, which gives really detailed system metrics. Had this charts laying around for a while, committing before they get lost :)

**Description**

Add node exporter metrics section

Screenshot from 2022-03-01 17-27-57

Screenshot from 2022-03-01 17-28-15


dapplion:
Any ideas for an important chart to fill the bottom right hole?

wemeetagain (approved):

PR #3823, opened 2022-03-03
Raise peer limit

dapplion:
**Motivation**

Our node struggles to have sufficient peers in beacon_attestation subnets and could be a root cause of bad validator performance. All other clients have higher target peer counts than 25, so we should explore the effects of raising this limit.

**Description**

Raise peer limit to 50

**Early findings**

This branch is deployed at 12:30, in a Prater node with 920 validators

Screenshot from 2022-03-03 14-57-04

Now we have way more mesh peers per subnet, with always > 0 per subnet

Screenshot from 2022-03-03 14-54-49

Network traffic increases significantly both in / out

Screenshot from 2022-03-03 14-55-37

We have more incoming PING messages

Screenshot from 2022-03-03 14-59-52

We process less? beacon_attestation messages

Screenshot from 2022-03-03 15-02-37


dapplion:
This PR is a draft since this needs extensive testing to understand the effects

wemeetagain:
What do we need to do to feel comfortable merging this?

dapplion:
> What do we need to do to feel comfortable merging this?

I would like to fix our gossipsub scoring issues first, then bump peer count

twoeths:
gossipsub PR is merged

twoeths:
gossipsub score has been stable for days in feature group, it's been good in master after we merge new gossipsub too

Screen Shot 2022-04-05 at 18 50 24

twoeths (approved):

PR #3826, opened 2022-03-04
Make PeerId ephemeral by default

dapplion:
**Motivation**

We suspect our node tends to get down-scored by peers. While issues like https://github.com/ChainSafe/lodestar/issues/3811 are being worked on, it's an okay strategy to clear the PeerId on each process session. AFAIK Nimbus also does this for privacy reasons.

**Description**

- Make PeerId ephemeral by default.

- Add flag `--peerIdPersist` to re-use an existing PeerId in disk


wemeetagain:
Closing this since last discussion about this we decided against doing this. Feel free to reopen at a later date if things change

PR #3836, opened 2022-03-08
execution engine fcU call refinement

g11tech:
**Motivation**
Nethermind(@MarekM25) reported it does not get an fcU event on starting from a weak subjectivity state while the lighthouse apparently does.

When the beacon starts from a weak subjectivity state or from a saved DB state it now sends fcU event to the execution engine so that it can start syncing.


g11tech:
converting to draft as nethermind team reported that they would also need a new payload before fcU at initialization point, which will require full payload, but at this point we might only have payload header (different is former has full transactions object while later has transaction root only)

g11tech:
not relevant anymore

PR #3838, opened 2022-03-08
Use in-memory peer store for ephemeral data

dapplion:
**Motivation**

From bumping libp2p here https://github.com/ChainSafe/lodestar/pull/3661 it required to turn the peerStore into async methods. I believe persisting all this data to disk is unnecessary, which adds a performance cost and polutes the disk. Also it makes it harder to maintain leaking the need for async functions in many new places

**Description**

Use in-memory peer store for ephemeral data:

- score: It's just a number. You can keep thousands and not cause a memory issue. Prune after X time, like 1 hour after disconnection. In this implementation the score is let to decay until a threshold on which it is pruned
- addressBook: Only relevant while the peer is connected, prune after disconnection so bounded by peer count
- metadata: Same as above, only relevant for connected peers
- encoding: Same as above, only relevant for connected peers

The only extre requirement / todo is to persist the banned peers to disk. This can be done in an interval heartbeat or when the node is about to close. Since it's not critical at the moment it can be left as a todo


twoeths (approved):
Loos good to me!

PR #3841, opened 2022-03-09
Cache AgentVersion and PeerData in network globals

dapplion:
**Motivation**

Currently we read AgentVersion and ClientKind from the peerstore everytime it's needed. In the libp2p bump https://github.com/ChainSafe/lodestar/pull/3661 this can trigger disk reads. This data is not necessary to persist besides the lifetime of the peer.

This PR extends https://github.com/ChainSafe/lodestar/pull/3838 adding agentVersion and clientKind to the ephemeral data attached to connected peers, which is dropped on disconnection.

In the future we should do a PR upstream to libp2p to emit the response of the identity protocol to prevent having to poll agentVersions from the peerstore

**Description**

- Cache AgentVersion and PeerData in network globals
- Move encodingPreference to PeerData

### DIFF
--- a/packages/lodestar/src/network/globals.ts
+++ b/packages/lodestar/src/network/globals.ts
@@ -1,0 +1,51 @@
+import {altair} from "@chainsafe/lodestar-types";
+import PeerId from "peer-id";
+import {ClientKind} from "./peers/client";
+import {Encoding} from "./reqresp/types";
+
+type PeerIdStr = string;
+
+export enum RelevantPeerStatus {
+  Unknown = "unknown",
+  relevant = "relevant",
+  irrelevant = "irrelevant",
+}
+
+export type PeerData = {
+  lastReceivedMsgUnixTsMs: number;
+  lastStatusUnixTsMs: number;
+  connectedUnixTsMs: number;
+  relevantStatus: RelevantPeerStatus;
+  direction: "inbound" | "outbound";
+  peerId: PeerId;
+  metadata: altair.Metadata | null;
+  agentVersion: string | null;
+  agentClient: ClientKind | null;
+  encodingPreference: Encoding | null;
+};
+
+/**
+ * Make data available to multiple components in the network stack.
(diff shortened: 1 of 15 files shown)


wemeetagain:
There's an event emitted from the peerstore.
`'change:metadata', {peerId, metadata}: {peerId: PeerId; metadata: Map<string, Uint8Array>}`

twoeths (dismissed):

packages/lodestar/src/network/globals.ts
```diff
@@ -0,0 +1,55 @@
...
+};
+
+/**
+ * Make data available to multiple components in the network stack.
+ * Due to class dependencies some modules have circular dependencies, like PeerManager - ReqResp.
+ * This third party class allows data to be available to both.
+ *
+ * The pruning and bounding of this class is handled by the PeerManager
+ */
+export class NetworkGlobals {
```
wemeetagain: Maybe this would be better called `PeerMetadata` and `peerMetadata` throughout.

Thoughts?

twoeths: we have consensus MetaData already, so maybe make this one `PeersData`? there is internal `PeerData` inside

packages/lodestar/src/network/peers/peerManager.ts
```diff
@@ -553,15 +567,20 @@ export class PeerManager {
   /** Register peer count metrics */
   private async runPeerCountMetrics(metrics: IMetrics): Promise<void> {
     let total = 0;
+
+    // TODO: Should we use here `this.connectedPeers`?
```
wemeetagain: I don't think we need/want to use `this.connectedPeers` here?

packages/lodestar/src/network/globals.ts
wemeetagain: would still like to see the name changed to be more specific than "network globals"

wemeetagain (approved):
LGTM

PR #3842, opened 2022-03-09
bumping blst to v0.2.4

### LINKED ISSUE #3692: Docker containers on Docker Hub for x64, arm64, and with a single multi-arch manifest that contains both

dadepo:
**Motivation**

Upgrades the blst dependency to the latest verstion (v0.2.4)

Closes #3692

### DIFF
--- a/packages/lodestar/test/e2e/chain/bls/multithread.test.ts
+++ b/packages/lodestar/test/e2e/chain/bls/multithread.test.ts
@@ -88,7 +88,7 @@ describe("chain / bls / multithread queue", function () {
       isValidPromiseArr.push(pool.verifySignatureSets(sets, {batchable: true}));
     }
 
-    await expect(isInvalidPromise).to.rejectedWith("BLST_ERROR");
+    await expect(isInvalidPromise).to.rejectedWith("BLST_INVALID_SIZE");
 
     const isValidArr = await Promise.all(isValidPromiseArr);
     for (const [i, isValid] of isValidArr.entries()) {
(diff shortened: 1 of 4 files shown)


packages/lodestar/test/e2e/chain/bls/multithread.test.ts
```diff
@@ -88,7 +88,7 @@ describe("chain / bls / multithread queue", function () {
       isValidPromiseArr.push(pool.verifySignatureSets(sets, {batchable: true}));
     }
 
-    await expect(isInvalidPromise).to.rejectedWith("BLST_ERROR");
+    await expect(isInvalidPromise).to.rejectedWith("BLST_INVALID_SIZE");
```
dadepo, line 91: The test was failing because the error is now `BLST_INVALID_SIZE`. 

The test is supposed to assert that _If the first signature is invalid it should not make the rest throw_ which is still true, but I am not sure if the change in the error is a sign of something else being wrong. Any thoughts?

wemeetagain: Not sure how the test was passing before, I believe the `signature` on L84 should be 96 bytes...
So there is an "invalid size" present.

dadepo: Ok. I take that to mean that the upgrade somewhat made the error more specific. Which is not a bad thing...

wemeetagain (approved):

dapplion (approved):

dapplion:
After merging this please attempt to run a full Lodestar client in arm

PR #3844, opened 2022-03-10
Kiln testnet config

g11tech:
**Motivation**
Kiln testnet is due for launch and configs are out!

This PR adds Kiln public testnet configuration in lodestar 
- [x] standalone lodestar config via `./lodestar --network kiln ...`, needs to be paired with a corresponding Kiln testnet complaint EL 
- [x] "single command merge test" script update: now run 
  ```
  cd kiln/devnets
  ./setup.sh --dataDir kiln-data --elClient nethermind --devnetVars ./kiln.vars --dockerWithSudo --withTerminal "gnome-terminal --disable-factory --"
  ```
  change `--elClient nethermind` to `--elClient geth` to run with geth instead!
**Description**


dapplion (approved):
Let's goooo

PR #3854, opened 2022-03-13
Added retry mechanism in executionEngine for executePayload

### LINKED ISSUE #3567: Add retry mechanism in executionEngine for executePayload
Add retry for following scenarios

- Execution Engine is unavailable
  -  Explore all unavailability scenarios and the behavior of the json rpc client we are using
 - Execution Engine errors: Sometimes EL can give internal errors because of the software issues for e.g. nethermind sometimes give internal errors on some concurrency race conditions. Without explicit handling the error basis the error message, a basic retry (like 3 retires) could be done to see if EL can recover.

Basically there are 3 fundamental apis: new payload (retry needed), forkchoice updated (no retry needed without payloadattributes as it tells the EL where out head is, so in 12 seconds it will be called again so no worries, but with payload attributes max tries 3 within 1-2 second interval), and engine_getPayloadV1 again 3 tries with 2 sec interval.

Conclusion: In other to keep things simple, have _3 retries within 1-2 second interval_ for all the endpoints that needs to be retried.

Also add metrics for:

- histogram of retry attempts per call
- histogram of overall requests times to EL

dadepo:
**Motivation**

Adding retries when CL calls EL

Closes #3567


packages/cli/src/options/beaconNodeOptions/execution.ts
```diff
@@ -5,13 +5,17 @@ import {ICliCommandOptions, extractJwtHexSecret} from "../../util";
 export type ExecutionEngineArgs = {
   "execution.urls": string[];
   "execution.timeout": number;
+  "retry.attempts": number;
```
g11tech: I think it would be better if we prefix the name space here, `execution.retryAttempts`, `execution.retryDelay`

packages/cli/src/options/beaconNodeOptions/execution.ts
```diff
@@ -41,4 +45,20 @@ export const options: ICliCommandOptions<ExecutionEngineArgs> = {
...
+  "retry.attempts": {
+    description: "Number of retry attempts when calling execution engine API",
+    type: "number",
+    defaultDescription:
+      defaultOptions.executionEngine.mode === "http" ? String(defaultOptions.executionEngine.retryAttempts) : "1",
+    group: "execution",
+  },
+
+  "retry.delay": {
+    description: "Delay time between retries when retrying calls to the execution engine API",
```
g11tech: "Delay time in ms..." would be a bit more precise

packages/cli/src/options/beaconNodeOptions/execution.ts
```diff
@@ -41,4 +45,20 @@ export const options: ICliCommandOptions<ExecutionEngineArgs> = {
...
+    type: "number",
+    defaultDescription:
+      defaultOptions.executionEngine.mode === "http" ? String(defaultOptions.executionEngine.retryDelay) : "0",
```
g11tech, line 58: do we really need to tailor make it for http? even if there is for example a non http mode ever (like ws) this would stay the same i believe.

packages/lodestar/src/executionEngine/http.ts
```diff
@@ -96,20 +105,29 @@ export class ExecutionEngineHttp implements IExecutionEngine {
...
-        if (e instanceof HttpRpcError || e instanceof ErrorJsonRpcResponse) {
-          return {status: ExecutePayloadStatus.ELERROR, latestValidHash: null, validationError: e.message};
-        } else {
-          return {status: ExecutePayloadStatus.UNAVAILABLE, latestValidHash: null, validationError: e.message};
-        }
-      });
+    const {status, latestValidHash, validationError} = await retry(
+      async (attempt) => {
+        return await this.rpc
+          .fetch<EngineApiRpcReturnTypes[typeof method], EngineApiRpcParamTypes[typeof method]>({
```
g11tech: just thinking aloud, would it be better if we push retry logic in the jsonrpc class? then this would be something like this.rpc.fetchWithRetries, also apart from number of attempts, it would be more flexible if we can pass a callback which takes an error to determine if the error is retry worth or not.

packages/lodestar/test/unit/executionEngine/http.test.ts
```diff
@@ -18,72 +21,90 @@ describe("ExecutionEngine / http", () => {
...
-      reqJsonRpcPayload = req.body;
-      delete (reqJsonRpcPayload as {id?: number}).id;
-      return returnValue;
+      if (errorResponsesBeforeSuccess === 0) {
+        reqJsonRpcPayload = req.body;
+        delete (reqJsonRpcPayload as {id?: number}).id;
+        return returnValue;
+      } else {
+        --errorResponsesBeforeSuccess;
+        throw Error(`Will succeed after ${errorResponsesBeforeSuccess} more attempts`);
```
g11tech: if we have fetchWithRetries as part of jsonrpc class, can we add retries test as separate file? just to test retry logic and not mix it up with the execution specs.

packages/cli/src/options/beaconNodeOptions/execution.ts
dadepo: This was done so as to make it uniform with already existing `execution.urls` and `execution.timeout` options which are tailor made for http.

dapplion: I would just do
```
defaultDescription: String(defaultOptions.executionEngine.retryDelay)
```

packages/lodestar/src/executionEngine/http.ts
```diff
@@ -97,13 +105,19 @@ export class ExecutionEngineHttp implements IExecutionEngine {
...
+        {
+          retryAttempts: this.retryAttempts,
+          retryDelay: this.retryDelay,
+        }
+      )
       // If there are errors by EL like connection refused, internal error, they need to be
-      // treated seperate from being INVALID. For now, just pass the error upstream.
-      .catch((e: Error): EngineApiRpcReturnTypes[typeof method] => {
+      // treated separate from being INVALID. For now, just pass the error upstream.
+      .catch(async (e: Error) => {
```
dapplion: I'm not sure if that's allowed, Can a catch callback handle rejects without causing an unhandled promise?

packages/lodestar/src/executionEngine/http.ts
```diff
@@ -261,13 +285,19 @@ export class ExecutionEngineHttp implements IExecutionEngine {
...
+    >(
+      {
+        method,
+        params: [payloadId],
+      },
+      {
+        retryAttempts: this.retryAttempts,
+        retryDelay: this.retryDelay,
+      }
+    );
```
dapplion: Here it would be very important to track retry behaviour. Meaning adding metrics for:
- histogram of retry attempts per call
- histogram of overall requests times to EL

So to de-duplicate code you can add a private method `fetchEl` that handles metrics and calls `fetchWithRetries`. You can even move the logic in `fetchWithRetries` to here since no-one else is using it.

packages/lodestar/src/eth1/provider/jsonRpcHttpClient.ts
```diff
@@ -75,6 +77,23 @@ export class JsonRpcHttpClient implements IJsonRpcHttpClient {
...
+      async (_attempt) => {
+        return await this.fetchJson({jsonrpc: "2.0", id: this.id++, ...payload}, opts);
+      },
+      {
+        retries: opts?.retryAttempts ?? 1,
+        retryDelay: opts?.retryDelay ?? 0,
+        shouldRetry: opts?.shouldRetry,
+      }
+    );
+    return parseRpcResponse(res, payload);
```
dapplion: The should retry logic here should be based exclusively in networking errors. You only want to retry when the EL is unavailable. Here it's a bit tough because the underlying httpclient can change. So you should add a lot of unit (or e2e) tests that spin up a real server and try different things like:
- bad URL
- bad port
- NGINX rejects to forward
- etc

And ensure that only those are retried. Otherwise you can try to guess when a response is actually from an EL client, and don't retry only when you know the error is an "app layer" EL error

dapplion (changes requested):

packages/lodestar/test/unit/executionEngine/httpRetry.test.ts
```diff
@@ -0,0 +1,282 @@
...
+      controller.signal
+    );
+  });
+
+  describe("getPayload", async () => {
+    it("getPayload with retry", async function () {
+      this.timeout("10 min");
+      /**
+       * curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"engine_getPayload","params":["0x0"],"id":67}' http://localhost:8545
+       */
```
g11tech: you don't need to rewrite all the execution api cases, as it will be difficult to maintain/update any changes in two places, rather just test out `fetchWithRetries` functionality exhaustively against any hypothetical request response (in the same way its happening here) basically write its test cases independentaly of the execution api logic

packages/lodestar/src/eth1/provider/jsonRpcHttpClient.ts
dadepo: > The should retry logic here should be based exclusively in networking errors. You only want to retry when the EL is unavailable

I believe this the case already. Line 94 (`parseRpcResponse`) would only be reached if the call returns 200, and  `parseRpcResponse` parses that and throws `ErrorJsonRpcResponse`. If the call fails for any other reason (if response is not ok, ie network errors) that is only when the request is retried. 

> So you should add a lot of unit (or e2e) tests that spin up a real server 

I'll add these

packages/lodestar/test/unit/executionEngine/httpRetry.test.ts
dadepo: What do you say to removing the tests as you suggested but leaving just:
 
- `notifyForkchoiceUpdate no retry when no pay load attributes` and 
- `notifyForkchoiceUpdate with retry when pay load attributes` tests? 

as this are specific to how the node calls the EL, and when those calls should be retried or not.

g11tech: ok, lets leave it at that, no need to remove :+1:

packages/lodestar/src/executionEngine/http.ts
dapplion: bump

docker/grafana/datasource.local.yml
```diff
@@ -4,6 +4,6 @@ datasources:
   - name: Prometheus
     type: prometheus
     access: proxy
-    url: http://localhost:9090
+    url: http://host.docker.internal:9090
```
dapplion: Why is localhost:9090 not reachable in MacOS?

dadepo: I think that has to do with the fact that docker on MacOS does not have access to the host directly and it sits on top of a linux vm. Which makes it impossible to use `localhost` to refer to the actual host.

dapplion: Then I would move this changes to another PR to keep this one in scope to `dadepo/retry-el-executepayload`

dadepo: This changes was made due to the request to add metrics to the retry mechanism. Since it makes it easier to see the metrics via the Prometheus/Grafana setup locally, instead of having to first push and run on a linux server. 

But I can move it to a separate PR as you suggested and just use the /metrics endpoint exposed by the node

packages/lodestar/src/executionEngine/http.ts
dadepo: > You can even move the logic in fetchWithRetries to here since no-one else is using it.

I decided to leave `fetchWithRetries` as is, because as mentioned by @g11tech [here](https://github.com/ChainSafe/lodestar/pull/3854#pullrequestreview-908101788) it makes it easier to test the retry mechanism separately.

@dapplion can you help clarify what you meant by `histogram of retry attempts per call`? I interpreted that to mean another histogram that captures the duration for each retry, while @g11tech interpreted that to mean a counter to try each request. 

Also I was wondering, this suggestion only focuses on adding metrics to retries to done in `executionEngine/http`, while there are other places the `JsonRpcHttpClient` is used, for example in `eth1Provider.ts`. Would it be an idea to move metrics tracking totally into `JsonRpcHttpClient`?  Then every client who uses  `JsonRpcHttpClient` do not need to have metrics separate, it would be generated automatically from using `JsonRpcHttpClient` to make requests.

If yes, then I can do that as another PR. Let me know what you think

g11tech:
@dadepo @dapplion doing a bit of refac, moving metrics inside jsonrpc, adding dashboard, hopefully things will become cleaner and abstracted out from execution engine view point!

packages/lodestar/src/executionEngine/http.ts
```diff
@@ -113,6 +126,9 @@ export class ExecutionEngineHttp implements IExecutionEngine {
         {
           retryAttempts: this.retryAttempts,
           retryDelay: this.retryDelay,
+          onEachRetryFn: () => {
+            this?.metrics?.executionEngineRequestCount.inc({method});
```
dapplion: this can be undefined?

packages/lodestar/src/executionEngine/http.ts
dapplion: Actually might be good idea to add metrics to JsonRpcHttpClient. eth1 calls to not the execution engine should have metrics too

dapplion: Actually might be good idea to add metrics to JsonRpcHttpClient. eth1 calls to not the execution engine should have metrics too

packages/lodestar/src/executionEngine/http.ts
g11tech: I am removing this because i am moving metrics inside json rpc, also intend is to capture the retry count, not the total count

g11tech:
grafana dashboard cc: @dadepo @dapplion :
image

packages/cli/src/options/beaconNodeOptions/execution.ts
g11tech: i think typescript isn't happy otherwise, mixes it up with mock options, so i guess this is alright!

dadepo:
> grafana dashboard cc: @dadepo @dapplion : image

Graph looks good. Just noticed that `engine_getPayloadV1` is missing from the screenshot. Is it because it is actually missing or it was not just part of the screenshoot

g11tech:
> > grafana dashboard cc: @dadepo @dapplion : image
> 
> Graph looks good. Just noticed that `engine_getPayloadV1` is missing from the screenshot. Is it because it is actually missing or it was not just part of the screenshoot

That method is called when the validator builds and proposes, this is just from my local setup without validator on the kiln network

g11tech (dismissed):

dadepo:
> @dadepo @dapplion doing a bit of refac, moving metrics inside jsonrpc, adding dashboard, hopefully things will become cleaner and abstracted out from execution engine view point!

Hi @dapplion. Is this PR fine by you and can be merged? Or there are things you still will like improved?

packages/lodestar/src/util/retry.ts
```diff
@@ -28,7 +28,7 @@ export async function retry<A>(fn: (attempt: number) => A | Promise<A>, opts?: I
   const shouldRetry = opts?.shouldRetry;
 
   let lastError: Error = Error("RetryError");
-  for (let i = 1; i <= maxRetries; i++) {
+  for (let i = 0; i < maxRetries; i++) {
```
dapplion: Warning! This change can break all existing usages of retry downstream! Please review carefully

dapplion (changes requested):

dapplion:
> Hi @dapplion. Is this PR fine by you and can be merged? Or there are things you still will like improved?

Overall looks good! Some changes I've done
- merged the metrics definitions into lodestar file, no need to keep them separate
- de-duplicate test code

Check the issue with [packages/lodestar/src/util/retry.ts](https://github.com/ChainSafe/lodestar/pull/3854/files/52d8600b224e7017c4d73d3e8cd0889aea0918bb#diff-7100b73f62366510d2e962dd00fc207a53a4ac300e5463e10baec29a5537f8e7) changes and should be good to go.

If I've broken the tests please I appreciate if you can review them :pray:

packages/lodestar/src/util/retry.ts
dadepo: Hi @g11tech can you help look into this? Given you made the modifications, it is probably faster to share any insights you had explaining the reason for the change. As far as I can see, everything still looks fine.

g11tech: hey @dadepo  I think i did it to not count first attempt as retry, you can revert it and may be correct a test that might fail because of retry comparison.

wemeetagain (approved):
LGTM

PR #3858, opened 2022-03-16
Bump to new gossipsub

dapplion:
**Motivation**

Adds https://github.com/ChainSafe/js-libp2p-gossipsub/pull/206 to Lodestar

**Description**

TBD


twoeths:
The test in hetzner-test0 has been good:
+ Most gossip peer scores are positive
Screen Shot 2022-03-31 at 10 47 35

+ It helps us get a very stable mesh peers
Screen Shot 2022-03-31 at 10 48 42

+ Validator monitor metrics are almost the same: balance delta, missed attestations ratio

+ Gossip validation time is slightly increased  but it does not affect the gossip score

+ Gossip block processed time is slightly increased, however "till received" is slightly better so overall "till processed" is the same

Some down side monitored on contabo nodes: the better gossip scores we have, the more mesh peers we have and the more gossip messages we receive and it causes performance issue (since messages have to get thru a lot of middle libraries before it reaches lodestar). So contabo can run stably only if:
+ We tweak gossipsub Dhi, D, Dlo to be 4, 3, 2 (default is 12, 8 , 6) (as in contabo-20)
+ Or we reduce connected validators in order to join less subnet topics, if connected validators is 20 it works fine, but with 30 validators it does not (as in contabo-19)

dapplion:
> and it causes performance issue (since messages have to get thru a lot of middle libraries before it reaches lodestar)

Can you post a CPU profile and detail where are the bottlenecks exactly?

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -113,6 +113,7 @@ export class Eth2Gossipsub extends Gossipsub {
       dataTransform: new DataTransformSnappy(gossipTopicCache),
       metricsRegister: modules.metrics ? ((modules.metrics.register as unknown) as MetricsRegister) : null,
       metricsTopicStrToLabel: modules.metrics ? getMetricsTopicStrToLabel(modules.config) : undefined,
+      asyncValidation: true,
```
dapplion, line 117: Oh lol

twoeths:
> > and it causes performance issue (since messages have to get thru a lot of middle libraries before it reaches lodestar)
> 
> Can you post a CPU profile and detail where are the bottlenecks exactly?

with latest update, our contabo node can handle default gossipsub params (D params of 6, 8, 12) and 50 validators

twoeths (approved):

PR #3859, opened 2022-03-16
Cleanup test nodes on success and error

dapplion:
**Motivation**

e2e or sim tests that run nodes are not cleaning up properly.

current code
```ts
const node = await start()
await runTest(node) // may throw
await node.cleanup()
```

If the test fails resources are not cleanup.

This PR does
```ts
const node = await start()
afterEachCallbacks.push(() => node.cleanup())
await runTest(node) // may throw
```

So whether the test succeeds or fails, clean up step is always called

**Description**

- Cleanup test nodes on success and error


wemeetagain:
https://github.com/ChainSafe/lodestar/runs/5565670372?check_suite_focus=true#step:9:400
`Error: listen EADDRINUSE: address already in use 127.0.0.1:10000`

Seems like its not shutting down nodes in time for the next test?

dapplion:
- Taken over by https://github.com/ChainSafe/lodestar/issues/4275

**Do not delete branch**

PR #3860, opened 2022-03-17
Return peer count

### LINKED ISSUE #3857: Beacon API peer_count call always returns 0
While using the beacon API to get some data from the client, always peer_count call returns 0

curl localhost:9596/eth/v1/node/peer_count
{"data":{"disconnected":0,"connecting":0,"connected":0,"disconnecting":0}}

It would be expected to receive the current amount of peers in each state
curl localhost:9596/eth/v1/node/peer_count
{"data":{"disconnected":10,"connecting":9,"connected":30,"disconnecting":2}}

Running Lodestar with Docker, version 0.34.0

**Screenshots**

- OS: Ubuntu 
- Version: 21.04 
- Image: chainsafe/lodestar:v0.34.0

dadepo:
**Motivation**

Implements the endpoint to get the peer count which was not implemented before

Closes #3857 

`curl localhost:9596/eth/v1/node/peer_count` should not return just 0 for all values

### DIFF
--- a/packages/lodestar/src/api/impl/node/index.ts
+++ b/packages/lodestar/src/api/impl/node/index.ts
@@ -1,6 +1,6 @@
 import {routes} from "@chainsafe/lodestar-api";
 import {createKeypairFromPeerId} from "@chainsafe/discv5";
-import {formatNodePeer} from "./utils";
+import {formatNodePeer, getRevelantConnection} from "./utils";
 import {ApiError} from "../errors";
 import {ApiModules} from "../types";
 import {IApiOptions} from "../../options";
@@ -57,22 +57,22 @@ export function getNodeApi(opts: IApiOptions, {network, sync}: Pick<ApiModules, 
       let disconnecting = 0;
 
       for (const connections of network.getConnectionsByPeer().values()) {
-        for (const connection of connections) {
-          switch (connection.stat.status) {
-            case "open":
-              connected++;
-              break;
-            case "closing":
-              disconnecting++;
-              break;
-            case "closed":
-              disconnected++;
-              break;
-            default:
-              connecting++;
-          }
+        const relevantConnection = getRevelantConnection(connections);
+        switch (relevantConnection?.stat.status) {
+          case "open":
+            connected++;
+            break;
+          case "closing":
+            disconnecting++;
+            break;
+          case "closed":
+            disconnected++;
+            break;
+          default:
+            connecting++;
         }
       }
+
       return {
         data: {
           disconnected,
--- a/packages/lodestar/src/api/impl/node/index.ts
+++ b/packages/lodestar/src/api/impl/node/index.ts
@@ -51,6 +51,7 @@ export function getNodeApi(opts: IApiOptions, {network, sync}: Pick<ApiModules, 
     },
 
     async getPeerCount() {
+      // TODO: Implement disconnect count with on-disk persistence
       let disconnected = 0;
       let connecting = 0;
       let connected = 0;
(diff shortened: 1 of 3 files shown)


dadepo:
Trying the same endpoint on lighthouse after running for about 5 minutes returns count for both connected and disconnected, while for lodestar, the endpoint only returns count for connected and nothing for disconnected or the other status. Still need to see if the current way is the most accurate way of getting the connection info from libp2p.

packages/lodestar/src/api/impl/node/index.ts
```diff
@@ -51,13 +51,34 @@ export function getNodeApi(opts: IApiOptions, {network, sync}: Pick<ApiModules,
...
+              connected++;
+              break;
+            case "closing":
+              disconnecting++;
+              break;
+            case "closed":
+              disconnected++;
+              break;
+            default:
+              connecting++;
```
dapplion: So if we have 1 peer with 20 open connections it will return `connected: 20`, is this the expected behavior?

dadepo: updated to use the most relevant connection for a peer in the counting

dadepo:
> Trying the same endpoint on lighthouse after running for about 5 minutes returns count for both connected and disconnected, while for lodestar, the endpoint only returns count for connected and nothing for disconnected or the other status. Still need to see if the current way is the most accurate way of getting the connection info from libp2p.

Looked into trying to get disconnected peer by making a diff with the peers found in the peerstore and the one returned by the connection manager (reasoning is if it is found in the peerstore but not in the connection manager, then those are disconnected). The issue with that is the peerstore persists across restarts, meaning the result of the diff would keep on increasing.

Leaving the implementation for now to use the connection manager for all the counting. Open to other suggestions, if that is not fit enough

wemeetagain (dismissed):
LGTM, I think its fine as is to just use the connection manager.

packages/lodestar/src/api/impl/node/index.ts
```diff
@@ -51,13 +51,34 @@ export function getNodeApi(opts: IApiOptions, {network, sync}: Pick<ApiModules,
     },
 
     async getPeerCount() {
-      // TODO: Implement
+      let disconnected = 0;
```
dapplion, line 55: So disconnected means all peers that have ever connected to us and are no longer connected? That sounds like a memory leak to track lol. How does Lighthouse track this?

dadepo: Yup. Lighthouse has a `network_globals` object that tracks this. So when a peer is disconnected, it is tracked, as can be seen here [here](https://github.com/sigp/lighthouse/blob/9bc9527998581deb5fd8674052bb2083b640bbbd/beacon_node/lighthouse_network/src/peer_manager/mod.rs#L714) which is then read and used to return the peer count as can be seen [here](https://github.com/sigp/lighthouse/blob/9bc9527998581deb5fd8674052bb2083b640bbbd/beacon_node/http_api/src/lib.rs#L1811) only that the `network_globals` is not persistent.

dapplion: It's persisted to disk that's how it doesn't leak memory. I would leave a `// TODO: Implement disconnect count with on-disk persistence`

dadepo: I assumed it is not being persisted as the values don't persist after each restarts. Will dig later to see how that is implemented... 

>  I would leave a // TODO: Implement disconnect count with on-disk persistence

Ok. Will do.

dapplion: > Will dig later to see how that is implemented...

I would ignore completely, tracking disconnected peers is an extremely unnecessary feature in my opinion haha

wemeetagain (approved):

PR #3868, opened 2022-03-23
Support running metrics via (grafana/prometheus) on macos

dadepo:
**Motivation**

The setup to run the metrics infra (grafana/prometheus) locally currently only works on Linux due to differences between docker on Linux and MacoS. This changes makes it possible to also vie metrics via (grafana/prometheus) on MacOS.

### DIFF
--- a/docker/grafana/datasource.local.yml
+++ b/docker/grafana/datasource.local.yml
@@ -4,6 +4,6 @@ datasources:
   - name: Prometheus
     type: prometheus
     access: proxy
-    url: http://localhost:9090
+    url: http://host.docker.internal:9090
     uid: prometheus_local
     isDefault: true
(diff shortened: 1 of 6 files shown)


docker/grafana/datasource.local.yml
```diff
@@ -4,6 +4,6 @@ datasources:
   - name: Prometheus
     type: prometheus
     access: proxy
-    url: http://localhost:9090
+    url: http://host.docker.internal:9090
```
dapplion: Duplicate these files or you'll break the linux setup

dadepo: I believe `host.docker.internal` is also supported on Linux. Is it broken at your end?

dapplion: No, doesn't work for me

```
$ curl host.docker.internal:80
curl: (6) Could not resolve host: host.docker.internal
```

dadepo: > No, doesn't work for me

Dang! :/

I have updated with separate config files for mac

dapplion (approved):

g11tech (approved):

PR #3875, opened 2022-03-28
Compute score using lodestar score and gossipsub score

twoeths:
**Motivation**

Some gossipsub peers' behavior is so bad that we'd like to disconnect them before others

**Description**

+ Compute final score using both lodestar score and gossipsub score
+ Lodestar score is decayed periodically
+ Pull scores from gossipsub periodically to update `PeerRpcScoreStore`
+ When we disconnect peers we choose peers with worse score (already existing feature), so this helps disconnect bad behavior gossip peers

### DIFF
--- a/packages/lodestar/package.json
+++ b/packages/lodestar/package.json
@@ -96,7 +96,7 @@
     "jwt-simple": "0.5.6",
     "libp2p": "^0.36.2",
     "libp2p-bootstrap": "^0.14.0",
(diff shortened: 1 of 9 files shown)


twoeths:
@dapplion I merged this PR as the test in contabo-20 went well, feel free to revise if any

dapplion:
> @dapplion I merged this PR as the test in contabo-20 went well, feel free to revise if any

Looks good! :+1: thank you!

PR #3880, opened 2022-04-03
Implement standard keymanager API - review

dapplion:
**Motivation**

Post-merge review of https://github.com/ChainSafe/lodestar/pull/3522

**Description**

Fixes some potential issues:
- Improve CLI options, if keymanager is enabled require importKeystorePath
- [Explain KeymanagerServer opts merging](https://github.com/ChainSafe/lodestar/commit/f11081e69eefdb875fc593aa0355462df01199e5)
- [Ensure importKeystoresPath is a directory](https://github.com/ChainSafe/lodestar/commit/87d9bfb4725d9036802accdb643c3f6c3a342f81)
- [Parse slashingProtectionStr to JSON](https://github.com/ChainSafe/lodestar/commit/77c7fb183fce9b1bf93cb1732572e2afaf6ee307) @dadepo how could this work before? If the import should require a JSON
- [Improve keystore lock logic](https://github.com/ChainSafe/lodestar/commit/1b92d71af1c797261e6c9c8d81f0d6d8208bc762) define the path logic concatenation in a single place

### DIFF
--- a/packages/keymanager-server/src/impl.ts
+++ b/packages/keymanager-server/src/impl.ts
@@ -60,9 +60,9 @@ export class KeymanagerApi implements Api {
    * Users SHOULD send slashing_protection data associated with the imported pubkeys. MUST follow the format defined in
    * EIP-3076: Slashing Protection Interchange Format.
    *
-   * @param keystores JSON-encoded keystore files generated with the Launchpad
+   * @param keystoresStr JSON-encoded keystore files generated with the Launchpad
    * @param passwords Passwords to unlock imported keystore files. `passwords[i]` must unlock `keystores[i]`
-   * @param slashingProtection Slashing protection data for some of the keys of `keystores`
+   * @param slashingProtectionStr Slashing protection data for some of the keys of `keystores`
    * @returns Status result of each `request.keystores` with same length and order of `request.keystores`
    *
    * https://github.com/ethereum/keymanager-APIs/blob/0c975dae2ac6053c8245ebdb6a9f27c2f114f407/keymanager-oapi.yaml
@@ -77,7 +77,7 @@ export class KeymanagerApi implements Api {
       message?: string;
     }[];
   }> {
-    const interchange = (slashingProtectionStr as unknown) as Interchange;
+    const interchange = JSON.parse(slashingProtectionStr) as Interchange;
     await this.validator.validatorStore.importInterchange(interchange);
 
     const statuses: {status: ImportStatus; message?: string}[] = [];
@@ -134,7 +134,7 @@ export class KeymanagerApi implements Api {
    * Slashing protection data must only be returned for keys from `request.pubkeys` for which a
    * `deleted` or `not_active` status is returned.
    *
-   * @param pubkeys List of public keys to delete.
+   * @param pubkeysHex List of public keys to delete.
    * @returns Deletion status of all keys in `request.pubkeys` in the same order.
    *
    * https://github.com/ethereum/keymanager-APIs/blob/0c975dae2ac6053c8245ebdb6a9f27c2f114f407/keymanager-oapi.yaml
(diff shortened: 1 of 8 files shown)


packages/keymanager-server/src/impl.ts
```diff
@@ -80,7 +76,7 @@ export class KeymanagerApi implements Api {
       message?: string;
     }[];
   }> {
-    const interchange = (slashingProtectionStr as unknown) as Interchange;
+    const interchange = JSON.parse(slashingProtectionStr) as Interchange;
```
dapplion: @dadepo do you know why JSON.parse() was not necessary if `slashingProtectionStr` is supposed to be a string?

packages/keymanager-server/src/impl.ts
```diff
@@ -172,14 +169,14 @@ export class KeymanagerApi implements Api {
...
+
+        const keystoreFilepath = this.getKeystoreFilepath(pubkeyHex);
+
         // Remove key from persistent storage
-        for (const keystoreFile of await fs.promises.readdir(this.importKeystoresPath)) {
-          if (keystoreFile.indexOf(pubkeyHex) !== -1) {
-            await fs.promises.unlink(keystorePathInfo.keystoreFilePath);
-            await fs.promises.unlink(keystorePathInfo.lockFilePath);
-          }
-        }
```
dapplion, line 182: @dadepo Why was it necessary to loop over all files in `importKeystoresPath`? The keystore file path should be predictable

packages/keymanager-server/src/impl.ts
dadepo: That's because the arguments to the function is passed in as an http body and fastify parses it already, so by the time our code calls finds the relevant handler  [here](https://github.com/ChainSafe/lodestar/blob/d61d898d0dbcde20335c4657137feb7014f9b2ef/packages/api/src/server/utils/server.ts#L60) it is already an object hence no need to use JSON.parse within the function. 

I also think the types safety of the function is weaker as it is decoupled from the calling logic in [here](https://github.com/ChainSafe/lodestar/blob/d61d898d0dbcde20335c4657137feb7014f9b2ef/packages/api/src/server/utils/server.ts#L60) so at runtime, it is an object that is passed in, which is then attempted to be parsed into JSON

My guess is that If it were a query path or params, then the parsing will be needed.

packages/keymanager-server/src/impl.ts
dadepo: I think the loop was from an initial implementation that did not use a stable path/filename. So indeed a loop is not necessary

packages/keymanager-server/src/impl.ts
dapplion: So can you add a nice long comment explaining this above my change and do conditional parsing?

```ts
const interchange: Interchange = typeof slashingProtectionStr === "string" ? JSON.parse(slashingProtectionStr) : slashingProtectionStr;
```

dapplion:
@dadepo Please if you can fix the test I broken and approve if it looks good then :pray:

packages/keymanager-server/src/impl.ts
dadepo: I added a comment but I feel that the conditional parsing is not necessary since the endpoint is going to be called with the payload within the body of the request. 

Also the snippet you shared won't work. What will work will require changing the type for `slashingProtectionStr: SlashingProtectionData | Record<string, unknown>` and then have the conditional parsing as follows:

```
    const interchange = typeof slashingProtectionStr === "string"
        ? (JSON.parse(slashingProtectionStr) as Interchange)
        : ((slashingProtectionStr as unknown) as Interchange);
```

Which I think is a bit too much since payload will always come within the body

dadepo:
> @dadepo Please if you can fix the test I broken and approve if it looks good then 🙏

Tests are now passing!

dapplion (commented):
Meta-Approve :heavy_check_mark: @dadepo it's my PR so you or someone else has to approve

g11tech (approved):

wemeetagain (approved):

Issue #3687, opened 2022-02-01
Rocketpool Integration Tracker


philknows:
This epic is to monitor/track outstanding issues and PRs relating to the RocketPool integration for Lodestar:

- Doppelgänger Protection #3689
- #3746
- #3692

philknows:
Implemented in [Rocketpool v1.7.0 beta](https://github.com/rocket-pool/smartnode-install/releases/tag/v1.7.0-b1) for testnet. Will keep open for tracking any issues related to seeing it move towards production after EF endorsement of Lodestar.

philknows:
Now integrated and published here: https://hub.docker.com/r/rocketpool/lodestar/tags
Will be added to the formal mainnet release schedule starting from release `v1.4.0`.

PR #3883, opened 2022-04-04
Add doppelganger support

### LINKED ISSUE #3689: Doppelgänger Protection Feature
Feature Request as part of #3687 RocketPool Integration and safety for validators.

Reference implementation from Lighthouse: https://lighthouse-book.sigmaprime.io/validator-doppelganger.html

Miss 2 attestations upon startup and listen for attestations coming from the loaded keys, disable validation duties if found because it would be a slashable offense to double attest.

- [x] Expose liveness API inspired by [Lighthouse's implementation](https://lighthouse-book.sigmaprime.io/api-lighthouse.html#lighthouseliveness)
- [x] Add the doppelganger service which polls BN's liveness API and keeps info about validators
- [x] Use the doppelganger in validator and check its doppelganger state before performing any slash-able action
- [x] Reasonable test coverage
- [x] Add metrics

dadepo:
**Motivation**

Part of https://github.com/ChainSafe/lodestar/issues/3687 RocketPool Integration and safety for validators

**Description**

Miss 2 attestations upon startup and listen for attestations coming from the loaded keys, disable validation duties if found because it would be a slashable offense to double attest

Closes #3689


packages/api/src/routes/lodestar.ts
```diff
@@ -83,6 +89,9 @@ export type Api = {
 
   /** Dump Discv5 Kad values */
   discv5GetKadValues(): Promise<{data: string[]}>;
+
+  /** Returns validator indices that has been observed to be active on the network */
+  getLiveness(indices: ValidatorIndex[], epoch: Epoch): Promise<{data: LivenessResponseData[]}>;
```
dapplion: The namespace lodestar is not enabled by default and should never be. This routes are dangerous and only meant for debugging. getLiveness should be moved to the validator namespace in a lack of a better place

packages/cli/src/cmds/dev/options.ts
```diff
@@ -58,6 +60,18 @@ const devOwnOptions: ICliCommandOptions<IDevOwnArgs> = {
...
+    description: "Enables Doppelganger protection",
+    defaultDescription: "false",
+    type: "boolean",
+  },
+
+  doppelgangerEpochsToCheck: {
+    description: "Number of epoch the doppelganger protection should miss validator duties for",
+    default: 3,
+    type: "number",
+  },
```
dapplion: Duplicated code, please declare only once

packages/cli/src/cmds/validator/options.ts
```diff
@@ -89,6 +91,19 @@ export const validatorOptions: ICliCommandOptions<IValidatorCliArgs> = {
...
+  enableDoppelganger: {
+    description: "Enables Doppelganger protection",
+    defaultDescription: "false",
+    default: false,
+    type: "boolean",
+  },
+
+  doppelgangerEpochsToCheck: {
+    description: "Number of epoch the doppelganger protection should miss validator duties for",
+    default: 3,
```
dapplion: What's the reasoning for this value? Whatever it is please justify it in detail and commit the reasoning in the code as a comment

packages/lodestar/src/api/impl/lodestar/index.ts
```diff
@@ -181,9 +187,52 @@ export function getLodestarApi({
...
       };
     },
+
+    async getLiveness(indices: ValidatorIndex[], epoch: Epoch): Promise<{data: LivenessResponseData[]}> {
+      if (indices.length === 0) {
+        return {
+          data: [],
+        };
+      }
+      const state = await chain.getHeadStateAtCurrentEpoch();
```
dapplion: To get the current epoch, just use the clock, no need to get the entire state

packages/lodestar/src/api/impl/lodestar/index.ts
```diff
@@ -181,9 +187,52 @@ export function getLodestarApi({
...
+    if (chain.seenBlockProposers.isKnown(slot, index)) {
+      proposedBlock = true;
+      break;
+    }
+  }
+
+  const hasAggregated = chain.seenAggregators.isKnown(epoch, index);
+  const hasAttested = chain.seenAttesters.isKnown(epoch, index);
+
+  return proposedBlock || hasAggregated || hasAttested;
```
dapplion: This checks only detects:
- attestations from gossip. If we are not subscribed to that subnet we will miss the attestation
- aggregates from gossip. Only tracks the aggregator, not all participants
- blocks from gossip: very rare

This checks are insufficient and do not cover the most likely ways to detect validators in the network:
- Inclusion in some else's aggregate, received via gossip and received via a block

Lighthouse covers all possible ways to detect it https://github.com/sigp/lighthouse/blob/aa72088f8fc91d41106a8afce7a0179cde64ce5d/beacon_node/beacon_chain/src/beacon_chain.rs#L4678

This PR from them discusses in depth the strategies, tradeoffs and risks, it's a great read for context. https://github.com/sigp/lighthouse/pull/2230

packages/lodestar/src/chain/chain.ts
```diff
@@ -242,7 +242,8 @@ export class BeaconChain implements IBeaconChain {
     const currentEpochStartSlot = computeStartSlotAtEpoch(this.clock.currentEpoch);
     const head = this.forkChoice.getHead();
     const bestSlot = currentEpochStartSlot > head.slot ? currentEpochStartSlot : head.slot;
-    return await this.regen.getBlockSlotState(head.blockRoot, bestSlot, RegenCaller.getDuties);
+    const beaconState = await this.regen.getBlockSlotState(head.blockRoot, bestSlot, RegenCaller.getDuties);
+    return beaconState;
```
dapplion: Why this change?

packages/lodestar/src/chain/regen/queued.ts
```diff
@@ -108,7 +108,8 @@ export class QueuedStateRegenerator implements IStateRegenerator {
     this.metrics?.regenFnCallTotal.inc({caller: rCaller, entrypoint: RegenFnName.getBlockSlotState});
 
     // The state is not immediately available in the caches, enqueue the job
-    return this.jobQueue.push({key: "getBlockSlotState", args: [blockRoot, slot, rCaller]});
+    const promise = this.jobQueue.push({key: "getBlockSlotState", args: [blockRoot, slot, rCaller]});
+    return promise;
```
dapplion: Why this change?

packages/lodestar/test/e2e/api/lodestar/lodestar.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      bn.chain.seenBlockProposers.add(bn.chain.clock.currentSlot, 1);
+
+      const client = getClient(config, new HttpClient({baseUrl: `http://127.0.0.1:${restPort}`}));
+
+      const currentEpoch = bn.chain.clock.currentEpoch;
+      const nextEpoch = currentEpoch + 1;
+      const previousEpoch = currentEpoch - 1;
+
+      // current epoch is fine
+      await expect(client.getLiveness([1], currentEpoch)).to.eventually.not.be.rejected;
```
dapplion: The `eventually` is required or just there for readability?

packages/lodestar/test/e2e/api/lodestar/lodestar.test.ts
```diff
@@ -0,0 +1,135 @@
...
+import {LogLevel, testLogger, TestLoggerOpts} from "../../../utils/logger";
+import {getDevBeaconNode} from "../../../utils/node/beacon";
+import {waitForEvent} from "../../../utils/events/resolver";
+import {ChainEvent} from "../../../../src/chain";
+import {SLOTS_PER_EPOCH} from "@chainsafe/lodestar-params";
+
+chai.use(chaiAsPromised);
+
+describe("api / impl / lodestar", function () {
+  describe("getLiveness endpoint", function () {
```
dapplion, line 16: Thank you for not using stubs :pray:

packages/lodestar/test/e2e/doppelganger/doppelganger.test.ts
```diff
@@ -0,0 +1,301 @@
...
+      if (callback) await callback();
+    }
+  });
+
+  const validatorCount = 1;
+  const SECONDS_PER_SLOT = 2;
+  const genesisSlotsDelay = 5;
+  const beaconParams: Pick<IChainConfig, "SECONDS_PER_SLOT"> = {
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    SECONDS_PER_SLOT: SECONDS_PER_SLOT,
```
dapplion: You can just do `SECONDS_PER_SLOT: 2` as you reference `beaconParams.SECONDS_PER_SLOT` below

packages/lodestar/test/e2e/doppelganger/doppelganger.test.ts
```diff
@@ -0,0 +1,301 @@
...
+      useRestApi: false,
+      testLoggerOpts,
+    });
+    afterEachCallbacks.push(() => Promise.all(validators.map((v) => v.stop())));
+
+    await connect(bn2.network as Network, bn.network.peerId, bn.network.localMultiaddrs);
+
+    await Promise.all([...validatorsWithDoppelganger, ...validators].map((validator) => validator.start()));
+
+    expect(validators[0].getStatus()).to.be.equals(
```
dapplion: To be more correct english I guess `to.equal` or `to.be.equal` would make more sense than `to.be.equals` right?

packages/lodestar/src/api/impl/lodestar/index.ts
```diff
@@ -181,9 +187,52 @@ export function getLodestarApi({
...
+
+  let proposedBlock = false;
+  for (let slot = startSlot; slot <= endSlot; slot++) {
+    if (chain.seenBlockProposers.isKnown(slot, index)) {
+      proposedBlock = true;
+      break;
+    }
+  }
+
+  const hasAggregated = chain.seenAggregators.isKnown(epoch, index);
```
dapplion: You should be very careful of false negatives. The `seenAggregators` cache prunes every epoch so I contains limited data.
```ts
  isKnown(targetEpoch: Epoch, validatorIndex: ValidatorIndex): boolean {
    return this.validatorIndexesByEpoch.get(targetEpoch)?.has(validatorIndex) === true;
  }
```
What we really want (in pseudo code) is:
```ts
  isKnown(targetEpoch: Epoch, validatorIndex: ValidatorIndex): boolean {
    const validatorIndexes = this.validatorIndexesByEpoch.get(targetEpoch)
    if (validatorIndexes) {
      return validatorIndexes.has(validatorIndex)
    } else {
      throw "=( i don't have this data anymore, not sure if seen or not"
    }
  }
```

dapplion (changes requested):
The current checks are not sufficient, more surface must be covered to provide more safety, see https://github.com/ChainSafe/lodestar/pull/3883#discussion_r848351293

See also a discussion in Nimbus for a wait-free doppelganger https://github.com/status-im/nimbus-eth2/issues/3083 you should also consider genesis. Really nice discussion in Prysm about it too https://github.com/prysmaticlabs/prysm/issues/7985#issuecomment-843209422

@dadepo before starting to implement code, let's come to a consensus here on what strategies to implement

packages/validator/src/services/doppelgangerService.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export enum DoppelgangerStatus {
+  VerifiedSafe = "VerifiedSafe",
+  Unverified = "Unverified",
+  Unknown = "Unknown",
+}
+
+export class DoppelgangerService {
+  private readonly doppelgangerStateByIndex = new Map<ValidatorIndex, DoppelgangerState>();
+  constructor(
```
twoeths, line 40: Some metrics we can add:
+ Statuses of DoppelGanger for each validator
+ Time to run DoppelGanger service for each epoch

that's not necessarily to be included in this PR, just something to think about

packages/validator/src/validator.ts
```diff
@@ -78,7 +82,24 @@ export class Validator {
 
     const clock = new Clock(config, logger, {genesisTime: Number(genesis.genesisTime)});
     const validatorStore = new ValidatorStore(config, slashingProtection, metrics, signers, genesis);
-    this.indicesService = new IndicesService(logger, api, validatorStore, metrics);
+    const indicesService = new IndicesService(logger, api, validatorStore, metrics);
+    // Do not enable doppelganger when started before genesis or first slot of genesis, because
```
dadepo: Maybe skip the whole first epoch? Instead of first slot? If a validator has duties in a slot (say the first slot) then per the spec, the validator should not have any other duties the reminder of that epoch? so the first epoch should actually be skipped

packages/lodestar/src/api/impl/lodestar/index.ts
dadepo: I looked into the pruning implementation of the seenAggregators and I see it only prunes up epoch less than `lowestPermissibleEpoch` which is basically epoch lesser than the current epoch - 3. Looks like this is also how Lighthouse is implemented. The `ObservedAttesters` data structures are `AutoPruningEpochContainer` container with similar pruning logic

packages/cli/src/options/doppelgangerOptions.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const doppelgangerOptions: ICliCommandOptions<DoppelgangerOps> = {
+  enableDoppelganger: {
+    description: "Enables Doppelganger protection",
+    defaultDescription: "false",
+    default: false,
+    type: "boolean",
+  },
+
+  doppelgangerEpochsToCheck: {
```
dadepo: consider removing this option (or have it hidden?) FYI LH does not have it

packages/validator/src/services/doppelgangerService.ts
```diff
@@ -0,0 +1,179 @@
...
+    if (indices.length !== 0) {
+      const previousEpoch = currentEpoch - 1;
+      // in the current epoch also request for liveness check for past epoch in case a validator index was live
+      // in the remaining 25% of the last slot of the previous epoch
+      const previous =
+        previousEpoch >= 0
+          ? (
+              await this.api.validator.getLiveness(indices, previousEpoch).catch((e) => {
+                this.logger.error(`Error getting liveness data for previous epoch ${previousEpoch}`, {}, e as Error);
+                return Promise.resolve({data: []});
```
dapplion: Why return a Promise.resolve instead of just `{data: []}`?

packages/cli/src/cmds/dev/options.ts
```diff
@@ -4,19 +4,23 @@ import {beaconOptions, IBeaconArgs} from "../beacon/options";
...
+import {doppelgangerOptions} from "../../options/doppelgangerOptions";
 
 type IDevOwnArgs = {
   genesisEth1Hash?: string;
   genesisValidators?: number;
   startValidators?: string;
   genesisTime?: number;
   reset?: boolean;
+  enableDoppelganger: boolean;
+  doppelgangerEpochsToCheck: number;
```
dapplion: This options should be optional

packages/cli/src/cmds/validator/options.ts
```diff
@@ -24,6 +25,8 @@ export type IValidatorCliArgs = IAccountValidatorArgs &
     externalSignerUrl?: string;
     externalSignerPublicKeys?: string[];
     externalSignerFetchPubkeys?: boolean;
+    enableDoppelganger: boolean;
+    doppelgangerEpochsToCheck: number;
```
dapplion: This options should be optional

packages/cli/src/cmds/account/cmds/validator/voluntaryExit.ts
```diff
@@ -111,6 +111,8 @@ BE UNTIL AT LEAST TWO YEARS AFTER THE PHASE 0 MAINNET LAUNCH.
       signers: [{type: SignerType.Local, secretKey}],
       logger: errorLogger(),
       graffiti: args.graffiti,
+      enableDoppelganger: args.enableDoppelganger,
+      doppelgangerEpochsToCheck: args.doppelgangerEpochsToCheck,
```
dapplion: voluntaryExit should not care about this options, remove once optional

packages/cli/src/options/doppelgangerOptions.ts
dapplion: We should remove this option, the epochs to check must be a researched constant that we client devs provide

packages/cli/src/options/doppelgangerOptions.ts
```diff
@@ -0,0 +1,21 @@
...
+type DoppelgangerOps = {
+  enableDoppelganger?: boolean;
+  doppelgangerEpochsToCheck?: number;
+};
+
+export const doppelgangerOptions: ICliCommandOptions<DoppelgangerOps> = {
+  enableDoppelganger: {
+    description: "Enables Doppelganger protection",
+    defaultDescription: "false",
+    default: false,
```
dapplion: No need to provide both `defaultDescription` and `default`, check yargs docs

packages/lodestar/src/chain/blocks/importBlock.ts
```diff
@@ -71,12 +71,10 @@ export type ImportBlockModules = {
  * - head_tracker.register_block(block_root, parent_root, slot)
  * - Send events after everything is done
  */
-export async function importBlock(chain: ImportBlockModules, fullyVerifiedBlock: FullyVerifiedBlock): Promise<void> {
+export async function importBlock(modules: ImportBlockModules, fullyVerifiedBlock: FullyVerifiedBlock): Promise<void> {
```
dapplion: Why rename `chain` to `modules`? If you believe this is an important change it should be done in another PR and keep this diff as small as possible

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -580,5 +584,47 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
...
+  let proposedBlock = false;
+  for (let slot = startSlot; slot <= endSlot; slot++) {
+    if (chain.seenBlockProposers.isKnown(slot, index)) {
+      proposedBlock = true;
+      break;
+    }
+  }
+
+  const hasAggregated = chain.seenAggregators.isKnown(epoch, index);
+  const hasAttested = chain.seenAttesters.isKnown(epoch, index);
```
dapplion, line 627: This is still insufficient. You are missing block attesters

dapplion: This is still insufficient. You are missing block attesters. Check Lighthouse impl for reference

packages/lodestar/src/chain/blocks/importBlock.ts
```diff
@@ -93,41 +91,41 @@ export async function importBlock(chain: ImportBlockModules, fullyVerifiedBlock:
   // it should always have epochBalances there bc it's a checkpoint state, ie got through processEpoch
   const justifiedCheckpoint = postState.currentJustifiedCheckpoint;
 
-  const onBlockPrecachedData: OnBlockPrecachedData = {
+  const onBlockPreCachedData: OnBlockPrecachedData = {
```
dapplion: Please don't add non-critical changes to the PR, do casing changes in another PR

packages/lodestar/src/chain/eventHandlers.ts
```diff
@@ -187,6 +187,8 @@ export async function onBlock(
     slot: block.message.slot,
     root: blockRoot,
   });
+  // TODO DA confirm if it is fine to use seenBlockProposers or create another data structure?
+  this.seenBlockProposers.add(block.message.slot, block.message.proposerIndex);
```
dapplion: This is not okay, seenBlockProposers is exclusively for blocks received via gossip. If you do something like this here, the data structure will be filled with junk during syncing too

packages/validator/src/services/doppelgangerService.ts
dapplion: i think it's critical to be added to this PR, to prove that the feature is working correctly before merging

dapplion (dismissed):

packages/lodestar/src/chain/blocks/importBlock.ts
dadepo: > Why rename chain to modules?

The type says `ImportBlockModules` with modules in it, and coming first to the code base, when I saw `chain` I initially thought this was referring to `IBeaconChain`.

> If you believe this is an important change it should be done in another PR and keep this diff as small as possible

Since this rename is contained to a file, I thought why not just change it here, instead of creating another PR just for it. But I can definitely revert and make another PR if that is the preference.

dapplion: > I thought why not just change it here

doppelganger is a critical feature. A bigger diff correlates with inferior review quality. So for the safety of the feature the diff must be minimal

> But I can definitely revert and make another PR if that is the preference.

please :pray:

dadepo: Gotcha. Already reverted the renaming. Will make another PR for that after this PR is merged.

dadepo:
I ran two vc with same validator twice and it was not detected as a doppelgänger. Temporarily switching to draft while I dig into this. 

Edit: Issue for this particular case has been resolved. Still will be busy with testing

packages/cli/src/options/doppelgangerOptions.ts
```diff
@@ -0,0 +1,13 @@
+import {ICliCommandOptions} from "../util/index.js";
+
+type DoppelgangerOps = {
+  enableDoppelganger?: boolean;
+};
+
+export const doppelgangerOptions: ICliCommandOptions<DoppelgangerOps> = {
+  enableDoppelganger: {
```
wemeetagain: To be more consistent with the naming of other cli args, maybe call this `doppelgangerProtection`

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -582,5 +585,45 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
...
+    async getLiveness(indices: ValidatorIndex[], epoch: Epoch): Promise<{data: LivenessResponseData[]}> {
+      if (indices.length === 0) {
+        return {
+          data: [],
+        };
+      }
+      const currentEpoch = chain.clock.currentEpoch;
+      if (epoch < currentEpoch - 1 || epoch > currentEpoch + 1) {
+        throw new Error(
+          `Request epoch ${epoch} is more than one epoch previous or after from the current epoch ${currentEpoch}`
```
wemeetagain:
```suggestion
          `Request epoch ${epoch} is more than one epoch before or after the current epoch ${currentEpoch}`
```

wemeetagain (commented):
structure looks good :+1:

packages/cli/src/options/doppelgangerOptions.ts
wemeetagain: bump

dadepo: What do you say to `doppelgangerEnabled` instead? Given our use of the past tense to convey options that _enables_ certain features: example: [`proposerBoostEnabled`](https://github.com/ChainSafe/lodestar/blob/04155c66fb28da65650f48d3394ce9762b02f374/packages/cli/src/options/beaconNodeOptions/chain.ts#L9), [`keymanagerEnabled`](https://github.com/ChainSafe/lodestar/blob/04155c66fb28da65650f48d3394ce9762b02f374/packages/cli/src/options/keymanagerOptions.ts#L5) and others like `api.rest.enabled`, `metrics.enabled` etc

wemeetagain: oky, yea

dadepo: renamed to `doppelgangerProtectionEnabled` instead to be more accurate.

packages/validator/src/validator.ts
```diff
@@ -92,8 +95,25 @@ export class Validator {
       defaultFeeRecipient ?? defaultDefaultFeeRecipient
     );
     const indicesService = new IndicesService(logger, api, validatorStore, metrics);
-    const emitter = new ValidatorEventEmitter();
-    const chainHeaderTracker = new ChainHeaderTracker(logger, api, emitter);
+    // Do not enable doppelganger when started before genesis or first slot of genesis, because
+    // there would not have been any activity/signing in the network, so it is pointless to wait.
+    // Doppelganger should be enabled at slot > genesis slot, for example at slot 1, because
+    // same validator could have been started at slot 0 with signature already published to the network
+    if (opts.doppelgangerProtectionEnabled && getCurrentSlot(config, genesis.genesisTime) > 0) {
```
wemeetagain: I think this is not a good check:
in certain cases the doppelganger service just won't be started (even when the user specified `--doppelgangerProtectionEnabled`)

I'd recommend either conditionally calling `assertDoppelgangerSafe` or pass slot information into `assertDoppelgangerSafe` to conditionally do the check there.

wemeetagain (changes requested):

packages/validator/src/services/attestationDuties.ts
```diff
@@ -203,10 +203,12 @@ export class AttestationDutiesService {
...
     const dependentRoot = toHexString(attesterDuties.dependentRoot);
-    const relevantDuties = attesterDuties.data.filter((duty) =>
-      this.validatorStore.hasVotingPubkey(toHexString(duty.pubkey))
-    );
+    const relevantDuties = attesterDuties.data.filter((duty) => {
+      const pubkeyHex = toHexString(duty.pubkey);
+      return this.validatorStore.hasVotingPubkey(pubkeyHex) && this.validatorStore.isDoppelgangerSafe(pubkeyHex);
+    });
 
     this.logger.debug("Downloaded attester duties", {
```
twoeths, line 213: we may want to log the total downloaded duties and the filtered duties

packages/cli/src/options/doppelgangerOptions.ts
```diff
@@ -0,0 +1,13 @@
+import {ICliCommandOptions} from "../util/index.js";
+
+type DoppelgangerOps = {
+  doppelgangerProtectionEnabled?: boolean;
+};
+
+export const doppelgangerOptions: ICliCommandOptions<DoppelgangerOps> = {
+  doppelgangerProtectionEnabled: {
+    description: "Enables Doppelganger protection",
+    default: false,
```
g11tech: should this be enabled by default?

packages/validator/src/services/doppelgangerService.ts
```diff
@@ -0,0 +1,232 @@
...
+      this.logger.error(
+        "Doppelganger(s) detected\n" +
+          "A doppelganger occurs when two different validator clients run the same public key. \
+          This validator client detected another instance of a local validator on the network \
+          and is shutting down to prevent potential slashable offences. Ensure that you are not \
+          running a duplicate or overlapping validator client",
+        violators
+      );
+
+      // TODO: Request process to shutdown
```
twoeths: what's the plan for this TODO?

twoeths (commented):
Looks good to me overall, dropped 1-2 small comments

packages/cli/src/options/doppelgangerOptions.ts
philknows: I don't believe Lighthouse enables it by default. They have an entire page dedicated to understanding the intricacies of it and the user must enable it themselves with the flag. I would assume the same for us too?
https://lighthouse-book.sigmaprime.io/validator-doppelganger.html

packages/validator/src/services/doppelgangerService.ts
dapplion: Will implement in a latter PR.

Current design prevents any validator from signing slash-able messages forever. Killing the process is an important signal to node operators but not critical for base functionality.

dapplion:
Tested doppelganger with a remoteKey on prater. To recreate:

**terminal 1**

Run beacon node to prater

```
cli/bin/lodestar.js beacon --network prater --weakSubjectivitySyncLatest --weakSubjectivityServerUrl https://1sla4tyOFn0bB1ohyCKaH2sLmHu:b8cdb9d881039fd04fe982a5ec57b0b8@eth2-beacon-prater.infura.io --api.rest.enabled --api.rest.api '*' --metrics.enabled --logLevel debug --eth1.enabled false
```

**terminal 2**

Run validator with keymanager API enabled and doppelgangerProtectionEnabled. It won't do anything awaiting to get keys

```
cli/bin/lodestar.js validator --network prater --rootDir .__testdb --keymanager.enabled --logLevel debug --doppelgangerProtectionEnabled
```

**terminal 3**

Add a random active key you grab from https://prater.beaconcha.in/

You need the auth token to authenticate, grab it from the filepath logged in terminal 2
```
cat .__testdb/validator-db/api-token.txt
```

```
curl http://localhost:5062/eth/v1/remotekeys -X POST -H "Content-Type: application/json" -H "Authorization: bearer api-token-0x7fd16fff6453982a5d8bf14617e7823b68cd18ade59985befe64e0a659300e7d" --data '{"remoteKeys": [{"pubkey": "0x998b75a5004802ef22b5583fb006dedc78ec8599754942e7f96fbc0352641b56d09016d156a632d07247ffd5456c752e", "url": "http://sign.er"}]}'
```

Note that the remote signer URL does not exist, but it doesn't matter since it will doppelganger before attempting to sign anything.

You should see

```
Jun-29 13:52:58.828[]                 info: Lodestar network=prater, version=v0.38.1/dadepo/doppelganger/60a0fd1, commit=60a0fd159c7fbc4a22a5968407b34a6c91d62697
Jun-29 13:52:58.844[]                 warn: No signers found with current args, expecting to be added via keymanager
Jun-29 13:52:58.884[]                 info: doppelganger protection enabled detectionEpochs=1
...
Jun-29 14:07:37.685[]                debug: Req req-1 127.0.0.1 importRemoteKeys
Jun-29 14:07:37.689[]                 info: Registered validator for doppelganger remainingEpochs=1, pubkeyHex=0x998b75a5004802ef22b5583fb006dedc78ec8599754942e7f96fbc0352641b56d09016d156a632d07247ffd5456c752e
Jun-29 14:07:37.692[]                debug: Res req-1 importRemoteKeys - 200
...
Jun-29 14:11:09.859[]                 info: Discovered new validators count=1
Jun-29 14:11:09.859[]                debug: doppelganger pollLiveness currentEpoch=104157, indicesCount=1
....
Jun-29 14:17:33.241[]                debug: HttpClient response routeId=getLiveness
Jun-29 14:17:33.247[]                debug: HttpClient response routeId=getLiveness
Jun-29 14:17:33.248[]                error: Doppelganger(s) detected
        A doppelganger occurs when two different validator clients run the same public key.
        This validator client detected another instance of a local validator on the network
        and is shutting down to prevent potential slashable offences. Ensure that you are not
        running a duplicate or overlapping validator client 84577
Jun-29 14:17:33.248[]                error: Process shutdown requested  Doppelganger(s) detected
Error: Doppelganger(s) detected
    at DoppelgangerService.detectDoppelganger (file:///home/lion/Code/eth2.0/lodestar/packages/validator/src/services/doppelgangerService.ts:205:36)
    at DoppelgangerService.pollLiveness (file:///home/lion/Code/eth2.0/lodestar/packages/validator/src/services/doppelgangerService.ts:139:10)
    at Clock.runAtMostEvery (file:///home/lion/Code/eth2.0/lodestar/packages/validator/src/util/clock.ts:86:7)
Jun-29 14:17:33.271[]                 info: Stopping gracefully, use Ctrl+C again to force process exit
```

and terminal 2 process exited with code > 0

packages/validator/src/services/doppelgangerService.ts
```diff
@@ -0,0 +1,266 @@
...
+    await sleep(this.clock.msToSlot(endSlotOfCurrentEpoch + 3 / 4));
+
+    // Collect indices that still need doppelganger checks
+    const pubkeysToCheckWithoutIndex: PubkeyHex[] = [];
+    const indicesToCheck: ValidatorIndex[] = [];
+    const indicesToCheckMap = new Map<ValidatorIndex, PubkeyHex>();
+
+    for (const [pubkeyHex, state] of this.doppelgangerStateByPubkey.entries()) {
+      if (state.remainingEpochs > 0) {
+        const index = this.indicesService.pubkey2index.get(pubkeyHex);
```
dadepo, line 99: Suggestion: instead of exposing the internal states of `indicesService` via `pubkey2index` and `pubkey2index` have a method that does the check, they were exposed for

packages/validator/src/services/doppelgangerService.ts
```diff
@@ -0,0 +1,266 @@
...
+  // This pubkey is known to the doppelganger service but has not been verified safe
+  Unverified = "Unverified",
+  // This pubkey is unknown to the doppelganger service
+  Unknown = "Unknown",
+  // This pubkey has been detected to be active on the network
+  DoppelgangerDetected = "DoppelgangerDetected",
+}
+
+/** Map a DoppelgangerStatus to an integer for rendering in Grafana */
+export const doppelgangerStatusMetrics: {[K in DoppelgangerStatus]: number} = {
```
dadepo: I see this is no longer used. If it won't be anymore maybe remove?

packages/api/src/beacon/routes/validator.ts
```diff
@@ -209,6 +215,9 @@ export type Api = {
   prepareSyncCommitteeSubnets(subscriptions: SyncCommitteeSubscription[]): Promise<void>;
 
   prepareBeaconProposer(proposers: ProposerPreparationData[]): Promise<void>;
+
+  /** Returns validator indices that has been observed to be active on the network */
```
wemeetagain:
```suggestion
  /** Returns validator indices that have been observed to be active on the network */
```

packages/validator/src/services/doppelgangerService.ts
wemeetagain: Yeah if its not used, just remove it. Looks like we now count per status, using the status as a label.

packages/validator/src/services/doppelgangerService.ts
```diff
@@ -0,0 +1,266 @@
...
+    // Collect indices that still need doppelganger checks
+    const pubkeysToCheckWithoutIndex: PubkeyHex[] = [];
+    const indicesToCheck: ValidatorIndex[] = [];
+    const indicesToCheckMap = new Map<ValidatorIndex, PubkeyHex>();
+
+    for (const [pubkeyHex, state] of this.doppelgangerStateByPubkey.entries()) {
+      if (state.remainingEpochs > 0) {
+        const index = this.indicesService.pubkey2index.get(pubkeyHex);
+        if (index !== undefined) {
+          indicesToCheck.push(index);
```
dadepo: given that every instance of `indicesToCheck.push` is accompanied with `indicesToCheckMap.set` maybe have a comment that explains why only `indicesToCheckMap` is not where `indicesToCheckMap.keys` can then be used to get the same values stored as `indicesToCheck`?

packages/validator/src/services/doppelgangerService.ts
```diff
@@ -0,0 +1,266 @@
...
+
+    const endSlotOfCurrentEpoch = computeStartSlotAtEpoch(currentEpoch + 1) - 1;
+    // Run the doppelganger protection check 75% through the last slot of this epoch. This
+    // *should* mean that the BN has seen the blocks and attestations for the epoch
+    await sleep(this.clock.msToSlot(endSlotOfCurrentEpoch + 3 / 4));
+
+    // Collect indices that still need doppelganger checks
+    const pubkeysToCheckWithoutIndex: PubkeyHex[] = [];
+    const indicesToCheck: ValidatorIndex[] = [];
+    const indicesToCheckMap = new Map<ValidatorIndex, PubkeyHex>();
```
wemeetagain: Why do we have both the array and map? Can't we then end up with duplicate indices in the array?
Can we just get the keys on L131?

eg:
...
```ts
if (indicesToCheckMap.size === 0) {
  return;
}

const indicesToCheck = Array.from(indicesToCheckMap.keys());
...
```

packages/validator/src/services/doppelgangerService.ts
```diff
@@ -0,0 +1,266 @@
...
+        `Doppelganger(s) detected
+        A doppelganger occurs when two different validator clients run the same public key.
+        This validator client detected another instance of a local validator on the network
+        and is shutting down to prevent potential slashable offences. Ensure that you are not
+        running a duplicate or overlapping validator client`,
+        violators
+      );
+
+      // Request process to shutdown
+      this.processShutdownCallback(Error("Doppelganger(s) detected"));
```
wemeetagain, line 196:
```suggestion
      this.processShutdownCallback(new Error("Doppelganger(s) detected"));
```

packages/validator/src/services/doppelgangerService.ts
```diff
@@ -0,0 +1,266 @@
...
+      // If a single doppelganger is detected, enable doppelganger checks on all validators forever
+      for (const state of this.doppelgangerStateByPubkey.values()) {
+        state.remainingEpochs = Infinity;
+      }
+
+      this.logger.error(
+        `Doppelganger(s) detected
+        A doppelganger occurs when two different validator clients run the same public key.
+        This validator client detected another instance of a local validator on the network
+        and is shutting down to prevent potential slashable offences. Ensure that you are not
```
wemeetagain:
```suggestion
        and is shutting down to prevent potential slashable offenses. Ensure that you are not
```

wemeetagain (changes requested):

packages/validator/src/services/doppelgangerService.ts
dapplion: As talked with @dadepo I'm happy to add it back latter if there is demand of per-validator metrics

packages/validator/src/services/doppelgangerService.ts
dapplion: Yeah this is a good suggestion. However checked in the code and would require significant changes. Could you open an issue to do latter?

packages/validator/src/services/doppelgangerService.ts
dapplion: It is as a micro-optimization to prevent to iterate the Map again. But agree with you and @dadepo that probably not justifiable. Taking your suggestion

packages/validator/src/services/doppelgangerService.ts
dapplion: Not necessary, built-in Error class does not need `new`

packages/validator/src/services/doppelgangerService.ts
dapplion: Both are correct, tho offenses is from United States english

wemeetagain (approved):
LGTM :rocket:

g11tech (approved):

PR #3885, opened 2022-04-05
Added the eslint rule to help with import order and ran the fix

dadepo:
**Motivation**

In a previous [PR](https://github.com/ChainSafe/lodestar/pull/3522#discussion_r810566341) it was pointed out that the preferred order for imports should be:

1. Non-chainsafe modules
2. @ chainsafe modules
3. Local paths

This PR enforces this via eslint to make it more applicable. It also updates existing import to adhere with the import rule


packages/api/src/routes/beacon/index.ts
```diff
@@ -1,9 +1,9 @@
 import {IChainForkConfig} from "@chainsafe/lodestar-config";
 import {phase0, ssz} from "@chainsafe/lodestar-types";
-import {RoutesData, ReturnTypes, reqEmpty, ContainerData} from "../../utils";
 import * as block from "./block";
 import * as pool from "./pool";
 import * as state from "./state";
+import {RoutesData, ReturnTypes, reqEmpty, ContainerData} from "../../utils";
```
dapplion: So we actually want the closest modules at the bottom, so the correct order would be the original one. @dadepo can you try to tweak the rule to achieve that?

dadepo: Updated!

dapplion (dismissed):
Let's goo! Thanks

twoeths (dismissed):

wemeetagain:
Can we block this until we land ESM support? This will be a bit annoying for that PR by creating a bunch of confiicts.
If this rule can be auto-fixed, it should be very easy to rebase this on master after that.

wemeetagain:
ready for rebase off master

wemeetagain (dismissed):

wemeetagain:
tests still erroring

wemeetagain:
I think I made conflicts with #4064

wemeetagain (dismissed):

packages/lodestar/src/network/metadata.ts
```diff
@@ -48,53 +48,49 @@ export class MetadataController {
...
   }
-
   get seqNumber(): bigint {
     return this._metadata.seqNumber;
   }
-
+  get attnets(): BitArray {
+    return this._metadata.attnets;
+  }
+  set attnets(attnets: BitArray) {
```
dapplion: Please preserve whitespace in all changes of this file. After each method and at least 1 new line

packages/lodestar/src/sync/sync.ts
```diff
@@ -101,35 +130,6 @@ export class BeaconSync implements IBeaconSync {
     return this.state === SyncState.Synced;
   }
 
-  get state(): SyncState {
```
dapplion: Why all this changes in a PR that's clearly scoped to "import order"? Could you please revert this commit and do this changes if applicable? We have really never followed the member-ordering rule in Lodestar

dapplion (dismissed):

wemeetagain (approved):

PR #3893, opened 2022-04-07
Remove gossip validation result log

twoeths:
**Motivation**

+ The gossipsub result does not bring any value, especially after the new gossipsub gets merged, and we have a lot of nice metrics over there
+ It's hard to locate other logs, for example the attestation journey of validators
+ The log file is quite big right now, it's 4GB per day

**Description**

+ Remove gossip result log


dapplion (approved):
LGTM!

In the future we should do a review of our logging policy and verbosity per module specially on debug

Issue #3895, opened 2022-04-08
Validator's performance issue


twoeths:
**Describe the bug**

This is when the slot started

```
Apr-08 05:19:24.000[CHAIN]         ^[[36mverbose^[[39m: Clock slot slot=2739997
```

At validator side, block came on time (1.4s after the slot started)
```
Apr-08 05:19:25.423[]              ^[[36mverbose^[[39m: Found new chain head slot=2739997, head=0xa233e16a0b32634e     ae6eecd32f1d38717d3dbd01d31d3dc59bb6331df46d28f3, previouDuty=0xb1b47f4c2df1e7f0bad5d9745f93d0ee36304924f8f79058db     29ddf3b130cda6, currentDuty=0x0065698b3eea05f0b7c379296842f41ca852e389799275c86761439f2dd0637c
```

however it took alsmot 6s until the attestations were signed
```
8860 Apr-08 05:19:31.250[]                ^[[34mdebug^[[39m: Signed attestation slot=2739997, index=13, head=0xa233e16a     0b32634eae6eecd32f1d38717d3dbd01d31d3dc59bb6331df46d28f3, validatorIndex=200305
8861 Apr-08 05:19:31.255[]                ^[[34mdebug^[[39m: Signed attestation slot=2739997, index=31, head=0xa233e16a     0b32634eae6eecd32f1d38717d3dbd01d31d3dc59bb6331df46d28f3, validatorIndex=200333
```

**Expected behavior**

Attestations should be signed right after beacon block come

This happened in contabo-20 which contains 50 connected validators

twoeths:
For a contabo node of 50 validators, it takes 10% of cpu time to sign SelectionProof
<img width="876" alt="Screen Shot 2022-04-08 at 15 10 38" src="https://user-images.githubusercontent.com/10568965/162393866-76e58b20-5609-458e-a229-e38789a6407c.png">

since we get attestation duties for current epoch and next epoch, and reorg rarely happens, we should be able to do a cache to divide that time by 2

twoeths:
Found another example in contabo-19 (30 connected validators atm)

+ The node
```
Apr-10 07:46:24.012[CHAIN]         ^[[36mverbose^[[39m: Clock slot slot=2755132
[...]

PR #3897, opened 2022-04-08
Cache selectionProofs

twoeths:
**Motivation**

+ For each validator, we create selection proof twice for the same slot.
+ We want to avoid this since it takes 10% of validator time in a contabo node of 50 connected validators

**Description**

Cache selection proof by pubkey hex by slot, prune it per epoch

part of #3895


wemeetagain (approved):
LGTM

dapplion:
While this solution is good, I would like to know why this happens in the first place

> For each validator, we create selection proof twice for the same slot.

Since we are tracking dependantRoots, why does that need to be computed twice?

twoeths:
> While this solution is good, I would like to know why this happens in the first place
> 
> > For each validator, we create selection proof twice for the same slot.
> 
> Since we are tracking dependantRoots, why does that need to be computed twice?

so I debug and found that we cache duties by dependent root already, so it works as designed

PR #3905, opened 2022-04-10
Re-org dashboards into higher folder

dapplion:
**Motivation**

We have invested a lot into having very useful dashboards. This dashboards are now buried into `docker/grafana/provisioning/dashboards`. This PR moves them into `dashboards`

**Description**

- Re-org dashboards into higher folder
- Re-work docker local setup to make it work with new directory structure
- De-duplicate files for local and MacOS users, i.e. merge 3 files in 1. @dadepo it should work fine and now you only need to edit two lines in a single file. If editing that file gets repetitive you can setup a .env file for the docker-compose.yml and read from there.


g11tech (approved):
LGTM! pushing dashboards :rocket: :heart:  !!!

Issue #2853, opened 2021-07-19
Low sync committee participation using dev command


wemeetagain:
**Describe the bug**
Running a transient devnet using the dev command, sync committee participation is very low (often zero).

**Expected behavior**

Using the dev command, the sync committee participation should be perfect.

**Steps to Reproduce**

```bash
node --inspect lodestar dev --api.rest.host 0.0.0.0 --api.rest.api beacon config events node validator lightclient --startValidators 0:7 --params.ALTAIR_FORK_EPOCH 0 --params.SECONDS_PER_SLOT 6 --sync.isSingleNode --network.discv5.enabled false --logLevel verbose --logLevelFile debug --logFile foo.log
```

Separately:

```bash
curl "http://localhost:9596/eth/v2/beacon/blocks/N" | jq .
```

dadepo:
From what I can see, it seems the sync committee participation is now ok. Given the following output:

```
➜  ~ curl "http://localhost:9596/eth/v1/beacon/states/head/sync_committees" | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   179  100   179    0     0  17640      0 --:--:-- --:--:-- --:--:-- 44750
{
  "data": {
    "validators": [
      "3",
      "2",
      "5",
      "6",
      "7",
      "4",
      "0",
      "1",
      "3",
      "2",
      "5",
      "6",
      "7",
      "4",
      "0",
      "1",
      "3",
      "2",
      "5",
      "6",
      "7",
      "4",
      "0",
      "1",
      "3",
      "2",
      "5",
      "6",
      "7",
      "4",
      "0",
      "1"
    ],
    "validator_aggregates": []
  }
}
```

also 

```
curl "http://localhost:9596/eth/v2/beacon/blocks/2" | jq -r '.data.message.body.sync_aggregate'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1883  100  1883    0     0   315k      0 --:--:-- --:--:-- --:--:--  919k
{
  "sync_committee_bits": "0xffffffff",
[...]

PR #3909, opened 2022-04-11
Including sync committees in the sim test reporter

dadepo:
**Motivation**

As a result of investigating #2853, a suggestion arose to also include the sync committees in the report for the SIM tests. This PR does that.


twoeths:
@dadepo I suppose we also need some assertions on the sync committee participations for altair sim tests?

dadepo:
> @dadepo I suppose we also need some assertions on the sync committee participations for altair sim tests?

I added an e2e test to include these assertions. I'll leave the changes in the SIM test, as I suppose having the information in the report is still useful.

twoeths (dismissed):

dapplion (changes requested):
The current approach of this PR is to add another sim test run to assert sync committee correct behaviour. Each run to finalization takes significant time, so this approach doesn't scale. We want to do way more assertions than only this one while keeping the CI runtime short.

Instead the goal is to run very few sim tests with strictly exclusive configurations (like forks), and do _all_ assertions on those few runs.

dadepo:
> The current approach of this PR is to add another sim test run to assert sync committee correct behaviour. Each run to finalization takes significant time, so this approach doesn't scale. We want to do way more assertions than only this one while keeping the CI runtime short.
> 
> Instead the goal is to run very few sim tests with strictly exclusive configurations (like forks), and do _all_ assertions on those few runs.

To be sure we are working with the same terminology here. I'll like to clarify my interpretation of sim (and hence e2e) test.

I consider the tests in `packages/lodestar/test/sim` to be the sim tests. These tests run the beacon node but do not have assertions. They only wait for chain related events: `finalized etc`. This PR does not add any new sim tests but updated the report to include sync committees.

There are the `e2e` tests which I consider to be the ones in `packages/lodestar/test/e2e`. These tests include ones that also run a beacon node but unlike `sim` tests, they also have assertions. This PR added another `e2e` test to confirm sync committees.

Will I be correct by saying, when you say `sim` tests, you are talking about the `e2e` tests? And you prefer instead of adding another one via the new `synccommittee.test.ts` one of the existing ones should be edited to include asserting for sync committees?

If so, which one will you suggest to be edited?

From a quick look, it looks as if what is needed is a proper consolidation of all e2e tests that runs a beacon node into one, where all assertions should be performed, and that looks like a separate effort, as this PR came from the issue https://github.com/ChainSafe/lodestar/issues/2853 which was specifically about sync committees not being enough, and after investigation the idea was to add a test to actually confirm that the correct amount of sync committees are included.

dapplion:
@dadepo yeah overall your picture is correct. To summarize, my points are:

e2e and sim test terminology is confusing, so it creates our current missorganization

**For this PR**
- Do not add this (e2e / sim) test `packages/lodestar/test/e2e/syncCommittee/synccommittee.test.ts`
- Extend current sim test `packages/lodestar/test/sim/singleNodeSingleThread.test.ts` to add an assertion about sync committee participation. Just adding information in the report is almost useless since no-one looks at those. We need automatic programatic assurances.

**Latter, sim test consolidation**
- Add way more assertions to `packages/lodestar/test/sim/singleNodeSingleThread.test.ts`, and the multinode one
- Any test in e2e that could be merged into `packages/lodestar/test/sim/singleNodeSingleThread.test.ts` because it uses the same configuration, do that and consolidate.

wemeetagain (dismissed):

dapplion (changes requested):
I would rather put this initiative in pause and implement a proper assertion infrastructure soon. This is not critical and definitely of less priority than other pending things

dapplion:
- Closing for https://github.com/ChainSafe/lodestar/issues/4276

PR #3912, opened 2022-04-12
Snappy frame encode big payloads as chunks as per the standard

### LINKED ISSUE #3903: investigate and fix snappy stream encoding of a big bellatrix block which was rejected by the lighthouse bn
Pawan from lighthouse raised an issue where there snappy decoder rejected a lodestar encoded block. Apparently it was a big `bellatrix` block from `goerli shadowfork 2`.

Investigation Updates:
- [x] After a bit of joint setup and investigation, pawan came up with theory that it was because of the snappy stream specification where a decoded/uncompressed chunk can't be > 65k bytes, and this particular block was around 130k bytes.
- [x] Was able to stream decode the provided data, as suspected it was just `1` chunk, was able to successfully ssz decode bellatrix block
  image
- [x] Chunkified the uncompressed data into three chunks and reencoded it as a stream and dumped it into a file. Passed that re-encoded chunkified stream against the test code to confirm that it was now being decoded as 3 chunks `(60k,60k,13k)` something bytes. Provided the file to pawan over discord for him to run through his decoder and confirm
- [x] Waiting on Pawan/Lighthouse team for confirmation on the chunk post, [details here in comment](https://github.com/ChainSafe/lodestar/issues/3903#issuecomment-1094113893)

Expectation: Upon successful confirmation from Pawan/lighthouse regarding the successful decoding of the chunkified stream, do a PR in lodestar to chunkify the snappy stream payloads as well as add tests for the same using the same block data

g11tech:
**Motivation**
On a bellatrix block of goerli shadowfork2, lighthouse frame decoder failed to decode block it req/resp from a lodestar node. Turned out it was big block, but being streamed as a single chunk which didnot confirm to the snappy frame encoding/decoding specs.

This PR chunkifies a big payload in the compress stream as per the standards and add tests for the same block

Closes #3903


packages/lodestar/src/network/reqresp/encodingStrategies/sszSnappy/encode.ts
```diff
@@ -4,6 +4,13 @@ import {createCompressStream} from "@chainsafe/snappy-stream";
 import {RequestOrOutgoingResponseBody, OutgoingSerializer} from "../../types";
 import {SszSnappyError, SszSnappyErrorCode} from "./errors";
 
+/**
+ * As per the snappy framing format for streams, the size of any uncompressed chunk can be
```
wemeetagain: I'm wondering if this behavior should be added upstream to the snappy-streams library?

g11tech: can be wrapped here: https://github.com/ChainSafe/node-snappy-stream/blob/master/lib/compress-stream.js#L56:L66

dapplion: Upstreaming if possible would be nice. Keep a copy of the test in Lodestar tho, for our safety :pray:

packages/lodestar/src/network/reqresp/encodingStrategies/sszSnappy/encode.ts
```diff
@@ -29,7 +29,7 @@ export async function* writeSszSnappyPayload<T extends RequestOrOutgoingResponse
  * Buffered Snappy writer
  */
 function encodeSszSnappy(bytes: Buffer): AsyncGenerator<Buffer> {
-  const stream = createCompressStream();
+  const stream = createCompressStream({asyncCompress: true});
```
g11tech: @dapplion @wemeetagain do we want to go with async or sync version? the sync version is 2x faster, and payloads are mostly 1 chunk (~`0.0229 ms` compress time) (the big bellatrix goerli block was 2 chunks ~`0.0562 ms` compress time).

wemeetagain: Imo lets go with sync. If we start dealing with larger payloads, we might revisit, selectively use async in those cases.

dapplion (approved):

packages/lodestar/src/network/reqresp/encodingStrategies/sszSnappy/encode.ts
```diff
@@ -29,6 +29,14 @@ export async function* writeSszSnappyPayload<T extends RequestOrOutgoingResponse
...
  */
 function encodeSszSnappy(bytes: Buffer): AsyncGenerator<Buffer> {
+  /**
+   * Use sync version (default) for compress as it is almost 2x faster than async
+   * one and most payloads are "1 chunk" and 100kb payloads (which would mostly be
+   * big bellatrix blocks with transactions) are just 2 chunks
+   *
+   * To use async version (for e.g. on big payloads) instantiate the stream with
+   * `createCompressStream({asyncCompress: true})`
+   */
```
dapplion, line 39: Thank you so much for adding a nice reasoning comment! :heart:

PR #3899, opened 2022-04-08
Add more API REST metrics

dapplion:
**Motivation**

Need more metrics to track our HTTP servers

**Description**

Existing:
- Successful response count

New:
- Total request count
- Total error count
- Current active socket count: REST API + metrics server
- Total count of bytes transmitted in + out
- Current buffer size in sockets

PR #3914, opened 2022-04-12
Add validator client metrics

### LINKED ISSUE #2007: Add validator client metrics
We do have validator monitor metrics, related to performance of validators in the Eth2.0 protocol (profitability, attestation performance)

However this issue refers to validator client metrics, which may include:

- VM stats: CPU, RAM, event loop lag, GC runs
- Signer metrics: total signed items, signature errors
- API client metrics: response times, timeouts, error count
- Slashing protection DB metrics

dapplion:
**Motivation**

We need visibility into the validator client.

Opening as draft first to get consensus of the metrics set

**Description**

Depends on https://github.com/ChainSafe/lodestar/pull/3899

Closes https://github.com/ChainSafe/lodestar/issues/2007

### DIFF
--- a/packages/cli/src/cmds/validator/handler.ts
+++ b/packages/cli/src/cmds/validator/handler.ts
@@ -1,5 +1,4 @@
 import {AbortController} from "@chainsafe/abort-controller";
-import {getClient} from "@chainsafe/lodestar-api";
 import {LevelDbController} from "@chainsafe/lodestar-db";
 import {SignerType, Signer, SlashingProtection, Validator} from "@chainsafe/lodestar-validator";
 import {getMetrics, MetricsRegister} from "@chainsafe/lodestar-validator";
@@ -92,7 +91,6 @@ export async function validatorHandler(args: IValidatorCliArgs & IGlobalArgs): P
   const controller = new AbortController();
   onGracefulShutdownCbs.push(async () => controller.abort());
 
-  const api = getClient(config, {baseUrl: args.server});
   const dbOps = {
     config: config,
     controller: new LevelDbController({name: dbPath}, {logger}),
@@ -131,7 +129,7 @@ export async function validatorHandler(args: IValidatorCliArgs & IGlobalArgs): P
   // It will wait for genesis, so this promise can be potentially very long
 
   const validator = await Validator.initializeFromBeaconNode(
-    {dbOps, slashingProtection, api, logger, signers, graffiti},
+    {dbOps, slashingProtection, api: args.server, logger, signers, graffiti},
     controller.signal,
     metrics
   );
--- a/packages/cli/src/cmds/validator/handler.ts
+++ b/packages/cli/src/cmds/validator/handler.ts
@@ -122,7 +122,6 @@ export async function validatorHandler(args: IValidatorCliArgs & IGlobalArgs): P
     const port = args["metrics.port"] ?? validatorMetricsDefaultOptions.port;
     const address = args["metrics.address"] ?? validatorMetricsDefaultOptions.address;
     const metricsServer = new HttpMetricsServer({port, address}, {register, logger});
-    logger.info("Starting metrics HTTP server", {port, address});
 
     onGracefulShutdownCbs.push(() => metricsServer.stop());
     await metricsServer.start();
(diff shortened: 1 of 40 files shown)


packages/validator/src/metrics.ts
```diff
@@ -0,0 +1,316 @@
...
+
+  return {
+    // Attestation journey:
+    // - Wait for block or 1/3, call prepare attestation
+    // - Get attestation, sign, call publish
+    // - Wait 2/3, call prepare aggregate
+    // - Get aggregate, sign, call publish
+
+    // Max wait time is 1 / 3 of slot = 12 / 3 = 4 sec
+    attesterStepCallProduceAttestation: register.histogram({
```
twoeths, line 98: right now there is I/O lag issue at the beacon node side in contabo nodes - feature group, so I'd like to track the time before we produce and right after we produce attestation.

packages/validator/src/services/attestation.ts
```diff
@@ -88,21 +98,37 @@ export class AttestationService {
...
     // validator that is elected to aggregate for this `slot` and
     // `committeeIndex`.
     await this.produceAndPublishAggregates(attestation, duties);
   }
 
+  /**
+   * Performs the first step of the attesting process: downloading one `Attestation` object.
+   * Beacon node's endpoint produceAttestationData return data is not dependant on committeeIndex.
+   * For a validator client with many validators this allows to do a single call for all committees
+   * in a slot, saving resources in both the vc and beacon node
```
twoeths: this is nice!

packages/validator/src/services/attestation.ts
```diff
@@ -52,16 +55,23 @@ export class AttestationService {
     // A validator should create and broadcast the attestation to the associated attestation subnet when either
     // (a) the validator has received a valid block from the expected block proposer for the assigned slot or
     // (b) one-third of the slot has transpired (SECONDS_PER_SLOT / 3 seconds after the start of slot) -- whichever comes first.
-    await Promise.race([sleep(this.clock.msToSlotFraction(slot, 1 / 3), signal), this.waitForBlockSlot(slot)]);
+    await Promise.race([sleep(this.clock.msToSlot(slot + 1 / 3), signal), this.waitForBlockSlot(slot)]);
+    this.metrics?.attesterStepCallProduceAttestation.observe(this.clock.secFromSlot(slot));
+
+    // Beacon node's endpoint produceAttestationData return data is not dependant on committeeIndex.
+    // Produce a single attestation for all committees, and clone mutate before signing
+    const attestationNoCommittee = await this.produceAttestation(slot);
```
twoeths: there could be errors at each of the step:
+ error producing attestation
+ error signing attestation
+ error publishing attestation

we can track each of the error type, or a general AttestationError metric. It could help us detect missed attestations right at the validator side.

packages/validator/src/services/attestation.ts
dapplion: Now in its own PR https://github.com/ChainSafe/lodestar/pull/3918

packages/validator/src/metrics.ts
dapplion: That is tracked by the REST API client metrics

packages/validator/src/services/attestation.ts
dapplion: Added

packages/cli/src/cmds/validator/handler.ts
```diff
@@ -1,17 +1,18 @@
 import {AbortController} from "@chainsafe/abort-controller";
-import {getClient} from "@chainsafe/lodestar-api";
 import {LevelDbController} from "@chainsafe/lodestar-db";
 import {SignerType, Signer, SlashingProtection, Validator} from "@chainsafe/lodestar-validator";
+import {getMetrics, MetricsRegister} from "@chainsafe/lodestar-validator";
 import {KeymanagerServer, KeymanagerApi} from "@chainsafe/lodestar-keymanager-server";
+import {RegistryMetricCreator, collectNodeJSMetrics, HttpMetricsServer} from "@chainsafe/lodestar";
```
wemeetagain, line 6: might be nice to pull this and the `IGauge` interface at some point?

wemeetagain (dismissed):
the metrics look great

packages/cli/src/cmds/validator/handler.ts
dapplion: Yeah.. but probably not worth the extra package or dependency relations. We can review in the future a cleaner approach, now just doing whatever works

wemeetagain: utilz

wemeetagain (approved):

packages/cli/src/cmds/validator/handler.ts
dapplion: urrg the recursive dependencies between lodestar/utils and ssz have been awful. Not looking forward to something similar

PR #3917, opened 2022-04-13
Do single call to produce attestation

### LINKED ISSUE #3098: Call produceAttestationData once for all validators
**What is your question?**

To reduce API calls cost on large validator counts, just call produceAttestationData once in the validator flow, and edit the committeeIndex

dapplion:
**Motivation**

Beacon node's endpoint produceAttestationData return data is not dependant on committeeIndex. We can do a single call for all committees and save resources

**Description**

Produce a single attestation for all committees, and clone mutate before signing

Closes https://github.com/ChainSafe/lodestar/issues/3098

### DIFF
--- a/packages/validator/src/services/attestation.ts
+++ b/packages/validator/src/services/attestation.ts
@@ -57,14 +57,20 @@ export class AttestationService {
     // (b) one-third of the slot has transpired (SECONDS_PER_SLOT / 3 seconds after the start of slot) -- whichever comes first.
     await Promise.race([sleep(this.clock.msToSlotFraction(slot, 1 / 3), signal), this.waitForBlockSlot(slot)]);
 
+    // Beacon node's endpoint produceAttestationData return data is not dependant on committeeIndex.
+    // Produce a single attestation for all committees, and clone mutate before signing
+    const attestationNoCommittee = await this.produceAttestation(slot);
+
     // await for all so if the Beacon node is overloaded it auto-throttles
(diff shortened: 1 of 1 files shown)


twoeths (approved):
lgtm!

nflaig:
@dapplion The DVT compatibility issue mentioned in https://github.com/ChainSafe/lodestar/issues/5103 is due to the changes in this PR. This also introduced a deviation compared to other CL clients as those send a request per committee index.

I am wondering if we really need this optimization as there is an upper bound of 64 requests due to [MAX_COMMITTEES_PER_SLOT](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#misc-1) but of course it will put more load on the BN either way.

I really want to keep the optimization while also fixing the compatibility issues we have.
As described in the issue, my idea is to introduce a flag `--distributed` (also already considering [aggregate committee selections](https://docs.google.com/document/d/1q9jOTPcYQa-3L8luRvQJ-M0eegtba4Nmon3dpO79TMk/mobilebasic)) and if set the attestation service behaves differently (disables the optimization). My main concern with all of this is that we introduce branching logic the attestation flow which I really want to avoid.

Considering all of that, instead of adding branches to the current attenstation service I think we might just be best of creating a 2nd "distributed" attestation service.

Edit: opened a PR to propose and discuss a possible solution https://github.com/ChainSafe/lodestar/pull/5258

PR #3921, opened 2022-04-14
Increase committee size in sim tests

twoeths:
**Motivation**

Increase committee size in sim tests to avoid any regressions we may have


dapplion (approved):

philknows:
Implemented due to #3919 not being detected from sim tests potentially from small committee sizes. Taking note that we should also seek to improve our sim tests with more depth when able.

PR #3926, opened 2022-04-15
Call validatorStore sign in parallel

dapplion:
**Motivation**

If using the remote signer we want to request all signatures at once.

**Description**

Call validatorStore sign in parallel

### DIFF
--- a/packages/validator/src/services/attestation.ts
+++ b/packages/validator/src/services/attestation.ts
@@ -194,20 +194,22 @@ export class AttestationService {
 
     const signedAggregateAndProofs: phase0.SignedAggregateAndProof[] = [];
 
-    for (const {duty, selectionProof} of duties) {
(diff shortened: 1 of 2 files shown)


dadepo:
> 
> If using the remote signer we want to request all signatures at once.
> 

Just to be sure if this is a typo or not, looking through the PR, there is nothing about remote signer. Is this description a typo?

dapplion:
> > If using the remote signer we want to request all signatures at once.
> 
> Just to be sure if this is a typo or not, looking through the PR, there is nothing about remote signer. Is this description a typo?

It's correct, each `await validatorStore.sing*`, if using a remote signer, awaits for a HTTP request to be completed. When using a local signer `await validatorStore.sing*` the function is sync so the behavior is equivalent as before this PR.

wemeetagain (approved):

PR #3927, opened 2022-04-15
New metric filtering missed blocks

### LINKED ISSUE #3925: New metric filtering missed blocks

In Prater, from Apr 08, seems like there are some validators send late gossip blocks, these blocks are missed blocks but it make our chart looks really urgly
Screen Shot 2022-04-15 at 14 15 32

One example is as below, the gossip block is sent almost 2 minutes after the slot started
```
Apr-15 08:58:15.844[NETWORK]       verbose: Received gossip block slot=2791482, root=0x4250…57ee, curentSlot=2791491, peerId=bafzaajiiaijccawxi3dft7goz4wuf6ugdy7j4s5ymjisbiwqxu6ddj7llhyg3tdsge, delaySec=111.80999994277954
```

and it's marked as missed in beaconcha.in
Screen Shot 2022-04-15 at 14 13 21

Create new metric to track "new chain head" block delay second

twoeths:
**Motivation**

Some gossip blocks come very late and become "missed/orphaned", we don't want to always see it

**Description**

New metric to track slot time till a block becomes new head

Closes #3925

### DIFF
--- a/packages/lodestar/src/chain/eventHandlers.ts
+++ b/packages/lodestar/src/chain/eventHandlers.ts
@@ -9,7 +9,6 @@ import {ChainEvent, IChainEvents} from "./emitter";
 import {BeaconChain} from "./chain";
 import {REPROCESS_MIN_TIME_TO_NEXT_SLOT_SEC} from "./reprocess";
 import {toCheckpointHex} from "./stateCache";
-import {BlockSource} from "./blocks/types";
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type AnyCallback = () => Promise<void>;
@@ -149,7 +148,7 @@ export async function onForkChoiceFinalized(this: BeaconChain, cp: CheckpointWit
   }
 }
 
-export function onForkChoiceHead(this: BeaconChain, head: IProtoBlock, source: BlockSource): void {
+export function onForkChoiceHead(this: BeaconChain, head: IProtoBlock): void {
   const delaySec = this.clock.secFromSlot(head.slot);
   this.logger.verbose("New chain head", {
     headSlot: head.slot,
@@ -159,7 +158,7 @@ export function onForkChoiceHead(this: BeaconChain, head: IProtoBlock, source: B
   this.syncContributionAndProofPool.prune(head.slot);
   this.seenContributionAndProof.prune(head.slot);
   this.metrics?.headSlot.set(head.slot);
-  this.metrics?.blockElapsedTimeTillBecomeHead.observe({src: source}, delaySec);
+  this.metrics?.gossipBlock.elapsedTimeTillBecomeHead.observe(delaySec);
 }
 
 export function onForkChoiceReorg(this: BeaconChain, head: IProtoBlock, oldHead: IProtoBlock, depth: number): void {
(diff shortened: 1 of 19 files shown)


twoeths:
New metric looks like

Screen Shot 2022-04-15 at 19 08 15

packages/lodestar/src/chain/eventHandlers.ts
```diff
@@ -149,13 +149,16 @@ export async function onForkChoiceFinalized(this: BeaconChain, cp: CheckpointWit
...
+  const delaySec = Date.now() / 1000 - (this.genesisTime + head.slot * this.config.SECONDS_PER_SLOT);
   this.logger.verbose("New chain head", {
     headSlot: head.slot,
     headRoot: head.blockRoot,
+    delaySec,
   });
   this.syncContributionAndProofPool.prune(head.slot);
   this.seenContributionAndProof.prune(head.slot);
   this.metrics?.headSlot.set(head.slot);
+  this.metrics?.gossipBlock.elappsedTimeTillBecomeHead.observe(delaySec);
```
dadepo: minor: typo in elapsed

packages/lodestar/src/chain/eventHandlers.ts
```diff
@@ -181,9 +184,11 @@ export async function onBlock(
 
   this.reprocessController.onBlockImported({slot: block.message.slot, root: blockRoot}, advancedSlot);
 
+  const delaySec = Date.now() / 1000 - (this.genesisTime + block.message.slot * this.config.SECONDS_PER_SLOT);
```
dadepo: Looks like this logic to calculate `delaySec` is already provided in `clock.secFromSlot(slot: Slot)`

packages/lodestar/src/metrics/metrics/lodestar.ts
```diff
@@ -514,6 +514,11 @@ export function createLodestarMetrics(
 
     // Gossip block
     gossipBlock: {
+      elappsedTimeTillBecomeHead: register.histogram({
+        name: "lodestar_gossip_block_elappsed_time_till_become_head",
+        help: "Time elappsed between block slot time and the time block becomes head",
+        buckets: [0.1, 1, 10],
```
dapplion: It's very unlikely for a block to come exactly between 0 and 0.1 seconds, better bucket values would be:
```
[0.5, 1, 2, 4, 6, 12]
```

packages/lodestar/src/chain/eventHandlers.ts
dapplion: @tuyennhv Are you sure this metric won't be triggered during sync, and unknown block root fetches?

twoeths: I added BlockSource enum and track that in the new metric, please have a look

dapplion (approved):
@tuyennhv I reverted your commit adding the block source since I think it's valuable to track blocks that become head late due to unknown block sync too. For example, a block through gossip can trigger unknown block sync and its parent won't be registered like it came from gossip. The PR adds less complexity too. If the metrics are not clear we can add the source thing

PR #3935, opened 2022-04-18
Extend error messages when voluntary exit errors because of present of lockfile

### LINKED ISSUE #2792: Voluntary exit fails when validator client is already running
**Describe the bug**

Voluntary exit fails when validator client is already running because of existing lockfile. Example:

```
yorick@ethlinux:~/eth-docker-devel$ docker-compose run --rm validator-exit
? Which validator do you want to voluntarily exit from the network? 0xad621fc2267c4cbf4ad774c3a86c372ff73473a309e3d71c0f56a7485188ccd4ea75b2f434afbde958d8486d448f5d6c
? Are you sure you want to permantently exit validator 0xad621fc2267c4cbf4ad774c3a86c372ff73473a309e3d71c0f56a7485188ccd4ea75b2f434afbde958d8486d448f5d6c from the prater network?

WARNING: THIS CANNOT BE UNDONE.

ONCE YOU VOLUNTARILY EXIT, YOU WILL NOT BE ABLE TO WITHDRAW
YOUR DEPOSIT UNTIL PHASE 2 IS LAUNCHED WHICH MAY NOT
BE UNTIL AT LEAST TWO YEARS AFTER THE PHASE 0 MAINNET LAUNCH.

 YES
Initiating voluntary exit for validator 0xad621fc2267c4cbf4ad774c3a86c372ff73473a309e3d71c0f56a7485188ccd4ea75b2f434afbde958d8486d448f5d6c
 ✖ Error: EEXIST: file already exists, open '/var/lib/lodestar/validators/keystores/0xad621fc2267c4cbf4ad774c3a86c372ff73473a309e3d71c0f56a7485188ccd4ea75b2f434afbde958d8486d448f5d6c/.lock'
```

**Expected behavior**

Given that a validator needs to continue validating until it has fully exited, being able to submit an exit while the validator client is running is the expected workflow. This is also how the "other four" clients work.

**Steps to Reproduce**

Run validator client with the following parameters:

```
    entrypoint:
      - node
      - --max-old-space-size=8192
      - /usr/app/node_modules/.bin/lodestar
      - validator
      - --rootDir
      - /var/lib/lodestar/validators
      - --server
      - http://consensus:9596
      - --graffiti
      - ${GRAFFITI}
      - --logLevel
      - ${LOG_LEVEL}
      - --network
      - ${NETWORK}
```

Then run a voluntary exit with these parameters:

```
    entrypoint:
      - node
      - --max-old-space-size=8192
      - /usr/app/node_modules/.bin/lodestar
      - account
      - validator
      - voluntary-exit
      - --rootDir
      - /var/lib/lodestar/validators
      - --server
      - http://consensus:9596
      -  --logLevel
      - ${LOG_LEVEL}
      - --network
      - ${NETWORK}
```

- OS: Ubuntu 20.04 with docker
- Version: version=0.25.1 master 88b23663

dadepo:
**Motivation**

Including message to use the force flag in case voluntary exit throws error because of the existence of lockfile

Closes #2792


packages/cli/src/validatorDir/ValidatorDir.ts
```diff
@@ -68,7 +67,7 @@ export class ValidatorDir {
       if (options && options.force) {
         // Ignore error, maybe log?
       } else {
-        throw e;
+        throw extendError(e as Error, "use --force to override");
```
dapplion: Doing this is extremely dangerous, we don't want to encourage this behaviour at all. Note that this code path is also used for validating

dapplion (changes requested):

dapplion (approved):

PR #3942, opened 2022-04-20
Integrate `besu` with the merge script along with some general cleanup 

g11tech:
This PR integrates `besu` with the merge script, and also add it to the upcoming `mainnet shadow fork 2` config as well as the `kiln` network:

`./setup.sh --dataDir kiln-data --elClient besu --devnetVars ./kiln.vars --dockerWithSudo --withTerminal "gnome-terminal --disable-factory --"`

Confirmed working on `kiln` and confirmed mainnet syncing on mainnet shadow fork 2

### DIFF
--- a/kiln/devnets/goerlishadow-4.vars
+++ b/kiln/devnets/goerlishadow-4.vars
@@ -1,21 +1,28 @@
 DEVNET_NAME=goerlishadow-4
+CONFIG_GIT_DIR=goerli-shadow-fork-4
+NETWORK_ID=5
+MERGE_TTD=9895202
+
+JWT_SECRET="0xdc6457099f127cf0bac78de8b297df04951281909db4f58b43def7c7151e765d"
 
 GETH_IMAGE=parithoshj/geth:master
 NETHERMIND_IMAGE=nethermindeth/nethermind:kiln_shadowfork
 ETHEREUMJS_IMAGE=g11tech/ethereumjs:kiln
-LODESTAR_IMAGE=chainsafe/lodestar:next
-CONFIG_GIT_DIR=goerli-shadow-fork-4
+BESU_IMAGE=hyperledger/besu:22.4.0-RC2-SNAPSHOT
 
-JWT_SECRET="0xdc6457099f127cf0bac78de8b297df04951281909db4f58b43def7c7151e765d"
+LODESTAR_IMAGE=chainsafe/lodestar:next
 
 LODESTAR_EXTRA_ARGS="--eth1.providerUrls http://127.0.0.1:8545 --execution.urls http://127.0.0.1:8551 --api.rest.enabled --api.rest.host 0.0.0.0 --api.rest.api '*'"
 
 LODESTAR_VALIDATOR_ARGS='--network kiln  --fromMnemonic "lens risk clerk foot verb planet drill roof boost aim salt omit celery tube list permit motor obvious flash demise churn hold wave hollow" --mnemonicIndexes 0..5'
 
-NETHERMIND_EXTRA_ARGS="--config  goerli_shadowfork --Network.DiscoveryPort=30303 --Network.P2PPort=30303 --Merge.Enabled=true --Merge.TerminalTotalDifficulty=9891031 --Init.DiagnosticMode=None --JsonRpc.Enabled=true --JsonRpc.Host=0.0.0.0 --JsonRpc.AdditionalRpcUrls \"http://localhost:8545|http;ws|net;eth;subscribe;engine;web3;client|no-auth,http://localhost:8551|http;ws|net;eth;subscribe;engine;web3;client\""
+NETHERMIND_EXTRA_ARGS="--config  goerli_shadowfork --Network.DiscoveryPort=30303 --Network.P2PPort=30303 --Merge.Enabled=true --Init.DiagnosticMode=None --JsonRpc.Enabled=true --JsonRpc.Host=0.0.0.0 --JsonRpc.AdditionalRpcUrls \"http://localhost:8545|http;ws|net;eth;subscribe;engine;web3;client|no-auth,http://localhost:8551|http;ws|net;eth;subscribe;engine;web3;client\""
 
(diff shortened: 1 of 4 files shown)


kiln/devnets/goerlishadow-4.vars
```diff
@@ -1,21 +1,28 @@
...
+MERGE_TTD=9895202
+
+JWT_SECRET="0xdc6457099f127cf0bac78de8b297df04951281909db4f58b43def7c7151e765d"
 
 GETH_IMAGE=parithoshj/geth:master
 NETHERMIND_IMAGE=nethermindeth/nethermind:kiln_shadowfork
 ETHEREUMJS_IMAGE=g11tech/ethereumjs:kiln
-LODESTAR_IMAGE=chainsafe/lodestar:next
-CONFIG_GIT_DIR=goerli-shadow-fork-4
+BESU_IMAGE=hyperledger/besu:22.4.0-RC2-SNAPSHOT
```
garyschulte: `22.4.0-RC2-SNAPSHOT`, is currently the latest build, but `develop` might be a more future-proof tag here

kiln/devnets/goerlishadow-4.vars
```diff
@@ -1,21 +1,28 @@
...
 
-GETH_EXTRA_ARGS="--http --http.api engine,net,eth,web3 --http.port 8545 --allow-insecure-unlock --http.addr 0.0.0.0 --http.corsdomain \"*\" --http.vhosts \"*\" --authrpc.port=8551 --networkid 5"
+GETH_EXTRA_ARGS="--http --http.api engine,net,eth,web3 --http.port 8545 --allow-insecure-unlock --http.addr 0.0.0.0 --http.corsdomain \"*\" --http.vhosts \"*\" --authrpc.port=8551"
 
 ETHEREUMJS_EXTRA_ARGS="--saveReceipts --rpc --rpcport 8545 --ws --rpcEngine --rpcEnginePort=8551 --rpcDebug --loglevel=debug"
 
+BESU_EXTRA_ARGS="--rpc-http-enabled=true --rpc-http-api=ADMIN,CLIQUE,MINER,ETH,NET,DEBUG,TXPOOL,ENGINE,TRACE --rpc-http-host=0.0.0.0 --rpc-http-port=8545 --Xmerge-support=true --engine-rpc-http-port=8551 --rpc-http-cors-origins=\"*\" --host-allowlist=\"*\" --engine-host-allowlist=\"*\" --p2p-enabled=true --engine-jwt-enabled=true"
```
garyschulte: you should remove `ENGINE` from rpc-http-api, it is not an allowed part of the public facing api

garyschulte (commented):
couple suggestions

g11tech:
thanks @garyschulte for looking into it :slightly_smiling_face: :pray: . updated as suggested :+1:

kiln/devnets/goerlishadow-4.vars
```diff
@@ -1,21 +1,28 @@
...
 
-GETH_EXTRA_ARGS="--http --http.api engine,net,eth,web3 --http.port 8545 --allow-insecure-unlock --http.addr 0.0.0.0 --http.corsdomain \"*\" --http.vhosts \"*\" --authrpc.port=8551 --networkid 5"
+GETH_EXTRA_ARGS="--http --http.api engine,net,eth,web3 --http.port 8545 --allow-insecure-unlock --http.addr 0.0.0.0 --http.corsdomain \"*\" --http.vhosts \"*\" --authrpc.port=8551"
 
 ETHEREUMJS_EXTRA_ARGS="--saveReceipts --rpc --rpcport 8545 --ws --rpcEngine --rpcEnginePort=8551 --rpcDebug --loglevel=debug"
 
+BESU_EXTRA_ARGS="--rpc-http-enabled=true --rpc-http-api=ADMIN,CLIQUE,MINER,ETH,NET,DEBUG,TXPOOL,TRACE --rpc-http-host=0.0.0.0 --rpc-http-port=8545 --Xmerge-support=true --engine-rpc-http-port=8551 --rpc-http-cors-origins=\"*\" --host-allowlist=\"*\" --engine-host-allowlist=\"*\" --p2p-enabled=true --engine-jwt-enabled=true"
```
fab-10, line 25: Suggest to add `--sync-mode=FAST` after this PR https://github.com/hyperledger/besu/pull/3655 is merged.

The same applies to the other vars file below

fab-10 (approved):
LGTM, I have also did a smoke test and it should work fine, just proposed to use fast sync when it is ready for the merge.

philknows (approved):

Muaytie23:
+This proposal extends [EIP-5792](./eip-5792.md) to allow dapps to downgrade their required atomicity guarantees and control the behaviour after a failed/reverted call. It introduces the batch-scope concept of `strict` vs. `loose` atomicity, where a `strict` batch remains atomic in the face of chain reorgs and a `loose` batch does not; and the per-call ability to continue after a failed/reverted call (`continue`) or stop processing (`halt`).
+
+## Motivation
+
+While the base EIP-5792 specification works extremely well for smart contract wallets, it does not allow the expression of the full range of flow control options that wallets can implement. For example, a dapp may only be submitting a batch for gas savings and not care about whether all calls are reverted on failure. A wallet may only be able to offer a limited form of atomicity through block builder backchannels, but that may be sufficient for a trading platform.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### RPC Interface
+
+The following subsections are modifications to the API endpoints from EIP-5792.
+
+If a request does not match the schema defined below, the wallet MUST reject the request with an error code of 0 .
i would prefer to create a params/constants table and refer them by their param/constant name here. brings it out as well as confine the TODO to a particualr section.

similarly with other TODOS

—

Issue #3943, opened 2022-04-22
Submitted attestation is not included in any AggregateAndProof


twoeths:
**Describe the bug**

Validator submitted a timely attestation with correct head, sentPeers = 2

```
Apr-22 01:09:01.632[]                ^[[34mdebug^[[39m: Signed attestation slot=3647143, index=10, head=0x2319727263ef848c02ea96936d3df9ce5a92970dfa66aceace301eb0521fa867, validatorIndex=270132
```

```
Apr-22 01:09:01.758[VMON]            ^[[34mdebug^[[39m: Local validator published unaggregated attestation validatorIndex=270132, slot=3647143, committeeIndex=10, subnet=10, sentPeers=2, delaySec=-1.3670001029968262
```

However it's not included in any AggregateAndProof and the attestation is missed

part of #3527

**Expected behavior**

Attestation is included in an AggregateAndProof and finally a Block

twoeths:
another example of an attestation that's not included in any AggregateAndProof
```
Apr-21 16:07:00.054[]                ^[[34mdebug^[[39m: Signed attestation slot=3644433, index=48, head=0xc073c0e442c4501350da9112b534a5f57e7afb27d1ee937938deab5eeea5c614, validatorIndex=270132
```

```
Apr-21 16:10:37.357[VMON]            ^[[34mdebug^[[39m: Local validator published unaggregated attestation validatorIndex=270132, slot=3644451, committeeIndex=59, subnet=59, sentPeers=8, delaySec=-1.7430000305175781
```

dapplion:
Current master has almost no cases of missed attestations

PR #3944, opened 2022-04-22
Add validator option not to submit attestation early

twoeths:
**Motivation**

As monitored in hetzner-c0, there are 2 missed attestations with:
+ correct head
+ sent peers is 2  or 8
+ delaySec is around -1.5

It's possible that the missed attestations were because we submit them too early, and peers ignore them (maybe due to unknown block root?). Note that lighthouse does not submit attestations early at this time

**Description**

Add an option to submit attestations right at 1/3 of slot so that we can test in mainnet nodes

part of #3943

### DIFF
--- a/packages/cli/src/cmds/validator/options.ts
+++ b/packages/cli/src/cmds/validator/options.ts
@@ -18,7 +18,7 @@ export type IValidatorCliArgs = IAccountValidatorArgs &
     server: string;
     force: boolean;
     graffiti: string;
-    dontSubmitAttestationEarly?: boolean;
+    earlyAttestationDelayMs?: number;
     importKeystoresPath?: string[];
     importKeystoresPassword?: string;
     externalSignerUrl?: string;
@@ -62,10 +62,10 @@ export const validatorOptions: ICliCommandOptions<IValidatorCliArgs> = {
     type: "string",
   },
 
-  dontSubmitAttestationEarly: {
-    description: "Wait for 1/3 of slot to submit attestation, by default submit attestation early if possible",
-    type: "boolean",
-    default: false,
+  earlyAttestationDelayMs: {
+    hidden: true,
+    description: "Delay early attestation in millisecond",
+    type: "number",
   },
 
   importKeystoresPath: {
--- a/packages/cli/src/cmds/validator/options.ts
+++ b/packages/cli/src/cmds/validator/options.ts
@@ -18,6 +18,7 @@ export type IValidatorCliArgs = IAccountValidatorArgs &
     server: string;
     force: boolean;
     graffiti: string;
+    dontSubmitAttestationEarly?: boolean;
     importKeystoresPath?: string[];
     importKeystoresPassword?: string;
     externalSignerUrl?: string;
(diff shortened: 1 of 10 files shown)


packages/cli/src/cmds/validator/options.ts
```diff
@@ -61,6 +62,12 @@ export const validatorOptions: ICliCommandOptions<IValidatorCliArgs> = {
     type: "string",
   },
 
+  dontSubmitAttestationEarly: {
```
wemeetagain: Can we hidden:true this?

dapplion:
Checkout this PR by Nimbus, they do a happy medium, wait at least 2 seconds but still get ahead https://github.com/status-im/nimbus-eth2/pull/3518

packages/cli/src/cmds/validator/options.ts
twoeths: I changed the option to `earlyAttestationDelayMs` and use `hidden:true` for it

packages/validator/src/services/attestation.ts
```diff
@@ -148,13 +154,25 @@ export class AttestationService {
...
       try {
+        const msToOneThirdSlot = this.clock.msToSlot(slot + 1 / 3);
+        const signedAttestationTime = Date.now();
+        // as monitored in hetzner nodes, most of the time we're before 1/3 of slot at this time
+        // i.e. msToOneThirdSlot > 0
+        if (msToOneThirdSlot > 0) {
+          // for early attestations, wait until now + 2000ms, or 1/3 of slot, whichever comes first
+          await sleep(Math.min(msToOneThirdSlot, this.earlyAttestationDelayMs));
+        }
+        this.metrics?.attesterStepCallPublishAttestation.observe(this.clock.secFromSlot(slot + 1 / 3));
```
dadepo: if it is `this.earlyAttestationDelayMs` that comes first, won't this metric be wrong?
edit: No. Metric tracks now (whenever that might be) up till slot + 1/3

dadepo:
Looking at the Nimbus PR [here](https://github.com/status-im/nimbus-eth2/pull/3518/files), initially the wait was 250ms, then it was changed to 1000ms and now 2000ms due to the latency introduced as the network grows. 

Won't a more robust approach be to have beacon clients cache early attestations instead? Not sure if doing this won't conflict with other parts of the spec/implementations

twoeths:
> Looking at the Nimbus PR [here](https://github.com/status-im/nimbus-eth2/pull/3518/files), initially the wait was 250ms, then it was changed to 1000ms and now 2000ms due to the latency introduced as the network grows.

> Won't a more robust approach be to have beacon clients cache early attestations instead? Not sure if doing this won't conflict with other parts of the spec/implementations

This is from the spec
> A validator should create and broadcast the `attestation` to the associated attestation subnet when either (a) the validator has received a valid block from the expected block proposer for the assigned `slot` or (b) `1 / INTERVALS_PER_SLOT` of the `slot` has transpired (`SECONDS_PER_SLOT / INTERVALS_PER_SLOT` seconds after the start of `slot`) -- whichever comes _first_.

so we still follow the spec. I don't think we want to introduce another cache at beacon node side unnecessarily

dapplion (approved):

PR #3955, opened 2022-04-27
Prune logic when we have too many peers

### LINKED ISSUE #3940: PeerManager failed to connect to subnet peers
**Describe the bug**

In hetzner-c0, there is a missed attestation after I deploy master there:
```
Apr-20 02:54:24.422[]                ^[[31merror^[[39m: Error publishing attestations slot=3633270 Internal Server Error: PublishError.InsufficientPeers
Error: Internal Server Error: PublishError.InsufficientPeers
    at HttpClient.request (/usr/app/node_modules/@chainsafe/lodestar-api/src/client/utils/httpClient.ts:109:15)
    at HttpClient.json (/usr/app/node_modules/@chainsafe/lodestar-api/src/client/utils/httpClient.ts:71:12)
    at Object.request [as submitPoolAttestations] (/usr/app/node_modules/@chainsafe/lodestar-api/src/client/utils/client.ts:74:19)
    at AttestationService.signAndPublishAttestations (/usr/app/node_modules/@chainsafe/lodestar-validator/src/services/attestation.ts:156:9)
    at AttestationService.runAttestationTasks (/usr/app/node_modules/@chainsafe/lodestar-validator/src/services/attestation.ts:69:5)
```

PeerManager is supposed to find subnet peers before validator submit subnet attestations but it failed to do that

**Expected behavior**

PeerManager should be able to find and connect to subnet peers before validators submit subnet attestations

twoeths:
**Motivation**

Right now when we have too many peers, we remove peers based on worse scored no-duty peers
+ We lack subnet peers
+ We have peers that have 0 long lived subnet
+ We want to continuously improve the long lived subnets of peer and balance the peer distribution per subnet

**Description**
Implement the prune logic when we have too many peers
- Remove peers that are not subscribed to a subnet and have no duties (they have less value)
- Remove worst scored no-duty peers
- Remove peers that we have many on any particular subne
  - Only consider removing peers on subnet that has > MAX_TARGET_SUBNET_PEERS to be safe
  - Do not remove peer that would drop us below targetPeersPerAttnetSubnet
  - Do not remove peer that would drop us below MIN_SYNC_COMMITTEE_PEERS
  - Choose peers that have less long lived subnet to remove first
 
Some other modification:
 - Number of peers per requested subnet is changed from 1 to 2
 - Due to the real score of lodestar node in mainnet, maxScore is 0 so we don't want to just remove peers with score < 0, I defined `LOW_SCORE_TO_PRUNE_IF_TOO_MANY_PEERS` constant for it, right now make it -2

closes #3940

### DIFF
--- a/packages/lodestar/src/metrics/metrics/lodestar.ts
+++ b/packages/lodestar/src/metrics/metrics/lodestar.ts
@@ -114,6 +114,13 @@ export function createLodestarMetrics(
       help: "network.reportPeer count by reason",
       labelNames: ["reason"],
     }),
+    peerManager: {
+      heartbeatDuration: register.histogram({
+        name: "lodestar_discovery_find_node_query_time_seconds",
+        help: "Time to complete a find node query in seconds in seconds",
+        buckets: [0.001, 0.01, 0.1, 1],
+      }),
+    },
 
     discovery: {
       peersToConnect: register.gauge({
--- a/packages/lodestar/src/metrics/metrics/lodestar.ts
+++ b/packages/lodestar/src/metrics/metrics/lodestar.ts
@@ -94,9 +94,10 @@ export function createLodestarMetrics(
       name: "lodestar_peers_requested_total_to_connect",
       help: "Priorization results total peers count requested to connect",
     }),
-    peersRequestedToDisconnect: register.gauge({
+    peersRequestedToDisconnect: register.gauge<"reason">({
       name: "lodestar_peers_requested_total_to_disconnect",
       help: "Priorization results total peers count requested to disconnect",
+      labelNames: ["reason"],
     }),
     peersRequestedSubnetsToQuery: register.gauge<"type">({
       name: "lodestar_peers_requested_total_subnets_to_query",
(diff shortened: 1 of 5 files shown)


twoeths:
This PR helps maintain a good number of long-lived subnets. This helps tested with mainnet nodes of 16 validators in the last 2 days
- contabo-5 could maintain >= 15 average long-lived subnets per peer, `TARGET_SUBNET_PEERS` is 6
Screen Shot 2022-05-02 at 10 58 16

- contabo-6 could maintain >= 10 average long-lived subnets per peer,  `TARGET_SUBNET_PEERS` is 6
- contabo-7 could maintain >= 20 average long-lived subnets per peer, `TARGET_SUBNET_PEERS` was changed to 12
- contabo-8 could maintain >= 25 average long-lived subnets per peer, `TARGET_SUBNET_PEERS` was changed to 18

dapplion (dismissed):
While logic is complicated, it's a necessary evil. We'll monitor the results and adjust in the future if it makes sense :+1:

dapplion (approved):

PR #3957, opened 2022-04-28
Fix bounded peersToConnect variable

twoeths:
**Motivation**

The `peersToConnect` variable in PeerDiscovery is not bounded and becomes negative most of the time

Screen Shot 2022-04-28 at 10 05 49

**Description**

Make it min 0.

part of #3940

### DIFF
--- a/packages/lodestar/src/network/peers/discover.ts
+++ b/packages/lodestar/src/network/peers/discover.ts
@@ -357,6 +357,8 @@ export class PeerDiscovery {
    * Peers that have been returned by discovery requests are dialed here if they are suitable.
    */
   private async dialPeer(cachedPeer: CachedENR): Promise<void> {
+    this.peersToConnect = Math.max(this.peersToConnect - 1, 0);
+
     const {peerId, multiaddrTCP} = cachedPeer;
 
     // Must add the multiaddrs array to the address book before dialing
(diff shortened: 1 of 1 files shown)


wemeetagain (dismissed):

packages/lodestar/src/network/peers/discover.ts
```diff
@@ -357,7 +357,7 @@ export class PeerDiscovery {
    * Peers that have been returned by discovery requests are dialed here if they are suitable.
    */
   private async dialPeer(cachedPeer: CachedENR): Promise<void> {
-    this.peersToConnect--;
+    this.peersToConnect = Math.max(this.peersToConnect - 1, 0);
```
dadepo, line 366: more of a question for my own understanding. In the attached screenshot, the value of peersToConnect  was in the range of -6000. Question is, should that be happening in the first place? Because this fix might be fixing the symptoms while the core issue of continually calling `dialPeer` which decrease `peersToConnect` till it enters negative value is not fixed...

twoeths: we dial a peer when
- `this.peersToConnect > 0`
- or the peer subscribes to a subnet that we want

if `this.peersToConnect` is 3 while we need to dial 5 subnet peers, in that case we want `this.peersToConnect` to be 0 instead of a negative value. The next heartbeat may increase `this.peersToConnect` again if some dials are not successful

dapplion (dismissed):

dapplion (approved):
Committed @tuyennhv rationale comment for context in the future. Please commit them for non-obvious fixes like this!

PR #3958, opened 2022-04-28
Add ability to update the fee recipient for execution via beacon and/or validator defaults

### LINKED ISSUE #3693: Implementing and calling prepare_beacon_proposer and providing `suggestedFeeRecipient` to the BN
Beacon node needs to provide suggestedFeeRecipient which could be a validator specific field. Since [timing preparePayload](https://github.com/ChainSafe/lodestar/issues/3488 ) could be an independent operation from produceBlock api call by the validator, the BN needs to be aware.

How this can be achieved is in a bit of flux with the debate over if its Beacon/Validator driven. A [`prepare_beacon_proposer endpoint`](https://github.com/ethereum/beacon-APIs/pull/178) has been suggested for the same.. However some issues/race conditions have come up(https://github.com/ethereum/beacon-APIs/issues/187)

Track and discuss how to best implement this in lodestar.

g11tech:
**Motivation**
Post `merge`, a `suggestedFeeRecipient` needs to be supplied per validator index while proposing blocks.

This PR adds the capability to supply a **default** `suggestedFeeRecipient` for 

- a validator process   (could be overridden per validator in a later PR via config or keymanager api)
- as well as for a BN in case the the proposer data is not updated for a validator

and to enable beacon node to maintain and use the updated/registered `suggestedFeeRecipient` while instantiating the block building corresponding to the validator index.

TODOs
- [x] BeaconProposerCache tests
- [x] merge interop beacon default fee recipient check
- [x] merge interop validator default fee recipient check

Closes #3693

### DIFF
--- a/packages/api/src/routes/validator.ts
+++ b/packages/api/src/routes/validator.ts
@@ -13,6 +13,7 @@ import {
   ssz,
   UintNum64,
   ValidatorIndex,
+  ExecutionAddress,
 } from "@chainsafe/lodestar-types";
 import {
   RoutesData,
@@ -42,6 +43,11 @@ export type SyncCommitteeSubscription = {
   validatorIndex: ValidatorIndex;
   syncCommitteeIndices: number[];
   untilEpoch: Epoch;
+};
+
+export type ProposerPreparationData = {
+  validatorIndex: ValidatorIndex;
+  feeRecipient: ExecutionAddress;
 };
 
 export type ProposerDuty = {
@@ -197,6 +203,8 @@ export type Api = {
   prepareBeaconCommitteeSubnet(subscriptions: BeaconCommitteeSubscription[]): Promise<void>;
 
   prepareSyncCommitteeSubnets(subscriptions: SyncCommitteeSubscription[]): Promise<void>;
+
+  prepareBeaconProposer(proposers: ProposerPreparationData[]): Promise<void>;
 };
 
 /**
@@ -215,6 +223,7 @@ export const routesData: RoutesData<Api> = {
   publishContributionAndProofs: {url: "/eth/v1/validator/contribution_and_proofs", method: "POST"},
   prepareBeaconCommitteeSubnet: {url: "/eth/v1/validator/beacon_committee_subscriptions", method: "POST"},
   prepareSyncCommitteeSubnets: {url: "/eth/v1/validator/sync_committee_subscriptions", method: "POST"},
+  prepareBeaconProposer: {url: "/eth/v1/validator/prepare_beacon_proposer", method: "POST"},
 };
 
 /* eslint-disable @typescript-eslint/naming-convention */
@@ -231,6 +240,7 @@ export type ReqTypes = {
   publishContributionAndProofs: {body: unknown};
   prepareBeaconCommitteeSubnet: {body: unknown};
   prepareSyncCommitteeSubnets: {body: unknown};
+  prepareBeaconProposer: {body: unknown};
 };
 
 export function getReqSerializers(): ReqSerializers<Api, ReqTypes> {
@@ -250,6 +260,14 @@ export function getReqSerializers(): ReqSerializers<Api, ReqTypes> {
       validatorIndex: ssz.ValidatorIndex,
       syncCommitteeIndices: ArrayOf(ssz.CommitteeIndex),
       untilEpoch: ssz.Epoch,
+    },
+    {jsonCase: "eth2"}
+  );
+
+  const ProposerPreparationData = new ContainerType(
+    {
+      validatorIndex: ssz.ValidatorIndex,
+      feeRecipient: ssz.ExecutionAddress,
     },
     {jsonCase: "eth2"}
   );
@@ -330,6 +348,7 @@ export function getReqSerializers(): ReqSerializers<Api, ReqTypes> {
     publishContributionAndProofs: reqOnlyBody(ArrayOf(ssz.altair.SignedContributionAndProof), Schema.ObjectArray),
     prepareBeaconCommitteeSubnet: reqOnlyBody(ArrayOf(BeaconCommitteeSubscription), Schema.ObjectArray),
     prepareSyncCommitteeSubnets: reqOnlyBody(ArrayOf(SyncCommitteeSubscription), Schema.ObjectArray),
+    prepareBeaconProposer: reqOnlyBody(ArrayOf(ProposerPreparationData), Schema.ObjectArray),
   };
 }
 
--- a/packages/api/src/routes/validator.ts
+++ b/packages/api/src/routes/validator.ts
@@ -45,6 +45,10 @@ export type SyncCommitteeSubscription = {
   untilEpoch: Epoch;
 };
 
+/**
+ * The types used here are string instead of ssz based because the use of proposer data
+ * is just validator --> beacon json api call for `beaconProposerCache` cache update.
+ */
 export type ProposerPreparationData = {
   validatorIndex: string;
   feeRecipient: string;
--- a/packages/api/src/routes/validator.ts
+++ b/packages/api/src/routes/validator.ts
@@ -13,7 +13,6 @@ import {
   ssz,
   UintNum64,
   ValidatorIndex,
-  ExecutionAddress,
 } from "@chainsafe/lodestar-types";
 import {
   RoutesData,
@@ -24,6 +23,7 @@ import {
   WithVersion,
   reqOnlyBody,
   ReqSerializers,
+  jsonType,
 } from "../utils";
 
 // See /packages/api/src/routes/index.ts for reasoning and instructions to add new routes
@@ -46,8 +46,8 @@ export type SyncCommitteeSubscription = {
 };
 
 export type ProposerPreparationData = {
-  validatorIndex: ValidatorIndex;
-  feeRecipient: ExecutionAddress;
+  validatorIndex: string;
+  feeRecipient: string;
 };
 
 export type ProposerDuty = {
@@ -264,14 +264,6 @@ export function getReqSerializers(): ReqSerializers<Api, ReqTypes> {
     {jsonCase: "eth2"}
   );
 
-  const ProposerPreparationData = new ContainerType(
-    {
-      validatorIndex: ssz.ValidatorIndex,
-      feeRecipient: ssz.ExecutionAddress,
-    },
-    {jsonCase: "eth2"}
-  );
-
   const produceBlock: ReqSerializers<Api, ReqTypes>["produceBlock"] = {
     writeReq: (slot, randaoReveal, grafitti) => ({
       params: {slot},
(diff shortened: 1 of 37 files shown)


packages/cli/src/options/beaconNodeOptions/execution.ts
```diff
@@ -41,4 +54,15 @@ export const options: ICliCommandOptions<ExecutionEngineArgs> = {
     type: "string",
     group: "execution",
   },
+
+  defaultSuggestedFeeRecipient: {
+    description:
+      "Specify fee recipient default for collecting the EL block fees and rewards (a hex string representing 20 bytes address: ^0x[a-fA-F0-9]{40}$) in case validator fails to update for a validator index before calling produceBlock.",
+    default:
+      defaultOptions.executionEngine.mode === "http"
+        ? String(defaultOptions.executionEngine.defaultSuggestedFeeRecipientHex)
```
dapplion: Is String() required here?

packages/cli/src/util/feeRecipient.ts
```diff
@@ -0,0 +1,16 @@
+import {ExecutionAddress} from "@chainsafe/lodestar-types";
+import {fromHex} from "@chainsafe/lodestar-utils";
+
+export function parseFeeRecipientHex(feeRecipientHexString: string): string {
+  const hexPattern = new RegExp(/^(0x|0X)(?<feeRecipientString>[a-fA-F0-9]{40})$/, "g");
+  const feeRecipientStringMatch = hexPattern.exec(feeRecipientHexString);
```
dapplion: Why not do just

```ts
if (!/^0x[a-f0-9]{40}$/i.test(feeRecipientHexString)) {
  // invalid
}
```

are users expected to pass garbage around the address?

packages/api/src/routes/validator.ts
```diff
@@ -254,6 +264,14 @@ export function getReqSerializers(): ReqSerializers<Api, ReqTypes> {
     {jsonCase: "eth2"}
   );
 
+  const ProposerPreparationData = new ContainerType(
+    {
+      validatorIndex: ssz.ValidatorIndex,
+      feeRecipient: ssz.ExecutionAddress,
```
dapplion: Just make this a string, and you can get rid of all the unnecessary type conversions to from hex

packages/lodestar/src/executionEngine/interface.ts
```diff
@@ -63,13 +64,19 @@ export type ApiPayloadAttributes = {
...
+  feeRecipient: ExecutionAddress;
+};
 /**
  * Execution engine represents an abstract protocol to interact with execution clients. Potential transports include:
  * - JSON RPC over network
  * - IPC
  * - Integrated code into the same binary
  */
 export interface IExecutionEngine {
+  proposers: MapDef<ValidatorIndex, {epoch: Epoch; feeRecipient: ExecutionAddress}>;
```
dapplion: Is it really the best place to put this data here? This data is agnostic of transport, and ExecutionEngineDisabled having to be aware of it shows it's maybe not the best place. If this map is unique to each beacon node, why not keep in the BeaconChain class? Does block production have access to it?

packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -580,5 +578,9 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
 
       network.prepareSyncCommitteeSubnets(subs);
     },
+
+    async prepareBeaconProposer(proposers) {
+      await chain.executionEngine.updateProposerPreparation(chain.clock.currentEpoch, proposers);
```
dapplion: Responding to https://github.com/ChainSafe/lodestar/pull/3958/files#r869010716

Yeah just add this data to the BeaconChain class, it doesn't really concern the executionEngine until block production

packages/validator/src/services/block.ts
```diff
@@ -80,7 +80,9 @@ export class BlockProposingService {
         this.metrics?.blockProposingErrors.inc({error: "produce"});
         throw extendError(e, "Failed to produce block");
       });
-      this.logger.debug("Produced block", debugLogCtx);
+      const blockFeeRecipient = (block.data as bellatrix.BeaconBlock).body.executionPayload?.feeRecipient;
+      const feeRecipient = blockFeeRecipient !== undefined ? toHexString(blockFeeRecipient) : undefined;
+      this.logger.debug("Produced block", {...debugLogCtx, feeRecipient});
```
twoeths, line 104: I suppose validator should confirm the `feeReceipient` to address this in this beacon node api spec: `Note that there is no guarantee that the beacon node will use the supplied fee recipient when creating a block proposal, so on receipt of a proposed block the validator should confirm that it finds the fee recipient within the block acceptable before signing it.`

an attacker can periodically poll our api periodically to provide its own addresses

packages/lodestar/src/chain/beaconProposerCache.ts
```diff
@@ -0,0 +1,37 @@
...
+      epoch: 0,
+      feeRecipient: opts.defaultFeeRecipient,
+    }));
+  }
+
+  add(epoch: Epoch, {validatorIndex, feeRecipient}: ProposerPreparationData): void {
+    this.feeRecipientByValidatorIndex.set(validatorIndex, {epoch, feeRecipient});
+  }
+
+  prune(epoch: Epoch): void {
```
twoeths, line 23: normally we use epoch as the key and the `prune()` is much simpler, why do we use `validatorIndex` as the key here?

packages/lodestar/src/chain/beaconProposerCache.ts
```diff
@@ -0,0 +1,37 @@
+import {MapDef} from "../util/map";
+import {Epoch} from "@chainsafe/lodestar-types";
+import {IMetrics} from "../metrics";
+
+const PROPOSER_PRESERVE_EPOCHS = 2;
+export type ProposerPreparationData = {
+  validatorIndex: string;
```
twoeths: why is `validatorIndex` is typed as string? we have our own `ValidatorIndex` type

packages/lodestar/src/chain/beaconProposerCache.ts
twoeths: nvm, we use `validatorIndex` as key because validator only provide feeRecipient for 1 epoch but we could use it for the next 2 epochs

packages/lodestar/src/chain/beaconProposerCache.ts
```diff
@@ -0,0 +1,37 @@
...
+export class BeaconProposerCache {
+  private readonly feeRecipientByValidatorIndex: MapDef<string, {epoch: Epoch; feeRecipient: string}>;
+  constructor(opts: {defaultFeeRecipient: string}, private readonly metrics?: IMetrics | null) {
+    this.feeRecipientByValidatorIndex = new MapDef<string, {epoch: Epoch; feeRecipient: string}>(() => ({
+      epoch: 0,
+      feeRecipient: opts.defaultFeeRecipient,
+    }));
+  }
+
+  add(epoch: Epoch, {validatorIndex, feeRecipient}: ProposerPreparationData): void {
```
twoeths, line 19: we should ~~validator~~ validate `epoch` here to make sure it's not too old

packages/lodestar/src/chain/beaconProposerCache.ts
```diff
@@ -0,0 +1,37 @@
...
+    for (const [validatorIndex, feeRecipientEntry] of this.feeRecipientByValidatorIndex.entries()) {
+      // We only retain an entry for PROPOSER_PRESERVE_EPOCHS epochs
+      if (feeRecipientEntry.epoch + PROPOSER_PRESERVE_EPOCHS < epoch) {
+        this.feeRecipientByValidatorIndex.delete(validatorIndex);
+      }
+    }
+  }
+
+  get(proposerIndex: number | string): string {
+    return this.feeRecipientByValidatorIndex.getOrDefault(`${proposerIndex}`).feeRecipient;
```
twoeths, line 34: if validators don't provide `feeRecipient` while the default value is `0x0000000000000000000000000000000000000000`, do we throw error here?

packages/validator/src/services/block.ts
g11tech: yes correct observation :+1: , one thing to consider is that the block build by mev-boost might not have the payout to `feeRecipient` as block's `feeRecipient` but in form of transaction included last minute. However we can address that in the mev's PR :thinking:

packages/lodestar/src/chain/beaconProposerCache.ts
g11tech: yea i was using that earlier, but Lion suggested  to avoid to/fro conversion between validator<>beacon for beacon proposer data  as this is mostly to populate a lookup index at the BN.

packages/lodestar/src/chain/beaconProposerCache.ts
g11tech: ummm, there has been a discussion regarding this in the R&D channels, mostly opinion-ed that its better not to throw and let the block proposal be successful and validator still earn the rewards.

packages/validator/src/services/block.ts
twoeths: just have a look at `ProduceBlindedBlockResponse`, I suppose we'll still have `feeRecipient` in `BlindedBlock` version, we'll only have transaction root instead of transaction list (correct me if I'm wrong, just follow (https://ethereum.github.io/beacon-APIs/#/Validator/produceBlindedBlock)

packages/lodestar/src/chain/beaconProposerCache.ts
twoeths: do you know how can the proposer still earn the reward if `feeRecipient` is `0x0000000000000000000000000000000000000000`?

packages/lodestar/src/chain/beaconProposerCache.ts
twoeths: 👍 so maybe just drop a comment here

packages/lodestar/src/chain/beaconProposerCache.ts
g11tech: block proposal rewards??

packages/validator/src/services/block.ts
g11tech: yes ofcourse, what i meant was that the mev builders can put their own fee Recipient and instead reward the fee Recipient by including a transaction in block. Will confirm with other teams how they are handling such scenario, but we can address that in MEV PR. right now, i will introduce the check.

packages/lodestar/src/chain/beaconProposerCache.ts
g11tech: `epoch` is actually coming from the clock, i guess we can skip check and avoid adding clock to this cache class. what do you think? :thinking:

twoeths: if it comes from beacon chain clock then it's fine, if it comes from api (not only validator, someone can just issue a `curl` to call the api) then we should check I think

g11tech: no api doesn't specifies this, this is added here: https://github.com/ChainSafe/lodestar/pull/3958/files#diff-b909bf67569ad6ffedfbd4ebe6c611c60e828d24b1e5d04fbafb058abc4eba50R587

twoeths: oh then it's totally fine 👍

twoeths (approved):

wemeetagain (approved):

PR #3964, opened 2022-05-02
ValidatorMonitor cli

twoeths:
**Motivation**
- Prater and mainnet are quite different in term of long lived subnet peers
- Initially this started as a dev command in my local environment to periodically polling `prepareBeaconCommitteeSubnet` so that I can launch a mainnet node and monitor `peers + long lived subnets + connected validators` there with specified validators without having to have real validators
- But then I think it could be helpful in different scenarios:
  - We can compare performance of lodestar validators vs some best-performed validators from other clients
  - We can monitor performance of lodestar validators launched by somebody given their indexes (and maybe they ask us for help to monitor those)
  - We can connect some lodestar validators to other clients' nodes and monitor their performance using lodestar metrics & Dashboard

**Description**

- Implement `ValidatorMonitor` cli, for example: `validatorMonitor --network prater --beaconApiUrl http://localhost:9596 --validatorIndexes 200008..200015`
- Similar to a Validator, we poll `prepareBeaconCommitteeSubnet()` to beacon node per epoch


dapplion:
@tuyennhv If I understand the goal is to track additional validators outside of the regular keys we run. To achieve the same we could modify the beacon node to:
- Provide a list of indexes to track as a BeaconNodeOptions argument. This indexes will be always tracked regardless of calls to prepareBeaconCommitteeSubnet. Then the beacon cli arg could read this indexes from a CLI arg, or from a file.
- Add a new debug or lodestar API endpoint where you pass a list of indexes and a number of epochs to track them, including infinite. We can also add another endpoint to remove indexes.

This approaches will prevent having to modify the validator package. Also since it's very simple to trigger then we won't need another CLI arg. The current name is a bit confusing tho since it's really a "trigger validatorMonitor" instead of just "validatorMonitor"

twoeths:
> @tuyennhv If I understand the goal is to track additional validators outside of the regular keys we run. To achieve the same we could modify the beacon node to:
> 
> * Provide a list of indexes to track as a BeaconNodeOptions argument. This indexes will be always tracked regardless of calls to prepareBeaconCommitteeSubnet. Then the beacon cli arg could read this indexes from a CLI arg, or from a file.
> * Add a new debug or lodestar API endpoint where you pass a list of indexes and a number of epochs to track them, including infinite. We can also add another endpoint to remove indexes.
> 
> This approaches will prevent having to modify the validator package. Also since it's very simple to trigger then we won't need another CLI arg. The current name is a bit confusing tho since it's really a "trigger validatorMonitor" instead of just "validatorMonitor"

that's a great idea 👍 , I'll replace with another PR

twoeths:
@dapplion one thing to add: I started this as a dev command so that I can test Peer Management in mainnet without requiring real validators. Since it calls `prepareBeaconCommitteeSubnet`, it'll request Attestation Subnets periodically. Will keep this for my own testing until our Peer Management is stable.

PR #3965, opened 2022-05-02
Issue advance fcU for builing the EL block

g11tech:
This PR adds the capability to issue advance FcUs using cached beacon proposer data in the engine and cache generated payload ids, which could be picked up while assembling block body in the produceBlock calls.

PS: WIP, and dependent upon https://github.com/ChainSafe/lodestar/pull/3958

TODOS:
- [x] update merge interop tests
 - [x] add check that prepareSlot is the next slot by clock
 - [x] prune payloadIdCache 

Some of the scenario's to issue fcU moved here for allowing this to merge: https://github.com/ChainSafe/lodestar/issues/4054
In these scenarios, since the head prediction might turn out to be wrong, in the produce block call, the previous payload will be not be picked up, but rather an empty payload will be generated by EL which is ok for now as the block proposal will still be successful albeit with the empty payload (current scenario)

### DIFF
--- a/packages/lodestar/src/chain/blocks/importBlock.ts
+++ b/packages/lodestar/src/chain/blocks/importBlock.ts
@@ -1,7 +1,6 @@
 import {ssz} from "@chainsafe/lodestar-types";
 import {SLOTS_PER_EPOCH} from "@chainsafe/lodestar-params";
 import {toHexString} from "@chainsafe/ssz";
-import {allForks} from "@chainsafe/lodestar-types";
 import {
   CachedBeaconStateAllForks,
   CachedBeaconStateAltair,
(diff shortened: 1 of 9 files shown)


packages/lodestar/src/chain/blocks/importBlock.ts
```diff
@@ -212,30 +215,33 @@ export async function importBlock(chain: ImportBlockModules, fullyVerifiedBlock:
...
-     * After BELLATRIX_EPOCH and TTD it's okay to send a zero hash block hash for the finalized block. This will happen if
-     * the current finalized block does not contain any execution payload at all (pre MERGE_EPOCH) or if it contains a
-     * zero block hash (pre TTD)
-     */
-    const finalizedBlockHash = chain.forkChoice.getFinalizedBlock().executionPayloadBlockHash;
-    if (headBlockHash !== null && headBlockHash !== ZERO_HASH_HEX) {
-      chain.executionEngine.notifyForkchoiceUpdate(headBlockHash, finalizedBlockHash ?? ZERO_HASH_HEX).catch((e) => {
-        chain.logger.error("Error pushing notifyForkchoiceUpdate()", {headBlockHash, finalizedBlockHash}, e);
-      });
+  void issueNextProposerEngineFcU(chain, postState).then((payloadId) => {
```
dapplion: Does this `issueNextProposerEngineFcU` call has to happen for all blocks, or only for slot we will propose?

dapplion:
FYI https://github.com/ChainSafe/lodestar/pull/3989 changes the control flow of importBlock and can affect this PR

packages/lodestar/src/chain/blocks/importBlock.ts
g11tech: ~~umm, basically it has to happen for **before**  all blocks because its looks through the propose data cached in engine to see if any the prepare payload call needs to be issued to execution engine before hand.~~
Rephrasing:  `issueNextProposerEngineFcU` checks if the next slot is to be proposed by any registered validators in `beaconProposerCache` and issue fcU accordingly, else it returns `null` and the normal fcU is issued just below to update the `head` of the El.

g11tech:
just tested this build in msf5, successful block production with geth, nethermind, besu:
image

https://beaconchain.mainnetshadowfork5.ethdevops.io/block/22469#transactions

packages/lodestar/src/chain/blocks/importBlock.ts
g11tech: renamed to :`maybeIssueNextProposerEngineFcU` to add a bit more context

wemeetagain (approved):

PR #3969, opened 2022-05-04
Mev/Builder api integration

### LINKED ISSUE #3953: Mev Boost / flashbots integration tracker
Currently its claimed that 80% of the pow mining rate (and hence potentially 80% of the pow blocks) could be MeV build block bundles.
Its proposed to natively integrate MeV with the CL clients for post merge block building, enabling solo stakers to integrate with the MeV ecosystem as well.

This is the tracker to document and track the MeV integration with lodestar
github: https://github.com/flashbots/mev-boost
architecture: https://hackmd.io/@flashbots/HJ3EbzVUY
implemenation overview: https://hackmd.io/@paulhauner/H1XifIQ_t#Change-2-Proposal-Promises
Specification 2.1: https://github.com/flashbots/mev-boost/blob/main/docs/specification.md

Tasks 
- [ ] setup the mev boost 
- [ ] sign and publish fee recipient to mev boost 
- [ ] call getheader on mevboost to get an execution header
- [ ] prepare, sign and publish a blind beacon block and fetch the full payload in response

g11tech:
**Motivation**
Reference: https://github.com/ethereum/execution-apis/pull/209/files
### Dev TODOs
- [x] Refactored to use the new api set ( as well as including the engine merged tracker PRs and esm changes on latest unstable)
- [x] parallel block building of by the validaor using engine + builder
  ~~- [ ] Choice logic to be included~~ moving to a followup issue https://github.com/ChainSafe/lodestar/issues/4208

### Dev Cleanup
- [x] Extract out the builder from the execution folder 
- [x] Cleanup the blinded beacon block's handling in beacon-state-transition for building state root post block body fetch
- [x] API req/resp tests in the unit tests
- [x] gasLimit constant handling
- [x] validator `builder.enabled` option
- [x] fix tests

###  Tests with mergemock
- [x] Signature validation over new domain
  - [x] of registerValidator
  - [x] of submit blinded block
- [x] Register validator
- [x] Get payload header
- [x] submit blinded block and get payload
- [x] engine accepting the new payload
- [x] Successful post merge scenario

### Tests with the full mev builder
- [x] Signature validation over new domain
  - [x] of registerValidator
  - [x] of submit blinded block
- [x] Register validator
- [x] Get payload header
- [x] submit blinded block and get payload
- [x] engine accepting the new payload
- [x] Successful post merge scenario

### Mergemock/local testnet relay integration with merge-interop
~~- [ ] path needs to be figured out yet~~ https://github.com/ChainSafe/lodestar/issues/4208

## Old work
This PR integrates the builder api and consists broadly of 3 parts:
- [x] container types
- [x] Builder module
   - [x] Options
      - [x] builder.urls, builder.enabled
   - [x] ~~Json rpc setup~~ It was decided to implement the api CL style
   - [x] builder api setup
   - [x] methods & types for
     - [x] builder_registerValidatorV1
     - [x] builder_getHeaderV1
     - [x] builder_getPayloadV1
- [x] validator registration : send signed `ValidatorRegistrationV1` in beacon proposer api which BN can propagate
- [x] perform `builder_registerValidatorV1` 
- [x] NOTE: there is no fcU to be issued here, mev ecosystem is expected to take of this using their own nodes
- [x] on produce block perform `builder_getHeaderV1` in parallel with fetching payload from engine while assembling body to get the `ExecutionPayloadHeaderV1` and signed `BuilderBidV1`
- [x] for a `BlindedSignedBeaconBlock` and send `builder_getPayloadV1` and get `ExecutionPayloadV1` which can be used be combined with the signed `BlindedBeaconBlock` to form SignedBeaconBlock and propogated
- [x] Validator checks to confirm and decide fee recipient on BlindedSignedBeaconBlock is what is updated with it before signing it for the publish via builder_getPayloadV1

Closes #3953 

UPDATE: integration/test SUCCESS with `mergemock` !!! :partying_face:  locally with some caveats
```bash
EL 65190: time="2022-05-16T17:31:23+05:30" level="info" msg="computed block from payload" blockHash="0xdcc802335d5f5919d3e76bf062a544ae7eeec2b6ece4ab4deb3645ea7c809da9" difficulty="0" extraData="" gasLimit="30000000" gasUsed="0" logsBloom="[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]" miner="0x0000000000000000000000000000000000000000" number="14" parentHash="0x14f76e1dedca806e74bec62596e0c0d34b2e2b2284306c5033a8c016932f7260" prevRandao="0x19404254604afb25269744cc999d21966f5476a78802ae0f2fb78f74ed1a1f01" receiptsRoot="0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421" sha3Uncles="0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347" stateRoot="0xca3149fa9e37db08d1cd49c9061db1002ef1cd58db2210f2115c8c989b2bdf45" timestamp="1652702483" transactionsRoot="0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421"
EL 65190: time="2022-05-16T17:31:23+05:30" level="info" msg="Imported new chain segment" blocks="1" dirty="0.00 B" elapsed="104.714µs" hash="0xdcc802335d5f5919d3e76bf062a544ae7eeec2b6ece4ab4deb3645ea7c809da9" mgas="0" mgasps="0" number="14" txs="0"
EL 65190: time="2022-05-16T17:31:23+05:30" level="info" msg="Executed payload" block_hash="0xdcc802335d5f5919d3e76bf062a544ae7eeec2b6ece4ab4deb3645ea7c809da9"
Eph 0/14 0.241[NODE-A]           info: > Block attestations slot=14, bits=1, inclDelay=1
Eph 0/14 0.241[VALI 0-31]        info: Published block slot=14, validator=0xa98e…1d99, graffiti=
EL 65190: time="2022-05-16T17:31:23+05:30" level="info" msg="Forkchoice updated" attributes="&{1652702485 0x4b75ecf2383e2bb573394d983975d282af5c9e26842f6cc3b235aa518a2111a6 0x0000000000000000000000000000000000000000}" finalized="0xccaf66b50e791f95d4b50bae4de28af9396824e7c29f99aeba19414fdf72673f" head="0xdcc802335d5f5919d3e76bf062a544ae7eeec2b6ece4ab4deb3645ea7c809da9" safe="0xdcc802335d5f5919d3e76bf062a544ae7eeec2b6ece4ab4deb3645ea7c809da9"
time="2022-05-16T17:31:23+05:30" level="info" msg="Preparing new payload" attributes="&{1652702485 0x4b75ecf2383e2bb573394d983975d282af5c9e26842f6cc3b235aa518a2111a6 0x0000000000000000000000000000000000000000}" payload_id="[0 0 0 0 0 0 0 15]"
Eph 0/14 0.349[VALI 0-31]        info: Published attestations slot=14, count=1
Eph 0/14 0.887[VALI 0-31]        info: Published SyncCommitteeMessage slot=14, count=32
Eph 0/14 1.000[NODE-A]           info: Synced - slot: 14 - head: 14 0x1f5d…7484 - execution: valid(0xdcc8…9da9) - finalized: 0x0000…0000:0 - peers: 0
```
Caveats
1. Had to override the transactionsRoot i was getting with the self computed one (mentioned before that the transactionsRoot is not matching vs locally computed for empty transactions array)
2. Running the test on mainnet preset, would be good if the mergemock can support minimal sync_commitee_bits is the variable thats plays spoils-sport while serializing/deserializing json requests
3. Had to turn off the signature validations on mergemock's submit blinded blocks because i)transactionsRoot issue ii) the blinded beacon block's signature was required on domain, otherwise the  full block reconstructed from the blinded beacon block + payload  wasn't getting published/accepted


packages/lodestar/src/executionEngine/index.ts
```diff
@@ -35,3 +42,11 @@ export function initializeExecutionEngine(opts: ExecutionEngineOpts, signal: Abo
...
   }
 }
+
+export function initializeExecutionBuilder(opts: ExecutionBuilderOpts, signal: AbortSignal): IExecutionBuilder {
+  switch (opts.mode) {
+    case "http":
+    default:
+      return new ExecutionBuilderHttp(opts, signal);
+  }
+}
```
dapplion: Should this be moved to `packages/lodestar/src/executionBuilder/index.ts` while re-using code from `packages/lodestar/src/executionEngine/*` if necessary?

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -29,26 +37,46 @@ export class PrepareBeaconProposerService {
...
-          this.logger.error("Error on prepareBeaconProposer", {epoch}, e);
+          this.logger.error("Error on prepareBeaconProposer", {slot}, e);
         }),
 
       // At the same time fetch any remaining unknown validator indices, then poll duties for those newIndices only
       this.indicesService
         .pollValidatorIndices()
-        .then((newIndices) => this.api.validator.prepareBeaconProposer(this.getProposerData(newIndices)))
+        .then((newIndices) => this.getProposerData(slot, newIndices))
+        .then((proposerData) => this.api.validator.prepareBeaconProposer(proposerData))
```
dapplion: At this point just use an async function instead of a then chain, for consistency with the rest of the codebase

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -29,26 +37,46 @@ export class PrepareBeaconProposerService {
...
-          this.logger.error("Error on poll indices and prepareBeaconProposer", {epoch}, e);
+          this.logger.error("Error on poll indices and prepareBeaconProposer", {slot}, e);
         }),
     ]);
   };
 
-  private getProposerData(indices: number[]): ProposerPreparationData[] {
-    return indices.map((validatorIndex) => ({validatorIndex, feeRecipient: this.defaultSuggestedFeeRecipient}));
-  }
+  private getProposerData = async (slot: Slot, indices: number[]): Promise<ProposerPreparationData[]> => {
```
dapplion: Why are this arrow functions? Do you really need binding to this?

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -29,26 +37,46 @@ export class PrepareBeaconProposerService {
...
+    const proposerData: ProposerPreparationData[] = [];
+    for (const validatorIndex of indices) {
+      const pubkeyHex = this.indicesService.index2pubkey.get(validatorIndex);
+      if (!pubkeyHex) throw Error(`Pubkey lookup failure for validatorIndex=${validatorIndex}`);
```
dapplion: Give some whitespace and breathing room to the function body

```suggestion
      if (!pubkeyHex) {
        throw Error(`Pubkey lookup failure for validatorIndex=${validatorIndex}`);
      }
```

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -29,26 +37,46 @@ export class PrepareBeaconProposerService {
...
         }),
     ]);
   };
 
-  private getProposerData(indices: number[]): ProposerPreparationData[] {
-    return indices.map((validatorIndex) => ({validatorIndex, feeRecipient: this.defaultSuggestedFeeRecipient}));
-  }
+  private getProposerData = async (slot: Slot, indices: number[]): Promise<ProposerPreparationData[]> => {
+    const proposerData: ProposerPreparationData[] = [];
+    for (const validatorIndex of indices) {
```
dapplion: To allow paralelizing requests to a remote signer (`signValidatorRegistration`), run this in `Promise.all(indices.map(...))`

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -29,26 +37,46 @@ export class PrepareBeaconProposerService {
...
+      const pubkeyHex = this.indicesService.index2pubkey.get(validatorIndex);
+      if (!pubkeyHex) throw Error(`Pubkey lookup failure for validatorIndex=${validatorIndex}`);
+      const feeRecipient = this.defaultSuggestedFeeRecipient;
+      const gasLimit = 10000;
```
dapplion: Why hardcode to 10000? Put this in an all caps constant at the top of the file with a comment explaining what will we do with this placeholder value. Like the zero feeRecipient

packages/validator/src/services/validatorStore.ts
```diff
@@ -284,6 +287,27 @@ export class ValidatorStore {
...
+    pubkey: BLSPubkey,
+    feeRecipient: ExecutionAddress,
+    gasLimit: number,
+    slot: Slot
+  ): Promise<bellatrix.SignedValidatorRegistrationV1> {
+    const domain = this.config.getDomain(DOMAIN_APPLICATION_MEV_BOOST, slot);
+    const timestamp = Math.floor(Date.now() / 1000);
+    const validatorRegistation: bellatrix.ValidatorRegistrationV1 = {
+      feeRecipient,
+      gasLimit: 10000,
```
dapplion: Use argument above, not hardcoded value

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -29,26 +37,46 @@ export class PrepareBeaconProposerService {
     clock.runEverySlot(this.prepareBeaconProposer);
   }
 
-  private prepareBeaconProposer = async (epoch: Epoch): Promise<void> => {
+  private prepareBeaconProposer = async (slot: Slot): Promise<void> => {
```
dapplion: Do we really need to run this every slot? With high validator counts that a lot of useless signatures.

packages/lodestar/src/executionEngine/index.ts
g11tech: yes makes sense, will move :slightly_smiling_face:

g11tech:
thanks @dapplion  for looking through :heart:, yes need to refactor as the bn <> boost interaction is now CL api style. 

will incorporate the feedback once am able to successfully implement the flow :slightly_smiling_face:

twoeths:
Builder spec was moved to https://github.com/ethereum/builder-specs

g11tech:
Update successful lodestar <> kiln builder
image

g11tech:
update: successful lodestar <> ropsten builder
image

packages/validator/src/services/block.ts
```diff
@@ -13,8 +13,13 @@ import {BlockDutiesService, GENESIS_SLOT} from "./blockDuties.js";
 type BlockProposingServiceOpts = {
   graffiti?: string;
   strictFeeRecipientCheck?: boolean;
+  builder: {enabled?: boolean};
 };
 
+type WrappedProduceBlockType =
+  | {type: allForks.BlockType.Full; data: allForks.FullOrBlindedBeaconBlock<allForks.BlockType.Full>}
+  | {type: allForks.BlockType.Blinded; data: allForks.FullOrBlindedBeaconBlock<allForks.BlockType.Blinded>};
```
dapplion: This `type` is from the spec or a wrapper here for convenience?

g11tech: for convenience :slightly_smiling_face:

packages/validator/src/services/prepareBeaconProposer.ts
g11tech: now runs every epoch :+1:

packages/beacon-state-transition/src/block/processExecutionPayload.ts
```diff
@@ -1,13 +1,14 @@
...
+import {bellatrix, ssz, allForks} from "@chainsafe/lodestar-types";
 import {toHexString, byteArrayEquals} from "@chainsafe/ssz";
 import {CachedBeaconStateBellatrix} from "../types.js";
 import {getRandaoMix} from "../util/index.js";
 import {ExecutionEngine} from "../util/executionEngine.js";
 import {isMergeTransitionComplete} from "../util/bellatrix.js";
 
-export function processExecutionPayload(
+export function processExecutionPayload<T extends allForks.BlockType>(
+  type: T,
```
dapplion: If would be nice if we can avoid having to pass this type everywhere

packages/beacon-state-transition/src/block/processExecutionPayload.ts
```diff
@@ -46,10 +47,19 @@ export function processExecutionPayload(
...
+    executionEngine &&
+    !executionEngine.notifyNewPayload(payload as bellatrix.ExecutionPayload)
+  ) {
     throw Error("Invalid execution payload");
   }
 
+  const transactionsRoot =
+    type === allForks.BlockType.Blinded
+      ? (payload as bellatrix.ExecutionPayloadHeader).transactionsRoot
+      : ssz.bellatrix.Transactions.hashTreeRoot((payload as bellatrix.ExecutionPayload).transactions);
```
dapplion: This can be solved without the type with type guards

```ts
function getTransactionsRoot(block: BlockFull | BlockBlinded): Root {
  if ((block as BlockBlinded).transactionsRoot) {
    return (block as BlockBlinded).transactionsRoot
  } else {
    // hashTreeRoot
  }
}

File: packages/beacon-state-transition/src/util/bellatrix.ts [hunk 12]
```diff
@@ -10,10 +10,22 @@ import {
...
+  state: BeaconStateBellatrix,
+  body: allForks.FullOrBlindedBellatrixBeaconBlockBody<T>
+): boolean {
   return (
     isMergeTransitionComplete(state) ||
-    !ssz.bellatrix.ExecutionPayload.equals(body.executionPayload, ssz.bellatrix.ExecutionPayload.defaultValue())
+    (type === allForks.BlockType.Blinded
+      ? !ssz.bellatrix.ExecutionPayloadHeader.equals(
+          (body as bellatrix.BlindedBeaconBlockBody).executionPayloadHeader,
+          ssz.bellatrix.ExecutionPayloadHeader.defaultValue()
```
dapplion: This code (before this PR) is really inneficient. For each block it has to:
- Create a new data structure zero'ed from scratch
- Do a full comparison of the data structure with the block's one

Can we do better an maybe avoid having to know if the block is blinded or not? Ideas:
- Cache and compare the root of ExecutionPayloadHeader which is equivalent blinded or full
- Compare only some fields, that if they are zero, everything else must be zero (like stateRoot)

packages/cli/src/options/beaconNodeOptions/index.ts
```diff
@@ -5,6 +5,7 @@ import * as api from "./api.js";
 import * as chain from "./chain.js";
 import * as eth1 from "./eth1.js";
 import * as execution from "./execution.js";
+import * as builder from "./builder.js";
```
dapplion: Please in long list of imports sort alphabetically to ease checking all options are present

packages/cli/src/options/beaconNodeOptions/builder.ts
```diff
@@ -0,0 +1,42 @@
...
+export const options: ICliCommandOptions<ExecutionBuilderArgs> = {
+  "builder.enabled": {
+    description: "Enable builder interface",
+    type: "boolean",
+    defaultDescription: `${
+      defaultOptions.executionBuilder.mode === "http" ? defaultOptions.executionBuilder.enabled : false
+    }`,
+    group: "builder",
+  },
+  "builder.urls": {
```
dapplion, line 28: Have consistent space between top level object properties

packages/lodestar/src/api/impl/beacon/blocks/index.ts
```diff
@@ -1,4 +1,6 @@
 import {routes} from "@chainsafe/lodestar-api";
+
+// eslint-disable-next-line no-restricted-imports
```
dapplion: Why is this line necessary?

packages/lodestar/src/network/gossip/gossipsub.ts
```diff
@@ -160,7 +160,7 @@ export class Eth2Gossipsub extends Gossipsub {
     const topicStr = this.getGossipTopicString(topic);
     const sszType = getGossipSSZType(topic);
     const messageData = (sszType.serialize as (object: GossipTypeMap[GossipType]) => Uint8Array)(object);
-    const sentPeers = await this.publish(topicStr, messageData);
+    const sentPeers = await this.publish(topicStr, messageData).catch((_e) => 0);
```
dapplion: This looks really dangerous, we don't want to swallow errors like this unless there's a really good reason. If there is please add a nice reasoning above explaining why is it okay to silence this call

packages/validator/src/services/block.ts
```diff
@@ -77,50 +82,109 @@ export class BlockProposingService {
...
+          return null;
+        })
+      : null;
+
+    const fullBlockPromise = this.produceBlock(slot, randaoReveal, graffiti).catch((e: Error) => {
+      this.logger.error("Failed to produce builder block", {}, e as Error);
+      return null;
+    });
+
+    await Promise.all([blindedBlockPromise, fullBlockPromise]);
```
dapplion, line 131: Could you explain in great detail the strategy here? From what I can see you attempt to produce blocks and publish one prioritizing the blindedBlock.

But, what if the `produceBlindedBlock` take too long to resolve? You will publish the regular block very late.

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -1,47 +1,71 @@
...
+    builderOpts: {enabled?: boolean}
   ) {
+    const currentEpoch = clock.getCurrentEpoch();
+
     clock.runEveryEpoch(this.prepareBeaconProposer);
+    void this.prepareBeaconProposer(currentEpoch);
+
+    if (builderOpts?.enabled) {
+      clock.runEveryEpoch(this.registerValidator);
+      void this.registerValidator(currentEpoch);
```
dapplion: Explain that `runEveryEpoch(this.registerValidator)` may take a while to run and requires to be run immediately because?

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -1,47 +1,71 @@
...
         }),
     ]);
   };
 
+  private prepareBeaconProposerInChunks = async (
+    proposers: routes.validator.ProposerPreparationData[]
+  ): Promise<void> => {
+    const proposerChunks = spliceIntoChunks(proposers, REGISTRATION_CHUNK_SIZE);
+    await Promise.all(proposerChunks.map((proposerChunk) => this.api.validator.prepareBeaconProposer(proposerChunk)));
+    return;
```
dapplion: What's the point of this return?

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -1,47 +1,71 @@
...
           this.logger.error("Error on poll indices and prepareBeaconProposer", {epoch}, e);
         }),
     ]);
   };
 
+  private prepareBeaconProposerInChunks = async (
+    proposers: routes.validator.ProposerPreparationData[]
+  ): Promise<void> => {
+    const proposerChunks = spliceIntoChunks(proposers, REGISTRATION_CHUNK_SIZE);
+    await Promise.all(proposerChunks.map((proposerChunk) => this.api.validator.prepareBeaconProposer(proposerChunk)));
```
dapplion: Would it better to not spam the beacon node with all chunks at once, but do this sequentially?

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -51,4 +75,64 @@ export class PrepareBeaconProposerService {
...
+          fromHexString(feeRecipient),
+          this.validatorStore.gasLimit,
+          slot
+        );
+      })
+    );
+  };
+}
+
+function spliceIntoChunks<T>(arr: T[], chunkSize: number): T[][] {
```
dapplion: Move to validator/src/utils and add many unit tests

packages/lodestar/src/network/gossip/gossipsub.ts
g11tech: sorry this a typo that got introduced while i was debugging

packages/validator/src/services/block.ts
g11tech: we can set the timeouts  for the builder api to be small? lets say 1/6 of the slot, since proposer boost is 1/3 of the slot.

packages/validator/src/services/prepareBeaconProposer.ts
g11tech: because the validators could have a turn proposing before next slot's registration

packages/validator/src/services/prepareBeaconProposer.ts
g11tech: will remove

packages/validator/src/services/prepareBeaconProposer.ts
g11tech: sure makes sense :+1:

packages/validator/src/services/validatorStore.ts
```diff
@@ -290,6 +299,27 @@ export class ValidatorStore {
...
+    _slot: Slot
+  ): Promise<bellatrix.SignedValidatorRegistrationV1> {
+    const timestamp = Math.floor(Date.now() / 1000);
+    const validatorRegistation: bellatrix.ValidatorRegistrationV1 = {
+      feeRecipient,
+      gasLimit,
+      timestamp,
+      pubkey,
+    };
+    const domain = computeDomain(DOMAIN_APPLICATION_BUILDER, this.config.GENESIS_FORK_VERSION, ZERO_HASH);
```
twoeths, line 363: Why do we go with genesis fork here?

shouldn't it be `const domain = this.config.getDomain(DOMAIN_APPLICATION_BUILDER, slot);`

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -51,4 +75,64 @@ export class PrepareBeaconProposerService {
...
+  };
+
+  private registerValidatorInChunks = async (
+    registrations: bellatrix.SignedValidatorRegistrationV1[]
+  ): Promise<void> => {
+    const registrationChunks = spliceIntoChunks(registrations, REGISTRATION_CHUNK_SIZE);
+    await Promise.all(
+      registrationChunks.map((registrationChunk) => this.api.validator.registerValidator(registrationChunk))
+    );
+    return;
```
twoeths: no need `return;` here

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -51,4 +75,64 @@ export class PrepareBeaconProposerService {
       ),
     }));
   }
+
+  private registerValidator = async (epoch: Epoch): Promise<void> => {
+    if (epoch < this.config.BELLATRIX_FORK_EPOCH - 1) return;
+    this.getValidatorRegistrations(epoch, this.indicesService.getAllLocalIndices())
```
twoeths: why do we have to register all validators for every epoch?

packages/lodestar/src/api/impl/beacon/blocks/index.ts
```diff
@@ -135,6 +137,13 @@ export function getBeaconBlockApi({
       return {data: config.getForkTypes(block.message.slot).BeaconBlock.hashTreeRoot(block.message)};
     },
 
+    async publishBlindedBlock(signedBlindedBlock) {
+      const exeutionBuilder = chain.executionBuilder;
+      if (!exeutionBuilder) throw Error("exeutionBuilder required to publish SignedBlindedBeaconBlock");
+      const signedBlock = await exeutionBuilder.submitSignedBlindedBlock(signedBlindedBlock);
+      return await this.publishBlock(signedBlock);
```
twoeths, line 143: do we need to double check the returned payload is respective to the payload header we got in the blinded block in order not to spread mismatch state root block to the network?

also maybe we need some log here so that we can trace any issues if happen later?

packages/lodestar/src/api/impl/beacon/blocks/index.ts
```diff
@@ -135,6 +137,13 @@ export function getBeaconBlockApi({
       return {data: config.getForkTypes(block.message.slot).BeaconBlock.hashTreeRoot(block.message)};
     },
 
+    async publishBlindedBlock(signedBlindedBlock) {
+      const exeutionBuilder = chain.executionBuilder;
```
twoeths: typo

packages/validator/src/services/prepareBeaconProposer.ts
dapplion: Not to me, but in a comment to the source code :pray:

packages/lodestar/src/api/impl/beacon/blocks/index.ts
dapplion:
```suggestion
      const executionBuilder = chain.executionBuilder;
```

packages/validator/src/services/validatorStore.ts
g11tech: thats as per the spec, i think the idea being that the validator registration doesn't need any fork safeguards

packages/validator/src/services/prepareBeaconProposer.ts
g11tech: because a validator registration might only be valid for 2 epochs on the builder, its just simpler to blindly register validator with them every epoch. :thinking:

packages/lodestar/src/api/impl/beacon/blocks/index.ts
g11tech: makes sense, can add the check in the execution/builder.ts itself for this to stay clean :+1:

packages/lodestar/src/api/impl/beacon/blocks/index.ts
g11tech: removed, may be got included with some rebasing artifacts

packages/validator/src/services/prepareBeaconProposer.ts
g11tech: replaced with already defined (with tests) `batchItems`

packages/validator/src/services/prepareBeaconProposer.ts
g11tech: modeled changes on the rebased function

packages/cli/src/options/beaconNodeOptions/index.ts
g11tech: done

packages/validator/src/services/prepareBeaconProposer.ts
g11tech: done

packages/lodestar/src/api/impl/beacon/blocks/index.ts
g11tech: added in the builder

packages/beacon-state-transition/src/block/processExecutionPayload.ts
g11tech: changed something on these lines everywhere except assemble block :+1:

packages/beacon-state-transition/src/util/bellatrix.ts
g11tech: tackled for blinded vs full by body inspection :+1: 
For efficiency issue would prefer to do a separate PR for efficiency , building on your idea: may be we can hash tree root and pass around the execution payload hash (as its anyway going to be computed/used in forkchoice if everything passes up good). similarly for cachedbeaconstate. if you like the idea, can open an issue and PR on it. :slightly_smiling_face:

packages/beacon-state-transition/src/block/processExecutionPayload.ts
g11tech: changed, now conditioned on the body's execution payload/header inspection of the block :+1:

packages/beacon-state-transition/src/block/index.ts
```diff
@@ -29,12 +29,14 @@ export function processBlock(
...
-      processExecutionPayload(
-        state as CachedBeaconStateBellatrix,
-        (block as bellatrix.BeaconBlock).body.executionPayload,
-        executionEngine
-      );
+    const fullOrBlindedPayload = ((block as bellatrix.BlindedBeaconBlock).body.executionPayloadHeader ??
+      (block as bellatrix.BeaconBlock).body.executionPayload) as allForks.FullOrBlindedExecutionPayload;
+
+    const fullOrBlindedBody = ((block as bellatrix.BlindedBeaconBlock).body ??
+      (block as bellatrix.BeaconBlock).body) as allForks.FullOrBlindedBellatrixBeaconBlockBody;
```
wemeetagain: Is this intended? It looks like without type assertions, this reads like `block.body ?? block.body`

packages/beacon-state-transition/src/block/processBlockHeader.ts
```diff
@@ -37,13 +37,18 @@ export function processBlockHeader(state: CachedBeaconStateAllForks, block: allF
     );
   }
 
+  // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+  const bodyRoot = (block.body as bellatrix.BlindedBeaconBlockBody).executionPayloadHeader
```
wemeetagain: maybe better to just do a nullish check instead of suppress linter

packages/beacon-state-transition/src/signatureSets/proposer.ts
```diff
@@ -1,32 +1,33 @@
...
   state: CachedBeaconStateAllForks,
-  signedBlock: allForks.SignedBeaconBlock
+  signedBlock: allForks.FullOrBlindedSignedBeaconBlock
 ): ISignatureSet {
   const {config, epochCtx} = state;
   const domain = state.config.getDomain(DOMAIN_BEACON_PROPOSER, signedBlock.message.slot);
 
+  // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+  const blockType = (signedBlock.message as bellatrix.BlindedBeaconBlock).body.executionPayloadHeader
+    ? ssz.bellatrix.BlindedBeaconBlock
```
wemeetagain, line 23: We should add the blinded variants to `getForkTypes` somehow.

packages/beacon-state-transition/src/block/index.ts
g11tech: good eye @wemeetagain! thanks :slightly_smiling_face:

packages/beacon-state-transition/src/block/processBlockHeader.ts
g11tech: done! :+1:

packages/beacon-state-transition/src/signatureSets/proposer.ts
g11tech: can I pick this in the followup ? Will add this to a followup issue tracker among few others i think. 
- https://github.com/ChainSafe/lodestar/issues/4208

packages/validator/src/services/block.ts
g11tech: as discussed over discord, we can just go with this for now and try to come up with a better strategy.
- https://github.com/ChainSafe/lodestar/issues/4208

packages/beacon-state-transition/src/block/processExecutionPayload.ts
```diff
@@ -46,10 +46,18 @@ export function processExecutionPayload(
   // if executionEngine is null, executionEngine.onPayload MUST be called after running processBlock to get the
   // correct randao mix. Since executionEngine will be an async call in most cases it is called afterwards to keep
   // the state transition sync
-  if (executionEngine && !executionEngine.notifyNewPayload(payload)) {
+  if (
+    (payload as bellatrix.ExecutionPayload).transactions !== undefined &&
```
dapplion: Unless it looks really annoying this should be a type guard. Like

```ts

function isExecutionPayload(
  payload: bellatrix.ExecutionPayload | bellatrix.ExecutionPayloadBlinded
): payload is bellatrix.ExecutionPayload {
  return (payload as bellatrix.ExecutionPayload).transactions !== undefined
}

if (isExecutionPayload(payload) && !executionEngine.notifyNewPayload(payload as bellatrix.ExecutionPayload)) {
      throw Error("Invalid execution payload");
}
```

packages/beacon-state-transition/src/block/processExecutionPayload.ts
```diff
@@ -46,10 +46,18 @@ export function processExecutionPayload(
...
+    executionEngine &&
+    !executionEngine.notifyNewPayload(payload as bellatrix.ExecutionPayload)
+  ) {
     throw Error("Invalid execution payload");
   }
 
+  const transactionsRoot =
+    (payload as bellatrix.ExecutionPayloadHeader).transactionsRoot ??
+    ssz.bellatrix.Transactions.hashTreeRoot((payload as bellatrix.ExecutionPayload).transactions);
```
dapplion: Same here, use type guards

packages/beacon-state-transition/src/signatureSets/proposer.ts
```diff
@@ -1,32 +1,33 @@
...
 export function getProposerSignatureSet(
   state: CachedBeaconStateAllForks,
-  signedBlock: allForks.SignedBeaconBlock
+  signedBlock: allForks.FullOrBlindedSignedBeaconBlock
 ): ISignatureSet {
   const {config, epochCtx} = state;
   const domain = state.config.getDomain(DOMAIN_BEACON_PROPOSER, signedBlock.message.slot);
 
+  const blockType =
+    (signedBlock.message as bellatrix.BlindedBeaconBlock).body.executionPayloadHeader !== undefined
```
dapplion: Use type guards

packages/beacon-state-transition/src/stateTransition.ts
```diff
@@ -18,13 +18,13 @@ import {processEpoch} from "./epoch/index.js";
...
   state: CachedBeaconStateAllForks,
-  signedBlock: allForks.SignedBeaconBlock,
+  signedBlock: allForks.FullOrBlindedSignedBeaconBlock,
   options?: {verifyStateRoot?: boolean; verifyProposer?: boolean; verifySignatures?: boolean},
   metrics?: IBeaconStateTransitionMetrics | null
 ): CachedBeaconStateAllForks {
   const {verifyStateRoot = true, verifyProposer = true, verifySignatures = true} = options || {};
 
-  const block = signedBlock.message;
+  const block = signedBlock.message as allForks.FullOrBlindedBeaconBlock;
```
dapplion: Is this casting necessary? If so why?

packages/beacon-state-transition/src/util/bellatrix.ts
```diff
@@ -10,10 +10,21 @@ import {
...
  */
-export function isExecutionEnabled(state: BeaconStateBellatrix, body: bellatrix.BeaconBlockBody): boolean {
+export function isExecutionEnabled(
+  state: BeaconStateBellatrix,
+  body: allForks.FullOrBlindedBellatrixBeaconBlockBody
+): boolean {
   return (
     isMergeTransitionComplete(state) ||
-    !ssz.bellatrix.ExecutionPayload.equals(body.executionPayload, ssz.bellatrix.ExecutionPayload.defaultValue())
+    ((body as bellatrix.BlindedBeaconBlockBody).executionPayloadHeader !== undefined
```
dapplion: Use type guards

dapplion:
I think we can have two directories named `executionEngine` and `executionBuilder` instead of both in the same `execution` dir. Less diff and there's always the opportunity to merge them once executionBuilder is final

dapplion (approved):
LGMT! @g11tech merging so be able to test live, let's keep iterating on it after the fact

PR #3973, opened 2022-05-04
Making comparison between local/remote BeaconParam less error prone. 

### LINKED ISSUE #3962: lodestar validator doesn't interop with infura BN
Context: Dhruv from Obol network is trying to integrate lodestar validator. For this they are hooking it up to an infura BN (most likely lighthouse) which is giving the initialization error:

image

on suggesting to pass `--paramsFile ` to match the config, it still didn't work:
> I tried overriding the config to match the with infura's response but as you can see in the above error screenshot, infura's response doesn't contain 2 fields: SHARDING_FORK_VERSION and SHARDING_FORK_EPOCH. 
For these 2 fields I need to override them to undefined but that is not possible because lodestar expects strings from config file. So I got parsing error.

>And for DEPOSIT_CONTRACT_ADDRESS I saw that lodestar converts this into byte array and then at the time of comparison it again converts it to hex string which eventually becomes what it is there in screenshot so overriding that is also not possible. 

Expectation: fix the config matching in the validator to make it interop with other BNs

dadepo:
**Motivation**

In issue #3962, it was surfaced that comparison between local/remote BeaconParam fails if:
- local config contains properties that is missing in remote
- difference in case of values. 

**Description**

The PR fills missing properties in remote config with default values specified in the spec before comparison. It also normalises the case of values before comparison.

Closes #3962

### DIFF
--- a/packages/validator/src/util/params.ts
+++ b/packages/validator/src/util/params.ts
@@ -1,12 +1,18 @@
-import {IChainConfig, chainConfigToJson} from "@chainsafe/lodestar-config";
+import {IChainConfig, chainConfigToJson, defaultChainConfig} from "@chainsafe/lodestar-config";
 
 export class NotEqualParamsError extends Error {}
 
 /**
  * Assert localConfig values match externalSpecJson. externalSpecJson may contain more values than localConfig.
  */
-export function assertEqualParams(localConfig: IChainConfig, remoteConfigJson: Record<string, string>): void {
+export function assertEqualParams(localConfig: IChainConfig, externalSpecJson: Record<string, string>): void {
   const localConfigJson = chainConfigToJson(localConfig) as Record<string, unknown>;
+
+  const remoteConfigJson = {
+    // fill missing properties in remote config with spec default values
+    ...chainConfigToJson(defaultChainConfig),
+    ...externalSpecJson,
+  };
 
   const errors: string[] = [];
 
--- a/packages/validator/src/util/params.ts
+++ b/packages/validator/src/util/params.ts
@@ -18,11 +18,11 @@ export function assertEqualParams(localConfig: IChainConfig, externalSpecJson: R
 
   // Ensure only that the localConfig values match the remote spec
   for (const key of Object.keys(localConfigJson)) {
-    if (
-      (localConfigJson[key] as string)?.toLocaleLowerCase() !==
-      (externalSpecJsonWithDefaults[key] as string)?.toLocaleLowerCase()
-    )
+    const localValue = String(localConfigJson[key]).toLocaleLowerCase();
+    const remoteValue = String(externalSpecJsonWithDefaults[key]).toLocaleLowerCase();
+    if (localValue !== remoteValue) {
       errors.push(`${key} different value: ${localConfigJson[key]} != ${externalSpecJsonWithDefaults[key]}`);
+    }
   }
 
   if (errors.length > 0) {
--- a/packages/validator/src/util/params.ts
+++ b/packages/validator/src/util/params.ts
@@ -8,7 +8,7 @@ export class NotEqualParamsError extends Error {}
 export function assertEqualParams(localConfig: IChainConfig, externalSpecJson: Record<string, string>): void {
   const localConfigJson = chainConfigToJson(localConfig) as Record<string, unknown>;
 
-  const remoteConfigJson = {
+  const externalSpecJsonWithDefaults = {
     // fill missing properties in remote config with spec default values
     ...chainConfigToJson(defaultChainConfig),
     ...externalSpecJson,
@@ -19,9 +19,10 @@ export function assertEqualParams(localConfig: IChainConfig, externalSpecJson: R
   // Ensure only that the localConfig values match the remote spec
   for (const key of Object.keys(localConfigJson)) {
     if (
-      (localConfigJson[key] as string)?.toLocaleLowerCase() !== (remoteConfigJson[key] as string)?.toLocaleLowerCase()
+      (localConfigJson[key] as string)?.toLocaleLowerCase() !==
+      (externalSpecJsonWithDefaults[key] as string)?.toLocaleLowerCase()
     )
-      errors.push(`${key} different value: ${localConfigJson[key]} != ${remoteConfigJson[key]}`);
+      errors.push(`${key} different value: ${localConfigJson[key]} != ${externalSpecJsonWithDefaults[key]}`);
   }
 
   if (errors.length > 0) {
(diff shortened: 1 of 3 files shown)


packages/validator/src/util/params.ts
```diff
@@ -1,20 +1,27 @@
...
 
 /**
  * Assert localConfig values match externalSpecJson. externalSpecJson may contain more values than localConfig.
  */
 export function assertEqualParams(localConfig: IChainConfig, externalSpecJson: Record<string, string>): void {
-  const params1Json = chainConfigToJson(localConfig) as Record<string, unknown>;
-  const params2Json = externalSpecJson;
+  const localConfigJson = chainConfigToJson(localConfig) as Record<string, unknown>;
+
+  const remoteConfigJson = {
```
dapplion: Better name is `externalSpecJsonWithDefaults`

packages/validator/src/util/params.ts
```diff
@@ -1,20 +1,27 @@
...
 
   const errors: string[] = [];
 
   // Ensure only that the localConfig values match the remote spec
-  for (const key of Object.keys(params1Json)) {
-    if (params1Json[key] !== params2Json[key])
-      errors.push(`${key} different value: ${params1Json[key]} != ${params2Json[key]}`);
+  for (const key of Object.keys(localConfigJson)) {
+    if (
+      (localConfigJson[key] as string)?.toLocaleLowerCase() !== (remoteConfigJson[key] as string)?.toLocaleLowerCase()
```
dapplion: Maybe better
```ts
const localValue = String(localConfigJson[key]).toLocaleLowerCase()
const remoteValue = String(remoteConfigJson[key]).toLocaleLowerCase()
       if (localValue !== remoteValue) {
         errors.push(`${key} different value: ${localValue} != ${remoteValue}`);
       }
```

dapplion: Sorry forgot to substitute `remoteValue` in the log

dadepo: It crossed my mind to also substitute but I thought it would be better to log the values as is, since this is the value returned from the config api and what the default values will be.

dapplion: But since it's not what's being used to compare the error will be very confusing if String() coercion does something stupid

dapplion:
So while this PR will fix some issues, the spec still has some params that may change at any time and are not relevant. Like SHARDING_* fork params or any future fork before it's activated. Probably we should whitelist params and test only the active fork params

g11tech:
> So while this PR will fix some issues, the spec still has some params that may change at any time and are not relevant. Like SHARDING_* fork params or any future fork before it's activated. Probably we should whitelist params and test only the active fork params

My suggestion was to divide the config variables hardwork wise and check for the hardwork epoch if not present set it to infinity value and fill it's other corresponding  params with the local defaults.

packages/validator/src/util/params.ts
dadepo: updated as suggested

dadepo:
> So while this PR will fix some issues, the spec still has some params that may change at any time and are not relevant. Like SHARDING_* fork params or any future fork before it's activated. Probably we should whitelist params and test only the active fork params

Won't that lead to more maintenance? where we need to update the whitelist once a property becomes an active fork params? Right now the checks is less error prone and will continue to work whenever there are new active fork params.

dadepo:
> My suggestion was to divide the config variables hardwork wise and check for the hardwork epoch if not present set it to infinity value and fill it's other corresponding params with the local defaults.

Indeed, but I decided not to divide the config variables hardwork wise, because they differ from the way the presets are. The presets are really different per hard forks and they are separated into different files even according to the spec https://github.com/ethereum/consensus-specs/tree/dev/presets/mainnet

But the specs do not separate hard fork related configs. It is just listed in the same file https://github.com/ethereum/consensus-specs/blob/8a4ea1d13db6a72ab2f79eaf5f7af75e1b64428d/configs/mainnet.yaml#L35 Also the fork related params are just 4 different pairs of values

g11tech:
> > My suggestion was to divide the config variables hardwork wise and check for the hardwork epoch if not present set it to infinity value and fill it's other corresponding params with the local defaults.
> 
> Indeed, but I decided not to divide the config variables hardwork wise, because they differ from the way the presets are. The presets are really different per hard forks and they are separated into different files even according to the spec https://github.com/ethereum/consensus-specs/tree/dev/presets/mainnet
> 
> But the specs do not separate hard fork related configs. It is just listed in the same file https://github.com/ethereum/consensus-specs/blob/8a4ea1d13db6a72ab2f79eaf5f7af75e1b64428d/configs/mainnet.yaml#L35 Also the fork related params are just 4 different pairs of values

hmmm yea i understand what you are saying :slightly_smiling_face:, just wanted to put out what I think would have been a good meaningful approach as well. (Though we wouldn't need to put them in files, but divi up the config field names up in separate with an indication of name of the hardfork epoch field) and then it would just be a matter of running a loop and filling up)

dadepo:
> > > My suggestion was to divide the config variables hardwork wise and check for the hardwork epoch if not present set it to infinity value and fill it's other corresponding params with the local defaults.
> > 
> > 
> > Indeed, but I decided not to divide the config variables hardwork wise, because they differ from the way the presets are. The presets are really different per hard forks and they are separated into different files even according to the spec https://github.com/ethereum/consensus-specs/tree/dev/presets/mainnet
> > But the specs do not separate hard fork related configs. It is just listed in the same file https://github.com/ethereum/consensus-specs/blob/8a4ea1d13db6a72ab2f79eaf5f7af75e1b64428d/configs/mainnet.yaml#L35 Also the fork related params are just 4 different pairs of values
> 
> hmmm yea i understand what you are saying 🙂, just wanted to put out what I think would have been a good meaningful approach as well. (Though we wouldn't need to put them in files, but divi up the config field names up in separate with an indication of name of the hardfork epoch field) and then it would just be a matter of running a loop and filling up)

oh sure :) wanted to add to the discussion the reason why I decided not to separate the configs out in files (as we have the presets)

dadepo:
> So while this PR will fix some issues, the spec still has some params that may change at any time and are not relevant. Like SHARDING_* fork params or any future fork before it's activated. Probably we should whitelist params and test only the active fork params

@dapplion can this PR be merged since it fixes the issues the obol guys are having 🙏. It also does not make the current situation worse and even makes the comparison between local/remote configs less error prone:

- In the case the user does not set any of the inactive params for example `SHARDING_*` we are currently filling up the local config with default values as specified in the spec. This PR does the same for remote config to make sure the comparison we do, does not break for unnecessary reason.
- In the case the user sets values for any of the inactive params (for whatever reason), the remote config will contain the default value and the validator will report the inconsistency on start up.
- When the inactive params become active, then the values set in local and remote can be taken into consideration when comparing.

if there is a room to improve how our configs are managed in general, can we do that as a separate issue?

dapplion (approved):
@dadepo Cool agree, let's merge this now but we need a better solution.

In the future forks params there's a high chance of missmatch in the future fairly frequently. While the solution I propose i more maintenance, we must ensure that there are no false positives which kill usability of our software.

PR #3975, opened 2022-05-05
Create codeql-analysis.yml

dapplion:
**Motivation**

Recommended by Github for code analysis

### DIFF
--- a/packages/cli/src/cmds/account/cmds/validator/import.ts
+++ b/packages/cli/src/cmds/account/cmds/validator/import.ts
@@ -171,7 +171,7 @@ async function getKeystorePassphrase(
       console.log("Imported passphrase successfully");
       return passphrase;
     } catch (e) {
-      console.log("Imported passphrase, but it's invalid");
+      console.log(`Imported passphrase, but it's invalid: ${(e as Error).message}`);
     }
   }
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/cmds/account/cmds/validator/import.ts
```diff
@@ -168,10 +168,10 @@ async function getKeystorePassphrase(
     const passphrase = fs.readFileSync(passphraseFile, "utf8");
     try {
       await keystore.decrypt(stripOffNewlines(passphrase));
-      console.log(`Imported passphrase ${passphraseFile}`);
+      console.log("Imported passphrase successfully");
       return passphrase;
     } catch (e) {
-      console.log(`Imported passphrase ${passphraseFile}, but it's invalid: ${(e as Error).message}`);
```
dapplion, line 174: Logging the error is very important, does codeql complain about that too?

dadepo: I'll check if the error message does not contain any information regarding the file path. Wanted to be on the safe side by removing it, but I'll confirm and if its safe and include.

dadepo: Updated to Include the error message

wemeetagain (approved):

PR #3978, opened 2022-05-05
ESM Support

wemeetagain:
Use ESM (only) across all lodestar packages.

- [x] Fix linter errors
- [x] Fix `check-types` errors
- [x] Fix merge sim tests
- [x] Fix sim tests
- [x] Fix e2e tests
- [x] Fix spec tests
- [x] Fix typescript-docs-verifier - disabled for now
- [x] Fix merge conflicts

PR #3979, opened 2022-05-05
Initialize libp2p with esm module

wemeetagain:
**Motivation**

Experiment to see what a minimal code change to support ESM libp2p would look like.

**Description**

Turn the bit of code that initialized a Libp2p instance into an ES Module. Use `await import` to import this.
Typescript tries to transpile `await import` into a `require` so a hack is required to force its hand.

### DIFF
--- a/packages/lodestar/src/network/nodejs/util.ts
+++ b/packages/lodestar/src/network/nodejs/util.ts
@@ -3,8 +3,7 @@
  */
 
 import PeerId from "peer-id";
-import LibP2p from "libp2p";
-import {NodejsNode} from "./bundle";
+import type LibP2p from "libp2p";
 import {defaultDiscv5Options, defaultNetworkOptions, INetworkOptions} from "../options";
 import {isLocalMultiAddr, clearMultiaddrUDP} from "..";
 import {ENR} from "@chainsafe/discv5";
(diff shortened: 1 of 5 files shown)


packages/lodestar/src/network/nodejs/util.ts
```diff
@@ -66,6 +65,12 @@ export async function createNodeJsLibp2p(
     }
   }
 
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const {NodejsNode} = await Function("return import('./bundle.mjs')")();
```
dapplion, line 71: Write a nice long comment explaining this hahaha

packages/lodestar/test/utils/network.ts
```diff
@@ -3,12 +3,17 @@ import {Multiaddr} from "multiaddr";
...
 import {Libp2pEvent} from "../../src/constants";
+import type Libp2p from "libp2p";
 
-export async function createNode(multiaddr: string, inPeerId?: PeerId): Promise<NodejsNode> {
+export async function createNode(multiaddr: string, inPeerId?: PeerId): Promise<Libp2p> {
   const peerId = inPeerId || (await createPeerId());
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const {NodejsNode} = await Function("return import('../../src/network/nodejs/bundle.mjs')")();
```
dapplion, line 15: Samee, long nice comment explaining this. Also, doesn't this require a type cast for safety?

packages/lodestar/src/network/nodejs/util.ts
dapplion: maybe you can import the type with `import type` just to make Typescript happy and not have everything as any

wemeetagain: lmao the most malicious looking code ever. Hopefully we finish #3978 instead of this.

wemeetagain:
Closing in favor of #3978

PR #3984, opened 2022-05-07
Fix the terminal validations of the merge block

### LINKED ISSUE #3982: Lodestar BN never validates the `assertValidTerminalPowBlock` checks in forkchoice
Context: While doing the PR for PoS banner transition, stumbled upon an issue:

In the forkchoice we are supposed to `assertValidTerminalPowBlock` in case the block is a merge transition block. This happens via
```typescript
if (
      preCachedData?.isMergeTransitionBlock ||
      (bellatrix.isBellatrixStateType(state) &&
        bellatrix.isBellatrixBlockBodyType(block.body) &&
        bellatrix.isMergeTransitionBlock(state, block.body))
    )
      assertValidTerminalPowBlock(this.config, (block as unknown) as bellatrix.BeaconBlock, preCachedData);
```

But the issue here is that `isMergeTransitionBlock` flag being passed never comes to be true even in case of merge transition block because it is evaluated on postState in `importBlock` from where the `forkChoice.onBlock` is being called.
Furthermore the second part of the condition also evaluates to `false` as the state here is also postState as well (again passed from `importBlock`

Suggested Fix:
Lighthose runs these validations in the block verification module where we have access to preState to proposerly evaluate `isMergeTransitionBlock`. So the assertions should move over there.

Note: Check if there are any spec tests which directly pass an invalid terminal block to forkchoice for validation. If so discuss and implement appropriate resolution.

### LINKED ISSUE #3986: Modify spec runner's `fork_choice.test.ts` to include testcases for `on_merge_block` scenario
Seems like the tests for merge block transitions have been included in the specs, but our `fork_choice.test.ts` doesn't yet address them.

Modify and make the spec runner run these tests successfully.

g11tech:
**Motivation**
It was discovered that terminal block validations weren't running on the merge block because of incorrect evaluation of the merge transition block on postState 

This PR fixes the evaluation on preState as well as move the assertions into the verifyBlock instead of forkChoice. The `assertValidTerminalPowBlock` still stays in forkchoice because it might be required to validate terminal conditions if it imported a merge transition block in syncing status, which later becomes valid

Closes #3982, Closes #3986

### DIFF
--- a/packages/fork-choice/src/forkChoice/forkChoice.ts
+++ b/packages/fork-choice/src/forkChoice/forkChoice.ts
@@ -1016,7 +1016,7 @@ export function assertValidTerminalPowBlock(
     if (computeEpochAtSlot(block.slot) < config.TERMINAL_BLOCK_HASH_ACTIVATION_EPOCH)
       throw Error(`Terminal block activation epoch ${config.TERMINAL_BLOCK_HASH_ACTIVATION_EPOCH} not reached`);
 
-    // powBock.blockhash is hex, so we just pick the corresponding root
+    // powBock.blockHash is hex, so we just pick the corresponding root
     if (!ssz.Root.equals(block.body.executionPayload.parentHash, config.TERMINAL_BLOCK_HASH))
       throw new Error(
         `Invalid terminal block hash, expected: ${toHexString(config.TERMINAL_BLOCK_HASH)}, actual: ${toHexString(
@@ -1032,7 +1032,7 @@ export function assertValidTerminalPowBlock(
 
     const {powBlock, powBlockParent} = preCachedData;
     if (!powBlock) throw Error("onBlock preCachedData must include powBlock");
-    if (!powBlockParent) throw Error("onBlock preCachedData must include powBlock");
+    if (!powBlockParent) throw Error("onBlock preCachedData must include powBlockParent");
 
     const isTotalDifficultyReached = powBlock.totalDifficulty >= config.TERMINAL_TOTAL_DIFFICULTY;
     const isParentTotalDifficultyValid = powBlockParent.totalDifficulty < config.TERMINAL_TOTAL_DIFFICULTY;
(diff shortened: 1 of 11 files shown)


twoeths:
@g11tech did you find out any spec test covering the `is_valid_terminal_pow_block()` function? Otherwise I think we should ask for a spec test for it

g11tech:
> 

I can see some tests that could be pertaining to them. I can create an issue and pursue it as a separate PR :thinking: 

Update: https://github.com/ChainSafe/lodestar/issues/3986

g11tech:
> @g11tech did you find out any spec test covering the `is_valid_terminal_pow_block()` function? Otherwise I think we should ask for a spec test for it

added

packages/lodestar/src/chain/blocks/verifyBlock.ts
```diff
@@ -272,6 +279,15 @@ export async function verifyBlockStateTransition(
           errorMessage: execResult.validationError,
         });
     }
+
+    if (isMergeTransitionBlock) {
```
dapplion, line 296: Add nice long comment here like in Lighthouse explaining why we do this here and not in the forkchoice. Link also to the spec's location where this code should be.

Since this is not documenation for JSDocs use style
```ts
//
//
```

packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -328,14 +328,6 @@ export class ForkChoice implements IForkChoice {
...
     }
 
-    if (
-      preCachedData?.isMergeTransitionBlock ||
-      (bellatrix.isBellatrixStateType(state) &&
-        bellatrix.isBellatrixBlockBodyType(block.body) &&
-        bellatrix.isMergeTransitionBlock(state, block.body))
-    )
-      assertValidTerminalPowBlock(this.config, (block as unknown) as bellatrix.BeaconBlock, preCachedData);
-
```
dapplion, line 338: Add nice long comment here explaining why we do NOT do this here and where it's happening instead. Link also to the spec's code.

Since this is not documenation for JSDocs use style
```ts
//
//
```

dapplion (approved):

PR #3989, opened 2022-05-08
Improve sync speed

dapplion:
**Motivation**

Lodestar doesn't have great sync speed compared to other clients. I've been analyzing why and what can we do better.

Verifying blocks requires 3 separate tasks:
- Verify signatures:
  - Collect signatures: 5 ms (index attestations)
  - Aggregate pubkeys: 45 ms
  - Verify signatures: 100-150 ms (in workers)
- Run state transition:
  - processBlock: 5 ms pure logic, 5 ms commitViewDU
  - hashTreeRoot: 22 ms
- Execution payload: (not benchmarked)

Same info but in a table

| Op                 | ms per block | ms per block x32 |
| ------------------ | ------------ | ---------------- |
| processBlock       | 32           | 1024             |
| getBlockSignatures | 45           | 1440             |
| verifySigs (102 s) | 100-200      | 3200-6400        |

As you can see the bulk of the work is for signature, which currently only the verifySig part is done in workers, getBlockSignatures happens in the main thread.

Current master processes blocks in series which submits 100 sigs per blocks to the workers. Since that's not a lot all signatures are submitted as a single job to worker id 0. So there's no use of parallelization.

**Description**

This PR attempts to parallelize better the tasks above.

- Submit all signatures at once to ensure full utilization of workers
- Ensure signatures and execution payloads are submitted first before state transition "blocks" main thread

As very rough estimate this PR reduces the time to process 32 blocks from 8s to 4s. From benchmark data below

Benchmark suite | Current: d7a22cb | Previous: d7acdcf | Ratio
-- | -- | -- | --
altair verifyImport mainnet_s3766816:31 | 3.6543  s/op | 6.6473  s/op | 0.55

Here are some CPU profiles that show this data plotted 

**1: Full block segment profile**

Signatures are collected first (blue wide sections) with some state transitions runs in between (narrow tall pink sections). Then the rest of block's state transitions are run. Then main thread is idle waiting for workers to finish verifying signatures.

Screenshot from 2022-05-08 22-33-48

**2: state transition run detail**

hashTreeRoot dominates

Screenshot from 2022-05-08 22-34-18

**TODO**
- [x] Do preliminary PR with perf test on master for reference https://github.com/ChainSafe/lodestar/pull/3995
- [ ] Benchmark with execution payload HTTP roundtrip time (just simulate)
- [ ] Long term: explore how to aggregate pubkeys in workers

### DIFF
--- a/packages/beacon-node/src/chain/blocks/verifyBlock.ts
+++ b/packages/beacon-node/src/chain/blocks/verifyBlock.ts
@@ -1,15 +1,7 @@
-import {
-  CachedBeaconStateAllForks,
-  isBellatrixStateType,
-  isBellatrixBlockBodyType,
-  isMergeTransitionBlock as isMergeTransitionBlockFn,
-  isExecutionEnabled,
-  getBlockSignatureSets,
-  stateTransition,
-} from "@lodestar/state-transition";
+import {CachedBeaconStateAllForks, computeEpochAtSlot} from "@lodestar/state-transition";
 import {allForks, bellatrix} from "@lodestar/types";
 import {toHexString} from "@chainsafe/ssz";
-import {IForkChoice, ExecutionStatus, assertValidTerminalPowBlock} from "@lodestar/fork-choice";
+import {IForkChoice, ExecutionStatus} from "@lodestar/fork-choice";
 import {IChainForkConfig} from "@lodestar/config";
 import {ILogger} from "@lodestar/utils";
 import {IMetrics} from "../../metrics/index.js";
@@ -19,11 +11,12 @@ import {IBeaconClock} from "../clock/index.js";
 import {BlockProcessOpts} from "../options.js";
 import {IStateRegenerator, RegenCaller} from "../regen/index.js";
 import {IBlsVerifier} from "../bls/index.js";
-import {ExecutePayloadStatus} from "../../execution/engine/interface.js";
-import {byteArrayEquals} from "../../util/bytes.js";
 import {IEth1ForBlockProduction} from "../../eth1/index.js";
 import {ImportBlockOpts} from "./types.js";
 import {POS_PANDA_MERGE_TRANSITION_BANNER} from "./utils/pandaMergeTransitionBanner.js";
+import {verifyBlocksStateTransitionOnly} from "./verifyBlocksStateTransitionOnly.js";
+import {verifyBlocksSignatures} from "./verifyBlocksSignatures.js";
+import {verifyBlocksExecutionPayload} from "./verifyBlocksExecutionPayloads.js";
 
 export type VerifyBlockModules = {
   bls: IBlsVerifier;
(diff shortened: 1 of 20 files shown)


dapplion:
- Blocked by https://github.com/ChainSafe/lodestar/issues/4017

wemeetagain:
failing tests and merge conflicts

wemeetagain:
I know this is lower priority, but it would be nice to not let this rot

dapplion:
- [ ] TODO: Ensure that the new verify+import benchmark runs last and / or ensure that all resources are cleaned-up to prevent messing with other benchmarks

dapplion:
Still needs a lot of work to be review-able, putting as draft till then

packages/beacon-node/src/chain/regen/queued.ts
```diff
@@ -69,8 +69,15 @@ export class QueuedStateRegenerator implements IStateRegenerator {
     // Check the checkpoint cache (if the pre-state is a checkpoint state)
     if (parentEpoch < blockEpoch) {
       const checkpointState = this.checkpointStateCache.getLatest(parentRoot, blockEpoch);
-      if (checkpointState) {
+      if (checkpointState && computeEpochAtSlot(checkpointState.slot) === blockEpoch) {
+        // TODO: Miss-use of checkpointStateCache here
```
wemeetagain, line 73: What does this comment mean?

packages/beacon-node/src/chain/blocks/verifyBlocksStateTransitionOnly.ts
```diff
@@ -0,0 +1,79 @@
...
+      throw new BlockError(block, {
+        code: BlockErrorCode.INVALID_STATE_ROOT,
+        root: postState.hashTreeRoot(),
+        expectedRoot: block.message.stateRoot,
+        preState,
+        postState,
+      });
+    }
+
+    postStates[i] = postState;
```
wemeetagain, line 60: Do we need to keep all states?

packages/beacon-node/src/chain/regen/queued.ts
dapplion: It doesn't conform to the method description: "state dialed to block.slot"

packages/beacon-node/src/chain/blocks/verifyBlocksStateTransitionOnly.ts
dapplion: Yes, current importBlock flow needs the state of every block. To only require the last we would need to refactor importBlock

dapplion:
I've restarted two contabos to sync Prater from genesis:

**With unstable**

Heavy use of a single worker. Does not compete for resources so the time for sig is lower, doing 480 sigs / sec

Screenshot from 2022-07-18 18-38-18

Screenshot from 2022-07-18 18-45-07



**With this branch**

Node is using the four cores with all the workers, cost per sig is higher, but can do 700 sigs / sec

Screenshot from 2022-07-18 18-36-12

Screenshot from 2022-07-18 18-46-42

packages/beacon-node/src/chain/blocks/verifyBlock.ts
```diff
@@ -38,249 +31,79 @@ export type VerifyBlockModules = {
...
-      //    On kintsugi devnet, this has been observed to cause contiguous proposal failures
-      //    as the network is geth dominated, till a non geth node proposes and moves network
-      //    forward
-      // For network/unreachable errors, an optimization can be added to replay these blocks
-      // back. But for now, lets assume other mechanisms like unknown parent block of a future
-      // child block will cause it to replay
+  try {
+    const [{postStates, proposerBalanceDeltas}, , {executionStatuses, mergeBlockFound}] = await Promise.all([
+      // Run state transition only
+      // TODO: Ensure it yields to allow flushing to workers and engine API
```
twoeths, line 84: remove this `TODO`?

packages/beacon-node/src/chain/blocks/verifyBlock.ts
```diff
@@ -38,249 +31,79 @@ export type VerifyBlockModules = {
...
-          return null;
-        }));
+      // Execution payloads
+      verifyBlocksExecutionPayload(chain, blocks, preState0, abortController.signal, opts),
+    ]);
 
-      assertValidTerminalPowBlock(chain.config, mergeBlock, {executionStatus, powBlock, powBlockParent});
+    if (mergeBlockFound !== null) {
+      // merge block found and is fully valid = state transition + signatures + execution payload.
+      // TODO: Will this banner be logged during syncing?
```
twoeths, line 96: it'd be nice to log the merge block so remove this `TODO` too?

packages/beacon-node/src/chain/blocks/verifyBlocksSignatures.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+type AllValidRes = {allValid: true} | {allValid: false; index: number};
+
+/**
+ * From an array of promises that resolve a boolean isValid
+ * - if all valid, await all and return
+ * - if one invalid, abort immediately and return index of invalid
+ */
+export function rejectFirstInvalidResolveAllValid(isValidPromises: Promise<boolean>[]): Promise<AllValidRes> {
```
twoeths, line 60: could you move this to utils module or any utils file? this could be reused

dapplion: The counter argument is that if we stop using this function it will stay in utils forever un-used. It's very specific to this code, if it gets re-used in the future then we can move to utils

packages/beacon-node/src/chain/blocks/verifyBlock.ts
dapplion: I don't think it should be logged during sync, tho would prefer to debate in a separate issue. This TODO was more to note that this can happen

packages/beacon-node/src/chain/blocks/verifyBlock.ts
dapplion: This goal is not accomplished, but it's just a hope. I haven't found a way to ensure from JS land that data has been sent to the worker which is the ultimate issue that cap throughput.

twoeths (approved):
this looks great 👍

PR #3990, opened 2022-05-09
Prepare custom version on next release

dapplion:
**Motivation**

We used `lerna publish --canary` option. However, since we now publish must version on branches, i.e. v0.35.x branch, lerna fails to detect the latest version and publishes canary versions as `0.34.0-dev.173+28e2c74cf0` instead of `0.36.0-dev.4+28e2c74cf0`, which creates confusion.

**Description**

- Prepare custom version on next release

### DIFF
--- a/.github/workflows/release-nightly.yml
+++ b/.github/workflows/release-nightly.yml
@@ -41,10 +41,20 @@ jobs:
         if: steps.cache-deps.outputs.cache-hit == 'true'
       # </common-build>
 
+      - name: Get version
+        id: version
+        run: |
+          PACKAGE_VERSION=$(node -p "require('./packages/lodestar/package.json').version")
+          export VERSION=${PACKAGE_VERSION}-dev.${GITHUB_SHA:0:10}
+          echo "::set-output name=version::$VERSION"
+          echo PACKAGE_VERSION $PACKAGE_VERSION GITHUB_SHA $GITHUB_SHA VERSION $VERSION
+      - name: Set version
+        run: node_modules/.bin/lerna version ${{ steps.version.outputs.version }} --no-git-tag-version --force-publish --yes
       - name: Publish to npm registry
-        # Just use lerna publish with --canary option. Using 'from-package' ignore other options
-        # and only compares against the verison in package.json, and skips release if already
-        # published.
+        # Note: before https://github.com/ChainSafe/lodestar/commit/28e2c74cf0f1bede8b09c8c9fec26f54b367e3fd
+        # We used `lerna publish --canary` option. However, since we now publish must version on branches,
+        # i.e. v0.35.x branch, lerna fails to detect the latest version and publishes canary versions as
+        # `0.34.0-dev.173+28e2c74cf0` instead of `0.36.0-dev.4+28e2c74cf0`, which creates confusion.
         #
         # --no-git-reset:
         #   Do not delete code version artifacts so the next step can pick the version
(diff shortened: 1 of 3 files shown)


.github/workflows/release-nightly.yml
```diff
@@ -41,10 +41,20 @@ jobs:
         if: steps.cache-deps.outputs.cache-hit == 'true'
       # </common-build>
 
+      - name: Get version
+        id: version
+        run: |
+          PACKAGE_VERSION=$(node -p "require('./packages/lodestar/package.json').version")
```
twoeths, line 46: looks like this command returns incorrect value as in https://github.com/ChainSafe/lodestar/runs/6338092886?check_suite_focus=true

```
Run VERSION=$(node -p "require('./packages/lodestar/package.json').version")
VERSION 0.34.0-dev.173+28e2c74cf0
```

so I'm afraid this time it'd get the wrong value too, can we test the output of this step in this branch?

dapplion: Should be fine because we run this now before lerna canary

g11tech (dismissed):
:crossed_fingers:

dapplion:
Publishing worked fine, see GA logs https://github.com/ChainSafe/lodestar/runs/6367861214?check_suite_focus=true#step:10:18

```
Found 14 packages to publish:
lerna info publish Publishing packages to npm...
 - @chainsafe/lodestar-api => 0.36.0-dev.3e6ce7b69e
 - @chainsafe/lodestar-beacon-state-transition => 0.36.0-dev.3e6ce7b69e
 - @chainsafe/lodestar-cli => 0.36.0-dev.3e6ce7b69e
 - @chainsafe/lodestar-config => 0.36.0-dev.3e6ce7b69e
 - @chainsafe/lodestar-db => 0.36.0-dev.3e6ce7b69e
 - @chainsafe/lodestar-fork-choice => 0.36.0-dev.3e6ce7b69e
 - @chainsafe/lodestar-keymanager-server => 0.36.0-dev.3e6ce7b69e
 - @chainsafe/lodestar-light-client => 0.36.0-dev.3e6ce7b69e
 - @chainsafe/lodestar => 0.36.0-dev.3e6ce7b69e
 - @chainsafe/lodestar-params => 0.36.0-dev.3e6ce7b69e
 - @chainsafe/lodestar-spec-test-util => 0.36.0-dev.3e6ce7b69e
 - @chainsafe/lodestar-types => 0.36.0-dev.3e6ce7b69e
 - @chainsafe/lodestar-utils => 0.36.0-dev.3e6ce7b69e
 - @chainsafe/lodestar-validator => 0.36.0-dev.3e6ce7b69e
```

Proof in NPM https://www.npmjs.com/package/@chainsafe/lodestar-cli/v/0.36.0-dev.3e6ce7b69e

.github/workflows/release-nightly.yml
mpetrunic: Why not take version from https://github.com/ChainSafe/lodestar/blob/master/lerna.json#L7 ?

.github/workflows/release-nightly.yml
```diff
@@ -71,17 +87,15 @@ jobs:
...
-          node_modules/.bin/lerna publish --yes --no-verify-access \
-          --canary --dist-tag next --no-git-reset --force-publish \
-          --preid dev --exact
+          node_modules/.bin/lerna publish from-package \
+          --yes \
+          --no-verify-access \
+          --dist-tag next \
+          --no-git-reset \
+          --force-publish \
+          --exact
```
mpetrunic, line 96: no more `--preid dev` ?

mpetrunic (commented):
LGTM, just left two comments to double check stuff

.github/workflows/release-nightly.yml
dapplion: `preid` flag is not necessary because the version is given, not computed by lerna in the canary logic. The `dev` word in used above when manually crafting the version string

dapplion: Can't comment above for some reason

> Why not take version from https://github.com/ChainSafe/lodestar/blob/master/lerna.json#L7 ?

Should be equivalent indeed, I don't have any preference in grabbing it from there, but the package.json seems more generic long term if we ever switch away from lerna

wemeetagain (approved):

g11tech (approved):

PR #3992, opened 2022-05-09
Simplify gitData and version guessing

dapplion:
**Background**

When publishing content to NPM the "git data" i.e. the exact commit at which the build was done is not possible to be known. So a `.git-data.json` file is created to persist this useful data:

 - For NPM release (stable): Only the version is persisted. Once must then track the version's tag in Github to resolve that version to a specific commit. While this is okay, git-data.json gives a gurantee of the exact commit at build time.

 - For NPM release (nightly): canary commits include the commit, so this feature is not really  necessary. However, it's more cumbersome to have conditional logic on stable / nightly.

 - For build from source: .git folder is available in the context of the built code, so it can extract  branch and commit directly without the need for .git-data.json.

 - For build from source dockerized: This feature is required to know the branch and commit, since git data is not persisted past the build. However, .dockerignore prevents .git folder from being  copied into the container's context, so .git-data.json can't be generated.

**Before**

The "version string" (what is displayed in logs and metrics) was manually crafted on each run from new git-data, persisted git-data and local package.json file. The priority order was:
- new git-data
- persisted git-data
- local package.json file

**After**

The priority order is changed such that the semver version is ALWAYS derived from the local package.json file and never from any git data. This will solve the incorrect tagging of versions.

Also, the number of commits is never tracked, so for nightly only the commit will be shown.

### DIFF
--- a/docker/from_source.Dockerfile
+++ b/docker/from_source.Dockerfile
@@ -30,6 +30,12 @@ RUN yarn install --non-interactive --frozen-lockfile --ignore-scripts
 COPY . .
 RUN yarn install --non-interactive --frozen-lockfile && yarn build
 
+# To have access to the specific branch and commit used to build this source,
+# a git-data.json file is created by persisting git data at build time. Then,
+# a version string like `v0.35.0-beta.0/HEAD/82219149 (git)` can be shown in
+# the terminal and in the logs; which is very useful to track tests better.
+RUN cd packages/cli && yarn write-git-data
+
 # Copy built src + node_modules to a new layer to prune unnecessary fs
 # Previous layer weights 7.25GB, while this final 488MB (as of Oct 2020)
 FROM node:16-alpine
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -3,9 +3,9 @@ import yargs from "yargs";
 import {cmds} from "./cmds";
 import {globalOptions, rcConfigOption} from "./options";
 import {registerCommandToYargs} from "./util";
-import {getVersion} from "./util/version";
+import {getVersionData} from "./util/version";
 
-const version = getVersion();
+const {version} = getVersionData();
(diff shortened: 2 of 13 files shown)


wemeetagain:
> When publishing content to NPM the "git data" i.e. the exact commit at which the build was done is not possible to be known. 

I don't think this is true?

```bash
$ yarn add @chainsafe/lodestar-cli
...
$ cat node_modules/@chainsafe/lodestar-cli/package.json | grep gitHead
  "gitHead": "822191497bcce6b0f45c8def07d24d6a6045f1c9"
```

dapplion:
> > When publishing content to NPM the "git data" i.e. the exact commit at which the build was done is not possible to be known.
> 
> I don't think this is true?
> 
> ```shell
> $ yarn add @chainsafe/lodestar-cli
> ...
> $ cat node_modules/@chainsafe/lodestar-cli/package.json | grep gitHead
>   "gitHead": "822191497bcce6b0f45c8def07d24d6a6045f1c9"
> ```

Oh I wasn't aware of this. We could def re-use this and burn the .git-data.json thing. Tho are we actually using it for our packages? I don't see it doing an install from NPM

dapplion:
> > When publishing content to NPM the "git data" i.e. the exact commit at which the build was done is not possible to be known.
> 
> I don't think this is true?
> 
> ```shell
> $ yarn add @chainsafe/lodestar-cli
> ...
> $ cat node_modules/@chainsafe/lodestar-cli/package.json | grep gitHead
>   "gitHead": "822191497bcce6b0f45c8def07d24d6a6045f1c9"
> ```

While this is a good suggestion it's actually complimentary to this PR. The main goal of the PR is to ensure that the version string is derived from the package.json, which is equivalent to @wemeetagain suggestion. In a latter PR we can effectively "merge" .git-data.json into the package.json using gitHead property, but I don't see it as a blocker.

This PR is very important to ensure we don't render confusing or outright wrong versions in the logs.

g11tech (approved):
LGTM!

Issue #4009, opened 2022-05-11
[PROD] Validator votes for incorrect head


twoeths:
**Describe the bug**

```
May-11 05:36:27.033[]                ^[[34mdebug^[[39m: Signed attestation slot=3785280, index=56, head=0xc4d2ce80c7723b88d996303a7f37ba05050f7c775277818bc689e7e6724ec000, validatorIndex=270127
```

chain head of same slot comes right after that

```
May-11 05:36:27.379[]              ^[[36mverbose^[[39m: Found new chain head slot=3785280, head=0x73e7c00453bb8383a0104fd62bb0af1032c58edfef4fbe44f56dced48871de60, previouDuty=0x952fc43d681ed80ef2f4779a6c8e98a5f7ef2037406548885947b4add17d9d8b, currentDuty=0xc4d2ce80c7723b88d996303a7f37ba05050f7c775277818bc689e7e6724ec000
```

attestation was missed probably because its AttestationData is not the same to others

**Expected behavior**

Beacon node turns out to be very busy at that time, it reached beacon node 4s later
```
May-11 05:36:31.551[VMON]            ^[[34mdebug^[[39m: Local validator published unaggregated attestation validatorIndex=270127, slot=3785280, committeeIndex=56, subnet=56, sentPeers=25, delaySec=2.236999988555908
```

twoeths:
The i/o lag only happened at that time - the below shows 4s delay between signing time and publishing time

```
May-11 05:36:27.033[]                ^[[34mdebug^[[39m: Signed attestation slot=3785280, index=56, head=0xc4d2ce80c7723b88d996303a7f37ba05050f7c775277818bc689e7e6724ec000, validatorIndex=270127
...
May-11 05:36:31.566[]                 ^[[32minfo^[[39m: Published attestations slot=3785280, count=3
```

right now it takes <0.1s to publish attestation

```
May-11 07:33:59.526[]                debug: Signed attestation slot=3785868, index=56, head=0x53634b7587e2404a3c7c573908b749d9bcec5b7320e763f44e5a2e8c2505cf98, validatorIndex=270103
May-11 07:33:59.589[]                 info: Published attestations slot=3785868, count=2
May-11 07:34:59.506[]                debug: Signed attestation slot=3785873, index=3, head=0x7e7b2c388b85fe46e2ea810687182fecfb14f8161f92166dc6a86b36653892bd, validatorIndex=270120
[...]

PR #4010, opened 2022-05-11
Observe attestation submission after the api called

twoeths:
**Motivation**

We want to monitor the delay time from 1/3 slot to the time attestation is published

**Description**

+ as shown in #4009, sometimes beacon node has i/o lag so we'd like to track `attesterStepCallPublishAttestation` metric after the attestation is published

part of #4009

### DIFF
--- a/packages/validator/src/services/attestation.ts
+++ b/packages/validator/src/services/attestation.ts
@@ -163,11 +163,10 @@ export class AttestationService {
     const afterBlockDelayMs = (1000 * this.clock.secondsPerSlot) / (this.opts?.afterBlockDelaySlotFraction ?? 6);
     await sleep(Math.min(msToOneThirdSlot, afterBlockDelayMs));
 
-    this.metrics?.attesterStepCallPublishAttestation.observe(this.clock.secFromSlot(slot + 1 / 3));
-
     // Step 2. Publish all `Attestations` in one go
     try {
       await this.api.beacon.submitPoolAttestations(signedAttestations);
+      this.metrics?.attesterStepCallPublishAttestation.observe(this.clock.secFromSlot(slot + 1 / 3));
       this.logger.info("Published attestations", {slot, count: signedAttestations.length});
       this.metrics?.publishedAttestations.inc(signedAttestations.length);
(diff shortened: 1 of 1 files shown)


wemeetagain (commented):
Do we still want the observation before calling publishing attestations?
Should the name of this metric be changed? eg: attesterStepFinishPublishAttestation or something

packages/validator/src/services/attestation.ts
```diff
@@ -163,11 +163,10 @@ export class AttestationService {
     const afterBlockDelayMs = (1000 * this.clock.secondsPerSlot) / (this.opts?.afterBlockDelaySlotFraction ?? 6);
     await sleep(Math.min(msToOneThirdSlot, afterBlockDelayMs));
 
-    this.metrics?.attesterStepCallPublishAttestation.observe(this.clock.secFromSlot(slot + 1 / 3));
-
     // Step 2. Publish all `Attestations` in one go
     try {
       await this.api.beacon.submitPoolAttestations(signedAttestations);
+      this.metrics?.attesterStepCallPublishAttestation.observe(this.clock.secFromSlot(slot + 1 / 3));
```
dapplion, line 169: Agree with @wemeetagain the metric name clearly states "before call to publish attestation". You can already get the metric you want by summing `attesterStepCallPublishAttestation + vc rest request time {routeId=submitPoolAttestations}`.

If you want to be more granular metrics, those should be added in the beacon chain not here

dapplion (changes requested):

packages/validator/src/services/attestation.ts
dapplion: If your goal is to measure how busy the beacon node is (potentially due to i/o lag), just use the api request time metric

wemeetagain:
@tuyennhv

Are you able to measure what you want using the API request time metric?
Can we close this PR?

twoeths:
> @tuyennhv
> 
> Are you able to measure what you want using the API request time metric? Can we close this PR?

yeah I'd use the API request time metric instead

Issue #3508, opened 2021-12-10
Optimistic Sync: Part 2 - Prune/Validate/Update fork-choice on lastValidHash response


g11tech:
As part 1 of optimistic sync, the execution status was tracked and validator duties were aborted on syncing head. Second part of the optimistic sync is to update fork-choice proto blocks and prune (on invalidation), and update the head

PR #4016, opened 2022-05-12
Handle merge block fetch error

g11tech:
**Motivation**
The merge block fetch (`getBlockByHash`) can error and stop the CL from syncing past merge transition point unless the corresponding EL also syncs upto the merge transition block.

This PR ignores the merge pow block fetch error for preparing data to run `assertValidTerminalPowBlock` validations on `isMergeTransitionBlock`. 
This fix will allow lodestar to keep syncing and importing the blocks even if the respective EL is behind the merge transition point or doesn't has the corresponding merge block (provided the validations pass as per `assertValidTerminalPowBlock` logic)

### DIFF
--- a/packages/lodestar/src/chain/blocks/verifyBlock.ts
+++ b/packages/lodestar/src/chain/blocks/verifyBlock.ts
@@ -296,8 +296,8 @@ export async function verifyBlockStateTransition(
     if (isMergeTransitionBlock) {
       const mergeBlock = block.message as bellatrix.BeaconBlock;
       const powBlockRootHex = toHexString(mergeBlock.body.executionPayload.parentHash);
-      const powBlock = await chain.eth1.getPowBlock(powBlockRootHex);
-      const powBlockParent = powBlock && (await chain.eth1.getPowBlock(powBlock.parentHash));
+      const powBlock = await chain.eth1.getPowBlock(powBlockRootHex).catch((_error) => null);
+      const powBlockParent = powBlock && (await chain.eth1.getPowBlock(powBlock.parentHash).catch((_error) => null));
 
       assertValidTerminalPowBlock(chain.config, mergeBlock, {executionStatus, powBlock, powBlockParent});
     }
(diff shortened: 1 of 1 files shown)


packages/lodestar/src/chain/blocks/verifyBlock.ts
```diff
@@ -296,8 +296,8 @@ export async function verifyBlockStateTransition(
     if (isMergeTransitionBlock) {
       const mergeBlock = block.message as bellatrix.BeaconBlock;
       const powBlockRootHex = toHexString(mergeBlock.body.executionPayload.parentHash);
-      const powBlock = await chain.eth1.getPowBlock(powBlockRootHex);
-      const powBlockParent = powBlock && (await chain.eth1.getPowBlock(powBlock.parentHash));
+      const powBlock = await chain.eth1.getPowBlock(powBlockRootHex).catch((_error) => null);
+      const powBlockParent = powBlock && (await chain.eth1.getPowBlock(powBlock.parentHash).catch((_error) => null));
```
dapplion: This is very serious error we must know about, at least log it

dapplion: If `isMergeTransitionBlock == true` and `powBlock == null` it's impossible for us to accept this block right?

g11tech: no, we can still import it if `executionStatus` is `SYNCING` and we are not  `safeSlotsToImportOptimistically` (128) blocks of the clock slot (i.e. merge has long happened, this is just syncing past merge even if EL doesn't has merge block), which is when this fetch actually throws error.

Another scenario is if `TERMINAL_BLOCK_HASH` is set, then also `powBlock` is not required as hashes are compared directly.

`assertValidTerminalPowBlock` takes care of these checks so we can safely assign it null and let it move to validation.

Logging them is a good idea! will update :+1:

dapplion: Why not move the fn `assertValidTerminalPowBlock` into the verifyBlock.ts, and do the fetching inside the function? Seems like things will be more straight forward to reason about.

g11tech: ummm well there are couple of considerations:
i) forkchoice spec tests have spec tests for terminal conditions, so a *pure* function is better to have
ii) the https://github.com/ChainSafe/lodestar/issues/3508 will require this validation as per spec once the merge block status changes from SYNCING to VALID. 

Lets leave it here for the moment and see if we can place it better when working on the ii) ? :thinking:

dapplion (approved):

PR #4020, opened 2022-05-13
Reject many skipped blocks

### LINKED ISSUE #3988: Reject importing blocks with many skipped slots
**Describe the bug**

Peers can DOS us by triggering expensive epoch transitions in cases of long non-finality. Lighthouse limits skipped slots to 100, we should do the same.

DOS will happen because we need to get the proposerShuffling. To get the shuffling we have to do a bunch of epochTransitions, the longer the distance between the parent and block, the more we have to do. epochTransitions are expensive ~750ms, so we must limit how many a single bad block can trigger.

Add this to gossip block validation

```ts
  if (blockSlot > parentBlock.slot + chain.importMaxSkipSlots) {
    throw new BlockGossipError(GossipAction.IGNORE, {
      code: BlockErrorCode.TOO_MANY_SKIPPED_SLOTS,
      parentSlot: parentBlock.slot,
      blockSlot,
    });
  }
```

dadepo:
**Motivation**

**Description**

Closes #3988

### DIFF
--- a/packages/cli/src/options/beaconNodeOptions/chain.ts
+++ b/packages/cli/src/options/beaconNodeOptions/chain.ts
@@ -6,12 +6,14 @@ export interface IChainArgs {
   "chain.blsVerifyAllMainThread": boolean;
   "chain.disableBlsBatchVerify": boolean;
   "chain.persistInvalidSszObjects": boolean;
+  // No need to define chain.persistInvalidSszObjects as part of IChainArgs
+  // as this is defined as part of IBeaconPaths
+  // "chain.persistInvalidSszObjectsDir": string;
   "chain.proposerBoostEnabled": boolean;
   "chain.defaultFeeRecipient": string;
   "chain.assertCorrectProgressiveBalances": boolean;
+  "chain.maxSkipSlots": number;
   "safe-slots-to-import-optimistically": number;
-  // this is defined as part of IBeaconPaths
-  // "chain.persistInvalidSszObjectsDir": string;
 }
 
 export function parseArgs(args: IChainArgs): IBeaconNodeOptions["chain"] {
@@ -25,6 +27,7 @@ export function parseArgs(args: IChainArgs): IBeaconNodeOptions["chain"] {
     proposerBoostEnabled: args["chain.proposerBoostEnabled"],
     defaultFeeRecipient: args["chain.defaultFeeRecipient"],
     assertCorrectProgressiveBalances: args["chain.assertCorrectProgressiveBalances"],
+    maxSkipSlots: args["chain.maxSkipSlots"],
     safeSlotsToImportOptimistically: args["safe-slots-to-import-optimistically"],
   };
 }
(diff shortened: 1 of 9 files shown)


packages/cli/src/options/beaconNodeOptions/chain.ts
```diff
@@ -66,6 +66,12 @@ Will double processing times. Use only for debugging purposes.",
     group: "chain",
   },
 
+  "chain.maxSkipSlots": {
```
dadepo, line 84: @dapplion should this option to be hidden? Also should it have a default value? I do not see a default value set in LH implementation though. If there is a need to set a default, what range of value will be advisable?

dapplion: Lighthouse doesn't have a default value because it's disabled by default. If the CLI value is not set the check does not apply

dadepo: Yeah. I was asking, should we also implement it that way (that is have it disabled by default, and hence do not have a default value)? Or do we consider the possibility of a DoS via this avenue high enough that we want to have a default value? (and hence have the check enabled by default)

dapplion: With this enabled by default you can also reject valid blocks so it's a dangerous feature if miss-used. I imagine that's why Lighthouse has disabled it by default

wemeetagain (dismissed):

packages/cli/src/options/beaconNodeOptions/chain.ts
```diff
@@ -8,9 +8,8 @@ export interface IChainArgs {
   "chain.persistInvalidSszObjects": boolean;
   "chain.proposerBoostEnabled": boolean;
   "chain.defaultFeeRecipient": string;
+  "chain.maxSkipSlots"?: number;
   "safe-slots-to-import-optimistically": number;
-  // this is defined as part of IBeaconPaths
-  // "chain.persistInvalidSszObjectsDir": string;
```
dapplion, line 14: Why is this comment deleted?

packages/lodestar/test/unit/chain/validation/block.test.ts
```diff
@@ -35,6 +36,9 @@ describe("gossip block validation", function () {
     chain.forkChoice = forkChoice;
     regen = chain.regen = sinon.createStubInstance(StateRegenerator);
 
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    (chain as any).opts = {maxSkipSlots};
```
dapplion: You can cast to the full class, which should allow editing. Better than any

dapplion (changes requested):

packages/cli/src/options/beaconNodeOptions/chain.ts
dadepo: Because I was not sure of the value of keeping the property commented out...it felt like something that was marked for removal. 

I have brought it back and updated it a bit to be more clear.

packages/lodestar/test/unit/chain/validation/block.test.ts
dadepo: Unfortunately on the full class ie `BeaconChain`, `opts` is a readonly value, so casting it to the full class will not allow editing

dadepo:
@dapplion I responded to your recent comments...if you are fine by them...is it possible to help merge?

packages/lodestar/test/unit/chain/validation/block.test.ts
dapplion: Please keep trying other strategies before casting to `any`, for example

```ts
((chain as Pick<BeaconChain, "opts">) as {opts: BeaconChain["opts"]}).opts = {maxSkipSlots}
```

Or what about just
```ts
chain.opts.maxSkipSlots = maxSkipSlots
```

Just keep iterating until you find the strongest chain of casting that will not compile if something changes

dapplion (approved):

PR #4022, opened 2022-05-16
Throw a more informative error on invalid keystore

wemeetagain:
**Motivation**

Seen in the private lodestar chat:

Has anyone ever seen this validator error before?
```
May-16 14:13:11.239[]                 info: Lodestar version=v0.34.0-beta.0/HEAD/+183/28e2c74c (git), network=prater
 ✖ SyntaxError: Unexpected token p in JSON at position 0
    at JSON.parse (<anonymous>)
    at Function.parse (/usr/app/node_modules/@chainsafe/bls-keystore/lib/class.js:53:41)
    at /usr/app/node_modules/@chainsafe/lodestar-cli/src/cmds/validator/keys.ts:63:44
    at Array.map (<anonymous>)
    at getLocalSecretKeys (/usr/app/node_modules/@chainsafe/lodestar-cli/src/cmds/validator/keys.ts:62:21)
    at Object.validatorHandler [as handler] (/usr/app/node_modules/@chainsafe/lodestar-cli/src/cmds/validator/handler.ts:68:68)
```

This error isn't very informative, but its due to an invalid keystore being loaded.

**Description**

This PR catches errors parsing a keystore, and throws an error with the keystore path

### DIFF
--- a/packages/cli/src/cmds/validator/keys.ts
+++ b/packages/cli/src/cmds/validator/keys.ts
@@ -60,9 +60,10 @@ export async function getLocalSecretKeys(
 
     const secretKeys = await Promise.all(
       keystorePaths.map(async (keystorePath) => {
+        const keystoreStr = fs.readFileSync(keystorePath, "utf8");
         let keystore;
         try {
-          keystore = Keystore.parse(fs.readFileSync(keystorePath, "utf8"));
+          keystore = Keystore.parse(keystoreStr);
         } catch (e) {
           throw new Error(`Error parsing keystore at ${keystorePath}: ${(e as Error).message}`);
         }
--- a/packages/cli/src/cmds/validator/keys.ts
+++ b/packages/cli/src/cmds/validator/keys.ts
@@ -61,12 +61,15 @@ export async function getLocalSecretKeys(
     const secretKeys = await Promise.all(
       keystorePaths.map(async (keystorePath) => {
         const keystoreStr = fs.readFileSync(keystorePath, "utf8");
+
         let keystore;
         try {
           keystore = Keystore.parse(keystoreStr);
         } catch (e) {
-          throw new Error(`Error parsing keystore at ${keystorePath}: ${(e as Error).message}`);
+          (e as Error).message = `Error parsing keystore at ${keystorePath}: ${(e as Error).message}`;
+          throw e;
         }
+
         return SecretKey.fromBytes(await keystore.decrypt(passphrase));
       })
     );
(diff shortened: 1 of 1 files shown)


philknows:
It would be helpful to indicate:
- What is the error specifically (ex. invalid JSON format, structure, etc.)
- Which file is the issue (ex. keystore_xxxx.json)

packages/cli/src/cmds/validator/keys.ts
```diff
@@ -59,9 +59,15 @@ export async function getLocalSecretKeys(
...
     const secretKeys = await Promise.all(
-      keystorePaths.map(async (keystorePath) =>
-        SecretKey.fromBytes(await Keystore.parse(fs.readFileSync(keystorePath, "utf8")).decrypt(passphrase))
-      )
+      keystorePaths.map(async (keystorePath) => {
+        let keystore;
+        try {
+          keystore = Keystore.parse(fs.readFileSync(keystorePath, "utf8"));
+        } catch (e) {
+          throw new Error("Error parsing keystore at " + keystorePath + ": " + (e as Error).message);
```
dadepo: I guess since the key store file does not contain any decrypted secrets, it should be fine to log the location (we recently had a PR that removed similar logging [here](https://github.com/ChainSafe/lodestar/pull/3975)

Suggestion: consider using `template literals` rather than concatenation.

dadepo (dismissed):

packages/cli/src/cmds/validator/keys.ts
wemeetagain: We removed the path of the passphrase from the logs since its more sensitive?

dadepo: > We removed the path of the passphrase from the logs since its more sensitive?

Yeah. The advisory that it triggered can be seen [here](https://codeql.github.com/codeql-query-help/python/py-clear-text-logging-sensitive-data/)

twoeths (dismissed):

packages/cli/src/cmds/validator/keys.ts
```diff
@@ -59,9 +59,19 @@ export async function getLocalSecretKeys(
...
+      keystorePaths.map(async (keystorePath) => {
+        const keystoreStr = fs.readFileSync(keystorePath, "utf8");
+
+        let keystore;
+        try {
+          keystore = Keystore.parse(keystoreStr);
+        } catch (e) {
+          (e as Error).message = `Error parsing keystore at ${keystorePath}: ${(e as Error).message}`;
+          throw e;
+        }
```
dapplion, line 71: @wemeetagain changed to only wrap the .parse step, not the file read. Also extended error to preserve stack trace

dapplion (approved):

PR #4029, opened 2022-05-17
Ensure light client update is in a single period

wemeetagain:
**Motivation**

Several reports of invalid light client updates from discord:

- Last week in sync committee period 48, I tested my light client and all the verifications passed. The period rolled over to 49, and my verification was no longer passing. I used the header slot to calculate in what period the header is, and saw that the headers are in period 48, still. I added a check in my code to check whether the calculated sync period and the headers in the committee update's period match, thinking that the headers might be updated to a better header further in sync period 49.

- API seems to give a wrong answer to period 49 (it belongs to period 48 though it's different from what is returned when asking for period 48). The next_committee_root is also the root of committee 49. The attested header is slot 401407 and the finalized header is 401312 (period 49 begins at 401408, so the attested header is right at the very end of the previous period). I believe both the attested and finalized headers should be in the requested period.

**Description**

Ensure that the signature period == attested header period === finalized header period in light client updates

- when getting a new sync aggregate, disregard it if the attested data is not in the same period
- when appending the finalized data to the update, don't append the finalized data if the finalized header isn't in the same period as the sync aggregate

### DIFF
--- a/packages/lodestar/src/chain/lightClient/index.ts
+++ b/packages/lodestar/src/chain/lightClient/index.ts
@@ -428,6 +428,8 @@ export class LightClientServer {
    * ```
    *
    * 3. On new blocks use `block.body.sync_aggregate`, `block.parent_root` and `block.slot - 1`
+   *
+   * @param syncPeriod The sync period of the sync aggregate and signed block root
    */
   private async onSyncAggregate(
     syncPeriod: SyncPeriod,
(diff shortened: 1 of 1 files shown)


dapplion (approved):

wemeetagain:
Confirmation from @zsfelfoldi that this fixes the issue

PR #4030, opened 2022-05-17
Simplify release process

wemeetagain:
**Motivation**

The release process was slightly error-prone and more tedious than necessary.

**Description**

- Add a release script with basic sanity checking, tagging, and pushing the tag.
- Update RELEASE.md to include usage.

The script automates the steps of "releasing" lodestar.
- optionally, bumping the version in package.jsons
- tag a specific commit
- push the tag to github

It ensures the following sanity checks:
- the tag follows the convention `v#.#.#` or `v#.#.#-beta.#` (eg: `v0.36.1` or `v0.36.1-beta.0`)
- the tag matches an existing branch `v#.#.x`
- the commit exists in the branch `v#.#.x`
- the user is prompted before continuing to the final step
- if the lerna.json matches the tag, the tag and push is performed manually using `git`
- if the lerna.json _doesn't_ match the tag, `lerna version` is used

Common usage:
- If beta release:
  - Checkout version branch (eg: `v0.36.x`)
  - Cherrypick patch commits on top
  - Run release script (eg: `yarn release -t v0.36.1-beta.0`)
- If non-beta release
  - Checkout version branch (eg: `v0.36.x`)
  - Ensure the checked out branch is the one to release
  - Run release script (eg: `yarn release -t v0.36.1`)

### DIFF
--- a/scripts/release.mjs
+++ b/scripts/release.mjs
@@ -129,6 +129,7 @@ async function lernaVersion(tag, yes) {
     await _lernaVersion({
       cwd: process.cwd(),
       bump: tag,
+      "force-publish": true,
       yes: yes,
     })
   } catch (e) {
--- a/scripts/release.mjs
+++ b/scripts/release.mjs
@@ -1,12 +1,15 @@
 import {execSync} from "node:child_process";
+import {readFileSync} from "node:fs";
 import yargs from "yargs";
 import {hideBin} from "yargs/helpers";
 import inquirer from "inquirer";
+import _lernaVersion from "@lerna/version";
 
 // Script to make releasing easier
 // Run with --help to see usage
 
 const cmd = (cmd, opts={}) => execSync(cmd, {encoding: "utf8", stdio: ["pipe", "pipe", "ignore"], ...opts}).trim();
+const exit = (...msg) => { console.log(...msg), process.exit(1) };
 
 const argv = yargs(hideBin(process.argv))
   .usage("Release lodestar")
@@ -23,7 +26,6 @@ const argv = yargs(hideBin(process.argv))
     commit: {
       alias: "c",
       type: "string",
-      default: cmd("git rev-parse --short HEAD"),
       describe: "The commit to tag",
     },
     yes: {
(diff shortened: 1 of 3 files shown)


scripts/release.mjs
```diff
@@ -0,0 +1,104 @@
...
+      name: "yes",
+      type: "confirm",
+      message: "Do you want to proceed?",
+    },
+  ]);
+  if (!input.yes) {
+    process.exit(1);
+  }
+}
+
```
twoeths, line 139: how do we prevent this error `TAG v0.36.1-beta.0 does not include LOCAL_VERSION 0.36.0` from https://github.com/ChainSafe/lodestar/runs/6476719731?check_suite_focus=true ? 

if yes, we should do a check here before we create a tag.

dapplion: That's a good idea! This condition should be checked here too

wemeetagain: The script now calls `lerna version $TAG` if the tag does not match `lerna.json` (which bumps the versions in package.jsons, commits, tags and pushes).
Otherwise, if the tag matches `lerna.json`, it tags and pushes manually with git.

This makes it very handy to just checkout the version branch (eg: v0.36.x), cherry pick commits on top, and run the script, in the case of a beta release.
Or just checkout the version branch and run the script, in the case of a non-beta release.

scripts/release.mjs
```diff
@@ -0,0 +1,166 @@
...
+  }
+  return lernaVersion === tag;
+}
+
+async function lernaVersion(tag, yes) {
+  try {
+    await _lernaVersion({
+      cwd: process.cwd(),
+      bump: tag,
+      yes: yes,
```
wemeetagain, line 133: @g11tech @tuyennhv does this call to `lerna version` need other flags too?

g11tech: I think `--force-publish` we rely on for both normal/nightly release and the additional `--exact` for nightly

g11tech (approved):

Issue #4027, opened 2022-05-17
[PROD] Validator Proposal Missed


twoeths:
**Describe the bug**

<img width="1451" alt="Screen Shot 2022-05-17 at 16 22 22" src="https://user-images.githubusercontent.com/10568965/168777506-8c85ce97-f33d-4cfd-a51a-b55554e08789.png">

No log was captured since it's 5 days ago, and this happened with version v0.36.0

**Expected behavior**

No error

twoeths:
issue is too old, lodestar is able to publish blocks consistently as with v1.2.1

PR #4032, opened 2022-05-18
Increase logMaxFiles

twoeths:
**Motivation**

We miss a block proposal and was not able to find its log because it's > 5 days ago

**Description**

+ Remove the "`Attestation processed`" log, this log account for roughly 80% of log file
+ Increase `logMaxFiles` to 14

part of #4027

### DIFF
--- a/docker-compose.validator.yml
+++ b/docker-compose.validator.yml
@@ -8,7 +8,7 @@ services:
       - logs:/logs
       - ./keystores:/keystores
     env_file: .env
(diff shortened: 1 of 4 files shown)


dapplion:
@tuyennhv I think we should only increase this numbers in our infrastructure. 14 days of files can be too much for regular consumers

dapplion:
@tuyennhv Tho could you do a separate PR to remove this log? That's valuable for everyone

https://github.com/ChainSafe/lodestar/pull/4032/files#diff-b8d5a4f6ccc549395e646ab2f2f7f5dcd5d62810fb91b0be8f47eee455f6e8a1L177-L183

PR #4019, opened 2022-05-13
Ignore gossip AggregateAndProof if aggregate is seen

twoeths:
**Motivation**

Address this spec when validating `beacon_aggregate_and_proof` gossip message:
```
[IGNORE] The valid aggregate attestation defined by hash_tree_root(aggregate) has not already been seen (via aggregate gossip, within a verified block, or through the creation of an equivalent aggregate locally).
```

**Description**

+ Implement `SeenAggregatedAttestations`
+ Populate seen aggregated attestations from api / gossip / blocks
+ Check if the aggregated attestation is seen when validating gossip `AggregateAndProof` message

Closes #3794

**TODO**
+ [x] Deploy and test in a node, make sure it takes less time to validate AggregateAndProof gossip messages

PR #4034, opened 2022-05-18
Oppool aggregates use BitArray only for set logic

dapplion:
**Motivation**

From https://github.com/ChainSafe/lodestar/pull/4019 I noticed that we don't really have to use the validator indexes to compare the bit sets of attestations and check if they are subset, superset or intersect. Since all the attestations to be compared are in the same slot and committee, the bits are sufficient for that.

From performance tests dealing with BitArray directly instead of arrays and ES6 Sets is much faster and (in my subjective opinion) simpler conceptually.

**Description**

- Use BitArrays only to compare attestation bit sets
- Remove attestingIndices from data structures as it's not necessary now
- Re-arrange tests for better maintainability
- Move getParticipation logic to standalone function to be able to test only that

### DIFF
--- a/packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
+++ b/packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
@@ -94,10 +94,8 @@ export class AggregatedAttestationPool {
     const stateSlot = state.slot;
     const stateEpoch = state.epochCtx.epoch;
     const statePrevEpoch = stateEpoch - 1;
-    const forkName = state.config.getForkName(stateSlot);
 
-    const getParticipationFn =
-      forkName === ForkName.phase0 ? this.getParticipationPhase0(state) : this.getParticipationAltair(state);
+    const getParticipation = getParticipationFn(state);
 
     const attestationsByScore: AttestationWithScore[] = [];
 
@@ -132,7 +130,7 @@ export class AggregatedAttestationPool {
         ) {
           continue;
         }
-        const participation = getParticipationFn(epoch, attestationGroup.committee);
+        const participation = getParticipation(epoch, attestationGroup.committee);
         if (participation === null) {
           continue;
         }
@@ -334,6 +332,7 @@ export class MatchingDataAttestationGroup {
       let notSeenAttesterCount = 0;
 
       for (let i = 0; i < committeeLen; i++) {
+        // TODO: Optimize aggregationBits.get() in bulk for the entire BitArray
         if (!committeeSeenAttesting[i] && aggregationBits.get(i)) {
           notSeenAttesterCount++;
         }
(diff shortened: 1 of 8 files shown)


twoeths:
The idea is great, not sure why `getAttestationsForBlock()` benchmark test (78ms/op) is worse than master (60ms/op) @dapplion , does it matter?

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -265,64 +271,80 @@ export class MatchingDataAttestationGroup {
...
+
+    for (const {attestation} of this.attestations) {
+      const {aggregationBits} = attestation;
       let notSeenAttesterCount = 0;
-      for (const attIndex of attestation.attestingIndices) {
-        if (!seenAttestingIndices.has(attIndex)) notSeenAttesterCount++;
+
+      for (let i = 0; i < committeeLen; i++) {
+        // TODO: Optimize aggregationBits.get() in bulk for the entire BitArray
+        if (!committeeSeenAttesting[i] && aggregationBits.get(i)) {
```
twoeths, line 283: this is the hot path. If we store `attestingIndices: number[]` in `AttestationWithIndex`, and change `seenAttestingIndices` param to `number[]` and make sure they are in the order of the committee,  this could be faster?
```typescript
let seenCount = 0;
let checkIndex = 0;
for (let i = 0; i < seenAttestingIndices.length; i++) {
  if (checkIndex >= attestingIndices.length) break;
  if(seenAttestingIndices[i] === attestingIndices[checkIndex) {
    seenCount ++;
    checkIndex++;
  }
}
const notSeenAttesterCount = attestingIndices.length - seenCount;
```

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -90,10 +94,8 @@ export class AggregatedAttestationPool {
     const stateSlot = state.slot;
     const stateEpoch = state.epochCtx.epoch;
     const statePrevEpoch = stateEpoch - 1;
-    const forkName = state.config.getForkName(stateSlot);
 
-    const getParticipationFn =
-      forkName === ForkName.phase0 ? this.getParticipationPhase0(state) : this.getParticipationAltair(state);
```
twoeths, line 96: need to remove `getParticipationPhase0` and `getParticipationAltair`

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
twoeths: nvm, this is incorrect

packages/lodestar/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -345,6 +363,62 @@ export function aggregateInto(attestation1: AttestationWithIndex, attestation2:
...
+      for (const validatorIndex of committee) {
+        if (flagIsTimelySource(participationStatus[validatorIndex])) {
+          seenValidatorIndices.add(validatorIndex);
+        }
+      }
+      return seenValidatorIndices;
+    };
+  }
+}
+
```
twoeths, line 368: if the below `intersection()` function is not used, remove it

dapplion:
> not sure why getAttestationsForBlock() benchmark test (78ms/op) is worse than master (60ms/op) @dapplion , does it matter?

Benchmarks in CI are very unstable, 60 -> 78 unless observed multiple times locally is inconclusive. Happy to merge your branch tho if you believe it's a worthwhile improvement

twoeths:
> > not sure why getAttestationsForBlock() benchmark test (78ms/op) is worse than master (60ms/op) @dapplion , does it matter?
> 
> Benchmarks in CI are very unstable, 60 -> 78 unless observed multiple times locally is inconclusive. Happy to merge your branch tho if you believe it's a worthwhile improvement

that's actually not a worthwhile improvement, we'll likely have less AggregateAndProof gossip messages thanks to #4019, we can consider that PR if this place becomes a bottle neck

twoeths (approved):

PR #4038, opened 2022-05-19
Implement countNotSeenAttester for OpPool

twoeths:
**Motivation**

Improve performance of aggregated op pool, started from #4034

**Description**

+ Implement `countNotSeenAttester` given committee, attesting indices, seen attesting indices in the same order
+ Clean up some code


twoeths:
@dapplion if we pass in attestingIndices (which we already do) and seenAttestingIndices in committee order, the `getAttestationsForBlock` benchmark test could finish in 35ms/op

twoeths:
While this PR proves a (2x) faster improvement, it's not a worth-wide improvement, will reconsider this PR if `getAttestationsForBlock()` is a bottle neck in the future

PR #4056, opened 2022-05-25
Assert tag in release workflow

### LINKED ISSUE #4050: Lodestar release unintended version
**Describe the bug**

When testing #4048 , I tried `lerna version` without `--no-git-tag-version` flag and release flow was triggered unintentionally with version `chore(release): v0.37.0-dev.ef9528aa59`

**Expected behavior**

Release workflow should validate the tag name, or the branch before creating the release

twoeths:
**Motivation**

An unintended release was created because I missed `--no-git-tag-version` param when I try `lerna version` command locally

**Description**

- Release workflow should assert tag version
- Create `assert-tag.mjs` to be used for both release workflow and `release.mjs` script

Closes #4050

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,6 +24,11 @@ jobs:
       - name: Get tag
         id: get_tag
         run: echo ::set-output name=tag::${GITHUB_REF/refs\/tags\//}
+      - name: Assert tag name is valid
+        id: assert_tag
+        run: node scripts/assert_tag.mjs
+        env:
+          TAG: ${{ steps.get_tag.outputs.tag }}
       - name: Get previous tag
         id: get_prev_tag
         run: node scripts/get_prev_tag.js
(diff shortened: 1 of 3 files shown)


g11tech (dismissed):

.github/workflows/release.yml
```diff
@@ -24,6 +24,11 @@ jobs:
       - name: Get tag
         id: get_tag
         run: echo ::set-output name=tag::${GITHUB_REF/refs\/tags\//}
+      - name: Assert tag name is valid
+        id: assert_tag
+        run: node scripts/assert_tag.mjs
```
wemeetagain, line 29: I think we need something like:
```ts
node -e 'import("./scripts/assert_tag.mjs").then(({assertTag}) => assertTag())'
```

wemeetagain (changes requested):

.github/workflows/release.yml
dapplion: Why?

twoeths:
closing the PR as there is new release process

Issue #4004, opened 2022-05-10
Run forkchoice update pre `assembleBlock`


g11tech:
Run forkchoice update pre `assembleBlock` in the produceBlock call to get the latest head at begining of proposal slot `s` to also include attestations of slot `s-1`'

Context; https://github.com/ethereum/consensus-specs/pull/2878

g11tech:
resolved in https://github.com/ChainSafe/lodestar/pull/4058

PR #4058, opened 2022-05-25
Run forkChoice to get correct head for proposal

g11tech:
**Motivation**
[Run forkchoice update pre assembleBlock](https://github.com/ChainSafe/lodestar/issues/4004)

**Description**
Runs updateTime and updateHead to process any queued attestations to find the correct head for the proposal slot.

Context Discussion/specs PR:
https://github.com/ethereum/consensus-specs/pull/2878
https://github.com/ethereum/consensus-specs/pull/2897

### DIFF
--- a/packages/lodestar/src/api/impl/validator/index.ts
+++ b/packages/lodestar/src/api/impl/validator/index.ts
@@ -176,6 +176,12 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
       notWhileSyncing();
       await waitForSlot(slot); // Must never request for a future slot > currentSlot
 
+      // Process the queued attestations in the forkchoice for correct head estimation
+      // forkChoice.updateTime() might have already been called by the onSlot clock
+      // handler, in which case this should just return.
+      chain.forkChoice.updateTime(slot);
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/api/impl/validator/index.ts
```diff
@@ -176,6 +176,12 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
       notWhileSyncing();
       await waitForSlot(slot); // Must never request for a future slot > currentSlot
 
+      // Process the queued attestations in the forkchoice for correct head estimation
+      // forkChoice.updateTime() might have already been called by the onSlot clock
+      // handler, in which case this should just return.
+      chain.forkChoice.updateTime(slot);
```
twoeths, line 182: calling `forkChoice.updateTime()` multiple times would calling `processAttestationQueue()` multiple times.

I think we should do early return in `forkChoice.updateTime(slot)` if `this.fcStore.currentSlot >= slot`

g11tech: makes sense, updating! :+1:

twoeths (approved):

PR #4062, opened 2022-05-25
Assert equal params with only whitelisted keys

dapplion:
**Motivation**

Lodestar validator client asserts that the beacon node is connecting to has the same configuration. This check is very important to exit early in case of misconfiguration. However, in the past it has broke interoperability with other clients due to the checks being too strict on params that are not guaranteed to be the same.

Latest consensus release changed `SHARDING_FORK_VERSION`. We don't know when each client will update this value since it doesn't matter now.

https://github.com/ethereum/consensus-specs/pull/2896/files#diff-fa42a5809dc8042fed427fd9d2d54c2595789c5cad751b03a8d118871285c26fL48

We need a more resilient check that only looks are "relevant" keys, and ignores the ones that are likely to change or not critical for interoperability.

**Description**

Assert equal params with only whitelisted keys.

So this check only compares a specific list of parameters that are consensus critical, ignoring the rest. Typed config and preset ensure new parameters are leveled critical or ignore, facilitating maintenance of the list

Please @g11tech and @dadepo can you ensure that this PR doesn't break interop with other CL clients?

### DIFF
--- a/packages/validator/src/util/params.ts
+++ b/packages/validator/src/util/params.ts
@@ -64,7 +64,7 @@ export function assertEqualParams(localConfig: IChainConfig, externalSpecJson: R
   }
 
   if (errors.length > 0) {
-    throw new NotEqualParamsError("Not equal BeaconParams\n" + errors.join("\n"));
+    throw new NotEqualParamsError("Local and remote configs are different\n" + errors.join("\n"));
   }
 }
 
--- a/packages/validator/src/util/params.ts
+++ b/packages/validator/src/util/params.ts
@@ -37,29 +37,29 @@ export function assertEqualParams(localConfig: IChainConfig, externalSpecJson: R
   const errors: string[] = [];
 
   for (const key of Object.keys(criticalParams) as (keyof typeof criticalParams)[]) {
-    // Ignore non-critical params
-    if (!criticalParams[key]) {
+    if (
+      // Ignore non-critical params
+      !criticalParams[key] ||
+      // This condition should never be true, but just in case
+      localSpecJson[key] === undefined ||
+      // The config/spec endpoint is poorly specified, so in practice each client returns a custom selection of keys.
+      // For example Lighthouse returns a manually selected list of keys that may be updated at any time.
+      // https://github.com/sigp/lighthouse/blob/bac7c3fa544495a257722aaad9cd8f72fee2f2b4/consensus/types/src/chain_spec.rs#L941
+      //
+      // So if we assert that spec critical keys are present in the spec we may break interoperability unexpectedly.
+      // So it's best to ignore keys are not defined in both specs and trust that the ones defined are sufficient
+      // to detect spec discrepancies in all cases.
+      externalSpecJson[key] === undefined
+    ) {
       continue;
-    }
-
-    // This condition should never be true, but just in case
-    if (localSpecJson[key] === undefined) {
-      errors.push(`${key} not defined in local config`);
-    }
-
-    // All consensus critical keys must be defined, otherwise we can't ensure interoperability
-    else if (externalSpecJson[key] === undefined) {
-      errors.push(`${key} not defined in external config`);
     }
 
     // Must compare JSON serialized specs, to ensure all strings are rendered in the same way
     // Must compare as lowercase to ensure checksum addresses and names have same capilatization
-    else {
-      const localValue = String(localSpecJson[key]).toLocaleLowerCase();
-      const remoteValue = String(externalSpecJson[key]).toLocaleLowerCase();
-      if (localValue !== remoteValue) {
-        errors.push(`${key} different value: ${localValue} != ${remoteValue}`);
-      }
+    const localValue = String(localSpecJson[key]).toLocaleLowerCase();
+    const remoteValue = String(externalSpecJson[key]).toLocaleLowerCase();
+    if (localValue !== remoteValue) {
+      errors.push(`${key} different value: ${localValue} != ${remoteValue}`);
     }
   }
 
--- a/packages/validator/src/util/params.ts
+++ b/packages/validator/src/util/params.ts
@@ -20,7 +20,7 @@ type ConfigWithPreset = IChainConfig & BeaconPreset;
  * parameters.
  *
  * So this check only compares a specific list of parameters that are consensus critical, ignoring the rest. Typed
- * config and preset ensure new parameters are leveled critical or ignore, faciliating manteinance of the list.
+ * config and preset ensure new parameters are labeled critical or ignore, facilitating maintenance of the list.
  */
 export function assertEqualParams(localConfig: IChainConfig, externalSpecJson: Record<string, string>): void {
   // Before comparing, add preset which is bundled in api impl config route.
--- a/packages/validator/src/util/params.ts
+++ b/packages/validator/src/util/params.ts
@@ -173,23 +173,23 @@ function getSpecCriticalParams(localConfig: IChainConfig): Record<keyof ConfigWi
     // # AltairPreset
     /////////////////
 
-    SYNC_COMMITTEE_SIZE: true,
-    EPOCHS_PER_SYNC_COMMITTEE_PERIOD: true,
-    INACTIVITY_PENALTY_QUOTIENT_ALTAIR: true,
-    MIN_SLASHING_PENALTY_QUOTIENT_ALTAIR: true,
-    PROPORTIONAL_SLASHING_MULTIPLIER_ALTAIR: true,
+    SYNC_COMMITTEE_SIZE: altairForkRelevant,
+    EPOCHS_PER_SYNC_COMMITTEE_PERIOD: altairForkRelevant,
+    INACTIVITY_PENALTY_QUOTIENT_ALTAIR: altairForkRelevant,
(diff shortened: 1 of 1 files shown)


packages/validator/src/util/params.ts
```diff
@@ -29,3 +65,131 @@ export function assertEqualParams(localConfig: IChainConfig, externalSpecJson: R
     throw new NotEqualParamsError("Not equal BeaconParams\n" + errors.join("\n"));
   }
 }
+
+function getSpecCriticalParams(localConfig: IChainConfig): Record<keyof ConfigWithPreset, boolean> {
```
dadepo, line 71: Maybe have this function in `@chainsafe/lodestar-params`? as that seems to be the package where we have params related functionality

packages/validator/src/util/params.ts
```diff
@@ -29,3 +65,131 @@ export function assertEqualParams(localConfig: IChainConfig, externalSpecJson: R
...
+    // # Config
+    ///////////
+
+    PRESET_BASE: false, // Not relevant, each preset value is checked below
+    TERMINAL_TOTAL_DIFFICULTY: bellatrixForkRelevant,
+    TERMINAL_BLOCK_HASH: bellatrixForkRelevant,
+    TERMINAL_BLOCK_HASH_ACTIVATION_EPOCH: bellatrixForkRelevant,
+
+    // Genesis
+    MIN_GENESIS_ACTIVE_VALIDATOR_COUNT: true,
```
dadepo, line 86: Instead of having a map of params to boolean, is it possible to have a list instead? Thus if a param is contained in the list, then it should be considered critical and be checked. 

I think with that, it becomes easier know the defined list and see what params are currently supported, instead of checking the boolean value in a map

packages/validator/src/util/params.ts
```diff
@@ -1,25 +1,61 @@
...
+ *
+ * This check however can't compare the full config as is, since some parameters are not critical to the spec and
+ * can be changed un-expectedly. Also, fork parameters can change un-expectedly, like their _FORK_VERSION or _EPOCH.
+ * Note that the config API endpoint is not precisely specified, so each clients can return a different set of
+ * parameters.
+ *
+ * So this check only compares a specific list of parameters that are consensus critical, ignoring the rest. Typed
+ * config and preset ensure new parameters are leveled critical or ignore, faciliating manteinance of the list.
  */
 export function assertEqualParams(localConfig: IChainConfig, externalSpecJson: Record<string, string>): void {
```
dadepo, line 25: it would also be nice to update the tests for this logic with this new requirement

packages/validator/src/util/params.ts
```diff
@@ -1,25 +1,61 @@
...
+ * Otherwise, signatures may be rejected, time may be un-equal and other bugs that are harder to debug caused
+ * by different parameters.
+ *
+ * This check however can't compare the full config as is, since some parameters are not critical to the spec and
+ * can be changed un-expectedly. Also, fork parameters can change un-expectedly, like their _FORK_VERSION or _EPOCH.
+ * Note that the config API endpoint is not precisely specified, so each clients can return a different set of
+ * parameters.
+ *
+ * So this check only compares a specific list of parameters that are consensus critical, ignoring the rest. Typed
+ * config and preset ensure new parameters are leveled critical or ignore, faciliating manteinance of the list.
```
wemeetagain:
```suggestion
 * config and preset ensure new parameters are labeled critical or ignore, facilitating maintenance of the list.
```

g11tech:
got this (lighthouse BN):
`./lodestar validator --network kiln --interopIndexes "0..5"  --server http://167.172.234.224:4000`
```typescript
May-25 18:11:16.064[]                 info: Genesis available
 ✖ Error: Not equal BeaconParams
INACTIVITY_PENALTY_QUOTIENT_BELLATRIX not defined in external config
INACTIVITY_PENALTY_QUOTIENT_BELLATRIX different value: 16777216 != undefined
MIN_SLASHING_PENALTY_QUOTIENT_BELLATRIX not defined in external config
MIN_SLASHING_PENALTY_QUOTIENT_BELLATRIX different value: 32 != undefined
PROPORTIONAL_SLASHING_MULTIPLIER_BELLATRIX not defined in external config
PROPORTIONAL_SLASHING_MULTIPLIER_BELLATRIX different value: 3 != undefined
MAX_BYTES_PER_TRANSACTION not defined in external config
MAX_BYTES_PER_TRANSACTION different value: 1073741824 != undefined
MAX_TRANSACTIONS_PER_PAYLOAD not defined in external config
MAX_TRANSACTIONS_PER_PAYLOAD different value: 1048576 != undefined
BYTES_PER_LOGS_BLOOM not defined in external config
BYTES_PER_LOGS_BLOOM different value: 256 != undefined
MAX_EXTRA_DATA_BYTES not defined in external config
MAX_EXTRA_DATA_BYTES different value: 32 != undefined
    at assertEqualParams (file:///usr/app/packages/validator/src/util/params.ts:65:11)
    at Function.initializeFromBeaconNode (file:///usr/app/packages/validator/src/validator.ts:161:5)

```

g11tech:
prsym:
`./lodestar validator --network kiln --interopIndexes "0..5"  --server http://164.92.130.39:4000`

```typescript
May-25 18:16:39.480[]                 info: Lodestar network=kiln, version=v0.37.0/c165a5c, commit=c165a5cf7f56b34d37492942db9b094d6ee3c0b5
May-25 18:16:39.571[]                 info: Decrypted 6 local keystores
May-25 18:16:39.571[]                 info: 0xb8720175cf41e0a67313224e85f86ad6f2ee0f775717646b21fb373eafeb015cdf88721387a989bfe71676405f01552c
May-25 18:16:39.571[]                 info: 0x93a8b44b4a6cee4611cf2b3efc689cf7ca9a6be3638fefe8ecf10da98b9f03e2bef867bf3a7d3ae77cd86a31c8d83cf5
May-25 18:16:39.572[]                 info: 0xb76d52873e0e1ebd2746af664a1ca015e5ff6c02287bf010b550123b752c31659c5207c70670ee34b07a7bb25b564529
May-25 18:16:39.572[]                 info: 0xa1da06aff8e369e5c871c4b63e930e8b5359f05591e6f336684f1e9ee8e952ed7541f082d9482f01d934501f3de58c1e
May-25 18:16:39.572[]                 info: 0xa3c9b8d51584e7f37c3bef871c9339eeb0ac805705a7aaf151becde5322edded237198ea798791d68b84444eb6dc0e24
May-25 18:16:39.572[]                 info: 0xb2fd9f2f484781147238cc0bc1cc80c0f4f05d28efd8813dcae7a5f8cde0ee2074dbca69fac9b25f9c4d87705732fd37
May-25 18:16:40.022[]                 info: Genesis available
 ✖ Error: Not equal BeaconParams
MAX_BYTES_PER_TRANSACTION not defined in external config
MAX_BYTES_PER_TRANSACTION different value: 1073741824 != undefined
MAX_TRANSACTIONS_PER_PAYLOAD not defined in external config
MAX_TRANSACTIONS_PER_PAYLOAD different value: 1048576 != undefined
BYTES_PER_LOGS_BLOOM not defined in external config
BYTES_PER_LOGS_BLOOM different value: 256 != undefined
MAX_EXTRA_DATA_BYTES not defined in external config
MAX_EXTRA_DATA_BYTES different value: 32 != undefined
    at assertEqualParams (file:///usr/app/packages/validator/src/util/params.ts:65:11)
    at Function.initializeFromBeaconNode (file:///usr/app/packages/validator/src/validator.ts:161:5)
    at Object.validatorHandler [as handler] (file:///usr/app/packages/cli/src/cmds/validator/handler.ts:122:21)

```

packages/validator/src/util/params.ts
dapplion: Yes but `@chainsafe/lodestar-params` doesn't care about this. This feature is very specific to a validator client that connects to the beacon node.

packages/validator/src/util/params.ts
dapplion: If we had a list of keys we lose the guarantee that Typescript provides that all keys have been manually assigned a "true" or "false" by a developer. Otherwise new keys will be assumed as "false" which is dangerous and not maintainable, as we would forget to add new keys.

dapplion:
@g11tech thanks for checking! Makes sense, all of those params are for future forks so I will ignore them conditionally.

dapplion:
Not that easy lol, checked Lighthouse source and they manually declare which parameters to expose. The list is a bit random and can be updated at any point without notice. What should we do?

https://github.com/sigp/lighthouse/blob/bac7c3fa544495a257722aaad9cd8f72fee2f2b4/consensus/types/src/chain_spec.rs#L941

dapplion:
Changed code to only consider keys that are:
- manually specified relevant i.e. won't change AND are critical
- are declared locally
- are declared in the remote spec

https://github.com/ChainSafe/lodestar/blob/d3a460f400d2bee57916e08918cb35a7f2c2903c/packages/validator/src/util/params.ts#L45-L52

@g11tech tests work
```
./lodestar validator --network kiln --interopIndexes "0..5" --server http://167.172.234.224:4000
```
```
./lodestar validator --network kiln --interopIndexes "0..5" --server http://164.92.130.39:4000
```

g11tech (approved):

wemeetagain (approved):

PR #4067, opened 2022-05-26
Bump typescript to v4.7.4

wemeetagain:
**Motivation**

Typescript 4.7 includes additional ESM compatibility that should be explored.

**Description**

Bump typescript now, figure out configuration later


dapplion:
:rocket: 

Since 4.7 supports `node16` and `nodenext` `compilerOptions.module` option, should we change it? I'm not sure what practical implications does it have to us or if we should use it.

https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/?q=1#esm-nodejs

wemeetagain:
maybe we can just merge this as-is and figure out the compiler options later. Will make an issue to track.

wemeetagain:
- https://github.com/ChainSafe/lodestar/issues/4152

philknows:
Anything blocking this upgrade from happening? Should know the priority of this so it doesn't just sit here.

wemeetagain:
Nothing blocking this. Also typescript is up to 4.7.4.
Imo low priority but also like a 1 line change so I'll update this and we can merge it.

packages/lodestar/src/chain/blocks/importBlock.ts
```diff
@@ -187,6 +192,35 @@ export async function importBlock(chain: ImportBlockModules, fullyVerifiedBlock:
     const cp = getCheckpointFromState(checkpointState);
     chain.checkpointStateCache.add(cp, checkpointState);
     pendingEvents.push(ChainEvent.checkpoint, cp, checkpointState);
+
+    // Note: in-lined code from previos handler of ChainEvent.checkpoint
+    chain.logger.verbose("Checkpoint processed", toCheckpointHex(cp));
```
dapplion: What's up with this diff? @wemeetagain

wemeetagain: I think its bc of the force-push to `unstable`

wemeetagain: should be fixed now

dapplion (approved):

PR #4070, opened 2022-05-26
Bump to v0.38.0 nightly builds

wemeetagain:
**Motivation**

As part of the release flow, start version v0.38.0. Bump master to produce nightly builds for this version.

**Description**

Bump package.json versions to v0.38.0


dapplion:
- We could adopt this strategy to prevent needing these bumps https://github.com/ChainSafe/lodestar/pull/4071

wemeetagain:
I think given #4071 we don't want do do this

Issue #4002, opened 2022-05-10
Extreme I/O lag on overloaded nodes


dapplion:
NodeJS due to its single threaded nature experiences exponential degradation of performance under heavy load. We have circumstantial evidence that in low power machines, when attempting to run too many keys performance suffers. Overall time to perform tasks increases, but any network call (vc call beacon api) time increases by x10, x100 respective to internal function times (block processing)

By running large amounts of keys on not shitty servers the issue is not apparent. However this issue still poses a future risk
- for solo stakers attempting to run Lodestar in under-powered machines
- post-merge due to extra calls to EL and mev-boost on time critical code paths

**Step 1: quantify**

Circumstantial evidence is not enough, we need proper documented metrics on when this issue manifest and how bad it is. We will run Lodestar in different configurations and different servers:

Validator configurations:
- beacon 1 validator
- beacon 1 validator --subscribeAllSubnets
- beacon + 8 validators
- beacon + 16 validators
- beacon + 32 validators
- beacon + 64 validators

Servers:
- Contabo Cloud VPS S
- Contabo Cloud VPS M
- Hetzner cloud CX41

Data to collect:
- Validator request times histogram `vc_rest_api_client_request_time_seconds_bucket{routeId="produceAttestationData"}`. Average, median, p95, p99.

**Step 2: reproduce**

If the data above proves that this issue of sufficient severity we need to have a reproducible simple case to collect help and potential solutions. A reproducible case must be:
- Standalone repo
- Contain no Lodestar specific code, nor any library at all, pure vanilla NodeJS code 
- Create similar conditions to which Lodestar experiences: very high load caused by many short async tasks
- Show that requests times by an external caller (in a separate process) show a similar pattern to the data collected in step 1

**Step 3: mitigate**

> Disclaimer: The root solution here is to reduce Lodestar load via optimizations of offloading to workers.

Provided that a Lodestar is under heavy load we want to minimize this I/O lag to acceptable levels, such that validator performance doesn't degrade. If an VC request time increases from 50ms to 500ms, it's tolerable. However if it increases from 50ms to 5000ms, the validator may miss an attestation and reduce profitability.

**Related issues**

May be the cause of this issues:

- https://github.com/ChainSafe/lodestar/issues/3110
- https://github.com/ChainSafe/lodestar/issues/3442
- https://github.com/ChainSafe/lodestar/issues/3584
- https://github.com/ChainSafe/lodestar/issues/3694
[...]

PR #4075, opened 2022-05-27
Use async version of crypto.randomBytes()

twoeths:
**Motivation**

`crypto.randomBytes()` has both sync and async version, as in the documentation of NodeJS:

```
The crypto.randomBytes() method will not complete until there is sufficient entropy available. This should normally never take longer than a few milliseconds. The only time when generating the random bytes may conceivably block for a longer period of time is right after boot, when the whole system is still low on entropy.
```

**Description**

Switch to async version of `crypto.randomBytes()`

part of #4002

### DIFF
--- a/packages/lodestar/src/network/peers/discover.ts
+++ b/packages/lodestar/src/network/peers/discover.ts
@@ -19,6 +19,8 @@ const MAX_CACHED_ENRS = 100;
 /** Max age a cached ENR will be considered for dial */
 const MAX_CACHED_ENR_AGE_MS = 5 * 60 * 1000;
 
+const randomBytesAsync = promisify(crypto.randomBytes);
+
 export type PeerDiscoveryOpts = {
   maxPeers: number;
   discv5FirstQueryDelayMs: number;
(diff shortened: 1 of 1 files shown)


packages/lodestar/src/network/peers/discover.ts
```diff
@@ -226,13 +226,18 @@ export class PeerDiscovery {
...
     }
 
-    const randomNodeId = crypto.randomBytes(64).toString("hex");
-
     this.randomNodeQuery = {code: QueryStatusCode.Active, count: 0};
     const timer = this.metrics?.discovery.findNodeQueryTime.startTimer();
 
     try {
-      const enrs = await this.discv5.findNode(randomNodeId);
+      // async version of crypto.randomBytes
```
wemeetagain: Consider using https://nodejs.org/api/util.html#utilpromisifyoriginal (this dependency doesn't work in browser but this module isn't expected to run in browser)

packages/lodestar/src/network/peers/discover.ts
```diff
@@ -226,13 +226,18 @@ export class PeerDiscovery {
...
+      const randomNodeId = await new Promise<Buffer>((resolve, reject) => {
```
dapplion: Why move randomBytes inside the try / catch for findNode()?

dapplion (approved):

wemeetagain (approved):

PR #4081, opened 2022-05-29
Correct Aggregate Inclusion metric

twoeths:
**Motivation**

Due to #4019, Aggregate Inclusion metric dropped. We want to get it back to the previous state.

Screen Shot 2022-05-29 at 15 23 21

**Description**

+ Call the metric before the `chain.seenAggregatedAttestations.isKnown()` check, about 70% of AggregateAndProof gossip messages were caught by that

### DIFF
--- a/packages/lodestar/src/api/impl/validator/index.ts
+++ b/packages/lodestar/src/api/impl/validator/index.ts
@@ -439,8 +439,9 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
           try {
             // TODO: Validate in batch
             const {indexedAttestation, committeeIndices} = await validateGossipAggregateAndProof(
-              chain,
-              signedAggregateAndProof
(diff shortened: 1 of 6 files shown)


twoeths:
Tested in contabo-8

Screen Shot 2022-05-29 at 17 01 31

wemeetagain (commented):
wont this introduce false positives too? invalid attestations being counted

dapplion (changes requested):
Initially I though it was important to track all aggregate for inclusion.

However, seeing the costs and dangers of doing so I am fine by only tracking novel aggregate inclusion. The metric will still fulfill it's purpose, which is showing that we got aggregated. To detect poor gossip performance we can still check the duplicate factor in gossip metrics. The combination of both should provide the same level of info as before.

So I would vote for closing this PR and ignoring my warning for v0.37.0 release

PR #4094, opened 2022-05-31
v0.38.0 release

g11tech:
v0.38.0 release


dapplion:
Possibility of a very slow memory leak in the beacon node. Does not prevent a release but want to leave a note in case this persists 

_hetzner-test1_

Screenshot from 2022-06-07 21-40-23

dapplion:
Some strange metrics in _hetzner-test1_ where metadata outgoing request errors jumps to 100%. This behaviour does not occur on other contabo beta deployments.

Screenshot from 2022-06-07 21-46-10

Screenshot from 2022-06-07 21-46-20

wemeetagain (approved):

PR #4107, opened 2022-06-03
Fix HttpClient event handler leak

dapplion:
**Motivation**

- https://github.com/ChainSafe/lodestar/issues/4087 reported a resource leak in the validator client

**Description**

Remove the listener added and not a different function

Closes https://github.com/ChainSafe/lodestar/issues/4087

PR #4108, opened 2022-06-03
Simplify HttpClient global AbortSignal handling

dapplion:
**Motivation**

Simplify how global signals are provided to the HttpClient instance.

- From https://github.com/ChainSafe/lodestar/pull/4107

**Description**

Instead of providing a function to get the global signal everytime, add a method to update the global signal only when necessary. In my opinion it makes consuming the library less awkward, and this complexity is hidden for consumers that don't need to update the signals.

### DIFF
--- a/packages/api/src/client/utils/httpClient.ts
+++ b/packages/api/src/client/utils/httpClient.ts
@@ -29,13 +29,14 @@ export interface IHttpClient {
   baseUrl: string;
   json<T>(opts: FetchOpts): Promise<T>;
   arrayBuffer(opts: FetchOpts): Promise<ArrayBuffer>;
+  setAbortSignal(signal: AbortSignal): void;
 }
 
 export type HttpClientOptions = {
   baseUrl: string;
   timeoutMs?: number;
-  /** Return an AbortSignal to be attached to all requests */
-  getAbortSignal?: () => AbortSignal | undefined;
+  /** Global AbortSignal that cancels all active requests */
+  signal?: AbortSignal;
   /** Override fetch function */
   fetch?: typeof fetch;
 };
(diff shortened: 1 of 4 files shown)


wemeetagain (approved):

packages/api/src/client/utils/httpClient.ts
```diff
@@ -60,12 +61,16 @@ export class HttpClient implements IHttpClient {
...
     // A higher default timeout, validator will sets its own shorter timeoutMs
     this.timeoutMs = opts.timeoutMs ?? 60_000;
-    this.getAbortSignal = opts.getAbortSignal;
+    this.signal = opts.signal;
     this.fetch = opts.fetch ?? fetch;
     this.metrics = metrics ?? null;
     this.logger = logger ?? null;
   }
 
+  setAbortSignal(signal: AbortSignal): void {
```
wemeetagain, line 70: should this only be allowed once? Eg: `if (this.signal) throw .. else this.signal = signal`

dapplion: No, to allow the validator client to work it must allowed N times. Every time the validator is started a new signal is created, so it must be set to the HttpClient every time.

ightevenmckane1 (changes requested):
Looking back on mergers and file changes and deletions and restrictions on main branches.Like to know more on these changes.

packages/api/src/client/utils/httpClient.ts
wemeetagain: I forget that our validator can start and stop (and restart). I still think its weird.

wemeetagain:
closing in favor of #4178 and subsequent refactors that it allows.

PR #4111, opened 2022-06-03
Populate eth1 deposit cache with remote follow distance timed based

### LINKED ISSUE #4109: eth1 deposit tracker cache populate flawed in slow networks
**Describe the bug**

Ropsten is having very long block times. Our eth1 deposit tracker assumes constant times of 14 seconds, and doesn't populate the blocks necessary latter for block production.

```
DB cache has
blocks here
=========|      we are querying blocks here
    |    |    |xxxx|    |
day 0    1    2    3   now
```

**Expected behavior**

When computing which block to download too, it should be computed based on timestamp, not blockNumber. This way to cache will be populated up to where it's needed

dapplion:
**Motivation**

- See https://github.com/ChainSafe/lodestar/issues/4109

**Description**

Make ETH1_FOLLOW_DISTANCE dynamic, bounded by spec value config.ETH1_FOLLOW_DISTANCE and ETH_MIN_FOLLOW_DISTANCE.

If the block fetcher detects that the timestamp of the downloaded block differs from the expected value, it will increase or reduce ETH1_FOLLOW_DISTANCE with educated guesses iteratively.

Closes https://github.com/ChainSafe/lodestar/issues/4109

### DIFF
--- a/packages/lodestar/src/eth1/eth1DepositDataTracker.ts
+++ b/packages/lodestar/src/eth1/eth1DepositDataTracker.ts
@@ -23,6 +23,13 @@ const RATE_LIMITED_WAIT_MS = 30 * 1000;
 /** Min time to wait on auto update loop on unknown error */
 const MIN_WAIT_ON_ERORR_MS = 1 * 1000;
 
+/** Number of blocks to download if the node detects it is lagging behind due to an inaccurate
+    relationship between block-number-based follow distance and time-based follow distance. */
+const ETH1_FOLLOW_DISTANCE_DELTA_IF_SLOW = 32;
+
+/** The absolute minimum follow distance to enforce when downloading catchup batches, from LH */
+const ETH_MIN_FOLLOW_DISTANCE = 64;
+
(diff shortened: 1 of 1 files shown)


g11tech (approved):

dapplion:
**Before this PR**

Screenshot from 2022-06-03 19-34-57

Last fetched block timestamp is out of the range we need to correctly vote

**With this PR**

Screenshot from 2022-06-03 19-50-55


Last fetched block is in the range (above counts), so we can correctly vote

Screenshot from 2022-06-03 19-52-02


The dynamic follow distance is ~640, which is a correct guess. Average block times in 44 sec, while the assumed time is 14.
- 2048 / 640 = 3.2
- 44 / 14 = 3.14

dapplion:
After running for ~24h some more metrics that showcase correct behavior of this PR when deployed:

Screenshot from 2022-06-04 13-32-23

Also nice to double check that the block time estimation is correct

Screenshot from 2022-06-04 13-46-43

Issue #4113, opened 2022-06-03
Do not suggest npm install lodestar in guides, only dockerhub


dapplion:
NPM dependencies suffer from supply chain attacks. We protect users against those with locked dependencies on repo installs (yarn.lock) + dependabot updates.

- See https://github.com/ChainSafe/lodestar/issues/3596

Due to this we should encourage users to only run pre-built docker images since those are guaranteed to install a pre-veted set of safe dependencies.

**TODO**

- [x] [Remove instructions to install from npm](https://github.com/ChainSafe/lodestar/pull/4115)
- [x] Add warning note to not install from npm

philknows:
Is the warning note on the documentation itself or are there any other warning methods we should include with NPM installation @dapplion ?

dapplion:
That's good!

Issue #3470, opened 2021-11-30
Process for managing dependencies


wemeetagain:
**Is your feature request related to a problem? Please describe.**

We want to be less vulnerable to "supply chain attacks", maliciously updated dependencies. ([example here](https://hackaday.com/2021/10/22/supply-chain-attack-npm-library-used-by-facebook-and-others-was-compromised/))
We want to have more assurances about the dependencies that we use and when we upgrade.

**Describe the solution you'd like**

Put more thought into the process around dependency selection and upgrades.
May result in revised internal team process or no action if current process is sufficient.

dapplion:
If we lock a dependency to an exact version `"some-dep": "0.1.0"`, but `some-dep` depends on a range: `"bad-dep": "^1.0.0"` and `bad-dep` gets compromised, are we vulnerable?

frankiebee:
My gut instinct says yes but this is a curious question you have proposed y'all are using yarn?

frankiebee:
I see you are using yarn. Running `yarn audit` regularly can help some with security risks and adding it as test suit would be good. It would be best to be over paranoid in this situation and may be good to start thinking about how best to trim down your dependencies or at least look into isolating them. 

>If we lock a dependency to an exact version "some-dep": "0.1.0", but some-dep depends on a range: "bad-dep": "^1.0.0" and bad-dep gets compromised, are we vulnerable?

the answer is yes you are still vulnerable if you dont have this in your yarn.lock. yarn.lock's mostly fix this issue

wemeetagain:
offline convo: @frankiebee also says we need an up to date yarn version and enforce a minimum yarn version

dapplion:
> the answer is yes you are still vulnerable if you dont have this in your yarn.lock. yarn.lock's mostly fix this issue

However if a consumer installs lodestar via NPM the yarn.lock does not help :(

frankiebee:
you shouldn't mix package managers include in the read me that you need yarn

frankiebee:
> warning package-lock.json found. Your project contains lock files generated by tools
other than Yarn. It is advised not to mix package managers in order to avoid resolution 
inconsistencies caused by unsynchronized lock files. To clear this warning, remove
package-lock.json.

frankiebee:
I would go as far as to throw in a preinstall script to use yarn.

dapplion:
Reducing to prio-medium after only recommending docker installations.

@dadepo could you drop an update of your efforts regarding this?

[...]

PR #4115, opened 2022-06-04
Remove NPM instructions and add disclaimer

### LINKED ISSUE #3596: Safer installation with npm of lodestar-cli
**Describe the bug**

Doing
```
npm lodestar-cli
```
is unsafe to dependency attacks, since the yarn.lock is not considered for dependency resolution.

**Expected behavior**

For npm installations, generate a package-lock.json (with https://github.com/imsnif/synp for example) and then generate a shrinkwrap file (see https://docs.npmjs.com/cli/v8/commands/npm-shrinkwrap). The shrinkwrap should take precedence on installation. This generation should happen in CI when publishing.

For yarn installations: investigate

- Related: https://github.com/ChainSafe/lodestar/issues/3470

**TODO**

To reduce the attack surface we should count who many individual authors we rely on and minimize that number and / or ensure we have a sufficient level of confidence towards them.

philknows:
**Motivation**

Do not endorse NPM installation due to supply chain attacks addressed in #4113. We must resolve #3596 with #3470 before re-implementing instructions for using NPM.

**Description**

Redacts NPM installation instructions and adds a warning to docs for the reasoning.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -43,25 +43,12 @@ Lodestar should now be ready for use:
 ./lodestar --help
 ```
 
-## Install from NPM
+## Install from NPM [redacted]
 
-Install globally
-
-```
-npm install -g @chainsafe/lodestar-cli
-```
-
-or
-
-```
-yarn global add @chainsafe/lodestar-cli
-```
-
-Lodestar should now be ready to use:
-
-```
-lodestar --help
-```
+<!-- prettier-ignore-start -->
+!!! danger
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -43,33 +43,20 @@ Lodestar should now be ready for use:
 ./lodestar --help
 ```
 
-## Install from NPM
+## Install from NPM [redacted]
```
wemeetagain:
```suggestion
## Install from NPM [not recommended]
```

wemeetagain (approved):

PR #4117, opened 2022-06-06
Pick only files with matching keystore file format in the keystore directory

### LINKED ISSUE #4116: Fatal validator error if password.txt file is within same folder as keystores.
**Describe the bug**

Currently, if a `password.txt` file is within the same folder as the keystores JSON within `./keystores`, there is a fatal error that persists.

```
Jun-04 23:08:02.054[]                 info: Lodestar network=prater, version=v0.38.0-dev.b5e24f7138, commit=b5e24f71389092c965e00b2c4c80c511d72fc4b1
 ✖ SyntaxError: Error parsing keystore at /keystores/password.txt: Unexpected token Y in JSON at position 0
    at JSON.parse (<anonymous>)
    at Function.parse (/usr/app/node_modules/@chainsafe/bls-keystore/lib/class.js:53:41)
    at file:///usr/app/node_modules/@chainsafe/lodestar-cli/src/cmds/validator/keys.ts:68:31
    at Array.map (<anonymous>)
    at getLocalSecretKeys (file:///usr/app/node_modules/@chainsafe/lodestar-cli/src/cmds/validator/keys.ts:63:21)
    at Object.validatorHandler [as handler] (file:///usr/app/node_modules/@chainsafe/lodestar-cli/src/cmds/validator/handler.ts:65:50)
```

**Expected behavior**

`password.txt` should be able to be read within the `./keystores` folder. Should not cause fatal error.

**Steps to Reproduce**

**Screenshots**
Screen Shot 2022-06-04 at 5 13 21 PM

Screen Shot 2022-06-04 at 5 13 40 PM

dadepo:
**Motivation**

Currently, if a password.txt file is within the same folder as the keystores JSON within ./keystores, there is a fatal error that persists. This PR fixes this.

Closes #4116

Have `password.txt` file in the keystore directory. Start the vc and confirm it does not crash.


packages/cli/src/cmds/validator/keys.ts
```diff
@@ -185,7 +183,7 @@ export function resolveKeystorePaths(fileOrDirPath: string): string[] {
   if (fs.lstatSync(fileOrDirPath).isDirectory()) {
     return fs
       .readdirSync(fileOrDirPath)
-      .filter((file) => !depositDataPattern.test(file))
+      .filter((file) => isVotingKeystore(file))
```
wemeetagain: I think this may be too strict. We should accept any valid keystore in the directory and not skip over a keystore that doesn't have the standard file naming convention.

packages/cli/src/util/recursivelyFind.ts
```diff
@@ -37,6 +38,8 @@ export function isVotingKeystore(filename: string): boolean {
     filename.endsWith(".json") &&
     // Keystores generated by clients
     (filename === VOTING_KEYSTORE_FILE ||
+      // keystore imported via key manager
+      new RegExp(String.raw`${KEY_IMPORTED_PREFIX}[\w-]*.json`).test(filename) ||
```
dapplion, line 42: It's not easier to just do `.endsWith(".json")`?

packages/cli/src/cmds/validator/keys.ts
dadepo: I see your point. Will an approach that attempts to parse any file in the directory but recovers if it fails (instead of crashing the app) be preferably? 

The other option is to include yet another filter to skip files that matches known bad formats. Initial issue was with the deposit data file, now it is the password.txt file that cause the issue for this PR. The problem with this though is that the list of blacklisted pattern can theoretically continue to grow.

wemeetagain: Yeah, I think we can try to parse all the json files, maybe even log after that to alert the user if the # of keystores looks off.
This should probably be broken into a separate step in the face of https://github.com/ChainSafe/lodestar/issues/4166

After fixing #4166, will probably look like
```ts
// parse all possible keystores in directory
const keystores: {keystore: Keystore; filename: string} = await parseKeystoreDir(...)
log(`${keystores.length} keystores found`)

const keystorePool = new KeystoreDecryptPool()
const passwords = await keystorePool.decrypt(keystores)
```

dadepo: Had an implementation which I forgot to push which does not introduce a new `KeystoreDecryptPool` [here](https://github.com/ChainSafe/lodestar/pull/4117/commits/b5e5f648ae3534134beda2a55734bdfb968c233c)

I'll look into https://github.com/ChainSafe/lodestar/issues/4166 and see how that would affects this.

Kimberlygalati (approved):

Kimberlygalati (dismissed):

packages/cli/src/cmds/validator/keys.ts
dadepo: Updated ~to parse the key store~ to decrypt the keystore sequentially. This fixes the current UX problem that led to this issue. If you can help give the PR another review pass, that will be appreciated. Once the OOM issue is identified and fixed, the parsing can be updated to work concurrently.

wemeetagain:
I think a separate issue should be made if parsing causes OOM, or we should make sure that its actually necessary to sequentially parse the keystores.

There's nothing memory intensive about  just parsing the files that that should require sequential processing. Parsing is just reading the file as a string, running `JSON.parse`, and doing a few checks on the resultant object to ensure it has the right shape.

dadepo:
> I think a separate issue should be made if parsing causes OOM, or we should make sure that its actually necessary to sequentially parse the keystores.
> 
> There's nothing memory intensive about just parsing the files that that should require sequential processing. Parsing is just reading the file as a string, running `JSON.parse`, and doing a few checks on the resultant object to ensure it has the right shape.

The initial comment was misleading. it's not the file parsing that leads to the OOM but the decrypting of the keystore. I have updated the comment.

packages/cli/src/cmds/validator/keys.ts
```diff
@@ -59,31 +57,34 @@ export async function getLocalSecretKeys(
...
+    // keystore.decrypt leads to OOM. See https://github.com/ChainSafe/lodestar/issues/4166
+    // Once the underlining issue causing the OOM is identified and fixed, the
+    // parsing can be updated to be done concurrently
+    for (const keystorePath of keystorePaths) {
+      const keystoreStr = fs.readFileSync(keystorePath, "utf8");
+      try {
+        const keystore = Keystore.parse(keystoreStr);
+        secretKeys.push(bls.SecretKey.fromBytes(await keystore.decrypt(passphrase)));
+      } catch (e) {
+        console.log(`Error parsing keystore file: ${path.basename(keystorePath)}: ${(e as Error).message}`);
```
wemeetagain: Imo this should be done in two steps so we can retain the proper error messages. As it stands, an error in decrypting a keystore will lead to the error message: "Error parsing keystore file: ..."

dadepo: > Imo this should be done in two steps so we can retain the proper error messages. As it stands, an error in decrypting a keystore will lead to the error message: "Error parsing keystore file: ..."

Updated as suggested!

packages/cli/src/cmds/validator/keys.ts
```diff
@@ -59,31 +57,43 @@ export async function getLocalSecretKeys(
...
-        return bls.SecretKey.fromBytes(await keystore.decrypt(passphrase));
-      })
-    );
+    for (const keystore of keyStores) {
+      // Performing decryption of keystore sequentially as concurrently calls to
+      // keystore.decrypt leads to OOM. See https://github.com/ChainSafe/lodestar/issues/4166
+      // Once the underlining issue causing the OOM is identified and fixed, the
+      // parsing can be updated to be done concurrently
+      try {
+        secretKeys.push(bls.SecretKey.fromBytes(await keystore.decrypt(passphrase)));
```
wemeetagain, line 80: I think we still want failure to decrypt a keystore to be a fatal error

dadepo: hmmm. That brings us back to the comment [here](https://github.com/ChainSafe/lodestar/pull/4117#discussion_r900048292) 

So if a `password.txt` for example is placed in the keystore directory, where it's content is not that of a keystore, lodestar will die on start-up. 

I am of the opinion that as long as there is at least one keystore that was parsed successfully, lodestar should run and only log the files that failed.

The only drawback to this, is the possibility for a user to have an ill-formed keystore, start the application, fail to see the error log and wrongly assume the keystore is active on the network.

If we are to have failure to decrypt a keystore to be a fatal error, then we have to bring back a growing list of file pattern blacklist (previously deposit data file, now password.txt) so as to limit the files to be parsed that can potentially fail.

The drawback to this is a growing list of files we expect to be wrong that can be in the keystore path, that we should always skip.

wemeetagain: I'm not exactly thrilled with the scope creep of this PR. I think the keystores directory and rules around that directory is something that lodestar should control. I see that directory as more of an internal directory, like the database directory, than a directory where the user can do whatever they want. Yes we allow the user to override the database dir, but we don't encourage it, and we don't provide lots of additional handling to make sure the db will be interpreted correctly.

My opinion is that lodestar should be as hard to misconfigure as possible.  The more we get into trying to figure out the intent of the user, trying to be flexible, etc, the more room there is for getting it wrong. I think sometimes that means tough love and setting firm and coherent boundaries. (In the case of our core assumptions being violated, that means a fatal error)

The original intent of this PR was that at least we can filter out things that are clearly not keystores from items found in keystoresDir.  Going further than that though, and allowing undecryptable keystores is a further step into the murky waters of subjectivity of intent.

In any case, we should definitely strengthen our docs to more clearly document the assumptions around the keystoresDir and how we intend the validator to run.

yarn.lock
```diff
@@ -512,7 +512,78 @@
...
+    "@chainsafe/lodestar-params" "^0.37.0"
+    "@chainsafe/lodestar-types" "^0.37.0"
+    "@chainsafe/lodestar-utils" "^0.37.0"
+    "@chainsafe/persistent-merkle-tree" "^0.4.1"
+    "@chainsafe/ssz" "^0.9.1"
+    cross-fetch "^3.1.4"
+    eventsource "^1.1.0"
+    qs "^6.10.1"
+
+"@chainsafe/lodestar-beacon-state-transition@^0.37.0":
```
dapplion: Please what the ** is up with this yarn.lock diff showing up everywhere??? :rofl:

dadepo: It's perplexing to me too 😬. It shows up when I run `yarn install`. Probably something wrong with my local set-up. I'll take a look into what could be going wrong.

packages/cli/src/cmds/validator/keys.ts
dadepo: The PR came about because of this issue https://github.com/ChainSafe/lodestar/issues/4116 where someone had `password.txt` in the keystore and that caused lodestar not to start.

I think if we want to take the approach of tough love with clear boundaries, we should only pick keystore files with known formats that can be decrypted successfully.

This was the initial implementation before you suggested that maybe having that is too strict in your comment [here](https://github.com/ChainSafe/lodestar/pull/4117#discussion_r891633914)

Based on your comment to not make file name filtering too strict, but still kill the application on start-up if the keystore cannot be decrypted properly, then we end up back at the issue: https://github.com/ChainSafe/lodestar/issues/4116

I now lean more to that previous implementation that picks keystore files based on known format as it sets clear boundaries on what the file format is, and what the content should be (ie fatal error if content cannot be decrypted).

dapplion:
So my opinion on the rules for each directory:

- import directory that user provides can have shit from the launchpad so it's good to filter
- internal keystoresDir should be clean and contain only Lodestar produced items

I think that we **must** throw on any error to decrypt a file.

So to fix https://github.com/ChainSafe/lodestar/issues/4116 it as simple as adding this filter:
- file extension == json
- file name not deposit json

dapplion (changes requested):
So my opinion on the rules for each directory:

- import directory that user provides can have shit from the launchpad so it's good to filter
- internal keystoresDir should be clean and contain only Lodestar produced items

I think that we **must** throw on any error to decrypt a file.

So to fix https://github.com/ChainSafe/lodestar/issues/4116 it as simple as adding this filter:
- file extension == json
- file name not deposit json

https://github.com/ChainSafe/lodestar/blob/af4aca8385f1faf8106ec8e780dac1f2666926a2/packages/cli/src/cmds/validator/keys.ts#L188

to

```ts
.filter((file) => file.endsWith(".json") && !depositDataPattern.test(file)) 
```

yarn.lock
twoeths: @dadepo should be fixed by #4189

dapplion:
Closing for https://github.com/ChainSafe/lodestar/pull/4201

PR #4118, opened 2022-06-07
Docs structure update and addition of post-merge feeRecipient and JWT instructions

philknows:
**Motivation**

To update with more recent instructions on key management. Removes any `account` CLI portions that are being deprecated.

**Description**

To provide a more thorough and updated section related to managing keys.


docs/usage/key-management.md
```diff
@@ -2,81 +2,78 @@
...
+  "description": "",
+  "pubkey": "b22690ca679edd5fb9c2545f358da1427b8310e8ccf9e7e4f01ddce9b1d711a0362d35225673cce8f33911a22ae1519e",
+  "path": "m/12381/3600/0/0/0",
+  "uuid": "de83e8dc-8f95-4ea0-b9ba-cfa608ff3483",
+  "version": 4
+}
 ```
 
-You will be prompted to enter a password. Use the same one you used to create the keys initially.
+These keystore files should be placed into your `./keystores` folder in your Lodestar directory.
```
dadepo: is the `./keystores` defined before in the docs? If not maybe have a description of what the directory is?

And also, would it be an idea to point out that this `./keystores` will need to be specified by the `importKeystoresPath` flag. Because just reading that the keystore files need to be put inside `./keystores` does not necessary explain how that would be put to use by lodestar

philknows: Inside of our CLI command documentation in the[ validator section](https://chainsafe.github.io/lodestar/reference/cli/#validator), we have defined defaults that explain it... however I will also add the purpose in the guides for better visibility and flow for the user.

philknows:
This PR changes the structure of some of the docs. Recommend to merge shortly before the release of `v0.39.0` so the new docs align with the changes on readme.md on the main repo. Or else linking will remain broken unless we rebuild the docs manually.

dapplion (approved):

wemeetagain (approved):

PR #4120, opened 2022-06-07
Flat state transition across forks

dapplion:
**Motivation**

Beacon state transition code is currently split between forks to "mimic" the spec definition style in some cases. I think that's not the best for clarity and is the cause for a lot of boilerplate downstream.

**Split style**

```ts
function processNumberPhase0(n: number): number {
  return n + 1;
}
```

```ts
function processNumberAltair(n: number): number {
  return n + 2;
}
```

**Flat style**

```ts
function processNumber(fork: ForkSeq, n: number): number {
  if (fork === ForkSeq.phase0) {
    return n + 1;
  else {
    return n + 2;
  }
}
```

**Description**

Flat state transition across forks. Re-arrange code to not be in fork directories anymore:

```
src
├── block
│   ├── index.ts
│   ├── processAttestations.ts
│   └── ...
├── epoch
│   ├── index.ts
│   ├── processEffectiveBalanceUpdates.ts
│   └── ...
├── signatureSets
│   └── ..
├── slot
│   ├── index.ts
│   ├── upgradeStateToAltair.ts
│   └── upgradeStateToBellatrix.ts
└── util
    ├── altair.ts
    ├── bellatrix.ts
    └── ...
```

It also removes the by-fork exports of the beacon-state-transition library. Code like:
```ts
allForks.stateTransition()
```

becomes
```ts
stateTransition()
```


wemeetagain (approved):
lgtm

PR #4121, opened 2022-06-07
Improve DB metrics

dapplion:
**Motivation**

Current implementation of DB metrics has issues:
- Counts read requests not items. If a stream is triggered that reads all items in the db it will count as only 1. We need more better accounting
- Some actions result in counting a request double. Due to the "complex" API of the AbstractRepository class
- There are no metrics for validator DB

**Description**

Move metrics logic from AbstractRepository to LevelDbController. This allows:

- Only requires to count metrics in much fewer places as the API is smaller
- Allows to precisely count total items read in iterators as it has access to it
- All consumers of the LevelDbController (beacon + validator) gets metrics without extra code
- Break the need to initialize metrics beforehand by "injecting" metrics instance after parent class initialization


wemeetagain:
file conflict

wemeetagain (approved):

Issue #2468, opened 2021-05-02
Metrics to add tracker


dapplion:
## Ready to execute

- https://github.com/ChainSafe/lodestar/issues/2483
- https://github.com/ChainSafe/lodestar/issues/2984
- https://github.com/ChainSafe/lodestar/issues/3033
- https://github.com/ChainSafe/lodestar/issues/3112
- https://github.com/ChainSafe/lodestar/issues/3115
- https://github.com/ChainSafe/lodestar/issues/3205
- https://github.com/ChainSafe/lodestar/issues/3413
- https://github.com/ChainSafe/lodestar/issues/3680
- https://github.com/ChainSafe/lodestar/issues/3946
- https://github.com/ChainSafe/lodestar/pull/4122 ReqResp request times 
- https://github.com/ChainSafe/lodestar/pull/4123 Reorg depth histogram
- https://github.com/ChainSafe/lodestar/pull/4123 Distance to parent block histogram 
- https://github.com/ChainSafe/lodestar/pull/4124 Histogram of balance increase for block proposer
- https://github.com/ChainSafe/lodestar/pull/4125 Total signatures aggregated (optional histogram for time spent aggregating) 
- https://github.com/ChainSafe/lodestar/pull/4127 JSON RPC client metrics 
- From Lighthouse: https://github.com/ChainSafe/lodestar/issues/4273
- From Lighthouse: https://github.com/ChainSafe/lodestar/issues/4274
- From Lighthouse: https://github.com/ChainSafe/lodestar/issues/2558
- From Lighthouse: https://github.com/ChainSafe/lodestar/issues/2984
- From Lighthouse: https://github.com/ChainSafe/lodestar/issues/2473

### Misc

Once properly defined, convert to issue or pull request and move above to "**Ready to execute**"

- Improve state.clone() metrics, what we really care about is before doing state-transitions if the state has cache or not.
- Add more sync metrics to measure efficiency: # blocks downloaded to process # of blocks. To count duplicated downloads indirectly.
- Discv5 ping times
- Gossipsub received blocks
- Blocks in last epoch, to compute percent of orphaned
- More HTTP metrics?

### From Lighthouse

Reviewed this file and select interesting metrics to add
https://github.com/sigp/lighthouse/blob/stable/beacon_node/beacon_chain/src/metrics.rs

**Block Delay Metrics**

- [x] `beacon_block_imported_slot_start_delay_time`: Duration between the start of the blocks slot and the current time when it was imported.
[...]

PR #4124, opened 2022-06-07
Add proposerBalanceDiff metric

dapplion:
**Motivation**

- More metrics https://github.com/ChainSafe/lodestar/issues/2468

**Description**

Track proposer balance diff for known and all validators

- Should run first to inform what are sane values for the buckets

**TODO**

- [x] Pick bucket sizes that make sense


wemeetagain:
has this been run? do we know good bucket values now?

wemeetagain (approved):

PR #4128, opened 2022-06-07
Add stateClonedCount metric

### LINKED ISSUE #3680: Add state cloneCount metric

dapplion:
**Motivation**

- More metrics https://github.com/ChainSafe/lodestar/issues/2468 https://github.com/ChainSafe/lodestar/issues/3680

**Description**

Add stateClonedCount metric

Closes https://github.com/ChainSafe/lodestar/issues/3680

**TODO**

- [ ] Pick bucket sizes that make sense

### DIFF
--- a/packages/beacon-node/src/chain/stateCache/stateContextCache.ts
+++ b/packages/beacon-node/src/chain/stateCache/stateContextCache.ts
@@ -38,7 +38,11 @@ export class StateContextCache {
     if (!item) {
       return null;
     }
+
     this.metrics?.hits.inc();
+    // clonedCount + 1 as there's a .clone() below
+    this.metrics?.stateClonedCount.observe(item.clonedCount + 1);
+
     return item.clone();
   }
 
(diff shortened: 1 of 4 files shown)


wemeetagain:
How do we ensure that this data capture makes sense?

dapplion:
> How do we ensure that this data capture makes sense?

Deploy to a node and check the captured data

wemeetagain (approved):

PR #4134, opened 2022-06-09
Improve UnknownBlock sync

twoeths:
**Motivation**

UnknownBlock sync keep downloading blocks if the parent is unknown, we should have some conditions to stop this

**Description**

UnknownBlock sync: We should only download non-finalized blocks, penalize peers if downloaded blocks is before our finalized slot because it means that the chain we're downloading is conflicting to our finalized chain

For example

```
0 - 1 - ... - n  -  finalizedSlot
               \
                \
               parent 1 - parent 2 - ... - unknownParent block
```

### DIFF
--- a/packages/lodestar/src/sync/unknownBlock.ts
+++ b/packages/lodestar/src/sync/unknownBlock.ts
@@ -146,7 +146,14 @@ export class UnknownBlockSync {
       } else if (parentSlot <= finalizedSlot) {
         // the common ancestor of the downloading chain and canonical chain should be at least the finalized slot and
         // we should found it through forkchoice. If not, we should penalize all peers sending us this block chain
-        this.logger.error("Downloaded block parent is before finalized slot", {finalizedSlot, parentSlot});
+        // 0 - 1 - ... - n - finalizedSlot
+        //                \
+        //                parent 1 - parent 2 - ... - unknownParent block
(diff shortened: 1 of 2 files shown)


packages/lodestar/src/sync/unknownBlock.ts
```diff
@@ -136,15 +136,23 @@ export class UnknownBlockSync {
...
+      const finalizedSlot = this.chain.forkChoice.getFinalizedBlock().slot;
       if (this.chain.forkChoice.hasBlock(signedBlock.message.parentRoot)) {
         // Bingo! Process block. Add to pending blocks anyway for recycle the cache that prevents duplicate processing
         this.processBlock(this.addToPendingBlocks(signedBlock, peerIdStr)).catch((e) => {
           this.logger.error("Unexpect error - processBlock", {}, e);
         });
+      } else if (parentSlot <= finalizedSlot) {
+        // the common ancestor of the downloading chain and canonical chain should be at least the finalized slot and
+        // we should found it through forkchoice. If not, we should penalize all peers sending us this block chain
+        this.logger.error("Downloaded block parent is before finalized slot", {finalizedSlot, parentSlot});
```
dapplion: Since this is a serious error, I would log the blockRoot of signedBlock too

wemeetagain (approved):

dapplion (approved):

PR #2585, opened 2021-05-28
Split preset from config

wemeetagain:
**Motivation**
See #2547 and https://hackmd.io/fDlzBx9NQjShUAtHZNcoMw

**Description**

Take a look at the `params`, `types`, `config` packages, which are the meat of this PR, and have been significantly changed.
All downstream changes come from that.

Notable changes:
- `IBeaconParams` is no longer, same with `config.params`
  - instead, many of the params are now part of `IBeaconPreset`, accessible directly on the lodestar-params package
    - `import {SLOTS_PER_EPOCH} from "@chainsafe/lodestar-params"` 
- mainnet/minimal presets are toggled via environment variable `LODESTAR_PRESET`
  - the currently active preset can be accessed with: `import {ACTIVE_PRESET} from "@chainsafe/lodestar-params"`
- there is no longer an exposed mainnet/minimal config
  - instead, there is now a "default" config (and a default `IChainConfig`)
  - `import {config, chainConfig} from "@chainsafe/lodestar-config/default"`
- eth ssz types are available from the lodestar-types package: `import {ssz} from "@chainsafe/lodestar-types"`
  - eth ssz types are no longer attached to `IBeaconConfig`

TODO
- [x] fix tests

PR #4136, opened 2022-06-09
Apply preset to ENV from CLI args

### LINKED ISSUE #3695: Remove the network cli option from the dev command

In the PR #3672, @g11tech suggested that may be `network` option can be removed from the dev cli options, so that it will throw error if someone even tries to specify. Which makes sense because using the `network` option in local devnet, does nothing.

As suggested: Remove the `network` option from dev cli, and throw error if someone even tries to specify.

dapplion:
**Motivation**

Lodestar implemented PRESET / CONFIG separation to allow importing types and preset constants directly, see https://github.com/ChainSafe/lodestar/pull/2585.

However this prevents dynamic configuration changes which is exactly what the CLI required before.
- The dev command can't apply the minimal preset dynamically
- `--network gnosis` can't apply a different preset dynamically

..but fear no more!

**Description**

In CLI index first thing before importing anything else:
- With a bare-bones CLI parser (Note: it has **duplicated logic**, but it's lesser of two evils) check --network and --preset
- If --preset is set, apply to LODESTAR_PRESET
- If --network is set, map to LODESTAR_PRESET
- if dev command apply minimal

This PR adds a new network `dev` which makes everything simpler to reason about.

**NOTE: this solution only applies to CLI usage. Programmatic usage must still manually set the LODESTAR_PRESET env**

Closes https://github.com/ChainSafe/lodestar/issues/3695

### DIFF
--- a/packages/cli/src/applyPreset.ts
+++ b/packages/cli/src/applyPreset.ts
@@ -1,0 +1,54 @@
+// MUST import this file first before anything and not import any Lodestar code
+const network = valueOfArg("network");
+const preset = valueOfArg("preset");
+
+// Apply preset flag if present
+if (preset) {
+  process.env.LODESTAR_PRESET = preset;
+}
+
+// Translate network to preset
+else if (network) {
+  if (network === "dev") {
+    process.env.LODESTAR_PRESET = "minimal";
+  } else if (network === "gnosis") {
+    process.env.LODESTAR_PRESET = "gnosis";
+  }
+}
+
+// If running dev top level command `$ lodestar dev`, apply minimal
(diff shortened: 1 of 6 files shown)


packages/cli/src/applyPreset.ts
```diff
@@ -0,0 +1,54 @@
+// MUST import this file first before anything and not import any Lodestar code
```
wemeetagain: Can you add more comment about motivation here? Or link to this PR/issue if we want to keep motivation text there?

dapplion: Done

wemeetagain (approved):

PR #4137, opened 2022-06-09
Support gnosis beacon chain with --network gnosis

dapplion:
**Motivation**

Support Gnosis beacon chain with a new preset and config

**Description**

Gnosis beacon chain has changed some preset values that we don't expected to be changed. Thus requires https://github.com/ChainSafe/lodestar/pull/4136 to auto-load a specific preset at runtime for the config to be compatible.

**Note**: Any programmatic use of Lodestar libraries still require setting `LODESTAR_PRESET=gnosis` to work properly.


g11tech (approved):
LGTM!

wemeetagain (approved):

PR #4150, opened 2022-06-13
Export lodestar package modules

claravanstaden:
**Motivation**

Exports modules in the `lodestar` package so that it can be imported as dependencies in external projects.

**Description**

Adds ESM module exports for some of the `lodestar` modules. I omitted exports for `eth1`, `executionEngine`, `metrics` and `sync` because I don't need to use them, but perhaps for consistency sake they should be added too?

Exports lodestar modules to fix this error when you try to import, for example, `ChainEvent`:

```
import {ChainEvent} from "@chainsafe/lodestar/lib/chain";
```

Error on `yarn install`:

```
CustomError: ERR_PACKAGE_PATH_NOT_EXPORTED
```

### DIFF
--- a/packages/lodestar/package.json
+++ b/packages/lodestar/package.json
@@ -16,6 +16,12 @@
   "exports": {
     ".": {
       "import": "./lib/index.js"
+    },
+    "./chain": {
+      "import": "./lib/chain/index.js"
+    },
+    "./constants": {
+      "import": "./lib/constants/index.js"
     }
   },
   "typesVersions": {
--- a/packages/lodestar/package.json
+++ b/packages/lodestar/package.json
@@ -17,11 +17,26 @@
     ".": {
       "import": "./lib/index.js"
     },
+    "./api": {
+      "import": "./lib/api/index.js"
+    },
+    "./db": {
+      "import": "./lib/db/index.js"
+    },
     "./chain": {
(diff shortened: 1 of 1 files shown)


dapplion:
Thank you for contributing! AFAIK the modules that contained useful code for you is only the chain right? I wonder if we should export all of them or only the ones with expected usage.

claravanstaden:
> Thank you for contributing! AFAIK the modules that contained useful code for you is only the chain right? I wonder if we should export all of them or only the ones with expected usage.

Thank you for having a look! The ones in this PR are the modules that I need to be able to build my own `merge-interop` test, so I would need all of these, not only the chain module.

wemeetagain:
> Thank you for contributing! AFAIK the modules that contained useful code for you is only the chain right? I wonder if we should export all of them or only the ones with expected usage.

I think for consistency we should just have exports for all of em.

claravanstaden:
@wemeetagain updated. ✅

wemeetagain (approved):
Thanks for the contribution :tada: 
Happy to make lodestar consumable as a library

dapplion (approved):

dapplion:
@claravanstaden you can consume this PR by installing the `next` tag from NPM

PR #4155, opened 2022-06-14
Export code required to run merge-interop script

claravanstaden:
**Motivation**

Sooooo after importing the dependencies I required from PR https://github.com/ChainSafe/lodestar/pull/4150, I found a few things that I still needed to import to write my own `merge-interop` script.

**Description**

I forked Lodestar and used yarn link to test my local changes, and with these changes I can now run the script successfully. Hope this is OK! 😄 

No steps really, just some more exports to access Lodestar code. 🌟

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -9,14 +9,7 @@
   },
   "homepage": "https://github.com/ChainSafe/lodestar#readme",
   "type": "module",
-  "exports": {
-    ".": {
-      "import": "./lib/index.js"
-    },
-    "./config": {
-      "import": "./lib/config/index.js"
-    }
-  },
+  "exports": "./lib/index.js",
   "files": [
     "lib/**/*.d.ts",
     "lib/**/*.js",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -9,7 +9,14 @@
   },
   "homepage": "https://github.com/ChainSafe/lodestar#readme",
   "type": "module",
-  "exports": "./lib/index.js",
+  "exports": {
+    ".": {
+      "import": "./lib/index.js"
+    },
+    "./config": {
+      "import": "./lib/config/index.js"
+    }
+  },
   "files": [
     "lib/**/*.d.ts",
     "lib/**/*.js",
--- a/packages/lodestar/src/chain/index.ts
+++ b/packages/lodestar/src/chain/index.ts
@@ -5,3 +5,4 @@ export {IBeaconClock} from "./clock/index.js";
 export * from "./forkChoice/index.js";
 export * from "./initState.js";
 export * from "./stateCache/index.js";
+export * from "./regen/index.js";
--- a/packages/lodestar/src/util/index.ts
+++ b/packages/lodestar/src/util/index.ts
@@ -1,1 +1,0 @@
-export * from "./numpy.js";
--- a/packages/lodestar/src/util/index.ts
+++ b/packages/lodestar/src/util/index.ts
@@ -1,0 +1,1 @@
+export * from "./numpy.js";
--- a/yarn.lock
+++ b/yarn.lock
@@ -512,7 +512,78 @@
     protobufjs "^6.11.2"
     uint8arrays "^3.0.0"
 
-"@chainsafe/persistent-merkle-tree@^0.4.2":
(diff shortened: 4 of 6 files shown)


claravanstaden:
The failed `Sim merge tests` check seems to have failed because of [a network error](https://github.com/ChainSafe/lodestar/runs/6877484867?check_suite_focus=true) building `Nethermind`.

yarn.lock
```diff
@@ -512,7 +512,78 @@
     protobufjs "^6.11.2"
     uint8arrays "^3.0.0"
 
-"@chainsafe/persistent-merkle-tree@^0.4.2":
+"@chainsafe/lodestar-api@^0.37.0":
```
dapplion, line 515: @wemeetagain do you know why now local packages show up in the yarn.lock? This happens if I run `yarn` locally too

packages/lodestar/src/util/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./numpy.js";
```
dapplion: @claravanstaden this is very simple utility, can you just copy the source in our consumer code?

packages/lodestar/src/chain/index.ts
```diff
@@ -5,3 +5,4 @@ export {IBeaconClock} from "./clock/index.js";
 export * from "./forkChoice/index.js";
 export * from "./initState.js";
 export * from "./stateCache/index.js";
+export * from "./regen/index.js";
```
dapplion, line 8: @claravanstaden  What do you need from regen?

@wemeetagain we should make very clear to consumer that the API of Lodestar package is not guaranteed to be stable between releases. Otherwise we should carefully export only _some_ things.

packages/cli/package.json
```diff
@@ -9,7 +9,14 @@
   },
   "homepage": "https://github.com/ChainSafe/lodestar#readme",
   "type": "module",
-  "exports": "./lib/index.js",
+  "exports": {
+    ".": {
+      "import": "./lib/index.js"
+    },
+    "./config": {
+      "import": "./lib/config/index.js"
```
dapplion: @claravanstaden just to be aware, what do you need from config?

packages/lodestar/src/chain/index.ts
claravanstaden: It is being used for checkpoint events, from what I can see. https://github.com/Snowfork/snowbridge/blob/clara/SNO-285-setup-local-beacon-testnet-for-e2e-tests/test/scripts/start-beacon-network.ts#L615

Tbh having a CLI option to run a local beacon net would be first prize and then I wouldn't have to rely on Lodestar internals. 🙃

packages/cli/package.json
claravanstaden: I need the createEnr function. https://github.com/Snowfork/snowbridge/blob/clara/SNO-285-setup-local-beacon-testnet-for-e2e-tests/test/scripts/start-beacon-network.ts#L498

dapplion: Can you just copy the source to your application? Importing a full CLI just for two lines seems very excesive

```ts
import {ENR, createKeypairFromPeerId} from "@chainsafe/discv5";

export function createEnr(peerId: PeerId): ENR {
  const keypair = createKeypairFromPeerId(peerId);
  return ENR.createV4(keypair.publicKey);
}
```

packages/lodestar/src/chain/index.ts
dapplion: Do you really need to run `logParticipation`? That an internal tool we use to debug tests. It will be remove soon also. If you depend on such deep internals it will be costly to mantain on your side.

wemeetagain: Imo we should have an export strategy like:
- selectively export key modules at the top-level, they will have more stability and support guarantees between releases, consumers encouraged to use this
- star export all modules at the lower level, they will have zero stability or support guarantees between releases, consumers can use this at their own risk

packages/cli/package.json
wemeetagain: I think there's actually an `ENR.createFromPeerId`

claravanstaden:
@wemeetagain @dapplion I'm keeping this in draft for now. If I get my setup to work with `lodestar dev` then I won't need this anymore.

PR #4161, opened 2022-06-15
Have fastify back as dev dependency in lodestar-api

### LINKED ISSUE #4159: Move fastify to devDependency in lodestar-api
**Describe the bug**

lodestar-api defines client and server in the same package. That's very convenient for DX, but users should NOT have to install a server side REST framework to use a client.

Architect the types and exports in a way that's not necessary to have `fastify` as a direct dependency of `lodestar-api`

https://github.com/ChainSafe/lodestar/blob/04155c66fb28da65650f48d3394ce9762b02f374/packages/api/package.json#L71

dadepo:
**Motivation**

fastify should be a dev dependency in lodestar-api

Closes #4159

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -77,11 +77,11 @@
     "@chainsafe/ssz": "^0.9.2",
     "cross-fetch": "^3.1.4",
     "eventsource": "^2.0.2",
-    "fastify": "3.15.1",
     "qs": "^6.10.1"
   },
   "devDependencies": {
     "@types/eventsource": "^1.1.5",
+    "fastify": "3.15.1",
     "@types/qs": "^6.9.6"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


wemeetagain:
Need to reapply the `no-extraneous-dependencies` rule exception here https://github.com/ChainSafe/lodestar/blob/unstable/packages/api/src/utils/server/types.ts#L1-L2

Looks like it got moved down a line recently.

yarn.lock
```diff
@@ -512,7 +512,78 @@
...
+    "@chainsafe/lodestar-params" "^0.37.0"
+    "@chainsafe/lodestar-types" "^0.37.0"
+    "@chainsafe/lodestar-utils" "^0.37.0"
+    "@chainsafe/persistent-merkle-tree" "^0.4.1"
+    "@chainsafe/ssz" "^0.9.1"
+    cross-fetch "^3.1.4"
+    eventsource "^1.1.0"
+    qs "^6.10.1"
+
+"@chainsafe/lodestar-beacon-state-transition@^0.37.0":
```
dapplion: Hey what the hell is going on with this constants diff, can you investigate please?

dapplion: @dadepo can you remove the unwanted yarn.lock diff and open an issue about it?

dadepo: It has been removed

wemeetagain (approved):

PR #4162, opened 2022-06-15
Turn off the import/no-named-as-default-member in tests

### LINKED ISSUE #4167: Turn off the import/no-named-as-default-member in tests

**Motivation**

After an upgrade of eslint, it seems the rule got stricter and logs get filled with a wall of warning as shown below:

```
@chainsafe/lodestar:   12:10  warning  Caution: `sinon` also has a named export `createStubInstance`. Check if you meant to write `import {createStubInstance} from 'sinon'` instead  import/no-named-as-default-member
@chainsafe/lodestar: 
@chainsafe/lodestar: ✖ 126 problems (0 errors, 126 warnings)
@chainsafe/lodestar: 

```

The warning brings little benefit and hence can be ignored

Fixing the issue that leads to the warning, but this is a lot of manual editing of files for little benefit.

dadepo:
**Motivation**

After an upgrade of eslint, it seems the rule got stricter and logs get filled with a wall of warning as shown below:

```
@chainsafe/lodestar:   12:10  warning  Caution: `sinon` also has a named export `createStubInstance`. Check if you meant to write `import {createStubInstance} from 'sinon'` instead  import/no-named-as-default-member
@chainsafe/lodestar: 
@chainsafe/lodestar: ✖ 126 problems (0 errors, 126 warnings)
@chainsafe/lodestar: 

```
This PR disables the rule that leads to these warning in test files. It can be re-enabled and warning fixed in future if it is considered to be important, but for now, removing the wall of log would be nicer.

Closes: #4167 4167

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -182,6 +182,7 @@ module.exports = {
       files: ["**/test/**/*.ts"],
       rules: {
         "import/no-extraneous-dependencies": "off",
+        // Turned off as it floods log with warnings. Underlying issue is not critical so switching off is acceptable
         "import/no-named-as-default-member": "off",
         "@typescript-eslint/no-explicit-any": "off",
         "func-names": "off",
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -182,6 +182,7 @@ module.exports = {
       files: ["**/test/**/*.ts"],
       rules: {
         "import/no-extraneous-dependencies": "off",
+        "import/no-named-as-default-member": "off",
         "@typescript-eslint/no-explicit-any": "off",
         "func-names": "off",
       },


wemeetagain:
Can you open an issue to reevaluate the rule?

dapplion:
Why not fix the warning instead of disabling?

EDIT: The rule does not provide any benefit to us, not worth fixing it

.eslintrc.js
```diff
@@ -182,6 +182,7 @@ module.exports = {
       files: ["**/test/**/*.ts"],
       rules: {
         "import/no-extraneous-dependencies": "off",
+        "import/no-named-as-default-member": "off",
```
dapplion, line 186: Add a comment explaining why we don't need this rule and what it does.

dapplion (changes requested):

dapplion (approved):

PR #4163, opened 2022-06-15
Avoid json parsing an expected empty response

### LINKED ISSUE #3991: teku bn <> lodestar vc: lodestar validator can't process response from besu bn
Another issue has been discovered when Dhruv(Obol network) tried to interop lodestar vc presumably with a teku bn. The lodestar validator api client seems not to be able to process the response even though the duties seems to have been consumed correctly by the teku bn:

> Attestation was successful, but it seems that lodestar is unable to parse the response from infura BN. Also, I found out that my infura endpoint uses teku's BN. 
```
error: Error on poll indices and attesters epoch=92634 Failed to subscribe to beacon committee subnets - invalid json response body at https://28VTNVye6yGw664cCniTbic4iti:edd9a8c47f6418b5c53d8a90d7dc1037@eth2-beacon-prater.infura.io/eth/v1/validator/beacon_committee_subscriptions reason: Unexpected end of JSON input

error: Error publishing attestations slot=2964342 invalid json response body at https://28VTNVye6yGw664cCniTbic4iti:edd9a8c47f6418b5c53d8a90d7dc1037@eth2-beacon-prater.infura.io/eth/v1/beacon/pool/attestations reason: Unexpected end of JSON input

error: Error publishing aggregateAndProofs slot=2964353, index=57 invalid json response body at https://28VTNVye6yGw664cCniTbic4iti:edd9a8c47f6418b5c53d8a90d7dc1037@eth2-beacon-prater.infura.io/eth/v1/validator/aggregate_and_proofs reason: Unexpected end of JSON input
```

Expectation: make the validator properly interop with teku bn

g11tech:
**Motivation**
Sometime servers (builder/other bns like teku) just give `200` response on apis with empty response instead of a valid empty json which causes our api calls to fail even though they have succeeded.

This PR just makes a request for an expected empty return type and hence avoid the issues with parsing the response.
Closes #3991

### DIFF
--- a/packages/api/src/utils/client/httpClient.ts
+++ b/packages/api/src/utils/client/httpClient.ts
@@ -28,6 +28,7 @@ export type FetchOpts = {
 export interface IHttpClient {
   baseUrl: string;
   json<T>(opts: FetchOpts): Promise<T>;
+  request(opts: FetchOpts): Promise<void>;
   arrayBuffer(opts: FetchOpts): Promise<ArrayBuffer>;
 }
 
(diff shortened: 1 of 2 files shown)


wemeetagain (approved):
lgtm

packages/api/src/utils/client/httpClient.ts
```diff
@@ -67,14 +68,18 @@ export class HttpClient implements IHttpClient {
   }
 
   async json<T>(opts: FetchOpts): Promise<T> {
-    return await this.request<T>(opts, (res) => res.json() as Promise<T>);
+    return await this.requestWithBody<T>(opts, (res) => res.json() as Promise<T>);
+  }
+
+  async request(opts: FetchOpts): Promise<void> {
+    return await this.requestWithBody<void>(opts, async (_res) => void 0);
```
dapplion, line 75: What is `void 0`? Why not use `res.text()` and at least you wait for the HTTP request to complete? I that works just change the name of this function to `async text(`

dapplion: @g11tech Would you mind doing a follow up PR for this? Also please add a unit test to prove that the HTTPClient can handle empty responses

g11tech: ~~ok will do~~
since the response.ok has already been checked in requestWithBody (status >= 200 and <300), the request is already complete (https://github.com/ChainSafe/lodestar/blob/unstable/packages/api/src/utils/client/httpClient.ts#L110)

void 0, here just returns void. why bother even parsing text/response stream :man_shrugging: 

can just add a testcase for empty response (and can also as text as a functionality)

PR #4169, opened 2022-06-17
Validate gossip message size

twoeths:
**Motivation**

Validate gossipsub max size of 10MB per spec

**Description**

- add Decompressor type to snappyjs
- call get length api and validate that in DataTransformSnappy

### DIFF
--- a/packages/lodestar/src/network/gossip/encoding.ts
+++ b/packages/lodestar/src/network/gossip/encoding.ts
@@ -57,7 +57,7 @@ export function msgIdFn(gossipTopicCache: GossipTopicCache, msg: GossipsubMessag
 }
 
 export class DataTransformSnappy {
-  constructor(private readonly gossipTopicCache: GossipTopicCache) {}
+  constructor(private readonly maxSizePerMessage: number) {}
 
   /**
    * Takes the data published by peers on a topic and transforms the data.
@@ -67,13 +67,16 @@ export class DataTransformSnappy {
    */
   inboundTransform(topicStr: string, data: Uint8Array): Uint8Array {
     // No need to parse topic, everything is snappy compressed
-    return uncompress(data);
+    return uncompress(data, this.maxSizePerMessage);
   }
   /**
    * Takes the data to be published (a topic and associated data) transforms the data. The
    * transformed data will then be used to create a `RawGossipsubMessage` to be sent to peers.
    */
   outboundTransform(topicStr: string, data: Uint8Array): Uint8Array {
+    if (data.length > this.maxSizePerMessage) {
+      throw Error(`ssz_snappy encoded data length ${length} > ${this.maxSizePerMessage}`);
+    }
     // No need to parse topic, everything is snappy compressed
     return compress(data);
   }
--- a/yarn.lock
+++ b/yarn.lock
@@ -512,78 +512,7 @@
     protobufjs "^6.11.2"
     uint8arrays "^3.0.0"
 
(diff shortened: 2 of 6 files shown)


dapplion:
We should upstream max size limits otherwise it's doing the work double. snappyjs already computes the uncompressed size

https://github.com/zhipeng-jia/snappyjs/blob/378fc9f8a119d53e4f0c25cd1d3295e28a3f1252/index.js#L69

```ts
  var decompressor = new SnappyDecompressor(compressed)
  var length = decompressor.readUncompressedLength()
  if (length === -1) {
    throw new Error('Invalid Snappy bitstream')
  }
```

This is how all other implementations work
- Lighthouse: https://github.com/sigp/lighthouse/blob/79db2d4deb6a47947699d8a4a39347c19ee6e5d6/beacon_node/lighthouse_network/src/types/pubsub.rs#L62
- Nimbus: https://github.com/status-im/nimbus-eth2/blob/stable/beacon_chain/networking/eth2_network.nim#L2219
- Teku: https://github.com/ConsenSys/teku/blob/b7fc946f9313801476b7fce2e4a02d91e7bb891b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/encoding/SnappyPreparedGossipMessage.java#L122
- Prysm: https://github.com/prysmaticlabs/prysm/blob/develop/beacon-chain/p2p/message_id.go#L52

twoeths:
I created https://github.com/zhipeng-jia/snappyjs/pull/14

packages/lodestar/src/network/gossip/encoding.ts
```diff
@@ -74,6 +80,9 @@ export class DataTransformSnappy {
    * transformed data will then be used to create a `RawGossipsubMessage` to be sent to peers.
    */
   outboundTransform(topicStr: string, data: Uint8Array): Uint8Array {
+    if (data.length > this.maxSizePerMessage) {
+      throw Error(`ssz_snappy encoded data length ${length} > ${this.maxSizePerMessage}`);
+    }
```
dapplion, line 79: Do we really need to limit outbound size? Do other implementations do it?

yarn.lock
```diff
@@ -512,7 +512,78 @@
...
+
+"@chainsafe/lodestar-types@^0.37.0":
+  version "0.37.0"
+  resolved "https://registry.yarnpkg.com/@chainsafe/lodestar-types/-/lodestar-types-0.37.0.tgz#ebc585830051d30e0c877a50b431bfe02742db39"
+  integrity sha512-snDFbYUrtav9JawdMZk8ytr65Y8gZC4uSy60TCUCnHvvi+TqojZvjlX3eAwxktjlf8uailoZqDVNEh5ICzEfJw==
+  dependencies:
+    "@chainsafe/lodestar-params" "^0.37.0"
+    "@chainsafe/ssz" "^0.9.1"
+
+"@chainsafe/lodestar-utils@^0.37.0":
```
dapplion: Please rebase to unstable and delete this diff

packages/lodestar/src/network/gossip/encoding.ts
dapplion: EDIT: We can keep it, since those messages would be rejected by our peers anyway

dapplion:
Mark as ready if this is good to go!

dapplion (dismissed):
LGTM!

dapplion (approved):

PR #4182, opened 2022-06-22
Update forkchoice using latestValidHash updates from execution engine

### LINKED ISSUE #3508: Optimistic Sync: Part 2 - Prune/Validate/Update fork-choice on lastValidHash response

g11tech:
**Motivation**
Post merge, engine can give optimistic `syncing` status on fcU heads which can (should be) eventually resolved into `valid`/`invalid`.
This need to be propagated to forkchoice and remove the consideration of invalids and their descendant from the forkchoice,

This PR implements the fcU updates coming from from the execution via latestValidHash

TODO testcases:
- [x] for propagate valid
~~- [ ] verify is_merge_block if the mergeblock has not yet been validated~~ moved to a separate issue as doesn't addressing this may be a bit more evolved, though EL already should have run terminal validations and hence de duping work isn't really necessary.
- [x] for propagate invalid
  - [x] with latest hash found in forkchoice
  - [x] with latesthash  not found in forkchoice

- [x] real life optimistic sync testing against ELs

Closes #3508

Testcases summary:
```shell
  executionStatus / normal updates
    ✔ preValidation forkchoice setup should be correct
    ✔ correcly invalidate 3C and validate 2C only
    ✔ Validate 3B, 2B, 1A
    ✔ Invalidate 3A, 2A with 2A loosing its bestChild, bestDescendant

  executionStatus / invalidate all postmerge chain
    ✔ preValidation forkchoice setup should be correct
    ✔ all post merge blocks should be invalidated except Cs
    ✔ pre merge block should be the FC head

  executionStatus / poision forkchoice if we invalidate previous valid
    ✔ preValidation forkchoice setup should be correct
    ✔ Validate 3B, 2B, 1A
    ✔ protoarray should be poisioned with a buggy LVH response

  executionStatus / poision forkchoice if we validate previous invalid
    ✔ preValidation forkchoice setup should be correct
    ✔ Inalidate 3B, 2B, 1A
    ✔ protoarray should be poisioned with a buggy LVH response


  13 passing (8ms)
```

### DIFF
--- a/packages/beacon-node/src/chain/blocks/index.ts
+++ b/packages/beacon-node/src/chain/blocks/index.ts
@@ -68,29 +68,44 @@ export async function processBlocks(
 
     // Fully verify a block to be imported immediately after. Does not produce any side-effects besides adding intermediate
     // states in the state cache through regen.
-    const {postStates, executionStatuses, proposerBalanceDeltas} = await verifyBlocksInEpoch.call(
+    const {postStates, executionStatuses, proposerBalanceDeltas, segmentExecStatus} = await verifyBlocksInEpoch.call(
       this,
       parentBlock,
       relevantBlocks,
       opts
     );
 
-    const fullyVerifiedBlocks = relevantBlocks.map(
-      (block, i): FullyVerifiedBlock => ({
-        block,
-        postState: postStates[i],
-        parentBlockSlot: parentSlots[i],
-        executionStatus: executionStatuses[i],
-        proposerBalanceDelta: proposerBalanceDeltas[i],
-        // TODO: Make this param mandatory and capture in gossip
-        seenTimestampSec: opts.seenTimestampSec ?? Math.floor(Date.now() / 1000),
-      })
-    );
+    // If segmentExecStatus has lvhForkchoice then, the entire segment should be invalid
+    // and we need to further propagate
+    if (segmentExecStatus.invalidSegmentLHV !== undefined) {
+      this.forkChoice.validateLatestHash(segmentExecStatus.invalidSegmentLHV);
+    } else if (segmentExecStatus.mayBeValidTillIndex >= 0) {
+      // Only import blocks if atleast one block was valid
+      const relevantBlocksPostExec = relevantBlocks.slice(0, segmentExecStatus.mayBeValidTillIndex + 1);
+      const fullyVerifiedBlocks = relevantBlocksPostExec.map(
+        (block, i): FullyVerifiedBlock => ({
+          block,
+          postState: postStates[i],
+          parentBlockSlot: parentSlots[i],
+          executionStatus: executionStatuses[i],
+          proposerBalanceDelta: proposerBalanceDeltas[i],
+          // TODO: Make this param mandatory and capture in gossip
+          seenTimestampSec: opts.seenTimestampSec ?? Math.floor(Date.now() / 1000),
+        })
+      );
 
-    for (const fullyVerifiedBlock of fullyVerifiedBlocks) {
-      // No need to sleep(0) here since `importBlock` includes a disk write
-      // TODO: Consider batching importBlock too if it takes significant time
-      await importBlock.call(this, fullyVerifiedBlock, opts);
+      for (const fullyVerifiedBlock of fullyVerifiedBlocks) {
+        // No need to sleep(0) here since `importBlock` includes a disk write
+        // TODO: Consider batching importBlock too if it takes significant time
+        await importBlock.call(this, fullyVerifiedBlock, opts);
+      }
+    }
+
+    // If segmentExecStatus has an execError, even if few blocks may be imported
+    // we still need to throw for this segment as it block processor will think
+    // segment imported till end, and will result next into unknown block error
+    if (segmentExecStatus.execAborted !== null) {
+      throw segmentExecStatus.execAborted.execError;
     }
   } catch (e) {
     // above functions should only throw BlockError
--- a/packages/beacon-node/src/chain/blocks/index.ts
+++ b/packages/beacon-node/src/chain/blocks/index.ts
@@ -68,7 +68,7 @@ export async function processBlocks(
 
     // Fully verify a block to be imported immediately after. Does not produce any side-effects besides adding intermediate
     // states in the state cache through regen.
-    const {postStates, executionStatuses, proposerBalanceDeltas, segmentExecStatus} = await verifyBlocksInEpoch.call(
+    const {postStates, proposerBalanceDeltas, segmentExecStatus} = await verifyBlocksInEpoch.call(
       this,
       parentBlock,
       relevantBlocks,
@@ -77,35 +77,30 @@ export async function processBlocks(
 
     // If segmentExecStatus has lvhForkchoice then, the entire segment should be invalid
     // and we need to further propagate
-    if (segmentExecStatus.invalidSegmentLHV !== undefined) {
-      this.forkChoice.validateLatestHash(segmentExecStatus.invalidSegmentLHV);
-    } else if (segmentExecStatus.mayBeValidTillIndex >= 0) {
-      // Only import blocks if atleast one block was valid
-      const relevantBlocksPostExec = relevantBlocks.slice(0, segmentExecStatus.mayBeValidTillIndex + 1);
-      const fullyVerifiedBlocks = relevantBlocksPostExec.map(
-        (block, i): FullyVerifiedBlock => ({
-          block,
-          postState: postStates[i],
-          parentBlockSlot: parentSlots[i],
-          executionStatus: executionStatuses[i],
-          proposerBalanceDelta: proposerBalanceDeltas[i],
-          // TODO: Make this param mandatory and capture in gossip
-          seenTimestampSec: opts.seenTimestampSec ?? Math.floor(Date.now() / 1000),
-        })
-      );
-
-      for (const fullyVerifiedBlock of fullyVerifiedBlocks) {
-        // No need to sleep(0) here since `importBlock` includes a disk write
-        // TODO: Consider batching importBlock too if it takes significant time
-        await importBlock.call(this, fullyVerifiedBlock, opts);
-      }
-    }
-
-    // If segmentExecStatus has an execError, even if few blocks may be imported
-    // we still need to throw for this segment as it block processor will think
-    // segment imported till end, and will result next into unknown block error
     if (segmentExecStatus.execAborted !== null) {
+      if (segmentExecStatus.invalidSegmentLHV !== undefined) {
+        this.forkChoice.validateLatestHash(segmentExecStatus.invalidSegmentLHV);
+      }
+
       throw segmentExecStatus.execAborted.execError;
+    }
+    const {executionStatuses} = segmentExecStatus;
+    const fullyVerifiedBlocks = relevantBlocks.map(
+      (block, i): FullyVerifiedBlock => ({
+        block,
+        postState: postStates[i],
+        parentBlockSlot: parentSlots[i],
+        executionStatus: executionStatuses[i],
+        proposerBalanceDelta: proposerBalanceDeltas[i],
+        // TODO: Make this param mandatory and capture in gossip
+        seenTimestampSec: opts.seenTimestampSec ?? Math.floor(Date.now() / 1000),
+      })
+    );
+
+    for (const fullyVerifiedBlock of fullyVerifiedBlocks) {
+      // No need to sleep(0) here since `importBlock` includes a disk write
+      // TODO: Consider batching importBlock too if it takes significant time
+      await importBlock.call(this, fullyVerifiedBlock, opts);
     }
   } catch (e) {
     // above functions should only throw BlockError
--- a/packages/beacon-node/src/chain/blocks/index.ts
+++ b/packages/beacon-node/src/chain/blocks/index.ts
@@ -81,9 +81,9 @@ export async function processBlocks(
       if (segmentExecStatus.invalidSegmentLHV !== undefined) {
         this.forkChoice.validateLatestHash(segmentExecStatus.invalidSegmentLHV);
       }
-
       throw segmentExecStatus.execAborted.execError;
     }
+
     const {executionStatuses} = segmentExecStatus;
     const fullyVerifiedBlocks = relevantBlocks.map(
       (block, i): FullyVerifiedBlock => ({
--- a/packages/beacon-node/src/chain/blocks/verifyBlock.ts
+++ b/packages/beacon-node/src/chain/blocks/verifyBlock.ts
@@ -76,7 +76,7 @@ export async function verifyBlocksInEpoch(
       verifyBlocksExecutionPayload(this, parentBlock, blocks, preState0, abortController.signal, opts),
     ]);
 
-    if (segmentExecStatus.mergeBlockFound !== null) {
+    if (segmentExecStatus.execAborted === null && segmentExecStatus.mergeBlockFound !== null) {
       // merge block found and is fully valid = state transition + signatures + execution payload.
       // TODO: Will this banner be logged during syncing?
       logOnPowBlock(this.logger, this.config, segmentExecStatus.mergeBlockFound);
--- a/packages/beacon-node/src/chain/blocks/verifyBlock.ts
+++ b/packages/beacon-node/src/chain/blocks/verifyBlock.ts
@@ -1,7 +1,7 @@
 import {CachedBeaconStateAllForks, computeEpochAtSlot} from "@lodestar/state-transition";
 import {allForks, bellatrix} from "@lodestar/types";
 import {toHexString} from "@chainsafe/ssz";
-import {ExecutionStatus, ProtoBlock} from "@lodestar/fork-choice";
+import {MaybeValidExecutionStatus, ProtoBlock} from "@lodestar/fork-choice";
 import {IChainForkConfig} from "@lodestar/config";
 import {ILogger} from "@lodestar/utils";
 import {BlockError, BlockErrorCode} from "../errors/index.js";
@@ -12,7 +12,7 @@ import {ImportBlockOpts} from "./types.js";
 import {POS_PANDA_MERGE_TRANSITION_BANNER} from "./utils/pandaMergeTransitionBanner.js";
 import {verifyBlocksStateTransitionOnly} from "./verifyBlocksStateTransitionOnly.js";
 import {verifyBlocksSignatures} from "./verifyBlocksSignatures.js";
-import {verifyBlocksExecutionPayload} from "./verifyBlocksExecutionPayloads.js";
+import {verifyBlocksExecutionPayload, SegmentExecStatus} from "./verifyBlocksExecutionPayloads.js";
 
 /**
  * Verifies 1 or more blocks are fully valid; from a linear sequence of blocks.
@@ -32,8 +32,9 @@ export async function verifyBlocksInEpoch(
   opts: BlockProcessOpts & ImportBlockOpts
 ): Promise<{
   postStates: CachedBeaconStateAllForks[];
-  executionStatuses: ExecutionStatus[];
+  executionStatuses: MaybeValidExecutionStatus[];
   proposerBalanceDeltas: number[];
+  segmentExecStatus: SegmentExecStatus;
 }> {
   if (blocks.length === 0) {
     throw Error("Empty partiallyVerifiedBlocks");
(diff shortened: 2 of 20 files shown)


packages/fork-choice/src/protoArray/protoArray.ts
```diff
@@ -214,6 +232,135 @@ export class ProtoArray {
...
+   * 2. Throw critical error and exit if a block in finalized chain gets invalidated
+   */
+  validateLatestHash(latestValidExecHash: RootHex, invalidateTillBlockHash: RootHex | null): void {
+    // We can skip calling this if the block was successfully validated and is
+    // to be imported, as on_block would already propagate valid hash
+
+    // Look reverse because its highly likely node with latestValidExecHash is towards the
+    // the leaves of the forkchoice
+    const latestValidHashIndexReverse = this.nodes
+      .slice()
```
twoeths: do we really need this `slice()` call?

packages/fork-choice/src/protoArray/protoArray.ts
```diff
@@ -214,6 +232,135 @@ export class ProtoArray {
...
+   * Optimistic sync validate till validated latest hash, invalidate any decendant branch if invalidate till hash provided
+   * TODO: implementation:
+   * 1. verify is_merge_block if the mergeblock has not yet been validated
+   * 2. Throw critical error and exit if a block in finalized chain gets invalidated
+   */
+  validateLatestHash(latestValidExecHash: RootHex, invalidateTillBlockHash: RootHex | null): void {
+    // We can skip calling this if the block was successfully validated and is
+    // to be imported, as on_block would already propagate valid hash
+
+    // Look reverse because its highly likely node with latestValidExecHash is towards the
```
twoeths, line 250: is this latestValidExecHash unique in `this.nodes()`? If yes maybe we don't need to call `reverse()` for the performance reason?

packages/fork-choice/src/protoArray/protoArray.ts
```diff
@@ -214,6 +232,135 @@ export class ProtoArray {
...
+          break;
+
+        default:
+      }
+
+      node = node.parent !== undefined ? this.getNodeByIndex(node.parent) : undefined;
+    }
+  }
+
+  propagateInValidExecutionStatusByIndex(invalidateFromIndex: number, latestValidHashIndex: number): void {
```
twoeths: can you make the 1st param name consistent to the consumer? this is called `invalidateFromIndex` while the consumer calls `invalidateTillBlockIndex`

also it's nice to have some descriptions in the method header, or draw some ascii texts representing forkchoice branches

packages/fork-choice/src/protoArray/protoArray.ts
```diff
@@ -214,6 +232,135 @@ export class ProtoArray {
...
+    // We need to do a two pass invalidation:
+    //   1. we go up and mark all nodes invalid and then
+    //   2. we need do iterate down and mark all children of invalid nodes invalid
+    //
+    // Pass 1, mark all parents invalid
+    /** invalidateAll is a flag which indicates detection of consensus failure */
+    let invalidateAll = false;
+    let invalidateIndex: number | undefined = invalidateFromIndex;
+    while (invalidateIndex !== undefined && invalidateFromIndex > latestValidHashIndex) {
```
twoeths: should it be 
```typescript
while (invalidateIndex !== undefined && invalidateIndex > latestValidHashIndex) {
```

twoeths: I suppose adding some unit tests for those new methods should help

g11tech: fixed! thanks for spotting it @tuyennhv ! yes will add a test case around this as well.

packages/fork-choice/src/protoArray/protoArray.ts
g11tech: Most likely, atleast when a validator hasn't produced two (slashable) versions of the block for a particular slot, but the one to one correspondence still needs to be established (with prevRando + timestamp combo should make it unique across forks/skipped slots).
As discussed with @dapplion in the discord channel, we can may be put this up in R & D discord and come implement the index.

however once the EL has synced up, most likely (99% of the time) this is going to be just one of the leaves (again most likely head block)

may be we can tackle it separately then. (I can add the above analysis to a comment for some context)

packages/fork-choice/src/protoArray/protoArray.ts
g11tech: umm, may be i can  loop reverse (without slicing it off), would be efficient as well

packages/fork-choice/src/protoArray/protoArray.ts
```diff
@@ -214,6 +236,177 @@ export class ProtoArray {
...
+    //
+    // We can also implement the index to lookup for exec hash => proto block, but it
+    // still needs to be established properly (though is highly likely) than a unique
+    // exec hash maps to a unique beacon block.
+    // For more context on this please checkout the following conversation:
+    // https://github.com/ChainSafe/lodestar/pull/4182#discussion_r914770167
+
+    if (execResponse.executionStatus === ExecutionStatus.Valid) {
+      const {latestValidExecHash} = execResponse;
+      let latestValidHashIndex = this.nodes.length - 1;
```
twoeths: the default latest valid hash index should not be the last node, maybe make it `undefined` or `null`?

packages/fork-choice/src/protoArray/protoArray.ts
```diff
@@ -214,6 +236,177 @@ export class ProtoArray {
...
+
+      for (; latestValidHashIndex >= 0; latestValidHashIndex--) {
```
twoeths: perhaps extract this to an array util so that it's reusable and testable

packages/beacon-node/src/chain/blocks/verifyBlock.ts
```diff
@@ -193,17 +206,23 @@ export async function verifyBlockStateTransition(
     switch (execResult.status) {
       case ExecutePayloadStatus.VALID:
         executionStatus = ExecutionStatus.Valid;
-        chain.forkChoice.validateLatestHash(execResult.latestValidHash, null);
+        chain.forkChoice.validateLatestHash({
```
twoeths: @g11tech my understanding is `latestValidHash` is the current execution payload which we have not imported into forkchoice, is it an issue?

g11tech: @tuyennhv its not an issue here, since the latestValidHash should be the parent of the current payload, as current payload should not have become part of the chain as it has not been fcU-ed yet. But I will validate this.

twoeths: @g11tech  what I see is `go-ethereum` returns the current execution payload as `latestValidHash` most of the time so we should handle it https://github.com/ethereum/go-ethereum/blob/e108d36575fd93e6088259cdabeed6dddf76458c/eth/catalyst/api.go#L324

g11tech: modified, just directly passing the parent's exechash now, so no matter what EL responds as LVH (block or its parent), we pass the correct one

packages/fork-choice/src/protoArray/protoArray.ts
g11tech: its not really generic, and is covered by the validation test cases.

packages/fork-choice/src/protoArray/protoArray.ts
g11tech: thanks for the input, i have initialized to -1, keeping in line with "not found" assignments of the indices. if you still feel strongly for undefined or null to represent `not found`, do respond :slightly_smiling_face:

packages/beacon-node/src/chain/blocks/verifyBlock.ts
```diff
@@ -115,17 +131,27 @@ export async function verifyBlockStateTransition(
...
+        if (parentBlock.executionPayloadBlockHash !== null) {
+          chain.forkChoice.validateLatestHash({
+            executionStatus,
+            latestValidExecHash: parentBlock.executionPayloadBlockHash,
+            invalidateTillBlockHash: null,
+          });
+        }
         break; // OK
 
       case ExecutePayloadStatus.INVALID: {
```
wemeetagain: should there be `executionStatus = ExecutionStatus.Invalid` here?

g11tech: we are just throwing below after `validateLatestHash` invalid status propagation, so its not getting used.

g11tech: also added some typesafety to it:

```diff
-  let executionStatus: ExecutionStatus;
+  let executionStatus: MaybeValidExecutionStatus;
```

twoeths:
@g11tech take a look at https://github.com/ethereum/consensus-specs/pull/2954

g11tech:
> @g11tech take a look at [ethereum/consensus-specs#2954](https://github.com/ethereum/consensus-specs/pull/2954)

yes, the PR is the result of the discussions we already had on discord interop channel. 
And this PR already incorporates this (if latestValidHash is not found in forkChoice, only the block is invalidated)

packages/fork-choice/src/protoArray/protoArray.ts
```diff
@@ -214,10 +237,245 @@ export class ProtoArray {
...
+            ...this.lvhError,
+          });
+
+        case ExecutionStatus.Syncing:
+          // If the node is valid and its parent is marked syncing, we need to
+          // propagate the execution status up
+          node.executionStatus = ExecutionStatus.Valid;
+          break;
+
+        default:
```
dapplion: Empty default clause?

packages/fork-choice/src/protoArray/protoArray.ts
```diff
@@ -482,6 +747,7 @@ export class ProtoArray {
    * head.
    */
   nodeIsViableForHead(node: ProtoNode): boolean {
+    if (node.executionStatus === ExecutionStatus.Invalid) return false;
```
dapplion:
```suggestion
    if (node.executionStatus === ExecutionStatus.Invalid) {
      return false;
    }
    
```

packages/fork-choice/src/protoArray/protoArray.ts
```diff
@@ -214,10 +237,245 @@ export class ProtoArray {
...
+    }
+
+    // Pass 2: mark all children of invalid nodes as invalid
+    for (let nodeIndex = 0; nodeIndex < this.nodes.length; nodeIndex++) {
+      const node = this.getNodeFromIndex(nodeIndex);
+      const parent = node.parent !== undefined ? this.getNodeByIndex(node.parent) : undefined;
+      // Only invalidate if this is post merge, and either parent is invalid or the
+      // concensus has failed
+      if (node.executionStatus !== ExecutionStatus.PreMerge && parent?.executionStatus === ExecutionStatus.Invalid) {
+        if (node.executionStatus === ExecutionStatus.Valid) {
```
dapplion: Should this two if statements be merged?

g11tech: simplified the conditions, hopefully its more clear now

packages/beacon-node/src/chain/blocks/index.ts
```diff
@@ -73,14 +73,21 @@ export async function processBlocks(
 
     // Fully verify a block to be imported immediately after. Does not produce any side-effects besides adding intermediate
     // states in the state cache through regen.
-    const {postStates, executionStatuses, proposerBalanceDeltas} = await verifyBlocksInEpoch(
+    const {postStates, executionStatuses, proposerBalanceDeltas, segmentExecStatus} = await verifyBlocksInEpoch(
```
dapplion, line 76: Maybe better to name `segmentExecStatus` `execStatuses`?

packages/beacon-node/src/chain/blocks/verifyBlocksExecutionPayloads.ts
```diff
@@ -114,14 +141,14 @@ export async function verifyBlocksExecutionPayload(
     parentBlock.executionStatus !== ExecutionStatus.PreMerge ||
     lastBlock.message.slot + opts.safeSlotsToImportOptimistically < currentSlot;
 
-  for (const block of blocks) {
+  for (let blockIndex = 0; blockIndex < blocks.length; blockIndex++) {
```
dapplion, line 144: Is blockIndex used now?

packages/beacon-node/src/chain/blocks/verifyBlocksExecutionPayloads.ts
```diff
@@ -294,11 +334,58 @@ export async function verifyBlockExecutionPayload(
...
     case ExecutePayloadStatus.ELERROR:
-    case ExecutePayloadStatus.UNAVAILABLE:
-      throw new BlockError(block, {
+    case ExecutePayloadStatus.UNAVAILABLE: {
+      const execError = new BlockError(block, {
         code: BlockErrorCode.EXECUTION_ENGINE_ERROR,
         execStatus: execResult.status,
         errorMessage: execResult.validationError,
       });
+      return {executionStatus: null, execError} as VerifyBlockExecutionResponse;
```
dapplion, line 343: Why is this casting necessary?

packages/fork-choice/src/protoArray/protoArray.ts
```diff
@@ -84,7 +89,7 @@ export class ProtoArray {
     currentSlot,
   }: {
     deltas: number[];
-    proposerBoost: ProposerBoost | null;
+    proposerBoost?: ProposerBoost | null;
```
dapplion: Why adding `?` here?

packages/fork-choice/src/protoArray/protoArray.ts
```diff
@@ -17,6 +21,7 @@ export class ProtoArray {
   finalizedRoot: RootHex;
   nodes: ProtoNode[] = [];
   indices = new Map<RootHex, number>();
+  lvhError?: LVHExecError;
```
dapplion, line 23: This PR should include the mechanism to kill beacon process from the beacon package. Just check every slot or at some predicatable spot

packages/fork-choice/src/protoArray/protoArray.ts
```diff
@@ -1,12 +1,16 @@
...
-import {ProtoBlock, ProtoNode, HEX_ZERO_HASH} from "./interface.js";
-import {ProtoArrayError, ProtoArrayErrorCode} from "./errors.js";
+import {ProtoBlock, ProtoNode, HEX_ZERO_HASH, ExecutionStatus, LVHExecResponse} from "./interface.js";
+import {ProtoArrayError, ProtoArrayErrorCode, LVHExecError, LVHExecErrorCode} from "./errors.js";
 
 export const DEFAULT_PRUNE_THRESHOLD = 0;
 type ProposerBoost = {root: RootHex; score: number};
 
+const ZERO_HASH = Buffer.alloc(32, 0);
+const ZERO_HASH_HEX = toHexString(ZERO_HASH);
```
dapplion: Just merge this two together
```suggestion
const ZERO_HASH_HEX = toHexString(Buffer.alloc(32, 0));
```

packages/fork-choice/src/protoArray/protoArray.ts
```diff
@@ -218,10 +242,220 @@ export class ProtoArray {
...
+      validNode.executionStatus = ExecutionStatus.Valid;
+    }
+    return validNode;
+  }
+
   /**
    * Follows the best-descendant links to find the best-block (i.e., head-block).
    */
   findHead(justifiedRoot: RootHex, currentSlot: Slot): RootHex {
+    if (this.lvhError !== undefined) {
```
dapplion: If someday someone changes lvhError to be null instead of undefined
```suggestion
    if (!this.lvhError) {
```

packages/fork-choice/test/unit/protoArray/executionStatusUpdates.test.ts
```diff
@@ -0,0 +1,645 @@
...
+      executionStatus: ExecutionStatus.Invalid,
+      latestValidExecHash: "2C",
+      invalidateFromBlockHash: "3C",
+    },
+    3
+  );
+
+  const invalidate3CValidate2CForkChoice = collectProtoarrayValidationStatus(fc);
+  it("correcly invalidate 3C and validate 2C only", () => {
+    expect(invalidate3CValidate2CForkChoice).to.be.deep.equal([
```
dapplion: This assert state can be declared in one line? Or are longer than 120 characters?

dapplion (commented):
:heavy_check_mark: meta-approving! Looks really good, thanks for the simplifications.

Leaving un-approved until we cut 0.42 release. Left some minor comments, if you want to address before then

packages/fork-choice/test/unit/protoArray/executionStatusUpdates.test.ts
g11tech: yea missed updating this one

packages/beacon-node/src/chain/blocks/verifyBlocksExecutionPayloads.ts
g11tech: blockIndex is used in getSegmentErrorResponse, to iterate backwards to see if lvh present in the segment itself

packages/fork-choice/test/unit/protoArray/executionStatusUpdates.test.ts
g11tech: updated

packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -146,6 +146,13 @@ export class ForkChoice implements IForkChoice {
     }
   }
 
+  /**
+   * Successful call if no critical errors, else throws
+   */
+  checkSaneForkChoice(): void {
+    this.protoArray.findHead(this.fcStore.justified.checkpoint.rootHex, this.fcStore.currentSlot);
```
dapplion: Why not just return the stored error?

```ts
isProtoArrayInIrrecoverableError(): Error | null
```

packages/beacon-node/src/chain/chain.ts
```diff
@@ -479,6 +479,12 @@ export class BeaconChain implements IBeaconChain {
     this.logger.verbose("Clock slot", {slot});
 
     // CRITICAL UPDATE
+    try {
+      this.forkChoice.checkSaneForkChoice();
+    } catch (e) {
+      this.logger.error("Critical error in forkChoice, shutting down gracefully", {}, e as Error);
+      process.kill(process.pid, "SIGINT");
```
dapplion: No you **must** never call process.kill outside of the cli package. This code must be consumed programatically

dapplion: See

https://github.com/ChainSafe/lodestar/blob/20edb58ca3207b7cf8e1450172f9575cd1defaea/packages/cli/src/cmds/validator/handler.ts#L45

packages/fork-choice/src/forkChoice/forkChoice.ts
g11tech: aren't the kind of errors thrown in proto-array's `findHead` all irrecoverable? lvh Error is just one kind
If they are, then i can do a small refactor of findHead and break it into two parts, one to check and get the error (which can be used as you suggested), and original findHead will use the new error inspect fn and throw if there is one.

dapplion: Calling findHead over and over just to check if there's an error feels wrong

dapplion: approach a:
```ts
if (forkChoice.irrecoverableError) {
  processShutdownCallback(forkChoice.irrecoverableError)
}
```
approach b
```ts
try {
  forkChoice.protoArray.findHead(this.fcStore.justified.checkpoint.rootHex, this.fcStore.currentSlot);
} catch (e) {
  processShutdownCallback(forkChoice.irrecoverableError)
}
```

approach a looks much more sane to me. Given how serious is calling processShutdownCallback() I want that to happen only for very explicit code paths

packages/fork-choice/src/protoArray/protoArray.ts
```diff
@@ -218,10 +241,220 @@ export class ProtoArray {
...
+      } else {
+        this.propagateInValidExecutionStatusByIndex(invalidateFromIndex, latestValidHashIndex, currentSlot);
+      }
+    }
+  }
+
+  private propagateValidExecutionStatusByIndex(validNodeIndex: number): void {
+    let nodeIndex: number | undefined = validNodeIndex;
+    // propagate till we keep encountering syncing status
+    while (nodeIndex !== undefined) {
```
wemeetagain, line 324: This loop should also stop when the parent is already valid as a cases in addition to going until invalid, premerge, or thru all ancestors.

wemeetagain (changes requested):

packages/fork-choice/src/protoArray/protoArray.ts
g11tech: updated the loop break condition to !== syncing  :+1:

wemeetagain (approved):
lgtm

PR #4194, opened 2022-06-26
Move IndicesService inside of ValidatorStore

dapplion:
**Motivation**

- While reviewing https://github.com/ChainSafe/lodestar/pull/3883 I noticed that ValidatorStore, IndicesService and DopplegangerService are unnecessarily complex because they three depend on each other, creating cyclic dependencies.

By moving IndicesService into ValidatorStore things simplify a bunch and would allow for a simpler https://github.com/ChainSafe/lodestar/pull/3883 implementation

**Description**

- Move IndicesService inside of ValidatorStore
- Drop unnecessary stubs in dutiesServices tests
- Make PrepareBeaconProposerService a function since it holds no state


g11tech (approved):
awesome! should also make builder PR cleaner! :rocket:

PR #4199, opened 2022-06-26
Await workers temination on stop

dapplion:
**Motivation**

There are many seg faults occurring in our CI. Most of them suggest to come from the multi-threaded workers.

One suspicious practice that we have is to not await worker termination. BeaconNode consumer just aborts an AbortController which immediately returns and gives the impression downstream that everything is cleaned up and finished.

**Description**

Await workers temination on stop

**NOTE: I have no evidence to support that this change has anything to do with the motivation above**

Even it does not help directly on test stability, it makes sense to await this specific step given that workers are resource intensive.


g11tech (approved):
LGTM!

PR #4201, opened 2022-06-27
Join cli+keymanager account managment

dapplion:
**Motivation**

Current unstable has some serious issues:
- Keymanager does not persist keystores in a way that are immediately runnable, rendering the feature not usable
- There are multiple code paths to import keystores, between import cmd, validator cmd, and keymanager. This inconsistencies are confusing, and have caused issues in the past.
- lockfile logic is dangerously scattered across multiple code paths, which can lead to the lock not being detected.
- RemoteKey server is incomplete pending this PR

**Description**

Join cli+keymanager account managment into cli

- Move keymanager code into validator cli
- keystores are _only_ written and read to disk via PersistedKeysBackend. import cmd, validator cmd and keymanager all use this class
- Make keymanager persisted keystores and remoteKeys readable after a restart


g11tech:
structure looks good! :+1:

packages/cli/src/cmds/validator/keymanager/persistedKeys.ts
```diff
@@ -0,0 +1,271 @@
...
+    passphraseFilepath: string;
+  } {
+    // TODO: Ensure correct formating 0x prefixed
+
+    const dirpath = path.join(this.paths.keystoresDir, pubkeyHex);
+
+    return {
+      dirpath,
+      keystoreFilepath: path.join(this.paths.secretsDir, pubkeyHex),
+      passphraseFilepath: path.join(dirpath, "voting-keystore.json"),
```
wemeetagain: are these swapped?

dapplion: Yep, good pick

dapplion:
Last commits add e2e tests that programatically tests this flows:
- run validation directly with keystores in disk
- import keystores from disk then run
- import keystores from API then run
- import remote keys from API then run

Plus
- Delete keys flows
- Test .lock if run twice
- Test authentication

dapplion:
Test this build's cli manually to ensure validators are read correctly :+1:

wemeetagain (approved):

packages/cli/src/util/passphrase.ts
```diff
@@ -1,15 +1,13 @@
...
 
 /**
  * Utility to read file as UTF8 and strip any trailing new lines.
  * All passphrase files must be read with this function
  */
 export function readPassphraseFile(passphraseFile: string): string {
   const data = fs.readFileSync(passphraseFile, "utf8");
-  const passphrase = stripOffNewlines(data);
+  // Remove trailing new lines '\n' or '\r' if any
+  const passphrase = data.replace(/[\n|\r]+$/g, "");
```
nflaig, line 10: @dapplion  what was the reason for changing this line?

I would recommend that we clean this up, either use `stripOffNewlines` here or get rid of the function altogether

https://github.com/ChainSafe/lodestar/blob/9f7b83037978ceaba49119b9506567078d12fdec/packages/cli/src/util/stripOffNewlines.ts#L4

dapplion: I don't remember to be honest

nflaig: let's clean this up https://github.com/ChainSafe/lodestar/pull/5283

PR #4209, opened 2022-06-28
PrepareNextSlot scheduler

### LINKED ISSUE #4054: Optimize issuing advance engine fcU for prepare payload
Currently the engine fcU is issued just after a  successful block import (https://github.com/ChainSafe/lodestar/pull/3965), predicting the new head to be the head in the next slot if BN has any beacon proposer. Normally this would work just fine, but this isn't sufficient and the following scenarios needs to be addressed:

- [ ] wait till the last interval of slot before issuing fcU unless the head is already last imported
- [ ] issue the fcU
  - [ ] for skipped slot
  - [ ] for head change

twoeths:
**Motivation**

+ We want to prepare for next payload for skipped slot if one of our connected validators is the proposer for next slot
+ The current PrecomputeEpochTransition does not work in Bellatrix because we always do advanced state transition in `importBlock()`

**Description**

- Rewrite PrepareNextSlot scheduler to be run 4s before next slot
  - For all forks, we'll precompute epoch transition 4s before next epoch (this is the same to what we've implemented so far)
  - For Bellatrix, always do advanced state transition in order to prepare next payload if one of connected validators is block proposer
    - Max 1 epoch transition, this usually happens at the end of an epoch
    - The 4s duration was double checked with lighthouse, I think this should be enough to prepare for next payload
    - Technically we can get proposer list if `prepareEpoch === headEpoch` but `processSlot()` is cheap enough, we'll need the next state to prepare payload anyway. If `prepareEpoch === headEpoch+1`, we need to do epoch transition first anyway to know the proposers
- Remove the `processSlot()` call in `importBlock()`

Screen Shot 2022-06-28 at 09 54 40 

Closes #4054


packages/lodestar/src/chain/prepareNextSlot.ts
```diff
@@ -0,0 +1,129 @@
...
+    const isLastEpochSlot = (clockSlot + 1) % SLOTS_PER_EPOCH === 0;
+    if (this.config.getForkSeq(prepareEpoch) < ForkSeq.bellatrix && !isLastEpochSlot) {
+      return;
+    }
+
+    const slotMs = this.config.SECONDS_PER_SLOT * 1000;
+    // At 1/3 slot time before the next slot, we either prepare payload or precompute epoch transition
+    await sleep(slotMs - slotMs / SCHEDULER_LOOKAHEAD_FACTOR, this.signal);
+
+    const {slot: headSlot, blockRoot: headRoot} = this.chain.forkChoice.getHead();
```
g11tech: should we try optimizing head for next slot, i.e. process attestations of this `clockSlot` as well and call updateHead. it would reduce any reorgs possibility immensely.

g11tech:
LGTM @tuyennhv looks amazing! just a small optimization point if we want to do it.

wemeetagain (dismissed):
LGTM, file conflict tho

wemeetagain (approved):

PR #4216, opened 2022-06-28
Move types dir to root

dapplion:
**Motivation**

extra types definitions are duplicated across lodestar cli and spec-test-utils packages. This causes weird type issues since the same library has different types depending on where it's loaded.

**Description**

Move types dir to root, and re-use across all packages


dadepo:
Except the failing CI, looks fine. The CI break looks related to libp2p-mplex so may not be a direct result of the changes?

dapplion:
> Except the failing CI, looks fine. The CI break looks related to libp2p-mplex so may not be a direct result of the changes?

The issue is the root path, which is only used by the benchmarks action

https://github.com/ChainSafe/lodestar/blob/ff0fe787c26633c54e121f72eebe28371547b924/tsconfig.json#L7

wemeetagain (approved):

PR #4220, opened 2022-06-29
Adjust currentTargetUnslashedBalanceIncrements if effectiveBalance is changed

twoeths:
**Motivation**

+ Fix this error

```
Jun-29 04:22:54.602[CHAIN]           error: Block error slot=1176512, errCode=BLOCK_ERROR_PRESTATE_MISSING error=previousTargetUnslashedBalanceIncrements is wrong, expect 6214032, got 6214033,current epoch 36765
```

+ Note that the difference is only 1 so I think it's due to the incorrect `effectiveBalance` of one of validators

**Description**

+ `previousTargetUnslashedBalanceIncrements` may went with the effectiveBalance of previous epoch while `prevTargetUnslStake` in EpochProcess always go with the current epoch's effectiveBalance

+ Adjust `currentTargetUnslashedBalanceIncrements` in `processEffectiveBalance()` before transfer it to `previousTargetUnslashedBalanceIncrements` in `afterEpochProcess`

### DIFF
--- a/packages/beacon-state-transition/src/epoch/processEffectiveBalanceUpdates.ts
+++ b/packages/beacon-state-transition/src/epoch/processEffectiveBalanceUpdates.ts
@@ -1,11 +1,16 @@
 import {
   EFFECTIVE_BALANCE_INCREMENT,
+  ForkSeq,
   HYSTERESIS_DOWNWARD_MULTIPLIER,
   HYSTERESIS_QUOTIENT,
   HYSTERESIS_UPWARD_MULTIPLIER,
   MAX_EFFECTIVE_BALANCE,
+  TIMELY_TARGET_FLAG_INDEX,
 } from "@chainsafe/lodestar-params";
-import {EpochProcess, CachedBeaconStateAllForks} from "../types.js";
+import {EpochProcess, CachedBeaconStateAllForks, BeaconStateAltair} from "../types.js";
+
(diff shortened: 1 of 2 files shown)


packages/beacon-state-transition/src/epoch/processEffectiveBalanceUpdates.ts
```diff
@@ -49,7 +55,17 @@ export function processEffectiveBalanceUpdates(state: CachedBeaconStateAllForks,
       // Should happen rarely, so it's fine to update the tree
       validators.get(i).effectiveBalance = effectiveBalance;
       // Also update the fast cached version
-      effectiveBalanceIncrement = Math.floor(effectiveBalance / EFFECTIVE_BALANCE_INCREMENT);
+      const newEffectiveBalanceIncrement = Math.floor(effectiveBalance / EFFECTIVE_BALANCE_INCREMENT);
+      const deltaEffectiveBalanceIncrement = newEffectiveBalanceIncrement - effectiveBalanceIncrement;
+      if (forkSeq > ForkSeq.phase0) {
+        const currentParticipation = (state as BeaconStateAltair).currentEpochParticipation.getAll();
```
wemeetagain: Is this slow? should we be using `state.currentEpochParticipation.get(i)` ?

twoeths: yes, will update it 👍

twoeths:
@dapplion merging to master as I can get through the errored slot

PR #4224, opened 2022-06-30
Remove event handlers file

dapplion:
**Motivation**

IMO the current split of logic between various flows makes the code unnecessarily complex and error prone. While the event driven pubsub model is fine, in this specific case of chain events it has bad trade-offs. Importing a block for example, requires some processes to happen atomically and sequentially, which events do not guarantee.

I've have been moving code out of event handlers for a while, now it's time to just remove the file entirely to prevent more code being added there.

@dadepo told me he got confused on where to add liveness data capture since he could do it **either** in importBlock or in the event handler. Which one is better is not obvious at first glance. If the event handler for block / attestation didn't exist, there won't be confusion

**Description**

- Remove event handlers file
- Move all code to where the event is emitted if done in Chain itself, or move the event listeners to the BeaconChain class
- Remove error events, since no longer necessary


wemeetagain (approved):
LGTM, agree with the premise

PR #4225, opened 2022-06-30
Stable

wemeetagain:


wemeetagain:
conflicts :(

dapplion:
Proposed solution from other Git flow users: https://medium.com/@jshvarts/dealing-with-conflicts-when-merging-release-to-develop-da289a572f0d

> Step 1. Create a new branch based on release. Let’s call it resolve-conflict
> Step 2. Merge develop into resolve-conflict as a [merge commit](https://git-scm.com/docs/git-merge)
> Step 3. Resolve conflicts in resolve-conflict branch— now you should be able to merge this branch into develop.
> Step 4. Merge resolve-conflict into develop

Discussion about the topic, which a simple conclusion: https://stackoverflow.com/questions/71267538/pr-from-release-to-main-has-conflicts

>  It's perfectly normal to have conflicts when merging main down into develop, which typically happens when the same files are modified on both develop and the release branch after it branches off. That's just a normal course of development **unless you are willing to implement code freezes**.

Lighthouse does code freezes, but I think we should not.

And the same solution is brought up:

> Every time we merge a release branch into main, we then merge main into develop, usually with a new branch off of main where we resolve conflicts first (since they're protected branches).

dapplion:
Conflicts fixed manually with https://github.com/ChainSafe/lodestar/commit/138225fed450b623ffce8131b96c51807733a481

As the references I pasted above, it's very likely to have conflicts merging stable into unstable. For example, this time:
- keymanager-server package does not exist anymore in unstable, but still does in stable and got its version bumped

PR #4230, opened 2022-07-04
Add option to attach beacon node to globalThis

wemeetagain:
**Motivation**

It can be useful to inspect a running beacon node for debugging purposes.

**Description**

Add a hidden cli option `--attachToGlobalThis` which attaches the beacon node to `globalThis.bn`

### DIFF
--- a/packages/cli/src/cmds/beacon/handler.ts
+++ b/packages/cli/src/cmds/beacon/handler.ts
@@ -85,6 +85,8 @@ export async function beaconHandler(args: IBeaconArgs & IGlobalArgs): Promise<vo
       metricsRegistries,
     });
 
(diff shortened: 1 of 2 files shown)


twoeths (approved):

dapplion (commented):
@wemeetagain is this really necessary for debugging? Current modern debugging tools seem to give you access to any variable in execution scopes.

wemeetagain:
Iits definitely not strictly necessary for debugging. But its something that I've had locally for a long time that I finally managed to put into a PR. It's definitely been useful in the past and occasionally to attach to a debug console and inspect the running node. Not sure about you but I've had trouble getting the same UX from debuggers / breakpoints.

PR #4234, opened 2022-07-04
Add support to run validator using deposit cli keystores

g11tech:
**Motivation**
Currently in our merge scripts we only support mnemonic for validator key/secret derivation.

This PR provides the ability to run validator via importing the deposit cli keystores

cc: @philknows (Please test out and approve if this works)

### DIFF
--- a/merge-scripts/goerlishadow-4.vars
+++ b/merge-scripts/goerlishadow-4.vars
@@ -1,3 +1,15 @@
+# Uncomment the following if you want to start validator with deposit cli keystores
+# Keystores and the pass.txt will be picked from the current dir (mounted as /currentDir in container)
+
+# LODESTAR_VALIDATOR_SECRETS_ARGS="--importKeystoresPath /currentDir/keystores --importKeystoresPassword /currentDir/pass.txt"
+
+# Comment this if you want to start validators with deposit cli and uncommented above
+
+LODESTAR_VALIDATOR_SECRETS_ARGS="--fromMnemonic \"lens risk clerk foot verb planet drill roof boost aim salt omit celery tube list permit motor obvious flash demise churn hold wave hollow\" --mnemonicIndexes 0..5"
+
+#---------------- Only Modify below if you know what you are doing ----------------
+#----------------------------------------------------------------------------------
+
 DEVNET_NAME=goerlishadow-4
 CONFIG_GIT_DIR=goerli-shadow-fork-4
 NETWORK_ID=5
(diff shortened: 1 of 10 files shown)


merge-scripts/goerlishadow-4.vars
```diff
@@ -1,3 +1,15 @@
+# Uncomment the following if you want to start validator with deposit cli keystores
```
dadepo: Maybe have the possibility to choose this via a command line argument? I feel having to uncomment/comment is a bit clunky and might not give the best UX

merge-scripts/goerlishadow-4.vars
```diff
@@ -1,3 +1,15 @@
...
+# Keystores and the pass.txt will be picked from the current dir (mounted as /currentDir in container)
+
+# LODESTAR_VALIDATOR_SECRETS_ARGS="--importKeystoresPath /currentDir/keystores --importKeystoresPassword /currentDir/pass.txt"
+
+# Comment this if you want to start validators with deposit cli and uncommented above
+
+LODESTAR_VALIDATOR_SECRETS_ARGS="--fromMnemonic \"lens risk clerk foot verb planet drill roof boost aim salt omit celery tube list permit motor obvious flash demise churn hold wave hollow\" --mnemonicIndexes 0..5"
+
+#---------------- Only Modify below if you know what you are doing ----------------
```
philknows, line 10: I would remove these lines in all network configs TBH since we do want the user to modify JWT secret as a minimum. And the guide already teaches you how to do so.

merge-scripts/goerlishadow-4.vars
philknows: In regards to this alternative, maybe instead of `--withValidator`, we separate it into `--withValidatorMnemonic` and `--withValidatorKeystore` as part of startup command would be better? And we keep the formatting nice and simple in these config files. The guide will teach them how to modify the mnemonic if they choose to do so (already explained in setup guide). And then for keystores, the setup guide can just tell them where to put the files and password and they just start with `--withValidatorKeystore`

g11tech: umm... the issue will be people forgetting to add the flag between runs (highly likely, happens all the time), could lead to messed up imports.

or we can have an interactive input to pick this data for first time (and the script will persist the choice and options next time around in a file). This can also be asked for the jwt-secret, i.e. the personalization options which are a bit different from configuration options. so for the @philknows 's guide, one wouldn't need to open the vars file at all.

(for which the person will have to edit vars assuming they know what they are doing)

dadepo: Sounds good to me. As long as the cost of implementation is not more than the benefit...

dadepo (approved):

merge-scripts/goerlishadow-4.vars
philknows: From a user flow perspective that would be easier to use, as long as the cost if implementation is not that high. We haven't seen many people use it yet or have enough feedback to say this ux flow needs to change. The format we already have works at this point with the guide.

If it's not that difficult to have personalization options, editing the vars files should only remain as an advanced user task and can be a future upgrade for better UX.

Personalization includes: custom jwt token, set feeRecipient and selecting between mnemonic input vs keystore. 

I will test this implementation as it is today then.

philknows (approved):

philknows (approved):

philknows:
LGTM! 🚀

philknows:
Draft updated merge-script guide now includes these instructions:
https://hackmd.io/@philknows/SJyFDNJj5

merge-scripts/goerlishadow-4.vars
g11tech: we can push jwt like a above these two lines

g11tech:
> LGTM! rocket
splitting might be a good idea, as short implementation, will update this PR

g11tech:
@philknows split ted the flag, you may adjust your guide, I have updated usage in README.md here
Also moved JWT_SECRET line above ---edit if you know what you are doing --- segment, you can ask people, oh fee recipient is left, let me add that too

g11tech:
> @philknows split ted the flag, you may adjust your guide, I have updated usage in README.md here Also moved JWT_SECRET line above ---edit if you know what you are doing --- segment, you can ask people, oh fee recipient is left, let me add that too

fee recipient added too! i guess that simplies your setup guide to a good extent!

PR #4279, opened 2022-07-11
Clone states only when necessary

dapplion:
**Motivation**

SSZ v2 introduced a new caching strategy:
- cache in not structurally share, only exists in a single instance
- cache is transfered "forward" on each clone
- The state source of the clone "loses" its cache

**Description**

To prevent losing the cache to only-read clones, clone the state only when it's mutated via a state transition run

@wemeetagain @tuyennhv This PR can have serious implications, please review in depth and think of corner cases. This should be tested on nodes too
- Deployed to contabo-5,contabo-18,contabo-19  Jul 11 2022 19:54:58 GMT+0000


packages/state-transition/src/stateTransition.ts
```diff
@@ -27,6 +27,7 @@ export function stateTransition(
   const block = signedBlock.message;
   const blockSlot = block.slot;
 
+  // .clone() before mutating state in state transition
   let postState = state.clone();
```
dapplion, line 37: Here and line 85 are the only two instances where state is cloned

dapplion:
In all contabo-5,contabo-18,contabo-19 there's a noticeable improvement in block processing times (deployed branch at the blue line)

Screenshot from 2022-07-12 08-35-56

packages/state-transition/src/block/initiateValidatorExit.ts
```diff
@@ -22,7 +23,10 @@ import {CachedBeaconStateAllForks} from "../types.js";
  * ```
  * Forcing consumers to pass the SubTree of `validator` directly mitigates this issue.
  */
-export function initiateValidatorExit(state: CachedBeaconStateAllForks, validator: phase0.Validator): void {
+export function initiateValidatorExit(
+  state: CachedBeaconStateAllForks,
+  validator: CompositeViewDU<typeof ssz.phase0.Validator>
```
dapplion, line 28: Using `CompositeViewDU` to express the need for mutability here

packages/beacon-node/src/chain/regen/regen.ts
```diff
@@ -242,18 +242,19 @@ async function processSlotsToNearestCheckpoint(
...
     let nextEpochSlot = computeStartSlotAtEpoch(preEpoch + 1);
     nextEpochSlot <= postSlot;
     nextEpochSlot += SLOTS_PER_EPOCH
   ) {
+    // processSlots calls .clone() before mutating
     postState = processSlots(postState, nextEpochSlot, metrics);
 
     // Cache state to preserve epoch transition work
-    const checkpointState = postState.clone();
+    const checkpointState = postState;
```
twoeths, line 257: @dapplion I think it's safer to do `postState.clone(false)`?

from your implementation

```typescript
  clone(dontTransferCache?: boolean): this {
    if (dontTransferCache) {
      return this.type.getViewDU(this.node) as this;
    } else {
      const cache = this.cache;
      this.clearCache();
      return this.type.getViewDU(this.node, cache) as this;
    }
  }
```

dapplion: Sorry I don't follow, what would it be safer to do a clone here?

dapplion: @tuyennhv The point of this PR is to think critically about when our state can actually be mutated. My opinion is that we clone is too many places that a state can't be mutated so cloning is an unnecessary expense that slows than Lodestar for no gain

twoeths: @dapplion in the past we used to have an issue of missing a `clone()` call, it really took time to debug in that case, @wemeetagain may experience it

so what I mean is instead of dropping the `clone()` call, can we do `clone(false)` with less performance impact?

if both `clone()` and `clone(false)` cause performance issue then we need to benchmark to improve performance?

twoeths: My confusion: I thought the cache is merely used for the mutation, it also contains the index of nodes so that we only have to traverse to the node once.

if we use `clone(false)`, the consumer does not have a chance to use the cached `nodes` of the tree so it's all bad going with either `clone() or clone(false)`

with ssz v2, we should not  ever have a state mutation unless in state transition, thanks @dapplion for your explanation 👍

packages/beacon-node/src/chain/stateCache/stateContextCache.ts
```diff
@@ -41,21 +41,12 @@ export class StateContextCache {
...
-    // clonedCount + 1 as there's a .clone() below
-    this.metrics?.stateClonedCount.observe(item.clonedCount + 1);
-    if (!stateInternalCachePopulated(item)) {
-      this.metrics?.stateInternalCacheMiss.inc();
-    }
-
-    // Clone first to account for metrics below
-    const itemCloned = item.clone();
-
     this.metrics?.stateClonedCount.observe(item.clonedCount);
```
twoeths, line 44: since we don't call `clone()` anymore, should metric observe `clonedCount` here?

twoeths (approved):
inspected all the `state.clone()` call, I think the change is safe

wemeetagain (approved):
LGTM but might be worth deploying to a node to test first

wemeetagain:
Oh, just noticed the metrics above

PR #4290, opened 2022-07-13
Checkpoint balances cache

twoeths:
**Motivation**

Right now whenever we provide a justified checkpoint to forkchoice, we provide balances too but forkchoice may switch justified checkpoint so we want to provide a way to cache balances from checkpoints

**Description**

- Implement a CheckpointBalancesCache, provide checkpoint balances per block process
- Implement `justifiedBalancesGetter` function:
  - Get from the cache first
  - If not find the closest state to checkpoint and extract balances from there

### DIFF
--- a/packages/beacon-node/src/chain/balancesCache.ts
+++ b/packages/beacon-node/src/chain/balancesCache.ts
@@ -9,11 +9,20 @@ import {CheckpointWithHex} from "@lodestar/fork-choice";
 import {Epoch, RootHex} from "@lodestar/types";
 import {toHexString} from "@lodestar/utils";
 
+/** The number of validator balance sets that are cached within `CheckpointBalancesCache`. */
+const MAX_BALANCE_CACHE_SIZE = 4;
+
+type BalancesCacheItem = {
+  rootHex: RootHex;
+  epoch: Epoch;
+  balances: EffectiveBalanceIncrements;
+};
+
 /**
  * Cache EffectiveBalanceIncrements of checkpoint blocks
  */
 export class CheckpointBalancesCache {
-  private readonly balancesByRootByEpoch = new Map<Epoch, Map<RootHex, EffectiveBalanceIncrements>>();
+  private readonly items: BalancesCacheItem[] = [];
 
   /**
    * Inspect the given `state` and determine the root of the block at the first slot of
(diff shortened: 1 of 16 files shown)


packages/beacon-node/src/chain/balancesCache.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+export class CheckpointBalancesCache {
+  private readonly items: BalancesCacheItem[] = [];
+
+  /**
+   * Inspect the given `state` and determine the root of the block at the first slot of
+   * `state.current_epoch`. If there is not already some entry for the given block root, then
+   * add the effective balances from the `state` to the cache.
+   */
+  processState(blockRootHex: RootHex, state: CachedBeaconStateAllForks): void {
```
dapplion, line 32: Is this called somewhere?

twoeths: added, thanks 👍

packages/beacon-node/src/chain/balancesCache.ts
```diff
@@ -0,0 +1,52 @@
...
+    const epoch = state.epochCtx.currentShuffling.epoch;
```
dapplion, line 33: Use `state.epochCtx.currentEpoch`

packages/beacon-node/src/chain/chain.ts
```diff
@@ -388,40 +389,71 @@ export class BeaconChain implements IBeaconChain {
...
     if (state) {
       // most of the time it should go here
       return state;
     }
 
-    // If it's not found, then find the oldest state in the same chain as this one
-    const checkpointSlot = computeStartSlotAtEpoch(checkpoint.epoch);
-    const head = this.forkChoice.getHead();
+    // Find a state in the same branch of checkpoint at same epoch. Balances should exactly the same
+    for (const descendantBlock of this.forkChoice.forwardIterateDescendants(checkpoint.rootHex)) {
```
twoeths, line 424: should we consume `forwardIterateDescendants` only once to improve performance a bit? either extract to an array, or that function could return an array itself

packages/beacon-node/src/chain/blocks/importBlock.ts
```diff
@@ -93,19 +88,12 @@ export async function importBlock(chain: ImportBlockModules, fullyVerifiedBlock:
...
-    const state = getStateForJustifiedBalances(chain, postState, block);
-    onBlockPrecachedData.justifiedBalances = getEffectiveBalanceIncrementsZeroInactive(state);
-  }
-
   const prevFinalizedEpoch = chain.forkChoice.getFinalizedCheckpoint().epoch;
-  chain.forkChoice.onBlock(block.message, postState, onBlockPrecachedData);
+  const blockDelaySec = (Math.floor(Date.now() / 1000) - postState.genesisTime) % chain.config.SECONDS_PER_SLOT;
+  const blockRoot = toHexString(chain.config.getForkTypes(block.message.slot).BeaconBlock.hashTreeRoot(block.message));
+  // Should compute checkpoint balances before forkchoice.onBlock
+  chain.checkpointBalancesCache.processState(blockRoot, postState);
```
dapplion, line 95: Calling fc_store.onVerifiedBlock() in a similar pattern as lighthouse sounds better than here.

packages/beacon-node/src/chain/chain.ts
dapplion: Since it's very unlikely to hit this code path I would not do that for simplicity. Metrics will clearly show if we have to run the iterator twice too often, only then optimize

dapplion (approved):

dapplion (approved):

twoeths:
Tested for a while in nightly group, metrics show that we always hit balances cache

Screen Shot 2022-07-14 at 09 08 30

PR #4292, opened 2022-07-13
Add e2e testing support for browsers 

### LINKED ISSUE #2723: Add e2e testing support for browsers

We have no assurance in CI that code doesn't break in the browser, and it does.

Add Karma like in bls repo with sample webpack config https://github.com/ChainSafe/bls/blob/master/karma.conf.js

nazarhussain:
**Motivation**

Improve the test coverage. 

**Description**

Add testing support for the browsers, using the cypress. For now it covers **electron**, **chrome** and **firefox**

Closes #2723


dapplion:
@nazarhussain we have not used emojis in commit messages before. While not opposed to it let's keep the codebase consistent and if you want bring it up to the team if we should adopt emojis or not

nazarhussain:
The only item left in this PR is to find a workaround for the SSE (sever sent events) for the mocking server and run it inside browsers.

packages/light-client/src/utils/utils.ts
```diff
@@ -74,3 +74,7 @@ export function isEmptyHeader(header: BeaconBlockHeader): boolean {
   const emptyValue = ssz.phase0.BeaconBlockHeader.defaultValue();
   return ssz.phase0.BeaconBlockHeader.equals(emptyValue, header);
 }
+
+// Thanks https://github.com/iliakan/detect-node/blob/master/index.esm.js
+export const isNode =
+  Object.prototype.toString.call(typeof process !== "undefined" ? process : 0) === "[object process]";
```
g11tech, line 80: should we do this determination by inspecting if window object is present? we inject process in eth2-light-client

nazarhussain: I used the same logic we used in the `bls` library. There could be different ways just tried to be consistent. 

https://github.com/ChainSafe/bls/blob/710e7f9f5e5c9ab79ac08460a3d7d5a937b379f1/src/getImplementation.ts#L4

g11tech (commented):
looks awesome! leaving for other to reveiw/approve

karma.base.config.js
```diff
@@ -0,0 +1,18 @@
...
+  exclude: [],
+  preprocessors: {
+    "test/**/*.ts": ["webpack"],
+  },
+  reporters: ["spec"],
+  browsers: ["ChromeHeadlessCI", "Electron", "FirefoxHeadless"],
+  customLaunchers: {
+    ChromeHeadlessCI: {
+      base: "ChromeHeadless",
+      flags: ["--no-sandbox"],
```
dapplion: Why `--no-sandbox`?

packages/light-client/test/unit/syncNaive.test.ts
```diff
@@ -1,25 +1,31 @@
...
-/* eslint-disable @typescript-eslint/naming-convention */
-
-describe("Lightclient flow", () => {
-  before("BLS sanity check", () => {
-    const sk = bls.SecretKey.fromBytes(Buffer.alloc(32, 1));
-    expect(sk.toPublicKey().toHex()).to.equal(
-      "0xaa1a1c26055a329817a5759d877a2795f9499b97d6056edde0eea39512f24e8bc874b4471f0501127abb1ea0d9f68ac1"
-    );
-  });
+describe("syncNaive", function () {
```
dapplion: I added the naive name here but I think we can use something more descriptive. Like "sync with in-memory API (no HTTP)"

dapplion:
Browser tests are pretty quick in CI! 42s, that's cool

dapplion (approved):
Excellent! Thank you for finding the root cause of the issue. We can iterate in the future on having better DX on bls for such usecases

karma.base.config.js
nazarhussain: I was testing some error. Removed it now.

packages/light-client/test/unit/syncNaive.test.ts
nazarhussain: Renamed the test.

PR #4294, opened 2022-07-14
Progressively compute target stake

twoeths:
**Motivation**

Research into computing target stake progressively to compute participation through an epoch instead of at the end of epoch.

**Description**

+ Add `currentTargetUnslashedBalanceIncrements, previoustTargetUnslashedBalanceIncrements` and compute it per block processing
+ Add assertion in `beforeEpochProcess`, by default do the assertion in spec tests
+ Also add cli flag to do the assertion, which allow us to sync from genesis and test

### DIFF
--- a/packages/state-transition/src/cache/epochProcess.ts
+++ b/packages/state-transition/src/cache/epochProcess.ts
@@ -163,7 +163,10 @@ export interface EpochProcess {
   isActiveNextEpoch: boolean[];
 }
 
-export function beforeProcessEpoch(state: CachedBeaconStateAllForks): EpochProcess {
+export function beforeProcessEpoch(
+  state: CachedBeaconStateAllForks,
+  assertCorrectProgressiveBalances = false
+): EpochProcess {
   const {config, epochCtx} = state;
   const forkName = config.getForkName(state.slot);
   const currentEpoch = epochCtx.currentShuffling.epoch;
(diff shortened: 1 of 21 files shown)


dapplion (approved):

dapplion:
For some reason build:docs script breaks in this PR https://github.com/ChainSafe/lodestar/runs/7339174005?check_suite_focus=true

```
@chainsafe/lodestar: Error [ERR_LOADER_CHAIN_INCOMPLETE]: "file:///home/runner/work/lodestar/lodestar/node_modules/ts-node/child-loader.mjs 'resolve'" did not call the next hook in its chain and did not explicitly signal a short circuit. If this is intentional, include `shortCircuit: true` in the hook's return.
@chainsafe/lodestar:     at new NodeError (node:internal/errors:387:5)
@chainsafe/lodestar:     at ESMLoader.resolve (node:internal/modules/esm/loader:860:13)
@chainsafe/lodestar:     at async ESMLoader.getModuleJob (node:internal/modules/esm/loader:439:7)
@chainsafe/lodestar:     at async Promise.all (index 0)
@chainsafe/lodestar:     at async ESMLoader.import (node:internal/modules/esm/loader:541:24)
@chainsafe/lodestar:     at async loadESM (node:internal/process/esm_loader:83:5)
@chainsafe/lodestar:     at async handleMainPromise (node:internal/modules/run_main:65:12) {
@chainsafe/lodestar:   code: 'ERR_LOADER_CHAIN_INCOMPLETE'
@chainsafe/lodestar: }
```

Running locally:
- Node 16: it works 
- Node 18: TBD


Fix:
- Blocked by https://github.com/ChainSafe/lodestar/pull/4297

packages/state-transition/src/cache/epochProcess.ts
```diff
@@ -369,6 +376,23 @@ export function beforeProcessEpoch(state: CachedBeaconStateAllForks): EpochProce
       currTargetUnslStake += effectiveBalanceByIncrement;
     }
   }
+
+  if (opts?.assertCorrectProgressiveBalances) {
```
wemeetagain, line 380: why is this configurable? because it was only used during debugging? and in normal practice it just adds overhead?

wemeetagain: Seems like it could/should just be 'always on'

twoeths: @wemeetagain  yeah I think we can turn it on when:
+ code is unchanged for a while
+ sync through prater
+ sync through mainnet
+ other testnets?

PR #4306, opened 2022-07-16
Preserve backslash character in panda banner

dapplion:
**Motivation**

Current banner prints broken as `\` character is not displayed messing the alignments

**Description**

String.raw preserve backslash character
- see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/raw
- from https://stackoverflow.com/questions/30466581/preserve-all-backslashes-in-a-string


wemeetagain (approved):

wemeetagain:
Good find

PR #4307, opened 2022-07-17
Review bellatrix assembleBody

dapplion:
**Motivation**

Working on a separate issue got stuck for a long time due to this try / catch that ignores all errors

https://github.com/ChainSafe/lodestar/blob/2f0641af6bc195968302926b67a4a293646d198f/packages/beacon-node/src/chain/factory/block/body.ts#L150-L155

After reviewing the assembleBody flow found some issues:
- Heavy use of `null` with different separate meanings which is very dangerous as it can leave to semantic confusion and bugs
- Try / catch swallows errors
- Unclear flow for pre-merge case

**Description**

Review bellatrix assembleBody

### DIFF
--- a/packages/beacon-node/src/chain/factory/block/body.ts
+++ b/packages/beacon-node/src/chain/factory/block/body.ts
@@ -29,7 +29,7 @@ import {IChainForkConfig} from "@lodestar/config";
 import {toHex} from "@lodestar/utils";
 
 import {IBeaconChain} from "../../interface.js";
-import {IExecutionEngine, IExecutionBuilder} from "../../../execution/index.js";
+import {PayloadId, IExecutionEngine, IExecutionBuilder} from "../../../execution/index.js";
 import {ZERO_HASH, ZERO_HASH_HEX} from "../../../constants/index.js";
 import {IEth1ForBlockProduction} from "../../../eth1/index.js";
 import {numToQuantity} from "../../../eth1/provider/utils.js";
@@ -132,13 +132,19 @@ export async function assembleBody<T extends BlockType>(
         proposerPubKey
       );
     } else {
-      (blockBody as bellatrix.BeaconBlockBody).executionPayload = await prepareExecutionPayload(
+      const payloadId = await prepareExecutionPayload(
         chain,
         safeBlockHash,
         finalizedBlockHash ?? ZERO_HASH_HEX,
         currentState as CachedBeaconStateBellatrix,
         feeRecipient
       );
+      // payloadId is only null when execution is pre-merge,
+      // Else it always ether returns payloadId or errors.
+      (blockBody as bellatrix.BeaconBlockBody).executionPayload =
+        payloadId !== null
+          ? await chain.executionEngine.getPayload(payloadId)
+          : ssz.bellatrix.ExecutionPayload.defaultValue();
     }
   }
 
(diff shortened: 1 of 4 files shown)


packages/beacon-node/src/chain/factory/block/body.ts
```diff
@@ -122,39 +122,26 @@ export async function assembleBody<T extends BlockType>(
           feeRecipient
         );
       }
+
```
dapplion, line 125: @g11tech this call here is confusing
```ts

      // This path will not be used in the production, but is here just for merge mock
      // tests because merge-mock requires an fcU to be issued prior to fetch payload
      // header.
      if (chain.executionBuilder.issueLocalFcUForBlockProduction) {
        await prepareExecutionPayload(
          chain,
          safeBlockHash,
          finalizedBlockHash ?? ZERO_HASH_HEX,
          currentState as CachedBeaconStateBellatrix,
          feeRecipient
        );
      }
```
You say that this needs to can fcU, but instead calls the entire `prepareExecutionPayload` flow which does more things

packages/beacon-node/src/chain/factory/block/body.ts
```diff
@@ -175,28 +162,45 @@ export async function prepareExecutionPayload(
...
+  }
+
+  // prepareExecutionPayload will throw error via notifyForkchoiceUpdate if
+  // the EL returns Syncing on this request to prepare a payload
+  // TODO: Handle only this case, DO NOT put a generic try / catch that discards all errors
+  const payloadId = await chain.executionEngine.notifyForkchoiceUpdate(parentHash, safeBlockHash, finalizedBlockHash, {
+    timestamp,
+    prevRandao,
+    suggestedFeeRecipient,
+  });
```
dapplion: @g11tech please re-add the try / catch here only and ignore only a range of expected errors, always logging to some appropriate logging level, else throw

packages/beacon-node/src/chain/factory/block/body.ts
```diff
@@ -175,28 +162,45 @@ export async function prepareExecutionPayload(
   finalizedBlockHash: RootHex,
   state: CachedBeaconStateBellatrix,
   suggestedFeeRecipient: string
-): Promise<PayloadId | null> {
```
g11tech, line 178: `prepareExecutionPayload` is used over here as well: https://github.com/ChainSafe/lodestar/blob/unstable/packages/beacon-node/src/chain/prepareNextSlot.ts#L116 as  a way to early signal the execution layer to produce the blocks so that transactions can be bundled, this change will break that

packages/beacon-node/src/chain/factory/block/body.ts
g11tech: @dapplion this was just a hack to run merge mock, i wanted to it for our mev boost tests, previously `prepareExecutionPayload` was just issuing fcU (and may be catching the payloadId) but pretty harmless for a test.

g11tech (approved):
lgtm! leaving PR merge to @dapplion for review

g11tech:
> lgtm! leaving PR merge to @dapplion for review

as discussed over discord, a followup PR to be done to further simplify `prepareExecutionPayload` with some extra logging and metrics :+1:

PR #4308, opened 2022-07-17
Consistent use of RootHex in notifyForkchoiceUpdate

dapplion:
**Motivation**

No necessary to support a mix of `Root` and `Root | RooHex` of parameters in `notifyForkchoiceUpdate()`

This caused an type issue not guarded by Typescript:
- An interface defines method `fn(arg: Root | RootHex)`
- An implementer class uses `fn(arg: RootHex)`, which compiles fine
- Then `fn()` is called with `RootHex` = TypeError

**Description**

Consistent use of RootHex in notifyForkchoiceUpdate


g11tech (approved):
LGTM!

PR #4299, opened 2022-07-14
Run fork-choice spec tests against production code

dapplion:
**Motivation**

Our current verify / import block flow are quite complex but are not covered by spec tests. Instead the fork-choice spec tests rely on a lot of custom copied code that may differ from production. This is a huge issue as issues may be fixed in the tests but not on production.

**Description**

- Run fork-choice spec tests against production code

Closes https://github.com/ChainSafe/lodestar/issues/3227

PR #4309, opened 2022-07-17
Update execution engine mock

dapplion:
**Motivation**

From https://github.com/ChainSafe/lodestar/pull/4299 brush-up execution engine mock to work well in both upcoming fork-choice spec tests and sim tests.

Also document with latest execution engine APIs spec for context.

**Description**

- Update execution engine mock


g11tech (approved):
LGTM!

PR #4314, opened 2022-07-17
Make dev command handler transparent

dapplion:
**Motivation**

dev command features duplicated initialization logic from beacon and validator. Hard to test some features like keymanager server for https://github.com/ChainSafe/lodestar/pull/4313

**Description**

Make dev command handler transparent
- just call beacon cmd handler as is
- then, just call validator cmd handler as is

https://github.com/ChainSafe/lodestar/blob/e6e273608c8dc97e1742814ec23d94017ffe7cd0/packages/cli/src/cmds/dev/handler.ts#L65-L77


nazarhussain (approved):

wemeetagain:
Test failing https://github.com/ChainSafe/lodestar/runs/7390042459?check_suite_focus=true#step:18:2566

wemeetagain (approved):

PR #4317, opened 2022-07-17
Merge single and multi block processors

dapplion:
**Motivation**

- Step towards https://github.com/ChainSafe/lodestar/pull/3989

**Description**

Merge single and multi block processors

- Equivalent logic
- Moves options from wrapper block type to an opts object
- Drops ChainSegmentError. Once blocks are verified in batched it would not be trivial to know if a previous block was actually valid. Since this is just an optimization for rare cases, not worth the engineering effort now.

### DIFF
--- a/packages/beacon-node/src/chain/blocks/index.ts
+++ b/packages/beacon-node/src/chain/blocks/index.ts
@@ -1,15 +1,15 @@
 /* eslint-disable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment */
 import {allForks} from "@lodestar/types";
-import {sleep, toHex} from "@lodestar/utils";
+import {toHex} from "@lodestar/utils";
 import {JobItemQueue} from "../../util/queue/index.js";
-import {BlockError, BlockErrorCode, ChainSegmentError} from "../errors/index.js";
+import {BlockError, BlockErrorCode} from "../errors/index.js";
 import {BlockProcessOpts} from "../options.js";
 import {IBeaconChain} from "../interface.js";
 import {verifyBlock, VerifyBlockModules} from "./verifyBlock.js";
 import {importBlock, ImportBlockModules} from "./importBlock.js";
 import {assertLinearChainSegment} from "./utils/chainSegment.js";
-import {PartiallyVerifiedBlock} from "./types.js";
-export {PartiallyVerifiedBlockFlags} from "./types.js";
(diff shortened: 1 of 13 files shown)


packages/beacon-node/src/chain/blocks/index.ts
```diff
@@ -58,92 +50,41 @@ export class BlockProcessor {
...
-      // Flag ignoreIfKnown causes BlockErrorCodes ALREADY_KNOWN, GENESIS_BLOCK to resolve.
-      // Return before emitting to not cause loud logging.
-      return;
+      // No need to sleep(0) here since `importBlock` includes a disk write
+      // TODO: Consider batching importBlock too if it takes significant time
+      await importBlock(chain, fullyVerifiedBlock, opts);
     }
+  } catch (e) {
+    // above functions should only throw BlockError
+    const err = getBlockError(e, blocks[0]);
```
wemeetagain, line 74: Shouldn't the block error be for the block that errored?

packages/beacon-node/src/sync/range/chain.ts
```diff
@@ -439,9 +438,10 @@ export class SyncChain {
 
       // At least one block was successfully verified and imported, so we can be sure all
       // previous batches are valid and we only need to download the current failed batch.
-      if (res.err instanceof ChainSegmentError && res.err.importedBlocks > 0) {
-        this.advanceChain(batch.startEpoch);
-      }
+      // TODO: Disabled for now
```
wemeetagain, line 441: can you make an issue for this? I guess it's a small thing, an optimization to avoid reprocessing up to batch.length - 1 blocks.

packages/beacon-node/src/chain/blocks/index.ts
dapplion: After completing this work everything is processed in batch so you don't know which block caused the error. Each function should take care of adding the right block to BlockError which is already happening, but here, at this line you have no clue which block caused this. blocks[0] is a placeholder, it shouldn't really happen. In future PRs we can review what to do in this case

packages/beacon-node/src/sync/range/chain.ts
dapplion: We can do an issue but I don't see this ever being implemented judging for our priorities in the last 2 years. i.e. to help process sync we have tens higher priority items to go after

wemeetagain (approved):

PR #4324, opened 2022-07-18
Rename proposerBalanceDiff to proposerBalanceDelta

dapplion:
**Motivation**

proposerBalanceDelta is more correct since the difference between a number before and after is a delta

**Description**

Rename proposerBalanceDiff to proposerBalanceDelta


wemeetagain (approved):

wemeetagain:
https://github.com/ChainSafe/lodestar/runs/7392294501?check_suite_focus=true#step:18:1349
`FATAL ERROR: v8::FromJust Maybe value is Nothing.`
:imp:

PR #4336, opened 2022-07-22
Produce block: filter out incompatible shuffling attestations

### LINKED ISSUE #4333: Lodestar produced block with invalid attestation during Sepolia instability
Post merge Sepolia has 2 unstable epochs

- the first merge slot was https://sepolia.beaconcha.in/block/115193
- Up until slot 115232, we don't see any of the config issue related validators missing a single slot. 115233 however is missed by lodestar and subsequently 115234&115235 are also missed by Besu and Nimbus. These are providers where no config issue is known, hence it could be a shuffling issue or something else
- The trigger point could be slot 115233, see slot dump from Nimbus node [invalid.tar.gz](https://github.com/ChainSafe/lodestar/files/9150536/invalid.tar.gz)
- Teku also logged that invalid slot. They failed validating the signature but the proposer_index matches the expected one (709) according to beaconcha.in [115233.zip](https://github.com/ChainSafe/lodestar/files/9150573/115233.zip)
- Seems like the block contains an attestation with invalid signature [state_at_115232.ssz.zip](https://github.com/ChainSafe/lodestar/files/9150582/state_at_115232.ssz.zip)
- The offending attestation is
```
IndexedAttestation{attesting_indices=SszList{size=17: 50, 198, 311, 389, 556, 639, 804, 1121, 1238, 1324, 1435, 1448, 1487, 1497, 1612, 1640, 1779}, data=AttestationData{slot=115232, index=0, beacon_block_root=0xacd63ca2b12641cb281fe40a479a5bf37debbba9ae324e4c9089742fdbbe1403, source=Checkpoint{epoch=3599, root=0x5a4b89da91bfe310be27b6cccaf4ead3ec62280d1ed9f4356e6df525a1f2d19b}, target=Checkpoint{epoch=3601, root=0xacd63ca2b12641cb281fe40a479a5bf37debbba9ae324e4c9089742fdbbe1403}}, signature=SszByteVector{0x86330d22c822f54c36456a82420ec64150024016a9ca1dd59218cec47d4d0be8182dbad6eacfff6debecd9ab2c7390840321416e0cb7ce01e6428ddef73f2777ca69e7d7fe448fc1febd86c6bff4d8a509bc29a3ec32e19e3f79aee8661c30ff}}
```
- source is correct, the target is weird. And gap between source and target is a couple of epochs. Source pre-merge and target post-merge.

**Hypothesis**
- attestation is valid wrt side chain and makes it into pool
- beacon node produces block on mainchain (with divergent shuffling) and pulls thsi attestation from the pool (because it was pre-validated)
- block is invalid because bitfield represents side-chain shuffling and is invalid to the assumed mainchain shuffling in the block

_continue_
- the gap 3599-3601 is present in other attestations https://sepolia.beaconcha.in/epoch/3601 - the target root is wrong
- But it might be correct wrt to the side chain, which the node had blocks available at the time
- The gap here at the point merge implies that there was still agreement in latest finalized and thus the source correct and thus lodestar still importing side-chain blocks

twoeths:
**Motivation**

Lodestar includes incompatible shuffling attestations when producing a block, see https://github.com/ChainSafe/lodestar/issues/4333#issuecomment-1191471214

**Description**

+ Filter out incompatible shuffling attestations based on forkchoice and attester dependent root in state
+ Refer to lighthouse's implementation: https://github.com/sigp/lighthouse/blob/21dec6f603edd53ad8d2372941c77ab8098d3b5a/beacon_node/beacon_chain/src/beacon_chain.rs#L1933

Closes #4333

**TODOs**

- [x] Deploy to a node
- [x] Issue curl to produce blocks, make sure there are enough attestations returned

### DIFF
--- a/packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts
+++ b/packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts
@@ -12,8 +12,11 @@ import {
   CachedBeaconStatePhase0,
   CachedBeaconStateAltair,
   computeEpochAtSlot,
+  computeStartSlotAtEpoch,
+  getBlockRootAtSlot,
 } from "@lodestar/state-transition";
 import {toHexString} from "@chainsafe/ssz";
+import {IForkChoice} from "@lodestar/fork-choice";
 import {MapDef} from "../../util/map.js";
 import {intersectUint8Arrays, IntersectResult} from "../../util/bitArray.js";
 import {pruneBySlot} from "./utils.js";
@@ -102,7 +105,7 @@ export class AggregatedAttestationPool {
   /**
    * Get attestations to be included in a block. Returns $MAX_ATTESTATIONS items
    */
-  getAttestationsForBlock(state: CachedBeaconStateAllForks): phase0.Attestation[] {
+  getAttestationsForBlock(forkChoice: IForkChoice, state: CachedBeaconStateAllForks): phase0.Attestation[] {
     const stateSlot = state.slot;
     const stateEpoch = state.epochCtx.epoch;
     const statePrevEpoch = stateEpoch - 1;
@@ -112,7 +115,6 @@ export class AggregatedAttestationPool {
     const attestationsByScore: AttestationWithScore[] = [];
 
     const slots = Array.from(this.attestationGroupByDataHashBySlot.keys()).sort((a, b) => b - a);
-    const {previousJustifiedCheckpoint, currentJustifiedCheckpoint} = state;
     slot: for (const slot of slots) {
       const attestationGroupByDataHash = this.attestationGroupByDataHashBySlot.get(slot);
       // should not happen
@@ -132,14 +134,7 @@ export class AggregatedAttestationPool {
 
       const attestationGroups = Array.from(attestationGroupByDataHash.values());
       for (const attestationGroup of attestationGroups) {
-        if (
-          !isValidAttestationData(
-            stateEpoch,
-            previousJustifiedCheckpoint,
-            currentJustifiedCheckpoint,
-            attestationGroup.data
-          )
-        ) {
+        if (!isValidAttestationData(forkChoice, state, attestationGroup.data)) {
           continue;
         }
         const participation = getParticipation(epoch, attestationGroup.committee);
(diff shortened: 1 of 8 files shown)


twoeths:
I was able to produce blocks with enough attestations on contabo-18, ready for review

packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -403,23 +398,56 @@ export function extractParticipation(
...
+  // minus the shuffling lookahead (usually 2). We call this the "pivot".
+  const pivotSlot = computeStartSlotAtEpoch(targetEpoch - 1) - 1;
+  const statePivotBlockRoot = toHexString(getBlockRootAtSlot(state, pivotSlot));
+
+  // Use fork choice's view of the block DAG to quickly evaluate whether the attestation's
+  // pivot block is the same as the current state's pivot block. If it is, then the
+  // attestation's shuffling is the same as the current state's.
+  // To account for skipped slots, find the first block at *or before* the pivot slot.
+  let pivotBlockRoot = "";
+  for (const protoBlock of forkChoice.iterateAncestorBlocks(toHexString(data.beaconBlockRoot))) {
```
dapplion: This can be optimized using the targetRoot of each protoBlock. So it always takes 2 hops at most:
- Get target of beaconBlockRoot.
- If target protoBlock > pivotSlot, get target's parent. Else return target

If you need another hop of epochs, do again

dapplion (commented):
Overall looks good, one minor optimization for faster fork-choice traversal

g11tech (approved):
LGTM!

packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -753,6 +753,43 @@ export class ForkChoice implements IForkChoice {
     return;
   }
 
+  /**
+   * Find dependent root of a head block
+   * epoch - 2 | epoch - 1 | epoch
+   * pivotSlot |     -     | blockSlot
+   */
+  findAttesterDependentRoot(headBlockHash: Root): RootHex | null {
```
dapplion, line 761: Should this have an epoch argument? I guess this `headBlockHash` comes from a checkpoint so the `headBlockHash` could be from older epochs. In that case you need the epoch argument to now how back you have to go.

twoeths: no we don't need that, it comes from the attestation's `beaconBlockRoot`, not the checkpoint

```typescript
const pivotBlockRoot = forkChoice.findAttesterDependentRoot(data.beaconBlockRoot);
```

dapplion: If we use the target root from the attestation we could save some hops right?

twoeths: @dapplion  1st hop was already the target root from the attestation

```typescript
// 1st hop: target block
block = this.getBlockHex(block.targetRoot);
```

wemeetagain (approved):

PR #4340, opened 2022-07-23
Fix safe optimistic import

g11tech:
**Motivation**
The safe optimistic import got broken with the PR to optimize the sync speed:
because in safe optimistic import, we used to check if are we syncing near head and the block's parent in forkchoice is not post merge then its not safe to import block with optimistic status.

But this was giving error (EL still syncing and spinned up the lodestar on kiln with weak subjectivity sync, so we are syncing near head), as the parent block would not be found in the forkchoice as we no longer verify then import block by block, now we verify a segment and then import

So updated verification to where the segment's safeness is evaluated and updated and based on that the block import is accepted if its safe.

Also removes the check for justified block being post-merge for safe import as it was a redundant check, as removed in this PR: https://github.com/ethereum/consensus-specs/pull/2881

### DIFF
--- a/packages/beacon-node/src/chain/blocks/verifyBlocksExecutionPayloads.ts
+++ b/packages/beacon-node/src/chain/blocks/verifyBlocksExecutionPayloads.ts
@@ -135,7 +135,7 @@ export async function verifyBlocksExecutionPayload(
     }
 
     const {executionStatus} = await verifyBlockExecutionPayload(chain, block, preState0, opts, isOptimisticallySafe);
-    // It becomes optimistically  safefor following blocks if a post-merge block is deemed fit
+    // It becomes optimistically  safe for following blocks if a post-merge block is deemed fit
     // for import. If it would not have been safe verifyBlockExecutionPayload would throw error
     // and we would not be here.
     isOptimisticallySafe ||= executionStatus !== ExecutionStatus.PreMerge;
--- a/packages/beacon-node/src/chain/blocks/verifyBlocksExecutionPayloads.ts
+++ b/packages/beacon-node/src/chain/blocks/verifyBlocksExecutionPayloads.ts
@@ -4,12 +4,10 @@ import {
   isBellatrixBlockBodyType,
   isMergeTransitionBlock as isMergeTransitionBlockFn,
   isExecutionEnabled,
-  isMergeTransitionComplete,
-  BeaconStateBellatrix,
 } from "@lodestar/state-transition";
 import {bellatrix, allForks} from "@lodestar/types";
 import {toHexString} from "@chainsafe/ssz";
-import {IForkChoice, ExecutionStatus, assertValidTerminalPowBlock} from "@lodestar/fork-choice";
+import {IForkChoice, ExecutionStatus, assertValidTerminalPowBlock, ProtoBlock} from "@lodestar/fork-choice";
 import {IChainForkConfig} from "@lodestar/config";
 import {ErrorAborted, ILogger} from "@lodestar/utils";
 import {IExecutionEngine} from "../../execution/engine/index.js";
@@ -35,6 +33,7 @@ type VerifyBlockModules = {
  */
 export async function verifyBlocksExecutionPayload(
   chain: VerifyBlockModules,
+  parentBlock: ProtoBlock,
   blocks: allForks.SignedBeaconBlock[],
   preState0: CachedBeaconStateAllForks,
   signal: AbortSignal,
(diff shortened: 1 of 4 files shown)


packages/beacon-node/src/chain/blocks/verifyBlocksExecutionPayloads.ts
```diff
@@ -41,14 +43,102 @@ export async function verifyBlocksExecutionPayload(
...
     if (signal.aborted) {
       throw new ErrorAborted("verifyBlockExecutionPayloads");
     }
 
-    const {executionStatus} = await verifyBlockExecutionPayload(chain, block, preState0, opts);
+    const {executionStatus} = await verifyBlockExecutionPayload(chain, block, preState0, opts, isOptimisticallySafe);
+    // It becomes optimistically  safefor following blocks if a post-merge block is deemed fit
+    // for import. If it would not have been safe verifyBlockExecutionPayload would throw error
+    // and we would not be here.
+    isOptimisticallySafe ||= executionStatus !== ExecutionStatus.PreMerge;
```
dapplion: While the `||=` operand is pretty cool, can you use a regular if statement since it's more readable?

packages/beacon-node/src/chain/blocks/verifyBlocksExecutionPayloads.ts
```diff
@@ -41,14 +43,102 @@ export async function verifyBlocksExecutionPayload(
...
   for (const block of blocks) {
     // If blocks are invalid in consensus the main promise could resolve before this loop ends.
     // In that case stop sending blocks to execution engine
     if (signal.aborted) {
       throw new ErrorAborted("verifyBlockExecutionPayloads");
     }
 
-    const {executionStatus} = await verifyBlockExecutionPayload(chain, block, preState0, opts);
+    const {executionStatus} = await verifyBlockExecutionPayload(chain, block, preState0, opts, isOptimisticallySafe);
+    // It becomes optimistically  safefor following blocks if a post-merge block is deemed fit
```
dapplion:
```suggestion
    // It becomes optimistically  safe for following blocks if a post-merge block is deemed fit
```

packages/beacon-node/src/chain/blocks/verifyBlocksExecutionPayloads.ts
```diff
@@ -41,14 +43,102 @@ export async function verifyBlocksExecutionPayload(
...
+  //    This means that the merge transition could be underway and we can't afford to import
+  //    a block which is not fully validated as it could affect liveliness of the network.
+  //
+
+  // For this segment of blocks:
+  //   We are optimistically safe with respec to this entire block segment if:
+  //    - we have transitioned to post-merge world or
+  //    - all the blocks are way behind the current slot
+  //    - or we have already imported a post-merge parent of first block of this chain in forkchoice
+  const lastBlock = blocks[blocks.length - 1];
```
dapplion, line 110: Does this function ensure that there are more than 0 blocks?

packages/beacon-node/src/chain/blocks/verifyBlocksExecutionPayloads.ts
```diff
@@ -41,14 +43,102 @@ export async function verifyBlocksExecutionPayload(
...
+  let isOptimisticallySafe =
+    (isBellatrixStateType(preState0) && isMergeTransitionComplete(preState0 as BeaconStateBellatrix)) ||
+    lastBlock.message.slot + opts.safeSlotsToImportOptimistically < chain.clock.currentSlot;
+
+  if (!isOptimisticallySafe) {
+    const firstBlock = blocks[0];
+    const parentRoot = toHexString(firstBlock.message.parentRoot);
+    const parentBlock = chain.forkChoice.getBlockHex(parentRoot);
```
dapplion: You could return the parent block of the segment in sanity checks

https://github.com/ChainSafe/lodestar/blob/3320c206fe846c759799948ad3643ea0fbba2121/packages/beacon-node/src/chain/blocks/verifyBlocksSanityChecks.ts#L98

and re-use here

packages/beacon-node/src/chain/blocks/verifyBlocksExecutionPayloads.ts
g11tech: may be somewhere from where its called? because it didn't give any type issues, but will return early to be safe

g11tech: added a throw just to be safe

dapplion (approved):
Looks great! Beautifully documented too thank you!!! :heart:

PR #4342, opened 2022-07-24
msgIdFn: use Buffer.subarray()

twoeths:
**Motivation**

`Buffer.slice()` only create a view of the Buffer instead of creating a copy so we should use it instead of `Uint8Array.slice()`

**Description**

- Use `Buffer.slice()` in `msgIdFn` function like other places

### DIFF
--- a/packages/beacon-node/src/network/gossip/encoding.ts
+++ b/packages/beacon-node/src/network/gossip/encoding.ts
@@ -13,7 +13,7 @@ import {GossipTopicCache} from "./topic.js";
  */
 export function fastMsgIdFn(rpcMsg: RPC.IMessage): string {
   if (rpcMsg.data) {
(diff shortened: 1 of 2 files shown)


packages/beacon-node/src/network/gossip/encoding.ts
```diff
@@ -53,7 +53,7 @@ export function msgIdFn(gossipTopicCache: GossipTopicCache, msg: GossipsubMessag
     }
   }
 
-  return digest(Buffer.concat(vec)).slice(0, 20);
+  return Buffer.from(digest(Buffer.concat(vec))).slice(0, 20);
```
twoeths: `Buffer.slice()` is deprecated in Node.js 18, should we migrate similar pattern to `Uint8Array.subarray()`?

dapplion: Yes

wemeetagain (approved):

Issue #2857, opened 2021-07-20
sync hashBlocks performance issue


dapplion:
**Describe the bug**

Sync module tracks the blocks downloaded to do peer scoring latter. However a CPU profile reveals it's a very expensive operation. How can the sync module track the blocks in a succinct unique manner without having to hash all the blocks upfront?

![Screenshot from 2021-07-20 14-27-42](https://user-images.githubusercontent.com/35266934/126324057-3e8d82bd-aac5-449c-a2ea-9f736b187ff4.png)

**Expected behavior**

Track blocks in a cheap manner

**Steps to Reproduce**

**Screenshots**

**Desktop (please complete the following information):**

- OS: 
- Version: 
- Branch: 
- Commit hash:

dapplion:
ssz v2 should cache the blocks root so not extra processing is done in the blocks

PR #4343, opened 2022-07-25
Refactor sync hashBlocks

twoeths:
**Motivation**

Using `Buffer.concat` is better than concatenating Uint8Arrays manually both in term of memory and performance

```
bytes utils
    ✔ byteArrayConcat 32 items                                         4.916421e+8 ops/s    2.034000 ns/op        -    1639474 runs   3.96 s
    ✔ Buffer.concat 32                                                 7.434944e+8 ops/s    1.345000 ns/op        -     476782 runs  0.808 s
``` 

**Description**

- `Buffer.from(Uint8Array.buffer)` gives no memory allocation
- then do `Buffer.concat`

### DIFF
--- a/packages/beacon-node/src/sync/range/utils/hashBlocks.ts
+++ b/packages/beacon-node/src/sync/range/utils/hashBlocks.ts
@@ -1,14 +1,23 @@
-import {allForks} from "@lodestar/types";
+import {allForks, RootHex} from "@lodestar/types";
 import {IChainForkConfig} from "@lodestar/config";
-import {byteArrayConcat} from "../../../util/bytes.js";
+import {toHex} from "@lodestar/utils";
 
 /**
- * Hash SignedBeaconBlock in a byte form easy to compare only
- * @param blocks
- * @param config
+ * String to uniquely identify block segments. Used for peer scoring and to compare if batches are equivalent.
  */
-export function hashBlocks(blocks: allForks.SignedBeaconBlock[], config: IChainForkConfig): Uint8Array {
-  return byteArrayConcat(
-    blocks.map((block) => config.getForkTypes(block.message.slot).SignedBeaconBlock.hashTreeRoot(block))
-  );
+export function hashBlocks(blocks: allForks.SignedBeaconBlock[], config: IChainForkConfig): RootHex {
+  switch (blocks.length) {
+    case 0:
+      return "0x";
+    case 1:
+      return toHex(config.getForkTypes(blocks[0].message.slot).SignedBeaconBlock.hashTreeRoot(blocks[0]));
+    default: {
+      const block0 = blocks[0];
+      const blockN = blocks[blocks.length - 1];
+      return (
+        toHex(config.getForkTypes(block0.message.slot).SignedBeaconBlock.hashTreeRoot(block0)) +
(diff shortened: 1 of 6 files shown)


g11tech:
didn't we move away from using buffers? not sure though why we did it?

twoeths:
> didn't we move away from using buffers?

I think for modules that run in Browser like `as-sha256` or `lightclient`, we have to use Uint8Array. For modules that we're sure they run in Node.js environment, we can use Buffer if performance matters in that context

> not sure though why we did it?

Buffer turns out to be better than Uint8Array in some scenarios:
- In the past @dapplion found out that using `Buffer.toString("hex")` is way better than the ssz `toHexString()`
- In this context, concatenating Uint8Array manually is not efficient, see https://github.com/achingbrain/uint8arraylist/issues/24#issue-1316422497

Looking into the `Buffer` doc, I found that Buffer.concat is better because it uses the internal preallocated Buffer instances, see https://nodejs.org/api/buffer.html
`The Buffer module pre-allocates an internal Buffer instance of size [Buffer.poolSize] that is used as a pool for the fast allocation of new Buffer instances created using [Buffer.allocUnsafe()], [Buffer.from(array)], [Buffer.concat()], and the deprecated new Buffer(size)`

dapplion (dismissed):
@tuyennhv we should be very explicit about this function being meant for NodeJS only. And until we don't have browser tests in Lodestar merging this is not safe as it can break everything again.

g11tech:
> @tuyennhv we should be very explicit about this function being meant for NodeJS only. And until we don't have browser tests in Lodestar merging this is not safe as it can break everything again.

I think its fine now, because in eth2-light-client demo, we inject buffer using craco https://github.com/ChainSafe/eth2-light-client-demo/blob/master/craco.config.js#L37:L39

g11tech (approved):

twoeths:
@dapplion this function is only used in range sync of the BeaconNode, we can make it clear by:
- have a separate nodejs util folder
- or add more comments to it

what do you think?

dapplion:
@tuyennhv Since byteArrayConcat was used only in hashBlocks I've refactored the function to not even need it.
- It should indirectly fix https://github.com/ChainSafe/lodestar/issues/2857

packages/beacon-node/src/sync/range/utils/hashBlocks.ts
```diff
@@ -1,14 +1,23 @@
...
-  );
+export function hashBlocks(blocks: allForks.SignedBeaconBlock[], config: IChainForkConfig): RootHex {
+  switch (blocks.length) {
+    case 0:
+      return "0x";
+    case 1:
+      return toHex(config.getForkTypes(blocks[0].message.slot).SignedBeaconBlock.hashTreeRoot(blocks[0]));
+    default: {
+      const block0 = blocks[0];
+      const blockN = blocks[blocks.length - 1];
```
wemeetagain, line 16: Is it possible to halt syncing with a batch like: `block ... bad block ... block`?

dapplion: This hash is only used for peer scoring, and attribute processing errors to the right set of peers

wemeetagain (approved):

PR #4344, opened 2022-07-25
Add Eth1 and builder metrics

g11tech:
Add Eth1 and builder metrics in grafana dashboard

# Eth1 stats 
(below scrape stats)

image
image

# Builder Metrics
(below Block production stats)

image

More metrics PRs will follow for 
- Merge stats
- execution engine stats (this is pending on the retry PR merge)

### DIFF
--- a/dashboards/lodestar_general.json
+++ b/dashboards/lodestar_general.json
@@ -3808,11 +3808,1051 @@
     },
     {
       "collapsed": true,
+      "datasource": null,
       "gridPos": {
         "h": 1,
         "w": 24,
         "x": 0,
         "y": 20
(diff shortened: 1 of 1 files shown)


twoeths (approved):
looks good to me, not sure why Builder Metric "Rest API response time" is so big 👀

g11tech:
> looks good to me, not sure why Builder Metric "Rest API response time" is so big eyes

yea, registerValidator has like 500 vals in it in one query so its understandable, but there are a few produceBlindedBlocks calls which call builder's getPayloadHeader api which consume like 3+sec

PR #4345, opened 2022-07-25
Use NodeJS native stream utils

dapplion:
**Motivation**

NodeJS stream utils are pretty nice and work very well with async iterables. No need to use it-* utils for NodeJS only code.

**Description**

- Use NodeJS native stream utils
- Reduce one level of piping in sendRequest
  - This requires a deep change on timeout tests


wemeetagain:
The reducing one level of piping looks good.

Is using the native pipeline function required for that refactor? I'm mildly wary of using the native node functions 'just because' and reserve their use for cases where it resolves a known cpu or memory bottleneck. Reason is that it makes running the chain / experimenting with alternative runtime environments unnecessarily harder.

dapplion:
- Closing for https://github.com/ChainSafe/lodestar/pull/4385

PR #4359, opened 2022-08-03
Build from source by default

dapplion:
**Motivation**

It doesn't make sense that we discourage installation via NPM and then distribute and image that installs from NPM. 

- Note @dadepo has not been able to find a solution to issue https://github.com/ChainSafe/lodestar/issues/3596 after 5 months.

**Description**

Build from source by default

- This image size: 277MB (from src)
- Previous image size: 251MB (from NPM)

The difference is due to extra files making it to the final image. With some `rm` statements size should be the same


wemeetagain (approved):
LGTM

g11tech (approved):
LGTM!

PR #4365, opened 2022-08-05
Serve Lightclient related data via p2p

### LINKED ISSUE #3077: Light client integration into p2p network
Integrating consensus light clients into the existing p2p network (rather than solely relying on a REST API) opens the door for light client gossip and other more efficient and decentralized communication.

At a high level, what changes would be required and what should be considered when integrating a light client, or light client functionality, into the p2p network?

Possible outcomes:
- Add websocket libp2p transport to consensus specs
- Transpose lodestar light client REST API to libp2p protocols, specification & implementation
- Explore light client update gossip, specification & implementation
- Explore ENR & MetaData extensions, if required

dadepo:
**Motivation**

Adds ability for a lodestar full node to serve light client related data via p2p. Implements the [Altair Light Client -- Networking](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/p2p-interface.md) spec

**Description**

- [x] Implement the Req/Resp domain
- [x] Implement the gossip domain: gossipsub
- [x] Process all TODOs
- [x] Add tests

Confirmed the new gossip messages get's published. Also confirmed that changes did not break existing lightclient which uses the REST endpoint and SSE:

```
➜  lodestar git:(dadepo/p2p-lightclient) ✗ ./lodestar lightclient --network goerli --logFile none --beaconApiUrl http://localhost:9596 --checkpointRoot 0x1946bf174cc3589b1b38a0bd8374732de84a3328bcf5f7258cb92ac0cabb6dad
Sep-30 19:42:37.140[]                 info: Head updated slot=4004312, root=0x89d90a7a40f3c93c1a094967e137f55dd39a15a6038acd186edee79995e9be64
Sep-30 19:42:49.011[]                 info: Head updated slot=4004313, root=0x18c097fdc52e2d206547072f3647b65e2d306dd852304c51b95a3611f4825312
```

Using a minimalist libp2p node with gossip enabled (see [here](https://github.com/dadepo/lightclient_consumer)), the optimistic update message got picked up:

```
➜  lightclient_consumer git:(master) ✗ node src/index.js /ip4/127.0.0.1/tcp/9000/p2p/16Uiu2HAky1xHb8Zn5WZFZ9UCrSNL1cmnsYMmBJigRKUXHKi1YFvD
started node with peer id: 12D3KooWSu7B4U8RTGhC2eCGyCnMpMec49DRED87kwfW2jMoeoiR
Connection established to: 16Uiu2HAky1xHb8Zn5WZFZ9UCrSNL1cmnsYMmBJigRKUXHKi1YFvD
Got binary data: 152,2,12,78,99,62,0,1,1,8,223,189,4,1,7,244,8,1,0,214,71,36,210,147,38,25,171,48,243,206,9,83,144,102,75,68,156,207,96,7,237,16,226,149,96,219,215,240,191,228,94,122,176,240,36,85,21,9,105,27,87,189,193,46,91,59,47,215,173,29,205,229,25,37,112,123,46,251,36,42,174,227,218,150,157,10,54,177,184,103,61,132,80,135,229,160,212,199,80,102,194,105,111,177,214,86,227,249,137,250,208,33,163,92,180,162,79,238,155,110,255,119,126,115,253,186,103,127,199,250,251,95,111,63,127,173,119,50,211,237,63,239,79,237,220,251,191,239,95,93,222,127,253,229,254,189,189,235,183,255,183,126,255,221,154,126,223,221,207,223,119,175,93,255,127,255,123,244,253,164,217,64,57,50,137,63,75,1,184,207,208,23,44,229,25,43,208,85,210,91,25,85,41,72,222,25,211,244,223,230,35,59,243,20,99,41,98,179,91,154,192,23,40,163,106,49,49,18,214,246,28,6,141,55,225,130,48,107,201,255,87,80,234,199,210,151,253,110,162,246,25,11,142,32,197,118,125,141,127,149,210,125,112,36,34,160,191,3,33,176,166,68,86,189,29,80,99,62,0,0,0,0,0 on topic /eth2/c2ce3aa8/light_client_optimistic_update/ssz_snappy

```

Consuming the light client related data will be implemented in https://github.com/ChainSafe/lodestar/issues/4618

Partially closes #3077


packages/beacon-node/src/chain/lightClient/index.ts
```diff
@@ -455,9 +415,10 @@ export class LightClientServer {
       throw new Error("attested data period different than signature period");
     }
 
-    const headerUpdate: routes.lightclient.LightclientOptimisticHeaderUpdate = {
+    const headerUpdate: altair.LightClientOptimisticUpdate = {
       attestedHeader: attestedData.attestedHeader,
       syncAggregate,
+      signatureSlot: attestedData.attestedHeader.slot, //TODO DA confirm
```
wemeetagain: This should be the slot of the block with the included signature rather than the slot of the header which is signed over.

packages/beacon-node/src/network/reqresp/handlers/lightclientUpdate.ts
```diff
@@ -0,0 +1,9 @@
+import {altair} from "@lodestar/types";
+import {IBeaconChain} from "../../../chain/index.js";
+
+export async function* onLightclientUpdate(
+  requestBody: altair.LightClientUpdateByRangeRequest,
+  chain: IBeaconChain
+): AsyncIterable<altair.LightClientUpdate[]> {
+  yield await chain.lightClientServer.getUpdates(requestBody.startPeriod, requestBody.count);
```
wemeetagain: should be wrapped in try/catch?

packages/beacon-node/src/network/reqresp/reqResp.ts
```diff
@@ -253,7 +253,9 @@ export class ReqResp implements IReqResp {
       case Method.BeaconBlocksByRoot:
         yield* this.reqRespHandlers.onBeaconBlocksByRoot(requestTyped.body);
         break;
-
+      case Method.LightClientBootstrap:
+        yield* this.reqRespHandlers.onLightClientBootstrap(requestTyped.body);
+        break;
```
wemeetagain, line 303: all the rest of the methods should be added here too

packages/beacon-node/src/network/reqresp/request/index.ts
```diff
@@ -109,7 +109,7 @@ export async function sendRequest<T extends IncomingResponseBody | IncomingRespo
     const protocol = protocols.get(protocolId);
     if (!protocol) throw Error(`dialProtocol selected unknown protocolId ${protocolId}`);
 
-    logger.debug("Req  sending request", {...logCtx, body: renderRequestBody(method, requestBody)});
```
wemeetagain: I think the reason there's an extra space here is to align in the logs with `Resp blah`

packages/beacon-node/src/network/reqresp/types.ts
```diff
@@ -52,6 +61,10 @@ export const isSingleResponseChunkByMethod: {[K in Method]: boolean} = {
   [Method.Metadata]: true,
   [Method.BeaconBlocksByRange]: false, // A stream, 0 or more response chunks
   [Method.BeaconBlocksByRoot]: false,
+  [Method.LightClientBootstrap]: true, // TODO DA: confirm
+  [Method.LightClientUpdate]: false, // TODO DA: confirm
+  [Method.LightClientFinalityUpdate]: true, // TODO DA: confirm
+  [Method.LightClientOptimisticUpdate]: true, // TODO DA: confirm
```
wemeetagain: looks good, everything but the LightClientUpdate protocol should be a single response chunk

packages/beacon-node/src/network/reqresp/types.ts
```diff
@@ -92,11 +110,18 @@ export function getRequestSzzTypeByMethod(method: Method) {
...
       return null;
     case Method.BeaconBlocksByRange:
       return ssz.phase0.BeaconBlocksByRangeRequest;
     case Method.BeaconBlocksByRoot:
       return ssz.phase0.BeaconBlocksByRootRequest;
+    case Method.LightClientBootstrap:
+      return ssz.altair.BlockRoot;
+    case Method.LightClientUpdate:
+      return ssz.altair.LightClientUpdate;
+    case Method.LightClientOptimisticUpdate:
```
wemeetagain: Seems this should be null too

packages/types/src/altair/sszTypes.ts
```diff
@@ -205,9 +208,37 @@ export const LightClientUpdate = new ContainerType(
...
+export const LightClientOptimisticUpdate = new ContainerType(
+  {
+    attestedHeader: phase0Ssz.BeaconBlockHeader,
+    syncAggregate: SyncAggregate,
+    signatureSlot: Slot,
+  },
+  {typeName: "LightClientOptimisticUpdate", jsonCase: "eth2"}
+);
+
+export const LightClientByRangeRequest = new ContainerType(
```
wemeetagain: maybe should be named `LightClientUpdateRequest` or at least match the name of the type

wemeetagain (commented):
I know its not quite ready, just giving this a light pass
Looks good so far

packages/beacon-node/src/network/reqresp/request/index.ts
dapplion: Yep

packages/types/src/altair/sszTypes.ts
dadepo: ~the type is actually called `LightClientOptimisticUpdate ` see [here](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/sync-protocol.md#lightclientoptimisticupdate) for reference...or I misunderstood your comment?~

Never mind. Got the comment wrong.

packages/beacon-node/src/network/reqresp/handlers/lightclientUpdate.ts
dadepo: I intentionally left out the try/catch since the spec did not explicitly say `RESOURCE_UNAVAILABLE` should be returned as it did in the other cases. I will look into this again, and if need be, make a PR to the spec.

packages/beacon-node/src/network/reqresp/handlers/index.ts
```diff
@@ -20,13 +28,25 @@ export function getReqRespHandlers({db, chain}: {db: IBeaconDb; chain: IBeaconCh
...
+    },
+    async *onLightClientUpdate(req) {
+      // TODO DA confirm MAX_REQUEST_LIGHT_CLIENT_UPDATES is adhered to.
+      // TODO DA Harmonize the capitalization of lightClient vs lightclient
+      yield* onLightclientUpdate(req, chain);
+    },
+    async *onLightClientFinalityUpdate() {
+      yield* onLightClientFinalityUpdate(chain);
+    },
+    async *onLightClientOptimisitcUpdate() {
```
g11tech: i am not sure a peer would do req/resp over the header as the new header should be easily available over `gossip`, but we can still keep it as the marginal cost to support this is `nil`

packages/beacon-node/src/network/reqresp/handlers/lightclientUpdate.ts
g11tech: it makes sense to have `RESOURCE_UNAVAILABLE` response, may be the spec just missed it

packages/beacon-node/test/e2e/network/network.test.ts
```diff
@@ -247,7 +256,26 @@ describe("network", function () {
 
     expect(netA.gossip.getTopics().length).to.equal(0);
     netA.subscribeGossipCoreTopics();
-    expect(netA.gossip.getTopics().length).to.equal(5);
+    /**
+     * TODO DA confirm
+     * Topics from previous fork are still subscribed together
+     * with same topic of the new fork. Is this expected behaviour?
```
g11tech: sorry what is the confusion here, could you elaborate a bit more pls?

packages/beacon-node/src/network/reqresp/handlers/lightclientUpdate.ts
dadepo: wrapped in try/catch and made a PR to update the spec [here](https://github.com/ethereum/consensus-specs/pull/3013)

packages/beacon-node/test/e2e/network/network.test.ts
dadepo: for example after altair, you have both `'/eth2/f5a5fd42/beacon_block/ssz_snappy',` and `'/eth2/16abab34/beacon_block/ssz_snappy',` subscribed to. The comment was to confirm if this is fine, or if the old topics should be unsubscribed from, when the new topic with the new fork is subscribed to.

g11tech: yes for 2 epochs there is overlap as far as i can remember

packages/beacon-node/src/network/reqresp/types.ts
```diff
@@ -201,6 +250,25 @@ export const reqRespBlockResponseSerializer = {
   },
 };
 
+export const reqRespLightClientUpdateSerializer = {
+  serialize: (chunks: LightClientUpdate[]): Uint8Array => {
```
dadepo: Is there a more idiomatic way to serialise this with ssz?

packages/api/src/beacon/routes/events.ts
```diff
@@ -5,18 +5,6 @@ import {RouteDef, TypeJson} from "../../utils/index.js";
 
 // See /packages/api/src/routes/index.ts for reasoning and instructions to add new routes
 
-export type LightclientOptimisticHeaderUpdate = {
```
dadepo, line 8: This was removed in other to use the ssz types directly...instead of redefining the same types. Is there a reason not to remove and have the types redefined?

packages/api/src/beacon/routes/lightclient.ts
```diff
@@ -102,39 +97,12 @@ export function getReqSerializers(): ReqSerializers<Api, ReqTypes> {
 }
 
 export function getReturnTypes(): ReturnTypes<Api> {
-  const lightclientSnapshotWithProofType = new ContainerType(
```
dadepo, line 105: same here. Any reason to have these types redefined instead of using the ssz types?

packages/api/src/beacon/routes/events.ts
wemeetagain: No thats perfect. I think they were here "temporarily" and then forgotten about

packages/api/src/beacon/routes/lightclient.ts
wemeetagain: Same. Best to just use the ssz types and keep that as the single source of truth

packages/beacon-node/src/network/reqresp/types.ts
wemeetagain: I think we only want to serialize a single chunk at a time here
From the spec:
> The response MUST consist of zero or more response_chunk. Each successful response_chunk MUST contain a single LightClientUpdate payload.

packages/beacon-node/src/network/reqresp/handlers/lightClientUpdate.ts
```diff
@@ -0,0 +1,17 @@
...
+import {ResponseError} from "../response/index.js";
+import {RespStatus} from "../../../constants/index.js";
+
+export async function* onLightclientUpdate(
+  requestBody: altair.LightClientUpdatesByRange,
+  chain: IBeaconChain
+): AsyncIterable<altair.LightClientUpdate[]> {
+  try {
+    const count = Math.min(MAX_REQUEST_LIGHT_CLIENT_UPDATES, requestBody.count);
+    yield await chain.lightClientServer.getUpdates(requestBody.startPeriod, count);
```
wemeetagain: this should be a `yield*` so each update gets yielded individually, rather than an array of updates being yielded.

packages/beacon-node/src/network/reqresp/handlers/lightClientUpdate.ts
```diff
@@ -0,0 +1,17 @@
+import {altair} from "@lodestar/types";
+import {MAX_REQUEST_LIGHT_CLIENT_UPDATES} from "@lodestar/params";
+import {IBeaconChain} from "../../../chain/index.js";
+import {ResponseError} from "../response/index.js";
+import {RespStatus} from "../../../constants/index.js";
+
+export async function* onLightclientUpdate(
+  requestBody: altair.LightClientUpdatesByRange,
+  chain: IBeaconChain
+): AsyncIterable<altair.LightClientUpdate[]> {
```
wemeetagain: should be iterable of updates, rather than array of updates

packages/beacon-node/src/network/reqresp/handlers/lightClientBootstrap.ts
```diff
@@ -0,0 +1,15 @@
...
+import {RespStatus} from "../../../constants/index.js";
+
+export async function* onLightclientBootstrap(
+  requestBody: altair.BlockRoot,
+  chain: IBeaconChain
+): AsyncIterable<altair.LightClientBootstrap> {
+  try {
+    yield await chain.lightClientServer.getBootstrap(requestBody);
+  } catch (e) {
+    throw new ResponseError(RespStatus.RESOURCE_UNAVAILABLE, (e as Error).message);
```
dadepo: Note: The `chain.lightClientServer.getBootstrap` could fail to produce a result for various reasons...this, according to the spec, should lead to an error code 3: ResourceUnavailable. The `(e as Error).message` provides additional textual information what error it was

packages/beacon-node/src/network/reqresp/handlers/lightClientUpdate.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export async function* onLightclientUpdate(
+  requestBody: altair.LightClientUpdatesByRange,
+  chain: IBeaconChain
+): AsyncIterable<altair.LightClientUpdate> {
+  try {
+    const count = Math.min(MAX_REQUEST_LIGHT_CLIENT_UPDATES, requestBody.count);
+    yield* await chain.lightClientServer.getUpdates(requestBody.startPeriod, count);
+  } catch (e) {
+    throw new ResponseError(RespStatus.RESOURCE_UNAVAILABLE, (e as Error).message);
```
dadepo: Note: The `chain.lightClientServer.getUpdates` could fail to produce a result for various reasons...this, according to the spec, should lead to an error code 3: ResourceUnavailable. The `(e as Error).message` provides additional textual information what error it was

packages/beacon-node/src/network/network.ts
```diff
@@ -350,4 +356,20 @@ export class Network implements INetwork {
...
+    } catch (e) {
+      this.logger.debug("Error on BeaconGossipHandler.onLightclientFinalityUpdate", {}, e as Error);
+    }
+  }
+
+  private async onLightclientOptimisticUpdate(optimisticUpdate: altair.LightClientOptimisticUpdate): Promise<void> {
+    try {
+      return await this.gossip.publishLightClientOptimisticUpdate(optimisticUpdate);
+    } catch (e) {
+      this.logger.debug("Error on BeaconGossipHandler.onLightclientOptimisticUpdate", {}, e as Error);
```
dadepo: consciously have this at debug level to prevent unnecessary flooding of logs with `PublishError.InsufficientPeers` which will be the case given there are likely no subscribers to this topics yet

packages/beacon-node/src/network/reqresp/handlers/lightclientUpdate.ts
etan-status: Alternative is to just return an empty result, or omit the periods that are unavailable, same as for beaconBlocksByRange / beaconBlocksByRoot.

packages/beacon-node/src/chain/lightClient/index.ts
```diff
@@ -274,65 +276,27 @@ export class LightClientServer {
    * - Has the most bits
    * - Signed header at the oldest slot
    */
-  async getUpdates(period: SyncPeriod): Promise<altair.LightClientUpdate> {
```
dapplion, line 277: Leave this function as is, and move the code

```ts
    const periods: number[] = Array.from({length: count}, (_ignored, i) => i + startPeriod);
    return await Promise.all(periods.map((period) => this.doGetUpdate(period)));
```

this is very opinionated and each consumer should decide how to handle it

packages/beacon-node/src/chain/validation/lightClientFinalityUpdate.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  // [IGNORE] The received finality_update matches the locally computed one exactly
+  if (lightClientFinalityUpdate != chain.lightClientServer.getFinalityUpdate()) {
+    throw new LightClientError(GossipAction.IGNORE, {
+      code: LightClientErrorCode.FINALITY_UPDATE_NOT_MATCHING_LOCAL,
+    });
+  }
+
+  chain.lightClientServer.latestForwardedFinalitySlot = lightClientFinalityUpdate.finalizedHeader.slot;
```
dapplion: This side-effect here in the middle of a validation function looks really bad. Bad patterns like this are best to be avoided if possible. Since you know that lightClientFinalityUpdate is equal to the one produced locally, just set `latestForwardedFinalitySlot` internally in `lightClientServer` whenever finalityUpdate changes

packages/beacon-node/src/chain/validation/lightClientFinalityUpdate.ts
```diff
@@ -0,0 +1,43 @@
...
+  const timeAtSignatureSlot = computeTimeAtSlot(config, signatureSlot, chain.genesisTime);
+  const currentSlotWithGossipDisparity = chain.clock.currentSlotWithGossipDisparity;
+  if (currentSlotWithGossipDisparity < timeAtSignatureSlot) {
+    throw new LightClientError(GossipAction.IGNORE, {
+      code: LightClientErrorCode.FINALITY_UPDATE_RECEIVED_TOO_EARLY,
+    });
+  }
+
+  // [IGNORE] The received finality_update matches the locally computed one exactly
+  if (lightClientFinalityUpdate != chain.lightClientServer.getFinalityUpdate()) {
```
dapplion: I guess we will never forward anything :stuck_out_tongue_closed_eyes:

packages/beacon-node/src/network/network.ts
twoeths: does a full node always publish an optimistic update message? I see this in the spec

```
Whenever fork choice selects a new head block with a sync aggregate participation >= MIN_SYNC_COMMITTEE_PARTICIPANTS and a post-Altair parent block, full nodes with at least one validator assigned to the current sync committee at the block's slot SHOULD broadcast derived light client data as follows
```

packages/beacon-node/src/chain/validation/lightClientFinalityUpdate.ts
twoeths: @dadepo  the unit test has all negative flows, I think we need to add a positive flow there to cover this case

packages/beacon-node/src/chain/validation/lightClientOptimisticUpdate.ts
```diff
@@ -0,0 +1,44 @@
...
+  const timeAtSignatureSlot = computeTimeAtSlot(config, signatureSlot, chain.genesisTime);
+  const currentSlotWithGossipDisparity = chain.clock.currentSlotWithGossipDisparity;
+  if (currentSlotWithGossipDisparity < timeAtSignatureSlot) {
+    throw new LightClientError(GossipAction.IGNORE, {
+      code: LightClientErrorCode.OPTIMISTIC_UPDATE_RECEIVED_TOO_EARLY,
+    });
+  }
+
+  // [IGNORE] The received optimistic_update matches the locally computed one exactly
+  if (lightClientOptimisticUpdate != chain.lightClientServer.getOptimisticUpdate()) {
```
twoeths: same issue to the finality update, need to use `ssz.equals()` method here, make sure to add a unit test for positive flow to scan through this

packages/beacon-node/src/network/network.ts
twoeths: also it's important not to publish these messages before 1/3 of slot according to the spec
```
These messages SHOULD be broadcasted after one-third of slot has transpired
```

packages/light-client/src/index.ts
```diff
@@ -327,15 +327,11 @@ export class Lightclient {
...
-          controller.signal,
-          this.onSSE
-        );
-        this.api.events.eventstream(
-          [routes.events.EventType.lightclientFinalizedUpdate],
+          [routes.events.EventType.lightClientOptimisticUpdate],
           controller.signal,
           this.onSSE
         );
+        this.api.events.eventstream([routes.events.EventType.lightClientFinalityUpdate], controller.signal, this.onSSE);
```
twoeths: could we do a single `eventstream()` api call?

packages/beacon-node/src/network/gossip/validation/queue.ts
```diff
@@ -17,6 +17,8 @@ const gossipQueueOpts: {[K in GossipType]: Pick<JobQueueOpts, "maxLength" | "typ
   [GossipType.attester_slashing]: {maxLength: 4096, type: QueueType.FIFO},
   [GossipType.sync_committee_contribution_and_proof]: {maxLength: 4096, type: QueueType.LIFO, maxConcurrency: 16},
   [GossipType.sync_committee]: {maxLength: 4096, type: QueueType.LIFO, maxConcurrency: 64},
+  [GossipType.light_client_finality_update]: {maxLength: 1024, type: QueueType.FIFO},
+  [GossipType.light_client_optimistic_update]: {maxLength: 1024, type: QueueType.FIFO},
```
dapplion, line 21: What the reasoning for 1024? This numbers should be researched to be a max of what's necessary for regular operations and limiting memory as much as possible

packages/beacon-node/src/network/reqresp/handlers/lightClientBootstrap.ts
dapplion: This is not okay, getBootstrap may throw for other reasons that are not RESOURCE_UNAVAILABLE. For internal errors it must return ServerError, check the spec

packages/beacon-node/src/network/reqresp/reqResp.ts
```diff
@@ -141,6 +142,46 @@ export class ReqResp implements IReqResp {
...
+  }
+
+  async lightClientUpdate(
+    peerId: PeerId,
+    request: altair.LightClientUpdatesByRange
+  ): Promise<altair.LightClientUpdate[]> {
+    return await this.sendRequest<altair.LightClientUpdate[]>(
+      peerId,
+      Method.LightClientUpdate,
+      [Version.V2],
```
dapplion: Why V2? Please double check your code with the spec

packages/types/src/altair/sszTypes.ts
```diff
@@ -182,13 +182,15 @@ export const BeaconState = new ContainerType(
   {typeName: "BeaconState", jsonCase: "eth2"}
 );
 
-export const LightClientSnapshot = new ContainerType(
+export const BlockRoot = Root;
```
dapplion: This type is redundant just use Root

packages/light-client/src/index.ts
dapplion: Yes, you must do a single eventstream call, doing 2 is very wasteful

packages/db/test/unit/controller/level.test.ts
```diff
@@ -113,6 +113,8 @@ describe("LevelDB controller", () => {
     expect(result.length).to.be.equal(2);
   });
 
+  // TODO Fails on any other platform part from Linux
+  // Issue here https://github.com/ChainSafe/lodestar/issues/4525
```
dapplion: Then add:
```ts
if (!isLinux) this.skip()
```

packages/beacon-node/test/unit/network/reqresp/utils.ts
```diff
@@ -37,6 +37,22 @@ export function generateEmptySignedBlocks(n = 3): phase0.SignedBeaconBlock[] {
...
+
+export function generateDefaultLightClientFinalityUpdate(): altair.LightClientFinalityUpdate {
+  return ssz.altair.LightClientFinalityUpdate.defaultValue();
+}
+
+export function generateDefaultLightClientOptimisticUpdate(): altair.LightClientOptimisticUpdate {
+  return ssz.altair.LightClientOptimisticUpdate.defaultValue();
+}
+
+export function generateDefaultLightClientUpdate(): altair.LightClientUpdate[] {
```
dapplion: Are these helpers really necessary? Why not just write `[ssz.altair.LightClientUpdate.defaultValue()]` in the consumer code?

packages/beacon-node/src/chain/validation/lightClientFinalityUpdate.ts
```diff
@@ -0,0 +1,43 @@
+import {IChainForkConfig} from "@lodestar/config";
+import {altair} from "@lodestar/types";
+import {computeTimeAtSlot} from "@lodestar/state-transition";
+import {IBeaconChain} from "../interface.js";
+import {LightClientError, LightClientErrorCode} from "../errors/lightClientError.js";
+import {GossipAction} from "../errors/index.js";
+
+// https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/p2p-interface.md#light_client_finality_update
+export async function validateLightClientFinalityUpdate(
```
dapplion: Unnecessary async function

packages/beacon-node/test/unit/chain/validation/lightClientFinalityUpdate.test.ts
```diff
@@ -80,9 +84,11 @@ describe("Light Client Finality Update validation", function () {
       return ssz.altair.LightClientFinalityUpdate.defaultValue();
     };
 
-    await expectRejectedWithLodestarError(
-      Promise.resolve(validateLightClientFinalityUpdate(config, chain, lightClientFinalityUpdate)),
```
dapplion: Can you try this syntax to see if it works with `expectRejectedWithLodestarError`?
```ts
await expectRejectedWithLodestarError(
  (async () => validateLightClientFinalityUpdate(config, chain, lightClientFinalityUpdate))(),
  LightClientErrorCode.FINALITY_UPDATE_NOT_MATCHING_LOCAL
)
```

packages/beacon-node/src/chain/lightClient/index.ts
wemeetagain: I think perhaps the chain lightclient module should not even have a `getUpdates` but rather a `getUpdate` since the return type is just a singular `LightClientUpdate`.

packages/beacon-node/src/chain/lightClient/index.ts
```diff
@@ -517,9 +529,11 @@ export class LightClientServer {
    */
   private async maybeStoreNewBestPartialUpdate(
     syncPeriod: SyncPeriod,
-    syncAggregate: altair.SyncAggregate,
+    syncAggregateBlock: altair.BeaconBlock,
```
wemeetagain: Can you just add the signature slot as an additional param here?
Passing in the whole block will prob be annoying to deal with future block types.

packages/beacon-node/src/chain/lightClient/index.ts
g11tech: i would find it useful to req/resp multi updates to get to the current period fast :thinking:

packages/beacon-node/src/network/network.ts
g11tech: i agree, this check should be done before emitting the event at `onSyncAggregate` function, optimistic update is the powerful indicator of where chain might finalize rather than just where the head is,

packages/beacon-node/src/network/reqresp/handlers/lightClientOptimisticUpdate.ts
```diff
@@ -0,0 +1,15 @@
+import {altair} from "@lodestar/types";
+import {IBeaconChain} from "../../../chain/index.js";
+import {ResponseError} from "../response/index.js";
+import {RespStatus} from "../../../constants/index.js";
+
+export async function* onLightClientOptimisticUpdate(
+  chain: IBeaconChain
+): AsyncIterable<altair.LightClientOptimisticUpdate> {
+  const optimisticUpdate = chain.lightClientServer.getOptimisticUpdate();
+  if (optimisticUpdate == null) {
```
g11tech: `===`?

packages/beacon-node/src/chain/lightClient/index.ts
```diff
@@ -501,14 +468,58 @@ export class LightClientServer {
...
         this.metrics?.lightclientServer.onSyncAggregate.inc({event: "update_latest_finalized_update"});
       }
     }
 
     // Check if this update is better, otherwise ignore
-    await this.maybeStoreNewBestPartialUpdate(syncPeriod, syncAggregate, attestedData);
+    await this.maybeStoreNewBestPartialUpdate(syncPeriod, syncAggregate, signatureSlot, attestedData);
+  }
+
+  private async doGetUpdate(period: number): Promise<altair.LightClientUpdate> {
```
wemeetagain: Can you call this `getUpdate`? and it can also be public.

g11tech: +1

packages/beacon-node/src/network/reqresp/handlers/lightClientOptimisticUpdate.ts
dadepo: > `===`?

in this case it should not matter, as its comparison with literal null. But I agree, using type safe comparison is generally a good this. Created a PR to enforce this https://github.com/ChainSafe/lodestar/pull/4647

packages/beacon-node/src/chain/lightClient/index.ts
dadepo: > Can you call this `getUpdate`? and it can also be public.

Doing this will make it a public function that only ever get used within this object as there is no instance where such method call is required by other part of the codebase. Any reason why you will want this?

Also take into consideration that the current use case maps to the query params `?start_period={}&count={}` where even if you need a single update you need to set count to 1.

wemeetagain: Vaguely, the reason for exposing this was: I like interfaces that expose the underlying functionality / primitives as simply as possible, rather than interfaces with higher-level logic, or logic based on the considerations of an external consumer. Ofc its somewhat subjective so :shrug: 
I think this is idea behind this comment too https://github.com/ChainSafe/lodestar/pull/4365#discussion_r987120494

packages/beacon-node/src/chain/lightClient/index.ts
dadepo: > I think perhaps the chain lightclient module should not even have a `getUpdates` but rather a `getUpdate` since the return type is just a singular `LightClientUpdate`.

Do not forget that the lightclient module also serves this rest endpoint, wherein, if the return type is specified as json, it does require `LightClientUpdate[]` to be returned.

packages/beacon-node/src/chain/lightClient/index.ts
dadepo: Gotcha. But it can also be _subjectively_ said that the underlying functionality to be exposed is ability to retrieve the updates, given the `start_period` and `count` needed. If the function returns one singular update as being suggested for `getUpdate`. Then usage will be something like:

```
for n in counts {
   lightClientServer.getUpdate(n)
}
```

instead of

```
 lightClientServer.getUpdates(start_period, counts)
``` 

Which I think is clearer.

I left a comment to the comment you pointed to

packages/beacon-node/test/unit/chain/validation/lightClientFinalityUpdate.test.ts
dadepo: The function being tested is not `asynchronous`. `expectRejectedWithLodestarError` expects an asynchronous input. Your suggestion feels like shoe-horning the test to satisfy  `expectRejectedWithLodestarError`'s need. Is there any benefit to doing that?

packages/beacon-node/src/chain/lightClient/index.ts
g11tech: i think @wemeetagain wants to expose both as public fns, which should be fine.

dadepo: > i think @wemeetagain wants to expose both as public fns, which should be fine.

Indeed. And I asked the reasoning behind that, which he also explained and rightly clarified as partly being subjective. I'll rename this in a bit. It does no harm which ever way

packages/db/test/unit/controller/level.test.ts
dadepo: I think a better approach is to fix the underlying issue. Instead of having platform dependent code that only asserts test behaviour for a particular platform.

packages/beacon-node/src/network/reqresp/handlers/lightClientBootstrap.ts
dadepo: > it must return ServerError, check the spec

The spec here https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/p2p-interface.md#getlightclientbootstrap which I am mainly following for the implementation says nothing about ServerError. Please do share the part of the spec you are referring to.

packages/beacon-node/src/chain/lightClient/index.ts
```diff
@@ -278,7 +278,18 @@ export class LightClientServer {
    */
   async getUpdates(startPeriod: SyncPeriod, count = 1): Promise<altair.LightClientUpdate[]> {
     const periods: number[] = Array.from({length: count}, (_ignored, i) => i + startPeriod);
-    return await Promise.all(periods.map((period) => this.getUpdate(period)));
+    const updates = [];
+    for (const period of periods) {
+      try {
+        updates.push(this.getUpdate(period));
```
dapplion: Why wrapping in a try catch if this is not awaited?

packages/beacon-node/src/network/reqresp/handlers/lightClientUpdate.ts
```diff
@@ -1,17 +1,11 @@
...
   chain: IBeaconChain
 ): AsyncIterable<altair.LightClientUpdate> {
-  try {
-    const count = Math.min(MAX_REQUEST_LIGHT_CLIENT_UPDATES, requestBody.count);
-    yield* await chain.lightClientServer.getUpdates(requestBody.startPeriod, count);
-  } catch (e) {
-    throw new ResponseError(RespStatus.RESOURCE_UNAVAILABLE, (e as Error).message);
-  }
+  const count = Math.min(MAX_REQUEST_LIGHT_CLIENT_UPDATES, requestBody.count);
+  yield* await chain.lightClientServer.getUpdates(requestBody.startPeriod, count);
```
dapplion: This yield* is given the false impression of streaming. getUpdates implementation will load all items in memory, and only then return the array. Why not load each update here and yield individually?

packages/beacon-node/src/network/reqresp/handlers/lightClientUpdate.ts
```diff
@@ -7,5 +7,7 @@ export async function* onLightclientUpdate(
   chain: IBeaconChain
 ): AsyncIterable<altair.LightClientUpdate> {
   const count = Math.min(MAX_REQUEST_LIGHT_CLIENT_UPDATES, requestBody.count);
-  yield* await chain.lightClientServer.getUpdates(requestBody.startPeriod, count);
+  for (const period of Array.from({length: count}, (_ignored, i) => i + requestBody.startPeriod)) {
```
dapplion: Just use a regular for loop

packages/beacon-node/src/network/network.ts
```diff
@@ -123,6 +123,8 @@ export class Network implements INetwork {
     );
 
     this.chain.emitter.on(ChainEvent.clockEpoch, this.onEpoch);
+    this.chain.emitter.on(ChainEvent.lightClientFinalityUpdate, this.onLightclientFinalityUpdate.bind(this));
+    this.chain.emitter.on(ChainEvent.lightClientOptimisticUpdate, this.onLightclientOptimisticUpdate.bind(this));
```
dapplion: You must unbind from foreign class emitter on close

packages/beacon-node/src/network/network.ts
```diff
@@ -350,4 +358,24 @@ export class Network implements INetwork {
       }
     }
   };
+
+  private async onLightclientFinalityUpdate(finalityUpdate: altair.LightClientFinalityUpdate): Promise<void> {
```
wemeetagain: make these methods like:
```ts
private onBlah = async (...): Promise<...> => {}
```
so that you don't have to do `.bind(this)` in the constructor

packages/beacon-node/src/network/network.ts
```diff
@@ -350,4 +358,24 @@ export class Network implements INetwork {
...
+    try {
+      // messages SHOULD be broadcasted after one-third of slot has transpired
+      await this.clock.waitForSlot(finalityUpdate.signatureSlot + 1 / 3);
```
wemeetagain: I don't think this does what you're hoping, this method only works with integer values. Needs to be done a different way.

packages/beacon-node/src/network/network.ts
```diff
@@ -350,4 +358,24 @@ export class Network implements INetwork {
...
+      return await this.gossip.publishLightClientFinalityUpdate(finalityUpdate);
```
wemeetagain: I believe according to spec these messages should only be published when we have a sync committee participant.

wemeetagain (changes requested):

packages/beacon-node/src/network/network.ts
dadepo: Indeed. I should think that has been covered [here](https://github.com/ChainSafe/lodestar/pull/4365/files#diff-4f9e095020a315582142d159413da6239fbd40c87ce2acf1f400dec9f8bee79cR476) where the actual publication occurs.

packages/beacon-node/src/network/network.ts
dadepo: Updated the implementation and also moved to `onSyncAggregate` where the updates are actually emitted. Let me know if that works fine.

packages/beacon-node/src/network/network.ts
wemeetagain: I don't see that. The link isn't working for me.

Looking for the code that that implements this part:
https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/p2p-interface.md#sync-committee
Specifically this part
> full nodes with at least one validator assigned to the current sync committee at the block's slot SHOULD broadcast

packages/beacon-node/src/chain/lightClient/index.ts
```diff
@@ -494,21 +508,32 @@ export class LightClientServer {
...
         this.metrics?.lightclientServer.onSyncAggregate.inc({event: "update_latest_finalized_update"});
       }
     }
 
+    // Emit update
+    // messages SHOULD be broadcast after one-third of slot has transpired
+    // https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/p2p-interface.md#sync-committee
+    const minPubTime = computeTimeAtSlot(this.config, signatureSlot, genesisTime) + this.config.SECONDS_PER_SLOT / 3;
+    const waitTime = minPubTime - Date.now() / 1000;
+    await sleep(waitTime);
```
wemeetagain: This sleep shouldn't happen here, should happen where you had it before. We don't want the pause in the beacon API. We only need it for p2p.

packages/beacon-node/src/network/network.ts
```diff
@@ -350,4 +358,34 @@ export class Network implements INetwork {
...
+      return await this.gossip.publishLightClientOptimisticUpdate(optimisticUpdate);
+    } catch (e) {
+      this.logger.debug("Error on BeaconGossipHandler.onLightclientOptimisticUpdate", {}, e as Error);
+    }
+  };
+
+  private waitOneThirdOfSlot = async (slot: number): Promise<void> => {
+    const minPubTime = computeTimeAtSlot(this.config, slot, this.chain.genesisTime) + this.config.SECONDS_PER_SLOT / 3;
+    const waitTime = minPubTime - Date.now() / 1000;
+    await sleep(waitTime);
```
wemeetagain: I think sleep takes a second argument, an abort signal. Can you add the signal from the constructor here?
Without that, if we're sleeping here and we exit lodestar, we may still have to sleep the full duration.

packages/beacon-node/src/network/network.ts
wemeetagain: bump on this. There should be a conditional here like:
```ts
if (this.hasAttachedSyncCommitteeMember()) {
  sleep(..)
  publish(..)
}
```

dadepo: I have seen how sync committee participation is calculated in the code base, but not come across where we check if a validator attached to the node is part of the current sync committee...

I know the validator registry can be found in the state...if I were to implement this check, my approach will be to get access to the validator registry somehow and check against the current sync committee. Will this be the most optimal approach? If there are parts of the code base where similar check is already being performed, please do share.

One meta question: The full node has already processed the block, has been able to create the lc update and also confirm there is enough sync committee participation. Do you know the rationale for this extra check?

wemeetagain: I think the rationale for this condition is to avoid flooding the message on the network. Only max 512 nodes will initially publish the message, rather than every full-node on the network.

Look at `beacon-node/src/network/subnets/syncnetsService.ts`. You can use `SyncnetsService#getActiveSubnets()`. If the returned array is length > 0 then we can assume we have an attached sync committee member.

dadepo: Thanks for the clarification.

wemeetagain: this isn't resolved yet

dadepo: opsie. I forgot to push. It is pushed now.

packages/beacon-node/src/network/network.ts
```diff
@@ -350,4 +360,46 @@ export class Network implements INetwork {
...
+      } catch (e) {
+        this.logger.debug("Error on BeaconGossipHandler.onLightclientOptimisticUpdate", {}, e as Error);
+      }
+    }
+  };
+
+  private waitOneThirdOfSlot = async (slot: number): Promise<void> => {
+    const minPubTime = computeTimeAtSlot(this.config, slot, this.chain.genesisTime) + this.config.SECONDS_PER_SLOT / 3;
+    const waitTime = minPubTime - Date.now() / 1000;
+    await sleep(waitTime, this.signal);
```
dapplion: sleep takes miliseconds, looks like it's passing seconds which would cause the sleep to wait 1000 times less than expected. Fixing on a commit locally, will push soon

wemeetagain:
I pushed some commits on top addressing some unaddressed comments to help expedite this.

dapplion (approved):
LGMT, some comments for future work:

- Do not re-calculate `light_client_updates_by_range` replies every time. Store in DB an SSZ compressed version of that data and just stream. This optimization should be done also for block by range
- Consider moving lightclient class outside of Chain so it has access to network. It's now really something like the sync class
- Should the lightclient server validate messages before broadcasting?
- Emit logic from the lightclient server provider probably requires more validation
- Apply `is_better_update` consistently in our lightclient data generator code, ref https://github.com/ethereum/consensus-specs/blob/be3c774069e16e89145660be511c1b183056017e/specs/altair/light-client/sync-protocol.md#is_better_update

dapplion (approved):

wemeetagain (approved):

Issue #3998, opened 2022-05-10
Remove RootCache


dapplion:
RootCache is an optimization that prevent having to walk the tree repeatedly when processing attestations.

After merging https://github.com/ChainSafe/lodestar/pull/3760 the optimization may not be necessary anymore:

- [x] Add performance test with current impl and RootCache https://github.com/ChainSafe/lodestar/pull/4374
- [ ] Remove RootCache and ensure performance is as good or better https://github.com/ChainSafe/lodestar/pull/4375

dapplion:
After discussing with @tuyennhv RootCache still provides a noticeable performance improvement even with ssz v2, so we are keeping the optimization

PR #4374, opened 2022-08-06
Add RootCache performance test

dapplion:
**Motivation**

RootCache may become unnecessary after merging SSZ v2.

- From https://github.com/ChainSafe/lodestar/issues/3998

**Description**

- Add performance test for RootCache
- Move RootCache out of registerAttestationInBlock

Locally I get

```
  RootCache.getBlockRootAtSlot
    ✔ RootCache.getBlockRootAtSlot - 250000 vs - 7PWei                 1.164144e+8 ops/s    8.590000 ns/op        -     573120 runs   7.15 s
    ✔ state getBlockRootAtSlot - 250000 vs - 7PWei                        974725.4 ops/s    1.025930 us/op        -       7795 runs   1.47 s
```

RootCache is faster (always will be since it's just a Map.get). But state is somewhat fast too. `1us` * 128 attestations per block is `0.13 ms` of processing time that would be added if we remove the RootCache. @tuyennhv what do you think?

### DIFF
--- a/packages/beacon-node/src/chain/blocks/importBlock.ts
+++ b/packages/beacon-node/src/chain/blocks/importBlock.ts
@@ -110,33 +110,31 @@ export async function importBlock(
   if (!opts.skipImportingAttestations && blockEpoch >= currentEpoch - FORK_CHOICE_ATT_EPOCH_LIMIT) {
     const attestations = block.message.body.attestations;
     const rootCache = new RootCache(postState);
-    const parentSlot = chain.forkChoice.getBlock(block.message.parentRoot)?.slot;
     const invalidAttestationErrorsByCode = new Map<string, {error: Error; count: number}>();
 
     for (const attestation of attestations) {
       try {
         const indexedAttestation = postState.epochCtx.getIndexedAttestation(attestation);
-        const targetEpoch = attestation.data.target.epoch;
+        const {target, slot, beaconBlockRoot} = attestation.data;
 
         const attDataRoot = toHexString(ssz.phase0.AttestationData.hashTreeRoot(indexedAttestation.data));
         chain.seenAggregatedAttestations.add(
-          targetEpoch,
+          target.epoch,
           attDataRoot,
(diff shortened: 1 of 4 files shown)


wemeetagain (approved):

twoeths:
> RootCache is faster (always will be since it's just a Map.get). But state is somewhat fast too. 1us * 128 attestations per block is 0.13 ms of processing time that would be added if we remove the RootCache. @tuyennhv what do you think?

@dapplion I think ssz-v2 should have same mechanism to cache loaded nodes, do you know why it takes more time than RootCache? I'd drop it only when the performances between the two are comparable

dapplion:
> I think ssz-v2 should have same mechanism to cache loaded nodes, do you know why it takes more time than RootCache?

To get a root each impl has to:
- ssz v1: traverse the state tree, traverse blockRoots tree, deserialize node to root
- ssz v2: use cache to jump directly to root node, deserialize node to root
- RootCache: single Map.get()

RootCache will **always** be faster, since it doesn't have to deserialize anything. I don't think it's ever possible for the ssz impl to be as fast as the RootCache. The question is if the optimization justifies having a RootCache there.

twoeths:
> > I think ssz-v2 should have same mechanism to cache loaded nodes, do you know why it takes more time than RootCache?
> 
> To get a root each impl has to:
> 
> * ssz v1: traverse the state tree, traverse blockRoots tree, deserialize node to root
> * ssz v2: use cache to jump directly to root node, deserialize node to root
> * RootCache: single Map.get()
> 
> RootCache will **always** be faster, since it doesn't have to deserialize anything. I don't think it's ever possible for the ssz impl to be as fast as the RootCache. The question is if the optimization justifies having a RootCache there.

I see. Tbh, I still don't feel convenient for any changes that's slower 😃 
- RootCache is also used for validatorMonitor metric
- This is all about calling `hashObjectToByteArray()` 128 times (and with validatorMonitor, it's 256) for the same data in same flow vs calling it once and cache it, I'd go with the 2nd

PR #4375, opened 2022-08-06
Remove root cache

dapplion:
**Motivation**

RootCache may become unnecessary after merging SSZ v2.

- From https://github.com/ChainSafe/lodestar/issues/3998
- See https://github.com/ChainSafe/lodestar/pull/4374 for a performance test

Marking as draft to decide first if this is a good idea

**Description**

- Remove RootCache


dapplion:
After discussing with @tuyennhv RootCache still provides a noticeable performance improvement even with ssz v2, so we are keeping the optimization

PR #4377, opened 2022-08-06
Use BeaconChain this in processBlock fns

dapplion:
**Motivation**

Process block functions have a huge list of dependencies, since it's a complex flow that requires interactions of many dependencies of the class `BeaconChain`. Seeing that huge list of dependencies I think that those functions should really be a private member of the BeaconChain class.

https://github.com/ChainSafe/lodestar/blob/1da9558e078af9c0f42c62d446a70b28162e1f08/packages/beacon-node/src/chain/chain.ts#L211-L237

At the very least they should be "external binded protected" properties to modularize the code.

**Description**

Replace dependency injection "chain" variable with actual this BeaconChain class.
- If we want to keep this functions in separate classes we can just use this approach
- If we don't want to have to pass around a this object let's just move this code to BeaconChain class body. Importing a block is probably the most important task that the class does, so it makes sense to be there in its body


wemeetagain (approved):
LGTM, good change

Issue #2015, opened 2021-01-31
Error processing finalized checkpoint


twoeths:
I got this when syncing pyrmont locally:

```
2021-01-31 20:36:17 [SYNC]             info: Sync progress 1957/16662 - 1.1 days left - 5.08 slots/s
2021-01-31 20:36:18 [CHORES]          error: Error processing finalized checkpoint epoch=1953 code=PROTO_ARRAY_ERROR_FINALIZED_NODE_UNKNOWN, root=0x631413d98355f04a73d701e8c19c3428fc1ef6702f032dc125883fea65f1e814
Error: PROTO_ARRAY_ERROR_FINALIZED_NODE_UNKNOWN
    at ProtoArray.maybePrune (/Users/tuyennguyen/Projects/workshop/lodestar/packages/lodestar-fork-choice/src/protoArray/protoArray.ts:241:13)
    at LodestarForkChoice.prune (/Users/tuyennguyen/Projects/workshop/lodestar/packages/lodestar-fork-choice/src/forkChoice/forkChoice.ts:488:28)
    at ChainEventEmitter.<anonymous> (/Users/tuyennguyen/Projects/workshop/lodestar/packages/lodestar/src/tasks/index.ts:108:29)
2021-01-31 20:36:32 [CHAIN]            info: Checkpoint justified epoch=1959, root=0x85a38819f760677973679cb24892959e521b46949a6acd54957cc5ce14636d1b
```

dapplion:
I've seen it too!

3xtr4t3rr3str14l:
> I've seen it too!

ditto

twoeths:
this is because we don't manage the order of archived tasks
+ archived task `n` runs and hasn't finished - maybe because it has to archive state
+ archived task `n+1` runs and finish - maybe because it doesn't have to archive state, it deletes all nodes all the way to its finalized node including the finalized node of task `n`
+ archived task reaches its last step which is `this.chain.forkChoice.prune(finalized.root);` and found that it's finalized root was deleted (by archived task `n+1`)

for this specific error, it causes no harm but we may have bigger issue after a big unfinalized period. To completely resolve this, I think we need to use `JobQueue` to order the execution of archived tasks.

PR #4379, opened 2022-08-06
Move archive logic to BeaconChain class

### LINKED ISSUE #4378: Archive states uses epoch as a key instead of slot
**Describe the bug**

In `maybeArchiveState` function epoch is used a key instead of slot. This probably means that Lodestar is not pruning states are expected having a larger db than what we designed

https://github.com/ChainSafe/lodestar/blob/1da9558e078af9c0f42c62d446a70b28162e1f08/packages/beacon-node/src/chain/archiver/archiveStates.ts#L48-L52

**Expected behavior**

Index archiveStates repo by slot

### LINKED ISSUE #4417: StateArchiver use epoch as key instead of slot
**Describe the bug**

In StateArchiver logic to prune states, epoch is used a key to index and prune states, which are actually indexed by slot

https://github.com/ChainSafe/lodestar/blob/db9450ba2ee4ac90469999450a16d143f6127961/packages/beacon-node/src/chain/archiver/archiveStates.ts#L48-L52

Current parameters in to persists "recent" states every 32 epochs, older states every 1024 epochs

https://github.com/ChainSafe/lodestar/blob/db9450ba2ee4ac90469999450a16d143f6127961/packages/beacon-node/src/chain/archiver/archiveStates.ts#L30-L39

However, now Lodestar is persisting states every 32 epochs always. Since states are reasonably heavy for Mainnet and Prater (~40MB) that means (82187 epochs in 1 year)

| Store state every N epochs | State DB growth per year |
| - | - |
| every 32 | 102 GB |
| every 1024 | 3.2 GB |

**Expected behavior**

Apply designed parameters, index by slot

dapplion:
**Motivation**

Having archive logic far away from BeaconChain class has the negative side-effect of complicating logic and making it harder to provide flow control guarantees that don't break prunning. The motivation is similar to these:
- https://github.com/ChainSafe/lodestar/pull/4224
- https://github.com/ChainSafe/lodestar/pull/3221

For example, having a flow more controlled, not dependant on event handler interactions would have prevent this issue without needing a JobQueue

- https://github.com/ChainSafe/lodestar/issues/2015

**Description**

Move code to archive blocks and states into BeaconChain class, to be triggered through the event handlers already present there.

- Get fork-choice state and finalized state from cache immediately sync, then interact with DBs async
- Fix issue https://github.com/ChainSafe/lodestar/issues/4378

Closes https://github.com/ChainSafe/lodestar/issues/4378
Closes https://github.com/ChainSafe/lodestar/issues/4417


packages/beacon-node/src/chain/chain.ts
```diff
@@ -567,6 +571,68 @@ export class BeaconChain implements IBeaconChain {
     if (headState) {
       this.opPool.pruneAll(headState);
     }
+
+    if (!this.opts.disableArchiveOnCheckpoint) {
+      // Archive state and blocks
+      // TODO: When is the best time to perform this actions?
+      // TODO: Only perform one archive at a time
```
wemeetagain, line 568: Doesn't this reintroduce https://github.com/ChainSafe/lodestar/issues/2015?

dapplion: It should not because all fork-choice ops are done before any async call in `archiveFinalizedStateAndBlocks`. So because we have a queue that ensures only one block is imported at once, fork-choice .prune() won't be called with stale data

wemeetagain (dismissed):
LGTM

packages/beacon-node/src/chain/archiver/archiveStates.ts
```diff
@@ -15,77 +16,70 @@ export const PERSIST_STATE_EVERY_EPOCHS = 1024;
...
-  for (const epoch of storedEpochs) {
-    const epochBucket = epoch - (epoch % persistEveryEpochs);
-    if (epochBuckets.has(epochBucket)) {
-      toDelete.add(epoch);
+export function computeStateSlotsToDelete(storedStateSlots: number[], persistEveryEpochs: number): number[] {
+  const persistEverySlots = persistEveryEpochs * SLOTS_PER_EPOCH;
+  const intervalsWithStates = new Set<number>();
+  const stateSlotsToDelete = new Set<number>();
+
+  for (const epoch of storedStateSlots) {
```
twoeths: variable name should be `slot` instead of `epoch`

packages/beacon-node/src/chain/chain.ts
```diff
@@ -557,6 +561,68 @@ export class BeaconChain implements IBeaconChain {
...
+
+    // Move canonical blocks to cold db, delete non-canonical blocks
+    await archiveBlocks(this.db, this.lightClientServer, this.logger, finalized, {
+      finalizedCanonicalBlocks,
+      finalizedNonCanonicalBlocks,
+    });
+
+    if (finalizedState) {
+      // should be after ArchiveBlocksTask to handle restart cleanly
+      const {archivedState, deletedSlots: deletedEpochs} = await maybeArchiveState(this.db, finalizedState, finalized);
```
twoeths: the naming is confused, should be `deletedSlots`, looks like a refactor issue

twoeths (approved):

PR #4380, opened 2022-08-06
Move produceBlock main function to BeaconChain class

dapplion:
**Motivation**

- Very similar motivation to https://github.com/ChainSafe/lodestar/pull/4379

The current code organization of this files is extremely old (from almost 3y ago), and requires many dependencies of BeaconChain to be made public when they shouldn't.

**Description**

- Move produceBlock main function to BeaconChain class
- Leave produceBlockBody (which has all the complex logic) modularized in a different file

@g11tech I'm getting a strange type error where Typescript considers the return value of the function different depending if it's from the interface or class. Could you investigate?


g11tech (dismissed):

g11tech:
working on this will update by tomorrow

g11tech (approved):
LGTM

PR #4385, opened 2022-08-07
Reduce one pipe level in ReqResp handler

dapplion:
**Motivation**

- Spin-off from https://github.com/ChainSafe/lodestar/pull/4345 with a useful change

**Description**

- Reduce one pipe level in ReqResp handler
- Move timeout test to main ReqResp handler function


wemeetagain:
```
@lodestar/beacon-node: src/network/reqresp/request/index.ts:59:44 - error TS2345: Argument of type 'string' is not assignable to parameter of type 'PeerId'.
@lodestar/beacon-node: 59   const peerIdStrShort = prettyPrintPeerId(peerIdStr);
```

wemeetagain (approved):

PR #4399, opened 2022-08-11
Add option to run beacon node in follow mode to EL

g11tech:
**Motivation**
We in our test/dev setups want to connect multiple beacon nodes to same EL,  the only complication for this is multiple EL requesting head change to the EL causing unknown EL head transitions and issues.

This PR enables to run beacon node in "follow" mode to an EL, which will be driven by only one beacon node in the normal way.
Assuming the EL and its driving beacon node are synced well, this should work out well for other bns as well as they will also be able to follow heads as well as propose when its time for their validators.

This is achieved by adding a flag which can be turned on in the follow nodes `disableImportExecutionFcU` which will turn off giving head fcU to the EL.


g11tech:
deployed on `unstable` dev nodes, all synced up sharing the same EL 
image

validators also attesting (no block proposal yet)
image

wemeetagain (approved):

g11tech:
Block production also looks successful:
image

g11tech:
merging since all aspects look good!

PR #4401, opened 2022-08-12
Reduce heap memory

### LINKED ISSUE #4397: 0.42.0: Heap memory increased
**Describe the bug**

Heap memory increased in v0.42.0

Screen Shot 2022-08-11 at 14 59 58

in v0.41.0
Screen Shot 2022-08-11 at 15 00 59

it's reproducable by this state transition code

```typescript
import fs from "fs";
import {createIBeaconConfig, createIChainForkConfig} from "@lodestar/config";
import {praterChainConfig as chainConfig} from "@lodestar/config/networks";
import {stateTransition, createCachedBeaconState, PubkeyIndexMap} from "@lodestar/state-transition";
import {StateContextCache} from "../../src/chain/stateCache/stateContextCache.js";

global.gc?.();
const cache = new StateContextCache({metrics: null});

const config = createIChainForkConfig(chainConfig);

const path = "/Users/tuyennguyen/tuyen/memTest";

const firstState = config.getForkTypes(3628800).BeaconState.deserializeToViewDU(fs.readFileSync(`${path}/state_3628800.ssz`));

// similar to chain.ts
let state = createCachedBeaconState(firstState, {
  config: createIBeaconConfig(config, firstState.genesisValidatorsRoot),
  pubkey2index: new PubkeyIndexMap(),
  index2pubkey: [],
});

cache.add(state);

// const altairState = state as BeaconStateAltair;
// console.log("@@@ sync committee", altairState.currentSyncCommittee);

for (let n = 0; n < 7; n++) {
  for (let i = 0; i < 10; i++) {
    // skip 3628800
    if (n === 0 && i === 0) continue;
    const blockFilePath = `${path}/block_36288${n}${i}.json`;
    if (fs.existsSync(blockFilePath)) {
      const blockBytes = fs.readFileSync(blockFilePath);
      const str = new TextDecoder().decode(blockBytes);
      const json = JSON.parse(str);
      const signedBlock = config.getForkTypes((json.data.message.slot as unknown) as number).SignedBeaconBlock.fromJson(json.data);
      state = stateTransition(state, signedBlock, {
        verifyProposer: false,
        verifySignatures: false,
        verifyStateRoot: true,
      });
      cache.add(state);
      console.log("@@@ successfully process block", signedBlock.message.slot, "cache size", cache.size);
    }
  }
}

const memoryUsage = process.memoryUsage();
console.log("@@@ memory used in MB:", memoryUsage.heapTotal / 1e6);
```

after running through different commits, it seems it happened from [`5c90db69c6706b37f01d4ef242150b8c2563234c`](https://github.com/ChainSafe/lodestar/pull/4294)

**Expected behavior**

No memory issue

twoeths:
**Motivation**

This PR #4294 causes the memory to increased a lot, the reason is we access `state.validators.get()` on every indexed attestations (20_000 times per block), see https://github.com/ChainSafe/lodestar/issues/4397#issuecomment-1212652592

**Description**

- Cache `slashed` property of validator in EpochContext, clone at every block only if needed
- The performance between BitArray vs boolean[]:
  - On read: 1.5x slower
  - On write: 5x slower
  - On clone: 60x faster

Pick BitArray because we mostly read (20_000 per block) and rarely write (max 16 + 2 = 18 slashed validators per block). Although the read is slower than boolean[], it should be faster than tree backed traversal, and the `clone()` is great

- Also wrap it in a class called `LazyBitArray` which only clone the BitArray on the first set. So on a block with no slashed validators (most of the time), we even don't need to do a `clone()` on every block, only build the `slashedArr` at epoch transition

Closes #4397

**Test**
- [x] The memory test shows it consumed 0-30MB more than v0.41.0 (vs ~500MB in the case of v0.42.0, or unstable)
- [ ] Deploy and test on a prater node

### DIFF
--- a/packages/beacon-node/test/memory/stateTransitionMemTest.ts
+++ b/packages/beacon-node/test/memory/stateTransitionMemTest.ts
@@ -1,0 +1,50 @@
+import fs from "fs";
+import {createIBeaconConfig, createIChainForkConfig} from "@lodestar/config";
+import {praterChainConfig as chainConfig} from "@lodestar/config/networks";
+import {stateTransition, createCachedBeaconState, PubkeyIndexMap} from "@lodestar/state-transition";
+import {StateContextCache} from "../../src/chain/stateCache/stateContextCache.js";
+
+global.gc?.();
+const cache = new StateContextCache({metrics: null});
+
+const config = createIChainForkConfig(chainConfig);
+
+const path = "/Users/tuyennguyen/tuyen/memTest";
+
+const firstState = config.getForkTypes(3628800).BeaconState.deserializeToViewDU(fs.readFileSync(`${path}/state_3628800.ssz`));
+
+// similar to chain.ts
+let state = createCachedBeaconState(firstState, {
+  config: createIBeaconConfig(config, firstState.genesisValidatorsRoot),
+  pubkey2index: new PubkeyIndexMap(),
+  index2pubkey: [],
+});
+
+cache.add(state);
+
+// const altairState = state as BeaconStateAltair;
+// console.log("@@@ sync committee", altairState.currentSyncCommittee);
+
+for (let n = 0; n < 7; n++) {
+  for (let i = 0; i < 10; i++) {
+    // skip 3628800
+    if (n === 0 && i === 0) continue;
+    const blockFilePath = `${path}/block_36288${n}${i}.json`;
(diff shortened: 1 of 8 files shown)


twoeths:
replaced by https://github.com/ChainSafe/lodestar/pull/4404

PR #4414, opened 2022-08-15
Add error handling to exchangeTransitionConfiguration

dapplion:
**Motivation**

- PR https://github.com/ChainSafe/lodestar/pull/4411 introduced exchangeTransitionConfiguration checks

However it has some issues:

- There is no error handling, and errors are very likely to happen here
- Geth returns an error if there's a mismatch so the comparision code would never be reached
- We should not polute the prepareNextSlot unnecessarily

**Description**

- Add error handling to the exchangeTransitionConfiguration calls
- Log that config is validated once per consecutive success calls
- Move code to BeaconChain

Still I fear this will contribute to our bad error / logging UX.
- How would this look like when the EL node is disconnected?
- For Geth's case, do RPC errors render nice enough to print the value from Geth?


g11tech (approved):
LGTM!

wemeetagain (approved):

PR #4430, opened 2022-08-17
Consider state slot in getDomain

### LINKED ISSUE #4428: Only include voluntaryExit in block for curr or prev forks
**Describe the bug**

In 

https://github.com/ChainSafe/lodestar/blob/7703607b53616d83550ac9b2e04ae55cbe43e474/packages/beacon-node/src/chain/opPools/opPool.ts#L184

we must check that voluntaryExit epoch is within current or previous forks. Else other nodes will reject the block as get_domain() can only verify signatures for current or previous blocks.

**Second issue**

Our getDomain implementation should be careful to consider only the current and previous forks as per the spec

```py
def get_domain(state: BeaconState, domain_type: DomainType, epoch: Epoch=None) -> Domain:
    """
    Return the signature domain (fork version concatenated with domain type) of a message.
    """
    epoch = get_current_epoch(state) if epoch is None else epoch
    fork_version = state.fork.previous_version if epoch < state.fork.epoch else state.fork.current_version
    return compute_domain(domain_type, fork_version, state.genesis_validators_root)
```

https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#get_domain

wemeetagain:
Resolves https://github.com/ChainSafe/lodestar/issues/4428

Add a `stateSlot: Slot` parameter to `getDomain`. This matches the ergonomics of the consensus function.
Note: `getDomain` still relies on the configured fork schedule rather than the on-chain fork schedule (`state.fork`)

### DIFF
--- a/packages/config/src/genesisConfig/index.ts
+++ b/packages/config/src/genesisConfig/index.ts
@@ -1,4 +1,4 @@
-import {ForkName} from "@lodestar/params";
+import {ForkName, SLOTS_PER_EPOCH} from "@lodestar/params";
 import {DomainType, ForkDigest, phase0, Root, Slot, ssz, Version} from "@lodestar/types";
 import {toHexString} from "@chainsafe/ssz";
 import {IChainForkConfig} from "../beaconConfig.js";
@@ -22,8 +22,13 @@ export function createICachedGenesis(chainForkConfig: IChainForkConfig, genesisV
   }
 
   return {
-    getDomain(domainType: DomainType, slot: Slot): Uint8Array {
-      const forkInfo = chainForkConfig.getForkInfo(slot);
+    getDomain(stateSlot: Slot, domainType: DomainType, messageSlot?: Slot): Uint8Array {
+      const stateForkInfo = chainForkConfig.getForkInfo(stateSlot);
+      const epoch = Math.floor(messageSlot ?? stateSlot / SLOTS_PER_EPOCH);
+      const forkInfo =
+        epoch < stateForkInfo.epoch
+          ? chainForkConfig.forksAscendingEpochOrder[Math.max(stateForkInfo.seq - 1, 0)] // previous
+          : stateForkInfo; // current
       let domainByType = domainCache.get(forkInfo.name);
       if (!domainByType) {
         domainByType = new Map<DomainType, Uint8Array>();
--- a/packages/config/src/genesisConfig/index.ts
+++ b/packages/config/src/genesisConfig/index.ts
@@ -23,12 +23,23 @@ export function createICachedGenesis(chainForkConfig: IChainForkConfig, genesisV
 
   return {
     getDomain(stateSlot: Slot, domainType: DomainType, messageSlot?: Slot): Uint8Array {
-      const stateForkInfo = chainForkConfig.getForkInfo(stateSlot);
+      // ```py
+      // def get_domain(state: BeaconState, domain_type: DomainType, epoch: Epoch=None) -> Domain:
+      //   """
+      //   Return the signature domain (fork version concatenated with domain type) of a message.
+      //   """
+      //   epoch = get_current_epoch(state) if epoch is None else epoch
+      //   fork_version = state.fork.previous_version if epoch < state.fork.epoch else state.fork.current_version
+      //   return compute_domain(domain_type, fork_version, state.genesis_validators_root)
+      // ```
+
       const epoch = Math.floor(messageSlot ?? stateSlot / SLOTS_PER_EPOCH);
-      const forkInfo =
-        epoch < stateForkInfo.epoch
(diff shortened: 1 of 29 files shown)


packages/config/src/genesisConfig/index.ts
```diff
@@ -22,8 +22,19 @@ export function createICachedGenesis(chainForkConfig: IChainForkConfig, genesisV
   }
 
   return {
-    getDomain(domainType: DomainType, slot: Slot): Uint8Array {
-      const forkInfo = chainForkConfig.getForkInfo(slot);
+    getDomain(fork: phase0.Fork | undefined, domainType: DomainType, slot: Slot): Uint8Array {
```
dapplion: It has to be an optional parameter why not putting it last? 
```suggestion
    getDomain(domainType: DomainType, slot: Slot, stateFork?: phase0.Fork): Uint8Array {
```

packages/config/src/genesisConfig/types.ts
```diff
@@ -11,5 +11,10 @@ export interface IForkDigestContext {
 }
 
 export interface ICachedGenesis extends IForkDigestContext {
-  getDomain(domainType: DomainType, slot: Slot): Uint8Array;
+  /**
+   * Return the signature domain (fork version concatenated with domain type) of a message.
+   *
+   * @param fork The domain will be selected the previous or current version if provided. Otherwise, the chain config fork schedule is used.
```
dapplion: Use curly brackets in argument notion for JSDocs
```suggestion
   * @param {stateFork} The domain will be selected the previous or current version if provided. Otherwise, the chain config fork schedule is used.
```

dapplion (changes requested):
I'm not sure this solution is correct. What about doing this?

Pushed the solution only on config package here https://github.com/ChainSafe/lodestar/compare/cayman/getDomain-alternative?expand=1

```ts
  return {
    getDomain(domainType: DomainType, slot: Slot, state?: {slot: Slot}): Uint8Array {
      // ```py
      // def get_domain(state: BeaconState, domain_type: DomainType, epoch: Epoch=None) -> Domain:
      //   """
      //   Return the signature domain (fork version concatenated with domain type) of a message.
      //   """
      //   epoch = get_current_epoch(state) if epoch is None else epoch
      //   fork_version = state.fork.previous_version if epoch < state.fork.epoch else state.fork.current_version
      //   return compute_domain(domain_type, fork_version, state.genesis_validators_root)
      // ```

      // If not state argument is passed, assume requested slot is the state slot
      const stateSlot = state?.slot ?? slot;
      const epoch = Math.floor(slot / SLOTS_PER_EPOCH);
      // Get pre-computed fork schedule, which _should_ match the one in the state
      const forkInfo = chainForkConfig.getForkInfo(stateSlot);
      // Only allow to select either current or previous fork respective of the fork schedule at stateSlot
      const forkName = epoch < forkInfo.epoch ? forkInfo.prevForkName : forkInfo.name;

      let domainByType = domainCache.get(forkName);
      if (!domainByType) {
        domainByType = new Map<DomainType, Uint8Array>();
        domainCache.set(forkInfo.name, domainByType);
      }
      let domain = domainByType.get(domainType);
      if (!domain) {
        domain = computeDomain(domainType, forkInfo.version, genesisValidatorsRoot);
        domainByType.set(domainType, domain);
      }
      return domain;
    },
```

wemeetagain:
refactored similarly to your solution, please rereview

g11tech (dismissed):
LGTM!

packages/validator/src/services/validatorStore.ts
```diff
@@ -182,7 +182,7 @@ export class ValidatorStore {
     // Duties are filtered before-hard by doppelganger-safe, this assert should never throw
     this.assertDoppelgangerSafe(pubkey);
 
-    const proposerDomain = this.config.getDomain(DOMAIN_BEACON_PROPOSER, blindedOrFull.slot);
+    const proposerDomain = this.config.getDomain(currentSlot, DOMAIN_BEACON_PROPOSER, blindedOrFull.slot);
```
dapplion: Should this really be the currentSlot? Without checking the specs, sounds it should be the state slot, so block's slot

g11tech: currentslot should be same as blindedorfull slot same as state slot for signing block here

dapplion: _should_ is not enough, this can lead to bad signatures on the fork slot if there are clock issues. There's no solid reason to not use `blindedOrFull.slot`

g11tech: updated :+1:

g11tech (approved):

dapplion (approved):

PR #4438, opened 2022-08-18
unregisterValidator when removing key from API

### LINKED ISSUE #4434: Implement unregistering a validator function when pubkey is removed from keystore

If you remove a pubkey from the keystore via API, the key is not removed from the map `doppelgangerStateByPubkey`.

Implement a "unregister validator" type of function within the Doppelganger service whenever a pubkey is removed from the keystore.

dapplion:
**Motivation**

In case a public key is removed from the keystore using API, the key is not removed from the map doppelgangerStateByPubkey within the Doppelganger service. This behavior may result in a situation where the validator is killed unintentionally when a mis-
takenly added public key which is already in use by another validator is again removed from the keystore leading to a DoS situation of the validator as the surveillance of the public key within the Doppelganger service is still active.

**Description**

- unregisterValidator when removing key from API
- Remove unnecessary call in removeDutiesForKey(), API calls both validator.removeDutiesForKey() + validator.validatorStore.removeSigner()

Closes #4434

### DIFF
--- a/packages/validator/src/validator.ts
+++ b/packages/validator/src/validator.ts
@@ -206,7 +206,6 @@ export class Validator {
   }
 
   removeDutiesForKey(pubkey: PubkeyHex): void {
-    this.validatorStore.removeSigner(pubkey);
     this.blockProposingService.removeDutiesForKey(pubkey);
     this.attestationService.removeDutiesForKey(pubkey);
     this.syncCommitteeService.removeDutiesForKey(pubkey);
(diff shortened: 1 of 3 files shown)


packages/validator/src/validator.ts
```diff
@@ -206,7 +206,6 @@ export class Validator {
   }
 
   removeDutiesForKey(pubkey: PubkeyHex): void {
-    this.validatorStore.removeSigner(pubkey);
```
wemeetagain, line 209: why removed here?

wemeetagain: because it already happens in the keymanager api

wemeetagain (approved):

PR #4440, opened 2022-08-18
Improve isValidAddress validation

dapplion:
**Motivation**

Properly validate hex address

**Description**

Improve isValidAddress validation

### DIFF
--- a/packages/beacon-node/src/util/address.ts
+++ b/packages/beacon-node/src/util/address.ts
@@ -1,3 +1,3 @@
 export function isValidAddress(address: string): boolean {
-  return !!address && address.startsWith("0x") && address.length === 42;
+  return /^0x[0-9a-fA-F]{40}$/.test(address);
 }


packages/beacon-node/src/util/address.ts
```diff
@@ -1,3 +1,3 @@
 export function isValidAddress(address: string): boolean {
-  return !!address && address.startsWith("0x") && address.length === 42;
+  return /^0x[0-9a-fA-F]{40}$/.test(address);
```
wemeetagain, line 2: not sure how much this is used, I think it may be faster to create a singleton regex object, so it can be reused each time isValidAddress is called.

wemeetagain (approved):

packages/beacon-node/src/util/address.ts
dapplion: NodeJS should cache those when interpreting the code (in theory)

Issue #4208, opened 2022-06-27
Followup tracker for MeV related pending issues/cleanup


g11tech:
This is a followup tracker for the 
 - https://github.com/ChainSafe/lodestar/pull/3969

#### Tasks
- [x]  [add the blinded variants to getForkTypes ](https://github.com/ChainSafe/lodestar/pull/3969#discussion_r907601731)
- [x] [blinded vs full block strategy] (https://github.com/ChainSafe/lodestar/pull/3969#discussion_r898818937)
- [x] https://github.com/ChainSafe/lodestar/pull/4703
- [x] Add metrics
~~- [ ] Validate the signature in signed builder bid received and validate for amount > min block rewards~~
- [x] https://github.com/ChainSafe/lodestar/pull/4488
- [x] [Do not create  a new validator registration data unless change in reg data](https://github.com/ChainSafe/lodestar/pull/4447)
  - this has been requested by the mev boost team to avoid load on relays, still send registrations but don't create new ones
  - ![image](https://user-images.githubusercontent.com/76567250/186832977-22ecb01a-33c3-484b-a497-deb6e98c3633.png)

PR #4447, opened 2022-08-19
Cache and reuse old validator registration data if no params change

g11tech:
**Motivation**
The FB team requested not to create/send new registration objects unless params change.

This PR caches the registration objects and reuses them to register them to builder.
Reference: 
- https://github.com/ChainSafe/lodestar/issues/4208


packages/validator/src/services/validatorStore.ts
```diff
@@ -79,6 +80,8 @@ type ValidatorData = {
  */
 export class ValidatorStore {
   private readonly validators = new Map<PubkeyHex, ValidatorData>();
+  private readonly validatorRegistrationCache = new ValidatorRegistrationCache();
```
wemeetagain: Can you include a jsdoc comment here or on the class?

g11tech: done!

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -95,11 +94,7 @@ export function pollBuilderValidatorRegistration(
               const pubkeyHex = validatorStore.getPubkeyOfIndex(index);
               if (!pubkeyHex) throw Error(`Pubkey lookup failure for index=${index}`);
               const feeRecipient = validatorStore.getFeeRecipient(pubkeyHex);
-              return validatorStore.signValidatorRegistration(
-                fromHexString(pubkeyHex),
-                fromHexString(feeRecipient),
-                slot
-              );
+              return validatorStore.getValidatorRegistration(pubkeyHex, feeRecipient, slot);
```
twoeths: > the FB team requested not to create/send new registration objects unless params change.

@g11tech I still see we send every validator registration per epoch, is it expected?

g11tech: yes, they also only keep a registration for 2 epochs, so its recommended to refresh every epoch :+1:

packages/validator/src/util/validatorRegistrationCache.ts
```diff
@@ -0,0 +1,56 @@
...
+ * This cache stores the bellatrix.SignedValidatorRegistrationV1 objects as mev boosts wants
+ * us to send an old regsitration object if there are no changes in the registration data.
+ *
+ * This could potentially be to prevent a DOS attack as every epoch the public builders would have
+ * a hard time processing and verifying the registrations from potentially half a million validators
+ */
+export class ValidatorRegistrationCache {
+  private readonly validatorRegistrationMap = new Map<
+    string,
+    {validatorRegistration: bellatrix.SignedValidatorRegistrationV1; fullKey: string}
```
dapplion: Add alias on type string such that

```ts
type PubkeyHex = string
/** `${feeRecipient}-${gasLimit}` */
type AttributesKey = string
```

dapplion: Since fullKey is nested by first level of pubkey, does it really need to include the pubkey?

packages/validator/src/services/validatorStore.ts
dapplion: Why not add this inside `ValidatorData` so it gets automatically pruned when removing validators?

packages/validator/src/util/validatorRegistrationCache.ts
dapplion: If you add this cache inside `ValidatorData` you don't need the 2d Map. And then you can prune by number of attribute combinations which is what makes more sense. pubkeys are naturally bounded already

packages/validator/src/services/validatorStore.ts
g11tech: moved

wemeetagain (dismissed):

wemeetagain (approved):

PR #4471, opened 2022-08-23
Chunkify prepareBeaconCommitteeSubnet() for large request

twoeths:
**Motivation**

On a node of 10k validators, it is not able to detect connected validators because `prepareBeaconCommitteeSubnet()` api does not work due to this error:
```
Aug-23 00:00:04.679[]                ^[[31merror^[[39m: Error on poll attesters epoch=116400 Payload Too Large: Request body is too large - Failed to subscribe to beacon committee subnets
Error: Payload Too Large: Request body is too large - Failed to subscribe to beacon committee subnets
    at HttpClient.requestWithBody (file:///usr/src/lodestar/packages/api/src/utils/client/httpClient.ts:121:15)
```

**Description**

- Chunkify the `prepareBeaconCommitteeSubnet` api at validator side
- Since we have nodes with 1k validators for a while, choose `SUBSCRIPTIONS_PER_REQUEST` as 1k

**Test**

Screen Shot 2022-08-23 at 11 22 22

### DIFF
--- a/packages/validator/src/services/attestationDuties.ts
+++ b/packages/validator/src/services/attestationDuties.ts
@@ -14,10 +14,12 @@ import {ChainHeaderTracker, HeadEventData} from "./chainHeaderTracker.js";
 const HISTORICAL_DUTIES_EPOCHS = 2;
 
 /**
- * On lodestar test nodes, we can have up to 1k validators per node. On a node with too many validators,
- * it can cause "Request body is too large" issue for http post.
+ * This is to prevent the "Request body is too large" issue for http post.
+ * Typical server accept up to1MB (2 ** 20 bytes) of request body, for example fastify and nginx.
+ * A typical subscription request is 107 bytes in length, make it 120 to buffer.
+ * This number is Math.floor(2 ** 20 / 120)
  **/
-const SUBSCRIPTIONS_PER_REQUEST = 1000;
+const SUBSCRIPTIONS_PER_REQUEST = 8738;
 
 /** Neatly joins the server-generated `AttesterData` with the locally-generated `selectionProof`. */
 export type AttDutyAndProof = {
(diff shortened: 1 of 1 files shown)


packages/validator/src/services/attestationDuties.ts
```diff
@@ -184,10 +190,15 @@ export class AttestationDutiesService {
 
     // If there are any subscriptions, push them out to the beacon node.
     if (beaconCommitteeSubscriptions.length > 0) {
-      // TODO: Should log or throw?
-      await this.api.validator.prepareBeaconCommitteeSubnet(beaconCommitteeSubscriptions).catch((e: Error) => {
-        throw extendError(e, "Failed to subscribe to beacon committee subnets");
-      });
+      const subscriptionsBatches = batchItems(beaconCommitteeSubscriptions, {batchSize: SUBSCRIPTIONS_PER_REQUEST});
+      await Promise.all(
```
g11tech: do we want to loop and await each chunk rather than parallel await?

dapplion: @tuyennhv I would like to compute the real max size of this call. The limiting factor should be request size.

wemeetagain: @g11tech I think parallel await is better here since we don't need to wait for each request to fully resolve before triggering the next.

twoeths: @dapplion a request size is around 107 bytes, make it 120 to buffer, typically a server accept a max body size of 1MB. I changed the constant to `Math.floor(2 ** 20 / 120) = 8738`

so for the 10k validators node, it takes only 2 requests. Tested in that node and it works well.

I added all of this in the comment of `SUBSCRIPTIONS_PER_REQUEST` constant

g11tech: can we atleast wrapError these promises and throw a single error (incase that happens) else it would fill up the screen in case there are 3+ such requests, not sure how we can achieve something similar on the bn side

OR we can merge this and add this to the list of UX errors and do it there

g11tech: another reason regarding sending this one by one could be: that bn will process requests one by one even if we send them in parallel, and there could be timeout issues for lining up the calls, though don't know what would be request to response timeline for a full chunk

twoeths: @g11tech `Promise.all()` will reject immediately upon any promise reject https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

g11tech (approved):
LGTM

dapplion (approved):

PR #4473, opened 2022-08-24
Split dashboards and review metrics

dapplion:
**Motivation**

Our current dashboard is way too big and defeats the purpose of good design where each dashboard has a well defined goal.

**Description**

Review existing metrics:
- Use better histogram buckets
- Use histogram buckets instead of avgMinMax
- Remove index label from validator monitor

Split dashboards
- Move existing charts from main dashboard to smaller ones by themes


dapplion:
- Closing for https://github.com/ChainSafe/lodestar/pull/4474 and https://github.com/ChainSafe/lodestar/pull/4475

PR #4474, opened 2022-08-24
Split main dashboard

dapplion:
**Motivation**

Our current dashboard is way too big and defeats the purpose of good design where each dashboard has a well defined goal.

**Description**

- Split main dashboard into 

```
lodestar_block_processor.json
lodestar_block_production.json
lodestar_bls_thread_pool.json
lodestar_debug_gossipsub.json
lodestar_discv5.json
lodestar_execution_engine.json
lodestar_multinode.json
lodestar_networking.json
lodestar_rest_api.json
lodestar_state_cache_regen.json
lodestar_summary.json
lodestar_sync.json
lodestar_validator_client.json
lodestar_validator_monitor.json
lodestar_vm_host.json
```

- Add script `scripts/assert_metrics_usage.mjs` to check which metrics are not used. **Note**: does not enforce usage


wemeetagain:
devils advocate: Its useful to debug nodes by seeing all available data on a single page. Sometimes you don't know what you're looking for, just looking for visual cues in the sea of data.  If you have a specific time frame and node selected, these get reset between dashboard views. Toggling between dashboards can be cumbersome in that case.

It might be worth maintaining the single-page metrics dashboard also?

dapplion:
> devils advocate: Its useful to debug nodes by seeing all available data on a single page. Sometimes you don't know what you're looking for, just looking for visual cues in the sea of data. If you have a specific time frame and node selected, these get reset between dashboard views. Toggling between dashboards can be cumbersome in that case.
> 
> It might be worth maintaining the single-page metrics dashboard also?

After being a very heavy user of the single dashboard I say now it's a terrible idea. This single dashboard is un-mantainable now, draining time from me to update it. Also, all players I've seen around do use of multiple dashboards, which if properly linked and tag are super easy to navigate around. If you get used to the multi-dashboard workflow you'll discover that it much better to work with.

For the purpose of comparing data points across domains, Grafana has very powerful features to do data comparision, checkout "Grafana Explore" and it's split views.

wemeetagain:
Ok, lets try it out :rocket:

wemeetagain (approved):

Issue #4246, opened 2022-07-06
Unknown block sync peerId conversion issue


dapplion:
In Sepolia, seems an error of conversion

```
Jul-06 14:24:35.555[SYNC]            error: Downloaded block parent is before finalized slot finalizedSlot=115200, parentSlot=115192, parentRoot=0xd1670e16e8614871ea41daddd96aeab17f55b0a331afb5577e99100a7a83c1db
Jul-06 14:24:35.559[SYNC]            error: Removing unknown parent block root=0x1815dbd1a77bf2530893ef2c9bf6250a84be6b913cfe76d164d5f9c716d8728d, slot=115208
Jul-06 14:24:35.559[SYNC]            error: Removing unknown parent block root=0x01da944e86c181f131a3d2bf50fd9b599150e687b53c991d969715d8d86a5e5e, slot=115210
Jul-06 14:24:35.559[SYNC]            error: Removing unknown parent block root=0xacd63ca2b12641cb281fe40a479a5bf37debbba9ae324e4c9089742fdbbe1403, slot=115224
Jul-06 14:24:35.559[SYNC]            error: Removing unknown parent block root=0x6363c1500f9f25e3ce68e81444f9a5a5229b6ddad8c2b32e2abff5bc02b8c350, slot=115245
Jul-06 14:24:35.559[SYNC]            error: Removing unknown parent block root=0x4cd26ac871cbbd8c78558bf3f7be212c0da4fd549947b79461ede7164a8592be, slot=115255
Jul-06 14:24:35.559[SYNC]            error: Removing unknown parent block root=0xa88ce720f7907631dc7a0bc2e6ef6934eaa38737c914e7c5ac728b28e9ea6d9e, slot=115256
Jul-06 14:24:35.560[SYNC]            error: Removing unknown parent block root=0xf730df20bae1f30477f016bfd7820997dcf81e113fc44b7b5d0ce747520efc28, slot=115264
Jul-06 14:24:35.560[SYNC]            error: Removing unknown parent block root=0x46877c82c449986ed63db6e04b235674b83131803f78cc35a1ca931565766d9f, slot=115274
Jul-06 14:24:35.560[SYNC]            error: Removing unknown parent block root=0x0479c0245a7500c042e97643d27b09755a9d89e366c49c5a0ce1713b26c986d8, slot=115299
[...]

PR #4479, opened 2022-08-25
Log bad PeerId format

dapplion:
**Motivation**

- This issue https://github.com/ChainSafe/lodestar/issues/4246 is hard to debug as we don't know the format of the string that was considered invalid 

**Description**

Wrap conversion and log error in case it's invalid

@wemeetagain maybe in #4246 there was a mix of different encodings used? Is it possible to commit to use PeerId or PeerIdStr everywhere?


wemeetagain:
> Is it possible to commit to use PeerId or PeerIdStr everywhere?

at least for now, libp2p plans to use PeerId objects (although i'm still an advocate for https://github.com/libp2p/js-libp2p/issues/680)

I'd still recommend we use PeerIdStr everywhere in lodestar .
Else we'd need to convert all Map<PeerIdStr, T> to PeerMap<T>

wemeetagain (approved):

PR #4482, opened 2022-08-25
Set libp2p agent version

### LINKED ISSUE #4465: Specify agent version for lodestar

Right now lodestar is always identified as the default agent version of libp2p, for example `js-libp2p/0.36.2`, see https://github.com/libp2p/js-libp2p/blob/1f38ab7ac8380c9501b252d076bb356662978882/src/identify/consts.ts#L5

- We should be able to provide our own version, for example `lodestar/0.41.0` using `identify` option https://github.com/libp2p/js-libp2p/blob/1f38ab7ac8380c9501b252d076bb356662978882/src/index.ts#L112
- Inform other clients to identify lodestar using this new agent version

dapplion:
**Motivation**

- See https://github.com/ChainSafe/lodestar/issues/4465

**Description**

Set libp2p agentVersion to Lodestar's version in a way that other clients should be able to parse it. Lighthouse has to match the string before the first `/` to be equal to `js-libp2p` to consider the client to be Lodestar.

- agentVersion before: `js-libp2p/0.32.4`
- agentVersion with this PR: `js-libp2p/lodestar-v1.0.0`

Closes https://github.com/ChainSafe/lodestar/issues/4465

### DIFF
--- a/packages/beacon-node/src/network/nodejs/bundle.ts
+++ b/packages/beacon-node/src/network/nodejs/bundle.ts
@@ -18,6 +18,7 @@ export interface ILibp2pOptions {
   bootMultiaddrs?: string[];
   maxConnections?: number;
   minConnections?: number;
+  lodestarVersion?: string;
 }
 
 export class NodejsNode extends LibP2p {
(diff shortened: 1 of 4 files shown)


dapplion:
TODO:
- [ ] Request other clients to switch to `lodestar`
- [ ] Replace js-libp2p to lodestar

PR #4487, opened 2022-08-26
Simplify beacon cmd init sequence

dapplion:
**Motivation**

Due to lots of legacy code the beacon cmd init sequence is more complicated that necessary.

**Description**

- I've moved all code to the handler itself which makes it much easier to reason about. All code that does not produce hard side-effects (like starting the beacon node) is in a separate function `beaconHandlerInit()` which can be easily unit tested
- By default no peer-id nor ENR is read from disk. The peer-id is now ephemeral unless using `--peerIdFile` flag. Both files are persisted at default locations for debugging
- init command is removed. The command has not done anything since many versions ago. It is a leftover from 2020 lodestar mechanics
- Lodestar can handle using custom networks with their name defined as `CONFIG_NAME` in the config.yml file. This is how many devnets are launched and lodestar would default to "mainnet" in those cases


packages/cli/src/cmds/beacon/handler.ts
```diff
@@ -94,3 +80,68 @@ export async function beaconHandler(args: IBeaconArgs & IGlobalArgs): Promise<vo
...
+
+  // Persist ENR and PeerId in beaconDir fixed paths for debugging
+  const pIdPath = path.join(beaconPaths.beaconDir, "peer_id.json");
+  const enrPath = path.join(beaconPaths.beaconDir, "enr");
+  writeFile(pIdPath, peerId.toJSON());
+  const fileENR = FileENR.initFromENR(enrPath, peerId, enr);
+  fileENR.saveToFile();
+
+  // Inject ENR to beacon options
+  beaconNodeOptions.set({network: {discv5: {enr: fileENR, enrUpdate: !enr.ip && !enr.ip6}}});
```
dapplion, line 127: @wemeetagain I don't understand why enr.ip and enr.ip6 decides if we should update the ENR or not. Probably this is legacy code could you check what we wanna do here?

packages/cli/src/cmds/beacon/options.ts
```diff
@@ -1,20 +1,17 @@
...
+import {beaconNodeOptions, paramsOptions, IBeaconNodeArgs} from "../../options/index.js";
 import {defaultLogMaxFiles, ICliCommandOptions, ILogArgs} from "../../util/index.js";
 import {defaultBeaconPaths, IBeaconPaths} from "./paths.js";
 
 interface IBeaconExtraArgs {
   forceGenesis?: boolean;
   genesisStateFile?: string;
+  configFile?: string;
+  bootnodesFile?: string;
+  peerIdFile?: string;
```
dapplion: This paths are now not relative to beaconDir so must be defined independently

packages/cli/src/cmds/beacon/options.ts
```diff
@@ -86,41 +115,58 @@ export const beaconPathsOptions: ICliCommandOptions<IBeaconPaths> = {
...
-    defaultDescription: defaultBeaconPaths.enrFile,
-    type: "string",
-  },
+interface IENRArgs {
+  "enr.ip"?: string;
+  "enr.tcp"?: number;
+  "enr.ip6"?: string;
+  "enr.udp"?: number;
+  "enr.tcp6"?: number;
+  "enr.udp6"?: number;
```
dapplion, line 137: This options are only used by the beacon cmd, there's no point in defining them close to global options

packages/cli/src/config/beaconNodeOptions.ts
```diff
@@ -1,36 +1,13 @@
...
-    beaconNodeOptionsCli: RecursivePartial<IBeaconNodeOptions>;
-  }) {
-    this.beaconNodeOptions = mergeBeaconNodeOptions(
-      network ? getNetworkBeaconNodeOptions(network) : {},
-      configFile ? readBeaconNodeOptions(configFile) : {},
-      bootnodesFile ? getInjectableBootEnrs(bootnodesFile) : {},
-      beaconNodeOptionsCli
-    );
-  }
+  constructor(private beaconNodeOptions: RecursivePartial<IBeaconNodeOptions>) {}
```
dapplion, line 10: Option merging is done in the beacon cmd handler, no need for all this complicated infrastructure

packages/cli/src/cmds/beacon/handler.ts
wemeetagain: if the enr ip is explicitly set, we assume that the user doesn't want it to be updated further.

wemeetagain:
I feel the changes around enr and peer-id persistence deserve deeper discussion than being rolled into this cleanup. Otherwise lgtm.

dapplion:
> I feel the changes around enr and peer-id persistence deserve deeper discussion than being rolled into this cleanup. Otherwise lgtm.

What concerns do you have? While unfortunate to be bundled on this PR, keeping that old behaviour defeats the simplicity of this PR. Plus Lodestar would greatly benefit from reseting the peerId as it's obvious that we are getting downscored. Can we discuss your concerns now before blocking this PR indefinitely?

wemeetagain (commented):
We talked offline. We will create an issue to reintroduce persistent peer id

wemeetagain (dismissed):
LGTM

wemeetagain (approved):

PR #4488, opened 2022-08-27
Add circuit breaker for builder

### LINKED ISSUE #4483: Failsafe circuit breaker for builder API

Lodestar will need to include https://github.com/ethereum/builder-specs/pull/47 to switch off builder network to preserve liveness. 

Implement in our builder API.

g11tech:
**Motivation**
In case there is a malevolent or buggy builder, it can take down the entire network by not propagating the blocks 

This PR add a circuit breaker for builder in bad network conditions
**Description**
Closes #4483
Manual Test chronology on `ropsten-1`
1. BN started, builder is initalized as disabled , validator produceBlock call comes in: builder flow errors as builder is disabled (and engine produced block is used)
  - image
2. After sometime, prepareNextSlot scheduler runs for an impending proposal, the health is evaluated, builder status checked via api and builder enabled and the next slot is proposed using builder block 
  - image
3. Post this builder stays enabled and validator keeps using the builder
 - image

### DIFF
--- a/packages/beacon-node/src/chain/prepareNextSlot.ts
+++ b/packages/beacon-node/src/chain/prepareNextSlot.ts
@@ -120,13 +120,13 @@ export class PrepareNextSlotScheduler {
         const proposerIndex = prepareState.epochCtx.getBeaconProposer(prepareSlot);
         const feeRecipient = this.chain.beaconProposerCache.get(proposerIndex);
         if (feeRecipient) {
-          // Update if the builder flow can be used with this proposer, it might lead to
-          // an api call to the builder to check its status, so we will not await for
-          // its completion here as we want to trigger local execution here without
-          // delay
-          this.chain.updateBuilderStatus(clockSlot).catch((e) => {
-            this.logger.error("Updating builder status failed", {prepareSlot}, e as Error);
-          });
+          // Update the builder status, if enabled shoot an api call to check status
+          this.chain.updateBuilderStatus(clockSlot);
+          if (this.chain.executionBuilder?.status) {
+            this.chain.executionBuilder.checkStatus().catch((e) => {
+              this.logger.error("Builder disabled as the check status api failed", {prepareSlot}, e as Error);
+            });
+          }
 
           const preparationTime =
             computeTimeAtSlot(this.config, prepareSlot, this.chain.genesisTime) - Date.now() / 1000;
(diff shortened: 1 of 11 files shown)


packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -236,6 +236,15 @@ export class ForkChoice implements IForkChoice {
     return (this.head = headNode);
   }
 
+  /**
+   * An iteration over protoArray to get present slots, to be called pre-emptively
+   * from prepareNextSlot to prevent delay on produceBlindedBlock
+   * @param windowStart is the slot after which (excluding) to provide present slots
+   */
+  getSlotsPresent(windowStart: number): number {
```
dapplion, line 244: Better name would be `getBlockCountAfterSlot(afterSlot: Slot): number`

packages/beacon-node/src/chain/prepareNextSlot.ts
```diff
@@ -120,6 +120,14 @@ export class PrepareNextSlotScheduler {
         const proposerIndex = prepareState.epochCtx.getBeaconProposer(prepareSlot);
         const feeRecipient = this.chain.beaconProposerCache.get(proposerIndex);
         if (feeRecipient) {
+          // Update if the builder flow can be used with this proposer, it might lead to
+          // an api call to the builder to check its status, so we will not await for
+          // its completion here as we want to trigger local execution here without
+          // delay
+          this.chain.updateBuilderStatus(clockSlot).catch((e) => {
```
dapplion: Without a more in-depth review this looks problematic. Since this is async there's a racing condition on producing the block and updating the status. Could be that the builder is disabled too late and the block is already produced?

Why not check at the end of every epoch (or some point, say at 2nd slot) the status of the previous epoch and update the status of the builder with it?
- That allows to have constant metrics of status
- You can subscribe to block event instead of `getSlotsPresent()` and have real-time cumulative status of blocks per epoch, which can also go into a metric

g11tech: :+1:  for counting the slots via event and the corresponding metric

However with regard to calling `updateBuilderStatus` here: for `disabling`, this is synchronous because the builder's check status api is only being called for `enabling`, which if it doesn't respond within a second should anyway be ok to have builder disabled for the next slot's block.

I think its a good idea to check for builder status  here before proposing next slot.

dapplion: > However with regard to calling updateBuilderStatus here: for disabling, this is synchronous because the builder's check status api is only being called for enabling

Got it, then can you find a way to make it more obvious on the code that disabling is sync?

g11tech: done

dapplion (dismissed):
Excellent!

Discussed offline to change the ALLOWED_FAULTS value to not follow the spec, as allowing only 1 fault is too sensitive.

dapplion (approved):

PR #4493, opened 2022-08-30
Check API definitions against OpenAPI spec

dapplion:
**Motivation**

We have no programmatic guarantee that Lodestar follows beacon APIs spec. In multiple instances it has been found that Lodestar had formatting issues, breaking interoperability.

**Description**

This PR, ensure that Lodestar follows a specific beacon APIs spec version having correct:
- URL + HTTP method
- Request format, case and properties (params, query, body)
- Response format, case and properties (request body on 200 response)

Closes https://github.com/ChainSafe/lodestar/issues/4403
Closes https://github.com/ChainSafe/lodestar/issues/4063

PR #4496, opened 2022-09-01
Enable builder api spec tests

dapplion:
**Motivation**

We have no programmatic guarantee that Lodestar follows beacon APIs spec. In multiple instances it has been found that Lodestar had formatting issues, breaking interoperability.

- Extends https://github.com/ChainSafe/lodestar/pull/4493

**Description**

This PR, ensure that Lodestar follows a specific builder APIs spec version having correct:
- URL + HTTP method
- Request format, case and properties (params, query, body)
- Response format, case and properties (request body on 200 response)


wemeetagain (approved):

g11tech (approved):
LGTM

PR #4502, opened 2022-09-01
Fix remote signing

### LINKED ISSUE #4424: Retrieving PubKeys from Remote Signer Fails
**Describe the bug**

When using remote signer, in the flow, there is a request to fetch the public key from the remote signer, via this call:

```
  const res = await fetch(`${externalSignerUrl}/keys`, {
    method: "GET",
    headers: {"Content-Type": "application/json"},
  });
```

This currently fails with the error:

```
Aug-15 17:16:26.532[]                error: Failed to download attester duties epoch=114818 <html><body><h1>Resource not found</h1></body></html>
Error: <html><body><h1>Resource not found</h1></body></html>
    at handlerExternalSignerResponse (file:///home/seamonkey/Downloads/goerli/lodestar/packages/validator/src/util/externalSignerClient.ts:52:11)
    at externalSignerPostSignature (file:///home/seamonkey/Downloads/goerli/lodestar/packages/validator/src/util/externalSignerClient.ts:32:16)
    at ValidatorStore.getSignature (file:///home/seamonkey/Downloads/goerli/lodestar/packages/validator/src/services/validatorStore.ts:391:32)
    at ValidatorStore.signAttestationSelectionProof (file:///home/seamonkey/Downloads/goerli/lodestar/packages/validator/src/services/validatorStore.ts:310:12)
    at AttestationDutiesService.getDutyAndProof (file:///home/seamonkey/Downloads/goerli/lodestar/packages/validator/src/services/attestationDuties.ts:323:28)
    at AttestationDutiesService.pollBeaconAttestersForEpoch (file:///home/seamonkey/Downloads/goerli/lodestar/packages/validator/src/services/attestationDuties.ts:225:30)
    at AttestationDutiesService.pollBeaconAttesters (file:///home/seamonkey/Downloads/goerli/lodestar/packages/validator/src/services/attestationDuties.ts:155:7)
    at async Promise.all (index 1)
    at AttestationDutiesService.runDutiesTasks (file:///home/seamonkey/Downloads/goerli/lodestar/packages/validator/src/services/attestationDuties.ts:116:5)
```

Source: From [Discord](https://discord.com/channels/593655374469660673/593655641445367808/1008854994323255336)

Reason seems to be that the API lodestar is expecting at the remote signer side has changed.

**Expected behavior**

Be able to retrieve pub keys from remote signer.

dadepo:
**Motivation**

Using remote signer currently does not work. This is because the web3signer URL is different from what lodestar expects.

PS: currently draft PR as the branch still need to be tested on goerli test net.

**Description**

Update implementation to use correct web3signer endpoints.

Closes #4424


packages/validator/src/services/validatorStore.ts
```diff
@@ -299,17 +349,35 @@ export class ValidatorStore {
...
+      signature: await this.getSignature(duty.pubkey, signingRoot, signableMessage),
     };
   }
 
   async signAttestationSelectionProof(pubkey: BLSPubkeyMaybeHex, slot: Slot): Promise<BLSSignature> {
     const domain = this.config.getDomain(slot, DOMAIN_SELECTION_PROOF);
     const signingRoot = computeSigningRoot(ssz.Slot, slot, domain);
 
-    return await this.getSignature(pubkey, signingRoot);
+    const signableMessage: SignableMessage = {
```
dapplion, line 482: If the web3signer is not enabled, all this objects are being created for nothing right?

dadepo: Updated to only create the Singable messages when web3signer is enabled.

dapplion:
Please add an e2e test that run the web3signer docker container and request signatures for all available objects. I would really need to have programmatic certainty that this works before merging.

packages/validator/src/util/externalSignerClient.ts
```diff
@@ -1,31 +1,199 @@
...
+        version: data.version,
+        block: ssz.phase0.BeaconBlock.toJson(data.block as phase0.BeaconBlock),
+      };
+    } else if (data.version === "ALTAIR") {
+      return {
+        version: data.version,
+        block: ssz.altair.BeaconBlock.toJson(data.block as altair.BeaconBlock),
+      };
+    } else {
+      throw new Error(`version ${data.version} not supported by remote signer`);
```
dapplion: What about the bellatrix fork? Using this pattern doesn't scale and will be painful to mantain

dadepo: I basically followed the pattern used in lighthouse [here](https://github.com/sigp/lighthouse/blob/aa022f46855df2a1420a6a80a788c73dc2779aa7/validator_client/src/signing_method/web3signer.rs#L76). Had an older version so bellatrix was not included. Added it now. As regards observation about maintenance. I'll revisit this after setting up all the tests

dadepo: Yet to do the final sanity check but it occurred to me, that maybe the maintenance won't be too hard. Because:

1. Having another fork is a coordinated effort that will be planned out and include coordinated code changes
2. Even if lodestar can handle new forks without code changes, it should not, because it also depends on making sure that the Web3Signer has been updated to also handle the new fork. If not, lodestar will pass the call to web3signer but it will fail.

packages/validator/src/services/validatorStore.ts
dadepo: @dapplion Your recent changes now makes it such that the objects are being created even when web3signer is not enabled. Is that intentional/okay now?

packages/validator/src/util/externalSignerClient.ts
dapplion: Right I wasnt aware of how weird the web3signer request types are 😅 made such that any fork after bellatrix with use the bellatrix path

packages/cli/src/cmds/validator/signers/index.ts
```diff
@@ -39,7 +39,12 @@ export async function getSignersFromArgs(args: IValidatorCliArgs & IGlobalArgs,
   // ONLY USE FOR TESTNETS - Derive interop keys
   if (args.interopIndexes) {
     const indexes = parseRange(args.interopIndexes);
-    return indexes.map((index) => ({type: SignerType.Local, secretKey: interopSecretKey(index)}));
+    // Using a remote signer with TESTNETS
+    if (args["externalSigner.pubkeys"] || args["externalSigner.fetch"]) {
+      return getRemoteSigners(args);
+    } else {
+      return indexes.map((index) => ({type: SignerType.Local, secretKey: interopSecretKey(index)}));
+    }
```
dapplion, line 47: What's the purpose of this? I don't follow

dapplion: To consume the interop indexes from the web3signer you should  load the interop keystores in the web3signer and then use .fetch

wemeetagain (approved):

dapplion (approved):

PR #4510, opened 2022-09-05
Add assertions to sim tests 

### LINKED ISSUE #4276: Add assertions to sim tests

**Proposed assertions**

Assertion on epoch:

**Background**

nazarhussain:
**Motivation**

Created a simulation environment helpers and added assertions. 

**Description**

Add following assertions: 

- node assertions 
- missed blocks assertions
- attestation participations

Closes #4276

```sh
git checkout <feature_branch>
cd packages/cli 
LODESTAR_PRESET=minimal yarn mocha test/simulation/singleNodeSingleValidator.test.ts
```


packages/cli/src/cmds/beacon/handler.ts
```diff
@@ -28,6 +28,13 @@ export async function beaconHandler(args: IBeaconArgs & IGlobalArgs): Promise<vo
   mkdir(beaconPaths.beaconDir);
   mkdir(beaconPaths.dbDir);
 
+  if (
+    args.dev &&
+    (process.env.NODE_ENV === "dev" || process.env.NODE_ENV === "development" || process.env.NODE_ENV === "test")
+  ) {
+    options = {...options, executionEngine: {mode: "mock"}} as IBeaconNodeOptions;
```
dapplion: This should be a regular CLI option, hidden if you need. Would prefer to not have this ENV switch is a production code path. Just add a hidden option --execution.engineMock

packages/cli/test/utils/simulation/assertions.ts
```diff
@@ -0,0 +1,307 @@
...
+      });
+    }
+  });
+}
+
+export function finalityAssertions(env: SimulationEnvironment): void {
+  describe("finality", () => {
+    for (let n = 0; n < env.params.beaconNodes; n++) {
+      describe(`beacon node "${n}"`, () => {
+        it("should have proper finality", async function () {
```
dapplion: If there's a single it, should this be un-nested to:
```
describe( finality )
  it( `beacon node ${n} some_more_text`)
```

packages/cli/test/utils/simulation/assertions.ts
```diff
@@ -0,0 +1,307 @@
...
+
+        describe("validator - n", () => {
+          it("should have correct status");
+          it("should have correct number of keys loaded");
+        });
+      });
+    }
+
+    it("all nodes should have same head", async () => {
+      const heads = await Promise.all(env.nodes.map((n) => n.api.beacon.getBlockHeader("head")));
```
dapplion: Nodes may receive blocks with some delay, so the head may be different without anything going wrong

packages/cli/test/utils/simulation/assertions.ts
```diff
@@ -0,0 +1,307 @@
...
+      const headRoots = heads.map((h) => h.data.root);
+      const firstHeadRoot = headRoots[0];
+
+      for (let n = 1; n < env.params.beaconNodes; n++) {
+        expect(headRoots[n]).to.equal(
+          firstHeadRoot,
+          `node ${n} has different head than node 0. firstNodeHead: ${Buffer.from(firstHeadRoot).toString(
+            "hex"
+          )}, node${n}Head: ${Buffer.from(headRoots[n]).toString("hex")}`
```
dapplion: You must log root and slot

packages/cli/test/utils/simulation/assertions.ts
```diff
@@ -0,0 +1,307 @@
...
+        // Make sure the its not end of epoch
+        expect(currentEpochParticipation.every((p) => p === 0)).to.be.eql(
+          true,
+          `Node ${n} has not set current epoch participation to zero. Probably its not the end of epoch.`
+        );
+
+        const previousEpochParticipation = (states[n].data as altair.BeaconState).previousEpochParticipation;
+        const totalAttestingBalance = previousEpochParticipation
+          .map((p, index) => ((p & TIMELY_HEAD) !== 0 ? states[n].data.balances[index] : 0))
+          .reduce((a, b) => a + b, 0);
```
dapplion: Write helper functions to compute participation. And consider just using a regular for loop would take one loop less over the data

packages/cli/test/simulation/fourNodeSingleValidator.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    await env.clock.waitForEndOfEpoch(2);
+  });
+
+  after(async () => {
+    await env.stop();
+  });
+
+  nodeAssertions(env);
+  missedBlocksAssertions(env);
+  participationAssertions(env);
```
dapplion: Participation is checked immediately?

nazarhussain: After 2 epochs.

packages/cli/test/utils/simulation/assertions.ts
nazarhussain: I was using a pattern, to group assertion by type and then for each node.

packages/cli/test/utils/simulation/assertions.ts
nazarhussain: Yes, but in our constraint simulation environment we expect that nodes should have same heads. 

Is there any alternative approach you suggest to check this? Should we check for a slot rather the head?

packages/cli/test/utils/simulation/index.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export function waitForEvent<T>(
+  emitter: EventEmitter,
+  event: string,
+  timeout = 3000,
+  condition: (e: T) => boolean = () => true
+): Promise<T> {
+  return new Promise((resolve, reject) => {
+    const timer = setTimeout(() => reject(new TimeoutError(`event ${event} not received`)), timeout);
+    emitter.on(event, (e) => {
```
wemeetagain: does this event handler need to be cleared?

packages/cli/test/simulation/fourNodeSingleValidator.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    // We need to use the attesting participation so have to switch to altair
+    altairEpoch: 1,
+    // Use a larger value instead of Infinity
+    // https://github.com/ChainSafe/lodestar/issues/4505
+    bellatrixEpoch: 10 ** 12,
+  });
+
+  before(async function () {
+    await env.start();
+    await env.network.connectAllNodes();
```
wemeetagain: Should this be done on all sim tests?

packages/cli/test/utils/simulation/index.ts
nazarhussain: It was an unused code. Cleaned it.

packages/cli/test/simulation/fourNodeSingleValidator.test.ts
nazarhussain: Yes it should, but we don't run multiple nodes in all tests (for now). Also there is an issue with multiple nodes causing duplication publication error need to debug further. 

This PR is focused only for these assertions: 

- Missed Blocks
- Participation on Head
- Participation on FFG

Rest will come in upcoming PRs and will be improved further as well.

dapplion (changes requested):
There's way to much data being logged in console to understand if the tests worked as intented. Doing text searches I found that some nodes exit immediately after starting

```
2022-09-08T22:43:18.3235817Z Eph 0/1 0.322[NODE-5]           [32minfo[39m: Stopping gracefully, use Ctrl+C again to force process exit
2022-09-08T22:43:18.3254731Z Eph 0/1 0.325[NODE-6]           [32minfo[39m: Stopping gracefully, use Ctrl+C again to force process exit
2022-09-08T22:43:18.3338321Z Eph 0/1 0.330[NODE-8]           [32minfo[39m: Stopping gracefully, use Ctrl+C again to force process exit
```

```
2022-09-08T22:43:18.2346537Z Eph 0/1 0.232[VAL-5]            [32minfo[39m: Stopping gracefully, use Ctrl+C again to force process exit
2022-09-08T22:43:18.2368150Z Eph 0/1 0.234[VAL-6]            [32minfo[39m: Stopping gracefully, use Ctrl+C again to force process exit
2022-09-08T22:43:18.2369576Z Eph 0/1 0.235[VAL-8]            [32minfo[39m: Stopping gracefully, use Ctrl+C again to force process exit
2022-09-08T22:43:18.2372852Z Eph 0/1 0.236[VAL-7]            [32minfo[39m: Stopping gracefully, use Ctrl+C again to force process exit
2022-09-08T22:43:18.2398433Z Eph 0/1 0.239[VAL-9]            [32minfo[39m: Stopping gracefully, use Ctrl+C again to force process exit
2022-09-08T22:43:18.2418074Z Eph 0/1 0.240[VAL-11]           [32minfo[39m: Stopping gracefully, use Ctrl+C again to force process exit
2022-09-08T22:43:18.2441290Z Eph 0/1 0.242[VAL-12]           [32minfo[39m: Stopping gracefully, use Ctrl+C again to force process exit
2022-09-08T22:43:18.2468207Z Eph 0/1 0.246[VAL-10]           [32minfo[39m: Stopping gracefully, use Ctrl+C again to force process exit
```

this is very concerning, why are nodes stopping that early?

source https://pipelines.actions.githubusercontent.com/serviceHosts/ef0bfa01-ab2d-480d-b599-9c7b1c6bb923/_apis/pipelines/1/runs/62462/signedlogcontent/2?urlExpires=2022-09-09T08%3A40%3A32.8449321Z&urlSigningMethod=HMACV1&urlSignature=8v2IgpbfkUhZZXoFGkLYHjrcIg69TjoUnPjW%2F6u0fqQ%3D

Also now there are sim tests running in the beacon node + CLI which are redundant. Bringing the total run time to 29 minutes. We must not do duplicate work.

----

After reading all the logs this is what I can guess it's happening:

- First run starts 8 nodes, hits genesis, does one run of assertions and stops all nodes immediatelly in the genesis slot
- Then a single node [NODE-9] starts again at Eph -3/5 and runs until slot 0 of epoch 2, then it's killed.
- Then a single node [NODE-10] starts at Eph -3/5 and runs until slot 0 of epoch 5, then it's killed

.github/workflows/test-sim.yml
```diff
@@ -52,9 +52,15 @@ jobs:
       #   run: yarn test:sim:multiThread
       #   working-directory: packages/beacon-node
       #   env: {RUN_ONLY_SIM_TEST: altair-epoch2}
-      - name: Upload debug log test files
+      - name: Upload debug log test files for "packages/beacon-node"
```
dapplion, line 54: Tests must be de-duplicated. This PR should replace the existing sim tests and cover all their cases + augment them by adding assertions.

packages/cli/test/simulation/multiNodeConfiguration.test.ts
```diff
@@ -1,18 +1,19 @@
...
 describe("multiNodeConfiguration", function () {
   this.timeout("5m");
   const env = new SimulationEnvironment({
     beaconNodes: 4,
-    validatorClients: 2,
+    validatorClients: 1,
     validatorsPerClient: 16,
     // Use a larger value instead of Infinity
     // https://github.com/ChainSafe/lodestar/issues/4505
     altairEpoch: 10 ** 12,
```
dapplion: Create a variable `FAR_FUTURE_EPOCH` and the comment once as JSDoc

packages/cli/test/simulation/singleNodeForks.test.ts
```diff
@@ -12,6 +12,7 @@ describe("singleNodeForks", function () {
       validatorsPerClient: 128,
       altairEpoch: 1,
       bellatrixEpoch: 2,
+      logFilesDir: `${logFilesDir}/singleNodeForks`,
```
dapplion: Test identification should be much more specific. Please check existing sim tests

packages/cli/test/utils/simulation/LodestarBeaconNodeProcess.ts
```diff
@@ -67,6 +67,9 @@ export const LodestarBeaconNodeProcess: BeaconNodeConstructor = class LodestarBe
       "params.BELLATRIX_FORK_EPOCH": String(this.params.bellatrixEpoch),
       logPrefix: this.id,
       logFormatGenesisTime: `${this.params.genesisTime}`,
+      logFile: `${this.params.logFilesDir}/${this.id}.log`,
+      logFileLevel: "debug",
+      logLevel: process.env.SHOW_LOGS ? "info" : "error",
```
dapplion: This is not enough, log info to sub-process stdout and don't pipe std to main process

packages/cli/test/simulation/singleNodeForks.test.ts
nazarhussain: I mapped the test files name with the log directory name. This is a lot easier approach to identify logs for a test file. 

In earlier tests, the processes are created multiple times in one test file, so why have a prefix for configuration.

packages/cli/test/utils/simulation/LodestarBeaconNodeProcess.ts
nazarhussain: If we don't pipe std of child process to main process, then it does not matter what its been logged in the child process at all.

I believe this is better approach for someone to easily debug in development.

.github/workflows/test-sim.yml
nazarhussain: As discussed earlier we will add assertions in multiple PRs. Once we reach to a satisfactory level, we can remove existing sim tests in the beacon-node.

nazarhussain:
@dapplion I am debugging the multi-node test which is passing locally and failing on CI. Rest is ready to review.

packages/cli/test/simulation/fourNodesForks.test.ts
```diff
@@ -0,0 +1,71 @@
...
+        await env.start();
+        await env.network.connectAllNodes();
+      });
+
+      after(async () => {
+        await env.stop();
+      });
+
+      it("all nodes should reach to finality", async () => {
+        await expect(Promise.all(env.nodes.map((node) => env.waitForEvent(event, node)))).be.fulfilled;
```
dapplion: No other assertions? Why not do all assertions available here?

packages/cli/test/simulation/fourNodesForks.test.ts
```diff
@@ -0,0 +1,86 @@
...
+chai.use(chaiAsPromised);
+
+describe.skip("fourNodesForks", function () {
+  this.timeout("5m");
+  let env: SimulationEnvironment;
+
+  const testCases: {
+    title: string;
+    params: {
+      event: routes.events.EventType;
```
dapplion: Since this test always wait for finalization, not need to declare the same event everytime

packages/cli/test/simulation/fourNodesForks.test.ts
```diff
@@ -0,0 +1,86 @@
...
+      bellatrixEpoch: number;
+      withExternalSigner?: boolean;
+    };
+  }[] = [
+    {
+      title: "phase0 fork only",
+      params: {
+        event: routes.events.EventType.finalizedCheckpoint,
+        altairEpoch: FAR_FUTURE_EPOCH,
+        bellatrixEpoch: FAR_FUTURE_EPOCH + 10,
```
dapplion: Why +10? Should not be necessary

packages/cli/test/simulation/fourNodesForks.test.ts
```diff
@@ -0,0 +1,86 @@
+import chai, {expect} from "chai";
+import chaiAsPromised from "chai-as-promised";
+import {routes} from "@lodestar/api/beacon";
+import {logFilesDir, SimulationEnvironment, FAR_FUTURE_EPOCH} from "../utils/simulation/index.js";
+
+chai.use(chaiAsPromised);
+
+describe.skip("fourNodesForks", function () {
+  this.timeout("5m");
```
dapplion: How is this 5m figure computed? Does it provide enough time for tests to complete? You should be able to compute the required timeout

packages/cli/test/utils/simulation/LodestarBeaconNodeProcess.ts
```diff
@@ -0,0 +1,137 @@
...
+  static totalProcessCount = 0;
+  readonly params: SimulationParams;
+  readonly secretKeys: Record<number, SecretKey[]> = {};
+  readonly address: string;
+  readonly port: number;
+  readonly restPort: number;
+  readonly id: string;
+  api!: Api;
+  peerId!: string;
+  multiaddrs!: string[];
```
dapplion: Do not use the non-null assertion operator, find a better way to be typesafe. At least acknowledge that the type may be undefined and use `multiaddrs: string[] | undefined`

packages/cli/test/utils/simulation/SimulationEnvironment.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    this.params = {
+      ...paramsWithDefaults,
+      genesisTime,
+      slotsPerEpoch: activePreset.SLOTS_PER_EPOCH,
+    } as SimulationParams;
+
+    this.controller = new AbortController();
+    this.id = getSimulationId(this.params);
+    this.rootDir = `${tmp.dirSync({unsafeCleanup: true}).name}/${this.id}`;
```
dapplion: Use `node:path` library `path.join` to do path manipulation

packages/cli/test/utils/simulation/SimulationEnvironment.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  async stop(): Promise<void> {
+    this.controller.abort();
+    await this.tracker.stop();
+    await Promise.all(this.nodes.map((p) => p.stop()));
+    await rm(this.rootDir, {recursive: true});
+  }
+
+  // TODO: Add timeout support
```
dapplion, line 97: Mocha's timeout should be sufficient to kill the process, just pick a timeout value dynamically that adjusts to the test's needs

packages/cli/test/utils/simulation/assertions.ts
```diff
@@ -0,0 +1,250 @@
...
+          const finalizedSlot = checkpoints.data.finalized.epoch;
+
+          expect(finalizedSlot).to.equal(currentSlot - 2);
+          expect(previousJustifiedSlot + 1).to.equal(
+            currentJustifiedSlot,
+            `there should be no gaps between current and previous justified epochs, received current ${currentJustifiedSlot} and previous ${previousJustifiedSlot}`
+          );
+          expect(currentJustifiedSlot + 1).to.equal(
+            finalizedSlot,
+            `there should be no gaps between finalized and previous justified epochs, received finalized ${finalizedSlot} and previous ${previousJustifiedSlot}`
```
dapplion: There can be gaps between finalized and previous justified epochs. The spec allows for that. Instead you want to assert is that every single epoch is finalized within that epoch. By testing that participation is > 66% on every epoch all this finality assertions are automatically true. If you want you can double check that, check that at the end of epoch N, currentJustified.epoch = N.

packages/cli/test/utils/simulation/assertions.ts
dapplion: The slot is by definition the same, since all nodes have the same clock (they are in the same host). Here it's very important the timing of when you are querying this. It can happen that (say Slot N starts at 17:00:00)
- Node0: set head at time 17:00:00.012
- Node1: set head at time 17:00:00.803
- Node2: set head at time 17:00:00.445
- Node3: set head at time 17:00:00.116

If you call heads at 17:00:00 exactly, the result you are gonna get is uncertain. The specific point through a slot at which you call heads is crucial to prevent false positives

packages/cli/test/simulation/fourNodesForks.test.ts
nazarhussain: I wanted `fourNodesForks.test.ts` to only cover the forks related assertions. We can add all assertions but not be scope specific test then.

packages/cli/test/simulation/singleNodeSingleValidatorMultipleEpoch.test.ts
```diff
@@ -0,0 +1,51 @@
+import {FAR_FUTURE_EPOCH, logFilesDir, SimulationEnvironment} from "../utils/simulation/index.js";
+import {missedBlocksAssertions, participationAssertions} from "../utils/simulation/assertions.js";
+
+const epochLimit = 5;
+
+describe("singleNodeSingleValidatorMultipleEpoch", function () {
```
dapplion: Is this test redundant to singleNodeForks? What is this test covering that singleNodeForks is not? Time is CI is precious for productivity, tests should add value comparable to the time they take

packages/cli/test/simulation/fourNodesForks.test.ts
nazarhussain: Just wanted to have different epochs for both forks.

packages/cli/test/simulation/fourNodesForks.test.ts
nazarhussain: Yes it works fine with 5m. We can compute the time but that would be a best case scenario, it something is broken and tests are not passing then it could timeout before assertions and we will not be able to see which assertions are failing.

packages/cli/test/simulation/singleNodeSingleValidatorMultipleEpoch.test.ts
nazarhussain: It's covering `missedBlocksAssertions` and `participationAssertions`, the `singleNodeForks` only covers the forks. As mentioned earlier I wanted it to be a scope specific test files, but it seems you want to merge them together. Will do that.

dapplion:
To ease immediate debugging of tests could you tweak rendering of:

Turn validator this logs 
```
Waiting for validator to start..
Waiting for validator to start...
Waiting for validator to start....
Waiting for validator to start.....
Waiting for validator to start......
Waiting for validator to start.......
Waiting for validator to start........
Waiting for validator to start.........
Waiting for validator to start..........
Waiting for validator to start...........
Waiting for validator to start............
Waiting for validator to start.............
Waiting for validator to start..............

```
Into just `Waiting for validators to start..` (plural) and log a single dot `.` to stdout on each sleep cycle.

Render the env object

```
%%%%%%%%% Starting test env %%%%%%%%%
{
  beaconNodes: 1,
  validatorClients: 1,
  validatorsPerClient: 128,
  altairEpoch: 0,
  bellatrixEpoch: 1000000000000,
  withExternalSigner: undefined
}
```
 in a single line

Logging participation observed by each node is redundant. Instead log the head of each node (root + slot) and then log participation of the node with the highest head.

dapplion:
AFAIK current tests fail as soon as some assertion fails. So if the test has to run for 5 epochs and participation is low in epoch 2 it will stop in epoch 2. A much better behavior for debugging would be always run until epoch 5 and then fail with the list of assertions that failed, for example:
- low participation in epoch 2,3,4,5
- missed slots in epochs 3,4,5

And render the table with full results above that

dapplion:
- Instead of running twice with external signer and not, for the multi node test run half of the validators with external signers. If they don't work properly the assertions will fail the test anyway
- Note: **you must run on bellatrix fork** before merging this PR, in all test cases bellatrix epoch is FAR_FUTURE
- On the multinode test all validators are running the same set of keys!

nazarhussain:
@dapplion 

> Into just Waiting for validators to start.. (plural) and log a single dot . to stdout on each sleep cycle.

Yes that's already there, you will see the trialing dashes while you run in terminal. But that does not work with the Github CI.

nazarhussain:
@dapplion  

> A much better behavior for debugging would be always run until epoch 5 and then fail with the list of assertions that failed

The tests runs till the checkpoint event. I will check if and why it's not logging all the failed assertions.

dapplion (approved):

PR #4511, opened 2022-09-06
Add feerecipient and gas_limit routes

### LINKED ISSUE #4392: Feature: Implement key manager API `gas_limit`

Implement the `gas_limit` API as per https://github.com/ethereum/keymanager-APIs/blob/master/apis/gas_limit.yaml and https://ethereum.github.io/keymanager-APIs/#/Gas%20Limit

### LINKED ISSUE #4393: Feature: Implement keymanager API `feerecipient`

Implement the `feerecipient` API as per https://github.com/ethereum/keymanager-APIs/blob/master/apis/fee_recipient.yaml and https://ethereum.github.io/keymanager-APIs/#/Fee%20Recipient

g11tech:
Add feerecipient and gas_limit routes

Passes the oapi spec test checker locally (since the opai release json is not available, constructed it locally, have asked for an alpha release on keymanager api so the api spec tests check can be enabled)
image

Closes #4392 Closes #4393

TODO:
 ~~- [ ] Enable oapi spec tests (once the keymanager alpha release is done making the json asset avaiilable to be fetched)~~ can be followed up in a separate PR as not show stopper
 - [x] Add persisting strategy (complexity is how to handle collisions with proposerSettingFile config, especially on restart
   - [x] not allow keymanager proposer updates if `proposerSettingsFile` provided
   - [x] update to keymanager should also persist configs in the validators dir (add a seprate topfolder for storing these)
   - [x] not let validator start with `proposerSettingsFile`  if there is proposer data saved in the keymanagar proposer dir unless provided with a flush flag: `--flushKeymanagerProposerConfigs` in which case the directory is deleted at startup
   - [x] add e2e tests


packages/api/src/keymanager/routes.ts
```diff
@@ -166,16 +175,40 @@ export type Api = {
...
 
   listRemoteKeys: {url: "/eth/v1/remotekeys", method: "GET"},
   importRemoteKeys: {url: "/eth/v1/remotekeys", method: "POST"},
   deleteRemoteKeys: {url: "/eth/v1/remotekeys", method: "DELETE"},
+
+  listFeeRecipient: {url: "/eth/v1/validator/{pubkey}/feerecipient", method: "GET"},
+  setFeeRecipient: {url: "/eth/v1/validator/{pubkey}/feerecipient", method: "POST"},
+  deleteFeeRecipient: {url: "/eth/v1/validator/{pubkey}/feerecipient", method: "DELETE"},
+
+  getGasLimit: {url: "/eth/v1/validator/{pubkey}/gas_limit", method: "GET"},
```
wemeetagain, line 209: should this be called `listGasLimit` to be consistent with the naming of other routes?

packages/validator/src/services/validatorStore.ts
```diff
@@ -191,18 +223,79 @@ export class ValidatorStore {
...
+    if (Number.isNaN(gasLimit) || gasLimit === 0) {
+      throw Error(`Gas Limit is not valid gasLimit=${gasLimit}`);
+    }
+
+    if (this.validators.has(pubkeyHex)) {
+      const validatorData = this.validators.get(pubkeyHex);
+      if (validatorData !== undefined) {
+        this.validators.set(pubkeyHex, {...validatorData, builder: {...validatorData.builder, gasLimit}});
+      } else {
+        throw Error(`ValidatorData for pubkey ${pubkeyHex} not set `);
```
wemeetagain: trailing space in string here and in other errors

packages/validator/src/services/validatorStore.ts
```diff
@@ -191,18 +223,79 @@ export class ValidatorStore {
...
 
   getGasLimit(pubkeyHex: PubkeyHex): number {
-    return (
-      (this.validators.get(pubkeyHex)?.builder || {}).gasLimit ??
-      this.defaultProposerConfig?.builder.gasLimit ??
-      defaultOptions.defaultGasLimit
-    );
+    if (this.validators.has(pubkeyHex)) {
+      return (
+        (this.validators.get(pubkeyHex)?.builder || {}).gasLimit ??
```
wemeetagain: can this be `this.validators.get(pubkeyHex)?.builder?.gasLimit ??`

packages/validator/src/services/validatorStore.ts
```diff
@@ -168,14 +168,46 @@ export class ValidatorStore {
   }
 
   getFeeRecipient(pubkeyHex: PubkeyHex): Eth1Address {
-    return this.validators.get(pubkeyHex)?.feeRecipient ?? this.defaultProposerConfig.feeRecipient;
+    if (this.validators.has(pubkeyHex)) {
```
wemeetagain: on all these methods, it might be better to use this pattern:
```ts
const validatorData = this.validators.get(pubkeyHex);
if (validatorData === undefined) {
  throw Error(...);
}
return validatorData.x.y.z;
```

packages/cli/src/cmds/validator/keymanager/impl.ts
```diff
@@ -15,10 +15,89 @@ import {
...
 export class KeymanagerApi implements Api {
-  constructor(private readonly validator: Validator, private readonly persistedKeysBackend: IPersistedKeysBackend) {}
+  constructor(
+    private readonly validator: Validator,
+    private readonly persistedKeysBackend: IPersistedKeysBackend,
+    private readonly proposerConfigWriteDisabled?: boolean
+  ) {}
+
+  private checkIfProposerWriteAvailable(): void {
+    if (this.proposerConfigWriteDisabled == true) {
```
wemeetagain: best to use === unless there's a good reason otherwise

packages/cli/src/cmds/validator/keymanager/impl.ts
```diff
@@ -15,10 +15,89 @@ import {
...
+      throw Error("proposerSettingsFile option activated");
+    }
+  }
+
+  async listFeeRecipient(pubkeyHex: string): ReturnType<Api["listFeeRecipient"]> {
+    try {
+      return {data: {pubkey: pubkeyHex, ethaddress: this.validator.validatorStore.getFeeRecipient(pubkeyHex)}};
+    } catch (e) {
+      throw Error((e as Error).message);
```
wemeetagain: why are you catching and rethrowing errors here? Is there a reason you want to overwrite the stacks?

packages/api/src/keymanager/routes.ts
g11tech: this is how the spec does it, and with the spec matcher there isn't much we can do :shrug: https://github.com/ethereum/keymanager-APIs/blob/master/apis/gas_limit.yaml#L2 as we have been trying to directly use operationId as our route name.

packages/cli/src/cmds/validator/keymanager/impl.ts
g11tech: i just followed convention from other methods, i think the this would return a 400 something error with the message as message. I will confirm this.

dapplion: This is the impl of the API, should not make assumptions like this. Fastify (rest server) is the one in charge of deciding to forward stack traces or not. The effect of your code is to drop the original stack trace, which will make errors way harder to debug. If this is done somewhere else, please point it out because it's a dangerous pattern

packages/cli/src/cmds/validator/options.ts
```diff
@@ -152,6 +161,12 @@ export const validatorOptions: ICliCommandOptions<IValidatorCliArgs> = {
     type: "string",
   },
 
+  flushKeymanagerProposerConfigs: {
+    description: "If keymanager's persisted proposer configs are found, delete them all!",
```
dapplion: This is a very strange flag, is this really necessary?

packages/validator/src/services/validatorStore.ts
```diff
@@ -168,14 +168,46 @@ export class ValidatorStore {
...
+      const validatorData = this.validators.get(pubkeyHex);
+      if (validatorData !== undefined) {
+        //const newValidatorData = {...validatorData, feeRecipient}
+        this.validators.set(pubkeyHex, {...validatorData, feeRecipient});
+      } else {
+        throw Error(`ValidatorData for pubkey ${pubkeyHex} not set `);
+      }
+    } else {
+      throw Error(`Validator pubkey ${pubkeyHex} not known`);
+    }
```
dapplion, line 187: Don't use has, then get. Call get then check for undefined. Please apply this comment to the entire PR

packages/validator/src/services/validatorStore.ts
```diff
@@ -191,18 +223,79 @@ export class ValidatorStore {
...
+  }
+
+  setGasLimit(pubkeyHex: PubkeyHex, gasLimitInput: string | number): void {
+    if ((typeof gasLimitInput !== "string" && typeof gasLimitInput !== "number") || `${gasLimitInput}`.trim() === "") {
+      throw Error("Not valid Gas Limit");
+    }
+    const gasLimit = Number(gasLimitInput);
+    if (Number.isNaN(gasLimit) || gasLimit === 0) {
+      throw Error(`Gas Limit is not valid gasLimit=${gasLimit}`);
+    }
```
dapplion: The ValidatorStore in not a place to parse unsanitized data. Please parse gasLimit into a usable typed value where it's collected from the user be it API or CLI

packages/validator/src/services/validatorStore.ts
```diff
@@ -191,18 +223,79 @@ export class ValidatorStore {
...
+
+        // if anything is set , i.e not default then return
+        if (
+          graffiti !== undefined ||
+          strictFeeRecipientCheck !== undefined ||
+          feeRecipient !== undefined ||
+          builder.enabled !== undefined ||
+          builder.gasLimit !== undefined
+        ) {
+          proposerConfig = {graffiti, strictFeeRecipientCheck, feeRecipient, builder};
```
dapplion: Does the ValidatorStore really need to know about the idea of a proposerConfig? Feels like it's the API role to construct this object from the other methods

packages/cli/src/cmds/validator/handler.ts
```diff
@@ -138,19 +148,40 @@ export async function validatorHandler(args: IValidatorCliArgs & IGlobalArgs): P
...
+    valProposerConfig = {proposerConfig: proposerConfigFromKeymanager, defaultConfig};
   } else {
-    valProposerConfig = {defaultConfig} as ValidatorProposerConfig;
+    // from Proposer Settings File
+    if (args.proposerSettingsFile) {
+      // parseProposerConfig will override the defaults with the arg created defaultConfig
+      valProposerConfig = parseProposerConfig(args.proposerSettingsFile, defaultConfig);
+    } else {
+      valProposerConfig = {defaultConfig} as ValidatorProposerConfig;
+    }
```
dapplion, line 190: Consider moving to a function below the handler `getValProposerConfig(args): ValidatorProposerConfig`

packages/cli/src/cmds/validator/keymanager/interface.ts
```diff
@@ -26,4 +27,9 @@ export interface IPersistedKeysBackend {
 
   /** Returns true some item is deleted from disk */
   deleteRemoteKey(pubkey: PubkeyHex): boolean;
+
+  writeProposerConfig(pubkey: PubkeyHex, proposerConfig: ProposerConfig | null): void;
+  deleteProposerConfig(pubkeyHex: PubkeyHex): void;
+  readAllProposerConfigs(): {[index: string]: ProposerConfig};
```
dapplion: I don't think the `All` part of the name is necessary. The plural at the end conveys enough meaning. It also breaks the name pattern of the other APIs

packages/cli/src/cmds/validator/options.ts
g11tech: this to to clear keymanager proposer configs, although name can be better :slightly_smiling_face: , any suggestions?

packages/cli/src/cmds/validator/keymanager/interface.ts
g11tech: sounds good, should i also change `readAllKeystores` to `readKeystores` from the previous work?

packages/cli/src/cmds/validator/options.ts
dapplion: What's the usecase? There's the delete API for that right?

packages/cli/src/cmds/validator/handler.ts
g11tech: you mean separate out this parseProposerConfig section from `getProposerConfigFromArgs`?

packages/cli/src/cmds/validator/options.ts
g11tech: the use case is you used keymanager api to make proposer data updates, and now want to discard it and switch to proposerFileConfig. Since now `proposerFileConfig` flag will error since we don't accept `proposerFileConfig` if there is keymanager persisted proposer data, this flag can be used to discard and clean keymanagaer persisted configs.

packages/cli/src/cmds/validator/keymanager/interface.ts
dapplion: Oh sorry brain fart, though this was the public rest API. naming is good :+1:

packages/validator/src/services/validatorStore.ts
g11tech: ummm i don't find the idea of ValidatorStore knowing about the proposer config odd as we also pass proposer data in the addSigner now. 
Other getters anyway return defaults if the that particular property is not assigned.

packages/cli/src/cmds/validator/handler.ts
```diff
@@ -138,19 +148,40 @@ export async function validatorHandler(args: IValidatorCliArgs & IGlobalArgs): P
...
-    valProposerConfig = parseProposerConfig(args.proposerSettingsFile, defaultConfig);
+
+  let valProposerConfig: ValidatorProposerConfig;
+  const proposerConfigFromKeymanager = persistedKeysBackend.readProposerConfigs();
+
+  if (Object.keys(proposerConfigFromKeymanager).length > 0) {
+    // from persistedBackend
+    if (args.proposerSettingsFile) {
+      throw new YargsError(
+        "Cannot accept --proposerSettingsFile since some proposer configs seemed to have been previously persisted via Keymanager api. Use --flushKeymanagerProposerConfigs in conjuction with --proposerSettingsFile if you want to discard them and instead use proposerSettingsFile"
```
wemeetagain:
```suggestion
        "Cannot accept --proposerSettingsFile since it conflicts with proposer configs previously persisted via the keymanager api. Use --flushKeymanagerProposerConfigs in conjuction with --proposerSettingsFile to discard them and instead use the proposerSettingsFile"
```

packages/cli/src/cmds/validator/options.ts
g11tech: removed the flag! and recommended manual deletion in case of conflict :+1:

dapplion (approved):
Looks good thanks for those sweet e2e tests :heart: 

Some comments that are addressable post merge:
- Should builder.enabled and builder.gasLimit be flattened in ProposerConfig?
- The logic of getProposerConfig says that if some property is set everything else won't be applied defaults. Is that expected behaviour? This function is only used in the API but not in the validator store to perform duties right?

g11tech:
> Looks good thanks for those sweet e2e tests heart
> 
> Some comments that are addressable post merge:
> 
>     * Should builder.enabled and builder.gasLimit be flattened in ProposerConfig?
> 
>     * The logic of getProposerConfig says that if some property is set everything else won't be applied defaults. Is that expected behaviour? This function is only used in the API but not in the validator store to perform duties right?

ummm actually, i should rename `getProposerConfig` to `getOverrideProposerConfig`, this is to only get the overrides to persist for keymanager.

Else, individual gets of `feeRecipient`, `gasLimit` which is what we actually use are backed by `defaults`

I like your idea idea on flattening the builder  params :+1: ,  how about `builder.enabled` and `builder.gasLimit`?

PR #4508, opened 2022-09-03
FIx archive finalized states db entry key typo

dapplion:
**Motivation**

Archive finalized states logic has a typo where it indexes states based on epoch instead of slot. That causes the archiver to not prune states as expected thus Lodestar DB grows faster than it should.

**Description**

- Use Slot to index archiveStates DB repository
- Add a new step on process start to prune existing states that do not align with configured buckets

Opening as draft to deploy and test before merging

Closes https://github.com/ChainSafe/lodestar/issues/4378
Closes https://github.com/ChainSafe/lodestar/issues/4417

PR #4514, opened 2022-09-07
Bump level db to 8.0.0

dapplion:
**Motivation**

- Testing https://github.com/ChainSafe/lodestar/pull/4508 discovered level has a new version 8.0.0 which includes types, a proper iterator and other goodies.

**Description**

- Bump level to 8.0.0
- Drop custom wrapper code for the iterator
- Use existing .keys() and .values() methods where possible

Functionality should be equivalent, only less wanky code to mantain

### DIFF
--- a/packages/db/test/unit/controller/level.test.ts
+++ b/packages/db/test/unit/controller/level.test.ts
@@ -1,3 +1,4 @@
+import {execSync} from "node:child_process";
 import {expect} from "chai";
 import leveldown from "leveldown";
 import all from "it-all";
@@ -20,6 +21,11 @@ describe("LevelDB controller", () => {
         else resolve();
       });
     });
+  });
+
+  it("test get not found", async () => {
+    const key = Buffer.from("not-existing-key");
+    expect(await db.get(key)).to.equal(null);
   });
 
   it("test put/get/delete", async () => {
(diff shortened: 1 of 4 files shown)


packages/db/test/unit/controller/level.test.ts
```diff
@@ -106,4 +113,33 @@ describe("LevelDB controller", () => {
...
+    const maxKey = Buffer.from([0xff]);
+
+    await db.batchPut(keys.map((key, i) => ({key, value: values[i]})));
+    await db.batchDelete(keys);
+
+    const sizeBeforeCompact = getDbSize();
+    await db.compactRange(minKey, maxKey);
+    const sizeAfterCompact = getDbSize();
+
+    expect(sizeAfterCompact).lt(sizeBeforeCompact, "Expected sizeAfterCompact < sizeBeforeCompact");
```
twoeths, line 131: how much disk space was saved with this test? I think we should expect some threshold here

twoeths (approved):
lgtm!

packages/db/test/unit/controller/level.test.ts
dapplion: I'm not that familiar with level internals so I think it's dangeours to do assumptions here. LevelDB behaviour is not trivial

PR #4519, opened 2022-09-07
Update default.env to current network values

philknows:
**Motivation**

This comment is outdated.

**Description**

This PR changes the comment text to ensure the user understands that `mainnet, `goerli` and `sepolia` are the currently supported networks. Other networks have been deprecated.

### DIFF
--- a/default.env
+++ b/default.env
@@ -1,5 +1,5 @@
 # To specify a specific network (defaults to mainnet) set this value.
-# Allowed values are: mainnet, gnosis, goerli, ropsten and sepolia. (Source for currently supported networks: /lodestar/packages/cli/src/networks/index.ts#L19)
+# Allowed values are: mainnet, gnosis, goerli, ropsten and sepolia. Source for currently supported networks: https://github.com/ChainSafe/lodestar/blob/unstable/packages/cli/src/networks/index.ts#L19
 LODESTAR_NETWORK=mainnet
 
 # Set a custom admin password to prevent having to signup.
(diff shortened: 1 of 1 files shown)


wemeetagain:
might be better to just link to the source. https://github.com/ChainSafe/lodestar/blob/unstable/packages/cli/src/networks/index.ts#L19

This list is already missing gnosis and ropsten.

philknows:
My only argument against putting a source link in there is that it's not necessarily accessible if someone is running Lodestar on a machine without a GUI. I can add the missing networks and the source link to cater to both needs though.

philknows:
Disregard my comment above, I think you meant the local source. It's been added now.

default.env
```diff
@@ -1,5 +1,5 @@
 # To specify a specific network (defaults to mainnet) set this value.
-# Allowed values are: mainnet and pyrmont (others may work, but are deprecated)
+# Allowed values are: mainnet, gnosis, goerli, ropsten and sepolia. (Source for currently supported networks: /lodestar/packages/cli/src/networks/index.ts#L19)
```
wemeetagain: can you use the full github link?

wemeetagain:
> Disregard my comment above, I think you meant the local source. It's been added now.

No I meant the github link. My feeling is that the vast majority of people who use machines w/o GUIs remote into those machines from machines with GUIs.

The idea is that, to the extent we can, we should deduplicate docs information, and rely on linking or auto-generation.
Otherwise we have to rely on remembering to synchronize things on updates.
And that's a lot more work, and we haven't historically been good at doing that.

wemeetagain (approved):

PR #4532, opened 2022-09-10
Make batch sizes dynamic for eth1 fetch of blocks/logs

g11tech:
**Motivation**
The batched requests of getBlocksByNumber to nethermind would timeout because it seems that the current code flow apparently whips out all 1000 requests in one go via Promise.all

This PR tries to control the concurrency and at any given time doesn't whip out more than `ETH1_CONCURRENCY` (set to 10) requests to the EL.

Seems to have resolved the problem for user: 
https://discord.com/channels/593655374469660673/593655641445367808/1018266744789418055

### DIFF
--- a/dashboards/lodestar_execution_engine.json
+++ b/dashboards/lodestar_execution_engine.json
@@ -1001,9 +1001,25 @@
           "interval": "",
           "legendFormat": "eth1_follow_distance_dynamic",
           "refId": "A"
+        },
+        {
+          "exemplar": false,
+          "expr": "lodestar_eth1_blocks_batch_size_dynamic",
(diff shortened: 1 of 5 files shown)


dapplion:
Oh damn, that's a fun side effect! To prevent this issue from happening again I think a unit test should be possible. You can create a mock provider and check that the pattern of calls is retried and divided as expected. Let me know if you think that's worth it.

dapplion:
Sorry hit the wrong button publishing the comment :pray: :heart:

wemeetagain (approved):
LGTM

the dynamic fetch is nifty

PR #4534, opened 2022-09-10
Increase stateCache size to fix headState missing in checkpoint sync start

### LINKED ISSUE #4523: Node notifier error  headState does not exist

When the execution layer client is not synced up yet and running Lodestar, the user experiences a `Node notifier error: headState does not exist` and output journal entries stop. Which is not useful to the end user.

Is there a better log output that is more relevant to a user of Lodestar? Can we output something that tells the user to wait until their EL client is finished syncing? Output resumes once the EL is synced, but during this time there is no context given to the user.

**Additional context**

This was noted when Somer was developing a Lodestar guide and provided this UX improvement suggestion for user friendliness.

UPDATE:
could be related to 
- https://github.com/ChainSafe/lodestar/issues/4005
instead of EL/CL syncing

g11tech:
In checkpoint sync start, the stateCache of 3 * 32 caches seem to be not sufficient since generally we start from finalized which is almost 3 epochs behind 
So before bn can sync blocks and generate more states nearer to head, the getHeadState functions seem to be running out of the states.

This PR does a quick-fix for expanding the stateCache's cache to 6 epochs (to give  margin on 3 epochs) so that bns can stay behind 3 epochs + while still trying to find/sync peers.
There could be a better solution, but this is a hotfix to the current issue users might run into.

(see the notifier crossing 96+ skipped slots here since last finalized checkpoint sync comforably without error) 
image

User issue also seems to have resolved.

Closes #4523

### DIFF
--- a/packages/beacon-node/src/chain/stateCache/stateContextCache.ts
+++ b/packages/beacon-node/src/chain/stateCache/stateContextCache.ts
@@ -6,7 +6,7 @@ import {IMetrics} from "../../metrics/index.js";
 import {MapTracker} from "./mapMetrics.js";
 import {stateInternalCachePopulated} from "./stateContextCheckpointsCache.js";
 
-const MAX_STATES = 3 * 32;
+const MAX_STATES = 6 * 32;
 
 /**
  * In memory cache of CachedBeaconState


packages/beacon-node/src/chain/stateCache/stateContextCache.ts
```diff
@@ -6,7 +6,7 @@ import {IMetrics} from "../../metrics/index.js";
 import {MapTracker} from "./mapMetrics.js";
 import {stateInternalCachePopulated} from "./stateContextCheckpointsCache.js";
 
-const MAX_STATES = 3 * 32;
+const MAX_STATES = 6 * 32;
```
twoeths, line 9: we need to investigate the impact of this with regards to the heap memory (which lead to `gc` percentage metrics). In the past when chain was not finalized for a while I saw the heap memory spiked at the same time.

dapplion (changes requested):
We must not merge this increase in cache size. The problem here is that the head state should never be dropped; period. This change will dramatically increase the change of OOM in Lodestar

g11tech:
> We must not merge this increase in cache size. The problem here is that the head state should never be dropped; period. This change will dramatically increase the change of OOM in Lodestar

:+1:  let me see how we can handle this particular case in a better way

g11tech:
fixed in https://github.com/ChainSafe/lodestar/pull/4562

PR #4535, opened 2022-09-12
Duplicate actions in the CI pipeline

### LINKED ISSUE #4352: Duplicate actions in the CI pipeline
**Describe the bug**

There are few actions in the CI pipeline, which duplicates with others. e.g. The tests runs twice on a PR when you push new commit.

**Expected behavior**

The CI tasks should not be duplicate on any case

**Steps to Reproduce**

- Create a PR 
- Push some commit to it

**Additional Information**

There are certain tasks in the PR which can be run in parallel to speed up the CI. e.g. Linting and Unit tests don't need to be run in the sequence. 

Some reference: 

https://github.com/marketplace/actions/skip-duplicate-actions

nazarhussain:
**Motivation**

Avoid running CI actions twice to save time. 

**Description**

Fix the duplicate CI actions. 

Closes  #4352

### DIFF
--- a/.github/workflows/test-browser.yml
+++ b/.github/workflows/test-browser.yml
@@ -1,6 +1,9 @@
 name: Browser tests
 
-on: [pull_request, push]
+on: 
+  pull_request:
+    types: [opened, reopened]
+  push:
 
 jobs:
   tests-main:
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -1,6 +1,9 @@
 name: E2E tests
 
-on: [pull_request, push]
+on: 
+  pull_request:
+    types: [opened, reopened]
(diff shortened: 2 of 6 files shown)


dapplion:
We must run on both events. While it looks like the same action runs twice, it's actually testing different commits
- Run 1 (push): on the HEAD commit of source branch
- Run 2 (pr): Github automatically merges HEAD commit of source branch into target branch, THEN runs CI

Please read about this detail on tooling such as bors (used by lighthouse) https://bors.tech/ which explain about this issue and even go further than running on push + pr.

Imagine a case where from unstable we have two branches
- branch A add a test that asserts that file X exists
- branch B removes file X

Depending on the timing of merging both PRs there's a risk of breaking (getting red CI) in unstable

PR #4538, opened 2022-09-12
Clean up cli logger args

dapplion:
**Motivation**

- Some clean-up and organizational changes from https://github.com/ChainSafe/lodestar/pull/4531

**Description**

- De-duplicate logic to disable logFile
- Move logFile option into logger options (do path manipulation outside of getXYZPaths)
- Move logger options to its own file. Now used by beacon validator and lightclient cmds
- Re-use validator.log file for slashing import export cmds
- Remove need to pass logger to LevelDbController


g11tech (approved):
looks awesome! much cleaner

PR #4545, opened 2022-09-14
Export beacon api client

wemeetagain:
It's useful (eg: for web use) to have an export that only loads client code


dapplion:
The index route does not export the server, it exports the client getter + the types (routes)

https://github.com/ChainSafe/lodestar/blob/0886d751c52a9fc9b306bb26a328c0408bbfcd7d/packages/api/src/beacon/index.ts#L3

PR #4548, opened 2022-09-15
isValidAttestationData returns invalid result for earlier epochs 

### LINKED ISSUE #4547: `isValidAttestationData` returns invalid result for earlier epochs
**Describe the bug**

When called `isValidAttestationData` for the an attestation which is targeting **last slot**  of the **second epoch** it results to zero. 

It's because of the following condition always evaluates to `-1` for which we don't have a block header. 

https://github.com/ChainSafe/lodestar/blob/0886d751c52a9fc9b306bb26a328c0408bbfcd7d/packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts#L435

**Expected behavior**

Instead of a hardcoded condition for first **2** epochs, we should simply rely if the `pivotSlot` is not the negative. That will suffice the few first epochs condition as well. 

https://github.com/ChainSafe/lodestar/blob/0886d751c52a9fc9b306bb26a328c0408bbfcd7d/packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts#L430-L432

- OS: Mac
- Branch: [unstable]

nazarhussain:
**Motivation**

Fix the `isValidAttestationData` to use the right pivot slot. 

**Description**

The pivot slot calculation for the `isValidAttestationData` was wrong and was causing issue for attestations in the first few epochs. 

Closes #4547

### DIFF
--- a/packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts
+++ b/packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts
@@ -3,6 +3,7 @@ import {
   ForkName,
   MAX_ATTESTATIONS,
   MIN_ATTESTATION_INCLUSION_DELAY,
+  MIN_SEED_LOOKAHEAD,
   SLOTS_PER_EPOCH,
   TIMELY_SOURCE_FLAG_INDEX,
 } from "@lodestar/params";
@@ -12,7 +13,7 @@ import {
   CachedBeaconStatePhase0,
   CachedBeaconStateAltair,
   computeEpochAtSlot,
-  computeStartSlotAtEpoch,
+  computeEndSlotAtEpoch,
   getBlockRootAtSlot,
 } from "@lodestar/state-transition";
 import {toHexString} from "@chainsafe/ssz";
@@ -427,15 +428,15 @@ export function isValidAttestationData(
 
   if (!ssz.phase0.Checkpoint.equals(data.source, justifiedCheckpoint)) return false;
 
-  // Shuffling can't have changed if we're in the first few epochs
-  if (stateEpoch < 2) {
-    return true;
-  }
   // Otherwise the shuffling is determined by the block at the end of the target epoch
   // minus the shuffling lookahead (usually 2). We call this the "pivot".
-  const pivotSlot = computeStartSlotAtEpoch(targetEpoch - 1) - 1;
-  const stateDependentRoot = toHexString(getBlockRootAtSlot(state, pivotSlot));
+  const pivotSlot = computeEndSlotAtEpoch(targetEpoch - MIN_SEED_LOOKAHEAD);
 
+  if (pivotSlot < 0) {
+    return true;
+  }
+
+  const stateDependentRoot = toHexString(getBlockRootAtSlot(state, pivotSlot));
   // Use fork choice's view of the block DAG to quickly evaluate whether the attestation's
   // pivot block is the same as the current state's pivot block. If it is, then the
   // attestation's shuffling is the same as the current state's.
--- a/packages/state-transition/src/util/epoch.ts
+++ b/packages/state-transition/src/util/epoch.ts
@@ -16,6 +16,13 @@ export function computeStartSlotAtEpoch(epoch: Epoch): Slot {
 }
 
 /**
+ * Return the starting slot of the given epoch.
+ */
+export function computeEndSlotAtEpoch(epoch: Epoch): Slot {
+  return (epoch + 1) * SLOTS_PER_EPOCH - 1;
+}
+
+/**
  * Return the epoch at which an activation or exit triggered in ``epoch`` takes effect.
  */
 export function computeActivationExitEpoch(epoch: Epoch): Epoch {


packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -781,9 +783,15 @@ export class ForkChoice implements IForkChoice {
     let block = this.getBlock(headBlockHash);
     if (!block) return null;
     const {slot} = block;
+    const blockEpoch = computeEpochAtSlot(slot);
+
     // The shuffling is determined by the block at the end of the target epoch
     // minus the shuffling lookahead (usually 2). We call this the "pivot".
-    const pivotSlot = computeStartSlotAtEpoch(computeEpochAtSlot(slot) - 1) - 1;
+    const pivotSlot = computeEndSlotAtEpoch(blockEpoch - MIN_SEED_LOOKAHEAD);
```
nazarhussain: As the description suggests we need to look head from the end of target epoch ( minus seed look ahead). Earlier expression was subtracting to two epoch.

packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -426,15 +427,15 @@ export function isValidAttestationData(
 
   if (!ssz.phase0.Checkpoint.equals(data.source, justifiedCheckpoint)) return false;
 
-  // Shuffling can't have changed if we're in the first few epochs
```
twoeths, line 430: I think we still need this condition

nazarhussain: I think that's not the case. e.g. Even if we have stateEpoch `9999` the `findAttesterDependentRoot` is just gonna check `two hopes` from the beacon head. So it can't allow attestations from the epoch 1, unless the two hopes reach to a slot in there.

packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -426,15 +427,15 @@ export function isValidAttestationData(
...
-  if (stateEpoch < 2) {
-    return true;
-  }
   // Otherwise the shuffling is determined by the block at the end of the target epoch
   // minus the shuffling lookahead (usually 2). We call this the "pivot".
-  const pivotSlot = computeStartSlotAtEpoch(targetEpoch - 1) - 1;
-  const statePivotBlockRoot = toHexString(getBlockRootAtSlot(state, pivotSlot));
+  const pivotSlot = computeEndSlotAtEpoch(targetEpoch - MIN_SEED_LOOKAHEAD);
 
+  if (pivotSlot < 0) {
```
twoeths, line 435: it's too relaxed to have this condition as a `stateEpoch` 1000 can include attestations from epoch 1

as the issue only happens for stateEpoch 2 and attestations in epoch 1 (I suppose?) I think we can just change the condition above to:

```typescript
if (stateEpoch <= 2) {
  return true;
}
```

nazarhussain: I think we would rather check that difference between `targetEpoch` and `stateEpoch` is not greater than `MIN_SEED_LOOKAHEAD + 1`. This makes more descriptive approach than hardcoding the `2` epoch. 

```ts

if (targetEpoch < stateEpoch - (MIN_SEED_LOOKAHEAD + 1)) {
   return false; 
}
```

nazarhussain: Lighthouse check this condition as follows that target epoch should be either current epoch or previous epoch. 

```rs
let pivot_slot =
            if target_epoch == state.previous_epoch() || target_epoch == state.current_epoch() {
                (target_epoch - shuffling_lookahead).end_slot(slots_per_epoch)
            } else {
                return false;
            };
```

https://github.com/sigp/lighthouse/blob/aa022f46855df2a1420a6a80a788c73dc2779aa7/beacon_node/beacon_chain/src/beacon_chain.rs#L2002

nazarhussain: @dapplion What's your opinion on this?

twoeths: > it's too relaxed to have this condition as a stateEpoch 1000 can include attestations from epoch 1

I missed the 1st check to make sure `targetEpoch` is only current epoch or previous epoch

so I think all ways below work

```typescript
  if (stateEpoch <= 2) {
    return true;
  }
```

or 

```typescript
  if (stateEpoch <= MIN_SEED_LOOKAHEAD + 1) {
    return true;
  }
```
or

```typescript
  if (pivotSlot < 0) {
    return true;
  }
```

I'm fine with whatever way, as long as we have a good comment (I prefer 1 & 2 thru)

packages/fork-choice/src/forkChoice/forkChoice.ts
twoeths: @nazarhussain  the code already goes with the description, if you want to use the new util, it should be

```typescript
const pivotSlot = computeEndSlotAtEpoch(blockEpoch - MIN_SEED_LOOKAHEAD - 1);
```

@dapplion could also help double check this

nazarhussain: I think, we need to start looking past in epoch with minimum seed look ahead value `MIN_SEED_LOOKAHEAD`. If we want to skip past one extra epoch we should change `MIN_SEED_LOOKAHEAD` param.

nazarhussain:
Closed as a lot of relevant changes done in other PR.

PR #4552, opened 2022-09-15
Duplicate actions in the CI pipeline

### LINKED ISSUE #4352: Duplicate actions in the CI pipeline

nazarhussain:
**Motivation**

We want to optimize the CI workflow to have quick developer feedback. 

**Description**

We need to run flow on pushing changes to stable/unstable branches. As long as the tests are passing after merging the branch it's not important for us if these passes on local branch or not. If required to run tests for the some branches which don't have a PR, you can use the manual trigger. 

Closes #4535

### DIFF
--- a/.github/workflows/test-browser.yml
+++ b/.github/workflows/test-browser.yml
@@ -1,6 +1,9 @@
 name: Browser tests
 
-on: [pull_request, push]
+on: 
+  pull_request:
+    types: [opened, reopened]
+  push:
 
 jobs:
   tests-main:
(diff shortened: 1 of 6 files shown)


dapplion (approved):
Agree with the rationale :+1:

PR #4553, opened 2022-09-16
Remove ChainEvent.forkChoiceHead event

dapplion:
**Motivation**

It's unnecessary to have this intermediary event between what the eventstream requires and where it's emitted. Splitting sensitive code like this has been a source of inconsistencies and bugs in the past. In the case for example it hide away this really ugly code

```ts
      const state = chain.stateCache.get(head.stateRoot);
      if (!state) {
        throw Error("cannot get state for head " + head.stateRoot);
      }

      const currentEpoch = state.epochCtx.epoch;
      const [previousDutyDependentRoot, currentDutyDependentRoot] = [currentEpoch - 1, currentEpoch].map((epoch) =>
        toHexString(getBlockRootAtSlot(state, Math.max(computeStartSlotAtEpoch(epoch) - 1, 0)))
      );
```

This code forces dangerous assumptions:
- When is the forkchoice:head event emitted? Is it emitted BEFORE or AFTER storing the state in the cache?
- Is there always a state available for the head? If not, under what circumstances? If it's not available should the event be delayed?

**Description**

- Remove intermediary head event
- Move event data logic into sequential importBlock flow

### DIFF
--- a/packages/beacon-node/src/api/impl/events/index.ts
+++ b/packages/beacon-node/src/api/impl/events/index.ts
@@ -1,16 +1,15 @@
-import {computeEpochAtSlot, computeStartSlotAtEpoch, getBlockRootAtSlot} from "@lodestar/state-transition";
+import {computeEpochAtSlot} from "@lodestar/state-transition";
 import {routes} from "@lodestar/api";
 import {toHexString} from "@chainsafe/ssz";
 import {ApiModules, IS_OPTIMISTIC_TEMP} from "../types.js";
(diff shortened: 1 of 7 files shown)


dapplion:
- Marked as draft to wait for https://github.com/ChainSafe/lodestar/pull/4555

and use 

```ts
      previousDutyDependentRoot: this.forkChoice.getDependantRoot(newHead, EpochDifference.previous),
      currentDutyDependentRoot: this.forkChoice.getDependantRoot(newHead, EpochDifference.current),
```

wemeetagain (approved):

PR #4382, opened 2022-08-06
Review dependantRoot usage

dapplion:
**Motivation**

General clean-up, drop the need to have an intermediary event and re-use code from findAttesterDependentRoot

**Description**

- Generalize `findAttesterDependentRoot` to support previous epoch and current epoch
- Use getDependantRoot to compute the head event dependant roots
- Cache shuffling dependant root and re-use in API
- Make dependantRoot hex in API to prevent unnecessary type conversions

---

@tuyennhv testing this branch with 

```
lodestar.js dev --startValidators 0..7 --api.rest.enabled --api.rest.api '*'
```

I get a lot of logs for re-orgs so some logic must be wrong somewhere between the validator consumer or beacon node producer. For every slot after epoch > 0
```
Eph 1/0 0.061[]                 warn: Potential next epoch attester duties reorg slot=8, dutyEpoch=2, priorDependentRoot=0xdca272c39cb493f10e3c400af6267632a78172f6636894d0e6166d25d8cca104, newDependentRoot=0x8649261fe405bcf7672e12d1b1a3e29aae3be28d228204d108af49c42bc6d87f
```
and for every slot after epoch > 1

```Eph 2/0 0.042[]                 warn: Potential next epoch attester duties reorg slot=16, dutyEpoch=3, priorDependentRoot=0x8649261fe405bcf7672e12d1b1a3e29aae3be28d228204d108af49c42bc6d87f, newDependentRoot=0x41fc2d22696e18c614a33dcc5d1a44d335a95fa5b48a2682c2c391678fe43a29
Eph 2/0 0.046[]                 warn: Potential current epoch attester duties reorg slot=16, dutyEpoch=2, priorDependentRoot=0xdca272c39cb493f10e3c400af6267632a78172f6636894d0e6166d25d8cca104, newDependentRoot=0x8649261fe405bcf7672e12d1b1a3e29aae3be28d228204d108af49c42bc6d87f
```

would you mind taking a look? I don't want to create a regression for the issue at https://github.com/ChainSafe/lodestar/issues/3211

PR #4554, opened 2022-09-16
Set dependantRoot type to string in api

dapplion:
**Motivation**

- spin off from https://github.com/ChainSafe/lodestar/pull/4382

Setting dependant root type to bytes requires unnecessary transformations. And it's something you only do comparisons with

**Description**

- Set dependantRoot type to string in api


g11tech (approved):
LGTM!

PR #4555, opened 2022-09-16
Replace findAttesterDependentRoot with getDependantRoot

dapplion:
**Motivation**

- Spin off from https://github.com/ChainSafe/lodestar/pull/4382

Generalize logic from `findAttesterDependentRoot` to work for both proposer and attester dependant root computation. To be used latter in API impl

**Description**

- Replace findAttesterDependentRoot with getDependantRoot

### DIFF
--- a/packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts
+++ b/packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts
@@ -16,7 +16,8 @@ import {
   getBlockRootAtSlot,
 } from "@lodestar/state-transition";
 import {toHexString} from "@chainsafe/ssz";
-import {IForkChoice} from "@lodestar/fork-choice";
+import {IForkChoice, EpochDifference} from "@lodestar/fork-choice";
+import {toHex} from "@lodestar/utils";
 import {MapDef} from "../../util/map.js";
 import {intersectUint8Arrays, IntersectResult} from "../../util/bitArray.js";
 import {pruneBySlot} from "./utils.js";
@@ -433,14 +434,20 @@ export function isValidAttestationData(
   // Otherwise the shuffling is determined by the block at the end of the target epoch
   // minus the shuffling lookahead (usually 2). We call this the "pivot".
   const pivotSlot = computeStartSlotAtEpoch(targetEpoch - 1) - 1;
-  const statePivotBlockRoot = toHexString(getBlockRootAtSlot(state, pivotSlot));
+  const stateDependantRoot = toHexString(getBlockRootAtSlot(state, pivotSlot));
 
   // Use fork choice's view of the block DAG to quickly evaluate whether the attestation's
   // pivot block is the same as the current state's pivot block. If it is, then the
   // attestation's shuffling is the same as the current state's.
   // To account for skipped slots, find the first block at *or before* the pivot slot.
-  const pivotBlockRoot = forkChoice.findAttesterDependentRoot(data.beaconBlockRoot);
-  return pivotBlockRoot === statePivotBlockRoot;
+  const beaconBlockRootHex = toHex(data.beaconBlockRoot);
+  const beaconBlock = forkChoice.getBlockHex(beaconBlockRootHex);
+  if (!beaconBlock) {
+    throw Error(`Attestation data.beaconBlockRoot ${beaconBlockRootHex} not found in forkchoice`);
+  }
+
+  const attestationDependantRoot = forkChoice.getDependantRoot(beaconBlock, EpochDifference.previous);
+  return attestationDependantRoot === stateDependantRoot;
 }
 
 function flagIsTimelySource(flag: number): boolean {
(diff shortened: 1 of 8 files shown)


packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -433,14 +434,20 @@ export function isValidAttestationData(
   // Otherwise the shuffling is determined by the block at the end of the target epoch
   // minus the shuffling lookahead (usually 2). We call this the "pivot".
   const pivotSlot = computeStartSlotAtEpoch(targetEpoch - 1) - 1;
```
nazarhussain, line 436: Suggest to create a utility function `computeEndSlotAtEpoch` and change the code to below. 

```ts
computeEndSlotAtEpoch(targetEpoch - MIN_SEED_LOOKAHEAD);
```

Also we don't need to go back in pat one extra epoch. We should just go back what that const `MIN_SEED_LOOKAHEAD` defined.

packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -773,40 +773,56 @@ export class ForkChoice implements IForkChoice {
...
-    throw Error(
-      `Not able to find attester dependent root for head ${headBlockHash}, slot ${slot}, pivotSlot ${pivotSlot}, last tried slot ${block.slot}`
-    );
+  getDependantRoot(block: ProtoBlock, epochDifference: EpochDifference): RootHex {
+    // The navigation at the end of the while loop will always progress backwards,
+    // jumping to a block with a strictly less slot number. So the condition `blockEpoch < atEpoch`
+    // is guaranteed to happen. Given the use of target blocks for faster navigation, it will take
+    // at most `2 * (blockEpoch - atEpoch + 1)` iterations to find the dependant root.
+
+    const beforeSlot = block.slot - (block.slot % SLOTS_PER_EPOCH) - epochDifference * SLOTS_PER_EPOCH;
```
nazarhussain, line 796: Would be better to improve the readability of code by using helper methods. I am from the philosophical group believing that **comments are the code smell** to improve better readable code. 

```
const beforeSlot = computeEndSlotAtEpoch(block.slot - epochDifference);
```

packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -773,40 +773,56 @@ export class ForkChoice implements IForkChoice {
...
+
+    // Special case close to genesis block, return the genesis block root
+    if (beforeSlot <= 0) {
+      const genesisBlock = this.protoArray.nodes[0];
```
nazarhussain, line 800: If that's the usage may be we make the getNodeFromIndex function public and then use it here.

packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -773,40 +773,56 @@ export class ForkChoice implements IForkChoice {
...
+      if (genesisBlock === undefined || genesisBlock.slot !== 0) {
+        throw Error("Genesis block not available");
+      }
+      return genesisBlock.blockRoot;
+    }
+
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
```
nazarhussain, line 808: The earlier logic was explicit of not looking too far in the history. Should be limiting it to MAX_SEED_LOOKAHEAD make sense here?

In any case there must be a breaking condition for such iterators.

nazarhussain (changes requested):

packages/fork-choice/src/forkChoice/forkChoice.ts
dapplion: `MAX_SEED_LOOKAHEAD` is a concept foreign to the fork-choice at this level. This function is generic with any EpochDifference, which is an enum with 0 or -1 just convenience

packages/fork-choice/src/forkChoice/forkChoice.ts
dapplion: Agree in general, each tool (including comments) must be used where appropiate. Your code snippet is wrong, mixing slot and epoch. The right implementation is:
```ts
const beforeSlot = computeEndSlotAtEpoch(computeEpochAtSlot(block.slot) - epochDifference))
```
which isn't that much readable TBH

packages/fork-choice/src/forkChoice/forkChoice.ts
dapplion: Yes, but `getNodeFromIndex` would throw `PROTO_ARRAY_ERROR_INVALID_NODE_INDEX`, while this implementation would throw `Genesis block not available`, which would be much easier to debug if a user reports. Would prefer to stick to this one-line re-implementation, as it's quite peculiar use of the nodes array

packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts
dapplion: This line is out of scope of this PR, would prefer to tackle such change in another PR

wemeetagain:
dependant -> dependent

nazarhussain (dismissed):

nazarhussain:
@wemeetagain @dapplion Interesting information. Seems both are valid spellings. 

image

nazarhussain (approved):

PR #4557, opened 2022-09-16
Use web EventSource if available

wemeetagain:
Current implementation of `EventSource`, from package `eventsource`, uses many nodejs imports and apis.
All of that complicates web usage of our api client and is unnecessary since there's a native implementation in-browser.

This PR adds a switch to use the native `EventSource` if it exists.


dapplion:
failed CI

twoeths (approved):

PR #4558, opened 2022-09-16
Improve validator UX

g11tech:
Users are not clear about whats going on with the current validator logging

Addressed some issues discovered during troubleshooting user setups


dapplion (approved):
Thank you!

PR #4560, opened 2022-09-17
Use consistent type-guards in blinded block checks

dapplion:
**Motivation**

Use type-guards consistently

**Description**

Use consistent type-guards in blinded block checks


twoeths (approved):

g11tech (approved):
LGTM!

PR #4562, opened 2022-09-17
Add strong ref to head state in state cache

dapplion:
**Motivation**

StateCache prune logic may drop head state if it gets too old. Also there's no mechanism to regen a head state if it's not available.
- https://github.com/ChainSafe/lodestar/issues/4523

**Description**

- Add strong ref to head state in state cache to prevent pruning
- Trigger regen if head state is not available only in importBlock

### DIFF
--- a/packages/beacon-node/src/chain/stateCache/stateContextCache.ts
+++ b/packages/beacon-node/src/chain/stateCache/stateContextCache.ts
@@ -1,5 +1,5 @@
 import {toHexString} from "@chainsafe/ssz";
-import {Epoch, Root, RootHex} from "@lodestar/types";
+import {Epoch, RootHex} from "@lodestar/types";
 import {CachedBeaconStateAllForks} from "@lodestar/state-transition";
 import {routes} from "@lodestar/api";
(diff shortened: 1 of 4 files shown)


twoeths (approved):

dapplion:
Deploying to `nogroup-ctvpss-0, nogroup-ctvpss-1` for testing

dapplion:
> Deploying to `nogroup-ctvpss-0, nogroup-ctvpss-1` for testing

It can follow the chain fine :+1:

packages/beacon-node/src/chain/stateCache/stateContextCache.ts
```diff
@@ -40,6 +45,10 @@ export class StateContextCache {
       return null;
     }
 
+    if (this.head?.stateRoot === rootHex) {
```
twoeths, line 48: when head state is not in `this.cache` it will not hit `this.head` check

this should be chained to line 43 above, this is why issue https://github.com/ChainSafe/lodestar/issues/4700 still happened

PR #4565, opened 2022-09-19
Overhaul blinded types into multi-fork

g11tech:
The blinded types currently used (for dev simplicity) are only of `bellatrix`

This PR overhauls blinded types into multi-fork in preparation for the `capella` hardfork

### DIFF
--- a/packages/api/src/beacon/routes/validator.ts
+++ b/packages/api/src/beacon/routes/validator.ts
@@ -445,7 +445,7 @@ export function getReturnTypes(): ReturnTypes<Api> {
     produceBlockV2: WithVersion((fork: ForkName) => ssz[fork].BeaconBlock),
     produceBlindedBlock: WithVersion((fork: ForkName) => {
       if (fork === ForkName.phase0 || fork === ForkName.altair) {
-        throw Error(`Invalid fork=${fork} for produceBlindedBlock`);
+        throw Error(`No BlindedBlock for fork ${fork} previous to bellatrix`);
       }
       return ssz[fork].BlindedBeaconBlock;
     }),
--- a/packages/api/src/beacon/routes/validator.ts
+++ b/packages/api/src/beacon/routes/validator.ts
@@ -169,7 +169,7 @@ export type Api = {
     slot: Slot,
     randaoReveal: BLSSignature,
     graffiti: string
-  ): Promise<{data: bellatrix.BlindedBeaconBlock; version: ForkName}>;
+  ): Promise<{data: allForks.BlindedBeaconBlock; version: ForkName}>;
 
   /**
    * Produce an attestation data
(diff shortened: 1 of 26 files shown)


packages/api/src/beacon/routes/validator.ts
```diff
@@ -443,7 +443,12 @@ export function getReturnTypes(): ReturnTypes<Api> {
     getSyncCommitteeDuties: ContainerDataExecutionOptimistic(ArrayOf(SyncDuty)),
     produceBlock: ContainerData(ssz.phase0.BeaconBlock),
     produceBlockV2: WithVersion((fork: ForkName) => ssz[fork].BeaconBlock),
-    produceBlindedBlock: WithVersion((_fork: ForkName) => ssz.bellatrix.BlindedBeaconBlock),
+    produceBlindedBlock: WithVersion((fork: ForkName) => {
+      if (fork === ForkName.phase0 || fork === ForkName.altair) {
```
dapplion, line 447: Use ForkSeq and < ForkSeq.bellatrix comparator

packages/api/src/beacon/routes/validator.ts
```diff
@@ -443,7 +443,12 @@ export function getReturnTypes(): ReturnTypes<Api> {
...
+        throw Error(`Invalid fork=${fork} for produceBlindedBlock`);
```
dapplion: Better error: `No BlindedBlock for fork ${fork} previous to bellatrix`

packages/api/src/beacon/routes/validator.ts
dapplion: Same for `packages/config/src/forkConfig/index.ts`

packages/state-transition/src/block/processBlockHeader.ts
```diff
@@ -29,25 +29,19 @@ export function processBlockHeader(state: CachedBeaconStateAllForks, block: allF
...
-    : types.BeaconBlockBody.hashTreeRoot(block.body);
-
+  const blockHeader = blindedOrFullBlockToHeader(state.config, block as allForks.FullOrBlindedBeaconBlock);
   // cache current block as the new latest block
+  // TODO: do we need to ZERO_HASH the stateRoot here??
   state.latestBlockHeader = ssz.phase0.BeaconBlockHeader.toViewDU({
-    slot: slot,
-    proposerIndex: block.proposerIndex,
-    parentRoot: block.parentRoot,
+    ...blockHeader,
```
dapplion: Would prefer not to spread here, map all properties manually. About ZERO_HASH, yes that's required by the spec

dapplion (changes requested):
Overall looks good! Some minor comments

packages/api/src/beacon/routes/validator.ts
g11tech: in `packages/config/src/forkConfig/index.ts` typescript can't infer with `seq` comparision that fork is not pre bellatrix causing type issues for accessing `ssz.allForksBlinded[forkName]`

g11tech: also same here as well

g11tech: ok figure it out thanks!

g11tech: duplicated the bellatrix blinded types for phase0 and altair for type consistency, the check here is no more required, while the check in  `packages/config/src/forkConfig/index.ts` is now using `seq` comparision. let me know if this approach is good

dapplion: Oh right, yeah then it's okay typesafety always wins

packages/types/src/allForks/sszTypes.ts
```diff
@@ -29,3 +29,27 @@ export const allForks = {
...
+  phase0: {
+    BeaconBlockBody: bellatrix.BlindedBeaconBlockBody,
+    BeaconBlock: bellatrix.BlindedBeaconBlock,
+    SignedBeaconBlock: bellatrix.SignedBlindedBeaconBlock,
+  },
+  // altair types would not be used, just set to have type consistency
+  altair: {
+    BeaconBlockBody: bellatrix.BlindedBeaconBlockBody,
+    BeaconBlock: bellatrix.BlindedBeaconBlock,
+    SignedBeaconBlock: bellatrix.SignedBlindedBeaconBlock,
```
dapplion: I actually prefer your original approach of using multiple ForkName comparisions and only defining the types for forks with blinded blocks. Sorry for the back and forth!

dapplion (approved):

PR #4573, opened 2022-09-20
Add assertions for the head and finality

### LINKED ISSUE #4276: Add assertions to sim tests

### LINKED ISSUE #4547: `isValidAttestationData` returns invalid result for earlier epochs

nazarhussain:
**Motivation**

Add assertions for the head and finality tracking per slot. 

Resolves #4276, Closes: https://github.com/ChainSafe/lodestar/issues/4547


packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -428,7 +428,8 @@ export function isValidAttestationData(
   if (!ssz.phase0.Checkpoint.equals(data.source, justifiedCheckpoint)) return false;
 
   // Shuffling can't have changed if we're in the first few epochs
-  if (stateEpoch < 2) {
+  // Also we can't look back 2 epochs if target epoch is 1 or less
+  if (stateEpoch < 2 || targetEpoch <= 1) {
```
nazarhussain: The following issue https://github.com/ChainSafe/lodestar/issues/4547  is fixed with above change.

packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts
```diff
@@ -428,7 +428,8 @@ export function isValidAttestationData(
   if (!ssz.phase0.Checkpoint.equals(data.source, justifiedCheckpoint)) return false;
 
   // Shuffling can't have changed if we're in the first few epochs
-  if (stateEpoch < 2) {
+  // Also we can't look back 2 epochs if target epoch is 1 or less
+  if (stateEpoch < 2 || targetEpoch < 2) {
```
nazarhussain, line 432: The following issue https://github.com/ChainSafe/lodestar/issues/4547  is fixed with above change.

twoeths (approved):
@nazarhussain the simulation tests look great!

PR #4575, opened 2022-09-20
Fix computed db checkpoint for weak subjectivity checks

g11tech:
**Motivation**
The checkpoint computed from db state for weak subjetcivity checks is wrongly based on its latestBlockHeader's slot's epoch
which causes a mismatch while comparing epochs in `isWithinWeakSubjectivityPeriod` and can cause a restarted node to be stuck if the last state written's latestBlockHeader's slot doesn't align

This PR fixes the calculation of a state's checkpoint epoch

### DIFF
--- a/packages/cli/src/cmds/beacon/initBeaconState.ts
+++ b/packages/cli/src/cmds/beacon/initBeaconState.ts
@@ -26,7 +26,7 @@ export function getCheckpointFromState(state: BeaconStateAllForks): Checkpoint {
   return {
     // the correct checkpoint is based on state's slot, its latestBlockHeader's slot's epoch can be
     // behind the state
-    epoch: computeCheckpointEpochAtStateSlot(state.slot / SLOTS_PER_EPOCH),
+    epoch: computeCheckpointEpochAtStateSlot(state.slot),
     root: getLatestBlockRoot(state),
   };
 }
--- a/packages/cli/src/cmds/beacon/initBeaconState.ts
+++ b/packages/cli/src/cmds/beacon/initBeaconState.ts
@@ -1,12 +1,7 @@
 import {ssz} from "@lodestar/types";
 import {createIBeaconConfig, IBeaconConfig, IChainForkConfig} from "@lodestar/config";
 import {ILogger} from "@lodestar/utils";
-import {
-  computeEpochAtSlot,
-  getLatestBlockRoot,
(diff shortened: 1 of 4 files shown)


packages/state-transition/src/util/weakSubjectivity.ts
```diff
@@ -111,7 +111,7 @@ export function isWithinWeakSubjectivityPeriod(
   wsState: BeaconStateAllForks,
   wsCheckpoint: Checkpoint
 ): boolean {
-  const wsStateEpoch = computeEpochAtSlot(wsState.slot);
+  const wsStateEpoch = Math.ceil(wsState.slot / SLOTS_PER_EPOCH);
```
twoeths: @g11tech do we really need this for the fix? it's quite confusing here so either not to include this change, or drop some comments to clarify

packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -23,7 +19,9 @@ import {IBeaconArgs} from "./options.js";
 
 export function getCheckpointFromState(state: BeaconStateAllForks): Checkpoint {
   return {
-    epoch: computeEpochAtSlot(state.latestBlockHeader.slot),
+    // the correct checkpoint is based on state's slot, its latestBlockHeader's slot's epoch can be
+    // behind the state
+    epoch: Math.ceil(state.slot / SLOTS_PER_EPOCH),
```
twoeths: using `state.slot` makes sense, thru I think `computeEpochAtSlot(state.slot)` is enough

if `SLOTS_PER_EPOCH` = 32 and state.slot = 40 then this function return an epoch of 2 does not sounds right, otherwise please drop some comments to avoid the confusing

also note that there is another `getCheckpointFromState` function defined somewhere else which expect the real checkpoint state (state.slot % 32 = 0)

g11tech: my reasoning: if state slot is >  some epoch boundary (ideally it should be on the boundary), and it its a checkpoint state, it should be for next epoch+1 only right? could be possible that the latestBlockHeader belongs from > epoch*32 slot, and hence the checkpoint root might belong to epoch+1
Also this: https://github.com/ChainSafe/lodestar/blob/unstable/packages/beacon-node/src/chain/initState.ts#L226

twoeths (approved):

Issue #3758, opened 2022-02-15
Remove deepmerge util


dapplion:
**Describe the bug**

Unnecessary and breaks things since ssz types can't be YOLO modified the way deepmerge does

dapplion:
Also introduces the risk of prototype pollution

wemeetagain:
@nflaig is this still relevant?

nflaig:
> ssz types can't be YOLO modified the way deepmerge does

not clear what code this is/was referencing, but we are not using deepmerge on ssz types right now and we don't use it on untrusted data, so prototype pollution is not a big issue

The only place I can see we run this on user input is in the CLI

https://github.com/ChainSafe/lodestar/blob/0d87c28aedb2e0556b027fc0714349542ff17542/packages/cli/src/config/beaconNodeOptions.ts#L38-L41

But not that we only run it on plain objects, this should prevent any security concerns + cli is trusted input by the user.

Other than that, we only use it in some tests / helpers

PR #4576, opened 2022-09-20
chore: typesafe options and cli args

### LINKED ISSUE #4577: All lodestar CLI commands are broken if not setting --dataDir
**Describe the bug**

Running lodestar beacon command breaks with

```
$ ./packages/cli/bin/lodestar.js beacon
 ✖ TypeError: The "path" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:372:5)
    at validateString (node:internal/validators:120:11)
    at Object.join (node:path:1172:7)
    at getCliLogger (file:///home/lion/Code/eth2.0/lodestar/packages/cli/src/util/logger.ts:75:43)
    at Object.beaconHandler [as handler] (file:///home/lion/Code/eth2.0/lodestar/packages/cli/src/cmds/beacon/handler.ts:32:18)
```

- The regression was introduced here https://github.com/ChainSafe/lodestar/pull/4538

https://github.com/ChainSafe/lodestar/blob/7ec526a34beedc773d99cd372b380de47afdd529/packages/cli/src/cmds/beacon/handler.ts#L32

`args` is typed with `dataDir` property being a non-optional string. However, the object generated by Yargs may not include that property.

**Expected behavior**

If `dataDir` can be undefined, it must typed as such

dapplion:
**Motivation**

Our current approach to options and CLI arg processing is not typesafe.

**Description**

_Problem_: CLI args and options are defined such that it's not guaranteed for the arg to not be undefined.
_This PR_: modifies `ICliCommandOptions` to ensure that either a default is set or the option must be marked as optional

_Problem_: Programmatic use of BeaconNode class has bad DX since many non-essential options must be provided
_This PR_: Changed options definitions such that BeaconNode class can be instantiated without proving any option

_Problem_: BeaconNodeOptions current default merging strategy is not typesafe. Options that are typed as not optional may receive undefined
_This PR_: Remove option deepmerging at the CLI level, and let's each subclass to handle default option setting without nesting

**Extra issues**

Fix typo bad option merging, mixing retryAttempts and retryDelay

https://github.com/ChainSafe/lodestar/blob/34305f34e86363223a88dfc66b0fd653d33105ec/packages/beacon-node/src/eth1/provider/jsonRpcHttpClient.ts#L129

Set maxPeers to a value that greater than targetPeers if only the latter is set

Closes https://github.com/ChainSafe/lodestar/issues/4577


dapplion:
Anyone understands what is this error?

```
 @chainsafe/lodestar: $ tsc -p tsconfig.build.json && yarn write-git-data
@chainsafe/lodestar: ../beacon-node/lib/api/options.d.ts:17:28 - error TS7016: Could not find a declaration file for module '@lodestar/api/lib'. '/home/runner/work/lodestar/lodestar/packages/api/lib/index.js' implicitly has an 'any' type.
@chainsafe/lodestar:   If the '@lodestar/api' package actually exposes this module, try adding a new declaration (.d.ts) file containing `declare module '@lodestar/api/lib';`
@chainsafe/lodestar: 17         api: (keyof import("@lodestar/api/lib").Api)[];
```

The file referenced `packages/beacon-node/lib/api/options.d.ts` is 
```ts
import { BeaconRestApiServerOpts } from "./rest/index.js";
export interface IApiOptions {
    maxGindicesInProof?: number;
    rest?: BeaconRestApiServerOpts & {
        enabled?: boolean;
    };
    version?: string;
}
export declare const defaultApiOptions: {
    maxGindicesInProof: number;
    rest: {
        enabled: boolean;
        port: number;
        address: string;
        bodyLimit: number;
        cors: string;
        api: (keyof import("@lodestar/api/lib").Api)[];
    };
};
```

and `@lodestar/api/lib/index.d.ts`, `packages/api/lib/index.d.ts` exists :shrug:

packages/cli/src/util/command.ts
```diff
@@ -1,6 +1,13 @@
 import {Options, Argv} from "yargs";
 
-export type ICliCommandOptions<OwnArgs> = Required<{[key in keyof OwnArgs]: Options}>;
+export type ICliCommandOptions<OwnArgs> = Required<
```
g11tech: this is some nifty typescript! :heart:

packages/beacon-node/test/utils/node/validator.ts
```diff
@@ -84,7 +84,7 @@ export async function getAndInitDevValidators({
 }
 
 export function getNodeApiUrl(node: BeaconNode): string {
-  const address = node.opts.api.rest.address || "localhost";
-  const port = node.opts.api.rest.port || 19596;
+  const address = node.opts.api?.rest?.address ?? "localhost";
+  const port = node.opts.api?.rest?.port ?? 19596;
```
g11tech: the beacon node api was being hosted at `19596` earlier

packages/beacon-node/test/utils/node/beacon.ts
```diff
@@ -47,39 +43,16 @@ export async function getDevBeaconNode(
...
-  options = deepmerge(
-    // This deepmerge should NOT merge the array with the defaults but overwrite them
-    defaultOptions,
-    deepmerge(
-      // This deepmerge should merge all the array elements of the api options with the
-      // dev defaults that we wish, especially for the api options
-      {
-        db: {name: tmpDir.name},
-        eth1: {enabled: false},
-        api: {rest: {api: ["beacon", "config", "events", "node", "validator"], port: 19596}},
```
g11tech: this endpoint has been removed, but validator still uses it

dapplion:
Split original PR into two parts:
- This one: ONLY ensure that CLI args have proper defaults. Such there are no surpises with undefined args
- Future 1: remove deepmerge defaults in favor a typesafe transparent default setting
- Future 2: make as non required BeaconNode class args optional so that it can actually be used programatically 

Previous branch with all parts before rebase: https://github.com/ChainSafe/lodestar/tree/dapplion/cli-typesafe-args_BACKUP

packages/cli/src/options/beaconNodeOptions/execution.ts
```diff
@@ -38,8 +38,7 @@ export const options: CliCommandOptions<ExecutionEngineArgs> = {
   "execution.urls": {
     description: "Urls to execution client engine API",
     type: "array",
-    defaultDescription:
-      defaultOptions.executionEngine.mode === "http" ? defaultOptions.executionEngine.urls.join(" ") : "",
+    default: defaultOptions.executionEngine.mode === "http" ? defaultOptions.executionEngine.urls.join(",") : "",
```
nflaig: `--execution.urls "http://192.168.1.125:8551,http://192.168.1.126:8551"` does currently not work

Either have to do
- `--execution.urls "http://192.168.1.125:8551" "http://192.168.1.126:8551"` or
- `--execution.urls "http://192.168.1.125:8551" --execution.urls "http://192.168.1.126:8551"`

See [discord](https://discord.com/channels/593655374469660673/743858262864167062/1101568514860515328), we might wanna consider supporting comma-separated for all values which support multiple URLs by adding custom `coerce` function

packages/beacon-node/src/execution/engine/index.ts
```diff
@@ -17,7 +17,7 @@ export type ExecutionEngineOpts =
   | ({mode?: "http"} & ExecutionEngineHttpOpts)
```
nflaig: We could consider making `mode` required to ensure it is set in default options, else default values determined by `mode` will be empty

https://github.com/ChainSafe/lodestar/blob/568a956b97a8b0813f0c63259773ee1f7ee29616/packages/cli/src/options/beaconNodeOptions/execution.ts#L41

packages/cli/src/options/beaconNodeOptions/execution.ts
dapplion: > See [discord](https://discord.com/channels/593655374469660673/743858262864167062/1101568514860515328), we might wanna consider supporting comma-separated for all values which support multiple URLs by adding custom coerce function 

We should definitely do that! Should it be part of this PR? or maybe a pre-step

packages/beacon-node/src/execution/engine/index.ts
dapplion: I think the root problem is that CLI args don't play well with this style of enum types. A better solution would be to export `defaultExecutionEngineHttpOpts` and use it here unconditionally

packages/cli/src/options/beaconNodeOptions/execution.ts
nflaig: sounds better to split this into another PR, merge it and rebase this PR

nflaig: Opened a PR, will rebase this branch afterwards
- https://github.com/ChainSafe/lodestar/pull/5660

packages/beacon-node/src/execution/engine/index.ts
nflaig: Updated in latest commit (https://github.com/ChainSafe/lodestar/pull/4576/commits/8c01e0229dcf449633b35780d9191b17ce88b756), I think this makes more sense as `defaultOptions.executionEngine.mode === "http"` is anyhow only statically evaluated

nflaig (approved):
Resolved all conflicts and re-reviewed it. Updated PR description to not close https://github.com/ChainSafe/lodestar/issues/3758.

dapplion:
Thank you so much @nflaig  :heart:

wemeetagain:

PR #4583, opened 2022-09-21
Use mix of local key manager and external signer for sim tests

### LINKED ISSUE #4276: Add assertions to sim tests

nazarhussain:
**Motivation**

Use mix of both local key manager and external signer for sim tests. 

Resolves #4276


packages/cli/test/utils/simulation/LodestarValidatorProcess.ts
```diff
@@ -58,6 +58,9 @@ export const LodestarValidatorProcess: ValidatorConstructor = class LodestarVali
     });
     this.secretKeys = validatorSecretKeys;
 
+    // Split half of the keys to external signer
+    this.externalSigner = new ExternalSignerServer(this.secretKeys.slice(0, this.secretKeys.length / 2));
```
g11tech: can we define and set the ranges to be divided locally and externally as validables (and the default split as half/half)

g11tech: also `this.rcConfig["externalSigner.url"] = this.externalSigner.url;` isn't required for this using this external signer?

g11tech: >also this.rcConfig["externalSigner.url"] = this.externalSigner.url; isn't required for this using this external signer?

oh ok since we now start keymanager to use this external signer

nazarhussain: @g11tech Done.

g11tech (approved):
lgtm

PR #4586, opened 2022-09-22
Implement forkchoice equivocating validators

### LINKED ISSUE #4303: Remove equivocating validators from fork choice consideration
- See https://github.com/ethereum/consensus-specs/pull/2845

twoeths:
**Motivation**

We want to disregard equivocating validators, see https://github.com/ethereum/consensus-specs/pull/2845

**Description**

- Handle onAttesterSlashing in forkchoice
- Call it when importing a block or gossip
- Make sure spec tests passed

Closes #4303

**TODO**

- [x] Right now upon restart, we fetch all slashed validators as equivocating indices, need to clarify that => we don't have any attestations initially and node has to resync so `equivocating_indices` should be built up through syncing
- [x] Do we need onProposerSlashing() handler in forkchoice? => no need since that validator could be active for a while and its attestation is still valid, after that when it's not active it's not part of the committee


g11tech (approved):
LGTM!

PR #4589, opened 2022-09-23
gossip attestation validation: handle no committee error

### LINKED ISSUE #4396: Not able to validate attestation: Requesting slot committee out of range
**Describe the bug**

Around the prater/goerli merge, we got

```
Aug-11 02:29:30.011[]                 info: Synced - slot: 3639747 - head: 3639747 0xaace…1edd - execution: valid(0x62bd…24b8) - finalized: 0x9469…c31e:113739 - peers: 50
Aug-11 02:29:41.278[NETWORK]         error: Gossip validation beacon_attestation threw a non-GossipActionError  Requesting slot committee out of range epoch: 113742 current: 113740
Error: Requesting slot committee out of range epoch: 113742 current: 113740
    at EpochContext.getShufflingAtEpoch (file:///usr/app/node_modules/@lodestar/state-transition/src/cache/epochContext.ts:720:13)
    at EpochContext.getShufflingAtSlot (file:///usr/app/node_modules/@lodestar/state-transition/src/cache/epochContext.ts:709:17)
    at getCommitteeIndices (file:///usr/app/node_modules/@lodestar/beacon-node/src/chain/validation/attestation.ts:287:56)
    at validateGossipAttestation (file:///usr/app/node_modules/@lodestar/beacon-node/src/chain/validation/attestation.ts:91:28)
    at runNextTicks (node:internal/process/task_queues:61:5)
    at listOnTimeout (node:internal/timers:528:9)
    at processTimers (node:internal/timers:502:7)
    at validateGossipAttestationRetryUnknownRoot (file:///usr/app/node_modules/@lodestar/beacon-node/src/network/gossip/handlers/index.ts:349:14)
    at beacon_attestation (file:///usr/app/node_modules/@lodestar/beacon-node/src/network/gossip/handlers/index.ts:182:28)
```

note that the sync status stated the head 3639747 at epoch 113742, not sure why the head epoch of 113740 was used to validate attestation

**Expected behavior**

No error

twoeths:
**Motivation**

When an attestation came from an out of synced node, our logs look like:
```
Aug-11 02:29:41.278[NETWORK]         error: Gossip validation beacon_attestation threw a non-GossipActionError  Requesting slot committee out of range epoch: 113742 current: 113740
Error: Requesting slot committee out of range epoch: 113742 current: 113740
    at EpochContext.getShufflingAtEpoch (file:///usr/app/node_modules/@lodestar/state-transition/src/cache/epochContext.ts:720:13)
    at EpochContext.getShufflingAtSlot (file:///usr/app/node_modules/@lodestar/state-transition/src/cache/epochContext.ts:709:17)
    at getCommitteeIndices (file:///usr/app/node_modules/@lodestar/beacon-node/src/chain/validation/attestation.ts:287:56)
    at validateGossipAttestation (file:///usr/app/node_modules/@lodestar/beacon-node/src/chain/validation/attestation.ts:91:28)
```
Two issues here:
- the log look dangerous to the user
- we did not handle no committee error well, and we treated this case as `IGNORE` instead of `REJECT`

**Description**

- Handle the no committee scenario in gossip attestation/aggregateAndProof validation: return `REJECT` to gossipsub
- Correct log level just in case we got non-attestation errors

Closes #4396

### DIFF
--- a/packages/beacon-node/src/chain/validation/aggregateAndProof.ts
+++ b/packages/beacon-node/src/chain/validation/aggregateAndProof.ts
@@ -85,18 +85,7 @@ export async function validateGossipAggregateAndProof(
       });
     });
 
-  let committeeIndices: number[];
-  try {
-    committeeIndices = getCommitteeIndices(attHeadState, attSlot, attIndex);
-  } catch (e) {
-    // this may come from an out-of-synced node, the spec did not define it so should not REJECT
-    // see https://github.com/ChainSafe/lodestar/issues/4396
-    throw new AttestationError(GossipAction.IGNORE, {
-      code: AttestationErrorCode.NO_COMMITTEE_FOR_SLOT_AND_INDEX,
-      index: attIndex,
-      slot: attSlot,
-    });
-  }
+  const committeeIndices: number[] = getCommitteeIndices(attHeadState, attSlot, attIndex);
 
   const attestingIndices = aggregate.aggregationBits.intersectValues(committeeIndices);
   const indexedAttestation: phase0.IndexedAttestation = {
--- a/packages/beacon-node/src/chain/validation/aggregateAndProof.ts
+++ b/packages/beacon-node/src/chain/validation/aggregateAndProof.ts
@@ -29,6 +29,7 @@ export async function validateGossipAggregateAndProof(
   const attData = aggregate.data;
   const attDataRoot = toHexString(ssz.phase0.AttestationData.hashTreeRoot(attData));
   const attSlot = attData.slot;
+  const attIndex = attData.index;
   const attEpoch = computeEpochAtSlot(attSlot);
   const attTarget = attData.target;
   const targetEpoch = attTarget.epoch;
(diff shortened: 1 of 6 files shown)


wemeetagain:
I think the error is happening when a node hasn't been getting blocks for an epoch, but manages to publish an attestation the next epoch.

Eg: node last received block in epoch 1, publishes an attestation in epoch 3.
Then when we're verifying the attestation, we use regen to get the state at epoch 1, and can't get the shuffling for epoch 3.

In order to process the attestation, we could use `regen.getBlockSlotState` to dial the state to the right epoch.

But I don't think we should just reject, since this may be a valid attestation.

twoeths:
> In order to process the attestation, we could use regen.getBlockSlotState to dial the state to the right epoch.

yeah that's exactly what we did in the past, there's a performance concern with it as we may need extra epoch transition just to process an attestation, and we changed to just use the state at block which make our node more stable. The main point for this issue is just for the log

> But I don't think we should just reject, since this may be a valid attestation.

agree, we should not reject, I'll change to ignore (same behaviour to earlier too)

dapplion (changes requested):
Instead of adding a try / catch would prefer to make getShufflingAtEpoch or a separate function return null if out of range. Then the validation code can throw the error directly if returned value is null

dapplion (approved):

PR #4595, opened 2022-09-25
Serialize the fcus in the batch processing of blocks

g11tech:
**Motivation**
Since we moved to the batch processing of the blocks, even though we fire fcUs in a particular order, observed that the EL might  receive them in a jumbled manner. This might or mightnot cause interuptions for beacon sync (observed for ethereumjs)

This PR serializes the fcUs by awaiting on the in import block, even if we ignore their failure (since the following fcUs should resolve EL heads) atleast ensuring we don't jumble fcus.

### DIFF
--- a/packages/beacon-node/src/chain/blocks/importBlock.ts
+++ b/packages/beacon-node/src/chain/blocks/importBlock.ts
@@ -276,7 +276,10 @@ export async function importBlock(
     const safeBlockHash = this.forkChoice.getJustifiedBlock().executionPayloadBlockHash ?? ZERO_HASH_HEX;
     const finalizedBlockHash = this.forkChoice.getFinalizedBlock().executionPayloadBlockHash ?? ZERO_HASH_HEX;
     if (headBlockHash !== ZERO_HASH_HEX) {
-      this.executionEngine.notifyForkchoiceUpdate(headBlockHash, safeBlockHash, finalizedBlockHash).catch((e) => {
+      // Call and wait on the fcU even though failure of this may not be critical for us to import
+      // ( and hence catch log error). Otherwise this may result in out of order fcU on the execution
+      // engine especially during sync of multiple blocks.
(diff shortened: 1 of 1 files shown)


packages/beacon-node/src/chain/blocks/importBlock.ts
```diff
@@ -276,7 +276,10 @@ export async function importBlock(
     const safeBlockHash = this.forkChoice.getJustifiedBlock().executionPayloadBlockHash ?? ZERO_HASH_HEX;
     const finalizedBlockHash = this.forkChoice.getFinalizedBlock().executionPayloadBlockHash ?? ZERO_HASH_HEX;
     if (headBlockHash !== ZERO_HASH_HEX) {
-      this.executionEngine.notifyForkchoiceUpdate(headBlockHash, safeBlockHash, finalizedBlockHash).catch((e) => {
+      // Call and wait on the fcU even though failure of this may not be critical for us to import
+      // ( and hence catch log error). Otherwise this may result in out of order fcU on the execution
+      // engine especially during sync of multiple blocks.
+      await this.executionEngine.notifyForkchoiceUpdate(headBlockHash, safeBlockHash, finalizedBlockHash).catch((e) => {
```
twoeths, line 282: maybe add this to a new job queue to make sure EL receive fcu requests in order? if we leave `await` here we'll add some more time to block processing time - below is goerli metric

Screen Shot 2022-09-26 at 08 53 34

cc @dapplion @wemeetagain

g11tech: makes sense, actually entire engine calls can/should be put in the job queue since the order there is important

g11tech: making the updates...

dapplion (changes requested):
As @tuyennhv says this should be handled in a queue

g11tech:
closing in favour of 
- https://github.com/ChainSafe/lodestar/pull/4651

Issue #4596, opened 2022-09-26
Invalid state root error with ZERO_HASH


twoeths:
**Describe the bug**

Investigate this log in goerli nodes:
```
Sep-26 00:48:25.493[chain]           ^[[31merror^[[39m: Block error slot=3970442 root=0x2656995d89a183e60a7199a3e0755af6155dc9c2991af44348bf5f3786e10c2b, expectedRoot=0x0000000000000000000000000000000000000000000000000000000000000000
Error: BLOCK_ERROR_INVALID_STATE_ROOT
    at verifyBlocksStateTransitionOnly (file:///usr/app/packages/beacon-node/src/chain/blocks/verifyBlocksStateTransitionOnly.ts:51:13)
    at BeaconChain.verifyBlocksInEpoch (file:///usr/app/packages/beacon-node/src/chain/blocks/verifyBlock.ts:70:7)
    at runMicrotasks (<anonymous>)
    at runNextTicks (node:internal/process/task_queues:61:5)
    at listOnTimeout (node:internal/timers:528:9)
    at processTimers (node:internal/timers:502:7)
    at BeaconChain.processBlocks (file:///usr/app/packages/beacon-node/src/chain/blocks/index.ts:71:68)
    at Timeout.JobItemQueue.runJob [as _onTimeout] (file:///usr/app/packages/beacon-node/src/util/queue/itemQueue.ts:92:22)
```

note that the expectedRoot is somehow ZERO_HASH, for Sep 25 it happened 78 times

```
grep -e expectedRoot=0x0000000000000000000000000000000000000000000000000000000000000000 -rn beacon-2022-09-25.log  | wc -l
78
```

**Expected behavior**

No error, or a good explanation of the error

twoeths:
all of the blocks have ZERO_HASH state root and invalid, if we check the block explorer they are all missed (not included in chain)

```
beacon-2022-09-26.log:973523:Sep-26 13:51:37.247[chain]           error: Block error slot=3974358 root=0x6f261166376ebb03595a2d488fedc5bfb372a51a9843ec52b13db3c71c9f858f, expectedRoot=0x0000000000000000000000000000000000000000000000000000000000000000
[...]

PR #4114, opened 2022-06-03
Update libp2p to ESM version

wemeetagain:
**Motivation**

We want to get the latest libp2p, gossipsub, etc updates.

**Description**

TODO:
- more type errors
- [x] https://github.com/multiformats/js-multiaddr/pull/247 - not needed
- [x] https://github.com/libp2p/js-libp2p-interfaces/issues/227
- [x] merge conflicts
- [x] configure libp2p to not break

Closes #4140
Closes https://github.com/ChainSafe/lodestar/issues/4043

PR #4599, opened 2022-09-27
Fix peer id log

twoeths:
**Motivation**
The peer id logged is incorrect when lodestar receives a gossip block

```
Received gossip block slot=3977407, root=0x7bc6…52f8, curentSlot=3977407, peerId=bafzaajiiaijcca2szy2eje5bz2y6w4mucwv3pe7gi33p2b3a3z5u2qszjzt6eqnzxq, delaySec=2.4609999656677246
```

**Description**

- We should use `peerId.toB58String()` api

part of #4596


wemeetagain:
Closing as this isn't needed any more after #4114 (there is no more `toB58String`)

PR #4604, opened 2022-09-29
Add optional forwardCheckpoint validation on normal sync

g11tech:
Add ability to provide a checkpoint to validate in normal forward sync

Previous work: 
- https://github.com/ChainSafe/lodestar/pull/3589

PENDING:
- some tests

### DIFF
--- a/packages/beacon-node/src/chain/chain.ts
+++ b/packages/beacon-node/src/chain/chain.ts
@@ -639,10 +639,16 @@ export class BeaconChain implements IBeaconChain {
   }
 
   private onForkChoiceFinalized(this: BeaconChain, cp: CheckpointWithHex): void {
-    this.logger.info("Fork choice finalized", {epoch: cp.epoch, root: cp.rootHex});
+    this.logger.verbose("Fork choice finalized", {epoch: cp.epoch, root: cp.rootHex});
     if (this.isWSCheckpointValidationPending && this.opts.forwardWSCheckpoint) {
       try {
         this.isWSCheckpointValidationPending = !this.forkChoice.verifyForwardCheckpoint(this.opts.forwardWSCheckpoint);
+        if (!this.isWSCheckpointValidationPending) {
+          this.logger.info("forwardWSCheckpoint found and validated on the chain", {
+            epoch: this.opts.forwardWSCheckpoint.epoch,
+            root: toHex(this.opts.forwardWSCheckpoint.root),
+          });
+        }
       } catch (error) {
         // Critical error, exit the chain
         this.processShutdownCallback(error as Error);
--- a/packages/beacon-node/src/chain/chain.ts
+++ b/packages/beacon-node/src/chain/chain.ts
@@ -10,6 +10,7 @@ import {
   isCachedBeaconState,
   Index2PubkeyCache,
   PubkeyIndexMap,
+  isWithinWeakSubjectivityPeriod,
 } from "@lodestar/state-transition";
 import {IBeaconConfig} from "@lodestar/config";
 import {allForks, UintNum64, Root, phase0, Slot, RootHex, Epoch, ValidatorIndex} from "@lodestar/types";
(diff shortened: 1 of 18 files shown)


packages/cli/src/cmds/beacon/initBeaconState.ts
```diff
@@ -93,6 +93,13 @@ export async function initBeaconState(
   logger: ILogger,
   signal: AbortSignal
 ): Promise<{anchorState: BeaconStateAllForks; wsCheckpoint?: Checkpoint}> {
+  if (
+    (args.checkpointState !== undefined || args.checkpointSyncUrl !== undefined) &&
+    args.forwardWSCheckpoint !== undefined
+  ) {
+    throw Error("forwardWSCheckpoint can not be used in conjuction with checkpointState or checkpointSyncUrl");
```
dapplion, line 100: simpler word choice
```suggestion
    throw Error("forwardWSCheckpoint can not be used together with checkpointState or checkpointSyncUrl");
```

packages/beacon-node/src/chain/chain.ts
```diff
@@ -637,6 +641,37 @@ export class BeaconChain implements IBeaconChain {
...
+          };
+          if (wssCheck) {
+            this.logger.info("forwardWSCheckpoint found and valid", wsData);
+          } else {
+            this.logger.warn("forwardWSCheckpoint found on chain but is stale, recommend using a fresh one", wsData);
+          }
+        }
+      } catch (error) {
+        // Critical error, exit the chain
+        this.processShutdownCallback(error as Error);
```
dapplion, line 671: Being such a critical action (kill lodestar completely) putting here in a generic try / catch feels too dangerous. I would really like to do
```ts
if (condition) {
  this.processShutdownCallback(error as Error);
}
```

packages/beacon-node/src/node/nodejs.ts
```diff
@@ -159,6 +160,47 @@ export class BeaconNode {
       createLibp2pMetrics(libp2p, metrics.register);
     }
 
+    if (opts.chain.forwardWSCheckpoint !== undefined) {
+      const {epoch, root} = opts.chain.forwardWSCheckpoint;
+      // lets just check if we have the root in the finalized block archive, if not found then we should be syncing it
+      const block = (await db.blockArchive.getByRoot(root))?.message;
```
dapplion, line 166: Could you modularize this as:

```ts
if (opts.chain.forwardWSCheckpoint !== undefined) {
  const block = (await db.blockArchive.getByRoot(root))?.message;
  performForwardWSCheckpointCheck(...args)
}
```
And then unit test `performForwardWSCheckpointCheck`? Since this logic is really critical could you explain each case and play around with the if statements to see if the can be un-nested?

packages/beacon-node/src/sync/sync.ts
```diff
@@ -114,6 +114,8 @@ export class BeaconSync implements IBeaconSync {
       (currentSlot < 0 ||
         // head is behind clock but close enough with some tolerance
         (headSlot <= currentSlot && headSlot >= currentSlot - this.slotImportTolerance)) &&
+      // If there is no pending checkpoint validation
+      !this.chain.isWSCheckpointValidationPending &&
```
dapplion, line 118: Can you explain this in more detail? The logic to consider when the node is synced or not is extremely critical, as it prevents the node from validating. When is this set to true, and when is it set to false? What's the risk of false positives?

packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -1198,6 +1199,42 @@ export class ForkChoice implements IForkChoice {
       () => this.fcStore.unrealizedJustified.balances
     );
   }
+
+  verifyForwardCheckpoint({epoch, root}: phase0.Checkpoint): boolean {
```
dapplion, line 1203: Can you add a nice long JSDoc comment linking to specs and explaining what this is doing and why? What's the meaning of the boolean being returned?

dapplion:
Stale PR, closing for now. Please re-open once all conflicts are fixed

Issue #4431, opened 2022-08-18
Update outdated NPM dependencies


philknows:
This issue is to track dependencies which are outdated and fixes are available for implementation. The following `lodestar-cli` dependencies are outdated:

- [ ] `ini`: Latest is `v3.0.0` https://www.npmjs.com/package/ini
- [x] `minimist`: Latest is `v1.2.6` https://www.npmjs.com/package/minimist
- [ ] `node-forge`: Latest is `v1.3.1` https://www.npmjs.com/package/node-forge
- [x] `tar`: Latest is `v6.1.1` https://www.npmjs.com/package/tar

dapplion:
- `ini`, `minimist`, `tar`: Fixable by recreating yarn.lock
- `node-forge`: Dependency of libp2p-crypto, blocked until https://github.com/ChainSafe/lodestar/pull/4427

Other old packages there because:
- Our own dependency on "electron": "^19.0.9", @nazarhussain why not v20?
- typedoc: to be removed with https://github.com/ChainSafe/lodestar/pull/4436

nazarhussain:
The `node-forge` will be updated once `@chainsafe/discv5`, `peer-id` and `libp2p` is updated to latest versions.

nazarhussain:
The `ini` package is deep dependency of `lerna`, so no need to update.

nazarhussain:
All other critical dependencies are fixed in the PR https://github.com/ChainSafe/lodestar/pull/4597

nazarhussain:
Still WIP.

philknows:
@nazarhussain Is there still a need to keep this specific issue open based on the comments and PRs above?

nazarhussain:
We can close this issue.

PR #4605, opened 2022-09-29
Update test related dependencies

nazarhussain:
**Motivation**

Update all the dependencies related to tests. Noticeable changes are: 

1. Upgrade all tests related dependencies. 
2. Use the `chai-as-promise` and `sinon-chai` as global setup instead of each test file 
3. Use proper assertions syntax from `sinon-chai` 
4. Add `eslint-plugin-mocha` and fixed all lint errors related to it. 

Relates: #4431 

Run all tests.


packages/beacon-node/src/network/interface.ts
```diff
@@ -34,7 +34,7 @@ export interface INetwork {
   /** Subscribe, search peers, join long-lived syncnets */
   prepareSyncCommitteeSubnets(subscriptions: CommitteeSubscription[]): void;
   reStatusPeers(peers: PeerId[]): void;
-  reportPeer(peer: PeerId, action: PeerAction, actionName?: string): void;
```
nazarhussain, line 37: This type was different than the implementation.

nazarhussain:
For most of the files which shows huge refactoring, its just adding the missing top level `describe`.

.eslintrc.js
```diff
@@ -209,6 +206,25 @@ module.exports = {
...
+        // https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-setup-in-describe.md
+        "mocha/no-setup-in-describe": "off",
+        // We use to split before in small isolated tasks
+        // https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-sibling-hooks.md
+        "mocha/no-sibling-hooks": "off",
+        // We need to disable because we disabled "mocha/no-setup-in-describe" rule
+        // TODO: Move all setup code to before/beforeEach and then disable async describe
+        // https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-async-describe.md
+        "mocha/no-async-describe": "off",
+      },
```
dapplion, line 229: Rule selection looks great! :+1:

packages/beacon-node/test/e2e/eth1/jsonRpcHttpClient.test.ts
```diff
@@ -1,343 +1,342 @@
...
-      requestListener: (req, res) => res.end(),
-      error: "connect ECONNREFUSED",
-    },
-    {
-      id: "Not a JSON RPC endpoint",
-      requestListener: (req, res) => {
-        res.setHeader("Content-Type", "text/html");
-        res.end("<html></html>");
+describe("jsonRpcHttpClient", () => {
+  describe("eth1 / jsonRpcHttpClient", function () {
```
dapplion: Is it really necessary to nest? Would prefer to keep the diff small even tho it has two top level describes :pray:

packages/beacon-node/test/perf/eth1/pickEth1Vote.test.ts
```diff
@@ -4,95 +4,97 @@ import {ContainerType, ListCompositeType} from "@chainsafe/ssz";
 import {newFilledArray, BeaconStateAllForks} from "@lodestar/state-transition";
 import {fastSerializeEth1Data, pickEth1Vote} from "../../../src/eth1/utils/eth1Vote.js";
 
-describe("eth1 / pickEth1Vote", () => {
-  const ETH1_FOLLOW_DISTANCE_MAINNET = 2048;
-  const EPOCHS_PER_ETH1_VOTING_PERIOD_MAINNET = 64;
-  const SLOTS_PER_EPOCH_MAINNET = 32;
-  const eth1DataVotesLimit = EPOCHS_PER_ETH1_VOTING_PERIOD_MAINNET * SLOTS_PER_EPOCH_MAINNET;
+describe("pickEth1Vote", () => {
+  describe("eth1 / pickEth1Vote", () => {
```
dapplion: Please do not nest unless it's really necessary

packages/beacon-node/test/unit/api/impl/beacon/blocks/utils.test.ts
```diff
@@ -90,14 +90,14 @@ describe("block api utils", function () {
...
+      expect(forkChoiceStub.getCanonicalBlockAtSlot).to.be.calledOnceWithExactly(2);
     });
 
     it("should resolve finalized slot", async function () {
       forkChoiceStub.getCanonicalBlockAtSlot.withArgs(2).returns(null);
       await resolveBlockId(forkChoiceStub, dbStub, "2").catch(() => {});
-      expect(forkChoiceStub.getCanonicalBlockAtSlot.withArgs(2).calledOnce).to.equal(true);
-      expect(dbStub.blockArchive.get.withArgs(2).calledOnce).to.equal(true);
+      expect(forkChoiceStub.getCanonicalBlockAtSlot).to.be.calledOnceWithExactly(2);
+      expect(dbStub.blockArchive.get).to.be.calledOnceWithExactly(2);
```
dapplion, line 100: Nice API! Thanks was not aware :+1:

packages/beacon-node/test/e2e/eth1/jsonRpcHttpClient.test.ts
nazarhussain: It's not recommended to have multiple top level describe. Here is the analogy for it. 

> Multiple describe blocks is often a sign that a test should be broken down to multiple files. One of the possible problems if having multiple suites is that, if you are, for example, going through tests one by one and focusing them (using describe.only), you might not notice describe block(s) that are down there at the bottom of a file which may lead to tests being unintentionally skipped.

Find more details here. 
https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/max-top-level-suites.md

packages/beacon-node/test/perf/eth1/pickEth1Vote.test.ts
nazarhussain: This is added because of the lint error from the recommended linting rules. See the above comment for details.

packages/beacon-node/test/unit/chain/forkChoice/forkChoice.test.ts
```diff
@@ -1,21 +1,20 @@
...
   getEffectiveBalanceIncrementsZeroed,
-  processSlots,
 } from "@lodestar/state-transition";
 import {toHexString} from "@chainsafe/ssz";
 import {generateSignedBlock} from "../../../utils/block.js";
 import {generateState} from "../../../utils/state.js";
 import {ChainEventEmitter, computeAnchorCheckpoint, initializeForkChoice} from "../../../../src/chain/index.js";
 import {createCachedBeaconStateTest} from "../../../utils/cachedBeaconState.js";
 import {generateValidators} from "../../../utils/validator.js";
+import {createIndexedAttestation, makeChild, runStateTransition} from "../../../utils/forkChoice.js";
```
dapplion: Utilities that are used only in a single test and will never be used else where can be defined in the same file. The benefit is all code is together, and if they are no longer used, the linter will detected it as not used and make you delete it. Is there a reason to move it out beyond organizational?

packages/beacon-node/test/unit/chain/opPools/aggregatedAttestationPool.test.ts
```diff
@@ -28,250 +28,258 @@ const validSignature = fromHexString(
...
-  // eslint-disable-next-line @typescript-eslint/naming-convention
-  const config = createIChainForkConfig(Object.assign({}, defaultChainConfig, {ALTAIR_FORK_EPOCH: altairForkEpoch}));
-  const originalState = generateCachedState({slot: currentSlot + 1}, config, true);
-  let altairState: CachedBeaconStateAllForks;
-  const attestation = generateAttestation({data: {slot: currentSlot, target: {epoch: currentEpoch}}});
-  const committee = [0, 1, 2, 3];
-  let forkchoiceStub: SinonStubbedInstance<IForkChoice>;
-  const sandbox = sinon.createSandbox();
+describe("aggregatedAttestationPool", () => {
+  describe("AggregatedAttestationPool", function () {
```
dapplion: Please revert nesting

dapplion (changes requested):
Some items of this PR are very valuable, but it got some scope creep. I would recommend to include in this PR:
- Bumping dependencies
- Adding linters, and fixing its errors
- Using proper sinon API
- Using single setup for chai promise

And relegate to other PRs (or not change)
- Organizational changes (move utils to other file, group describes)
- Other purely cosmetic changes

nazarhussain:
> Organizational changes (move utils to other file, group describes)

@dapplion These organizational changes are done because of the lint errors introduced after adding the mocha recommended lint plugin. See the details at this comment https://github.com/ChainSafe/lodestar/pull/4605#discussion_r983512031. It's not recommend to export functions from tests files, so why have to move those to specific utils files. 

If we include the changes as you recommended then we should fix the linting errors introduced, otherwise it's not beneficial to add lint plugins and rules.

packages/beacon-node/test/unit/chain/forkChoice/forkChoice.test.ts
nazarhussain: It's not recommended to export functions from test files. Earlier there was a function which was exported from the test file. 

> Test files should have only one purpose, which is testing a specific unit. Using exports could mean the test file is also used to provide and expose utility or library functionalities, instead those should be moved to separate files.

For details https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-exports.md

packages/beacon-node/test/e2e/eth1/jsonRpcHttpClient.test.ts
dapplion: Right I see their point but I disagree. Again that's a purely organizational decision. Would prefer to have that in a different PR and discuss there, else this PR may be left often longer than necessary. Can you disable that rule?

packages/beacon-node/test/unit/chain/forkChoice/forkChoice.test.ts
dapplion: Oh sorry, I missed the exported function below. Then please only move exported functions, i.e. the createAttestation one (plus dependant) only.

dapplion:
> > Organizational changes (move utils to other file, group describes)
> 
> @dapplion These organizational changes are done because of the lint errors introduced after adding the mocha recommended lint plugin. See the details at this comment [#4605 (comment)](https://github.com/ChainSafe/lodestar/pull/4605#discussion_r983512031). It's not recommend to export functions from tests files, so why have to move those to specific utils files.
> 
> If we include the changes as you recommended then we should fix the linting errors introduced, otherwise it's not beneficial to add lint plugins and rules.

Right I would only disable the only-one-describe rule, no-export rule is something already follow almost everywhere

packages/beacon-node/tsconfig.json
```diff
@@ -2,6 +2,6 @@
   "extends": "../../tsconfig.json",
   "exclude": ["../../node_modules/it-pipe"],
   "compilerOptions": {
-    "typeRoots": ["../../node_modules/@types", "./node_modules/@types", "../../types"]
```
dapplion, line 5: Why had this to be removed?

packages/utils/test/unit/math.test.ts
```diff
@@ -1,3 +1,4 @@
+import "../setup.js";
```
dapplion, line 1: Is this setup really necessary? This test is not using promises

dapplion (approved):

PR #4607, opened 2022-09-29
Fix dashboard variables

dapplion:
**Motivation**

The adhoc variable name is not consistent across all dashboards so when jumping between them the instance name is not propagated.

Also, if the adhoc variable datasource is another variable, for some reason the variable doesn't work. So I set the adhoc variable to the hardcoded variable that we use. This is unfortunate but fixes our deployment for now


g11tech (approved):
LGTM!

PR #4621, opened 2022-10-02
Review sim tests

dapplion:
**Motivation**

- Extend https://github.com/ChainSafe/lodestar/pull/4510 but moving out of mocha assertions for better control

**Description**

- Generalize design to make it simpler to add other clients
- Print succinct status messages during simulation run
- Improve metrics tracking to be more resilient, and not rely on clock for epoch participation
- Guarantee network is started in time


packages/cli/test/utils/simulation/eth2clients/index.ts
```diff
@@ -0,0 +1,18 @@
+import {BeaconProcessOpts, ValidatorProcessOpts, Eth2Client, SpwanOpts} from "./interface.js";
+import {prepareBeaconNodeLodestarArgs, prepareValidatorLodestarArgs} from "./lodestar.js";
+
+export function prepareBeaconNodeArgs(opts: BeaconProcessOpts): SpwanOpts {
+  if (opts.client === Eth2Client.lodestar) {
```
g11tech: future prep for multi clients, nice:+1:

packages/cli/test/utils/simulation/eth2clients/lodestar.ts
```diff
@@ -0,0 +1,101 @@
...
+    paramsFile: opts.configFilepath,
+    rest: true,
+    "rest.address": "127.0.0.1",
+    "rest.port": opts.restPort,
+    "rest.namespace": allNamespaces,
+    "sync.isSingleNode": opts.isSingleNode,
+    "network.allowPublishToZeroPeers": opts.isSingleNode,
+    eth1: false,
+    discv5: !opts.isSingleNode,
+    // TODO - TEMP: Use an actual EL latter
```
g11tech: yes we should be able to plug el later

g11tech (dismissed):
looks amazing! much needed overhaul and structure to simulation.  will let others review and weight in! :heart:

g11tech (dismissed):

nazarhussain (commented):
I have following observations. 

1. The output is not understable (at leat to me). When we started we had that single summary table approach, but decided to show stats for all nodes running. In the following output we can't see and compare each node to find if one issue is across the nodes or only on one. If that's not a requirement, we could simply the exiting implementation as well. 

```
time      | slo | block    | prop | attnP | incS | syncP | fiE | peers   
0/1 0.052 | 1   | 0x20fc[24](https://github.com/ChainSafe/lodestar/actions/runs/3170483871/jobs/5163093983#step:8:25) | 73   | 2     | 0    | -     | 0   | 3,3,3,3 
0/2 0.039 | 2   | 0x99da91 | 82   | 3     | 0    | -     | 0   | 3,3,3,3 
0/3 0.084 | 3   | 0x9d72d0 | 63   | 10    | ❌11 | -     | 0   | 3,3,3,3
```

2. I have a particular philosophy about the code base. The readability of the code the most important thing, as it have direct relation with the maintenance cost. So if we are using any specific testing framework (our case mocha and chai), if we open a test file it should look like a test file and it should make assertions the way our testing framework do. If we find some limitation in testing framework we can for sure identity and may be create our own, until then we should stick to the syntax and pattern from the framework. 

3. Data collection and data assertions should be two different domains and not be mixed together. 

4. The problem we have on hand is the the unstable assertions, so we should focus on solving that problem in the PR, if there is a need to change the implementation approach, rewrite we should first document it in the issue and then plan it in the sprint.

dapplion:
> The output is not understable (at leat to me)

We can add a guide in the code to understand the output. The goal for us (experienced with the code) to debug issues as quickly as possible

> but decided to show stats for all nodes running

Tracking the stats for all nodes at once doesn't scale and would quickly overflow the table. Tracking peer count takes very little space and is extremely useful. Tracking the head of all nodes takes more space and isn't that useful with my experience.

wemeetagain:
Whats the status of this PR @dapplion @nazarhussain 
Should this be closed since @nazarhussain has done some work refactoring the sim tests?

dapplion:
@nazarhussain would you mind checking if the ideas from here have been evaluated? Then please just succinctly list which are already in master and which ones should be ignored. Then happy to close this PR

nazarhussain:
I can figure out following customization to SIM tests in this PR. 

- [ ] Configurable Lodestar startup time in addition genesis delay 
- [ ] Custom error handler for the simulation environment 
- [ ] Prefix the process id to the logs 
- [x] Add  run clock to check for each slot on each node 
- [x] Support to add different CL clients 
- [x] Support to add different EL clients 
- [x] Custom table rendere 
- [x] Pass `ChainConfig` as parameter to the env 
- [x] Custom assertion structure

@dapplion  If you think the remaining items are important, we can create separate issues for it as with current code it would be easy to do it in new PRs.

PR #4622, opened 2022-10-02
Move map utils to utils package

dapplion:
**Motivation**

Map utils are useful for multiple packages, at least validator, beacon, cli

- Reduce duplicated code between packages
- Reduce diff from https://github.com/ChainSafe/lodestar/pull/4621

**Description**

Move map utils to utils package


g11tech (approved):
looks good!

PR #4629, opened 2022-10-03
Replace geth binary in sim merge tests with docker

g11tech:
Building geth takes time and resources (3-4 minutes of cloning + build), whill pulling geth image is so much easier. This would significantly reduce the merge sim time and will be followed up with nethermind build replacement if success.

geth build time:
image
image pull time:
image

similar for nethermind (will be addressed in a separate PR as docker setup for nethermind needs to be added)

### DIFF
--- a/.github/workflows/test-sim-merge.yml
+++ b/.github/workflows/test-sim-merge.yml
@@ -41,7 +41,7 @@ jobs:
       # </common-build>
 
       - name: Pull Geth
-        run: docker run $GETH_IMAGE --help
+        run: docker pull $GETH_IMAGE
 
       - name: Test Lodestar <> Geth interop
         run: yarn test:sim:merge-interop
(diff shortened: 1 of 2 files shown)


.github/workflows/test-sim-merge.yml
```diff
@@ -40,21 +40,15 @@ jobs:
...
-      # Install Geth merge interop
-      - uses: actions/setup-go@v2
-        with:
-          go-version: '1.17'
-      - name: Clone Geth merge interop branch
-        run: git clone -b master https://github.com/g11tech/go-ethereum.git && cd go-ethereum && git reset --hard $GETH_COMMIT && git submodule update --init --recursive
-      - name: Build Geth
-        run: cd go-ethereum && make
+      - name: Pull Geth
+        run: docker run $GETH_IMAGE --help
```
dapplion: Why not
```suggestion
        run: docker pull $GETH_IMAGE
```

dapplion (approved):

PR #4630, opened 2022-10-03
Use xxhash with salt for fastMsgIdFn

### LINKED ISSUE #4603: Improve fastMsgIdFn

As of v1.2.0 rc.0, fastMsgIdFn takes 1.57% of cpu time since the more mesh peers we have, the more fastMsgIdFn get called

```typescript
export function fastMsgIdFn(rpcMsg: RPC.IMessage): string {
  if (rpcMsg.data) {
    return Buffer.from(digest(rpcMsg.data)).subarray(0, 8).toString("hex");
  } else {
    return "0000000000000000";
  }
}
```
Screen Shot 2022-09-29 at 11 28 30

Instead of copying to Buffer and convert to hex from there, we can just create String directly from there

```typescript
const hash = digest(data);
String.fromCharCode(hash[0], hash[1], hash[2], hash[3], hash[4], hash[5], hash[6], hash[7]);
```

quick benchmark
```
  encoding.test.ts
    ✔ same to fastMsgIdFn                                              4.564126e+11 ops/s 0.002191000 ns/op   x0.000    3899710 runs   10.0 s
    ✔ new fastMsgIdFn                                                  6.195787e+11 ops/s 0.001614000 ns/op   x0.000    5125581 runs   10.1 s
```

it's ~30% faster

if we don't count the hash, `String.fromCharCode()` is 3x faster
```
✔ Buffer.from(hash).subarray(0, 8).toString                        9.216590e+11 ops/s 0.001085000 ns/op   x0.002    6948395 runs   10.1 s
✔ String.fromCharcode 8 first bytes                                3.134796e+12 ops/s 0.0003190000 ns/op   x0.010   15844892 runs   10.0 s
✔ hash64 bytes                                                     6.570302e+11 ops/s 0.001522000 ns/op        -    5346980 runs   10.0 s
```

dapplion:
**Motivation**

fastMsgIdFn() must produce a unique output per message within the message of the last few heartbeats. We can use any hash function so we should go for the cheapest possible with good safety.

sha256 on a node subscribed to all subnets it can take 0.5% of total CPU time.

xxhash is a cryptographically unsafe function that's has really good performance and good distribution of output. This PR also adds salt so an attacker can't pre-compute collisions.

**Description**

- Use xxhash with salt for fastMsgIdFn

Closes #4603

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "test:spec-fast": "lerna run test:spec-fast --no-bail",
     "test:spec-main": "lerna run test:spec-main --no-bail",
     "benchmark": "yarn benchmark:files 'packages/*/test/perf/**/*.test.ts'",
-    "benchmark:files": "LODESTAR_PRESET=mainnet NODE_OPTIONS='--max-old-space-size=4096 --loader=ts-node/esm' benchmark --config .benchrc.yaml",
+    "benchmark:files": "LODESTAR_PRESET=mainnet NODE_OPTIONS='--max-old-space-size=4096 --loader=ts-node/esm' benchmark --config .benchrc.yaml --defaultBranch unstable",
     "release:create-rc": "node scripts/release/create_rc.mjs",
     "release:tag-rc": "node scripts/release/tag_rc.mjs",
     "release:tag-stable": "node scripts/release/tag_stable.mjs",
--- a/packages/beacon-node/package.json
+++ b/packages/beacon-node/package.json
@@ -116,7 +116,6 @@
     "@libp2p/mplex": "^5.2.4",
     "@libp2p/peer-id-factory": "^1.0.18",
     "@libp2p/tcp": "^3.1.2",
-    "@multiformats/multiaddr": "^11.0.0",
     "@lodestar/api": "^1.1.0",
     "@lodestar/config": "^1.1.0",
     "@lodestar/db": "^1.1.0",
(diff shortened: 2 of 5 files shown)


twoeths:
the benchmark looks great, suggest to test this on a node for 1 day before merge

twoeths:
@dapplion  I deployed this branch to feat1 lg1k

twoeths:
this looks promising, initially it takes 200ms - 300ms `submitAttestationPool` request time while unstable (feat2) returns 1.5s-2s (both deployed at the same time, attestations are submitted right at 1/3 of slot)

Screen Shot 2022-10-04 at 17 07 20

this correlates to the memory issue with this branch

Screen Shot 2022-10-04 at 17 07 56

to avoid the String creation, we should try https://github.com/ChainSafe/js-libp2p-gossipsub/pull/355

dapplion:
This is a low-priority optimization, not worth risking a memory leak for this. @tuyennhv if you want confirm that number causes the leak or not else let's close this PR

twoeths:
closing this PR as it still has a memory issue after I tried numbered FastMsgIdFn version (200ms - 300ms - this branch vs 1.5s - 2s - unstable `submitAttesttionPool`  request time), at least it proves how well we can improve I/O lag issue at 1/3 of slot by improving `fastMsgIdFn`

PR #4639, opened 2022-10-05
Use xxhash for fastMsgIdFn

### LINKED ISSUE #4603: Improve fastMsgIdFn

twoeths:
**Motivation**

- Same motivation to #4630, fastMsgIdFn turns out to be very important for I/O lagged issue
- `xxhashjs` is way slower than `xxhash-wasm`, however it still double speed compared to our current hash function
- TODO: test for 12h or more to see if it has any memory issues like in `xxhash-wasm`

**Description**
- Use `xxhashjs`
- `fastMsgIdFn` to return a number

Closes #4603

Some first tests show good performance of `submitPoolAttestations` if we submit attestations right at 1/3 of slot
- This branch

Screen Shot 2022-10-05 at 11 08 10

- unstable

Screen Shot 2022-10-05 at 11 08 50


twoeths:
it has same memory issue due to data collision with `xx-hash` `h32`, see https://github.com/ChainSafe/js-libp2p-gossipsub/issues/356

PR #4641, opened 2022-10-05
Support content negotiation of light client data served via rest api

### LINKED ISSUE #4635: Support Content Negotiation for returning Light Client related data as either JSON or SSZ
The REST API for light client related data should also be able to return results as SSZ.

Ref: https://github.com/ethereum/beacon-APIs/pull/247
 
> Depending on the `Accept` header it can be returned either as JSON or SSZ-serialized bytes.

### LINKED ISSUE #4812: Align light client APIs with beacon-APIs
https://github.com/ethereum/beacon-APIs/pull/247
```diff
 diff --unified \
      <(curl -s "https://lodestar-goerli.chainsafe.io/eth/v1/beacon/light_client/updates?start_period=500&count=1" | jq) \
      <(curl -s "http://unstable.prater.beacon-api.nimbus.team/eth/v1/beacon/light_client/updates?start_period=500&count=1" | jq)
--- /dev/fd/11  2022-11-25 15:06:10
+++ /dev/fd/12  2022-11-25 15:06:10
@@ -1,12 +1,13 @@
-{
-  "data": [
-    {
+[
+  {
+    "version": "bellatrix",
+    "data": {
       "attested_header": {
-        "slot": "4097055",
-        "proposer_index": "274054",
-        "parent_root": "0x4987da1c0b42cf2633ee2f9b8d690b9040245a92f0fa8befaa3caaca0fc952e7",
-        "state_root": "0xe263d9b02eb03ea6854630d405ba351ffa264c13b5a69fcc83d6f7a6102f42a9",
-        "body_root": "0x74f334bfce2f0bf41c185f602f2761b09add84f249d02a94f85c0d63dc56fb93"
+        "slot": "4097477",
+        "proposer_index": "63871",
+        "parent_root": "0x9d087996c4179d17753af1c78e3c923fea283562f67c756c21db97379552996b",
+        "state_root": "0x6060993d4bed7ef680d181674b368f3ba4826cbf4a917add8b5fd4c6db44fe02",
+        "body_root": "0x1b48df98a57d496a543edddaad2f1efba06a9722b5075b675573dd722c65f8a0"
       },
       "next_sync_committee": {
         "pubkeys": [
@@ -527,31 +528,31 @@
       },
       "next_sync_committee_branch": [
         "0x20383b704227adbd41a2ba5e6344a261fac69faadb37323ab2aa1c4f58bf490e",
-        "0x57e9c5bf57a4210d7d2652e5b2d988f5eb6bea94aad14f7a8936b5eef4930cfd",
-        "0x50112f4a626d6ecd5e4d492ec087649a4bc1e9c5feb3341b5ed265dc7c7f93b5",
-        "0xd5a712758eb28ae961ece9f00530415cfea98ae53d0b2c134338d5afac78292d",
-        "0xac45a2e44099354cea7aa5ca37182a9a17fcbe6239c5d13230b4e47ccbea6993"
+        "0xdb48f8b78bef3096d6672da19c92743baddd1bbbfa44059ffc47f8ab3ba8a791",
+        "0x29719a872f27c9b602ca648daf9bde7f2121aef960683f4cbef8af0924337ee4",
+        "0x0de42b5c575d6a872c01eb8943308c3f9570674677a9b55a7d1e2748bdf602b6",
+        "0xd000cb9ebca43ea71bbc998a613c7ad0dc9c06548b35601f084b7cafb0d1aa5e"
       ],
       "finalized_header": {
-        "slot": "4096960",
-        "proposer_index": "135111",
-        "parent_root": "0x8f55fddf1801a59c9903e93b7de723763525667e77207670d8f7282cfdb67d18",
-        "state_root": "0x1602f6c3c7a024294101485f3dd4da4584176224aea016db9d0595f708e0a216",
-        "body_root": "0x16126bf0285422e60ba3c1926ca7bb21204e59e753738b6db9d90f8436aee6fa"
+        "slot": "4097408",
+        "proposer_index": "40202",
+        "parent_root": "0x0e7ec3696acd3db042c7c8b1c9a216df4e7fc509bb4d77c2ec6ae8cdf31d2eae",
+        "state_root": "0x360abd21d0b179c698dae8bfdb8a2aa1a0cb9279ed8dcd125440e12cb705aedb",
+        "body_root": "0xf1f579f9b5fd334713b9cac285a4ffa0ab5a065309543ce6a5c0a3befc6733ad"
       },
       "finality_branch": [
-        "0x1ef4010000000000000000000000000000000000000000000000000000000000",
+        "0x2cf4010000000000000000000000000000000000000000000000000000000000",
         "0xb9e27e742c646ed456256e87a5dd7d1c9b0806813bcf5417226eb9270a394cd0",
         "0xfa2ee10325d0d3fbb385ea2d1c020d98a2e7ac5e56887f741fb493f6dd5ff503",
-        "0x50112f4a626d6ecd5e4d492ec087649a4bc1e9c5feb3341b5ed265dc7c7f93b5",
-        "0xd5a712758eb28ae961ece9f00530415cfea98ae53d0b2c134338d5afac78292d",
-        "0xac45a2e44099354cea7aa5ca37182a9a17fcbe6239c5d13230b4e47ccbea6993"
+        "0x29719a872f27c9b602ca648daf9bde7f2121aef960683f4cbef8af0924337ee4",
+        "0x0de42b5c575d6a872c01eb8943308c3f9570674677a9b55a7d1e2748bdf602b6",
+        "0xd000cb9ebca43ea71bbc998a613c7ad0dc9c06548b35601f084b7cafb0d1aa5e"
       ],
       "sync_aggregate": {
-        "sync_committee_bits": "0xfbd3dfc7ff5ffff2ff7dddf4fee2f763d6fdfff6ff7fcbf7e2ffdfdf8bbbfefdfffff77ffffb6ff5faffbff7e7ff1eefaf5affbef77ff7febfff9c5ff7defd5d",
-        "sync_committee_signature": "0x99ecf0563b315823799550b35daf8569bd9091c8a4e6f2d4f33ef13fc49c00f524b38a0530f6774d0171b28aae9d005d109df81589618f21530ae5129f243a4dedd5f6b830ab40d39a25d85963a8451a1f84da71d4b6a70637cf91517b58b89f"
+        "sync_committee_bits": "0xfbd3dfc7ff5ffff3ff7dddb4fee2f77fd6fdfff6ff7fcbf7e2ffdfdf8bbbfefdfffff77ffffb6ff5faffbff7e7ff1eefaf5affbef77ff7febfff9c5ff7defd5d",
+        "sync_committee_signature": "0xb58cc16285825dff3e62d254ca5dc8ba58f3c2192b8dee3361eaabd0416eda471384d7ca62f11fc31adfb666988be1a10b82d58ce8c2b2677b037c356d7e74f6d0af970c1dfcc9ffe7770d1790b794edafac8f97b5c47e4cfa1f468f4a2a4c5d"
       },
-      "signature_slot": "4097056"
+      "signature_slot": "4097481"
     }
-  ]
-}
+  }
+]

```

dadepo:
**Motivation**

The REST API for light client related data should also be able to return results as SSZ.

Closes #4635

This PR also includes changes to align the JSON object returned by the API to include version.

Closes #4812


packages/api/src/beacon/routes/lightclient.ts
```diff
@@ -80,17 +85,40 @@ export function getReqSerializers(): ReqSerializers<Api, ReqTypes> {
...
     },
 
-    getOptimisticUpdate: reqEmpty,
-    getFinalityUpdate: reqEmpty,
+    getOptimisticUpdate: {
+      writeReq: (format) => ({
+        headers: {accept: format === "ssz" ? mimeTypeSSZ : ""},
+      }),
+      parseReq: ({headers}) => [headers.accept === mimeTypeSSZ ? "ssz" : "json"],
+      schema: {},
```
dadepo: In this scenario, is it correct to have the schema as `{}`? If not what should the definition look like?

packages/beacon-node/src/api/impl/lightclient/index.ts
```diff
@@ -16,6 +18,23 @@ export function getLightclientApi(
   // We want some some sort of resistance against this DoS vector.
   const maxGindicesInProof = opts.maxGindicesInProof ?? 512;
 
+  const serializeSSZArray = (chunks: LightClientUpdate[]): Uint8Array => {
```
dadepo: Is there a more ssz way to perform these serialization?

twoeths: using `ssz.altair.LightClientUpdate.serialize()` for each LightClientUpdate is correct
however the way we serialize there may not be correct, it should have some length prefixed data + fork_digest in order for the client to parse:
```
(
                response_chunk_len: Little-endian Uint64 byte length of `response_chunk`
                response_chunk: (
                  context: 4 byte `ForkDigest`
                  payload: SSZ serialized payload bytes
                )
```

we need to have some encode/decode separate functions and make sure it works in unit tests first

packages/api/test/unit/beacon/testData/lightclient.ts
```diff
@@ -57,7 +57,7 @@ export const testData: GenericServerTestCases<Api> = {
     },
   },
   getBootstrap: {
-    args: [toHexString(root)],
+    args: [toHexString(root), "json"],
```
dapplion, line 37: You must add unit tests for the ssz variant. Note just setting this `"ssz"` won't actually touch that code path. Double check code is being tested in your approach

packages/beacon-node/src/api/impl/lightclient/index.ts
dadepo: Hey @tuyennhv, I took some time to dig into `ssz` and found `ListCompositeType` which I was able to use to decode/encode the response. I added test to confirm this. Seems this should be ok? Let me know if you have any other observation/comments.

packages/api/src/beacon/server/lightclient.ts
```diff
@@ -24,5 +25,53 @@ export function getRoutes(config: IChainForkConfig, api: Api): ServerRoutes<Api,
         return Buffer.from(serializeProof(proof));
       },
     },
+    getBootstrap: {
+      ...serverRoutes.getBootstrap,
```
dadepo, line 19: tried de-duplicating some of the code here, but was a bit tricky getting the types to line up right. If this is a blocker, then I'll revisit.

packages/beacon-node/src/api/impl/lightclient/index.ts
twoeths: @dadepo  no, that is not correct as ListCompositeType is not mentioned in the spec https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/light_client/updates.yaml

as noted above and in the spec, it's something like:

```typescript
const result: Buffer[] = [];
for (const update of lightClientUpdates) {
  const payload = ssz.LightClientUpdate.serialize(update);
  const slot = update.attestedHeader.slot;
  const forkName = config.getForkName(slot);
  const context = config.forkName2ForkDigest(forkName);
  const responseChunk = Buffer.concat([context, payload]);
  result.push(Buffer.concat[Buffer.from(varint.encode(responseChunk.length)), responseChunk]);
}

return Buffer.concat(result);
```

dadepo: @tuyennhv Thanks for the clarification. I misinterpreted how the serialisation/deserialisation was supposed to work. I have updated the implementation.

twoeths: oh I misread it, it said `Little-endian Uint64 byte length of response_chunk` so maybe no varint here, just encode/decode with little-endian uint64, in lodestar we have `UintNum64` ssz type so just use it

cc @wemeetagain if you also read the spec the same way

wemeetagain: yeah agree with your reading @tuyennhv 
You also could do the concat in one step.
`Buffer.concat([ssz.UintNum64.serialize(payload.length + 4), context, payload])`

dapplion (changes requested):
While this PR fullfils it's intent it's re-using an API (that I introduced in /debug/beacon/states) that doesn't scale really well.

If this feature is not urgent, since it's technically just a bandwidth optimization for the lightclient. We can sit on this and look for a design of dual JSON / SSZ that scales to multiple routes without so much boilerplate code.

There's an open discussion to support dual JSON / SSZ on the engine API + beacon API so we may want to wait for that

dadepo:
> While this PR fullfils it's intent it's re-using an API (that I introduced in /debug/beacon/states) that doesn't scale really well.
>
> There's an open discussion to support dual JSON / SSZ on the engine API + beacon API so we may want to wait for that

@dapplion do you know when this discussion will be concluded? Because I do not think it makes sense to keep this PR indefinitely at limbo. 

My suggestion is to review and merge so at least we have complaint with this part of the API spec. The optimisation of the API you introduced can follow later.

dapplion:
This PR has two different scopes:
1. Be able to return results as SSZ
2. Align the JSON object returned by the API to include version.

Point 1. I do not like the approach since it doesn't scale and we want to eventually support many more routes with the same functionality. Point 1. is not urgent and can wait to figure out on other design

Point 2. is a necessary bug fix to align with the spec, which is (a) urgent, and (b) clearly defined so it can be merged immediately.

In general having PRs with multiple scopes is not great but in this case it's an obvious example to split so we can get the bug fix asap.

dadepo:
Both of these changes are bug fixes, in the sense that both are needed due to the fact that the spec got updated/properly defined, and our implementation needed to be modified to be compliant with the updated spec. 

But If you think one is more "buggier" than the other and deserves urgent attention, while the other does not? Then I guess that's fine. Only that i would go about it differently. I would have the "less than perfect changes" reviewed and merged now, instead of waiting to figure out an improved design, while we keep differing from the spec.

But I guess here we disagree. 

Anyways, here is a separate PR https://github.com/ChainSafe/lodestar/pull/4877 that adds versioning to the LC API. Please help review 🙏

dapplion:
Closing as this approach does not scale for reasons outlined above

PR #4643, opened 2022-10-05
Fix datadir CLI arg type

### LINKED ISSUE #4577: All lodestar CLI commands are broken if not setting --dataDir

dapplion:
**Motivation**

Fix type inconsistency in datadir CLI arg

- https://github.com/ChainSafe/lodestar/pull/4576 is the proper long term solution, but merging this first for upcoming release

**Description**

- Set dataDir to `string | undefined`
- Add test to ensure dev command can start

Closes https://github.com/ChainSafe/lodestar/issues/4577

### DIFF
--- a/packages/cli/src/cmds/beacon/handler.ts
+++ b/packages/cli/src/cmds/beacon/handler.ts
@@ -30,7 +30,7 @@ export async function beaconHandler(args: IBeaconArgs & IGlobalArgs): Promise<vo
   mkdir(beaconPaths.dbDir);
 
   const abortController = new AbortController();
-  const logger = getCliLogger(args, {defaultLogFile: "beacon.log"}, config);
+  const logger = getCliLogger(args, {defaultLogFilepath: path.join(beaconPaths.dataDir, "beacon.log")}, config);
 
   onGracefulShutdown(async () => {
     abortController.abort();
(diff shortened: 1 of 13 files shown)


g11tech (approved):
lgtm!

PR #4647, opened 2022-10-06
Enforce strict equality checking

dadepo:
**Motivation**

In a PR review, usage of non-type safe equality was spotted. Since It is a good practice to use type safe equality, this PR enforces it via linting rules and also fixes the various places in the code base non-type safe equality has been used.

### DIFF
--- a/packages/api/test/unit/beacon/oapiSpec.test.ts
+++ b/packages/api/test/unit/beacon/oapiSpec.test.ts
@@ -122,7 +122,7 @@ describe("eventstream event data", () => {
       const exampleDataJson = JSON.parse(exampleDataStr.slice(5).trim()) as unknown;
 
       const testEvent = eventTestData[topic as keyof typeof eventTestData];
-      if (testEvent == null) {
+      if (testEvent === null) {
         throw Error(`No eventTestData for ${topic}`);
       }
 
--- a/packages/beacon-node/src/api/impl/beacon/state/index.ts
+++ b/packages/beacon-node/src/api/impl/beacon/state/index.ts
@@ -63,7 +63,7 @@ export function getBeaconStateApi({
       if (filters?.id) {
         for (const id of filters.id) {
           const validatorIndex = getStateValidatorIndex(id, state, pubkey2index);
-          if (validatorIndex != null) {
+          if (validatorIndex !== undefined) {
             const validator = validators.getReadonly(validatorIndex);
             if (filters.status && !filters.status.includes(getValidatorStatus(validator, currentEpoch))) {
               continue;
@@ -108,7 +108,7 @@ export function getBeaconStateApi({
       const {pubkey2index} = chain.getHeadState().epochCtx;
 
       const validatorIndex = getStateValidatorIndex(validatorId, state, pubkey2index);
-      if (validatorIndex == null) {
+      if (validatorIndex === undefined) {
         throw new ApiError(404, "Validator not found");
       }
 
@@ -137,7 +137,7 @@ export function getBeaconStateApi({
             balances.push({index: id, balance: state.balances.get(id)});
           } else {
             const index = headState.epochCtx.pubkey2index.get(id);
-            if (index != null && index <= state.validators.length) {
+            if (index !== undefined && index <= state.validators.length) {
               balances.push({index, balance: state.balances.get(index)});
             }
           }
--- a/packages/beacon-node/src/chain/balancesCache.ts
+++ b/packages/beacon-node/src/chain/balancesCache.ts
@@ -35,7 +35,7 @@ export class CheckpointBalancesCache {
     const epochBoundaryRoot =
       epochBoundarySlot === state.slot ? blockRootHex : toHexString(getBlockRootAtSlot(state, epochBoundarySlot));
 
-    const index = this.items.findIndex((item) => item.epoch === epoch && item.rootHex == epochBoundaryRoot);
+    const index = this.items.findIndex((item) => item.epoch === epoch && item.rootHex === epochBoundaryRoot);
     if (index === -1) {
       if (this.items.length === MAX_BALANCE_CACHE_SIZE) {
         this.items.shift();
--- a/packages/beacon-node/src/chain/initState.ts
+++ b/packages/beacon-node/src/chain/initState.ts
@@ -90,7 +90,7 @@ export async function initStateFromEth1({
     logger,
     signal,
     pendingStatus:
-      statePreGenesis && depositTree !== undefined && lastProcessedBlockNumber != null
+      statePreGenesis && depositTree !== undefined && lastProcessedBlockNumber !== null
         ? {state: statePreGenesis, depositTree, lastProcessedBlockNumber}
         : undefined,
   });
@@ -118,7 +118,7 @@ export async function initStateFromEth1({
 
     return genesisResult.state;
   } catch (e) {
-    if (builder.lastProcessedBlockNumber != null) {
+    if (builder.lastProcessedBlockNumber !== null) {
       logger.info("Persisting genesis state", {block: builder.lastProcessedBlockNumber});
 
       // Commit changed before serializing
--- a/packages/beacon-node/src/chain/validation/block.ts
+++ b/packages/beacon-node/src/chain/validation/block.ts
@@ -70,7 +70,7 @@ export async function validateGossipBlock(
   }
 
   // [REJECT] The current finalized_checkpoint is an ancestor of block -- i.e.
-  // get_ancestor(store, block.parent_root, compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)) == store.finalized_checkpoint.root
+  // get_ancestor(store, block.parent_root, compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)) === store.finalized_checkpoint.root
   const parentRoot = toHexString(block.parentRoot);
   const parentBlock = chain.forkChoice.getBlockHex(parentRoot);
   if (parentBlock === null) {
@@ -94,7 +94,7 @@ export async function validateGossipBlock(
   // single bad block can trigger
   // Note: Ensure this check is done before calling chain.regen.getBlockSlotStat as this is the function that does various epoch transitions.
   // Note: This validation check is not part of the spec.
-  if (chain.opts.maxSkipSlots != null && parentBlock.slot + chain.opts.maxSkipSlots < blockSlot) {
+  if (chain.opts.maxSkipSlots !== undefined && parentBlock.slot + chain.opts.maxSkipSlots < blockSlot) {
     throw new BlockGossipError(GossipAction.IGNORE, {
       code: BlockErrorCode.TOO_MANY_SKIPPED_SLOTS,
       parentSlot: parentBlock.slot,
@@ -124,7 +124,7 @@ export async function validateGossipBlock(
 
   // Extra conditions for merge fork blocks
   // [REJECT] The block's execution payload timestamp is correct with respect to the slot
-  // -- i.e. execution_payload.timestamp == compute_timestamp_at_slot(state, block.slot).
+  // -- i.e. execution_payload.timestamp === compute_timestamp_at_slot(state, block.slot).
   if (fork === ForkName.bellatrix) {
     if (!isExecutionBlockBodyType(block.body)) throw Error("Not merge block type");
     const executionPayload = block.body.executionPayload;
--- a/packages/beacon-node/src/chain/validation/block.ts
+++ b/packages/beacon-node/src/chain/validation/block.ts
@@ -70,7 +70,7 @@ export async function validateGossipBlock(
   }
 
   // [REJECT] The current finalized_checkpoint is an ancestor of block -- i.e.
-  // get_ancestor(store, block.parent_root, compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)) === store.finalized_checkpoint.root
+  // get_ancestor(store, block.parent_root, compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)) == store.finalized_checkpoint.root
   const parentRoot = toHexString(block.parentRoot);
   const parentBlock = chain.forkChoice.getBlockHex(parentRoot);
   if (parentBlock === null) {
@@ -124,7 +124,7 @@ export async function validateGossipBlock(
 
   // Extra conditions for merge fork blocks
   // [REJECT] The block's execution payload timestamp is correct with respect to the slot
-  // -- i.e. execution_payload.timestamp === compute_timestamp_at_slot(state, block.slot).
+  // -- i.e. execution_payload.timestamp == compute_timestamp_at_slot(state, block.slot).
   if (fork === ForkName.bellatrix) {
     if (!isExecutionBlockBodyType(block.body)) throw Error("Not merge block type");
     const executionPayload = block.body.executionPayload;
--- a/packages/beacon-node/src/network/peers/utils/prioritizePeers.ts
+++ b/packages/beacon-node/src/network/peers/utils/prioritizePeers.ts
@@ -311,7 +311,7 @@ function pruneExcessPeers({
       // If we have successfully found a candidate peer to prune, prune it,
       // otherwise all peers on this subnet should not be removed.
       // In this case, we remove all peers from the pruning logic and try another subnet.
-      if (removedPeer != null) {
+      if (removedPeer !== null) {
         // recalculate variables
         removePeerFromSubnetToPeers(subnetToPeers, removedPeer);
         decreaseSynccommitteePeerCount(syncCommitteePeerCount, removedPeer.syncnetsTrueBitIndices);
--- a/packages/beacon-node/src/sync/backfill/backfill.ts
+++ b/packages/beacon-node/src/sync/backfill/backfill.ts
@@ -506,7 +506,6 @@ export class BackfillSync extends (EventEmitter as {new (): BackfillSyncEmitter}
     if (this.syncAnchor.lastBackSyncedBlock === null) {
       throw Error("Invalid lastBackSyncedBlock for checkpoint validation");
     }
-    let x: string = this.wsCheckpointHeader;
     if (this.wsCheckpointHeader === null) {
       throw Error("Invalid null checkpoint for validation");
     }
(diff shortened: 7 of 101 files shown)


packages/beacon-node/src/chain/initState.ts
```diff
@@ -118,7 +118,7 @@ export async function initStateFromEth1({
 
     return genesisResult.state;
   } catch (e) {
-    if (builder.lastProcessedBlockNumber != null) {
+    if (builder.lastProcessedBlockNumber !== null) {
```
wemeetagain, line 121: Since this is has a slightly different meaning, I'd go through each one carefully to make sure no bug is introduced.
Double check each case in this PR that looked like `foo != null` or `foo == null` to make sure that the proper behavior is preserved. If `foo` can be either `null` or `undefined` then we need to check each case (or keep `== null`). If `foo` is either `null` OR `undefined` then it should be good to change and tighten the condition.

dadepo: I agree each diff should be carefully inspected. Also will it be possible to run this branch on our tests servers for a while before merging?

dapplion (changes requested):
Doing `var == null` or `var != null` is very nice syntax sugar to test null and undefined in succinct syntax. This PR has a very high risk of introducing nasty bugs with no clear benefit.

Even if you test this branch on servers, many of this code paths won't be hit even if you run Lodestar for a while. The only sensible solution is to replace all statements `var == null` with `var === null || var === undefined` but I don't really see the value in that.

With this PR you have spotted some missuses of `==` I would just correct those. I would be okay merging a strict equality rule that applies to everything except null and undefined, but this is very low priority ATM

packages/beacon-node/src/api/impl/beacon/state/index.ts
```diff
@@ -63,7 +63,7 @@ export function getBeaconStateApi({
       if (filters?.id) {
         for (const id of filters.id) {
           const validatorIndex = getStateValidatorIndex(id, state, pubkey2index);
-          if (validatorIndex != null) {
+          if (validatorIndex !== undefined) {
```
dadepo, line 66: `getStateValidatorIndex` returns `number | undefined`. This ensures that `null` can never be returned. If `null` is attempted, this will fail at compile time. 

The change catches a bug, because if `getStateValidatorIndex` returns `undefined`, then `validatorIndex != null` will return true. Which is not what is intended. Although it is worth nothing with the "smart" option set for the eslint rules, `validatorIndex != null` will take care of `validatorIndex != undefined`.

packages/beacon-node/src/api/impl/beacon/state/index.ts
```diff
@@ -108,7 +108,7 @@ export function getBeaconStateApi({
       const {pubkey2index} = chain.getHeadState().epochCtx;
 
       const validatorIndex = getStateValidatorIndex(validatorId, state, pubkey2index);
-      if (validatorIndex == null) {
+      if (validatorIndex === undefined) {
```
dadepo, line 111: same observation as [here](https://github.com/ChainSafe/lodestar/pull/4647/files#r992314583)

packages/beacon-node/src/api/impl/beacon/state/index.ts
```diff
@@ -137,7 +137,7 @@ export function getBeaconStateApi({
             balances.push({index: id, balance: state.balances.get(id)});
           } else {
             const index = headState.epochCtx.pubkey2index.get(id);
-            if (index != null && index <= state.validators.length) {
+            if (index !== undefined && index <= state.validators.length) {
```
dadepo, line 140: `headState.epochCtx.pubkey2index.get(id)` return type is `number | undefined` hence the compiler ensures `null` can never be returned. 

The code before the change works because when `index` is `undefined`, the `index != null` will return true (which is not what is expected), but `undefined <=  state.validators.length` will always return false.

The better way is to have strict checking against `undefined` which the diff introduces

packages/beacon-node/src/chain/validation/block.ts
```diff
@@ -94,7 +94,7 @@ export async function validateGossipBlock(
   // single bad block can trigger
   // Note: Ensure this check is done before calling chain.regen.getBlockSlotStat as this is the function that does various epoch transitions.
   // Note: This validation check is not part of the spec.
-  if (chain.opts.maxSkipSlots != null && parentBlock.slot + chain.opts.maxSkipSlots < blockSlot) {
+  if (chain.opts.maxSkipSlots !== undefined && parentBlock.slot + chain.opts.maxSkipSlots < blockSlot) {
```
dadepo, line 97: type of `chain.opts.maxSkipSlots` is `number | undefined`, so the compiler takes care that `null` can never be a value.

The code before the diff when `chain.opts.maxSkipSlots` is `undefined` works because even though `chain.opts.maxSkipSlots != null` returns true (ie undefined != null returns true),  `parentBlock.slot + chain.opts.maxSkipSlots < blockSlot` will always return false because `parentBlock.slot + undefined < blockSlot` will be false.

The better way is to perform strict check against `undefined`, this ensures the if check short-circuits on first check.

packages/api/test/unit/beacon/oapiSpec.test.ts
```diff
@@ -122,7 +122,7 @@ describe("eventstream event data", () => {
       const exampleDataJson = JSON.parse(exampleDataStr.slice(5).trim()) as unknown;
 
       const testEvent = eventTestData[topic as keyof typeof eventTestData];
-      if (testEvent == null) {
+      if (testEvent === null) {
```
dadepo, line 125: the `topic as keyof typeof eventTestData` in `eventTestData[topic as keyof typeof eventTestData]` ensures that `testEvent` will have a non-null/non-undefined value...if not, this fails at compile time. Hence the `testEvent === null)` check might even be considered redundant.

packages/beacon-node/src/chain/balancesCache.ts
```diff
@@ -35,7 +35,7 @@ export class CheckpointBalancesCache {
     const epochBoundaryRoot =
       epochBoundarySlot === state.slot ? blockRootHex : toHexString(getBlockRootAtSlot(state, epochBoundarySlot));
 
-    const index = this.items.findIndex((item) => item.epoch === epoch && item.rootHex == epochBoundaryRoot);
+    const index = this.items.findIndex((item) => item.epoch === epoch && item.rootHex === epochBoundaryRoot);
```
dadepo, line 38: both `item.rootHex` and `epochBoundaryRoot` are typed `string`. This diff is safe

packages/validator/src/slashingProtection/minMaxSurround/minMaxSurround.ts
```diff
@@ -47,7 +47,7 @@ export class MinMaxSurround implements IMinMaxSurround {
   private async assertNotSurrounding(pubKey: BLSPubkey, attestation: MinMaxSurroundAttestation): Promise<void> {
     const minSpan = await this.store.minSpan.get(pubKey, attestation.sourceEpoch);
     const distance = attestation.targetEpoch - attestation.sourceEpoch;
-    if (minSpan != null && minSpan > 0 && minSpan < distance) {
+    if (minSpan !== null && minSpan > 0 && minSpan < distance) {
```
dadepo, line 50: type of miinSpan is `number | null`. It can't have the value `undefined`, making this diff safe

packages/validator/src/slashingProtection/attestation/index.ts
```diff
@@ -147,7 +147,7 @@ export class SlashingProtectionAttestationService {
     // Pre-compute and store lower-bound
     const minSourceEpoch = minEpoch(attestations.map((attestation) => attestation.sourceEpoch));
     const minTargetEpoch = minEpoch(attestations.map((attestation) => attestation.targetEpoch));
-    if (minSourceEpoch != null && minTargetEpoch != null) {
+    if (minSourceEpoch !== null && minTargetEpoch !== null) {
```
dadepo, line 150: type of minSourceEpoch is number | null. It can't have the value undefined, making this diff safe

packages/utils/src/objects.ts
```diff
@@ -30,7 +30,7 @@ export function toExpectedCase(
 }
 
 function isObjectObject(val: unknown): boolean {
-  return val != null && typeof val === "object" && Array.isArray(val) === false;
+  return val !== null && typeof val === "object" && Array.isArray(val) === false;
```
dadepo, line 33: diff is safe. `isObjectObject` will continue to return false if `val` is `null` or `undefined`

packages/beacon-node/src/chain/initState.ts
dadepo, line 121: type of `builder.lastProcessedBlockNumber` is `number | null`. It is safe to do a strict check against `null` as the compiler ensures `builder.lastProcessedBlockNumber` won't be undefined

packages/beacon-node/src/chain/initState.ts
```diff
@@ -90,7 +90,7 @@ export async function initStateFromEth1({
     logger,
     signal,
     pendingStatus:
-      statePreGenesis && depositTree !== undefined && lastProcessedBlockNumber != null
+      statePreGenesis && depositTree !== undefined && lastProcessedBlockNumber !== null
```
dadepo, line 93: `depositTree` is ` ListCompositeTreeViewDU<ByteVectorType>`. Compiler ensures it can't be null or undefined so this check is actually not necessary.

type of `lastProcessedBlockNumber` is `number | null`. Won\t be `undefined` so strict checking against `null` is safe

packages/beacon-node/src/network/peers/utils/prioritizePeers.ts
```diff
@@ -311,7 +311,7 @@ function pruneExcessPeers({
       // If we have successfully found a candidate peer to prune, prune it,
       // otherwise all peers on this subnet should not be removed.
       // In this case, we remove all peers from the pruning logic and try another subnet.
-      if (removedPeer != null) {
+      if (removedPeer !== null) {
```
dadepo, line 314: type of `removedPeer` is `PeerInfo | null`. It won't be `undefined` hence safe to perform strict checking against only`null`

packages/beacon-node/src/sync/backfill/backfill.ts
```diff
@@ -318,7 +318,7 @@ export class BackfillSync extends (EventEmitter as {new (): BackfillSyncEmitter}
       //       parent of
       //       the anchorBlock via strategy a) as it could be multiple skipped/orphaned slots
       //       behind
-      if (this.syncAnchor.lastBackSyncedBlock != null) {
+      if (this.syncAnchor.lastBackSyncedBlock !== null) {
```
dadepo, line 321: type of `this.syncAnchor.lastBackSyncedBlock` is `BackfillBlock | null`. Safe to use strict equality check

packages/cli/src/config/peerId.ts
```diff
@@ -10,11 +10,11 @@ import {writeFile600Perm, readFile} from "../util/index.js";
 // See https://github.com/libp2p/js-libp2p-peer-id/pull/9 for more details
 
 async function createFromParts(multihash: Uint8Array, privKey?: Uint8Array, pubKey?: Uint8Array): Promise<PeerId> {
-  if (privKey != null) {
+  if (privKey !== undefined) {
```
dadepo, line 13: `privKey` is of type `Uint8Array | undefined`

packages/cli/src/config/peerId.ts
```diff
@@ -28,9 +28,9 @@ export type PeerIdJSON = {id: string; pubKey?: string; privKey?: string};
 export function exportToJSON(peerId: PeerId, excludePrivateKey?: boolean): PeerIdJSON {
   return {
     id: uint8ArrayToString(peerId.toBytes(), "base58btc"),
-    pubKey: peerId.publicKey != null ? uint8ArrayToString(peerId.publicKey, "base64pad") : undefined,
+    pubKey: peerId.publicKey !== undefined ? uint8ArrayToString(peerId.publicKey, "base64pad") : undefined,
```
dadepo, line 31: `peerId.publicKey` is `Uint8Array | undefined`

packages/cli/src/config/peerId.ts
```diff
@@ -28,9 +28,9 @@ export type PeerIdJSON = {id: string; pubKey?: string; privKey?: string};
...
     privKey:
-      excludePrivateKey === true || peerId.privateKey == null
+      excludePrivateKey === true || peerId.privateKey === undefined
```
dadepo, line 33: peerId.privateKey is `Uint8Array | undefined`

packages/cli/src/config/peerId.ts
```diff
@@ -39,8 +39,8 @@ export function exportToJSON(peerId: PeerId, excludePrivateKey?: boolean): PeerI
 export async function createFromJSON(obj: PeerIdJSON): Promise<PeerId> {
   return await createFromParts(
     uint8ArrayFromString(obj.id, "base58btc"),
-    obj.privKey != null ? uint8ArrayFromString(obj.privKey, "base64pad") : undefined,
-    obj.pubKey != null ? uint8ArrayFromString(obj.pubKey, "base64pad") : undefined
+    obj.privKey !== undefined ? uint8ArrayFromString(obj.privKey, "base64pad") : undefined,
```
dadepo, line 42: `obj.privKey && obj.pubKey` typed `string | undefined`

packages/cli/src/options/paramsOptions.ts
```diff
@@ -46,7 +48,9 @@ export function parseTerminalPowArgs(args: ITerminalPowArgs): IBeaconParamsUnpar
   const parsedArgs = ObjectKeys(terminalArgsToParamsMap).reduce((beaconParams: Partial<IBeaconParamsUnparsed>, key) => {
     const paramOption = terminalArgsToParamsMap[key];
     const value = args[key];
-    if (paramOption != null && value != null) beaconParams[paramOption] = value;
+    if (paramOption !== undefined && value !== undefined) {
```
dadepo, line 51: `value && paramOption` is typed `a type | undefined`

packages/beacon-node/src/sync/backfill/backfill.ts
```diff
@@ -503,10 +503,10 @@ export class BackfillSync extends (EventEmitter as {new (): BackfillSyncEmitter}
    * the initialization point is the same block tree as the DB once backfill
    */
   private async checkIfCheckpointSyncedAndValidate(): Promise<void> {
-    if (this.syncAnchor.lastBackSyncedBlock == null) {
+    if (this.syncAnchor.lastBackSyncedBlock === null) {
```
dadepo, line 506: type `this.syncAnchor.lastBackSyncedBlock` is `BackfillBlock | null`. Compiler ensure this won't be null so safe to do strict checking against `null`

packages/beacon-node/src/sync/backfill/backfill.ts
```diff
@@ -503,10 +503,10 @@ export class BackfillSync extends (EventEmitter as {new (): BackfillSyncEmitter}
...
       throw Error("Invalid lastBackSyncedBlock for checkpoint validation");
     }
-    if (this.wsCheckpointHeader == null) {
+    if (this.wsCheckpointHeader === null) {
```
dadepo, line 509: type of `this.wsCheckpointHeader` is `BackfillBlockHeader | null`. Same observation [here](https://github.com/ChainSafe/lodestar/pull/4647/files#r996684857) applies

packages/beacon-node/src/sync/backfill/backfill.ts
```diff
@@ -581,7 +581,7 @@ export class BackfillSync extends (EventEmitter as {new (): BackfillSyncEmitter}
             // we only need to check if prevFinalizedCheckpointBlock is in db
             const prevBackfillCpBlock = await this.db.blockArchive.getByRoot(this.prevFinalizedCheckpointBlock.root);
             if (
-              prevBackfillCpBlock != null &&
+              prevBackfillCpBlock !== null &&
```
dadepo, line 584: type of `prevBackfillCpBlock` is ` allForks.SignedBeaconBlock | null`. Same observation [here](https://github.com/ChainSafe/lodestar/pull/4647/files#r996684857) applies

packages/beacon-node/src/sync/backfill/backfill.ts
```diff
@@ -744,7 +744,7 @@ export class BackfillSync extends (EventEmitter as {new (): BackfillSyncEmitter}
 
   private async syncBlockByRoot(peer: PeerId, anchorBlockRoot: Root): Promise<void> {
     const [anchorBlock] = await this.network.reqResp.beaconBlocksByRoot(peer, [anchorBlockRoot]);
-    if (anchorBlock == null) throw new Error("InvalidBlockSyncedFromPeer");
+    if (anchorBlock === null) throw new Error("InvalidBlockSyncedFromPeer");
```
dadepo, line 747: type of `anchorBlock` is `allForks.SignedBeaconBlock[]` the compiler enforces this is not null, so this check is actually frivolous.

packages/beacon-node/src/sync/backfill/backfill.ts
```diff
@@ -866,7 +866,7 @@ async function extractPreviousFinOrWsCheckpoint(
   )[0];
 
   let prevFinalizedCheckpointBlock: BackfillBlockHeader;
-  if (nextPrevFinOrWsBlock != null) {
+  if (nextPrevFinOrWsBlock !== undefined) {
```
dadepo, line 869: type of `nextPrevFinOrWsBlock` is ` allForks.SignedBeaconBlock` compiler ensures can't be null. The actual value is access via array index. If at runtime, nothing is found at that index, `undefined` will be set not `null`. So safe to do strict check against `undefined`

packages/cli/src/options/paramsOptions.ts
```diff
@@ -19,7 +19,9 @@ const getArgKey = (key: keyof IBeaconParamsUnparsed): string => `params.${key}`;
 export function parseBeaconParamsArgs(args: Record<string, string | number>): IBeaconParamsUnparsed {
   return ObjectKeys(chainConfigTypes).reduce((beaconParams: Partial<IBeaconParamsUnparsed>, key) => {
     const value = args[getArgKey(key)];
-    if (value != null) beaconParams[key] = value;
+    if (value !== undefined) {
```
dadepo, line 22: array index can return `undefined` at runtime not null.

dadepo:
Closing for now, as not priority and no consensus on the improvement the lint rule provides.

PR #4648, opened 2022-10-06
Add vc bn fallback

### LINKED ISSUE #4337: Support multiple bn per vc (fallback)
**Describe the bug**

This option should support an array of endpoints

https://github.com/ChainSafe/lodestar/blob/ef0091e69b601243d4cf60561e32676def205f97/packages/cli/src/cmds/validator/options.ts#L130

dapplion:
**Motivation**

- See https://github.com/ChainSafe/lodestar/issues/4337

**Description**

A good vc bn fallback must fulfill these goals
- if first server is stable and responding do not query fallbacks
- if first server errors, retry that same request on fallbacks
- until first server is shown to be reliable again, contact all servers

This behaviour is not simple, so classic patterns like Promise.all() or for try catch are not enough.

The solution of this PR appears complex, but that's what's needed to achieve the goals above. Please check the inline comments for pointers on the logic.

A unit tests proves that it works, but I haven't tested yet on a deployed beacon node.

Closes https://github.com/ChainSafe/lodestar/issues/4337

### DIFF
--- a/packages/api/src/utils/client/httpClient.ts
+++ b/packages/api/src/utils/client/httpClient.ts
@@ -6,6 +6,7 @@ import {Metrics} from "./metrics.js";
 
 /** A higher default timeout, validator will sets its own shorter timeoutMs */
 const DEFAULT_TIMEOUT_MS = 60_000;
+const DEFAULT_ROUTE_ID = "unknown";
 const URL_SCORE_MAX = 10;
 const URL_SCORE_MIN = 0;
 const URL_SCORE_DELTA_ERROR = 1;
@@ -70,11 +71,11 @@ export class HttpClient implements IHttpClient {
   private readonly metrics: null | Metrics;
   private readonly logger: null | ILogger;
 
-  private readonly urls: URLOpts[] = [];
+  private readonly urlOpts: URLOpts[] = [];
   private readonly urlScore: number[];
 
   get baseUrl(): string {
-    return this.urls[0].baseUrl;
+    return this.urlOpts[0].baseUrl;
   }
 
   /**
@@ -82,7 +83,7 @@ export class HttpClient implements IHttpClient {
    */
   constructor(opts: HttpClientOptions, {logger, metrics}: HttpClientModules = {}) {
     if (opts.baseUrl) {
-      this.urls.push({
+      this.urlOpts.push({
         baseUrl: opts.baseUrl,
         bearerToken: opts.bearerToken,
         timeoutMs: opts.timeoutMs,
@@ -91,7 +92,7 @@ export class HttpClient implements IHttpClient {
 
     if (opts.urls) {
       for (const urlOpts of opts.urls) {
-        this.urls.push(urlOpts);
+        this.urlOpts.push(urlOpts);
       }
     }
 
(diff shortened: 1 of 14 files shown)


wemeetagain (dismissed):
You might be able to use `Promise.any`, maybe not worth it tho :P
looks good

packages/validator/src/validator.ts
```diff
@@ -71,18 +71,25 @@ export class Validator {
...
+    let api: Api;
+    if (typeof opts.api === "string" || Array.isArray(opts.api)) {
+      const urls = typeof opts.api === "string" ? [opts.api] : opts.api;
+      const urlOpts = urls.map((url) => ({baseUrl: url}));
+      // This new api instance can make do with default timeout as a faster timeout is
+      // not necessary since this instance won't be used for validator duties
+      api = getClient(
+        {
+          baseUrl: urls[0],
+          urls: urlOpts,
```
twoeths: do we have any issues if `urls` includes `baseUrl`?

packages/api/src/utils/client/httpClient.ts
```diff
@@ -49,62 +68,167 @@ export type HttpClientModules = {
...
-    this.bearerToken = opts.bearerToken;
+    if (opts.baseUrl) {
+      this.urlOpts.push({
+        baseUrl: opts.baseUrl,
+        bearerToken: opts.bearerToken,
+        timeoutMs: opts.timeoutMs,
+      });
+    }
+
+    if (opts.urls) {
```
twoeths: should we do some filters to avoid duplicate in `this.urlOpts`?

dapplion:
> You might be able to use Promise.any, maybe not worth it tho :P

Looked into it and doesn't simplify the code that much since I have to track which URL completed the request successfully. Also Promise.any is available in NodeJS >= 15.0.0 and our engine is still > 12.9

packages/api/src/utils/client/httpClient.ts
dapplion: Yes good point, added

philknows:
Is there any sort of notification for us we can use when a fallback is triggered?

dapplion:
> Is there any sort of notification for us we can use when a fallback is triggered?

Yes, this PRs adds metrics that will clearly show if a URL is down

packages/api/src/utils/client/httpClient.ts
```diff
@@ -49,62 +65,183 @@ export type HttpClientModules = {
...
+            const i_ = i; // Keep local copy of i variable to index urlScore after requestWithBody() resolves
+
+            this.requestWithBody(urlOpts, opts, getBody).then(
+              (res) => {
+                this.urlsScore[i_] = Math.min(URL_SCORE_MAX, this.urlsScore[i_] + URL_SCORE_DELTA_SUCCESS);
+                // Resolve immediately on success
+                resolve(res);
+              },
+              (err) => {
+                this.urlsScore[i_] = Math.max(URL_SCORE_MIN, this.urlsScore[i_] - URL_SCORE_DELTA_ERROR);
```
g11tech, line 183: should the score be reduced by half, so that the url will have to prove `working` multiple times to come back to max

wemeetagain: Looks like currently score decreases twice as fast as score increases

wemeetagain (approved):

PR #4651, opened 2022-10-07
Serialize calls to the engine api

g11tech:
Serialize calls to the engine api

Follow up of https://github.com/ChainSafe/lodestar/pull/4595

Metrics for engine job processor:

image

### DIFF
--- a/packages/beacon-node/src/execution/engine/http.ts
+++ b/packages/beacon-node/src/execution/engine/http.ts
@@ -15,6 +15,7 @@ import {
 import {IJsonRpcHttpClient, ReqOpts} from "../../eth1/provider/jsonRpcHttpClient.js";
 import {IMetrics} from "../../metrics/index.js";
 import {JobItemQueue} from "../../util/queue/index.js";
+import {EPOCHS_PER_BATCH} from "../../sync/constants.js";
 import {
   ExecutePayloadStatus,
   ExecutePayloadResponse,
@@ -58,7 +59,11 @@ export const defaultExecutionEngineHttpOpts: ExecutionEngineHttpOpts = {
   timeout: 12000,
 };
 
-const QUEUE_MAX_LENGHT = 256;
+/**
+ * Size for the serializing queue for fcUs and new payloads, the max length could be equal to
+ * EPOCHS_PER_BATCH * 2 in case new payloads are also not awaited serially
+ */
+const QUEUE_MAX_LENGHT = EPOCHS_PER_BATCH * 2;
 
 // Define static options once to prevent extra allocations
 const notifyNewPayloadOpts: ReqOpts = {routeId: "notifyNewPayload"};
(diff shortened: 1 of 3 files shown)


packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -328,6 +328,27 @@ export function createLodestarMetrics(
...
+      }),
+      jobTime: register.histogram({
+        name: "lodestar_engine_http_processor_queue_job_time_seconds",
+        help: "Time to process engine http processor queue job in seconds",
+        buckets: [0.1, 1, 10, 100],
+      }),
+      jobWaitTime: register.histogram({
+        name: "lodestar_engine_http_processor_queue_job_wait_time_seconds",
+        help: "Time from job added to the engine http processor queue to starting in seconds",
+        buckets: [0.1, 1, 10, 100],
```
dapplion: Bucket values must be realistic on expected values else not useful at all. If this sort of x10 prev value buckets are still in Lodestar is because we add them a long time ago, or we haven't figured out the right values. The wait time metric is crucial so buckets must be good. See below as example of buckets values specific to a metric given expected values

https://github.com/ChainSafe/lodestar/blob/1cba8368ac433752d7f60df30fd0b551013226e3/packages/beacon-node/src/metrics/metrics/lodestar.ts#L369-L370

packages/beacon-node/src/execution/engine/http.ts
```diff
@@ -266,16 +277,11 @@ export class ExecutionEngineHttp implements IExecutionEngine {
...
-        method,
-        params: [payloadId],
-      },
-      getPayloadOpts
-    );
+    const executionPayloadRpc = await (this.jobQueue.push({
+      method,
+      params: [payloadId],
+      methodOpts: getPayloadOpts,
+    }) as Promise<EngineApiRpcReturnTypes[typeof method]>);
```
dapplion: getPayload doesn't need to be sequential right? Don't feel it needs to be queued

packages/beacon-node/src/execution/engine/http.ts
```diff
@@ -75,6 +78,14 @@ const exchageTransitionConfigOpts: ReqOpts = {routeId: "exchangeTransitionConfig
 export class ExecutionEngineHttp implements IExecutionEngine {
   readonly payloadIdCache = new PayloadIdCache();
   private readonly rpc: IJsonRpcHttpClient;
+  private readonly jobQueue: JobItemQueue<[EngineRequest], EngineResponse>;
```
dapplion: Since it's not generic to multiple jobs I would call it `rpcFetchQueue` or something like that. Also add a comment explaining why this is necessary and reason which methods need it and which do not

packages/beacon-node/src/execution/engine/http.ts
```diff
@@ -57,6 +58,8 @@ export const defaultExecutionEngineHttpOpts: ExecutionEngineHttpOpts = {
   timeout: 12000,
 };
 
+const QUEUE_MAX_LENGHT = 256;
```
dapplion: Write a JSDoc comment reasoning the choice of this number. What the tradeoff here? A job queue holds items in memory which in this case are full de-serialized blocks. Are there any memory concerns if the queue fills? What's a reasonable length given the usage of this queue? Since blocks are imported strictly sequential the queue length should always be pretty short.

packages/beacon-node/src/metrics/metrics/lodestar.ts
g11tech: hmmm right, most of the time will be taken up by new payload, and it can vary from 100s of ms to 2-4 seconds (if EL is really slow) and may be on average 300-400ms, so should be bucket for [0.1, 0.5, 2, 5] ?

packages/beacon-node/src/execution/engine/http.ts
g11tech: right, can keep it out of the queue, but ideally there shouldn't be anything else in the queue at that moment anyways. still can keep it out.

packages/beacon-node/src/execution/engine/http.ts
g11tech: Will check from the blocks processor queue what limit it has, ideally there could be fcUs corresponding to the last batch of blocks plus new payload

packages/beacon-node/src/metrics/metrics/lodestar.ts
dapplion: Then do for example
```
[0.05, 0.1, 0.2, 0.3, 0.5, 0.75, 1, 2, 5, 10, 25]
```

packages/beacon-node/src/execution/engine/http.ts
```diff
@@ -57,6 +59,12 @@ export const defaultExecutionEngineHttpOpts: ExecutionEngineHttpOpts = {
   timeout: 12000,
 };
 
+/**
+ * Size for the serializing queue for fcUs and new payloads, the max length could be equal to
+ * EPOCHS_PER_BATCH * 2 in case new payloads are also not awaited serially
+ */
+const QUEUE_MAX_LENGHT = EPOCHS_PER_BATCH * SLOTS_PER_EPOCH * 2;
```
wemeetagain: lenght -> length

g11tech: :facepalm:

wemeetagain (approved):
Lgtm, appreciate the metrics

PR #4655, opened 2022-10-10
Enable specifying bootnodes cli arg as comma separated values

g11tech:
**Motivation**
Pari's ansible uses comma separated bootnodes and he endeavour to use them in a custom devnet failed as we don't accept bootnodes as comma separated,

This PR enables providing the bootnodes as comma separated arg value

### DIFF
--- a/packages/cli/src/options/beaconNodeOptions/network.ts
+++ b/packages/cli/src/options/beaconNodeOptions/network.ts
@@ -27,17 +27,13 @@ export function parseArgs(args: INetworkArgs): IBeaconNodeOptions["network"] {
   const listenAddress = args.listenAddress || defaultListenAddress;
   const udpPort = args.discoveryPort ?? args.port ?? defaultP2pPort;
   const tcpPort = args.port ?? defaultP2pPort;
-  // Each bootnode entry could be comma separated, just deserialize it into a single array
-  // as comma separated entries are generally most friendly in ansible kind of setups, i.e.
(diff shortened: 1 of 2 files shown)


packages/cli/src/options/beaconNodeOptions/network.ts
```diff
@@ -27,13 +27,17 @@ export function parseArgs(args: INetworkArgs): IBeaconNodeOptions["network"] {
   const listenAddress = args.listenAddress || defaultListenAddress;
   const udpPort = args.discoveryPort ?? args.port ?? defaultP2pPort;
   const tcpPort = args.port ?? defaultP2pPort;
+  // Each bootnode entry could be comma separated, just deserialize it into a single array
+  // as comma separated entries are generally most friendly in ansible kind of setups, i.e.
+  // [ "en1", "en2,en3" ] => [ 'en1', 'en2', 'en3' ]
+  const bootnodes = (args["bootnodes"] ?? []).reduce((acc: string[], elem: string) => acc.concat(elem.split(",")), []);
```
dapplion: This logic should be in the coerce function of yargs here 

https://github.com/ChainSafe/lodestar/blob/b1ae325ddc0ef483535e6c3f110a2b48685c0ad4/packages/cli/src/options/beaconNodeOptions/network.ts#L86-L91

check some validators args for this exact usecase

wemeetagain (approved):

PR #4658, opened 2022-10-13
Set enr.tcp to default to port/defaultP2pPort

g11tech:
Earlier we used to default to set the port/defaultP2pPort for enr.tcp.

 But the latest code doesn't seem to be setting enr.tcp which is missing in the generated ENR. This PR re-adds the default back and also simpifies enr.udp assignment.

### DIFF
--- a/packages/cli/src/cmds/beacon/handler.ts
+++ b/packages/cli/src/cmds/beacon/handler.ts
@@ -14,6 +14,7 @@ import {BeaconNodeOptions, exportToJSON, FileENR, getBeaconConfigFromArgs} from 
 import {onGracefulShutdown, getCliLogger, mkdir, writeFile600Perm} from "../../util/index.js";
 import {getNetworkBootnodes, getNetworkData, readBootnodes} from "../../networks/index.js";
 import {getVersionData} from "../../util/version.js";
+import {defaultP2pPort} from "../../options/beaconNodeOptions/network.js";
 import {IBeaconArgs} from "./options.js";
 import {getBeaconPaths} from "./paths.js";
 import {initBeaconState} from "./initBeaconState.js";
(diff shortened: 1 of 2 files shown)


packages/cli/src/cmds/beacon/handler.ts
```diff
@@ -150,16 +151,15 @@ export async function beaconHandlerInit(args: IBeaconArgs & IGlobalArgs) {
...
+  enr.tcp = args["enr.tcp"] ?? args.port ?? defaultP2pPort;
+  // TODO: add `port`/`defaultP2pPort` port as backup as well once we
+  // fix the below discv5 issue
+  //
   // See https://github.com/ChainSafe/discv5/issues/201
-  // if (args.port != null) enr.udp = args.port;
-  if (args.discoveryPort != null) enr.udp = args.discoveryPort;
-
+  const udpPort = args["enr.udp"] ?? args.discoveryPort;
+  if (udpPort != null) enr.udp = udpPort;
```
twoeths, line 161: we should not set udp to enr until the above discv5 issue is resolved

twoeths:
`tcp` is not missing in the generated enr, just checked the `enr` file in unstable group via https://enr-viewer.com/

Screen Shot 2022-10-14 at 08 54 18

packages/cli/src/cmds/beacon/handler.ts
g11tech: this is just previous logic rehashed, 
```
 if (args.discoveryPort != null) enr.udp = args.discoveryPort;
if (args["enr.udp"] != null) enr.udp = args["enr.udp"];
```
so it will not assign udp if  `discoveryPort` and `enr.udp` args are not externally provided

g11tech:
> `tcp` is not missing in the generated enr, just checked the `enr` file in unstable group via https://enr-viewer.com/
> 
> Screen Shot 2022-10-14 at 08 54 18

in a proper network, ENR gets updated because of p2p protocols wherein the node gets poll on what other peers see, but in a testnet where lodestar node is acting as bootnode, this will fail as no node will be able to connect to lodestar bootnode and no updates to ENR will happen.

Observed and verified from a testnet, you can also confirm by starting lodestar on an empty datadir (on any network) and then pickup the generated ENR right at the moment.

packages/cli/src/cmds/beacon/handler.ts
twoeths: thanks for the clarification @g11tech  👍

twoeths (approved):

PR #4659, opened 2022-10-14
Use signatureSlot to verify signatures at period boundary

### LINKED ISSUE #4627: Implement the spec patch that allows light client to verify signatures at period boundary

The changes to the spec in https://github.com/ethereum/consensus-specs/pull/2805 takes care of scenarios at epoch boundary where it was not clear if an attested_header was signed by the current or next sync committee. The update to the spec introduced `signatureSlot` and updated the light client validator logic.

Signature Slot is being added here https://github.com/ChainSafe/lodestar/pull/4365

This issue is for updating the light clients verification logic to take care of the epoch boundary situation.

dadepo:
**Motivation**

Implements changes made to the spec in the PR https://github.com/ethereum/consensus-specs/pull/2805/files

Closes #4627


wemeetagain:
Blocked by #4365 tointroduce the `signatureSlot` field

packages/light-client/src/index.ts
```diff
@@ -392,31 +392,34 @@ export class Lightclient {
...
     if (attestedHeader.slot > slotWithFutureTolerance(this.config, this.genesisTime, MAX_CLOCK_DISPARITY_SEC)) {
       throw Error(`header.slot ${attestedHeader.slot} is too far in the future, currentSlot: ${this.currentSlot}`);
     }
 
-    const period = computeSyncPeriodAtSlot(attestedHeader.slot);
-    const syncCommittee = this.syncCommitteeByPeriod.get(period);
+    const attestedPeriod = computeSyncPeriodAtSlot(attestedHeader.slot);
+    const signaturePeriod = computeSyncPeriodAtSlot(signatureSlot);
+    // if at sync period boundary, the signature is constructed by sync committee in the next period
+    const syncCommittee = this.getSigningSyncCommitteeAtBoundary(attestedPeriod, signaturePeriod);
```
wemeetagain: we shouldn't be doing any switching here,
should always use signaturePeriod
```suggestion
    const syncCommittee = this.syncCommitteeByPeriod.get(signaturePeriod);
```

packages/light-client/test/utils/utils.ts
```diff
@@ -111,8 +111,14 @@ export function getInteropSyncCommittee(period: SyncPeriod): SyncCommitteeKeys {
...
  */
-export function computeLightclientUpdate(config: IBeaconConfig, period: SyncPeriod): altair.LightClientUpdate {
-  const updateSlot = period * EPOCHS_PER_SYNC_COMMITTEE_PERIOD * SLOTS_PER_EPOCH + 1;
+export function computeLightclientUpdate(
+  config: IBeaconConfig,
+  period: SyncPeriod,
+  atBoundary = false
+): altair.LightClientUpdate {
+  const updateSlot = atBoundary
+    ? period * EPOCHS_PER_SYNC_COMMITTEE_PERIOD * SLOTS_PER_EPOCH - 1
```
wemeetagain: is this right? wouldn't this make the update for the previous period?

wemeetagain (changes requested):

packages/light-client/src/index.ts
```diff
@@ -571,4 +585,13 @@ export class Lightclient {
       // TODO: Metrics, updated syncCommittee
     }
   }
+
+  private getSigningSyncCommitteeAtBoundary(
```
wemeetagain: This should be removed, we always use the period of signatureSlot

packages/light-client/src/index.ts
```diff
@@ -392,31 +392,34 @@ export class Lightclient {
...
-    const syncCommittee = this.syncCommitteeByPeriod.get(period);
+    const attestedPeriod = computeSyncPeriodAtSlot(attestedHeader.slot);
+    const signaturePeriod = computeSyncPeriodAtSlot(signatureSlot);
+    // if at sync period boundary, the signature is constructed by sync committee in the next period
+    const syncCommittee = this.syncCommitteeByPeriod.get(signaturePeriod);
+
     if (!syncCommittee) {
       // TODO: Attempt to fetch committee update for period if it's before the current clock period
-      throw Error(`No syncCommittee for period ${period}`);
+      throw Error(`No syncCommittee for attested period ${attestedPeriod} and signaturePeriod ${signaturePeriod}`);
```
wemeetagain: we can remove the diff for this line and the other similar error below

wemeetagain:
```suggestion
      throw Error(`No syncCommittee for period ${signaturePeriod}`);

Comment (dadepo) [reply packages/light-client/src/index.ts]: If at period 1 and at the period boundary, If no sync committee is found, the error would come back saying
(`No syncCommittee for period 2`). Won't that throw the user off? 

Maybe `No syncCommittee for attested period ${attestedPeriod} *with* signaturePeriod ${signaturePeriod}`?

File: packages/light-client/src/index.ts [hunk 5]
```diff
@@ -529,44 +531,69 @@ export class Lightclient {
...
+    // in the scenario where only update for period is requested at the boundary,
+    // there won't have been an update to process that would have enabled next sync committee to be
+    // cached in this.syncCommitteeByPeriod, hence directly use one in update
+    if (isLastSlotInPeriod(updateSlot) && signingSyncCommittee === undefined) {
+      nextSyncCommittee = deserializeSyncCommittee(update.nextSyncCommittee);
+      signingSyncCommittee = {
+        ...newNextSyncCommitteeStats,
+        ...nextSyncCommittee,
+      };
+    }
```
wemeetagain: this is incorrect please remove this.
we cannot use data from the update to verify the authenticity of the update.

packages/light-client/src/index.ts
```diff
@@ -529,44 +531,69 @@ export class Lightclient {
...
-    };
 
     if (!existingNextSyncCommittee || isBetterUpdate(existingNextSyncCommittee, newNextSyncCommitteeStats)) {
       this.logger.info("Stored SyncCommittee", {nextPeriod, replacedPrevious: existingNextSyncCommittee != null});
-      this.emitter.emit(LightclientEvent.committee, updatePeriod);
+      this.emitter.emit(LightclientEvent.committee, attestedPeriod);
       this.syncCommitteeByPeriod.set(nextPeriod, {
         ...newNextSyncCommitteeStats,
-        ...deserializeSyncCommittee(update.nextSyncCommittee),
+        ...(nextSyncCommittee !== undefined ? nextSyncCommittee : deserializeSyncCommittee(update.nextSyncCommittee)),
```
wemeetagain: can use ?? syntax here

packages/light-client/src/index.ts
dadepo: I will. But maybe you can point out how the correct implementation should work? My reading of the logic is that, on every update, the next sync committee is cached for the next period.

But in the case where the update comes at the boundary there won't be a next sync committee cached to use for it, since no previous update was processed.

How will such edge case be handled? 

My thinking was the next sync committee can be taken directly from the update and used...but that is wrong as you pointed out

dadepo: The CI now fails a test with the following log:

```
Error sync {} Error: No syncCommittee for attested period 1 and signaturePeriod 2
```

This is because in the test case, the update for period 1 is set to happen at the boundary. This means that update would have been signed by sync committee in the next period (period 2) but that does not exist in the cache yet. 

as an example, if the bootstrap using the checkpoint root is initiated at the period boundary, the header and the current sync committee is returned. From there the period is calculated and the sync starts. But the lightclient update requested will be done at the period boundary...that means the returned update will be signed by the sync committee in the next period...I am not sure how the chain of sync committee can then be validated?

wemeetagain (approved):

wemeetagain:
superseded by #4908

Issue #4656, opened 2022-10-11
Possible typescript type improvements


wemeetagain:
I was thinking about how we could better handle consensus types changing across different forks, fork capabilities, and also blinded and unblinded variants.

We can/should use generics to help switch between both forks/capabilities and blinded/unblinded if we restructure types a bit.

Here's a proposal using an example:
```ts
export type BlindedOrFull = "full" | "blinded";

export type ExecutionFork = ForkName.bellatrix | ForkName.capella;

type forkTypes = {
  [ForkName.phase0]: {
    BeaconBlockBody: phase0.BeaconBlockBody;
  };
  [ForkName.altair]: {
    BeaconBlockBody: altair.BeaconBlockBody;
  };
  [ForkName.bellatrix]: {
    BeaconBlockBody: bellatrix.BeaconBlockBody;
    BlindedBeaconBlockBody: bellatrix.BlindedBeaconBlockBody;
  };
  [ForkName.capella]: {
    BeaconBlockBody: capella.BeaconBlockBody;
    BlindedBeaconBlockBody: capella.BlindedBeaconBlockBody;
  };
};

export type BeaconBlockBody<
  F extends ForkName = ForkName,
  B extends BlindedOrFull = "full"
> = F extends ExecutionFork
  ? B extends "full"
    ? forkTypes[F]["BeaconBlockBody"]
    : forkTypes[F]["BlindedBeaconBlockBody"]
  : B extends "full"
  ? forkTypes[F]["BeaconBlockBody"]
  : never; // <- we can exclude non-execution blinded blocks :)

// 
export function isBlindedBlockBody<F extends ExecutionFork>(
  body: BeaconBlockBody<F, BlindedOrFull>
): body is BeaconBlockBody<F, "blinded"> { // <- the fork type isn't lost if its known :)
  (body as BeaconBlockBody<F, "blinded">).executionPayloadHeader !== undefined;
  return true;
}
```

So functions could be gated on different forks or unions of fork names:

```ts
export function doSomethingWithBellatrix(body: BeaconBlockBody<ForkName.bellatrix>): void {}

export function doSomethingWithExecution(body: BeaconBlockBody<ExecutionFork>): void {
  body.executionPayload;
}
```

And functions that deal with blinded or unblinded blocks can also be kept typesafe:

```ts
export function doSomethingWithBlindedOrUnblinded(body: BeaconBlockBody<ExecutionFork, BlindedOrFull>): void {
[...]

PR #4671, opened 2022-10-19
Multifork types

wemeetagain:
**Motivation**

See https://github.com/ChainSafe/lodestar/issues/4656

**Description**

Just playing around with the ideas in #4656 

Some cool ideas here, especially the typing around these methods: https://github.com/ChainSafe/lodestar/compare/cayman/multifork-types?expand=1#diff-164a51e7deba45ff2747f0e311044818bbcbb62d902fcb344057cc3e4105e91cR33-R35

~Currently blocked by the type error on these lines: https://github.com/ChainSafe/lodestar/compare/cayman/multifork-types?expand=1#diff-ae18c534732a3db368114c1c79fdf064b41236620dc61bc6e899579da049ed5fR92-R93~

Ok, made it past the first hurdle.

Now hitting a second one. It seems `SSZTypes` isn't working as intended.
```
const x: SSZTypes<ForkName.capella | ForkName.phase0>["SignedBeaconBlock"] = null as any;

const y: SignedBeaconBlock<ForkName.phase0> = null as any;

x.toJson(y); // erroring!
```

### DIFF
--- a/packages/api/src/beacon/routes/beacon/block.ts
+++ b/packages/api/src/beacon/routes/beacon/block.ts
@@ -1,8 +1,8 @@
 import {ContainerType} from "@chainsafe/ssz";
-import {ForkName} from "@lodestar/params";
+import {ExecutionFork, ExecutionForks, ForkName} from "@lodestar/params";
 import {IChainForkConfig} from "@lodestar/config";
 import {phase0, Slot, Root, ssz, RootHex} from "@lodestar/types";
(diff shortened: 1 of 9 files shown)


dapplion:
I don't disagree with the general idea but it has not become a necessity after developing both capella and deneb forks. Would you be ok with closing for now to reduce the count of stale PRs? @wemeetagain

wemeetagain:
closing this to reduce the count of stale PRs

PR #4676, opened 2022-10-20
Return READY on getHealth pre-genesis

### LINKED ISSUE #4619: Lodestar /health endpoint returns SYNCING pre-genesis
**Describe the bug**

Pre-genesis the node should return READY

g11tech:
Return READY on getHealth pre-genesis

Closes #4619


wemeetagain (approved):

dapplion:
@g11tech Not sure if that's a proper solution. Why is the internal status syncing? The fact that the return value of the API no longer aligns with what the node has internally could be a real headache when debugging

PR #4679, opened 2022-10-21
Add state upgrade to capella and enable spec tests

g11tech:
- Add state upgrade to capella 
- enable spec tests for capella which only has `transition` tests enabled in the spec 1.2.0/rcs versions
- upgrade spec to 1.2.0 from tc
- exclude new `sync` tests as we will need to write spec runners for it, task created

Part of
  -  https://github.com/ChainSafe/lodestar/issues/4680

### DIFF
--- a/packages/state-transition/src/block/processDeposit.ts
+++ b/packages/state-transition/src/block/processDeposit.ts
@@ -88,8 +88,9 @@ export function processDeposit(fork: ForkSeq, state: CachedBeaconStateAllForks, 
       (validators as CachedBeaconStateCapella["validators"]).push(
         ssz.capella.Validator.toViewDU({
           ...newValidatorFields,
-          // Very strange that this needs to be set at 0 else statroot mismatches with spec tests blocks,
-          // may be this will get updated later on to a proper epoch or field likely will be junked
+          // Very strange that this needs to be set at 0 rather than FAR_FUTURE_EPOCH else stateRoot mismatches
+          // with spec tests blocks.
+          // May be this will get updated later on to a proper epoch or field likely will be junked
           fullyWithdrawnEpoch: 0,
         })
       );
--- a/packages/state-transition/src/block/processDeposit.ts
+++ b/packages/state-transition/src/block/processDeposit.ts
@@ -72,25 +72,31 @@ export function processDeposit(fork: ForkSeq, state: CachedBeaconStateAllForks, 
 
     // add validator and balance entries
     const effectiveBalance = Math.min(amount - (amount % EFFECTIVE_BALANCE_INCREMENT), MAX_EFFECTIVE_BALANCE);
-    const newValidatorFields = {
-      pubkey,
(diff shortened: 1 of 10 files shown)


packages/state-transition/src/block/processDeposit.ts
```diff
@@ -72,31 +72,26 @@ export function processDeposit(fork: ForkSeq, state: CachedBeaconStateAllForks,
...
-          exitEpoch: FAR_FUTURE_EPOCH,
-          withdrawableEpoch: FAR_FUTURE_EPOCH,
-          effectiveBalance,
-          slashed: false,
-          fullyWithdrawnEpoch: FAR_FUTURE_EPOCH,
+          ...newValidatorFields,
+          // Very strange that this needs to be set at 0 rather than FAR_FUTURE_EPOCH else stateRoot mismatches
+          // with spec tests blocks.
+          // May be this will get updated later on to a proper epoch or field likely will be junked
+          fullyWithdrawnEpoch: 0,
```
dapplion, line 100: This field is no longer there in the current specs right? validator object is not modified in capella. So just set to 0 and we'll remove latter

packages/state-transition/src/block/processDeposit.ts
```diff
@@ -72,31 +72,26 @@ export function processDeposit(fork: ForkSeq, state: CachedBeaconStateAllForks,
...
+    const newValidatorFields = {
+      pubkey,
+      withdrawalCredentials: deposit.data.withdrawalCredentials,
+      activationEligibilityEpoch: FAR_FUTURE_EPOCH,
+      activationEpoch: FAR_FUTURE_EPOCH,
+      exitEpoch: FAR_FUTURE_EPOCH,
+      withdrawableEpoch: FAR_FUTURE_EPOCH,
+      effectiveBalance,
+      slashed: false,
+    };
```
dapplion: To reduce diff latter, just duplicate declaration

packages/state-transition/src/slot/upgradeStateToCapella.ts
```diff
@@ -0,0 +1,74 @@
...
+export function upgradeStateToCapella(stateBellatrix: CachedBeaconStateBellatrix): CachedBeaconStateCapella {
+  const {config} = stateBellatrix;
+
+  // Get underlying node and cast altair tree to bellatrix tree
+  //
+  // An bellatrix BeaconState tree can be safely casted to a capella BeaconState tree because:
+  // - Deprecated fields are replaced by new fields at the exact same indexes
+  // - All new fields are appended at the end
+  //
+  // altair                           | op   | altair
```
dapplion: bellatrix | op | capella

packages/state-transition/src/block/processDeposit.ts
g11tech: yes, correct

dapplion (approved):

PR #4682, opened 2022-10-21
Add warnings about starting API on public ip

wemeetagain:
**Motivation**

It may be a DoS vector to start the API on a public IP.
Users should be warned against doing this accidentally.

**Description**

- Add cli help warning
- Add log warning


wemeetagain:
Closing in favor of  #4683

PR #4684, opened 2022-10-22
Functionally disable identify-push protocol

wemeetagain:
**Motivation**

Without rate limits, identify-push protocol may be a DoS vector

**Description**

Disallow all inbound identify-push streams

A fix upstream should also allow toggling off the protocol entirely.


wemeetagain:
upstream change https://github.com/libp2p/js-libp2p/pull/1464

PR #4688, opened 2022-10-25
Try out Lavamoat for protection against supply chain attacks 

### LINKED ISSUE #3470: Process for managing dependencies

dadepo:
**Motivation**

Lavamoat is a set of tools for mitigating supply chain attacks. It does this via two main ways:

- ability to control lifecycle scripts. Provided by [@lavamoat/allow-scripts](https://github.com/LavaMoat/LavaMoat/blob/main/packages/allow-scripts)
- ability to freeze objects and control permitted privileges. Requires a runtime. Provided by [lavamoat-node](https://github.com/LavaMoat/LavaMoat/blob/main/packages/node)

This PR integrates `@lavamoat/allow-scripts` and configures it to run only dependency lifecycle hooks that has been specified.

 Exploring how lavamoat-node can be useful will be attempted in another PR.

An overview of how the tool works is as follows:

1. The tool works on the assumption that running lifecycle scripts has been disabled. By using `--ignore-scripts` flag or by configuration via `.yarnrc` or `.npmrc`
2. The module @lavamoat/preinstall-always-fail is to prevent accidentally running `yarn | yarn install` without disabling lifecycle script. With this module as dependency, the build fails on `yarn | yarn install` if lifecycle script is not disabled
3. the tool provides a way to whitelist the modules that should run lifecycle scripts
4. Those lifecycle scripts now need to be explicitly triggered by `yarn allow-scripts`

Running `yarn` directly will now fail with error like this:

```
Exit code: 1
Command: echo "Don't run npm lifecycle scripts by default, whitelist them with @lavamoat/allow-scripts" && exit 1
Arguments: 
Directory: ~chainsafe/official/lodestar/node_modules/@lavamoat/preinstall-always-fail
Output:
Don't run npm lifecycle scripts by default, whitelist them with @lavamoat/allow-scripts
```

`yarn setup` needs to be used in place of `yarn` or `yarn install`

- [x] revisit the usage of .yarnrc and how it can be removed
- [x] revisit and de-duplicate the setup command

Closes #3470


packages/db/test/unit/controller/level.test.ts
```diff
@@ -113,7 +113,7 @@ describe("LevelDB controller", () => {
     expect(result.length).to.be.equal(2);
   });
 
-  it("test compactRange + approximateSize", async () => {
+  it.skip("test compactRange + approximateSize", async () => {
```
dapplion: Do not unconditionally skip, add
```
if (!linux) this.skip()
```

dadepo: I still think the better approach is not to have platform dependent checks like this...but if you insist I'll add it. In the meantime this PR is in draft and I will make sure not to have the skip checked in when I mark it as ready for review.

.github/workflows/benchmark.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           key: ${{ runner.os }}-${{ steps.node.outputs.v8CppApiVersion }}-${{ hashFiles('**/yarn.lock', '**/package.json') }}
       - name: Install & build
         if: steps.cache-deps.outputs.cache-hit != 'true'
-        run: yarn install --frozen-lockfile --ignore-optional && yarn build
+        run: yarn install --frozen-lockfile --ignore-optional --ignore-scripts && yarn build
```
dapplion: Does this flag `--ignore-scripts` ignore all scripts? Or gets intercepted by lavamoat somehow to only apply to a subset of scripts?

dadepo: It ignores all scripts. I updated the description of the PR to give an overview of how lavamoat works

dapplion:
Stale PR, closing for now. Best to re-open once all conflicts are fixed

PR #4696, opened 2022-10-26
Update libp2p dashboard

dapplion:
**Motivation**

Improve libp2p dashboard for clarity

**Description**

- Add rate() operator
- Select adhoc filter per instance

_before_

Screenshot from 2022-10-26 16-47-46

_after_

Screenshot from 2022-10-26 16-45-48


wemeetagain (approved):
thanks, much better

Issue #4690, opened 2022-10-25
v1.2.0: investigate fromBytes performance issue


twoeths:
**Describe the bug**

With v1.2.0, on a node of 1000 keys, `fromBytes()` takes more than 8% of cpu time

<img width="1255" alt="Screen Shot 2022-10-25 at 21 49 27" src="https://user-images.githubusercontent.com/10568965/197806723-6658be0a-748a-43cb-9ac6-963e9c4c75d8.png">
[1025_lg1k_chacha20poly1305_no_mem_alloc.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/9861357/1025_lg1k_chacha20poly1305_no_mem_alloc.cpuprofile.zip)

**Expected behavior**

In v1.1.1, it takes only around 1% with the same type of node

<img width="1277" alt="Screen Shot 2022-10-25 at 21 50 37" src="https://user-images.githubusercontent.com/10568965/197807046-342b495f-7211-497d-98c1-71f6434df089.png">

[0525_lg1k_v1.1.1.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/9861372/0525_lg1k_v1.1.1.cpuprofile.zip)

dapplion:
Unrelated but important, noticed that on [0525_lg1k_v1.1.1.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/9861372/0525_lg1k_v1.1.1.cpuprofile.zip) afterProcessEpoch is run 3 times between 250500 ms and 255500 ms (~5s range)

![Screenshot from 2022-10-25 21-21-59](https://user-images.githubusercontent.com/35266934/197919536-f85dc9d8-6ecd-4070-b5b5-8309eaec54cc.png)

twoeths:
v1.2.0 process more attestation, looking into the number of `aggregateAttestationInto` calls it's 3.3x compared to v1.1

![Screen Shot 2022-10-26 at 11 42 17](https://user-images.githubusercontent.com/10568965/197936402-d785b60a-7b77-413f-85a6-f3443249b7c2.png)

v1.1.0
<img width="1291" alt="Screen Shot 2022-10-26 at 11 42 52" src="https://user-images.githubusercontent.com/10568965/197936474-10074323-ce27-46c3-a364-8fa97031c345.png">

in both nodes it receives the same number of valid attestations per second
[...]

PR #4697, opened 2022-10-27
Add metrics for aggregated attestations

twoeths:
**Motivation**

- I want to tweak SLOTS_RETAINED in attestation pools
- See performance of our published aggregated attestations

**Description**

- Add metrics to our attestation pool to see how many attestations are preaggregated
- Add metrics to show number of participations in our published AggregatedAttestation
- Since metrics are the same if I change `SLOTS_RETAINED` to 1, just leave it as is for now

part of #4690

The result shows fewer participations in our AggregatedAttestations than expected, probably because we drop so many unaggregated attestations
Screen Shot 2022-10-27 at 10 23 26

### DIFF
--- a/packages/validator/src/metrics.ts
+++ b/packages/validator/src/metrics.ts
@@ -169,6 +169,12 @@ export function getMetrics(register: MetricsRegister, gitData: LodestarGitData) 
       help: "Total published aggregates",
     }),
 
+    numParticipantsInAggregate: register.histogram({
+      name: "vc_attestation_service_participants_in_aggregate_total",
+      help: "Number of attestations in the published AggregatedAttestation",
+      buckets: [0, 50, 200, 500],
+    }),
+
     attestaterError: register.gauge<{error: "produce" | "sign" | "publish"}>({
       name: "vc_attestation_service_errors",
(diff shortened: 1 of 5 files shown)


wemeetagain (approved):
lgtm

packages/validator/src/metrics.ts
```diff
@@ -169,6 +169,12 @@ export function getMetrics(register: MetricsRegister, gitData: LodestarGitData)
       help: "Total published aggregates",
     }),
 
+    numParticipantsInAggregate: register.histogram({
+      name: "vc_attestation_service_participants_in_aggregate_total",
+      help: "Number of attestations in the published AggregatedAttestation",
+      buckets: [0, 50, 200, 500],
```
dapplion, line 175: Would be good to refine this buckets after watching metrics for a bit

twoeths: what I see on goerli is not great, it's 60-70 maximum
right now a sync committee on mainnet has 219 members, on goerli it's 189 and they'll gradually grow

dapplion: You mean beacon attestation committee?

Issue #4689, opened 2022-10-25
v1.2.0: isIP and isIPv4 performance issue


twoeths:
**Describe the bug**

this is from a node with novc running v1.2.0

<img width="1327" alt="Screen Shot 2022-10-25 at 20 35 43" src="https://user-images.githubusercontent.com/10568965/197788045-e36f5a83-99de-4936-b5b7-5f54b7e9cbbc.png">

[1025_as_chacha20poly1305_novc.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/9860771/1025_as_chacha20poly1305_novc.cpuprofile.zip)

**Expected behavior**

with v1.1.1, it almost takes 0% cpu time there

[1025_v1.1.1_sm1v.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/9860776/1025_v1.1.1_sm1v.cpuprofile.zip)

wemeetagain:
Difference is that newer versions of `is-ip` package uses the `super-regex` which is way overkill and adds a lot of overhead.

We probably want to revert https://github.com/multiformats/js-multiaddr/pull/260

wemeetagain:
See https://github.com/ChainSafe/is-ip/pull/1

PR #4701, opened 2022-10-29
Force is-ip v4.0.0

wemeetagain:
**Motivation**

See #4689 

**Description**

Until we fix things upstream, force v4.0.0


dapplion (approved):

wemeetagain:
I think this doesn't work :(

PR #4702, opened 2022-10-29
Save and use full LightClientUpdate instead of partial

g11tech:
In the previous scheme of things, the sync committee branch corresponding to finalized checkpoint's state was fetched and bundled, but that shifted from finalized checkpoints state to the attested header's state.

But the non checkpoint syncComitteeWitness data gets pruned later and hence the best partial update was no more servicable causing broken get update api.

This PR fixes by storing the full update instead of partial because all corresponding data is available while storing the update, so that the data availability doesn't become the issue post pruning.

Console log of the light client demo which can now properly process the lightclient updates to sync to the current (the new url fetch errors are because of the latest merged p2p PR which might have updated some routes, the light client demo to be independently updated)

image


packages/db/src/schema.ts
```diff
@@ -60,7 +60,7 @@ export enum Bucket {
   lightClient_syncCommittee = 52, // Root(altair.SyncCommittee) -> altair.SyncCommittee
   // TODO: May be redundant to block stores
   lightClient_checkpointHeader = 53, // BlockRoot -> phase0.BeaconBlockHeader
-  lightClient_bestPartialLightClientUpdate = 54, // SyncPeriod -> PartialLightClientUpdate
+  lightClient_bestLightClientUpdate = 54, // SyncPeriod -> LightClientUpdate
```
wemeetagain: Should we use a new bucket for a non-errored migration?

wemeetagain (approved):

PR #4704, opened 2022-10-30
Use newer @multiformats/multiaddr

wemeetagain:
**Motivation**

Better alternative to #4701 

**Description**

Update multiaddr to the latest. Should be updated again after https://github.com/multiformats/js-multiaddr/pull/281 is released


twoeths (approved):
lgtm, do we need to also change `@multiformats/multiaddr` to 11.0.6 in package.json @wemeetagain ?

PR #4712, opened 2022-11-02
Lazy aggregation

### LINKED ISSUE #4690: v1.2.0: investigate fromBytes performance issue

twoeths:
**Motivation**

- For attestation pools, right now we do preaggregation which has `fromBytes()` performance issue and the aggregated attestations may not be used if no connected validator is an aggregator.

**Description**

- Attestation pool:
  - Do aggregation only once getAggregate() is called
  - Cached the aggregated attestations because there could be up to 16 aggregators per committee
- Aggregated Attestation pool: only do aggregation when we call `getAttestationsForBlock`
- Add benchmark for `bls.Signature.aggregate`

Closes #4690

### DIFF
--- a/packages/beacon-node/src/chain/opPools/attestationPool.ts
+++ b/packages/beacon-node/src/chain/opPools/attestationPool.ts
@@ -36,6 +36,11 @@ type AggregateFast = {
    *   which is a less busy time than 1/3 of the slot.
    */
   signatures: Uint8Array[];
+  /**
+   * There could be up to TARGET_AGGREGATORS_PER_COMMITTEE aggregator per attestation data and all getAggregate() call
+   * tends to be at the same time so we want to cache the aggregated attestation, invalidate cache per new unaggregated attestation.
+   */
+  aggregatedAttestation: phase0.Attestation | null;
 };
 
 /** Hex string of DataRoot `TODO` */
@@ -192,6 +197,7 @@ function aggregateAttestationInto(aggregate: AggregateFast, attestation: phase0.
 
   aggregate.aggregationBits.set(bitIndex, true);
   aggregate.signatures.push(attestation.signature);
+  aggregate.aggregatedAttestation = null;
   return InsertOutcome.Aggregated;
 }
 
@@ -204,6 +210,7 @@ function attestationToAggregate(attestation: phase0.Attestation): AggregateFast 
     // clone because it will be mutated
     aggregationBits: attestation.aggregationBits.clone(),
     signatures: [attestation.signature],
+    aggregatedAttestation: null,
   };
 }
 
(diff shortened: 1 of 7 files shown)


packages/beacon-node/src/chain/opPools/attestationPool.ts
```diff
@@ -25,7 +25,22 @@ const MAX_ATTESTATIONS_PER_SLOT = 16_384;
...
+   * - Defer aggregation until getAggregate(). Consumes more memory but prevents extra work by only doing the
+   *   aggregation if the connected validator is an aggregator. The aggregation is done during 2/3 of the slot
+   *   which is a less busy time than 1/3 of the slot.
+   */
+  signatures: Uint8Array[];
+  /**
+   * There could be up to TARGET_AGGREGATORS_PER_COMMITTEE aggregator per attestation data and all getAggregate() call
+   * tends to be at the same time so we want to cache the aggregated attestation, invalidate cache per new unaggregated attestation.
+   */
+  aggregatedAttestation: phase0.Attestation | null;
```
wemeetagain, line 43: I don't see where this is ever being set, only ever being invalidated or instantiated as null.

twoeths: oh good catch, thanks

dapplion:
Converting to draft until this is being proven to be good for performance in heavy networks

twoeths:
There were some concerns the last time probably due to source deployment vs unstable #4727 (and it makes an impression that `unstable` is always better than feature branches), I will deploy to some nodes and monitor them again

twoeths:
The metrics are clearly worse than unstable/stable group, especially on `md16` node:
- Attestation gossip validation time (vs unstable 250ms): 
Screen Shot 2022-11-14 at 10 40 49
- Block processor (vs unstable 1.2s - 1.8s)
Screen Shot 2022-11-14 at 10 42 08

Since we didn't touch above code paths, the real difference should come from vm metrics. The prominent different metric is `Process heap bytes`
- On feat1 md16 (this branch):
Screen Shot 2022-11-14 at 10 44 43

- vs unstable md16
Screen Shot 2022-11-14 at 10 45 10

looks like aggregating a lot of signatures at the same time causes a memory issue

twoeths:
should rebase against `unstable` and try again since there was some improvement in #4838

twoeths:
the initial implementation already ignored signature verification

attestation gossip validaiton time is not bad (300ms-500ms), it's not as good as unstable since it's a github source deployment) but memory pattern on `md16` node is the same which is a concern

twoeths:
should not be an issue anymore after we merge #5284

PR #4714, opened 2022-11-02
Add `--mdns` flag

### LINKED ISSUE #4628: Make MDNS optional with --mdns
Currently mdns is always on due to legacy logic in `packages/beacon-node/src/network/nodejs/bundle.ts`, should be optional instead

wemeetagain:
**Motivation**

- resolve #4628 

**Description**

Add `--mdns` flag
Default mdns to disabled


dapplion (changes requested):
This flag should also cause peers from MDNS to be connected to by the node which is not the case. Peer Manager should also consider MDNS peers, but now only listens to discv5 peers

packages/beacon-node/src/network/peers/discover.ts
```diff
@@ -131,17 +133,32 @@ export class PeerDiscovery {
   async start(): Promise<void> {
     await this.discv5.start();
     this.discv5StartMs = Date.now();
-    this.discv5.on("discovered", this.onDiscovered);
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
+    for (const service of (this.libp2p as any).services) {
```
dapplion: Cast to a Partial type that has the properties you need

wemeetagain:
```suggestion
    for (const service of ((this.libp2p as unknown) as {services: unknown[]}).services) {
```

packages/beacon-node/src/network/peers/discover.ts
```diff
@@ -245,18 +261,69 @@ export class PeerDiscovery {
...
+   */
+  private onDiscoveredPeer = async (evt: CustomEvent<PeerInfo>): Promise<void> => {
+    const {id, multiaddrs} = evt.detail;
+    const status = await this.handleDiscoveredPeer(id, multiaddrs[0]);
+    this.metrics?.discovery.discoveredStatus.inc({status});
+  };
+  /**
+   * Progressively called by libp2p peer discovery as a result of any query.
+   */
+  private async handleDiscoveredPeer(peerId: PeerId, multiaddrTCP: Multiaddr): Promise<DiscoveredPeerStatus> {
```
dapplion, line 264: Is this repeated code from below?

dapplion:
Failed e2e tests is:

``` 
@lodestar/beacon-node:   1) network
@lodestar/beacon-node:        should delete a peer on disconnect:
@lodestar/beacon-node:       AssertionError: expected 1 to equal +0
@lodestar/beacon-node:       + expected - actual
@lodestar/beacon-node:       -1
@lodestar/beacon-node:       +0
@lodestar/beacon-node:       
@lodestar/beacon-node:       at Context.<anonymous> (file:///home/runner/work/lodestar/lodestar/packages/beacon-node/test/e2e/network/network.test.ts:150:72)
```

packages/beacon-node/src/network/peers/discover.ts
wemeetagain: its similar. doesn't include everything but I'm not sure how to dedupe. thoughts or commits welcome

wemeetagain:
ready for re-review, tests passing yay

wemeetagain:
@ChainSafe/lodestar bump

packages/beacon-node/src/network/peers/discover.ts
```diff
@@ -131,17 +133,31 @@ export class PeerDiscovery {
...
+      this.discv5.kadValues().forEach((enr) => this.onDiscoveredENR(enr));
     }
   }
 
   async stop(): Promise<void> {
-    this.discv5.off("discovered", this.onDiscovered);
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
+    for (const service of (this.libp2p as any).services) {
+      if (isPeerDiscovery(service)) {
+        service.removeEventListener("peer", this.onDiscoveredPeer);
```
dapplion: Why don't we bake mdns into peer discovery instead of having to add it to libp2p?

packages/beacon-node/src/network/peers/discover.ts
```diff
@@ -245,18 +261,69 @@ export class PeerDiscovery {
     }
   }
 
+  /**
+   * Progressively called by libp2p peer discovery as a result of any query.
+   */
+  private onDiscoveredPeer = async (evt: CustomEvent<PeerInfo>): Promise<void> => {
+    const {id, multiaddrs} = evt.detail;
+    const status = await this.handleDiscoveredPeer(id, multiaddrs[0]);
+    this.metrics?.discovery.discoveredStatus.inc({status});
```
dapplion, line 259: Would be best to treat mdns discovered peers via different metrics. Or at least know how many peers are from MDNS

packages/beacon-node/src/network/peers/discover.ts
```diff
@@ -245,18 +261,69 @@ export class PeerDiscovery {
     }
   }
 
+  /**
+   * Progressively called by libp2p peer discovery as a result of any query.
+   */
+  private onDiscoveredPeer = async (evt: CustomEvent<PeerInfo>): Promise<void> => {
```
twoeths, line 256: so now we have onDiscoveredPeer and onDiscoveredENR, could them be called together on the same peer? 

that's already an issue with v1.2.1
https://github.com/ChainSafe/lodestar/issues/4790

dapplion (dismissed):
LGTM! Some points, but this can be merged as-is

twoeths:
@wemeetagain currently we don't have to handle `"peer"` events, why do we have to handle it in this PR?

also I see libp2p already handle `"peer"` event, then add to "addressBook" and "protoBook" and reemit `"peer:discovery"`, should we handle `"peer:discovery"` instead, or we don't need to handle any of them at all like what we do currently?

wemeetagain:
merging and making issues for the open comments

wemeetagain:
@tuyennhv your comment makes sense. We can/should use the "peer:discovery" event instead

packages/beacon-node/src/network/peers/discover.ts
wemeetagain: Ended up going the other direction. Listening to libp2p's "peer:discovery" event

twoeths (approved):

PR #4717, opened 2022-11-02
Update libp2p to 0.41.0

wemeetagain:
**Motivation**

We want upstream fixes and other goodies. Current version

https://github.com/ChainSafe/lodestar/blob/7d662d7951db27edcee15fd7266bed322f0a3882/packages/beacon-node/package.json#L142

**Description**

Minimal edits to update to latest libp2p and networking dependencies


wemeetagain:
It seems there's a memory leak.

Screenshot shows high process memory (the spike back down is process being restarted)

Screenshot from 2023-01-04 11-27-54

wemeetagain:
This leak may also be present in the unstable branch. Current theory is its related to the rate of inbound peer connections.

dapplion (approved):

twoeths:
> This leak may also be present in the unstable branch. Current theory is its related to the rate of inbound peer connections.

yes, specifically it's due to https://github.com/ChainSafe/node-snappy-stream/issues/16

PR #4715, opened 2022-11-02
Update multistream-select

wemeetagain:
**Motivation**

- Resolve https://github.com/ChainSafe/lodestar/issues/4709

**Description**

Bump multistream-select dependency

Issue #4709, opened 2022-11-01
v1.2.0: multistream-select performance issue


twoeths:
**Describe the bug**

`multistream-select` calls `it-handshake` which call `it-map` and it takes 6% of cpu time in `md16` node

<img width="1333" alt="Screen Shot 2022-11-01 at 19 46 24" src="https://user-images.githubusercontent.com/10568965/199235573-0ccf4e58-4000-4dd6-89f4-64720ef5f68d.png">

[0111_multi_stream_select_3.1.1_lg1k.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/9909903/0111_multi_stream_select_3.1.1_lg1k.cpuprofile.zip)

8.75% in a `novc` node

<img width="1432" alt="Screen Shot 2022-11-01 at 19 49 15" src="https://user-images.githubusercontent.com/10568965/199236009-856054b8-d918-4711-995d-366f81201fcc.png">

[1031_v1.2.0_rc.2_novc.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/9909940/1031_v1.2.0_rc.2_novc.cpuprofile.zip)

the root cause is https://github.com/libp2p/js-libp2p-multistream-select/pull/28

**Expected behavior**

The above PR is included in multistream-select 3.1.1, try it and make sure there is no performance issue

dapplion:
Upstream PR merged, reopen is still happens

PR #4718, opened 2022-11-03
it-handshake 4.1.2

twoeths:
**Motivation**

The it-handshake has bad performance, we want to migrate to its latest version

**Description**

- #4715 upgrade multistream-select to use it-handshake 4.1.2 in the end but we don't have good mesh peers with that update
- This effort is to only update it-handshake 4.1.2 to avoid side effects in multistream-select

part of #4709


twoeths:
duplicate to https://github.com/ChainSafe/lodestar/pull/4719

PR #4720, opened 2022-11-04
BLS: use worker with lowest latency

### LINKED ISSUE #4653: Unstable BLS performance
**Describe the bug**

On deployment of `unstable` or `v1.2.0-rc.0` to `unstable, beta, feat1, feat2`, I see different performance of BLS, which leads to different time on gossip validation.

- On beta since we deploy `v1.2.0-rc.0`, we have worse BLS performance: longer `from worker latency`, longer job wait time, lower utilization rate

Screen Shot 2022-10-09 at 15 51 17

- On unstable group, both the job wait time is smaller/more stable, also the `job wait time`

Screen Shot 2022-10-09 at 15 52 38

Worse BLS performance leads to longer gossip validation time
- Beta

Screen Shot 2022-10-09 at 15 58 00

- Unstable

Screen Shot 2022-10-09 at 15 59 00

The worse BLS performance, the worse performance of node overall including the I/O lagged issue in #4600

Similar metrics are monitored in feat1 and feat2

**Expected behavior**

Stable performance of BLS

twoeths:
**Motivation**

There is different performance between `beta` and `unstable` groups although they have same commit.

What was found in #4653 is we have different performance based on worker latency.

**Description**

- Metrics show that if we tend to keep using a worker for BLS even though  that cpu could be used by any jobs outside of lodestar
- This change records the latency every time we run job and pick a worker with lowest latency the next time

Closes #4653

### DIFF
--- a/packages/beacon-node/src/chain/bls/multithread/index.ts
+++ b/packages/beacon-node/src/chain/bls/multithread/index.ts
@@ -85,6 +85,7 @@ type WorkerStatus =
 type WorkerDescriptor = {
   worker: Worker;
   status: WorkerStatus;
+  latency: number;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


twoeths:
see https://github.com/ChainSafe/lodestar/issues/4653#issuecomment-1303186375

PR #4721, opened 2022-11-04
Update the ping time for inboud and outbound

### LINKED ISSUE #4645: Socket read timeout in CI tests
**Describe the bug**

libp2p-tcp closes streams upon 30 seconds on inactivity (both inbound and outbound), and will close the connection
https://github.com/libp2p/js-libp2p-tcp/blob/ceac538e1e08f74fe6e509ca2fa5546876b7f51e/src/socket-to-conn.ts#L62

libp2p-tcp uses Socket setTimeout() which is a NodeJS feature from it's data stream. So timeout means:
- No data seen in that stream for $timeout miliseconds

Current sim tests suffer from this timeouts:
- Are really the nodes not sending any data for 30 seconds?
- Is the 30 seconds timeout a good value?
- Why is this issue only showing up now?
- There should be metrics reporting the reasons to severe connections

nazarhussain:
**Motivation**

Avoid socket timeout if the node tries to load with higher genesis delay. 

**Description**

Here are some stats from this change extracted for last **24 hours** and drawn over **1h interval**.

**nogroup-ctvpss-0 - current branch**
image

**nogroup-ctvpss-1 - current branch**
image

**nogroup-ctvpss-2 - unstable**
image

**nogroup-ctvpss-3 - unstable**
image

Closes  #4645 

Run tests

### DIFF
--- a/packages/beacon-node/src/network/peers/peerManager.ts
+++ b/packages/beacon-node/src/network/peers/peerManager.ts
@@ -31,8 +31,8 @@ import {
 /** heartbeat performs regular updates such as updating reputations and performing discovery requests */
 const HEARTBEAT_INTERVAL_MS = 30 * 1000;
 /** The time in seconds between PING events. We do not send a ping if the other peer has PING'd us */
-const PING_INTERVAL_INBOUND_MS = 4 * 60 * 1000 - 11 * 1000; // Offset to not ping when outbound reqs
-const PING_INTERVAL_OUTBOUND_MS = 4 * 60 * 1000;
+const PING_INTERVAL_INBOUND_MS = 15 * 1000; // Offset to not ping when outbound reqs
+const PING_INTERVAL_OUTBOUND_MS = 20 * 1000;
 /** The time in seconds between re-status's peers. */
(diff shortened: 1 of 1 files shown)


nazarhussain:
My understanding to this change. 

- As I suspected the decreasing the ping time, increases the ping requests 
- Inbound ping requests increases to almost double 
- Outbound ping requests increases drastically
- As of my understanding of the Grafana metics dashboard it shows these values **per minute**
- So even with the change in the PR, having maximum 2 ping requests per minute does not seems to aggressive.

dapplion:
@nazarhussain Could you analyze overall bandwidth changes?

nazarhussain:
@dapplion Here are the network packet traffic. Observations are: 

1. The maximum number of packets/s are almost the same. 
2. While the average is quite hight 
3. I am not sure if this is because the nodes were syncing previously, but in that case the maximum packets should be higher as well.
4. A two nodes which this change and two with unstable show similar metrics, that implies there are no false positive. 



image

image

image

image

dapplion (approved):
IMO metrics look good so I would merge

PR #4729, opened 2022-11-06
Fix typos in variables, comments and metric names

nflaig:
**Motivation**

I was reviewing the code and just fixed some typos along the way.

**Description**

This PR fixes typos in variables and comments. I also updated typos in Prometheus metric names in both the code and the dashboard but not sure about other implications as metric names could be considered an external API if there are other consumers of those metrics. If renaming the metrics could be problematic I would rather keep the current names with the typos.

### DIFF
--- a/packages/beacon-node/src/chain/errors/lightClientError.ts
+++ b/packages/beacon-node/src/chain/errors/lightClientError.ts
@@ -22,7 +22,7 @@ export class LightClientError extends GossipActionError<LightClientErrorType> {}
 // Errors for the light client server
 
 export enum LightClientServerErrorCode {
-  RESOURCE_UNAVAILABLE = "RESOURCE_UNAVALIABLE",
+  RESOURCE_UNAVAILABLE = "RESOURCE_UNAVAILABLE",
 }
 
 export type LightClientServerErrorType = {code: LightClientServerErrorCode.RESOURCE_UNAVAILABLE};
--- a/packages/beacon-node/src/metrics/metrics/lodestar.ts
+++ b/packages/beacon-node/src/metrics/metrics/lodestar.ts
@@ -93,21 +93,21 @@ export function createLodestarMetrics(
     }),
     peersRequestedToConnect: register.gauge({
       name: "lodestar_peers_requested_total_to_connect",
-      help: "Priorization results total peers count requested to connect",
+      help: "Prioritization results total peers count requested to connect",
     }),
     peersRequestedToDisconnect: register.gauge<"reason">({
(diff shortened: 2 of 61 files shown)


packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -93,21 +93,21 @@ export function createLodestarMetrics(
     }),
     peersRequestedToConnect: register.gauge({
       name: "lodestar_peers_requested_total_to_connect",
-      help: "Priorization results total peers count requested to connect",
+      help: "Prioritization results total peers count requested to connect",
```
nflaig, line 96: Here I am not sure what `Prioritization results`  in the `help` sentence means to be honest, the metric name on the other hand is clear to me. Maybe someone could clarify this.

packages/beacon-node/src/chain/errors/lightClientError.ts
```diff
@@ -22,7 +22,7 @@ export class LightClientError extends GossipActionError<LightClientErrorType> {}
 // Errors for the light client server
 
 export enum LightClientServerErrorCode {
-  RESOURCE_UNAVAILABLE = "RESOURCE_UNAVALIABLE",
+  RESOURCE_UNAVAILABLE = "RESOURCE_UNAVAILABLE",
```
nflaig, line 25: Are error codes parsed/checked somewhere? I did not find a reference to `RESOURCE_UNAVALIABLE` anywhere in the code itself but if this error code is referenced somewhere else should probably keep the value despite the typo.

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -572,19 +572,19 @@ export function createLodestarMetrics(
     // Gossip block
     gossipBlock: {
       elapsedTimeTillReceived: register.histogram({
-        name: "lodestar_gossip_block_elappsed_time_till_received",
-        help: "Time elappsed between block slot time and the time block received via gossip",
+        name: "lodestar_gossip_block_elapsed_time_till_received",
```
nflaig, line 575: Changing metric names is potentially problematic, needs to be confirmed by someone with more insights that this is not a problem as long as the corresponding value is also updated in the Grafana dashboard.

wemeetagain (approved):
Thanks for the contribution and thoughtful consideration of the different cases.

I'm in favor of ripping off the bandaid and updating everything.

nflaig:
@wemeetagain thanks for the review :thumbsup:

PR #4734, opened 2022-11-07
Fix beacon/states/validator endpoint

### LINKED ISSUE #4731: Fix `validators/{validator_id}` API endpoint
**Describe the bug**
From discord: https://discord.com/channels/593655374469660673/593655641445367808/1039215264023191612
Thanks. The beacon chain API spec says that I should be able to access an individual validator's state at /eth/v1/beacon/states/{state_id}/validators/{validator_id} . I've been specifically /eth/v1/beacon/states/head/validators/{validator_id}. I always get "Validator not found" error.  In closed issue 3852, the syntax is shown as /eth/v1/beacon/states/head/validators?id={validator_id}. Maybe there is something I don't understand about the API spec, but I did not think passing the validator ID as a query param was part of the spec. In any case, even when I call it using this syntax, I get {"data":[],"execution_optimistic":false} for any validator ID I try. Data is empty. Is there something I'm doing wrong? I can confirm that endpoint /eth/v1/beacon/states/head/validators without a validator ID at all does work correctly to return all validators, but I  am unable to query for just a single validator's state. 

**Expected behavior**

Endpoint matches API spec and returns valid data

wemeetagain:
**Motivation**

Resolves #4731 

**Description**

Refactor `getStateValidatorIndex` to properly handle inputs required of it and to return more information in failure cases.

### DIFF
--- a/packages/beacon-node/src/api/impl/beacon/state/index.ts
+++ b/packages/beacon-node/src/api/impl/beacon/state/index.ts
@@ -62,8 +62,9 @@ export function getBeaconStateApi({
       const validatorResponses: routes.beacon.ValidatorResponse[] = [];
       if (filters?.id) {
         for (const id of filters.id) {
-          const validatorIndex = getStateValidatorIndex(id, state, pubkey2index);
-          if (validatorIndex != null) {
+          const resp = getStateValidatorIndex(id, state, pubkey2index);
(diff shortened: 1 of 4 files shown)


twoeths (approved):
lgtm

Issue #4733, opened 2022-11-07
Support serving sync committee hashes


philknows:
**Is your feature request related to a problem? Please describe.**

For light clients, we are experimenting with new ways to ensure we can enable light client sync in the most optimal way possible, with the best assurances of truth. Using https://github.com/lightclients/kevlar for optimistic syncing, we can easily ask for hashes of the sync committee to make comparisons amongst a set of provers. We would just need an easy way to get this information from clients like Lodestar to serve this data through an endpoint.

**Describe the solution you'd like**

To help with the optimistic construction, it would be ideal to add an endpoint to fetch the hashes off the sync committee such as `getCommitteeHash` or `getOptimisticHash` with a route of something similar to `/eth/v1/beacon/light_client/committee_hash`. We should implement this in Lodestar, test it and then push upstream to include it in the beaconAPIs spec. 

**Additional context**

This feature is to help with supporting better syncing methods for software such as Kevlar using the light client protocol.

shresthagrawal:
Thanks, @philknows for creating the initial issue. A small addition to the endpoint structure. It would be good to have support for downloading these hashes in batches. So something like `/eth/v1/beacon/light_client/committee_hash?start_period=x&period_cnt=y`. Feel free to check out the existing prover implementation [here](https://github.com/lightclients/kevlar/blob/main/src/provers/light-optimistic/server.ts#L17)

twoeths:
when it comes to the spec, I think it's more suitable to go with sync committee root instead of hash to compliant to all of the specs in ethereum consensus layer. Also in lodestar, we track sync committee root (not hash).

The PR #4736 still go with committee hash to serve the need of kevlar atm.

what do you think @shresthagrawal ?

shresthagrawal:
@tuyennhv sync committee root should work as well. I guess that would be preferred as we would like to push it to the official spec as well

philknows:
Closed and implemented, but there's more work to be done to get this standardized in the spec. And in the meantime, move this to the Lodestar namespace.

PR #4736, opened 2022-11-08
Implement committee_hash lightclient api

twoeths:
**Motivation**

- Kevlar https://github.com/lightclients/kevlar needs this for its optimistic sync

**Description**

- From a period, get best update
- Extract attester header than beacon block root from there
- Get sync committee witness from `syncCommitteeWitness` db
- Get current sync committee from `syncCommittee` db

part of #4733


shresthagrawal:
Thanks for taking the task! I guess you haven't added the hashing part yet? The hash should be simply the keccak256 of the concatenated public keys of the sync committee. [Here](https://github.com/lightclients/kevlar/blob/main/src/provers/light-optimistic/memory-store.ts#L20) is the current prover implementation

packages/api/src/beacon/routes/lightclient.ts
```diff
@@ -60,6 +64,7 @@ export const routesData: RoutesData<Api> = {
   getOptimisticUpdate: {url: "/eth/v1/beacon/light_client/optimistic_update", method: "GET"},
   getFinalityUpdate: {url: "/eth/v1/beacon/light_client/finality_update", method: "GET"},
   getBootstrap: {url: "/eth/v1/beacon/light_client/bootstrap/{block_root}", method: "GET"},
+  getCommitteeHash: {url: "/eth/v1/beacon/light_client/committee_hash", method: "GET"},
```
dapplion: Use `v0` until this route is spec'ed in beacon-APIs repo

dapplion (dismissed):

packages/beacon-node/src/chain/lightClient/index.ts
```diff
@@ -291,6 +293,31 @@ export class LightClientServer {
     return update;
   }
 
+  /**
+   * API ROUTE to get the current sync committee from the best available update for `period`.
```
dadepo: minor: the comment here is semantically different from the corresponding comment on api route

dadepo:
Maybe consider adding tests that covers the added endpoint? apart from that LGTM.

twoeths:
> Maybe consider adding tests that covers the added endpoint? apart from that LGTM.

it already covered the end point as long as we update the testData for lightclient, doubled checked it via the unit test output

```
  beacon / lightclient
onError: Error: aborted
    ✔ getStateProof
    ✔ getUpdates
    ✔ getOptimisticUpdate
    ✔ getFinalityUpdate
    ✔ getBootstrap
    ✔ getCommitteeHash
```

dapplion:
> Thanks for taking the task! I guess you haven't added the hashing part yet? The hash should be simply the keccak256 of the concatenated public keys of the sync committee. [Here](https://github.com/lightclients/kevlar/blob/main/src/provers/light-optimistic/memory-store.ts#L20) is the current prover implementation

@shresthagrawal Why don't you do the hashTreeRoot of the SyncCommittee instead of concat pubkeys + hash? It's kindof tangent to your classic object hashing in the beacon chain. Plus the hashTreeRoot of the SyncCommittee is usually already available due to state hashing.

shresthagrawal:
@dapplion yes that works, I already mentioned that in the GitHub issue earlier https://github.com/ChainSafe/lodestar/issues/4733#issuecomment-1311457596

dadepo:
> > Maybe consider adding tests that covers the added endpoint? apart from that LGTM.
> 
> it already covered the end point as long as we update the testData for lightclient, doubled checked it via the unit test output
> 
> ```
>   beacon / lightclient
> onError: Error: aborted
>     ✔ getStateProof
>     ✔ getUpdates
>     ✔ getOptimisticUpdate
>     ✔ getFinalityUpdate
>     ✔ getBootstrap
>     ✔ getCommitteeHash
> ```

Unfortunately that test only seem to test some of the type machinery and not the actual implementation. 

I updated the implementation to throw an error like this:

```
    async getCommitteeHash(startPeriod: SyncPeriod, count: number) {
      throw new Error("error");
      const maxAllowedCount = Math.min(MAX_REQUEST_LIGHT_CLIENT_COMMITTEE_HASHES, count);
      const periods = Array.from({length: maxAllowedCount}, (_ignored, i) => i + startPeriod);
      const committeeHashes = await Promise.all(
        periods.map((period) => chain.lightClientServer.getCommitteeHash(period))
      );
      return {data: committeeHashes};
    },
```

Ran the test and also got:

```
  beacon / lightclient
    ✔ getStateProof
    ✔ getUpdates
    ✔ getOptimisticUpdate
    ✔ getFinalityUpdate
    ✔ getBootstrap
    ✔ getCommitteeHash
```

You can also confirm this. 

What will be needed is e2e test, and the few existing ones can be found in `packages/beacon-node/test/e2e/api/impl/beacon`

twoeths:
> What will be needed is e2e test, and the few existing ones can be found in packages/beacon-node/test/e2e/api/impl/beacon

@dadepo thanks, implemented it

packages/beacon-node/src/chain/lightClient/index.ts
```diff
@@ -291,6 +293,31 @@ export class LightClientServer {
...
+    }
+
+    const currentSyncCommittee = await this.db.syncCommittee.get(syncCommitteeWitness.currentSyncCommitteeRoot);
+    if (!currentSyncCommittee) {
+      throw new LightClientServerError(
+        {code: LightClientServerErrorCode.RESOURCE_UNAVAILABLE},
+        `currentSyncCommittee not available ${toHexString(blockRoot)} for period ${period}`
+      );
+    }
+    return digest(concat(currentSyncCommittee.pubkeys));
```
dapplion: Do we agree on doing hashTreeRoot here or not?

twoeths: I can switch to use root unless @shresthagrawal wants to use hash for now

shresthagrawal: Yes please go ahead with the root!

wemeetagain (approved):
LGTM

PR #4737, opened 2022-11-08
Update @chainsafe/discv5 to v1.4.0

wemeetagain:


dapplion:
Build is breaking since metrics now require extra properties. discv5 should switch to the metrics generator pattern

wemeetagain:
added new metrics, should be good to go

will open an issue there to switch to generator pattern

dapplion (approved):

Issue #4743, opened 2022-11-10
Failed to produce block


twoeths:
**Describe the bug**

This happens on mainnet, v1.2.0

Validator log

```
34943:Nov-09 19:07:11.219[]                error: Error proposing block slot=5099734, validator=0x9458…b482 Failed to produce engine or builder block - Failed to produce block
```

Beacon log

```
Nov-09 19:07:11.175[rest]            ^[[31merror^[[39m: Req req-54l produceBlockV2 error  No block for root 0x858b12df7d85d42459c04ae99d4142a593dd0a990d720cf3b68cc98a5e0cccc8
Error: No block for root 0x858b12df7d85d42459c04ae99d4142a593dd0a990d720cf3b68cc98a5e0cccc8
    at ProtoArray.getBlockReadonly (file:///usr/app/packages/fork-choice/src/protoArray/protoArray.ts:879:13)
    at ForkChoice.getDependentRoot (file:///usr/app/packages/fork-choice/src/forkChoice/forkChoice.ts:842:29)
    at isValidAttestationData (file:///usr/app/packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts:449:47)
    at AggregatedAttestationPool.getAttestationsForBlock (file:///usr/app/packages/beacon-node/src/chain/opPools/aggregatedAttestationPool.ts:137:14)
    at BeaconChain.produceBlockBody (file:///usr/app/packages/beacon-node/src/chain/produceBlock/produceBlockBody.ts:91:55)
    at BeaconChain.produceBlockWrapper (file:///usr/app/packages/beacon-node/src/chain/chain.ts:366:36)
    at Object.produceBlock (file:///usr/app/packages/beacon-node/src/api/impl/validator/index.ts:228:21)
    at Object.handler (file:///usr/app/packages/api/src/utils/server/genericJsonServer.ts:41:23)
```

**Expected behavior**

Lodestar should be able to produce block

twoeths:
this issue does happen on goerli but it's so rare

```
grep -e "produceBlockV2" -rn beacon-2022-11-* | wc -l
124
```

```
grep -e " produceBlockV2 error  No block for root" -rn beacon-2022-11-* | wc -l
2
```

twoeths:
at slot 5099733, epoch 159364 was finalized (slot 5099648)

```
[...]

PR #4745, opened 2022-11-10
Produce block: ignore old attestations

twoeths:
**Motivation**

Failed to produce attestations, this happens if the `produceBlock` function try to include an old attestation (which rarely happens), for the scenario in #4743:
- Lodestar call `produceBlock` for slot 5099734
- We try to include an attestation of an old slot (suppose it < 5099712)
- It tried to get dependent root at 5099647
- But finalized slot is at 5099648

**Description**

- For v1.1.x, `findAttesterDependentRoot()` returns null in this case
- For v1.2.0, due to #4555, the new function `getDependentRoot()` throws error instead
- The fix is to catch the error and simply ignore that attestation data
- Also add a new metric to track number of aggregated attestations in lodestar blocks

Closes #4555


g11tech (approved):
lgtm!

twoeths:
there are some failed tests that not related to this PR

PR #4756, opened 2022-11-11
Run sim tests on self hosted runner

dapplion:
**Motivation**

Github runners are not powerful enough to handle sim tests reliably. We want to be sure that sim tests are run with enough resources

**Description**

Run sim tests on self hosted runner

### DIFF
--- a/.github/workflows/test-sim.yml
+++ b/.github/workflows/test-sim.yml
@@ -13,7 +13,7 @@ env:
 jobs:
   tests-sim:
     name: Sim tests
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, ubuntu, ax161, lodestar]
     steps:
       # <common-build> - Uses YAML anchors in the future
       - uses: actions/checkout@v2
@@ -40,7 +40,7 @@ jobs:
       # </common-build>
 
       - name: Pull Geth
-        run: | 
+        run: |
           TAR_NAME=$(basename $GETH_IMAGE)
           BINARY_FOLDER="${TAR_NAME%.*.*}"
           curl -sS $GETH_IMAGE > $TAR_NAME


dapplion:
Putting as draft after some experiments:

Self hosted runners operate differently: they don't start from a clean slate but just run on bare metal and **do not clean up**. Instead Github self hosted runners run on fresh VMs that are destroyed after the fact.

From an offline exchange with Pari:

> L: When using Github self hosted runners, every registered runner can run exactly 1 job at a time? Do you know if that can be customized? In your usage of it with kurtosis is there any issue with clean-up after the fact? Being used to Github own runners where you get a fresh environment every time that is destroyed after the fact, using self hosted has some challenges
> P: yeah so you can't customize it - we had the same issue on our system. So the fix is to register multiple runners on the same physical host so you "hack" it in. When no jobs are running, nothing changes about the usage anyway. The correct way to do it would be to use something like:
> - Actuated: https://blog.alexellis.io/blazing-fast-ci-with-microvms/
> - Kubernetes with the autoscaling option to spin up more runners when there are pending jobs
> - Amazon's free autoscaling service for github runners
> But in practice, Github makes this hard so incentivize using their runner services. I have an explicit job that cleans up kurtosis post run irrespective of fail/success - so i haven't had any stale file issues there.

IMO we should not run every sim test on the self hosted runner, but instead do:
- Run on Github runners on commit a light version of the sim tests that are okay with low power machines. Sim tests have shown to be very stable with correct business logic
- Run on self hosted runners **only on demand** for key commits on key PRs + on merge to master + before release. This less frequent tests could do much longer runs with more nodes.

This approach makes sense because the workflow must be different:
- Requires more setup phase
- Must assume another test is running in the background, so ports, paths must not conflict
- Requires a tear-down phase

dapplion:
- Closing for https://github.com/ChainSafe/lodestar/issues/4758

Faithtosin:
There's another option, running sim tests on [AWS Codebuild runners](https://github.com/aws-actions/aws-codebuild-run-build#:~:text=CodeBuild%20compute%20types,GPU%20hardware%20devices).  I have implemented AWS Codebuild runners for Github action in the past just to solve this issue of having more resources for Jobs. 

With AWS Codebuild we get clean environments every time, logs streamed straight to Github actions and machines:
up to 72 x86_64 vCPUs
up to 255 GB RAM
up to 8 ARM64 vCPUs
GPU hardware devices

With this setup, we'll only use Github managed runners to stream the logs and orchestrate the process, Codebuild containers will do the heavy lifting.

dapplion:
@Faithtosin The thing is we don't have evidence yet that the current simulation tests fail due to resource exhaustion. From what I've seen it's business logic errors.

nitindudeja2107 (commented):
#4788 #4812 #4798 #4811

nitindudeja2107 (commented):
#4788 #4756 #4812

PR #4765, opened 2022-11-14
WIP Add ethers provider lightclient class

dapplion:
**Motivation**

Basic ground-work for integrating our light-client as an ethers provider, such that's a drop-in replacement for web3 dapps and wallets.

**Description**

- `LightclientRestProvider` extends `ethers.JsonRpcProvider`
- `LightclientRestProvider` intercepts some calls and replaces them with `eth_getProof`
- Uses `RootResolver.resolveQuantityOrTag()` to get a state root from a block number or block tag
- `RootResolver` uses the light-client protocol to keep an internal cache of heads to resolve

CC: @g11tech

### DIFF
--- a/packages/light-client/package.json
+++ b/packages/light-client/package.json
@@ -63,12 +63,16 @@
     "@chainsafe/bls": "7.1.1",
     "@chainsafe/persistent-merkle-tree": "^0.4.2",
     "@chainsafe/ssz": "^0.9.2",
+    "@ethereumjs/statemanager": "^1.0.1",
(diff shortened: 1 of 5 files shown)


dapplion:
@nazarhussain Could you add a snippet of how the same concept would be implemented for web3.js 1.x and 4.x?

nazarhussain:
The `@lodestar/prover` package can handle the following use case. For details please check the [README](https://github.com/ChainSafe/lodestar/blob/unstable/packages/prover/README.md)

PR #4770, opened 2022-11-16
Extract out transport layer for light client and include ability to drive execution layer

dadepo:
**Description**

This PR is currently in draft. The purpose is to showcase the direction of changes based on our last meeting. It contains the following changes:

- Extract a transport layer with a Rest implementation. The Gossip implementation is already being worked on in the branch [here](https://github.com/ChainSafe/lodestar/tree/dadepo/add_gossip_transport) but can't be completed until the work to extract the ResReq and Gossip is done
- Added ability to drive an EL.

With the changes in this PR, we will have a light client implementation that will be able to drive an EL

What I would like to do before updating the PR to ready for review

- [x] Extract the code to interact with EL so it can be re-used by beacon-node and light client. Right now I only copied part of the necessary code over. PS: if there is any alternative approach other than this, let me know
- [ ] Test on goerli/sepolia that the light client can indeed drive an EL successfully
- [x] Revisit the logic to parse command line arguments for light client 
- [ ] Address implementation TODOs


packages/beacon-node/src/network/gossip/validation/queue.ts
```diff
@@ -1,6 +1,6 @@
 import {mapValues} from "@lodestar/utils";
+import {JobItemQueue, JobQueueOpts, QueueType} from "@lodestar/engine-api-client";
```
wemeetagain, line 2: Queue should probably be moved to @lodestar/utils if possible

packages/beacon-node/src/node/nodejs.ts
```diff
@@ -9,14 +9,14 @@ import {Api} from "@lodestar/api";
 import {BeaconStateAllForks} from "@lodestar/state-transition";
 import {ProcessShutdownCallback} from "@lodestar/validator";
 
+import {initializeExecutionBuilder, initializeExecutionEngine} from "@lodestar/engine-api-client";
```
wemeetagain, line 12: You might break this PR up into two parts.
Just move the queue to utils and move execution to this new package.
We can get these merged more quickly if they are just moving some directories around.

Also agree with @g11tech 's naming suggestion re the new package: `@lodestar/execution`

dadepo:
closing this Pr. It will be split into two as suggested

Issue #4769, opened 2022-11-15
API endpoint for block proofs


wemeetagain:
**Is your feature request related to a problem? Please describe.**

We have an endpoint for state proofs, we don't have one for block proofs

**Describe the solution you'd like**

Add an endpoint similar to the state proof endpoint, but for block proofs

wemeetagain:
Already implemented

`/eth/v0/beacon/proof/block/{block_id}?format={hex_proof_descriptor}`

PR #4771, opened 2022-11-16
Move proof endpoints to new namespace

dapplion:
**Motivation**

After discusion with the community it became clear to separate lightclient consensus endpoints to lightclient data provider endpoints.

- Merged routes into spec apply this distinction, dropping the getStateProof endpoint https://github.com/ethereum/beacon-APIs/pull/247

We should keep exposing this endpoint but makes sense to offer it in a new namespace since it's probable that this will be satisfied by different entities.

**Description**

Move proofs endpoints to new namespace

- Related https://github.com/ChainSafe/lodestar/issues/4769


twoeths:
some unit tests failed

wemeetagain (approved):

PR #4774, opened 2022-11-16
Support EIP-4844 - WIP

dapplion:
**Motivation**

Full implementation of the EIP-4844 hard-fork logic. EIP-4844 rationale -> [B L O B S P A C E](https://twitter.com/protolambda/status/1584105020697432064)

**Description**

Implementation list:
- [x] Add params, config and types https://github.com/ChainSafe/lodestar/pull/4568
- [x] Make `c-kzg` library available for Lodestar. Add native bindings for `ethereum/c-kzg-eip4844` https://github.com/dankrad/c-kzg/pull/3
- [x] Block proposing
  - Add new engine API `engine_getBlobsV1`
  - Cache returned blobs in beacon to not modify validator cycle code
  - Publish to gossip topic, depending on fork
- [x] Block processing / import
  - Update chain `processBlock` to take block + (maybe) blobs pair
  - Validate blobs `is_data_available`, running KZG crypto
  - Store blobs to database + migrate from hot to cold on finalization + prune old blobs < MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS
- [x] Add p2p ResResp methods `BeaconBlockAndBlobsSidecarByRoot` + `BlobsSidecarsByRange`
  - Update register / unregister logic to active those topics only post-eip4844
- [x] Add p2p gossip topic `beacon_block_and_blobs_sidecar`

Test list:
- [ ] Integrate with interop repo https://github.com/Inphi/eip4844-interop/pull/66

PR #4778, opened 2022-11-17
Add eip-4844 types and config

dapplion:
**Motivation**

- Part of https://github.com/ChainSafe/lodestar/pull/4774

**Description**

- Update config for eip-4844
- Move fork scheduling from config preset to network config. Else testnets can get forks scheduled if we don't overwrite all _FORK_EPOCH variables. Added a test to ensure that never happens
- Add params for eip-4844
- Add types for eip-4844


g11tech (approved):
lgtm!

PR #4779, opened 2022-11-17
Use node 18

### LINKED ISSUE #4768: Upgrade to node 18
According to the [nodejs release schedule](https://github.com/nodejs/release#release-schedule), node 16 went into maintenance mode 2022-10-18.
The new active LTS is node 18.

We should upgrade all infrastructure, docs, guides, etc. to node 18.

Checklist for completing upgrade:
- [ ] CI/sim tests
- [ ] Docker builds 
- [ ] Individual dev environments
- [ ] Documentation upgrades
- [ ] Upgrade test/production fleet of servers

wemeetagain:
**Motivation**

Resolve #4768 

**Description**

Minimal diff to test/build/recommend using node 18


twoeths:
not sure why `computeProposerBoostScoreFromBalances` benchmark test is 3x slower than NodeJS 16

dapplion:
> not sure why `computeProposerBoostScoreFromBalances` benchmark test is 3x slower than NodeJS 16

Can you retry multiple times and double check in local to make sure it's not a false positive?

wemeetagain:
> > not sure why `computeProposerBoostScoreFromBalances` benchmark test is 3x slower than NodeJS 16
> 
> Can you retry multiple times and double check in local to make sure it's not a false positive?

confirmed locally and also in CI multiple times.

wemeetagain:
also e2e tests consistently timing out in a few places :eyes:

wemeetagain:
tests fixed, ready for re-review

packages/cli/test/e2e/runDevCmd.test.ts
```diff
@@ -9,11 +9,12 @@ describeCliTest("Run dev command", function ({spawnCli}) {
   itDone("Run dev command with no --dataDir until beacon api is listening", async function (done) {
     const beaconPort = 39011;
 
-    const devProc = spawnCli({pipeStdToParent: false, printOnlyOnError: true, logPrefix: "dev"}, [
+    const devProc = spawnCli({pipeStdToParent: true, printOnlyOnError: true, logPrefix: "dev"}, [
```
dapplion: Warning we do not want to pollute test output. If you need that enable with a temporary flag on ENV variable. Or upload output in a test artifact

wemeetagain: whoops, stray diff from debugging these tests

wemeetagain: should be removed in latest commit of this branch

twoeths:
- Mesh peers and gossip peer scores are great (while rpc count is the same)

Screen Shot 2022-12-09 at 10 09 38

- EventLoop lag is better than Nodejs 16

Screen Shot 2022-12-09 at 10 10 29

- That's confirmed by attestation produce/publish delay

Screen Shot 2022-12-09 at 10 11 09

However:

- Heap memory (and gc) is not stable so we need to monitor it for a long time

Screen Shot 2022-12-09 at 10 12 37

- And some benchmark tests failed, especially `Buffer.concat()` is >2000x slower is a concern

dapplion:
Where are all the places we use Buffer.concat in Lodestar's regular runtime?

twoeths (approved):
Heap memory is an issue of unstable, not this branch. As discussed, we'd merge this anyway as the metrics look great. Some more things we need to do:
- Create 2 issues for the 2 failed performance tests
  - Find some alternatives
  - Look for same usages in our dependencies
- Open a performance issue of `Buffer.concat` for NodeJS team, really hope they'd tackle this soon

wemeetagain:
> Where are all the places we use Buffer.concat in Lodestar's regular runtime?

Probably the most run:
https://github.com/ChainSafe/lodestar/blob/798ec16d301e9db20f94cd818205277678d6f213/packages/beacon-node/src/network/gossip/encoding.ts#L62

Others:
https://github.com/ChainSafe/lodestar/blob/e5dabac124933bcadc650d19d6b128dcbfcb6c43/packages/beacon-node/src/util/graffiti.ts#L7

A few in these files:
https://github.com/ChainSafe/lodestar/blob/eccde2f3d71448ad5b5026f87e4699c29b5b04f7/packages/state-transition/src/util/seed.ts#L34
https://github.com/ChainSafe/lodestar/blob/2883368c84efc23a679cafc4c98d430f45444e51/packages/utils/src/verifyMerkleBranch.ts#L4

wemeetagain:
opened issue upstream in node repo:
- https://github.com/nodejs/node/issues/45802

dapplion:
Ok to me to defer issue for latter and unblock v18

PR #4788, opened 2022-11-21
Default afterBlockDelaySlotFraction to 0

### LINKED ISSUE #4600: Missed attestation after v1.1.0
**Describe the bug**

With v1.1.0 we have ~2% wrong head attestations and ~1% missed attestation

Screen Shot 2022-09-27 at 14 48 44

This is because in v1.1.0 we improved gossipsub hence we have a lot of mesh peers, and that introduced the I/O lag issue
```
Local validator published unaggregated attestation validatorIndex=429161, slot=4786296, committeeIndex=5, subnet=5, sentPeers=17, delaySec=2.5950000286102295
```

at the validator side
```
Sep-27 06:19:39.000[]                ^[[34mdebug^[[39m: HttpClient request routeId=submitPoolAttestations
Sep-27 06:19:43.733[]                ^[[34mdebug^[[39m: HttpClient response routeId=submitPoolAttestations
```
Screen Shot 2022-09-27 at 14 55 05

**Expected behavior**
- Should improve `vc REST API request times` `submitPoolAttestations` and `beacon REST API response times` `submitPoolAttestations`

twoeths:
**Motivation**

It's been a while since we test the `afterBlockDelaySlotFraction=0` on mainnet nodes, the result is good so we should apply that by default to benefit everybody.

**Description**

- By default set `afterBlockDelaySlotFraction` as 0 instead of 1/6 slot

Closes #4600

### DIFF
--- a/packages/validator/src/services/attestation.ts
+++ b/packages/validator/src/services/attestation.ts
@@ -135,9 +135,11 @@ export class AttestationService {
     // never beyond the 1/3 cutoff time.
     // https://github.com/status-im/nimbus-eth2/blob/7b64c1dce4392731a4a59ee3a36caef2e0a8357a/beacon_chain/validators/validator_duties.nim#L1123
     const msToOneThirdSlot = this.clock.msToSlot(slot + 1 / 3);
-    // Default = 1/6, which is half of attestation offset
-    const afterBlockDelayMs = 1000 * this.clock.secondsPerSlot * (this.opts?.afterBlockDelaySlotFraction ?? 1 / 6);
-    await sleep(Math.min(msToOneThirdSlot, afterBlockDelayMs));
+    // submitting attestations asap to avoid busy time at around 1/3 of slot
+    const afterBlockDelayMs = 1000 * this.clock.secondsPerSlot * (this.opts?.afterBlockDelaySlotFraction ?? 0);
+    if (afterBlockDelayMs > 0) {
+      await sleep(Math.min(msToOneThirdSlot, afterBlockDelayMs));
+    }
 
     this.metrics?.attesterStepCallPublishAttestation.observe(this.clock.secFromSlot(slot + 1 / 3));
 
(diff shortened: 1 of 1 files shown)


packages/validator/src/services/attestation.ts
```diff
@@ -135,9 +135,11 @@ export class AttestationService {
     // never beyond the 1/3 cutoff time.
     // https://github.com/status-im/nimbus-eth2/blob/7b64c1dce4392731a4a59ee3a36caef2e0a8357a/beacon_chain/validators/validator_duties.nim#L1123
     const msToOneThirdSlot = this.clock.msToSlot(slot + 1 / 3);
-    // Default = 1/6, which is half of attestation offset
-    const afterBlockDelayMs = 1000 * this.clock.secondsPerSlot * (this.opts?.afterBlockDelaySlotFraction ?? 1 / 6);
-    await sleep(Math.min(msToOneThirdSlot, afterBlockDelayMs));
+    // submitting attestations asap to avoid busy time at around 1/3 of slot
+    const afterBlockDelayMs = 1000 * this.clock.secondsPerSlot * (this.opts?.afterBlockDelaySlotFraction ?? 0);
```
dapplion: Since this is delicate value with a lot of background, please add a constant at the top of the file with context
```ts
/**
 * Explain the history of the value, why it was 1/6, why it's now 0. Link to relevant issues for the constant
 */
const DEFAULT_AFTER_BLOCK_DELAY_SLOT_FRACTION = 0
```

packages/validator/src/services/attestation.ts
```diff
@@ -135,9 +135,11 @@ export class AttestationService {
...
+    if (afterBlockDelayMs > 0) {
```
dapplion: This is if statement is not necessary, sleep() implementation already does this

wemeetagain (approved):

PR #4792, opened 2022-11-21
Optimize capella worst case block processing

dapplion:
**Motivation**

Worst case for capella block processing is very slow = read all validators and all balances from the state's tree.

**Description**

Add caches to prevent having to read validator's tree unless there's a withdrawal

TODO:
- Ensure all assumptions made with this caches are correct before merging
- Ensure worst case speed up is worth it


dapplion:
Not necessary with latest spec.

- Max swipe per block is limited
- Swipe is pretty fast with latest unstable benchmarks

PR #4797, opened 2022-11-22
ReqResp package tests

nazarhussain:
**Motivation**

Improve the test coverage. 

**Description**

Add unit tests for the reqresp package. 

Run all unit tests.


packages/reqresp/src/types.ts
```diff
@@ -97,4 +97,5 @@ export interface TypeSerializer<T> {
   deserialize(bytes: Uint8Array): T;
   maxSize: number;
   minSize: number;
+  equals(a: T, b: T): boolean;
```
dapplion: Why do you need equals?
- If you are using this for tests only do not add unnecessary stuff on src code
- If you are using this for test assertion, it's really bad to debug. Always compare actual data, assert serialized == serialized or json == json, else the failed assertion will be "expected true, got false" which does not give any data point to debug

packages/reqresp/test/fixtures/messages.ts
```diff
@@ -0,0 +1,168 @@
...
+  chunks: ["0x08", "0xff060000734e61507059010c00000175de410100000000000000"].map(
+    (s) => new Uint8ArrayList(fromHexString(s))
+  ),
+  asyncChunks: [
+    "0x08", // length prefix
+    "0xff060000734e61507059", // snappy frames header
+    "0x010c00000175de410100000000000000", // snappy frames content
+  ]
+    .map(fromHexString)
+    .map(Buffer.from),
```
dapplion: I know it's a test, but let's not accept that doing two iterations instead of one is acceptable in Lodestar's codebase :pray:

packages/reqresp/test/utils/response.ts
```diff
@@ -0,0 +1,33 @@
...
+import {allForks} from "@lodestar/types";
+import {responseEncodeError, responseEncodeSuccess} from "../../src/encoders/responseEncode.js";
+import {RespStatus} from "../../src/interface.js";
+import {EncodedPayload, EncodedPayloadType, ProtocolDefinition} from "../../src/types.js";
+import {ResponseChunk} from "../fixtures/encoders.js";
+import {beaconConfig} from "../fixtures/messages.js";
+import {blocksToReqRespBlockResponses} from "../utils/block.js";
+import {arrToSource} from "../utils/index.js";
+
+export async function* responseEncode(
```
dapplion, line 8: What is the purpose of this function?

packages/reqresp/test/mocks/logger.ts
```diff
@@ -0,0 +1,11 @@
+import sinon from "sinon";
+import {ILogger} from "@lodestar/utils";
+
+export const createStubbedLogger = (): ILogger => ({
```
dapplion, line 4: Why is this thing needed?

nazarhussain: To reuse the stub where-ever logger required.

packages/reqresp/test/utils/response.ts
nazarhussain: @dapplion  To be used in the tests for the `test/unit/encoders/responseEncode.test.ts`. Particularly used to same assertions data as it exists earlier.

packages/reqresp/test/mocks/logger.ts
dapplion: I have some PTSD with unnecessary usage of Sinon https://github.com/ChainSafe/lodestar/issues/1523 Please avoid sinon unless you actually want to assert calls, even then there always better patterns to achieve the same test guarantees

In this specific case you can do

```ts
export const loggerNoop = {
  info: () => {}
  ...
}
```

dapplion: Or no need to re-invent the wheel, and just re-use testLogger which has been incredibly helpful to me to debug tests

packages/reqresp/src/types.ts
nazarhussain: It was added to re-use existing helper `isEqualSszType`. This equality builtin to `ssz` and I believe it's makes sense to use when we are not sure about what SSZ types are used.

packages/reqresp/test/utils/response.ts
```diff
@@ -0,0 +1,33 @@
...
+): AsyncIterable<Buffer> {
+  for (const chunk of responseChunks) {
+    if (chunk.status === RespStatus.SUCCESS) {
+      if (chunk.payload.type === EncodedPayloadType.bytes) {
+        return [chunk.payload.bytes];
+      }
+
+      const lodestarResponseBodies = protocol.method.startsWith("beacon_blocks")
+        ? blocksToReqRespBlockResponses(([chunk.payload.data] as unknown) as allForks.SignedBeaconBlock[], beaconConfig)
+        : [chunk.payload];
```
dapplion: This lines look really nasty the fact that this burried utility uses `beacon_blocks` prefix to switches logic. The `EncodedPayload` should be passed as an argument to this function, and decided by the test caller if use bytes or ssz.

packages/reqresp/test/utils/response.ts
dapplion: This looks like src/ code that somehow ended in tests, given that `responseEncodeSuccess` and `responseEncodeError` are used in prod. But this function is really awkward and won't be used in production. It encodes this obscure type "ResponseChunk" that only exists in tests

packages/reqresp/test/utils/response.ts
nazarhussain: I can refactor that part to better approach, but for the PR scope it was intended to reuse existing tests, so why this ported as it is.

packages/reqresp/src/types.ts
dapplion: If `isEqualSszType` is used only on tests, please do not add here

packages/reqresp/test/utils/response.ts
dapplion: oh okay

packages/reqresp/test/utils/response.ts
nazarhussain: Simplified the `responseEncode` method. Its a utility function as many others for tests only, so definitely would not used in production.

packages/reqresp/test/mocks/logger.ts
nazarhussain: Related to this observation I commented in detail https://github.com/ChainSafe/lodestar/issues/1523#issuecomment-1345676484

dapplion (approved):

PR #4798, opened 2022-11-22
Extract engine api to a separate package

dadepo:
**Motivation**

Both the beacon node and light client would have to communicate with an execution layer. Hence a need to have the code that does it extracted out of the beacon-node to enable re-use.


wemeetagain:
can you move the JsonHttpClient to utils?

packages/engine-api-client/package.json
```diff
@@ -0,0 +1,60 @@
...
+  "type": "module",
+  "author": "ChainSafe Systems",
+  "homepage": "https://github.com/ChainSafe/lodestar#readme",
+  "license": "Apache-2.0",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/ChainSafe/lodestar.git"
+  },
+  "exports": "./lib/index.js",
+  "types": "lib/index.d.ts",
```
wemeetagain, line 14: check the other package.jsons, need a `typesVersions` here, just copy/paste from the other package.jsons.

wemeetagain: This should fix that type issue where you're needing `/src` in the imports

twoeths (commented):
lgtm overall!

packages/engine-api-client/README.md
```diff
@@ -0,0 +1,17 @@
...
+[![Eth Consensus Spec v1.1.10](https://img.shields.io/badge/ETH%20consensus--spec-1.1.10-blue)](https://github.com/ethereum/consensus-specs/releases/tag/v1.1.10)
+![ES Version](https://img.shields.io/badge/ES-2020-yellow)
+![Node Version](https://img.shields.io/badge/node-16.x-green)
+
+> This package is part of [ChainSafe's Lodestar](https://lodestar.chainsafe.io) project
+
+Lodestar client for [Engine JSON-RPC API](https://github.com/ethereum/execution-apis/tree/main/src/engine)
+
+> **Note**
+> No version of this package exist prior to version 1.2.1.
```
dapplion, line 13: Add usage example. That should make it clear how hard / easy the API is to consume

packages/engine-api-client/src/engine/http.ts
```diff
@@ -1,21 +1,39 @@
...
+ * blocks per batch are requested _at most_. A batch may request less blocks to account for
+ * already requested slots. There is a timeout for each batch request. If this value is too high,
+ * we will negatively report peers with poor bandwidth. This can be set arbitrarily high, in which
+ * case the responder will fill the response up to the max request size, assuming they have the
+ * bandwidth to do so.
+ *
+ * Jul2022: Current batch block processor wants only blocks in the same epoch. So we'll process only
+ * one batch at a time. Metrics can confirm preliminary tests that speed is as good.
+ */
+export const EPOCHS_PER_BATCH = 1;
```
dapplion: This constant should be in the beacon node package, and this class must expose a queueMaxLength option. It can default to a sane value similar to this too.

packages/engine-api-client/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./engine/index.js";
+export * from "./engine/interface.js";
```
dapplion: Do not nest contents, just dump engine/ contents into src/

packages/engine-api-client/src/metrics/index.ts
```diff
@@ -0,0 +1,23 @@
+import {IGauge, IHistogram} from "@lodestar/utils/metrics";
```
dapplion: Move packages/engine-api-client/src/metrics/index.ts to packages/engine-api-client/src/metrics.ts

packages/engine-api-client/src/metrics/index.ts
```diff
@@ -0,0 +1,23 @@
...
+  engineHttpProcessorQueue: IQueueMetrics;
+};
+
+export type JsonRpcHttpClientMetrics = {
+  requestTime: IHistogram<"routeId">;
+  requestErrors: IGauge<"routeId">;
+  requestUsedFallbackUrl: IGauge<"routeId">;
+  activeRequests: IGauge<"routeId">;
+  configUrlsCount: IGauge;
+  retryCount: IGauge<"routeId">;
```
dapplion: If the JsonRpcHttpClient class is in utils, this should be too

packages/reqresp/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/ChainSafe/lodestar/issues"
   },
-  "version": "1.2.1",
+  "version": "0.1.0",
```
dapplion: :warning: watch out, dangerous unwanted diff!

packages/engine-api-client/package.json
dadepo: > This should fix that type issue where you're needing `/src` in the imports

the issue ended up being something else...The`typesVersion` has also been added as suggested

packages/utils/src/constants/constants.ts
```diff
@@ -0,0 +1,16 @@
...
+export const ZERO_HASH_HEX = "0x" + "00".repeat(32);
+export const EMPTY_SIGNATURE = Buffer.alloc(96, 0);
+export const GRAFFITI_SIZE = 32;
+export const MAX_VARINT_BYTES = 10;
+
+/**
+ * The maximum milliseconds of clock disparity assumed between honest nodes.
+ */
+export const MAXIMUM_GOSSIP_CLOCK_DISPARITY = 500;
+export const MAXIMUM_GOSSIP_CLOCK_DISPARITY_SEC = MAXIMUM_GOSSIP_CLOCK_DISPARITY / 1000;
```
dapplion: Why do all this constants need to be here? Please keep the PR diff to the min possible. Best to open issues if you have tangential improvement proposals

packages/utils/src/constants/network.ts
```diff
@@ -0,0 +1,91 @@
...
+  // Lighthouse-defined codes
+  129: "Client has too many peers",
+  250: "Peer score too low",
+  251: "Peer banned this node",
+};
+
+/** Until js-libp2p types its events */
+export enum Libp2pEvent {
+  peerConnect = "peer:connect",
+  peerDisconnect = "peer:disconnect",
```
dapplion: By no means the utils package has to be aware of all these deep implementation details of the beacon-node. Please remove and keep the PR diff minimal

packages/utils/package.json
```diff
@@ -36,15 +62,17 @@
...
   "dependencies": {
     "@chainsafe/as-sha256": "^0.3.1",
     "any-signal": "2.1.2",
     "bigint-buffer": "^1.1.5",
     "case": "^1.6.3",
     "chalk": "^4.1.2",
     "js-yaml": "^3.13.1",
-    "winston": "^3.3.3"
+    "winston": "^3.3.3",
+    "prom-client": "^14.1.0",
```
dapplion: prom-client is heavy dependency assumed to be run in NodeJS, why is this dependency necessary? Please remove

packages/utils/package.json
```diff
@@ -36,15 +62,17 @@
...
+    "cross-fetch": "^3.1.4",
+    "jwt-simple": "0.5.6"
```
dapplion: jwt is node only dependency, please review the source of the dependency before including in utils. Utils must be consumable from browser code, so this line will break

https://github.com/hokaccha/node-jwt-simple/blob/a812607fa498e147a0816cbbbbf386df60d9fd6c/lib/jwt.js#L13

packages/utils/src/metrics/index.ts
```diff
@@ -0,0 +1,12 @@
+import {Gauge, Histogram} from "prom-client";
```
dapplion: Do not import prom-client, define a look-a-like type like here 
https://github.com/ChainSafe/lodestar/blob/unstable/packages/reqresp/src/metrics.ts

dapplion (changes requested):

packages/reqresp/package.json
dadepo: spotted it earlier and updated, just did not push yet.

dapplion:
@dadepo There's a 6,000 line diff in the yarn.lock, this should not happen please review. You should only clean the dependency added here 
https://github.com/ChainSafe/lodestar/pull/4798/commits/f6f401d6e09d5148bfb9cfe9f499690dddaee235#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR2325

dadepo:
> can you move the JsonHttpClient to utils?

@wemeetagain Lion spotted [here](https://github.com/ChainSafe/lodestar/pull/4798#discussion_r1030303789) that jwt-token can't be used in the browser...and JsonHttpClient depends on it. Since utils package is supposed to be agnostic to both browser and node, I will have to move `JsonRpcHttpClient` away from the `utils` package. Unless you have other suggestions to go about this?

packages/utils/src/constants/constants.ts
dadepo: The constants were moved in anticipation that they will be shared when the light-client is updated. But I agree it is premature in this PR. When working on the pr for the light-client, if they are actually needed then I will move it then and it will be obvious why.

dadepo:
> @dadepo There's a 6,000 line diff in the yarn.lock, this should not happen please review. You should only clean the dependency added here [f6f401d#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR2325](https://github.com/ChainSafe/lodestar/commit/f6f401d6e09d5148bfb9cfe9f499690dddaee235#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR2325)

when making changes that involves changing package structures, I often delete node_modules folder and yarn.lock to install everything afresh to be sure I am not getting false positives, where things continue to work because of already resolved, and cached dependencies. This is why the yarn.lock end up with the diffs.

How will you go about this instead in other to be assured no cached files are leading to false positives?

packages/utils/src/metrics/index.ts
```diff
@@ -1,12 +1,57 @@
...
 
-export type IAvgMinMax = {
-  addGetValuesFn(getValuesFn: () => number[]): void;
-  set(values: number[]): void;
+export type IMetrics = {
+  executionEnginerHttpClient: JsonRpcHttpClientMetrics;
+  engineHttpProcessorQueue: IQueueMetrics;
 };
+
+export type JsonRpcHttpClientMetrics = {
```
dapplion: `packages/utils/src/metrics/index.ts` should not know about `JsonRpcHttpClientMetrics`, please abstract to their respective scopes

packages/utils/src/metrics/index.ts
```diff
@@ -1,12 +1,57 @@
...
+  requestTime: IHistogram<"routeId">;
+  requestErrors: IGauge<"routeId">;
+  requestUsedFallbackUrl: IGauge<"routeId">;
+  activeRequests: IGauge<"routeId">;
+  configUrlsCount: IGauge;
+  retryCount: IGauge<"routeId">;
+};
+
+export interface IQueueMetrics {
```
dapplion, line 37: Same as above comment

packages/utils/src/provider/jsonRpcHttpClient.ts
```diff
@@ -43,15 +43,6 @@ export type ReqOpts = {
   shouldRetry?: (lastError: Error) => boolean;
 };
 
-export type JsonRpcHttpClientMetrics = {
```
dapplion: Why remove from here?

packages/beacon-node/src/eth1/eth1DepositDataTracker.ts
```diff
@@ -3,6 +3,8 @@ import {IChainForkConfig} from "@lodestar/config";
 import {BeaconStateAllForks, becomesNewEth1Data} from "@lodestar/state-transition";
 import {ErrorAborted, TimeoutError, fromHex, ILogger, isErrorAborted, sleep} from "@lodestar/utils";
 
+import {HttpRpcError, isJsonRpcTruncatedError} from "@lodestar/engine-api-client";
```
dapplion, line 5: Unnecessary newline above this import

packages/beacon-node/test/sim/merge-interop.test.ts
```diff
@@ -108,7 +106,7 @@ describe("executionEngine / ExecutionEngineHttp", function () {
 
     //const controller = new AbortController();
     const executionEngine = new ExecutionEngineHttp(
-      {urls: [engineRpcUrl], jwtSecretHex, retryAttempts, retryDelay},
+      {urls: [engineRpcUrl], jwtSecretHex, retryAttempts, retryDelay, queueMaxLength},
```
dapplion: If it's passing the default, why does it have to be specified?

packages/engine-api-client/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+## Usage
+
+```
```
dapplion, line 17: Color as typescript with `ts` tag

packages/engine-api-client/README.md
```diff
@@ -0,0 +1,68 @@
...
+yarn add @lodestar/engine-api-client
+```
+
+Provides method for interacting with the Engine API of the execution layer client as defined in [Engine JSON-RPC API](https://github.com/ethereum/execution-apis/tree/main/src/engine)
+
+```
+ const baseUrl = http://localhost:8551 // running execution layer client
+ const controller = new AbortController()
+ const jwtSecretHex = "" // "jwt secret shared with execution layer client"
```
dapplion: Set to a random token of expected format, not empty string. Keep the comment

packages/engine-api-client/README.md
```diff
@@ -0,0 +1,68 @@
...
+   {signal: controller.signal}
+ );
+ 
+ // Prepare a payload
+ const payloadId = await executionEngine.notifyForkchoiceUpdate(
+  genesisBlockHash,
+  safeBlockHash,
+  finalizedBlockHash,
+  {
+      timestamp: quantityToNum("0x..."),
```
dapplion: User should never have to use this data transformers. This are meant to interact directly with execution layer data which is exactly what this package is abstracting

packages/engine-api-client/src/provider/metrics.ts
```diff
@@ -0,0 +1,15 @@
+import {IGauge, IHistogram, IQueueMetrics} from "@lodestar/utils/metrics";
+
+export type IMetrics = {
+  executionEnginerHttpClient: JsonRpcHttpClientMetrics;
+  engineHttpProcessorQueue: IQueueMetrics;
+};
+
+export type JsonRpcHttpClientMetrics = {
```
dapplion: I recall seeing this twice in the PR, can you double check it's not duplicated?

dapplion (changes requested):

packages/utils/src/metrics/index.ts
dadepo: not sure why the diff is still showing `JsonRpcHttpClientMetrics` is present in this file. It is not as can be seen if the file is viewed directly on this branch https://github.com/ChainSafe/lodestar/blob/dadepo/extract_out_engine_api/packages/utils/src/metrics/index.ts

packages/utils/src/metrics/index.ts
dadepo: @dapplion `IQueueMetrics` is used by `JobItemQueue` (see [here](https://github.com/ChainSafe/lodestar/blob/8634572c177ba74d271bc71e1c5a4bb0abc73e71/packages/utils/src/queue/itemQueue.ts#L23)) which is used in multiple places in `beacon-node` package. The same  `JobItemQueue` is also used in the `engine-api-client` (see [here](https://github.com/ChainSafe/lodestar/blob/8634572c177ba74d271bc71e1c5a4bb0abc73e71/packages/engine-api-client/src/http.ts#L106))

Hence the queue functionality (which makes use of `IQueueMetrics`) is reused across the `beacon-node` package and `engine-api-client` package, hence why the queue related modules are in the `utils` package.

Given the above, where else would you suggest to put it?

packages/utils/src/provider/jsonRpcHttpClient.ts
dadepo: @dapplion this got moved to `/packages/engine-api-client/src/provider/metrics.ts` so as to be part of the `engine-api-client` package. Will you have this be somewhere else instead?

packages/engine-api-client/README.md
dadepo: This example was actually copied from the `merge-interop.test.ts` file (see [here](https://github.com/ChainSafe/lodestar/blob/8634572c177ba74d271bc71e1c5a4bb0abc73e71/packages/beacon-node/test/sim/merge-interop.test.ts#L121))

I am not sure why because the API do not actually need those data transformers as the actually value can be set directly. I will update the example to have something like:

```
 // Prepare a payload
 const payloadId = await executionEngine.notifyForkchoiceUpdate(
  ForkName.bellatrix,
  genesisBlockHash,
  safeBlockHash,
  finalizedBlockHash,
  {
      timestamp: 1670268513458, // QUANTITY, integer value as defined in ethereum execution layer JSON RPC https://eth.wiki/json-rpc/API
      prevRandao: new Uint8Array(32), // DATA, 32 Bytes as defined in ethereum execution layer JSON RPC https://eth.wiki/json-rpc/API,
      suggestedFeeRecipient: "0x...",
  }
 );
```

Is that acceptable by you?

packages/engine-api-client/src/provider/metrics.ts
dadepo: I did a grep and confirmed this exist only once. Seems github is skewing the diff and that might be responsible for the duplication you saw.

packages/engine-api-client/README.md
dapplion: > Is that acceptable by you?

It's not what's acceptable to me, but what's the most useful for consumers of this API. The goal of the README example is to show a quick and concise example to get a feel of how this library is used in the real world

dapplion (commented):
Pushed some changes to address review comments, if conflicts are fixed should be good to go

packages/utils/src/hex.ts
```diff
@@ -106,10 +97,10 @@ export function bytesToData(bytes: Uint8Array): DATA {
 /**
  * DATA as defined in ethereum execution layer JSON RPC https://eth.wiki/json-rpc/API
  */
-export function dataToBytes(hex: DATA, fixedLength: number | null): Uint8Array {
+export function dataToBytes(hex: DATA, fixedLength?: number): Uint8Array {
```
dapplion: keep the null type, forces you as consumer to think on the length. Else we'll forget to add it

packages/engine-api-client/test/unit/executionEngine/httpRetry.test.ts
```diff
@@ -45,18 +46,12 @@ describe("ExecutionEngine / http ", () => {
...
-        retryDelay: defaultExecutionEngineHttpOpts.retryDelay,
-      },
-      {signal: controller.signal}
-    );
+    executionEngine = new ExecutionEngineHttp({urls: [baseUrl], retryAttempts}, {signal: controller.signal});
   });
 
   describe("notifyForkchoiceUpdate", function () {
     it("notifyForkchoiceUpdate no retry when no pay load attributes", async function () {
+      this.timeout("10 min");
```
dapplion: Why is the 10 minutes timeout necessary? Sounds very extreme, this is a unit test!

packages/engine-api-client/src/interface.ts
```diff
@@ -72,6 +71,21 @@ export type BlobsBundle = {
   blobs: Blob[];
 };
 
+export interface IExecutionBuilder {
```
dapplion: Why does the engine-api package declare the builder class interface?

dadepo: Part of unintended changes when resolving merge conflicts. Resolved some in past commits. Missed this. It is removed now.

packages/engine-api-client/test/unit/executionEngine/httpRetry.test.ts
dadepo: Not sure why this is showing as a diff now. It is currently same on unstable
https://github.com/ChainSafe/lodestar/blob/70796d3ab8736e15551598fe44029b11b3741091/packages/beacon-node/test/unit/executionEngine/httpRetry.test.ts#L91

I'll change this to 60s

packages/utils/src/hex.ts
dadepo: Reverted as you requested... 
but IMHO, I don't think it makes for a cleaner API having to pass null like this `dataToBytes(data.extraData, null)` as the `null` provides no information and might even be confusing when reading the code.

g11tech (dismissed):
lgtm!

packages/engine-api-client/package.json
```diff
@@ -0,0 +1,71 @@
...
+  },
+  "dependencies": {
+    "@lodestar/utils": "^1.2.1",
+    "@lodestar/types": "^1.2.1",
+    "@lodestar/config": "^1.2.1",
+    "@lodestar/db": "^1.2.1",
+    "@lodestar/state-transition": "^1.2.1",
+    "cross-fetch": "^3.1.4",
+    "jwt-simple": "0.5.6",
+    "c-kzg": "^1.0.7"
```
dapplion: - This will break Lodestar installations, see https://github.com/ChainSafe/lodestar/issues/4886 c-kzg must not be installed (for now) by default.
- It was fixed in https://github.com/ChainSafe/lodestar/pull/4888 but this commits undo the fix

dapplion (changes requested):

packages/engine-api-client/package.json
dadepo: I believe this should now be fixed with this commit https://github.com/ChainSafe/lodestar/pull/4798/commits/46a9b4d5f843f2ae7a376431db114f99ac7621d8 I had to duplicate the `kzg.ts` as it is not possible to move it to the util package as the content is node dependent.

Let me know if this is fine or you rather it be dealt another way.

packages/engine-api-client/package.json
```diff
@@ -61,6 +60,9 @@
     "chai-as-promised": "^7.1.1",
     "chai": "^4.3.6"
   },
+  "peerDependencies": {
+    "c-kzg": "^1.0.7"
```
dapplion, line 64: This rationale of this package is "abstract communication with an execution layer client". Thus it should not be aware even that c-kzg exists, else it's a bad abstraction.
- In this PR https://github.com/ChainSafe/lodestar/pull/4931 I refactor the role of the mock which IMO should not end up in the engine-api-client package as it has nothing to do with the goal stated above

dapplion (changes requested):

nflaig:
> @wemeetagain Lion spotted https://github.com/ChainSafe/lodestar/pull/4798#discussion_r1030303789 that jwt-token can't be used in the browser...and JsonHttpClient depends on it. Since utils package is supposed to be agnostic to both browser and node, I will have to move JsonRpcHttpClient away from the utils package. Unless you have other suggestions to go about this?

We could think about changing the library to [sign JWTs](https://github.com/panva/jose/blob/b73eace68f555789f450e13b1965c9a80468e2a6/docs/classes/jwt_sign.SignJWT.md) to [jose](https://www.npmjs.com/package/jose) which supports all sorts of runtimes and is a really high quality library. The downside is that it has a much bigger code base and supports a lot of features we don't need at the moment.

wemeetagain:
> > @wemeetagain Lion spotted [#4798 (comment)](https://github.com/ChainSafe/lodestar/pull/4798#discussion_r1030303789) that jwt-token can't be used in the browser...and JsonHttpClient depends on it. Since utils package is supposed to be agnostic to both browser and node, I will have to move JsonRpcHttpClient away from the utils package. Unless you have other suggestions to go about this?
> 
> We could think about changing the library to [sign JWTs](https://github.com/panva/jose/blob/b73eace68f555789f450e13b1965c9a80468e2a6/docs/classes/jwt_sign.SignJWT.md) to [jose](https://www.npmjs.com/package/jose) which supports all sorts of runtimes and is a really high quality library. The downside is that it has a much bigger code base and supports a lot of features we don't need at the moment.

Also thought about just writing our own lib that does the minimal functionality. I coded this up somewhere, tho it's maybe lost now. But it ends up being like 40 LoC max.

dapplion:
Closing for now, best to re-open with a fresh take if needed

PR #4803, opened 2022-11-23
Submit and handle blsToExecutionChanges for setting withdrawals

g11tech:
Add submitting and handling blsToExecutionChanges

TODO: 
- [x] figure out the signature generation for withdrawals
- [x] tests
- [x] correct pruning logic


packages/beacon-node/test/unit-mainnet/network/gossip/scoringParameters.test.ts
```diff
@@ -80,7 +80,7 @@ describe("computeGossipPeerScoreParams", function () {
     expect(params.firstMessageDeliveriesWeight).closeTo(1.8407, TOLERANCE);
     expect(params.firstMessageDeliveriesDecay).closeTo(0.99856, TOLERANCE);
     expect(params.firstMessageDeliveriesCap).closeTo(21.73035, TOLERANCE);
-    expect(params.invalidMessageDeliveriesWeight).closeTo(-2150, TOLERANCE);
+    expect(params.invalidMessageDeliveriesWeight).closeTo(-2200, TOLERANCE);
```
g11tech, line 83: @dapplion  I am assuming the weights changed because of addition of new topic, in case that assumption is incorrect, will dig deeper into it

dapplion: CC: @tuyennhv

packages/beacon-node/src/chain/opPools/opPool.ts
```diff
@@ -190,7 +220,17 @@ export class OpPool {
       }
     }
 
-    return [attesterSlashings, proposerSlashings, voluntaryExits];
+    const blsToExecutionChanges: capella.SignedBLSToExecutionChange[] = [];
+    for (const blsToExecutionChange of blsToExecutionChanges.values()) {
+      if (isValidBlsToExecutionChange(state as CachedBeaconStateCapella, blsToExecutionChange, false).valid) {
```
dapplion: Here we want to be extremely careful and potentially re-implement this function again thinking each condition. In the past we have bricked block production due to including bad objects. I would add a new function `isValidBlsToExecutionChangeForBlockProduction` or similar.

Lets think each condition from https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#new-process_bls_to_execution_change

1. `assert address_change.validator_index < len(state.validators)`: If valid before will always be valid in the future, no need to check
2. `assert validator.withdrawal_credentials[:1] == BLS_WITHDRAWAL_PREFIX`: Must be checked again, since it can already be changed by now.
3. `assert validator.withdrawal_credentials[1:] == hash(address_change.from_bls_pubkey)[1:]`: If valid before will always be valid in the future, no need to check
4. `assert bls.Verify(address_change.from_bls_pubkey, signing_root, signed_address_change.signature)`: May become invalid in the future if the fork changes. No need to check the full BLS signature, but check that the fork it was valid, is still the current fork

To prevent having to check this messages over and over they have to be pruned. Conditions 2. and 4. can be checked on finalized state and drop the no longer valid messages

packages/beacon-node/src/chain/opPools/opPool.ts
```diff
@@ -268,6 +314,20 @@ export class OpPool {
...
+
+  /**
+   * Call after finalizing
+   * Prune blsToExecutionChanges for validators which are exited in the finalized epoch.
+   */
+  private pruneBlsToExecutionChanges(headState: BeaconStateAllForks): void {
+    const finalizedEpoch = headState.finalizedCheckpoint.epoch;
+    for (const [key, blsToExecutionChange] of this.blsToExecutionChanges.entries()) {
+      const index = blsToExecutionChange.message.validatorIndex;
+      if (headState.validators.getReadonly(index).exitEpoch <= finalizedEpoch) {
```
dapplion: Wrong condition, must check that withdrawal credentials are already changed

packages/beacon-node/src/network/network.ts
```diff
@@ -392,6 +392,11 @@ export class Network implements INetwork {
       this.gossip.subscribeTopic({type: GossipType.light_client_finality_update, fork});
     }
 
+    //for fork capella or greater ?? bls_to_execution_change introduced in capella fork. convert to forkseq > forkseq.capella
```
dapplion: Add space after `//`, and yes use ForkSeq[fork]

packages/beacon-node/test/utils/blsToExecutionChange.ts
```diff
@@ -0,0 +1,16 @@
+import {capella} from "@lodestar/types";
+
+export function generateEmptyBlsToExecutionChange(): capella.BLSToExecutionChange {
```
dapplion: Please do not add this functions, just use `ssz.capella.BlsToExecutionChange.defaultValue()`

packages/db/src/schema.ts
```diff
@@ -69,6 +69,7 @@ export enum Bucket {
   validator_metaData = 41,
 
   backfilled_ranges = 42, // Backfilled From to To, inclusive of both From, To
+  capella_blsToExecutionChange = 56, // ValidatorIndex -> SignedBLSToExecutionChange
```
dapplion: Why 56? We should have some guidelines for picking a bucket number. For now just take the first available, which appears to be 29

packages/state-transition/src/block/processBlsToExecutionChange.ts
```diff
@@ -10,11 +10,16 @@ export function processBlsToExecutionChange(
...
 ): void {
-  assertValidBlsToExecutionChange(state, signedBlsToExecutionChange, true);
-
   const addressChange = signedBlsToExecutionChange.message;
-  const validator = state.validators.get(addressChange.validatorIndex);
 
+  const validation = isValidBlsToExecutionChange(state, signedBlsToExecutionChange, true);
+  if (!validation.valid) {
+    throw (
+      validation.error ?? Error(`Invalid signedBlsToExecutionChange validatorIndex=${addressChange.validatorIndex}`)
```
dapplion: Type the return as `{valid: true} | {valid: false, error: Error}` so you don't need to default on a random error here.

packages/state-transition/src/block/processBlsToExecutionChange.ts
```diff
@@ -23,39 +28,46 @@ export function processBlsToExecutionChange(
...
-): void {
+): {valid: boolean; error?: Error} {
   const addressChange = signedBLSToExecutionChange.message;
 
   if (addressChange.validatorIndex >= state.validators.length) {
-    throw Error(
+    const error = Error(
       `withdrawalValidatorIndex ${addressChange.validatorIndex} > state.validators len ${state.validators.length}`
     );
+    return {valid: false, error};
```
dapplion: Just do everywhere, no need to declare error variable
```suggestion
    return {valid: false, error: Error(...)};
```

packages/state-transition/src/block/processBlsToExecutionChange.ts
```diff
@@ -23,39 +28,46 @@ export function processBlsToExecutionChange(
...
   }
 
   if (verifySignature && !verifyBlsToExecutionChangeSignature(state, signedBLSToExecutionChange)) {
-    throw `Signature could not be verified for BLS to Execution Change for validatorIndex${addressChange.validatorIndex}`;
+    const error = Error(
+      `Signature could not be verified for BLS to Execution Change for validatorIndex${addressChange.validatorIndex}`
+    );
+    return {valid: false, error};
   }
+  return {valid: true};
```
dapplion, line 79: Whitespace above this line =D :pray:

packages/validator/src/services/validatorStore.ts
```diff
@@ -531,6 +534,28 @@ export class ValidatorStore {
     };
   }
 
+  async signBlsToExecutionChange(
```
dapplion: Is this used by someone? What's the usecase?

packages/beacon-node/test/unit-mainnet/network/gossip/scoringParameters.test.ts
twoeths: `expectedMessageRate` is quite tricky but we can tweak this later. Looking good for now

packages/beacon-node/src/chain/opPools/opPool.ts
g11tech: yes, fixing this

packages/api/src/beacon/routes/beacon/pool.ts
```diff
@@ -94,6 +102,15 @@ export type Api = {
...
   submitPoolVoluntaryExit(exit: phase0.SignedVoluntaryExit): Promise<void>;
 
+  /**
+   * Submit SignedBLSToExecutionChange object to node's pool
+   * Submits SignedBLSToExecutionChange object to node's pool and if passes validation node MUST broadcast it to network.
+   * @param requestBody
+   * @returns any BLSToExecutionChange is stored in node and broadcasted to network
+   * @throws ApiError
+   */
+  submitPoolBlsToExecutionChange(blsToExecutionChange: capella.SignedBLSToExecutionChange): Promise<void>;
```
twoeths: @g11tech  looks like we'll submit in batches for next devnets https://github.com/ethereum/beacon-APIs/pull/279

g11tech: thanks @tuyennhv ! :slightly_smiling_face: updated

packages/beacon-node/src/chain/chain.ts
```diff
@@ -703,8 +704,9 @@ export class BeaconChain implements IBeaconChain {
 
     // TODO: Improve using regen here
     const headState = this.stateCache.get(this.forkChoice.getHead().stateRoot);
+    const finalizedState = this.checkpointStateCache.get(cp);
     if (headState) {
-      this.opPool.pruneAll(headState);
+      this.opPool.pruneAll(headState, finalizedState);
```
dapplion, line 709: You must not assume the finalized state to be available

packages/beacon-node/src/network/gossip/interface.ts
```diff
@@ -23,6 +23,8 @@ export enum GossipType {
   sync_committee = "sync_committee",
   light_client_finality_update = "light_client_finality_update",
   light_client_optimistic_update = "light_client_optimistic_update",
+  // capella
+  bls_to_execution_change = "bls_to_execution_change",
```
dapplion, line 26: Dont list the fork here, just declare the topic

packages/beacon-node/src/chain/chain.ts
g11tech: correct, 
but pruning is based off on checking against finalized state if the validator has its credential updated or not, not sure if there is another way we can do this.

packages/beacon-node/src/api/impl/beacon/pool/index.ts
```diff
@@ -91,6 +96,33 @@ export function getBeaconPoolApi({
       await network.gossip.publishVoluntaryExit(voluntaryExit);
     },
 
+    async submitPoolBlsToExecutionChange(blsToExecutionChanges) {
```
twoeths, line 99: this api is only available from capella, should throw error if we're not there yet?

packages/beacon-node/src/chain/chain.ts
dapplion: Let's merge for now, but please open an issue to review latter

dapplion (approved):
LGTM! Added one pending `TODO CAPELLA` that can block Lodestar from proposing blocks

PR #4817, opened 2022-11-27
Chill capella owl

dapplion:
**Motivation**

Tested all characters in this table for eyes https://www.utf8-chartable.de/unicode-utf8-table.pl?start=9600&number=128

**Description**

Chill capella owl


wemeetagain (approved):
(◡_◡):+1:

PR #4820, opened 2022-11-28
Follow spec for processSyncAggregate

dapplion:
**Motivation**

Current implementation does not follow spec, in extreme bizarre network conditions it will lead to a consensus failure. For example:
- Proposer is withdrawn in in the block it proposes (impossible with current duties functionality)
- Proposer appears multiple times in a sync committee (impossible in networks with > 1000 validators)
- Pubkey associated with proposer does not participate and participates simultaneously in the SyncAggregate, in order yes -> no.

**Description**

Follow spec for processSyncAggregate
- Ensure sync aggregate is scanned by strict ascending order of bit index
- Calculate block proposer rewards immediately


g11tech (dismissed):
lgtm

g11tech (approved):
lgtm

PR #4828, opened 2022-11-29
Add grca rate limiter

### LINKED ISSUE #4827: Inbound rate limiter for reqresp failed for remaining limit
**Describe the bug**

The rate limiter implementation for the `reqresp` failed to validate the limit for a remaining limit. 

e.g. If remaining limit is 100 and currently you consumed 80 requests, you can still ask for the 200 more requests because the implementation validate the consumption before the increment. 

https://github.com/ChainSafe/lodestar/blob/222a91f3ac659db0feb6f7e9621ebeea8ebb085c/packages/beacon-node/src/network/reqresp/rateTracker.ts#L30

https://github.com/ChainSafe/lodestar/blob/222a91f3ac659db0feb6f7e9621ebeea8ebb085c/packages/beacon-node/src/network/reqresp/rateTracker.ts#L36

**Expected behavior**

It should not allow to over-limit in any case.

nazarhussain:
**Motivation**

Fix the logic for rate limiting. 

**Description**

- Add a new GRCA rate limiter as utility 
- Replace the existing rate limiter 
- Update the implementation of inbound rate limiter. 

Closes: https://github.com/ChainSafe/lodestar/issues/4827

Run all unit tests.


packages/utils/src/rateLimiterGRCA.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+    const msPerBucket = quota.replenishAllEvery;
+    if (msPerBucket === 0) {
+      throw Error("Replenish time must be positive");
+    }
+    const msPerToken = msPerBucket / quota.maxTokens;
+    return new RateLimiterGRCA(msPerBucket, msPerToken);
+  }
+
+  currentBucketRemainingTime(key: Key, msSinceStart?: number): number {
```
nazarhussain: @dapplion I used this rate limiter implementation from discv5. I believe this additional parameters `msSinceStart` was added for the testing purpose only and don't have a real use case. Should we remove it?

dapplion: You should copy all the unit tests to ensure this copied source code doesn't break

dapplion: Hmm un-sure would have to re-review the logic

packages/beacon-node/src/network/reqresp/inboundRateLimiter.ts
```diff
@@ -44,13 +43,13 @@ export class InboundRateLimiter {
...
-  private requestCountTrackersByPeer: MapDef<string, RateTracker>;
+  private requestCountByPeerTracker: RateLimiterGRCA<string>;
   /**
    * This rate tracker is specific to lodestar, we don't want to serve too many blocks for peers at the
    * same time, even through we limit block count per peer as in blockCountTrackersByPeer
    */
-  private blockCountTotalTracker: RateTracker;
-  private blockCountTrackersByPeer: MapDef<string, RateTracker>;
+  private blockCountTotalTracker: RateLimiterGRCA<null>;
+  private blockCountByPeerTracker: RateLimiterGRCA<string>;
```
dapplion: Change this to be generic per method. Add a quota per overall method calls, and per method items. Move the rate limiter to the reqResp package and make it generic to any route. 

```ts
interface RateLimiterOpts {
  totalRequestsQuota?: RateLimiterQuota
  methodsQuotas?: Record<string, {
    requestsQuota: RateLimiterQuota;
    itemsQuota?: RateLimiterQuota;
  }>
}
```

packages/reqresp/src/messages/BeaconBlocksByRange.ts
```diff
@@ -15,5 +16,16 @@ export const BeaconBlocksByRange: ProtocolDefinitionGenerator<
...
+    inboundRateLimits: {
+      /**
+       * Nodes uses these endpoint during the sync to fetch blocks.
+       * If we restrict too much we can get into a situation where nodes can't sync from us.
+       * Higher range may end-up in a DOS attack.
+       * So we try to use optimistic values. We can always tune this later.
+       */
+      byPeer: {quota: 500, quotaTime: minutes(1)},
+      total: {quota: 2000, quotaTime: minutes(1)},
+      getRequestCount: (req) => req.count,
```
dapplion: De-duplicate from v2, for example create a beaconBlocksByRangeInboundRateLimits, export from here and import in v2

packages/reqresp/src/messages/Goodbye.ts
```diff
@@ -12,5 +13,13 @@ export const Goodbye: ProtocolDefinitionGenerator<phase0.Goodbye, phase0.Goodbye
     responseType: () => ssz.phase0.Goodbye,
     renderRequestBody: (req) => req.toString(10),
     contextBytes: {type: ContextBytesType.Empty},
+    inboundRateLimits: {
+      /**
+       * Metadata is part of handshake process, so we keep it
+       * equivalent to status message
+       */
+      byPeer: {quota: 2, quotaTime: minutes(1)},
```
dapplion: This is not metadata, but Ping. 2 / minute is very restrictive. This can potentially kill Lodestar since we would ban all peers. What do other clients use?

packages/reqresp/src/messages/LightClientOptimisticUpdate.ts
```diff
@@ -15,5 +15,12 @@ export const LightClientOptimisticUpdate: ProtocolDefinitionGenerator<null, alta
     requestType: () => null,
     responseType: () => ssz.altair.LightClientOptimisticUpdate,
     contextBytes: getContextBytesLightclient((update) => modules.config.getForkName(update.signatureSlot), modules),
+    inboundRateLimits: {
+      /**
+       * Updates happens every block for the head. Can be optimized later on.
```
dapplion: having 1 / min in LightClientFinalityUpdate and 500 / min in LightClientOptimisticUpdate is way to disparate. The way to think about this numbers is = what is the max rate we can afford and what's the min the network needs to function? Such questions should be answered for each route, and justify the numbers

packages/cli/src/options/beaconNodeOptions/network.ts
```diff
@@ -140,38 +132,6 @@ export const options: ICliCommandOptions<INetworkArgs> = {
     group: "network",
   },
 
-  "network.requestCountPeerLimit": {
```
dapplion: Expose the rate limit factor to allow disabling.

dapplion: Also note that removing this options from here will cause potential consumers from Lodestar starting at all. There's not reason to cause that here, I would mark them as deprecated until the next major release

packages/reqresp/src/ReqResp.ts
```diff
@@ -48,29 +51,52 @@ export class ReqResp {
...
     this.controller = new AbortController();
+    this.rateLimiter.start();
     // We set infinity to prevent MaxListenersExceededWarning which get logged when listeners > 10
     // Since it is perfectly fine to have listeners > 10
     setMaxListeners(Infinity, this.controller.signal);
 
     for (const [protocolID, protocol] of this.supportedProtocols) {
-      await this.libp2p.handle(protocolID, this.getRequestHandler(protocol));
+      try {
+        await this.libp2p.handle(protocolID, this.getRequestHandler(protocol));
```
dapplion: What the rationale of this code change? What does this have to do with rate limit?

packages/reqresp/src/protocols/LightClientFinalityUpdate.ts
```diff
@@ -15,5 +15,17 @@ export const LightClientFinalityUpdate: ProtocolDefinitionGenerator<null, altair
...
+    inboundRateLimits: {
+      /**
+       * Finality updates can't be passed more frequently than once per epoch.
+       * So for one peer we allow more relaxed double.
+       *
+       * 384 seconds is chosen to be fair and equivalent to 1 epoch. Can be updated in future.
+       *
+       * For total we multiply with `10` to have lower peer count on light client.
+       */
+      byPeer: {quota: 2, quotaTime: seconds(384)},
```
dapplion: This limits are very restrictive, it's almost free to use to produce this updates. Just use 2 / 12sec like for optimistic

packages/reqresp/src/protocols/Metadata.ts
```diff
@@ -1,5 +1,15 @@
 import {allForks, ssz} from "@lodestar/types";
 import {ContextBytesType, Encoding, ProtocolDefinitionGenerator} from "../types.js";
+import {seconds} from "./utils.js";
+
+export const metadataInboundRateLimit = {
+  /**
+   * One peer does not request a lot of metadata
+   * For total we multiply with `defaultNetworkOptions.maxPeers`
```
dapplion: > For total we multiply with `defaultNetworkOptions.maxPeers`

No, you are just multiplying by 55 which may or may not be maxPeers. Other consumers may use different maxPeers thus causing un-wanted rate-limits.

By not setting any total limit at all, you are already limiting the total to `2 * maxPeers`, since otherwise peers will be down-scored. The only point of using total over byPeer is to allow higher usage by some peers but prevent an overall overload if byPeer.quota * maxPeers > what we can manage. For any protocol that you are setting total = byPeer * maxPeers, just remove the total limit.

packages/reqresp/src/protocols/LightClientOptimisticUpdate.ts
```diff
@@ -15,5 +15,17 @@ export const LightClientOptimisticUpdate: ProtocolDefinitionGenerator<null, alta
...
+      /**
+       * Optimistic updates can't be passed more frequently than once per slot.
+       * So for one peer we allow more relaxed double rate limit.
+       *
+       * 12 seconds is chosen to be fair and relates to slot but can be updated in future.
+       *
+       * For total we multiply with `10` to have lower peer count on light client.
+       */
+      byPeer: {quota: 2, quotaTime: seconds(12)},
+      total: {quota: 20, quotaTime: seconds(10)},
```
dapplion: Setting this rate limit values here is fundamentally wrong since you are "forcing" the rate limit values of a full-node to every consumer of this library which includes full nodes and light client consumers.

packages/reqresp/src/rate_limiter/ReqRespRateLimiter.ts
```diff
@@ -0,0 +1,153 @@
...
+
+      this.reportPeer(peerId);
+
+      if (this.metrics) {
+        this.metrics.rateLimitErrors.inc({tracker: "requestCountPeerTracker"});
+      }
+
+      throw new RequestError(
+        {
+          code: RequestErrorCode.REQUEST_RATE_LIMITED,
```
dapplion: Remove new line after `{` for consistency with rest of codebase. Same for `{
        peerId: peerIdStr,
      }`

dapplion (changes requested):

packages/reqresp/src/rate_limiter/ReqRespRateLimiter.ts
nazarhussain: I formatted this manually for now, but such things should be left to the `prettier` and should not be engaged in the code review cycle.

dapplion: Unfortunately I think prettier in flexible on this instance specifically

packages/beacon-node/src/network/options.ts
```diff
@@ -11,6 +11,12 @@ export interface INetworkOptions extends PeerManagerOpts, ReqRespBeaconNodeOpts,
   mdns: boolean;
   connectToDiscv5Bootnodes?: boolean;
   version?: string;
+
+  // TODO: These options  are deprecated and will be removed in next major release.
+  requestCountPeerLimit?: number;
+  blockCountPeerLimit?: number;
+  blockCountTotalLimit?: number;
+  rateTrackerTimeoutMs?: number;
```
dapplion: Let me know your opinion, but currently we do API breakages for programmatic usage options. I would argue to keep breaking until https://github.com/ChainSafe/lodestar/pull/4576 so I'm in favor of dropping the options from here, but keep the CLI flags marked as deprecated

packages/reqresp/src/protocols/Metadata.ts
nazarhussain: Yes `maxPeers` can be changed runtime. That's why I mentioned `defaultNetworkOptions.maxPeers` and not `network.maxPeers`. If user change that option, he can also change limits by `network.rateLimitMultiplier`.

We don't know how many maximum inbound peers are allowed on the node, so we assume and initialize with that the default value is, which is 55. 

I understand the point of having implicit `maxPeers` limit, but since `total` quota was not set as optional so had to initialize it's value.

packages/reqresp/src/protocols/LightClientOptimisticUpdate.ts
nazarhussain: It's not **forcing** the rate limits rather setting **default** values. Any consumer of `reqresp` package can change these independently at the time of protocol registration.

packages/beacon-node/src/network/options.ts
nazarhussain: Yes I agree, we can remove the options and just keep the flags. As the CLI value parsing actually don't have any impact. To not break the CLI scripts user may have, we can keep the CLI options.

nazarhussain: For reference it was kept because of the code usage. But since the argument is deprecated we can remove those descriptions. 

```ts
defaultDescription: String(defaultOptions.network.rateTrackerTimeoutMs),
```

packages/reqresp/src/protocols/LightClientOptimisticUpdate.ts
dapplion: Right, tho there's a debate if the default should be no value at all (completely agnostic), or we apply the default that's convenient to our specific usecase

packages/reqresp/src/protocols/Metadata.ts
dapplion: total and byPeer should be optional params

packages/beacon-node/src/network/options.ts
dapplion: Yes, for deprecated options I would just keep the minimum code necessary to not make yargs throw an error

packages/reqresp/src/protocols/LightClientOptimisticUpdate.ts
nazarhussain: I believe our use case should be similar to lot others, so setting default values is much convenient. If one had a unique use can he still can register the protocol with different limits, there is no restriction on it.

dapplion (dismissed):

dapplion (approved):

PR #4829, opened 2022-11-30
Do not unsubscribe subnets if not less than 64 validators

### LINKED ISSUE #4818: v1.2.1: Low mesh peers due to p7
**Describe the bug**

Gossipsub score has been low since v1.2.1

Screen Shot 2022-11-27 at 12 54 15

Screen Shot 2022-11-27 at 12 55 03

due to P7

Screen Shot 2022-11-27 at 12 55 23

**Expected behavior**

Mesh peers should be the same to v1.1.0

twoeths:
**Motivation**

- When there are >= 64 validators and some random subnets are expired, right now we do redundant unsubscribe() then subscribe() again to same subnets. In gossipsub 0.14.1 it's not an issue but v4.1.1 just ignore the subscription.
- That's actually a TODO in AttnetsService

**Description**

Check if we have to subscribe to all subnets, if yes just extend the random subnets subscriptions instead

part of #4818

### DIFF
--- a/packages/beacon-node/src/network/subnets/attnetsService.ts
+++ b/packages/beacon-node/src/network/subnets/attnetsService.ts
@@ -172,7 +172,16 @@ export class AttnetsService implements IAttnetsService {
    */
   private unsubscribeExpiredRandomSubnets(slot: Slot): void {
     const expired = this.subscriptionsRandom.getExpired(slot);
-    // TODO: Optimization: If we have to be subcribed to all subnets, no need to unsubscribe. Just extend the timeout
+    const currentSlot = this.chain.clock.currentSlot;
+
+    if (this.knownValidators.size * RANDOM_SUBNETS_PER_VALIDATOR >= ATTESTATION_SUBNET_COUNT) {
+      // Optimization: If we have to be subcribed to all subnets, no need to unsubscribe. Just extend the timeout
+      for (const subnet of expired) {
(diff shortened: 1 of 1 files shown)


dapplion (approved):

packages/beacon-node/src/network/subnets/attnetsService.ts
```diff
@@ -172,7 +172,16 @@ export class AttnetsService implements IAttnetsService {
...
-    // TODO: Optimization: If we have to be subcribed to all subnets, no need to unsubscribe. Just extend the timeout
+    const currentSlot = this.chain.clock.currentSlot;
+
+    if (this.knownValidators.size * RANDOM_SUBNETS_PER_VALIDATOR >= ATTESTATION_SUBNET_COUNT) {
+      // Optimization: If we have to be subcribed to all subnets, no need to unsubscribe. Just extend the timeout
+      for (const subnet of expired) {
+        this.subscriptionsRandom.request({subnet, toSlot: randomSubscriptionSlotLen() + currentSlot});
+      }
+      return;
+    }
```
dapplion, line 183: Does this also cover the --subscribe-all-subnets flag ?

twoeths: @dapplion  it's already covered https://github.com/ChainSafe/lodestar/blob/1fde79dad53c91c6df05efb04b12a09bd1275d00/packages/beacon-node/src/network/subnets/attnetsService.ts#L276

PR #4841, opened 2022-12-05
EIP-4844: Use BlockInput

dapplion:
**Motivation**

**Description**

This PR should not include any changes in business logic, just replace a function input type, used widely on the codebase

- Use `BlockInput` struct enum to accommodate block OR blocks+blobs in the future


packages/beacon-node/src/network/network.ts
```diff
@@ -38,7 +39,7 @@ interface INetworkModules {
 
 export class Network implements INetwork {
   events: INetworkEventBus;
-  reqResp: IReqRespBeaconNode;
```
twoeths: why do we have to switch to implementation instead of interface?

g11tech (commented):
looks good apart from the same query @tuyennhv  has raised, is it because of the stub type issues since I think i faced a similar situation earlier?

twoeths (approved):

packages/beacon-node/src/network/network.ts
dapplion: Removed unwanted diff

PR #4847, opened 2022-12-05
EIP-4844: Add blobsSidecar db repos

dapplion:
**Motivation**

**Description**

BlobsSidecar have to be stored, served and prune with identical logic to blocks. So this two new repos mimic existing for blocks. See https://github.com/ChainSafe/lodestar/pull/4774 for how this repos will be used specifically


g11tech (approved):
lgtm!

PR #4851, opened 2022-12-05
EIP-4844: Add c-kzg library and current trusted setup

dapplion:
**Motivation**

EIP-4844 needs a new crypto backend, which will co-exist to BLST. It also needs access to a network specific trusted setup, which is a list of G1 and G2 points

**Description**

- Add c-kzg library and current trusted setup


packages/beacon-node/src/util/kzg.ts
```diff
@@ -0,0 +1,113 @@
...
+import fs from "node:fs";
+import {fileURLToPath} from "node:url";
+import {FIELD_ELEMENTS_PER_BLOB, loadTrustedSetup} from "c-kzg";
+import {fromHex, toHex} from "@lodestar/utils";
+
+// Global variable __dirname no longer available in ES6 modules.
+// Solutions: https://stackoverflow.com/questions/46745014/alternative-for-dirname-in-node-js-when-using-es6-modules
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+export const TRUSTED_SETUP_BIN_FILEPATH = path.join(__dirname, "../../trusted_setup.bin");
```
g11tech, line 11: would this be network dependent or same setup will work for all networks (and then the question of chains like gnosis etc)

g11tech (approved):
lgtm, just a query regarding the trusted bin customization w.r.t. networks

packages/beacon-node/src/util/kzg.ts
dapplion: Still unknown, probably it will be shared since it's only dependant on the size of blobs. I would bet big on being the same file for all eth2 networks

twoeths (approved):

PR #4862, opened 2022-12-06
EIP-4844: Archive hot blobs and prune old blobs

dapplion:
**Motivation**

**Description**

- On finalization, move cannonical blobs to archive, delete others
- After retention period, prune old blobs. Simple logic that even on restarts will prune any range of old blocks.

### DIFF
--- a/packages/beacon-node/src/node/nodejs.ts
+++ b/packages/beacon-node/src/node/nodejs.ts
@@ -149,6 +149,9 @@ export class BeaconNode {
 
     // start db if not already started
     await db.start();
+    // Prune hot db repos
+    // TODO: Should this call be awaited?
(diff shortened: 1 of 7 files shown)


packages/beacon-node/src/node/nodejs.ts
```diff
@@ -149,6 +149,9 @@ export class BeaconNode {
 
     // start db if not already started
     await db.start();
+    // Prune hot db repos
+    // TODO: Should this call be awaited?
```
wemeetagain, line 153: Probably so unless we can be sure the pruning is done by the time we're syncing.

twoeths (dismissed):

twoeths (approved):

PR #4843, opened 2022-12-05
EIP-4844: Add engine API getBlobsBundle method

dapplion:
**Motivation**

- From https://github.com/ChainSafe/lodestar/pull/4774

**Description**

- Add engine API getBlobsBundle method
- Update consumers to use ForkName

PR #4864, opened 2022-12-06
EIP-4844: Update produce block flow

dapplion:
**Motivation**

**Description**

- Update block production flow to produce blobs if necessary. Since validator does not have to sign the blobsSidecar, it stays cached in the beacon node until the validator calls publishBlock back.
- Uses https://github.com/ChainSafe/lodestar/pull/4843 to create blobs

### DIFF
--- a/packages/beacon-node/src/chain/chain.ts
+++ b/packages/beacon-node/src/chain/chain.ts
@@ -1,4 +1,5 @@
 import path from "node:path";
+import {computeAggregateKzgProof} from "c-kzg";
 import {
   BeaconStateAllForks,
   CachedBeaconStateAllForks,
@@ -12,12 +13,12 @@ import {
   PubkeyIndexMap,
 } from "@lodestar/state-transition";
 import {IBeaconConfig} from "@lodestar/config";
-import {allForks, UintNum64, Root, phase0, Slot, RootHex, Epoch, ValidatorIndex} from "@lodestar/types";
+import {allForks, UintNum64, Root, phase0, Slot, RootHex, Epoch, ValidatorIndex, eip4844} from "@lodestar/types";
 import {CheckpointWithHex, ExecutionStatus, IForkChoice, ProtoBlock} from "@lodestar/fork-choice";
 import {ProcessShutdownCallback} from "@lodestar/validator";
-import {ILogger, toHex} from "@lodestar/utils";
+import {ILogger, pruneSetToMax, toHex} from "@lodestar/utils";
 import {CompositeTypeAny, fromHexString, TreeView, Type} from "@chainsafe/ssz";
-import {SLOTS_PER_EPOCH} from "@lodestar/params";
+import {ForkSeq, SLOTS_PER_EPOCH} from "@lodestar/params";
 
 import {GENESIS_EPOCH, ZERO_HASH} from "../constants/index.js";
 import {IBeaconDb} from "../db/index.js";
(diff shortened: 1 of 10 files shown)


packages/beacon-node/src/chain/produceBlock/produceBlockBody.ts
```diff
@@ -188,6 +211,34 @@ export async function produceBlockBody<T extends BlockType>(
           if (payload.transactions.length === 0) {
             this.metrics?.blockPayload.emptyPayloads.inc({prepType});
           }
+
+          // Capella and later forks have withdrawals on their ExecutionPayload
+          if (fork === ForkName.capella || fork === ForkName.eip4844) {
+            // TODO EIP-4844 Remove this when the EC includes `withdrawals`
+            (blockBody as capella.BeaconBlockBody).executionPayload.withdrawals = [];
```
g11tech: this will break the withdrawals, so we should do this only in an unmerged PR

g11tech: removing from here, so that we can only keep this in the unmerged eip-4844 parent PR

dapplion: Sure, we should merge withdrawal logic here ASAP. Is it there already on unstable or no-where yet?

packages/beacon-node/src/chain/chain.ts
```diff
@@ -359,27 +371,65 @@ export class BeaconChain implements IBeaconChain {
...
+   *   return BlobsSidecar(
+   *       beacon_block_root=hash_tree_root(block),
+   *       beacon_block_slot=block.slot,
+   *       blobs=blobs,
+   *       kzg_aggregated_proof=compute_proof_from_blobs(blobs),
+   *   )
+   */
+  getBlobsSidecar(beaconBlock: eip4844.BeaconBlock): eip4844.BlobsSidecar {
+    const blockHash = toHex(beaconBlock.body.executionPayload.blockHash);
+    const blobsSidecar = this.producedBlobsSidecarCache.get(blockHash);
```
twoeths, line 425: seems here we get from beacon block hash while we cache by payload block hash

packages/beacon-node/src/chain/produceBlock/produceBlockBody.ts
```diff
@@ -188,6 +211,34 @@ export async function produceBlockBody<T extends BlockType>(
...
+
+          if (fork === ForkName.eip4844) {
+            // SPEC: https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/validator.md#blob-kzg-commitments
+            // After retrieving the execution payload from the execution engine as specified in Bellatrix, use the
+            // payload_id to retrieve blobs and blob_kzg_commitments via get_blobs_and_kzg_commitments(payload_id)
+            // TODO EIP-4844: getBlobsBundle and getPayload must be either coupled or called in parallel to save time.
+            const blobsBundle = await this.executionEngine.getBlobsBundle(payloadId);
+
+            // Sanity check consistency between getPayload() and getBlobsBundle()
+            const blockHash = toHex(payload.blockHash);
```
twoeths, line 223: here we return payload block hash but seems later on we get by beacon block hash

dapplion: Get by execution payload blockHash

https://github.com/ChainSafe/lodestar/blob/95739bde2cc380ab0089910f0a22eac51e63be21/packages/beacon-node/src/chain/chain.ts#L424-L425

Set by execution payload blockHash

https://github.com/ChainSafe/lodestar/blob/95739bde2cc380ab0089910f0a22eac51e63be21/packages/beacon-node/src/chain/chain.ts#L397

Which is computed here from the payload

https://github.com/ChainSafe/lodestar/blob/95739bde2cc380ab0089910f0a22eac51e63be21/packages/beacon-node/src/chain/produceBlock/produceBlockBody.ts#L229

packages/beacon-node/src/chain/produceBlock/produceBlockBody.ts
g11tech: merge withdrawal logic is already in the executin/http where withdrawals are bundeled on execution payload depending on fork

g11tech: we set withdrawal to attributes here: https://github.com/ChainSafe/lodestar/pull/4864/files#diff-c331f7b8e7121839c55f5f20f21d1efe1375b69c6fbbe5db12a4baf1ed4acf5dR350, and rest execution apis handle it appropriately

g11tech (approved):
lgtm!

packages/beacon-node/src/chain/chain.ts
g11tech: `const blockHash = toHex(beaconBlock.body.executionPayload.blockHash);` is  payload blockHash

PR #4865, opened 2022-12-06
EIP-4844: Add onBlobsSidecarsByRange

dapplion:
**Motivation**

**Description**

- Adds handler for ReqResp protocol BlobsSidecarsByRange
- Logic very similar to blocks by range but IMO similar. @tuyennhv can you please review this logic is safe for non-finalized blobs?

### DIFF
--- a/packages/beacon-node/src/network/reqresp/handlers/blobsSidecarsByRange.ts
+++ b/packages/beacon-node/src/network/reqresp/handlers/blobsSidecarsByRange.ts
@@ -1,0 +1,91 @@
+import {GENESIS_SLOT, MAX_REQUEST_BLOCKS} from "@lodestar/params";
+import {eip4844} from "@lodestar/types";
+import {fromHex} from "@lodestar/utils";
+import {ContextBytesType, EncodedPayload, EncodedPayloadType, ResponseError, RespStatus} from "@lodestar/reqresp";
+import {IBeaconChain} from "../../../chain/index.js";
+import {IBeaconDb} from "../../../db/index.js";
+
+// TODO EIP-4844: Unit test
+
+export async function* onBlobsSidecarsByRange(
+  requestBody: eip4844.BlobsSidecarsByRangeRequest,
+  chain: IBeaconChain,
+  db: IBeaconDb
+): AsyncIterable<EncodedPayload<eip4844.BlobsSidecar>> {
+  const {startSlot} = requestBody;
+  let {count} = requestBody;
+
+  if (count < 1) {
+    throw new ResponseError(RespStatus.INVALID_REQUEST, "count < 1");
+  }
+  // TODO: validate against MIN_EPOCHS_FOR_BLOCK_REQUESTS
+  if (startSlot < GENESIS_SLOT) {
+    throw new ResponseError(RespStatus.INVALID_REQUEST, "startSlot < genesis");
+  }
+
+  if (count > MAX_REQUEST_BLOCKS) {
+    count = MAX_REQUEST_BLOCKS;
+  }
+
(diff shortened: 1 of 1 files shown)


g11tech (approved):
lgtm!

packages/beacon-node/src/network/reqresp/handlers/blobsSidecarsByRange.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  // Non-finalized tram of blobs
+
+  if (endSlot > finalizedSlot) {
+    const headRoot = chain.forkChoice.getHeadRoot();
+    // TODO EIP-4844: forkChoice should mantain an array of canonical blocks, and change only on reorg
+    const headChain = chain.forkChoice.getAllAncestorBlocks(headRoot);
+
+    // Iterate head chain with ascending block numbers
+    for (const block of headChain) {
```
twoeths, line 64: @dapplion  `getAllAncestorBlocks` returns blocks in in descending order, so a `reverse()` is needed?

packages/beacon-node/src/network/reqresp/handlers/blobsSidecarsByRange.ts
```diff
@@ -0,0 +1,91 @@
...
+      // Must include only blocks in the range requested
+      if (block.slot >= startSlot && block.slot < endSlot) {
+        // Note: Here the forkChoice head may change due to a re-org, so the headChain reflects the cannonical chain
+        // at the time of the start of the request. Spec is clear the chain of blobs must be consistent, but on
+        // re-org there's no need to abort the request
+        // Spec: https://github.com/ethereum/consensus-specs/blob/a1e46d1ae47dd9d097725801575b46907c12a1f8/specs/eip4844/p2p-interface.md#blobssidecarsbyrange-v1
+
+        const blockBytes = await db.blobsSidecar.getBinary(fromHex(block.blockRoot));
+        if (blockBytes) {
```
twoeths, line 73: if we don't have `blockBytes`, maybe throw Error otherwise the response of this is not in synchronized to `beaconBlocksByRange`

packages/beacon-node/src/network/reqresp/handlers/blobsSidecarsByRange.ts
dapplion: Good point! Pushing a fix

packages/beacon-node/src/network/reqresp/handlers/blobsSidecarsByRange.ts
dapplion: Hmm, I'm undecided here, our db should never become out of sync with the fork-choice. However they could be some edge cases. Should we fail the entire response if we can't find the block in the db?

twoeths: I think we should, otherwise our response is not really a chain (one block may miss its parent for example) and the client node can't process all of the returned blocks

also found this in the spec https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#beaconblocksbyrange

```
Clients MUST respond with blocks that are consistent from a single chain within the context of the request.
```

dapplion: @tuyennhv then we should change the blocks by range implementation to also throw if the block binary is not found right?

twoeths: @dapplion  yeah I think so, gonna make a PR for it 👍

PR #4866, opened 2022-12-06
EIP-4844: Add onBeaconBlockAndBlobsSidecarByRoot

dapplion:
**Motivation**

- From https://github.com/ChainSafe/lodestar/pull/4774

**Description**

- Add ReqResp handler for BeaconBlockAndBlobsSidecarByRoot protocol
- Logic is pretty simple, just couples getting blocks and blobs by root

PR #4869, opened 2022-12-06
EIP-4844: Implement beaconBlocksMaybeBlobsByRoot

dapplion:
**Motivation**

- From https://github.com/ChainSafe/lodestar/pull/4774/files

**Description**

- Implement beaconBlocksMaybeBlobsByRoot to logically couple beaconBlocksByRoot + beaconBlockAndBlobsSidecarByRoot
- Consumes the protocol handled by https://github.com/ChainSafe/lodestar/pull/4866
- The nasty logic is there to handle the edge case explained in https://github.com/ethereum/consensus-specs/issues/3113

### DIFF
--- a/packages/beacon-node/src/network/network.ts
+++ b/packages/beacon-node/src/network/network.ts
@@ -289,7 +289,7 @@ export class Network implements INetwork {
         request.map(
           async (beaconBlockRoot): Promise<BlockInput | null> => {
             const [resultBlockBlobs, resultBlocks] = await Promise.allSettled([
-              this.reqResp.beaconBlockAndBlobsSidecarByRoot(peerId, []),
+              this.reqResp.beaconBlockAndBlobsSidecarByRoot(peerId, [beaconBlockRoot]),
               this.reqResp.beaconBlocksByRoot(peerId, [beaconBlockRoot]),
             ]);
 
--- a/packages/beacon-node/src/network/network.ts
+++ b/packages/beacon-node/src/network/network.ts
@@ -268,9 +268,67 @@ export class Network implements INetwork {
   }
 
   async beaconBlocksMaybeBlobsByRoot(peerId: PeerId, request: phase0.BeaconBlocksByRootRequest): Promise<BlockInput[]> {
-    // TODO EIP-4844: Will throw an error for blocks post EIP-4844
-    const blocks = await this.reqResp.beaconBlocksByRoot(peerId, request);
-    return blocks.map((block) => getBlockInput.preEIP4844(this.config, block));
+    // Assume all requests are post EIP-4844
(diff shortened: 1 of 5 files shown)


packages/beacon-node/src/network/network.ts
```diff
@@ -206,9 +206,67 @@ export class Network implements INetwork {
...
+    }
+
+    // NOTE: Consider blocks may be post or pre EIP-4844
+    // TODO EIP-4844: Request either blocks, or blocks+blobs
+    else {
+      const results = await Promise.all(
+        request.map(
+          async (beaconBlockRoot): Promise<BlockInput | null> => {
+            const [resultBlockBlobs, resultBlocks] = await Promise.allSettled([
+              this.reqResp.beaconBlockAndBlobsSidecarByRoot(peerId, []),
```
twoeths: I suppose it should not empty array?

dapplion: Oh good catch!

packages/beacon-node/src/network/network.ts
```diff
@@ -243,9 +243,67 @@ export class Network implements INetwork {
   }
 
   async beaconBlocksMaybeBlobsByRoot(peerId: PeerId, request: phase0.BeaconBlocksByRootRequest): Promise<BlockInput[]> {
-    // TODO EIP-4844: Will throw an error for blocks post EIP-4844
-    const blocks = await this.reqResp.beaconBlocksByRoot(peerId, request);
-    return blocks.map((block) => getBlockInput.preEIP4844(this.config, block));
+    // Assume all requests are post EIP-4844
```
g11tech, line 271: shouldn't we just do a normal beaconBlocksByRoot request followed by blobs downloading for those blocks which are >= eip4844 epoch?

dapplion: Not sure I follow, why would you query blocks, then blocks+blobs when you can query blocks+blobs directly and save 1 request?

g11tech: `this.config.getForkSeq(this.chain.forkChoice.getFinalizedBlock().slot) >= ForkSeq.eip4844` is not sufficient condition to establish it all post eip4844 blocks, thinking of backfil

g11tech: looks good with the knowledge that BeaconBlocksByRoot is to be primarily used for post finalized

g11tech (dismissed):
lgtm

packages/beacon-node/src/network/network.ts
dapplion: Backfill should not use this method. p2p spec says that only unfinalized blocks are server by root. Else use by range. So once the network has finalized the eip4844 no node will serve via p2p a by root request without blobs

g11tech (approved):
lgtm

PR #4882, opened 2022-12-07
EIP-4844: Handle BlobsSidecar

dapplion:
**Motivation**

**Description**

Complete handling logic for BlobsSidecar
- On gossip, handle BlockInput and call processBlock
- On block verify conditionally run validateBlobsSidecar, skipping for gossip since it's already verified
- On block import persist BlobsSidecar to db


g11tech (approved):
lgtm!

PR #4879, opened 2022-12-07
Declare spec tests skipped in preset test runner

dapplion:
**Motivation**

Currently we declare what spec tests are skipped in an obscure utility. Such a critical decision should be surfaced to the test runner for better clarity.

**Description**

- Declare spec tests skipped in preset test runner

PR #4883, opened 2022-12-07
Move all spec test skip statements to top file

dapplion:
**Motivation**

- Extends https://github.com/ChainSafe/lodestar/pull/4879 but to every single skip statement on spec tests

**Description**

- Move all spec test skip statements to top file


g11tech (approved):
lgtm!

PR #4884, opened 2022-12-07
feat: render named stack traces on ssz type errors

dapplion:
**Motivation**

Whenever there's an SSZ error, the stack trace looks like this

```
Error: First offset must equal to fixedEnd 376 != 344
  at readVariableOffsets (/home/lion/Code/eth2.0/lodestar/node_modules/@chainsafe/ssz/src/type/container.ts:448:15)
  at ContainerType.getFieldRanges (/home/lion/Code/eth2.0/lodestar/node_modules/@chainsafe/ssz/src/type/container.ts:399:21)
  at ContainerType.value_deserializeFromBytes (/home/lion/Code/eth2.0/lodestar/node_modules/@chainsafe/ssz/src/type/container.ts:201:30)
  at ContainerType.value_deserializeFromBytes (/home/lion/Code/eth2.0/lodestar/node_modules/@chainsafe/ssz/src/type/container.ts:208:36)
  at ContainerType.deserialize (/home/lion/Code/eth2.0/lodestar/node_modules/@chainsafe/ssz/src/type/abstract.ts:135:17)
```

which sucks because you wonder, which is the offending property? Where should I go look next?

Would be awesome if the stack trace also included the path of what property is causing the error. Previous attempts:
- Add try / catch on every method and concatenate the path of the error, for example: `state.executionPayloadHeader.excessGas`. However this requires adding tons of boilerplate code + has a serious performance penalty.

But what if instead of rendering `ContainerType` in stack trace rows it rendered `BeaconStateCapella`? Well, that's what this PR does

**Description**

The above error renders with this PR like
```
Error: First offset must equal to fixedEnd 376 != 344
  at readVariableOffsets (/home/lion/Code/eth2.0/lodestar/node_modules/@chainsafe/ssz/src/type/container.ts:448:15)
  at ExecutionPayloadHeaderCapellaType.getFieldRanges (/home/lion/Code/eth2.0/lodestar/node_modules/@chainsafe/ssz/src/type/container.ts:399:21)
  at ExecutionPayloadHeaderCapellaType.value_deserializeFromBytes (/home/lion/Code/eth2.0/lodestar/node_modules/@chainsafe/ssz/src/type/container.ts:201:30)
  at BeaconStateCapellaType.value_deserializeFromBytes (/home/lion/Code/eth2.0/lodestar/node_modules/@chainsafe/ssz/src/type/container.ts:208:36)
  at BeaconStateCapellaType.deserialize (/home/lion/Code/eth2.0/lodestar/node_modules/@chainsafe/ssz/src/type/abstract.ts:135:17)
```

Much nicer! The key here is to force JS to pick up a dedicated "object name" for each class instance. I've tried many hacks to do that dynamically but all fail. 
- All engines I've tried consider object.name a read-only immutable property
- They appear to pick the name of the prototype, not of the instance. So if there are multiple instances of ContainerType they will all have the same name
- The only viable option I've found is to declare a new class for each type, so they have a custom name.

Relevant sources on dynamic object names
- https://stackoverflow.com/questions/33605775/es6-dynamic-class-names

**Drawbacks**
- This change will cause all SSZ operations to require one extra hop up the prototype chain. I've not seen that to cause any noticeable performance penalties, and modern engine should cache those paths anyway
- Typescript compilation seems to take the same time, no observed issue there

### DIFF
--- a/packages/api/src/beacon/routes/beacon/block.ts
+++ b/packages/api/src/beacon/routes/beacon/block.ts
@@ -285,15 +285,21 @@ export function getReqSerializers(config: ChainForkConfig): ReqSerializers<Api, 
 }
 
 export function getReturnTypes(): ReturnTypes<Api> {
-  const BeaconHeaderResType = new ContainerType({
-    root: ssz.Root,
-    canonical: ssz.Boolean,
-    header: ssz.phase0.SignedBeaconBlockHeader,
-  });
+  const BeaconHeaderResType = ContainerType.named(
+    {
+      root: ssz.Root,
+      canonical: ssz.Boolean,
+      header: ssz.phase0.SignedBeaconBlockHeader,
+    },
+    {typeName: "BeaconHeaderResType", jsonCase: "eth2"}
+  );
 
-  const RootContainer = new ContainerType({
-    root: ssz.Root,
-  });
+  const RootContainer = ContainerType.named(
+    {
+      root: ssz.Root,
+    },
+    {typeName: "RootContainer", jsonCase: "eth2"}
+  );
 
   return {
     getBlock: ContainerData(ssz.phase0.SignedBeaconBlock),
--- a/packages/api/src/beacon/routes/beacon/state.ts
+++ b/packages/api/src/beacon/routes/beacon/state.ts
@@ -295,52 +295,55 @@ export function getReqSerializers(): ReqSerializers<Api, ReqTypes> {
 }
 
 export function getReturnTypes(): ReturnTypes<Api> {
-  const RootContainer = new ContainerType({
-    root: ssz.Root,
-  });
+  const RootContainer = ContainerType.named(
+    {
+      root: ssz.Root,
+    },
+    {typeName: "RootContainer", jsonCase: "eth2"}
+  );
 
-  const FinalityCheckpoints = new ContainerType(
+  const FinalityCheckpoints = ContainerType.named(
     {
       previousJustified: ssz.phase0.Checkpoint,
       currentJustified: ssz.phase0.Checkpoint,
       finalized: ssz.phase0.Checkpoint,
     },
-    {jsonCase: "eth2"}
+    {typeName: "FinalityCheckpoints", jsonCase: "eth2"}
   );
 
-  const ValidatorResponse = new ContainerType(
+  const ValidatorResponse = ContainerType.named(
     {
       index: ssz.ValidatorIndex,
       balance: ssz.UintNum64,
       status: new StringType<ValidatorStatus>(),
       validator: ssz.phase0.Validator,
     },
-    {jsonCase: "eth2"}
+    {typeName: "ValidatorResponse", jsonCase: "eth2"}
   );
 
-  const ValidatorBalance = new ContainerType(
+  const ValidatorBalance = ContainerType.named(
     {
       index: ssz.ValidatorIndex,
       balance: ssz.UintNum64,
     },
-    {jsonCase: "eth2"}
+    {typeName: "ValidatorBalance", jsonCase: "eth2"}
   );
 
-  const EpochCommitteeResponse = new ContainerType(
+  const EpochCommitteeResponse = ContainerType.named(
     {
       index: ssz.CommitteeIndex,
       slot: ssz.Slot,
       validators: ssz.phase0.CommitteeIndices,
     },
-    {jsonCase: "eth2"}
+    {typeName: "EpochCommitteeResponse", jsonCase: "eth2"}
   );
 
-  const EpochSyncCommitteesResponse = new ContainerType(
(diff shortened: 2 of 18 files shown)


wemeetagain:
Please use something like this:

```ts
import {ContainerType, Type} from "@chainsafe/ssz";

export function namedContainerType<Fields extends Record<string, Type<unknown>>>(
  name: string,
  fields: Fields,
  opts?: ConstructorParameters<typeof ContainerType>[1]
): ContainerType<Fields> {
  const NamedContainerType = new Function("superClass", `return class ${name} extends superClass {}`)(
    ContainerType
  ) as typeof ContainerType;

  return (new NamedContainerType(fields, opts) as unknown) as ContainerType<Fields>;
}
```

```ts

const BeaconBlockHeader = namedContainerType(
  "BeaconBlockHeaderType",
   {
    slot: Slot,
    proposerIndex: ValidatorIndex,
    parentRoot: Root,
    stateRoot: Root,
    bodyRoot: Root,
  },
  {typeName: "BeaconBlockHeader", jsonCase: "eth2", cachePermanentRootStruct: true}
);
```

dapplion:
@wemeetagain  wow that's a wonderful hack! Much cleaner than my approach. I've upstream this to SSZ, please review there and let's cut a release
- https://github.com/ChainSafe/ssz/pull/286

wemeetagain:
> benchmarks wtf

maybe we should deploy this and make sure there's no performance regression :sweat_smile:

wemeetagain:
It would be good to add a test to ensure naming consistency

eg:
- for every type
  - if the type is only in one fork, that the name of the class matches the exported name of the type
  - if the type is only in multiple forks, that the name of the class matches the exported name of the type with the fork name appended

dapplion:
I've been trying to get a CPU profile to understand the performance decrease but all profiles return no data, as the application is in idle when running this script

```ts
import {generatePerfTestCachedStatePhase0} from "../../../../../../state-transition/test/perf/util.js";
import {getPubkeysForIndices} from "../../../../../src/api/impl/validator/utils.js";
import {linspace} from "../../../../../src/util/numpy.js";

// Preparation
const state = generatePerfTestCachedStatePhase0();

// Infinite benchmark
for (let i = 0; true; i++) {
  const indexes = linspace(0, 1000);
  getPubkeysForIndices(state.validators, indexes);

  if (i % 1000 === 0) {
    await new Promise((r) => setTimeout(r, 0)); // yield for tooling to capture data
  }
}
```

matthewkeil:
@dapplion I was able to get this running. I think this is what you are looking for but will be happy to update if not. See the flamegraph below.  I ran the script above for 1000 iterations.  I included the zip as well.  Took a bit of hacking but basically built the test folders to js and then built the script to js as well and profiled with 0x.  Let me know if you need anything else or help getting it going.

steps to reproduce:
1) add `packages/beacon-node/src/test.ts`
2) add `test` to `includes` array in `packages/state-transition/tsconfig.build.json`
3) `yarn build` the whole repo
4) `npm i -g 0x http-server`
5) double check the imports look correct in `packages/beacn-node/lib/src/test.js` (i had to manually update)
6) `0x packages/beacn-node/lib/src/test.js`
7) `http-server PATH_TO_OUPUT` the path will be in the logging from `0x` when it finished

Screen Shot 2023-05-26 at 1 22 28 PM


[getPubkeysForIndices-flamegraph.zip](https://github.com/ChainSafe/lodestar/files/11578423/getPubkeysForIndices-flamegraph.zip)

dapplion:
@matthewkeil thank you! However the profile you shared captures the data preparation only (generate state and compute it's root). I upddated the code snippet with comments, could you capture a profile only after the comment `// Infinite benchmark`?

matthewkeil:
@dapplion here are two zips.  Both were 1M runs through the loop.  Unstable takes 32s and `dapplion/named_types` takes 1177s (33x longer).  You can serve them locally with `http-server` or similar.  I looked at them both and there is a lot of difference and a lot of garbage collection on the named-types branch.  I am going to analyze more deeply and look at the ssz code to try and sort through the differences.  In the viewer webpage there are buttons at the bottom to see the internals/cpp/native etc and turning them all on makes it more apparent what is going on but not clear why its taking so long.  Perhaps your keen eye will be helpful.

```sh
npm i -g http-server
cd /folder/with/both/unzipped
http-server
```
Then open a browser to `http://localhost:8080` and navigate to the two `flamegraph.html`.

[unstable_32s_1M_runs.zip](https://github.com/ChainSafe/lodestar/files/11648011/unstable_32s_1M_runs.zip)
[named_types_1177s_1M_runs.zip](https://github.com/ChainSafe/lodestar/files/11648012/named_types_1177s_1M_runs.zip)

wemeetagain:
closing as stale, discussion can continue in https://github.com/ChainSafe/lodestar/issues/6356, PR can reopen on progress

PR #4889, opened 2022-12-08
Handle missing blocks in onBeaconBlocksByRange

twoeths:
**Motivation**

- The returned data of onBeaconBlocksByRange should be a chain segment, if we miss any blocks, throw Error
- From the discussion in https://github.com/ChainSafe/lodestar/pull/4865#discussion_r1041858081

**Description**

- Handle similarly for `onBlobsSidecarsByRange`

### DIFF
--- a/packages/beacon-node/src/network/reqresp/handlers/beaconBlocksByRange.ts
+++ b/packages/beacon-node/src/network/reqresp/handlers/beaconBlocksByRange.ts
@@ -112,6 +112,9 @@ async function getUnfinalizedBlocksAtSlots(
         slot: slots[i],
         bytes: block,
       });
+    } else {
+      // otherwise the returned block is not really a chain segment if it misses a block in the middle
+      throw new ResponseError(RespStatus.RESOURCE_UNAVAILABLE, `No block found for slot ${slots[i]}`);
     }
   }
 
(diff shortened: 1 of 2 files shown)


packages/beacon-node/src/network/reqresp/handlers/beaconBlocksByRange.ts
```diff
@@ -112,6 +112,9 @@ async function getUnfinalizedBlocksAtSlots(
         slot: slots[i],
         bytes: block,
       });
+    } else {
+      // otherwise the returned block is not really a chain segment if it misses a block in the middle
+      throw new ResponseError(RespStatus.RESOURCE_UNAVAILABLE, `No block found for slot ${slots[i]}`);
```
dapplion: Reverse the condition above 

```ts
if (not) {
  throw
}
return ok
```

dapplion: @tuyennhv The status ResourceUnavailable must be used according to the specification. https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/p2p-interface.md#blobssidecarsbyrange-v1

> Peers that are unable to reply to block requests within the MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS epoch range SHOULD respond with error code 3: ResourceUnavailable. Such peers that are unable to successfully reply to this range of requests MAY get descored or disconnected at any time.

This is not what's happening here. This error means that our logic got corrupted or that our DB went out of sync. Both cases are considered "internal errors" so should not use ResourceUnavailable

dapplion: Apply both comments to next file too

twoeths:
implemented in #4891

PR #4891, opened 2022-12-08
De-duplicate ReqResp ByRange code

dapplion:
**Motivation**

- PR added a slightly different implementation than blocks by range https://github.com/ChainSafe/lodestar/pull/4865

Given that both do fundamentally the same it's possible to de-duplicate the logic and make it consistent

**Description**

- De-duplicate ReqResp ByRange code
- Handle missing hot block / blob as in https://github.com/ChainSafe/lodestar/pull/4889


twoeths (approved):
lgtm!

PR #4892, opened 2022-12-08
Merge preset params into a single all forks file

dapplion:
**Motivation**

- From https://github.com/ChainSafe/lodestar/pull/4890 noted that params are split across fork files. Which is un-necessary and would look simpler in a single file.

**Description**

- Merge preset params into a single all forks file


wemeetagain:
pls also update the package subpath exports

wemeetagain (approved):
:rocket:

PR #4893, opened 2022-12-08
Add mock EL support to sim tests

nazarhussain:
**Motivation**

Add Mock EL support to Sim tests. 

**Description**

To test some features which are not yet present on EL we need to have EL Mock enabled. 

Run all sim tests .


packages/cli/test/simulation/multi_fork.test.ts
```diff
@@ -53,10 +53,11 @@ const env = SimulationEnvironment.initWithDefaults(
     },
   },
   [
-    {id: "node-1", cl: CLClient.Lodestar, el: ELClient.Geth, keysCount: 32, mining: true},
+    {id: "node-1", cl: CLClient.Lodestar, el: ELClient.Mock, keysCount: 32, mining: true},
```
dapplion: While fine to test now if you need, I would not run any Mock in the main test run. It breaks the realism of this sim test for no benefit. Best to add a second file that runs only with Mock to be taken over by eip-4844 runs

nazarhussain: Yes definitely, I was added here to test the changes on the CI.

dapplion (approved):
LGTM!

@nazarhussain I've edited your PR to run each sim test in a separate CI step. This is crucial for better DX on knowing what failed quickly.

PR #4897, opened 2022-12-11
Add scrape_location as default label to beacon and validator metrics

### LINKED ISSUE #4809: Dashboard metrics using `scrape_location` label in expression are broken
**Problem**

At the moment all dashboard metrics which use the label `scrape_location` in the expression are broken because this label is not set anywhere. Currently the expected label values used in the dashboards are either `validator` or `beacon`.

**Possible solutions**

I see three options how this could be fixed

1. set `scrape_location` as [default label](https://github.com/siimon/prom-client#default-labels-segmented-by-registry)
2. use `job` name or `instance` to differentiate between beacon node and validator
3. set it in `prometheus.yml` file
```
    static_configs:
      - targets: ["localhost:8008"]
        labels:
          scrape_location: "beacon"
```

In my opinion option 1. is the best solution because it does not rely on the deployment configuration.

**Screenshots**

image

nflaig:
**Motivation**

Fixes metrics that use the `scrape_location` label and makes it independent from the deployment configuration i.e. it should not be required to set this label in prometheus configuration file which is currently also not the case, e.g. see [eth-docker lodestar jobs](https://github.com/eth-educators/eth-docker/blob/main/prometheus/ls-prom.yml).

**Description**

Adds the `scrape_location` as default label to beacon and validator node. See [prom-client default labels](https://github.com/siimon/prom-client#default-labels-segmented-by-registry) for more details about default labels.

Closes #4809

### DIFF
--- a/packages/beacon-node/src/metrics/metrics.ts
+++ b/packages/beacon-node/src/metrics/metrics.ts
@@ -33,6 +33,9 @@ export function createMetrics(
     lodestar.unhandledPromiseRejections.inc();
   });
 
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  register.setDefaultLabels({scrape_location: "beacon"});
+
   collectNodeJSMetrics(register);
 
   // Merge external registries
--- a/packages/beacon-node/test/unit/metrics/beacon.test.ts
+++ b/packages/beacon-node/test/unit/metrics/beacon.test.ts
@@ -12,10 +12,10 @@ describe("BeaconMetrics", () => {
     expect(metricsAsText).to.not.equal("");
 
     // check updating beacon-specific metrics
-    await expect(metrics.register.getSingleMetricAsString("libp2p_peers")).eventually.include("libp2p_peers 0");
+    await expect(metrics.register.getSingleMetricAsString("libp2p_peers")).eventually.match(/libp2p_peers({.+})? 0/);
     metrics.peers.set(1);
-    await expect(metrics.register.getSingleMetricAsString("libp2p_peers")).eventually.include("libp2p_peers 1");
+    await expect(metrics.register.getSingleMetricAsString("libp2p_peers")).eventually.match(/libp2p_peers({.+})? 1/);
     metrics.peers.set(20);
-    await expect(metrics.register.getSingleMetricAsString("libp2p_peers")).eventually.include("libp2p_peers 20");
+    await expect(metrics.register.getSingleMetricAsString("libp2p_peers")).eventually.match(/libp2p_peers({.+})? 20/);
   });
 });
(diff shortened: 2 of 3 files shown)


dapplion (changes requested):
Thank you so much for the contribution!

First of all I want to help others consume our dashboards. However adding a default label to all metrics has a non-negligible cost on scrape I'm not willing to pay for given that this happens on the main thread.

Other ideas I got:
- Add labels on eth-docker setup. Metrics are purely additive and we could push for that standard to differentiate process metrics. Probably other clients have the same issue: for example for Go GC metrics
- In our dashboard set the metric to a negative (scrape_location != validator instead of scrape_location == beacon)
- Add a doc on notes on how to consume our dashboards with a list of expected labels and a proposed setup

nflaig:
> a default label to all metrics has a non-negligible cost on scrape I'm not willing to pay for given that this happens on the main thread

I already had a feeling you would say that, especially because the label is not used by a lot of dashboard metrics at the moment

> Add labels on eth-docker setup. Metrics are purely additive and we could push for that standard to differentiate process metrics. Probably other clients have the same issue: for example for Go GC metrics

I will look into this, they have specific prom config per client so it should not be a problem. Also this needs to be added to the prom config files in lodestar repository itself

> In our dashboard set the metric to a negative (scrape_location != validator instead of scrape_location == beacon)

could be a solution but I generally don't like doing this, as it would become to unspecific if there are more `scrape_location` labels in the future

> Add a doc on notes on how to consume our dashboards with a list of expected labels and a proposed setup

Definitely, will do the PR on eth-docker and then see how and where this could be documented

nflaig:
@dapplion  just as a follow up , I quickly checked other dashboards, the [prysm dashboard](https://raw.githubusercontent.com/metanull-operator/eth2-grafana/master/eth2-grafana-dashboard-single-source-beacon_node.json) relies on the job name to differentiate the metrics. I think this is slightely better than adding another label, just need to ensure the jobs are properly named, could use the same job name `validator` and `beacon node` to use the same values as prysm.

image

[Another prysm dashboard](https://raw.githubusercontent.com/ynager/grafana-eth-staking/main/dashboard.json) adds another `name` label, so even the within the different prysm dashboards there seem to be no standardization but to be fair those all come from different repositories.

image

dapplion:
Yeah it's a bit chaotic. I tend to base on this guides for sanity https://prometheus.io/docs/concepts/jobs_instances/ from those, what do you think is the best set of label names? In our case for those hosts the validator and beacon are in the same host that's why we use scrape_location. However would be more consistent to just use job

nflaig:
I think just using the existing `job` label is the most inituititve and best solution as this makes it easy for developers/operators to understand where the label is coming from. I also think it is good to have a separate job per service.

About the label values, I think calling the jobs just `beacon` and `validator` is fine and would be inline with the lodestar cli args.

PR #4907, opened 2022-12-13
Only register active or pending validators to builder

### LINKED ISSUE #4800: Lodestar should only submit `active` and `pending` validators during builder registration
Hi all! I'm the maintainer of the Rocket Pool Smartnode stack and am updating our Goerli beta with the latest Lodestar in prep for a mainnet release soon. I ran into a bug during testing we saw with most of the other CL clients.

**Describe the bug**

The bug is simple - according to the official [builder API spec](https://ethereum.github.io/builder-specs/#/Builder/registerValidator) you should only send `active` or `pending` validators to the external builder during registration. In other words, you can't send it `slashed` or `exited` validators. Doing so causes the following output in MEV-Boost:

```
rocketpool_mev-boost  | time="2022-11-23T05:15:12Z" level=info msg="http: POST /eth/v1/builder/validators 502" duration=0.161919 method=POST path=/eth/v1/builder/validators status=502 version=v1.4.0

rocketpool_mev-boost  | time="2022-11-23T05:21:36Z" level=warning msg="error calling registerValidator on relay" error="HTTP error response: 400 / {\"code\":400,\"message\":\"failed to verify validator signature for 0x9188c03e733ef2ad8bf778480e0cda1f76eb8166a27e9f12c73ee6ec294f29238b72c01e139a3370afe7ce7ec32b94ef\"}\n{\"code\":400,\"message\":\"not a known validator: 0x95901b2ab0202da7859b1f279fdd0015162e56b5500f2a6614fb9d4c554c8e4099e96ccf32f1a47b568efd55411583a5\"}\n" method=registerValidator numRegistrations=3 ua="node-fetch/1.0 (+https://github.com/bitinn/node-fetch)" url="https://relay-goerli.edennetwork.io/eth/v1/builder/validators?id=rocketpool" version=v1.4.0
```

and a corresponding error message in Lodestar's VC:

```
rocketpool_validator  | Nov-23 05:28:00.225[]                error: Failed to register validator registrations with builder epoch=137170 Bad Gateway: Bad Gateway: no successful relay response
rocketpool_validator  | Error: Bad Gateway: Bad Gateway: no successful relay response
rocketpool_validator  |     at HttpClient.requestWithBody (file:///usr/app/packages/api/src/utils/client/httpClient.ts:258:15)
rocketpool_validator  |     at runMicrotasks (<anonymous>)
rocketpool_validator  |     at processTicksAndRejections (node:internal/process/task_queues:96:5)
rocketpool_validator  |     at HttpClient.request (file:///usr/app/packages/api/src/utils/client/httpClient.ts:132:12)
rocketpool_validator  |     at Object.request [as registerValidator] (file:///usr/app/packages/api/src/utils/client/client.ts:82:9)
rocketpool_validator  |     at registerValidator (file:///usr/app/packages/validator/src/services/prepareBeaconProposer.ts:110:11)
rocketpool_validator  |     at Clock.runAtMostEvery (file:///usr/app/packages/validator/src/util/clock.ts:96:7)
``` 

The example validator in question here is [0x95901b2ab0202da7859b1f279fdd0015162e56b5500f2a6614fb9d4c554c8e4099e96ccf32f1a47b568efd55411583a5](https://prater.beaconcha.in/validator/0x95901b2ab0202da7859b1f279fdd0015162e56b5500f2a6614fb9d4c554c8e4099e96ccf32f1a47b568efd55411583a5#attestations), which voluntarily exited long ago.

As I mentioned, most of the other CL clients had the same problem. We [created an issue](https://github.com/flashbots/mev-boost/issues/245) on Flashbots' mev-boost repo to track it. Lighthouse, Nimbus, Prysm, and Teku have resolved it - links to each of their own issues is provided in the above.

**Expected behavior**

Lodestar should only register validators that are in one of the following states:
```
- pending_initialized
- pending_queued
- active_ongoing
- active_exiting
- active_slashed
```

**Steps to Reproduce**

1. Make a validator
2. Voluntarily exit it
3. Observe the above behavior

- OS: Any
- Version: v1.2.1 via the Docker container
- Branch: ---
- Commit hash: ---

twoeths:
**Motivation**

Only register active or pending validators to builder

**Description**
- Make sure we don't send duplicate validators to builder
- Filter active or pending validators

Closes #4800

### DIFF
--- a/packages/beacon-node/src/api/impl/validator/index.ts
+++ b/packages/beacon-node/src/api/impl/validator/index.ts
@@ -6,8 +6,6 @@ import {
   attesterShufflingDecisionRoot,
   getBlockRootAtSlot,
   computeEpochAtSlot,
-  isActiveValidator,
-  isPendingValidator,
 } from "@lodestar/state-transition";
 import {GENESIS_SLOT, SLOTS_PER_EPOCH, SLOTS_PER_HISTORICAL_ROOT, SYNC_COMMITTEE_SUBNET_SIZE} from "@lodestar/params";
 import {Root, Slot, ValidatorIndex, ssz, Epoch} from "@lodestar/types";
@@ -25,6 +23,7 @@ import {validateSyncCommitteeGossipContributionAndProof} from "../../../chain/va
 import {CommitteeSubscription} from "../../../network/subnets/index.js";
 import {ApiModules} from "../types.js";
 import {RegenCaller} from "../../../chain/regen/index.js";
+import {getValidatorStatus} from "../beacon/state/utils.js";
 import {computeSubnetForCommitteesAtSlot, getPubkeysForIndices} from "./utils.js";
 
 /**
@@ -659,17 +658,22 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
       // Spec: https://ethereum.github.io/builder-specs/#/Builder/registerValidator
       const headState = chain.getHeadState();
       const currentEpoch = chain.clock.currentEpoch;
-      const seenValidators = new Set<ValidatorIndex>();
 
       const filteredRegistrations = registrations.filter((registration) => {
         const {pubkey} = registration.message;
         const validatorIndex = headState.epochCtx.pubkey2index.get(pubkey);
-        if (validatorIndex === undefined || seenValidators.has(validatorIndex)) return false;
+        if (validatorIndex === undefined) return false;
 
-        seenValidators.add(validatorIndex);
         const validator = headState.validators.getReadonly(validatorIndex);
-
-        return isActiveValidator(validator, currentEpoch) || isPendingValidator(validator, currentEpoch);
+        const status = getValidatorStatus(validator, currentEpoch);
+        return (
+          status === "active" ||
+          status === "active_exiting" ||
+          status === "active_ongoing" ||
+          status === "active_slashed" ||
+          status === "pending_initialized" ||
+          status === "pending_queued"
+        );
       });
 
       return chain.executionBuilder?.registerValidator(filteredRegistrations);
--- a/packages/beacon-node/src/api/impl/validator/index.ts
+++ b/packages/beacon-node/src/api/impl/validator/index.ts
@@ -657,16 +657,21 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
     async registerValidator(registrations) {
       // should only send active or pending validator to builder
       // Spec: https://ethereum.github.io/builder-specs/#/Builder/registerValidator
+      const headState = chain.getHeadState();
+      const currentEpoch = chain.clock.currentEpoch;
+      const seenValidators = new Set<ValidatorIndex>();
+
       const filteredRegistrations = registrations.filter((registration) => {
         const {pubkey} = registration.message;
-        const headState = chain.getHeadState();
         const validatorIndex = headState.epochCtx.pubkey2index.get(pubkey);
-        if (validatorIndex === undefined) return false;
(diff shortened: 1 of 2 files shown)


dapplion:
@tuyennhv How often does the validator call this endpoint? Could it become a performance issue?

twoeths:
> @tuyennhv How often does the validator call this endpoint? Could it become a performance issue?

@dapplion it's called once per epoch, at validator side we cache the registration data already and it has been in prod for a while. Do you think we need to limit to some max number of validators here (this is mostly for a public node)? If yes we need to inspect all apis and apply that. I don't see any performance issue except for that

packages/beacon-node/src/api/impl/validator/index.ts
```diff
@@ -653,7 +655,24 @@ export function getValidatorApi({chain, config, logger, metrics, network, sync}:
...
+      // should only send active or pending validator to builder
+      // Spec: https://ethereum.github.io/builder-specs/#/Builder/registerValidator
+      const headState = chain.getHeadState();
+      const currentEpoch = chain.clock.currentEpoch;
+      const seenValidators = new Set<ValidatorIndex>();
+
+      const filteredRegistrations = registrations.filter((registration) => {
+        const {pubkey} = registration.message;
+        const validatorIndex = headState.epochCtx.pubkey2index.get(pubkey);
+        if (validatorIndex === undefined || seenValidators.has(validatorIndex)) return false;
```
g11tech: seems like `seenValidators` is only deduping the request from what it seems and the purpose of `isActiveValidator`/`isPendingValidator` is to filter out exited/slashed validators?

g11tech:
> > @tuyennhv How often does the validator call this endpoint? Could it become a performance issue?
> 
> @dapplion it's called once per epoch, at validator side we cache the registration data already and it has been in prod for a while. Do you think we need to limit to some max number of validators here (this is mostly for a public node)? If yes we need to inspect all apis and apply that. I don't see any performance issue except for that

i can't see how we  can have duplicate validators in the same request since the request originated from `prepareBeaconProposer` shouldn't have duplicates.
Also the validators per request are already limited as to `const REGISTRATION_CHUNK_SIZE = 512;` as `prepareBeaconProposer` chunkifies this

packages/beacon-node/src/api/impl/validator/index.ts
g11tech: i don't think we need `seenValidators` processing as this particular requests would/should not contain any duplicates

Also should the `ValidatorStatus` be generated/maintained as part of CachedBeaconState generation? that would be optimized to generate and keep it as well as serve it to the validator client, which can also track/logout the groupings of the validators keys w.r.t. statues

twoeths:
> i can't see how we can have duplicate validators in the same request since the request originated from prepareBeaconProposer shouldn't have duplicates.

@g11tech yeah it should not have duplicate registrations if it comes from lodestar validator client. Having beacon api double check that could help 2 scenarios:
- a public node where someone could send whatever data to our api
- another validator client works with a lodestar beacon node

since we haven't applied this for other apis I'm ok to drop it for now

packages/beacon-node/src/api/impl/validator/index.ts
twoeths: I dropped `seenValidators`. Since this really goes with `ValidatorStatus` in the api spec, I use `getValidatorStatus()` util to avoid adding another function to check for pending validator.

Regarding adding `ValidatorStatus` to `CachedBeaconState`, it'd introduce some memory increase because we need to add that for all validators, I don't see the need for now

dapplion (approved):
LGTM! Concerns addressed

g11tech (approved):
lgtm!

PR #4908, opened 2022-12-13
Implement light-client spec tests

dapplion:
**Motivation**

Implement `light-client` spec tests from https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/light_client

**Description**

- [x] `single_merkle_proof`
- [x] `sync`: Re-used new code in actual light-client class
- [x] `update_ranking`


packages/light-client/src/spec/validateLightClientBootstrap.ts
```diff
@@ -0,0 +1,26 @@
...
+  if (
+    !isValidMerkleBranch(
+      ssz.altair.SyncCommittee.hashTreeRoot(bootstrap.currentSyncCommittee),
+      bootstrap.currentSyncCommitteeBranch,
+      CURRENT_SYNC_COMMITTEE_DEPTH,
+      CURRENT_SYNC_COMMITTEE_INDEX,
+      bootstrap.header.stateRoot
+    )
+  ) {
+    throw Error("Invalid next sync committee merkle branch");
```
dadepo: I believe the error message should be about the current sync committee and not the next?

packages/light-client/src/spec/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+export const ZERO_HASH = new Uint8Array(32);
+export const ZERO_PUBKEY = new Uint8Array(48);
+export const ZERO_SYNC_COMMITTEE = ssz.altair.SyncCommittee.defaultValue();
+export const ZERO_NEXT_SYNC_COMMITTEE_BRANCH = Array.from({length: NEXT_SYNC_COMMITTEE_DEPTH}, () => ZERO_HASH);
+export const ZERO_HEADER = ssz.phase0.BeaconBlockHeader.defaultValue();
+export const ZERO_FINALITY_BRANCH = Array.from({length: FINALIZED_ROOT_DEPTH}, () => ZERO_HASH);
+/** From https://notes.ethereum.org/@vbuterin/extended_light_client_protocol#Optimistic-head-determining-function */
+const SAFETY_THRESHOLD_FACTOR = 2;
+
+export function sumBits(bits: BitArray): number {
```
dadepo, line 15: This function is defined in 3 different places across the code base now. Here, in the beacon node package, and in the light client src. Might be good to have it only in the util package.

packages/beacon-node/src/chain/lightClient/index.ts
```diff
@@ -534,9 +540,29 @@ export class LightClientServer {
...
+
+      const nextBestUpdate: LightClientUpdateSummary = {
+        activeParticipants: sumBits(syncAggregate.syncCommitteeBits),
+        attestedHeaderSlot: attestedData.attestedHeader.slot,
+        signatureSlot,
+        // The actual finalizedHeader is fetched below. To prevent a DB read we approximate the actual slot.
+        // If update is not finalized finalizedHeaderSlot does not matter (see is_better_update), so setting
+        // to zero to set it some number.
+        finalizedHeaderSlot: attestedData.isFinalized
+          ? computeStartSlotAtEpoch(attestedData.finalizedCheckpoint.epoch)
```
dadepo, line 555: is it safe to assume the finalised header slot will always be the first slot at the epoch?

packages/light-client/src/spec/validateLightClientUpdate.ts
```diff
@@ -0,0 +1,133 @@
...
+  update: altair.LightClientUpdate,
+  syncCommittee: SyncCommitteeFast
+): void {
+  // Verify sync committee has sufficient participants
+  if (sumBits(update.syncAggregate.syncCommitteeBits) < MIN_SYNC_COMMITTEE_PARTICIPANTS) {
+    throw Error("Sync committee has not sufficient participants");
+  }
+
+  // Sanity check that slots are in correct order
+  if (update.signatureSlot <= update.attestedHeader.slot) {
```
dadepo, line 30: The spec also include the `current_slot >= update.signature_slot` assertion. Is it a conscious decision to not include that?

packages/beacon-node/test/spec/presets/light_client/sync.ts
```diff
@@ -0,0 +1,207 @@
...
+        fromHex(testcase.meta.genesis_validators_root)
+      );
+
+      const lightClientOpts = {
+        allowForcedUpdates: true,
+        updateHeadersOnForcedUpdate: true,
+      };
+      const lightClient = new LightclientSpec(config, lightClientOpts, testcase.bootstrap);
+
+      // fromHex(testcase.meta.trusted_block_root)
```
dadepo: comment can be removed.

packages/light-client/src/index.ts
```diff
@@ -239,12 +180,12 @@ export class Lightclient {
...
 
   getHead(): phase0.BeaconBlockHeader {
-    return this.head.header;
+    return this.lightclientSpec.store.optimisticHeader;
   }
 
   /** Returns header since head may change during request */
   async getHeadStateProof(paths: JsonPath[]): Promise<{proof: TreeOffsetProof; header: phase0.BeaconBlockHeader}> {
-    const header = this.head.header;
+    const header = this.lightclientSpec.store.optimisticHeader;
```
dadepo: `this.getHead()` can also be used here as it was used in L238

packages/light-client/src/spec/validateLightClientUpdate.ts
dapplion: Yes, to remove the dependency on the clock from this function. That check is done on the caller of this function

packages/beacon-node/src/chain/lightClient/index.ts
dapplion: In this very specific case, IMO yes. That `finalizedHeaderSlot` is used only to break ties in the is_better_update function. If the finalized checkpoint advances is extremely unlikely that the referenced block's slot does not increase. Even if it didn't the consequence is just persisting a slightly different update under some circumstances

packages/light-client/src/spec/utils.ts
dapplion: Such a tinny function, which is just an alias for readability on the SSZ method. I think it's better to have 2 extra lines than a random function in the utils package.

g11tech (approved):
lgtm!

PR #4916, opened 2022-12-16
Make withdrawal probes bounded

g11tech:
**Motivation**

Spec 1.3.0-alpha 2 
- https://github.com/ethereum/consensus-specs/blob/v1.3.0-alpha.2/specs/capella/beacon-chain.md

makes withdrawal probes bounded. This PR implements the same for upcoming withdrawals testnet.


twoeths (approved):

twoeths:
we should merge this first which skip eip4844 spec tests, then https://github.com/ChainSafe/lodestar/pull/4917

g11tech:
> we should merge this first which skip eip4844 spec tests, then #4917

yes, we can remove the corresponding code from #4917 and focus on passing the eip4844 spec vector fixes would be great

PR #4922, opened 2022-12-19
Find more topic peers when subscribeAllSubnets

### LINKED ISSUE #4921: Lack topic peers when subscribe to all subnets
**Describe the bug**

Right now we have `subscribeToAllSubnets` flag but it's quite random to pick peers with required subnets (since we only prioritize committee subnets". In the worse scenario, a node with that flag has very few topic peers

Screen Shot 2022-12-19 at 11 21 16

on another view

Screen Shot 2022-12-19 at 11 21 33

it's because most of the peers have 0 long lived subnets

Screen Shot 2022-12-19 at 11 22 26

**Expected behavior**

When we have `subscribeToAllSubnets` flag, we should:
- subscribe to all subnets (already done)
- set to enr `attnets` (already done)
- actively search for peers with all subnets as if we have >= 64 validators connect to our node

twoeths:
**Motivation**

- If we enable `subscribeAllSubnets` flag, we want the node to have some certain topic/mesh peers in each subnet, but right now we don't actively search for it

**Description**

- If we enable that flag, act like we have `ATTESTATION_SUBNET_COUNT` validators connecting to our node, i.e. actively search for nodes subscribing to requested subnets (which are all subnets in this case)
- This only happens on mainnet, on goerli I think there are a lot of peers subscribing to many/all subnets so it does not happen

Closes #4921

### DIFF
--- a/packages/beacon-node/src/network/subnets/attnetsService.ts
+++ b/packages/beacon-node/src/network/subnets/attnetsService.ts
@@ -54,7 +54,15 @@ export class AttnetsService implements IAttnetsService {
     private readonly metadata: MetadataController,
     private readonly logger: ILogger,
     private readonly opts?: SubnetsServiceOpts
-  ) {}
+  ) {
+    // if subscribeAllSubnets, we act like we have >= ATTESTATION_SUBNET_COUNT validators connecting to this node
+    // so that we have enough subnet topic peers, see https://github.com/ChainSafe/lodestar/issues/4921
+    if (this.opts?.subscribeAllSubnets) {
+      for (let subnet = 0; subnet < ATTESTATION_SUBNET_COUNT; subnet++) {
+        this.committeeSubnets.request({subnet, toSlot: Infinity});
+      }
+    }
+  }
 
   start(): void {
     this.chain.emitter.on(ChainEvent.clockSlot, this.onSlot);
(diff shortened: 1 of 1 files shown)


dapplion (changes requested):
Why is using the randomSubscriptionSloLen() here necessary?

I think this branch should achieve the same with simpler strategy https://github.com/ChainSafe/lodestar/compare/tuyen/more_topic_peers_subscribeAllSubnets-2?expand=1

twoeths:
> Why is using the randomSubscriptionSloLen() here necessary?

@dapplion it's really not necessary, just want to have same code to real random subscriptions, I changed to `Infinity`

dapplion:
@tuyennhv see in my branch I'm using the existing TTL infra and setting it to infinity. Do you think that approach won't work? See diff between both branches https://github.com/ChainSafe/lodestar/compare/tuyen/more_topic_peers_subscribeAllSubnets...tuyen/more_topic_peers_subscribeAllSubnets-2?expand=1

twoeths:
> @tuyennhv see in my branch I'm using the existing TTL infra and setting it to infinity. Do you think that approach won't work? See diff between both branches https://github.com/ChainSafe/lodestar/compare/tuyen/more_topic_peers_subscribeAllSubnets...tuyen/more_topic_peers_subscribeAllSubnets-2?expand=1

@dapplion  oh I see your idea now, yes that should work, merged that branch to this PR

dapplion (approved):

twoeths:
Tested on a mainnet node 

Screen Shot 2022-12-20 at 13 14 01

PR #4923, opened 2022-12-19
Ensure BLSToExecutionChange block inclusions are safe

dapplion:
**Motivation**

BLSToExecutionChange on block processed are expected to have a signature domain at the state's epoch. Since we don't want to re-validate the BLS signature on block inclusion from the op-pool, we must track for what epoch that object was valid.

See related proposal to allow changes through multiple forks: https://github.com/ethereum/consensus-specs/pull/3168

**Description**

- Append `signature_epoch` to data structure.
- Prune on clock fork advance


g11tech (approved):
lgtm!

PR #4819, opened 2022-11-28
Extract out transport layer for light client

dadepo:
**Description**

This PR is currently in draft. The purpose is to showcase the direction of changes based on our last meeting. It contains the following changes:

- Extract a transport layer with a Rest implementation. The Gossip implementation is already being worked on in the branch [here](https://github.com/ChainSafe/lodestar/tree/dadepo/add_gossip_transport)
- Added ability to drive an EL.

With the changes in this PR, we will have a light client implementation that will be able to drive an EL

What I would like to do before updating the PR to ready for review

- [ ] Add integration test
- [x] Test on goerli/sepolia that the light client can indeed drive an EL successfully
- [ ] Address implementation TODOs

PR #4925, opened 2022-12-19
Move Lightclient REST transport to separate class

dapplion:
**Motivation**

- PR https://github.com/ChainSafe/lodestar/pull/4819 is blocked on extracting execution engine package

We don't need to wait for that to happen to apply the LightClientTransport pattern now. This PRs moves the REST logic into a separate class.

**Description**

This move is very easy because the existing lightclient implements assumes that transport is REST only. Once we bring p2p, things will get **much** more complex.

Some open questions:
- get proof capability must be dropped from a generic transport since current p2p does not support that
- the p2p transport requires as dependencies: ReqResp, Gossip, discv5, Peer managment, etc.
- the transport instance of p2p requires access to the lightclient state to respond to ResResp requests. This creates a ciruclar dependency that makes the pattern `const lightclient = new Lightclient(lightclientTransportP2p)` very ugly

### DIFF
--- a/packages/beacon-node/test/e2e/chain/lightclient.test.ts
+++ b/packages/beacon-node/test/e2e/chain/lightclient.test.ts
@@ -1,11 +1,14 @@
 import {expect} from "chai";
 import {IChainConfig} from "@lodestar/config";
-import {ssz} from "@lodestar/types";
-import {fromHexString, toHexString} from "@chainsafe/ssz";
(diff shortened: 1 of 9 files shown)


dadepo (approved):

wemeetagain (approved):
Lgtm, we can figure out the proof thing later.
Maybe we can just have another 'transport' for providing proofs

g11tech (approved):

PR #4926, opened 2022-12-19
Track BlsToExecutionChange fork signature instead of head's fork

dapplion:
**Motivation**

- After reviewing https://github.com/ChainSafe/lodestar/pull/4923 we should not assume the head's fork is correct but instead use the "submitted fork" as correct

**Description**

@g11tech Let's sit on this one for some hours to ensure it feels right.

- Track BlsToExecutionChange fork signature instead of head's fork
- TODO: beacon API should provide with what fork that message was signed

### DIFF
--- a/packages/beacon-node/src/api/impl/beacon/pool/index.ts
+++ b/packages/beacon-node/src/api/impl/beacon/pool/index.ts
@@ -99,11 +99,14 @@ export function getBeaconPoolApi({
     async submitPoolBlsToExecutionChange(blsToExecutionChanges) {
       const errors: Error[] = [];
 
+      // TODO CAPELLA: Which fork should assume to validate submitted execution changes?
+      const signatureFork = chain.config.getForkName(chain.clock.currentSlot);
+
       await Promise.all(
         blsToExecutionChanges.map(async (blsToExecutionChange, i) => {
           try {
-            const {signatureEpoch} = await validateBlsToExecutionChange(chain, blsToExecutionChange);
-            chain.opPool.insertBlsToExecutionChange(blsToExecutionChange, signatureEpoch);
+            await validateBlsToExecutionChange(chain, blsToExecutionChange, signatureFork);
+            chain.opPool.insertBlsToExecutionChange(blsToExecutionChange, signatureFork);
             await network.gossip.publishBlsToExecutionChange(blsToExecutionChange);
           } catch (e) {
             errors.push(e as Error);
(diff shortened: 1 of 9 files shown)


packages/beacon-node/src/chain/opPools/utils.ts
```diff
@@ -63,7 +63,7 @@ export function isValidBlsToExecutionChangeForBlockInclusion(
   //    the fork it was valid, is still the current fork
   //
   // TODO CAPELLA: Ensure that the fork at witch the signature was verified is the same as the state's fork
-  if (signedBLSToExecutionChange.signatureEpoch !== computeEpochAtSlot(state.slot)) {
+  if (signedBLSToExecutionChange.signatureForkSeq !== state.config.getForkSeq(state.slot)) {
```
twoeths, line 66: this is correct 👍

packages/beacon-node/src/api/impl/beacon/pool/index.ts
```diff
@@ -99,11 +99,14 @@ export function getBeaconPoolApi({
     async submitPoolBlsToExecutionChange(blsToExecutionChanges) {
       const errors: Error[] = [];
 
+      // TODO CAPELLA: Which fork should assume to validate submitted execution changes?
+      const signatureFork = chain.config.getForkName(chain.clock.currentSlot);
```
g11tech, line 103: i think this would be correct fork to assume as we can assume that submitted message is ready to be included in the current slot

g11tech (approved):
lgtm!

dapplion:
@g11tech @tuyennhv do VoluntaryExits have the same issues, where we may include in a block a message from the previous fork and create an invalid block?

twoeths:
> @g11tech @tuyennhv do VoluntaryExits have the same issues, where we may include in a block a message from the previous fork and create an invalid block?

@dapplion it's not an issue for attester slashing, proposer slashing and voluntary exit because they all have `slot` in their data structure, when we get domain to verify signatures we based on that `slot`, not the state slot

`BlsToExecutionChange` does not have `slot` in its data structure so we have to do the check to filter valid signature messages when producing block as in this PR

dapplion:
@tuyennhv Oh right, from spec https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#voluntary-exits
```py
domain = get_domain(state, DOMAIN_VOLUNTARY_EXIT, voluntary_exit.epoch)
```

PR #4927, opened 2022-12-19
Publishes Grafana and Prometheus images for monitoring.

MicahZoltu:
**Motivation**

I run lodestar in a fully dockerized environment and it is non-trivial to add grafana/prometheus at the moment.  This is because configuration of these two images is done at build time rather than runtime, and they aren't published anywhere.  As a user, if I want to add monitoring to my docker hosted Lodestar instance I would need to:
1. clone this repository somewhere
2. setup some configuration in .env files and such
3. docker build
4. docker publish
5. go to my production environment and add new entries for each.

After this change, I can skip to step 5.

**Description**

Prometheus only supports configuration by file, so in order to support runtime configuration via environment we have to switch over to using an entrypoint script. `exec` will make it so `/bin/prometheus` will take over PID 1, which ensures that signals get processed correctly. I picked some reasonable defaults for BEACON_URL and VC_URL, but I'm happy to change them to whatever.

Grafana requires building with a second context because the dashboards folder is outside of the docker folder. This requires a relatively up-to-date version of docker and I'm *hoping* GitHub actions supports it. Users should be able to mount a volume of their choosing targeting `/dashboards` if they want their own dashboards to be retained on top of the default set.

Grafana build was setting the admin password at build time, but that should be set at runtime. I set the default password to `admin` under the assumption that the user will have to go out of their way to expose the service and hopefully they'll change the password when they do.  😬

I added two new build/push steps to each of the GitHub action workflows that follow the tagging convention of `chainsafe/lodestar`. I suspect some other name will be desired like `chainsafe/lodestar-grafana` and I'm happy to change it to whatever.

### DIFF
--- a/.github/workflows/publish-rc.yml
+++ b/.github/workflows/publish-rc.yml
@@ -155,3 +155,18 @@ jobs:
       # Display history to know byte size of each layer
       # Image is available only because of the previous `docker run` command
       - run: docker image history chainsafe/lodestar:${{ needs.tag.outputs.tag }}
+
+      - name: Build and push custom Grafana
+        run: >
+          docker buildx build ./docker/grafana/ --push
+          --file ./docker/grafana/Dockerfile
+          --build-context dashboards=./dashboards
+          --tag chainsafe/lodestar-grafana:${{ needs.tag.outputs.tag }}
+          --platform linux/amd64,linux/arm64
+
+      - name: Build and push custom Prometheus
+        run: >
+          docker buildx build ./docker/prometheus/ --push
+          --file ./docker/prometheus/Dockerfile
+          --tag chainsafe/lodestar-prometheus:${{ needs.tag.outputs.tag }}
+          --platform linux/amd64,linux/arm64
--- a/docker/prometheus/entrypoint.sh
+++ b/docker/prometheus/entrypoint.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+sed -i 's/#BEACON_URL/'"$BEACON_URL"'/' /etc/prometheus/prometheus.yml
+sed -i 's/#VC_URL/'"$VC_URL"'/' /etc/prometheus/prometheus.yml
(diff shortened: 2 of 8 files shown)


docker/prometheus/entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
```
g11tech, line 1: nice replacing urls on starup making it env friendly :+1:

.github/workflows/publish-rc.yml
```diff
@@ -155,3 +155,18 @@ jobs:
       # Display history to know byte size of each layer
       # Image is available only because of the previous `docker run` command
       - run: docker image history chainsafe/lodestar:${{ needs.tag.outputs.tag }}
+
+      - name: Build and push custom Grafana
+        run: >
+          docker buildx build ./docker/grafana/ --push
+          --file ./docker/grafana/Dockerfile
+          --build-context dashboards=./dashboards
+          --tag chainsafe/grafana:${{ needs.tag.outputs.tag }}
```
g11tech: needs appropriate repo created in chainsafe dockerhub cc @philknows  for grafana and prometheus and as @MicahZoltu suggested with some lodestar prefix ( e.g. lodestar-grafana?)

philknows: @wemeetagain has access to create this repo. I'm going to advocate for a prefix like `lodestar-grafana` due to the high potential of others in the ChainSafe namespace using Grafana for their own needs and to minimize any potential confusion.

MicahZoltu: I have renamed the images to have `lodestar-` prefix.

wemeetagain:
the `chainsafe/lodestar-grafana` and `chainsafe/lodestar-prometheus` dockerhub repos should be ready

wemeetagain (approved):
thanks @MicahZoltu

PR #4931, opened 2022-12-22
Run optimistic sync spec tests

dapplion:
**Motivation**

Last remaining item we were not running for spec tests. The `sync` spec tests use the same runner as the `fork_choice` tests but with optimistic sync. It requires to inject execution engine replies ahead of submitting blocks.

**Description**

- Refactor execution engine mock to allow modifying its backend and inject responses, while re-using our production-path logic to manage status responses

to use the mock execution backend with this PR:

```ts
const executionEngineBackend = new ExecutionEngineMockBackend({genesisBlockHash: "some-hash"});
const executionEngine = getExecutionEngineFromBackend(executionEngineBackend, {signal});
// Pass `executionEngine` as module to BeaconChain

// Latter in the test
executionEngineBackend.addPowBlock(powBlock);
executionEngineBackend.addPredefinedPayloadStatus(step.block_hash, step.payload_status);
```

### DIFF
--- a/packages/beacon-node/src/chain/blocks/importBlock.ts
+++ b/packages/beacon-node/src/chain/blocks/importBlock.ts
@@ -15,7 +15,7 @@ import {ChainEvent} from "../emitter.js";
 import {REPROCESS_MIN_TIME_TO_NEXT_SLOT_SEC} from "../reprocess.js";
 import {RegenCaller} from "../regen/interface.js";
 import type {BeaconChain} from "../chain.js";
-import {BlockInputType, FullyVerifiedBlock, ImportBlockOpts} from "./types.js";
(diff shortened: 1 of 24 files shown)


dapplion:
@g11tech The spec for optimistic sync is failing, looking into seems some issue with propagating payload status on the forkchoice could you look into it? To run the spec tests and get helpful info do
```
DEBUG=true yarn test:spec-minimal --grep bellatrix/sync
```

dapplion:
Tests *should* pass now :+1: there was a type error on the JsonRpc mock server not spreading the params array

EDIT: Test `eip4844/sync/optimistic/pyspec_tests/from_syncing_to_invalid` is failing

g11tech (approved):
lgtm!

Issue #4881, opened 2022-12-07
[vc] Investigate late attestations published on mainnet


twoeths:
**Describe the bug**
- This is about the time we publish attestations compared to 1/3 of slot (so -2 means 1s after slot, +2 means 6s after slot
- On goerli it rarely happens

<img width="1291" alt="Screen Shot 2022-12-07 at 16 00 48" src="https://user-images.githubusercontent.com/10568965/206134702-0473c9ff-f98d-40cd-b245-53271d4e8e45.png">

- However some mainnet nodes (v1.2.x) experienced this issue more

<img width="1285" alt="Screen Shot 2022-12-07 at 16 00 27" src="https://user-images.githubusercontent.com/10568965/206134616-a0d5e9b3-a7fc-4de4-a356-c020d00341b6.png">

**Expected behavior**

Attestations should be published not later than 1/3 of slots

twoeths:
This is because `produceAttestationData` being called after 4s (when vc hasn't received new chain head on time)

<img width="1290" alt="Screen Shot 2022-12-07 at 19 32 37" src="https://user-images.githubusercontent.com/10568965/206180250-9bcef3a4-2092-43b8-8c13-a2634b20389c.png">

and produceAttestationData request time sometimes take >1s or even 5s

<img width="1267" alt="Screen Shot 2022-12-07 at 19 34 21" src="https://user-images.githubusercontent.com/10568965/206180543-a709c19d-560e-4e6f-9add-c114e93b9cc8.png">

root cause is the I/O lag issue

twoeths:
We could try http 1.1 keep-alive (or http 2, that's a longer plan) to reuse http connection

philknows:
> This is because `produceAttestationData` being called after 4s (when vc hasn't received new chain head on time)
> 
> <img alt="Screen Shot 2022-12-07 at 19 32 37" width="1290" src="https://user-images.githubusercontent.com/10568965/206180250-9bcef3a4-2092-43b8-8c13-a2634b20389c.png">
> 
> and produceAttestationData request time sometimes take >1s or even 5s
> 
> <img alt="Screen Shot 2022-12-07 at 19 34 21" width="1267" src="https://user-images.githubusercontent.com/10568965/206180543-a709c19d-560e-4e6f-9add-c114e93b9cc8.png">
> 
> root cause is the I/O lag issue

Referencing #4002

twoeths:
[...]

PR #4932, opened 2022-12-22
Mock validator client

twoeths:
**Motivation**

We want to track the I/O lag issue statically

**Description**

- Write a mock vc that produce attestation right at 1/3 of slot
- New sub-command, for example `validator mock --beaconNode http://localhost:9596 --network goerli --metrics true`

part of #4002, #4881


dapplion:
@tuyennhv Can you check if this tool already serves your needs? https://github.com/michaelsproul/blockdreamer

dapplion (changes requested):
I would prefer to not merge such a niche tool an our production CLI, specially if it shows in --help. Please develop as a standalone repo or just use blockdreamer

twoeths:
> I would prefer to not merge such a niche tool an our production CLI, specially if it shows in --help. Please develop as a standalone repo or just use blockdreamer

@dapplion  it's a hidden subcommand so will not be shown in `--help`, we want to issue `produceAttestationData` right at 1/3 of slot (the busiest time of our node) so cannot use `blockdreamer`. Moreover we want to utilize our api client metric to measure the http call time and show it in our grafana dashboard.

I'd keep this branch to see how it work first, will create a repo when we think we really need to.

PR #4934, opened 2022-12-23
Do not include VoluntaryExit messages from old fork

dapplion:
**Motivation**

- Similar reasoning to https://github.com/ChainSafe/lodestar/pull/4926 to not include messages from the op pool with signature from old fork.

**Description**

- Filter out messages by signature fork
- Prune messages from old signature fork

### DIFF
--- a/packages/beacon-node/src/chain/opPools/opPool.ts
+++ b/packages/beacon-node/src/chain/opPools/opPool.ts
@@ -2,6 +2,7 @@ import {
   CachedBeaconStateAllForks,
   CachedBeaconStateCapella,
   computeEpochAtSlot,
+  computeStartSlotAtEpoch,
   getAttesterSlashableIndices,
   isValidVoluntaryExit,
 } from "@lodestar/state-transition";
@@ -173,7 +174,9 @@ export class OpPool {
     phase0.SignedVoluntaryExit[],
     capella.SignedBLSToExecutionChange[]
   ] {
+    const {config} = state;
     const stateEpoch = computeEpochAtSlot(state.slot);
+    const stateFork = config.getForkName(state.slot);
     const toBeSlashedIndices = new Set<ValidatorIndex>();
     const proposerSlashings: phase0.ProposerSlashing[] = [];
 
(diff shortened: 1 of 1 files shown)


packages/beacon-node/src/chain/opPools/opPool.ts
```diff
@@ -316,8 +323,17 @@ export class OpPool {
    * Prune if validator has already exited at or before the finalized checkpoint of the head.
    */
   private pruneVoluntaryExits(headState: CachedBeaconStateAllForks): void {
+    const {config} = headState;
+    const headStateFork = config.getForkSeq(headState.slot);
```
wemeetagain, line 327: Shouldn't we be deleting exits later than the fork based on clock slot? I guess it doesn't much matter if we delete them eventually anyway and they're filtered out during block building.

dapplion: Yes, but using the head state makes the API simpler, and functionally quasi equivalent. The only difference is that exists may be pruned a bit latter in case of fully skipped epochs.

wemeetagain (approved):

twoeths:
@wemeetagain @dapplion my understanding is we can include VoluntaryExit from previous fork in our block. Here's the verify_signagure part of `process_voluntary_exit`

```python
# Verify signature
    domain = get_domain(state, DOMAIN_VOLUNTARY_EXIT, voluntary_exit.epoch)
    signing_root = compute_signing_root(voluntary_exit, domain)
    assert bls.Verify(validator.pubkey, signing_root, signed_voluntary_exit.signature)
```

and `get_domain`:
```python
def get_domain(state: BeaconState, domain_type: DomainType, epoch: Epoch=None) -> Domain:
    """
    Return the signature domain (fork version concatenated with domain type) of a message.
    """
    epoch = get_current_epoch(state) if epoch is None else epoch
    fork_version = state.fork.previous_version if epoch < state.fork.epoch else state.fork.current_version
    return compute_domain(domain_type, fork_version, state.genesis_validators_root)
```

in general we don't have to do the logic in this PR for gossip messages with slot/epoch data inside because it based on that slot/epoch to get fork version to verify signature. We only need that for `SignedBLSToExecutionChange` since there is no slot/epoch there

dapplion:
Oh, right good point. Then we should assert that a voluntary exit in the pool is for state's current or previous fork

PR #4936, opened 2022-12-23
Handle skipped blobs side car for empty blobs in blobsSidecarsByRange

g11tech:
Handle skipped blobs side car for empty blobs in blobsSidecarsByRange
see: 
- https://github.com/ethereum/consensus-specs/pull/3174

### DIFF
--- a/packages/beacon-node/src/network/network.ts
+++ b/packages/beacon-node/src/network/network.ts
@@ -12,6 +12,7 @@ import {altair, eip4844, Epoch, phase0} from "@lodestar/types";
 import {IMetrics} from "../metrics/index.js";
 import {ChainEvent, IBeaconChain, IBeaconClock} from "../chain/index.js";
 import {BlockInput, BlockInputType, getBlockInput} from "../chain/blocks/types.js";
+import {ckzg} from "../util/kzg.js";
 import {INetworkOptions} from "./options.js";
 import {INetwork} from "./interface.js";
 import {ReqRespBeaconNode, ReqRespHandlers, doBeaconBlocksMaybeBlobsByRange} from "./reqresp/index.js";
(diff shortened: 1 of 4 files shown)


packages/beacon-node/src/network/network.ts
```diff
@@ -243,22 +244,53 @@ export class Network implements INetwork {
...
+          if (blobKzgCommitmentsLen !== 0) {
+            throw Error(
+              `Missing blobsSidecar for blockSlot=${block.message.slot} with blobKzgCommitmentsLen=${blobKzgCommitmentsLen}`
+            );
+          }
+          blobsSidecar = {
+            beaconBlockRoot: this.config.getForkTypes(block.message.slot).BeaconBlock.hashTreeRoot(block.message),
+            beaconBlockSlot: block.message.slot,
+            blobs: [],
+            kzgAggregatedProof: ckzg.computeAggregateKzgProof([]),
```
dapplion: Why do you need to create an empty BlobsSidecar?

g11tech: because empty BlobsSidecar may be skipped in the req/resp, thats whats happening in devnet3

dapplion (commented):
Given that this function is growing in complexity would make sense to move it to a file with signature

```ts
function beaconBlocksMaybeBlobsByRange(config, reqResp, peerId, request) {}
```

And unit test it by injecting a mock reqResp that resolves with predefined data

twoeths (approved):

packages/beacon-node/test/unit/network/doBeaconBlocksMaybeBlobsByRange.ts
```diff
@@ -0,0 +1,88 @@
+import sinon, {SinonStubbedInstance} from "sinon";
+import {expect} from "chai";
+import {peerIdFromString} from "@libp2p/peer-id";
+import {ssz, eip4844} from "@lodestar/types";
+import {createIBeaconConfig, createIChainForkConfig, defaultChainConfig} from "@lodestar/config";
+
+import {doBeaconBlocksMaybeBlobsByRange, ReqRespBeaconNode} from "../../../src/network/reqresp/index.js";
+import {BlockInputType} from "../../../src/chain/blocks/types.js";
+
+describe("doBeaconBlocksMaybeBlobsByRange", function () {
```
dapplion, line 10: Please always append `.test.ts` extension to test files

g11tech: ahh thats my bad, here is the fix: https://github.com/ChainSafe/lodestar/pull/4949

PR #4937, opened 2022-12-24
Force usage of Uint8Array.prototype.slice in effectiveBalanceIncrements

### LINKED ISSUE #4390: getEffectiveBalanceIncrementsZeroInactive mutates cache
**Describe the bug**

Function `getEffectiveBalanceIncrementsZeroInactive` says

https://github.com/ChainSafe/lodestar/blob/0de01443f1f4cfa6efc12783c5783f0035e4e29f/packages/state-transition/src/util/balance.ts#L57-L58

But immediately after mutates the balances setting them to zero 

https://github.com/ChainSafe/lodestar/blob/0de01443f1f4cfa6efc12783c5783f0035e4e29f/packages/state-transition/src/util/balance.ts#L66-L67

So the .slice() **MUST** copy the array as it will be mutated right? Note that Buffer.slice() does not copy while Uint8Array.slice() copies data. We must ensure that the data is copied.

@tuyennhv do you think this can be an issue?

**Expected behavior**

Don't mutate cache

Instead do

```ts
    // Buffer.prototype.slice does not copy memory, Enforce Uint8Array usage https://github.com/nodejs/node/issues/28087
    const uint8Array = Uint8Array.prototype.slice.call(data, start, end - 1);
```

dapplion:
**Motivation**

- See https://github.com/ChainSafe/lodestar/issues/4390

Uint8Array.slice copy while Buffer.slice does not copy. Prevent a future footgun by forcing the former.

**Description**

Closes https://github.com/ChainSafe/lodestar/issues/4390

### DIFF
--- a/packages/state-transition/src/cache/epochContext.ts
+++ b/packages/state-transition/src/cache/epochContext.ts
@@ -516,8 +516,9 @@ export class EpochContext {
   }
 
   beforeEpochTransition(): void {
-    // Clone before being mutated in processEffectiveBalanceUpdates
-    this.effectiveBalanceIncrements = this.effectiveBalanceIncrements.slice(0);
+    // Clone (copy) before being mutated in processEffectiveBalanceUpdates
+    // NOTE: Force to use Uint8Array.slice (copy) instead of Buffer.call (not copy)
+    this.effectiveBalanceIncrements = Uint8Array.prototype.slice.call(this.effectiveBalanceIncrements, 0);
   }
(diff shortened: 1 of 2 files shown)


wemeetagain (approved):
Were these the only troublesome cases?

dapplion:
> Were these the only troublesome cases?

I have not done a deep search again, but possibly so since we looked into it at the time of writing the original issue

PR #4942, opened 2022-12-24
Keep at least 10% outbound peers

### LINKED ISSUE #2215: Keep at least 10% of outbound peers
If you run current master with ports open the node will eventually be connected to 100% of inbound peers.

From https://discord.com/channels/595666850260713488/646778960830660623/811507775389696050

> - Is there any reason or issue to keep a certain ratio of inbound vs outbound peers? i.e. any issues with having ~100% of your peers inbound?
> - You'd be much more susceptible to an eclipse attack in that scenario.  Having a minimum number of outbound peers ensures that you chose at least some of your peers.
> - What specific numbers are you targeting in your implmentation?
> - I don’t remember off the top of my head. I think it’s 10% are entirely randomly selected outbound connections.

dapplion:
**Motivation**

To protect against eclipse attacks always keep some outbound peers. Outbound peers are chosen at random from walking the discv5 DHT, while an attacker can easily direct target_peers worth of nodes to connect to us, such that we only listen to attack nodes. Taking over the discv5 DHT is much more difficult, thus the min ratio of outbound peers.

- See https://github.com/ChainSafe/lodestar/issues/2215

**Description**

- Keep at least 10% outbound peers

Closes https://github.com/ChainSafe/lodestar/issues/2215

### DIFF
--- a/packages/beacon-node/src/network/peers/peerManager.ts
+++ b/packages/beacon-node/src/network/peers/peerManager.ts
@@ -406,6 +406,7 @@ export class PeerManager {
         const peerData = this.connectedPeers.get(peer.toString());
         return {
           id: peer,
+          direction: peerData?.direction ?? null,
           attnets: peerData?.metadata?.attnets ?? null,
           syncnets: peerData?.metadata?.syncnets ?? null,
           score: this.peerRpcScores.getScore(peer),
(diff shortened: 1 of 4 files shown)


wemeetagain (approved):
Lgtm

Issue #4657, opened 2022-10-11
node can't find peers


TimDaub:
Hi I have a lodestar client running that seems to never find peers, shall I create a github issue?

```
Oct-12 00:51:12.642[]                 warn: Low peer count peers=0
Oct-12 00:36:41.000[]                 info: Searching peers - peers: 0 - slot: 4891981 (skipped 4891981) - head: 0 0x4d61…9360 - finalized: 0x0000…0000:0
Oct-12 00:36:53.001[]                 info: Searching peers - peers: 0 - slot: 4891982 (skipped 4891982) - head: 0 0x4d61…9360 - finalized: 0x0000…0000:0
Oct-12 00:37:05.000[]                 info: Searching peers - peers: 0 - slot: 4891983 (skipped 4891983) - head: 0 0x4d61…9360 - finalized: 0x0000…0000:0
Oct-12 00:37:17.000[]                 info: Searching peers - peers: 0 - slot: 4891984 (skipped 4891984) - head: 0 0x4d61…9360 - finalized: 0x0000…0000:0
Oct-12 00:37:29.000[]                 info: Searching peers - peers: 0 - slot: 4891985 (skipped 4891985) - head: 0 0x4d61…9360 - finalized: 0x0000…0000:0
 ```
- branch v1.1.0, but no difference
- how I run it: `/lodestar beacon --data-dir=./data --jwt-secret=../datadir/jwt.hex`
- `name -a
Linux ethnode 5.4.0-90-generic #101-Ubuntu SMP Fri Oct 15 20:00:55 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

TimDaub:
As discussed in Discord, I've now:

- Updated to [v1.2.0-rc.1](https://github.com/ChainSafe/lodestar/releases/tag/v1.2.0-rc.1)
- Added `--logLevel debug` and `DEBUG=*`

Here's what I've got and it looks like there is an error reading the peer list.

[cl.txt](https://github.com/ChainSafe/lodestar/files/9778644/cl.txt)

could be this? https://github.com/libp2p/js-libp2p-mdns/blob/36d68fc819316ec7f7a215a38310d90130770e0f/src/compat/querier.ts#L103

TimDaub:
So I checked the logs and as I said, there is this error showing up:

[...]

PR #4957, opened 2022-12-29
Log error if there is no discv5 kad value

twoeths:
**Motivation**
- discv5 can't fetch enrs if there is no kad value
- See https://github.com/ChainSafe/lodestar/issues/4657#issuecomment-1366335998

**Description**

Add a check and log error if there is no kad value

part of #4657

### DIFF
--- a/packages/beacon-node/src/network/peers/discover.ts
+++ b/packages/beacon-node/src/network/peers/discover.ts
@@ -120,7 +120,12 @@ export class PeerDiscovery {
       },
     });
     opts.discv5.bootEnrs.forEach((bootEnr) => this.discv5.addEnr(bootEnr));
-    this.logger.verbose("PeerDiscovery number of bootEnrs", {bootEnrs: opts.discv5.bootEnrs.length});
+    const numBootEnrs = opts.discv5.bootEnrs.length;
+    if (numBootEnrs === 0) {
+      this.logger.error("PeerDiscovery: discv5 has no boot enr");
+    } else {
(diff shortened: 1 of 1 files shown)


packages/beacon-node/src/network/peers/discover.ts
```diff
@@ -220,6 +220,10 @@ export class PeerDiscovery {
    * Request to find peers. First, looked at cached peers in peerStore
    */
   private async runFindRandomNodeQuery(): Promise<void> {
+    if (this.discv5.kadValues().length === 0) {
+      this.logger.error("discv5 has no kad value");
+    }
```
dapplion: This error could get very noisy, so I'm thinking how often this should log. Questions:
- If the kad table has some value at the start, can it get empty after time? I think no
- Then, the only way for kad table to be empty is there are no bootnodes

So, options:
- If the above is true, just logging if there are 0 bootnodes at the start should be equivalent?
- Else, log only on the first run with a `has_warned_empty_kad` state variable
- Or if the kad can become empty, then log once only per time interval the kad became empty

dapplion (changes requested):

packages/beacon-node/src/network/peers/discover.ts
twoeths: Update: log error if there is no boot enr

wemeetagain (approved):

dapplion (approved):

PR #4973, opened 2023-01-04
Remove all intermediary ChainEvents

dapplion:
**Motivation**

BeaconNode source emits events meant for the API through intermediary ChainEvents. That's not necessary since the few users of the internal events can just listen to the events defined as in the API.

- Extends https://github.com/ChainSafe/lodestar/pull/4972

**Description**

Remove all intermediary ChainEvents

### DIFF
--- a/packages/api/src/beacon/routes/events.ts
+++ b/packages/api/src/beacon/routes/events.ts
@@ -5,7 +5,7 @@ import {RouteDef, TypeJson} from "../../utils/index.js";
 
 // See /packages/api/src/routes/index.ts for reasoning and instructions to add new routes
 
-export enum EventType {
+export const enum EventType {
   /**
    * The node has finished processing, resulting in a new head. previous_duty_dependent_root is
    * `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 1) - 1)` and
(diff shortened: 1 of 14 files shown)


packages/beacon-node/test/sim/withdrawal-interop.test.ts
```diff
@@ -350,3 +349,26 @@ describe("executionEngine / ExecutionEngineHttp", function () {
     console.log("\n\nDone\n\n");
   }
 });
+
+async function retrieveCanonicalWithdrawals(bn: BeaconNode, fromSlot: Slot, toSlot: Slot): Promise<number> {
```
dapplion, line 353: @g11tech Can you review if this assertion is good for your test? Re-wrote the previous event based, into an after the fact one

packages/beacon-node/test/sim/withdrawal-interop.test.ts
```diff
@@ -217,6 +219,9 @@ describe("executionEngine / ExecutionEngineHttp", function () {
     // delay a bit so regular sync sees it's up to date and sync is completed from the beginning
     const genesisSlotsDelay = 8;
 
+    // TODO for g11tech: Why 4? Provide rationale for the number
```
dapplion, line 222: @g11tech: Why 4? Provide rationale for the number

twoeths (approved):

packages/beacon-node/test/sim/withdrawal-interop.test.ts
g11tech: yes looks good

packages/beacon-node/test/sim/withdrawal-interop.test.ts
g11tech: from the test length run i noted that if there is no issue, withdrawal blocks come out to be 6, so i chose 4 as safe number incase there are any skipped slots for whatever reasons (sometimes have seen it happening)

g11tech (approved):
lgtm!

packages/api/src/beacon/routes/events.ts
```diff
@@ -5,7 +5,7 @@ import {RouteDef, TypeJson} from "../../utils/index.js";
 
 // See /packages/api/src/routes/index.ts for reasoning and instructions to add new routes
 
-export enum EventType {
+export const enum EventType {
```
nflaig, line 8: @dapplion What was the reason for changing this to a `const` enum besides just the minor reduction in generated code? The problem with const enums are that they are completely removed during compilation which means others using the library can not reuse them, see [discord](https://discord.com/channels/593655374469660673/667606053877317642/1065521781026332682)

wemeetagain: IMO this should be reinstated as an enum

wemeetagain: Also re-exported as necessary to allow for downstream reuse

nflaig: changed in #5042, seems to be re-exported already so we should be fine there

dapplion: Potentially unwanted side-effect from attempting to make EventType from API + ChainEvent work for the same handlers. Happy to revert it to enum

Issue #4680, opened 2022-10-21
Pending spec test tracker


g11tech:
Spec 1.3.0-alpha.0
- https://github.com/ChainSafe/lodestar/pull/4679
- https://github.com/ChainSafe/lodestar/pull/4707

has spec tests for 

- [x] `light_client`
- [x]  `sync` (most likely optimistic sync)
- [x]  add runners for 
  - [x] full_withdrawals" 
  - [x] "partial_withdrawals"
  - [x] "bls_to_execution_change"
  - [x] "withdrawals"
  - [x] "historical_summaries_update"
- [x] capella/random/random/pyspec_tests/randomized_11, capella/sanity/blocks/pyspec_tests, 
- [x] full_withdrawal_in_epoch_transition" (These tests specifically seem to have a full validator withdrawal with a validatorstill in sync committee. Remove once spec tests are fixed )
- [x] seek clarification on "eip4844/transition/core/pyspec_tests/transition_randomized_state" and fix runner or the test gets fixed

These spec runners need to be added into our spec test runner utils.

Completed:
- [x] handle genesis changes for the capella and enable genesis specs for capella

g11tech:
closing as tasks are addressed

PR #4974, opened 2023-01-04
Add historicalSummaries capella onwards

g11tech:
Add historicalSummaries capella onwards

Ref
- https://github.com/ethereum/consensus-specs/pull/3165

TODO:
- [x] get it merge ready when new spec tests are released


packages/state-transition/test/unit/stateTransition.test.ts
```diff
@@ -1,48 +0,0 @@
-import fs from "node:fs";
```
g11tech, line 1: this test was based on beacon state structure from devnet1, since it has changed, is no longer relvant here

packages/beacon-node/test/spec/presets/index.test.ts
```diff
@@ -22,7 +22,16 @@ import {transition} from "./transition.js";
...
-  skippedHandlers: ["full_withdrawals", "partial_withdrawals", "bls_to_execution_change", "withdrawals"],
+  skippedHandlers: [
+    "full_withdrawals",
+    "partial_withdrawals",
+    "bls_to_execution_change",
+    "withdrawals",
+    "historical_summaries_update",
+  ],
+  // This test right now has been escalted as history summary should have been pushed
+  // in the state, but in the expected state as per test, nothing has been pushed
```
g11tech: escalated the test to the CL group, waiting on revert but seems like lodestar is doing the right thing

packages/beacon-node/test/spec/presets/index.test.ts
```diff
@@ -22,7 +22,16 @@ import {transition} from "./transition.js";
 const skipOpts: SkipOpts = {
   skippedForks: [],
   skippedRunners: [],
-  skippedHandlers: ["full_withdrawals", "partial_withdrawals", "bls_to_execution_change", "withdrawals"],
+  skippedHandlers: [
+    "full_withdrawals",
+    "partial_withdrawals",
+    "bls_to_execution_change",
+    "withdrawals",
+    "historical_summaries_update",
```
g11tech: added to pending tracker list 
-https://github.com/ChainSafe/lodestar/issues/4680, 
but other transition tests cover summaries update as well  so we can take it up as a separate task

dapplion (dismissed):
LGMT! 
- But can we merge this one first? https://github.com/ChainSafe/lodestar/pull/4987

packages/beacon-node/test/spec/presets/index.test.ts
g11tech: added the epochfn handler for historical_summaries_update, so its not skipped now :+1:

dapplion (approved):

Issue #4700, opened 2022-10-28
headState does not exist


wemeetagain:
Report from discord: https://discord.com/channels/593655374469660673/743858262864167062/1035585735983644702

`version=v1.1.1/9d07393, commit=9d07393709f6a870f9eeda9bca7a9
5196ca67c68`

```
Oct-28 15:55:18.000[]                error: Node notifier error  headState does not exist
Error: headState does not exist
    at BeaconChain.getHeadState (file:///usr/app/packages/beacon-node/src/chain/chain.ts:319:27)
    at runNodeNotifier (file:///usr/app/packages/beacon-node/src/node/notifier.ts:55:31)
```

We thought this was fixed in v1.1.0:
- https://github.com/ChainSafe/lodestar/issues/4523

twoeths:
logs from when the issue was reported

```
Oct-28 15:55:06.001[]                 info: Searching peers - peers: 13 - slot: 4205375 (skipped 95) - head: 4205280 0x8b1e…8d4e - execution: valid(0xab62…7e3e) - finalized: 0x0e57…6e29:131413
Oct-28 15:55:18.000[]                error: Node notifier error  headState does not exist
Error: headState does not exist
    at BeaconChain.getHeadState (file:///usr/app/packages/beacon-node/src/chain/chain.ts:319:27)
    at runNodeNotifier (file:///usr/app/packages/beacon-node/src/node/notifier.ts:55:31)
Oct-28 15:55:40.250[NETWORK]          info: Subscribed gossip core topics
Oct-28 15:55:40.258[SYNC]             info: Subscribed gossip core topics
```

SeaMonkey82:
~~Still seeing the same error on Version: v1.2.1/cdde39f~~

Slightly different error.
```
Dec-27 13:22:24.008[rest]            error: Req req-13q getStateValidators error  No state found for id 'head'
Error: No state found for id 'head'
    at resolveStateId (file:///home/seamonkey/Downloads/goerli/lodestar/packages/beacon-node/src/api/impl/beacon/state/utils.ts:43:11)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at Object.getStateValidators (file:///home/seamonkey/Downloads/goerli/lodestar/packages/beacon-node/src/api/impl/beacon/state/index.ts:57:21)
[...]

PR #4976, opened 2023-01-05
Review importBlock order of operations

dapplion:
**Motivation**

We noted with @g11tech that the import order of importBlock is not safe since it can import a block into the fork-choice without it being available in the DB. That can cause the node to get stuck if it ever needs to regen, and other potential undefined behavior.

**Description**

ImportBlock order of operations must guarantee that BeaconNode does not end in an unknown state:
1. Persist block to hot DB (pre-emptively)
   - Done before importing block to fork-choice to guarantee that blocks in the fork-choice lwaysare persisted
     in the DB. Otherwise the beacon node may end up in an unrecoverable state. If a block is persisted in the hot
     db but is unknown by the fork-choice, then it will just use some extra disk space. On restart is will be
     pruned regardless.
   - Note that doing a disk write first introduces a small delay before setting the head. An improvement where disk
     write happens latter requires the ability to roll back a fork-choice head change if disk write fails
2. Import block to fork-choice
3. Import attestations to fork-choice
4. Import attester slashings to fork-choice
5. Compute head. If new head, immediately stateCache.setHeadState()
6. Queue notifyForkchoiceUpdate to engine api
7. Add post state to stateCache


packages/fork-choice/src/forkChoice/forkChoice.ts
```diff
@@ -309,7 +309,7 @@ export class ForkChoice implements IForkChoice {
     blockDelaySec: number,
     currentSlot: Slot,
     executionStatus: MaybeValidExecutionStatus
-  ): void {
+  ): ProtoBlock {
```
dapplion, line 312: Make onBlock return the blockSummary to prevent having to consider in the forkchoice that the new block is not found in the fork-choice

packages/beacon-node/src/chain/blocks/importBlock.ts
```diff
@@ -11,12 +11,11 @@ import {ForkChoiceError, ForkChoiceErrorCode, EpochDifference, AncestorStatus} f
 import {ZERO_HASH_HEX} from "../../constants/index.js";
 import {toCheckpointHex} from "../stateCache/index.js";
 import {isOptimisticBlock} from "../../util/forkChoice.js";
-import {ChainEvent} from "../emitter.js";
+import {ChainEvent, ReorgEventData} from "../emitter.js";
 import {REPROCESS_MIN_TIME_TO_NEXT_SLOT_SEC} from "../reprocess.js";
 import {RegenCaller} from "../regen/interface.js";
 import type {BeaconChain} from "../chain.js";
 import {BlockInputType, FullyVerifiedBlock, ImportBlockOpts, AttestationImportOpt} from "./types.js";
-import {PendingEvents} from "./utils/pendingEvents.js";
```
dapplion, line 20: - From https://github.com/ChainSafe/lodestar/pull/4973 all remaining events are safe to be emitted in the order of this PR

twoeths:
@dapplion what happened in https://github.com/ChainSafe/lodestar/issues/4700 is head block existed as head but post state was not in the cache due to an error in the middle, should we add post state to state cache before adding block to forkchoice?

dapplion:
> @dapplion what happened in #4700 is head block existed as head but post state was not in the cache due to an error in the middle, should we add post state to state cache before adding block to forkchoice?

The only `await` points are the DB writes, so all steps after 1. are executed in sequence. It's not possible for anything else to do a request to the cache, so adding to the fork-choice and adding to the cache will happen "atomically" if nothing throws.

If there are not awaits it's more of a question of what is allowed to throw before what.

Regarding the head state keep in mind there's a possibility the head state is not available, so it triggers regen and will be null for a while.

dapplion:
Closing + re-opening to trigger CI, somehow it got stuck

packages/beacon-node/src/chain/blocks/importBlock.ts
wemeetagain: it may be safe from a logic perspective, but we may want to keep this to avoid a future footgun?

Now we will be adding additional work, running all event handlers, interspersed with importBlock, which we want to finish ASAP.

dapplion: It's safe because none of those events trigger side effects, unlike they did before. They are solely consumed by API eventsream clients

wemeetagain (approved):

g11tech (commented):
looks good! reviewing post merge :+1:

PR #4980, opened 2023-01-05
Fix first query discovery delay

dapplion:
**Motivation**

- PR https://github.com/ChainSafe/lodestar/issues/3423 introduced a delay on the first discv5 query

However in practice it delays by 30 seconds which is the peer manager loop interval, since it just cancels the first request

**Description**

- Fix first query discovery delay, by awaiting the param amount, instead of just ignoring the first request

### DIFF
--- a/packages/beacon-node/src/network/peers/discover.ts
+++ b/packages/beacon-node/src/network/peers/discover.ts
@@ -224,7 +224,7 @@ export class PeerDiscovery {
     // See https://github.com/ChainSafe/lodestar/issues/3423
     const msSinceDiscv5Start = Date.now() - this.discv5StartMs;
     if (msSinceDiscv5Start <= this.discv5FirstQueryDelayMs) {
-      await sleep(msSinceDiscv5Start);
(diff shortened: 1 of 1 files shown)


packages/beacon-node/src/network/peers/discover.ts
```diff
@@ -222,8 +222,9 @@ export class PeerDiscovery {
   private async runFindRandomNodeQuery(): Promise<void> {
     // Delay the 1st query after starting discv5
     // See https://github.com/ChainSafe/lodestar/issues/3423
-    if (Date.now() - this.discv5StartMs <= this.discv5FirstQueryDelayMs) {
-      return;
+    const msSinceDiscv5Start = Date.now() - this.discv5StartMs;
+    if (msSinceDiscv5Start <= this.discv5FirstQueryDelayMs) {
+      await sleep(msSinceDiscv5Start);
```
twoeths: should be `await sleep(this.discv5FirstQueryDelayMs - msSinceDiscv5Start);` ?

wemeetagain (approved):

PR #4985, opened 2023-01-06
Clarify Keystore Import Options

ansermino:
**Motivation**

Clarify the options for importing keys.

**Description**

- Clearly explains the two options for importing keys
- Removes unnecessary keystore example (not very helpful for user)
- Adds warning about secret store being ignored if keys are imported at run time

Closes N/A

```sh
git checkout ansermino/keystore-import-docs
pip3 install mkdocs
pip3 install -r docs/requirements.txt
mkdocs serve
```

Screenshot 2023-01-06 at 4 54 31 PM

### DIFF
--- a/docs/usage/validator-management.md
+++ b/docs/usage/validator-management.md
@@ -36,57 +36,41 @@ Validators are represented by a BLS keypair. Use your generated mnemonic from on
 
 To import a validator keystore that was created via one of the methods described above, you must locate the validator keystore JSONs exported by those tools (ex. `keystore-m_12381_3600_0_0_0-1654128694.json`).
 
-Inside the keystore JSON file, you should have an [EIP-2335 conformant keystore file](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2335.md#json-schema) such as the example below:
+Inside the keystore JSON file, you should have an [EIP-2335 conformant keystore file](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2335.md#json-schema).
 
-```
-{
-  "crypto": {
-    "kdf": {
-      "function": "scrypt",
-      "params": {
-        "dklen": 32,
-        "n": 262144,
-        "r": 8,
-        "p": 1,
-        "salt": "30bb9ef21d9f1f946c3c7ab70e27f453180a49d473a2a3e79ca2bc715ac4e898"
-      },
-      "message": ""
-    },
-    "checksum": {
-      "function": "sha256",
-      "params": {},
-      "message": "ba3cf1c8ba5be4f90c36bcf44ee37a779eac8c54b72121e4755b6722e95164a7"
-    },
-    "cipher": {
-      "function": "aes-128-ctr",
-      "params": {
-        "iv": "90f76d9d4d1b089e89802eac2f80b6b7"
-      },
-      "message": "8de2b0f55da54719822db6c083f0436ff94cd638be96c57b91339b438e9355f6"
-    }
-  },
-  "description": "",
-  "pubkey": "b22690ca679edd5fb9c2545f358da1427b8310e8ccf9e7e4f01ddce9b1d711a0362d35225673cce8f33911a22ae1519e",
-  "path": "m/12381/3600/0/0/0",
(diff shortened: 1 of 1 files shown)


docs/usage/validator-management.md
```diff
@@ -36,58 +36,42 @@ Validators are represented by a BLS keypair. Use your generated mnemonic from on
...
-  "pubkey": "b22690ca679edd5fb9c2545f358da1427b8310e8ccf9e7e4f01ddce9b1d711a0362d35225673cce8f33911a22ae1519e",
-  "path": "m/12381/3600/0/0/0",
-  "uuid": "de83e8dc-8f95-4ea0-b9ba-cfa608ff3483",
-  "version": 4
-}
-```
+You will also need the passphrase used the encrypt the keystore. This can be specified interactively, or provided in a plaintext file.
 
-These keystore files can be imported into your Lodestar's keystores folder with the `--importKeystores` command.
+#### Import Keys To Secret Store
```
dapplion: What is the "Secret Store"? Import all it literally does is to copy a json file into another location, not sure if such a name gives the impression something more complex is going on

docs/usage/validator-management.md
```diff
@@ -36,58 +36,42 @@ Validators are represented by a BLS keypair. Use your generated mnemonic from on
...
 ```bash
-./lodestar validator \
-  --importKeystores keystore-m_12381_3600_0_0_0-1654128694.json
+# Interactive passphrase input
+./lodestar validator import --keystore ./validator_keys
+
+# Plaintext passphrase file input
+./lodestar validator import --keystore ./validator_keys --passphraseFile ./password.txt
 ```
+These will be automatically loaded at runtime.
```
dapplion: Instead of runtime, I would say when running the validator command specifically

docs/usage/validator-management.md
```diff
@@ -36,58 +36,42 @@ Validators are represented by a BLS keypair. Use your generated mnemonic from on
...
+
+#### Import Keys at Runtime
 
-Similarly, create a `password.txt` file with the password you set for your keystores and import it with the `--importKeystoresPassword` command.
+To import keys at runtime specify the `--importKeystores` and `--importKeystoresPassword` flags with the `validator` command:
```
dapplion: I'm not sure if this sentence clarifies the previous. Technically when any command is running that's runtime which includes any action. We should find a word to specifically means "running the validator cmd" specifically, instead of any runtime.

wemeetagain:
@dapplion I recommend you take over this PR and merge it

docs/usage/validator-management.md
ansermino: Fair point, I had assumed this was standard language but it's clearly not. Will update with "Lodestar's keystores folder", which is how the existing docs describe it.

docs/usage/validator-management.md
ansermino: We do need to be careful not to confuse people here as we are running the validator command to import these keys, but I think this can be phrased as "running the validator". Will update.

docs/usage/validator-management.md
ansermino: IMO this is a tool that helps configure the actual application, so referring to runtime would imply the validator is running. I don't love the word choice myself, but it was the best I could come up with at the time. 

I will replace this with something more specific :+1:

ansermino:
> @dapplion I recommend you take over this PR and merge it

I have no opposition to this. Generally my turn around is going to be pretty slow. Just want to see this iterated on :)

dapplion:
Assigning @philknows as I realize I'm too bias from developing the internals of the tool, and may produce docs that confuse newcomers.

wemeetagain:
> I have no opposition to this. Generally my turn around is going to be pretty slow. Just want to see this iterated on :)

Yeah I mentioned this bc I know you probably have a lot on your plate. :)

I know we can make any additional edits we want and get it merged without lots of back and forth before it gets stale.

philknows:
Made some edits to clarify the two methods. Already removed the alias flags which we don't want to use anymore in our documentation. Some of these flags may change if we push forward with #4986 also. But will work for now if it's clear.

Screenshot 2023-01-10 at 5 46 48 PM

wemeetagain (approved):

PR #4988, opened 2023-01-09
Discv5 worker

wemeetagain:
WIP, opening for initial review / discussion.

Deployed on `nogroup-hzax41-0` and the `beta` group


packages/beacon-node/src/network/discv5/index.ts
```diff
@@ -0,0 +1,116 @@
...
+    await this.status.workerApi.close();
+    await Thread.terminate((this.status.workerApi as unknown) as Thread);
+  }
+
+  onDiscoveredStr(enrStr: Uint8Array): void {
+    try {
+      this.emit("discovered", ENR.decode(Buffer.from(enrStr)));
+    } catch (e) {
+      // TODO there is a bug in enr encoding(?) that causes many enrs to be incorrectly encoded (and thus incorrectly decoded)
+      // this.logger.error("Unable to decode enr", {enr: Buffer.from(enrStr).toString("hex")}, e as Error);
```
wemeetagain, line 78: There seems to be a bug in ENR encoding?
Uncommenting this results in a flood of error logs.

packages/beacon-node/src/network/metadata.ts
```diff
@@ -35,7 +34,7 @@ export interface IMetadataModules {
  * https://github.com/ethereum/consensus-specs/blob/v1.1.10/specs/phase0/p2p-interface.md#metadata
  */
 export class MetadataController {
-  private enr?: ENR;
+  private setEnrValue?: (key: string, value: Uint8Array) => Promise<void>;
```
wemeetagain, line 37: Someone has to 'own' the ENR.
Current choice is for the discv5 worker to own the ENR and just pass a function here to update it.

The FileENR should also be removed and we just write the ENR to disk on close.

dapplion:
Good starting point! observables seems sufficient to fullfil the need of events here. To safely merge this we need high observability of the metrics of the worker. This blog is great at how to collect basically all data points we need from the worker https://prog.world/monitoring-multithreaded-node-js-applications/

Regarding metrics the approach of just fetching the metrics str is good for now. Also I think we should move to the pattern of the discv5 library creating the metrics names itself, like we do in gossipsub now.

wemeetagain:
Screenshot from 2023-01-09 12-36-33

Added some vm-level metrics and appended to the discv5 dashboard

wemeetagain:
@dapplion @tuyennhv let me know what kind of other metrics you'd like to see

wemeetagain:
> Also I think we should move to the pattern of the discv5 library creating the metrics names itself, like we do in gossipsub now.

https://github.com/ChainSafe/discv5/pull/233 :)

twoeths:
> @dapplion @tuyennhv let me know what kind of other metrics you'd like to see

looks great 👍 , can't think of other metrics to add

packages/beacon-node/src/network/discv5/index.ts
dapplion: Any updates here?

philknows (approved):

PR #4989, opened 2023-01-09
Drop FileENR

dapplion:
**Motivation**

While it's nice to write ENR immediately to disk on set, users should instead use the REST API /eth/v1/node/identity to learn the current ENR.

**Description**

- Drop FileENR, related to https://github.com/ChainSafe/lodestar/pull/4988


wemeetagain:
We should write the ENR on shutdown too, no? only if `--peerIdFile` is set.

dapplion:
> We should write the ENR on shutdown too, no? only if --peerIdFile is set.

peerId is never re-use with current code. Once that is added yes, the ENR should be persisted on shutdown

twoeths (approved):

wemeetagain (approved):

PR #4991, opened 2023-01-09
Fix mimetype for JSON response

### LINKED ISSUE #4990: Wrong header value for JSON content type for API client
**Describe the bug**

The `headers` values for the JSON type data consistent of `<MIME_type>/<MIME_subtype>` but in API client only `<MIME_subtype>` is specified. 

**Expected behavior**

An http endpoint returning the `JSON` data should use proper `Accept:` header. 

**Steps to Reproduce**

1. Send `client.debug.getStateV2("head")` to Lighthouse node 
2. Lighthouse has strict mimetype implementation which will return only if correct mime type. 

- OS: Mac
- Version: `1.3.0`
- Branch: `unstable`

nazarhussain:
**Motivation**

Make sure the API client works with other CL clients. 

**Description**

Fix the mimetype for JSON response. 

Closes #4990

Run all tests.

### DIFF
--- a/packages/api/src/beacon/routes/debug.ts
+++ b/packages/api/src/beacon/routes/debug.ts
@@ -91,7 +91,7 @@ export function getReqSerializers(): ReqSerializers<Api, ReqTypes> {
   const getState: ReqSerializer<Api["getState"], ReqTypes["getState"]> = {
     writeReq: (state_id, format) => ({
       params: {state_id: String(state_id)},
-      headers: {accept: format === "ssz" ? mimeTypeSSZ : ""},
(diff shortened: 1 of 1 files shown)


nazarhussain:
For reference: 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept
https://www.iana.org/assignments/media-types/media-types.xhtml
https://www.iana.org/assignments/media-types/application/json

g11tech (approved):
lgtm

wemeetagain (approved):

PR #4994, opened 2023-01-09
Update API response object

### LINKED ISSUE #4993: API client implementation does not fully compliance with the specs
**Describe the bug**

The API client implementation is not fully compliance with the API specs. There are following observations. 

1. The API specs tend to specify two key information 1) Status Code 2) Response body 
2. The API client implementation only `resolves` with the response body and skip the the other piece of information 
3. For endpoints which are not returning body API client resolves to `undefined`
4. For the endpoints which are returning multiple status codes e.g. `GET /eth/v1/node/health`, `POST /eth/v1/beacon/blocks`  there are no way to identify what response code was returned the API call. 

With these observations, it's not easy to use the API client with any CL node which have fully compliance implementation. 

Lodestar is missing some edge cases on implementation e.g. Lodestar writes the node health code as response body as well as status code. While reading it just reads the body and ignore the status code. 

**Proposals**

There are two approaches we can fix this issue. 

1. Resolve API calls with the either `response body` or `status code`. When response is available we resolve with it otherwise status code. This will introduce minimum change, but could introduce some confusion the API client users. 

2. Create a new response structure given below and always resolves with it. 

```js
{ok: boolean, status: number, [body,result]?: T}
```
With this approach API client user can get both information as response body and status code and use on his convenience. This approach would require a bit extra refactoring compared to option 1, but is more steady and future proof approach. 

We can additionally create an optional config parameter for the API client `resolveErrorResponse: boolean`, if user set we can also resolve error response with `ok: false`. That should be a optional flag, if user want to avoid try/catch and alway check the response in data centric approach.

nazarhussain:
**Motivation**

Make the `/eth/v1/node/health` compliance with the spec.

**Description**

Read/Write the http status code for the node health. 

Closes #4993

Run all tests.


dapplion:
This API is very strange, there must be another way such that routes do not have to expose the status code
```ts
  publishBlock(block: allForks.SignedBeaconBlock): Promise<HttpStatusCode>;
```

packages/api/src/beacon/client/proof.ts
```diff
@@ -1,22 +1,48 @@
...
-import {IHttpClient, getFetchOptsSerializers} from "../../utils/client/index.js";
+import {IHttpClient, getFetchOptsSerializers, ClientOptions, HttpError} from "../../utils/client/index.js";
+import {HttpStatusCode} from "../../utils/client/httpStatusCode.js";
+import {APIClientResponse} from "../../utils/types.js";
 
 /**
  * REST HTTP client for lightclient routes
  */
-export function getClient(_config: IChainForkConfig, httpClient: IHttpClient): Api {
+export function getClient<ErrorAsResponse extends boolean = false>(
```
wemeetagain: Can you document the behavior of ErrorAsResponse somewhere?

Seems that depending on the option passed, you can make errors either throw or return.

Do we need both behaviors? Might be simpler to pick a convention (ie always return) and stick with it.

wemeetagain (commented):
the approach generally looks good to me.
looking for more discussion about the ErrorAsResponse feature

packages/api/src/beacon/client/proof.ts
nazarhussain: @wemeetagain My preference is to use `always return` case. But to avoid backward compatibility, I made it optional for now. If we all agree to one approach we can stick to it.

packages/api/README.md
```diff
@@ -11,8 +11,10 @@ Typescript REST client for the [Ethereum Consensus API spec](https://github.com/
 
 ## Usage
 
+We use more typesafe approach for the API client, where all the errors are returned not thrown. This approach is more easy to document and better to handle all possible error cases.
```
wemeetagain, line 14: Maybe also reference the similarity to the `Response` type.

packages/api/src/beacon/client/debug.ts
```diff
@@ -12,38 +16,55 @@ export function getClient(_config: IChainForkConfig, httpClient: IHttpClient): A
...
+    returnTypes as ReturnTypes<Api>,
+    httpClient
+  );
   // For `getState()` generate request serializer
   const fetchOptsSerializers = getFetchOptsSerializers<Api, ReqTypes>(routesData, reqSerializers);
 
   return {
     ...client,
 
+    // TODO: Debug the type issue
```
wemeetagain, line 31: Can you also think about how we would be able to incorporate this? https://github.com/ethereum/beacon-APIs/issues/250

packages/api/src/utils/client/httpStatusCode.ts
```diff
@@ -0,0 +1,394 @@
+"use strict";
+
+/**
+ * Hypertext Transfer Protocol (HTTP) response status codes.
+ * @see {@link https://www.rfc-editor.org/rfc/rfc7231#section-6}
+ */
+export enum HttpStatusCode {
```
dapplion, line 7: Who is using all these protocols? What's the point of declaring this?

dapplion (commented):
Please note this will introduce a breaking change where we don't want to just have to bump the version of the monorepo to 2.0.0 for this.

wemeetagain:
> Please note this will introduce a breaking change where we don't want to just have to bump the version of the monorepo to 2.0.0 for this.

Do we need to change the monorepo to 2.0 for this?
We've been doing breaking changes (eg: to the types) and not cutting a major release.

We should consider:
- either just not calling this a breaking release
- or doing independent versioning so we can do a breaking release of only this single package
- or just being ok doing a monorepo-wide breaking release.

packages/api/src/beacon/client/debug.ts
nazarhussain: We already have support for SZZ encoding for few endpoints, similarly we can support for others when required. 

https://github.com/ethereum/beacon-APIs/blob/6669d981486b7034fc9911fdf7be157343b6ea5e/apis/debug/state.v2.yaml#L40-L42

nazarhussain:
@dapplion @wemeetagain  I think we consider from a product mindset. This is not a breaking change for the Lodestar users. Where as it concerns for `@lodestar/api` as library, either we should be ok with such changes or have an independant versions for those reusable libraries.

packages/api/src/utils/client/httpStatusCode.ts
nazarhussain: To create types for the API which is completely matching the specification.

dapplion: I mean just declare the codes actually used, not every single code out there

packages/api/src/beacon/client/debug.ts
wemeetagain: This beacon-apis PR is about systematically supporting SSZ for all endpoints, not just supporting one-by-one.

Edit: I'm not suggesting to tackle this in this PR, just more an FYI of possible improvement that will be desired.

wemeetagain (approved):

Issue #4623, opened 2022-10-02
Lodestar does not add UDP to ENR


dapplion:
**Describe the bug**

Jorge from Netheremind got the ENR from Lodestar shortly after start-up and the ENR did not had a UDP port.

**Expected behavior**

Add UDP port to ENR on startup from configured --port flag

twoeths:
refer to https://github.com/ChainSafe/lodestar/pull/4579

the blocker for this issue is https://github.com/ChainSafe/discv5/issues/201

cc @wemeetagain

twoeths:
if we set UDP to ENR, most of our peers are inbound
- this branch
<img width="1304" alt="Screen Shot 2023-01-07 at 15 08 21" src="https://user-images.githubusercontent.com/10568965/211140722-6fa65e48-15ac-4945-8274-a709f7383e30.png">

- unstable
<img width="1269" alt="Screen Shot 2023-01-07 at 15 08 52" src="https://user-images.githubusercontent.com/10568965/211140741-012c5352-fef3-410c-8ed7-f8dac712b2da.png">

we received/sent ways more goodbye requests from/to peers
- this branch
<img width="1264" alt="Screen Shot 2023-01-07 at 15 10 40" src="https://user-images.githubusercontent.com/10568965/211140796-72506636-406e-4052-a466-88f26acf173f.png">
<img width="1270" alt="Screen Shot 2023-01-07 at 15 11 01" src="https://user-images.githubusercontent.com/10568965/211140809-8d394865-6586-46b7-9c03-6821c526e7d6.png">

- unstable
<img width="1273" alt="Screen Shot 2023-01-07 at 15 11 47" src="https://user-images.githubusercontent.com/10568965/211140836-d09f79b1-7948-4413-b0b9-06964aa9d5a6.png">
<img width="1275" alt="Screen Shot 2023-01-07 at 15 13 58" src="https://user-images.githubusercontent.com/10568965/211140925-7ef817ae-1817-4e49-a702-b8a5d1005588.png">

lodestar tends to switch peers a lot with this branch, this is peer connection length
- this branch
<img width="1287" alt="Screen Shot 2023-01-07 at 15 33 32" src="https://user-images.githubusercontent.com/10568965/211141560-a806387d-7bf3-4024-8de8-42c929233b3c.png">

- unstable
[...]

PR #4997, opened 2023-01-10
Tag relevant peer

twoeths:
**Motivation**

- When there are many peers (> 55), libp2p randomly delete peer and it makes our peers unstable. See https://github.com/libp2p/js-libp2p/blob/v0.41.0/src/connection-manager/index.ts#L635
- The reason is we don't tag any peers

**Description**

- Tag relevants peers in use, untag when we disconnect peers
- Keep `maxConnections` option in libp2p's connection manager

found it from #4623

**Test result**
- Deployed to feat1 and "Peer connection length` metric improved
Screen Shot 2023-01-10 at 11 15 40

### DIFF
--- a/packages/beacon-node/src/network/nodejs/bundle.ts
+++ b/packages/beacon-node/src/network/nodejs/bundle.ts
@@ -66,8 +66,11 @@ export async function createNodejsLibp2p(options: ILibp2pOptions): Promise<Libp2
       dialTimeout: 30_000,
 
       autoDial: false,
-      // This is the default value, we reject connections based on tcp() option above so this is not important
-      maxConnections: Infinity,
+      // DOCS: the maximum number of connections libp2p is willing to have before it starts disconnecting.
+      // If ConnectionManager.size > maxConnections calls _maybeDisconnectOne() which will sort peers disconnect
+      // the one with the least `_peerValues`. That's a custom peer generalized score that's not used, so it always
+      // has the same value in current Lodestar usage.
+      maxConnections: options.maxConnections,
       // DOCS: the minimum number of connections below which libp2p not activate preemptive disconnections.
       // If ConnectionManager.size < minConnections, it won't prune peers in _maybeDisconnectOne(). If autoDial is
       // off it doesn't have any effect in behaviour.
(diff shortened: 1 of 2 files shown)


maschad:
I think this duplicates the work of https://github.com/ChainSafe/js-libp2p-gossipsub/issues/380  

the plan [outlined here](https://github.com/libp2p/js-libp2p/issues/369) is to have the option to configure priority peers to prevent the connection manager pruning connections to high-value peers.

in the interim we could just leave [maxConnections](https://github.com/ChainSafe/lodestar/blob/unstable/packages/beacon-node/src/network/nodejs/bundle.ts#L72)  and https://github.com/ChainSafe/lodestar/blob/unstable/packages/beacon-node/src/network/nodejs/util.ts#L73 as the default  (both `Infinity`) to prevent pruning for now (until https://github.com/libp2p/js-libp2p/issues/369 is released)

wemeetagain:
@maschad @tuyennhv

Was thinking about this recently, I think we can do both this PR and https://github.com/ChainSafe/js-libp2p-gossipsub/issues/380

While they both tag peers, I think each has its own merit.

Tagging mesh peers in gossipsub is helpful for any users of gossipsub to avoid disconnecting these important peers.
Likewise, its ok for libp2p applications to tag important peers from their perspective.

If both of these PRs were implemented, if libp2p peer manager were to disconnect a peer, hopefully both tags would work together to give better behavior:
1. first disconnect irrelevant peers
2. next disconnect relevant but non-mesh peers
3. next disconnect relevent mesh peers

maschad:
@wemeetagain good point, I suppose as long as there isn't the prospect of a performance penalty as suggested [here](https://github.com/ChainSafe/js-libp2p-gossipsub/pull/383#issuecomment-1377659379)  then I agree that having both tags would serve as a safeguard against this type of disconnection.

twoeths:
thanks @wemeetagain @maschad for your feeback

@dapplion also suggested to use `maxConnection` option in `@libp2p/tcp` as our `maxPeers` https://github.com/libp2p/js-libp2p-tcp/blob/b93f4c0a921923401264aa560ae259afc9430a93/src/index.ts#L38, so we limit our peers to 55 at the transport layer and `libp2p` never has to prune based on tags, so tagging peers is optional in this case, any thoughts?

we can also do both and add `tcp.maxConnections` feature flag

twoeths:
#5001 was merged which set `maxConnection` to `tcp`, however it only affect inbound connections, still need the tagging solution here @wemeetagain , otherwise we may have too many peers at some points

Screen Shot 2023-01-15 at 05 39 06

wemeetagain (approved):

PR #4999, opened 2023-01-11
Fix peer count by client metric

wemeetagain:
**Motivation**

See https://discord.com/channels/593655374469660673/1049274098490036235/1062627172352790629

**Description**

Fix bug where peer counts were not reset to 0.

### DIFF
--- a/packages/beacon-node/src/network/peers/peerManager.ts
+++ b/packages/beacon-node/src/network/peers/peerManager.ts
@@ -620,6 +620,11 @@ export class PeerManager {
     metrics.peerLongLivedAttnets.reset();
     metrics.peerConnectionLength.reset();
 
+    // reset client counts _for each client_ to 0
+    for (const client of Object.values(ClientKind)) {
+      peersByClient.set(client, 0);
+    }
+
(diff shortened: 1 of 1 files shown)


packages/beacon-node/src/network/peers/peerManager.ts
```diff
@@ -620,6 +620,11 @@ export class PeerManager {
     metrics.peerLongLivedAttnets.reset();
     metrics.peerConnectionLength.reset();
 
+    // reset client counts _for each client_ to 0
+    for (const client of Object.values(ClientKind)) {
+      peersByClient.set(client, 0);
```
twoeths, line 625: we default to 0 at the line 631/636 below already, do you mean `metrics.peersByClient.set({client}, 0);`?

wemeetagain: If all connections for a client gets pruned, no entry for that client is added to the map, so metrics never get set to 0

twoeths: oh I see it now 👍

twoeths (approved):

PR #5003, opened 2023-01-13
Add and propogate blockValue to select full or blinded block

g11tech:
Since capella onwards, blockValue would also be now available from execution generated payloads, its now feasible to choose between engine or execution on max profit naive strategy.

This PR adds and propogates the same.

TODO: 
- [ ] Compare blockValue and choose between execution and blinded block in validator's block.ts

### DIFF
--- a/packages/api/src/beacon/routes/validator.ts
+++ b/packages/api/src/beacon/routes/validator.ts
@@ -16,6 +16,7 @@ import {
   ValidatorIndex,
   RootHex,
   StringType,
+  Wei,
 } from "@lodestar/types";
 import {
   RoutesData,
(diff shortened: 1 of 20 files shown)


dapplion (approved):

philknows:
@g11tech Is this done completely autonomously and by default? It will just compare blinded block bid with what the EL block value is and automatically select the higher one?

g11tech:
> @g11tech Is this done completely autonomously and by default? It will just compare blinded block bid with what the EL block value is and automatically select the higher one?

This one just propogates the value to validator, i have a follow-up PR that provides selection option to either choose builder always or choose basis maximum value 

Will push in couple of hours

g11tech:
@philknows this PR will enable the actual selection:
 - https://github.com/ChainSafe/lodestar/pull/5064

PR #5005, opened 2023-01-13
Use genesis fork version for signing and verifying `BLSToExecutionChange`

g11tech:
Use genesis fork version for signing and verifying withdrawals
ref:
- https://github.com/ethereum/consensus-specs/pull/3206

### DIFF
--- a/packages/beacon-node/test/spec/presets/index.test.ts
+++ b/packages/beacon-node/test/spec/presets/index.test.ts
@@ -65,6 +65,9 @@ specTestIterator(
     ssz_static: {
       type: RunnerType.custom,
       fn: sszStatic([
+        // The LightClient types are disabled momentarily as there is LightClientHeader introduced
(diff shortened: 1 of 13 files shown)


wemeetagain (dismissed):

packages/beacon-node/test/spec/presets/index.test.ts
```diff
@@ -62,7 +62,16 @@ specTestIterator(
     rewards: {type: RunnerType.default, fn: rewards},
     sanity: {type: RunnerType.default, fn: sanity},
     shuffling: {type: RunnerType.default, fn: shuffling},
-    ssz_static: {type: RunnerType.custom, fn: sszStatic()},
+    ssz_static: {
```
g11tech, line 65: these LightClient* types are disabled momentarily as there is LightClientHeader introduced in this spec release and is being addressed in separate PR which will fix and reenable all these types

wemeetagain (approved):

PR #4952, opened 2022-12-28
Use job name to differentiate beacon and validator metrics

nflaig:
**Motivation**

See problem described in #4809 and discussions in #4897.

**Description**

Grafana dashboards now use the `job` default label instead of `scrape_location` custom label to differentiate between beacon and validator metrics.

The prometheus config files are updated to use the expected job name.

Closes #4809

PR #5020, opened 2023-01-18
Fix Memory usage panel in VM + host dashboard

nflaig:
**Motivation**

As introduced in #4952 the `job` default label should be used to differentiate between beacon and validator metrics.

**Description**

Replaces `scrape_location` with `job` label.

### DIFF
--- a/dashboards/lodestar_vm_host.json
+++ b/dashboards/lodestar_vm_host.json
@@ -346,7 +346,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "exemplar": false,
-          "expr": "process_heap_bytes{scrape_location=\"beacon\"}",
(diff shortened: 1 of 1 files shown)


nflaig:
@tuyennhv You added the memory usage panel in this PR #4981. Where are you setting the `scrape_location` label, it does not exist in the default prometheus configuration and other 3rd party deployements are not setting this label either which means the dashboards panel which use `scrape_location` are broken, see issue #4809.

I think we should use the default `job` label to differentiate between beacon node and validator as discussed here #4897.

Would be great to get your opinion on this

wemeetagain (approved):

wemeetagain:
Agree with you @nflaig

PR #5021, opened 2023-01-18
Progress indicator for loading keys

### LINKED ISSUE #4823: validator: indicator for loading keys
Loading more than 100 keys or loading a validator on a low-resource machine lets the user believe the client is "stuck" as it does not indicate any status.

```bash
10% .. Loaded 10/100 keys (0xbb3d65dd002050ac12d963226f41dd88)
20% .. Loaded 20/100 keys (0xb0f4f2afa59948b642fd831d8dae30c6)
30% .. Loaded 30/100 keys (0x6362f93a37edaa6337ca0eb670872bbc)
40% .. Loaded 40/100 keys (0x9dc4325370645bc9ff1440519db86db3)
50% .. Loaded 50/100 keys (0xe257ca039b831fb548cdc37bd6b35439)
```

Getting a drink

nazarhussain:
**Motivation**

Keep user informed in long running processes. 

**Description**

Show progress while loading keys. 

Closes #4823 

Run all tests.

### DIFF
--- a/packages/cli/src/cmds/validator/signers/index.ts
+++ b/packages/cli/src/cmds/validator/signers/index.ts
@@ -3,12 +3,16 @@ import {deriveEth2ValidatorKeys, deriveKeyFromMnemonic} from "@chainsafe/bls-key
 import {interopSecretKey} from "@lodestar/state-transition";
 import {externalSignerGetKeys, Signer, SignerType} from "@lodestar/validator";
 import {toHexString} from "@chainsafe/ssz";
+import {ILogger} from "@lodestar/utils";
 import {defaultNetwork, IGlobalArgs} from "../../../options/index.js";
 import {assertValidPubkeysHex, isValidHttpUrl, parseRange, YargsError} from "../../../util/index.js";
 import {getAccountPaths} from "../paths.js";
 import {IValidatorCliArgs} from "../options.js";
 import {decryptKeystoreDefinitions, PersistedKeysBackend} from "../keymanager/persistedKeys.js";
+import {showProgress} from "../../../util/progress.js";
 import {importKeystoreDefinitionsFromExternalDir, readPassphraseOrPrompt} from "./importExternalKeystores.js";
+
+const KEYSTORE_IMPORT_PROGRESS_MS = 1000;
 
 /**
  * Options processing heriarchy
@@ -35,7 +39,11 @@ import {importKeystoreDefinitionsFromExternalDir, readPassphraseOrPrompt} from "
  * - Fetched directly from remote signer API
  * - Remote signer definition imported from keymanager api
  */
-export async function getSignersFromArgs(args: IValidatorCliArgs & IGlobalArgs, network: string): Promise<Signer[]> {
+export async function getSignersFromArgs(
+  args: IValidatorCliArgs & IGlobalArgs,
+  network: string,
+  {logger, signal}: {logger: Pick<ILogger, "info">; signal: AbortSignal}
+): Promise<Signer[]> {
   // ONLY USE FOR TESTNETS - Derive interop keys
   if (args.interopIndexes) {
     const indexes = parseRange(args.interopIndexes);
@@ -71,7 +79,15 @@ export async function getSignersFromArgs(args: IValidatorCliArgs & IGlobalArgs, 
       password: await readPassphraseOrPrompt(args),
     });
 
-    return await decryptKeystoreDefinitions(keystoreDefinitions, args);
+    const needle = showProgress({
+      total: keystoreDefinitions.length,
+      frequencyMs: KEYSTORE_IMPORT_PROGRESS_MS,
+      signal: signal,
+      progress: ({ratePerSec, percentage, current, total}) => {
+        logger.info(`${percentage.toFixed(0)}% of keystores imported. ${current} of ${total}. ${ratePerSec} keys/s`);
+      },
+    });
+    return await decryptKeystoreDefinitions(keystoreDefinitions, {...args, onDecrypt: needle});
   }
 
   // Remote keys declared manually with --externalSignerPublicKeys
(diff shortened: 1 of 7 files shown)


packages/cli/src/util/progress.ts
```diff
@@ -0,0 +1,42 @@
...
+  signal: AbortSignal;
+  frequencyMs: 1000;
+  progress: (opts: {current: number; total: number; percentage: number; ratePerSec: number}) => void;
+}): NeedleFunc {
+  let current = 0;
+  let last = 0;
+
+  const needle: NeedleFunc = (needle: number) => {
+    // zero is considered first index in the range
+    current = needle + 1;
```
wemeetagain, line 22: can we check against total here and finish immediately?

packages/cli/src/cmds/validator/signers/index.ts
```diff
@@ -3,13 +3,17 @@ import {deriveEth2ValidatorKeys, deriveKeyFromMnemonic} from "@chainsafe/bls-key
...
+import {ILogger} from "@lodestar/utils";
 import {defaultNetwork, IGlobalArgs} from "../../../options/index.js";
 import {assertValidPubkeysHex, isValidHttpUrl, parseRange, YargsError} from "../../../util/index.js";
 import {getAccountPaths} from "../paths.js";
 import {IValidatorCliArgs} from "../options.js";
 import {decryptKeystoreDefinitions, PersistedKeysBackend} from "../keymanager/persistedKeys.js";
+import {showProgress} from "../../../util/progress.js";
 import {importKeystoreDefinitionsFromExternalDir, readPassphraseOrPrompt} from "./importExternalKeystores.js";
 
+const KEYSTORE_IMPORT_PROGRESS_MS = 1000;
```
wemeetagain: I think 1 second may be too fast. Imo 3 or 5 seems better. Thoughts?

packages/cli/src/util/progress.ts
nazarhussain: We have to calculate the progress here as well but yes we can do that and finish immediately.

nazarhussain: @wemeetagain Just thought of it again, In case of finishing it right away we will not be able to calculate the `ratePerSec`, as we don't know in how much ms the needle was moved.

wemeetagain: I don't think we need the `ratePerSec` feature

packages/cli/src/util/progress.ts
```diff
@@ -0,0 +1,42 @@
...
+  };
+
+  const internalId = setInterval(() => {
+    progress({
+      current,
+      total,
+      ratePerSec: ((current - last) / frequencyMs) * 1000,
+      percentage: (current / total) * 100,
```
q9f, line 37: might make sense to have these two fields as a formatted float.

Screenshot at 2023-01-18 19-36-11

dapplion: Expected speed is 1 key every 7 seconds. Display keys per minute

packages/cli/src/cmds/validator/signers/index.ts
nazarhussain: As @dapplion mentioned the usual rate is 1 key / 7 sec, so I changed the progress time to 10s, so it show some useful progress. 

 https://github.com/ChainSafe/lodestar/pull/5021#discussion_r1080712022

wemeetagain (approved):
lgtm

packages/cli/src/cmds/validator/signers/index.ts
```diff
@@ -86,7 +106,20 @@ export async function getSignersFromArgs(args: IValidatorCliArgs & IGlobalArgs,
...
+
+    const needle = showProgress({
+      total: keystoreDefinitions.length,
+      frequencyMs: KEYSTORE_IMPORT_PROGRESS_MS,
+      signal: signal,
+      progress: ({ratePerSec, percentage, current, total}) => {
+        logger.info(
+          `${percentage.toFixed(0)}% of local keystores imported. current=${current} total=${total}${current} rate=${(
+            ratePerSec * 60
+          ).toFixed(2)}keys/m`
```
q9f, line 118: This duplicates the same code from lines 82-93 btw. Might be worth putting in a function.

packages/cli/src/cmds/validator/signers/index.ts
```diff
@@ -71,7 +79,19 @@ export async function getSignersFromArgs(args: IValidatorCliArgs & IGlobalArgs,
...
 
-    return await decryptKeystoreDefinitions(keystoreDefinitions, args);
+    const needle = showProgress({
+      total: keystoreDefinitions.length,
+      frequencyMs: KEYSTORE_IMPORT_PROGRESS_MS,
+      signal: signal,
+      progress: ({ratePerSec, percentage, current, total}) => {
+        logger.info(
+          `${percentage.toFixed(0)}% of keystores imported. current=${current} total=${total}${current} rate=${(
+            ratePerSec * 60
```
q9f, line 89: the rate is always a multiple of 6 due to some rounding earlier on.

packages/cli/src/util/progress.ts
```diff
@@ -0,0 +1,55 @@
...
+  };
+
+  const processProgress = (): void => {
+    const currentTime = Date.now();
+    const processTime = currentTime - lastProcessTime;
+
+    progress({
+      current,
+      total,
+      ratePerSec: processTime === 0 ? 0 : ((current - last) / processTime) * 1000,
```
q9f, line 36: unless we need to calculate the rate per second, we should use rate per minute directly and make sure to use floating point numbers. currently the rate per minute is always a multiple of 6:

- 0.00
- 6.00
- 12.00
- 18.00

PR #5031, opened 2023-01-21
Cache and retransmit bls changes if submitted early

g11tech:
A user (most likely ) is expected to start submitting bls to execution changes earlier than capella hf . This would result in gossip error as the topic is not yet been subscribed by peers.

This PR would cache the bls changes and e re-gossip if there is any gossip publishing error.

### DIFF
--- a/packages/beacon-node/src/api/impl/beacon/pool/index.ts
+++ b/packages/beacon-node/src/api/impl/beacon/pool/index.ts
@@ -104,7 +104,11 @@ export function getBeaconPoolApi({
           try {
             await validateBlsToExecutionChange(chain, blsToExecutionChange);
             chain.opPool.insertBlsToExecutionChange(blsToExecutionChange);
-            await network.gossip.publishBlsToExecutionChange(blsToExecutionChange);
+            try {
+              await network.gossip.publishBlsToExecutionChange(blsToExecutionChange);
+            } catch (e) {
+              await chain.cacheBlsToExecutionChanges(blsToExecutionChange);
+            }
           } catch (e) {
             errors.push(e as Error);
             logger.error(
--- a/packages/beacon-node/src/api/impl/beacon/pool/index.ts
+++ b/packages/beacon-node/src/api/impl/beacon/pool/index.ts
@@ -104,9 +104,9 @@ export function getBeaconPoolApi({
           try {
             await validateBlsToExecutionChange(chain, blsToExecutionChange);
             chain.opPool.insertBlsToExecutionChange(blsToExecutionChange);
-            try {
+            if (chain.clock.currentEpoch >= chain.config.CAPELLA_FORK_EPOCH) {
               await network.gossip.publishBlsToExecutionChange(blsToExecutionChange);
-            } catch (e) {
+            } else {
               await chain.cacheBlsToExecutionChanges(blsToExecutionChange);
             }
           } catch (e) {
--- a/packages/beacon-node/src/api/impl/beacon/pool/index.ts
+++ b/packages/beacon-node/src/api/impl/beacon/pool/index.ts
@@ -112,7 +112,7 @@ export function getBeaconPoolApi({
           } catch (e) {
             errors.push(e as Error);
             logger.error(
-              `Error on submitPoolSyncCommitteeSignatures [${i}]`,
+              `Error on submitPoolBlsToExecutionChange [${i}]`,
               {validatorIndex: blsToExecutionChange.message.validatorIndex},
               e as Error
             );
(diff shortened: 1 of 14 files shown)


packages/beacon-node/src/api/impl/beacon/pool/index.ts
```diff
@@ -104,7 +104,11 @@ export function getBeaconPoolApi({
           try {
             await validateBlsToExecutionChange(chain, blsToExecutionChange);
             chain.opPool.insertBlsToExecutionChange(blsToExecutionChange);
-            await network.gossip.publishBlsToExecutionChange(blsToExecutionChange);
+            try {
+              await network.gossip.publishBlsToExecutionChange(blsToExecutionChange);
+            } catch (e) {
+              await chain.cacheBlsToExecutionChanges(blsToExecutionChange);
```
dapplion, line 115: Swallowing errors generically is very ugly, why not do
```
if (after_capella)
  publish
else
  persist
```

packages/db/src/schema.ts
```diff
@@ -33,6 +33,7 @@ export enum Bucket {
   phase0_proposerSlashing = 14, // ValidatorIndex -> ProposerSlashing
   phase0_attesterSlashing = 15, // Root -> AttesterSlashing
   capella_blsToExecutionChange = 16, // ValidatorIndex -> SignedBLSToExecutionChange
+  capella_blsToExecutionChangeCache = 17, // ValidatorIndex -> SignedBLSToExecutionChange
```
dapplion, line 36: You should be able to recycle the existing in-memory + DB pool. Why is this extra bucket required?

g11tech: umm i guess the current db pool before capella can be used as to recycle, i was going for caching any generic publishing errors but i guess its over kill

packages/beacon-node/src/api/impl/beacon/pool/index.ts
g11tech: i was trying to go for handling any persitent errors, but this looks simpler in terms of UX, so will modify :+1:

packages/db/src/schema.ts
g11tech: the problem could be, once capella is hit, it would become unclear which ones are received from gossip and which ones have been submitted, as there is always a possibility of restarts while the gossip job is not complete, so would like to retain this as a separate repo

g11tech: we can clean it post capella

dapplion: It does not matter where an object is received. You persist to db changes that are still valid to be included in a block

packages/beacon-node/src/api/impl/beacon/pool/index.ts
```diff
@@ -104,7 +104,11 @@ export function getBeaconPoolApi({
           try {
             await validateBlsToExecutionChange(chain, blsToExecutionChange);
             chain.opPool.insertBlsToExecutionChange(blsToExecutionChange);
-            await network.gossip.publishBlsToExecutionChange(blsToExecutionChange);
+            if (chain.clock.currentEpoch >= chain.config.CAPELLA_FORK_EPOCH) {
+              await network.gossip.publishBlsToExecutionChange(blsToExecutionChange);
+            } else {
+              await chain.cacheBlsToExecutionChanges(blsToExecutionChange);
```
dapplion, line 115: Why is it necessary to explicitly call this? `chain.opPool.insertBlsToExecutionChange` should eventually write to DB on shutdown

packages/beacon-node/src/network/network.ts
```diff
@@ -411,6 +412,17 @@ export class Network implements INetwork {
...
+      // If we are subscribed and post capella fork epoch, try gossiping the cached bls changes
+      if (
+        this.isSubscribedToGossipCoreTopics() &&
+        epoch >= this.config.CAPELLA_FORK_EPOCH &&
+        !this.cachedBlsChangesPromise
+      ) {
+        this.cachedBlsChangesPromise = this.gossipCachedBlsChanges().then(() => {
+          this.cachedBlsChangesPromise = null;
+        });
+      }
```
dapplion, line 429: Just move this code above under the `// On fork transition` comment, with some retry in case there are no peers

packages/db/src/schema.ts
dapplion: > the problem could be, once capella is hit, it would become unclear which ones are received from gossip and which ones have been submitted

That's a fair point. However, it's not a consensus issue to re-broadcast other validator messages. After thinking about it I'm not opposed to mark those objects in some way as "submitted pre-capella"

g11tech:
closing in favor of https://github.com/ChainSafe/lodestar/pull/5049

PR #5043, opened 2023-01-23
Optimize UX loading for keystores

### LINKED ISSUE #4179: Optimize UX loading +100 keystores on vc
**Describe the bug**

Next steps:

nazarhussain:
**Motivation**

Improve the UX for the keystores loading. 

**Description**

Add key caching to the keystore loading logic. 

Closes #4179

- Run all unit tests

### DIFF
--- a/packages/cli/src/cmds/validator/keymanager/keystoreCache.ts
+++ b/packages/cli/src/cmds/validator/keymanager/keystoreCache.ts
@@ -1,0 +1,79 @@
+import fs from "node:fs";
+import bls from "@chainsafe/bls";
+import {Keystore} from "@chainsafe/bls-keystore";
+import {SignerLocal, SignerType} from "@lodestar/validator";
+import {fromHex, toHex} from "@lodestar/utils";
+import {PointFormat} from "@chainsafe/bls/types";
+import {lockFilepath, unlockFilepath} from "../../../util/lockfile.js";
+
+export async function loadKeystoreCache(
+  cacheFilepath: string,
+  keystores: Pick<Keystore, "uuid" | "pubkey">[],
+  passwords: string[]
+): Promise<SignerLocal[]> {
+  if (keystores.length !== passwords.length) {
+    throw new Error(
+      `Number of keystores and passwords must be equal. keystores=${keystores.length}, passwords=${passwords.length}`
+    );
+  }
+
+  if (!fs.existsSync(cacheFilepath)) {
+    throw new Error(`Cache file ${cacheFilepath} does not exists.`);
+  }
+
+  lockFilepath(cacheFilepath);
+
+  const password = passwords.join("");
+  // We can't use Keystore.parse as it validates the `encrypted message` to be only 32 bytes.
+  const keystore = new Keystore(JSON.parse(fs.readFileSync(cacheFilepath, "utf8")));
+  const secretKeyConcatenatedBytes = await keystore.decrypt(password);
+
+  const result: SignerLocal[] = [];
+  for (const [index, k] of keystores.entries()) {
+    const secretKeyBytes = Uint8Array.prototype.slice.call(secretKeyConcatenatedBytes, index * 32, (index + 1) * 32);
+    const secretKey = bls.SecretKey.fromBytes(secretKeyBytes);
+    const publicKey = secretKey.toPublicKey().toBytes(PointFormat.compressed);
+
+    if (toHex(publicKey) !== toHex(fromHex(k.pubkey))) {
+      throw new Error(
+        `Keystore ${k.uuid} does not match the expected pubkey. expected=${toHex(fromHex(k.pubkey))}, found=${toHex(
+          publicKey
+        )}`
+      );
+    }
+
+    result.push({
+      type: SignerType.Local,
+      secretKey,
+    });
+  }
+
+  unlockFilepath(cacheFilepath);
+
+  return result;
+}
+
+export async function writeKeystoreCache(
+  cacheFilepath: string,
+  keystores: Pick<Keystore, "uuid" | "pubkey">[],
+  passwords: string[],
+  secretKeys: Uint8Array[]
+): Promise<void> {
+  if (keystores.length !== passwords.length) {
+    throw new Error(
+      `Number of keystores and passwords must be equal. keystores=${keystores.length}, passwords=${passwords.length}`
+    );
+  }
+
+  if (keystores.length !== secretKeys.length) {
+    throw new Error(
+      `Number of keystores and secretkeys must be equal. keystores=${keystores.length}, secretKeys=${secretKeys.length}`
+    );
+  }
+
+  const password = passwords.join("");
+  const secretKeyConcatenatedBytes = Buffer.concat(secretKeys);
+  const publicConcatenatedBytes = Buffer.concat(keystores.map((k) => fromHex(k.pubkey)));
+  const keystore = await Keystore.create(password, secretKeyConcatenatedBytes, publicConcatenatedBytes, cacheFilepath);
+  fs.writeFileSync(cacheFilepath, keystore.stringify());
+}
--- a/packages/cli/src/cmds/validator/keymanager/keystoreCache.ts
+++ b/packages/cli/src/cmds/validator/keymanager/keystoreCache.ts
@@ -83,6 +83,7 @@ export async function writeKeystoreCache(
 
 export async function clearKeystoreCache(cacheFilepath: string): Promise<void> {
   if (fs.existsSync(cacheFilepath)) {
+    unlockFilepath(cacheFilepath);
     fs.unlinkSync(cacheFilepath);
   }
 }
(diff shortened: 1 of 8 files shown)


packages/cli/src/cmds/validator/keymanager/keystoreCache.ts
```diff
@@ -0,0 +1,89 @@
...
+  if (!fs.existsSync(cacheFilepath)) {
+    throw new Error(`Cache file ${cacheFilepath} does not exists.`);
+  }
+
+  lockFilepath(cacheFilepath);
+
+  const password = passwords.join("");
+  // We can't use Keystore.parse as it validates the `encrypted message` to be only 32 bytes.
+  const keystore = new Keystore(JSON.parse(fs.readFileSync(cacheFilepath, "utf8")));
+  const secretKeyConcatenatedBytes = await keystore.decrypt(password);
```
nflaig, line 36: What's the point of encrypting the keystore at all if we store the passwords in the same place? I assume the decrypting still takes a long time if the keystore is big but probably still better then decryption every single key separately as it was done before

It does not seem ideal to me that we store passwords in plain text at all which is currently done by Lodestar as we have a `keystore` and a `secrets` (with passwords) file. Those files are stored on the same device (and even the same folder) so it pointless in terms of security.

In my opinion, we should **never** store the password, only the first time the user imports a encrypted keystore the password should be interactively provided through the cli or keymanager API. After keystores are imported they should be stored unecrypted which makes subsequently loading them much faster and we never store passwords in plain text.

packages/cli/src/cmds/validator/keymanager/persistedKeys.ts
```diff
@@ -246,11 +247,32 @@ export class PersistedKeysBackend implements IPersistedKeysBackend {
...
 ): Promise<Signer[]> {
-  const signers: Signer[] = [];
+  if (opts.cacheFilePath) {
+    try {
+      const signers = await loadKeystoreCache(opts.cacheFilePath, keystoreDefinitions);
+      if (opts?.onDecrypt) {
+        opts?.onDecrypt(signers.length - 1, signers[signers.length - 1]);
+      }
+      return signers;
+    } catch (e) {
```
nflaig: `e` seems to be unused

if we want to keep it for debugging purposes

```suggestion
    } catch (_e) {
```

or could just completely remove it

```suggestion
    } catch {
```

packages/cli/test/unit/util/progress.test.ts
```diff
@@ -53,6 +53,26 @@ describe("progress", () => {
       expect(progress.secondCall.args[0]).to.eql({total, current: total, ratePerSec: 0, percentage: 100});
     });
 
+    it("should call progress with correct values directly reaches to total", () => {
+      const progress = sandbox.spy();
+      const frequencyMs = 50;
+      const total = 8;
+      const needle = showProgress({total, signal: new AbortController().signal, frequencyMs, progress});
```
nflaig, line 60: if we only need to abort signal could consider the following to save some resources

```suggestion
      const needle = showProgress({total, signal: new AbortSignal(), frequencyMs, progress});
```

packages/cli/src/cmds/validator/keymanager/keystoreCache.ts
nazarhussain: 1. Local keystores is a feature, which is consistent across all CL clients. As you mentioned it's not the best way in terms of security. 
2. To overcome security issue, there is a feature of remote key manager, which does not involve storing passing in local file. 

It's upto user to use one feature or other. In regard to decryption, for cache decryption it will take time upto one keystore. So if we have 100 validators keys we can save upto 99% of time with the cache.

nflaig: That's what I figured, looks like all CLs implement it like this. We also quickly discussed this topic in the standup, I will create a separate issue for this to get some ideas how we could improve the current security model but it might just be the case that a remoter signer/key manager is the only solution.

> So if we have 100 validators keys we can save upto 99% of time with the cache.

that's a great performance improvement, I was assuming it will be much faster but having constant time is really a big deal if you have a lot of validators

packages/cli/src/cmds/validator/signers/index.ts
```diff
@@ -91,7 +91,11 @@ export async function getSignersFromArgs(
         );
       },
     });
-    return await decryptKeystoreDefinitions(keystoreDefinitions, {...args, onDecrypt: needle});
+    return await decryptKeystoreDefinitions(keystoreDefinitions, {
```
nflaig: just minor thing I missed earlier, the `await` here should not be needed. Could consider adding the [no-return-await](https://eslint.org/docs/latest/rules/no-return-await) eslint rule to detect this

```suggestion
    return decryptKeystoreDefinitions(keystoreDefinitions, {
```

packages/cli/src/util/progress.ts
```diff
@@ -16,6 +16,7 @@ export function showProgress({
   let current = 0;
   let last = 0;
   let lastProcessTime: number = Date.now();
+  let internalId: NodeJS.Timeout;
```
nflaig: could use a more descriptive name for the interval, I think it will also not be just an ID but the whole `Timeout` object

packages/cli/test/unit/cmds/validator/keymanager/keystoreCache.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+const numberOfSigners = 10;
+
+describe("keystoreCache", () => {
+  let definitions: LocalKeystoreDefinition[];
+  let signers: SignerLocal[];
+  let secretKeys: Uint8Array[];
+  let passwords: string[];
+  let keystoreCacheFile: string;
+  // tmp.setGracefulCleanup();
```
nflaig: commented out code

packages/cli/test/unit/cmds/validator/keymanager/keystoreCache.test.ts
```diff
@@ -0,0 +1,89 @@
...
+      definitions.push({password: secretKey.toHex(), keystorePath});
+      passwords.push(password);
+      secretKeys.push(secretKey.toBytes());
+    }
+  });
+
+  describe("writeKeystoreCache", () => {
+    it("should write a valid keystore cache file", async () => {
+      await expect(writeKeystoreCache(keystoreCacheFile, signers, passwords)).to.fulfilled;
+      expect(fs.existsSync(keystoreCacheFile)).to.be.true;
```
nflaig, line 63: is there an easy way to do some basic sanitiy checks on the written file? I guess the proper check happens in the others test were the file is written and then loaded, so maybe should not bother too much here

nflaig (dismissed):
LGTM! Really good UX and performance improvement!

packages/cli/src/util/progress.ts
nazarhussain: It's common convention among JS community to suffix output of `setTimeout` or `setInterval` as `Ids` because these are just the pointers to v8 internal timers.

nazarhussain: I changed the name to be more descriptive.

packages/cli/test/unit/cmds/validator/keymanager/keystoreCache.test.ts
nazarhussain: The authenticity of files could better be checked in e2e tests.

nflaig (approved):

packages/cli/src/util/progress.ts
nflaig: > It's common convention among JS community to suffix output of setTimeout or setInterval as Ids because these are just the pointers to v8 internal timers

gotcha, thanks for pointing that out, I didn't know about that

wemeetagain (approved):

packages/cli/src/cmds/validator/signers/index.ts
```diff
@@ -91,7 +91,11 @@ export async function getSignersFromArgs(
         );
       },
     });
-    return await decryptKeystoreDefinitions(keystoreDefinitions, {...args, onDecrypt: needle});
+    return decryptKeystoreDefinitions(keystoreDefinitions, {
+      ...args,
+      onDecrypt: needle,
+      cacheFilePath: `${args.importKeystores[0]}.cache`,
```
nflaig, line 97: @nazarhussain I noticed that the `validator_keys.cache` is stored in the same directory as my `validator_keys` folder which I set via `--importKeystores`. This is a not ideal in a containerized environment as the cache file is likely not mounted to the container or persistent in a volume. Maybe it would be better to store the cache file in the data directory (`dataDir`).

In our `docker-compose.validator.yml` file we also only mount the `/keystores` which would not include the cache file.

https://github.com/ChainSafe/lodestar/blob/2042c3b1c1916ceed6bc825057515a4456c33a08/docker-compose.validator.yml#L4-L11

wemeetagain:

PR #5053, opened 2023-01-25
Review eip-4844 sync code

dapplion:
**Motivation**

Spec clarified a couple points that influence our implementation:
- Old blobs must not be considered available if its data availability has not been checked in the past. This means that Lodestar cannot sync from outside the blobs prune window. A better UX strategy should be communicated on boot for users. So far this PR removes `BlockInputType.postEIP4844OldBlobs`. This causes blobs by range calls to a range previous to prune window to have to throw an error, stalling sync effectively. See https://github.com/ethereum/consensus-specs/pull/3141
- To handle the edge case on a by root request after eip-4844 fork and before finalizing it, the server will return a specific error code which we can handle to trigger a retry. See https://github.com/ethereum/consensus-specs/pull/3154

**Description**

- [Compute emptyKzgAggregatedProof once lazily](https://github.com/ChainSafe/lodestar/commit/88b0eee2180e961169ed101c50ff43a00d47c0f9)
- [Fetch blocks and blobs in concurrently](https://github.com/ChainSafe/lodestar/commit/8b749903dcf63330fd6eb3cf4b709dd3f749dd7a)
- [Remove BlockInput case postEIP4844OldBlobs](https://github.com/ChainSafe/lodestar/commit/868df381d6eb21f2513978ceefcd95889ccdbb2b)
- [Review beaconBlocksMaybeBlobsByRoot](https://github.com/ChainSafe/lodestar/commit/7026386f6adea611550bf82b83d6e6a1459dc2b6)
- [Rename beaconBlocksMaybeBlobsByRange](https://github.com/ChainSafe/lodestar/commit/cf5afaaa33f8c02f38f7065565c2e03613b57a5b)
- [Ensure beaconBlocksMaybeBlobsByRange is same epoch](https://github.com/ChainSafe/lodestar/commit/20111de6daaaa4a77521ed05d290300ec9e4d875)

### DIFF
--- a/packages/beacon-node/package.json
+++ b/packages/beacon-node/package.json
@@ -131,6 +131,7 @@
     "@multiformats/multiaddr": "^11.0.0",
     "@types/datastore-level": "^3.0.0",
     "buffer-xor": "^2.0.2",
+    "c-kzg": "^1.0.9",
     "cross-fetch": "^3.1.4",
     "datastore-core": "^8.0.1",
     "datastore-level": "^9.0.1",
--- a/packages/beacon-node/src/chain/blocks/types.ts
+++ b/packages/beacon-node/src/chain/blocks/types.ts
@@ -7,13 +7,11 @@ import {IChainForkConfig} from "@lodestar/config";
 export enum BlockInputType {
   preDeneb = "preDeneb",
   postDeneb = "postDeneb",
-  postDenebOldBlobs = "postDenebOldBlobs",
 }
 
 export type BlockInput =
   | {type: BlockInputType.preDeneb; block: allForks.SignedBeaconBlock}
-  | {type: BlockInputType.postDeneb; block: allForks.SignedBeaconBlock; blobs: deneb.BlobsSidecar}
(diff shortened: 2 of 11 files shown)


dapplion:
@g11tech can you take over this PR and get it merge ready?

g11tech (approved):
lgtm leaving `unmerged` for second pair of eyes

wemeetagain (approved):
Looks good. Appreciate you calling out the changes in the description

packages/beacon-node/package.json
```diff
@@ -131,6 +131,7 @@
     "@multiformats/multiaddr": "^11.0.0",
     "@types/datastore-level": "^3.0.0",
     "buffer-xor": "^2.0.2",
+    "c-kzg": "^1.0.9",
```
dapplion, line 134: @g11tech  Do we want to install c-kzg always now? Current strategy was to keep as optional dependency right?

g11tech: yes, we can install it always :+1:

dapplion: Are you sure? deneb is not necessary now and it can crash Lodestar installations

wemeetagain:

PR #5054, opened 2023-01-25
Fix own enr encoding

wemeetagain:
**Motivation**

We update our enr as part of startup, we need to include the private key when serialize the enr to send to the worker.


g11tech (approved):
lgtm! lodestar is synced on devnet4

wemeetagain:

PR #5064, opened 2023-01-27
Add validatior option to specify builder block selection strategy

g11tech:
PR 
- https://github.com/ChainSafe/lodestar/pull/5003
enabled propagating `blockValue` to validator for builder and engine blocks but the selection logic was unchanged.

This PR takes #5003 to completion by adding a validator option to specify the builder vs engine block strategy with two options:
- either  `--builder.selection builderalways`
- or `--builder.selection maxprofit`
with the self explanatory selection behavior

PS:
Even though builder `blockValue` is always available, but engine's blockValue is/will be only available capella onwards where execution engine provide get_payloadV2 endpoint which returns `blockValue` of the engine block. 
In the case of non-availability, the engine blockValue will be defauled to `0` and hence builder will always be picked so `maxprofit` startegy collapses to `builderalways` in that case


twoeths (approved):

wemeetagain:

PR #5078, opened 2023-01-30
Fix github action warnings

### LINKED ISSUE #4967: Update github workflows to avoid `::set-output`
We need to update our github workflows to not use `::set-output`. We have by May 31 2023 to do so.

See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

twoeths:
**Motivation**
- Fix warnings in https://github.com/ChainSafe/lodestar/actions/runs/4040952032

**Description**

- `set-output`: same fix to #5002 but for `js` file
- `Node.js 12 actions`: migrate to latest version of actions
- `set-state`: I don't see we use it in lodestar, not sure where it comes from

Closes #4967

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -26,8 +26,8 @@ jobs:
 
     steps:
       # <common-build> - Uses YAML anchors in the future
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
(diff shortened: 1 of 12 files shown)


wemeetagain (approved):

wemeetagain:

Issue #5084, opened 2023-01-31
Discv5 Observable performance issue


twoeths:
**Describe the bug**

- Since we introduce Discv5 in worker thread, it takes almost 10% of cpu time on a mainnet node subscribing to all subnets

<img width="1307" alt="Screen Shot 2023-01-31 at 13 12 45" src="https://user-images.githubusercontent.com/10568965/215680881-6d5e3b6e-84aa-4ea3-9689-094c299ae993.png">

[mainnet_node_v1.4.0.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/10543375/mainnet_node_v1.4.0.cpuprofile.zip)

cc @wemeetagain 

**Expected behavior**

- Discv5 should take less cpu to emit a discovered peer to main thread

**Screenshots**

~~This is from a contabo vpss node (v1.3.0 vs v1.4.0)~~ __Update__: ignore the below metric, it's the common situation of all of lodestar nodes on goerli, v1.4.0 has only been deployed since Jan 30

<img width="1265" alt="Screen Shot 2023-01-31 at 13 11 26" src="https://user-images.githubusercontent.com/10568965/215680655-1e8df7c0-ca1a-4eb3-a640-c4713d291ab5.png">

- Version: v1.4.0

wemeetagain:
can you try with https://github.com/ChainSafe/lodestar/pull/5073 applied?

twoeths:
> can you try with #5073 applied?

I will deploy that PR to a node, I will also try using the native thread communication `postMessage` instead of Observable to see if it's better

twoeths:
Further investigation shows that this only happens when we have a lot of discovered ENRs from discv5, somehow this only happens on test mainnet nodes subscribing to all subnets, not on goerli nodes or prod mainnet nodes

we may not need it for v1.4.1

twoeths:
Wasn't able to take same profile, also #5073 should improve it and it was shipped with v1.4.1

PR #5073, opened 2023-01-28
Update to @chainsafe/discv5 v3.0.0

wemeetagain:
**Motivation**

Several things:
- Our ENR api was error-prone, and was refactored in 3.0.0
- We had to re-encode ENRs across the worker boundary (which costs us a signature verification in the main thread)
- We frequently had errors decoding ENRs (that we had successfully decoded and re-encoded within the worker)

**Description**

- Update to @chainsafe/discv5 v3.0.0
- Use new toObject / fromObject to pass ENRs as objects across the thread boundary (no encoding/decoding necessary, no decoding errors possible)

PR #5086, opened 2023-01-31
Use postMessage for Discv5 worker thread

twoeths:
**Motivation**

- Right now we use Observable inside `threads` library to inform a new enr to main thread from discv5 worker thread, the profiler shows that it takes almost 10% on a mainnet node subscribing to all subnets

**Description**

- Use `postMessage` instead
- Cannot see any differences after I run this branch for 3 hours so need to take a profile and a benchmark test to see

**TODO**
- [x] Make sure this works and we can have a lot of discovered peers/ENRs
- [ ] Let this branch run for a while and take another profile to see if `postMessage` is better or not
- [ ] Write a benchmark to see this is better or not

### DIFF
--- a/packages/beacon-node/src/network/discv5/index.ts
+++ b/packages/beacon-node/src/network/discv5/index.ts
@@ -72,8 +72,8 @@ export class Discv5Worker extends (EventEmitter as {new (): StrictEventEmitter<E
 
   onMessage = (msg: Event): void => {
     const messageData = ((msg as unknown) as {data?: unknown})?.data as Discv5EventData | undefined;
-    if (messageData && messageData.type === "discv5-enr") {
-      this.onDiscovered(messageData.data);
+    if (messageData && messageData.type === "discv5-enr-result") {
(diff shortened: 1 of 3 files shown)


twoeths:
`Observable` is only 2.8% slower than `postMessage` (68s vs 70s) if I sent 10M messages from worker thread to main thread, benchmark is in https://gist.github.com/tuyennhv/c8dfcaa8931e19b1ffa6b9558ee28de5 and https://gist.github.com/tuyennhv/d6f441da0aa174899821fc708fd9b8ad

dapplion:
@tuyennhv can you post a screenshot of what specifically is using the 10% of cpu time?

wemeetagain:
this isn't happening anymore after #5073 see description

twoeths:
> @tuyennhv can you post a screenshot of what specifically is using the 10% of cpu time?

@dapplion it was originally from https://github.com/ChainSafe/lodestar/issues/5084 , I don't think it happen anymore after some enhancement/improvement @wemeetagain made

PR #5087, opened 2023-01-31
Add metrics to capture beacon node and validator db size

nflaig:
**Motivation**

At the moment as an operator you need to look on OS level to find out how big the beacon node and validator db is. This is not ideal and there is no metric to see the db growth over time. The db size is also a value we need for the client monitoring implemenation (#5037). Other CLs such as lighthouse and prysm also expose this as a metric.

**Description**

- ~~Adds support for async collect functions to support calling `db.approximateSize`, prom-client itself also supports this, see [CollectFunction](https://github.com/siimon/prom-client/blob/master/index.d.ts#L137) and [get()](https://github.com/siimon/prom-client/blob/master/lib/gauge.js#L107)~~
- Adds `dbSizeTotal` to validator and lodestar metrics
- Add interval to capture metric for db size
- Add DB size panels to VM + host dashboard

image

### DIFF
--- a/packages/db/src/controller/level.ts
+++ b/packages/db/src/controller/level.ts
@@ -33,6 +33,7 @@ export class LevelDbController implements IDatabaseController<Uint8Array, Uint8A
   constructor(opts: ILevelDBOptions, {metrics}: LevelDbControllerModules) {
     this.opts = opts;
     this.metrics = metrics ?? null;
+    if (metrics) this.addDbSizeMetricCollector(metrics);
     this.db = opts.db || new Level(opts.name || "beaconchain", {keyEncoding: "binary", valueEncoding: "binary"});
   }
 
@@ -56,6 +57,7 @@ export class LevelDbController implements IDatabaseController<Uint8Array, Uint8A
       throw Error("metrics can only be set once");
     } else {
       this.metrics = metrics;
+      this.addDbSizeMetricCollector(metrics);
     }
   }
 
@@ -174,6 +176,17 @@ export class LevelDbController implements IDatabaseController<Uint8Array, Uint8A
     }
 
     this.metrics?.dbWriteItems.inc({bucket}, itemsRead);
+  }
+
+  /** Add collect function to capture metric for db size */
+  private addDbSizeMetricCollector(metrics: ILevelDbControllerMetrics): void {
+    const minKey = Buffer.from([0x00]);
+    const maxKey = Buffer.from([0xff]);
+
+    metrics.dbSizeTotal.addCollect(async () => {
+      const dbSize = await this.approximateSize(minKey, maxKey);
+      metrics.dbSizeTotal.set(dbSize);
+    });
   }
 }
 
--- a/packages/db/src/controller/level.ts
+++ b/packages/db/src/controller/level.ts
@@ -20,6 +20,9 @@ export type LevelDbControllerModules = {
 
 const BUCKET_ID_UNKNOWN = "unknown";
 
+/** Time between capturing metric for db size, every few minutes is sufficient */
+const DB_SIZE_METRIC_INTERVAL_MS = 5 * 60 * 1000;
+
 /**
  * The LevelDB implementation of DB
  */
@@ -29,11 +32,11 @@ export class LevelDbController implements IDatabaseController<Uint8Array, Uint8A
 
   private readonly opts: ILevelDBOptions;
   private metrics: ILevelDbControllerMetrics | null;
+  private dbSizeMetricInterval?: NodeJS.Timer;
 
   constructor(opts: ILevelDBOptions, {metrics}: LevelDbControllerModules) {
     this.opts = opts;
     this.metrics = metrics ?? null;
-    if (metrics) this.addDbSizeMetricCollector(metrics);
     this.db = opts.db || new Level(opts.name || "beaconchain", {keyEncoding: "binary", valueEncoding: "binary"});
   }
 
@@ -42,11 +45,19 @@ export class LevelDbController implements IDatabaseController<Uint8Array, Uint8A
     this.status = Status.started;
 
     await this.db.open();
+
+    if (this.metrics) {
+      this.collectDbSizeMetric();
+    }
   }
 
   async stop(): Promise<void> {
     if (this.status === Status.stopped) return;
     this.status = Status.stopped;
+
+    if (this.dbSizeMetricInterval) {
+      clearInterval(this.dbSizeMetricInterval);
+    }
 
     await this.db.close();
   }
@@ -57,7 +68,9 @@ export class LevelDbController implements IDatabaseController<Uint8Array, Uint8A
       throw Error("metrics can only be set once");
     } else {
       this.metrics = metrics;
-      this.addDbSizeMetricCollector(metrics);
+      if (this.status === Status.started) {
+        this.collectDbSizeMetric();
+      }
     }
   }
 
@@ -178,15 +191,24 @@ export class LevelDbController implements IDatabaseController<Uint8Array, Uint8A
     this.metrics?.dbWriteItems.inc({bucket}, itemsRead);
   }
 
-  /** Add collect function to capture metric for db size */
-  private addDbSizeMetricCollector(metrics: ILevelDbControllerMetrics): void {
+  /** Start interval to capture metric for db size */
+  private collectDbSizeMetric(): void {
+    this.dbSizeMetric();
+    this.dbSizeMetricInterval = setInterval(this.dbSizeMetric.bind(this), DB_SIZE_METRIC_INTERVAL_MS);
+  }
+
+  /** Capture metric for db size */
+  private dbSizeMetric(): void {
     const minKey = Buffer.from([0x00]);
     const maxKey = Buffer.from([0xff]);
 
-    metrics.dbSizeTotal.addCollect(async () => {
-      const dbSize = await this.approximateSize(minKey, maxKey);
-      metrics.dbSizeTotal.set(dbSize);
-    });
+    this.approximateSize(minKey, maxKey)
+      .then((dbSize) => {
+        this.metrics?.dbSizeTotal.set(dbSize);
+      })
+      .catch(() => {
+        // ignore errors
+      });
   }
 }
 
--- a/dashboards/lodestar_vm_host.json
+++ b/dashboards/lodestar_vm_host.json
@@ -3517,6 +3517,228 @@
           ],
           "title": "DB - write req / sec",
           "type": "timeseries"
+        },
+        {
+          "fieldConfig": {
+            "defaults": {
+              "color": {
+                "mode": "palette-classic"
+              },
+              "custom": {
+                "axisLabel": "",
+                "axisPlacement": "auto",
+                "barAlignment": 0,
+                "drawStyle": "line",
+                "fillOpacity": 30,
+                "gradientMode": "none",
+                "hideFrom": {
+                  "legend": false,
+                  "tooltip": false,
+                  "viz": false
+                },
+                "lineInterpolation": "linear",
+                "lineWidth": 1,
+                "pointSize": 5,
+                "scaleDistribution": {
+                  "type": "linear"
+                },
+                "showPoints": "auto",
+                "spanNulls": false,
+                "stacking": {
+                  "group": "A",
+                  "mode": "none"
+                },
+                "thresholdsStyle": {
+                  "mode": "off"
+                }
+              },
+              "mappings": [],
+              "thresholds": {
+                "mode": "absolute",
+                "steps": [
+                  {
+                    "color": "green",
+                    "value": null
+                  },
+                  {
+                    "color": "red",
+                    "value": 80
+                  }
+                ]
+              },
+              "unit": "bytes"
+            },
+            "overrides": [
+              {
+                "__systemRef": "hideSeriesFrom",
+                "matcher": {
+                  "id": "byNames",
+                  "options": {
+                    "mode": "exclude",
+                    "names": [
+                      "db size"
+                    ],
+                    "prefix": "All except:",
+                    "readOnly": true
+                  }
+                },
+                "properties": [
+                  {
+                    "id": "custom.hideFrom",
+                    "value": {
+                      "legend": false,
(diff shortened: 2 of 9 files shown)


packages/db/src/controller/level.ts
```diff
@@ -175,6 +177,17 @@ export class LevelDbController implements IDatabaseController<Uint8Array, Uint8A
...
     this.metrics?.dbWriteItems.inc({bucket}, itemsRead);
   }
+
+  /** Add collect function to capture metric for db size */
+  private addDbSizeMetricCollector(metrics: ILevelDbControllerMetrics): void {
+    const minKey = Buffer.from([0x00]);
+    const maxKey = Buffer.from([0xff]);
+
+    metrics.dbSizeTotal.addCollect(async () => {
+      const dbSize = await this.approximateSize(minKey, maxKey);
```
wemeetagain: Should this be this.db.approximateSize ?

packages/db/src/controller/level.ts
```diff
@@ -175,6 +177,17 @@ export class LevelDbController implements IDatabaseController<Uint8Array, Uint8A
 
     this.metrics?.dbWriteItems.inc({bucket}, itemsRead);
   }
+
+  /** Add collect function to capture metric for db size */
+  private addDbSizeMetricCollector(metrics: ILevelDbControllerMetrics): void {
+    const minKey = Buffer.from([0x00]);
+    const maxKey = Buffer.from([0xff]);
```
wemeetagain, line 203: I think the min and max keys are optional? Not setting = whole db?

wemeetagain, line 203: I think the min and max keys are optional? Not setting = whole db?

nflaig: Omitting the start and end key causes an error `TypeError: The arguments 'start' and 'end' are required`, see [approximateSize input validation](https://github.com/Level/classic-level/blob/8a4717b2a2177ad8ad6b35963a61bd7c86bb279f/index.js#L95). The downstream calls also rely on those parameters to be set.

I am not 100% sure if the min and max key I am setting at the moment have the proper values to get the total db size. There is not explicit test in classic-level repository which tests that the full db size is returned by `approximateSize`.

packages/db/src/controller/level.ts
nflaig: I could call `this.db.approximateSize` but there is a wrapper method in the class itself which restrict start and end key to be of type `Uint8Array`. 

https://github.com/ChainSafe/lodestar/blob/cbb761900ab1858b8227e6cec2f5e4cd6abb150d/packages/db/src/controller/level.ts#L141-L143

Shouldn't really matter which method of those two we call, should it be changed to `this.db.approximateSize`?

nflaig: @wemeetagain calling `approximateSize` could throw an error if the database is closed for some reason. In that case the scraping the metrics from `/metrics` whould return a 500 error. Do you think we should just catch any errors here and ignore them?

packages/db/src/controller/level.ts
```diff
@@ -175,6 +177,17 @@ export class LevelDbController implements IDatabaseController<Uint8Array, Uint8A
...
+
+    metrics.dbSizeTotal.addCollect(async () => {
```
nflaig: @wemeetagain I am a bit concerned about the performance impact of this, since the collect function is called every time metrics are fetched. I think for db size we don't need real-time values and could definitely have more delay, we could interally limit this to once every 5-10 minutes maybe but I have not tested this against a bigger database size (20-30GB) so I can't really tell, do we have a quick way to run this branch on a existing system? My testing setup is not ready at the moment

wemeetagain: Yeah, I was hesitant on this PR since it introduced async collection functions when we had none before. (this might have some annoying side effects of making it harder to diagnose event loop lag, since in the past, we've used latency in collecting metrics as a way to measure event loop lag.)

Like you said, you could create a heartbeat function that get gets registered on start, stopped on close, that caches the approx size.

Definitely a way to test this, let me know when you want to test it

packages/db/src/controller/level.ts
wemeetagain: ah, yeah my bad, didn't see this

packages/db/src/controller/level.ts
nflaig: > this might have some annoying side effects of making it harder to diagnose event loop lag, since in the past, we've used latency in collecting metrics as a way to measure event loop lag

thanks for mentioning this, I think we should not introduce async collection then just for this one use case. I think it is better then to use a interval to collect the metrics as you mentioned

> Definitely a way to test this, let me know when you want to test it

Could maybe test it in the current state, we already track the prom scrape duration and could use that as a indicator

nflaig: > that get gets registered on start, stopped on close, that caches the approx size.

@wemeetagain do you know why the `start` and `stop` is not explicitly called in case of the validator LevelDB? We just instantiate it and implicitly open it by calling `db.get`, maybe it is by design that we lazy open the db but I did not find any explanation for this

https://github.com/ChainSafe/lodestar/blob/6793305c50255b0530438d0a2cb1960d96a3be3b/packages/cli/src/cmds/validator/handler.ts#L101-L104

wemeetagain: its not intentional, just an oversight

packages/db/src/controller/level.ts
nflaig: the specified `minKey` and `maxKey` seem to properly capture the whole db size

packages/db/src/controller/level.ts
nflaig: for now, will just catch errors and ignore them. The db package does not utitlize the logger at the moment but I think it should be fine for this use case to just ignore the error as I have not observed a single error as of now. If size a single call to `approximateSize` would fail it is also not an issue as it is called periodically anyways

packages/db/src/controller/level.ts
```diff
@@ -20,6 +20,9 @@ export type LevelDbControllerModules = {
 
 const BUCKET_ID_UNKNOWN = "unknown";
 
+/** Time between capturing metric for db size, every few minutes is sufficient */
+const DB_SIZE_METRIC_INTERVAL_MS = 5 * 60 * 1000;
```
nflaig, line 24: every 5 minutes seems to be sufficient as we don't really need real-time data for this, it is just an approximation after all

packages/db/src/controller/level.ts
nflaig: fixed in https://github.com/ChainSafe/lodestar/pull/5106, we can now rely on start/stop hook to manage heartbeat function to collect db size metric

packages/db/src/controller/level.ts
nflaig: `approximateSize` takes around 0.05 - 0.5ms and it is executed in a different thread, so we could definitely also increase this interval if necessary

packages/db/src/controller/level.ts
```diff
@@ -175,6 +190,26 @@ export class LevelDbController implements IDatabaseController<Uint8Array, Uint8A
...
+    this.dbSizeMetric();
+    this.dbSizeMetricInterval = setInterval(this.dbSizeMetric.bind(this), DB_SIZE_METRIC_INTERVAL_MS);
+  }
+
+  /** Capture metric for db size */
+  private dbSizeMetric(): void {
+    const minKey = Buffer.from([0x00]);
+    const maxKey = Buffer.from([0xff]);
+
+    this.approximateSize(minKey, maxKey)
```
dapplion, line 205: I did some tests back in the day and this was both extremely expensive and inaccurate. Please test in depth before consider merging

dapplion (dismissed):
See above

packages/db/src/controller/level.ts
nflaig: Accuracy looks pretty good

image

image

In terms of performance, I am just testing this on a goerli node at the moment and the DB is not that big. On the 2GB db, `approximateSize` takes around ~0.3ms. This will probably go up a bit if the DB gets bigger but we only execute this once at startup and then every 5 minutes.

I think it makes sense to test this on a mainnet archive node, is there any way we could do this?

dapplion: > I think it makes sense to test this on a mainnet archive node, is there any way we could do this?

Yes sure, provision a node and sync from genesis.

dapplion:
@nflaig from PR description:

> Other CLs such as lighthouse and prysm also expose this as a metric.

Can you check how they collect that metric specifically?

nflaig:
> Can you check how they collect that metric specifically?


### Prysm

Prysm uses the stats function from go os package to [getCurrentDbBytes](https://github.com/prysmaticlabs/prysm/blob/63186c8b0f67ca30641a1d8315882c1e8dfe2115/beacon-chain/node/prometheus.go#L51), this seems to be executed as a collect function, i.e. every time metrics are scraped

```go
func (bc *bcnodeCollector) getCurrentDbBytes() (float64, error) {
	fs, err := os.Stat(bc.dbPath)
	if err != nil {
		return 0, fmt.Errorf("could not collect database file size for prometheus, path=%s, err=%s", bc.dbPath, err)
	}
	return float64(fs.Size()), nil
}
```

### Lighthouse

Lighthouse also uses native fs to get the size [size_of_dir](https://github.com/sigp/lighthouse/blob/38514c07f222ff7783834c48cf5c0a6ee7f346d0/common/directory/src/lib.rs#L79), looks like they also execute this every time metrics are scraped

```rust
/// Get the approximate size of a directory and its contents.
///
/// Will skip unreadable files, and files. Not 100% accurate if files are being created and deleted
/// while this function is running.
pub fn size_of_dir(path: &Path) -> u64 {
    if let Ok(iter) = fs::read_dir(path) {
        iter.filter_map(std::result::Result::ok)
            .map(size_of_dir_entry)
            .sum()
    } else {
        0
    }
}

fn size_of_dir_entry(dir: fs::DirEntry) -> u64 {
    dir.metadata().map(|m| m.len()).unwrap_or(0)
}
```

They both seem to collect on every scrape but as mentioned above I think this is not necessary at all as a data point every 5 minutes is enough. I would rather reduce in the interval further if we want more "real-time" stats about the size.


### Lodestar alternative

Based how prysm and lighthouse measure the size of the db we could follow a similar approach by using `fs` package `stat`

```ts
/** Get approximate size of db folder */
async function getDbFolderSize(dbPath: string): Promise<number> {
  const files = await fs.readdir(dbPath);

  const statsPromises = files.map((file) => {
    const filePath = path.join(dbPath, file);
    return fs.stat(filePath);
  });

  const stats = await Promise.all(statsPromises);

  return stats.reduce((totalSize, stat) => totalSize + stat.size, 0);
}
```

The result of `getDbFolderSize` and `approximateSize` is the same which further indicates that `approximateSize` is accurate. However, `getDbFolderSize` takes way longer to execute ~10ms whereas `approximateSize` takes ~0.3ms. 

### Lodestar alternative 2

Use [get-folder-size](https://www.npmjs.com/package/get-folder-size) package, but this seems to just use `fs` internally and adds some logic to work with nested directories which we do not need. Expect same performance here as with `getDbFolderSize` custom function, or even a bit worse.

PS: `get-folder-size` is even worse, takes ~15ms

packages/db/src/controller/level.ts
nflaig: Did some testing on a mainnet node with 30GB db, `approximateSize` still executes mostly under 1ms

```
mainnet-consensus-1  | approximateSize: 0.182ms
mainnet-consensus-1  | approximateSize: 0.596ms
mainnet-consensus-1  | approximateSize: 0.592ms
mainnet-consensus-1  | approximateSize: 0.211ms
mainnet-consensus-1  | approximateSize: 0.923ms
mainnet-consensus-1  | approximateSize: 1.36ms
mainnet-consensus-1  | approximateSize: 1.185ms
mainnet-consensus-1  | approximateSize: 0.961ms
mainnet-consensus-1  | approximateSize: 0.213ms
```

wemeetagain (approved):
LGTM

Initial reports of performance look good, all metrics are in place to monitor the situation.

packages/db/src/controller/level.ts
```diff
@@ -175,6 +190,26 @@ export class LevelDbController implements IDatabaseController<Uint8Array, Uint8A
...
+      .then((dbSize) => {
+        this.metrics?.dbSizeTotal.set(dbSize);
+      })
+      .catch(() => {
+        // ignore errors
```
dapplion, line 210: @nflaig can you add metrics to track count of errors here + log them to debug? We should never swallow errors. Also, can you track in a histogram the time it takes to run approximateSize?

nflaig: The only reason this would throw an error is if the database is closed which only happens if the process is shutdown. We also clear the interval to call `approximateSize` before that so its really unlikely to happen, also wouldn't care at that point.

> We should never swallow errors

Yes, I agree with this in general but the db package does not use the logger at the moment and I did not want to introduce it just for this case.

> histogram the time it takes to run approximateSize

Was considering that but after looking at the time this takes even for "big" DB sizes I went with "we don't need it"

Let's add a histogram to track the time and also set a `status` label with `success` and `error` value then we see if this ever throws an error during normal operation.

dapplion: Re logger: then you can move the metric caller outside, so that the interval is controlled from somewhere else which has access to the logger. 

Re metrics: would prefer to have the histogram tracking time for all cases, and the error counter as a separate.

nflaig: > Re logger: then you can move the metric caller outside, so that the interval is controlled from somewhere else which has access to the logger.

If we really want a debug error log here which (I still think we don't need it), then I would rather go with injecting the logger to `LevelDbController`, this is the pattern we usually use. In my opinion it is not a good idea to move the logic to track the db size outside of this class/file.

> Re metrics: would prefer to have the histogram tracking time for all cases, and the error counter as a separate.

sure sounds good as well

nflaig: implemented in https://github.com/ChainSafe/lodestar/pull/5190 and to be further discussed there

wemeetagain:

PR #5090, opened 2023-02-01
Fix the args order in evaluating isBetterUpdate for new partial update

g11tech:
As reported by @claravanstaden in discord, the light client updates were not seemingly correct, on debugging figured out that the first assigned update of the period might not be getting replaced by better updates which may arrive later.

On log debugging + code inspection it was discovered that the arg order to isBetterUpdate function was incorrect leading to opposite evalution.

this PR fixes the same


twoeths (approved):

wemeetagain:

Issue #5074, opened 2023-01-30
Low gossipsub score due to p7


twoeths:
**Describe the bug**

This fix https://github.com/ChainSafe/js-libp2p-gossipsub/pull/385 helps minimize broken promises for a while, however since Jan 25 we catch less promises from duplicate and there are more broken promises

This is monitored from a mainnet node subscribing to all subnets. __Update__: this happens on all mainnet nodes

<img width="789" alt="Screen Shot 2023-01-30 at 09 22 54" src="https://user-images.githubusercontent.com/10568965/215374276-d20a1827-63ca-480f-89c6-3e4eddb6ab88.png">

<img width="1296" alt="Screen Shot 2023-01-30 at 09 23 29" src="https://user-images.githubusercontent.com/10568965/215374342-8bd8e248-ab2a-4cff-a136-3b3a3b1484a0.png">

<img width="797" alt="Screen Shot 2023-01-30 at 09 23 53" src="https://user-images.githubusercontent.com/10568965/215374396-2ae045a1-8d9e-4b6d-826e-7cf8954bf5a3.png">

**Expected behavior**

Low broken promises

twoeths:
On mainnet, the percentage of promise broken ratio is quite high since Jan 25

<img width="1297" alt="Screen Shot 2023-01-30 at 10 10 39" src="https://user-images.githubusercontent.com/10568965/215379492-71a3a253-ee60-4e94-9afd-cad6de9dd842.png">

twoeths:
since Jan 25 on mainnet nodes there are more Nimbus peers

<img width="793" alt="Screen Shot 2023-02-06 at 08 42 48" src="https://user-images.githubusercontent.com/10568965/216862869-260c7c4c-cfec-419d-9d19-bee81ddba16d.png">

there are a lot of IWANT promise that's never delivered since then

<img width="1284" alt="Screen Shot 2023-02-06 at 08 43 43" src="https://user-images.githubusercontent.com/10568965/216862971-8f280ce4-0a68-462c-a72b-96a608fba23a.png">

this PR also confirms Nimbus peer score is low https://github.com/ChainSafe/lodestar/pull/5094#issuecomment-1418365878

need to work with Nimbus to improve its gossip peer score

twoeths:
Some issues/PRs in progress in order to debug this issue:
- [x] https://github.com/ChainSafe/lodestar/pull/5175
[...]

PR #5092, opened 2023-02-02
Track gossipsub iwant promises that never delivered percentage

twoeths:
**Motivation**

Iwant promises that's never delivered are not tracked correctly because we only keep `iwantPromiseDeliveryTime` items in gossipsub tracer up to 120s then they are pruned 

**Description**

- Correct the formation: `1 - sum(rate(gossipsub_iwant_promise_delivery_seconds_count[$__rate_interval])) / sum(rate(gossipsub_iwant_promise_sent_total[$__rate_interval]))`
- It shows from Jan 25 there are a lot of Iwant promises are never delivered on mainnet nodes:
Screen Shot 2023-02-02 at 13 51 25

part of #5074

### DIFF
--- a/dashboards/lodestar_debug_gossipsub.json
+++ b/dashboards/lodestar_debug_gossipsub.json
@@ -52,7 +52,10 @@
     "list": [
       {
         "builtIn": 1,
-        "datasource": "-- Grafana --",
+        "datasource": {
+          "type": "datasource",
+          "uid": "grafana"
+        },
         "enable": true,
         "hide": true,
         "iconColor": "rgba(0, 211, 255, 1)",
@@ -70,8 +73,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 5,
-  "iteration": 1649607068536,
+  "id": 1,
   "links": [
     {
       "asDropdown": true,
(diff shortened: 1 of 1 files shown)


dashboards/lodestar_debug_gossipsub.json
```diff
@@ -92,6 +45,10 @@
   "panels": [
     {
       "collapsed": false,
+      "datasource": {
+        "type": "prometheus",
+        "uid": "prometheus_local"
```
nflaig, line 99: need to use `${DS_PROMETHEUS}` here, you can just run `node scripts/lint-grafana-dashboards.mjs ./dashboards` to fix the dashboard

```suggestion
        "uid": "${DS_PROMETHEUS}"
```

dashboards/lodestar_debug_gossipsub.json
```diff
@@ -1,58 +1,12 @@
 {
-  "__inputs": [
```
nflaig: should this `__inputs` header be removed? I think if you do `Export for sharing externally` it is added

twoeths: I'm not sure, not related to this change so I reverted

wemeetagain (approved):

wemeetagain:

PR #5094, opened 2023-02-02
Fix PeerScore.update() function

twoeths:
**Motivation**

1. Some incorrect computation where final score is 0, lodestar score is 0 and gossipScore < 0 as below:

```
{"peer_id":"16Uiu2HAmVcU5cjAGqRwCM1PyniSbhMPEEAgCM5aJ6qZwSFfsDgX5","lodestar_score":0,"gossip_score":-6810.915930589742,"ignore_negative_gossip_score":false,"score":0,"last_update":1675327150937}
```

2. Right now we ignore 5 biggest negative gossip score to allow them to recover but some of them are very small, for example -6000

4. We have no idea how clients perform in our peers

**Description**

1. In the above case the final score should be < 0, should remove the small optimization to check `lodestarScore !== 0` (which was done before we take gossipsub score into account)
2. Only ignore negative gossipsub if it's > -1000
3. Track scores by client

part of #5074

### DIFF
--- a/packages/beacon-node/src/metrics/metrics/lodestar.ts
+++ b/packages/beacon-node/src/metrics/metrics/lodestar.ts
@@ -49,11 +49,12 @@ export function createLodestarMetrics(
       help: "Histogram of current count of long lived attnets of connected peers",
       buckets: [0, 4, 16, 32, 64],
     }),
-    peerScore: register.histogram({
+    peerScoreByClient: register.histogram<"client">({
       name: "lodestar_app_peer_score",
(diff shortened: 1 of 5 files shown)


twoeths:
It shows that Nimbus gossipsub score is low on mainnet

Screen Shot 2023-02-06 at 08 34 21

while other clients perform well, like Lighthouse

Screen Shot 2023-02-06 at 08 34 50

wemeetagain (approved):

wemeetagain:

PR #5097, opened 2023-02-02
fix: update `getConnectionsMap`  and `getConnection` functions 

maschad:
**Motivation**
Following the discussion https://github.com/libp2p/js-libp2p/issues/1567#issuecomment-1414075599 this will reduce some of the complexities of utility functions.

**Description**
This along with a solution for https://github.com/libp2p/js-libp2p/issues/1574 will help us to better inspect, debug and observe the libp2p object overall


wemeetagain (changes requested):
I would prefer to not have additional runtime logic here for the benefit of making the types prettier.

Also we've had trouble in the past with string->peerId conversions causing poor performance:
https://github.com/libp2p/js-libp2p/issues/680#issuecomment-741066845

PR #5027, opened 2023-01-20
Add execution payload header to light client header capella onwards

g11tech:
Add execution payload header to light client header capella onwards

Part of 
 - https://github.com/ChainSafe/lodestar/issues/5102

PR #5101, opened 2023-02-05
Fix and extract payload to header conversion

g11tech:
Earlier blinded header on/post capella was assumed to be different  fromexecution header in terms of having only transactions replaced by transaction root.

However since, blinded header is just same as execution header now  so blinded will already have withdrawalsRoot rather than withdrawals
This is relevant from capella onwards and doesn't affect bellatrix

and the conversion is abstracted and imported to be used in PR:
 - https://github.com/ChainSafe/lodestar/pull/5027


wemeetagain (approved):

wemeetagain:

PR #5104, opened 2023-02-06
Filter enrs in worker

wemeetagain:
**Motivation**

Seen in metrics, most ENRs discovered are not relevant. If irrelevant ENRs can be filtered in a separate thread, than that's fewer messages needed to be sent to the main thread.

Screenshot from 2023-02-05 18-58-52

**Description**

- Check for relevant ENR kvs in worker thread (move logic from `onDisoveredENR`)
- Metric to track ENR relevance of all discovered ENRs
- Refactor `onDiscoveredENR` and `onDiscoveredPeer` into a single function (remove duplicate logic)
- add `discoverKadValues` to emit all enrs in `kadValues` for proper ENR relevance filtering.

### DIFF
--- a/packages/beacon-node/src/network/discv5/index.ts
+++ b/packages/beacon-node/src/network/discv5/index.ts
@@ -11,6 +11,7 @@ import {
   SignableENR,
 } from "@chainsafe/discv5";
 import {spawn, Thread, Worker} from "@chainsafe/threads";
+import {IBeaconConfig} from "@lodestar/config";
 import {ILogger} from "@lodestar/utils";
 import {IMetrics} from "../../metrics/metrics.js";
 import {Discv5WorkerApi, Discv5WorkerData} from "./types.js";
@@ -19,6 +20,7 @@ export type Discv5Opts = {
   peerId: PeerId;
   discv5: Omit<IDiscv5DiscoveryInputOptions, "metrics" | "searchInterval" | "enabled">;
   logger: ILogger;
+  config: IBeaconConfig;
   metrics?: IMetrics;
 };
 
@@ -56,6 +58,8 @@ export class Discv5Worker extends (EventEmitter as {new (): StrictEventEmitter<E
       config: this.opts.discv5,
       bootEnrs: this.opts.discv5.bootEnrs as string[],
       metrics: Boolean(this.opts.metrics),
+      chainConfig: this.opts.config,
+      genesisValidatorsRoot: this.opts.config.genesisValidatorsRoot,
     };
     const worker = new Worker("./worker.js", {workerData} as ConstructorParameters<typeof Worker>[1]);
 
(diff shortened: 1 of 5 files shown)


g11tech (dismissed):
lgtm!

packages/beacon-node/src/network/discv5/index.ts
```diff
@@ -112,6 +116,12 @@ export class Discv5Worker extends (EventEmitter as {new (): StrictEventEmitter<E
     }
   }
 
+  async discoverKadValues(): Promise<void> {
+    if (this.status.status === "started") {
+      await this.status.workerApi.discoverKadValues();
```
nazarhussain, line 121: The `this.status.status` looks really weird. Can we rename it to `this.status` to `this.statusWorker` and `Discv5WorkerStatus` to `Discv5Worker`.

nazarhussain, line 121: The `this.status.status` looks really weird. Can we rename it to `this.status` to `this.statusWorker` and `Discv5WorkerStatus` to `Discv5Worker`.

packages/beacon-node/src/network/discv5/types.ts
```diff
@@ -27,6 +30,8 @@ export type Discv5WorkerApi = {
 
   /** Return the ENRs currently in the kad table */
   kadValues(): Promise<ENRData[]>;
+  /** emit thethe ENRs currently in the kad table */
```
nazarhussain:
```suggestion
  /** emit the ENRs currently in the kad table */
```

packages/beacon-node/src/network/peers/discover.ts
```diff
@@ -260,100 +258,53 @@ export class PeerDiscovery {
...
+    }
+    // async due to some crypto that's no longer necessary
+    const peerId = await enr.peerId();
+    // tcp multiaddr is known to be be present, checked inside the worker
+    const multiaddrTCP = enr.getLocationMultiaddr(ENRKey.tcp);
+    if (!multiaddrTCP) {
+      this.logger.error("Discv5 worker sent enr without tcp multiaddr", {enr: enr.encodeTxt()});
+      this.metrics?.discovery.discoveredStatus.inc({status: DiscoveredPeerStatus.error});
+      return;
+    }
```
nazarhussain, line 282: Can we handle this case inside the worker? so the worker always sent fully verified ENRs.

nazarhussain (changes requested):

nazarhussain:
@wemeetagain The SIM tests failing with `checkpoint-sync` node not able to sync. Can this change be the cause some issue with the syncing?

packages/beacon-node/src/network/discv5/index.ts
wemeetagain: Not really part of the scope of this PR. But happy to resolve separately.

packages/beacon-node/src/network/peers/discover.ts
wemeetagain: This case is handled inside the worker. This condition is included here just for type safety

wemeetagain:
>  Can this change be the cause some issue with the syncing?
Looks like it was just a flaky sim test

nazarhussain (approved):

wemeetagain:

PR #5105, opened 2023-02-06
Release v1.4.2

wemeetagain:
Hotfix to include
- #5090
- #5094


twoeths:
- Lodestar sent more goodbye requests on beta/mainnet nodes, on a 7-day timeframe

Screen Shot 2023-02-07 at 13 35 09

Screen Shot 2023-02-07 at 13 35 19

- Mesh peers are a bit low but I assume it'd take time to build up as we don't change gossipsub part (this is still the same mesh peers I see in other prod nodes)

Screen Shot 2023-02-07 at 13 37 10

Except for that I don't see other differences, this is not a big change anyway

twoeths (approved):

PR #5109, opened 2023-02-06
Hex encode graffiti in transit when calling produce block APIs

nflaig:
**Motivation**

According to beacon node API specification the graffiti should be a hex encoded string, see [produceBlockV2](https://ethereum.github.io/beacon-APIs/#/ValidatorRequiredApi/produceBlockV2) and have a [length of 66 characters](https://github.com/ethereum/beacon-APIs/blob/181575da112fe6556e48a5b7e0ad4a2154e95da0/types/primitive.yaml#L74).

Issue highlighted by @dB2510 in [discord](https://discord.com/channels/849256203614945310/936375861131755540/1072039316752171090).

**Description**

Hex encodes graffiti when writing `produceBlock` request and decodes it when parsing it on the beacon node. The functions for converting a uft8 string to hex and vice-versa are browser compatible.

If the graffiti is too long, it removes characters from the end and if it is too short, the hex value will be right-padded with zeros.

Could consider to explicitly fail validator on start if configured graffiti is too long instead of implicitly trimming it but should be discussed first if we want this behavior.

### DIFF
--- a/packages/api/src/utils/serdes.ts
+++ b/packages/api/src/utils/serdes.ts
@@ -89,7 +89,7 @@ export function fromGraffitiHex(hex: string): string {
   try {
     return new TextDecoder("utf8").decode(fromHexString(hex));
   } catch {
-    // allow malformed graffiti
-    return "";
+    // allow malformed graffiti hex string
+    return hex;
   }
 }
--- a/packages/api/src/utils/serdes.ts
+++ b/packages/api/src/utils/serdes.ts
@@ -78,8 +78,8 @@ export function toGraffitiHex(utf8: string): string {
   }
 
   if (hex.length < GRAFFITI_HEX_LENGTH) {
-    // left-pad with zeros if hex string is too short
-    return "0x" + hex.slice(2).padStart(GRAFFITI_HEX_LENGTH - 2, "0");
+    // right-pad with zeros if hex string is too short
+    return hex.padEnd(GRAFFITI_HEX_LENGTH, "0");
   }
 
   return hex;
--- a/packages/api/src/utils/serdes.ts
+++ b/packages/api/src/utils/serdes.ts
@@ -86,5 +86,10 @@ export function toGraffitiHex(utf8: string): string {
 }
 
 export function fromGraffitiHex(hex: string): string {
-  return new TextDecoder("utf8").decode(fromHexString(hex));
+  try {
+    return new TextDecoder("utf8").decode(fromHexString(hex));
+  } catch {
+    // allow malformed graffiti
+    return "";
+  }
 }
--- a/packages/api/src/utils/serdes.ts
+++ b/packages/api/src/utils/serdes.ts
@@ -1,4 +1,4 @@
-import {JsonPath} from "@chainsafe/ssz";
+import {fromHexString, JsonPath, toHexString} from "@chainsafe/ssz";
 
 /**
  * Serialize proof path to JSON.
@@ -65,4 +65,26 @@ export function fromU64StrOpt(u64Str: U64Str | undefined): U64 | undefined {
 
 export function toU64StrOpt(u64: U64 | undefined): U64Str | undefined {
   return u64 !== undefined ? toU64Str(u64) : undefined;
+}
+
+const GRAFFITI_HEX_LENGTH = 66;
+
+export function toGraffitiHex(utf8: string): string {
+  const hex = toHexString(new TextEncoder().encode(utf8));
+
+  if (hex.length > GRAFFITI_HEX_LENGTH) {
+    // remove characters from the end if hex string is too long
+    return hex.slice(0, GRAFFITI_HEX_LENGTH);
+  }
+
+  if (hex.length < GRAFFITI_HEX_LENGTH) {
+    // left-pad with zeros if hex string is too short
+    return "0x" + hex.slice(2).padStart(GRAFFITI_HEX_LENGTH - 2, "0");
+  }
+
+  return hex;
+}
+
+export function fromGraffitiHex(hex: string): string {
+  return new TextDecoder("utf8").decode(fromHexString(hex));
 }
--- a/packages/api/test/unit/beacon/testData/validator.ts
+++ b/packages/api/test/unit/beacon/testData/validator.ts
@@ -6,7 +6,7 @@ import {GenericServerTestCases} from "../../../utils/genericServerTest.js";
 const ZERO_HASH = Buffer.alloc(32, 0);
 const ZERO_HASH_HEX = "0x" + ZERO_HASH.toString("hex");
 const randaoReveal = Buffer.alloc(96, 1);
-const graffiti = "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2";
+const graffiti = "a".repeat(32);
 
 export const testData: GenericServerTestCases<Api> = {
   getAttesterDuties: {
--- a/packages/api/src/beacon/routes/validator.ts
+++ b/packages/api/src/beacon/routes/validator.ts
@@ -33,7 +33,7 @@ import {
   ContainerDataExecutionOptimistic,
   ContainerData,
 } from "../../utils/index.js";
-import {fromU64Str, toU64Str, U64Str} from "../../utils/serdes.js";
+import {fromU64Str, fromGraffitiHex, toU64Str, U64Str, toGraffitiHex} from "../../utils/serdes.js";
 import {ExecutionOptimistic} from "./beacon/block.js";
 
 // See /packages/api/src/routes/index.ts for reasoning and instructions to add new routes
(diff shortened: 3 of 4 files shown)


packages/api/test/unit/beacon/testData/validator.ts
```diff
@@ -6,7 +6,7 @@ import {GenericServerTestCases} from "../../../utils/genericServerTest.js";
 const ZERO_HASH = Buffer.alloc(32, 0);
 const ZERO_HASH_HEX = "0x" + ZERO_HASH.toString("hex");
 const randaoReveal = Buffer.alloc(96, 1);
-const graffiti = "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2";
+const graffiti = "a".repeat(32);
```
wemeetagain, line 9: why is this changed?

packages/api/src/utils/serdes.ts
```diff
@@ -66,3 +66,25 @@ export function fromU64StrOpt(u64Str: U64Str | undefined): U64 | undefined {
...
+  const hex = toHexString(new TextEncoder().encode(utf8));
+
+  if (hex.length > GRAFFITI_HEX_LENGTH) {
+    // remove characters from the end if hex string is too long
+    return hex.slice(0, GRAFFITI_HEX_LENGTH);
+  }
+
+  if (hex.length < GRAFFITI_HEX_LENGTH) {
+    // left-pad with zeros if hex string is too short
+    return "0x" + hex.slice(2).padStart(GRAFFITI_HEX_LENGTH - 2, "0");
```
wemeetagain: I think this should be right-padded or this will change how our graffiti is written. Currently smaller-than-32 byte input is implicitly right-padded.

wemeetagain (changes requested):

packages/api/test/unit/beacon/testData/validator.ts
nflaig: `0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2` this value did not make too much sense in the first place, it is from the [produceBlockV2](https://ethereum.github.io/beacon-APIs/#/ValidatorRequiredApi/produceBlockV2) which makes sense there becasue it is is the value which should be passed to the beacon API but in our case it should be the actual validator graffiti value which is a UFT-8 string and not hex encoded.

The [`runGenericServerTest` ](https://github.com/ChainSafe/lodestar/blob/c152fe7ec0064c567c70dce74e5a1b6934873ee2/packages/api/test/utils/genericServerTest.ts#L21) function currently uses the `args` as input to the [api client](https://github.com/ChainSafe/lodestar/blob/c152fe7ec0064c567c70dce74e5a1b6934873ee2/packages/api/test/utils/genericServerTest.ts#L52) and it uses that same value to check if it was received by the [mock api](https://github.com/ChainSafe/lodestar/blob/c152fe7ec0064c567c70dce74e5a1b6934873ee2/packages/api/test/utils/genericServerTest.ts#L61). This means at the moment we generally assume we sent some value do some kind of encoding in `writeReq` and then decode in `parseReq` without modyfing the value.

The problem is that `toGraffitiHex` will trim or pad the value depending on its length, meaning the graffiti needs to be exactly 32 bytes for the test to pass.

packages/api/src/utils/serdes.ts
nflaig: sure let's right-pad it instead, it is also simpler in terms of the implementations since we only need to `padEnd` until `GRAFFITI_HEX_LENGTH`

packages/api/src/utils/serdes.ts
```diff
@@ -66,3 +66,25 @@ export function fromU64StrOpt(u64Str: U64Str | undefined): U64 | undefined {
...
+  const hex = toHexString(new TextEncoder().encode(utf8));
+
+  if (hex.length > GRAFFITI_HEX_LENGTH) {
+    // remove characters from the end if hex string is too long
+    return hex.slice(0, GRAFFITI_HEX_LENGTH);
+  }
+
+  if (hex.length < GRAFFITI_HEX_LENGTH) {
+    // right-pad with zeros if hex string is too short
+    return hex.padEnd(GRAFFITI_HEX_LENGTH, "0");
```
nflaig, line 82: @wemeetagain after decoding in `fromGraffitiHex` the string will have null bytes `\u0000` at the end, do you think this could cause any issues?

packages/api/src/utils/serdes.ts
```diff
@@ -66,3 +66,25 @@ export function fromU64StrOpt(u64Str: U64Str | undefined): U64 | undefined {
 export function toU64StrOpt(u64: U64 | undefined): U64Str | undefined {
   return u64 !== undefined ? toU64Str(u64) : undefined;
 }
+
+const GRAFFITI_HEX_LENGTH = 66;
```
nflaig, line 70: @wemeetagain I don't really like that the constant is defined here, do you have a better idea where we could place it?

packages/api/src/utils/serdes.ts
wemeetagain: Thats fine, it seems most graffitis on-chain are that way. Its just that block explorers strip off trailing 0s and convert to utf8 for display

packages/api/src/utils/serdes.ts
wemeetagain: imo it seems fine here, next to these graffiti hex utility functions. I don't think there's anything really reusable or more broad about the graffiti hex length right now.

wemeetagain (dismissed):

packages/api/src/utils/serdes.ts
```diff
@@ -66,3 +66,25 @@ export function fromU64StrOpt(u64Str: U64Str | undefined): U64 | undefined {
...
+  if (hex.length < GRAFFITI_HEX_LENGTH) {
+    // right-pad with zeros if hex string is too short
+    return hex.padEnd(GRAFFITI_HEX_LENGTH, "0");
+  }
+
+  return hex;
+}
+
+export function fromGraffitiHex(hex: string): string {
+  return new TextDecoder("utf8").decode(fromHexString(hex));
```
wemeetagain: is there an edgecase where a graffiti is passed in that throws here?

The case I'm thinking of is a graffiti with 31 alphanumeric characters, then an emoji. toGraffitiHex then chops off part of the emoji resulting in invalid utf8?

Might be safer to wrap this in a try/catch.

g11tech (dismissed):

packages/api/src/utils/serdes.ts
nflaig: > The case I'm thinking of is a graffiti with 31 alphanumeric characters, then an emoji. toGraffitiHex then chops off part of the emoji resulting in invalid utf8?

I have tested that scenario with `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa🐼` which results in hex `0x61616161616161616161616161616161616161616161616161616161616161f0` which is decoded to `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa�`, so looks like it can't display the last character of the UFT-8 string but it does not throw an error. 

`fromGraffitiHex` will mostly throw an error if [`fromHexString`](https://github.com/ChainSafe/ssz/blob/b0e99993b86c669b0e9a95882c1e7c85fdd97075/packages/ssz/src/util/byteArray.ts#L17) fails


> Might be safer to wrap this in a try/catch.

Before the API input validation was basically non existent, you could provide any kind of graffiti and the beacon node would somehow make it work. Now we are definitely stricter, an invalid hex value for example would throw an error. I guess what we do not want is that proposing a block fails due to an invalid graffiti.

Should we add a try-catch which would return an empty string as fallback value for the graffiti? And maybe add a warning log message?

wemeetagain: Yeah, I think its preferable to allow a malformed graffiti for now. try/catch + warning sounds sufficient.

Seems that fromHexString is the only throwable thing. TextDecoder seems to default to a "non-fatal" mode: https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/TextDecoder 
> the decoder will substitute malformed data with a replacement character.

nflaig: Wrapped decoding into try-catch, if an invalid hex string is provided ~~we will now just use an empty string (`""`) as graffiti.~~

Regarding the warning, it is not possible to access the logger here and I do not want to add a `console.log`. ~~Maybe it might anyway better if we do some basic validation of the graffiti on startup of the VC and show the warning there if the graffiti is malformed. If another client is interacting with the BN API we can not do much, it is up to that client to warn the user about the malformed graffiti, I think it should anyway not be the concern of the BN.~~

PS: not sure if we need to validate the graffiti on startup, maybe some basic checks regarding the length could make sense

nflaig: Regarding the default return value, changed it now to just return the provided value if decoding fails. This should not cause any issues downstream as it is how the BN currently handles the graffiti (without any modification).

wemeetagain (dismissed):
LGTM, will wait for another ack for this one before merging

g11tech (dismissed):
can we add unit tests for to/from graffitiHex for normal and special character strings?

g11tech:
also check if we require similar change in keymanager api / external signer api

nflaig:
> can we add unit tests for to/from graffitiHex for normal and special character strings?

added some unit test, those should cover all edge cases but let me know if I missed something

nflaig:
> also check if we require similar change in keymanager api / external signer api

did not find any references to the graffiti in the [keymanager api spec](https://ethereum.github.io/keymanager-APIs/) or any other place in the code where we should apply graffiti encoding/decoding

wemeetagain (approved):

wemeetagain:

PR #5116, opened 2023-02-08
Allow to set empty graffiti

### LINKED ISSUE #5110: Allow to set empty graffiti
At the moment, it is not possible to set an empty graffiti value by providing `--graffiti=""` as omitting the graffiti or setting it to an empty string both results in the default graffiti being used which is `Lodestar-<version>/<short-commit-sha>`. 

I think we should allow to explicitly set an empty graffiti as an operator might not want to use the default graffiti but also not a custom graffiti.

This is quite easy to change in terms of the implementation, it is just a matter of if we want this behavior or not.

Current code which disallows empty graffiti

https://github.com/ChainSafe/lodestar/blob/bc7759674f5c54013c238cb369a9d60c0635b5f9/packages/cli/src/cmds/validator/handler.ts#L184-L185

Fix to allow empty graffiti

```ts
 const defaultConfig = { 
   graffiti: args.graffiti ?? getDefaultGraffiti(),
```

nflaig:
`--graffiti=""` should result in an empty graffiti being used instead of the default graffiti which is `Lodestar-<version>/<short-commit-sha>`. If graffiti CLI option is omitted it will still use the default one.

Closes #5110

### DIFF
--- a/packages/cli/src/cmds/validator/handler.ts
+++ b/packages/cli/src/cmds/validator/handler.ts
@@ -185,7 +185,7 @@ function getProposerConfigFromArgs(
   }: {persistedKeysBackend: IPersistedKeysBackend; accountPaths: {proposerDir: string}}
 ): ValidatorProposerConfig {
   const defaultConfig = {
-    graffiti: args.graffiti || getDefaultGraffiti(),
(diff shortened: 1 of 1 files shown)


wemeetagain (approved):

wemeetagain:

PR #5117, opened 2023-02-08
Remove unnecessary return await

nflaig:
**Motivation**

from eslint docs [no-return-await](https://eslint.org/docs/latest/rules/no-return-await)

> This rule aims to prevent a likely common performance hazard due to a lack of understanding of the semantics of async function.

> Using return await inside an async function keeps the current function in the call stack until the Promise that is being awaited has resolved, at the cost of an extra microtask before resolving the outer Promise

**Benefit**

Improve performance by avoiding the extra microtask

**Trade off**

The function no longer being a part of the stack trace if an error is thrown asynchronously from the Promise being returned. This can make debugging more difficult.

I have personally not experienced any debugging issues due to missing `await` in a return statement, if a developers wants a piece of code to be more debuggable should be explicit about it by assigning the return value to a variable before returning.

```ts
async function foo() {
    const x = await bar();
    return x;
}
```

**Description**

- Add eslint rule to disallow unnecessary return await
- Fix unnecessary return await statements

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -134,6 +134,7 @@ module.exports = {
     "no-consecutive-blank-lines": 0,
     "no-console": "error",
     "no-var": "error",
+    "no-return-await": "error",
     "object-curly-spacing": ["error", "never"],
     "object-literal-sort-keys": 0,
     "no-prototype-builtins": 0,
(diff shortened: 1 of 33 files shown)


wemeetagain:
@dapplion should review this since he had some opinions on this in the past

dapplion:
As long as you don't do
```ts
try {
  await foo() 
} catch () {}
```
all those return await are useless

dapplion (approved):

wemeetagain (approved):

wemeetagain:
reopened to get CI to restart

wemeetagain:

PR #5123, opened 2023-02-09
Fix prometheus in local metrics docker setup

nflaig:
**Motivation**

Dockerized setup to collect metrics from local beacon node and validator client is broken

**Description**

need to set in `environment` now instead of `args` else default values set in Dockerfile will be used

https://github.com/ChainSafe/lodestar/blob/d5e05d3d850e5fe6374b3c832250036f0da4c9f5/docker/prometheus/Dockerfile#L11-L13

also fixed default value for `VC_URL` to use default metric port

One other thing, the local grafana docker build is also broken but not sure how to go about that and I think the issue is already known

this is not docker build compliant

https://github.com/ChainSafe/lodestar/blob/d5e05d3d850e5fe6374b3c832250036f0da4c9f5/docker/grafana/Dockerfile#L10-L11

should not rely on `buildx` features here if possible

### DIFF
--- a/docker/docker-compose.local.yml
+++ b/docker/docker-compose.local.yml
@@ -11,11 +11,11 @@ services:
   prometheus:
     build:
       context: prometheus
-      args:
-        # Linux:  http://localhost:8008
-        # MacOSX: http://host.docker.internal:8008
-        BEACON_URL: localhost:8008
-        VC_URL: localhost:5064
+    environment:
+      # Linux:  http://localhost:8008
+      # MacOSX: http://host.docker.internal:8008
+      BEACON_URL: localhost:8008
+      VC_URL: localhost:5064
     restart: always
     network_mode: host
     volumes:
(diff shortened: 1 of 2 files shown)


wemeetagain:
Agree the reliance on buildx is not good. Open to suggestions to make life easier

wemeetagain (approved):

nflaig:
@wemeetagain I am not sure if there is a approach without some trade-offs, at the moment running `docker-compose -f docker/docker-compose.local.yml up -d` will fail due to as it tries to load dashboards from `docker.io/library/dashboards:latest`. 
To make the build work have to comment out this line `COPY --from=dashboards . /dashboards/` which is not ideal.

I see two possible solutions for this

1. move the build context to root folder which would allow to copy dashboards without relying on `buildx`
2. publish a separate image `chainsafe/lodestar-dashboards` which just bundles the dashboards then load the dashboards from that image. This image could potentially also be useful for 3rd party services which want to make use of lodestar dashboards in their custom builds.

wemeetagain:

PR #5124, opened 2023-02-10
Track lodestar published blocks after 1/3 of slot

twoeths:
**Motivation**

If lodestar publishes a block after 4s, it may be reorged by other nodes so we have to track it (although it maybe rare on mainnet). But right now the bucket value is 5 instead of 4

**Description**

- Change the bucket value of `validator_monitor_beacon_block_delay_seconds` metric

### DIFF
--- a/packages/beacon-node/src/metrics/metrics/lodestar.ts
+++ b/packages/beacon-node/src/metrics/metrics/lodestar.ts
@@ -912,7 +912,7 @@ export function createLodestarMetrics(
         name: "validator_monitor_beacon_block_delay_seconds",
         help: "The delay between when the validator should send the block and when it was received",
         labelNames: ["src"],
-        buckets: [0.1, 0.25, 0.5, 1, 2, 5, 10],
+        buckets: [0.1, 0.25, 0.5, 1, 2, 4, 6, 10],
       }),
 
       // Only for known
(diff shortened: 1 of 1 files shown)


wemeetagain (dismissed):

wemeetagain:
We may want even more granular timing?
It's not just that we must have published a block by 4 seconds, the next proposer must have actually received our block by 4 seconds. I'm not sure how long it takes for blocks to propagate through the network, but perhaps we should adjust the buckets to account for that.

twoeths:
> We may want even more granular timing? It's not just that we must have published a block by 4 seconds, the next proposer must have actually received our block by 4 seconds. I'm not sure how long it takes for blocks to propagate through the network, but perhaps we should adjust the buckets to account for that.

@wemeetagain agree, I'm not sure how much time it needs to reach other nodes (and process it) so added bucket 3 and 3.5, maybe a bit redundant but I think it does not have any performance impact

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -912,7 +912,8 @@ export function createLodestarMetrics(
         name: "validator_monitor_beacon_block_delay_seconds",
         help: "The delay between when the validator should send the block and when it was received",
         labelNames: ["src"],
-        buckets: [0.1, 0.25, 0.5, 1, 2, 5, 10],
+        // we also want other nodes to received our published before 4s so add bucket 3 and 3.5
+        buckets: [0.1, 0.25, 0.5, 1, 2, 3, 3.5, 4, 6, 10],
```
wemeetagain:
```suggestion
        buckets: [0.1, 0.25, 0.5, 1, 2, 3, 4, 6, 10],
```

Removing 3.5 as I don't think it gives us any benefit. From our metrics, we consistently get blocks with 1s delay. Any block sent after 3s will likely not reach enough of the network, the half second doesn't help us.

wemeetagain (approved):

wemeetagain:

Issue #4856, opened 2022-12-05
Move libp2p,discv5,backfill to workers


wemeetagain:
During team retreat we discussed the idea of parallelizing work across multiple threads as makes sense.

If anyone has more notes from that session add here please.

Some open questions:
- how should metrics be handled across workers?

dapplion:
> how should metrics be handled across workers?

I've been playing around with discv5 on a thread on this is the biggest hurdle I found so far. 

Requirements:
- Functions can't be called directly, everything must be a message / event
- Metrics server will remain in the main thread, which can have metrics data already available or collect it from somewhere

Solutions:
- On every metrics call (`metric.inc()`) fire an event to upstream that data to main thread
- Collect metrics on worker, and then on scrape collect all of them with a single event

dapplion:
Some notes

![IMG_20221201_201337](https://user-images.githubusercontent.com/35266934/205808837-af728bdd-ac1f-46ac-b70c-6e21cd549fe0.png)
![IMG_20221201_201319](https://user-images.githubusercontent.com/35266934/205808847-7f9cacd3-34cc-4f41-9804-057615195937.png)

wemeetagain:
> Collect metrics on worker, and then on scrape collect all of them with a single event

Yeah thats what I was leaning towards. The worker maintains its own Registry. And the worker exposes a `async metrics(): Promise<string>` that basically looks like `return await registry.metrics()`

wemeetagain:
> I've been playing around with discv5 on a thread on this is the biggest hurdle I found so far.

Same :joy: 

The other weird thing I saw is that the `threads` library doesn't support the event listener pattern, only "observables". You can create an observable per event to listen on, its just a little weird.

dapplion:
> The other weird thing I saw is that the threads library doesn't support the event listener pattern, only "observables". You can create an observable per event to listen on, its just a little weird.

I think we should skip the threads library for now. Or connect to the raw's Worker events for events, and use threads for an easy wrapper for direct calls from the main thread

wemeetagain:
Discv5 done, see #4988

wemeetagain:
Looking into moving the network into a thread. Some annoyances/hurdles (that could be treated as refactoring prereqs):
- ~some consensus objects can't be `structuredClone`'d. Any BitArray gets converted into an object.~
  - ~convert BitArray methods to functions?~
[...]

PR #5131, opened 2023-02-11
Refactor network/libp2p initialization

wemeetagain:
**Motivation**

See discussion in #4856 

Initializing libp2p inside the network class, rather than inside the cli, is a first step towards moving the network into a separate thread.

**Description**

- Refactor the `Network` class to use the async initialization pattern used elsewhere in the codebase (eg: `BeaconNode`)
  - `await Network.init(...)`
  - `await network.close()`
- Initialize libp2p inside `Network.init`
- A circular dependency was uncovered between the attnetsService and gossipsub. Resolved by adding a level of indirection rather than passing the entire network class around.


g11tech (approved):
lgtm!

wemeetagain:

PR #5139, opened 2023-02-13
Fix the zero finalized light client header upgrade

g11tech:
Starting on a checkpoint sync, when the finalized header for the light client update store will be zero (till a finalization happens post startup), the upgrade would fail as zero finalized header's slot (use to upgrade from) is 0 and would result in such kind of error as reported by @philknows 

image

However the upgrade neednot be applied as zero finalized already belongs to the highest fork, this PR fixes the same


wemeetagain (approved):

wemeetagain:

PR #5143, opened 2023-02-13
`structuredClone`able INetworkOptions

wemeetagain:
**Motivation**

See description in #4856 
If the Network class is to move to a thread, its options need to be structuredClone-able (ref: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm).

**Description**

Ensure (by hand) that the INetworkOptions is structuredClone-able.


dapplion (approved):

wemeetagain:

PR #5154, opened 2023-02-14
Add chain.archiveStateEpochFrequency cli option

wemeetagain:
**Motivation**

A few lodestar consumers have wanted finer grained control over historical state retrieval.
This is a cheap win, since this merely exposes a constant as a configurable value, doesn't require any additional implementation work.

**Description**

- Replace `PERSIST_STATE_EVERY_EPOCHS` constant with `chain.archiveStateEpochFrequency` hidden cli option
- Handle case where `chain.archiveStateEpochFrequency` can now be smaller than `PERSIST_TEMP_STATE_EVERY_EPOCHS`
  - Ensures that `chain.archiveStateEpochFrequency` will not have a de-facto minimum of `PERSIST_TEMP_STATE_EVERY_EPOCHS`

### DIFF
--- a/packages/beacon-node/src/chain/archiver/archiveStates.ts
+++ b/packages/beacon-node/src/chain/archiver/archiveStates.ts
@@ -7,14 +7,17 @@ import {IBeaconDb} from "../../db/index.js";
 import {CheckpointStateCache} from "../stateCache/index.js";
 
 /**
- * Minimum number of epochs between archived states
- */
-export const PERSIST_STATE_EVERY_EPOCHS = 1024;
-/**
  * Minimum number of epochs between single temp archived states
  * These states will be pruned once a new state is persisted
  */
 const PERSIST_TEMP_STATE_EVERY_EPOCHS = 32;
+
+export interface StatesArchiverOpts {
+  /**
+   * Minimum number of epochs between archived states
+   */
+  archiveStateEpochFrequency: number;
+}
 
 /**
  * Archives finalized states from active bucket to archive bucket.
@@ -25,7 +28,8 @@ export class StatesArchiver {
   constructor(
     private readonly checkpointStateCache: CheckpointStateCache,
     private readonly db: IBeaconDb,
-    private readonly logger: ILogger
+    private readonly logger: ILogger,
+    private readonly opts: StatesArchiverOpts
   ) {}
 
   /**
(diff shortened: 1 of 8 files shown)


ec2:
Much awaited feature! Thanks a lot!

twoeths (dismissed):

packages/beacon-node/src/chain/archiver/archiveStates.ts
```diff
@@ -43,22 +47,23 @@ export class StatesArchiver {
   async maybeArchiveState(finalized: CheckpointWithHex): Promise<void> {
     const lastStoredSlot = await this.db.stateArchive.lastKey();
     const lastStoredEpoch = computeEpochAtSlot(lastStoredSlot ?? 0);
+    const archiveStateEpochFrequency = this.opts.archiveStateEpochFrequency;
```
nflaig: could be destructured, depends on what you prefer :)

```suggestion
    const {archiveStateEpochFrequency} = this.opts;
```

packages/beacon-node/src/chain/archiver/archiveStates.ts
```diff
@@ -43,22 +47,23 @@ export class StatesArchiver {
...
 
-    if (finalized.epoch - lastStoredEpoch > PERSIST_TEMP_STATE_EVERY_EPOCHS) {
+    if (finalized.epoch - lastStoredEpoch > Math.min(PERSIST_TEMP_STATE_EVERY_EPOCHS, archiveStateEpochFrequency)) {
```
nflaig, line 52: why is this line changed? seems unrelated to the PR. I know it is probably because `archiveStateEpochFrequency` can now be lower than `PERSIST_TEMP_STATE_EVERY_EPOCHS` but should add explainer in PR

packages/cli/src/options/beaconNodeOptions/chain.ts
```diff
@@ -17,6 +17,7 @@ export interface IChainArgs {
   "chain.assertCorrectProgressiveBalances": boolean;
   "chain.maxSkipSlots": number;
   "safe-slots-to-import-optimistically": number;
+  "chain.archiveStateEpochFrequency": number;
```
nflaig, line 20: could be grouped with other `chain.` args

nflaig (dismissed):
LGTM, just some minor remarks

packages/cli/src/options/beaconNodeOptions/chain.ts
wemeetagain: yeah, debated doing that. Then settled with "add at the bottom"

packages/beacon-node/src/chain/archiver/archiveStates.ts
wemeetagain: Added some words to the PR description

nflaig (approved):

packages/cli/src/options/beaconNodeOptions/chain.ts
```diff
@@ -133,4 +135,11 @@ Will double processing times. Use only for debugging purposes.",
     defaultDescription: String(defaultOptions.chain.safeSlotsToImportOptimistically),
     group: "chain",
   },
+
+  "chain.archiveStateEpochFrequency": {
+    hidden: true,
+    description: "Minimum number of epochs between archived states",
+    type: "number",
+    group: "chain",
+  },
```
dapplion, line 144: I think this flag should be public, with maybe a better name. Lighthouse exposes this and its quite useful to fine tune node behaviour

wemeetagain: Current behavior is annoying in that it isn't "store a state when epoch % N == 0".
It's "maybe store a state after N epochs since the last stored state if the state is finalized".
Not sure what the current Lighthouse behavior is, but I'm guessing its not what we're doing.

Preference to merge this as is (as a hidden option and only semi-supported) since there are people who want this now.
And figure out better behavior later (or do more research to justify the current behavior) and at that point make a public cli option.

dapplion (approved):

wemeetagain:

Issue #5158, opened 2023-02-16
Add VC metric to track validator statuses


nflaig:
**Is your feature request related to a problem? Please describe.**

As a node operator I want to know the status of my validators without relying on 3rd party solutions such as [beaconchai.in](https://beaconcha.in/).

The following questions should be answered

- how many of my validators are active vs just imported without deposit?
- how many validators have a pending deposit?
- how many of my validators exited and was it voluntary or forced?
- what is the state of my withdrawal if exited?

**Describe the solution you'd like**

Add VC metric to track [validator statuses](https://hackmd.io/ofFJ5gOmQpu1jjHilHbdQQ) based on [getValidatorStatus](https://github.com/ChainSafe/lodestar/blob/0607fef8d9b576fd127952122cbf60adc589c629/packages/beacon-node/src/api/impl/beacon/state/utils.ts#L78).

There is already a metric to track total count of validators (`vc_indices_count`), might just need to add `status` label.

https://github.com/ChainSafe/lodestar/blob/0607fef8d9b576fd127952122cbf60adc589c629/packages/validator/src/metrics.ts#L241-L244

This metric should then be visualized on a dashboard, [Lodestar validator client](https://github.com/ChainSafe/lodestar/blob/0607fef8d9b576fd127952122cbf60adc589c629/dashboards/lodestar_validator_client.json) looks like a good candidate as it already displays the VC indices.

Finally, also need to update the metric value used for [`validator_active`](https://github.com/ChainSafe/lodestar/blob/63479b4e9a910c0d0b49fe323d367a12455b7a8a/packages/beacon-node/src/monitoring/clientStats.ts#L185) in the client monitoring to correctly only count the active validators instead of total.

philknows:
This kind of ties into re-thinking a little bit about how the validator UX should look like (#4785) to the end user and how we can improve from our side.

maschad:
Agreed @philknows and I think this can inform how we approach https://github.com/ChainSafe/lodestar/issues/5192

Along with the `status` label we may want to track other metadata like:

- `fee_recipient` set for that validator
- `pubkey`
- `validator index`

maschad:
[Lighthouse's Validator Monitor Metrics](https://github.com/sigp/lighthouse-metrics/blob/master/dashboards/ValidatorMonitor.json#L38) tracks a few things we may be interested in such as:

[...]

Issue #4767, opened 2022-11-15
Installation fails macOS Ventura 13.0.1 version


AmanRaj1608:
**Describe the bug**

While installing and building on new macos version. This is due to the version of [blst-ts package](https://github.com/ChainSafe/blst-ts/issues/70)
Need to bump up from `@chainsafe/blst:^0.2.4` to  `@chainsafe/blst:^0.2.6`

**Expected behavior**
Should Install the packages and build.

**Steps to Reproduce**
- Clone project
- Install via `yarn install --ignore-optional`

**Screenshots**
![image](https://user-images.githubusercontent.com/42104907/201940574-8407e9b4-6328-43fb-b4f3-ae7df46ea892.png)

- OS:  OSX Ventura
- Version: 13.0.1
- Branch: stable / unstable
- Commit hash: 1ceaa20ebef7a7adcf9709c323b128d4b4fbeba6

dapplion:
- Linking https://github.com/ChainSafe/blst-ts/issues/75

wemeetagain:
@AmanRaj1608 can you retry with latest lodestar v1.4.2 ?

patrickmao1:
looks like it's the good ol' "package for darwin-arm64 doesn't exist" issue? 
```
Retrieving BLST native bindings /Users/patrickmao/repos/lodestar/node_modules/@chainsafe/blst/prebuild/darwin-arm64-93-binding.node ...
Error: Error importing BLST native bindings: 404 Not Found
```
Where can I get this prebuild for darwin-arm64?

dapplion:
> looks like it's the good ol' "package for darwin-arm64 doesn't exist" issue?
> 
> ```
> Retrieving BLST native bindings /Users/patrickmao/repos/lodestar/node_modules/@chainsafe/blst/prebuild/darwin-arm64-93-binding.node ...
> Error: Error importing BLST native bindings: 404 Not Found
> ```
> 
> Where can I get this prebuild for darwin-arm64?

We currently don't have the capability to provide darwin-arm64 prebuilds, see https://github.com/ChainSafe/blst-ts#supported-environments

maschad:
> looks like it's the good ol' "package for darwin-arm64 doesn't exist" issue?
> 
> ```
> Retrieving BLST native bindings /Users/patrickmao/repos/lodestar/node_modules/@chainsafe/blst/prebuild/darwin-arm64-93-binding.node ...
[...]

PR #5161, opened 2023-02-17
Fix: Log validator index, status and pubkey each epoch

### LINKED ISSUE #4785: Improve validator status logging

There has been multiple occasions where feedback was given to clarify the meaning of the `Discovered new validators = x` as to what this statement actually means for the user. 

Please correct if not true: Currently, this line is logged when a user starts the validator client. When the user's validator pubkey is active (deposit is recognized from logs) on the beacon chain, that increments this count by 1. 

However there is a period of time, generally when deposits have been made recently on the execution chain (e.g. a new staker and now setting up beacon infrastructure), it can be misleading to see their logs state:

```
info: 3 local keystores
info: 0x000000...
info: 0x000001...
info: 0x000002...
info: Discovered new validator count = 0
```

Although we've clearly stated that the validator client recognizes 3 local keystores successfully decrypted and ready to process duties once activated, `Discovered new validator count = 0` in its current phrasing is counter-intuitive as these are technically new validators that haven't been discovered _yet_ by the beacon chain. This is a sensitive time period for stakers and better UX is justified to reduce confusion. This confusion is common as users generally make deposits to their validators slightly before, during or after setting up their beacon infrastructure. 

In addition, if a user adds an additional key to their client, the way it's phrased confuses a user about whether it reflects "new validators" added since the previous run.  

This is not an issue after pubkeys have been recognized and activated on the beacon chain since the proper number of validator keys will be reflected in the log.

Would like to see if there is better phrasing to reflect the reality of what we're logging to the user. An example:
`Validators currently activated on the beacon chain = x`

Do we need to add a small disclaimer for users about needing to wait for some time so it can be activated?
`(New deposits may take some time for activation)` 

**Additional context**
image

maschad:
Closes https://github.com/ChainSafe/lodestar/issues/4785


packages/validator/src/services/indices.ts
```diff
@@ -76,6 +76,14 @@ export class IndicesService {
   private async pollValidatorIndicesInternal(pubkeysHex: PubkeyHex[]): Promise<ValidatorIndex[]> {
     const pubkeysHexToDiscover = pubkeysHex.filter((pubkey) => !this.pubkey2index.has(pubkey));
 
+    pubkeysHex.forEach((pubkey) => {
+      if (this.pubkey2index.has(pubkey)) {
+        this.logger.info("Validator status: ", {status: "active"});
```
nflaig: do we know that the status is `active` here? This is related to https://github.com/ChainSafe/lodestar/issues/5158, we currently do not keep track of the validator status as far as I am aware

nflaig: I think this should be a one liner, having multiple lines is not that easy to read in my opinion, also it might happen that they are not grouped together, generally can't rely on the order of logs as they are written asynchronously

```suggestion
        this.logger.info("Validator", {index: this.pubkey2index.get(pubkey), status: "active", pubkey}});
```

nflaig (changes requested):

nflaig (dismissed):
looks good to me from a technical point of view, I am not really opinionated on how the log message should be displayed. Since this is a UX topic makes sense to get actual user feedback so we might just wanna push this out asap.

maschad:
Would be good to get @g11tech opinion on this about log formatting

wemeetagain (changes requested):
https://github.com/ChainSafe/lodestar/issues/4785 says
every epoch, we should log the pubkey count with the validator grouped by status
as lighthouse is doing

I don't think this is doing that. Also don't think this is necessarily the right place to add the log.

packages/validator/src/services/indices.ts
```diff
@@ -108,6 +108,13 @@ export class IndicesService {
         this.pubkey2index.set(pubkeyHex, validatorState.index);
         this.index2pubkey.set(validatorState.index, pubkeyHex);
         newIndices.push(validatorState.index);
+      } else {
+        this.logger.info(
+          `Validator:
```
wemeetagain: We try to format logs like:
`logger[method](message, obj)`
Where we think about each of the three params:
- method - very conscious about the info / error logs we add as they can overwhelm, confuse or scare users
- message - Informative but not overly verbose
- obj - a single-level object (check its type) with useful and relevant metadata

so this one could look like:
`logger.info("Discovered validator", {status: validatorState.status ?? "unknown, index: this.pubkey2index.get(pubkeyHex), pubkey: pubkeyHex})`

g11tech:
> #4785 says every epoch, we should log the pubkey count with the validator grouped by status as lighthouse is doing
> 
> I don't think this is doing that. Also don't think this is necessarily the right place to add the log.

yes #4785 is meant to do as suggested by @wemeetagain :+1:

packages/validator/src/services/indices.ts
```diff
@@ -2,6 +2,7 @@ import {ValidatorIndex} from "@lodestar/types";
 import {Logger} from "@lodestar/utils";
 import {toHexString} from "@chainsafe/ssz";
 import {Api, ApiError} from "@lodestar/api";
+import {ValidatorResponse, ValidatorStatus} from "@lodestar/api/lib/beacon/routes/beacon/state.js";
```
nflaig: why is this imported like this? shouldnt this be just `@lodestar/api` and if that it not possible I think we need to properly re-export from api package

packages/validator/src/services/indices.ts
```diff
@@ -110,6 +116,48 @@ export class IndicesService {
...
+
+    logValidatorStatuses(allValidators, this.logger);
+
     return newIndices;
   }
 }
+
+function logValidatorStatuses(allValidators: Map<ValidatorStatus, ValidatorResponse[]>, logger: Logger): void {
+  allValidators.get("active")?.forEach((validatorState: ValidatorResponse) => {
+    logger.info("Validator activated", {
```
nflaig: this code block is repeated mutliple times and really only the message (`"Validator activated"`) is changed, if we extract this it is ensure we use the same format for all related logs and only need to update in one place if we decide to change the log format later on

```ts
logger.info("Validator activated", {
      pubkey: toHexString(validatorState.validator.pubkey),
      index: validatorState.index,
});
```

packages/validator/src/services/indices.ts
```diff
@@ -110,6 +116,48 @@ export class IndicesService {
...
+      pubkey: toHexString(validatorState.validator.pubkey),
+      index: validatorState.index,
+    });
+  });
+
+  [...(allValidators.get("pending_initialized") ?? []), ...(allValidators.get("pending_queued") ?? [])]?.forEach(
```
nflaig: IMO this line is a bit hard to read and reason about, the logic is also repeated in multiple places

maybe can do something like this:

add a function/method (maybe should use class method and store allValidators as class property? and then access via `this.allValidators` instead of passing it into multiple functions)

```ts
function getValidatorStatesByStatuses(statuses: ValidatorStatus[], allValidators: Map): ValidatorResponse[] {
   return statuses.filter((status)=> allValidators.has(status)).map((status) => allValidators.get(status));
}

```

and then here we can just do

```suggestion
  getValidatorStatesByStatuses(["pending_initialized", "pending_queued"]).forEach(
```

packages/validator/src/services/indices.ts
```diff
@@ -110,6 +116,48 @@ export class IndicesService {
...
+  allValidators.get("withdrawal_done")?.forEach((validatorState: ValidatorResponse) => {
+    logger.info("Validator withdrawn", {
+      pubkey: toHexString(validatorState.validator.pubkey),
+      index: validatorState.index,
+    });
+  });
+
+  [...(allValidators.get("exited_slashed") ?? []), ...(allValidators.get("exited_unslashed") ?? [])]?.forEach(
+    (validatorState: ValidatorResponse) => {
+      logger.info("Validator Exited", {
```
nflaig: is the capital `E` intended here? seems inconsistent compared to other logs

```suggestion
      logger.info("Validator exited", {
```

packages/validator/src/services/indices.ts
```diff
@@ -110,6 +116,48 @@ export class IndicesService {
...
+        pubkey: toHexString(validatorState.validator.pubkey),
+        index: validatorState.index,
+      });
+    }
+  );
+
+  logger.info("Initialized Validators: ", {
```
nflaig: usually colons are not used for such kind of logs any specific reason why it is used here?

and to be more consistent with existing logs such as `Discovered validator` should probably also use lower case `v`

```suggestion
  logger.info("Initialized validators", {
```

packages/validator/src/services/indices.ts
```diff
@@ -110,6 +116,48 @@ export class IndicesService {
...
+    total: allValidators.size,
+    available: allValidators.get("active")?.length,
```
nflaig: why is the term `available` used here instead of `active`? IMO this should be called `active`

```suggestion
    active: allValidators.get("active")?.length,
```

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -1,9 +1,10 @@
 import {Epoch, bellatrix} from "@lodestar/types";
-import {Api, ApiError, routes} from "@lodestar/api";
-import {BeaconConfig} from "@lodestar/config";
+import {Api, ApiError} from "@lodestar/api";
 import {SLOTS_PER_EPOCH} from "@lodestar/params";
+import {BeaconConfig} from "@lodestar/config";
 
-import {IClock, LoggerVc, batchItems} from "../util/index.js";
+import {ProposerPreparationData} from "@lodestar/api/lib/beacon/routes/validator.js";
```
nflaig: same as above, import from `"@lodestar/api"` and should avoid deep imports from  `lib` folder of packages

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -37,13 +38,19 @@ export function pollPrepareBeaconProposer(
...
+
+          proposers.push({
             validatorIndex: String(index as number),
-            feeRecipient: validatorStore.getFeeRecipientByIndex(index),
-          })
-        );
+            feeRecipient,
+          });
+
+          logger.info("Validator exists in Beacon Chain", {validatorIndex: index, feeRecipient, pubKey});
```
nflaig: to be considered if `Beacon Chain` should use upper case, based on current logs we rarely/never do this

```suggestion
          logger.info("Validator exists in beacon chain", {validatorIndex: index, feeRecipient, pubKey});
```

nflaig (changes requested):

maschad:
Thanks for the feedback @nflaig :)  made the updates!

packages/validator/src/services/indices.ts
```diff
@@ -110,6 +115,44 @@ export class IndicesService {
...
+    getValidatorStatesByStatuses(loggedValidatorStates[key], allValidators).forEach(
+      (validatorState: ValidatorResponse) => {
+        logFormatter(key, validatorState);
+      }
+    );
+  });
+
+  logger.info("Initialized validators", {
+    total: allValidators.size,
+    active: allValidators.get("active")?.length,
```
nflaig: what do we want to log here if there are no `active` validators? I think it should be the number `0`

```suggestion
    active: allValidators.get("active")?.length ?? 0,
```

packages/validator/src/services/indices.ts
```diff
@@ -110,6 +115,44 @@ export class IndicesService {
...
+      index: validatorState.index,
+    });
+  };
+  Object.keys(loggedValidatorStates).map((key) => {
+    getValidatorStatesByStatuses(loggedValidatorStates[key], allValidators).forEach(
+      (validatorState: ValidatorResponse) => {
+        logFormatter(key, validatorState);
+      }
+    );
+  });
```
nflaig: If you want to further improve readability :)

```suggestion
  for (const [message, statuses] of Object.entries(loggedValidatorStates)) {
    getValidatorStatesByStatuses(statuses, allValidators).forEach((validatorState) => {
      logFormatter(message, validatorState);
    });
}
```

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -1,9 +1,10 @@
 import {Epoch, bellatrix} from "@lodestar/types";
-import {Api, ApiError, routes} from "@lodestar/api";
-import {BeaconConfig} from "@lodestar/config";
+import {Api, ApiError} from "@lodestar/api";
 import {SLOTS_PER_EPOCH} from "@lodestar/params";
+import {BeaconConfig} from "@lodestar/config";
 
-import {IClock, LoggerVc, batchItems} from "../util/index.js";
+import {ProposerPreparationData} from "@lodestar/api";
```
nflaig: `ProposerPreparationData` should probably be imported with the other `from "@lodestar/api";` imports on line 2

nflaig (commented):
Looks pretty clean now, great refactoring!

packages/validator/src/services/indices.ts
maschad: Good idea! Thanks

packages/validator/src/services/prepareBeaconProposer.ts
maschad: Thanks my `yarn install` is broken because of https://github.com/ChainSafe/lodestar/issues/4767 so I think my ts auto-imports are messed up

packages/validator/src/services/indices.ts
maschad: [Not possible at the moment](https://github.com/ChainSafe/lodestar/actions/runs/4245050914/jobs/7380036254#step:7:29)  but it can be addressed in another PR

packages/validator/src/services/prepareBeaconProposer.ts
maschad: [It's not properly exported](https://github.com/ChainSafe/lodestar/actions/runs/4245050914/jobs/7380036254#step:7:32) we should create an issue for this to address it in a separate PR

nflaig: Not sure if the extra effort would be worth it and if it is done in this PR someone looking at why this was exported will see it here. But would like to hear one of the maintainers opinion on this.

packages/validator/src/services/indices.ts
```diff
@@ -110,6 +116,49 @@ export class IndicesService {
...
       }
     }
+
+    logValidatorStatuses(allValidators, this.logger);
+
     return newIndices;
   }
 }
+
+function logValidatorStatuses(allValidators: Map<ValidatorStatus, ValidatorResponse[]>, logger: Logger): void {
```
dapplion: We usually sort modules first, arguments second. May be good to add to contributing guidelines

packages/validator/src/services/indices.ts
```diff
@@ -2,6 +2,7 @@ import {ValidatorIndex} from "@lodestar/types";
 import {Logger} from "@lodestar/utils";
 import {toHexString} from "@chainsafe/ssz";
 import {Api, ApiError} from "@lodestar/api";
+import {ValidatorStatus, ValidatorResponse} from "@lodestar/api/lib/beacon/routes/beacon/state.js";
```
wemeetagain: I think we've been doing somehting like
```ts
import {routes} from "@lodestar/api"

const x: routes.beacon.ValidatorStatus;
```

packages/validator/src/services/indices.ts
```diff
@@ -101,7 +101,13 @@ export class IndicesService {
...
+
+    const allValidators = new Map<ValidatorStatus, ValidatorResponse[]>();
+
     for (const validatorState of res.response.data) {
+      // Group all validators by status
+      allValidators.set(validatorState.status, [validatorState]);
+
       const pubkeyHex = toHexString(validatorState.validator.pubkey);
       if (!this.pubkey2index.has(pubkeyHex)) {
         this.logger.debug("Discovered validator", {pubkey: pubkeyHex, index: validatorState.index});
```
wemeetagain: This log should be replaced with the "Validator exists in beacon chain" log.
```suggestion
        this.logger.debug("Validator exists in beacon chain", {validatorIndex: validatorState.index, feeRecipient, pubkey: pubkeyHex});
```

packages/validator/src/services/indices.ts
```diff
@@ -101,7 +101,13 @@ export class IndicesService {
     ApiError.assert(res, "Can not fetch state validators from beacon node");
 
     const newIndices = [];
+
+    const allValidators = new Map<ValidatorStatus, ValidatorResponse[]>();
+
     for (const validatorState of res.response.data) {
+      // Group all validators by status
+      allValidators.set(validatorState.status, [validatorState]);
```
wemeetagain: This set will overwrite previous sets when validators share a status.

packages/validator/src/services/indices.ts
```diff
@@ -110,6 +116,49 @@ export class IndicesService {
...
+
+  const logFormatter = (message: string, validatorState: ValidatorResponse): void => {
+    logger.info(message, {
+      pubkey: toHexString(validatorState.validator.pubkey),
+      index: validatorState.index,
+    });
+  };
+
+  for (const [message, statuses] of Object.entries(loggedValidatorStates)) {
+    getValidatorStatesByStatuses(statuses, allValidators).forEach((validatorState) => {
```
wemeetagain: project coding preferences thing: we try to avoid `x.forEach()` and instead do `for (const y of x) {`

packages/validator/src/services/indices.ts
```diff
@@ -110,6 +116,49 @@ export class IndicesService {
...
+
+function logValidatorStatuses(allValidators: Map<ValidatorStatus, ValidatorResponse[]>, logger: Logger): void {
+  const loggedValidatorStates: {[key: string]: ValidatorStatus[]} = {
+    "Validator activated": ["active"],
+    "Validator pending": ["pending_initialized", "pending_queued"],
+    "Validator withdrawn": ["withdrawal_done"],
+    "Validator exited": ["exited_slashed", "exited_unslashed"],
+  };
+
+  const logFormatter = (message: string, validatorState: ValidatorResponse): void => {
```
wemeetagain: This function appears to be only used once and is just a single `logger.info`. Preference would be to avoid one-line functions like this unless there's a good reason (reuse of logic, etc).

packages/validator/src/services/indices.ts
```diff
@@ -110,6 +116,49 @@ export class IndicesService {
...
+    getValidatorStatesByStatuses(statuses, allValidators).forEach((validatorState) => {
+      logFormatter(message, validatorState);
+    });
+  }
+
+  logger.info("Initialized validators", {
+    total: allValidators.size,
+    active: allValidators.get("active")?.length ?? 0,
+  });
+}
```
wemeetagain: We should avoid printing one line per validator here.
Can you refactor the one log line to be:
```ts
logger.info("Current validator statuses", {
  active: 5,
  pending: 4,
  exited: 4,
  withdrawn: 3,
  total: 16,
});
```
Be sure to only print the each status total if the status is non-zero.

packages/validator/src/services/indices.ts
maschad: not sure what you mean by this

wemeetagain (changes requested):

packages/validator/src/services/indices.ts
maschad: Good catch! Thanks

packages/validator/src/services/indices.ts
nflaig: can just make `logValidatorStatuses` a private method of the class then we do not have to worry about the order of function parameters here. Having `allValidators` as persistent as private property of the class might also make it easier to implement https://github.com/ChainSafe/lodestar/issues/5158. Though could refactor that later once we know how to add the metric for this.

packages/validator/src/services/indices.ts
```diff
@@ -2,6 +2,7 @@ import {ValidatorIndex} from "@lodestar/types";
 import {Logger} from "@lodestar/utils";
 import {toHexString} from "@chainsafe/ssz";
 import {Api, ApiError} from "@lodestar/api";
+import {routes} from "@lodestar/api";
```
nflaig: can be added to the imports one line above

packages/validator/src/services/indices.ts
nflaig: shouldn't we show `total` first?

packages/validator/src/services/indices.ts
maschad: What's the significance of the order of the function parameters? The reason I didn't make it a class method was because I don't think it relates directly to the Indices service but rather it is a helper function.

nflaig: > The reason I didn't make it a class method was because I don't think it relates directly to the Indices service but rather it is a helper function.

IMO if the function is just used in this class would go for a private method. Also isn't one of the jobs of the indices service to print out information about the validators? Either way is fine, just personal preference

packages/validator/src/services/indices.ts
```diff
@@ -101,15 +101,73 @@ export class IndicesService {
...
+
+  for (const [message, statuses] of Object.entries(loggedValidatorStates)) {
+    for (const validatorState of getValidatorStatesByStatuses(statuses, allValidators)) {
+      logger.info(message, {
+        pubkey: toHexString(validatorState.validator.pubkey),
+        index: validatorState.index,
+      });
+    }
+  }
+
```
wemeetagain:
```suggestion
```
We should avoid printing one line per validator here. We can address printing individual validator lifecycle logs later, but this isn't the right place (will be reprinted every epoch!)

When you do this, you can make an `type SimpleValidatorStatus = "active" | "pending" ...;`, make `allValidators` a `MapDef<SimpleValidatorStatus, number>`, and directly increment each value in `allValidators`.
Should simplify the code a lot and will no longer need `getValidatorStatesByStatuses`.

packages/validator/src/services/indices.ts
```diff
@@ -101,15 +101,73 @@ export class IndicesService {
...
+  logger.info("Validator statuses", {
+    active: allValidators.get("active")?.length ?? 0,
+    pending: getValidatorStatesByStatuses(loggedValidatorStates["Validator pending"], allValidators)?.length ?? 0,
+    exited: getValidatorStatesByStatuses(loggedValidatorStates["Validator exited"], allValidators)?.length ?? 0,
+    withdrawn: getValidatorStatesByStatuses(loggedValidatorStates["Validator withdrawn"], allValidators)?.length ?? 0,
```
wemeetagain: only print the each status total if the status is non-zero.
Something like
```ts
const statuses: Record<SimpleStatus, number> = Object.fromEntries(
  Array.from(allValidators.entries()).filter((entry) => entry[1] > 0)
);
const total = Object.values(statuses).reduce((a, b) => a + b, 0);
logger.info("Validator statuses", {total, ...statuses});
```

packages/validator/src/services/indices.ts
```diff
@@ -101,15 +101,73 @@ export class IndicesService {
     ApiError.assert(res, "Can not fetch state validators from beacon node");
 
     const newIndices = [];
+
+    const allValidators = new Map<routes.beacon.ValidatorStatus, routes.beacon.ValidatorResponse[]>();
```
wemeetagain: You can use a `MapDef` here
```ts
import {MapDef} from "@lodestar/utils"
```
Then below you will be able to do something like:
```ts
allValidators.getOrDefault(simpleStatuses[validatorState.status]).push(validatorState);
```

wemeetagain (changes requested):

packages/validator/src/services/indices.ts
maschad: Ok I will address the lifecycle logs in another PR, I had though this was desired.

packages/validator/src/services/indices.ts
maschad: Good idea,  I still feel like this could be more readable but I can't think of a way atm.

packages/validator/src/services/indices.ts
```diff
@@ -101,15 +103,34 @@ export class IndicesService {
     ApiError.assert(res, "Can not fetch state validators from beacon node");
 
     const newIndices = [];
+
+    const allValidators = new MapDef<SimpleValidatorStatus, number>(() => 0);
+
     for (const validatorState of res.response.data) {
+      // Group all validators by status
+      allValidators.set(
+        validatorState.status as SimpleValidatorStatus,
```
wemeetagain: You will need to map from validator status to simple validator status.
I recommend defining SimpleValidatorStatus separately and independently from validator status.

Something like:
```ts
type SimpleValidatorStatus = "pending" | "active" | "exited" | "withdrawn";
const statusToSimpleStatus: (status: ValidatorStatus): SimpleValidatorStatus => {
  switch (status) {
    case x:
    case y:
    case z:
      return "active";
    ...
  }
}
```

Then here:
```suggestion
        statusToSimpleStatus(validatorState.status),
```

packages/validator/src/services/indices.ts
```diff
@@ -101,15 +103,34 @@ export class IndicesService {
...
+          pubkey: pubkeyHex,
+        });
         this.pubkey2index.set(pubkeyHex, validatorState.index);
         this.index2pubkey.set(validatorState.index, pubkeyHex);
         newIndices.push(validatorState.index);
       }
     }
+
+    // Filter out validator statuses that are greater than 0
+    const statuses = Object.fromEntries(Array.from(allValidators.entries()).filter((entry) => entry[1] > 0));
```
wemeetagain: Should also conditionally add "pending" validators (validators not known by the beacon chain) to `statuses`.

```ts
const pendingCount = pubkeysHex.length - res.response.data.length;
if (pendingCount > 0) {
  statuses["pending"] = pendingCount;
}
```

wemeetagain (changes requested):

packages/validator/src/services/indices.ts
maschad: Thanks for the suggestion, upon reconsideration is this substantially more readable than [what we had previously?](https://github.com/ChainSafe/lodestar/pull/5161/commits/f1f0e36265ebe457d6230f080db1100f2ecf954b#diff-eb9b1b96c2b62069854c2962a24c1983f831836c63455470fe0d0d78c8f5812cR126-R144) 

Given this will require a mapping and a function to set these values,  I could have just updated the keys in the [previous mapping](https://github.com/ChainSafe/lodestar/pull/5161/commits/f1f0e36265ebe457d6230f080db1100f2ecf954b#diff-eb9b1b96c2b62069854c2962a24c1983f831836c63455470fe0d0d78c8f5812cR126-R130)  to be:

```typescript
const loggedValidatorStates: {[key: string]: ValidatorStatus[]} = {
    "activated": ["active"],
    "pending": ["pending_initialized", "pending_queued"],
    "withdrawn": ["withdrawal_done"],
    "exited": ["exit_slashed", "exit_unslashed"],
  };
```  
and then had:

```typescript
	const statuses = {}

	for (const key in loggedValidatorStates) {
	  const values = loggedValidatorStates[key];
	  let total = 0;
	  for (const value of values) {
	    if (allValidatorStates[value] !== undefined) {
	      total += allValidatorStates[value].length;
	    }
	  }
	  result[key] = total;
	}
    this.logger.info("Validator statuses", {...statuses}); // Output: { activated: 60, pending: 0, .... }
```

I will make these changes now but I think we could have had it either way.

packages/validator/src/services/indices.ts
```diff
@@ -14,6 +14,26 @@ const PUBKEYS_PER_REQUEST = 10;
...
+    return "exited";
+  }
+  if (status.includes("pending")) {
+    return "pending";
+  }
+  if (status === "withdrawal_done") {
+    return "withdrawn";
+  }
+
+  return "unknown";
```
wemeetagain: Every ValidatorStatus has a SimpleValidatorStatus that it maps to, there shouldn't be an unknown

packages/validator/src/services/indices.ts
```diff
@@ -14,6 +14,26 @@ const PUBKEYS_PER_REQUEST = 10;
...
 type PubkeyHex = string;
 
+// To assist with logging statuses, we only log the statuses that are not active_exiting or withdrawal_possible
+type SimpleValidatorStatus = "pending" | "active" | "exited" | "withdrawn" | "unknown";
+
+const statusToSimpleStatusMapping = (status: routes.beacon.ValidatorStatus): SimpleValidatorStatus => {
+  if (status.includes("active") || status === "withdrawal_possible") {
+    return "active";
+  }
+  if (status.includes("exiting")) {
```
wemeetagain: ~~There is no "exiting_*" ValidatorStatus...~~ This is missing some exited statuses, It may be better to just use a switch case here

packages/validator/src/services/indices.ts
```diff
@@ -14,6 +14,28 @@ const PUBKEYS_PER_REQUEST = 10;
...
+// To assist with logging statuses, we only log the statuses that are not active_exiting or withdrawal_possible
+type SimpleValidatorStatus = "pending" | "active" | "exited" | "withdrawn";
+
+const statusToSimpleStatusMapping = (status: routes.beacon.ValidatorStatus): SimpleValidatorStatus => {
+  switch (status) {
+    case "active":
+    case "active_exiting":
+    case "active_slashed":
+    case "active_ongoing":
+    case "withdrawal_possible":
```
wemeetagain: withdrawal_possible is exited

packages/validator/src/services/indices.ts
```diff
@@ -101,15 +122,37 @@ export class IndicesService {
...
         newIndices.push(validatorState.index);
       }
     }
+
+    // Filter out validator statuses that are greater than 0
+    const statuses = Object.fromEntries(Array.from(allValidators.entries()).filter((entry) => entry[1] > 0));
+    // The number of validators that are not in the beacon chain
+    const pendingCount = pubkeysHex.length - res.response.data.length;
+    if (pendingCount > 0) {
+      statuses["pending"] += pendingCount;
```
wemeetagain: This will be NaN if there hasn't been at least one pending entry.
Maybe do this before making `statuses` so you can use the `allValidators.set(s, allValidators.getOrDefault(s) + pendingCount)` pattern

packages/validator/src/services/indices.ts
```diff
@@ -101,15 +122,37 @@ export class IndicesService {
...
+          validatorIndex: validatorState.index,
+          pubkey: pubkeyHex,
+        });
         this.pubkey2index.set(pubkeyHex, validatorState.index);
         this.index2pubkey.set(validatorState.index, pubkeyHex);
         newIndices.push(validatorState.index);
       }
     }
+
+    // Filter out validator statuses that are greater than 0
```
wemeetagain: comment is reversed

packages/validator/src/services/indices.ts
```diff
@@ -101,15 +122,37 @@ export class IndicesService {
...
+    const statuses = Object.fromEntries(Array.from(allValidators.entries()).filter((entry) => entry[1] > 0));
+    // The number of validators that are not in the beacon chain
+    const pendingCount = pubkeysHex.length - res.response.data.length;
+    if (pendingCount > 0) {
+      statuses["pending"] += pendingCount;
+    }
+    // The total number of validators
+    const total = Object.values(statuses).reduce((a, b) => a + b, 0);
```
wemeetagain: realized this is also `pubkeysHex.length` :joy:

packages/validator/src/services/indices.ts
```diff
@@ -101,15 +122,37 @@ export class IndicesService {
     ApiError.assert(res, "Can not fetch state validators from beacon node");
 
     const newIndices = [];
+
+    const allValidators = new MapDef<SimpleValidatorStatus, number>(() => 0);
```
wemeetagain: small thing, this could be renamed to allStatuses or something similar

packages/validator/src/services/indices.ts
maschad: That's for clarifying.

packages/validator/src/services/indices.ts
```diff
@@ -101,15 +125,37 @@ export class IndicesService {
...
       }
     }
+
+    // Retrieve the number of validators for each status
+    const statuses = Object.fromEntries(Array.from(allValidatorStatuses.entries()).filter((entry) => entry[1] > 0));
+    // The number of validators that are not in the beacon chain
+    const pendingCount = pubkeysHex.length - res.response.data.length;
+    if (pendingCount > 0) {
+      allValidatorStatuses.set("pending", allValidatorStatuses.getOrDefault("pending") + pendingCount);
+    }
```
wemeetagain: This needs to be before L149...
Also debatably remove the if (pendingCount > 0) condition

maschad: Good point I think it can be removed since we our default function sets it to zero if it's undefined.

packages/validator/src/services/prepareBeaconProposer.ts
```diff
@@ -37,13 +38,19 @@ export function pollPrepareBeaconProposer(
...
+
+          proposers.push({
             validatorIndex: String(index as number),
-            feeRecipient: validatorStore.getFeeRecipientByIndex(index),
-          })
-        );
+            feeRecipient,
+          });
+
+          logger.info("Validator exists in beacon chain", {validatorIndex: index, feeRecipient, pubKey});
```
wemeetagain: Can you revert the changes in this file?
There's already the other "Validator exists in beacon chain" log

maschad: Given this includes the `feeReceipient` as well, perhaps we should remove the [_debug log_](https://github.com/ChainSafe/lodestar/pull/5161/files/50825d16cc85a747ce41c21f1f1a990fd799112d#diff-eb9b1b96c2b62069854c2962a24c1983f831836c63455470fe0d0d78c8f5812cR138-R141) in favour of this _info log_

nflaig: printing out the fee recipient would be great, at the moment I always have to query the keymanager api to check this

wemeetagain: This is not the right place to add this log. This will get run every epoch XD
And its where we're preparing for epoch proposers.

Really recommend reverting all changes to this file and attempting the log in the other file. Really even `fetchValidatorIndices` isn't probably the best place for it, but at least it will have the right behavior (print once upon discovery)

wemeetagain (dismissed):

packages/validator/src/services/indices.ts
```diff
@@ -101,15 +125,40 @@ export class IndicesService {
...
+
     for (const validatorState of res.response.data) {
+      // Group all validators by status
+      const status = statusToSimpleStatusMapping(validatorState.status);
+      allValidatorStatuses.set(status, allValidatorStatuses.getOrDefault(status) + 1);
+
       const pubkeyHex = toHexString(validatorState.validator.pubkey);
       if (!this.pubkey2index.has(pubkeyHex)) {
-        this.logger.debug("Discovered validator", {pubkey: pubkeyHex, index: validatorState.index});
+        //#TODO: We want to include the fee_recipient here, but it's not available in the API
```
nflaig: general thing, I really dislike TODOs in code without an according issue, maybe can introduce something like `TODO(#<issue_number>)` , this approach is for example used by Prysm and I think it is a good idea because from my experience TODOs a lot the time are just excuses not to do things right away and a lot of the time TODOs are there for years...even decades in some code bases

If we connect an issue to the TODO we can keep track of it and if we close the issue because the problem was resolved or it is no longer relevant then we can also remove the TODO in the code.

maschad: Good point I will link the issue

nflaig: Still questionable to me if we need TODOs here at all or if the issues are sufficient and if the issues are not sufficient maybe we need to add more context to them instead.

To be fair, I really hate TODOs with a passion, so that might just be me. In my previous projects I would not allow a branch to be merged into the main branch until all TODOs were resolved.

maschad: The comment is to guide the implementation details which are also mentioned to a degree in the issue, but there needs to be a wider discussion on how to achieve this which will happen on the issue.

wemeetagain: IMO the TODOs aren't really needed and the issues would be sufficient.
@maschad and I have discussed this PR and the related issues at length and I feel he has an understanding of what needs to be changed without tagging these lines explicitly as TODO.

nflaig (approved):

wemeetagain (approved):

packages/validator/src/services/indices.ts
```diff
@@ -101,15 +125,39 @@ export class IndicesService {
...
+
     for (const validatorState of res.response.data) {
+      // Group all validators by status
+      const status = statusToSimpleStatusMapping(validatorState.status);
+      allValidatorStatuses.set(status, allValidatorStatuses.getOrDefault(status) + 1);
+
       const pubkeyHex = toHexString(validatorState.validator.pubkey);
       if (!this.pubkey2index.has(pubkeyHex)) {
-        this.logger.debug("Discovered validator", {pubkey: pubkeyHex, index: validatorState.index});
+        this.logger.info("Validator exists in beacon chain", {
```
nflaig, line 138: @maschad running this on goerli with quite a lot of validators and this inital log is quite noisy in my opinion, we are printing out `Validator exists in beacon chain` for each validator which seems bit redundant because these logs all happen after each other.

I think we should reconsider this and maybe only print out `Validator exists in beacon chain` once and then just list all the validators, with index, pubkey and ideally also fee recipient (if possible)

maschad: Good point

maschad: Good point, I've included that change in be5a7e4ae5f9e70c62a3b46d2639e35b3ccedd30

wemeetagain:

packages/validator/src/services/indices.ts
```diff
@@ -101,15 +125,39 @@ export class IndicesService {
...
+
+    allValidatorStatuses.set("pending", allValidatorStatuses.getOrDefault("pending") + pendingCount);
+
+    // Retrieve the number of validators for each status
+    const statuses = Object.fromEntries(Array.from(allValidatorStatuses.entries()).filter((entry) => entry[1] > 0));
+
+    // The total number of validators
+    const total = pubkeysHex.length;
+
+    this.logger.info("Validator statuses", {...statuses, total});
```
nflaig, line 159: @maschad wondering here, wasn't this intended to be logged every epoch? 

Once we discovered all pubkey we are not calling this due to this check

https://github.com/ChainSafe/lodestar/blob/ddc58cebfc0cbe66b29e2c83124c1e98bb11874f/packages/validator/src/services/indices.ts#L102-L106

I guess this should be fixed by the other PR https://github.com/ChainSafe/lodestar/pull/5239

PR #5164, opened 2023-02-19
Implement verge spec feature

### LINKED ISSUE #4681: verkle trees: add 2 fields to ExecutionData
With the verge, proofs and witness of the pre-state of an (EL) block execution are included in the blocks.  In order to build a PoS testnet that provides proofs in blocks, it would be great if it were possible to create a branch with the required fields to be added to the `ExecutionData`.

The `ExecutionData` needs to have two more, optional fields:

 * `VerkleProof`, a byte array of variable length
 * `VerkleKeyVals`, an array of (key, value) pairs, in which the key is a 32-byte vector, and value is a byte vector of length either 0 (missing value) or 32.

At the moment, there is no need to support forking as for this testnet, verkle trees are used since the genesis.

If present, these two fields should be propagated over the network with the rest of the ExecutionData so that other clients can inject these fields in their verkle-enabled EL.

dapplion:
**Motivation**

- Develop "verge" spec feature on top of Deneb. https://github.com/ChainSafe/lodestar/issues/4681
- Spec: https://github.com/ethereum/consensus-specs/pull/3230

@gballet current spec is based on capella. Are there plans to rebase it on deneb?

**Description**

This fork currently only updates the execution payload type. There's no additional logic required in consensus side.

Closes https://github.com/ChainSafe/lodestar/issues/4681

### DIFF
--- a/packages/beacon-node/test/spec/presets/fork.ts
+++ b/packages/beacon-node/test/spec/presets/fork.ts
@@ -4,6 +4,7 @@ import {
   CachedBeaconStateAltair,
   CachedBeaconStatePhase0,
   CachedBeaconStateCapella,
+  CachedBeaconStateDeneb,
 } from "@lodestar/state-transition";
 import * as slotFns from "@lodestar/state-transition/slot";
 import {phase0, ssz} from "@lodestar/types";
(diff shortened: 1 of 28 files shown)


g11tech:
superseded by  
- https://github.com/ChainSafe/lodestar/pull/5805

PR #5169, opened 2023-02-20
Expand block processor dashboard

dapplion:
**Motivation**

For better experience with block processor dashboard metrics

**Description**

- Expand all rows of the block processor dashboard
- Add 2 new charts, see below

Screenshot from 2023-02-20 10-34-12

PR #5171, opened 2023-02-20
Improve state transition metrics

dapplion:
**Motivation**

Current time metrics miss 2 important time spenders in state transition function:
- commit step
- hashTreeRoot step

**Description**

- Increase the timer range to capture commit step
- Time hashTreeRoot step
- Time commit step individually
- Track if caches are populated after running state transition function
- Rename to more "normal" names

@tuyennhv note that time metrics for the state transition will likely be worse since those were not accounting for all steps

PR #5172, opened 2023-02-20
Increase chance of ssz cache populated

dapplion:
**Motivation**

- Metrics from https://github.com/ChainSafe/lodestar/pull/5169 show that the internal SSZ cache of state data structures is almost never populated. Their design should ensure that unless there's heavy forking they should be populated in high probability. This PR intends to research the problem while being deployed in testing fleet

**Description**

Transfer cache to the cloned state only when importing blocks

### DIFF
--- a/packages/beacon-node/src/chain/blocks/importBlock.ts
+++ b/packages/beacon-node/src/chain/blocks/importBlock.ts
@@ -5,6 +5,7 @@ import {
   CachedBeaconStateAltair,
   computeEpochAtSlot,
   computeStartSlotAtEpoch,
+  isStateValidatorsNodesPopulated,
   RootCache,
 } from "@lodestar/state-transition";
 import {routes} from "@lodestar/api";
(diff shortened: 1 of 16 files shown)


packages/state-transition/src/epoch/computeUnrealizedCheckpoints.ts
```diff
@@ -33,8 +34,8 @@ export function computeUnrealizedCheckpoints(
 
     // Clone state and use progressive balances
     else {
-      // Clone state to mutate below           false = do not transfer cache
-      stateRealizedCheckpoints = state.clone(false);
+      // Clone state to mutate below         true = do not transfer cache
+      stateRealizedCheckpoints = state.clone(true);
```
dapplion, line 38: This is the main fix: to prevent transferring the cache to this temporary state `true` must be passed, not `false`

twoeths: 👍

packages/beacon-node/src/chain/blocks/importBlock.ts
```diff
@@ -329,6 +330,10 @@ export async function importBlock(
   // This adds the state necessary to process the next block
   this.stateCache.add(postState);
 
+  if (!isStateValidatorsNodesPopulated(postState)) {
+    this.logger.verbose("After importBlock caching postState without SSZ cache", {slot: postState.slot});
```
twoeths, line 334: do we need any metrics for this?

dapplion: - This is already covered in the state transition metric set from https://github.com/ChainSafe/lodestar/pull/5171 not sure if a metric adds value. The purpose of adding this was meant more for debugging specific instances

twoeths (approved):

wemeetagain:

PR #5145, opened 2023-02-14
Remove unnecessary I prefix interfaces

dapplion:
**Motivation**

- From https://github.com/ChainSafe/lodestar/issues/3652 remove all legacy cases of unnecessary I prefixed types that are interfaces

**Description**

Almost all changes are performed automatically with script below. Non script changes:
- Remove 2 unused interfaces
- Rename duplicate type `Logger` imported from winston

```bash
#!/bin/bash

# Removes all unnecessary uses of `interface IType`, replacing them for `type Type`
# Skips instances with there's some class that already exists with name Type
# Note: this script takes multiple minutes to execute, and modifies hundreds of files

LODESTAR_DIRECTORY="/home/lion/Code/eth2.0/lodestar"
SRC_DIRECTORY="$LODESTAR_DIRECTORY/packages"

items=$(find "$SRC_DIRECTORY" -name "*.ts" -type f -exec grep -H "interface I" {} \; \
  | sed -E 's/^.+interface[[:space:]]+([[:alnum:]]+).+$/\1/g' \
  | sort -u)

echo Found matches: $items

for item in $items
do
  item_no_prefix=${item:1}  # remove first character
  
  if grep -rl "class $item_no_prefix" --include \*.ts $SRC_DIRECTORY >/dev/null; then
    echo "File(s) containing 'class $item_no_prefix' found, skipping changes for $item"
  else
    echo "replacing $item"
    find "$SRC_DIRECTORY" -name "*.ts" -type f -exec sed -i "s/interface[[:space:]]\+$item[[:space:]]*{/type $item_no_prefix = {/g" {} \;
    find "$SRC_DIRECTORY" -name "*.ts" -type f -exec sed -i "s/$item/$item_no_prefix/g" {} \;
  fi
done

# Format with prettier to add ; to type declarations

$LODESTAR_DIRECTORY/node_modules/.bin/prettier "$SRC_DIRECTORY/**/*.ts" --write
```

The script output is 
```

[...]

PR #5174, opened 2023-02-20
Remove i prefix

nflaig:


dapplion:
@nflaig Why is the diff so big if it's only a few remaining items?

nflaig:
@dapplion no idea tbh, some weird github behavior/bug?

I force-pushed the first commit of this branch because there was a typo in the commit message `Manuel` --> `Manual` but that's it.

The changes shown here do not make sense, for example these changes is in both branches, so it shouldn't show up as a diff here

image

nflaig:
actual changes are done here https://github.com/ChainSafe/lodestar/pull/5145

nflaig:
@dapplion this must be some weird bug on the github side, after you cherry-picked the last commit there is 0 diff between the two branches, at least according to `git diff`

image

PR #5175, opened 2023-02-20
`persistNetworkIdentity` cli option

### LINKED ISSUE #4992: Add CLI option to persist peerId and ENR
> We should write the ENR on shutdown too, no? only if --peerIdFile is set.

peerId is never re-use with current code. Once that is added yes, the ENR should be persisted on shutdown

_Originally posted by @dapplion in https://github.com/ChainSafe/lodestar/issues/4989#issuecomment-1375096000_

twoeths:
**Motivation**

- To have a fixed peer id to integrate with other clients' nodes

**Description**

- Add `peerIdFile` option as global option
- Load peer id from `peerIdFile`, if it does exist it'll throw error `Error: ENOENT: no such file or directory`
- Persist enr when node is shutdown

Closes #4992

### DIFF
--- a/packages/cli/src/cmds/beacon/handler.ts
+++ b/packages/cli/src/cmds/beacon/handler.ts
@@ -10,7 +10,7 @@ import {ACTIVE_PRESET, PresetName} from "@lodestar/params";
 import {ProcessShutdownCallback} from "@lodestar/validator";
 
 import {GlobalArgs, parseBeaconNodeArgs} from "../../options/index.js";
-import {BeaconNodeOptions, exportToJSON, getBeaconConfigFromArgs} from "../../config/index.js";
+import {BeaconNodeOptions, exportToJSON, getBeaconConfigFromArgs, readPeerId} from "../../config/index.js";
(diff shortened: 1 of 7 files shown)


packages/cli/src/cmds/beacon/handler.ts
```diff
@@ -140,7 +145,7 @@ export async function beaconHandlerInit(args: BeaconArgs & GlobalArgs) {
   }
 
   // Create new PeerId everytime by default, unless peerIdFile is provided
-  const peerId = await createSecp256k1PeerId();
+  const peerId = args.peerIdFile ? await readPeerId(args.peerIdFile) : await createSecp256k1PeerId();
   const enr = SignableENR.createV4(createKeypairFromPeerId(peerId));
```
wemeetagain: I think we need something like:
```
// If peerIdFile
//   attempt to read enr from disk, beaconDir/enr
//   accept enr if enr.peerId == peerId
// else
//    create new enr
```

Otherwise, when using peerIdFile, then restarting, the discv5 network will ignore our enr updates since the enr seqno has been reset.

wemeetagain (approved):

wemeetagain:

PR #5178, opened 2023-02-20
refactor: move data availability validation out of state transition to allow optimistic sync

g11tech:
ref
 - https://github.com/ethereum/consensus-specs/pull/3185

### DIFF
--- a/packages/beacon-node/src/chain/blocks/index.ts
+++ b/packages/beacon-node/src/chain/blocks/index.ts
@@ -57,7 +57,11 @@ export async function processBlocks(
   }
 
   try {
-    const {relevantBlocks, parentSlots, parentBlock} = verifyBlocksSanityChecks(this, blocks, opts);
+    const {relevantBlocks, dataAvailabilityStatuses, parentSlots, parentBlock} = verifyBlocksSanityChecks(
+      this,
+      blocks,
+      opts
+    );
 
     // No relevant blocks, skip verifyBlocksInEpoch()
     if (relevantBlocks.length === 0 || parentBlock === null) {
@@ -71,6 +75,7 @@ export async function processBlocks(
       this,
       parentBlock,
       relevantBlocks,
+      dataAvailabilityStatuses,
       opts
     );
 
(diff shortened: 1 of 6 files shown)


dapplion (commented):
While this was changed in the spec we don't really have to change anything right? I really like the need to have to explicitly state if the availability of the blob on import as it forces critical decisions publicly. Then if a block is not available it wont ever be imported to the forkchoice so it is not a concern to it

g11tech:
> While this was changed in the spec we don't really have to change anything right? I really like the need to have to explicitly state if the availability of the blob on import as it forces critical decisions publicly. Then if a block is not available it wont ever be imported to the forkchoice so it is not a concern to it

I can accumulate and make the dataAvailable flags pass upstream to forkchoice's `onBlock` where we can check and throw on availability status to surface the considerations

i feel it doesnot belong to block state transition, also feel the same for execution status too that we pass there.

wemeetagain:
Moving to draft and adding discussion tag

packages/beacon-node/src/api/impl/beacon/state/utils.ts
```diff
@@ -254,7 +253,6 @@ async function getFinalizedState(
     state = stateTransition(state, block, {
       // Replaying finalized blocks, all data is considered valid
       executionPayloadStatus: ExecutionPayloadStatus.valid,
-      dataAvailableStatus: DataAvailableStatus.available,
```
dapplion: The main purpose of this notation was to force all consumers to explicitly declare assumptions about blobs. We will have to support different paths like regen block replay where we don't want to load blobs, but already know they are available. I'm aware it's technically pointless but being verbose about such critical decisions sounded good to me. I'm fine moving DA check outside of state transition, but this explicit statements should stay IMO

g11tech: ok sounds good, will make modifications

dapplion (dismissed):

g11tech:
rebased to resolve conflicts

dapplion (approved):

wemeetagain:

PR #5182, opened 2023-02-20
Collect system data for client monitoring

nflaig:
**Motivation**

Follow up PR for https://github.com/ChainSafe/lodestar/pull/5037 to implement collection of system data to support full feature set of beaconcha.in monitoring app.

**Description**

Replaces currently hard-coded values in system stats with proper provider functions. Most of the system data is collected using [systeminformation](https://www.npmjs.com/package/systeminformation) which is a widely used and well maintained package with [0 dependencies](https://www.npmjs.com/package/systeminformation?activeTab=dependencies). There are a few issues running in a containerized environment as some data might be inaccurate or missing but the important data is collected correctly even there. These issues might also be addressed in the future, see https://github.com/sebhildebrandt/systeminformation/issues/777, although it is really not that important.

**Performance**

Dashboards are already in place to keep track of the performance. The functions to collect system data are not executed in the main thread and a lot of the time use child processes which execute native system calls depending on OS. The performance measurements from the last few days looks really promising and I did not see any issues as of now both validator efficiency and system load seem unaffected by enabling client monitoring.

Added some screenshots of prometheus metrics, just let me now if I should provide further metrics.

<details>
<summary markdown="span">Prometheus metrics</summary>
<pre>

Takes between 0.5 - 1 second to collect system data, depending on the environment (on windows ~3 seconds).

image

It should not cause any spikes in Lodestar CPU usage as it is executed in a different process

image

Overall container CPU usage looks got to me as well, it does not seem like collecting system data consumes too much CPU

image

Event Loop Lag seems unaffected as well

image

Process memory usage looks stable

image

Container memory usage as well (no rss memory leak)

image

</pre>

</details>

**Beaconcha.in mobile app**

We support all dashboards in the beaconcha.in mobile app and also all notifications which is the most valuable feature of the app in my opinion. A node operator gets a lot of value from the app which you normally would only get if you run prometheus + grafana and invest additional time to configure proper alerting.

<details>
<summary markdown="span">Mobile app screenshots</summary>
<pre>

Screenshot_20230220_135448_Beaconchain 
Screenshot_20230220_141016_Beaconchain
Screenshot_20230218_212040_Beaconchain
Screenshot_20230218_212106_Beaconchain
Screenshot_20230219_232736_Beaconchain
Screenshot_20230218_212159_Beaconchain
Screenshot_20230218_212227_Beaconchain
Screenshot_20230218_212242_Beaconchain
Screenshot_20230218_212305_Beaconchain
Screenshot_20230218_212320_Beaconchain
Screenshot_20230218_212338_Beaconchain
Screenshot_20230218_212403_Beaconchain
Screenshot_20230218_212417_Beaconchain
</pre>

</details>

### DIFF
--- a/packages/beacon-node/src/monitoring/service.ts
+++ b/packages/beacon-node/src/monitoring/service.ts
@@ -6,6 +6,7 @@ import {HistogramExtra} from "../metrics/utils/histogram.js";
 import {defaultMonitoringOptions, MonitoringOptions} from "./options.js";
 import {createClientStats} from "./clientStats.js";
 import {ClientStats} from "./types.js";
+import {collectSystemData} from "./system.js";
 
 type MonitoringData = Record<string, string | number | boolean>;
 
@@ -146,6 +147,10 @@ export class MonitoringService {
     const timer = this.collectDataMetric.startTimer();
     const data: MonitoringData[] = [];
     const recordPromises = [];
+
+    if (this.options.collectSystemStats) {
+      await collectSystemData(this.logger);
+    }
 
     for (const [i, s] of this.clientStats.entries()) {
       data[i] = {};
--- a/packages/beacon-node/src/monitoring/system.ts
+++ b/packages/beacon-node/src/monitoring/system.ts
@@ -1,0 +1,268 @@
+import os from "node:os";
+import path from "node:path";
+import system from "systeminformation";
+import {Logger} from "@lodestar/utils";
+
+type MiscOs = "lin" | "win" | "mac" | "unk";
+
+// static data only needs to be collected once
+let staticDataCollected = false;
+// disk I/O is not measurable in some environments
+let diskIOMeasurable = true;
+
+let cpuCores = 0;
+let cpuThreads = 0;
+let cpuNodeSystemSecondsTotal = 0;
+let cpuNodeUserSecondsTotal = 0;
+let cpuNodeIdleSecondsTotal = 0;
+let memoryNodeBytesTotal = 0;
+let memoryNodeBytesFree = 0;
+let memoryNodeBytesCached = 0;
+let memoryNodeBytesBuffers = 0;
+let diskNodeBytesTotal = 0;
+let diskNodeBytesFree = 0;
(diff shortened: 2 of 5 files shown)


packages/beacon-node/src/monitoring/service.ts
```diff
@@ -147,6 +148,10 @@ export class MonitoringService {
     const data: MonitoringData[] = [];
     const recordPromises = [];
 
+    if (this.options.collectSystemStats) {
```
nflaig, line 152: **Note:** it is intentional that `collectSystemStats` uses the term stats and `collectSystemData` the term data as the options value refers to not just collecting system data but also constructing system stats as defined [here](https://github.com/ChainSafe/lodestar/blob/9ac65e1b94701c9f3856f85e7ab86bb61563951d/packages/beacon-node/src/monitoring/clientStats.ts#L197)

wemeetagain (approved):
lgtm
More review could be done on `packages/beacon-node/src/monitoring/system.ts` where the majority of the functionality lives.

packages/beacon-node/src/monitoring/system.ts
```diff
@@ -0,0 +1,268 @@
...
+export function getCpuCores(): number {
+  return cpuCores;
+}
+
+/**
+ * Number of CPU threads available
+ */
+export function getCpuThreads(): number {
+  return cpuThreads;
+}
```
nazarhussain, line 147: Overall seems good to me. I don't prefer to have `get` prefix for so many functions. Instead an object with getter properties or a class with getters would be preferred.

nflaig: I wanted to use a class initially but then I was no longer sure it makes due to the way the values should be cached. Should I define it as a class, instantiate it in this file and export the instantiated class only from the file? I could see that, should be the same result but everything wrapped into a class

nflaig: Refactored in https://github.com/ChainSafe/lodestar/pull/5202

wemeetagain:

PR #5187, opened 2023-02-21
Simplify INetwork Interface

wemeetagain:
**Motivation**

Another step towards #4856 
The `INetwork` class should be simplified in preparation for moving it to a worker.

**Description**

- remove unnecessary methods/properties
  - many properties were internal details, unused, or factored away
- replace large interfaces with small interfaces
  - create a simplified gossipsub interface, `GossipsubBeaconNode`
- most methods become async/return promises
  - things that will be easy to retain outside the worker can stay sync
- avoid reaching into deep internals in favor of dumpXXX methods

### DIFF
--- a/packages/beacon-node/src/network/gossip/interface.ts
+++ b/packages/beacon-node/src/network/gossip/interface.ts
@@ -122,6 +122,21 @@ export type GossipModules = {
   chain: IBeaconChain;
 };
 
+export type GossipBeaconNode = {
+  publishBeaconBlock(signedBlock: allForks.SignedBeaconBlock): Promise<void>;
+  publishSignedBeaconBlockAndBlobsSidecar(item: deneb.SignedBeaconBlockAndBlobsSidecar): Promise<void>;
+  publishBeaconAggregateAndProof(aggregateAndProof: phase0.SignedAggregateAndProof): Promise<number>;
+  publishBeaconAttestation(attestation: phase0.Attestation, subnet: number): Promise<number>;
+  publishVoluntaryExit(voluntaryExit: phase0.SignedVoluntaryExit): Promise<void>;
+  publishBlsToExecutionChange(blsToExecutionChange: capella.SignedBLSToExecutionChange): Promise<void>;
+  publishProposerSlashing(proposerSlashing: phase0.ProposerSlashing): Promise<void>;
+  publishAttesterSlashing(attesterSlashing: phase0.AttesterSlashing): Promise<void>;
+  publishSyncCommitteeSignature(signature: altair.SyncCommitteeMessage, subnet: number): Promise<void>;
+  publishContributionAndProof(contributionAndProof: altair.SignedContributionAndProof): Promise<void>;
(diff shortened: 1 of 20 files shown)


packages/beacon-node/src/network/gossip/interface.ts
```diff
@@ -122,6 +122,21 @@ export type GossipModules = {
   chain: IBeaconChain;
 };
 
+export type GossipBeaconNode = {
```
nflaig, line 125: shouldn't this rather be an `interface`?

packages/beacon-node/test/unit/api/impl/node/node.test.ts
```diff
@@ -52,15 +49,13 @@ describe("node api implementation", function () {
...
-      networkStub.metadata = {
-        get json(): altair.Metadata {
-          return {
-            attnets: BitArray.fromBoolArray([true]),
-            syncnets: BitArray.fromBitLen(0),
-            seqNumber: BigInt(1),
-          };
-        },
-      } as MetadataController;
+      networkStub.getMetadata.returns(
```
nflaig, line 52: might just be able to do 

```suggestion
      networkStub.getMetadata.resolves(
```

and remove the `Promise.resolve` one line below

packages/beacon-node/test/unit/api/impl/node/node.test.ts
```diff
@@ -52,15 +49,13 @@ describe("node api implementation", function () {
...
+        Promise.resolve({
+          attnets: BitArray.fromBoolArray([true]),
+          syncnets: BitArray.fromBitLen(0),
+          seqNumber: BigInt(1),
+        })
+      );
```
nflaig, line 58: might just be able to do

```suggestion
      networkStub.getMetadata.resolves(
        attnets: BitArray.fromBoolArray([true]),
        syncnets: BitArray.fromBitLen(0),
        seqNumber: BigInt(1),
      );
```

there are more cases below

nflaig (commented):
It does look a lot cleaner and I did not find any big issues but can't really approve this as I am not familiar with the code and the impact of those changes.

dapplion:
Noted we should probably have two classes, one interacting with everything else in the main thread and one for the worker - main thread comms

g11tech (approved):
lgtm

packages/beacon-node/src/network/gossip/interface.ts
wemeetagain: Definitely could be an interface. I don't know that we have clear project-wide opinions on the use of one vs the either.

nflaig: IMO if it is implemented by a class within the project directly or even outside in an external library I would use `interface` otherwise should just use `type`

wemeetagain: Yeah, `type` here is a bit sloppy. Can clean it up in a future PR.
I'm continuing to do some work to clean this up, aiming to move the network to a worker, so will continue touching these pieces of code.

wemeetagain:

PR #5183, opened 2023-02-20
Refactor client monitoring

nflaig:
**Motivation**

Some follow up changes/refactoring for #5037 to clean up few things and fix some minor issues.

**Description**

- Simplify dynamic client stats property
- Add 5 seconds bucket to collect data metric
- Add 2 seconds to monitoring interval to avoid rate limit errors
- Add comment on why only remote host should be logged
- Override colors in send data duration panel
- Add jsdoc to client stats properties
- Use `CONSTANT_CASE` for client stats schema fixtures

See [commits](https://github.com/ChainSafe/lodestar/pull/5183/commits) for further details.

PR #5189, opened 2023-02-21
Fix client monitoring rate limit errors

nflaig:
**Motivation**

Follow up change for #5037 which fixes beaconcha.in rate limit errors. 

**Description**

The current implementation might send the next request too early if the previous one took a bit longer due to the nature of how `setInterval` works, it does not wait for async functions to finish.

This change ensure we are only starting the next interval after the previous request has finished.

Reverts changes done in [this commit](https://github.com/ChainSafe/lodestar/pull/5183/commits/286a1cf3d5075e9de84b3a119be2f2bc89e35d58) as setting interval to 62 seconds did not actually fix the issue just made it less likely to happen.

We can safely set the interval to 60 seconds now as we will only start the next interval after we the get previous response which means the sent interval might vary a bit (usually just a few milliseconds) depending on the response times of the remote service.

### DIFF
--- a/packages/beacon-node/src/monitoring/service.ts
+++ b/packages/beacon-node/src/monitoring/service.ts
@@ -43,7 +43,7 @@ export class MonitoringService {
 
   private status = Status.Stopped;
   private initialDelayTimeout?: NodeJS.Timeout;
-  private monitoringInterval?: NodeJS.Timer;
+  private monitoringInterval?: NodeJS.Timeout;
   private fetchAbortController?: AbortController;
   private pendingRequest?: Promise<void>;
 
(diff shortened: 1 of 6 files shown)


packages/beacon-node/src/monitoring/service.ts
```diff
@@ -211,6 +208,16 @@ export class MonitoringService {
     }
   }
 
+  private nextMonitoringInterval(): void {
+    if (this.status === Status.Stopped) return;
```
nflaig, line 212: This check just prevents an edge case which can happen due to a race condition where `this.send()` has not yet finished but service was stopped, it would invoke `this.nextMonitoringInterval()` again and keep the interval running even though service is already stopped.

This is probably never relevant if the BN or VC run as standalone services because the whole process will be stopped either way but if it is embedded in another software this might matter.

wemeetagain (approved):

wemeetagain:

packages/beacon-node/src/monitoring/service.ts
nflaig: Just noting here that this change is also quite important if BN or VC are standalone services as setting another timeout would prevent the process from exiting
- https://github.com/ChainSafe/lodestar/pull/5330

PR #5190, opened 2023-02-21
Improve approximate db size instrumentation

nflaig:
**Motivation**

As discussed in https://github.com/ChainSafe/lodestar/pull/5087 in https://github.com/ChainSafe/lodestar/pull/5087#discussion_r1111193591 we should improve approximate db size instrumentation.

**Description**

- Add debug log to capture approximate db size errors
- Add histogram to track approximate db size time (bucket sizes are chosen based on initial measurements, see https://github.com/ChainSafe/lodestar/pull/5087#discussion_r1104650506)
- Add dashboard panel to display approximate db size duration

image

@dapplion You also mentioned that an error counter should be added but this is really not necessary in my opinion as this should never throw an error and if it does, we will now log it to debug.

### DIFF
--- a/dashboards/lodestar_vm_host.json
+++ b/dashboards/lodestar_vm_host.json
@@ -1,9 +1,49 @@
 {
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS",
+      "label": "prometheus",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
+      "pluginName": "Prometheus"
+    }
+  ],
+  "__elements": {},
+  "__requires": [
+    {
+      "type": "grafana",
+      "id": "grafana",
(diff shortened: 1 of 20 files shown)


dashboards/lodestar_vm_host.json
```diff
@@ -1,9 +1,49 @@
 {
+  "__inputs": [
```
nflaig, line 2: This was added because I used the `Export for sharing externally` option which should be correct in my opinion.
Not sure about the other additions, probably related to a newer grafana version, I made sure that the dashboards is not changed anywhere else, just added one panel for approximate db size duration.

Would make sense if someone reviewing quickly imports this as well to double check it

wemeetagain (approved):
lgtm

wemeetagain:

PR #5194, opened 2023-02-22
Add Result Err helpers for batched flow

dapplion:
**Motivation**

To implement a batched flow like in batched attestation validation https://github.com/ChainSafe/lodestar/blob/97f00b83a986aa9a4f94ae7f0aebef93005f713a/packages/beacon-node/src/chain/validation/attestation.ts#L19-L26 the following semantics are necessary

Given a function that may return an error
```ts
function squareNum(n: number): Result<number, Error>
```

That function can now be consumed as
```ts
function getNumSquare(): Result<number, Error> {
  const value = getNum();
  if (isErr(value)) {
    return value; // return as error
  }
  return value ** 2;
}
```

**Description**

Introduce Result, Err and isErr building blocks


packages/beacon-node/src/util/err.ts
```diff
@@ -0,0 +1,106 @@
+const symErr = Symbol("err");
```
wemeetagain: Can you add this to lodestar/utils? This might be a nice pattern to apply across other packages, eg api or state transition

wemeetagain (approved):

wemeetagain:

Issue #5176, opened 2023-02-20
Gossip beacon_attestation queue always full if subscribe-all-subnets


dapplion:
**Describe the bug**

Running Lodestar beacon node on a very powerful host (https://www.hetzner.com/dedicated-rootserver/ax41-nvme) results in Lodestar dropping a constant % of received attestations. Other system metrics point to the system not being at capacity.

This indicates the gossip queue strategy is not optimal and rejects messages that _could_ be processed. 

_rate of dropped messages on the beacon_attestation topic_

![Screenshot from 2023-02-20 16-41-01](https://user-images.githubusercontent.com/35266934/220057562-09d385b4-a1e1-4c06-8846-b5f6374b207d.png)

_beacon_attestation topic message queue length (always at almost max)_

![Screenshot from 2023-02-20 16-41-14](https://user-images.githubusercontent.com/35266934/220057695-fcb134fa-ff7f-41cd-a6be-275c6ed4f468.png)

The limiting factors are:
- Main thread at capacity
- BLS workers at capacity

_Most BLS worker threads are iddle_

![Screenshot from 2023-02-20 16-33-19](https://user-images.githubusercontent.com/35266934/220058046-f252d80a-8390-4a71-804b-d51140f6d1b5.png)

_The BLS worker manager queue in the main thread is never full_

![Screenshot from 2023-02-20 16-34-49](https://user-images.githubusercontent.com/35266934/220058145-d2709995-d8e2-4d8a-a411-b5537aec25f8.png)

Having the gossip queues have the downside of keeping objects for too long and processing many of them pretty late
![Screenshot from 2023-02-20 16-39-54](https://user-images.githubusercontent.com/35266934/220058425-85e084d7-d2f6-497a-b033-d2af42d60397.png)
![Screenshot from 2023-02-20 16-40-06](https://user-images.githubusercontent.com/35266934/220058441-514aa663-4fa1-46c7-bb7d-75af234cdc1d.png)

The host CPU stats do not show high usage, hinting at idle capacity available to process those attestations

![Screenshot from 2023-02-20 16-43-10](https://user-images.githubusercontent.com/35266934/220058593-e190905d-5c5e-4e4b-a983-fb714dc8b82b.png)

**Expected behavior**

Given enough resources Lodestar should be able to process all attestations

twoeths:
Implemented in above PRs

PR #5195, opened 2023-02-22
Pull gossip queues for better throughput

dapplion:
**Motivation**

Lodestar's gossip queues are too restrictive to handle big volumes of objects such when subscribed to all subnets

- See https://github.com/ChainSafe/lodestar/issues/5176

Our existing queues throttle by limiting concurrency, irrespective of downstream worker's capacity. They have the authority to push jobs, and do so independently from each other

**Description**

This PR updates the queue design to:
- Queues just hold data
- NetworkProcessor has the authority to get data from the queues and execute it
- NetworkProcessor throttles based on downstream workers availability to take more work
- Since there's now a central scheduling function it can prioritize topics over others, for example process blocks before attestations

_Marking as draft until testing shows all good_


dapplion:
Deployed to `feat3`, results show good queue efficiency but mesh health is bad.

_Queues are never full, message drop rate is 0, and job wait time is almost 0_

Screenshot from 2023-02-23 12-15-14

_But mesh peers is now at minimum and don't have peers on most subnets_

Screenshot from 2023-02-23 12-16-36

_We receive way less messages in the beacon_attestation topic_

Screenshot from 2023-02-23 12-17-51

_Peer quality is much different, we have mostly nodes with no validators connected_

Screenshot from 2023-02-23 12-20-10

**EDIT**: From Tuyen

> feat3-lg1k-hzax41 has same issue to beta-lg1k-hzax41, this node contains only low long lived subnets peer, peer count is consistently >=56 peers so there is no chance for new peer. low long lived subnets peer causes low mesh peers see https://github.com/ChainSafe/lodestar/issues/5198


- `feat3-novc-ctvpss`: Queue times for aggregates reduced from 1s to 50ms. Other metrics good.
- `feat3-sm1v-ctvpss`: Queue times for aggregates reduced from 1s to 50ms. Other metrics good.
- `feat3-md16-ctvpsm`: Queue times for aggregates + attestations reduced from 1s to 50ms. Other metrics good.
- `feat3-md64-hzax41`: Affected to issue of peers with low subnet count
- `feat3-lg1k-hzax41`: Affected to issue of peers with low subnet count

twoeths:
peers are not pruned because:
1. most/all peers have duties
2. most/all peers subscribe to >= 1 subnet
3. most/all peers have good score
4. not too grouped to a subnet

this is the same issue to https://github.com/ChainSafe/lodestar/issues/5198

 The reason it happens more to this branch maybe because gossip score it so great (point 3 above - I compared gossipsub score in feat3 and other groups)

packages/api/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/ChainSafe/lodestar/issues"
   },
-  "version": "1.4.3",
+  "version": "1.5.0",
```
twoeths: I wonder why it shows the diff here while `unstable` really have version as `1.5.0`, should not be an issue anyway

twoeths:
This is awesome work, I can see good metrics like "Job Wait Time" or "Dropped jobs %". However I also notice this push some pressure to I/O lag issue on #4002:

- A little increasement on Event Loop Lag
- A little increasement on Lodestar CPU time (160% -> 170%)
- Especially the `submitPoolAttestations` at validator side (500ms vs 100vs as in stable 1k validator node)
Screen Shot 2023-03-01 at 10 25 47

twoeths:
@dapplion is it possible to introduce some flags to mitigate the issue?

packages/api/package.json
dapplion: Rebased on unstable and fixed the bad diff

dapplion:
> @dapplion is it possible to introduce some flags to mitigate the issue?

If any this PR is allowing nodes to process all network objects they are required to process as per the spec. "The issue" is actually correct behavior, where current stable drops too many messages. To mitigate "the issue" we should force the node into processing less messages than they can, sort of re-introducing a forced limiting throughput. I'm softly against introducing data, and instead spend our energy understanding what's taking CPU time.

If we were to add flags we could limit the max concurrency per topic.

twoeths:
[0301_gossip-queue-pull_1k.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/10857742/0301_gossip-queue-pull_1k.cpuprofile.zip)

I'm attaching the profile, there is no special thing except for `handleReceivedMessages()` (then `executeWork`) was called many times inside `runMicroTasks`

In the current implementation we yield to the macro queue every 50ms through `yieldEveryMs` inside `JobItemQueue`, do we have equivalent mechanism in this implementation @dapplion ?

dapplion:
> In the current implementation we yield to the macro queue every 50ms through `yieldEveryMs` inside `JobItemQueue`, do we have equivalent mechanism in this implementation @dapplion ?

That _should_ not be necessary in theory, I added metrics and each gossip packet tends to trigger work on itself. From looking at the profile I don't see issues related with this code specifically. Since everything is run sync now it makes sense to show up on the call stack of handleReceivedMessages and executeWork but their self time is small

packages/beacon-node/src/network/processor/worker.ts
```diff
@@ -0,0 +1,52 @@
...
+export type NetworkWorkerModules = ValidatorFnsModules &
+  ValidatorFnModules & {
+    chain: IBeaconChain;
+    events: NetworkEventBus;
+    metrics: Metrics | null;
+    // Optionally pass custom GossipHandlers, for testing
+    gossipHandlers?: GossipHandlers;
+  };
+
+export class NetworkWorker {
```
wemeetagain, line 18: might be worth adding a tsdoc comment here describing the purpose or mentioning that this isn't related to thread workers

packages/beacon-node/src/network/processor/index.ts
```diff
@@ -0,0 +1,161 @@
...
+    events: NetworkEventBus;
+    logger: Logger;
+    metrics: Metrics | null;
+  };
+
+export type NetworkProcessorOpts = GossipHandlerOpts & {
+  maxGossipTopicConcurrency?: number;
+};
+
+const executeGossipWorkOrderObj: Record<GossipType, true> = {
```
wemeetagain, line 23: might be worth comment here mentioning the order here defines the order that gossip topics will be processed

packages/beacon-node/src/network/processor/index.ts
```diff
@@ -0,0 +1,161 @@
...
+export type NetworkProcessorModules = NetworkWorkerModules &
+  ValidatorFnsModules & {
+    chain: IBeaconChain;
+    events: NetworkEventBus;
+    logger: Logger;
+    metrics: Metrics | null;
+  };
+
+export type NetworkProcessorOpts = GossipHandlerOpts & {
+  maxGossipTopicConcurrency?: number;
```
wemeetagain, line 20: nice to have a tsdoc here

wemeetagain (approved):
Looks really good. Can you follow up w a PR to add all new metrics to dashboard?
It seems we'll have to tune things based on what we see there.

packages/beacon-node/src/network/processor/worker.ts
dapplion: Yeah naming sucks, I think this specific class should be merged into the NetworkProcessor

wemeetagain:

wemeetagain:

wemeetagain:

PR #5197, opened 2023-02-22
Use more deterministic keystore cache path 

### LINKED ISSUE #5141: `validator_keys.cache` is stored in the same directory as my `validator_keys`
@nazarhussain I noticed that the `validator_keys.cache` is stored in the same directory as my `validator_keys` folder which I set via `--importKeystores`. This is a not ideal in a containerized environment as the cache file is likely not mounted to the container or persistent in a volume. Maybe it would be better to store the cache file in the data directory (`dataDir`).

In our `docker-compose.validator.yml` file we also only mount the `/keystores` which would not include the cache file.

https://github.com/ChainSafe/lodestar/blob/2042c3b1c1916ceed6bc825057515a4456c33a08/docker-compose.validator.yml#L4-L11

_Originally posted by @nflaig in https://github.com/ChainSafe/lodestar/pull/5043#discussion_r1103874698_

nazarhussain:
**Motivation**

Use more deterministic keystore cache. 

**Description**

Create a new cache directory and use the keystore cache stored in there. 

Closes #5141

Ru all tests.

### DIFF
--- a/packages/cli/src/cmds/validator/signers/index.ts
+++ b/packages/cli/src/cmds/validator/signers/index.ts
@@ -1,3 +1,4 @@
+import {join} from "node:path";
 import bls from "@chainsafe/bls";
 import {deriveEth2ValidatorKeys, deriveKeyFromMnemonic} from "@chainsafe/bls-keygen";
 import {interopSecretKey} from "@lodestar/state-transition";
@@ -94,7 +95,7 @@ export async function getSignersFromArgs(
     return decryptKeystoreDefinitions(keystoreDefinitions, {
       ...args,
       onDecrypt: needle,
-      cacheFilePath: `${args.importKeystores[0]}.cache`,
+      cacheFilePath: join(args.importKeystores[0], "keystore.cache"),
     });
   }
 
@@ -126,7 +127,7 @@ export async function getSignersFromArgs(
     const keystoreSigners = await decryptKeystoreDefinitions(keystoreDefinitions, {
       ...args,
       onDecrypt: needle,
-      cacheFilePath: `${accountPaths.keystoresDir}.cache`,
+      cacheFilePath: join(accountPaths.cacheDir, "keystores.cache"),
     });
 
     // Read local remote keys, imported via keymanager api
--- a/packages/cli/src/cmds/validator/signers/index.ts
+++ b/packages/cli/src/cmds/validator/signers/index.ts
@@ -45,6 +45,8 @@ export async function getSignersFromArgs(
   network: string,
   {logger, signal}: {logger: Pick<ILogger, "info">; signal: AbortSignal}
 ): Promise<Signer[]> {
+  const accountPaths = getAccountPaths(args, network);
+
   // ONLY USE FOR TESTNETS - Derive interop keys
   if (args.interopIndexes) {
     const indexes = parseRange(args.interopIndexes);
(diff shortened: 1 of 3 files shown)


packages/cli/src/cmds/validator/signers/index.ts
```diff
@@ -94,7 +97,7 @@ export async function getSignersFromArgs(
     return decryptKeystoreDefinitions(keystoreDefinitions, {
       ...args,
       onDecrypt: needle,
-      cacheFilePath: `${args.importKeystores[0]}.cache`,
+      cacheFilePath: join(accountPaths.cacheDir, "imported_keystores.cache"),
```
nflaig, line 100: highly opinionated (feel free to ignore) but I really don't like just having `join` here, someone has to hover the function to know it from `path` library or scroll up to the imports to see where this is coming from

since path uses such general names for it's functions I generally like to do `path.`, to make it more obvious we are joining paths here

```suggestion
      cacheFilePath: path.join(accountPaths.cacheDir, "imported_keystores.cache"),
```

nflaig (approved):
LGTM! This should address the issue noted in https://github.com/ChainSafe/lodestar/pull/5043#discussion_r1103874698

packages/cli/src/cmds/validator/signers/index.ts
nazarhussain: We use destructuring convention all over the code, so seems fine to me. In my preference, I always tend to relate a function to standard library function. If I see a function `join` I will consider it `path.join`. If we define a user function which joins string, I would declare it as `joinStr` to avoid confusion with standard library.

nflaig: > We use destructuring convention all over the code, so seems fine to me.

I would disagree  on when an import should be  destructured or not. I do think this should not be based on convention in the code but rather on the design of the library, meaning how did the author/s want you to use it.

In the case of most of the native nodejs libraries those are designed to not use destructuring, one indicator of this are the ambiguous names of functions like for example `path` has things like [delimiter](https://nodejs.org/api/path.html#pathdelimiter), [normalize(path)](https://nodejs.org/api/path.html#pathnormalizepath) or [parse(path)](https://nodejs.org/api/path.html#pathparsepath). If we just take `parse` as an example, this could mean anything, parse a URL, a encoded value, something else? If the authors would have written the library to be more destructuring friendly they would have called this function `parsePath`.

And if you just look at the docs, for any nodejs library it becomes also clear how they want the libraries to be used, e.g. [path](https://nodejs.org/api/path.html), [crypto](https://nodejs.org/api/crypto.html), [os](https://nodejs.org/api/os.html)

> If I see a function join I will consider it path.join.

That might be fine for you as you might know all of the nodejs APIs but for a more unexperienced developer it might not. Functions should clearly convey their purpose without having to think or consider anything and if it requires the code to be more verbose to achieve this it is fine to me.

> so seems fine to me.

it definitely is fine in this case, especially since the property right before is called `cacheFilePath` and the variable is called `accountPaths.cacheDir` is should be clear that we are joining paths here

This is just more general thing how I think native libraries should be used due to their ambiguous names.

dapplion: I really like to see `path.join` over just `join` too

nazarhussain: As I mentioned in some earlier discussion, if any style/pattern can't be enforced by linting then we should keep it flexible. 

I am creating a separate issue to enforce not to use the destructuring for the some module imports e.g. `path`. Once we have that rule, the same style will be enforced everywhere.

nazarhussain: Here is the issue https://github.com/ChainSafe/lodestar/issues/5201

wemeetagain:

PR #5200, opened 2023-02-23
Add dashboard variables to customize beacon/validator job name

nflaig:
**Motivation**

Ongoing effort to make Lodestar dashboards more user friendly and easier to consume. We do not want users to change the job name in their prometheus config file as this would break historical data. The ideal case is that users do not have to touch their existing prometheus config, neither should they have to add custom labels or be forced to use a specific `job_name`.

As mentioned by metanull.eth in [discord](https://discord.com/channels/593655374469660673/593655641445367808/1076151935746461806) 

> Adding variables to Lodestar dashboards would help because it just gives one more point of configurability between two different dashboards. If you are starting with a fresh installation of everything, historical data isn't going to matter, and the variables can be easily changed to get dashboards compatible without losing access to historical data.

**Description**

- use variables for to match job names instead of hard coding values
- when importing the dashboard, user can configure the job names they want, defaults will be `beacon` and `validator` as before
- a lot of the diffs in dashboards are due to newer grafana version

image

### DIFF
--- a/dashboards/lodestar_block_processor.json
+++ b/dashboards/lodestar_block_processor.json
@@ -3046,7 +3046,7 @@
               {
                 "matcher": {
                   "id": "byName",
-                  "options": "beacon_fork_choice_errors_total{group=\"beta\", instance=\"contabo-13\", job=\"$beacon_job\"}"
+                  "options": "beacon_fork_choice_errors_total{group=\"beta\", instance=\"contabo-13\", job=~\"$beacon_job|beacon\"}"
                 },
                 "properties": [
                   {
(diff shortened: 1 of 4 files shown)


dapplion (changes requested):
Please can you split this PR into a PR first:
- Update Grafana version and related changes, target unstable
- Add Job names variables, target first PR's branch

nflaig:
@dapplion split the PR into two (https://github.com/ChainSafe/lodestar/pull/5210 and https://github.com/ChainSafe/lodestar/pull/5211), the changes to `dashboards/lodestar_summary.json`, `dashboards/lodestar_validator_client.json` and `dashboards/lodestar_vm_host.json` are exactly the same, just separated in two PRs. Regarding `dashboards/lodestar_block_processor.json`, I dropped most of the grafana related changes as there was a massiv update to the dashboard in [this commit](https://github.com/ChainSafe/lodestar/commit/c8d244ff872ced4bf2fbf75a4140f87e5ff8cda4) which uses an even newer grafana version.

PR #5202, opened 2023-02-24
Use singleton class to manage system data

nflaig:
**Motivation**

https://github.com/ChainSafe/lodestar/pull/5182#discussion_r1114932094

**Description**

Use singleton class to manage system data

### DIFF
--- a/packages/beacon-node/src/monitoring/properties.ts
+++ b/packages/beacon-node/src/monitoring/properties.ts
@@ -1,24 +1,24 @@
 import {Registry} from "prom-client";
 import {JsonRecord, JsonType, MetricObject, MetricValue, MetricWithGetter, RecordValue} from "./types.js";
 
-interface PropertyDefinition {
+type PropertyDefinition = {
   /** Key of value to be sent to remote service */
   jsonKey: string;
   /** Description of the property */
   description?: string;
-}
+};
 
-interface StaticPropertyDefinition<T extends RecordValue> extends PropertyDefinition {
+type StaticPropertyDefinition<T extends RecordValue> = PropertyDefinition & {
   /** Static value */
   value: T;
-}
+};
 
-interface DynamicPropertyDefinition<T extends RecordValue> extends PropertyDefinition {
+type DynamicPropertyDefinition<T extends RecordValue> = PropertyDefinition & {
   /** Value provider function */
   provider: () => T;
-}
+};
 
-interface MetricPropertyDefinition<T extends RecordValue> extends PropertyDefinition {
+type MetricPropertyDefinition<T extends RecordValue> = PropertyDefinition & {
   /** Type of value to be sent to remote service */
   jsonType: JsonType;
   /** Name of the metric */
@@ -37,7 +37,7 @@ interface MetricPropertyDefinition<T extends RecordValue> extends PropertyDefini
   cacheResult?: boolean;
   /** Default value if metric does not exist */
   defaultValue: T;
-}
+};
 
 /**
  * Interface to be implemented by client stats properties
--- a/packages/beacon-node/src/monitoring/system.ts
+++ b/packages/beacon-node/src/monitoring/system.ts
@@ -9,6 +9,7 @@ type MiscOs = "lin" | "win" | "mac" | "unk";
  * Singleton class to collect and provide system information
  */
 class System {
+  private static instance?: System;
   // static data only needs to be collected once
   private staticDataCollected = false;
   // disk I/O is not measurable in some environments
@@ -31,6 +32,11 @@ class System {
   private _networkNodeBytesTotalTransmit = 0;
   private _miscNodeBootTsSeconds = 0;
   private _miscOs: MiscOs = "unk";
+
+  constructor() {
+    if (System.instance) return System.instance;
+    System.instance = this;
+  }
 
   /**
    * Collect system data and update cached values
--- a/packages/beacon-node/src/monitoring/system.ts
+++ b/packages/beacon-node/src/monitoring/system.ts
@@ -5,264 +5,214 @@ import {Logger} from "@lodestar/utils";
 
 type MiscOs = "lin" | "win" | "mac" | "unk";
 
-// static data only needs to be collected once
-let staticDataCollected = false;
-// disk I/O is not measurable in some environments
-let diskIOMeasurable = true;
-
-let cpuCores = 0;
-let cpuThreads = 0;
-let cpuNodeSystemSecondsTotal = 0;
-let cpuNodeUserSecondsTotal = 0;
-let cpuNodeIdleSecondsTotal = 0;
-let memoryNodeBytesTotal = 0;
-let memoryNodeBytesFree = 0;
-let memoryNodeBytesCached = 0;
-let memoryNodeBytesBuffers = 0;
-let diskNodeBytesTotal = 0;
-let diskNodeBytesFree = 0;
-let diskNodeReadsTotal = 0;
-let diskNodeWritesTotal = 0;
-let networkNodeBytesTotalReceive = 0;
-let networkNodeBytesTotalTransmit = 0;
-let miscNodeBootTsSeconds = 0;
-let miscOs: MiscOs = "unk";
-
 /**
- * Collect system data and update cached values
+ * Singleton class to collect and provide system information
  */
-export async function collectSystemData(logger: Logger): Promise<void> {
-  const debug = (dataType: string, e: Error): void => logger.debug(`Failed to collect ${dataType} data`, {}, e);
+class System {
+  // static data only needs to be collected once
+  private staticDataCollected = false;
+  // disk I/O is not measurable in some environments
+  private diskIOMeasurable = true;
 
-  await Promise.all([
-    collectStaticData().catch((e) => debug("static system", e)),
-    collectCpuData().catch((e) => debug("CPU", e)),
-    collectMemoryData().catch((e) => debug("memory", e)),
-    collectDiskData().catch((e) => debug("disk", e)),
-    collectNetworkData().catch((e) => debug("network", e)),
-  ]);
+  private _cpuCores = 0;
+  private _cpuThreads = 0;
+  private _cpuNodeSystemSecondsTotal = 0;
+  private _cpuNodeUserSecondsTotal = 0;
+  private _cpuNodeIdleSecondsTotal = 0;
+  private _memoryNodeBytesTotal = 0;
+  private _memoryNodeBytesFree = 0;
+  private _memoryNodeBytesCached = 0;
+  private _memoryNodeBytesBuffers = 0;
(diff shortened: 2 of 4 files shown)


packages/beacon-node/src/monitoring/properties.ts
```diff
@@ -1,24 +1,24 @@
 import {Registry} from "prom-client";
 import {JsonRecord, JsonType, MetricObject, MetricValue, MetricWithGetter, RecordValue} from "./types.js";
 
-interface PropertyDefinition {
+type PropertyDefinition = {
```
nflaig, line 4: This change is not really related to the PR, I want to sneak it in tho :)

see https://github.com/ChainSafe/lodestar/pull/5187#discussion_r1116167127 on why this is changed

packages/beacon-node/src/monitoring/system.ts
```diff
@@ -5,264 +5,211 @@ import {Logger} from "@lodestar/utils";
...
-
-  miscNodeBootTsSeconds = getSystemBootTime();
-}
+class System {
+  // static data only needs to be collected once
+  private staticDataCollected = false;
+  // disk I/O is not measurable in some environments
+  private diskIOMeasurable = true;
+
+  private _cpuCores = 0;
```
nflaig: I generally would not prefix private properties with `_` but needs to be done in that case as we also have a `get` accessor for each property which uses the same name but without `_`.

I saw this convention used in other places as well so I assume it should be fine to use here

https://github.com/ChainSafe/lodestar/blob/b64a67c71513e5d8b0ff2b8479ee90ed84da94d4/packages/beacon-node/src/chain/clock/LocalClock.ts#L20-L21

packages/beacon-node/src/monitoring/system.ts
```diff
@@ -5,264 +5,211 @@ import {Logger} from "@lodestar/utils";
...
-  const cpuTimes: Record<string, number> = {};
+  private async collectCpuData(): Promise<void> {
+    const cpuTimes: Record<string, number> = {};
 
-  os.cpus().forEach((cpu) => {
-    // sum up CPU times per mode and convert to seconds
-    for (const [mode, time] of Object.entries(cpu.times)) {
-      if (cpuTimes[mode] == null) cpuTimes[mode] = 0;
-      cpuTimes[mode] += Math.floor(time / 1000);
+    for (const cpu of os.cpus()) {
```
nflaig, line 73: changed this from `os.cpus().forEach((cpu) => {})` to `for (const cpu of os.cpus())`, see https://github.com/ChainSafe/lodestar/pull/5161#discussion_r1116132778

packages/beacon-node/src/monitoring/system.ts
```diff
@@ -5,264 +5,211 @@ import {Logger} from "@lodestar/utils";
...
 
-function getSystemBootTime(): number {
-  return Math.floor(Date.now() / 1000 - os.uptime());
-}
+  private getSystemBootTime(): number {
+    return Math.floor(Date.now() / 1000 - os.uptime());
+  }
 
-/**
- * Number of CPU cores available
```
nflaig, line 136: I removed all of the function docs as those are already documented in `clientStats.ts` and I think it is better not to duplicate this, also not sure how much value the docs provide here so I went with reducing duplication and verbosity

https://github.com/ChainSafe/lodestar/blob/b64a67c71513e5d8b0ff2b8479ee90ed84da94d4/packages/beacon-node/src/monitoring/clientStats.ts#L201-L205

packages/beacon-node/src/monitoring/system.ts
```diff
@@ -5,264 +5,211 @@ import {Logger} from "@lodestar/utils";
...
- * Operating system, can be one of: lin, win, mac, unk for unknown
- */
-export function getMiscOs(): MiscOs {
-  return miscOs;
+  get miscOs(): MiscOs {
+    return this._miscOs;
+  }
 }
+
+export default new System();
```
nflaig, line 144: IMO the best singleton pattern is to just initialize the class and export is as default, this means that the class will always be initialized but I don't see a problem with that. The important thing is that it ensure that only ever a single instance can exist

packages/beacon-node/src/monitoring/system.ts
```diff
@@ -5,264 +5,211 @@ import {Logger} from "@lodestar/utils";
...
-    collectStaticData().catch((e) => debug("static system", e)),
-    collectCpuData().catch((e) => debug("CPU", e)),
-    collectMemoryData().catch((e) => debug("memory", e)),
-    collectDiskData().catch((e) => debug("disk", e)),
-    collectNetworkData().catch((e) => debug("network", e)),
-  ]);
-
-  miscNodeBootTsSeconds = getSystemBootTime();
-}
+class System {
```
nazarhussain, line 11: As this class is designed in a way to only be used as a singleton, so try adding a constructor and throw an error if more instances are created.

nazarhussain (changes requested):

packages/beacon-node/src/monitoring/system.ts
nflaig: The class it not exported and therefore can not be initialized somewhere else.

I know there is a singleton pattern were you have a static method called `getInstance` which handles the singleton state and will only ever initialize the class once and then in the `constructor` you throw an error. However, this pattern does not make sense here in my opinion since the singleton is ensured by the default export itself, see https://github.com/ChainSafe/lodestar/pull/5202#discussion_r1116821503.

packages/beacon-node/src/monitoring/system.ts
wemeetagain: why do we need private properties and getters? Rather than public properties (or public readonly properties).

nflaig:
rebased against `unstable` to get https://github.com/ChainSafe/lodestar/commit/fc33339a218a467ffb39ed242a249c2b08a9d6d8 in which fixes network e2e tests

packages/beacon-node/src/monitoring/system.ts
nflaig: I would kinda be fine with public readonly properties as well but I cannot make them readonly as they are modified within the class itself.

https://github.com/ChainSafe/lodestar/blob/0cd2d205744557d9ab67e8f1d631c2785c0efe7d/packages/beacon-node/src/monitoring/system.ts#L46

But even if I can make them readonly I still think it is not ideal in most cases, I would rather prefer to have explicit getters and setters to retrieve/modify properties of a class and make them private by default at the cost of a bit more code/verbosity.

Also the code below should never be allowed, that is why I can only use private properties + getters

image

packages/beacon-node/src/monitoring/system.ts
nflaig: @nazarhussain I am still trying to wrap my head around what pattern is best to use here, but I still think that what I am currently doing is correct for reasons as explained in the [comment above](https://github.com/ChainSafe/lodestar/pull/5202#discussion_r1118639955).


Just to get a better picture of what you are suggesting, would the diff look something like this?

```diff
diff --git a/packages/beacon-node/src/monitoring/system.ts b/packages/beacon-node/src/monitoring/system.ts
index a13ba28db6..f69adc8c94 100644
--- a/packages/beacon-node/src/monitoring/system.ts
+++ b/packages/beacon-node/src/monitoring/system.ts
@@ -29,6 +29,19 @@ class System {
   private _miscNodeBootTsSeconds = 0;
   private _miscOs: MiscOs = "unk";
 
+  private static instance?: System;
+
+  static getInstance(): System {
+    return System.instance ?? (System.instance = new System());
+  }
+
+  private constructor() {
+    if (System.instance) {
+      throw new Error("Use System.getInstance instead of new");
+    }
+    System.instance = this;
+  }
+
   /**
    * Collect system data and update cached values
    */
@@ -212,4 +225,4 @@ class System {
   }
 }
 
-export default new System();
+export default System.getInstance();

```

nazarhussain: The above structure looks good to me. You can optimize it further with `static getter` and a `returning construtor`

```ts
let num = 0; 

class System {
    private static _instance: System; 
    public prop = 120; 

    constructor() {
        if(System._instance) return System._instance; 
        System._instance = this; 
        this.prop = num++; 
    }

    static get instance(): System {
        if(System._instance) return this._instance;

        this._instance = new System();
        return this._instance;
    }
}

const sys1 = System.instance; 
const sys2 = new System(); 
console.log(sys1.prop, sys2.prop);

```

This way no matter if someone call `System.instance` or `new System()` there will always be **ONLY ONE** instance.

packages/beacon-node/src/monitoring/system.ts
nazarhussain: The `readonly` only works for the TS environment. Once compiled we didn't had that function anymore. This code is not for external use so its fine to use readonly.

nazarhussain (dismissed):

packages/beacon-node/src/monitoring/system.ts
wemeetagain: Malicious user could still corrupt the values like `(system as any)._cpuCores = 0`.
Or recreate new instances like `new Object.getPrototypeOf(system).constructor()`.
The general assumption we're currently using in the project: internal lodestar code is running in a trusted environment. You don't need to have real data hiding at runtime, you don't need to build protections against malicious misuse. No need to protect against "What if someone sets this module's private counter to zero?"

(If we do end up supporting running lodestar in an untrusted environment, we would wrap the entirety and expose only a controlled api that doesn't allow access to internals)

So here, I would use some judgement about how we expect this module to be used, as an internal datasource for the monitoring service. My feeling is that this PR should really be focusing on matters of maintainability of the code. How easy / hard am I making our lives when we inevitably come back in 6 months or a year and need to add / delete / etc system data to expose to the monitoring api?

nflaig: @wemeetagain It is not really about a malicious user but just keeping the external API as simple/limited as possible but you are probably right, since the use case of this is currently limited to as an internal datasource for the monitoring service, we should be fine with public modifiable properties as well.

If we go with that approach I think changes done in [this commit](https://github.com/ChainSafe/lodestar/pull/5202/commits/45c8cfd5aea26ca7123a1157843a79eaf5f525a6) also do not make sense, as the class is not exported and can therefore not be initialized twice.

nflaig:
@nazarhussain I removed the custom constructor to ensure a singleton after doing some further refactoring. The reasoning is still the same as explained in https://github.com/ChainSafe/lodestar/pull/5202#discussion_r1118639955.

Thanks for the suggestions and feedback on how this could be implemented with just 3 lines of code, I think this will be really valuable once we need a singleton class which is also exported and potentially even used externally.

wemeetagain (approved):
lgtm :+1:

wemeetagain:

PR #5203, opened 2023-02-24
Restrict import destructuring for some modules

### LINKED ISSUE #5201: Not use the import destructuring for some modules e.g. `path`

It's not a problem and a styling issue and consistent use of it. During the discussion it's raised by @nflaig  and @dapplion  to not use the import destructuring for the modules like `path`. So if we have to use `join` or `resolve` we should always use `path.join` and `path.resolve`. 

It's not a functionality problem, but a styling and it must be enforced by linting rule to have consistent usage. 

Create a new eslint rule to enforce not use the destructuring for the given list of packages. 

Tried to find existing eslint plugins but there is none with such behavior that came up in my search. 

**Additional context**

> highly opinionated (feel free to ignore) but I really don't like just having `join` here, someone has to hover the function to know it from `path` library or scroll up to the imports to see where this is coming from
>
> since path uses such general names for it's functions I generally like to do `path.`, to make it more obvious we are joining paths here
>
> ```suggestion
>     cacheFilePath: path.join(accountPaths.cacheDir, "imported_keystores.cache"),
> ```

_Originally posted by @nflaig in https://github.com/ChainSafe/lodestar/pull/5197#discussion_r1114801898_

nflaig:
**Motivation**

Closes https://github.com/ChainSafe/lodestar/issues/5201

**Description**

Adds eslint rule to restrict import destructuring for some modules (`"node:fs", "node:os", "node:path"`). Further modules could be added but those are the ones where I saw the highest inconsistency on how the imports are done and also these use really ambiguous function names.

Also refactored the eslint rule for restricted node module imports.

cc @nazarhussain

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -146,16 +146,18 @@ module.exports = {
       {
         patterns: ["../lib/*", "@chainsafe/*/lib/*"],
         paths: [
-          {name: "child_process", message: "Please use node:child_process instead."},
-          {name: "crypto", message: "Please use node:crypto instead."},
-          {name: "fs", message: "Please use node:fs instead."},
-          {name: "http", message: "Please use node:http instead."},
-          {name: "net", message: "Please use node:net instead."},
-          {name: "os", message: "Please use node:os instead."},
-          {name: "path", message: "Please use node:path instead."},
-          {name: "stream", message: "Please use node:stream instead."},
-          {name: "util", message: "Please use node:util instead."},
-          {name: "url", message: "Please use node:url instead."},
+          ...restrictNodeModuleImports(
+            "child_process",
+            "crypto",
+            "fs",
+            "http",
+            "net",
+            "os",
+            "path",
+            "stream",
+            "util",
+            "url"
+          ),
         ],
       },
     ],
@@ -242,6 +244,10 @@ module.exports = {
     },
   ],
 };
+
+function restrictNodeModuleImports(...modules) {
+  return modules.map((module) => ({name: module, message: `Please use 'node:${module}' instead.`}));
+}
 
 function restrictImportDestructuring(...modules) {
   return modules.map((module) => ({
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -159,6 +159,7 @@ module.exports = {
         ],
       },
     ],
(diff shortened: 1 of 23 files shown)


packages/cli/src/util/file.ts
```diff
@@ -1,4 +1,4 @@
-import fs, {WriteFileOptions} from "node:fs";
```
nflaig, line 1: I do not like it if we have to do `fs.WriteFileOptions` for the types but I guess this is a trade-off we will have to accept here.

Other solution would be be more specific and use the solution suggested here https://github.com/ChainSafe/lodestar/issues/5201#issuecomment-1442394412

.eslintrc.js
```diff
@@ -241,3 +244,14 @@ module.exports = {
...
 };
+
+function restrictNodeModuleImports(...modules) {
+  return modules.map((module) => ({name: module, message: `Please use 'node:${module}' instead.`}));
+}
+
+function restrictImportDestructuring(...modules) {
+  return modules.map((module) => ({
+    selector: `ImportDeclaration[source.value='${module}'] ImportSpecifier`,
+    message: `Importing from '${module}' using destructuring is restricted.`,
```
nflaig, line 255: not sure about the message here, was using `is not allowed` instead of `is restricted` but that sounded a bit harsh :)

dapplion (dismissed):
LGTM

.eslintrc.js
nazarhussain: Seems fine, as it's just the message for our team.

nazarhussain (dismissed):

packages/cli/src/util/file.ts
nazarhussain: That seems fine. We already using namespaced types on few other places.

.eslintrc.js
nflaig: yeah but also other contributors, I am not a native english speaker but I think as it is currently should be accurate and also formally correct.

nflaig:
rebased against `unstable` to get https://github.com/ChainSafe/lodestar/commit/fc33339a218a467ffb39ed242a249c2b08a9d6d8 in which fixes network e2e tests

dapplion (approved):

wemeetagain:

PR #5056, opened 2023-01-25
Update source installation page

nflaig:
**Motivation**

Should highlight that `git clone` will check out the `unstable` branch by default which is not recommended to use in production, see [discord discussion](https://discord.com/channels/593655374469660673/743858262864167062/1067833393393442846). 

**Description**

Add warning about branch to source installation page. Also added sections to better separate the different steps.

PR #5207, opened 2023-02-27
Update source.md - Default to stable branch

coincashew:
**Motivation**

**Safer**: To make the default git cloning behavior for "build from source" production safe, thus avoiding any misunderstandings.
**Common default branch**: All other consensus clients github repos default to their respective stable branch.

**Description**

Adds stable branch flag to git clone operation.


twoeths (dismissed):

nflaig:
@coincashew Thanks for updating this but we can not merge changes directly to `stable` branch as those need to be included to `unstable` first and then will be merged from there.

We actually did a update to the source installation docs a while ago but for it to update the actual docs page it needs to be in `stable` but it currently just is in `unstable`, see PR https://github.com/ChainSafe/lodestar/pull/5056.

I opted for adding a note instead of directly specifying which branch should be cloned as I think if you run in prod it might even be best to specify a tag aka version instead of a branch.

dapplion (dismissed):

dapplion:
@nflaig users don't read notes they just copy commands (me included!) I think adding the stable branch as a safe default is good and additive to your note

nflaig:
just need to adapt the note then as it currently says that `unstable` will be pulled by default, we have a bit of a conflict here since the note was added on `unstable` branch whereas this PR targets the `stable` branch

nflaig:
> Common default branch: All other consensus clients github repos default to their respective stable branch.

@coincashew looks like [prysm](https://github.com/prysmaticlabs/prysm) does not do this either, they default to `develop`, [in their docs](https://docs.prylabs.network/docs/install/install-with-script#step-2-install-prysm) they reference the `master` branch though. In your guide it points to `master` as well so should be fine

nflaig (changes requested):
@coincashew This is definitely a good update to the docs and I think as @dapplion mentioned we should just add the `-b stable` by default. Need to reopen this against the `unstable` branch and do the docs updates there to avoid merge conflicts when merging `unstable` to `stable`.

wemeetagain (dismissed):

nflaig (changes requested):
we also need to update the warning or just remove it entirely

https://github.com/ChainSafe/lodestar/blob/e98d8e9fadb9c3d2660fd583a56d5d44c3ca8eb9/docs/install/source.md?plain=1#L30-L32

coincashew:
Requested change made. Good feedback all around. Thanks.

nflaig (approved):
lgtm, thanks for the update @coincashew

wemeetagain:

PR #5217, opened 2023-02-28
prioritizePeers: ensure to prune to target peers

### LINKED ISSUE #5198: Ensure to prune to target peers
**Describe the bug**

In rare case where all peers subscribe to at least 1 subnet, but they are not grouped to 1 subnet, and all performs well, lodestar doesn't prune peers to target peers. When that happens, we see peers stay >= 55 peers like below:

Screen Shot 2023-02-23 at 10 05 58

this does not allow new connections (implemented by libp2p). in worse case scenario, we don't have enough subnet peers (from same node like the above)

Screen Shot 2023-02-23 at 10 09 29

and it causes a lot of missed attestations (since this node has 1k validators on goerli and it's supposed to subscribe to all subnets)

**Expected behavior**

Make sure we always prune to target peers (50) to allow more better peers to come, like below:

Screen Shot 2023-02-23 at 10 11 41

twoeths:
**Motivation**

- In rare cases we'll have all low long lived subnets peers, they all have good score and don't group to a subnet, in that case lodestar does not prune any peers
- This causes libp2p not to accept any new connections and we're not able to improve peers' long lived subnets
- This causes missed attestations with "InsufficientPeers" error, see https://github.com/ChainSafe/lodestar/issues/5198#issuecomment-1445879443

**Description**

- Ensure to always prune to target peers
  -  change `peerHasDuty` to `dutiesByPeer`: count dutied subnets of peer (modified)
  - peersEligibleForPruning: sort peers based on dutied subnets then long lived subnets then score (modified)
  - prune no long lived subnets (no change)
  - prune low score peers (no change)
  - prune peers that are too grouped to a subnet (no change)
  - prune more peers if there are still more peers to delete (new)
    - based on the sorted peers above
    - don't care `peersEligibleForPruning` since we want to ensure having exactly `targetPeers`

Closes #5198

**TODO**

- [x] test this once this is merged #5195

### DIFF
--- a/packages/beacon-node/src/network/peers/utils/prioritizePeers.ts
+++ b/packages/beacon-node/src/network/peers/utils/prioritizePeers.ts
@@ -62,6 +62,7 @@ export enum ExcessPeerDisconnectReason {
   LOW_SCORE = "low_score",
   NO_LONG_LIVED_SUBNET = "no_long_lived_subnet",
   TOO_GROUPED_SUBNET = "too_grouped_subnet",
+  FIND_BETTER_PEERS = "find_better_peers",
 }
 
 /**
@@ -106,7 +107,7 @@ export function prioritizePeers(
     })
   );
 
-  const {attnetQueries, syncnetQueries, peerHasDuty} = requestAttnetPeers(
+  const {attnetQueries, syncnetQueries, dutiesByPeer} = requestAttnetPeers(
     connectedPeers,
     activeAttnets,
     activeSyncnets,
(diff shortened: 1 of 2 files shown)


dapplion:
@tuyennhv can you rebase on latest unstable?

twoeths:
Metrics on `feat1` is expected, peers are cut to 50 frequently

Screen Shot 2023-03-03 at 09 01 18

sometimes we disconnect peers with the new reason: "find better peers"

Screen Shot 2023-03-03 at 09 05 12

all other metrics are the same

dapplion (approved):
Looks good! Good point to add the distinct reason to track on metrics. Let's keep an eye on how often the nodes now get stuck with bad peers

wemeetagain:

PR #5220, opened 2023-03-01
Update default builder selection strategy to MaxProfit

g11tech:
Shift builder vs execution default selection strategy to max profit

making this as a default UX makes more sense for a retail user who has to supply less args :+1:


nflaig (commented):
It would be helpful to have an explainer in the PR of what the reasoning is behind this change

why did we use `BuilderAlways` before instead of `MaxProfit`?

g11tech:
> It would be helpful to have an explainer in the PR of what the reasoning is behind this change
> 
> why did we use `BuilderAlways` before instead of `MaxProfit`?

making this as a default UX makes more sense for a retail user who has to supply less args :+1:

wemeetagain (approved):

wemeetagain:

PR #5210, opened 2023-02-27
Minor dashboards fixes, grafana update related changes

nflaig:
**Motivation**

As suggested in https://github.com/ChainSafe/lodestar/pull/5200#pullrequestreview-1314823368, the initial PR https://github.com/ChainSafe/lodestar/pull/5200 should be split into two separate PRs

**Description**

- Almost all the changes in this PR are related to a newer grafana version
- Also fixes broken panel in vm+host dashboard which could not be edited, I think this was related to incorrect x/y coordinates
- Typo fix and few title updates, nothing special..

PR #5221, opened 2023-03-01
Fix grafana dashboard lint script

nflaig:
**Motivation**

Dashboard linter script was not catching some references to internal prometheus uids, see https://github.com/ChainSafe/lodestar/pull/5210#discussion_r1119555169.

**Description**

- Nested dashboard panels are now recursively checked
- Panel datasource must point to the datasource variable
- Fixed dashboard by running `scripts/validate-grafana-dashboards.sh`

**Further observations**

There are two observations unrelated to the changes in this PR which I am not sure are an issue or not

1. Grafana says that the `DS_PROMETHEUS` variable is not referenced by any variable or dashboard, this happens for all dashboards where we have `DS_PROMETHEUS` in `__inputs`, I think this happens though because on import it will replaces the variable with the acutal uid, in my case this seems to be `PBFA97CFB590B2093` and if I exported without sharing externally `"uid": "${DS_PROMETHEUS}"` is replaced by `"uid": "PBFA97CFB590B2093"`

image

2. if I import a dashboard and then export it without modification using `Export for sharing externally` there will always be a diff even if same grafana version is used, I am not sure why this replacements happens...it should not set some internal uid here. I guess just have to run `./scripts/validate-grafana-dashboards.sh` locally to fix this before commiting

```diff
@@ -84,7 +84,7 @@
       "collapsed": false,
       "datasource": {
         "type": "prometheus",
-        "uid": "${DS_PROMETHEUS}"
+        "uid": "PBFA97CFB590B2093"
       },
```

### DIFF
--- a/dashboards/libp2p.json
+++ b/dashboards/libp2p.json
@@ -698,7 +698,7 @@
         {
           "datasource": {
             "type": "prometheus",
-            "uid": "prometheus_local"
+            "uid": "${DS_PROMETHEUS}"
           },
           "fieldConfig": {
             "defaults": {
@@ -775,7 +775,7 @@
             {
               "datasource": {
                 "type": "prometheus",
-                "uid": "prometheus_local"
+                "uid": "${DS_PROMETHEUS}"
               },
               "editorMode": "code",
               "expr": "rate(libp2p_tcp_listener_events_total[$rate_interval])",
@@ -790,7 +790,7 @@
         {
           "datasource": {
             "type": "prometheus",
-            "uid": "prometheus_local"
+            "uid": "${DS_PROMETHEUS}"
           },
           "fieldConfig": {
             "defaults": {
@@ -867,7 +867,7 @@
             {
               "datasource": {
                 "type": "prometheus",
-                "uid": "prometheus_local"
+                "uid": "${DS_PROMETHEUS}"
               },
               "editorMode": "code",
               "expr": "rate(libp2p_tcp_listener_errors_total[$rate_interval])",
@@ -882,7 +882,7 @@
         {
           "datasource": {
             "type": "prometheus",
-            "uid": "prometheus_local"
+            "uid": "${DS_PROMETHEUS}"
           },
           "fieldConfig": {
             "defaults": {
@@ -977,7 +977,7 @@
             {
               "datasource": {
                 "type": "prometheus",
-                "uid": "prometheus_local"
+                "uid": "${DS_PROMETHEUS}"
               },
               "editorMode": "code",
               "expr": "libp2p_tcp_listener_status_info",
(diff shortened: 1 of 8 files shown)


nflaig:
@dapplion do you know why it is required to hard code a value here? it seems like it does not matter what value is used here as long as it is not `"${DS_PROMETHEUS}"`

https://github.com/ChainSafe/lodestar/blob/466c7547b1d9e36e1e5d5dab7218c145f4d0e0c5/scripts/lint-grafana-dashboards.mjs#L171

nflaig:
diff of lint script is hard to reason about because logic to assert panels was extracted into function to be able to call it recursively 

here is a better diff to understand changes done to `lint-grafana-dashboards.mjs`

```diff
diff --git a/scripts/lint-grafana-dashboards.mjs b/scripts/lint-grafana-dashboards.mjs
index 4082300d53..754045bc05 100755
--- a/scripts/lint-grafana-dashboards.mjs
+++ b/scripts/lint-grafana-dashboards.mjs
@@ -214,17 +214,12 @@ function assertTemplatingListItemContent(json, varName, item) {
  */
 function assertPanels(panels) {
   for (const panel of panels) {
+    // Panel datasource must point to the datasource variable
+    if (panel.datasource) {
+      panel.datasource.type = "prometheus";
+      panel.datasource.uid = `\${${variableNameDatasource}}`;
+    }
     if (panel.targets) {
       for (const target of panel.targets) {
         // All panels must point to the datasource variable
         if (target.datasource) {
           target.datasource.type = "prometheus";
+          target.datasource.uid = `\${${variableNameDatasource}}`;
-          target.datasource.uid = "${DS_PROMETHEUS}";
         }
 
         // Disable exemplar
@@ -238,9 +233,5 @@ function assertPanels(panels) {
         }
       }
     }
+    // Recursively check nested panels
+    if (panel.panels) {
+      assertPanels(panel.panels);
+    }
   }
 }
```

nflaig:
other interesting approach as mentioned in https://github.com/grafana/grafana/issues/10786#issuecomment-1277000930, recommends to replace all uids with null-string (`""`) but If we do that and click go to edit on a panel it wont preview the query.

image


Other repositories using lodestar dashboards + auto-provisioning are replacing `${DS_PROMETHEUS}` with the actual datasource id defined in the grafana config, for example see [eth-docker dashboard provisioning](https://github.com/eth-educators/eth-docker/blob/d1e21eb86f60cc7c9f57a9b845a1880faa38a255/grafana/provision-dashboards.sh#L66). This is not specific to Lodestar dashboard as all dashboards that allow to configure prometheus as variable have this issue.

Based on my observations, we could support auto-provisioning without workarounds if we remove the prometheus variable altogether or remove uid references as mentioned above. However, both of those approaches are not ideal as it would not longer be possible to specific which prometheus datasource to use in case there are multiple.

wemeetagain (approved):
Thanks for the thorough explanation

wemeetagain:

PR #5222, opened 2023-03-01
Lodestar prover for execution api 

### LINKED ISSUE #4706: Lodestar prover for execution api
### Description 

Execution land apps so far has been using JSON-RPC api (locally hosted or hosted by an infra or app provider). This places requirement of trust in the api provider which has lead to centralization of these services.

With ethereum PoS light client protocol, its now possible to verify the data served by the JSON-RPC `client` side. This creates a space for a "prover" library with which user can further decentralize trust.
Currently, a few provers a.la. kevlar and nimbus lc proxy have come up which basically offers a proxy to the web3 calls. This requires running another server which does the job of intercepting the calls to the backend rpc service and marry/prove it with the light client data.

However lodestar wants to approach it a bit differently. Lodestar wants to "programmatically" proxy so that the entire proving happens on the `client`. Obviously this can also be used to host a proxy rpc sever.

This  is the task for building such a module:

**ProofedProvider**
Takes a
  - a web3 provider  or options
  - a proofed root provider or options
 
and  returns a `safeProvider` which responses to the request with the proved data 

**Expected Usecase**

```typescript
import {ProofedWeb3Provider, ProofedRootProvider} from "@lodestar/prover"

const rpc:Web3Provider|URL = ....
const rootProvider: ProofedRootProvider|{mode: LightNode.Rest,apiUrl: string}|{mode: LightNode.P2P, bootnodes:string[]} = ....
const safeProvider = ProofedProvider(rpc ,rootProvider); 

// wait for safeProvider to sync 
await safeProvider.sync({checkpoint:"0x.."})

// you may stop query safe provider at any time synced or not
const {latest,finalized,status} = safeProvider.getStatus()

// you may call any web3 api at any time synced or not, but will error if number, latest. finalized can't be resolved
// calls by block root will not be proxided
safeProvider.eth_call(....)
safeProvider.eth_getBalance(...)
```

### Task List

- [x] Create a package skeleton `@lodestar/prover` 
- [x] Create a factory method to create a proofed web3 provider 
- [x] Create a factory method to create a proxy server 
- [x] Create a small CLI app to start the proxy and expose it as binary 
- [x] Add Rest interface for the lightclient 
- [x] Add P2P interface for the lightclient (Moved to https://github.com/ChainSafe/lodestar/issues/5644)
- [x] Optional logging support for the web3 provider 
- [x] Logging support for the proxy server to stream to std
- [x] Support to provide white list of fallback web3 methods for non-verified requests 
- [x] Check the given execution URLs for compatibility before start 
- [x] Run the e2e tests in the browser for the web3 provider (Covered with the existing browser tests)
- [x] Run the e2e tests in the Node for the web3 proxy (Moved to https://github.com/ChainSafe/lodestar/issues/5647)
- [x] Handle the re-org scenario for the proof provider (Moved to https://github.com/ChainSafe/lodestar/issues/5645)
- [x] Support batch RPC requests
- [x] Setup local EL/CL nodes for the CI
- [x] Use batch request inside `getExecutionPayloads` to optimize the performance 
- [x] Update all external requests to be restricted with timeout 
- [x] Update the logic for pruning the unfinalized headers (Moved to https://github.com/ChainSafe/lodestar/issues/5646)
- [x] Reduce dependency on etheruemjs packages (Out of scope for now)

### Web3 Methods to support 

The method support list is now moved to the [README](https://github.com/ChainSafe/lodestar/blob/unstable/packages/prover/README.md)

nazarhussain:
**Motivation**

Make it easier for third party developers to utilize the light client and verify the non-trusted eth responses. 

**Description**

Create a lightclient prover utility for web3 providers. 

Closes #4706

Run all tests. 

```sh
git checkout <feature_branch>
lodestar beacon --new-flag option1
```


dapplion (commented):
Overall direction looks good! :fire: looking forward to the polished PR

nazarhussain:
This PR functionality is ready and tested manually with the example given in the readme file. Now adding the unit and integration tests. 

Please share your earlier feedback.

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,200 @@
...
+
+export class ProofProvider {
+  private payloads: OrderedMap<string, allForks.ExecutionPayload> = new OrderedMap();
+  private finalizedPayloads: OrderedMap<string, allForks.ExecutionPayload> = new OrderedMap();
+  private readyPromise: Promise<void>;
+
+  // Map to match CL slot to EL block number
+  private slotsMap: {[slot: number]: string} = {};
+
+  lightClient?: Lightclient;
```
g11tech, line 29: wouldn't make it optional but assign in constructor

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  async waitToBeReady(): Promise<void> {
+    return this.readyPromise;
+  }
+
+  static buildWithRestApi(urls: string[], opts: RootProviderInitOptions): ProofProvider {
+    const config = createChainForkConfig(networksChainConfig[opts.network]);
+    const api = getClient({urls}, {config});
+    const transport = new LightClientRestTransport(api);
```
g11tech, line 46: should construct the lightclient here itself and send in the constructor

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  constructor(private options: RootProviderOptions) {
+    this.readyPromise = this.sync(options.checkpoint);
+  }
+
+  async waitToBeReady(): Promise<void> {
+    return this.readyPromise;
+  }
+
+  static buildWithRestApi(urls: string[], opts: RootProviderInitOptions): ProofProvider {
```
g11tech: change it into static method `init` and take the transport type as an option and do if then else for transport below and throw not implemented error for p2p

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,200 @@
...
+    const config = createChainForkConfig(networksChainConfig[opts.network]);
+    const api = getClient({urls}, {config});
+    const transport = new LightClientRestTransport(api);
+
+    return new ProofProvider({
+      ...opts,
+      config,
+      api,
+      transport,
+    });
```
g11tech, line 53: call sync from here itself than in the constructor

g11tech: call sync from here itself than in the constructor and assign to the ready promise

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,200 @@
...
+      transport,
+      genesisData,
+    });
+
+    assertLightClient(this.lightClient);
+    this.lightClient.start();
+    this.registerEvents();
+
+    const headSlot = this.lightClient.getHead().beacon.slot;
+    await this.lightClient?.sync(
```
g11tech: isn't what lightClient.start does?

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,200 @@
...
+  private async sync(checkpoint?: string): Promise<void> {
+    if (this.lightClient !== undefined) {
+      throw Error("Light client already initialized and syncing.");
+    }
+
+    const {api, config, transport} = this.options;
+    const checkpointRoot = await getSyncCheckpoint(api, checkpoint);
+    const genesisData = await getGenesisData(api);
+
+    this.lightClient = await Lightclient.initializeFromCheckpointRoot({
```
g11tech, line 74: this should just happen in init and just start should be called

readyPromise is a wait on the lightClient.status to turn into `started`

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    const headSlot = this.lightClient.getHead().beacon.slot;
+    await this.lightClient?.sync(
+      computeSyncPeriodAtSlot(headSlot - MAX_REQUEST_LIGHT_CLIENT_UPDATES),
+      computeSyncPeriodAtSlot(headSlot)
+    );
+
+    // Load the payloads from the CL
+    const startSlot = headSlot;
+    const endSlot = headSlot - MAX_PAYLOAD_HISTORY;
```
g11tech: you can't go beyond the finalized checkpoint's epoch as the beacon node itself might not have that data, so a max condition also has to be applied unless max payload history < 2 epochs

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,200 @@
...
+    const payloads = await getExecutionPayloads(this.options.api, startSlot, endSlot);
```
g11tech: would it be beneficial to fetch them dynamically on requirement? also why slot => endslot mapping is required?

may be we just need to track head, and finalized slot's number?

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,200 @@
...
+      const blockNumber = numberToHex(payload.blockNumber);
+      this.payloads.set(blockNumber, payload);
+      this.slotsMap[parseInt(slot)] = blockNumber;
+    }
+
+    // Load the finalized payload from the CL
+    const finalizedSlot = this.lightClient.getFinalized().beacon.slot;
+    const finalizedPayload = await getExecutionPayloads(this.options.api, finalizedSlot, finalizedSlot);
+    const finalizedBlockNumber = numberToHex(finalizedPayload[finalizedSlot].blockNumber);
+    this.finalizedPayloads.set(finalizedBlockNumber, finalizedPayload[finalizedSlot]);
```
g11tech: again why do we need this mapping, once a slot is finalized, all numbers corresponding to finalized's block number and before are finalized, why do we need this mapping

packages/prover/src/proof_provider/proof_provider.ts
nazarhussain: No. I believe `start` and `sync` are doing two different things.

packages/prover/src/proof_provider/proof_provider.ts
nazarhussain: I want the `ProofProvider` ready to when it's synced.

packages/prover/src/proof_provider/proof_provider.ts
nazarhussain: To construct lightclient we need to call some `async` behavior. I strictly don't want to make `init` as `async` to make it useful in the web3_provider context.

packages/prover/src/proof_provider/proof_provider.ts
nazarhussain: Can't do that because of the comment below. https://github.com/ChainSafe/lodestar/pull/5222#discussion_r1134191019

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,200 @@
...
+    if (typeof blockNumber === "number") {
+      const payload = this.payloads.get(numberToHex(blockNumber));
+      if (!payload) throw new Error(`No payload for blockNumber ${blockNumber}`);
+      return payload;
+    }
+
+    throw new Error(`Invalid blockNumber "${blockNumber}"`);
+  }
+
+  async update(update: allForks.LightClientHeader, finalized = false): Promise<void> {
```
g11tech: lets rename the arg `update` here to  or `lcHeader`

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,200 @@
...
+    const fork = this.options.config.getForkName(update.beacon.slot);
+
+    if (!isForkExecution(fork)) {
+      return;
+    }
+
+    if (isForkExecution(fork) && !("execution" in update)) {
```
g11tech: another check you need to make is that execution is not the defaultValue() as the upgraded `lcHeader` will have `defaultValue()` assigned to execution

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  private registerEvents(): void {
+    assertLightClient(this.lightClient);
+
+    this.options.signal.addEventListener("abort", () => {
+      this.lightClient?.stop();
+    });
+
+    this.lightClient.emitter.on(LightclientEvent.lightClientFinalityUpdate, async (data) => {
+      await this.update(data, true).catch((e) => {
```
g11tech: aren't you `typing` arg of lightclient header to this.update, data you should get here is of type update

Also it would be better to emit `latest` and `finalized` events from lightclient as thats what you mean to process here i think

nazarhussain: I didn't get that comment. Please explain it a bit.

packages/prover/src/proof_provider/proof_provider.ts
nazarhussain: Do we need to be worried if `execution` have default value as long as it's an execution fork?

packages/prover/src/utils/execution.ts
```diff
@@ -0,0 +1,108 @@
...
+  );
+
+  // Shresth Agrawal (2022) Patronum source code. https://github.com/lightclients/patronum
+  const account = Account.fromAccountData({
+    nonce: BigInt(proof.nonce),
+    balance: BigInt(proof.balance),
+    storageRoot: proof.storageHash,
+    codeHash: proof.codeHash,
+  });
+  return account.serialize().equals(expectedAccountRLP ? expectedAccountRLP : emptyAccountSerialize);
```
g11tech, line 81: this looks redundant, verify proof already checks for match of these things i think, will double check and revert what the correct condition should be

nazarhussain: I believe this is additional check for `storageHash` and `codeHash` because `trie.verifyProof` does not check both. I took this logic from `patronum` referred that references in our code.

packages/prover/src/proof_provider/proof_provider.ts
g11tech: you may use .then(lightclient => lightclient.start())

packages/prover/src/proof_provider/proof_provider.ts
g11tech: start does sync and much more, so external sync is not required.
you need to return a promise that resolves when lightClient.status is shifted to started, for that you may need to emit a syncronized event in the lightclient which will resolve this promise

packages/prover/src/proof_provider/proof_provider.ts
g11tech: yes, because execution is actually being bundled on/post `capella` only in the updates

packages/prover/src/proof_provider/proof_provider.ts
g11tech: https://github.com/ChainSafe/lodestar/pull/5222/files#diff-229bc1a2e4f16b99b96ae3bd25f0429ff996cf967786ac6235f53f8f4a5b1e2cR142 
not sure the types match

g11tech:
looks exciting on the first pass :clap: , will need to give more brain/review cycles to the PR

packages/prover/src/proof_provider/proof_provider.ts
nazarhussain: Yes the type matches to the `lightclient` events else TS would not have let pass it. Remember it's not the direct event handler, rather a function called from within event handler without any type casting.

packages/prover/src/cli/cmds/start/index.ts
```diff
@@ -0,0 +1,17 @@
+import {CliCommand, CliCommandOptions} from "../../../utils/command.js";
+import {startHandler} from "./handler.js";
+import {StartArgs, startOptions} from "./options.js";
+
+export const start: CliCommand<StartArgs> = {
```
g11tech: can we rename it to `proverProxyHandler` or something more meaningful

packages/prover/src/interfaces.ts
```diff
@@ -0,0 +1,51 @@
+import {ChainForkConfig} from "@lodestar/config";
+import {NetworkName} from "@lodestar/config/networks";
+import {ProofProvider} from "./proof_provider/proof_provider.js";
+import {ELRequestPayload, ELResponse} from "./types.js";
+
+export enum LightNode {
```
g11tech: LCTransport is more appropriate name

packages/prover/src/cli/cmds/start/options.ts
```diff
@@ -0,0 +1,81 @@
+import {NetworkName, networksChainConfig} from "@lodestar/config/networks";
+import {LightNode} from "../../../interfaces.js";
+import {CliCommandOptions} from "../../../utils/command.js";
+
+export type StartArgs = {
+  port: number;
+  network: string;
+  "execution-rpc": string;
+  mode: "rest" | "p2p";
```
g11tech: would like to use transport as the arg name

packages/prover/src/cli/cmds/start/options.ts
```diff
@@ -0,0 +1,81 @@
+import {NetworkName, networksChainConfig} from "@lodestar/config/networks";
+import {LightNode} from "../../../interfaces.js";
+import {CliCommandOptions} from "../../../utils/command.js";
+
+export type StartArgs = {
+  port: number;
+  network: string;
+  "execution-rpc": string;
```
g11tech: use use `executionUrls` and make it an array which we may support later

packages/prover/src/cli/cmds/start/options.ts
```diff
@@ -0,0 +1,81 @@
...
+  mode: "rest" | "p2p";
+  "beacon-rpc"?: string[];
```
g11tech: would prefer beaconUrls

packages/prover/src/cli/cmds/start/options.ts
```diff
@@ -0,0 +1,81 @@
...
+  "beacon-bootnode"?: string[];
+  checkpoint?: string;
+};
+
+export type StartOptions = {
+  network: NetworkName;
+  executionRpcUrl: string;
```
g11tech, line 17: keep them same name as in args for easy correlation

packages/prover/src/cli/cmds/start/options.ts
```diff
@@ -0,0 +1,81 @@
...
+  port: number;
+  mode: LightNode;
+  beaconRpcUrls: string[];
+  beaconBootNodes: string[];
+  checkpoint?: string;
```
g11tech: would prefer wsCheckpoint to reflect it shouldn't be older than weak subjectivity period, also reflect the same in description

packages/prover/tsconfig.build.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../tsconfig.build.json",
+  "include": ["src"],
+  "compilerOptions": {
+    "outDir": "lib",
+    // Had to add it because compiling error `Duplicate identifier 'Buffer'`
```
g11tech, line 6: i think the commend here will break json or just check why github is showing it red

packages/prover/src/proof_provider/proof_provider.ts
g11tech: ahh, so this event comes from light-client package  which has same name as one in beacon-node

I would tend to rename these events in light-client package as: `lightClientOptimisticHeader` and `lightClientFinalityHeader` and rename `this.update` to `this.processUpdateHeader`

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,205 @@
...
+    if (typeof blockNumber === "number") {
+      const payload = this.payloads.get(numberToHex(blockNumber));
+      if (!payload) throw new Error(`No payload for blockNumber ${blockNumber}`);
+      return payload;
+    }
+
+    throw new Error(`Invalid blockNumber "${blockNumber}"`);
+  }
+
+  async update(lcHeader: allForks.LightClientHeader, finalized = false): Promise<void> {
```
g11tech: rename to `processUpdateHeader` for better clarity or `processLCHeader`

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,205 @@
...
+    const fork = this.options.config.getForkName(lcHeader.beacon.slot);
+
+    if (!isForkExecution(fork)) {
+      return;
+    }
+
+    if (isForkExecution(fork) && !("execution" in lcHeader)) {
+      throw new Error("Execution payload is required for execution fork");
+    }
```
g11tech: !isForkExecution has already been checked

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,205 @@
...
+      computeSyncPeriodAtSlot(headSlot)
+    );
+
+    // Load the payloads from the CL
+    const {start, end} = await getSlotLimitsForPayloads(this.lightClient);
+    const payloads = await getExecutionPayloads(this.options.api, start, end);
+    for (const [slot, payload] of Object.entries(payloads)) {
+      const blockNumber = numberToHex(payload.blockNumber);
+      this.payloads.set(blockNumber, payload);
+      this.slotsMap[parseInt(slot)] = blockNumber;
```
g11tech: i don't understand slot to number mapping, all numbers <= finalized's number are finalized, all numbers <= head's number are present

if there is a mapping that needs to be stored is beacon block root => execution block root for the non-finalized set and tracking what is head block

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    // Load the finalized payload from the CL
+    const finalizedSlot = this.lightClient.getFinalized().beacon.slot;
+    const finalizedPayload = await getExecutionPayloads(this.options.api, finalizedSlot, finalizedSlot);
+    const finalizedBlockNumber = numberToHex(finalizedPayload[finalizedSlot].blockNumber);
+    this.finalizedSlot = finalizedSlot;
+    this.slotsMap[finalizedSlot] = finalizedBlockNumber;
```
g11tech: track finalized slot => number mapping separately and thats the only slot => number reference that makes sense

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  getStatus(): {latest: number; finalized: number; status: RunStatusCode} {
+    assertLightClient(this.lightClient);
```
g11tech: instead of throwing on unitialized light client, have an uninitialized status

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,205 @@
...
+
+    if (!isForkExecution(fork)) {
+      return;
+    }
+
+    if (isForkExecution(fork) && !("execution" in lcHeader)) {
+      throw new Error("Execution payload is required for execution fork");
+    }
+
+    const newPayload = (lcHeader as capella.LightClientHeader).execution;
```
g11tech: rename it to newPayloadHeader for clarity as this is not a full payload with transactions

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,205 @@
...
+    const blockNumber = numberToHex(newPayload.blockNumber);
+    const blockSlot = lcHeader.beacon.slot;
+    const existingPayload = this.payloads.get(blockNumber);
+
+    if (existingPayload && existingPayload.blockHash === newPayload.blockHash) {
```
g11tech: change it into nested if with outer if (existingPayload!==undeined)
and inner an if else

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    if (isForkExecution(fork) && !("execution" in lcHeader)) {
+      throw new Error("Execution payload is required for execution fork");
+    }
+
+    const newPayload = (lcHeader as capella.LightClientHeader).execution;
+    const blockNumber = numberToHex(newPayload.blockNumber);
+    const blockSlot = lcHeader.beacon.slot;
+    const existingPayload = this.payloads.get(blockNumber);
```
g11tech: for non finalized, only maintain mappings of roots to roots and refrain from using mapping of numbers as they are susceptible to reorg

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,205 @@
...
+    if (existingPayload && existingPayload.blockHash === newPayload.blockHash) {
+      if (finalized) {
+        this.finalizedSlot = blockSlot;
+      }
+
+      // We payload have the payload for this block
+      return;
+    }
+
+    if (existingPayload && existingPayload.blockHash !== newPayload.blockHash) {
```
g11tech: if you refer unfinalized by hashes and not numbers this problem will not arise

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,205 @@
...
+      // TODO: Need to decide either
+      // 1. throw an error
+      // 2. or just ignore it
+      // 3. or update the blockHash
+    }
+
+    const newPayloads = await getExecutionPayloads(this.options.api, blockSlot, blockSlot);
```
g11tech: newPayloadHeaders

packages/prover/src/utils/consensus.ts
```diff
@@ -0,0 +1,87 @@
...
+  if (res.ok) return res.response.data;
+
+  if (!res.ok && res.error.code === 404) {
+    return fetchNearestBlock(api, direction === "down" ? slot - 1 : slot + 1);
+  }
+
+  throw new Error(`Can not fetch nearest block for slot=${slot}`);
+}
+
+export async function getSlotLimitsForPayloads(lightClient: Lightclient): Promise<{start: number; end: number}> {
```
g11tech: could be renamed to getUnFinaliziedRange

packages/prover/src/utils/consensus.ts
```diff
@@ -0,0 +1,87 @@
...
+  [startSlot, endSlot] = [Math.min(startSlot, endSlot), Math.max(startSlot, endSlot)];
+
+  const payloads: Record<number, allForks.ExecutionPayload> = {};
+
+  let slot = endSlot;
+  let block = (await fetchNearestBlock(api, slot, "down")) as capella.SignedBeaconBlock;
+  payloads[block.message.slot] = block.message.body.executionPayload;
+
+  while (slot >= startSlot) {
+    const previousBlock = (await fetchNearestBlock(api, block.message.slot - 1, "down")) as capella.SignedBeaconBlock;
```
g11tech, line 49: you should do an array fetch here may be on a batch of request promises (for loop for batching in batch size for e.g.)

packages/prover/src/utils/consensus.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    if (block.message.body.executionPayload.parentHash === previousBlock.message.body.executionPayload.blockHash) {
+      payloads[block.message.slot] = block.message.body.executionPayload;
```
g11tech, line 52: this will easily break on reorgs, just store unfinalized payloads referenced by hash

only condition you need to check in the end if if finalized is ancesor of the head

packages/prover/test/e2e/web3_provider.test.ts
```diff
@@ -0,0 +1,45 @@
...
+        // `getProof` will always remain the non-verified method
+        // as we use it to create proof and verify
+        await expect(web3.eth.getProof("0xf97e180c050e5Ab072211Ad2C213Eb5AEE4DF134", [], "latest")).fulfilled;
+      });
+    });
+
+    describe("ethers", () => {
+      it("should connect to the network and call non-verified method", async () => {
+        const {provider} = createVerifiedExecutionProvider(
+          new ethers.JsonRpcProvider("https://lodestar-sepoliarpc.chainsafe.io"),
```
g11tech, line 33: i am not sure if we should rely on connecting to real world servers as the server uptimes can't be guaranteed espeicially during client software upgrades

packages/prover/src/web3_proxy.ts
```diff
@@ -0,0 +1,122 @@
...
+          fetchResponseBody(res).then(resolve).catch(reject);
+        }
+      );
+      req.write(JSON.stringify(payload));
+      req.end();
+    });
+  }
+
+  const proxyServer = http.createServer(function proxyRequestHandler(req, res) {
+    if (req.url === "/proxy") {
```
g11tech, line 69: would like to keep the url as `/prover` for better presentation

packages/prover/src/verified_requests/eth_getBalance.ts
```diff
@@ -0,0 +1,29 @@
...
+export const ethGetBalance: ELVerifiedRequestHandler<[address: string, block?: number | string], string> = async ({
+  handler,
+  payload,
+  rootProvider,
+}) => {
+  const {
+    params: [address, block],
+  } = payload;
+  const executionPayload = rootProvider.getExecutionPayload(block ?? "latest");
+  const proof = await getELProof(handler, [address, [], numberToHex(executionPayload.blockNumber)]);
```
g11tech: use hash not number to fetch

packages/prover/src/proof_provider/proof_provider.ts
nazarhussain: I don't want to inject the lightclient to proof provider. 

1. Lightclient initialization is strictly an `async` process because of some BLS issue documented there. So don't want to use it in a strictly non-async `init` method of proof provider. 
2. Directly calling the `Lightclient` constructor would require to duplicate some logic used in light client initialization. 

I think we should proceed with the current implementation as of now. Once we add other transport modes, there may appear a need refactor that logic.

packages/prover/src/cli/cmds/start/options.ts
nazarhussain: Renamed to `execution-url`. I don't see a reason to have it array for now. If we make it plural for now user can mislead to behavior.

packages/prover/src/proof_provider/proof_provider.ts
nazarhussain: No it's not checked already.

packages/prover/tsconfig.build.json
nazarhussain: Github parse JSON as normal json files where comments are not allowed. And TS parse these as JSON5 files where comments are allowed.

packages/prover/test/e2e/web3_provider.test.ts
nazarhussain: I had added it as task in the main issue to setup a local EL/CL servers for CI.

packages/prover/src/web3_proxy.ts
nazarhussain: The `/` is the actual prover url and `/proxy` is just forwarding endpoint. So this seems ok to me.

packages/prover/src/utils/consensus.ts
nazarhussain: I had added an extra task to epic to refactor this logic to use the batch RPC requests.

packages/prover/src/utils/consensus.ts
nazarhussain: that's just an extra safety check while fetching the blocks from the API. It can be skipped. The re-org scenario is covered on different place.

packages/prover/src/cli/cmds/start/options.ts
```diff
@@ -0,0 +1,81 @@
...
+  port: number;
+  network: string;
+  "execution-url": string;
+  transport: "rest" | "p2p";
+  "beacon-urls"?: string[];
+  "beacon-bootnodes"?: string[];
+  "ws-checkpoint"?: string;
+};
+
+export type StartOptions = {
```
g11tech, line 15: can we change this to {...} & ({ transport:LCTransport.rest,beaconUrls:..} | {transport:LCTransport.p2p,beaconBootnodes:...}

packages/prover/src/interfaces.ts
```diff
@@ -0,0 +1,51 @@
...
+  Rest = "Rest",
+  P2P = "P2P",
+}
+
+export type RootProviderInitOptions = {
+  network: NetworkName;
+  signal: AbortSignal;
+  config?: ChainForkConfig;
+  wsCheckpoint?: string;
+} & ({transport: LCTransport.Rest; urls: string[]} | {transport: LCTransport.P2P; bootnodes: string[]});
```
g11tech, line 16: lets keep the same name of `urls` or `bootnodes` in cli/options and here as well so as to just have a clear correspondence that they are all the same

packages/prover/src/proof_provider/payload_store.ts
```diff
@@ -0,0 +1,154 @@
...
+
+type BlockELRoot = string;
+type BlockCLRoot = string;
+
+/**
+ * The in-memory store for the execution payloads to be used to verify the proofs
+ */
+export class PayloadStore {
+  // We store the block numbers only for finalized blocks
+  private finalizedRoots = new OrderedMap<BlockELRoot>();
```
g11tech, line 16: nice

packages/prover/src/proof_provider/payload_store.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  // Unfinalized blocks are stored by the roots of the beacon chain
+  private unfinalizedRoots = new Map<BlockCLRoot, BlockELRoot>();
+
+  // Payloads store with BlockELRoot as key
+  private payloads = new Map<BlockELRoot, allForks.ExecutionPayload>();
```
g11tech, line 22: :+1:

packages/prover/src/proof_provider/payload_store.ts
```diff
@@ -0,0 +1,154 @@
...
+    if (finalizedMaxRoot) {
+      return this.payloads.get(finalizedMaxRoot);
+    }
+
+    return undefined;
+  }
+
+  get latest(): allForks.ExecutionPayload | undefined {
+    if (this.latestBlockRoot) {
+      return this.payloads.get(this.latestBlockRoot);
```
g11tech, line 39: :+1:

packages/prover/src/proof_provider/payload_store.ts
```diff
@@ -0,0 +1,154 @@
...
+      return this.getOrFetchFinalizedPayload(blockId);
+    }
+
+    return undefined;
+  }
+
+  async getOrFetchFinalizedPayload(blockNumber: number): Promise<allForks.ExecutionPayload | undefined> {
+    if (blockNumber > this.finalizedRoots.max) {
+      throw new Error(
+        `Block number ${blockNumber} is higher than the latest finalized block number. We recommend to use block hash for unfinalized blocks.`
```
g11tech, line 72: :+1:

packages/prover/src/proof_provider/payload_store.ts
```diff
@@ -0,0 +1,154 @@
...
+      );
+    }
+
+    let blockELRoot = this.finalizedRoots.get(blockNumber);
```
g11tech, line 76: a comment here would be nice:
`\\ check if we have payload cached locally else fetch from api`

packages/prover/src/proof_provider/payload_store.ts
```diff
@@ -0,0 +1,154 @@
...
+      return this.payloads.get(blockELRoot);
+    }
+
+    return undefined;
+  }
+
+  set(payload: allForks.ExecutionPayload, finalized: boolean): void {
+    const blockRoot = bufferToHex(payload.blockHash);
+    this.payloads.set(blockRoot, payload);
+    this.latestBlockRoot = blockRoot;
```
g11tech: if you are fetching payloads in getOrFetchFinalizedPayload and setting there, the latestBlockRoot is getting updated as well from finalized roots, is this correct behavior?

packages/prover/src/proof_provider/payload_store.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    if (finalized) {
+      this.finalizedRoots.set(payload.blockNumber, blockRoot);
+    }
+  }
+
+  async processLCHeader(header: capella.LightClientHeader, finalized = false): Promise<void> {
```
g11tech, line 109: why are we not using allForks.LightClientHeader?

packages/prover/src/proof_provider/payload_store.ts
```diff
@@ -0,0 +1,154 @@
...
+    const blockSlot = header.beacon.slot;
+    const blockNumber = header.execution.blockNumber;
+    const blockELRoot = bufferToHex(header.execution.blockHash);
+    const blockCLRoot = bufferToHex(header.beacon.stateRoot);
+    const existingELRoot = this.unfinalizedRoots.get(blockCLRoot);
+
+    if (finalized) {
+      if (existingELRoot) {
+        this.unfinalizedRoots.delete(blockCLRoot);
```
g11tech, line 124: we might need to delete all `unfinalizedRoots` which have slot <= finalized

packages/prover/src/proof_provider/payload_store.ts
```diff
@@ -0,0 +1,154 @@
...
+
+      return;
+    }
+
+    if (existingELRoot && existingELRoot === blockELRoot) {
+      // We already have the payload for this block
+      return;
+    }
+
+    if (existingELRoot && existingELRoot !== blockELRoot) {
```
g11tech, line 146: can we chain all above three `if` as `if else if else if...` for better understanding of the flow?

packages/prover/src/proof_provider/payload_store.ts
```diff
@@ -0,0 +1,154 @@
...
+    if (finalized) {
+      this.finalizedRoots.set(payload.blockNumber, blockRoot);
+    }
+  }
+
+  async processLCHeader(header: capella.LightClientHeader, finalized = false): Promise<void> {
+    const blockSlot = header.beacon.slot;
+    const blockNumber = header.execution.blockNumber;
+    const blockELRoot = bufferToHex(header.execution.blockHash);
+    const blockCLRoot = bufferToHex(header.beacon.stateRoot);
```
g11tech, line 113: can we add a jsdoc here that blockCLRoot is tracking a beacon block by stateRoot

packages/prover/src/proof_provider/payload_store.ts
```diff
@@ -0,0 +1,154 @@
...
+      return;
+    }
+
+    if (existingELRoot && existingELRoot === blockELRoot) {
+      // We already have the payload for this block
+      return;
+    }
+
+    if (existingELRoot && existingELRoot !== blockELRoot) {
+      // Re-org happened, we need to update the payload
```
g11tech: If i understand correctly, I think el root by CL stateroot can never differ since stateRoot has executionPayloadHeader 's merkel hash included in it, if this happens, this is failure of consensus and is a critical error and you should just reject the header

packages/prover/src/proof_provider/payload_store.ts
```diff
@@ -0,0 +1,154 @@
...
+      this.payloads.delete(existingELRoot);
+      this.unfinalizedRoots.set(blockCLRoot, blockELRoot);
+    }
+
+    this.payloads.set(blockELRoot, (await getExecutionPayloads(this.opts.api, blockSlot, blockSlot))[blockSlot]);
+    this.latestBlockRoot = blockELRoot;
```
g11tech: if the event is finalized, then latestBlockRoot should not be updated only if slot >= current latestBlockRoot's slot

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,182 @@
...
+type RootProviderOptions = Omit<RootProviderInitOptions, "transport"> & {
+  transport: LightClientRestTransport;
+  api: Api;
+  config: ChainForkConfig;
+};
+
+export class ProofProvider {
+  private store: PayloadStore;
+
+  private readyPromise?: Promise<void>;
```
g11tech, line 28: add comment: `make sure readyPromise doesn't throw unhandled exceptions`

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,182 @@
...
+    const transport = new LightClientRestTransport(api);
+
+    const provider = new ProofProvider({
+      ...opts,
+      config,
+      api,
+      transport,
+    });
+
+    provider.readyPromise = provider.sync(opts.wsCheckpoint);
```
g11tech: add a .catch to `provider.sync(...).catch(e=>{})` to make sure it doesn't throw unhandled exceptions

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,182 @@
...
+    const {start, end} = await getUnFinalizedRangeForPayloads(this.lightClient);
+    const payloads = await getExecutionPayloads(this.opts.api, start, end);
+    for (const payload of Object.values(payloads)) {
+      this.store.set(payload, false);
+    }
+
+    // Load the finalized payload from the CL
+    const finalizedSlot = this.lightClient.getFinalized().beacon.slot;
+    const finalizedPayload = await getExecutionPayloads(this.opts.api, finalizedSlot, finalizedSlot);
+    this.store.set(finalizedPayload[finalizedSlot], true);
```
g11tech, line 105: you can use a better pattern of Object.values().forEach(payload=>this.store.set(payload,true/false)) here and everywhere else

packages/prover/src/proof_provider/proof_provider.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    if (!isForkExecution(fork)) {
+      return;
+    }
+
+    if (
+      isForkExecution(fork) &&
+      (!("execution" in lcHeader) || lcHeader.execution === ssz.capella.ExecutionPayloadHeader.defaultValue())
+    ) {
+      throw new Error("Execution payload is required for execution fork");
```
g11tech, line 160: you should check for isForkWithdrawals or something here as only they will have a non zero execution, also you can't === objects you will need to use : `ssz.capella.ExecutionPayloadHeader.equals(lcHeader.execution,ssz.capella.ExecutionPayloadHeader.defaultValue())

also the ssz type for above will depend upon the fork, so you will actually have to do:

`const sszType =config.getForkExecutionTypes(lcHeader.beacon.slot).ExecutionPayloadHeader`
and then check `sszType.equals(lcHeader.execution,sszType.defaultValue())`

packages/prover/src/utils/execution.ts
```diff
@@ -0,0 +1,108 @@
...
+export async function processAndVerifyRequest({
+  payload,
+  handler,
+  proofProvider,
+}: {
+  payload: ELRequestPayload;
+  handler: ELRequestMethod;
+  proofProvider: ProofProvider;
+}): Promise<ELResponse | undefined> {
+  await proofProvider.waitToBeReady();
```
g11tech, line 27: should we race here against a timeout?

packages/prover/src/utils/execution.ts
```diff
@@ -0,0 +1,108 @@
...
+  // eslint-disable-next-line no-console
+  console.warn(`Request handler for ${payload.method} is not implemented.`);
+  return handler(payload);
+}
+
+export async function getELProof(
+  handler: ELRequestMethod,
+  args: [address: string, storageKeys: string[], block: number | string]
+): Promise<ELProof> {
+  // TODO: Find better way to generate random id
```
g11tech, line 43: shouldn't id be increasing sequence? i am not exactly sure what jsonrpc spec says

packages/prover/src/utils/consensus.ts
g11tech: yes just skip the check as anyway this mapping is not canonical since it unfinalized and will cause intractable logical errors

packages/prover/src/proof_provider/payload_store.ts
nazarhussain: Because we are intended to use `capella` upward, earlier forks don't have execution payloads.

packages/prover/src/proof_provider/payload_store.ts
nazarhussain: I prefer guard conditions over `if else` structure. Had simplified it conditions a bit.

packages/prover/src/proof_provider/payload_store.ts
nazarhussain: Can understand JSDoc for a local variable? Had added doc comments to class members.

packages/prover/src/proof_provider/payload_store.ts
nazarhussain: My understanding is, the reorg can happen on the CL and EL side. Each fork on CL can point to different fork on EL. So incase we receive a header update which for which we already have a payload in record but have different root that's a sign of such reorg. In that case we consider new version to be right but yet unfinalized.

packages/prover/src/proof_provider/proof_provider.ts
nazarhussain: `Object.values` is used where it's intended use. On this line we need to set a single payload so didn't used it.

packages/prover/src/proof_provider/proof_provider.ts
nazarhussain: Updated the `sszType` logic. For the `isForkExecution` seems correct to me. If that's not correct then we may need to update `isForkExecution`.

packages/prover/src/utils/execution.ts
nazarhussain: Added an action item to the list, better to be done on multiple places in a separate PR.

packages/prover/src/utils/execution.ts
nazarhussain: `jsonrpc` standard is flexible about the ID. The incremental behavior may collide with internal provider requests.

packages/prover/src/proof_provider/payload_store.ts
g11tech: yes pls add for more clarity to the code reader (or comment) it took me sometime to follow what is goin on with the condition checks below

packages/prover/src/proof_provider/payload_store.ts
g11tech: then use Exclude<allforks, altair|bellatrix>

packages/prover/src/proof_provider/payload_store.ts
g11tech: nothing else needs to be change, just chain them with `else if`

packages/prover/src/proof_provider/proof_provider.ts
g11tech: bellatrix which is execution fork might not have `execution` in LCHeader and should just return instead of throwing error, also could you chain this with above confition on L151 with `if else`

nazarhussain: I don't get your point. If `bellatrix` is an execution fork why would it not have `execution` in the header?

packages/prover/src/proof_provider/payload_store.ts
nazarhussain: I prefer the guard conditions over `if else` that makes the code more readable. I think we should just pass on this formatting preference.

packages/prover/src/proof_provider/payload_store.ts
nazarhussain: I believe the names of the variables `blockCLRoot, `blockELRoot`, `blockNumber` and `blockSlot` are very clear and self explanatory. No need to add comments for these.

packages/prover/src/proof_provider/payload_store.ts
nazarhussain: We can't use the type namespaces with the `utility types`. I believe Caymen was wroking on type refactor where we can get such extended types. We can revisit this logic when we have those types.

packages/prover/src/proof_provider/payload_store.ts
nazarhussain: We don't store the `slot` numbers for the `unfinalizedRoots` so can't delete those with that condition. I added an action item to revisit that logic for future. Once we add extended unit tests this logic will be easy to verify and refactor.

packages/prover/src/proof_provider/proof_provider.ts
g11tech: execution is only bundled with LightClient header on/after capella

nazarhussain: @g11tech Updated the code to `isForkWithdrawals`

packages/prover/src/proof_provider/payload_store.ts
g11tech: but you have payloads stored for them, so do a look and cleanup

packages/prover/src/proof_provider/payload_store.ts
g11tech: i think: `Exclude<allforks.LightClientHeader, altair.LightClientHeader | bellatrix.LightClientHeader>` can be done, no?

g11tech (approved):
approving with the caveat that followup PRs would need to polish and address some of the unaddressed issues here to productionize the `prover`

But an amazing effort here to bring such an amazing functionality to lodestar codebase! :clap:

nazarhussain:
> approving with the caveat that followup PRs would need to polish and address some of the unaddressed issues here to productionize the `prover`
> 
> But an amazing effort here to bring such an amazing functionality to lodestar codebase! 👏

All the remaining points are added as action item to this issue https://github.com/ChainSafe/lodestar/issues/4706 so we will not forget it.

wemeetagain:

PR #5232, opened 2023-03-04
chore: cancel previous workflows on new push to the PRs

g11tech:
Cancel previous workflows on new push to the PRs

grouping strategy to cancel any previous runs:
```
  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```

head_ref is only defined on PRs and will group the runs per PR and on stable/unstable the unique run id will cause each PR merge commit to be grouped separately resulting into  CI being run on each merged PR commit.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,9 +1,10 @@
 name: Benchmark
 
-# only one can run at a time.
 # Actions access a common cache entry and may corrupt it.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  # If PR, cancel prev commits. head_ref = source branch name on pull_request, null if push
+  # if push, default to github.ref so retries of the same commit do not overwrite each other
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
   cancel-in-progress: true
 
 on:
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,8 +1,10 @@
 name: Benchmark
 
-# only one can tun at a time.
+# only one can run at a time.
 # Actions access a common cache entry and may corrupt it.
-concurrency: cd-benchmark-${{ github.ref }}
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
 
 on:
   push:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -11,6 +11,11 @@
 #
 name: "CodeQL"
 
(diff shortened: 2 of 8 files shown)


.github/workflows/benchmark.yml
```diff
@@ -1,8 +1,10 @@
 name: Benchmark
 
-# only one can tun at a time.
+# only one can run at a time.
 # Actions access a common cache entry and may corrupt it.
-concurrency: cd-benchmark-${{ github.ref }}
```
dapplion, line 5: Note this change breaks the original intent of running strictly once per github ref

.github/workflows/codeql-analysis.yml
```diff
@@ -11,6 +11,11 @@
 #
 name: "CodeQL"
 
+# only one can run at a time.
```
dapplion: This looks like it would cause PRs with the same head_ref to cancel each other right?

dapplion (changes requested):

.github/workflows/benchmark.yml
g11tech: you mean strictly one at a time?

if i change it to : `${{ github.workflow }}-${{ github.ref }}` would it be ok to cancel the previous ongoing run?

or I can copy the benchmark which runs in original way on stable, unstable and benchmark , and modify this to run on the PRs only with the cancel-in-progess true

.github/workflows/codeql-analysis.yml
g11tech: yes this causes a new commit on the same PR to cancel run for the previous commit (if still running) and only the new commit workflows work 

for e.g. to test i pushed another commit just after starting this PR while original worksflows were running resulting into previous ones being cancelled shortly

image

and only the last commit ones running.

dapplion: Right a meant that 2 or more PRs with the same head branch will cancel each other. Is there a way to identify individual PRs?

g11tech: umm head_ref is the source branch for the PR, would it be separate for separate PRs?

image

dapplion: What about using `github.event.number`? Pull request events are:
- on open: `open` https://docs.github.com/webhooks-and-events/webhooks/webhook-events-and-payloads?actionType=opened#pull_request
- on push: `synchronize` https://docs.github.com/webhooks-and-events/webhooks/webhook-events-and-payloads?actionType=synchronize#pull_request

g11tech: but event number will not make the same PR commits in the same group for the new ones to cancel the last running

dapplion (approved):
Looks good to me! Added comments to explain behavior

wemeetagain:

PR #5234, opened 2023-03-06
Comment on PRs included in release

nflaig:
**Motivation**

Make it easier to track in which release a PR was included.

**Description**

Comment on PRs included in stable release by adding [GitHub Action: Comment on Pull Requests included in Release](https://github.com/marketplace/actions/comment-on-pull-requests-included-in-release) to publish stable workflow.

By default the message looks like this:

🎉 This PR is included in [v1.5.0](https://github.com/ChainSafe/lodestar/releases/tag/v1.5.0) 🎉

We can customize that if needed by setting the `message`, see [inputs](https://github.com/marketplace/actions/comment-on-pull-requests-included-in-release#inputs).

### DIFF
--- a/.github/workflows/publish-stable.yml
+++ b/.github/workflows/publish-stable.yml
@@ -114,6 +114,16 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+  comment:
+    name: Comment on included PRs
+    runs-on: buildjet-4vcpu-ubuntu-2204
+    needs: [tag, npm]
+    if: needs.tag.outputs.is_stable == 'true'
+    steps:
+      - uses: nflaig/release-comment-on-pr@v1
+        with:
+          token: ${{ secrets.GH_PAGES_TOKEN }}
+
   docker:
     name: Publish to Docker Hub
     runs-on: buildjet-4vcpu-ubuntu-2204


.github/workflows/publish-stable.yml
```diff
@@ -48,6 +48,16 @@ jobs:
...
 
+  comment:
+    name: Comment on PRs included in release
+    runs-on: buildjet-4vcpu-ubuntu-2204
+    needs: tag
+    if: needs.tag.outputs.is_stable == 'true'
+    steps:
+      - uses: nflaig/release-comment-on-pr@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
nflaig, line 59: @wemeetagain as noted in the [usage section](https://github.com/marketplace/actions/comment-on-pull-requests-included-in-release#usage) of the gh action, this token needs the `public_repo` permission which is not the case by default. Might need to create a [personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) in that case and use that one instead.

image

.github/workflows/publish-stable.yml
```diff
@@ -114,6 +114,16 @@ jobs:
...
 
+  comment:
+    name: Comment on PRs
+    runs-on: buildjet-4vcpu-ubuntu-2204
+    needs: [tag, npm]
+    if: needs.tag.outputs.is_stable == 'true'
+    steps:
+      - uses: nflaig/release-comment-on-pr@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
nflaig: @wemeetagain as noted in the [usage section](https://github.com/marketplace/actions/comment-on-pull-requests-included-in-release#usage) of the gh action, this token needs the `public_repo` permission which is not the case by default. Might need to create a [personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) in that case and use that one instead.

image

.github/workflows/publish-stable.yml
```diff
@@ -114,6 +114,16 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+  comment:
+    name: Comment on PRs
+    runs-on: buildjet-4vcpu-ubuntu-2204
+    needs: [tag, npm]
```
nflaig, line 120: to be considered if we want to run this before or after `docker` job, technically can run it before as it is currently implemented

.github/workflows/publish-stable.yml
wemeetagain: We have a secret named GH_PAGES_TOKEN which should have that permission

.github/workflows/publish-stable.yml
wemeetagain: Only thing I'd want to make sure of is that it can't/won't fail and stop the docker step from running.

nflaig: yes definitely, I haven't worked that much with github workflows but my assumption here is that it will run at the same time as the `docker` job as they are independent.

.github/workflows/publish-stable.yml
nflaig: wondering if we can dy-run that somehow, the comment will then be added by the github pages bot account I would assume, so the only requirement is that the account is able to post comments.

wemeetagain (approved):

wemeetagain:

PR #5239, opened 2023-03-07
Track validator client statuses as metrics

### LINKED ISSUE #5158: Add VC metric to track validator statuses

maschad:
Closes https://github.com/ChainSafe/lodestar/issues/5158


packages/beacon-node/src/monitoring/clientStats.ts
```diff
@@ -187,11 +187,46 @@ function createValidatorStats(): ClientStats {
     }),
     validatorActive: new MetricProperty({
       jsonKey: "validator_active",
-      metricName: "vc_indices_count",
+      metricName: "vc_active_validators_count",
       jsonType: JsonType.Number,
       defaultValue: 0,
       description: "Number of validator keys that are currently active",
     }),
+    validatorPending: new MetricProperty({
```
nflaig: this is the definition of client stats based on [this spec](https://docs.google.com/document/d/1qPWAVRjPCENlyAjUBwGkHMvz9qLdd_6u9DPZcNxDBpc/edit), only need to update `validatorActive` property

packages/beacon-node/src/monitoring/clientStats.ts
```diff
@@ -187,11 +187,46 @@ function createValidatorStats(): ClientStats {
...
+      jsonKey: "validator_pending",
+      metricName: "vc_pending_validators_count",
```
nflaig: where are these metrics coming from?

nflaig (changes requested):

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -821,7 +821,31 @@ export function createLodestarMetrics(
         name: "validator_monitor_validators",
         help: "Count of validators that are specifically monitored by this beacon node",
       }),
-
+      validatorsActive: register.gauge({
+        name: "validator_monitor_validators_active",
+        help: "Count of validators monitored by this beacon node that are active",
+      }),
```
dapplion: All of this should be the same metric with label: "status" where status can have values: "active", "pending", etc

maschad: I had originally [done something similar](https://github.com/ChainSafe/lodestar/pull/5239#discussion_r1127489060) but it goes against [the spec](https://docs.google.com/document/d/1qPWAVRjPCENlyAjUBwGkHMvz9qLdd_6u9DPZcNxDBpc/edit?usp=drivesdk) , but I agree with you in that this would be more simplified programmatically.

packages/validator/src/metrics.ts
```diff
@@ -325,6 +325,12 @@ export function getMetrics(register: MetricsRegister, gitData: LodestarGitData)
       help: `Current health status of the beacon(s) the validator is connected too. ${renderEnumNumeric(BeaconHealth)}`,
     }),
 
+    validatorStatuses: register.gauge<{status: string}>({
+      name: "vc_validator_status",
```
nflaig: could consider plural in the metric name

```suggestion
      name: "vc_validator_statuses",
```

or acutally I think to align it more with other metrics such as [vc_doppelganger_validator_status_count](https://github.com/ChainSafe/lodestar/blob/7cbfa97282e8b4d99dbabb2105bc99a208bff689/packages/validator/src/metrics.ts#LL437C32-L437C54), need to do this

```suggestion
      name: "vc_validator_status_count",
```

packages/beacon-node/src/monitoring/clientStats.ts
```diff
@@ -187,7 +187,7 @@ function createValidatorStats(): ClientStats {
     }),
     validatorActive: new MetricProperty({
       jsonKey: "validator_active",
-      metricName: "vc_indices_count",
+      metricName: "vc_active_validators_count",
```
nflaig: to get value from a metric based on label see `libp2p_data_transfer_bytes_total` as an example

https://github.com/ChainSafe/lodestar/blob/9c7c08b4adf9d04cd2e5f1084ed999a1a230a422/packages/beacon-node/src/monitoring/clientStats.ts#L112-L119

packages/beacon-node/src/metrics/validatorMonitor.ts
```diff
@@ -287,6 +294,39 @@ export function createValidatorMonitor(
...
+        case "exited_unslashed":
+        case "withdrawal_done":
+          metrics.validatorMonitor.validatorStatus.inc({status: "exited"});
+          break;
+        case "pending_initialized":
+        case "pending_queued":
+          metrics.validatorMonitor.validatorStatus.inc({status: "pending"});
+          break;
+        case "withdrawal_possible":
+          metrics.validatorMonitor.validatorStatus.inc({status: "withdrawalable"});
```
nflaig: shouldn't these be the same as the simple validator statuses (`SimpleValidatorStatus`)?

if we keep it as is, need to fix typo

```suggestion
          metrics.validatorMonitor.validatorStatus.inc({status: "withdrawable"});
```

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -821,7 +821,16 @@ export function createLodestarMetrics(
         name: "validator_monitor_validators",
         help: "Count of validators that are specifically monitored by this beacon node",
       }),
-
+      validatorStatus: register.gauge<"status">({
+        name: "validator_monitor_validators_status",
+        help: "Count of validators monitored by this beacon node that are active",
```
nflaig: help text seems outdated

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -821,7 +821,16 @@ export function createLodestarMetrics(
...
+        labelNames: ["status"],
+      }),
+      validatorBalances: register.gauge<"index">({
+        name: "validator_monitor_validator_balances",
```
nflaig: the naming of `validatorStatus` and `validatorBalances` is a bit inconsistent

packages/beacon-node/src/monitoring/clientStats.ts
```diff
@@ -187,7 +187,8 @@ function createValidatorStats(): ClientStats {
...
```
nflaig:
```suggestion
      metricName: "vc_validator_status_count",
```

packages/beacon-node/src/metrics/validatorMonitor.ts
```diff
@@ -306,21 +306,21 @@ export function createValidatorMonitor(
         case "active_exiting":
         case "active_slashed":
         case "active_ongoing":
-          metrics.validatorMonitor.validatorsActive.inc();
+          metrics.validatorMonitor.validatorStatus.inc({status: "active"});
```
dapplion: What are you tracking here exactly? The total number of counts validators with a specific status were seem in an epoch? Not sure this metrics represents the instantaneous count of connected validators with a specific status

maschad: I am not sure I understand. This function would be called while iterating through an array of the validator indices, and each registered validator should have a status, therefore the metrics object would then increment each status.

nflaig: we want to track the status of local validators here right? and assuming `registerValidatorStatuses` is called every epoch we would just increment this indefinitely

packages/beacon-node/src/metrics/validatorMonitor.ts
```diff
@@ -294,37 +293,43 @@ export function createValidatorMonitor(
...
+      for (const monitoredValidator of validators.values()) {
+        const index = monitoredValidator.index;
 
-      metrics.validatorMonitor.validatorBalances.set({index}, balance);
+        switch (statuses[index]) {
+          case "active":
+          case "active_exiting":
+          case "active_slashed":
+          case "active_ongoing":
+            metrics.validatorMonitor.validatorStatuses.inc({status: "active"});
```
dapplion: Much more efficient to compute this value first and call .set() once

maschad: are you sure? It would require two iterations as opposed to one, as well as some extra space allocation for the incremented variables,  I don't think the call to `inc()` is more expensive than that.

packages/validator/src/validator.ts
```diff
@@ -303,9 +312,13 @@ export class Validator {
     // The total number of validators
     const total = pubkeysHex.length;
 
-    this.logger.info("Validator statuses", {...statuses, total});
+    if (metrics) {
+      for (const [status, amount] of allValidatorStatuses) {
+        metrics.validatorStatuses.inc({status}, amount);
```
nflaig: shouldn't this use `set` instead?

```suggestion
        metrics.validatorStatuses.set({status}, amount);
```

packages/validator/src/validator.ts
```diff
@@ -159,6 +162,24 @@ export class Validator {
       this.clock.start(this.controller.signal);
       this.chainHeaderTracker.start(this.controller.signal);
 
+      const validatorKeys = this.validatorStore.getValidatorKeys();
+
+      for (const pubkeyHex in validatorKeys) {
+        if (!indicesService.validatorPubKeyExists(pubkeyHex)) {
+          this.logger.info("Validator exists in beacon chain", {
```
nflaig, line 177: won't this still print out `Validator exists in beacon chain` for each validator?

definitely looking forward to having the fee recipient printed out 🙌

packages/validator/src/validator.ts
maschad: Thanks! Good catch!

packages/validator/src/validator.ts
maschad: It shouldn't, given that a validator index would have existed in the `pubkey2index` Mapping had it been discovered prior to this. Therefore it would only be for newly discovered validators by the indices service.

packages/beacon-node/src/metrics/validatorMonitor.ts
dapplion: > I don't think the call to inc() is more expensive than that.

I would say so, please check prom-client source for the case of increasing with labels

maschad: Good point, it could be more expensive depending on the amount of statuses given this [validateLabel](https://github.com/siimon/prom-client/blob/2c37e80c1bc14cd27eaeecd73cc36ad268856a2e/lib/validation.js#L17) check which iterates through all the labels.

packages/validator/src/validator.ts
nflaig: that's if the service is already running and we import and new validator I would assume but what happens if the service starts, isn't `pubkey2index` just stored in-memory?

maschad: Yes in that case it would print it for every validator, but I would suggest that's technically accurate since at the start of the service the validator client would be discovering all indices since they would all be new to the client.

nflaig: yeah I think should print that for each validator, just thought you wanted to address what I mentioned in https://github.com/ChainSafe/lodestar/pull/5161#discussion_r1133312778

maschad: Sorry I think I had misinterpreted what was written there, following the [lighthouse ux](https://github.com/ChainSafe/lodestar/issues/4785#issuecomment-1324702738) I think we should print this out for each validator, I do agree it can be a bit noisy upon initialization but I think it's useful information.

nflaig: It definitely is needed information if we discover a new validator while the VC is running, however I am not so sure at startup, I could imagine a log where if validators are >= 2 we print out `Validators exists in beacon chain`, followed by a list of validators (index, pubkey, fee recipient). 

We can definitely also print this out for each and gather more feedback from users, I don't think there is a right or wrong way to do it.

maschad: Good point, I think since it was originally agreed upon to be this way let's move forward with this and then see what the user feedback is

packages/validator/src/validator.ts
```diff
@@ -159,6 +162,24 @@ export class Validator {
...
+            validatorIndex: validatorStore.getIndexOfPubkey(pubkeyHex),
+            pubKey: pubkeyHex,
```
nflaig: this log is quite verbose and will most of the time not be a one-liner (depending on screen size) which makes is hard to read. I am wondering if we really need to print out the full pubkey here or just the first and last 4 bytes using `prettyBytes` https://github.com/ChainSafe/lodestar/blob/df1ef97e9050a588b74f21bfae314d675caea288/packages/utils/src/format.ts#L7

we already print out the full pubkey on when loading from keystore so it might be fine to not do it here

image

packages/validator/src/validator.ts
```diff
@@ -159,6 +162,24 @@ export class Validator {
       this.clock.start(this.controller.signal);
       this.chainHeaderTracker.start(this.controller.signal);
 
+      const validatorKeys = this.validatorStore.getValidatorKeys();
+
+      for (const pubkeyHex in validatorKeys) {
+        if (!indicesService.validatorPubKeyExists(pubkeyHex)) {
+          this.logger.info("Validator exists in beacon chain", {
+            validatorIndex: validatorStore.getIndexOfPubkey(pubkeyHex),
```
nflaig: redundant to call this `validatorIndex` here, it should be pretty clear we are talking about the validator index due to the message itself

```suggestion
            index: validatorStore.getIndexOfPubkey(pubkeyHex),
```

packages/validator/src/validator.ts
maschad: Good idea with `prettyBytes` that will be easier to read. The log is actually printed out by the [indicesService](https://github.com/ChainSafe/lodestar/blob/unstable/packages/validator/src/services/indices.ts#L138) which is ran per epoch, so I moved it here to separate concerns.

packages/validator/src/validator.ts
maschad: I don't have a strong opinion on this, I can see a rationale behind utilizing either terminology, to be clear I was following the format used in [lighthouse's VC](https://github.com/ChainSafe/lodestar/issues/4785#issuecomment-1324702738)

nflaig: I would say adding a prefix to an index property only makes sense if there can be multiple types of indexes in the context of the log message. This is not the case here and why aren't we using the term `validatorPubkey` then as well. Would be in favor of removing the `validator` prefix to improve readablity.

https://github.com/ChainSafe/lodestar/issues/4785#issuecomment-1324702738 also log out the fee recipient first, which is questionable, I would probably not look at their log as a good example of how it should be done.

maschad: Understood. Good point.

packages/beacon-node/src/metrics/validatorMonitor.ts
```diff
@@ -287,6 +293,57 @@ export function createValidatorMonitor(
...
+
+      const statusCount = {
+        active: 0,
+        exited: 0,
+        pending: 0,
+        withdrawable: 0,
+        slashed: 0,
+      };
+
+      metrics.validatorMonitor.validatorStatuses.reset();
```
nflaig: reset should not be required since `set` is used below which will override current values. This also creates a potential race-condition where prometheus might scrape the metrics before they are set again

packages/validator/src/services/indices.ts
```diff
@@ -61,6 +36,11 @@ export class IndicesService {
     return this.pubkey2index.get(pubKey);
   }
 
+  /** Returns the validator pubkey for a given validator index */
```
nflaig: documentation differs from a what the function does

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -821,7 +821,16 @@ export function createLodestarMetrics(
         name: "validator_monitor_validators",
         help: "Count of validators that are specifically monitored by this beacon node",
       }),
-
+      validatorStatuses: register.gauge<"status">({
+        name: "validator_monitor_validators_status",
+        help: "Statuses of the various validators monitored by this beacon node",
+        labelNames: ["status"],
+      }),
+      validatorBalances: register.gauge<"index">({
```
nflaig: there is still an inconsistency between this metric and `validatorStatuses`

`validator_monitor_validators_status` vs `validator_monitor_validator_balances`, one of those has to be aligned with the other. Please check what naming conventions similar metrics use in validator monitor

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -821,7 +821,16 @@ export function createLodestarMetrics(
         name: "validator_monitor_validators",
         help: "Count of validators that are specifically monitored by this beacon node",
       }),
-
+      validatorStatuses: register.gauge<"status">({
```
nflaig, line 824: what's the difference between this metric and `validatorsConnected`? does it make sense to just add `status` label to the existing metric?

https://github.com/ChainSafe/lodestar/blob/df1ef97e9050a588b74f21bfae314d675caea288/packages/beacon-node/src/metrics/metrics/lodestar.ts#L820

Something else to consider, currently all validator montior metrics are collected in `scrapeMetrics` https://github.com/ChainSafe/lodestar/blob/df1ef97e9050a588b74f21bfae314d675caea288/packages/beacon-node/src/metrics/validatorMonitor.ts#L477

and those metrics are collected whenever Prometheus is scraping

https://github.com/ChainSafe/lodestar/blob/df1ef97e9050a588b74f21bfae314d675caea288/packages/beacon-node/src/metrics/metrics.ts#L28

I haven't looked into the implementation in detail but should definitely check if statuses and balances can also be collected and set there to align with other validator monitor metrics

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -821,7 +821,16 @@ export function createLodestarMetrics(
...
+        name: "validator_monitor_validators_status",
+        help: "Statuses of the various validators monitored by this beacon node",
+        labelNames: ["status"],
+      }),
+      validatorBalances: register.gauge<"index">({
+        name: "validator_monitor_validator_balances",
+        help: "Balance of validator",
```
nflaig:
```suggestion
        help: "Balance of validators",
```

packages/beacon-node/src/metrics/metrics/lodestar.ts
nflaig: what the difference between this metric and `prevEpochOnChainBalance`? Can't we just use the existing one to track the balance?

https://github.com/ChainSafe/lodestar/blob/df1ef97e9050a588b74f21bfae314d675caea288/packages/beacon-node/src/metrics/metrics/lodestar.ts#L832

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -821,7 +821,16 @@ export function createLodestarMetrics(
...
+        labelNames: ["index"],
```
nflaig: I think we need to be careful with introducing a lot of metrics that use `index` as label as validator index is in theory an unbounded value set (not really, due to limited ether supply). This might blow up the metrics if someone runs 5000+ validators.

packages/validator/src/metrics.ts
```diff
@@ -325,6 +325,12 @@ export function getMetrics(register: MetricsRegister, gitData: LodestarGitData)
       help: `Current health status of the beacon(s) the validator is connected too. ${renderEnumNumeric(BeaconHealth)}`,
     }),
 
+    validatorStatuses: register.gauge<{status: string}>({
+      name: "vc_validator_status_count",
+      help: "Current statuses of the validators.",
```
nflaig: super nit picky but other metrics don't end with a period, let's improve consistency

```suggestion
      help: "Current statuses of the validators",
```

packages/validator/src/validator.ts
```diff
@@ -38,6 +38,9 @@ export type ValidatorOptions = {
   valProposerConfig?: ValidatorProposerConfig;
 };
 
+// To assist with logging statuses, we only log the statuses that are not active_exiting or withdrawal_possible
+export type SimpleValidatorStatus = "pending" | "active" | "exited" | "withdrawn";
```
nflaig, line 42: are these supposed to differ from the ones used in validator monitor?

```ts
const statusCount = {
  active: 0,
  exited: 0,
  pending: 0,
  withdrawable: 0,
  slashed: 0,
};
```

packages/validator/src/services/validatorStore.ts
```diff
@@ -228,6 +232,10 @@ export class ValidatorStore {
     );
   }
 
+  getValidatorKeys(): PubkeyHex[] {
```
nflaig: this function seems to do the same as `votingPubkeys`

https://github.com/ChainSafe/lodestar/blob/df1ef97e9050a588b74f21bfae314d675caea288/packages/validator/src/services/validatorStore.ts#L312-L314

if we want to keep it I would at least improve naming consistency and use the term pubkey

```suggestion
  getValidatorPubkeys(): PubkeyHex[] {
```

packages/validator/src/validator.ts
```diff
@@ -159,6 +162,32 @@ export class Validator {
       this.clock.start(this.controller.signal);
       this.chainHeaderTracker.start(this.controller.signal);
 
+      const validatorKeys = this.validatorStore.getValidatorKeys();
+
+      for (const pubkeyHex in validatorKeys) {
+        let feeRecipient = "";
```
nflaig: setting an empty string here is not ideal, please check if it is possible to use a better default value

https://github.com/ChainSafe/lodestar/blob/df1ef97e9050a588b74f21bfae314d675caea288/packages/validator/src/services/validatorStore.ts#L149

https://github.com/ChainSafe/lodestar/blob/df1ef97e9050a588b74f21bfae314d675caea288/packages/validator/src/services/validatorStore.ts#L187

packages/validator/src/validator.ts
```diff
@@ -159,6 +162,32 @@ export class Validator {
...
+
+        try {
+          feeRecipient = validatorStore.getFeeRecipient(pubkeyHex);
```
nflaig, line 171: do we really expect that this throws an error? we are iterating over the validator keys which we got from `validatorStore` and `getFeeRecipient` only throw an error if pubkey does not exist

https://github.com/ChainSafe/lodestar/blob/df1ef97e9050a588b74f21bfae314d675caea288/packages/validator/src/services/validatorStore.ts#L184-L186

packages/validator/src/validator.ts
```diff
@@ -159,6 +162,32 @@ export class Validator {
...
+        } catch (e) {
+          this.logger.warn("Cannot retrieve Fee Recipient", {}, e as Error);
```
nflaig: generally error messages use lower case consistently

```suggestion
          this.logger.warn("Cannot retrieve fee recipient", {}, e as Error);
```

packages/validator/src/validator.ts
```diff
@@ -265,6 +294,40 @@ export class Validator {
...
+  private async fetchValidatorStatuses(pubkeysHex: string[], metrics: Metrics | null): Promise<void> {
+    const res = await this.api.beacon.getStateValidators("head", {
+      id: pubkeysHex,
+    });
+    ApiError.assert(res, "Can not fetch state validators from beacon node");
+
+    const allValidatorStatuses = new MapDef<SimpleValidatorStatus, number>(() => 0);
+
+    for (const validatorState of res.response.data) {
+      // Group all validators by status
```
nflaig: does this comment make sense inside the for-loop?

packages/validator/src/validator.ts
```diff
@@ -265,6 +294,40 @@ export class Validator {
...
+
+    // Retrieve the number of validators for each status
+    const statuses = Object.fromEntries(Array.from(allValidatorStatuses.entries()).filter((entry) => entry[1] > 0));
+
+    // The total number of validators
+    const total = pubkeysHex.length;
+
+    if (metrics) {
+      for (const [status, amount] of allValidatorStatuses) {
+        metrics.validatorStatuses.set({status}, amount);
```
nflaig: shouldn't this rather be called `count`?

```suggestion
      for (const [status, count] of allValidatorStatuses) {
        metrics.validatorStatuses.set({status}, count);
```

packages/beacon-node/src/metrics/validatorMonitor.ts
maschad: Good catch! That was there before I decided to increment and then set.

packages/beacon-node/src/metrics/metrics/lodestar.ts
maschad: we could just remove the `validatorsConnected` actually since it would require a greater refactor to collect statuses and balances in `scrapeMetrics`. I had left it just for backwards compatibility.

packages/validator/src/validator.ts
maschad: This was just for tests but good suggestion, I may have to add this default to the tests.

packages/validator/src/validator.ts
maschad: See https://github.com/ChainSafe/lodestar/pull/5239#discussion_r1147803452

packages/validator/src/validator.ts
maschad: it could be outside the loop as well, indifferent to whether it improves readability.

nflaig: doesn't matter that much but I would probably put it outside as the combined iterations "group all validators by status" and not a single iteration

packages/beacon-node/src/metrics/metrics/lodestar.ts
maschad: That's worth considering perhaps using `status` may actually be a better label but that may not be as useful for those running far less validators

packages/validator/src/validator.ts
maschad: Yes this is for the purposes of logging, the validator monitor needs greater exactitude for the metrics.

packages/validator/src/services/validatorStore.ts
maschad: Good catch, I will just remove the function instead.

packages/beacon-node/src/metrics/metrics/lodestar.ts
maschad: No difference, we can use the existing one, I am wondering if that is more intuitive name though.

nflaig: well since both of those use `index` label they are quite expensive, so keeping both does not make sense IMO and we should be careful with removing/renaming metrics. Would generally vote for keeping `prevEpochOnChainBalance` in that case

packages/beacon-node/src/metrics/metrics/lodestar.ts
maschad: The `validatorConnected` stat is useful for variety of other calculations and metrics so I don't think it's wise to remove it.

nflaig: yeah, I don't think it is a good idea to remove it, would be more about adding the label to that metric instead but should be fine to introduce a separate metric

dashboards/lodestar_validator_monitor.json
```diff
@@ -167,7 +167,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "exemplar": false,
-          "expr": "validator_monitor_validators_status",
+          "expr": "validator_monitor_validators_statuses",
```
nflaig: not native english speaker but double plural sounds odd to me

```suggestion
          "expr": "validator_monitor_validator_statuses",
```

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -837,12 +837,6 @@ export function createLodestarMetrics(
       }),
 
       // Validator Monitor Metrics (per-epoch summaries)
-      // Only track prevEpochOnChainBalance per index
-      prevEpochOnChainBalance: register.gauge<"index">({
```
nflaig: I'm really not sure about removing this if it does the same as the newly added metric

packages/validator/src/services/indices.ts
```diff
@@ -36,7 +36,7 @@ export class IndicesService {
     return this.pubkey2index.get(pubKey);
   }
 
-  /** Returns the validator pubkey for a given validator index */
+  /** Returns if a pubkey for a given validator index exists */
```
nflaig:
```suggestion
  /** Returns true if a pubkey for a given validator index exists */
```

packages/beacon-node/src/metrics/validatorMonitor.ts
```diff
@@ -532,8 +525,6 @@ export function createValidatorMonitor(
      * Should be called whenever Prometheus is scraping.
      */
     scrapeMetrics(slotClock) {
-      metrics.validatorMonitor.validatorsConnected.set(validators.size);
```
nflaig: where is `validatorsConnected` metric set now?

maschad:
Thanks for the feedback @nflaig , I think @wemeetagain or @dapplion should give a final review before merging.

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -821,7 +821,11 @@ export function createLodestarMetrics(
         name: "validator_monitor_validators",
         help: "Count of validators that are specifically monitored by this beacon node",
       }),
-
+      validatorStatuses: register.gauge<"status">({
+        name: "validator_monitor_validators_statuses",
```
nflaig:
```suggestion
        name: "validator_monitor_validator_statuses",
```

packages/beacon-node/src/api/impl/validator/index.ts
```diff
@@ -746,6 +750,8 @@ export function getValidatorApi({
         );
       });
 
+      metrics?.registerValidatorStatuses(currentEpoch, validatorStatuses);
```
dapplion, line 753: This is a very odd pattern. IMO we should not trigger status metrics in response to an API call. The status of validators materially change over time not in response to API calls. Would make much more sense to track status after an epoch transition like it's already done in the validator monitor. It's also not necessary for the validator client to be involved at all, since the beacon node already has all required information

dapplion (changes requested):

packages/beacon-node/src/api/impl/validator/index.ts
maschad: While that's true the `registerValidator` will be called every epoch. The same pattern has been applied here for other metrics invoked in this file in response to API calls, nevertheless I think the requirements have changed since the original issue so we perhaps don't need the `validatorStatuses` tracked by the [beacon node anymore](https://github.com/ChainSafe/lodestar/pull/5239/files#diff-61206b531bae1c15627a06074c8d264e9d833f3d8f9b12cb6fa65b5cb845fbdcR824) since they are in the [VC metrics](https://github.com/ChainSafe/lodestar/pull/5239/files#diff-6b01906a41fe4b3c383bda4797a4d02c2c81e9c86399fbb408600792e0ec88f0R328)

nflaig: > perhaps don't need the validatorStatuses tracked by the [beacon node anymore](https://github.com/ChainSafe/lodestar/pull/5239/files#diff-61206b531bae1c15627a06074c8d264e9d833f3d8f9b12cb6fa65b5cb845fbdcR824) since they are in the [VC metrics](https://github.com/ChainSafe/lodestar/pull/5239/files#diff-6b01906a41fe4b3c383bda4797a4d02c2c81e9c86399fbb408600792e0ec88f0R328)

We should see the BN and VC as two separate services and not make any assumptions that those will be deployed together.

packages/beacon-node/src/metrics/validatorMonitor.ts
```diff
@@ -287,6 +289,53 @@ export function createValidatorMonitor(
...
     },
 
+    registerValidatorStatuses(currentEpoch, statuses) {
+      // Prevent registering status for the same epoch twice. processEpoch() may be ran more than once for the same epoch.
+      if (currentEpoch <= lastRegisteredStatusEpoch) {
+        return;
+      }
+      lastRegisteredStatusEpoch = currentEpoch;
+
+      const statusCount = {
```
wemeetagain, line 299: For these metrics, I'd recommend just using the raw statuses and not using these aggregated ststuses.

maschad: The intention was to [mimic lighthouse's metrics](https://github.com/ChainSafe/lodestar/issues/5158#issuecomment-1452597916) given consumers may be already acquainted with this format.

wemeetagain: I see, sounds good

wemeetagain: I think the logic for selecting statuses needs to be updated.

There are some raw statuses that translate to multiple aggregated statuses.

Eg:
active_slashed = active and slashed
exited_slashed = exited and slashed
etc

packages/validator/src/validator.ts
```diff
@@ -159,6 +162,32 @@ export class Validator {
...
+          this.logger.info("Validator exists in beacon chain", {
+            index: validatorStore.getIndexOfPubkey(pubkeyHex),
+            pubKey: prettyBytes(pubkeyHex),
+            feeRecipient,
+          });
+        }
+      }
+
+      this.clock.runEveryEpoch(() =>
+        this.fetchValidatorStatuses(validatorKeys, metrics).catch((e) =>
```
nflaig, line 186: just a note on this, currently the `Validator statuses` log does not work as expected, we call `fetchValidatorIndices` in batches of 10

https://github.com/ChainSafe/lodestar/blob/887064d7f497adf123a3839ec4c71c323c24bbdd/packages/validator/src/services/indices.ts#L113-L114

The produced log on startup looks like this at the moment which is not correct

```
debug: HttpClient request routeId=getStateValidators
Apr-05 09:48:01.056[]                debug: HttpClient response routeId=getStateValidators
Apr-05 09:48:01.056[]                 info: Validator exists in beacon chain validatorIndex=461922, pubKey=0xa00421a5fad45881c14e2fa23f43aa93c21fd6610f96cc1e9be57c33178648730bdea843d4150d41cc610b857f4a2a59
Apr-05 09:48:01.057[]                 info: Validator exists in beacon chain validatorIndex=461923, pubKey=0xaed788793b98d6bed220cfa06ab357dc9c2bd1512f42f894c76deade02c404826f01758f0d90b6f5544d66a2788a27fe
Apr-05 09:48:01.057[]                 info: Validator exists in beacon chain validatorIndex=461924, pubKey=0xaa052e0d9603559793fea8515e0c6b6ff0e2410d0da08413ff84f1fe371b42f3e3381eb0e7506d423a870de58dc1b9fd
Apr-05 09:48:01.057[]                 info: Validator exists in beacon chain validatorIndex=461925, pubKey=0x8256e393a0332d0bc59f63b4490b3ed838240e6f32ac2409b95cced191542fb944e3866ae2d05d15e5122339c7281677
Apr-05 09:48:01.057[]                 info: Validator exists in beacon chain validatorIndex=461926, pubKey=0x8bfc76d96f53aa973acae88fd6b56a0c2b595e5ed3da2fb6b8353ace436fa0b0278b3ab0ee2c74ee2c93c03729f03958
Apr-05 09:48:01.057[]                 info: Validator exists in beacon chain validatorIndex=461927, pubKey=0xade304c8dda06bcf3cd70906e0f4f278577d530b997ed00159ce4c2dec4ea6586ddabfdbd2b470b5f16a1fdb0bf67c83
Apr-05 09:48:01.057[]                 info: Validator exists in beacon chain validatorIndex=461928, pubKey=0xb72e9beb76cbd5d37fdebff9cd65a2e019816f10b827decf36e149f8c32859c2837feb6edc066dda832253bd1a2d6624
Apr-05 09:48:01.057[]                 info: Validator exists in beacon chain validatorIndex=461929, pubKey=0xb52535102a3ea2d128bab7383c60c2c3e92e5153bbb80aa6b57ffe34051b6a82d5d960833369024f8c5e0392f3fb58fa
Apr-05 09:48:01.057[]                 info: Validator exists in beacon chain validatorIndex=461930, pubKey=0xa1107ab5b3d572a9198519d1a96ee4fc02846fe6f1328348e3e74f79ead408c2c7ed4fb5824cf4a3ae94ccbeb1e7c716
Apr-05 09:48:01.057[]                 info: Validator exists in beacon chain validatorIndex=461931, pubKey=0xb8581e1c4f6280977d5cfd84341d0f429963dfe0614b3efaaf01b299b9d07ffb766e92d94f9b33ff76c841ccf10d74b1
Apr-05 09:48:01.057[]                 info: Validator statuses active=10, total=10
Apr-05 09:48:01.057[]                debug: HttpClient request routeId=getStateValidators
Apr-05 09:48:01.059[]                debug: HttpClient response routeId=getStateValidators
Apr-05 09:48:01.059[]                 info: Validator exists in beacon chain validatorIndex=461932, pubKey=0xaf0bdd50360d92fdeba5123f7a053627ebdaaeb8b176898b1ea13bed307fdcad8081e33ab171416104bebd8827bdbf93
Apr-05 09:48:01.059[]                 info: Validator exists in beacon chain validatorIndex=461933, pubKey=0xb36e5520af098a0b163d493f4fe9b0e9406ed2eb490effc87d9bc8d1c11cdb6d1aed2b7c84db053b67005a1ce969b801
Apr-05 09:48:01.059[]                 info: Validator exists in beacon chain validatorIndex=461934, pubKey=0xa33c063d4d4cf5b7ae19fceeb0ff0a22abb11feb53cb195a0b4e6f53f3863216624774f5d95f9bfcb1cbf9f03fe52d66
Apr-05 09:48:01.059[]                 info: Validator exists in beacon chain validatorIndex=461935, pubKey=0xa54358dfa9975cd884c05034c39fa51a4b8288f82d5cc063e52ac0ca3e537f316aa68eb421718f4bbef85b63c7d83b3c
Apr-05 09:48:01.059[]                 info: Validator exists in beacon chain validatorIndex=461936, pubKey=0x94d5e85442021044a72a705fdb03a768c1080e34eb977ec55b729887ce09b62abb112ce8b65de83e99d538466cf4735c
Apr-05 09:48:01.059[]                 info: Validator exists in beacon chain validatorIndex=461937, pubKey=0xa9f88897d1d0cc422517faf6c4c2211186c1807c4920d08ace23e95fa580c074b16f98851ce9d8b70e230af32bba9e9b
Apr-05 09:48:01.059[]                 info: Validator exists in beacon chain validatorIndex=461938, pubKey=0x8806312a61d339cd607517114d1c8decec82afceb85c2e10e4c7a95590aed620a5ab150119e701e23fad03b2c2d583d3
Apr-05 09:48:01.059[]                 info: Validator exists in beacon chain validatorIndex=461939, pubKey=0xb4e2a7aa20ed22adc97c8c7164b113c0a3ade2915f4d1263f1b1278a2ecbad3d7383df0ca0dffc3f685e843305d87234
Apr-05 09:48:01.059[]                 info: Validator exists in beacon chain validatorIndex=461940, pubKey=0xb99a86e1adc5baf0be2f4f688a90c259e53e3057aa4b0e931496a25f4583f13b3cb83d4a01ae6e783c8476cd8bad80e1
Apr-05 09:48:01.059[]                 info: Validator exists in beacon chain validatorIndex=461941, pubKey=0xb341016be21bee2a500d2ab70a2ceb239ff30a938996b5eb7ebba4f99f18f92939458a57e5cad5685541bd80b8999250
Apr-05 09:48:01.059[]                 info: Validator statuses active=10, total=10
Apr-05 09:48:01.060[]                debug: HttpClient request routeId=getStateValidators
Apr-05 09:48:01.134[]                debug: HttpClient response routeId=getStateValidators
```


Now here we passing in all pubkeys to `getStateValidators` which does not work if you have 500+ keys

philknows:
Decided with @nflaig that these "nice to have" features are not required in the short term and this PR addresses way too many changes that would be ideal for one PR. We'll close this and address the improvements 1 by 1 and when it becomes necessary. PR here is also getting stale.

Issue #5247, opened 2023-03-08
v1.6.0-rc.0: Increased block process time


twoeths:
**Describe the bug**

This is from the beta mainnet running v1.6.0, Gossip Block Process time has been increased

<img width="1430" alt="Screen Shot 2023-03-08 at 16 04 56" src="https://user-images.githubusercontent.com/10568965/223669978-261f5f49-c0fc-462f-9113-9e32a0337d9f.png">

<img width="1296" alt="Screen Shot 2023-03-08 at 16 05 30" src="https://user-images.githubusercontent.com/10568965/223670091-122250dc-db71-4075-8a86-facef5e4fea7.png">

**Expected behavior**

Gossip Block Process time should be the same to `stable` group (v1.4.2)

twoeths:
Same metric in stable mainnet node

<img width="1298" alt="Screen Shot 2023-03-08 at 16 07 32" src="https://user-images.githubusercontent.com/10568965/223670552-370dbdc0-d83b-4e15-8028-8b5dd297706e.png">

<img width="1300" alt="Screen Shot 2023-03-08 at 16 08 09" src="https://user-images.githubusercontent.com/10568965/223670651-5ada39fd-b8c8-4f77-9afa-abd793fcdda3.png">

twoeths:
this is mainly due to the `notifyNewPayload` engine api call

<img width="1302" alt="Screen Shot 2023-03-08 at 17 15 30" src="https://user-images.githubusercontent.com/10568965/223686019-208446f5-9177-42d7-80ca-9edd2c672207.png">

in the stable mainnet node

<img width="1294" alt="Screen Shot 2023-03-08 at 17 16 29" src="https://user-images.githubusercontent.com/10568965/223686254-d1c8540c-3693-4416-aa8b-dff47ff8540d.png">

twoeths:
CPU time in v1.6.0 increased a lot

<img width="808" alt="Screen Shot 2023-03-08 at 18 17 06" src="https://user-images.githubusercontent.com/10568965/223699639-32b86a83-2800-4057-86eb-7fc93396c926.png">

GC Pause Time Rate tends keeps increasing

<img width="785" alt="Screen Shot 2023-03-08 at 18 20 41" src="https://user-images.githubusercontent.com/10568965/223700398-fdcf8519-7eaf-4cb7-89c1-eaa1bb95806e.png">

Execution engine notifyNewPayload body streaming time is high too

[...]

PR #5256, opened 2023-03-10
Limit preaggregating attestations

twoeths:
**Motivation**

As shown on a subscribe-all-subnets goerli node, preaggregating attestations takes 8% of cpu time, on mainnet it could be worse. I think some preaggregation are redundant because validators only get AggregatedAttestation at 2/3 of clock slot, including:
- Attestations with `slot < clockSlot` and `> clockSlot - 3`
- Attestations with `slot = clockSlot` but come to the pool at > 2/3 of slot

I noticed that on mainnet node, the attestation job wait time is 7s in average at some points so they are not useful to preaggregate anymore

**Description**

- Do not preaggregate attestations from old slots
- Do no preaggregate attestations in clock slot but come to the pool at > 2/3 of slot
- Same to SyncCommitteeMessage

part of #5247


dapplion (changes requested):
I'm not sure about the safety of this change in adverse conditions. This settings should be flags in case they need to be customized in the future

twoeths:
@dapplion are you worry about the `lowestPemissibleSlot` condition or `cutOffSecFromSlot` condition or both?
- I don't see it's reasonable to preaggregate attestations of old slots as aggregators only need to aggregate attestations of current slot
- For `cutOffSecFromSlot` condition we could accept `toleranceSec` which is 0.5s => do you mean a flag for this?

dapplion:
Only `lowestPemissibleSlot`, default to 1 (only same slot) but allow to increase to more

packages/beacon-node/src/chain/opPools/syncCommitteeMessagePool.ts
```diff
@@ -66,6 +73,11 @@ export class SyncCommitteeMessagePool {
       throw new OpPoolError({code: OpPoolErrorCode.SLOT_TOO_LOW, slot, lowestPermissibleSlot});
     }
 
+    // validator gets SyncCommitteeContribution at 2/3 of slot, it's no use to preaggregate later than that time
+    if (this.clock.secFromSlot(slot) > this.cutOffSecFromSlot) {
+      throw new OpPoolError({code: OpPoolErrorCode.LATE_MESSAGE, slot});
```
dapplion: How loud would be this error?

twoeths: we throw error here and gossipHandler would log error there.
since this is mainly for debugging purpose, not for end user so I changed log level to debug.

dapplion: Why not return an insert outcome? Then the caller can decide to maybe log a debug if result is not add

packages/beacon-node/src/chain/opPools/attestationPool.ts
```diff
@@ -130,12 +143,13 @@ export class AttestationPool {
   }
 
   /**
-   * Removes any attestations with a slot lower than `current_slot` and bars any future
-   * attestations with a slot lower than `current_slot - SLOTS_RETAINED`.
+   * Removes any attestations with a slot lower than `current_slot - SLOTS_RETAINED`.
+   * Not intested in attestations in old slots, we only preaggregate attestations for the current slot.
```
wemeetagain:
```suggestion
   * Removes any attestations with a slot lower than `current_slot - preaggregateSlotDistance`.
   * By default, not interested in attestations in old slots, we only preaggregate attestations for the current slot.
```

dapplion (dismissed):

dapplion (approved):

wemeetagain (approved):

wemeetagain:

Issue #4687, opened 2022-10-25
Sending 0 as committee_index in GET AttestationData endpoint


dB2510:
**Describe the bug**

While integrating lodestar to our Obol distributed validator cluster we found out that when calling [GET attestation data endpoint](https://ethereum.github.io/beacon-APIs/#/Validator/produceAttestationData) from beacon node, lodestar sends committee index as 0.
https://github.com/ChainSafe/lodestar/blob/bc20c21c6782e1917134576e577af3b292415c4d/packages/validator/src/services/attestation.ts#L91
Whereas other clients for ex, lighthouse and teku send committee index obtained at the time of getting attester duties endpoint. This results in timeout since we store attestation data with (slot, committee_index) as key.

**Expected behavior**

Lodestar should send actual `committee_index` as query parameter to GET attestation data endpoint rather than sending 0 obtained from [attester duties endpoint](https://ethereum.github.io/beacon-APIs/#/Validator/getAttesterDuties) so that it would be consistent with other clients and spec as well.

**Steps to Reproduce**

**Screenshots**

**Desktop (please complete the following information):**

- OS:  macOS Monterey
- Version:  v1.1.1
- Branch:  v1.1.1
- Commit hash:  90a92dd74f

wemeetagain:
We currently call the produceAttestationData once per slot for all connected validators with duties at that slot. We're calling with committee index 0 since the resulting attestation data isn't actually dependent on the committee index. We reuse the result for all connected validators with duties at that slot.

We probably wouldn't want to get rid of this optimization, but we could make the produceAttestationData call with a "real" committee index, eg: the committee index from the first duty in the list of duties for that slot.

Would that resolve yall's issue?

dB2510:
@wemeetagain yeah that would work for us 👍

philknows:
Closing via #4695

dB2510:
Hey @wemeetagain 👋 
Sorry for bumping up this issue. Actually the fix doesn't work for us. We call [GET attestation data endpoint](https://ethereum.github.io/beacon-APIs/#/Validator/produceAttestationData) endpoint for each validator public key like every other client and since AttestationData consists of beacon_block_root (LMD ghost vote) which is head block root, it can change between calls if we have multiple validators attesting in the same slot. This causes conflicts between lodestar and other VCs in the cluster as lodestar queries attestation_data once with 0 as committee_index.

[...]

PR #3923, opened 2022-04-14
Submit unaggregated attestations in one go

twoeths:
**Motivation**

There are a lot of same logs like below in hetzner nodes:
```
Apr-13 23:55:16.172[]                 ^[[32minfo^[[39m: Published attestations slot=2782176, index=8, count=1
Apr-13 23:55:16.182[]                 ^[[32minfo^[[39m: Published attestations slot=2782176, index=35, count=1
Apr-13 23:55:16.192[]                 ^[[32minfo^[[39m: Published attestations slot=2782176, index=27, count=1
Apr-13 23:55:16.202[]                 ^[[32minfo^[[39m: Published attestations slot=2782176, index=22, count=1
...
Apr-13 23:55:16.210[]                 ^[[32minfo^[[39m: Published attestations slot=2782176, index=44, count=1
Apr-13 23:55:16.219[]                 ^[[32minfo^[[39m: Published attestations slot=2782176, index=9, count=1
Apr-13 23:55:16.228[]                 ^[[32minfo^[[39m: Published attestations slot=2782176, index=29, count=1
```

+ We could submit all attestations of the same slot in one go, to save I/O interactions to the beacon node
+ Contabo nodes of 50 validators suffer from I/O lag, this should mitigate the issue

**Description**

Before
- Wait for 1/3 of slot or chain head
- For each committee index
  - Sign attestations
  - Submit attestations
  - Wait for 2/3 of slots
  - Produce and publish AggregateAndProofs
 
After
- Wait for 1/3 of slot or chain head
- For each committee index
  - Sign attestations
- Collect all signed attestations, submit all in 1 go
- Wait for 2/3 of slots
- For each committee index
  - Produce and publish AggregateAndProofs
  

part of #3922

PR #5258, opened 2023-03-10
Fix attestation compatibility issue in distributed validator cluster

### LINKED ISSUE #5103: Add CLI option to change validator behavior in distributed cluster (DVT)

There are compatibility issues running Lodestar VC in a distributed validator cluster due to an [optimization](https://github.com/ChainSafe/lodestar/issues/4687#issuecomment-1291021144).

The problem is that if there are multiple validators in a cluster and there are `N` attester duties in the same slot, then Lodestar queries the attestation data only **once** with the `committee_index` of the first duty and uses  that attestation data for all the validators by overriding the correct `committee_index` internally from the attester duty object for that validator. Other clients query `N` times to get the attestation data for each attester duty.

Distributed validator middleware clients such as [Charon](https://github.com/ObolNetwork/charon) calls [`/eth/v1/validator/attestation_data`](https://ethereum.github.io/beacon-APIs/#/Validator/produceAttestationData) endpoint for each validator public key and there is a slight chance that the `beacon_block_root` (LMD GHOST vote) changes between calls which would cause a conflict between attestations submitted by Lodestar and other VCs in the cluster, resulting in aggregation failure as explained in [this comment](https://github.com/ChainSafe/lodestar/issues/4687#issuecomment-1403716069) and further outlined [here](https://github.com/ObolNetwork/charon-distributed-validator-cluster/issues/23#issuecomment-1308428065).

The distributed validator middleware requires the actual `committee_index` retrieved from [`/eth/v1/validator/duties/attester/{epoch}`](https://ethereum.github.io/beacon-APIs/#/Validator/getAttesterDuties) to be sent to [`/eth/v1/validator/attestation_data`](https://ethereum.github.io/beacon-APIs/#/Validator/produceAttestationData) which means we need to disable the optimization. Since we only need to do this if the VC is part of a distributed cluster we could introduce a `--distributed` CLI option to achieve this.
This flag will also be useful when Lodestar implements the new DVT specific endpoints introduced in https://github.com/ethereum/beacon-APIs/pull/224.

Could completely remove the optimization which we probably don't want to do as it makes sense to do if you directly communicate with the beacon node as `committee_index` is just passed around. 

cc @dB2510

nflaig:
**Motivation**

Closes https://github.com/ChainSafe/lodestar/issues/5103 and https://github.com/ChainSafe/lodestar/issues/4687

**Description**

- allow to disable attestation service grouping optimization, attestation tasks will be executed per committee instead of just once for all committees
- add cli option to disable attestation service grouping optimization (`--disableAttestationGrouping`)
- add cli option to run as part of a distributed validator cluster (`--distributed`), this will also disable attestation grouping

**Open tasks**

- [x] add tests for  `runAttestationTasksPerCommittee`
- [x] remove TODOs

### DIFF
--- a/packages/cli/src/cmds/validator/handler.ts
+++ b/packages/cli/src/cmds/validator/handler.ts
@@ -51,6 +51,7 @@ export async function validatorHandler(args: IValidatorCliArgs & GlobalArgs): Pr
 
   const {version, commit} = getVersionData();
   logger.info("Lodestar", {network, version, commit});
+  if (args.distributed) logger.info("Client is configured to run as part of a distributed validator cluster");
   logger.info("Connecting to LevelDB database", {path: validatorPaths.validatorsDbDir});
 
   const dbPath = validatorPaths.validatorsDbDir;
@@ -167,6 +168,7 @@ export async function validatorHandler(args: IValidatorCliArgs & GlobalArgs): Pr
       scAfterBlockDelaySlotFraction: args.scAfterBlockDelaySlotFraction,
       disableAttestationGrouping: args.disableAttestationGrouping,
       valProposerConfig,
+      distributed: args.distributed,
     },
     metrics
   );
--- a/packages/cli/src/cmds/validator/handler.ts
+++ b/packages/cli/src/cmds/validator/handler.ts
@@ -165,6 +165,7 @@ export async function validatorHandler(args: IValidatorCliArgs & GlobalArgs): Pr
       doppelgangerProtectionEnabled,
       afterBlockDelaySlotFraction: args.afterBlockDelaySlotFraction,
       scAfterBlockDelaySlotFraction: args.scAfterBlockDelaySlotFraction,
+      disableAttestationGrouping: args.disableAttestationGrouping,
       valProposerConfig,
     },
     metrics
--- a/packages/cli/src/cmds/validator/options.ts
+++ b/packages/cli/src/cmds/validator/options.ts
@@ -36,6 +36,7 @@ export type IValidatorCliArgs = AccountValidatorArgs &
     graffiti: string;
     afterBlockDelaySlotFraction?: number;
     scAfterBlockDelaySlotFraction?: number;
+    disableAttestationGrouping?: boolean;
     suggestedFeeRecipient?: string;
     proposerSettingsFile?: string;
     strictFeeRecipientCheck?: boolean;
@@ -187,6 +188,14 @@ export const validatorOptions: CliCommandOptions<IValidatorCliArgs> = {
     description:
       "Delay before publishing SyncCommitteeSignature if block comes early, as a fraction of SECONDS_PER_SLOT (value is from 0 inclusive to 1 exclusive)",
     type: "number",
+  },
+
+  disableAttestationGrouping: {
+    hidden: true,
+    description:
+      "Disables attestation service grouping optimization, attestation tasks will be executed per committee instead of just once for all committees.",
+    default: false,
+    type: "boolean",
   },
 
   proposerSettingsFile: {
--- a/packages/cli/src/cmds/validator/options.ts
+++ b/packages/cli/src/cmds/validator/options.ts
@@ -53,6 +53,8 @@ export type IValidatorCliArgs = AccountValidatorArgs &
     "externalSigner.pubkeys"?: string[];
     "externalSigner.fetch"?: boolean;
 
+    distributed?: boolean;
+
     interopIndexes?: string;
     fromMnemonic?: string;
     mnemonicIndexes?: string;
@@ -284,6 +286,14 @@ export const validatorOptions: CliCommandOptions<IValidatorCliArgs> = {
     description: "Fetch then list of pubkeys to validate from an external signer",
     type: "boolean",
     group: "externalSignerUrl",
+  },
+
+  // Distributed validator
+
+  distributed: {
+    description: "Enables specific features required to run as part of a distributed validator cluster",
+    default: false,
+    type: "boolean",
   },
 
   // Metrics
--- a/packages/cli/src/cmds/validator/options.ts
+++ b/packages/cli/src/cmds/validator/options.ts
@@ -196,7 +196,6 @@ export const validatorOptions: CliCommandOptions<IValidatorCliArgs> = {
     hidden: true,
     description:
       "Disables attestation service grouping optimization, attestation tasks will be executed per committee instead of just once for all committees.",
-    default: false,
     type: "boolean",
   },
 
@@ -292,7 +291,6 @@ export const validatorOptions: CliCommandOptions<IValidatorCliArgs> = {
 
   distributed: {
     description: "Enables specific features required to run as part of a distributed validator cluster",
-    default: false,
     type: "boolean",
   },
 
--- a/packages/validator/src/services/attestation.ts
+++ b/packages/validator/src/services/attestation.ts
@@ -76,7 +76,9 @@ export class AttestationService {
       const dutiesByCommitteeIndex = groupAttDutiesByCommitteeIndex(duties);
       await Promise.all(
         Array.from(dutiesByCommitteeIndex.entries()).map(([index, duties]) =>
-          this.runAttestationTasksPerCommittee(duties, slot, index, signal)
+          this.runAttestationTasksPerCommittee(duties, slot, index, signal).catch((e) => {
+            this.logger.error("Error on attestation routine", {slot, index}, e);
+          })
         )
       );
     } else {
--- a/packages/validator/src/services/attestation.ts
+++ b/packages/validator/src/services/attestation.ts
@@ -12,7 +12,7 @@ import {groupAttDutiesByCommitteeIndex} from "./utils.js";
 import {ChainHeaderTracker} from "./chainHeaderTracker.js";
 import {ValidatorEventEmitter} from "./emitter.js";
 
-type AttestationServiceOpts = {
+export type AttestationServiceOpts = {
   afterBlockDelaySlotFraction?: number;
   disableAttestationGrouping?: boolean;
 };
--- a/packages/validator/src/services/attestation.ts
+++ b/packages/validator/src/services/attestation.ts
@@ -14,6 +14,7 @@ import {ValidatorEventEmitter} from "./emitter.js";
 
 type AttestationServiceOpts = {
   afterBlockDelaySlotFraction?: number;
+  disableAttestationGrouping?: boolean;
 };
 
 /**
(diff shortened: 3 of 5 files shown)


packages/validator/src/services/attestation.ts
```diff
@@ -66,9 +68,7 @@ export class AttestationService {
 
     // Beacon node's endpoint produceAttestationData return data is not dependant on committeeIndex.
     // Produce a single attestation for all committees, and clone mutate before signing
-    // Downstream tooling may require that produceAttestation is called with a 'real' committee index
-    // So we pick the first duty's committee index - see https://github.com/ChainSafe/lodestar/issues/4687
-    const attestationNoCommittee = await this.produceAttestation(duties[0].duty.committeeIndex, slot);
+    const attestationNoCommittee = await this.produceAttestation(slot);
```
nflaig: reverted changes done in https://github.com/ChainSafe/lodestar/pull/4695

packages/validator/src/services/distributedAttestation.ts
```diff
@@ -0,0 +1,63 @@
...
+import {sleep} from "@lodestar/utils";
+import {AttestationService, AttestationServiceArgs} from "./attestation.js";
+import {groupAttDutiesByCommitteeIndex} from "./utils.js";
+
+export class DistributedAttestationService extends AttestationService {
+  constructor(...args: AttestationServiceArgs) {
+    super(...args);
+  }
+
+  protected runAttestationTasks = async (slot: Slot, signal: AbortSignal): Promise<void> => {
```
nflaig: the logic implemented here will further deviate from normal attestation service if we wan to support [attestation committee aggregation selection](https://docs.google.com/document/d/1q9jOTPcYQa-3L8luRvQJ-M0eegtba4Nmon3dpO79TMk/mobilebasic)

packages/validator/src/services/distributedAttestation.ts
```diff
@@ -0,0 +1,63 @@
...
+    // First, wait until the `aggregation_production_instant` (2/3rds of the way though the slot)
+    await sleep(this.clock.msToSlot(slot + 2 / 3), signal);
+    this.metrics?.attesterStepCallProduceAggregate.observe(this.clock.secFromSlot(slot + 2 / 3));
+
+    // Then download, sign and publish a `SignedAggregateAndProof` for each
+    // validator that is elected to aggregate for this `slot` and
+    // `committeeIndex`.
+    await Promise.all(
+      Array.from(dutiesByCommitteeIndex.entries()).map(([index, duties]) => {
+        // const attestation = signedAttestations.find((a) => a.data.index === index);
```
nflaig, line 57: we could find the actual attestation here but to publish the aggregate this is not really needed, I just use the same approach we use in normal attestation service (mutating the attestation data)

dapplion:
Is this solution really simpler than just branching on the original class?

nflaig:
@dapplion I need to further refactor this so get a better picture, the 4 options I see are

1. implement subclass which overrides `runAttestationTasks` method
2. add if-else branch to `runAttestationTasks`
3. add 2nd method `runDistributedAttestationTasks`
4. remove the optimization implemented in https://github.com/ChainSafe/lodestar/pull/3917

IMO our attestation service is really easy to reason about at the moment and I really want to keep it like that. Completely isolating the logic seems like the best idea.

dapplion (changes requested):
IMO it's much more simple to have branching logic, see full implementation here https://github.com/ChainSafe/lodestar/compare/dapplion/attestation-service-no-grouping?expand=1

packages/cli/src/cmds/validator/handler.ts
```diff
@@ -51,6 +51,7 @@ export async function validatorHandler(args: IValidatorCliArgs & GlobalArgs): Pr
 
   const {version, commit} = getVersionData();
   logger.info("Lodestar", {network, version, commit});
+  if (args.distributed) logger.info("Client is configured to run as part of a distributed validator cluster");
```
nflaig, line 54: added this log at the beginning as I think it is quite important to highlight

packages/cli/src/cmds/validator/options.ts
```diff
@@ -189,6 +192,14 @@ export const validatorOptions: CliCommandOptions<IValidatorCliArgs> = {
     type: "number",
   },
 
+  disableAttestationGrouping: {
```
nflaig, line 195: added this as hidden cli option, I think it does not hurt and it allows us to disable the optimization independent of running in a distributed validator cluster

packages/validator/src/services/attestation.ts
```diff
@@ -64,41 +65,84 @@ export class AttestationService {
...
+  private async runAttestationTasksGrouped(
+    dutiesAll: AttDutyAndProof[],
+    slot: Slot,
+    signal: AbortSignal
+  ): Promise<void> {
     // Produce a single attestation for all committees, and clone mutate before signing
-    // Downstream tooling may require that produceAttestation is called with a 'real' committee index
-    // So we pick the first duty's committee index - see https://github.com/ChainSafe/lodestar/issues/4687
-    const attestationNoCommittee = await this.produceAttestation(duties[0].duty.committeeIndex, slot);
+    const attestationNoCommittee = await this.produceAttestation(0, slot);
```
nflaig, line 119: reverted changes done in https://github.com/ChainSafe/lodestar/pull/4695

packages/validator/src/services/attestation.ts
```diff
@@ -64,41 +65,84 @@ export class AttestationService {
...
-    await this.signAndPublishAttestations(slot, attestationNoCommittee, duties);
+    await this.signAndPublishAttestations(slot, attestationNoCommittee, dutiesAll);
 
     // Step 2. after all attestations are submitted, make an aggregate.
     // First, wait until the `aggregation_production_instant` (2/3rds of the way though the slot)
     await sleep(this.clock.msToSlot(slot + 2 / 3), signal);
     this.metrics?.attesterStepCallProduceAggregate.observe(this.clock.secFromSlot(slot + 2 / 3));
 
-    const dutiesByCommitteeIndex = groupAttDutiesByCommitteeIndex(this.dutiesService.getDutiesAtSlot(slot));
+    const dutiesByCommitteeIndex = groupAttDutiesByCommitteeIndex(dutiesAll);
```
nflaig, line 129: it should not be required to call `this.dutiesService.getDutiesAtSlot(slot)` again, see [comment](https://github.com/ChainSafe/lodestar/pull/3923/files#r1131374597)

packages/validator/src/services/attestation.ts
```diff
@@ -122,7 +166,10 @@ export class AttestationService {
 
     await Promise.all(
       duties.map(async ({duty}) => {
+        // TODO:  duty <> attestation consistency check required? slot and index are checked by lighthouse
```
nflaig: to be removed, really not sure if this is necessary, can this ever happen? @dapplion

packages/validator/src/services/attestation.ts
```diff
@@ -122,7 +166,10 @@ export class AttestationService {
...
+        // see https://github.com/sigp/lighthouse/blob/319cc61afeb1dbf3692e280dfa18e7b455542b16/validator_client/src/attestation_service.rs#L367
         const index = duty.committeeIndex;
+        // TODO: mutating is not required if attestation grouping is disabled as actual committee index is already set
```
nflaig: not a big deal here that we also mutate attestation if grouping is disabled as this would not be necessary since the committee index is already correct. What do you think @dapplion?

nflaig:
> IMO it's much more simple to have branching logic, see full implementation here

@dapplion thanks for the input, after further refactoring and reviewing I think you are right, I also like the branching solution more

Updated the code based on your initial implementation

packages/validator/src/services/attestation.ts
```diff
@@ -64,41 +65,84 @@ export class AttestationService {
...
+  ): Promise<void> {
+    // Produce attestation with actual committee index
+    const attestation = await this.produceAttestation(index, slot);
+
+    // Step 1. Sign `Attestation` for each validator. Then publish all `Attestations` in one go
+    await this.signAndPublishAttestations(slot, attestation, dutiesSameCommittee);
+
+    // Step 2. after all attestations are submitted, make an aggregate.
+    // First, wait until the `aggregation_production_instant` (2/3rds of the way though the slot)
+    await sleep(this.clock.msToSlot(slot + 2 / 3), signal);
```
nflaig, line 105: @dapplion this logic is duplicated in both `runAttestationTasksPerCommittee` and `runAttestationTasksGrouped`, was wondering if extracting this into a separate function makes sense

something like this

```ts
    /**
   * If a validator needs to publish an aggregate attestation,
   * wait until the`aggregation_production_instant` (2/3rds of the way though the slot)
   */
  private async waitTwoThirdsOfSlot(slot: Slot, signal: AbortSignal): Promise<void> {
    await sleep(this.clock.msToSlot(slot + 2 / 3), signal);
    this.metrics?.attesterStepCallProduceAggregate.observe(this.clock.secFromSlot(slot + 2 / 3));
  }
```

could then also use similar approach for 1/3 / block recevied wait time


```ts
 /**
   * A validator should create and broadcast the attestation to the associated attestation subnet when either
   *
   * (a) the validator has received a valid block from the expected block proposer for the assigned slot or
   *
   * (b) one-third of the slot has transpired (SECONDS_PER_SLOT / 3 seconds after the start of slot) -- whichever comes first.
   */
  private async waitOneThirdOfSlotOrReceivedBlock(slot: Slot, signal: AbortSignal): Promise<void> {
    await Promise.race([sleep(this.clock.msToSlot(slot + 1 / 3), signal), this.emitter.waitForBlockSlot(slot)]);
    this.metrics?.attesterStepCallProduceAttestation.observe(this.clock.secFromSlot(slot + 1 / 3));
  }
```

packages/validator/src/services/attestation.ts
```diff
@@ -64,41 +65,84 @@ export class AttestationService {
...
+    this.metrics?.attesterStepCallProduceAggregate.observe(this.clock.secFromSlot(slot + 2 / 3));
+
+    // Then download, sign and publish a `SignedAggregateAndProof` for each
```
nflaig, line 108: few comments are duplicated between `runAttestationTasksPerCommittee` and `runAttestationTasksGrouped` but I am not sure if further refactoring would improve clarity, and if someone looks at the code I would assume they will most of the time just look at one flow

packages/validator/src/services/attestation.ts
dapplion: :shrug: it's just 2 lines, no worries

packages/cli/src/cmds/validator/handler.ts
```diff
@@ -165,7 +166,9 @@ export async function validatorHandler(args: IValidatorCliArgs & GlobalArgs): Pr
       doppelgangerProtectionEnabled,
       afterBlockDelaySlotFraction: args.afterBlockDelaySlotFraction,
       scAfterBlockDelaySlotFraction: args.scAfterBlockDelaySlotFraction,
+      disableAttestationGrouping: args.disableAttestationGrouping,
```
dapplion, line 169: I introduced this option with the intention to replace `distributed`. Having both options do not make sense. Pick one or the other. I don't have strong feeling towards one or the other

nflaig: My reasoning behind this is the following

- `distributed` is specific to running the VC in a distributed validator cluster, this flag will enable further features later, specifically talking about [aggregate committee selections](https://docs.google.com/document/d/1q9jOTPcYQa-3L8luRvQJ-M0eegtba4Nmon3dpO79TMk/mobilebasic)
- `disableAttestationGrouping` is to disable the grouping optimization individually. This gives us more flexibility if there are other similar compatibility issues in the future.

If we just keep `disableAttestationGrouping` we would need another flag to enable aggregate committee selections to be fully dvt compliant. 

Therefore, we either keep both or just `distributed`.

nflaig: @dapplion what do you think about using `distributed` as cli option but still keeping `disableAttestationGrouping` as option in attestation service and then just setting that to `true` if `distributed=true`?

packages/cli/src/cmds/validator/options.ts
```diff
@@ -277,6 +288,14 @@ export const validatorOptions: CliCommandOptions<IValidatorCliArgs> = {
     group: "externalSignerUrl",
   },
 
+  // Distributed validator
+
+  distributed: {
+    description: "Enables specific features required to run as part of a distributed validator cluster",
+    default: false,
```
dapplion: No need to set the default value to the primitive's default value, please remove

packages/validator/src/services/attestation.ts
dapplion: Consistency is checked in the ValidatorStore

packages/validator/src/services/attestation.ts
dapplion: Yeah but otherwise complicates the logic more to save an object allocation, which is whatever

dapplion (commented):
Looks good to merge! Just a few nits

packages/cli/src/cmds/validator/options.ts
nflaig: Reasoning for setting this was to improve clarity on what the default is and don't we show this in the cli docs?

nflaig: ok, seems to be more consistent with other `boolean` values, I removed it

dapplion (dismissed):

twoeths (dismissed):
lgtm

nflaig:
This should not be merged yet, waiting for confirmation from Obol that this actually fixes the DVT compatibility issues.

philknows (dismissed):

packages/validator/src/services/attestation.ts
```diff
@@ -64,41 +65,84 @@ export class AttestationService {
...
+      // middleware clients such as Charon (https://github.com/ObolNetwork/charon) expect the actual committee index
+      // to be sent to produceAttestationData endpoint. This is required because the middleware client itself
+      // calls produceAttestationData on the beacon node for each validator and there is a slight chance that
+      // the `beacon_block_root` (LMD GHOST vote) changes between calls which would cause a conflict between
+      // attestations submitted by Lodestar and other VCs in the cluster, resulting in aggregation failure.
+      // See https://github.com/ChainSafe/lodestar/issues/5103 for further details and references.
+      const dutiesByCommitteeIndex = groupAttDutiesByCommitteeIndex(duties);
+      await Promise.all(
+        Array.from(dutiesByCommitteeIndex.entries()).map(([index, duties]) =>
+          this.runAttestationTasksPerCommittee(duties, slot, index, signal)
```
nflaig: @dapplion was just wondering if we do want to catch errors thrown by `runAttestationTasksPerCommittee` as there are multiple function calls (e.g. `produceAttestation`) which could throw an error.

I would assume as the code currently works we would have an unhandled promise rejection if more than one `runAttestationTasksPerCommittee` throws an error, although it would still finish all tasks per committee.

packages/validator/src/services/attestation.ts
```diff
@@ -64,41 +65,86 @@ export class AttestationService {
...
+      // to be sent to produceAttestationData endpoint. This is required because the middleware client itself
+      // calls produceAttestationData on the beacon node for each validator and there is a slight chance that
+      // the `beacon_block_root` (LMD GHOST vote) changes between calls which would cause a conflict between
+      // attestations submitted by Lodestar and other VCs in the cluster, resulting in aggregation failure.
+      // See https://github.com/ChainSafe/lodestar/issues/5103 for further details and references.
+      const dutiesByCommitteeIndex = groupAttDutiesByCommitteeIndex(duties);
+      await Promise.all(
+        Array.from(dutiesByCommitteeIndex.entries()).map(([index, duties]) =>
+          this.runAttestationTasksPerCommittee(duties, slot, index, signal).catch((e) => {
+            this.logger.error("Error running attestation tasks", {slot, index}, e);
```
nflaig, line 80: errors thrown by `runAttestationTasksPerCommittee` needs to be caught to avoid unhandled promise rejections and properly log all errors if multiple runs fails

this is also inline with the previous implementation

https://github.com/ChainSafe/lodestar/blob/106a05a281fa63b0d1b26571a573c7d46545626b/packages/validator/src/services/attestation.ts#L65-L67

packages/validator/src/services/attestation.ts
```diff
@@ -64,41 +65,86 @@ export class AttestationService {
...
+      // to be sent to produceAttestationData endpoint. This is required because the middleware client itself
+      // calls produceAttestationData on the beacon node for each validator and there is a slight chance that
+      // the `beacon_block_root` (LMD GHOST vote) changes between calls which would cause a conflict between
+      // attestations submitted by Lodestar and other VCs in the cluster, resulting in aggregation failure.
+      // See https://github.com/ChainSafe/lodestar/issues/5103 for further details and references.
+      const dutiesByCommitteeIndex = groupAttDutiesByCommitteeIndex(duties);
+      await Promise.all(
+        Array.from(dutiesByCommitteeIndex.entries()).map(([index, duties]) =>
+          this.runAttestationTasksPerCommittee(duties, slot, index, signal).catch((e) => {
+            this.logger.error("Error on attestation routine", {slot, index}, e);
```
nflaig, line 80: errors thrown by `runAttestationTasksPerCommittee` need to be caught to avoid unhandled promise rejections and properly log all errors if multiple runs fails

this is also inline with the previous implementation

https://github.com/ChainSafe/lodestar/blob/106a05a281fa63b0d1b26571a573c7d46545626b/packages/validator/src/services/attestation.ts#L65-L67

nflaig:
> This should not be merged yet, waiting for confirmation from Obol that this actually fixes the DVT compatibility issues.

Testing was successful 👍

dapplion (approved):

wemeetagain:

PR #5259, opened 2023-03-11
Always add gossip attestations to forkchoice

twoeths:
**Motivation**

- Gossip handler: Right now we do early return if beacon node does not have any validators as aggregator on specified subnets, i.e we didn't add gossip attestations to forkchoice in that case

**Description**

- We should always add attestations to forkchoice
- The `attnetsService.shouldProcess` should only decide if we add attestation to our pool to prepare for aggregator's duty later

### DIFF
--- a/packages/beacon-node/src/network/processor/gossipHandlers.ts
+++ b/packages/beacon-node/src/network/processor/gossipHandlers.ts
@@ -255,15 +255,13 @@ export function getGossipHandlers(modules: ValidatorFnsModules, options: GossipH
       const {indexedAttestation} = validationResult;
       metrics?.registerGossipUnaggregatedAttestation(seenTimestampSec, indexedAttestation);
 
-      // Node may be subscribe to extra subnets (long-lived random subnets). For those, validate the messages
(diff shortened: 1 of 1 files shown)


dapplion (changes requested):
Are you sure this is required? What problem or correctness issue is it trying to fix?

twoeths:
@dapplion right now we only pass attestations to fork_choice if we have an aggregator on that subnet/slot, I don't see any requirements for that.

I checked from the spec: `Run ``on_attestation`` upon receiving a new ``attestation`` from either within a block or directly on the wire.` => we always have to do that as I understand

https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/fork-choice.md#on_attestation

also this was instructed from Cayman as I remember, could you double check @wemeetagain ?

dapplion:
Right, if we have already done the work of verifying the attestation, might as well include it in the fork-choice

wemeetagain (approved):

dapplion (approved):

wemeetagain:

PR #5263, opened 2023-03-13
Process gossip block immediately

twoeths:
**Motivation**

- Gossip Block needs to check bls and regen module like other gossip objects before it's processed, in the worse case the delay could be > 2s like the below test mainnet node

Screenshot 2023-03-13 at 14 20 08

- Gossip block's signature is verified on main thread so network processor should not check for bls thread pool before processing it

**Description**

- Ensure to process gossip block immediately by bypassing bls thread pool and regen check for `beacon_block*` topic

part of #5247

### DIFF
--- a/packages/beacon-node/src/network/processor/index.ts
+++ b/packages/beacon-node/src/network/processor/index.ts
@@ -20,24 +20,28 @@ export type NetworkProcessorOpts = GossipHandlerOpts & {
   maxGossipTopicConcurrency?: number;
 };
 
+type WorkOpts = {
+  bypassQueue?: boolean;
+};
+
 /**
  * True if we want to process gossip object immediately, false if we check for bls and regen
  * in order to process the gossip object.
  */
-const executeGossipWorkOrderObj: Record<GossipType, boolean> = {
+const executeGossipWorkOrderObj: Record<GossipType, WorkOpts> = {
   // gossip block verify signatures on main thread, hence we want to bypass the bls check
-  [GossipType.beacon_block]: true,
-  [GossipType.beacon_block_and_blobs_sidecar]: true,
-  [GossipType.beacon_aggregate_and_proof]: false,
-  [GossipType.beacon_attestation]: false,
-  [GossipType.voluntary_exit]: false,
-  [GossipType.proposer_slashing]: false,
-  [GossipType.attester_slashing]: false,
-  [GossipType.sync_committee_contribution_and_proof]: false,
-  [GossipType.sync_committee]: false,
-  [GossipType.light_client_finality_update]: false,
-  [GossipType.light_client_optimistic_update]: false,
-  [GossipType.bls_to_execution_change]: false,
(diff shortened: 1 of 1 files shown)


packages/beacon-node/src/network/processor/index.ts
```diff
@@ -20,19 +20,24 @@ export type NetworkProcessorOpts = GossipHandlerOpts & {
   maxGossipTopicConcurrency?: number;
 };
 
-const executeGossipWorkOrderObj: Record<GossipType, true> = {
+/**
+ * True if we want to process gossip object immediately, false if we check for bls and regen
+ * in order to process the gossip object.
+ */
+const executeGossipWorkOrderObj: Record<GossipType, boolean> = {
```
dapplion: Make this an Record of opts for better readibility:
```suggestion
type WorkOpts = {
  bypassQueue?: boolean
}

const executeGossipWorkOrderObj: Record<GossipType, WorkOpts> = {
  [GossipType.beacon_block]: {bypassQueue: true}
```

packages/beacon-node/src/network/processor/index.ts
```diff
@@ -123,13 +132,14 @@ export class NetworkProcessor {
...
-      if (!this.chain.blsThreadPoolCanAcceptWork() || !this.chain.regenCanAcceptWork()) {
-        this.metrics?.networkProcessor.canNotAcceptWork.inc();
-        break;
-      }
-
       for (const topic of executeGossipWorkOrder) {
+        // Check canAcceptWork before calling queue.next() since it consumes the items
+        // beacon block is guaranteed to be processed immedately
+        const bypassQueue = executeGossipWorkOrderObj[topic]?.bypassQueue ?? false;
+        if (!bypassQueue && (!this.chain.blsThreadPoolCanAcceptWork() || !this.chain.regenCanAcceptWork())) {
```
dapplion: This will cause to have to call canAcceptWork many more times. Why not:

```ts
const canAcceptWork = this.chain.blsThreadPoolCanAcceptWork() && this.chain.regenCanAcceptWork()

for (const topic of executeGossipWorkOrder) {
        // Check canAcceptWork before calling queue.next() since it consumes the items
        // beacon block is guaranteed to be processed immedately
        if ( !canAcceptWork && !executeGossipWorkOrderObj[topic].bypassQueue) {
           break job_loop;
        }
```

twoeths: good point 👍 , I modified accordingly

twoeths:
convert to draft, new gossipsub queue work will be reverted once we merge stable v1.6.0 to unstable

dapplion (dismissed):

twoeths:
Merged to [tuyen/new_gossip_queues](https://github.com/ChainSafe/lodestar/tree/tuyen/new_gossip_queues) as pre-approved by @dapplion

wemeetagain:

PR #5275, opened 2023-03-16
fix: Update VC logging for scenario where inactive validators

### LINKED ISSUE #5265: Log `Validator pubkey not found` as a warning
**Describe the bug**

In situations where a user's validator client may be running inactive validators, they may get a `404` error when trying to [get the state validator index](https://github.com/ChainSafe/lodestar/blob/15a9035ad030aaee303f13a5bbc0b409cc63385f/packages/beacon-node/src/api/impl/beacon/state/index.ts#L67)  this error case should perhaps be logged as a warning.

See [discussion here](https://discord.com/channels/593655374469660673/743858262864167062/1084527788435976323)

maschad:
Closes https://github.com/ChainSafe/lodestar/issues/5265

### DIFF
--- a/packages/validator/src/services/indices.ts
+++ b/packages/validator/src/services/indices.ts
@@ -122,6 +122,9 @@ export class IndicesService {
 
   private async fetchValidatorIndices(pubkeysHex: string[]): Promise<ValidatorIndex[]> {
     const res = await this.api.beacon.getStateValidators("head", {id: pubkeysHex});
+    if (res.status === 404) {
+      this.logger.warn("Validator pubkey not found in state");
+    }
     ApiError.assert(res, "Can not fetch state validators from beacon node");
 
     const newIndices = [];


packages/validator/src/services/indices.ts
```diff
@@ -122,6 +122,9 @@ export class IndicesService {
 
   private async fetchValidatorIndices(pubkeysHex: string[]): Promise<ValidatorIndex[]> {
     const res = await this.api.beacon.getStateValidators("head", {id: pubkeysHex});
+    if (res.status === 404) {
+      this.logger.warn("Validator pubkey not found in state");
```
nflaig, line 126: not sure what this would achieve, now we are logging a warning and error which makes it even more noisy

nflaig (changes requested):
This PR does not address the issue discussed [in discord](https://discord.com/channels/593655374469660673/743858262864167062/10845277884359763230), please have another look at what was discussed, it is not about the Lodestar validator client. Need to rephrase the issue https://github.com/ChainSafe/lodestar/issues/5265 as well probably to make it clearer this error is on the beacon node.

PR #5268, opened 2023-03-14
Fix array query string parsing

nflaig:
**Motivation**

The [query string parser](https://www.npmjs.com/package/querystring) we use at the moment does not support an array as comma-separated values and on top of that the package itself is deprecated and last release was 2 years ago, [querystring node API](https://nodejs.org/api/querystring.html) is considered legacy as well and does not support all parsing requirements we need.

The following request it not correctly parsed at the moment, it result in just one value

```sh
curl 'http://localhost:9596/eth/v1/beacon/states/head/validators?id=0xb89bebc699769726a318c8e9971bd3171297c61aea4a6578a7a4f94b547dcba5bac16a89108b6b6a1fe3695d1a874a0b,0xa99a76ed7796f7be22d5b7e85deeb7c5677e88e511e0b337618f8c4eb61349b4bf2d153f649f7b53359fe8b94a38e44c'
```

To be OpenAPI spec compliant we need to support this query string format, all [other 4 CLs support this format](https://discord.com/channels/593655374469660673/1062235362509996112/1085260177210605679) as well.

The issue was reported by jcrtp from rocketpool, see [discord thread](https://discord.com/channels/593655374469660673/1062235362509996112/1084551528381165568).

**Description**

Updates fastify [querystringParser](https://www.fastify.io/docs/latest/Reference/Server/#querystringparser) to use [qs](https://www.npmjs.com/package/qs) instead of [querystring](https://www.npmjs.com/package/querystring).

`qs` seems to be really popular and is actively maintained. The code looks clean as well and has a good test suite.

We also already use `qs` package in several other places (e.g. [here](https://github.com/ChainSafe/lodestar/blob/a198b7ed1a6328d277ac4dca5483f979b8999b25/packages/api/src/utils/client/format.ts#L8)) in the monorepo.

PR #5280, opened 2023-03-17
Update query parser options

nflaig:
**Motivation**

Follow up PR to https://github.com/ChainSafe/lodestar/pull/5268. Updates the parser options to make array query string parser more consistent/predictable.

**Description**

Update query parser options, drop support for array query strings like `id[0]=1&id[1]=2&id[2]=3` as those are not required to be OpenAPI spec compliant and results are inconsistent, see https://github.com/ljharb/qs/issues/331. The schema validation will catch this and throw an error as parsed query string results in an object.

**Parsing of different formats**

Here are just a few examples how array query strings will be parsed, most example query strings are from [OpenAPI Parameter Serialization](https://swagger.io/docs/specification/serialization/)

```ts
"id[0]=3&id[1]=4&id[3]=5" => {"id":{"0":"3","1":"4","3":"5"}} // results in "should be array" error
"id=3&id=4&id=5"          => {"id":["3","4","5"]} // supported format
"id=3,4,5"                => {"id":["3","4","5"]} // supported format
"id=3|4|5"                => {"id":"3|4|5"} // results in a string (not supported)
"id=3%204%205"            => {"id":"3 4 5"} // results in a string (not supported)
"id=3%2C4%2C5"            => {"id":"3,4,5"} // results in a string (not supported)
"id"                      => {"id":""} // empty string here is expected
```

### DIFF
--- a/packages/beacon-node/src/api/rest/base.ts
+++ b/packages/beacon-node/src/api/rest/base.ts
@@ -45,12 +45,12 @@ export class RestApiServer {
       ajv: {customOptions: {coerceTypes: "array"}},
       querystringParser: (str) =>
         qs.parse(str, {
-          // defaults to 20 but Beacon API spec allows max items of 30
-          arrayLimit: 30,
-          // array as comma-separated values must be supported to be OpenAPI spec compliant
+          // Array as comma-separated values must be supported to be OpenAPI spec compliant
           comma: true,
-          // default limit of 1000 seems unnecessarily high, let's reduce it a bit
-          parameterLimit: 100,
+          // Drop support for array query strings like `id[0]=1&id[1]=2&id[2]=3` as those are not required to
+          // be OpenAPI spec compliant and results are inconsistent, see https://github.com/ljharb/qs/issues/331.
+          // The schema validation will catch this and throw an error as parsed query string results in an object.
+          parseArrays: false,
         }),
       bodyLimit: opts.bodyLimit,
     });
--- a/packages/api/test/utils/utils.ts
+++ b/packages/api/test/utils/utils.ts
@@ -10,7 +10,7 @@ export function getTestServer(): {baseUrl: string; server: FastifyInstance} {
 
   const server = fastify({
     ajv: {customOptions: {coerceTypes: "array"}},
-    querystringParser: (str) => qs.parse(str, {comma: true}),
+    querystringParser: (str) => qs.parse(str, {comma: true, parseArrays: false}),
   });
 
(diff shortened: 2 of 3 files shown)


packages/beacon-node/src/api/rest/base.ts
```diff
@@ -45,12 +45,12 @@ export class RestApiServer {
       ajv: {customOptions: {coerceTypes: "array"}},
       querystringParser: (str) =>
         qs.parse(str, {
-          // defaults to 20 but Beacon API spec allows max items of 30
-          arrayLimit: 30,
```
nflaig, line 49: removed `arrayLimit` as it does not acutally work as expected, see https://github.com/ljharb/qs/issues/294, in fact, completely removed array parsing (values like `id[0]=1&id[1]=2&id[2]=3`) to make the parsing more predictable. This kind of array query strings were also not supported before updating the query parser in https://github.com/ChainSafe/lodestar/pull/5268, so I would assume nobody relies on that functionality at the moment.

packages/beacon-node/src/api/rest/base.ts
```diff
@@ -45,12 +45,12 @@ export class RestApiServer {
...
-          // array as comma-separated values must be supported to be OpenAPI spec compliant
+          // Array as comma-separated values must be supported to be OpenAPI spec compliant
           comma: true,
-          // default limit of 1000 seems unnecessarily high, let's reduce it a bit
-          parameterLimit: 100,
```
nflaig, line 53: Parameter limit silently cuts off exceeding values which is not ideal, using the default value of `1000` here should be fine and is only relevant if someone sends array queries like this `id=3&id=4&id=5`.

The parameter limit is anyhow not what should be used to limit DoS vectors, instead we should rely on maximum header size of the http server which is 16KB.

Some downstream tooling such as rocketpool might require this limit to be increased as they send requests like this `eth/v1/beacon/states/head/validators?id=<1000 validator pubkeys, comma-separated>` which requires much higher max header size.

For now, this configurable using the `--max-http-header-size` node cli option, see [http.maxHeaderSize](https://nodejs.org/api/http.html#httpmaxheadersize)

dapplion (approved):
LGTM

wemeetagain:

packages/beacon-node/src/api/rest/base.ts
nflaig: Lodestar now also has a CLI flag `--rest.headerLimit` to configure this
- https://github.com/ChainSafe/lodestar/pull/5594

PR #5284, opened 2023-03-20
Only preaggregate attestations if there are connected aggregators

twoeths:
**Motivation**

- Preaggregating attestations could take up to 8% of cpu time while a lot of them are redundant, see https://github.com/ChainSafe/lodestar/issues/5247#issuecomment-1463396305. This PR is to preaggregate attestations if there are a connected aggregator at specified subnet/slot
- An effort to mitigate/fix the issue we have with new gossip queues

**Description**

- Right now we track committee subscriptions up until a slot, and we process **all** attestations up until that slot/subnet while an aggregator only has duty at a specified subnet/slot.
- Notice that `prepareBeaconCommitteeSubnet()` api is called for next epoch at `vc` side, so this causes a lot of redundancy to preaggregate attestations
- Instead of that track aggregator at slot/subnet and based on that to process attestations
- Also at `submitPoolAtestations` api, only add attestation to pool if necessary

part of #5247

### DIFF
--- a/packages/beacon-node/src/api/impl/beacon/pool/index.ts
+++ b/packages/beacon-node/src/api/impl/beacon/pool/index.ts
@@ -64,12 +64,13 @@ export function getBeaconPoolApi({
               slot,
               beaconBlockRoot
             );
-            const sentPeers = await network.gossip.publishBeaconAttestation(attestation, subnet);
-            metrics?.submitUnaggregatedAttestation(seenTimestampSec, indexedAttestation, subnet, sentPeers);
+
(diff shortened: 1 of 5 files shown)


twoeths:
Attestation Pool Size is reduced 95%

Screenshot 2023-03-20 at 17 03 19

Similar to Preaggregation per slot
Screenshot 2023-03-20 at 17 09 02

dapplion (approved):
LGTM

wemeetagain:

PR #5285, opened 2023-03-20
Log expected api errors as warning instead of error

### LINKED ISSUE #5265: Log `Validator pubkey not found` as a warning

nflaig:
**Motivation**

Expected API errors (`ApiError`) should be logged as warnings instead of errors to not unnecessarily scare users. Printing out the stacktrace does not help here either, it just creates noise.

**Description**

Check if error is `ApiError` and just log it as `warn` and only print out the `err.message`.

Unexpected errors (non `ApiError`) will still be logged as `error` including the stacktrace.

Closes https://github.com/ChainSafe/lodestar/issues/5265, related [discussion](https://discord.com/channels/593655374469660673/743858262864167062/1084567242437558292)

### DIFF
--- a/packages/beacon-node/src/api/rest/base.ts
+++ b/packages/beacon-node/src/api/rest/base.ts
@@ -97,7 +97,12 @@ export class RestApiServer {
       if (err instanceof ErrorAborted || err instanceof NodeIsSyncing) return;
 
       const {operationId} = res.context.config as RouteConfig;
-      this.logger.error(`Req ${req.id} ${operationId} error`, {}, err);
+
+      if (err instanceof ApiError && err.statusCode < 500) {
+        this.logger.warn(`Req ${req.id} ${operationId} failed`, {reason: err.message});
+      } else {
+        this.logger.error(`Req ${req.id} ${operationId} error`, {}, err);
+      }
       metrics?.errors.inc({operationId});
     });
 
--- a/packages/beacon-node/src/api/rest/base.ts
+++ b/packages/beacon-node/src/api/rest/base.ts
@@ -98,7 +98,7 @@ export class RestApiServer {
 
       const {operationId} = res.context.config as RouteConfig;
 
-      if (err instanceof ApiError && err.statusCode < 500) {
+      if (err instanceof ApiError) {
         this.logger.warn(`Req ${req.id} ${operationId} failed`, {reason: err.message});
       } else {
         this.logger.error(`Req ${req.id} ${operationId} error`, {}, err);


packages/beacon-node/src/api/rest/base.ts
```diff
@@ -97,7 +97,12 @@ export class RestApiServer {
       if (err instanceof ErrorAborted || err instanceof NodeIsSyncing) return;
 
       const {operationId} = res.context.config as RouteConfig;
-      this.logger.error(`Req ${req.id} ${operationId} error`, {}, err);
+
+      if (err instanceof ApiError && err.statusCode < 500) {
```
maschad: Are you sure that all **400 codes** should not be logged as errors? I would think particular instances such as a [404 for resolving a block ID](https://github.com/ChainSafe/lodestar/blob/unstable/packages/beacon-node/src/api/impl/beacon/blocks/utils.ts#L34) it should be thrown as an error, otherwise we can't fetch a block root for example.

Also [if we fail to provide a set of sync committee duties for a particular epoch](https://github.com/ChainSafe/lodestar/blob/unstable/packages/beacon-node/src/api/impl/validator/index.ts#L479) , a `400` error is thrown. Perhaps what could be considered is optionally returning the stack trace as suggested in the [spec](https://github.com/ethereum/beacon-APIs/blob/master/types/http.yaml#L141)

maschad (changes requested):
I am not sure all non-500 error codes are just warnings

packages/beacon-node/src/api/rest/base.ts
nflaig: > Are you sure that all 400 codes should not be logged as errors?

pretty much, in fact, could even remove the `err.statusCode < 500` in my opinion, there is anyway just one error at the moment were this would make a difference https://github.com/ChainSafe/lodestar/blob/cd2ec2ee9878a12c09742c76fb2681a4e7d551e2/packages/beacon-node/src/api/impl/validator/index.ts#L356

> [404 for resolving a block ID](https://github.com/ChainSafe/lodestar/blob/unstable/packages/beacon-node/src/api/impl/beacon/blocks/utils.ts#L34) it should be thrown as an error, otherwise we can't fetch a block root for example.

I am not really sure how this change affects what you mentioned, all that changed now is that we log it as a warning with less verbosity (no stacktrace) instead of an error. The API response is still the same.
The two examples you linked just further enforce the change I want to introduce, there is no reason at all we should log a stacktrace when those errors happen.


> Perhaps what could be considered is optionally returning the stack trace as suggested in the [spec](https://github.com/ethereum/beacon-APIs/blob/master/types/http.yaml#L141)

For some unexpected errors it could make sense to also include the stacktrace in the HTTP error response but might as well just look at the logs instead. Maybe someone else in the team has an opinion on this, but I would not implement this unless there is a specific use case for it.

nflaig:
> I am not sure all non-500 error codes are just warnings

regarding this, if there are non-500 error codes which are not considered a warning but something critical then the error code must be changed

packages/beacon-node/src/api/rest/base.ts
nflaig: consider logging all api errors as warning

```suggestion
      if (err instanceof ApiError) {
```

wemeetagain: I'm in favor of this change

packages/beacon-node/src/api/rest/base.ts
```diff
@@ -97,7 +97,12 @@ export class RestApiServer {
       if (err instanceof ErrorAborted || err instanceof NodeIsSyncing) return;
 
       const {operationId} = res.context.config as RouteConfig;
-      this.logger.error(`Req ${req.id} ${operationId} error`, {}, err);
+
+      if (err instanceof ApiError) {
+        this.logger.warn(`Req ${req.id} ${operationId} failed`, {reason: err.message});
```
nflaig, line 102: @wemeetagain I want to establish this pattern that for non-critical issues we use the terms "failed"  and "reason" instead of "error" as based on my observations the term "error" scares a lot of users, especially less technical ones. What do you think?

wemeetagain: :+1: :+1:

wemeetagain (approved):

packages/beacon-node/src/api/rest/base.ts
g11tech: i think the stack trace is very useful, but we can move the log to `debug` will full trace, an additional `warn` log in this format is acceptable

nflaig: > but we can move the log to debug will full trace, an additional warn log in this format is acceptable

that's also what I was considering, sounds like a reasonable compromise. Not sure if it makes sense here though in case of `ApiError`, if someone feels strongly about having a stack trace for those we should probably log it as well to `debug`

wemeetagain:

PR #5286, opened 2023-03-21
[DNM] Reprocess gossip attestations

twoeths:
**Motivation**

- There are so many unknown block root attestations passing through `executeWork` and they are reprocessed in uncontrollable way because when block come, they are processed immediately
- On a mainnet node subscribing to all subnets, there are too many of them and the pool reached limit

Screenshot 2023-03-21 at 13 11 10

**Description**
- All attestations pass through gossip queues
- If an attestation has unknown block root, send it to a new queue in block processor (implemented as a Map of  unkown block root)
- When block comes, get from this new queue and send to the gossip queue again
- On 2nd validation, do not need to deserialize the gossip message again

part of #5247

Note: the new gossip queue will be reverted in `unstable` when we merge v1.6.0 so making this PR as Draft


twoeths:
replaced by #5327

PR #5287, opened 2023-03-21
Log monitoring errors as warning and reduce verbosity

nflaig:
**Motivation**

Similar to changes introduced in https://github.com/ChainSafe/lodestar/pull/5285, the monitoring service is just a sidecar to the main functionality/process and should be considered non-critical, meaning it should be fine here if we just log this as a warning.

**Description**

Log monitoring errors as `warn` instead of `error` and reduce verbosity (no stacktrace).

The expected errors here are

- invalid domain, cannot resolve
- timeout, remote service is unresponsive
- dns resolution issues

The error message is enough to convey the issue, the stacktrace is not needed. Based on my observations using this for several weeks the errors I usually see are becasue of timeouts which are kinda expected since we send request 24/7 every minute and the remote service might be unresponsive sometimes.

### DIFF
--- a/packages/beacon-node/test/unit/monitoring/service.test.ts
+++ b/packages/beacon-node/test/unit/monitoring/service.test.ts
@@ -176,8 +176,8 @@ describe("monitoring / service", () => {
 
         // Validation of sent data happens inside the mocked remote service
         // which returns a 500 error if data does not match expected schema.
-        // Fail test if error was logged due to a 500 response.
-        expect(logger.error).to.not.have.been.calledOnce;
+        // Fail test if warning was logged due to a 500 response.
+        expect(logger.warn).to.not.have.been.calledWithMatch("Failed to send client stats");
       });
     });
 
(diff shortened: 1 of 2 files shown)


wemeetagain (dismissed):

packages/beacon-node/test/unit/monitoring/service.test.ts
```diff
@@ -231,10 +231,9 @@ describe("monitoring / service", () => {
       setTimeout(() => service.stop(), 10);
     });
 
-    function assertError(error: {name?: string; message?: string}): void {
-      expect(logger.error).to.have.been.calledOnce;
-      // errors are not thrown and need to be asserted based on the error log
-      expect(logger.error).to.have.been.calledWithMatch("Failed to send client stats", {}, error);
+    function assertError(error: {message: string}): void {
+      // errors are not thrown and need to be asserted based on the log
+      expect(logger.warn).to.have.been.calledWithMatch("Failed to send client stats", {reason: error.message});
```
nflaig, line 236: The assertions here are a bit brittle due to checking the error messages, but should not break unless the log level is changed and having this stricter assertion is important to ensure we are actually testing correctly.

wemeetagain (approved):

wemeetagain:

PR #5290, opened 2023-03-21
void

ghost:


philknows:
Hey @abebeos , thanks for your contribution. We value your opinion and will leave this PR open for more commenting from other users and maintainers. We generally won't merge something in that doesn't have a majority consensus on whether the problem being solved here is in fact, a problem. And that this PR rectifies this problem.

abebeos:
I understand, just be aware that this problem would be more of a "Product Management" or "CTO" issue:

If your CTO or Product Manager does not see the problem (with the emojs), then you have a problem with your management (because it fails to introduce/enforce quality-criteria)

As for those who "vote":

* Please show me one manual of a semiconductor that contains low-quality emojis (don't waste your time, you won't find one).

(Just to clarify: those emojis are nicely blockable via e.g. AdBlockPlus - I do not see them anymore in my browser, the problem is solved for me personally.)

nflaig:
Would be in favor of keeping the emojis as they convey additional information ℹ️ 

> (Just to clarify: those emojis are nicely blockable via e.g. AdBlockPlus - I do not see them anymore in my browser, the problem is solved for me personally.)

it's great that you found a solution to get rid of these abominations 😄

abebeos:
This "fun oriented" mentality keeps Ethereum and (crypto in general) back.

Lets get serious:

> additional information

Over-information, redundant cognitive load

wemeetagain:
I like the emojis :clown_face: 

:-1: from me

PR #5292, opened 2023-03-21
void

ghost:


README.md
```diff
@@ -27,19 +27,22 @@
...
 - :gear: [NodeJS](https://nodejs.org/) (LTS)
 - :toolbox: [Yarn](https://yarnpkg.com/)/[Lerna](https://lerna.js.org/)
 
-###### Developer Quickstart:
+### Developer Quickstart:
 
 ```bash
 lerna bootstrap
 yarn build
+lerna run test:unit
```
nflaig, line 40: IMO the whole point of the quickstart is to be "quick", meaning get project setup locally and learn about the existing commands (`./lodestar --help`) to get started

I am not sure if running the tests will provide valuable information here if we assume someone is completely new to Lodestar, they might just be overwhelmed...

README.md
```diff
@@ -27,19 +27,22 @@
...
 ./lodestar --help
 ```
 
+More tests and dev-info within [CONTRIBUTING.md](https://github.com/ChainSafe/lodestar/blob/unstable/CONTRIBUTING.md#tests).
```
nflaig: for any reference within the docs should use relative paths

```suggestion
More tests and dev-info within [CONTRIBUTING.md](/CONTRIBUTING.md#tests).
```

README.md
abebeos: > valuable information
Tests are not valuable info for a developer code-level quick-start?

Maybe add some emojis then?

nflaig: > Tests are not valuable info for a developer code-level quick-start?

Like I said, we are assuming here someone is looking at Lodestar for the first time and just wants to get started. I still don't really see why they should run the tests at this point, maybe you can elaborate a bit on why you think this is important here. If someone wants to dig deeper and contribute we have a reference to `CONTRIBUTING.md`.

Let's wait and get more opinions on this

abebeos: > Like I said, we are assuming here someone is looking at Lodestar for the first time and just wants to get started.

There it is again: "we"

> Let's wait and get more opinions on this

Opinions do not matter where a (quasi) standard exist.

Leaving you to it, me unsubscribing from this PR, as I do not have the nerves to discuss basics.

nflaig: > There it is again: "we"

the "we" implies it should be a general assumption, I will be more precise next time

PR #5295, opened 2023-03-22
Increase bls changes queue length to handle expected spikes on capella

g11tech:
Increase bls changes queue length to handle expected spikes on capella

Lighthouse Ref: https://github.com/sigp/lighthouse/blob/unstable/beacon_node/network/src/beacon_processor/mod.rs#L177

cc @wemeetagain @philknows


wemeetagain (approved):

wemeetagain:

PR #5299, opened 2023-03-23
Add logging policy

dapplion:
**Motivation**

Lodestar's logging should be more consistent across all contributors. It also not clear what should be logged at error vs warn, or at verbose vs debug.

This PR adds my own strong opinions and is marked as draft to receive feedback from other contributors.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -154,22 +154,21 @@ We're currently experimenting with hosting the majority of lodestar packages and
 
 ### Logging Levels
 
+Contributors must chose the log level carefully to ensure a consistent experience for every type of user:
+
 - `error`: Critical issues that prevent the application from functioning correctly or cause significant disruption to users. Examples include failed network connections, crashes, or data corruption.
 - `warn`: Situations that may lead to critical issues if not addressed but do not prevent the application from functioning. Examples include configuration issues, deprecated features, or temporary network disruptions.
 - `info`: General sporadic informational about the node's state. Examples include initialization messages, infrequent periodic status updates, or high-level progress reports.
 - `debug`: Detailed diagnostic information that can help developers or users troubleshoot specific issues. Examples include individual request logs for every REST API, networking interactions, or internal components status changes. Alias to `verbose`.
 - `trace`: Reserved for extremely verbose logs such a step by step execution traces of individual network packets or similar granular actions
 
-### Guideliness
+### Logging guidelines
 
-1. Always use the appropriate log level as defined above.
-2. Avoid excessive logging. Log messages should be clear and concise, providing enough information to understand the context and severity of the issue.
-3. Do not log sensitive data, such as private keys, user credentials, or personal information.
-4. Do not log arbitrary data from the network as ASCII or UTF8 at levels higher or equal to `info`.
-5. Use clear and concise language. Prefer to log variables in JSON format `log.debug("Action", {slot})` instead of formating the text yourself `log.debug('slot=${slot}')`.
-6. Include only relevant context in log messages, sufficient to debug the issue or action it refers to.
-
-Contributors must chose the log level carefully to ensure a consistent experience for every type of user:
+- Avoid excessive logging. Log messages should be clear and concise, providing enough information to understand the context and severity of the issue.
+- Do not log sensitive data, such as private keys, user credentials, or personal information.
+- Do not log arbitrary data from the network as ASCII or UTF8 at levels higher or equal to `info`.
+- Use clear and concise language. Prefer to log variables in JSON format `log.debug("Action", {slot})` instead of formatting the text yourself `log.debug('slot=${slot}')`.
+- Include only relevant context in log messages, sufficient to debug the issue or action it refers to.
 
 ## Contributing to Grafana dashboards
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -157,15 +157,15 @@ We're currently experimenting with hosting the majority of lodestar packages and
 - `error`: Critical issues that prevent the application from functioning correctly or cause significant disruption to users. Examples include failed network connections, crashes, or data corruption.
 - `warn`: Situations that may lead to critical issues if not addressed but do not prevent the application from functioning. Examples include configuration issues, deprecated features, or temporary network disruptions.
 - `info`: General sporadic informational about the node's state. Examples include initialization messages, infrequent periodic status updates, or high-level progress reports.
-- `verbose`: High-level state changes to internal modules. Examples include chain and state updates.
-- `debug`: Detailed diagnostic information that can help developers or users troubleshoot specific issues. Examples include individual request logs for every REST API, networking interactions, or internal components status changes.
+- `debug`: Detailed diagnostic information that can help developers or users troubleshoot specific issues. Examples include individual request logs for every REST API, networking interactions, or internal components status changes. Alias to `verbose`.
+- `trace`: Reserved for extremely verbose logs such a step by step execution traces of individual network packets or similar granular actions
 
 ### Guideliness
 
-1. Always use the appropriate log level as defined above. Log messages should be clear and concise, providing enough information to understand the context and severity of the issue.
-2. Avoid excessive logging. Log only the information that is necessary for understanding and troubleshooting issues. Over-logging can make it more difficult for users to find relevant information.
+1. Always use the appropriate log level as defined above.
+2. Avoid excessive logging. Log messages should be clear and concise, providing enough information to understand the context and severity of the issue.
 3. Do not log sensitive data, such as private keys, user credentials, or personal information.
-4. Do not log arbitrary data from the network as ASCII or UTF8 at levels higher or equal to the current default level.
+4. Do not log arbitrary data from the network as ASCII or UTF8 at levels higher or equal to `info`.
 5. Use clear and concise language. Prefer to log variables in JSON format `log.debug("Action", {slot})` instead of formating the text yourself `log.debug('slot=${slot}')`.
 6. Include only relevant context in log messages, sufficient to debug the issue or action it refers to.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -157,7 +157,7 @@ We're currently experimenting with hosting the majority of lodestar packages and
 - `error`: Critical issues that prevent the application from functioning correctly or cause significant disruption to users. Examples include failed network connections, crashes, or data corruption.
 - `warn`: Situations that may lead to critical issues if not addressed but do not prevent the application from functioning. Examples include configuration issues, deprecated features, or temporary network disruptions.
 - `info`: General sporadic informational about the node's state. Examples include initialization messages, infrequent periodic status updates, or high-level progress reports.
-- `verbose`: _TBD_
+- `verbose`: High-level state changes to internal modules. Examples include chain and state updates.
 - `debug`: Detailed diagnostic information that can help developers or users troubleshoot specific issues. Examples include individual request logs for every REST API, networking interactions, or internal components status changes.
 
 ### Guideliness
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -150,6 +150,27 @@ We're currently experimenting with hosting the majority of lodestar packages and
   - Error: One or more main functionalities are not working, preventing some functions from working properly.
   - Fatal: One or more main functionalities are not working and preventing the application from fulfilling its duties.
 
+## Logging policy
+
+### Logging Levels
+
+- `error`: Critical issues that prevent the application from functioning correctly or cause significant disruption to users. Examples include failed network connections, crashes, or data corruption.
+- `warn`: Situations that may lead to critical issues if not addressed but do not prevent the application from functioning. Examples include configuration issues, deprecated features, or temporary network disruptions.
+- `info`: General sporadic informational about the node's state. Examples include initialization messages, infrequent periodic status updates, or high-level progress reports.
+- `verbose`: _TBD_
+- `debug`: Detailed diagnostic information that can help developers or users troubleshoot specific issues. Examples include individual request logs for every REST API, networking interactions, or internal components status changes.
+
+### Guideliness
+
+1. Always use the appropriate log level as defined above. Log messages should be clear and concise, providing enough information to understand the context and severity of the issue.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -150,6 +150,27 @@ We're currently experimenting with hosting the majority of lodestar packages and
...
   - Fatal: One or more main functionalities are not working and preventing the application from fulfilling its duties.
 
+## Logging policy
+
+### Logging Levels
+
+- `error`: Critical issues that prevent the application from functioning correctly or cause significant disruption to users. Examples include failed network connections, crashes, or data corruption.
+- `warn`: Situations that may lead to critical issues if not addressed but do not prevent the application from functioning. Examples include configuration issues, deprecated features, or temporary network disruptions.
+- `info`: General sporadic informational about the node's state. Examples include initialization messages, infrequent periodic status updates, or high-level progress reports.
+- `verbose`: _TBD_
```
wemeetagain:
```suggestion
- `verbose`: High-level state changes to internal modules. Examples include chain and state updates.
```

CONTRIBUTING.md
```diff
@@ -150,6 +150,27 @@ We're currently experimenting with hosting the majority of lodestar packages and
...
+- `debug`: Detailed diagnostic information that can help developers or users troubleshoot specific issues. Examples include individual request logs for every REST API, networking interactions, or internal components status changes.
+
+### Guideliness
```
wemeetagain: I feel like we need a guideline about when/where to include a stack trace

CONTRIBUTING.md
```diff
@@ -150,6 +150,27 @@ We're currently experimenting with hosting the majority of lodestar packages and
...
+
+1. Always use the appropriate log level as defined above. Log messages should be clear and concise, providing enough information to understand the context and severity of the issue.
+2. Avoid excessive logging. Log only the information that is necessary for understanding and troubleshooting issues. Over-logging can make it more difficult for users to find relevant information.
+3. Do not log sensitive data, such as private keys, user credentials, or personal information.
+4. Do not log arbitrary data from the network as ASCII or UTF8 at levels higher or equal to the current default level.
```
nflaig: it is not clear to me what is meant by `at levels higher or equal to the current default level`, specifically what does the term "level" refer to here?

nflaig:
sounds pretty good to me 👍 

I don't (yet) have that strong opinions on how the logging should be done within Lodestar myself but I have written (3 years ago) a log level usage/policy guide myself before which could be used as  additional inspiration.

Just to add some context first, the defined logging policy was for a SaaS solution, meaning the logs were mostly meant to be consumed by developers/administrators and were used to create a log-based alerting system on top. It also has to be noted that logs were not the only tool used to diagnose issues, there were also metrics and additionally traces ([jaeger](https://www.jaegertracing.io/)) which were directly linked to logs.

This is how the logging policy was defined:

- `error` - Any error or exception that is or might be critical to the operation of the application. 
  It is recommended use an automated alerting system which informs a system administrator or developer about the error. 
  Most of the time `warn` should be preferred if the issue is recoverable without manual intervention.
- `warn` - Any message that warns about potential issues that could cause application oddities. Usually, those issues
  are automatically recoverable. It might also be used to warn about other potential problems such as invalid requests
  which could indicate malfunctioning of the application or even a security threat, e.g. a lot of failed login attempts.
- `info` - Generally useful information to log, e.g. service start/stop, configuration assumptions, etc. Usually, those
  logs can be ignored under normal circumstances if the application is running without problems but are important to have
  if an issue occurs to help diagnose the potential cause.
- `debug` - Information that is diagnostically helpful while developing the application. Debug logs should be disabled in
  a production environment during normal operation of the application. In some cases debug logs can be enabled during
  production to get additional information which might be helpful to diagnose an issue.

CONTRIBUTING.md
dapplion: That's a good point. How could we summarize it? I would say that when the error message provides all possible location information the stack trace can be omitted. We do not support suppressing stack traces out of the box, and the uncontrolled nature of Javascript exceptions could be an issue if we start omitting stack traces. **very** relevant to this topic https://www.youtube.com/watch?v=6cap2HiBFVQ

CONTRIBUTING.md
dapplion: level refers to log level with values: "error", "warn", etc. Please let me know how could I re-write the sentence for better comprehension

nflaig: "current default level" will always be `info` so I think we can make it a bit more explicit

```suggestion
4. Do not log arbitrary data from the network as ASCII or UTF8 at levels other than `debug` and `verbose`.
```

or

```suggestion
4. Do not log arbitrary data from the network as ASCII or UTF8 at levels higher or equal to `info`.
```

CONTRIBUTING.md
```diff
@@ -150,6 +150,27 @@ We're currently experimenting with hosting the majority of lodestar packages and
...
+
+- `error`: Critical issues that prevent the application from functioning correctly or cause significant disruption to users. Examples include failed network connections, crashes, or data corruption.
+- `warn`: Situations that may lead to critical issues if not addressed but do not prevent the application from functioning. Examples include configuration issues, deprecated features, or temporary network disruptions.
+- `info`: General sporadic informational about the node's state. Examples include initialization messages, infrequent periodic status updates, or high-level progress reports.
+- `verbose`: High-level state changes to internal modules. Examples include chain and state updates.
+- `debug`: Detailed diagnostic information that can help developers or users troubleshoot specific issues. Examples include individual request logs for every REST API, networking interactions, or internal components status changes.
+
+### Guideliness
+
+1. Always use the appropriate log level as defined above. Log messages should be clear and concise, providing enough information to understand the context and severity of the issue.
```
maschad: I think `1.` and `2.` are essentially saying very similar things. I would combine them to say

```suggestion
1. Logs should state clearly and concisely only that which is required to understand / troubleshoot an issue. Including any error codes, stack traces, debug logs or environment variables which are applicable.
```

building off https://github.com/ChainSafe/lodestar/pull/5299/files#r1150637797

CONTRIBUTING.md
```diff
@@ -150,6 +150,27 @@ We're currently experimenting with hosting the majority of lodestar packages and
   - Error: One or more main functionalities are not working, preventing some functions from working properly.
   - Fatal: One or more main functionalities are not working and preventing the application from fulfilling its duties.
 
+## Logging policy
+
+### Logging Levels
+
+- `error`: Critical issues that prevent the application from functioning correctly or cause significant disruption to users. Examples include failed network connections, crashes, or data corruption.
```
maschad, line 159: In the logging levels we can include that: 

`error` would normally contain error codes/ crash logs.

`warn` may include environment variables or other context related variables.

`debug` would include logs about specific issues, api and function calls in code that would be useful in testing / outside of production.

`verbose` may include stack traces  and all logs relevant to monitor performance

CONTRIBUTING.md
dapplion: Simplified 1 and 2

dapplion:
TBH from https://github.com/ChainSafe/lodestar/issues/5320 it's not clear to me what log would be debug or trace. Do you have any specific test if a specific log should go in one or the other?

I would even argue to log everything that's not info to debug and call it a day, super simple. In our deployments we only track debug. In CI we only persist debug. And in Javascript logging is not free, if we start adding crazy frequent trace statements it's gonna impact performance

maschad:
> I would even argue to log everything that's not info to debug and call it a day, super simple. In our deployments we only track debug. In CI we only persist debug. And in Javascript logging is not free, if we start adding crazy frequent trace statements it's gonna impact performance

Given that we may want to consider whether it's ideal to have trace logs then. The primary intention of debug logs imo should be to assist developers rather than consumers, and so perhaps we should omit trace logs for now.

nflaig:
> I would even argue to log everything that's not info to debug and call it a day, super simple. In our deployments we only track debug. In CI we only persist debug. And in Javascript logging is not free, if we start adding crazy frequent trace statements it's gonna impact performance

IMO using logs to do some sort of tracing is not the best option (maybe during development). Something like [opentelemetry](https://github.com/open-telemetry/opentelemetry-js) is much better to trace the application because you can nicely visualize it in [jaeger](https://www.jaegertracing.io/) and not have to read through 1000 lines of logs.

Also I would assume that `trace` is not enabled by default, so not very useful to retroactively diagnose an issue.

CONTRIBUTING.md
```diff
@@ -150,6 +150,27 @@ We're currently experimenting with hosting the majority of lodestar packages and
   - Error: One or more main functionalities are not working, preventing some functions from working properly.
   - Fatal: One or more main functionalities are not working and preventing the application from fulfilling its duties.
 
+## Logging policy
```
nflaig, line 153: maybe add some text after this heading to explain in one sentence why the Logging policy section exists and what it is supposed to achieve

CONTRIBUTING.md
```diff
@@ -150,6 +150,27 @@ We're currently experimenting with hosting the majority of lodestar packages and
...
+### Guideliness
+
+1. Always use the appropriate log level as defined above.
+2. Avoid excessive logging. Log messages should be clear and concise, providing enough information to understand the context and severity of the issue.
+3. Do not log sensitive data, such as private keys, user credentials, or personal information.
+4. Do not log arbitrary data from the network as ASCII or UTF8 at levels higher or equal to `info`.
+5. Use clear and concise language. Prefer to log variables in JSON format `log.debug("Action", {slot})` instead of formating the text yourself `log.debug('slot=${slot}')`.
+6. Include only relevant context in log messages, sufficient to debug the issue or action it refers to.
+
+Contributors must chose the log level carefully to ensure a consistent experience for every type of user:
```
nflaig:
```suggestion
Contributors must chose the log level carefully to ensure a consistent experience for every type of user.
```

CONTRIBUTING.md
```diff
@@ -150,6 +150,27 @@ We're currently experimenting with hosting the majority of lodestar packages and
...
+
+### Logging Levels
+
+- `error`: Critical issues that prevent the application from functioning correctly or cause significant disruption to users. Examples include failed network connections, crashes, or data corruption.
+- `warn`: Situations that may lead to critical issues if not addressed but do not prevent the application from functioning. Examples include configuration issues, deprecated features, or temporary network disruptions.
+- `info`: General sporadic informational about the node's state. Examples include initialization messages, infrequent periodic status updates, or high-level progress reports.
+- `debug`: Detailed diagnostic information that can help developers or users troubleshoot specific issues. Examples include individual request logs for every REST API, networking interactions, or internal components status changes. Alias to `verbose`.
+- `trace`: Reserved for extremely verbose logs such a step by step execution traces of individual network packets or similar granular actions
+
+### Guideliness
```
nflaig:
```suggestion
### Guidelines
```

CONTRIBUTING.md
```diff
@@ -150,6 +150,27 @@ We're currently experimenting with hosting the majority of lodestar packages and
...
+
+1. Always use the appropriate log level as defined above.
+2. Avoid excessive logging. Log messages should be clear and concise, providing enough information to understand the context and severity of the issue.
+3. Do not log sensitive data, such as private keys, user credentials, or personal information.
+4. Do not log arbitrary data from the network as ASCII or UTF8 at levels higher or equal to `info`.
+5. Use clear and concise language. Prefer to log variables in JSON format `log.debug("Action", {slot})` instead of formating the text yourself `log.debug('slot=${slot}')`.
```
nflaig:
```suggestion
5. Use clear and concise language. Prefer to log variables in JSON format `log.debug("Action", {slot})` instead of formatting the text yourself `log.debug('slot=${slot}')`.
```

nflaig (dismissed):

CONTRIBUTING.md
nflaig: kinda like this `Contributors must chose the log level carefully to ensure a consistent experience for every type of user`, maybe should just add this here and rephrase it a bit

CONTRIBUTING.md
```diff
@@ -150,6 +150,26 @@ We're currently experimenting with hosting the majority of lodestar packages and
   - Error: One or more main functionalities are not working, preventing some functions from working properly.
   - Fatal: One or more main functionalities are not working and preventing the application from fulfilling its duties.
 
+## Logging policy
+
+### Logging Levels
+
+Contributors must chose the log level carefully to ensure a consistent experience for every type of user:
```
wemeetagain:
```suggestion
Contributors must choose the log level carefully to ensure a consistent experience for every type of user:
```

CONTRIBUTING.md
```diff
@@ -150,6 +150,26 @@ We're currently experimenting with hosting the majority of lodestar packages and
...
+
+- `error`: Critical issues that prevent the application from functioning correctly or cause significant disruption to users. Examples include failed network connections, crashes, or data corruption.
+- `warn`: Situations that may lead to critical issues if not addressed but do not prevent the application from functioning. Examples include configuration issues, deprecated features, or temporary network disruptions.
+- `info`: General sporadic informational about the node's state. Examples include initialization messages, infrequent periodic status updates, or high-level progress reports.
+- `debug`: Detailed diagnostic information that can help developers or users troubleshoot specific issues. Examples include individual request logs for every REST API, networking interactions, or internal components status changes. Alias to `verbose`.
+- `trace`: Reserved for extremely verbose logs such a step by step execution traces of individual network packets or similar granular actions
```
wemeetagain: Are we removing trace?

wemeetagain (approved):

CONTRIBUTING.md
nflaig: Are we ok with not adding a bit more details about stack traces? I think the `ApiError` is a pretty good example of when not to include the stack trace as changed in https://github.com/ChainSafe/lodestar/pull/5285

- it is thrown on a higher level in the stack
- it is explicitly thrown based on simple assertions
- the error is expected to happen due to invalid client requests
- the error message is usually good which makes it easy to find out where it was thrown

Not including the stack trace of course can come with a cost of not being able to easily debug an issue but this should be rarely the case if we are careful to only do it for some errors but let's not forget the cost of including the stack trace in terms of UX. Not being able to easily read the logs because of too much stack trace is also not ideal.

I would argue that `debug` logs can include stack traces more often/always but for levels higher than or equal `info` need to be more careful.

wemeetagain: I just approved the PR so this didn't sit for too long. Happy to merge more verbiage about when stack traces are acceptable. Totally agree with your comment.

wemeetagain:

PR #5300, opened 2023-03-23
[DNM] Network processor without emitting event

twoeths:
**Motivation**

- We have network I/O lag issue since we introduce the new gossip queue
- I found that emitting event in gossipsub makes our event loop more busy, see https://itnext.io/high-performance-node-js-concurrency-with-events-on-hexomter-da4472ae3061

**Description**

- Use callback instead of network event to communicate between network processor and gossipsub.

part of #5247

same effort in `js-libp2p-gossipsub`: https://github.com/ChainSafe/js-libp2p-gossipsub/pull/416

**Test**
- Deployed this on feat1, this is on 16 validator node
Screenshot 2023-03-23 at 10 01 29

- While the same node on `unstable` (confirm that they have same subnet mesh peers => same network load)
Screenshot 2023-03-23 at 10 02 34

### DIFF
--- a/packages/beacon-node/src/network/events.ts
+++ b/packages/beacon-node/src/network/events.ts
@@ -1,11 +1,9 @@
 import {EventEmitter} from "events";
 import {PeerId} from "@libp2p/interface-peer-id";
 import StrictEventEmitter from "strict-event-emitter-types";
-import {TopicValidatorResult} from "@libp2p/interface-pubsub";
 import {phase0} from "@lodestar/types";
 import {BlockInput} from "../chain/blocks/types.js";
 import {RequestTypedContainer} from "./reqresp/ReqRespBeaconNode.js";
-import {PendingGossipsubMessage} from "./processor/types.js";
 
 export enum NetworkEvent {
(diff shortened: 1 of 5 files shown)


twoeths:
[0301_gossip-queue-pull_1k.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/11049626/0301_gossip-queue-pull_1k.cpuprofile.zip)

`events.emit` could take up to 10% of cpu time, mostly from gossipsub

Screenshot 2023-03-23 at 17 22 38

dapplion:
@tuyennhv but 0.00% of self time. Total time tracks the time spent on deeper stack traces, which makes sense. CPU time spent on the handler will be registered on the emit total time

twoeths:
> @tuyennhv but 0.00% of self time. Total time tracks the time spent on deeper stack traces, which makes sense. CPU time spent on the handler will be registered on the emit total time

@dapplion  yes I'm closing this PR, plus we should not follow this approach for the worker thread work as informed by @wemeetagain . Just don't understand why I see better metrics in this PR 🤔

PR #5302, opened 2023-03-23
Add SharedArrayBuffer example perf test

wemeetagain:
**Motivation**

[`SharedArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer) is another tool in our toolbelt for handling cross-thread communication. It's worth benchmarking a simple example to see what the performance characteristics look like.

**Description**

Add perf test testing sending a `Status` message across a `MessageChannel` vs sending via `SharedArrayBuffer`

Using `SharedArrayBuffer` to pass a `Status` is much faster, seemingly because ssz is faster than `structuredClone`.

```
  SharedArrayBuffer vs MessageChannel
    ✔ set and get status - MessageChannel                              1.268392e+8 ops/s    7.884000 ns/op   x1.039     141832 runs   1.31 s
    ✔ set and get status - SharedArrayBuffer                           5.467469e+8 ops/s    1.829000 ns/op   x1.022     779467 runs   1.82 s
    ✔ serialize/deserialize status - structuredClone                   1.637465e+8 ops/s    6.107000 ns/op        -     168109 runs   1.21 s
    ✔ serialize/deserialize status - ssz                               6.265664e+8 ops/s    1.596000 ns/op        -     684291 runs   1.62 s
```

### DIFF
--- a/packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
+++ b/packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
@@ -46,18 +46,11 @@ class StatusCacheReader {
   }
 }
 
-/**
- * 16_000 items: push then shift  - LinkedList is >200x faster than regular array
- *               push then pop - LinkedList is >10x faster than regular array
- * 24_000 items: push then shift  - LinkedList is >350x faster than regular array
- *               push then pop - LinkedList is >10x faster than regular array
- */
 describe("SharedArrayBuffer vs MessageChannel", () => {
   setBenchOpts({noThreshold: true});
 
   itBench({
     id: "set and get status - MessageChannel",
-    runsFactor: 1000,
     before: () => {
       const channel = new MessageChannel();
       return {
@@ -84,7 +77,6 @@ describe("SharedArrayBuffer vs MessageChannel", () => {
 
   itBench({
     id: "set and get status - SharedArrayBuffer",
-    runsFactor: 1000,
     before: () => {
       const buffer = new SharedArrayBuffer(STATUS_BUFFER_SIZE);
       const reader = new StatusCacheReader(buffer);
@@ -104,14 +96,12 @@ describe("SharedArrayBuffer vs MessageChannel", () => {
 
   itBench({
     id: "serialize/deserialize status - structuredClone",
-    runsFactor: 1000,
     beforeEach: () => ssz.phase0.Status.defaultValue(),
     fn: (status) => void structuredClone(status),
   });
 
   itBench({
     id: "serialize/deserialize status - ssz",
-    runsFactor: 1000,
     beforeEach: () => ssz.phase0.Status.defaultValue(),
     fn: (status) => void ssz.phase0.Status.deserialize(ssz.phase0.Status.serialize(status)),
   });
--- a/packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
+++ b/packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
@@ -101,4 +101,18 @@ describe("SharedArrayBuffer vs MessageChannel", () => {
       reader.get();
     },
   });
+
+  itBench({
+    id: "serialize/deserialize status - structuredClone",
+    runsFactor: 1000,
+    beforeEach: () => ssz.phase0.Status.defaultValue(),
+    fn: (status) => void structuredClone(status),
+  });
+
+  itBench({
+    id: "serialize/deserialize status - ssz",
+    runsFactor: 1000,
+    beforeEach: () => ssz.phase0.Status.defaultValue(),
+    fn: (status) => void ssz.phase0.Status.deserialize(ssz.phase0.Status.serialize(status)),
+  });
 });
--- a/packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
+++ b/packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
@@ -1,0 +1,104 @@
+import {itBench, setBenchOpts} from "@dapplion/benchmark";
+import {phase0, ssz} from "@lodestar/types";
+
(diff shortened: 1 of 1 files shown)


packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    const newSyncVal = Atomics.load(this.buffer, 0);
+    if (newSyncVal !== this.syncVal) {
+      this.syncVal = newSyncVal;
+      this.status = ssz.phase0.Status.deserialize(this.buffer.subarray(1));
+    }
+    return this.status;
+  }
+}
+
+/**
```
twoeths: these comments are for other tests

packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
```diff
@@ -0,0 +1,118 @@
...
+  itBench({
+    id: "serialize/deserialize status - structuredClone",
+    runsFactor: 1000,
+    beforeEach: () => ssz.phase0.Status.defaultValue(),
+    fn: (status) => void structuredClone(status),
+  });
+
+  itBench({
+    id: "serialize/deserialize status - ssz",
+    runsFactor: 1000,
```
twoeths: `runsFactor` should only be used if we loop this exact time inside `fn`, @dapplion can confirm

packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    return this.status;
+  }
+}
+
+/**
+ * 16_000 items: push then shift  - LinkedList is >200x faster than regular array
+ *               push then pop - LinkedList is >10x faster than regular array
+ * 24_000 items: push then shift  - LinkedList is >350x faster than regular array
+ *               push then pop - LinkedList is >10x faster than regular array
+ */
```
wemeetagain:
```suggestion
```

packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
```diff
@@ -0,0 +1,118 @@
...
+describe("SharedArrayBuffer vs MessageChannel", () => {
+  setBenchOpts({noThreshold: true});
+
+  itBench({
+    id: "set and get status - MessageChannel",
+    runsFactor: 1000,
```
wemeetagain:
```suggestion
```

packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    },
+    fn: async ({setStatus, getStatus, status}) => {
+      setStatus(status);
+      await getStatus;
+    },
+  });
+
+  itBench({
+    id: "set and get status - SharedArrayBuffer",
+    runsFactor: 1000,
```
wemeetagain:
```suggestion
```

packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    beforeEach: (args) => args,
+    fn: ({reader, writer, status}) => {
+      writer.update(status);
+      reader.get();
+    },
+  });
+
+  itBench({
+    id: "serialize/deserialize status - structuredClone",
+    runsFactor: 1000,
```
wemeetagain:
```suggestion
```

packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
wemeetagain:
```suggestion
```

packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
```diff
@@ -0,0 +1,108 @@
...
+      return {
+        reader,
+        writer,
+        status: ssz.phase0.Status.defaultValue(),
+      };
+    },
+    beforeEach: (args) => args,
+    fn: ({reader, writer, status}) => {
+      writer.update(status);
+      reader.get();
```
twoeths, line 93: @wemeetagain now we only test on main thread, does the read/write from/to `SharedArrayBuffer` has same speed from different worker threads?

packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+      return {
+        setStatus,
+        getStatus,
+        status,
+      };
+    },
+    fn: async ({setStatus, getStatus, status}) => {
+      setStatus(status);
+      await getStatus;
```
twoeths, line 74: @wemeetagain I'm not sure if this test is comparable to the `SharedArrayBuffer` test below because one is async and one is sync. Also it's good to test send/receive of n messages instead of 1

packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
wemeetagain: Not sure, I should try adding more perf tests that test across workers

packages/beacon-node/test/perf/util/sharedArrayBuffer.test.ts
wemeetagain: This is one of the advantages of using SharedArrayBuffer, that it can provide a sync interface.
Will try adding a test for N messages.

dapplion:
lib2p worker PR is merged, and we don't have plans to use ShareArrayBuffer there

PR #5308, opened 2023-03-27
Add unit test to check for unused exports and remove them

### LINKED ISSUE #2209: Remove dead code
At least this util is dead code and could be removed https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/util/io.ts

It would be nice to run [`ts-prune`](https://www.npmjs.com/package/ts-prune) on each package of the monorepo to find and delete dead code

nicobao:
**Motivation**

The goal is this PR is to:
1. create a reproducible unit test to check for unused exports (values used within the same module - but unused in the same package or cross-package) or dead-code (exported values that's never used, even within the same module)
1. remove detected unnecessary `export` keywords and delete detected dead code

Only 1. has been done for now.

How does it work?
1. get the dependency tree
3. get the directories per package
4. get the list of subpackages per package
5. get the list of unused exports per subpackage (using `ts-prune`). It will only look for usage within the same package the subpackage is in
6. get the list of imported values per subpackage (using `ripgrep`)
7. compare the list of exports created in 4. with the list of imports created in 5. and conclude

Why not just use `ts-prune`?
- `ts-prune` does not support monorepo & lerna. There have been PRs - [here](https://github.com/nadeesha/ts-prune/pull/213) and [there](https://github.com/nadeesha/ts-prune/issues/208) but the repo is in maintenance mode so they were closed.
- `ts-prune` only search for usages within a package
- but in our case, almost all exports are used cross-package (think `@lodestar/api`)
- it is still useful to get a first list of potential unused exports

Why `ripgrep`?
- it is fast and widely used
- I know it
- there is an npm package to use it in nodejs, used by vscode

Known limitations and possible improvements:
- the way it detects actual usage from `import * as foo from 'subpkg'` is potentially incomplete or inaccurate
- it's relatively slow - it wasn't optimized for speed, I just wanted to make it work first
- it only supports ES Module import, not CommonJS (but we don't really care)
- it would be nice if it was just an external tool that we would use as a CLI instead of creating a specific package for this test
- maybe adding monorepo support to a fork of `ts-prune` would have been better

What I need from you:
- the test seems to work as expected: it fails with the list of unused exported values. The values I manually checked do seem to be unused across packages.
- I need to know which of the unused exported values I should really act upon? (removing 'export' or deleting dead-code)

For convenience, I have put [here](https://raw.githubusercontent.com/nicobao/lodestar/unused-exports/packages/check-dead-code/result.json) the list of unused exported values that the test has found. You can log the same value by running the failing test (it will just not output a beautiful json like the one I provided).

Closes #2209 

```sh
cd packages/check-dead-code
yarn test:unit
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "supertest": "^6.2.4",
     "ts-loader": "^9.3.1",
     "ts-node": "^10.9.1",
+    "ts-prune": "^0.10.3",
     "typescript": "4.9.3",
     "typescript-docs-verifier": "^2.3.0",
     "webpack": "^5.74.0"
--- a/packages/check-dead-code/LICENSE
+++ b/packages/check-dead-code/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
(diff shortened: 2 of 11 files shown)


dapplion:
Hey @nicobao :wave: 

Thank you so much for this contribution! This is indeed a very interesting approach. Your PR shows that there must be a way to say what exposed variables are ok and which are not. In most libraries almost everything is exposed, some of which could be considered dead code and others not. This libraries are published to NPM so we must account for arbitrary usage.

@wemeetagain a potential first step would be to restrict exports to only expose exactly what we want to be used externally. Then this tool cleans up the rest.

@nicobao while this tool is very useful I don't think it should live in the Lodestar monorepo as is completely independent. If that's of interest to you could you move it to an independent repo and PR the usage of it strictly?

nicobao:
Hi @dapplion :)

Thank you for the kind comments. I am glad that you appreciate my contribution.

The goal of the test is precisely to compare the [expected](https://github.com/nicobao/lodestar/blob/83e97b5927e80ea254796763ce570fe141407748/packages/check-dead-code/test/unit/assert.test.ts#L501) unused values (the one we want to expose as an api in our npm packages) with the [actual](https://github.com/nicobao/lodestar/blob/83e97b5927e80ea254796763ce570fe141407748/packages/check-dead-code/test/unit/assert.test.ts#L500) unsued values (generated by the tool I wrote).

The remaining unknown variable is the [expected](https://github.com/nicobao/lodestar/blob/83e97b5927e80ea254796763ce570fe141407748/packages/check-dead-code/test/unit/assert.test.ts#L501) values. It should not be empty, it was just a placeholder. I believe that's the data you mentioned to @wemeetagain.

By writing the exports we want to be used externally in this `expected` variable instead of on a comment in GitHub or in a .md file, we can compare it with the output of the tool, and we obtain a reproducible test.

@dapplion, where do you want to run the test? 
Can I keep it in its own lerna package? Or do you want the test to be fully CLI-based instead of using mocha?
The tool, when I externalize it - let's call it `ts-prune-monorepo`, will need to have access to the whole lerna project. It will run either as a TS library or a CLI. The tool assumes npx, yarn & lerna are available in the host computer: that's the only dependencies that I didn't internalize. There must be a way though, I didn't try much.

nazarhussain:
Moving forward we should also consider that `ts-prune` went into maintenance mode and not actively developed. It's future seems not clear and with that state I am not confident to use it in Lodestar.

nicobao:
@nazarhussain Maintenance mode means that they will not add new features, not that the current features are unmaintained:

```
However, due to the popularity of ts-prune, it has absorbed more use cases, and complexity has bloated to the point that I'm no longer comfortable to add more features or do any other changes to the core system.

The most important thing for ts-prune is to be backwards compatible and reliable for existing use cases.
```

I agree though, that long term it's probably not wise to rely on this tool.

Do you want to use the tool I made?
If yes, I can link only the usage but I need to know how you expect it to be used in Lodestar. How do you see it?

dapplion:

Even if we didn't merge, thank you so much for contributing @nicobao

nflaig:
This is a pretty good starting point and can be used without merging it when someone with a good knowledge of the codebase wants to take the time and look into https://github.com/ChainSafe/lodestar/issues/2209. I checked the json results, there are definitely false positives and changes need to be manually reviewed but will help a lot to get an overview. The fact that the Lodestar libraries can be used by anyone makes things a lot harder though, we have to assume everything that is exported at this points is used, should only remove exports in a new major release.

Generally, I would rather see this as a CLI / general purpose tool as it might also be useful to other projects.

nicobao:
Happy to have been of some help. Unfortunately, I won't have time to make it an external tool. My goal was not to build a general-purpose tool but to make a script that solves the specific problem lodestar has, and for that purpose the solution as is works. 

Turning this script into a general-purpose tool would require a bit of refactoring as to how things are loaded, and I am pretty sure the approach I took wasn't the most appropriate in terms of performance and reliability. If somebody wants to make a general-purpose tool, I'd rather look at ts-prune and other similar tool, eventually fork one of them and use the script I made as inspiration only.

PR #5309, opened 2023-03-27
Add Metrics to track blocks proposed and orphaned 

### LINKED ISSUE #5255: New metrics to track proposed blocks

Need new metrics to track blocks proposed by a lodestar node's local validators (same to `beaconcha.in`:
- Blocks proposed and included in canonical chain
- Orphaned blocks
- Missed blocks

maschad:
**Motivation**

We would like to have more insight into how a Lodestar's local validator is performing with regards to block production.

**Description**

This tracks how many blocks have been:
- Blocks proposed and included in canonical chain
- Orphaned blocks

related to #5255 

## Update
Tracking missed blocks will involve some other changes which I will include in another PR.


packages/validator/src/services/block.ts
```diff
@@ -110,6 +110,7 @@ export class BlockProposingService {
         builderSelection,
       }).catch((e: Error) => {
         this.metrics?.blockProposingErrors.inc({error: "produce"});
+        this.metrics?.missedBlocks.inc();
```
dapplion: There's already a metric in the previous line, why this duplicate metric necessary?

packages/beacon-node/src/sync/unknownBlock.ts
```diff
@@ -244,6 +246,8 @@ export class UnknownBlockSync {
       else {
         this.logger.error("Unknown error processing block from unknown parent sync", errorData, res.err);
         pendingBlock.status = PendingBlockStatus.pending;
+        // This potentially an orphaned block
+        this.metrics?.syncUnknownBlock.orphanedBlocks.inc();
```
dapplion: Why is this a potentially orphaned block? Orphaned block generally means a block that is included in the chain at time T1, but re-org out at time T2 > T1. You can only determine with certainty the "orphan" status at finalization or probabilistically before.

packages/validator/src/services/block.ts
dapplion: "missed block" is a fuzzy term that usually refers to a block that did not make it on-chain. Block production error is one of the reasons, but could be more which you cannot track from the validator client

packages/validator/src/services/block.ts
```diff
@@ -122,11 +123,13 @@ export class BlockProposingService {
 
       await this.publishBlockWrapper(signedBlock).catch((e: Error) => {
         this.metrics?.blockProposingErrors.inc({error: "publish"});
+        this.metrics?.missedBlocks.inc();
         throw extendError(e, "Failed to publish block");
       });
       this.logger.info("Published block", {...logCtx, graffiti, ...block.debugLogCtx});
       this.metrics?.blocksPublished.inc();
     } catch (e) {
+      this.metrics?.missedBlocks.inc();
```
dapplion: If `publishBlockWrapper` throws you will count missedBlocks twice

packages/validator/src/services/blockDuties.ts
```diff
@@ -182,4 +183,22 @@ export class BlockDutiesService {
       }
     }
   }
+
+  private async isIncludedInCanonicalChain(slot: Slot): Promise<boolean> {
+    //#TODO: figure out how to get the latest proposed block and compare it to the canonical block
```
dapplion: As noted above implementing this function is not easy and will rely on probabilistic assumptions. The beacon chain has complete info about the chain it should be done there

packages/validator/src/services/block.ts
maschad: Correct, just to clarify my understanding is a _missed block_ is one where a validator node fails to validate a block that it was selected to validate and yes this is just one instance of it (hence there is a separate metric i.e. `missedBlocks`)

packages/validator/src/services/blockDuties.ts
maschad: Good point

packages/validator/src/services/block.ts
dapplion: You should not create duplicate metrics, since in Grafana you can easily aggregate then with dedicated queries

packages/beacon-node/src/sync/unknownBlock.ts
maschad: Yah I suppose [Line 227](https://github.com/ChainSafe/lodestar/pull/5309/files#diff-593e2f511966ad9d8cb737688f30aadb67c896c68766d1b09a33cf1e251f7860R227) covers the situation where a block has an unknown parent

packages/beacon-node/src/chain/chain.ts
```diff
@@ -112,6 +112,8 @@ export class BeaconChain implements IBeaconChain {
   // Seen cache for liveness checks
   readonly seenBlockAttesters = new SeenBlockAttesters();
 
+  readonly blockProposerIndices = new Map<Uint8Array, ValidatorIndex>();
```
wemeetagain: You shouldn't use Uint8Array as a Map key (except in certain special circumstances). Map uses object equality to match keys, so two distinct Uint8Arrays with the same binary contents will be treated as two distinct keys by the Map.

```ts
const m = new Map<Uint8Array, string>();
m.set(Uint8Array.from([0]), "foo");
m.get(Uint8Array.from([0])); // returns undefined
```

maschad: Right, good catch

packages/beacon-node/src/api/impl/beacon/blocks/index.ts
```diff
@@ -213,6 +213,8 @@ export function getBeaconBlockApi({
 
       // TODO: Validate block
 
+      chain.blockProposerIndices.set(toHexString(signedBlock.message.stateRoot), signedBlock.message.proposerIndex);
```
wemeetagain, line 216: You will want to track blocks by block root, not state root.
You will need to use the fork choice to get all ancestors of the recently finalized block and iterate thru that list to check against this map.
You will need to ensure that you check the fork choice before it is pruned, since it is pruned on finalization.

g11tech:
Closing in favour of 
- https://github.com/ChainSafe/lodestar/pull/5448

PR #5311, opened 2023-03-28
Process api attestations in batch

twoeths:
**Motivation**

`submitPoolsAttestation` takes more time with the new gossip queues, part of the reason is we validate their signatures in different bls batch (because there are many bls signatures to validate with new gossip queues)

**Description**

- Validate all api attestations in same batch if they have same data, this will:
  - Validate a lot of common data once
  - `computeSigningRoot()` once, we can have a cheap way to get signature set, see `cloneIndexedAttestationSignatureSet()`
  - Generate state once
  - Validate signatures in one batch
- Name the new api `validateApiAttestations`
- If attestation datas are not the same, fall back to the old api `validateGossipAttestation`

### DIFF
--- a/packages/beacon-node/src/chain/validation/attestation.ts
+++ b/packages/beacon-node/src/chain/validation/attestation.ts
@@ -1,4 +1,4 @@
-import {phase0, Epoch, Root, Slot} from "@lodestar/types";
+import {phase0, Epoch, Root, Slot, ValidatorIndex} from "@lodestar/types";
 import {ProtoBlock} from "@lodestar/fork-choice";
 import {ATTESTATION_SUBNET_COUNT, SLOTS_PER_EPOCH} from "@lodestar/params";
 import {toHexString} from "@chainsafe/ssz";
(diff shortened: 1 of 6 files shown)


packages/beacon-node/src/chain/validation/attestation.ts
```diff
@@ -161,6 +164,161 @@ export async function validateGossipAttestation(
...
+    }
+
+    validationResults.push({indexedAttestation, subnet: expectedSubnet});
+  }
+
+  if (!(await chain.bls.verifySignatureSets(signatureSets, {batchable: true}))) {
+    throw new AttestationError(GossipAction.REJECT, {code: AttestationErrorCode.INVALID_SIGNATURE});
+  }
+
+  validatorIndices.forEach((validatorIndex) => chain.seenAttesters.add(targetEpoch, validatorIndex));
```
wemeetagain: can you use `for(const validatorIndex of validatorIndices) {`

packages/beacon-node/src/chain/validation/attestation.ts
```diff
@@ -161,6 +164,161 @@ export async function validateGossipAttestation(
...
+        validatorIndex,
+      });
+    }
+
+    validationResults.push({indexedAttestation, subnet: expectedSubnet});
+  }
+
+  if (!(await chain.bls.verifySignatureSets(signatureSets, {batchable: true}))) {
+    throw new AttestationError(GossipAction.REJECT, {code: AttestationErrorCode.INVALID_SIGNATURE});
+  }
```
dapplion, line 325: The behavior as implemented here will cause all attestations to be rejected if only of the them is wrong. This is not correct, and while you can argue that an APIi is _likely_ to submit attestations that are all valid this is not an assumption I'm comfortable taking. If you want to go the route of batchable verify functions I did an implementation in this branch which supports individual status reporting https://github.com/ChainSafe/lodestar/blob/dapplion/attestation-batch/packages/beacon-node/src/chain/validation/attestation.ts

twoeths: if it fails here, at the `submitPoolAttesatations()` api, it'll fallback to verify each attestation separately and will publish only valid attestation as the old flow https://github.com/ChainSafe/lodestar/pull/5311/files#diff-2381400ee782266412ea40f4ba10a08e49c7e60a3b63dbacaa8774d9ae583fcfR111

I should add some comments to this line

twoeths:
it was my wrong assumption that `vc` always submit attestations with same data, the reality is it may contain different committee indices. While we can sort attestation data and validate in batch at api side, it may not worth it as it's not the root cause of I/O lag issue. May consider it later, closing the PR for now.

PR #5316, opened 2023-03-29
Handle gossip messages asynchronously

twoeths:
**Motivation**

- There are I/O lag issue with the new gossip queue: more time to submit attestations from `vc`, more time to call `notifyNewPayload` engine api
- Lifecycle of a gossipsub message is all synchronous:
  - we call `onGossipsubMessage` synchronously
  - which call gossip validation function synchronously
  - which call the `reportMessageValidationResult` synchronously
  - this is really bad for the I/O in case `RPC` contains thousands of gossip messages
- In current implementation (`stable/unstable`) when we push gossip messages to `JobItemQueue` we always use `setTimeout()` which helps avoid the I/O lag issue, we also yield to the macro queue every 50ms there

**Description**

- Use `setTimeout` to break our gossip handler to multiple event loops which helps the I/O lag issue (which make it equivalent to the current behavior of `stable` branch)
- Metrics show that it does not effect performance at all: almost same job wait time, job process time, no dropped job

part of #5247

### DIFF
--- a/packages/beacon-node/src/network/gossip/gossipsub.ts
+++ b/packages/beacon-node/src/network/gossip/gossipsub.ts
@@ -405,7 +405,9 @@ export class Eth2Gossipsub extends GossipSub implements GossipBeaconNode {
     // Get seenTimestamp before adding the message to the queue or add async delays
     const seenTimestampSec = Date.now() / 1000;
 
-    // Emit message to network processor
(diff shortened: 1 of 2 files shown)


wemeetagain (approved):

dapplion (approved):

wemeetagain:

PR #5324, opened 2023-03-30
Unschedule capella in chiado and gnosis

g11tech:
Because of mainnet scheduling of Capella a bug was introduced in chiado and gnosis configs where which inherits configs from mainnet

This PR explicitly unschedules capella iin them


nflaig (approved):

wemeetagain:
:tada: This PR is included in [v1.7.2](https://github.com/ChainSafe/lodestar/releases/tag/v1.7.2) :tada:

dapplion:
Yeah I noticed the risk of scheduling forks accidentally due to how they are inherit. I think we should have some more definitive solution to prevent this from happening again

wemeetagain:

PR #5325, opened 2023-03-30
Add support for the dial only protocols

nazarhussain:
**Motivation**

Add support for the dial only protocol in the `reqresp` package. 

**Description**

While building Lightclient P2P transport for the Prover, came upon a scenario to use dial only protocols. Current `reqresp` package does not support it. So had to refactor the reqresp package. 

Ground work for https://github.com/ChainSafe/lodestar/issues/4706

**Note:** The usage of this refactor will come in the upcoming PRs. Didn't wanted to blow up the scope of that prover PR, so splitted tasks in different small PRs. 

- Run all tests.

### DIFF
--- a/packages/reqresp/src/types.ts
+++ b/packages/reqresp/src/types.ts
@@ -45,7 +45,7 @@ export interface InboundRateLimitQuota<Req = unknown> {
 }
 
 // `protocolPrefix` is added runtime so not part of definition
-export interface DuplexProtocolDefinition<Req = unknown, Resp = unknown> extends Omit<Protocol, "protocolPrefix"> {
+export interface ProtocolDefinition<Req = unknown, Resp = unknown> extends Omit<Protocol, "protocolPrefix"> {
   handler: ReqRespHandler<Req, Resp>;
   requestType: (fork: ForkName) => TypeSerializer<Req> | null;
   responseType: (fork: ForkName) => TypeSerializer<Resp>;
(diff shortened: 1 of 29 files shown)


dapplion:
Since DuplexProtocolDefinition is the most used, why not keep it named ProtocolDefinition and then use a special name for dial only protocols?

nazarhussain:
> Since DuplexProtocolDefinition is the most used, why not keep it named ProtocolDefinition and then use a special name for dial only protocols?

That could be confusing to name the union of both. What about using `MixedProtocolDefinition` for the union?

packages/reqresp/src/types.ts
```diff
@@ -55,13 +55,29 @@ export interface ProtocolDefinition<Req = unknown, Resp = unknown> extends Omit<
...
 
-export type ProtocolDefinitionGenerator<Req, Res> = (
+export interface DialOnlyProtocolDefinition<Req = unknown, Resp = unknown>
+  extends Omit<ProtocolDefinition<Req, Resp>, "handler" | "inboundRateLimits" | "renderRequestBody"> {
+  handler?: never;
+  inboundRateLimits?: never;
+  renderRequestBody?: never;
+}
+
+export type MixedProtocolDefinition<Req = unknown, Resp = unknown> =
```
wemeetagain, line 101: Please add jsdoc comments for these types / interfaces

nazarhussain: @wemeetagain done

wemeetagain (approved):

wemeetagain:

PR #5327, opened 2023-03-31
Handle unknown block root attestations

twoeths:
**Motivation**

Right now if a gossip attestation is validated and has unknown block root, it'll be reprocessed once the block comes. This has some drawback:
- Some redundancy to validate unknown block root attestations because we can extract it from the serialized and do the check first
- The number of attestations to reprocess could be up to 16_000, it's bad for the event loop to reprocess all of them at once

**Description**

- Extract the slot and root from Attestation first and put to a map of root/slot if the block has not come yet. Ignore the message if there are too many items in the waiting queue
- Process them once block come
- Remove waiting gossip messages if block hasn't come in same slot

### DIFF
--- a/packages/beacon-node/src/network/processor/extractSlotRootFns.ts
+++ b/packages/beacon-node/src/network/processor/extractSlotRootFns.ts
@@ -5,6 +5,7 @@ import {
   getSlotFromAttestationSerialized,
   getSlotFromSignedAggregateAndProofSerialized,
 } from "../../util/sszBytes.js";
+import {nullableResult} from "../../util/wrapError.js";
 import {GossipType} from "../gossip/index.js";
 import {ExtractSlotRootFns} from "./types.js";
 
@@ -14,17 +15,21 @@ import {ExtractSlotRootFns} from "./types.js";
  */
 export function createExtractBlockSlotRootFns(): ExtractSlotRootFns {
   return {
-    [GossipType.beacon_attestation]: (data: Uint8Array): SlotRootHex => {
-      return {
-        slot: getSlotFromAttestationSerialized(data),
-        root: getBlockRootFromAttestationSerialized(data),
-      };
-    },
-    [GossipType.beacon_aggregate_and_proof]: (data: Uint8Array): SlotRootHex => {
-      return {
-        slot: getSlotFromSignedAggregateAndProofSerialized(data),
-        root: getBlockRootFromSignedAggregateAndProofSerialized(data),
-      };
-    },
+    [GossipType.beacon_attestation]: nullableResult(getSlotRootFromAttestationSerialized),
+    [GossipType.beacon_aggregate_and_proof]: nullableResult(getSlotRootFromAggregateAndProofSerialized),
+  };
+}
+
+function getSlotRootFromAttestationSerialized(data: Uint8Array): SlotRootHex {
+  return {
+    slot: getSlotFromAttestationSerialized(data),
+    root: getBlockRootFromAttestationSerialized(data),
+  };
+}
+
+function getSlotRootFromAggregateAndProofSerialized(data: Uint8Array): SlotRootHex {
+  return {
+    slot: getSlotFromSignedAggregateAndProofSerialized(data),
+    root: getBlockRootFromSignedAggregateAndProofSerialized(data),
   };
 }
--- a/packages/beacon-node/src/network/processor/extractSlotRootFns.ts
+++ b/packages/beacon-node/src/network/processor/extractSlotRootFns.ts
@@ -5,7 +5,6 @@ import {
   getSlotFromAttestationSerialized,
   getSlotFromSignedAggregateAndProofSerialized,
 } from "../../util/sszBytes.js";
-import {nullableResult} from "../../util/wrapError.js";
 import {GossipType} from "../gossip/index.js";
 import {ExtractSlotRootFns} from "./types.js";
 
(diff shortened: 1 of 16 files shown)


twoeths:
Some metrics captured

Screenshot 2023-04-04 at 21 30 43

Screenshot 2023-04-04 at 21 32 22

Screenshot 2023-04-04 at 21 33 33

==> it'll be improved if we trigger block search once we receive an unknown block root attestation, that's to be done in a separate work once we merge `new_gossip_queues` branch to unstable

packages/beacon-node/src/network/processor/extractSlotRootFns.ts
```diff
@@ -0,0 +1,30 @@
+import {SlotRootHex} from "@lodestar/types";
+import {
+  getBlockRootFromAttestationSerialized,
+  getBlockRootFromSignedAggregateAndProofSerialized,
+  getSlotFromAttestationSerialized,
+  getSlotFromSignedAggregateAndProofSerialized,
```
dapplion, line 6: @tuyennhv Can you add bad data unit tests for 
- getBlockRootFromAttestationSerialized
- getBlockRootFromSignedAggregateAndProofSerialized,
- getSlotFromAttestationSerialized,
- getSlotFromSignedAggregateAndProofSerialized,

For each: decode an empty buffer, and decode a buffer that ends right after the offset such that this function produces incomplete data. The unit test should ensure that this functions never throw unexpectedly and let's think what to do in that case

packages/beacon-node/src/network/processor/index.ts
```diff
@@ -48,6 +51,31 @@ const executeGossipWorkOrder = Object.keys(executeGossipWorkOrderObj) as (keyof
...
 
+// How many attestations (aggregate + unaggregate) we keep before new ones get dropped.
+const MAX_QUEUED_UNKNOWN_BLOCK_GOSSIP_OBJECTS = 16_384;
+
+// We don't want to process too many attestations in a single tick
+// As seen on mainnet, utilization rate for attestation topic is 80_000% (800x) - 90_000% (900x)
+// so make this constant go with that number
+const MAX_UNKNOWN_BLOCK_GOSSIP_OBJECTS_PER_TICK = 1024;
+
+// Same motivation to JobItemQueue, we don't want to block the vent loop
```
dapplion:
```suggestion
// Same motivation to JobItemQueue, we don't want to block the event loop
```

packages/beacon-node/src/network/processor/index.ts
```diff
@@ -48,6 +51,31 @@ const executeGossipWorkOrder = Object.keys(executeGossipWorkOrderObj) as (keyof
 // TODO: Arbitrary constant, check metrics
 const MAX_JOBS_SUBMITTED_PER_TICK = 128;
 
+// How many attestations (aggregate + unaggregate) we keep before new ones get dropped.
+const MAX_QUEUED_UNKNOWN_BLOCK_GOSSIP_OBJECTS = 16_384;
+
+// We don't want to process too many attestations in a single tick
+// As seen on mainnet, utilization rate for attestation topic is 80_000% (800x) - 90_000% (900x)
```
dapplion: We should stop looking at the utilization rate as it does not really make sense and track the concurrency metric

https://github.com/ChainSafe/lodestar/blob/887064d7f497adf123a3839ec4c71c323c24bbdd/packages/beacon-node/src/metrics/metrics/lodestar.ts#L269-L270

packages/beacon-node/src/util/sszBytes.ts
```diff
@@ -38,6 +38,7 @@ export function getBlockRootFromAttestationSerialized(data: Uint8Array): BlockRo
 }
 
 export function getAttDataHashFromAttestationSerialized(data: Uint8Array): AttDataHash {
```
dapplion: If we are using base64 I would be extremely explicit everywhere. Same this function `getAttDataBase64FromAttestationSerialized` and when using it, name the variable `attDataBase64`

packages/beacon-node/src/network/processor/index.ts
twoeths: Screenshot 2023-04-05 at 08 44 16

The concurrency metric ranges from 1000 to 2000 so I'd make it a little bit conservative, 1024, too. Will change the comment to base on this metric.

packages/beacon-node/src/network/processor/extractSlotRootFns.ts
twoeths: I added invalid test data to those functions. I think we should let the consumer decide what to do instead of never throw and catch error silently
+ If getting block or slot throws error, we ignore the forkchoice check to handle unknown block root, when deserializing the data it'll throw error and reject
+ if getAttDataBase64 throws error, it acts like not having a cached Attestation Data. This is not likely to happen because gossip message should be deserialized successfully earlier

packages/beacon-node/src/util/sszBytes.ts
twoeths: I changed it accordingly 👍

packages/beacon-node/src/network/processor/extractSlotRootFns.ts
```diff
@@ -0,0 +1,35 @@
...
+import {GossipType} from "../gossip/index.js";
+import {ExtractSlotRootFns} from "./types.js";
+
+/**
+ * Extract the slot and block root of a gossip message form serialized data.
+ * Only applicable for beacon_attestation and beacon_aggregate_and_proof topics.
+ */
+export function createExtractBlockSlotRootFns(): ExtractSlotRootFns {
+  return {
+    [GossipType.beacon_attestation]: nullableResult(getSlotRootFromAttestationSerialized),
```
dapplion: Let's define the API directly as return base64 string or null if invalid, no need to wrap with this function if there's a single error case

packages/beacon-node/src/util/sszBytes.ts
```diff
@@ -28,29 +28,58 @@ export type AttDataHash = string;
 const ATTESTATION_SLOT_OFFSET = 4;
 const ATTESTATION_BEACON_BLOCK_ROOT_OFFSET = ATTESTATION_SLOT_OFFSET + 8 + 8;
 const ROOT_SIZE = 32;
+const SLOT_SIZE = 8;
+const ATTESTATION_DATA_SIZE = 128;
 
 export function getSlotFromAttestationSerialized(data: Uint8Array): Slot {
+  if (data.length < ATTESTATION_SLOT_OFFSET + SLOT_SIZE) {
+    throw Error(`Invalid attestation data length ${data.length}`);
```
dapplion: Since there is a single failure mode and this is a hot path let's just return null here and document in the JSDocs that null means data too short. Since all of these are SSZ objects we can compute the min size ahead of time and enforce it higher up in the stack, such as in gossip snappy decode step. What do you think? meaning to do both

packages/beacon-node/src/util/wrapError.ts
```diff
@@ -20,3 +20,20 @@ export async function wrapError<T>(promise: Promise<T>): Promise<Result<T>> {
...
+): (...args: Parameters<T>) => ReturnType<T> | null {
+  return (...args: Parameters<T>): ReturnType<T> | null => {
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+      return fn(...args);
+    } catch (e) {
+      return null;
+    }
+  };
+}
```
dapplion: Let's drop this function since the errors above are ignored anyway

packages/beacon-node/src/util/sszBytes.ts
twoeths: it needs to do after gossip snappy decode step, I think we should do at Lodestar side since we have `sszType.minSize`, `gossipsub.onGossipsubMessage` would be a good place

> return null here and document in the JSDocs that null means data too short.

yeah that's good to do 👍

twoeths: just checked our `DataTransrormSnappy`, it's convenient to check ssz type `minSize/maxSize` there

dapplion (approved):
LGTM

wemeetagain:

PR #5328, opened 2023-03-31
Enable metrics if monitoring endpoint is configured

nflaig:
**Motivation**

Improve monitoring UX by not requiring users to also supply `--metrics` flag. If users only want to use monitoring via remote service and do not have a local prometheus/grafana stack it is not required to start the metrics server.

**Description**

Monitoring relies on metrics data to work properly and as of now would throw an error if `--metrics` flag is not supplied. This change implicitly enables metrics if a monitoring endpoint is configured.

**Note:** metrics server will only be started if metrics are explicitly enabled


wemeetagain (approved):
LGTM. Good job not starting the metrics server unless actually required.

wemeetagain:

PR #5329, opened 2023-03-31
Reference gh issue with rationale for choosing vc http timeout

### LINKED ISSUE #5315: Default validator HTTP request timeout is too short
At the moment all HTTP requests done by the VC have a timeout of `SECONDS_PER_SLOT` (12 seconds).

https://github.com/ChainSafe/lodestar/blob/f136f952ddb861976b3cc9484914c49a06ee8d90/packages/validator/src/validator.ts#L83-L84

Most of the time the HTTP request timeout does not matter at all as we expect that the BN responds within few milliseconds but there might be unexpected network issues and we should account for this.

It should be allowed to take longer than the current slot for attestation tasks as late submission of attestations is still acceptable, it will just not give as good rewards but it is still better than completely missing the attestation due to the timeout.
See [Rewards and Penalties](https://kb.beaconcha.in/rewards-and-penalties).

The current timeout also does not really make sense, if a HTTP request is sent at 2/3 of the slot to `getAggregatedAttestation` it would mean the timeout happens after the slot is already over (20 seconds).

**Current behavior**

VC has a default HTTP timeout of `SECONDS_PER_SLOT` which could potentially cause missed attestations.

**Expected behavior**

VC should account for network issues and have a higher timeout, potentially something like `SECONDS_PER_SLOT * 5` (60 seconds). Timeout should be chosen based on [Attestation Reward Scale](https://www.attestant.io/posts/defining-attestation-effectiveness/) and technical implications of keeping the HTTP request open for too long.

Related https://github.com/ChainSafe/lodestar/issues/5314

nflaig:
Closes https://github.com/ChainSafe/lodestar/issues/5315

### DIFF
--- a/packages/validator/src/validator.ts
+++ b/packages/validator/src/validator.ts
@@ -81,6 +81,7 @@ export class Validator {
         {
           urls: typeof opts.api === "string" ? [opts.api] : opts.api,
           // Validator would need the beacon to respond within the slot
+          // See https://github.com/ChainSafe/lodestar/issues/5315 for rationale
           timeoutMs: config.SECONDS_PER_SLOT * 1000,
           getAbortSignal: () => this.controller.signal,
         },


wemeetagain (approved):

wemeetagain:

PR #5333, opened 2023-04-03
Cache gossip validation by attestation data 

twoeths:
**Motivation**

- `validateGossipAttestation` takes almost 12% on mainnet node with new gossip queues because we process all gossip attestations now
  - 5.8% for `computeSiningRoot()`
  - 3.9% for `validateHeadBlockAndTargetRoot()`

notice that a lot of time the above apis run with same attestation data

see https://github.com/ChainSafe/lodestar/issues/5247#issuecomment-1491175978

**Description**

Implement `SeenAttestationDatas` cache to avoid running the above apis with same data
- Key as attestation data hash `Buffer.from(attestationDataSerializedBytes).toString("base64")`, this is cheap enough as shown in benchmark
- value as
  - index2pubkey to avoid recomputing signing root, shared acoss application so does not take memory
  - committeeIndices to avoid regenerating state, part of shuffling data so does not take memory
  - signingRoot: to avoid recomputing signing root
  - attDataRootHex: to be used for forkchoice (right now we always do that for all attestations in forkchoice)
  - subnet: to be compliant to the current flow
- Bound the cache as 200 max size per slot (assuming, we have 3 variants per committees, with 64 committee indices per slot)
  - metrics show that this is a good number
  - Memory cost is less than 300kb per slot, for mainnet usually it takes only 150kb per slot
- Number of slots to cache is configurable, default to 3 which take < 1MB which is affordable to us

**Test Result**
- `validateGossipAttestation()` time is reduced to less than 3%. If we want to do full cache of attestation, which is 32 slots, we can reduce more
- Hit %
Screenshot 2023-04-03 at 11 13 28


dapplion (commented):
This optimization makes a lot of sense I think the tradeoffs are pretty good :+1: 

Implementation is solid, with bounded data structures and metrics. If CI passes, LGTM to merge

dapplion (approved):

wemeetagain:

Issue #1773, opened 2020-11-20
Enforce Conventional Commits


wemeetagain:
There are some possible benefits to using/enforcing conventional commits. Pls discuss

Github action to enforce?

dapplion:
In meetings it was agreed to not enforce at the commit level. Maybe at the PR title level (see https://github.com/ChainSafe/lodestar/issues/1812), but with our regular bi-weekly release schedule and not yet following semver's breaking policy it's not necessary

philknows:
Reopened based on our Standup call from March 28. We would like to have better and more human-readable updates to our releases. A manual release summary and enforcing conventional commits on our monorepo should better organize our changelog and make it more useful for users. 

We should:
- [ ] Add a Github action to enforce 
- [ ] Update our `contributions.md` to reflect this change

# Quick examples
* `feat: new feature`
* `fix(scope): bug in scope`
* `feat!: breaking change` / `feat(scope)!: rework API`
* `chore(deps): update dependencies`

# Commit types
* `build`: Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)
* `ci`: Changes to CI configuration files and scripts (example scopes: Travis, Circle, BrowserStack, SauceLabs)
* **`chore`: Changes which doesn't change source code or tests e.g. changes to the build process, auxiliary tools, libraries**
* `docs`: Documentation only changes
* **`feat`: A new feature**
* **`fix`: A bug fix**
* `perf`: A code change that improves performance
* `refactor`:  A code change that neither fixes a bug nor adds a feature
* `revert`: Revert something
* `style`: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
* `test`: Adding missing tests or correcting existing tests

More information on the spec here: https://www.conventionalcommits.org/en/v1.0.0/#specification

maschad:
Thanks for reopening this @philknows ! It will really help with standardizing our changes. A few points:

- I think we can distinguish deps a bit more to have it as it's own commit types i.e. `deps (dev):` or `deps!:` if upgrading the dep introduces a breaking change.
- We should ensure we outline all aspects of the [specification](https://www.conventionalcommits.org/en/v1.0.0/#specification) that we want to enforce.

We can setup a [husky pre-commit hook](https://github.com/typicode/husky) to enforce this once we agree upon it.

nflaig:
> deps!: if upgrading the dep introduces a breaking change.

[...]

PR #5342, opened 2023-04-04
Add conventional commit PR title checker

wemeetagain:
**Motivation**

See #1773 

**Description**

Add github action to enforce conventional commits on PR titles.
This is the same action currently configured in the ssz repo, although this one is configured slightly differently.

Please give feedback on the current configuration.

### DIFF
--- a/.github/workflows/lint-pr-title.yml
+++ b/.github/workflows/lint-pr-title.yml
@@ -1,0 +1,59 @@
+name: "Lint PR title"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+      - labeled
+      - unlabeled
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # Configure which types are allowed (newline-delimited).
+          # Default: https://github.com/commitizen/conventional-commit-types
+          #types:
+
+          # Configure which scopes are allowed (newline-delimited).
+          # These are regex patterns auto-wrapped in `^ $`.
+          scopes: |
+            \d+
+
+          # Configure that a scope must always be provided.
+          requireScope: true
+
+          # Configure which scopes are disallowed in PR titles (newline-delimited).
+          # For instance by setting the value below, `chore(release): ...` (lowercase)
+          # and `ci(e2e,release): ...` (unknown scope) will be rejected.
+          # These are regex patterns auto-wrapped in `^ $`.
+          disallowScopes: |
+            release
+            [A-Z]+
+
+          # Configure additional validation for the subject based on a regex.
+          # This example ensures the subject doesn't start with an uppercase character.
+          subjectPattern: ^(?![A-Z]).+$
+
+          # If `subjectPattern` is configured, you can use this property to override
+          # the default error message that is shown when the pattern doesn't match.
+          # The variables `subject` and `title` can be used within the message.
+          subjectPatternError: |
+            The subject "{subject}" found in the pull request title "{title}"
+            didn't match the configured pattern. Please ensure that the subject
+            doesn't start with an uppercase character.
+
+          # If the PR contains one of these newline-delimited labels, the
+          # validation is skipped. If you want to rerun the validation when
+          # labels change, you might want to use the `labeled` and `unlabeled`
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-pr-title.yml
```diff
@@ -0,0 +1,59 @@
...
+          # These are regex patterns auto-wrapped in `^ $`.
+          scopes: |
+            \d+
+
+          # Configure that a scope must always be provided.
+          requireScope: true
+
+          # Configure which scopes are disallowed in PR titles (newline-delimited).
+          # For instance by setting the value below, `chore(release): ...` (lowercase)
+          # and `ci(e2e,release): ...` (unknown scope) will be rejected.
```
nflaig, line 34: I am not sure what this comment it trying to say

all of these `chore(release): ...` and `ci(e2e,release): ...` will be rejected or just `(unknown scope)`?

based on the regex all of these are allowed because do not match `release` and no uppercase letter

.github/workflows/lint-pr-title.yml
```diff
@@ -0,0 +1,59 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # Configure which types are allowed (newline-delimited).
+          # Default: https://github.com/commitizen/conventional-commit-types
+          #types:
+
+          # Configure which scopes are allowed (newline-delimited).
+          # These are regex patterns auto-wrapped in `^ $`.
+          scopes: |
+            \d+
```
nflaig: this would mean scopes need one or more digits?

wemeetagain: Whoops, good catch. Not sure what should go here...

.github/workflows/lint-pr-title.yml
wemeetagain: Yeah the comment is a bit confusing. I'd just copied it from the readme.
I think it means that any message with a scope that is either "release" or has upper-case letters would be disallowed.

.github/workflows/lint-pr-title.yml
nflaig: not sure either, if we want to support multiple scopes like `ci(e2e,release)` is also questionable

`requireScope` is set to `true` right now so would always have to provide one, what if we do a global lint change or anything similar, just provide `*` as scope?

could also be strict, only numbers and lowercase `^[a-z0-9]+$`

wemeetagain: I think requiring a scope could be a good thing for consistency, but maybe we don't need to restrict the scope options, instead rely on maintainer discretion.

philknows: Personally a big fan of simplicity and minimizing/not requiring the scope. I don't think there is a huge benefit for complication and subjectivity if something falls into multiple scopes... the scope labels are already hard enough sometimes.

philknows:
We should close up #5299 and also modify https://github.com/ChainSafe/lodestar/blob/unstable/CONTRIBUTING.md?plain=1#L95 to add in the new standard defined by this github action.

dapplion:
No conventional commit PR title for this PR?

.github/workflows/lint-pr-title.yml
nflaig: there might be commits where a scope will be hard to define (e.g anything related to general docs) so I would also vote for making the scope optional

wemeetagain:
Ready for rereview

.github/workflows/lint-pr-title.yml
matthewkeil: I also agree with optional scopes. PRs don't often fit entirely into one package or "snippet".  Sometimes it is clear but often its muddy for what should be a very short string

.github/workflows/lint-pr-title.yml
```diff
@@ -0,0 +1,56 @@
+name: "Lint PR title"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+      - labeled
+      - unlabeled
```
matthewkeil, line 10: Curious why you are triggering for labels

matthewkeil (approved):

.github/workflows/lint-pr-title.yml
wemeetagain: We can disable/enable this check with the `ignore-semantic-pull-request` label.

Happy to remove if we don't think its worth it.

matthewkeil: thats a pretty neat feature actually!!

wemeetagain:

PR #5348, opened 2023-04-06
Add note about setting default nodejs version when using nvm

nflaig:
**Motivation**

nvm sets the node version to default after reboot which might break Lodestar because modules are compiled on a different node version during `yarn install`, see [discord](https://discord.com/channels/593655374469660673/743858262864167062/1093330636565131325).

**Description**

Add note to source installation page that a default version via `nvm alias default <version> && nvm use default` should bet set to ensure the node version does not change after reboot.


wemeetagain (approved):

wemeetagain:

PR #5351, opened 2023-04-07
refactor!: move network logic into NetworkCore

wemeetagain:
**Motivation**

In anticipation of moving the network to a separate thread, move all network functionality to a separate module with an interface that may be implemented as a worker thread.

**Description**

In anticipation of moving the network to a separate thread, move all network functionality to a separate module with an interface that may be implemented as a worker thread.

Pulled into separate PRs
- Move all network-related metrics to a separate metrics object controlled by (#5445)
- Refactor MetadataController without chain dependency (#5443)
- Turn `LocalClock` into an `EventEmitter` that emits `ClockEvent`s, Allows the `NetworkCore` to rely only on a clock for clock events without using the chain directoy. This will be used by the threaded implementation to instantiate its own clock. (#5446)
- Separate gossip publishing from the gossipsub implementation, will be used to serialize gossip objects on the chain thread. (#5450)
- Rely on a `StatusCache` instead of the full chain to allow `NetworkCore` to get the status without the chain. (#5451)

Important points
- Move the core network logic into `NetworkCore` which implements `INetworkCore`
- Track connected peers in the `Network` implementation to allow for synchronous `getConnectedPeers` and `getConnectedPeerCount`.
- Track `SimpleAttnetsService` info to allow for synchronous `shouldProcessAttestation` during gossip and api attestation processing
- A few renamings `networkEventBus`->`network.events`, `prepareBeaconCommitteeSubnet`->`prepareBeaconCommitteeSubnets`


packages/beacon-node/src/network/subnets/attnetsService.ts
```diff
@@ -379,3 +392,26 @@ export class AttnetsService implements IAttnetsService {
     metrics.attnetsService.aggregatorSlotSubnetCount.set(aggregatorCount);
   }
 }
+
+/**
+ * Some attnet subscription functionality is needed to be mirrored on the chain side of the thread boundary to determine if attestations should be processed
+ */
+export class SimpleAttnetsService {
```
dapplion, line 399: Why is this necessary? This data will be send from the worker to the main thread? IMO:
- Main thread receives validator subscription requests via rest API
- Main thread forward sends those network worker subscription updates
- Worker tracks internal state on what to process and what to subscribe to
- Worker emits gossip object to main thread along with a flag "should_process" or "process_before_slot"

dapplion (commented):
Very good direction, nice spin off from the main thread PR to lay some groundwork

packages/beacon-node/src/network/subnets/attnetsService.ts
wemeetagain: Also is used when submitting attestations via API, where a process_before_slot is not available

wemeetagain:
Converting back to draft
will pull out all major features into smaller PRs before reopening for review

PR #5354, opened 2023-04-10
Fix network processor jobsSubmitted metric

twoeths:
**Motivation**

The network processor `jobsSubmitted` metric was monitored as 1 while debug showed that it's usually 128 if there are jobs submitted

**Description**

- Only record `jobsSubmitted` if there are really jobs submitted
- a follow up of #5350

### DIFF
--- a/packages/beacon-node/src/network/processor/index.ts
+++ b/packages/beacon-node/src/network/processor/index.ts
@@ -292,6 +292,8 @@ export class NetworkProcessor {
       break;
     }
 
-    this.metrics?.networkProcessor.jobsSubmitted.observe(jobsSubmitted);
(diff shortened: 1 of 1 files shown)


dapplion (commented):
What's the problem with observing the case of no jobs submitted?

twoeths:
> What's the problem with observing the case of no jobs submitted?

@dapplion  most of the time it does not accept work, there is already `canNotAcceptWork` metric for that, I only want to monitor non-zero `jobsSubmitted`

Screenshot 2023-04-10 at 19 58 57

dapplion (approved):

wemeetagain:

Issue #5353, opened 2023-04-09
Dropped 100% sync committee message and sync committee contribution and proof


twoeths:
**Describe the bug**

This is from a test mainnet node subscribing to all subnets.

After #5350 after a while when there are 30% of dropped attestations, 100% of sync committee messages and sync contribution and proofs are dropped since then and node is not able to recover (after I restart it gets back to 0% drop)

<img width="804" alt="Screenshot 2023-04-09 at 17 27 44" src="https://user-images.githubusercontent.com/10568965/230767563-d85bbd70-6799-467b-b767-591fe79c2002.png">

while beacon_attestation topic is consistently at around its max size

<img width="791" alt="Screenshot 2023-04-09 at 17 29 49" src="https://user-images.githubusercontent.com/10568965/230767647-2c7effc5-69a7-4b13-9b09-31a575bb7fcc.png">

**Expected behavior**

Right now when adding a gossip message, we delete exactly 1 gossip message.
As we have to drop 30% - 40% of attestations anyway, how about drop 30% - 40% of attestation messages when queue is at max size to allow the node to recover @dapplion ?

twoeths:
this is stable test mainnet node v1.10.0

<img width="1292" alt="Screenshot 2023-08-20 at 13 22 15" src="https://github.com/ChainSafe/lodestar/assets/10568965/73413221-4bea-48c5-be10-bb25e55f3826">

PR #5355, opened 2023-04-11
Gossip queue to drop by ratio

twoeths:
**Motivation**

- With mainnet node, after a while node has to drop 30% - 40% attestations consistently and other lower priority queues are never processed, they are all dropped 100%
- Right now after add an item to queue, if it's >= maxLength we drop 1, this keeps the node busy, queue length = max length all the time and node is not able to recover

**Description**

The current strategy is considered "drop by count" strategy with `count=1`. This PR implements a new "drop by ratio" strategy:
- With `start / step` as its params, applied for `attestation` topic only, other topics use the same `drop by count` strategy (unchanged)
- Add a `dropRatio` initialized with `start`
- If node is overload, i.e queue hit max length, drop by `dropRatio`, and increase it with `step`, bound by `MAX_DROP_RATIO`
- If node is in good status, i.e queue has 0 item and not a recent drop, reset `dropRatio`

part of #5353

**Testing**
- The initial test shows that `dropRatio` hit `MAX_DROP_RATIO` after 10h-12h and stay that high but even with that, the total drop of attestation is only around 10% (as compared to 30% - 40%), and other lower priority queues have some chance to run (instead of having to drop 100% of them like currently)
- Need to continue monitor the latest commit

### DIFF
--- a/packages/beacon-node/src/metrics/metrics/lodestar.ts
+++ b/packages/beacon-node/src/metrics/metrics/lodestar.ts
@@ -248,6 +248,12 @@ export function createLodestarMetrics(
       help: "Count of total gossip validation queue length",
       labelNames: ["topic"],
     }),
+
+    gossipValidationQueueDropRatio: register.gauge<"topic">({
+      name: "lodestar_gossip_validation_queue_current_drop_ratio",
+      help: "Current drop ratio of gossip validation queue",
+      labelNames: ["topic"],
+    }),
     gossipValidationQueueDroppedJobs: register.gauge<"topic">({
       name: "lodestar_gossip_validation_queue_dropped_jobs_total",
       help: "Count of total gossip validation queue dropped jobs",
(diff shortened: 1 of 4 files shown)


twoeths:
- With drop by ratio (applied for `beac_attestation` topic only), even through the number of gossip messages are dropped more each time, the total attestations dropped are fewer than `unstable`. Also sync_committee messages are not consistently dropped (as 100% in unstable, although it's not perfect, still need to work on this)

feat1
Screenshot 2023-04-13 at 08 44 57

vs. unstable (note that there are more dropped attestations and almost 100% sync_committee messages are dropped consistently)
Screenshot 2023-04-13 at 08 45 43

- Drop Ratio metric is increased gradually each time and reset when node status is good
Screenshot 2023-04-13 at 08 48 42

dapplion (approved):
Happy to merge this but would really like to focus on getting those queue lengths down :pray:

- With more PRs to do less work like https://github.com/ChainSafe/lodestar/pull/5363

wemeetagain (approved):

wemeetagain:

PR #5357, opened 2023-04-11
Decrypt keystores in a thread pool

### LINKED ISSUE #4179: Optimize UX loading +100 keystores on vc

wemeetagain:
**Motivation**

Resolves https://github.com/ChainSafe/lodestar/issues/4179

**Description**

- Refactor the `decryptKeystoreDefinitions` function
- Use threads library worker pool implementation
- poolSize code copied from bls worker thread impl

### DIFF
--- a/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
+++ b/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
@@ -54,7 +54,7 @@ export async function decryptKeystoreDefinitions(
       }
     }
 
-    const task = pool.queue((thread) => thread.decryptKeystoreDefinition(definition, Boolean(opts.ignoreLockFile)));
+    const task = pool.queue((thread) => thread.decryptKeystoreDefinition(definition));
     tasks.push(task);
     task
       .then((secretKeyBytes: Uint8Array) => {
--- a/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
+++ b/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
@@ -28,8 +28,8 @@ export async function decryptKeystoreDefinitions(
     }
   }
 
-  const signers: SignerLocal[] = [];
-  const passwords: string[] = [];
+  const signers = new Array(keystoreDefinitions.length) as SignerLocal[];
+  const passwords = new Array(keystoreDefinitions.length) as string[];
   const tasks: QueuedTask<ModuleThread<DecryptKeystoreWorkerAPI>, Uint8Array>[] = [];
   const errors: Error[] = [];
   const pool = Pool(
@@ -63,8 +63,8 @@ export async function decryptKeystoreDefinitions(
           secretKey: bls.SecretKey.fromBytes(secretKeyBytes),
         };
 
-        signers.push(signer);
-        passwords.push(definition.password);
+        signers[index] = signer;
+        passwords[index] = definition.password;
 
         if (opts?.onDecrypt) {
           opts?.onDecrypt(index);
--- a/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
+++ b/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
@@ -46,7 +46,7 @@ export async function decryptKeystoreDefinitions(
       lockFilepath(definition.keystorePath);
     } catch (e) {
       if (opts.ignoreLockFile) {
-        opts.logger.warn("Keystore forcefully loaded even though lockfile could not be acquired", {
+        opts.logger.warn("Keystore forcefully loaded even though lockfile exists", {
           path: definition.keystorePath,
         });
       } else {
--- a/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
+++ b/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
@@ -1,0 +1,94 @@
+import {spawn, ModuleThread, Pool, QueuedTask, Worker} from "@chainsafe/threads";
+import {Signer, SignerLocal, SignerType} from "@lodestar/validator";
+import bls from "@chainsafe/bls";
+import {LocalKeystoreDefinition} from "../interface.js";
+import {clearKeystoreCache, loadKeystoreCache, writeKeystoreCache} from "../keystoreCache.js";
+import {lockFilepath, unlockFilepath} from "../../../../util/lockfile.js";
+import {defaultPoolSize} from "./poolSize.js";
+import {DecryptKeystoreWorkerAPI, KeystoreDecryptOptions} from "./types.js";
+
+/**
+ * Decrypt keystore definitions using a threadpool
+ */
+export async function decryptKeystoreDefinitions(
+  keystoreDefinitions: LocalKeystoreDefinition[],
+  opts: KeystoreDecryptOptions
+): Promise<Signer[]> {
+  if (opts.cacheFilePath) {
+    try {
+      const signers = await loadKeystoreCache(opts.cacheFilePath, keystoreDefinitions);
+      if (opts?.onDecrypt) {
+        opts?.onDecrypt(signers.length - 1);
+      }
+      return signers;
+    } catch {
+      // Some error loading the cache, ignore and invalidate cache
+      await clearKeystoreCache(opts.cacheFilePath);
+    }
+  }
+
+  const signers: SignerLocal[] = [];
+  const passwords: string[] = [];
+  const tasks: QueuedTask<ModuleThread<DecryptKeystoreWorkerAPI>, Uint8Array>[] = [];
+  const errors: Error[] = [];
+
+  const pool = Pool(() => spawn<DecryptKeystoreWorkerAPI>(new Worker("./worker.js")), defaultPoolSize);
+  for (const [index, definition] of keystoreDefinitions.entries()) {
+    try {
+      lockFilepath(definition.keystorePath);
+    } catch (e) {
+      if (opts.force) {
+        // Ignore error, maybe log?
+      } else {
+        throw e;
+      }
+    }
+
+    const task = pool.queue((thread) => thread.decryptKeystoreDefinition(definition, Boolean(opts.force)));
+    tasks.push(task);
+    task
+      .then((secretKeyBytes: Uint8Array) => {
+        const signer: SignerLocal = {
+          type: SignerType.Local,
+          secretKey: bls.SecretKey.fromBytes(secretKeyBytes),
+        };
+
+        signers.push(signer);
+        passwords.push(definition.password);
+
+        if (opts?.onDecrypt) {
+          opts?.onDecrypt(index);
+        }
+      })
+      .catch((e: Error) => {
+        // In-progress tasks can't be canceled, so there's a chance that multiple errors may be caught
+        // add to the list of errors
+        errors.push(e);
+        // cancel all pending tasks, no need to continue decrypting after we hit one error
+        for (const task of tasks) {
+          task.cancel();
+        }
+      });
+  }
+
+  try {
+    // only resolves if there are no errored tasks
+    await pool.completed();
+  } catch (e) {
+    // If an error occurs, the program isn't going to be running,
+    // so we should unlock all lockfiles we created
+    for (const {keystorePath} of keystoreDefinitions) {
+      unlockFilepath(keystorePath);
+    }
+
+    throw new AggregateError(errors);
+  } finally {
+    await pool.terminate();
+  }
+
+  if (opts.cacheFilePath) {
+    await writeKeystoreCache(opts.cacheFilePath, signers, passwords);
+  }
+
+  return signers;
+}
--- a/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
+++ b/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
@@ -4,7 +4,7 @@ import bls from "@chainsafe/bls";
 import {LocalKeystoreDefinition} from "../interface.js";
 import {clearKeystoreCache, loadKeystoreCache, writeKeystoreCache} from "../keystoreCache.js";
 import {lockFilepath, unlockFilepath} from "../../../../util/lockfile.js";
-import {defaultPoolSize} from "./poolSize.js";
+import {calculateThreadpoolConcurrency} from "./poolSize.js";
 import {DecryptKeystoreWorkerAPI, KeystoreDecryptOptions} from "./types.js";
 
 /**
@@ -32,6 +32,7 @@ export async function decryptKeystoreDefinitions(
   const passwords = new Array(keystoreDefinitions.length) as string[];
   const tasks: QueuedTask<ModuleThread<DecryptKeystoreWorkerAPI>, Uint8Array>[] = [];
   const errors: Error[] = [];
+  const {numWorkers, jobConcurrency} = calculateThreadpoolConcurrency();
   const pool = Pool(
     () =>
       spawn<DecryptKeystoreWorkerAPI>(new Worker("./worker.js"), {
@@ -39,7 +40,10 @@ export async function decryptKeystoreDefinitions(
         // the initialization to timeout. The number below is big enough to almost disable the timeout
         timeout: 5 * 60 * 1000,
       }),
-    defaultPoolSize
+    {
+      concurrency: jobConcurrency,
+      size: numWorkers,
+    }
   );
   for (const [index, definition] of keystoreDefinitions.entries()) {
     try {
--- a/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
+++ b/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
@@ -4,7 +4,7 @@ import bls from "@chainsafe/bls";
 import {LocalKeystoreDefinition} from "../interface.js";
 import {clearKeystoreCache, loadKeystoreCache, writeKeystoreCache} from "../keystoreCache.js";
 import {lockFilepath, unlockFilepath} from "../../../../util/lockfile.js";
-import {calculateThreadpoolConcurrency} from "./poolSize.js";
+import {defaultPoolSize} from "./poolSize.js";
 import {DecryptKeystoreWorkerAPI, KeystoreDecryptOptions} from "./types.js";
 
 /**
@@ -32,7 +32,6 @@ export async function decryptKeystoreDefinitions(
   const passwords = new Array(keystoreDefinitions.length) as string[];
   const tasks: QueuedTask<ModuleThread<DecryptKeystoreWorkerAPI>, Uint8Array>[] = [];
   const errors: Error[] = [];
-  const {numWorkers, jobConcurrency} = calculateThreadpoolConcurrency();
   const pool = Pool(
     () =>
       spawn<DecryptKeystoreWorkerAPI>(new Worker("./worker.js"), {
@@ -40,10 +39,7 @@ export async function decryptKeystoreDefinitions(
         // the initialization to timeout. The number below is big enough to almost disable the timeout
         timeout: 5 * 60 * 1000,
       }),
-    {
-      concurrency: jobConcurrency,
-      size: numWorkers,
-    }
+    defaultPoolSize
   );
   for (const [index, definition] of keystoreDefinitions.entries()) {
     try {
--- a/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
+++ b/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
@@ -45,7 +45,7 @@ export async function decryptKeystoreDefinitions(
     try {
       lockFilepath(definition.keystorePath);
     } catch (e) {
-      if (opts.force) {
+      if (opts.ignoreLockFile) {
         opts.logger.warn("Keystore forcefully loaded even though lockfile could not be acquired", {
           path: definition.keystorePath,
         });
@@ -54,7 +54,7 @@ export async function decryptKeystoreDefinitions(
       }
     }
 
-    const task = pool.queue((thread) => thread.decryptKeystoreDefinition(definition, Boolean(opts.force)));
+    const task = pool.queue((thread) => thread.decryptKeystoreDefinition(definition, Boolean(opts.ignoreLockFile)));
     tasks.push(task);
     task
       .then((secretKeyBytes: Uint8Array) => {
--- a/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
+++ b/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
@@ -20,6 +20,7 @@ export async function decryptKeystoreDefinitions(
       if (opts?.onDecrypt) {
         opts?.onDecrypt(signers.length - 1);
       }
+      opts.logger.debug("Loaded keystores via keystore cache");
       return signers;
     } catch {
       // Some error loading the cache, ignore and invalidate cache
@@ -45,7 +46,9 @@ export async function decryptKeystoreDefinitions(
       lockFilepath(definition.keystorePath);
     } catch (e) {
       if (opts.force) {
-        // Ignore error, maybe log?
+        opts.logger.warn("Keystore forcefully loaded even though lockfile could not be acquired", {
+          path: definition.keystorePath,
+        });
       } else {
         throw e;
       }
@@ -95,6 +98,7 @@ export async function decryptKeystoreDefinitions(
 
   if (opts.cacheFilePath) {
     await writeKeystoreCache(opts.cacheFilePath, signers, passwords);
+    opts.logger.debug("Written keystores to keystore cache");
   }
 
   return signers;
--- a/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
+++ b/packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
@@ -1,5 +1,5 @@
 import {spawn, ModuleThread, Pool, QueuedTask, Worker} from "@chainsafe/threads";
-import {Signer, SignerLocal, SignerType} from "@lodestar/validator";
+import {SignerLocal, SignerType} from "@lodestar/validator";
 import bls from "@chainsafe/bls";
 import {LocalKeystoreDefinition} from "../interface.js";
 import {clearKeystoreCache, loadKeystoreCache, writeKeystoreCache} from "../keystoreCache.js";
@@ -13,7 +13,7 @@ import {DecryptKeystoreWorkerAPI, KeystoreDecryptOptions} from "./types.js";
 export async function decryptKeystoreDefinitions(
   keystoreDefinitions: LocalKeystoreDefinition[],
   opts: KeystoreDecryptOptions
-): Promise<Signer[]> {
+): Promise<SignerLocal[]> {
   if (opts.cacheFilePath) {
     try {
       const signers = await loadKeystoreCache(opts.cacheFilePath, keystoreDefinitions);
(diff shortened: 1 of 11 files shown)


packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
```diff
@@ -0,0 +1,94 @@
...
+  const tasks: QueuedTask<ModuleThread<DecryptKeystoreWorkerAPI>, Uint8Array>[] = [];
+  const errors: Error[] = [];
+
+  const pool = Pool(() => spawn<DecryptKeystoreWorkerAPI>(new Worker("./worker.js")), defaultPoolSize);
+  for (const [index, definition] of keystoreDefinitions.entries()) {
+    try {
+      lockFilepath(definition.keystorePath);
+    } catch (e) {
+      if (opts.force) {
+        // Ignore error, maybe log?
```
dapplion: Yes we should log

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
```diff
@@ -0,0 +1,94 @@
...
+        // cancel all pending tasks, no need to continue decrypting after we hit one error
+        for (const task of tasks) {
+          task.cancel();
+        }
+      });
+  }
+
+  try {
+    // only resolves if there are no errored tasks
+    await pool.completed();
```
dapplion: What happens if the pool completes before all tasks are done? We should have some sanity checks on done

dapplion (dismissed):
Overall looks good, some minor comments

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
wemeetagain: The pool completes when all enqueued tasks resolve (and throws if tasks reject). Since the tasks are enqueued synchronously before this, it isn't possible to complete before all tasks are done.

nflaig:
the results importing 2000 keys on

```
CPU(s):                          12
Thread(s) per core:              2
Core(s) per socket:              6
Socket(s):                       1
```


**unstable:** 19.7 keys/m

```
Apr-15 13:43:59.725[]                 info: Lodestar network=goerli, version=v1.6.0/9eb2ca3, commit=9eb2ca36ad4f12fa7cf39cffd96a33e41a55bf4b
Apr-15 13:43:59.727[]                 info: Connecting to LevelDB database path=/home/devops/goerli/data/validator/validator-db
Apr-15 13:44:09.763[]                 info: 0% of keystores imported. current=1 total=2000 rate=6.00keys/m
Apr-15 13:44:19.763[]                 info: 0% of keystores imported. current=4 total=2000 rate=18.00keys/m
Apr-15 13:44:29.764[]                 info: 0% of keystores imported. current=8 total=2000 rate=24.00keys/m
Apr-15 13:44:39.763[]                 info: 1% of keystores imported. current=11 total=2000 rate=18.00keys/m
Apr-15 13:44:49.763[]                 info: 1% of keystores imported. current=14 total=2000 rate=18.00keys/m
Apr-15 13:44:59.764[]                 info: 1% of keystores imported. current=18 total=2000 rate=24.00keys/m
Apr-15 13:45:09.764[]                 info: 1% of keystores imported. current=21 total=2000 rate=18.00keys/m
Apr-15 13:45:19.763[]                 info: 1% of keystores imported. current=24 total=2000 rate=18.00keys/m
Apr-15 13:45:29.764[]                 info: 1% of keystores imported. current=27 total=2000 rate=18.00keys/m
Apr-15 13:45:39.764[]                 info: 2% of keystores imported. current=31 total=2000 rate=24.00keys/m
Apr-15 13:45:49.763[]                 info: 2% of keystores imported. current=34 total=2000 rate=18.00keys/m
Apr-15 13:45:59.763[]                 info: 2% of keystores imported. current=37 total=2000 rate=18.00keys/m
Apr-15 13:46:09.764[]                 info: 2% of keystores imported. current=40 total=2000 rate=18.00keys/m
Apr-15 13:46:19.764[]                 info: 2% of keystores imported. current=44 total=2000 rate=24.00keys/m
Apr-15 13:46:29.763[]                 info: 2% of keystores imported. current=47 total=2000 rate=18.00keys/m
```

**with transferable objects:** 100.5 keys/m

```
Apr-15 13:48:07.727[]                 info: Lodestar network=goerli, version=v1.6.0/cayman/parallel-keystore-decrypt/0096e30, commit=0096e304ac3333a291d28ff2f40b19deb948ff88
Apr-15 13:48:07.729[]                 info: Connecting to LevelDB database path=/home/devops/goerli/data/validator/validator-db
Apr-15 13:48:17.753[]                 info: 100% of keystores imported. current=0 total=2000 rate=0.00keys/m
Apr-15 13:48:27.755[]                 info: 1% of keystores imported. current=24 total=2000 rate=143.97keys/m
Apr-15 13:48:37.757[]                 info: 2% of keystores imported. current=36 total=2000 rate=71.99keys/m
Apr-15 13:48:47.756[]                 info: 3% of keystores imported. current=56 total=2000 rate=120.01keys/m
Apr-15 13:48:57.757[]                 info: 4% of keystores imported. current=71 total=2000 rate=89.99keys/m
Apr-15 13:49:07.756[]                 info: 4% of keystores imported. current=88 total=2000 rate=102.01keys/m
Apr-15 13:49:17.758[]                 info: 5% of keystores imported. current=108 total=2000 rate=119.98keys/m
Apr-15 13:49:27.761[]                 info: 6% of keystores imported. current=117 total=2000 rate=53.98keys/m
Apr-15 13:49:37.761[]                 info: 7% of keystores imported. current=143 total=2000 rate=156.00keys/m
Apr-15 13:49:47.761[]                 info: 8% of keystores imported. current=155 total=2000 rate=72.00keys/m
Apr-15 13:49:57.761[]                 info: 9% of keystores imported. current=179 total=2000 rate=144.00keys/m
```

**without transferable objects:** 102.2 keys/m

```
Apr-15 13:51:55.473[]                 info: Lodestar network=goerli, version=v1.6.0/65a8668, commit=65a86684c54005c935e99e5eeab7497801dab1d7
Apr-15 13:51:55.476[]                 info: Connecting to LevelDB database path=/home/devops/goerli/data/validator/validator-db
Apr-15 13:52:05.507[]                 info: 100% of keystores imported. current=0 total=2000 rate=0.00keys/m
Apr-15 13:52:15.507[]                 info: 1% of keystores imported. current=24 total=2000 rate=144.00keys/m
Apr-15 13:52:25.509[]                 info: 2% of keystores imported. current=36 total=2000 rate=71.99keys/m
Apr-15 13:52:35.508[]                 info: 3% of keystores imported. current=60 total=2000 rate=144.01keys/m
Apr-15 13:52:45.509[]                 info: 4% of keystores imported. current=72 total=2000 rate=71.99keys/m
Apr-15 13:52:55.509[]                 info: 5% of keystores imported. current=91 total=2000 rate=114.01keys/m
Apr-15 13:53:05.521[]                 info: 5% of keystores imported. current=108 total=2000 rate=101.87keys/m
Apr-15 13:53:15.520[]                 info: 6% of keystores imported. current=122 total=2000 rate=84.01keys/m
Apr-15 13:53:25.521[]                 info: 7% of keystores imported. current=141 total=2000 rate=113.99keys/m
Apr-15 13:53:35.521[]                 info: 8% of keystores imported. current=167 total=2000 rate=156.00keys/m
Apr-15 13:53:45.523[]                 info: 9% of keystores imported. current=178 total=2000 rate=65.99keys/m
```

nflaig (commented):
Issues found during testing

- incorrect initial log `info: 100% of keystores imported. current=0 total=2000 rate=0.00keys/m`, should show 0 instead of 100%
- keystore cache does not seem to work, the file is created after import but on next startup is removed and keys are loaded from scratch

wemeetagain:
I think we can do better than 1 decryption per thread at a time.

ChatGPT was helpful here:

--

The memory usage of Scrypt depends on the specific parameters chosen. These parameters include:

    N: The CPU/memory cost factor, which must be a power of 2 (e.g., 2^14)
    r: The block size (typically 8)
    p: The parallelization factor (typically 1)

The memory usage can be roughly calculated with the following formula:

Memory Usage = 128 * N * r * p bytes

For example, using the parameters N=2^14 (16,384), r=8, and p=1, the memory usage would be:

Memory Usage = 128 * 16,384 * 8 * 1 = 16,777,216 bytes, or approximately 16 MB.

-- 

The default used by the deposit cli is here
- https://github.com/ethereum/staking-deposit-cli/blob/d7b530442d6e0921c9db84b3a1cf6b3ecd6b9d35/staking_deposit/key_handling/keystore.py#L190-L195

Comes out to ~256MB.

So since each thread has its own heap, we can really do quite a bit more decryptions at a time. Ofc taking into account (conservatively) the total amount of free memory available (`os.freeMem()`).

I'd like to see what this looks like and add some benchmarks here so we can keep track of this.

wemeetagain:
> Issues found during testing
> 
>     * incorrect initial log `info: 100% of keystores imported. current=0 total=2000 rate=0.00keys/m`, should show 0 instead of 100%
> 
>     * keystore cache does not seem to work, the file is created after import but on next startup is removed and keys are loaded from scratch

:eyes: will look into that

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Decrypt keystore definitions using a threadpool
+ */
+export async function decryptKeystoreDefinitions(
+  keystoreDefinitions: LocalKeystoreDefinition[],
+  opts: KeystoreDecryptOptions
+): Promise<SignerLocal[]> {
+  if (opts.cacheFilePath) {
+    try {
+      const signers = await loadKeystoreCache(opts.cacheFilePath, keystoreDefinitions);
```
nflaig, line 19: When loading from keystore cache file we are currently not checking if a `.lock` for any of the keystores exists and the cache file itself does not have a lock file.

It is not possible to load the cache file is stored in `dataDir` and if you start a 2nd validator with same `dataDir` you would get `✖ Error: Database is not open` but if a VC has a cache file it will not create lock files which is dangerous.

cc @nazarhussain

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
```diff
@@ -0,0 +1,105 @@
...
+      unlockFilepath(keystorePath);
+    }
+
+    throw new AggregateError(errors);
+  } finally {
+    await pool.terminate();
+  }
+
+  if (opts.cacheFilePath) {
+    await writeKeystoreCache(opts.cacheFilePath, signers, passwords);
```
nflaig, line 99: due to what's mentioned in https://github.com/ChainSafe/lodestar/pull/5357#discussion_r1167822267, signers and passwords are not the same order when loading the keystore cache later

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/worker.ts
```diff
@@ -0,0 +1,26 @@
...
+  const keystore = Keystore.parse(fs.readFileSync(keystorePath, "utf8"));
+
+  // Memory-hogging function
+  const secret = await keystore.decrypt(password);
+  // Transfer the underlying ArrayBuffer back to the main thread: https://threads.js.org/usage-advanced#transferable-objects
+  // This small performance gain may help in cases where this is run for many keystores
+  return Transfer(secret, [secret.buffer]);
+}
+
+expose({decryptKeystoreDefinition} as unknown as DecryptKeystoreWorkerAPI);
```
nflaig, line 26: what does the type cast do here? isn't the exposed function type inferred from type provided to `spawn` function?

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
```diff
@@ -0,0 +1,105 @@
...
+    tasks.push(task);
+    task
+      .then((secretKeyBytes: Uint8Array) => {
+        const signer: SignerLocal = {
+          type: SignerType.Local,
+          secretKey: bls.SecretKey.fromBytes(secretKeyBytes),
+        };
+
+        signers.push(signer);
+        passwords.push(definition.password);
```
nflaig: > keystore cache does not seem to work, the file is created after import but on next startup is removed and keys are loaded from scratch

The issue is that the order of signers is non deterministic due to the fact that we run decryption in parallel and push them to the array asynchronously.

```
Keystore index: 5
Keystore index: 10
Keystore index: 4
Keystore index: 2
Keystore index: 11
Keystore index: 9
Keystore index: 0
Keystore index: 1
Keystore index: 8
Keystore index: 7
Keystore index: 6
Keystore index: 3
```

good news is, we have the index and can use that to ensure the order, updating the code as below fixes the issue

```suggestion
        signers[index] = signer;
        passwords[index] = definition.password;
```

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/types.ts
```diff
@@ -0,0 +1,14 @@
+import {Logger} from "@lodestar/utils";
+import {LocalKeystoreDefinition} from "../interface.js";
+
+export type DecryptKeystoreWorkerAPI = {
+  decryptKeystoreDefinition({keystorePath, password}: LocalKeystoreDefinition, force: boolean): Promise<Uint8Array>;
```
nflaig: why is the `force` param needed? seems to be unused in worker

nflaig:
> incorrect initial log info: 100% of keystores imported. current=0 total=2000 rate=0.00keys/m, should show 0 instead of 100%

this happens due to the fact that `processProgress` is called before `onDecrypt` is called to update the current value

https://github.com/ChainSafe/lodestar/blob/0096e304ac3333a291d28ff2f40b19deb948ff88/packages/cli/src/util/progress.ts#L30

It seems like the problem is just how we calculate the percentage here

https://github.com/ChainSafe/lodestar/blob/42ecf9799951a167d28c816cdb6fc53b40561611/packages/cli/src/util/progress.ts#L38

if current=0 (which is the default) then this will always result in 100 due to 0 being falsy

```node
> 0 && 2000 ? (0 / 2000) * 100 : 100
100
```

cc @nazarhussain

nazarhussain:
> Apr-15 13:48:17.753[]                 info: 100% of keystores imported. current=0 total=2000 rate=0.00keys/m

I believe there was a test for this case, if missed we should add it.

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/types.ts
dapplion: the force param is used to override the LOCK file protection, should be used by the worker

nflaig: We should consider renaming this to something like `ignoreLockfile`. In this particular case, looks like lockfiles are handled outside of the worker

wemeetagain: Its no longer used by the worker, its handled in the main thread. It can be removed. And we can rename the option there.

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
wemeetagain: Can we take this to an issue / separate PR?

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/worker.ts
wemeetagain: The typecast is not great (potentially hiding incompatibilities in the future) but is used to coerce the function after the buffer transfer feature changed the function signature.

wemeetagain:
@nflaig curious to see what your benchmark shows now after the latest commit

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
nflaig: yeah should be separate from this PR, created an issue https://github.com/ChainSafe/lodestar/issues/5374

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/worker.ts
nflaig: Ok was just wondering because I removed the typecast and did not see any issues

nflaig:
> curious to see what your benchmark shows now after the latest commit

It does not look great on my machine (or at least not better than before), it is overloaded which I think just causes bad overall performance (note: I am also running geth and several lodestar instance on that host)

image

This is the result of `calculateThreadpoolConcurrency`:
```
numWorkers: 12
jobConcurrency: 8
```
Those took around 30GB of RAM and put all CPUs under full load (also happened before) as you can see above, probably need to adjust the params somehow to be a bit more conservative.

Just as reference, `htop` from previously tested state, commit  `0096e304ac3333a291d28ff2f40b19deb948ff88`:

image

It does look like memory utilization is better but does not seem to improve performance, I think CPU is just the bottleneck?


This is the log output from latest commit, it looks like decrypt batches take too long now

```
Apr-18 08:22:21.205[]                 info: 0% of keystores imported. current=0 total=2000 rate=0.00keys/m
Apr-18 08:22:31.213[]                 info: 0% of keystores imported. current=0 total=2000 rate=0.00keys/m
Apr-18 08:22:41.215[]                 info: 0% of keystores imported. current=0 total=2000 rate=0.00keys/m
Apr-18 08:22:51.217[]                 info: 0% of keystores imported. current=0 total=2000 rate=0.00keys/m
Apr-18 08:23:01.222[]                 info: 0% of keystores imported. current=0 total=2000 rate=0.00keys/m
Apr-18 08:23:11.222[]                 info: 4% of keystores imported. current=88 total=2000 rate=528.00keys/m
Apr-18 08:23:21.223[]                 info: 4% of keystores imported. current=88 total=2000 rate=0.00keys/m
Apr-18 08:23:31.227[]                 info: 4% of keystores imported. current=88 total=2000 rate=0.00keys/m
Apr-18 08:23:41.229[]                 info: 4% of keystores imported. current=88 total=2000 rate=0.00keys/m
Apr-18 08:23:51.233[]                 info: 4% of keystores imported. current=88 total=2000 rate=0.00keys/m
Apr-18 08:24:01.245[]                 info: 4% of keystores imported. current=88 total=2000 rate=0.00keys/m
Apr-18 08:24:11.245[]                 info: 8% of keystores imported. current=152 total=2000 rate=384.00keys/m
Apr-18 08:24:21.253[]                 info: 7% of keystores imported. current=136 total=2000 rate=-95.92keys/m
Apr-18 08:24:31.265[]                 info: 7% of keystores imported. current=136 total=2000 rate=0.00keys/m
Apr-18 08:24:41.273[]                 info: 7% of keystores imported. current=136 total=2000 rate=0.00keys/m
Apr-18 08:24:51.281[]                 info: 7% of keystores imported. current=136 total=2000 rate=0.00keys/m
Apr-18 08:25:01.289[]                 info: 7% of keystores imported. current=136 total=2000 rate=0.00keys/m
Apr-18 08:25:11.299[]                 info: 12% of keystores imported. current=233 total=2000 rate=581.42keys/m
Apr-18 08:25:21.301[]                 info: 14% of keystores imported. current=288 total=2000 rate=329.93keys/m
Apr-18 08:25:31.309[]                 info: 14% of keystores imported. current=288 total=2000 rate=0.00keys/m
Apr-18 08:25:41.317[]                 info: 14% of keystores imported. current=288 total=2000 rate=0.00keys/m
Apr-18 08:25:51.325[]                 info: 14% of keystores imported. current=288 total=2000 rate=0.00keys/m
Apr-18 08:26:01.333[]                 info: 14% of keystores imported. current=288 total=2000 rate=0.00keys/m
```

Some logs show negative rate for keys/m, likely related to how `onDecrypt` is called, the rate will always be a bit inaccurate.

```
Apr-18 08:24:21.253[]                 info: 7% of keystores imported. current=136 total=2000 rate=-95.92keys/m
```


Problems mentioned earlier are fixed now

- loading from keystore cache ✅
- initial log shows 0% ✅

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/types.ts
```diff
@@ -0,0 +1,14 @@
+import {Logger} from "@lodestar/utils";
+import {LocalKeystoreDefinition} from "../interface.js";
+
+export type DecryptKeystoreWorkerAPI = {
+  decryptKeystoreDefinition({keystorePath, password}: LocalKeystoreDefinition): Promise<Uint8Array>;
+};
+
+export type KeystoreDecryptOptions = {
+  ignoreLockFile?: boolean;
```
nflaig, line 9: haven't tested this but renaming `force` --> `ignoreLockFile` likely breaks current `--force` CLI option

we just destructure `args` here, and top-level flag is not called `ignoreLockFile`

https://github.com/ChainSafe/lodestar/blob/a42fddd356c0511c686622d688a5133943b732a6/packages/cli/src/cmds/validator/signers/index.ts#L98-L99

1. Do we want to keep the name `force` for the CLI flag? will it ever do more than ignore lock files? and if not just rename?
2. Other option to 1., keep top-level name of CLI flag but use more descriptive and intuitive name like `ignoreLockFile`  downstream
3. I feel like we should explicitly assign `args` when calling `decryptKeystoreDefinitions` get type safety on options

wemeetagain:
> It does not look great

Ok will revert the last commit so we just have a job concurrency of 1.

> issues around force vs ignoreLockFile

Lets keep the top-level CLI flag for now (so as to not introduce a breaking change to users) and just use `ignoreLockFile` downstream.

And explicitly assign `args` when calling `decryptKeystoreDefinitions` to get type safety.

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  const signers = new Array(keystoreDefinitions.length) as SignerLocal[];
+  const passwords = new Array(keystoreDefinitions.length) as string[];
+  const tasks: QueuedTask<ModuleThread<DecryptKeystoreWorkerAPI>, Uint8Array>[] = [];
+  const errors: Error[] = [];
+  const pool = Pool(
+    () =>
+      spawn<DecryptKeystoreWorkerAPI>(new Worker("./worker.js"), {
+        // A Lodestar Node may do very expensive task at start blocking the event loop and causing
```
nflaig: This comment seems a bit out of place, should probably be revised. In case of loading the keys there is really nothing else going on as this is a blocking operation

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/index.ts
```diff
@@ -0,0 +1,105 @@
...
+        timeout: 5 * 60 * 1000,
+      }),
+    defaultPoolSize
+  );
+  for (const [index, definition] of keystoreDefinitions.entries()) {
+    try {
+      lockFilepath(definition.keystorePath);
+    } catch (e) {
+      if (opts.ignoreLockFile) {
+        opts.logger.warn("Keystore forcefully loaded even though lockfile could not be acquired", {
```
nflaig: Is this the right warning message? shouldn't it rather be

```suggestion
        opts.logger.warn("Keystore forcefully loaded even though lockfile exists", {
```

Mostly, not sure what "could not be acquired" is trying to say

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/poolSize.ts
```diff
@@ -0,0 +1,16 @@
+let defaultPoolSize: number;
+
+try {
+  if (typeof navigator !== "undefined") {
+    defaultPoolSize = navigator.hardwareConcurrency ?? 4;
+  } else {
+    defaultPoolSize = (await import("node:os")).cpus().length;
```
nflaig, line 8: noting here that we should change this line to use [os.availableParallelism()](https://nodejs.org/api/os.html#osavailableparallelism) once we upgrade to node v20

packages/cli/src/cmds/validator/keymanager/decryptKeystoreDefinitions/worker.ts
```diff
@@ -0,0 +1,26 @@
...
+import {expose} from "@chainsafe/threads/worker";
+import {Transfer, TransferDescriptor} from "@chainsafe/threads";
+import {Keystore} from "@chainsafe/bls-keystore";
+import {LocalKeystoreDefinition} from "../interface.js";
+import {DecryptKeystoreWorkerAPI} from "./types.js";
+
+/**
+ * Decrypt a single keystore definition, returning the secret key as a Uint8Array
+ *
+ * NOTE: This is memory-intensive process, since decrypting the keystore involves running a key derivation function (either pbkdf2 or scrypt)
```
nflaig: this note is correct but based on the performance tests it looks like CPU is rather the bottleneck that prevents us from increasing the job concurrency to further speed up the key import

With job concurrency of 1

image

packages/cli/test/utils/keymanagerTestRunners.ts
```diff
@@ -78,9 +79,10 @@ export function getKeymanagerTestRunner({args: {spawnCli}, afterEachCallbacks, d
 export async function expectKeys(keymanagerClient: Api, expectedPubkeys: string[], message: string): Promise<void> {
   const keys = await keymanagerClient.listKeys();
   ApiError.assert(keys);
-  expectDeepEquals(
-    keys.response.data,
-    expectedPubkeys.map((pubkey) => ({validatingPubkey: pubkey, derivationPath: "", readonly: false})),
-    message
-  );
+
+  // the order of keys isn't always deterministic so we can't use deep equal
```
nflaig: I would assume this change can be reverted since the keystore cache fix made the order of keys deterministic

packages/cli/test/unit/validator/decryptKeystoreDefinitions.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    try {
+      await decryptKeystoreDefinitions(definitions, {logger: console});
+      expect.fail("Second decrypt should fail due to failure to get lockfile");
+    } catch (e) {
+      expect((e as Error).message.startsWith("EEXIST: file already exists"), "Wrong error is thrown").to.be.true;
+    }
+  });
+
+  it("decrypt keystores if lockfiles already exist if force=true", async () => {
```
nflaig:
```suggestion
  it("decrypt keystores if lockfiles already exist if ignoreLockFile=true", async () => {
```

nflaig (commented):
lgtm, just few final remarks after more in-depth review. Everything works as expected now and the performance boost is pretty amazing, on average it is now **5x** faster and in combination with the keystore cache I think the UX for big operators is pretty good now

nflaig (approved):

nflaig:
@wemeetagain I noticed that we have a similar logic in the [keymanager API](https://github.com/ChainSafe/lodestar/blob/34c2a2af773bb14bb6e8de6ec32bc553062466e1/packages/cli/src/cmds/validator/keymanager/impl.ts#L111). Is supporting thread pool decryption something we should consider there as well or do we expect that this API will generally be used to import less overall keys?

https://github.com/ChainSafe/lodestar/blob/34c2a2af773bb14bb6e8de6ec32bc553062466e1/packages/cli/src/cmds/validator/keymanager/impl.ts#L146

wemeetagain:
Probably wouldn't hurt to support decryption in parallel here. :shrug:

dapplion:
> @wemeetagain I noticed that we have a similar logic in the [keymanager API](https://github.com/ChainSafe/lodestar/blob/34c2a2af773bb14bb6e8de6ec32bc553062466e1/packages/cli/src/cmds/validator/keymanager/impl.ts#L111). Is supporting thread pool decryption something we should consider there as well or do we expect that this API will generally be used to import less overall keys?
> 
> https://github.com/ChainSafe/lodestar/blob/34c2a2af773bb14bb6e8de6ec32bc553062466e1/packages/cli/src/cmds/validator/keymanager/impl.ts#L146

Yes definitely, that should be done in a thread and limit with a queue to prevent killing the vc process

wemeetagain:

Issue #2438, opened 2021-04-27
`--private` mode CLI option to not send impl details


3xtr4t3rr3str14l:
**Is your feature request related to a problem? Please describe.**

I noticed Lighthouse has a `--private` option in their CLI that prevents client info from being sent over the identity protocol.  Maybe if we do this too it would be nice to have.

**Describe alternatives you've considered**

**Additional context**

dapplion:
What's the usecase for that flag, when you don't want gossip, for local only testnets?

3xtr4t3rr3str14l:
> What's the usecase for that flag, when you don't want gossip, for local only testnets?

I think it's for not sending agent info across the network (client type, version, etc)

3xtr4t3rr3str14l:
But yeah not sure of a specific use case other than just someone not wanting to have that info sent out to minimize potentially identifying info.

philknows:
Latest discussion points from @wemeetagain :

Use case would be similar to setting a blank graffiti, ie if you're paranoid about getting identified, attacked w lodestar-specific attacks.
1. attacker learns about lodestar DoS
2. attacker scans thru the network, connecting and running identify on all
3. attacker builds list of lodestar nodes
4. attacker executes attack on all lodestar nodes
Issue is a bit misworded since the user agent is part of the identify protocol, not part of gossipsub
Imo still useful but low priority

PR #5361, opened 2023-04-13
Add User-Agent to builder requests

### LINKED ISSUE #5360: Send node impl/version as user agent on builder requests
Feature request from metachris#2573 of flashbots

> please start sending lodestar name and version as user agent for requests to the builder network

**Rationale**

Allow relays to support individual emergencies if need be, such that not all validators are caught up in blanked filters

dapplion:
**Motivation**

- see https://github.com/ChainSafe/lodestar/issues/5360

**Description**

- Set on all requests by default
- Set to current agent version by default
- Allow to customize and disable via visible flag `--builder.userAgent`

```
builder
      --builder                        Enable builder interface
                                                      [boolean] [default: false]
      --builder.urls                   Urls hosting the builder API      [array]
      --builder.timeout                Timeout in milliseconds for builder API H
                                       TTP client                       [number]
      --builder.faultInspectionWindow  Window to inspect missed slots for enabli
                                       ng/disabling builder circuit breaker
                                                                        [number]
      --builder.allowedFaults          Number of missed slots allowed in the fau
                                       ltInspectionWindow for builder circuit
                                                                        [number]
      --builder.userAgent              User-Agent header attached to all builder
                                        requests
        [string] [default: "Lodestar/v1.6.0/dapplion/builder-useragent/8846e16"]
```

Closes https://github.com/ChainSafe/lodestar/issues/5360

### DIFF
--- a/packages/api/src/utils/client/httpClient.ts
+++ b/packages/api/src/utils/client/httpClient.ts
@@ -54,6 +54,7 @@ export interface URLOpts {
   baseUrl: string;
   timeoutMs?: number;
   bearerToken?: string;
+  extraHeaders?: Record<string, string>;
 }
 
 export type FetchOpts = {
@@ -77,6 +78,7 @@ export interface IHttpClient {
 export type HttpClientOptions = ({baseUrl: string} | {urls: (string | URLOpts)[]}) & {
   timeoutMs?: number;
   bearerToken?: string;
+  extraHeaders?: Record<string, string>;
   /** Return an AbortSignal to be attached to all requests */
   getAbortSignal?: () => AbortSignal | undefined;
   /** Override fetch function */
@@ -93,6 +95,7 @@ export {Metrics};
 export class HttpClient implements IHttpClient {
   private readonly globalTimeoutMs: number;
   private readonly globalBearerToken: string | null;
+  private readonly globalExtraHeaders: Record<string, string> | null;
   private readonly getAbortSignal?: () => AbortSignal | undefined;
   private readonly fetch: typeof fetch;
   private readonly metrics: null | Metrics;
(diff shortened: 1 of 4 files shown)


packages/api/src/utils/client/httpClient.ts
```diff
@@ -265,7 +271,8 @@ export class HttpClient implements IHttpClient {
     try {
       const url = urlJoin(baseUrl, opts.url) + (opts.query ? "?" + stringifyQuery(opts.query) : "");
 
-      const headers = opts.headers || {};
+      const headers =
+        extraHeaders && opts.headers ? {...extraHeaders, ...opts.headers} : opts.headers || extraHeaders || {};
```
nflaig, line 275: doesn't the following achieve the same?

```suggestion
      const headers = {...extraHeaders, ...opts.headers};
```

as destructuring `undefined` and `null` will just be omitted

```node
> {...null}
{}
> {...undefined}
{}
```

apparently spread syntax on `null` is supposed to throw an error but can't reproduce that on node 18, but just in case could do

```suggestion
      const headers = {...(extraHeaders ?? {}), ...(opts.headers ?? {})};
```

packages/cli/src/options/beaconNodeOptions/builder.ts
```diff
@@ -56,4 +59,12 @@ export const options: CliCommandOptions<ExecutionBuilderArgs> = {
     description: "Number of missed slots allowed in the faultInspectionWindow for builder circuit",
     group: "builder",
   },
+
+  "builder.userAgent": {
```
nflaig, line 63: is this a value we want operators to mess with? defeats the whole purpose of what this PR is trying to achieve. Maybe hiding it would be better

packages/api/src/utils/client/httpClient.ts
dapplion: Functionally equivalent but your approach requires spreading all object's keys if only one headers option is specified

packages/cli/src/options/beaconNodeOptions/builder.ts
dapplion: Given that this leaks information I prefer to expose the option to mess with it. Either:
- A flag to prevent sending this information
- A flag to control what is sent on the user agent

We should do something similar for the p2p agent

nflaig (approved):

wemeetagain (approved):

packages/cli/src/options/beaconNodeOptions/builder.ts
nflaig: I like `--private` flag (or similar) suggested in https://github.com/ChainSafe/lodestar/issues/2438, sounds much more elegant option for users with privacy/security concerns, can of course still keep flags to individually override each of those values

dapplion: Feel free to PR on top of unstable

wemeetagain:

PR #5385, opened 2023-04-19
Fix the formatting of blockValue in the validator proposer logs

g11tech:
Since the blockvalue is most < 1 eth, current display of blockvalue in eth logs as zero because of bigint arithmetic

image

This PR fixes the formatting so that the above would now be displayed as `0.10379` upto 5 decimal places

### DIFF
--- a/packages/validator/src/services/block.ts
+++ b/packages/validator/src/services/block.ts
@@ -247,7 +247,7 @@ export class BlockProposingService {
     const debugLogCtx = {
       source: source,
       // winston logger doesn't like bigint
-      "blockValue(approx ETH)": formatBigDecimal(fullOrBlindedBlock.blockValue, ETH_TO_WEI, MAX_DECIMAL_FACTOR),
+      blockValue: `${formatBigDecimal(fullOrBlindedBlock.blockValue, ETH_TO_WEI, MAX_DECIMAL_FACTOR)} ETH`,
     };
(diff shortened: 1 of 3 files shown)


packages/validator/src/services/block.ts
```diff
@@ -243,7 +247,7 @@ export class BlockProposingService {
     const debugLogCtx = {
       source: source,
       // winston logger doesn't like bigint
-      "blockValue(eth)": `${fullOrBlindedBlock.blockValue / ETH_TO_WEI}`,
+      "blockValue(approx ETH)": formatBigDecimal(fullOrBlindedBlock.blockValue, ETH_TO_WEI, MAX_DECIMAL_FACTOR),
```
nflaig: I think `approx` can be omitted, it should be obvious due to the fact that only 5 decimals are shown, this is also common practice on other websites like etherscan or beaconcha.in

would either do

```suggestion
      "blockValue(ETH)": formatBigDecimal(fullOrBlindedBlock.blockValue, ETH_TO_WEI, MAX_DECIMAL_FACTOR),
```

or  what would feel more intuitive to me

```suggestion
      "blockValue": `${formatBigDecimal(fullOrBlindedBlock.blockValue, ETH_TO_WEI, MAX_DECIMAL_FACTOR)} ETH`,
```

g11tech: thanks updated!

nflaig (approved):
lgtm

wemeetagain:

PR #5388, opened 2023-04-19
feat: prune invalid SSZ objects

### LINKED ISSUE #4420: Prune old invalidSszObjects every some interval
**Describe the bug**

Persisting invalidSszObjects is very useful but not if persisted for months. Add flag to specific a retention window, after that clean.

Otherwise it risks filling up the host's disk with old junk

acuarica:
**Motivation**

This PR closes #4420. That is, it allows the user to prune old `invalidSszObjects` after a specified number of days.

**Description**

This PR adds a new CLI hidden option `--persistInvalidSszObjectsRetention` only applicable to `beacon` command. It works in conjunction with `--chain.persistInvalidSszObjects` and `--persistInvalidSszObjectsDir`.

When `--chain.persistInvalidSszObjects` is enabled, pruning is activated by default. However, you can disable it with `--persistInvalidSszObjectsRetention 0`

If `--persistInvalidSszObjectsRetention` is not specified, the retention period for invalid SSZ objects defaults to 7 days.

The application prunes invalid SSZ objects during start up and every hour.

```sh
git checkout acuarica/prune-invalid-ssz-objects
lodestar beacon --persistInvalidSszObjectsRetention 14
```

### DIFF
--- a/packages/cli/src/cmds/beacon/handler.ts
+++ b/packages/cli/src/cmds/beacon/handler.ts
@@ -105,7 +105,7 @@ export async function beaconHandler(args: BeaconArgs & GlobalArgs): Promise<void
               (args.persistInvalidSszObjectsRetentionHours ?? DEFAULT_RETENTION_SSZ_OBJECTS_HOURS) * HOURS_TO_MS
             );
           } catch (e) {
-            logger.warn("error pruning invalid SSZ objects", {persistInvalidSszObjectsDir}, e as Error);
+            logger.warn("Error pruning invalid SSZ objects", {persistInvalidSszObjectsDir}, e as Error);
           }
           // Run every ~1 hour
         }, HOURS_TO_MS)
--- a/packages/cli/src/cmds/beacon/handler.ts
+++ b/packages/cli/src/cmds/beacon/handler.ts
@@ -11,12 +11,22 @@ import {LoggerNode, getNodeLogger} from "@lodestar/logger/node";
 import {GlobalArgs, parseBeaconNodeArgs} from "../../options/index.js";
 import {BeaconNodeOptions, getBeaconConfigFromArgs} from "../../config/index.js";
 import {getNetworkBootnodes, getNetworkData, isKnownNetworkName, readBootnodes} from "../../networks/index.js";
-import {onGracefulShutdown, mkdir, writeFile600Perm, cleanOldLogFiles, parseLoggerArgs} from "../../util/index.js";
+import {
+  onGracefulShutdown,
+  mkdir,
+  writeFile600Perm,
+  cleanOldLogFiles,
+  parseLoggerArgs,
+  pruneOldFilesInDir,
+} from "../../util/index.js";
 import {getVersionData} from "../../util/version.js";
 import {BeaconArgs} from "./options.js";
 import {getBeaconPaths} from "./paths.js";
 import {initBeaconState} from "./initBeaconState.js";
 import {initPeerIdAndEnr} from "./initPeerIdAndEnr.js";
+
+const DEFAULT_RETENTION_SSZ_OBJECTS_HOURS = 15 * 24;
+const HOURS_TO_MS = 3600 * 1000;
 
 /**
  * Runs a beacon node.
@@ -80,8 +90,28 @@ export async function beaconHandler(args: BeaconArgs & GlobalArgs): Promise<void
       metricsRegistries,
     });
 
-    if (args.attachToGlobalThis) (globalThis as unknown as {bn: BeaconNode}).bn = node;
+    // dev debug option to have access to the BN instance
+    if (args.attachToGlobalThis) {
+      (globalThis as unknown as {bn: BeaconNode}).bn = node;
+    }
 
+    // Prune invalid SSZ objects every interval
+    const {persistInvalidSszObjectsDir} = args;
+    const pruneInvalidSSZObjectsInterval = persistInvalidSszObjectsDir
+      ? setInterval(() => {
+          try {
+            pruneOldFilesInDir(
+              persistInvalidSszObjectsDir,
+              (args.persistInvalidSszObjectsRetentionHours ?? DEFAULT_RETENTION_SSZ_OBJECTS_HOURS) * HOURS_TO_MS
+            );
+          } catch (e) {
+            logger.warn("error pruning invalid SSZ objects", {persistInvalidSszObjectsDir}, e as Error);
+          }
+          // Run every ~1 hour
+        }, HOURS_TO_MS)
+      : null;
+
+    // Intercept SIGINT signal, to perform final ops before exiting
     onGracefulShutdown(async () => {
       if (args.persistNetworkIdentity) {
         try {
@@ -93,6 +123,10 @@ export async function beaconHandler(args: BeaconArgs & GlobalArgs): Promise<void
         }
       }
       abortController.abort();
+
+      if (pruneInvalidSSZObjectsInterval !== null) {
+        clearInterval(pruneInvalidSSZObjectsInterval);
+      }
     }, logger.info.bind(logger));
 
     abortController.signal.addEventListener(
--- a/packages/cli/src/util/pruneOldFilesInDir.ts
+++ b/packages/cli/src/util/pruneOldFilesInDir.ts
@@ -1,0 +1,17 @@
+import fs from "node:fs";
+import path from "node:path";
+
+export function pruneOldFilesInDir(dirpath: string, maxAgeMs: number): void {
+  for (const entryName in fs.readdirSync(dirpath)) {
+    const entryPath = path.join(dirpath, entryName);
+
+    const stat = fs.statSync(entryPath);
+    if (stat.isDirectory()) {
+      pruneOldFilesInDir(entryPath, maxAgeMs);
+    } else if (stat.isFile()) {
+      if (Date.now() - stat.ctimeMs > maxAgeMs) {
+        fs.unlinkSync(entryPath);
+      }
+    }
+  }
+}
--- a/packages/cli/src/util/pruneOldFilesInDir.ts
+++ b/packages/cli/src/util/pruneOldFilesInDir.ts
@@ -9,7 +9,7 @@ export function pruneOldFilesInDir(dirpath: string, maxAgeMs: number): void {
     if (stat.isDirectory()) {
       pruneOldFilesInDir(entryPath, maxAgeMs);
     } else if (stat.isFile()) {
-      if (Date.now() - stat.ctimeMs > maxAgeMs) {
+      if (Date.now() - stat.mtimeMs > maxAgeMs) {
         fs.unlinkSync(entryPath);
       }
     }
--- a/packages/cli/src/util/pruneOldFilesInDir.ts
+++ b/packages/cli/src/util/pruneOldFilesInDir.ts
@@ -2,7 +2,7 @@ import fs from "node:fs";
 import path from "node:path";
 
 export function pruneOldFilesInDir(dirpath: string, maxAgeMs: number): void {
-  for (const entryName in fs.readdirSync(dirpath)) {
+  for (const entryName of fs.readdirSync(dirpath)) {
     const entryPath = path.join(dirpath, entryName);
 
     const stat = fs.statSync(entryPath);
--- a/packages/cli/src/cmds/beacon/options.ts
+++ b/packages/cli/src/cmds/beacon/options.ts
@@ -15,6 +15,7 @@ type BeaconExtraArgs = {
   beaconDir?: string;
   dbDir?: string;
   persistInvalidSszObjectsDir?: string;
+  persistInvalidSszObjectsRetentionHours?: number;
   peerStoreDir?: string;
   persistNetworkIdentity?: boolean;
 };
@@ -83,6 +84,12 @@ export const beaconExtraOptions: CliCommandOptions<BeaconExtraArgs> = {
     defaultDescription: defaultBeaconPaths.persistInvalidSszObjectsDir,
     hidden: true,
     type: "string",
+  },
+
+  persistInvalidSszObjectsRetentionHours: {
+    description: "Number of hours to keep invalid SSZ objects in local disk",
+    hidden: true,
+    type: "number",
   },
 
   peerStoreDir: {
--- a/packages/cli/src/cmds/beacon/options.ts
+++ b/packages/cli/src/cmds/beacon/options.ts
@@ -87,7 +87,7 @@ export const beaconExtraOptions: CliCommandOptions<BeaconExtraArgs> = {
   },
 
   persistInvalidSszObjectsRetentionHours: {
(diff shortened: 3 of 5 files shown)


packages/beacon-node/src/chain/chain.ts
```diff
@@ -505,6 +506,36 @@ export class BeaconChain implements IBeaconChain {
     }
   }
 
+  prunePastInvalidSszObjects(): void {
+    const invalidObjectsDir = this.opts.persistInvalidSszObjectsDir ?? "invalid_ssz_objects";
```
wemeetagain: Can you set the fallback at a higher level so it doesn't have to be repeated here and in `persistInvalidSszObject`?

packages/beacon-node/src/chain/chain.ts
```diff
@@ -505,6 +506,36 @@ export class BeaconChain implements IBeaconChain {
...
+    const invalidObjectsRetentionDays = this.opts.persistInvalidSszObjectsRetention ?? 7;
+    this.logger.debug("Prunning invalid ssz objects", {invalidObjectsDir, invalidObjectsRetentionDays});
```
wemeetagain: I think pruning has only one 'n' in it.

packages/beacon-node/src/chain/chain.ts
```diff
@@ -505,6 +506,36 @@ export class BeaconChain implements IBeaconChain {
...
+
+    const datesPastRetention = (date: string): boolean =>
+      new Date(date).getTime() < today.getTime() - retentionDays * DAYS_TO_MS;
+    fs.readdir(basePath, (err, list) => {
+      if (!err) {
+        for (const date of list.filter(datesPastRetention)) {
+          const dirpath = path.join(basePath, date);
+          fs.rm(dirpath, {recursive: true}, (err) => {
+            if (err) {
+              this.logger.error("Failed to remove invalid ssz object", {dir: dirpath, err: err.message});
```
wemeetagain: According to our recently added logging policy (https://github.com/ChainSafe/lodestar/pull/5299), these error logs should be debug logs.

packages/beacon-node/src/chain/chain.ts
```diff
@@ -505,6 +506,36 @@ export class BeaconChain implements IBeaconChain {
...
+
+    const retentionDays = invalidObjectsRetentionDays + 1;
+    const DAYS_TO_MS = 24 * 60 * 60 * 1000;
+    const today = new Date();
+    const basePath = invalidObjectsDir;
+
+    const datesPastRetention = (date: string): boolean =>
+      new Date(date).getTime() < today.getTime() - retentionDays * DAYS_TO_MS;
+    fs.readdir(basePath, (err, list) => {
+      if (!err) {
```
wemeetagain: Might be good for readability to align nested error, success cases consistently here (eg: error first, then success or vice versa) or just use promise-based API.

packages/beacon-node/src/chain/chain.ts
```diff
@@ -505,6 +506,36 @@ export class BeaconChain implements IBeaconChain {
...
+    const invalidObjectsRetentionDays = this.opts.persistInvalidSszObjectsRetention ?? 7;
+    this.logger.debug("Prunning invalid ssz objects", {invalidObjectsDir, invalidObjectsRetentionDays});
+
+    const retentionDays = invalidObjectsRetentionDays + 1;
+    const DAYS_TO_MS = 24 * 60 * 60 * 1000;
+    const today = new Date();
+    const basePath = invalidObjectsDir;
+
+    const datesPastRetention = (date: string): boolean =>
+      new Date(date).getTime() < today.getTime() - retentionDays * DAYS_TO_MS;
```
wemeetagain: small thing, but `today.getTime() - retentionDays * DAYS_TO_MS` can be precomputed outside the loop.
Figured if you're already doing so with `DAYS_TO_MS`, you might as well go all the way.

wemeetagain (dismissed):

packages/beacon-node/src/chain/chain.ts
acuarica: I have used a `private` property to avoid duplication 751969c03a. Does this solve the issue?

acuarica:
Hey @wemeetagain, thanks for your comments. I addressed all of them and left a question.

packages/beacon-node/src/chain/chain.ts
```diff
@@ -505,6 +506,36 @@ export class BeaconChain implements IBeaconChain {
     }
   }
 
+  prunePastInvalidSszObjects(): void {
```
dapplion: This code should be moved to the cli package. Unless extremely necessary filesystem interactions should live there

packages/beacon-node/src/chain/chain.ts
```diff
@@ -505,6 +506,36 @@ export class BeaconChain implements IBeaconChain {
...
+          fs.rm(dirpath, {recursive: true}, (err) => {
+            if (err) {
+              this.logger.debug("Failed to remove invalid ssz object", {dir: dirpath, err: err.message});
+            } else {
+              this.logger.debug("Removed invalid ssz object", {path: dirpath});
+            }
+          });
+        }
+      }
+    });
```
dapplion: We don't use the fs callback based API in Lodestar, just use the promise based and put a try catch at the top

packages/beacon-node/src/node/nodejs.ts
```diff
@@ -213,6 +213,11 @@ export class BeaconNode {
     // Load persisted data from disk to in-memory caches
     await chain.loadFromDisk();
 
+    if (opts.chain.persistInvalidSszObjects && opts.chain.persistInvalidSszObjectsRetention !== 0) {
+      chain.prunePastInvalidSszObjects();
+      setInterval(() => chain.prunePastInvalidSszObjects(), 1000 * 60 * 60);
```
dapplion: This interval must be handled and clean gracefully on shutdown

packages/beacon-node/src/chain/chain.ts
```diff
@@ -505,6 +506,36 @@ export class BeaconChain implements IBeaconChain {
...
+    const invalidObjectsRetentionDays = this.opts.persistInvalidSszObjectsRetention ?? 7;
+    this.logger.debug("Pruning invalid ssz objects", {invalidObjectsDir, invalidObjectsRetentionDays});
+
+    const retentionDays = invalidObjectsRetentionDays + 1;
+    const DAYS_TO_MS = 24 * 60 * 60 * 1000;
+    const today = new Date();
+    const retentionDateInMS = today.getTime() - retentionDays * DAYS_TO_MS;
+    const basePath = invalidObjectsDir;
+
+    const datesPastRetention = (date: string): boolean => new Date(date).getTime() < retentionDateInMS;
```
dapplion: Why define the function at the top instead of in-lining?

packages/beacon-node/src/chain/chain.ts
```diff
@@ -505,6 +506,36 @@ export class BeaconChain implements IBeaconChain {
...
+      if (err) {
+        this.logger.debug("Unable to delete invalid ssz objects", {err: err.message});
+      } else {
+        for (const date of list.filter(datesPastRetention)) {
+          const dirpath = path.join(basePath, date);
+          fs.rm(dirpath, {recursive: true}, (err) => {
+            if (err) {
+              this.logger.debug("Failed to remove invalid ssz object", {dir: dirpath, err: err.message});
+            } else {
+              this.logger.debug("Removed invalid ssz object", {path: dirpath});
```
dapplion: I'm not sure we want to know about every single file deleted. Probably just logging that N files where deleted up to D day is sufficient

dapplion (changes requested):
Good start! Main comment about moving code to CLI package

packages/beacon-node/src/chain/chain.ts
```diff
@@ -505,6 +506,36 @@ export class BeaconChain implements IBeaconChain {
     }
   }
 
+  prunePastInvalidSszObjects(): void {
+    const invalidObjectsDir = this.persistInvalidSszObjectsDir;
+    const invalidObjectsRetentionDays = this.opts.persistInvalidSszObjectsRetention ?? 7;
```
nflaig: I feel like specifying the default value here is not ideal, might be better to define default options object which can also be referenced in the `defaultDescription` in cmd options. That way it is ensured that those are always in sync

https://github.com/ChainSafe/lodestar/blob/751969c03ae443032e9e4d062d1e15ff89a2d00a/packages/cli/src/cmds/beacon/options.ts#L91

packages/beacon-node/src/chain/chain.ts
acuarica: I see. I put `prunePastInvalidSszObjects` here because there is the method `persistInvalidSszObject`, which writes to disk. The rationale being both write and delete being in the same module.

packages/beacon-node/src/node/nodejs.ts
acuarica: I used `signal.addEventListener("abort", [...])`, is that ok?

packages/beacon-node/src/chain/chain.ts
```diff
@@ -1,4 +1,6 @@
 import path from "node:path";
+import fs from "node:fs";
+import {promisify} from "node:util";
```
nflaig: it is better to just import the promise-based functions from `fs/promises`

```suggestion
import fs from "node:fs/promises";
```

acuarica: Thanks for the heads up. Just FYI I used `promisify` because it was used when writing invalid SSZ objects

https://github.com/ChainSafe/lodestar/blob/751969c03ae443032e9e4d062d1e15ff89a2d00a/packages/beacon-node/src/util/file.ts#L17

nflaig: yeah, not a big deal either way, but should avoid using `promisify` if native promise-based APIs exist

packages/beacon-node/src/chain/chain.ts
```diff
@@ -505,6 +506,30 @@ export class BeaconChain implements IBeaconChain {
...
+    const today = new Date();
+    const retentionDateInMS = today.getTime() - retentionDays * DAYS_TO_MS;
+    const basePath = invalidObjectsDir;
+
+    try {
+      const list = await fs.readdir(basePath);
+      const filteredList = list.filter((date) => new Date(date).getTime() < retentionDateInMS);
+      for (const date of filteredList) {
+        const dirpath = path.join(basePath, date);
+        void fs.rm(dirpath, {recursive: true});
```
nflaig: any reason why promises not handled here? if `fs.rm` runs into an error this will result in an unhandled promise rejection and not be caught by `catch`, the log ""Removed invalid ssz objects" will also be executed before the files are actually deleted

I would assume it is fine to delete the files in parallel, meaning can just do `.map` and resolve the promises with `await Promise.all`

packages/beacon-node/src/node/nodejs.ts
```diff
@@ -213,6 +213,14 @@ export class BeaconNode {
     // Load persisted data from disk to in-memory caches
     await chain.loadFromDisk();
 
+    if (opts.chain.persistInvalidSszObjects && opts.chain.persistInvalidSszObjectsRetention !== 0) {
+      void chain.prunePastInvalidSszObjects();
+      const pruneIntervalId = setInterval(() => chain.prunePastInvalidSszObjects(), 1000 * 60 * 60);
```
nflaig: highly opinionated but I can't intuitively read this, changing the order of units would be nice

```suggestion
      const pruneIntervalId = setInterval(() => chain.prunePastInvalidSszObjects(),  60 * 60 * 1000);
```

nflaig (commented):
Some basic unit tests for this would be good to make sure files are deleted correctly and errors are handled

packages/beacon-node/src/chain/chain.ts
acuarica: Good catch. Not sure about deleting in parallel though, so I just `await` for `fs.rm`.

nflaig: yeah you are right, this task runs fully async so there is no rush, better to run in sequence to keep file system load as low as possible

dapplion:
Redone, now with code in beacon handler only

packages/cli/src/cmds/beacon/handler.ts
```diff
@@ -80,8 +90,27 @@ export async function beaconHandler(args: BeaconArgs & GlobalArgs): Promise<void
       metricsRegistries,
     });
 
-    if (args.attachToGlobalThis) (globalThis as unknown as {bn: BeaconNode}).bn = node;
+    // dev debug option to have access to the BN instance
+    if (args.attachToGlobalThis) {
+      (globalThis as unknown as {bn: BeaconNode}).bn = node;
+    }
```
dapplion, line 96: Added comment and this section of code is getting crowded

dapplion (dismissed):

packages/cli/src/cmds/beacon/handler.ts
```diff
@@ -80,8 +90,28 @@ export async function beaconHandler(args: BeaconArgs & GlobalArgs): Promise<void
...
 
+    // Prune invalid SSZ objects every interval
+    const {persistInvalidSszObjectsDir} = args;
+    const pruneInvalidSSZObjectsInterval = persistInvalidSszObjectsDir
+      ? setInterval(() => {
```
nflaig, line 101: I feel like logic related to pruning ssz objects + constants should not be defined here, extracting to util might be better, similar to `cleanOldLogFiles`.

packages/cli/src/util/pruneOldFilesInDir.ts
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  for (const entryName of fs.readdirSync(dirpath)) {
+    const entryPath = path.join(dirpath, entryName);
+
+    const stat = fs.statSync(entryPath);
+    if (stat.isDirectory()) {
+      pruneOldFilesInDir(entryPath, maxAgeMs);
+    } else if (stat.isFile()) {
+      if (Date.now() - stat.mtimeMs > maxAgeMs) {
```
nflaig, line 12: Changed from `cTimeMs` to `mtimeMs` as [can't manipulate creation time](https://stackoverflow.com/a/44240159/10577550) from within Nodejs which makes it hard to test.

I am not sure what is technically more correct. Currently, it should not matter for ssz objects because I would assume those are just written once and not modified.

For more general purpose, I would not consider a file which was just recently modified "old" so might be anyways better to use `mTimeMs`

packages/cli/src/cmds/beacon/handler.ts
nflaig: Refactoring is not worth it right now as it would just make implementation less readable, might consider refactoring later if beacon handler growths further

nflaig (dismissed):

nflaig (approved):

dapplion (approved):

wemeetagain:

PR #5387, opened 2023-04-19
Add option to override preset with custom values from file

nflaig:
**Motivation**

Follow up to https://github.com/ChainSafe/lodestar/pull/4601 to expose `setActivePreset` through the CLI and not just at the library level.

Presets are more extensive than runtime configurations, and generally only applicable during compile-time. However, to support various private networks, simulation networks  or CI/CD integration tests we require more flexibility.

The main motivation for this PR is to support running Lodestar VC in Obol Network CI/CD integration tests where they use a custom configuration that requires to override preset values such as `SLOTS_PER_EPOCH` and `EPOCHS_PER_SYNC_COMMITTEE_PERIOD`.

**Description**

Adds CLI flag `--presetFile` which invokes `setActivePreset` with custom values from file to override values of the active preset.

**Usage**

> preset.yaml

```yml
SLOTS_PER_EPOCH: 16
EPOCHS_PER_SYNC_COMMITTEE_PERIOD: 2
```

```
lodestar beacon --presetFile "./preset.yaml"
```

PR #4601, opened 2022-09-27
Allow preset overrides

daniildxb:
**Motivation**

This PR allows users to override preset values (i.e. for use in ephemerals)
**Description**

Adds `setCustomPreset(presetName: PresetName, overrides: Partial<BeaconPreset>)` method that allows users to pass preset base and override object

Closes #4598 

**Steps to test or reproduce**
Added e2e test to params package to verify it works

PR #5389, opened 2023-04-19
Correctly override active preset

nflaig:
**Motivation**

Fix for https://github.com/ChainSafe/lodestar/pull/4601 and  https://github.com/ChainSafe/lodestar/pull/5387, `setActivePreset` now correctly overrides not only the variables that are exported individually/explicitly but also the exported `activePreset` object.

**Description**

- Correctly override the active preset
- Add beacon preset types to filter out extraneous keys
- Improve consistency with [ChainConfig](https://github.com/ChainSafe/lodestar/blob/unstable/packages/config/src/chainConfig/types.ts) by converting `BeaconPreset` to type and renaming the file to `types.ts`
- Add jsdoc to `BeaconPreset`
- Verify that preset values in `activePreset` are overridden

**Note:** Initial implementation in https://github.com/ChainSafe/lodestar/pull/5387 did not filter out extraneous keys due to the fact that the destructuring did that already. However, since we override the `activePreset` object now I think it is better to apply filtering to not pollute the object as it is done in [chainConfigToJson](https://github.com/ChainSafe/lodestar/blob/35fe214ef2c26bcb1cc53181142e376e03a5500a/packages/config/src/chainConfig/json.ts#L9).

### DIFF
--- a/packages/params/src/types.ts
+++ b/packages/params/src/types.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/naming-convention */
-export interface BeaconPreset {
+export type BeaconPreset = {
   // Misc
   MAX_COMMITTEES_PER_SLOT: number;
   TARGET_COMMITTEE_SIZE: number;
@@ -76,7 +76,7 @@ export interface BeaconPreset {
   ///////////
   FIELD_ELEMENTS_PER_BLOB: number;
   MAX_BLOBS_PER_BLOCK: number;
-}
+};
 
 /**
  * Presets only contain numbers, this is just used to filter out extraneous keys
--- a/packages/params/src/types.ts
+++ b/packages/params/src/types.ts
@@ -1,4 +1,8 @@
 /* eslint-disable @typescript-eslint/naming-convention */
+
+/**
+ * Compile-time chain configuration
+ */
 export type BeaconPreset = {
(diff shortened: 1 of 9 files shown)


packages/params/src/types.ts
```diff
@@ -0,0 +1,170 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+
+/**
+ * Compile-time chain configuration
```
nflaig, line 4: @wemeetagain not sure if "compile-time" is the right wording here but want to add some context how this differs from `ChainConfig`

https://github.com/ChainSafe/lodestar/blob/35fe214ef2c26bcb1cc53181142e376e03a5500a/packages/config/src/chainConfig/types.ts#L6

wemeetagain: load-time? import-time?

compile-time also seems fine even tho its not exactly right

wemeetagain (commented):
lgtm

packages/params/src/types.ts
nflaig: lets keep compile-time then, I think it conveys the difference between config and preset the best, even though it is technically not accurate in the context of JS

wemeetagain (approved):

wemeetagain:

Issue #5384, opened 2023-04-19
High regen queue job wait time


twoeths:
**Describe the bug**

With the new gossip flow, the Job Wait Time of regen queue is almost 1s, this is likely one of the reason of I/O lag issue

<img width="1306" alt="Screenshot 2023-04-19 at 13 32 46" src="https://user-images.githubusercontent.com/10568965/232986401-ab0fdc99-d051-4e5e-ad0c-4260075598d4.png">

right now the regen queue length is limited to 16 to accept new gossip work

<img width="1287" alt="Screenshot 2023-04-19 at 13 34 43" src="https://user-images.githubusercontent.com/10568965/232986769-d808bc8c-a74a-4b70-89e1-4f528162c339.png">

**Expected behavior**

- This is from stable (v1.7): regen queue is only 1 or 0 and job wait time is way lower
<img width="1291" alt="Screenshot 2023-04-19 at 13 40 31" src="https://user-images.githubusercontent.com/10568965/232987994-c2864af0-d3cb-4a65-b238-0e1a7a1b989c.png">

<img width="1292" alt="Screenshot 2023-04-19 at 13 40 52" src="https://user-images.githubusercontent.com/10568965/232988070-acc1c838-5c4a-495d-9c63-237667c27f1b.png">

**Question**
- We have a state cache of 96 items, why are there some attestations with block head not in that cache, is it worth to always regen them? maybe we should skip some of them, see https://github.com/ChainSafe/lodestar/blob/3ee1d8a5acb09160827b185c79d2ead80442c0e7/packages/beacon-node/src/chain/validation/attestation.ts#L139
- Consider reducing REGEN_CAN_ACCEPT_WORK_THRESHOLD (right now it's 16) cc @dapplion

twoeths:
when I zoom in, some Job Wait Time is > 4s which is too big

<img width="1287" alt="Screenshot 2023-04-19 at 15 00 19" src="https://user-images.githubusercontent.com/10568965/233008742-20fdf910-d821-4db8-9d46-5517b7c7857a.png">

dapplion:
@tuyennhv for this type of metric is best to capture the histogram to understand the real distribution of values through a longer timeframe. That would allow to answer the question: what % of regen jobs over a day take > 1s

twoeths:
@dapplion it's about 25% of them take >1s

[...]

PR #5390, opened 2023-04-20
Add TOO_MANY_SKIPPED_SLOTS attestation error

twoeths:
**Motivation**

- Regen module is overloaded, see #5384
- We want to have DOS protection to avoid having to process attestations with very old head block that can potentially cause our regen to be overloaded

**Description**

- Follow Lighthouse to add `TOO_MANY_SKIPPED_SLOTS` error
- By default, configure `maxSkipSlots` of chain to be 32:
  - for gossip block validation, it's unlikely we see a reorg with 32 slots
  - for attestation validation, having this value ensures we don't have to regen states most of the time (because attestation has to be in the last 32 slots + `maxSkipSlots` of 32 above => headBlock of attestation should be in the last 64 slots and it's likely we can get its state from the cache). Also most of Attestations (>90%) are caught by the `SeenAttestationDatas` cache, it's no problem not to import these weird attestations.
- Import block: add postState to state cache before we emit "block" event since some handlers need that
- Add metrics to track this new error

part of #5384

### DIFF
--- a/packages/beacon-node/src/chain/blocks/importBlock.ts
+++ b/packages/beacon-node/src/chain/blocks/importBlock.ts
@@ -89,7 +89,12 @@ export async function importBlock(
     this.clock.currentSlot,
     executionStatus
   );
-  this.logger.verbose("Added block to forkchoice", {slot: block.message.slot, root: blockRootHex});
+
+  // This adds the state necessary to process the next block
+  // Some block event handlers require state being in state cache so need to do this before emitting EventType.block
+  this.stateCache.add(postState);
+
+  this.logger.verbose("Added block to forkchoice and state cache", {slot: block.message.slot, root: blockRootHex});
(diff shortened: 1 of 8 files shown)


twoeths:
**feat1**
- Was able to cache some attestations with very old block head

Screenshot 2023-04-20 at 19 32 14
- While there is no cache miss due to `getState()` entry point
Screenshot 2023-04-20 at 19 34 30

**unstable**
- It shows there are always some cache miss due to `getState()` consistently (from attestation validation) and it'll cause regen
Screenshot 2023-04-20 at 19 35 26

g11tech (approved):
lgtm

nflaig (approved):
lgtm, great comments, makes it even for me easy to understand 👍

wemeetagain (approved):

wemeetagain:

PR #5401, opened 2023-04-22
Only call regen.getBlockSlotState() if necessary

### LINKED ISSUE #5400: High voluntary_exit validation time
**Describe the bug**

Validation time of `voluntary_exit` message is sometimes very high although its logic is simple

Screenshot 2023-04-22 at 13 29 16

probably it's due to the Job Wait time when we try to get head state at current epoch
Screenshot 2023-04-22 at 13 30 44

the bigger concern is we may use the regen queue unnecessarily here, a lot of time head state should be in the same epoch so we don't need the regen queue. Note that `getBlockSlotState() api` for gossip block validation should be as fast as possible so other consumers should not use this api unless it's really necessary

**Expected behavior**

Low validation time consistently

twoeths:
**Motivation**

- Regen Job Wait Time is sometimes high, see https://github.com/ChainSafe/lodestar/issues/5384
  - For validator_exit gossip message, sometimes it takes up to 4s due to this, see https://github.com/ChainSafe/lodestar/issues/5400 , note that we can validate up to 128 `validator_exit` messages at the same time
  - This api is used by gossip block validation too so we don't want it to be slow

**Description**

- `regen.getBlockSlotState()` always use queue implementation so we use it with care, only call it if necessary
- Modify/add `getHeadStateAtCurrentEpoch()` and `getHeadStateAtEpoch()` apis in chain so it only calls `regen` if necessary, using the `checkpointStateCache` (used inside `getHeadState()`) it'll reach the queue 1 time at most per epoch
- Add RegenCaller for voluntary_exit validation

close #5400

### DIFF
--- a/packages/beacon-node/src/chain/chain.ts
+++ b/packages/beacon-node/src/chain/chain.ts
@@ -333,11 +333,22 @@ export class BeaconChain implements IBeaconChain {
     return headState;
   }
 
-  async getHeadStateAtCurrentEpoch(): Promise<CachedBeaconStateAllForks> {
-    const currentEpochStartSlot = computeStartSlotAtEpoch(this.clock.currentEpoch);
+  async getHeadStateAtCurrentEpoch(regenCaller: RegenCaller): Promise<CachedBeaconStateAllForks> {
+    return this.getHeadStateAtEpoch(this.clock.currentEpoch, regenCaller);
+  }
+
+  async getHeadStateAtEpoch(epoch: Epoch, regenCaller: RegenCaller): Promise<CachedBeaconStateAllForks> {
+    // using getHeadState() means we'll use checkpointStateCache if it's available
+    const headState = this.getHeadState();
+    // head state is in the same epoch, or we pulled up head state already from past epoch
+    if (epoch <= computeEpochAtSlot(headState.slot)) {
+      // should go to this most of the time
+      return headState;
+    }
+    // only use regen queue if necessary, it'll cache in checkpointStateCache if regen gets through epoch transition
     const head = this.forkChoice.getHead();
-    const bestSlot = currentEpochStartSlot > head.slot ? currentEpochStartSlot : head.slot;
-    return this.regen.getBlockSlotState(head.blockRoot, bestSlot, {dontTransferCache: true}, RegenCaller.getDuties);
+    const startSlot = computeStartSlotAtEpoch(epoch);
+    return this.regen.getBlockSlotState(head.blockRoot, startSlot, {dontTransferCache: true}, regenCaller);
   }
 
   async getCanonicalBlockAtSlot(slot: Slot): Promise<allForks.SignedBeaconBlock | null> {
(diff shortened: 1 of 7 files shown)


twoeths:
Tested in the last 5h on test mainnet node (feat2 - this branch vs unstable)

- feat2: voluntary_exit Job Time is < 200ms consistently

Screenshot 2023-04-22 at 21 10 38

- unstable: some takes up to 3.5s
Screenshot 2023-04-22 at 21 10 58

- feat2: Regen Job Wait Time: no item > 500ms
Screenshot 2023-04-22 at 21 35 01

- unstable: Regen Job Wait Time: a couple of items > 500ms

Screenshot 2023-04-22 at 21 35 59


- feat2: Regen job queue length is 2 most of the time (1 for PrepareNextSlot and 1 for validateGossipBlock)

Screenshot 2023-04-22 at 21 40 08

- unstable: randomly has job queue length >2 depending on voluntary_exit messages

Screenshot 2023-04-22 at 21 41 56

wemeetagain (approved):
lgtm

wemeetagain:

packages/beacon-node/src/chain/chain.ts
```diff
@@ -333,11 +333,22 @@ export class BeaconChain implements IBeaconChain {
...
-    const currentEpochStartSlot = computeStartSlotAtEpoch(this.clock.currentEpoch);
+  async getHeadStateAtCurrentEpoch(regenCaller: RegenCaller): Promise<CachedBeaconStateAllForks> {
+    return this.getHeadStateAtEpoch(this.clock.currentEpoch, regenCaller);
+  }
+
+  async getHeadStateAtEpoch(epoch: Epoch, regenCaller: RegenCaller): Promise<CachedBeaconStateAllForks> {
+    // using getHeadState() means we'll use checkpointStateCache if it's available
+    const headState = this.getHeadState();
+    // head state is in the same epoch, or we pulled up head state already from past epoch
+    if (epoch <= computeEpochAtSlot(headState.slot)) {
```
nflaig, line 344: @tuyennhv shouldn't this be a strict equals check? I noticed this method always returns the same state if a previous epoch is passed in

twoeths: @nflaig no this should be "<=", the function means to get head state dialing to an epoch. If provided epoch < head state epoch just return head state

also if we let it pass through the below `regen.getBlockSlotState` call, if provided slot < head state slot it'll throw error

nflaig: That makes sense, I was trying to figure out what would be required to get proposer duties of a finalized epoch. The right method to use in that case would be `getStateBySlot` I'd assume. 
https://github.com/ChainSafe/lodestar/blob/f77b774be201b30a4451435ad90a48c5dee48029/packages/beacon-node/src/chain/chain.ts#L362
 You also noted in https://github.com/ChainSafe/lodestar/issues/5846 is that Lodestar does not persists finalized states very frequently which would be an issue in that case. We could solve this issue if we add `--chain.archiveStateEpochFrequency 1` flag but this significantly increases storage requirements (but that might be fine for someone that really needs that data).

PR #5417, opened 2023-04-25
refactor(reqresp)!: support byte based handlers

### LINKED ISSUE #5378: ReqResp encoded bytes API
- Issue https://github.com/ChainSafe/lodestar/pull/5229 would benefit greatly from ReqResp dealing with non deserialized SSZ data structures

Since ReqResp requests and responses must cross the worker boundary it's best to just wait to de-serialize them on the main thread. For it the API exposed by the reqresp package must deal exclusively with Uint8Array instead of T representing the deserialized object. For reference of how the API should look like PR https://github.com/ChainSafe/lodestar/pull/5229 is quasi complete reference.

nazarhussain:
**Motivation**

Update `reqresp` package to only support the `bytes` for handler. 

**Description**

To make `network` more flexible to work form the worker, need to make the reqresp more consistent. 

Closes #5378

- Run all tests


nflaig:
Regarding the PR title, if a change is specific to a package, we could use the package name as scope e.g. `refactor(reqresp)`. This allows to quickly search for commits for a specific subpackage.

dapplion:
Pushed a significant refactor to align this to what is eventually needed to work within a worker thread. We really need to have this asap so please review correctness and not organizational approaches that we can review latter after landing the worker thread issue

packages/reqresp/src/response/index.ts
```diff
@@ -77,24 +77,28 @@ export async function handleRequest<Req, Resp>({
           }
         });
 
-        logger.debug("Req  received", {...logCtx, body: protocol.renderRequestBody?.(requestBody)});
+        logger.debug("Req  received", logCtx);
 
-        const requestCount = protocol?.inboundRateLimits?.getRequestCount?.(requestBody) ?? 1;
+        // TODO: Consider counting max count by request for byRange and byRoot
+        // const requestCount = protocol?.inboundRateLimits?.getRequestCount?.(requestBody) ?? 1;
+        const requestCount = 1;
```
twoeths: this is a side effect of this change regarding rate limiter. Can we decode requests before this as before? Otherwise we should track this in a github issue

packages/beacon-node/src/network/reqresp/handlers/index.ts
```diff
@@ -13,48 +14,54 @@ import {onLightClientUpdatesByRange} from "./lightClientUpdatesByRange.js";
...
   return {
-    async *onStatus(protocol) {
-      yield* onStatus(protocol, chain);
+    async *onStatus() {
+      yield* onStatus(chain);
     },
-    async *onBeaconBlocksByRange(protocol, req) {
-      yield* onBeaconBlocksByRange(protocol, req, chain, db);
+    async *onBeaconBlocksByRange(req) {
+      const body = ssz.phase0.BeaconBlocksByRangeRequest.deserialize(req.data);
```
twoeths, line 37: same to previous comment, why do we have to deserialize each protocol separately here instead of in `handleRequest()` method of reqresp package?

packages/beacon-node/src/network/reqresp/types.ts
```diff
@@ -34,6 +39,88 @@ type RequestBodyByMethod = {
...
+  [ReqRespMethod.BeaconBlockAndBlobsSidecarByRoot]: deneb.SignedBeaconBlockAndBlobsSidecar;
+  [ReqRespMethod.LightClientBootstrap]: altair.LightClientBootstrap;
+  [ReqRespMethod.LightClientUpdatesByRange]: altair.LightClientUpdate;
+  [ReqRespMethod.LightClientFinalityUpdate]: altair.LightClientFinalityUpdate;
+  [ReqRespMethod.LightClientOptimisticUpdate]: altair.LightClientOptimisticUpdate;
+};
+
+/** Request SSZ type for each method and ForkName */
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/explicit-function-return-type
+export function getRequestSzzTypeByMethod(method: ReqRespMethod) {
```
twoeths: typo
```suggestion
export function getRequestSszTypeByMethod(method: ReqRespMethod) {
```

twoeths (dismissed):
Generally this looks good to me, there's a typo and a question about the rate limiter change

packages/beacon-node/src/network/reqresp/handlers/index.ts
wemeetagain: I think its because of the desired architecture here: https://github.com/ChainSafe/lodestar/issues/5378#issuecomment-1528125740

Handlers will be in the main thread and the `handleRequest` will be in the network thread.

twoeths: thanks for clarifying @wemeetagain , that makes a lot of sense 👍

packages/reqresp/src/response/index.ts
dapplion: Re-added

dapplion (approved):

wemeetagain:

Issue #5420, opened 2023-04-26
Performance issue when there are too many mesh peers


twoeths:
**Describe the bug**

Right now mesh peer param Dhigh=12, there could be too many mesh peers per subnet on a test mainnet node

<img width="802" alt="Screenshot 2023-04-26 at 09 12 03" src="https://user-images.githubusercontent.com/10568965/234448448-bb2b9941-68a7-4643-82b4-b90c6e7fa249.png">

Gossip Block process time

<img width="1476" alt="Screenshot 2023-04-26 at 09 16 11" src="https://user-images.githubusercontent.com/10568965/234449024-26cac769-35a4-4e21-ac1c-cd932e67c8d4.png">

Gossip Block Received Delay

<img width="1425" alt="Screenshot 2023-04-26 at 09 24 22" src="https://user-images.githubusercontent.com/10568965/234450224-fa57bd4c-1674-4caf-9c42-0fb3e87e23e3.png">

If we configure `network.gossipsubDHigh=9` so that there is no subnet has more than 9 mesh peers

<img width="803" alt="Screenshot 2023-04-26 at 09 18 17" src="https://user-images.githubusercontent.com/10568965/234449337-0e3f11ca-3e4f-415d-9d08-77fbd96d382d.png">

we got better metric

<img width="1496" alt="Screenshot 2023-04-26 at 09 19 56" src="https://user-images.githubusercontent.com/10568965/234449554-35055cfd-359b-4a84-85e1-8ca78fd48991.png">

<img width="1419" alt="Screenshot 2023-04-26 at 09 24 42" src="https://user-images.githubusercontent.com/10568965/234450269-279c1140-37ac-453b-890c-d4a4fadd9c35.png">

**Expected behavior**

Lodestar should able to handle high number of mesh peers per spec (with Dhigh=12)

twoeths:
`gossipsubDHigh` was set back to 12, closing the issue

PR #5427, opened 2023-04-27
fix: set gossipsubDHigh as 9

twoeths:
**Motivation**

When there are too many mesh peers, gossip block cannot be processed timely, see https://github.com/ChainSafe/lodestar/issues/5420

**Description**

- Set `gossipsubDHigh=9` for now, should work on removing it on upcoming release and keep the issue open

part of #5420


dapplion (approved):
Chatted offline for rationale, let's merge for now temporarily while Lodestar attestation process queue is optimized to handle full mainnet traffic. Then revert to spec value

wemeetagain:

PR #5432, opened 2023-04-28
fix: update builder mergemock sim with builder block assertions

g11tech:
Previously builder sim was running without any validations.

Modified the mergemock builder interaction to enable differentiating by fee recipient and use that to add engine and builder block assertions

### DIFF
--- a/packages/beacon-node/test/sim/mergemock.test.ts
+++ b/packages/beacon-node/test/sim/mergemock.test.ts
@@ -206,7 +206,7 @@ describe("executionEngine / ExecutionEngineHttp", function () {
     let builderBlocks = 0;
     await new Promise<void>((resolve, _reject) => {
       bn.chain.emitter.on(routes.events.EventType.block, async (blockData) => {
-        const {data: fullOrBlindedBlock} = (await bn.api.beacon.getBlockV2(blockData.block).catch((_e) => null)) ?? {};
+        const {data: fullOrBlindedBlock} = await bn.api.beacon.getBlockV2(blockData.block);
         if (fullOrBlindedBlock !== undefined) {
           const blockFeeRecipient = toHexString(
             (fullOrBlindedBlock as bellatrix.SignedBeaconBlock).message.body.executionPayload.feeRecipient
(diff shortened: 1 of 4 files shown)


packages/beacon-node/test/sim/mergemock.test.ts
```diff
@@ -159,7 +172,7 @@ describe("executionEngine / ExecutionEngineHttp", function () {
       defaultConfig: {
         graffiti: "default graffiti",
         strictFeeRecipientCheck: true,
-        feeRecipient: "0xcccccccccccccccccccccccccccccccccccccccc",
+        feeRecipient: "0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
```
dapplion: Store this addresses in variables to link param setting and assertion. With names like: feeRecipientHappyCase, feeRecipientLocal, feeRecipientMevBoost, etc

g11tech: sounds good updated :+1:

packages/beacon-node/test/sim/mergemock.test.ts
```diff
@@ -183,7 +202,22 @@ describe("executionEngine / ExecutionEngineHttp", function () {
       await Promise.all(validators.map((v) => v.close()));
     });
 
+    let engineBlocks = 0;
+    let builderBlocks = 0;
     await new Promise<void>((resolve, _reject) => {
+      bn.chain.emitter.on(routes.events.EventType.block, async (blockData) => {
+        const {data: fullOrBlindedBlock} = (await bn.api.beacon.getBlockV2(blockData.block).catch((_e) => null)) ?? {};
```
dapplion: If there's a block event there must be a block available in the API right? This catch all to null seems dangerous

packages/beacon-node/test/sim/mergemock.test.ts
```diff
@@ -221,6 +255,14 @@ describe("executionEngine / ExecutionEngineHttp", function () {
       );
     }
 
+    // 2. builder and engine blocks are as expected
+    if (builderBlocks < expectedBuilderBlocks || engineBlocks < expectedEngineBlocks) {
+      throw Error(
+        `Incorrect builderBlocks=${builderBlocks} (expected=${expectedBuilderBlocks}) or engineBlocks=${engineBlocks} (expected=${expectedEngineBlocks})`
+      );
+    }
```
dapplion, line 266: Split this into two separate ifs and throws

packages/beacon-node/test/sim/mergemock.test.ts
```diff
@@ -221,6 +255,14 @@ describe("executionEngine / ExecutionEngineHttp", function () {
...
+    console.log({engineBlocks, builderBlocks});
```
dapplion: Log clutter must be controlled, is this console log really necessary?

packages/beacon-node/test/sim/mergemock.test.ts
g11tech: it is to just update the stats, but nevertheless removed the catch :+1:

packages/beacon-node/test/sim/mergemock.test.ts
g11tech: cleaned up

packages/beacon-node/test/sim/mergemock.test.ts
g11tech: done :+1:

dapplion (approved):

wemeetagain:

PR #5435, opened 2023-04-29
fix(validator)!: align db metric names with other metrics

nflaig:
**Motivation**

Rename metric names to improve consistency.

**Description**

Renamed validator db metrics to be more inline with other validator metrics, those metrics now use `vc_` as prefix instead of `validator_`.

This is a breaking change to the metrics but I am assuming that there are no external dashboards that consume these metrics and therefore should be fine to change the name now. Still tagging as breaking change.

**Note:** Those metrics were introduced in https://github.com/ChainSafe/lodestar/pull/4121 but I don't think using `validator_` as prefix was intentional.

### DIFF
--- a/dashboards/lodestar_vm_host.json
+++ b/dashboards/lodestar_vm_host.json
@@ -4369,7 +4369,7 @@
                 "uid": "${DS_PROMETHEUS}"
               },
               "exemplar": false,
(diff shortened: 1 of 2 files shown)


wemeetagain (approved):

wemeetagain:

PR #5436, opened 2023-04-29
feat(validator): add bucket labels to db metrics

nflaig:
**Motivation**

Without bucket labels it is hard to investigate issues such as https://github.com/ChainSafe/lodestar/issues/5356.

**Description**

Adds bucket labels to validator db metrics. This makes it more consistent with BN db metrics and allows to better investigate issues.

**Before**

image

**After**

image

Issue #5356, opened 2023-04-11
Validator db read/write ops too high on first startup with empty db


nflaig:
**Describe the bug**

Validator db read/write ops too high on first startup with empty db which causes I/O lag and instability in VC. After one or two hours the process stabilizes and read/write ops per second drastically decrease.

**Expected behavior**

VC should not be unstable for one or two hours if started with an empty db.

**Steps to Reproduce**

Run VC with a lot of keys (500+) and make sure validator-db is empty. If VC was active before and db was removed, be careful to not get slashed as this also removes the slashing protection db.

**Metrics**

![image](https://user-images.githubusercontent.com/38436224/231142907-383ff81c-c90f-424f-bb06-d7956d389ca9.png)

![image](https://user-images.githubusercontent.com/38436224/231146316-e59d1ad1-1b11-46e2-b803-16f3b557b36e.png)

![image](https://user-images.githubusercontent.com/38436224/231146425-b146d21c-1157-4342-84e3-3b2db3db7843.png)

dapplion:
Can you paste the promquery code for the first metrics screen capture?

nflaig:
I just changed the prom query to use the vc metrics, the rate interval is 1m, would probably have to add bucket labels to further investigate

- `rate(validator_db_read_items_total[$rate_interval])` (top left)
- `rate(validator_db_write_items_total[$rate_interval])` (top right)
- `rate( validator_db_read_req_total [$rate_interval])` (bottom left)
- `rate( validator_db_write_req_total [$rate_interval])` (bottol right)

nflaig:
[...]

PR #5437, opened 2023-04-29
fix(slashing-protection): filter by db bucket when loading pubkeys

### LINKED ISSUE #4261: slashing protection exports goes OOM while exporting 500 validators kiln data
From: https://github.com/ChainSafe/lodestar/pull/4257#issuecomment-1177936632
On our kiln node with 500 dbs which has `8GB` ram, but the process got killed as it went OOM:
```typescript
Jul-07 16:33:38.807[]                 info: Extracting the slashing protection logs version=4, format=complete, dbPath=/root/kiln-data/lodestar/validator-db, file=/root/kiln-data/lodestar/spexport.json
Jul-07 16:33:38.885[]              verbose: Fetching the pubkeys from the slashingProtection db
./lodestar: line 7: 2437989 Killed                  node --trace-deprecation --max-old-space-size=4096 ./packages/cli/bin/lodestar.js "$@"
```

as per our new logs, it happened while listing pub keys from the `slashingProtection` module, it could be same issue that affected the validator decoding in the validator start that we fixed via serializing.

Investigate and fix, the run shouldn't go OOM

nflaig:
**Motivation**

Closes https://github.com/ChainSafe/lodestar/issues/4261

**Description**

Applies min-max key filter to db query to only get pubkeys from bucket used by repository. The OOM mentioned in https://github.com/ChainSafe/lodestar/issues/4261 was caused by a combination of https://github.com/ChainSafe/lodestar/issues/5356 (causing the db state to be quite huge) and the db query not applying any filtering.

**Note:** There is currently no limit on how many attestation are stored in the db. In some extrem cases could still run into OOM issues. This is likely only relevant for big operators that run many validators but pruning strategy for attestation would be something useful to implement in the future.

https://github.com/ChainSafe/lodestar/blob/f0dec8c89957ef0653bb40e3a969750db8d0a869/packages/validator/src/slashingProtection/attestation/index.ts#L46

Other solution that can be implemented is to provide the option to just export the slashing protection for a subset of validators

https://github.com/ChainSafe/lodestar/blob/f0dec8c89957ef0653bb40e3a969750db8d0a869/packages/cli/src/cmds/validator/slashingProtection/export.ts#L49

Depends on https://github.com/ChainSafe/lodestar/pull/5436 to be merged first to avoid conflicts.

### DIFF
--- a/packages/validator/src/slashingProtection/attestation/attestationByTargetRepository.ts
+++ b/packages/validator/src/slashingProtection/attestation/attestationByTargetRepository.ts
@@ -45,10 +45,10 @@ export class AttestationByTargetRepository {
 
   async getAll(pubkey: BLSPubkey, limit?: number): Promise<SlashingProtectionAttestation[]> {
     const attestations = await this.db.values({
-      ...this.dbReqOpts,
       limit,
       gte: this.encodeKey(pubkey, 0),
       lt: this.encodeKey(pubkey, Number.MAX_SAFE_INTEGER),
+      bucketId: this.bucketId,
     });
     return attestations.map((attestation) => this.type.deserialize(attestation));
   }
@@ -69,7 +69,7 @@ export class AttestationByTargetRepository {
   }
 
   async listPubkeys(): Promise<BLSPubkey[]> {
-    const keys = await this.db.keys({...this.dbReqOpts, gte: this.minKey, lt: this.maxKey});
+    const keys = await this.db.keys({gte: this.minKey, lt: this.maxKey, bucketId: this.bucketId});
     return uniqueVectorArr(keys.map((key) => this.decodeKey(key).pubkey));
   }
 
--- a/packages/validator/src/slashingProtection/attestation/attestationByTargetRepository.ts
+++ b/packages/validator/src/slashingProtection/attestation/attestationByTargetRepository.ts
@@ -27,6 +27,9 @@ export class AttestationByTargetRepository {
   private readonly bucketId: string;
   private readonly dbReqOpts: DbReqOpts;
 
+  private readonly minKey: Uint8Array;
+  private readonly maxKey: Uint8Array;
+
   constructor(opts: DatabaseApiOptions) {
     this.db = opts.controller;
     this.type = new ContainerType({
@@ -36,6 +39,8 @@ export class AttestationByTargetRepository {
     }); // casing doesn't matter
     this.bucketId = getBucketNameByValue(this.bucket);
     this.dbReqOpts = {bucketId: this.bucketId};
+    this.minKey = encodeKey(this.bucket, Buffer.alloc(0));
+    this.maxKey = encodeKey(this.bucket + 1, Buffer.alloc(0));
   }
 
   async getAll(pubkey: BLSPubkey, limit?: number): Promise<SlashingProtectionAttestation[]> {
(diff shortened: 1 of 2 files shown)


packages/validator/src/slashingProtection/attestation/attestationByTargetRepository.ts
```diff
@@ -36,14 +39,16 @@ export class AttestationByTargetRepository {
...
+    this.maxKey = encodeKey(this.bucket + 1, Buffer.alloc(0));
   }
 
   async getAll(pubkey: BLSPubkey, limit?: number): Promise<SlashingProtectionAttestation[]> {
     const attestations = await this.db.values({
-      ...this.dbReqOpts,
       limit,
       gte: this.encodeKey(pubkey, 0),
       lt: this.encodeKey(pubkey, Number.MAX_SAFE_INTEGER),
+      bucketId: this.bucketId,
```
nflaig, line 51: Unrelated change, just minor follow up to https://github.com/ChainSafe/lodestar/pull/5436

`FilterOptions` does not extend `DbReqOpts` and therefore is not guaranteed to be a superset of it, to avoid type issues in the future should rather not destructure `this.dbReqOpts` here

https://github.com/ChainSafe/lodestar/blob/3ed8d04c7971f753570f0d054a3c44bc2c382be1/packages/db/src/controller/interface.ts#L10

also inline with implementation here

https://github.com/ChainSafe/lodestar/blob/3ed8d04c7971f753570f0d054a3c44bc2c382be1/packages/db/src/abstractRepository.ts#L249-L252

packages/validator/src/slashingProtection/block/blockBySlotRepository.ts
```diff
@@ -68,7 +73,7 @@ export class BlockBySlotRepository {
   }
 
   async listPubkeys(): Promise<BLSPubkey[]> {
-    const keys = await this.db.keys(this.dbReqOpts);
+    const keys = await this.db.keys({gte: this.minKey, lt: this.maxKey, bucketId: this.bucketId});
```
dapplion, line 76: How is this not taken care of by the underlying db repository impl?

nflaig: The repositories on the validator do not extend [Repository](https://github.com/ChainSafe/lodestar/blob/3ed8d04c7971f753570f0d054a3c44bc2c382be1/packages/db/src/abstractRepository.ts#L19) and the [db itself](https://github.com/ChainSafe/lodestar/blob/3ed8d04c7971f753570f0d054a3c44bc2c382be1/packages/db/src/controller/level.ts#L143) does not handle that.

This is mostly do to the fact that there are incompatibilities with parameters and returns types which does not allow to override methods such as `encodeKey`.

Another more historic decision I found is that it was desired to keep db interface required by validator to be more minimal.
As mentioned in https://github.com/ChainSafe/lodestar/pull/1954 and a separate type [LodestarValidatorDatabaseController](https://github.com/ChainSafe/lodestar/blob/3ed8d04c7971f753570f0d054a3c44bc2c382be1/packages/validator/src/types.ts#L22).

Maybe a good refactor would be to create a base repository for the validator repositories with a minimal set of methods.

nflaig: I can look into creating a base repository for subset of methods required by validator repositories. Will have to see if that refactor even makes sense but would prefer to do it in another PR.

nflaig:
Just to reconfirm, exporting a 200MB slashing protection file works on a VC with 2k keys, would consider the OOM issue fixed.

packages/validator/src/slashingProtection/block/blockBySlotRepository.ts
dapplion: I'll merge this for now, and feel free to open PR to dedup

dapplion (approved):

wemeetagain:

Issue #5379, opened 2023-04-18
ReqResp async iterable to events


dapplion:
- PR https://github.com/ChainSafe/lodestar/pull/5229 **requires** to handle ReqResp request as events, since messages must be sent via worker parent port

Since the reqresp package is tightly integrated with libp2p and expects async iterables there must be some adopter code in between. IMO async iterables are great for DX to write the handlers and network code so we should keep it. What's necessary is:
- [ ] Provide a middleware for the handler: convert async iterable -> events -> async iterable
- [ ] Provide a middleware for the request: convert async iterable -> events -> async iterable

A sample implementation with context of how this will be used can be found here packages/beacon-node/src/network/reqresp/utils/handlerToEvents.ts but as long as the full API of async iterables is covered, and memory is bounded another implementation could be used.

dapplion:
Implemented with libp2p thread

PR #5442, opened 2023-05-01
refactor(reqresp)!: add ReqResp async iterable to events

dapplion:
**Motivation**

- See https://github.com/ChainSafe/lodestar/issues/5379

**Description**

Implement async iterable bridge to handle both incoming and outgoing requests. Emitter and handler are in the same thread, which is currently pointless but necessary for moving libp2p to a thread.


dapplion:
Superseeded by libp2p thread PR

Issue #5336, opened 2023-04-04
Expose validator monitor via logs


nflaig:
**Description**

Currently, we just expose validator monitor via metrics but it would be desired to also expose this information via logs.
Potentially opt-in behind a feature flag such as `--validator-monitor-log`.

**Example**

Lighthouse validator monitor logs example (when adding `--validator-monitor-auto` flag):

`INFO Previous epoch attestation(s) success   validators: ["xxx"], epoch: 192112, service: val_mon`

`INFO Previous epoch attestation(s) failed to match head, validators: ["xxx"], epoch: 192096, service: val_mon`
(same for source, target and suboptimal inclusion distance) or if missing entirely. Happens at the start of each epoch.

`INFO Unaggregated attestation                validator: xxxx, src: api, slot: 6147625, epoch: 192113, delay_ms: 39, index: 42, head: 0x9eecb3839976ccf00c8635ed77302077ae858e39cba5cc8d8ba48aefd35baba3, service: val_mon`
When attestation is submitted and then afterwards:

`INFO Attestation included in aggregate       validator: xxxx, src: gossip, slot: 6147625, epoch: 192113, delay_ms: 207, index: 42, head: 0x9eecb3839976ccf00c8635ed77302077ae858e39cba5cc8d8ba48aefd35baba3, service: val_mon`

` INFO Attestation included in block           validator: xxxx, slot: 6147625, epoch: 192113, inclusion_lag: 0 slot(s), index: 42, head: 0x9eecb3839976ccf00c8635ed77302077ae858e39cba5cc8d8ba48aefd35baba3, service: val_mon`

**Related**

- https://discord.com/channels/593655374469660673/593655641445367808/1092417950092378122
- https://discord.com/channels/593655374469660673/593655641445367808/1092724003879272458

nflaig:
Based on [discussion in discord](https://discord.com/channels/593655374469660673/1105347927548907600/1110574619292872765), the validator monitor logs should be disabled if node is still syncing (to be considered) or if no validators are attached.

It also needs to be further discussed if we want validator monitor logs by default or opt-in via feature flag.

nflaig:
Lighthouse documentation on [validator monitoring](https://lighthouse-book.sigmaprime.io/validator-monitoring.html) can serve as a guide on how we want to implement this. Their reasoning and implementation seems to be really well defined and thought through.

PR #5448, opened 2023-05-01
feat: collect and add proposal stats in grafana metrics

### LINKED ISSUE #4636: Metric to track missed blocks
Guarding against re-orgs, track missed blocks and register as metric both at network level and to validator monitor

g11tech:
Compute and add proposal stats in grafana metrics on finalization

TODO:
 - [x] Test extensively
 - [x] Add Grafana metrics
   image
 - [x] Add unit tests
 
 Closes https://github.com/ChainSafe/lodestar/issues/4636

### DIFF
--- a/packages/beacon-node/src/chain/archiver/index.ts
+++ b/packages/beacon-node/src/chain/archiver/index.ts
@@ -25,6 +25,14 @@ type ProposalStats = {
   missed: number;
 };
 
+export type FinalizedStats = {
+  allValidators: ProposalStats;
+  attachedValidators: ProposalStats;
+  finalizedCanonicalCheckpointsCount: number;
+  finalizedFoundCheckpointsInStateCache: number;
+  finalizedAttachedValidatorsCount: number;
+};
+
 /**
  * Used for running tasks that depends on some events or are executed
  * periodically.
@@ -174,13 +182,7 @@ export class Archiver {
     finalizedData: FinalizedData,
     finalized: CheckpointWithHex,
     lastFinalized: CheckpointWithHex
-  ): {
-    allValidators: ProposalStats;
-    attachedValidators: ProposalStats;
-    finalizedCanonicalCheckpointsCount: number;
-    finalizedFoundCheckpointsInStateCache: number;
-    finalizedAttachedValidatorsCount: number;
-  } {
+  ): FinalizedStats {
     const {finalizedCanonicalCheckpoints, finalizedCanonicalBlocks, finalizedNonCanonicalBlocks} = finalizedData;
 
     // Range to consider is:
--- a/packages/beacon-node/src/chain/archiver/index.ts
+++ b/packages/beacon-node/src/chain/archiver/index.ts
@@ -1,6 +1,6 @@
 import {Logger} from "@lodestar/utils";
-import {CheckpointWithHex, IForkChoice, ProtoBlock} from "@lodestar/fork-choice";
-import {Epoch, ValidatorIndex, Slot} from "@lodestar/types";
+import {CheckpointWithHex, IForkChoice} from "@lodestar/fork-choice";
+import {ValidatorIndex, Slot} from "@lodestar/types";
 import {SLOTS_PER_EPOCH} from "@lodestar/params";
 
 import {IBeaconDb} from "../../db/index.js";
@@ -194,7 +194,7 @@ export class Archiver {
     finalizedProposersCheckpoints.push(lastFinalized);
     finalizedProposersCheckpoints.push(finalized);
 
-    // Sort the data to in following structure to make further inferences
+    // Sort the data to in following structure to make inferences
     const slotProposers = new Map<Slot, {canonicalVals: ValidatorIndex[]; nonCanonicalVals: ValidatorIndex[]}>();
 
     //  1. Process canonical blocks
--- a/packages/beacon-node/src/chain/archiver/index.ts
+++ b/packages/beacon-node/src/chain/archiver/index.ts
@@ -1,6 +1,6 @@
 import {Logger} from "@lodestar/utils";
 import {CheckpointWithHex, IForkChoice, ProtoBlock} from "@lodestar/fork-choice";
-import {Epoch, ValidatorIndex} from "@lodestar/types";
+import {Epoch, ValidatorIndex, Slot} from "@lodestar/types";
 import {SLOTS_PER_EPOCH} from "@lodestar/params";
 
 import {IBeaconDb} from "../../db/index.js";
@@ -18,6 +18,13 @@ export type ArchiverOpts = StatesArchiverOpts & {
   disableArchiveOnCheckpoint?: boolean;
 };
 
+type ProposalStats = {
+  expected: number;
+  finalized: number;
+  orphaned: number;
+  missed: number;
+};
+
 /**
  * Used for running tasks that depends on some events or are executed
  * periodically.
@@ -25,6 +32,7 @@ export type ArchiverOpts = StatesArchiverOpts & {
 export class Archiver {
   private jobQueue: JobItemQueue<[CheckpointWithHex], void>;
 
+  private prevFinalized: CheckpointWithHex;
   private readonly statesArchiver: StatesArchiver;
 
   constructor(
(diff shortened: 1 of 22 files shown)


packages/beacon-node/src/metrics/metrics/beacon.ts
```diff
@@ -58,6 +58,66 @@ export function createBeaconMetrics(register: RegistryMetricCreator) {
 
     // Non-spec'ed
 
+    // Finalized block and proposal stats
+    allValidators: {
+      expected: register.gauge({
+        name: "all_validators_expected_count",
```
nflaig: Would suggest prefixing metrics with either `beacon_` or `lodestar_`

packages/beacon-node/src/chain/archiver/index.ts
```diff
@@ -143,4 +166,165 @@ export class Archiver {
...
+    }
+
+    const attachedValidators: ProposalStats = {
+      expected: expectedAttachedValidatorsProposalsCount,
+      finalized: finalizedAttachedValidatorsProposalsCount,
+      orphaned: finalizedAttachedValidatorsOrphanCount,
+      missed: finalizedAttachedValidatorsMissedCount,
+    };
+
+    this.logger.info("All validators finalized stats", {
```
nflaig: is this something we always want to log to info?

packages/beacon-node/src/metrics/metrics/beacon.ts
```diff
@@ -58,6 +58,66 @@ export function createBeaconMetrics(register: RegistryMetricCreator) {
 
     // Non-spec'ed
 
+    // Finalized block and proposal stats
+    allValidators: {
```
nflaig: Just reading the metric name `all_validators_expected_count` or `allValidators.expected` does not make it obvious that this is about blocks

nflaig:
Should this PR also add dashboard panels in grafana to visualize the metrics? The title suggests that but right now it only added the prometheus metrics

packages/beacon-node/src/chain/archiver/index.ts
g11tech: finalized stats people should find useful in normal usage

g11tech:
> Should this PR also add dashboard panels in grafana to visualize the metrics? The title suggests that but right now it only added the prometheus metrics

updated

packages/beacon-node/src/metrics/metrics/beacon.ts
wemeetagain: `beacon_` prefix usually means its a standardized metric (across clients)
`lodestar_` prefix means its a lodestar beacon node metric.

We should prefix these with `lodestar_` and also move these definitions to lodestar.ts sibling file.

packages/beacon-node/src/chain/archiver/index.ts
```diff
@@ -143,4 +174,159 @@ export class Archiver {
...
+
+    this.logger.info("All validators finalized proposal stats", {
+      ...allValidators,
+      finalizedCanonicalCheckpointsCount,
+      finalizedFoundCheckpointsInStateCache,
+    });
+    this.logger.info("Attached validators finalized proposal stats", {
+      ...attachedValidators,
+      finalizedAttachedValidatorsCount,
+    });
```
wemeetagain: Would like more eyes here @philknows @dapplion since these are new info logs and need to be extremely polished and well thought out.

My first thought is that I don't think we want the "all validators" log, and only want the "attached validators" log.
Also I think we should remove `finalizedAttachedValidatorsCount` (or at the very least rename it to be less wordy).

My justification for ^ is that info logs should probably not be added for extensive general chain metrics (rather use prom metrics for that), but rather be reserved for individual node status and individual validator statuses.

nflaig: Could just hide behind a feature flag, similar to validator monitor logs (to be added)
- https://github.com/ChainSafe/lodestar/issues/5336

g11tech: we can push all validators log (and finalizedAttachedValidatorsCount) to debug and retain attached validator logs for info 

reasoning: not all deployments have metrics set, but all deployments by default write in debug file which has been helpful in past for debugging

packages/beacon-node/test/unit/chain/archive/collectFinalizedProposalStats.test.ts
```diff
@@ -0,0 +1,207 @@
...
+      null
+    );
+  });
+
+  // Each test case is:
+  //   name, canonical slots [], non canonical slots [], finalized checkpoints [],
+  //     [prev finalized, latest finalized], attached validators[], exected stats
+  const testCases1: [string, Slot[], Slot[], Epoch[], [Epoch, Epoch], ValidatorIndex[], FinalizedStats][] = [
+    [
+      "all - 1 orphaned, 0 missed, attached - 1 orphan 0 missed",
```
g11tech: fyi: renaming them better for clearer context

dapplion (changes requested):
While the functionality is important I think this PR breaks a bunch of separation of concerns in our code that makes me very reluctant to approve
- Now archiver class is mixed with the role of the validator monitor
- Dependency between archiver / BeaconProposerCache is nasty
- Having to bloat the fork-choice data structure for a purely metric feature is also not pretty
- Archiver doing a validator status info log at non-regular intervals is nasty, which should be the responsability of the validator monitor, see https://github.com/ChainSafe/lodestar/issues/5336

Also that random `const checkpointState = checkpointStateCache.get(checkpointHex);` in the middle of the archiver is not good, as we want to eventually move to a more black-box style regen and there's no guarantee the state is available at all

I don't have good suggestions for all the points brought up but the current implementation seems expensive to maintain long term

g11tech:
> While the functionality is important I think this PR breaks a bunch of separation of concerns in our code that makes me very reluctant to approve
> 
>     * Now archiver class is mixed with the role of the validator monitor
> 
>     * Dependency between archiver / BeaconProposerCache is nasty
> 
>     * Having to bloat the fork-choice data structure for a purely metric feature is also not pretty
> 
>     * Archiver doing a validator status info log at non-regular intervals is nasty, which should be the responsability of the validator monitor, see [Expose validator monitor via logs #5336](https://github.com/ChainSafe/lodestar/issues/5336)
> 
> 
> Also that random `const checkpointState = checkpointStateCache.get(checkpointHex);` in the middle of the archiver is not good, as we want to eventually move to a more black-box style regen and there's no guarantee the state is available at all
> 
> I don't have good suggestions for all the points brought up but the current implementation seems expensive to maintain long term

hmmm, let me suggest ways to address these concerns :

1. what we need is the state is what proposers in `BeaconProposerCache` have to propose in that epoch, may be this can be computed and cached in `BeaconProposerCache` itself in prepareForNextSlot's epoch transition. 
2. move the computation to forkchoice where finalization event is thrown
3. collect proposerIndex in the block db although imo putting proposerIndex in forkchoice isn't much of the bloat but i can understand the concern

If you like this approach I can refac this

dapplion (approved):
Did a second pass and I think the approach is good. Attempted to do alternative solutions and they have their own trade-offs, specifically:
- Register in validator monitor seen blocks by attached validators
- Extend the EpochContext to include the prev epoch proposers
- Track orphan data at the end of each epoch cross referencing those data points

That approach does not require to change the fork-choice but requires another cache in the validator monitor. It think we can explore that latter, but for now this PR is good. Also there's a bug that prevents that approach from working always
- https://github.com/ChainSafe/lodestar/issues/5494

packages/beacon-node/src/chain/archiver/index.ts
nflaig: @g11tech  After seeing this log (`Attached validators finalized proposal stats expected=0, finalized=0, orphaned=0, missed=0, finalizedAttachedValidatorsCount=0`) on different beacon node instances I find it not very useful as a info log which should be printed out by default.

I think should reconsider and only enable it via flag as I suggested above. Right now, it seems really random as we do report stats on block proposals but not attestations. 

It also just prints out the proposal stats during which the beacon node was running, on restart, it will reset all values to 0. Unless someone is running hundreds of validators this log will not change for months.

Focusing the BN logs on just sync and peer info by default feels a lot cleaner to be as there might not even be validators attached.

dapplion: I do agree here

philknows: We had a discussion about this in Discord here: https://discord.com/channels/593655374469660673/1105347927548907600/1110563872152236133

The compromise made is that this will not show on `info` if node is still syncing, there are no validators attached or if all stats are 0. There will also be a flag to disable this.

wemeetagain:

PR #5450, opened 2023-05-01
refactor(network): add gossip publisher module

wemeetagain:
**Motivation**

- Spin-off from https://github.com/ChainSafe/lodestar/pull/5351

**Description**

Move `publish*` functions to a "gossip publisher" module that exists separately from the gossipsub implementation. This separation will allow us to serialize gossip objects in the main thread and then send them to the gossipsub implementation in the network thread.


dapplion:
Should the publisher API just be part of the main thread network public API? I wrote some thoughts on how the API of each components should look like

dapplion (approved):

wemeetagain:

PR #5451, opened 2023-05-01
refactor(network)!: use StatusCache in PeerManager

wemeetagain:
**Motivation**

- Spin-off from https://github.com/ChainSafe/lodestar/pull/5351

**Description**

Use `StatusCache` (and clock) in PeerManager instead of using the entire chain as a module dependency.
Note: this relaxes our peer relevance calculation by removing the check in the case that a peer is on an old finalized chain.


dapplion (approved):

wemeetagain:

PR #5229, opened 2023-03-03
feat: move libp2p to network worker thread

dapplion:
**Motivation**

- Allow network to perform time sensitive actions (gossip heartbeat, pings, etc) timely. DOS protection sandboxing resource spikes. See https://github.com/ChainSafe/lodestar/issues/4856

**Goals**

- Merge first minimum viable solution, going for simplicity first and optimization latter
- Allow libp2p to be run in main thread or worker by switching a runtime flag

**Description**

TBD

**TODO**

- [x] Wire metrics on libp2p worker
- [x] Connect NetworkEventBus through the worker boundary
- [x] Complete reqresp async iterables via events
- [x] Handle serialization/deserialization of req/resp across worker boundary
- [x] ~Move discv5 worker out of libp2p worker, and connect them via main thread~
- [x] Logger in worker with user provided settings
- [ ] Add metrics to track async iterable bridge Maps
- [x] Add tests for all possible events and methods crossing the worker boundary to programmatically ensure structures are clonable. Otherwise runtime error and Lodestar dead
- [x] Add more tests for both usage modes (in worker, outside worker)

Closes https://github.com/ChainSafe/lodestar/issues/5447

PR #5453, opened 2023-05-02
refactor!: move getNetworkIdentity to network class

dapplion:
**Motivation**

- Spin-off from https://github.com/ChainSafe/lodestar/pull/5229

**Description**

Move getNetworkIdentity logic into network class to reduce interface surface

### DIFF
--- a/packages/beacon-node/test/unit/api/impl/node/node.test.ts
+++ b/packages/beacon-node/test/unit/api/impl/node/node.test.ts
@@ -3,8 +3,6 @@ import {PeerId} from "@libp2p/interface-peer-id";
 import {expect} from "chai";
 import {multiaddr} from "@multiformats/multiaddr";
 import {createSecp256k1PeerId} from "@libp2p/peer-id-factory";
-import {createKeypairFromPeerId, SignableENR} from "@chainsafe/discv5";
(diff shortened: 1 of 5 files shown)


wemeetagain (dismissed):

packages/beacon-node/test/unit/api/impl/node/node.test.ts
```diff
@@ -43,36 +41,6 @@ describe("node api implementation", function () {
     sinon.stub(networkStub, "localMultiaddrs").get(() => [multiaddr("/ip4/127.0.0.1/tcp/36000")]);
   });
 
-  describe("getNetworkIdentity", function () {
```
matthewkeil, line 46: Does this unit test need to be reimplemented?

matthewkeil (approved):
Nice!! Just the ? about the unit test moving somewhere instead of getting deleted

packages/beacon-node/test/unit/api/impl/node/node.test.ts
dapplion: The API impl now just calls a function, a unit test for it would be pretty stupid, testing that function a calls function b

wemeetagain:

PR #5454, opened 2023-05-02
fix(slashing-protection): limit min-max surround epoch lookback to 4096

### LINKED ISSUE #5356: Validator db read/write ops too high on first startup with empty db

nflaig:
**Motivation**

Closes https://github.com/ChainSafe/lodestar/issues/5356

**Description**

Limits number of epochs in the past to check for surrounding attestations (min-span) to **4096**.

**DB size**

Running with 2k keys, the DB size was initially **4GB** and is now reduced to **100MB**.

**Metrics**

These are the metrics when running a VC with 2k keys and an empty db.

There are still quite a lot of read/writes but it goes down very fast compared to initially and does no longer cause instability in VC.

image

Request times do not show any I/O lag, previously was running into timeouts (10 seconds).

image

For the first 1-2 minutes there is still a increased step times diff but it goes down very fast, after a few minutes instead of hours as before.

image

Metrics look good to me now, there are no big issues and also need to keep in mind that those metrics are collected running 2k validators which is not a normal setup. This also only happens if the db is empty, meaning for most setups only on initial installation.

### DIFF
--- a/packages/validator/src/slashingProtection/minMaxSurround/minMaxSurround.ts
+++ b/packages/validator/src/slashingProtection/minMaxSurround/minMaxSurround.ts
@@ -5,13 +5,29 @@ import {SurroundAttestationError, SurroundAttestationErrorCode} from "./errors.j
 // surround vote checking with min-max surround
 // https://github.com/protolambda/eth2-surround#min-max-surround
 
+/**
+ * Number of epochs in the past to check for surrounding attestations.
+ *
+ * This value can be limited to a reasonable high amount as Lodestar does not solely rely on this strategy but also
+ * implements the minimal strategy which has been formally proven to be safe (https://github.com/michaelsproul/slashing-proofs).
+ *
+ * Limiting this value is required due to practical reasons as otherwise there would be a min-span DB read and write
+ * for each validator from current epoch until genesis which massively increases DB size and causes I/O lag, resulting in
+ * instability on first startup with an empty DB. See https://github.com/ChainSafe/lodestar/issues/5356 for more details.
+ *
+ * The value 4096 has been chosen as it is the default used by slashers (https://lighthouse-book.sigmaprime.io/slasher.html#history-length)
+ * and is generally higher than the weak subjectivity period. However, it would still be risky if we just relied on min-max surround
+ * for slashing protection, as slashers can be configured to collect slashable attestations over a longer period.
+ */
+const DEFAULT_MAX_EPOCH_LOOKBACK = 4096;
+
 export class MinMaxSurround implements IMinMaxSurround {
   private store: IDistanceStore;
   private maxEpochLookback: number;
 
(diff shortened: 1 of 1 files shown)


packages/validator/src/slashingProtection/minMaxSurround/minMaxSurround.ts
```diff
@@ -5,13 +5,29 @@ import {SurroundAttestationError, SurroundAttestationErrorCode} from "./errors.j
 // surround vote checking with min-max surround
 // https://github.com/protolambda/eth2-surround#min-max-surround
 
+/**
+ * Number of epochs in the past to check for surrounding attestations.
+ *
+ * This value can be limited to a reasonable high amount as Lodestar does not solely rely on this strategy but also
+ * implements the minimal strategy which has been formally proven to be safe (https://github.com/michaelsproul/slashing-proofs).
```
nflaig, line 12: Minimal strategy refers to [conditions 4 and 5](https://eips.ethereum.org/EIPS/eip-3076#conditions) from EIP-3076 and is implemented in Lodestar

https://github.com/ChainSafe/lodestar/blob/3ed8d04c7971f753570f0d054a3c44bc2c382be1/packages/validator/src/slashingProtection/attestation/index.ts#L99-L121

packages/validator/src/slashingProtection/minMaxSurround/minMaxSurround.ts
```diff
@@ -5,13 +5,29 @@ import {SurroundAttestationError, SurroundAttestationErrorCode} from "./errors.j
...
+ *
+ * Limiting this value is required due to practical reasons as otherwise there would be a min-span DB read and write
+ * for each validator from current epoch until genesis which massively increases DB size and causes I/O lag and instability
+ * on first startup with an empty db. See https://github.com/ChainSafe/lodestar/issues/5356 for more details on the issue.
+ *
+ * The value 4096 has been chosen as it is the default used by slashers (https://lighthouse-book.sigmaprime.io/slasher.html#history-length)
+ * and is generally higher than the weak subjectivity period. However, it would still be risky if we just relied on min-max surround
```
nflaig, line 19: Right now, the weak subjectivity period for goerli is 3686 epochs and for mainnet 3442 epochs.

wemeetagain (approved):

dapplion (approved):

wemeetagain:

PR #5456, opened 2023-05-02
refactor!: simplify discv5 initialization

wemeetagain:
**Motivation**

Simplification of initialization of discv5 for the purpose of simplifying passing the enr across a thread boundary


packages/beacon-node/src/network/options.ts
```diff
@@ -20,22 +19,14 @@ export interface NetworkOptions
...
-
 export const defaultNetworkOptions: NetworkOptions = {
   maxPeers: 55, // Allow some room above targetPeers for new inbound peers
   targetPeers: 50,
   discv5FirstQueryDelayMs: 1000,
   localMultiaddrs: ["/ip4/0.0.0.0/tcp/9000"],
   bootMultiaddrs: [],
   mdns: false,
-  discv5: defaultDiscv5Options,
+  discv5: null,
```
dapplion, line 30: Note that defaulting to null means discv5 is disabled, is that the intention? If so we should add a comment like "discv5 = null disables discv5 by default"

packages/beacon-node/src/network/peers/peerManager.ts
```diff
@@ -61,6 +61,12 @@ const ALLOWED_NEGATIVE_GOSSIPSUB_FACTOR = 0.1;
 // The Node should compute a recomended value every interval and log a warning
 // to terminal if it deviates significantly from the user's settings
 
+export type LodestarDiscv5Opts = Omit<IDiscv5CreateOptions, "enr" | "multiaddr"> & {
```
dapplion: We should not extend `IDiscv5CreateOptions` with Omit, Either select the few properties we want to expose or just use a new object. From looking into it we just need

```ts
export type Discv5Opts = {
  enr: string;
  bindAddr: string;
  bootEnrs: string[];
};
```

wemeetagain:
@dapplion ready for re-review

dapplion (approved):

wemeetagain:

Issue #3062, opened 2021-09-02
Chain logs "expected" errors


dapplion:
**Describe the bug**

In Prater fork many blocks were missed, causing many errors for the chain processor. However this errors are expected in such network conditions and should not be logged as full errors, maybe warning or just info.

```
beacon_node_1  | Sep-02 12:40:40.494 [CHAIN]           error: Block error slot=1173203 code=BLOCK_ERROR_PARENT_UNKNOWN, parentRoot=0x520990c851ba76fec4bfdbe08a06ade7da6cce87d233c72a237be4cec0da7c05
beacon_node_1  | Error: BLOCK_ERROR_PARENT_UNKNOWN
beacon_node_1  |     at Object.processBlock (/usr/app/packages/lodestar/src/chain/blocks/process.ts:36:11)
beacon_node_1  |     at processBlockJob (/usr/app/packages/lodestar/src/chain/blocks/processor.ts:66:11)
beacon_node_1  |     at JobItemQueue.BlockProcessor.jobQueue.queue_1.JobItemQueue.maxLength.maxLength [as itemProcessor] (/usr/app/packages/lodestar/src/chain/blocks/processor.ts:34:18)
beacon_node_1  |     at Timeout.JobItemQueue.runJob [as _onTimeout] (/usr/app/packages/lodestar/src/util/queue/itemQueue.ts:85:33)
beacon_node_1  |     at listOnTimeout (internal/timers.js:554:17)
beacon_node_1  |     at processTimers (internal/timers.js:497:7)
beacon_node_1  | Sep-02 12:40:40.854 [CHAIN]           error: Block error slot=1173202 code=BLOCK_ERROR_PARENT_UNKNOWN, parentRoot=0x9c583b08f4f000776b74934ddf8ab4e11c0e17e933f0f6133844c35b97115f52
beacon_node_1  | Error: BLOCK_ERROR_PARENT_UNKNOWN
beacon_node_1  |     at Object.processBlock (/usr/app/packages/lodestar/src/chain/blocks/process.ts:36:11)
beacon_node_1  |     at processBlockJob (/usr/app/packages/lodestar/src/chain/blocks/processor.ts:66:11)
beacon_node_1  |     at JobItemQueue.BlockProcessor.jobQueue.queue_1.JobItemQueue.maxLength.maxLength [as itemProcessor] (/usr/app/packages/lodestar/src/chain/blocks/processor.ts:34:18)
[...]

PR #5457, opened 2023-05-03
DNM: logging ux changes based on new logging policy

### LINKED ISSUE #4400: Tracker: Improve Lodestar error UX
This is a tracker to collect all issues regarding to improving lodestar error UX. Most of the times we are pretty aggressive in throwing error even if its intermittent/self resolvable and requires no user action/attention.

For such issues, an error accumulator should be used and only throw if accumulator exceeds threshold (accum resets and starts from 0 if the progress is made between consecutive tries)

Issues:
- [x] Block Parent Unknown https://github.com/ChainSafe/lodestar/issues/3062#issuecomment-1211457758 as raised here https://github.com/ChainSafe/lodestar/issues/3062 and https://github.com/flashbots/mev-boost/issues/157
- [x] https://github.com/ChainSafe/lodestar/issues/4348
- [x] https://github.com/ChainSafe/lodestar/issues/5938
- [x] https://github.com/ChainSafe/lodestar/issues/4396
-  [x] https://github.com/ChainSafe/lodestar/issues/4146
-  [x] https://github.com/ChainSafe/lodestar/issues/3569
-  [x] https://github.com/ChainSafe/lodestar/issues/3605
-  [x] https://github.com/ChainSafe/lodestar/issues/4522
-  [x] https://github.com/ChainSafe/lodestar/issues/4523
-  [x] https://github.com/ChainSafe/lodestar/issues/2311
-  [x] https://github.com/ChainSafe/lodestar/issues/4540
-  [x] https://github.com/ChainSafe/lodestar/issues/4559

### LINKED ISSUE #5359: Implement new logging policy
This issue is to track the implementation of #5299 into Lodestar's logging. After the Shapella hard fork, we experienced a consistent, ignorable error messages relating to the `QUEUE_ERROR_QUEUE_MAX_LENGTH` which created a poor logging UX to users of Lodestar during the fork transition due to the flow of large batches of BLStoExecution messages. This is a wake up call for us to urgently implement the new logging policy for the next attainable release to fix our poor logging UX.

There are many errors we would like to fix listed in #4400 which should now also abide by the new policy.

philknows:
**Motivation**

This PR is a collection of changes committed by the team, dedicated to discussion and refactoring based on our newly implemented logging policy #5299. Lodestar is implementing a better logging policy to improve our UX for continued adoption by the community.

Meant to track and discuss specific portions of the code that require log UX improvements. Do not merge.

**Description**

Contributors are asked to submit commits to this PR with potential changes and motivation for why a specific change is being proposed.  

Closes #5359 
Partially closes #4400

### DIFF
--- a/packages/api/test/unit/client/httpClient.test.ts
+++ b/packages/api/test/unit/client/httpClient.test.ts
@@ -128,6 +128,7 @@ describe("httpClient json client", () => {
       return Promise.reject(Error("did not throw"));
     } catch (e) {
       if (!(e instanceof HttpError)) throw Error(`Not an HttpError: ${(e as Error).message}`);
+      // Discuss error handling: https://github.com/ChainSafe/lodestar/issues/5359#issuecomment-1529027981
       expect(e.message).to.equal("Service Unavailable: Node is syncing");
       expect(e.status).to.equal(503, "Wrong error status code");
     }
--- a/packages/beacon-node/src/chain/blocks/index.ts
+++ b/packages/beacon-node/src/chain/blocks/index.ts
@@ -110,7 +110,7 @@ export async function processBlocks(
     if (!(err instanceof BlockError)) {
       this.logger.error("Non BlockError received", {}, err);
     } else if (!opts.disableOnBlockError) {
-      this.logger.error("Block error", {slot: err.signedBlock.message.slot}, err);
+      this.logger.warn("Block error", {slot: err.signedBlock.message.slot}, err);
 
       if (err.type.code === BlockErrorCode.INVALID_SIGNATURE) {
         const {signedBlock} = err;
--- a/packages/beacon-node/src/network/peers/peerManager.ts
+++ b/packages/beacon-node/src/network/peers/peerManager.ts
@@ -610,7 +610,7 @@ export class PeerManager {
     try {
       await this.libp2p.hangUp(peer);
     } catch (e) {
-      this.logger.warn("Unclean disconnect", {peer: prettyPrintPeerId(peer)}, e as Error);
+      this.logger.debug("Unclean disconnect", {peer: prettyPrintPeerId(peer)}, e as Error);
     }
   }
 
--- a/packages/validator/src/services/attestationDuties.ts
+++ b/packages/validator/src/services/attestationDuties.ts
@@ -168,7 +168,7 @@ export class AttestationDutiesService {
     for (const epoch of [currentEpoch, nextEpoch]) {
       // Download the duties and update the duties for the current and next epoch.
       await this.pollBeaconAttestersForEpoch(epoch, indexArr).catch((e: Error) => {
-        this.logger.error("Failed to download attester duties", {epoch}, e);
+        this.logger.warn("Failed to download attester duties", {epoch}, e);
       });
     }
 
--- a/packages/validator/src/services/syncCommitteeDuties.ts
+++ b/packages/validator/src/services/syncCommitteeDuties.ts
@@ -183,7 +183,7 @@ export class SyncCommitteeDutiesService {
     for (const epoch of [currentEpoch, nextPeriodEpoch]) {
       // Download the duties and update the duties for the current and next period.
       await this.pollSyncCommitteesForEpoch(epoch, indexArr).catch((e: Error) => {
-        this.logger.error("Failed to download SyncDuties", {epoch}, e);
+        this.logger.warn("Failed to download SyncDuties", {epoch}, e);
       });
     }
 


packages/beacon-node/src/network/peers/peerManager.ts
```diff
@@ -610,7 +610,7 @@ export class PeerManager {
     try {
       await this.libp2p.hangUp(peer);
     } catch (e) {
-      this.logger.warn("Unclean disconnect", {peer: prettyPrintPeerId(peer)}, e as Error);
+      this.logger.debug("Unclean disconnect", {peer: prettyPrintPeerId(peer)}, e as Error);
```
philknows, line 613: Reference: https://github.com/ChainSafe/lodestar/issues/5359#issuecomment-1519129570 by @nflaig for moving from `warn` to `debug`.

packages/beacon-node/src/chain/blocks/index.ts
```diff
@@ -110,7 +110,7 @@ export async function processBlocks(
     if (!(err instanceof BlockError)) {
       this.logger.error("Non BlockError received", {}, err);
     } else if (!opts.disableOnBlockError) {
-      this.logger.error("Block error", {slot: err.signedBlock.message.slot}, err);
+      this.logger.warn("Block error", {slot: err.signedBlock.message.slot}, err);
```
philknows, line 113: Reference: https://github.com/ChainSafe/lodestar/issues/3062
Block errors are sometimes expected, especially for unstable/testnet networks and `error` doesn't feel like the correct verbosity as the user cannot be expected to do something about it. As mentioned in the issue, `warn` or `info` may be more appropriate. Should we also consider removing the stacktrace?

packages/api/test/unit/client/httpClient.test.ts
```diff
@@ -128,6 +128,7 @@ describe("httpClient json client", () => {
       return Promise.reject(Error("did not throw"));
     } catch (e) {
       if (!(e instanceof HttpError)) throw Error(`Not an HttpError: ${(e as Error).message}`);
+      // Discuss error handling: https://github.com/ChainSafe/lodestar/issues/5359#issuecomment-1529027981
       expect(e.message).to.equal("Service Unavailable: Node is syncing");
```
philknows, line 132: Reference: https://github.com/ChainSafe/lodestar/issues/5359#issuecomment-1529027981
As @nflaig mentioned in the comment, we should remove the stacktrace here, log as `warn` and consider logging something that is expected like "Node is currently syncing..."

packages/validator/src/services/attestationDuties.ts
```diff
@@ -168,7 +168,7 @@ export class AttestationDutiesService {
     for (const epoch of [currentEpoch, nextEpoch]) {
       // Download the duties and update the duties for the current and next epoch.
       await this.pollBeaconAttestersForEpoch(epoch, indexArr).catch((e: Error) => {
-        this.logger.error("Failed to download attester duties", {epoch}, e);
+        this.logger.warn("Failed to download attester duties", {epoch}, e);
```
philknows, line 171: Reference: https://github.com/ChainSafe/lodestar/issues/5359#issuecomment-1529027981
Not sure if this should be an `error`? We also should suppress it if it's triggered due to `Node is syncing` state.

packages/validator/src/services/syncCommitteeDuties.ts
```diff
@@ -183,7 +183,7 @@ export class SyncCommitteeDutiesService {
     for (const epoch of [currentEpoch, nextPeriodEpoch]) {
       // Download the duties and update the duties for the current and next period.
       await this.pollSyncCommitteesForEpoch(epoch, indexArr).catch((e: Error) => {
-        this.logger.error("Failed to download SyncDuties", {epoch}, e);
+        this.logger.warn("Failed to download SyncDuties", {epoch}, e);
```
philknows, line 186: Reference: https://github.com/ChainSafe/lodestar/issues/5359#issuecomment-1529027981
Not sure if this should be an error? We also should suppress it if it's triggered due to `Node is syncing` state.

packages/beacon-node/src/chain/blocks/index.ts
dapplion: This log here does not make much sense to be above debug level IF ONLY the consumers handle the error properly. Invalid blocks can be submitted via:
- gossip: a peer is free to submit and invalid block. That's an uncontrolled event the user can't do much about, so we may want to not log it at all. However, if there's a consensus bug the node will just freeze without warning
- sync: same as above
- API: this _should_ never happen as long as the API is trusted. The API also logs errors so this will probably be a duplicate

For now, let's manually review every caller of this path and check if they already log the error and at what level

packages/beacon-node/src/network/peers/peerManager.ts
dapplion: Definitely must be debug level

packages/api/test/unit/client/httpClient.test.ts
dapplion: This is a test, the actual logger statement to change should be somewhere else

packages/validator/src/services/attestationDuties.ts
dapplion: This is a broader issue around handling bad beacon node connections more gracefully. This make sense to be an error because it's critical, AND the user can do something about it (say if the beacon node is down or unsynced). However the problem is the possible spam of errors

nflaig: would agree that `error` here is the correct log level, I opened a separate PR to fix this log if node is syncing
- https://github.com/ChainSafe/lodestar/pull/5467

packages/validator/src/services/syncCommitteeDuties.ts
nflaig: same as https://github.com/ChainSafe/lodestar/pull/5457#discussion_r1186128363 above

packages/api/test/unit/client/httpClient.test.ts
nflaig: There are no more verbose "Node is syncing" error logs on the VC side, those are all now logged as `warn` (single line log)

Related PRs
- https://github.com/ChainSafe/lodestar/pull/5467
- https://github.com/ChainSafe/lodestar/pull/5511

packages/beacon-node/src/network/peers/peerManager.ts
nflaig: To be fixed by
- https://github.com/ChainSafe/lodestar/pull/5541

We might wanna include that in 1.9 as there have been concerns from several users in regard to this log

packages/beacon-node/src/chain/blocks/index.ts
nflaig: This log has been moved to debug
- https://github.com/ChainSafe/lodestar/pull/5485

wemeetagain:
Can this be closed?

philknows:
Closing. Stale and it probably makes better sense to discuss specific UX logging issues as its own issues for better tracking.

Issue #5413, opened 2023-04-25
Optimize block time to head


dapplion:
In unstable metrics show that the time to process a block = seen in gossip until set as head ranges 600-900ms. This includes calling the execution client to notify the payload. Running the state transition function for a block takes ~50ms and computing the head another ~50ms. So the long process times are due to the shared resources with other items like processing attestations.

**Block process timeline**

| Step | Avg time |
| -- | -- | 
| Gossip validation | 30 ms |
| Send notify payload | 10 ms |
| State transition | 45 ms |
| State hashing | 25 ms |
| Verify block sigs | 25 ms |
| notify new payload idle | 300 ms |
| import attestations | 10 ms |
| recompute head | 100 ms |
| persist LC data | 25 ms |
| **total** | 570 ms |

CPU profiles show that attestation processing interlaces with block processing, It's unclear now if the 300 ms time for the new payload response is caused by that or not. See two examples of idle and busy loop waiting for notify new payload response

![image](https://user-images.githubusercontent.com/35266934/234184688-9c8320fb-8033-4608-aaa6-32e83fe32ee3.png)

![image](https://user-images.githubusercontent.com/35266934/234186189-e3591930-24b7-46d1-8c3d-5b8474a51be4.png)

**Research / gather more data**

- What's the actual notify new payload time from the point of view of the execution client
  - On unstable mainnet, average time is 100ms
```
rpc_duration_engine_newPayloadV2_success {quantile="0.5"} 104116.5
rpc_duration_engine_newPayloadV2_success {quantile="0.75"} 132396.75
rpc_duration_engine_newPayloadV2_success {quantile="0.95"} 225751.89999999988
rpc_duration_engine_newPayloadV2_success {quantile="0.99"} 364260.3900000002
rpc_duration_engine_newPayloadV2_success {quantile="0.999"} 1.0846435890000006e+06
rpc_duration_engine_newPayloadV2_success {quantile="0.9999"} 1.090077e+06
```

- If significantly less than 300ms, add a snooper to watch traffic and understand where is the delay
- At what point into the slot is the new payload HTTP request actually dispatched from the OS?

**Optimization targets**

- [x] Algorithmic optimization of recompute head
- [ ] ~~Stack recompute head with new payload waiting~~ - investigated, found that it doesn't improve timing
- [ ] ~~Complete block import before new payload response setting it as optimistic head and set as VALID on response latter~~ - likely to cause missed attestations
[...]

PR #5458, opened 2023-05-03
feat: block network processor when processing current slot block

### LINKED ISSUE #5441: Performance issue due to unlimited maxGossipTopicConcurrency
**Describe the bug**

If we don't limit `maxGossipTopicConcurrency` (by default), it takes more time for a gossip to become head. If we limit it to 512, on test mainnet node it drops 35% of attestations, however the time for a block till becomes head is quite similar to stable

Screenshot 2023-05-01 at 13 02 39

**Expected behavior**

Unlimited `maxGossipTopicConcurrency` or configure it to a value that:
- No dropped attestations
- The time for a block to become head is the same to above

twoeths:
**Motivation**

- We want to process current slot block asap
- Right now we do that by limiting maxGossipTopicConcurrency to 512 which is not really necessary because the "hot" time is only ~600ms to process block while we applied for the whole slot

**Description**

- When processing current slot block, block network processor. Most of the gossip messages at this time are attestations with unknown block root so should not be an issue
- Remove the default value of `maxGossipTopicConcurrency`
- Add "reason" to "lodestar_network_processor_can_not_accept_work_total" metric

part of #5413
Closes #5441

**Testing**
- [x] Initial test on test mainnet passed
- [x] Test for 1 more day with latest version

### DIFF
--- a/packages/beacon-node/src/chain/chain.ts
+++ b/packages/beacon-node/src/chain/chain.ts
@@ -133,6 +133,7 @@ export class BeaconChain implements IBeaconChain {
   private readonly exchangeTransitionConfigurationEverySlots: number;
 
   private processShutdownCallback: ProcessShutdownCallback;
+  private _isProcessingCurrentSlotBlock = false;
 
   constructor(
     opts: IChainOptions,
@@ -282,6 +283,10 @@ export class BeaconChain implements IBeaconChain {
     this.stateCache.clear();
     this.checkpointStateCache.clear();
     await this.bls.close();
+  }
+
+  isProcessingCurrentSlotBlock(): boolean {
+    return this._isProcessingCurrentSlotBlock;
   }
 
   regenCanAcceptWork(): boolean {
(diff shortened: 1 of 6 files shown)


twoeths:
metrics look good compared to beta:
- feat2: shorter block processs time

Screenshot 2023-05-04 at 08 47 06

- vs beta

Screenshot 2023-05-04 at 08 46 05

- Blocked messages per slot due to processing current block slot
Screenshot 2023-05-04 at 08 48 06

- Almost no dropped attestations (vs 30% - 40 in beta)
Screenshot 2023-05-04 at 08 49 02

packages/beacon-node/src/chain/chain.ts
```diff
@@ -284,6 +285,10 @@ export class BeaconChain implements IBeaconChain {
     await this.bls.close();
   }
 
+  isProcessingCurrentSlotBlock(): boolean {
+    return this._isProcessingCurrentSlotBlock;
+  }
```
dapplion: Should this logic and `_isProcessingCurrentSlotBlock` be in the network processor class? It knows the block's slot and current clock

twoeths: yes I can move the logic there. It means we block network processor messages right after we receive a gossip block message instead of at the time we process block

wemeetagain (approved):
lgtm

wemeetagain:

PR #5461, opened 2023-05-04
fix(slashing-protection): resolve `minEpoch` max call stack issue

### LINKED ISSUE #5392: Cannot import big slashing protection
`Internal server error. Error: Maximum call stack size exceeded`

Running with `--keymanager.bodyLimit 11125829120`

`28M` file

nflaig:
**Motivation**

Closes https://github.com/ChainSafe/lodestar/issues/5392

**Description**

Epochs array might have more than ~10^5 elements, cannot use `Math.min` as it would throw `RangeError: Maximum call stack size exceeded` if `epochs.length >= 125626`.

This change makes the solution to determine min epoch more robust and adds few unit tests.

### DIFF
--- a/packages/validator/src/slashingProtection/utils.ts
+++ b/packages/validator/src/slashingProtection/utils.ts
@@ -28,7 +28,7 @@ export function numToString(num: number): string {
 }
 
 export function minEpoch(epochs: Epoch[]): Epoch | null {
-  return epochs.length > 0 ? Math.min(...epochs) : null;
+  return epochs.length > 0 ? epochs.reduce((minEpoch, epoch) => (minEpoch < epoch ? minEpoch : epoch)) : null;
 }
 
 export function uniqueVectorArr(buffers: Uint8Array[]): Uint8Array[] {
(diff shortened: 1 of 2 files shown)


packages/validator/src/slashingProtection/utils.ts
```diff
@@ -28,7 +28,7 @@ export function numToString(num: number): string {
 }
 
 export function minEpoch(epochs: Epoch[]): Epoch | null {
-  return epochs.length > 0 ? Math.min(...epochs) : null;
+  return epochs.length > 0 ? epochs.reduce((minEpoch, epoch) => (minEpoch < epoch ? minEpoch : epoch)) : null;
```
nflaig, line 31: alternative implementation

```ts
export function minEpoch(epochs: Epoch[]): Epoch | null {
  if (epochs.length === 0) {
    return null;
  }

  let minEpoch = epochs[0];
  for (let i = 1; i < epochs.length; i++) {
    if (epochs[i] < minEpoch) {
      minEpoch = epochs[i];
    }
  }

  return minEpoch;
}
```

but there is no noticeable difference in performance

dapplion (approved):

wemeetagain:

Issue #5429, opened 2023-04-28
Investigate high execution payload verification time


twoeths:
**Describe the bug**

Sometimes it takes significant time to verify execution engine payload on test mainnet nodes

<img width="1309" alt="Screenshot 2023-04-28 at 08 41 45" src="https://user-images.githubusercontent.com/10568965/235034066-4cc06ca7-ba56-4582-a671-3e51d234ad05.png">

Some logs:

```
1143794:Apr-26 23:46:31.425[chain]         verbose: Verified execution payload slot=6310730, recvToVerifiedExecPayload=6.816999912261963
Apr-26 23:48:05.877[chain]         verbose: Verified execution payload slot=6310738, recvToVerifiedExecPayload=5.330000162124634
```
it's likely lodestar is busy at that time, not the execution layer's issue

**Expected behavior**
- <0.5s consistently

twoeths:
was able to take a profile of this block

```
Apr-28 02:37:16.797[chain]         verbose: Verified execution payload slot=6318784, recvToVerifiedExecPayload=1.9059998989105225
```

full log:
```
beacon-2023-04-29.log:292138:Apr-29 05:07:50.846[chain]         verbose: Verified execution payload slot=6326737, recvToVerifiedExecPayload=1.9059998989105225
devops@feat1-mainnet-hzax41:~/beacon$ grep -e "6326737" -rn beacon-2023-04-29.log
292041:Apr-29 05:07:43.367[chain]         verbose: Running prepareForNextSlot nextEpoch=197711, prepareSlot=6326737, headSlot=6326736, headRoot=0xadaca0a5724a5ccca2d7cbd56277c2ab7a5e24d1ec510bc54a8710a6f7166902, isEpochTransition=false
292108:Apr-29 05:07:47.000[chain]         verbose: Clock slot slot=6326737
292129:Apr-29 05:07:48.961[network]       verbose: Received gossip block slot=6326737, root=0x6510…54a5, curentSlot=6326737, peerId=16Uiu2HAmVCyDPHdUYYv9mMY2Z2DDCm2WVRd1PBK5nc4i4F5y6P4K, delaySec=1.940000057220459, recvToVal=0.020999908447265625
292130:Apr-29 05:07:49.301[chain]         verbose: Transitioned gossip block slot=6326737, recvToTransition=0.36100006103515625
[...]

PR #5473, opened 2023-05-09
feat: batch io operations when verifying & importing block

### LINKED ISSUE #5415: Improve importBlock time
**Describe the bug**

ImportBlock could takes up to 162ms (this is from a test mainnet node)

Screenshot 2023-04-25 at 13 17 17

**Expected behavior**

- Add metric to track it
- `recomputeForkchoiceHead` takes 130ms, a lot of tasks inside it could be done while waiting for `engine.notifyNewPayload`
- `LightClientServer.onImportBlockHead` takes 30ms, right now most of nodes don't act as a LightClientServer by default, should we turn the flag to not calling this function by default? If not find a better time to call it, could be in the next event loop

twoeths:
**Motivation**

- We want to import block asap
- Persisting block may take up to 1s - 2s, in the worse scenario it could be up to 3s due to the I/O lag (see #5429)

Screenshot 2023-05-09 at 14 43 58

**Description**

- We already have a `Promise.all()` in `verifyBlock()`, take that chance to persist blocks (and blobs) to db
- If there are errors, check if blocks are not in forkchoice then remove them from db
- Also process import block to LightClientServer in the next event loop to improve block time to head

Closes #5415

### DIFF
--- a/packages/beacon-node/src/chain/blocks/index.ts
+++ b/packages/beacon-node/src/chain/blocks/index.ts
@@ -8,7 +8,7 @@ import type {BeaconChain} from "../chain.js";
 import {verifyBlocksInEpoch} from "./verifyBlock.js";
 import {importBlock} from "./importBlock.js";
 import {assertLinearChainSegment} from "./utils/chainSegment.js";
-import {BlockInput, FullyVerifiedBlock, ImportBlockOpts} from "./types.js";
+import {BlockInput, BlockInputType, FullyVerifiedBlock, ImportBlockOpts} from "./types.js";
 import {verifyBlocksSanityChecks} from "./verifyBlocksSanityChecks.js";
 export {ImportBlockOpts, AttestationImportOpt} from "./types.js";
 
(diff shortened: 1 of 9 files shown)


packages/beacon-node/src/chain/blocks/index.ts
```diff
@@ -141,6 +141,23 @@ export async function processBlocks(
...
+        removeBlockPromises.push(this.db.block.remove(block));
+        if (type === BlockInputType.postDeneb) {
+          const {blobs} = blockInput;
+          removeBlockPromises.push(this.db.blobsSidecar.remove(blobs));
+        }
+      }
+    }
+    Promise.all(removeBlockPromises).catch((e) => {
+      this.logger.verbose("Failed to remove blocks", {slots: blocks.map((b) => b.block.message.slot).join(",")}, e);
+    });
```
dapplion: I would not bundle in a single Promise.all since then separate errors are reported once, just attach a .catch (two total) to each call above

packages/beacon-node/src/chain/blocks/verifyBlock.ts
```diff
@@ -101,6 +136,7 @@ export async function verifyBlocksInEpoch(
 
       // All signatures at once
       verifyBlocksSignatures(this.bls, this.logger, this.metrics, preState0, blocks, opts),
+      ...persistBlockPromises,
```
dapplion: Move the new code to a function called `writeBlockInputToDb()` and follow similar pattern to the functions above. Do the Promise.all inside that function. Place the coment explaining why we do this eagerly above line 139, with a space between line 138 and the start of the comment

dapplion (commented):
Pushed some changes, see

https://github.com/ChainSafe/lodestar/blob/4ce29676d56654b1d342fc006cd2d73f78f2a781/packages/beacon-node/src/chain/blocks/index.ts#L145-L159

@tuyennhv we have to be careful here since this may open up attacks. We are exposing DB growth to potentially untrusted data. Blocks from gossip have proposer sig validated but blocks from sync have no validation at all. We may want to disable this feature for sync and only do for payloads with some validation like gossip or unknown block sync

twoeths:
@dapplion I added `eagerPersistBlock` flag to `ImportBlockOpts` so that only blocks coming from gossip or unknown block sync gets this feature

g11tech:
> We

agreed, anyway eager saving on sync makes no effective impact because of checkpoint sync

dapplion (approved):
Looks great! Both gossip and unknown block root are safe for `eagerPersistBlock`. Can you open an issue to add a test that asserts that unknown block sync won't send to the chain blocks with invalid root?

twoeths:
@dapplion I added that test case to this commit https://github.com/ChainSafe/lodestar/pull/5485/commits/c425a246c85e2ce6639e8814aeea7ec8cd321160

see #5485

wemeetagain:

PR #5474, opened 2023-05-09
fix(validator): lock keystores when loading from cache file

### LINKED ISSUE #5374: Lockfiles are not created/checked when loading from keystore cache
When loading from keystore cache file we are currently not checking if a `.lock` for any of the keystores exists and the cache file itself does not have a lock file.

It is not possible to load the cache file is stored in `dataDir` and if you start a 2nd validator with same `dataDir` you would get `✖ Error: Database is not open` but if a VC has a cache file it will not create lock files which is dangerous.

_Originally posted by @nflaig in https://github.com/ChainSafe/lodestar/pull/5357#discussion_r1167813508_

nflaig:
**Motivation**

Closes https://github.com/ChainSafe/lodestar/issues/5374

**Description**

- Lock keystores when loading from cache file (this will also check if lockfiles already exist)
- Update tests to run with and without existing keystore cache


dapplion (approved):
LGTM!

wemeetagain:

PR #5466, opened 2023-05-05
chore: update year on the chainsafe cli notice

nflaig:
**Motivation**

It's 2023 already

**Description**

Update year on the chainsafe cli notice

PR #5476, opened 2023-05-10
chore: dynamic year on the chainsafe cli notice

nflaig:
**Motivation**

As discussed in https://github.com/ChainSafe/lodestar/pull/5466 we don't really know why the year is even there but it "looks correct" so let's keep it and make sure it always up to date.

**Description**

Makes year on the chainsafe cli notice dynamic.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -9,7 +9,7 @@ import {getVersionData} from "./util/version.js";
 const {version} = getVersionData();
 const topBanner = `🌟 Lodestar: TypeScript Implementation of the Ethereum Consensus Beacon Chain.
   * Version: ${version}
-  * by ChainSafe Systems, 2018-2023`;
+  * by ChainSafe Systems, 2018-${new Date().getFullYear()}`;
 const bottomBanner = `📖 For more information, check the CLI reference:
   * https://chainsafe.github.io/lodestar/reference/cli
 
--- a/packages/flare/src/cli.ts
+++ b/packages/flare/src/cli.ts
@@ -10,7 +10,7 @@ Flare is a sudden brief burst of bright flame or light.
 In the wrong hands, can lead people astray.
 Use with care.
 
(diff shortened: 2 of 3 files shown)


dapplion:
Q: Is it really necessary to print the year?
ChatGTP: --

While it has been a common practice to include the year in copyright notices, it's not universally required, especially in many jurisdictions, including the United States. This tradition started because under the old U.S. copyright law (before 1978), the copyright would only last for 28 years from the date of publication, but could be renewed for another 28 years if the copyright was properly noticed, and the renewal registered. This is why you often saw, and still see, copyright notices with a year in them.

However, under the Berne Convention for the Protection of Literary and Artistic Works, which the U.S. joined in 1989, copyright is automatic and does not require registration or a copyright notice. Many other countries are also signatories to this convention.

So, in many cases, including the year in a copyright notice is not strictly necessary. However, it's still a common practice because it can provide certain benefits:

1. If you are not in a Berne Convention country, or are dealing with someone who is not, the notice may be required or beneficial.
2. It makes it clear to everyone when the work was produced, which can be useful information.
3. It may help in a legal dispute where the date of creation is relevant.

In the case of software projects, it's also common to include the range of years during which the project has been updated. This can help indicate the lifespan and active maintenance period of the project. However, it's worth noting that every jurisdiction may have slightly different rules about what is required or beneficial in a copyright notice.

It's always a good idea to consult with a legal expert if you're unsure about the specific requirements or benefits for your situation.

----

I would vote to remove the year

nflaig:
>  it's also common to include the range of years during which the project has been updated. This can help indicate the lifespan and active maintenance period of the project

I kinda like this part but can also just remove it, no strong opinion either way as long as it does not display outdated information

wemeetagain:
I like it too, no real reason but for the vibe. Feels like it gives a sense of the progression of the project, gives confidence in the project

dapplion (approved):

wemeetagain:

PR #5181, opened 2023-02-20
feat: free the blobs

g11tech:
Free the blobs

the branch is now functional! (cleanup and some blob gossip validations still wip)

Below are the logs of a local devnet (with a corresponding ethereumjs branch) where peer1 is producing blobs
And peer2 syncs block and blobs
1. Initially via req/resp to get `Synced`
2. Then via gossip to stay `Synced`
image
3. Unknown Parent root sync using beacon/blob side cars by root
image

ref:
- https://github.com/ethereum/consensus-specs/pull/3244

TODOS: 
 - [x] modified block/blob production
 - [x] modified blob dbs
 - [x] handled block/blob gossip 
 - [x] marry blob and block gossip for BlockInput
 - [x] updated references    
 - [x] fixed gossip types
 - [x] updated fetch blocksMaybeBlobsbyRange
 - [x] updated fetch blocksMaybeBlobsbyRoot
 - [x] handle block/blob archival to finalizeddbs    
 - [x] blob publish
 - [x] fix block publish
 - [x] handle onBlobSideCarByRange
 - [x] handle onBlockSideCarbyRoot
 - [x] prune gossip blockinput cache
 - [x] handle genesis state blobs db handling
 - [x] validator blob signing and publishing
   - [x] modify get produced blobs
   - [x] sign and propagate
 - [x] integrate c-kzg with computeBlobKzgProof
   - [x]  update interfaces

TODO:
- [x] single node run with ethereumjs courtsey: - https://github.com/ethereumjs/ethereumjs-monorepo/pull/2567
- [x] Multi peer run to test gosssip
- [x] Multi peer run to sync to test req/resp
  - [x] normal sync via blobssidecar by range
  - [x] unknown block sync via by root
- [x]  Fix beaconBlockMayBeBlobsByRoot request: review and make it multi block

PR #5484, opened 2023-05-12
fix: modify minimal preset's field elements per blob to mainnet for dev runs

g11tech:
As part of by parts integration of free the blobs PR
 - https://github.com/ChainSafe/lodestar/pull/5181

For devnets started using cli in minimal and for the merge interop/inline tests, the `FIELD_ELEMENTS_PER_BLOB` needs to be set to `4096` which is the mainnet value

because all kzg libraries and ELs process the blobs on this size and is a hardcoded value in them. We only need minimal value for spec tests which does not use the cli path or will not be providing DEV_RUN flag on test setup

### DIFF
--- a/packages/cli/src/applyPreset.ts
+++ b/packages/cli/src/applyPreset.ts
@@ -14,7 +14,7 @@
 // set LODESTAR_PRESET manually every time.
 
 // IMPORTANT: only import Lodestar code here which does not import any other Lodestar libraries
-import {setActivePreset, presetFromJson} from "@lodestar/params/setPreset";
+import {setActivePreset, presetFromJson, PresetName} from "@lodestar/params/setPreset";
 import {readFile} from "./util/file.js";
 
 const network = valueOfArg("network");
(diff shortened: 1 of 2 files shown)


packages/cli/src/applyPreset.ts
```diff
@@ -44,6 +47,9 @@ else if (network) {
 else if (process.argv[2] === "dev") {
   process.env.LODESTAR_PRESET = "minimal";
   process.env.LODESTAR_NETWORK = "dev";
+  // "c-kzg" has hardcoded the mainnet value, do not use presets
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  setActivePreset(PresetName.minimal, {FIELD_ELEMENTS_PER_BLOB: 4096});
```
dapplion, line 52: Adding this in the production code path just for a local test of us is pretty ugly. Please use presetFile functionality in that test to override the values on the test runtime only.

dapplion (changes requested):

packages/cli/src/applyPreset.ts
g11tech: this for running lodestar via `--dev` network flag on cli (mostly via prebuild/docker images), test override is a different file

dapplion: Could you detail more who / where this is necessary? Is someone else dependant on this fix?

g11tech: generally its used in ethereumjs sims for 4844 (https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/test/sim/4844.md) , also we would want our `--dev` network to be 4844 compatible with EL clients

dapplion (approved):

wemeetagain:

PR #5488, opened 2023-05-12
ci: add support for the CI env for e2e tests

nazarhussain:
**Motivation**

Reduce dependencies of external resources during the e2e tests. 

**Description**

- Removed dependency of e2e tests to external sepolia nodes
- Created a script to spinup the environment for the e2e tests
- Take backup of the logs just in case need to debug

Run e2e tests.


.github/workflows/test-e2e.yml
```diff
@@ -51,7 +52,22 @@ jobs:
...
+      - name: Wait for the node to be ready
+        run: npx wait-port -t 60000 0.0.0.0:5001
+
       - name: E2E tests
         run: yarn test:e2e
         env:
           GOERLI_RPC_URL: ${{ secrets.GOERLI_RPC_URL!=0 && secrets.GOERLI_RPC_URL || env.GOERLI_RPC_DEFAULT_URL }}
+
+      - name: Stop the e2e test environment
+        run: kill -9 $(lsof -t -i:5001)
```
dapplion: Could you move this logic onto a bash script with instructions so it can be recreated locally? Counter argument is that breaking the steps in CI steps gives better debuggability on what broke, but let me know your thoughts

nazarhussain: Recreating locally have more benefits for everyone. Will update the scripts.

packages/cli/test/scripts/e2e_test_env.ts
```diff
@@ -0,0 +1,44 @@
...
+    },
+  },
+  [
+    {id: "node-1", cl: CLClient.Lodestar, el: ELClient.Geth, keysCount: 32, mining: true},
+    {id: "node-2", cl: CLClient.Lodestar, el: ELClient.Nethermind, keysCount: 32},
+  ]
+);
+
+await env.start({runTimeoutMs: 0});
+await connectAllNodes(env.nodes);
```
dapplion, line 44: Does this approach leave a network alive for the entire duration of the tests? Which is going through forks? If yes, this means test outcomes of different e2e tests influence each other since they share global state, is that correct?

nazarhussain: Yes the network will be alive through out execution of **all e2e tests**. It's like connecting to external node, but with this the node is configured and running locally.

Every e2e tests should scaffold their specific network condition on this live network. e.g. If one test need to a contract, can deploy it. Or one network need a specific account, can transfer funds from genesis account. This way execution of tests would not effect each other.

wemeetagain (approved):
Approving/merging to unblock other PRs but lets make sure the dicussion here doesn't get lost

wemeetagain:

PR #5490, opened 2023-05-13
feat!: worker friendly logger instances

dapplion:
**Motivation**

Our current Logger instances are not DX friendly to support workers.

Let's list the requirements:
1. Downstream dependencies like `@lodestar/light-client` or `@lodestar/reqresp` just want access to a generic simple Logger instance to write data at some LogLevel. No need to expose them to more complicated loggers or to force dependencies on NodeJS specific code
2. Complex intermediate depencies like `@lodestar/beacon-node` need access to a generic logger with more features, specifically:
  2.1. Create child logger with custom settings, which **must** re-use references to the same transport instances
  2.2. Serialize a logger instance settings and transport settings for a worker downstream
3. A worker must be able to instantiate a logger with the same settings as its main thread handler, with new transport instances but that have the same I/O target (stdout, file)
4. Top level dependency `@lodestar/cli` has to be able to produce compatible logger instances for the above cases
5. plus misc use cases for testing (empty logger, test logger)

**Description**

So to fulfill 1. and 2. there must exist separate insterfaces, which are:
- `Logger`
- `LoggerNode`

`LoggerNode` can export its settings with `toOpts()` and be recreate with those


packages/logger/package.json
```diff
@@ -0,0 +1,78 @@
...
+    "./browser": {
+      "import": "./lib/loggers/browser.js"
+    },
+    "./node": {
+      "import": "./lib/loggers/node.js"
+    },
+    "./empty": {
+      "import": "./lib/loggers/empty.js"
+    }
+  },
```
nazarhussain: Instead of subpath exports why not not use conditional exports https://nodejs.org/api/packages.html#conditional-exports. And only default export. This way user can use a consistent import pattern. 

```json
"exports": {
    ".": {
      "node": "./lib/loggers/node.js",
      "browser": "./lib/loggers/browser.js",
      "default": "./lib/index.js"
    },
  },
```

wemeetagain (approved):

packages/logger/src/node.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  static fromNewTransports(opts: LoggerNodeOpts): WinstonLoggerNode {
+    return new WinstonLoggerNode(opts, getNodeLoggerTransports(opts));
+  }
+
+  // Return a new logger instance with different module and log level
+  // but a reference to the same transports, such that there's only one
+  // transport instance per tree of child loggers
+  child(opts: LoggerNodeChildOpts): LoggerNode {
```
nflaig, line 146: @dapplion This changed how we create child loggers, it creates a new event listener for each child logger
- https://github.com/ChainSafe/lodestar/issues/5529

Was the previous implementation not compatible when using workers?
https://github.com/ChainSafe/lodestar/blob/375d660ed1e7ba13945dff41e63154a26c4be008/packages/utils/src/logger/winston.ts#L82

Using native `winston.child` does not work as mention in previous comment.

Workaround for now is to just increase max listeners which should be fine to do as explained in the PR
- https://github.com/ChainSafe/lodestar/pull/5537

packages/logger/src/node.ts
```diff
@@ -0,0 +1,159 @@
...
+  //   logFileDailyRotate: undefined,
+  // }
+  // `lodestar --logFileDailyRotate` -> enabled daily rotate with default value
+  // `lodestar --logFileDailyRotate 10` -> set daily rotate to custom value 10
+  // `lodestar --logFileDailyRotate 0` -> disable daily rotate and accumulate in same file
+  if (opts.file) {
+    const filename = opts.file.filepath;
+
+    transports.push(
+      opts.file.dailyRotate != null
```
nflaig, line 98: What's the expected approach to disable log file rotation now?

The comment above does not longer match behavior as setting `--logFileDailyRotate 0` will still initialize `DailyRotateFile`. Default `winston.transports.File` is effectively never used now.

packages/logger/src/node.ts
dapplion: Hmm unsure of the reason forgot to document, so may be an oversight. Could you test if the previous approach works fine still?

packages/logger/src/node.ts
dapplion: Right, logFileDailyRotate is a bit a weird UX flag. Does yargs support accepting both syntax with no arg `--logFileDailyRotate` and with arg `--logFileDailyRotate 5`? That would make handling easier

packages/logger/src/node.ts
nflaig: Child loggers are only created in the main thread, does not affect workers at all because we initialize a completely new logger instance there https://github.com/ChainSafe/lodestar/blob/e2e5417f48cac848cc0eb0b34389c9a6b7673444/packages/beacon-node/src/network/core/networkCoreWorker.ts#L38

Opened a PR to revert `logger.child` implementation back to original approach
- https://github.com/ChainSafe/lodestar/pull/5597

packages/logger/src/node.ts
nflaig: Both works, but setting `--logFileDailyRotate` (without value) and omitting the flag will result in `5` due to the fact that we set it as default here https://github.com/ChainSafe/lodestar/blob/e2e5417f48cac848cc0eb0b34389c9a6b7673444/packages/cli/src/options/logOptions.ts#L40

Setting it to `0` to disable daily file rotation seems logical to me (and is most explicit), that's also how it worked previously
- https://github.com/ChainSafe/lodestar/pull/5598

wemeetagain:

PR #5492, opened 2023-05-14
feat: decouple the deneb blob and block production

g11tech:
As part of by parts integration of free the blobs PR
 - https://github.com/ChainSafe/lodestar/pull/5181

Update the blob and block production and signing flow as per the decoupled blob spec

implements:
 - https://github.com/ethereum/beacon-APIs/pull/302


packages/beacon-node/src/api/impl/beacon/blocks/index.ts
```diff
@@ -205,56 +233,26 @@ export function getBeaconBlockApi({
...
-              chain.getBlobsSidecar(signedBlock.message as deneb.BeaconBlock)
-            )
-          : getBlockInput.preDeneb(config, signedBlock, BlockSource.api);
-
-      await promiseAllMaybeAsync([
+      metrics?.registerBeaconBlock(OpSource.api, seenTimestampSec, blockForImport.block.message);
+      const publishPromises = [
         // Send the block, regardless of whether or not it is valid. The API
         // specification is very clear that this is the desired behaviour.
-        () => network.gossip.publishBeaconBlockMaybeBlobs(blockForImport) as Promise<unknown>,
```
twoeths: maybe remove this API in upcoming PRs?

packages/beacon-node/src/chain/chain.ts
```diff
@@ -683,11 +687,21 @@ export class BeaconChain implements IBeaconChain {
...
 
-    // Prune old blobsSidecar for block production, those are only useful on their slot
-    if (this.config.getForkSeq(slot) >= ForkSeq.deneb && this.producedBlobsSidecarCache.size > 0) {
-      for (const [key, blobsSidecar] of this.producedBlobsSidecarCache) {
-        if (slot > blobsSidecar.beaconBlockSlot + MAX_RETAINED_SLOTS_CACHED_BLOBS_SIDECAR) {
-          this.producedBlobsSidecarCache.delete(key);
+    // Prune old blobSidecars for block production, those are only useful on their slot
+    if (this.config.getForkSeq(slot) >= ForkSeq.deneb) {
+      if (this.producedBlobSidecarsCache.size > 0) {
+        for (const [key, {slot: blobSlot}] of this.producedBlobSidecarsCache) {
```
twoeths, line 696: not related to this PR, why don't we have `producedBlobSidecarsCache` with slot or block number as key so that we don't have to do a full scan here? I think we should have slot/block number in all of the consumer of the cache

packages/beacon-node/src/chain/chain.ts
```diff
@@ -134,7 +134,8 @@ export class BeaconChain implements IBeaconChain {
   readonly checkpointBalancesCache: CheckpointBalancesCache;
   // TODO DENEB: Prune data structure every time period, for both old entries
   /** Map keyed by executionPayload.blockHash of the block for those blobs */
-  readonly producedBlobsSidecarCache = new Map<RootHex, deneb.BlobsSidecar>();
+  readonly producedBlobSidecarsCache = new Map<RootHex, {blobSidecars: deneb.BlobSidecars; slot: Slot}>();
```
twoeths: maybe define a new type of `BlockHash` as we use execution block hash here, having `RootHex` is a bit confuse even through Typescript compiles ok

packages/validator/src/services/block.ts
```diff
@@ -270,35 +311,93 @@ export class BlockProposingService {
...
+      }
+
+      case ForkName.deneb:
       default: {
         const res = await this.api.validator.produceBlockV2(slot, randaoReveal, graffiti);
         ApiError.assert(res, "Failed to produce block: validator.produceBlockV2");
-        return res.response;
+
+        const {response} = res;
+        if ((response.data as allForks.BlockContents).block === undefined) {
```
twoeths: should we also assert `blobSidecars !== undefined` here?

packages/validator/src/services/block.ts
```diff
@@ -270,35 +311,93 @@ export class BlockProposingService {
...
+        }
+        const {data: block, blockValue} = response as {data: allForks.BlindedBeaconBlock; blockValue: Wei};
+        return {block, blockValue};
+      }
+
+      case ForkName.deneb:
+      default: {
+        if ((response as {data: allForks.BlindedBlockContents}).data.blindedBlock === undefined) {
+          throw Error(`Expected BlockContents response at fork=${fork}`);
+        }
```
twoeths, line 403: we may need to also assert `blindedBlobSidecars !== undefined` here

packages/validator/src/util/externalSignerClient.ts
```diff
@@ -229,5 +232,9 @@ function serializerSignableMessagePayload(config: BeaconConfig, payload: Signabl
 
     case SignableMessageType.BLS_TO_EXECUTION_CHANGE:
       return {BLS_TO_EXECUTION_CHANGE: ssz.capella.BLSToExecutionChange.toJson(payload.data)};
+
+    case SignableMessageType.BLOB:
+      // TODO: freetheblobs
```
twoeths: suggest commenting this with `TODO DENEB` prefix so that we don't miss any of these TODOs by just doing a quick search

packages/beacon-node/src/chain/chain.ts
g11tech: because its not re-org safe

packages/validator/src/services/block.ts
g11tech: well the api will actually fail in deserializing json response from the beacon block if not all fields are in order, this extra complication is because there are now two different type structures being mixed here, but if for code clarity you feel adding would be a good idea, i can add additional asset :+1:

packages/beacon-node/src/api/impl/beacon/blocks/index.ts
g11tech: hmmm makes sense :+1:

packages/types/src/deneb/sszTypes.ts
```diff
@@ -154,6 +156,57 @@ export const BlobSidecar = new ContainerType(
...
+  {typeName: "BlockContents", jsonCase: "eth2"}
+);
+
+export const SignedBlockContents = new ContainerType(
+  {
+    signedBlock: SignedBeaconBlock,
+    signedBlobSidecars: SignedBlobSidecars,
+  },
+  {typeName: "SignedBlockContents", jsonCase: "eth2"}
+);
```
dapplion: Is this a spec'ed type? Types in `packages/types` should exclusively include spec'ed types, which need merkleization and / or serialization. Else define in beacon-node package

dapplion (changes requested):

packages/types/src/deneb/sszTypes.ts
g11tech: i think its just beacon-apis serialization type, might be useful in eventually supporting ssz encoded/decoded responses in api, but for now will move into beacon-api package itself and import/use from there :+1:

packages/beacon-node/src/chain/chain.ts
twoeths: I mean we cache by slot, and then by block hash, and when we search we filter by slot first, then by block hash (I notice all of the consumers have slot).
it'd be easier to prune by slot

g11tech: hmmm can be done but sounds like an overkill since the freshest blobs are the most likely to be used in the proposal and are added to cache by increasing slot order, so pruning out by max naturally follows that order

packages/types/src/deneb/sszTypes.ts
g11tech: @dapplion  have done some refactoring on this, let me know if you find it suitable now

dapplion (approved):
Looks great!

wemeetagain:

PR #5500, opened 2023-05-17
fix: use proper state to verify attestations

### LINKED ISSUE #4831: Wrong state in attestation validation
**Describe the bug**

`packages/beacon-node/src/chain/validation/attestation.ts:81` retrieves the wrong state.
- Must be a state in the same chain as attHeadBlock, but dialed to target.epoch

It causes signature validation to fail if the first slot of a fork is missed. The domain of the state will be wrong, since it will pick the domain of the previous fork, due to updated getDomain logic

### LINKED ISSUE #5498: Use proper state to verify attestations
- Now that we have https://github.com/ChainSafe/lodestar/pull/5390, we may be able to re-enable https://github.com/ChainSafe/lodestar/pull/4849
- use `maxSkipSlots` as the parameter to avoid regen rather than just using a state at the wrong epoch (expose via CLI too)
- related https://github.com/ChainSafe/lodestar/issues/4831

twoeths:
**Motivation**

Right now we always use head state to verify attestations

**Description**

We should only use head block state if:
- head block slot and attestation slot are in the same fork
- and head block state includes committees of target epoch

otherwise we should use head state dialed to target epoch. Notice that with `maxSkippedSlot=32` we'll always use head state unless it's at epoch boundary

Also add some metrics to monitor slot distances in attestation

Closes #5498 Closes #4831

**TODOs**
- [x] Unit test
- [x] Deploy and monitor on a test mainnet node

### DIFF
--- a/packages/beacon-node/src/chain/validation/aggregateAndProof.ts
+++ b/packages/beacon-node/src/chain/validation/aggregateAndProof.ts
@@ -12,7 +12,12 @@ import {AttestationError, AttestationErrorCode, GossipAction} from "../errors/in
 import {RegenCaller} from "../regen/index.js";
 import {getAttDataBase64FromSignedAggregateAndProofSerialized} from "../../util/sszBytes.js";
 import {getSelectionProofSignatureSet, getAggregateAndProofSignatureSet} from "./signatureSets/index.js";
-import {getCommitteeIndices, verifyHeadBlockAndTargetRoot, verifyPropagationSlotRange} from "./attestation.js";
+import {
+  getCommitteeIndices,
+  getStateForAttestationVerification,
+  verifyHeadBlockAndTargetRoot,
+  verifyPropagationSlotRange,
+} from "./attestation.js";
 
 export type AggregateAndProofValidationResult = {
   indexedAttestation: phase0.IndexedAttestation;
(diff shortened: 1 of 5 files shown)


twoeths:
Tested for some days on a mainnet node `lodestar_gossip_attestation_use_head_block_state_dialed_to_target_epoch_count` shows no value. Ready for review.

packages/beacon-node/src/chain/validation/aggregateAndProof.ts
```diff
@@ -106,14 +117,13 @@ export async function validateGossipAggregateAndProof(
   // Using the target checkpoint state here caused unstable memory issue
   // See https://github.com/ChainSafe/lodestar/issues/4896
   // TODO: https://github.com/ChainSafe/lodestar/issues/4900
```
wemeetagain:
```suggestion
```

packages/beacon-node/src/chain/validation/attestation.ts
```diff
@@ -163,14 +175,13 @@ export async function validateGossipAttestation(
     // Using the target checkpoint state here caused unstable memory issue
     // See https://github.com/ChainSafe/lodestar/issues/4896
     // TODO: https://github.com/ChainSafe/lodestar/issues/4900
```
wemeetagain:
```suggestion
```

wemeetagain (commented):
remove stale comments

wemeetagain (approved):
lgtm

wemeetagain:

PR #5501, opened 2023-05-17
test: update the sim tests to simplify assertion logic

nazarhussain:
**Motivation**

Make the sim tests more readable and easy to maintain. 

**Description**

- Updated the assertion logic to make it flexible and more readable. 
- Updated the matcher to be applied for both capture and assert step using bit flag
- Move the capture step based on the matcher not every slot
- Move the assert step to be applied once every node, not on all nodes at once. 
- Fixed the head assertion failure 
- Updated the error logging to make it more consistent among assertions  

Run all tests


.github/workflows/test-sim.yml
```diff
@@ -6,6 +6,11 @@ on:
     branches: [unstable, stable]
   pull_request:
   workflow_dispatch:
+    inputs:
+      debug:
```
matthewkeil, line 10: What `type` of input will this render as?  Does it default to `string` with the `default: ''`?

packages/cli/test/utils.ts
```diff
@@ -38,6 +38,6 @@ export function getTestdirPath(filepath: string): string {
   return fullpath;
 }
 
-export function isTruthy<T = unknown>(value: T | undefined | null): value is T {
-  return value !== undefined && value !== null;
+export function isNullish<T = unknown>(value: T | undefined | null): value is undefined | null {
+  return value === undefined || value === null;
```
matthewkeil, line 42: This is a nice change.  Is more readable 🦅 👁️

matthewkeil (approved):
Nice job.  I reviewed for code quality.  Only had the one question about workflow input.  Approving as is

.github/workflows/test-sim.yml
nazarhussain: If you invoke the workflow manually you can set any value to it e.g. `lodestar:sim:*` will show all sim tests related logs. For regular run which is invoked with push or pull request, it's value will not be defined.

matthewkeil: sweet.  it defaults to a `type: string` control

Screen Shot 2023-05-22 at 11 41 46 AM

dapplion:
Looks good!

wemeetagain:

packages/cli/test/utils/simulation/SimulationTracker.ts
```diff
@@ -279,52 +270,115 @@ export class SimulationTracker {
...
-      }
     }
 
-    this.reporter.progress(slot);
-    this.processRemoveAssertionQueue();
-  }
-
-  private processRemoveAssertionQueue(): void {
-    for (const id of this.removeAssertionQueue) {
+    for (const id of removeAssertions) {
```
nflaig, line 375: @nazarhussain Noticed that removeAssertions is always empty here as there are no items pushed to array after initializing it

https://github.com/ChainSafe/lodestar/blob/29314c9e4515662fd882b9a47b8aa0c95b0ff88c/packages/cli/test/utils/simulation/SimulationTracker.ts#L330

PR #4475, opened 2022-08-24
Review metrics

dapplion:
**Motivation**

Misc clean-up of existing metrics

**Description**

- Remove un-used metrics
- Remove index label from validator monitor
- Use histograms instead of avgMinMax
- Use proper bucket sizes
- Review metrics
- Move fork-choice metrics to beacon-node package

PR #5502, opened 2023-05-17
chore: fix dashboard panels with outdated metric queries

nflaig:
**Motivation**

Looked through metrics and fixed few metric queries along the way.

**Description**

Some metrics were changed from gauge to histogram but the queries in the dashboards were not updated
- https://github.com/ChainSafe/lodestar/pull/4475

Other updates are related to panels where we have a old and a new metric queries, just removed the old one as it should no longer be relevant at this point (updated Sep 1, 2022).


wemeetagain (approved):

wemeetagain:

Issue #5347, opened 2023-04-05
Print the set feeRecipient on startup of the VC


nflaig:
**Description**

Print the set feeRecipient on startup of the VC for each validator

**Example**

`info: Validator exists in beacon chain validatorIndex=xx, pubKey=0xxx, feeRecipient=Ox39`

**Related**

- https://discord.com/channels/593655374469660673/593655641445367808/1093177663511863356

nflaig:
Implementation started in https://github.com/ChainSafe/lodestar/pull/5239

Issue #5192, opened 2023-02-22
Log Validator life cycle changes


maschad:
**Is your feature request related to a problem? Please describe.**

Following up from the discussion on [improving validator status logging](https://github.com/ChainSafe/lodestar/issues/4785) we would like to log the status of a validator lifecycle to give the user more incremental updates about the [status of the validators](https://hackmd.io/ofFJ5gOmQpu1jjHilHbdQQ).

**Describe the solution you'd like**

Ideally, we could log a Validator's voting status when it is discovered by the beacon chain along with some meta-data.

i.e. is a validator's lifecycle going through `pending` -> `active` -> `exited` -> `withdrawn` and then we can log the total validators.

**Additional context**

Upon initialization **Lighthouse** would log

```bash

Nov 23 08:24:23.745 INFO Initialized validators                  enabled: 2, disabled: 0
Nov 23 08:24:23.757 INFO Connected to beacon node                endpoint: http://localhost:5052/, version: Lighthouse/v3.2.1-6d5a2b5/x86_64-linux
Nov 23 08:24:23.758 INFO Initialized beacon node connections     available: 1, total: 1
Nov 23 08:24:23.758 INFO Loaded validator keypair store          voting_validators: 2
Nov 23 08:24:23.764 INFO Validator exists in beacon chain        fee_recipient: 0x0000…abcd, validator_index: 123456, pubkey: 0x00001, service: duties
Nov 23 08:24:23.764 INFO Validator exists in beacon chain        fee_recipient: 0x0000…abcd, validator_index: 123457, pubkey: 0x00001, service: duties
``` 

for instance. As mentioned [here](https://github.com/ChainSafe/lodestar/issues/4785#issuecomment-1324702738) Certainly it owuld be useful to have the `fee_receipient` , `signing_method` and `total_voting_validators` included.

dapplion:
The text `Validator exists in beacon chain` sounds very reasonable and correct. Logging it only at initialization is fine

philknows:
As mentioned in https://github.com/ChainSafe/lodestar/pull/5507#issuecomment-1558752371, there are still some outstanding questions relating to this log change:

- Define how status changes should be handled
- Define how newly discovered validators should be handled

PR #5507, opened 2023-05-17
refactor: revise validator log to output seen instead of exists

philknows:
**Motivation**

Referencing feedback from a user on Discord: https://discord.com/channels/593655374469660673/593655641445367808/1108123487840763977

On the `VC`, due to the frequency of this log being output: `Validator exists in beacon chain`, there is a probable chance that the word `exists` can be easily misinterpreted with the word `exits` - which should be avoided for bad UX.  

**Description**

An alternative syntax for this log output is `Validator seen on beacon chain` which will avoid any misinterpretation.

### DIFF
--- a/packages/validator/src/services/indices.ts
+++ b/packages/validator/src/services/indices.ts
@@ -135,7 +135,7 @@ export class IndicesService {
 
       const pubkeyHex = toHexString(validatorState.validator.pubkey);
       if (!this.pubkey2index.has(pubkeyHex)) {
-        this.logger.info("Validator exists in beacon chain", {
+        this.logger.info("Validator seen on beacon chain", {
           validatorIndex: validatorState.index,
           pubKey: pubkeyHex,
         });


matthewkeil (approved):

matthewkeil:
You may need to update the title to a conventional commit

dapplion (approved):

nflaig:
Just a thought, this log seems a bit verbose as it is printed out per validator. If this is done on startup it might make sense to keep it simple (`Validator index=461468, pubkey=0x8042...10a0 feeRecipient=0x0000`) or maybe as suggested in https://github.com/ChainSafe/lodestar/pull/5161#discussion_r1133312778.

The current implementation also does not work correctly as mentioned in https://github.com/ChainSafe/lodestar/pull/5161#discussion_r1159416159 and https://github.com/ChainSafe/lodestar/pull/5239#discussion_r1158157879.

I think the validator statuses should be printed out each epoch based on clock (data from cache) and not based on fetching `getStateValidators` while discovery/lifecycle logs should printed as soon as status of a validator changes.

Might have to rework this a bit when adding fee recipient and also think about logs for life cycle changes which could be useful for operators
- https://github.com/ChainSafe/lodestar/issues/5347
- https://github.com/ChainSafe/lodestar/issues/5192

This PR addresses some issues but it requires refactoring
- https://github.com/ChainSafe/lodestar/pull/5239

philknows:
Definitely. I'm all for changes that achieve the same messaging in the most efficient way possible. Did we agree upon including `status` for the validator lifecycle in #5192? If yes, that should be included as well on initialization.

As suggested in https://github.com/ChainSafe/lodestar/pull/5161#discussion_r1133312778, it would make sense to just do this once on startup to print:

`Validator index=461468, pubkey=0x8042...10a0 feeRecipient=0x0000 status=pending | active | exited | withdrawn` 

The only time I would foresee this message coming up again is upon re-initialization of the VC or if one of the running validators has a `status` change.

nflaig:
> Did we agree upon including status for the validator lifecycle in https://github.com/ChainSafe/lodestar/issues/5192?

Not sure if that was discussed but printing the status on initialization sounds good to me, just need to define how status changes and newly discovered validators should be handled.

philknows:
>  just need to define how status changes and newly discovered validators should be handled.

We should probably continue discussing this on PR #5239 itself if it's being refactored to include this or directly on the #5192 issue.

wemeetagain:

PR #5509, opened 2023-05-18
fix: correct seen attestation data count per slot metric

twoeths:
**Motivation**

This is on stable mainnet node, count of seen attestation data per slot is not correct because the scrape time is at random part of a slot

Screenshot 2023-05-18 at 11 06 19

**Description**

Count it at the previous slot instead of the current slot


dapplion (approved):

wemeetagain:

Issue #5314, opened 2023-03-28
Attestations are skipped for slots where previous slot tasks take > `SECONDS_PER_SLOT`


nflaig:
Lodestar VC consciously skips attestations for slots if the previous slot tasks take > `SECONDS_PER_SLOT` due to `runAtMostEvery` function.

https://github.com/ChainSafe/lodestar/blob/f136f952ddb861976b3cc9484914c49a06ee8d90/packages/validator/src/util/clock.ts#L90

**Rational behind this**
> If a task happens to take more than one slot to run, we might skip a slot. This is unfortunate,
   however the alternative is to *always* process every slot, which has the chance of creating a
   theoretically unlimited backlog of tasks. It was a conscious decision to choose to drop tasks
   on an overloaded/latent system rather than overload it even more.
   

The rational is understandable, however, based on my observations, the overload will most likely not be on the VC side but rather on the BN side in which case the BN should make sure to not accept more tasks. If `runAtMostEvery` is just implemented to "protect" the BN we should consider removing it. We should also avoid making assumptions on what BN implementation we are dealing with, might be a DVT middleware or any other CL (lighthouse, prysm, etc.).

Other VCs seem to not care if the tasks of the previous slot are still pending and just start tasks for the next slot. I think we should follow the same approach in Lodestar VC to make it more robust against (expected) errors such as timeouts.

**Steps to reproduce**

This issue was observed while testing Lodestar VC with DVT middleware Charon (https://github.com/ObolNetwork/charon) where `getAggregatedAttestation` would not return a response due to missing implementation of [distributed aggregation selections](https://docs.google.com/document/d/1q9jOTPcYQa-3L8luRvQJ-M0eegtba4Nmon3dpO79TMk/mobilebasic) which causes the attestation routine for slot *N* to take more than `SECONDS_PER_SLOT` due to the VC default http timeout of [SECONDS_PER_SLOT](https://github.com/ChainSafe/lodestar/blob/f136f952ddb861976b3cc9484914c49a06ee8d90/packages/validator/src/validator.ts#L83-L84) causing the VC to skip slot *N + 1*.

**Current behavior**

Attestation routines are not properly decoupled. Issues (such as timeouts) in an attestation routine for slot *N* can lead to missing ALL attestations in slot *N + 1*.

**Expected behavior**

Attestation routines for each slot should be independent/decoupled from each other. Issues in a previous routine should not affect the next routine.

[...]

PR #5511, opened 2023-05-21
fix(validator): improve error handling in attestation service

nflaig:
**Motivation**

Errors thrown during attestation routine are not handled within `AttestationService` but instead will be will be logged by `Clock` which does not use [LoggerVC](https://github.com/ChainSafe/lodestar/blob/bc88bddd96a54c8f5fb34666d99e8f0faa329d5a/packages/validator/src/util/logger.ts#L9) due to circular dependency (LoggerVC depends on clock).

https://github.com/ChainSafe/lodestar/blob/bc88bddd96a54c8f5fb34666d99e8f0faa329d5a/packages/validator/src/validator.ts#L73-L74

When the EL is syncing the BN will throw an error during attestation routine of VC which produces an verbose error

```
error: Error on runEvery fn  Error producing attestation - Service Unavailable: Node is syncing - Block's execution payload not yet validated, executionPayloadBlockHash=0xfaf4067706a82a716245fcbd963b3ed502a8fc4a6d1c896c44e9cd7efe1ec425 number=9039112
Error: Error producing attestation - Service Unavailable: Node is syncing - Block's execution payload not yet validated, executionPayloadBlockHash=0xfaf4067706a82a716245fcbd963b3ed502a8fc4a6d1c896c44e9cd7efe1ec425 number=9039112
    at Function.assert (file:///home/devops/goerli/lodestar/packages/api/src/utils/client/httpClient.ts:44:13)
    at AttestationService.produceAttestation (file:///home/devops/goerli/lodestar/packages/validator/src/services/attestation.ts:167:14)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at AttestationService.runAttestationTasksGrouped (file:///home/devops/goerli/lodestar/packages/validator/src/services/attestation.ts:134:36)
    at AttestationService.runAttestationTasks (file:///home/devops/goerli/lodestar/packages/validator/src/services/attestation.ts:102:7)
    at Clock.runAtMostEvery (file:///home/devops/goerli/lodestar/packages/validator/src/util/clock.ts:96:7)
```

**Expected error log**

```
warn: Node is syncing - Error producing attestation - Service Unavailable: Node is syncing - Block's execution payload not yet validated, executionPayloadBlockHash=0x1c11477b79fd8d7d0c72411897fb96efa25c15781f7b78a714377af7b02bc02a number=9039134
```

**Description**

Catch errors thrown by `runAttestationTasksGrouped` to properly handle "Node is syncing" errors and improve logging UX.

### DIFF
--- a/packages/validator/src/services/attestation.ts
+++ b/packages/validator/src/services/attestation.ts
@@ -99,9 +99,11 @@ export class AttestationService {
     } else {
       // Beacon node's endpoint produceAttestationData return data is not dependant on committeeIndex.
       // Produce a single attestation for all committees and submit unaggregated attestations in one go.
-      await this.runAttestationTasksGrouped(duties, slot, signal).catch((e) => {
-        this.logger.error("Error on attestation routine", {slot}, e);
-      });
+      try {
+        await this.runAttestationTasksGrouped(duties, slot, signal);
+      } catch (e) {
+        this.logger.error("Error on attestation routine", {slot}, e as Error);
+      }
     }
   };
 
--- a/packages/validator/src/services/attestation.ts
+++ b/packages/validator/src/services/attestation.ts
@@ -92,14 +92,16 @@ export class AttestationService {
       await Promise.all(
         Array.from(dutiesByCommitteeIndex.entries()).map(([index, duties]) =>
           this.runAttestationTasksPerCommittee(duties, slot, index, signal).catch((e) => {
-            this.logger.error("Error on attestation routine", {slot, index}, e);
+            this.logger.error("Error on committee attestation routine", {slot, index}, e);
           })
         )
       );
     } else {
       // Beacon node's endpoint produceAttestationData return data is not dependant on committeeIndex.
       // Produce a single attestation for all committees and submit unaggregated attestations in one go.
-      await this.runAttestationTasksGrouped(duties, slot, signal);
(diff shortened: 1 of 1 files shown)


packages/validator/src/services/attestation.ts
```diff
@@ -92,14 +92,16 @@ export class AttestationService {
...
           })
         )
       );
     } else {
       // Beacon node's endpoint produceAttestationData return data is not dependant on committeeIndex.
       // Produce a single attestation for all committees and submit unaggregated attestations in one go.
-      await this.runAttestationTasksGrouped(duties, slot, signal);
+      await this.runAttestationTasksGrouped(duties, slot, signal).catch((e) => {
+        this.logger.error("Error on attestation routine", {slot}, e);
+      });
```
twoeths: using both `await` and `catch` is a bit confused. If we don't want to wait for the result of `runAttestationTasksGrouped` we should remove `await` and we can use `catch` there, if we want to wait and catch the error, we should use `try/catch` pattern

nflaig: Let's use try-catch then, waiting for the result is important right now due to `runAtMostEvery`, this might change in the future as discussed here https://github.com/ChainSafe/lodestar/issues/5314. But unless there is a specific use case should probably prefer awaiting results in most cases.

twoeths (approved):

wemeetagain:

PR #5517, opened 2023-05-22
chore: add eslint rule to ensure await in try-catch

nflaig:
**Motivation**

Current eslint rules do not ensure that promises are awaited within a try-catch. This can lead to unhandled promise rejections and unexpected behavior if missed.

**Description**

https://github.com/ChainSafe/lodestar/pull/5117 already added a similar rule but [no-return-await](https://eslint.org/docs/latest/rules/no-return-await) does not consider return await within try-catch. [return-await](https://typescript-eslint.io/rules/return-await/) supersedes this rule and adds support for optionally requiring return await in certain cases by default it will be required [in-try-catch](https://typescript-eslint.io/rules/return-await/#in-try-catch).

- Add eslint rule to ensure await in try-catch
- Fix missing await inside try-catch
- Disable return-await false positives

**Related**

- https://github.com/typescript-eslint/typescript-eslint/issues/994

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -41,6 +41,7 @@ module.exports = {
       },
     ],
     "@typescript-eslint/await-thenable": "error",
+    "@typescript-eslint/return-await": "error",
     "@typescript-eslint/naming-convention": [
       "error",
       {selector: "default", format: ["camelCase"]},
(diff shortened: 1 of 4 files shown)


dapplion (approved):

wemeetagain:

PR #5519, opened 2023-05-22
docs: generating and using flamegraphs

### LINKED ISSUE #5341: Document getting flamechart of live node
To understand what's affecting Lodestar performance our current strategy is to attach a chromium dev tools instance into a node running with `node --inspect`

Those dev tools can render a stack chart by time, but not a regular flamechart by purely stack occurrences. The information exposed by `node --inspect` _should_ be enough to produce a flamechart.

CC: @matthewkeil @nflaig @tuyennhv 

- [x] Research existing tools that can produce and render a flamechart attaching into a running process; i.e. run process as normal for 1 hour, then capture stacks for 12 minutes, then detach.
- [x] Write reproducible steps + scripts for other contributors

matthewkeil:
**Motivation**

We often need do do processor profiling and using the `--inspect` flag has detrimental effect on performance.  It is also difficult to profile for a short period of time and revert performance back to normal so its not great for production nodes.  The existing flamegraph libraries (`0x` is a good example) are not capable of turning on and off for a node in production to get real world data.

**Description**

Includes instructions for how to do cpu profiling on a running production node.

Closes #5341

Follow instructions in `docs/tools/flamegraphs.md`

**Note for Future**

There is a flamegraph panel for Grafana so it should be possible to pull them with a cron job and visualize from our dashboard

https://grafana.com/docs/grafana/latest/panels-visualizations/visualizations/flame-graph/


docs/tools/flamegraphs.md
```diff
@@ -0,0 +1,167 @@
...
+## Modifying Lodestar
+
+Use the following two commands to install `perf` for generating the stack traces.  You may get a warning about needing to restart the VM due to kernel updates. This is nothing to be concerned with and if so, cancel out of the restart dialog.
+
+```bash
+sudo apt-get install linux-tools-common linux-tools-generic
+sudo apt-get install linux-tools-`uname -r`  # empirically this throws if run on the same line above
+```
+
+SSH into the Lodestar instance and modify `~/beacon/beacon_run.sh` to add a necessary flag.
```
dapplion: Each lodestar instance may be running in different configurations. In this case you should specify:
- Best to run Lodestar not dockerized but as a service.
- To run the service use a script that invokes node with <these flags>

Instead of actually printing such a concrete example of `beacon_run.sh` script

docs/tools/flamegraphs.md
```diff
@@ -0,0 +1,167 @@
...
+  --max-old-space-size=4096 \
+  /usr/src/lodestar/packages/cli/bin/lodestar \
+  beacon \
+  --rcConfig /home/devops/beacon/rcconfig.yml
+```
+
+After updating `beacon_run.sh` file, restart the beacon service.
+
+```sh
+admin@12.34.56.78: sudo systemctl restart beacon.service
```
dapplion: Lodestar may be run with different service names, you should specify at the beginning what environment you are assuming for this guide

docs/tools/flamegraphs.md
```diff
@@ -0,0 +1,167 @@
...
+<img src="../images/flamescope/unfiltered-flamegraph.png" alt="flamescope home screen" width="1024"/>
+<img src="../images/flamescope/zoom-in.png" alt="flamescope home screen" width="1024"/>
+
+## Filtering Results
+
+There can be a lot of "noise" in the stack traces with `libc`, `v8` and `libuv` calls.  It is possible to filter the results to make it more useful, but note this will skew the results. Looking at the graph both filtered and unfiltered can be beneficial. The following `sed` command will remove the noise from the stack traces.  You can also use the bottom script programmatically to add/remove filters.
+
+```sh
+sed -r -e "/( __libc_start| uv_| LazyCompile | v8::internal::| node::| Builtins_| Builtin:| Stub:| LoadIC:| \\[unknown\\]| LoadPolymorphicIC:)/d" -e 's/ LazyCompile:[*~]?/ /'
+```
```
dapplion, line 123: With the one-liner sed command the ts script below should not be necessary

dapplion (commented):
nice guide thank you! some minor comments

docs/tools/flamegraphs.md
matthewkeil: Updated.

docs/tools/flamegraphs.md
matthewkeil: Updated.

docs/tools/flamegraphs.md
matthewkeil: Updated.

matthewkeil:
@dapplion I went through the issue one last time to make sure i covered everything and noticed that I had some references stored in my work repo.  I added them to the bottom.

dapplion (approved):
Looks great thanks!

wemeetagain:

PR #5525, opened 2023-05-24
chore: add script to pull dashboards from remote API

dapplion:
**Motivation**

Our DX to update dashboards is not good, since it requires to manually download files and perform some changes to minimize diff. We want to promote faster experimentation with dashboards to ensure they are up to date and persisted in git.

**Download script**

With one command run it lists all dashboards in the API, downloads then, cleans them up ready to be commited. This allows to very fast iteration of dashboards:
- Log into Grafana UI
- Visually edit a dashboard
- Run download script and commit

To use the script:

Create a file `.secrets.env` with envs
```
GRAFANA_API_KEY=$token
GRAFANA_URL=https://yourgrafanaapi.io
```

Run
```
node scripts/download_dashboards.mjs
```

Check git diff of resulting files, commit, push and open PR

**New linter rules**

To ensure the download command is deterministic we need more linter rules. Using the HTTP API does not allow to set "Export for sharing externally" to true. So the fields starting with `__` must be set manually, see https://community.grafana.com/t/export-dashboard-for-external-use-via-http-api/50716

- Force all panels to be expanded. Toggling expand is huge source of big diffs. By forcing panels to always be expanded there's less opportunity for diff and gives full visibility to panels. If a dashboard now feels to long, it should be split into two, not collapsed.
- Insert fixed __inputs field and remove __requirements

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,6 +62,9 @@ packages/beacon-node/trusted_setup.txt
 # docker-compose .env file
 .env
 
+# Grafana secret
+.secrets.env
+
 # Git artifacts
 packages/cli/.git-data.json
 
--- a/scripts/download_dashboards.mjs
+++ b/scripts/download_dashboards.mjs
@@ -1,0 +1,109 @@
+/* eslint-disable
+  @typescript-eslint/explicit-function-return-type,
+  @typescript-eslint/naming-convention,
+  import/no-extraneous-dependencies,
+  no-console
+*/
+
+import fs from "node:fs";
+import path from "node:path";
+import axios from "axios";
+import dotenv from "dotenv";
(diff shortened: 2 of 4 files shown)


dapplion:
Opened a second PR with only dashboard diff to allow easier review of the download scripts here https://github.com/ChainSafe/lodestar/pull/5526

wemeetagain (approved):
amazing, love this

nflaig:
Might wanna update [CONTRIBUTING.md](https://github.com/ChainSafe/lodestar/blob/unstable/CONTRIBUTING.md#contributing-to-grafana-dashboards) to reflect new workflow in another PR.

wemeetagain:

PR #5526, opened 2023-05-24
chore: grafana dashboard lint rules for API exports

dapplion:
**Motivation**

- PR https://github.com/ChainSafe/lodestar/pull/5525 requires more lint rules to make dashboard export deterministic

**Description**

To ensure the download command is deterministic we need more linter rules. Using the HTTP API does not allow to set "Export for sharing externally" to true. So the fields starting with __ must be set manually, see https://community.grafana.com/t/export-dashboard-for-external-use-via-http-api/50716

- Force all panels to be expanded. Toggling expand is huge source of big diffs. By forcing panels to always be expanded there's less opportunity for diff and gives full visibility to panels. If a dashboard now feels to long, it should be split into two, not collapsed.
- Insert fixed __inputs field and remove __requirements

### DIFF
--- a/dashboards/lodestar_block_processor.json
+++ b/dashboards/lodestar_block_processor.json
@@ -1,52 +1,19 @@
 {
   "__inputs": [
     {
-      "name": "DS_PROMETHEUS",
-      "label": "Prometheus",
       "description": "",
-      "type": "datasource",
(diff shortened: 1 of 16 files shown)


wemeetagain (approved):

wemeetagain:

PR #4258, opened 2022-07-07
Drop lerna

dapplion:
**Motivation**

lerna has been used in this monorepo for +3y (added in https://github.com/ChainSafe/lodestar/pull/324). While useful we have had to battle this tool more than treat it like a helpful item:
- https://github.com/ChainSafe/lodestar/pull/4052
- https://github.com/ChainSafe/lodestar/pull/4049
- https://github.com/ChainSafe/lodestar/pull/3990
- https://github.com/ChainSafe/lodestar/pull/3679
- https://github.com/ChainSafe/lodestar/pull/3637
- https://github.com/ChainSafe/lodestar/pull/3601
- https://github.com/ChainSafe/lodestar/pull/2656
- https://github.com/ChainSafe/lodestar/pull/2650
- https://github.com/ChainSafe/lodestar/pull/2649
- https://github.com/ChainSafe/lodestar/pull/1909
- https://github.com/ChainSafe/lodestar/pull/666

It also deprecates scripts introduced here, not used currently AFAIK
- https://github.com/ChainSafe/lodestar/pull/1142

**Description**

We use lerna for three actions:
- Run scripts in all packages: Now done `yarn workspaces run <cmd>`
- Bump version and dependencies of local packages: Now done with `scripts/release/set_version.mjs`
- Publish all packages: Now done `scripts/release/publish.mjs`

Please check the new scripts and you'll see they are really simple. I don't think they justify a tool so heavy and problematic like lerna. I don't guarantee this PR won't introduce bugs for each type of release, but fixing those should be trivial now that we control the tooling.

**Please note that we have dropped 2_300 lines of dependencies from yarn.lock :tada:**

PR #5531, opened 2023-05-24
chore: fix build:watch scripts

nflaig:
**Motivation**

Currently `build:watch` scripts (added in https://github.com/ChainSafe/lodestar/pull/1142) are broken as they reference outdated commands removed in https://github.com/ChainSafe/lodestar/pull/2933.

PR to drop lerna (https://github.com/ChainSafe/lodestar/pull/4258) suggests to remove those scripts but I find them quite handy for faster feedback cycles and generally observing build errors. Running `yarn build:ifchanged` works but is not ideal in my opinion.

**Description**

- removes `build:types:watch` as `build:types` no longer exists
- rename `build:lib:watch` to `build:watch` and update reference to correct build command
- add missing `build:watch` script to packages

~~Running this from root directory causes `MaxListenersExceededWarning` warnings but might not be worth further investigate how to prevent this as it does not impact functionality~~ 

Max listeners warning is fixed by using `lerna exec` instead of `lerna run` (see [stackoverflow](https://stackoverflow.com/a/55655749/10577550)).


dapplion (dismissed):
LGTM

wemeetagain (approved):

wemeetagain:

PR #5538, opened 2023-05-24
refactor: improve the attached validator finalized proposal stats log ux

g11tech:
as per discord discussion, it was felt to only log to info attached proposal stats when relevant:

    - No need to explicitly track SYNCED state since no validators attached would be there to show
    - debug log if validators attached but no proposals were scheduled
    - info log if proposals were scheduled (canonical) or there were orphans (non canonical)


packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -216,8 +216,8 @@ export function createLodestarMetrics(
 
     // Finalized block and proposal stats
     allValidators: {
-      expected: register.gauge({
-        name: "lodestar_all_validators_expected_count",
+      total: register.gauge({
+        name: "lodestar_all_validators_total_count",
```
nflaig, line 220: non-intuitive metric name, might wanna refactor now or later

```suggestion
        name: "lodestar_all_validators_proposals_total_count",
```

or

```suggestion
        name: "lodestar_all_validators_blocks_total_count",
```

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -216,8 +216,8 @@ export function createLodestarMetrics(
 
     // Finalized block and proposal stats
     allValidators: {
-      expected: register.gauge({
-        name: "lodestar_all_validators_expected_count",
+      total: register.gauge({
```
nflaig, line 219: Do we consider these metrics to be part of validator monitor now? Maybe should add them to `validatorMonitor`
https://github.com/ChainSafe/lodestar/blob/f96c5a18bf3e1c636abe069bcde22169f6f01284/packages/beacon-node/src/metrics/metrics/lodestar.ts#L675

packages/cli/src/options/beaconNodeOptions/chain.ts
```diff
@@ -53,6 +55,12 @@ export const options: CliCommandOptions<ChainArgs> = {
     group: "chain",
   },
 
+  validatorMonitor: {
```
nflaig: Not sure about this flag, I feel like capturing metrics should always be enabled (if metrics enabled) whereas logs should be opt-in as those are way more intrusive.

Also Lodestar has a validator monitor already where we capture most (or even all?) metrics about attestations
https://github.com/ChainSafe/lodestar/blob/f96c5a18bf3e1c636abe069bcde22169f6f01284/packages/beacon-node/src/metrics/validatorMonitor.ts#L181

Maybe better to wait adding a cli flag that disables stuff until we implement validator monitor logs
- https://github.com/ChainSafe/lodestar/issues/5336

I like the strategy and implementation from Lighthouse ([validator monitoring](https://lighthouse-book.sigmaprime.io/validator-monitoring.html)), could get some inspiration from there.

packages/beacon-node/src/chain/archiver/index.ts
```diff
@@ -301,21 +303,37 @@ export class Archiver {
...
+          finalizedAttachedValidatorsCount,
+        });
+      }
+
+      this.metrics?.attachedValidators.total.set(attachedValidators.total);
+      this.metrics?.attachedValidators.finalized.set(attachedValidators.finalized);
+      this.metrics?.attachedValidators.orphaned.set(attachedValidators.orphaned);
+      this.metrics?.attachedValidators.missed.set(attachedValidators.missed);
+    } else {
+      this.logger.debug("No attached Proposers to the beacon", {finalizedEpoch: finalized.epoch});
```
nflaig:
```suggestion
      this.logger.debug("No proposers attached to beacon node", {finalizedEpoch: finalized.epoch});
```

packages/beacon-node/src/chain/archiver/index.ts
```diff
@@ -301,21 +303,37 @@ export class Archiver {
...
+        this.logger.info("Attached validators finalized proposal stats", {
+          ...attachedValidators,
+          finalizedAttachedValidatorsCount,
+        });
+      } else {
+        this.logger.debug("Attached validators finalized proposal stats", {
+          ...attachedValidators,
+          finalizedAttachedValidatorsCount,
+        });
+      }
```
nflaig: Why do we need to check both `total` and `orphaned`? I would assume orphaned blocks to be counted in total blocks therefore checking if total is 0 should be sufficient.

Potential refactoring/deduplication

```suggestion
      this.logger[attachedValidators.total !== 0 ? LogLevel.debug : LogLevel.info](
        "Attached validators finalized proposal stats",
        {
          ...attachedValidators,
          finalizedAttachedValidatorsCount,
        }
      );
```

packages/beacon-node/src/chain/archiver/index.ts
```diff
@@ -301,21 +303,37 @@ export class Archiver {
...
-    this.metrics?.allValidators.expected.set(allValidators.expected);
+    // Only log to info if there is some relevant data to show
+    //  - No need to explicitly track SYNCED state since no validators attached would be there to show
+    //  - debug log if validators attached but no proposals were scheduled
+    //  - info log if proposals were scheduled (canonical) or there were orphans (non canonical)
+    if (finalizedAttachedValidatorsCount !== 0) {
+      if (attachedValidators.total !== 0 || attachedValidators.orphaned !== 0) {
+        this.logger.info("Attached validators finalized proposal stats", {
+          ...attachedValidators,
+          finalizedAttachedValidatorsCount,
```
nflaig: Not a fan of `finalizedAttachedValidatorsCount` as it is way to hard to read

```suggestion
          validators: finalizedAttachedValidatorsCount,
```

We do not omit any information by simplifying this as
- we know that those are attached validators as it is already part of message
- what are even finalized validators? This name might make sense in code but not in the log
- should be obvious it is the count as we show a number `validators=10`

packages/beacon-node/src/chain/beaconProposerCache.ts
```diff
@@ -7,19 +7,28 @@ const PROPOSER_PRESERVE_EPOCHS = 2;
 
 export type ProposerPreparationData = routes.validator.ProposerPreparationData;
 
+/** Cache to track the validators registored with beacon node. Could eventually be promoted as a
```
nflaig:
```suggestion
/** Cache to track the validators registered with beacon node. Could eventually be promoted as a
```

nflaig: ` Could eventually be promoted as a full monitor` is not clear to me, what is remaining for monitoring block proposals?

packages/beacon-node/src/chain/beaconProposerCache.ts
```diff
@@ -7,19 +7,28 @@ const PROPOSER_PRESERVE_EPOCHS = 2;
...
+ * full monitor
+ */
 export class BeaconProposerCache {
+  /** Flag to switch on/off validator stats default set to true */
+  readonly validatorMonitor: boolean;
```
nflaig: recommend removing for now

nflaig (changes requested):

packages/beacon-node/src/metrics/metrics/lodestar.ts
nflaig: @dapplion is validator monitor something defined in the spec (haven't found anything). It seems so well defined between clients (or at least compared to Lighthouse) or did we just copy what they did?

**Edit:** Answering myself, does look like our implementation is inspired by Lighthouse
- https://github.com/ChainSafe/lodestar/pull/2474

packages/beacon-node/src/chain/beaconProposerCache.ts
g11tech: you can repurpose the flag later on if required, whats the point of removing it since this is what you have been asking for?

packages/beacon-node/src/chain/beaconProposerCache.ts
g11tech: tbh i am not sure, monitoring can be for other things as well, but this class holds the registered validators so can naturally be become a validator monitor

packages/beacon-node/src/chain/archiver/index.ts
g11tech: pls check comment at L310

packages/beacon-node/src/metrics/metrics/lodestar.ts
g11tech: hmmm... i think expected was more intutive than total for me ... but anyway whatever makes things more intuitive to the users. :+1:

packages/beacon-node/src/metrics/metrics/lodestar.ts
g11tech: yes we can move metrics there, nice suggestion

packages/beacon-node/src/chain/archiver/index.ts
nflaig: ok maybe I am not understanding this correctly, what does expected (now called total) mean? My naive assumption without looking to deep into the code was it just summing up finalized + missed + orphaned

packages/cli/src/options/beaconNodeOptions/chain.ts
g11tech: hmm, we can think about validator monitoring in a more holistic way again not the scope of the PR, if the flag is causing confusion lets rename it 

(my though process of a validator monitor is a module/call enable and hold and crunch monitoring metadata, not just a property to hold metrics, but i concede i haven't though a lot about it since the scope of the PR is limited)

packages/beacon-node/src/metrics/metrics/lodestar.ts
nflaig: As per discussion in discord it was not clear what is meant by "expected", user might think it means "scheduled" aka lookahead of upcoming blocks.

Not sure if total is the correct term, see my other comment, maybe I am misunderstanding it

packages/cli/src/options/beaconNodeOptions/chain.ts
nflaig: yeah not sure myself, let's keep this PR simple and rethink later.

packages/beacon-node/src/chain/beaconProposerCache.ts
nflaig: can keep it here, mostly about the CLI flag. I am not sure having a flag which enabled validator monitoring metrics is correct approach

packages/beacon-node/src/chain/archiver/index.ts
g11tech: there can be multiple forks in the chain, so the `expected` is as per the canonical chain (and one may have produced blocks in non canonical chain so they are now oprhaned)

for e.g. as per the fork which was finalized, you may not have any proposal, but in the fork in which you proposed, you might have proposed a block which is now orphaned 

so orphaned could be part of expected (for e.g. fork was correct but you send block late which other validators didn't include) or it could be the above scenario.

packages/beacon-node/src/chain/beaconProposerCache.ts
g11tech: removed the validatorMonitor flag complications for future PRs

packages/beacon-node/src/chain/beaconProposerCache.ts
g11tech: removed

packages/beacon-node/src/metrics/metrics/lodestar.ts
g11tech: not mixing in other goals with the PR as of now, lets track and address separately

nflaig (approved):
LGTM!

All points mentioned in https://github.com/ChainSafe/lodestar/issues/5255#issuecomment-1559627103 have been addressed

- Make it sync aware (so it doesn't show while syncing) ✅ 
- Disable if no validators are attached ✅ 
- Only display in `info` if non-zero. else it goes to `debug` ✅ 
- Change the word `expected` to `total` to minimize confusion on logger ✅

wemeetagain (approved):

nflaig:
I think it is pretty good that we moved this to debug now if stats are 0 because it looks like it really just shows this log once (maybe twice?) if there is a proposal and then it shows 0 again (potentially confusing to users). I would assume this is due to pruning the state.

```
May-26 09:12:00.079[chain]           debug: Attached validators finalized proposal stats total=0, finalized=0, orphaned=0, missed=0, validators=1962
May-26 09:18:24.066[chain]           debug: Attached validators finalized proposal stats total=0, finalized=0, orphaned=0, missed=0, validators=1962
May-26 09:24:48.098[chain]            info: Attached validators finalized proposal stats total=2, finalized=2, orphaned=0, missed=0, validators=1962
May-26 09:31:12.067[chain]           debug: Attached validators finalized proposal stats total=0, finalized=0, orphaned=0, missed=0, validators=1962
May-26 09:37:36.105[chain]           debug: Attached validators finalized proposal stats total=0, finalized=0, orphaned=0, missed=0, validators=1962
```

wemeetagain:

PR #5541, opened 2023-05-25
refactor(network): move unclean disconnect log to debug

nflaig:
**Motivation**

https://github.com/ChainSafe/lodestar/issues/5359#issuecomment-1519129570

**Description**

Move unclean disconnect log to from `warn` to `debug`

### DIFF
--- a/packages/beacon-node/src/network/peers/peerManager.ts
+++ b/packages/beacon-node/src/network/peers/peerManager.ts
@@ -628,7 +628,7 @@ export class PeerManager {
     try {
       await this.libp2p.hangUp(peer);
     } catch (e) {
-      this.logger.warn("Unclean disconnect", {peer: prettyPrintPeerId(peer)}, e as Error);
+      this.logger.debug("Unclean disconnect", {peer: prettyPrintPeerId(peer)}, e as Error);
     }
   }
 


packages/beacon-node/src/network/peers/peerManager.ts
```diff
@@ -628,7 +628,7 @@ export class PeerManager {
     try {
       await this.libp2p.hangUp(peer);
     } catch (e) {
-      this.logger.warn("Unclean disconnect", {peer: prettyPrintPeerId(peer)}, e as Error);
+      this.logger.debug("Unclean disconnect", {peer: prettyPrintPeerId(peer)}, e as Error);
```
nflaig, line 631: Anyone knows why `prettyPrintPeerId` only prints out the first 2 characters? Those seem to be always 16 , e.g. `peer=16...LaZW3U`.

I saw users wonder why peer 16 is always disconnecting 🤔

twoeths: yeah that's a good question. For this context I suppose we should print the whole peer id as it's for developer, not for user. We can ask other client for a specific peer id for example

nflaig: Moving this from warn to debug will solve most of the confusion for users as most do not look at debug logs. We are using `prettyPrintPeerId` all over the place in peer manager, maybe could print first 4 characters instead of just 2.

> We can ask other client for a specific peer id for example

Not sure how they would look up the peer id but would assume if it just based on searching the logs it should even be sufficient to just have the truncated peer id.

Anyways, changes on how we want to log the peer ids should probably be addressed in another PR.

twoeths (approved):

wemeetagain:

Issue #5566, opened 2023-05-29
Add more eslint rules


dapplion:
typescript-eslint has many rules that we don't currently use but would be benefitial. I've read all of them and propose this new eslintrc. I'm opening this issue first to:
- Get :+1: from contributors on the set of rules here
- Progressively implement them. Running this new eslintrc triggers +6000 errors and warnings

```js
{
  extends: [
    "eslint:recommended",
    "plugin:import/errors",
    "plugin:import/warnings",
    "plugin:import/typescript",
    "plugin:@typescript-eslint/strict",
    "plugin:@typescript-eslint/recommended-requiring-type-checking",
  ],
  rules: {
    // @typescript-eslint/*, sorted
    "@typescript-eslint/array-type": "error",
    "@typescript-eslint/await-thenable": "error",
    "@typescript-eslint/ban-ts-comment": "error",
    "@typescript-eslint/ban-types": "error",
    "@typescript-eslint/consistent-generic-constructors": "error",
    "@typescript-eslint/consistent-indexed-object-style": "error",
    "@typescript-eslint/consistent-type-assertions": "error",
    "@typescript-eslint/consistent-type-exports": "error",
    "@typescript-eslint/consistent-type-imports": "error",
    "@typescript-eslint/explicit-function-return-type": ["error", {allowExpressions: true}],
    "@typescript-eslint/explicit-member-accessibility": ["error", {accessibility: "no-public"}],
    "@typescript-eslint/func-call-spacing": "error",
    "@typescript-eslint/naming-convention": [
      "error",
      {selector: "default", format: ["camelCase"]},
      {
        selector: ["classProperty", "objectLiteralProperty", "classMethod", "parameter"],
        format: ["camelCase"],
        leadingUnderscore: "allow",
      },
      //variable must be in camel or upper case
      {selector: "variable", format: ["camelCase", "UPPER_CASE"], leadingUnderscore: "allow"},
      //classes and types must be in PascalCase
      {selector: ["typeLike", "enum"], format: ["PascalCase"]},
      {selector: "enumMember", format: null},
      //ignore rule for quoted stuff
      {
        selector: [
          "classProperty",
          "objectLiteralProperty",
          "typeProperty",
          "classMethod",
          "objectLiteralMethod",
          "typeMethod",
          "accessor",
          "enumMember",
        ],
        format: null,
        modifiers: ["requiresQuotes"],
      },
      //ignore rules on destructured params
      {
        selector: "variable",
[...]

PR #5567, opened 2023-05-29
ci: assert eslintrc rules are sorted

dapplion:
**Motivation**

- Minimize diff for rule addition, to ease efforts like https://github.com/ChainSafe/lodestar/issues/5566

**Description**

Add CI step to ensure eslintrc rules are sorted

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,30 +18,21 @@ module.exports = {
   plugins: ["@typescript-eslint", "eslint-plugin-import", "@chainsafe/eslint-plugin-node", "prettier"],
   extends: [
     "eslint:recommended",
-    "plugin:import/errors",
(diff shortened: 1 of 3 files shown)


nflaig (approved):
LGTM!

Just a hint from GPT-4

> It's important to note that the order of rules in eslint configurations can matter. For example, if you're using a plugin that extends eslint's functionality and there's a conflict between two rules, the rule that appears last in the configuration will take precedence. So, even if the two configurations have the same rules, if they're in a different order, they might not function identically.

It does not seem to matter in our case as there are no lint errors.

twoeths (approved):

wemeetagain (approved):

wemeetagain:

PR #5562, opened 2023-05-29
chore: add rule @typescript-eslint/restrict-template-expressions

dapplion:
**Motivation**

We have had many instances of attempting to log data in bad formats, this rule prevents such errors

- See https://github.com/ChainSafe/lodestar/issues/4650

**Description**

Closes https://github.com/ChainSafe/lodestar/issues/4650

PR #5568, opened 2023-05-29
feat: add --private flag

### LINKED ISSUE #2438: `--private` mode CLI option to not send impl details

dapplion:
**Motivation**

Add option to prevent node from sending impl/version data over indentify protocol

**Description**

Closes https://github.com/ChainSafe/lodestar/issues/2438

### DIFF
--- a/packages/cli/src/cmds/beacon/handler.ts
+++ b/packages/cli/src/cmds/beacon/handler.ts
@@ -156,8 +156,15 @@ export async function beaconHandlerInit(args: BeaconArgs & GlobalArgs) {
   const {peerId, enr} = await initPeerIdAndEnr(args, beaconPaths.beaconDir, logger);
   // Inject ENR to beacon options
   beaconNodeOptions.set({network: {discv5: {enr: enr.encodeTxt(), config: {enrUpdate: !enr.ip && !enr.ip6}}}});
-  // Add simple version string for libp2p agent version
-  beaconNodeOptions.set({network: {version: version.split("/")[0]}});
+
+  if (args.private) {
+    beaconNodeOptions.set({network: {private: true}});
+  } else {
+    // Add simple version string for libp2p agent version
+    beaconNodeOptions.set({network: {version: version.split("/")[0]}});
+    // Casting to `as string` so that if version type changes this line does not compile
+    beaconNodeOptions.set({executionBuilder: {userAgent: `Lodestar/${getVersionData().version as string}`}});
+  }
 
   // Render final options
   const options = beaconNodeOptions.getWithDefaults();
--- a/packages/cli/src/cmds/beacon/handler.ts
+++ b/packages/cli/src/cmds/beacon/handler.ts
@@ -162,7 +162,7 @@ export async function beaconHandlerInit(args: BeaconArgs & GlobalArgs) {
   } else {
     // Add simple version string for libp2p agent version
     beaconNodeOptions.set({network: {version: version.split("/")[0]}});
-    // Casting to `as string` so that if version type changes this line does not compile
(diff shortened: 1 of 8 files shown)


nflaig:
Shall we also consider a global CLI flag, `--private` which prevents all kinds of information leaks as mentioned in https://github.com/ChainSafe/lodestar/pull/5361#discussion_r1181517044?

E.g. also prevent sending user-agent to relayers
- https://github.com/ChainSafe/lodestar/pull/5361

This would make it easier for users with privacy/security concerns as they would just have to add a single flag without having to know which parts of the client leaks information.

packages/cli/src/cmds/beacon/handler.ts
```diff
@@ -156,8 +156,15 @@ export async function beaconHandlerInit(args: BeaconArgs & GlobalArgs) {
...
   beaconNodeOptions.set({network: {discv5: {enr: enr.encodeTxt(), config: {enrUpdate: !enr.ip && !enr.ip6}}}});
-  // Add simple version string for libp2p agent version
-  beaconNodeOptions.set({network: {version: version.split("/")[0]}});
+
+  if (args.private) {
+    beaconNodeOptions.set({network: {private: true}});
+  } else {
+    // Add simple version string for libp2p agent version
+    beaconNodeOptions.set({network: {version: version.split("/")[0]}});
+    // Casting to `as string` so that if version type changes this line does not compile
```
nflaig: Curious about the comment, isn't the point of using types to catch those issue. I mean this line should not compile if it is broken because we changed some type of `version`. 

As per typescript compiler, it does not even complain if `version` is an object. Good news is we catch this now via eslint rule added in https://github.com/ChainSafe/lodestar/pull/5562.

packages/cli/src/cmds/beacon/handler.ts
```diff
@@ -156,8 +156,15 @@ export async function beaconHandlerInit(args: BeaconArgs & GlobalArgs) {
...
+    beaconNodeOptions.set({executionBuilder: {userAgent: `Lodestar/${getVersionData().version as string}`}});
```
nflaig:
```suggestion
    beaconNodeOptions.set({executionBuilder: {userAgent: `Lodestar/${version}`}});
```

packages/cli/src/cmds/beacon/options.ts
```diff
@@ -97,6 +98,11 @@ export const beaconExtraOptions: CliCommandOptions<BeaconExtraArgs> = {
     description: "Whether to reuse the same peer-id across restarts",
     type: "boolean",
   },
+
+  private: {
+    description: "Do not send implementation and version name over p2p identify protocol and builder requests",
```
nflaig: Potential simplification

```suggestion
    description: "Do not send implementation details over p2p identify protocol and in builder requests",
```

packages/cli/src/cmds/beacon/handler.ts
dapplion: Indeed, with https://github.com/ChainSafe/lodestar/pull/5562 it's not necessary. It's annoying tho how type relaxed typescript is

nflaig: I guess typescript is just lax here because of how template literals work in javascript, but it's definitely not ideal

nflaig (approved):

wemeetagain (approved):

wemeetagain:

PR #5569, opened 2023-05-29
feat: change the blob txs encoding assumption to be opaque rlp

g11tech:
Treat the blob txs as opaque rlp like any other tx,
- https://github.com/ethereum/EIPs/pull/6985

now the essential validation of beacon block commitments and payload's versioned hashes is entrusted to EL via new payload v3
- https://github.com/ethereum/consensus-specs/pull/3359

TODO
- [x] Update the commitment to transaction helper fn used in mock and test


wemeetagain (approved):

wemeetagain:

Issue #2491, opened 2021-05-05
Move API consumer code to chain class


dapplion:
**Feature request**

Basic ops like getBlock or getState and not clearly defined in Lodestar. The code in the API to do that could be moved to BeaconChain class so it can be consumed programatically from there too

3xtr4t3rr3str14l:
@dapplion i guess since these ops have been moved to the `lodestar-api` package that we can close this issue?

dapplion:
> @dapplion i guess since these ops have been moved to the `lodestar-api` package that we can close this issue?

No, this issue is about moving the code from api/impl to chain

PR #5570, opened 2023-05-29
refactor: move block getter from API impl to chain

dapplion:
**Motivation**

Our strategy to store blocks in not trivial and should be managed by the chain class not the API impl. The API impl should be concerned only with resolving the request payload and then forwarding to the chain.

**Description**

Chain exposes two methods for it

```ts
  /**
   * Since we can have multiple parallel chains,
   * this methods returns blocks in current chain head according to
   * forkchoice. Works for finalized slots as well
   */
  getCanonicalBlockAtSlot(
    slot: Slot
  ): Promise<{block: allForks.SignedBeaconBlock; executionOptimistic: boolean} | null>;
  /**
   * Get local block by root, does not fetch from the network
   */
  getBlockByRoot(root: RootHex): Promise<{block: allForks.SignedBeaconBlock; executionOptimistic: boolean} | null>;
```

An edge case important to note is what happens if an archive operation happens **during** a getBlock request. In the future we may want to have a lock mechanism on during archiving, which will be handled by the chain class

- See https://github.com/ChainSafe/lodestar/issues/2491

Move block getter from API impl to chain


wemeetagain (approved):

wemeetagain:

PR #5572, opened 2023-05-29
refactor: move state getter from API impl to chain

### LINKED ISSUE #2491: Move API consumer code to chain class

### LINKED ISSUE #3683: Move pseudo-regen code from beacon API to regen
**Describe the bug**

Beacon API code includes regen code that's very sensitive in terms of performance and memory.

https://github.com/ChainSafe/lodestar/blob/aba0815ee5b6571c882335717bdf2aa304e3c193/packages/lodestar/src/api/impl/beacon/state/utils.ts#L269-L277

This code should be moved into regen and unified there to ensure the caches and its usage is consistent and safe.

dapplion:
**Motivation**

Regenerating states is a very sensitive part of the become node. Currently there's a code path buried deep in /api/impl/utils that can take significant resources from API queries. We want to have a tight control over what and when is reading states from disk and calling regen.

For the reason the api impl should only be concerned in understanding the request of the API consumer and querying the chain class for a state.

- Related to https://github.com/ChainSafe/lodestar/pull/5570

**Description**

Expose two need methods in BeaconChain class to be called from API code

```ts
  /** Returns a local state cannonical at `slot` */
  getStateBySlot(
    slot: Slot,
    opts?: StateGetOpts
  ): Promise<{state: BeaconStateAllForks; executionOptimistic: boolean} | null>;
  /** Returns a local state by state root */
  getStateByStateRoot(
    stateRoot: RootHex,
    opts?: StateGetOpts
  ): Promise<{state: BeaconStateAllForks; executionOptimistic: boolean} | null>;
```

Replace the custom code reading from state cache, db and running potentially unbounded state transitions with calls to regen. This PR surfaces the shortcomings of our API capability to return data at arbitrary states. However those shortcomings do not impact stakers at all and should be handled properly in another PR.

Closes https://github.com/ChainSafe/lodestar/issues/2491
Closes https://github.com/ChainSafe/lodestar/issues/3683


wemeetagain (approved):

wemeetagain:

PR #5573, opened 2023-05-29
feat: skip serializing block after fetching from network

### LINKED ISSUE #3657: Skip serializing blocks when persisting to db

dapplion:
**Motivation**

When we fetch blocks from network we don't need to re-serialize them again.

**Description**

Avoid the extra work of serialization by keeping a reference to the serialized payload as the block travels through the codepaths.

Generalizes @tuyennhv 's approach to apply not only to gossip but to all syncs, by modifying the BlockInput:

https://github.com/ChainSafe/lodestar/blob/1a5a578f4d32497ddbf09601652e643ffdc2b266/packages/beacon-node/src/chain/blocks/types.ts#L20-L22

Closes https://github.com/ChainSafe/lodestar/issues/3657

### DIFF
--- a/packages/beacon-node/src/api/impl/beacon/blocks/index.ts
+++ b/packages/beacon-node/src/api/impl/beacon/blocks/index.ts
@@ -218,12 +218,14 @@ export function getBeaconBlockApi({
             config,
             signedBlock,
             signedBlobs.map((sblob) => sblob.message)
-          )
+          ),
+          null
         );
       } else {
         signedBlock = signedBlockOrContents;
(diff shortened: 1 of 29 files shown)


twoeths (approved):

wemeetagain:

PR #5485, opened 2023-05-12
feat: trigger block search for unknown block attestations

twoeths:
**Motivation**

- At gossip time, we usually get unknown block attestations because gossip block comes after attestations, and we want to actively search for that block instead of waiting for gossip block passively

**Description**
- New `NetworkEvent.unknownBlock` network event, emitted when unknown block attestations
- Refactor unknownBlock sync to handle both `unknownBlockParent` and `unknownBlock` event
- At each block there is usually a race between gossip block and rpc block. Most of the time gossip block wins.
- Prevent unbundling attack

Closes #3613

PR #5574, opened 2023-05-30
chore: add metrics to unknown block sync panel

twoeths:
**Motivation**

#5485 was merged, at each slot there's a race between gossip block and unknown block sync so we want to track unknown block sync received time

**Description**

- Correct "Action/slot" panel: add "type" label
- Insert "Received Time" panel to "Unknown Block Sync" row as 2nd panel (that makes it more diff than usual)

Screenshot 2023-05-30 at 09 51 38


wemeetagain (approved):

wemeetagain:

PR #5582, opened 2023-05-30
test: remove chain mocks

dapplion:
**Motivation**

MockBeaconChain class is very annoying to have to update everytime we change its interface. It's actually unnecessary, so this PR just drops it completely.

**Description**

Remove manual chain mocks of:
- BeaconChain
- ForkChoice


wemeetagain:
I like the motivation, but the e2e tests are failing

packages/beacon-node/src/network/network.ts
```diff
@@ -105,7 +103,6 @@ export class Network implements INetwork {
     this.chain.emitter.on(routes.events.EventType.head, this.onHead);
     this.chain.emitter.on(routes.events.EventType.lightClientFinalityUpdate, this.onLightClientFinalityUpdate);
     this.chain.emitter.on(routes.events.EventType.lightClientOptimisticUpdate, this.onLightClientOptimisticUpdate);
-    modules.signal.addEventListener("abort", this.close.bind(this), {once: true});
```
dapplion, line 108: This is a terrible idea, debugging tests I discovered that close() is being called twice:
- first time through AbortController, not awaited, uncontrolled
- second time via close() but since it's second time it has no effect

This broke the expected close flow from the caller, causing issues in subsequent tests

dapplion (commented):
While attempting to fix the use of BeaconChain mock in network tests, I noticed that they can use the **actual** BeaconChain class and it works fine.

So the big refactor of the last commit is adding a new function `getNetworkForTest()` which instantiates a network class with the real BeaconChain thing as dependency

wemeetagain (approved):

wemeetagain:

PR #5584, opened 2023-05-30
chore: consistent file naming convention

### LINKED ISSUE #5111: Consistent file naming convention

Not having a consistent naming convention cause problems different issues.

1. Different conventions used by different team members. 
2. Difficult to quickly find a particular file
3. Sometimes difficulty relating context of piece of code with other

Some examples are: 

_camelCase_
https://github.com/ChainSafe/lodestar/blob/unstable/packages/beacon-node/src/chain/blocks/importBlock.ts
https://github.com/ChainSafe/lodestar/blobunstable/packages/reqresp/src/encoders/requestDecode.ts

_CapitalCase_
https://github.com/ChainSafe/lodestar/blob/unstable/packages/reqresp/src/ReqResp.ts
https://github.com/ChainSafe/lodestar/blob/unstable/packages/beacon-node/src/chain/clock/LocalClock.ts

I would like to recommend to have consistent naming convention which follow these rules. 

1. We should strictly use `snake_case` for file names
2. If file exports one class, the file name should be the class name in `snake_case`
3. If file exports multiple classes then we should split it to different files. 
5. If file exports multiple functions, then we should have a generic name in `snake_case`
6. All tests files should have name related to top level `describe` in snake_case. 
7. All directories should follow `snake_case` names.

Other approach was to follow `camelCase` which is already followed on most of the files. The problems with that approach are following: 

1. Windows OS does have [case-insensitive file system](https://learn.microsoft.com/en-us/windows/wsl/case-sensitivity). So having `camelCase` reduces our target community for contribution and can cause unseen problem on Windows OS. 
2. Same applies to [Mac OS](https://discussions.apple.com/thread/251191099) which is more widely used development platform. 
3. `camelCase` does not have better readability and also does not match the export names that we have in `CapitalCase`. 

**Additional context**

IN many different softwares, there is a known set of observations and errors that caused due to not handling the proper case sensitive file names. Please check the list in following link.

https://cwe.mitre.org/data/definitions/178.html
https://cwe.mitre.org/data/definitions/19.html

nazarhussain:
**Motivation**

Have a consistent and clean codebase. 

**Description**

Use a consistent file name convention for all packages. 

Closes #5111

Run all tests.


philknows:
To reduce risk of merge conflicts, we should look at this after we release v1.9.0 and merge v1.9.0 `stable` back to `unstable`. This one will require coordination as mentioned in Discord on May 30.

dapplion (changes requested):
Before doing a full impl of the change the whole time should agree on what casing to use. While snake_case is nice it's not what we are used to and not the most common casing in this project. To minimize diff let's start with camelCase which is the most used in the codebase.

Also for this type of huge changes you should outline exactly what programmatic approach did you use to apply this change in the codebase. Otherwise you force contributors into auditing the results, not the process

nazarhussain:
> Before doing a full impl of the change the whole time should agree on what casing to use. While snake_case is nice it's not what we are used to and not the most common casing in this project. To minimize diff let's start with camelCase which is the most used in the codebase.
> 
> Also for this type of huge changes you should outline exactly what programmatic approach did you use to apply this change in the codebase. Otherwise you force contributors into auditing the results, not the process

1. I thought the thumbs on the issue shows the agreement, and also no one highlighted anything during the meeting. So I thought this issue is good to go. 
2. Related to changes, if we use the `camelCase` we are not actually covering the issue of case sensitivity outlined in the issue, and making a change is not effective for the project. 
3. Once the PR would have been ready, I would have highlighted the process and special cases in the PR to focus the review.

dapplion:

Issue #4501, opened 2022-09-01
Make network classes startable once


dapplion:
Network classes can be started and stopped. This pattern is "forced" on us because of libp2p, but we don't needed.

This forces Lodestar to mantain multiple AbortControllers, since you need a new one for each start. However if we change the flow to only be startable once a single AbortController can be passed as constructor parameter, and throw on start if it's already aborted.

PR #5587, opened 2023-05-30
refactor: class BeaconDb as close only

dapplion:
**Motivation**

Lodestar does not support start stop syntax in general. It's unnecessary for downstream classes to expose such behavior

- Part of https://github.com/ChainSafe/lodestar/issues/4501

**Description**

BeaconDb as close only class


packages/beacon-node/src/db/beacon.ts
```diff
@@ -23,7 +24,12 @@ import {
 } from "./repositories/index.js";
 import {PreGenesisState, PreGenesisStateLastProcessedBlock} from "./single/index.js";
 
-export class BeaconDb extends DatabaseService implements IBeaconDb {
+export type BeaconDbModules = {
```
nflaig, line 27: unused type

packages/validator/src/validator.ts
```diff
@@ -224,6 +225,7 @@ export class Validator {
   async close(): Promise<void> {
     if (this.state === Status.closed) return;
     this.controller.abort();
+    await this.db.close();
```
nflaig, line 228: db is already close here

https://github.com/ChainSafe/lodestar/blob/1f4ba2728ae400083d0c77684e4d8cc4bed8a544/packages/cli/src/cmds/validator/handler.ts#L106

maybe it is good to call it here as well, just to make sure..also if validator is used without CLI it is required

wemeetagain (approved):

wemeetagain:

PR #5588, opened 2023-05-30
refactor: drop start stop pattern from Network where possible

### LINKED ISSUE #4501: Make network classes startable once

dapplion:
**Motivation**

Lodestar does not support start stop syntax in general. It's unnecessary for downstream classes to expose such behavior

**Description**

upstream dependencies have the start stop pattern baked in and is not worth it to change (libp2p itself, gossip, reqresp).

This PR just changes local classes which don't really need the start stop pattern

Closes https://github.com/ChainSafe/lodestar/issues/4501


wemeetagain (approved):

wemeetagain:

PR #5537, opened 2023-05-24
fix(logger): increase max listeners of file transports to 20

nflaig:
**Motivation**

Closes https://github.com/ChainSafe/lodestar/issues/5529

**Description**

Increase max listeners of file transports to prevent `MaxListenersExceededWarning` warnings. Each child logger (`logger.child`) adds a new listener. Setting a reasonable limit to still detect potential memory leaks. 

This workaround is also suggested in https://github.com/winstonjs/winston/issues/2247#issuecomment-1335191439 and seems appropriate in our case as based on https://github.com/winstonjs/winston/issues/1685#issuecomment-729627681 it does not look like this is an issue, it should ok to increase max listeners limit to suppress the warnings.

Issue #5529, opened 2023-05-24
`MaxListenersExceededWarning` when creating >10 child loggers


nflaig:
`MaxListenersExceededWarning` warnings for `DailyRotateFile` on current `unstable` branch, might be related to changes done in
- https://github.com/ChainSafe/lodestar/pull/5490

Logs with `--trace-warnings`

```
(node:13306) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 unpipe listeners added to [DailyRotateFile]. Use emitter.setMaxListeners() to increase limit
    at _addListener (node:events:587:17)
    at DailyRotateFile.addListener (node:events:605:10)
    at DerivedLogger.Readable.pipe (/home/nico/projects/ethereum/lodestar/node_modules/readable-stream/lib/_stream_readable.js:631:8)
    at DerivedLogger.add (/home/nico/projects/ethereum/lodestar/node_modules/winston/lib/winston/logger.js:364:10)
    at /home/nico/projects/ethereum/lodestar/node_modules/winston/lib/winston/logger.js:124:44
    at Array.forEach (<anonymous>)
    at DerivedLogger.configure (/home/nico/projects/ethereum/lodestar/node_modules/winston/lib/winston/logger.js:124:18)
    at new Logger (/home/nico/projects/ethereum/lodestar/node_modules/winston/lib/winston/logger.js:42:10)
    at new DerivedLogger (/home/nico/projects/ethereum/lodestar/node_modules/winston/lib/winston/create-logger.js:44:7)
(node:13306) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added to [DailyRotateFile]. Use emitter.setMaxListeners() to increase limit
    at _addListener (node:events:587:17)
    at DailyRotateFile.prependListener (node:events:619:14)
    at prependListener (/home/nico/projects/ethereum/lodestar/node_modules/readable-stream/lib/_stream_readable.js:98:69)
    at DerivedLogger.Readable.pipe (/home/nico/projects/ethereum/lodestar/node_modules/readable-stream/lib/_stream_readable.js:708:3)
    at DerivedLogger.add (/home/nico/projects/ethereum/lodestar/node_modules/winston/lib/winston/logger.js:364:10)
    at /home/nico/projects/ethereum/lodestar/node_modules/winston/lib/winston/logger.js:124:44
    at Array.forEach (<anonymous>)
[...]

PR #5597, opened 2023-06-01
fix(logger): avoid creating new winston instance for each child

nflaig:
**Motivation**

As discussed in https://github.com/ChainSafe/lodestar/pull/5490#discussion_r1209317683

Method to create child loggers was updated in
- https://github.com/ChainSafe/lodestar/pull/5490

This causes issues with no clear benefit
- https://github.com/ChainSafe/lodestar/issues/5529

**Description**

Reverts `logger.child` implementation back to original approach https://github.com/ChainSafe/lodestar/blob/375d660ed1e7ba13945dff41e63154a26c4be008/packages/utils/src/logger/winston.ts#L82

This avoids `MaxListenersExceededWarning` warnings as it no longer creates a new winston instance for each child logger.

Reverts https://github.com/ChainSafe/lodestar/pull/5537 as the workaround is no longer needed.

### DIFF
--- a/packages/logger/src/node.ts
+++ b/packages/logger/src/node.ts
@@ -2,19 +2,12 @@ import path from "node:path";
 import DailyRotateFile from "winston-daily-rotate-file";
 import TransportStream from "winston-transport";
 import winston from "winston";
-import {Logger, LogLevel, logLevelNum, TimestampFormat} from "./interface.js";
+import type {Logger as Winston} from "winston";
+import {Logger, LogLevel, TimestampFormat} from "./interface.js";
 import {ConsoleDynamicLevel} from "./utils/consoleTransport.js";
-import {getFormat} from "./utils/format.js";
 import {WinstonLogger} from "./winston.js";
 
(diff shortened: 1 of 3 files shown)


packages/logger/src/node.ts
```diff
@@ -128,36 +121,34 @@ interface DefaultMeta {
...
   }
 
   static fromNewTransports(opts: LoggerNodeOpts): WinstonLoggerNode {
-    return new WinstonLoggerNode(opts, getNodeLoggerTransports(opts));
+    return WinstonLoggerNode.fromOpts(opts, getNodeLoggerTransports(opts));
   }
 
-  // Return a new logger instance with different module and log level
-  // but a reference to the same transports, such that there's only one
-  // transport instance per tree of child loggers
```
nflaig, line 152: @dapplion This comment should no longer be relevant also it seems inaccurate as we are not returning a new logger instance with different log level.

wemeetagain (approved):

wemeetagain:

PR #5598, opened 2023-06-01
fix(logger): restore previous --logFileDailyRotate behavior

nflaig:
**Motivation**

https://github.com/ChainSafe/lodestar/pull/5490#discussion_r1209337604

**Description**

Restores previous `--logFileDailyRotate` behavior

- `lodestar --logFileDailyRotate` -> enable daily rotate with default value (same as omitting the flag)
- `lodestar --logFileDailyRotate 10` -> set daily rotate to custom value 10
- `lodestar --logFileDailyRotate 0` -> disable daily rotate and accumulate in same file

This logic was previously handled by the CLI package

https://github.com/ChainSafe/lodestar/blob/a4b29cfe38de109c44992977074869a8c01cfc83/packages/cli/src/util/logger.ts#L72-L82

### DIFF
--- a/packages/logger/test/unit/loggerTransport.test.ts
+++ b/packages/logger/test/unit/loggerTransport.test.ts
@@ -127,7 +127,7 @@ describe("winston transport log to file", () => {
     const filenameRx = /^child-logger-test/;
     const filepath = path.join(tmpDir, filename);
 
-    const logger = getNodeLoggerTest({module: "a", file: {filepath, level: LogLevel.info}});
+    const logger = getNodeLoggerTest({module: "a", file: {filepath, level: LogLevel.info, dailyRotate: 5}});
 
     stdoutHook = hookProcessStdout();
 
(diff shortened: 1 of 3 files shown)


packages/logger/test/unit/loggerTransport.test.ts
```diff
@@ -127,7 +127,7 @@ describe("winston transport log to file", () => {
     const filenameRx = /^child-logger-test/;
     const filepath = path.join(tmpDir, filename);
 
-    const logger = getNodeLoggerTest({module: "a", file: {filepath, level: LogLevel.info}});
+    const logger = getNodeLoggerTest({module: "a", file: {filepath, level: LogLevel.info, dailyRotate: 5}});
```
nflaig, line 130: IMO better to set `dailyRotate` here to use `DailyRotateFile` which is the more commonly used file transport

wemeetagain (approved):

wemeetagain:

Issue #4005, opened 2022-05-10
Regen refactor > safer resilient strategy


dapplion:
## Background

Consensus clients need to cache some states to fully participate in the network. States are very heavy so you can't cache all the states that you may need. Writing all the possible states you may need to disk is not practical either. So what do you do?

- Keep in memory the few states that you need the most
- Regenerate from memory or disk (i.e. re-process blocks) to access states that probabilistically are less useful.

So the stateCache and checkpointStateCache handle the first point: deciding which states to keep in memory. The regen module handles the second: provide the ability to regenerate _any_ state, within some boundary.

## Current strategy

- Keep in memory the most recent 96 states regardless of forks
- Keep in memory checkpoint states for the most recent ~4 epochs
- Keep in memory the latest finalized state + the head state
- Write to disk the latest finalized state, delete the previous finalized state

This approach works well for good network conditions. Thanks for tree structural sharing the cost of those 96 states in a linear chain is very low multiple of the cost of a single state (~1.2x).

However, during attacks, bugs or highly forked network our node quickly runs out of memory or can become unable to follow the chain.
- If those 96 states are significantly different between each other structural sharing is not useful so the total memory could become 96x the cost of a single state.
- Same for checkpoint states, see past example of this, which caused fast OOMs https://github.com/ChainSafe/lodestar/issues/3171
- Structural sharing is only useful for states that are close to each other. In long periods of non-finality we would regen from latest finalized state which could be hours old, potentially DOS-ing ourselves.

Relevant issues:
- https://github.com/ChainSafe/lodestar/issues/3099

## Improvements goals

So, we can do better. Specifically:

1. Don't let the state cache cause an OOM if states are too expensive
2. Limit max regen cost in all cases = reduce DOS risk
3. Make regen as cheap as possible using both regen from memory states and disk states

## Proposed strategies

**1. Regen from memory and disk**

On every checkpoint write a state to disk to a "hot state db" bucket. On finalization, move some of those states "cold state db" or "archive db" bucket. Then on regen, use those states depending on the max distance of the closest available state in memory if any. This would allow to drop the need to keep the finalized state in memory.

- https://github.com/ChainSafe/lodestar/issues/2154

**2. Bound regen depending on consumer**

Depending on the caller, restrict the work triggered by regen

- https://github.com/ChainSafe/lodestar/issues/3683
- https://github.com/ChainSafe/lodestar/issues/2846

**3. WeakRef state cache**

[...]

PR #5599, opened 2023-06-01
refactor: do not expose state caches outside regen

dapplion:
**Motivation**

State cache handling is a very sensitive part of Lodestar and should be handled with care. Exposing the state cache data structure directly encourages assumptions about which state may or may not be available. IMO consumers should always go through regen as the ultimate provider of _any_ state under _some_ constraints (sync only, allow replay, etc)

**NOTE**: there's no change of logic**, just added enough new methods to map 1:1 current unstable

- Related to https://github.com/ChainSafe/lodestar/issues/4005

**Description**

Make state caches data structure internal to regen only


wemeetagain (approved):
lgtm

wemeetagain:

PR #5602, opened 2023-06-02
feat: add validator monitor metric for missed attestation reason

### LINKED ISSUE #5376: Automatically debug missed attestations
While our fleet performance is pretty good there are still some missed attestations sometimes. Some of them are caused by external factors, other for internal ones.

A beacon node has enough information to do a fairly good guess of why an attestation was missed. Usually @tuyennhv performs that process manually but it should be feasible to automate it as part of the validator monitor. That would help us focus efforts and quantitatively measure our efforts in improving Lodestar's performance.

A possible mechanism for this:
- After 32 slot (or some safe reorg distance) observe which validators where expected to attest and if the did so
- If the state claims a missed attestation check:
  - has the attestation been seen?
  - has the attestation been published late?
  - how many peers was the attestation published to?
  - was the attestation seen as part of an aggregate?
  - Is the attestation voting for the correct head and target?

dapplion:
**Motivation**

Beacon nodes sometimes miss attestations, either due to network conditions or internal problems. This PR extends the validator monitor with heuristics to try to guess _why_ a registered validator did not obtain the maximum attester reward possible.

- https://github.com/ChainSafe/lodestar/issues/5376

**Description**

Assumptions:
- Must handle short re-orgs of 1-5 slot depth
- Assumes somewhat healthy network conditions: no slashings, no deep re-orgs of > 32 slots

First, what's a missed attestation?
- Half way through the last slot of the epoch, the validator monitor is called with the current head state
- Iterates previous epoch participation and if some flag is set to false
- Uses tracked data from API submission and gossip inputs to guess the reason of penalty

The main function to review is this one :point_down: 

https://github.com/ChainSafe/lodestar/blob/9a5828d37c58d71f0a952d73ee2ff1f55bcbf7f1/packages/beacon-node/src/metrics/validatorMonitor.ts#L683-L695

Closes https://github.com/ChainSafe/lodestar/issues/5376

### DIFF
--- a/packages/beacon-node/src/metrics/validatorMonitor.ts
+++ b/packages/beacon-node/src/metrics/validatorMonitor.ts
@@ -1,8 +1,18 @@
-import {computeEpochAtSlot, AttesterStatus, parseAttesterFlags} from "@lodestar/state-transition";
-import {Logger} from "@lodestar/utils";
-import {allForks, altair} from "@lodestar/types";
-import {ChainForkConfig} from "@lodestar/config";
-import {MIN_ATTESTATION_INCLUSION_DELAY, SLOTS_PER_EPOCH} from "@lodestar/params";
+import {
+  computeEpochAtSlot,
+  AttesterStatus,
+  parseAttesterFlags,
+  CachedBeaconStateAllForks,
+  CachedBeaconStateAltair,
+  parseParticipationFlags,
+  computeStartSlotAtEpoch,
+  getBlockRootAtSlot,
+  ParticipationFlags,
+} from "@lodestar/state-transition";
+import {Logger, MapDef, MapDefMax, toHex} from "@lodestar/utils";
+import {RootHex, allForks, altair} from "@lodestar/types";
+import {ChainConfig, ChainForkConfig} from "@lodestar/config";
+import {ForkSeq, INTERVALS_PER_SLOT, MIN_ATTESTATION_INCLUSION_DELAY, SLOTS_PER_EPOCH} from "@lodestar/params";
 import {Epoch, Slot, ValidatorIndex} from "@lodestar/types";
 import {IndexedAttestation, SignedAggregateAndProof} from "@lodestar/types/phase0";
 import {LodestarMetrics} from "./metrics/lodestar.js";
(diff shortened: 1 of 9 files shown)


packages/beacon-node/src/metrics/validatorMonitor.ts
```diff
@@ -1,16 +1,32 @@
...
  * Historical data will be kept around for `HISTORIC_EPOCHS` before it is pruned.
  */
-const HISTORIC_EPOCHS = 4;
+const MAX_CACHED_EPOCHS = 4;
+
+const MAX_CACHED_DISTINCT_TARGETS = 4;
+
+const INTERVALS_LATE_ATTESTATION_SUBMISSION = 1.5;
+
+const RETAIN_REGISTERED_VALIDATORS_MS = 12 * 3600 * 1000; // 12 hours
```
dapplion, line 29: Taking advantage of `onceEveryEndOfEpoch` this PR also cleans validators not seen in a while. 12 hours is an arbitrary value that seems long enough, but please feel free to suggest others

packages/beacon-node/src/metrics/validatorMonitor.ts
```diff
@@ -156,26 +180,47 @@ function withEpochSummary(validator: MonitoredValidator, epoch: Epoch, fn: (summ
...
+
 /// A validator that is being monitored by the `ValidatorMonitor`. */
 type MonitoredValidator = {
-  /// The validator index in the state. */
-  index: number;
   /// A history of the validator over time. */
   summaries: Map<Epoch, EpochSummary>;
   inSyncCommitteeUntilEpoch: number;
+  // Unless the validator slashes itself, there MUST be one attestation per target checkpoint
+  attestations: MapDefMax<Epoch, MapDefMax<TargetRoot, AttestationSummary>>;
```
dapplion, line 222: This PR assumes:
- There's strictly one attestation per tuple of (validator index, target_checkpoint)
- If validator submits the same attestation twice, the second event is ignored
- An attestation can be included in multiple blocks, which may or may not be cannonical

packages/state-transition/src/block/processAttestationsAltair.ts
```diff
@@ -26,6 +26,7 @@ const PROPOSER_REWARD_DOMINATOR = ((WEIGHT_DENOMINATOR - PROPOSER_WEIGHT) * WEIG
 const TIMELY_SOURCE = 1 << TIMELY_SOURCE_FLAG_INDEX;
 const TIMELY_TARGET = 1 << TIMELY_TARGET_FLAG_INDEX;
 const TIMELY_HEAD = 1 << TIMELY_HEAD_FLAG_INDEX;
+const SLOTS_PER_EPOCH_SQRT = intSqrt(SLOTS_PER_EPOCH);
```
dapplion, line 29: Micro-optimization I noted from this PR, there's no need to compute the square root over and over for a constant value

g11tech (approved):
lgtm

dapplion:
Dashboard example

image

wemeetagain:

Issue #5438, opened 2023-04-30
Keymanager API: decrypt keystores in a thread pool


nflaig:
Similar to https://github.com/ChainSafe/lodestar/pull/5357, when importing keystores through the keymanager API we should decrypt in a thread pool to reduce the time it takes to import a lot of keys.

- decrypt in a thread pool (in parallel)
- limit with a queue to prevent killing the vc process

Based on existing tooling and use cases, there seems to be less of an requirement to import a lot of keys via API. Most operators with a lot of keys use `--importKeystores` or pre-import keys via `validator import`. Other solutions that use keymanager API often call import per keystore separately as there is also the max body limit restriction enforced by the API which needs to be overriden separately by setting `--keymanager.bodyLimit`.

nflaig:
Might fix [issues faced by operators](https://twitter.com/Stefa2kEth/status/1628407296790343680) that import a lot of keys in a single request as it offloads keystore decryption to a worker thread.

PR #5608, opened 2023-06-05
fix: adjust decryption worker pool size based on keystore count

nflaig:
**Motivation**

Spawning more worker threads than the number of keystores to decrypt does not make any difference as those workers would just be idle.

Especially during testing where we call `decryptKeystoreDefinitions` multiple times, the extra overhead of spawning those additional workers is quite noticeable.

This change reduces the time of running decrypt tests by at least 50%.

This change is especially important once https://github.com/ChainSafe/lodestar/issues/5438 is implemented as spawning too many workers when handling multiple concurrent requests to `/eth/v1/keystores` (with possibly just a single keystore in each request) would blow up the validator client.

**Description**

Adjust decryption worker pool size based on keystore count.


wemeetagain (approved):
LGTM :)

wemeetagain:

PR #5614, opened 2023-06-06
feat(prover): add whitelist unverified methods support to prover

nazarhussain:
**Motivation**

Make the prover proxy useable for real world Dapp development. 

**Description**

Add support for providing a white list of methods those can be forwarded. 

Partially #4706 

Run all tests


packages/prover/src/cli/cmds/start/options.ts
```diff
@@ -52,6 +55,14 @@ export const startOptions: CliCommandOptions<StartArgs> = {
       "The trusted checkpoint root to start the lightclient. If not provided will initialize from the latest finalized slot. It shouldn't be older than weak subjectivity period",
     type: "string",
   },
+
+  "unverified-whitelist": {
+    description: `Comma separated list of methods which are allowed to forward. If not provided, all methods are allowed.${alwaysAllowedMethods.join(
+      ","
```
matthewkeil:
```suggestion
    description: `Comma separated list of methods which are allowed to forward. If not provided, all methods are allowed.  ${alwaysAllowedMethods.join(
```

packages/prover/src/interfaces.ts
```diff
@@ -13,12 +13,13 @@ export enum LCTransport {
...
 
-export type RootProviderInitOptions = {
-  signal: AbortSignal;
-  logger: Logger;
-  wsCheckpoint?: string;
-} & ConsensusNodeOptions &
-  NetworkOrConfig;
+export type RootProviderInitOptions = ConsensusNodeOptions &
+  NetworkOrConfig & {
+    signal: AbortSignal;
```
matthewkeil, line 18: It's much easier to see these at the top!  Very nice touch!

packages/prover/src/utils/process.ts
```diff
@@ -24,22 +24,42 @@ export const supportedELRequests: Record<string, ELVerifiedRequestHandler<any, a
...
 
   for (const pay of isBatchRequest(payload) ? payload : [payload]) {
-    if (isRequest(pay) && !isNullish(supportedELRequests[pay.method])) {
+    if (isRequest(pay) && verifiableMethods.includes(pay.method)) {
       verifiable.push(pay);
-    } else {
+      continue;
+    }
+
+    // If unverifiedWhitelist is not set that implies all methods are allowed
```
matthewkeil, line 48: This was a helpful comment!

matthewkeil (dismissed):
Approving because looks great! Just the spacing in the first comment. Running the tests now

matthewkeil:
@lodestar/prover unit tests all passed!  🎉

packages/prover/src/cli/cmds/start/options.ts
```diff
@@ -52,6 +55,14 @@ export const startOptions: CliCommandOptions<StartArgs> = {
       "The trusted checkpoint root to start the lightclient. If not provided will initialize from the latest finalized slot. It shouldn't be older than weak subjectivity period",
     type: "string",
   },
+
+  "unverified-whitelist": {
```
nflaig: On the casing topic, almost all of the CLI flags on Lodestar are using camelCase. For the prover to have same look and feel would probably make sense to align this.

nazarhussain: I updated and renamed all cli options for the prover to camelCase.

packages/prover/README.md
```diff
@@ -53,7 +53,9 @@ lodestar-prover start \
 
 ⌛ - Todo
 
-➡️ - Request will be forward to EL without any intermediary manipulations.
+➡️ - Request will be forward to EL without any intermediary manipulations. You can limit these by providing `unverifiedWhitelist` option for provider or `--unverified-whitelist` from the cli. If not set then all methods will be forwarded.
```
nflaig: any reference to CLI flags need to be updated, maybe in another refactor PR if there are more changes than just here

```suggestion
➡️ - Request will be forward to EL without any intermediary manipulations. You can limit these by providing `unverifiedWhitelist` option for provider or `--unverifiedWhitelist` from the cli. If not set then all methods will be forwarded.
```
I guess another benefit for using camelCase is that CLI flags and library options are the same now 👍

packages/prover/src/cli/cmds/start/options.ts
```diff
@@ -33,35 +36,42 @@ export const startOptions: CliCommandOptions<StartArgs> = {
...
   },
 
-  "ws-checkpoint": {
+  wsCheckpoint: {
     description:
       "The trusted checkpoint root to start the lightclient. If not provided will initialize from the latest finalized slot. It shouldn't be older than weak subjectivity period",
     type: "string",
   },
+
+  unverifiedWhitelist: {
```
nflaig: Just curious, is there a specific reason why this is not handled as an array? 

The pattern I usually see is that if a list of values is allowed we use `array` which gives the option to provide the flag multiple times or alternatively space-separated values.
This would work by default and if you also want to support comma-separated values, could add custom `coerce` function as we do here
https://github.com/ChainSafe/lodestar/blob/7c8e5f8e31e29bdd78e6c66551919819a2dd07d7/packages/cli/src/cmds/validator/options.ts#L280

packages/prover/src/interfaces.ts
```diff
@@ -78,6 +79,10 @@ export type ConsensusNodeOptions =
...
   | {transport: LCTransport.P2P; bootnodes: string[]};
 
-export type VerifiedExecutionInitOptions = LogOptions &
-  ConsensusNodeOptions &
-  NetworkOrConfig & {wsCheckpoint?: string; signal?: AbortSignal};
+export type RootProviderOptions = {
+  wsCheckpoint?: string;
+  signal?: AbortSignal;
+  unverifiedWhitelist?: string[];
+};
```
nflaig: This is really nuanced but sorting options based on actual options (e.g. wsCheckpoint) and modules (e.g. signal) might be helpful

```suggestion
export type RootProviderOptions = {
  signal?: AbortSignal;
  wsCheckpoint?: string;
  unverifiedWhitelist?: string[];
};
```

packages/prover/src/utils/process.ts
```diff
@@ -24,22 +24,42 @@ export const supportedELRequests: Record<string, ELVerifiedRequestHandler<any, a
   eth_estimateGas: eth_estimateGas,
 };
 
-export function splitRequestsInChunks(payload: JsonRpcRequestOrBatch): {
+export const verifiableMethods = Object.keys(verifiableMethodHandlers);
+export const alwaysAllowedMethods = ["eth_subscribe", "eth_unsubscribe", "eth_getProof"];
```
nflaig, line 28: does it make sense to have an enum or defined typed for all the available methods?

nflaig (dismissed):
LGTM! just left a few remarks

packages/prover/src/utils/process.ts
nazarhussain: Don't think so enum is gonna help here. We are not gonna reuse these anywhere else as types.

packages/prover/src/cli/cmds/start/options.ts
nazarhussain: I didn't knew `yargs` allows space separated values as flag options. Normally that's not the case for a Linux CLI interfaces. Any value without a flag is considered as argument not as flag options. 

I will double check and if it is then update the PR.

nflaig (approved):

matthewkeil (approved):

wemeetagain:

PR #5616, opened 2023-06-06
chore(scripts): add clean:nm to root package.json

matthewkeil:
**Motivation**

I find that when switching branches I occasionally need to delete the node_modules folder and re-run yarn.  This is a helpful script I added and deleted a couple times.  Figured may be helpful for others.

**Description**

1-line package.json script to delete the root and package `node_modules` folders

from root folder

```sh
git checkout mkeil/clean-nm-script
yarn clean:nm
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   ],
   "scripts": {
     "clean": "rm -rf ./packages/*/lib ./packages/*/*.tsbuildinfo",
+    "clean:nm": "rm -rf ./packages/*/node_modules ./node_modules",
     "build": "lerna run build",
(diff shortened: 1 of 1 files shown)


package.json
nflaig, line 1: Let's reserve `feat` for actual user facing features. Would recommend `chore` here instead

nflaig (approved):
Looks good! I have found myself running that exact command as well. 

Alternative solution to the "switching branches problem" would be to use git worktrees but not using them myself currently.

package.json
matthewkeil: Good idea!  Updated.

wemeetagain:

PR #5617, opened 2023-06-06
fix: simplify snappy frame encoding

wemeetagain:
**Motivation**

During upgrade to node 20, I noticed that our `snappy` dependency was out of date.
But worse, this dependency was hidden in our snappy frame compression library, which is a poorly maintained, untyped fork at https://github.com/chainsafe/node-snappy-stream

**Description**
This is a minimal implementation of snappy frame compression that doesn't rely on node streams, stream-to-it, and half-a-dozen other libraries in node-snappy-stream. Instead it uses an async generator to add the framing.

The `snappy` dependency has also been updated to v7.x.x


dapplion (commented):
Can you add a test with pre-generated compress data? Current tests check the loop data -> compress -> uncompress -> check eq. Would be good to have data -> compress --> check eq, uncompress -> check eq, where the compressed data is generated with the previous version proven to work

wemeetagain:
We already have those tests :) `reqresp/test/unit/encoders/encodingStrategies/sszSnappy/{encode,decode}.ts`

wemeetagain:
@ChainSafe/lodestar ready for rereview

dapplion (dismissed):
LGTM!

dapplion (approved):

wemeetagain:

PR #5618, opened 2023-06-07
refactor: remove intermediary not-started state from peer discovery

dapplion:
**Motivation**

Noted that our node takes longer than necessary to start the first discv5 query. This is because there's a race condition between calling the first heartbeat loop and discv5 worker being ready.

The race condition is silenced here, where consumers do not realize underlying worker is not ready yet

https://github.com/ChainSafe/lodestar/blob/c775ad7bba0a4399bcc347b2551283c0fafff792/packages/beacon-node/src/network/discv5/index.ts#L117-L123

This is not ideal and can lead to issues, this hidden intermediary state can be avoided by either:
- throw error if not started
- await for worker to be initialized
- remove the state where worker is not ready yet completely

**Description**

remove intermediary not-started state from peer discovery, by adopting the `static async init` + `close` pattern. IMO is the solution most simple to reason about with less gotchas downstream


wemeetagain (dismissed):

wemeetagain (approved):

wemeetagain:

PR #5619, opened 2023-06-07
refactor: drop observable usage in Discv5Worker for vanila events

dapplion:
**Motivation**

After merging the network worker PR we are managing events ourselves, no need to depend on Observables if we don't use their features

**Description**

Drop observable usage in Discv5Worker for vanila events

Builds on https://github.com/ChainSafe/lodestar/pull/5618


dapplion:

PR #5623, opened 2023-06-08
refactor: reusable thread pool to decrypt keystores

nflaig:
**Motivation**

Decrypting keystores is required in multiple places
- when importing from keystore definition files ([#5357](https://github.com/ChainSafe/lodestar/pull/5357))
- when importing keystores through the keymanager API ([#5438](https://github.com/ChainSafe/lodestar/issues/5438))

This PR makes the thread pool implementation reusable, improves pool lifecycle management and error handling.

**Description**

Reusable thread pool to decrypt keystores, addresses the following issues
- terminates workers if process receives exit signal
- ensures consistent configuration of thread pool
- enables worker to decrypt from file or keystore string
- more graceful and deterministic error handling
- improved error handling and formatting 👇 

**Current error handling**

Current error thrown if decryption of keystores fails is not that helpful if there are multiple. There is no indication on which keystore file is affected and only reports one error.

```
✖ Error: Invalid password
    at Object.<anonymous> (/home/devops/goerli/lodestar/node_modules/@chainsafe/bls-keystore/lib/functional.js:86:19)
    at Generator.next (<anonymous>)
    at fulfilled (/home/devops/goerli/lodestar/node_modules/@chainsafe/bls-keystore/lib/functional.js:5:58)
```

**Updated error handling**

Updated error handling provides more details on which keystore import failed and why it failed.

Single keystore import error

```
 ✖ Error importing keystore

keystore-m_12381_3600_2_0_0-1673696854.json: Invalid password
```

Multiple keystore import errors

```
 ✖ Multiple errors importing keystores

keystore-m_12381_3600_2_0_0-1673696854.json: Invalid password
keystore-m_12381_3600_8_0_0-1673696857.json: Invalid password
keystore-m_12381_3600_5_0_0-1673696855.json: Invalid password
```

Single error, multiple aborted due to error

```
 ✖ Error importing keystore

keystore-malformed.json: Unexpected token s in JSON at position 0

Aborted 100 pending keystore imports
```

Multiple errors, multiple aborted due to errors

```
 ✖ Multiple errors importing keystores

keystore-m_12381_3600_1028_0_0-1680958115.json: Invalid password
keystore-m_12381_3600_1031_0_0-1680958117.json: Invalid password
keystore-m_12381_3600_1032_0_0-1680958118.json: Invalid password
keystore-m_12381_3600_1045_0_0-1680958125.json: Invalid password
keystore-m_12381_3600_1049_0_0-1680958127.json: Invalid password
[...]

PR #5624, opened 2023-06-08
feat: decrypt keystores imported through API in thread pool

### LINKED ISSUE #5438: Keymanager API: decrypt keystores in a thread pool

nflaig:
**Motivation**

Closes https://github.com/ChainSafe/lodestar/issues/5438

**Description**

Decrypt keystores in a thread pool when importing through keymanager API
- reduces the time it takes to import a lot of keys (~5x faster)
- offloads keystore decryption to a worker thread to keep main thread stable
- keeps VC stable during keystore import, it is still able to perform all validator duties in time
- thread pool queues tasks to prevent killing the VC process

Depends on
- https://github.com/ChainSafe/lodestar/pull/5623

### DIFF
--- a/packages/cli/src/cmds/validator/keymanager/impl.ts
+++ b/packages/cli/src/cmds/validator/keymanager/impl.ts
@@ -148,28 +148,24 @@ export class KeymanagerApi implements Api {
         decryptKeystores.queue(
           {keystoreStr, password},
           (secretKeyBytes: Uint8Array) => {
-            try {
-              const secretKey = bls.SecretKey.fromBytes(secretKeyBytes);
+            const secretKey = bls.SecretKey.fromBytes(secretKeyBytes);
 
-              // Persist the key to disk for restarts, before adding to in-memory store
-              // If the keystore exist and has a lock it will throw
-              this.persistedKeysBackend.writeKeystore({
-                keystoreStr,
-                password,
-                // Lock immediately since it's gonna be used
-                lockBeforeWrite: true,
-                // Always write, even if it's already persisted for consistency.
-                // The in-memory validatorStore is the ground truth to decide duplicates
-                persistIfDuplicate: true,
-              });
+            // Persist the key to disk for restarts, before adding to in-memory store
+            // If the keystore exist and has a lock it will throw
+            this.persistedKeysBackend.writeKeystore({
+              keystoreStr,
+              password,
+              // Lock immediately since it's gonna be used
+              lockBeforeWrite: true,
+              // Always write, even if it's already persisted for consistency.
+              // The in-memory validatorStore is the ground truth to decide duplicates
+              persistIfDuplicate: true,
+            });
 
-              // Add to in-memory store to start validating immediately
-              this.validator.validatorStore.addSigner({type: SignerType.Local, secretKey});
+            // Add to in-memory store to start validating immediately
+            this.validator.validatorStore.addSigner({type: SignerType.Local, secretKey});
 
-              statuses[i] = {status: ImportStatus.imported};
-            } catch (e) {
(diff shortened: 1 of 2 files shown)


packages/cli/src/cmds/validator/keymanager/impl.ts
```diff
@@ -142,30 +145,43 @@ export class KeymanagerApi implements Api {
...
+              statuses[i] = {status: ImportStatus.error, message: (e as Error).message};
+            }
+          },
+          (e: Error) => {
+            statuses[i] = {status: ImportStatus.error, message: e.message};
+          }
+        );
       } catch (e) {
         statuses[i] = {status: ImportStatus.error, message: (e as Error).message};
       }
```
matthewkeil: Why the nested catch and the error callback here.  Could we have this catch in `queue` by making it async and awaiting the tasks and passing errors sync/async to `onError`

https://github.com/ChainSafe/lodestar/blob/d89400cc46525061304a4eabdce1ea860d2f617f/packages/cli/src/cmds/validator/keymanager/decryptKeystores/threadPool.ts#L43-45

matthewkeil (changes requested):

packages/cli/src/cmds/validator/keymanager/impl.ts
nflaig: I am not a big fan of having these 3 error handlers/catch blocks here but each of them handles different error cases

Lets just look at what each of those cases are

---
**The outer try-catch:** https://github.com/ChainSafe/lodestar/blob/d89400cc46525061304a4eabdce1ea860d2f617f/packages/cli/src/cmds/validator/keymanager/impl.ts#L178
1. no password error (L136)
2. parse error (L139)
3. getPubkeyHexFromKeystore error (L140)
4. and very unlikely queue errors (L148)

I think could move that block up to only handle cases 1-3 as it is unlikely that `queue` throws an error

---
**The `onError` handler:** https://github.com/ChainSafe/lodestar/blob/d89400cc46525061304a4eabdce1ea860d2f617f/packages/cli/src/cmds/validator/keymanager/impl.ts#L174

This one just catches any errors that are thrown by the worker that decrypt the keystore e.g. invalid password

---
**The inner try-catch:** https://github.com/ChainSafe/lodestar/blob/d89400cc46525061304a4eabdce1ea860d2f617f/packages/cli/src/cmds/validator/keymanager/impl.ts#L170

Catches error due to failure to persist keystore to disk `writeKeystore` (L156)

---

Is your idea to make `onDecrypted` callback async? If that is the case I don't think `onError` would catch it unless we would do something like this

```ts
task.then(result => {
  try {
    onDecrypted(result);
  } catch (error) {
    onError(error);
  }
}).catch(onError);
```

Maybe would help to add some comments to individual error handlers to make it bit easier to follow

nflaig: ok nvm my comment ☝️  I think you are right can just handle it in `onError` handler and don' even need to make it async (it is wrapped into a promise automagically)

matthewkeil (approved):
LGTM 🚀

twoeths (approved):

wemeetagain:

Issue #5404, opened 2023-04-23
BLOCK_ERROR_ALREADY_KNOWN publishing blinded block


twoeths:
**Describe the bug**

Error was reported from here https://discord.com/channels/593655374469660673/593655641445367808/1099742962499461221

![](https://cdn.discordapp.com/attachments/593655641445367808/1099742962323312670/image.png)

Block was published successfully by the relay, error was thrown by lodestar unlike https://github.com/ChainSafe/lodestar/issues/5140 , could be because we process block after publishing it https://github.com/ChainSafe/lodestar/blob/ad0c58ffd3a875c869b1e28cc1afc23c2f962696/packages/beacon-node/src/api/impl/beacon/blocks/index.ts#L234

**Expected behavior**

- No error to user, add verbose/debug log for developer

PR #5627, opened 2023-06-11
fix: prevent unhandled rejections in racePromisesWithCutoff

nflaig:
**Motivation**

Prevent unhandled promise rejections in `racePromisesWithCutoff`, as reported [in discord](https://discord.com/channels/593655374469660673/743858262864167062/1117247288759304282)

```
10:52:23.013[]                error: uncaughtException: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined".
UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined". This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined".
UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined"
```

and already previously reported [in discord](https://discord.com/channels/593655374469660673/593655641445367808/1099556361471283281)

```
validator_1    | UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined". This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined".
validator_1    | UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined".
```

**Description**

Resolves `cutoffPromise` and `timeoutPromise` promises when timeout is reached instead of rejecting and throwing an error.

This prevents unhandled rejections due to `timeoutPromise` not being handled if we return earlier
https://github.com/ChainSafe/lodestar/blob/f2458491b96a8e3177568928268d5499fa5e20fa/packages/utils/src/promise.ts#L120

we never handle the `timeoutPromise` here
https://github.com/ChainSafe/lodestar/blob/f2458491b96a8e3177568928268d5499fa5e20fa/packages/utils/src/promise.ts#L130

**Related issues**
- https://github.com/ChainSafe/lodestar/issues/5404

### DIFF
--- a/packages/utils/src/promise.ts
+++ b/packages/utils/src/promise.ts
@@ -80,11 +80,11 @@ export async function racePromisesWithCutoff<T>(
 ): Promise<(Error | T)[]> {
   // start the cutoff and timeout timers
   let cutoffObserved = false;
-  const cutoffPromise = new Promise((resolve) => setTimeout(resolve, cutoffMs)).then(() => {
+  const cutoffPromise = sleep(cutoffMs).then(() => {
     cutoffObserved = true;
   });
   let timeoutObserved = false;
-  const timeoutPromise = new Promise((resolve) => setTimeout(resolve, timeoutMs)).then(() => {
+  const timeoutPromise = sleep(timeoutMs).then(() => {
     timeoutObserved = true;
   });
   const startTime = Date.now();
--- a/packages/utils/src/promise.ts
+++ b/packages/utils/src/promise.ts
@@ -80,14 +80,12 @@ export async function racePromisesWithCutoff<T>(
 ): Promise<(Error | T)[]> {
   // start the cutoff and timeout timers
   let cutoffObserved = false;
-  const cutoffPromise = new Promise((_resolve, reject) => setTimeout(reject, cutoffMs)).catch((e) => {
+  const cutoffPromise = new Promise((resolve) => setTimeout(resolve, cutoffMs)).then(() => {
     cutoffObserved = true;
-    throw e;
   });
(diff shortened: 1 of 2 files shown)


g11tech (dismissed):
lgtm

packages/utils/src/promise.ts
```diff
@@ -80,14 +80,12 @@ export async function racePromisesWithCutoff<T>(
 ): Promise<(Error | T)[]> {
   // start the cutoff and timeout timers
   let cutoffObserved = false;
-  const cutoffPromise = new Promise((_resolve, reject) => setTimeout(reject, cutoffMs)).catch((e) => {
+  const cutoffPromise = new Promise((resolve) => setTimeout(resolve, cutoffMs)).then(() => {
```
nflaig: @g11tech  wanted keep the diff as minimal as possible but we could also just do now, maybe makes the code more readble

```
 const cutoffPromise = sleep(cutoffMs).then(() => {
    cutoffObserved = true;
  });
```

wemeetagain: Yeah, I think thats more readable

wemeetagain (approved):

wemeetagain:

PR #5633, opened 2023-06-13
feat(beacon-node): add napi blst for attestations and aggregateWithProof

matthewkeil:
**NOTE**: Most of the diff is the `blst` package and can be ignored.  PR only covers `pacakges/beacon-node` changes.

A couple of attempts were made to bring in as a bundle with yarn but all failed.
- `GitPkg` was unable to accommodate using the root of the `bls` repo (swig bls dependency is needed for existing code) along with a subfolder (napi dependency) as separate builds.  
- NPM suffered the same issues attempting to use a git import instead of the normal semver import.  

Once the POC is validated and published the blst-ts `package` can be removed and installed normally.

**Motivation**

The @chainsafe/blst library was rebuilt with Napi to put bls work on the libuv thread pool in-lieu of worker threads

**Description**

- Swap libuv verification for worker thread verification of attestations and aggregatesWithProofs coming from gossip.
- Add metrics for libuv verification process
- Add memory test to understand footprint of napi publicKeys


matthewkeil:
Closing for now to stop email triggers.  Will reopen once ready

PR #5636, opened 2023-06-13
feat: make few improvements to the prover

nazarhussain:
**Motivation**

Make the `@lodestar/prover` useful for production.

**Description**

- Utilize batch request to improve performance 
- Expose api object on the root provider
- Verify the execution compatibility on start
- Add support for the timeout to the proxy

Partially #4706

- Run all tests


packages/prover/scripts/generate_fixtures.ts
```diff
@@ -105,9 +114,9 @@ async function generateFixture(label: string, generator: Generator, network: NET
   };
 
   const dependentRequests = [];
-  for (const {method, params} of opts.dependentRequests ?? []) {
-    const {payload, response} = await rawEth(network, method, params);
-    if (response.error || !response.result) {
+  for (const req of opts.dependentRequests ?? []) {
```
nflaig, line 117: why is there `opts.dependentRequests` and `const dependentRequests = [];`, are those the same thing?

packages/prover/src/cli/cmds/start/options.ts
```diff
@@ -41,6 +44,13 @@ export const startOptions: CliCommandOptions<StartArgs> = {
     group: "execution",
   },
 
+  requestTimeout: {
+    description: `Number of ms to wait for a response from the execution node. Default is ${DEFAULT_PROXY_REQUEST_TIMEOUT}.`,
+    type: "number",
+    demandOption: false,
```
nflaig, line 51: since it is not hidden, could set `defaultDescription` property

packages/prover/src/utils/conversion.ts
```diff
@@ -95,3 +95,8 @@ export function cleanObject<T extends Record<string, unknown> | unknown[]>(obj:
 
   return obj;
 }
+
+export function chunkIntoN<T extends unknown[]>(arr: T, n: number): T[] {
```
nflaig, line 105: would be great to have tsdoc for this function and add a tldr of what it does

packages/prover/src/utils/evm.ts
```diff
@@ -67,35 +72,55 @@ export async function getVMWithState({
     storageKeysMap[address] = storageKeys;
   }
 
-  if (storageKeysMap[tx.from ?? ZERO_ADDRESS] === undefined) {
-    storageKeysMap[tx.from ?? ZERO_ADDRESS] = [];
+  // If from address is not present then we have to fetch it with for all keys
```
nflaig: either

```suggestion
  // If from address is not present then we have to fetch it for all keys
```

or 

```suggestion
  // If from address is not present then we have to fetch it with all keys
```

there is another similar case below

packages/prover/src/constants.ts
```diff
@@ -3,3 +3,4 @@ export const MAX_REQUEST_LIGHT_CLIENT_UPDATES = 128;
 export const MAX_PAYLOAD_HISTORY = 32;
 export const UNVERIFIED_RESPONSE_CODE = -33091;
 export const ZERO_ADDRESS = "0x0000000000000000000000000000000000000000";
+export const DEFAULT_PROXY_REQUEST_TIMEOUT = 3000;
```
nflaig, line 6: 3 seconds seems kinda low, is that really sufficient for slow networks?

packages/prover/src/utils/rpc.ts
```diff
@@ -0,0 +1,105 @@
...
+    return mergeBatchReqResp(payloads, response as JsonRpcBatchResponse) as E extends false
+      ? {request: JsonRpcRequest; response: JsonRpcResponse}[]
+      : {request: JsonRpcRequest; response: JsonRpcResponseWithResultPayload<unknown>}[];
+  }
+
+  async verifyCompatibility(): Promise<void> {
+    try {
+      await this.request("eth_getProof", [ZERO_ADDRESS, [], "latest"], {raiseError: true});
+    } catch (err) {
+      this.logger.error("Execution compatibility failed.", undefined, err as Error);
```
nflaig, line 96: usually just pass an empty object if there is no context but I guess either way is fine

packages/prover/test/unit/web3_provider.test.ts
```diff
@@ -1,14 +1,25 @@
...
 import {createVerifiedExecutionProvider} from "../../src/web3_provider.js";
+import {ELRpc} from "../../src/utils/rpc.js";
 
 describe("web3_provider", () => {
+  const sandbox = sinon.createSandbox();
+
+  beforeEach(() => {
+    sandbox.restore();
+  });
+
```
nflaig, line 16: might be good to also make sure stubs are cleaned up after all tests are executed

```
after(() => {
    sandbox.restore();
});
```

packages/prover/src/constants.ts
nazarhussain: Our proxy is just forwarding the request to an eth node. e.g. Geth. I think 3s is good for an RPC all for that use case.

nflaig: Sure, can always adjust later if it is deemed to be too low

packages/prover/scripts/generate_fixtures.ts
nazarhussain: `opts.dependentRequests` contains only partial requests while `dependentRequests` contains full `payload` and `response`.

packages/prover/src/cli/cmds/start/options.ts
nazarhussain: Used `default` instead as don't need special string parsing for that value.

nflaig (approved):
LGTM!

wemeetagain:

PR #5637, opened 2023-06-13
fix: fix big decimal fomatting of zeros after '.'

g11tech:
fixes a bug where we missed putting zeros after decimal for logging block rewards: for.e.g

`1037977392756968` WEI should be printed as `0.00103` ETH upto 5 decimal places.
This PR makes the fix and add tests cases for the same


nflaig (dismissed):
lgtm

packages/validator/src/util/format.ts
```diff
@@ -5,5 +5,8 @@ export function isValidatePubkeyHex(pubkeyHex: string): boolean {
 export function formatBigDecimal(numerator: bigint, denominator: bigint, maxDecimalFactor: bigint): string {
   const full = numerator / denominator;
   const fraction = ((numerator - full * denominator) * maxDecimalFactor) / denominator;
-  return `${full}.${fraction}`;
+
+  // zeros to be added post decimal are number of zeros in maxDecimalFactor - number of digits in fraction
+  const zerosPostDecimal = `${maxDecimalFactor}`.length - 1 - `${fraction}`.length;
```
wemeetagain:
```suggestion
  const zerosPostDecimal = String(maxDecimalFactor).length - 1 - String(fraction).length;
```

I think this is clearer.

wemeetagain (approved):

wemeetagain:

PR #5639, opened 2023-06-13
style: fix unnecessary type casting

nazarhussain:
**Motivation**

Make sure the types are safe and used with best practices. 

**Description**

Enable `unnecessary type casting` linting rule and fix the code. 

Build and run all code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "build:watch": "lerna exec --parallel -- 'yarn run build:watch'",
     "build:ifchanged": "lerna exec -- ../../scripts/build_if_changed.sh",
     "lint": "eslint --color --ext .ts packages/*/src packages/*/test",
+    "lint:fix": "yarn lint --fix",
     "check-build": "lerna run check-build",
     "check-types": "lerna run check-types --no-bail",
     "coverage": "lerna run coverage --no-bail",
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -68,6 +68,9 @@ module.exports = {
     "@typescript-eslint/no-floating-promises": "error",
     "@typescript-eslint/no-non-null-assertion": "error",
     "@typescript-eslint/no-require-imports": "error",
+    // We usually type-cast these standard types because the concerned function accepts any type
+    // and we want to TS detect error if original variable type changes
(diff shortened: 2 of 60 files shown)


packages/beacon-node/test/spec/presets/light_client/sync.ts
```diff
@@ -124,7 +124,7 @@ export const sync: TestRunnerFn<SyncTestCase, void> = (fork) => {
       for (const [i, step] of testcase.steps.entries()) {
         try {
           if (isProcessUpdateStep(step)) {
-            const currentSlot = Number(step.process_update.current_slot as bigint);
```
nazarhussain: This was intended in case we change the type for `current_slot` because `Number` accepts any type. If anyone have idea about it's workaround do share.

package.json
```diff
@@ -15,6 +15,7 @@
     "build:watch": "lerna exec --parallel -- 'yarn run build:watch'",
     "build:ifchanged": "lerna exec -- ../../scripts/build_if_changed.sh",
     "lint": "eslint --color --ext .ts packages/*/src packages/*/test",
+    "lint:fix": "eslint --fix --color --ext .ts packages/*/src packages/*/test",
```
dapplion: Would this work?
```suggestion
    "lint:fix": "yarn lint --fix",
```

packages/beacon-node/test/spec/presets/light_client/sync.ts
dapplion: The goal of this cast was to guard against the changes in the `current_slot` type. If we change `current_slot` to say an object in the future, this function will pass the typechecker but produce NaN, breaking at runtime. We could ban the use of the Number constructor and expose a bigintToNumber helper, but that has a runtime cost

dapplion (commented):
I think this rule is great but I'm really concerned by the usage of Number, String, and similar constructors that take any as argument. Can you find a solution that does not reduce the typesafety of `Number(surprise_var as number`?

nazarhussain:
@dapplion I found a workaround by adding `"string", "bigint", "number"` types to ignore list for that lint rule. Please re-review.

dapplion (approved):
LGTM! thanks!

wemeetagain:

PR #5640, opened 2023-06-13
fix: fix the blobs side cars by range response

g11tech:
fix a typo which was in the yielding the response for blob side car bytes which cause all blobs for the block to be passed rather than the extracted blobsidecar bytes.

PR fixes that and also renames variables a bit to avoid confusion


wemeetagain (approved):

wemeetagain:

PR #5641, opened 2023-06-13
test: make results of keymanager e2e tests deterministic

nflaig:
**Motivation**

Keystores are decrypted in parallel now (https://github.com/ChainSafe/lodestar/pull/5624) which means the order in which they are added to the validator store  (`validatorStore.addSigner`) is no longer deterministic.
https://github.com/ChainSafe/lodestar/blob/f07d97f5ff45e64ad45a00d9c829a4f0a899f055/packages/cli/src/cmds/validator/keymanager/impl.ts#L166

This makes the result of keymanager e2e non-deterministic and causes e2e tests to fail randomly.

image

However, the order of signers in validator store does not matter which means the assertion can be updated to check the response in an unordered way.

**Description**

This change makes results of keymanager e2e tests deterministic, we only need to check the presence of all keys in response, irrespective of their order.


wemeetagain (approved):

g11tech (commented):
looks great

wemeetagain:

PR #5628, opened 2023-06-12
feat: persist network thread profile

twoeths:
**Motivation**

- We want to take profile of network thread without promise tracking

**Description**

- Use `node:inspector` module to persist network thread profile
- 
part of #5604

PR #5643, opened 2023-06-14
fix!: match beacon APIs URL casing

dapplion:
**Motivation**

lodestar API routes, should match beacon APIs URL casing

- See https://github.com/ChainSafe/lodestar/pull/5628#discussion_r1226187456

**Description**

Set all existing routes casing to snake case. This is a breaking change but I'm not aware of us explicitly guaranteeing backwards compatibility with them. Maybe it should be clear somewhere in another PR that those are dev routes


wemeetagain (approved):

wemeetagain:

PR #5654, opened 2023-06-16
fix: unknown block sync to subscribe/unsubscribe to network events

### LINKED ISSUE #5653: Only enable unknown block sync after block is synced
### Describe the bug

As found in https://github.com/ChainSafe/lodestar/issues/5650#issuecomment-1592822730, when node is syncing it may also receive unknown block root event (especially in a setup with fallback urls), there is no need to search for block at that time

also this is to avoid this case https://github.com/ChainSafe/lodestar/blob/2eddf466394680cb1c6a24700b9a3a2958a3ef90/packages/beacon-node/src/sync/unknownBlock.ts#L228

### Expected behavior

Only start/init unknown block sync once the node is synced

### Steps to reproduce

_No response_

### Additional context

_No response_

### Operating system

Linux

v1.9.0-rc.2

twoeths:
**Motivation**

Unknown block sync received unknown block roots while the node is syncing, this lead to:
- Pending blocks may be filled up quickly
- Incorrect catch of this case https://github.com/ChainSafe/lodestar/blob/2eddf466394680cb1c6a24700b9a3a2958a3ef90/packages/beacon-node/src/sync/unknownBlock.ts#L228

**Description**

- Add subscribe/unsubscribe methods to UnknownBlock sync
- Only subscribe once the node is synced, similar to `network.subscribeGossipCoreTopics()`

Closes #5653

### DIFF
--- a/packages/beacon-node/src/sync/sync.ts
+++ b/packages/beacon-node/src/sync/sync.ts
@@ -196,36 +196,48 @@ export class BeaconSync implements IBeaconSync {
     const state = this.state; // Don't run the getter twice
 
     // We have become synced, subscribe to all the gossip core topics
-    if (
-      state === SyncState.Synced &&
-      !this.network.isSubscribedToGossipCoreTopics() &&
-      this.chain.clock.currentEpoch >= MIN_EPOCH_TO_START_GOSSIP
-    ) {
-      this.network
-        .subscribeGossipCoreTopics()
-        .then(() => {
-          this.metrics?.syncSwitchGossipSubscriptions.inc({action: "subscribed"});
-          this.logger.info("Subscribed gossip core topics");
-        })
-        .catch((e) => {
-          this.logger.error("Error subscribing to gossip core topics", {}, e);
-        });
+    if (state === SyncState.Synced && this.chain.clock.currentEpoch >= MIN_EPOCH_TO_START_GOSSIP) {
+      if (!this.network.isSubscribedToGossipCoreTopics()) {
+        this.network
+          .subscribeGossipCoreTopics()
+          .then(() => {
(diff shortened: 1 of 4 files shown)


packages/beacon-node/src/sync/sync.ts
```diff
@@ -196,36 +199,48 @@ export class BeaconSync implements IBeaconSync {
...
+        this.network
+          .subscribeGossipCoreTopics()
+          .then(() => {
+            this.metrics?.syncSwitchGossipSubscriptions.inc({action: "subscribed"});
+            this.logger.info("Subscribed gossip core topics");
+          })
+          .catch((e) => {
+            this.logger.error("Error subscribing to gossip core topics", {}, e);
+          });
+      }
```
dapplion, line 213: Can we rebundle this if statement as before? If you need you can re-order the conditions
```js
    if (
      state === SyncState.Synced &&
      this.chain.clock.currentEpoch >= MIN_EPOCH_TO_START_GOSSIP &&
      !this.network.isSubscribedToGossipCoreTopics()
    ) {
```

dapplion (commented):
I can't think of any immediate issues with this, let's merge it. Added a minor comment to reduce diff

dapplion (approved):

packages/beacon-node/src/sync/sync.ts
twoeths: yes we can do it but then the below `this.unknownBlockSync.subscribeToNetwork();` has to depend on `this.network.isSubscribedToGossipCoreTopics()`. Logically it's the same but it'll make it hard to reason about later.

wemeetagain:

PR #5660, opened 2023-06-16
fix(cli): consistently support URLs as comma-separated values

nflaig:
**Motivation**

As discussed in https://github.com/ChainSafe/lodestar/pull/4576#discussion_r1232086829 we should support providing URLs as comma-separated values more consistently. This is also the most intuitive pattern to provide multiple values.

**Description**

We already support this for `--beaconNodes`
https://github.com/ChainSafe/lodestar/blob/81a6da0a7b93556d6cc62dcc9389625c08d17dcb/packages/cli/src/cmds/validator/options.ts#L165-L172

This PR just applies this pattern to CLI flag that support providing multiple URLs.

Opted to just copy-paste, extracting it into reusable function could make sense but found it to be less intuitive.


dapplion (approved):

wemeetagain:

Issue #5604, opened 2023-06-02
Network worker extremely busy / high event loop lag / high I/O lag


dapplion:
### Problem description

General issue to recap observations of enabling the network worker on high load nodes.

On June 25th network worker is enabled on Goerli on the unstable group nodes. The lg1k node with 1000 validators connected, shows bad network performance with long response times and other metrics indicating very slow network handling.

On the week of June 25th we deploy network worker on feat1 group nodes, turning the network worker on and off (moving network stack to the main thread). The metrics below capture

- after enabling the worker the main thread got less busy (green line decreasing)
- network worker is extremely busy, causing bad network behaviour

_Below, active handles in pink = network worker on_
_Note there's an unrelated memory leak going on that's being addressed in another issue_
_feat1-mainnet node running on mainnet with --subscribe-all-subnets, no validators attached, version v1.8.0/888c599, timeframe: 5/24 till 5/30_

![image1](https://github.com/ChainSafe/lodestar/assets/35266934/f03cfe75-3e83-4685-8a8b-832dc92acac0)
![image2](https://github.com/ChainSafe/lodestar/assets/35266934/f8592c3a-7aa4-4d30-9090-11d8245ff25e)
![image3](https://github.com/ChainSafe/lodestar/assets/35266934/79582127-7db6-42e9-a64b-506c852af54f)

The biggest difference between using the network worker is the huge increase in outbound network traffic. Outbound gossip traffic jumped from 0.3MB/s to 2MB/s (~6x). Before this node dropped 50% of incoming attestations, now process everything relatively timely.

_feat1-mainnet node running on mainnet with --subscribe-all-subnets, no validators attached, version v1.8.0/888c599, timeframe: 5/27 till 6/02_

![image](https://github.com/ChainSafe/lodestar/assets/35266934/8c1fde3c-5e58-4b8d-a3a6-899070bfe129)

confirmed with system metrics

![image](https://github.com/ChainSafe/lodestar/assets/35266934/b4867e50-6342-487c-bb3d-8ac8f8c263b1)

**Questions**

- What is the network worker doing that chokes so much?
- Why is the overall load of the node up so much?
- Is there any overhead of having most OS I/O interactions in a worker?

### Solution description

--

matthewkeil:
@dapplion I have come up with two strategies to poll the worker to generate profiles.

[...]

Issue #5556, opened 2023-05-26
v1.9.0-rc.0: long incoming req/resp handler in network thread


twoeths:
### Describe the bug

Long incoming request handler for `beacon_blocks_by_root, ping and status`

<img width="1300" alt="Screenshot 2023-05-26 at 19 38 10" src="https://github.com/ChainSafe/lodestar/assets/10568965/4fd5c5cc-015d-435e-8873-ca49fa70b37b">

### Expected behavior

same incoming request handler like in `stable`

### Operating system

Linux

### Lodestar version or commit hash

v1.9.0-rc.0

twoeths:
I added below log to this line https://github.com/ChainSafe/lodestar/blob/84903697c8fedc96d876aeb3a6631343482d4814/packages/reqresp/src/response/index.ts#L60

```
logger.debug("Req  received before decoding request", logCtx);
```

it logged a delay in `requestDecode()` function or I/O lag

it took 1s to receive and decode a ping request
```
May-30 09:39:14.734[network]         debug: Req  received before decoding request method=ping, client=Lighthouse, peer=16...8DShj2, requestId=5273
May-30 09:39:15.670[network]         debug: Req  received method=ping, client=Lighthouse, peer=16...8DShj2, requestId=5273
```

it took 3s to receive and decode a `beacon_blocks_by_range` request
```
May-30 09:44:39.768[network]         debug: Req  received before decoding request method=beacon_blocks_by_range, client=Lighthouse, peer=16...kY2mne, requestId=7098
May-30 09:44:43.069[network]         debug: Req  received method=beacon_blocks_by_range, client=Lighthouse, peer=16...kY2mne, requestId=7098
```

twoeths:
@dapplion found I/O on the network thread to be super lag

<img width="1292" alt="Screenshot 2023-06-02 at 05 16 32" src="https://github.com/ChainSafe/lodestar/assets/10568965/a1cd78c1-c32b-4145-b04b-47d1319fa19a">

twoeths:
issue was resolved, this is on a mainnet node with `useWorker = true`

<img width="1294" alt="Screenshot 2023-08-20 at 13 15 34" src="https://github.com/ChainSafe/lodestar/assets/10568965/092ee7c5-4c0d-40cd-9e56-f0ea23ba3f1d">

it's even better than `unstable` mainnet node where `useWorker=false`

[...]

PR #5664, opened 2023-06-18
fix: gossipsub to yield more to the macro queue

twoeths:
**Motivation**

- Network event loop is super lag, see #5604

**Description**

- Apply the same code change in v1.8.0, see https://github.com/ChainSafe/lodestar/pull/5316
- Although it does not completely resolve the below issue, it helps a lot

part of #5556
part of #5604

**Test result on mainnet node**

Screenshot 2023-06-18 at 15 29 36

the 1st event loop lag metric is not a big difference , but it's still better than unstable (from 12h)

Screenshot 2023-06-18 at 15 29 20

the 2nd event loop lag is clearly better

Screenshot 2023-06-18 at 15 35 49

especially the incoming request handler time

Screenshot 2023-06-18 at 15 30 59

also outgoing request time
Screenshot 2023-06-18 at 15 31 47

it helps our peers more stable, more peers want to stay with us
Screenshot 2023-06-18 at 15 33 06

### DIFF
--- a/packages/beacon-node/src/network/gossip/gossipsub.ts
+++ b/packages/beacon-node/src/network/gossip/gossipsub.ts
@@ -280,20 +280,25 @@ export class Eth2Gossipsub extends GossipSub {
     const seenTimestampSec = Date.now() / 1000;
 
     // Emit message to network processor
-    this.events.emit(NetworkEvent.pendingGossipsubMessage, {
-      topic,
-      msg,
-      msgId,
-      // Hot path, use cached .toString() version
-      propagationSource: propagationSource.toString(),
-      seenTimestampSec,
-      startProcessUnixSec: null,
-    });
+    setTimeout(() => {
+      this.events.emit(NetworkEvent.pendingGossipsubMessage, {
+        topic,
+        msg,
+        msgId,
+        // Hot path, use cached .toString() version
+        propagationSource: propagationSource.toString(),
+        seenTimestampSec,
+        startProcessUnixSec: null,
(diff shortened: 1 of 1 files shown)


dapplion (approved):

twoeths:
attaching the profile of network thread of this branch

[network_thread_2023-06-19T08:39:44.576Z.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/11786215/network_thread_2023-06-19T08.39.44.576Z.cpuprofile.zip)

there are still some `runMicroTasks` call takes up to 400ms (but previously I can found one with 1.1s)

Screenshot 2023-06-19 at 15 52 13

wemeetagain:
:tada: This PR is included in [v1.9.1](https://github.com/ChainSafe/lodestar/releases/tag/v1.9.1) :tada:

nflaig:
On the topic of `setImmediate` vs `setTimeout(0)`, I have gathered metrics on this (changes in this branch https://github.com/ChainSafe/lodestar/commit/d674bd1afd254bfa288ee62b9ca3cf969aefd8ba) and there is no noticeable improvement.

I think this is due to the fact that we yield to the event loop after I/O polling most of the time so it does not really make a difference. In fact, I think `setTimeout(0)` might even be better because this queues those tasks to be executed after already pending timers.

There might be some potential if we only use `setImmediate` in few specific places but that will likely be hard to measure.

cc @matthewkeil

wemeetagain:

PR #5667, opened 2023-06-18
fix!: update validator liveness API to be spec compliant

nflaig:
**Motivation**

Current implementation of validator liveness API is not spec compliant, see  [getLiveness](https://ethereum.github.io/beacon-APIs/#/Validator/getLiveness).

The API was initially introduced in
- https://github.com/ethereum/beacon-APIs/pull/131

and further updated here
- https://github.com/ethereum/beacon-APIs/pull/253

**Description**

Update validator liveness API to conform with latest beacon API spec ([liveness.yaml](https://github.com/ethereum/beacon-APIs/blob/a62aa810fb283873de0d8025405f7a8ce6991f51/apis/validator/liveness.yaml))

- use POST instead of GET
- remove epoch from response data
- pass epoch as path param instead of query
- pass validator indices in request body instead of query
- update URL to `/eth/v1/validator/liveness/{epoch}`
- throw `ApiError` instead of normal `Error` to improve error handling/logging

This change is breaking as it introduces an incompatibility with older Lodestar versions but due to the fact that our implementation was never spec compliant I doubt it will cause any issues with other downstream tooling.

### DIFF
--- a/packages/validator/src/services/doppelgangerService.ts
+++ b/packages/validator/src/services/doppelgangerService.ts
@@ -1,5 +1,5 @@
 import {Epoch, ValidatorIndex} from "@lodestar/types";
-import {Api, ApiError} from "@lodestar/api";
+import {Api, ApiError, routes} from "@lodestar/api";
 import {Logger, sleep} from "@lodestar/utils";
 import {computeStartSlotAtEpoch} from "@lodestar/state-transition";
 import {ProcessShutdownCallback, PubkeyHex} from "../types.js";
@@ -12,10 +12,10 @@ import {IndicesService} from "./indices.js";
 const DEFAULT_REMAINING_DETECTION_EPOCHS = 1;
 const REMAINING_EPOCHS_IF_DOPPLEGANGER = Infinity;
 
-export type LivenessResponseData = {
-  index: ValidatorIndex;
+/** Liveness responses for a given epoch */
+type EpochLivenessData = {
   epoch: Epoch;
-  isLive: boolean;
+  responses: routes.validator.LivenessResponseData[];
 };
 
 export type DoppelgangerState = {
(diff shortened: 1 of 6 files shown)


dapplion:
@nflaig could you look into why the API spec tests did not catch this?

nflaig:
@dapplion yeah was wondering the same, haven't investigated it yet but will do. It might also be worth to see if we can validate pattern to ensure we are encoding correctly to prevent issues like the graffiti not being hex encoded.

packages/validator/src/services/doppelgangerService.ts
```diff
@@ -209,21 +209,16 @@ export class DoppelgangerService {
       //
       // Do not bother iterating through the current epoch responses since they've already been
       // checked for violators and they don't result in updating the state.
-      for (const response of previousEpochLiveness) {
-        if (response.epoch !== previousEpoch) {
-          // Server sending bad data
-          throw Error(`Inconsistent livenessResponseData epoch ${response.epoch} != ${previousEpoch}`);
```
nflaig, line 215: Epoch is no longer part of response, no need to check this

wemeetagain (dismissed):

wemeetagain (approved):

nflaig:
> could you look into why the API spec tests did not catch this?

@dapplion this was not caught because our oapi spec tests check against v2.3.0 and not the latest spec (v2.4.1)
https://github.com/ChainSafe/lodestar/blob/50daa01c80b22136e70b7cfb5ab66c735f9641a8/packages/api/test/unit/beacon/oapiSpec.test.ts#L26

> It might also be worth to see if we can validate pattern to ensure we are encoding correctly

This seems to be validated already
https://github.com/ChainSafe/lodestar/blob/50daa01c80b22136e70b7cfb5ab66c735f9641a8/packages/api/test/utils/checkAgainstSpec.ts#L20

wemeetagain:

PR #5668, opened 2023-06-19
chore: track block source in lodestar summary dashboard

twoeths:
**Motivation**

- If blocks come from either range sync or unknown block sync that's an issue to our node

**Description**

- Track block source in lodestar dashboard as a new panel
- Change the `uid` of `lodestar_summary.json` to be `lodestar_summary` so it works with `download_dashboards.mjs` script (see https://github.com/ChainSafe/lodestar/blob/9a8a6f4824fdb2eb36a06fc237ce8663bad412b8/scripts/download_dashboards.mjs#L65). After this is merged we should reimport this dashboard to grafana and remove the old uid dashboard

Screenshot 2023-06-19 at 13 33 12

### DIFF
--- a/dashboards/lodestar_summary.json
+++ b/dashboards/lodestar_summary.json
@@ -1,26 +1,12 @@
 {
   "__inputs": [
     {
-      "name": "DS_PROMETHEUS",
-      "label": "Prometheus",
       "description": "",
-      "type": "datasource",
+      "label": "Prometheus",
+      "name": "DS_PROMETHEUS",
       "pluginId": "prometheus",
-      "pluginName": "Prometheus"
-    },
-    {
-      "name": "VAR_BEACON_JOB",
-      "type": "constant",
-      "label": "Beacon node job name",
-      "value": "beacon",
-      "description": ""
-    },
-    {
-      "name": "VAR_VALIDATOR_JOB",
-      "type": "constant",
-      "label": "Validator client job name",
-      "value": "validator",
-      "description": ""
+      "pluginName": "Prometheus",
+      "type": "datasource"
     }
   ],
   "annotations": {
@@ -1743,7 +1729,7 @@
               }
             ]
           },
-          "unit": "s"
+          "unit": "percentunit"
         },
         "overrides": []
       },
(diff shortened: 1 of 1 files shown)


dashboards/lodestar_summary.json
```diff
@@ -3219,7 +3372,7 @@
   },
   "timezone": "utc",
   "title": "Lodestar",
-  "uid": "lodestar",
+  "uid": "lodestar_summary",
```
twoeths, line 3389: `uid` needs to map to the file name

dashboards/lodestar_summary.json
```diff
@@ -3156,46 +3323,32 @@
...
-          "text": "${VAR_VALIDATOR_JOB}",
-          "selected": false
-        },
-        "options": [
-          {
-            "value": "${VAR_VALIDATOR_JOB}",
-            "text": "${VAR_VALIDATOR_JOB}",
-            "selected": false
-          }
-        ]
```
dapplion, line 3198: @nflaig can you take a look?

nflaig: The change here seems fine, not sure why it was updated but dashboard variables still work, just need to re-add them to `__inputs`

dashboards/lodestar_summary.json
```diff
@@ -1,26 +1,12 @@
...
-      "label": "Prometheus",
       "description": "",
-      "type": "datasource",
+      "label": "Prometheus",
+      "name": "DS_PROMETHEUS",
       "pluginId": "prometheus",
-      "pluginName": "Prometheus"
-    },
-    {
-      "name": "VAR_BEACON_JOB",
```
nflaig, line 12: @tuyennhv any idea when/how the variables were updated? The script should not modify `__inputs` if those are already defined

nflaig: > So the fields starting with __ must be set manually, see https://community.grafana.com/t/export-dashboard-for-external-use-via-http-api/50716

As Lion mentioned here, looks like it happens on download. Will update the script to add job inputs if dashboard uses them.

nflaig (approved):

dashboards/lodestar_summary.json
nflaig: Issue will be fixed by
- https://github.com/ChainSafe/lodestar/pull/5672

twoeths: thanks for the fix @nflaig , it could be because I did some hack in order to download this dashboard

wemeetagain:

PR #5669, opened 2023-06-19
feat: allow builder only proposal flow for benefit of DVT protocols

g11tech:
DVT protocols block proposal may depend on sourcing block for proposal only through relayer so that the distributed validators are signing off on same block for proposal.

This PR allows for the same via `--builder.selection=builderonly` option

### DIFF
--- a/packages/cli/src/cmds/validator/handler.ts
+++ b/packages/cli/src/cmds/validator/handler.ts
@@ -254,6 +254,8 @@ function parseBuilderSelection(builderSelection?: string): BuilderSelection | un
         break;
       case "builderalways":
         break;
+      case "builderonly":
+        break;
       default:
         throw Error("Invalid input for builder selection, check help.");
     }
--- a/packages/cli/src/cmds/validator/options.ts
+++ b/packages/cli/src/cmds/validator/options.ts
@@ -237,7 +237,7 @@ export const validatorOptions: CliCommandOptions<IValidatorCliArgs> = {
 
   "builder.selection": {
     type: "string",
-    description: "Default builder block selection strategy: maxprofit or builderalways",
+    description: "Default builder block selection strategy: maxprofit or builderalways or builderonly",
     defaultDescription: `${defaultOptions.builderSelection}`,
     group: "builder",
   },
--- a/packages/cli/src/cmds/validator/options.ts
+++ b/packages/cli/src/cmds/validator/options.ts
@@ -237,7 +237,7 @@ export const validatorOptions: CliCommandOptions<IValidatorCliArgs> = {
 
   "builder.selection": {
     type: "string",
-    description: "Default builder block selection strategy: maxprofit or builderalways or builderonly",
+    description: "Default builder block selection strategy: maxprofit, builderalways, or builderonly",
     defaultDescription: `${defaultOptions.builderSelection}`,
     group: "builder",
   },
--- a/packages/validator/src/services/block.ts
+++ b/packages/validator/src/services/block.ts
@@ -240,7 +240,7 @@ export class BlockProposingService {
       fullBlock = await fullBlockPromise;
     } else {
       throw Error(
-        `Neither builder not execution proposal flow activated: isBuilderEnabled=${isBuilderEnabled} builderSelection=${builderSelection}`
+        `Neither builder nor execution proposal flow activated: isBuilderEnabled=${isBuilderEnabled} builderSelection=${builderSelection}`
       );
     }
     this.logger.info("Started block production", {});
--- a/packages/validator/src/services/block.ts
+++ b/packages/validator/src/services/block.ts
@@ -197,7 +197,16 @@ export class BlockProposingService {
     // Start calls for building execution and builder blocks
     const blindedBlockPromise = isBuilderEnabled ? this.produceBlindedBlock(slot, randaoReveal, graffiti) : null;
     const fullBlockPromise =
-      builderSelection !== BuilderSelection.BuilderOnly ? this.produceBlock(slot, randaoReveal, graffiti) : null;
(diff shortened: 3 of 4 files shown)


packages/cli/src/cmds/validator/handler.ts
nflaig, line 1: CLI description should probably be updated accordingly, also I think once we add `builderonly` there is should be clear that `builderalways` does not exclusively only produce blinded blocks.

https://github.com/ChainSafe/lodestar/blob/ef64e04389eaa989b01efbefd1c822c7af9bf1da/packages/cli/src/cmds/validator/options.ts#L238-L240

Could potentially rename the option to `builderprefer` (or similar) but likely not worth it and would be a breaking change.

packages/validator/src/services/block.ts
```diff
@@ -225,10 +232,18 @@ export class BlockProposingService {
...
-      fullBlock = await fullBlockPromise;
+    } else if (blindedBlockPromise !== null && fullBlockPromise === null) {
+      blindedBlock = await blindedBlockPromise;
+      fullBlock = null;
+    } else if (blindedBlockPromise === null && fullBlockPromise !== null) {
       blindedBlock = null;
+      fullBlock = await fullBlockPromise;
+    } else {
+      throw Error(
+        `Neither builder nor execution proposal flow activated: isBuilderEnabled=${isBuilderEnabled} builderSelection=${builderSelection}`
```
nflaig: In my opinion this error condition should not be possible as this error would only popup at the time a block proposal has to be be done but at this point it is too late and the block will be missed.

I would suggest to either
- throw error on startup if `isBuilderEnabled=false` and `builderSelection=builderonly` or
- just ignore builder selection if builder is disabled, `builder.selection` is option of builder so from a user perspective this should be expected to not be handled in that case

packages/validator/src/services/block.ts
```diff
@@ -190,10 +196,11 @@ export class BlockProposingService {
   > => {
     // Start calls for building execution and builder blocks
     const blindedBlockPromise = isBuilderEnabled ? this.produceBlindedBlock(slot, randaoReveal, graffiti) : null;
-    const fullBlockPromise = this.produceBlock(slot, randaoReveal, graffiti);
+    const fullBlockPromise =
+      builderSelection !== BuilderSelection.BuilderOnly ? this.produceBlock(slot, randaoReveal, graffiti) : null;
```
nflaig: as mentioned in other comment, should probably just ignore `builderSelection ` if `isBuilderEnabled=false`

packages/validator/src/services/block.ts
```diff
@@ -225,10 +232,18 @@ export class BlockProposingService {
...
+      fullBlock = null;
+    } else if (blindedBlockPromise === null && fullBlockPromise !== null) {
       blindedBlock = null;
+      fullBlock = await fullBlockPromise;
+    } else {
+      throw Error(
+        `Neither builder nor execution proposal flow activated: isBuilderEnabled=${isBuilderEnabled} builderSelection=${builderSelection}`
+      );
     }
+    this.logger.info("Started block production", {});
```
nflaig: Let's also add slot to the log, it ensures it shows up if you query for slot specific logs

```suggestion
    this.logger.info("Started block production", {slot});
```

g11tech: oops sorry this log is not necessary as we have a debug log to which i added the relevant data

packages/validator/src/services/block.ts
g11tech: throwing on startup makes sense, this is to just cover all cases for typescript

UPDATE
Actually there is a problem : we allow  `builderSelection` to be configured via proposerSettingFile or also via keymanager api, so throwing on startup is not a viable solution, better would be to ignore builder selection if builder is disabled

packages/validator/src/services/block.ts
nflaig: ah true, this log would also be incorrect as it is logged after all promises are already awaited

packages/validator/src/services/block.ts
nflaig: Ignoring it was also my first thought when I looked at the code and how we could implement this. Also what I suggested here https://github.com/ChainSafe/lodestar/pull/5669#discussion_r1233955364,

packages/validator/src/services/block.ts
g11tech: yes this seems to be the best startegy

packages/cli/src/cmds/validator/handler.ts
g11tech: done, thanks for bringing to notice :heart:

packages/cli/src/cmds/validator/options.ts
```diff
@@ -237,7 +237,7 @@ export const validatorOptions: CliCommandOptions<IValidatorCliArgs> = {
 
   "builder.selection": {
     type: "string",
-    description: "Default builder block selection strategy: maxprofit or builderalways",
+    description: "Default builder block selection strategy: maxprofit or builderalways or builderonly",
```
nflaig: Might be easier to read like this
```suggestion
    description: "Default builder block selection strategy: maxprofit, builderalways, or builderonly",
```

packages/cli/src/cmds/validator/handler.ts
nflaig: I think naming of options is good now, it should be clear to a user that the right option to set  if they only want builder block is `builderonly` and not `builderalways`

nflaig (approved):

wemeetagain:
:tada: This PR is included in [v1.9.2](https://github.com/ChainSafe/lodestar/releases/tag/v1.9.2) :tada:

wemeetagain:

PR #5677, opened 2023-06-21
chore!: update prometheus-gc-stats dependency

wemeetagain:
**Motivation**

When updating to node v20, the `gc-stats` package was found to be out of date and incompatible with node 20. This dependency is a sub-dependency of `prometheus-gc-stats`, which we use to gather information about gc runs.

While investigating, I found that the information from `gc-stats` is now available natively from the `node:v8` package (as of node v18.15.0).

Seeing as `prometheus-gc-stats` was not well maintained, and `gc-stats` is completely abandoned, I took this opportunity to simplify the supply chain and use the native functionality from `node:v8` in a new [`@chainsafe/prometheus-gc-stats` package](https://github.com/ChainSafe/node-prometheus-gc-stats) (pull requests and a review welcome).

**Description**

- Use `@chainsafe/prometheus-gc-stats` instead of `prometheus-gc-stats`.
- Calling `gcStats(...)` now requires cleanup by calling the returned function.
- Always use the latest lts in our workflows
- Explicitly state the minimum node version (v18.15.x) in the README

### DIFF
--- a/packages/beacon-node/src/metrics/nodeJsMetrics.ts
+++ b/packages/beacon-node/src/metrics/nodeJsMetrics.ts
@@ -1,7 +1,7 @@
 import {collectDefaultMetrics, Registry} from "prom-client";
-import gcStats from "prometheus-gc-stats";
+import {gcStats} from "@chainsafe/prometheus-gc-stats";
 
-export function collectNodeJSMetrics(register: Registry, prefix?: string): void {
+export function collectNodeJSMetrics(register: Registry, prefix?: string): () => void {
   collectDefaultMetrics({
     register,
     prefix,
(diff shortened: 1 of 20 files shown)


twoeths (approved):

packages/beacon-node/src/metrics/nodeJsMetrics.ts
```diff
@@ -13,5 +13,7 @@ export function collectNodeJSMetrics(register: Registry, prefix?: string): void
   // - nodejs_gc_runs_total: Counts the number of time GC is invoked
   // - nodejs_gc_pause_seconds_total: Time spent in GC in seconds
   // - nodejs_gc_reclaimed_bytes_total: The number of bytes GC has freed
-  gcStats(register, {prefix})();
+  // `close` must be called to stop the gc collection process from continuing
+  const close = gcStats(register, {collectionInterval: 6000, prefix});
```
nflaig, line 17: why is the interval every 6 seconds? Is it just randomly selected or is there some reason for that specific value?

wemeetagain:

PR #5681, opened 2023-06-22
fix: fix max responses that can be received by blobs side car by range

g11tech:
After quite a decent amount of debugging where lodestar would stall syncing via beacon blocks may be blobs by range on devnet6 from almost all clients on a particular epoch, it was figured out that the other clients were properly sending the response.

So it became apparent that lodestar was chunking out some of the blob data. This PR fixes the same as `count` in the blob side cars by range is `count` of `slots` for which blob data is requested.


nflaig (approved):

wemeetagain:

PR #5689, opened 2023-06-23
refactor: clean up validator import options

nflaig:
**Motivation**

The [validator import docs](https://chainsafe.github.io/lodestar/reference/cli/#validator-import) show all options that are available to validator which does not make sense as the only essential (non-hidden) options are `--importKeystores` and `importKeystoresPassword`.

**Description**

Fix validator import docs to only show relevant options. Validator options and global options are still available as parent args but do not need to be specifically documented for the sub-command.

image


wemeetagain (approved):

wemeetagain:

PR #5691, opened 2023-06-23
chore: use availableParallelism to determine thread pool size

nflaig:
**Motivation**

From [node docs](https://nodejs.org/api/os.html#oscpus)

> os.cpus().length should not be used to calculate the amount of parallelism available to an application. Use os.availableParallelism() for this purpose.

**Description**

Use `os.availableParallelism()` instead of `os.cpus().length` to determine thread pool size.

This is supported from node v18.14.0, see [availableParallelism](https://nodejs.org/api/os.html#osavailableparallelism).


wemeetagain (approved):

wemeetagain:

PR #5719, opened 2023-06-29
chore: fix local testnet dev scripts

nflaig:
**Motivation**

Current dev script do not work but those are quite useful to quickly spin up a [local testnet](https://chainsafe.github.io/lodestar/usage/local/)

**Description**

Adjust dev scripts to work with latest version of CLI.

### DIFF
--- a/scripts/dev/node1.sh
+++ b/scripts/dev/node1.sh
@@ -2,7 +2,7 @@
 
 GENESIS_TIME=$(date +%s)
 
-packages/cli/bin/lodestar dev \
+./lodestar dev \
   --genesisValidators 8 \
   --startValidators 0..7 \
   --genesisTime $GENESIS_TIME \
@@ -13,7 +13,4 @@ packages/cli/bin/lodestar dev \
   --rest.namespace '*' \
   --metrics \
   --logLevel debug \
-  --eth1 false \
-  --network.requestCountPeerLimit 1000000 \
-  --network.blockCountTotalLimit 1000000 \
-  --network.blockCountPeerLimit 1000000
+  --eth1 false
--- a/scripts/dev/node2.sh
+++ b/scripts/dev/node2.sh
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 
 # Fetch node1 data
-ENR=$(curl -s http://localhost:9596/eth/v1/node/identity | jq .data.enr)
-GENESIS_TIME=$(curl -s http://localhost:9596/eth/v1/beacon/genesis | jq .data.genesis_time)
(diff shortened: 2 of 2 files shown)


wemeetagain (approved):

scripts/dev/node1.sh
```diff
@@ -13,7 +13,4 @@ packages/cli/bin/lodestar dev \
   --rest.namespace '*' \
   --metrics \
   --logLevel debug \
-  --eth1 false \
-  --network.requestCountPeerLimit 1000000 \
-  --network.blockCountTotalLimit 1000000 \
-  --network.blockCountPeerLimit 1000000
```
dapplion, line 19: Why remove the rate limit changes?

nflaig: No specific reason other than simplifying the scripts. Since those were added in https://github.com/ChainSafe/lodestar/pull/3858, I just assumed those parameters were specifically used to test this PR.

dapplion: Those set the reqresp limits close to infinity, otherwise dev nodes with ban each other very quickly when syncing

nflaig: Those seem to be deprecated and unused
https://github.com/ChainSafe/lodestar/blob/0f0f533e1ac6163082eb0b7baa7499c4e02ae568/packages/cli/src/options/beaconNodeOptions/network.ts#L30-L31


This PR https://github.com/ChainSafe/lodestar/pull/4828 added a new flag
https://github.com/ChainSafe/lodestar/blob/0f0f533e1ac6163082eb0b7baa7499c4e02ae568/packages/cli/src/options/beaconNodeOptions/network.ts#L250

To disable rate limiting now would have to set `--network.rateLimitMultiplier 0`

nflaig: > dev nodes with ban each other very quickly when syncing

Haven't seen that so far, I guess the nodes are instantly synced since we have `--reset` flag and always start from an empty db.

dapplion: > Haven't seen that so far, I guess the nodes are instantly synced since we have --reset flag and always start from an empty db.

You need to start external nodes after a while such that they need to sync a sufficient number of slots (thousands)

nflaig: Was not aware of that but makes sense, shall we add `--network.rateLimitMultiplier 0` to dev scripts?

It seems to be disabled during testing as well
https://github.com/ChainSafe/lodestar/blob/0f0f533e1ac6163082eb0b7baa7499c4e02ae568/packages/beacon-node/test/utils/network.ts#L127-L128

wemeetagain:

PR #5722, opened 2023-06-30
fix: abort and headers already sent errors for the rest api

### LINKED ISSUE #5706: Validator client unclean disconnect on shutdown
### Describe the bug

When shutting down the VC is causes the following error on the BN

```
Jun-25 02:58:26.174[rest]            error: Req req-5 eventstream error  aborted
Error: aborted
    at connResetException (node:internal/errors:717:14)
    at abortIncoming (node:_http_server:754:17)
    at socketOnClose (node:_http_server:748:3)
    at Socket.emit (node:events:525:35)
    at TCP.<anonymous> (node:net:322:12)
```

In rare cases it might also lead to further error on the BN

```
Jun-25 02:58:26.192[]                error: uncaughtException: Cannot set headers after they are sent to the client
Error: Cannot set headers after they are sent to the client
    at new NodeError (node:internal/errors:399:5)
    at ServerResponse.setHeader (node:_http_outgoing:645:11)
    at ServerResponse.writeHead (node:_http_server:378:21)
    at onSendEnd (/mnt/goerli/goerli/lodestar/node_modules/fastify/lib/reply.js:596:7)
    at onSendHook (/mnt/goerli/goerli/lodestar/node_modules/fastify/lib/reply.js:529:5)
    at fallbackErrorHandler (/mnt/goerli/goerli/lodestar/node_modules/fastify/lib/error-handler.js:127:3)
    at handleError (/mnt/goerli/goerli/lodestar/node_modules/fastify/lib/error-handler.js:61:5)
    at onErrorHook (/mnt/goerli/goerli/lodestar/node_modules/fastify/lib/reply.js:742:5)
    at Reply.send (/mnt/goerli/goerli/lodestar/node_modules/fastify/lib/reply.js:133:5)
    at defaultErrorHandler (/mnt/goerli/goerli/lodestar/node_modules/fastify/lib/error-handler.js:92:9) Cannot set headers after they are sent to the client
Error: Cannot set headers after they are sent to the client
    at new NodeError (node:internal/errors:399:5)
    at ServerResponse.setHeader (node:_http_outgoing:645:11)
    at ServerResponse.writeHead (node:_http_server:378:21)
    at onSendEnd (/mnt/goerli/goerli/lodestar/node_modules/fastify/lib/reply.js:596:7)
    at onSendHook (/mnt/goerli/goerli/lodestar/node_modules/fastify/lib/reply.js:529:5)
    at fallbackErrorHandler (/mnt/goerli/goerli/lodestar/node_modules/fastify/lib/error-handler.js:127:3)
    at handleError (/mnt/goerli/goerli/lodestar/node_modules/fastify/lib/error-handler.js:61:5)
    at onErrorHook (/mnt/goerli/goerli/lodestar/node_modules/fastify/lib/reply.js:742:5)
    at Reply.send (/mnt/goerli/goerli/lodestar/node_modules/fastify/lib/reply.js:133:5)
    at defaultErrorHandler (/mnt/goerli/goerli/lodestar/node_modules/fastify/lib/error-handler.js:92:9)
```

It looks like the VC abruptly aborts the connection. Might worth to investigate if we can more gracefully close the connection or if that's not possible at all reduce the verbosity of the log on the BN side.

`Error: Cannot set headers after they are sent to the client` should ideally never happen, this is caused when trying to set headers on a request that is already sent or aborted
https://github.com/ChainSafe/lodestar/blob/f5d12a893f5b913a04938275bb248ace1ca47539/packages/api/src/beacon/server/events.ts#L26-L27

### Expected behavior

Cleanly close all connections when shutting down VC, no error logs on the BN side.

### Steps to reproduce

Shutdown VC, observe BN logs

### Additional context

_No response_

### Operating system

Linux

v1.9.0

nazarhussain:
**Motivation**

Gracefully shutdown the connection. 

**Description**

When validator is loading up and waiting for genesis that can take a lot of time and during this time if the user press Ctrl+C then the following error is shown. 

> Error: aborted

So with this change we carefully wait for the validator to load and if user abort we skip that error. 

Closes #5706 

- Run all tests

### DIFF
--- a/packages/api/src/beacon/server/events.ts
+++ b/packages/api/src/beacon/server/events.ts
@@ -37,6 +37,9 @@ export function getRoutes(config: ChainForkConfig, api: ServerApi<Api>): ServerR
           await new Promise<void>((resolve, reject) => {
             void api.eventstream(req.query.topics, controller.signal, (event) => {
               try {
+                // If the request is already aborted, we don't need to send any more events.
+                if (req.raw.destroyed) return;
+
                 const data = eventSerdes.toJson(event);
                 res.raw.write(serializeSSEEvent({event: event.type, data}));
               } catch (e) {
--- a/packages/api/src/beacon/server/events.ts
+++ b/packages/api/src/beacon/server/events.ts
@@ -1,4 +1,5 @@
 import {ChainForkConfig} from "@lodestar/config";
+import {ErrorAborted} from "@lodestar/utils";
 import {Api, ReqTypes, routesData, getEventSerdes} from "../routes/events.js";
 import {ServerRoutes} from "../../utils/server/index.js";
 import {ServerApi} from "../../interfaces.js";
@@ -49,7 +50,12 @@ export function getRoutes(config: ChainForkConfig, api: ServerApi<Api>): ServerR
             // The client may disconnect and we need to clean the subscriptions.
             req.raw.once("close", () => resolve());
             req.raw.once("end", () => resolve());
-            req.raw.once("error", (err) => reject(err));
+            req.raw.once("error", (err) => {
+              if ("code" in err && (err as unknown as {code: string}).code === "ECONNRESET") {
+                return reject(new ErrorAborted());
+              }
+              return reject(err);
+            });
           });
 
           // api.eventstream will never stop, so no need to ever call `res.raw.end();`
--- a/packages/api/src/beacon/server/events.ts
+++ b/packages/api/src/beacon/server/events.ts
@@ -54,7 +54,7 @@ export function getRoutes(config: ChainForkConfig, api: ServerApi<Api>): ServerR
             req.raw.once("close", () => resolve());
             req.raw.once("end", () => resolve());
             req.raw.once("error", (err) => {
-              if ("code" in err && (err as unknown as {code: string}).code === "ECONNRESET") {
+              if ((err as unknown as {code: string}).code === "ECONNRESET") {
                 return reject(new ErrorAborted());
               }
               return reject(err);
--- a/packages/cli/src/cmds/validator/handler.ts
+++ b/packages/cli/src/cmds/validator/handler.ts
@@ -150,6 +150,7 @@ export async function validatorHandler(args: IValidatorCliArgs & GlobalArgs): Pr
 
   // This promise resolves once genesis is available.
   // It will wait for genesis, so this promise can be potentially very long
+
   const validator = await Validator.initializeFromBeaconNode(
     {
       db,
@@ -171,6 +172,7 @@ export async function validatorHandler(args: IValidatorCliArgs & GlobalArgs): Pr
   );
 
   onGracefulShutdownCbs.push(() => validator.close());
+
   // Start keymanager API backend
   // Only if keymanagerEnabled flag is set to true
   if (args["keymanager"]) {
--- a/packages/cli/src/cmds/validator/handler.ts
+++ b/packages/cli/src/cmds/validator/handler.ts
@@ -16,7 +16,6 @@ import {
   MonitoringService,
 } from "@lodestar/beacon-node";
 import {getNodeLogger} from "@lodestar/logger/node";
-import {isErrorAborted} from "@lodestar/utils";
 import {getBeaconConfigFromArgs} from "../../config/index.js";
 import {GlobalArgs} from "../../options/index.js";
 import {YargsError, cleanOldLogFiles, getDefaultGraffiti, mkdir, parseLoggerArgs} from "../../util/index.js";
@@ -151,8 +150,6 @@ export async function validatorHandler(args: IValidatorCliArgs & GlobalArgs): Pr
 
   // This promise resolves once genesis is available.
   // It will wait for genesis, so this promise can be potentially very long
-  // During the wait time user can abort the process with Ctrl+C
-  // So we have to wrap this code with the try/catch
   const validator = await Validator.initializeFromBeaconNode(
     {
       db,
@@ -171,13 +168,7 @@ export async function validatorHandler(args: IValidatorCliArgs & GlobalArgs): Pr
       distributed: args.distributed,
     },
     metrics
-  ).catch((err) => {
-    if (isErrorAborted(err)) {
-      throw new Error("Validator initialization aborted. Exiting.");
-    } else {
-      throw err;
-    }
-  });
+  );
 
   onGracefulShutdownCbs.push(() => validator.close());
   // Start keymanager API backend
--- a/packages/cli/src/cmds/validator/handler.ts
+++ b/packages/cli/src/cmds/validator/handler.ts
@@ -16,6 +16,7 @@ import {
   MonitoringService,
 } from "@lodestar/beacon-node";
 import {getNodeLogger} from "@lodestar/logger/node";
+import {isErrorAborted} from "@lodestar/utils";
 import {getBeaconConfigFromArgs} from "../../config/index.js";
 import {GlobalArgs} from "../../options/index.js";
 import {YargsError, cleanOldLogFiles, getDefaultGraffiti, mkdir, parseLoggerArgs} from "../../util/index.js";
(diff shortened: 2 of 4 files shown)


nazarhussain:
To reproduce the error run following command in `unstable` branch and then try in this branch. 

```bash
./lodestar dev --reset --startValidators=0..7 --rest.port=3400
```

After running this command wait for few seconds and press Ctrl+C.

nflaig (changes requested):
The errors mentioned in https://github.com/ChainSafe/lodestar/issues/5706 are solely on the beacon node side and are not addressed by this PR, although the unclean disconnect might be a fix in the validator client but it happens at any point when shutting down and not just at startup.

I am not sure if it is a good idea to swallow aborted errors during initialization of the validator client (or beacon node) as in that case those kinda make sense. Specifically in the case of "waiting for genesis..." it feels alright to me to print out a abort error if user aborts (CTRL+C).

nazarhussain:
@nflaig Please suggest the steps which you used to produce the issue mentioned in #5706. The only way I was able to produce was the steps mentioned above and I fixed it accordingly. 

Related to your other comment. When user presses Ctrl+C and we show a message
> Stopping gracefully, use Ctrl+C again to force process exit

Then it does not make useful to show abort error.

nflaig:
> Please suggest the steps which you used to produce the issue mentioned in https://github.com/ChainSafe/lodestar/issues/5706

Can be reproduce by running a beacon node

```sh
./lodestar beacon \
    --dataDir /home/devops/goerli/data/beacon \
    --rest \
    --rest.address "0.0.0.0" \
    --rest.namespace '*' \
    --metrics \
    --execution.urls http://localhost:8551 \
    --jwt-secret /home/devops/goerli/data/jwtsecret \
    --logLevel info \
    --network goerli
```

and a validator client

```sh
./lodestar validator \
    --dataDir /home/devops/goerli/data/validator \
    --beaconNodes http://localhost:9596 \
    --metrics \
    --logLevel info \
    --network goerli \
```

and then just shutting down (CTRL+C) the validator client which produces this log **on the beacon node**

```sh
Jun-30 17:00:44.746[rest]            error: Req req-1 eventstream error  aborted
Error: aborted
    at connResetException (node:internal/errors:720:14)
    at abortIncoming (node:_http_server:771:17)
    at socketOnClose (node:_http_server:765:3)
    at Socket.emit (node:events:523:35)
    at TCP.<anonymous> (node:net:334:12)
```

the other issue `Error: Cannot set headers after they are sent to the client` is hard to reproduce as it depends on the timing when shutting down the validator client.

nazarhussain:
@nflaig Thanks for sharing the snippet. Will check and fix the issue.

packages/cli/src/cmds/validator/handler.ts
```diff
@@ -150,60 +151,63 @@ export async function validatorHandler(args: IValidatorCliArgs & GlobalArgs): Pr
...
-    const keymanagerApi = new KeymanagerApi(
-      validator,
-      persistedKeysBackend,
-      abortController.signal,
-      proposerConfigWriteDisabled
-    );
-    const keymanagerServer = new KeymanagerRestApiServer(
+  // During the wait time user can abort the process with Ctrl+C
+  // So we have to wrap this code with the try/catch
+  try {
```
nflaig: do we really have to wrap the whole thing in try-catch, I feel like this is only relevant for "waiting for genesis ..." case

packages/api/src/beacon/server/events.ts
```diff
@@ -49,7 +50,12 @@ export function getRoutes(config: ChainForkConfig, api: ServerApi<Api>): ServerR
             // The client may disconnect and we need to clean the subscriptions.
             req.raw.once("close", () => resolve());
             req.raw.once("end", () => resolve());
-            req.raw.once("error", (err) => reject(err));
+            req.raw.once("error", (err) => {
+              if ("code" in err && (err as unknown as {code: string}).code === "ECONNRESET") {
```
nflaig: `"code" in err` why is this check needed? if it is undefined the 2nd case will fail anyhow

nflaig: I am not really sure if this is ideal, we might swallow cases where validator client crashed or disconnected for other reasons than shutting it down.

Was really hoping that this could be somehow addressed on the validator client side, it should not cause `ECONNRESET` error on the beacon node but disconnect gracefully.

nflaig (commented):
`Error: Cannot set headers after they are sent to the client` not yet addressed, can't close the issue with this in current state

packages/api/src/beacon/server/events.ts
```diff
@@ -36,6 +37,9 @@ export function getRoutes(config: ChainForkConfig, api: ServerApi<Api>): ServerR
           await new Promise<void>((resolve, reject) => {
             void api.eventstream(req.query.topics, controller.signal, (event) => {
               try {
+                // If the request is already aborted, we don't need to send any more events.
+                if (req.raw.destroyed) return;
```
nazarhussain, line 41: This check will try to avoid headers already sent scenario. Though that edge case is very tricky to reproduce. I tried a lot of ticks but was not able to produce it locally.

packages/api/src/beacon/server/events.ts
nazarhussain: We are not swallowing something, trying to log more relevant information. If the client crashed internally it's still aborted from the context of AbortController on client. So we can't differentiate on the server side of both scenarios.

nflaig: I haven't looked at the code specifically but shouldn't the validator client be able to more gracefully close the connection is it receives a abort signal?

packages/api/src/beacon/server/events.ts
nflaig: Are you sure this handles all cases where this can happen, I was think maybe to check `res.raw.headersSent` before setting hearders but would also have to look into this more if that the correct condition to check for

packages/cli/src/cmds/validator/handler.ts
```diff
@@ -169,10 +171,15 @@ export async function validatorHandler(args: IValidatorCliArgs & GlobalArgs): Pr
       distributed: args.distributed,
     },
     metrics
-  );
+  ).catch((err) => {
```
nflaig: As mentioned in https://github.com/ChainSafe/lodestar/pull/5511#discussion_r1200304394 by @tuyennhv we shouldn't use await and catch.

packages/cli/src/cmds/validator/handler.ts
```diff
@@ -169,10 +171,15 @@ export async function validatorHandler(args: IValidatorCliArgs & GlobalArgs): Pr
...
+    if (isErrorAborted(err)) {
+      throw new Error("Validator initialization aborted. Exiting.");
```
nflaig: Why do we want to throw an error? This will still print out a stack trace and there is not much difference compared to just throwing the abort error

After all, as you mentioned, we are printing "Stopping gracefully.." already

packages/api/src/beacon/server/events.ts
nazarhussain: headers already sent error occurred, when you try to write to a response, which related request is completed. That completion can be because request was aborted, or server already finished writing response and closed the stream with `res.end()`.  

I can't see the later case happens in our code, as we write the response only from one place. So the only case left in my mind is the first one, so this condition will cover it. 

If you can find a way to reproduce `headers already sent` scenario manually, please share then we can test.

packages/cli/src/cmds/validator/handler.ts
nazarhussain: I earlier used the `try/catch` pattern as suggested by @tuyennhv. You asked it does not feel good wrap it. So alternative is the following approach. As we don't prefer using `let` in our code.

packages/cli/src/cmds/validator/handler.ts
nazarhussain: It shows a better message to suer why stopping, than simply error stating `aborted`

packages/api/src/beacon/server/events.ts
nflaig: Based on the stack trace in the issue, the error is not caused by us setting headers but something fastify does.

Looks like here [res.writeHead]( https://github.com/fastify/fastify/blob/cc347d7c0b4266097b61b126158b797878668353/lib/reply.js#L597), might be an issue on fastify or related to us not closing correcty.

But `res.raw.write`  looks like a good candidate that could be causing this. If we are sure this does not cause unwanted side effects the change looks good.

> If you can find a way to reproduce headers already sent scenario manually

I just saw this once or twice on my end over a period of several months and it was reported by a user once. Seems to be highly timing specific, I tried to reproduce it consistently but no luck so far

packages/api/src/beacon/server/events.ts
```diff
@@ -49,7 +53,12 @@ export function getRoutes(config: ChainForkConfig, api: ServerApi<Api>): ServerR
             // The client may disconnect and we need to clean the subscriptions.
             req.raw.once("close", () => resolve());
             req.raw.once("end", () => resolve());
-            req.raw.once("error", (err) => reject(err));
+            req.raw.once("error", (err) => {
+              if ((err as unknown as {code: string}).code === "ECONNRESET") {
```
nflaig, line 57: I reviewed our code and did some research on how `EventSource` behaves. It looks like we are closing this correctly on the client by calling `.close()`. 

It is invoking both the `close` and `error` event on the server, this is a bit strange to me but it looks like that's how `EventSource` works, it doesn't gracefully close for some reason.

Another thing is that usually this would be handled by fastify which also just ignore errors with code `ECONNRESET`, see https://github.com/fastify/fastify/blob/cc347d7c0b4266097b61b126158b797878668353/fastify.js#L667C23-L667C33

But since the event API is not handled by fastify we need to do it ourselves and applying same pattern sounds good to me.

We might even consider just resolving here instead of rejecting with abort error but I guess if this error is not logged due to it being ignored upstream this is also fine.

packages/cli/src/cmds/validator/handler.ts
nflaig: Yeah, all of the proposed solutions look sub-optimal

packages/cli/src/cmds/validator/handler.ts
nflaig: I really don't know about this change, I am not sure we are fixing a real problem here. I never saw a user complaint about this and it shouldn't cause confusion and bad UX if we throw an aborted error here.

As this does not address points mentioned in the issue I would be in favor of removing it from this PR and opening a separate PR.

Maybe can get further opinions on this from others

packages/api/src/beacon/server/events.ts
nazarhussain: I don't see any side effects for this change.

packages/cli/src/cmds/validator/handler.ts
nazarhussain: Removed.

packages/api/src/beacon/server/events.ts
nazarhussain: I feel we should abort here on `ECONNRESET` and let the higher level stack handle that error, as it's currently doing. Once we have more resources to knew if the reset was actually abort or gracefully closed by client then we can decide to resolve the later scenario.

nflaig: Yeah, sounds good to me 👍

nflaig (approved):
Looks good to me 👍 

Let's also update the PR description to reflect the changes implemented

packages/api/src/beacon/server/events.ts
nflaig: > It is invoking both the close and error event on the server, this is a bit strange to me but it looks like that's how EventSource works, it doesn't gracefully close for some reason.

This might be related to
- https://github.com/nodejs/node/issues/42154

nflaig: This is also simliar to issue reported here
- https://github.com/websockets/ws/issues/1256

packages/api/src/beacon/server/events.ts
nflaig: @nazarhussain I was going through logs from sim tests and it looks like the error still happens there.
```
[node-1-cl-lodestar] [8460]: Eph 6/0 2.220[node-1-cl-lodestar] [31merror[39m: uncaughtException: Cannot set headers after they are sent to the client
Error: Cannot set headers after they are sent to the client
    at new NodeError (node:internal/errors:405:5)
    at ServerResponse.setHeader (node:_http_outgoing:648:11)
    at ServerResponse.writeHead (node:_http_server:381:21)
    at onSendEnd (/home/runner/actions-runner/_work/lodestar/lodestar/node_modules/fastify/lib/reply.js:597:7)
    at onSendHook (/home/runner/actions-runner/_work/lodestar/lodestar/node_modules/fastify/lib/reply.js:530:5)
    at fallbackErrorHandler (/home/runner/actions-runner/_work/lodestar/lodestar/node_modules/fastify/lib/error-handler.js:127:3)
    at handleError (/home/runner/actions-runner/_work/lodestar/lodestar/node_modules/fastify/lib/error-handler.js:61:5)
    at onErrorHook (/home/runner/actions-runner/_work/lodestar/lodestar/node_modules/fastify/lib/reply.js:743:5)
    at Reply.send (/home/runner/actions-runner/_work/lodestar/lodestar/node_modules/fastify/lib/reply.js:133:5)
    at defaultErrorHandler (/home/runner/actions-runner/_work/lodestar/lodestar/node_modules/fastify/lib/error-handler.js:92:9) Cannot set headers after they are sent to the client
Error: Cannot set headers after they are sent to the client
```

See [node-1-cl-lodestar.log](https://github.com/ChainSafe/lodestar/files/12111895/node-1-cl-lodestar.log) from [this sim tests run](https://github.com/ChainSafe/lodestar/actions/runs/5613234520). This error seems to happen quite frequently in sim tests.

I would suggest we revert this change here as it does not solve the issue. It might also be a bug in fastify and not directly solvable on our end.

**Edit:** I created an issue to keep track of this
- https://github.com/ChainSafe/lodestar/issues/5783

wemeetagain:

PR #5723, opened 2023-06-30
feat: support el_offline in eth/v1/node/syncing

### LINKED ISSUE #5542: Feature: Support `el_offline` in `eth/v1/node/syncing`

When using Lighthouse VC, it'd be immensely useful if Lodestar could signal that its EL is offline, so the VC can mark it down.

This is handled by the `el_offline` field for `eth/v1/node/syncing`

See https://github.com/sigp/lighthouse/pull/4295 for reference

nazarhussain:
**Motivation**

Add support for the status of execution engine in syncing status endpoint. 

**Description**

- Add support for the status of execution engine. 

Closes #5542 

- Run all tests.


X-oss-byte (approved):

packages/beacon-node/src/api/impl/node/index.ts
```diff
@@ -63,11 +63,11 @@ export function getNodeApi(
     },
 
     async getSyncingStatus() {
-      return {data: sync.getSyncStatus()};
+      return {data: await sync.getSyncStatus()};
     },
 
     async getHealth(_req, res) {
-      if (sync.getSyncStatus().isSyncing) {
+      if ((await sync.getSyncStatus()).isSyncing) {
```
wemeetagain: How often are these called? and will this effect validator performance that now each response will likely take twice as long as before (because each call now awaits a response by the EL)?

packages/beacon-node/src/execution/engine/http.ts
```diff
@@ -398,6 +397,18 @@ export class ExecutionEngineHttp implements IExecutionEngine {
     });
     return response.map(deserializeExecutionPayloadBody);
   }
+
+  async isOffline(): Promise<boolean> {
+    try {
+      await this.rpc.fetch({method: "eth_chainId", params: []}, {});
```
nflaig: how should timeouts be handled? I feel waiting 12 seconds here for a response is not great. 

AFAIK only Lighthouse VC uses el_offline at the moment, we might wanna check how their beacon node handles this.

packages/beacon-node/test/e2e/api/impl/beacon/node/endpoints.test.ts
```diff
@@ -0,0 +1,86 @@
...
+import {LogLevel, testLogger} from "../../../../../utils/logger.js";
+import {getDevBeaconNode} from "../../../../../utils/node/beacon.js";
+import {BeaconNode} from "../../../../../../src/node/nodejs.js";
+
+describe("beacon node api", function () {
+  this.timeout("30s");
+
+  const restPort = 9596;
+  const config = createBeaconConfig(chainConfigDef, Buffer.alloc(32, 0xaa));
+  const validatorCount = 512;
```
nflaig: should consider reducing validator count to 1 or even 0, not relevant for /eth/v1/node/ API tests

packages/beacon-node/test/e2e/api/impl/beacon/node/endpoints.test.ts
```diff
@@ -0,0 +1,86 @@
...
+  after(async () => {
+    await bn.close();
+  });
+
+  describe("getSyncingStatus", async () => {
+    it("should return valid syncing status", async () => {
+      const res = await client.node.getSyncingStatus();
+      ApiError.assert(res);
+
+      expect(res.response.data).to.include.keys("headSlot", "syncDistance", "isSyncing", "isOptimistic", "elOffline");
```
nflaig: keys will already be checked by oapi spec tests, might wanna do some sanity checks on the actual values here

packages/beacon-node/test/e2e/api/impl/beacon/node/endpoints.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    });
+
+    it("should return 'elOffline' as 'true' for default dev node", async () => {
```
nflaig: should we rather use spec value in test description?

```suggestion
    it("should return 'el_offline' as 'true' for default dev node", async () => {
```

packages/beacon-node/src/sync/sync.ts
```diff
@@ -78,15 +78,18 @@ export class BeaconSync implements IBeaconSync {
     this.unknownBlockSync.close();
   }
 
-  getSyncStatus(): SyncingStatus {
+  async getSyncStatus(): Promise<SyncingStatus> {
     const currentSlot = this.chain.clock.currentSlot;
+    const elOffline = await this.chain.executionEngine.isOffline();
```
dapplion: We should deeply review the implications of adding this call here, sync is delicate. Also, what's somewhat concerning is that this is a different code path than other executionEngine consumers. It's possible that:
- block processing attempts to call executionEngine and errors because it's offline
- this getSyncStatus calls eth_chainId and succeeds, considering el online
How to resolve this inconsistency? It can happen both ways

nazarhussain: @dapplion I explicitly checked for `eth_chainId` for `ECONNREFUSED` which can happen only if server is offline, not in case server respond error. 

As far I knew it's the same `executionEngine` used by the block processing. If there is a some error say 500 or 400 range that's concerns to block processing, it still means that server is online.

packages/beacon-node/src/execution/engine/http.ts
```diff
@@ -82,6 +86,14 @@ const getPayloadOpts: ReqOpts = {routeId: "getPayload"};
  * https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.1/src/engine/interop/specification.md
  */
 export class ExecutionEngineHttp implements IExecutionEngine {
+  readonly emitter = new ExecutionEngineEventEmitter();
```
wemeetagain: Why do we need this emitter?

nazarhussain: For this use case. 

https://github.com/ChainSafe/lodestar/blob/df7a0c6650acf283ac9f0e5b7d18315a4c9b19a9/packages/beacon-node/src/node/nodejs.ts#L225-L237

g11tech: since we are not fcUing on state change as of now (as per discord discussions) may be we can remove it to kkeep techncal debt low

g11tech:
I would like to simplfy the implementation :
 
 1. For offline/online/auth state latch on to httpClient (or wrap it), so any/all calls to the engine can be used to update state
 2. For syncing/synced track the status and update (as you have correctly done this)
 
 So for 1. would like track on level of httpClient

packages/beacon-node/src/node/nodejs.ts
```diff
@@ -66,6 +66,7 @@ enum LoggerModule {
   backfill = "backfill",
   chain = "chain",
   eth1 = "eth1",
+  executionEngine = "execution_engine",
```
nflaig: what about just using `execution`? I feel like this is too long and inconsistent with other module names.

For CLI options we also just use `execution.` prefix

https://github.com/ChainSafe/lodestar/blob/96820d72b14fdeb4b1c21f6f1c6c8dc604bc8331/packages/cli/src/options/beaconNodeOptions/execution.ts#L6

wemeetagain (approved):
LGTM

g11tech (commented):
post merge review LGTM :+1:

wemeetagain:

PR #5729, opened 2023-07-05
feat: verify attestation gossip messages in batch

### LINKED ISSUE #5416: Batch gossip validation

Right now validate gossip messages separately (but do them concurrently), we could try to verify them in batch

- Refactor gossip queue with keeping `add()` as is (add 1 item) but `next()` takes multiple items from the queue
- all `validateGossip*` functions should take an array of messages instead of 1 message
  - Phase 0: do all verifications except for signature verification
  - Phase 1: do signature verification for all *valid* messages after phase 0
    - Aggregate pubkeys/signatures before doing signature verification
    - If one of them fails, retry with doing them separately (like the current flow)

- Consider if it's worth to index the attestation queue by target root, in order to regen state leveraging the work of @dapplion 
- Alternative would be to use the current SeenAttestationDatas cache, just improve the signature verification, state regen may not be the bottle neck for now

twoeths:
**Motivation**

- verifying multiple messages are >=10x faster than `verifyMultipleSignatures` - this includes aggregating pubkey and signatures and do a simple `aggregatedSignature.verify()`, this is the number from the `bls.test.ts` performance test

|num messages|32|64|128|
|---------------|--|---|----|
|verifyMultipleSignatures|33.1 ms/op|65.4 ms/op|130.5 ms/op|
|batch verify signatures with same message|3.49 ms/op|5.16 ms/op|8.97 ms/op|

- as of Jul 2023, there are up to 316 validators signing for the same attestation data https://beaconcha.in/block/17620266#attestations

**Description**

- Define new `IndexedGossipQueue`, when `beacon_attestation` gossip messages come, we index by attestation data base64, new param for this queue
  - `indexFn`: function to index to store as key
  - `minChunkSize`: keys with at least min chunk size will be processed first, this is 32
  - `maxChunkSize`: get up to `maxChunkSize` to process in batch, this is 128
- Define new `validateGossipAttestationsSameAttData` function to verify attestations with the same attestation data:
  - Phase0: Verify attestations following the spec without the signature check, and get signatures from there
  - Phase1: Verify signatures of these attestations
    - Do a quick check: aggregate pubkeys and signatures and verify, this takes 10ms for 128 signatures
    - If one of the signature is invalid, retry with the old way (`bls.verifySignatureSets`) with `batchable: false`

Closes #5416

**Test result**
- can process all attestations on a mainnet node subscribing to all subnets
- time to process an attestation is reduced to <10ms per message
- around 90% of attestations are verified in batch and main thread, so cpu time is reduced a lot to 140% -160%(>=300% on unstable) 
- `IndexedGossipQueue` works well: 0.06% cpu for `next()` and 0.28% for `add()`

**TODOs**

- `validateGossipAttestationsSameAttData` take 15% of cpu time, verifying signatures on main thread seem only suitable for network thread (`useWorker=true`)
- we may still want to offload the new batch signature verification to a worker thread if `useWorker=false`

### DIFF
--- a/packages/beacon-node/src/chain/blocks/importBlock.ts
+++ b/packages/beacon-node/src/chain/blocks/importBlock.ts
@@ -89,11 +89,14 @@ export async function importBlock(
 
   this.metrics?.importBlock.bySource.inc({source});
   this.logger.verbose("Added block to forkchoice and state cache", {slot: block.message.slot, root: blockRootHex});
-  this.emitter.emit(routes.events.EventType.block, {
-    block: toHexString(this.config.getForkTypes(block.message.slot).BeaconBlock.hashTreeRoot(block.message)),
-    slot: block.message.slot,
-    executionOptimistic: blockSummary != null && isOptimisticBlock(blockSummary),
-  });
+  // We want to import block asap so call all event handler in the next event loop
+  setTimeout(() => {
+    this.emitter.emit(routes.events.EventType.block, {
+      block: toHexString(this.config.getForkTypes(block.message.slot).BeaconBlock.hashTreeRoot(block.message)),
+      slot: block.message.slot,
+      executionOptimistic: blockSummary != null && isOptimisticBlock(blockSummary),
+    });
+  }, 0);
 
   // 3. Import attestations to fork choice
   //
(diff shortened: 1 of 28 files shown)


twoeths:
replaced by #5896

Issue #5611, opened 2023-06-05
Update to Nodejs v20


wemeetagain:
### Problem description

Nodejs v20 is released and holds promise to be somewhat more performant (possibly resolving #4903)

### Solution description

Update to v20, resolving any build issues that arise.

- [x] blst https://github.com/ChainSafe/lodestar/pull/5612
- [x] cpu_features https://github.com/ChainSafe/lodestar/pull/5674
- [x] snappy https://github.com/ChainSafe/lodestar/pull/5617
- [x] gc-stats https://github.com/ChainSafe/lodestar/pull/5677 https://github.com/ChainSafe/lodestar/pull/5683 https://github.com/ChainSafe/lodestar/pull/5686 https://github.com/ChainSafe/lodestar/pull/5690

philknows:
Would also do a quick scan of the documentation to ensure anywhere we state node requirements to use node 20.

wemeetagain:
`unstable` branch with Node 20 has been deployed to our `feat1` group for several days

wemeetagain:
Closed with 1.10.0

PR #5730, opened 2023-07-05
chore!: use node v20 throughout monorepo

wemeetagain:
**Motivation**

Since #5611, we support node v20, but we aren't using it everywhere

**Description**

Update docs, workflows, types, and docker images to use node 20

TODO
- Update sim tests to node 20


dapplion:
Can you confirm the performance regression locally? Seems legit

wemeetagain:
This error is only happening in CI, not locally
https://github.com/ChainSafe/lodestar/actions/runs/5522181144/jobs/10071384231?pr=5730#step:11:6935

Dockerfile
```diff
@@ -1,7 +1,7 @@
 
 # --platform=$BUILDPLATFORM is used build javascript source with host arch
 # Otherwise TS builds on emulated archs and can be extremely slow (+1h)
-FROM --platform=${BUILDPLATFORM:-amd64} node:18-alpine as build_src
+FROM --platform=${BUILDPLATFORM:-amd64} node:20-alpine as build_src
```
nflaig, line 4: This can lead to non-deterministic builds, we might install a different node version on the release compared to the rc.

Would suggest we at least specify the minor version to reduce this risk with the drawback that we have to more frequently update the value here.

```suggestion
FROM --platform=${BUILDPLATFORM:-amd64} node:20.4-alpine as build_src
```

nflaig:
> This error is only happening in CI, not locally
https://github.com/ChainSafe/lodestar/actions/runs/5522181144/jobs/10071384231?pr=5730#step:11:6935

Can't reproduce this either, installed several different node versions. Might have to check which node version is actually used in the CI.

Dockerfile
nflaig: @wemeetagain would have to apply this change more consistently, not just here. We ideally want to have the same node version in CI as in docker builds.

Im still not 100% sure if it is worth to fix minor but had bad experience in the past not doing it.

nflaig:
No idea what is going on with sim tests, @nazarhussain any ideas?

Dockerfile
wemeetagain: Lets save this for another PR and deeper discussion

nazarhussain:
Based on the error, seems like the beacon node had not started. 

>   <rejected> FetchError: request to http://127.0.0.1:5004/eth/v1/node/peer_count failed, reason: socket hang up

nflaig:
I think it is somehow related with error thrown after updating cross-fetch, still can't make sense of why this produces correct results locally.

nazarhussain:
Internally the error says lodestar could not connect to the EL. But Geth is running fine, so it seems like some firewall issue on docker which didn't allowed connecting the host machine to docker container on local IP. 

> [node-1-cl-lodestar] [4384]: Eph 0/1 1.779[node-1-cl-lodestar/eth1] [31merror[39m: Error updating eth1 chain cache  request to http://127.0.0.1:9001/ failed, reason: connect ECONNREFUSED 127.0.0.1:9001
FetchError: request to http://127.0.0.1:9001/ failed, reason: connect ECONNREFUSED 127.0.0.1:9001

wemeetagain:
I updated cross-fetch because the error was present, but it didn't fix it.
I think the problem is related to updated behavior of the native fetch implementation, something that changed between v18 and v20.

wemeetagain:
It seems that node 20 now has `fetch` available natively without needing the `--experimental-fetch` flag.
So that probably explains the difference in behavior.

philknows (dismissed):

nflaig (dismissed):

packages/beacon-node/test/e2e/eth1/jsonRpcHttpClient.test.ts
```diff
@@ -148,7 +153,16 @@ describe("eth1 / jsonRpcHttpClient", function () {
       const controller = new AbortController();
       if (abort) setTimeout(() => controller.abort(), 50);
       const eth1JsonRpcClient = new JsonRpcHttpClient([url], {signal: controller.signal});
-      await expect(eth1JsonRpcClient.fetch(payload, {timeout})).to.be.rejectedWith(error);
+
+      try {
+        await eth1JsonRpcClient.fetch(payload, {timeout});
```
nflaig: @nazarhussain try/catch is problematic here, test case would pass if no error is thrown

Either have to do explicit fail if there is no error like it is done here
https://github.com/ChainSafe/lodestar/blob/cdca9fbd589e450bcdb190032e97e16ea4ba0f13/packages/cli/test/unit/validator/decryptKeystoreDefinitions.test.ts#L77

or can use `.to.be.rejected.then` pattern

README.md
```diff
@@ -8,7 +8,7 @@
 [![Ethereum Consensus Spec v1.1.10](https://img.shields.io/badge/ETH%20consensus--spec-1.1.10-blue)](https://github.com/ethereum/consensus-specs/releases/tag/v1.1.10)
 [![codecov](https://codecov.io/gh/ChainSafe/lodestar/branch/unstable/graph/badge.svg)](https://codecov.io/gh/ChainSafe/lodestar)
 ![ES Version](https://img.shields.io/badge/ES-2020-yellow)
-![Node Version](https://img.shields.io/badge/node-18.15.x-green)
+![Node Version](https://img.shields.io/badge/node-20.x-green)
```
nflaig, line 11: @wemeetagain I updated this to 20.x as well, not sure if it was intentionally not updated before but as we discussed that we want to advocate for using node 20 I think we should update it here as well.

Last reference to 18.x is in package.json but that might be to disruptive for users to update to 20.x, although we could ensure everyone is using node 20 this way
https://github.com/ChainSafe/lodestar/blob/18872876dd61793700710a9b47a2a80dfee920ab/package.json#L5

docs/install/source.md
```diff
@@ -2,7 +2,7 @@
 
 ## Prerequisites
 
-Make sure to have [Yarn installed](https://classic.yarnpkg.com/en/docs/install). It is also recommended to [install NVM (Node Version Manager)](https://github.com/nvm-sh/nvm) and use the LTS version (currently v18) of [NodeJS](https://nodejs.org/en/).
+Make sure to have [Yarn installed](https://classic.yarnpkg.com/en/docs/install). It is also recommended to [install NVM (Node Version Manager)](https://github.com/nvm-sh/nvm) and use the LTS version (currently v20) of [NodeJS](https://nodejs.org/en/).
```
nflaig, line 5: Just a note: v20 is not yet LTS version until 2023-10-24, see [nodejs-release-working-group](https://github.com/nodejs/Release#nodejs-release-working-group) for schedule

nflaig (approved):
Looks good to me now, everything besides sim tests are using node 20 now.

wemeetagain:
Ok, when CI is finished running, I'll change the branch rules and merge this

packages/beacon-node/test/e2e/eth1/jsonRpcHttpClient.test.ts
```diff
@@ -153,16 +153,13 @@ describe("eth1 / jsonRpcHttpClient", function () {
...
+      await expect(eth1JsonRpcClient.fetch(payload, {timeout})).to.be.rejected.then((error) => {
         if (testCase.errorCode) {
-          expect((error as FetchError).code).to.eql(testCase.errorCode);
+          expect((error as FetchError).code).to.be.equal(testCase.errorCode);
         } else {
-          expect((error as Error).message).includes(testCase.error);
+          expect((error as Error).message).to.include(testCase.error);
         }
-      }
+      });
```
nazarhussain: I think we discussed to no use the `nested then/catch` and prefer `try/catch`.

packages/beacon-node/test/e2e/eth1/jsonRpcHttpClient.test.ts
nazarhussain: Yes I knew that pattern, but we discussed earlier to not use the `nested then/catch` and prefer `try/catch`.

nflaig: In that case need to explicitly fail the test if there is no error

packages/beacon-node/test/e2e/eth1/jsonRpcHttpClient.test.ts
nflaig: I was not part of that discussion but in my opinion try/catch in tests should be last resort if there are no other options to use built-in assertion syntax

nflaig:
Regarding the failing sim tests on node 20, it looks like the issue is related to node-fetch and Lighthouse.

I am using the script mentioned here https://github.com/ChainSafe/lodestar/issues/5737#issuecomment-1627305176 to fetch the state.

For `getStateV2` I am getting an error after few milliseconds which looks similar to the error in sim tests

```
/home/nico/projects/test/debug-getStatev2/node_modules/node-fetch/lib/index.js:1505
                        reject(new FetchError(`request to ${request.url} failed, reason: ${err.message}`, 'system', err));
                               ^
FetchError: request to http://localhost:5052/eth/v2/debug/beacon/states/6896350 failed, reason: socket hang up
    at ClientRequest.<anonymous> (/home/nico/projects/test/debug-getStatev2/node_modules/node-fetch/lib/index.js:1505:11)
    at ClientRequest.emit (node:events:512:28)
    at Socket.socketOnEnd (node:_http_client:519:9)
    at Socket.emit (node:events:524:35)
    at endReadableNT (node:internal/streams/readable:1378:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  type: 'system',
  errno: 'ECONNRESET',
  code: 'ECONNRESET'
}
```

There is an open issue for it already
- https://github.com/node-fetch/node-fetch/issues/1735

**Edit:** I created an issue to track this
- https://github.com/ChainSafe/lodestar/issues/5765

wemeetagain:

PR #5731, opened 2023-07-05
feat: extend attestation validity till end of next epoch - eip7045

g11tech:
ref: 
https://github.com/ethereum/consensus-specs/pull/3360

- [x] upgrade and check spec tests
~~- [ ] modify upgrade deneb state?~~ no upgrade changes needed

### DIFF
--- a/packages/state-transition/src/block/processAttestationPhase0.ts
+++ b/packages/state-transition/src/block/processAttestationPhase0.ts
@@ -1,7 +1,7 @@
 import {toHexString} from "@chainsafe/ssz";
-import {phase0, ssz} from "@lodestar/types";
+import {Slot, phase0, ssz} from "@lodestar/types";
 
-import {MIN_ATTESTATION_INCLUSION_DELAY, SLOTS_PER_EPOCH} from "@lodestar/params";
+import {MIN_ATTESTATION_INCLUSION_DELAY, SLOTS_PER_EPOCH, ForkSeq} from "@lodestar/params";
 import {computeEpochAtSlot} from "../util/index.js";
 import {CachedBeaconStatePhase0, CachedBeaconStateAllForks} from "../types.js";
 import {isValidIndexedAttestation} from "./index.js";
@@ -22,7 +22,7 @@ export function processAttestationPhase0(
   const slot = state.slot;
   const data = attestation.data;
 
-  validateAttestation(state, attestation);
+  validateAttestation(ForkSeq.phase0, state, attestation);
 
   const pendingAttestation = ssz.phase0.PendingAttestation.toViewDU({
     data: data,
@@ -56,7 +56,11 @@ export function processAttestationPhase0(
   }
 }
 
-export function validateAttestation(state: CachedBeaconStateAllForks, attestation: phase0.Attestation): void {
+export function validateAttestation(
+  fork: ForkSeq,
+  state: CachedBeaconStateAllForks,
+  attestation: phase0.Attestation
+): void {
   const {epochCtx} = state;
   const slot = state.slot;
   const data = attestation.data;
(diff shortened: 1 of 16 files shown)


packages/state-transition/src/block/processAttestationsAltairPlus.ts
```diff
@@ -152,7 +161,8 @@ export function getAttestationParticipationStatus(
 
   let flags = 0;
   if (isMatchingSource && inclusionDelay <= SLOTS_PER_EPOCH_SQRT) flags |= TIMELY_SOURCE;
-  if (isMatchingTarget && inclusionDelay <= SLOTS_PER_EPOCH) flags |= TIMELY_TARGET;
+  // post deneb attestation is valid till end of next epoch for target
+  if (isMatchingTarget && (fork >= ForkSeq.deneb || inclusionDelay <= SLOTS_PER_EPOCH)) flags |= TIMELY_TARGET;
```
matthewkeil: If forkSeq >= deneb and the attestation is more than the allowed end of next epoch old does this fall through as timely?

packages/state-transition/src/block/processAttestationPhase0.ts
```diff
@@ -80,7 +84,14 @@ export function validateAttestation(state: CachedBeaconStateAllForks, attestatio
         `targetEpoch=${data.target.epoch} computedEpoch=${computedEpoch}`
     );
   }
-  if (!(data.slot + MIN_ATTESTATION_INCLUSION_DELAY <= slot && slot <= data.slot + SLOTS_PER_EPOCH)) {
+
+  // post deneb, the attestations are valid till end of next epoch
+  if (
+    !(
+      data.slot + MIN_ATTESTATION_INCLUSION_DELAY <= slot &&
+      (fork >= ForkSeq.deneb || slot <= data.slot + SLOTS_PER_EPOCH)
```
matthewkeil: Does there need to be a check for the extra epoch worth of slots if deneb here?

matthewkeil (commented):
A couple of questions do just double check the conditionals.  Otherwise things look good!!

packages/state-transition/src/block/processAttestationsAltairPlus.ts
g11tech: attestation target is valid for this/previous epoch (see other validations) ... so not possible for inclusionDelay to be beyond next epoch (see other checks)

packages/state-transition/src/block/processAttestationPhase0.ts
g11tech: same reason as previous, attestation is already invalid if target not in current or previous epoch

g11tech:
> A couple of questions do just double check the conditionals. Otherwise things look good!!

thanks for looking into this, target epoch validations check this and this implementation closely follow the specs (check spec PR)

packages/state-transition/src/block/processAttestationsAltairPlus.ts
```diff
@@ -28,7 +29,8 @@ const TIMELY_TARGET = 1 << TIMELY_TARGET_FLAG_INDEX;
 const TIMELY_HEAD = 1 << TIMELY_HEAD_FLAG_INDEX;
 const SLOTS_PER_EPOCH_SQRT = intSqrt(SLOTS_PER_EPOCH);
 
-export function processAttestationsAltair(
+export function processAttestationsAltairPlus(
```
dapplion: Should this be called by fork name?
```suggestion
export function processAttestationsDeneb(
```

g11tech: umm.. it also processes post altair pre deneb attestations (so a unified fn for processing attestations on/after altair) like how we handle everything else in a generic way (except phase0 which are so different , so we can let it stay separate)

twoeths:
@g11tech is p2p change supposed for this PR or another one? https://github.com/ethereum/consensus-specs/pull/3360/files#diff-21cd1dde2a4815e714ef42d0cefa1bbd2997a4358ecf6e9e51025332c1d642fdR196

g11tech:
> @g11tech is p2p change supposed for this PR or another one? https://github.com/ethereum/consensus-specs/pull/3360/files#diff-21cd1dde2a4815e714ef42d0cefa1bbd2997a4358ecf6e9e51025332c1d642fdR196

was thinking of doing a followup to this PR

packages/state-transition/src/block/processAttestationsAltairPlus.ts
g11tech: how about `processAttestationsMultiFork`?

dapplion:
@g11tech pushed to the PR a quick review:
- I don't think there's any need to rename the function to *Plus, let's just keep the names as is
- Used a more verbose if else pattern in timelyTarget to make it more clear the fork branching logic

Can you please add the p2p changes in this PR too? I should not be that big

g11tech:
> this

sounds good :+1:

g11tech:
for the p2p validations, should we just update to the new ones irrespective of the fork since new validations are more relaxed than the pre-deneb. Even if for pre-deneb if some attestations gets passed through, the actual process attestation will reject it pre-deneb.. wdyt @tuyennhv @dapplion 

---------------
UPDATE: 

as per discussion with @dapplion : "p2p validation must match the spec for all forks"
will update the validations accordingly

wemeetagain (approved):

wemeetagain:

PR #5734, opened 2023-07-07
feat: verifySignatureSetsSameSigningRoot bls api

twoeths:
**Motivation**

- verifying signature sets of the same signing root is 6x to >10x faster than `verifyMultipleSignatures()` api
- as a preparation for #5729 where we'll validate signature sets with same signing root
- we want to leverage our worker pool to control `canAcceptWork()` (if we should process a gossip message or not) 

**Description**

- new `verifyManySignatureSetsSameMessage()` worker api
- enhance `verifySignatureSetsMaybeBatch()` to add new param `isSameMessage: boolean`
- separate `QueueItem` type (an item in "jobs" queue) and `JobItem` type (which contain a resolve/reject function of a promise)

```typescript
/**
 * A job item we want to get a result for.
 */
export type JobItem<W = BlsWorkReq | SerializedSet, R = boolean> = {
  resolve: (result: R | PromiseLike<R>) => void;
  reject: (error?: Error) => void;
  addedTimeMs: number;
  workReq: W;
};

/**
 * An item in the queue. Could be BlsWorkReq or SerializedSet[] of the same message.
 */
export type QueueItem = JobItem<BlsWorkReq> | JobItem<SerializedSet>[];
```
- an item in the queue could be a "multi sig sets" job item or an array of job item of same message signature set
- `verifySignatureSetsSameSigningRoot` has "verifyOnMainThread" to verify on main thread or not, consumer can decide it later (in #5729)

part of #5416

**Notes**
- This introduces new api, noone consume this new api atm
- cc @matthewkeil @dapplion

### DIFF
--- a/packages/beacon-node/src/chain/bls/maybeBatch.ts
+++ b/packages/beacon-node/src/chain/bls/maybeBatch.ts
@@ -16,9 +16,11 @@ export type SignatureSetDeserialized = {
 export function verifySignatureSetsMaybeBatch(sets: SignatureSetDeserialized[], isSameMessage = false): boolean {
   if (sets.length >= MIN_SET_COUNT_TO_BATCH) {
     if (isSameMessage) {
-      // TODO: return false if not same message or throw error?
+      // Consumers need to make sure that all sets have the same message
       const aggregatedPubkey = bls.PublicKey.aggregate(sets.map((set) => set.publicKey));
       const aggregatedSignature = bls.Signature.aggregate(
+        // As of Jul 2023 we could skip the signature validation here, no attack scenario found
+        // however this does not have the same security guarantee as the regular verify
         sets.map((set) => bls.Signature.fromBytes(set.signature, CoordType.affine, false))
       );
       return aggregatedSignature.verify(aggregatedPubkey, sets[0].message);
--- a/packages/beacon-node/src/chain/bls/maybeBatch.ts
+++ b/packages/beacon-node/src/chain/bls/maybeBatch.ts
@@ -19,9 +19,8 @@ export function verifySignatureSetsMaybeBatch(sets: SignatureSetDeserialized[], 
       // Consumers need to make sure that all sets have the same message
       const aggregatedPubkey = bls.PublicKey.aggregate(sets.map((set) => set.publicKey));
       const aggregatedSignature = bls.Signature.aggregate(
-        // As of Jul 2023 we could skip the signature validation here, no attack scenario found
-        // however this does not have the same security guarantee as the regular verify
-        sets.map((set) => bls.Signature.fromBytes(set.signature, CoordType.affine, false))
+        // true = validate signature
+        sets.map((set) => bls.Signature.fromBytes(set.signature, CoordType.affine, true))
       );
       return aggregatedSignature.verify(aggregatedPubkey, sets[0].message);
     }
--- a/packages/beacon-node/src/chain/bls/maybeBatch.ts
+++ b/packages/beacon-node/src/chain/bls/maybeBatch.ts
@@ -13,8 +13,17 @@ export type SignatureSetDeserialized = {
  * Verify signatures sets with batch verification or regular core verify depending on the set count.
  * Abstracted in a separate file to be consumed by the threaded pool and the main thread implementation.
  */
-export function verifySignatureSetsMaybeBatch(sets: SignatureSetDeserialized[]): boolean {
+export function verifySignatureSetsMaybeBatch(sets: SignatureSetDeserialized[], isSameMessage = false): boolean {
   if (sets.length >= MIN_SET_COUNT_TO_BATCH) {
+    if (isSameMessage) {
+      // TODO: return false if not same message or throw error?
+      const aggregatedPubkey = bls.PublicKey.aggregate(sets.map((set) => set.publicKey));
+      const aggregatedSignature = bls.Signature.aggregate(
+        sets.map((set) => bls.Signature.fromBytes(set.signature, CoordType.affine, false))
+      );
+      return aggregatedSignature.verify(aggregatedPubkey, sets[0].message);
+    }
+
     return bls.Signature.verifyMultipleSignatures(
       sets.map((s) => ({
         publicKey: s.publicKey,
--- a/packages/beacon-node/src/chain/bls/multithread/index.ts
+++ b/packages/beacon-node/src/chain/bls/multithread/index.ts
@@ -16,6 +16,7 @@ import {IBlsVerifier, VerifySignatureOpts} from "../interface.js";
 import {getAggregatedPubkey, getAggregatedPubkeysCount} from "../utils.js";
 import {verifySignatureSetsMaybeBatch} from "../maybeBatch.js";
 import {
+  ApiName,
   BlsWorkReq,
   BlsWorkResult,
   JobItem,
@@ -182,7 +183,7 @@ export class BlsMultiThreadWorkerPool implements IBlsVerifier {
           }))
         );
       } finally {
-        if (timer) timer();
+        if (timer) timer({api: ApiName.verifySignatureSets});
       }
     }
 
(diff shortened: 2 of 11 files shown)


wemeetagain:
> verifySignatureSetsSameSigningRoot has "verifyOnMainThread" to verify on main thread or not

Not sure that we'd ever want to verify on the main thread?

twoeths:
> > verifySignatureSetsSameSigningRoot has "verifyOnMainThread" to verify on main thread or not
> 
> Not sure that we'd ever want to verify on the main thread?

@wemeetagain  I think the theory of not to do heavy operations on main thread is because normally main thread is for I/O operation. Now we do I/O operations on worker thread, we can test verifying signatures (the 1st round) on main thread (and leave all fallback verifications on worker thread), this is when `useWorker=true`. The network processor is designed not to do anything when we receive/process gossip block already

I tested `verifyOnMainThread=true` with `useWorker=false` (no network thread), it takes 15% of cpu profile and mesh peers range from 2 to 6 which is not too bad (verification time of an attestation is <10ms). I'm not sure `verifyOnMainThread=true` works fine with `useWorker=true` but would like to give us a chance to test that

twoeths:
another use case for `verifyOnMainThread=true` is #5739

packages/beacon-node/src/chain/bls/maybeBatch.ts
```diff
@@ -13,8 +13,18 @@ export type SignatureSetDeserialized = {
  * Verify signatures sets with batch verification or regular core verify depending on the set count.
  * Abstracted in a separate file to be consumed by the threaded pool and the main thread implementation.
  */
-export function verifySignatureSetsMaybeBatch(sets: SignatureSetDeserialized[]): boolean {
+export function verifySignatureSetsMaybeBatch(sets: SignatureSetDeserialized[], isSameMessage = false): boolean {
   if (sets.length >= MIN_SET_COUNT_TO_BATCH) {
+    if (isSameMessage) {
```
matthewkeil, line 18: MIN_SET_COUNT_BATCH is set to 2 so batches of 2 with the same message will fall through to the standard singular verify.  Would this conditional maybe want to be outside of the set.length check?

packages/beacon-node/src/chain/bls/multithread/index.ts
```diff
@@ -106,9 +125,9 @@ export class BlsMultiThreadWorkerPool implements IBlsVerifier {
 
   private readonly format: PointFormat;
   private readonly workers: WorkerDescriptor[];
-  private readonly jobs: JobQueueItem[] = [];
+  private readonly jobs: QueueItem[] = [];
   private bufferedJobs: {
```
matthewkeil, line 131: I wonder if it is worth changing how we buffer the jobs considering how much faster it will be to use the verify for the same message.  Perhaps we could use a map where the key is the message and the value is an array of jobs or something like that?  That would reduce the complexity of different cases to submit work to.  It will also capture cases where the same attestation is being verified across many network messages. What do you think?

packages/beacon-node/src/chain/bls/multithread/worker.ts
```diff
@@ -88,21 +134,14 @@ function verifyManySignatureSets(workReqArr: BlsWorkReq[]): BlsWorkResult {
...
 
-  return {
-    workerId,
-    batchRetries,
-    batchSigsSuccess,
-    workerStartNs: startNs,
-    workerEndNs: process.hrtime.bigint(),
-    results,
-  };
+  return {results, batchRetries, batchSigsSuccess};
```
matthewkeil, line 144: I am curious why you are not returning the metrics if it is still possible to send these to the workers?  Are you thinking they will always be done on main thread?  I tend to agree with @wemeetagain that the bls work should be done by the worker pool but definitely trust your opinion when it comes to cpu efficiency @tuyennhv!!

packages/beacon-node/test/perf/bls/bls.test.ts
```diff
@@ -1,20 +1,26 @@
+import crypto from "node:crypto";
 import {itBench} from "@dapplion/benchmark";
 import bls from "@chainsafe/bls";
-import type {PublicKey, SecretKey, Signature} from "@chainsafe/bls/types";
+import {CoordType, type PublicKey, type SecretKey} from "@chainsafe/bls/types";
```
matthewkeil, line 4: very cool usage of the `type` keyword here!!! never seen it used like this before!

matthewkeil (changes requested):
I am curious what you think about simplifying the API back to:

```ts
type WorkerApi = {
  verifyManySignatureSets(workReqArr: BlsWorkReq[]): Promise<BlsWorkResult>;
};
```

That way there is only a single api from the chain code perspective and the decision making about when and to check messages for being the same and to verify accordingly all happens within the `IBlsVerifier`.  We would be able to achieve the grouping of similar attestations across multiple gossip messages and I think it will simplify the worker API.

If you don't think this is a good idea let me know and I will come back and finish the review with the code as-is.

packages/beacon-node/src/chain/bls/multithread/index.ts
matthewkeil: Will PR after this if there is a possible optimization after seeing full implementation.

packages/beacon-node/src/chain/bls/multithread/index.ts
```diff
@@ -190,14 +209,56 @@ export class BlsMultiThreadWorkerPool implements IBlsVerifier {
     return results.every((isValid) => isValid === true);
   }
 
+  async verifySignatureSetsSameSigningRoot(
+    sets: ISignatureSet[],
+    opts?: Pick<VerifySignatureOpts, "verifyOnMainThread">
+  ): Promise<boolean[]> {
+    if (opts?.verifyOnMainThread && !this.blsVerifyAllMultiThread) {
+      const isSameMessage = true;
```
matthewkeil, line 219: What do you think about a metric for main thread time similar to the other case?

matthewkeil: Should we also be counting the sets and keys to see what kind of gains we get under load?

packages/beacon-node/src/chain/bls/multithread/worker.ts
twoeths: > I am curious why you are not returning the metrics if it is still possible to send these to the workers? Are you thinking they will always be done on main thread? 

for some metrics (for example latencyToworker, latencyFromWorker) we need to return results and compute on main thread
also the main thing for this PR is to add new api so I follow the same way of the old PR

packages/beacon-node/src/chain/bls/maybeBatch.ts
twoeths: if it's 1 we use the standard verify, if it's >=2 we'll aggregate pubkeys and signatures to verify

twoeths:
> I am curious what you think about simplifying the API back to:
> 
> ```ts
> type WorkerApi = {
>   verifyManySignatureSets(workReqArr: BlsWorkReq[]): Promise<BlsWorkResult>;
> };
> ```
> 
> That way there is only a single api from the chain code perspective and the decision making about when and to check messages for being the same and to verify accordingly all happens within the `IBlsVerifier`. We would be able to achieve the grouping of similar attestations across multiple gossip messages and I think it will simplify the worker API.
> 
> If you don't think this is a good idea let me know and I will come back and finish the review with the code as-is.

@matthewkeil  thanks for looking into this PR. As discussed, the part of filtering signature sets with the same message is done in #5729 so we really need a new api here. I'm happy if you have any suggestions or simpler way to achieve this.

packages/beacon-node/test/perf/bls/bls.test.ts
twoeths: thanks to IDE it helps us do it :)

packages/beacon-node/src/chain/bls/multithread/index.ts
twoeths: thanks for reminding this, all metrics are available, I only need to add "api" label there

matthewkeil (commented):
Thanks for taking the time to chat about this.  Ill definitely spend some more time looking through the draft PR implementing this over the week!!  Awesome idea 🚀 for optimization of the gossip work

packages/beacon-node/src/chain/bls/multithread/index.ts
twoeths: > Should we also be counting the sets and keys to see what kind of gains we get under load?

we counted the api call, regarding the gain we can count on `bls.test.ts` perf test

dapplion:
@tuyennhv I think I found a simpler approach extending your idea. It re-uses the existing queue, such that:
- same message batch are lazily aggregated and sent to the worker
- on failure are put back into the queue
- plus nice metrics and use of enums to split job types

Please take a look and consider incorporating

https://github.com/ChainSafe/lodestar/compare/tuyen/verifyManySignatureSetsSameMessage-lion-take?expand=1

twoeths:
> @tuyennhv I think I found a simpler approach extending your idea. It re-uses the existing queue, such that:
> 
> * same message batch are lazily aggregated and sent to the worker
> * on failure are put back into the queue
> * plus nice metrics and use of enums to split job types
> 
> Please take a look and consider incorporating
> 
> https://github.com/ChainSafe/lodestar/compare/tuyen/verifyManySignatureSetsSameMessage-lion-take?expand=1

@dapplion that's great, it looks really simpler, I created https://github.com/ChainSafe/lodestar/pull/5747

closing this PR for now

PR #5738, opened 2023-07-08
chore: misc dashboards update

dapplion:
**Motivation**

After a debugging session I've done multiple changes to dashboards:
- add missing metrics
- migrate old graph to timeline
- adjust sizes
- add avg_over_time to smooth out data

Plus I noticed that downloading the charts with 

```
node scripts/download_dashboards.mjs
```

is not deterministic due to `fieldConfig.defaults.thresholds` sometimes adding `value: null` sometimes not. Since we do not need those defaults, I've added a lint rule to delete all thresholds. If we ever need them we can review.


dashboards/lodestar_bls_thread_pool.json
```diff
@@ -599,7 +519,7 @@
       "pluginVersion": "7.4.5",
       "targets": [
         {
-          "expr": "delta(lodestar_bls_thread_pool_queue_job_wait_time_seconds_sum[$rate_interval])/delta(lodestar_bls_thread_pool_queue_job_wait_time_seconds_count[$rate_interval])",
+          "expr": "rate(lodestar_bls_thread_pool_queue_job_wait_time_seconds_sum[$rate_interval])/rate(lodestar_bls_thread_pool_queue_job_wait_time_seconds_count[$rate_interval])",
```
matthewkeil, line 524: Quick question.  I noticed that you made this change in many places.  I read the docs for both function and the major difference is the tolerance to restarts not skewing metrics.  Is that the reason you made this change?  Or was it directly related to how the calculation happens?

dashboards/lodestar_debug_gossipsub.json
```diff
@@ -183,11 +170,13 @@
             "type": "prometheus",
             "uid": "${DS_PROMETHEUS}"
           },
+          "editorMode": "code",
           "exemplar": false,
-          "expr": "lodestar_gossip_peer_score_by_threshold_count",
+          "expr": "avg_over_time( lodestar_gossip_peer_score_by_threshold_count [$rate_interval])",
```
matthewkeil, line 179: ~~For the moving averages what do you think about having a smaller $rate_interval than the window size so the graphs represent more localized values?~~

N.m.  I read more closely in the docs. This is correct!!

dashboards/lodestar_networking.json
```diff
@@ -3023,28 +2769,38 @@
...
             "uid": "${DS_PROMETHEUS}"
           },
-          "editorMode": "code",
-          "expr": "rate(lodestar_execution_engine_http_client_request_time_seconds_sum{routeId=\"notifyNewPayload\"} [$rate_interval])\n/\nrate(lodestar_execution_engine_http_client_request_time_seconds_count{routeId=\"notifyNewPayload\"} [$rate_interval])",
+          "exemplar": false,
+          "expr": "rate(lodestar_gossip_block_elapsed_time_till_processed_sum[$rate_interval])\n/\nrate(lodestar_gossip_block_elapsed_time_till_processed_count[$rate_interval])",
           "hide": false,
-          "legendFormat": "notifyNewPayload roundtrip",
-          "range": true,
+          "interval": "",
```
matthewkeil, line 2792: Out of curiosity, what does the empty `interval` here do?

dashboards/lodestar_networking.json
```diff
@@ -3425,8 +3140,8 @@
           "mode": "scheme",
           "reverse": false,
           "scale": "exponential",
-          "scheme": "Oranges",
-          "steps": 64
+          "scheme": "Magma",
```
matthewkeil, line 3151: 🔥

dashboards/lodestar_validator_monitor.json
```diff
@@ -1059,7 +1196,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "exemplar": false,
-          "expr": "32*12*rate(validator_monitor_prev_epoch_on_chain_inclusion_distance_bucket [$rate_interval])",
+          "expr": "32*12*rate(validator_monitor_prev_epoch_on_chain_inclusion_distance_bucket [$rate_interval]) - 10000",
```
matthewkeil: Interesting.  Why the scalar reduction?  Would love to learn from this!

matthewkeil: I see that this was a metric that was moved from below.  Still curious about the scalar though.  Will look when I load the dashboards to see how things change.  Feel free to learn me somethin'!!

scripts/lint-grafana-dashboard.mjs
```diff
@@ -320,10 +327,25 @@ function assertPanels(panels) {
...
         if (target.expr) {
           target.expr.replace(/\$__rate_interval/g, `$${variableNameRateInterval}`);
+
+          // ban use of delta and increase functions
+          if (target.expr.includes("delta(")) {
+            throw Error(`promql function 'delta' is not allowed, use 'rate' instead: ${target.expr}`);
+          }
+          if (target.expr.includes("increase(")) {
+            throw Error(`promql function 'increase' is not allowed, use 'rate' instead: ${target.expr}`);
+          }
```
matthewkeil, line 349: Should we also disallow "decrease"?

```suggestion
          const disallowed = ["delta", "increase", "decrease"];
          for (const disallow of disallowed) {
            if (target.expr.includes(disallow + "(")) {
              throw Error(`promql function '${disallow}' is not allowed, use 'rate' instead: ${target.expr}`);
            }
          }
```

matthewkeil (dismissed):
This is awesome!  I am really glad you started this process.  I was thinking one of the next things I attempt is to clean up the dashboards a bit and perhaps reorganize them a bit.  I will put up an issue to discuss and see how the team feels when I finish with the tasks I am working on.

dashboards/lodestar_vm_host.json
```diff
@@ -1030,7 +941,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "nodejs_eventloop_lag_seconds",
+          "expr": "nodejs_eventloop_lag_seconds{job=\"beacon\"}",
```
nflaig: Job name variable needs to be supported, else job name is not configurable

```suggestion
          "expr": "nodejs_eventloop_lag_seconds{job=~\"$beacon_job|beacon\"}",
```

**Edit:** Updated lint script to ensure this ☝️

dashboards/lodestar_networking.json
```diff
@@ -2003,28 +1830,32 @@
...
           "interval": "",
           "legendFormat": "to disconnect {{reason}}",
+          "range": true,
           "refId": "B"
         }
       ],
       "thresholds": [],
       "timeRegions": [],
-      "title": "Requested peers to connect and disconnect",
+      "title": "Requested peers to connect and disconnect (rate / minute)",
```
nflaig: consistency to other titles

```suggestion
      "title": "Requested peers to connect and disconnect (rate / min)",
```

scripts/lint-grafana-dashboard.mjs
```diff
@@ -320,10 +327,25 @@ function assertPanels(panels) {
         // Force usage of interval variable
         if (target.expr) {
           target.expr.replace(/\$__rate_interval/g, `$${variableNameRateInterval}`);
+
+          // ban use of delta and increase functions
```
nflaig: Would be great to add a rationale here on why those functions are banned

dashboards/lodestar_bls_thread_pool.json
dapplion: delta() / delta() _should_ behave like rate() / rate(). But just delta() is very different read to just rate(). Many times I been reading a chart, think this numbers are impossible! And then found out it was a delta :unamused: . For this reason I impose a strict rate() only usage across the board to keep things consistent.

dashboards/lodestar_networking.json
dapplion: No idea grafana exports add so much random crap

dashboards/lodestar_validator_monitor.json
dapplion: The contant substraction causes heatmaps to ignore the first row. This is very useful in histograms where the first bucket (inclusion distance 1) takes 99% of the values, but you want good resolution of the other values. Without this trick, non 1 values are dark, and with this they pop and are readable.

scripts/lint-grafana-dashboard.mjs
dapplion: I can't imagine anyone using decrease ever

scripts/lint-grafana-dashboard.mjs
```diff
@@ -10,6 +10,7 @@ import fs from "node:fs";
 @typescript-eslint/no-unsafe-assignment,
 @typescript-eslint/explicit-function-return-type,
 @typescript-eslint/naming-convention,
+quotes,
```
nflaig, line 13: There is a conflict between eslint and prettier if you try to defined a string such as `'job=~"$beacon_job|beacon"'`, it won't let you escape double quotes `\"` and eslint will complain if single quote or backtick is used.

dashboards/lodestar_vm_host.json
dapplion: Good point! We should add a lint rule for it

nflaig (approved):

wemeetagain (approved):

wemeetagain:

PR #5742, opened 2023-07-10
chore: remove winston from packages other than logger

nflaig:
**Motivation**

Noticed when installing @lodestar/api, it also installs winston and tons of other sub deps which are not required.

**Description**

Remove winston* dependencies from all packages other than @lodestar/logger.

### DIFF
--- a/packages/beacon-node/package.json
+++ b/packages/beacon-node/package.json
@@ -124,7 +124,6 @@
     "@lodestar/db": "^1.9.1",
     "@lodestar/fork-choice": "^1.9.1",
     "@lodestar/light-client": "^1.9.1",
-    "@lodestar/logger": "^1.9.1",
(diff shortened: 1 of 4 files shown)


packages/beacon-node/package.json
```diff
@@ -124,7 +124,6 @@
     "@lodestar/db": "^1.9.1",
     "@lodestar/fork-choice": "^1.9.1",
     "@lodestar/light-client": "^1.9.1",
-    "@lodestar/logger": "^1.9.1",
```
nflaig, line 127: @lodestar/logger should only be a dependency of @lodestar/cli, other packages use dependency injection to get access to logger (could be different implementation)

wemeetagain (approved):

wemeetagain:

PR #5747, opened 2023-07-12
feat: add verifySignatureSetsSameMessage BLS api

twoeths:
**Motivation**

- verifying signature sets of the same signing root is 6x to >10x faster than `verifyMultipleSignatures()` api
- as a preparation for #5729 where we'll validate signature sets with same signing root
- we want to leverage our worker pool to control `canAcceptWork()` (if we should process a gossip message or not) 

**Description**

- new `verifySignatureSetsSameMessage()` bls api
- the current job queue item is JobQueueItemDefault, add another job queue item JobQueueItemSameMessage
- defer the pubkey aggregation of JobQueueItemDefault when `runJob()`, take that chance to aggregate pubkeys and signatures of JobQueueItemSameMessage and transform it to a regular BlsWorkReq, see `jobItemWorkReq`
- for JobQueueItemSameMessage if it fails to verify, transform it to multiple JobQueueItemDefault and retry, see `jobItemSameMessageToMultiSet`

part of #5416


packages/beacon-node/src/chain/bls/singleThread.ts
```diff
@@ -34,6 +37,31 @@ export class BlsSingleThreadVerifier implements IBlsVerifier {
...
   }
 
+  async verifySignatureSetsSameMessage(
+    sets: {publicKey: PublicKey; signature: Uint8Array}[],
+    message: Uint8Array
+  ): Promise<boolean[]> {
+    const startNs = process.hrtime.bigint();
+    const pubkey = bls.PublicKey.aggregate(sets.map((set) => set.publicKey));
+    // validate signature = true
+    const signatures = sets.map((set) => bls.Signature.fromBytes(set.signature, CoordType.affine, true));
```
dapplion: Watch out, this can throw here and should be handled on a per-item basis

dapplion (commented):
Looks good! Can you add some tests for invalid signature format? Such that it throws an error on the fromBytes call for both worker and main thread, default and sameMessage

packages/beacon-node/src/chain/bls/multithread/index.ts
```diff
@@ -346,44 +373,111 @@ export class BlsMultiThreadWorkerPool implements IBlsVerifier {
...
+              // Create new jobs for each pubkey set, and Promise.all all the results
+              for (const j of jobItemSameMessageToMultiSet(job)) {
+                if (j.opts.priority) {
+                  this.jobs.unshift(j);
+                } else {
+                  this.jobs.push(j);
+                }
+              }
+              this.metrics?.blsThreadPool.sameMessageRetryJobs.inc(1);
+              this.metrics?.blsThreadPool.sameMessageRetrySets.inc(job.sets.length);
```
wemeetagain: Is it worth pulling this into a method since this block is repeated? eg: `retryJobItemSameMessage`

twoeths: 👍 I extracted it to new method `retryJobItemSameMessage`

twoeths:
Tested on a mainnet node subscribing to all subnets
- With `useWorker=false`, this PR does not have stable mesh peers
- With `useWorker=true`, mesh peers are stable
Screenshot 2023-07-19 at 18 01 23

- `beacon_attestation` job time is a bit better (~131ms vs ~178ms on unstable)
Screenshot 2023-07-19 at 18 02 28

- `beacon_attestation` job wait time is way better than unstable (~821ms vs ~2.5s)
Screenshot 2023-07-19 at 18 03 18

- cpu time is better than unstable (~200% vs > 300ms) because we send way less signature sets through worker threads
Screenshot 2023-07-19 at 18 06 51


we'll likely have a flag to turn this batch attestation feature on/off later (to be done on another PR)

wemeetagain (dismissed):

dapplion (approved):
Looks great! Pushed a commit with mostly cosmetic changes

wemeetagain:

PR #5752, opened 2023-07-12
fix: fix eslint for import/no-extraneous-dependencies

nazarhussain:
**Motivation**

All the packages should have valid dependencies. 

**Description**

- Fix the eslint rule `import/no-extraneous-dependencies` which was not effective for mono-repo packages. 
- Fix the package dependencies 
- Update the linting dependencies setup to work with latest Typescript 
- Fix all fixable linting errors
- Fix few useable linting warnings.

Run all tests

**NOTE: Would be better to review by commit.**


nazarhussain:
Would be better to review by commit.

matthewkeil (approved):
LGTM 🚀 

Just leaving at note for meta searches.  There were 4 strict null checks that were explicitly converted to `if (x !== null)` or vice versa for positive case.

.eslintrc.js
```diff
@@ -152,7 +156,6 @@ module.exports = {
     semi: "off",
   },
   settings: {
-    "import/internal-regex": "^@chainsafe/",
```
nflaig, line 155: why was this removed? should add some context in the PR

packages/beacon-node/src/network/core/networkCore.ts
```diff
@@ -2,15 +2,15 @@ import {PeerId} from "@libp2p/interface-peer-id";
...
 import {Connection} from "@libp2p/interface-connection";
 import {PublishOpts} from "@chainsafe/libp2p-gossipsub/types";
-import {routes} from "@lodestar/api";
 import {PeerScoreStatsDump} from "@chainsafe/libp2p-gossipsub/dist/src/score/peer-score.js";
+import {fromHexString} from "@chainsafe/ssz";
+import {ENR} from "@chainsafe/discv5";
+import {routes} from "@lodestar/api";
 import {BeaconConfig} from "@lodestar/config";
-import {LoggerNode} from "@lodestar/logger/node";
+import type {LoggerNode} from "@lodestar/logger/node";
```
nflaig, line 10: @lodestar/logger is a dependency of beacon-node, what is the reason for adding explicit type only import?

packages/beacon-node/test/perf/api/impl/validator/attester.test.ts
```diff
@@ -1,5 +1,6 @@
 import {itBench} from "@dapplion/benchmark";
 import {PointFormat} from "@chainsafe/bls/types";
+// eslint-disable-next-line import/no-relative-packages
 import {generatePerfTestCachedStatePhase0, numValidators} from "../../../../../../state-transition/test/perf/util.js";
```
nflaig, line 4: I feel like we should get rid of those cross-package relative imports. Those functions could be moved to @lodestar/test-utils and then imported from there.

nflaig (commented):
Looks good to me, just few remarks.

We should try to fix all those relative imports, they are easy to find now searching for `import/no-relative-packages`  and reusable test utils can be moved to @lodestar/test-utils.

packages/beacon-node/src/network/core/networkCore.ts
nflaig: On 2nd thought, maybe it makes sense. The only reason the beacon-node packages requires @lodestar/logger as dependency is because the workers need to initialize their own instance using `getNodeLogger`. I don't think this can be solved though as it will not be possible to inject the logger to the worker process.

.eslintrc.js
nazarhussain: First it was contextually not correct. Secondly it was making the mono-repo packages as external.

packages/beacon-node/src/network/core/networkCore.ts
nazarhussain: For type only import, it was changed before I realized that worker are requiring `getNodeLogger`.

packages/beacon-node/test/perf/api/impl/validator/attester.test.ts
nazarhussain: Yes we can do that and `@lodestar/test-utils` will be helpful in this context.

wemeetagain:

PR #5754, opened 2023-07-13
feat: prioritize api signature sets

### LINKED ISSUE #5739: Prioritize BLS signatures from APIs
### Problem description

When gossip blocks are late, node has to download block through `beacon_block_by_root`

Screenshot 2023-07-09 at 21 07 09

gossip attestations have to be reprocessed at the same time which send too many signature sets to bls worker thread pool, job length is consistently at >= 500

Screenshot 2023-07-09 at 21 12 44

which cause bls job wait time to be very high during that time

Screenshot 2023-07-09 at 21 13 08

which cause `sumitPoolAttestations` time to be very high too

Screenshot 2023-07-09 at 21 15 02

the submission was late so attestations are not included in any AggregateAndProof (below is `timely_target_no_aggregate_inclusion` type in attestation performance summary panel)

Screenshot 2023-07-09 at 21 17 54

### Solution description

root cause of the issue is gossip performance, however we could prioritize verifying bls signatures from `submitPoolAttestations` api

with https://github.com/ChainSafe/lodestar/pull/5734, it should take <7ms per slot to validate 64 signature sets of same attestation data, we could consider `verifyOnMainThread` option for `submitPoolAttestations` api

### Additional context

_No response_

twoeths:
**Motivation**

- When bls is overload, it takes a lot of time (>1s) to verify a signature
- There are so many gossip signature sets while just few api siganture sets so we should prioritize the latter

**Description**

- Enhance `LinkedList` to add `unshift()` method
- Change `jobs` queue in bls multithread to be a `LinkedList`, this makes it very cheap to do an `unshift()`
- Add `priority` flag to `verifySignatureSets` api
- Use `priority: true` from api end point to prioritize bls signatures

Closes #5739

### DIFF
--- a/packages/beacon-node/src/api/impl/beacon/pool/index.ts
+++ b/packages/beacon-node/src/api/impl/beacon/pool/index.ts
@@ -1,12 +1,12 @@
 import {routes, ServerApi} from "@lodestar/api";
 import {Epoch, ssz} from "@lodestar/types";
 import {SYNC_COMMITTEE_SUBNET_SIZE} from "@lodestar/params";
-import {validateGossipAttestation} from "../../../../chain/validation/index.js";
-import {validateGossipAttesterSlashing} from "../../../../chain/validation/attesterSlashing.js";
-import {validateGossipProposerSlashing} from "../../../../chain/validation/proposerSlashing.js";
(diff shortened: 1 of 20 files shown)


wemeetagain (approved):

wemeetagain:

PR #5758, opened 2023-07-14
feat: add IPv6 support

### LINKED ISSUE #5380: Feature: IPv6 support

With CGNAT becoming more prevalent, peering is hard for users when IPv4 incoming connections are not available.

Support IPv6 for P2P. This could be done by doing dual-stack by default, and as well by allowing one v4 and one v6 `--listenAddress`. Lighthouse for example does `--listen-address 0.0.0.0 --listen-address ::` for dual-stack, and specifying only either gets you a v4 or v6 solo stack.

**Additional context**

P2P support is the driver, here. Metrics, REST, key manager, all can stay v4, at least initially. If it's trivial to make them dual-stack as well, by all means do, but for the bulk of users where this is an issue - home users with CGNAT - they'll have v4 available and they won't need to reach REST via v6 from "the Internet"

wemeetagain:
**Motivation**

- Resolves #5380  

**Description**

- Update discv5 to v5.0.0 (which includes IPv6 support)
- Add new CLI args `--listenAddress6`, `--port6`, `--discoveryPort6`
  - `port6` and `discoveryPort6` follow the same behavior as `port` and `discoveryPort`
  - default `port6` is 9090
  - If neither `listenAddress` nor `listenAddress6` is set, a default `listenAddress` will be used
  - Else, if a `listenAddress` or `listenAddress6` is set explicitly, only those that are set will be used
  - Eg: `--port 9090` - listen on IPv4 on port 9090 for both libp2p and discv5 connections, no IPv6
  - Eg: `--listenAddress6 :: --port6 9000` - listen on IPv6 on port 9000 for both libp2p and discv5 connections, no IPv4
  - Eg: `--listenAddress 0.0.0.0 --listenAddress6 ::` - listen on IPv4 on port 9000, IPv6 on port 9090 for both libp2p and discv5 connections


packages/cli/src/options/beaconNodeOptions/network.ts
```diff
@@ -113,6 +170,27 @@ export const options: CliCommandOptions<NetworkArgs> = {
...
   },
 
+  listenAddress6: {
+    type: "string",
+    description: "The IPv6 address to listen for p2p UDP and TCP connections",
+    group: "network",
+  },
+
+  port6: {
+    description: "The TCP/UDP port to listen on. The UDP port can be modified by the --discovery-port6 flag.",
```
nflaig:
```suggestion
    description: "The TCP/UDP port to listen on. The UDP port can be modified by the --discoveryPort6 flag.",
```

For ipv4 is also references `--discovery-port` instead of `--discoveryPort`

packages/cli/src/options/beaconNodeOptions/network.ts
```diff
@@ -38,10 +43,59 @@ export type NetworkArgs = {
...
+
+  for (const key of [
+    "listenAddress",
+    "port",
+    "discoveryPort",
+    "listenAddress6",
+    "port6",
+    "discoveryPort6",
+  ] as (keyof typeof muArgs)[]) {
+    validateMultiaddrArg(muArgs, key);
```
wemeetagain, line 92: this allows us to fail fast if the user sets an incorrect ip address or port
eg: `--listenAddress ::` and `--listenAddress6 0.0.0.0` should error

packages/cli/src/options/beaconNodeOptions/network.ts
```diff
@@ -63,7 +120,7 @@ export function parseArgs(args: NetworkArgs): IBeaconNodeOptions["network"] {
       : null,
     maxPeers: maxPeers ?? defaultOptions.network.maxPeers,
     targetPeers: targetPeers ?? defaultOptions.network.targetPeers,
-    localMultiaddrs: [`/ip4/${listenAddress}/tcp/${tcpPort}`],
+    localMultiaddrs: [localMu, localMu6].filter(Boolean) as string[],
```
wemeetagain, line 123: libp2p was already configured to listen to an array of multiaddrs.
So nothing further than passing both multiaddrs here needs to be done.

packages/beacon-node/src/network/discv5/worker.ts
```diff
@@ -48,7 +48,10 @@ const config = createBeaconConfig(workerData.chainConfig, workerData.genesisVali
 const discv5 = Discv5.create({
   enr: SignableENR.decodeTxt(workerData.enr, keypair),
   peerId,
-  multiaddr: multiaddr(workerData.bindAddr),
+  bindAddrs: {
+    ip4: (workerData.bindAddrs.ip4 ? multiaddr(workerData.bindAddrs.ip4) : undefined) as Multiaddr,
+    ip6: workerData.bindAddrs.ip6 ? multiaddr(workerData.bindAddrs.ip6) : undefined,
+  },
```
wemeetagain, line 54: discv5 5.0.0 now requires bindAddrs to be passed in like so.
So we thread this through to the cli layer.

nflaig (approved):
LGTM!

Thanks for comments and outlining configuration options, behavior sounds good to me

twoeths (approved):

wemeetagain:

Issue #5761, opened 2023-07-16
BroadcastValidations tracer


g11tech:
implement - https://github.com/ethereum/beacon-APIs/pull/317
 - [x] add basic validations - https://github.com/ChainSafe/lodestar/pull/5762
 - [x] for blocks not produced via beacon, do a consensus val (state transition check)
 - [x] for gossip, also do gossip vals first before broadcasting
 - [ ] for gossip/consensus_and_equivocation 
 - [x] change warnings to errors for consensus chec

philknows:
As discussed @g11tech , we should include validations for blobs if they are not local. Not sure if you want to keep it to this issue or open a new one?

wemeetagain:
@g11tech is this issue complete?

PR #5762, opened 2023-07-16
feat: add basic broadcast validation to the block publishing

g11tech:
implement
 - https://github.com/ethereum/beacon-APIs/pull/317
Part of - https://github.com/ChainSafe/lodestar/issues/5761

TODO:
 - [x] use the new methods in validator publish calls with `none` param as default (expose via a cli option), primarily to support block publishing via fallback beacon nodes
 - [x] add a cli option to beacon node  for broadcast validation strickness - "warning" or "error" as the two options and use the accordingly in the validation part (warning as default)
 - [x] for `consensus` vals 
   - [x] for the blocks (full or blinded produced from the same beacon node) keep a map in chain to check via beacon block hash
   - for blocks not produced via beacon, do a consensus val (state transition check)
- [x] instead of throwing right now if validation is not supported only log a warnng and change to throw as feature matures

### DIFF
--- a/packages/api/src/beacon/routes/beacon/block.ts
+++ b/packages/api/src/beacon/routes/beacon/block.ts
@@ -45,6 +45,7 @@ export type BlockHeaderResponse = {
 };
 
 export enum BroadcastValidation {
+  none = "none",
   gossip = "gossip",
   consensus = "consensus",
   consensusAndEquivocation = "consensus_and_equivocation",
--- a/packages/api/src/beacon/routes/beacon/block.ts
+++ b/packages/api/src/beacon/routes/beacon/block.ts
@@ -177,7 +177,7 @@ export type Api = {
 
   publishBlockV2(
     blockOrContents: allForks.SignedBeaconBlock | SignedBlockContents,
-    opts: {broadcastValidation: BroadcastValidation}
+    opts: {broadcastValidation?: BroadcastValidation}
   ): Promise<
     ApiClientResponse<
       {
@@ -204,7 +204,7 @@ export type Api = {
 
   publishBlindedBlockV2(
     blindedBlockOrContents: allForks.SignedBlindedBeaconBlock | SignedBlindedBlockContents,
-    opts: {broadcastValidation: BroadcastValidation}
+    opts: {broadcastValidation?: BroadcastValidation}
   ): Promise<
     ApiClientResponse<
       {
@@ -256,9 +256,9 @@ export type ReqTypes = {
   getBlockHeaders: {query: {slot?: number; parent_root?: string}};
   getBlockRoot: BlockIdOnlyReq;
   publishBlock: {body: unknown};
-  publishBlockV2: {body: unknown; query: {broadcast_validation: string}};
+  publishBlockV2: {body: unknown; query: {broadcast_validation?: string}};
   publishBlindedBlock: {body: unknown};
-  publishBlindedBlockV2: {body: unknown; query: {broadcast_validation: string}};
+  publishBlindedBlockV2: {body: unknown; query: {broadcast_validation?: string}};
   getBlobSidecars: BlockIdOnlyReq;
 };
 
--- a/packages/api/src/beacon/routes/beacon/block.ts
+++ b/packages/api/src/beacon/routes/beacon/block.ts
@@ -44,6 +44,12 @@ export type BlockHeaderResponse = {
   header: phase0.SignedBeaconBlockHeader;
 };
 
+export enum BroadcastValidation {
+  gossip = "gossip",
+  consensus = "consensus",
+  consensusAndEquivocation = "consensus_and_equivocation",
+}
+
 export type Api = {
   /**
    * Get block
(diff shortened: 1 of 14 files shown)


dapplion (dismissed):
This PR changes current behaviour that ignores broadcastValidation, to rejecting calls with values != consensus. I'm not sure if it's an strictly improvement, but I can see the reasoning why. Do you think there value in merging a throw before the full actual implementation?

g11tech:
> This PR changes current behaviour that ignores broadcastValidation, to rejecting calls with values != consensus. I'm not sure if it's an strictly improvement, but I can see the reasoning why. Do you think there value in merging a throw before the full actual implementation?

added a flag to enforce strictness (mostly for dev purposes) I agree will push follow up PRs on this  tracking here: 
- https://github.com/ChainSafe/lodestar/issues/5761

currently default broadCastValdaton requested by valdator s `none` as well as defaut `strictness` on beacon s `warn`. so should not impact current behavor but allow us to add valdatons n progressve manner (as per tracker above)

packages/api/src/beacon/routes/beacon/block.ts
```diff
@@ -44,6 +44,13 @@ export type BlockHeaderResponse = {
   header: phase0.SignedBeaconBlockHeader;
 };
 
+export enum BroadcastValidation {
+  none = "none",
```
wemeetagain, line 48: according to the beacon APIs this shouldn't exist

packages/api/src/beacon/routes/beacon/block.ts
```diff
@@ -167,6 +174,20 @@ export type Api = {
       HttpStatusCode.BAD_REQUEST | HttpStatusCode.SERVICE_UNAVAILABLE
     >
   >;
+
+  publishBlockV2(
+    blockOrContents: allForks.SignedBeaconBlock | SignedBlockContents,
+    opts: {broadcastValidation: BroadcastValidation}
```
wemeetagain: broadcastValidation shouldn't be required according to beacon API spec

wemeetagain (changes requested):

packages/api/src/beacon/routes/beacon/block.ts
g11tech: updated thanks for spotting :+1:

packages/api/src/beacon/routes/beacon/block.ts
g11tech: `none` is what our current case is, just added it for convenience and incase one doesn't want to actually do any validations before publishing... may be we can remove it later

packages/beacon-node/src/api/impl/beacon/blocks/index.ts
```diff
@@ -37,6 +39,127 @@ export function getBeaconBlockApi({
...
+    switch (broadcastValidation) {
+      case routes.beacon.BroadcastValidation.none: {
+        if (blockLocallyProduced) {
+          chain.logger.debug("No broadcast validation requested for the block", valLogMeta);
+        } else {
+          chain.logger.warn("No broadcast validation requested for the block", valLogMeta);
+        }
+        break;
+      }
+      case routes.beacon.BroadcastValidation.consensus: {
```
wemeetagain, line 90: I think this PR should implement the different broadcast validations

g11tech: would prefer to do followups unless you feel strongly about this

packages/cli/src/cmds/validator/options.ts
```diff
@@ -127,6 +128,14 @@ export const validatorOptions: CliCommandOptions<IValidatorCliArgs> = {
   ...logOptions,
   ...keymanagerOptions,
 
+  broadcastValidation: {
+    // TODO: flag hidden till validations fully implemented
+    hidden: true,
+    description: "Broadcast validation to be requested before publishing the block",
+    defaultDescription: defaultOptions.broadcastValidation,
+    type: "string",
+  },
```
wemeetagain: I don't think this needs to be implemented on the validator's side? I don't think we ever want extra validations to run before publishing its blocks

g11tech: hmm.... default can be set to none whch is what its currently set, but some builder using lodestar validator to publish can provide this flag

wemeetagain: imo since this feature is specifically for relay client diversification, we shouldn't touch validator code here, and it makes it a safer PR since we know there are no validator changes.
A builder using lodestar to publish should enable broadcast validation on its own part of the stack, not in the validator.

g11tech: how about those using the DVT where dvt only sits in the middleware and the distributed validators do want to validate the block... but i guess we can leave that out of scope for now in the least. 

so just remove the broadcastValidation configuration option or not even move validator calls to produce[Block|BlindedBlock] to v2?

wemeetagain (approved):

wemeetagain:

PR #5467, opened 2023-05-05
fix(validator): log api errors as `warn` if node is syncing

nflaig:
**Motivation**

Logs mentioned in https://github.com/ChainSafe/lodestar/issues/5359#issuecomment-1529027981 should not be shown as errors on the VC if beacon node is syncing.

**Description**

`HttpError` is not thrown anymore, instead API client returns a `ApiClientResponse`

https://github.com/ChainSafe/lodestar/blob/913a51d8807d2111d1e67e50f19277e9cc953eac/packages/api/src/utils/client/client.ts#L81-L85

which will be handled by `ApiError.assert` and if not ok, will throw `ApiError`

https://github.com/ChainSafe/lodestar/blob/913a51d8807d2111d1e67e50f19277e9cc953eac/packages/api/src/utils/client/httpClient.ts#L42-L46

**Before**

```
Error: Error on getProposerDuties - Service Unavailable: Node is syncing - headSlot 5529152 currentSlot 5529417
    at Function.assert (file:///usr/app/packages/api/src/utils/client/httpClient.ts:44:13)
    at BlockDutiesService.pollBeaconProposers (file:///usr/app/packages/validator/src/services/blockDuties.ts:155:14)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at BlockDutiesService.pollBeaconProposersAndNotify (file:///usr/app/packages/validator/src/services/blockDuties.ts:129:5)
    at BlockDutiesService.runBlockDutiesTask (file:///usr/app/packages/validator/src/services/blockDuties.ts:87:9)
    at Clock.runAtMostEvery (file:///usr/app/packages/validator/src/util/clock.ts:96:7)
```

**After**

```
warn: Node is syncing - Error on getProposerDuties - Service Unavailable: Node is syncing - headSlot 5564286 currentSlot 5565452
warn: Node is syncing - Error on getProposerDuties - Service Unavailable: Node is syncing - headSlot 5564383 currentSlot 5565453
warn: Node is syncing - Error on getProposerDuties - Service Unavailable: Node is syncing - headSlot 5564479 currentSlot 5565454
```

PR #5771, opened 2023-07-18
refactor: update execution client state change logs

nflaig:
**Motivation**

These logs are meant for end users as they are info/warn/error and not debug only. The term "ExecutionEngine" is quite technical in my opinion and to avoid any confusion would recommend we just stick to "Execution client" which is the term used by most users.

The second change of this PR is the update of log level to warn from info if execution client is syncing. This better aligns it with the validator log if beacon node is syncing. We have discussed in https://github.com/ChainSafe/lodestar/pull/5467 if info or warn should be used there.

**Description**

- Update execution client state change messages
- Use warn instead of info if execution client is syncing

### DIFF
--- a/packages/beacon-node/src/execution/engine/http.ts
+++ b/packages/beacon-node/src/execution/engine/http.ts
@@ -428,19 +428,19 @@ export class ExecutionEngineHttp implements IExecutionEngine {
 
     switch (newState) {
       case ExecutionEngineState.ONLINE:
-        this.logger.info("ExecutionEngine became online");
+        this.logger.info("Execution client became online");
         break;
       case ExecutionEngineState.OFFLINE:
-        this.logger.error("ExecutionEngine went offline");
+        this.logger.error("Execution client went offline");
         break;
       case ExecutionEngineState.SYNCED:
-        this.logger.info("ExecutionEngine is synced");
+        this.logger.info("Execution client is synced");
         break;
       case ExecutionEngineState.SYNCING:
-        this.logger.info("ExecutionEngine is syncing");
+        this.logger.info("Execution client is syncing");
         break;
       case ExecutionEngineState.AUTH_FAILED:
-        this.logger.error("ExecutionEngine authentication failed");
+        this.logger.error("Execution client authentication failed");
         break;
     }
 
--- a/packages/beacon-node/src/execution/engine/http.ts
+++ b/packages/beacon-node/src/execution/engine/http.ts
@@ -437,7 +437,7 @@ export class ExecutionEngineHttp implements IExecutionEngine {
         this.logger.info("Execution client is synced");
         break;
       case ExecutionEngineState.SYNCING:
-        this.logger.info("Execution client is syncing");
+        this.logger.warn("Execution client is syncing");
         break;
       case ExecutionEngineState.AUTH_FAILED:
         this.logger.error("Execution client authentication failed");


wemeetagain (approved):
lgtm, I agree with you re the log level discussion

nazarhussain (approved):

packages/beacon-node/src/execution/engine/http.ts
```diff
@@ -428,19 +428,19 @@ export class ExecutionEngineHttp implements IExecutionEngine {
...
-        this.logger.error("ExecutionEngine went offline");
+        this.logger.error("Execution client went offline");
         break;
       case ExecutionEngineState.SYNCED:
-        this.logger.info("ExecutionEngine is synced");
+        this.logger.info("Execution client is synced");
         break;
       case ExecutionEngineState.SYNCING:
-        this.logger.info("ExecutionEngine is syncing");
+        this.logger.warn("Execution client is syncing");
```
nazarhussain, line 440: I think this should not be a warning. It's a normal scenario that will happen a lot of times and syncing is a normal process not something a user should be worried about.

nazarhussain (changes requested):

packages/beacon-node/src/execution/engine/http.ts
nflaig: My assumption here is that this is only printed out if the state changes, therefore this should not happen often during normal operation, if the EL gets into syncing mode there is degraded performance, the node might not be able to perform correct head votes / miss block proposals.

g11tech: i think warn level is good here because it will only come if execution goes into syncing from synced, which is actually not supposed to happen often once execution syncs up with the CL, and users getting notified that node won't be able to perform its duties till execution is back up

g11tech, line 440:
```suggestion
        this.logger.warn("Execution client is syncing hence client would not be able to fully participate in the network till its synced");
```

packages/beacon-node/src/execution/engine/http.ts
```diff
@@ -428,19 +428,19 @@ export class ExecutionEngineHttp implements IExecutionEngine {
...
-        this.logger.info("ExecutionEngine became online");
+        this.logger.info("Execution client became online");
         break;
       case ExecutionEngineState.OFFLINE:
-        this.logger.error("ExecutionEngine went offline");
+        this.logger.error("Execution client went offline");
         break;
       case ExecutionEngineState.SYNCED:
-        this.logger.info("ExecutionEngine is synced");
+        this.logger.info("Execution client is synced");
```
g11tech, line 437:
```suggestion
        this.logger.info("Execution client is synced and client can now participate fully in the network");
```

packages/beacon-node/src/execution/engine/http.ts
nflaig: Not really sure about adding this, we are not printing this out when the beacon node itself is syncing, my general feeling about this is that users/operators generally understand what a syncing client implies (I guess it becomes more complicated in case of EL, see my other comment below)

nflaig: > hence client would not be able to fully participate in the network till its synced

@g11tech on this topic, what does "not fully" mean in that case, I saw users complain about missed head votes which was due to EL lacking behind / unsynced.

My understanding is
1. if EL is really far behind --> missed attestations / proposals
2. if EL is close to being synced (~ 1-2 epochs behind) --> incorrect / missing head votes, att target / source votes still work, missed proposals
3. if EL is slow to process blocks (can't quite keep up), same as degraded performance as point 2

nazarhussain: I think it's the sync stats among the EL nodes, not the with the CL. And EL nodes can get unsynced and synced a lot?

nflaig: > sync stats among the EL nodes

That's an interesting topic, how does this actually work if you specify multiple ELs in `--execution.urls`.

I see issues with this
- if one beacon node has multiple ELs only one gets payload and fork choice updates, how are the fallback ELs able to sync? I thought this requires some sort of multiplexer / middleware that sends request to all ELs.
- if a execution client is connected to multiple beacon nodes it causes frequent reorgs as it might received different fork choice updates

nflaig: > EL nodes can get unsynced and synced a lot

At least in the normal case with just one execution client I have not yet seen that the EL is getting back to syncing mode if it was already synced before and has been up 24/7

g11tech: if el is not synced with the cl head then except gossip we do not participate in the network at all including attests, proposals aggregates etc. so none of validator duties

g11tech: > I think it's the sync stats among the EL nodes, not the with the CL. And EL nodes can get unsynced and synced a lot?

no they can't once synced they should stay synced with the CL head as long as CL doesn't go out of sync and then CL doesn't has to do batch updates to EL (our req/resp is batch processing) to get back to head

nflaig: > we do not participate in the network at all

In that case I think it should be fine to just print "Execution client is syncing" as warn. Maybe the scenario where all head votes are missed is related to the CL not being able to keep up with head (but close to it).

nflaig: I think warn should be better then, let's keep the change

packages/beacon-node/src/execution/engine/http.ts
nflaig: Just a note on this log "ExecutionEngine is synced", right now it is printed out any time the beacon node starts because the default state is `online` and then it goes to `synced` state if EL is synced.

Therefore, printing "can now participate" is not that accurate in that case. I would suggest we keep the logs simple for now and revisit it if we get user feedback on it or if we notice that there is confusion around what this means.

packages/beacon-node/src/execution/engine/http.ts
philknows: I'm not a fan really of long explanations in the logs. I think something more concise would be: `Execution client is syncing. Validator duties on hold.`

packages/beacon-node/src/execution/engine/http.ts
philknows: How long does it take usually for Lodestar to know whether the EL is actually in a `synced` state or if this is just the default `online` state due to beacon startup? If there's some sort of way to delay/timeout this message to ensure it actually is `synced`, that might work better?

nflaig: > How long does it take usually for Lodestar to know whether the EL is actually in a synced state

After first request to EL we know the actual status and print out a status change message

> ensure it actually is synced

We know the correct state at this point in time since we made an request to EL, "can now participate" is just unnecessary here because beacon node was just starting up and EL was likely always synced.
The message only makes sense to me if EL goes from syncing to synced.

packages/beacon-node/src/execution/engine/http.ts
nflaig: > I'm not a fan really of long explanations in the logs

Yeah, I think we should keep logs simple and use commonly understood terms. For errors that crash the process or fail a CLI command we can add more details as it is a one time message and we should help the user to fix the issue without having to guess what went wrong or having to ask in discord because the error is unclear.

> Validator duties on hold

hmm personally not a fan, would have to add this to offline state change message as well and then add another message to online, synced that indiciates that those are not "on hold" anymore.

nazarhussain (approved):

wemeetagain:

PR #5774, opened 2023-07-18
refactor(beacon-node): simplify libp2p init

wemeetagain:
**Motivation**

Looking at libp2p initialization, I noticed that the `connectToDiscv5Bootnodes` feature was bugged, and no longer functioned.
Upon fixing, I noticed an opportunity to simplify the whole thing.

**Description**

- fix `connectToDiscv5Bootnodes`
- combine the two `createNodeJsLibp2p` functions
- rename `src/network/nodejs` directory to `src/network/libp2p`


packages/beacon-node/src/network/libp2p/index.ts
```diff
@@ -1,72 +1,93 @@
...
+  disablePeerDiscovery?: boolean;
   metrics?: boolean;
   metricsRegistry?: Registry;
-  lodestarVersion?: string;
-  hideAgentVersion?: boolean;
-  mdns?: boolean;
 };
 
-export async function createNodejsLibp2p(options: Libp2pOptions): Promise<Libp2p> {
+export async function getDiscv5Multiaddrs(bootEnrs: string[]): Promise<string[]> {
```
twoeths, line 24: we may want to try/catch to handle the case user provides an invalid enr

twoeths (approved):

nflaig:
After looking at different behavior with and without `--network.connectToDiscv5Bootnodes` it is not clear to me what the expected behavior of this flag is. The description says "Attempt direct connection to discv5 bootnodes" and before my PR https://github.com/ChainSafe/lodestar/pull/5834 it also had "from network.discv5.bootEnrs option" which I removed because it didn't make sense to me.

But there seems to be two places this is used

**In case of libp2p**
https://github.com/ChainSafe/lodestar/blob/549aa03a40e4090d752e7275d3e29046e6214b4b/packages/beacon-node/src/network/libp2p/index.ts#L55
- With `connectToDiscv5Bootnodes` it always adds all bootnodes from file + from github url  + `--bootnodes` flag to `bootMultiaddrs` which is passed to `createLibp2p`
- Without `connectToDiscv5Bootnodes` it seems like `bootMultiaddrs` is always empty

**In case of discv5**
https://github.com/ChainSafe/lodestar/blob/549aa03a40e4090d752e7275d3e29046e6214b4b/packages/beacon-node/src/network/peers/discover.ts#L132
- With `connectToDiscv5Bootnodes` the enr value is decoded right at start and we call `onDiscoveredENR`(**Edit:** This seems to dial the bootnodes if they were discovered through discv5), if an invalid enr value is provided this will throw the error reported here https://github.com/ChainSafe/lodestar/issues/5833
- Without `connectToDiscv5Bootnodes` it seems to lazily connect bootnodes (**Edit:** is it connecting? I guess it just uses those nodes to discover peers, maybe someone with better discv5 understanding can clarify this), in this case we just call `addEnr` on the worker and `ENR.decodeTxt` is handled within discv5 which creates a DEBUG log but ignores the error ([worker addEnr call](https://github.com/ChainSafe/lodestar/blob/549aa03a40e4090d752e7275d3e29046e6214b4b/packages/beacon-node/src/network/discv5/worker.ts#L61), [discv5 addEnr](https://github.com/ChainSafe/discv5/blob/75304d5a067f0ec2af839a280265b3d489bd4829/src/service/service.ts#L282-L288))


Also note that the internal option `network.discv5.bootEnrs` always contains all bootnodes (file + from github url  + `--bootnodes` flag) not matter if `--network.connectToDiscv5Bootnodes` flag is set or not.
https://github.com/ChainSafe/lodestar/blob/549aa03a40e4090d752e7275d3e29046e6214b4b/packages/cli/src/cmds/beacon/handler.ts#L175-L179


There is a comment in the code which I am not sure it is accurate since we dial `discv5.bootEnrs` either way, just not on start? (**Edit:** Maybe just my misunderstanding what connect means, i.e. it means connect as peers (libp2p) and not connect as discv5 nodes to discover more peers?)
https://github.com/ChainSafe/lodestar/blob/549aa03a40e4090d752e7275d3e29046e6214b4b/packages/beacon-node/src/network/peers/discover.ts#L132-L136

When the flag was initially added it seems like the purpose was limited to "append parsed bootnode ENRs to bootMultiaddrs" which as far as I understand are the nodes we dial over libp2p.
- see https://github.com/ChainSafe/lodestar/pull/3343


Based on my understanding, the core functionality of `--network.connectToDiscv5Bootnodes` is to connect to discv5 bootnodes directly on the network (libp2p) as peers and not just for discovery.

The second behavior "connect on start" was added in this PR https://github.com/ChainSafe/lodestar/pull/3672 and it seems like  this is required as otherwise starting two nodes [using dev scripts](https://github.com/ChainSafe/lodestar/tree/549aa03a40e4090d752e7275d3e29046e6214b4b/scripts/dev) won't connect to each other, this will dial the enrs directly through libp2p as if they were discovered by discv5.
https://github.com/ChainSafe/lodestar/blob/549aa03a40e4090d752e7275d3e29046e6214b4b/packages/beacon-node/src/network/peers/discover.ts#L382

If my understanding is correct I think it makes sense to update the CLI description to indicate that it specifically means that Lodestar should dial bootnodes through/over libp2p as peers, perhaps `Attempt direct connection to discv5 bootnodes over libp2p` to make it a bit more concise.
https://github.com/ChainSafe/lodestar/blob/549aa03a40e4090d752e7275d3e29046e6214b4b/packages/cli/src/options/beaconNodeOptions/network.ts#L271-L273

wemeetagain:
Yes, `connectToDiscv5Bootnodes` is meant to establish libp2p connections with the connectable enrs (those with ip+tcp records).

It seems that there are some duplicate `connectToDiscv5Bootnodes` behavior.
We use libp2p's "bootstrap" peer discovery module to have libp2p emit peer discovery events. Which is handled in our peer discovery module, `onDiscoveredPeer`.
We use our peer discovery module to simulate emitting newly discovered ENRs, `onDiscoveredENR`.

wemeetagain:
>Without connectToDiscv5Bootnodes it seems to lazily connect bootnodes (Edit: is it connecting? I guess it just uses those nodes to discover peers, maybe someone with better discv5 understanding can clarify this)

It's adding ENRs to our discv5 kad table, which will be used when we perform a discv5 lookup, to discover more peers.

wemeetagain:

Issue #5768, opened 2023-07-17
Metrics server is listening on IPv6 interface and not shutting down


nflaig:
### Describe the bug

It looks like Lodestar beacon node metrics server starts listening on IPv6 interface.

```sh
devops@Ubuntu-2204-jammy-amd64-base:~/goerli/lodestar$ lsof -i :8008
COMMAND     PID   USER   FD   TYPE    DEVICE SIZE/OFF NODE NAME
node    2301537 devops  102u  IPv6 973623086      0t0  TCP localhost.localdomain:8008->localhost.localdomain:56578 (ESTABLISHED)
node    2301537 devops  128u  IPv6 973428666      0t0  TCP *:8008 (LISTEN)
```

```sh
devops@Ubuntu-2204-jammy-amd64-base:~/goerli/lodestar$ sudo netstat -ltnp | grep -w ':8008'
tcp6       0      0 :::8008                 :::*                    LISTEN      2301537/node
```

It is also no longer possible to shut down the process, even force closing does not help. Had to use `kill -9` to close process as just `kill` didn't do anything.

```
Jul-17 17:08:54.000[]                 info: Synced - slot: 6092144 - head: 0xe00e…5e94 - exec-block: valid(9361504 0x8772…) - finalized: 0x7710…765c:190377 - peers: 51
^CJul-17 17:09:01.307[]                 info: Stopping gracefully, use Ctrl+C again to force process exit
Jul-17 17:09:06.176[]                 info: Synced - slot: 6092145 - head: (slot -1) 0xe00e…5e94 - exec-block: valid(9361504 0x8772…) - finalized: 0x7710…765c:190377 - peers: 50
^CJul-17 17:09:10.028[]                 info: Forcing process exit
^C^C^C^C^C./lodestar: line 7: 2357120 Killed                  node --trace-deprecation --max-old-space-size=4096 ./packages/cli/bin/lodestar.js "$@
```

### Expected behavior

Metrics server should not be listening on IPv6 and beacon node should cleanly shut down when receiving SIGTERM / SIGINT.

### Steps to reproduce

Running with default network and listen options, metrics are enabled

```sh
./lodestar beacon \
    --dataDir /home/devops/goerli/data/beacon \
    --metrics \
    --execution.urls http://localhost:8551 \
    --jwt-secret /home/devops/goerli/data/jwtsecret \
    --logLevel info \
    --network goerli \
[...]

PR #5777, opened 2023-07-19
fix: metrics server listen on localhost by default

nflaig:
**Motivation**

We are currently not setting the beacon node metrics server host to localhost / 127.0.0.1 by default. This causes it to listen on IPv6 network interface and makes it externally reachable by default (unless firewall is enabled), see https://github.com/ChainSafe/lodestar/issues/5768#issuecomment-1640866352 for further details.

Metric servers should only be explicitly exposed to an external network inferface by setting `--metrics.address "0.0.0.0"`.

**Description**

Set `127.0.0.1` as default beacon node metrics address.

This is already the case for validator metrics server
https://github.com/ChainSafe/lodestar/blob/0884fdfa1b2b80e3538633cba70b0ce3585f0c5c/packages/cli/src/cmds/validator/options.ts#L17

### DIFF
--- a/packages/beacon-node/src/metrics/options.ts
+++ b/packages/beacon-node/src/metrics/options.ts
@@ -18,4 +18,5 @@ export type MetricsOptions = HttpMetricsServerOpts & {
 export const defaultMetricsOptions: MetricsOptions = {
   enabled: false,
   port: 8008,
+  address: "127.0.0.1",
 };


twoeths (approved):

dapplion (approved):

wemeetagain (approved):

wemeetagain:

PR #5778, opened 2023-07-19
refactor: align metrics and REST API server logs

nflaig:
**Motivation**

I noticed that the UX / logs of the metrics and REST API server is different.

**Description**

Align metrics and REST API server logs:
- use "started" instead of "starting" log
- print out actual host the metrics server is listening on
- more details in error log if metrics server startup fails

### DIFF
--- a/packages/beacon-node/src/metrics/server/http.ts
+++ b/packages/beacon-node/src/metrics/server/http.ts
@@ -1,4 +1,5 @@
 import http from "node:http";
+import {AddressInfo} from "node:net";
 import {Registry} from "prom-client";
 import {Logger} from "@lodestar/utils";
 import {wrapError} from "../../util/wrapError.js";
@@ -72,12 +73,17 @@ export async function getHttpMetricsServer(
 
   const activeSockets = new HttpActiveSocketsTracker(server, socketsMetrics);
 
-  const {port, address} = opts;
-  logger.info("Starting metrics HTTP server", {port, address: address ?? "127.0.0.1"});
-
   await new Promise<void>((resolve, reject) => {
-    server.once("error", reject);
-    server.listen(port, address, resolve);
+    server.once("error", (err) => {
+      logger.error("Error starting metrics HTTP server", opts, err);
+      reject(err);
+    });
+    server.listen(opts.port, opts.address, () => {
+      const {port, address: host, family} = server.address() as AddressInfo;
+      const address = `http://${family === "IPv6" ? `[${host}]` : host}:${port}`;
+      logger.info("Started metrics HTTP server", {address});
+      resolve();
+    });
   });
 
   return {


packages/beacon-node/src/metrics/server/http.ts
```diff
@@ -72,12 +73,17 @@ export async function getHttpMetricsServer(
...
-
   await new Promise<void>((resolve, reject) => {
-    server.once("error", reject);
-    server.listen(port, address, resolve);
+    server.once("error", (err) => {
+      logger.error("Error starting metrics HTTP server", opts, err);
+      reject(err);
+    });
+    server.listen(opts.port, opts.address, () => {
+      const {port, address, family} = server.address() as AddressInfo;
```
nflaig, line 82: should be safe to assume that is this `AddressInfo` in our case
- `null` can't happen since we call it in listen callback
- `string` is only the case if it is a path to the pipe or socket but we always start server with address and port

packages/beacon-node/src/metrics/server/http.ts
```diff
@@ -72,12 +73,17 @@ export async function getHttpMetricsServer(
...
+      const url = `http://${family === "IPv6" ? `[${address}]` : address}:${port}`;
```
nflaig, line 83: IPv6 is supported, this just ensures we print out a valid URL

packages/beacon-node/src/metrics/server/http.ts
```diff
@@ -72,12 +73,17 @@ export async function getHttpMetricsServer(
...
-
   await new Promise<void>((resolve, reject) => {
-    server.once("error", reject);
-    server.listen(port, address, resolve);
+    server.once("error", (err) => {
+      logger.error("Error starting metrics HTTP server", opts, err);
+      reject(err);
+    });
+    server.listen(opts.port, opts.address, () => {
+      const {port, address: host, family} = server.address() as AddressInfo;
```
nflaig, line 82: should be safe to assume that this is a `AddressInfo` in our case
- `null` can't happen since we call it in listen callback
- `string` is only the case if it is a path to the pipe or socket but we always start server with address and port

packages/beacon-node/src/metrics/server/http.ts
```diff
@@ -72,12 +73,17 @@ export async function getHttpMetricsServer(
...
+      const address = `http://${family === "IPv6" ? `[${host}]` : host}:${port}`;
```
nflaig, line 83: IPv6 is supported, this just ensures we print out a valid URL. 

Same approach is used by fastify
- https://github.com/fastify/fastify/pull/4902

wemeetagain (approved):

wemeetagain:

PR #5779, opened 2023-07-20
fix: use compressed point format for bls

twoeths:
**Motivation**

We used uncompressed signature/pubkey when passing through thread boundary because of `_wrapDeserialize` in the past but now I find it's not necessary anymore

**Description**

- Using compressed form of pubkey/signature could reduce data passing through boundary as below:

```
const PUBLIC_KEY_LENGTH_COMPRESSED = 48;
const PUBLIC_KEY_LENGTH_UNCOMPRESSED = 48 * 2;
const SIGNATURE_LENGTH_COMPRESSED = 96;
const SIGNATURE_LENGTH_UNCOMPRESSED = 96 * 2;
```

**Tested on md16**

- Forwarded messages are the same, it means most of BLS signatures are valid. Also we have an e2e test to pass signature sets to worker threads
- Attestation validation job time and job wait time are a bit better

Screenshot 2023-07-20 at 09 14 38

- vs unstable md16

Screenshot 2023-07-20 at 09 15 28

### DIFF
--- a/packages/beacon-node/src/chain/bls/multithread/index.ts
+++ b/packages/beacon-node/src/chain/bls/multithread/index.ts
@@ -128,10 +128,7 @@ export class BlsMultiThreadWorkerPool implements IBlsVerifier {
     // TODO: Allow to customize implementation
     const implementation = bls.implementation;
 
-    // Use compressed for herumi for now.
-    // THe worker is not able to deserialize from uncompressed
-    // `Error: err _wrapDeserialize`
-    this.format = implementation === "blst-native" ? PointFormat.uncompressed : PointFormat.compressed;
+    this.format = PointFormat.compressed;
     this.workers = this.createWorkers(implementation, defaultPoolSize);
 
     if (metrics) {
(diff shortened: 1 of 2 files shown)


dapplion (changes requested):
Having to decompress a point takes non trivial amount of compute time, I would bet it's more expensive than serializing an extra 48 bytes. Can you benchmark it?

twoeths:
> Having to decompress a point takes non trivial amount of compute time, I would bet it's more expensive than serializing an extra 48 bytes. Can you benchmark it?

@dapplion you're right, `serialize + deserialize` uncompressed pubkey is 4 times faster
```
 ✔ Serialize and deserialize public key, point format compressed       35231.41 ops/s    28.38377 us/op        -         25 runs   1.22 s
 ✔ Serialize and deserialize public key, point format uncompressed     154207.5 ops/s    6.484770 us/op        -        192 runs   1.75 s
```

however serializing pubkey in compressed form is the same or even a bit faster, and the deserialization happens on worker thread anyway

```
BLS ops
    ✔ Serialize public key, point format compressed                        1410358 ops/s    709.0400 ns/op        -      41542 runs   30.0 s
    ✔ Serialize public key, point format uncompressed                      1078653 ops/s    927.0820 ns/op        -      31826 runs   30.1 s
```

we should also care about the amount of data passing through thread boundary - using compressed form save 48 bytes per set, the `md16` test suggests that using compressed form of pubkey could be better, I'll deploy to other nodes to verify

twoeths:
there are inconsistent performance between nodes, closing the PR

PR #5331, opened 2023-04-01
Handle timeout errors as api errors

nflaig:
**Motivation**

Error logs due to a http timeout are hard to read as those are not handled as by `ApiError.assert` which adds more context to the error.

**Current error log**

Error message itself it not detailed, need to look into stack trace to find out where it failed

```log
Eph 0/3 0.022[]                error: Error on runEvery fn  Timeout request
Error: Timeout request
    at HttpClient.requestWithBody (file:///home/nico/projects/ethereum/lodestar/packages/api/src/utils/client/httpClient.ts:302:17)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at Object.request [as produceAttestationData] (file:///home/nico/projects/ethereum/lodestar/packages/api/src/utils/client/client.ts:69:23)
    at AttestationService.produceAttestation (file:///home/nico/projects/ethereum/lodestar/packages/validator/src/services/attestation.ts:151:17)
    at AttestationService.runAttestationTasksGrouped (file:///home/nico/projects/ethereum/lodestar/packages/validator/src/services/attestation.ts:119:36)
    at AttestationService.runAttestationTasks (file:///home/nico/projects/ethereum/lodestar/packages/validator/src/services/attestation.ts:87:7)
    at Clock.runAtMostEvery (file:///home/nico/projects/ethereum/lodestar/packages/validator/src/util/clock.ts:96:7)
```

**Expected error log**

Error message includes all details, we know it was timeout and that it happened on produce attestation data request

```log
Eph 0/1 2.002[]                error: Error on runEvery fn  Error producing attestation - Timeout request
Error: Error producing attestation - Timeout request
    at Function.assert (file:///home/nico/projects/ethereum/lodestar/packages/api/src/utils/client/httpClient.ts:44:13)
    at AttestationService.produceAttestation (file:///home/nico/projects/ethereum/lodestar/packages/validator/src/services/attestation.ts:152:14)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at AttestationService.runAttestationTasksGrouped (file:///home/nico/projects/ethereum/lodestar/packages/validator/src/services/attestation.ts:119:36)
    at AttestationService.runAttestationTasks (file:///home/nico/projects/ethereum/lodestar/packages/validator/src/services/attestation.ts:87:7)
    at Clock.runAtMostEvery (file:///home/nico/projects/ethereum/lodestar/packages/validator/src/util/clock.ts:96:7)
```

**Description**

[...]

PR #5780, opened 2023-07-20
refactor: update distributed aggregation selection error logs

nflaig:
**Motivation**

The error that distributed aggregation selection failed because client did not receive combined selection proofs before certain time in slot happens quite frequently if peers are offline and threshold (3/4) is not fulfilled.

After seeing this error quite a few times I think it should be less technical (not use operation id from spec).

**Description**

- Update error logs if client failed to receive combined selection proofs before x time of slot
- Remove unnecessary catch block, there is no scenario where we would swallow an error here, i.e. an error that is thrown after x time of slot such as timeouts but those are handled as API errors now (see https://github.com/ChainSafe/lodestar/pull/5331)


wemeetagain (approved):

wemeetagain:

Issue #5323, opened 2023-03-30
Error using Lodestar as a failover


wemeetagain:
From discord (https://discord.com/channels/593655374469660673/743858262864167062/1090175408340733962):

I have a setup where a teku validator uses Lodestar as a failover.
![image](https://user-images.githubusercontent.com/1348242/228867643-937028ca-b220-429d-8af1-964a99a05dd8.png)

however when it tries to connect, Lodestar returns this error

Chozowarrior:
also tested on Goerli with same results

Chozowarrior:
The other way around. Teku as failover for Lodestar VC does work but does display some errors in the logging: 
![image](https://user-images.githubusercontent.com/43718356/232802482-da186c72-3e7f-4419-b430-dc99fac956e6.png)

dapplion:
> The other way around. Teku as failover for Lodestar VC does work but does display some errors in the logging: ![image](https://user-images.githubusercontent.com/43718356/232802482-da186c72-3e7f-4419-b430-dc99fac956e6.png)

What version of Teku are you using? execution_optimistic should be implemented by all clients now

Chozowarrior:
> > The other way around. Teku as failover for Lodestar VC does work but does display some errors in the logging: ![image](https://user-images.githubusercontent.com/43718356/232802482-da186c72-3e7f-4419-b430-dc99fac956e6.png)
> 
> What version of Teku are you using? execution_optimistic should be implemented by all clients now

yes the otherway around issue has been resolved but the main issue (Lodestar as failover for Teku has not been resolved yet).

dapplion:
@Chozowarrior Could you share the version of the Teku beacon node that the Lodestar validator client of your screen capture? Then I can check with Teku devs to see if they are complying with the beacon API spec

Chozowarrior:
> @Chozowarrior Could you share the version of the Teku beacon node that the Lodestar validator client of your screen capture? Then I can check with Teku devs to see if they are complying with the beacon API spec

You mean validator client right? I was using Teku 23.3.1 as a validator at the time. I was also in contact with the Teku team regarding this issue. since there was also a different issue on their side because after failing to connect to Lodestar it would not successfully attest after connecting to a different failover. That was this issue in case it is of any interest https://github.com/ConsenSys/teku/pull/7016 I was in contact with Stefan Bratanov regarding this.

nflaig:
[...]

PR #5784, opened 2023-07-21
fix: prevent eventstream errors on shutdown or client aborts

### LINKED ISSUE #5783: Error: Cannot set headers after they are sent to the client
### Describe the bug

When shutting down the beacon node it can happen that there are uncaught exceptions trying to set headers on a response which is already sent / aborted /closed and no longer writable.

```
[node-1-cl-lodestar] [8460]: Eph 6/0 2.220[node-1-cl-lodestar] �[31merror�[39m: uncaughtException: Cannot set headers after they are sent to the client
Error: Cannot set headers after they are sent to the client
    at new NodeError (node:internal/errors:405:5)
    at ServerResponse.setHeader (node:_http_outgoing:648:11)
    at ServerResponse.writeHead (node:_http_server:381:21)
    at onSendEnd (/home/runner/actions-runner/_work/lodestar/lodestar/node_modules/fastify/lib/reply.js:597:7)
    at onSendHook (/home/runner/actions-runner/_work/lodestar/lodestar/node_modules/fastify/lib/reply.js:530:5)
    at fallbackErrorHandler (/home/runner/actions-runner/_work/lodestar/lodestar/node_modules/fastify/lib/error-handler.js:127:3)
    at handleError (/home/runner/actions-runner/_work/lodestar/lodestar/node_modules/fastify/lib/error-handler.js:61:5)
    at onErrorHook (/home/runner/actions-runner/_work/lodestar/lodestar/node_modules/fastify/lib/reply.js:743:5)
    at Reply.send (/home/runner/actions-runner/_work/lodestar/lodestar/node_modules/fastify/lib/reply.js:133:5)
    at defaultErrorHandler (/home/runner/actions-runner/_work/lodestar/lodestar/node_modules/fastify/lib/error-handler.js:92:9) Cannot set headers after they are sent to the client
Error: Cannot set headers after they are sent to the client
```

See [node-1-cl-lodestar.log](https://github.com/ChainSafe/lodestar/files/12111895/node-1-cl-lodestar.log) from [this sim tests run](https://github.com/ChainSafe/lodestar/actions/runs/5613234520). This error seems to happen consistently in sim tests.

There was a previous issue https://github.com/ChainSafe/lodestar/issues/5706 and an attempt to fix it https://github.com/ChainSafe/lodestar/pull/5722 but the error still happens.

### Expected behavior

No uncaught exceptions when shutting down the beacon node.

### Steps to reproduce

The issue happens consistently in sim tests.

### Additional context

This error can also happen while node is running and client aborts
- see https://github.com/ChainSafe/lodestar/issues/5323

### Operating system

Linux

unstable (6a5225f89ff3f207e12fde53155c79779c48f975)

nflaig:
**Motivation**

Closes https://github.com/ChainSafe/lodestar/issues/5783

**Description**

Before writing to the `.raw` stream it is recommended to call `.hijack`, see https://github.com/fastify/fastify/issues/3979#issuecomment-1150797373.
This prevents Fastify from sending the response and avoids "Error: Cannot set headers after they are sent to the client" errors during shutdown or client aborts.

See mention of `.hijack()` in [Lifecycle.md](https://github.com/fastify/fastify/blob/main/docs/Reference/Lifecycle.md). Further details can be found in [Reply.md#hijack](https://github.com/fastify/fastify/blob/main/docs/Reference/Reply.md#hijack)

The previous changes done in https://github.com/ChainSafe/lodestar/pull/5722 and changes of this PR will fix all the errors reported in  https://github.com/ChainSafe/lodestar/issues/5323. We might consider closing this issue as well after checking with user that reported it.

### DIFF
--- a/packages/api/src/beacon/server/events.ts
+++ b/packages/api/src/beacon/server/events.ts
@@ -18,7 +18,13 @@ export function getRoutes(config: ChainForkConfig, api: ServerApi<Api>): ServerR
         const controller = new AbortController();
 
         try {
-          // Add injected headers from other pluggins. This is required for fastify-cors for example
+          // Prevent Fastify from sending the response, this is recommended before writing to the `.raw` stream
+          // and avoids "Cannot set headers after they are sent to the client" errors during shutdown or client aborts.
+          // See https://github.com/fastify/fastify/issues/3979, https://github.com/ChainSafe/lodestar/issues/5783
+          // eslint-disable-next-line @typescript-eslint/no-floating-promises
+          res.hijack();
+
+          // Add injected headers from other plugins. This is required for fastify-cors for example
           // From: https://github.com/NodeFactoryIo/fastify-sse-v2/blob/b1686a979fbf655fb9936c0560294a0c094734d4/src/plugin.ts
           Object.entries(res.getHeaders()).forEach(([key, value]) => {
             if (value !== undefined) res.raw.setHeader(key, value);
@@ -37,9 +43,6 @@ export function getRoutes(config: ChainForkConfig, api: ServerApi<Api>): ServerR
           await new Promise<void>((resolve, reject) => {
             void api.eventstream(req.query.topics, controller.signal, (event) => {
               try {
-                // If the request is already aborted, we don't need to send any more events.
-                if (req.raw.destroyed) return;
-
                 const data = eventSerdes.toJson(event);
                 res.raw.write(serializeSSEEvent({event: event.type, data}));
               } catch (e) {


packages/api/src/beacon/server/events.ts
```diff
@@ -38,7 +38,9 @@ export function getRoutes(config: ChainForkConfig, api: ServerApi<Api>): ServerR
             void api.eventstream(req.query.topics, controller.signal, (event) => {
               try {
                 // If the request is already aborted, we don't need to send any more events.
-                if (req.raw.destroyed) return;
+                // This prevents "Cannot set headers after they are sent to the client" errors
+                // during shutdown or when client aborts abruptly.
+                if (res.raw.writable) return;
```
nazarhussain: I believe you want here if response is not writeable. 

```suggestion
                if (!res.raw.writable) return;
```

nazarhussain (changes requested):

packages/api/src/beacon/server/events.ts
nazarhussain: If req is aborted then response will not be writeable, both are strictly related. Not sure why the earlier is not working. Seems like a bug in fastify implementation.

nflaig: yeah, you are right. Was looking so many different things, completely missed that. This fixed this issue but obviously broke the eventstream api.

packages/api/src/beacon/server/events.ts
```diff
@@ -37,9 +43,6 @@ export function getRoutes(config: ChainForkConfig, api: ServerApi<Api>): ServerR
           await new Promise<void>((resolve, reject) => {
             void api.eventstream(req.query.topics, controller.signal, (event) => {
               try {
-                // If the request is already aborted, we don't need to send any more events.
-                if (req.raw.destroyed) return;
```
nflaig, line 41: Removed as it did not resolve the issue (added in https://github.com/ChainSafe/lodestar/pull/5722)

packages/api/src/beacon/server/events.ts
```diff
@@ -18,7 +18,13 @@ export function getRoutes(config: ChainForkConfig, api: ServerApi<Api>): ServerR
         const controller = new AbortController();
 
         try {
-          // Add injected headers from other pluggins. This is required for fastify-cors for example
+          // Prevent Fastify from sending the response, this is recommended before writing to the `.raw` stream
+          // and avoids "Cannot set headers after they are sent to the client" errors during shutdown or abrupt client aborts.
+          // See https://github.com/fastify/fastify/issues/3979, https://github.com/ChainSafe/lodestar/issues/5783
+          // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
nflaig, line 24: `hijack()` is not async, see [reply.js#L116](https://github.com/fastify/fastify/blob/b31b1b48a148e8fae4d65a7b8eb2d271389d02db/lib/reply.js#L116)

nazarhussain (approved):
Seems very tricky, but as per the referenced issue it seems to works.

nflaig:
I researched a bit more to make sure "hijacking" the response (`res.hijack()`) does not have any negative side-effects in our case.
Found a [great article](https://lirantal.com/blog/avoid-fastify-reply-raw-and-reply-hijack-despite-being-a-powerful-http-streams-tool/) that mentions a few risks.

The most important consideration is about not execution hooks in our case as closing the evenstream api via `res.raw.end()` does not do anything, the socket has to be force closed on the server unless client closes it.

There are custom hooks that are skipped now, namely [`onResponse`](https://github.com/ChainSafe/lodestar/blob/d2bf4473c7b6df42f2182597e2844ed6b1e02842/packages/beacon-node/src/api/rest/base.ts#L100) (which was never executed anyhow) and [`onError`](https://github.com/ChainSafe/lodestar/blob/d2bf4473c7b6df42f2182597e2844ed6b1e02842/packages/beacon-node/src/api/rest/base.ts#L106) which just created noise before by logging `error: Req req-5 eventstream error  aborted` (not a helpful/relevant error). Once the connection is established, the server-sent events do not throw any errors. There is a bug currently when validating topics as this error is never returned to client and connection is just kept open but this will be fixed by https://github.com/ChainSafe/lodestar/pull/5787.

All other hooks are called before handler is invoked, meaning CORS headers are still set by [plugin](https://github.com/ChainSafe/lodestar/blob/0e0feca25c781048d2cb2567481c01174a98832b/packages/beacon-node/src/api/rest/base.ts#L84) and [`onRequest`](https://github.com/ChainSafe/lodestar/blob/0e0feca25c781048d2cb2567481c01174a98832b/packages/beacon-node/src/api/rest/base.ts#L93) hook is still executed as well.

wemeetagain:

PR #5330, opened 2023-04-01
fix(beacon-node): cleanly exit process on graceful shutdown

nflaig:
**Motivation**

Cleanly exit beacon node on graceful shutdown without requiring `process.exit`, closes https://github.com/ChainSafe/lodestar/issues/5317

**Description**

Uses [why-is-node-running](https://github.com/mafintosh/why-is-node-running) to identify active handles. (alternativ tool: [wtfnode](https://github.com/myndzi/wtfnode))

~~Most of the time process exits but there are errors such as `Error: QUEUE_ERROR_QUEUE_ABORTED`.
In some cases the process keeps running, I guess depends on the timing...good news is that workers are all closed correctly~~

<details>
  <summary>why-is-node-running output</summary>
  <p>
There are 1411 handle(s) keeping the process running

# TTYWRAP
node:internal/async_hooks:202
node:internal/bootstrap/switches/is_main_thread:51
node:internal/bootstrap/switches/is_main_thread:170
/home/nico/projects/ethereum/lodestar/node_modules/siginfo/index.js:3               - if (process.stderr.isTTY || force === true) {
/home/nico/projects/ethereum/lodestar/node_modules/why-is-node-running/include.js:3 - require('siginfo')(why, true)
node:internal/modules/cjs/loader:1159
node:internal/modules/cjs/loader:1213
node:internal/modules/cjs/loader:1037

# FILEHANDLE
node:internal/async_hooks:202

# FILEHANDLE
node:internal/async_hooks:202

# FILEHANDLE
node:internal/async_hooks:202

... more node:internal/async_hooks:202

# TTYWRAP
node:internal/async_hooks:202
node:internal/bootstrap/switches/is_main_thread:51
node:internal/bootstrap/switches/is_main_thread:148
node:internal/bootstrap/loaders:189
node:internal/bootstrap/loaders:311
node:internal/bootstrap/loaders:291
node:internal/bootstrap/loaders:296

# TTYWRAP
node:internal/async_hooks:202
node:internal/bootstrap/switches/is_main_thread:196
node:internal/bootstrap/loaders:189
node:internal/bootstrap/loaders:311
node:internal/bootstrap/loaders:291
node:internal/bootstrap/loaders:296
node:internal/bootstrap/loaders:276

# DNSCHANNEL
node:internal/async_hooks:202
node:internal/dns/utils:79
node:internal/dns/utils:74
node:internal/dns/utils:241
node:internal/dns/utils:251
node:internal/dns/promises:367
[...]

PR #5786, opened 2023-07-21
fix: gracefully terminate connections when closing http server

nflaig:
**Motivation**

While looking into https://github.com/ChainSafe/lodestar/issues/5783 I noticed that our REST API server is not really "friendly" to clients, it abruptly terminates connections. 

We should give the beacon node some time to complete pending requests and only forcefully close connections if timeout is reached. This improves client experience and prevents potential data corruption if request is writing data.

**Description**

Gracefully close HTTP server by waiting for all connections to drain until timeout
- it immediately destroys all sockets without an attached HTTP request
- it allows graceful timeout to sockets with ongoing HTTP requests
- it informs connections using keep-alive that server is shutting down by setting a `Connection: close` header
- it forcefully terminates all connections after timeout (1 second) is reached

**Further considerations**

It might make sense to close server immediately to stop accepting new connections.

Right now, what happens is that while terminating, new connections are just force closed by destroying the socket.
https://github.com/ChainSafe/lodestar/blob/485a996a1c2244015c40ab0c8f576ad790b79f7a/packages/beacon-node/src/api/rest/activeSockets.ts#L28-L29

This causes `curl: (56) Recv failure: Connection reset by peer` error on the client.

If the server is closed immediately it will cause `curl: (7) Failed to connect to localhost port 9596 after 0 ms: Connection refused` on the client as server stops listening for connections.

The difference is likely insignificant in our case as the shutdown period is really short and the beacon node generally does not receive that many requests per second and in both cases the client will receive an error.

This has been discussed in http-terminator repository already https://github.com/gajus/http-terminator/issues/22, but since our current and now updated implementation follows that of http-terminator closely, I kept the implementation to close server after terminating sockets.

### DIFF
--- a/packages/beacon-node/src/api/rest/base.ts
+++ b/packages/beacon-node/src/api/rest/base.ts
@@ -153,12 +153,14 @@ export class RestApiServer {
 
     // In NodeJS land calling close() only causes new connections to be rejected.
     // Existing connections can prevent .close() from resolving for potentially forever.
-    // In Lodestar case when the BeaconNode wants to close we will just abruptly terminate
-    // all existing connections for a fast shutdown.
+    // In Lodestar case when the BeaconNode wants to close we will attempt to gracefully
+    // close all existing connections but forcefully terminate after timeout for a fast shutdown.
     // Inspired by https://github.com/gajus/http-terminator/
-    this.activeSockets.destroyAll();
+    await this.activeSockets.terminate();
 
     await this.server.close();
+
+    this.logger.debug("REST API server closed");
   }
 
   /** For child classes to override */
--- a/packages/beacon-node/src/api/rest/base.ts
+++ b/packages/beacon-node/src/api/rest/base.ts
@@ -29,11 +29,6 @@ export type RestApiServerMetrics = SocketMetrics & {
   errors: IGauge<"operationId">;
 };
 
-enum Status {
-  Listening = "listening",
-  Closed = "closed",
-}
-
 /**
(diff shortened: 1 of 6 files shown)


packages/beacon-node/src/api/rest/base.ts
```diff
@@ -106,8 +99,7 @@ export class RestApiServer {
     server.addHook("onError", async (req, _res, err) => {
       // Don't log ErrorAborted errors, they happen on node shutdown and are not useful
       // Don't log NodeISSyncing errors, they happen very frequently while syncing and the validator polls duties
-      // Don't log eventstream aborted errors if server instance is being closed on node shutdown
-      if (err instanceof ErrorAborted || err instanceof NodeIsSyncing || this.status === Status.Closed) return;
```
nflaig, line 110: I added this in https://github.com/ChainSafe/lodestar/pull/5330 to avoid noisy errors on shutdown, however eventstream aborts are now properly handled and no more error are logged on shutdown.

Since we attempt to gracefully close connections now, it is highly unlikely to log any errors on shutdown since we should only get `ErrorAborted` which is already checked and not logged. Any error that is not caught here is unexpected and should be logged.

dapplion (approved):
Looks good!

wemeetagain (approved):
:+1:

matthewkeil (commented):
Post merge review.  This is great!!!

wemeetagain:

Issue #5128, opened 2023-02-10
API SSZ response


wemeetagain:
**Is your feature request related to a problem? Please describe.**
- See https://github.com/ethereum/beacon-APIs/issues/250

**Describe the solution you'd like**

Systematically support returning ssz-encoded responses throughout all/most API endpoints.
Systematically return the [Eth-Consensus-Version](https://github.com/ethereum/beacon-APIs/blob/093890f163788d9f78cbf0b8cd529e524cf6fe48/apis/validator/block.v2.yaml#L43-L45) header 

Figure out how to expose encoding choice + raw response in our api client. (Take notes from fetch API?)

**Additional context**

There are a benefit from using ssz across the board in production. SSZ encoding is more performant than JSON encoding.

We also already have some consumers that are requesting this feature.

maschad:
The [getReqSerializers()](https://github.com/ChainSafe/lodestar/blob/0006d8662d4e082a85496235b6b5431189776bbb/packages/api/src/beacon/routes/debug.ts#L147) function checks the headers for the requested format in the header. Given this function is implemented across a variety of other clients I am thinking it should be an interface that does this check and then based on that [generates a JSON client](https://github.com/ChainSafe/lodestar/blob/0006d8662d4e082a85496235b6b5431189776bbb/packages/api/src/utils/client/client.ts#L52) or an SSZ client. Wdty @nazarhussain @wemeetagain @dapplion

nazarhussain:
I suggest API clients should not be encoding specific but each endpoint. I can initiate a client, and for some endpoints I need SSZ and for some I need JSON. 

We have one endpoint already implemented with this approach. Check this reference. 

https://github.com/ChainSafe/lodestar/blob/0006d8662d4e082a85496235b6b5431189776bbb/packages/api/src/beacon/routes/debug.ts#L84-L98

The second argument from the API client should be passed as HEADER to the request and than parsed by **ReqeustSerializers**.

dapplion:
@nazarhussain I imagine an API ergonomics like this would be nice:

```ts
api.beacon.getGenesis() // queries with mime-type json
api.beacon.ssz.getGenesis() // queries with mime-type ssz
```
WDYT

nazarhussain:
My preferences would be 

```ts

api.beacon.getGenesis() // queries with mime-type json
api.beacon.getGenesis('ssz') // queries with mime-type ssz
```

For the following reasons:

[...]

PR #5790, opened 2023-07-23
feat: add syncing_status query parameter to eth/v1/node/health

### LINKED ISSUE #5702: Add `syncing_status` query parameter to health API
### Problem description

Latest beacon API spec requires `syncing_status` query parameter
- https://github.com/ethereum/beacon-APIs/pull/251

### Solution description

Add syncing_status query parameter to health API
- https://ethereum.github.io/beacon-APIs/#/Node/getHealth

### Additional context

_No response_

nflaig:
**Motivation**

Closes https://github.com/ChainSafe/lodestar/issues/5702

**Description**

- Adds `syncing_status` query parameter to `eth/v1/node/health`
- Fixes current issue of always returning 200 status code irrespective of node sync status
- Returns status code based on node sync status
  - 200: Node is ready  
  - 206: Node is syncing but can serve incomplete data
  - 100-599: Customized syncing status while node is syncing
  - 400: Invalid syncing status code provided in query parameter
  - 503:  Lodestar does not start its API until fully initialized, so this status can never be served

### DIFF
--- a/packages/api/src/beacon/routes/node.ts
+++ b/packages/api/src/beacon/routes/node.ts
@@ -176,7 +176,9 @@ export function getReqSerializers(): ReqSerializers<Api, ReqTypes> {
     getNodeVersion: reqEmpty,
     getSyncingStatus: reqEmpty,
     getHealth: {
-      writeReq: (options) => ({query: {syncing_status: options?.syncingStatus}}),
+      writeReq: (options) => ({
+        query: options?.syncingStatus !== undefined ? {syncing_status: options.syncingStatus} : {},
+      }),
       parseReq: ({query}) => [{syncingStatus: query.syncing_status}],
       schema: {query: {syncing_status: Schema.Uint}},
     },
--- a/packages/api/src/beacon/routes/node.ts
+++ b/packages/api/src/beacon/routes/node.ts
@@ -70,6 +70,10 @@ export enum NodeHealth {
   NOT_INITIALIZED_OR_ISSUES = HttpStatusCode.SERVICE_UNAVAILABLE,
 }
 
+export type NodeHealthOptions = {
+  syncingStatus?: number;
+};
+
 /**
  * Read information about the beacon node.
  */
@@ -120,10 +124,10 @@ export type Api = {
   /**
    * Get health check
    * Returns node health status in http status codes. Useful for load balancers.
-   *
-   * NOTE: This route does not return any value
    */
-  getHealth(): Promise<
+  getHealth(
+    options?: NodeHealthOptions
+  ): Promise<
     ApiClientResponse<
       {[HttpStatusCode.OK]: void; [HttpStatusCode.PARTIAL_CONTENT]: void},
       HttpStatusCode.SERVICE_UNAVAILABLE
@@ -150,7 +154,7 @@ export type ReqTypes = {
   getPeerCount: ReqEmpty;
   getNodeVersion: ReqEmpty;
   getSyncingStatus: ReqEmpty;
-  getHealth: ReqEmpty;
+  getHealth: {query: {syncing_status?: number}};
 };
 
 export function getReqSerializers(): ReqSerializers<Api, ReqTypes> {
@@ -171,7 +175,11 @@ export function getReqSerializers(): ReqSerializers<Api, ReqTypes> {
     getPeerCount: reqEmpty,
     getNodeVersion: reqEmpty,
     getSyncingStatus: reqEmpty,
-    getHealth: reqEmpty,
+    getHealth: {
+      writeReq: (options) => ({query: {syncing_status: options?.syncingStatus}}),
+      parseReq: ({query}) => [{syncingStatus: query.syncing_status}],
+      schema: {query: {syncing_status: Schema.Uint}},
+    },
   };
 }
 
--- a/packages/api/src/beacon/routes/node.ts
+++ b/packages/api/src/beacon/routes/node.ts
@@ -130,7 +130,7 @@ export type Api = {
   ): Promise<
     ApiClientResponse<
       {[HttpStatusCode.OK]: void; [HttpStatusCode.PARTIAL_CONTENT]: void},
-      HttpStatusCode.SERVICE_UNAVAILABLE
+      HttpStatusCode.BAD_REQUEST | HttpStatusCode.SERVICE_UNAVAILABLE
     >
   >;
 };
--- a/packages/api/src/interfaces.ts
+++ b/packages/api/src/interfaces.ts
@@ -23,12 +23,11 @@ export type ApiClientResponse<
 
 export type ApiClientResponseData<T extends ApiClientResponse> = T extends {ok: true; response: infer R} ? R : never;
 
-export type GenericRequestObject = Record<string, unknown>;
-export type GenericResponseObject = {code: (code: number) => void};
+export type GenericOptions = Record<string, unknown>;
 
 export type ServerApi<T extends Record<string, APIClientHandler>> = {
   [K in keyof T]: (
-    ...args: [...args: Parameters<T[K]>, req?: GenericRequestObject, res?: GenericResponseObject]
+    ...args: [...args: Parameters<T[K]>, opts?: GenericOptions]
   ) => Promise<ApiClientResponseData<Resolves<T[K]>>>;
 };
 
--- a/packages/api/src/interfaces.ts
+++ b/packages/api/src/interfaces.ts
@@ -4,7 +4,8 @@ import {Resolves} from "./utils/types.js";
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
 export type APIClientHandler = (...args: any) => PromiseLike<ApiClientResponse>;
-export type APIServerHandler = (...args: any) => PromiseLike<ApiServerResponse<ApiClientResponse>>;
+export type APIServerHandler = (...args: any) => PromiseLike<unknown>;
+
 export type ApiClientSuccessResponse<S extends keyof any, T> = {ok: true; status: S; response: T; error?: never};
 export type ApiClientErrorResponse<S extends Exclude<HttpStatusCode, HttpSuccessCodes>> = {
   ok: false;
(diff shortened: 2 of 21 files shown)


packages/api/src/utils/server/genericJsonServer.ts
```diff
@@ -42,7 +42,7 @@ export function getGenericJsonServer<
 
       handler: async function handler(this: FastifyInstance, req, resp): Promise<unknown | void> {
         const args: any[] = routeSerdes.parseReq(req as ReqGeneric as ReqTypes[keyof Api]);
-        const data = (await api[routeId](...args)) as Resolves<Api[keyof Api]>;
+        const data = (await api[routeId](...args, req, resp)) as Resolves<Api[keyof Api]>;
```
nflaig: Types are already correct but request / response objects were never passed in here.

packages/beacon-node/src/api/impl/node/index.ts
```diff
@@ -66,10 +66,18 @@ export function getNodeApi(
       return {data: sync.getSyncStatus()};
     },
 
-    async getHealth(_req, res) {
+    async getHealth(options, _req, res) {
```
dapplion: Uhmm I didn't noticed that request and response objects where being leaked to the API implementation. I see it was introduced by https://github.com/ChainSafe/lodestar/pull/4994 but why not keep the interaction in the Fastify's response object in `packages/api/src/beacon/server/node.ts`?

nflaig: Looked at this as well but the only "clean way" to achieve this without too many type hacks is approach used in https://github.com/ChainSafe/lodestar/pull/3349 but in this case the `getHealth` type is incorrect, it doesn't actually return a `res.response` of type `number`.

If you think passing response object to API implementation is not desired we should update types accordingly and use the other approach to implementing this.

dapplion: @nazarhussain any opinions here?

packages/api/src/utils/client/client.ts
```diff
@@ -80,14 +80,17 @@ export function generateGenericJsonClient<
         }
       } catch (err) {
         if (err instanceof HttpError) {
-          return {ok: false, error: {code: err.status, message: err.message, operationId: routeId}} as ReturnType<
-            Api[keyof Api]
-          >;
+          return {
+            ok: false,
+            status: err.status,
```
nflaig, line 85: Assuming this was just missed previously as the code here is not really type-safe, didn't find a quick solution to fix that but we should always set status in client response.

packages/api/src/utils/client/client.ts
```diff
@@ -80,14 +80,17 @@ export function generateGenericJsonClient<
...
+            error: {code: err.status, message: err.message, operationId: routeId},
+          } as ReturnType<Api[keyof Api]>;
         }
 
         if (err instanceof TimeoutError) {
           return {
             ok: false,
+            status: HttpStatusCode.INTERNAL_SERVER_ERROR,
```
nflaig, line 93: This status code is kinda wrong but 408 would also be wrong as it would indicate a server timeout whereas this error happens due to client timeout.

If server doesn't respond within client timeout we can assume a server issue, setting 500 here is the most accurate. Alternative would be to just throw the error here but this makes the error logs less informative if a timeout happens.
- see https://github.com/ChainSafe/lodestar/pull/5331

packages/api/src/utils/client/client.ts
```diff
@@ -26,7 +26,7 @@ export function getFetchOptsSerializer<Fn extends (...args: any) => any, ReqType
     return {
       url: urlFormater(req.params ?? {}),
       method: routeDef.method,
-      query: req.query,
+      query: Object.keys(req.query ?? {}).length ? req.query : undefined,
```
nflaig, line 29: Fixes appending '?' to URL if query params are omitted, e.g. `/eth/v1/node/health` instead of `/eth/v1/node/health?`.

I would assume most/(all) HTTP servers will ignore this anyways, but it might be misleading if req.url is logged on the server.

wemeetagain (dismissed):
lgtm!

nflaig:
Would like to get @nazarhussain thoughts on https://github.com/ChainSafe/lodestar/pull/5790#discussion_r1272059306 first before we merge this. Putting as draft for now. Thanks for all the reviews :)

packages/beacon-node/src/api/impl/node/index.ts
nazarhussain: Passing `res` directly to API handler is a pattern which does not meant it's leaking anything. 

We can define a data structure for the implementation of API endpoints, that way we can handle the status code at the server implementation side. 

So all implementation handlers should return; 
```
{
   body: unknown;
   status?: number; 
}
```

And if `status` is not present then we consider `200` as default.

packages/api/src/beacon/routes/node.ts
```diff
@@ -171,7 +175,13 @@ export function getReqSerializers(): ReqSerializers<Api, ReqTypes> {
     getPeerCount: reqEmpty,
     getNodeVersion: reqEmpty,
     getSyncingStatus: reqEmpty,
-    getHealth: reqEmpty,
+    getHealth: {
+      writeReq: (options) => ({
+        query: options?.syncingStatus !== undefined ? {syncing_status: options.syncingStatus} : {},
+      }),
+      parseReq: ({query}) => [{syncingStatus: query.syncing_status}],
+      schema: {query: {syncing_status: Schema.Uint}},
```
nazarhussain, line 183: I would recommend moving the range check in the schema.

packages/beacon-node/src/api/impl/node/index.ts
nflaig: Health API is really the only one right now that needs to modify the status code.

The interface of response is limited to setting the status code
https://github.com/ChainSafe/lodestar/blob/7b5fc63e678c4d9010b8b6993a17a8db84667013/packages/api/src/interfaces.ts#L27

packages/api/src/beacon/routes/node.ts
nflaig: Do you have an example API that does this? Based on current types this is not possible as far as I can see, we only support pre-defined schemas
https://github.com/ChainSafe/lodestar/blob/a12771447337de2189bce31846ad4b11f621ec28/packages/api/src/utils/schema.ts#L25

We enforce basic validation through schema but the error response and logging is not ideal in my opinion

### JSON API response

Schema validation error

```json
[
  {
    "instancePath": "/syncing_status",
    "schemaPath": "#/properties/syncing_status/minimum",
    "keyword": "minimum",
    "params": {
      "comparison": ">=",
      "limit": 0
    },
    "message": "must be >= 0"
  }
]
```

API error

```json
{
  "statusCode": 400,
  "error": "Bad Request",
  "message": "Invalid syncing status code: 600"
}
```

### Error log

Schema validation error

```
Eph 0/2 1.126[rest]            error: Req req-o getHealth error  querystring/syncing_status must be >= 0
Error: querystring/syncing_status must be >= 0
    at defaultSchemaErrorFormatter (/home/nico/projects/ethereum/lodestar/node_modules/fastify/lib/context.js:108:10)
    at wrapValidationError (/home/nico/projects/ethereum/lodestar/node_modules/fastify/lib/validation.js:228:17)
    at validate (/home/nico/projects/ethereum/lodestar/node_modules/fastify/lib/validation.js:157:16)
    at preValidationCallback (/home/nico/projects/ethereum/lodestar/node_modules/fastify/lib/handleRequest.js:92:25)
    at handler (/home/nico/projects/ethereum/lodestar/node_modules/fastify/lib/handleRequest.js:76:7)
    at handleRequest (/home/nico/projects/ethereum/lodestar/node_modules/fastify/lib/handleRequest.js:24:5)
    at runPreParsing (/home/nico/projects/ethereum/lodestar/node_modules/fastify/lib/route.js:569:5)
    at next (/home/nico/projects/ethereum/lodestar/node_modules/fastify/lib/hooks.js:179:7)
    at handleResolve (/home/nico/projects/ethereum/lodestar/node_modules/fastify/lib/hooks.js:196:5)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

API error

```
Eph 0/0 5.017[rest]             warn: Req req-9 getHealth failed reason=Invalid syncing status code: 600
```

In both cases as it is right now, it is better to use API error. Schema validation could be made more flexible but there are not that many values in the spec that enforce things like minimum / maximum.

nflaig:
Noticed something else while testing, right now we always return a json body in the response.

```sh
> curl -s http://localhost:9596/eth/v1/node/health
{}
```

In case of success, this is due to default behavior if no returnType is defined
https://github.com/ChainSafe/lodestar/blob/a12771447337de2189bce31846ad4b11f621ec28/packages/api/src/utils/server/genericJsonServer.ts#L51-L55

And in case of errors, we return them in response body as well
```sh
> curl -s http://localhost:9596/eth/v1/node/health?syncing_status=600 | jq
{
  "statusCode": 400,
  "error": "Bad Request",
  "message": "Invalid syncing status code: 600"
}
```

At least for the error response, I think this is even a good thing as just returning the status code is not that helpful.

packages/beacon-node/src/api/impl/node/index.ts
nazarhussain: In that case we can make the response status code structure as optional. e.g.

```
() => T | {body: T, status; number}
```

This will require no refactor on API implementation side and minimum change on the server side.

nflaig: True, that could be a good pattern to give some API implementations more control over the response. Maybe we want to allow to set `null` explicitly as body as well, that way we can more closely follow the spec for APIs that should not return a json body (see https://github.com/ChainSafe/lodestar/pull/5790#issuecomment-1648412274).

Proposed type would look like this then

```ts
() => T | {body: T | null, status: number}
```

nazarhussain: So will you make the change in this PR or separate?

nflaig: Better to do it in this PR, allows to remove the workaround in tests, should reduce overall diff

packages/api/src/utils/server/genericJsonServer.ts
nflaig: Types are updated and no longer support passing in request / response objects to API implementation

packages/beacon-node/src/api/impl/node/index.ts
nflaig: Request and response objects are no longer passed to API implementation. After looking into updating return type to something like `() => T | {body: T | null, status: number}` it does not look that clean and requires too many type checks in other place unrelated to health API.

Better to address in a separate PR.

nflaig:
https://github.com/ChainSafe/lodestar/pull/5790#issuecomment-1648412274 has been addressed now by updating the handler.

```sh
> curl -s http://localhost:9596/eth/v1/node/health
```

No longer returns a response body. Errors will still return a response body with more error details.

packages/api/src/interfaces.ts
```diff
@@ -18,17 +18,16 @@ export type ApiClientResponse<
...
 export type ApiClientResponseData<T extends ApiClientResponse> = T extends {ok: true; response: infer R} ? R : never;
 
-export type GenericRequestObject = Record<string, unknown>;
-export type GenericResponseObject = {code: (code: number) => void};
+export type GenericOptions = Record<string, unknown>;
 
 export type ServerApi<T extends Record<string, APIClientHandler>> = {
   [K in keyof T]: (
-    ...args: [...args: Parameters<T[K]>, req?: GenericRequestObject, res?: GenericResponseObject]
+    ...args: [...args: Parameters<T[K]>, opts?: GenericOptions]
```
nflaig, line 30: I removed req / res but added generic opts instead, we have API implementations that pass additional options, like `publishBlock`. Might not be a bad thing in general to allow this as API implementations are called internally as well.

https://github.com/ChainSafe/lodestar/blob/19b723c55042090bcb0442f83002ca129e37817d/packages/beacon-node/src/api/impl/beacon/blocks/index.ts#L202

nflaig:
@nazarhussain Thanks for suggesting an implementation on how we could resolve the issue of setting the status code. This is really similar to what I attempted as well locally but it creates too much noise in my opinion just for solving an issue of a single API.

I think if we want to implement a solution that gets rid of all custom route handlers, e.g. getState or getStateProof API

https://github.com/ChainSafe/lodestar/blob/19b723c55042090bcb0442f83002ca129e37817d/packages/api/src/beacon/server/debug.ts#L22

https://github.com/ChainSafe/lodestar/blob/19b723c55042090bcb0442f83002ca129e37817d/packages/api/src/beacon/server/proof.ts#L19

To achieve this we would have to give the API implementation the option to return a "raw" response, no modification at all in generic handler, skip calls like `returnType.toJson` which restricts the API to only return JSON.

https://github.com/ChainSafe/lodestar/blob/19b723c55042090bcb0442f83002ca129e37817d/packages/api/src/utils/server/genericJsonServer.ts#L52

I went for solution now which only modifies the health API and does some minor clean up of other things but I really want to keep the diff of this PR smaller.

Another refactoring PR might make sense which addresses more than just the requirements of getHealth API or we could just revisit this when implementing https://github.com/ChainSafe/lodestar/issues/5128.

nazarhussain:
I don't see the change as noise rather a feature to make the API implementation flexible. It's upto us we want that feature, and either we want that feature in this PR or a separate PR. 

I would not suggest to remove the support for custom handlers, it's a flexibility that we may need anytime if not now.

nflaig:
I looked at this for like 4 hours, tried different type and different places how to modify the type. Nothing seemed like a good solution to me, for example having to add this `const data = "status" in apiRes ? apiRes.response?.data : apiRes.data;` in 4-5 other places doesn't seem right to me.

I would like a solution that addresses more than just the status code issue of health API. For this PR, I feel like 2 type casts with comments that explain why it is done seems alright to me (health API is just a bit of an outlier). It also keeps the changes this of this PR isolated to health API, I noticed that forwarding request for example has unwanted side-effects as we override the options some API implementations like `publishBlock` add.

Also considering that health API is broken at the moment (always returns 200) and there are now tests that ensure it works correctly reduces the risk of a regression due to missing type safety by casting the types.

nflaig:
> I would not suggest to remove the support for custom handlers

Shouldn't remove support for them but rather make them less needed. Still need to consider the benefit of this but having the API implementation in a single place seems beneficial to me.

wemeetagain (approved):

wemeetagain:

PR #5794, opened 2023-07-24
feat: poll proposer duties of next epoch in advance

### LINKED ISSUE #5792: Poll block proposer duties before an epoch starts
### Problem description

Right now Lodestar has a risk of missing block proposal at start slot of epoch since it has to poll for proposer duties (and due to I/O lag it may take up to 2s for it which make the block proposal very late into the slot)

### Solution description

We already have `PrepareNextSlotScheduler` run at 2/3 of slot, if we can make sure it runs in less than 2s consistently we can poll for block proposer duties for next epoch 2s before the slot

One blocker for this is https://github.com/ChainSafe/lodestar/issues/5409

### Additional context

_No response_

twoeths:
**Motivation**

We have some missed block proposal at start slot of epoch because there are some delay (up to 2s polling proposer duties right at start slot of epoch

```
Jul-23 12:06:49.167[]                debug: Detected new block proposer currentSlot=6940832
```

in this example, we wasted 2s after slot starts and the block was missed

**Description**

- We already have `PrepareNextSlotScheduler` to run state transition 1/3 slot before next epoch so we can poll block proposer duties 1s-2s before the next epoch start
- A lot of time state transition finishes in 3s, so I defined the look ahead time as 1s before next epoch start. In the future when we improve state transition, we can tweak this number

Closes #5792

### DIFF
--- a/packages/beacon-node/src/api/impl/validator/index.ts
+++ b/packages/beacon-node/src/api/impl/validator/index.ts
@@ -405,11 +405,11 @@ export function getValidatorApi({
       const head = chain.forkChoice.getHead();
       let state: CachedBeaconStateAllForks | undefined = undefined;
       const prepareNextSlotLookAheadMs = (config.SECONDS_PER_SLOT * 1000) / SCHEDULER_LOOKAHEAD_FACTOR;
-      const cpState = chain.regen.getCheckpointStateSync({rootHex: head.blockRoot, epoch});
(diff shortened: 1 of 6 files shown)


twoeths:
was able to request proposer duties for next epoch consistently on goerli

Screenshot 2023-07-25 at 12 39 28

wemeetagain (approved):

wemeetagain:

PR #5797, opened 2023-07-24
fix: offline error message when node is shutting down

nazarhussain:
**Motivation**

There should not be a false error log. 

**Description**

Currently when node is shutting down, it shows an error message that EL engine is turned offline, which is not the valid case. 

- Run the beacon-node
- Wait for log message to show EL engine is syncing
- Shutdown the beacon-node


packages/beacon-node/src/execution/engine/http.ts
```diff
@@ -131,9 +129,13 @@ export class ExecutionEngineHttp implements IExecutionEngine {
       this.updateEngineState(ExecutionEngineState.ONLINE);
       return res;
     } catch (err) {
-      if (!isErrorAborted(err)) {
-        this.updateEngineState(getExecutionEngineState({payloadError: err}));
+      const newState = getExecutionEngineState({payloadError: err, oldState: this.state});
```
nflaig: why does this return `undefined` at all? Wouldn't it be better to just return the oldState if nothing changed?

packages/beacon-node/src/execution/engine/http.ts
```diff
@@ -205,22 +207,11 @@ export class ExecutionEngineHttp implements IExecutionEngine {
...
-          this.updateEngineState(getExecutionEngineState({payloadError: e}));
-        }
-        if (e instanceof HttpRpcError || e instanceof ErrorJsonRpcResponse) {
-          return {status: ExecutePayloadStatus.ELERROR, latestValidHash: null, validationError: e.message};
-        } else {
-          return {status: ExecutePayloadStatus.UNAVAILABLE, latestValidHash: null, validationError: e.message};
-        }
-      });
-    this.updateEngineState(getExecutionEngineState({payloadStatus: status}));
+    const {status, latestValidHash, validationError} = await (this.rpcFetchQueue.push(engineRequest) as Promise<
```
nflaig: Is it ok to throw errors here? Why is previous comment no longer valid?

> If there are errors by EL like connection refused, internal error, they need to be
   treated separate from being INVALID. For now, just pass the error upstream.

packages/beacon-node/test/unit/execution/engine/utils.test.ts
```diff
@@ -0,0 +1,191 @@
...
+import {ExecutePayloadStatus, ExecutionEngineState} from "../../../../src/execution/index.js";
+import {
+  HTTP_CONNECTION_ERROR_CODES,
+  HTTP_FATAL_ERROR_CODES,
+  getExecutionEngineState,
+} from "../../../../src/execution/engine/utils.js";
+import {QueueError, QueueErrorCode} from "../../../../src/util/queue/errors.js";
+import {ErrorJsonRpcResponse, HttpRpcError} from "../../../../src/eth1/provider/jsonRpcHttpClient.js";
+
+describe("execution/engine/utils", () => {
```
nflaig: to be more consistent with existing test titles

```suggestion
describe("execution / engine / utils", () => {
```

packages/beacon-node/test/unit/execution/engine/utils.test.ts
```diff
@@ -0,0 +1,191 @@
...
+  describe("getExecutionEngineState", () => {
```
nflaig, line 14: These tests are great! 

~~As this relies on `error.code` returned from fetch, we might wanna merge https://github.com/ChainSafe/lodestar/pull/5811 first to make sure EL status is still works.~~

Tests would not catch if `error.code` returned from fetch changes, will have to  cover all cases in [fetch.test.ts](https://github.com/ChainSafe/lodestar/blob/41f21a9fa294e3760de7486230509c9ceeb38631/packages/api/test/unit/client/fetch.test.ts) to make sure we can rely on those error codes.

packages/beacon-node/src/execution/engine/http.ts
```diff
@@ -131,9 +129,13 @@ export class ExecutionEngineHttp implements IExecutionEngine {
...
     } catch (err) {
-      if (!isErrorAborted(err)) {
-        this.updateEngineState(getExecutionEngineState({payloadError: err}));
+      const newState = getExecutionEngineState({payloadError: err, oldState: this.state});
+      if (newState !== undefined) {
+        this.updateEngineState(newState);
       }
+
+      // TODO: This is case where we can't determine the nature of error.
+      // We should throw only for such cases not for the which we knew the status
```
nflaig: What exactly does this mean, right now all errors are thrown? Which ones should not be thrown if this TODO is resolved?

packages/beacon-node/test/unit/execution/engine/utils.test.ts
```diff
@@ -0,0 +1,191 @@
...
+    const testCasesPayload: Record<
+      ExecutePayloadStatus,
+      [oldState: ExecutionEngineState, newState: ExecutionEngineState][]
```
nflaig, line 17: In test cases, instead of calling it old and new, wouldn't it be clearer to call it from and to?

```suggestion
      [fromState: ExecutionEngineState, toState: ExecutionEngineState][]
```

Both seems fine though

nflaig (commented):
As the EL status changes rely on `error.code` returned from fetch, we might wanna merge https://github.com/ChainSafe/lodestar/pull/5811 first to make sure EL status still works and adapt error codes here in this PR if required.

I added a bunch of tests in the PR to make sure error codes still work as expected, see [fetch.test.ts](https://github.com/ChainSafe/lodestar/blob/41f21a9fa294e3760de7486230509c9ceeb38631/packages/api/test/unit/client/fetch.test.ts).

Based on codes defined here
https://github.com/ChainSafe/lodestar/blob/59ec5feaa0b7ef815b5752d3e4bb476e5c316151/packages/beacon-node/src/execution/engine/utils.ts#L39-L40

It seems like `EAI_AGAIN`, `ECONNRESET`, and `ECONNABORTED` are not covered, @nazarhussain if by any chance you know how to produce them I would add them to the test cases.

packages/beacon-node/test/unit/execution/engine/utils.test.ts
nazarhussain: These are very standard error codes must comply with any `fetch` implementation. Testing regressively native implementations is not suggested. But for node/browser compatibility of fetch we should have some basic tests.

nflaig: Yes, I doubt the error codes will ever change, I just tested different cases in regard to custom `FetchError` added in https://github.com/ChainSafe/lodestar/pull/5811 but these are rather meant to test if `FetchError` works as expected.

Since we use native fetch now it is even more unlikely we get unexpected errors. As per my understanding the codes you are checking right now for EL offline/auth failed should be alright and it is anyhow hard produce specific error codes (https://github.com/ChainSafe/lodestar/pull/5811#issuecomment-1658556984).

nazarhussain:
> It seems like `EAI_AGAIN`, `ECONNRESET`, and `ECONNABORTED` are not covered, @nazarhussain if by any chance you know how to produce them I would add them to the test cases.


For those e need to create a custom server implementation which reset the client connections. I don't think it's important enough to mock a server for now.

packages/beacon-node/src/execution/engine/http.ts
nazarhussain: Rephrased the comment to elaborate further.

packages/beacon-node/src/execution/engine/http.ts
nazarhussain: We treat the EL connection related error in new code where we wrap the `fetchRetries`. So this comment is not longer relevant.

packages/beacon-node/test/unit/execution/engine/utils.test.ts
nazarhussain: The `oldState` | `newState` is used across other places in code. So kept it as it.

packages/beacon-node/src/execution/engine/http.ts
nflaig: It seems to be still the same comment, it still unclear to me what exactly this means

packages/beacon-node/src/execution/engine/utils.ts
```diff
@@ -53,17 +50,58 @@ export function getExecutionEngineState({
 
     case ExecutePayloadStatus.UNAVAILABLE:
       return ExecutionEngineState.OFFLINE;
+
+    default:
+      // In case we can't determine the state, we assume it stays in old state
```
nflaig, line 55: this comment seems updated, wasn't the previous one more accurate?

```
// In case we can't determine the state, we assume it's online
// This assumption is better than considering offline, because the offline state may trigger some notifications
```

nflaig (commented):
Just some minor remarks, would suggest another review from someone else, maybe @g11tech who is more familiar with this code.

packages/beacon-node/src/execution/engine/http.ts
nazarhussain: Updated now, missed to pushed the last commit.

nflaig: Now I got what you mean, thanks 👍

packages/beacon-node/src/execution/engine/utils.ts
```diff
@@ -30,15 +32,10 @@ export class ExecutionEngineMockJsonRpcClient implements IJsonRpcHttpClient {
...
 
-export function getExecutionEngineState({
-  payloadError,
-  payloadStatus,
-}:
-  | {payloadStatus: ExecutePayloadStatus; payloadError?: never}
-  | {payloadStatus?: never; payloadError: unknown}): ExecutionEngineState {
+function getExecutionEngineStateForPayloadStatus(payloadStatus: ExecutePayloadStatus): ExecutionEngineState {
   switch (payloadStatus) {
     case ExecutePayloadStatus.ACCEPTED:
```
g11tech:
```suggestion
    case ExecPayloadStatus.ACCEPTED:
```
either we rename to this `ExecPayloadStatus` or to `ExecutionPayloadStatus`

packages/beacon-node/src/execution/engine/utils.ts
```diff
@@ -30,15 +32,10 @@ export class ExecutionEngineMockJsonRpcClient implements IJsonRpcHttpClient {
...
+export const HTTP_FATAL_ERROR_CODES = ["ECONNREFUSED", "ENOTFOUND", "EAI_AGAIN"];
+export const HTTP_CONNECTION_ERROR_CODES = ["ECONNRESET", "ECONNABORTED"];
 
-export function getExecutionEngineState({
-  payloadError,
-  payloadStatus,
-}:
-  | {payloadStatus: ExecutePayloadStatus; payloadError?: never}
-  | {payloadStatus?: never; payloadError: unknown}): ExecutionEngineState {
+function getExecutionEngineStateForPayloadStatus(payloadStatus: ExecutePayloadStatus): ExecutionEngineState {
```
g11tech:
```suggestion
function getELStateFromPayloadStatus(payloadStatus: ExecutePayloadStatus): ExecutionEngineState {
```
for brevity

packages/beacon-node/src/execution/engine/utils.ts
```diff
@@ -53,17 +50,59 @@ export function getExecutionEngineState({
 
     case ExecutePayloadStatus.UNAVAILABLE:
       return ExecutionEngineState.OFFLINE;
+
+    default:
+      // In case we can't determine the state, we assume it stays in old state
+      // This assumption is better than considering offline, because the offline state may trigger some notifications
+      return ExecutionEngineState.ONLINE;
```
g11tech, line 57:
```suggestion
      return ELState.ONLINE;
```
for brevity

packages/beacon-node/src/execution/engine/utils.ts
```diff
@@ -53,17 +50,59 @@ export function getExecutionEngineState({
...
+  }
+}
+
+function getExecutionEngineStateForPayloadError(
```
g11tech, line 61:
```suggestion
function getELStateFromPayloadError(
```

packages/beacon-node/src/execution/engine/http.ts
```diff
@@ -205,22 +206,11 @@ export class ExecutionEngineHttp implements IExecutionEngine {
       };
     }
 
-    const {status, latestValidHash, validationError} = await (
-      this.rpcFetchQueue.push(engineRequest) as Promise<EngineApiRpcReturnTypes[typeof method]>
-    )
-      // If there are errors by EL like connection refused, internal error, they need to be
-      // treated separate from being INVALID. For now, just pass the error upstream.
-      .catch((e: Error): EngineApiRpcReturnTypes[typeof method] => {
```
g11tech, line 213: this error handling is required upstream in processBlocks (as well as to not downgrade the peer in case of UNAVAILABLE)

g11tech (changes requested):
restore the previous error hadling in new payload

packages/beacon-node/src/execution/engine/utils.ts
nazarhussain: Renamed to `ExecutionPayloadStatus`.

packages/beacon-node/src/execution/engine/utils.ts
nazarhussain: I prefer it to be `getExecutionStateFromPayloadStatus` which makes the use of same vocabulary across codebase.

packages/beacon-node/src/execution/engine/utils.ts
nazarhussain: Renamed to `ExecutionState`

packages/beacon-node/src/execution/engine/interface.ts
```diff
@@ -30,7 +30,7 @@ export enum ExecutePayloadStatus {
   UNSAFE_OPTIMISTIC_STATUS = "UNSAFE_OPTIMISTIC_STATUS",
 }
 
-export enum ExecutionEngineState {
+export enum ExecutionState {
```
g11tech: i am thinking if `ExecutionState` might cause confusion over reference to actual execution state, in that case original `ExecutionEngineState` is preferable

nazarhussain: If I use `ExecutionEngineState`, then would prefer to use the same vocabulary/wording to other places for variables and functions. That will lead to almost similar to what we had earlier. Will it be ok?

g11tech (approved):
lgtm

PR #5800, opened 2023-07-25
feat(beacon-node): network worker event latency metrics

### LINKED ISSUE #5604: Network worker extremely busy / high event loop lag / high I/O lag

matthewkeil:
**NOTE** panel updates moved to PR#5827

**Motivation**

Add metrics for network event latency and message port latency to network worker.  Similar to how we collect data for latency across bls worker message port.  Also captures latency from initial emit of network event until re-emit of event on other thread event bus.

**Description**

Add additional metrics to help resolve #5604


packages/beacon-node/src/network/events.ts
```diff
@@ -27,13 +28,13 @@ export enum NetworkEvent {
...
-  [NetworkEvent.unknownBlock]: {rootHex: RootHex; peer?: PeerIdStr};
-  [NetworkEvent.pendingGossipsubMessage]: PendingGossipsubMessage;
-  [NetworkEvent.gossipMessageValidationResult]: {
+  [NetworkEvent.peerConnected]: EmittedAt & {peer: PeerIdStr; status: phase0.Status};
+  [NetworkEvent.peerDisconnected]: EmittedAt & {peer: PeerIdStr};
+  [NetworkEvent.reqRespRequest]: EmittedAt & {request: RequestTypedContainer; peer: PeerId};
+  [NetworkEvent.unknownBlockParent]: EmittedAt & {blockInput: BlockInput; peer: PeerIdStr};
+  [NetworkEvent.unknownBlock]: EmittedAt & {rootHex: RootHex; peer?: PeerIdStr};
+  [NetworkEvent.pendingGossipsubMessage]: EmittedAt & PendingGossipsubMessage;
+  [NetworkEvent.gossipMessageValidationResult]: EmittedAt & {
```
dapplion: Instead of adding this property on the business logic you can just track timings on the event bridge wrapper object

packages/beacon-node/src/network/gossip/gossipsub.ts
```diff
@@ -283,6 +283,7 @@ export class Eth2Gossipsub extends GossipSub {
     // See https://github.com/ChainSafe/lodestar/issues/5604
     setTimeout(() => {
       this.events.emit(NetworkEvent.pendingGossipsubMessage, {
+        emittedAt: Date.now(),
```
dapplion: milisecond precision may not be enough, BLS worker latency metrics are often < 1 ms, check the implementation there

dapplion (changes requested):

packages/beacon-node/src/network/events.ts
matthewkeil: Updated.

packages/beacon-node/src/network/gossip/gossipsub.ts
matthewkeil: Updated.

packages/beacon-node/src/network/core/networkCoreWorker.ts
```diff
@@ -85,11 +85,12 @@ new AsyncIterableBridgeHandler(getReqRespBridgeReqEvents(reqRespBridgeEventBus),
 );
 const reqRespBridgeRespCaller = new AsyncIterableBridgeCaller(getReqRespBridgeRespEvents(reqRespBridgeEventBus));
 
+let networkCoreWorkerMetrics: NetworkCoreWorkerMetrics | null = null;
 // respBridgeCaller metrics
 if (metricsRegister) {
-  const networkCoreWorkerMetrics = getNetworkCoreWorkerMetrics(metricsRegister);
+  networkCoreWorkerMetrics = getNetworkCoreWorkerMetrics(metricsRegister);
```
dapplion:
```ts
const networkCoreWorkerMetrics = metricsRegister ? getNetworkCoreWorkerMetrics(metricsRegister) : null;
if (networkCoreWorkerMetrics) {
  //
}
```
has the benefit of not compiling if someone deletes the line assigning `networkCoreWorkerMetrics`

matthewkeil: very good idea!!  🚀

matthewkeil: I will make sure to do it like this in the future as well

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -110,6 +110,11 @@ export function createLodestarMetrics(
         help: "Current count of pending items in reqRespBridgeReqCaller data structure",
       }),
     },
+    networkWorkerWireEventsOnMainThreadLatency: register.histogram({
```
twoeths: it'd be great to have suffix in the name representing unit, like 
```suggestion
    networkWorkerWireEventsOnMainThreadLatencySec: register.histogram({
```

matthewkeil: Updated.

nflaig: aren't all metrics in seconds?

matthewkeil: yes.  Dapplion had me change it to be in ISU units

dapplion: > aren't all metrics in seconds?

@nflaig point is that since all metrics are seconds it may not be able to specify so in the name. Actually we don't do that anywhere else (please check if this is true). So for consistency we should: use seconds everywhere, and don't declare that on the name. For code variables we use multiple units so in that case it makes sense to specify the unit

nflaig: The only metrics I found that are suffixed with `Seconds` are in `packages/beacon-node/src/metrics/metrics/lodestar.ts`.

E.g. `prevEpochAttestationsMinDelaySeconds` 
https://github.com/ChainSafe/lodestar/blob/201dfc805e7f7a60c9b03474d6c96ba452ca7094/packages/beacon-node/src/metrics/metrics/lodestar.ts#L799

Most metrics don't have this suffix, a more common suffix is `Time` for time based metrics
https://github.com/ChainSafe/lodestar/blob/201dfc805e7f7a60c9b03474d6c96ba452ca7094/packages/beacon-node/src/metrics/metrics/lodestar.ts#L1398

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -110,6 +110,12 @@ export function createLodestarMetrics(
         help: "Current count of pending items in reqRespBridgeReqCaller data structure",
       }),
     },
+    networkWorkerWireEventsOnMainThreadLatencySec: register.histogram<"eventName">({
+      name: "lodestar_network_worker_wire_events_on_main_thread_latency",
+      help: "Latency to transmit network events to main thread across worker port",
```
nflaig: The metric name does not follow the [recommended naming convention](https://prometheus.io/docs/practices/naming/#metric-names). In my opinion, no need to suffix the variable name (since we use seconds everywhere) but the metric name on the other hand must be suffixed.

```suggestion
    networkWorkerWireEventsOnMainThreadLatency: register.histogram<"eventName">({
      name: "lodestar_network_worker_wire_events_on_main_thread_latency_seconds",
      help: "Latency in seconds to transmit network events to main thread across worker port",
```

packages/beacon-node/src/network/core/metrics.ts
```diff
@@ -309,12 +309,20 @@ export function createNetworkCoreMetrics(register: RegistryMetricCreator) {
...
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
 export function getNetworkCoreWorkerMetrics(register: RegistryMetricCreator) {
   return {
     reqRespBridgeRespCallerPending: register.gauge({
       name: "lodestar_network_worker_reqresp_bridge_caller_pending_count",
       help: "Current count of pending elements in respBridgeCaller",
     }),
+    networkWorkerWireEventsOnWorkerThreadLatencySec: register.histogram<"eventName">({
+      name: "lodestar_network_worker_wire_events_on_worker_thread_latency",
+      help: "Latency to transmit network events to worker thread across parent port",
```
nflaig:
```suggestion
    networkWorkerWireEventsOnWorkerThreadLatency: register.histogram<"eventName">({
      name: "lodestar_network_worker_wire_events_on_worker_thread_latency_seconds",
      help: "Latency in seconds to transmit network events to worker thread across parent port",
```

dashboards/lodestar_networking.json
nflaig, line 1: There are issues with the dashboard changes but it looks like lint job still succeeded https://github.com/ChainSafe/lodestar/actions/runs/5724598511/job/15511407857?pr=5800#step:10:9
Need to exit with status code if error it thrown from lint script.

It looks like there are collapsed panels
```
Run scripts/validate-grafana-dashboards.sh
file:///home/runner/actions-runner/_work/lodestar/lodestar/scripts/lint-grafana-dashboard.mjs:30[6](https://github.com/ChainSafe/lodestar/actions/runs/5724598511/job/15511407857?pr=5800#step:10:7)
      throw Error(`all panels must be expanded, go to the UI and expand panel with title ${panel.title}`);
            ^

Error: file dashboards/lodestar_networking.json: all panels must be expanded, go to the UI and expand panel with title Summary
```

matthewkeil:
Screenshot 2023-08-01 at 6 37 40 AM
Screenshot 2023-08-01 at 6 37 47 AM

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -110,6 +110,12 @@ export function createLodestarMetrics(
         help: "Current count of pending items in reqRespBridgeReqCaller data structure",
       }),
     },
+    networkWorkerWireEventsOnMainThreadLatencySec: register.histogram<"eventName">({
```
nflaig: As per discussion I think we should remove `Sec`-suffix from metric variable names, @tuyennhv since you suggested that any objections from your side?

packages/beacon-node/src/util/workerEvents.ts
```diff
@@ -34,6 +38,12 @@ export function wireEventsOnWorkerThread<EventData>(
       // This check is not necessary but added for safety in case of improper implemented events
       isWorkerToMain[data.event] === EventDirection.mainToWorker
     ) {
+      const [sec, nanoSec] = process.hrtime(data.posted);
```
nflaig, line 46: Node docs suggest that `hrtime` is deprecated (https://nodejs.org/api/process.html#processhrtimetime) and `hrtime.bigint()` should be used instead.

packages/beacon-node/src/util/workerEvents.ts
```diff
@@ -34,6 +38,12 @@ export function wireEventsOnWorkerThread<EventData>(
...
+      const networkWorkerLatency = sec + nanoSec / 1e9;
```
nflaig: Maybe would help readability if we declare a `const NS_PER_SEC = 1e9` somewhere

matthewkeil:
@dapplion this was updated to just have the worker metrics and the dashboards were mergerd in #5827

dapplion (dismissed):

packages/beacon-node/src/util/workerEvents.ts
```diff
@@ -1,9 +1,14 @@
 import {MessagePort, Worker} from "node:worker_threads";
+import {Metrics} from "../metrics/metrics.js";
+import {NetworkCoreWorkerMetrics} from "../network/core/metrics.js";
 import {StrictEventEmitterSingleArg} from "./strictEvents.js";
 
+const nanoToSecondConversion = 1e9;
```
nflaig: Just minor remark, we use CONSTANT_CASE for such values in most parts of the code base

For example

https://github.com/ChainSafe/lodestar/blob/e690ac6df8b6c3e766fd5c245254575dad061f7b/packages/cli/src/cmds/beacon/handler.ts#L29

nflaig (changes requested):
We should clarify in this PR if we want `Sec`-suffix or not before merging it.

I think we should omit that see https://github.com/ChainSafe/lodestar/pull/5800#discussion_r1276126809 and conversation here https://github.com/ChainSafe/lodestar/pull/5800#discussion_r1276117993.

packages/beacon-node/src/util/workerEvents.ts
```diff
@@ -34,6 +40,10 @@ export function wireEventsOnWorkerThread<EventData>(
       // This check is not necessary but added for safety in case of improper implemented events
       isWorkerToMain[data.event] === EventDirection.mainToWorker
     ) {
+      metrics?.networkWorkerWireEventsOnWorkerThreadLatencySec.observe(
+        {eventName: data.event as string},
+        Number(process.hrtime.bigint() - data.posted) / nanoToSecondConversion
```
nflaig: Should be changed to `process.hrtime()`
- see https://github.com/ChainSafe/lodestar/pull/5926

nflaig:
> @nflaig point is that since all metrics are seconds it may not be able to specify so in the name. Actually we don't do that anywhere else (please check if this is true). So for consistency we should: use seconds everywhere, and don't declare that on the name. For code variables we use multiple units so in that case it makes sense to specify the unit

I agree with what @dapplion mentioned in https://github.com/ChainSafe/lodestar/pull/5800#discussion_r1276090932
- should use seconds for all metrics
- do not declare that on the name (remove `Sec`-suffix)
- For code variables (other variables in the code that are time related, not including metric variables) we should use a unit suffix

As outlined here https://github.com/ChainSafe/lodestar/pull/5800#discussion_r1276117993, we do not suffix metric variable names with a time unit right now, or at least not besides a few outliers.

packages/beacon-node/src/metrics/metrics/lodestar.ts
```diff
@@ -110,6 +110,12 @@ export function createLodestarMetrics(
...
+      name: "lodestar_network_worker_wire_events_on_main_thread_latency_seconds",
```
nflaig, line 125: While looking at networking metrics I noticed that the name here does not match with the value used in the dashboard

https://github.com/ChainSafe/lodestar/blob/e42d6cc6457051cf67ab7a3cdbfada25028d10f7/dashboards/lodestar_networking.json#L940

nflaig (approved):
LGTM 🚀

wemeetagain (approved):

wemeetagain:

PR #5803, opened 2023-07-26
feat: implement IndexedGossipQueue

twoeths:
**Motivation**

- We want to process attestations of the same data in batch so we want to group attestations of the same data by attestation data base64
- extracted from #5729

**Description**
- Separate to `GossipQueue` interface, the current implementation is `LinearGossipQueue`. This does not affect the current performance of node since noone is using the new gossip queue for now
- Implement new `IndexedGossipQueue`
  - index items by indexFn using a map, in the case of attestation it's attestation data base64
  - store keys with at least minChunkSize
  - on next, pick the the last key with minChunkSize, pop up to maxChunkSize items
  - on delete, pick the 1st key in the map and delete the 1st item in the list
- Implement some utility collections:
  - OrderedSet that's backed by LinkedList in order to get the first and last item
  - OrderedMap that's backed by OrderedSet in order to get first and last key/value

part of #5416


packages/beacon-node/test/perf/util/set.test.ts
```diff
@@ -0,0 +1,62 @@
...
+                if (deleteType === DeleteType.First) {
+                  set.delete(item, true);
+                } else {
+                  set.delete(item, false);
+                }
+                i++;
+              }
+            }
+          },
+          runsFactor: 1000,
```
dapplion: Make this 1000 a variable a dedup from above to ensure consistency

packages/beacon-node/src/network/processor/gossipQueues/indexed.ts
```diff
@@ -0,0 +1,153 @@
...
+      return 0;
+    }
+  }
+
+  /**
+   * Try to get items of last key with minChunkSize first.
+   * If not, pick the last key
+   */
+  next(): T[] | null {
+    let key: string | null = this.minChunkSizeKeys.last();
```
dapplion, line 106: I don't think this approach is best, consider this case:

- a key receives minChunkSize-1 items at time T1 and then 1 extra item at time T2
- It is appended to minChunkSizeKeys at time T2 and will be picked up first, but most of its items are at time T1

Similarly if a key does not receive minChunkSize it will always be de-prioritized. This approach risks respecting FIFO for some load patterns.

The goals for a FIFO queue with indexing that intends to batch should be:
1. Prioritize latest items
2. Buffer items to allow batching, but limit buffering by max wait or max length

For 1. in batches, what decide the timeliness of a batch?
- First arrival
- Latest arrival
- Avg of items

twoeths: yes, this is not ideal. Initially I did like this:
- always pick the last item (like LIFO)
- process it along with `min(64, number of messages with same data/key)`

the performance is far from the current approach since it can't batch much messages. I think it's not easy to pick a strategy, we need to:
- define some criterias to pick a strategy, one is: job wait time + job time + avg(forwarded messages)
- do some dynamic configurations here, configure it at runtime and compare

dapplion:
It's not easy to enforce FIFO and fairness rules having multiple list of objects that arrive at different times. We do FIFO for attestations because the most recent attestations have the most value to us.

Given a list of single-bit gossip attestations (assuming all valid) their value is equal the sum of `1 / incl_delay`. Prioritizing each list by first or last arrival time ignores the count of attestations a list have and could then, ignore big lists of not recent attestations.

One approach that would be fair but maybe expensive is to compute this "score" everytime

```ts

type MsgList = {
  avgRecvTimestampMs: number;
  items: GossipMsg[];
};
type GossipMsgId = string;
type GossipMsg = any;

const msgListScore = (a: MsgList): number => a.items.length / Math.max(1000, Date.now() - a.avgRecvTimestampMs);
const getKey = (_: GossipMsg): GossipMsgId => "TODO";

class Queue {
  queuesMap: Map<GossipMsgId, MsgList>;
  queuesList: MsgList[];

  add(item: GossipMsg): void {
    const key = getKey(item);
    let queue = this.queuesMap.get(key);

    if (!queue) {
      queue = {avgRecvTimestampMs: Date.now(), items: [item]};
      this.queuesMap.set(key, queue);
    } else {
      queue.items.push(item);
      // Compute average adding 1 value: avg = (avg * N + x) / (N + 1)
      queue.avgRecvTimestampMs =
        (queue.items.length * queue.avgRecvTimestampMs + Date.now()) / (queue.items.length + 1);
    }
  }

  getNext(): GossipMsg[] | undefined {
    // Sort in place by score, can be optimized runing this sometimes only
    this.queuesList.sort((a, b) => msgListScore(a) - msgListScore(b));

    return this.queuesList.shift()?.items;
  }
}
```

twoeths:
> One approach that would be fair but maybe expensive is to compute this "score" everytime

@dapplion  yes that's expensive, tested the `IndexedGossipQueueAvgTime` in the last 2h

Screenshot 2023-08-02 at 16 14 50

dapplion:
> > One approach that would be fair but maybe expensive is to compute this "score" everytime
> 
> @dapplion yes that's expensive, tested the `IndexedGossipQueueAvgTime` in the last 2h
> Screenshot 2023-08-02 at 16 14 50

please @tuyennhv  do not post spiky graphs, post either histograms or % of values above a certain threshold, or an avg_over_time

dapplion:
The queue must have some mechanism to pick up groups of attestation datas that do not reach the minimum count.

twoeths:
the current mechanism works as below:
```
attDataHex0: 30 items
attDataHex1: 35 items
attDataHex2: 1 items
attDataHex3: 2 items
attDataHex4: 2 item
```
with `minChunkSize=32` it'll pick 35 items of attDatHex1, then 2 items of attDataHex4, then 2 items of attDataHex3, then 1 item of attDataHex2 then 30 items of attDataHex0. If after we process `attDataHex1`, `attDataHex${n}` has more than 32 items then we'd process it next, the delay is worth because we batch more 

> The queue must have some mechanism to pick up groups of attestation datas that do not reach the minimum count.

@dapplion what do you expect, or want to improve with the above example? I can give it a try in a real node

dapplion:
Let's go with this and have histogram metrics to measure delay in the queue since first item

dapplion (approved):

wemeetagain:

PR #5810, opened 2023-07-27
feat: only subscribe to short lived subnets 2 slots in advanced

### LINKED ISSUE #5757: Do not subscribe to short-lived subnets too early
### Problem description

Right now when we have an aggregation duty at epoch n, we'll subscribe to the 1st slot of epoch `n-1` which force us to process a lot of attestations unnecessarily

cc @dapplion 

### Solution description

Investigate a good value (`x` slot) to subscribe to subnet in advance

### Additional context

_No response_

twoeths:
**Motivation**

- Right now if we have aggregator duty at epoch `n` we'll subscribe at the start of epoch `n-1` which is a waste of cpu and bandwidth

**Description**

- We still search for subnet peers at the start of epoch `n - 1`, that makes topic peers available, so it does not take time to get stable mesh peers
- Subscribe to dutied subnet 2 slots before dutied slot
- Use the current `aggregatorSlotSubnet` map, check it per clock slot to subscribe

Closes #5757

**Test**

- The produced aggregate has same number of participants on all nodes, for example `1v`

Screenshot 2023-07-27 at 14 48 39

- vs `unstable`

Screenshot 2023-07-27 at 14 49 13

- For short lived subnets, we subscribed in very short period

Screenshot 2023-07-27 at 14 50 55

- vs `unstable`

Screenshot 2023-07-27 at 14 51 20

- average number of mesh peers are good, this is on `1v` node

Screenshot 2023-07-27 at 14 57 54

### DIFF
--- a/packages/beacon-node/src/network/subnets/dllAttnetsService.ts
+++ b/packages/beacon-node/src/network/subnets/dllAttnetsService.ts
@@ -24,6 +24,9 @@ enum SubnetSource {
   longLived = "long_lived",
 }
 
+/** As monitored on goerli, we only need to subscribe 2 slots before aggregator dutied slot to get stable mesh peers */
+const SLOTS_TO_SUBSCRIBE_IN_ADVANCE = 2;
+
 /**
  * Manage deleterministic long lived (DLL) subnets and short lived subnets.
  * - PeerManager uses attnetsService to know which peers are required for duties and long lived subscriptions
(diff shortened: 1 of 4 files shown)


wemeetagain (approved):

packages/beacon-node/src/network/subnets/dllAttnetsService.ts
```diff
@@ -24,6 +24,9 @@ enum SubnetSource {
   longLived = "long_lived",
 }
 
+/** As monitored on goerli, we only need to subscribe 2 slots before aggregator dutied slot to get stable mesh peers */
+const SLOTS_TO_SUBSCRIBE_IN_ADVANCE = 2;
```
dapplion, line 28: @tuyennhv can you make this a hidden CLI option in another PR?

twoeths: @dapplion https://github.com/ChainSafe/lodestar/pull/5819

wemeetagain:

PR #5811, opened 2023-07-27
refactor: switch to native fetch implementation

### LINKED ISSUE #5765: Consecutive HTTP requests cause socket hang up on node v20
### Describe the bug

Consecutive HTTP requests cause socket hang up on node v20. This only happens with some servers, so far I could only reproduce this with Lighthouse. This issue causes the [sim tests to fail](https://github.com/ChainSafe/lodestar/actions/runs/5575358658/jobs/10185174387#step:8:94) and might cause other issues, e.g. when running Lodestar VC with a Lighthouse BN.

### Expected behavior

No socket hang up / requests failures

### Steps to reproduce

Running the following script will cause the first request to pass but the second one fails with socket hang up error

```ts
import { getClient } from "@lodestar/api";
import { config } from "@lodestar/config/default";

const lighthouse = "http://localhost:5052"
const api = getClient({ baseUrl: lighthouse }, { config });

await api.beacon.getBlockV2("head");
await api.beacon.getBlockV2("head");
```
```sh
> node index.js 

/home/nico/projects/test/debug-getStatev2/node_modules/node-fetch/lib/index.js:1505
                        reject(new FetchError(`request to ${request.url} failed, reason: ${err.message}`, 'system', err));
                               ^
FetchError: request to http://localhost:5052/eth/v2/beacon/blocks/head failed, reason: socket hang up
    at ClientRequest.<anonymous> (/home/nico/projects/test/debug-getStatev2/node_modules/node-fetch/lib/index.js:1505:11)
    at ClientRequest.emit (node:events:512:28)
    at Socket.socketOnEnd (node:_http_client:519:9)
    at Socket.emit (node:events:524:35)
    at endReadableNT (node:internal/streams/readable:1378:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  type: 'system',
  errno: 'ECONNRESET',
  code: 'ECONNRESET'
}

Node.js v20.4.0
```

Yielding to macro queue in-between the two requests does not produce an error
```ts
// ...
await api.beacon.getBlockV2("head");
setTimeout(async () => await api.beacon.getBlockV2("head"), 0);
```

### Additional context

The issue is already reported upstream

- https://github.com/node-fetch/node-fetch/issues/1735
- https://github.com/nodejs/node/issues/47130

There are also some [suggested workarounds](https://github.com/node-fetch/node-fetch/issues/1735) but those can't easily applied in our case since we are using cross-fetch and require browser compatiblity.

### Operating system

Linux

6e014217f9df78c0916299d357588a5bbab05d88

nflaig:
### Motivation

With recent issues we faced using cross-fetch (https://github.com/ChainSafe/lodestar/issues/5765 and `Failed to execute 'fetch' on 'Window': Illegal invocation`) I would suggest we get rid of it and use [native fetch](https://nodejs.org/api/globals.html#fetch) instead as technically there is no reason not to do it.

We already dropped support for node versions below 18.15.0 and native fetch is available on later versions and has to be explicitly be disabled by setting `--no-experimental-fetch` flag.

Native fetch in node is still [Stability: 1](https://nodejs.org/api/documentation.html#stability-index) but based on the requirement for it to be compatible with the browser implementation I don't expect there will be breaking changes in the future.

**Advantages**
- more consistent behavior node vs. browser
- less likely to have regressions when updating node
- better maintenance, cross-fetch is still on node-fetch v2 ([http://lquixada/cross-fetch#146](https://github.com/lquixada/cross-fetch/issues/146))

**Disadvantages**

Errors are not that great as details are inside `cause` property which itself is of type `Error`. The errors thrown by node-fetch are much better for us as they have their [own error handling](https://github.com/node-fetch/node-fetch/blob/main/docs/ERROR-HANDLING.md) which is different from native fetch in browser.

The error message is not detailed at all, it just says `fetch failed`, no context at all unless you look into cause
```sh
TypeError: fetch failed
    at Object.fetch (node:internal/deps/undici/undici:11576:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async file:///home/nico/projects/test/fetch-test/addr-not-found.js:2:3 {
  cause: Error: getaddrinfo ENOTFOUND non-existant-domain.com
      at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:118:26) {
    errno: -3008,
    code: 'ENOTFOUND',
    syscall: 'getaddrinfo',
    hostname: 'non-existant-domain.com'
  }
}
```

### Description

Removes cross-fetch and uses [native fetch](https://developer.mozilla.org/en-US/docs/Web/API/fetch) implementation instead. Inconsistent errors and intransparent messages are addressed by wrapping the native fetch implementation and throwing a custom `FetchError`, similar to [node-fetch error handling](https://github.com/node-fetch/node-fetch/blob/main/docs/ERROR-HANDLING.md).

Details about error handling can be found here
https://github.com/ChainSafe/lodestar/blob/33a0ceb08e2ccd1299b373c91ab9092bbc80dff1/packages/api/src/utils/client/fetch.ts#L24

### Compatibility

See [Fetch_API#browser_compatibility](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#browser_compatibility) for more details.

image

Closes https://github.com/ChainSafe/lodestar/issues/5765

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -75,7 +75,6 @@
     "@lodestar/params": "^1.9.2",
     "@lodestar/types": "^1.9.2",
     "@lodestar/utils": "^1.9.2",
-    "cross-fetch": "^3.1.8",
     "eventsource": "^2.0.2",
     "qs": "^6.11.1"
   },
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -10,6 +10,9 @@ export {
   HttpError,
   ApiError,
   Metrics,
+  FetchError,
+  isFetchError,
+  fetch,
 } from "./utils/client/index.js";
 export * from "./utils/routes.js";
 
--- a/packages/api/src/utils/client/fetch.ts
+++ b/packages/api/src/utils/client/fetch.ts
@@ -1,5 +1,7 @@
 /**
- * Wrapper around native fetch to improve error handling
+ * Native fetch with transparent and consistent error handling
+ *
+ * [MDN Reference](https://developer.mozilla.org/docs/Web/API/fetch)
  */
 async function wrappedFetch(url: string | URL, init?: RequestInit): Promise<Response> {
   try {
@@ -10,6 +12,10 @@ async function wrappedFetch(url: string | URL, init?: RequestInit): Promise<Resp
 }
 
 export {wrappedFetch as fetch};
+
+export function isFetchError(e: unknown): e is FetchError {
+  return e instanceof FetchError;
+}
 
 export type FetchErrorType = "failed" | "input" | "aborted" | "unknown";
 
--- a/packages/api/src/utils/client/fetch.ts
+++ b/packages/api/src/utils/client/fetch.ts
@@ -51,13 +51,6 @@ export class FetchError extends Error {
   }
 }
 
-/**
- * ```
- * TypeError: opaque message
- *   cause: Error: more detailed message
- *     code: 'ERROR_CODE'
- * ```
- */
 type NativeFetchError = TypeError & {
   cause: Error & {
     code?: string;
--- a/packages/api/src/utils/client/httpClient.ts
+++ b/packages/api/src/utils/client/httpClient.ts
@@ -1,6 +1,7 @@
-import {ErrorAborted, Logger, TimeoutError, fetch, FetchError} from "@lodestar/utils";
+import {ErrorAborted, Logger, TimeoutError} from "@lodestar/utils";
 import {ReqGeneric, RouteDef} from "../index.js";
 import {ApiClientResponse, ApiClientSuccessResponse} from "../../interfaces.js";
+import {fetch, isFetchError} from "./fetch.js";
 import {stringifyQuery, urlJoin} from "./format.js";
 import {Metrics} from "./metrics.js";
 import {HttpStatusCode} from "./httpStatusCode.js";
@@ -322,7 +323,7 @@ export class HttpClient implements IHttpClient {
 }
 
 function isAbortedError(e: Error): boolean {
-  return e instanceof FetchError && e.type === "aborted";
+  return isFetchError(e) && e.type === "aborted";
 }
 
 function getErrorMessage(errBody: string): string {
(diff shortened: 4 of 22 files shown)


wemeetagain (approved):
Looks good!

nflaig:
Just mentioned in EL offline PR (https://github.com/ChainSafe/lodestar/pull/5797), we might wanna cover all error codes used here

https://github.com/ChainSafe/lodestar/blob/59ec5feaa0b7ef815b5752d3e4bb476e5c316151/packages/beacon-node/src/execution/engine/utils.ts#L39-L40

I can try find out when those are thrown and add them in this PR

nflaig:
After looking at further error scenarios outlined here (https://betterstack.com/community/guides/scaling-nodejs/nodejs-errors/) and in Node.js docs (https://nodejs.org/api/errors.html) I think the test coverage is sufficient. The underlying http client implementation is still the same, fetch is just a higher level abstraction.

These errors are hard to test for locally / in tests:

 - `EAI_AGAIN`:  This happens due to a  DNS lookup timeout error (https://stackoverflow.com/a/40182520/10577550), this is non-deterministic and will produce false positives
 - `ECONNRESET`: It's a low-level, non-deterministic network error that's dependent on underlying TCP conditions. Can be reproduced by modifying client behavior but that's not that helpful in my opinion, this is the issue we faced here [#5765](https://github.com/ChainSafe/lodestar/issues/5765) due to client reusing a closed socket.
 - `ECONNABORTED`: Approach outlined here [#7-econnaborted](https://betterstack.com/community/guides/scaling-nodejs/nodejs-errors/#7-econnaborted) is not reproducible locally, I am guessing due to local / loopback connection.

It might be good to add `UND_ERR_SOCKET` to `HTTP_FATAL_ERROR_CODES` as this happens when server shuts down abruptly while there is a pending request but could cause false positives as this can happen due to other network issues as well.

https://github.com/ChainSafe/lodestar/blob/59ec5feaa0b7ef815b5752d3e4bb476e5c316151/packages/beacon-node/src/execution/engine/utils.ts#L39

wemeetagain:

PR #5816, opened 2023-07-28
refactor: rename data gas to blob gas for relevant deneb fields

g11tech:
pursuant to 
 - https://github.com/ethereum/consensus-specs/pull/3461

renaming data gas fields => blob gas

TODO: 
- [x] spec tests update

### DIFF
--- a/packages/beacon-node/test/spec/specTestVersioning.ts
+++ b/packages/beacon-node/test/spec/specTestVersioning.ts
@@ -15,7 +15,7 @@ import {DownloadTestsOptions} from "@lodestar/spec-test-util";
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
 export const ethereumConsensusSpecsTests: DownloadTestsOptions = {
-  specVersion: "v1.4.0-beta.0",
+  specVersion: "v1.4.0-beta.1",
   // Target directory is the host package root: 'packages/*/spec-tests'
   outputDir: path.join(__dirname, "../../spec-tests"),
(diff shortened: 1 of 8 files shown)


dapplion (dismissed):

packages/beacon-node/test/spec/specTestVersioning.ts
```diff
@@ -15,7 +15,7 @@ import {DownloadTestsOptions} from "@lodestar/spec-test-util";
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
 export const ethereumConsensusSpecsTests: DownloadTestsOptions = {
-  specVersion: "v1.4.0-beta.0",
+  specVersion: "v1.4.0-beta.1",
```
matthewkeil, line 18: Out of curiosity, why do we have this magic string in this file and in `packages/params/test/e2e/ensure-config-is-synced.test.ts`?

matthewkeil (approved):
Changes lgtm 🚀 

I noticed that we do not have a test for 
`tests/core/pyspec/eth2spec/test/helpers/sharding.py` from the spec.  Is that the only spec we still need to add to our suite?

Also the question about the magic string for specVersion.  Not sure if those would ever want to be different or not.

packages/beacon-node/test/spec/specTestVersioning.ts
g11tech: this corresponds to protocol spec test releases (which go hand in hand with spec releases) : https://github.com/ethereum/consensus-spec-tests/releases

wemeetagain:

PR #5820, opened 2023-07-31
fix(prover): allow payload provider to work with no finalized block

nazarhussain:
**Motivation**

Make the `@lodestar/prover` production ready. 

**Description**

- Make the `payload_provider` to work in situation when there is no finalized blocks
- Don't return `0` instead `undefined` for minimum finalized header incase of newly initialized object. 
- Return same payload header for `max` and `min` in case we have only one finalized header. 
- Some other minor fixes. 

Run all tests


wemeetagain (approved):

wemeetagain:

PR #5821, opened 2023-07-31
chore: format inputs on all dashboards to reduce diff

nflaig:
**Motivation**

Right now when running `node scripts/download_dashboards.mjs` it will produce a diff even if no dashboard was modified due to the fact that not all them have the same format of  `__inputs`  variables as defined in the lint script.

https://github.com/ChainSafe/lodestar/blob/487aef9445e4e83531d541054ea0b60ccf11d7e3/scripts/lint-grafana-dashboard.mjs#L99

**Description**

Format inputs on all dashboards to reduce diff when running download script.

### DIFF
--- a/dashboards/lodestar_block_processor.json
+++ b/dashboards/lodestar_block_processor.json
@@ -1,19 +1,19 @@
 {
   "__inputs": [
     {
-      "name": "DS_PROMETHEUS",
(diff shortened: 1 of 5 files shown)


dapplion (approved):

wemeetagain:

PR #5782, opened 2023-07-21
feat: only dial subnet peers if needed

twoeths:
**Motivation**

- For now we always dial subnet peers as much as we can, this lead to https://github.com/ChainSafe/lodestar/issues/5741
- This is because:
  - Most of the time, we have enough subnet peers very quickly (see `Peer requests to discovery` metrics)
  - But we always got ENRs very frequently (because node wants to connect to "maxPeers" so it issue FIND_NODES request frequently) and if the retuned ENRs have the subnets we want, we still connect to it
    - In the case of sync committee subscription, we usually have up to 50 long lived subnets per peer and most of them are outbound peers
  
**Description**

- Only connect to subnet peers if needed by adding "peersToConnect" to "subnetRequests"
- Based on the current metrics, this should help us have around 10 long lived subnets per peer instead of 50 like in the case of sync committee

Closes #5741

PR #5827, opened 2023-08-01
chore: add peer discovery metrics to grafana

twoeths:
**Motivation**

To know more about `PeerDiscovery` status

**Description**

- Add new metrics introduced in #5782 to "Peer requests to discovery" panel of networking dashboard
- Also add "Peer Manager" prefix to the other dashboards with similar name to avoid confusion, these metrics come from "PeerManager"

Screenshot 2023-08-01 at 14 01 35

### DIFF
--- a/dashboards/lodestar_networking.json
+++ b/dashboards/lodestar_networking.json
@@ -770,6 +770,246 @@
     },
     {
       "collapsed": false,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 30
+      },
+      "id": 608,
+      "panels": [],
+      "title": "Network Worker",
+      "type": "row"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${DS_PROMETHEUS}"
+      },
+      "description": "Async time to send a message across the worker/parent port",
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "palette-classic"
+          },
+          "custom": {
+            "axisCenteredZero": false,
+            "axisColorMode": "text",
+            "axisLabel": "",
+            "axisPlacement": "auto",
+            "barAlignment": 0,
+            "drawStyle": "line",
+            "fillOpacity": 23,
+            "gradientMode": "opacity",
+            "hideFrom": {
+              "graph": false,
+              "legend": false,
+              "tooltip": false,
+              "viz": false
+            },
+            "lineInterpolation": "linear",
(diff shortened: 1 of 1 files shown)


dashboards/lodestar_networking.json
```diff
@@ -768,6 +768,246 @@
...
+      "pluginVersion": "7.4.5",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "editorMode": "code",
+          "exemplar": false,
+          "expr": "(\n  sum(rate(\n    lodestar_network_worker_wire_events_on_worker_thread_latency_sum[$rate_interval]\n  )) \n  +\n  sum(rate(\n    lodestar_network_worker_wire_events_on_main_thread_latency_sum[$rate_interval]\n  ))\n)\n/\n(\n  sum(rate(\n    lodestar_network_worker_wire_events_on_worker_thread_latency_count[$rate_interval]\n  ))\n  +\n  sum(rate(\n    lodestar_network_worker_wire_events_on_main_thread_latency_count[$rate_interval]\n  ))\n)",
```
twoeths, line 861: @matthewkeil I found this metric from #5800, do you want it to be part of this PR too?

matthewkeil: I leave that up to you @tuyennhv.  I changed them around a few times so they may be a bit different from the last time you checked.  Feel free to either pull the updated ones into this PR or I can merge the additions from yours into mine.  Whichever works for me.  I leave the decision with you. @wemeetagain mentioned this to me as well so just let me know how to help because I think he wants to get these into 1.10

Here is a screenshot of the updated version I made.  I would go with your overall average because my calculation seems incorrect, but I kinda prefer the layout I settled on with one panel broken out by `eventName` and the other panel with just the three averages on it. `Average`, average of `workerToMain` and `mainToWorker` because they seem to be pretty different in the two directions and the individual metrics are also quite distinct in value for each event type so there will be a lot of opportunity to tune the individual events but they are hard to glance at to see if the averages is changing.

Screenshot 2023-08-01 at 6 37 40 AM

Screenshot 2023-08-01 at 6 37 47 AM

twoeths: ok then let's also add your panels to grafana, you can change them later @matthewkeil

twoeths:
this also include new panels from @matthewkeil , ready for review

matthewkeil (approved):
LGTM 🚀

Issue #2115, opened 2021-02-28
Consider increasing max_semi_space_size


dapplion:
According to GC metrics on master the beacon node spends 10% of the time doing Scavenge GC runs. 

![Screenshot from 2021-02-28 20-45-51](https://user-images.githubusercontent.com/35266934/109431313-37d48880-7a06-11eb-8560-95d6b0db8f40.png)

If the metrics are correct this article suggests increasing `max_semi_space_size` with the flag `--max_semi_space_size` to reduce the frequency of these runs.

https://www.alibabacloud.com/blog/better-node-application-performance-through-gc-optimization_595119

wemeetagain:
cc @matthewkeil re #5829

matthewkeil:
@wemeetagain I had this set on `beta` when we discovered the memory leak in #5851 so it was hard to see full effect but it did cut the Scavenge time in line with the worker new space adjustment before things started to get wonky when the heap grew from the leak.  It will most definitely be a nice tune-up though and I will get this set correctly once we get the leak addressed and can see the fruits of the change!!

matthewkeil:
As a note, the docs are not super clear about what "semi" space is.  But after digging in the codebase its related to young generation and `kNewLargeObjectSpaceToSemiSpaceRatio` which always equals `1`;

```c++
size_t Heap::YoungGenerationSizeFromSemiSpaceSize(size_t semi_space_size) {
  return semi_space_size * (2 + kNewLargeObjectSpaceToSemiSpaceRatio);
}
```

matthewkeil:
Memory leak #5851 resolved.  Deploying `unstable` to `feat2` with `--max-semi-space-size=64`.  The max scavenge witnessed currently was on mainnet in the unstable group.  All other instances were lower and some substantially lower. 
Unsure how the larger value will affect performance on the smaller instances so will let run for a few days without the worker to see what turns up.

`unstable-mainnet-hzax41`
![Screenshot 2023-08-10 at 1 23 13 PM](https://github.com/ChainSafe/lodestar/assets/18608739/93f1e801-5e15-43a2-b539-154545e26d9a)

`unstable-novc-ctvpss`
![Screenshot 2023-08-10 at 1 25 03 PM](https://github.com/ChainSafe/lodestar/assets/18608739/7cb0297e-2acc-4e31-bcde-b0c717196aef)

`unstable-sm1v-ctvpss`
[...]

PR #5829, opened 2023-08-01
feat(beacon-node): network worker new space adjustment 

matthewkeil:
**Motivation**

While going through the suggestions that Ben gave us I found that there was a very large amount of new space scavenge collection.  I updated the dashboards and realized it was over 200mb and 3% of the worker thread time spent on scavenge.  There was a lot of page-faults as well which lends itself to a bunch of data being moved during GC.  There was also not very much mark-and-sweep GC which means the objects were most likely getting moved only once before getting un-ref'd. This realization that most of the objects are very short lived, but likely high in volume to trigger the scavenge.  So in essence causing them to get moved back and forth between the `To` and `From` spaces a single time.

I played around with the values of new space to get around this.  I bumped the new space way up above the GC volume.  Now the objects do not live long enough to need to be moved in the new space.  The HandleScope and new space management system are quite efficient so there is little to no performance degradation from leaving the objects in new space once they are no longer ref'd.  They just either sit at the back of the line or get written over when the HandleScope pointers shift.

**Description**

- Updated dashboard panels for memory.
- Added panels for each thread (heap and GC stats) in a row
- Updated network worker new space to a sufficiently large number (over the existing `unstable` GC volume)

[X] - Figure out method to determine how large is too large
[X] - Verify that new space is not too large

Helps with #5604 

** Metrics **

Need to let the node stabilize for a few more hours so traffic is back up to levels on `unstable` for even comparison but things are looking pretty good so far.

Screenshot 2023-08-01 at 3 55 53 AM
Screenshot 2023-08-01 at 3 56 19 AM
Screenshot 2023-08-01 at 3 56 26 AM
Screenshot 2023-08-01 at 3 56 39 AM

Screenshot 2023-08-01 at 4 00 52 AM
Screenshot 2023-08-01 at 4 00 27 AM
Screenshot 2023-08-01 at 3 57 08 AM

Screenshot 2023-08-01 at 4 02 04 AM
Screenshot 2023-08-01 at 4 02 30 AM
Screenshot 2023-08-01 at 4 01 44 AM
Screenshot 2023-08-01 at 4 04 39 AM
Screenshot 2023-08-01 at 4 04 06 AM
Screenshot 2023-08-01 at 4 03 46 AM
Screenshot 2023-08-01 at 4 02 45 AM


packages/beacon-node/src/network/core/networkCoreWorkerHandler.ts
```diff
@@ -107,7 +107,10 @@ export class WorkerNetworkCore implements INetworkCore {
       loggerOpts: modules.logger.toOpts(),
     };
 
-    const worker = new Worker("./networkCoreWorker.js", {workerData} as ConstructorParameters<typeof Worker>[1]);
+    const worker = new Worker("./networkCoreWorker.js", {
+      workerData,
+      resourceLimits: {maxYoungGenerationSizeMb: 512},
```
dapplion: What's the default value?

dapplion:
Metrics looks promising! This PR does not add metrics, but changes a fundamental parameter, can you split:
- PR A: Add dashboards
- PR B: Change constant value

packages/beacon-node/src/network/core/networkCoreWorkerHandler.ts
matthewkeil: It's a [calculated value](https://github.com/nodejs/node/blob/4166d40d0873b6d8a0c7291872c8d20dc680b1d7/deps/v8/src/heap/heap.cc#L266) and ends up roughly 16mb with a `--max-old-space-size` of 4GB.

matthewkeil:
This was updated to just have the single line change.  I rebased this onto unstable from a discussion with @wemeetagain but there is a memory leak on unstable so its hard to see how this is actually performing at 128mb.  I will move to ready for review once we get the leak sorted and can see how this performs with @tuyennhv update with limited subscriptions.

Dashboard changes were moved to PR #5858

matthewkeil:
Value was set from research on #2115.  You can find the resolution to that issue [here](https://github.com/ChainSafe/lodestar/issues/2115#issuecomment-1687860046)


There is a 3x1 ratio of semi space to young generation space in the code but in practice it seems like setting the semi space to the average value of scavenge collection seemed optimal.  In previous empirical testing it seemed to also be the optimal value for the worker young generation setting despite the ratio.  I am not certain of the reason for this but its worth further investigation.

packages/beacon-node/src/network/core/networkCoreWorkerHandler.ts
```diff
@@ -107,7 +107,10 @@ export class WorkerNetworkCore implements INetworkCore {
       loggerOpts: modules.logger.toOpts(),
     };
 
-    const worker = new Worker("./networkCoreWorker.js", {workerData} as ConstructorParameters<typeof Worker>[1]);
+    const worker = new Worker("./networkCoreWorker.js", {
+      workerData,
+      resourceLimits: {maxYoungGenerationSizeMb: 152},
```
wemeetagain: Can you add a comment here linking / describing why 152?

wemeetagain: might be worth exposing as a hidden cli flag so we don't need to PR to test other values

matthewkeil: Updated.

wemeetagain (dismissed):

wemeetagain (approved):

wemeetagain:

PR #5830, opened 2023-08-01
chore: fix format of input in dashboard lint script

nflaig:
**Motivation**

After further looking into why we get random dashboard diffs on __inputs it looks like downloaded dashboards have __inputs fields and the format that Grafana uses it different from the format we use in the lint script.

Maybe depends on Grafana version or when dashboard was last time updated as it does not happen for all dashboards. Tested it with our grafana instance.

**Description**

- fix format of input in dashboard lint script
- fix inputs on remaining dashboards that were not updated

### DIFF
--- a/dashboards/lodestar_networking.json
+++ b/dashboards/lodestar_networking.json
@@ -1,12 +1,12 @@
 {
   "__inputs": [
     {
-      "name": "DS_PROMETHEUS",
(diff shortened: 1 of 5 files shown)


wemeetagain (approved):

wemeetagain:

Issue #5823, opened 2023-07-31
Block proposal improvement tracker


g11tech:
tacker for various items to make block production more robust

- [ ] register fee recipients to all fallback bs for the VC
- [x] https://github.com/ChainSafe/lodestar/pull/5831
- [ ] for the multiple fallback bns, resolve with the most successful blinded block, similarly check for the bns
  -  the block production should be multiple bn aware rather than fallback urls handled in httpClient  
- [x] https://github.com/ChainSafe/lodestar/pull/5880 
  - switch to produce block v3 
  - move the execution blinded race to the beacon
- [x] https://github.com/ChainSafe/lodestar/pull/6227
   -  allow execution block to send a blinded response for block production for low latency in publishing 
   - may be full body could be blinded than just the execution pa
   -  respond with a blocksource response for bn to know if the blinded produced block is bound to a bn or not 
- [x] https://github.com/ChainSafe/lodestar/pull/6236
- [x] https://github.com/ChainSafe/lodestar/issues/5826

philknows:
Can you update this issue with the latest status? 🙏

g11tech:
updated, will continue on this post latest changes are PRed for the deneb ... affects produce/publish api all over

wemeetagain:
@g11tech is this fully resolved?

PR #5831, opened 2023-08-01
feat: send and use validators fee recipient for engine block production

g11tech:
part of
 - https://github.com/ChainSafe/lodestar/issues/5823

Generally the validator registers fee recipient every epoch for engine (as well as builder block production). However in some edge cases like where registration was not updated (for e.g. fee recipient updated just few slots into epoch via keymanager) or registration issues or the block production requests routed to a fallback bn where validators registrations were not done .... (the edge case lodestar deployments witnessed on 31 july), its imperative that the produce block calls should carry the fee recipient and start a fresh engine block production with a new recipient if bn didn't have a prepared payload id which matches the requested fee recipient.

This PR achieves the same.

TODO:
- [x] add unit/mock test cases for the api
- [x] improve fee recipient logging for easy debugging

### DIFF
--- a/packages/api/src/beacon/routes/validator.ts
+++ b/packages/api/src/beacon/routes/validator.ts
@@ -201,7 +201,8 @@ export type Api = {
   produceBlock(
     slot: Slot,
     randaoReveal: BLSSignature,
-    graffiti: string
+    graffiti: string,
+    feeRecipient?: string
   ): Promise<
     ApiClientResponse<
       {[HttpStatusCode.OK]: {data: allForks.BeaconBlock; blockValue: Wei}},
@@ -222,7 +223,8 @@ export type Api = {
   produceBlockV2(
     slot: Slot,
     randaoReveal: BLSSignature,
-    graffiti: string
+    graffiti: string,
+    feeRecipient?: string
   ): Promise<
     ApiClientResponse<
       {[HttpStatusCode.OK]: {data: allForks.BeaconBlock | BlockContents; version: ForkName; blockValue: Wei}},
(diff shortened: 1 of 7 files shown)


packages/api/src/beacon/routes/validator.ts
```diff
@@ -428,7 +431,7 @@ export type ReqTypes = {
   getProposerDuties: {params: {epoch: Epoch}};
   getSyncCommitteeDuties: {params: {epoch: Epoch}; body: U64Str[]};
   produceBlock: {params: {slot: number}; query: {randao_reveal: string; graffiti: string}};
-  produceBlockV2: {params: {slot: number}; query: {randao_reveal: string; graffiti: string}};
+  produceBlockV2: {params: {slot: number}; query: {randao_reveal: string; graffiti: string; fee_recipient?: string}};
```
dapplion, line 434: This is not part of the spec right? https://github.com/ethereum/beacon-APIs/blob/ea5668708f2c6717fd2934c9a0b82b6705818a2e/apis/validator/block.v2.yaml#L13 adding this here will cause our API client to break with spec compliant impls

g11tech: yes you are correct :slightly_smiling_face:  its not in spec, plan to PR there as well (for produce block v3), 

but its spec compatible, in the sense that if the fee recipient is not passed, it will behave in the old spec complaint way

dapplion (approved):
Looks good! Please upstream this change to API specs

g11tech:
> Looks good! Please upstream this change to API specs

sounds good! will open PR

wemeetagain:

PR #5832, opened 2023-08-01
docs: update instructions on how to contribute to dashboards

nflaig:
**Motivation**

After using the download script a bit more myself noted that instructions are not that precise for current workflow.

It is no longer required to import a temporal dashboard with different name but more importantly it must be ensured that the UID is not modified, else the download script will not pick up the changes.

It should be pretty clear to existing maintainers that when using our Grafana cloud instance there is no need to import the dashboard, in fact, it is not possible to import it with same UID. Point 1. and 2. are rather meant for (new) contributors that modify dashboards on their own / local Grafana instance.

**Description**

Update instructions on how to contribute to dashboards to better reflect current workflow.

### DIFF
--- a/.wordlist.txt
+++ b/.wordlist.txt
@@ -59,6 +59,7 @@ TOC
 TTD
 TypeScript
 UI
+UID
 UTF
 VM
 Vitalik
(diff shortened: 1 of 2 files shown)


wemeetagain (approved):

wemeetagain:

PR #5842, opened 2023-08-03
chore: clarify event loop lag metrics in grafana

twoeths:
**Motivation**

- There are event loop lag metric A and metric B which causes some confusion to the team. I found that the Metric A `nodejs_eventloop_lag_seconds` actually comes from `prom-client` https://github.com/siimon/prom-client/blob/v14.2.0/lib/metrics/eventLoopLag.js#L54 while Metric B `nodejs_eventloop_lag_mean_seconds`  is computed by NodeJS api https://nodejs.org/api/perf_hooks.html#perf_hooksmonitoreventloopdelayoptions

- Today I found that metric A keeps increasing (v1.9.2 vs v1.10.0-rc.1) while metric B is consistently low, and the `vc` api is also consistently low so we should care more metric B (NodeJS event loop lag) cc @dapplion 

Screenshot 2023-08-03 at 16 02 15

Screenshot 2023-08-03 at 16 02 27

Screenshot 2023-08-03 at 16 03 19

**Description**

- Change name of event loop lag metric A to "prom-client Event Loop Lag"
- Change name of event loop lag metric B to "NodeJS Event Loop Lag"

The change is small but looks like this "VM" dashboard is out of date so there are a lot of changes, if someone saved/updated the dashboard intentionally please confirm in this PR

### DIFF
--- a/dashboards/lodestar_vm_host.json
+++ b/dashboards/lodestar_vm_host.json
@@ -62,28 +62,269 @@
   "panels": [
     {
       "collapsed": false,
-      "datasource": {
-        "type": "prometheus",
-        "uid": "${DS_PROMETHEUS}"
-      },
       "gridPos": {
         "h": 1,
         "w": 24,
         "x": 0,
         "y": 0
       },
-      "id": 12,
+      "id": 552,
       "panels": [],
+      "title": "Memory Overview",
+      "type": "row"
+    },
+    {
+      "datasource": {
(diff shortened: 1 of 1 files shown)


dapplion:
I may have changed some charts without updating the source

dapplion (approved):

wemeetagain (approved):

PR #5854, opened 2023-08-07
fix(prover): update web3_provider to detect web3.js 4x version

nazarhussain:
**Motivation**

Make the `@lodestar/prover` ready for production use. 

**Description**

- Make the `web3_provider` logic to detect the `Web3.js` 4x version. Which have a different implementation logic from legacy version. 

Run all tests.


packages/cli/package.json
```diff
@@ -78,10 +78,10 @@
     "@types/lockfile": "^1.0.2",
     "bip39": "^3.1.0",
     "deepmerge": "^4.3.1",
-    "ethers": "^5.0.2",
+    "ethers": "^6.7.0",
     "expand-tilde": "^2.0.2",
     "find-up": "^6.3.0",
-    "got": "12.4.1",
+    "got": "^11.8.6",
```
nazarhussain, line 84: Had to downgrade this because both `got 12.x` and `got 13.x` was causing types incompatibility issues with some other packages. 

https://github.com/jaredwray/cacheable/issues/174

wemeetagain (approved):

packages/cli/package.json
nflaig: What's the reason for using `got` instead of `fetch`?

nazarhussain: Probably old implementation. Don't have specific reason for its usage.

Issue #5118, opened 2023-02-09
Memory usage metrics panel should only track beacon node


philknows:
**Describe the bug**

Currently, the memory usage panel is showing data from both beacon node and validator which includes some confusing data points on the graph.

<img width="808" alt="Screenshot 2023-02-09 at 11 49 21 AM" src="https://user-images.githubusercontent.com/58080811/217739542-352302b0-0c26-46f8-a6cf-339969615f6b.png">

**Expected behavior**

For organization and simplicity, it should only be tracking beacon node memory usage.

twoeths:
this was broken after https://github.com/ChainSafe/lodestar/pull/5020

nflaig:
does the beacon node and validator use the same job? The prometheus config file needs to looks like this, with separate jobs for BN and VC

https://github.com/ChainSafe/lodestar/blob/db612ea82f618c4ab6509ebd88f96631788c93d2/prometheus.yml#L1-L12

nflaig:
This https://github.com/ChainSafe/lodestar/pull/5200 should fix all the remaining issues, the job names can now be customized when importing the dashboard. This means can just set the job name currently used and there is no longer the requirement to use `beacon` and `validator` as `job_name` as renaming jobs in prometheus config causes historical data to be lost which might not be desired.

nflaig:
The changes mention in https://github.com/ChainSafe/lodestar/issues/5118#issuecomment-1441936895 are now merged to `unstable`, let me know if there are still problems, else I would recommend that we close the issue.

nflaig:
Seems like the issue is related to how our internal prometheus instance is configured. We use the same job called `beacon` to scrape different services, including the validator client. Further details can be found [on discord](https://discord.com/channels/593655374469660673/605818244036821012/1081114490537521192).

nflaig:
I think we can close this in favor of https://github.com/ChainSafe/lodestar-ansible-development/issues/121 as the problem is not related to Lodestar or the dashboard contained in this repository.

nflaig:
Resolved by https://github.com/ChainSafe/lodestar-ansible-development/pull/141

PR #5858, opened 2023-08-07
feat(dashboards):  memory panel updates

matthewkeil:
**Motivation**

Breaks out memory panes by thread for greater insight.  Some of the changes were already merged and this fixes the remaining updates.  Fixes a couple of metrics that disappear when the network worker is turned on.

Also adds a package.json script to `lint-dashboards`

**Description**

PR broken out of #5829


wemeetagain (approved):

dashboards/lodestar_vm_host.json
```diff
@@ -148,7 +148,7 @@
           },
           "editorMode": "code",
           "exemplar": false,
-          "expr": "sum(process_resident_memory_bytes{job=~\"beacon|validator\"})",
+          "expr": "sum(process_resident_memory_bytes{job=~\"$beacon_job|beacon|$validator_job|validator\"})",
```
nflaig, line 151: what the point of matching both jobs? Can just remove the job matcher in that case

package.json
```diff
@@ -18,6 +18,7 @@
     "lint:fix": "yarn lint --fix",
     "lint-docs": "prettier '**/*.md' --check",
     "lint-docs:fix": "prettier '**/*.md' --write",
+    "lint-dashboards": "scripts/validate-grafana-dashboards.sh",
```
nflaig, line 21: I think we should not mix up validation and linting, the validate script does other things like git diff which is not required if you lint locally, it just happens to work right now but that might change in the future

Would recommend the following instead

`"lint-dashboards": "node scripts/lint-grafana-dashboards.mjs ./dashboards",`

nflaig (commented):
@matthewkeil bit late on the review, but suggested some changes that could be incorporated in another PR

dashboards/lodestar_vm_host.json
nflaig: @matthewkeil I looked at this again and I am not sure we want to track beacon and validator metrics in a single panel.

At least @philknows opened an issue for this previously, not sure what the desired behavior is now though.
- https://github.com/ChainSafe/lodestar/issues/5118

Regarding the 3rd `process_resident_memory_bytes` metric, seems to be from https://github.com/lidofinance/validator-ejector, this means job label matcher is required.

```
process_resident_memory_bytes{app="validator-ejector", client_name="lodestar", group="lido-goerli", host_type="lg1k", instance="lido-goerli-hzax41-0", job="ejector", network="goerli", scrape_location="ejector"}
```

matthewkeil: there is another job that I wanted to exclude.  only wanted to take those two

package.json
matthewkeil: i like the idea of being able to run the same command that the CI runs so I can see if it will pass/fail before pushing the commit.  I suppose we can change the script command name?

nflaig: We have a bunch of those scripts in `./scripts` folder
- [assert_eslintrc_sorted.mjs](https://github.com/ChainSafe/lodestar/blob/unstable/scripts/assert_eslintrc_sorted.mjs)
- [assert_exports.mjs](https://github.com/ChainSafe/lodestar/blob/unstable/scripts/assert_exports.mjs)
- [assert_no_yarn_warnings.sh](https://github.com/ChainSafe/lodestar/blob/unstable/scripts/assert_no_yarn_warnings.sh)

Could just run them from there directly.

nflaig: Would say lint should lint and not validate but it is a minor detail so no big deal.

wemeetagain:

Issue #4536, opened 2022-09-12
Enable `trace` log level


yorickdowne:
**Is your feature request related to a problem? Please describe.**

eth-docker uses the same log level for all running components. `trace` is an industry standard, but not supported by Lodestar. This causes Lodestar startup to fail when `trace` logging is configured.

**Describe the solution you'd like**

Support a `trace` log level as an alias to `silly`. That way `trace` logging can be enabled for a "stack" consisting of Lodestar, an EL, Grafana, etc.

dapplion:
Sounds good to me :+1:

PR #5861, opened 2023-08-08
feat(logger): remove unused trace log level

### LINKED ISSUE #5809: Broken `trace` logging
### Describe the bug

It appears that trace logging is broken.  Attempted to add a `logger.trace()` call and e2e tests and the node broke when deployed.

I removed the call and the error went away.  I changed the error to `debug` and the error also went away.  It looks like the name was changed from `silly` to `trace` in #4539 and winston expects the string `silly` to be passed.

We are [setting](https://github.com/ChainSafe/lodestar/blob/2c3d7e14f0082cfe70156e55a9213db952f815c4/packages/logger/src/winston.ts#L56) the [`logLevelNum`](https://github.com/ChainSafe/lodestar/blob/2c3d7e14f0082cfe70156e55a9213db952f815c4/packages/logger/src/interface.ts#L4) which the [docs say](https://github.com/winstonjs/winston#using-custom-logging-levels) should work but it may not be set correctly.

```sh
TypeError: colors[Colorizer.allColors[lookup]] is not a function
    at Colorizer.colorize (/usr/src/lodestar/node_modules/logform/colorize.js:73:49)
    at Colorizer.transform (/usr/src/lodestar/node_modules/logform/colorize.js:98:25)
    at Format.transform (/usr/src/lodestar/node_modules/logform/combine.js:20:24)
    at DerivedLogger._transform (/usr/src/lodestar/node_modules/winston/lib/winston/logger.js:313:29)
    at DerivedLogger.Transform._read (/usr/src/lodestar/node_modules/readable-stream/lib/_stream_transform.js:177:10)
    at DerivedLogger.Transform._write (/usr/src/lodestar/node_modules/readable-stream/lib/_stream_transform.js:164:83)
    at doWrite (/usr/src/lodestar/node_modules/readable-stream/lib/_stream_writable.js:409:139)
    at writeOrBuffer (/usr/src/lodestar/node_modules/readable-stream/lib/_stream_writable.js:398:5)
    at DerivedLogger.Writable.write (/usr/src/lodestar/node_modules/readable-stream/lib/_stream_writable.js:307:11)
    at DerivedLogger.log (/usr/src/lodestar/node_modules/winston/lib/winston/logger.js:225:14) colors[Colorizer.allColors[lookup]] is not a function
```

### Expected behavior

A trace log should work and not break the other log levels

### Steps to reproduce

1. Create a `logger.trace`
2. Run e2e tests and should work correctly 

### Additional context

_No response_

### Operating system

Linux

unstable

matthewkeil:
**Motivation**

`trace` does not follow our logging policy and it is broken as well.  Found an error and instead of fixing we decided to remove it.  

Closes  #5809 

**Description**

- Ensured no usage of the `log.trace()` 
- Update `Logger` interface to remove `trace` method
- Kept `LogLevel.trace` so that it can be passed into the CLI without throwing an error but will be handled similar to `debug` logging level
- Remove `trace` method from `WinstonLogger`, `getLoggerVc` and `getEmptyLogger`

### DIFF
--- a/packages/logger/src/empty.ts
+++ b/packages/logger/src/empty.ts
@@ -17,8 +17,5 @@ export function getEmptyLogger(): Logger {
     debug: function debug(): void {
       // Do nothing
     },
-    trace: function trace(): void {
-      // Do nothing
-    },
   };
 }
--- a/packages/logger/src/winston.ts
+++ b/packages/logger/src/winston.ts
@@ -77,8 +77,8 @@ export class WinstonLogger implements Logger {
     this.createLogEntry(LogLevel.debug, message, context, error);
   }
 
-  trace(message: string, context?: LogData, error?: Error): void {
-    this.createLogEntry(LogLevel.trace, message, context, error);
+  trace(): void {
+    throw new Error("Trace it not supported by Lodestar");
   }
 
   private createLogEntry(level: LogLevel, message: string, context?: LogData, error?: Error): void {
--- a/packages/logger/src/winston.ts
+++ b/packages/logger/src/winston.ts
@@ -77,10 +77,6 @@ export class WinstonLogger implements Logger {
     this.createLogEntry(LogLevel.debug, message, context, error);
   }
 
-  trace(message: string, context?: LogData, error?: Error): void {
-    this.createLogEntry(LogLevel.trace, message, context, error);
-  }
-
   private createLogEntry(level: LogLevel, message: string, context?: LogData, error?: Error): void {
     // Note: logger does not run format.transform function unless it will actually write the log to the transport
 
--- a/packages/logger/src/winston.ts
+++ b/packages/logger/src/winston.ts
@@ -77,8 +77,8 @@ export class WinstonLogger implements Logger {
     this.createLogEntry(LogLevel.debug, message, context, error);
   }
 
-  trace(): void {
-    throw new Error("Trace it not supported by Lodestar");
+  trace(message: string, context?: LogData, error?: Error): void {
+    this.createLogEntry(LogLevel.trace, message, context, error);
   }
 
   private createLogEntry(level: LogLevel, message: string, context?: LogData, error?: Error): void {
--- a/packages/utils/src/logger.ts
+++ b/packages/utils/src/logger.ts
@@ -1,7 +1,7 @@
 /**
  * Interface of a generic Lodestar logger. For implementations, see `@lodestar/logger`
  */
-export type Logger = Record<LogLevel, LogHandler>;
+export type Logger = Record<Exclude<LogLevel, LogLevel.trace>, LogHandler>;
 
 export enum LogLevel {
   error = "error",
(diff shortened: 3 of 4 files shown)


nflaig:
What happens if you pass `trace` as log level now from the CLI? See requirement mentioned in https://github.com/ChainSafe/lodestar/issues/4536/.

matthewkeil:
> What happens if you pass `trace` as log level now from the CLI? See requirement mentioned in #4536.

It will error as it's unsupported per our logging policy.  Should I throw an error on startup stating that its not supported?

Or perhaps should I leave the functions but throw a better error and kill the process if we use it so we know not to?

nflaig:
As per issue

>This causes Lodestar startup to fail when trace logging is configured.

> That way trace logging can be enabled for a "stack" consisting of Lodestar, an EL, Grafana, etc.

Right now we break this use case, I would suggest we go with what I suggested in https://github.com/ChainSafe/lodestar/issues/5809#issuecomment-1660358713.

> We should remove it from types to prevent accidental usage but still allow it to be passed through CLI to not break the use case mentioned in https://github.com/ChainSafe/lodestar/issues/4536.

dapplion (dismissed):
A runtime error like this is not the best way to handle it

matthewkeil:
> A runtime error like this is not the best way to handle it

I agree.  Was actually just messaging @nflaig about this on discord.  Was tossing around the idea of just having trace log to the debug LogLevel but that isnt great either because the function can still be used...

Another idea would be to have two "log levels" one for reporting with all of them and one for the "supported levels" that is used for interface generation so the classes do not export the symbol.  That way passing --trace as the reporting level just outputs debug level but the intellisense will not have trace as a method.  What do you prefer @dapplion and @nflaig?

nflaig:
We can simply update the logger interface
https://github.com/ChainSafe/lodestar/blob/e690ac6df8b6c3e766fd5c245254575dad061f7b/packages/utils/src/logger.ts#L4

to something like this

```ts
export type Logger = Record<Exclude<LogLevel, LogLevel.trace>, LogHandler>;
```

This will allow to set `--logLevel trace` while also preventing accidental usage in code

wemeetagain:
Yeah, would like to see the solution @nflaig proposed

nflaig (changes requested):
shouldn't this method be removed here as well?

https://github.com/ChainSafe/lodestar/blob/92cca53f7711debf2f21029344a5bcd5ecb037e1/packages/logger/src/winston.ts#L80

nflaig (approved):

wemeetagain (approved):
Lgtm, can you update the PR description with the latest changes / motivation before you merge?

nflaig:
On that note, does this close https://github.com/ChainSafe/lodestar/issues/5809 as well?

The expected behavior in the issue is
> A trace log should work and not break the other log levels

but after discussion, the expected behavior should rather be what I mentioend in https://github.com/ChainSafe/lodestar/pull/5861#issuecomment-1669872677

> This will allow to set --logLevel trace while also preventing accidental usage in code

matthewkeil:
@wemeetagain the PR body is updated but cannot merge because of e2e test not closing cleanly.  Ready to merge though I believe.

wemeetagain:

PR #5862, opened 2023-08-08
fix: move and fix execution engine state update checks for online state

g11tech:
Noticed that PR -  fix: offline error message when node is shutting down #5797  was updating execution state from SYNCED to ONLINE since the checks for the same were not located at correct place which caused excessive logging: here is an example log:

```typescript
ph 0/2 0.022[Node-A/execution] info: Execution client became online oldState=SYNCED, newState=ONLINE
EL 5743: INFO [08-07|18:59:44.030] Imported new potential chain segment     number=2 hash=fd4af3..63ebe2 blocks=1 txs=0 mgas=0.000 elapsed="190.628µs" mgasps=0.000 dirty=698.00B
Eph 0/2 0.031[Node-A/execution] info: Execution client is synced oldState=ONLINE, newState=SYNCED
```

This PR fixes the same

### DIFF
--- a/packages/beacon-node/src/execution/engine/http.ts
+++ b/packages/beacon-node/src/execution/engine/http.ts
@@ -421,6 +421,14 @@ export class ExecutionEngineHttp implements IExecutionEngine {
 
     if (oldState === newState) return;
 
+    // The ONLINE is initial state and can reached from offline or auth failed error
+    if (
+      newState === ExecutionEngineState.ONLINE &&
+      !(oldState === ExecutionEngineState.OFFLINE || oldState === ExecutionEngineState.AUTH_FAILED)
+    ) {
+      return;
+    }
+
     switch (newState) {
       case ExecutionEngineState.ONLINE:
         this.logger.info("Execution client became online", {oldState, newState});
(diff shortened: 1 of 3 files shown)


nflaig (commented):
Do we really want to print out the enum values in all caps? `oldState=SYNCED, newState=ONLINE`

newState is already part of the message and oldState should be logged previously. There are also not that much changes of state where this would be helpful, most of the time it is from syncing to synced.

nazarhussain:
@g11tech That change will make the tests fail, will fix those and update the PR.

nazarhussain:
@nflaig The message is more descriptive, and having extra information and context does not harm, rather useful for log parsing for someone.

nflaig:
> The message is more descriptive, and having extra information and context does not harm

@nazarhussain maybe just add previous state then? The log message already contains the newState hence it is duplicated information. Maybe not all caps would be preferable as well

One other note, on startup we print this

```
info: Execution client is synced oldState=ONLINE, newState=SYNCED
```

This is confusing to me, does this mean the EL was not synced before? I know it doesn't but will the user know?

Before it just sayed `info: Execution client is synced` which was fine since it was rather just the state without any indication of what the previous state might have been

g11tech:
i think the current logging is informative as well as easily understood , we always print out enum in caps i think so i guess that is fine as well

nazarhussain (approved):

g11tech (commented):
looks good to me, @nazarhussain you may approve the PR and merge (I as author can't do the same)

PR #5865, opened 2023-08-08
fix: reset gossipsubDHigh param to 12

wemeetagain:
**Motivation**

- We now are doing less work since https://github.com/ChainSafe/lodestar/issues/5420

**Todo**
- [ ] Test in running node

**Description**

- set `gossipsubDHigh` to 12


dapplion:
CC @tuyennhv

twoeths (approved):
there are not a lot of traffic since we subscribe only 2 subnets since v1.10.0 so I think there is no risk, we're good to go

dapplion (approved):

PR #5866, opened 2023-08-08
chore!: update target to es2021

wemeetagain:
**Motivation**

- We want to produce up-to-date javascript that doesn't have lots of polyfills

**Description**

- update the target to es2021 (es2022 still has poor support according to @nazarhussain )


nazarhussain:
For reference https://kangax.github.io/compat-table/es2016plus/

image

wemeetagain:
FYI the "red" in the attached image only shows polyfill/transpilation support rather than native browser/engine support.
All ES2021 (and below) features we're using are supported in all major browsers and engines.

nflaig:
We have ES version badges in all README files (`README.md` and `packages/*/README.md`), might wanna update those

https://github.com/ChainSafe/lodestar/blob/664820a35c610d86db459bff19493ddd0a1bcfd6/README.md?plain=1#L9

nflaig (approved):

Issue #5852, opened 2023-08-06
Improve recomputeForkChoiceHead


twoeths:
### Describe the bug

In mainnet node it could take up to 0.5s to `recomputeForkChoiceHead`

<img width="1085" alt="Screenshot 2023-08-06 at 15 10 27" src="https://github.com/ChainSafe/lodestar/assets/10568965/f23cce32-037a-4649-946d-4fce2eb27229">

and it takes 6% of cpu in total

<img width="1314" alt="Screenshot 2023-08-06 at 15 22 50" src="https://github.com/ChainSafe/lodestar/assets/10568965/262bfb37-e9d5-42e9-8b09-2d56a1b076ee">

while out benchmark says <= 100ms

see https://github.com/ChainSafe/lodestar/pull/5836#issuecomment-1666767135

### Expected behavior

Get back to the level of v1.9.x

<img width="1298" alt="Screenshot 2023-08-06 at 15 12 01" src="https://github.com/ChainSafe/lodestar/assets/10568965/0357c7e4-92bc-442e-9852-dd321ac937e9">

### Operating system

Linux

### Lodestar version or commit hash

v1.10.0

twoeths:
this could be because we processed so many attestations in v1.10.0 compared to v1.9.2

<img width="797" alt="Screenshot 2023-08-06 at 15 15 45" src="https://github.com/ChainSafe/lodestar/assets/10568965/316aaa14-f274-497b-ac0f-c6ac4237ee1d">

also looks like the benchmark test does not go with this number, need to update it

twoeths:
this is part of the effort to handle more validators coming to ethereum, see #5855

twoeths:
for 2M validator network, this could be up to 8s on 2M validator devnet

<img width="823" alt="Screenshot 2023-08-09 at 10 15 40" src="https://github.com/ChainSafe/lodestar/assets/10568965/8363a6ac-f605-44cc-8203-a162299cd331">

twoeths:
[0810_updateHead_perf_test.cpuprofile.zip](https://github.com/ChainSafe/lodestar/files/12308374/0810_updateHead_perf_test.cpuprofile.zip)

this issue is caused by `getAncestor()` call, we iterate too much from a node until finalized node to check if a node is descendant of finalized node

[...]

PR #5867, opened 2023-08-09
fix: improve forkchoice updateHead() time

twoeths:
**Motivation**

- Performance test of computeDeltas does not reflect the number we see in goerli/mainnet
- Update: was able to reproduce the big updateHead time in `forkChoice updateHead vc 600000 bc 7200 eq 0` test

**Description**

- Update performance tests to reproduce performance issue of "forkchoice.updateHead"
- Write "isFinalizedRootOrDescendant" specifically used for "nodeIsViableForHead()" function to avoid having to loop through the entire nodes a lot of time in "updateHead()"
  - This is exactly the same to https://github.com/sigp/lighthouse/blob/stable/consensus/proto_array/src/proto_array.rs#L975

part of #5852

### DIFF
--- a/packages/fork-choice/src/protoArray/computeDeltas.ts
+++ b/packages/fork-choice/src/protoArray/computeDeltas.ts
@@ -19,12 +19,26 @@ export function computeDeltas(
   newBalances: EffectiveBalanceIncrements,
   equivocatingIndices: Set<ValidatorIndex>
 ): number[] {
-  const deltas = Array.from({length: indices.size}, () => 0);
+  const deltas = Array<number>(indices.size).fill(0);
   const zeroHash = HEX_ZERO_HASH;
   // avoid creating new variables in the loop to potentially reduce GC pressure
   let oldBalance, newBalance: number;
   let currentRoot, nextRoot: string;
   let currentDeltaIndex, nextDeltaIndex: number | undefined;
+  // this function tends to get some very few roots from `indices` so create a small cache to improve performance
+  const cachedIndices = new Map<string, number>();
+
+  const getIndex = (root: string): number | undefined => {
+    let index = cachedIndices.get(root);
+    if (index === undefined) {
+      index = indices.get(root);
+      if (index !== undefined) {
+        cachedIndices.set(root, index);
+      }
+    }
+    return index;
+  };
+
   for (let vIndex = 0; vIndex < votes.length; vIndex++) {
     const vote = votes[vIndex];
     // There is no need to create a score change if the validator has never voted or both of their
@@ -52,7 +66,7 @@ export function computeDeltas(
     if (equivocatingIndices.size > 0 && equivocatingIndices.has(vIndex)) {
       // this function could be called multiple times but we only want to process slashing validator for 1 time
       if (currentRoot !== zeroHash) {
-        currentDeltaIndex = indices.get(currentRoot);
+        currentDeltaIndex = getIndex(currentRoot);
         if (currentDeltaIndex !== undefined) {
           if (currentDeltaIndex >= deltas.length) {
             throw new ProtoArrayError({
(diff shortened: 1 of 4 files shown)


twoeths:
I'm not sure if using common variables before the for loop help in the case lodestar uses so much memory, the performance is the same when I test in goerli. At least it does no harm, may want to consult Ben if it's a necessary change or not

Screenshot 2023-08-09 at 15 42 25

I'm not sure why the perf test is way faster even though I added all votes with different current root / next root. At least I see the benchmark is better, from more than 3ms for 250k validators

Screenshot 2023-08-09 at 15 48 31

to this new value

Screenshot 2023-08-09 at 15 49 07

packages/fork-choice/src/protoArray/computeDeltas.ts
```diff
@@ -77,7 +82,7 @@ export function computeDeltas(
       }
       // We ignore the vote if it is not known in `indices .
       // We assume that it is outside of our tree (ie: pre-finalization) and therefore not interesting
-      const nextDeltaIndex = indices.get(nextRoot);
+      nextDeltaIndex = indices.get(nextRoot);
```
dapplion: Can you create a local cache to resolve root -> index in the context of this function? the 4,000,000 lookups end up adding up I think. It can also cache the absence of a root with null to prevent cache misses in that case too

packages/fork-choice/src/protoArray/computeDeltas.ts
```diff
@@ -21,33 +21,38 @@ export function computeDeltas(
 ): number[] {
   const deltas = Array.from({length: indices.size}, () => 0);
```
dapplion: This can also be improved, it's faster to do a regular for loop than array.from

wemeetagain: What about Array.fill performance?

twoeths:
with https://github.com/ChainSafe/lodestar/pull/5867/commits/0aa5224597cc3e408749c24b993c9a9ccdfdeddb the `forkChoice updateHead vc 600000 bc 7200 eq 0` test time is reduced from 15s/op to 22ms/op in my local environment

wemeetagain (approved):

packages/fork-choice/src/protoArray/computeDeltas.ts
dapplion: Last time i checked regular for loop in the fastest way to populate an array. There should be benchmarks in the state transition package

twoeths:
```
  array creation
Array.from(() => 0): 9.6100125 ms
    ✔ Array.from(() => 0)
Array.from().fill(0): 8.389225 ms
    ✔ Array.from().fill(0)
Array.from(): 15.170741699999999 ms
    ✔ Array.from()
new Array(): 0.8613583 ms
    ✔ new Array()
new Array(); for loop: 1.2474166 ms
    ✔ new Array(); for loop
```

@dapplion I'll do `new Array() and a for loop in` in https://github.com/ChainSafe/lodestar/pull/5882 👍

dapplion: Have you benchmark just pushing to an empty array?

twoeths: not yet, how important is it? in this case there are always more than 0 protonodes in forkchoice

twoeths:
The root cause of this slow down is we call `getAncestor()` to check for descendant of finalized node too many times. If we have `n` node, `applyScoreChanges` will call from `n * getAncestor()` to `2 * n * getAncestor()` with `getAncestor(i)` ranges from 1 to n

The complexity of a function to check if all nodes are ancestor of finalized node (naive way, not the improvement in this PR) is `O(n * (n + 1) / 2)`, so worse case of `applyScoreChanges` would be `O(n * (n + 1))` 

We need to review:
- functions that call `getAncestor()`
- functions that loop through the entire proto array more than 1 time

I opened #5901 #5902 from this review cc @dapplion

PR #5880, opened 2023-08-13
feat: implement produce block v3 and move builder/execution race to beacon

g11tech:
# Breaking Change 
(for annoucements/release notes)
 - builder vs execution block production `race` has been moved to beacon (from validator)  (via both `produceBlockV3` and `produceBlindedBlock`) and hence both validation and beacon will need to be updated
 - ~~validator flag `--builder` is now deprecated and builder flow activation is solely decided upon `--builder.selection`~~
 - See https://github.com/ChainSafe/lodestar/pull/6081
   - propose config now also doesn't need `enabled` flag under `builder` section. if you do not want to use the builder on a validator that might still be hooked onto a builder enabled beacon, you may provide the builder `selection` as `executiononly` (by default its set to `maxprofit` but default still can be overridden with `--builder.selection` flag on the validator apart from the proposer config file level overrides)
     ```yaml
          proposer_config:
            '0xa4855c83d868f772a579133d9f23818008417b743e8447e235d8eb78b1d8f8a9f63f98c551beb7de254400f89592314d':
              fee_recipient: '0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb'
              builder:
                selection: "executiononly"
          default_config:
            fee_recipient: '0xcccccccccccccccccccccccccccccccccccccccc'
            builder:
              selection: "maxprofit"
      ```
   - DVTs which are based off on only builder based block production can still continue using `--builder.selection builderonly` flag on the validator to only trigger block production through the builder url configured on the beacon

----

Part of 
- https://github.com/ChainSafe/lodestar/issues/5823

TODO:
 - [x] keep the breaking flags in but hidden and in no-op/backward compatible mode
 - [x] depreacate usage

Implements :
- [x] switch full and blinded block production to produce block v3
   - [x] add in api routes
   - [x] imp in beacon-node
     - [x] move blinded vs full race to beacon 
   - [x] switch validator to use v3 
~~- [x] move produceblockv2 and produceblindedblock to the standard beacon api~~ reverted after dicusssion in R&D
 - [x] fix tests
   - [x] resolve any tests w.r.t block production
   - [x] move block production race test to beacon node and mock on produceBlockV3
 - [x] house keeping tasks (might need discussion)
    - [x] implement strict fee recipient checks and allow beacon to be configured
    - [x]  allow a way to supply builderSelection i.e. with builderOnly
    - [x] flags cleanup
 - [x] breaking change summary (for announcement/release notes benefit)

### PS: 
this PR doesn't add consensus block value which will be provided in a followup PR


packages/api/src/utils/routes.ts
```diff
@@ -89,7 +52,9 @@ export function AllForksSignedBlindedBlockContentsReqSerializer(
   };
 }
 
-export function AllForksBlindedBlockContentsResSerializer(getType: () => ForkBlobs): TypeJson<BlindedBlockContents> {
+export function AllForksBlindedBlockContentsResSerializer(
+  getType: () => ForkBlobs
+): TypeJson<allForks.BlindedBlockContents> {
```
dapplion: functions should follow camel case with first letter lowercase. If `allForksBlindedBlockContentsResSerializer` is always passed a closure with a static fork value, just pass `fork`

packages/beacon-node/src/api/impl/beacon/blocks/index.ts
```diff
@@ -138,17 +138,89 @@ export function getBeaconBlockApi({
...
+
+      if (signedBlindedBlobSidecars !== null) {
+        Object.assign(logCtx, {blobs: signedBlindedBlobSidecars.length});
+        if (executionPayload === null) {
+          throw Error("Missing locally produced executionPayload for deneb+ publishBlindedBlock");
+        }
+        const blockHash = toHex(executionPayload.blockHash);
+        const {blobSidecars} = chain.producedBlobSidecarsCache.get(blockHash) ?? {};
+        if (blobSidecars === undefined) {
+          throw Error("Missing blobSidecars from the local execution cache");
```
dapplion: Or there's an item in `producedBlobSidecarsCache` that does not have the property `blobSidecars`

packages/beacon-node/src/api/impl/beacon/blocks/index.ts
```diff
@@ -138,17 +138,89 @@ export function getBeaconBlockApi({
     signedBlindedBlockOrContents,
     opts: PublishBlockOpts = {}
   ) => {
-    const executionBuilder = chain.executionBuilder;
-    if (!executionBuilder) throw Error("exeutionBuilder required to publish SignedBlindedBeaconBlock");
-    // Mechanism for blobs & blocks on builder is not yet finalized
+    let signedBlindedBlock: allForks.SignedBlindedBeaconBlock;
+    let signedBlindedBlobSidecars: deneb.SignedBlindedBlobSidecars | null;
```
dapplion: Can you move this logic into a function "toFullBlock" or something similar, and add tests for the different cases. Lots of type casting going on here that could break.

packages/beacon-node/src/chain/interface.ts
```diff
@@ -95,7 +95,7 @@ export interface IBeaconChain {
   readonly checkpointBalancesCache: CheckpointBalancesCache;
   readonly producedBlobSidecarsCache: Map<BlockHash, {blobSidecars: deneb.BlobSidecars; slot: Slot}>;
   readonly producedBlindedBlobSidecarsCache: Map<BlockHash, {blobSidecars: deneb.BlindedBlobSidecars; slot: Slot}>;
-  readonly producedBlockRoot: Set<RootHex>;
+  readonly producedBlockRoot: Map<RootHex, allForks.ExecutionPayload | null>;
```
dapplion, line 97: Add metrics to track the size of this cache. Also for others related to block production added for deneb

packages/state-transition/src/block/processExecutionPayload.ts
```diff
@@ -77,45 +82,3 @@ export function processExecutionPayload(
     .getExecutionForkTypes(state.slot)
     .ExecutionPayloadHeader.toViewDU(payloadHeader) as typeof state.latestExecutionPayloadHeader;
 }
-
-export function executionPayloadToPayloadHeader(
```
dapplion, line 81: Note to self, util moved to `packages/state-transition/src/util/execution.ts`

packages/validator/src/services/block.ts
```diff
@@ -43,16 +33,35 @@ const MAX_DECIMAL_FACTOR = BigInt("100000");
...
-  expectedFeeRecipient: string;
-  strictFeeRecipientCheck: boolean;
-  isBuilderEnabled: boolean;
-  builderSelection: BuilderSelection;
-};
+// const BLOCK_PRODUCTION_RACE_CUTOFF_MS = 2_000;
+// /** Overall timeout for execution and block production apis */
+// const BLOCK_PRODUCTION_RACE_TIMEOUT_MS = 12_000;
+
+type FullOrBlindedBlockWithContents =
```
dapplion, line 33: Add comments explaining what each union type represent, and when they happen

packages/validator/src/services/block.ts
```diff
@@ -183,260 +192,133 @@ export class BlockProposingService {
...
     };
-    const blockFeeRecipient = (fullOrBlindedBlock.block as bellatrix.BeaconBlock).body.executionPayload?.feeRecipient;
-    const feeRecipient = blockFeeRecipient !== undefined ? toHexString(blockFeeRecipient) : undefined;
 
-    if (source === ProducedBlockSource.engine) {
-      if (feeRecipient !== undefined) {
-        if (feeRecipient !== expectedFeeRecipient && strictFeeRecipientCheck) {
-          throw Error(`Invalid feeRecipient=${feeRecipient}, expected=${expectedFeeRecipient}`);
-        }
+    let fullOrBlindedBlockWithContents: FullOrBlindedBlockWithContents;
```
dapplion: Please move the pattern of
```ts
let x;
if () {
  x = 1
} else {
  x = 2
}
```
to
```ts
function get_x() {
  if () {
    return 1
  } else {
    return 2
  }
}
```
and test it

packages/validator/test/unit/services/block.test.ts
```diff
@@ -48,13 +49,19 @@ describe("BlockDutiesService", function () {
     });
 
     const clock = new ClockMock();
-    const blockService = new BlockProposingService(config, loggerVc, api, clock, validatorStore, null);
+    // use produceBlockV3
+    const blockService = new BlockProposingService(config, loggerVc, api, clock, validatorStore, null, true);
```
dapplion: pass an options object instead of just `true`

dapplion:
Mostly cosmetic comments, otherwise looks good

g11tech:
> Mostly cosmetic comments, otherwise looks good

thank you will address the feedback :heart: :pray:

wemeetagain:
Can you update the description with the lastest?

dapplion:
Is this really a breaking change if the flag is set to false by default?

g11tech:
> Is this really a breaking change if the flag is set to false by default?

it is a lodestar cross version  compatibility issue change as the RACE between execution and builder has now moved to beacon (for non produceblockv3 as well), so its both beacon and validator would need to be updated

g11tech:
> Can you update the description with the lastest?

updated

wemeetagain (dismissed):
LGTM, thanks for fixing this PR up

packages/beacon-node/src/api/impl/beacon/blocks/index.ts
dapplion: this does not appear resolved. let, assign on if else pattern still here

packages/beacon-node/src/api/impl/beacon/blocks/index.ts
```diff
@@ -138,17 +138,79 @@ export function getBeaconBlockApi({
...
+      }
+
+      chain.logger.verbose("Publishing block assembled from locally cached payload", logCtx);
+    } else {
+      // Mechanism for blobs & blocks on builder is implemenented separately in a followup deneb-builder PR
+      if (isSignedBlindedBlockContents(signedBlindedBlockOrContents)) {
+        throw Error("exeutionBuilder not yet implemented for deneb+ forks");
+      }
+      const executionBuilder = chain.executionBuilder;
+      if (!executionBuilder) throw Error("exeutionBuilder required to publish SignedBlindedBeaconBlock");
```
dapplion:
```suggestion
      if (!executionBuilder) {
        throw Error("exeutionBuilder required to publish SignedBlindedBeaconBlock");
      }
```

g11tech:
@dapplion further resolved the feedback :+1: :heart:

dapplion (approved):
LGTM!

wemeetagain:

PR #5884, opened 2023-08-14
fix: set user credentials in URL as Authorization header

### LINKED ISSUE #5881: Basic Auth userdata no longer working in validator client as of v1.10.0
### Describe the bug

Using a url of the form `https://user:password@bn.com` no longer works with lodestar, presenting this error:

```info: Waiting for genesis message=Request cannot be constructed from a URL that includes credentials: https://user:password@bn.com```

### Expected behavior

Lodestar should parse the URI per spec and set the `Authorization` header as it did in previous versions

### Steps to reproduce

_No response_

### Additional context

_No response_

### Operating system

Linux

v1.10.0

### PR DESCRIPTION
**Pull request by nflaig**
**Motivation**

Closes https://github.com/ChainSafe/lodestar/issues/5881

**Description**

- Add functions to encode / decode base64
- Move URL validation function to utils package
- Improve URL validation and fail on startup
- Set user credentials in URL as Authorization header

**Note:** Basic authorization header is only added in [HttpClient](https://github.com/ChainSafe/lodestar/blob/f146bbc2cb1f9d54b1c08e60f01c489cfbb9f209/packages/api/src/utils/client/httpClient.ts#L95) and **not** [JsonRpcHttpClient](https://github.com/ChainSafe/lodestar/blob/f146bbc2cb1f9d54b1c08e60f01c489cfbb9f209/packages/beacon-node/src/eth1/provider/jsonRpcHttpClient.ts#L51) as it implements it's own authentication scheme using JWT as bearer tokens.

### DIFF
--- a/packages/api/src/utils/client/httpClient.ts
+++ b/packages/api/src/utils/client/httpClient.ts
@@ -276,7 +276,7 @@ export class HttpClient implements IHttpClient {
     const timer = this.metrics?.requestTime.startTimer({routeId});
 
     try {
-      const url = urlJoin(baseUrl, opts.url) + (opts.query ? "?" + stringifyQuery(opts.query) : "");
+      const url = new URL(urlJoin(baseUrl, opts.url) + (opts.query ? "?" + stringifyQuery(opts.query) : ""));
 
       const headers =
         extraHeaders && opts.headers ? {...extraHeaders, ...opts.headers} : opts.headers || extraHeaders || {};
@@ -285,6 +285,12 @@ export class HttpClient implements IHttpClient {
       }
       if (bearerToken && headers["Authorization"] === undefined) {
         headers["Authorization"] = `Bearer ${bearerToken}`;
+      }
+      if ((url.username || url.password) && headers["Authorization"] === undefined) {
+        headers["Authorization"] = `Basic ${toBase64(`${url.username}:${url.password}`)}`;
+        // Remove the username and password from the URL
+        url.username = "";
+        url.password = "";
       }
 
       this.logger?.debug("HttpClient request", {routeId});
@@ -298,7 +304,7 @@ export class HttpClient implements IHttpClient {
 
       if (!res.ok) {
         const errBody = await res.text();
-        throw new HttpError(`${res.statusText}: ${getErrorMessage(errBody)}`, res.status, url);
+        throw new HttpError(`${res.statusText}: ${getErrorMessage(errBody)}`, res.status, url.toString());
       }
 
       const streamTimer = this.metrics?.streamTime.startTimer({routeId});
--- a/packages/api/src/utils/client/httpClient.ts
+++ b/packages/api/src/utils/client/httpClient.ts
@@ -1,4 +1,4 @@
-import {ErrorAborted, Logger, TimeoutError} from "@lodestar/utils";
+import {ErrorAborted, Logger, TimeoutError, isValidHttpUrl, toBase64} from "@lodestar/utils";
 import {ReqGeneric, RouteDef} from "../index.js";
 import {ApiClientResponse, ApiClientSuccessResponse} from "../../interfaces.js";
 import {fetch, isFetchError} from "./fetch.js";
@@ -123,8 +123,15 @@ export class HttpClient implements IHttpClient {
 
     // opts.baseUrl is equivalent to `urls: [{baseUrl}]`
     // unshift opts.baseUrl to urls, without mutating opts.urls
-    for (const urlOrOpts of [...(baseUrl ? [baseUrl] : []), ...urls]) {
+    for (const [i, urlOrOpts] of [...(baseUrl ? [baseUrl] : []), ...urls].entries()) {
       const urlOpts: URLOpts = typeof urlOrOpts === "string" ? {baseUrl: urlOrOpts, ...allUrlOpts} : urlOrOpts;
+
+      if (!urlOpts.baseUrl) {
+        throw Error(`HttpClient.urls[${i}] is empty or undefined: ${urlOpts.baseUrl}`);
+      }
+      if (!isValidHttpUrl(urlOpts.baseUrl)) {
+        throw Error(`HttpClient.urls[${i}] must be a valid URL: ${urlOpts.baseUrl}`);
+      }
       // De-duplicate by baseUrl, having two baseUrls with different token or timeouts does not make sense
       if (!this.urlsOpts.some((opt) => opt.baseUrl === urlOpts.baseUrl)) {
         this.urlsOpts.push(urlOpts);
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: packages/api/src/utils/client/httpClient.ts [hunk 1]
```diff
@@ -279,6 +286,14 @@ export class HttpClient implements IHttpClient {
       if (bearerToken && headers["Authorization"] === undefined) {
         headers["Authorization"] = `Bearer ${bearerToken}`;
       }
+      if (url.username || url.password) {
+        if (headers["Authorization"] === undefined) {
+          headers["Authorization"] = `Basic ${toBase64(`${url.username}:${url.password}`)}`;
+        }
+        // Remove the username and password from the URL
+        url.username = "";
+        url.password = "";
```
nazarhussain, line 295: Would prefer to completely remove these fields. 

```suggestion
        // Remove the username and password from the URL
        delete url.username;
        delete url.password;
```

packages/api/src/utils/client/httpClient.ts
```diff
@@ -279,6 +286,14 @@ export class HttpClient implements IHttpClient {
       if (bearerToken && headers["Authorization"] === undefined) {
         headers["Authorization"] = `Bearer ${bearerToken}`;
       }
+      if (url.username || url.password) {
+        if (headers["Authorization"] === undefined) {
+          headers["Authorization"] = `Basic ${toBase64(`${url.username}:${url.password}`)}`;
+        }
```
nazarhussain, line 292: Throw an error if the user sets the `Authorization` header attribute as well as the `username/password` as options.

packages/api/src/utils/client/httpClient.ts
```diff
@@ -279,6 +286,14 @@ export class HttpClient implements IHttpClient {
       if (bearerToken && headers["Authorization"] === undefined) {
         headers["Authorization"] = `Bearer ${bearerToken}`;
       }
+      if (url.username || url.password) {
```
nazarhussain, line 289: This should be a compound condition for both, we can't generate a valid Authorization header otherwise.

packages/api/src/utils/client/httpClient.ts
```diff
@@ -269,7 +276,7 @@ export class HttpClient implements IHttpClient {
     const timer = this.metrics?.requestTime.startTimer({routeId});
 
     try {
-      const url = urlJoin(baseUrl, opts.url) + (opts.query ? "?" + stringifyQuery(opts.query) : "");
+      const url = new URL(urlJoin(baseUrl, opts.url) + (opts.query ? "?" + stringifyQuery(opts.query) : ""));
```
nazarhussain, line 279: This function `stringifyQuery` uses a third-party library. See if we can replace it with the native `node:querystring` module.

packages/api/src/utils/client/httpClient.ts
```diff
@@ -279,6 +286,14 @@ export class HttpClient implements IHttpClient {
       if (bearerToken && headers["Authorization"] === undefined) {
         headers["Authorization"] = `Bearer ${bearerToken}`;
       }
+      if (url.username || url.password) {
+        if (headers["Authorization"] === undefined) {
+          headers["Authorization"] = `Basic ${toBase64(`${url.username}:${url.password}`)}`;
```
nazarhussain, line 291: If we can move this logic to a utility function `generateAuthHeader(username, password)`, that will be easy to manage in future.

nazarhussain (dismissed):

packages/api/src/utils/client/httpClient.ts
nflaig: TypeScript won't let me `The operand of a 'delete' operator must be optional.`

And I think this good because downstream code might assume this is a defined string as type is `URL.username: string`, likely wouldn't cause issues but I prefer to not mess with the type if I pass it to third party code.

If you create a URL object without credentials you also get empty strings

```node
new URL("https://google.com")
URL {
  href: 'https://google.com/',
  origin: 'https://google.com',
  protocol: 'https:',
  username: '',
  password: '',
  host: 'google.com',
  hostname: 'google.com',
  port: '',
  pathname: '/',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

packages/api/src/utils/client/httpClient.ts
nflaig: I was thinking about it but after looking at other http client implementations (e.g. native node http) I figured usually they just have a defined order of precedence, and we don't do that for bearer token either.

From [node docs](https://nodejs.org/api/http.html):
> Sending an Authorization header will override using the auth option to compute basic authentication.


In our case it would be: explicit header > bearer header > basic auth header

As another side note, if we want to throw an error, this can not be implemented here as it would lazily throw the first time a request is done which is not ideal. We have to throw misconfiguration errors on startup. This is why I also added URL validation in the constructor of the http client, these runtime errors are bad, especially if you use fallback URLs which might not be queried for hours / days and when they do you notice that the URL is invalid.

packages/api/src/utils/client/httpClient.ts
nflaig: As per RFC (https://datatracker.ietf.org/doc/html/rfc2617) those are optional and can be omitted (`*`, i.e. 0 or more).

```
     user-pass   = userid ":" password
      userid      = *<TEXT excluding ":">
      password    = *TEXT
```

In both cases fetch also throws an error as it does not allow credentials in the URL

```
fetch("http://user@localhost")
TypeError: Request cannot be constructed from a URL that includes credentials: http://user@localhost
```

```
fetch("http://:password@localhost")
> Uncaught:
TypeError: Request cannot be constructed from a URL that includes credentials: http://:password@localhost
```

packages/api/src/utils/client/httpClient.ts
nflaig: We can't as it does not support all formats required to be openapi compliant.

Specifically we want the "repeat" format like this `topic=topic1&topic=topic2`, although I think comma-separated should be supported by all clients as well

https://github.com/ChainSafe/lodestar/blob/f9c71074df49c4bdd1b01d8ac434719fda1a1dbf/packages/api/src/utils/client/format.ts#L5-L7

packages/api/src/utils/client/httpClient.ts
nflaig: Was considering that as well but opted against it because
- not reused anywhere else
- simple enough
- format of auth header as defined in RFC will never change

Same argument could be made for setting bearer token above but I think the code is clearer as is.

packages/api/src/utils/client/httpClient.ts
nflaig: As another side note, we use the same library for parsing, last time I looked at this, there was no native alternative for it and `qs` was the best library I could find

https://github.com/ChainSafe/lodestar/blob/1ff56f6ab3b74fa9998639f0f6c8773fb8181078/packages/beacon-node/src/api/rest/base.ts#L51-L58

wemeetagain (approved):

PR #5886, opened 2023-08-14
feat: add yamux stream muxer

### LINKED ISSUE #5804: Add yamux as first streamMuxer
### Problem description

Beacon chain spec expects implementations to support yamux. 

Yamux is a better multiplexer than yamux per-spec as it supports per stream back-pressure. However Javascript mplex implementation is more battle tested and has served us well, while yamux is new and may have bugs or performance issues

### Solution description

`@chainsafe/libp2p-yamux` is fully developed now, we should attempt to incorporate it by heavily testing it

- See some testing notes in: https://github.com/ChainSafe/lodestar/pull/4845

### Additional context

_No response_

wemeetagain:
**Motivation**

- Since https://github.com/sigp/lighthouse/pull/4431, lighthouse doesn't support mplex. In order to peer with other clients, we need yamux
- Resolve https://github.com/ChainSafe/lodestar/issues/5804

**Description**

- Add yamux as first stream muxer


twoeths:
some monitors on this branch:

- `rss` is higher than `unstable` (only 9.5GB to 10.5GB - 1k node) while heap memories are the same

Screenshot 2024-01-03 at 15 40 31

- `gc` rate is a little bit (~0.1%) higher than `stable/unstable` (1% to 1.2% - 1k node)

Screenshot 2024-01-03 at 15 43 29

Some other metrics are the same:
- % head late (>4s)
- Block source (should be mostly from gossip)
- number of mesh peers on `1k` is better than that of stable, but on `*16` is a little bit worse. I'd say this metric is the same as well
- missed attestations ratio
- gossip block received delay
- gossipsub forward messages/peers

I think it's pretty close, if we can resolve the issue in `rss` then it's good to go

wemeetagain:
@tuyennhv this has been running for the past 4 days on feat3

Seems to be similar results, specifically, rss is still higher, 10.5 vs 9.5

philknows:
Can we do another round of this testing? Especially docker. Adding to v1.20.

philknows:
Based on standup today, needs another round of testing. We need to deploy this one also for testing via Docker deploy. Higher priority given to this over snappyjs due to PRs being pushed to deprecate mplex for yamux. Tagged for v1.20 target if it can make it in. No longer a draft to signal for testing/reviews.

twoeths:
metrics look all good to me
rss is just the same to `feat1

twoeths (approved):

wemeetagain:
:tada: This PR is included in [v1.21.0](https://github.com/ChainSafe/lodestar/releases/tag/v1.21.0) :tada:

PR #5895, opened 2023-08-17
chore: update discv5 and varint

wemeetagain:
- Most of our use of varint now moved to faster uint8-varint v2.x dependency


packages/reqresp/src/encodingStrategies/sszSnappy/encode.ts
```diff
@@ -1,4 +1,4 @@
-import varint from "varint";
+import {encode} from "uint8-varint";
```
nflaig: Would prefer for imports of a library to be consistent across different files.

I am guessing there is no default export, could do this instead

```ts
import * as varint from "uint8-varint";
```

or have it renamed consistently to `varintEncode`

nflaig (approved):

